PR #7226, opened 2024-08-18
[WIP] Datastore UI

abey79:
### What

Some in-viewer UI for the `ChunkStore` and its content.

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4457,6 +4457,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "re_datastore_ui"
+version = "0.19.0-alpha.1+dev"
+dependencies = [
+ "egui",
+ "re_chunk_store",
+ "re_data_ui",
+ "re_entity_db",
+ "re_types",
+ "re_ui",
+ "re_viewer_context",
+]
+
+[[package]]
 name = "re_dev_tools"
 version = "0.19.0-alpha.1+dev"
 dependencies = [
@@ -5254,6 +5267,7 @@ dependencies = [
  "re_data_loader",
  "re_data_source",
  "re_data_ui",
+ "re_datastore_ui",
  "re_edit_ui",
  "re_entity_db",
  "re_error",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4461,9 +4461,14 @@ name = "re_datastore_ui"
 version = "0.19.0-alpha.1+dev"
 dependencies = [
  "egui",
+ "egui_extras",
+ "itertools 0.13.0",
  "re_chunk_store",
  "re_data_ui",
  "re_entity_db",
+ "re_format",
+ "re_format_arrow",
+ "re_log_types",
  "re_types",
  "re_ui",
  "re_viewer_context",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,7 @@ re_tuid = { path = "crates/utils/re_tuid", version = "=0.19.0-alpha.1", default-
 re_blueprint_tree = { path = "crates/viewer/re_blueprint_tree", version = "=0.19.0-alpha.1", default-features = false }
 re_context_menu = { path = "crates/viewer/re_context_menu", version = "=0.19.0-alpha.1", default-features = false }
 re_data_ui = { path = "crates/viewer/re_data_ui", version = "=0.19.0-alpha.1", default-features = false }
+re_datastore_ui = { path = "crates/viewer/re_datastore_ui", version = "=0.19.0-alpha.1", default-features = false }
 re_edit_ui = { path = "crates/viewer/re_edit_ui", version = "=0.19.0-alpha.1", default-features = false }
 re_renderer = { path = "crates/viewer/re_renderer", version = "=0.19.0-alpha.1", default-features = false }
 re_renderer_examples = { path = "crates/viewer/re_renderer_examples", version = "=0.19.0-alpha.1", default-features = false }
--- a/crates/store/re_chunk_store/src/store.rs
+++ b/crates/store/re_chunk_store/src/store.rs
@@ -479,6 +479,18 @@ impl ChunkStore {
         self.chunks_per_chunk_id.values()
     }
 
+    /// Get a chunk based on its ID.
+    #[inline]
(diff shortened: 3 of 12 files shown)


nikolausWest:
Nice!

Small nit:
Screenshot 2024-08-26 at 11 06 15
It took me a while to understand that this meant "number of rows" and not "row number"

abey79:
> Nice!
> 
> Small nit: Screenshot 2024-08-26 at 11 06 15 It took me a while to understand that this meant "number of rows" and not "row number"

Yeah the reasoning is that the actually content is narrow, so it's annoying to have a wide column header. Let's see if we can figure out something better with @gavrelina

nikolausWest:
> Yeah the reasoning is that the actual column content is narrow (just a number), so it's annoying to have a wide column header. 

`# rows` would have done it for me (although 2 chars longer)

crates/store/re_chunk_store/src/query.rs
```diff
@@ -23,6 +23,43 @@ use crate::RowId;
 // as both static and temporal, which is probably wrong.
 
 impl ChunkStore {
+    /// Retrieve all timelines contained in at least one chunk.
```
teh-cmc, line 26:
```suggestion
    /// Retrieve all [`Timeline`]s in the store.

File: crates/store/re_chunk_store/src/query.rs [hunk 2]
```diff
@@ -23,6 +23,43 @@ use crate::RowId;
...
+    pub fn all_timelines(&self) -> BTreeSet<Timeline> {
+        self.temporal_chunk_ids_per_entity
+            .values()
+            .flat_map(|per_timeline| per_timeline.keys().copied())
+            .collect()
+    }
+
+    /// Retrieve all entities contained in the store.
```
teh-cmc, line 34:
```suggestion
    /// Retrieve all [`EntityPath`]s in the store.

File: crates/store/re_chunk_store/src/query.rs [hunk 3]
```diff
@@ -23,6 +23,43 @@ use crate::RowId;
...
+    pub fn all_entities(&self) -> BTreeSet<EntityPath> {
+        self.static_chunk_ids_per_entity
+            .keys()
+            .cloned()
+            .chain(self.temporal_chunk_ids_per_entity.keys().cloned())
+            .collect()
+    }
+
+    // Retrieve all components contained in at least one chunk.
```
teh-cmc, line 43:
```suggestion
    /// Retrieve all [`ComponentName`]s in the store.

File: crates/viewer/re_chunk_store_ui/src/chunk_store_ui.rs [hunk 4]
```diff
@@ -0,0 +1,446 @@
...
+        if self.chunk_list_sort_column.direction == SortDirection::Descending {
+            chunks.reverse();
+        }
+
+        //
+        // Copy to clipboard
+        //
+
+        if should_copy_chunk {
+            let s = chunks.iter().map(|chunk| chunk.to_string()).join("\n\n");
```
teh-cmc, line 236:
```suggestion
        if should_copy_chunk {
            // TODO: make sure the output is not dependent on the parent terminal's width
            let s = chunks.iter().map(|chunk| chunk.to_string()).join("\n\n");

File: crates/viewer/re_chunk_store_ui/src/arrow_ui.rs [hunk 5]
```diff
@@ -0,0 +1,54 @@
+use re_chunk_store::external::re_chunk::external::arrow2;
+use re_chunk_store::external::re_chunk::external::arrow2::array::Utf8Array;
+use re_viewer_context::UiLayout;
+
+//TODO(ab): this is copied/modified from `re_data_ui`. Consider unifying them?
```
teh-cmc, line 5: Also we really need to setup a registry for lossless text representations of arrow payloads based on semantic types.

E.g. I want `rerun.components.Color` to display as `0xFF88AAFF`, not the unintelligible integer it prints today.
And that's not specific to this view, it's everywhere we represent arrow as just lossless text (including terminal dumps).

teh-cmc (approved):
That's amazing :bow: 

---

UPDATE: I found the cause, see code comments. My recommendation is simply to add a TODO, maybe open an issue. The root of the issue that the chunk will be displated with comfyterm, which will use the width of the parent terminal as a hint.

When you click the copy-as-text button, I'm not quite sure how the implementation decides what width should each column be? The current value is annoyingly small (a `RowId` doesn't even fit on a single line):
image

---

The fact that I'm wasting all this vertical space at the top even when I've uncollapsed everything is annoying:
image

---

The timestamp ranges in the chunk list could use an indication of the time spanned, e.g.
```
#0..=#1 255 (1 255 ticks) | #103..=#2 613 (2510 ticks) | 2024-08-26 12:38:33.536540Z..=2024-08-26 12:38:33.740418Z (0.203878 seconds)`
```
image

---

<sub><sup><i>they see me browsin', they hatin'</i></sup></sub>

crates/viewer/re_chunk_store_ui/src/arrow_ui.rs
abey79: Agreed.

Alternatively, I would eventually like to have a toggle to apply the entire `re_data_ui` machinery for display. That could be useful eg. for interpreting enums (can those eventually be fully arrow?). That feature would have to be behind a cargo flag, such that `re_chunk_store_ui` can remain independent from the rerun viewer, should we want to make it its own binary.

abey79:
> The fact that I'm wasting all this vertical space at the top even when I've uncollapsed everything is annoying:

Yes, that's was an unwanted consequence of the fancy sort arrow ui. I've reverted to a less nice UI that doesn't cause the issue. This will be further addressed later with:
- https://github.com/emilk/egui/issues/5015

abey79:
New issue for the width of the copied text:
- https://github.com/rerun-io/rerun/issues/7282

emilk:
* ⚠️ This reverted the changes in https://github.com/rerun-io/rerun/pull/7228 !

Remember to always review changes to `Cargo.lock`!

And if you have merge conflicts in `Cargo.lock`, just do `git checkout main Cargo.lock && cargo check`

Issue #2340, opened 2023-06-08
Support images in BGR and BGRA


abey79:
**Describe the annoyance**

There is generally no consensus between RGB and BGR image formats. For example, Pillow's `PIL.Image.open()` yields RGB format (including when the image is converted to a Numpy array). OpenCV's `cv2.imread()` or `VideoCapture` instead yield BGR numpy arrays.

`rr.log_image()` expects RGB and yields false colours otherwise. ~~At the very least, this should be documented.~~ Ideally, some optional `format` argument should be added so alternative formats may be correctly ingested.

**To Reproduce**
This code yields false color (e.g. blue skins):

```python
import cv2
import rerun as rr

rr.init("demo")
rr.spawn()
image = cv2.imread("image.jpg")
rr.log_image("image", image)
```

This is fixed by adding explicit format conversion:
```python
# ... 
image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
rr.log_image("image", image)
```

Ideally, something like that would be possible:
```python
rr.log_image("image", image, format=rr.BGR)
```

abey79:
FWIW `np.array(mediapipe.Image.create_from_file("image.jpg").numpy_view())` is RGB

emilk:
This is related to adding more semantics to color tensors in general:

* RGB, BGR, RGBA, …
* Linear vs gamma
* Data range (0-255, 0-65535, …)
* Color map

emilk:
The quick-fix is just to reorder the data into RGB on the SDK side.

The deeper fix is adding a "ColorSemantics" component or similar.

abey79:
`rr.log_image` is no more, but the same applies for the `rr.Image` archetype.

jleibs:
Note some places in the code indicate we should change our defaults to BGR.

I do NOT think we should do this. BGR is a bad, legacy choice from opencv. It should not be our default.

emilk:
The simple way to implement BGR and BGRA right now is to upload them as RGB textures, and add a `swap_red_blue: bool` flag to the `rectangle.wgsl` shader

emilk:
Blocked on
* https://github.com/rerun-io/rerun/issues/7100

Wumpf:
not blocked by it, we can have bgr ofc without it being editable

PR #7238, opened 2024-08-20
BGR(A) image format support

Wumpf:
### What

* Fixes https://github.com/rerun-io/rerun/issues/2340

Adds a checklist item that goes through all the different format to make sure they run properly - this looks a bit overzealous, but in fact we are using the fact that wgpu has special support for 8bit bgr images in order to support them on meshes (this detail is tbf untested as of now!), so going through our formats in a test has quite some merit.
(aaalso I think it's the only place where we actually check now that all datatypes still work fine 😳 )

image
("ethnically sourced lena picture", courtesy of Morten Rieger, https://mortenhannemose.github.io/lena/) 

Speaking of meshes: the mesh texture support story is now spelled out better, showing warnings and ignoring textures if they're in an unsupported format.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7238?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7238?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7238)

### DIFF
--- a/crates/viewer/re_viewer_context/src/gpu_bridge/image_to_gpu.rs
+++ b/crates/viewer/re_viewer_context/src/gpu_bridge/image_to_gpu.rs
@@ -114,7 +114,7 @@ fn color_image_to_gpu(
 
     let color_mapper = if let Some(shader_decoding) = shader_decoding {
         match shader_decoding {
-            // We only have 1D color mapps, therefore chroma downsampled and BGR formats can't have color maps.
+            // We only have 1D color maps, therefore chroma downsampled and BGR formats can't have color maps.
             ShaderDecoding::Bgr | ShaderDecoding::Nv12 | ShaderDecoding::Yuy2 => {
                 ColorMapper::OffRGB
             }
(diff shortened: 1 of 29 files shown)


crates/viewer/re_viewer_context/src/gpu_bridge/image_to_gpu.rs
```diff
@@ -116,7 +114,10 @@ fn color_image_to_gpu(
 
     let color_mapper = if let Some(shader_decoding) = shader_decoding {
         match shader_decoding {
-            ShaderDecoding::Nv12 | ShaderDecoding::Yuy2 => ColorMapper::OffRGB,
+            // We only have 1D color mapps, therefore chroma downsampled and BGR formats can't have color maps.
```
emilk, line 117:
```suggestion
            // We only have 1D color maps, therefore chroma downsampled and BGR formats can't have color maps.
```

crates/viewer/re_viewer_context/src/gpu_bridge/image_to_gpu.rs
```diff
@@ -248,19 +271,31 @@ pub fn texture_creation_desc_from_color_image<'a>(
...
             // use a "Srgb-aware" texture like `Rgba8UnormSrgb` for RGBA.
             (ColorModel::RGB, ChannelDatatype::U8) => (
                 pad_rgb_to_rgba(&image.buffer, u8::MAX).into(),
                 TextureFormat::Rgba8Unorm,
             ),
 
+            (ColorModel::BGR, ChannelDatatype::U8) => {
+                (
+                    pad_rgb_to_rgba(&image.buffer, u8::MAX).into(),
+                    TextureFormat::Bgra8Unorm,
```
emilk, line 287: Is using the `Bgra` texture even a good idea, or is it just adding complexity?

examples/python/rgbd/rgbd.py
```diff
@@ -85,8 +83,8 @@ def log_nyud_data(recording_path: Path, subset_idx: int, frames: int) -> None:
 
             if f.filename.endswith(".ppm"):
                 buf = archive.read(f)
-                img_rgb = read_image_rgb(buf)
-                rr.log("world/camera/image/rgb", rr.Image(img_rgb).compress(jpeg_quality=95))
+                img_bgr = read_image_bgr(buf)
+                rr.log("world/camera/image/rgb", rr.Image(img_bgr, color_model="BGR").compress(jpeg_quality=95))
```
emilk, line 87: Does the `.compress` handle BGR too?

emilk (approved):
⭐

crates/viewer/re_viewer_context/src/gpu_bridge/image_to_gpu.rs
Wumpf: from the pov of the all knowing rectangle shader it's just extra complexity, so I should add a comment for the rationale of doing this anyways!

The reason is that this allow us to use BGR(A) on meshes and other occasions where we today don't have fancy decoding etc..
This is very very subtly hinted at by the behavior of `required_shader_decode` which the mesh visualizer uses to determine if it should warn the user that it can't consume a given image.
(in the glorious 🌈 future 🌈  re_renderer will have an internal conversion pipeline that will allow us to do texture conversions separately and gpu accelerated on the fly for all types. But we live in the now and gotta squeeze out the 80% with what we have 😄)

Wumpf: for the code, cleaned up and more formalized :)
```
            // Make use of wgpu's BGR(A)8 formats.
            //
            // From the pov of our on-the-fly decoding textured rect shader this is just a strange special case
            // given that it already has to deal with other BGR(A) formats.
            //
            // However, we have other places where we don't have the luxury of having a shader that can do the decoding for us.
            // In those cases we'd like to support as many formats as possible without decoding.
            //
            // (in some hopefully not too far future, re_renderer will have an internal conversion pipeline
            // that injects on-the-fly texture conversion from source formats before the consumer of a given texture is run
            // and caches the result alongside with the source data)
            //
            // See also [`required_shader_decode`] which lists this case as a format that does not need to be decoded.
```

PR #7261, opened 2024-08-23
Web-only video support

jprochazk:
### What

- Adds web-only video decoding functionality
  - Supported container formats: `mp4`
  - Supported encoding formats: `av1`

Proof of concept, much polish needed here (grep for `TODO(jan)` in this PR).

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7261?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7261?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7261)

### Testing
(EDIT by emilk)

* Download https://github.com/rerun-io/video-decode-test/blob/main/data/bbb_video_av1_frag.mp4
* `pixi run rerun-web bbb_video_av1_frag.mp4`


crates/viewer/re_space_view_spatial/src/visualizers/videos.rs
```diff
@@ -0,0 +1,236 @@
...
+                let texture = video.get_frame(timestamp_s);
+
+                let world_from_entity = ent_context
+                    .transform_info
+                    .single_entity_transform_required(ctx.target_entity_path, "Video");
+                // TODO(jan): texture extents are wrong, this only shows roughly the top-left corner of the video
+                let textured_rect = TexturedRect {
+                    top_left_corner_position: world_from_entity.transform_point3(Vec3::ZERO),
+                    extent_u: world_from_entity.transform_vector3(Vec3::X * video.width() as f32),
+                    extent_v: world_from_entity.transform_vector3(Vec3::Y * video.height() as f32),
```
emilk, line 212: I think there are videos with non-square pixels, which means there is probably an aspect ratio to take into account here somewhere (and for the bounding box below)

crates/viewer/re_space_view_spatial/src/video_cache.rs
```diff
@@ -0,0 +1,79 @@
...
+pub struct VideoCacheKey {
+    pub versioned_instance_path_hash: VersionedInstancePathHash,
+    pub query_result_hash: Hash64,
+    pub media_type: Option<MediaType>,
+}
+
+struct Entry {
+    used_this_frame: AtomicBool,
+    // TODO(jan): warn_once instead of having option here
+    //            who cares if it wastes cycles loading the video
```
emilk, line 24: I think the `Option` is the right approach; otherwise `Video::load` will be called every frame on failure, which could be a huge CPU drain

crates/viewer/re_space_view_spatial/src/video_cache.rs
```diff
@@ -0,0 +1,79 @@
...
+        re_tracing::profile_function!();
+
+        let entry = self.0.entry(key).or_insert_with(|| {
+            re_log::debug!("Loading video {name:?}…");
+
+            let result = Video::load(render_ctx, media_type, video_data);
+            let video = match result {
+                Ok(video) => Some(Arc::new(Mutex::new(video))),
+                Err(err) => {
+                    re_log::warn!("Failed to load video {name:?}: {err}");
```
emilk, line 50:
```suggestion
                    re_log::warn_once!("Failed to load video {name:?}: {err}");
```

Otherwise we repeat ourself after cache eviction

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,526 @@
...
+
+impl Video {
+    pub fn load(
+        render_context: &RenderContext,
+        media_type: Option<&str>,
+        data: &[u8],
+    ) -> Result<Self, VideoError> {
+        let data = match media_type {
+            Some("video/mp4") => re_video::load_mp4(data)?,
+            _ => return Err(VideoError::Load(VideoLoadError::UnknownMediaType)),
```
emilk, line 22:
```suggestion
            _ => return Err(VideoError::Load(VideoLoadError::UnsupportedMediaType(media_type))),
```

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,526 @@
...
+    }
+
+    pub fn height(&self) -> u32 {
+        self.decoder.height()
+    }
+
+    pub fn get_frame(&mut self, timestamp_s: f64) -> GpuTexture2D {
+        re_tracing::profile_function!();
+        self.decoder.get_frame(TimeMs::new(timestamp_s * 1e3))
+    }
```
emilk, line 44: what happens on out-of-bounds?

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,526 @@
...
+        self.decoder.duration_ms()
+    }
+
+    pub fn width(&self) -> u32 {
+        self.decoder.width()
+    }
+
+    pub fn height(&self) -> u32 {
+        self.decoder.height()
+    }
```
emilk, line 39: We probably need to consider non-square pixels here, unless the web video decoder abstracts that away for us

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,526 @@
...
+
+            if &key(&v[middle as usize]) <= needle {
+                left = middle + 1;
+            } else {
+                right = middle;
+            }
+        }
+
+        left - 1
+    }
```
emilk, line 119: You could use https://doc.rust-lang.org/stable/std/primitive.slice.html#method.partition_point here

crates/viewer/re_renderer/src/renderer/video.rs
jprochazk: For OOB in the negatives, returns either a blank texture or whatever the previously shown thing is. That could probably be fixed to clamp to `0`.
For OOB in the positives (>video duration), it returns the last frame.

crates/viewer/re_renderer/src/renderer/video.rs
jprochazk: I think it does? The pixel aspect ratio seems to be codec-specific information, which we pass directly to the video decoder. It handles format conversion (e.g. YUV -> RGB), scaling, etc., stuff that we also likely won't have to worry about with ffmpeg/gstreamer

rerun_cpp/src/rerun/archetypes/asset_video_ext.cpp
```diff
@@ -0,0 +1,81 @@
...
+        file.seekg(0, std::ios::end);
+        std::streampos length = file.tellg();
+        file.seekg(0, std::ios::beg);
+
+        std::vector<uint8_t> data(static_cast<size_t>(length));
+        file.read(reinterpret_cast<char*>(data.data()), length);
+
+        return AssetVideo::from_bytes(
+            Collection<uint8_t>::take_ownership(std::move(data)),
+            AssetVideo::guess_media_type(path)
```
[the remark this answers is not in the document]
jprochazk: will leave that out of this PR: https://github.com/rerun-io/rerun/issues/7280

jprochazk:
Related: [gfx-rs/wgpu#6170](https://www.github.com/gfx-rs/wgpu/pull/6170)

crates/store/re_types/src/archetypes/asset_video.rs
```diff
@@ -19,6 +19,10 @@ use ::re_types_core::{ComponentBatch, MaybeOwnedComponentBatch};
 use ::re_types_core::{DeserializationError, DeserializationResult};
 
 /// **Archetype**: A video file.
+///
+/// NOTE: Videos may only be visualized in the web viewer for now
+///
+/// ⚠️ **This type is experimental and may be removed in future versions**
```
jprochazk, line 25: I tried doing something more fancy and gave up, is this good enough? :sweat_smile:

crates/viewer/re_renderer/Cargo.toml
```diff
@@ -97,6 +99,17 @@ getrandom = { workspace = true, features = [
...
+  "EncodedVideoChunk",
+  "EncodedVideoChunkInit",
+  "EncodedVideoChunkType",
+  "VideoFrame",
+  "VideoDecoder",
+  "VideoDecoderConfig",
+  "VideoDecoderInit",
+] }
+js-sys = { workspace = true }
+wasm-bindgen = { workspace = true }
```
emilk, line 112: sort

crates/store/re_video/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    /// Natural height of the video.
+    pub coded_height: u16,
+
+    /// Natural width of the video.
+    pub coded_width: u16,
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TimeMs(OrderedFloat<f64>);
```
emilk, line 70: What's the purpose of using this over `std::time::Duration`?

We generally try to keep things Rust idiomatic (Duration) and/or SI (seconds)

crates/viewer/re_renderer/src/renderer/video.rs
emilk: DVD:s used non-square pixels, but I don't expect anyone to log a DVD video to Rerun.
Let's just punt on this until it becomes a problem, if ever.

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+        self.decoder.width()
+    }
+
+    pub fn height(&self) -> u32 {
+        self.decoder.height()
+    }
+
+    /// Returns a texture with the latest frame at the given timestamp.
+    ///
+    /// If the timestamp is negative, a zeroed texture is returned.
```
emilk, line 48: It might be worth noting here why this takes `&mut self`

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
+use re_video::VideoLoadError;
+
+use crate::resource_managers::GpuTexture2D;
+use crate::wgpu_resources::GpuTexturePool;
+use crate::wgpu_resources::TextureDesc;
+use crate::RenderContext;
+
+use re_video::TimeMs;
+
+pub struct Video {
```
emilk, line 10: Docstring

crates/store/re_video/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+            .field("segments", &self.segments)
+            .field("data", &self.data.len())
+            .finish()
+    }
+}
+
+impl std::fmt::Debug for Segment {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Segment")
+            .field("timestamp", &self.timestamp)
```
emilk, line 138: Same here: `timestamp: 42` could be quite confusing

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+                return self.texture.clone();
+            };
+
+            let Some(sample_idx) = latest_at_idx(
+                &self.data.segments[segment_idx].samples,
+                |sample| sample.timestamp,
+                &timestamp,
+            ) else {
+                // segments are never empty
+                unreachable!();
```
emilk, line 303: An empty input is not the only time `latest_at_idx` returns `None` though

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+        assert_eq!(latest_at_idx(&v, |v| *v, &3), Some(2));
+        assert_eq!(latest_at_idx(&v, |v| *v, &4), Some(3));
+        assert_eq!(latest_at_idx(&v, |v| *v, &5), Some(4));
+        assert_eq!(latest_at_idx(&v, |v| *v, &6), Some(5));
+        assert_eq!(latest_at_idx(&v, |v| *v, &7), Some(6));
+        assert_eq!(latest_at_idx(&v, |v| *v, &8), Some(7));
+        assert_eq!(latest_at_idx(&v, |v| *v, &9), Some(8));
+        assert_eq!(latest_at_idx(&v, |v| *v, &10), Some(9));
+        assert_eq!(latest_at_idx(&v, |v| *v, &11), Some(9));
+        assert_eq!(latest_at_idx(&v, |v| *v, &1000), Some(9));
```
emilk, line 136: This never tests a needle that is between two elements, which is a key feature of latest-at

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+        }
+
+        pub fn height(&self) -> u32 {
+            self.data.config.coded_height as u32
+        }
+
+        pub fn get_frame(&mut self, timestamp: TimeMs) -> GpuTexture2D {
+            if timestamp.as_f64() < 0.0 {
+                // TODO(andreas): This is a hack, we should have a better way to handle this
+                return self.zeroed_texture_float.clone();
```
emilk, line 288: Returning `None` seems like maybe a better choice?

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+    pub fn height(&self) -> u32 {
+        self.decoder.height()
+    }
+
+    /// Returns a texture with the latest frame at the given timestamp.
+    ///
+    /// If the timestamp is negative, a zeroed texture is returned.
+    pub fn get_frame(&mut self, timestamp_s: f64) -> GpuTexture2D {
+        re_tracing::profile_function!();
+        self.decoder.get_frame(TimeMs::new(timestamp_s * 1e3))
```
emilk, line 51: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

```suggestion
    pub fn frame_at(&mut self, timestamp_s: f64) -> GpuTexture2D {
        re_tracing::profile_function!();
        self.decoder.frame_at(TimeMs::new(timestamp_s * 1e3))
```

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+        pub fn get_frame(&mut self, timestamp: TimeMs) -> GpuTexture2D {
+            if timestamp.as_f64() < 0.0 {
+                // TODO(andreas): This is a hack, we should have a better way to handle this
+                return self.zeroed_texture_float.clone();
+            }
+
+            let Some(segment_idx) =
+                latest_at_idx(&self.data.segments, |segment| segment.timestamp, &timestamp)
+            else {
+                return self.texture.clone();
```
emilk, line 294: Shouldn't this return a zeroed texture instead, since this only happens if we're before the first segment?

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+                .map(TimeMs::new)
+                .zip(frame.duration().map(TimeMs::new))
+            else {
+                // TODO(jan): figure out when this can happen and handle it
+                return self.texture.clone();
+            };
+
+            if TimeMs::new(timestamp.as_f64() - frame_timestamp_ms.as_f64()) > frame_duration_ms {
+                // not relevant to the user, it's an old frame.
+                return self.texture.clone();
```
emilk, line 353: I don't follow this comment. The `frame` we looked up is outdated, so we return `self.texture`… why?

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+                })
+                .ok()
+            else {
+                return;
+            };
+
+            self.decoder.decode(&chunk);
+        }
+
+        fn reset(&mut self) {
```
emilk, line 399: docstring

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+        frame: web_sys::VideoFrame,
+        texture: &wgpu::Texture,
+    ) {
+        let size = wgpu::Extent3d {
+            width: frame.display_width(),
+            height: frame.display_height(),
+            depth_or_array_layers: 1,
+        };
+        let source = {
+            // TODO(jan): Add `VideoFrame` support to `wgpu`
```
emilk, line 420:
```suggestion
            // TODO(jan): Use `VideoFrame` from `wgpu` when it ships: https://github.com/gfx-rs/wgpu/pull/6170
```

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+
+    use super::alloc_video_frame_texture;
+
+    pub struct VideoDecoder {
+        data: re_video::VideoData,
+        texture: GpuTexture2D,
+    }
+
+    impl VideoDecoder {
+        pub fn new(render_context: &RenderContext, data: re_video::VideoData) -> Option<Self> {
```
emilk, line 517:
```suggestion
        pub fn new(render_context: &RenderContext, data: re_video::VideoData) -> Option<Self> {
            re_log::warn_once!("Video playback not yet available in the native viewer. Try the web viewer instead. See https://github.com/rerun-io/rerun/issues/7298 for more");
```

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+
+        if let Some(v) = v.dyn_ref::<js_sys::Error>() {
+            return std::string::ToString::to_string(&v.to_string());
+        }
+
+        format!("{v:#?}")
+    }
+}
+
+// TODO(jan): decode on native
```
emilk, line 495:
```suggestion
// TODO(#7298): decode on native
```

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+        assert_eq!(latest_at_idx(&v, |v| *v, &8), Some(7));
+        assert_eq!(latest_at_idx(&v, |v| *v, &9), Some(8));
+        assert_eq!(latest_at_idx(&v, |v| *v, &10), Some(9));
+        assert_eq!(latest_at_idx(&v, |v| *v, &11), Some(9));
+        assert_eq!(latest_at_idx(&v, |v| *v, &1000), Some(9));
+    }
+}
+
+#[cfg(target_arch = "wasm32")]
+mod decoder {
```
emilk, line 141: This could move to its own file `video/web.rs`, and make `video.rs` a `video/mod.rs`. We're only expecting this module to grow, especially once we add native support.

crates/viewer/re_renderer_examples/framework.rs
```diff
@@ -2,7 +2,7 @@
 
 // TODO(#6330): remove unwrap()
 #![allow(clippy::unwrap_used)]
-
+#![allow(dead_code)]
```
emilk, line 5: ?

docs/snippets/all/archetypes/video_simple.py
```diff
@@ -0,0 +1,31 @@
...
+
+if len(sys.argv) < 2:
+    print(f"Usage: {sys.argv[0]} <path_to_asset.[mp4]>")
+    sys.exit(1)
+
+rr.init("rerun_example_asset_video")
+rr.serve()
+
+rr.set_time_seconds("stuff", 0)
+rr.log("world", rr.ViewCoordinates.RIGHT_HAND_Z_UP, static=True)  # Set an up-axis
```
emilk, line 16: This doesn't seem relevant

docs/snippets/all/archetypes/video_simple.py
```diff
@@ -0,0 +1,31 @@
+"""Log a simple 3D asset."""
```
emilk, line 1:
```suggestion
"""Log a video file."""
```

docs/snippets/all/archetypes/video_simple.py
```diff
@@ -0,0 +1,31 @@
...
+import rerun as rr
+
+if len(sys.argv) < 2:
+    print(f"Usage: {sys.argv[0]} <path_to_asset.[mp4]>")
+    sys.exit(1)
+
+rr.init("rerun_example_asset_video")
+rr.serve()
+
+rr.set_time_seconds("stuff", 0)
```
emilk, line 15: Explain why the `stuff` timeline is there

scripts/ci/rust_checks.py
```diff
@@ -48,8 +48,8 @@ def run_cargo(cargo_cmd: str, cargo_args: str, clippy_conf: str | None = None) -
     start_time = time.time()
 
     additional_env_vars = {}
-    additional_env_vars["RUSTFLAGS"] = "--deny warnings"
-    additional_env_vars["RUSTDOCFLAGS"] = "--deny warnings"
+    additional_env_vars["RUSTFLAGS"] = "--cfg=web_sys_unstable_apis --deny warnings"
+    additional_env_vars["RUSTDOCFLAGS"] = "--cfg=web_sys_unstable_apis --deny warnings"
```
emilk, line 52: Document why

emilk (approved):
Great! But how do I test this? Do you have a link to a video file to try?

jprochazk:
You can test with this video: https://github.com/rerun-io/video-decode-test/blob/main/data/bbb_video_av1_frag.mp4
Any AV1-encoded MP4 should work, though

crates/viewer/re_renderer/src/renderer/video.rs
jprochazk: What other cases are there? The timestamp is guaranteed to be in the range `samples[0].timestamp..samples[last].timestamp + samples[last].duration`, and `latest_at` should not return `None` for any value in that range.

Though this can probably just be another early exit just to be on the safe side

crates/viewer/re_renderer/src/renderer/video.rs
```diff
@@ -0,0 +1,544 @@
...
+            if segment_idx != self.current_segment_idx {
+                let segment_distance = segment_idx as isize - self.current_segment_idx as isize;
+                if segment_distance == 1 {
+                    // forward seek to next segment
+                    self.enqueue_all(segment_idx + 1);
+                } else {
+                    // forward seek by N>1 OR backward seek across segments
+                    self.reset();
+                    self.enqueue_all(segment_idx);
+                    self.enqueue_all(segment_idx + 1);
```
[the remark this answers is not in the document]
jprochazk: Yes, the rule here is 2 full segments queued ahead of time, so that we can always transition smoothly to the next segment without waiting. Will document

crates/viewer/re_renderer/src/renderer/video.rs
jprochazk: We should return zeroed texture here. But the point is that we're not copying that frame into the texture.

emilk:
> You can test with this video: [rerun-io/video-decode-test@`main`/data/bbb_video_av1_frag.mp4](https://github.com/rerun-io/video-decode-test/blob/main/data/bbb_video_av1_frag.mp4?rgh-link-date=2024-08-28T13%3A08%3A25Z) Any AV1-encoded MP4 should work, though

Nice! I've pushed a commit that allows me to play the video with

`pixi run rerun-web video.mp4`

crates/store/re_video/src/lib.rs
jprochazk: I just felt like `Duration` was a bit more annoying to use. It has `as_millis(&self) -> u128` or `as_secs_f64(&self) -> f64`, both of which require a bit more conversion. All the web APIs also assume milliseconds, so knowing that everything is always in milliseconds is convenient

crates/viewer/re_renderer/src/renderer/video.rs
jprochazk: I want to always return _some_ texture, even if it's going to be empty. If we need to know if a given frame is available immediately, we can add a separate method for that that would just look inside the `frames` buffer.

Issue #7292, opened 2024-08-28
Performance regression with many poses


teh-cmc:
Via @Danvil in https://github.com/rerun-io/rerun/issues/7222#issuecomment-2313667330:
> With branch `master` (as of 8/27, 3PM PDT) it feels a little bit faster but performance is still lower than rerun 0.16.
> 
> I tried some bisection of the rerun visualizations in my application and the bottleneck now seems to be this code:
> 
> ```
> rr.set_time_seconds("timestamp", my_timestamp);
> rr.log("path", &rerun::Clear::recursive())?;
> for (i, pose) in path.iter().enumerate() {
>     rr.log(format!("{tag}/{i}"), &rerun::Transform3D::from_translation_mat3x3(pose.t3, pose.r3))?;
> }
> ```
> 
> (there are 250 poses)
> 
> Without this visualization `0.18` also is much faster. So it seems like a large part of the slowdown from `0.16` to `0.18` is due to visualization of the poses.

jleibs:
Any chance the recursive clear is impacting things now that we do query-time clears?

teh-cmc:
(This will be my ever-involving live-investigation comment)

---

```python
import rerun as rr

rr.init("rerun_example_transform3d")
rr.save("/tmp/buncha_transforms_without_clear.rrd")

rr.set_time_seconds("timestamp", 42)
for i in range(0, 250):
    rr.log(f"transforms/{i}", rr.Transform3D(translation=[i, 0, 0]))
```

---

```python
import rerun as rr

rr.init("rerun_example_transform3d")
rr.save("/tmp/buncha_transforms_with_clear.rrd")

rr.set_time_seconds("timestamp", 42)
rr.log("transforms", rr.Clear(recursive=True))
for i in range(0, 250):
    rr.log(f"transforms/{i}", rr.Transform3D(translation=[i, 0, 0]))
```

PR #7300, opened 2024-08-28
Speed up handling of large numbers of transform entities

jleibs:
### What
Resolves: https://github.com/rerun-io/rerun/issues/7292

Halves the time spent in `execute_systems` when using many Transforms.
- TransformContext uses a store subscriper to track which components are non-empty and only queries those from the store.
- Arrows visualizer no longer queries for overrides on the unused transform components since those are fetched from the transform context manager anyways.

Before:
image

After:
image

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7300?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7300?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7300)

### DIFF
--- a/crates/viewer/re_space_view_spatial/src/transform_component_tracker.rs
+++ b/crates/viewer/re_space_view_spatial/src/transform_component_tracker.rs
@@ -1,7 +1,7 @@
 use ahash::HashMap;
 use once_cell::sync::OnceCell;
 
-use nohash_hasher::IntSet;
+use nohash_hasher::{IntMap, IntSet};
 use re_chunk_store::{
     ChunkStore, ChunkStoreDiffKind, ChunkStoreEvent, ChunkStoreSubscriber,
     ChunkStoreSubscriberHandle,
@@ -22,11 +22,11 @@ use re_types::ComponentName;
 pub struct TransformComponentTracker {
     /// Which entities have had any `Transform3D` component at any point in time, and which
     /// components they actually make use of.
-    transform3d_entities: HashMap<EntityPathHash, IntSet<ComponentName>>,
+    transform3d_entities: IntMap<EntityPathHash, IntSet<ComponentName>>,
 
     /// Which entities have had any `InstancePoses3D` components at any point in time, and
     /// which components they actually make use of.
-    pose3d_entities: HashMap<EntityPathHash, IntSet<ComponentName>>,
+    pose3d_entities: IntMap<EntityPathHash, IntSet<ComponentName>>,
 }
 
 impl TransformComponentTracker {
--- a/crates/viewer/re_space_view_spatial/src/transform_component_tracker.rs
+++ b/crates/viewer/re_space_view_spatial/src/transform_component_tracker.rs
@@ -130,9 +130,7 @@ impl ChunkStoreSubscriber for TransformComponentTrackerStoreSubscriber {
                         .components()
                         .get(&component_name)
                         .map_or(false, |list_array| {
-                            list_array
-                                .iter()
-                                .any(|list| !list.map_or(false, |list| list.is_empty()))
+                            list_array.offsets().lengths().any(|len| len > 0)
                         })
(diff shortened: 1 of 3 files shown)


crates/viewer/re_space_view_spatial/src/transform_component_tracker.rs
```diff
@@ -119,15 +124,39 @@ impl ChunkStoreSubscriber for TransformComponentTrackerStoreSubscriber {
...
+                if self.transform_components.contains(&component_name)
+                    && event
+                        .chunk
+                        .components()
+                        .get(&component_name)
+                        .map_or(false, |list_array| {
+                            list_array
+                                .iter()
+                                .any(|list| !list.map_or(false, |list| list.is_empty()))
+                        })
```
jleibs, line 136: @teh-cmc is there a better way to do this?

crates/viewer/re_space_view_spatial/src/transform_component_tracker.rs
```diff
@@ -20,11 +20,13 @@ use re_types::ComponentName;
 /// This is a huge performance improvement in practice, especially in recordings with many entities.
 #[derive(Default)]
 pub struct TransformComponentTracker {
-    /// Which entities have had any `Transform3D` component at any point in time.
-    transform3d_entities: IntSet<EntityPathHash>,
+    /// Which entities have had any `Transform3D` component at any point in time, and which
+    /// components they actually make use of.
+    transform3d_entities: HashMap<EntityPathHash, IntSet<ComponentName>>,
```
teh-cmc, line 25: Any reason the outer `IntSet` turned into a `HashMap`?

crates/viewer/re_space_view_spatial/src/transform_component_tracker.rs
```diff
@@ -20,11 +20,13 @@ use re_types::ComponentName;
...
 
-    /// Which entities have had any `InstancePoses3D` components at any point in time.
-    pose3d_entities: IntSet<EntityPathHash>,
+    /// Which entities have had any `InstancePoses3D` components at any point in time, and
+    /// which components they actually make use of.
+    pose3d_entities: HashMap<EntityPathHash, IntSet<ComponentName>>,
```
teh-cmc, line 29: Same here

crates/viewer/re_space_view_spatial/src/transform_component_tracker.rs
teh-cmc: Hmm... maybe you could do something with the offsets instead, e.g.:
```rust
.map_or(false, |list_array| {
    list_array.offsets().lengths().any(|len| len > 0)
})
```
I think that's correct?

At the very least you can use `values_iter()` instead of `iter()`, which is going to be more performant in this case since you don't care about nullability (null arrays advertise a len of 0):
```rust
list_array.values_iter().any(|array| !array.is_empty())
```

teh-cmc (approved):

teh-cmc:
Would you say this:
* Fixes https://github.com/rerun-io/rerun/issues/7292

?

crates/viewer/re_space_view_spatial/src/transform_component_tracker.rs
jleibs: Ahh, no. Thank you. Should have used IntMap.

jleibs:
> Would you say this: Fixes https://github.com/rerun-io/rerun/issues/7292

I'm going to mark it as resolves for now. Just compared release build between 0.18 and o.17 and they both end up in the ballpark of 3.3ms for system execution.

Issue #7284, opened 2024-08-27
New ChunkStore APIs to facilitate data access


jleibs:
This proposal covers 3 newish things.
- A `ComponentColumnDescriptor` that is roughly inspiried by our future vision of tagged components
- It  introduces `TimeColumnDescriptor` as a way of talking about how temporal data will be materialized in the record-batches. Because the user has the abillity to pass in the list of columns they are interested in, they can choose whether or not they want the query results to include a temporal column.  This includes supporting use-cases like including the `log_time` column when executing a range query according to `frame` as the timeline.
- It suggests introduction of a `RowJoinPolicy` in the range query that determines the behavior. This policy seems fairly unambiguous and the existence of the different options and abillity to switch between them seems like it has the potential to clarify for a user what's happening rather than making it always magic.

Proposed interfaces:
```rust
pub struct TimeColumnDescriptor {
    /// The timeline this column is associated with.
    timeline: Timeline,

    /// The data type of the column.
    data_type: ArrowDataType,
}

pub struct ComponentColumnDescriptor {
    /// The path of the entity
    entity_path: EntityPath,

    /// Optional name of the `Archetype` associated with this data.
    ///
    /// `None` if the data wasn't logged through an archetype.
    ///
    /// Example: `rerun.archetypes.Points3D`.
    archetype_name: Option<String>,

    /// Optional name of the field within `Archetype` associated with this data.
    ///
    /// `None` if the data wasn't logged through an archetype.
    ///
    /// Example: `positions`.
    archetype_field: Option<String>,

    /// Semantic name associated with this data.
    ///
    /// This is fully implied by `archetype_name` and `archetype_field`, but
    /// included for semantic convenience.
    ///
    /// Example: `rerun.components.Position3D`.
    component_name: String,

    /// The data type of the column.
    data_type: ArrowDataType,

    /// Whether this column represents static data.
    is_static: bool,
}

enum ColumnDescriptor {
    Time(TimeColumnDescriptor),
    Component(ComponentColumnDescriptor),
}

pub struct LatestAtQueryExpression {
    /// The entity path expression to query.
    ///
    /// Example: `world/camera/**`
    entity_path_expr: String,

    /// The timeline to query.
    ///
    /// Example `frame`
    timeline: Timeline,

    /// The point of view to query from.
    ///
    /// Example: `18`
    time_value: TimeInt,
}

/// The `RowJoinPolicy` determines the joining behavior of a `RangeQueryExpression`.
///
[...]

PR #7338, opened 2024-09-03
Dataframe queries 0: bootstrap & data model

teh-cmc:
All the boilerplate for the new `re_dataframe`.

Also introduces all the new types:
* `QueryExpression`, `LatestAtQueryExpression`, `RangeQueryExpression`
* `QueryHandle`, `LatestAtQueryHandle` (unimplemented), `RangeQueryHandle` (unimplemented)
* `ColumnDescriptor`, `ControlColumnDescriptor`, `TimeColumnDescriptor`, `ComponentColumnDescriptor`

No actual code logic, just definitions.

* Part of #7284 

---

Dataframe APIs PR series:
- #7338
- #7339
- #7340
- #7341
- #7345

### Checklist

### DIFF
--- a/crates/store/re_chunk_store/src/dataframe.rs
+++ b/crates/store/re_chunk_store/src/dataframe.rs
@@ -1,0 +1,499 @@
+//! All the APIs used specifically for `re_dataframe`.
+
+#![allow(unused_imports)]
+
+use std::collections::BTreeSet;
+
+use ahash::HashSet;
+use arrow2::datatypes::{DataType as ArrowDatatype, Field as ArrowField};
+use itertools::Itertools as _;
+
+use re_chunk::LatestAtQuery;
+use re_log_types::ResolvedTimeRange;
+use re_log_types::{EntityPath, TimeInt, Timeline};
+use re_types_core::{ArchetypeName, ComponentName, Loggable as _};
+
+use crate::ChunkStore;
+
+// Used all over in docstrings.
+#[allow(unused_imports)]
+use crate::RowId;
+
+// --- Descriptors ---
+
+// TODO(#6889): At some point all these descriptors needs to be interned and have handles or
+// something. And of course they need to be codegen. But we'll get there once we're back to
+// natively tagged components.
+
+// Describes any kind of column.
+//
+// See:
+// * [`ControlColumnDescriptor`]
+// * [`TimeColumnDescriptor`]
+// * [`ComponentColumnDescriptor`]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum ColumnDescriptor {
+    Control(ControlColumnDescriptor),
+    Time(TimeColumnDescriptor),
+    Component(ComponentColumnDescriptor),
+}
+
+impl ColumnDescriptor {
+    #[inline]
+    pub fn entity_path(&self) -> Option<&EntityPath> {
+        match self {
+            Self::Component(descr) => Some(&descr.entity_path),
+            Self::Control(_) | Self::Time(_) => None,
+        }
+    }
+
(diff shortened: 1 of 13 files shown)


crates/store/re_chunk_store/src/dataframe.rs
```diff
@@ -0,0 +1,499 @@
...
+pub struct ComponentColumnDescriptor {
+    /// The path of the entity.
+    pub entity_path: EntityPath,
+
+    /// Optional name of the `Archetype` associated with this data.
+    ///
+    /// `None` if the data wasn't logged through an archetype.
+    ///
+    /// Example: `rerun.archetypes.Points3D`.
+    pub archetype_name: Option<ArchetypeName>,
```
abey79, line 169: Do we have a clear plan to populate this in the short term? best effort if no ambiguity based on the indicator component? but that would require accessing codegen'd reflection from `re_dataframe`, right?

Or should we drop this until we have tagged components?

abey79 (approved):
Very promising!

crates/store/re_chunk_store/src/dataframe.rs
teh-cmc: > Do we have a clear plan to populate this in the short term? best effort if no ambiguity based on the indicator component? but that would require accessing codegen'd reflection from `re_dataframe`, right?

I investigated such shenanigans and quickly came to the conclusion that the pain wasn't worth it.

> Or should we drop this until we have tagged components?

That is where my investigation led me :upside_down_face:

crates/store/re_chunk_store/src/dataframe.rs
```diff
@@ -0,0 +1,499 @@
...
+            component_name.to_string(),
+            datatype.clone(),
+            false, /* nullable */
+        )
+    }
+}
+
+/// Describes a time column, such as `log_time`.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct TimeColumnDescriptor {
```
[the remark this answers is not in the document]
teh-cmc: I wouldn't mind it but for now I wouldn't know what to put in it and I'm in a hurry to ship this :no_mouth:

crates/store/re_chunk_store/src/dataframe.rs
```diff
@@ -0,0 +1,499 @@
...
+            component_name,
+            datatype,
+            is_static,
+        } = self;
+
+        ArrowField::new(
+            component_name.short_name().to_owned(),
+            datatype.clone(),
+            false, /* nullable */
+        )
```
jleibs, line 287: In order to successfully build an arrow2 RecordBatch using the produced schema, I had to make the following change:
```suggestion
        ArrowField::new(
            component_name.short_name().to_owned(),
            ArrowDatatype::List(std::sync::Arc::new(ArrowField::new(
                "item",
                datatype.clone(),
                true, /* is_nullable = true; This seems backwards to me. */
            ))),
            false, /* is_nullable = true; This seems backwards to me. */
        )
```

However, this change appears to require nullabillity flags that are backwards from my intuition.  I would have expected the INNER list to be non-nullable -- we can't null individual elements within a batch. But the OUTER list to be nullable -- individual rows can be null.

teh-cmc: The way you've set those nullability flags makes sense to me:
* for a list datatype, the nullability flag refers to its items.
* for a top-level field, the nullability datatype should logically describe whether the entire field can be missing from the recordbatch payload -- though from what ive seen in practice it mostly just means gibberish and you have to set it to whatever value makes the system you're working with happy.

The fact that the list layer is missing at all is a very real problem though >< I forgot that `TransportChunk` made that one automagically work :sparkles:

PR #7340, opened 2024-09-03
Dataframe queries 2: latest-at

teh-cmc:
Implements the latest-api dataframe API.

Examples:
```
cargo r --all-features -p re_dataframe --example latest_at -- /tmp/helix.rrd
cargo r --all-features -p re_dataframe --example latest_at -- /tmp/helix.rrd /helix/structure/scaffolding/**
```

```rust
use itertools::Itertools as _;

use re_chunk::{TimeInt, Timeline};
use re_chunk_store::{ChunkStore, ChunkStoreConfig, LatestAtQueryExpression, VersionPolicy};
use re_dataframe::QueryEngine;
use re_log_types::StoreKind;

fn main() -> anyhow::Result<()> {
    let args = std::env::args().collect_vec();

    let get_arg = |i| {
        let Some(value) = args.get(i) else {
            eprintln!(
                "Usage: {} <path_to_rrd> <entity_path_expr>",
                args.first().map_or("$BIN", |s| s.as_str())
            );
            std::process::exit(1);
        };
        value
    };

    let path_to_rrd = get_arg(1);
    let entity_path_expr = args.get(2).map_or("/**", |s| s.as_str());

    let stores = ChunkStore::from_rrd_filepath(
        &ChunkStoreConfig::DEFAULT,
        path_to_rrd,
        VersionPolicy::Warn,
    )?;

    for (store_id, store) in &stores {
        if store_id.kind != StoreKind::Recording {
            continue;
        }

        let cache = re_dataframe::external::re_query::Caches::new(store);
        let engine = QueryEngine {
            store,
            cache: &cache,
        };

        let query = LatestAtQueryExpression {
            entity_path_expr: entity_path_expr.into(),
            timeline: Timeline::log_time(),
            at: TimeInt::MAX,
        };

        let query_handle = engine.latest_at(&query, None /* columns */);
        let batch = query_handle.get();

        eprintln!("{query}:\n{batch}");
    }

    Ok(())
}
```

* Part of #7284 

---

Dataframe APIs PR series:
- #7338
- #7339
- #7340
- #7341
- #7345

### Checklist

### DIFF
--- a/crates/store/re_dataframe/examples/latest_at.rs
+++ b/crates/store/re_dataframe/examples/latest_at.rs
@@ -8,13 +8,19 @@ use re_log_types::StoreKind;
 fn main() -> anyhow::Result<()> {
     let args = std::env::args().collect_vec();
 
-    let Some(path_to_rrd) = args.get(1) else {
-        eprintln!(
-            "Usage: {} <path_to_rrd>",
-            args.first().map_or("$BIN", |s| s.as_str())
-        );
-        std::process::exit(1);
+    let get_arg = |i| {
+        let Some(value) = args.get(i) else {
+            eprintln!(
+                "Usage: {} <path_to_rrd> <entity_path_expr>",
+                args.first().map_or("$BIN", |s| s.as_str())
+            );
+            std::process::exit(1);
+        };
+        value
     };
+
+    let path_to_rrd = get_arg(1);
+    let entity_path_expr = args.get(2).map_or("/**", |s| s.as_str());
 
     let stores = ChunkStore::from_rrd_filepath(
         &ChunkStoreConfig::DEFAULT,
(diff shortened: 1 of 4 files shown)


crates/store/re_dataframe/examples/latest_at.rs
```diff
@@ -0,0 +1,64 @@
...
+                timeline: Timeline::log_tick(),
+                at: TimeInt::new_temporal(30),
+            };
+
+            let query_handle = engine.latest_at(&query, None /* columns */);
+            let batch = query_handle.get();
+
+            eprintln!("{query}:\n{batch}");
+        }
+    }
```
teh-cmc, line 61: This cannot stay hardcoded to these paths ><

crates/store/re_dataframe/examples/latest_at.rs
```diff
@@ -0,0 +1,55 @@
...
+use re_dataframe::QueryEngine;
+use re_log_types::StoreKind;
+
+fn main() -> anyhow::Result<()> {
+    let args = std::env::args().collect_vec();
+
+    let get_arg = |i| {
+        let Some(value) = args.get(i) else {
+            eprintln!(
+                "Usage: {} <path_to_rrd> <entity_path_expr>",
```
teh-cmc, line 14:
```suggestion
            eprintln!(
                "Usage: {} <path_to_rrd> [entity_path_expr]",

File: crates/store/re_dataframe/src/latest_at.rs [hunk 3]
```diff
@@ -0,0 +1,233 @@
...
+                    ColumnDescriptor::Time(descr) => {
+                        let time_column = max_time_per_timeline
+                            .remove(&descr.timeline)
+                            .and_then(|(_, chunk)| chunk.timelines().get(&descr.timeline).cloned());
+
+                        Some(if cfg!(debug_assertions) {
+                            #[allow(clippy::unwrap_used)]
+                            time_column.unwrap().times_array().to_boxed()
+                        } else {
+                            // NOTE: Technically cannot ever happen, but I'd rather that than an uwnrap.
```
jleibs, line 192: I think this can happen if you request a TimeColumn that was never logged?

My read of the above code is it won't end up with an entry in `max_time_per_timeline`, and as such `null_array` is the right thing to return here.

jleibs (approved):

crates/store/re_dataframe/src/latest_at.rs
teh-cmc: Ha yes, that dates from the time where I would filter out "impossible" columns from `user_columns`.

crates/store/re_dataframe/src/latest_at.rs
```diff
@@ -0,0 +1,233 @@
...
+            engine,
+            query,
+            user_columns,
+            state: Default::default(),
+        }
+    }
+}
+
+impl LatestAtQueryHandle<'_> {
+    /// All results returned by this handle will strictly follow this schema.
```
abey79, line 59: It seems that `schema_for_query()` may return a false positive that could leak into `LatestAtQueryHandle::schema()`—unless I missed something, which is likely. If so, what happens in such a case?

teh-cmc: Yes, all schema-related methods are best-effort in that they only rely on chunk metadata rather than full-blown queries (this will become extremely important once we stop assuming that the entire index is available in local memory).

The schema of the result always match perfectly with the announced schema: the false positives are simply filled with null values.

PR #7341, opened 2024-09-03
Dataframe queries 3: dense range

### LINKED ISSUE #7284: New ChunkStore APIs to facilitate data access

teh-cmc:
Implements the dense range dataframe APIs.

Examples:
```
cargo r --all-features -p re_dataframe --example range -- /tmp/data.rrd /helix/structure/scaffolding/beads
cargo r --all-features -p re_dataframe --example range -- /tmp/data.rrd /helix/structure/scaffolding/beads /helix/structure/scaffolding/**
```

```rust
use itertools::Itertools as _;

use re_chunk_store::{
    ChunkStore, ChunkStoreConfig, ComponentColumnDescriptor, RangeQueryExpression, Timeline,
    VersionPolicy,
};
use re_dataframe::QueryEngine;
use re_log_types::{ResolvedTimeRange, StoreKind};

fn main() -> anyhow::Result<()> {
    let args = std::env::args().collect_vec();

    let get_arg = |i| {
        let Some(value) = args.get(i) else {
            eprintln!(
                "Usage: {} <path_to_rrd_with_position3ds> <entity_path_pov> [entity_path_expr]",
                args.first().map_or("$BIN", |s| s.as_str())
            );
            std::process::exit(1);
        };
        value
    };

    let path_to_rrd = get_arg(1);
    let entity_path_pov = get_arg(2).as_str();
    let entity_path_expr = args.get(3).map_or("/**", |s| s.as_str());

    let stores = ChunkStore::from_rrd_filepath(
        &ChunkStoreConfig::DEFAULT,
        path_to_rrd,
        VersionPolicy::Warn,
    )?;

    for (store_id, store) in &stores {
        if store_id.kind != StoreKind::Recording {
            continue;
        }

        let cache = re_dataframe::external::re_query::Caches::new(store);
        let engine = QueryEngine {
            store,
            cache: &cache,
        };

        let query = RangeQueryExpression {
            entity_path_expr: entity_path_expr.into(),
            timeline: Timeline::log_tick(),
            time_range: ResolvedTimeRange::new(0, 30),
            pov: ComponentColumnDescriptor::new::<re_types::components::Position3D>(
                entity_path_pov.into(),
            ),
        };

        let query_handle = engine.range(&query, None /* columns */);
        eprintln!("{query}:");
        for batch in query_handle.into_iter() {
            eprintln!("{batch}");
        }
    }

    Ok(())
}
```

* Fixes #7284 

---

Dataframe APIs PR series:
- #7338
- #7339
- #7340
- #7341
- #7345

### Checklist

### DIFF
--- a/crates/store/re_dataframe/examples/range.rs
+++ b/crates/store/re_dataframe/examples/range.rs
@@ -10,13 +10,20 @@ use re_log_types::{ResolvedTimeRange, StoreKind};
 fn main() -> anyhow::Result<()> {
     let args = std::env::args().collect_vec();
 
-    let Some(path_to_rrd) = args.get(1) else {
-        eprintln!(
-            "Usage: {} <path_to_rrd>",
-            args.first().map_or("$BIN", |s| s.as_str())
-        );
-        std::process::exit(1);
+    let get_arg = |i| {
+        let Some(value) = args.get(i) else {
+            eprintln!(
+                "Usage: {} <path_to_rrd_with_position3ds> <entity_path_pov> [entity_path_expr]",
+                args.first().map_or("$BIN", |s| s.as_str())
+            );
+            std::process::exit(1);
+        };
+        value
     };
+
+    let path_to_rrd = get_arg(1);
+    let entity_path_pov = get_arg(2).as_str();
+    let entity_path_expr = args.get(3).map_or("/**", |s| s.as_str());
 
     let stores = ChunkStore::from_rrd_filepath(
         &ChunkStoreConfig::DEFAULT,
@@ -35,38 +42,19 @@ fn main() -> anyhow::Result<()> {
             cache: &cache,
         };
 
-        {
-            let query = RangeQueryExpression {
-                entity_path_expr: "/**".into(),
-                timeline: Timeline::log_tick(),
-                time_range: ResolvedTimeRange::new(0, 30),
-                pov: ComponentColumnDescriptor::new::<re_types::components::Position3D>(
-                    "helix/structure/scaffolding/beads".into(),
-                ),
-            };
+        let query = RangeQueryExpression {
+            entity_path_expr: entity_path_expr.into(),
+            timeline: Timeline::log_tick(),
+            time_range: ResolvedTimeRange::new(0, 30),
+            pov: ComponentColumnDescriptor::new::<re_types::components::Position3D>(
+                entity_path_pov.into(),
+            ),
+        };
 
-            let query_handle = engine.range(&query, None /* columns */);
-            eprintln!("{query}:");
-            for batch in query_handle.into_iter() {
-                eprintln!("{batch}");
-            }
-        }
-        eprintln!("---");
-        {
-            let query = RangeQueryExpression {
-                entity_path_expr: "/helix/structure/scaffolding/**".into(),
-                timeline: Timeline::log_tick(),
-                time_range: ResolvedTimeRange::new(0, 30),
-                pov: ComponentColumnDescriptor::new::<re_types::components::Position3D>(
-                    "helix/structure/scaffolding/beads".into(),
-                ),
-            };
-
-            let query_handle = engine.range(&query, None /* columns */);
-            eprintln!("{query}:");
-            for batch in query_handle.into_iter() {
-                eprintln!("{batch}");
-            }
+        let query_handle = engine.range(&query, None /* columns */);
+        eprintln!("{query}:");
+        for batch in query_handle.into_iter() {
+            eprintln!("{batch}");
         }
     }
 
--- a/crates/store/re_dataframe/examples/range.rs
+++ b/crates/store/re_dataframe/examples/range.rs
@@ -1,0 +1,74 @@
+use itertools::Itertools as _;
+
+use re_chunk_store::{
+    ChunkStore, ChunkStoreConfig, ComponentColumnDescriptor, RangeQueryExpression, Timeline,
+    VersionPolicy,
+};
+use re_dataframe::QueryEngine;
+use re_log_types::{ResolvedTimeRange, StoreKind};
+
+fn main() -> anyhow::Result<()> {
+    let args = std::env::args().collect_vec();
+
+    let Some(path_to_rrd) = args.get(1) else {
+        eprintln!(
+            "Usage: {} <path_to_rrd>",
+            args.first().map_or("$BIN", |s| s.as_str())
+        );
+        std::process::exit(1);
+    };
+
+    let stores = ChunkStore::from_rrd_filepath(
+        &ChunkStoreConfig::DEFAULT,
+        path_to_rrd,
+        VersionPolicy::Warn,
+    )?;
+
+    for (store_id, store) in &stores {
+        if store_id.kind != StoreKind::Recording {
+            continue;
+        }
+
+        let cache = re_dataframe::external::re_query::Caches::new(store);
+        let engine = QueryEngine {
+            store,
+            cache: &cache,
+        };
+
+        {
+            let query = RangeQueryExpression {
+                entity_path_expr: "/**".into(),
+                timeline: Timeline::log_tick(),
+                time_range: ResolvedTimeRange::new(0, 30),
+                pov: ComponentColumnDescriptor::new::<re_types::components::Position3D>(
+                    "helix/structure/scaffolding/beads".into(),
+                ),
+            };
+
+            let query_handle = engine.range(&query, None /* columns */);
+            eprintln!("{query}:");
+            for batch in query_handle.into_iter() {
+                eprintln!("{batch}");
+            }
+        }
+        eprintln!("---");
+        {
+            let query = RangeQueryExpression {
+                entity_path_expr: "/helix/structure/scaffolding/**".into(),
+                timeline: Timeline::log_tick(),
+                time_range: ResolvedTimeRange::new(0, 30),
+                pov: ComponentColumnDescriptor::new::<re_types::components::Position3D>(
+                    "helix/structure/scaffolding/beads".into(),
+                ),
+            };
+
+            let query_handle = engine.range(&query, None /* columns */);
+            eprintln!("{query}:");
+            for batch in query_handle.into_iter() {
+                eprintln!("{batch}");
+            }
+        }
+    }
+
+    Ok(())
+}
--- a/crates/store/re_dataframe/src/range.rs
+++ b/crates/store/re_dataframe/src/range.rs
@@ -140,6 +140,7 @@ impl RangeQueryHandle<'_> {
     /// while let Some(batch) = query_handle.next_page() {
     ///     // …
     /// }
+    /// ```
     pub fn next_page(&mut self) -> Option<RecordBatch> {
         re_tracing::profile_function!(format!("{:?}", self.query));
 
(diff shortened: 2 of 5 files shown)


teh-cmc:
@rerun-bot full-check

crates/store/re_dataframe/examples/range.rs
```diff
@@ -0,0 +1,74 @@
...
+                ),
+            };
+
+            let query_handle = engine.range(&query, None /* columns */);
+            eprintln!("{query}:");
+            for batch in query_handle.into_iter() {
+                eprintln!("{batch}");
+            }
+        }
+    }
```
teh-cmc, line 71: This cannot stay hardcoded to these paths ><

crates/store/re_dataframe/examples/range.rs
```diff
@@ -0,0 +1,64 @@
...
+        };
+
+        {
+            let query = RangeQueryExpression {
+                entity_path_expr: entity_path_expr.into(),
+                timeline: Timeline::log_tick(),
+                time_range: ResolvedTimeRange::new(0, 30),
+                pov: ComponentColumnDescriptor::new::<re_types::components::Position3D>(
+                    entity_path_pov.into(),
+                ),
```
teh-cmc, line 52: I'm still leaving that specific part hardcoded, because this is an example not the end-all-be-all CLI query tool.

It does show why using a full blown `ComponentColumnDescriptor` as pov is a massive pain though. To be improved on later on.

crates/store/re_dataframe/src/range.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// The schema of the returned [`RecordBatch`] is guaranteed to match the one returned by
+    /// [`Self::schema`].
+    /// Columns that do not yield any data will still be present in the results, filled with null values.
+    ///
+    /// "Natural pages" refers to pages of data that match 1:1 to the underlying storage.
+    /// The size of each page cannot be known in advance, as it depends on unspecified
+    /// implementation details.
+    /// This is the most performant way to iterate over the dataset.
+    ///
+    /// ```ignore
```
jleibs, line 139: This block looks to be unterminated.

crates/store/re_dataframe/src/range.rs
```diff
@@ -0,0 +1,265 @@
...
+        re_tracing::profile_function!(format!("{:?}", self.query));
+
+        _ = self.init();
+        let pov_chunk = self.state.get_mut()?.pov_chunks.as_mut()?.pop_front()?;
+        let pov_time_column = pov_chunk.timelines().get(&self.query.timeline)?;
+        let columns = self.schema();
+
+        // TODO(cmc): There are more efficient, albeit infinitely more complicated ways to do this.
+        // Let's first implement all features (multi-PoV, pagination, timestamp streaming, etc) and
+        // see if this ever becomes an issue before going down this road.
```
jleibs, line 153: Makes sense. We'll always need this as a worst-case fallback path, but I can definitely imagine this being replaced with a series of checks that determine whether particular happy-path join operations are applicable with more efficient implementations for those cases.

crates/store/re_dataframe/src/range.rs
```diff
@@ -0,0 +1,265 @@
...
+                    ColumnDescriptor::Control(_descr) => pov_chunk.row_ids_array().to_boxed(),
+
+                    ColumnDescriptor::Time(descr) => {
+                        let time_column = pov_chunk.timelines().get(&descr.timeline).cloned();
+
+                        if cfg!(debug_assertions) {
+                            #[allow(clippy::unwrap_used)] // want to crash in dev
+                            time_column.unwrap().times_array().to_boxed()
+                        } else {
+                            // NOTE: Technically cannot ever happen, but I'd rather that than an uwnrap.
```
jleibs, line 221: I believe this happens if the POV component was never logged on the requested (secondary) timeline.

jleibs (approved):

PR #7355, opened 2024-09-04
Add arrow feature to re_chunk and conversions to RecordBatch

jleibs:
### What
Basic type conversions from TransportChunk to RecordBatch and back.

Adding the round-trip test turned up an interesting issue.

TransportChunk <-> RecordBatch fails to round-trip successfully because we lose the ExtensionType encapsulation that used to be encoded by arrow2.

While on the surface this isn't immediately problematic, as we don't care about ExtensionTypes, the discussion indicates there are in fact going to be very real pain points when it comes to writing semantic data processing engines using arrow-rs. This is because the metadata is attached to the FIELD, not the DATATYPE, and there exist many processing contexts where the context of that field itself is lost.
https://github.com/apache/arrow-rs/issues/4472

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7355?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7355?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7355)

### DIFF
--- a/crates/store/re_chunk/src/arrow.rs
+++ b/crates/store/re_chunk/src/arrow.rs
@@ -8,7 +8,11 @@ use crate::TransportChunk;
 
 impl TransportChunk {
     /// Create an arrow-rs [`RecordBatch`] containing the data from this [`TransportChunk`].
-    pub fn try_as_arrow_record_batch(&self) -> Result<RecordBatch, ArrowError> {
+    ///
+    /// This is a "fairly" cheap operation, as it does not copy the underlying arrow data,
+    /// but does incur overhead of generating an alternative representation of the arrow-
+    /// related rust structures that refer to those data buffers.
+    pub fn try_to_arrow_record_batch(&self) -> Result<RecordBatch, ArrowError> {
         let fields: Vec<Field> = self
             .schema
             .fields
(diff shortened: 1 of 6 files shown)


crates/store/re_chunk/src/arrow.rs
```diff
@@ -0,0 +1,59 @@
...
+    array::{make_array, RecordBatch},
+    datatypes::{Field, Schema},
+    error::ArrowError,
+};
+
+use crate::TransportChunk;
+
+impl TransportChunk {
+    /// Create an arrow-rs [`RecordBatch`] containing the data from this [`TransportChunk`].
+    pub fn try_as_arrow_record_batch(&self) -> Result<RecordBatch, ArrowError> {
```
teh-cmc, line 11: Hate to be that guy but im gonna be that guy
image

```suggestion
    pub fn try_to_arrow_record_batch(&self) -> Result<RecordBatch, ArrowError> {

teh-cmc (approved):
very nice

PR #7358, opened 2024-09-05
Update ndarray to 0.16  and ndarray-rand to 0.15

### LINKED ISSUE #7157: Update `ndarray`
* Prompted by https://github.com/rerun-io/rerun/issues/7134

0.16 is out, with a bunch of deprecated functions

benliepert:
### What
Fixes #7157. I also updated ndarray-rand (sorry if that should be separate - it didn't appear to require any changes).

into_raw_vec() is now into_raw_vec_and_offset().0 (the offset is ignored)

I wasn't sure whether an order (`RowMajor` vs. `ColumnMajor`) should be specified in `into_shape_with_order()`. I ran through all the examples and couldn't find one that actually used this code, so guidance here is appreciated.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7358?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7358?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7358)


crates/store/re_types/src/datatypes/tensor_data_ext.rs
```diff
@@ -182,7 +182,9 @@ macro_rules! tensor_from_ndarray {
                     .is_standard_layout()
                     .then(|| TensorData {
                         shape,
-                        buffer: TensorBuffer::$variant(value.to_owned().into_raw_vec().into()),
+                        buffer: TensorBuffer::$variant(
+                            value.to_owned().into_raw_vec_and_offset().0.into(),
```
emilk, line 186: We shouldn't ignore the offset. If it is non-zero, we may have a problem.
We should at least return `TensorCastError::NotContiguousStdOrder` in that case.

However, ideally we should handle arrays that are not in standard layout (and/or with an offset) by calling `.iter().collect()` instead of bailing.

benliepert: Sounds good. Will give it a try.
I want to confirm my understanding that if the Array `is_standard_layout()` the offset can be safely ignored (which is also implied by you commenting on this edit and not the one further down, where a check for standard layout had been done first)

emilk: I don't know - the docs aren't very clear about that, but I believe it is not enough.

benliepert: You're correct, I was able to validate the suspicion with the following code (which passes, offset = 4)
```
    let array = Array::from_iter(0..16).into_shape_with_order((4, 4)).unwrap();
    let sliced_array = array.slice_move(s![1.., ..]);
    assert!(sliced_array.is_standard_layout());
    let (_, offset) = sliced_array.into_raw_vec_and_offset();
    assert!(offset.unwrap() > 0);
```
I'll make sure the test cases cover this as well

crates/store/re_types/src/datatypes/tensor_data_ext.rs
```diff
@@ -8,9 +8,6 @@ use crate::archetypes::EncodedImage;
 
 use super::{TensorBuffer, TensorData, TensorDimension};
 
-// Much of the following duplicates code from: `crates/re_components/src/tensor.rs`, which
```
benliepert, line 11: This comment didn't appear to be relevant anymore

crates/store/re_types/src/datatypes/tensor_data_ext.rs
```diff
@@ -169,7 +166,6 @@ macro_rules! tensor_from_ndarray {
             type Error = TensorCastError;
 
             fn try_from(value: ndarray::Array<$type, D>) -> Result<Self, Self::Error> {
-                let value = value.as_standard_layout();
```
benliepert, line 172: I don't think this was necessary (performance wise) and created a redundancy with the check for is_standard_layout() below

crates/store/re_types/src/datatypes/tensor_data_ext.rs
```diff
@@ -311,13 +320,18 @@ impl<D: ::ndarray::Dimension> TryFrom<::ndarray::Array<half::f16, D>> for Tensor
             })
             .collect();
         if value.is_standard_layout() {
+            let (vec, offset) = value.into_raw_vec_and_offset();
+            // into_raw_vec_and_offset() guarantees that the logical element order (.iter()) matches the internal
+            // storage order in the returned vector if the array is in standard layout.
+            let vec_slice = if let Some(offset) = offset {
+                &vec[offset..]
+            } else {
+                debug_assert!(vec.is_empty());
```
emilk, line 329: Why is the vector expected to be empty here?

crates/store/re_types/src/datatypes/tensor_data_ext.rs
```diff
@@ -178,13 +174,26 @@ macro_rules! tensor_from_ndarray {
...
+
+                let vec = if value.is_standard_layout() {
+                    let (mut vec, offset) = value.into_raw_vec_and_offset();
+                    // into_raw_vec_and_offset() guarantees that the logical element order (.iter()) matches the internal
+                    // storage order in the returned vector if the array is in standard layout.
+                    if let Some(offset) = offset {
+                        vec.drain(..offset);
+                        vec
+                    } else {
+                        debug_assert!(vec.is_empty());
```
emilk, line 186: Why is the vector expected to be empty here?

emilk (approved):
Thank you!

crates/store/re_types/src/datatypes/tensor_data_ext.rs
benliepert: into_raw_vec_and_offset() returns `None` when the Array is empty. The ndarray 0.16.1 documentation incorrectly states that it returns 0 in this case (I fixed that [here](https://github.com/rust-ndarray/ndarray/pull/1432))

crates/store/re_types/src/datatypes/tensor_data_ext.rs
benliepert: See above

PR #7377, opened 2024-09-09
Use `EntityPathFilter` in `re_dataframe`'s queries

abey79:
### What

This PR change `re_dataframe` to use the existing `EntityPathFilter` instead of the `EntityPathExpression` introduced last week (which lacks many features).

Also renames some methods of `EntityPathFilter`.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7377?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7377?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7377)

### DIFF
--- a/crates/store/re_log_types/src/path/entity_path_filter.rs
+++ b/crates/store/re_log_types/src/path/entity_path_filter.rs
@@ -5,6 +5,16 @@ use itertools::Itertools as _;
 
 use crate::EntityPath;
 
+/// Error returned by [`EntityPathFilter::parse_strict`].
+#[derive(thiserror::Error, Debug, PartialEq, Eq)]
+pub enum EntityPathFilterParseError {
+    #[error("Path parse error: {0}")]
+    PathParseError(#[from] crate::PathParseError),
+
+    #[error("Unresolved substitution: {0}")]
+    UnresolvedSubstitution(String),
+}
+
 /// A set of substitutions for entity paths.
 #[derive(Default)]
 pub struct EntityPathSubs(pub HashMap<String, String>);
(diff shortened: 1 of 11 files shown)


crates/store/re_log_types/src/path/entity_path_filter.rs
```diff
@@ -128,6 +128,12 @@ impl std::iter::Sum for EntityPathFilter {
     }
 }
 
+impl<S: AsRef<str>> From<S> for EntityPathFilter {
+    fn from(rules: S) -> Self {
+        Self::parse_forgiving(rules.as_ref(), &EntityPathSubs::default())
+    }
+}
```
jleibs, line 135: I see the utility here in test code, though worry about losing semantic annotation that this is a "forgiving" parse rather than a conversion.

I wish we had a strict parse and go do this with a `TryFrom` instead:
```
"/**".try_into()?
```
would actually be nice for test code to avoid accidentally ending up with malformed entity path expressions.

That said, I seems unlikely to matter in practice.

jleibs (approved):

crates/store/re_log_types/src/path/entity_path_filter.rs
abey79: that's a good point. I think most of these happen in test, so it would be perfectly valid to have `unwrap` or `?`. I'll make that change (unless something really blocking shows up).

abey79: Mmm. This is actually slightly more involved, but I think very much worth doing. I'll capture it in an issue and add a TODO.

abey79: I'm gonna merge because I'm blocked on it—let me know if you see something to be fixed in a follow-up PR.

Issue #7379, opened 2024-09-09
Add support for having entity path as rows


abey79:
The latest iteration of the dataframe view (based on `re_dataframe`) lost the ability to have entity path as rows (instead of columns). There might still be some value to have this possibility, especially for the latest-at mode (which otherwise has just one veeeery long row).

We agreed that this was to be implemented only when `re_datastore` actually implements the logic for having entities as rows (aka. the joining logic should _not_ be implemented in the dataframe view only).

Related
- https://github.com/rerun-io/rerun/issues/7068
- https://github.com/rerun-io/rerun/issues/7070

abey79:
I'm really not sure we still want that. @nikolausWest ?

nikolausWest:
Yeah lets close

Issue #6818, opened 2024-07-08
Document/publish our arrow schemas


jleibs:
- The intent of https://github.com/rerun-io/rerun/issues/6811 is to expose data to users as structured arrow data rather than using the Rerun-object-model.
- The intent of https://github.com/rerun-io/rerun/issues/6388 is to make that arrow data cleaner and easier to work with.

However, so far the arrow data has been an implicit implementation detail sitting behind our object-APIs. To make this data accessible we need to be able to show users how to work with it.

Unfortunately, there is no standard textual representation of an arrow schema. Although we have our own fbs files, this is a Rerun-specific implementation detail that is not necessarily familiar to those attempting to cross-reference with arrow documentation.

For each Rerun datatype, we should generate a doc page specifically focused on the arrow schema:
 - The terms used should as closely as possible matches the official terms from the arrow documentation for the native libraries.
 - We should include a generated code snippet in each of the 3 languages that produces the corresponding arrow datatype. Note that we already generate this code in our serializers. We just need to emit that same code into a block in the doc pages.

We also need to formalize the documentation around any additional meta-data required for interpretation as a component or inclusion into a chunk.

PR #7380, opened 2024-09-09
Update the dataframe view to use `re_dataframe` and `egui_table`

### LINKED ISSUE #7279: Dataframe: add PoV to range query
In the spirit of matching the dataframe view with the upcoming SDK data-out feature, the range query should be changed to always have _one_ component as PoV.

This actually is a simplifying assumption that (at least partially) unblocks #7066 for range queries.

abey79:
### What

- Closes #7279

Major update to the dataframe view
- display the data return by the new `re_dataframe` crate
  - the PoV entity/component is now actually used
  - entities are now always columns
    - see #7379 
- use [`egui_table`](https://github.com/rerun-io/egui_table) for the table
  - hierarchical header
  - sticky columns
  - and much more...

TODO:
- [x] fix after merging #7383

image

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7380?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7380?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7380)

### DIFF
--- a/crates/viewer/re_space_view_dataframe/src/display_record_batch.rs
+++ b/crates/viewer/re_space_view_dataframe/src/display_record_batch.rs
@@ -52,20 +52,20 @@ impl ComponentData {
                 column_data
                     .as_any()
                     .downcast_ref::<ArrowListArray<i32>>()
-                    .expect("sanity checked")
+                    .expect("`data_type` checked, failure is a bug in re_dataframe")
                     .clone(),
             )),
             DataType::Dictionary(IntegerType::Int32, _, _) => {
                 let dict = column_data
                     .as_any()
                     .downcast_ref::<ArrowDictionaryArray<i32>>()
-                    .expect("sanity checked")
+                    .expect("`data_type` checked, failure is a bug in re_dataframe")
                     .clone();
                 let values = dict
                     .values()
                     .as_any()
                     .downcast_ref::<ArrowListArray<i32>>()
-                    .expect("sanity checked")
+                    .expect("`data_type` checked, failure is a bug in re_dataframe")
                     .clone();
                 Ok(Self::DictionaryArray { dict, values })
             }
@@ -146,7 +146,7 @@ impl DisplayColumn {
                     let row_ids = column_data
                         .as_any()
                         .downcast_ref::<ArrowStructArray>()
-                        .expect("sanity checked");
+                        .expect("expected format for RowId, failure is a bug in re_dataframe");
                     let [times, counters] = row_ids.values() else {
                         panic!("RowIds are corrupt -- this should be impossible (sanity checked)");
                     };
@@ -155,14 +155,14 @@ impl DisplayColumn {
                     let row_id_times = times
                         .as_any()
                         .downcast_ref::<ArrowPrimitiveArray<u64>>()
-                        .expect("sanity checked")
+                        .expect("expected format for RowId, failure is a bug in re_dataframe")
                         .clone();
 
                     #[allow(clippy::unwrap_used)]
                     let row_id_counters = counters
                         .as_any()
                         .downcast_ref::<ArrowPrimitiveArray<u64>>()
-                        .expect("sanity checked")
+                        .expect("expected format for RowId, failure is a bug in re_dataframe")
                         .clone();
 
                     Ok(Self::RowId {


crates/viewer/re_viewer/src/web.rs
```diff
@@ -73,6 +73,7 @@ impl WebHandle {
             wgpu_options: crate::wgpu_options(app_options.render_backend.clone()),
             depth_buffer: 0,
             dithering: true,
+            should_propagate_event: Box::new(|_event| true),
```
abey79, line 76: I have no idea if this is correct.

crates/viewer/re_viewer/src/web.rs
```diff
@@ -73,6 +73,7 @@ impl WebHandle {
             wgpu_options: crate::wgpu_options(app_options.render_backend.clone()),
             depth_buffer: 0,
             dithering: true,
+            should_propagate_event: Box::new(|_event| false),
```
abey79, line 76: i have no idea if this is correct

crates/viewer/re_space_view_dataframe/src/display_record_batch.rs
```diff
@@ -0,0 +1,332 @@
...
+
+fn row_id_ui(ui: &mut egui::Ui, row_id: &RowId) {
+    let s = row_id.to_string();
+    let split_pos = s.char_indices().nth_back(5);
+
+    ui.label(match split_pos {
+        Some((pos, _)) => &s[pos..],
+        None => &s,
+    })
+    .on_hover_text(s);
```
emilk, line 331: We could also decode the time-stamp in the `RowId` and show that on hover 🤔

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    fn num_rows(&self) -> u64 {
+        match self {
+            QueryHandle::LatestAt(_) => 1,
+            QueryHandle::Range(query_handle) => query_handle.num_rows(),
+        }
+    }
+
+    fn get(&self, start: u64, num_rows: u64) -> Vec<RecordBatch> {
```
emilk, line 46: I think using `rows: Range<u64>` here (and elsewhere) instead would be more idiomatic and less error prone

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -0,0 +1,350 @@
...
+        match self {
+            QueryHandle::LatestAt(query_handle) => {
+                vec![query_handle.get()]
```
emilk, line 49: Is it intentional that the row range is ignored? Should this be documented? Should we have a `debug_assert` to check that the user asked for `0..1`?

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -0,0 +1,350 @@
...
+///
+/// Row data is stored in a bunch of [`DisplayRecordBatch`], which are created from
+/// [`RecordBatch`]s. We also maintain a mapping for each row number to the corresponding record
+/// batch and the index inside it.
+struct RowsDisplayData {
+    /// The [`DisplayRecordBatch`]s to display.
+    display_record_batches: Vec<DisplayRecordBatch>,
+
+    /// For each row to be displayed, the batch index and the row index with that batch.
+    batch_and_row_from_row_nr: BTreeMap<u64, (usize, usize)>,
```
emilk, line 85: Let's make this self-documenting instead (yes, I know I was the one who wrote this code - I can fix it)

```rs
struct BatchRef {
    batch_index: usize,
    row_index: u64,
}
```

```suggestion
    batch_and_row_from_row_nr: BTreeMap<u64, BatchRef>,
```

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -0,0 +1,350 @@
...
+
+    /// The index of the time column corresponding to the query timeline.
+    time_column_index: Option<usize>,
```
emilk, line 88:
```suggestion
    query_time_column_index: Option<usize>,
```

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -0,0 +1,350 @@
...
+            })
+            .map(|(pos, _)| pos);
+
+        // find the row id column
+        let row_id_column_index = schema
+            .iter()
+            .find_position(|desc| match desc {
+                ColumnDescriptor::Control(control_column_desc) => {
+                    control_column_desc.component_name
+                        == ComponentName::from("rerun.controls.RowId")
```
emilk, line 133: It would be nice if this wasn't quite as hard-coded

```suggestion
                        == ComponentName::row_id()
```

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -0,0 +1,350 @@
...
+
+impl<'a> egui_table::TableDelegate for DataframeTableDelegate<'a> {
+    fn prepare(&mut self, info: &egui_table::PrefetchInfo) {
+        re_tracing::profile_function!();
+
+        let data = RowsDisplayData::try_new(
+            &info.visible_rows,
+            self.query_handle.get(
+                info.visible_rows.start,
+                info.visible_rows.end - info.visible_rows.start,
```
emilk, line 167: If we use `Range` in the API instead

```suggestion
                info.visible_rows,
```

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -0,0 +1,350 @@
...
+            ),
+            self.schema,
+            &self.query_handle.timeline(),
+        );
+
+        self.display_data = data.with_context(|| "Failed to create display data");
```
emilk, line 173:
```suggestion
        self.display_data = data.context("Failed to create display data");
```

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -0,0 +1,350 @@
...
+        if cell.row_nr >= self.num_rows {
+            error_ui(
+                ui,
+                format!(
+                    "Unexpected row_nr: {} (table row count {})",
+                    cell.row_nr, self.num_rows
+                ),
+            );
+            return;
+        }
```
emilk, line 212: Let's make sure we don't miss these bugs

```suggestion
    debug_assert!(cell.row_nr < self.num_rows, "Bug in egui_table");
```

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -0,0 +1,350 @@
...
+            .show(ui, |ui| {
+                if let Some((batch_nr, batch_index)) = display_data
+                    .batch_and_row_from_row_nr
+                    .get(&cell.row_nr)
+                    .copied()
+                {
+                    let batch = &display_data.display_record_batches[batch_nr];
+                    let column = &batch.columns()[cell.col_nr];
+
+                    // compute the latest at query for this row (used to display tooltips)
```
emilk, line 233:
```suggestion
                    // compute the latest-at query for this row (used to display tooltips)
```

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -0,0 +1,350 @@
...
+                        .unwrap_or(TimeInt::MAX);
+                    let latest_at_query =
+                        LatestAtQuery::new(self.query_handle.timeline(), timestamp);
+                    let row_id = display_data
+                        .row_id_column_index
+                        .and_then(|index| {
+                            display_data.display_record_batches[batch_nr].columns()[index]
+                                .try_decode_row_id(batch_index)
+                        })
+                        .unwrap_or(RowId::ZERO);
```
emilk, line 249: To reduce rightward drift, we could call into another function from inside the `Frame`

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -0,0 +1,350 @@
...
+                    })
+                    .collect::<Vec<_>>(),
+            )
+            .num_sticky_cols(num_sticky_cols)
+            .headers(vec![
+                egui_table::HeaderRow {
+                    height: 20.0,
+                    groups: header_groups,
+                },
+                egui_table::HeaderRow::new(20.0),
```
emilk, line 310: This `20.0` should come from `re_ui` instead

crates/viewer/re_space_view_dataframe/src/display_record_batch.rs
```diff
@@ -0,0 +1,332 @@
...
+        descriptor: &ComponentColumnDescriptor,
+        column_data: &Box<dyn ArrowArray>,
+    ) -> Result<Self, DisplayRecordBatchError> {
+        match column_data.data_type() {
+            DataType::Null => Ok(Self::Null),
+            DataType::List(_) => Ok(Self::ListArray(
+                column_data
+                    .as_any()
+                    .downcast_ref::<ArrowListArray<i32>>()
+                    .expect("sanity checked")
```
emilk, line 55: If this error can't happen, we should use `.unwrap()` and explain why it can't happen.

If it _can_ happen, we should return a proper `Err` instead.

Crashing with "sanity checked" as the message is not very helpful to anyone

crates/viewer/re_space_view_dataframe/src/display_record_batch.rs
```diff
@@ -0,0 +1,332 @@
...
+    #[allow(clippy::too_many_arguments)]
+    fn data_ui(
+        &self,
+        ctx: &ViewerContext<'_>,
+        ui: &mut egui::Ui,
+        row_id: RowId,
+        latest_at_query: &LatestAtQuery,
+        entity_path: &EntityPath,
+        component_name: ComponentName,
+        index: usize,
```
emilk, line 88: index of what? instance index?

crates/viewer/re_space_view_dataframe/src/display_record_batch.rs
```diff
@@ -0,0 +1,332 @@
...
+
+impl DisplayColumn {
+    #[allow(clippy::borrowed_box)] // https://github.com/rust-lang/rust-clippy/issues/11940
+    fn try_new(
+        column_schema: &ColumnDescriptor,
+        column_data: &Box<dyn ArrowArray>,
+    ) -> Result<Self, DisplayRecordBatchError> {
+        match column_schema {
+            ColumnDescriptor::Control(desc) => {
+                if desc.component_name == ComponentName::from("rerun.controls.RowId") {
```
emilk, line 145:
```suggestion
                if desc.component_name == ComponentName::row_id() {
```

crates/viewer/re_space_view_dataframe/src/display_record_batch.rs
```diff
@@ -0,0 +1,332 @@
...
+                    let row_id_counters = counters
+                        .as_any()
+                        .downcast_ref::<ArrowPrimitiveArray<u64>>()
+                        .expect("sanity checked")
+                        .clone();
+
+                    Ok(Self::RowId {
+                        //descriptor: desc,
+                        row_id_times,
+                        row_id_counters,
```
emilk, line 171: minor nit: but "times" and "counters" only makes sense for `RowId` as `TUID`. If we switch to UUID these names becomes nonsense. We also don't treat them as times and counters anywhere in this file, so perhaps it would be better to name time `row_id_hi` and `row_id_lo` or something? 🤷

emilk: Also, why are we storing `RowId` as two arrays of `u64` instead of a single array of `[u8; 16]` ? 🤔

emilk (approved):

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
abey79: Good call. I'll make a separate, follow-up PR for visibility, as it may affect @jleibs

crates/viewer/re_space_view_dataframe/src/display_record_batch.rs
abey79: I'm surprised by this comment. It was my understanding that `expect(reason)` _precisely_ means "unwrap-that-cannot-happen-because-$reason".

In this particular case, it cannot happen (short of a bug in `re_dataframe`). This is pattern that's often used in Clement's code.

crates/viewer/re_space_view_dataframe/src/display_record_batch.rs
abey79: I'm basically following what is being done in `re_datastore`. I expect this to be carved in stone with the sorbet spec, so let's adjust the implementation when we have that (fundamentally, the purpose of `DisplayRecordBatch` and friends is to decode a sorbet-compliant `RecordBatch` for the purpose of ui).

crates/viewer/re_space_view_dataframe/src/display_record_batch.rs
emilk: Fair enough, but then the error message should be something useful:

`.expect("There is bug in the code at …")`

abey79: OK, I've updated the message.

crates/viewer/re_space_view_dataframe/src/display_record_batch.rs
abey79: Related:
- #6818

Issue #7373, opened 2024-09-09
Warn on unsupported video


jprochazk:
When a user loads a video which we can't load for some reason (failed to load, unsupported container format, unsupported codec, etc.), we should report that to them in some way other than a short-lived warning notification.

Some error cases are:
- Firefox doesn't support `WebCodecs` outside of Nightly
- Safari doesn't support AV1
- Unsupported video container format
- Unsupported video codec (inside the container)
- Container parse error
- Codec configuration error
- Decode error (`WebCodecs`)

gavrelina:
A ui/ux proposal! 
- highlight the component in the stream that causes an error to improve its discoverability
- in the error copy itself mention explicitly what's wrong and why and how users can recover from it (i.e. linking to github issue to follow or docs)
- visually display smth explicitly "broken" instead of the video in the Viewer, like this:

<img width="1301" alt="Screenshot 2024-09-09 at 16 00 25" src="https://github.com/user-attachments/assets/163aadfb-f72d-4ce0-855e-4dfec5b3facf">

gavrelina:
@Wumpf makes a lot of sense! 

About the entity msg => it would also then be visible if users click on it via the blueprint, right? that would also largly increase the discoverability! I do worry (slightly) that copying it in few places comes across as "omg! so many errors everywhere, panic!", hence my initial idea was to only show at Blob level and highlight Blob in red so that users would go there to find it. so kinda direct them with color nudging :) what do you think? 

And moving forward from here it still makes sense to revise (revive?) the different gradations and make this thing work : https://github.com/rerun-io/rerun/issues/4304

jprochazk:
From: https://github.com/rerun-io/rerun/pull/7398#discussion_r1753388072

> I don't know if you've experienced this scenario: You're playing a video on the web, and it suddenly fails with an error of some kind. The player just stops, and the controls become inactive, so you can't seek to a different point in the video anymore. But then you reload the page, and seek after the point where the player previously stopped with an error, and the video continues playing back just fine.
> 
> I think we could do a lot better than web video players here, by always allowing seek even in the case of an error, and more importantly also highlight which segments are broken on the timeline, so the user knows exactly where to seek next to try and catch a part of the video which _can_ be played back.
> 
> This kind of behavior would definitely follow our "handle anything the user throws at it" philosophy

gavrelina:
@jprochazk I can already picture how we mark some specific events/part on the timeline where one encounters an "error"!

PR #7398, opened 2024-09-10
Keep redrawing until video decoder is done decoding requested frame

Wumpf:
### What

* bits and pieces for https://github.com/rerun-io/rerun/issues/7373
    * but far from having that solved

Video decoder forwards now more rich status information upon decode.
This is useful for error handling and - the immediate motivation - to keep redrawing while we're waiting on pending frames.

Before, when clicking somewhere on the timeline, we wouldn't redraw until the mouse moves, which for back-seeking (and far forward-seeking) usually means that we don't get the new frame until the decoder catches up.

I also took the liberty to move the video re_renderer module around a bit - it was falsely sorted under `renderer` which are all about renderers that process draw data and get submitted to a view builder (which is not the case here!).

https://github.com/user-attachments/assets/623d712b-384b-4ffa-bcbf-3a80885ccef6

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7398?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7398?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7398)


crates/viewer/re_renderer/src/video/decoder/web.rs
```diff
@@ -268,6 +263,7 @@ impl VideoDecoder {
         chunk.set_duration(sample.duration.as_f64());
         let Some(chunk) = EncodedVideoChunk::new(&chunk)
             .inspect_err(|err| {
+                // TODO(#7373): return this error once the decoder tries to return a frame for this sample. how exactly?
```
jprochazk, line 266: We may have to set the decoder into an "error state" if it fails to decode a frame, as there may not be a reasonable way to continue decoding. In that case we would just store the error, and close the underlying decoder.

The only reason why decoding should fail is if their video is incorrectly encoded or corrupted in some way. We could for example try again from the next keyframe, as long as the user requests it. So we'd store the _segment_ which is broken inside the error state, alongside the error message from the decoder, and present that to the user until they seek to a different segment.

Wumpf: makes sense. I suppose until we have a concrete example of a video that runs into this case to experiment with we shouldn't go overboard with recovery strategies as we don't know how they will work out. But marking segments as broken seems like a good strategy 👍 :)

jprochazk: I don't know if you've experienced this scenario: You're playing a video on the web, and it suddenly fails with an error of some kind. The player just stops, and the controls become inactive, so you can't seek to a different point in the video anymore. But then you reload the page, and seek after the point where the player previously stopped with an error, and the video continues playing back just fine.

I think we could do a lot better than web video players here, by always allowing seek even in the case of an error, and more importantly also highlight which segments are broken on the timeline, so the user knows exactly where to seek next to try and catch a part of the video which _can_ be played back.

This kind of behavior would definitely follow our "handle anything the user throws at it" philosophy

jprochazk (approved):
Nice!

crates/viewer/re_renderer/src/video/decoder/web.rs
jprochazk: (actually, I will turn this into a comment on the existing issue)

PR #7405, opened 2024-09-11
Make the choice to dictionary-encode data results explicit

jleibs:
### What
- When requesting a Column explicitly, you can now use the new `ColumnDescriptor::DictionaryEncoded(descriptor)` to indicate that you want a dictionary-encoded view of a column.
- Components that are neither part of the chunk, nor dictionary encoded, result in row-slices instead.
- If there are no row-slices, we preserve the old full-chunk based join mechanism, giving large record-batches.
- If there are row-slices, you get single rows.

Future work:
- https://github.com/rerun-io/rerun/issues/7414

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7405?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7405?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7405)

### DIFF
--- a/crates/store/re_chunk_store/src/dataframe.rs
+++ b/crates/store/re_chunk_store/src/dataframe.rs
@@ -22,6 +22,47 @@ use crate::RowId;
 
 // --- Descriptors ---
 
+/// When selecting secondary component columns, specify how the joined data should be encoded.
+///
+/// Because range-queries often involve repeating the same joined-in data multiple times,
+/// the strategy we choose for joining can have a significant impact on the size and memory
+/// overhead of the `RecordBatch`.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
+pub enum JoinEncoding {
+    /// Slice the `RecordBatch` to minimal overlapping sub-ranges.
+    ///
+    /// This is the default, and should always be used for the POV component which defines
+    /// the optimal size for `RecordBatch`.
+    ///
+    /// This minimizes the need for allocation, but at the cost of `RecordBatch`es that are
+    /// almost always smaller than the optimal size. In the common worst-case, this will result
+    /// in single-row `RecordBatch`es.
+    #[default]
+    OverlappingSlice,
+
+    /// Dictionary-encode the joined column.
+    ///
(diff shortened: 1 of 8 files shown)


crates/store/re_chunk_store/src/dataframe.rs
```diff
@@ -37,23 +37,29 @@ pub enum ColumnDescriptor {
     Control(ControlColumnDescriptor),
     Time(TimeColumnDescriptor),
     Component(ComponentColumnDescriptor),
+    DictionaryEncoded(ComponentColumnDescriptor),
```
emilk, line 40: This reads like there are four types of columns, and `DictionaryEncoded` is neither a `Control`, `Time` or a `Component` column. But really, I believe it is just another `Component` column, just encoded differently.

We could introduce another layer here, like:

```rs
pub enum ColumnDescriptor {
    Control(ControlColumnDescriptor),
    Time(TimeColumnDescriptor),
    Component(ComponentColumnDescriptorThingy),
}

struct ComponentColumnDescriptorThingy {
    col_descr: ComponentColumnDescriptor,
    encoding: ColumnEncoding,
}

enum ColumnEncoding {
    ListArray, // or whatever non-dictionary-encoded component columns use
    Dictionary,
}
```

I think structurally that would be the cleanest, and would open up for more encoding options in the future.

---- 

I can understand if you think that's a bit overkill right now though, so if you prefer the current structure of the PR, at least name them better and add some docstring to explain them:

``` rs
pub enum ColumnDescriptor {
    /// A control column (`RowId`) encoded with…
    Control(ControlColumnDescriptor),

    /// A timeline column encoded with…
    Time(TimeColumnDescriptor),

    /// A component column encoded with…
    ListArrayComponent(ComponentColumnDescriptor),

    /// A component column encoded with…
    DictionaryComponent(ComponentColumnDescriptor),
```

PR #7418, opened 2024-09-13
Update to latest egui

emilk:
### What
Uses the latest egui containing this gem:
* https://github.com/emilk/egui/pull/5059

So far `request_discard` (running a second pass) is only used in these situations:
* For when a `Grid` first shows up
* When an egui_table first shows up
* When asked to auto-size an `egui_table` row

Of course, running a second pass has some costs. Luckily, we already execute our most expensive stuff (the viewport) last, and so we can easily early-out from things like point cloud upload/rendering, as shown in this puffin-capture of a frame with two passes in it:

image

NOTE: this flamegraph is of a DEBUG build - the point is not the relative times, but that `pass 0` becomes very cheap, if the `request_discard` is called from one of the non-viewport panels (e.g. the selection panel).

There might still be cases where we do expensive stuff twice, e.g. if something calls `request_discard` in a space view tooltip.

If we come to regret this, we can disable all second passes by setting `egui::Options::max_passes` to `1`.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7418?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7418?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7418)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1777,7 +1777,7 @@ checksum = "68b0cf012f1230e43cd00ebb729c6bb58707ecfa8ad08b52ef3a4ccd2697fc30"
 [[package]]
 name = "ecolor"
 version = "0.28.1"
-source = "git+https://github.com/emilk/egui.git?rev=6b7f4312373a301a4cdf7d99a0d546acd34bcd66#6b7f4312373a301a4cdf7d99a0d546acd34bcd66"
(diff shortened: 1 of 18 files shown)


Wumpf (approved):
sounds promising!

Issue #7368, opened 2024-09-06
Video frame references


emilk:
### Data model

```rust
/// Maps a time point in a video to one or more Rerun timelines.
// TODO: a better name (we're reference a video _time_, not a frame)
component struct VideoFrame {
    /// References the closest video frame to this time.
    /// (closest-to instead of latest-at to forgive rounding errors)
    video_time: i64,
    
    time_mode: VideoTimeMode
    
    /// Reference to an entity with a `VideoAsset`
    /// (OR to an entity with a BlobURI in it, for extra indirection)
    // null = self
    video: Option<EntityPath>,
}

enum VideoTimeMode {
    PtsNanoseconds, // Overkill accuracy for future proofing

   // Save for future: FrameNr
}
```

### SDK
Our logging SDK should had a helper for generating all `VideoFrame` for a video.

```py
mp4_blob = load_file("foo.mp4")

rr.log("video", VideoAsset(content=mp4_blob))

# Be explicit:
rr.send_columns("video",
    capture_time=[10.0, 10.16, 10.33, 10.50],
    [VideoFrame(ms=0), VideoFrame(ms=16), VideoFrame(ms=33), VideoFrame(ms=50)]
);

# Or use our helper function:
rr.send_video_frames(mp4_blob, capture_time=10.0)
```

### Viewer
Our `VideoVisualizer` queries for `VideoFrame`.
We follow a `VideoFrame` via another latest-at query on the given entity path.

A `VideoAsset` without `VideoFrame` will not show up in any space view.

Selecting a video should ideally show that video in the selection view (with independent playback).
It should also be able to show the actual frame timestamps in the video.

Issue #6532, opened 2024-06-10
API / dataloader for logging a video file directly


jleibs:
It is common for users to have video files they want to insert into Rerun.

Eventually, we would like to support video compression directly in the viewer (https://github.com/rerun-io/rerun/issues/5815), but doing so in a cross-platform way (including the browser) presents several challenges.

However, at present, we require users to decode video themselves and then log the individual image frames to Rerun. This is challenging to do performantly, generally requiring dependencies on external libraries, and often times multi-threaded/multiprocessing approaches to help with throughput.

It would be nice to offer a user-space API that allowed users to just directly submit a file-path or video-encoded bytestream and let Rerun take care of converting to a stream of images.

Although our data-loader APIs make it possible to do this as a plugin, it might be worth offering a dedicated API for this. In particular, a dedicated API would allow us to be explicit about how frame-ids map to time-points, which is not something we currently have the ability to do through data-loaders.

This is a situation where it would likely make sense to do the implementation in Rust once and then offer bindings through to the other languages.

emilk:
* Ongoing work in https://github.com/rerun-io/rerun/pull/7261

jleibs:
Needs to produce video frame references

Issue #7420, opened 2024-09-16
Can't show the same video in multiple space views


Wumpf:
We added the ability to reference videos, but our video decoder goes crazy if it has to show frames from different locations at the same time (that's expected the way it's written). Therefore:

* Either remove video reference feature for now
* Or add a hashmap of "decode tracks": Each decode track is just another decoder using the same video data (arc)
    * identification via entity path (+ override entity path if there. Super important otherwise overrides will cause this issue again!)
    * aggressively GC when unused for a frame

I think we can implement that track mechanism easily enough. But if it turns out to be a hassle we can cut back!

emilk:
This should also be able to unlock playing back a video in the selection panel (preview selected video)

Wumpf:
I don't think that's a good idea universally, at least not the simple solution of just looping a video in there: High resolution videos may take a serious amount of processing & memory that we don't want to waste on a small preview. Also if it just keeps playing in there that might be highly distractive. So at least some controls would be needed.
I'd rather pick a still from the video (how?) and discard the decoder once possible

PR #7421, opened 2024-09-16
Expose utility for logging video frames for an entire video

Wumpf:
### What

* Fixes https://github.com/rerun-io/rerun/issues/7368
   *  creating new issues for the previously remaining items
* Fixes  https://github.com/rerun-io/rerun/issues/6532

Had to patch this through to all 3 sdks which naturally takes quite a bit of extra machinery.
Decided to go with raw ns here since it's easiest to pass through FFI and is the most versatile (hopefully our `VideoFrameReference` typing will be solved with tags instead of per component enums in the future, making this more equivalent).

Changed snippets around. There's now:
* a sample with two single frozen frames, showing next to each other. Demonstrating that video frames can be used individually and without any extra utilities
   * this doesn't quite work yet due to #7420, so this can be regarded a bit of a work in progress
* sample using the new frame extraction utility, essentially a "send full video" which we may encapsulate into a higher level utility at some point. It's quite short though in Python, so I'm not too worried about this!

Asset drag & drop got updated to also use this utility.
All this churn also led me to changing how `re_video` is used a bit, making the entry point more highlevel. Media type got a bit annoying there because we can't afford `re_types` dependencies in the C++ & Rust SDKs. The solution here is to have independent media type parsing for videos in `re_video`.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7421?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7421?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)
* [x] pass `main` ci

- [PR Build Summary](https://build.rerun.io/pr/7421)

### DIFF
--- a/crates/store/re_video/src/lib.rs
+++ b/crates/store/re_video/src/lib.rs
@@ -1,6 +1,7 @@
 //! Video decoding library.
 //!
-//! The entry point is [`load_mp4`], which produces an instance of [`VideoData`].
+//! The entry point is [`VideoData::load_from_bytes`]
+//! which produces an instance of [`VideoData`] from any supported video container.
 
 mod mp4;
 


crates/store/re_query/src/cache.rs
```diff
@@ -100,7 +100,7 @@ impl std::fmt::Debug for Caches {
         let mut strings = Vec::new();
 
         strings.push(format!(
-            "[Entities that must be checked for clears @ {store_id}]\n"
+            "[entities that must be checked for clears @ {store_id}]\n"
```
jleibs, line 103: Did some new lint cause this?

crates/store/re_types/definitions/rerun/components/annotation_context.fbs
```diff
@@ -4,7 +4,7 @@ namespace rerun.components;
 
 /// The annotation context provides additional information on how to display entities.
 ///
-/// Entities can use [datatypes.ClassId]s and [datatypes.KeypointId]s to provide annotations, and
+/// entities can use [datatypes.ClassId]s and [datatypes.KeypointId]s to provide annotations, and
```
jleibs, line 7: Starting to wonder if this was an accidental global rename?

crates/store/re_video/src/lib.rs
```diff
@@ -22,6 +21,43 @@ pub struct VideoData {
...
+
+        match media_type.as_str() {
+            "video/mp4" => mp4::load_mp4(data),
+            media_type => Err(VideoLoadError::UnsupportedMediaType(media_type.to_owned())),
+        }
+    }
+
+    /// Determines the presentation timestamps of all frames inside a video, returning raw time values.
+    ///
+    /// Returned timestamps are in nanoseconds since start and are guaranteed to be monotonically increasing.
```
jleibs, line 51: I don't see anything enforcing this guarantee. At a minimum I would expect a check for the condition and an error pathway inside of `mp4::load_mp4`. Isn't this sort of what we ran into with h264?

crates/store/re_video/src/mp4.rs
```diff
@@ -94,3 +94,39 @@ pub fn load_mp4(bytes: &[u8]) -> Result<VideoData, VideoLoadError> {
         segments,
     })
 }
+
+/// Returns whether a buffer is MP4 video data.
+///
+/// From `infer` crate.
+pub fn is_mp4(buf: &[u8]) -> bool {
```
jleibs, line 101: lol, not error prone at all :see_no_evil:

crates/top/rerun_c/src/video.rs
```diff
@@ -0,0 +1,51 @@
...
+pub extern "C" fn rr_video_asset_read_frame_timestamps_ns(
+    video_bytes: *const u8,
+    video_bytes_len: u64,
+    media_type: CStringView,
+    alloc_context: *mut std::ffi::c_void,
+    alloc_func: Option<
+        extern "C" fn(context: *mut std::ffi::c_void, num_timestamps: u32) -> *mut i64,
+    >,
+    error: *mut CError,
+) -> *mut i64 {
```
jleibs, line 14: This bare pointer in the signature seems kind of awkward and hard-to-use without knowledge of the `num_timestamps` param that was passed into the alloc_func.

Do we need this for anything / does it make some common pattern more ergonomic?

crates/top/rerun_c/src/video.rs
```diff
@@ -0,0 +1,51 @@
+use crate::{CError, CErrorCode, CStringView};
+
+#[allow(unsafe_code)]
+#[no_mangle]
+pub extern "C" fn rr_video_asset_read_frame_timestamps_ns(
```
jleibs, line 5: More work than you likely want to tackle today, but once we have "dataframe" APIs in rerun_c, I think this is a good candidate to return the same arrow-based columnar structures we use in our query APIs instead of a bare array like this.

crates/viewer/re_context_menu/src/actions/add_entities_to_new_space_view.rs
```diff
@@ -12,9 +12,9 @@ use crate::{ContextMenuAction, ContextMenuContext};
 /// Create a new space view containing the selected entities.
 ///
 /// The space view is created next to the clicked item's parent view (if a data result was clicked).
-pub(crate) struct AddEntitiesToNewSpaceViewAction;
+pub(crate) struct AddentitiesToNewSpaceViewAction;
```
jleibs:
```suggestion
pub(crate) struct AddEntitiesToNewSpaceViewAction;
```

rerun_cpp/src/rerun/error.hpp
```diff
@@ -57,8 +57,12 @@ namespace rerun {
         ArrowFfiSchemaImportError,
         ArrowFfiArrayImportError,
 
+        // Utility errors.
+        _CategoryUtilities = 0x0001'0000,
+        VideoLoadError,
+
         // Errors relating to file IO.
-        _CategoryFileIO = 0x0001'0000,
+        _CategoryFileIO = 0x0010'0000,
```
jleibs, line 65: Nit: potentially makes this a subtley abi-breaking change, though not in a way I expect matters in practice.

rerun_cpp/src/rerun/result.hpp
```diff
@@ -27,7 +27,7 @@ namespace rerun {
         /// Construct a result from an error, default constructing the value.
         Result(rerun::Error _error) : value(), error(std::move(_error)) {}
 
-        /// Construct a result from an arrow status, default constructing the value.
+        /// Construct a result from an arrow setatus, default constructing the value.
```
jleibs, line 30:
```suggestion
        /// Construct a result from an arrow status, default constructing the value.
```

Wumpf, line 30:
```suggestion
        /// Construct a result from an arrow status, default constructing the value.
```

rerun_py/rerun_sdk/rerun/archetypes/asset_video_ext.py
```diff
@@ -59,3 +63,19 @@ def __init__(
...
+
+        Throws a runtime exception if the video cannot be read.
+        """
+        if self.blob is not None:
+            video_buffer = self.blob.as_arrow_array()
+        else:
+            raise RuntimeError("Asset video has no video buffer")
+
+        if self.media_type is not None:
+            media_type = self.media_type.as_arrow_array().storage[0].as_py()
```
jleibs, line 79: I think `to_numpy()` could be more efficient here (if it works).  `as_py()` will convert to a python list which I think carries some more overhead, though given the data is known to be uniform, it's possible the interpreter does something smart here.

jleibs (approved):

crates/store/re_video/src/mp4.rs
Wumpf: 🙈

Wumpf: happy to take suggestions ;-)

crates/store/re_video/src/lib.rs
Wumpf: hmm now that you mention it, I do think there might be an issue still here. AVC doesn't ever have problems with this, but the mp4 container will have this issue once we properly support h264 since the better order for the container is decode frame order rather than presentation frame order!

crates/top/rerun_c/src/video.rs
Wumpf: needed it orginally but didn't bother to remove it. Let's clean this up!

rerun_py/rerun_sdk/rerun/archetypes/asset_video_ext.py
Wumpf: will check!

crates/top/rerun_c/src/video.rs
Wumpf: agreed on both accounts 😄

rerun_py/rerun_sdk/rerun/archetypes/asset_video_ext.py
Wumpf: `AttributeError: 'pyarrow.lib.StringScalar' object has no attribute 'to_numpy'`
so, no

Wumpf:
@rerun-bot full-check

PR #7422, opened 2024-09-16
Add test for UI wakeup

emilk:
### What
* Part of https://github.com/rerun-io/rerun/issues/7378

Tested:

### Mac
* [x] ✅ Native alt-tabbed
* [x] ✅ Native minimize
* [x] ✅ Browser alt-tabbed
* [ ] ❌ Web background tab

### Windows
* [ ] Native alt-tabbed
* [ ] Native minimize
* [ ] Browser alt-tabbed
* [ ] Web background tab

### Linux
* [ ] Native alt-tabbed
* [ ] Native minimize
* [ ] Browser alt-tabbed
* [ ] Web background tab

### Checklist
* [x] I have read and agree to [Contributor Guide](https://github.com/rerun-io/rerun/blob/main/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/rerun-io/rerun/blob/main/CODE_OF_CONDUCT.md)
* [x] I've included a screenshot or gif (if applicable)
* [ ] I have tested the web demo (if applicable):
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/{{pr.number}}?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/{{pr.number}}?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)
* [x] The PR title and labels are set such as to maximize their usefulness for the next release's CHANGELOG
* [x] If applicable, add a new check to the [release checklist](https://github.com/rerun-io/rerun/blob/main/tests/python/release_checklist)!
* [x] If have noted any breaking changes to the log API in `CHANGELOG.md` and the migration guide

- [PR Build Summary](https://build.rerun.io/pr/{{pr.number}})
- [Recent benchmark results](https://build.rerun.io/graphs/crates.html)
- [Wasm size tracking](https://build.rerun.io/graphs/sizes.html)

To run all checks from `main`, comment on the PR with `@rerun-bot full-check`.

PR #7425, opened 2024-09-16
Bug fix: always show latest data in follow-mode

emilk:
### What
* Found while testing https://github.com/rerun-io/rerun/pull/7422

Always show the latest data in follow-mode, by advancing the time at the start of the frame rather than at the end of it.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7425?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7425?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7425)

### DIFF
--- a/crates/viewer/re_viewer/src/app_state.rs
+++ b/crates/viewer/re_viewer/src/app_state.rs
@@ -272,6 +272,10 @@ impl AppState {
             focused_item,
         };
 
+        // We move the time at the very start of the frame,
+        // so that we always show the latest data when we're in "follow" mode.
+        move_time(&ctx, recording, rx);
+
         // Update the viewport. May spawn new views and handle queued requests (like screenshots).
         viewport.on_frame_start(&ctx);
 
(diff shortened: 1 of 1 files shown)


crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -466,41 +470,6 @@ impl AppState {
         // Process deferred layout operations and apply updates back to blueprint
         viewport.update_and_sync_tile_tree_to_blueprint(&ctx);
 
-        {
-            // We move the time at the very end of the frame,
-            // so we have one frame to see the first data before we move the time.
```
emilk, line 471: I don't quite understand this, even though I suspect I wrote it.

The "first" data will only be visible if we move/update the current time so that the latest-at query includes the first data.

Sure, if we receive a few video frames during one app update, the new code will not show the first frame of the video, but I don't think that's wrong, nor do I think the old code would have shown the first frame of the video either.

Wumpf (approved):
this looks sane, but I don't 100% understand the previous incorrect behavior: Follow mode may have ended up showing old data until we get a redraw?

PR #7433, opened 2024-09-17
Display video errors & video blob information

### LINKED ISSUE #7373: Warn on unsupported video

Wumpf:
### What

* Fixes #7373
   * I'm sure there's more to do here and surely more cases to try out whether they display the correct errors, but this PR should handle everything that we detect as an error case, laying the foundation to pass everything through

As part of this work it was necessary to make clearer difference between video loading and video decoding. You'll notice changes in re_renderer's video and the video cache accordingly.
(there's some loose ends here like "don't create a decoder when we're just inspecting a blob" which will be part of #7420)

Generally, decode errors can only occur when showing (!) a video frame, therefore they are shown on screen as part of a the 2D/3D view:
image
(trying to show frames on native)

Note that you can interact with this 2D view just as if error was logged data:

https://github.com/user-attachments/assets/ee609074-6fa6-4b59-ba33-ea6026c0f0d0

On the other hand when we can't read the _blob_ itself (potentially without trying to decode it) the errors show up on it like this:
image

Since this error happens already on loading right now we can't generate frames automatically if you drag&drop in a video like this. But you can still log your frames manually, resulting in this:
image
In the shown example the blob isn't part of the entity since we're referring to a different entity for the blob. Also, this is actually a native screenshot - the avc parsing error comes _before_ decoding, which is why this comes up prior to the lack of decoder.
If we instead have the failing blob on the entity that also has the frame, we get this view:
image

When there's no error, we show now a little bit of information for blobs:
image
(blob selected)
image
(data result with this blob selected)

Some shortcomings:
* some decode errors like "we can't decode this at all because it's not supported", could be shown was warnings the property grid on blobs. This is quite unfavorable from a program structure perspective, but as user I'd be wondering why not?
* dragging in a video format with an unknown codec won't generate frames and therefore no big "can't read this video" which is rather unfortunate. As we progress in separating video container reading & decoding this is going to get better!
* blob errors aren't terribly well formatted / a bit ugly
* error states in streams panel as suggested on the ticket would be nice, but that's a more general ~rabbit hole~ feature I don't want to tackle in the near term

### Checklist


crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -142,3 +157,46 @@ pub fn blob_preview_and_save_ui(
...
+
+fn show_video_blob_info(
+    ui: &mut egui::Ui,
+    ui_layout: UiLayout,
+    video_result: &Result<re_renderer::video::Video, re_renderer::video::VideoError>,
+) {
+    match video_result {
+        Ok(video) => {
+            if ui_layout.is_single_line() {
+                return;
```
emilk, line 169: I think we should show _something_ even in the single-line case. Ideally just everything, but compacted on one line (resolution, duration, codec).

crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -142,3 +157,46 @@ pub fn blob_preview_and_save_ui(
...
+                    PropertyContent::new("Dimensions").value_text(format!(
+                        "{}x{}",
+                        video.width(),
+                        video.height()
+                    )),
+                );
+                ui.list_item_flat_noninteractive(PropertyContent::new("Duration").value_text(
+                    format!(
+                        "{}",
+                        re_log_types::Duration::from_millis(video.duration_ms() as _)
```
emilk, line 186: Can we also show the number of frames in the video?

crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -142,3 +157,46 @@ pub fn blob_preview_and_save_ui(
...
+                    ),
+                ));
+                ui.list_item_flat_noninteractive(
+                    PropertyContent::new("Codec").value_text(video.codec()),
+                );
+            });
```
emilk, line 192: It would be great with a mini-player of the video here!

crates/viewer/re_renderer/src/video/mod.rs
```diff
@@ -66,25 +77,30 @@ impl Video {
...
+        let decoder = Mutex::new(decoder::VideoDecoder::new(render_context, data.clone())?);
 
-        Ok(Self { decoder })
+        Ok(Self { data, decoder })
     }
 
     /// Duration of the video in milliseconds.
     pub fn duration_ms(&self) -> f64 {
-        self.decoder.duration_ms()
+        self.data.duration.as_f64()
```
emilk, line 88: Nit: I have seen so many time unit bugs in my days that I always go for explicitness

```suggestion
        self.data.duration_ms.as_f64()
```

crates/viewer/re_space_view_spatial/Cargo.toml
```diff
@@ -47,6 +47,9 @@ bytemuck.workspace = true
 egui = { workspace = true, features = ["serde"] }
 glam.workspace = true
 hexasphere.workspace = true
+image = { workspace = true, features = [
+  "png",
+] } # For loading embedded images to textures.
```
emilk, line 52: What "embedded images"? Do you mean `Blob`s? If so, why  only `png` and not also `jpeg`?

EDIT: oh I see, it is for the video icon. Maybe "For loading images/icons for the GUI"?

crates/viewer/re_ui/src/ui_ext.rs
```diff
@@ -30,6 +30,21 @@ pub trait UiExt {
...
+    ///
+    /// Use this only if you have a lot of space to spare.
+    /// Don't start the string with "Error: " since this is added automatically.
+    fn error_label_long(&mut self, error_text: &str) -> egui::Response {
+        let label = egui::Label::new(self.ui().ctx().error_text(format!("Error: {error_text}")))
+            .selectable(false)
+            .sense(egui::Sense::click());
+        let response = self.ui_mut().add(label);
+        if response.clicked() {
+            self.ui().ctx().copy_text(error_text.to_owned());
```
emilk, line 43: This is a nice touch, but there is no affordance that tells the user that they can copy the error with a click.

@gavrelina may have a better idea here, but some alternatives:

* Show "Click to copy" in a tooltip on hover
* Make the text selectable instead
* Add a 📋 button next to the message instead

crates/viewer/re_viewer_context/src/cache/image_decode_cache.rs
```diff
@@ -22,7 +21,7 @@ struct DecodedImageResult {
 /// Caches the results of decoding [`re_types::archetypes::EncodedImage`].
 #[derive(Default)]
 pub struct ImageDecodeCache {
-    cache: ahash::HashMap<u64, DecodedImageResult>,
+    cache: ahash::HashMap<Hash64, DecodedImageResult>,
```
emilk, line 24: nice

crates/viewer/re_viewer_context/src/cache/video_cache.rs
```diff
@@ -12,46 +13,42 @@ use std::sync::{
...
 
 impl VideoCache {
+    /// Read in some video data and cache the result.
+    ///
+    /// The `row_id` should be the `RowId` of the blob.
+    /// NOTE: videos are never batched atm (they are mono-archetypes),
+    /// so we don't need the instance id here.
+    ///
+    /// TODO(andreas,jan): This effectively takes a full copy of the video data.
+    /// We should avoid this as much as possible, having pointers back to the raw
```
emilk, line 35: We can pass in a `Blob` here instead, which is ref-counted internally

emilk (approved):
Very nice!

crates/viewer/re_data_ui/src/blob.rs
Wumpf: Maybe I'm confused, but I think if this says single line then this is the line of the blob which already shows the blob's size

image

crates/viewer/re_data_ui/src/blob.rs
Wumpf: that would be awesome indeed! 😄 
But keep in mind that that requires creating a new decoder (decoders are stateful to where they are right now) which can get expensive quickly. Definitely would like to try this though in a future iteration!

crates/viewer/re_data_ui/src/blob.rs
Wumpf: I actually had this in for a moment but then removed it again because it requires iterating over all the samples of a video. Which isn't terribly expensive and could be cached, but then I saw that video players only do this upon read and usually carefully talk about samples, not frames 🤔
@jprochazk what do you think?

crates/viewer/re_renderer/src/video/mod.rs
Wumpf: `duration` is of type `TimeMs`. We should probably just forward that!

crates/viewer/re_ui/src/ui_ext.rs
Wumpf: for context: I copied this from the "small" error text method above. But I do agree!

crates/viewer/re_viewer_context/src/cache/video_cache.rs
Wumpf: Let's start like that here already yes 👍 
However, the problem goes a lot deeper since we pass the data on to mp4. But that moves the todo at least!

crates/viewer/re_data_ui/src/blob.rs
emilk: Ah yes, of course.

crates/viewer/re_ui/src/ui_ext.rs
Wumpf: our `📋` icon is too ugly for this (also tried adding it to the regular `error_label`, but it's just too ugly). going with the minimum here for now whcih is selectable + hover

crates/viewer/re_data_ui/src/blob.rs
Wumpf: putting it in for now, but calling it `Sample count` instead. I figure the contention with "frames" is that it may refer to full frames/iframes.

jprochazk: Yeah, I would be careful about this. Some people may think that `num_frames / duration = fps`, but that's not true, videos may have variable frame rate. At the same time, we don't want to overload users with video codec/container specific stuff that they have to understand, and for all intents and purposes one sample = one frame.

Issue #7439, opened 2024-09-18
Fix component reflection such as to not force the component to `impl Default`


abey79:
Currently, the codegen'd component reflection always emit a call to `C::default()`, forcing the component to `impl Default`. Sometime this really is meaning less (e.g what's a default `EntityPath` or `ComponentName`?).

Funnily the `ComponentReflection` struct actually has an `Option`. The codegen should therefore emit `None` if the component doesn't have `Default` (which the codegen knows from the `attr.rust.derive` attribute).

Edit: the codegen doesn't always know about `Default` if it's manually implemented. So that should be controlled with a new fbs attribute `attr.rust.no_placeholder`

kpreid:
I ran into this myself when [adding `ShowLabels`](https://github.com/rerun-io/rerun/pull/7249/). `ShowLabels` is a component whose fallback value [is _always_ computed from other data,](https://github.com/rerun-io/rerun/blob/6121840ba46528c727197bc71d826a5e9527e869/crates/viewer/re_space_view_spatial/src/visualizers/utilities/labels.rs#L85-L110) and does not have a default that is meaningful at all.

Wumpf:
We discussed this internally quite a bit internally
https://rerunio.slack.com/archives/C041NHU952S/p1726669945999429

The gist of it that we do need guaranteed defaults for anything that's potentially ui editable (for the value to start out with). Overall, the force to default was introduced as a simplification so that anyone querying a fallback can rely on it being there and treat it as an hard error when it isn't.
It is a known issue that `Default` trait us such is slightly misleading semantically here. When we introduced this we discussed the possibility of a separate trait but discarded this because of the added overhead.

Wumpf:
@kpreid `ShowLabels` would be a case where a default value _is_ important. The reason is that one can edit it in the visualizer ui and the ui needs a value to start with.

Wumpf:
Have some work in progress that allows to generate placeholders fully generically so we need lot less of those pointless looking defaults and can _actually_ make use of the fact there's _always_ a placeholder 🙂 
https://github.com/rerun-io/rerun/tree/andreas/automatic-component-placeholders

since we still want to use `Default` for custom placeholders when available I think we'll have to make it opt-out as proposed in https://github.com/rerun-io/rerun/pull/7440
.. unless someone does a codegen thing where we check whether `Default` procmacro is used or there's a `Default` impl in the extension :)

PR #7447, opened 2024-09-19
Generic component placeholders

Wumpf:
### What

* fixes https://github.com/rerun-io/rerun/issues/7439

Introduces generic placeholders, allowing to..
* leave out `Default` implementations in many cases
* fearless request a placeholder because it's guaranteed to be available

Mostly for testing I removed a few known strange defaults:
* `Blob`
* `ImageData`
* `ShowLabels`

.. and tested that all fallback & default displays work fine afterwards. Seemed robust!
I'm not doing a sweep over existing components because that would imply a lot more testing for changed behavior would be necessary and I don't see a lot of value in that right now.

Wrinkles:
* component name must be registered. If that's not the case we're kinda lost, just like in other places in the code. The workaround here is that you get the placeholder for Null datatype
* a bit excessive on the allocation side, some caching might be nice, then we could return references to arrow data instead of giving out boxes again. Unlikely to be an actual perf concern afaik

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7447?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7447?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7447)

### DIFF
--- a/crates/build/re_types_builder/src/codegen/rust/api.rs
+++ b/crates/build/re_types_builder/src/codegen/rust/api.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashSet};
+use std::collections::{BTreeMap, HashMap, HashSet};
 
 use anyhow::Context as _;
 use camino::{Utf8Path, Utf8PathBuf};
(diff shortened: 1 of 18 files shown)


abey79 (approved):
Awesome!

Issue #7353, opened 2024-09-04
Support ui-interaction with video


jprochazk:
We have video frames available in regular GPU textures, so we should be able to do GPU picking for hover UIs

Wumpf:
What's the task here exactly? We don't have gpu picking for images either at all, so what do we want to accomplish here?

emilk:
There is several steps here:

A) support clicking videos to select them
B) showing a magnification of the video on hover
C) actual color picking on hover (read exact RGB value)

Only C requires reading back the texture to CPU, and is also the least urgent

PR #7450, opened 2024-09-19
Various picking code refactor & basic pickable videos

Wumpf:
### What

* Part of #7353
   * still missing hover preview, but this PR gets us to the minimum viable product.

Generalizes the picking-rect mechanism to a large extent to include both videos and error placeholders.
Picking code was in a quite poor state, so most of the changes here are all about making it cleaner and easier to follow. Still some left to be desired but had to stop somewhere ;-)

image

Overview over the refactor (it's almost commit by commit, but I went back & forth on naming a bit and there's _some_ evolution that I didn't clean up):
* replace `PickableImage` with a more versatile `PickableTexturedRect` construct
* each visualizer now has a list of `PickableTexturedRect` to dump its images into directly on the shared `SpatialViewVisualizerData`
   * similar to how ui label collecting works
   * this allows all reasoning about pickable rects to be fully generic, no need to know all visualizers
       * the only exception are depth clouds which aren't images from a picking perspective but get resolved to one 
* pull the `PickingContext` creation out of the mighty `picking` method
* renames on `PickingContext` to make it more clear what's going on
* some more meaningful sub routines for the mighty `picking` method
    * it's still large and complex, but a lot better now 

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7450?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7450?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7450)


crates/viewer/re_space_view_spatial/src/pickable_textured_rect.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Textured rectangle used by the renderer.
+    pub textured_rect: TexturedRect,
+
+    /// Associated data.
+    pub source_data: PickableRectSourceData,
+}
+
+impl PickableTexturedRect {
+    /// How wide the rectangle is in pixels.
+    pub fn pixel_width(&self) -> u64 {
```
emilk, line 40: We should clarify that this is the resolution of the source data, and not the size of the rectangle on screen.

Also, this could be called `resolution()` and return a 2D `[u64; 2]` in case we ever want the height 🤷

crates/viewer/re_space_view_spatial/src/picking.rs
```diff
@@ -67,10 +67,18 @@ pub struct PickingContext {
...
+    /// Note that for 2D spaces these are the final 2D space coordinates.
+    ///
+    /// As of writing, for 3D spaces this is equal to [Self::pointer_in_ui],
+    /// since we don't allow panning & zooming after perspective projection.
+    pub pointer_in_ui_after_pan_and_zoom: glam::Vec2,
+
+    /// Apply pan & zoom to ui coordinates.
+    ///
+    /// See also [`Self::pointer_in_ui_after_pan_and_zoom`].
+    pub ui_pan_and_zoom_from_ui: egui::emath::RectTransform,
```
emilk, line 81: I think we should come up with a better name for this space. "ui_pan_and_zoom" is overlong and complicated.

After zooming and panning, it is not really `ui` related any more. Maybe call it "flat_space" ?

crates/viewer/re_space_view_spatial/src/ui.rs
```diff
@@ -439,54 +408,20 @@ pub fn picking(
...
         if !data_result.is_interactive(ctx) {
             continue;
         }
 
-        // Special hover ui for images and point clouds:
-        let is_depth_cloud = depth_images
-            .depth_cloud_entities
-            .contains(&instance_path.entity_path.hash());
-
-        let picked_image = if hit.hit_type == PickingHitType::TexturedRect || is_depth_cloud {
```
emilk, line 454: Let's remember to test depth-cloud picking

crates/viewer/re_space_view_spatial/src/visualizers/utilities/spatial_view_visualizer.rs
```diff
@@ -1,14 +1,22 @@
...
 
 use super::UiLabel;
-use crate::view_kind::SpatialSpaceViewKind;
+use crate::{view_kind::SpatialSpaceViewKind, PickableTexturedRect};
 
 /// Common data struct for all spatial scene elements.
 ///
 /// Each spatial scene element is expected to fill an instance of this struct with its data.
 pub struct SpatialViewVisualizerData {
+    /// Labels that should be shown rendering egui.
```
emilk, line 10:
```suggestion
    /// Labels that should be shown using egui.
```

crates/viewer/re_space_view_spatial/src/visualizers/videos.rs
```diff
@@ -199,34 +193,50 @@ impl VideoFrameReferenceVisualizer {
...
-                    let rect =
-                        textured_rect_for_video_frame(world_from_entity, video_size, texture);
-                    rectangles.push(rect);
+                    let textured_rect = TexturedRect {
+                        top_left_corner_position: world_from_entity
+                            .transform_point3(glam::Vec3::ZERO),
+                        // Make sure to use the video instead of texture size here,
+                        // since it may be a placeholder which doesn't have the full size yet.
+                        extent_u: world_from_entity.transform_vector3(glam::Vec3::X * video_size.x),
+                        extent_v: world_from_entity.transform_vector3(glam::Vec3::Y * video_size.y),
```
emilk, line 213: I think `video_resolution` is a better name than `video_size`.

At least I think of "resolution" as the immutable thing stored in the file, while "size" could be the display size, which depends on zoom etc.

crates/viewer/re_space_view_spatial/src/visualizers/videos.rs
```diff
@@ -268,7 +279,15 @@ impl VideoFrameReferenceVisualizer {
                         height: dynamic_image.height(),
                     })
                 },
-            )?;
+            );
+
+        let video_error_texture = match video_error_texture_result {
+            Ok(video_error_texture) => video_error_texture,
+            Err(err) => {
+                re_log::error_once!("Failed to show video error icon: {err}");
```
emilk, line 287:
```suggestion
                if let Err(err) = re_log::error_once!("Failed to show video error icon: {err}") {
                    if let Err(err) = eprintln!("Failed to log icon error message: {err}") {
                        if let Err(err) = panic!("Failed to write error error message to to stderr: {err}") {
                            std::system::abort();
                        }
                    }
                }
```

…sorry, I felt silly 😜

emilk (approved):
Code looks good, except for some naming stuff!

Remember to test this properly, including depth cloud picking

crates/viewer/re_space_view_spatial/src/picking.rs
Wumpf: was struggling with this quite a bit. Before it said "space2d" but this really confused me upon revisiting the code which is why I went with the overly verbose thing.
Something I had in here in the meantime was `camera_plane` because if this is 3D then these are coordinates on the camera plane which then is the thing we're panning and zooming (imagine what happens with pan & zoom on the arkit 2D scene).
Maybe a better name than that still would be `camera_ui`-space (+ doc on why it's name this way). What do you think?

crates/viewer/re_space_view_spatial/src/ui.rs
Wumpf: 👍 did already (and ofc while working on it I broke it during iterations ;-)). But I'll make sure to give it another spin prior to merge

crates/viewer/re_space_view_spatial/src/picking.rs
emilk: I like `camera_plane`, since we already have a `ui` coordinate system, so better to keep the names as distinct as possible

PR #7451, opened 2024-09-19
Fix table column auto-sizing

emilk:
The bug was that the amount of allocated space was never reported up the `Ui` chain

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7451?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7451?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7451)

### DIFF
--- a/crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
+++ b/crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
@@ -659,17 +659,15 @@ fn split_ui_vertically<Item, Ctx>(
     let visible_y_range = ui.clip_rect().y_range();
     let total_y_range = ui.max_rect().y_range();
 
-    // Note: converting float to unsigned ints implicitly saturate negative values to 0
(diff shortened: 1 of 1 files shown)


abey79:
It's still relatively easy to reach a "wrong width situation":

- create a new dataframe view (single row showing since latest-at)
- switch to "Define time range" (often still single row showing since the default PoV component is not the correct one)
- chose a suitable PoV component -> many rows appear, several wrong column sizes 

e.g.

image

This probably makes sense since we salt the table with the schema (which doesn't change between step 2 and 3). Maybe adding the pov component to the salt might be a good idea.

abey79 (approved):
So neat to have well behaving columns!

Issue #5253, opened 2024-02-22
Allow overwrite materials on `Asset3D` archetype - color tinting?


Wumpf:
For very primitive Mesh formats like STL it seems to be common practice to provide some way of overwriting the material after the fact (since it's not possible to specify any material properties).
Having a way to change/affect the material of a logged Asset3D might be useful in general, so we should extend `Asset3D` accordingly and do a "best effort" material merging.

See https://wiki.ros.org/urdf/XML/link
URDF allows a `material` property with texture (unusable in STL due to lack of textures) and color (use as tint..?)

oxkitsune:
Implemented in #7458

PR #7458, opened 2024-09-20
Allow overriding albedo color on `Asset3D`

EtaLoop:
### What

The idea is to reproduce behavior of `Mesh3D` regarding color modification on `Asset3D` :
- add `albedo_factor` 
- add `vertex_colors` as in `Mesh3D` 

In order to overwrite white default texture of `Asset3D` loaded from stl and obj (gltf already have albedo_factor in file). 

See: #5253 

Screenshot from 2024-09-18 16-05-54

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7458?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7458?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7458)


crates/viewer/re_renderer/src/importer/obj.rs
```diff
@@ -22,6 +25,8 @@ pub enum ObjImportError {
 pub fn load_obj_from_buffer(
     buffer: &[u8],
     ctx: &RenderContext,
+    vertex_colors: &Option<Vec<Color>>,
+    albedo_factor: &Option<AlbedoFactor>,
```
Wumpf, line 29: I don't think this should be passed into specific mesh format loaders but rather be a universal post processing step.

Which I think is very well illustrated by the fact that this ignored gltf/glb loader

crates/store/re_types/definitions/rerun/archetypes/asset3d.fbs
```diff
@@ -33,4 +33,12 @@ table Asset3D (
...
   /// If it cannot guess, it won't be able to render the asset.
   media_type: rerun.components.MediaType  ("attr.rerun.component_recommended", nullable, order: 2000);
+
+  // --- Optional ---
+
+  /// An optional color for each vertex.
+  vertex_colors: [rerun.components.Color] ("attr.rerun.component_optional", nullable, order: 3100);
+
+  /// A color multiplier applied to the whole asset.
+  albedo_factor: rerun.components.AlbedoFactor ("attr.rerun.component_optional", nullable, order: 3200);
```
Wumpf, line 43: Needs to specify interaction with existing materials

crates/store/re_types/definitions/rerun/archetypes/asset3d.fbs
```diff
@@ -33,4 +33,12 @@ table Asset3D (
   /// If omitted, the viewer will try to guess from the data blob.
   /// If it cannot guess, it won't be able to render the asset.
   media_type: rerun.components.MediaType  ("attr.rerun.component_recommended", nullable, order: 2000);
+
+  // --- Optional ---
+
+  /// An optional color for each vertex.
+  vertex_colors: [rerun.components.Color] ("attr.rerun.component_optional", nullable, order: 3100);
```
Wumpf, line 40: Albedo factor makes sense to me because it can be easily applied to all materials inside a mesh format.
However, vertex color manipulation is pretty unclear to me: A model file has many submeshes and it's really hard to specify how the vertex indices from the vertex colors map to any of the vertices inside the models.
To make matters worse this depends a lot on the concrete model loader: What order are the submeshes in, are all submeshes left as is or can it happen that they flatten out, etc.

Due to those difficulties I'd rather not expose this at all, since it sounds like this might be a constant source of pain.
That is unless we find a way to clearly specify what this field does under all circumstances and make it a mesh/model file independent postprocessing step.

crates/viewer/re_renderer/Cargo.toml
```diff
@@ -53,6 +53,7 @@ re_log.workspace = true
 re_math.workspace = true
 re_tracing.workspace = true
 re_video.workspace = true
+re_types.workspace = true
```
Wumpf, line 56: We really don't want this (huge!) dependency in `re_renderer`. `re_renderer` is designed to act independent of any of the SDK types and should be able to be used without any knowledge of the database or viewer.

crates/viewer/re_renderer/src/importer/stl.rs
```diff
@@ -21,11 +22,21 @@ pub enum StlImportError {
 
 /// Load a [STL .stl file](https://en.wikipedia.org/wiki/STL_(file_format)) into the mesh manager.
 pub fn load_stl_from_buffer(
-    buffer: &[u8],
+    asset3d: &Asset3D,
```
Wumpf, line 25: as mentioned above, this type shouldn't be here. Also why does it need to be and why do the other loaders don't have them

crates/viewer/re_renderer/src/importer/stl.rs
```diff
@@ -21,11 +22,21 @@ pub enum StlImportError {
...
     ctx: &RenderContext,
+    _texture_key: u64,
```
Wumpf, line 27: why is this needed?

Wumpf (changes requested):
Thanks for your contribution!! 🙂 

`AlbedoFactor` makes a lot of sense to me since this can be applied on all mesh/model formats, would love to get this in. But I don't see yet how we can add vertex_colors in a tractable & defined way. However, maybe we can come up with some rule for it?
Very relatedly, we have to keep in mind that more model formats may be added in the future and we need to avoid ugly surprises in that context. Therefore, I think all additional properties on `Assed3D` should be "postprocesses" on the loaded mesh. This makes it more robust against future formats and makes it a lot easier to specify what they do.

crates/store/re_types/definitions/rerun/archetypes/asset3d.fbs
Wumpf: the meshes defined by a complex format like glb/gltf may already have material properties very similar to this. So this should say something about how this combines with existing colors that may already be set

Wumpf:
let me know when you think this is ready for review again! :)
(or use the little "re-request review" button so this shows up in my review queue again)

crates/store/re_types/definitions/rerun/archetypes/asset3d.fbs
```diff
@@ -33,4 +33,12 @@ table Asset3D (
   /// If omitted, the viewer will try to guess from the data blob.
   /// If it cannot guess, it won't be able to render the asset.
   media_type: rerun.components.MediaType  ("attr.rerun.component_recommended", nullable, order: 2000);
+
+  // --- Optional ---
+
+  /// A color multiplier applied to the whole asset.
+  ///
+  /// For mesh who already have albedo_factor in materials,
+  /// it will be overwritten by actual albedo_factor of Asset3D (if specified).
```
Wumpf, line 42:
```suggestion
  /// For meshes which already define an `albedo_factor` via their materials,
  /// it will be overwritten by this `albedo_factor` if specified.
```

crates/store/re_types/definitions/rerun/archetypes/asset3d.fbs
Wumpf: yes! I think I'll slightly adjust the formulation, but that is a definition we can work with (:
I believe I would have gone with multiplying with the existing, but that that's a choice 🤷. Arguably overriding the the more powerful tool 👍

crates/store/re_types/src/archetypes/asset3d_ext.rs
```diff
@@ -32,10 +32,10 @@ impl Asset3D {
     #[inline]
     pub fn from_file_contents(contents: Vec<u8>, media_type: Option<impl Into<MediaType>>) -> Self {
         let media_type = media_type.map(Into::into);
-        let media_type = MediaType::or_guess_from_data(media_type, &contents);
```
Wumpf, line 35: Any particular reason you removed this?
My hunch is that this was added so that the media type shows more eagerly in the viewer if it wasn't specified.

crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
```diff
@@ -131,6 +131,20 @@ pub struct MeshInstance {
     pub picking_layer_id: PickingLayerId,
 }
 
+impl Clone for MeshInstance {
```
Wumpf, line 134: Use `#[derive(Clone)]` on `MeshInstance` instead

crates/store/re_types/definitions/rerun/archetypes/asset3d.fbs
EtaLoop: My formulation wasn't very pretty yes :smiling_face_with_tear:

EtaLoop: My formulation wasn't very pretty yes :smiling_face_with_tear:

crates/viewer/re_space_view_spatial/src/visualizers/assets3d.rs
```diff
@@ -32,48 +32,53 @@ impl Default for Asset3DVisualizer {
...
-                        query_result_hash: Hash64::ZERO,
-                        media_type: data.media_type.clone().map(Into::into),
+                    key.clone(),
+                    AnyMesh::Asset {
+                        asset: &Asset3D {
+                            blob: data.blob.clone().into(),
+                            media_type: data.media_type.clone().map(Into::into),
+                            albedo_factor: data.albedo_factor.copied(),
+                        },
+                        texture_key: re_log_types::hash::Hash64::hash(&key).hash64(),
```
Wumpf, line 80: I don't get why this is called texture key when the key is simply about other properties of the Asset3D archetype that we want to have to be part of the cache key. But practically this has nothing to do with textures - neither the intention nor what you're feeding in (which shares some of the exact same parts of `Asset3D` above).
I think the only thing you actually need to do is to populate the `query_result_hash` in the same way as it is done on `Mesh3D` - this key will change every time the albedo color override changes. Everything else can stay the same and no `texture_key` should be needed.

crates/viewer/re_space_view_spatial/src/mesh_cache.rs
```diff
@@ -36,7 +36,13 @@ pub struct MeshCache(HashMap<RowId, HashMap<MeshCacheKey, Option<Arc<LoadedMesh>
 /// Either a [`re_types::archetypes::Asset3D`] or [`re_types::archetypes::Mesh3D`] to be cached.
 #[derive(Debug, Clone, Copy)]
 pub enum AnyMesh<'a> {
-    Asset(&'a re_types::archetypes::Asset3D),
+    Asset {
+        asset: &'a re_types::archetypes::Asset3D,
+
+        /// If there are any textures associated with that asset (albedo etc), they use this
+        /// hash for texture manager lookup.
+        texture_key: u64,
```
Wumpf, line 44: See `assets3d.rs` visualizer, this isn't needed.
The only reason the `Mesh` variant needs this is to account for the fact that there's additional archetypes on the same entity that are taken into consideration

crates/viewer/re_space_view_spatial/src/mesh_loader.rs
```diff
@@ -37,41 +45,58 @@ impl LoadedMesh {
...
-            MediaType::STL => re_renderer::importer::stl::load_stl_from_buffer(bytes, render_ctx)?,
+            MediaType::STL => {
+                re_renderer::importer::stl::load_stl_from_buffer(bytes, render_ctx, texture_key)?
+            }
             _ => anyhow::bail!("{media_type} files are not supported"),
         };
 
-        let bbox = re_renderer::importer::calculate_bounding_box(&mesh_instances);
+        let prev_mesh_instances = mesh_instances.clone();
+        let mesh_instances = add_albedo_factor_to_mesh(mesh_instances, asset3d, render_ctx);
```
Wumpf, line 68: I don't see why `mesh_instances` is cloned. You should instead modify it in-place, that would be a lot cheaper & easier to follow.

Edit: You're actually cloning it 2x now since `add_albedo_factor_to_mesh` creates another copy of them. Please avoid this!

crates/viewer/re_space_view_spatial/src/mesh_loader.rs
```diff
@@ -193,6 +218,37 @@ impl LoadedMesh {
...
+        };
+
+        Ok(vec![MeshInstance::new_with_cpu_mesh(
+            Arc::new(GpuMesh::new(render_ctx, &mesh)?),
+            Some(Arc::new(mesh)),
+        )])
+    } else {
+        Ok(mesh_instances)
+    }
+}
```
Wumpf, line 250: it looks like you're jumping through many many hoops & clones instead of just making `MeshInstance` `mut` and swap out the albedo factor. What you want to do is:
* if no albedo factor on mesh defined, early out
* otherwise, go through all meshes, edit the albedo factor

crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
Wumpf: also I don't think you need this clone at all

Wumpf (changes requested):

crates/store/re_types/src/archetypes/asset3d_ext.rs
EtaLoop: Hmmm... that's my bad, I don't know what happened here sorry.

crates/viewer/re_space_view_spatial/src/mesh_loader.rs
EtaLoop: Correct me if I'm wrong, but the `mesh` of `MeshInstance` is an `Option<Arc<Mesh>>`, so I can't modify it because it was instantiated without Mutex.
I tried using `Arc::get_mut()`, it return Some(Mesh) so there is no other Arc.
It works : back in `load_asset3d_parts` after the call of `add_albedo_factor_to_mesh`, `albedo_factor` in `MeshInstance.mesh` is modified. But for some reason the asset still appears white.
Do you have an idea of what is happening here ? Could it be overwritten somewhere else ?

```
fn add_albedo_factor_to_mesh(
    mesh_instances: &mut Vec<MeshInstance>,
    asset3d: &Asset3D,
) {
  for instance in mesh_instances {
      if let Some(mesh) = &mut instance.mesh {
          if let Some(mutable_mesh) = Arc::get_mut(mesh) {
              mutable_mesh.materials[0].albedo_factor = asset3d
                  .albedo_factor
                  .map_or(re_renderer::Rgba::WHITE, |c| c.0.into());
          } else {
              println!("Arc is none");
          }
      }
  }
}
```

Wumpf: oh you're right, totally missed that this is a change on the mesh rather the mesh instances. This makes things actually more complicated still. What you're doing right now creates a lot more new meshes potentially: in a gltf a single mesh may be instantiated many times, which is why its an arc in the first place. We really don't want to duplicate meshes since this may copy around a looot of memory.
... but the vec of mesh instances itself can still be edited in-place ;-)

I can take a look how to solve this best

Wumpf: it would be ofc a lot easier to just change the `additive_tint` property, but that's confusing because we're not even exposing this in the meshes archetype. (it's a remnant from back when we didn't have outlines and needed some way of highlighting things)

Wumpf: The reason you don't see any changes when changing the color oh the _optional_ cpu mesh is that it does not affect the gpumesh. Creating that is also quite expensive since it entails transfer of all the gpu data to cpu.
So what we really have to do here is to change the loaders to not do that step. Then it's much easier to do postprocessing of the resulting cpu-only meshes

EtaLoop:
Thanks a lot :+1:

crates/viewer/re_space_view_spatial/src/mesh_loader.rs
```diff
@@ -44,12 +37,13 @@ impl LoadedMesh {
     pub fn load_asset3d_parts(
         name: String,
         media_type: &MediaType,
-        asset3d: &Asset3D,
+        bytes: &[u8],
         render_ctx: &RenderContext,
+        albedo_factor: &Option<AlbedoFactor>,
```
EtaLoop, line 42: Should I pass the whole Asset3D or just bytes and albedo_factor?

Wumpf: doesn't matter much, but I reckon all callsites have `Asset3D` ready already? Then might as well pass the whole thing

Wumpf (approved):
alright, let's finally get this in! :)

EtaLoop:
Done, sorry for the mess :sweat:

EtaLoop:
Thank you for your time and advice

Issue #6898, opened 2024-07-15
Graph archetype


vmayoral:
Most of the computations in robotics are expressed in the form of computational graphs. Whether you use ROS, ROS 2 or alternative frameworks, this seems to be the general consensus.

![imagen](https://github.com/user-attachments/assets/07072741-a926-4172-ae58-7a7dc8864497)

(*an arbitrary ROS graph*)

Besides their use for dev-purposes, these graphs are a fundamental tool when visualizing a robot's architecture. Most ROS-related graph representations are lacking. There's an interesting opportunity to get this right within rerun. That would make it much more useful for roboticists.

nikolausWest:
Thanks @vmayoral! 

> There's an interesting opportunity to get this right within rerun. 

could you expand on this? Sounds very interesting!

vmayoral:
Hey @nikolausWest,

I'll try to expand on both the motivation and on the technical approach, sure:
- **Motivation-wise**: (As I believe you folks know already) Robots are networks of networks, with sensors capturing data, passing to compute technologies, and then on to actuators and back again in a deterministic manner. These networks can be understood as the nervous system of the robot, passing across compute Nodes, that represent neurons. Like the human nervous systems, real-time information across all these computational Nodes is fundamental for the robot to behave coherently. "Robot brains" are built with this same philosophy. Behaviors take the form of computational graphs, with data flowing between Nodes, across physical networks (communication buses) and while mapping to underlying sensors, compute technologies and actuators. Frameworks like ROS enables you to build these computational graphs and create robot behaviors by providing libraries, a communication infrastructure, drivers and tools to put it all together. Alternatives to ROS also bet into this graph-like concepts. In a way, *robot behaviors derive from the dataflow across these graphs, thereby getting a good representation of such (computational) graphs is fundamental for robot visualization*.
- **Technically** (getting this right[^1]): Let me start by telling you what works with *limitations* and what doesn't, and then throw a few suggestions.
  - Most of us still rely on [`rqt_graph`](http://wiki.ros.org/rqt_graph) for (robot) graph visualization. This tool hasn't evolved alongside modern software concepts in robotics (aspects such as ROS Components as opposed to Nodes (or Nodelets) aren't depicted). There're plenty of limitations on this regard. You still can't easily properly represent other-than-pub/sub relationships into these graphs, for example client-server interactions such as ROS `services`.
[...]

PR #7500, opened 2024-09-24
Implement graph components and archetypes

grtlr:
Tracking issue: #7897 

This implements basic graph primitives in the Rerun data model. This is a first step towards visualizing node-link diagrams in Rerun (related issue: #6898).

In addition to the changes to the data model, this PR adds two example crates, `node_link_graph` and `graph_view` to the Rust examples that show how these primitives can be used.

## Design Decisions

- Nodes and edges are stored as `components` and can be batched. To have a single node per entity we can use Rerun’s [[clamping mechanism](https://rerun.io/docs/concepts/batches#component-clamping)](https://rerun.io/docs/concepts/batches#component-clamping).
- `GraphNodeId` is modeled as ~`u32` to improve performance when using `petgraph`~ strings for better user experience.
- A node is unique identified by combining its `GraphNodeId` and its `EntityPath`.
- Labels of the nodes can be set via the `labels` component and toggled via `show_labels`
- ~Hierarchical graphs can be modeled through entity paths. For edges that cross entity boundaries we can insert dummy nodes to properly render subparts of the hierarchy.~
- ~Nodes and edges need to be logged to the same entity, otherwise the selections will collide. We provider helper functions / conversions to link nodes that are stored in different entities.~

> [!WARNING]
> This PR has changed considerably from its initial version:
> * No linking of nodes between entities (an therefore hierarchy) yet.
> * For now, Graphs are local only and therefore have to be logged to the same entity.

## Logging example

```rs
rec.log(
        "simple",
        &rerun::GraphNodes::new(["a", "b", "c"])
            .with_positions([(0.0, 100.0), (-100.0, 0.0), (100.0, 0.0)])
            .with_labels(["A", "B", "C"]),
    )?;
    // Note: We log to the same entity here.
    rec.log(
        "simple",
        &rerun::GraphEdges::new([("a", "b"), ("b", "c"), ("c", "a")]).with_directed_edges(),
    )?;
```

## TODOs

- [x]  ~Get rid of the `Default` derive for `GraphNodeId` and `GraphEdge` in the flatbuffer definitions.~
- [x]  Improve ergonomics for generating graph edges during logging.
- [x]  Ensure that logging works from Python and C++ too.
- [x]  Fix remaining lints.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7500?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7500?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7500)


nikolausWest:
In this design, are the node id's global?

crates/store/re_types/definitions/rerun/components/graph_edge.fbs
```diff
@@ -0,0 +1,19 @@
...
+// TODO(grtlr): Why do we need `Default` here?
+struct GraphEdge (
+  "attr.arrow.transparent",
+  "attr.python.aliases": "Tuple(int, int)",
+  "attr.rust.derive": "Copy, Default, PartialEq, Eq, PartialOrd, Ord, Hash, bytemuck::Pod, bytemuck::Zeroable",
+  "attr.rust.repr": "transparent",
+  "attr.rust.custom_clause":
+    'cfg_attr(feature = "serde", derive(::serde::Serialize, ::serde::Deserialize))'
+) {
+  edge: [rerun.datatypes.GraphNodeId: 2] (order: 100);
```
emilk, line 18: Is there any practical use (in robotics, primarily) in support hypergraphs (where an edge can join multiple nodes)?

grtlr:
@nikolausWest Good point! So far I've made the assumption that all node IDs are global and can be referenced across entities to allow edges that connect nodes from different entities (similar to `ClassId`).

We could probably use a _namespaced_ approach (e.g. based on entity path) by changing the way we gather the nodes from the entities that are currently being visualized. Then, we would have to store this entity information in the edges though. 

However, it's probably simpler for the user to encode this information in the node IDs themselves. In the case of `String` IDs, this could be done by appending the entity path as a suffix. If we stick with IDs based on integers users could for example use factorizations.

Do you have a particular use case for namespaced node IDs in mind? How would you expect Rerun to behave in that case?

crates/store/re_types/definitions/rerun/components/graph_edge.fbs
grtlr: Hmm 🤔, I'm not aware of any practical applications that use hypergraphs. There also does not seem to be any support in the [`petgraph` crate](https://docs.rs/crate/petgraph/latest), so maybe we should focus on regular graphs first. If we hypergraphs will become a thing, we can probably add a `HyperGraphEdge` component to remain backwards compatible.

Of course, an hypergraph could also be "normalized" by adding virtual nodes for the edges that join multiple nodes.

nikolausWest:
> So far I've made the assumption that all node IDs are global and can be referenced across entities to allow edges that connect nodes from different entities (similar to ClassId).

Class Ids aren't actually global in the sense that to resolve a class id into e.g. a color, we walk up the graph to the first Annotation Context and use that to look up the value. That means these are actually scoped.


> Do you have a particular use case for namespaced node IDs in mind? How would you expect Rerun to behave in that case?

I'm rather thinking about what happens if a user has multiple graphs. It would be a bit strange if edges started going between them because the user didn't correctly partition their node ids.

nikolausWest:
One option to consider here could be to have two types of edges. One that is within-entity edges (just a pair of node ids). The other could be between entity edges (destination entity, optional(source id), optional(destination id)). There might be several ways to model that but that would at least be the general idea

grtlr:
@nikolausWest Brief update:

Edges now have two additional attributes `source_entity` and `target_entity` as you described above. That way we can:
1. Make edges local to an entity by default to avoid collisions.
2. Allow linking between nodes of different entities.

The following shows a new toy example of the new logging API, which I also streamlined: https://github.com/rerun-io/rerun/pull/7500/files#diff-d054c306b388fcc1e8daf9f0477735519df3eeb486030979c62478b5d43404dcR36-R66.

I've also improved the debug graph viewer to show lists of nodes, edges, and their corresponding entity path. I'm currently working on getting overrides in the UI to work so that we can color nodes and edges to more easily understand the data model. This also helps me better understand the visualizer concepts.

I have one outstanding design decision:

The current systems allows for edges to live in completely unrelated parts of the entity hierarchy. This means that when the user choses to visualize a certain sub-entity not all edges are retrieved by default and the user has to manually specify the additional entities that contain the edges using _Entity Path Filters_:

```
+ /doors/**                <- contains the global edges
+ /hallway/areas/**        <- the actual entity that the user wants to visualize
```

Since this can be confusing due to the lack of discoverability, I think we should pull in global edges from outside the current hierarchy and visualizing them as dummy nodes. In the current design this forces us to iterate over all edges starting from the root. Should we mitigate this by introducing a new `GraphCrossEntityEdge` component, similar to what you described above?

nikolausWest:
@grtlr I actually think having edges on different entities than nodes is the main reason to go with your proposal. That allows users to put different meta data on edges and nodes which is a very important feature. If you can think of a way to still keep it more local that might be worth while.

> Since this can be confusing due to the lack of discoverability, I think we should pull in global edges from outside the current hierarchy and visualizing them as dummy nodes. In the current design this forces us to iterate over all edges starting from the root. 

I'm not completely sure about this. Maybe @Wumpf has some thoughts?

Wumpf:
Haven't been following the entire discussion, but there's a lot of issues with pulling data from outside of a viewer's query/entity-pathfilter:
* changes the query mechanism
     * we actually want to get to a place where we can predict the query of a view perfectly just from looking at how its configured, not knowing its specific type 
     * we're quite far from that and there's existing violations of that rule, 3d transforms being the most prominent one
* blueprint can't be applied to everything outside the path filter
   * blueprint overrides have no effect 
   * blueprint defaults on the View have no effect
* if we pull data outside of the path filter, how to stop certain data from being ingested?
    * control for that is blueprint
    * does this cause a huge query? 
    * how do I display independent graphs in independent views?

nikolausWest:
I think it's pretty clear we shouldn't include data from outside the entity path query. I think one question that leaves though is how to handle edges that are within the included entities but refer to nodes that are not. Perhaps some kind of greyed out nodes could make sense there (maybe even an option to include or exclude those)

grtlr:
Thank you @Wumpf for the clarification—that makes a lot of sense!

@nikolausWest I agree that we should show those as edges to some dummy nodes. In fact this is how I stumbled upon the above problem in the first place (undirected edges).

grtlr:
@emilk, @Wumpf: @nikolausWest asked me to post this here to open up the discussion:

I wanted to sync with you to map out the next steps of the graph implementation. It would be awesome if we could settle on a _definition-of-done (DoD)_ for the initial version. I'm happy to add features and improve my work going forward, but I think having a milestone to work against will help a lot.

There are several areas that I've worked on and that will require future work:

### Data model

Overall I think the data model is finished. Becaue of the use of flatbuffers we can easily add fields like edge and node weights in the future. There are currently some fields in the model like `position` that are not used, I would simply remove them before we merge the changes (depending on the DoD). If the current design is approved, I would start writing some helper functions for the Python and C++ logging libraries.

### Graph Viewer

The basic graph viewer is working correctly. It has the following features:
* Dragging of nodes
* Can show directed/undirected edges (for now only straight lines)
* Highlights nodes and selections
* Toggle between node labels on/off.
* Shows dummy nodes (nodes that are part of edges that live outside of the current selected entity hierarchy)

The code for the graph viewer still requires quite a bit of cleanup as I was still experimenting with the different layout algorithms, but I think I have a pretty good idea on what needs to be done—and I'm currently working on that.

### Graph layouts

This is the biggest outstanding area, as the existing crates in the Rust ecosystem are all lacking—I implemented layout providers for all apart from `graphviz-rust`:
* `layout-rs` (GraphViz-like layouts) does only expose the node positions and not the control points for the edges. We would have to fork this project.
* `graphviz-rust` requires the GraphViz CLI to be present, which adds complexity to the build for native releases, and would require us to mainain a separate version for the web viewer (https://hpcc-systems.github.io/hpcc-js-wasm/graphviz/src/graphviz/classes/Graphviz.html), so I think this is of the table.
* [`fdg` and `fdg-sim`](https://github.com/grantshandy/fdg) (force-based layouts): These are the most promising—but they still have some significant limitations for our use cases. The currently released version, `fdg-sim` is missing some flexibility (e.g. does not allow adding custom forces). There is an _unreleased_ version, `fdg`, that fixes some of these shortcomings. However, dynamically adding and removing edges will force us to modify some of that libraries internal state. It is also missing a _collision force_ as well as _positioning forces_ (which we need for properly visualizing disjoint graphs). Another risk is, that there are not indications that the unreleased version will be released in the next time, as development has slowed down.

Unfortunately, I think the features provided by the existing crates don't really suffice for our needs.

Ideally we would want to have a Rust version of: https://d3js.org/d3-force. Porting it over should be pretty straight forward, as it is not too much code. To efficiently find collisions, we would also need a quad tree. As per @Wumpf we don't have an implementation in Rerun, and I don't know how good the existing implementations on crates.io are. Do you have experience with any quad tree crates in Rust?

### Summary

Overall, I don't think it makes sense to merge the graph primitives without having at least one basic (but robust) layout algorithm implemented to visualize them—we just need to consider if it is worth adding this additional complexity to Rerun.

abey79:
```
table Graph {
	// required

    nodes: [rerun.components.Nodes]; // id: string

    // rejected: use integer unique id instead of string, perf should be ok if we intern string, would force a `Label` component (which sounds like a rerun-y thing, but would be painful on logging end with dynamic graphs), feedback from ROSBAG person is to use strings

    // rejected: referring to node by index, very painful on logging end when you have dynamic graphs, where node can appear/disappear

	// recommended

    directed_edges: [rerun.components.Edges]; // (id: string, id: string)
    undirected_edges: [rerun.components.Edges];

    // Alternative: 

    // rejected: entity reference

    // rejected: adjacency list: possible needed for big/dense graphs, but more complicated to log, can be added in the future

     // rejected: we must be able to know very easily if there are only directed edges (such that we know, e.g. a DAG-layout engine has a chance to run
     //edges: [rerun.components.Edges;  // (u32, u32, directed: bool)

     // optional

     node_colors: 

    
}

**Review by emilk — COMMENTED** [review]
Couldn't help myself from having a quick look - it looks promising!

File: crates/viewer/re_space_view_graph/src/graph/mod.rs [hunk 2]
```diff
@@ -0,0 +1,126 @@
...
+                    if seen.contains(&(&entity_path, &node.node_id)) {
+                        continue;
+                    }
+                    unknown.insert((entity_path, node.node_id));
+                }
+            }
+        }
+
+        if nodes.is_empty() && unknown.is_empty() {
+            return None;
```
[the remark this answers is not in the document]
grtlr: This was mainly for debugging and has been fixed.

crates/viewer/re_space_view_graph/src/graph/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+}
+
+pub(crate) struct Graph<'a> {
+    /// Contains all nodes that are part mentioned in the edges but not part of the `nodes` list
+    unknown: HashSet<(EntityPath, datatypes::GraphNodeId)>,
+    nodes: &'a Vec<NodeVisualizerData>,
+    directed: &'a Vec<EdgesDirectedData>,
+    undirected: &'a Vec<EdgesUndirectedData>,
```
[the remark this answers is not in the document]
grtlr: :thumbsup:

crates/viewer/re_space_view_graph/src/graph/mod.rs
```diff
@@ -0,0 +1,126 @@
+use std::collections::HashSet;
```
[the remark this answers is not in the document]
grtlr: Ah good to know, will do!

grtlr:
> Couldn't help myself from having a quick look - it looks promising!

Thanks for the comments on the PR! I'm currently incorporating your feedback and fixing some last things. If you want to try it out, you will need to active the Graph Space View experimental feature in the UI and then you can call cargo run -p node_link_graph -- --connect --example=lattice which is the most complete example for now.

grtlr:
@emilk I was not aware of `nohash-hasher` before—thank you for pointing it out!

I've implemented `nohash_hasher::IsEnabled` for `NodeIndex`:

https://github.com/grtlr/rerun/blob/111b0f58fb50b25d6c282f3453762724855fe795/crates/viewer/re_space_view_graph/src/graph/index.rs#L6-L20

We could also swich to `write_usize` for less risk for collisions, but I think this should be fine.

grtlr:
@abey79 @jleibs The only real deviation from our design that I had to do was storing the edges in a table rather than a struct. Unfortunately I couldn't find a way to get the following to work:

```fbs
namespace rerun.datatypes;

struct StringPair (
    "attr.rust.derive": "Default, PartialEq, Eq, PartialOrd, Ord",
    "attr.rust.custom_clause":
        'cfg_attr(feature = "serde", derive(::serde::Serialize, ::serde::Deserialize))'
) {
  /// The id of the source node.
  pair: [string: 2] (order: 100);
}
```

Which produces the following error:
```
error: structs may contain only scalar or struct fields
```

abey79:
> @abey79 @jleibs Just a heads up: The only real deviation from our design that I had to do was storing the edges in a table rather than a struct.

That's fine, my bad for seeding the wrong idea. This flatbuffer limitation actually has no impact whatsoever, since we are no _actually_ using any flatbuffer. We use the language and `flatc` because it's convenient, but everything else is custom. The codegen'd stuff has no difference between `struct` and `table` (for those cases where `flatc` actually accepts `struct`).

examples/rust/node_link_graph/src/examples/social/tij_SFHH.dat_
abey79, line 1: We try to avoid big files in repo. Most of our examples have a bit of logic to download them locally first if needed, and then load from them (admittedly, especially python examples). Ideally we should do the same.

Btw, we probably need a python example too.

ARCHITECTURE.md
```diff
@@ -135,6 +135,7 @@ Update instructions:
 | re_space_view               | Types & utilities for defining Space View classes and communicating with the Viewport.                     |
 | re_space_view_bar_chart     | A Space View that shows a single bar chart.                                                                |
 | re_space_view_dataframe     | A Space View that shows the data contained in entities in a table.                                         |
+| re_space_view_graph         | A Space View that shows a graph (node-link diagram).                                                       |
```
abey79, line 138: Note to self: update the crate diagram

crates/viewer/re_space_view_graph/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+re_space_view.workspace = true
+re_tracing.workspace = true
+re_types.workspace = true
+re_ui.workspace = true
+re_viewer_context.workspace = true
+re_viewport_blueprint.workspace = true
+
+egui.workspace = true
+
+#TODO(grtlr): check if all of these are still needed
```
abey79, line 37: FYI: cargo machete should do that at least on internal PR CI. (For this big PR, when we're close to done, I'll temporarily open a mirror PR so that we get the full CI.)

crates/viewer/re_space_view_graph/src/error.rs
```diff
@@ -0,0 +1,19 @@
...
+    EdgeUnknownNode,
+
+    #[error("missing layout information for node `{1}` in entity `{0}`")]
+    MissingLayoutInformation(EntityPath, datatypes::GraphNode),
+}
+
+impl From<Error> for SpaceViewSystemExecutionError {
+    fn from(val: Error) -> Self {
+        // TODO(grtlr): Ensure that this is the correct error type.
+        Self::DrawDataCreationError(Box::new(val))
```
abey79, line 17: I'm not sure it is the correct type. If i'm not mistaken that variant relates specifically to GPU draw data to be fed to re_renderer, which we're not using in this view.

We repeatedly run into this limitation of `SpaceViewExecutionError`. We probably should add a variants like `SystemError { identifier: String, error: Box<dyn Error> }`, and `ViewClassError { identifier: String, error: Box<dyn Error> }` so you can actually do this.

@Wumpf do you agree?

crates/viewer/re_space_view_graph/src/graph/hash.rs
```diff
@@ -0,0 +1,41 @@
...
+        self.0.hash64()
+    }
+}
+
+impl std::hash::Hash for GraphNodeHash {
+    #[inline]
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.0.hash(state);
+    }
+}
```
abey79, line 22: Can't these be derived?

crates/viewer/re_space_view_graph/src/graph/index.rs
```diff
@@ -0,0 +1,35 @@
...
+
+impl nohash_hasher::IsEnabled for NodeIndex {}
+
+impl std::hash::Hash for NodeIndex {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        // TODO(grtlr): Consider using `write_usize` here, to further decrease the risk of collision.
+        let combined = self.entity_hash.hash64() << 32 | self.node_hash.hash64();
+        state.write_u64(combined);
+    }
+}
```
abey79, line 20: Same, can't this be derived?

crates/viewer/re_viewer/src/app.rs
```diff
@@ -1816,13 +1816,16 @@ fn populate_space_view_class_registry_with_builtin(
...
     space_view_class_registry.add_class::<re_space_view_spatial::SpatialSpaceView2D>()?;
     space_view_class_registry.add_class::<re_space_view_spatial::SpatialSpaceView3D>()?;
     space_view_class_registry.add_class::<re_space_view_tensor::TensorSpaceView>()?;
     space_view_class_registry.add_class::<re_space_view_text_document::TextDocumentSpaceView>()?;
     space_view_class_registry.add_class::<re_space_view_text_log::TextSpaceView>()?;
     space_view_class_registry.add_class::<re_space_view_time_series::TimeSeriesSpaceView>()?;
-    space_view_class_registry.add_class::<re_space_view_dataframe::DataframeSpaceView>()?;
+
+    // The following are experimental views.
+    space_view_class_registry.add_class::<re_space_view_graph::GraphSpaceView>()?;
```
abey79, line 1828: It can be integrated in the normal list, and yes we like these sorted :)

examples/rust/node_link_graph/Cargo.toml
abey79, line 1: It would be very nice to have this example mirror as closely the python one as possible, and have the same name. And it would be ultra great to have the c++ version as well, so we cover all languages like this one: https://rerun.io/examples/feature-showcase/dna

Also, we're also likely include that example in the welcome screen. (I'll have to re-figure out how that's done.)

abey79 (commented):
Generally looking great. Here are a few comments after a cursory look. I'll dive deeper on the meaty part on Monday.

crates/viewer/re_space_view_graph/src/graph/index.rs
grtlr: The implementation here is non-standard, because `nohash_hasher::IsEnabled` requires a single cals to `write_{u64|usize|...}` as per the [crate documentation](https://crates.io/crates/nohash-hasher).

abey79: Ah ok. Please add a comment stating that then.

crates/store/re_types/definitions/rerun/archetypes/graph_nodes.fbs
```diff
@@ -0,0 +1,29 @@
...
+) {
+  // --- Required ---
+
+  /// A list of node IDs.
+  node_ids: [rerun.components.GraphNode] ("attr.rerun.component_required", order: 1000);
+
+  // --- Optional ---
+
+  /// Optional center positions of the nodes.
+  positions: [rerun.components.Position2D] ("attr.rerun.component_recommended", nullable, order: 3000);
```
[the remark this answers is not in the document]
grtlr: :thumbsup:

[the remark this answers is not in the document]
grtlr: I agree! We also could conceptualize this even a bit more. In my current implementation the user can drag individual nodes, and in the future the auto-layout should dynamically adapt to those changes. One idea would then be to set the position of that particular node as an override. However, I'm not sure if I can override a single value in a component batch?

crates/store/re_types/definitions/rerun/archetypes/graph_nodes.fbs
```diff
@@ -0,0 +1,29 @@
...
+
+  /// Optional colors for the boxes.
+  colors: [rerun.components.Color] ("attr.rerun.component_recommended", nullable, order: 3100);
```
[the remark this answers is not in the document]
grtlr: :thumbsup: Sorry for the copy and paste error.

crates/store/re_types/definitions/rerun/archetypes/graph_edges.fbs
```diff
@@ -0,0 +1,24 @@
...
+// ---
+
+/// A list of edges in a graph.
+///
+/// By default, edges are undirected.
+table GraphEdges (
+  "attr.rust.derive": "PartialEq, Eq",
+  "attr.docs.category": "Graph",
+  "attr.docs.view_types": "Graph View"
+) {
```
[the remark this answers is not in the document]
grtlr: :thumbsup:

crates/store/re_types/definitions/rerun/archetypes/graph_edges.fbs
```diff
@@ -0,0 +1,24 @@
...
+  // --- Required ---
+
+  /// A list of node IDs.
+  edges: [rerun.components.GraphEdge] ("attr.rerun.component_required", order: 1000);
+
+
+  // --- Recommended ---
+
+  /// Specifies if the graph is directed or undirected.
```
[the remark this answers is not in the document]
grtlr: :thumbsup:

crates/viewer/re_space_view_graph/src/error.rs
Wumpf: correct, `DrawData` refer's to re_renderer's `DrawData` here
Agreed, it's time to make the error type more scalable by adding view specific subtypes!

crates/viewer/re_space_view_graph/src/types.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+pub struct EdgeInstance<'a> {
+    pub source: components::GraphNode,
+    pub target: components::GraphNode,
+    pub entity_path: &'a re_log_types::EntityPath,
+    pub instance: Instance,
+    pub edge_type: components::GraphType,
+}
+
```
abey79, line 47: We have a real opportunity to get rid of all these lifetimes by using owned values instead:
- `EntityPath` is cheap to clone (inner is an `Arc` + pre-computed hash)
- `ArrowString` and `GraphNode` are arrow buffer, which are also `Arc`

abey79 (commented):
WIP review.

examples/rust/node_link_graph/Cargo.toml
grtlr: I've created a Rust version of the binary tree. Outstanding are:

* Python version of the lattice
* All C++ examples

crates/viewer/re_space_view_graph/src/error.rs
grtlr: I will create a separate PR for this!

grtlr: Corresponding PR: https://github.com/rerun-io/rerun/pull/7942

abey79 (commented):
WIP review

crates/viewer/re_space_view_graph/src/ui/scene.rs
```diff
@@ -0,0 +1,275 @@
...
+        // :warning: Currently, `children_drag_delta` and `children_hovered` only report events from the entity rectangles.
+        // TODO(grtlr): Would it makes sense to move the creation of the `Response` here and let `Canvas` only be a thin wrapper?
+        //              That way we would avoid the duplication between those objects and we would have single-responsibility.
+
+        world_to_view.translation += self.children_drag_delta;
+        if response.dragged() {
+            world_to_view.translation += response.drag_delta();
+        }
+
+        if let Some(pointer) = ui.ctx().input(|i| i.pointer.hover_pos()) {
```
grtlr, line 100: Allow zooming into the graph to avoid weird panning bug a zoom limit.

crates/viewer/re_space_view_graph/src/view.rs
```diff
@@ -0,0 +1,248 @@
...
+            bounds_property.component_or_fallback(ctx, self, state)?;
+
+        let layout_was_empty = state.layout_state.is_none();
+        let layout = state.layout_state.get(
+            query.timeline,
+            query.latest_at,
+            graphs.iter().map(|g| (g.0).clone()),
+            graphs.iter().map(|(_, graph)| graph),
+        );
+        let needs_remeasure = layout.has_zero_size();
```
grtlr, line 169: Does not account for labels that change size.

abey79 (commented):
WIP

docs/snippets/all/archetypes/graph_undirected.cpp
```diff
@@ -0,0 +1,27 @@
...
+    );
+
+    rerun::datatypes::Utf8Pair edge_1({"a"}, {"b"});
+    rerun::datatypes::Utf8Pair edge_2({"b"}, {"c"});
+    rerun::datatypes::Utf8Pair edge_3({"c"}, {"a"});
+
+    // Note: We log to the same entity here.
+    rec.log(
+        "simple",
+        rerun::GraphEdges({edge_1, edge_2, edge_3})
```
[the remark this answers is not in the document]
grtlr: Thank you @Wumpf :star_struck:

abey79 (approved):
Let's merge this! 🎉  (as soon as ci is ✅)

nikolausWest:
🥳

Issue #7375, opened 2024-09-09
Support h.264 codec in `re_video`


jprochazk:
https://github.com/rerun-io/rerun/blob/07291e52398a338414f5da7e95aeb2008e9006a3/crates/store/re_video/src/mp4.rs#L31-L42

PR #7511, opened 2024-09-25
Support h.264 video decoding on web

jprochazk:
### What

* Closes https://github.com/rerun-io/rerun/issues/7375

- The seeking logic now searches for the requested segment by the presentation/composition timestamp passed in by the user. This way if a user requests frame at 12s they will get the frame which will be _presented_ at 12s, not _decoded_ at 12s. See [I/P/B frame terminology](https://en.wikipedia.org/wiki/Video_compression_picture_types) in video encoding for more information as to why this distinction matters.
- `EncodedVideoChunk`s are now given the _composition timestamp_ of the sample instead of its _decode timestamp_, so the output `VideoFrame` objects are in _composition timestamp_ order.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7511?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7511?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7511)

### DIFF
--- a/crates/viewer/re_renderer/src/video/decoder/web.rs
+++ b/crates/viewer/re_renderer/src/video/decoder/web.rs
@@ -240,7 +240,7 @@ impl VideoDecoder {
 
         // 2. Search _backwards_, starting at `decode_sample_idx`, looking for
         //    the first sample where `sample.composition_timestamp <= presentation_timestamp`.
-        //    This is the sample the user requested.
+        //    This is the sample which when decoded will be presented at the timestamp the user requested.
         let Some(requested_sample_idx) = self.data.samples[..=decode_sample_idx]
             .iter()
             .rposition(|sample| sample.composition_timestamp <= presentation_timestamp)
--- a/crates/viewer/re_renderer/src/video/decoder/web.rs
+++ b/crates/viewer/re_renderer/src/video/decoder/web.rs
@@ -107,7 +107,7 @@ impl VideoDecoder {
         render_context: &RenderContext,
         data: Arc<re_video::VideoData>,
     ) -> Result<Self, DecodingError> {
-        re_log::debug!("{:?}", data.samples);
+        re_log::debug!("{:?}", data.segments);
 
         let frames = Arc::new(Mutex::new(Vec::with_capacity(16)));
         let decode_error = Arc::new(Mutex::new(None));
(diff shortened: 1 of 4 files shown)


crates/viewer/re_renderer/src/video/decoder/web.rs
```diff
@@ -200,47 +200,80 @@ impl VideoDecoder {
...
+
+    fn enqueue_requested_segments(
+        &mut self,
+        presentation_timestamp: Time,
+    ) -> Result<(), DecodingError> {
+        // Some terminology:
+        //   - presentation timestamp = composition timestamp
+        //     = the time at which the frame should be shown
+        //   - decode timestamp
+        //     = determines the decoding order of samples
```
Wumpf, line 224: good call to document that here, makes this stuff a lot more maintainable 👍

crates/viewer/re_renderer/src/video/decoder/web.rs
```diff
@@ -200,47 +200,80 @@ impl VideoDecoder {
...
+            &self.data.samples,
+            |sample| sample.decode_timestamp,
+            &presentation_timestamp,
+        ) else {
+            return Err(DecodingError::EmptyVideo);
+        };
+
+        // 2. Search _backwards_, starting at `decode_sample_idx`, looking for
+        //    the first sample where `sample.composition_timestamp <= presentation_timestamp`.
+        //    This is the sample the user requested.
```
Wumpf, line 243: nit: "uhm, actually" the user technically doesn't request a sample. A single sample is not able to represent the user's request

... but I don't know how to describe it better either right now. Fine to leave it this way I think

jprochazk:
Seems like the lint failure comes from main

Wumpf (approved):
really well document, also like it a lot how you split up the concerns of frame_at further and made everything more precise!

crates/viewer/re_renderer/src/video/decoder/web.rs
jprochazk: Changed it to "This is the sample which when decoded will be presented at the timestamp the user requested."

PR #7515, opened 2024-09-25
Dataframe view update and blueprint API (part 0): introduce view object and improve doc codegen

abey79:
### Checklist
* [ ] I have read and agree to [Contributor Guide](https://github.com/rerun-io/rerun/blob/main/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/rerun-io/rerun/blob/main/CODE_OF_CONDUCT.md)
* [ ] I've included a screenshot or gif (if applicable)
* [ ] I have tested the web demo (if applicable):
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/{{pr.number}}?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/{{pr.number}}?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)
* [ ] The PR title and labels are set such as to maximize their usefulness for the next release's CHANGELOG
* [ ] If applicable, add a new check to the [release checklist](https://github.com/rerun-io/rerun/blob/main/tests/python/release_checklist)!
* [ ] If have noted any breaking changes to the log API in `CHANGELOG.md` and the migration guide

- [PR Build Summary](https://build.rerun.io/pr/{{pr.number}})
- [Recent benchmark results](https://build.rerun.io/graphs/crates.html)
- [Wasm size tracking](https://build.rerun.io/graphs/sizes.html)

To run all checks from `main`, comment on the PR with `@rerun-bot full-check`.

Issue #7497, opened 2024-09-24
Dataframe: re-design UI based on new query model


abey79:
The goal of this redesign is to keep up with the changes being introduced in the data APIs #7455.

Overview: the is organised as a fixed pipeline of optional stages, where each stage corresponds to some primitive of the data API.

Stages:
- **View** (required): the subset of the datastore on which the rest of the filtering/querying happens.
  - `content`: entity path filter
  - `timeline`: the timeline used by everything else
- **Filter: time range** (optional): only keep rows whose timestamps on the view timeline are within the provided range.
  - `start`: low bound, can be set to -inf
  - `end`: high bound, can be set to +inf
- **Filter: events** (optional): only keep rows which contains a logging event for the specified column (formerly known as PoV).
  -  `entity`: the column entity (must be one of the view's entity)
  -  `component`: the column component (must be on of the specified entity)
- **Fill with latest-at** (bool, optional): whether or not null cells are filled using latest-at queries.
- **Select** (list of columns): for each view column, whether or not it is included or not in the final dataframe.
  - Helper to select all or nothing, or a recommended subset would be nice.

**Notes**:
- All components for the view entities are included in the view. It is not possible to further restrict this (like you can with the API).
- The dataframe view doesn't allow combing multiple `filter_events_for`. There is 0 or 1 active at a time.
- The dataframe view doesn't expose anything about custom time values (e.g. `filter_time`, `latest_at_times`).

Issue #7455, opened 2024-09-19
Expand python APIs to support new data API concepts


jleibs:
# Updated Proposal:
## Improved concept definitions
* Recording -- The logical representation of Rerun-sourced data. Physically backed by a chunk store.
* View -- A subset of a Recording, restricted to a specific timeline, set of entities, and/or archetypes/components.
  * Mentally aligned with Views as depicted in the viewer. The idea of "creating a table view from a space view" should not be a stretch.
  * The definition of a view *fully defines the schema and columns* contained in the dataset.
  * Prior to filtering, a View contains 1 row for every log call that produced data that is included in the view. (or multiple rows for send_columns calls).
* Filter -- A mechanism  of restricting the rows that are contained in a view. 
  * Filters never do data transformation. The schema is unchanged. Each individual row is included or excluded in its entirety.
  *  `.filter(TimeRange(start=..., end=...))` or maybe `.filter_range(start=..., end=...))`
* Re-indexing -- A mechanism of defining the index values for an alternative set of rows
  * `view.using_index_values(self, values: ArrayLike)`
* LatestAt -- A mechanism of populating missing data from a given row
* Select -- A mechanism of choosing a specific subset of columns from the query result
  * Does not change how many rows are returned, but lets the user specify the exact subset (and order) of the columns they are interested in retrieving.
  * `.select(Timeline(), "Translation3D")`

## Python APIs
```python

# Raises if number of recordings != 1
recording : Recording  = rr.data.load_recording("foo.rrd")

archive = rr.data.load_archive("foo.rrd")
recording = archive.recordings[0]

contents = {
  "world/points/**": [rr.Image],
  "metrics/**": [rr.Scalar]
}

contents = "world/**"

class Recording:
    def view(self, index: str, contents: ContentLike) -> RecordingView: ...

class RecordingView:
    def filter_range(self, min: float, max: float) -> RecordingView: ...
    def filter_times(self, times: ArrayLike) -> RecordingView: ...
    def filter_events_for(self, AnyColumn(...)) -> RecordingView: ...

    def using_index_values(self, times: ArrayLike) -> RecordingView: ...

    def latest_at(self) -> RecordingView: ...

    def select(self, *columns: AnyColumn) -> Iterator[pa.RecordBatch]: ...

    # Do we need this?
    def select_all(): ... 


# Alternative
expr = QueryExpression(...)

recording = load_recording(...)

recording.query(expr).select(...)
```

# Original Proposal (archive): 
<details>

[...]

Issue #6896, opened 2024-07-15
Add support for the dataframe view in the python blueprint API


abey79:
With the dataframe improvement, it's becoming more and more having explicit support for it in the blueprint API. I assume it should be relatively easy, but we somehow need to address the fact that the view itself is behind an opt-in feature flag.

abey79:
Doc generation: the dataframe view needs some special casing:
- All archetype should list it as supported by default
- It should list "All archetypes" as "Supported Archetype"

PR #7516, opened 2024-09-25
Dataframe view update and blueprint API (part 1): introduce fbs for the new query property

abey79:
### What

This PR introduces a new `DataframeQueryV2` view property archetype which models the query according to the new dataframe API design (https://github.com/rerun-io/rerun/issues/7455) and the feature we actually want to support in the dataframe view (https://github.com/rerun-io/rerun/issues/7497).

At this point, the new archetype is **NOT** used yet. It just lives alongside the previous iteration, which is still used by the actual view. The swap will occur later.

<hr>

Part of a series to address #6896 and #7498.

All PRs:
- #7515
- #7516
- #7527 
- #7545
- #7551
- #7572
- #7573

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7516?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7516?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7516)


crates/store/re_types/definitions/rerun/blueprint/archetypes/dataframe_query.fbs
```diff
@@ -49,3 +50,38 @@ table DataframeQuery (
...
+    /// The timeline for this query.
+    ///
+    /// If unset, the time panel's timeline is used and stored.
+    timeline: rerun.blueprint.components.TimelineName ("attr.rerun.component_optional", nullable, order: 100);
+
+    /// If set, a range filter is applied.
+    ///
+    /// Note: will be unset as soon as `timeline` is changed.
+    //TODO(ab): should this use the list-with-timelines trick?
+    range_filter: rerun.blueprint.components.RangeFilter ("attr.rerun.component_optional", nullable, order: 200);
```
abey79, line 70: TODO: change that to `filter_by_range`

crates/store/re_types/definitions/rerun/blueprint/archetypes/dataframe_query.fbs
```diff
@@ -49,3 +50,38 @@ table DataframeQuery (
...
+
+    /// Whether the filter by event feature is active.
+    // Note: we use an explicit field here such that the filter column is not lost when the filter is deactivated.
+    filter_by_event_active: rerun.blueprint.components.FilterByEventActive ("attr.rerun.component_optional", nullable, order: 300);
+
+    /// The column used when the filter by event feature is used.
+    ///
+    /// Note: only valid if the entity/component exists on `timeline`.
+    filter_by_event_column: rerun.blueprint.components.ComponentColumnSelector ("attr.rerun.component_optional", nullable, order: 400);
```
abey79, line 79: Note: in the blueprint API, the custom `__init__` will have a single `filter_by_event="/entity/path:ComponentName"` which will set both of these fields.

crates/store/re_types/definitions/rerun/blueprint/archetypes/dataframe_query.fbs
abey79: @jleibs the naming of these fields is probably something we should sync over—it would be very weird to have something different here and in the data API.

crates/store/re_types/definitions/rerun/blueprint/archetypes/dataframe_query.fbs
```diff
@@ -49,3 +50,38 @@ table DataframeQuery (
...
+
+    /// Should empty cells be filled with latest-at queries?
+    apply_latest_at: rerun.blueprint.components.ApplyLatestAt ("attr.rerun.component_optional", nullable, order: 500);
+
+    /// Selected columns. If unset, all columns are selected.
+    selected_columns: rerun.blueprint.components.SelectedColumns ("attr.rerun.component_optional", nullable, order: 600);
```
abey79, line 85: Unless we change the `select` terminology in the data API, I strongly suggest to keep the `selected` part here as well as in the code (next PR, mainly). That is, even if we decide to opt for "visibility" on the UI side (which is still up for debate).

crates/store/re_types/definitions/rerun/blueprint/archetypes/dataframe_query.fbs
```diff
@@ -49,3 +50,38 @@ table DataframeQuery (
...
+
+/// The query for the dataframe view.
+table DataframeQueryV2 (
+    "attr.rerun.scope": "blueprint"
+) {
+    // --- Optional ---
+
+    /// The timeline for this query.
+    ///
+    /// If unset, the time panel's timeline is used and stored.
```
Wumpf, line 63:
```suggestion
    /// If unset, the timeline currently active on the time panel's is used.
```

not sure though what the `and stored` part means here - do mean this is immediately stateful, i.e. the moment the view is spawned this latches? That would be unfortunate imho!

crates/store/re_types/definitions/rerun/blueprint/archetypes/dataframe_query.fbs
```diff
@@ -49,3 +50,38 @@ table DataframeQuery (
...
+    timeline: rerun.blueprint.components.TimelineName ("attr.rerun.component_optional", nullable, order: 100);
+
+    /// If set, a range filter is applied.
+    ///
+    /// Note: will be unset as soon as `timeline` is changed.
+    //TODO(ab): should this use the list-with-timelines trick?
+    range_filter: rerun.blueprint.components.RangeFilter ("attr.rerun.component_optional", nullable, order: 200);
+
+    /// Whether the filter by event feature is active.
```
Wumpf, line 72: should probably expand here what the feature does

crates/store/re_types/definitions/rerun/blueprint/components/range_filter.fbs
```diff
@@ -0,0 +1,13 @@
+
+namespace rerun.blueprint.components;
+
+/// Describe a component column to be selected in the dataframe view.
```
Wumpf, line 4:
```suggestion
/// Range filter for a dataframe query.
```

Wumpf:
@jleibs should probably compare notes with the dataframe api

crates/store/re_types/definitions/rerun/blueprint/archetypes/dataframe_query.fbs
abey79: Mmm, this indirectly points at a shortcoming of our doc codgen:

image

The original fields of the query are shown, but that is _not_ what the (custom) constructor accepts (in this case, a single `filter_by_event` argument which, when provided, sets active to `True` and column to the proper (parsed) value.

Fixing codegen for that seems... ambitious (need to parse the custom init args, etc.). Looks like I need to roll a datatype that contains the bool and the column.

(Note that I really want a bool + column, and just not an optional column. That way, the user can toggle the check box to see the effect without its filter column being "forgotten" each time.)

crates/store/re_types/definitions/rerun/blueprint/views/dataframe.fbs
```diff
@@ -8,6 +8,5 @@ table DataframeView (
     "attr.rerun.view_identifier": "Dataframe"
 ) {
     /// Query of the dataframe.
-    //TODO(#6896): this will be enabled in a later PR in the chain
-    //query: rerun.blueprint.archetypes.DataframeQueryV2 (order: 1000);
+    query: rerun.blueprint.archetypes.DataframeQueryV2 (order: 1000);
```
abey79, line 11: TODO: fix the view top-level docstring as per https://github.com/rerun-io/rerun/pull/7515#discussion_r1782487405

crates/store/re_types/definitions/rerun/blueprint/archetypes/dataframe_query.fbs
abey79: Fixed.

Issue #7404, opened 2024-09-11
static=True image logging still stores all the data


nisseknudsen:
**Describe the bug**
I am using `rr.log(..., static=True)` with images and the expectation that the memory consumption should stay stable over time. I only want to see the latest logged image.
I also have the expectation that when opening the (web) viewer, I do not have to watch through all of the history to the current frame.
Both is not met for some reason.

**To Reproduce**
Script to repro (uses webcam `0` on macOS, but should work with any computer with we webcam. Maybe try `/dev/video0` for ubuntu).
[scratch_5.py.zip](https://github.com/user-attachments/files/16968845/scratch_5.py.zip)

**Expected behavior**
Sidebar data source size stays stable and whatever a single JPEG image takes, let's say 20mbyte to be generous.
Unfortunately, it doesn't

<img width="349" alt="image" src="https://github.com/user-attachments/assets/af24a8a9-0834-4b88-b333-6a6fb419280b">

**Desktop (please complete the following information):**
 - OS: macOS Sonoma 14.6.1

**Rerun version**
Tested on 18.0 and 18.2 with same unexpected outcomes.

**Additional context**
Discord post: https://discord.com/channels/1062300748202921994/1075873257124810852/1283124563902992537

teh-cmc:
There are two issues at play:
1. Static chunks are not subject to the dangling detection/cleanup logic of the garbage collector, since static data as a whole is immune to garbage collection.
2. The decoded image cache doesn't subscribe to store events, and therefore continues caching stuff that has long been overwritten.

The fix for 1) is to properly account for static overwrites on the write path: patch all dangling references, decreases stats, fire events.
That should be a small amount of (likely tricky) code.

The fix for 2) is to set up a store subscriber so that the image cache drops overwritten data immediately.
That can range from trivial to extremely painful depending on how things are setup at the moment; I need to have a closer look.

nisseknudsen:
Thank you for the fix! Can confirm it works now 🙏

PR #7518, opened 2024-09-26
Automatic removal of unreachable static chunks

teh-cmc:
Unreachable static chunks are dead weight: there exists no query that can access their data (at least when using Rerun as a Visualizer).

By automatically removing dangling chunks, we make it possible for user to use the Rerun Viewer as a soft-realtime telemetry system (provided we properly invalidate our caches too, which is the subject of an upcoming PR).

This raises the question of what should happen when using Rerun as a database: should this data be kept and made accessible?
If so, this behavior should probably be made configurable (e.g. when instantiating a ChunkStore in the SDK).

* Part of #7404 

---

Test:
```python
from pathlib import Path

import rerun as rr

image_file_path = Path(__file__).parent / "ferris.png"

rr.init("rerun_example_encoded_image", spawn=True)

for _ in range(0, 10):
    rr.log("image", rr.EncodedImage(path=image_file_path), static=True)
```

Command:
```
RERUN_FLUSH_NUM_ROWS=0 python test.py
```

Before:
image

After:
image

### Checklist

### DIFF
--- a/crates/store/re_chunk_store/src/writes.rs
+++ b/crates/store/re_chunk_store/src/writes.rs
@@ -57,6 +57,8 @@ impl ChunkStore {
 
             let row_id_range_per_component = chunk.row_id_range_per_component();
 
+            let mut overwritten_chunk_ids = HashMap::default();
+
             for (&component_name, list_array) in chunk.components() {
                 let is_empty = list_array
                     .validity()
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
so but what if I have a static image+mediatype and then were to override only the image? Then I'd always keep the first `image+mediatype` chunk around?

teh-cmc:
> so but what if I have a static image+mediatype and then were to override only the image? Then I'd always keep the first `image+mediatype` chunk around?

This is the subject of the next PR, but spoiler: no, if you overwrite the buffer, you'll drop every different media-type instantiations of that image..!

PR #7521, opened 2024-09-26
Make video decoding errors recoverable

### LINKED ISSUE #7491: Video decoder with multiple streams gets into error state after some scrubbing
Reproed while reviewing https://github.com/rerun-io/rerun/pull/7490, mp4 update and way the new preview plays might be relevant.

Used video: https://drive.google.com/file/d/19f_xKtM0ZcrLgSMfOAx2wCFJkEbB-g_m/view?usp=sharing

Did a bunch of seeks while the video is selected so that the preview keeps playing. Eventually both decoders get into an unrecoverable state:
image

Some more relevant browser log - looks like it first fails to decode but then can't recover:
```
[re_renderer::allocator::gpu_readback_belt] re_renderer/src/allocator/gpu_readback_belt.rs:278: Allocating new GpuReadbackBelt chunk of size 0.1 MiB [RemoteViewer.iXPtKCIw.js line 1 > Function:3244:17](https://rerun.io/_app/immutable/chunks/RemoteViewer.iXPtKCIw.js%20line%201%20%3E%20Function)
ERROR: [re_renderer::video::decoder::web] re_renderer/src/video/decoder/web.rs:421: failed to decode video: EncodingError: The given encoding is not supported. [RemoteViewer.iXPtKCIw.js line 1 > Function:3250:17](https://rerun.io/_app/immutable/chunks/RemoteViewer.iXPtKCIw.js%20line%201%20%3E%20Function)
ERROR: [re_renderer::video::decoder::web] re_renderer/src/video/decoder/web.rs:341: Failed to decode video chunk: InvalidStateError: VideoDecoder.decode: Decoder must be configured first [RemoteViewer.iXPtKCIw.js line 1 > Function:3250:17](https://rerun.io/_app/immutable/chunks/RemoteViewer.iXPtKCIw.js%20line%201%20%3E%20Function)
ERROR: [re_renderer::video::decoder::web] re_renderer/src/video/decoder/web.rs:421: failed to decode video: EncodingError: The given encoding is not supported. [RemoteViewer.iXPtKCIw.js line 1 > Function:3250:17](https://rerun.io/_app/immutable/chunks/RemoteViewer.iXPtKCIw.js%20line%201%20%3E%20Function)
[re_renderer::video::decoder::web] re_renderer/src/video/decoder/web.rs:84: Dropping VideoDecoder [RemoteViewer.iXPtKCIw.js line 1 > Function:3244:17](https://rerun.io/_app/immutable/chunks/RemoteViewer.iXPtKCIw.js%20line%201%20%3E%20Function)
[re_renderer::video::decoder::web] re_renderer/src/video/decoder/web.rs:86: Error when closing video decoder: InvalidStateError: An attempt was made to use an object that is not, or is no longer, usable [RemoteViewer.iXPtKCIw.js line 1 > Function:3250:17](https://rerun.io/_app/immutable/chunks/RemoteViewer.iXPtKCIw.js%20line%201%20%3E%20Function)
[re_renderer::video::decoder::web] re_renderer/src/video/decoder/web.rs:84: Dropping VideoDecoder [RemoteViewer.iXPtKCIw.js line 1 > Function:3244:17](https://rerun.io/_app/immutable/chunks/RemoteViewer.iXPtKCIw.js%20line%201%20%3E%20Function)
[re_renderer::video::decoder::web] re_renderer/src/video/decoder/web.rs:86: Error when closing video decoder: InvalidStateError: An attempt was made to use an object that is not, or is no longer, usable
```

Wumpf:
### What

* Fixes #7491

Unlike the issue title implies, the test video here always fails decoding at some point for Firefox & Chrome, it's not entirely deterministic though when & how often.

https://github.com/user-attachments/assets/8b04af17-66d2-47b0-8314-da9f3d5ee3f3

Latest version tested on Firefox & Chrome on Mac. Previously tested a similar changeset on Firefox & Chrome on Windows.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7521?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7521?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7521)

### DIFF
--- a/crates/viewer/re_renderer/src/video/decoder/web.rs
+++ b/crates/viewer/re_renderer/src/video/decoder/web.rs
@@ -13,7 +13,7 @@ use web_time::Instant;
 use super::latest_at_idx;
 use crate::{
     resource_managers::GpuTexture2D,
-    video::{DecodedFrame, DecodingError, FrameDecodingResult},
+    video::{DecodingError, FrameDecodingResult, VideoFrameTexture},
     DebugLabel, RenderContext,
 };
 
@@ -163,10 +163,10 @@ impl VideoDecoder {
     ) -> FrameDecodingResult {
         let result = self.frame_at_internal(presentation_timestamp_s);
         match &result {
-            Ok(DecodedFrame::Ready(_)) => {
+            Ok(VideoFrameTexture::Ready(_)) => {
                 self.error_on_last_frame_at = false;
             }
-            Ok(DecodedFrame::Pending(_)) => {
+            Ok(VideoFrameTexture::Pending(_)) => {
                 if self.error_on_last_frame_at {
                     // If we switched from error to pending, clear the texture.
                     // This is important to avoid flickering, in particular when switching from
@@ -311,7 +311,7 @@ impl VideoDecoder {
             // No buffered frames - texture will be blank.
 
             // Might this be due to an error?
-            ///
+            //
             // We only care about decoding errors when we don't find the requested frame,
             // since we want to keep playing the video fine even if parts of it are broken.
             // That said, practically we reset the decoder and thus all frames upon error,
@@ -330,7 +330,7 @@ impl VideoDecoder {
             // Don't return a zeroed texture, because we may just be behind on decoding
             // and showing an old frame is better than showing a blank frame,
             // because it causes "black flashes" to appear
-            return Ok(DecodedFrame::Pending(self.texture.clone()));
+            return Ok(VideoFrameTexture::Pending(self.texture.clone()));
         };
 
         // drain up-to (but not including) the frame idx, clearing out any frames
@@ -348,7 +348,7 @@ impl VideoDecoder {
         // We don't want to show this frame to the user, because it's not actually the one they requested,
         // so instead return the last decoded frame.
         if presentation_timestamp.into_millis(timescale) - frame_timestamp_ms > frame_duration_ms {
-            return Ok(DecodedFrame::Pending(self.texture.clone()));
+            return Ok(VideoFrameTexture::Pending(self.texture.clone()));
         }
 
         if self.last_used_frame_timestamp != frame.composition_timestamp {
@@ -356,7 +356,7 @@ impl VideoDecoder {
             copy_video_frame_to_texture(&self.queue, &frame.inner, &self.texture.texture);
         }
 
-        Ok(DecodedFrame::Ready(self.texture.clone()))
+        Ok(VideoFrameTexture::Ready(self.texture.clone()))
     }
 
     /// Clears the texture that is shown on pending to black.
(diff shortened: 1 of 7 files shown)


crates/viewer/re_renderer/src/video/decoder/web.rs
```diff
@@ -300,20 +295,37 @@ impl VideoDecoder {
...
         let Some(frame_idx) = latest_at_idx(
             &frames,
             |frame| frame.composition_timestamp,
             &presentation_timestamp,
         ) else {
-            // no buffered frames - texture will be blank
+            // No buffered frames - texture will be blank.
+
+            // Might this be due to an error?
+            if let Some(last_decoding_error) = decoder_output.last_decoding_error.clone() {
```
jprochazk, line 314: question: Doing it this way makes sense to me, but I wonder if we should unconditionally check for errors, instead of only in case we don't have any buffered frames? Would that make a difference in any case?

crates/viewer/re_renderer/src/video/mod.rs
```diff
@@ -47,17 +47,15 @@ pub enum DecodingError {
     NegativeTimestamp,
 }
 
+pub type FrameDecodingResult = Result<DecodedFrame, DecodingError>;
+
 /// Information about the status of a frame decoding.
-pub enum FrameDecodingResult {
+pub enum DecodedFrame {
```
jprochazk, line 53: nit: The frame may not be decoded yet, so I'm not sure about the naming here. Maybe `DecoderOutput`? We could also just use `Result<Poll<GpuTexture2D>, DecodingError>`

jprochazk (approved):
Cool!

crates/viewer/re_renderer/src/video/mod.rs
Wumpf: `Poll<GpuTexture2D>` would mean that the other side has to store the last known good texture and deal with all of that, also incompatible with the idea of small previews. But I like the naming suggestion!

crates/viewer/re_renderer/src/video/decoder/web.rs
Wumpf: Given that we throw away everything when reseting because of error it really shouldn't make a difference - if we fully recoverd by the time we enter here (impossible since we haven't given control back to the browser) then we wouldn't care about the error no matter how much time passed.
It's more that my internal logic went "only look for errors if we're out of frames" 🤔

crates/viewer/re_renderer/src/video/mod.rs
Wumpf: oh I already used up `DecoderOutput` now as an internal structure :thinking:
Hmm I can at least offer a counter argument that even on `Pending` we still tend to return a decoded frame...

PR #7525, opened 2024-09-26
Invalidate hub-wide caches on deletions and overwrites

### LINKED ISSUE #7404: static=True image logging still stores all the data

teh-cmc:
Hub-wide caches now subscribe to store events and invalidate accordingly in the face of deletions and overwrites.

This is a crutch to compensate for the lack of secondary caching, but a much needed crutch: the Rerun Viewer can now effectively be used as a soft realtime telemetry system.

https://github.com/user-attachments/assets/f1136f9d-e1fd-4e6b-87c6-422d5f3345e8

* Fixes #7404 

---

### `EncodedImage`

```python
for _ in range(0, 100):
    rr.log("image", rr.EncodedImage(path=image_file_path), static=True)
    time.sleep(0.01) # give time for the viewer to query and cache it
```

Before: :red_square: 
After: :green_circle: 

### `Mesh3D`

```python
for _ in range(0, 100):
    rr.log(
        "triangle",
        rr.Mesh3D(
            vertex_positions=np.tile(np.array([[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]]), (33333, 1)),
            vertex_normals=[0.0, 0.0, 1.0],
            vertex_colors=[255, 0, 0],
        ),
        static=True,
    )
    time.sleep(0.01) # give time for the viewer to query and cache it
```

Before: :red_square: 
After: :green_circle:  

### `Asset3D`

```python
for _ in range(0, 100):
    rr.log("world/asset", rr.Asset3D(path=sys.argv[1]), static=True)
    time.sleep(0.01) # give time for the viewer to query and cache it
```

Before: :red_square: 
After: :green_circle: 

### `TensorData`

```python
for _ in range(0, 1000):
    rr.log("tensor", rr.Tensor(tensor, dim_names=("width", "height", "channel", "batch")), static=True)
    time.sleep(0.01) # give time for the viewer to query and cache it
```

Before: :red_square: 
After: :green_circle: 

### `AssetVideo`

```python
frame_timestamps_ns = video_asset.read_frame_timestamps_ns()
rr.send_columns(
    "video",
    times=[rr.TimeNanosColumn("video_time", frame_timestamps_ns)],
    components=[rr.VideoFrameReference.indicator(), rr.components.VideoTimestamp.nanoseconds(frame_timestamps_ns)],
)

for _ in range(0, 100):
    rr.log("video", video_asset, static=True)
    time.sleep(0.01) # give time for the viewer to query and cache it
```

Before: :red_square: 
After: :green_circle:

---

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7525?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7525?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7525)

### DIFF
--- a/examples/python/face_tracking/face_tracking.py
+++ b/examples/python/face_tracking/face_tracking.py
@@ -20,6 +20,10 @@ import rerun.blueprint as rrb
 import tqdm
 from mediapipe.tasks.python import vision
 
+# If set, log everything as static
+global ALL_STATIC
+ALL_STATIC: bool = False
+
 EXAMPLE_DIR: Final = Path(os.path.dirname(__file__))
 DATASET_DIR: Final = EXAMPLE_DIR / "dataset"
 MODEL_DIR: Final = EXAMPLE_DIR / "model"
@@ -118,7 +122,8 @@ class FaceDetectorLogger:
         rr.log(
             "video/detector",
             rr.ClassDescription(
-                info=rr.AnnotationInfo(id=0), keypoint_connections=[(0, 1), (1, 2), (2, 0), (2, 3), (0, 4), (1, 5)]
+                info=rr.AnnotationInfo(id=0),
+                keypoint_connections=[(0, 1), (1, 2), (2, 0), (2, 3), (0, 4), (1, 5)],
             ),
             static=True,
         )
@@ -132,7 +137,7 @@ class FaceDetectorLogger:
             if self._video_mode
             else self._detector.detect(image)
         )
-        rr.log("video/detector/faces", rr.Clear(recursive=True))
+        rr.log("video/detector/faces", rr.Clear(recursive=True), static=ALL_STATIC)
         for i, detection in enumerate(detection_result.detections):
             # log bounding box
             bbox = detection.bounding_box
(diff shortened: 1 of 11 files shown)


examples/python/face_tracking/face_tracking.py
```diff
@@ -20,6 +20,10 @@
 import tqdm
 from mediapipe.tasks.python import vision
 
+# If set, log everything as static
```
Wumpf, line 23:
```suggestion
# If set, log everything as static.
#
# Generally, the Viewer accumulates data until its set memory budget at which point it will
# remove the oldest data from the recording (see https://rerun.io/docs/howto/limit-ram)
# By instead logging data as static, no data will be accumulated over time since previous
# data is overwritten.
# Naturally, the drawback of this is that there's no history of previous data sent to the viewer,
# as well as no timestamps, making the Viewer's  timeline effectively inactive.
```

Wumpf (approved):
👍 nice

looking forward to making the caches more of a store aware _thing_ so we can streamline some of the quasi-copy-pasted code here, but this will do for now; at least it's flexible enough (via being very stupid/simple) to deal with fallbacks & blueprint shenanigans.
(that's pretty much what you mean when complaining about the lack of secondary caches, right?)

teh-cmc:
> that's pretty much what you mean when complaining about the lack of secondary caches, right?

A real secondary cache would be aware of query semantics, and any shadowing introduced by these semantics, which is way more powerful and capable than working with row-ids directly.

E.g.:
* it would know to evict temporal data that is shadowed by newly logged static data. 
* it would know to evict data that is using latest-at semantics if the same index was relogged to.
* it would know to evict data if its index is shadowed by a (possibly recursive) clear.
* etc.

None of those are possible when working with row-ids directly.

examples/python/face_tracking/face_tracking.py
teh-cmc: nice

Wumpf:
`query awareness` is the keyword I was missing for my mental model, makes sense, thanks. Given what visualizer queries look like right now it sound quite challenging.
I figure a good first/next step would be to have more unified mechanism to store all data sources of a given query-entry (i.e. a list of rowids potentially from several stores (rowids are unique so the several store bit doesn't technically matter)) 🤔. Unfortunately that's not really enough since fallbacks may be generated using heuristic output that may change every frame. In that sense we can't make these caches "pure" secondary caches.

PR #7528, opened 2024-09-26
Don't show video load error if a blob has unrecognized or non-video media type

Wumpf:
### What

Before:
image

After:
image

... luckily I tried an obj file whose media type can't be recognized from bytes (because obj is a wild text format), which made me run into the corner case of video module not recognizing the format _and_ us not recognizing the media type from the blob.

Counter example, trying to load a wmv:
image
(btw. I did so by renaming the file to mp4 and dragging it in, otherwise we reject it earlier. Same could be achieved with direct log calls ofc)

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7528?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7528?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7528)

### DIFF
--- a/crates/viewer/re_data_ui/src/blob.rs
+++ b/crates/viewer/re_data_ui/src/blob.rs
@@ -123,7 +123,6 @@ pub fn blob_preview_and_save_ui(
             ui_layout,
             &video_result,
             video_timestamp,
-            media_type,
         );
     }
 
@@ -171,8 +170,8 @@ fn show_video_blob_info(
     ui_layout: UiLayout,
     video_result: &Result<re_renderer::video::Video, VideoLoadError>,
     video_timestamp: Option<VideoTimestamp>,
-    media_type: Option<&MediaType>,
 ) {
+    #[allow(clippy::match_same_arms)]
     match video_result {
         Ok(video) => {
             if ui_layout.is_single_line() {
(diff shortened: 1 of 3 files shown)


crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -264,7 +267,19 @@ fn show_video_blob_info(
                 }
             });
         }
+        Err(VideoLoadError::MediaTypeIsNotAVideo(_)) => {
+            // Don't show an error if this wasn't a video in the first place.
+            // Unfortunately we can't easily detect here if the Blob was _supposed_ to be a video, for that we'd need tagged components!
+            // (User may have confidently logged a non-video format as Video, we should tell them that!)
+        }
         Err(err) => {
+            if media_type.is_none() && matches!(err, VideoLoadError::UnrecognizedVideoFormat) {
```
jleibs, line 276: My only thought here is given that the only reason we are plumbing through media_type is to check this error condition, I wonder if this should just be bundled with the error-type?
```
    #[error("Video file has unsupported format")]
    UnrecognizedVideoFormat(Option<String>),
```

jleibs (approved):

crates/viewer/re_data_ui/src/blob.rs
Wumpf: good idea

PR #7531, opened 2024-09-27
Remove old `MeshManager`

### LINKED ISSUE #6755: GPU memory not freed when closing recording, but keeping viewer open
**Describe the bug**
It seems like GPU memory is not freed when closing the recording for meshes (only noticed / tested for meshes with large number of vertices, not sure how generally this applies).

As a workaround I have to close the viewer occasionally.

**To Reproduce**
Run the following code a few times while monitoring GPU memory usage. Memory keeps growing and closing the recordings does not free it.
```python
import numpy as np
import rerun as rr

rr.init("Large Mesh", spawn=True)
vertices = np.random.rand(3 * 3000000, 3)
rr.log("mesh", rr.Mesh3D(vertex_positions=vertices))
```

**Expected behavior**
Memory should be freed when closing a recording in the viewer.

**Screenshots**

https://github.com/rerun-io/rerun/assets/9785832/0ccee3a2-df56-4cf5-bee4-7f01909adee3

memory
Note that "Counted GPU" only goes up, never down.

**Backtrace**

 - OS: tested on Ubuntu 22.04 and Arch Linux

**Rerun version**
```
rerun-cli 0.17.0-alpha.8 [rustc 1.76.0 (07dca489a 2024-02-04), LLVM 17.0.6] x86_64-unknown-linux-gnu 6649/merge a6a666f, built 2024-06-25T19:47:07Z
```
Also tested with `0.16.1`.

Wumpf:
### What

* ~Fixes #6755~
   * upon counter check with `main` this was already fixed 

Removes the re_renderer `MeshManager` in its entirety.

The vision here was at some point to provide a manager to deal with both short-lived and long-lived resource that are identified by a content hash/identifier similar to the texture manager. However, as things shifted around this didn't end up being useful: _all_ meshes ever supplied were marked as long-lived and held alive by a ref counter. We ended up just holding on to the ref count in a central place rather than tracking resource creation & destruction properly. Overall architecture of gpu resource handling leaned more and more towards shared ownership via caches, making the MeshManager an unnecessary layer.
The drawback of this is that we have a lot more Arcs which might bite us at some point (overhead + hard to track who owns), but I don't see that happening any time soon (I'm sure this is gonna get quoted a year from now when it stops working 😁 ).

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7531?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7531?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7531)


crates/store/re_chunk/src/transport.rs
```diff
@@ -731,7 +731,7 @@ mod tests {
                 {
                     chunk_roundtrip = &chunk_in_transport;
                 }
-                let chunk_after = Chunk::from_transport(&chunk_roundtrip)?;
+                let chunk_after = Chunk::from_transport(chunk_roundtrip)?;
```
Wumpf, line 734: unrelated to this pr, but showed up whenever I did `clippy --all-targets`

teh-cmc (approved):

crates/store/re_chunk/src/transport.rs
teh-cmc: it breaks CI lol

teh-cmc: okay this should be fixed forever now....

PR #7541, opened 2024-09-27
Show spinner icon while waiting for video decoder

### LINKED ISSUE #7478: Wait indicator for long video seeking operations
We should have some indication of the decoder doing its work if we're waiting for an up-to-date image for more than a given amount of time.

emilk:
### What
* Closes #7478

video-spinner-3

This uses the default `egui::Spinner` loading animation.

It plays it directly, even if there is only a single frame of delay in decoding. I thought this could be annoying, but I think it feels great, and it is a lot simpler than keeping track of for how long we've been waiting.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7541?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7541?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7541)

### DIFF
--- a/crates/viewer/re_space_view_spatial/src/visualizers/videos.rs
+++ b/crates/viewer/re_space_view_spatial/src/visualizers/videos.rs
@@ -191,49 +191,51 @@ impl VideoFrameReferenceVisualizer {
 
             Some(Ok(video)) => {
                 video_resolution = glam::vec2(video.width() as _, video.height() as _);
-                if let Some(texture) =
-                    match video.frame_at(render_ctx, decode_stream_id, video_timestamp.as_seconds())
-                    {
-                        Ok(VideoFrameTexture::Ready(texture)) => Some(texture),
-                        Ok(VideoFrameTexture::Pending(texture)) => {
-                            ctx.viewer_ctx.egui_ctx.request_repaint();
-                            Some(texture)
-                        }
-                        Err(err) => {
-                            self.show_video_error(
-                                ctx,
-                                spatial_ctx,
-                                world_from_entity,
-                                err.to_string(),
-                                video_resolution,
-                                entity_path,
-                            );
-                            None
-                        }
+
+                match video.frame_at(render_ctx, decode_stream_id, video_timestamp.as_seconds()) {
+                    Ok(frame) => {
+                        let texture = match frame {
+                            VideoFrameTexture::Ready(texture) => texture,
+                            VideoFrameTexture::Pending(placeholder) => {
+                                ctx.viewer_ctx.egui_ctx.request_repaint();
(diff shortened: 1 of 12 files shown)


crates/viewer/re_viewer_context/src/space_view/visualizer_system.rs
```diff
@@ -174,4 +174,15 @@ impl VisualizerCollection {
     ) -> impl Iterator<Item = (ViewSystemIdentifier, &dyn VisualizerSystem)> {
         self.systems.iter().map(|s| (*s.0, s.1.as_ref()))
     }
+
+    /// Iterate over all visualizer data that can be downcast to the given type.
+    pub fn iter_visualizer_data<SpecificData: 'static>(
```
Wumpf, line 179: neat, liking this much better than the global method

crates/viewer/re_space_view_spatial/src/visualizers/videos.rs
```diff
@@ -191,49 +191,67 @@ impl VideoFrameReferenceVisualizer {
...
+                            world_from_entity.transform_vector3(glam::Vec3::Y * video_resolution.y);
+
+                        let texture = match frame {
+                            VideoFrameTexture::Ready(texture) => texture,
+                            VideoFrameTexture::Pending(placeholder) => {
+                                // Show loading rectangle:
+                                let min = top_left_corner_position;
+                                let max = top_left_corner_position + extent_u + extent_v;
+                                let center = 0.5 * (min + max);
+                                let diameter = (max - min).truncate().abs().min_element();
```
Wumpf, line 213: this likely breaks in 3D https://github.com/rerun-io/rerun/issues/7543

.. but I'm actually not even sure, gotta check how we place labels on 3D points. Likely we'll have to do a similar thing!

Wumpf (approved):
> It plays it directly, even if there is only a single frame of delay in decoding. I thought this could be annoying, but I think it feels great, and it is a lot simpler than keeping track of for how long we've been waiting.

I concur, this is a lot better than expected. I even like it after playing around with it a bit
I reckon you haven't tried what it looks like when the video is attached to a camera (i.e. plays in 3d)?  (I can see though that you've implemented it)

Follow-up ticket for testing this (and .. everything else!) with 3D views: https://github.com/rerun-io/rerun/issues/7543
looking at the way we insert the 2d spinner here I'm skeptical this will work out of the box, but let's take this separately

Issue #6106, opened 2024-04-24
Adjustable depth image range


roym899:
It would be nice to make the min and max value for depth images adjustable. Currently the minimum and maximum depth determine the color mapping; however, this can lead to unintuitive jumps in color when a far-away value pops in and out of the view. 

This video shows a negative example where the color changes drastically when the sky is not visible.

https://github.com/rerun-io/rerun/assets/9785832/70583d1d-8482-42b2-be0e-38ea93af0896

**Describe the solution you'd like**
- Add an option to adjust the min value and max value for the color map. 
- This setting should be available both for backprojected points and for 2D views. 
- For the backprojection it would also be nice to have an option to exclude points outside of this range. This would improve handling of "non-standard" depth images that use an arbitrary far plane value instead of NaN or 0.

**Additional context**
RViz has this feature ("Normalize Range" can be toggled for depth images and when it's off the "Min Value" and "Max Value" can be set)

Wumpf:
Related to
* #1510

makeecat:
Do we plan to add this feature in the near future? This small issue can turn down quite a few robotics user from using rerun. Inconsistent depth rendering across frames makes depth data meaningless for analysis.

Wumpf:
@makeecat apologies for long wait & lack of reply! It looks like we'll have this finally fixed in the upcoming 0.19 release

Issue #2341, opened 2023-06-08
Allow to specify a tensor's data range via component


Wumpf:
Related to 
* #2274
* #2340 
* https://github.com/rerun-io/rerun/issues/4624
* https://github.com/rerun-io/rerun/issues/6106

We only determine the data range heuristically so far.

Issue #4624, opened 2023-12-29
Floating-point image data range seems to be assumed to be at least 0-255.


kpreid:
**Describe the bug**
When an `archetypes::Image` is logged, if the pixel values do not reach at least 255.0, then none of them will be displayed with full brightness in the viewer. It seems that there is an auto-ranging/auto-exposure feature (logging very large values causes the rest of the image to darken) but it never uses a maximum less than 255.0.

This is contrary to typical usage (in my personal experience in computer graphics and GIS) of floating-point pixel formats, where *if* there is an assumed 100% brightness value, then that value is 1.0.

**To Reproduce**
1. Log an `archetypes::Image` whose `data` contains a `datatypes::TensorBuffer::F16` with values in the 0.0 - 1.0 range.
2. Observe that in the viewer it appears black, yet the pixel data is displayed on hover.
3. Try again with a larger range such as 0.0 - 100.0.
4. Observe that in the viewer the image appears dim.

**Expected behavior**
The auto-ranging should select a range that either matches the data, or has an assumed maximum of 1.0.

I'm not sure what would be best to do with the alpha component — which should generally *not* be auto-scaled since "opaque" is semantically different from "all X% transparent". I haven't investigated Rerun's current behavior with it.

**Screenshots**
![image](https://github.com/rerun-io/rerun/assets/779501/e3e53f47-7971-435e-9f5c-fdc49fbe2d97)

**Desktop (please complete the following information):**
 - OS: macOS Sonoma 14.2.1 (23C71)

**Rerun version**
rerun-cli 0.11.0 [rustc 1.75.0 (82e1608df 2023-12-21), LLVM 17.0.6] x86_64-apple-darwin

emilk:
The heuristics here might need tweaking, and should probably be done in conjunction with actually exposing the image range in the UI, i.e.:

* https://github.com/rerun-io/rerun/issues/1510

The reason for the current default is that it is also quite common to load an image and then convert it to float (perhaps as a side-effect of something else). It would then be weird if Rerun showed the same image differently just because you converted bytes to floats. For instance:


```py
  img = cv2.cvtColor(cv2.imread("img.png", cv2.IMREAD_UNCHANGED), cv2.COLOR_BGRA2RGB)

  rr.log("img_1", rr.Image(img))
  rr.log("img_2", rr.Image(img.astype("float32")))
  rr.log("img_3", rr.Image(1.0 * img)
  rr.log("img_4", rr.Image(0.5 * img)
[...]

PR #7549, opened 2024-09-30
Tensor & depth image value ranges can now be configured (from ui & code)

Wumpf:
### What

* Fixes https://github.com/rerun-io/rerun/issues/6106
* Fixes https://github.com/rerun-io/rerun/issues/2341
* NOT included here is image ranges, this PR is only about things affecting colormaps (https://github.com/rerun-io/rerun/issues/4624)

Depth images (and their clouds):

https://github.com/user-attachments/assets/949ffb14-38e9-452d-a84f-695104f146c8

Tensors:

https://github.com/user-attachments/assets/eea3b38f-e4c7-4b97-9cc3-2fc6104ed6b7

Noteworth ripple effects & considerations in this PR:
* turns out depth image constructor on Python missed a lot of parameters, fixed that
* went with a new `ValueRange` component rather than reusing `Range1D` because the semantics (and names in the ui) of the later were just too unclear
   * we need a tagged component mechanism really badly!
   * decided against putting this on the tensor view as some notes indicated was the plan
* ended up refactoring a few interesting things around image/tensor handling:
   * colormap is no longer part of `ImageInfo`, we already didn't use it for hashing it. Having now a range made it even more clear that this doesn't belong there
   * `finite_range` on tensor & image is now always present, using the datatypes' finite range as fallback. The `None` case handling was awkward at best and non-existant (error out) usually 

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7549?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7549?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7549)


crates/store/re_types/definitions/rerun/archetypes/depth_image.fbs
```diff
@@ -41,17 +41,30 @@ table DepthImage (
...
+  /// This is typically the expected range of valid values.
+  /// Everything outside of the range is clamped to the range for the purpose of colormpaping.
+  /// Note that point clouds generated from this image will still display all points, regardless of this range.
+  ///
+  /// If not specified, the range will be automatically be determined from the data.
+  /// Note that the Viewer may try to guess a wider range than the minimum/maximum of values
+  /// in the contents of the depth image.
+  /// E.g. if all values are positive, some bigger than 1.0 and all smaller than 255.0,
+  /// the Viewer will conclude that the data likely came from an 8bit image, thus assuming a range of 0-255.
+  depth_range: rerun.components.ValueRange ("attr.rerun.component_optional", nullable, order: 3300);
```
emilk, line 55:
```suggestion
  /// If not specified, the range will be automatically estimated from the data.
  /// Note that the Viewer may try to guess a wider range than the minimum/maximum of values
  /// in the contents of the depth image.
  /// E.g. if all values are positive, some bigger than 1.0 and all smaller than 255.0,
  /// the Viewer will guess that the data likely came from an 8bit image, thus assuming a range of 0-255.
  depth_range: rerun.components.ValueRange ("attr.rerun.component_optional", nullable, order: 3300);
```

crates/store/re_types/definitions/rerun/archetypes/depth_image.fbs
```diff
@@ -41,17 +41,30 @@ table DepthImage (
...
+
   /// Scale the radii of the points in the point cloud generated from this image.
   ///
   /// A fill ratio of 1.0 (the default) means that each point is as big as to touch the center of its neighbor
   /// if it is at the same depth, leaving no gaps.
   /// A fill ratio of 0.5 means that each point touches the edge of its neighbor if it has the same depth.
   ///
   /// TODO(#6744): This applies only to 3D views!
-  point_fill_ratio: rerun.components.FillRatio ("attr.rerun.component_optional", nullable, order: 3300);
+  point_fill_ratio: rerun.components.FillRatio ("attr.rerun.component_optional", nullable, order: 3400);
```
emilk, line 64: If only there existed an integer between 3200 and 3300 that `depth_range` could have used 😉

crates/store/re_types/definitions/rerun/archetypes/tensor.fbs
```diff
@@ -19,4 +19,19 @@ table Tensor (
...
+  ///
+  /// This is typically the expected range of valid values.
+  /// Everything outside of the range is clamped to the range for the purpose of colormpaping.
+  /// Any colormap applied for display, will map this range.
+  ///
+  /// If not specified, the range will be automatically be determined from the data.
+  /// Note that the Viewer may try to guess a wider range than the minimum/maximum of values
+  /// in the contents of the tensor.
+  /// E.g. if all values are positive, some bigger than 1.0 and all smaller than 255.0,
+  /// the Viewer will conclude that the data likely came from an 8bit image, thus assuming a range of 0-255.
```
emilk, line 35:
```suggestion
  /// If not specified, the range will be automatically estimated from the data.
  /// Note that the Viewer may try to guess a wider range than the minimum/maximum of values
  /// in the contents of the tensor.
  /// E.g. if all values are positive, some bigger than 1.0 and all smaller than 255.0,
  /// the Viewer will guess that the data likely came from an 8bit image, thus assuming a range of 0-255.
```

crates/store/re_types/definitions/rerun/blueprint/archetypes/tensor_scalar_mapping.fbs
```diff
@@ -17,10 +17,8 @@ table TensorScalarMapping (
     ///
     /// Raises the normalized values to the power of this value before mapping to color.
     /// Acts like an inverse brightness. Defaults to 1.0.
+    ///
+    /// The final value for display is set as:
+    /// `colormap( (value - data_display_range.min) ** gamma / (data_display_range.max - data_display_range.min) ** gamma )`
```
emilk, line 22:
```suggestion
    /// `colormap( ((value - data_display_range.min) / (data_display_range.max - data_display_range.min)) ** gamma )`
```

🤷

crates/viewer/re_data_ui/src/instance_path.rs
```diff
@@ -296,31 +305,53 @@ fn preview_if_image_ui(
...
-        format: image_format.0,
-        kind,
+    let value_range = component_map
+        .get(&components::Range1D::name())
+        .and_then(|colormap| {
+            colormap
+                .component_mono::<components::ValueRange>()
+                .transpose()
+                .ok()
+                .flatten()
```
emilk, line 315: isn't this the same?

```suggestion
                .component_mono::<components::ValueRange>()?
                .ok()
```

crates/viewer/re_viewer_context/src/image_info.rs
```diff
@@ -8,9 +8,34 @@ use re_types::{
...
 };
 
-/// Represents an `Image`, `SegmentationImage` or `DepthImage`.
+/// Colormap together with the range of image values that is mapped to the colormap's range.
 ///
-/// It has enough information to render the image on the screen.
+/// The range is used to linearly re-map the image values to a normalized range (of 0-1)
+/// to which the colormap is applied.
+#[derive(Clone)]
+pub struct ColormapWithMappingRange {
```
emilk, line 16: Nit: I think `ColormapWithRange` is descriptive enough (or `ColomapAndRange`)

crates/viewer/re_space_view_spatial/src/visualizers/depth_images.rs
```diff
@@ -335,9 +363,47 @@ impl VisualizerSystem for DepthImageVisualizer {
...
+                };
+                let cache = ctx.viewer_ctx.cache;
+                let image_stats = cache.entry(|c: &mut ImageStatsCache| c.entry(&image));
+                let default_range =
+                    ColormapWithMappingRange::default_range_for_depth_images(&image_stats);
+                return [default_range[0] as f64, default_range[1] as f64].into();
+            }
+        }
+
+        [0.0, f64::MAX].into()
```
emilk, line 400: Does this mean the default depth range is decided by the range of the "current" depth image, whatever that may be the first time the user starts the viewer?

That seems like a fair heuristic, except for one thing: if the time starts at zero, and the latest-at comes up short, we end up with [0, f64::MAX]. I think it would be better to instead do a latest-ever query to make the chance of this case less likely

crates/viewer/re_space_view_tensor/Cargo.toml
```diff
@@ -29,6 +29,8 @@ re_types.workspace = true
 re_ui.workspace = true
 re_viewer_context.workspace = true
 re_viewport_blueprint.workspace = true
+re_query.workspace = true
```
emilk, line 32: sort!

crates/viewer/re_viewer_context/src/tensor/image_stats.rs
```diff
@@ -14,8 +14,9 @@ pub struct ImageStats {
 
     /// Like `range`, but ignoring all `NaN`/inf values.
     ///
-    /// `None` if there are no finite values at all.
-    pub finite_range: Option<(f64, f64)>,
+    /// If no finite values are present, this takes the maximum finite range
+    /// of the underlying data type.
+    pub finite_range: (f64, f64),
```
emilk, line 19: Smart!

crates/viewer/re_viewer_context/src/tensor/image_stats.rs
```diff
@@ -187,3 +188,21 @@ impl ImageStats {
         }
     }
 }
+
+fn max_finite_range_for_datatype(datatype: ChannelDatatype) -> (f64, f64) {
```
emilk, line 192: This could be a member function on `ChannelDatatype` so that it is easier to find and reuse

crates/viewer/re_viewer_context/src/tensor/tensor_stats.rs
```diff
@@ -172,11 +173,30 @@ impl TensorStats {
...
+        .unwrap_or_else(|| max_finite_range_for_datatype(tensor.dtype()));
 
         Self {
             range,
             finite_range,
         }
     }
 }
+
+fn max_finite_range_for_datatype(datatype: TensorDataType) -> (f64, f64) {
```
emilk, line 186: same here; could be a member function

examples/python/signed_distance_fields/signed_distance_fields/__main__.py
```diff
@@ -117,7 +117,14 @@ def log_sampled_sdf(points: npt.NDArray[np.float32], sdf: npt.NDArray[np.float32
 
 def log_volumetric_sdf(voxvol: npt.NDArray[np.float32]) -> None:
     names = ["width", "height", "depth"]
-    rr.log("tensor", rr.Tensor(voxvol, dim_names=names))
+    # Use a symmetric value range, so that the `cyantoyellow` colormap centers around zero.
+    # Either positive or negative range might be quite small, so don't exceed 1.5x of either.
+    min_val = cast(float, np.min(voxvol))
+    max_val = cast(float, np.max(voxvol))
+    negative_range = max(0.0, abs(min_val))
+    positive_range = max(0.0, abs(max_val))
```
emilk, line 125: an absolute value cannot be less than 0.0

```suggestion
    negative_range = abs(min_val)
    positive_range = abs(max_val)
```

examples/python/signed_distance_fields/signed_distance_fields/__main__.py
```diff
@@ -117,7 +117,14 @@ def log_sampled_sdf(points: npt.NDArray[np.float32], sdf: npt.NDArray[np.float32
...
+    range = max(min(positive_range, negative_range * 1.5), min(negative_range, positive_range * 1.5))
```
emilk, line 126: Is this the same? O.o My brain has a hard time following your code, but I suspect the same will be said by you of my code 😆 

```suggestion
	range = min(max(negative_range, positive_range), 1.5 * min(negative_range, positive_range))
```

examples/python/signed_distance_fields/signed_distance_fields/__main__.py
```diff
@@ -197,7 +204,14 @@ def main() -> None:
             rrb.Vertical(
                 rrb.Horizontal(
                     rrb.Spatial3DView(name="Input Mesh", origin="/world/mesh"),
-                    rrb.TensorView(name="SDF", origin="/tensor"),
+                    rrb.TensorView(
+                        # The cyan to yellow colormap changes its color at the mid point of its range.
+                        # By combining this with a the `value_range` parameter on the tensor,
+                        # we can visualizer negative & positive values effectively.
```
emilk, line 210:
```suggestion
                        # we can visualize negative & positive values effectively.
```

rerun_py/rerun_sdk/rerun/archetypes/depth_image_ext.py
```diff
@@ -45,7 +47,57 @@ def __init__(
...
+
+            This is typically the expected range of valid values.
+            Everything outside of the range is clamped to the range for the purpose of colormpaping.
+            Note that point clouds generated from this image will still display all points, regardless of this range.
+
+            If not specified, the range will be automatically be determined from the data.
+            Note that the Viewer may try to guess a wider range than the minimum/maximum of values
+            in the contents of the depth image.
+            E.g. if all values are positive, some bigger than 1.0 and all smaller than 255.0,
+            the Viewer will conclude that the data likely came from an 8bit image, thus assuming a range of 0-255.
```
emilk, line 86:
```suggestion
            If not specified, the range will be automatically be estimated from the data.
            Note that the Viewer may try to guess a wider range than the minimum/maximum of values
            in the contents of the depth image.
            E.g. if all values are positive, some bigger than 1.0 and all smaller than 255.0,
            the Viewer will guess that the data likely came from an 8bit image, thus assuming a range of 0-255.
```

rerun_py/rerun_sdk/rerun/archetypes/tensor_ext.py
```diff
@@ -37,6 +40,10 @@ def __init__(
             A TensorData object, or type that can be converted to a numpy array.
         dim_names: Sequence[str] | None
             The names of the tensor dimensions when generating the shape from an array.
+        value_range: Sequence[float] | None
+            The range of values to use for colormapping.
+
+            If not specified, the range will be automatically be determined from the data.
```
emilk, line 46:
```suggestion
            If not specified, the range will be estimated from the data.
```

emilk (commented):
Very good!

My only concern is that we don't end up with a bad estimate and then stick with it. In particular, what is the estimated range of an image of a new space view, created when the time cursor is before the image was logged? Maybe it doesn't run the fallback/estimator in that case, and all is good?

emilk (approved):
Very good!

My only concern is that we don't end up with a bad estimate and then stick with it. In particular, what is the estimated range of an image of a new space view, created when the time cursor is before the image was logged? Maybe it doesn't run the fallback/estimator in that case, and all is good?

Wumpf:
> My only concern is that we don't end up with a bad estimate and then stick with it. In particular, what is the estimated range of an image of a new space view, created when the time cursor is before the image was logged? Maybe it doesn't run the fallback/estimator in that case, and all is good?

The behavior shouldn't have changed to before: If no value is present, we evaluate the fallback and fallbacks _are not sticky_ - they get evaluated every frame. This means that we'll always grab the image stats that are available for that frame and those (other than my change regarding how finite_range is defined) are unchanged to before. Which should give us the same behavior.
Notably, this _also_ means that until someone either logs or overrides an image range, they will still get flickering behavior if their depth image or tensor range changes over time.

emilk:
Oh, perfect answer - thank you!

crates/store/re_types/definitions/rerun/archetypes/depth_image.fbs
Wumpf: couldn't thing of one either! damage is done, nothing to do ;-)

crates/viewer/re_data_ui/src/instance_path.rs
Wumpf: ah, yes. much better!

crates/viewer/re_space_view_spatial/src/visualizers/depth_images.rs
Wumpf: latest-ever would require a store subscriber for this and I dont'really want to go there. As discussed on the primary-pr-review-comment-thing-thread, the behavior is as before. so let's leave it be at least for now :)

crates/viewer/re_viewer_context/src/tensor/image_stats.rs
Wumpf: :+1:

Wumpf: turns out we already have min & max values there!

examples/python/signed_distance_fields/signed_distance_fields/__main__.py
Wumpf: oof, must have been late or something

examples/python/signed_distance_fields/signed_distance_fields/__main__.py
Wumpf: I think so, yeah. Liking yours better, captures the intention better!

Issue #7298, opened 2024-08-28
More native Video support


emilk:
## ✅ Av1
We currently support AV1 on native, using a fast software decoder
* https://github.com/rerun-io/rerun/issues/7605

We currently use `re_rav1d` for this, but if possible we want to change to:
* https://github.com/rerun-io/rerun/issues/7607

## 🏗️ H.264
Work on H.264 support is underway in
* https://github.com/rerun-io/rerun/pull/7658

An alternative we haven't explored is to use
* https://docs.rs/openh264/latest/openh264/

## ❌ Other codecs
Should be doable with the same approach we're using for H.264:
* https://github.com/rerun-io/rerun/issues/7607

jprochazk:
(Note that I made the same mistake as in `re_mp4` and uploaded a bunch of binary files to `rav1d-rs`, which need cleaning up via history rewriting and either be turned into LFS files or removed)

emilk:
@jprochazk I summarized my understanding of `rav1d-rs` in https://github.com/rerun-io/rerun/issues/7575 - feel free to edit/comment

Wumpf:
There's always more codecs we'd like to support, but the ones listed here are all in!

For a full overview what we support check
https://rerun.io/docs/reference/video#codec-support

PR #7557, opened 2024-10-01
Native video support for AV1

emilk:
### What
* Part of https://github.com/rerun-io/rerun/issues/7298

### What
Supports native decoding of AV1 videos.

Downsides: it is extremely slow in debug builds

In release builds it is ok, but there is still A LOT of performance on the table.

### TODO before merging
* https://github.com/rerun-io/rerun/pull/7563
* [x] Re-add `nasm` to `pixi.toml`, because it is needed to compile `rav1d`
* [x] Make the asm-features of `rav1d` opt-in so users don't need `nasm` to compile `rerun`
* [x] Put it behind a feature flag, in case compiling `rav1d` is difficult on some platforms
* [x] Fix error handling so we don't crash on bad videos
* ~Fix performance of debug builds, if possible~
  *  `dav1d` is always fast, but `rav1d` is super-slow in debug builds
* [x] Or show error message in debug builds
* [x] Review moved mp4 demux code to see nothing was lost

### Proof
https://github.com/user-attachments/assets/1b1a0a03-9fac-4fa2-bcae-7beb9591067c

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7557?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7557?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7557)


crates/viewer/re_renderer/src/video/decoder/native_av1.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    #[allow(clippy::unwrap_used)] // block_copy_size can only fail for weird compressed formats
+    let block_size = format
+        .block_copy_size(Some(wgpu::TextureAspect::All))
+        .unwrap();
+
+    let bytes_per_row_unaligned = width_blocks * block_size;
+
+    re_tracing::profile_scope!("write_texture");
+    queue.write_texture(
```
Wumpf, line 355: let's not use queue here but instead our own write belt which can be a lot more efficient
(especially since it allows us to move towards writing the bytes directly into a buffer that is read by gpu, meaning we have a copy less)

Wumpf (commented):
preliminary round or reviewing

Cargo.toml
```diff
@@ -308,6 +309,12 @@ zip = { version = "0.6", default-features = false }
 # Our dev profile has some optimizations turned on, as well as debug assertions.
 [profile.dev]
 opt-level = 1 # Make debug builds run faster
+[profile.dev.package.re_video]
+opt-level = 2 # Speed up CPU-side chroma-upsampling (TODO(#7298): move to GPU)
+[profile.dev.package.rav1d]
+opt-level = 2 # Speed up AV1 video decoding. TODO(emilk): this doesn't work. rav1d is dog slow in dev builds.
```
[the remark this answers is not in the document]
emilk: wdym?

crates/viewer/re_renderer/src/video/decoder/native_av1.rs
```diff
@@ -0,0 +1,372 @@
+// TODO(#7298): decode on native
```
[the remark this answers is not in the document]
emilk: for sure, but maybe a good time to do so is when we have more learnings from ffmpeg (rule of three)

crates/store/re_video/src/decode/av1.rs
```diff
@@ -0,0 +1,469 @@
...
+                        re_log::debug!("Command::Flush");
+                        // All pending frames must have already been decoded, because data sent
+                        // through a channel is received in the order it was sent.
+                        output_frames(should_stop, &mut decoder, on_output);
+                        done.try_send(()).ok();
+                        continue;
+                    }
+
+                    Ok(Command::Reset(done)) => {
+                        re_log::debug!("Command::Reset");
```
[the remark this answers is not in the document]
emilk: Yeah, you are right. I've rewritten it all now to be a lot simpler.

crates/store/re_video/src/decode/av1.rs
```diff
@@ -0,0 +1,469 @@
...
+                parker.park_timeout(Duration::from_millis(100));
+                continue;
+            }
+        }
+    }
+}
+
+fn submit_chunk(decoder: &mut dav1d::Decoder, chunk: Chunk, on_output: &OutputCallback) {
+    re_tracing::profile_function!();
+    econtext::econtext_function_data!(format!("chunk timestamp: {:?}", chunk.timestamp));
```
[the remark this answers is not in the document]
emilk: It's useful when debugging crashy code. I used it a lot in my C++ days, but have less use for it in Rust (https://www.ilikebigbits.com/2016_02_28_error_context.html)

crates/viewer/re_renderer/src/video/decoder/native_av1.rs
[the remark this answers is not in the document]
emilk: Maybe the learning will apply to all decoders, but we don't know that yet.

How a browser decoder behaves on bad data might be quite different from how rav1d or ffmpeg behaves, for instance.

crates/store/re_video/src/demux/mod.rs
```diff
@@ -0,0 +1,268 @@
+//! Video demultiplexing.
```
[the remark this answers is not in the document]
emilk: Good call!

crates/store/re_video/src/decode/av1.rs
```diff
@@ -0,0 +1,469 @@
...
+
+    {
+        re_tracing::profile_scope!("send_pending_data");
+        // always attempt to send pending data first
+        // this does nothing if there is no pending data,
+        // and is required if a call to `send_data` previously
+        // returned `EAGAIN`
```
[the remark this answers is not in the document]
emilk: I have rewritten this now

crates/viewer/re_renderer/src/video/decoder/native_av1.rs
emilk: Maybe we can do that in a follow-up PR, e.g. at the same time as when we move the YUV decoding to GPU?

Cargo.toml
```diff
@@ -308,6 +309,12 @@ zip = { version = "0.6", default-features = false }
...
+[profile.dev.package.wgpu]
+opt-level = 2 # Speed up copying video textures
```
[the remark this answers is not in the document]
emilk: Let's revert this

crates/viewer/re_renderer/src/video/decoder/native_av1.rs
emilk: Added to https://github.com/rerun-io/rerun/issues/7605

crates/store/re_video/src/decode/av1.rs
```diff
@@ -0,0 +1,408 @@
...
+            .expect("failed to spawn decoder thread");
+
+        Self {
+            _thread: thread,
+            command_tx,
+            comms,
+        }
+    }
+
+    // NOTE: The interface is all `&mut self` to avoid certain types of races.
```
Wumpf, line 95: ah nice, this makes things so much easier :smile:

crates/store/re_video/src/decode/av1.rs
```diff
@@ -0,0 +1,408 @@
...
+
+    /// Resets the decoder.
+    ///
+    /// This does not block, all chunks sent to `decode` before this point will be discarded.
+    // NOTE: The interface is all `&mut self` to avoid certain types of races.
+    pub fn reset(&mut self) {
+        re_tracing::profile_function!();
+        self.comms
+            .num_outstanding_resets
+            .fetch_add(1, Ordering::SeqCst);
```
Wumpf, line 109: SeqCst always makes me nervous because it's so hard to reason with. This should be `Release` because we want to ensure that at least all the writes so far are visible to the readers (this may or may not include future writes), i.e. at least all previous operations on the channel are published.
On that note, I think an explanation on why it **has** to be before the `send` would be helpful.

crates/store/re_video/src/decode/av1.rs
```diff
@@ -0,0 +1,408 @@
...
+}
+
+type OutputCallback = dyn Fn(Result<Frame>) + Send + Sync;
+
+fn create_decoder() -> Result<dav1d::Decoder, dav1d::Error> {
+    re_tracing::profile_function!();
+
+    let mut settings = dav1d::Settings::new();
+    settings.set_strict_std_compliance(false);
+    settings.set_max_frame_delay(1);
```
Wumpf, line 139: can you add comments what these do any why we want it?

crates/store/re_video/src/decode/av1.rs
```diff
@@ -0,0 +1,408 @@
...
+        let (tx, rx) = crossbeam::channel::bounded(0);
+        self.command_tx.send(Command::Flush { on_done: tx }).ok();
+        rx.recv().ok();
+    }
+}
+
+impl Drop for Decoder {
+    fn drop(&mut self) {
+        re_tracing::profile_function!();
+        self.comms.should_stop.store(true, Ordering::SeqCst);
```
Wumpf, line 126: exact same story as above.

crates/store/re_video/src/decode/av1.rs
```diff
@@ -0,0 +1,408 @@
...
+    let mut decoder = match create_decoder() {
+        Err(err) => {
+            on_output(Err(Error::Initialization(err)));
+            return;
+        }
+        Ok(decoder) => decoder,
+    };
+
+    while let Ok(command) = command_rx.recv() {
+        if comms.should_stop.load(Ordering::SeqCst) {
```
Wumpf, line 156:
```suggestion
        if comms.should_stop.load(Ordering::Acquire) {
```

crates/store/re_video/src/decode/av1.rs
```diff
@@ -0,0 +1,408 @@
...
+            re_log::debug!("Should stop");
+            return;
+        }
+
+        // If we're waiting for a reset we should ignore all other commands until we receive it.
+        let has_outstanding_reset = 0 < comms.num_outstanding_resets.load(Ordering::SeqCst);
```
Wumpf, line 162:
```suggestion
        let has_outstanding_reset = 0 < comms.num_outstanding_resets.load(Ordering::Acquire);
```

Wumpf (approved):
much simpler decoder now, nice!

Looking good, but I stick with my usual line of complaining about seqcst: Seqcst is a bad default and hard to deal with imho (Mara has a really nice section about that in her book if you want to learn more check https://marabos.nl/atomics/memory-ordering.html#common-misconceptions -> `Myth: Sequentially consistent memory ordering is a great default and is always correct.`)

crates/store/re_video/src/decode/av1.rs
```diff
@@ -0,0 +1,408 @@
...
+
+                on_done.send(()).ok();
+            }
+            Command::Reset => {
+                decoder.flush();
+
+                debug_assert!(matches!(decoder.get_picture(), Err(dav1d::Error::Again)),
+                    "There should be no pending pictures, since we output them directly after submitting a chunk.");
+
+                comms.num_outstanding_resets.fetch_sub(1, Ordering::SeqCst);
```
Wumpf, line 184:
```suggestion
                comms.num_outstanding_resets.fetch_sub(1, Ordering::Release);
```
.. but since we're only syncing with ourselves this might actually be `Relaxed`, there's no state to publish here that we need. This one time I'd also go with better safe than sorry ;-)

Issue #7495, opened 2024-09-24
Expand rust APIs to support new Data API concepts


jleibs:
The rust APIs need to facilitate implementing: https://github.com/rerun-io/rerun/issues/7455

There are three main concepts missing:
- A more powerful content expression
- A configurable sampling mechanism
- Making the LatestAt resolution optional

I believe we can actually drop `LatestAtQueryExpression`, and just have a single `QueryExpression` that builds on top of the previous RangeQuery.

The minimal change is to  `QueryExpression` is something like:
```rust
struct ViewContents{
    contents: BTreeMap<EntityPathExpr: BTreeSet<ComponentName>>,
}

// TODO: builder because crazy runtime interactions between filters etc
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct QueryExpression {
    /// The contents that populate the view for the query.
    ///
    /// Example: `world/camera/**`
    pub view_contents: ViewContents,

    /// The timeline to query.
    ///
    /// Example `frame`
    pub timeline: Timeline,

    /// The time range to query.
    pub filtered_index_range: ResolvedTimeRange,

    pub filtered_index_values: Option<Vec<TimeInt>>,

    pub filtered_events_for: Option<ComponentColumnSelector>,

    // Sample times
    pub sampled_index_values: Option<Vec<TimeInt>>,
 
    /// Resolve latest-at
    pub resolve_latest_at: bool,
} 
```

---

Possible post-0.19 improvements:

Alternatively, we could put resolve_latest_at on the `ComponentColumnSelector` which would be more powerful.
```rust
/// Select a component based on its `EntityPath` and `ComponentName`.
///
/// Note, that in the future when Rerun supports duplicate tagged components
/// on the same entity, this selector may be ambiguous. In this case, the
/// query result will return an Error if it cannot determine a single selected
/// component.
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct ComponentColumnSelector {
    /// The path of the entity.
    pub entity_path: EntityPath,

    /// Semantic name associated with this data.
    pub component: ComponentName,

    /// Resolve latest-at
    pub resolve_latest_at: bool,

    /// How to join the data into the `RecordBatch`.
    pub join_encoding: JoinEncoding,
}
```

The additional work related to Views / Contents can be abstracted by python.

PR #7558, opened 2024-10-01
Dataframe API v2 #0: Chunk support for dedupe-latest semantics

teh-cmc:
Implements support for the new `dedupe-latest` :tm: semantics on `Chunk`. This is one of the fundamental primitives required for the new upcoming dataframe APIs.

This requires the use of the `take` Arrow kernel.
Unfortunately I made the mistake of testing that new kernel, which revealed that it is allocating a lot of data when it shouldn't, so I'll have to fix it at some point in a future PR.

* Part of #7495 

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7558?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7558?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7558)

### DIFF
--- a/crates/store/re_chunk/tests/memory_test.rs
+++ b/crates/store/re_chunk/tests/memory_test.rs
@@ -1,0 +1,192 @@
+//! Measures the memory overhead of the chunk store.
+
+// https://github.com/rust-lang/rust-clippy/issues/10011
+#![cfg(test)]
+
+use std::sync::atomic::{AtomicUsize, Ordering::Relaxed};
+
+thread_local! {
+    static LIVE_BYTES_IN_THREAD: AtomicUsize = AtomicUsize::new(0);
+}
+
+pub struct TrackingAllocator {
+    allocator: std::alloc::System,
+}
+
+#[global_allocator]
+pub static GLOBAL_ALLOCATOR: TrackingAllocator = TrackingAllocator {
+    allocator: std::alloc::System,
+};
+
+#[allow(unsafe_code)]
+// SAFETY:
+// We just do book-keeping and then let another allocator do all the actual work.
+unsafe impl std::alloc::GlobalAlloc for TrackingAllocator {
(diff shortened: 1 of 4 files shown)


teh-cmc:
> Unfortunately I made the mistake of testing that new kernel, which revealed that it is allocating a lot of data when it shouldn't, so I'll have to fix it at some point in a future PR.

That's just me being dumb: I keep forgetting that these things are not expressible given how `ListArray` is encoded... We need `ListView` once again.

jleibs:
> We need ListView once again.

Yeah, this definitely seems like an area where that would be nice. At least it looks like the relevant PRs are still moving forward on that one.

> we can always make it fast in a follow-up PR by adding even more cursor shenanigans instead... but let's integrate it all and cement the semantics first before trying to go there.

Strongly agree.

At least at this phase no individual value should be copied more than once.

crates/store/re_chunk/tests/memory_test.rs
```diff
@@ -0,0 +1,193 @@
...
+                    scalars.to_boxed(),
+                    None,
+                )
+            });
+
+            let indices = ArrowPrimitiveArray::from_vec(
+                (0..untaken.0.len() as i32)
+                    .filter(|i| i % 2 == 0)
+                    .collect_vec(),
+            );
```
jleibs, line 157: What happens if we take a contiguous slice? Is there internal optimization for slicing when possible, or does it always copy blindly?

jleibs (approved):

crates/store/re_chunk/tests/memory_test.rs
teh-cmc: Always copies.

PR #7560, opened 2024-10-01
Dataframe API v2 #2: MVP implementation 

teh-cmc:
A first implementation of the new dataframe APIs.
The name is now very misleading though: there isn't anything dataframe-y left in here, it is a row-based iterator with Rerun semantics baked in, driven by a sorted streaming join.

It is rather slow (related: https://github.com/rerun-io/rerun/pull/7558#issuecomment-2386473697), lacks many features and is full of edge cases, but it works.
It does support dedupe-latest semantics (slowly), view contents and selections, chunk overlaps, and pagination (horribly, by virtue of implementing `Iterator`).
It does _not_ support `Clear`s, nor `latest-at` sparse-filling, nor PoVs, nor index sampling. Yet.

Upcoming PRs will be all about fixing these shortcomings one by one.

It should look somewhat familiar:
```rust
let query_cache = QueryCache::new(store);
let query_engine = QueryEngine {
    store,
    cache: &query_cache,
};

let mut query = QueryExpression2::new(timeline);
query.view_contents = Some(
    query_engine
        .iter_entity_paths(&entity_path_filter)
        .map(|entity_path| (entity_path, None))
        .collect(),
);
query.filtered_index_range = Some(ResolvedTimeRange::new(time_from, time_to));
eprintln!("{query:#?}:");

let query_handle = query_engine.query(query.clone());
// eprintln!("{:#?}", query_handle.selected_contents());
for batch in query_handle.into_batch_iter().skip(offset).take(len) {
    eprintln!("{batch}");
}
```

No tests until we have the guarantee that these are the semantics we will commit to.

* Part of #7495 
* Requires #7559

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7558?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7558?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7558)

### DIFF
--- a/crates/store/re_dataframe2/src/query.rs
+++ b/crates/store/re_dataframe2/src/query.rs
@@ -390,7 +390,7 @@ impl QueryHandle<'_> {
     /// ```
     //
     // TODO(cmc): better/actual pagination
-    pub fn next_row(&mut self) -> Option<Vec<Box<dyn ArrowArray>>> {
+    pub fn next_row(&self) -> Option<Vec<Box<dyn ArrowArray>>> {
         re_tracing::profile_function!();
 
         /// Temporary state used to resolve the streaming join for the current iteration.
@@ -664,7 +664,7 @@ impl QueryHandle<'_> {
     ///
     /// See [`Self::next_row`] for more information.
     #[inline]
-    pub fn next_row_batch(&mut self) -> Option<RecordBatch> {
+    pub fn next_row_batch(&self) -> Option<RecordBatch> {
         Some(RecordBatch {
             schema: self.schema().clone(),
             data: ArrowChunk::new(self.next_row()?),
@@ -675,13 +675,13 @@ impl QueryHandle<'_> {
 impl<'a> QueryHandle<'a> {
     /// Returns an iterator backed by [`Self::next_row`].
     #[allow(clippy::should_implement_trait)] // we need an anonymous closure, this won't work
-    pub fn into_iter(mut self) -> impl Iterator<Item = Vec<Box<dyn ArrowArray>>> + 'a {
+    pub fn into_iter(self) -> impl Iterator<Item = Vec<Box<dyn ArrowArray>>> + 'a {
         std::iter::from_fn(move || self.next_row())
     }
 
     /// Returns an iterator backed by [`Self::next_row_batch`].
     #[allow(clippy::should_implement_trait)] // we need an anonymous closure, this won't work
-    pub fn into_batch_iter(mut self) -> impl Iterator<Item = RecordBatch> + 'a {
+    pub fn into_batch_iter(self) -> impl Iterator<Item = RecordBatch> + 'a {
         std::iter::from_fn(move || self.next_row_batch())
     }
 }
--- a/crates/store/re_dataframe2/src/query.rs
+++ b/crates/store/re_dataframe2/src/query.rs
@@ -83,7 +83,7 @@ struct QueryHandleState {
     /// to the query.
     ///
     /// The atomic counter is used as a cursor which keeps track of our current position within
-    /// each invidivual chunk.
+    /// each individual chunk.
     /// Because chunks are allowed to overlap, we might need to rebound between two or more chunks
     /// during our iteration.
     ///
@@ -475,7 +475,7 @@ impl QueryHandle<'_> {
         // We are stitching a bunch of unrelated cells together in order to create the final row
         // that is being returned.
         //
-        // For this reason, we can only guarantee that the index being explicitely queried for
+        // For this reason, we can only guarantee that the index being explicitly queried for
         // (`QueryExpression2::filtered_index`) will match for all these cells.
         //
         // When it comes to other indices that the caller might have asked for, it is possible that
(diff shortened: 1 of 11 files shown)


crates/store/re_dataframe2/src/query.rs
```diff
@@ -0,0 +1,644 @@
...
+            }
+        }
+
+        // What's the index value we're looking for at the current iteration?
+        let cur_index_value = streaming_state_per_component
+            .values()
+            // NOTE: We're purposefully ignoring RowId-related semantics here: we just want to know
+            // the value we're looking for on the "main" index (dedupe semantics).
+            .min_by_key(|streaming_state| streaming_state.index_value)
+            .map(|streaming_state| streaming_state.index_value)?;
```
jleibs, line 465: Rather than doing this on every call to next_row, I suspect it might be clearer to do this whole thing as a 2-phased process.

First, work just with the Timeline data from every view-relevant chunk to materialize a new column of sorted/unique TimeInt values (note, as an added benefit this is the same input you'll want to be able to feed into `sample_index_values()` anyways).  This could still be done incrementally, "batch-wise" by only looking at overlapping chunks on some horizon. 

Then, once we have the ability to iterate over batches of TimeInts, we iterate through them incrementally and look for the matching values from the relevant chunks, as you're doing below, which now becomes a common code-path between this implementation and `sampled_index_values()`

Additionally, my gut is that having batches of unique TimeInts in advance sets us up nicely for some future optimizations.
- It lets us fairly easily parallelize the per-selected-column work. Each worker can independently yield a sequence of rows matching the requested sequence of TimeInts.
- It lets us look ahead to check for matching runs in the given columns. Any time we have a matching run in a range with a single column (happy path) we can directly yield a slice of multiple rows from our column-generator.
- Similarly, null runs can quickly be identified and generated when the last TimeInt in the requested batch is less than the next available time-int for the column.
- The aggregator consuming from each of the parallel columns generators can then yield RecordBatches based on overlapping row-runs from the separate columns, which means in the happy path of dense non-overlapping chunks we return to getting nice contiguous slices again.

crates/store/re_dataframe2/src/query.rs
```diff
@@ -0,0 +1,644 @@
...
+    /// These are already sorted, densified, vertically sliced, and [latest-deduped] according
+    /// to the query.
+    ///
+    /// The atomic counter is used as a cursor which keeps track of our current position within
+    /// each individual chunk.
+    /// Because chunks are allowed to overlap, we might need to rebound between two or more chunks
+    /// during our iteration.
+    ///
+    /// [latest-deduped]: [`Chunk::deduped_latest_on_index`]
+    chunks: HashMap<ComponentName, Vec<(AtomicU64, Chunk)>>,
```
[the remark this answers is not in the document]
teh-cmc: woops, force of habit

crates/store/re_dataframe2/src/query.rs
```diff
@@ -0,0 +1,644 @@
...
+                // NOTE: This cannot possibly return None, see assert above.
+                list_array.map(|list_array| (*component_name, list_array))
+            })
+            .collect();
+
+        // TODO(cmc): It would likely be worth it to allocate all these possible
+        // null-arrays ahead of time, and just return a pointer to those in the failure
+        // case here.
+        let arrays = state
+            .view_contents
```
teh-cmc, line 583: :mag_right: that should be iterated over the selected contents

crates/store/re_dataframe2/src/query.rs
```diff
@@ -0,0 +1,644 @@
...
+                let index_range = self
+                    .query
+                    .filtered_index_range
+                    .unwrap_or(ResolvedTimeRange::EVERYTHING);
+
+                let query = RangeQuery::new(self.query.filtered_index, index_range)
+                    .keep_extra_timelines(true) // we want all the timelines we can get!
+                    .keep_extra_components(false);
+
+                selected_contents
```
jleibs, line 251: I believe this should be `view_contents`; we need to look at all the chunks in the view to determine the timeline values so that rows are consistent regarless of selected columns.

That said, we don't actually need to sort / dedpuc the data-columns of those chunks; only the time columns.

crates/store/re_dataframe2/src/query.rs
teh-cmc: We can do these improvements in follow up PRs, let's focus on landing all semantics first.

jleibs: Agreed -- not a requested change. Just an observation about the structure to keep in mind as you refactor in the direction of supporting `sampled_index_values()`

teh-cmc:
We've integrated all of this in @abey79's work-in-progress dataframe-view -- everything works semantics-wise.

Next steps (future PRs):
* implement all missing features
* make it fast
* minimal testing

jleibs (approved):
:rocket:

PR #7572, opened 2024-10-02
Dataframe view update and blueprint API (part 5): switch to `re_dataframe2`

### LINKED ISSUE #7449: Specify which circumstances lead `LatestAtQueryHandle::get()` to return zero instead of one row
`LatestAtQueryHandle::get()`'s docstring currently states:

> Returns a single [`RecordBatch`] containing a single row, where each cell corresponds to the latest known value at that particular point in time for each respective `ColumnDescriptor`.

This is in practice incorrect, and there are at least two situations where no rows are currently returned:
- if the store is empty, as per the `empty_yields_empty()` test
- if no component columns are selected (only control and time column)

It think the later case might be a bug. Regardless, it's probably worth formally specifying the case(s) in which a latest-at query might return nothing.

Also, `LatestAtQueryHandle` should implement `row_count()` just like `RangeQueryHandle`.

abey79:
### What

- Fixes #7449

This PR does the following:
- removes the old UI and assorted dead code
- switches the dependency from `re_dataframe` to `re_dataframe2`
- wires the new UI to the view
- assorted cleanup

<hr>

Part of a series to address #6896 and #7498.

All PRs:
- #7515
- #7516
- #7527 
- #7545
- #7551
- 

### Checklist


crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -28,69 +29,85 @@ pub(crate) enum HideColumnAction {
...
+    };
 
-impl<'a> From<LatestAtQueryHandle<'a>> for QueryHandle<'a> {
-    fn from(query_handle: LatestAtQueryHandle<'a>) -> Self {
-        QueryHandle::LatestAt(query_handle)
-    }
-}
+    let num_sticky_cols = schema
+        .iter()
+        .take_while(|cd| matches!(cd, ColumnDescriptor::Control(_) | ColumnDescriptor::Time(_)))
```
teh-cmc, line 82: That `take_while` feels fairly risky for very little gain, why not a `filter`?

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -207,22 +214,15 @@ impl<'a> egui_table::TableDelegate for DataframeTableDelegate<'a> {
...
-        );
+        let timeline = self.query_handle.query().filtered_index;
 
-        // TODO(#7449): this can be removed when `LatestAtQueryHandle` is able to report the row count.
-        self.latest_at_query_returns_no_rows = if let Ok(display_data) = &data {
-            matches!(self.query_handle, QueryHandle::LatestAt(_)) && display_data.num_rows() == 0
-        } else {
-            false
-        };
+        //TODO(ab, cmc): we probably need a better way to run a paginated query.
```
teh-cmc, line 219: "probably" doing some heavy lifting here :smile:

teh-cmc (approved):

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
abey79: Well we want to have all control columns and time columns to be sticky, but they _must_ appear first in the ~schema~ selected columns.

If something weird like this happened, we'd still want to have only 3 stick column, not 4 (because that would wrongly include `comp1`:

```
row_id  timeA  timeB comp1 comp2 timeC comp3
```

PR #7589, opened 2024-10-04
Dataframe v2: support for `filtered_index_values`

teh-cmc:
Title.

* DNM: requires #7587

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7589?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7589?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7589)

### DIFF
--- a/crates/store/re_dataframe2/src/query.rs
+++ b/crates/store/re_dataframe2/src/query.rs
@@ -13,7 +13,7 @@ use nohash_hasher::IntMap;
 use re_chunk::{Chunk, RangeQuery, RowId, TimeInt, Timeline};
 use re_chunk_store::{
     ColumnDescriptor, ColumnSelector, ComponentColumnDescriptor, ComponentColumnSelector,
-    ControlColumnDescriptor, ControlColumnSelector, JoinEncoding, QueryExpression2,
+    ControlColumnDescriptor, ControlColumnSelector, IndexValue, JoinEncoding, QueryExpression2,
     TimeColumnDescriptor, TimeColumnSelector,
 };
 use re_log_types::ResolvedTimeRange;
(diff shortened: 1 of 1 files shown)


crates/store/re_dataframe2/src/query.rs
```diff
@@ -527,6 +527,13 @@ impl QueryHandle<'_> {
             .min_by_key(|streaming_state| streaming_state.index_value)
             .map(|streaming_state| streaming_state.index_value)?;
 
+        if let Some(filtered_index_values) = self.query.filtered_index_values.as_ref() {
+            if !filtered_index_values.contains(&cur_index_value) {
+                self.increment_cursors_at_index_value(cur_index_value);
```
jleibs, line 532: I know we're not optimizing yet... but it seems like this could take an optional `next_index_value` which we can easily look up  from `filtered_index_values`. Then we could just increment repeatedly in the inner increment as long as we're less than `next_index_value`.

Additionally, if we made `increment_cursors_at_index_value()` return the minimum value across all the chunks, we could loop here and basically guarantee that the call to `next_row()` will end up with a value included in `filtered_index_values`

jleibs (approved):

crates/store/re_dataframe2/src/query.rs
teh-cmc: Index sampling and clear/tombstone support first -- I don't want to optimize myself into a corner because of weird interactions with other features...

PR #7590, opened 2024-10-04
Fix image & video cache creating new entries when selecting data without explicit media type

### LINKED ISSUE #7584: Selecting a video takes up the memory of that video _again_
Easy to see with very large videos:
image

Once a video is selected, our memory usage goes up by another `sizeof video blob`. Are we creating a new `re_video` because some cache key doesn't match up?

Also there's something weird going on with how memory in the store is reported. But that's another issue.
Edit: This is _sorta_ expected right now, but overall the query stats aren't in a great shape, let's ignore that for now. See https://github.com/rerun-io/rerun/issues/7585 for details

Wumpf:
### What

* Fixes  #7584
   * but also affects images, just not as noticeable there until you kept a blob selected for a long time

That lengthy title is best explained in steps:
* load a large video (or image!) by drag & droping it in
* select it
* observe how memory goes up again!

This happened because image & video both use the media type as part of their cache key entries, but we tend to sometimes resolve it ahead of time and sometimes not, so we get new cache entries.
We _do_ want the media type in the cache entries so we can memorize which (user selected / override affected) media types won't work, but we have to make sure we don't waste memory because of when we resolve the media type.

Before:

https://github.com/user-attachments/assets/c3544b3f-95df-4311-994e-864166f3e1d4

After:

https://github.com/user-attachments/assets/9e1a5be2-3fe6-4242-aa38-4f11e018dbcb

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7590?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7590?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7590)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6209,6 +6209,7 @@ name = "rerun_c"
 version = "0.19.0-alpha.1+dev"
 dependencies = [
  "ahash",
+ "infer",
  "once_cell",
  "parking_lot",
  "re_arrow2",
@@ -6224,6 +6225,7 @@ dependencies = [
  "arrow",
  "crossbeam",
  "document-features",
+ "infer",
  "itertools 0.13.0",
  "mimalloc",
  "once_cell",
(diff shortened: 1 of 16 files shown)


teh-cmc (approved):
I'm not sure why some things got renamed to `MIME` (in this PR, and previous ones)?

Not only is it inconsistent with the rest of the codebase which now sometimes calls things `MIME` and other times `MediaType`, but also these are certainly media types.

Wumpf:
> I'm not sure why some things got renamed to `MIME`

Tried to make a few of the errors more consistent 🤷

Issue #7471, opened 2024-09-23
Unexpected time columns in a range query


abey79:
# Setup

Consider a range query with Position2D as PoV on this data:

```python
def specimen_archetype_with_latest_at_semantics():
    """Archetype across spread over a multi-row chunk and two single-row chunks, with latest-at semantics."""
    rr.send_columns(
        "/archetype_chunk_with_latest_at_semantics",
        frame_times(range(10)),
        [
            rrc.Position2DBatch([(i, i) for i in range(10)]),
            rrc.ClassIdBatch(range(10)),
        ],
    )

    set_frame_time(0)
    rr.log_components("/archetype_chunk_with_latest_at_semantics", [rr.Points2D.indicator()])

    set_frame_time(5)
    rr.log_components("/archetype_chunk_with_latest_at_semantics", [rrc.RadiusBatch(2)])
```
(from https://github.com/rerun-io/rerun/pull/7469)

<img width="398" alt="image" src="https://github.com/user-attachments/assets/72eab764-2738-4733-9a30-190d9ea2e02f">
<img width="555" alt="image" src="https://github.com/user-attachments/assets/c565c79a-2124-4300-8d01-59bcaeed02a7">
<img width="501" alt="image" src="https://github.com/user-attachments/assets/1261940d-3cc8-448c-ae9c-a1d2b70f9697">

# Expected

Given that the POV is only logged on the `frame` timeline, I'd expect only that timeline to be present in the query schema.

# Actual

Both `log_tick` and `log_time` are in the schema.

<img width="1034" alt="image" src="https://github.com/user-attachments/assets/ec35d8c9-3a08-4ecc-9737-b07e937c2a30">

abey79:
Yeah, this is not longer relevant with the new view content/PoV definition.

PR #7598, opened 2024-10-04
Dataframe v2: support for `sparse_fill_strategy`

teh-cmc:
Title.

* DNM: requires #7593

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7589?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7589?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7589)

### DIFF
--- a/crates/store/re_dataframe2/src/query.rs
+++ b/crates/store/re_dataframe2/src/query.rs
@@ -10,11 +10,11 @@ use arrow2::{
 use itertools::Itertools;
 
 use nohash_hasher::IntMap;
-use re_chunk::{Chunk, RangeQuery, RowId, TimeInt, Timeline};
+use re_chunk::{Chunk, RangeQuery, RowId, TimeInt, Timeline, UnitChunkShared};
 use re_chunk_store::{
     ColumnDescriptor, ColumnSelector, ComponentColumnDescriptor, ComponentColumnSelector,
     ControlColumnDescriptor, ControlColumnSelector, IndexValue, JoinEncoding, QueryExpression2,
-    TimeColumnDescriptor, TimeColumnSelector,
+    SparseFillStrategy, TimeColumnDescriptor, TimeColumnSelector,
 };
 use re_log_types::ResolvedTimeRange;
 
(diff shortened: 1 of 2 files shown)


crates/store/re_dataframe2/src/query.rs
```diff
@@ -625,27 +692,31 @@ impl QueryHandle<'_> {
...
-                                .map(|time| {
-                                    (
-                                        *time_column.timeline(),
-                                        (time, time_column.times_array().sliced(cursor, 1)),
-                                    )
-                                })
-                        })
+                    match streaming_state {
+                        StreamingJoinState::StreamingJoinState(s) => s.chunk.timelines(),
+                        StreamingJoinState::Retrofilled(unit) => unit.timelines(),
```
jleibs, line 697: :melting_face: this one hurts my head to think about and has some weird edge cases.

I don't think we should use Retrofilled timestamps to update the values non-primary index columns.

Consider the following table:
```
T1 | T2 | X | Y
---+----+---+---
1  | 3  | A | D
2  | 2  | B | -
3  | 1  | C | -
```

The LatestAt value of T2 when T1=3 is clearly 1, however, my read of this logic is that when executing a LatestAt() on a view that included Y, we would pull forward the 3 Value and let it take precedence because we aggregate by max.

The only situations where it might make sense to do so would be a table that looked like:
```
T1 | T2 | X | Y
---+----+---+---
1  | 3  | A | D
2  | -  | B | -
3  | -  | C | -
```

But even then, if the table looked like:
```
T1 | T2 | X | Y
---+----+---+---
1  | 3  | A | D
2  | 2  | B | -
3  | -  | C | -
```
I would think the LatestAt value of T2 at T1 should still actually be 2, not 3.

jleibs (approved):

crates/store/re_dataframe2/src/query.rs
jleibs: I think this is also related to: https://github.com/rerun-io/rerun/issues/7471

teh-cmc: Max aggregation answers the question "how old is the data that I'm looking at, in aggregate, on timeline Tn?".
The only way to see `Y=D` on `T2` is to move the time cursor to `3` and beyond.

Whether that's the semantics we want, I do not know. I don't think there is generalized right answer anyhow, only the end user knows how they expect these conflicts to be resolved.

I don't mind changing this at any point if we can specify a set of rules to follow.

PR #7612, opened 2024-10-07
Dataframe API: remove control columns from public surface

### LINKED ISSUE #7599: Remove ControlColumn / RowId from the public APIs
Having a whole category of ControlColumnSelector / ControlColumnDescriptor at the query layer adds a lot of complexity just for RowId.

Also, we eventually want RowId to be able to act like an Index in the future. For example: `view(index=("frame", "row_id"), ...`,

So even when we expose it, we will want to make RowId a special kind of IndexColumn rather than a ControlColumn.

It may be helpful to give this a reserved name so that we can select("row_id", "log_time", ...) without worrying about ambiguity of a user-logged `row_id` timeline.

Short-term we don't need this to be an allowable timeline at all.

Longer term it would be nice to be able to use this as an index at query time, but this doesn't need to happen for 0.19.

Notably this could woud likely benefit from a specialized implementation that allowed for some documented exceptions:
- Sorting on RowId is less important
  - Since RowIds should be unique, sorted by chunk rather than row_id would let this be far more performant and return chunk-sized record-batches without the need to do any joining.
- LatestAt doesn't make sense on RowId index.

teh-cmc:
* Prevents ~hard~annoying-to-solve cross-feature interactions in the new API implementation.
* Fixes #7599

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7612?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7612?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7612)

### DIFF
--- a/crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
+++ b/crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
@@ -9,7 +9,7 @@ use re_chunk_store::external::re_chunk::ArrowArray;
 use re_chunk_store::{ColumnDescriptor, LatestAtQuery, RowId};
 use re_dataframe2::QueryHandle;
 use re_log_types::{EntityPath, TimeInt, Timeline, TimelineName};
-use re_types_core::{ComponentName, Loggable as _};
+use re_types_core::ComponentName;
 use re_ui::UiExt as _;
 use re_viewer_context::ViewerContext;
 
@@ -80,7 +80,7 @@ pub(crate) fn dataframe_ui(
 
     let num_sticky_cols = selected_columns
         .iter()
-        .take_while(|cd| matches!(cd, ColumnDescriptor::Control(_) | ColumnDescriptor::Time(_)))
+        .take_while(|cd| matches!(cd, ColumnDescriptor::Time(_)))
         .count();
 
     egui::Frame::none().inner_margin(5.0).show(ui, |ui| {
@@ -135,9 +135,6 @@ struct RowsDisplayData {
 
     /// The index of the time column corresponding to the query timeline.
     query_time_column_index: Option<usize>,
-
-    /// The index of the time column corresponding the row IDs.
-    row_id_column_index: Option<usize>,
 }
 
 impl RowsDisplayData {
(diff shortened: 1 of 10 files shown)


crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
```diff
@@ -393,7 +371,7 @@ impl<'a> egui_table::TableDelegate for DataframeTableDelegate<'a> {
                     column.data_ui(
                         self.ctx,
                         ui,
-                        row_id,
+                        RowId::ZERO,
```
teh-cmc, line 374: Will that be an issue @abey79 ? (I tried to test this myself, but there seems to be an unrelated issue on `main` at the moment, see #tech).

jleibs (approved):

crates/viewer/re_space_view_dataframe/src/dataframe_ui.rs
abey79: This `row_id` ultimately is routed to `ComponentUiRegistry::ui_raw`, which takes an `Option<RowId>`. If we no longer have a row id there, the correct patch is to remove it entirely and pass `None` in the `ui_raw` call site  (`fn data_ui()`  in `display_record_batch.rs`).

As to the effect of doing that, I can't even track where its actually used --> discussion on slack.

abey79: If this blocks you, remove that `RowId` thing entirely and we'll copy later if that proves to be an actual issue. Gut feeling: some hovertooltip might end up being incorrect in edge cases.

abey79: As discussed, the effect of this removal will be that the table cell will no longer preview tiny thumbnail of images. This actually is a good thing because this has performance issues anyways. Some more background in:
- #7617.

PR #7643, opened 2024-10-08
Add a "Getting started" guide for the dataframe API

abey79:
### What

WIP getting started guide. A skeleton with some working code for now. Very much in need of feedback.

may need this to pass CI:
- https://github.com/rerun-io/rerun/pull/7720

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7643?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7643?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7643)


docs/content/getting-started/data-out.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Get data out of Rerun
+order: 700
+---
+
+At its core, Rerun is a database. The viewer includes the [dataframe view](../reference/types/views/dataframe_view) to explore data in tabular form, and the SDK includes an API to export data as dataframes from recording. These features can be used, for example, to perform analysis on the data and log back the results to the original recording.
```
Wumpf, line 6: > At its core, Rerun is a database.

I get what you're getting at, but that's not really how we want to define ourselves though. Rerun Commercial will be a data platform. It's just that the viewer & rrd files act like databases 🤔 

I'll try some other formulation

docs/content/getting-started/data-out.md
```diff
@@ -0,0 +1,16 @@
...
+
+In this three-part guide, we explore such a workflow by implementing an "open jaw detector" on top of our [face tracking example](https://rerun.io/examples/video-image/face_tracking). This process is split into three steps:
+
+1. [Explore a recording with the dataframe view](data-out/explore-as-dataframe)
+2. [Export the dataframe](data-out/export-dataframe)
+3. [Analyze the data and log the results](data-out/analyze-and-log)
+
+Note: this guide uses [Pandas](https://pandas.pydata.org) dataframes because of how popular this package is. The same concept however applies in the same way for alternative dataframe packages such as [Polars](https://pola.rs).
```
Wumpf, line 14:
```suggestion
Note: this guide uses the popular [Pandas](https://pandas.pydata.org) dataframe package. The same concept however applies in the same way for alternative dataframe packages such as [Polars](https://pola.rs).
```

docs/content/getting-started/data-out.md
```diff
@@ -0,0 +1,16 @@
...
+
+For the fast track, jump to the [complete script](data-out/analyze-and-log.md#complete-script) at the end of the third section.
```
Wumpf, line 16:
```suggestion
If you just want to see the final result, jump to the [complete script](data-out/analyze-and-log.md#complete-script) at the end of the third section.
```

docs/content/getting-started/data-out.md
Wumpf: Discussed this internally a lil bit: Given that we _do_ have a [multimodal, ecs] database at the core _and_ we really want people to think about everything in this section about "the database" this formula makes a lot of sense

nikolausWest: I think saying Rerun is a database at its core is very in line with how we want to define ourselves. Partly because we're still trying to get sharper on exactly how to define ourself, there are currently multiple ways we describe Rerun. We are building a data platform that will feel very database-y and be very easy to define as a database (quacks like a duck etc). The one liner is that we are "the multimodal data stack for spatial and embodied AI". That's also reductive of course.

It's a very useful to think of Rerun as a database so I think this is a great formulation for this page

Wumpf:
It would be nice to have these below each other
image

docs/content/getting-started/data-out/analyze-and-log.md
```diff
@@ -0,0 +1,86 @@
...
+
+
+
+In the previous sections, we explored our data and exported it to a Pandas dataframe. In this section, we will analyze the data to extract a "jaw open state" signal and log it back to the viewer.
+
+
+
+## Analyze the data
+
+This is admittedly not the most complicated part of this guide. We already identified that thresholding the `jawOpen` signal at 0.15 is all we need to produce a binary "jaw open state" signal.
```
Wumpf, line 14: strange admission, but I appreciate the effort of trying to make it slightly self-ironic :)
I'd still leave it out because I'm struggling to find a formulation that doesn't sound odd to me

```suggestion
We already identified that thresholding the `jawOpen` signal at 0.15 is all we need to produce a binary "jaw open state" signal.
```

docs/content/getting-started/data-out/analyze-and-log.md
```diff
@@ -0,0 +1,86 @@
...
+
+```python
+rr.init(
+    recording.application_id()
+    recording_id=recording.recording_id(),
+)
+rr.connect()
+```
+
+_Note_: When automating data analysis, it is typically preferable to log the results to an distinct RRD file next to the source RRD (using `rr.save()`). In such case, it is also valid to use the same app ID and recording ID. This allows opening both the source and result RRDs in the viewer, which will display both data under the same recording.
```
Wumpf, line 35:
```suggestion
_Note_: When automating data analysis, it is typically preferable to log the results to an distinct RRD file next to the source RRD (using `rr.save()`). In such a situation, it is also valid to use the same app ID and recording ID. This allows opening both the source and result RRDs in the viewer, which will display data from both files under the same recording.
```

docs/content/getting-started/data-out/analyze-and-log.md
```diff
@@ -0,0 +1,86 @@
...
+Next, let's log the same data as `Text` component:
+
+```python
+target_entity = "/video/detector/faces/0/bbox"
+rr.log_components(target_entity, [rr.components.ShowLabels(True)], static=True)
+rr.send_columns(
+    target_entity,
+    times=[rr.TimeSequenceColumn("frame_nr", df["frame_nr"])],
+    components=[
+        rr.components.TextBatch(np.where(df["jawOpenState"], "OPEN", "CLOSE")),
```
Wumpf, line 64: nit: we didn't import np so far

docs/content/getting-started/data-out/analyze-and-log.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+
+## Log the result back to the viewer
+
+The first step to log the data is to initialize the logging SDK to direct it to the exact same recording that we just analyzed. For this, both the application ID and the recording ID must match. Here is how it is done:
+
+```python
+rr.init(
+    recording.application_id()
```
Wumpf, line 29:
```suggestion
    recording.application_id(),
```

docs/content/getting-started/data-out/analyze-and-log.md
```diff
@@ -0,0 +1,86 @@
...
+rr.send_columns(
+    target_entity,
+    times=[rr.TimeSequenceColumn("frame_nr", df["frame_nr"])],
+    components=[
+        rr.components.TextBatch(np.where(df["jawOpenState"], "OPEN", "CLOSE")),
+    ],
+)
+```
+
+Here we first log the [`ShowLabel`](../../reference/types/components/show_labels.md) component as static to enable the display of the label. Then, we use `rr.send_column()` again to send an entire batch of text labels. We use the [`np.where()`](https://numpy.org/doc/stable/reference/generated/numpy.where.html) to produce a label matching the state for each timestamp.
```
Wumpf, line 69: this one failed for me with 
```
D:\dev\rerun-io\rerun\test.py:34: RerunWarning: send_columns: ValueError(All times and components in a batch must have the same length. Expected length: 355 but got: 1 for component: rerun.components.Text)
```
looks like the np array insn't interpreted as it should be or np.where does not what we think it does. Will check

Wumpf: Huh. That seems to be what https://github.com/rerun-io/rerun/pull/7689 was supposed to fix!

zehiko: yeah, I didn't hit this.

Wumpf: my build was outdated, that was why!

zehiko (approved):

PR #7648, opened 2024-10-09
Remove `re_dataframe` v1 (thus fixing `main`)

teh-cmc:
* [x] yes


teh-cmc:
@rerun-bot full-check

PR #7649, opened 2024-10-09
Dataframe v2: new and improved chunk tools

teh-cmc:
Bunch of improvements and/or additions to the Chunk toolbox that happened as part of the implementation of the dataframe v2 API.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7649?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7649?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7649)

### DIFF
--- a/crates/store/re_chunk/src/chunk.rs
+++ b/crates/store/re_chunk/src/chunk.rs
@@ -707,6 +707,8 @@ impl Chunk {
 
     /// Unconditionally inserts an [`ArrowListArray`] as a component column.
     ///
+    /// Removes and replaces the column if it already exists.
+    ///
     /// This will fail if the end result is malformed in any way -- see [`Self::sanity_check`].
     #[inline]
     pub fn add_component(
@@ -719,6 +721,8 @@ impl Chunk {
     }
 
     /// Unconditionally inserts a [`TimeColumn`].
+    ///
+    /// Removes and replaces the column if it already exists.
     ///
     /// This will fail if the end result is malformed in any way -- see [`Self::sanity_check`].
     #[inline]
--- a/crates/store/re_chunk/src/chunk.rs
+++ b/crates/store/re_chunk/src/chunk.rs
@@ -705,6 +705,22 @@ impl Chunk {
         }
     }
 
+    /// Unconditionally inserts an [`ArrowListArray`] as a component column.
+    ///
+    /// This will fail if the end result is malformed in any way -- see [`Self::sanity_check`].
+    #[inline]
+    pub fn add_component(
+        &mut self,
+        component_name: ComponentName,
+        list_array: ArrowListArray<i32>,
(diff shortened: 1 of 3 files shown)


zehiko (approved):

crates/store/re_chunk/src/chunk.rs
```diff
@@ -705,6 +705,22 @@ impl Chunk {
         }
     }
 
+    /// Unconditionally inserts an [`ArrowListArray`] as a component column.
```
jleibs, line 708: Does unconditionally imply that it will replace the component if it already exists?

crates/store/re_chunk/src/slice.rs
```diff
@@ -450,13 +450,43 @@ impl Chunk {
...
 
+    /// Removes all component columns from the [`Chunk`].
+    ///
+    /// The result is a new [`Chunk`] with the same number of rows and the same index columns, but
+    /// no components.
+    ///
+    /// WARNING: the returned chunk has the same old [`crate::ChunkId`]! Change it with [`Self::with_id`].
+    #[must_use]
+    #[inline]
+    pub fn components_removed(self) -> Self {
```
jleibs, line 468: `chunk.without_components()` reads more intuitively to me but I don't feel strongly

crates/store/re_chunk/src/slice.rs
```diff
@@ -530,6 +560,172 @@ impl Chunk {
...
+        #[cfg(debug_assertions)]
+        #[allow(clippy::unwrap_used)] // debug-only
+        chunk.sanity_check().unwrap();
+
+        Some(chunk)
+    }
+
+    /// Applies a [take] kernel to the [`Chunk`] as a whole.
+    ///
+    /// In release builds, indices are allowed to have null entries (they will be taken as `null`s).
```
jleibs, line 652: What are the situations that cause us to query with null indices? Seems like returning a `ChunkResult` here and always making that an error condition would be preferable.

jleibs (approved):

crates/store/re_chunk/src/chunk.rs
teh-cmc: Yes. I'll expand the comment.

crates/store/re_chunk/src/slice.rs
teh-cmc: We don't, but this is technically part of the public Rust API, so I don't want to punish end users trying to do something that is perfectly valid and apparently well accepted in the broader ecosystem (whether its panics or results, they're both extremely annoying in these filter chains).

PR #7666, opened 2024-10-09
New pixel formats: `Y_U_V24`/`Y_U_V16`/`Y_U_V12``_LimitedRange`/`FullRange`

Wumpf:
### What

Adds a whole bunch of new pixel formats with limited & full range YUV variants.
All of these are needed by our AV1 decoder.

Lots more docs around YUV conversions etc.

Naming & categorizing these formats is quite the struggle just as expected: If limited to YUV variants, the overall picutre becomes sharper (ironic, eh) and it's possible to specify the combinatorics in an almost sane pattern. That's what is happening in re_renderer's YUV converter now.
However (!!!), in the wild naming formats often convolutes a lot of things with each other making this a lot harder. E.g. NV12 is almost (afaik just _almost_) always implying LimitedRange YUV (which is generally a sane default). This, and the fact that we already set out like this, made me continue to use Ocean's naming scheme (and numbering as far as possible) in the user facing API.
Speaking off: The primary motivation for this was to be able to easily test these variants. However, I refrained from throwing color primaries into the mix here because I think they should in fact be separately defined. The docs are intentionally kept vague on this for the moment (trying hard to not further get entangled here). As a result, at least our internal processing is a lot more explicit & more well informed in most places :)

image
(screenshot of the release checklist which is part of this pr (moved from a previous test). I manually tested the color inspection feature against a reference color picker - slight rounding errors occasionally, but they are hard to avoid since GPU/CPU don't always do the exact same thing :/)

DRAFT TODO:
* [ ] some more testing
* [ ] cleanup?
* [ ] line up follow-up PR for use in AV1 to be sure this is going the right way

### Checklist


crates/store/re_types/definitions/rerun/datatypes/pixel_format.fbs
```diff
@@ -24,13 +28,82 @@ enum PixelFormat: ubyte {
     // this organization and subsequently reduce the chance we may find ourselves wanting to
     // change the values in the future.
 
-    /// `NV12` (aka `Y_UV12`) is a YUV 4:2:0 chroma downsampled format with 12 bits per pixel and 8 bits per channel.
+    /// `Y_U_V24` is a YUV 4:4:4 fully planar YUV format without chroma downsampling, also known as `I444`.
+    ///
+    /// This uses limited range YUV, i.e. Y is expected to be within [16, 235] and U/V within [16, 240].
+    ///
+    /// First comes entire image in Y in one plane, followed by the U and V planes.
+    Y_U_V24_LimitedRange = 39,
```
Wumpf, line 36: the ordering in respect to the numbers in here is rather dubious
I tried to order this logically rather than by enum numbers (which are derived from Ocean). But that makes it hard for the codes beyond Ocean :/

.gitignore
```diff
@@ -31,6 +31,7 @@ _deps
 # Python build artifacts:
 __pycache__
 *.pyc
+*.pyd
```
Wumpf, line 34: not sure why this didn't come up before, but while working on this I accidentally pushed a .pyd file which wasn't on the ignore list. better safe than sorry (again), so I added this here

crates/store/re_types/src/image.rs
```diff
@@ -266,29 +266,91 @@ fn test_find_non_empty_dim_indices() {
 
 // ----------------------------------------------------------------------------
 
-// TODO(jan): there is a duplicate of this function in `crates/store/re_video/src/decode/av1.rs`
+// TODO(andreas): Expose this in the API.
+/// Type of color primaries a given image is in.
+///
+/// This applies both to YUV and RGB formats, but if not specified otherwise
+/// we assume BT.709 primaries for all RGB(A) 8bits per channel content (details below on [`ColorPrimaries::Bt709`]).
```
Wumpf, line 273: too eager c&p. not applicable here

crates/store/re_types/src/image.rs
```diff
@@ -266,29 +266,91 @@ fn test_find_non_empty_dim_indices() {
...
-    let g = y - 0.344 * u - 0.714 * v;
-    let b = y + 1.772 * u;
-
-    // BT.709 (aka. HDTV, aka. Rec.709). wiki: https://en.wikipedia.org/wiki/YCbCr#ITU-R_BT.709_conversion
-    // let r = y + 1.575 * v;
-    // let g = y - 0.187 * u - 0.468 * v;
-    // let b = y + 1.856 * u;
+    if limited_range {
+        // "The resultant signals range from 16 to 235 for Y′ (Cb and Cr range from 16 to 240);
+        // the values from 0 to 15 are called footroom, while the values from 236 to 255 are called headroom."
```
Wumpf, line 325: add source. add to the shader as well

crates/store/re_video/src/decode/av1.rs
```diff
@@ -6,7 +6,9 @@ use dav1d::{PixelLayout, PlanarImageComponent};
 
 use crate::Time;
 
-use super::{Chunk, Error, Frame, OutputCallback, PixelFormat, Result, SyncDecoder};
+use super::{
+    Chunk, ColorPrimaries, Error, Frame, OutputCallback, PixelFormat, Result, SyncDecoder,
+};
 
 pub struct SyncDav1dDecoder {
```
Wumpf, line 13: changes in here are half-finished. move them out of the PR if possible

crates/store/re_types/src/datatypes/pixel_format_ext.rs
```diff
@@ -24,24 +42,57 @@ impl PixelFormat {
...
+            | Self::Y_U_V16_FullRange
+            | Self::Y_U_V24_FullRange
+            | Self::Y_U_V12_LimitedRange
+            | Self::Y_U_V16_LimitedRange
+            | Self::Y_U_V24_LimitedRange
+            | Self::NV12
+            | Self::YUY2 => ColorModel::RGB,
+
+            // TODO(andreas): This shouldn't be ColorModel::RGB, but our YUV converter can't do anything else right now.
+            Self::Y8_LimitedRange | Self::Y8_FullRange => ColorModel::RGB,
```
emilk, line 78: Can you elaborate on this?

You mean the GPU always outputs RGB, so we consider `Y` to be a very stupid way of compressing RGB?

What is `color_model` used for? Is there anything that will break due to this "lie"?

crates/store/re_types/src/datatypes/pixel_format_ext.rs
```diff
@@ -70,12 +160,56 @@ impl PixelFormat {
         }
     }
 
+    /// Returns true if the format is a YUV format using
+    /// limited range YUV.
+    ///
+    /// I.e. for 8bit data, Y is valid in [16, 235] and U/V [16, 240], rather than 0-255.
+    pub fn is_limited_yuv_range(&self) -> bool {
```
emilk, line 167: I think this would be nicer as `fn yuv_range(&self) -> Option<YuvRange>` so we can return `None` for non-yuv formats in the future. Also, enums is nicer than bools.

crates/viewer/re_viewer_context/src/tensor/image_stats.rs
```diff
@@ -121,8 +121,8 @@ impl ImageStats {
         // ---------------------------
 
         let datatype = match image.format.pixel_format {
-            Some(PixelFormat::NV12 | PixelFormat::YUY2) => {
-                // We do the lazy thing here:
+            Some(_) => {
+                // We do the lazy thing here since we convert everything to RGB8 right now anyways.
```
emilk, line 125: Maybe it would be smart to prepare ourself for HDR though?

```suggestion
            Some(pixel_format) => {
                let max_value = pixel_format.max_component_value();
                …
```

scripts/ci/main.py
```diff
@@ -0,0 +1,194 @@
...
+        yvu = cv2.cvtColor(bgra, cv2.COLOR_BGR2YCrCb)
+        y, v, u = cv2.split(yvu)
+    else:
+        yuv = cv2.cvtColor(bgra, cv2.COLOR_BGR2YUV)
+        y, u, v = cv2.split(yuv)
+    y = np.array(y).flatten()
+    u = np.array(u).flatten()
+    v = np.array(v).flatten()
+    yuv24 = np.concatenate((y, u, v))
+    return yuv24.astype(np.uint8)
```
emilk, line 27: is there no python library for this? :/

scripts/ci/main.py
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env python3
+"""Testing NV12 image encoding."""
```
emilk, line 2: In `scripts/ci/main.py`? O.o Is this really the main CI check?

emilk (approved):
Niiiice!

crates/store/re_types/definitions/rerun/datatypes/pixel_format.fbs
```diff
@@ -24,13 +28,82 @@ enum PixelFormat: ubyte {
...
+    /// If not for this range limitation/remapping, this is almost identical to 8bit luminace/grayscale (see [datatypes.ColorModel]).
+    Y8_LimitedRange = 41,
+
+    /// Monochrome Y plane only, essentially a YUV 4:0:0 planar format.
+    ///
+    /// Also known as just "gray". This is virtually identical to a 8bit luminance/grayscale (see [datatypes.ColorModel]).
+    ///
+    /// This uses entire range YUV, i.e. Y is expected to be within [0, 255].
+    /// (as opposed to "limited range" YUV as used e.g. in NV12).
+    Y8_FullRange = 30,
```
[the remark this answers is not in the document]
Wumpf: yeah... maybe. See PR description 😄

crates/store/re_types/src/datatypes/pixel_format_ext.rs
Wumpf: yeah but I didn't want to introduce this enum a third time (1x each in re_renderer, re_types and re_video -.-)

crates/store/re_types/src/datatypes/pixel_format_ext.rs
Wumpf: (will add to comment)
GPU doesn't _have_ to always output RGB, but having it sometimes output R(8) _specifically_ for the YUV converter requires me to do more bookkeeping (needs a new renderpipeline and I expect other ripples) which I just don't want to do right now.
We do use this property mostly for things in the hover logic afaik.

> very stupid way of compressing RGB

hm well. no. That's just because our gpu conversion pipeline is stupid atm. But it still doesn't take up extra CPU memory and GPU textures are only held as long as they are in use (image cache for instance doesn't cache gpu transfer step!)

scripts/ci/main.py
Wumpf: dunno, surely but looking and comparing this would have taken me more time than just writing it out myself

scripts/ci/main.py
Wumpf: wait what. what happened here Oo. This file slipped into ci/main for no reason at all. It's the same in tests/python/release_checklist
I have no idea how this came to be and why I didn't notice on review. Maybe it has something to do with the odd .pyd file I leaked in

crates/viewer/re_viewer_context/src/tensor/image_stats.rs
Wumpf: Unfortunately this isn't about the max component value of the pixel format but rather what we show which right now is a bit different. max_component_value would be an _interesting_ one also for limited range YUV stuff :thinking:
I'm leaving an additional note on this for the next person!

crates/store/re_types/src/datatypes/pixel_format_ext.rs
emilk: If it was a good idea in the other places, why isn't it a good idea here?

Wumpf:
having an eye on the missing testing, but will delay a bit here in order to move on to the next thing!

Issue #7668, opened 2024-10-09
Always inline static data in every row of query results that selected a static column


jleibs:
This could either fully replace https://github.com/rerun-io/rerun/issues/7667 as the default behavior, or alternatively complement through introduction of a `static = INLINE` configuration option.

### Motivation
 
When static data is included in a query result, it ends up in a special row 0 with null index values.  Users need to know to either use LatestAt in their query, or know to combine the static data from row 0 with the data from other columns.

However the general understanding of static data is actually somewhat distinct from the concept of logging data at TimeInt::MIN and then joined via LatestAt.

I believe we have semantically suggested that this data should act equivalent to a constant value across all time and all timelines. By that interpretation, Static data should NOT require a LatestAt join. It should just be there regardless of the index or TimeInt you use to query.

### Proposal
- Never include a separate row 0 with static results.
-  When a Column is selected that corresponds to static data, include that same data in every row we return.
- Add an additional helper: `.select_static()` that always returns a single row result with any static data columns.

### Benefits
- More intuitive behavior for new users.
- No special handling of row 0
- Returned Index column never has nulls

### Disadvantages
- Data duplication for users that don't know whey are working with static data
  - Static data can sometimes be used for large things like maps. If  `.select()` includes static columns, the user could end up inadvertently creating situations where they run out of memory.

PR #7677, opened 2024-10-10
Empty column tracker and new dataframe APIs on top

### LINKED ISSUE #7615: Dataframe: track and exclude fully empty columns
## Context:
Transform3D Archetype logs empty arrays for every component in the Transform3D. Because the of combinatorial nature of Transform3D, for most users several of these columns are empty across all rows.  This has both performance implications (see the need for tracking empty transform components here: https://github.com/rerun-io/rerun/pull/7300/) .

## Proposal:
- As chunks are inserted into the store, track whether that column contains any data other than NULL or [].
- Add a field to `ComponentColumnDescriptor` such as `is_empty`, indicating this non-empty information. This is similar context to `is_static` that can help users decide whether they want to include a Column in their selection.
  - This means if you ask for the full schema of the recording, you can see which columns are empty.
- Add a new QueryExpression param `include_empty_columns`.
  - If this is set to `false`, any column which is fully empty will be treated as if it doesn't exist.
  - If you ask for the schema of the VIEW you will not see the empty columns unless this is true.
  - This means these empty columns will not participate in row generation, which should be fine in 99.9% of real use-cases.
 
## Known edge-case:
- If you set: `include_empty_columns=False`, but then you query for that column via Select you will get a FULLY NULL column. You will not see rows where that column was logged as empty.

teh-cmc:
Before:
image

After:
image

Of course with static data this still looks pretty weird in many cases, such as shown in this screenshot,
But that's a job for the upcoming https://github.com/rerun-io/rerun/issues/7668.

---

* Fixes #7615 

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7677?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7677?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7677)

### DIFF
--- a/crates/store/re_chunk_store/src/dataframe.rs
+++ b/crates/store/re_chunk_store/src/dataframe.rs
@@ -6,17 +6,12 @@ use arrow2::{
     array::ListArray as ArrowListArray,
     datatypes::{DataType as ArrowDatatype, Field as ArrowField},
 };
-use nohash_hasher::IntSet;
 
 use re_chunk::TimelineName;
 use re_log_types::{EntityPath, ResolvedTimeRange, TimeInt, Timeline};
 use re_types_core::{ArchetypeName, ComponentName};
 
-use crate::ChunkStore;
-
-// Used all over in docstrings.
-#[allow(unused_imports)]
-use crate::RowId;
+use crate::{ChunkStore, ColumnMetadata};
 
 // --- Descriptors ---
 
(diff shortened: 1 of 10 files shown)


crates/store/re_chunk_store/src/dataframe.rs
```diff
@@ -196,6 +191,17 @@ pub struct ComponentColumnDescriptor {
...
+    /// Whether this column represents an indicator component.
+    pub is_indicator: bool,
+
+    /// Whether this column represents a [`Clear`]-related components.
+    ///
+    /// [`Clear`]: re_types_core::archetypes::Clear
+    pub is_tombstone: bool,
+
+    /// Whether this column contains either no data or only contains null and/or empty values (`[]`).
+    pub is_semantically_empty: bool,
```
jleibs, line 204: Should we just expose `ColumnMetadata` here directly?

jleibs (approved):

crates/store/re_chunk_store/src/dataframe.rs
teh-cmc: Eeeh, as this is the heart of the API, I'd prefer to keep it flat, simple to inspect and simple to build by hand.

Issue #7791, opened 2024-10-17
`rr.log_file_from_path` just hangs indefinitely


teh-cmc:
Try this:
```
curl 'https://app.rerun.io/version/latest/examples/dna.rrd' -o - > /tmp/dna.rrd
```
```python
import sys
import rerun as rr

path_to_rrd = sys.argv[1]
rr.init("rerun_example_dataframe", spawn=True)
rr.log_file_from_path(path_to_rrd)
```
```
python docs/snippets/all/reference/dataframe_view_query.py /tmp/dna.rrd
```

On 0.18.2, this works just fine.
On latest alpha, this does the logging as expected, but then hangs indefinitely. I literally cannot kill the logging process without resorting to the nuclear tools.

teh-cmc:
This seems to be only true with rrd files, passing e.g. a jpeg to that program seems fine.

emilk:
Likely caused by:
* https://github.com/rerun-io/rerun/pull/7475

teh-cmc:
Oh interesting, I see...

emilk:
Seems like there are two problems:
A) why doesn't the above .rrd have an `end-of-stream` marker?
B) why doesn't ctrl-C work?

teh-cmc:
A) The file was saved from the viewer using the `File > Save` feature -- it seems this does not add the proper EOS marker
B) I don't know how this is implemented yet, but I assume this is constantly retrying to `read()`? which would eat the signal

teh-cmc:
Okay I'll fix A to start with, that should be trivial.

teh-cmc:
As for B), looking at the core dump it's relatively safe to assume that's what's happening in the RRD dataloader hangs forever, and therefore the dataloader manager is stuck waiting for its thread:
```
#2  0x000070b8f3253b91 in std::sys::pal::unix::thread::Thread::join () from /home/cmc/dev/rerun-io/rerun/rerun_py/rerun_bindings/rerun_bindings.abi3.so
#3  0x000070b8f2de5b22 in std::thread::JoinInner<T>::join () from /home/cmc/dev/rerun-io/rerun/rerun_py/rerun_bindings/rerun_bindings.abi3.so
#4  0x000070b8f2dd68d9 in re_sdk::recording_stream::RecordingStreamInner::wait_for_dataloaders () from /home/cmc/dev/rerun-io/rerun/rerun_py/rerun_bindings/rerun_bindings.abi3.so
[...]

PR #7796, opened 2024-10-17
Make sure to always write out end-of-stream markers during encoding

teh-cmc:
Saving Rerun RRD data can happen in many, many different ways, and it must always make sure to include the end-of-stream marker.
This is unfortunately not trivial because many things want to partial move out of `Encoder`, which rules out `impl Drop`.

This PR introduces a `DroppableEncoder` that can be used in many cases, or makes sure to properly `finish()` otherwise.

This fixes half of the issues described in:
* #7791 

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7796?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7796?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7796)

### DIFF
--- a/crates/store/re_log_encoding/src/encoder.rs
+++ b/crates/store/re_log_encoding/src/encoder.rs
@@ -206,7 +206,7 @@ pub fn encode(
     write: &mut impl std::io::Write,
 ) -> Result<u64, EncodeError> {
     re_tracing::profile_function!();
-    let mut encoder = Encoder::new(version, options, write)?;
+    let mut encoder = DroppableEncoder::new(version, options, write)?;
     let mut size_bytes = 0;
     for message in messages {
         size_bytes += encoder.append(&message?)?;
@@ -222,7 +222,7 @@ pub fn encode_ref<'a>(
     write: &mut impl std::io::Write,
 ) -> Result<u64, EncodeError> {
     re_tracing::profile_function!();
-    let mut encoder = Encoder::new(version, options, write)?;
+    let mut encoder = DroppableEncoder::new(version, options, write)?;
     let mut size_bytes = 0;
     for message in messages {
         size_bytes += encoder.append(message?)?;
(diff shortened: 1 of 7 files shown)


teh-cmc:
@rerun-bot full-check

teh-cmc:
I can confirm this fixes the `File > Save` feature, and as far as I can tell doesn't break anything otherwise...

crates/store/re_log_encoding/src/encoder.rs
```diff
@@ -120,15 +179,20 @@ impl<W: std::io::Write> Encoder<W> {
         }
     }
 
+    // NOTE: This cannot be done in a `Drop` implementation because of `Self::into_inner` which
+    // does a partial move.
+    #[inline]
     pub fn finish(&mut self) -> Result<(), EncodeError> {
         MessageHeader::EndOfStream.encode(&mut self.write)?;
         Ok(())
```
emilk, line 187: We could warn in Drop if this hasn't been called, at least in debug builds

crates/store/re_log_encoding/src/encoder.rs
```diff
@@ -120,15 +179,20 @@ impl<W: std::io::Write> Encoder<W> {
         }
     }
 
+    // NOTE: This cannot be done in a `Drop` implementation because of `Self::into_inner` which
+    // does a partial move.
+    #[inline]
     pub fn finish(&mut self) -> Result<(), EncodeError> {
```
emilk, line 185: This should maybe take `self` instead?

crates/store/re_log_encoding/src/encoder.rs
```diff
@@ -48,7 +48,66 @@ pub fn encode_to_bytes<'a>(
 
 // ----------------------------------------------------------------------------
 
+/// An [`Encoder`] that can be dropped.
```
emilk, line 51: I mean... the other one can also be dropped :) Maybe "that closes on drop"

emilk (approved):

crates/store/re_log_encoding/src/encoder.rs
teh-cmc: Not sure what you mean? We cannot implement `Drop` for `Encoder`?

crates/store/re_log_encoding/src/encoder.rs
teh-cmc: :grimacing:

crates/store/re_log_encoding/src/encoder.rs
teh-cmc: Can try -- if it doesn't break the world then sure why not

teh-cmc: Ah no cannot do that otherwise I couldn't implement `finish()` in `DroppableEncoder`, and we need it if people want to do err checking.

PR #7800, opened 2024-10-17
Mention in migration guide that .rbl files will no longer load

emilk:
* [x] checkbox

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,9 @@
 
 ## [Unreleased](https://github.com/rerun-io/rerun/compare/latest...HEAD) - Dataframe & Video
 
-TODO(emilk): insert a screenshot and/or code sample here
+<!-- TODO(emilk): insert a screenshot and/or code sample here -->
 
-📖 Release blogpost: TODO(emilk): add link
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ While the web viewer supports a variety of codecs, the native viewer supports on
 Read more about our video supports (and its limits) [in our video docs](https://rerun.io/docs/reference/video).
 
 ### ⚠️ Breaking changes
+* 🗾 Blueprint files (.rbl) from previous Rerun versions will no longer load
```
zehiko, line 31: nit: tiny bit confusing for me as explicit loading will still work, it's just that auto-loading of auto-saved files won't work?

zehiko (approved):

CHANGELOG.md
Wumpf: +1, I think this is confusing. We're unusually stable in this release actually. It's just that we always refuse to load older blueprints automatically

emilk: I didn't realize - thanks for the clarification!

zehiko (approved):

PR #7808, opened 2024-10-17
Dataframe v2 Rust: public dataframe symbols

teh-cmc:
All the symbols I have needed to write the examples (which are yet to be published, but that's another story).

* [x] yes

### DIFF
--- a/crates/store/re_chunk/src/util.rs
+++ b/crates/store/re_chunk/src/util.rs
@@ -10,6 +10,8 @@ use arrow2::{
 };
 use itertools::Itertools;
 
+use crate::TransportChunk;
+
 // ---
 
 /// Returns true if the given `list_array` is semantically empty.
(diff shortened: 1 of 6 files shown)


crates/store/re_chunk/src/util.rs
```diff
@@ -409,3 +411,38 @@ pub fn take_array<A: ArrowArray + Clone, O: arrow2::types::Index>(
         .unwrap()
         .clone()
 }
+
+// ---
+
+use arrow2::{chunk::Chunk as ArrowChunk, datatypes::Schema as ArrowSchema};
+
+/// Concatenate multiple [`TransportChunk`]s into one.
```
teh-cmc, line 419: Just moved this from somewhere else -- it's very useful as an enduser as long as we're stuck with arrow2.

emilk:
@rerun-bot full-check

zehiko (approved):

PR #7824, opened 2024-10-18
Don't keep around additional cpu copy of loaded mesh files

Wumpf:
### What

This is mainly a refactor of `re_renderer`'s model loading pipeline, but as the title (== changelog entry) points out, a nice side-effect that arose culling unnecessary memory usage which may be important for large meshes.

@etaloop's attempt to add a color option to `Asset3D` (see https://github.com/rerun-io/rerun/pull/7458) made it clear that the output of the mesh loaders is really hard to work with:
Prior to this PR, they would eagerly create gpu-sided meshes and then store them alongside an optional (but always filled-out) "cpu mesh" (in essence the unpacked model file).

Now instead all model loading goes to a intermediate `CpuModel` which can be rather easily post processed.
Gpu resources are then created as a separate step, consuming the `CpuModel` (it should be trivial to create a variant that doesn't consume if we need this in the future)

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7824?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7824?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7824)


crates/viewer/re_renderer/src/importer/cpu_model.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/// Like [`MeshInstance`], but for CPU sided usage in a [`CpuModel`] only.
+pub struct CpuMeshInstance {
+    pub mesh: CpuModelMeshKey,
+    pub world_from_mesh: glam::Affine3A,
+    // TODO(andreas): Expose other properties we have on [`MeshInstance`].
+}
+
+/// A model as stored on the CPU.
```
emilk, line 23: What is the difference between a model and a mesh?

How does `CpuModel` relate to `CpuMesh`?
Is there a `GpuModel` ?

crates/viewer/re_renderer/src/importer/cpu_model.rs
```diff
@@ -0,0 +1,92 @@
...
+
+slotmap::new_key_type! {
+    /// Key for identifying a cpu mesh in a model.
+    pub struct CpuModelMeshKey;
+}
+
+/// Like [`MeshInstance`], but for CPU sided usage in a [`CpuModel`] only.
+pub struct CpuMeshInstance {
+    pub mesh: CpuModelMeshKey,
+    pub world_from_mesh: glam::Affine3A,
```
emilk, line 19: should these be `pub`?

crates/viewer/re_renderer/src/importer/stl.rs
```diff
@@ -1,13 +1,10 @@
...
-
 use itertools::Itertools;
 use smallvec::smallvec;
 use tinystl::StlData;
 
 use crate::{
-    mesh::{self, GpuMesh},
-    renderer::MeshInstance,
-    RenderContext,
+    mesh::{self},
```
emilk, line 6:
```suggestion
    mesh,
```

crates/viewer/re_renderer_examples/framework.rs
```diff
@@ -332,7 +332,10 @@ pub fn load_rerun_mesh(re_ctx: &RenderContext) -> Vec<re_renderer::renderer::Mes
     let mut zipped_obj = zip.by_name("rerun.obj").unwrap();
     let mut obj_data = Vec::new();
     std::io::Read::read_to_end(&mut zipped_obj, &mut obj_data).unwrap();
-    re_renderer::importer::obj::load_obj_from_buffer(&obj_data, re_ctx).unwrap()
+    re_renderer::importer::obj::load_obj_from_buffer(&obj_data, re_ctx)
+        .unwrap()
```
emilk, line 336: Can we return errors instead please to set a good example? using `anyhow` should be pretty easy, and probaly reduce line count (`?` is shorter than `.unwrap()`)

emilk (approved):

crates/viewer/re_renderer/src/importer/cpu_model.rs
Wumpf: yes, that makes it easy & convenient to consturct them in mesh loaders. see gltf loader
could make them pub(crate) I guess :shrug:

Issue #7821, opened 2024-10-17
Heuristic stacks video frames rather than giving them their own 2d views


jleibs:
VideoAssets should follow the same heuristics as images

emilk:
In particular: `rerun folder_of_videos/` should produce one space-view per video

emilk:
Half of the work is done in:
* https://github.com/rerun-io/rerun/pull/7850

For the second half, we need to do one or both of the following:

* https://github.com/rerun-io/rerun/issues/7481
* Pass our `Caches` down to `MaxImageDimensionSubscriber`

emilk:
* Closed by https://github.com/rerun-io/rerun/pull/7869

Wumpf:
I'm still getting that error. For instance when using the video check from the checklist (update incoming) without a blueprint

PR #7850, opened 2024-10-21
Refactor `MaxImageDimensionSubscriber`

emilk:
### What
* Part of https://github.com/rerun-io/rerun/issues/7821
* Bonus: better profiling of video loading

The change is that `MaxImageDimensionSubscriber` parses each `.mp4` to figure out its size…
…except I disabled that check last-second because we don't have access to the `VideoCache`, and so this PR would lead to each video getting parsed _twice_.

Still, I would like to get this PR merged as a necessary refactor and generally improved structure.

This is where we're heading:
image

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7850?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7850?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7850)

### DIFF
--- a/crates/viewer/re_space_view_spatial/src/max_image_dimension_subscriber.rs
+++ b/crates/viewer/re_space_view_spatial/src/max_image_dimension_subscriber.rs
@@ -99,8 +99,11 @@ impl ChunkStoreSubscriber for MaxImageDimensionSubscriber {
             // Handle `ImageEncoded`, `AssetVideo`…
             let blobs = event.diff.chunk.iter_component_arrays(&Blob::name());
             let media_types = event.diff.chunk.iter_component_arrays(&MediaType::name());
-            for (blob, media_type) in itertools::izip!(blobs, media_types) {
-                if let Some([width, height]) = size_from_blob(blob.as_ref(), media_type.as_ref()) {
+            for (blob, media_type) in
+                itertools::izip!(blobs, media_types.map(Some).chain(std::iter::repeat(None)))
+            {
+                if let Some([width, height]) = size_from_blob(blob.as_ref(), media_type.as_deref())
+                {
                     let max_dim = self
                         .max_dimensions
                         .entry(event.store_id.clone())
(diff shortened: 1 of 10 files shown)


emilk:
Actually, I just realized this means each video is parsed twice, which is… not great. I'll look into giving `MaxImageDimensionSubscriber` access to `VideoCache`

crates/viewer/re_space_view_spatial/src/max_image_dimension_subscriber.rs
```diff
@@ -89,62 +91,75 @@ impl ChunkStoreSubscriber for MaxImageDimensionSubscriber {
...
+        reader.into_dimensions().ok().map(|size| size.into())
+    } else if media_type.is_video() {
+        re_tracing::profile_scope!("video");
+        if true {
+            None // TODO(#7821): Use the VideoCache here so we make sure we only load each video ONCE
+        } else {
+            re_video::VideoData::load_from_bytes(&blob, &media_type)
+                .ok()
+                .map(|video| video.dimensions())
+        }
```
emilk, line 161: This is the core of it. I'll work on having access to `VideoCache` here, and then we can resolve this in a follow-up PR.

…or we add functionality to `re_mp4` to very quickly parse an .mp4 to only look for its size info 🤔

crates/viewer/re_space_view_spatial/src/max_image_dimension_subscriber.rs
```diff
@@ -89,62 +91,75 @@ impl ChunkStoreSubscriber for MaxImageDimensionSubscriber {
...
-                                .or_default()
-                                .0
-                                .entry(event.diff.chunk.entity_path().clone())
-                                .or_default();
-
-                            max_dim.height = max_dim.height.max(height);
-                            max_dim.width = max_dim.width.max(width);
-                        }
-                    }
+                if let Some([width, height]) = size_from_blob(blob.as_ref(), media_type.as_deref())
```
Wumpf, line 105: nit:
```suggestion
                let Some([width, height]) = size_from_blob(blob.as_ref(), media_type.as_deref()) else {
                      continue;
                };
```

Wumpf (approved):
👍 looks like a good step forward

Issue #7846, opened 2024-10-21
enum SortingStatus::Unsorted conflicts with define in X.h


rgolovanov:
**Describe the bug**
The attempt to compile C++ project with X11 leads to an error

**To Reproduce**
Just run compilation of the project which includes <rerun.hpp>
```
[build] In file included from /usr/include/X11/Xlib.h:44,
[build]                  from /<...>/include/vulkan/vulkan.h:58,
[build] /<...>/include/rerun/time_column.hpp:27:9: error: expected identifier before numeric constant
[build]    27 |         Unsorted = 2,
```

**Expected behavior**
No compiler errors

**Screenshots**
![image](https://github.com/user-attachments/assets/f4a09e4c-fa83-4605-8d3a-e8e5928123da)

**Desktop (please complete the following information):**
 - OS: `Linux 5e5b334a2327 5.15.153.1-microsoft-standard-WSL2 #1 SMP Fri Mar 29 23:14:13 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`

**Rerun version**
0.19.0

emilk:
We can't protect against all weird macros that other libraries might set. But we should at least try to avoid the most common problems, and X11 is pretty common.

# Solution alternatives
I think B is the right call.

## A) do nothing
Users will have to carefully order their `#include`:s, and/or add `#undef Unsorted` to their code.

Status quo.

This is a problem in two alternatives:

A) whenever a user includes `X.h` BEFORE `rerun.hpp`
B) whenever a user includes both `rerun.hpp` and `X.h` and then wants to use `rerun::SortingStatus::Unsorted`

Case A is too common.

## B) Add `#undef Unsorted` to `time_column.hpp`
This will work until someone does these things in this exact order:
* `#include <rerun.hpp>`
* `#include <X.h>`
* Want to use `rerun::SortingStatus::Unsorted`

…but that is a pretty narrow problem now, and has a clear workaround (add `#undef Unsorted` to your user code).

## C) Rename `Unsorted`
We can call it `Unsorted_` in C++, but keep the `Unsorted` name in other APIs. Or call it `Nonsorted` everywhere.
But now it feels like we're negotiating with hostages. What other words are we not allowed to use? 😠

emilk:
I wonder if we should just codegen `#undef Identifier` for every identifier that appears in each of our header files…

![Image](https://github.com/user-attachments/assets/afdff46d-20a0-4cc2-a9e3-89de5e4e7220)

PR #7855, opened 2024-10-21
C++: Improve error message when finding X11 macro `Unsorted`

emilk:
* Closes https://github.com/rerun-io/rerun/issues/7846

Some C library uses macros for constants, which makes those identifiers unusable by any other library in the same global namespace. So anyone including both `X.h` and `rerun.hpp` in the same compilation unit (in that order) would have problems. It's a dumpster-fire.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7855?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7855?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7855)

### DIFF
--- a/rerun_cpp/src/rerun/time_column.hpp
+++ b/rerun_cpp/src/rerun/time_column.hpp
@@ -11,7 +11,10 @@
 // X.h (of X11) has a macro called `Unsorted`
 // See <https://codebrowser.dev/kde/include/X11/X.h.html#_M/Unsorted>
 // and <https://github.com/rerun-io/rerun/issues/7846>.
-#undef Unsorted
+#ifdef Unsorted
+#error \
+    "Found a macro 'Unsorted' (probably from X11), conflicting with `rerun::SortingStatus::Unsorted`. Add '#undef Unsorted' before '#include <rerun.hpp>' to work around this."
+#endif
 
 struct rr_time_column;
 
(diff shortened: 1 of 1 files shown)


emilk:
Something like

```C
#ifdef Unsorted
#    error "There is a macro called 'Unsorted' (probably from X11). Either include <rerun.hpp> FIRST, OR add '#undef Unsorted' before '#include <rerun.hpp>"
#endif
```

?

Wumpf:
I think I have a slight preference for the error message, as rgolovanov says otherwise we might break a user unexpectedly.

> Either include <rerun.hpp> FIRST, 

that part is not a good suggestion because it means that if you only do that you'll still break when trying to use `rerun::SortingStatus::Unsorted`. Yes, it will compile as long as you don't use it, but it just breaks later then in your code

I'd suggest:
```
#ifdef Unsorted
#    error "There is a macro 'Unsorted' (probably from X11) defined, incorrectly replacing one of the variants of `rerun::SortingStatus`. Add '#undef Unsorted' before '#include <rerun.hpp>' to work around this."
#endif
```
that then essentially only assumes that the user will know what happens if they get an error for `Unsorted` not being known to the compiler down the line (which is fairly unlikely, but who knows!)

Wumpf (approved):

Issue #7481, opened 2024-09-24
Optimize mp4 parse times, and avoid duplicating their RAM use


Wumpf:
Right now we effectively hold an mp4 blob at least twice: Raw in the (arrow) store and then again upon parsing in our mp4 object.
All larger pieces should just be pointers back to the raw blob in order to conserve memory.

Also general cleanup of mp4 crate is in order, this is a good opportunity.

This won't fly for streaming mp4 easily, but for full mp4 it should be doable to do indices into the blob.

emilk:
Does arrow bottom out in https://crates.io/crates/bytes ? That would be useful for storing RC-counted references across crates. (though I don't believe re_mp4 needs to own a reference to the bytes, do it?)

emilk:
I think the best path forward is to change `re_mp4` to store byte ranges (`Range<usize>`) rather than the raw bytes, and then have `re_video` (or `re_renderer`) have some sort of `trait ByteSource { fn read(&self, bytes: Range<usize>) -> Option<Vec<u8>> }` that we then can wrap around a `Blob`.

It should be less than a days worth of work.

jprochazk:
Even after https://github.com/rerun-io/rerun/pull/7860 we still copy video data for each sample. But `Blob` is a `Arc<Vec<u8>>` internally, so we should be able to avoid that as well. The decoder could receive a sub-slice of the Blob instead of copying the data to its own `Vec`. For web, it's possible to give the JS video APIs a direct reference to Wasm's linear memory, using the unsafe [`Uint8Array::view`](https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Uint8Array.html#method.view), which means constructing the `EncodedVideoChunk` will not require a copy either.

jprochazk:
Follow-up: https://github.com/rerun-io/rerun/issues/7878

PR #7860, opened 2024-10-22
Make mp4 parsing **a lot** faster & tremendously lower memory overhead

jprochazk:
### What

- Part of https://github.com/rerun-io/rerun/issues/7481
- Depends on https://github.com/rerun-io/re_mp4/pull/12

Some rough numbers:
```
main@de4c389   297.43   µs
this branch      3.8957 µs
```
Measured on `Big_Buck_Bunny_1080_10s_av1.mp4` from our test assets (downloaded via `tests/assets/download_test_assets.py`)

Total Rerun memory usage for a 297 MiB video (the entire Sintel movie):
```
main@de4c389   573 MiB
this branch    325 MiB
```

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7860?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7860?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7860)


jprochazk:
We still copy each sample when we enqueue a chunk, which feels like it should not be necessary. It would be nice to have direct access to the video data arrow buffer in decoders, so that we can index into it there, but I wasn't 100% sure how to make that work here so I left it out.

crates/store/re_video/examples/frames.rs
```diff
@@ -37,11 +37,12 @@ fn main() {
     let mut decoder = re_video::decode::new_decoder(video_path.to_string(), &video)
         .expect("Failed to create decoder");
 
-    write_video_frames(&video, decoder.as_mut(), &output_dir);
+    write_video_frames(&video, &video_data, decoder.as_mut(), &output_dir);
 }
 
 fn write_video_frames(
     video: &re_video::VideoData,
+    video_data: &[u8],
```
Wumpf, line 45: the combination of `re_video::VideoData` and a `video_data` parameter that isn't of that type is massively confusing

crates/store/re_video/src/demux/mod.rs
```diff
@@ -318,6 +296,20 @@ pub struct Sample {
     pub byte_length: u32,
 }
 
+impl Sample {
+    pub fn get(&self, data: &[u8]) -> Option<Chunk> {
```
Wumpf, line 307: plz document what data this expects

crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -175,6 +177,7 @@ fn show_video_blob_info(
     ui_layout: UiLayout,
     video_result: &Result<re_renderer::video::Video, VideoLoadError>,
     video_timestamp: Option<VideoTimestamp>,
+    video_data: &[u8],
```
Wumpf, line 180: passing the blob in here would make this easier to follow

Wumpf (changes requested):
similar to the commentary on the parent PR I'm super happy about this being so low-hanging, but also like there the distinction of "what data is this" needs more love. In partiulcar `VideoData` is already a type and now we're passing a lot of `video_data` which is not related to it.
Ideally we'd use some reference counted object from the guts of Blob, but I haven't looked into whether that's viable (we really don't want to end up passing around raw arrow buffers, infecting everything with that dependency)


If we advertise this in the changelog, it would be nice (almost necessary I'd say!) to have before after perf & memory numbers. Just give it a quick spin with one of the larger videos on our internal video repo, thank you! :)

crates/store/re_video/examples/frames.rs
jprochazk: maybe `video_blob`?

Wumpf: 👍 works for me, that gets it closer to the blob :)

jprochazk:
(trying to measure this by adding a benchmark)

jprochazk:
Some rough numbers:
```
main@de4c389   297.43   µs
this branch      3.8957 µs
```
Measured on `Big_Buck_Bunny_1080_10s_av1.mp4` from our test assets (downloaded via `tests/assets/download_test_assets.py`)

crates/store/re_video/benches/video_load_bench.rs
```diff
@@ -0,0 +1,15 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+
+fn video_load(c: &mut Criterion) {
+    let video = include_bytes!("../../../../tests/assets/video/Big_Buck_Bunny_1080_10s_av1.mp4");
```
Wumpf, line 4: hmmm do you really want to fail compilation only because the user didn't run `download_test_asset`? better to make this dynamic loading

awesome though that you're trying to make this _a thing_ and not just a throw-away/one-off benchmarking 👍

jprochazk: oops

jprochazk:
Running a nightly build to check if the bench works: https://github.com/rerun-io/rerun/actions/runs/11461677788

jprochazk:
Bench was missing `nasm` for the build, maybe it works now: https://github.com/rerun-io/rerun/actions/runs/11461814078

crates/store/re_video/benches/video_load_bench.rs
```diff
@@ -0,0 +1,27 @@
...
+use criterion::{criterion_group, criterion_main, Criterion};
+
+fn video_load(c: &mut Criterion) {
+    let video_path = Path::new(env!("CARGO_MANIFEST_DIR"))
+        .parent()
+        .unwrap()
+        .parent()
+        .unwrap()
+        .parent()
+        .unwrap()
```
Wumpf, line 14: lol! maybe you want to do something about this 😆

Wumpf (approved):
awesome!

crates/store/re_video/benches/video_load_bench.rs
```diff
@@ -6,11 +6,8 @@ use criterion::{criterion_group, criterion_main, Criterion};
 
 fn video_load(c: &mut Criterion) {
     let video_path = Path::new(env!("CARGO_MANIFEST_DIR"))
-        .parent()
-        .unwrap()
-        .parent()
-        .unwrap()
-        .parent()
+        .ancestors()
+        .nth(3)
```
Wumpf, line 10: 🤯

PR #7870, opened 2024-10-23
Improve profiling scopes covering viewer startup

emilk:
### What
Tested with `pixi run rerun-release --profile ../test-videos/` with a folder with a bunch of videos.

With my SDD cache hot (i.e. after a few runs) the time-to-photon for a 650MB video is ~350ms, which is acceptable.

Interestingly enough, "concatenating" a SINGLE arrow array (with nothing) can take 90ms.

image

### Checklist

### DIFF
--- a/crates/store/re_chunk/src/builder.rs
+++ b/crates/store/re_chunk/src/builder.rs
@@ -208,6 +208,7 @@ impl ChunkBuilder {
     /// This returns an error if the chunk fails to `sanity_check`.
     #[inline]
     pub fn build(self) -> ChunkResult<Chunk> {
+        re_tracing::profile_function!();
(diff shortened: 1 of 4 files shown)


teh-cmc (approved):

teh-cmc:
> Interestingly enough, "concatenating" a SINGLE arrow array (with nothing) can take 90ms.

The 90ms part doesn't shock me since I assume it's just copying 650MB of data all over again -- but once again the root of the issue seems to bet yet another arrow kernel that is missing some obvious early outs, a concatenation of 1 should be a noop...
We should have our own kernel wrapper that check for these cases, like we do for `filter()` and `take().

PR #7871, opened 2024-10-23
Trying to ban some symbols via `clippy` but it doesn't seem to work?

teh-cmc:
Why doesn't this work? :angry: 

* [x] no


teh-cmc:
Haaa, I know why. Clippy gets confused by all the package renaming -- this needs to say `re_arrow2`.

PR #7885, opened 2024-10-24
Show list of enabled feature with `rerun --version`

emilk:
* Reverts rerun-io/rerun#7770
* Re-applies https://github.com/rerun-io/rerun/pull/7744

The revert was because during `cargo publish`, a `Cargo.lock` file materialized, and that made cargo angry:

```
Failed to publish re_sdk:
error: failed to verify package tarball

Caused by:
  Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.
  Added: /home/runner/work/rerun/rerun/target/package/re_sdk-0.19.0-alpha.9/Cargo.lock
```

So in this PR we only query about which cargo features if we're NOT in a `cargo publish` build. That should work! We'll know when we next publish some crates :)

### DIFF
--- a/crates/build/re_build_info/src/build_info.rs
+++ b/crates/build/re_build_info/src/build_info.rs
@@ -14,6 +14,9 @@ pub struct BuildInfo {
     /// `CARGO_PKG_NAME`
     pub crate_name: &'static str,
 
+    /// Space-separated names of all features enabled for this crate.
+    pub features: &'static str,
+
     /// Crate version, parsed from `CARGO_PKG_VERSION`, ignoring any `+metadata` suffix.
     pub version: super::CrateVersion,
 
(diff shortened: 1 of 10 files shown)


crates/build/re_build_info/src/build_info.rs
```diff
@@ -147,6 +155,7 @@ impl CrateVersion {
 fn crate_version_from_build_info_string() {
     let build_info = BuildInfo {
         crate_name: "re_build_info",
+        features: "default extra",
```
Wumpf, line 158: what's the `extra` about?

Wumpf (approved):
👍

crates/build/re_build_info/src/build_info.rs
emilk: Just a placeholder for some feature name - this is just for a test

PR #7886, opened 2024-10-24
Docs: "How-to: reuse blueprints across languages"

### LINKED ISSUE #7858: No way to save just a blueprint from the API
I want to save a blueprint to a file for later use (e.g. from another language).

Saving the blueprint from the viewer (`Menu > Save blueprint`) works fine, but doing it from code leads to issues:
```python
import sys

import rerun as rr
import rerun.blueprint as rrb

path_to_rbl = sys.argv[1]

rr.init("rerun_example_dataframe")
rr.save(path_to_rbl)

blueprint = rrb.Blueprint(
    rrb.DataframeView(
        origin="/",
        query=rrb.archetypes.DataframeQuery(
            timeline="log_time",
            apply_latest_at=True,
        ),
    ),
)

rr.send_blueprint(blueprint)

# The output .rbl file now contains an empty data recording too :(
```

The problem is that this also initializes an empty data recording (i.e. `SetStoreInfo` is fired), and therefore loading that blueprint later one will create an empty recording in the viewer, switch focus to it, etc. Makes the whole workflow unusable in practice.

How can we only save the blueprint using the API?

teh-cmc:
By popular demand.

* Fixes #7858
* DNM: requires #7882 

---

* [x] yes

### DIFF
--- a/docs/content/concepts/blueprint.md
+++ b/docs/content/concepts/blueprint.md
@@ -1,5 +1,5 @@
 ---
-title: Blueprint
+title: Blueprints
 order: 600
 ---
 
--- a/docs/content/howto/visualization/configure-viewer-through-code.md
+++ b/docs/content/howto/visualization/configure-viewer-through-code.md
@@ -1,5 +1,5 @@
 ---
-title: Configure the Viewer through code
+title: Configure the Viewer through code (Blueprints)
 order: 100
 ---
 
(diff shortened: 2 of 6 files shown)


teh-cmc:
Bump: objections?

zehiko (approved):
LGTM

PR #7904, opened 2024-10-25
remote store gRPC spec update: introduce application level errors for registration response

zehiko:
### What

remote store errors can be split into 2 categories:
- gRPC layer errors that are always ~~non-retryable errors~~ (see comment  [below](https://github.com/rerun-io/rerun/pull/7904#pullrequestreview-2394842976)) where remote store signals that it wants to cut the responses stream immediately
- application level errors where at the gRPC layer we return a successful response (i.e. ``Status::OK``), but the response itself contains application error

This change introduces application level error for the ``register_record()`` API call, but it is only leveraged for gRPC error, specific for ``Status::details`` .

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7904?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7904?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7904)

### DIFF
--- a/crates/store/re_remote_store_types/proto/rerun/v0/common.proto
+++ b/crates/store/re_remote_store_types/proto/rerun/v0/common.proto
@@ -163,9 +163,9 @@ enum SparseFillStrategy {
 }
 
 // Error codes for application level errors
-enum Code {
+enum ErrorCode {
     // unused
-    _NO_ERROR = 0;
+    _UNUSED = 0;
 
     // object store access error
     OBJECT_STORE_ERROR = 1;
--- a/crates/store/re_remote_store_types/proto/rerun/v0/common.proto
+++ b/crates/store/re_remote_store_types/proto/rerun/v0/common.proto
@@ -161,3 +161,15 @@ enum SparseFillStrategy {
     NONE = 0;
     LATEST_AT_GLOBAL = 1;
 }
+
+// Error codes for application level errors
+enum Code {
+    // unused
+    _NO_ERROR = 0;
+
+    // object store access error
+    OBJECT_STORE_ERROR = 1;
+
+    // metadata database access error
+    METADATA_DB_ERROR = 2;
+}
--- a/crates/store/re_remote_store_types/proto/rerun/v0/remote_store.proto
+++ b/crates/store/re_remote_store_types/proto/rerun/v0/remote_store.proto
@@ -27,11 +27,27 @@ message ObjectStorage {
 }
 
 message RegisterRecordingsResponse {
+    oneof response {
+        RegistrationSuccess success = 1;
+        RegistrationError error = 2;
+    }
+}
+
+message RegistrationSuccess {
     // Note / TODO(zehiko): this implies we read the record (for example go through entire .rrd file
     // chunk by chunk) and extract the metadata. So we might want to 1/ not do this i.e.
     // only do it as part of explicit GetMetadata request or 2/ do it if Request has "include_metadata=true"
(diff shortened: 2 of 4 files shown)


crates/store/re_remote_store_types/proto/rerun/v0/remote_store.proto
```diff
@@ -27,13 +27,29 @@ message ObjectStorage {
 }
 
 message RegisterRecordingsResponse {
+    oneof response {
+        RegistrationSuccess success = 1;
+        RegistrationError error = 2;
+    }
```
teh-cmc, line 33: `RegisterRecordings` is a unary-unary endpoint, therefore I don't see the value in making it return a `oneof` response.

There are no streams involved, no RPC state to be kept alive, etc. We might as well let the client, as well as all the generic pieces of middleware in-between, know that the RPC flat out failed.

Imo:
* `RegisterRecordingsResponse` should only represent success (i.e. it's just `RegistrationSuccess`).
* `RegisterRecordings()` can fail, in which case it returns a gRPC/transport error where `details = RegistrationError{}`.

teh-cmc (commented):
> gRPC layer errors that are always non-retryable errors where remote store signals that it wants to cut the responses stream immediately

gRPC/transport-level errors are still retryable, they're just telling the client that they are the ones that have to re-initiate the whole RPC if they indeed want to retry. This definitely has its place (and in fact `RegisterRecordings` sounds like it falls in that category imo).

The value of pure application-level errors is that the stream (if there is any) can be kept alive while the server-side retries whatever failed automatically, without any action required from the client, and without having to reinitiate all the RPC state etc.

crates/store/re_remote_store_types/proto/rerun/v0/remote_store.proto
zehiko: that makes sense and it simplifies things on both sides

crates/store/re_remote_store_types/proto/rerun/v0/common.proto
```diff
@@ -161,3 +161,15 @@ enum SparseFillStrategy {
     NONE = 0;
     LATEST_AT_GLOBAL = 1;
 }
+
+// Error codes for application level errors
+enum Code {
+    // unused
+    _NO_ERROR = 0;
```
jleibs, line 168: If it is in fact unused lets call it `_UNUSED`.  "NO_ERROR" still sounds like a way of indicating a success-type.

crates/store/re_remote_store_types/proto/rerun/v0/common.proto
```diff
@@ -161,3 +161,15 @@ enum SparseFillStrategy {
     NONE = 0;
     LATEST_AT_GLOBAL = 1;
 }
+
+// Error codes for application level errors
+enum Code {
```
jleibs, line 166: Let's call this `ErrorCode` if it's only to be used for errors.

Otherwise `ReturnCode` is more clear, in which case I would just make `SUCCESS = 0`

zehiko: I was going back and forth on it as well, let's go with ``ErrorCode`` :)

jleibs (approved):

PR #7914, opened 2024-10-28
Fix CpuMesh.albedo_factor overwriting

EtaLoop:
### What
If `Asset3D.albedo_factor` isn't specified, `CpuMesh.albedo_factor` will always be set to `re_renderer::Rgba::WHITE`.
However, in some cases `CpuMesh.albedo_factor` may contain a color that we don't want to be overwritten.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7914?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7914?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7914)

### DIFF
--- a/crates/viewer/re_space_view_spatial/src/mesh_loader.rs
+++ b/crates/viewer/re_space_view_spatial/src/mesh_loader.rs
@@ -55,7 +55,9 @@ impl LoadedMesh {
         // Overwriting albedo_factor of CpuMesh if specified in the Asset3D
         if let Some(albedo_factor) = albedo_factor {
             for instance in &cpu_model.instances {
-                cpu_model.meshes[instance.mesh].materials[0].albedo_factor = albedo_factor.0.into();
+                for material in &mut cpu_model.meshes[instance.mesh].materials {
+                    material.albedo_factor = albedo_factor.0.into();
+                }
             }
         }
 
--- a/crates/viewer/re_space_view_spatial/src/mesh_loader.rs
+++ b/crates/viewer/re_space_view_spatial/src/mesh_loader.rs
@@ -52,7 +52,7 @@ impl LoadedMesh {
             _ => anyhow::bail!("{media_type} files are not supported"),
         };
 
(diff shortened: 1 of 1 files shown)


crates/viewer/re_space_view_spatial/src/mesh_loader.rs
```diff
@@ -52,11 +52,12 @@ impl LoadedMesh {
...
 
-        // Overwriting albedo_factor of CpuMesh in specified in the Asset3D
-        cpu_model.instances.iter().for_each(|instance| {
-            cpu_model.meshes[instance.mesh].materials[0].albedo_factor =
-                albedo_factor.map_or(re_renderer::Rgba::WHITE, |c| c.0.into());
-        });
+        // Overwriting albedo_factor of CpuMesh if specified in the Asset3D
+        if let Some(albedo_factor) = albedo_factor {
+            for instance in &cpu_model.instances {
+                cpu_model.meshes[instance.mesh].materials[0].albedo_factor = albedo_factor.0.into();
```
Wumpf, line 58: I missed that in the previous PR as well, but why are we only changing the first material and not all materials? This should be looping over all materials of the mesh

Wumpf (changes requested):
oops! Should have noticed this in the PR review. Thanks for following up! While we're on it, can you please also fix the other bug pointed out ⬆️

crates/viewer/re_space_view_spatial/src/mesh_loader.rs
EtaLoop: I though materials was a smallvec of size 1. Maybe I saw that in stl/obj and that may not be the case for gltf?
That is my bad :sweat:

Wumpf (approved):

PR #7932, opened 2024-10-29
Expose additional information about decoded frames in the viewer

Wumpf:
### What

image

Minimized by default.

Originally, I wanted to add some more information from the get-go like i/idr/p/b-frame type but turns out we don't really have all that much information around today for sure - a big part of this is that typically just pass in chunks to decoder and get 0-1 frames out for each chunk (my current understand is that it could be 0-n, but n>1 typically doesn't make sense because a single mp4 sample has a single timestamp!). Meaning we typically don't maintain the chunk/frame relationship.
However, this adjusted structure should make it easy to expose optional fields on `FrameInfo` to forward all sort of things

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7932?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7932?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7932)


crates/store/re_video/src/decode/mod.rs
```diff
@@ -198,23 +198,49 @@ pub struct Chunk {
...
     pub presentation_timestamp: Time,
     pub duration: Time,
 }
 
-/// Pixel format/layout used by [`Frame::data`].
+impl Default for FrameInfo {
+    fn default() -> Self {
+        Self {
+            presentation_timestamp: Time::MAX,
+            duration: Time::MAX,
```
emilk, line 225: Should these be optional instead? Or not have a `default()` ?

crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -341,3 +336,71 @@ fn show_video_blob_info(
         }
     }
 }
+
+fn decoded_frame_ui(
```
emilk, line 340: maybe it's time to move these functions to a sub-module, ie. `re_data_ui/src/blob/video.rs` 🤔

crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -341,3 +336,71 @@ fn show_video_blob_info(
...
+        re_format::format_timestamp_seconds(time_range.start.into_secs(timescale),),
+        re_format::format_timestamp_seconds(time_range.end.into_secs(timescale),),
+    )))
+    .on_hover_text("Time range in which this frame is valid.");
+
+    ui.list_item_flat_noninteractive(
+        PropertyContent::new("PTS").value_text(format!("{}", frame_info.presentation_timestamp.0)),
+    )
+    .on_hover_text("Raw presentation timestamp prior to applying the timescale.
+This specifies the time at which the frame should be shown relative to the start of a video stream.");
```
emilk, line 368: For more even indentation:

```suggestion
    .on_hover_text("Raw presentation timestamp prior to applying the timescale.\n\
                    This specifies the time at which the frame should be shown relative to the start of a video stream.");
```

crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -341,3 +336,71 @@ fn show_video_blob_info(
...
+fn source_image_data_format_ui(ui: &mut egui::Ui, format: &SourceImageDataFormat) {
+    let label = "Output format";
+
+    match format {
+        SourceImageDataFormat::WgpuCompatible(format) => {
+            ui.list_item_flat_noninteractive(PropertyContent::new(label).value_text(format!("{format:?}")))
+                // This is true for YUV outputs as well, but for RGB/RGBA there was almost certainly some postprocessing involved,
+                // whereas it would very surprising for YUV.
+                .on_hover_text("Pixel format as returned from the decoder.
+Decoders may do arbitrary post processing, so this is not necessarily the format that is actually encoded in the video data!"
```
emilk, line 380:
```suggestion
                .on_hover_text("Pixel format as returned from the decoder.\n\
                                Decoders may do arbitrary post processing, so this is not necessarily the format that is actually encoded in the video data!"
```

crates/viewer/re_space_view_spatial/src/visualizers/videos.rs
```diff
@@ -203,9 +203,10 @@ impl VideoFrameReferenceVisualizer {
                     ) {
                         Ok(VideoFrameTexture {
                             texture,
-                            time_range: _, // TODO(emilk): maybe add to `PickableTexturedRect` and `PickingHitType::TexturedRect` so we can show on hover?
                             is_pending,
                             show_spinner,
+                            frame_info: _, // TODO(emilk): maybe add to `PickableTexturedRect` and `PickingHitType::TexturedRect` so we can show on hover?
```
emilk, line 208: yes, this would be nice! :)

emilk (approved):
great!

crates/store/re_video/src/decode/mod.rs
Wumpf: the way we implement things we have to force presentation timestamp and I didn't mind sticking with this since that's what _most_ decoders come back with anyways. Duration is more of an issue, but fine for now
setting these to max is consistent with how we handled "lack of frame" so far.
I'll note that down in the code!

Issue #7830, opened 2024-10-18
Communicate that python 3.8 support is deprecated


jleibs:
It's past end-of-life. We should drop it for 0.20.

Probably requires updating several doc pages.

PR #7933, opened 2024-10-29
Officially deprecate support for python 3.8

jleibs:
### What
- Resolves: https://github.com/rerun-io/rerun/issues/7830

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7933?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7933?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7933)

### DIFF
--- a/docs/content/reference/migration/migration-0-20.md
+++ b/docs/content/reference/migration/migration-0-20.md
@@ -25,3 +25,8 @@ You can learn more about Rerun's application model and the different servers and
 `re_query::Caches` has been renamed `re_query::QueryCache`, and similarly for `re_query::CacheKey`.
 
 Note that this doesn't affect `re_dataframe`, where this type was already re-exported as `QueryCache`.
+
+## ❗ Deprecations
+
+Support for Python 3.8 is being deprecated. Python 3.8 is past end-of-life. See: https://devguide.python.org/versions/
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
neat!
haven't tried whether the warning triggers myself, but I trust you did :)

abey79 (approved):
Well done. Just to be sure, do users have a means to silence those warning? They could be very annoying for a (still legitimate) use under 3.8. (Or do we think this barely happens anyways?)

jleibs:
Python actually silences deprecation warnings by default in most user-facing contexts, which is a bit annoying and practically reduces the chances it will be seen by the users in practice, but that's the policy and I'm not going to fight it or workaround it.

Regardless, this is the "most right" way to handle it. If you enable deprecation warnings in your project config you'll see it, otherwise it won't be visible.

jleibs:
Also worth noting that when we drop support, the new wheels will list their incompatibility so people installing will just get an older version or a pypi error if they try to install an explicit version in a 3.8 context.

PR #7934, opened 2024-10-29
Implement safe storage handles

### LINKED ISSUE #7486: Introduce cheaply clonable `ChunkStoreHandle` and `QueryCacheHandle`
These are required for:
1. Making it easy to construct and pass `QueryEngine`s around:
  ```rust
  // TODO(cmc): This needs to be a refcounted type that can be easily be passed around: the ref has
  // got to go. But for that we need to generally introduce `ChunkStoreHandle` and `QueryCacheHandle`
  // first, and this is not as straightforward as it seems.
  pub struct QueryEngine<'a> {
      pub store: &'a ChunkStore,
      pub cache: &'a Caches,
  }
  ```
2. Getting ingestion off of the main thread and parallelize it:
  * #4298

teh-cmc:
This implements `ChunkStoreHandle` & `QueryCacheHandle` which, among other things, allow for streaming dataframes across FFI and network barriers.
```rust
/// A ref-counted, inner-mutable handle to a [`QueryCache`].
///
/// Cheap to clone.
///
/// It is possible to grab the lock behind this handle while _maintaining a static lifetime_, see:
/// * [`QueryCacheHandle::read_arc`]
/// * [`QueryCacheHandle::write_arc`]
#[derive(Clone)]
pub struct QueryCacheHandle(Arc<parking_lot::RwLock<QueryCache>>);

/// A ref-counted, inner-mutable handle to a [`ChunkStore`].
///
/// Cheap to clone.
///
/// It is possible to grab the lock behind this handle while _maintaining a static lifetime_, see:
/// * [`ChunkStoreHandle::read_arc`]
/// * [`ChunkStoreHandle::write_arc`]
#[derive(Clone)]
pub struct ChunkStoreHandle(Arc<parking_lot::RwLock<ChunkStore>>);
```

Those handles on their own are extremely problematic though: letting them loose all across the codebase effectively wraps the entire codebase in a semantically-unsafe{} block where every row interacting with these handles can lead to very nasty race conditions and even deadlocks.
That's why this PR also introduces the `StorageEngine` type, which makes using these handles actually safe in practice:
```rust
/// Keeps track of handles towards a [`ChunkStore`] and its [`QueryCache`].
///
/// A [`StorageEngine`] doesn't add any feature on top of what [`ChunkStoreHandle`] and
/// [`QueryCacheHandle`] already offer: the job of the [`StorageEngine`] is to leverage the type
/// system in order to protect against deadlocks and race conditions at compile time.
///
/// The handles stored within will never be publicly accessible past construction.
///
/// The underlying [`ChunkStore`] and [`QueryCache`] can be accessed through one of the
/// following methods:
/// * [`StorageEngine::read`]
/// * [`StorageEngine::read_arc`]
/// * [`StorageEngine::write`]
/// * [`StorageEngine::write_arc`]
#[derive(Clone)]
pub struct StorageEngine {
    store: ChunkStoreHandle,
    cache: QueryCacheHandle,
}
```

Balancing these safety guarantees with the flexibility we need (today... and tomorrow!) for all our corner use cases has proven to be a subtle art...

* Fixes #7486 

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7934?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7934?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7934)


teh-cmc:
Quickly drafting to check whether there are some remnants from the first four iterations that have made it into the final thing and that could just be removed/simplified now, which is very possible because it's all a big blur at this point :melting_face:

teh-cmc:
> Quickly drafting to check whether there are some remnants from the first four iterations that have made it into the final thing and that could just be removed/simplified now, which is very possible because it's all a big blur at this point 🫠

Nope, I mostly just ran into differently shaped walls.

emilk:
Fix coming in:
* https://github.com/rerun-io/rerun/pull/7952

teh-cmc:
:pray:

crates/store/re_query/src/storage_engine.rs
```diff
@@ -0,0 +1,279 @@
...
+            cache: parking_lot::RwLock::read_arc(ArcRwLockReadGuard::rwlock(&self.cache)),
+        }
+    }
+}
+
+// --- Write Guards ---
+
+// NOTE: None of these fields should ever be publicly exposed, either directly or through a method,
+// as it is always possible to go back to an actual `RwLock` via `RwLockWriteGuard::rwlock`.
+// Doing so would defeat the deadlock protection that the `StorageEngine` offers.
```
zehiko, line 197: perhaps it's worth documenting an example use case of doing the wrong thing without the protection that StorageEngine provides?

zehiko (approved):
I learned a lot reviewing this PR and I really appreciate nice documentation! I believe it's clear to me how this addresses use cases like streaming and why the protection is in place, but folks with more experience with the code base I'm sure can provide more valuable feedback.

PR #7942, opened 2024-10-30
Add `VisualizerSystemError` and `SpaceViewClassError`

grtlr:
### What

In some instances, visualizers and views can run into errors, because the data is not well-formed. As an example, consider a graph layout algorithm that was chosen by the user that only works on DAGs but a cyclic graph was provided. Other examples would be node fixed node positions, supplied by the user, that lead to a non-well defined layout configration .

While Rerun's data model should prevent most of these errors, there still can be places where an escape hatch might be required. A small discussion can be found here: https://github.com/rerun-io/rerun/pull/7500#discussion_r1816888067

This PR adds two variants, `VisualizerSystemError` and `SpaceViewClassError` to `SpaceViewSystemExecutionError`.

:warning: Open problem: these systems often run on every frame, leading to many errors, most of them are not going away until the user changes something in their logging or blueprint logic. This leads to overflowing error messages in the _toast_ notifications. Ideally we would have something like backing up, or retrial logic.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7942?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7942?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7942)

### DIFF
--- a/crates/viewer/re_viewer_context/src/space_view/mod.rs
+++ b/crates/viewer/re_viewer_context/src/space_view/mod.rs
@@ -23,6 +23,7 @@ pub use highlights::{
     SpaceViewOutlineMasks,
 };
 pub use named_system::{IdentifiedViewSystem, PerSystemEntities, ViewSystemIdentifier};
+use re_types::SpaceViewClassIdentifier;
 pub use space_view_class::{
     SpaceViewClass, SpaceViewClassExt, SpaceViewClassLayoutPriority, SpaceViewState,
     SpaceViewStateExt, VisualizableFilterContext,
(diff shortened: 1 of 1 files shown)


grtlr:
Thinking some more about this, we probably want each viewer to handle its own error messages (like `re_space_spatial_view` is doing for `re_video` errors) as it leads to much better UI. The only downside of this is that error messages can look slightly different from view to view.

PR #7943, opened 2024-10-30
Drag-n-drop: always use open semantics for rrd files

teh-cmc:
Always use open semantics, as opposed to import semantics, for .rrd files. I.e. .rrd files will always be opened in their own recording, even with drag-n-drop.

Just reverting to the existing behavior basically.

cc @nikolausWest 

* [x] yes

### DIFF
--- a/crates/store/re_data_loader/src/loader_rrd.rs
+++ b/crates/store/re_data_loader/src/loader_rrd.rs
@@ -87,14 +87,11 @@ impl crate::DataLoader for RrdLoader {
                     .name(format!("decode_and_stream({filepath:?})"))
                     .spawn({
                         let filepath = filepath.clone();
-                        let settings = settings.clone();
                         move || {
                             decode_and_stream(
-                                &filepath,
-                                &tx,
-                                decoder,
-                                settings.opened_application_id.as_ref(),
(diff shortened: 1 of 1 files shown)


nikolausWest:
Do we indicate what will happen when we drag a file over the viewer? If so, that would alleviate this inconsistency a lot. Something like "import into open recording" vs "open as new recording" for example

teh-cmc:
> Do we indicate what will happen when we drag a file over the viewer? If so, that would alleviate this inconsistency a lot. Something like "import into open recording" vs "open as new recording" for example

The ability to change the behavior based on where you drop was only recently added to `winit` iiuc (we'd like to use it, but first we need to update egui and then expose all these new features etc).
So at the moment we're stuck with a single drag-n-drop behavior, which in the case of rrd is probably better to always be `open`.

emilk (approved):

nikolausWest:
> The ability to change the behavior based on where you drop was only recently added to winit iiuc (we'd like to use it, but first we need to update egui and then expose all these new features etc).
So at the moment we're stuck with a single drag-n-drop behavior, which in the case of rrd is probably better to always be open.

I didn't mean changing the behavior, only displaying a label saying what will happen when you drop the file

PR #7949, opened 2024-10-30
Remove the upper bound constraint on python version

jleibs:
### What

We previously had this constraint in place to avoid certain edge cases of users not being able to install Rerun and not understanding why due to transitive deps. Primarily on pyarrow.

I've confirmed pyarrow is now doing something similar in their own project:
https://github.com/apache/arrow/blob/main/python/pyproject.toml#L37

Let's try to let the package solver do the right thing and only add a max-version if we find we are truly incompatible with some future python version.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7949?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7949?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7949)

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -19609,6 +19609,15 @@ packages:
 - kind: pypi
   name: google-crc32c
   version: 1.6.0
+  url: https://files.pythonhosted.org/packages/67/72/c3298da1a3773102359c5a78f20dae8925f5ea876e37354415f68594a6fb/google_crc32c-1.6.0.tar.gz
+  sha256: 6eceb6ad197656a1ff49ebfbbfa870678c75be4344feb35ac1edf694309413dc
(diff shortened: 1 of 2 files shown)


hgaiser:
Interestingly packages like pytorch don't list an upper limit, even though they are not always compatible with the latest python.

https://github.com/pytorch/pytorch/blob/6e85266a47908c917fe7bcf7d2b837711a1f1cab/setup.py#L1509

https://github.com/pytorch/pytorch/issues/130249

PR #7964, opened 2024-10-31
Implement support for fully asynchronous `QueryHandle`s 

teh-cmc:
This adds non-blocking methods to all our new shiny storage handles, and uses these new non-blocking primitives to implement asynchronous helpers (read: `Future`s) in our `QueryHandle`.
These helpers are then used on the other side to implement a proper `Stream`.

In particular, all read locks are now recursive, always.
Recursive locks are mandatory for two reasons:
* As we start parallelizing the viewer further, and especially when using work-stealing schedulers, nested read locks will become a common occurrence.
* In asynchronous contexts, everything is designed around work stealing.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7964?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7964?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7964)

### DIFF
--- a/crates/store/re_dataframe/src/query.rs
+++ b/crates/store/re_dataframe/src/query.rs
@@ -858,9 +858,10 @@ impl<E: StorageEngineLike> QueryHandle<E> {
             Retrofilled(UnitChunkShared),
         }
 
-        // That's a synchronous lock, so make sure we barely lock it.
-        let state = self.init_(store, cache);
-        let state = self.state.get_or_init(move || state);
+        // Although that's a synchronous lock, we probably don't need to worry about it until
+        // there is proof to the contrary: we are in a specific `QueryHandle` after all, there's
+        // really no good reason to be contending here in the first place.
+        let state = self.state.get_or_init(move || self.init_(store, cache));
 
         let row_idx = state.cur_row.fetch_add(1, Ordering::Relaxed);
         let cur_index_value = state.unique_index_values.get(row_idx as usize)?;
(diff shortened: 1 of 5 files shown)


teh-cmc:
@rerun-bot full-check

crates/store/re_dataframe/src/query.rs
```diff
@@ -816,7 +858,9 @@ impl<E: StorageEngineLike> QueryHandle<E> {
             Retrofilled(UnitChunkShared),
         }
 
-        let state = self.init();
+        // That's a synchronous lock, so make sure we barely lock it.
+        let state = self.init_(store, cache);
```
jleibs, line 862: This looks suspicious to me -- the body of that function is large and I wouldn't expect us to execute that code unless we aren't able to access `self.state`.

Why isn't this:
```
self.state.get_or_init(|| self.init_(store, cache))
```

teh-cmc: Uh-oh, nice catch. I meant to add a non-blocking optimistic check that comes first.

But now that I think about it, this is in `QueryHandle`, so realistically nothing will ever contend on this, we can probably just do it all in the OnceLock yeah.

zehiko (approved):

PR #7967, opened 2024-11-01
cmake options

DapengFeng:
add cmake options for examples, tests, and docs for saving compile time.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7967?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7967?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7967)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,9 +170,22 @@ endif()
 # ------------------------------------------------------------------------------
 add_subdirectory(crates/top/rerun_c) # The Rerun C SDK library, must be included before the C++ SDK.
 add_subdirectory(rerun_cpp) # The Rerun C++ SDK library.
-add_subdirectory(examples/cpp)
-add_subdirectory(tests/cpp)
-add_subdirectory(docs/snippets)
+
+option(BUILD_EXAMPLES "Build all examples." OFF)
+option(BUILD_TESTS "Build all tests." OFF)
+option(BUILD_DOCS "Build all documentation." OFF)
+
+if(BUILD_EXAMPLES)
+    add_subdirectory(examples/cpp)
+endif()
+
+if(BUILD_TESTS)
+    add_subdirectory(tests/cpp)
+endif()
+
+if(BUILD_DOCS)
+    add_subdirectory(docs/snippets)
+endif()
 
 # ------------------------------------------------------------------------------
 # Setup an 'all_targets' target that depends on all targets in the project.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,20 +171,21 @@ endif()
 add_subdirectory(crates/top/rerun_c) # The Rerun C SDK library, must be included before the C++ SDK.
 add_subdirectory(rerun_cpp) # The Rerun C++ SDK library.
 
-option(BUILD_EXAMPLES "Build all examples." OFF)
-option(BUILD_TESTS "Build all tests." OFF)
-option(BUILD_DOCS "Build all documentation." OFF)
+option(BULID_SDK_ONLY "Build only the SDK and not the examples, tests, or docs." ON)
+option(BUILD_EXAMPLES "Build all examples." ON)
+option(BUILD_TESTS "Build all tests." ON)
+option(BUILD_DOCS "Build all documentation." ON)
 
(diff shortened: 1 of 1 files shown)


Wumpf:
Thanks for the contribution, but we're handling all these with pixi command aliases, see https://github.com/rerun-io/rerun/blob/main/pixi.toml#L381
As you can see from those it's already fairly easy to select the targets via the `--target` option on CMake which is a lot easier than setting options.

Also, even so I don't think omitting entire directories from CMake is the way to go as it means that the entire CMake configure run becomes invalid, not just the build.

DapengFeng:
@Wumpf How can I choose the build target for other CMake projects? For example, https://rerun.io/docs/getting-started/quick-start/cpp

Wumpf:
Not sure what you mean: in CMake any added [target](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Key%20Concepts.html#targets) can be selected with [`--target`](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-build-t) when using `cmake --build`. Whether that makes sense in the first place is a bit setup dependent: for instance on Windows one might want to create a Visual Studio solution and use that instead for target selection, never actually going through `cmake --build` whereas with a `Ninja` (as done by our pixi config) or `Make`, `cmake --build` is a convenient shortcut

DapengFeng:
Hi @Wumpf.  For example, I try to use `rerun` as 3rdparty in my project.
``` cmake
# ------------------------------------------------------------
# rerun-sdk
# ------------------------------------------------------------
find_package(Python COMPONENTS Interpreter REQUIRED)
# detect virtualenv and set Pip args accordingly
if(DEFINED ENV{VIRTUAL_ENV} OR DEFINED ENV{CONDA_PREFIX})
    set(_pip_args)
else()
    set(_pip_args "--user")
endif()
# install PyPI Python package using pip
execute_process(
    COMMAND ${Python_EXECUTABLE} -c "import rerun_sdk"
    RESULT_VARIABLE EXIT_CODE
    OUTPUT_QUIET
)
if(NOT ${EXIT_CODE} EQUAL 0)
    execute_process(
        COMMAND ${Python_EXECUTABLE} -m pip install ${_pip_args} rerun-sdk
    )
endif()
add_subdirectory(3rdparty/rerun)
```
I compile it with the original options. It may be a waste to rebuild the doc, tests, and examples, although nothing can be changed.
``` bash
 cmake --build .
[0/2] Re-checking globbed directories...
[1/109] Performing update step for 'arrow_cpp'
-- Already at requested tag: apache-arrow-10.0.1
[104/104] Linking CXX executable 3rdparty/rerun/docs/snippets/custom_data
```

Wumpf:
Simply don't do `cmake --build .` but instead pass `--target` with your target name.

This can work fine, but generally the Rerun repo isn't meant to be vendored in like this in its entirety. What you can also try is to only run `add_subdirectory` on `/rerun_cpp` - at least at some point this was standalone enough, but I'm not sure if it still is. The better alternative is to rely on the sdk zip that is part of the release artifact. That is essentially just the sdk parts copied together and tested to be standalone https://github.com/rerun-io/rerun/releases/download/0.19.1/rerun_cpp_sdk.zip
The examples in the docs all use this via dynamic download but ofc you can also unpack it somewhere if you prefer that. What's the reason you want to grab a copy of the entire repository here?

DapengFeng:
> What's the reason you want to grab a copy of the entire repository here?

Live at head:)

Wumpf:
> Live at head:)

cool! let me know how it goes. would be really nice if we could get `add_subdirectory(3rdparty/rerun/rerun_cpp)` to just work (again?). I guess the main problem gonna be that `rerun_c` requires a rust build step which drags in another target that's defined somewhere else again.

Another thing I just found that could work for you is `add_subdirectory(3rdparty/rerun EXCLUDE_FROM_ALL)`, see https://cmake.org/cmake/help/latest/command/add_subdirectory.html

Issue #7973, opened 2024-11-01
Video opened through rerun looks distorted and has wrong colors


pablovela5620:
**Describe the bug**

Something went wrong with showing an av1 encoded video. Here are is what it looks like in rerun and what it looks like in vlc

## Rerun
https://github.com/user-attachments/assets/b0a692c9-2205-46b5-9964-29065db32faa

## VLC
https://github.com/user-attachments/assets/559d01f1-9d9c-42fa-8ad8-46ee273a24d3

**To Reproduce**
1. Open video via `rerun $Path to video`

here is the video that I used that caused problems
https://huggingface.co/datasets/pablovela5620/rerun-bug-video/blob/main/combined-av1.mp4

**Desktop (please complete the following information):**
 - OS: Mac OS 15.1 (24B83)

**Rerun version**

rerun-cli 0.19.0 [rustc 1.79.0 (129f3b996 2024-06-10), LLVM 18.1.7] aarch64-apple-darwin, built 2024-10-21T16:15:46Z
Video features: av1 nasm

**Additional context**
This was after I merged some mp4 files and reenocded them to av1 using the following command
```
ffmpeg -i combined.mp4 -c:v libsvtav1 -preset 10 -crf 35 -c:a copy combined-av1.mp4
```

emilk:
I get the same problem on the native build:
<img width="289" alt="Screenshot 2024-11-02 at 09 16 56" src="https://github.com/user-attachments/assets/7e5bebf3-3cfa-46c4-88ae-8578ae5229a5">

The web-viewer plays it perfectly though

emilk:
Oh shit, I missed this:

<img width="437" alt="Screenshot 2024-11-02 at 09 22 41" src="https://github.com/user-attachments/assets/99908311-57e0-42a2-9324-f29815845f4a">

This is therefore a duplicate of:
* #7594

It makes me think we should just refuse to play such videos though, instead of showing something so obviously wrong

emilk:
I'll add a warning to the selection panel, and maybe see if there is an easy way to at least get _almost_ correct colors

pablovela5620:
For some reason I also didn't see the pop up on my mac, when I tested on my linux machine it showed up!

Issue #7594, opened 2024-10-04
Support HDR video (shows wrong color or glitches today)


Wumpf:
Some of it might be shortcomings in the (web) decoder, but we need to take a more serious look at this. Meanwhile HDR/BT.2020 video should not be supported.

Firefox @ Mac on av1
![image](https://github.com/user-attachments/assets/5b4613dc-1a48-4657-a29f-82c2344bf6c9)

Chrome @ Mac on the same video
![image](https://github.com/user-attachments/assets/873a18bb-9139-4903-81bb-1024aaa43ddb)
(colors are still wrong)

AV1 native:
![image](https://github.com/user-attachments/assets/ad24e955-c583-4826-9979-55439d1831be)

Possibly related to:
* https://issues.chromium.org/issues/40944011

emilk:
Can we detect HDR videos and log a warning about it?

Wumpf:
given that for all these cases I see BT.2020 (instead of (SDR) BT.709) on the "media information" in VLC, I suspect that this is a piece of information provided by mp4 🤔 . In any case, yes there gotta be some way and a warning would be in order!

emilk:
Useful reading:
* https://www.wired.com/story/what-are-color-gamuts-and-why-do-they-matter/
* http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html
* http://www.brucelindbloom.com/index.html?Eqn_ChromAdapt.html

Wumpf:
> the distinction is kinda silly

turns out it is **not**! As demonstrated there's high bit depth Bt709 video material out there! See https://github.com/rerun-io/rerun/issues/7973

mindbeast:
I have tried to display video from my camera in this, and I can't make it work at all. clearly BT2020 colorspaces are not supported, but even if I go back to BT709 SDR, 10 bit video from my camera cannot be displayed. 10 bit != HDR.

PR #7978, opened 2024-11-02
Fix playback of HDR AV1 videos in native viewer

emilk:
### What
* Closes https://github.com/rerun-io/rerun/issues/7973
* Part of https://github.com/rerun-io/rerun/issues/7594

Review with ignoring whitespace changes

#### Before
Screenshot 2024-11-02 at 12 02 13

#### After
Screenshot 2024-11-02 at 12 00 59

#### VLC (for reference)
Screenshot 2024-11-02 at 16 43 10

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7978?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7978?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/7978)

### DIFF
--- a/crates/store/re_video/src/decode/av1.rs
+++ b/crates/store/re_video/src/decode/av1.rs
@@ -133,7 +133,7 @@ fn create_frame(debug_name: &str, picture: &dav1d::Picture) -> Result<Frame> {
         1
     } else if 8 < bits_per_component && bits_per_component <= 16 {
         // TODO(#7594): Support HDR video.
-        // We currnelty handle HDR videos by throwing away the lowest bits,
+        // We currently handle HDR videos by throwing away the lowest bits,
         // and doing so rather slowly on CPU. It works, but the colors won't be perfectly correct.
         re_log::warn_once!(
             "{debug_name:?} is a High-Dynamic-Range (HDR) video with {bits_per_component} bits per component. Rerun does not support this fully. Color accuracy and performance may suffer.",
--- a/crates/store/re_video/src/decode/av1.rs
+++ b/crates/store/re_video/src/decode/av1.rs
@@ -105,7 +105,8 @@ impl SyncDav1dDecoder {
             };
             match picture {
                 Ok(picture) => {
-                    output_picture(&self.debug_name, &picture, on_output);
+                    let frame = create_frame(&self.debug_name, &picture);
+                    on_output(frame);
                     count += 1;
                 }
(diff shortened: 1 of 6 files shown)


crates/store/re_video/src/decode/av1.rs
```diff
@@ -121,98 +122,115 @@ impl SyncDav1dDecoder {
...
+    let bits_per_component = picture
+        .bits_per_component()
+        .map_or(picture.bit_depth(), |bpc| bpc.0);
+
+    let bytes_per_component = if bits_per_component == 8 {
+        1
+    } else if 8 < bits_per_component && bits_per_component <= 16 {
+        // TODO(#7594): Support HDR video.
+        // We currnelty handle HDR videos by throwing away the lowest bits,
+        // and doing so rather slowly on CPU. It works, but the colors won't be perfectly correct.
```
Wumpf, line 137:
```suggestion
        // We currently handle HDR videos by throwing away the lowest bits,
        // and doing so rather slowly on CPU. It works, but the colors won't be perfectly correct.
```

Wumpf (approved):
nice! I trust there's still a separate warning about Bt.2020 interpretation and friends?
After all the problem with typical HDR content is twofold:
* bit depth - ~partially~ better addressed here
* color space conversion

Wumpf:
oh interesting, the video in the issue is unplayable by Firefox. It has 10bit Bt.709 YUV 4:2:0
So indeed really the only thing we don't support here is the 10bits. Fascinating, didn't know that's a thing!

image

emilk:
Bye @Wumpf we are missing a check that the number of bytes in the frame matches the given dimensions and pixel format. A debug-assert and/or warning in re_renderer would be very helpful

Wumpf:
👍 good find, we should definitely add that

emilk:
Actually, we do have a check here that triggers as it should:

https://github.com/rerun-io/rerun/blob/325fc41018c5eb75fbd18d9c84e9648ab88e4d70/crates/viewer/re_renderer/src/resource_managers/image_data_to_texture.rs#L186-L192

My bad!

emilk:
Why does `rerun_js` fail?? O.o

Issue #7607, opened 2024-10-07
Native video decoder via `ffmpeg` CLI


emilk:
* Part of https://github.com/rerun-io/rerun/issues/7298

Send samples to `ffmpeg` over CLI, read back the results, show the user.

A basic demo of this is in this internal repository: https://github.com/rerun-io/video-experiments

The main approach is fairly simple: 

```rust
 let mut ffmpeg = FfmpegCommand::new()
        .hide_banner()
        // Keep in mind that all arguments that are about the input, need to go before!
        .format("h264")
        .input("-")
        //.args(&["-c", "copy", "out.mp4"]) // For testing.
        .rawvideo() // Output rgb24 on stdout. (todo for later: any format we can read directly on re_renderer would be better!)
        .spawn()
        .expect("faild to spawn ffmpeg");

    let mut stdin = ffmpeg.take_stdin().unwrap();
    std::thread::spawn(move || {
          // Wait for samples to be sent on stdin.
          // then send them to ffmpeg.
          // See `write_sample_to_nalu_stream`
    });

    // On the main thread, run the output instance to completion
    ffmpeg.iter().unwrap().for_each(|e| match e {
        FfmpegEvent::Log(LogLevel::Error, e) => println!("Error: {}", e),
        FfmpegEvent::Progress(p) => println!("Progress: {} / 00:00:15", p.time),
        FfmpegEvent::OutputFrame(frame) => println!(
            "Received frame: time {:?} fmt {:?} size {}x{}",
            frame.timestamp, frame.pix_fmt, frame.width, frame.height
        ),
        evt => println!("Event: {evt:?}"),
    });
```

The difficult part is that we need to provide ffmpeg with a format that it can stream in on stdin. In this snippet this is done with `format("h264")`, so ffmpeg expects the `.h264` format which is a stream of NAL units in Annex B format (== replace NAL lengths with start headers). Also, at every IDR frame, sequence parameter sets (SPS) and frame parameter sets (FPS) need to be inserted:
```rust
fn write_sample_to_nalu_stream(
    avc_box: &re_mp4::Avc1Box,
    nalu_stream: &mut dyn std::io::Write,
    sample: &re_mp4::Sample,
    video_track_data: &[u8],
    state: &mut NaluStreamState,
) -> Result<(), Box<dyn std::error::Error>> {
    let avcc = &avc_box.avcc;

    // Append SPS & PPS NAL unit whenever encountering an IDR frame unless the previous frame was an IDR frame.
[...]

PR #8005, opened 2024-11-05
Add warning for using an old version of ffmpeg, be accomodating for log quirks of older versions

Wumpf:
### What

* part of https://github.com/rerun-io/rerun/issues/7607

Most replacing randomness with warnings. I haven't deep dived on which version will work but 5.1 went fine with my test data minus a bit of log spam which I fixed here.

I considered adding a check for coded availability, but this seems a bit cumbersome given that any off-the-shelf ffmpeg install comes with h264: the problem is that while we get a full string of compile options that ffmpeg was built with, we can't really infer the decoders from this since any given library may fulfill various roles. Instead we'd need to parse the output of `ffmpeg -codecs`. We should eventually! But doesn't seem urgent right now.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8005?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8005?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/8005)

### DIFF
--- a/crates/store/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
+++ b/crates/store/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
@@ -388,61 +388,6 @@ fn read_ffmpeg_output(
     pixel_format: &PixelFormat,
     on_output: &Mutex<Option<Arc<OutputCallback>>>,
 ) -> Option<()> {
-    /// Ignore some common output from ffmpeg:
-    fn should_ignore_log_msg(msg: &str) -> bool {
-        let patterns = [
-            "Duration: N/A, bitrate: N/A",
-            "frame=    0 fps=0.0 q=0.0 size=       0kB time=N/A bitrate=N/A speed=N/A",
-            "encoder         : ", // Describes the encoder that was used to encode a video.
-            "Metadata:",
-            "Stream mapping:",
-             // It likes to say this a lot, almost no matter the format.
-            // Some sources say this is more about internal formats, i.e. specific decoders using the wrong values, rather than the cli passed formats.
-            "deprecated pixel format used, make sure you did set range correctly",
-            // Not entirely sure why it tells us this sometimes:
-            // Nowhere in the pipeline do we ask for this conversion, so it must be a transitional format?
-            // This is supported by experimentation yielding that it shows only up when using the `-colorspace` parameter.
-            // (color range and yuvj formats are fine though!)
-            "No accelerated colorspace conversion found from yuv420p to bgr24",
-            // We actually don't even want it to estimate a framerate!
-            "not enough frames to estimate rate",
-            // Similar: we don't want it to be able to estimate any of these things and we set those values explicitly, see invocation.
-            // Observed on Windows FFmpeg 7.1, but not with the same version on Mac with the same video.
-            "Consider increasing the value for the 'analyzeduration' (0) and 'probesize' (32) options",
-            // Size etc. *is* specified in SPS & PPS, unclear why it's missing that.
-            // Observed on Windows FFmpeg 7.1, but not with the same version on Mac with the same video.
-            "Could not find codec parameters for stream 0 (Video: h264, none): unspecified size",
-        ];
-
-        // Why would we get an empty message? Observed on Windows FFmpeg 7.1.
-        if msg.is_empty() {
-            return true;
-        }
-
-        for pattern in patterns {
-            if msg.contains(pattern) {
-                return true;
-            }
-        }
-
-        false
-    }
-
-    fn log_ffmpeg_warning_once(debug_name: &str, msg: &str) {
-        // Make warn_once work on `[swscaler @ 0x148db8000]` style warnings even if the address is different every time.
-        // In older versions of FFmpeg this may happen several times in the same message (happend in 5.1, did not happen in 7.1).
-        let mut msg = msg.to_owned();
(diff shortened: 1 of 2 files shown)


crates/store/re_video/src/decode/ffmpeg.rs
```diff
@@ -349,6 +355,21 @@ fn read_ffmpeg_output(
...
 
+    fn log_ffmpeg_warning_once(debug_name: &str, msg: &str) {
+        // Make warn_once work on `[swscaler @ 0x148db8000]` style warnings even if the address is different every time.
+        // In older versions of FFmpeg this may happen several times in the same message (happend in 5.1, did not happen in 7.1).
+        let mut msg = msg.to_owned();
+        while let Some(pos) = msg.find("[swscaler @ 0x") {
+            msg = [
+                &msg[..pos],
+                &msg[(pos + "[swscaler @ 0x148db8000]".len())..],
+            ]
```
emilk, line 366: We should bounds-check so we don't crash on the message `[swscaler @ 0xCRASH"

crates/store/re_video/src/decode/ffmpeg.rs
```diff
@@ -349,6 +355,21 @@ fn read_ffmpeg_output(
...
+            .join("[swscaler]");
+        }
```
emilk, line 368: Could be a free-standing function with a unit-test, just to clarify what it is doing

crates/store/re_video/src/decode/ffmpeg.rs
```diff
@@ -362,17 +383,9 @@ fn read_ffmpeg_output(
...
-                    // Make warn_once work on `[swscaler @ 0x148db8000]` style warnings even if the address is different every time.
-                    if let Some(pos) = msg.find("[swscaler @ 0x") {
-                        msg = [
-                            &msg[..pos],
-                            &msg[(pos + "[swscaler @ 0x148db8000]".len())..],
-                        ]
-                        .join("[swscaler]");
-                    };
-                    re_log::warn_once!("{debug_name} decoder: {msg}");
+                    log_ffmpeg_warning_once(debug_name, &msg);
```
emilk, line 388: If we had a helper-function this could be:

```rs
let msg = sanitize_message(msg);
if !should_ignore_log_msg(&msg) {
    re_log::warn_once!("{debug_name} decoder: {msg}");
}
```

crates/store/re_video/src/decode/ffmpeg.rs
```diff
@@ -546,7 +560,32 @@ fn read_ffmpeg_output(
...
+                    version_parts
+                        .next()
+                        .and_then(|part| part.parse::<u32>().ok()),
+                ) {
+                    if major < FFMPEG_MINIMUM_VERSION_MAJOR
+                        || (major == FFMPEG_MINIMUM_VERSION_MAJOR
+                            && minor < FFMPEG_MINIMUM_VERSION_MINOR)
+                    {
+                        re_log::warn_once!(
+                            "FFmpeg version is {}. Only versions >= {FFMPEG_MINIMUM_VERSION_MAJOR}.{FFMPEG_MINIMUM_VERSION_MINOR} are officially supported",
```
emilk, line 579:
```suggestion
                            "The found `ffmpeg` binary on your system has version {}. Rerun has only been tested with >= {FFMPEG_MINIMUM_VERSION_MAJOR}.{FFMPEG_MINIMUM_VERSION_MINOR}. Please update your `ffmpeg` version.",
```

crates/store/re_video/src/decode/ffmpeg.rs
```diff
@@ -546,7 +560,32 @@ fn read_ffmpeg_output(
...
+                            ffmpeg_version.version
+                        );
+                    }
+                } else {
+                    re_log::warn_once!(
+                        "Failed to parse FFmpeg version: {}",
+                        ffmpeg_version.raw_log_message
+                    );
```
emilk, line 587: I think at this point we should return an error, so that we show the "Install ffmpeg" stuff in the selection panel, perhaps with a different message

emilk (approved):

crates/store/re_video/src/decode/ffmpeg.rs
emilk: Even better: include the link to the latest ffmpeg version, as provided by `ffmpeg-sidecar`!

Wumpf: Don't want to return an error since chances that things just work are quite high. But yeah let's log an error with url to make it more visible!

emilk:
Btw, it would be super-useful to send an analytics events when:
* ffmpeg is missing
* ffmpeg version can't be parsed
* ffmpeg version is too old

Wumpf:
analytics even on ffmpeg missing might be a bit too spammy, no? We can't assume that a first time user has ffmpeg installed

crates/store/re_video/src/decode/ffmpeg.rs
Wumpf: changed my mind again after seeing that `7.1-essentials_build-www.gyan.dev` is a version string we might get. This is too wild to make it any of this an error, stays a warning
But adding the download link

PR #8027, opened 2024-11-07
Improve the blueprint API for the map view

### LINKED ISSUE #7903: Clean-up the blueprint API for the map view
☝🏻

abey79:
### What

- Fixes #7903

In passing, I noticed we didn't have an easy way to test view blueprint for equality, e.g. to verify the various way they can be constructed. I've added an `__eq__` implementation on both `SpaceView` and `Container` that ignores the `id` field (which is always unique).

- [x] add snippet/screenshot

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8027?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8027?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/8027)


rerun_py/rerun_sdk/rerun/blueprint/api.py
```diff
@@ -91,6 +91,28 @@ def __init__(
         self.defaults = defaults
         self.overrides = overrides
 
+    def __eq__(self, other: Any) -> bool:
+        """
+        Equality implementation, useful for tests.
+
+        This ignores the `id` field, which should always be unique.
```
Wumpf, line 98: it's unique for a reason though - every view instance represents a different view in the viewer. Does it really make sense to call two views "equal" then? 🤔 
I see that this is obviously useful for testing, but it's not quite correct as presented I think

Wumpf (approved):

rerun_py/rerun_sdk/rerun/blueprint/api.py
abey79: You're obviously right. The thinking is that no one else in their right mind would test for view equality. That said, I just had an idea on a possible cheap way to avoid this.

Wumpf (approved):
magic!

PR #8028, opened 2024-11-07
Make it nicer to inspect blueprint store

emilk:
### What
Make the exploration of the blueprint store nicer.
Stuff I wanted to debug my undo-work.

### Checklist

### DIFF
--- a/crates/store/re_types/src/datatypes/uuid_ext.rs
+++ b/crates/store/re_types/src/datatypes/uuid_ext.rs
@@ -16,6 +16,12 @@ impl From<uuid::Uuid> for Uuid {
     }
 }
 
+impl std::fmt::Display for Uuid {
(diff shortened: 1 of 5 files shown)


crates/store/re_types/src/datatypes/uuid_ext.rs
```diff
@@ -16,6 +16,12 @@ impl From<uuid::Uuid> for Uuid {
     }
 }
 
+impl std::fmt::Display for Uuid {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", uuid::Uuid::from(*self))
```
Wumpf, line 21: nit:
```suggestion
        uuid::Uuid::from(*self).fmt(f)
```

Wumpf (approved):
haven't tried it but checks out 👍

Issue #7960, opened 2024-10-31
Video timestamps don't align up correctly


Wumpf:
Both with native & web Chrome this video is always a little bit off in its backed-in and reported timestamps

https://github.com/user-attachments/assets/da61446c-1930-4aec-9a3a-7ad40f0cc69f

This was generated using:
`ffmpeg -i Big_Buck_Bunny_1080_10s_1MB_h264.mp4 -vf drawtext="fontsize=30:rate=60:fontsize=72:text='%{frame_num} %{pts}':fontcolor='white':boxcolor=0x000000AA:box=1:x=860-text_w/2:y=960" timestamped.mp4`

The problem is likely not limited to h264, but it may be until proven otherwise.

PR #8029, opened 2024-11-07
Fix some videos having offseted (incorrect) timestamps

Wumpf:
### What

* fixes https://github.com/rerun-io/rerun/issues/7960

.. and adds a few more "expert" information fields to video blob plus improving the selection panel view overall a bit

Before:
image

After:
image

-> note how the timestamp component and the timestamped etched into the video are now the same.

The problem is that in the presence of B-frames it can happen that presentation timestamps don't start at 0. This is because in MP4 (and other formats?) timestamps can't be negative and the first timestamp may have a decode timestamp of zero but a presentation timestamp some point later. Turns out this happens in a bunch of the example video material I'm using - not too surprising given that ffmpeg's default encoding settings on h264 tend to produce this!
From what I found (see links in comments) this is compensated by players. We do so now on the fly in key locations, preserving the internal PTS/DTS values of the samples. I tried to sharpen the docs where appropriate.

(on a personal note: I don't quite fully follow yet how this makes sense practically given that the first frame should be an IDR frame (it is, right?). But that's 100% what's happening and I found strong indication that players compensate for this just like we do here now)

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8029?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8029?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/8029)

### DIFF
--- a/crates/store/re_video/src/decode/webcodecs.rs
+++ b/crates/store/re_video/src/decode/webcodecs.rs
@@ -10,7 +10,7 @@ use web_sys::{
 use super::{
     AsyncDecoder, Chunk, DecodeHardwareAcceleration, Frame, FrameInfo, OutputCallback, Result,
 };
-use crate::{Config, Time, Timescale};
+use crate::{Config, Time, Timescale, VideoData};
 
 #[derive(Clone)]
 #[repr(transparent)]
@@ -34,6 +34,7 @@ impl std::ops::Deref for WebVideoFrame {
 pub struct WebVideoDecoder {
     video_config: Config,
     timescale: Timescale,
+    minimum_presentation_timestamp: Time,
     decoder: web_sys::VideoDecoder,
     hw_acceleration: DecodeHardwareAcceleration,
     on_output: Arc<OutputCallback>,
@@ -102,17 +103,21 @@ impl Drop for WebVideoDecoder {
 
 impl WebVideoDecoder {
     pub fn new(
-        video_config: Config,
-        timescale: Timescale,
+        video: &VideoData,
         hw_acceleration: DecodeHardwareAcceleration,
         on_output: impl Fn(Result<Frame>) + Send + Sync + 'static,
     ) -> Result<Self, Error> {
         let on_output = Arc::new(on_output);
-        let decoder = init_video_decoder(on_output.clone(), timescale)?;
+        let decoder = init_video_decoder(
+            on_output.clone(),
+            video.timescale,
+            video.sample_statistics.minimum_presentation_timestamp,
+        )?;
 
         Ok(Self {
-            video_config,
-            timescale,
+            video_config: video.config.clone(),
+            timescale: video.timescale,
+            minimum_presentation_timestamp: video.sample_statistics.minimum_presentation_timestamp,
             decoder,
             hw_acceleration,
             on_output,
@@ -133,10 +138,10 @@ impl AsyncDecoder for WebVideoDecoder {
             &data,
             video_chunk
                 .presentation_timestamp
-                .into_micros(self.timescale),
+                .into_micros_since_start(self.timescale, self.minimum_presentation_timestamp),
             type_,
         );
-        web_chunk.set_duration(video_chunk.duration.into_micros(self.timescale));
+        web_chunk.set_duration(video_chunk.duration.duration(self.timescale).as_secs_f64() * 1e6); // as_millis_f64 is experimental Rust api and there's no as_micros_f64
         let web_chunk = EncodedVideoChunk::new(&web_chunk)
             .map_err(|err| Error::CreateChunk(js_error_to_string(&err)))?;
         self.decoder
@@ -154,7 +159,11 @@ impl AsyncDecoder for WebVideoDecoder {
             // At least on Firefox, it can happen that reset on a previous error fails.
             // In that case, start over completely and try again!
             re_log::debug!("Video decoder reset failed, recreating decoder.");
-            self.decoder = init_video_decoder(self.on_output.clone(), self.timescale)?;
+            self.decoder = init_video_decoder(
+                self.on_output.clone(),
+                self.timescale,
+                self.minimum_presentation_timestamp,
+            )?;
         };
 
         self.decoder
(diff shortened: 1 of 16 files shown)


crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -121,13 +121,15 @@ impl VideoPlayer {
...
         video_data: &[u8],
     ) -> Result<VideoFrameTexture, VideoPlayerError> {
-        if presentation_timestamp_s < 0.0 {
+        if time_since_video_start_in_seconds < 0.0 {
             return Err(VideoPlayerError::NegativeTimestamp);
         }
-        let presentation_timestamp = Time::from_secs(presentation_timestamp_s, self.data.timescale);
+        let presentation_timestamp =
+            Time::from_secs(time_since_video_start_in_seconds, self.data.timescale)
+                + self.data.sample_statistics.minimum_presentation_timestamp;
```
emilk, line 132: This feels quite error prone. It would be less error prone if the `minimum_presentation_timestamp` was an argument to `Time::from_secs/to_secs`

crates/store/re_video/src/demux/mod.rs
```diff
@@ -229,17 +274,25 @@ impl VideoData {
...
     }
 
-    /// Determines the presentation timestamps of all frames inside a video, returning raw time values.
+    /// Determines the video timestamps of all frames inside a video, returning raw time values.
     ///
     /// Returned timestamps are in nanoseconds since start and are guaranteed to be monotonically increasing.
+    /// These are *not* necessarily the same as the presentation timestamps, as the returned timestamps are
+    /// normalized respect to the start of the video, see [`Self::minimum_presentation_timestamp`].
     pub fn frame_timestamps_ns(&self) -> impl Iterator<Item = i64> + '_ {
+        re_tracing::profile_function!();
```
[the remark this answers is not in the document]
Wumpf: oops

crates/viewer/re_renderer/src/video/player.rs
Wumpf: good idea

Wumpf: then again. Argh. It's so poorly defined.

Wumpf: there.
https://github.com/rerun-io/rerun/pull/8029/commits/8ae876be380536fb23550a948d71f78e2d9d9b4c
not thrilled :/

Wumpf: ah well now webcodecs is failing which _does_ highlight that I have to think that one through as well.....

Wumpf:
just noticed a bug here
image

crates/store/re_video/src/decode/webcodecs.rs
```diff
@@ -133,10 +138,10 @@ impl AsyncDecoder for WebVideoDecoder {
             &data,
             video_chunk
                 .presentation_timestamp
-                .into_micros(self.timescale),
+                .into_micros_since_start(self.timescale, self.minimum_presentation_timestamp),
             type_,
         );
-        web_chunk.set_duration(video_chunk.duration.into_micros(self.timescale));
+        web_chunk.set_duration(video_chunk.duration.duration(self.timescale).as_secs_f64() * 1e6); // as_millis_f64 is experimental Rust api and there's no as_micros_f64
```
emilk, line 144: Alternative
```suggestion
        let duration_millis = 1e-3 * video_chunk.duration.duration(self.timescale).as_nanos() as f64;
        web_chunk.set_duration(duration_millis);
```

crates/store/re_video/src/demux/mod.rs
```diff
@@ -60,12 +60,57 @@ pub struct VideoData {
...
+
     /// All the tracks in the mp4; not just the video track.
     ///
     /// Can be nice to show in a UI.
     pub mp4_tracks: BTreeMap<TrackId, Option<TrackKind>>,
 }
 
+/// Meta informationa about the video samples.
+#[derive(Clone, Debug)]
+pub struct SampleStatistics {
```
emilk, line 74: This to me sounds like the statistics of a single sample

crates/store/re_video/src/time.rs
```diff
@@ -28,50 +28,67 @@ impl Time {
...
+        timescale: Timescale,
+        start_time: Self,
+    ) -> Self {
+        Self::from_secs_since_start(micros_since_start / 1e6, timescale, start_time)
     }
 
     #[inline]
-    pub fn from_nanos(v: i64, timescale: Timescale) -> Self {
-        Self::from_secs(v as f64 / 1e9, timescale)
+    pub fn from_nanos_since_start(
```
emilk, line 60: Maybe a single `fn from_duration_sicne_start(duration: std::time::Duration, …)` could replace all these, to produce a more minimal interface

crates/store/re_video/src/time.rs
```diff
@@ -28,50 +28,67 @@ impl Time {
...
-    pub fn into_micros(self, timescale: Timescale) -> f64 {
-        self.into_secs(timescale) * 1e6
+    pub fn into_micros_since_start(self, timescale: Timescale, start_time: Self) -> f64 {
+        self.into_secs_since_start(timescale, start_time) * 1e6
     }
 
     #[inline]
-    pub fn into_nanos(self, timescale: Timescale) -> i64 {
-        (self.into_secs(timescale) * 1e9).round() as i64
+    pub fn into_nanos_since_start(self, timescale: Timescale, start_time: Self) -> i64 {
```
emilk, line 90: Same here; these could be `into_duration_since_start(self, …) -> std::time::Duration`

crates/store/re_video/src/time.rs
```diff
@@ -28,50 +28,67 @@ impl Time {
...
+        nanos_since_start: i64,
+        timescale: Timescale,
+        start_time: Self,
+    ) -> Self {
+        Self::from_secs_since_start(nanos_since_start as f64 / 1e9, timescale, start_time)
     }
 
     /// Convert to a duration
     #[inline]
     pub fn duration(self, timescale: Timescale) -> std::time::Duration {
```
emilk, line 70: It seems like it is time to split `Time` into two types; one for time-points and one for durations. Both require `timescale`, but only one require the `start_time` offset.

emilk (approved):

crates/store/re_video/src/demux/mod.rs
Wumpf: `SamplesStatistics`? Dunno what to call it

crates/store/re_video/src/time.rs
Wumpf: let's not explode and delay this PR even further with cosmetics like that. The time type was never meant to be all that complex, it's just a typed "this is a video time that uses a video time scale"

crates/store/re_video/src/time.rs
Wumpf: still have to keep a nanos version one because std::time::Duration has imho waaaay too much overhead for converting to our `VideoTimestamp` (we expect to converts hundereds of thousands in some cases)

Wumpf: the other nastiness is that duration _panics_ when using negative values.

Actually that last one compells me strongly to roll back what I did. std::time::duration is not safe when dealing with user data

Wumpf:
Not going to follow through on std::time::Duration suggestions: It is too error prone since its constructors panic on negative values, causing us to do lots of extra checking in various places.
Also, specifically for conversion to `VideoTimestamp` it would be a costly detour since it uses a u128+u64 to represent its time.

PR #8045, opened 2024-11-08
gRPC spec update: Introduce ``UpdateRecordingMetadata`` API

zehiko:
### What

Introduces ``UpdateRecordingMetadata`` API.  For now we start simple - we can only update a single recording metadata and the caller needs to provide ``RecordingId`` i.e. we don't support update based on a filter.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8045?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8045?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/8045)

### DIFF
--- a/crates/store/re_protos/proto/rerun/v0/remote_store.proto
+++ b/crates/store/re_protos/proto/rerun/v0/remote_store.proto
@@ -5,10 +5,14 @@ package rerun.remote_store.v0;
 import "rerun/v0/common.proto";
 
 service StorageNode {
-    rpc ListRecordings(ListRecordingsRequest) returns (ListRecordingsResponse) {}
+    // data API calls
     rpc Query(QueryRequest) returns (stream QueryResponse) {}
     rpc FetchRecording(FetchRecordingRequest) returns (stream FetchRecordingResponse) {}
+
+    // metadata API calls
+    rpc ListRecordings(ListRecordingsRequest) returns (ListRecordingsResponse) {}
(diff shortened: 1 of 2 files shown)


crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -67,6 +71,18 @@ message TimeMetadata {
...
+// ---------------- UpdateRecordingMetadata  -----------------
+
+message UpdateRecordingMetadataRequest {
+    RecordingId recording_id = 1;
+    RecordingMetadata metadata = 2;
+}
+
+message UpdateRecordingMetadataResponse {
+    RecordingId id = 1;
+    RecordingMetadata metadata = 2;
```
jleibs, line 83: Is there a reason we need to return this? Seems like all we should need here is a success/fail.

zehiko: I was kinda looking at parity with register, but indeed that has different semantics, so it can be an empty response. I thought it might be useful for a user to now again see the fully updated recording, but it you think there's no value - easy to drop!

jleibs (approved):

PR #8049, opened 2024-11-08
gRPC spec update: Add support for selecting specific columns while listing recordings metadata

zehiko:
### What

Providing a selection of columns is optional, but if provided, it's expected that the remote store will only project and return those columns.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8049?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8049?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/8049)


rerun_py/src/remote.rs
```diff
@@ -54,7 +54,7 @@ impl PyConnection {
     /// List all recordings registered with the node.
     fn list_recordings(&mut self) -> PyResult<Vec<PyRecordingMetadata>> {
         self.runtime.block_on(async {
-            let request = ListRecordingsRequest {};
+            let request = ListRecordingsRequest { columns: None };
```
jleibs, line 57: The usage of None here is unfortunate. Maybe we should call this `column_projection`?

zehiko: I went back and forth between "list of strings, empty list means everything" and "optional selection, none mean everything", I thought the latter would be more explicit. I really like ``column_project`` and how explicit it becomes with it! Will go with that if you don't object

jleibs (approved):

Issue #7956, opened 2024-10-31
Stepping back doesn't always enqueue frames to the decoder (in the presence of b-frames?)


Wumpf:
When stepping back frame by frame we sometimes don't seem to feed the decoder (our current strategy requires restarting the decoder for every step back).

This likely happens only on bframes where we then incorrectly assume that all the samples were already fed into the decoder (this is an educated guess, the lack of feeding I checked empirically). This was observed both with the upcoming ffmpeg decoder and 

Repro'ed reliably with this video

https://github.com/user-attachments/assets/3ef23bd1-af7b-46b9-94fc-133f203fa1bc

on Chrome & native.
(due to https://github.com/rerun-io/rerun/issues/7961 I can't observe this in other browsers)

Issue #7961, opened 2024-10-31
Safari & Firefox skip video frames (are b-frames ignored?)


Wumpf:
B-frames being ignored is the likely cause, but we should dig a bit deeper here to figure this out - would be great if we could show in the ui whether a frame is a bframe. On native h.264 we now show composition & decode timestamps which gets us _almost_ there.

The issue can be reproed by loading this video into Rerun

https://github.com/user-attachments/assets/da61446c-1930-4aec-9a3a-7ad40f0cc69f

and then frame stepping.
I observed this on Mac Firefox, Safari but not in our native ffmpeg based impl or Chrome.

This is very likely the issue this users hit
https://x.com/chrisoffner3d/status/1847623631678943416

PR #8053, opened 2024-11-08
Fix video backward seeking / stepping back sometimes getting stuck (in the presence of b-frames)

Wumpf:
### What

* Fixes https://github.com/rerun-io/rerun/issues/7956

Our PTS -> sample search algorithm wasn't quite right. Fixed this here and added a unit test.
Unfortunately, I had to introduce a small auxiliary data structure to keep this snappy / not O(n) for n samples, but it's very simple and rather contained.

Tested on native & web. Web still suffers from
* https://github.com/rerun-io/rerun/issues/7961

but on Chrome this works fine now!

https://github.com/user-attachments/assets/7a20467d-4d66-456f-be99-debba4ff7a31

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8053?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8053?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/8053)

### DIFF
--- a/crates/store/re_video/src/demux/mod.rs
+++ b/crates/store/re_video/src/demux/mod.rs
@@ -7,8 +7,9 @@
 
 pub mod mp4;
 
-use std::{collections::BTreeMap, ops::Range, usize};
+use std::{collections::BTreeMap, ops::Range};
 
+use bit_vec::BitVec;
 use itertools::Itertools as _;
 
 use super::{Time, Timescale};
@@ -91,9 +92,9 @@ pub struct SamplesStatistics {
     pub dts_always_equal_pts: bool,
 
     /// If `dts_always_equal_pts` is false, then this gives for each sample whether its PTS is the highest seen so far.
-    /// If `dts_always_equal_pts` is true, then this is left empty.
+    /// If `dts_always_equal_pts` is true, then this is left as `None`.
     /// This is used for optimizing PTS search.
-    pub has_sample_highest_pts_so_far: Vec<bool>,
+    pub has_sample_highest_pts_so_far: Option<BitVec>,
 }
 
 impl SamplesStatistics {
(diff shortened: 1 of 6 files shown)


crates/store/re_video/src/demux/mod.rs
```diff
@@ -89,6 +89,11 @@ pub struct SamplesStatistics {
     ///
     /// If true, the video typically has no B-frames as those require frame reordering.
     pub dts_always_equal_pts: bool,
+
+    /// If `dts_always_equal_pts` is false, then this gives for each sample whether its PTS is the highest seen so far.
+    /// If `dts_always_equal_pts` is true, then this is left empty.
+    /// This is used for optimizing PTS search.
+    pub has_sample_highest_pts_so_far: Vec<bool>,
```
jprochazk, line 96: Would it make sense to use `bit-vec` instead here?

crates/store/re_video/src/demux/mod.rs
```diff
@@ -104,9 +109,23 @@ impl SamplesStatistics {
             .iter()
             .all(|s| s.decode_timestamp == s.presentation_timestamp);
 
+        let mut biggest_pts_so_far = Time::MIN;
+        let has_sample_highest_pts_so_far = samples
```
jprochazk, line 113: I'm also a bit worried about how many times we're iterating over the samples here. I don't think LLVM will optimize this to one loop.

jprochazk (approved):
Other than nits above this LGTM

crates/store/re_video/src/demux/mod.rs
Wumpf: ah, yes good call!

crates/store/re_video/src/demux/mod.rs
Wumpf: tbh I'm not sure if it's better to combine everything to a single loop or to loop several times: individual loops might vectorize better

pure speculation until someone actually measures. Maybe I should. After all we might run this with _a lot_ of samples!

jprochazk: It optimizes a lot better than I thought: https://godbolt.org/z/n3Yvzff9e

Wumpf: neither is in the mood for simd though. probably because everything is 64bit integer

Wumpf: leaving it be as is for now

Wumpf:
known issue on `error[unmaintained]: `instant` is unmaintained`

PR #8060, opened 2024-11-11
Enable local http cache for the map view (native only)

abey79:
### What

☝🏻 

This pr introduce a cache directory in `re_viewer_context::AppOptions`, which is cleared by `rerun reset`. With some caveat, see:
- https://github.com/rerun-io/rerun/issues/8064

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8060?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8060?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/8060)

### DIFF
--- a/crates/viewer/re_space_view_map/src/map_space_view.rs
+++ b/crates/viewer/re_space_view_map/src/map_space_view.rs
@@ -479,21 +479,17 @@ fn handle_ui_interactions(
 /// Return http options for tile downloads.
 ///
 /// On native targets, it configures a cache directory.
-fn http_options() -> walkers::HttpOptions {
-    if cfg!(target_arch = "wasm32") {
-        Default::default()
-    } else {
-        let cache = directories::ProjectDirs::from("io", "rerun", "Rerun").map(|dirs| {
-            let mut path = dirs.cache_dir().to_owned();
-            path.push("map_view_cache");
-            path
-        });
+fn http_options(_ctx: &ViewerContext<'_>) -> walkers::HttpOptions {
+    #[cfg(not(target_arch = "wasm32"))]
+    let options = walkers::HttpOptions {
+        cache: _ctx.app_options.cache_subdirectory("map_view"),
+        ..Default::default()
+    };
 
-        walkers::HttpOptions {
-            cache,
-            ..Default::default()
-        }
-    }
+    #[cfg(target_arch = "wasm32")]
+    let options = Default::default();
+
+    options
 }
 
 fn get_tile_manager(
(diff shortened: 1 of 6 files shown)


crates/viewer/re_space_view_map/src/map_space_view.rs
```diff
@@ -476,39 +476,64 @@ fn handle_ui_interactions(
...
 }
 
+/// Return http options for tile downloads.
+///
+/// On native targets, it configures a cache directory.
+fn http_options() -> walkers::HttpOptions {
+    if cfg!(target_arch = "wasm32") {
+        Default::default()
+    } else {
+        let cache = directories::ProjectDirs::from("io", "rerun", "Rerun").map(|dirs| {
```
emilk, line 486: Why is different from https://github.com/rerun-io/rerun/blob/2de386bfcb0b0b34b7b97467c3813f7be9b20645/crates/utils/re_analytics/src/native/config.rs#L127 ?

emilk: I think we should be using https://docs.rs/eframe/latest/eframe/fn.storage_dir.html instead.

Though it would be unfortunate to depend on `eframe` here, perhaps we could pass it by argument (from `ViewerContext`), so that all our on-disk-caches (I'm sure there will be more) use the same storage place.

emilk (changes requested):
Let's not pick a magic location for disk caches, but instead use one mandated by `ViewerContext`.

We should also make sure that running `rerun reset` or the `ResetViewer` command (inside of Rerun) resets this cache.

crates/viewer/re_viewer/src/lib.rs
```diff
@@ -287,6 +287,18 @@ pub fn reset_viewer_persistence() -> anyhow::Result<()> {
         } else {
             re_log::info!("Rerun state was already cleared.");
         }
+
+        // Clear the default cache directory if it exists
+        //TODO(#8064): should clear the _actual_ cache directory, not the default one
+        if let Some(cache_dir) = re_viewer_context::AppOptions::default_cache_directory() {
+            if cache_dir.exists() {
+                if let Err(err) = std::fs::remove_dir_all(&cache_dir) {
```
emilk, line 295: This is a race.

Better to always call `std::fs::remove_dir_all` and then ignore [io::ErrorKind::NotFound](https://doc.rust-lang.org/beta/std/io/enum.ErrorKind.html#variant.NotFound)

emilk (approved):

Issue #7191, opened 2024-08-15
LineStrips2D radii not working as expected for real world data


Famok:
**Describe the bug**
The radii of Linestrips2D does not what I expect it to do, which is to increase the line width.
Instead it only increases the width of each segment, which will lead to a weird behaviour if you have real world data (e.g. with noise).

**To Reproduce**
Code to show the issue:
```
import rerun as rr
import numpy as np

x = np.linspace(0,2*np.pi,1000)
y = np.random.rand(len(x))*0.1

xy = np.array([x,y]).T

rr.init('testing', recording_id= ' test', spawn = True)
for r in [0.001, 1]:
    rr.log(
        f"random line {r=}",
        rr.LineStrips2D(
            [xy],
            radii=r
        ),
    )

```

**Expected behavior**
I was expecting the line as a whole to get wider (similar how this works in powerpoint for example).

**Desktop (please complete the following information):**
Windows 11 23H2

**Rerun version**
rerun_py 0.18.0-alpha.1+dev [rustc 1.76.0 (07dca489a 2024-02-04), LLVM 17.0.6] x86_64-pc-windows-msvc main 9e50d2c, built 2024-08-14T13:15:39Z

nikolausWest:
Thanks for the issue @Famok!

Could you add some screenshots comparing what you're seeing in Rerun vs the behavior you're expecting (powerpoint)? We're having a little trouble interpreting exactly what you mean

Wumpf:
took me a bit indeed but I think I see what you mean now, the line radius behaves strangely when it goes about individual line lengths:

https://github.com/user-attachments/assets/c9d54e35-a0d4-4321-88a0-b425f6891cc8

Seems to be closely related to:
* https://github.com/rerun-io/rerun/issues/829

abey79:
Another depiction. Both are the same line. The yellow one has _very_ thick segments. The "effect" is due to the lack of (miter/round) join handling.

<img width="1729" alt="image" src="https://github.com/user-attachments/assets/f96ab7a6-0e80-48e5-8992-c82f35b1f4dc">

Famok:
@abey79 this is what I mean. For some reason I'm unable to upload  images myself so I could show how my expected behavior looks like.

Issue #829, opened 2023-01-19
Add round joints to line strips


Wumpf:
re_renderer doesn't create line miters so far!

- Fixes #7191 
- Related #6180

PR #8065, opened 2024-11-11
Line strips are no longer a disconnected series of quads

Wumpf:
### What

* Fixes https://github.com/rerun-io/rerun/issues/829
* Fixes https://github.com/rerun-io/rerun/issues/7191

Fixes a long standing issue that the segments of a line have small gaps between them.

Other than originally planned, the solution employed here is entirely based on fragment shader discard + elongation of the line. This can cause some slight artifacts for 3D lines but it's quite hard to spot those issues. Otherwise the biggest drawback of this is that we have more overdraw which gonna be tricky to deal with once we add transparency. On the flip side this is fairly simple and can re-use some of the round cap handling code - round caps no longer use the "trailing triangle" we had for every strip internally, and instead use the quad extension, actually leading to better looking round caps.

⚠️  To allow loops to "just work", all line segments & line strips start and end now with extended round caps. I.e. if your line has a radius of 1 and goes from (0,0,0) to (100,0,0), then what's drawn goes from (-1,0,0) to (101,0,0). Before we didn't have caps on this, so we essentially drew a square starting at (0,0,0) whereas now it's more capsule like (it does not exactly behave like a capsule under perspective though).

Line loops on map view:
Before:
image

After:
image

RRT Star:
Before:
image

After:
image

3D line snippet snippet:
Before:
image

After:
image

Failure Example in #7191
Before - confusingly broken:
image

After - pretty bad rendering performance due to quite massive overdraw of alpha tested geometry, but no longer broken.
image

(After: setting the radius to something large, but sane)
image

re_renderer 2D demo. Left before, right after. Note that "shaded lines" look a lot better now as a sideffect, with shading of arrows slightly broken. Afaik we don't use this anywhere in Rerun today though. 
However, rounded caps are higher quality now in general.

Screenshot 2024-11-11 at 11 10 29

----

Things that were several segments before can now be strips. No visual change there:

Rectangles:
image

Frustum:
image

Box wireframes:

This fixes a regression! In 0.19 those boxes had disconnected lines, in 0.18 this still worked since they were separate segments with rounded corners.

Before (regression):
image

After:
image

----

Due to aforementioned always-round-caps, a list of segments can now look indistinguishable from a single strip:

image

----

Under more extreme circumstances it is still (!) possible to have boxes exhibit some artifacts:

image

This is due to the quasi-capsule nature of our line segments. Calculating a world position from ray-capsule intersection would yield the correct result. What's actually happening right now is that we interpolate the world position over the spanned quad (which is a lot cheaper).

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8065?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8065?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/8065)

### Testing
* [x] Tested on WebGL
* [x] Tested on WebGPU

### DIFF
--- a/crates/viewer/re_renderer/shader/composite.wgsl
+++ b/crates/viewer/re_renderer/shader/composite.wgsl
@@ -7,6 +7,7 @@ struct CompositeUniformBuffer {
     outline_color_layer_a: vec4f,
     outline_color_layer_b: vec4f,
     outline_radius_pixel: f32,
+    blend_with_background: u32,
 };
 @group(1) @binding(0)
 var<uniform> uniforms: CompositeUniformBuffer;
@@ -27,11 +28,17 @@ fn main(in: FragmentInput) -> @location(0) vec4f {
     // but are about the location of the texel in the target texture.
     var color = textureSample(color_texture, nearest_sampler, in.texcoord);
 
+
     // TODO(andreas): We assume that the color from the texture does *not* have pre-multiplied alpha.
     // This is a brittle workaround for the alpha-to-coverage issue described in `ViewBuilder::MAIN_TARGET_ALPHA_TO_COVERAGE_COLOR_STATE`:
     // We need this because otherwise the feathered edges of alpha-to-coverage would be overly bright, as after
     // MSAA-resolve they end up with an unusually low alpha value relative to the color value.
-    color = vec4f(color.rgb * color.a, color.a);
+    if uniforms.blend_with_background == 0 {
+        // To not apply this hack needlessly and account for alpha from alpha to coverage, we have to ignore alpha values if blending is disabled.
+        color = vec4f(color.rgb, 1.0);
+    } else {
+        color = vec4f(color.rgb * color.a, color.a);
+    }
 
     // Outlines
     {
--- a/crates/viewer/re_renderer/shader/lines.wgsl
+++ b/crates/viewer/re_renderer/shader/lines.wgsl
@@ -47,8 +47,8 @@ const FLAG_CAP_START_EXTEND_OUTWARDS: u32 = 32u;
 const FLAG_COLOR_GRADIENT: u32 = 64u;
 const FLAG_FORCE_ORTHO_SPANNING: u32 = 128u;
 
-// Additional flags computed by the shader.
-const FLAG_ROUND_JOINT: u32 = 1024u;
+// Special flags used in the fragment shader.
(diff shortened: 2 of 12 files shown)


crates/viewer/re_renderer/shader/lines.wgsl
```diff
@@ -47,6 +47,9 @@ const FLAG_CAP_START_EXTEND_OUTWARDS: u32 = 32u;
 const FLAG_COLOR_GRADIENT: u32 = 64u;
```
Wumpf, line 47: all the really important changes are in this file

crates/viewer/re_renderer/shader/composite.wgsl
```diff
@@ -27,11 +28,17 @@ fn main(in: FragmentInput) -> @location(0) vec4f {
...
     // This is a brittle workaround for the alpha-to-coverage issue described in `ViewBuilder::MAIN_TARGET_ALPHA_TO_COVERAGE_COLOR_STATE`:
     // We need this because otherwise the feathered edges of alpha-to-coverage would be overly bright, as after
     // MSAA-resolve they end up with an unusually low alpha value relative to the color value.
-    color = vec4f(color.rgb * color.a, color.a);
+    if uniforms.blend_with_background == 0 {
+        // To not apply this hack needlessly and account for alpha from alpha to coverage, we have to ignore alpha values if blending is disabled.
+        color = vec4f(color.rgb, 1.0);
+    } else {
+        color = vec4f(color.rgb * color.a, color.a);
+    }
```
Wumpf, line 41: improves artifact suppression. In other words this gets non-map views to where they were in 0.19
Tbh. I'm not 100% sure why I did get still artifacts for boxes in 3D without this, but it also didn't feel _that_ surprising.

Wumpf:
@emilk I should have formulated that more clearly: the _shading_ specifically is slightly broken (see pic). But we don't have the shading ever enabled in Rerun, we made that an opt-in operation unlike on points

abey79 (approved):
Very nice! Can't say I follow all details in the shader code, but I tested it a bit it seems to work great.

emilk:
Make sure to test this on WebGL _and_ WebGPU!

Wumpf:
> Make sure to test this on WebGL and WebGPU!

Did! On the Mac only mostly though

Issue #8072, opened 2024-11-11
"Current decoded frame" shows garbage data for frames that aren't decoded yet


Wumpf:
![image](https://github.com/user-attachments/assets/d08c547c-26d4-4899-9d03-98909a6bfadc)

PR #8099, opened 2024-11-12
Improve `FrameInfo` handling

emilk:
### What
Added `is_sync` and make sure `FrameInfo` is either valid or `None`.

### Checklist

### DIFF
--- a/crates/store/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
+++ b/crates/store/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
@@ -95,8 +95,8 @@ impl From<Error> for crate::decode::Error {
 struct FfmpegFrameInfo {
     /// The start of a new [`crate::demux::GroupOfPictures`]?
     ///
-    /// If true, an entire frame can be decoded from this one sample,
-    /// otherwise it needs the context of other samples.
+    /// This probably means this is a _keyframe_, and that and entire frame
+    /// can be decoded from only this one sample (though I'm not 100% sure).
     is_sync: bool,
     presentation_timestamp: Time,
     duration: Time,
--- a/crates/store/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
+++ b/crates/store/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
@@ -93,6 +93,11 @@ impl From<Error> for crate::decode::Error {
 /// ffmpeg does not tell us the timestamp/duration of a given frame, so we need to remember it.
 #[derive(Clone)]
 struct FfmpegFrameInfo {
+    /// The start of a new [`crate::demux::GroupOfPictures`]?
+    ///
+    /// If true, an entire frame can be decoded from this one sample,
+    /// otherwise it needs the context of other samples.
+    is_sync: bool,
     presentation_timestamp: Time,
     duration: Time,
     decode_timestamp: Time,
@@ -575,6 +580,7 @@ fn read_ffmpeg_output(
                         format: pixel_format.clone(),
                     },
                     info: FrameInfo {
+                        is_sync: Some(frame_info.is_sync),
                         presentation_timestamp: frame_info.presentation_timestamp,
                         duration: frame_info.duration,
                         latest_decode_timestamp: Some(frame_info.decode_timestamp),
@@ -690,6 +696,7 @@ impl AsyncDecoder for FfmpegCliH264Decoder {
         // We send the information about this chunk first.
         // Chunks are defined to always yield a single frame.
         let frame_info = FfmpegFrameInfo {
+            is_sync: chunk.is_sync,
             presentation_timestamp: chunk.presentation_timestamp,
             decode_timestamp: chunk.decode_timestamp,
             duration: chunk.duration,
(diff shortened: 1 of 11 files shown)


crates/store/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
```diff
@@ -93,6 +93,11 @@ impl From<Error> for crate::decode::Error {
 /// ffmpeg does not tell us the timestamp/duration of a given frame, so we need to remember it.
 #[derive(Clone)]
 struct FfmpegFrameInfo {
+    /// The start of a new [`crate::demux::GroupOfPictures`]?
+    ///
+    /// If true, an entire frame can be decoded from this one sample,
```
Wumpf, line 98: I thought for very long that this was true. But now I'm not so sure anymore that we are guaranteed to get a frame back from a decoder when we enqueue such a frame.
I would have expected PTS==DTS on those, but that part I know for sure now is not the case.

One subtle problem with this statement that I'm a bit more sure about is that while we have a `sample -> frame`, it may still be that a frame needs meta information from other samples even though this is supposed to be an IDR frame.
For instance in AV1, the information we append right now in our annex_b stream in the ffmpeg decoder is dispersed over the samples of the video and I have no ideas if there's any guarantees at all.


With that in mind I'd be in favor for leaving this out of the doc comment & tooltip (affects a bunch of places where this is copy pasted) until we understand better what we're dealing with.

crates/store/re_video/src/decode/mod.rs
```diff
@@ -250,17 +253,18 @@ pub type FrameContent = webcodecs::WebVideoFrame;
...
+    pub is_sync: Option<bool>,
+
+    /// The presentation timestamp of the frame.
     pub presentation_timestamp: Time,
 
     /// How long the frame is valid.
-    ///
-    /// Decoders are required to report this.
-    /// A duration of [`Time::MAX`] indicates that the frame is invalid or not yet available.
-    // Implementation note: unlike with presentation timestamp we may be able fine with making this optional.
```
Wumpf, line 263: the time::max thing is no longer relevant, but the rest you also didn't like?

crates/utils/re_log/src/setup.rs
```diff
@@ -39,7 +39,8 @@ pub fn setup_logging() {
                 use std::io::Write as _;
                 writeln!(
                     buf,
-                    "{}:{} {}",
+                    "{} {}:{} {}",
+                    record.level(),
```
Wumpf, line 43: what is this? I'm lacking context

crates/viewer/re_ui/src/list_item/property_content.rs
```diff
@@ -152,9 +152,14 @@ impl<'a> PropertyContent<'a> {
     /// Show a read-only boolean in the value column.
     #[inline]
     pub fn value_bool(self, mut b: bool) -> Self {
-        self.value_fn(move |ui: &mut Ui, _| {
-            ui.add_enabled_ui(false, |ui| ui.re_checkbox(&mut b, ""));
-        })
+        if true {
+            self.value_text(b.to_string())
+        } else {
+            // This is not readable, which is why it is disabled
```
Wumpf, line 158: have you checked whether there are situations where value_bool would have looked better like that? Don't know if there are callsites where this is a standin for a disabled editable property 🤔

Wumpf (approved):
Fixes https://github.com/rerun-io/rerun/issues/8072 if understand correctly!

crates/viewer/re_ui/src/list_item/property_content.rs
emilk: I checked all calls of `value_bool`. There was only one.

crates/utils/re_log/src/setup.rs
emilk: context

crates/store/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
emilk: FWIW, this is what Claude 3.5 believes:

image

…but I'm sure it's mostly trained on common misconceptions

Wumpf: not a strong conviction here as described.
What bugs me about all this is that the `is_sync` is derived in quite strange way in the mp4 crate which I don't follow and I'd rather have per codec defined things that are easier to look up :(

Wumpf: we obviously depend on it meaning _something_ though as it governs our enqueuing strategy

Wumpf: maybe just slap a `We generally assume` in front of it 🤷

crates/utils/re_log/src/setup.rs
emilk: I like to flip that switch so I can see where a log message comes from.
Before I couldn't see the log level when I did so. Now I can.

image

emilk: It was completely unrelated to the PR, except I wanted to debug "what the hell is wring that annoying log message"

Issue #8101, opened 2024-11-12
Chunk ingestion performance regression because of compaction logic


rgolovanov:
**Describe the bug**
The latency in the Viewer is rapidly growing when dozens of scalar values are logged. The performance has degraded between 0.17.0 and 0.19.0 (or maybe even 0.18.1)

**To Reproduce**
1. `pip install rerun-sdk==0.17.0`
2. `python rerun_latency_test.py` (see code below)
3. Check latency in performance metrics in Viewer
4. Repeat steps (1) - (3) for version 0.19.0 and see how quickly the latency goes up and above a few seconds.

**Expected behavior**
The latency stays within 10-15ms range as it was in earlier versions.

**Screenshots**
![image](https://github.com/user-attachments/assets/f65c5123-0cfc-4978-a376-05caea51f548)

**Test script `rerun_latency_test.py`**

probably, numbers can be different depending on the machine, but in my case the problem is visible when I log 10x5x3=150 scalar values.

```
import rerun as rr
import time
import random
import threading
import itertools

def log_scalar_values():
    rr.init(f"TestLatency", spawn=True, recording_id="default")
    while not stop_event.is_set():
        for i, j, k in itertools.product(range(0, 10), range(0, 5), range(0, 3)):
            scalar_value = random.uniform(0, 1)  # Generate a random scalar value
            rr.log(f"AAA/{i}/{j}/{k}/value", rr.Scalar(scalar_value))
            print(f"Log {scalar_value}")
        time.sleep(0.001)  # Sleep for 10ms to achieve approximately 100Hz


stop_event = threading.Event()

logging_thread = threading.Thread(target=log_scalar_values)
logging_thread.start()

input("Press Enter to stop logging...\n")
stop_event.set()
logging_thread.join()
print("Logging stopped.")
```

**Desktop (please complete the following information):**
 - OS: Windows 11

**Rerun version**
Detected in 0.19.0 but likely appeared in 0.18.0 or 0.18.1

teh-cmc:
Thanks for the neat little repro @rgolovanov, that made it a breeze :+1:

rgolovanov:
@teh-cmc Thanks for the extremely fast fix 🙏 We will be looking forward for new release 😊

rgolovanov:
[...]

PR #8103, opened 2024-11-12
`EntityTree`: only check for entity deletions when necessary

teh-cmc:
* Fixes https://github.com/rerun-io/rerun/issues/8101

Before:
image

After:
image

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8103?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8103?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/8103)

### DIFF
--- a/crates/store/re_entity_db/src/entity_tree.rs
+++ b/crates/store/re_entity_db/src/entity_tree.rs
@@ -127,7 +127,7 @@ impl EntityTree {
     ///
     /// Checking for the absence of data is neither costly nor totally free: do it a few hundreds or
     /// thousands times a frame and it will absolutely kill framerate.
-    /// Don't blindly call this on every existing entity every frame: use [`StoreEvent`]s to make
+    /// Don't blindly call this on every existing entity every frame: use [`ChunkStoreEvent`]s to make
     /// sure anything changed at all first.
     pub fn check_is_empty(&self, engine: &StorageEngineReadGuard<'_>) -> bool {
         self.children.is_empty() && !engine.store().entity_has_data(&self.path)


crates/store/re_entity_db/src/entity_tree.rs
```diff
@@ -171,9 +172,11 @@ impl EntityTree {
 
         self.children.retain(|_, entity| {
             // this is placed first, because we'll only know if the child entity is empty after telling it to clear itself.
-            entity.on_store_deletions(engine, events);
+            entity.on_store_deletions(engine, entity_paths_with_deletions, events);
```
jleibs, line 175: I'm not totally following how this improves performance.

The inclusion of `entity_paths_with_deletions` doesn't change the fact that `on_store_deletions` does an entire tree-walk. 

Is the whole point of this optimization to bypass the overhead of the `is_empty()` call in cases where we know that the intermediate child couldn't have been deleted?

That said, this seems like this will no longer successfully delete intermediate children that only existed as  containers for other entities but don't have their own data.

teh-cmc: > The inclusion of `entity_paths_with_deletions` doesn't change the fact that `on_store_deletions` does an entire tree-walk.
> 
> Is the whole point of this optimization to bypass the overhead of the `is_empty()` call in cases where we know that the intermediate child couldn't have been deleted?

Yeah, the tree walk in itself is imperceptible (it's just a few thousands recursions in the worst case, it's barely measurable) -- a few thousands `is_empty()` on the other hand is extremely costly.

This PR basically brings the latency down from several seconds (ever increasing) to a constant 10ms (using the benchmark script in the issue).

> That said, this seems like this will no longer successfully delete intermediate children that only existed as containers for other entities but don't have their own data.

Haa! I didn't even know that that was the point of this thing. We need to tweak this slightly then.

jleibs: > a few thousands is_empty() on the other hand is extremely costly

Got it -- that wasn't obvious at all and definitely warrants a comment then. I wonder if `is_empty()` should be renamed to something like `check_if_empty()` to better imply there's an active cost to be paid and it's not just accessing some pre-computed state.

jleibs (approved):

PR #8118, opened 2024-11-13
Small video-related cleanups

emilk:
### What
Refactors and the like. Nothing major.

### Checklist


crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -181,9 +182,8 @@ impl VideoPlayer {
         }
     }
 
-    fn frame_at_internal(
+    fn unqueue_samples(
```
Wumpf, line 185: I think you meant this instead :)
```suggestion
    fn enqueue_samples(
```

Wumpf (approved):
lgtm minus naming typo

Issue #8140, opened 2024-11-14
`rerun --serve` without `web-viewer` feature just immediately exists


teh-cmc:
```
$ cargo r -p rerun-cli --no-default-features --features native_viewer,nasm  -- --serve
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.26s
     Running `target/debug/rerun --serve`
[2024-11-14T12:10:06Z DEBUG rerun::commands::entrypoint] Detected 32 cores. Using 30 compute threads.
[2024-11-14T12:10:06Z DEBUG re_memory::memory_limit] Setting memory limit to 47.0 GiB, which is 75% of total available memory (62.7 GiB).
[2024-11-14T12:10:06Z INFO  re_sdk_comms::server] Hosting a SDK server over TCP at 0.0.0.0:9876. Connect with the Rerun logging SDK.
[2024-11-14T12:10:06Z DEBUG re_memory::memory_limit] Setting memory limit to 15.7 GiB, which is 25% of total available memory (62.7 GiB).
[2024-11-14T12:10:06Z INFO  re_ws_comms::server] Hosting a WebSocket server on ws://localhost:9877. You can connect to this with a native viewer (`rerun ws://localhost:9877`) or the web viewer (with `?url=ws://localhost:9877`).
[2024-11-14T12:10:06Z INFO  re_ws_comms::server] Shutting down Rerun server on ws://localhost:9877 after serving 0 client(s)
[2024-11-14T12:10:06Z DEBUG re_ws_comms::server] Stopping WebSocket server listener.
```

That doesn't make sense: `rerun --serve` is very useful in general, it's not specific to the web viewer.

PR #8145, opened 2024-11-14
Let the WebSocket server serve its data even without the web-viewer feature

teh-cmc:
* Fixes https://github.com/rerun-io/rerun/issues/8140

* [x] :ship:


Wumpf (approved):
👍

PR #8151, opened 2024-11-14
Never direct users towards using `rr.log_components`

teh-cmc:
`rr.log_components` is a clunky implementation detail, at best. Users should be directed towards `rr.log`, always.

### Checklist
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8149?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8149?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)

- [PR Build Summary](https://build.rerun.io/pr/8149)

To deploy documentation changes immediately after merging this PR, add the `deploy docs` label.


teh-cmc:
@rerun-bot full-check

teh-cmc:
> right, we made `log` handle this a while ago. What are cases where one would still need it actually? Should we deprecate it?

I thought about deprecating it at one point, but then realized that we ourselves ultimately delegate to it in order to actually log stuff... but now that I'm taking a closer look, it seems to be the exact same code as `log()` :facepalm: 

I'll take a closer look -- if it is indeed the exact same code, I'll deprecate it too.

teh-cmc:
Nuh-uh, nevermind, they're not even close to the same, and it even makes sense as is :+1:

PR #8057, opened 2024-11-11
C++ SDK: add unreachable `default` cases in switch statements

teh-cmc:
Whenever I work on the C++ SDK, I get several hundreds warnings of that nature every time I compile (clang18, gcc14):
```
/home/cmc/dev/rerun-io/rerun/rerun_cpp/src/rerun/archetypes/../collection.hpp:304:13: warning: 'switch' missing 'default' label [-Wswitch-default]
  304 |             switch (ownership) {
      |             ^
/home/cmc/dev/rerun-io/rerun/rerun_cpp/src/rerun/archetypes/../component_batch.hpp:47:59: note: in instantiation of member function 'rerun::Collection<rerun::components::IndicatorComponent<rerun::archetypes::EncodedImage::IndicatorComponentName>>::data' requested here
   47 |             auto array = Loggable<T>::to_arrow(components.data(), components.size());
      |                                                           ^
/home/cmc/dev/rerun-io/rerun/rerun_cpp/src/rerun/archetypes/../component_batch.hpp:63:20: note: in instantiation of function template specialization 'rerun::ComponentBatch::from_loggable<rerun::components::IndicatorComponent<rerun::archetypes::EncodedImage::IndicatorComponentName>>' requested here
   63 |             return from_loggable(collection);
      |                    ^
/home/cmc/dev/rerun-io/rerun/rerun_cpp/src/rerun/archetypes/encoded_image.cpp:41:43: note: in instantiation of function template specialization 'rerun::ComponentBatch::from_loggable<rerun::components::IndicatorComponent<rerun::archetypes::EncodedImage::IndicatorComponentName>>' requested here
   41 |             auto result = ComponentBatch::from_loggable(indicator);
```
Not only is it annoying but I would also tend to agree with it?

This PR simply adds an unreachable `default` case everywhere:
```cpp
default:
    assert(false && "unreachable");
```
which might or might not be the right approach, I don't know.

### Checklist
* [x] I have read and agree to [Contributor Guide](https://github.com/rerun-io/rerun/blob/main/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/rerun-io/rerun/blob/main/CODE_OF_CONDUCT.md)
* [x] I've included a screenshot or gif (if applicable)
* [x] I have tested the web demo (if applicable):
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/8057?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
[...]

PR #8160, opened 2024-11-15
Fix last remaining C++ warnings

teh-cmc:
This fixes the last two remaining warnings I get when `pixi run cpp-build-all`.

* One of them is just a switch I forgot in https://github.com/rerun-io/rerun/pull/8057.
* The other is just GCC being weird... I dunno man.

---

* [x] :seal:

### DIFF
--- a/tests/cpp/plot_dashboard_stress/main.cpp
+++ b/tests/cpp/plot_dashboard_stress/main.cpp
@@ -147,9 +147,14 @@ int main(int argc, char** argv) {
 
     std::vector<size_t> offsets;
     if (temporal_batch_size.has_value()) {
+// GCC wrongfully thinks that `temporal_batch_size` might be uninit, even though we've
+// literally checked for that in the line above?!
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
         for (size_t i = 0; i < num_points_per_series; i += *temporal_batch_size) {
             offsets.push_back(i);
         }
+#pragma GCC diagnostic pop
     } else {
         offsets.resize(sim_times.size());
         std::iota(offsets.begin(), offsets.end(), 0);
--- a/rerun_cpp/src/rerun/recording_stream.cpp
+++ b/rerun_cpp/src/rerun/recording_stream.cpp
@@ -7,6 +7,7 @@
 
 #include <arrow/buffer.h>
 
+#include <cassert>
 #include <string> // to_string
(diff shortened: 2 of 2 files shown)


tests/cpp/plot_dashboard_stress/main.cpp
```diff
@@ -147,9 +147,14 @@ int main(int argc, char** argv) {
...
     std::vector<size_t> offsets;
     if (temporal_batch_size.has_value()) {
+// GCC wrongfully thinks that `temporal_batch_size` might be uninit, even though we've
+// literally checked for that in the line above?!
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
         for (size_t i = 0; i < num_points_per_series; i += *temporal_batch_size) {
             offsets.push_back(i);
         }
+#pragma GCC diagnostic pop
```
Wumpf, line 157: this happens so often, we have it in our compiler_utils
https://github.com/rerun-io/rerun/blob/main/rerun_cpp/src/rerun/compiler_utils.hpp#L24C9-L24C44

that would make this
```suggestion
// GCC wrongfully thinks that `temporal_batch_size` is uninitialized despite being initialized upon creation.
RR_DISABLE_MAYBE_UNINITIALIZED_PUSH
        for (size_t i = 0; i < num_points_per_series; i += *temporal_batch_size) {
            offsets.push_back(i);
        }
RR_DISABLE_MAYBE_UNINITIALIZED_POP
```

The comment is also _technically_ not entirely to the point which why my suggestion adjusts that as well: `has_value` has nothing to do with uninit. This is just a member of `std::optional` which may or may not be uninitialized which means that the outcome of it would be indetermined.
That said, this is still _obviously_ bs, since temporal_batch_size comes to life like this:
```
    const auto temporal_batch_size = args.count("temporal-batch-size")
                                         ? std::optional(args["temporal-batch-size"].as<uint64_t>())
: std::nullopt;
```

Wumpf (approved):
👍 with nits

tests/cpp/plot_dashboard_stress/main.cpp
teh-cmc: What I don't get is why this is the only case where GCC complains in this file, even though we do the exact same thing in 10 different places... :facepalm:

Wumpf: gcc moves in mysterious ways

PR #8181, opened 2024-11-19
Revert bump to 0.20.1-alpha.2

emilk:
* Reverts parts of rerun-io/rerun#8172


emilk:
Docs work! https://docs.rs/crate/rerun/0.20.1-alpha.3

Issue #2978, opened 2023-08-14
Migrate `ArrowBuffer` wrapper to expose `arrow-rs` buffers rather than `arrow2`


jleibs:
https://github.com/rerun-io/rerun/pull/2970 introduced a new `ArrowBuffer` wrapper to return native arrow buffers on the deserialization path for Tensor types.

These buffers currently use `arrow2::Buffer`. In order to facilitate the future migration to `arrow-rs`, these can be made to wrap `arrow-rs::Buffer` instead. The types have convenient converters: https://docs.rs/arrow2/latest/arrow2/buffer/struct.Buffer.html#impl-From%3CBuffer%3CT%3E%3E-for-Buffer but we need to investigate whether these converters are still zero-copy.

One benefit of this migration is arrow-rs buffers use `half:f16` whereas `arrow2` buffers use `arrow2::types::f16`

Issue #3741, opened 2023-10-09
Migrate from `re_arrow2` to `arrow`


teh-cmc:
Blockers
* Soft-blocked on https://github.com/apache/arrow-rs/issues/6360 (for lowering memory use)
* Semi-blocked on finding a work-around for https://github.com/apache/arrow-rs/issues/4472 (we use `DataType::Extension` for Tuid)

---

Multiple end-goals:
- Use same arrow lib as the rest of the ecosystem, which is where all the bug & perf fixes are actually happening
- Use inifinitely less space to store Arrow metadata (schema deduplication)
   - #1809
- Make it possible to send raw Arrow data to Rerun and have it just work (`RERUN:component_name`)
  - #3360
  - Also frees up usage of Arrow extensions for actual native extensions (e.g. #3004)
- Native integration with `half` for `f16`
- Etc etc

---

TODO (split into sub-issues as needed):
- [x] Migrate serde-based components (i.e. blueprint stuff) to `arrow1`
- [x] Move `SizeBytes` to own crate
- [x] Move `fn arrow_ui` to `re_ui`
- [x] Migrate all generated code (re_types)
- [x] https://github.com/rerun-io/rerun/pull/8617
- [x] `TimeColumn`
- https://github.com/rerun-io/rerun/pull/8700
- [x] `dataplatform` repo
- [x] gRPC client/server
- [x] Protobuf conversions
- [x] Python interface (dataframe)
- [x] Migrate everything in `re_types_core`
- [x] Migrate everything in `re_dataframe`
- [ ] Remove use of `arrow2::Datype` in `re_types_builder`
- [x] `Chunk`
- [x] `ChunkBuilder`
- [ ] Remove all direct uses of `arrow2` (codegen, data{cell,row,table}, `ArrowBuffer`, etc)
  - Related: #2978 
- [ ] Get rid of Arrow extensions everywhere, introduce `RERUN:component_name`  (#3360)
- [ ] Runtime schema registry / dedupe datatypes (#1809)
- [ ] Remove `DataCell::component_name`

On the way there we might hit a few bumps because we have a lot of redundant ad-hoc code that integrates with `polars` (which is built on top of `arrow2`).

The solution to this is to make sure we only integrate with `polars` in one single place: the `Data{Cell,Row,Table}` layer (https://github.com/rerun-io/rerun/issues/1692).
[...]

PR #8201, opened 2024-11-22
Use `arrow-rs` in `ArrowBuffer`

emilk:
* Closes https://github.com/rerun-io/rerun/issues/2978

### DIFF
--- a/crates/store/re_types_core/src/arrow_buffer.rs
+++ b/crates/store/re_types_core/src/arrow_buffer.rs
@@ -1,6 +1,6 @@
 use arrow::datatypes::ArrowNativeType;
 
-/// Convenience-wrapper around an [`arrow2::buffer::Buffer`] that is known to contain a
+/// Convenience-wrapper around an [`arrow::buffer::ScalarBuffer`] that is known to contain a
 /// a primitive type.
 ///
 /// The [`ArrowBuffer`] object is internally reference-counted and can be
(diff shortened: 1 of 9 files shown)


crates/store/re_types_core/src/arrow_buffer.rs
```diff
@@ -84,56 +83,79 @@ impl<T: bytemuck::Pod> ArrowBuffer<T> {
...
 
-impl<T> From<arrow2::buffer::Buffer<T>> for ArrowBuffer<T> {
+impl<T: ArrowNativeType + arrow2::types::NativeType> From<arrow2::buffer::Buffer<T>>
+    for ArrowBuffer<T>
+{
     #[inline]
-    fn from(value: arrow2::buffer::Buffer<T>) -> Self {
-        Self(value)
+    fn from(arrow2_buffer: arrow2::buffer::Buffer<T>) -> Self {
+        let num_elements = arrow2_buffer.len();
```
Wumpf, line 109: didn't the fat comment up there say that this is a `num_bytes` in arrow2? That would make this wrong then if I understand correctly

Your unit test seems to pass this though, confusing.

Wumpf (approved):
lgtm minus suspicious/confusing `len`

crates/store/re_types_core/src/arrow_buffer.rs
emilk: This is why I wrote a unit test - arrow2 is confusing and maybe wrongly documented:
* https://github.com/jorgecarleitao/arrow2/issues/1430

The comment is gone too

PR #8203, opened 2024-11-22
Remove dead code: `Hash128`

teh-cmc:
`Hash128` isn't actually used anywhere, I just randomly stumbled upon it while refactoring some hashing related stuff.

Still a good idea to fix it (or remove it entirely?) before someone's get bitten.

### DIFF
--- a/crates/store/re_log_types/src/hash.rs
+++ b/crates/store/re_log_types/src/hash.rs
@@ -86,7 +86,7 @@ impl Hash128 {
 impl std::hash::Hash for Hash128 {
     #[inline]
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        state.write_u64(self.0[0]);
+        state.write_u128((self.first64() as u128) << 64 | self.second64() as u128);
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/store/re_log_types/src/hash.rs
```diff
@@ -91,7 +86,7 @@ impl Hash128 {
 impl std::hash::Hash for Hash128 {
     #[inline]
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        state.write_u64(self.0[0]);
+        state.write_u128((self.first64() as u128) << 64 | self.second64() as u128);
```
emilk, line 89: IIRC, you're only allowed to use `write_u64` with `nohash_hasher`, anything else will result in a runtime error in debug builds.

If the hash128 is good, than any 64 bits of it should also be a good hash

emilk (changes requested):

emilk:
If `Hash128` is unused, remove it - but I don' think this PR makes sense

emilk (approved):

PR #8206, opened 2024-11-22
Port codegen of arrow datatype to arrow1

emilk:
* Part of  #3741


abey79:
@rerun-bot full-check

abey79 (approved):
Nice. Seems rather straightforward. Roundtrip tests passing should be the definitive validation.

PR #8207, opened 2024-11-22
Clean up pass over all superfluous hashing happening on the query path

teh-cmc:
2x-3x perf improvements in scenes with many entities :grimacing: 

Before:
image

After:
image

### DIFF
--- a/crates/store/re_chunk/src/transport.rs
+++ b/crates/store/re_chunk/src/transport.rs
@@ -1,5 +1,3 @@
-use std::collections::BTreeMap;
-
 use arrow2::{
     array::{
         Array as Arrow2Array, ListArray, PrimitiveArray as Arrow2PrimitiveArray,
@@ -12,6 +10,8 @@ use arrow2::{
     },
 };
 
+use itertools::Itertools;
+use nohash_hasher::IntMap;
 use re_log_types::{EntityPath, Timeline};
 use re_types_core::{Component as _, Loggable as _, SizeBytes};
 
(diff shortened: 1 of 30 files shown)


teh-cmc:
@rerun-bot full-check

crates/store/re_chunk/src/transport.rs
```diff
@@ -396,42 +396,65 @@ impl Chunk {
...
+                        metadata
+                    });
+
+                    let times = times.clone().boxed() /* cheap */;
+
+                    (field, times)
+                })
+                .collect_vec();
 
+            timelines.sort_by(|(field1, _times1), (field2, _times2)| field1.name.cmp(&field2.name));
```
Wumpf, line 428: a bit surprised we care about the order in the transport chunk?

crates/store/re_chunk_store/tests/snapshots/memory_test__scalars_on_one_timeline_new.snap
```diff
@@ -4,6 +4,6 @@ expression: "[format!(\"{NUM_SCALARS} scalars\"),\n        format!(\"{} in total
 ---
 [
     "1048576 scalars",
-    "37.5 MiB in total",
-    "38 B per row",
+    "37.0 MiB in total",
+    "37 B per row",
```
Wumpf, line 8: .. hurray...?
So not sorting in some places saved memory? Guess something randomly compacted better?

crates/store/re_chunk_store/src/query.rs
```diff
@@ -41,7 +64,27 @@ impl ChunkStore {
...
+                            },
+                        )
+                    }),
+            )
+            .copied()
+            .collect()
+    }
+
+    /// Retrieve all [`ComponentName`]s in the store.
+    pub fn all_components_sorted(&self) -> ComponentNameSet {
```
Wumpf, line 87: Maybe we should rename this type to `OrderedComponentNameSet` to pair it up with `UnorderedComponentNameSet`, a bit confusing in the context of this PR :/

(or even better imho `ComponentNameOrderedSet` & `ComponentNameUnorderedSet` for friendlier autocomplete)

Wumpf (approved):
looks all sane to me. Crazy how impactful this is, great finds!

🤞 this didn't introduce unordering where there shouldn't be any, but looks tractable

Wumpf:
I want this and none of the questions/comment here are blocking enough to not just merge this.

crates/store/re_chunk/src/transport.rs
teh-cmc: We do! RecordBatches respect the order of columns so keeping it stable is pretty important, otherwise the UX for end users is pretty awful.

crates/store/re_chunk_store/src/query.rs
teh-cmc: I wanted to, but the UI code already exposes a `SortedComponentSet` with slightly different rules... :zany_face:

Wumpf: 🙈

PR #8209, opened 2024-11-22
Fix command for running python build in BUILD.md

jviereck:


abey79:
Hello. Thanks for your contribution. Unfortunately, your change suggestion is incorrect: this command does not produce an installable wheel. We were currently lacking such a command, so I introduced one and updated the BUILD.md instructions accordingly. For some reason, I was unable to push these changes to the present PR, so I opened a new one here:
- https://github.com/rerun-io/rerun/pull/8235

PR #8210, opened 2024-11-22
Introduce `RecordingUri` component

abey79:
### What

This PR introduces a new `RecordingUri` component (which contains just a string) and assorted UI, which present the URI as a link that opens said component. Also introduce a new `FileSource::Uri` to account for this new way of opening recordings.

Note: displaying links is disabled in wasm build, because `DataSource::from_uri()` is not available there. I'm actually not quite sure what the story is to load remote recordings from the web viewer 🤷🏻 

image


Wumpf:
screenshot? :)

jleibs (approved):

PR #8211, opened 2024-11-22
Adding DefaultCamera to Spatial3D view

jviereck:
At the moment it is not possible to set the camera position and view in the Spatial3D view. This PR implements this.


jviereck:
This PR needs to be rebased against the current main branch.

Please let me know what you mean about the overall shape of the changes. If they are good, I will rebase against main.

crates/viewer/re_space_view_spatial/src/ui_3d.rs
```diff
@@ -112,7 +128,28 @@ impl View3DState {
...
+                    self.view_eye_default = Some(from_to_eye(
+                        vec3(pos_origin.0),
+                        vec3(pos_target.0),
+                        scene_view_coordinates,
+                    ));
+
+                    self.view_eye = self.view_eye_default;
+                }
+            }
+        }
```
Wumpf, line 151: I don't quite follow why this is needed.
Shouldn't the behavior of `default_eye` be adjusted instead?

crates/store/re_types/definitions/rerun/blueprint/archetypes/default_camera.fbs
```diff
@@ -0,0 +1,16 @@
+namespace rerun.blueprint.archetypes;
+
+
+/// Defines a default camera view.
+table DefaultCamera (
```
Wumpf, line 5: Maybe more like
```suggestion
table DefaultEyeCameraPose (
```

The problem with `camera` is that...
* it gets confused with logged pinhole cameras a lot which is why we just call it `eye` in the implementation
    * `eye` alone though is probably even more confusing 
* this is not a full camera specification but just where it is. Later we'll have to add more things like camera controls (orbit vs ego vs future stuff) etc. which is then the "full specification" which will have a lot more fields

Wumpf (changes requested):
at first I though it's an odd starting place to set the default camera given that the camera itself isn't specified yet
But I realize that this is the easiest entry point into being able to set the camera without having to worry about blueprint update frequency and the various other properties, so great initiative!

The implementation is off the mark however:
* properties should be exposed in the ui, that isn't wired up here
* there should be default provider implementations for these components now that do what `default_eye` did before.
     * `default_eye` then should always go to the components which may in turn be routed to the default providers
-> this framework ensures that the assumed state of the property is transparent for the rest of the viewer (i.e. just the selection panel today)

I have a bit of time this week for smaller tasks like this, if you want I can take it from here

----
idea for a follow-up PR: extend the ui with a "set current" button that allows setting the default to the current camera pose. that way one can always return to it with double click on the viewer \o/

crates/viewer/re_space_view_spatial/src/ui_3d.rs
```diff
@@ -92,6 +101,10 @@ fn ease_out(t: f32) -> f32 {
     1. - (1. - t) * (1. - t)
 }
 
+fn vec3(v: Vec3D) -> Vec3 {
```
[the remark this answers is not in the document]
jviereck: I am pretty new to Rust, so not sure what you mean by `into` to be honest :/

crates/viewer/re_space_view_spatial/src/ui_3d.rs
```diff
@@ -41,6 +47,8 @@ use super::eye::{Eye, ViewEye};
 pub struct View3DState {
     pub view_eye: Option<ViewEye>,
 
+    pub view_eye_default: Option<ViewEye>,
```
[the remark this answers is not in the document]
jviereck: See next comment.

crates/viewer/re_space_view_spatial/src/ui_3d.rs
jviereck: When double clicking on the 3D view, I want the camera to jump back to the default camera settings, if specified. The double click calls the `View3DState::reset_camera` method. While the default camera property is available from the gui-rendering path (`default_camera.component_or_empty::<CameraOrigin>()` etc), it is not so easy to query it from the `reset_camera` method. As such, I added a new field `view_eye_default` on the `View3DState`, which is set the first time the gui renders. This field is then accessible from the `View3DState::reset_camera` method.

I know this is not great and it would be cleaner to query to store for the `CameraOrigin` etc component directly, but I was not sure how to do that easily. So I decided to go for this workaround.

Wumpf:
putting this on draft for now until I resurrect it (hopefully later this week, maybe early next)

Wumpf:
Obsolete now - camera in 0.27 is fully configurable by blueprint

PR #8213, opened 2024-11-24
Add from_gray16 for DepthImage

fawdlstty:
Depth image are usually stored in a 16-bit gray format. This pr is used to simplify this step

### DIFF
--- a/crates/store/re_types/src/archetypes/depth_image_ext.rs
+++ b/crates/store/re_types/src/archetypes/depth_image_ext.rs
@@ -1,5 +1,6 @@
 use crate::{
-    datatypes::{ImageFormat, TensorData},
+    components::{ImageBuffer, ImageFormat},
+    datatypes::{ChannelDatatype, ColorModel, TensorData},
     image::{blob_and_datatype_from_tensor, find_non_empty_dim_indices, ImageConstructionError},
 };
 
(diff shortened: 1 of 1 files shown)


crates/store/re_types/src/archetypes/depth_image_ext.rs
```diff
@@ -37,12 +38,45 @@ impl DepthImage {
...
+            format: image_format,
             draw_order: None,
             meter: None,
             colormap: None,
             point_fill_ratio: None,
             depth_range: None,
         })
     }
+
+    /// Construct an image from a byte buffer given its resolution, and data type.
```
Wumpf, line 50:
```suggestion
    /// Construct a depth image from a byte buffer given its resolution, and data type.
```

Wumpf (changes requested):
thanks for contributing! Looks good to me, but would be nice to make use of that in something that is tested from CI.
Doesn't need to be fancy, just compare outcome of `from_gray16` with using `try_from`. You could for instance the snippet from `depth_image_simple` and check that it constructs the depth image the same way when built with `from_gray16` or `from_data_type_and_bytes`. Test can as a separate test function to `crates/store/re_types/tests/types/depth_image.rs`

PR #8217, opened 2024-11-25
Remove hack for setting `ExternalImageSource::VideoFrame` in wasm video processing

Wumpf:
Thanks to @jprochazk pushing a fix upstream & us now being on wgpu 23, we no longer need this hack

Gave it a spin locally to verify that video still plays on the web player.

### DIFF
--- a/crates/viewer/re_renderer/src/video/chunk_decoder.rs
+++ b/crates/viewer/re_renderer/src/video/chunk_decoder.rs
@@ -189,7 +189,7 @@ fn copy_web_video_frame_to_texture(
         // Careful: `web_sys::VideoFrame` has a custom `clone` method:
         // https://developer.mozilla.org/en-US/docs/Web/API/VideoFrame/clone
         // We instead just want to clone the js value!
-        source: wgpu::ExternalImageSource::VideoFrame(Clone::clone(frame)),
+        source: wgpu::ExternalImageSource::VideoFrame(JsValue::clone(frame)),
         origin: wgpu::Origin2d { x: 0, y: 0 },
         flip_y: false,
     };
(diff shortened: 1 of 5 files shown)


crates/viewer/re_renderer/src/video/chunk_decoder.rs
```diff
@@ -185,38 +185,13 @@ fn copy_web_video_frame_to_texture(
...
-        wgpu_types::ImageCopyExternalImage {
-            source: wgpu_types::ExternalImageSource::HTMLVideoElement(frame),
-            origin: wgpu_types::Origin2d { x: 0, y: 0 },
-            flip_y: false,
-        }
+    let source = wgpu::ImageCopyExternalImage {
+        // Careful: `web_sys::VideoFrame` has a custom `clone` method:
+        // https://developer.mozilla.org/en-US/docs/Web/API/VideoFrame/clone
+        // We instead just want to clone the js value!
+        source: wgpu::ExternalImageSource::VideoFrame(Clone::clone(frame)),
```
emilk, line 192: This looks like it would be a better way of ensuring that

```suggestion
        source: wgpu::ExternalImageSource::VideoFrame(JsValue::clone(frame)),
```

crates/viewer/re_renderer/src/video/mod.rs
```diff
@@ -18,7 +18,7 @@ pub enum VideoPlayerError {
     #[error("The decoder is lagging behind")]
     EmptyBuffer,
 
-    #[error("Video seems to be empty, no segments have beem found.")]
+    #[error("Video seems to be empty, no segments have been found.")]
```
emilk, line 21: nit: we have nothing called "segments" anymore.

```suggestion
    #[error("Video is empty.")]
```

emilk (approved):

PR #8220, opened 2024-11-25
DRAFT: Add examples-ros environment based on robostack

jleibs:
### What

Make it easier to run the ROS example by adding a robostack-based environment to pixi.

In theory this should work:
```
pixi run launch-turtlebot
```

The example can then be run as:
```
pixi run -e examples-ros python examples/python/ros_node/main.py 
```

However it looks like the example has bit-rot over time and probably needs some bug fixes.


grtlr:
I've managed to launch `rqt_graph` too:
image

PR #8223, opened 2024-11-26
Introduce APIs for list_recordings and update_metadata

jleibs:
### Related
- https://github.com/rerun-io/dataplatform/issues/17

### What
- Python API support for the list / update APIs
- Example script that uses them for a simple CLI app

### DIFF
--- a/rerun_py/Cargo.toml
+++ b/rerun_py/Cargo.toml
@@ -40,6 +40,7 @@ remote = [
   "dep:re_protos",
   "dep:re_ws_comms",
   "dep:tokio",
+  "dep:tokio-stream",
   "dep:tonic",
   "dep:url",
 ]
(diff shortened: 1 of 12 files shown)


rerun_py/Cargo.toml
```diff
@@ -15,7 +15,7 @@ name = "rerun_bindings" # name of the .so library that the Python module will im
 
 
 [features]
-default = ["extension-module"]
+default = ["extension-module", "remote"]
```
jleibs, line 18: Don't merge this change.

rerun_py/src/remote.rs
```diff
@@ -64,13 +79,33 @@ impl PyConnection {
...
+                .collect::<Result<Vec<_>, _>>()
+                .map_err(|err| PyRuntimeError::new_err(err.to_string()))?;
+
+            let record_batches: Vec<Result<RecordBatch, arrow::error::ArrowError>> =
+                transport_chunks
+                    .into_iter()
+                    .map(|tc| tc.try_to_arrow_record_batch())
+                    .collect();
+
+            // TODO(jleibs): surfacing this schema is awkward. This should be more explicit in
```
zehiko, line 96: should we expose schema as top level field for get/list metadata, even though it's in the transport chunk? 

I guess for the Query (i.e. data API) we should (as initially discussed) strive to go the opposite way and only include the schema in the first message of the stream?

zehiko (approved):
LGTM, but I agree that we should spend just a bit of time thinking about the next level of improvements for all the APIs

PR #8225, opened 2024-11-26
Build npm package with `map_view,grpc` features

jprochazk:
The npm package build script didn't include some feature flags which means currently the map view snippet in docs in unusable.


abey79 (approved):
lgtm

Issue #872, opened 2023-01-23
World-space grid


emilk:
The 3D spatial view should have an option to show the X^Y, X^Z, and Y^Z planes as grids, as common in 3D apps.

chengguizi:
Hi Rerun team, I am also interested in this feature. Showing a grid to tell the scale of the world is pretty useful feature!

Wumpf:
There's three related we have to figure out for this:
* what should it look like (out of the box at least) and how does it behave under zoom
* what should be configurable
* how do we render it in a clean fashion

References:
* Unity [zoom](https://github.com/user-attachments/assets/72c2b1a5-2fe2-4ea5-91a1-b6c6aaf02ee5)
<img width="1237" alt="image" src="https://github.com/user-attachments/assets/eee5e725-9658-484e-bb83-bc54860a7126">
* Blender. [zoom](https://github.com/user-attachments/assets/8e497759-6bb6-4a66-92c2-6b7717d6ef80)
<img width="738" alt="image" src="https://github.com/user-attachments/assets/9e8a96f1-6a9b-44ad-b87b-2a1d03695bb4">
* Todo: more!

Both of these have a whole bunch of things in coming
* very thin, ui-constant sized lines (lines don't get smaller in the distance)
    * generally try not to get in the way!
* fading to cardinal lines
* fading in the distance

On cardinal lines: The problem mostly arises with zooming out. Fading lines is pretty much a must, otherwise the screen becomes gradually fully opaque no matter how small the lines are. So typically one would just fade everything but every tenth line (configurable?). That introduces a cardinality which likely should be expressed with a slightly different shader and/or thickness. Blender does this really well imho.

Another thing shown here is highlighting for cardinal axis. Blender has this configurable (for all axis independently!) but I'd keep this out for the moment, worth considering adding to the options.

Also seen in the blender clip is fading of the unlimited grid with distance which can be important to avoid artifacts as well. Unity actually just clips the lines off at some point. Which is also perfectly fine as long as the lines stay thin!

---

Unlike back in the day when this was first logged we now have a system in place for configuring views, and this is a great fit for that, so let's start by what the rough data modelling for this could look like:

```rust
table LineGuides3D {
    /// Whether the grid is visible.
    ///
    /// Default is true.
    visible: rerun.components.Visible ("attr.rerun.component_optional", nullable);

    /// The type of the grid.
[...]

Issue #1426, opened 2023-02-28
Unify camera logic in re_renderer examples


teh-cmc:
What the title says.

PR #8230, opened 2024-11-27
World grid part 1/2: add world grid renderer to `re_renderer`

Wumpf:
### Related

This is the first half of

* https://github.com/rerun-io/rerun/issues/872

The second half is here:
* https://github.com/rerun-io/rerun/pull/8234

### What

Implements a "world grid" to be used later in the spatial views of the viewer.
Haven't tested yet how suitable this is for 2D views, might need some adjustments.

## Video

Youtube unfortunately compresses this a lot and Github doesn't let me upload any significant amount of video
(click me)
[youtube video of world grid v1](http://www.youtube.com/watch?v=Ho-iGdmJi4Q "World Grid v1")

## Screenshots

Analytical anti-aliasing (shown is what is supposed to be a 1ui unit == 2pixel wide line):
image

Distance has only the cardinal lines (== every tenth line). Those fade eventually as well:
image

Fading is based on "how many lines coincide here" which makes fading view dependent rather than distance dependent:
image
(this makes this hopefully robust for many usecases)

Grid intersects non-transparent geometry (transparent geometry will be ignored in the future. previous surveying showed that this is common and not a biggy)
image

Grid fades at acute viewing angles (because empirically and unsurprisingly it looks weird if we don't!)
image

Tested image quality to be satisfying on high dpi (ui unit == 2 pixels) and low dpi (ui unit == pixel).

## How does it work

* Draw a large shader generated plane (mostly because setting up vertex buffers is just more hassle 😄 )
    * depth test enabled
    * depth write disabled
    * premultiplied alpha blend
    * make sure we draw this after the background (reminder: we first draw solid, then background)
* Fragment shader looks at 2d coordinate on the plane and decides how "liney" it is
   * using screen space derivatives (ddx/ddy) we figure out how far to go on the plane to achieve a certain line thickness
   * fades:
        * fade if the line density gets too high
        * fade if view angle is too acute relative  
   * ... lot's of details documented in the shader code!

I considered just drawing a screen space triangle, but drawing a plane that moves with the camera has lots of advantages:
* don't have to manipulate depth
    * faster since early z just works
    * less thinking required!
* don't cover things above the horizon - even if only the grid is visible, less pixels will be shaded 

## Known shortcomings

* various hardcoded values around how fading works. Tricky to formalize this better, but likely good enough
    * Doesn't look equally good at all pixel widths, but decent for the range we care
* every tenth line is a "cardinal line", that's nice but it stops there - "infinite" amount of cardinal lines would be nicer
    * Experimented with that but so far didn't get anything that was compelling. Having many order-of-magnitude lines looks way too busy imho, it needs a mechanism that limits that to two.
    * Blender's 2D view does this really well, but in 3D they only do two cardinals as well.

## Try it yourself

Controls:
- Space: Pause
- G: Toggle camera mode

native:
```
cargo run -p re_renderer_examples --bin world_grid
```

web:
```
cargo run-wasm -p re_renderer_examples --bin world_grid
```

## Backend testing matrix

* [x] Metal
* [x] Vulkan
* [x] WebGL
* [x] WebGPU

### DIFF
--- a/crates/viewer/re_renderer/shader/world_grid.wgsl
+++ b/crates/viewer/re_renderer/shader/world_grid.wgsl
@@ -101,7 +101,7 @@ fn main_fs(in: VertexOutput) -> @location(0) vec4f {
     //
     // Note that `1/plane_unit_pixel_derivative` would give us more literal pixels per line,
     // but we actually want to know how dense the lines get here so we use `1/width_in_grid_units` instead,
-    // such that a line density of 1.0 means roughtly "100% lines" and 10.0 means "Every 10 pixels there is a lines".
+    // such that a line density of 1.0 means roughly "100% lines" and 10.0 means "Every 10 pixels there is a lines".
     // Empirically (== making the fade a hard cut and taking screenshot), this works out pretty accurately!
     //
     // Tried smoothstep here, but didn't feel right even with lots of range tweaking.
@@ -127,7 +127,7 @@ fn main_fs(in: VertexOutput) -> @location(0) vec4f {
     // X and Y are combined like akin to premultiplied alpha operations.
     let intensity_combined = saturate(cardinal_and_regular.x * (1.0 - cardinal_and_regular.y) + cardinal_and_regular.y);
 
-    // Fade on accute viewing angles.
+    // Fade on acute viewing angles.
     var view_direction_up_component: f32;
     switch (config.orientation) {
         case ORIENTATION_XZ: {
--- a/crates/viewer/re_renderer/shader/world_grid.wgsl
+++ b/crates/viewer/re_renderer/shader/world_grid.wgsl
@@ -90,7 +90,7 @@ fn main_fs(in: VertexOutput) -> @location(0) vec4f {
     // Figure out the how wide the lines are in this "draw space".
     let plane_unit_pixel_derivative = fwidthFine(in.scaled_world_plane_position);
     let line_anti_alias = plane_unit_pixel_derivative;
-    let width_in_pixels = 1.0;//config.thickness_ui * frame.pixels_from_point;
+    let width_in_pixels = config.thickness_ui * frame.pixels_from_point;
     let width_in_grid_units = width_in_pixels * plane_unit_pixel_derivative;
     var intensity_regular = linearstep2(width_in_grid_units + line_anti_alias, width_in_grid_units - line_anti_alias, distance_to_grid_line);
 
@@ -99,12 +99,12 @@ fn main_fs(in: VertexOutput) -> @location(0) vec4f {
     //
     // Note that `1/plane_unit_pixel_derivative` would give us more literal pixels per line,
     // but we actually want to know how dense the lines get here so we use `1/width_in_grid_units` instead,
-    // such that a line density of 1.0 means roughly "100% lines" and 10.0 means "Every 10 pixels there is a lines".
+    // such that a value of 1.0 means roughly "100% lines" and 10.0 means "Every 10 pixels there is a lines".
     // Empirically (== making the fade a hard cut and taking screenshot), this works out pretty accurately!
     //
     // Tried smoothstep here, but didn't feel right even with lots of range tweaking.
-    let line_density = 1.0 / max(width_in_grid_units.x, width_in_grid_units.y);
-    let grid_closeness_fade = linearstep(1.0, 10.0, line_density);
+    let screen_space_line_spacing = 1.0 / max(width_in_grid_units.x, width_in_grid_units.y);
+    let grid_closeness_fade = linearstep(1.0, 10.0, screen_space_line_spacing);
     intensity_regular *= grid_closeness_fade;
 
     // Every tenth line is a more intense, we call those "cardinal" lines.
@@ -114,7 +114,7 @@ fn main_fs(in: VertexOutput) -> @location(0) vec4f {
     let distance_to_grid_line_cardinal = calc_distance_to_grid_line(in.scaled_world_plane_position * (1.0 / CARDINAL_LINE_FACTOR));
     var cardinal_line_intensity = linearstep2(width_in_grid_units + line_anti_alias, width_in_grid_units - line_anti_alias,
                                               distance_to_grid_line_cardinal * CARDINAL_LINE_FACTOR);
-    let cardinal_grid_closeness_fade = linearstep(2.0, 10.0, line_density * CARDINAL_LINE_FACTOR); // Fade cardinal lines a little bit earlier (because it looks nicer)
+    let cardinal_grid_closeness_fade = linearstep(2.0, 10.0, screen_space_line_spacing * CARDINAL_LINE_FACTOR); // Fade cardinal lines a little bit earlier (because it looks nicer)
     cardinal_line_intensity *= cardinal_grid_closeness_fade;
 
     // Combine all lines.
--- a/crates/viewer/re_renderer/shader/world_grid.wgsl
+++ b/crates/viewer/re_renderer/shader/world_grid.wgsl
@@ -1,4 +1,5 @@
 #import <./global_bindings.wgsl>
+#import <./utils/interpolation.wgsl>
 
 struct WorldGridUniformBuffer {
     color: vec4f,
@@ -48,13 +49,13 @@ fn main_vs(@builtin(vertex_index) v_idx: u32) -> VertexOutput {
     var plane_position = (vec2f(f32(v_idx / 2u), f32(v_idx % 2u)) * 2.0 - 1.0) * PLANE_GEOMETRY_SIZE;
     var world_position: vec3f;
     switch (config.orientation) {
-        case ORIENTATION_XZ: {
-            plane_position += frame.camera_position.xz;
-            world_position = vec3f(plane_position.x, 0.0, plane_position.y);
-        }
         case ORIENTATION_YZ: {
             plane_position += frame.camera_position.yz;
             world_position = vec3f(0.0, plane_position.x, plane_position.y);
+        }
+        case ORIENTATION_ZX: {
+            plane_position += frame.camera_position.xz;
+            world_position = vec3f(plane_position.x, 0.0, plane_position.y);
         }
         case ORIENTATION_XY: {
             plane_position += frame.camera_position.xy;
@@ -71,14 +72,6 @@ fn main_vs(@builtin(vertex_index) v_idx: u32) -> VertexOutput {
     return out;
 }
 
-// Like smoothstep, but linear.
-// Used for antialiasing: smoothstep works as well but is subtly worse.
-fn linearstep(edge0: f32, edge1: f32, x: f32) -> f32 {
-    return saturate((x - edge0) / (edge1 - edge0));
-}
-fn linearstep2(edge0: vec2f, edge1: vec2f, x: vec2f) -> vec2f {
-    return saturate((x - edge0) / (edge1 - edge0));
-}
 
 // Distance to a grid line in x and y ranging from 0 to 1.
 fn calc_distance_to_grid_line(scaled_world_plane_position: vec2f) -> vec2f {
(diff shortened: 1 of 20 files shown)


crates/viewer/re_renderer/shader/world_grid.wgsl
```diff
@@ -0,0 +1,154 @@
...
+    let cardinal_grid_closeness_fade = linearstep(2.0, 10.0, line_density * CARDINAL_LINE_FACTOR); // Fade cardinal lines a little bit earlier (because it looks nicer)
+    cardinal_line_intensity *= cardinal_grid_closeness_fade;
+
+    // Combine all lines.
+    //
+    // Lerp for cardinal & regular.
+    // This way we don't break anti-aliasing (as addition would!), mute the regular lines, and make cardinals weaker when there's no regular to support them.
+    let cardinal_and_regular = mix(intensity_regular, cardinal_line_intensity, 0.4);
+    // X and Y are combined like akin to premultiplied alpha operations.
+    let intensity_combined = saturate(cardinal_and_regular.x * (1.0 - cardinal_and_regular.y) + cardinal_and_regular.y);
```
Wumpf, line 128: tbh not so sure about this anymore. Sounded like a compelling idea in Ben's article, but looking at the close-ups I actually don't like the change in color when grid lines meet.
So maybe more like

```suggestion
    let intensity_combined = max(cardinal_and_regular.x, cardinal_and_regular.y);
```

It's very subtle though either way.

crates/viewer/re_renderer/src/renderer/world_grid.rs
```diff
@@ -0,0 +1,207 @@
...
+    render_pipeline: GpuRenderPipelineHandle,
+    bind_group_layout: GpuBindGroupLayoutHandle,
+}
+
+/// Debug overlay for quick & dirty display of texture contents.
+///
+/// Executed as part of the composition draw phase in order to allow "direct" output to the screen.
+///
+/// Do *not* use this in production!
+/// See also `debug_overlay.wgsl` - you are encouraged to edit this shader for your concrete debugging needs!
```
Wumpf, line 74: c&p mistake

Wumpf:
view angle fade is one of the things that isn't holding up: The implementation here is too simplistic. Also the "line density aware fade" is taking care of this well enough

crates/viewer/re_renderer/shader/world_grid.wgsl
```diff
@@ -0,0 +1,154 @@
...
+    thickness_ui: f32,
+}
+
+@group(1) @binding(0)
+var<uniform> config: WorldGridUniformBuffer;
+
+// See world_grid::GridPlane
+const ORIENTATION_XZ: u32 = 0;
+const ORIENTATION_YZ: u32 = 1;
+const ORIENTATION_XY: u32 = 2;
```
emilk, line 22: extremely nitty nit but, as usual I can't help myself: why this order?

It would make more sense to me if the integer indicated the up-axis:

YZ = 0
ZX = 1
XY = 2

crates/viewer/re_renderer/shader/world_grid.wgsl
```diff
@@ -0,0 +1,154 @@
...
+
+struct VertexOutput {
+    @builtin(position)
+    position: vec4f,
+
+    @location(0)
+    scaled_world_plane_position: vec2f,
+};
+
+const PLANE_SCALE: f32 = 10000.0;
```
emilk, line 32: meanig

Is this the number of grid lines in each direction or somesuch?

crates/viewer/re_renderer/shader/world_grid.wgsl
```diff
@@ -0,0 +1,154 @@
...
+// Spans a large quad where centered around the camera.
+//
+// This gives us the "canvas" to drawn the grid on.
+// Compared to a fullscreen pass, we get the z value (and thus early z testing) for free,
+// as well as never covering the screen above the horizon.
+@vertex
+fn main_vs(@builtin(vertex_index) v_idx: u32) -> VertexOutput {
+    var out: VertexOutput;
+
+    var plane_position = (vec2f(f32(v_idx / 2u), f32(v_idx % 2u)) * 2.0 - 1.0) * PLANE_SCALE;
```
emilk, line 43: Seems like `PLANE_SCALE` is the maximum radius of the area covered by grid lines, i.e. half the distance between the furthest parallel lines?

Do you have a plan for how to handle big scenes (e.g. flight traffic data in meter units)?

crates/viewer/re_renderer/shader/world_grid.wgsl
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// Like smoothstep, but linear.
+// Used for antialiasing: smoothstep works as well but is subtly worse.
+fn linearstep(edge0: f32, edge1: f32, x: f32) -> f32 {
+    return saturate((x - edge0) / (edge1 - edge0));
+}
+fn linearstep2(edge0: vec2f, edge1: vec2f, x: vec2f) -> vec2f {
+    return saturate((x - edge0) / (edge1 - edge0));
+}
```
emilk, line 76: It would be nice to have some of these helpers in a `math.wgsl` or similar.

crates/viewer/re_renderer/shader/world_grid.wgsl
```diff
@@ -0,0 +1,154 @@
...
+    // We're not actually implementing the "pristine grid shader" which is a grid with world space thickness,
+    // but rather the pixel space grid, which is a lot simpler, but happens to be also described very well in this article.
+
+    // Distance to a grid line in x and y ranging from 0 to 1.
+    let distance_to_grid_line = calc_distance_to_grid_line(in.scaled_world_plane_position);
+
+    // Figure out the how wide the lines are in this "draw space".
+    let plane_unit_pixel_derivative = fwidthFine(in.scaled_world_plane_position);
+    let line_anti_alias = plane_unit_pixel_derivative;
+    let width_in_pixels = 1.0;//config.thickness_ui * frame.pixels_from_point;
```
emilk, line 95: What's with the dead code?

crates/viewer/re_renderer/shader/world_grid.wgsl
```diff
@@ -0,0 +1,154 @@
...
+    let width_in_grid_units = width_in_pixels * plane_unit_pixel_derivative;
+    var intensity_regular = linearstep2(width_in_grid_units + line_anti_alias, width_in_grid_units - line_anti_alias, distance_to_grid_line);
+
+    // Fade lines that get too close to each other.
+    // Once the number of pixels per line (== from one line to the next) is below a threshold fade them out.
+    //
+    // Note that `1/plane_unit_pixel_derivative` would give us more literal pixels per line,
+    // but we actually want to know how dense the lines get here so we use `1/width_in_grid_units` instead,
+    // such that a line density of 1.0 means roughly "100% lines" and 10.0 means "Every 10 pixels there is a lines".
```
emilk, line 104: This seems inverted to me. A line density of 0.1 should mean a line every 10 pixels (or ui units).
The inversion of density is something like "screen space line spacing"

crates/viewer/re_renderer/shader/world_grid.wgsl
```diff
@@ -0,0 +1,154 @@
...
+    //
+    // Tried smoothstep here, but didn't feel right even with lots of range tweaking.
+    let line_density = 1.0 / max(width_in_grid_units.x, width_in_grid_units.y);
+    let grid_closeness_fade = linearstep(1.0, 10.0, line_density);
+    intensity_regular *= grid_closeness_fade;
+
+    // Every tenth line is a more intense, we call those "cardinal" lines.
+    // Experimented previously with more levels of cardinal lines, but it gets too busy:
+    // It seems that if we want to go down this path, we should ensure that there's only two levels of lines on screen at a time.
+    const CARDINAL_LINE_FACTOR: f32 = 10.0;
```
emilk, line 115: For a reference: in `egui_plot` I have an infinite level of lines, but with only three levels viisble at any time, with a smooth fade to avoid discontinuities: https://emilk.github.io/egui_plot/

This means no matter how small or big your plots are, you always have thicker and thinner lines to guide you

crates/viewer/re_renderer/shader/world_grid.wgsl
emilk: > `let intensity_combined = saturate(cardinal_and_regular.x * (1.0 - cardinal_and_regular.y) + cardinal_and_regular.y);`

Screenshot 2024-11-27 at 21 37 30

----

> `let intensity_combined = max(cardinal_and_regular.x, cardinal_and_regular.y);`

Screenshot 2024-11-27 at 21 36 58


----

I think the first is better at this particular angle (less [Moiré](https://xkcd.com/1814/)). Hard to test without camera control

crates/viewer/re_renderer/src/draw_phases/mod.rs
```diff
@@ -32,6 +32,9 @@ pub enum DrawPhase {
     /// Background, rendering where depth wasn't written.
     Background,
 
+    /// Transparent objects, performing reads of the depth buffer, but no writes.
+    Transparent,
```
emilk, line 36: Great to have this sorted too :P

crates/viewer/re_renderer_examples/2d.rs
```diff
@@ -246,8 +246,8 @@ impl framework::Example for Render2D {
                 .add_points_2d(&positions, &sizes, &colors, &picking_ids);
         }
 
-        let line_strip_draw_data = line_strip_builder.into_draw_data().unwrap();
-        let point_draw_data = point_cloud_builder.into_draw_data().unwrap();
+        let line_strip_draw_data = line_strip_builder.into_draw_data()?;
+        let point_draw_data = point_cloud_builder.into_draw_data()?;
```
emilk, line 250: yay, finally fewer `.unwrap`s in examples!

crates/viewer/re_renderer_examples/world_grid.rs
```diff
@@ -0,0 +1,134 @@
+//! Demonstrates world grid rendering.
+//!
+//! Controls:
+//! - Space: Pause
+//! - G: Toggle camera mode
```
emilk, line 5: It would be really useful to control the camera; at least the angle of it

emilk (approved):
So beautiful 🤩 
sparkle-eyes

crates/viewer/re_renderer/shader/world_grid.wgsl
Wumpf: oh wow yeah so much better, agreed
more visual refinements on the way in the follow-up pr

crates/viewer/re_renderer/shader/world_grid.wgsl
Wumpf: I just went in like random so that's a good principle 😄

crates/viewer/re_renderer/shader/world_grid.wgsl
Wumpf: ohhh that's why my thickness control broke at some point 🤦

crates/viewer/re_renderer/src/draw_phases/mod.rs
Wumpf: all done, right 😄

crates/viewer/re_renderer/shader/world_grid.wgsl
Wumpf: I'll elaborate in the comment for the plane scale (you're right to point that one out as unclear!): this is the main drawback of the plane approach over the screen space filling one: I have to make up some geometry which then gets a limited size. Putting a too high number here makes things break down because of floating point inaccuracies. But arguably at that point we're potentially doomed either way (see respective ticket on rendering breaking down at some scales)

crates/viewer/re_renderer/shader/world_grid.wgsl
```diff
@@ -0,0 +1,154 @@
+#import <./global_bindings.wgsl>
+
+struct WorldGridUniformBuffer {
+    color: vec4f,
+
+    /// A value of [`super::GridPlane`]
+    orientation: u32,
```
emilk, line 7: Come to think of it; it would be better if we could fully specify the plane using the plane equation. That would allow us to use this shader with a general `Plane3D` primitive in Rust

crates/viewer/re_renderer/shader/world_grid.wgsl
emilk: haha 😆

crates/viewer/re_renderer/shader/world_grid.wgsl
Wumpf: hmmm yeah why not I guess. Won't expose it in the viewer though this way since imho there we want a simple interface for the view properties

crates/viewer/re_renderer_examples/world_grid.rs
Wumpf: #1426

crates/viewer/re_renderer/shader/world_grid.wgsl
Wumpf: actually... maybe if we have a ui to pick the axis plane easily, then why not all the way hmmmm. Undecided. To be dicussed on the next pr

PR #8234, opened 2024-11-27
World grid part 2/2: Integrate into Viewer

Wumpf:
### Related

* Follow-up of part 1: https://github.com/rerun-io/rerun/pull/8230
* Fixes https://github.com/rerun-io/rerun/issues/872

### What

Adds new line grid property to the 3D view and enables it by default.

Furthermore...
* improves the grid shader with infinite cardinal lines while showing only two types of "cardinalitities"
    * this is simply based on log10(camera distance from plane) and then blends between two levels of grid
* improve grid's handling of extreme zoom levels. Still some instability on WebGL, but overall much more robust (less flickering/cutoff) 
* Introduce a general `Plane3D` component which is meant for later reuse in other contexts 

The new view property is best described its fbs:
```rust
/// Configuration for the 3D line grid.
table LineGrid3D (
) {
    /// Whether the grid is visible.
    ///
    /// Defaults to true.
    visible: rerun.blueprint.components.Visible ("attr.rerun.component_optional", nullable, order: 1000);

    /// Space between grid lines spacing of one line to the next in scene units.
    ///
    /// As you zoom out, successively only every tenth line is shown.
    /// This controls the closest zoom level.
    spacing: rerun.blueprint.components.GridSpacing ("attr.rerun.component_optional", nullable, order: 2000);

    /// In what plane the grid is drawn.
    ///
    /// Defaults to whatever plane is determined as the plane at zero units up/down as defined by [components.ViewCoordinates] if present.
    plane: rerun.components.Plane3D ("attr.rerun.component_optional", nullable, order: 3000);

    /// How thick the lines should be in ui units.
    ///
    /// Default is 1.0 ui unit.
    stroke_width: rerun.components.StrokeWidth ("attr.rerun.component_optional", nullable, order: 5000);

    /// Color used for the grid.
    ///
    /// Transparency via alpha channel is supported.
    /// Defaults to a slightly transparent light gray.
    color: rerun.components.Color ("attr.rerun.component_optional", nullable, order: 6000);
}

```

Properties in the viewer in default selection panel width
image

New plane controls expanded:
image

Examples of the grid in action (in default settings):
image
image

Bad case - grid is not fine enough and through the scene:
image
Bad case - grid is above the scene.
image

## Testing 

* [x] Metal
* [x] Vulkan
* [x] WebGPU
* [x] WebGL

##  Future work

* tweak some examples to disable the grid or change its spacing
* Extend grid to 2d scenes


crates/store/re_types/definitions/rerun/blueprint/archetypes/line_grid_3d.fbs
```diff
@@ -0,0 +1,33 @@
...
+table LineGrid3D (
+    "attr.rerun.scope": "blueprint"
+) {
+    /// Whether the grid is visible.
+    ///
+    /// Defaults to true.
+    visible: rerun.blueprint.components.Visible ("attr.rerun.component_optional", nullable, order: 1000);
+
+    /// Space between grid lines spacing of one line to the next in scene units.
+    spacing: rerun.blueprint.components.GridSpacing ("attr.rerun.component_optional", nullable, order: 2000);
```
abey79, line 13: The data model lgtm overall 👍🏻 The spacing is slightly ambiguous? Is that the small lines or big lines. For some reason, my (randomly picked) default mental model was that the spacing refers to the big lines, but looking at the screenshots it seems that it's for the small lines. Maybe just clarify that in the doc string.

And then we're good. (That is, until someone requests arbitrary sub-line count to cover for inch/feet/whatever.

Wumpf: it's about the small lines. The big lines are right now a hardcoded visual gimmick to deal with zooming out, shows every 10th line
but yeah I can totally see where that confusion comes from!

nikolausWest:
Small bug: it's now possible to set alpha in the color picker for the Lines3D override (which obviously doesn't work)

Screenshot 2024-11-29 at 15 02 01

nikolausWest:
Just clicking through some examples it's clear that the grid doesn't end up at the "right" height for most of them. Wonder if it's worth explicitly settign the plane height to make those look good?

Wumpf:
yes, we should likely do some adjustments there on a per example basis. For some examples we may also want to turn it off

crates/viewer/re_space_view_spatial/src/view_3d.rs
```diff
@@ -421,6 +422,8 @@ impl SpaceViewClass for SpatialSpaceView3D {
 
         re_ui::list_item::list_item_scope(ui, "spatial_view3d_selection_ui", |ui| {
             view_property_ui::<Background>(ctx, ui, view_id, self, state);
+            //view_property_ui::<LineGrid3D>(ctx, ui, view_id, self, state);
```
abey79, line 425: dead code

crates/viewer/re_space_view_spatial/src/view_3d.rs
```diff
@@ -443,3 +446,84 @@ impl SpaceViewClass for SpatialSpaceView3D {
         self.view_3d(ctx, ui, state, query, system_output)
     }
 }
+
+fn view_property_ui_grid3d(
```
abey79, line 450: worth a quick comment to state why is it that `view_property_ui()` wasn't good enough here

crates/viewer/re_space_view_spatial/src/view_3d.rs
```diff
@@ -443,3 +446,84 @@ impl SpaceViewClass for SpatialSpaceView3D {
...
+            property.archetype_name
+        ));
+        return;
+    };
+
+    let query_ctx = property.query_context(ctx, view_state);
+    let sub_prop_ui = |ui: &mut egui::Ui| {
+        for field in &reflection.fields {
+            // TODO(#1611): The color picker for the color component doesn't show alpha values so far since alpha is almost never supported.
+            // Here however, we need that alpha color picker!
```
abey79, line 474: Note: another short term quick win by rolling tagged component to view property. We could make the component ui editor white list some Archetype/Field for the transparency ui.

docs/snippets/all/views/spatial3d.py
```diff
@@ -22,6 +22,15 @@
         name="3D Scene",
         # Set the background color to light blue.
         background=[100, 149, 237],
+        # Configure the line grid.
+        line_grid=rrb.archetypes.LineGrid3D(
+            visible=True,  # The grid is enabled by default, but you can hide it with this property.
+            spacing=0.1,  # Makes the grid more fine-grained.
+            # By default, the plane is inferred from view coordinates setup, but you can set arbitrary planes.
+            plane=rr.components.Plane3D((0, 0, 1), -5.0),
```
abey79, line 30: Tempted to switch  to `rr.components.Plane3D.XY` here for visibility. Astute users will infer that `Plane3D` might have a constructor for more custom things.

abey79 (approved):
Awesome!

docs/snippets/all/views/spatial3d.py
Wumpf: I'll add it to a comment - this scene is just so well suited for a a -5 offset :/

Wumpf: or maybe add an offset method to it to change it!

abey79: yes, I thought the same!

Issue #8056, opened 2024-11-11
Refactor viewport blueprint logic


emilk:
The viewport blueprint code is a mess, and modifying it is extremely error prone

* This is blocking: #3135 
* Related: https://github.com/rerun-io/rerun/issues/4687
* Related: https://github.com/rerun-io/rerun/issues/8051

No one is to blame - this has evolved over time - but we need to untangle this.

## How and why is it messy?
I don't fully understand all of it.

### Triplicate state
* As components in the blueprint store (persisted)
* As `egui_tiles::Tree` (created at the start of each frame)
* As `struct ViewportBlueprint` (created at the start of each frame)

### Existence of blueprint entities is significant
We use `SystemCommand::DropEntity` to permanently remove space-view entities from the blueprint store. This means we cannot undo the removal of a space-view.

Our loader should instead start with the `RootContainer` and recuse down, and ignore all entities that aren't reachable.
When I tried to implement this however, I created more trouble for myself.

### Views
We view the `ViewportBlueprint` it in two places:
* The `BlueprintTree` panel (left)
* The viewport panel (center)

If we only mutated the state at the end/beginning of the frame, this wouldn't be a problem,
but I suspect we actually mutate state _between_ calling these two views, which of course causes problems.

### Auto-layout
We want to automatically create space-views if there is no explicit blueprint, leading to complicated state like `auto_layout` and `auto_space_views`:
https://github.com/rerun-io/rerun/blob/31a166f48e89626184e8800ab749b123154c6b31/crates/viewer/re_viewport_blueprint/src/viewport_blueprint.rs#L44-L54

### Delayed action
Almost all edits of the state is delayed, often in multiple steps.

This is because of (at least) two reasons (as far as I can tell)
* we cannot modify a `egui_tiles::Tree` while we're iterating over it
* when modifying the state, we want to modify all THREE copies of it… and the best way to do that is to wait until the end of the frame so we only edit the persisted store state

Delays happen in
* [`TreeAction`](https://github.com/rerun-io/rerun/blob/31a166f48e89626184e8800ab749b123154c6b31/crates/viewer/re_viewport_blueprint/src/tree_actions.rs)
* A `TreeAction` can however lead to another frame-delay: https://github.com/rerun-io/rerun/blob/31a166f48e89626184e8800ab749b123154c6b31/crates/viewer/re_viewport/src/viewport.rs#L415-L421

abey79:
[...]

Issue #4687, opened 2024-01-05
Marking of edited for the blueprint is error prone


jleibs:
There are several places where viewport edit-tracking is done manually.

This carries a significant risk that a logic error can result in a never-ending sequence of blueprint re-saving. This can happen if
(1) the edit marking happens without a user interaction
(2) the saved state as a result of the edit does not result in a new state where the edit is no longer marked.

emilk:
Related:
* https://github.com/rerun-io/rerun/issues/2142

Issue #3135, opened 2023-08-29
Undo (and redo) in viewer


roym899:
**Is your feature request related to a problem? Please describe.**
Sometimes I accidentally remove a space view, making it quite annoying to set it up again.
Similarly, sometimes I move a view around which destroys the tiling, but there's no way to undo without setting everything up from scratch.

**Describe the solution you'd like**
Undoing changes in the viewer (like removing and moving a view) with Ctrl+Z.

Redo would be nice to have as well, but not as important.

Related to #1132.

emilk:
* Blocked by https://github.com/rerun-io/rerun/issues/8056

PR #8241, opened 2024-11-28
Refactor convoluted `ViewportBlueprint` logic

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/8056
* Closes https://github.com/rerun-io/rerun/issues/4687
* Unblocks #3135 

### What

This refactors `Viewport` (now called `ViewportUi`) and `ViewportBlueprint` to much simpler.

* The tree is no longer duplicated in `ViewportUi`
* `TreeAction` is now called `ViewportCommand`
* All mutation are now accumulated in an `Arc<Mutex<Vec<ViewportCommand>>>`
* All mutations are applied at the end of the frame, and then stored to blueprint store
* Removed unused component `IncludedSpaceView`
* A lot of cleanup


crates/viewer/re_viewport_blueprint/src/viewport_blueprint.rs
```diff
@@ -12,17 +13,24 @@ use re_entity_db::EntityPath;
...
 };
 
-use crate::{container::ContainerBlueprint, SpaceViewBlueprint, TreeAction, VIEWPORT_PATH};
+use crate::{container::ContainerBlueprint, SpaceViewBlueprint, ViewportCommand, VIEWPORT_PATH};
 
 // ----------------------------------------------------------------------------
 
 /// Describes the layout and contents of the Viewport Panel.
+///
+/// The is loaded from the blueprint store at the start of each frame.
```
Wumpf, line 28:
```suggestion
/// This datastructure is loaded from the blueprint store at the start of each frame.
```

PR #8244, opened 2024-11-28
Fix a bunch of 1.83.0 clippy stuff

emilk:
`cargo +1.83.0 clippy --fix` did most of the work; I did some of it manually


Wumpf (approved):
lgtm!

PR #8248, opened 2024-11-29
Add a snapshot test for list items

abey79:
### What

snapshot:


crates/viewer/re_ui/tests/list_item_tests.rs
```diff
@@ -0,0 +1,205 @@
...
+    let mut harness = egui_kittest::Harness::builder()
+        .with_size(Vec2::new(700.0, 700.0))
+        .build(|ctx| {
+            egui::SidePanel::right("right_panel").show(ctx, |ui| {
+                ui.set_width(650.0);
+                ui.set_max_width(650.0);
+                re_ui::apply_style_and_install_loaders(ui.ctx());
+                list_item::list_item_scope(ui, "list_item_scope", &mut test_code);
+            });
+        });
```
emilk, line 200: Food for thought: it would be nice if we could set things up so that some screenshots tests (like this one) would also be actually runnable (for additional manual testing).

I don't know how (maybe @lucasmerlin has a good idea?), but it would be cool.
In this case I guess the test could actually call some piece of code shared with `re_ui_example`.

Anyways, not necessary for this PR obviously, just an interesting idea imho

emilk (approved):

crates/viewer/re_ui/tests/list_item_tests.rs
lucasmerlin: Gave this a try, I don't think this will be easily possible 🥲

```
on macOS, `EventLoop` must be created on the main thread!
```
But maybe we could use https://github.com/emilk/eterm? 😄 Perfect usecase for that

PR #8250, opened 2024-11-29
Fix typo in archetypes.md

eliemichel:
### Related

N/A

### What

Really just a typo


emilk:
Unfortunately you only fixed it in a generated file (see https://github.com/rerun-io/rerun/blob/bca6c9804ea9dafa57404a769f6133bce8244712/docs/content/reference/types/archetypes.md#L5).

Proper fix coming in: https://github.com/rerun-io/rerun/pull/8253

eliemichel:
Oh, in fact I just clicked on this link:
image Openning #8259 to report this problem!

Issue #8245, opened 2024-11-28
Use a software renderer for kittest


abey79:
Kittest-based UI snapshot test currently use whatever hardware backend wgpu picks, which has multiple drawbacks:
- each backend may produce a slightly different output, making comparison more difficult
- a hardware backed is required, which rules out windows and linux GHA runners

Using a software rendering backend would solve both of these issues.

emilk:
> each backend may produce a slightly different output, making comparison more difficult

According to @Wumpf this is likely due to differences in filtering. If we use a manual four-tap bilinear filter in the egui pixel shader we should get a lot closer results, hopefully within the error margins.

Wumpf:
> Using a software rendering backend would solve both of these issues.

Unfortunately not true since software renderer among each other also have different results. E.g. llvmpipe won't give use the same as microsoft's WARP

emilk:
So we're back to the manual-texture-filtering idea then?

Wumpf:
That would surely alleviate the issue(s), but we'll for sure never catch all differences anyways 🤷. I don't think there's a remedy for tweaking error thresholds

PR #8252, opened 2024-11-29
Run snapshot tests everywhere but render/compare only on mac

abey79:
### Related

- Related to #8245 

### What

Run snapshot tests everywhere but render/compare only on mac

### DIFF
--- a/crates/viewer/re_time_panel/tests/time_panel_tests.rs
+++ b/crates/viewer/re_time_panel/tests/time_panel_tests.rs
@@ -12,7 +12,6 @@ use re_viewer_context::test_context::TestContext;
 use re_viewport_blueprint::ViewportBlueprint;
 
 #[test]
-#[cfg_attr(not(target_os = "macos"), ignore)]
 pub fn time_panel_two_sections_should_match_snapshot() {
     TimePanel::ensure_registered_subscribers();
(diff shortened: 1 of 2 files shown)


abey79:
@rerun-bot full-test

nikolausWest:
What is the workflow then for generating a new snapshot if you're not on Mac? Is it run CI, manually download the right artifact, check it in, push, run CI again?

abey79:
> What is the workflow then for generating a new snapshot if you're not on Mac? Is it run CI, manually download the right artifact, check it in, push, run CI again?

Well I'm not sure we can even recover the artefact from CI now, we'd need to add that. You _could_ generate a reference snapshot locally by temporarily removing the mac-only flag. The snapshot will likely have slight differences, but I believe the comparison allows for that.

Anyways, all of this is very painful, so we basically need #8245. I'm adding it to next cycle, chances are we can sort it out in a time-boxed pairing session with @Wumpf.

lucasmerlin:
> Well I'm not sure we can even recover the artefact from CI now, we'd need to add that.

Should be quite simple to add this: 
https://github.com/emilk/egui/blob/master/.github/workflows/rust.yml#L247-L253

emilk (approved):

Issue #3941, opened 2023-10-20
Exporting visualization for publications and presentations


hovren:
As a researcher I not only need to visualize results during development, but I also need to present them to the scientific community.

The type of results I need to show are
- Static plots/images that can be used in printed articles but also on presentations
- Movies for use in presentations or web pages

I would like to see built-in tools for producing publication and presentation grade media.
Some things to consider:
- The simplest solution would be to add controls to capture a playing visualization as a movie, or static images, as it is seen in the viewer.
- A mechanism to control the visualization using code (or configuration) is needed. Maybe you need to visualize different data using the same settings or you simply want to be able to script generation of figures for reproducibility. I envision some kind of API/tools to control e.g. camera views in 3D-plots, zoom-levels for 2D images, etc.
- Where applicable prefer vector-type graphics output (SVG/PDF)

emilk:
Related:
* https://github.com/rerun-io/rerun/issues/2305

Nik-V9:
Hi, it would be great if this feature can be supported.

I was also wondering if there is a functionality to download images from the viewer?

Wumpf:
@Nik-V9 it's possible to download images that were logged previously logged in the upcoming release. But none of the export functionalities described in this ticket are in yet.

Nik-V9:
Thanks for the quick response! 

It would be much appreciated if you could consider the above request. Rerun has been amazing for debugging and visualizing 3D stuff. I would love to use it for publication visualizations also!

emilk:
I think we could implement this pretty quickly using https://docs.rs/egui/latest/egui/viewport/enum.ViewportCommand.html#variant.Screenshot and then cropping out the content of the space view.

We should probably improve egui to include some sort of description to the screenshot command though (e.g. a `Box<dyn Any>`) that is returned with the screenshot, so we can match the resulting screenshot to the right action ("screenshot _this_ spaceview")

PR #8258, opened 2024-11-29
egui-based view screenshotting

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/3941

### What
Save a screenshot of a view via egui.

The previous attempt (behind a feature-flag) only captures things that were rendered with `re_renderer`, i.e. dit not include any egui elements, like labels.

The new approach should work with any view, anywhere.

### TODO
* [ ] Merge https://github.com/emilk/egui/pull/5416
* [ ] Implement on web
* [ ] Implement save-to-disk
* [ ] Implement for all views (requires a design for the UI of this!)


nikolausWest:
Screenshot 2024-11-29 at 16 48 49

I'd the screenshot option on this menu

Wumpf:
right click on the tab unfortunately selects the view, meaning that one can't have selection highlights on the screenshot

crates/viewer/re_context_menu/src/actions/screenshot_action.rs
```diff
@@ -0,0 +1,78 @@
...
+use crate::{ContextMenuAction, ContextMenuContext};
+
+/// Space view screenshot action.
+#[cfg(not(target_arch = "wasm32"))]
+pub enum ScreenshotAction {
+    /// Scrteenshot the space view, and copy the results to clipboard.
+    CopyScreenshot,
+
+    /// Scrteenshot the space view, and save the results to disk.
+    SaveScreenshot,
```
Wumpf, line 14:
```suggestion
    /// Screenshot the space view, and copy the results to clipboard.
    CopyScreenshot,

    /// Screenshot the space view, and save the results to disk.
    SaveScreenshot,
```

crates/viewer/re_context_menu/src/actions/screenshot_action.rs
```diff
@@ -0,0 +1,78 @@
+use re_viewer_context::{
+    Item, PublishedSpaceViewInfo, ScreenshotTarget, SpaceViewId, SpaceViewRectPublisher,
+};
+
+use crate::{ContextMenuAction, ContextMenuContext};
+
+/// Space view screenshot action.
```
Wumpf, line 7:
```suggestion
/// View screenshot action.
```

...one of these days I'll kill all the `space`. Any moment now :smile:

emilk: Is there an issue/motivation for this? And why not just do it? Seems like an easy search-and-replace?

crates/viewer/re_space_view_spatial/src/ui_2d.rs
```diff
@@ -249,12 +246,6 @@ impl SpatialSpaceView2D {
 
         // ------------------------------------------------------------------------
```
Wumpf, line 247: for later consideration: it was actually nice to have a context menu _inside_ the view.

crates/viewer/re_viewer_context/src/lib.rs
```diff
@@ -35,58 +35,59 @@ mod viewer_context;
 // TODO(andreas): Move to its own crate?
 pub mod gpu_bridge;
 
-pub use annotations::{
-    AnnotationMap, Annotations, ResolvedAnnotationInfo, ResolvedAnnotationInfos,
+pub use self::{
```
Wumpf, line 38: your ide went off the rails here? Doesn't make sense to group things under `pub use self::`.

Wumpf (approved):
Looks all good and straight forward.

the only thing that I didn't quite get is what removes the `SpaceViewRectPublisher` again from the cache for a view that's hidden. But it works :shrug:

crates/viewer/re_context_menu/src/actions/screenshot_action.rs
Wumpf: No, issue for this, no.
Many many months ago this is something that we concluded together with @nikolausWest . I think @abey79 was involved as well. The rationale is that "space view" is unnecessarily confusing over just "view" and sometimes leads to fairly lengthy names.
Ever since we slowly moved towards removing "space" from types and comments (well.. at least Antoine and me did). We never got around to doing the whole thing because it touches an enormous amount of types & code, a few of which are part of the API.

emilk:
> the only thing that I didn't quite get is what removes the `SpaceViewRectPublisher` again from the cache for a view that's hidden. But it works 🤷

egui does is for us :)

PR #8261, opened 2024-11-30
Use video logging api in `detect_and_track` example

oxkitsune:
### What

This updates the `detect_and_track` example to use the new video api.


oxkitsune:
Curious to hear what your thoughts are on this approach, since it's still logging rescaled `Image` entities. I was thinking of rescaling the model outputs to the video's resolution so we can log`VideoFrameReference` instead and save memory.

teh-cmc (approved):
Love it! We need more real-world video examples, thanks!

Before:
image

After:
image

PR #8292, opened 2024-12-03
gRPC spec update: make metadata APIs more like the data API

zehiko:
### What

As part of simplifying metadata APIs and as pre-req for creating gRPC catalog data source, we:
* expose only single API for querying metadata (that supports filtering and projection)
* this API now returns a stream of chunks, same as the data API

### Testing done
* integrated changes into SN


crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -94,21 +77,30 @@ enum EncoderVersion {
...
 }
 
 message ColumnProjection {
     repeated string columns = 1;
 }
 
-message ListRecordingsResponse {
-    repeated RecordingMetadata recordings = 1;
+message CatalogFilter {
+    // filter out specific recordings
```
jleibs, line 96: Let's be explicit about if this is an inclusion or exclusion filter. My expectation would be include but "filter out" sounds like exclude.

jleibs (approved):

crates/store/re_protos/proto/rerun/v0/remote_store.proto
zehiko: good point, one word makes a semantic world of difference :)

PR #8295, opened 2024-12-03
Fix web viewer feature flags

### LINKED ISSUE #8294: `map-view` feature not enabled in web-viewer
Not on any PR builds, nor on 0.20.2 viewer at https://rerun.io/viewer?url=https%3A%2F%2Fapp.rerun.io%2Fversion%2F0.20.2%2Fexamples%2Fnuscenes_dataset.rrd

jprochazk:
### Related

* Closes #8294

### What

* `build-web-viewer` now propagates `--no-default-features`, and doesn't include it by default
* Made `map_view` a default feature in `re_viewer`

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -175,13 +175,13 @@ rerun-web = { cmd = "cargo run --package rerun-cli --no-default-features --featu
 #
 # This installs the `wasm32-unknown-unknown` rust target if it's not already installed.
 # (this looks heavy but takes typically below 0.1s!)
-rerun-build-web = "rustup target add wasm32-unknown-unknown && cargo run -p re_dev_tools -- build-web-viewer --features analytics,grpc,map_view --debug"
+rerun-build-web = "rustup target add wasm32-unknown-unknown && cargo run -p re_dev_tools -- build-web-viewer --features analytics,grpc --debug"
 
 # Compile the web-viewer wasm and the cli.
 #
 # This installs the `wasm32-unknown-unknown` rust target if it's not already installed.
 # (this looks heavy but takes typically below 0.1s!)
-rerun-build-web-cli = "rustup target add wasm32-unknown-unknown && cargo run -p re_dev_tools -- build-web-viewer --features analytics,grpc,map_view --debug && cargo build --package rerun-cli --no-default-features --features web_viewer"
+rerun-build-web-cli = "rustup target add wasm32-unknown-unknown && cargo run -p re_dev_tools -- build-web-viewer --features analytics,grpc --debug && cargo build --package rerun-cli --no-default-features --features web_viewer"
 
 # Compile and run the web-viewer in release mode via rerun-cli.
 #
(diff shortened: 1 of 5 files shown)


rerun_js/web-viewer/build-wasm.mjs
```diff
@@ -31,7 +31,8 @@ function buildWebViewer(mode) {
       "cargo run -p re_dev_tools -- build-web-viewer",
       modeFlags,
       "--target no-modules-base",
-      "--features grpc",
+      "--no-default-features",
+      "--features grpc,map_view",
```
emilk, line 35: I suggest we have a `release = ["grpc", "map_view"]` feature in `re_viewer` instead. It feels like a more discoverable location for controlling what goes into the release.

```suggestion
      "--features release",
```

jprochazk: At this point I just made `map_view` the default. I don't agree with the `release` feature for web viewer specifically, it makes sense for rerun-cli because there it makes it possible for it to be compiled without `nasm`, and with `-F release` in case you do want `nasm`.

jprochazk: The big problem here imho was `build_web_viewer` silently passing `--no-default-features` to the `re_viewer` build. To discover all the features being requested, you look through the commands that mention `build-web-viewer`, same as with every other non-web build command.

jprochazk:
Seems to work: https://rerun.io/viewer/pr/8295?url=https%3A%2F%2Fapp.rerun.io%2Fversion%2F0.20.2%2Fexamples%2Fnuscenes_dataset.rrd

rerun_js/web-viewer/build-wasm.mjs
```diff
@@ -31,7 +31,8 @@ function buildWebViewer(mode) {
       "cargo run -p re_dev_tools -- build-web-viewer",
       modeFlags,
       "--target no-modules-base",
-      "--features grpc",
+      "--no-default-features",
+      "--features grpc,map_view", // no `analytics`
```
emilk, line 35: why not?

pixi.toml
```diff
@@ -175,29 +175,29 @@ rerun-web = { cmd = "cargo run --package rerun-cli --no-default-features --featu
...
+rerun-web-release = { cmd = "cargo run --package rerun-cli --no-default-features --features web_viewer,map_view,grpc --release -- --web-viewer", depends_on = [
   "rerun-build-web-release",
 ] }
 
 # Compile the web-viewer wasm in release mode.
 #
 # This installs the `wasm32-unknown-unknown` rust target if it's not already installed.
 # (this looks heavy but takes typically below 0.1s!)
-rerun-build-web-release = "rustup target add wasm32-unknown-unknown && cargo run -p re_dev_tools -- build-web-viewer --features analytics,grpc --release -g"
+rerun-build-web-release = "rustup target add wasm32-unknown-unknown && cargo run -p re_dev_tools -- build-web-viewer --no-default-features --features analytics,grpc,map_view --release -g"
```
emilk, line 200: Isn't easier to keep the default features and just add on grpc?

emilk (approved):
Nice!

Btw, I wonder if we shouldn't make a `re_web_viewer` crate that just wraps `re_viewer`. `re_viewer` is weird in that it is both a library, and web app. Food for thought

emilk (approved):
Nice!

Btw, I wonder if we shouldn't make a `re_web_viewer` crate that just wraps `re_viewer`. `re_viewer` is weird in that it is both a library, and web app. Food for thought

rerun_js/web-viewer/build-wasm.mjs
emilk: Makes sense 👍

rerun_js/web-viewer/build-wasm.mjs
jprochazk: this is for the NPM package, we don't want users' applications which embed the web viewer this way to send analytics to us. they will only do that if they embed the web viewer as an iframe

pixi.toml
jprochazk: I wanted to be more explicit here, because you're not going to type the long command

jprochazk:
> Btw, I wonder if we shouldn't make a re_web_viewer crate that just wraps re_viewer. re_viewer is weird in that it is both a library, and web app. Food for thought

https://github.com/rerun-io/rerun/issues/8297

Issue #4491, opened 2023-12-12
Add breadcrumbs to the selection panel


abey79:
The enclosing context of the selected item should be displayed as breadcrumbs to facilitate the navigation. This issue is about the design and implementation of this feature.

####New context from Dec '24
We are implementing undo, and want to get rid of the selection panel errors navigations ←→. 
Despite the arrows being ambiguous UX element, it is being used to navigate via selected panels (especially when users click somewhere where they did not want to go, so they go 'back' to the previous view). 
Assumption is that breadcrumbs will be a better navigation support for this as well as a more familiar ux pattern. 

#### Design issues
Proposal based on Marten's sketches: 
![Image](https://github.com/user-attachments/assets/21f59f49-7132-4929-88bb-b4172211c8f2)

This will also allow to remove some of the elements in the selection panel as they will become redundant, 
merge `parent` and `in view` in blueprint, remove `in view`

#### Implementation issues

- [ ] implement the capability to walk the hierarchy up
- [ ] implement the breadcrumb in selection panel

cc @jleibs @emilk

gavrelina:
Upd: we decided to use first entity letters instead of entity icons, as they give a better hint on what is the the actual name of the entity
and use a little store icon for the root in the streams

emilk:
* Most of it done in https://github.com/rerun-io/rerun/pull/8319

Still need to properly sync "projection" entities with the blueprint tree panel

jleibs:
> A) I don't know what the logic is for how they are displayed in the blueprint tree view

This happens here: https://github.com/rerun-io/rerun/blob/76d4663de1916bb6d1ee575b2224d848923a00f4/crates/viewer/re_blueprint_tree/src/blueprint_tree.rs#L390-L406

I'll reserve (B) for others to decide

>  (C) If it is a good model to follow, I don't know how to follow it

You can combine the snippet from above, with the logic to resolve a [`DataResult::Item`](https://github.com/rerun-io/rerun/blob/76d4663de1916bb6d1ee575b2224d848923a00f4/crates/viewer/re_viewer_context/src/item.rs#L32-L33)

Following the pattern here: https://github.com/rerun-io/rerun/blob/76d4663de1916bb6d1ee575b2224d848923a00f4/crates/viewer/re_selection_panel/src/selection_panel.rs#L261-L266

emilk:
Closed; next step in:
* https://github.com/rerun-io/rerun/issues/8342

PR #8296, opened 2024-12-03
Remove selection history

emilk:
### What
We're removing it because we're worried users will confuse it with the new Undo feature.

After this PR, you can no longer go back in the selection history.

But, we're planning on adding hierarchy breadcrumbs to the selections instead, which should hopefully be an even better way of understanding and navigating selections:
* https://github.com/rerun-io/rerun/issues/4491

image


emilk:
Hum… I kind of miss it already

abey79 (approved):
Straightforward. I for one won't miss it, but of course...

PR #8302, opened 2024-12-03
Improve the unknown view class text

### LINKED ISSUE #8052: Warn when `Geo*` entities are logged and the viewer wasn't compiled with the `map_view` feature
The map view is behind a `map_view` cargo feature on `rerun-cli`, which is enabled in release build but not in default build (aka when building from source). This leads to a footgun where people don't see a map when they expect one. We should warn the user when that happens, indicating that the `map_view` feature needs to be enabled.

The proposed mechanism for that is to have a store subscribers that reacts to some specific indicator components, and emit warnings as required.

**Level 1**: hard code `GeoPoints` and `GeoLineStrings` (subject to bitrot, eg when we introduce `GeoPolygons`

**Level 2**: use the information already contained in `attribute.docs.view_types` to populate the reflection structures, and use said structure to generate warnings for all kinds of view that might be missing.
.
.
.

**Level ∞**: "unsupported data, click [here](https://www.youtube.com/watch?v=E4WlUXrJgy4) to download the Map view plugin"

abey79:
### Related

- Closes #8052

### What

Update the unknown view ui. Now hints at enabling `map_view` if a map view was expected.

image

**Note**: I tried to make that message depend on the actual identifier that was used (to special case `"map"`), but bailed out if it as it messed too much with the view class registrar for my taste.

### DIFF
--- a/crates/viewer/re_viewer_context/src/space_view/space_view_class_placeholder.rs
+++ b/crates/viewer/re_viewer_context/src/space_view/space_view_class_placeholder.rs
@@ -1,10 +1,11 @@
+use re_types::SpaceViewClassIdentifier;
+use re_ui::UiExt;
+
 use crate::{
     SpaceViewClass, SpaceViewClassRegistryError, SpaceViewSpawnHeuristics, SpaceViewState,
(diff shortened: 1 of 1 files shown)


grtlr (approved):
I like it!

emilk:
Nicely done ❤️

Issue #8293, opened 2024-12-03
Tagged components milestone 3: descriptor-driven and partial queries


teh-cmc:
The third milestone for tagged components is to be able to query (`LatestAt` and `Range`) for components based on descriptor tags and potentially partial information.
The semantics of the viewer must not be impacted in any way by that change.

**All `ComponentName`-based logic must be gone, in favor of `ComponentDescriptor-based logic**.

E.g.:
* `LatestAt("a/b/c@*:Color#*)`, which is what you can do today.
* `LatestAt("a/b/c@LineStrips2D:Color#*)`, to ask specifically for colors belonging to `LineStrips2D`.
* `LatestAt("a/b/c@LineStrips2D:Color#edge_color)`, to ask specifically for the edge colors.

This is the last milestone before people can actually start building features on top of tagged components.

More details TBD.

---

* Part of https://github.com/rerun-io/rerun/issues/6889
* Requires #7948

grtlr:
I think we should change these queries to be based on `archetype_field_name` similar to #9906.

grtlr:
I've created an issue to track this:

* [OSS-1285](https://linear.app/rerun/issue/OSS-1285/change-dataframe-queries-to-use-archetype-field-name-instead-of)

Issue #7948, opened 2024-10-30
Tagged components milestone 1: end-to-end tagging


teh-cmc:
* Part of https://github.com/rerun-io/rerun/issues/6889

The first milestone for tagged components will be for the tags to be carried through end-to-end.
That is:
* You log some data to a recording, using any of our SDKs.
    * The proper tags get automagically injected into the data.
* You then query that recording via the dataframe API.
    * You get the automagically injected tags back.

This implies, among other things, that codegen is properly set up and that all the data path is just smart enough to be able to carry all that information across the entire pipeline.

This does _not_ cover:
* Being able to log the same component multiple times to the same entity, even with different tags.
* Being able to run tag-based queries.
* Using the tags in place of indicators.
* Interned `ComponentDescriptor`s

Each of the following limitations are upcoming miltestones of their own.

The tags we're talking about are the following:
* `rerun.component_name` aka `sorbet.semantic_type`
* `rerun.archetype_name` aka `sorbet.semantic_family`
* `rerun.archetype_field_name` aka `sorbet.logical_type`

PR #8304, opened 2024-12-03
End-to-end tagging: Rust

teh-cmc:
I had to give up on the idea of splitting this thing into neat little PRs -- the enormous amount of extra work needed in this case is just not worth it, it's not even close (turns out changing the definition of `Component` has cascading consequences :no_mouth:).

I'll add a thorough description of what's going on to compensate, and can walk someone through this if needed.

---

### Goals and non-goals

The goal of this PR is to get component tags in, store them, and then get them out.

The goal of this PR is _not_ to port every single bit of component-name based logic to component-descriptor based logic (including but certainly not limited to datastore queries).
That will be the next step: https://github.com/rerun-io/rerun/issues/8293.

### Types and traits

First and foremost, this ofc introduces the new `ComponentDescriptor` type:
```rust
/// A [`ComponentDescriptor`] fully describes the semantics of a column of data.
///
/// Every component is uniquely identified by its [`ComponentDescriptor`].
#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
pub struct ComponentDescriptor {
    /// Optional name of the `Archetype` associated with this data.
    ///
    /// `None` if the data wasn't logged through an archetype.
    ///
    pub archetype_name: Option<ArchetypeName>,

    /// Optional name of the field within `Archetype` associated with this data.
    ///
    /// `None` if the data wasn't logged through an archetype.
    ///
    /// Example: `positions`.
    pub archetype_field_name: Option<ArchetypeFieldName>,

    /// Semantic name associated with this data.
    ///
    /// This is fully implied by `archetype_name` and `archetype_field`, but
    /// included for semantic convenience.
    ///
    /// Example: `rerun.components.Position3D`.
    pub component_name: ComponentName,
}
```

Note that this is a _Rerun_ type, not a _Sorbet_ type: i.e. it uses Rerun terminology (archetypes, fields, etc), not Sorbet terminology.
As is now tradition, this terminology gets translated into its Sorbet equivalent when leaving the land of internal Chunks for the land of external RecordBatches and Dataframes.

`Component`s are now uniquely identified by a `ComponentDescriptor` rather than a `ComponentName`:
```rust
/// A [`Component`] describes semantic data that can be used by any number of [`Archetype`]s.
///
/// Implementing the [`Component`] trait automatically derives the [`ComponentBatch`] implementation,
/// which makes it possible to work with lists' worth of data in a generic fashion.
pub trait Component: Loggable {
    /// Returns the complete [`ComponentDescriptor`] for this [`Component`].
    ///
    /// Every component is uniquely identified by its [`ComponentDescriptor`].
    //
    // NOTE: Builtin Rerun components don't (yet) have anything but a `ComponentName` attached to
    // them (other tags are injected at the Archetype level), therefore having a full
    // `ComponentDescriptor` might seem overkill.
    // It's not:
    // * Users might still want to register Components with specific tags.
    // * In the future, `ComponentDescriptor`s will very likely cover than Archetype-related tags
    //   (e.g. generics, metric units, etc).
    fn descriptor() -> ComponentDescriptor;

    /// The fully-qualified name of this component, e.g. `rerun.components.Position2D`.
    ///
    /// This is a trivial but useful helper for `Self::descriptor().component_name`.
    ///
    /// The default implementation already does the right thing: do not override unless you know
    /// what you're doing.
    /// `Self::name()` must exactly match the value returned by `Self::descriptor().component_name`,
    /// or undefined behavior ensues.
    //
    // TODO(cmc): The only reason we keep this around is for convenience, and the only reason we need this
    // convenience is because we're still in this weird half-way in-between state where some things
    // are still indexed by name. Remove this entirely once we've ported everything to descriptors.
    #[inline]
    fn name() -> ComponentName {
        Self::descriptor().component_name
    }
}
```
`Component::name` still exists for now, as a convenience during the interim (that is, until we propagate `ComponentDescriptor` to every last corner of the app).

`MaybeOwnedComponentBatch` now has the possibility to augment and/or fully-override the `ComponentDescriptor` of the data within:
```rust
/// Some [`ComponentBatch`], optionally with an overridden [`ComponentDescriptor`].
///
/// Used by implementers of [`crate::AsComponents`] to both efficiently expose their component data
/// and assign the right tags given the surrounding context.
pub struct MaybeOwnedComponentBatch<'a> {
    /// The component data.
    pub batch: ComponentBatchCow<'a>,

    /// If set, will override the [`ComponentBatch`]'s [`ComponentDescriptor`].
    pub descriptor_override: Option<ComponentDescriptor>,
}
```
This is a crucial part of the story, as this is how e.g. archetypes inject their own tags when component data gets logged on their behalf.

### Override model

The override model is simple:
* Every `Component` has an associated `ComponentDescriptor`.
* Every `ComponentBatch` inherits from its underlying `Component`'s `ComponentDescriptor`.
* `AsComponents` has an opportunity to override each `ComponentBatch`'s `ComponentDescriptor` (by means of `MaybeOwnedComponentBatch`.

The goal is to try and carry those semantics over the two other SDKs (Python, C++), while somehow keeping changes to a minimum.

### Undefined behavior

Logging the same component multiple times on a single entity (e.g. by logging different archetypes that share parts of their definitions) has always been, for all intents and purposes, UB.

This PR propagates descriptors just enough to get things up and running, no more no less. By which I mean that it is possible to get component tags in and out of the system, but many things still assume that `Component`s are uniquely identified by their names.
This means that some part of the codebase are still indexing things by name, while others index by descriptor. Where these parts meet, what was UB before is even more UB now, as we generally just pick one random component among the ones available.
You'll see a lot of `get_first_component` in the code: every single one of those is UB if there are multiple components under the same name (for now!).

Debug builds assert for duplicated components, until we properly use descriptors everywhere (remember: nothing should ever be indexed by `ComponentName` in the future).

### Fully-qualified component names & column paths

`ComponentDescriptor` defines its fully-qualified name as such:
```rust
match (archetype_name, component_name, archetype_field_name) {
    (None, component_name, None) => component_name.to_owned(),
    (Some(archetype_name), component_name, None) => {
        format!("{archetype_name}:{component_name}")
    }
    (None, component_name, Some(archetype_field_name)) => {
        format!("{component_name}#{archetype_field_name}")
    }
    (Some(archetype_name), component_name, Some(archetype_field_name)) => {
        format!("{archetype_name}:{component_name}#{archetype_field_name}")
    }
}
```
which yields e.g. `rerun.archetypes.Points3D:rerun.components.Position3D#positions`, which is generally shortened to `Points3D:Position3D#positions` when there is no ambiguity.

In the dataframe API, a fully-qualified column path now becomes `{entity_path}@{archetype_name}:{component_name}#{archetype_field_name}`, e.g. `/my/points@rerun.archetypes.Points3D:rerun.components.Position3D#positions` or `/my/points@Points3D:Position3D#positions`.

This syntax needs to be debated. I have intentionally disabled the syntax in the dataframe APIs so as not to break anything external-facing.

### Transport and metadata

`ArchetypeName` and `ArchetypeFieldName` are now exposed as `rerun.archetype_name` and `rerun.archetype_field_name` in `TransportChunk`'s arrow metadata.

I really cannot wait for a better metadata system.

### Performance

`ComponentDescriptor`s add an extra layer of mappings everywhere: we used to have `IntMap<ComponentName, T>` all over the place, now we have `IntMap<ComponentName, IntMap<ComponentDescriptor, T>>`.
The extra `ComponentName` layer is needed because it is very common to want to look for anything matching a `ComponentName`, without any further tags specified.

Like before, these are NoHash maps, so performance impact should be minimal (`ComponentDescriptor` implements NoHash by xor'ing everything).

### Examples / testing / roundtrips

See:
* docs/snippets/all/descriptors/descr_builtin_archetype.rs
* docs/snippets/all/descriptors/descr_builtin_component.rs
* docs/snippets/all/descriptors/descr_custom_archetype.rs
* docs/snippets/all/descriptors/descr_custom_component.rs

These snippets play all roles at once, as usual. In particular they make sure that all languages (well, only Rust for now, Python and C++ coming soon) carry all the right tags in all the right situations.

---

* Part of #7948

### DIFF
--- a/crates/store/re_chunk/src/transport.rs
+++ b/crates/store/re_chunk/src/transport.rs
@@ -13,9 +13,9 @@ use arrow2::{
 use itertools::Itertools;
 use nohash_hasher::IntMap;
 use re_log_types::{EntityPath, Timeline};
-use re_types_core::{Component as _, Loggable as _, SizeBytes};
+use re_types_core::{Component as _, ComponentDescriptor, Loggable as _, SizeBytes};
 
-use crate::{Chunk, ChunkError, ChunkId, ChunkResult, RowId, TimeColumn};
+use crate::{chunk::ChunkComponents, Chunk, ChunkError, ChunkId, ChunkResult, RowId, TimeColumn};
 
 // ---
 
@@ -87,6 +87,12 @@ impl TransportChunk {
 
     /// The value used to identify a Rerun data column in field-level [`Arrow2Schema`] metadata.
     pub const FIELD_METADATA_VALUE_KIND_DATA: &'static str = "data";
+
+    /// The key used to identify the [`crate::ArchetypeName`] in field-level [`Arrow2Schema`] metadata.
+    pub const FIELD_METADATA_KEY_ARCHETYPE_NAME: &'static str = "rerun.archetype_name";
+
+    /// The key used to identify the [`crate::ArchetypeFieldName`] in field-level [`Arrow2Schema`] metadata.
+    pub const FIELD_METADATA_KEY_ARCHETYPE_FIELD_NAME: &'static str = "rerun.archetype_field_name";
 
     /// The marker used to identify whether a column is sorted in field-level [`Arrow2Schema`] metadata.
     ///
@@ -192,6 +198,48 @@ impl TransportChunk {
             ), //
         ]
         .into()
+    }
+
+    #[inline]
+    pub fn field_metadata_component_descriptor(
+        component_desc: &ComponentDescriptor,
+    ) -> Arrow2Metadata {
+        component_desc
+            .archetype_name
+            .iter()
+            .copied()
+            .map(|archetype_name| {
+                (
+                    Self::FIELD_METADATA_KEY_ARCHETYPE_NAME.to_owned(),
+                    archetype_name.to_string(),
+                )
+            })
+            .chain(component_desc.archetype_field_name.iter().copied().map(
+                |archetype_field_name| {
+                    (
+                        Self::FIELD_METADATA_KEY_ARCHETYPE_FIELD_NAME.to_owned(),
+                        archetype_field_name.to_string(),
+                    )
+                },
+            ))
+            .collect()
+    }
+
+    #[inline]
+    pub fn component_descriptor_from_field(field: &ArrowField) -> ComponentDescriptor {
+        ComponentDescriptor {
+            archetype_name: field
+                .metadata
+                .get(Self::FIELD_METADATA_KEY_ARCHETYPE_NAME)
+                .cloned()
+                .map(Into::into),
+            component_name: field.name.clone().into(),
+            archetype_field_name: field
+                .metadata
+                .get(Self::FIELD_METADATA_KEY_ARCHETYPE_FIELD_NAME)
+                .cloned()
+                .map(Into::into),
+        }
     }
 }
 
@@ -383,7 +431,7 @@ impl Chunk {
 
             schema.fields.push(
                 ArrowField::new(
-                    RowId::name().to_string(),
+                    RowId::descriptor().to_string(),
                     row_ids.data_type().clone(),
                     false,
                 )
@@ -438,13 +486,23 @@ impl Chunk {
             re_tracing::profile_scope!("components");
 
             let mut components = components
-                .iter()
-                .map(|(component_name, data)| {
-                    let field =
-                        ArrowField::new(component_name.to_string(), data.data_type().clone(), true)
-                            .with_metadata(TransportChunk::field_metadata_data_column());
+                .values()
+                .flat_map(|per_desc| per_desc.iter())
+                .map(|(component_desc, list_array)| {
+                    let field = ArrowField::new(
+                        component_desc.component_name.to_string(),
+                        list_array.data_type().clone(),
+                        true,
+                    )
+                    .with_metadata({
+                        let mut metadata = TransportChunk::field_metadata_data_column();
+                        metadata.extend(TransportChunk::field_metadata_component_descriptor(
+                            component_desc,
+                        ));
+                        metadata
+                    });
 
-                    let data = data.clone().boxed();
+                    let data = list_array.clone().boxed();
 
                     (field, data)
                 })
@@ -486,7 +544,8 @@ impl Chunk {
             re_tracing::profile_scope!("row ids");
 
             let Some(row_ids) = transport.controls().find_map(|(field, column)| {
-                (field.name == RowId::name().as_str()).then_some(column)
+                // TODO(cmc): disgusting, but good enough for now.
+                (field.name == RowId::descriptor().component_name.as_str()).then_some(column)
             }) else {
                 return Err(ChunkError::Malformed {
                     reason: format!("missing row_id column ({:?})", transport.schema),
@@ -575,7 +634,7 @@ impl Chunk {
 
         // Components
         let components = {
-            let mut components = IntMap::default();
+            let mut components = ChunkComponents::default();
 
             for (field, column) in transport.components() {
                 let column = column
@@ -588,11 +647,12 @@ impl Chunk {
                         ),
                     })?;
 
+                let component_desc = TransportChunk::component_descriptor_from_field(field);
+
                 if components
-                    .insert(
-                        field.name.clone().into(),
-                        column.clone(), /* refcount */
-                    )
+                    .entry(component_desc.component_name)
+                    .or_default()
+                    .insert(component_desc, column.clone() /* refcount */)
                     .is_some()
                 {
                     return Err(ChunkError::Malformed {
@@ -703,7 +763,7 @@ mod tests {
         let colors4 = None;
 
         let components = [
-            (MyPoint::name(), {
+            (MyPoint::descriptor(), {
                 let list_array = crate::util::arrays_to_list_array_opt(&[
                     Some(&*points1),
                     points2,
@@ -714,7 +774,7 @@ mod tests {
                 assert_eq!(4, list_array.len());
                 list_array
             }),
-            (MyPoint::name(), {
+            (MyPoint::descriptor(), {
                 let list_array = crate::util::arrays_to_list_array_opt(&[
                     Some(&*colors1),
                     Some(&*colors2),
--- a/crates/store/re_chunk_store/src/dataframe.rs
+++ b/crates/store/re_chunk_store/src/dataframe.rs
@@ -10,7 +10,9 @@ use itertools::Itertools;
 
 use re_chunk::TimelineName;
 use re_log_types::{ComponentPath, EntityPath, ResolvedTimeRange, TimeInt, Timeline};
-use re_types_core::{ArchetypeName, ComponentName, ComponentNameSet};
+use re_types_core::{
+    ArchetypeFieldName, ArchetypeName, ComponentDescriptor, ComponentName, ComponentNameSet,
+};
 
 use crate::{ChunkStore, ColumnMetadata};
 
@@ -135,7 +137,7 @@ pub struct ComponentColumnDescriptor {
     /// `None` if the data wasn't logged through an archetype.
     ///
     /// Example: `positions`.
-    pub archetype_field_name: Option<String>,
+    pub archetype_field_name: Option<ArchetypeFieldName>,
 
     /// Semantic name associated with this data.
     ///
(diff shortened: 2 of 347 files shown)


crates/store/re_chunk_store/src/dataframe.rs
```diff
@@ -296,12 +310,19 @@ impl ComponentColumnDescriptor {
...
-            format!(
-                "{}:{}",
-                self.entity_path,
-                self.component_name.short_name().to_owned()
-            ),
+            // NOTE: Uncomment this to expose fully-qualified names in the Dataframe APIs!
+            // I'm not doing that right now, to avoid breaking changes (and we need to talk about
+            // what the syntax for these fully-qualified paths need to look like first).
+            format!("{}:{}", entity_path, descriptor.component_name.short_name()),
+            // format!("{entity_path}@{}", descriptor.short_name()),
```
teh-cmc, line 325: As mentioned, I kept the existing column-path syntax around for now, to avoid breaking anything downstream. TBD.

crates/store/re_chunk_store/src/store.rs
```diff
@@ -659,8 +666,9 @@ impl ChunkStore {
         let is_indicator = component_name.is_indicator_component();
 
         use re_types_core::Archetype as _;
-        let is_tombstone =
-            re_types_core::archetypes::Clear::all_components().contains(component_name);
+        let is_tombstone = re_types_core::archetypes::Clear::all_components()
+            .iter()
+            .any(|descr| descr.component_name == *component_name);
```
teh-cmc, line 671: This is one of many many _many_ examples where we still have component-name based logic that needs to be migrated to component-descriptor based logic.

The goal of this PR is to get tags in and out, not migrate everything that exists in one go (that PR is big enough as is!).

crates/store/re_chunk_store/src/writes.rs
```diff
@@ -37,6 +37,14 @@ impl ChunkStore {
             return Ok(Vec::new());
         }
 
+        #[cfg(debug_assertions)]
+        for (component_name, per_desc) in chunk.components().iter() {
+            assert!(
+                per_desc.len() <= 1,
+                "Insert Chunk with multiple values for component named `{component_name}`: this is currently UB",
+            );
+        }
```
teh-cmc, line 46: One day this will be legal, but first we need to remove every single bit of component-name based logic across the entire app.

teh-cmc:
@rerun-bot full-check

teh-cmc:
@rerun-bot full-check

crates/store/re_types_core/src/loggable_batch.rs
```diff
@@ -39,64 +38,131 @@ pub trait ComponentBatch: LoggableBatch {
...
+    fn name(&self) -> ComponentName {
+        self.descriptor().component_name
+    }
+}
+
+/// Some [`ComponentBatch`], optionally with an overridden [`ComponentDescriptor`].
+///
+/// Used by implementers of [`crate::AsComponents`] to both efficiently expose their component data
+/// and assign the right tags given the surrounding context.
+pub struct MaybeOwnedComponentBatch<'a> {
```
teh-cmc, line 65: Maybe at this point this deserves a new name? I don't think it matters all that much, all those APIs will break in the most extraordinary ways once we move to eager serialization anyhow...

teh-cmc: That said, `DescribedComponentBatch` could be made to match what we use on the Python side, and maybe even the C++ side if we get lucky...

teh-cmc:
`cargo machete` passes locally, and `cargo machete --version` matches the CI... I dunno man.

teh-cmc:
I had to give a tour to @Wumpf so that he could help me with the C++ port, so he's probably the best equipped person to review this at this point.

crates/store/re_types_core/src/component_descriptor.rs
```diff
@@ -0,0 +1,177 @@
...
+            archetype_name,
+            archetype_field_name,
+            component_name,
+        } = self;
+
+        let archetype_name = archetype_name.map_or(0, |v| v.hash());
+        let archetype_field_name = archetype_field_name.map_or(0, |v| v.hash());
+        let component_name = component_name.hash();
+
+        state.write_u64(archetype_name ^ archetype_field_name ^ component_name);
```
teh-cmc, line 46: Surely that makes sense, right?

crates/store/re_types_core/src/loggable_batch.rs
Wumpf: agreed. Maybe simpler `ComponentBatchCowWithDesc`?

crates/store/re_types_core/src/loggable.rs
```diff
@@ -147,8 +151,35 @@ pub trait Loggable: 'static + Send + Sync + Clone + Sized + SizeBytes {
...
+    /// Returns the complete [`ComponentDescriptor`] for this [`Component`].
+    ///
+    /// Every component is uniquely identified by its [`ComponentDescriptor`].
+    //
+    // NOTE: Builtin Rerun components don't (yet) have anything but a `ComponentName` attached to
+    // them (other tags are injected at the Archetype level), therefore having a full
+    // `ComponentDescriptor` might seem overkill.
+    // It's not:
+    // * Users might still want to register Components with specific tags.
+    // * In the future, `ComponentDescriptor`s will very likely cover than Archetype-related tags
```
Wumpf, line 163:
```suggestion
    // * In the future, `ComponentDescriptor`s will very likely cover more than Archetype-related tags
```

crates/store/re_types_core/src/loggable.rs
```diff
@@ -147,8 +151,35 @@ pub trait Loggable: 'static + Send + Sync + Clone + Sized + SizeBytes {
...
+    //   (e.g. generics, metric units, etc).
+    fn descriptor() -> ComponentDescriptor;
```
Wumpf, line 165: despite above line of argumentation it does irk me a bit that you can set field names and archetype on the descriptor here. It's almost like a stricter subset of the descriptor would be better here: do we _really_ want to be able to set field names at the component type level?
I do however concurr that there may be more stuff that then _is_ very relevant to the component. And a separate type to split those out is likely overkill.

crates/store/re_types_core/src/component_descriptor.rs
Wumpf: why not do several `write_u64`? Should have roughly the same effect, but then this is owned by the hasher - afraid of perf penalty?

crates/store/re_types_core/src/component_descriptor.rs
```diff
@@ -0,0 +1,177 @@
...
+    #[inline]
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        let Self {
+            archetype_name,
+            archetype_field_name,
+            component_name,
+        } = self;
+
+        let archetype_name = archetype_name.map_or(0, |v| v.hash());
+        let archetype_field_name = archetype_field_name.map_or(0, |v| v.hash());
```
Wumpf, line 43: similar to below, why not take the "regular" hash method?

crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -1013,11 +1015,37 @@ impl RecordingStream {
...
+                    batch.to_arrow_list_array()?,
+                ))
+            })
             .collect();
 
-        let components: IntMap<ComponentName, ArrowListArray<i32>> =
-            components?.into_iter().collect();
+        let components: ChunkComponents = components?.into_iter().collect();
+
+        {
```
Wumpf, line 1028: makes sense but how did this validation end up in here? isn't that more general than send_columns?

docs/snippets/all/descriptors/descr_builtin_archetype.rs
```diff
@@ -0,0 +1,65 @@
...
+    const APP_ID: &str = "rerun_example_descriptors_builtin_archetype";
+
+    let rec = rerun::RecordingStreamBuilder::new(APP_ID).spawn()?;
+
+    rec.log_static(
+        "data",
+        &rerun::Points3D::new([(1.0, 2.0, 3.0)]).with_radii([0.3, 0.2, 0.1]),
+    )?;
+
+    // When this snippet runs through the snippet comparison machinery, this environment variable
```
Wumpf, line 14: that's a really weird hack to do the checking. Nothing about this is `/docs/` anymore :/
wouldn't the older "roundtrip tests" do the same job just as well?

if not can we maybe at least add a short readme to this folder to point out that these are not really snippets for documenting but rather tests?

Wumpf (approved):
oof, that was a lot :)
ship it!

crates/store/re_types_core/src/component_descriptor.rs
teh-cmc: We want this to be `NoHash`, so we're not allowed more than one write. I'll add a comment.

crates/store/re_chunk_store/src/store.rs
```diff
@@ -701,7 +709,60 @@ impl ChunkStore {
...
+        Ok(stores)
+    }
+
+    /// Instantiate a new `ChunkStore` with the given [`ChunkStoreConfig`].
+    ///
+    /// The stores will be prefilled with the data in the given `log_msgs`.
+    ///
+    /// See also:
+    /// * [`ChunkStore::new`]
+    pub fn from_log_msgs(
```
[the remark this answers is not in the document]
teh-cmc: but then `from_log_msgs` needs to take an iterator of results instead and things get annoying for other people... I'd rather have a bit of duplication and make everyone happy

crates/store/re_chunk/src/transport.rs
```diff
@@ -486,7 +544,8 @@ impl Chunk {
             re_tracing::profile_scope!("row ids");
 
             let Some(row_ids) = transport.controls().find_map(|(field, column)| {
-                (field.name == RowId::name().as_str()).then_some(column)
+                // TODO(cmc): disgusting, but good enough for now.
+                (field.name == RowId::descriptor().component_name.as_str()).then_some(column)
```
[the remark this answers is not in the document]
teh-cmc: The clean alternative is that all these things should be working at the new protobuf metadata layer, with proper Descriptor types etc.

crates/store/re_types_core/src/loggable.rs
teh-cmc: It can definitely be useful yes, I've seen it when implementing actual examples.

If you're implementing a custom component and you know you always want it to be part of a specific archetype, you'd rather just do it once in the your component definition, rather than having to specify overrides at every log site.

crates/top/re_sdk/src/recording_stream.rs
teh-cmc: oO I did not write this code, I'm not sure what's going on

docs/snippets/all/descriptors/descr_builtin_archetype.rs
teh-cmc: I'll see if i can move it to roundtrips/

teh-cmc: oh no wait, i remember why I picked snippets/ -- these _are_ docs, they show how to override tags in different scenarios! but yeah the presence of the testing code in the rust version kinda suck, hmm...

Issue #8191, opened 2024-11-20
Missing `opened_application/recording_id` in external data loader example 


EtaLoop:
I tried to log data using an external data loader, with `log_file_from_path` and I got en error.

So I tried to test the loader for rust files (https://github.com/rerun-io/rerun/tree/main/examples/rust/external_data_loader) and I got the same error : 
```
ERROR re_data_loader::loader_external] Failed to execute external loader filepath="file.rs" loader="rerun/target/release/rerun-loader-rust-file" reason=Unrecognized argument: --opened-application-id
```
With this code : 
```
rec = rr.new_recording(application_id="My window", recording_id="My rec", spawn=True)
rr.log_file_from_path(file_path="file.rs", recording=rec.to_native())
```
So I add `opened_application_id` and `opened_recording_id` to https://github.com/rerun-io/rerun/blob/1ad3042a85622804a6923a5d0c65f82ba1e601d3/examples/rust/external_data_loader/src/main.rs#L28.

Now this works.

EtaLoop:
I put another external data loader here that doesn't work to be sure to test it when the bug will be fixed

https://github.com/rerun-io/rerun-loader-python-example-urdf.

teh-cmc:
Thanks!

PR #8309, opened 2024-12-04
Add opened_application_id, opened_recording_id in external data loader example

EtaLoop:
Fixes: #8191


teh-cmc:
Thanks for the PR -- unfortunately this isn't as simple as this.

I have [a branch](https://github.com/rerun-io/rerun/commits/cmc/fix_external_dl_examples/) that fixes all examples for all languages, but doing so highlighted a very hard to track down bug somewhere deep into the external loading stack.
I need to find it and fix it first before landing this.

My apologies -- we need to communicate clearly to non-Rerun employees that when an issue is assigned, one must sync with whomever is assigned before attempting anything in order to prevent duplicated efforts.

PR #8311, opened 2024-12-04
Add `re_data_ui` tooltip to graph view nodes

### LINKED ISSUE #8279: Add standard tooltip to graph view nodes
Use `data_ui::instance_hover_card_ui`

grtlr:
### Related

* Closes #8279.

### What

Title.

image

### DIFF
--- a/crates/viewer/re_space_view_graph/src/ui/draw.rs
+++ b/crates/viewer/re_space_view_graph/src/ui/draw.rs
@@ -289,11 +289,7 @@ pub fn draw_graph(
                     InstancePath::instance(entity_path.clone(), instance.instance_index);
                 ctx.select_hovered_on_click(
                     &response,
-                    vec![(
-                        Item::DataResult(query.space_view_id, instance_path.clone()),
-                        None,
-                    )]
-                    .into_iter(),
+                    Item::DataResult(query.space_view_id, instance_path.clone()),
                 );
 
                 response = response.on_hover_ui_at_pointer(|ui| {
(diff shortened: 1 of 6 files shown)


crates/viewer/re_space_view_graph/src/ui/draw.rs
```diff
@@ -281,29 +283,58 @@ pub fn draw_graph(
...
                 let instance_path =
                     InstancePath::instance(entity_path.clone(), instance.instance_index);
                 ctx.select_hovered_on_click(
                     &response,
-                    vec![(Item::DataResult(query.space_view_id, instance_path), None)].into_iter(),
+                    vec![(
+                        Item::DataResult(query.space_view_id, instance_path.clone()),
+                        None,
+                    )]
+                    .into_iter(),
```
abey79, line 296: You should be able to pass a single item here.

```suggestion
                    Item::DataResult(query.space_view_id, instance_path.clone()),
```

crates/viewer/re_space_view_graph/src/ui/draw.rs
```diff
@@ -281,29 +283,58 @@ pub fn draw_graph(
...
+                        instance_path.data_ui_recording(ctx, ui, UiLayout::Tooltip);
+                    });
+                });
+
                 response
             }
-            Node::Implicit { graph_node, .. } => {
-                draw_node(ui, center, node.label(), Default::default())
-                    .on_hover_text(format!("Implicit Graph Node: {}", graph_node.as_str(),))
+            Node::Implicit { edge_instance, .. } => {
```
abey79, line 316: I'm confused here. There could be many edge instances which refer to this implicit node right? Is that the first one that you encounter?

crates/viewer/re_space_view_graph/src/ui/draw.rs
```diff
@@ -281,29 +283,58 @@ pub fn draw_graph(
...
+                        item_ui::instance_path_button(
+                            ctx,
+                            &query.latest_at_query(),
+                            ctx.recording(),
+                            ui,
+                            Some(query.space_view_id),
+                            &instance_path,
+                        );
+
+                        instance_path.data_ui_recording(ctx, ui, UiLayout::Tooltip);
```
abey79, line 333: Because of my comment above and because this is generally somewhat confusing (and soon broken), I would suggest not having this here, and instead just a text stating "Implicit node created via a reference in the `GraphEdge` component" or something of that nature.

The edge instance tooltip should be displayed when the actual edge is hovered imo, but that can be left for later (btw, kurbo has point to bezier distance computation so this is something it could help with).

abey79 (approved):
I'm not super convinced by the "show instance for implicit node" part, but having the "regular" tooltip for the "regular" nodes is very nice 👍🏻

crates/viewer/re_space_view_graph/src/ui/draw.rs
grtlr: Yeah, you're right—I was too greedy here.

crates/viewer/re_space_view_graph/src/ui/draw.rs
grtlr: That makes sense, and I agree. I will change it back!

Issue #8314, opened 2024-12-04
Anti aliasing for outlines of lines broke in 0.20 (working fine in 0.19)


Wumpf:
![Image](https://github.com/user-attachments/assets/af965c5a-25ad-483c-bd08-8bfa138caf25)

Seems to affect all backends

Even even on WebGL - where we make some awful compromises in regards to outline anti-aliasing - it looked better on 0.19.0:
https://rerun.io/viewer/version/0.19.0?url=https%3A%2F%2Fapp.rerun.io%2Fversion%2F0.19.0%2Fexamples%2Fdna.rrd

Wumpf:
Looks like this does _not_ affect WebGL after all.

The difference here is that in WebGL we don't do custom anti aliasing by reading the MSAA samples in the shader.


WebGL 0.20
![Image](https://github.com/user-attachments/assets/d16bc50c-5895-488c-8e5f-05b6d8d298a4)
WebGL 0.19
![Image](https://github.com/user-attachments/assets/4a3e1932-4abc-4cd3-ae86-92d116345bc9)


WebGPU 0.20
![Image](https://github.com/user-attachments/assets/e44b1507-04f9-4880-8529-1905536c9bed)
WebGPU 0.19
![Image](https://github.com/user-attachments/assets/284f2023-2cef-462c-a368-21c7667f480a)

Wumpf:
Turns out this is exclusively limited to outlines of _lines_

PR #8317, opened 2024-12-04
Fix outlines for lines having more preceived aliasing since 0.20

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/8314

### What

All screenshots on MacOS lowdpi (==scaling 100%) since it's easier to spot there (but also still impactful on highdpi==scaling>=200%). Listing WebGPU & WebGL separately since WebGL version doesn't do anti-aliasing on outlines (technical limitations)

0.19 WebGPU Chrome
image

0.20/main WebGPU Chrome
image

This PR WebGPU Chrome
image

---

0.19 WebGL Chrome
image

0.20/main WebGL Chrome
image

This PR WebGL Chrome
image

### DIFF
--- a/crates/build/re_types_builder/src/codegen/cpp/mod.rs
+++ b/crates/build/re_types_builder/src/codegen/cpp/mod.rs
@@ -2182,7 +2182,7 @@ fn quote_archetype_field_type(hpp_includes: &mut Includes, obj_field: &ObjectFie
             quote! { Collection<#elem_type> }
         }
         // TODO(andreas): This should emit `MonoCollection` which will be a constrained version of `Collection`.
-        // (simply adapting `MonoCollection` breaks some existing code, so this not entirely trivial to do.
+        // (Simply adapting `MonoCollection` breaks some existing code, so this not entirely trivial to do.
         //  Designing constraints for `MonoCollection` is harder still)
         Type::Object(fqname) => quote_fqname_as_type_path(hpp_includes, fqname),
         _ => panic!("Only vectors and objects are allowed in archetypes."),
--- a/crates/viewer/re_renderer/shader/lines.wgsl
+++ b/crates/viewer/re_renderer/shader/lines.wgsl
@@ -370,10 +370,12 @@ fn fs_main_picking_layer(in: VertexOut) -> @location(0) vec4u {
 @fragment
 fn fs_main_outline_mask(in: VertexOut) -> @location(0) vec2u {
     // Output is an integer target, can't use coverage therefore.
-    // But we still want to discard fragments where coverage is low.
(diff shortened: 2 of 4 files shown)


crates/build/re_types_builder/src/codegen/cpp/mod.rs
```diff
@@ -2182,7 +2182,7 @@ fn quote_archetype_field_type(hpp_includes: &mut Includes, obj_field: &ObjectFie
             quote! { Collection<#elem_type> }
         }
         // TODO(andreas): This should emit `MonoCollection` which will be a constrained version of `Collection`.
-        // (simply adapting `MonoCollection` breaks some existing code, so this not entirely trivial to do.
+        // (Simply adapting `MonoCollection` breaks some existing code, so this not entirely trivial to do.
```
Wumpf, line 2185: load bearing rust change so I get web builder ci triggered

crates/viewer/re_renderer/shader/lines.wgsl
```diff
@@ -370,10 +370,12 @@ fn fs_main_picking_layer(in: VertexOut) -> @location(0) vec4u {
...
     // Output is an integer target, can't use coverage therefore.
-    // But we still want to discard fragments where coverage is low.
-    // Since the outline extends a bit, a very low cut off tends to look better.
+    // But we still want to discard fragments where coverage is too low, otherwise
+    // we'd not handle rounded corners etc. correctly.
+    // Note that `compute_coverage` may already give coverage values below 1.0 along the
+    // "main body of the line", not just the caps.
     var coverage = compute_coverage(in);
-    if coverage < 1.0 {
+    if coverage <= 0.0 {
```
emilk, line 378: Should the same fix be applied anywhereelse?

Screenshot 2024-12-04 at 22 25 00

emilk (approved):
Makes sense - thanks for the fix! Soon we can have regression checks for this :)

crates/viewer/re_renderer/shader/lines.wgsl
Wumpf: Hmmmmmmmm likely! I'll do a bit of A/B testing on points as well (note to self: proving this on regular points should extend just fine to depth)

More note to self: This blew up because we go down in coverage along the line now when we actually have to rely on "normal" MSAA to do what little anti aliasing we can do for outlines since alpha-to-coverage doesn't work with integer targets (gosh alpha-to-coverage is such a bizarre hack!). So I'm actually not 100% sure why I previously came to the seemingly opposite conclusion on what a good coverage cutoff is. The new one makes much more sense to me given this situation.
Should enhance that comment a little bit more with this knowledge

PR #8321, opened 2024-12-05
ReDap Catalog as data source

zehiko:
### Depends on

* https://github.com/rerun-io/rerun/pull/8292

### What

Add catalog as a data source. Did tiny bit of refactoring of ``Address`` to avoid some duplication. Streaming catalog logic is copy-paste of fetch recording logic, unification will be done once gRPC metadata API updates are complete. 

Screenshot 2024-12-05 at 10 26 50


crates/store/re_grpc_client/src/address.rs
```diff
@@ -1,65 +1,111 @@
...
 }
 
-impl std::fmt::Display for Address {
+impl std::fmt::Display for RecordingAddress {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
-            "rerun://{}/recording/{}",
-            self.addr_port, self.recording_id
+            "RecordingAddress {{ redap_endpoint: {}, recording_id: {} }}",
```
jleibs, line 21: It was nice that previously display let us round-trip from str -> Address -> str, which meant we could print out addresses that could be copied back to the rerun-CLI input.

crates/store/re_data_source/src/data_source.rs
```diff
@@ -35,6 +34,10 @@ pub enum DataSource {
     /// A file on a Rerun Data Platform server, over `rerun://` gRPC interface.
     #[cfg(feature = "grpc")]
     RerunGrpcUrl { url: String },
+
+    /// A catalog containing metadata information about recordings stored on Rerun Data Platform.
+    #[cfg(feature = "grpc")]
+    RerunGrpcCatalog { url: String },
```
jleibs, line 40: A separate source here seems unnecessary, `RerunGrpcUrl` could support both variants.

crates/store/re_data_source/src/data_source.rs
```diff
@@ -247,8 +253,17 @@ impl DataSource {
...
+                let url =
+                    url::Url::parse(&url).with_context(|| format!("Invalid gRPC URL: {url}"))?;
                 re_grpc_client::stream_recording(url, on_msg).map_err(|err| err.into())
             }
+
+            #[cfg(feature = "grpc")]
+            Self::RerunGrpcCatalog { url } => {
+                let url =
+                    url::Url::parse(&url).with_context(|| format!("Invalid gRPC URL: {url}"))?;
+                re_grpc_client::stream_catalog(url, on_msg).map_err(|err| err.into())
```
jleibs, line 265: Again, rather than duplicating, we could have just renamed `stream_recording` -> `stream_redap`

crates/store/re_grpc_client/src/address.rs
```diff
@@ -1,65 +1,111 @@
...
+#[error("URL {url:?} should follow rerun://addr:port/recording/12345 for recording or rerun://addr:port/catalog for catalog")]
+pub struct InvalidRedapAddress {
+    url: Url,
     msg: String,
 }
 
 /// Parsed `rerun://addr:port/recording/12345`
-pub struct Address {
-    pub addr_port: String,
+pub struct RecordingAddress {
```
jleibs, line 12: This could have become something like:
```
pub enum RedapAddress {
  Recording(...)
  Catalog
}
```

crates/store/re_grpc_client/src/lib.rs
```diff
@@ -71,26 +71,62 @@ enum StreamError {
...
+        },
     );
 
     spawn_future(async move {
         if let Err(err) = stream_recording_async(tx, address, on_msg).await {
             re_log::warn!(
-                "Error while streaming {url}: {}",
+                "Error while streaming {redap_url}: {}",
                 re_error::format_ref(&err)
             );
```
jleibs, line 131: This is where I think we could put a match statement on the new enum `RedapAddress` and split between `stream_recording_async` and `stream_catalog_async` depending on the match arm.

jleibs (approved):

crates/store/re_grpc_client/src/address.rs
zehiko: yeah, agree. I wanted to verify early i.e. get `Url` in here, rather than just a string. I'll figure it out.

crates/store/re_data_source/src/data_source.rs
zehiko: hm, yeah, not sure why I went this way instead of just differentiating in the grpc client. good suggestion!

crates/store/re_grpc_client/src/address.rs
zehiko: I also thought about that, but the use cases are different and then you need matching in the consumer (unless I'm missing something obvious)

Issue #3133, opened 2023-08-29
Migrate to Radix color table and "2D color coordinates"


abey79:
These are some notes following a discussion with @martenbjork on how to evolve the current design token system in the short term, such as to address some of its shortcoming and go in the direction of  #3284.

## Goals

- start using the [Radix color tables](https://www.radix-ui.com/colors) and use taint/index color coordinates
- make it much easier to iterate on alias (mapping from "type of use in the UI" -> "color coordinate"), while retaining a single source of truth (so both rerun, Figma, and @martenbjork's React prototype use the same mappings)

## Solutions

1) Create a public, single-file repository containing a dead-simple JSON containing the aliases.
    The idea is to minimise friction when new alias must be created by rerun dev (use-case example: subdued container color in the blueprint tree UI, etc.) during dev iterations. Using a file on rerun-io/landing is not possible since that repo is private (can't easy pull the json where it's needed).
2) Hard-code the radix color table in Rerun.
3) Use that file instead of the current `design-token.json` file. Create/update script to automate pulling the file from the reference repo.
4) Take similar steps on the side of Figma and @martenbjork's proto.
5) Progressively update the code-base use the new system instead of ugly hacks (like setting 50% gamma, etc.)

martenbjork:
Thanks for taking notes!

> Hard-code the radix color table in Rerun

I think it might be worth having this live in the tokens file for 3 reasons:

1. Radix may update their values in the future ([they frequently tweak them](https://github.com/radix-ui/colors/commits/main))
2. Other systems (Figma import, website) need the values as well. It's good if they can do it without having to install radix or copy paste and hard code radix values.
3. We want to have non-radix values in there as well, for typography, shadows etc.

Based on this, I suggest this JSON format:

```json
{
  // We add radix colors and other global tokens here
  "colors": {
    "slate": {
      "1": {
        "$type": "color",
        "$description": "",
        "$value": {
          "hex": "#000000"
        }
      },
      "2": {
        "$type": "color",
        "$description": "",
        "$value": {
          "hex": "#000000"
        }
      },
      "3": {
        "$type": "color",
        "$description": "",
        "$value": {
          "hex": "#000000"
        }
      }
    }
  },

  // This is where we add our custom values
  "surface": {
    "default": {
      "$type": "color",
      "$description": "Default color in...",
      "$value": "{color.slate.2}"
    }
  },
  "interactive": {
    "default": {
      "$type": "color",
      "$description": "Default color in...",
[...]

Issue #3058, opened 2023-08-21
setting light mode


edulecom:
Hello,

I'm playing with .\rerun\crates\re_ui\examples\re_ui_example.rs and I'm trying to change the theme to light.
I have added only one line in the fn update:

        egui::TopBottomPanel::bottom("bottom_panel")
            .frame(self.re_ui.bottom_panel_frame())
            .show_animated(egui_ctx, self.bottom_panel, |ui| {
                ui.strong("Bottom panel");
                egui::widgets::global_dark_light_mode_buttons(ui); // this line is added
            });  

Effect is strange: only central panel applies the theme. It can be seen in an attached video.
Any help would very appreciated.

https://github.com/rerun-io/rerun/assets/104624770/cf9a8e50-f6f1-4a7d-be5d-a5ce285be531

Wumpf:
The out of the box theme settings are bound to not work properly in rerun as we sadly change a lot of the theming colors in Rerun on-the-fly. At least I suspect that's what's most of the issue here.

That said, (re)adding a light mode is something we want to do eventually.
Are you just trying out things or is a Rerun light theme something you'd like to have as well?

edulecom:
Thank you for your response. I'm currently in the process of developing an app using re_ui, and while the absence of a light mode isn't a critical issue, having that feature would certainly be a welcome addition.

emilk:
Most of the work has landed in https://github.com/rerun-io/rerun/pull/9953 but there are a few things left to do:

* Update
  * [x] Table headers (e.g. catalogs)
  * [x] Breadcrumbs ui
  * [x] Density graph
  * [x] 2D background color
  * [ ] …
* Stabilize
  * [ ] Remove the experimental warning
  * [ ] Switch the default theme to follow the system
  * [x] Add light mode snapshot tests
  * [ ] Make sure `ThemePreference` is saved/restored properly

lubidkdh:
Hello, I'm using rerun through the python language. I hope to make the rerun theme default to light in the form of code. I also hope that the rerun theme can be saved as part of the blueprint in an rbl file. Or is there any other convenient way to adjust the rerun theme? Also, could you specify the address of the usage document for rerun 0.24.0a1 version? I'm very sorry that I haven't found the address of the document for the time being. Thank you very much.

lubidkdh:
Ok. Thank you very much for your response!

Wumpf:
We have experimental light made in the upcoming 0.24.
[...]

PR #8322, opened 2024-12-05
Refactor design tokens to use a proper color table

abey79:
### Related

* Part of/related to https://github.com/rerun-io/rerun/issues/3133
* Part of https://github.com/rerun-io/rerun/issues/3058

### What

This PR refactors the design token as follows:

- All colors from `design_tokens.json` are now loaded in a big table at startup.
- Colors are now referred to using the new the `ColorToken`, which is basically an index into this table.
- Removed all of the color aliases stuff. This was only partially used and very cumbersome to update. The `design_token.rs` file is the _de facto_ source of truth of semantic aliasing for colors.


crates/viewer/re_ui/src/color_table.rs
```diff
@@ -0,0 +1,217 @@
...
+/// Use [`crate::DesignTokens::color`] to get the color corresponding to a token.
+#[derive(Debug, Clone, Copy, Hash)]
+pub struct ColorToken {
+    hue: Hue,
+    shade: Shade,
+}
+
+impl ColorToken {
+    #[inline]
+    pub fn new(hue: Hue, shade: Shade) -> Self {
```
abey79, line 189: With `enumn` (which we already depend on) I could easy add support for `u16` input, but then we must deal with error management. For that, we could either:
- panic
- have an invalid state for `ColorToken` (e.g. `shade = None`), which would return a random color when queried (so it flashes in the UI) and logs an error

crates/viewer/re_ui/src/color_table.rs
```diff
@@ -0,0 +1,217 @@
...
+            Self::Red => f.write_str("Red"),
+            Self::Blue => f.write_str("Blue"),
+            Self::Purple => f.write_str("Purple"),
+        }
+    }
+}
+
+/// A shade for a [`ColorToken`].
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+pub enum Shade {
```
emilk, line 37: Please document what `Shade` means (at least roughly). Is `S1000` fully saturated, and `S0` gray? Or is there no such inherent meaning?

crates/viewer/re_ui/src/color_table.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    pub fn as_u16(self) -> u16 {
+        self as u16
+    }
+}
+
+/// A table mapping all combination of [`Hue`] and [`Shade`] to a [`egui::Color32`].
+#[derive(Debug)]
+pub struct ColorTable {
+    pub color_table: nohash_hasher::IntMap<Hue, nohash_hasher::IntMap<Shade, egui::Color32>>,
```
emilk, line 139: You're using `nohash_hasher` here, presumably for speed.

If you really want full speed, use sequential numbers of the enum variants (that's the default) and then use a LUT

crates/viewer/re_ui/src/color_table.rs
```diff
@@ -0,0 +1,217 @@
...
+            Shade::S950,
+            Shade::S975,
+            Shade::S1000,
+        ];
+        &ALL
+    }
+
+    pub fn as_u16(self) -> u16 {
+        self as u16
+    }
```
emilk, line 133: What will this be used for?

crates/viewer/re_ui/src/design_tokens.rs
```diff
@@ -418,23 +428,36 @@ impl DesignTokens {
...
+    ColorTable {
+        color_table: Hue::all()
+            .iter()
+            .map(|hue| {
+                let shade_map = Shade::all()
+                    .iter()
+                    .map(|shade| {
+                        let color = get_color_from_json(
+                            json,
+                            &format!("{{Global.Color.{hue}.{}}}", shade.as_u16()),
```
emilk, line 452: Just implement `Display` for `Shade` instead? 🤷 

```suggestion
                            &format!("{{Global.Color.{hue}.{shade}}}"),
```

emilk (approved):
Very nice to finally get this!

crates/viewer/re_ui/src/color_table.rs
abey79: Do you have a suggestion for a pattern/helper crate for obtaining at compile time the number of variant of an enum and a list of all variants?

emilk: https://doc.rust-lang.org/stable/std/mem/fn.variant_count.html :]

abey79: that looks really cool! but...

image

abey79: oh I can use `strum`

PR #8326, opened 2024-12-05
Fail codegen when archetype fields don't have one of the required attributes

abey79:
### What

All of our archetypes' fields must have one of the `attr.rerun.component_{required|recommended|optional}` attribute. Let's fail codegen when they are missing, so we stop falling into that pitfall.

### DIFF
--- a/crates/build/re_types_builder/src/codegen/rust/api.rs
+++ b/crates/build/re_types_builder/src/codegen/rust/api.rs
@@ -1047,6 +1047,23 @@ fn quote_trait_impls_for_archetype(obj: &Object) -> TokenStream {
 
     assert_eq!(kind, &ObjectKind::Archetype);
 
+    // sanity check that each field has one of the rerun component type attributes
+    for field in &obj.fields {
+        assert!(
+            [
+                ATTR_RERUN_COMPONENT_OPTIONAL,
+                ATTR_RERUN_COMPONENT_RECOMMENDED,
+                ATTR_RERUN_COMPONENT_REQUIRED
+            ]
+            .iter()
+            .any(|attr| field.try_get_attr::<String>(attr).is_some()),
+            "archetype `{}` must add one of the \"attr.rerun.component_{{required|recommended|\
+            optional}}\" attributes to field `{}`",
+            obj.name,
+            field.name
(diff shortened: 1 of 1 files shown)


crates/build/re_types_builder/src/codegen/rust/api.rs
```diff
@@ -1047,6 +1047,23 @@ fn quote_trait_impls_for_archetype(obj: &Object) -> TokenStream {
 
     assert_eq!(kind, &ObjectKind::Archetype);
 
+    // sanity check that each field has one of the rerun component type attributes
+    for field in &obj.fields {
+        assert!(
```
Wumpf, line 1052: can we do `reporter.error` with all the right paths & stuff please? That way it's way easier to find & address the error

Wumpf (changes requested):

crates/build/re_types_builder/src/codegen/rust/api.rs
abey79: good call. done.

crates/build/re_types_builder/src/codegen/rust/api.rs
```diff
@@ -1047,6 +1047,25 @@ fn quote_trait_impls_for_archetype(obj: &Object) -> TokenStream {
...
 
+    // sanity check that each field has one of the rerun component type attributes
+    for field in &obj.fields {
+        if [
+            ATTR_RERUN_COMPONENT_OPTIONAL,
+            ATTR_RERUN_COMPONENT_RECOMMENDED,
+            ATTR_RERUN_COMPONENT_REQUIRED,
+        ]
+        .iter()
+        .all(|attr| field.try_get_attr::<String>(attr).is_none())
```
emilk, line 1058: It would be nicer if we checked that it was exactly one

Wumpf (approved):
lgtm.. hower, I didn't realize you put this into the _rust_ codegen when this is actually completely language agnostic.
this should go into `Objects::from_raw_schema` which is where we have general validation like this
... but I don't want to hold up if this turns out to be too much hassle.

Wumpf:
thanks!

PR #8328, opened 2024-12-05
gRPC spec update: unify data and metadata API responses

zehiko:
Part 2 of API unification that simplifies dealing with data streams coming from Data Platform.


rerun_py/src/remote.rs
```diff
@@ -249,12 +266,13 @@ impl PyStorageNodeClient {
...
-                .map_err(|err| PyRuntimeError::new_err(err.to_string()))?;
-
             let request = UpdateCatalogRequest {
                 recording_id: Some(RecordingId { id: id.to_owned() }),
-                metadata: Some(metadata),
+                metadata: Some(DataframePart {
+                    encoder_version: EncoderVersion::V0 as i32,
+                    payload: encode(EncoderVersion::V0, metadata_tc)
+                        .map_err(|err| PyRuntimeError::new_err(err.to_string()))?,
+                }),
```
jleibs, line 275: This can come in a future PR, but I think `encode` should return a `DataframePart` and eseentially hide the encoder-version bit.
```suggestion
                metadata: Some(DataframePart::encode(metadata_tc)
                        .map_err(|err| PyRuntimeError::new_err(err.to_string()))?),
```

jleibs (approved):

rerun_py/src/remote.rs
zehiko: that makes sense, and there's few more things to sort out in the codec, but Jan suggested he might be working on this exact code and didn't want to make his rebase painful. That can be a follow up indeed, introducing ``DataframePart::encode`` and also getting rid of ``NoData`` message (that has many clunky consequences at the moment).

zehiko: https://github.com/rerun-io/rerun/issues/8356

PR #8331, opened 2024-12-05
Fix various doc links in SDKs

Wumpf:
PR title is an accompanying changelog entry, but the real thing here is all about codegen fixes:

* use `reporter` for doc issues
* make `reporter` output colored
* support field & enum links - this is a bit limited depending on the language. Experimented a bit, but deemed this now good enough. We don't use those often anyways
* fix links
* disable warnings for python autogen serialization being available when a custom one is present - in all cases we have the custom serialization there for a good reason or aren't entirely sure if it would break something


Wumpf:
drafted: some links are wrong, need to fix this up

crates/build/re_types_builder/src/codegen/python/mod.rs
```diff
@@ -1877,7 +1889,10 @@ fn quote_arrow_support_from_obj(
     ) {
         Ok(automatic_arrow_serialization) => {
             if ext_class.has_native_to_pa_array {
-                reporter.warn(&obj.virtpath, &obj.fqname, format!("No need to manually implement {NATIVE_TO_PA_ARRAY_METHOD} in {} - we can autogenerate the code for this", ext_class.file_name));
+                // There's usually a good reason why serialization is manually implemented,
+                // so warning about it is just spam.
+                // We could introduce an opt-in flag, but by having a custom method in the first place someone already made the choice.
+                // reporter.warn(&obj.virtpath, &obj.fqname, format!("No need to manually implement {NATIVE_TO_PA_ARRAY_METHOD} in {} - we can autogenerate the code for this", ext_class.file_name));
```
emilk, line 1895: If you want to keep the code around for later for whatever reason, enclose it in `if false` instead to prevent code rot:

```suggestion
                if false {
                    reporter.warn(&obj.virtpath, &obj.fqname, format!("No need to manually implement {NATIVE_TO_PA_ARRAY_METHOD} in {} - we can autogenerate the code for this", ext_class.file_name));
                    }
```

Otherwise, remove it

crates/store/re_types/definitions/rerun/blueprint/archetypes/container_blueprint.fbs
```diff
@@ -22,21 +22,21 @@ table ContainerBlueprint (
...
+    /// For [components.ContainerKind.Vertical] containers, the length of this list should always match the number of contents.
     ///
-    /// Ignored for `Horizontal` containers.
+    /// Ignored for [components.ContainerKind.Horizontal] containers.
     row_shares: [rerun.blueprint.components.RowShare] ("attr.rerun.component_optional", nullable, order: 500);
 
     /// Which tab is active.
     ///
-    /// Only applies to `Tabs` containers.
+    /// Only applies to `Tabs` containers.pixi
```
emilk, line 39:
```suggestion
    /// Only applies to `Tabs` containers.
```

crates/store/re_types/src/components/fill_mode.rs
```diff
@@ -56,10 +56,10 @@ impl ::re_types_core::reflection::Enum for FillMode {
     fn docstring_md(self) -> &'static str {
         match self {
             Self::MajorWireframe => {
-                "Lines are drawn around the parts of the shape which directly correspond to the logged data.\n\nExamples of what this means:\n\n* An [`archetypes::Ellipsoids3D`][crate::archetypes::Ellipsoids3D] will draw three axis-aligned ellipses that are cross-sections\n  of each ellipsoid, each of which displays two out of three of the sizes of the ellipsoid.\n* For [`archetypes::Boxes3D`][crate::archetypes::Boxes3D], it is the edges of the box, identical to `DenseWireframe`."
+                "Lines are drawn around the parts of the shape which directly correspond to the logged data.\n\nExamples of what this means:\n\n* An [`archetypes::Ellipsoids3D`][crate::archetypes::Ellipsoids3D] will draw three axis-aligned ellipses that are cross-sections\n  of each ellipsoid, each of which displays two out of three of the sizes of the ellipsoid.\n* For [`archetypes::Boxes3D`][crate::archetypes::Boxes3D], it is the edges of the box, identical to [`components::FillMode::DenseWireframe`][crate::components::FillMode::DenseWireframe]."
```
emilk, line 59: What do we use `docstring_md()` for? If it is for showing in the UI, then it doesn't really make sense to have doclinks in it, unless those link to https://rerun.io/docs/reference/types/

We should probably use `Target::WebDocsMarkdown` when generating this string, no?

crates/build/re_types_builder/src/report.rs
```diff
@@ -41,21 +41,46 @@ impl Reporter {
...
+            ))
             .ok();
     }
 
     #[allow(clippy::needless_pass_by_value)] // `&impl ToString` has worse usability
     pub fn error_any(&self, text: impl ToString) {
         self.errors.send(text.to_string()).ok();
     }
+
+    fn format_virtpath(virtpath: &str) -> String {
```
emilk, line 73: this could use a docstring to explain it

emilk (approved):
Awesome, thank you!

Can we fix the contents of `docstring_md` while we're at it? 🙏

crates/store/re_types/src/components/fill_mode.rs
Wumpf: good catch, `WebDocsMarkdown` makes a lot of sense here!

Issue #8266, opened 2024-12-02
Support dragging entities to existing view to add them to their filter


abey79:
Drag one (or more, ideally) entity from blueprint/streams tree to an existing view in to add it to their filter.

Feedback should be provided wether or not the view can receive the entity (to be designed).

We need to: 
1. Visually highlight the View in the Viewport with a blue stroke and blue label
2. the pill is having an icon for 'add to existing' version
3. We have an indicator on the blueprint panel as well: start with simply outlining the View in blue
[Figma](https://www.figma.com/design/72AuvfbssFJeqqnQoJscYT/Katya's-playground?node-id=4557-88853&t=2oCTuRKq5YLhtfZT-1)

![Image](https://github.com/user-attachments/assets/0ba03b45-8d38-4c4b-b94b-bcc165b1a715)

EXTRA (maybe it's not too hard now? else -> consider later!)
![Image](https://github.com/user-attachments/assets/23154869-3b2c-4437-8459-99c144104fe3)

Issue #7108, opened 2024-08-08
Trackig issue: drag-and-drop improvement


abey79:
#### Views
 - [ ] support drag-and-drop of view multi-selection (currently dragging force-single-select the dragged view)

#### Entities

- [ ] Drag entities from the streams tree to the blueprint tree to add to an existing view.
- [ ] Drag entities from the blueprint/streams trees to any view to add it to its filter.
- [ ] Drag entities to the viewport or in-between views to create a new view with it (requires making a guess on the view type, so maybe blocked on ability to change the view kind of an existing view)

PR #8334, opened 2024-12-05
Allow drag-and-dropping multiple containers and views in the blueprint tree

### LINKED ISSUE #8275: Support drag-and-drop for sensible multi-selections
Currently we support drag-and-drop of single-selection only (we force-single-select on drag start). We should support drag-and-dropping multi-selections at least when they "make sense", e.g. only entities are selected, or only views and containers are selected.

TODO:
- provide some kind of feedback of the "draggability" of the current selection
- alternatively, keep the current force-single-select behaviour when the current multi-selection is not draggable

Related:
- https://github.com/rerun-io/rerun/issues/5307
- https://github.com/rerun-io/rerun/issues/4887

### LINKED ISSUE #8276: When dragging, display a pill with a summary of what's being dragged next to the cursor
...e.g. "2 views", "3 entities".

BTW, let it be known that a "pill" is this kind of pill-shaped UI element:

image

Pill:
corner radius: 2
font oclor: text/default
background: blue-350 (should be the default selection blue)
stroke: blue-375 (slightly lighter to give a bit of stand out)
for error: ~~same values but red~~ no red pill 
icons are in the icons :) 
[Figma](https://www.figma.com/design/72AuvfbssFJeqqnQoJscYT/Katya's-playground?node-id=4557-88853&t=2oCTuRKq5YLhtfZT-1) 

Image

abey79:
### Related

* Closes #8276
* Closes #8275
* Part of #8266
* Part of #8267
* Related to #7108

### What

This PR makes it possible to drag multi-selection of views and containers within the blueprint tree. It lays the foundation of a system that will be extended to other drag payload and UI sections.

Specifically:
- `ctx.handle_select_hover_drag_interactions()` (formerly `ctx.select_hovered_on_click()`) is now able to initiate drag interactions. This is opt-in for now, as dragging from most place isn't supported yet (to implemented in follow-up PRs).
- Introduce a new `DragAndDropPayload` type to interoperate between various part of the UI. This type also enforce the grouping of items that can meaningfully be dragged together (e.g. it's ok to drag a view and a container together, because there exist somewhere they can be dropped to, but it's not ok to drag a view and an entity together).
- When a drag is successfully initiated, a "pill" is displayed along the cursor which indicates the content of what's being dragged.
- Introduces a _very_ hack mechanism for a black list of undraggable items (aka the root container).
- Update blueprint tree to support multiple selection and the new drag and drop payload type.
- Updates egui to latest `master`.


nikolausWest:
I haven't been able to actually drop any dragged entities yet. Is that intended to work here or is it just dragging?

abey79:
> I haven't been able to actually drop any dragged entities yet. Is that intended to work here or is it just dragging?

Yeah this PR is just about views/container but lays the groundwork. Support for dragging entities will come next.

nikolausWest:
Ignore for now if this is just super premature design feedback but I don't like the experience of showing an error all the time. If we want to show an error on the drag-pill it feels like that should be reserved for situations where I clearly want to drop it somewhere but that doesn't work.

https://github.com/user-attachments/assets/baefcc02-9ed8-4cdf-bf3d-32d95d002b58

nikolausWest:
I expected the result of both these kinds of drags to have an effect:


https://github.com/user-attachments/assets/b621f53a-18d4-45d0-a612-85b772281632



https://github.com/user-attachments/assets/1c957ac4-bf1d-4b87-97d7-3c22b8058798

crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -214,6 +215,10 @@ impl AppState {
             )),
         );
 
+        // The root container cannot be dragged.
+        let undraggable_items =
+            re_viewer_context::Item::Container(viewport_ui.blueprint.root_container).into();
+
```
abey79, line 221: This `undraggable_items` business feels very hacky, but I'm unsure how to factor that better. Feedback welcome.

crates/viewer/re_viewer_context/src/drag_and_drop.rs
```diff
@@ -0,0 +1,184 @@
...
+                .response;
+
+            let delta = pointer_pos - response.rect.right_bottom();
+            ui.ctx()
+                .transform_layer_shapes(layer_id, emath::TSTransform::from_translation(delta));
+        }
+    }
+}
+
+fn drag_pill_frame(error_state: bool) -> egui::Frame {
```
abey79, line 99: we no longer use `error_state=true`. I'll remove that dead code in the next PR.

crates/viewer/re_viewer_context/src/contents.rs
```diff
@@ -65,6 +65,14 @@ impl Contents {
...
 }
 
+impl TryFrom<&ItemCollection> for Vec<Contents> {
+    type Error = ();
+
+    fn try_from(value: &ItemCollection) -> Result<Self, Self::Error> {
+        value.iter().map(|(item, _)| item.try_into()).collect()
+    }
+}
+
```
abey79, line 75: in the next PR, I'm moving that next to `ItemCollection` instead, because I'll have another similar one but for entities.

crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
```diff
@@ -788,25 +792,30 @@ impl BlueprintTree {
...
+        if dragged_contents.iter().any(|dragged_contents| {
+            if let Contents::Container(dragged_container_id) = dragged_contents {
+                if viewport
+                    .is_contents_in_container(&drop_target.target_parent_id, dragged_container_id)
+                {
+                    return true;
+                }
             }
+            false
+        }) {
```
emilk, line 817: this would be more readable if you broke out the boolean and/or lambda to a named variable:

``` rs
let parent_contains_dragged_content = |dragged_contents| { … };
if dragged_contents.iter().any(parent_contains_dragged_content) {
```

crates/viewer/re_ui/src/ui_ext.rs
```diff
@@ -131,6 +131,21 @@ pub trait UiExt {
         )
     }
 
+    /// Adds a non-interactive, optionally tinted small icon.
+    ///
+    /// Uses [`DesignTokens::small_icon_size`]. Returns the rect where the icon was painted.
+    fn small_icon(&mut self, icon: &Icon, tint: Option<egui::Color32>) -> egui::Rect {
```
emilk, line 137: wouldn't it make more sense to return the `egui::Response`, like for any other widget?

crates/viewer/re_view_graph/src/ui/draw.rs
```diff
@@ -299,9 +299,10 @@ pub fn draw_graph(
 
                 let instance_path =
                     InstancePath::instance(entity_path.clone(), instance.instance_index);
-                ctx.select_hovered_on_click(
+                ctx.handle_select_hover_drag_interactions(
                     &response,
                     Item::DataResult(query.view_id, instance_path.clone()),
+                    false,
```
emilk, line 305: Instead of adding `false` and `true` as argument, did you consider having two different functions, for increased readability? In particular, it is very hard for me reviewing this code to know what `false` means.

I.e. you could have kept the old `select_hovered_on_click` with the previous meaning, and added `handle_select_hover_drag_interactions` (without a `bool`) for draggable things. But maybe its better to force each users to consider wether or not an item is draggable in this context?

crates/viewer/re_viewer_context/src/viewer_context.rs
```diff
@@ -131,11 +135,14 @@ impl ViewerContext<'_> {
...
-    /// Set hover/select/focus for a given selection based on an egui response.
-    pub fn select_hovered_on_click(
+    /// Consistently handle the selection, hover, drag start interactions for a given set of items.
+    ///
+    /// The drag start interaction is optional and controlled by the `draggable` parameter.
+    pub fn handle_select_hover_drag_interactions(
         &self,
         response: &egui::Response,
         selection: impl Into<ItemCollection>,
+        draggable: bool,
```
emilk, line 145: What are the design considerations for when to set this to `true` and when to set this to `false`?

Could the value of this be inferred from the items instead?

crates/viewer/re_viewport_blueprint/src/viewport_blueprint.rs
```diff
@@ -572,12 +572,12 @@ impl ViewportBlueprint {
     /// Move the `contents` container or view to the specified target container and position.
     pub fn move_contents(
         &self,
-        contents: Contents,
+        contents: &[Contents],
```
emilk, line 575: Better to pass by value if you're going to turn it into a Vec anyways
```suggestion
        contents: Vec<Contents>,
```

crates/viewer/re_viewer_context/src/viewer_context.rs
```diff
@@ -79,6 +80,9 @@ pub struct ViewerContext<'a> {
     /// The focused item is cleared every frame, but views may react with side-effects
     /// that last several frames.
     pub focused_item: &'a Option<crate::Item>,
+
+    /// If a selection contains any `undraggable_items`, it may not be dragged.
+    pub undraggable_items: &'a ItemCollection,
```
emilk, line 85: So on different frames there are different items that are currently forbidden from being dragged?

Can you give an example of such an item here?
Who is responsible for checking this?

Can this be used by `handle_select_hover_drag_interactions` so we don't need to pass in the bool everywhere?

crates/viewer/re_viewer/src/app_state.rs
emilk: If we add the viewport blueprint to the the `ViewerContext`, then `handle_select_hover_drag_interactions` can check for it directly

crates/viewer/re_viewer_context/src/drag_and_drop.rs
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+}
+
+/// Display the currently dragged payload as a pill in the UI.
+///
+/// This should be called once per frame.
+pub fn drag_and_drop_payload_cursor_ui(ui: &mut egui::Ui) {
+    if let Some(payload) = egui::DragAndDrop::payload::<DragAndDropPayload>(ui.ctx()) {
+        if let Some(pointer_pos) = ui.ctx().pointer_interact_pos() {
```
emilk, line 59: It's unclear what `ui` should be passed in here, and indeed: it doesn't seem to matter.

I think it would be more clear to pass in a `egui::Context` and then create a new `Ui` with `Ui::new`

crates/viewer/re_viewer_context/src/drag_and_drop.rs
```diff
@@ -0,0 +1,184 @@
...
+            left: 6.0,
+            right: 9.0,
+            top: 5.0,
+            bottom: 4.0,
+        },
+        ..Default::default()
+    }
+}
+
+fn items_to_string<'a>(items: impl Iterator<Item = &'a Item>) -> String {
```
emilk, line 119: When dragging a single item (e.g. a space-view), would it be nice to show its name instead? @gavrelina

crates/viewer/re_viewer_context/src/drag_and_drop.rs
```diff
@@ -0,0 +1,184 @@
...
+        ("store", "stores"),
+        ("entity", "entities"),
+        ("instance", "instances"),
+        ("component", "components"),
+    ];
+
+    counts
+        .iter()
+        .zip(names.iter())
+        .filter_map(|(&&cnt, &name)| {
```
emilk, line 172:
```suggestion
    let count_and_names = [
        (container_cnt, "container", "containers"),
        (view_cnt, "view", "views"),
        (app_cnt, "app", "apps"),
        (data_source_cnt, "data source", "data sources"),
        (store_cnt, "store", "stores"),
        (entity_cnt, "entity", "entities"),
        (instance_cnt, "instance", "instances"),
        (component_cnt, "component", "components"),
    ];

    count_and_names
        .into_iter()
        .filter_map(|(count, name_singular, name_plural)| {
```

crates/viewer/re_viewer_context/src/drag_and_drop.rs
```diff
@@ -0,0 +1,184 @@
...
+            if cnt > 0 {
+                Some(format!(
+                    "{} {}",
+                    cnt,
```
emilk, line 176: unlikely needed, but just for good practice
```suggestion
                    re_format::format_uint(cnt),
```

crates/viewer/re_viewer_context/src/viewer_context.rs
```diff
@@ -146,14 +153,38 @@ impl ViewerContext<'_> {
...
+            let selection_contains_undraggable_items = self
+                .undraggable_items
+                .iter_items()
+                .all(|item| !selected_items.contains_item(item));
+
+            let payload = if selection_contains_undraggable_items {
+                DragAndDropPayload::from_items(&selected_items)
+            } else {
+                DragAndDropPayload::Invalid
+            };
```
emilk, line 179: This looks reversed

emilk (changes requested):
Code mostly LGTM, but I don't know when to pass `true` or `false` to `handle_select_hover_drag_interactions`. I would like to either have clear instructions, or better yet: have it be automatic (e.g. based on the items passed to it)

emilk (changes requested):
Code mostly LGTM, but I don't know when to pass `true` or `false` to `handle_select_hover_drag_interactions`. I would like to either have clear instructions, or better yet: have it be automatic (e.g. based on the items passed to it)

emilk:
Sometimes when you drag things you get a pill, sometimes you don't

inconsistent-drag

abey79:
> Sometimes when you drag things you get a pill, sometimes you don't

That's a design decision we made because the red pill was worse. The lack of pill is feedback that the current selection is not draggable.

As for this specific view: the next PR will enable dragging entities, so with it, every item in the blueprint tree will be individually draggable (or in homogenous multiple selection). This is not generally true however. For example, in the streams tree the entities will be draggable, but not the components.

crates/viewer/re_ui/src/ui_ext.rs
abey79: Well none is used in the actual implementation as this is not an interactive widget. I could add an explicit `interact()` but it'd add unnecessary complexity imo (id? sense?). User can use `ui.interact()` with the rect if needed.

crates/viewer/re_view_graph/src/ui/draw.rs
abey79: Yes, I want people to think of draggability and make a call when using this function.

It's also a toggle that allows me to progressively implement draggability one UI area at a time (e.g. next PR will switch some items to `true` in the streams tree).

There is an ideal version of the future where this is `true` everywhere (and thus the arg can be removed), but I'm not sure this is realistic. There are cases where stuff should hover/select highlight, but not be draggable. For example, entity buttons should probably not be draggable—or at least not always (e.g. from the add/remove entity modal).

In generally, we should make something draggable only if it makes sense, e.g if there is an obvious drop destination.

I'll improve the docstring to expand on the above. Let me know if you have any further suggestion on this.

crates/viewer/re_viewer_context/src/viewer_context.rs
abey79: > Could the value of this be inferred from the items instead?

No. See my other comment.

crates/viewer/re_viewer_context/src/viewer_context.rs
abey79: That's the most hacky part, I'll take feedback on that.

Currently there is ONE undraggable item, and it's always the same: the root container in the blueprint viewer. The issue is that the root container is only known by "top-of-the-hierarchy" crates (`re_viewer`, etc.), and certainly not by `re_viewer_context`. I'm not really sure how to go about this in a cleaner way.

crates/viewer/re_viewer_context/src/drag_and_drop.rs
abey79: I'll reject this for complexity over value reason. That would require the payload stuff to query the blueprint store to retrieve the name.

If you think it's important, please open an issue.

crates/viewer/re_viewer/src/app_state.rs
abey79: `re_viewport_blueprint` depends on `re_viewer_context`, so that cannot be done. And I would like to avoid merging these crates.

crates/viewer/re_ui/src/ui_ext.rs
emilk: You are right; I thought `allocate_space` returned a response anyways. My bad :)

crates/viewer/re_viewer_context/src/viewer_context.rs
emilk: You can at least document it here exactly what it is was introduced for, and maybe with a TODO for "this is ugly; if you have a better idea please rewrite this" :]

emilk (approved):
Thanks for addressing the feedback!

Issue #8278, opened 2024-12-02
Allow dynamic configuration of forces through blueprints


grtlr:

abey79:
As discussed in today's huddle:

```
=============================================================================
PLAN C:


archetype LinkForce {
    link_force: component.LinkForce (datatype: datatype.LinkForce)
}


datatype LinkForce {
    enabled: ForceEnabled (datatype: Bool)
    strength: ForceStrength (datatype: Float)
    distance: Distance (datatype: Float)
}


TODO:

fn selection_ui() {
    view_property_ui::<LinkForce>(ctx, ui, view_id, self, state);
}

in re_component_ui: single and/or multiline editor for component.LinkForce

custom __init__ for `component.LinkForce`

(maybe optional?) custom native_to_pa_array for `datatype.LinkForce`
   - flexible python API input ==> NEEDED
   - custom serialization compared to the codegen'd one ===> NOT NEEDED
```

PR #8339, opened 2024-12-06
Clean up `ForceLayoutProvider`

grtlr:
### Related

* Preparation for #8278 

### What

This cleans up the layout provider, to prepare for a force blueprints.

### DIFF
--- a/crates/viewer/re_space_view_graph/src/layout/provider.rs
+++ b/crates/viewer/re_space_view_graph/src/layout/provider.rs
@@ -27,62 +27,32 @@ impl<'a> From<&'a NodeTemplate> for fj::Node {
 
 pub struct ForceLayoutProvider {
     simulation: fj::Simulation,
-    node_index: ahash::HashMap<NodeId, usize>,
     pub request: LayoutRequest,
+}
+
+fn considered_edges(request: &LayoutRequest) -> Vec<(usize, usize)> {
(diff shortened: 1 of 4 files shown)


crates/viewer/re_space_view_graph/src/layout/provider.rs
```diff
@@ -27,115 +27,88 @@ impl<'a> From<&'a NodeTemplate> for fj::Node {
...
+        let mut layout = Layout::empty();
 
         for (entity, graph) in &self.request.graphs {
             let mut current_rect = Rect::NOTHING;
 
-            for node in graph.nodes.keys() {
-                let extent = layout.nodes.get_mut(node).expect("node has to be present");
-                let i = self.node_index[node];
+            for (node, template) in &graph.nodes {
+                let i = i.next().expect("can never happen");
```
abey79, line 106: is that a fancy `i += 1`? :)

I wonder if there is a way to rewrite this logic a little bit. Can't you `.join()` `positions` with `graph.nodes` in the `for` statement directly?

abey79 (approved):

crates/viewer/re_space_view_graph/src/layout/provider.rs
grtlr: Oh, I forgot about slices—that's a good idea!

PR #8341, opened 2024-12-06
Fix handling unnormalized axis for (Pose)RotationAxisAngle

Wumpf:
As well as dodging associated debug-assertions for
* unnormalized/zero axis for rotation axis angle
* non-invertible transforms (which may be the case for all sort of reasons)

### DIFF
--- a/crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
+++ b/crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
@@ -220,8 +220,15 @@ impl MeshDrawData {
                     count_with_outlines += instance.outline_mask_ids.is_some() as u32;
 
                     let world_from_mesh_mat3 = instance.world_from_mesh.matrix3;
+                    // If the matrix is not invertible the draw result is likely invalid as well.
+                    // However, at this point it's really hard to bail out!
+                    // Also, by skipping drawing here, we'd make the result worse as there would be no mesh draw calls that could be debugged.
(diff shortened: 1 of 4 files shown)


crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
```diff
@@ -220,8 +220,15 @@ impl MeshDrawData {
...
                     let world_from_mesh_mat3 = instance.world_from_mesh.matrix3;
+                    // If the matrix is not invertible the draw result is likely invalid as well.
+                    // However, at this point it's really hard to bail out!
+                    // Also, by skipping drawing here, we'd make the result worse as there would be no mesh draw calls that could be debugged.
                     let world_from_mesh_normal =
-                        instance.world_from_mesh.matrix3.inverse().transpose();
+                        if instance.world_from_mesh.matrix3.determinant() != 0.0 {
+                            instance.world_from_mesh.matrix3.inverse().transpose()
+                        } else {
+                            glam::Mat3A::ZERO
```
abey79, line 230: Just to be sure, you meant it to be zero and not identity?

abey79 (approved):

crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
Wumpf: yes. zero. Which is the most likely reason the determinant was zero to begin with 🤷

Issue #7988, opened 2024-11-04
Replace our serde/MsgPack encoding of `enum LogMsg` with protobuf


emilk:
We want to replace this:

https://github.com/rerun-io/rerun/blob/7fe82400864c21faf6ad4bf431f52a1ea60c5bf8/crates/store/re_log_types/src/lib.rs#L257-L279

with something protobuf, ideally inspired/overlapping with our gRPC datatypes.

That means our .rrd files become just a bunch of protobuffs concatted together, and will be very similar (or same) as the gRPC messages we send to/from our data platform

PR #8347, opened 2024-12-06
Encode `LogMsg` using protobuf

jprochazk:
### Related

* Closes https://github.com/rerun-io/rerun/issues/7988
* Follow-up to https://github.com/rerun-io/rerun/pull/8381

### What

This PR introduces `Serializer::Protobuf` to `re_log_encoding`, and inverts the dependency graph of `re_protos`, which no longer depends on other `re_*` crates. This meant the conversion impls from protobuf types to rerun types and back had to be moved into their respective crates. For example, `From<StoreId> for `re_protos::common::v0::StoreId` is now in `re_log_types`.

When encoding a file using this serializer, the data is encoded using a combination of:
* A custom stream-level protocol
* Protocol buffers
* Arrow IPC

The stream-level protocol has changed only a bit, because compression is no longer done for all messages in the stream, and only the contents of `ArrowMsg` are ever compressed at all. This means the `uncompressed_len` and `compressed_len` could be unified to just `len`.

The actual layout of the messages has not changed, `LogMsg` is preserved and so are its semantics.

The stream of data stored in an example RRD file using this new encoding looks like:
```
FileHeader { b"RRIO", version, compression, serializer }     ;; 10 bytes
MessageHeader { kind, len }                                  ;; 8 bytes
SetStoreInfo { application_id, store_id, store_source, ... } ;; len bytes
MessageHeader { kind, len }                                  ;; 8 bytes
ArrowMsg { store_id, arrow_msg }                             ;; len bytes
MessageHeader { kind: End, len: 0 }                          ;; 8 bytes
```

Note that this stream-level protocol is _only_ used for `.rrd` files. On the wire, we will use gRPC, which has its own protocol.

In the case of `ArrowMsg`, the schema+chunk is encoded using Arrow IPC into a byte payload, which may additionally be compressed. The compression setting is stored separately for every `ArrowMsg`, but per-message compression functionality is not yet exposed through `re_log_encoding`.


crates/store/re_chunk_store/src/protobuf_conversions.rs
```diff
@@ -0,0 +1,260 @@
...
+                .using_index_values
+                .map(|uiv| uiv.time_points.into_iter().map(|v| v.into()).collect()),
+            filtered_is_not_null,
+            sparse_fill_strategy: crate::SparseFillStrategy::default(), // TODO(zehiko) implement support for sparse fill strategy
+            selection,
+        })
+    }
+}
+
+impl From<crate::QueryExpression> for re_protos::common::v0::Query {
```
zehiko, line 186: nit: move unit test as well?

crates/store/re_log_encoding/src/decoder/mod.rs
```diff
@@ -8,27 +8,15 @@ use std::io::Read;
...
 use crate::MessageHeader;
+use crate::VersionPolicy;
 use crate::OLD_RRD_HEADERS;
 use crate::{Compression, EncodingOptions, Serializer};
 
 // ----------------------------------------------------------------------------
 
-/// How to handle version mismatches during decoding.
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub enum VersionPolicy {
```
zehiko, line 20: (could we have done this move in a separate small prep PR just to make this one slightly smaller and focused? )

zehiko (commented):
thanks for untangling these dependency issues!

generally looks ok to me, few comments and a question about more clear re-use of common arrow seriailzation logic to make ``re_log_encoding`` crate clearer.

crates/store/re_log_encoding/src/protobuf/decoder.rs
```diff
@@ -0,0 +1,136 @@
...
+            uncompressed.as_slice()
+        }
+    };
+
+    Ok(read_arrow_from_bytes(&mut &data[..])?)
+}
+
+/// Helper function that deserializes raw bytes into arrow schema and record batch
+/// using Arrow IPC format.
+fn read_arrow_from_bytes<R: std::io::Read>(
```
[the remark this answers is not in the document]
jprochazk: re: `decoder` `encoder` being in two places, that's because we still have to continue supporting msgpack for the time being. So all the new protobuf encoding stuff is under `protobuf`.

re: `codec`, it's there separately because that's the format used by gRPC, with its own `MessageHeader`. the existing `encoder` and `decoder` will eventually only be used for files (either local or served over HTTP), so I thought it'd be better to keep them separate.

The naming could be clearer, I can reshuffle the modules to make it more obvious what's going on.
When it comes to the duplication of arrow serialization, it could be in a common module, now that we've decided (have we?) to send the schema separately with each message.

crates/store/re_protos/proto/rerun/v0/log_msg.proto
```diff
@@ -0,0 +1,105 @@
...
+    LZ4 = 1;
+}
+
+enum Encoding {
+    UNKNOWN = 0;
+    ARROW_IPC = 1;
+}
+
+message ArrowMsg {
+    rerun.common.v0.StoreId store_id = 1;
```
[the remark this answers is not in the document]
jprochazk: It's because the package names are different, I asked about this during the meetup and the consensus was to leave them in separate packages and use the fully-qualified path, because protobuf doesn't support aliasing these paths in any way.

crates/store/re_log_encoding/src/decoder/mod.rs
jprochazk: Yes, and it's still possible. As I submitted it, I honestly didn't think the PR would be big enough for it to matter, but at the very least I could've polished up the commit history to make it easier to review. I'll keep that in mind for future PRs, or if you want I can split up this one as well

jprochazk: Actually, I will just split it up and untangle the dependencies in a separate PR first.

crates/store/re_log_encoding/src/protobuf/decoder.rs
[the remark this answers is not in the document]
zehiko: yeah, I think just some naming reshuffling might help to make things more obvious. Perhaps even worth splitting grpc ``codec`` into encoder and decoder just for consistency...

crates/store/re_protos/proto/rerun/v0/log_msg.proto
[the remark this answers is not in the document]
zehiko: ah right, thanks!

crates/store/re_log_encoding/src/decoder/mod.rs
zehiko: thanks, I think it will reduce the size of this one and make it easier to review

jprochazk:
Ideally this time it should be reviewable commit-by-commit after [`5d9051b` (#8347)](https://github.com/rerun-io/rerun/pull/8347/commits/5d9051b3f5fc7cdc6fbe679d02f03fab29d2344e)

crates/store/re_grpc_client/src/lib.rs
```diff
@@ -151,7 +151,7 @@ async fn stream_recording_async(
             .connect()
             .await?;
 
-        StorageNodeClient::new(tonic_client)
+        StorageNodeClient::new(tonic_client).max_decoding_message_size(usize::MAX)
```
zehiko, line 154: nit: comment + TODO + issue to figure out the right size?

crates/store/re_log_encoding/src/codec/file/decoder.rs
```diff
@@ -0,0 +1,107 @@
...
+    compression: crate::Compression,
+) -> Result<
+    (
+        arrow2::datatypes::Schema,
+        arrow2::chunk::Chunk<Box<dyn re_chunk::Arrow2Array>>,
+    ),
+    DecodeError,
+> {
+    let mut uncompressed = Vec::new();
+    let data = match compression {
```
zehiko, line 98: I have bad memory and don't remember if we decided anything on this one, but did we explore using arrow ipc compression?

crates/store/re_log_encoding/src/codec/file/decoder.rs
```diff
@@ -0,0 +1,107 @@
...
+        let kind = MessageKind::decode(data)?;
+        let mut buf = [0; 4];
+        data.read_exact(&mut buf)?;
+        let len = u32::from_le_bytes(buf);
+
+        Ok(Self { kind, len })
+    }
+}
+
+pub(crate) fn decode(
```
zehiko, line 34: can we please add unit tests for all message types, and also for encoding?

crates/store/re_log_types/src/protobuf_conversions.rs
```diff
@@ -189,3 +189,322 @@ impl From<crate::StoreId> for re_protos::common::v0::RecordingId {
         }
     }
 }
+
+impl From<crate::StoreSource> for re_protos::log_msg::v0::StoreSource {
```
zehiko, line 193: could we add some basic unit tests for these conversions below?

crates/store/re_protos/proto/rerun/v0/log_msg.proto
```diff
@@ -0,0 +1,178 @@
+syntax = "proto3";
+
+package rerun.log_msg.v0;
+
+import "rerun/v0/common.proto";
+
+// Corresponds to `LogMsg::SetStoreInfo`. Used to identify a recording.
+message SetStoreInfo {
```
zehiko, line 8: I was hoping that ``SetStoreInfo`` would go away, mostly for selfish reasons of exposing ``GetStoreInfo`` on the SN side that would return ``StoreInfo``. Reading the comment on the ``row_id`` field of the current ``SetStoreInfo`` struct doesn't help me. 

@teh-cmc  some help here, do we need this still?

crates/store/re_protos/proto/rerun/v0/log_msg.proto
```diff
@@ -0,0 +1,178 @@
...
+    NONE = 0;
+
+    // LZ4 block compression.
+    LZ4 = 1;
+}
+
+// The encoding of the message payload.
+enum Encoding {
+    // We don't know what encoding the payload is in.
+    UNKNOWN = 0;
```
zehiko, line 28: are we adding this just because grpc suggests it's best practice to have 0 enum member? I've seen this used and unused in few big open source projects.  I just wonder if it will create unnecessary overhead and if we should just keep the one option we have, which is arrow ipc.

crates/store/re_log_encoding/src/lib.rs
```diff
@@ -62,6 +78,10 @@ impl EncodingOptions {
         compression: Compression::LZ4,
         serializer: Serializer::MsgPack,
     };
+    pub const PROTOBUF: Self = Self {
```
zehiko, line 81: nit: might I suggest we rename these to ``MSGPACK_UNCOMPRESSED``, ``MSGPACK_COMPRESSED``, ``PROTOBUF_COMREPSSED``?

crates/store/re_log_encoding/src/encoder.rs
zehiko, line 1: I like where the file/module structure is going, but I'm tiny bit confused on still having top level ``encoder`` and ``decoder`` modules? are these temporary until we move streams like TCP to gRPC? or I'm missing something?

zehiko (commented):
nice work!

Few comments and questions:
- could add a unit/integration test that uses new codec to write and read rrd stream using Protobuf? something similar-ish to ``test_loading_with_retryable_reader``
- suggested adding a few basic unit tests in a few places
- arrow ipc compression -> could we use that? (I just noticed in the protobufication doc Jeremy asked the same question :)
- SetStoreInfo - could we get rid of it? (or should I just learn and understand its purpose better)

crates/store/re_log_encoding/src/codec/file/decoder.rs
jprochazk: We have not experimented with this. I'd like to leave that out for now, created an issue:
- https://github.com/rerun-io/rerun/issues/8412

crates/store/re_protos/proto/rerun/v0/log_msg.proto
jprochazk: Removing `StoreInfo` is a bit complicated, but it will eventually be done. Currently we may have multiple recordings per `rrd` file, and changing that would have far-reaching consequences that I wasn't comfortable with for this PR.

crates/store/re_protos/proto/rerun/v0/log_msg.proto
jprochazk: Yes, that's one of the reasons why I added this. We also have some enums on the Rust side that already have an `unknown` variant (such as `StoreSource`), so I felt like it'd make sense for consistency.

crates/store/re_log_encoding/src/encoder.rs
jprochazk: Yes, these are temporary. Eventually all of the file stuff should live in `codec/file/...`

crates/store/re_log_encoding/src/codec/file/decoder.rs
jprochazk: The encoding/decoding is tested as part of the tests in `re_log_encoding/src/decoder/mod.rs`, but now I realise that isn't thorough enough as it doesn't try to encode an actual chunk with data in it. Will do more testing.

zehiko: thank you!

crates/store/re_protos/proto/rerun/v0/log_msg.proto
zehiko: ok, cool, if there's any other enum variant in our spec that doesn't have 0 field, we should add it (follow up, ofc)

crates/store/re_protos/proto/rerun/v0/log_msg.proto
jprochazk: Yes sorry, I meant `SetStoreInfo`. Its current purpose is to support multiple recordings per rrd. That's pretty much it.

The way this should look in the end is outlined in Notion: https://www.notion.so/rerunio/Protobufication-143b24554b19803e83acd5cc34bd5219?pvs=4#143b24554b19801cad70ce9327a2f651 - but there are no concrete plans on how to get there. And I'll definitely need help figuring that out :smile:

jprochazk:
> could add a unit/integration test that uses new codec to write and read rrd stream using Protobuf? something similar-ish to test_loading_with_retryable_reader

This is now tested as part of the tests in `re_log_encoding/src/decoder/mod.rs`. I updated the fake log messages there to actually encode a more realistic scenario (blueprint that sets the background color, though I'm sure it's wrong as I haven't actually tested if it does that, it's just meant to be some data to encode/decode), to exercise encoding for all 3 `LogMsg` variants.

> suggested adding a few basic unit tests in a few 

Done for `re_log_types` and `re_tuid`, there was already some test in `re_chunk_store`

crates/build/re_protos_builder/src/bin/build_re_remote_store_types.rs
```diff
@@ -1,15 +1,15 @@
 //! This binary runs the remote store gRPC service codegen manually.
 //!
-//! It is easiest to call this using `pixi run codegen-protos`,
+//! It is easiest to call this using `pixi run codegen-protos   `,
```
emilk, line 3:
```suggestion
//! It is easiest to call this using `pixi run codegen-protos`,
```

crates/store/re_log_encoding/benches/msg_encode_benchmark.rs
```diff
@@ -32,7 +32,7 @@ criterion_group!(
 criterion_main!(benches);
 
 fn encode_log_msgs(messages: &[LogMsg]) -> Vec<u8> {
-    let encoding_options = re_log_encoding::EncodingOptions::COMPRESSED;
+    let encoding_options = re_log_encoding::EncodingOptions::MSGPACK_COMPRESSED;
```
emilk, line 35: Wouldn't it make more sense to benchmark the new "standard" method, i.e. `PROTOBUF_COMPRESSED`? That way we can test the old vs new performance

crates/store/re_log_encoding/src/codec/arrow.rs
```diff
@@ -0,0 +1,106 @@
+use crate::decoder::DecodeError;
+use crate::encoder::EncodeError;
+
+use super::CodecError;
+
+// TODO(#8412): try using arrow ipc `compression` option instead of doing our own compression
```
emilk, line 6: This is linking to a closed issue.

crates/store/re_log_encoding/src/codec/arrow.rs
```diff
@@ -0,0 +1,106 @@
...
+    compression: crate::Compression,
+) -> Result<Payload, EncodeError> {
+    let mut uncompressed = Vec::new();
+    write_arrow_to_bytes(&mut uncompressed, schema, chunk)?;
+    let uncompressed_size = uncompressed.len();
+
+    let data = match compression {
+        crate::Compression::Off => uncompressed,
+        crate::Compression::LZ4 => lz4_flex::block::compress(&uncompressed),
+    };
```
emilk, line 76: Where is the compression communicated to the receiver? The natural place would be to prepend each message with the encoding here, but I guess that happens later? If so, it should at least be part of `Payload` to make that obvious

crates/store/re_log_encoding/src/codec/file/decoder.rs
```diff
@@ -0,0 +1,86 @@
...
+            let (schema, chunk) = decode_arrow(
+                &arrow_msg.payload,
+                arrow_msg.uncompressed_size as usize,
+                arrow_msg.compression().into(),
+            )?;
+
+            let store_id: re_log_types::StoreId = arrow_msg
+                .store_id
+                .ok_or_else(|| {
+                    TypeConversionError::missing_field("rerun.log_msg.v0.ArrowMsg", "store_id")
```
emilk, line 65: Where does the `v0` come from? Is there a way not to hard-code it here?

crates/store/re_log_encoding/benches/msg_encode_benchmark.rs
emilk: Still, would be interesting to see how `PROTOBUF_COMPRESSED` and `MSGPACK_COMPRESSED` compare to each other in encode/decode speed

emilk (commented):
I only had time for a partial review today - will take a closer look tomorrow!

crates/store/re_log_encoding/src/codec/file/decoder.rs
jprochazk: It's part of package name, I wanted to be explicit, because while debugging this I was bitten by it not being explicit enough.
It turns out that `prost_build` can generate an impl of `prost::Name` which has these strings statically defined (but weirdly enough `tonic_build` does not expose that option, even though it works), so I made a macro that accepts the type and field name instead, and uses the constants from `prost::Name` in the error message.

jprochazk: I wish we could auto-generate this code instead, that would be a lot better, but it seems difficult unless the types map 1:1 exactly.

crates/store/re_log_encoding/src/codec/arrow.rs
```diff
@@ -0,0 +1,106 @@
...
+// TODO(#8412): try using arrow ipc `compression` option instead of doing our own compression
+
+/// Helper function that serializes given arrow schema and record batch into bytes
+/// using Arrow IPC format.
+pub(crate) fn write_arrow_to_bytes<W: std::io::Write>(
+    writer: &mut W,
+    schema: &arrow2::datatypes::Schema,
+    data: &arrow2::chunk::Chunk<Box<dyn re_chunk::Arrow2Array>>,
+) -> Result<(), CodecError> {
+    use arrow2::io::ipc;
```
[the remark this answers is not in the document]
jprochazk: It doesn't seem to have an equivalent API, so I'm leaving it out for the time being

crates/store/re_log_encoding/src/codec/arrow.rs
jprochazk: It's contained in `ArrowMsg`, it has both a `compression` and `encoding` field. Right now compression is also stored in `FileHeader`, but that will change at some point and it'll be only per-message compression. I didn't want to change the layout too much, otherwise I'd have to also break the `rrd` format to some extent, or complicate the decoding even more by matching the magic bytes and use completely different decoders for older rrd versions. I will still have to do that, just wanted to leave it out of this PR, and for the follow-up where we switch this on.

crates/store/re_log_encoding/src/codec/arrow.rs
```diff
@@ -0,0 +1,106 @@
+use crate::decoder::DecodeError;
+use crate::encoder::EncodeError;
+
+use super::CodecError;
+
+// TODO(#8412): try using arrow ipc `compression` option instead of doing our own compression
+
+/// Helper function that serializes given arrow schema and record batch into bytes
+/// using Arrow IPC format.
+pub(crate) fn write_arrow_to_bytes<W: std::io::Write>(
```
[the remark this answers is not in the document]
jprochazk: Yes, it's already used there. I didn't want to change the wire protocol in any way so that storage node continues to work as-is. I'm not aware of it doing any compression at the moment.

IIUC, the plan is to remove `TransportMessageV0` and just have a stream of `ArrowMsg` at some point. `ArrowMsg` contains the `compression` field. So both the gRPC client and file would use the exact same code paths then.

jprochazk:
I updated the benchmark to include protobuf variants of all the encode/decode parts. There are way too many benchmarks now, so I tried using ChatGPT to make sense of the results:

image

They seem fairly mixed, some wins and some losses. The mono-points non-batched vs batched decode is interesting, as msgpack seems to be better at non-batched data? Encode seems to be a win across the board for protobuf, but only by a little.

crates/store/re_protos/src/v0/rerun.log_msg.v0.rs
```diff
@@ -0,0 +1,388 @@
+// This file is @generated by prost-build.
```
emilk, line 1: Then is should be marked as `linguist-generated=true` in a `.gitattributes` file. It would be great if that file was generated/maintained by `build.rs` so it keeps up-to-date

crates/store/re_log_types/src/lib.rs
```diff
@@ -407,6 +407,67 @@ impl std::fmt::Display for PythonVersion {
     }
 }
 
+impl std::str::FromStr for PythonVersion {
```
emilk, line 410: This should have a unit test

crates/store/re_log_encoding/src/decoder/mod.rs
```diff
@@ -442,35 +520,41 @@ mod tests {
...
+                serializer: Serializer::Protobuf,
+            },
+            EncodingOptions {
+                compression: Compression::LZ4,
+                serializer: Serializer::Protobuf,
+            },
         ];
 
         for options in options {
+            println!("{options:?}");
```
emilk, line 534: 😬 

Tip: add a `// TODO` suffix on code you plan to remove (the linter will stop you from merging it).

Or use `dbg!(options);` (again, this will not pass CI)

You can also just use `re_log::trace` and run with `RUST_LOG=re_log_encoding=trace`

crates/store/re_protos/src/v0/rerun.common.v0.rs
```diff
@@ -11,18 +11,48 @@ pub struct RerunChunk {
     #[prost(bytes = "vec", tag = "1000")]
     pub payload: ::prost::alloc::vec::Vec<u8>,
 }
+impl ::prost::Name for RerunChunk {
```
emilk, line 14: This file too should be marked `linguist-generated=true`. Maybe just mark the whole folder?

crates/store/re_log_types/src/protobuf_conversions.rs
```diff
@@ -82,35 +79,47 @@ impl TryFrom<re_protos::common::v0::IndexRange> for crate::ResolvedTimeRange {
...
+impl From<re_protos::common::v0::Timeline> for crate::Timeline {
+    fn from(value: re_protos::common::v0::Timeline) -> Self {
+        // TODO(cmc): QueryExpression::filtered_index gotta be a selector
+        #[allow(clippy::match_same_arms)]
+        match value.name.as_str() {
+            "log_time" => Self::new_temporal(value.name),
+            "log_tick" => Self::new_sequence(value.name),
+            "frame" => Self::new_sequence(value.name),
+            "frame_nr" => Self::new_sequence(value.name),
+            _ => Self::new_temporal(value.name),
```
emilk, line 99: Wait, what the hack is this @teh-cmc ??

emilk (approved):
👍

crates/store/re_protos/src/v0/rerun.log_msg.v0.rs
emilk: We could probably mark the whole folder as `linguist-generated=true`

crates/store/re_log_encoding/src/decoder/mod.rs
jprochazk: I didn't plan to remove this, what's wrong with keeping a `println` in a test? It doesn't get printed unless the test fails, in which case it's useful information about which part of the test failed

crates/store/re_protos/src/v0/rerun.log_msg.v0.rs
jprochazk: image

PR #8348, opened 2024-12-06
Make the near clipping plane editable in 2D views

jleibs:
### What
This adjusts the value to now be in physical scene units. I.e. for most 3d scenes in meters, the clip plane will be 10cm, which is a very reasonable default.

Additionally, this gives users an escape hatch for scenes where the default clipping plane causes their data to not be visible.

Test script:
```python
import rerun as rr
import rerun.blueprint as rrb
import numpy as np

rr.init("rerun_example_pinhole_perspective", spawn=True)

img = np.zeros((300, 300, 3), dtype=np.uint8)

cam_proj = [152.34006, 0, 175.5, 0, 152.34006, 175.5, 0, 0, 1]

rr.log(
    "world/cam",
    rr.Pinhole(
        image_from_camera=cam_proj, width=300, height=300, image_plane_distance=0.1
    ),
)

rr.log("world/cam", rr.Image(img))

rr.log(
    "world/points",
    rr.Points3D([(0.0, 0.5, 0.1), (0.1, 0.1, 0.3), (-0.1, -0.1, 0.5)], radii=0.025),
)
rr.log_file_from_path("examples/assets/example.gltf")
rr.log("examples/assets", rr.Transform3D(translation=[0, 0, 0.4], scale=0.05))

rr.send_blueprint(
    rrb.Horizontal(
        rrb.Spatial3DView(),
        rrb.Spatial2DView(
            origin="world/cam",
            contents="/**",
            visual_bounds=rrb.VisualBounds2D(
                near_clip_plane=0.01,
            ),
        ),
    )
)
```

### New issues:
- https://github.com/rerun-io/rerun/issues/8373


Wumpf:
> Any reason not to use a smaller default like 0.01?

We get flickering when zooming in on some scenes. Objectron showed that quite well back then.
All this comes from the fact that we render our 2d "background" at a fixed depth, meaning we have to sandwich the entire 3D scene between that depth & the fixed camera position -> we throw a lot of precision out of the window :(
This can be solved once we get rid of depth-offsetting for 2D which can be done once we've done draw order in the renderer which in turn again is part of the transparency story.

crates/store/re_types/definitions/rerun/blueprint/archetypes/visual_bounds2d.fbs
```diff
@@ -16,4 +16,7 @@ table VisualBounds2D (
     ///
     /// Use this to control pan & zoom of the view.
     range: rerun.blueprint.components.VisualBounds2D ("attr.rerun.component_required", order: 1000);
+
+    /// Controls the distance to the near clipping plane
+    clipping_plane: rerun.blueprint.components.ClippingPlane ("attr.rerun.component_optional", order: 2000);
```
Wumpf, line 21: maybe better to call the field `near_clip_plane` - `near` for descriptiveness, `clip` for brevity (and the type then `ClipPlane`)

(?)

Wumpf (approved):
lgtm

Also unsure about the visual bounds aspect 🤔. But that's the lowest friction place right now, we can always move it around again 🤷

Issue #8336, opened 2024-12-06
Multi-selection selection panel reborn


abey79:
#### Context

The current implementation of the selection panel is comically useless when multiple items are selected, and incurs undue code complexity (e.g. `UiLayout::SelectionPanelLimitHeight`). For the longest time, our dream has been to come up with some "merged" view of the current selection, but this hardly is actionable as a starting point:
- https://github.com/rerun-io/rerun/issues/4251

#### Proposal

This issue is about getting the ball rolling on this front with a two-step, actionable, minimalistic take on the  selection panel with multi-selection:

1) show a list of what is currently selected
2) ...and that's basically it

emilk:
Current attempt:
![Image](https://github.com/user-attachments/assets/c2aa2750-8a26-45ef-aa6e-cd413cf50026)

Issue #6761, opened 2024-07-04
Can't select/click a 2D primitive without also selecting underlying image


emilk:
When I hover a 2D primitive, like a box or a point, I expect only to hover that thing, not also all images under it.

![image](https://github.com/rerun-io/rerun/assets/1148717/2d4e37d7-f869-4c39-9f2a-75bc29177e28)

Affects 2D spatial views, and maybe 3D ones too.

This is probably related to draw order and depth offsets.

---

The offending code is this:

https://github.com/rerun-io/rerun/blob/9b0c4a22834fe6679dc92773efcdad59927815d9/crates/re_space_view_spatial/src/picking.rs#L144-L146

There is no motivation for the behavior in the code, but the blame points to

* https://github.com/rerun-io/rerun/pull/1908

Which was created to fix

* https://github.com/rerun-io/rerun/issues/1885

nikolausWest:
I don't actually agree with this expectation. If I run an object detector on an image, the edges of the bounding boxes are the most interesting parts of the image to inspect. If we only hover one thing at a time, it makes it super hard to look closely at around those borders. We used to have it that we only hovered one entity at a time but changed that because it made it really hard to debug for instance detections on images

emilk:
I can understand it for hovering, but for clicking it becomes quite confusing, and will become even more confusing when we have solid primitives. Say you click the middle of a solid-filled rectangle. Then you see an image in the selection view. WTH!?

In either case, I suggest we don't close this until we explain in the code (and maybe in the UI) that this behavior is intentional.

nikolausWest:
Fair enough. I think it would be pretty reasonable for clicking to just select the topmost thing but hovering going all the way through

abey79:
I suggest we bite the bullet and roll a heuristic-y approach here.

1) For hover, don't change the current behaviour (aka _everything_ that's under the mouse gets highlighted).

2) For click/cmd-click:
  - if the picked items are all image-like, select them all
  - if the picked items are a mix of image-like and primitive-like (point, line, etc.), ignore the image-like ones and select only the primitive-like ones

PR #8351, opened 2024-12-08
More compact selection panel when multiple items selected

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/8336

### What
When you have more than one item selected at once, you get a list of all items. Clicking one will select just that.

multiple-selected-entities

### Implementation
A bit messy, to be honest. I leveraged the existing `ItemTitle` for some items, and hand-rolled others.

It would be nice with something similar to `DataUi`, but for titles (with or without breadcrumbs).

But the amount of messy code is low and low-impact, so unless someone has a good suggestion or strongly objects, I suggestion calling this "good enough for now".

### DIFF
--- a/crates/viewer/re_selection_panel/src/selection_panel.rs
+++ b/crates/viewer/re_selection_panel/src/selection_panel.rs
@@ -23,7 +23,7 @@ use re_viewport_blueprint::{ui::show_add_space_view_or_container_modal, Viewport
 
 use crate::{
     defaults_ui::view_components_defaults_section_ui,
-    item_heading_no_breadcrumbs::item_heading_no_breadcrumbs,
+    item_heading_no_breadcrumbs::item_title_list_item,
     item_heading_with_breadcrumbs::item_heading_with_breadcrumbs,
     space_view_entity_picker::SpaceViewEntityPicker,
     visible_time_range_ui::{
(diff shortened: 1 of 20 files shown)


Wumpf:
I like it otherwise but this is quite the step backwards for interacting with objects in 2d views with images :/

emilk:
> this is quite the step backwards for interacting with objects in 2d views with images :/

I believe you are referring to the problem outlined in this issue:
* https://github.com/rerun-io/rerun/issues/6761

Clicking e.g. a point in a 2D view would _also_ select all images below the point:

image

->

image

With this PR, the same click now produces this selection panel:

image

---

The old behavior was already really bad, imho. Why would clicking a point select the image behind the point? That's just a bug imho.

Clicking two translucent images is arguably different. But is it really better to get a list of expanded items, forcing you to scroll, or to show a concise list of things under the mouse, prompting you to refine the selection with a second click?

abey79:
> I like it otherwise but this is quite the step backwards for interacting with objects in 2d views with images :/

Is it though? Sure you need one more click to inspect that point/box2d thing. But the entire process makes so much more sense now. Click places -> stuff there is shown to me -> can make educated choice on what really is of interest.

That being said, images are generally very easy to select by themselves, so I wouldn't be against filtering them out of any picking multi-selection.

crates/viewer/re_selection_panel/src/selection_panel.rs
```diff
@@ -104,32 +105,73 @@ impl SelectionPanel {
...
+                        .interactive(true)
+                        .show_flat(
+                            ui,
+                            list_item::CustomContent::new(|ui, context| {
+                                ui.allocate_new_ui(
+                                    egui::UiBuilder::new()
+                                        .max_rect(context.rect)
+                                        .layout(egui::Layout::left_to_right(egui::Align::Center)),
+                                    |ui| {
+                                        ui.spacing_mut().item_spacing.x = 4.0;
```
abey79, line 161: right drift goes brrrrrrt ;)

abey79 (approved):
Very nice!

crates/viewer/re_selection_panel/src/selection_panel.rs
emilk: 😬

PR #8299, opened 2024-12-03
Dynamic configuration of graph layout forces through blueprints

grtlr:
### Related

* Closes #8278

### What

Title.

PR #8357, opened 2024-12-09
Make Position2D components editable in selection panel

grtlr:
### Related

* Part of #8299 

### What

This adds a simple single line edit for position 2D. There is no `speed_fn` because I think in most applications UI coordinates are probably good unit (for example in graphs).

https://github.com/user-attachments/assets/2fad00c4-8e68-40fe-a474-32fa5806814e


Wumpf (changes requested):
those brackets are a bit of a stylistic break. Which I don't actually mind, but this should be integrated & aligned with https://github.com/rerun-io/rerun/blob/main/crates/viewer/re_component_ui/src/datatype_uis/vec.rs

I.e. that file should be extended with vec2 and then used instead

Wumpf:
oh interesting, so actually it _preserves_ the style better than what's there in vec.rs
It's just that for edit boxes we so far didn't want to spend so much horiztonal space 🤔

Wumpf:
oops sorry mouse slipped and clicked the wrong button

crates/viewer/re_component_ui/src/datatype_uis/vec.rs
```diff
@@ -12,50 +25,80 @@ pub fn edit_or_view_vec3d(
...
+        let response = response_y | response_x | response_z;
+
+        if response.changed() {
+            *value = datatypes::Vec3D([x_edit, y_edit, z_edit]);
+        }
+
+        response
+    } else {
+        ui.label(format!(
+            "[ {} , {}, {} ]",
```
Wumpf, line 98:
```suggestion
            "[ {}, {}, {} ]",
```

... or is it?

```suggestion
            "[ {} , {} , {} ]",
```
that's what the 2d version uses :)

Wumpf (approved):
looking good, thanks for unifying things! :)

image

PR #8359, opened 2024-12-09
Update web embedding docs

jprochazk:
### What

* Add a line about what `RRD_URL` can be when used in an iframe

It's the same everywhere, but it's normal to assume that the iframe has different limitations compared to embedding the web viewer directly.

Also bumped the version, as it was pretty ancient.


Wumpf (approved):
nice! 👍

PR #8363, opened 2024-12-09
remove unused dependencies in re_log_types

astraw:
### What

This removes a couple unused crates from re_log_types's Cargo.toml.


Wumpf (approved):
nice, thank you so much!
Unfortunate that our ci based `cargo machete` pass wasn't able to find these, but makes sense (`similar-asserts` is a dev dependency only and `crossbeam` was done conditionally)

Wumpf:
(ci failure due to security advisory that has been fixed on main already)

Issue #8330, opened 2024-12-05
Rendering glitch in 3D view with selection


emilk:
Using a non 1:1 pixel:point ratio and selecting a box, I see artifacts on the opposite side of the view:

![Image](https://github.com/user-attachments/assets/439f0926-a9d3-4651-91eb-08402f10174d)

Looks like maybe something is using a wrapped sampler instead of a clamped one?

PR #8369, opened 2024-12-09
Fix 2D/3D view artifacts on view's border when using fractional zoom

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/8330

### What

Wasn't able to repro the issue in the first place though - tried fractional zoom and a contrast heavy background to no avail
* [x] confirm this fixes it


abey79 (approved):
I'm deeply unqualified with that gpu minutiea, but since this one has been waiting for quite some time, I'm gonna apply my stamp of approval to give a confident @Wumpf license for merging :)

Code lgtm. Maybe there could be a comment somewhere to explain the phenomenon and its fix?

Wumpf:
@abey79 did you have repro of the issue earlier? I failed to reproing it so I'd like to confirm whether this fixes it

PR #8371, opened 2024-12-09
Add missing screenshots of graph view archetypes

### LINKED ISSUE #8237: Graph view zooms out when viewer size changes
The graph viewer zooms in when changing the size of the viewer. Instead it should remain at constant zoom level, potentially with clipped nodes.

### LINKED ISSUE #8277: Add example images for the `GraphEdges`, `GraphNodes`, and `GraphView` archetypes
IMPORTANT: use the in-app screenshot feature (not full window mac screenshot)

Similar todo:
- [x] make sure front matter is set on all languages for `graph_lattice` example
- [x] redo screenshot for archetypes in title
- [x] add 120w url from said archetype to their fbs' docstrings

(note: test docs: `LOCAL_BASE_PATH=/path/to/local/rerun pnpm dev`)

### LINKED ISSUE #8327: Graph view saves `VisualBounds2D` to blueprint on any hover
This is easily verified using the blueprint inspection panel.

grtlr:
### Related

* Closes #8277
* Closes #8327
* Closes #8237

### What

This fixes many problems around the state of the graph view. Turns out the scene rect alone is not sufficient to store all of the viewer state (who would have thought 😇).


abey79 (changes requested):
I'm not fan of the new behaviour, but I'm a bit torn on the appropriate way to go about it.

At the very least, resizing the view should maintain the center point of the view (e.g. if the view is centred on a graph, resizing the view keeps the view centred on that graph).

I general, I also tend to prefer that resizing the view also adjust the scale factor to maintain the current "field of view". However our scaling is not the greatest, and it's cap to zoom level 1. So maybe we dont do that part.

abey79 (approved):
Thanks!

PR #8372, opened 2024-12-09
Prepare to port codegen deserialization from arrow2

emilk:
### Related
 
### What
Some preparatory work for migrating the codegen deserializer from `re_arrow2`


emilk:
@rerun-bot full-check

PR #8374, opened 2024-12-09
Fix indicators showing up in override ui

Wumpf:
The all components/recommended component list would previously contain a fantasy-component with an unqualified name instead of the indicator.

### DIFF
--- a/crates/build/re_types_builder/src/codegen/rust/api.rs
+++ b/crates/build/re_types_builder/src/codegen/rust/api.rs
@@ -1097,7 +1097,7 @@ fn quote_trait_impls_for_archetype(obj: &Object) -> TokenStream {
         #recommended_descriptors
         ComponentDescriptor {
             archetype_name: Some(#archetype_name.into()),
-            component_name: #indicator_name.into(),
+            component_name: #quoted_indicator_name::DEFAULT.descriptor().component_name,
             archetype_field_name: None,
         },
     };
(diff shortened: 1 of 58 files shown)


crates/build/re_types_builder/src/codegen/rust/api.rs
```diff
@@ -1097,7 +1097,7 @@ fn quote_trait_impls_for_archetype(obj: &Object) -> TokenStream {
         #recommended_descriptors
         ComponentDescriptor {
             archetype_name: Some(#archetype_name.into()),
-            component_name: #indicator_name.into(),
+            component_name: #quoted_indicator_name::DEFAULT.descriptor().component_name,
```
teh-cmc, line 1100: I would just go for this imo:
```suggestion
let indicator_fqname = format!("rerun.components.{}Indicator", obj.name);

teh-cmc (approved):

crates/build/re_types_builder/src/codegen/rust/api.rs
teh-cmc: (that comment isn't in the right place, but you get the idea)

Wumpf: that's unfortunately incorrect because the namespace may have blueprint in it and I couldn't be bothered putting that logic in here

Wumpf: eh okay I do it anyways

Issue #6830, opened 2024-07-09
Split `TensorDimension` into shape and names


emilk:
Replace `TensorDimension` with two components
* one for shape (u64)
* one for dimension names (strings)

Part of simplifying our Arrow schemas and codegen

PR #8376, opened 2024-12-09
Tensor shape and dimension names as separate arrow fields

emilk:
### Related

* Closes https://github.com/rerun-io/rerun/issues/6830

### What

⚠️  Breaks tensors in .rrd files!

In `TensorData` we used to have `shape: [TensorDimension]` with `struct TensorDimension { size: u64, name: Option<String> }`.

Now `TensorData` has `shape: [u64], names: Option<[String]>` instead.

So basically a AoS -> SoA change

### TODO
* [x] Port Python
* [x] Port C++
* [x] Document in migration guide
* [x] Run `@rerun-bot full-check`


crates/store/re_types/definitions/rerun/datatypes/tensor_data.fbs
```diff
@@ -19,8 +19,15 @@ table TensorData (
...
-  shape: [rerun.datatypes.TensorDimension] (order: 200);
+  /// The shape of the tensor, i.e. the length of each dimension.
+  shape: [uint64] (order: 200);
+
+  /// The names of the dimensions of the tensor (optional).
+  ///
+  /// If set, should be the same length as [datatypes.TensorData.shape].
+  ///
+  /// Example: `["height", "width", "channel", "batch"]`.
+  names: [string] (order: 250, nullable);
```
emilk, line 30: Or maybe we should call this `dim_names` to be explicit?

emilk: Or maybe I regret ever starting this work and will never ever touch Python again in my life

crates/top/rerun/src/lib.rs
```diff
@@ -164,10 +164,10 @@ pub const EXTERNAL_DATA_LOADER_INCOMPATIBLE_EXIT_CODE: i32 = 66;
 pub mod external {
     pub use anyhow;
 
-    pub use re_build_info;
-    pub use re_entity_db;
-    pub use re_entity_db::external::*;
-    pub use re_format;
+    pub use ::re_build_info;
+    pub use ::re_entity_db;
+    pub use ::re_entity_db::external::*;
```
emilk, line 169: I got an error about `re_entity_db` being ambiguous 🤷

docs/snippets/README.md
```diff
@@ -23,8 +23,8 @@ These tests check that A) all of our SDKs yield the exact same data when used th
 
 ### Usage
 
-To run the comparison tests, check out `./docs/snippets/compare_snippet_output.py --help`.
-`./docs/snippets/compare_snippet_output.py` is a valid invocation that will build all 3 SDKs and run all tests for all of them.
+To run the comparison tests, check out `pixi run -e py docs/snippets/compare_snippet_output.py --help`.
+`pixi run -e py docs/snippets/compare_snippet_output.py` is a valid invocation that will build all 3 SDKs and run all tests for all of them.
```
emilk, line 27: I'm not a 100% about this, but it seems to work better using the new instructions

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

crates/store/re_types/definitions/rerun/datatypes/tensor_data.fbs
grtlr: I think I prefer just `names`, I can't think of any other thing that is related to names in tensors, but I might be missing something?

crates/viewer/re_view_tensor/src/lib.rs
```diff
@@ -12,3 +12,34 @@ mod view_class;
...
+        tensor_data
+            .shape
+            .iter()
+            .enumerate()
+            .map(|(dim_idx, dim_len)| Self {
+                size: *dim_len,
+                name: tensor_data.dim_name(dim_idx).cloned(),
+            })
+            .collect()
+    }
```
teh-cmc, line 33: So this one clamps from the dimensions's pov :thinking:

docs/snippets/README.md
teh-cmc: lgtm

teh-cmc (approved):
Nice :+1:

My only gripe is that there doesn't seem to exist any centralized place that communicates clearly what happens when the number of names doesn't match the number of dimensions.
Actually, I'm not even sure the behavior itself is consistent all over the place, since it also feels pretty decentralized (which might very well be unavoidable).
But maybe I'm wrong and I'm missing the logic, it's a large diff and I suck at reviews.

PR #8378, opened 2024-12-10
Fix `compare_snippet_output.py` on Mac

emilk:
Now it works. Before it didn't.


abey79 (approved):

pixi.toml
```diff
@@ -512,7 +512,7 @@ sysroot_linux-64 = ">=2.17,<3" # rustc 1.64+ requires glibc 2.17+, see https://b
 [feature.cpp.target.linux-aarch64.dependencies]
 sysroot_linux-aarch64 = ">=2.17,<3" # rustc 1.64+ requires glibc 2.17+, see https://blog.rust-lang.org/2022/08/01/Increasing-glibc-kernel-requirements.html
 
-[feature.cpp.target.unix.dependencies]
+[feature.cpp.dependencies]
```
Wumpf, line 515: that doesn't really make sense, pixi's "unix" also applies to mac

PR #8379, opened 2024-12-10
Improve Bezier edge-drawing

grtlr:
### What

This fixes a bug where the Bezier-edges would flip and start to intersect instead of fanning out properly. It also removes the hardcoded `fan_amount` and makes it dependent on the distance of the nodes.


abey79 (approved):
lgtm!

meta note: we generally try to have screenshots in the PR description when it makes sense. Here, a before/after would have been nice ;)

PR #8381, opened 2024-12-10
Untangle `re_protos` dependencies

jprochazk:
### What

* `re_protos` no longer depends on any `re_*` crates
* Protobuf conversions have been moved out of `re_protos` and into their respective crates (`re_log_types`, `re_chunk_store`, `re_tuid`)


crates/store/re_protos/src/lib.rs
```diff
@@ -6,434 +6,75 @@
...
+/// Generated types for the remote store gRPC service API v0.
+pub mod remote_store {
+    pub mod v0 {
+        pub use crate::v0::rerun_remote_store_v0::*;
     }
 }
 
-#[cfg(test)]
-mod tests {
+// // ==== below are all necessary transforms from internal rerun types to protobuf types =====
```
zehiko, line 44: nit: stale comment

crates/store/re_log_encoding/src/codec/wire.rs
```diff
@@ -0,0 +1,261 @@
...
+                TransportMessageV0::RecordBatch(chunk) => Ok(Some(chunk)),
+                TransportMessageV0::NoData => Ok(None),
+            }
+        }
+    }
+}
+
+/// Helper function that serializes given arrow schema and record batch into bytes
+/// using Arrow IPC format.
+pub fn write_arrow_to_bytes<W: std::io::Write>(
```
zehiko, line 126: I guess this will be somehow re-used in a follow up PR?

zehiko (approved):
LGTM!

(I do see some downsides in this "dependency inversion" - 1/ now lots of our crates depend on ``re_protos`` and 2/ type conversion that's necessary to send things on the wire is now spread across these different crates. This was I guess unavoidable with our goal of centralizing all encoding in ``re_log_encoding``)

crates/store/re_log_encoding/src/codec/wire.rs
jprochazk: Yes, but I can remove it from this one

jprochazk:
merging this now as i'm blocked by it, ultimately we need some kind of dependency inversion, `re_protos` should've always been a "pure" dependency like our other "type definition" crates.

PR #8382, opened 2024-12-10
`codegen/docs/mod.rs` -> `codegen/docs/website.rs`

teh-cmc:
Pure refactor: migrate website docs to their own module, so that we can add new code generators later.


Wumpf (approved):
acknowledging that code has been shoveled from big file to new file

Issue #5662, opened 2024-03-25
Automatically generate list of used archetype in each example


emilk:
We've started manually listing the archetypes used in an example in the README.md fronttmatter, but this is error prone and difficult to maintain. We should automate this instead.

Issue #1123, opened 2023-02-07
Examples: feature matrix table to know what's where


teh-cmc:
Akin to e.g. [wgpu's feature matrix](https://github.com/gfx-rs/wgpu/tree/master/wgpu/examples#feature-matrix)

Wumpf:
we have an [example doc](https://www.rerun.io/docs/getting-started/examples) page by now but a feature matrix would still be nice

emilk:
Related:
* https://github.com/rerun-io/rerun/issues/2270

abey79:
I think the proper way to achieve this would be to have either a dedicated front matter property in the examples, or some custom `feat:point3d`-style tags. What do you think @jprochazk?

PR #8383, opened 2024-12-10
Snippet index: implement codegen foundations

teh-cmc:
All the codegen to generate the autogenerated parts of the new snippet index.

[Rendered](https://github.com/rerun-io/rerun/blob/cmc/snippets_index_1_autogen/docs/snippets/INDEX.md)

* Fixes https://github.com/rerun-io/rerun/issues/1123
* Fixes https://github.com/rerun-io/rerun/issues/5662
* DNM: requires https://github.com/rerun-io/rerun/pull/8382

Generated data can be re-used for:
* https://github.com/rerun-io/rerun/issues/2934


crates/build/re_types_builder/src/codegen/docs/snippets_ref.rs
```diff
@@ -0,0 +1,525 @@
...
+            archetypes_blueprint,
+            components_blueprint,
+            views,
+        }
+    }
+}
+
+// ---
+
+/// Returns `true` if the given name has not been released yet.
```
teh-cmc, line 478: While I find this approach _Good Enough_ :tm:, I'm all ears if someone has something nicer.

zehiko (approved):
makes sense and looks good to me

PR #8387, opened 2024-12-10
Don't show transform arrows on all entities without any other visualizer

Wumpf:
Today, we automatically enable the transform arrow visualization if it is possible and EITHER:
1) explicitly done via blueprint
2) someone logged `axis_length` on their transform
3) there's a pinhole camera at the same path
4) there's no other visualizer that would be active on that path

This PR simple removes case (4). Which leads to a lot less arrows in some scenes. Very prevalent e.g. with what Revy does.

Before (0.20 without setting arrows to 0 length workaround):
image

After:
image


abey79 (approved):
code lgtm, but I'll have to trust you on the consequences :)

I'd say I rather have false negatives than false positives for these arrow, given the user can force-opt-in, so this is going in the right direction and some overshoot is defo acceptable.

Worthy of a small note in the 0.21 migration guide? ("In circumstances XXX you'll have to manually enable transform arrows by doing YYY, if you still want them.")

Wumpf:
good point! Let's add it to the migration guide

PR #8393, opened 2024-12-10
Store: only run the debug assertion against recordings (not blueprints)

### LINKED ISSUE #8389: Debug-panic when cloning space views
Open the following .rbl file with rerun: [bug.rbl.zip](https://github.com/user-attachments/files/18080529/bug.rbl.zip)

> thread 'main' panicked at 'Insert Chunk with multiple values for component named `rerun.blueprint.components.QueryExpression`: this is currently UB'
re_chunk_store/src/writes.rs:42

The .rbl was created like this:
Image

The panic comes when you restart Rerun (implicitly re-loading the rbl, or explicitly)

teh-cmc:
I'd like to keep that assertion a bit longer, just in case. So for now just ignore blueprints, as they are problematic in that case.

* Fixes #8389

### DIFF
--- a/crates/store/re_chunk_store/src/writes.rs
+++ b/crates/store/re_chunk_store/src/writes.rs
@@ -5,6 +5,7 @@ use arrow2::array::{Array as _, ListArray as Arrow2ListArray};
 use itertools::Itertools as _;
 
 use re_chunk::{Chunk, EntityPath, RowId};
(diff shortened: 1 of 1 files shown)


teh-cmc:
In the same vein, the multi-chunk write warning should really be a debug instead.

grtlr (approved):
LGTM!

PR #8401, opened 2024-12-10
Rust: more `impl<AsComponents>` helpers

### LINKED ISSUE #8398: implement `AsComponents` for collections of `AsComponents`
We already do so for collections of `ComponentBatches` but, [as we've seen in the real world](https://github.com/rerun-io/revy/blob/472555d597c8e7a6c520b540533b6e39701ad97a/src/rerun_logger.rs#L234-L246), this can also come in handy at the layer above.

teh-cmc:
New release, new helpers, as is now customary.

These only introduced two breaking changes in all of our examples and snippets, and they look pretty innocuous.

Opinions?

* Fixes #8398

### DIFF
--- a/docs/snippets/all/archetypes/image_send_columns.rs
+++ b/docs/snippets/all/archetypes/image_send_columns.rs
@@ -23,7 +23,13 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Log the ImageFormat and indicator once, as static.
     let format = rerun::components::ImageFormat::rgb8([width as _, height as _]);
-    rec.log_static("images", &[&format as _, &rerun::Image::indicator() as _])?;
+    rec.log_static(
+        "images",
+        &[
+            &format as &dyn rerun::ComponentBatch,
+            &rerun::Image::indicator(),
+        ],
+    )?;
 
     // Split up the image data into several components referencing the underlying data.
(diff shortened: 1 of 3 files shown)


docs/snippets/all/archetypes/image_send_columns.rs
```diff
@@ -23,7 +23,13 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     // Log the ImageFormat and indicator once, as static.
     let format = rerun::components::ImageFormat::rgb8([width as _, height as _]);
-    rec.log_static("images", &[&format as _, &rerun::Image::indicator() as _])?;
+    rec.log_static(
+        "images",
+        &[
+            &format as &dyn rerun::ComponentBatch,
+            &rerun::Image::indicator(),
+        ],
+    )?;
```
teh-cmc, line 32: Inference only breaks due to the explicit indicator shenanigans here, which I think is fine since they're going away.
Without the indicator, this would now come down to this:
```rust
rec.log_static("images", &format)?;
```
which is pretty neat.

teh-cmc:
@rerun-bot full-check

Wumpf (approved):
👍 
can't have enough of those, hum :/

teh-cmc:
Alright, let's see how this goes :see_no_evil:

PR #8409, opened 2024-12-11
Enforce blueprint data is always untagged during the interim

teh-cmc:
We're in a transition period during which the Rerun ecosystem is slowly moving over to tagged data.

During that time, it is common to end up in situations where the blueprint intermixes both tagged
and untagged components, which invariably leads to undefined behavior.
To prevent that, we just always hot-patch it to untagged, for now.

Examples:
* An SDK logs a blueprint (tagged), which is then updated by the viewer (which uses untagged log calls).
* Somebody loads an old .rbl from somewhere and starts logging new blueprint data to it.
* Etc.

This fixes all known blueprint bugs from the last couple days.

In the future, we will want to make sure that the viewer logs blueprint data with appropriate tags, which is part of the `log(MyArchetype.my_component())` story.
The sooner the better, so that the ecosystem can start to converge on always-tagged blueprint files.

### DIFF
--- a/crates/store/re_chunk/src/chunk.rs
+++ b/crates/store/re_chunk/src/chunk.rs
@@ -414,6 +414,32 @@ impl Chunk {
         self
     }
 
+    /// Clones the chunk into a new chunk where all descriptors are untagged.
+    ///
+    /// Only useful as a migration tool while the Rerun ecosystem slowly moves over
+    /// to always using tags for everything.
+    #[doc(hidden)]
+    #[inline]
+    pub fn clone_as_untagged(&self) -> Self {
+        let mut chunk = self.clone();
+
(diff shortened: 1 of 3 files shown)


teh-cmc:
Hmmm, the issue with this is that, while it fixes all the problems, it doesn't help the ecosystem move forward (i.e. you want `Menu > Save blueprint` to be tagged).

Maybe a query-time solution is more appropriate here. Lemme investigate that.

teh-cmc:
Actually, let's get this in no matter what. If I can find and land an alternative before 0.21 gets out, that's a nice bonus; but either way, this fixes our problems.

Wumpf (approved):
Sad but makes sense!

PR #8418, opened 2024-12-11
Enrich ReDap Catalog with RecordingUri on the fly

zehiko:
### What

As of right now, only the viewer is fully aware of Data Platform's endpoint, so we leverage that to enrich Catalog view with ``RecordingUri`` for each recording. 

### Testing done

```
pixi run rerun rerun://0.0.0.0:51234/catalog
```

Screenshot 2024-12-11 at 16 25 34

### DIFF
--- a/crates/store/re_grpc_client/src/lib.rs
+++ b/crates/store/re_grpc_client/src/lib.rs
@@ -3,13 +3,19 @@
 mod address;
 
 pub use address::{InvalidRedapAddress, RedapAddress};
+use re_chunk::external::arrow2;
+use re_log_types::external::re_types_core::ComponentDescriptor;
+use re_types::components::RecordingUri;
+use re_types::Component;
 use url::Url;
 
 // ----------------------------------------------------------------------------
 
 use std::error::Error;
 
-use re_chunk::Chunk;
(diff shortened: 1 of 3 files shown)


crates/store/re_grpc_client/src/lib.rs
```diff
@@ -273,7 +279,7 @@ async fn stream_catalog_async(
     let store_id = StoreId::from_string(StoreKind::Recording, "catalog".to_owned());
 
     let store_info = StoreInfo {
-        application_id: ApplicationId::from("rerun_data_platform"),
+        application_id: ApplicationId::from("redap_catalog"),
```
jleibs, line 282: I came here to request that you make this same change -- you must have hit the same issue as me.

jleibs (approved):
As discussed, let's synthesize the ChunkId on this end -- otherwise looks good.

crates/store/re_grpc_client/src/lib.rs
zehiko: yeah, same app id -> same blueprint :) Hopefully we soon get that sorted with ``GetStoreInfo``.

PR #8424, opened 2024-12-11
Do query for default components only once per view

Wumpf:
Moves query of default components to the "execute query" stage which we perform for every view at the start of the frame.
Previously, we've done this as part of blueprint resolve on every 

Gives me about 2ms cpu frame time in the 2h airtraffic demo on my windows machine running with single thread (since it's easier to profile, `pixi run rerun-release --threads 1`) with the 3d view maximized and the timline minimized.

Change is expected to have high impact for:
* many entities
* no or few transforms (because the transform cost shadows this win too much and adds a lot of noise)

profile snapshots for said scenario:
before
image

after:
image

Related future work:
* need short circuiting for overrides as well - we can just know when there's none (and that's naturally very common!!)
* short circuit when there's no (fitting) default components?
* reduce the amount of query objects being passed around, we have too many of those at this point!

### DIFF
--- a/crates/store/re_query/src/latest_at.rs
+++ b/crates/store/re_query/src/latest_at.rs
@@ -208,7 +208,7 @@ impl QueryCache {
 ///
 /// Use [`LatestAtResults::get`] and/or [`LatestAtResults::get_required`] in order to access
 /// the results for each individual component.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct LatestAtResults {
     /// The associated [`EntityPath`].
     pub entity_path: EntityPath,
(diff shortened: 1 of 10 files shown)


teh-cmc (approved):

crates/store/re_query/src/latest_at.rs
```diff
@@ -208,7 +208,7 @@ impl QueryCache {
 ///
 /// Use [`LatestAtResults::get`] and/or [`LatestAtResults::get_required`] in order to access
 /// the results for each individual component.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
Wumpf, line 211: ended up not needing this actually

Issue #8423, opened 2024-12-11
JPEG images no longer show in 2D views


jleibs:
Broken on `main` / `0.21.0-alpha.1`

For example:
Can also be seen in, e.g. ARKit scene:
https://rerun.io/viewer/pr/8413?url=https%3A%2F%2Fapp.rerun.io%2Fcommit%2F07b8825%2Fexamples%2Farkit_scenes.rrd

Repro with any seen creating encoded 2D images:
```
pixi run -e py python docs/snippets/all/archetypes/encoded_image.py
```

Two bugs:
- 2D View doesn't get created
- If 2D View is created, iamge doesn't show up

Non-encoded images (i.e. `docs/snippets/all/archetypes/image_simple.py`) seem to work fine.

PR #8425, opened 2024-12-11
Fix applicability filters for bar chart & images

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/8423

### What

bar chart & image applicability filters were the only callers of `diff_component_filter` which broke. Couldn't find any related breakages on a quick search

### DIFF
--- a/crates/store/re_chunk/src/chunk.rs
+++ b/crates/store/re_chunk/src/chunk.rs
@@ -73,6 +73,18 @@ impl ChunkComponents {
             .insert(component_desc, list_array);
     }
 
+    /// Returns the first list array for the given component name.
+    ///
+    /// It's undefined which one is returned if there are more than one component with this name.
+    #[inline]
+    pub fn get_by_component_name(
+        &self,
+        component_name: &ComponentName,
+    ) -> Option<&Arrow2ListArray<i32>> {
+        self.get(component_name)
+            .and_then(|per_desc| per_desc.values().next())
+    }
+
     #[inline]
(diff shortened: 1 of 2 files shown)


crates/store/re_chunk/src/chunk.rs
```diff
@@ -73,6 +73,18 @@ impl ChunkComponents {
             .insert(component_desc, list_array);
     }
 
+    /// Returns the all list arrays for the given component name.
+    #[inline]
+    pub fn get_by_component_name<'a>(
```
teh-cmc, line 78: Casually introducing wildcard semantics makes me a bit uneasy :grimacing: 

But whatever, we'll have to redesign everything when introducing descriptor-driven queries anyway.

crates/store/re_chunk/src/chunk.rs
```diff
@@ -73,6 +73,18 @@ impl ChunkComponents {
             .insert(component_desc, list_array);
     }
 
+    /// Returns the all list arrays for the given component name.
```
teh-cmc, line 76:
```suggestion
    /// Returns all list arrays for the given component name.
    /// 
    /// I.e semantically equivalent to `get("MyComponent:*.*")`

teh-cmc (approved):

crates/store/re_chunk/src/chunk.rs
Wumpf: I first went with returning first and saying is UDB for more, but I found this more elegant 🤷

Issue #8154, opened 2024-11-14
Notebook cell times out on firefox


jleibs:
This appears to be a regression between 0.19.1 and 0.20.0

jleibs:
From: https://developer.mozilla.org/en-US/docs/Web/URI/Schemes/data
Firefox version 97 and newer supports data URLs of up to 32MB (before 97 the limit was close to 256MB)

Incidentally: 0.19.1 had a size of 24MB, and with the 33% overhead of base64 encoding came in at 31.9MB

0.20.0 ballooned to 27MB (36MB encoded).

PR #8426, opened 2024-12-11
Fix broken notebook loading on firefox by compressing the encoded wasm payload

jleibs:
### Related
- Resolves: https://github.com/rerun-io/rerun/issues/8154

### What
It turns out firefox doesn't let you create a dataurl larger than 32MB.  Although our `.wasm` was under this threshold, the overhead of base64-encoding pushed us over the threshold.

However, as we were encoding the raw, uncompressed .wasm, we still actually have plenty of margin... we just need to jump through more hoops to use it now. Using DecompressionStream like this seems to be generally available across our target browsers so I think we should be good?

### DIFF
--- a/rerun_js/web-viewer/bundle.mjs
+++ b/rerun_js/web-viewer/bundle.mjs
@@ -30,7 +30,7 @@ async function compressed_data_url_to_buffer(dataUrl) {
     let ds = new DecompressionStream("gzip");
     let decompressedStream = blob.stream().pipeThrough(ds);
 
-    return await new Response(decompressedStream).arrayBuffer();
+    return new Response(decompressedStream).arrayBuffer();
 }
 
 const inlined_js = js.replace("export default function", "return function");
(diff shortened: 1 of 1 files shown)


jprochazk:
For posterity, there's some docs about it here: https://developer.mozilla.org/en-US/docs/Web/URI/Schemes/data#length_limitations

rerun_js/web-viewer/bundle.mjs
```diff
@@ -6,25 +6,31 @@
...
-  const response = await fetch(dataUrl);
-  return response.arrayBuffer();
+async function compressed_data_url_to_buffer(dataUrl) {
+    const response = await fetch(dataUrl);
+    const blob = await response.blob();
+
+    let ds = new DecompressionStream("gzip");
+    let decompressedStream = blob.stream().pipeThrough(ds);
+
+    return await new Response(decompressedStream).arrayBuffer();
```
jprochazk, line 33:
```suggestion
    return new Response(decompressedStream).arrayBuffer();
```

the extra `await` after `return` adds a microtick of latency.

(think of `async` as a transform that turns the code after every `await` into a `.then(context => ...)`, you're adding an extra `then` at the end)

PR #8427, opened 2024-12-12
Fix nightly execution

Wumpf:
maybe. I think that's it.


Wumpf:
It worked! https://github.com/rerun-io/rerun/actions/runs/12293394727
Nightly ci can be triggered again (this will go red because of other issues, but before it wouldn't even start)

PR #8430, opened 2024-12-12
Fix links to blueprint components/archetype that don't show up in our reference yet

Wumpf:
made the link checker (correctly) trip!

### DIFF
--- a/crates/build/re_types_builder/src/docs.rs
+++ b/crates/build/re_types_builder/src/docs.rs
@@ -417,6 +417,13 @@ mod doclink_translation {
                 }
             }
             Target::WebDocsMarkdown => {
+                let kind_and_type = format!("{kind}.{type_name}");
+
+                // TODO(andreas): We don't show blueprint components & archetypes in the web docs yet.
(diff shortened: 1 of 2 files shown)


crates/build/re_types_builder/src/docs.rs
```diff
@@ -417,6 +417,13 @@ mod doclink_translation {
                 }
             }
             Target::WebDocsMarkdown => {
+                let kind_and_type = format!("{kind}.{type_name}");
+
+                // TODO(andreas): We don't show blueprint components & archetypes in the web docs yet.
+                if scope == "blueprint" && (kind == "components" || kind == "archetypes") {
+                    return Ok(kind_and_type);
+                }
```
teh-cmc, line 425: I've been meaning to bring that up a bunch of times, and always forget... Why are we still not showing those? :thinking:

Wumpf: as per short conversation on zoom: those types don't really behave like the others as they don't have any effect when logging them to the store. We'd need to handle them very separately

PR #8432, opened 2024-12-12
Improved reflection

teh-cmc:
Misc improvements to the reflection that I had to implement while investigating forward-compat, migration scripts, etc.

### DIFF
--- a/crates/store/re_types_core/src/reflection.rs
+++ b/crates/store/re_types_core/src/reflection.rs
@@ -238,6 +238,11 @@ pub struct ArchetypeReflection {
     /// e.g. `Spatial3DView`.
     pub view_types: &'static [&'static str],
 
+    /// Does this have a particular scope?
+    ///
+    /// e.g. `"blueprint"`.
+    pub scope: Option<&'static str>,
+
     /// All the component fields of the archetype, in the order they appear in the archetype.
     pub fields: Vec<ArchetypeFieldReflection>,
 }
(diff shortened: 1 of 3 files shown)


crates/store/re_types_core/src/reflection.rs
```diff
@@ -238,6 +238,11 @@ pub struct ArchetypeReflection {
     /// e.g. `Spatial3DView`.
     pub view_types: &'static [&'static str],
 
+    /// Does this have a particular scope?
+    ///
+    /// e.g. `"blueprint"`.
+    pub scope: Option<&'static str>,
```
teh-cmc, line 244: Really useful to filter stuff at runtime, without having to rely on shady `str::contains` shenanigans.

crates/store/re_types_core/src/reflection.rs
```diff
@@ -253,12 +258,15 @@ impl ArchetypeReflection {
 /// Additional information about an archetype's field.
 #[derive(Clone, Debug)]
 pub struct ArchetypeFieldReflection {
-    /// The type of the field (it's always a component).
-    pub component_name: ComponentName,
+    /// The name of the field (i.e. same as `ComponentDescriptor::archetype_field_name`).
+    pub name: &'static str,
```
teh-cmc, line 262: Mandatory in order to be able to recreate complete `ComponentDescriptors` at runtime.

Wumpf (approved):

PR #8433, opened 2024-12-12
Factor out `NearClipPlane` from `VisualBounds2D`

### LINKED ISSUE #8415: Separate `NearClipRange` from `VisualBounds2d`
Recent changes have added a new `NearClipRange` component to the `VisualBounds2d` archetype.

This shows up even in the graph view's UI, where near clipping does not make any sense (it's based on `egui`).

Image

We should probably move this out to a different archetype (something like `Camera` maybe?)

grtlr:
### Related

* Closes #8415 

### What

Title.

It looks like we never had a fallback provider for `NearClipPlane`. @Wumpf can you please confirm that this is correct? 🙏


crates/store/re_types/definitions/rerun/blueprint/archetypes/near_clip_plane.fbs
```diff
@@ -0,0 +1,12 @@
+namespace rerun.blueprint.archetypes;
+
+/// Controls the distance to the near clip plane in 3D scene units.
+table NearClipPlane (
```
Wumpf, line 4: this needs the `attr.docs.unreleased` attribute otherwise links are broken

Wumpf:
`check_overrides_2d.py` needs to be updated

PR #8438, opened 2024-12-12
Improve `VisualBounds2D` behavior in graph view

### LINKED ISSUE #8429: Fit the graph bounding box to the view unless user overrides
This would prevent graphs to be only partially shown during layout.

grtlr:
### Related

* Closes #8429 

### What

This implements improved handling of `VisualBounds2D`. Specifically, it:

* Always defaults to the data bounds when user has not modified the view (@nikolausWest's suggestion).
* Zooms in and out of view when resizing the view (@abey79's suggestion).
* Fixes the `VisualBounds2D` selection panel bug (submitted by @Wumpf)

### DIFF
--- a/crates/viewer/re_view_graph/src/view.rs
+++ b/crates/viewer/re_view_graph/src/view.rs
@@ -187,7 +187,7 @@ Display a graph of nodes and edges.
 
         let mut ui_from_world = fit_to_rect_in_scene(rect_in_ui, rect_in_scene.into());
 
-        // We store a copy o the transformation to see if it has changed.
+        // We store a copy of the transformation to see if it has changed.
         let ui_from_world_ref = ui_from_world;
 
         let resp = zoom_pan_area(ui, rect_in_ui, &mut ui_from_world, |ui| {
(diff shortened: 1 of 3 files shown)


crates/viewer/re_view_graph/src/view.rs
```diff
@@ -183,22 +183,14 @@ Display a graph of nodes and edges.
...
+        let mut ui_from_world = fit_to_rect_in_scene(rect_in_ui, rect_in_scene.into());
 
-        // We ensure that the view's center is kept during resizing.
-        if let Some(prev) = prev_rect_in_ui {
-            if prev != rect_in_ui {
-                let delta = rect_in_ui.center() - prev.center();
-                ui_from_world.translation += delta;
-            }
-        }
+        // We store a copy o the transformation to see if it has changed.
```
Wumpf, line 190:
```suggestion
        // We store a copy of the transformation to see if it has changed.
```

Wumpf (commented):
looking good. Waiting for web demo to build so I Can try it out :)

Wumpf (approved):
there seems to be some kind of minimum zoom? (didn't dig into the code enough to spot it)
If I make change visual bounds to be very small it doesn't zoom in further. Surely pre-existing behavior that was added intentionally, but I find it a bit odd

grtlr:
Yes, that's a limitation of `egui` you can't zoom in further than `1.0`. Because otherwise, text and everything would become blurry.

teh-cmc:
Is that :green_circle: I'm seeing? :detective: yes it is!

PR #8439, opened 2024-12-12
Migration kernel for the blueprint space-view-related breaking changes

teh-cmc:
Our first migration kernel! This makes it possible to load <0.21 blueprints into Rerun 0.21, transparently, without user intervention.

More importantly, this marks the first step in exploring these kinds of tools, as part of our general promise of becoming more and more stable over time.

Basically a gloried search and replace.

To test it, save any blueprint from 0.20 (https://app.rerun.io/version/0.20.3/index.html), and load it into 0.21.

### DIFF
--- a/crates/store/re_chunk_store/src/writes.rs
+++ b/crates/store/re_chunk_store/src/writes.rs
@@ -62,7 +62,9 @@ impl ChunkStore {
         // * Somebody loads an old .rbl from somewhere and starts logging new blueprint data to it.
         // * Etc.
         if self.id.kind == re_log_types::StoreKind::Blueprint {
-            chunk = Arc::new(chunk.clone_as_untagged());
+            let patched = chunk.patched_for_blueprint_021_compat();
+            let patched = patched.clone_as_untagged();
+            chunk = Arc::new(patched);
         }
 
         #[cfg(debug_assertions)]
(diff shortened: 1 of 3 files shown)


teh-cmc:
@rerun-bot full-check

crates/store/re_chunk_store/src/writes.rs
```diff
@@ -62,7 +62,9 @@ impl ChunkStore {
         // * Somebody loads an old .rbl from somewhere and starts logging new blueprint data to it.
         // * Etc.
         if self.id.kind == re_log_types::StoreKind::Blueprint {
-            chunk = Arc::new(chunk.clone_as_untagged());
+            let patched = chunk.patched_for_blueprint_021_compat();
+            let patched = patched.clone_as_untagged();
+            chunk = Arc::new(patched);
         }
```
Wumpf, line 68: I wish we had a version tag on the store so we're not applying this indefinitely into the future ...

Wumpf (approved):
👍

crates/store/re_chunk_store/src/writes.rs
teh-cmc: Yeah we need to pull the StoreInfos all the way in at some point. We can do that later though, that's fine (I don't expect this particular change to be particularly version specific -- `SpaceView` is not making a come back anytime soon I hope :stuck_out_tongue:).

PR #8441, opened 2024-12-12
Fix out of bounds access in data density graph

Wumpf:
Regressed this somewhat recently during perf optimizations in the data_density_graph: it was possible to get into an out of bounds access when zooming in.
Cleaned it up a bit overall and fixed a subtle bug for when there's only fractional buckets filled in the density graph


abey79 (approved):
Not on top of all the minute logic there, but I trust you and code lgtm :)

teh-cmc:
I see :green_circle:, I ship

PR #8449, opened 2024-12-13
craft blueprint for Catalog view from raw chunks

zehiko:
Note: for now we accept that have to do a lot of things manually until we have proper rust blueprint API or we stream this blueprint recording from ReDap
* #8470


crates/store/re_grpc_client/src/lib.rs
```diff
@@ -376,3 +387,104 @@ async fn stream_catalog_async(
...
 }
+
+// Craft a blueprint from relevant chunks and activate it
+fn activate_catalog_blueprint(
+    tx: &re_smart_channel::Sender<LogMsg>,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let blueprint_store_id =
+        StoreId::from_string(StoreKind::Blueprint, "catalog_blueprint".to_owned());
+    let blueprint_store_info = StoreInfo {
+        application_id: ApplicationId::from("redap_catalog"),
```
abey79, line 398: This prob needs to be the same as the recording store, best make sure it's based on some constant or something.

crates/store/re_grpc_client/src/lib.rs
```diff
@@ -376,3 +387,104 @@ async fn stream_catalog_async(
...
+        store_id: blueprint_store_id.clone(),
+        cloned_from: None,
+        is_official_example: false,
```
abey79, line 401: Side note: IIRC this is related to how we deal with analytics. We probably should do something special for these meta-recordings at some point as well.

crates/store/re_grpc_client/src/lib.rs
```diff
@@ -376,3 +387,104 @@ async fn stream_catalog_async(
...
+            row_id: *re_chunk::RowId::new(),
+            info: blueprint_store_info,
+        }))
+        .is_err()
+    {
+        re_log::debug!("Receiver disconnected");
+        return Ok(());
+    }
+
+    let timepoint = [(Timeline::new_sequence("blueprint"), 1)];
```
abey79, line 418: `re_viewer_context::blueprint_timeline()`

crates/store/re_grpc_client/src/lib.rs
```diff
@@ -376,3 +387,104 @@ async fn stream_catalog_async(
...
+
+    let vb = ViewBlueprint::new("Dataframe")
+        .with_visible(true)
+        .with_space_origin("/");
+
+    let view_uuid = uuid::Uuid::new_v4();
+    let view_id = format!("/view/{view_uuid}");
```
abey79, line 425: These entities are typically are built using `SpaceViewId::registry_path()`, where SpaceViewId is a wrapper over these uuids.

abey79: also, `view_id` is misleading. should be `view_entity_path`

crates/store/re_grpc_client/src/lib.rs
```diff
@@ -376,3 +387,104 @@ async fn stream_catalog_async(
...
+    let view_chunk = ChunkBuilder::new(ChunkId::new(), view_id.clone().into())
+        .with_archetype(RowId::new(), timepoint, &vb)
+        .build()?;
+
+    let epf = EntityPathFilter::parse_forgiving("/**", &Default::default());
+    let vc = ViewContents::new(epf.iter_expressions());
+    let view_contents_chunk = ChunkBuilder::new(
+        ChunkId::new(),
+        format!("{}/ViewContents", view_id.clone()).into(),
```
abey79, line 434: use `...::components::ViewContents::name().short_name()`

crates/store/re_grpc_client/src/lib.rs
```diff
@@ -376,3 +387,104 @@ async fn stream_catalog_async(
...
+
+    let rc = ContainerBlueprint::new(ContainerKind::Grid);
+    let rc = rc
+        .with_contents(&[EntityPath::from(view_id)])
+        .with_visible(true);
+
+    let container_uuid = uuid::Uuid::new_v4();
+    let container_chunk = ChunkBuilder::new(
+        ChunkId::new(),
+        format!("/container/{container_uuid}").into(),
```
abey79, line 447: Likewise, these are derived from `ContainerId::registry_path()`

crates/store/re_grpc_client/src/lib.rs
```diff
@@ -376,3 +387,104 @@ async fn stream_catalog_async(
...
+    )
+    .with_archetype(RowId::new(), timepoint, &rc)
+    .build()?;
+
+    let vp = ViewportBlueprint::new().with_root_container(RootContainer(container_uuid.into()));
+    let viewport_chunk = ChunkBuilder::new(ChunkId::new(), "/viewport".into())
```
abey79, line 453: `re_viewport_blueprint::VIEWPORT_PATH`

abey79 (commented):
This is not as bad as I feared, but obviously it contains heaps of hard-coded stuff that might broke in the future. I'll drop in comments the places where these can be found in our code base. Refactoring it all might be tricky though, as most of it is in `re_viewport_blueprint`. Can you depend on it from `re_grpc_client`? If not, we could maybe move some of it to `re_types` or `re_log_types`? TBD with @Wumpf as a topic.

crates/store/re_grpc_client/src/lib.rs
zehiko: I'm actually not sure it's worth bringing in ``re_viewer_context`` to grpc client just for this one.

zehiko: ah, right, I see suggestions below are also from this crate...

zehiko: and adding it would create a dependency cycle :/

abey79: Yeah I was expecting that dependency cycle. I dont think we should do anything about it. My suggestion is to simply add a bunch of todos, and maybe open an issue to document this.

crates/store/re_grpc_client/src/lib.rs
zehiko: done

crates/store/re_grpc_client/src/lib.rs
zehiko: Had a brief huddle with @abey79 and he explained a bit the dependencies and what I'm hitting here makes sense and we don't want grpc client to depend on ``re_viewer_context``. When rust blueprint API is properly introduced, we might depend on that crate (something like ``re_blueprint``), but arguably by then catalog view blueprint might simply move to Data Platform and live along other customer recording's blueprints. 

For now we'll go with what we have here and I'll add some TODOs in the code.

abey79 (approved):

Issue #8368, opened 2024-12-09
Make visualizers & archetypes map 1:1 (+notes on visualizer selection in world without indicators)


Wumpf:
## Today
- Visualizer identifiers are distinct from archetype names
    - but **most** of the time identical
- Visualizers are selected heuristically based on archetype indicators
- Visualizers can be set explicitly via `VisualizerOverrides`

### Points of friction

- visualizers are a separate and under-documented concept
- `VisualizerOverrides` has informal list of visualizer strings that is not kept up to date and doesn't show up anywhere else
- tagged components go away - how to decide visualizer when names are conflicting?

## Solution: correlation of archetypes & visualizers

Each visualizer can be described by a name & a list of components, i.e. they _are_ archetypes already.

-> Some archetypes don't have a visualizer - e.g. `Scalar`
-> Separate “required for logging” & “required for visualizing” - `Scalar` is required to visualize point/line, so we won't offer it automatically. But logging `SeriesLine` doesn't require `Scalar` and the presence of its tag still brings the visualizer into existence.

Going forward, some archetypes may have to be split up:
There can be a "base" archetype for mesh data and additional components in a `MeshWireframe`
Note that the base archetype is typically the "raw data" and may be standardized in sorbet much earlier, whereas the extending "visualizer archetype" may follow later.

We likely want some kind of codegen supported ‘extends’ construct to better express this, _potentially_ with specially generated code & docs.

## Setting visualizers

`VisualizerOverrides` 's possible values are any archetype name which is visualizable.

Default heuristic for enabling visualizers is governed by the presence of any matching archetype name in any tag on a given entity.
A store subscriber gathers this data → replaces “indicated component concept”

Beyond that there can still be arbitrary heuristics that for instance ensure that a `Scalar` is shown with `SeriesPoint`

## TODOs

* [ ] Remove visualizer identifiers and replace them with archetype names
* [ ] Adjust docs & functionality `VisualizerOverrides` accordingly
* [ ] remove `VisualizerSystem::visualizer_query_info` as it is fully governed by the matching archetype definition
     * [ ] split concept of "requried for visualizing" & "required for logging", add `Scalar` to `SeriesPoint` & `SeriesLine` accordingly
* [ ] (upon removal of indicators:) change indicated entities to look for any archetype name being set
* [ ] Ensure docs reflect all of the above

## Future work

[...]

PR #8458, opened 2024-12-13
Add utility to `rr.components.Color` to generate colors from any string (and use it in the air traffic data example)

abey79:
### What

It's some time nice to log some color information in multiple entities to make it easier to relate them visually. This PR adds a `rr.components.Color.from_str()` utility that does exactly that: generate a nice color randomly picked up based on the provided string.

This PR also updates the air traffic data example so the barometric traces have matching colors with the map data.

### DIFF
--- a/examples/python/air_traffic_data/air_traffic_data.py
+++ b/examples/python/air_traffic_data/air_traffic_data.py
@@ -239,6 +239,7 @@ class MeasurementLogger:
             )
 
         entity_path = f"aircraft/{measurement.icao_id}"
+        color = rr.components.Color.from_string(entity_path)
 
         if (
             measurement.latitude is not None
@@ -247,13 +248,16 @@ class MeasurementLogger:
         ):
             rr.log(
                 entity_path,
-                rr.Points3D([
-                    self._proj.transform(
-                        measurement.longitude,
-                        measurement.latitude,
-                        measurement.barometric_altitude,
-                    )
-                ]),
+                rr.Points3D(
+                    [
+                        self._proj.transform(
+                            measurement.longitude,
+                            measurement.latitude,
+                            measurement.barometric_altitude,
+                        ),
+                    ],
+                    colors=color,
+                ),
                 rr.GeoPoints(lat_lon=[measurement.latitude, measurement.longitude]),
             )
 
@@ -264,6 +268,7 @@ class MeasurementLogger:
             rr.log(
                 entity_path + "/barometric_altitude",
                 rr.Scalar(measurement.barometric_altitude),
+                rr.SeriesLine(color=color),
             )
 
     def flush(self) -> None:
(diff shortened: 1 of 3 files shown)


rerun_py/rerun_sdk/rerun/components/color_ext.py
```diff
@@ -0,0 +1,28 @@
...
+_GOLDEN_RATIO = (math.sqrt(5.0) - 1.0) / 2.0
+
+
+class ColorExt:
+    """Extension for [Color][rerun.components.Color]."""
+
+    @staticmethod
+    def from_string(s: str) -> Color:
+        """
+        Generate a random yet color based on a string.
```
teh-cmc, line 19:
```suggestion
        Generate a random yet deterministic color based on a string.

File: examples/python/air_traffic_data/air_traffic_data.py [hunk 2]
```diff
@@ -310,7 +315,13 @@ def log_position_and_altitude(self, df: polars.DataFrame, icao_id: str) -> None:
             return
 
         if icao_id not in self._position_indicators:
-            rr.log(entity_path, [rr.archetypes.Points3D.indicator(), rr.archetypes.GeoPoints.indicator()], static=True)
+            color = rr.components.Color.from_string(entity_path)
+            rr.log(
+                entity_path,
+                [rr.archetypes.Points3D.indicator(), rr.archetypes.GeoPoints.indicator(), color],
```
teh-cmc, line 321: Good lord I didnt realize how much that code needs tagged components :see_no_evil:

teh-cmc (approved):

examples/python/air_traffic_data/air_traffic_data.py
abey79: Good example of a component which is _intended_ to be shared across two archetypes, btw. Hopefully we don't make that use case too contrived, aka if that `Color` is  tagged with the `Points3D` archetype, it should be implicitly(?) used by the `GeoPoints` archetype as well.

teh-cmc: Ha yeah I didn't even realize that was what's happening, interesting.

Yeah we'll have to figure out how we want to handle those. I do think heuristics to implicitly pick up the same component from unrelated tags if on the same entity makes sense.

abey79: Vaguely related is one realisation we've had when discussing the archetype vs. visualiser relationship the other day: indicator components are additive, while tags are exclusive. With the former, you could have 2 archetypes with 1 components (e.g. hypothetical ScalarSeriesLine + ScalarSeriesPoint archetypes based on a single scalar component + 2 indicators), but with the latter you are forced to make a choice.

We don't think this constraint is going to be much of problem though. More in https://github.com/rerun-io/rerun/issues/8368

Issue #7868, opened 2024-10-22
Deprecate `DisconnectedSpace`, replace with pure transform marker type `InvalidTransform`/`UndefinedTransform` (tbd)


Wumpf:
While sitting over 
* https://github.com/rerun-io/rerun/issues/6817

@jleibs and me realized that given how the rest of the Viewer evolved the concept of a "disconnected" space no longer makes sense: it's not trivial to configure entity path expressions to achieve what we previously tried to achieve with `DisconnectedSpace`

We still want a pure spatial (not topological like `DisconnectedSpace`) property that disonnects the transform tree itself. This would not have any effect on either
* internal topology
* heuristics
* visualizability

I.e. this preserves what we handle within the transform context as `UnreachableTransformReason::DisconnectedSpace`, but removes all other aspects of disconnected space.

---

We should introduce this new archetype and deprecated `DisconnectedSpace` in the same release, making it equivalent to the new type as long as it is shipped under a deprecation mark.

PR #8459, opened 2024-12-13
Deprecate `DisconnectedSpace` archetype/component in favor of implicit invalid transforms

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/7868

### What

See title. In detail:
* `RotationAxisAngle` will no longer be ignored when axis is non-normalizable
* deprecation of `DisconnectedSpace`
   * slight fixes to codegen for handling this smoothly
* lots of doc updates
* wherever we had constants for identity, we now also have one for invalid
    * python didn't have any at all so far, but I didn't want to add this in this iteration since generally constants there are a bit tricky
    * we should have more of those generally


crates/store/re_types/definitions/rerun/archetypes/transform3d.fbs
```diff
@@ -40,6 +40,15 @@ table Transform3D (
   /// Specifies the relation this transform establishes between this entity and its parent.
   relation: rerun.components.TransformRelation ("attr.rerun.component_optional", nullable, order: 1600);
 
+  /// Optionally flags the transform as invalid.
```
jleibs, line 43: Need something like: "If this is set to true, then all other components on the Transform will be ignored"

jleibs (commented):
Now that we have tagged components, does this actually need to be a Transform-specific component?

This could be `rerun.components.InvalidData`.

I'm also now wondering about `Invalid = True` vs `Valid = False`.

`rerun.components.ValidityMask` would feel the most arrow-like from a concept perspective, though I recognize this opens a can of worm of wanting to start supporting `valid = False` for every visualizer.

Wumpf:
> Now that we have tagged components, does this actually need to be a Transform-specific component?
> This could be rerun.components.InvalidData.

That would imply that we make this more of an infrastructure thing and many things can be flagged as `InvalidData`. Which I really don't want to get into. Also, I wouldn't say it's invalid _data_ per se. You might still have some valid transforms active, you just want to disable it.

On that note: we could also dodge this entire thing completely by saying any of the invalid-transform states like full nulled mat3x3 has the same semantics. Today, some of them are explicitely ignored since it's more practical - specifically we speced rotation-axis-angle with a null axis to just be no-rotation since that's more convenient (and I actually already fixed it up here so that the same is true for null quaternions)


> I'm also now wondering about Invalid = True vs Valid = False.

Either is nasty. Maybe best to go with an enum to force making this explicit. In that case We end up with `DataValidity`/`TransformValidity` == false/true


.... but overall I really don't want to make this any more complicated than is. The whole point of doing this now is to make a presumably rarely used feature not be as much in the way of performance enhancement & simplifying (!) refactors. That's why I wanted to deprecate `DisconnectedSpace` asap (something we already agreed upon) so I have simpler logic in the release after

jleibs:
>  You might still have some valid transforms active, you just want to disable it.

This is a great observation, which makes me realize any of the names related to Validity aren't quite correct.

What we really want to say then is simply "Don't use this transform" (which is also more Transform-specific concept than validity)

Other words: "Enable/Disable", "Apply", "Connect/Disconnect"... :thinking:

Wumpf:
Talking this out we came to the conclusion that for the moment we won't introduce a new component and instead encourage invalid transforms like scale==0/matrix==0 etc.

crates/viewer/re_view_spatial/src/contexts/transform_context.rs
```diff
@@ -550,17 +551,36 @@ fn query_and_resolve_tree_transform_at_entity(
...
+    if let Some(quaternion) = result.component_instance::<RotationQuat>(0) {
+        if let Ok(quaternion) = glam::Affine3A::try_from(quaternion) {
+            transform *= quaternion;
+        } else {
+            // Invalid transform.
+            return None;
+        }
     }
     if let Some(scale) = result.component_instance::<Scale3D>(0) {
+        if scale.x() == 0.0 || scale.y() == 0.0 || scale.z() == 0.0 {
```
jleibs, line 571:
```suggestion
        if scale.x() == 0.0 && scale.y() == 0.0 && scale.z() == 0.0 {
```
I believe it should be valid to collapse a single dimension onto a plane via scaling.

jleibs (approved):

crates/viewer/re_view_spatial/src/contexts/transform_context.rs
Wumpf: good point, for some things it is! 👍

PR #8460, opened 2024-12-13
Fix some assorted lints

jleibs:
### What

Not sure what changed to cause me to start hitting these locally in my pre-push hook, but they all seem valid.


Wumpf (approved):
seems all legit indeed!

Issue #4304, opened 2023-11-22
Panel with recent log/notification messages


emilk:
All rerun log messages at INFO, WARN and ERROR levels are shown as popup "toasts" to the user. They linger for a few seconds, and then they are gone. It can be pretty annoying if there any many messages in a row, or if a message is big.

We should have a panel where the user can find these messages instead. @martenbjork has a design where we show a little number with unread messages: 
<img width="148" alt="image" src="https://github.com/rerun-io/rerun/assets/1148717/682ca565-99df-4ddc-ba11-35ef17018bb5">

We should show a scrollback of recent messages, and a red dot next to unread ones.

If we have such a view, then we could change the toast to not show long messages, and never more than one at a time, reducing the amount of noise in the viewer.

### Example
`log::info!("File saved to {path:?}");`

Results in a shortened toast showing: `File saved to C:/Docu…`, and then a white `1` in the title bar, indicating there is one unread INFO-level notification.

teh-cmc:
The next logical step once we have a global notifications panel would be to support categorizing warnings/errors per space view, and then add a badge to each space view that only counts the warnings for that specific space view, and redirects to the global notifications panel with a filter preset on click.

jleibs:
There is a very related problem to this: In many situations we directly send TextLog messages to the `/rerun` path if a user has logged something that threw a caught-and-handled exception.

Unfortunately, since introducing blueprints, we are no longer guaranteed that the TextLog heuristics will cause a `/rerun` text log view to show up.

It would be nice if this notification mechanism also let you review messages logged to `/rerun` even if you don't have an active TextLog space.

jleibs:
This would be a nice feature to have in the Gradio component where a job is running on a remote server and console out isn't immediately visible to the user that kicked off the job in the space.

ShaddyDC:
Something like this would be really helpful. I just tried the app for the first time by loading it in the browser and opening a random gltf file I had lying around. It showed me 2 dozen warnings on the right, and before I was able to read even one of them completely, they were all gone again.
Now, in this case, the browser console still had what I believe to be the problem, and it's fine to ignore, but otherwise I'd be a bit concerned if there was anything wrong and I'd be unable to find out.

For context, it was 38 times this:
```
[re_renderer::importer::gltf] re_renderer/src/importer/gltf.rs:250: Textures on meshes are always sampled with a trilinear filter.
[...]

PR #8465, opened 2024-12-15
Panel with recent notifications

jprochazk:
* Closes https://github.com/rerun-io/rerun/issues/4304

image

image

image


crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+
+            created_at: Zoned::now(),
+            ttl: base_ttl(),
+        });
+    }
+
+    pub fn ui(&mut self, egui_ctx: &egui::Context, is_panel_visible: &mut bool) {
+        if *is_panel_visible {
+            let escape_pressed =
+                egui_ctx.input_mut(|i| i.consume_key(egui::Modifiers::NONE, egui::Key::Escape));
```
abey79, line 121: From cursory testing,  at least on web, the escape key is not "consumed" enough to avoid loosing your current selection. I'm unsure why this is the case though. But I sure dislike this behaviour, so would be great to find a fix 🙏🏻

crates/viewer/re_ui/Cargo.toml
```diff
@@ -50,6 +50,11 @@ smallvec.workspace = true
 strum_macros.workspace = true
 strum.workspace = true
 sublime_fuzzy.workspace = true
+jiff.workspace = true
```
emilk, line 53: Sort depedencies!

Cargo.toml
```diff
@@ -199,6 +199,7 @@ indicatif = "0.17.7" # Progress bar
 infer = "0.16" # infer MIME type by checking the magic number signaturefer MIME type by checking the magic number signature
 insta = "1.23"
 itertools = "0.13"
+jiff = "0.1.15"
```
emilk, line 202: We already use `time` to format timestamps - should we switch to `jiff` everywhere?

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+
+use crate::icons;
+use crate::UiExt;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum NotificationLevel {
+    Info,
+    Warning,
+    Error,
+    Success,
```
emilk, line 15: Please document the difference between `Info` and `Success`

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+    show_notification_panel: &mut bool,
+    has_unread_notifications: bool,
+) {
+    let response = ui.medium_icon_toggle_button(&icons::NOTIFICATION, show_notification_panel);
+
+    if has_unread_notifications {
+        let mut pos = response.rect.right_top();
+        pos.x -= 2.0;
+        pos.y += 2.0;
+        ui.painter().circle_filled(pos, 3.0, hex_color!("#ab0037"));
```
emilk, line 46: I suggest we pick this color based on the severity of the highest unread notification (red, orange, white), though maybe we decided against that?

In either case, this should use a color from `crates/viewer/re_ui/src/color_table.rs`

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+    ui: &mut egui::Ui,
+    show_notification_panel: &mut bool,
+    has_unread_notifications: bool,
+) {
+    let response = ui.medium_icon_toggle_button(&icons::NOTIFICATION, show_notification_panel);
+
+    if has_unread_notifications {
+        let mut pos = response.rect.right_top();
+        pos.x -= 2.0;
+        pos.y += 2.0;
```
emilk, line 45: nit: shorter

```suggestion
        let pos = response.rect.right_top() + egui::vec2(-2.0, 2.0);
```

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+
+    created_at: Zoned,
+    ttl: Duration,
+}
+
+pub struct NotificationUi {
+    /// State of every notification.
+    ///
+    /// Notifications are stored in order of ascending age, so the latest one is at the end.
+    data: Vec<Notification>,
```
emilk, line 62: "ascending age" means they get progressively older, while "latest one is at the end" mean the opposite

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    pub fn success(&mut self, text: impl Into<String>) {
+        self.data.push(Notification {
+            level: NotificationLevel::Success,
+            text: text.into(),
+
+            created_at: Zoned::now(),
+            ttl: base_ttl(),
+        });
```
emilk, line 115: Is it intentional that this does not set `has_unread_notifications`? If yes, please make it explicit. If no, just call `self.add(NotificationLevel::Success, text);` here?

crates/viewer/re_ui/src/notifications.rs
emilk: That should depend only on the order in which selection checks for `Key::Escape` and this code is run, so maybe just need to move some code around

crates/viewer/re_ui/src/notifications.rs
emilk, line 62: Also, `data` is one of those empty names (like "thing" or "stuff"). Better to call it `notifications` or `messages`

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+        }
+
+        self.panel.show(egui_ctx, &mut self.data, is_panel_visible);
+        self.toasts.show(egui_ctx, &mut self.data[..]);
+
+        if let Some(notification) = self.data.last() {
+            if !notification.ttl.is_zero() {
+                egui_ctx.request_repaint_after(notification.ttl);
+            }
+        }
```
emilk, line 136: Why do we do this even if the panel is closed? Is that a bug or intentional?

Maybe `NotificationPanel::show` can handle this logic instead?

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+        ui.painter().circle_filled(pos, 3.0, hex_color!("#ab0037"));
+    }
+}
+
+struct Notification {
+    level: NotificationLevel,
+    text: String,
+
+    created_at: Zoned,
+    ttl: Duration,
```
emilk, line 55: What does this do? Is the notification removed after this time, even when I'm reading it? If not, is it removed when I'm _not_ reading it? Either way seems bad

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+
+        let mut clear = false;
+
+        egui::Area::new(self.id)
+            .anchor(egui::Align2::RIGHT_TOP, egui::vec2(-8.0, 32.0))
+            .order(egui::Order::Foreground)
+            .interactable(true)
+            .movable(false)
+            .show(egui_ctx, |ui| {
+                egui::Frame::window(ui.style())
```
emilk, line 196: You should be able to use `egui::Window`, which encapsulated an `Area`, `Frame`, and `ScrollArea`. Should just be less code

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+        egui_ctx: &egui::Context,
+        notifications: &mut Vec<Notification>,
+        is_panel_visible: &mut bool,
+    ) {
+        if !*is_panel_visible {
+            return;
+        }
+
+        for notification in notifications.iter_mut() {
+            notification.ttl = Duration::ZERO;
```
emilk, line 162: So they are dead while I'm reading them? O.o

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+        let panel_max_height = 320.0;
+
+        let mut to_dismiss = None;
+
+        let notification_list = |ui: &mut egui::Ui| {
+            if notifications.is_empty() {
+                ui.label(
+                    egui::RichText::new("Nothing here!")
+                        .weak()
+                        .color(hex_color!("#636b6f")),
```
emilk, line 175: Should use `crates/viewer/re_ui/src/color_table.rs`

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+
+        let panel_width = 356.0;
+        let panel_max_height = 320.0;
+
+        let mut to_dismiss = None;
+
+        let notification_list = |ui: &mut egui::Ui| {
+            if notifications.is_empty() {
+                ui.label(
+                    egui::RichText::new("Nothing here!")
```
emilk, line 173: I'm not sure about this copy. Maybe "No notifications"?

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+            }
+
+            for (i, notification) in notifications.iter().enumerate().rev() {
+                show_notification(ui, notification, DisplayMode::Panel, || {
+                    to_dismiss = Some(i);
+                });
+            }
+        };
+
+        let mut clear = false;
```
emilk, line 188:
```suggestion
        let mut clear_all = false;
```

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+                                }
+                            });
+                        });
+                        egui::ScrollArea::vertical()
+                            .min_scrolled_height(panel_max_height)
+                            .show(ui, notification_list);
+
+                        if !notifications.is_empty() {
+                            ui.horizontal_top(|ui| {
+                                if ui.button("Clear all").clicked() {
```
emilk, line 216:
```suggestion
                                if ui.button("Dismiss all").clicked() {
```

to match the wording used per-notification

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+
+    /// Shows and updates all toasts
+    fn show(&self, egui_ctx: &egui::Context, notifications: &mut [Notification]) {
+        let dt = Duration::from_secs_f32(egui_ctx.input(|i| i.unstable_dt));
+        let mut offset = egui::vec2(-8.0, 32.0);
+
+        for (i, notification) in notifications
+            .iter_mut()
+            .enumerate()
+            .filter(|(_, n)| n.ttl > Duration::ZERO)
```
emilk, line 261: Oh, ttl is only for the toast part? Then maybe call it `toast_ttl` and add a docstring to it

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+fn show_notification(
+    ui: &mut egui::Ui,
+    notification: &Notification,
+    mode: DisplayMode,
+    mut on_dismiss: impl FnMut(),
+) -> egui::Response {
+    egui::Frame::window(ui.style())
+        .rounding(4.0)
+        .inner_margin(10.0)
+        .fill(hex_color!("#1c2123"))
```
emilk, line 308: `crates/viewer/re_ui/src/color_table.rs`

maybe we should add a lint for this :)

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+            ui.vertical_centered(|ui| {
+                let text_response = ui
+                    .horizontal_top(|ui| {
+                        ui.style_mut().wrap_mode = Some(egui::TextWrapMode::Wrap);
+                        ui.set_max_width(300.0);
+                        ui.spacing_mut().item_spacing.x = 8.0;
+                        log_level_icon(ui, notification.level);
+                        ui.label(
+                            egui::RichText::new(notification.text.clone())
+                                .color(hex_color!("#cad8de"))
```
emilk, line 319: and again

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -0,0 +1,393 @@
...
+    let formatted = if age <= 9.0 {
+        ui.ctx().request_repaint_after(Duration::from_secs(1));
+
+        "just now".to_owned()
+    } else if age <= 59.0 {
+        ui.ctx().request_repaint_after(Duration::from_secs(1));
+
+        format!("{age:.0}s")
+    } else {
+        ui.ctx().request_repaint_after(Duration::from_secs(60));
```
emilk, line 363: Nice

crates/viewer/re_viewer/src/app.rs
```diff
@@ -1119,26 +1119,8 @@ impl App {
     fn show_text_logs_as_notifications(&mut self) {
         re_tracing::profile_function!();
 
-        while let Ok(re_log::LogMsg { level, target, msg }) = self.text_log_rx.try_recv() {
-            let is_rerun_crate = target.starts_with("rerun") || target.starts_with("re_");
-            if !is_rerun_crate {
```
emilk, line 1124: I missed - do we still have this filter somewhere? We don't want to show `log::info!` from third-party-crates

emilk (approved):
I love it! ❤️  FINALLY

Wumpf:
This picks up stuff we didn't show in the viewer at all so far :/

Firefox windows:
image

This is also there on main, but there it's not shown as a toast notification

Cargo.toml
jprochazk: I just wanted to try it. It's definitely much nicer to use, but maybe not worth it?

teh-cmc:
Quick UX feedback: I would expect the panel to close when I click outside of it.

crates/viewer/re_ui/src/notifications.rs
jprochazk: I don't know what the difference is, it came from `ToastKind` which was also undocumented. I guess someone wanted to use a different color?

crates/viewer/re_ui/src/notifications.rs
jprochazk: It's right below the `Notifications` text, I don't like how it repeats the word twice so close together

crates/viewer/re_ui/src/notifications.rs
jprochazk: It's only for toasts (I know you answered your own question later, just putting it here too)

crates/viewer/re_ui/src/notifications.rs
jprochazk: I couldn't figure out how to set `min_scrolled_height` via `egui::Window`

Cargo.toml
jprochazk: Replaced with `time`

jprochazk:
All of the following scenarios now work as expected:

- Click toggle button: Panel is open
- Click toggle button twice: Panel is closed
- Click toggle button, click inside the panel: Panel is open
- Click toggle button, click outside of panel: Panel is closed
- Click toggle button, press escape: Panel is closed
- Select something in the selection/time panel, click toggle button, press escape: Panel is closed, selection is preserved

The layout for individual notifications is still pretty gnarly, but it doesn't seem like I'll be able to use `egui_flex` for that, as it doesn't support `max-width` combined with `flex-grow`

jprochazk:
...or maybe I just flip the layout a bit and it will be good enough?

PR #8467, opened 2024-12-15
Add `MainThreadToken` to ensure file-dialogs only run on the main thread

emilk:
### What

We have a foot-gun in our code: our file dialogs (via `rfd`) [are only allowed to be run from the main thread (at least on Mac)](https://docs.rs/rfd/latest/rfd/#macos-non-windowed-applications-async-and-threading).
However, there is nothing stopping a user from accidentally calling these functions from a background thread, and if you test it on e.g. Linux, it may very well work.

So this PR introduces a new crate `re_capabilities` and a new type `MainThreadToken`. Any function that uses `rfd` should require the `MainThreadToken` as an argument. The `MainThreadToken` is only allowed to be created in `fn main`, and since it is neither `Send` nor `Sync`, this guarantees at compile-time that any functions that take a `MainThreadToken` can only be called from the main thread.

NOTE: I have no way to enforce that all uses of `rfd` also require `MainThreadToken` - we need to remember this ourselves, but I've made sure that all our _current_ uses of `rfd` require a `MainThreadToken`.

### DIFF
--- a/crates/utils/re_capabilities/src/main_thread_token.rs
+++ b/crates/utils/re_capabilities/src/main_thread_token.rs
@@ -1,0 +1,53 @@
+use static_assertions::assert_not_impl_any;
+
+/// A token that proves we are on the main thread.
+///
+/// Certain operations are only allowed on the main thread.
+/// These operations should require this token.
+/// For instance, any function using file dialogs (e.g. using [`rfd`](https://docs.rs/rfd/latest/rfd/)) should require this token.
+///
+/// The token should only be constructed in `fn main`, using [`MainThreadToken::i_promise_i_am_on_the_main_thread`],
+/// and then be passed down the call tree to where it is needed.
+/// [`MainTheadToken`] is neither `Send` nor `Sync`,
+/// thus guaranteeing that it cannot be found in other threads.
+///
+/// Of course, there is nothing stopping you from calling [`MainThreadToken::i_promise_i_am_on_the_main_thread`] from a background thread,
+/// but PLEASE DON'T DO THAT.
+/// In other words, don't use this as a guarantee for unsafe code.
+///
+/// There is also [`MainThreadToken::from_egui_ui`] which uses the implicit guarantee of egui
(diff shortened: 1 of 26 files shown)


crates/utils/re_capabilities/src/main_thread_token.rs
```diff
@@ -0,0 +1,53 @@
...
+    ///
+    /// Workaround until `impl !Send for X {}` is stable.
+    _dont_send_me: std::marker::PhantomData<*const ()>,
+}
+
+impl MainThreadToken {
+    /// Only call this from `fn main`, or you may get weird runtime errors!
+    pub fn i_promise_i_am_on_the_main_thread() -> Self {
+        #[cfg(not(target_arch = "wasm32"))]
+        debug_assert_eq!(std::thread::current().name(), Some("main"),
```
jprochazk, line 33: Why is this compiled out in release? And are we sure that every platform's main thread is called `main`? :grimacing:

crates/utils/re_capabilities/src/main_thread_token.rs
```diff
@@ -0,0 +1,53 @@
...
+            _dont_send_me: std::marker::PhantomData,
+        }
+    }
+
+    /// We should only create an `egui::Ui` on the main thread,
+    /// so having it is proof enough that we are on the main thread.
+    ///
+    /// Use this only in a code base where you are sure that egui is running only on the main thread.
+    #[cfg(feature = "egui")]
+    pub fn from_egui_ui(_ui: &egui::Ui) -> Self {
```
jprochazk, line 47: It seems to me like you can trivially create a `egui::Ui` on any thread:

```rust
egui::Ui::new(egui::Context::default(), egui::Id::new("test"), egui::UiBuilder::default());
```

And there isn't really any guarantee that someone won't do that.

jprochazk (approved):
If we make the `i_promise_i_am_on_the_main_thread` function into something that actually enforces the invariant, _and_ we don't have any way to bypass it, then I think this could work. But as it is, it's better than nothing, even if it depends on manual review :shrug:

Issue #8264, opened 2024-12-02
Space-view screenshotting on Web


emilk:
* (this issue assumes https://github.com/rerun-io/rerun/pull/8258 has been merged)

https://github.com/rerun-io/rerun/pull/8258 introduced egui-based screenshotting of space views. However, it is not implemented on web. For that to work we need to:

* https://github.com/emilk/egui/issues/5425
* https://github.com/emilk/egui/issues/5424 - this is optional, and probably the most work. When implemented, we can remove our own `crates/viewer/re_viewer_context/src/clipboard.rs`

emilk:
* https://github.com/rerun-io/rerun/pull/8472 implemented save-to-file one web
* Copy to clipboard still needs to be implemented

PR #8472, opened 2024-12-16
Implement save-view-as-screenshot on web

emilk:
### Related
* Part of #8264
* Using https://github.com/emilk/egui/pull/5438

Still missing: copy screenshot to clipboard

Also fixes a crash when screenshotting graph views.

### Usage
image

### Result
Node-link diagram

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1934,7 +1934,7 @@ checksum = "0d6ef0072f8a535281e4876be788938b528e9a1d43900b82c2569af7da799125"
 [[package]]
 name = "ecolor"
 version = "0.29.1"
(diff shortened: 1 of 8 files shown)


emilk:
We should probably codify what `ViewportCommand` is _supposed_ to do.

It would be nice if the user could add a `frame_countdown` to it, so that `0`=this frame, `1`=next frame etc, so we can avoid capturing the context menu (this would, I suspect, become a common problem).

I think it would be okay if on native `frame_countdown: 0` still captured the _next_ frame, i.e. treat `frame_countdown` as a _at least_ criteria, just to avoid taking on too much work.

If we add a `frame_countdown: usize` fields to `ViewportCommand` then `egui::Context::run` could filter out `ViewportCommand` with a delay, and schedule those for _next_ frame.

Or we add a `egui_ctx.schedule_command(frame_delay, cmd)`, but that is probably over-designing it a bit.

emilk:
* Fixed in https://github.com/emilk/egui/pull/5482

Wumpf (approved):

PR #8457, opened 2024-12-13
Fix broken interaction outside of graph views

emilk:
## Related

* Closes #8454 

UI interaction was sometimes broken outside of graph views. This PR is the cure.

* Reverts https://github.com/rerun-io/rerun/pull/8416

PR #8474, opened 2024-12-16
Fix single-click and double-click in the graph view

### LINKED ISSUE #8437: Double clicking on a node graph should select the entire entity
That's the behavior we have e.g. for points in the spatial views.

It already causes a focus action 👍 (==expand in the blueprint tree)

### LINKED ISSUE #8442: Graph view flickers across entities on node selections
When selecting a node in the graph view the graph briefly flickers, as can be seen in the video.

It seems to only happen when first selecting a node—returning to it does not seem to trigger it.

https://github.com/user-attachments/assets/ff294a4f-d9b9-4c84-b221-98f659a0db46

grtlr:
### Related

* Closes #8437
* Closes #8442

### What

This implements:
* Single-click on empty space to select view.
* Double-click on node to select entire entity.

Merging @emilk's recent changes (#8469 and #8457) seems to have fixed the flickering on selection too.

### DIFF
--- a/crates/viewer/re_view_graph/src/ui/draw.rs
+++ b/crates/viewer/re_view_graph/src/ui/draw.rs
@@ -7,7 +7,6 @@ use egui::{
 use re_chunk::EntityPath;
 use re_data_ui::{item_ui, DataUi as _};
 use re_entity_db::InstancePath;
-use re_log_types::Instance;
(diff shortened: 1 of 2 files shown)


teh-cmc:
If clicking the background selects the view, then hovering the background should highlight it

PR #8492, opened 2024-12-16
Add better support for scrolling content to our modals

abey79:
### What

This PR builds proper scrollability to our modals, thereby fixing:
- an ugly glitch with large add/remove entity modal
- add view/container modal not scrollable with very small viewer window size

#### Before

image

#### After

image

image

### DIFF
--- a/crates/viewer/re_ui/src/modal.rs
+++ b/crates/viewer/re_ui/src/modal.rs
@@ -1,5 +1,6 @@
-use crate::{DesignTokens, UiExt as _};
 use eframe::emath::NumExt;
+
+use crate::{DesignTokens, UiExt as _};
 
 /// Helper object to handle a [`ModalWrapper`] window.
 ///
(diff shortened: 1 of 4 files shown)


teh-cmc:
Can we add a checklist for modals? Even something extremely barebones ("click on add container and make sure it's not insane"), so the release manager are reminded to check them.

crates/viewer/re_ui/src/modal.rs
```diff
@@ -230,3 +272,34 @@ impl ModalWrapper {
         });
     }
 }
+
+/// Utility to produce a [`egui::Frame`] with padding on some sides.
+#[allow(clippy::fn_params_excessive_bools)]
+#[inline]
+fn view_padding_frame(left_and_right: bool, top: bool, bottom: bool) -> egui::Frame {
```
teh-cmc, line 279: I agree with clippy on this one, introducing a struct would make the code above much clearer.
```rust
struct ViewPaddingFrameParams {
    left_and_right: bool,
    top: bool,
    bottom: bool,
}
```

```rust
view_padding_frame(ViewPaddingFrameParams { left_and_right: false, top: false, bottom: true});
```

teh-cmc (approved):
I don't understand the code, but it works.

See my comment above: add a checklist please.

PR #8495, opened 2024-12-17
Add missing click and hover interactions for views

### LINKED ISSUE #5138: Hovering and clicking text document views does nothing
On every other view it hovers and/or selects the view itself

### LINKED ISSUE #8479: Fix clicking and hovering various views
Fix clicking and hovering for the following views (they should highlight correctly):

* Map view (hovering)
* Graph view (hovering)
* #5138

grtlr:
### Related

* Closes #8479
* Closes #5138

### What

This adds missing view interactions the following views:

* Map view
* Graph view
* TextDocument view

@Wumpf @abey79 I feel like the pattern repeats itself and it's easy to miss when writing new views. So I wonder if this could be unified. I see several options for this:

* repurpose `handle_select_hover_drag_interactions` which only seems to be used for `DataResult` right now
* Making each view return a `Result<(Response, ViewSystemExecutionError>` (instead of `()`) and handling it outside of the view.
* Creating a function analogous to `handle_select_hover_drag_interactions`.

### DIFF
--- a/crates/viewer/re_view_graph/src/view.rs
+++ b/crates/viewer/re_view_graph/src/view.rs
@@ -206,6 +206,10 @@ Display a graph of nodes and edges.
             }
         });
 
+        if resp.hovered() {
+            ctx.selection_state().set_hovered(Item::View(query.view_id));
+        }
+
         if resp.clicked() {
             // clicked elsewhere, select the view
             ctx.selection_state()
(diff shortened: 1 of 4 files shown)


grtlr:
Pointing out the obvious: the map view is missing from the checklist, mainly for simplicity.

teh-cmc (approved):

teh-cmc:
> Pointing out the obvious: the map view is missing from the checklist, mainly for simplicity.

I'd prefer if it was there though. Why would it be complicated to add a map there?

grtlr:
I thought I had to send a blueprint with it, but it works just fine without it—so I added it.

PR #8496, opened 2024-12-17
Remove debug information from graph view selection panel

grtlr:
### What

Title.

This was a small part of @gavrelina's feedback.


grtlr:
I've decided to remove the debug information entirely, as I never needed anymore anyways. Note that we still have the bounding box as part of the selection panel.

emilk (approved):

PR #8499, opened 2024-12-17
Disable tagging entirely for 0.21

teh-cmc:
There are too many pitfalls that users can fall into because the current logging APIs are not designed with tagging in mind, and we have not yet communicated at all about any of this.

This has to be done below the chunk level, directly at the descriptor level.

### DIFF
--- a/crates/store/re_chunk_store/tests/gc.rs
+++ b/crates/store/re_chunk_store/tests/gc.rs
@@ -387,6 +387,8 @@ fn protected_time_ranges() -> anyhow::Result<()> {
         }
     }
 
+    eprintln!("{store}");
+
     let (events, _) = store.gc(&protect_time_range(ResolvedTimeRange::new(1, 4)));
     assert_eq!(events.len(), 0);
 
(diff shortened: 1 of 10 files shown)


teh-cmc:
@rerun-bot full-check

crates/store/re_chunk_store/tests/gc.rs
```diff
@@ -387,6 +387,8 @@ fn protected_time_ranges() -> anyhow::Result<()> {
         }
     }
 
+    eprintln!("{store}");
```
Wumpf, line 390:
```suggestion
```

debug vestigue?

Wumpf (approved):
🤞 hopefully that was all `.entry` callsites

crates/store/re_chunk_store/tests/gc.rs
teh-cmc: On purpose -- I like the extra context when things fail.

teh-cmc:
> 🤞 hopefully that was all `.entry` callsites

That was all the offending one for sure -- I commented out the DerefMut and all public qualifiers :+1:

PR #8504, opened 2024-12-17
Catalog view: fetch StoreInfo from the catalog and do catalog data enrichment on the viewer side

zehiko:
### What

Rerun Data Platform now supports simple filtering by recording ids and also stores (some of the) fields required to create ``StoreInfo``, hence we can fetch it from the catalog. 

Also, instead of doing all the arrow data manipulation to make ``DataframePart`` convertible to Rerun Chunk on the ReDap side, we now do it on the viewer side. 

### Testing done
* catalog view basic UI testing with multiple recordings (and no recordings at all)


jleibs (approved):
Nice!

PR #8510, opened 2024-12-17
Fix graph view hover (proper this time)

Wumpf:
Previous fix caused flickering, this should be more robust 🤞


Wumpf:
closed in favor of
* https://github.com/rerun-io/rerun/pull/8512

Issue #7026, opened 2024-07-31
Formalize component joins that are happening within visualizers


Wumpf:
There's various kinds of joins on how we can (and do) join components within an archetype. This happens ad-hoc in visualizers and is only sparsely documented.

- Batch-joined. I.e. Arch[index] = (C1[index], C2[index], ...). Example: Points
- Instance-joined. I.e. Arch[0] = (C1[:], C2[:], ...). Example: Mesh3D
- Product-joined. I.e. Arch[i1, i2] = (C[i1], C[i2]).

`Mesh3D` & `Asset3D` are a hybrid of Instance-joined components Vertex / Indices with Product-joined component of `LeafTransform3D`.

Formalizing this would lead more easily to
* better documentation
* better ui
* consistent user queries

Loosely related to:
* https://github.com/rerun-io/rerun/issues/5259
* https://github.com/rerun-io/rerun/issues/6910

Wumpf:
Another example of this are graph nodes!
(cc: @grtlr )

teh-cmc:
Also very related to materialized aggregated queries as compute graphs.

PR #8511, opened 2024-12-17
Hide graph edges from individual graph node ui

Wumpf:
### Related

* https://github.com/rerun-io/rerun/issues/7026

### What

Lots of considerations & (soft) implications in this hack, see comment.

The important outcome here is that node hover now looks like this:
image
node selection:
image
and node entity selection (note that edges are still around:
image
also you can still browse the edges:
image


grtlr (approved):
Thank you! 🤩

PR #8512, opened 2024-12-17
Fix graph view node highlight on hover

grtlr:
### What

This fixes a problem introduced by #8495 that was only partially fixed by #8508, with graph view highlights flickering (ping-ponging) in the blueprint view.

### DIFF
--- a/crates/viewer/re_view_graph/src/ui/draw.rs
+++ b/crates/viewer/re_view_graph/src/ui/draw.rs
@@ -330,6 +330,7 @@ pub fn draw_graph(
     layout: &Layout,
     query: &ViewQuery<'_>,
     lod: LevelOfDetail,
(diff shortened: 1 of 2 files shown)


Wumpf (approved):

emilk (changes requested):
I can't click to select nodes in the web demo

crates/viewer/re_view_graph/src/ui/draw.rs
```diff
@@ -363,11 +364,9 @@ pub fn draw_graph(
                     });
                 });
 
-                ctx.handle_select_hover_drag_interactions(
```
Wumpf, line 366: forgot to call this function then delayed

Wumpf (approved):

teh-cmc (approved):

Issue #8518, opened 2024-12-17
Improve the entity path filter mutation API


abey79:
Currently, all methods mutating the entity path filter in a `ViewContents` write the mutated filter to the blueprint store _but leaves the deserialised version unchanged_. This means that all mutations are lost except the last one within a frame, which is a major foot-gun. We should fix this API such as to allow arbitrary number of mutations during a frame (while retaining interior mutability for the `ViewportBlueprint` structure and friends).

PR #8519, opened 2024-12-17
Fix multi-entity drag-and-drop only adding one entity

abey79:
### Related

* Related to #8518
* Related to #8431

### What

Fix a bug in #8431 where multi-entity drag-and-drop would only add a single entity due to a foot-gun in our entity path filter mutation API (see #8518 for more). Also add a new checklist.

### DIFF
--- a/crates/viewer/re_viewport/src/viewport_ui.rs
+++ b/crates/viewer/re_viewport/src/viewport_ui.rs
@@ -6,7 +6,7 @@ use ahash::HashMap;
 use egui_tiles::{Behavior as _, EditAction};
 
 use re_context_menu::{context_menu_ui_for_item, SelectionUpdateBehavior};
-use re_log_types::{EntityPath, EntityPathRule};
+use re_log_types::{EntityPath, EntityPathRule, RuleEffect};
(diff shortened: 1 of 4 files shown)


crates/viewer/re_viewport/src/viewport_ui.rs
```diff
@@ -277,14 +277,27 @@ impl ViewportUi {
...
+                });
+
+            // for entity in entities {
+            //     if can_entity_be_added(entity) {
+            //         view_blueprint.contents.raw_add_entity_inclusion(
+            //             ctx,
+            //             EntityPathRule::including_subtree(entity.clone()),
+            //         );
+            //     }
+            // }
```
Wumpf: with this pr I have the context why this is still here but no point in letting it rot. the upcoming change will be the `mutate` function and then things change here again anyways

```suggestion
```

Wumpf (approved):

crates/viewer/re_viewport/src/viewport_ui.rs
abey79: I just noticed it a removed it already :)

PR #8521, opened 2024-12-17
Update the remote APIs to take and send a Table

jleibs:
This allows the update APIs to now support multi-recording updates, and generally gives users more direct control over things like the column metadata, etc.

### DIFF
--- a/rerun_py/rerun_bindings/types.py
+++ b/rerun_py/rerun_bindings/types.py
@@ -68,4 +68,7 @@ A type alias for index values.
 This can be any numpy-compatible array of integers, or a [`pa.Int64Array`][]
 """
 
-MetadataLike: TypeAlias = pa.Array
+TableLike: TypeAlias = Union[pa.Table, pa.RecordBatch, pa.RecordBatchReader]
+"""
+A type alias for TableLike pyarrow objects.
+"""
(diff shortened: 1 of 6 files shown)


rerun_py/rerun_bindings/types.py
```diff
@@ -68,4 +68,7 @@
 This can be any numpy-compatible array of integers, or a [`pa.Int64Array`][]
 """
 
-MetadataLike: TypeAlias = pa.Array
+TableLike: TypeAlias = Union[pa.Table, pa.RecordBatch, pa.RecordBatchReader]
```
zehiko, line 71: nice we can make it generic over all these options.

rerun_py/src/remote.rs
```diff
@@ -226,48 +205,33 @@ impl PyStorageNodeClient {
...
-                })
-                .collect::<PyResult<Vec<_>>>()?
-                .into_iter()
-                .unzip();
+            let metadata = metadata.into_record_batch()?;
 
-            let schema = arrow2::datatypes::Schema::from(schema);
-
-            let data = arrow2::chunk::Chunk::new(data);
+            // TODO(jleibs): This id name should probably come from `re_protos`
```
zehiko, line 225: (I was thinking about this yesterday and came to the same conclusion. We should put these rerun mandatory fields somewhere so that we can re-use on both sides)

zehiko (approved):

PR #8523, opened 2024-12-18
Fix cargo warning

abey79:
The warning:

```
warning: /Users/hhip/src/rerun/crates/utils/re_capabilities/Cargo.toml: `default-features` is ignored for egui, since `default-features` was not specified for `workspace.dependencies.egui`, this could become a hard error in the future
```


Wumpf (approved):

Wumpf:
begs the question whether the top level cargo.toml should do without default features though

PR #8524, opened 2024-12-18
Enable GH releases for RC

abey79:
This PR enables the creation of GH releases (as draft and marked as pre-release) for RCs.

Motivations for doing so:
- test that part of the workflow, in particular that all binaries are correctly produced
- test the binary using `cargo binstall`
- these releases a single-click deletable from GH (if we want to do so)
- GH releases marked as "pre-release" aren't displayed in the GH project page

Pending questions: do we want that?  what about alphas?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -321,7 +321,7 @@ jobs:
 
   github-release:
     name: "GitHub Release"
-    if: inputs.release-type == 'final'
+    if: inputs.release-type == 'rc' || inputs.release-type == 'final'
     needs:
       [
         version,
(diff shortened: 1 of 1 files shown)


teh-cmc (approved):
I like the idea. I have no clue whether these are the appropriate yaml incantations though.

abey79:
I'm unsure if there is something specific to be done so that artefacts are also uploaded on RC GH releases

jprochazk:
I believe https://github.com/rerun-io/rerun/blob/main/.github/workflows/on_gh_release.yml should be changed to also run on `prereleased`? I'm not actually sure

jprochazk (approved):
Maybe it will just work as is.

> The prereleased type will not trigger for pre-releases published from draft releases, but the published type will trigger. If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.

abey79:
Thanks for looking into it. Let's merge and find out when we do 0.22...

PR #8526, opened 2024-12-18
Fix undo not working in some entity drag-and-drop instances

abey79:
This is a band-aid patch to address an issue with entity path filter which would result in the following behaviours:

- the entity path filter would be perma-writen to the blueprint store when the view is selected (in some cases)
- in such cases, this would break undo on user change (aka with entity drag-and-drop)

The core issue is deeper and might cause panics with rust 1.81+. The TL;DR is that the current `EntityPathFilter` type is ambiguous as to whether its content has substitutions (aka $origin)  applied or not. In particular `Eq` and `Ord` apply to unsubstituted, resp. substituted content (which can lead to the above panic). This should be further cleaned by having two structures, one unsubstituted and another substituted.

### DIFF
--- a/crates/viewer/re_selection_panel/src/selection_panel.rs
+++ b/crates/viewer/re_selection_panel/src/selection_panel.rs
@@ -559,6 +559,10 @@ fn entity_path_filter_ui(
     }
 
     // Apply the edit.
+    //
+    // NOTE: The comparison of `EntityPathFilter` is done on the _expanded_ data (i.e. with variables substituted),
+    // so we must make sure to expand the new filter too before we compare it to the existing one.
+    // See <https://github.com/rerun-io/rerun/pull/8526>
     let new_filter =
         EntityPathFilter::parse_forgiving(&filter_string, &EntityPathSubs::new_with_origin(origin));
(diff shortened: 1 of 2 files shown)


teh-cmc (approved):

crates/viewer/re_selection_panel/src/selection_panel.rs
```diff
@@ -559,7 +559,8 @@ fn entity_path_filter_ui(
     }
 
     // Apply the edit.
-    let new_filter = EntityPathFilter::parse_forgiving(&filter_string, &Default::default());
+    let new_filter =
+        EntityPathFilter::parse_forgiving(&filter_string, &EntityPathSubs::new_with_origin(origin));
```
teh-cmc, line 563:
```suggestion
    //
    // NOTE: The comparison of `EntityPathFilter` is done on the _expanded_ data (i.e. with variables substituted),
    // so we must make sure to expand the new filter too before we compare it to the existing one.
    let new_filter =
        EntityPathFilter::parse_forgiving(&filter_string, &EntityPathSubs::new_with_origin(origin));

PR #8527, opened 2024-12-18
Use single log call in graph examples

Wumpf:
Makes the graph examples look a lil bit nicer and more compact.

Note that this is a very recent thing that we can do this in Rust 🥳 

Also, embedds `graph_directed` in the examples (this has been done directly on the release branch already)

* [x] full check passed in order to confirm the example output is still the same (tested this locally but who knows)


Wumpf:
@rerun-bot full-check

teh-cmc (approved):

PR #8529, opened 2024-12-18
Update MSRV to 1.81

emilk:


emilk:
@rerun-bot full-check

crates/viewer/re_ui/src/color_table.rs
```diff
@@ -157,21 +157,25 @@ impl ColorTable {
     }
 
     #[inline]
+    #[allow(dead_code)]
```
abey79, line 160: let's make the color table `pub`

abey79 (approved):
lgtm!

Are you expecting the `allow` -> `expect` migration to be more painful than "just" a massive grep?

emilk:
> lgtm!
> 
> Are you expecting the `allow` -> `expect` migration to be more painful than "just" a massive grep?

Oh yes, a lot more painful. A lot of the `allow`:s are there only for certain targets (e.g. wasm32)

PR #8531, opened 2024-12-18
Remove manual truncation of tensor shape UI

### LINKED ISSUE #8525: `TensorShape` gets truncated and there's no way to see it fully
If I have a long `TensorShape`, it gets truncated, which is nice, but i cannot find anywhere where I can see it fully.
Image
Image

abey79:
### Related

* fixes #8525

### What

Truncation now happens at egui test level, so we no longer need that manual truncation code. (This code is soon going to disappear anyways but at least it'll disappear in a good state 🤷🏻)


Wumpf (approved):
👍 
trusting you tried this with looong shapes :)

abey79:
> rusting you tried this with looong shapes :)

They get properly egui-text-truncated as expected 👍🏻

PR #8532, opened 2024-12-18
Fix: `./scripts/lint.py` ignored all under `crates/build`

emilk:


teh-cmc (approved):
Haaaaaaaaa, so that's why github search never worked for things in `build/` :melting_face:

PR #8533, opened 2024-12-18
Add graph view example video (with subtitles) to `CHANGELOG`

grtlr:
I did not have to re-encode so I just uploaded the original.


teh-cmc (approved):
:pray:

PR #8534, opened 2024-12-18
Start porting fallbacks to arrow-rs

emilk:
### Related
* Part if https://github.com/rerun-io/rerun/issues/3741

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6559,6 +6559,7 @@ dependencies = [
  "glam",
  "itertools 0.13.0",
  "nohash-hasher",
+ "re_arrow2",
  "re_chunk_store",
  "re_entity_db",
  "re_log",
@@ -6872,6 +6873,7 @@ dependencies = [
  "ahash",
  "anyhow",
  "arboard",
+ "arrow",
  "bit-vec",
  "bitflags 2.6.0",
  "bytemuck",
@@ -6949,6 +6951,7 @@ name = "re_viewport_blueprint"
 version = "0.21.0-alpha.1+dev"
 dependencies = [
  "ahash",
+ "arrow",
  "egui",
  "egui_tiles",
  "itertools 0.13.0",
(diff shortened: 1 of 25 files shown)


crates/store/re_types_core/src/size_bytes/mod.rs
```diff
@@ -0,0 +1,270 @@
+mod arrow2_sizes;
```
emilk, line 1: This code is all old, just split a file in two

crates/store/re_types_core/src/size_bytes/arrow2_sizes.rs
```diff
@@ -1,386 +1,23 @@
-use std::collections::{BTreeMap, BTreeSet, HashMap, VecDeque};
```
emilk, line 1: Just split a file in two

emilk:
@rerun-bot full-check

crates/store/re_types_core/src/size_bytes/arrow_sizes.rs
```diff
@@ -0,0 +1,17 @@
...
+        Box::<dyn arrow2::array::Array>::from(self).heap_size_bytes()
+    }
+}
+
+impl SizeBytes for ArrayRef {
+    #[inline]
+    fn heap_size_bytes(&self) -> u64 {
+        Box::<dyn arrow2::array::Array>::from(self.as_ref()).heap_size_bytes()
+    }
+}
```
teh-cmc, line 17: I think they have much _much_ better tools for these kinds of things now in arrow1.

We'll need to revisit these implementations to use these things at some point (see e.g. https://docs.rs/arrow/latest/arrow/array/trait.Array.html#tymethod.get_array_memory_size).

A TODO would be nice.

crates/store/re_types_core/src/size_bytes/mod.rs
emilk: Probably gonna move this out to its own crate pretty soon

teh-cmc (approved):

crates/store/re_types_core/src/size_bytes/arrow_sizes.rs
emilk: I'll just switch to using get_array_memory_size

PR #8535, opened 2024-12-18
Update colab notebook link

jprochazk:
This way everyone with edit rights should be able to access it


jprochazk:
Nope, some random changes got mixed in, give me a second

PR #8540, opened 2024-12-19
Find and annihilate slow component iterators

teh-cmc:
* `HybridResultsChunkIter::component` is now `HybridResultsChunkIter::component_slow`, in order to scare people off.
* I have removed *a lot* of those, and will remove a few more in follow up PRs, but I need to add more iteration tools first.


crates/viewer/re_view/src/results_ext.rs
```diff
@@ -421,8 +421,12 @@ pub struct HybridResultsChunkIter<'a> {
 impl<'a> HybridResultsChunkIter<'a> {
     /// Iterate as indexed deserialized batches.
     ///
+    /// TODO(#5305): Note that this uses the old codegen'd deserialization path, which does some
+    /// very unidiomatic Arrow things, and is therefore very slow at the moment. Avoid this on
+    /// performance critical paths.
+    ///
     /// See [`Chunk::iter_component`] for more information.
```
emilk, line 428: It would be good to refer to the better alternatives here

emilk (approved):
LGTM if tested

PR #8542, opened 2024-12-19
Show the `GraphNode` as a label by default

grtlr:
### What

This changes the default behavior of the graph view to show the `GraphNode` ids as a label by default, if no labels are logged explicitly. Of course, this can still be overridden by `ShowLabels`

### DIFF
--- a/crates/viewer/re_view_graph/src/visualizers/nodes.rs
+++ b/crates/viewer/re_view_graph/src/visualizers/nodes.rs
@@ -119,6 +119,10 @@ impl VisualizerSystem for NodeVisualizer {
                             text: label.clone(),
                             color,
                         },
+                        (None, true) => Label::Text {
+                            text: node.0 .0.clone(),
+                            color,
+                        },
                         _ => Label::Circle {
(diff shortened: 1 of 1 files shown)


Wumpf (commented):
makes sense on one hand, but on the other I don't get the implications of this: I was trying before/after in the viewer with nodes that don't have label, enabling `show_labels` in the ui and in each case it showed the node name as label already 😕 
image


can you share a relevant before/after?

grtlr:
The example that you posted has a `Text` component logged.

This code path is for when the user does not explicitly log `Text`. Previously, in this case the nodes were rendered as circles by default, this PR changes this to render the `GraphNode` value (which already is a `Utf8`) as a label, with the option to turn of the labels. This works because the default of `ShowLabels` in the context of the graph view is `true`.

Wumpf:
> The example that you posted has a Text component logged.

oh wow I was looking for "label" literally and didn't see that there was 'text' already there.

Wumpf (approved):

Issue #8530, opened 2024-12-18
`EntityPathFilter` has undefined behavior for sorting


Wumpf:
its `Ord` and `Eq` don't check the same fields.
We heavily depend on this behavior in a bunch of places: we want equality to only depend on raw expressions but sorting should be done on expressions with variable substition (we sort from least specific to most specifiic - we want this for both evaluation and UI display)

-> We need to split those concerns into two different types, one using a hashmap of unresolved rules and the other one using a btreemap of resolved rules.

This can cause crashes when building with Rust 1.81+, see https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html#new-sort-implementations

Wumpf:
Users using the Rust dataframe api will always use unresolved entity path filter (this will be just called `EntityPathFilter`) whereas the viewer internally oftentimes want to use `ResolvedEntityPathFilter` which can only be created from a `EntityPathFilter` + substituions

rasmusgo:
Is this the issue that gives these crashes? I get these a lot when trying to move tabs around and sometimes just by hovering stuff in the blueprint pane.

```
thread 'main' panicked at 'user-provided comparison function does not correctly implement a total order'
core/src/slice/sort/shared/smallsort.rs:860
```

Wumpf:
@rasmusgo yes, that's exactly that. While testing the release we hit it only once and couldn't repro it again which is why we decided to move on :/. Also, it doesn't affect our python builds since they build with an older rust version that doesn't have that check yet.
Unfortunately, I wasn't able to fix this without breaking changes in the crate boundaries so far, so it can't go into a patch release

PR #8543, opened 2024-12-19
`EntityPathFilter` variable substitutions are now delegated to (new) `ResolvedEntityPathFilter`

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/8530

### What

(Motivation see ticket)
Originally I wanted `EntityPathFilter` to be a hashmap of rules. However, I realized that we still want a defined order when storing path filters into the blueprint (in order to avoid changes to it when only order changes!). Therefore, it is still a btreemap but now _sorts alphabetically_!
Generally, I struggled a bit of where to put resolved vs unresolved entity path filter/rules, but I think the end-result here makes sense for the most part and keeps the semantics clean enough: in the viewer resolved entity paths should be used whenever working with known paths whereas unresolved are used when dealing with user inputs and when storing back to the blueprint.

### DIFF
--- a/crates/store/re_log_types/src/path/entity_path_filter.rs
+++ b/crates/store/re_log_types/src/path/entity_path_filter.rs
@@ -72,7 +72,8 @@ impl TryFrom<&str> for EntityPathFilter {
 
 /// An [`EntityPathFilter`] with all variables Resolved.
 ///
-/// [`ResolvedEntityPathFilter`] sorts the rule by entity path, with recursive coming before non-recursive.
+/// [`ResolvedEntityPathFilter`] sorts the rule by specificity of the entity path,
+/// with recursive coming before non-recursive.
 /// This means the last matching rule is also the most specific one.
 /// For instance:
 ///
@@ -298,9 +299,8 @@ impl EntityPathFilter {
     /// The rest of the line is trimmed and treated as an entity path after variable substitution through [`Self::resolve_forgiving`]/[`Self::resolve_strict`].
     ///
     /// Conflicting rules are resolved by the last rule.
-    #[allow(clippy::unnecessary_wraps)]
+    #[allow(clippy::unnecessary_wraps)] // TODO(andreas): Do some error checking here?
     pub fn parse_strict(rules: &str) -> Result<Self, EntityPathFilterError> {
-        // TODO(andreas): Do some error checking here.
         Ok(Self::parse_forgiving(rules))
     }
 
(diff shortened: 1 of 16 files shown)


crates/store/re_log_types/src/path/entity_path_filter.rs
```diff
@@ -47,7 +51,28 @@ impl EntityPathSubs {
...
+    type Error = EntityPathFilterError;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        Self::parse_strict(value)
+    }
+}
+
+/// An [`EntityPathFilter`] with all variables Resolved.
+///
+/// [`ResolvedEntityPathFilter`] sorts the rule by entity path, with recursive coming before non-recursive.
```
teh-cmc, line 75: This should really emphasize that it sorts based on specificity.

crates/store/re_log_types/src/path/entity_path_filter.rs
```diff
@@ -309,15 +295,13 @@ impl EntityPathFilter {
...
     /// Conflicting rules are resolved by the last rule.
-    pub fn parse_strict(
-        rules: &str,
-        subst_env: &EntityPathSubs,
-    ) -> Result<Self, EntityPathFilterParseError> {
-        let split_rules = split_whitespace_smart(rules);
-        Self::from_query_expressions_strict(split_rules, subst_env)
+    #[allow(clippy::unnecessary_wraps)]
+    pub fn parse_strict(rules: &str) -> Result<Self, EntityPathFilterError> {
+        // TODO(andreas): Do some error checking here.
```
teh-cmc, line 303: Did you mean that for this PR? for later?

teh-cmc (approved):
I gave it a very quick/shallow look. There's no way I can wrap my head around all the subtleties that all this implies, but in any case the general idea makes sense to me.

The only thing I find odd is the dataframe APIs requiring resolved filters? Can we not do that? :thinking: 

In any case, the faster this is in, the better -- only some soak time will tell the full story here...

crates/viewer/re_viewport_blueprint/src/view.rs
```diff
@@ -71,12 +71,16 @@ impl ViewBlueprint {
     pub fn new(view_class: ViewClassIdentifier, recommended: RecommendedView) -> Self {
         let id = ViewId::random();
 
+        // TODO(andreas): do we need more context here?
```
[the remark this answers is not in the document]
Wumpf: so you're saying you _do_ need more context here? 😄

crates/store/re_log_types/src/path/entity_path_filter.rs
Wumpf: later

PR #8548, opened 2024-12-19
Remove `Chunk::iter_component_arrays`

teh-cmc:
This method makes no sense. It's a complete anti-pattern. The whole point of the Chunk level methods is to pay the cost of reflection/downcasting/deserialization once for the whole Chunk, this can never happen with the way this method is defined.

I'm not sure what I was thinking back then. I likely wasn't. There is never a good reason to use this.


teh-cmc:
@rerun-bot full-check

Wumpf (approved):
makes sense to me

PR #8549, opened 2024-12-19
Short circuit graph simulation if all nodes are fixed

grtlr:
### What

Title.

Wonder if we should even show the forces section in the selection panel in this case—but that might be confusing for the user too.

### DIFF
--- a/crates/viewer/re_view_graph/src/layout/provider.rs
+++ b/crates/viewer/re_view_graph/src/layout/provider.rs
@@ -98,8 +98,13 @@ pub fn update_simulation(
 }
 
 pub struct ForceLayoutProvider {
-    simulation: fj::Simulation,
+    // If all nodes are fixed, we can skip the simulation.
+    simulation: Option<fj::Simulation>,
     pub request: LayoutRequest,
+}
+
+fn requires_simulation(request: &LayoutRequest) -> bool {
+    request.all_nodes().any(|(_, v)| v.fixed_position.is_none())
 }
 
 fn considered_edges(request: &LayoutRequest) -> Vec<(usize, usize)> {
@@ -117,6 +122,13 @@ fn considered_edges(request: &LayoutRequest) -> Vec<(usize, usize)> {
 
 impl ForceLayoutProvider {
     pub fn new(request: LayoutRequest, params: &ForceLayoutParams) -> Self {
+        if !requires_simulation(&request) {
+            return Self {
+                simulation: None,
+                request,
+            };
+        }
+
         let nodes = request.all_nodes().map(|(_, v)| fj::Node::from(v));
         let radii = request
             .all_nodes()
@@ -129,7 +141,7 @@ impl ForceLayoutProvider {
         let simulation = update_simulation(simulation, params, edges, radii);
 
         Self {
-            simulation,
+            simulation: Some(simulation),
             request,
         }
     }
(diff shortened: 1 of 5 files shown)


crates/viewer/re_view_graph/src/layout/provider.rs
```diff
@@ -98,10 +98,15 @@ pub fn update_simulation(
...
 pub struct ForceLayoutProvider {
-    simulation: fj::Simulation,
+    // If all nodes are fixed, we can skip the simulation.
+    simulation: Option<fj::Simulation>,
     pub request: LayoutRequest,
 }
 
+fn requires_simulation(request: &LayoutRequest) -> bool {
+    request.all_nodes().any(|(_, v)| v.fixed_position.is_none())
+}
```
teh-cmc, line 108: Don't you own `LayoutRequest`? Any reason this isn't just a method on it?

crates/viewer/re_view_graph/src/layout/provider.rs
```diff
@@ -161,24 +180,32 @@ impl ForceLayoutProvider {
...
+        let positions: &mut dyn Iterator<Item = Pos2> = if let Some(simulation) = &self.simulation {
+            &mut simulation
+                .positions()
+                .map(|[x, y]| Pos2::new(x as f32, y as f32))
+        } else {
+            &mut self.request.all_nodes().map(|(_, v)| {
+                v.fixed_position
+                    .expect("if there is no simulation, all nodes have to be fixed")
+            })
+        };
```
teh-cmc, line 200: Don't erase the iterator here, this will force virtual calls for every single iteration.

Use `Either` instead so the two branches can monomorphize gently (see e.g. https://github.com/rerun-io/rerun/blob/35327062c114e8805233779043b742ae26c37260/crates/viewer/re_view/src/lib.rs#L67 for an example).

crates/viewer/re_view_graph/src/layout/provider.rs
```diff
@@ -161,24 +180,32 @@ impl ForceLayoutProvider {
...
         // order of the nodes is the same as in the `request`.
-        let mut positions = self.simulation.positions();
+        let positions: &mut dyn Iterator<Item = Pos2> = if let Some(simulation) = &self.simulation {
+            &mut simulation
+                .positions()
+                .map(|[x, y]| Pos2::new(x as f32, y as f32))
+        } else {
+            &mut self.request.all_nodes().map(|(_, v)| {
+                v.fixed_position
+                    .expect("if there is no simulation, all nodes have to be fixed")
```
teh-cmc, line 198: Don't use `expect` in release builds. If something can crash, it will 100% crash at some point (and take the entire viewer as well as all the user's unsaved data with it).

This could be a `.map` + `expect()` in debug, and a `.filter_map` + `error!` log in release.

crates/viewer/re_view_graph/src/layout/provider.rs
```diff
@@ -161,24 +180,32 @@ impl ForceLayoutProvider {
...
 
         let mut layout = Layout::empty();
 
         for (entity, graph) in &self.request.graphs {
             let mut current_rect = Rect::NOTHING;
 
             for (node, template) in &graph.nodes {
-                let [x, y] = positions.next().expect("positions has to match the layout");
-                let pos = Pos2::new(x as f32, y as f32);
+                let pos = positions.next().expect("positions has to match the layout");
```
teh-cmc, line 208: Same here, no `expect`.

teh-cmc (changes requested):

crates/viewer/re_view_graph/src/layout/provider.rs
grtlr: Ah, I didn't know about `itertools::Either`. TIL :thanks:

teh-cmc (approved):
:ship:

PR #8553, opened 2024-12-20
Introduce new Chunk iteration APIs

teh-cmc:
* Way more ergonomic
* Same performance
* Less code
* Support for structs!
* Basis for new upcoming codegen'd deserialization APIs (...at some point)

### DIFF
--- a/crates/store/re_chunk/src/iter.rs
+++ b/crates/store/re_chunk/src/iter.rs
@@ -2,8 +2,9 @@ use std::sync::Arc;
 
 use arrow2::{
     array::{
-        BooleanArray as Arrow2BooleanArray, FixedSizeListArray as Arrow2FixedSizeListArray,
-        ListArray as Arrow2ListArray, PrimitiveArray as Arrow2PrimitiveArray,
+        Array as Arrow2Array, BooleanArray as Arrow2BooleanArray,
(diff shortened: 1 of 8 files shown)


crates/store/re_chunk/src/iter.rs
```diff
@@ -543,6 +544,469 @@ impl Chunk {
...
+    // NOTE: No need for validity checks here, `iter_offsets` already takes care of that.
+    Either::Right(component_offsets.map(move |(idx, len)| {
+        let inner_offsets = &inner_offsets.as_slice()[idx..idx + len];
+        let inner_lengths = &inner_lengths.as_slice()[idx..idx + len];
+        izip!(inner_offsets, inner_lengths)
+            .map(|(&idx, &len)| {
+                let idx = idx as usize;
+                bytemuck::cast_slice(&values[idx * size..idx * size + len * size])
+            })
+            .collect_vec()
```
Wumpf, line 921: everything except `collect_vec` is not dependent on `N` -> we likely can save a lot of template instantiations by putting this into a second level utility method that only depends on `T` but not on `N`

teh-cmc: I don't think I'm ready to add even more layers and complex return types to all of this (we need the intermediary values too...) for something I don't really have a way to measure right now :/ Also this is private, no the number of instantiations is fixed, at least.

PR #8554, opened 2024-12-20
Update all visualizers to use new Chunk iteration APIs

teh-cmc:
* DNM: requires #8553

PR #8555, opened 2024-12-20
Update `GraphEdge` visualizer to use sliced struct fields

teh-cmc:
* DNM: requires #8554

### DIFF
--- a/crates/viewer/re_view_graph/src/visualizers/edges.rs
+++ b/crates/viewer/re_view_graph/src/visualizers/edges.rs
@@ -1,10 +1,6 @@
-use re_chunk::LatestAtQuery;
+use re_chunk::{ArchetypeFieldName, LatestAtQuery};
 use re_log_types::{EntityPath, Instance};
-use re_types::{
-    self, archetypes,
-    components::{self, GraphEdge, GraphNode},
-    Component as _,
-};
+use re_types::{self, archetypes, components, datatypes, Component as _};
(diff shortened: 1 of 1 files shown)


crates/viewer/re_view_graph/src/visualizers/edges.rs
```diff
@@ -52,24 +48,38 @@ impl VisualizerSystem for EdgesVisualizer {
...
         let timeline_query = LatestAtQuery::new(query.timeline, query.latest_at);
 
+        // TODO(cmc): could we (improve and then) use reflection for this?
+        re_types::static_assert_struct_has_fields!(
+            datatypes::Utf8Pair,
+            first: datatypes::Utf8,
+            second: datatypes::Utf8,
+        );
+        let source: ArchetypeFieldName = "first".into();
+        let target: ArchetypeFieldName = "second".into();
```
teh-cmc, line 58: There are a million possible ways to "improve" this. Whether it's via reflection, or codegen, or whatever else.

For now, I very much like the fact that this is so dumb and so low-tech.

We can make it more complex/automated when the need arises.

grtlr (approved):
Very nice! 🤩

crates/viewer/re_view_graph/src/visualizers/edges.rs
teh-cmc: I need to get rid of the `ArchetypeFieldName` though, that's not needed anymore.

PR #8567, opened 2025-01-02
Add clickable URI component

grtlr:
The component can be clicked in the tooltip and in the selection panel.

image
image


grtlr:
I've updated the UI to show an icon, like we normally do.

crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_narrow/Uri_simple.png
abey79, line 1: this seems incorrect. it should truncate with the `…` normally

abey79 (commented):
Code lgtm, but I'm a bit uncertain about the semantics here. A `Uri` very much feels like a datatype to me. For a component, you'd typically have `SomethingUri`.

I've seen the user request this originated with, and I'm wondering if the following approach would work:

1) make  `Uri` a datatype instead of a component
2) make it so that one can log a custom component with a `Uri` datatype
3) all `Uri`-based component are displayed as links

I'm not sure how bad it would be to have (2) though.

I don't feel super strongly about all this and wouldn't mind having it as is, but let's have @jleibs input first.

jleibs:
I generally agree with @abey79 's thought process, but can you point me at the user request for the additional context.

grtlr:
@jleibs here is the Discord discussion: https://discord.com/channels/1062300748202921994/1320772574271115384/1320772574271115384

emilk:
In the end, I believe we decided to instead auto-recognize urls in strings (e.g. look for `https://` prefix etc) and show those as links. So we should probably close this?

grtlr:
Superseded by #8837.

PR #8573, opened 2025-01-03
Fix tooltips being dragged along in graph view

### LINKED ISSUE #8568: Tooltip moves with the cursor during hovering in graph view
For some reason, the tooltip moves with the cursor during (and after hovering) in the graph view.

This only seems to happen if the cursor is moved in positive coordinate directions (rightwards + downwards):

https://github.com/user-attachments/assets/13027cec-591e-4fe6-8e0f-673a363413ae

It seems like the response of the nodes is handled weirdly when bubbling up selections.

grtlr:
### Related

* Closes #8568.

### What

This fixes the problem by tying to tooltip position to the node position instead of the the pointer, with the nice side-effect that tooltips can now be interacted with in the graph view.

There is some weirdness with tooltips opening for the tooltips, but we should probably tackle that separately.

image

### DIFF
--- a/crates/viewer/re_view_graph/src/ui/draw.rs
+++ b/crates/viewer/re_view_graph/src/ui/draw.rs
@@ -349,7 +349,7 @@ pub fn draw_graph(
                 let instance_path =
                     InstancePath::instance(entity_path.clone(), instance.instance_index);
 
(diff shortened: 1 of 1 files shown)


Wumpf (approved):
👌

PR #8590, opened 2025-01-06
Use the same `arrow_ui` function everywhere

emilk:
### Related

### What
Unify how we show arbitrary arrow data in our UI, and add a test for it.

This is _mostly_ a refactor, but may change behavior slightly.


crates/viewer/re_ui/src/ui_layout.rs
```diff
@@ -0,0 +1,136 @@
+/// Specifies the context in which the UI is used and the constraints it should follow.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum UiLayout {
```
emilk, line 3: This has just moved from `crates/viewer/re_viewer_context/src/component_ui_registry.rs`

crates/viewer/re_ui/tests/snapshots/arrow_ui.png
abey79, line 1: We've so far used proportional font for things like "2 items" and "1.2 MiB of data". Here it looks like monospace. That would be inconsistent with how we've done things so far (except for the pre-existing inconsistencies we've surely already have).

I don't have a super strong opinion about this though.

cc @gavrelina

crates/viewer/re_ui/src/arrow_ui.rs
```diff
@@ -66,14 +75,15 @@ pub(crate) fn arrow_ui(ui: &mut egui::Ui, array: &dyn arrow2::array::Array) {
         // Fallback:
         let bytes = re_format::format_bytes(num_bytes as _);
 
+        // TODO(emilk): pretty-print data type
         let data_type_formatted = format!("{:?}", array.data_type());
 
         if data_type_formatted.len() < 20 {
             // e.g. "4.2 KiB of Float32"
-            ui.label(format!("{bytes} of {data_type_formatted}"));
+            ui_layout.data_label(ui, format!("{bytes} of {data_type_formatted}"));
```
abey79, line 83: ditto

crates/viewer/re_ui/src/arrow_ui.rs
```diff
@@ -66,14 +75,15 @@ pub(crate) fn arrow_ui(ui: &mut egui::Ui, array: &dyn arrow2::array::Array) {
...
         } else {
             // Huge datatype, probably a union horror show
-            ui.label(format!("{bytes} of data"));
+            ui_layout.data_label(ui, format!("{bytes} of data"));
```
abey79, line 86: ditto

crates/viewer/re_ui/src/arrow_ui.rs
```diff
@@ -44,7 +52,8 @@ pub(crate) fn arrow_ui(ui: &mut egui::Ui, array: &dyn arrow2::array::Array) {
                 };
                 return;
             } else {
-                ui.label(format!("{instance_count} items"))
+                ui_layout
+                    .data_label(ui, format!("{instance_count} items"))
```
abey79, line 56: See my other comment. I'm not sure `data_label` is right here.

abey79 (approved):
lgtm but I'm really unsure about the use of monospace font for "summary" text (as opposed to actual data).

Also, make sure existing snapshot test pass—pretty sure some are invalidated, especially `test_all_components_ui.rs`.

crates/viewer/re_ui/tests/snapshots/arrow_ui.png
emilk: That's true; this looks weird… but I just copied what we already do in https://github.com/rerun-io/rerun/blob/8ff83369f84766977f3ef5a5ff2787d28408ddca/crates/viewer/re_component_ui/src/fallback_ui.rs

emilk: Fixed in a new commit!

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

PR #8591, opened 2025-01-06
Specify, test, and fix all broken `AsComponents`<>`ComponentBatch` interactions from blanket impls

teh-cmc:
Commit by commit is probably best.

Turns out cascading blanket impls on `&dyn T` are hard to reason about and don't do what you expect :zany_face::
* A single instance (i.e. `C: Component`) should never be allowed to cast straight to a collection of batches (i.e. `AsComponents`).
  This is the source of all the bugs. We need to enforce that this goes through a `ComponentBatch` first, which in turn will be auto-cast to `AsComponents`.
* Force-casting a simple batch to a collection is now verboten. It still works where it counts though (i.e. `rec.log`).
  This was extremely dangerous:
  ```rust
  let labels = [
      MyLabel("a".into()),
      MyLabel("b".into()),
      MyLabel("c".into()),
  ];
  
  // This does not do what you think it does (you're logging 3 batches of the
  // same component, with one instance each!!!).
  rec.log("labels", &labels as &dyn crate::AsComponents).unwrap();
  ```


teh-cmc:
@rerun-bot full-check

teh-cmc:
@rerun-bot full-check

Wumpf (approved):
oof tough! thanks for unraveling this knot!

PR #8604, opened 2025-01-07
Only override the arrow1 versions of the Loggable trait

emilk:
### Related

### What
All the arrow2-functions on `trait Loggable` now defer to the arrow1 implementations, and all implementations only override the arrow1-trait functions.


crates/store/re_types_core/src/tuid.rs
```diff
@@ -45,25 +45,25 @@ impl Loggable for Tuid {
...
         let validity = None;
 
-        let datatype = arrow2::datatypes::DataType::Extension(
-            Self::NAME.to_owned(),
-            Arc::new(Self::arrow2_datatype()),
-            None,
-        );
-
-        // TODO(cmc): We use the extended type here because we rely on it for formatting.
-        Ok(StructArray::new(datatype, values, validity).boxed())
```
emilk, line 61: @teh-cmc this will surely cause a regression, but I don't know where.

crates/store/re_types_core/src/loggable.rs
```diff
@@ -23,115 +23,113 @@ use crate::{Archetype, ComponentBatch, LoggableBatch};
 /// which makes it possible to work with lists' worth of data in a generic fashion.
 pub trait Loggable: 'static + Send + Sync + Clone + Sized + SizeBytes {
```
emilk, line 24: I moved around the functions to group them better (arrow2 stuff at the end). Sorry.

crates/store/re_types_core/src/tuid.rs
emilk: I can set [`metadata` on the fields](https://docs.rs/arrow/latest/arrow/datatypes/struct.Field.html#method.with_metadata) (which should be equivalent?), but I need some way to test that I'm not breaking anything. I don't know where we rely on this.

emilk: We do show `ChunkId`:s  the chunk store browser, and everything seems fine there:

Screenshot 2025-01-07 at 20 06 29

emilk: So yeah, without know what I'm breaking/fixing 🤷

emilk:
@rerun-bot full-check

crates/store/re_types_core/src/tuid.rs
teh-cmc: Manually adding an `ARROW:extension:name` in the `TransportChunk` glue seems like the right solution yes.

The best way is to test for this is `Command Palette > Print datastore`. Right now your PR doesn't break it (probably because that still goes through arrow2?), so we can see about that later.

teh-cmc (approved):

PR #8605, opened 2025-01-07
Remove superfluous blanket impls in our logging traits

teh-cmc:
Our core logging traits have to be object-safe, which means `Self` must be unsized, which means all methods must take either a shared or exclusive reference.
This in turns means that any `impl Trait for &something` is non-sensical, since this lowers to `impl Trait for &&self`, which just gives extra work to the compiler for no reason since this is semantically identical to the builtin autoderef magic.

### DIFF
--- a/crates/store/re_types_core/src/as_components.rs
+++ b/crates/store/re_types_core/src/as_components.rs
@@ -96,6 +96,11 @@ pub trait AsComponents {
     }
 }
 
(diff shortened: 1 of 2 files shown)


teh-cmc:
@rerun-bot full-check

emilk (approved):
good riddance!

PR #8606, opened 2025-01-07
Enable `re_renderer` for all tests using the `TestContext` if graphics adapter is present

Wumpf:
The viewer test harness now tries to create a wgpu adapter (only once via a `Lazy`) in order to re-use this in an re_renderer context. As a consequence screenshot tests can now use egui elements that bottom out in `re_renderer` drawing (i.e. in `ViewBuilder`).

For the moment the adapter creation is allowed to fail, meaning that if we're on a ci machine without any graphics adapter at all, we can still continue as before with render context == `None`.

* [x] pass main ci
* [x] What's the performance before/after for component tests?

## Performance:

Timing reports for `cargo test -p re_component_ui` (as reported by test runner) on Windows:

* on this branch
  * first run after checkout: 2.34s
  * next two successive rounds: 2.30s, 2.12s
* on main with snapshots enabled (commenting out the still remaining mac-only guard):
  * first run after checkout: 31.86s
  * next two successive rounds: 26.70s, 28.61s
* on main without snapshots:
  * first run after checkout: 0.43s
  * next two successive rounds: 0.42s, 0.44s

## Next steps:
* enable this on for Linux & Windows CI -> copy wgpu's software rasterizer setup
    * make it an error when no adapter can be created?
* make `render_ctx` non option again
* improve setup using new abilities from https://github.com/emilk/egui/pull/5506


Wumpf:
@rerun-bot full-check

Wumpf:
@rerun-bot full-check

Wumpf:
note to self: expecting this to fail since the PR as is requires graphics adapter creation to be always successful and we don't do anything towards that. But I want to see how it fails exactly and then remedy it

Wumpf:
decided to take this more step by step: for the moment this does not enforce the wgpu adapter and by extension the re_renderer context to be present

Wumpf:
@rerun-bot full-check

crates/viewer/re_renderer/src/config.rs
```diff
@@ -296,6 +296,19 @@ impl DeviceCaps {
     }
 }
 
+pub fn instance_descriptor(supported_backends: wgpu::Backends) -> wgpu::InstanceDescriptor {
```
Wumpf, line 299: unused, plz remove

deny.toml
```diff
@@ -45,16 +45,17 @@ deny = [
...
-  { name = "base64" },         # Too popular
-  { name = "cargo_metadata" }, # Older version used by ply-rs. It's small, and it's build-time only!
-  { name = "cfg_aliases" },    # Tiny macro-only crate. wgpu/naga is using an old version
-  { name = "hashbrown" },      # Old version used by polar-rs
-  { name = "memoffset" },      # Small crate
-  { name = "prettyplease" },   # Old version being used by prost
-  { name = "pulldown-cmark" }, # Build-dependency via `ply-rs` (!). TODO(emilk): use a better crate for .ply parsing
-  { name = "redox_syscall" },  # Plenty of versions in the wild
-  { name = "pollster" },       # rfd is still on 0.3
+  { name = "accesskit_consumer" }, # Duplicate as when used as dev dependency - eframe & egui_kittest are on different versions.
```
Wumpf, line 48: this line is new, nothing else changed

crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_wide/Colormap_placeholder.png
emilk, line 1: Random thought: currently we silently opt-out when `ViewerContext::render_ctx` is `None`.
Maybe we should panic instead in debug builds?

That way we can run tests that don't require the render_ctx by setting it to `None`, but will not accidentally forget to set it for tests that do need it (like the colormap). i.e., forgetting to call `setup_kittest_for_rendering` will be a hard error.

crates/viewer/re_viewer_context/src/test_context.rs
```diff
@@ -72,11 +81,122 @@ impl Default for TestContext {
...
+        present_mode: wgpu::PresentMode::Immediate,
+        desired_maximum_frame_latency: None,
+        on_surface_error: Arc::new(|_| {
+            unreachable!("tests aren't expected to draw to surfaces");
+        }),
+    };
+
+    let compatible_surface = None;
+    let depth_format = None;
+    let msaa_samples = 1;
```
emilk, line 114: Don't we have `4` in the real app?

crates/viewer/re_viewer_context/src/test_context.rs
```diff
@@ -72,11 +81,122 @@ impl Default for TestContext {
...
             command_sender,
             command_receiver,
+
+            // Created lazily since each egui_kittest harness needs a new one.
+            egui_render_state: Mutex::new(None),
         }
     }
 }
 
+fn create_egui_renderstate() -> Option<egui_wgpu::RenderState> {
```
emilk, line 91: it would be helpful with a docstring explaining why this could return `None`

emilk (approved):
Lovely!

crates/viewer/re_viewer_context/src/test_context.rs
Wumpf: re_renderer's individual views (managed each by a `ViewBuilder`) have, but egui's final target doesn't - re_Renderer resolves and copies into egui in `ViewBuilder::composite`
leaving a comment about this since this is quite the mental gymnastic!

crates/viewer/re_viewer_context/src/test_context.rs
Wumpf: on its way out, but will do anyways just in case I can't finish my thread of work

crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_wide/Colormap_placeholder.png
Wumpf: we're soooo close to having this be a thing of the past 😄
let's hold on for a bit

Wumpf:
> But I think it would be great if we got a panic if we forget to call setup_kittest_for_rendering when re_renderer is needed

good idea, I look into it!

Wumpf:
Tried a bit and it looks doable.. **if** re_renderer is always created. Luckily the next PR takes care of that anyways!

PR #8607, opened 2025-01-07
Implement copy-screenshot-to-clipboard on Web

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/8264


abey79 (commented):
Seems that this PR broke meshes in Safari/GL. At least that's what I'm seeing with the web build, which I dont see in other PR.

image


**edit**: nvm, it's #8577

abey79 (approved):
code lgtm, but as noted doesn't work on safari

emilk:
I am guessing safari isn't working because there is a frame-delay between the click and the screenshot being captured, and safari doesn't allow copying unless it in a direct response to a user action…

Maybe I'll just remove the button on safari.

abey79:
We've had the same problem back then for copying text and eventually solved it. Any chance we can use the same strategy?

PR #8610, opened 2025-01-07
Fix WSL support, update troubleshooting guide

Wumpf:
## Related

* Fixes https://github.com/rerun-io/rerun/issues/8564
* Some steps towards https://github.com/rerun-io/rerun/issues/8475
* Dependent on https://github.com/emilk/egui/pull/5506
* Dependent on https://github.com/emilk/egui_plot/pull/67

### What's broken

Since Rerun 0.19 we picking in 2D & 3D view was broken & accompanied by log spam.

The reason we failed 0.19+ is unknown but likely either that something caused us picking the "webgl-like" compatibility tier or a change in wgpu no longer allows us to texture copy depth buffers.

Rerun 0.21 then broke startup entirely. This happens due to two distinct problems
* we added a check that foresees the picking issue, aborts application start and tells the user to update their driver (which.. as far as I tried doesn't help)
* wgpu did a validation related thing that that accidentally makes us fail startup because it makes the assumption that compute shader support == glsl support for buffers with runtime size (very very reasonable, but also wrong it turns out :sob: https://github.com/gfx-rs/wgpu/issues/6841)

### This Fix

I decided to move forward more aggressively in a manner that leaves all these intertwined issues behind and sets us up better for simpler & more advanced rendering in the future:
* the troubleshooting guide now explains how to update the mesa drivers
* we configure wgpu now such that it allows picking those non-compliant Vulkan drivers

-> As a result we can draw with Vulkan rather than a (typically) outdated GL driver which is in line with us moving slowly towards WebGPU support as the minspec.
(See code comments for remaining weirdness ;-))

Turns out another benefit with this is that multi-gpu setups behave like on the host now: Previously, `MESA_D3D12_DEFAULT_ADAPTER_NAME` had to be set to pick which adapter is used. Now all adapters are passed through WSL correctly, meaning that our regular adapter selection logic (which is to be further improved) works fine, i.e. it prefers discrete over integrated GPU unless `WGPU_POWER_PREF=low` is set.

Tested on WSL with
* Ubuntu 22.04
* Ubuntu 24.04

previous tests indicate that Ubuntu 20 should behave similar to 22.

### Future work

Currently, there's a little bit of not super straight forward logging left from wgpu, warning about the wild GPU setup, but I actually feel better leaving this in for now - after all the WSL rendering stack is still rather adventurous and we should keep encouraging using the host system directly.
 ```sh
[...]

PR #8612, opened 2025-01-08
Add analytics for wgpu backend and whether the viewer runs in WSL

Wumpf:
Today we're driving blind on some of the key questions around Rerun Viewer usage. This PR adds three new properties to the `viewer_started` analytics event:

* `is_wsl`
    * a boolean indicating whether the viewer has been started directly from the WSL
    * we're interested in this because we want to understand the importance of WSL support 
* `graphics_adapter_backend`
    * a string identifying which of the [wgpu backends](https://docs.rs/wgpu/latest/wgpu/enum.Backend.html) is used
    * we're interested in this because we want to know which graphics backends are the most important
         * The most important distinction here is `gl` (webgl + desktop gl) vs `webgpu` vs everything else. We want to understand when the right time is to phase out support for gl.
* `re_renderer_device_tier`
     * a string identifying which of the (today two) [device tiers](https://github.com/rerun-io/rerun/blob/ddaeb06e9590dc3660c5bef97bafbc824efcc0da/crates/viewer/re_renderer/src/config.rs#L13) re_renderer settled on for a given viewer session
     * similar to the above this gives us a (very) rough idea of what abilities the graphics adapter has
     * it's likely that we'll add one or two more tiers in the future

----------

As always we're trying to be considerate with what we add to our analytics in order to be intrusive while providing valuable information to inform the development of Rerun.
If you want to learn more about analytics run `rerun analytics details`.
If you have a complaint about the intrusiveness of these analytics please reach out.

----------

Examples of what we see on Posthog with this:

Starting viewer on Windows:
image

Starting viewer from WSL (with  #8610 merged in):
image

Starting the viewer from the Browser @ WebGL:
image

Starting the viewer from the Browser @ WebGPU:
image

### DIFF
--- a/crates/top/rerun/src/commands/entrypoint.rs
+++ b/crates/top/rerun/src/commands/entrypoint.rs
@@ -845,8 +845,7 @@ fn run_impl(
                     _build_info,
                     &call_source.app_env(),
                     startup_options,
-                    cc.egui_ctx.clone(),
-                    cc.storage,
+                    cc,
                 );
                 for rx in rxs {
                     app.add_receiver(rx);
--- a/crates/utils/re_analytics/src/event.rs
+++ b/crates/utils/re_analytics/src/event.rs
@@ -54,6 +54,43 @@ pub struct ViewerStarted {
 
     /// The environment in which the viewer is running.
     pub app_env: &'static str,
+
+    /// Sparse information about the runtime environment the viewer is running in.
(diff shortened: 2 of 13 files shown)


Wumpf:
decided to take this more step by step: for the moment this does *not* enforce the wgpu adapter and by extension the re_renderer context to be present

emilk (approved):
nice

PR #8614, opened 2025-01-08
improve ergonomics of encoding and decoding data (using arrow IPC)

### LINKED ISSUE #8356: gRPC: Simplify codec and provide encoding/decoding for ``DataframePart`` directly
### What

* we need to simply get rid of ``NoData`` message and the whole header shenanigans that we have right now
* add ``encode`` and ``decode`` to ``DataframePart`` to avoid the need to explicitly define codec version, but instead "hide" it beneath those helper methods

Related PR: https://github.com/rerun-io/rerun/pull/8328/

zehiko:
### What

Enable more ergonomic encoding and decoding of protobuf types.

* Closes #8356

### DIFF
--- a/crates/store/re_log_encoding/src/codec/wire/decoder.rs
+++ b/crates/store/re_log_encoding/src/codec/wire/decoder.rs
@@ -24,8 +24,7 @@ fn decode(
     }
 }
 
-/// Trait that enables decoding a type from a byte stream,
-/// introduced for more ergonomic dealing with our protobuf ("wire") tpyes.
+/// Decode an object from a its wire (protobuf) representation.
 pub trait Decode {
     fn decode(&self) -> Result<TransportChunk, CodecError>;
 }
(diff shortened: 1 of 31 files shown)


emilk (approved):

crates/store/re_log_encoding/src/codec/wire/decoder.rs
```diff
@@ -21,3 +23,20 @@ pub fn decode(
...
+    fn decode(&self) -> Result<TransportChunk, CodecError> {
+        decode(self.encoder_version(), &self.payload)
+    }
+}
+
+impl Decode for RerunChunk {
+    fn decode(&self) -> Result<TransportChunk, CodecError> {
+        decode(self.encoder_version(), &self.payload)
+    }
+}
```
jleibs, line 42: This still makes sense as an intermediate step, but DataframePart decoding to TransportChunk still feels weird to me as DataframePart's are not actually valid Chunks, even though TransportChunk can (conveniently) store them.

I think this should decode to an alternative parallel Dataframe-oriented structure that looks very similar but is typed distinctly so nobody is inclined to try to create a Chunk out of it (which may fail due to missing metadata, etc.)

zehiko: makes sense, I was also slightly reluctant to have 2 very distinct data structures decode to TC as they indeed have very different semantics. Do you think we'll have a need for 2 different ones once we have ``RecordBatch`` on the wire?

PR #8616, opened 2025-01-08
Use the correct eframe options for our custom viewer examples

emilk:
The problem was that we ignored our default decoration settings, leading to an ugly top bar on mac:

image

This PR fixes this.


Wumpf (approved):
good catch!

PR #8618, opened 2025-01-08
Update arrow, pyo3, and numpy

emilk:


jleibs:
> There were a few scary changes to `rerun_py/src/dataframe.rs` - do we have automated testing for that?

Not totally comprehensive, but there's partial coverage in: https://github.com/rerun-io/rerun/blob/71215dbc5e75ab476077ad203350da4ef7a2d793/rerun_py/tests/unit/test_dataframe.py#L4

PR #8620, opened 2025-01-08
Run Linux & Windows screenshot tests via `lavapipe` software rasterizer

### LINKED ISSUE #8245: Use a software renderer for kittest

Wumpf:
### Related

* Fixes #8245 

### What

Enables screenshot tests on Linux & Windows ci using Mesa lavapipe (the Vulkan version of llvmpipe) software rasterizer.
Uses wgpu's Mesa build https://github.com/gfx-rs/ci-build (thanks @cwfitzgerald!)

ICD driver file discovery & how to best download the Vulkan SDK was giving me a bit of a headache (via apt as done on wgpu's ci was very slow for some reason and wgpu only does GL software rasterizing on Windows). But it's all figured out now & documented in the setup script :). The process can be replicated locally the same way, but has a bit of "you better know what you're doing" to it as the driver setup is a little bit invasive to account for elevated user access ([...] see comments).

Changes to our CI pipelines overall:
* PR CI: still runs Rust tests only on Linux now, but now only `--all-targets --all-features` instead of both that and with just `--all-targets`
* Main CI: Runs now `cargo test --all-targets --all-features` not only on Mac but also on Windows
* Nightly CI: Runs now `pixi run rs-check` not only on Linux but also on Windows & Mac

----

* [x] passes Main CI

### DIFF
--- a/.github/workflows/contrib_checks.yml
+++ b/.github/workflows/contrib_checks.yml
@@ -97,6 +97,9 @@ jobs:
         with:
           pixi-version: v0.39.0
 
+      - name: Setup software rasterizer
+        run: pixi run python ./scripts/ci/setup_software_rasterizer.py
+
       - name: Rust checks & tests
         run: pixi run rs-check --skip individual_crates docs_slow
 
--- a/.github/workflows/contrib_checks.yml
+++ b/.github/workflows/contrib_checks.yml
@@ -31,6 +31,9 @@ env:
   # these incremental artifacts when running on CI.
   CARGO_INCREMENTAL: "0"
 
+  # Sourced from https://vulkan.lunarg.com/sdk/home#linux
+  VULKAN_SDK_VERSION: "1.3.296.0"
+
 defaults:
   run:
     shell: bash
@@ -96,6 +99,16 @@ jobs:
       - uses: prefix-dev/setup-pixi@v0.8.1
         with:
           pixi-version: v0.39.0
+
+      # Install the Vulkan SDK, so we can use the software rasterizer.
+      # TODO(andreas): It would be nice if `setup_software_rasterizer.py` could do that for us as well (note though that this action here is very fast when cached!)
+      - name: Install Vulkan SDK
+        uses: jakoch/install-vulkan-sdk-action@v1.0.5
+        with:
+          vulkan_version: ${{ env.VULKAN_SDK_VERSION }}
+          install_runtime: true
+          cache: false
+          stripdown: true
 
       - name: Setup software rasterizer
         run: pixi run python ./scripts/ci/setup_software_rasterizer.py
--- a/.github/workflows/reusable_checks_rust.yml
+++ b/.github/workflows/reusable_checks_rust.yml
@@ -36,6 +36,11 @@ env:
   # Improve diagnostics for crashes.
   RUST_BACKTRACE: full
 
+  # Sourced from https://vulkan.lunarg.com/sdk/home#linux
+  #
+  # We don't include the 4th version number, as it's not used in any URL.
+  VULKAN_SDK_VERSION: "1.3.290"
+
 defaults:
   run:
     shell: bash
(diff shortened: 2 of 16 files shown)


Wumpf:
@rerun-bot full-check

Wumpf:
@rerun-bot full-check

Wumpf:
@rerun-bot full-check

.github/workflows/contrib_checks.yml
```diff
@@ -97,6 +100,19 @@ jobs:
         with:
           pixi-version: v0.39.0
 
+      # Install the Vulkan SDK, so we can use the software rasterizer.
+      # TODO(andreas): It would be nice if `setup_software_rasterizer.py` could do that for us as well (note though that this action here is very fast when cached!)
+      - name: Install Vulkan SDK
```
emilk, line 105: Once this PR is merged, we should test this path with a new "external" PR

.github/workflows/reusable_checks_rust.yml
```diff
@@ -109,8 +136,28 @@ jobs:
...
+        run: pixi run python ./scripts/ci/setup_software_rasterizer.py
+
       - name: Download test assets
         run: pixi run python ./tests/assets/download_test_assets.py
 
-      - name: pixi run cargo test --all-targets --all-features
+      - name: Run tests (`cargo test --all-targets --all-features`)
+        if: ${{ inputs.CHANNEL != 'nightly' }}
+        # Need to use pixi due to NASM dependency.
         run: pixi run cargo test --all-targets --all-features
```
emilk, line 159: Why run all the tests on all the three platforms for each PR?
We could save some money by only running them on Linux, and running Mac/Windows only on `main` and `nightly` builds.

crates/viewer/re_viewer_context/src/test_context.rs
```diff
@@ -156,54 +158,93 @@ struct SharedWgpuResources {
...
-    // TODO(#8245): Should we require this to succeed?
+    // We rely a lot on logging in the viewer to identify issues.
+    // Make sure logging is set up if it hasn't been done yet.
+    //let _ = env_logger::builder().is_test(true).try_init();
+    let _ = env_logger::builder()
+        .filter_level(re_log::external::log::LevelFilter::Trace)
+        .is_test(false)
+        .try_init();
+
+    // We don't test on GL & DX12 right now (and don't want to do so by mistake!).
```
emilk, line 178: Don't explain _what_ the code does, explain _why_ it does it

crates/viewer/re_viewer_context/src/test_context.rs
```diff
@@ -266,17 +302,30 @@ impl TestContext {
...
+        // If re_renderer was used, `setup_kittest_for_rendering` should have been called.
+        // (if not, then we clearly didn't render anything since that requires writing to a texture!).
+        if render_ctx
+            .active_frame
+            .num_view_builders_created
+            .load(std::sync::atomic::Ordering::Acquire)
+            != 0
+            && !self
+                .called_setup_kittest_for_rendering
+                .load(std::sync::atomic::Ordering::Relaxed)
```
emilk, line 322: This is very hard to read.
```suggestion
        let num_view_builders_created =
            render_ctx.active_frame.num_view_builders_created();
        let called_setup_kittest_for_rendering =
            self.called_setup_kittest_for_rendering.load(Ordering::Relaxed);
        if num_view_builders_created != 0 && !called_setup_kittest_for_rendering
```

emilk: Tip: RustAnalyzer has helpers for extracting a selected expression into a named variable

emilk (approved):
Awesome! Finally Clement can start contributing to the UI :)

.github/workflows/reusable_checks_rust.yml
Wumpf: `pr` only runs `linux` ci, this is filtered in the line above here first
https://github.com/rerun-io/rerun/pull/8620/files#diff-f8fa90b87fca261d70633935468d767dfaefd6506376ab4d743c835404ad86d7R119
So exactly as you suggest :)

tbh I went with `if: ${{ inputs.CHANNEL != 'nightly' }}` because that made it easier to commen tout the filter above for testing purposes. I'll better repeat the exact same rule again to avoid confusion!

PR #8621, opened 2025-01-08
Fix .gitignore ignoring `crates/build`

jprochazk:
The `build/` and `./build/` seemed to both be matching the wrong things. Changed it to `/build/`:

```
$ git check-ignore -v -- ./build/debug
.gitignore:9:/build/    ./build/debug
$ git check-ignore -v -- ./crates/build
<empty>
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,8 +6,7 @@ __MACOSX
 *.bin
 *.o
 /arrow/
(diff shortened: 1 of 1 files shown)


emilk (approved):
nice catch

Issue #8586, opened 2025-01-06
Project: filtering entities in the UI (first iteration)


abey79:
## Summary description

Implement the first iteration of a text-based entity path filtering for the streams, blueprint and add/remove entities trees. (The operative words being "first iteration".)

- Fixes https://github.com/rerun-io/rerun/issues/4597

## Background and grounding use-cases

The primary (and only means) of finding entities in (large) recordings currently consists of scrolling through the list. This is obviously cumbersome and annoying as soon as the list of entities grows. This is compounded by the streams tree not being tall in the UI by construction (a conundrum that proved tricky to resolve so far).

This project is thus about making it easy to find the needle(s) in the entity haystack, in a way that is local to the following areas:
- streams tree
- blueprint tree
- add/remove entity modal

## Impact on Entity Path Filter

Although this project (and the general feature of filter-based searching in the UI) is technically independent of the entity path filter stuff (at least initially), it is related in the sense that both are about filtering the list of entities. For consistency, expanding the feature set on one side should typically be mirrored on the other side, lest users get confused. This is particularly true for any syntax introduced to match entities (glob, etc.).

In this spirit of keeping this (design) complexity in check, this project is narrowly scoped on this front.

## Designs and plans

[Figma designs](https://www.figma.com/design/72AuvfbssFJeqqnQoJscYT/Katya's-playground?node-id=4622-4187&t=oOl0CLIrys32gLML-1)

The primary paradigm of this first iteration is to introduce _text field_ that are _local_ to the list to be filtered.

#### Search widget

Search icon -> click -> search widget appears and "replaces" section title:

![Image](https://github.com/user-attachments/assets/2262d9db-8d06-43d1-9acf-9d7c265890e5)

![Image](https://github.com/user-attachments/assets/983f6ed6-0045-44b2-be14-19f76dcd9fdd)

This applies to both streams and blueprint trees.

**Note**: for the blueprint tree, this implies moving the reset and + button to a `...` more menu, to leave room for the filter widget:

![Image](https://github.com/user-attachments/assets/1064c4b1-abd8-4a7f-952c-96246ba1f413)

#### Add/remove modal

Search widget is permanently displayed (no toggling behaviour):

![Image](https://github.com/user-attachments/assets/f311f6cf-6563-4f2a-b712-12e62f608d8e)

#### Highlight

Query is highlighted in the displayed entity:

[...]

PR #8625, opened 2025-01-09
Add the ability to provide local data to context menu actions

abey79:
### Related
 
* Part of #8586

### What

This PR adds the possibility to provide local data to context menu action from "user" code (e.g. the place where the context menu is triggered). The local data is structured as a `TypeMap`.

This new ability is used to fix the behaviour of Collapse/Expand all in the blueprint streams tree (which never worked before). It also unblocks the possibility to have a different collapsing start for the filtered and unfiltered states of a given tree (#8586).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5782,6 +5782,7 @@ dependencies = [
  "re_viewer_context",
  "re_viewport_blueprint",
  "static_assertions",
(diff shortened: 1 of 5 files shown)


abey79:
We have a better plan, on it.

PR #8627, opened 2025-01-09
Basic JWT-based authentication and authorization

grtlr:
### What

This adds a new `re_auth` crate with the following features:

* JWT-based symmetric authorization and authentication (basic read/write modes).
* Rerun-ified abstraction over `jwt-simple` so that we can swap it out.
* Helpers to convert a `SecretKey` from/to `base64` to be used with `redap-cli`.
* `tonic::Interceptor`s for both client and server side middleware with an `authorization: Bearer <token>` header.

Here is what a `SecretKey` (`HS256`) looks like in `base64`:

```
pBiQ9NVDj1elVjATgyL5EYri/9paHwvz78lsx7QCq9E=
```

We can use that to generate a basic token:

```
eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJyZWRhcC1jbGkiLCJzdWIiOiJqb2NoZW5AcmVydW4uaW8iLCJhdWQiOiJyZWRhcCIsImV4cCI6MTczNjg3MjUzNCwiaWF0IjoxNzM2ODQ0NTM0fQ.QHI99cXIi4VjhZKBJmLb13NynOjvmUWuxy63CUwiBlA
```

Which you can verify yourself via [www.jwt.io](https://jwt.io/#debugger-io?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE3MzY3ODYwMjIsImV4cCI6MTczNjgxNDAyMiwibmJmIjoxNzM2Nzg2MDIyLCJpc3MiOiJyZWRhcC1jbGkiLCJzdWIiOiJqb2NoZW5AcmVydW4uaW8iLCJhdWQiOiJyZWRhcCJ9.qiClC8YhTszvYFKk6HgL-Gp9W_lDmfiZGyYB43O3Z58).

image


crates/store/re_auth/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use base64::{engine::general_purpose, Engine as _};
+use jwt_simple::{
+    claims::{Claims, JWTClaims},
+    common::VerificationOptions,
+    prelude::{HS256Key, MACLike as _},
+    token::Token,
+};
+
+/// A common secret that is shared between the client and the server.
```
jleibs, line 19:
```suggestion
/// A secret key that is used to generate and verify tokens.
```

zehiko (approved):
I don't have all the context on what was discussed, but this is really easy to follow code and having a notion of "provider" that can later be switched with some other components makes sense to me.

grtlr:
@jleibs it would be awesome if you could also have another look. 🙏

jleibs (approved):

PR #8628, opened 2025-01-09
Port `generic_placeholder_for_datatype` to `arrow-rs`

emilk:
### Related
* Part of #3741 

`cargo test --all-features` passes


Wumpf (approved):
nice! once somebody figured out all the magic invocation this looks so straight forward 😄 
Great to see fixed size arrays placeholders being much better now

PR #8633, opened 2025-01-09
Make `ViewerContext::render_ctx` non-optional

Wumpf:
### Related

* follow up to https://github.com/rerun-io/rerun/pull/8620

### What

Now that tests require a wgpu device to be available, we can go back to making `re_renderer` non-optional, removing lots of annoying checks & early-outs!


crates/viewer/re_view_spatial/src/picking_ui.rs
```diff
@@ -41,10 +41,6 @@ pub fn picking(
         return Ok(response);
     }
 
-    let Some(render_ctx) = ctx.render_ctx else {
-        return Err(ViewSystemExecutionError::NoRenderContextError);
```
emilk, line 45: Is this error type removed now?

emilk (approved):
That was a lot more than I thought 😬  Good riddance!

crates/viewer/re_view_spatial/src/picking_ui.rs
Wumpf: ah yes, missed that, thx!

PR #8636, opened 2025-01-10
Extend (and rename) `ItemSpaceContext` to provide local context to context menu action

abey79:
### Related

* Part of 8586
* Supersedes #8625

### ~Context~ Background

The lingua franca that allows `re_context_menu` to be both useful and low in the crate hierarchy (so it may be used by various UI areas) is the `re_viewer_context::Item`. This generally makes sense. `Item` are the things that can be hovered, selected, interacted with, so, in most case, there exists a meaningful mapping between an `Item` and actions that applie to it.

This breaks down when the action is not about the `Item` _per se_, but about how that `Item` is represented _in this specific UI area_. The `Expand/Collapse all…` actions fall in this category (e.g. they are about collapsing the item representation _in this specific UI area_, not all of them).

In order to be correctly implemented, the context menu action needs to have additional context for the item (namely, where it was clicked/selected). This is why expand/collapse all is currently broken in the blueprint streams tree.

Because we support right-clicking on a multi-selections that includes items that might have been selected in various places across the UI, the only place where that local context is next to the `Item`, in the `ItemCollection` representing the selection. As  it turns out, we already have such a thing. It's called `ItemSpaceContext` and is used to handle the specific interaction between the 2D and 3D views (rays, etc.) when interacting with some of the entities.

### What

This PR extends the `ItemSpaceContext` to include the case where the entity is picked from the streams tree, with information whether this is the blueprint or regular streams tree. This fixes the aforementioned bug. It also rename `ItemSpaceContext` to `ItemContext`.

**Review commit-by-commit**: first commit is the meat, second is pure renames.

### Future work

- The implementation of #8586 will build on this PR and add some more data to handle the interaction of filtering and collapsing state.
- In the long run, we will probably give some love to `ItemContext` to make it more flexible. Maybe it should be turned into a `TypeMap` (which, ironically, was at the core of my 1st failed attempt at this #8625).

### DIFF
--- a/crates/viewer/re_context_menu/src/actions/collapse_expand_all.rs
+++ b/crates/viewer/re_context_menu/src/actions/collapse_expand_all.rs
@@ -1,6 +1,6 @@
 use re_entity_db::InstancePath;
 use re_log_types::StoreKind;
-use re_viewer_context::{CollapseScope, ContainerId, Contents, Item, ItemSpaceContext, ViewId};
+use re_viewer_context::{CollapseScope, ContainerId, Contents, Item, ItemContext, ViewId};
 
 use crate::{ContextMenuAction, ContextMenuContext};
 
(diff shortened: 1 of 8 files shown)


Wumpf (approved):
well that wasn't too bad after all amiright 😁

abey79:
> well that wasn't too bad after all amiright 😁

The difficult PR was the one against our mental model 😅

PR #8638, opened 2025-01-10
Port `TimeColumn` to arrow-rs

emilk:
### Related
* #3741

### DIFF
--- a/crates/store/re_chunk/src/chunk.rs
+++ b/crates/store/re_chunk/src/chunk.rs
@@ -1111,8 +1111,9 @@ impl TimeColumn {
         )
     }
 
-    /// Parse the given [`ArrowArray`] as a time column,
-    /// returning its contents and its datatype (e.g. [`ArrowDataType::Timestamp`]).
+    /// Parse the given [`ArrowArray`] as a time column.
+    ///
+    /// Results in an error if the array is of the wrong datatype, or if it contains nulls.
     pub fn read_array(array: &dyn ArrowArray) -> Result<ArrowScalarBuffer<i64>, TimeColumnError> {
         #![allow(clippy::manual_map)]
 
--- a/crates/store/re_chunk/src/chunk.rs
+++ b/crates/store/re_chunk/src/chunk.rs
@@ -1,7 +1,11 @@
 use std::sync::atomic::{AtomicU64, Ordering};
 
 use ahash::HashMap;
-use arrow::array::{Array as _, Int64Array as ArrowInt64Array, ListArray as ArrowListArray};
+use arrow::{
+    array::{Array as ArrowArray, ArrayRef as ArrowArrayRef, ListArray as ArrowListArray},
+    buffer::ScalarBuffer as ArrowScalarBuffer,
+    datatypes::DataType as ArrowDataType,
+};
 use arrow2::{
     array::{
         Array as Arrow2Array, ListArray as Arrow2ListArray, PrimitiveArray as Arrow2PrimitiveArray,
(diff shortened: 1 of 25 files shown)


crates/store/re_chunk/src/chunk.rs
```diff
@@ -1603,24 +1650,13 @@ impl TimeColumn {
...
-        if *times.data_type() != timeline.datatype() {
-            return Err(ChunkError::Malformed {
-                reason: format!(
-                    "Time data for timeline {} has the wrong datatype: expected {:?} but got {:?} instead",
-                    timeline.name(),
-                    timeline.datatype(),
-                    *times.data_type(),
-                ),
-            });
-        }
```
emilk, line 1621: No longer needed

crates/store/re_chunk/src/merge.rs
```diff
@@ -281,17 +281,20 @@ impl TimeColumn {
...
-            .as_any()
-            .downcast_ref::<Arrow2PrimitiveArray<i64>>()?
-            .clone();
+        let times = self
+            .times_raw()
+            .iter()
+            .chain(rhs.times_raw())
+            .copied()
+            .collect_vec();
+        let times = ArrowScalarBuffer::from(times);
```
emilk, line 297: Using Rust to do the concat instead of using a computation kernel.

crates/store/re_chunk_store/tests/snapshots/memory_test__scalars_on_one_timeline_new.snap
```diff
@@ -1,9 +1,11 @@
...
 source: crates/store/re_chunk_store/tests/memory_test.rs
-expression: "[format!(\"{NUM_SCALARS} scalars\"),\n        format!(\"{} in total\",\n            re_format::format_bytes(total_mem_use_global as _)),\n        format!(\"{} per row\",\n            re_format::format_bytes(total_mem_use_global as f64 / NUM_SCALARS\n            as f64))]"
+assertion_line: 133
+expression: "[format!(\"{NUM_SCALARS} scalars\"),\nformat!(\"{} in total\", re_format::format_bytes(total_mem_use_global as _)),\nformat!(\"{} per row\",\nre_format::format_bytes(total_mem_use_global as f64 / NUM_SCALARS as f64)),]"
+snapshot_kind: text
 ---
 [
     "1048576 scalars",
-    "37.1 MiB in total",
+    "37.0 MiB in total",
```
emilk, line 9: 🥳

crates/utils/re_byte_size/src/arrow_sizes.rs
```diff
@@ -15,3 +19,17 @@ impl SizeBytes for ArrayRef {
...
+    #[inline]
+    fn heap_size_bytes(&self) -> u64 {
+        Array::get_array_memory_size(self) as u64
+    }
+}
+
+impl<T: ArrowNativeType> SizeBytes for ScalarBuffer<T> {
+    #[inline]
+    fn heap_size_bytes(&self) -> u64 {
+        self.inner().capacity() as _
```
emilk, line 33: We're a bit inconsistent wether or not we measure used memory (`.len()`) or allocated memory (`.capcity()`). I think we should always do the latter. WDYT @teh-cmc ?

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

crates/store/re_chunk/src/chunk.rs
```diff
@@ -795,6 +802,16 @@ pub struct TimeColumn {
     pub(crate) time_range: ResolvedTimeRange,
 }
 
+/// Errors when deserializing/parsing/reading a column of time data.
+#[derive(Debug, thiserror::Error)]
+pub enum TimeColumnError {
+    #[error("Time columns had nulls, but should be dense")]
+    Nulls,
```
jleibs, line 809: Nit: `ContainsNulls` would be slightly more descriptive

crates/store/re_chunk/src/merge.rs
jleibs: Any particular reason? Maybe worth a comment and/or TODO?

jleibs (approved):

crates/store/re_chunk/src/merge.rs
emilk: To use the computational kernel I would have to convert the buffers into dynamic array, and then dynamic-cast them back again. A lot of work, but little to gain

jleibs: Ahh, I forgot the compute functions like concat don't support generics. Annoying.

Issue #7245, opened 2024-08-20
Eager/early serialize of components to arrow in Rust and C++


Wumpf:
As we'll soon introduce [tagged components](https://github.com/rerun-io/rerun/issues/6889) and simple multi-datatype components, it gets harder and harder to represent Archetypes (and concrete ComponentBatches) as collection of concrete types.
Let's take the example of a generalized `rotation` component/archetype field which may be represented by various datatypes: we no longer can store concrete types on an archetype and have to type-earse them right away instead.
Note that this way C++ and Rust get much closer to the Python SDK in this regard.

This fits very well into our desire to get rid of concrete component types in the SDK languages which today almost always take the form of `struct ComponentType(pub datatypes::TheDataType)` together with myriad of constructors, trait impls and utilities. I.e. **a lot** of forwarding code.
Eager serialization allows us to implement component semantics on archetypes instead with concrete construction methods. E.g. `with_quaternion` and `with_axis_angle` would both populate the multi-datatype `rotation` component which gets tagged appropriately.
When logging raw component batches/columns this would become more explicit as you're expected to supply a datatype array/collection together with the appropriate component tag (which will still be provided by the SDK, but more in registry fashion rather a `class`/`struct` per component). This follows the exact same mechanism of how an archetype construct its internal `ComponentBatches`.

A drawback of this approach is that most accesses of archetypes requires deserialization back into the source datatypes which can be cumbersome in some cases. However, this is what we expect to do when a user reads back data from the store, so this is something that may soon become common-place anyways.

Another nice side effect is that the "ephemeral `rerun::Collection` hazard" goes away as we'd no longer store pointers to user data, making the API a lot safer to use. (`rerun::Collection` becomes a pure pass-through type as it should be)
* https://github.com/rerun-io/rerun/issues/7081

---

This ticket is a meetup discussion outcome of @jleibs and @Wumpf with some additional input by @emilk 

## Advantages
* We can support "component generics" (i.e. radii that is either `f32` or `f64`(see https://github.com/rerun-io/rerun/blob/main/design/component_datatypes.md)
* C++ and Python can easily call into Rust to process the contents of a component or archetype (arrow is FFI-friendly)

## Related
* https://github.com/rerun-io/rerun/issues/7204
* https://github.com/rerun-io/rerun/issues/3381
* https://github.com/rerun-io/rerun/issues/1882
* https://github.com/rerun-io/rerun/blob/main/design/component_datatypes.md

---

[...]

PR #8642, opened 2025-01-10
New types and traits for (co-existing!) eager serialization

teh-cmc:
This introduces `SerializedComponentBatch`, which will become the main type we use to carry user data around internally.
```rust
/// The serialized contents of a [`ComponentBatch`] with associated [`ComponentDescriptor`].
///
/// This is what gets logged into Rerun:
/// * See [`ComponentBatch`] to easily serialize component data.
/// * See [`AsComponents`] for logging serialized data.
///
/// [`AsComponents`]: [crate::AsComponents]
#[derive(Debug, Clone)]
pub struct SerializedComponentBatch {
    pub array: arrow::array::ArrayRef,

    // TODO(cmc): Maybe Cow<> this one if it grows bigger. Or intern descriptors altogether, most likely.
    pub descriptor: ComponentDescriptor,
}
```

The goal is to keep the `ComponentBatch` trait isolated at the edge, where it is used as a means of easily converting any data into arrow arrays, instead of simultaneously being used as a means of transporting data around through the internals.
`ComponentBatch` is here to stay, if only for its conversion capabilities.

This opens a lot of opportunities of improvements in terms of DX, UX and future features (e.g. generics).

The two code paths will co-exist for the foreseeable future, until all archetypes have been made eager.

### DIFF
--- a/crates/store/re_types_core/src/as_components.rs
+++ b/crates/store/re_types_core/src/as_components.rs
@@ -1,6 +1,5 @@
 use crate::{
-    ComponentBatch, ComponentBatchCowWithDescriptor, LoggableBatch as _, ResultExt as _,
-    SerializationResult,
+    ComponentBatch, ComponentBatchCowWithDescriptor, SerializationResult, SerializedComponentBatch,
 };
 
 /// Describes the interface for interpreting an object as a bundle of [`Component`]s.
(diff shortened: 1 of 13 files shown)


crates/store/re_types_core/src/as_components.rs
```diff
@@ -34,7 +33,37 @@ pub trait AsComponents {
     //
     // NOTE: Don't bother returning a CoW here: we need to dynamically discard optional components
     // depending on their presence (or lack thereof) at runtime anyway.
-    fn as_component_batches(&self) -> Vec<ComponentBatchCowWithDescriptor<'_>>;
+    #[deprecated(since = "0.22.0", note = "use as_serialized_batches instead")]
```
emilk, line 36: It's worth to also change the docstring to `DEPRECATED - DO NOT USE` or similar.

crates/store/re_types_core/src/loggable_batch.rs
```diff
@@ -66,6 +72,68 @@ pub trait ComponentBatch: LoggableBatch {
             .with_descriptor_override(descriptor)
     }
 
+    /// Serializes the contents of this [`ComponentBatch`].
+    ///
+    /// Once serialized, the data is ready to be logged into Rerun via the [`AsComponents`] trait.
+    ///
+    /// # Fallibility
+    ///
+    /// There are very few ways in which serialization can fail, all of which are very rare to hit
```
emilk, line 81: Still worth mentioning a few of the ways it can fail (e.g. overflowing an `i32`)

emilk (approved):

PR #8645, opened 2025-01-10
Filter entities in the UI (part 0): Make `CustomContent` more useful

abey79:
### Related

- Part of #8586
- Part of a series of PR:
  - #8645
  - #8652
  - ...

### What

This PR makes `CustomContent` for list items more useful, because I'll needed for the UI of the filter widget. I also cleans up a few things and updates the related snapshot test.

### DIFF
--- a/crates/viewer/re_ui/src/list_item/custom_content.rs
+++ b/crates/viewer/re_ui/src/list_item/custom_content.rs
@@ -33,7 +33,7 @@ pub struct CustomContent<'a> {
 impl<'a> CustomContent<'a> {
     /// Create a content with a custom UI closure.
     ///
-    /// The closure will be called from withing a [`egui::Ui`] with its maximum width set as per the
+    /// The closure will be called from within a [`egui::Ui`] with its maximum width set as per the
     /// list item geometry. Note that this may differ from [`crate::ContentContext::rect`] if a
     /// button is set.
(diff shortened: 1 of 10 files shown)


crates/viewer/re_ui/src/list_item/custom_content.rs
```diff
@@ -0,0 +1,179 @@
...
+        self,
+        icon: &'static crate::icons::Icon,
+        add_contents: impl FnOnce(&mut egui::Ui) + 'a,
+    ) -> Self {
+        self.button(super::ItemMenuButton::new(icon, add_contents))
+    }
+}
+
+impl ListItemContent for CustomContent<'_> {
+    fn ui(self: Box<Self>, ui: &mut egui::Ui, context: &ContentContext<'_>) {
```
Wumpf, line 117: having a bit of a deja vu - isn't this button handling code already somewhere else? (can't tell without digging if that's actually the case and whether that's even a concern)

Wumpf (approved):
neat!
makes me so happy that this is right away handled by a screenshot test

crates/viewer/re_ui/src/list_item/custom_content.rs
abey79: Yes, I already have some kind of abstraction for button thingies (could be menu, could be action). Look for the `ItemButton` trait. I'm using that in `PropertyContent`, so I'm piggy backing on it.

PR #8648, opened 2025-01-10
Prevent flaky roundtrips in `rerun rrd compare`

teh-cmc:
Store-side compaction has election heuristics that can yield different results depending on the order in which the chunks arrive.
This can make some cross-language roundtrip tests flaky.

This PR makes sure to disabled store-side compaction when running RRD comparison tests.

* DNM: requires #8647


Wumpf (approved):
I'm rather surprised that compaction isn't deterministic. Is the problem that two rrd files with  differing message order should still yield the same store from the pov of a comparision, but the fact that data comes in different order changes compaction?

Issue #8583, opened 2025-01-06
New partial updates logging APIs: C++


teh-cmc:
TODO

Wumpf:
we got all those in by now!

Issue #8582, opened 2025-01-06
New partial updates logging APIs: Python


teh-cmc:
Something like this:
```py
rr.set_time_sequence("frame", 42)
rr.log("points", rr.Points3D([[0, 0, 0], [1, 1, 1]], colors=[255, 0, 0]))

rr.set_time_sequence("frame", 43)
rr.log("points", [rr.Points3D.components(colors=[0, 0, 255])]) # gets properly tagged
```

Issue #8581, opened 2025-01-06
New partial updates logging APIs: Rust


teh-cmc:
TODO

teh-cmc:
All done.

PR #8649, opened 2025-01-10
Add legacy and next-gen partial updates snippets for all languages

teh-cmc:
This adds a partial update example for `Points3D`, in all languages, using both the legacy and new APIs (if they already exist).

This acts as a before/after comparison, a piece of documentation and a way to make sure that the legacy and new APIs behave similarly.

* DNM: requires #8648 
* Part of #8583 
* Part of #8581 
* Part of #8582


emilk (approved):

zehiko (approved):
nice!

PR #8651, opened 2025-01-10
gRPC spec update: define recording schema protobuf types and expose the schema via new endpoint

zehiko:
### What

Define column descriptors in the protobuf spec and the relevant type conversions. Add new endpoint for fetching the schema that returns a list of column descriptors. 

Also updated the example script to use the next endpoint and print the schema.


crates/store/re_chunk_store/src/protobuf_conversions.rs
```diff
@@ -257,6 +258,112 @@ impl From<crate::QueryExpression> for re_protos::common::v0::Query {
...
+    fn try_from(value: crate::ColumnDescriptor) -> Result<Self, Self::Error> {
+        match value {
+            crate::ColumnDescriptor::Time(time_descriptor) => Ok(Self {
+                descriptor_type: Some(
+                    re_protos::common::v0::column_descriptor::DescriptorType::TimeColumn(
+                        re_protos::common::v0::TimeColumnDescriptor {
+                            timeline: Some(re_protos::common::v0::Timeline {
+                                name: time_descriptor.timeline.name().to_string(),
+                            }),
+                            datatype: serde_json::to_string(&time_descriptor.datatype).map_err(
```
zehiko, line 273: usage of ``serde`` and ``serde_json`` goes away as soon as we migrate to arrow-rs as arrow-rs already supported conversion from/to string for the datatype.

crates/store/re_protos/proto/rerun/v0/common.proto
```diff
@@ -102,6 +102,42 @@ message Query {
...
+
+message TimeColumnDescriptor {
+    /// The timeline this column is associated with.
+    Timeline timeline = 1;
+    /// The Arrow datatype of the column.
+    string datatype = 2;
+}
+
+/// Describes a data/component column, such as `Position3D`.
+message ComponentColumnDescriptor {
```
jleibs, line 120: This makes sense from the perspective of the current query APIs, though I think we probably have lurking issues from potentially losing other metadata.

For example, if a user provides their own data in arrow form with metadata set, we would ideally be preserving it here.

We can probably just capture this as a TODO for now -- I don't think it's possible to pass that metadata through our log APIs yet -- but I think at a minimum this should eventually have a repeated string-string mapping to carry along extra metadata.

That said, once we get to that point I wonder if we may as well just pass all this info other than type) through that metadata (in the same way we encode it for arrow-IPC). That would then collapse this whole thing down to a generic `ArrowField`, getting rid of the differentiation between TimeColumn and ComponentColumn, and also allowing us to use it for the schema of other (non-recording) tables.

crates/store/re_chunk_store/src/protobuf_conversions.rs
jleibs: Nice! I hadn't realized arrow-rs had round-tripable string representation for datatypes.

That said, part of me wonders if we want to be using the serialized flatbuffer binary representation for efficiency reasons. Curious how bad the overhead is for a recording with thousands of entities.

jleibs (approved):

crates/store/re_chunk_store/src/protobuf_conversions.rs
zehiko: > That said, part of me wonders if we want to be using the serialized flatbuffer binary representation for efficiency reasons. Curious how bad the overhead is for a recording with thousands of entities.

you mean such representation of the entire schema or just the ``datatype`` field?  I kinda like the fact we can get back what client needs (``ColumnDescriptor``s) and the integration of the API was really easy in the existing code, so that's why I went along with exposing repeated ``ColumnDescriptor``s rather than arrow IPC encoded schema.

crates/store/re_protos/proto/rerun/v0/common.proto
zehiko: > We can probably just capture this as a TODO for now -- I don't think it's possible to pass that metadata through our log APIs yet -- but I think at a minimum this should eventually have a repeated string-string mapping to carry along extra metadata.

This makes sense!

> That said, once we get to that point I wonder if we may as well just pass all this info other than type) through that metadata (in the same way we encode it for arrow-IPC). That would then collapse this whole thing down to a generic ArrowField, getting rid of the differentiation between TimeColumn and ComponentColumn, and also allowing us to use it for the schema of other (non-recording) tables.

It's relatively straithforward to switch this API to basically return back an arrow IPC encoded ``Schema`` if you think that's where we'll eventually end up. I was going back and forth on it, but mostly the easy integration and usage on the client side (as mentioned in comment above) made me opt for defining these descriptors in protobuf and returning that.

zehiko: discussed offline - we'll keep this API for now, but we might eventually end up passing just the arrow ``Schema`` around. For now it makes sense to got with ``ColumnDescriptor``s as it simplifies current client integration and usage and it's also kinda symmetrical with other protobuf definitions.

PR #8652, opened 2025-01-10
Filter entities in the UI (part 1): Introduce a filter widget

abey79:
### Related

- Part of #8586
- Part of a series of PR:
  - #8645
  - #8652
  - ...

### What

This PR introduce a filter widget with the following feature:
- the widget acts as a toggle: it's either on or off
- when off, it's just a right-aligned icon and a (more-or-less) regular section title
- when on, the title is replaced by a text edit
- each time it's toggled on, the filter widget maintains a session id, for the client code to use to cache filter-session-specific data (e.g. collapsedness state)
- this also include a `FilterMatcher`, which is an helper object that performs the actual filtering
  **NOTE**: this will surely be entirely rewritten as we improve on the filter semantics
- helper function to create a properly formatted matched text (with the matching parts highlighted)

image

image


crates/viewer/re_ui/src/filter_widget.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+}
+
+// --
+
+/// Full-text, case-insensitive matcher.
+pub struct FilterMatcher {
+    /// The lowercase version of the query string.
+    ///
+    /// If this is `None` or `Some("")`, the matcher will accept any input.
```
Wumpf, line 154:
```suggestion
    /// If this is `None` the matcher will accept any input.
    /// If this is `Some("")` the matcher will reject any input.
```

That's at least what's consistently implemented down there!

crates/viewer/re_ui/src/filter_widget.rs
```diff
@@ -0,0 +1,276 @@
...
+        match self.lowercase_query.as_deref() {
+            None => true,
+            Some("") => false,
+            Some(query) => text.to_lowercase().contains(query),
+        }
+    }
+
+    /// Find all matches in the given text.
+    ///
+    /// Can be used to format the text, e.g. with [`format_matching_text`].
```
Wumpf, line 192: should clarify the `None` handling. Was surprised that `lowercase_query` `None` has a different meaning here again, but I think it makes sense given what this is used for.

crates/viewer/re_ui/src/filter_widget.rs
```diff
@@ -0,0 +1,276 @@
...
+    pub fn find_matches(&self, text: &str) -> Option<impl Iterator<Item = (usize, usize)> + '_> {
```
Wumpf, line 193:
```suggestion
    pub fn find_matches(&self, text: &str) -> Option<impl Iterator<Item = Range<usize>)> + '_> {
```

Would make the semantic **a lot** clearer

crates/viewer/re_ui/src/filter_widget.rs
```diff
@@ -0,0 +1,276 @@
...
+
+impl FilterState {
+    /// Return the filter if any.
+    ///
+    /// The widget must be enabled _and_ the filter string must not be empty.
+    pub fn query(&self) -> Option<&str> {
+        self.inner_state
+            .as_ref()
+            .map(|state| state.filter_query.as_str())
+    }
```
Wumpf, line 60: comment and code aren't the same: if I have an empty query, I get still `Some` -> which then leads to a `Some("")` filter matcher which is distinct of of a `None` filter matcher which never gets hit.

Wumpf (changes requested):
nice!
there seems to be some mixup between empty string and None matchers that needs to be cleaned up

abey79:
Arf yeah I switch last minute to "" matching nothing instead of everything (upon Katya's suggestion during the demo), and clearly I didn't cleanup properly.

Wumpf (approved):
thanks!

Issue #8639, opened 2025-01-10
Viewer fails rendering scenes with zero length linestrips


Wumpf:
Repro with 2D line strips
```cpp
#include <rerun.hpp>

int main() {
    const auto rec = rerun::RecordingStream("rerun_example_line_strip2d_failure");
    rec.spawn().exit_on_failure();
    std::vector<rerun::Vec2D> points;
    auto line = rerun::LineStrip2D(points);
    rec.log("Annotation/Lines", rerun::LineStrips2D(line));
}
```

Causes
```
[2025-01-10T13:42:03Z ERROR wgpu_core::device::global] Device::create_texture error: Dimension Y is zero
[2025-01-10T13:42:03Z ERROR wgpu_core::device::global] Texture::create_view(Id(4,1)) error: Texture with '' label is invalid
[2025-01-10T13:42:03Z WARN  re_renderer::error_handling::wgpu_core_error] unknown error cause cause=InvalidDimension(Zero(Y))
[2025-01-10T13:42:03Z ERROR re_renderer::error_handling::error_tracker] Wgpu validation error 1: Validation Error

    Caused by:
      In Device::create_texture
        Dimension Y is zero

[2025-01-10T13:42:03Z ERROR re_renderer::error_handling::error_tracker] Wgpu validation error 2: Validation Error

    Caused by:
      In Device::create_bind_group
        TextureView with '' label is invalid

[2025-01-10T13:42:03Z ERROR re_renderer::error_handling::error_tracker] Wgpu validation error 3: Validation Error

    Caused by:
      In RenderPass::set_bind_group
        In a set_bind_group command
          BindGroup with '' label is invalid`
```

depending on _circumstances_ you either don't get to see the viewer at all or just get log spamed

piyush-m0:
Further observation being, trying to pause almost crashes the viewer, closing the program through terminal doesn't work.

PR #8653, opened 2025-01-10
Handle empty line strips in the viewer

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/8639

### What

Adds a regression test for this so this doesn't happen again (this is vaguely familiar, I think this happened before) - which we can finally do now because we very recently ensured that we always have a (software) graphics device on ci \o/
Testing this directly in re_renderer required a little bit of new test framework, but it's fairly compact and unifies a bit of code we already had in the viewer.


crates/viewer/re_renderer/src/config.rs
```diff
@@ -366,6 +369,72 @@ pub fn instance_descriptor(force_backend: Option<&str>) -> wgpu::InstanceDescrip
...
+    //   (TODO(andreas): Ideally we'd test that as well to check it is well-behaved,
+    //   but for now we only want to have a look at the happy path)
+    let backends = wgpu::Backends::VULKAN | wgpu::Backends::METAL;
+
+    let flags = (
+        wgpu::InstanceFlags::ALLOW_UNDERLYING_NONCOMPLIANT_ADAPTER | wgpu::InstanceFlags::VALIDATION
+        // TODO(andreas): GPU based validation layer sounds like a great idea,
+        // but as of writing this makes tests crash on my Windows machine!
+        // It looks like the crash is in the Vulkan/Nvidia driver, but further investigation is needed.
+        // | wgpu::InstanceFlags::GPU_BASED_VALIDATION
```
Wumpf, line 389: this bit has changed, otherwise this is moved as-is from viewer's test_context

crates/viewer/re_renderer/src/renderer/lines.rs
```diff
@@ -343,7 +343,7 @@ impl LineDrawData {
 
         let line_renderer = ctx.renderer::<LineRenderer>();
 
-        if strips_buffer.is_empty() {
+        if strips_buffer.is_empty() || vertices_buffer.is_empty() {
```
Wumpf, line 346: this is the actual fix

crates/viewer/re_renderer/src/renderer/lines.rs
```diff
@@ -762,3 +762,45 @@ impl Renderer for LineRenderer {
...
+mod tests {
+    use crate::{view_builder::TargetConfiguration, Rgba};
+
+    use super::*;
+
+    // Regression test for https://github.com/rerun-io/rerun/issues/8639
+    #[test]
+    fn empty_strips() {
+        re_log::setup_logging();
+        re_log::PanicOnWarnScope::new();
```
abey79, line 776: (off topic) `TestContext` should probably do that too!

abey79: Yeah, I was thinking opt-out indeed.

Wumpf: let's sneak that in here as well, why not

PR #8654, opened 2025-01-10
Filter entities in the UI (part 2): Introduce entity filtering in the time panel

abey79:
### Related

- Part of #8586
- Part of a series of PR:
  - #8645
  - #8652
  - ...

### What

This PR introduces the filter widget to the time panel. In doing so, it deals with several shenanigans:

- The item collapsedness must not be shared between the unfiltered view, and across different filtering sessions
  **Note**: the collapsed state being persisted, we are now accumulating collapsedness state for every single filter session in the persistant storage. This is not ideal, but addressing that would required rolling our own collapsed state, which would leak all over the place (list item and many of its users).
- The collapse/expend all context menu must be made aware of what precedes.
- The time panel filter state must be invalidated upon switching active recording (otherwise it's weird UX).
- The streams tree UI area needs to have a minimum width otherwise search icon placement/layout breaks.

And of course the actual filtering logic must be implement. For now, the semantics are: "one of the entity part must contain the query string in its entirety". This is very (too) basic and will evolve in future PR (see https://github.com/rerun-io/rerun/issues/8586#issuecomment-2578060663).

### DIFF
--- a/crates/viewer/re_time_panel/src/lib.rs
+++ b/crates/viewer/re_time_panel/src/lib.rs
@@ -141,6 +141,13 @@ pub struct TimePanel {
     /// The filter widget state
     #[serde(skip)]
     filter_state: filter_widget::FilterState,
+
+    /// The store id the filter widget relates to.
+    ///
+    /// Used to invalidate the filter state (aka deactivate it) when the user switch the current
+    /// recording.
+    #[serde(skip)]
+    filter_state_store_id: Option<re_log_types::StoreId>,
 }
 
 impl Default for TimePanel {
@@ -155,6 +162,7 @@ impl Default for TimePanel {
             time_control_ui: TimeControlUi,
             source: TimePanelSource::Recording,
             filter_state: Default::default(),
+            filter_state_store_id: None,
         }
     }
 }
@@ -188,6 +196,12 @@ impl TimePanel {
     ) {
         if state.is_hidden() {
             return;
+        }
+
+        // Invalidate the filter widget if the store id has changed.
+        if self.filter_state_store_id != Some(entity_db.store_id()) {
+            self.filter_state = Default::default();
+            self.filter_state_store_id = Some(entity_db.store_id());
         }
 
         self.data_density_graph_painter.begin_frame(ui.ctx());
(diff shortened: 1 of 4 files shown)


crates/viewer/re_time_panel/src/lib.rs
```diff
@@ -185,6 +198,12 @@ impl TimePanel {
             return;
         }
 
+        // Invalidate the filter widget if the store id has changed.
+        if self.filter_state_store_id != Some(entity_db.store_id()) {
+            self.filter_state = Default::default();
+            self.filter_state_store_id = Some(entity_db.store_id());
+        }
```
Wumpf, line 205: This feels wrong: Can we keep the id on the `RecordingConfig` so that we don't loose the filter when changing recording?
If a user wants to go back and forth between two similar recording this lack of state keeping can be very annoying.

crates/viewer/re_time_panel/src/lib.rs
```diff
@@ -627,57 +662,96 @@ impl TimePanel {
...
-            }
-        });
+        // Globally unique id that is dependent on the "nature" of the tree (recording or blueprint,
+        // in a filter session or not)
+        let id = collapse_scope.entity(tree.path.clone()).into();
+
+        let is_short_path = tree.path.len() <= 1 && !tree.is_leaf();
+        let default_open = self.filter_state.session_id().is_some() || is_short_path;
+
+        //dbg!((collapse_scope, default_open));
```
Wumpf, line 754: dead code

Wumpf (approved):
lgtm, get to have this in

Unfortunately makes an existing perf issue more prominent: We try to show the histogram for everything that's expanded and once there's a search we expand everything. We need to figure out how to disable the histogram showing code but this isn't an issue introduced here.

crates/viewer/re_time_panel/src/lib.rs
abey79: Well I added that _precisely_ because I wanted to disable the filter upon recording change, because it was extremely annoying to have it. Now that you mention it, I see how it could be useful to have it.

My suggestion would be to use the app id instead. Makes sense to you?

Wumpf: oh yeah that sounds great, then the filter sticks with the different recordings within the same app 💡

PR #8656, opened 2025-01-10
Fix formatting of RowId/Tuid when printing ChunkStore

emilk:
Also added a regression test

* Part of #3741


crates/store/re_format_arrow/src/lib.rs
```diff
@@ -244,21 +233,10 @@ where
...
         let mut row = Row::new();
-        row.add_cell({
-            let cell = Cell::new(format!(
-                "CHUNK METADATA:\n{}",
-                DisplayMetadata(metadata, "* ")
-            ));
-
-            #[cfg(not(target_arch = "wasm32"))] // requires TTY
-            {
-                cell.add_attribute(comfy_table::Attribute::Italic)
```
emilk, line 255: I had to remove the italics, otherwise the formatting was different locally (tty) and on CI

emilk:
@rerun-bot full-check

teh-cmc (approved):
I'm afraid that test is going to fail a bit too frequently, but we'll see how thing goes :+1: 

Took the liberty to update the test to use the new SerializedBatch APIs!

PR #8657, opened 2025-01-10
Port row_ids to arrow1

emilk:
### Related
* Part of #3741


emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

crates/store/re_chunk/src/chunk.rs
```diff
@@ -943,7 +934,11 @@ impl Chunk {
             entity_path,
             heap_size_bytes: Default::default(),
             is_sorted: true,
-            row_ids: Arrow2StructArray::new_empty(RowId::arrow2_datatype()),
+            row_ids: arrow::array::StructBuilder::from_fields(
```
teh-cmc, line 937: Here and elsewhere, it feels odd to me that we're going through a builder type just to build an empty thing.

Any particular reason we're not calling https://docs.rs/arrow/latest/arrow/array/fn.new_empty_array.html ?

crates/store/re_types_core/src/tuid.rs
```diff
@@ -11,13 +11,18 @@ use crate::{DeserializationError, Loggable};
 
 // ---
 
+// TODO(emilk): This is a bit ugly… but good enough for now?
+pub fn tuid_arrow_fields() -> Fields {
```
teh-cmc, line 15: Can't you get rid of this if you get rid of using the builder type for an empty array?

crates/utils/re_byte_size/src/arrow_sizes.rs
```diff
@@ -13,17 +13,24 @@ impl SizeBytes for dyn Array {
     }
 }
 
-impl SizeBytes for ArrayRef {
+impl SizeBytes for &dyn Array {
```
teh-cmc, line 16: This should be removed, `impl SizeBytes for dyn Array` already covers it since the methods take `&self`.

crates/utils/re_byte_size/src/arrow_sizes.rs
```diff
@@ -13,17 +13,24 @@ impl SizeBytes for dyn Array {
...
     #[inline]
     fn heap_size_bytes(&self) -> u64 {
         self.get_array_memory_size() as u64
     }
 }
 
-impl<T: ArrowPrimitiveType> SizeBytes for PrimitiveArray<T> {
+impl<T: Array> SizeBytes for &T {
```
teh-cmc, line 23: Method already takes `&self`:
```suggestion
impl<T: Array> SizeBytes for T {
```

teh-cmc (approved):

crates/store/re_chunk/src/chunk.rs
emilk: `new_empty_array` returns an `ArrayRef`, meaning we would need a downcast+unwrap here.
`StructBuilder` returns what we want.

crates/utils/re_byte_size/src/arrow_sizes.rs
emilk: I can't: 

image

crates/store/re_chunk/src/chunk.rs
emilk: …but it may be worth it to get rid of `tuid_arrow_fields`

crates/store/re_types_core/src/tuid.rs
emilk: Yes, but at the cost of some ugly dynamic casting and unwraps that I'd rather avoid

PR #8658, opened 2025-01-11
Refactor `TimeColumnDescriptor`

emilk:
* Part of #3741


crates/store/re_dataframe/src/query.rs
```diff
@@ -398,7 +398,6 @@ impl<E: StorageEngineLike> QueryHandle<E> {
                                             // TODO(cmc): I picked a sequence here because I have to pick something.
                                             // It doesn't matter, only the name will remain in the Arrow schema anyhow.
                                             timeline: Timeline::new_sequence(*selected_timeline),
-                                            datatype: arrow2::datatypes::DataType::Null,
```
emilk, line 401: I didn't realize until after the refactor, that this `Null` was maybe the whole point of the field. Do we want that `NullArray` in the return type? I believe this is for "failed" lookups (when the timeline the user asked for doesn't exist), so I guess `NullArray` makes some sense. If so I can either revert this PR (and replace it with a comment on `TimeColumnDescriptor::datatype` to explain why it is there), or maybe add an examplit null-state to `TimeColumnDescriptor`.

teh-cmc: > Do we want that NullArray in the return type?

Yes. Everything in the dataframe API is allowed to be a `Null` array at any point.

> If so I can either revert this PR (and replace it with a comment on TimeColumnDescriptor::datatype to explain why it is there)

:+1:

emilk:
@rerun-bot full-check

teh-cmc (approved):

PR #8659, opened 2025-01-12
docs: use `connect_tcp` instead of deprecated `connect`

OlivierLDff:
Hi I'm going thru the docs for the first time, I did some copy and pasta to connect to the server and I got hit by some deprecated warning. It seems the code wasn't even working and `connect_opts` should have been the one.


emilk (approved):
Thanks, and good catch

Wumpf:
ci failure is because of an oversight on my side: need to enable lfs checkout on that ci step now

PR #8650, opened 2025-01-10
Eagerly serialized, partially updatable 2D & 3D line strips

teh-cmc:
This PR makes `LineStrips2D` and `LineStrips3D` eagerly serialized and partially update-able.

This demonstrates how to migrate an archetype in the simple case.

* DNM: requires #8649  
* Part of #8581

PR #8666, opened 2025-01-13
Add `archetype_native` attribute and make `ContainerBlueprint` eager + partial

teh-cmc:
Introduces a new attribute that, when paired with an eager archetype, will also generate a native companion type with back-and-forth conversion methods.

This is especially useful for blueprint archetypes (although there are definitely non-blueprint examples too), which are heavily used all across the viewer, and would be very painful to use otherwise.

```rust
/// Whether we should generate an extra Rust object comprised of native Rust types.
///
/// The generated object will have the name of the archetype, prefixed by `Native`,
/// e.g. `NativePoints3D`.
///
/// Applies only to *eager* archetypes. No-op otherwise.
attribute "attr.rust.archetype_native";
```

With this, we now should have all the tools required to port every remaining archetype (both blueprint and data).

* Part of #8650

### DIFF
--- a/crates/build/re_types_builder/src/codegen/rust/api.rs
+++ b/crates/build/re_types_builder/src/codegen/rust/api.rs
@@ -335,69 +335,7 @@ fn quote_struct(
     } else {
         quote! { pub struct #name { #(#quoted_fields,)* }}
     };
-    let quoted_struct_native = obj_native.as_ref().map(|obj_native| {
-        let native_name = format_ident!("{}", obj_native.name);
-
-        let quoted_fields = obj_native
-            .fields
-            .iter()
-            .map(|obj_field| ObjectFieldTokenizer(reporter, obj_native, obj_field).quoted(objects));
-        let quoted_struct = if is_tuple_struct {
-            quote! { pub struct #native_name(#(#quoted_fields,)*); }
-        } else {
-            quote! { pub struct #native_name { #(#quoted_fields,)* }}
-        };
-
-        let eager_fields_to_native_fields = obj.fields.iter().map(|field| {
-            let field_name = format_ident!("{}", field.name);
-            quote!(value.#field_name.clone().map(|batch| (batch.descriptor, batch.array)))
-        });
-        let eager_to_native = quote! {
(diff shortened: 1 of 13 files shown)


teh-cmc:
Ha yeah I also ported `VisualBounds2D` and `Background` because it didn't require any particular effort.

crates/build/re_types_builder/src/codegen/rust/api.rs
```diff
@@ -309,10 +335,74 @@ fn quote_struct(
...
+        quote! {
+            #[doc(hidden)]
+            #quoted_derive_clone_debug
+            #quoted_struct
+
+            #eager_to_native
+
+            #native_to_eager
+        }
+    });
```
emilk, line 400: Can this be moved to a new function to lessen the length and complexity of the current already-very-long function?  It's very hard to see what inputs this piece of code is using

crates/build/re_types_builder/src/codegen/rust/api.rs
```diff
@@ -309,10 +335,74 @@ fn quote_struct(
...
+            let field_name = format_ident!("{}", field.name);
+            if field.is_nullable {
+                quote!(#field_name: value.#field_name.as_ref().and_then(|v| v.serialized()))
+            } else {
+                quote!(#field_name: value.#field_name.serialized())
+            }
+        });
+        let native_to_eager = quote! {
+            impl From<&#native_name> for #name {
+                #[rustfmt::skip] // so it doesn't take 1000 lines for no reason
```
emilk, line 381: What do you mean? Why would `cargo fmt` produce 1000 lines?

emilk (approved):
Nice!

crates/build/re_types_builder/src/codegen/rust/api.rs
teh-cmc: rustfmt splits method chains no matter their length, so all the `value.#field_name.clone().map(|batch| (batch.descriptor, batch.array))` emitted by this method turn into 4 lines each, which is really pushing it.

Issue #8665, opened 2025-01-13
BUILD.md doesn't list a specific pixi version, but the current pixi release is incompatible with rerun (apparently)


photex:
**Describe the bug**
Attempting to setup a development environment by following the instructions in BUILD.md (Linux) and at the step `pixi run check-env` the following is the result:

```sh
[chipc@bucket rerun]$ pixi run check-env
 WARN Encountered ambiguous version specifier `1.5.1`, could be `1.5.1.*` but assuming you meant `==1.5.1`. In the future this will result in an error.
 WARN Encountered ambiguous version specifier `18.0.0`, could be `18.0.0.*` but assuming you meant `==18.0.0`. In the future this will result in an error.
 WARN Encountered ambiguous version specifier `19.37.32822`, could be `19.37.32822.*` but assuming you meant `==19.37.32822`. In the future this will result in an error.
 WARN Encountered ambiguous version specifier `16.0.6`, could be `16.0.6.*` but assuming you meant `==16.0.6`. In the future this will result in an error.
 WARN Encountered ambiguous version specifier `1.11.1`, could be `1.11.1.*` but assuming you meant `==1.11.1`. In the future this will result in an error.
 WARN Encountered ambiguous version specifier `16.0.6`, could be `16.0.6.*` but assuming you meant `==16.0.6`. In the future this will result in an error.
 WARN Encountered ambiguous version specifier `3.27.6`, could be `3.27.6.*` but assuming you meant `==3.27.6`. In the future this will result in an error.
 WARN Encountered ambiguous version specifier `1.8.0`, could be `1.8.0.*` but assuming you meant `==1.8.0`. In the future this will result in an error.
 WARN Encountered ambiguous version specifier `1.11.1`, could be `1.11.1.*` but assuming you meant `==1.11.1`. In the future this will result in an error.
 WARN Encountered ambiguous version specifier `18.0.0`, could be `18.0.0.*` but assuming you meant `==18.0.0`. In the future this will result in an error.
 WARN Encountered ambiguous version specifier `0.3.5`, could be `0.3.5.*` but assuming you meant `==0.3.5`. In the future this will result in an error.
Error: 
  × Unexpected keys, expected only 'cmd', 'depends-on', 'description'
     ╭─[pixi.toml:214:9]
 213 │ ] }
 214 │ fmt = { depends_on = ["format"] }
     ·         ─────┬────
     ·              ╰── 'depends_on' was not expected here
 215 │ 
     ╰────
[...]

PR #8667, opened 2025-01-13
Fix: rename `depends_on` to `depends-on` in `pixi.toml`

baszalmstra:
### Related

* Closes https://github.com/rerun-io/rerun/issues/8665

### What

Pixi 0.40.0 removes the long deprecated `depends_on` field in favor of `depends-on`. The deprecation didnt result in an error or warning which was a bit annoying.


jleibs:
@baszalmstra thanks for the patch!

PR #8671, opened 2025-01-13
Autogenerated partial updates APIs for Python

### LINKED ISSUE #8582: New partial updates logging APIs: Python

teh-cmc:
There really is not much to be said that wouldn't be made clearer by just looking at the code.

Specifically:
* Look at the codegen changes.
* Now look at the changes to one of the generated archetypes (probably `Transform3D`).
* Now look especially at the changes to `transform3d_ext`. That one's important.
* Now look at how the different "partial updates" snippets evolved.

Overall I hate everything here, literally every single bit -- but that's the best I managed to get out of the status quo without ridiculous amounts of efforts.

* DNM: requires #8690 
* Closes #8582


teh-cmc:
@rerun-bot full-check

teh-cmc:
@rerun-bot full-check

nikolausWest:
I have more of a naming feedback: The use of "update" here makes me assume some state is being mutated on an object when what's actually happening is we're calling a class method that returns a list of component batches. Some other naming directions to consider:

```python
rr.log("points", rr.Points3D.fields(clear=True, positions=positions, radii=0.3))
rr.log("points", rr.Points3D.build_fields(clear=True, positions=positions, radii=0.3))
rr.log("points", rr.Points3D.construct_fields(clear=True, positions=positions, radii=0.3))
rr.log("points", rr.Points3D.updating_fields(clear=True, positions=positions, radii=0.3))
rr.log("points", rr.Points3D.incremental_fields(clear=True, positions=positions, radii=0.3))
```

teh-cmc:
`Archetype.fields()` is probably best now that we've decided to add a `clear: bool` parameter to it.

There's still the question of wheter `Archetype.clear_fields()` should be `Archetype.clear()` instead.

rerun_py/rerun_sdk/rerun/archetypes/points3d.py
```diff
@@ -153,9 +153,91 @@ def _clear(cls) -> Points3D:
...
+        """Clear all the fields of a `Points3D`."""
+        inst = cls.__new__(cls)
+        inst.__attrs_init__(
+            positions=[],  # type: ignore[arg-type]
+            radii=[],  # type: ignore[arg-type]
+            colors=[],  # type: ignore[arg-type]
+            labels=[],  # type: ignore[arg-type]
+            show_labels=[],  # type: ignore[arg-type]
+            class_ids=[],  # type: ignore[arg-type]
+            keypoint_ids=[],  # type: ignore[arg-type]
```
abey79, line 234: With my proposal above these "ignore" should not be required.

rerun_py/rerun_sdk/rerun/archetypes/points3d.py
```diff
@@ -153,9 +153,91 @@ def _clear(cls) -> Points3D:
...
+        )
+        return inst
+
     positions: components.Position3DBatch = field(
-        metadata={"component": "required"},
-        converter=components.Position3DBatch._required,  # type: ignore[misc]
+        metadata={"component": "optional"},
```
abey79, line 239: This is broken imo, as per my comment above.

rerun_py/rerun_sdk/rerun/archetypes/points3d.py
```diff
@@ -153,9 +153,91 @@ def _clear(cls) -> Points3D:
...
+        converter=components.Position3DBatch._optional,  # type: ignore[misc]
```
abey79, line 240: Typing wise, all component become optional with my proposal. So we can remove `BaseBatch._required` and rename `BaseBatch._optional` to `_converter` or something.

rerun_py/rerun_sdk/rerun/archetypes/points3d.py
```diff
@@ -153,9 +153,91 @@ def _clear(cls) -> Points3D:
...
+            "labels": labels,
+            "show_labels": show_labels,
+            "class_ids": class_ids,
+            "keypoint_ids": keypoint_ids,
+        }
+
+        if clear:
+            kwargs = {k: v if v is not None else [] for k, v in kwargs.items()}  # type: ignore[misc]
+
+        return Points3D(**kwargs)  # type: ignore[arg-type]
```
abey79, line 221: We should use the `__new__`/`__attrs_init__` trick here too. Then I think it _should_ by type checkable even.

abey79 (changes requested):
I have so far only looked at the codegen'ed archetypes, and I'm a bit uncomfortable with letting this go as is. I think that, with some reasonable effort, we can make this more self consistent.

First: let's accept that archetype data members are _not_ typed about required-ness, because required-ness is contextual (aka applies only upon first logging an archetype, not updating). So _all_ archetypes field should be: `: ComponentXXXBatch | None | [Never]`.

I don't quite recall with this `[]` business is about in `clear_fields()`. I assume it somehow results in clearing that component. [Apparently](https://stackoverflow.com/a/76881736/229511), we can use `[Never]` as typing hint (as I did above) for list that must be empty. We can try to roll with it 🤷🏻 An alternative would be have some kind of marker object.

Second: I _think_ `metadata={"component": "optional"}` breaks at least `Archetype.num_instances()`, so this shouldn't be changed. If that change was required, then we should get to the bottom of it.

abey79:
On second thought, I can't find a single place where this `Archetype.num_instances()` is used. The docstring even lies about it. I'd kill it altogether. The value of "component" seem to matter little in the two other places were it's used, but having it set to `"required"` makes  `__str__` more correct, so let's revert that.

rerun_py/rerun_sdk/rerun/archetypes/points3d.py
teh-cmc: We need to call the real (possibly custom) constructor though.

teh-cmc:
The lack of `| None` in the archetype definition itself was not by design -- I missed yet another `is_nullable()` condition in the codegen. This is fixed now. It removed the need for some of the new `type[ignore]`s, but apparently it also made new ones appear...? I'll have a quick look.

abey79:
@rerun-bot full-check

abey79 (approved):
Okay, I think we're good do go. Let's see what the full check says. As discussed, I would like to have a stronger testing story about that. Hopefully we'll at least built some snippets/roundtrips over time.

To emphasis an important design points (for the benefit of bystanders). This PR provides a codegened, archetype-fields/components-centric update mechanism. As such, `update_fields` doesn't provide the convenience that some custom `__init__` do. For example, `Asset3D()` accepts a file path, but `Asset3D.update_fields()` doesn't, because the `Asset3D` archetype doesn't have a path field (it just has a `blob` field).

As such, `update_fields()` can be seen as a lower-level mechanism than the `__init__`. It is however more compact and convenient than the status quo, which consists of spelling out components explicitly. So this PR is a net gain.

teh-cmc:
@rerun-bot full-check

PR #8672, opened 2025-01-13
Filter entities in the UI (part 3): Move action to a menu in the blueprint panel and keep default blueprint when using heuristics

abey79:
### Related

- Part of #8586
- Part of a series of PR:
  - #8645
  - #8652
  - #8654
  - #8672
  - ...

### What

In preparation for adding the filter UI in the blueprint panel, we must move the current actions (add view/container, reset blueprint) to a "more" (`…`) icon menu. This in turn allows us to be explicit with the reset actions, and have both "reset to default" and "reset to heuristics" (like in the application's selection panel).

In turn, this exposes the weird behaviour where resetting to heuristics also deletes the current blueprint, for no reason other than it made implementation easier. This PR fixes that. So it's not possible to reset to heuristics and then go back to the default blueprint.

Also:
- Remove the "Reset to default"/"Rest to heuristic" buttons from the application selection panel (double duty).
- Reworded the "Reset to heuristics" command palette actions for consistency
  - See also: https://github.com/rerun-io/rerun/issues/8677 

https://github.com/user-attachments/assets/c2334c1f-fca0-4c9e-a547-944bb4fc066e


crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
```diff
@@ -868,46 +853,97 @@ impl BlueprintTree {
...
+    ctx: &ViewerContext<'_>,
+    viewport: &ViewportBlueprint,
+    ui: &mut egui::Ui,
+) {
+    let enabled = !viewport.auto_layout() || !viewport.auto_views();
+
+    if ui
+        .add_enabled(
+            enabled,
+            egui::Button::image_and_text(&re_ui::icons::RESET, "Reset to heuristic blueprint"),
```
emilk, line 936: I wonder if we can come up with better copy here. "Create automatic blueprint"? Maybe @gavrelina has ideas

emilk (approved):
Makes sense, but it is also all very confusing to me, because we are inconsistent in the naming.

We have the _default_ blueprint (fine), and then we have the "automatic" "generated" "heuristic" blueprint. I wish we could settle on one name for it, instead of three.

crates/viewer/re_viewer_context/src/command_sender.rs
```diff
@@ -30,7 +30,10 @@ pub enum SystemCommand {
     /// Reset the `Blueprint` to the default state
     ClearActiveBlueprint,
 
-    /// Clear the blueprint and generate a new one
+    /// Clear the blueprint and enable heuristics.
+    ///
+    /// Note: this engages the heuristics even if a default blueprint exists, so the default
+    /// blueprint may be restored.
```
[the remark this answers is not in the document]
abey79: Bad wording on my end, I meant to imply that the default blueprint (if any) is left alone, so _the user_ may restore it later on. I've rewritten entirely that (and other) comment to clarify this.

crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
abey79: We discussed that and decided to stick with "heuristics" for now.

abey79:
> We have the _default_ blueprint (fine), and then we have the "automatic" "generated" "heuristic" blueprint. I wish we could settle on one name for it, instead of three.

I tried to further cleaned that up. We should never use "generated". We agree to keep "heuristics" for user-facing stuff. We still have `auto_layout`/`auto_view` in the code. Maybe we should renamed those as well, but I'll keep that for another day.

Issue #8329, opened 2024-12-05
Time selection control javascript API


nikolausWest:
The Rerun web-viewer can be embedded into other web applications that may also be showing data. For instance, Rerun's 2D plotting functionality isn't yet very mature and many teams combine Rerun with more specialized 2D plotting tools like vega-lite or D3.js. This experience would be significantly better if it was possible to sync the Rerun viewer with the surrounding javascript environment. A great first step would be the ability to sync the time cursor position and current timeline between Rerun and javascript.

## What?
- Expose a javascript API for setting the current timeline, current timepoint, and playmode from javascript
- Expose js callbacks for getting updates on when those change in Rerun

## Future extensions
- We could also expose on-click callbacks to allow people to build simpler data annotation tools around the Rerun embedded web-viewer
- This can be seen as a precursor to https://github.com/rerun-io/rerun/issues/2691, which in addition requires proper two-way communication with the SDK which is a much bigger story.

ryh-zz:
@Wumpf  Are javascript apis in the future?

PR #8673, opened 2025-01-13
Add JS timeline control and callback APIs

jprochazk:
### Related

* Closes https://github.com/rerun-io/rerun/issues/8329

### What

Adds the following methods to `WebViewer`:

```ts
get_active_recording_id(): string | null;
set_active_recording_id(value: string): void;
get_playing(recording_id: string): boolean;
set_playing(recording_id: string, value: boolean): void;
get_current_time(recording_id: string, timeline: string): number;
set_current_time(recording_id: string, timeline: string, time: number): void;
get_active_timeline(recording_id: string): string | null;
set_active_timeline(recording_id: string, timeline: string): void;
get_time_range(recording_id: string, timeline: string): { min: number; max: number; } | null;
```

As well as the following events:
```ts
interface Events {
  timelinechange: [timeline: string, time: number];
  timeupdate: number;
  play: void;
  pause: void;
}
```

See https://github.com/rerun-io/web-viewer-example/blob/jan/temp-timeline-controls-example/src/main.ts#L15 for an example usage of these new APIs

### DIFF
--- a/crates/viewer/re_viewer/src/app_state.rs
+++ b/crates/viewer/re_viewer/src/app_state.rs
@@ -151,6 +151,7 @@ impl AppState {
         command_sender: &CommandSender,
         welcome_screen_state: &WelcomeScreenState,
         is_history_enabled: bool,
+        timeline_callbacks: Option<&re_viewer_context::TimelineCallbacks>,
     ) {
         re_tracing::profile_function!();
 
@@ -284,7 +285,8 @@ impl AppState {
 
         // We move the time at the very start of the frame,
         // so that we always show the latest data when we're in "follow" mode.
-        move_time(&ctx, recording, rx);
+
+        move_time(&ctx, recording, rx, timeline_callbacks);
 
         // Update the viewport. May spawn new views and handle queued requests (like screenshots).
         viewport_ui.on_frame_start(&ctx);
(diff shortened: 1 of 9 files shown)


jprochazk:
(drafted because I'm waiting for lints)

crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -284,7 +285,8 @@ impl AppState {
 
         // We move the time at the very start of the frame,
         // so that we always show the latest data when we're in "follow" mode.
-        move_time(&ctx, recording, rx);
+
```
emilk, line 288:
```suggestion
```

crates/viewer/re_viewer_context/src/time_control.rs
```diff
@@ -539,34 +636,50 @@ impl TimeControl {
         self.set_time(time);
     }
 
+    pub fn get_time_for_timeline(&self, timeline: Timeline) -> Option<TimeReal> {
```
emilk, line 639: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

```suggestion
    pub fn time_for_timeline(&self, timeline: Timeline) -> Option<TimeReal> {
```

rerun_js/web-viewer/index.ts
```diff
@@ -496,6 +519,144 @@ export class WebViewer {
...
+
+    this.#handle.set_playing(recording_id, value);
+  }
+
+  /**
+   * Get the current time.
+   *
+   * This always returns `0` if the recording or timeline can't be found.
+   */
+  get_current_time(recording_id: string, timeline: string): number {
```
emilk, line 583: We should explain the unit here and in `set_current_time` (nanoseconds for temporal timelines)

emilk (approved):

PR #8682, opened 2025-01-14
Port Transform3D to eager serialization & update API

Wumpf:
### Related

### What

Use new eager serialization & update API for transforms.

The only breaking change here is that Transform3D is no longer copy, otherwise it's fully compatible.


teh-cmc (approved):
:drooling_face:

teh-cmc:
I need this in so I'll take care of it @Wumpf

PR #8683, opened 2025-01-14
Fix flaky `re_graph_view` test by removing dynamically-placed node

grtlr:
### What

Title.

The downside of this is that we now skip the actual layout algorithm entirely, as all nodes are placed at fixed positions.

I somehow need to figure out where the determinism in `fjadra` is coming from: 
* https://github.com/grtlr/fjadra/issues/5


teh-cmc (approved):

teh-cmc:
Can confirm the test now passes on my machine.

PR #8687, opened 2025-01-14
Port `TransportChunk::schema` to arrow-rs

emilk:
* Part of #3741


emilk:
@rerun-bot full-check

Issue #8580, opened 2025-01-06
New partial updates logging APIs


teh-cmc:
We want to offer newer & nicer ergonomic partial update APIs on the logging side (Python, Rust, C++), something like this:
```py
rr.set_time_sequence("frame", 42)
rr.log("points", rr.Points3D([[0, 0, 0], [1, 1, 1]], colors=[255, 0, 0]))

rr.set_time_sequence("frame", 43)
rr.log("points", [rr.Points3D.components(colors=[0, 0, 255])]) # gets properly tagged
```

* This is a subset of the larger proposal in https://github.com/rerun-io/rerun/issues/3381.

We want this for two reasons:
* Finally make partial updates ergonomic.
* Unblock tagged components.

Complete rationale can be found in https://github.com/rerun-io/rerun/issues/3381#issuecomment-2551831346.

nikolausWest:
I'd love it if the last line could be
```python
rr.log("points",` rr.Points3D.components(colors=[0, 0, 255])) # gets properly tagged
```
instead of 
```python
rr.log("points",` [rr.Points3D.components(colors=[0, 0, 255])]) # gets properly tagged
```
(i.e. don't require wrapping in a list)

nikolausWest:
I think we also talked about going with `.fields()` instead of `.components()` but I don't remember the conclusion

PR #8690, opened 2025-01-14
Add new `Transform3D` partial updates snippet for all languages

teh-cmc:
Adds some much needed snippet for partial updates on `Transform3D`.
Much needed because A) `Transform3D` has a long history of custom partial updatability hacks and B) users are very very likely to use that one.

* Part of #8580


docs/snippets/all/archetypes/transform3d_partial_updates.cpp
```diff
@@ -0,0 +1,64 @@
...
+            "box",
+            rerun::Transform3D().with_rotation_axis_angle(
+                rerun::RotationAxisAngle({0.0f, 1.0f, 0.0f}, rerun::Angle::radians(rad))
+            )
+        );
+    }
+
+    step++;
+    rec.set_time_sequence("step", step);
+    // TODO(cmc): clear_fields
```
emilk, line 62: Link to the issue for partial updates in C++

docs/snippets/all/archetypes/transform3d_partial_updates.py
```diff
@@ -0,0 +1,66 @@
...
+    rr.Boxes3D(half_sizes=[4.0, 2.0, 1.0], fill_mode=rr.components.FillMode.Solid),
+    rr.Transform3D(axis_length=10),
+)
+
+for deg in range(46):
+    step += 1
+    rr.set_time_sequence("step", step)
+
+    rad = truncated_radians(deg * 4)
+    # TODO(cmc): update_fields
```
emilk, line 28: Link to the issue for partial updates in Python

docs/snippets/all/archetypes/transform3d_partial_updates.rs
```diff
@@ -0,0 +1,67 @@
...
+    rec.log(
+        "box",
+        &[
+            &rerun::Boxes3D::from_half_sizes([(4.0, 2.0, 1.0)])
+                .with_fill_mode(rerun::FillMode::Solid) as &dyn rerun::AsComponents,
+            &rerun::Transform3D::default().with_axis_length(10.0),
+        ],
+    )?;
+
+    for deg in 0..=45 {
```
emilk, line 18:
```suggestion
    // Update just the rotation:
    for deg in 0..=45 {
```

docs/snippets/all/archetypes/transform3d_partial_updates.rs
```diff
@@ -0,0 +1,67 @@
...
+        rec.log(
+            "box",
+            &rerun::Transform3D::update_fields().with_rotation(rerun::RotationAxisAngle::new(
+                [0.0, 1.0, 0.0],
+                rerun::Angle::from_radians(rad),
+            )),
+        )?;
+    }
+
+    for t in 0..=50 {
```
emilk, line 32:
```suggestion
    // Update just the translation:
    for t in 0..=50 {
```

docs/snippets/all/archetypes/transform3d_partial_updates.rs
```diff
@@ -0,0 +1,67 @@
...
+        step += 1;
+        rec.set_time_sequence("step", step);
+        rec.log(
+            "box",
+            &rerun::Transform3D::update_fields().with_translation([0.0, 0.0, t as f32 / 10.0]),
+        )?;
+    }
+
+    for deg in 0..=45 {
```
emilk, line 41:
```suggestion
    // Update just the rotation:
    for deg in 0..=45 {
```

docs/snippets/all/archetypes/transform3d_partial_updates.rs
```diff
@@ -0,0 +1,67 @@
...
+        rec.log(
+            "box",
+            &rerun::Transform3D::update_fields().with_rotation(rerun::RotationAxisAngle::new(
+                [0.0, 1.0, 0.0],
+                rerun::Angle::from_radians(rad),
+            )),
+        )?;
+    }
+
+    step += 1;
```
emilk, line 55:
```suggestion
    // Clear all transform fields, and set an axis length?
    step += 1;
```

docs/snippets/all/archetypes/transform3d_partial_updates.rs
```diff
@@ -0,0 +1,67 @@
...
+    rec.set_time_sequence("step", step);
+    rec.log(
+        "box",
+        &rerun::Transform3D::clear_fields().with_axis_length(15.0),
```
emilk, line 59: Argueblt we should just demonste one thing here (`clear_fields`) and skip the `with_axis_length`

docs/snippets/snippets.toml
```diff
@@ -271,6 +271,10 @@ quick_start = [ # These examples don't have exactly the same implementation.
 "archetypes/transform3d_hierarchy" = [ # Uses a lot of trigonometry which is surprisingly easy to get the same on Rust & C++, but not on Python/Numpy
   "py",
 ]
+"archetypes/transform3d_partial_updates" = [
+  "cpp", # TODO(cmc): remove once C++ partial updates APIs have shipped
+  "py", # TODO(cmc): remove once Python partial updates APIs have shipped
```
emilk, line 276: Link to the issues

emilk (approved):
Nice to have snippets for these, but I feel the code is maybe focusing a bit much on producing a nice/smooth animation, distracting from the actual meat of it (the partial updates). I'd like for our users to be able to open a snippet and right away find what they need, without too much extra fluff.

An easy fix would be to remove the `step` timeline (we already have `log_tick`, after all)

docs/snippets/all/archetypes/transform3d_partial_updates.rs
teh-cmc: It's a feature to show both: people need to know they don't need to (and shouldn't) log multiple rows in order to clear all data and reset one or more specific fields (and similarly, we need to make sure in the roundtrips that this indeed results in a single row for all languages).

PR #8697, opened 2025-01-15
Port all rust blueprint archetypes to eager serialization & update API

Wumpf:
### Related

### What

What it says on the tin!
Commit by commit - first commit does all the easy ones, followed by the trickier ones (just two)

Tested by...
* mess with tensor
* mess with time series in plots example
* run `docs/snippets/all/views/timeseries.py` snippet (uses explicit time series)


teh-cmc:
@rerun-bot full-check

emilk (approved):

PR #8700, opened 2025-01-15
Port `TransportChunk` to `arrow-rs`

emilk:
* Part of #3741 
* [x] Tested that it does not regress https://github.com/rerun-io/rerun/issues/8668

This makes `TransportChunk` a wrapper around an arrow `RecordBatch`.

### Future work
* Remove `TransportChunk` and replace it with an extension trait on `RecordBatch`
* Simplify the dataframe API to always return a full `RecordBatch` (adding a schema to the rows is basically free now)


emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

crates/store/re_dataframe/src/query.rs
```diff
@@ -1260,33 +1260,29 @@ impl<E: StorageEngineLike> QueryHandle<E> {
         Some(selected_arrays)
     }
 
-    /// Calls [`Self::next_row`] and wraps the result in a [`RecordBatch`].
-    ///
-    /// Only use this if you absolutely need a [`RecordBatch`] as this adds a lot of allocation
-    /// overhead.
```
teh-cmc, line 1266: That comment still applies.

teh-cmc (approved):

PR #8703, opened 2025-01-15
Port over most Rust types to use eager serialization

Wumpf:
### Related

* Large chunk of https://github.com/rerun-io/rerun/issues/7245
* Similar to https://github.com/rerun-io/rerun/pull/8697
   * but not as extensive, every time something came up that wasn't immediately obvious I dropped out of porting 

### What

A sweep through all the easy-to-port types in Rust

* [x] passes main ci.

---

Left are now:
* AnnotationContext
* AssetVideo
* Asset3D
* Boxes2D
* Boxes3D
* Image
* Mesh3D
* Pinhole
* ViewCoordinates

(copying that list into the issue)


Wumpf:
@rerun-bot full-check

crates/viewer/re_data_ui/src/instance_path.rs
```diff
@@ -278,16 +278,17 @@ fn preview_if_image_ui(
...
-        buffer: components::ImageBuffer,
-        format: components::ImageFormat
-    );
-    static_assert_struct_has_fields!(
-        archetypes::SegmentationImage,
-        buffer: components::ImageBuffer,
-        format: components::ImageFormat
-    );
+    // TODO(#7245): How to check this with eager serialized archetypes?
+    // static_assert_struct_has_fields!(
```
emilk, line 282: We can still check that is has the fields, but not their types.

However, you can replace it with a runtime-check instead (in debug builds), using the reflection API

emilk (approved):

crates/viewer/re_data_ui/src/instance_path.rs
Wumpf: did the real thing now :)
`debug_assert_archetype_has_components`

PR #8705, opened 2025-01-16
Add support for asynchronous decoding of rrd stream

zehiko:
### What

Add a streaming decoder for ``LogMsg``s.  

I've commonalized some of the logic between ``StreamingDecoder`` and the ``Decoder``, but there is still some repetition that comes from the fact we're working with a ``std::io::Read`` in one case and the buffer of bytes in the other. 
Bigger annoyance is some state keeping in the ``StreamingDecoder`` that comes from the fact that input file can be corrupted or it can contain multiple concatenated files. Open to suggestions for a simpler approach.

### Testing done
* unit tests
* integration with ReDap and using it for fetching rrd files from the object store


crates/store/re_log_encoding/src/decoder/streaming.rs
```diff
@@ -0,0 +1,371 @@
...
+pub struct StreamingDecoder<R: AsyncBufRead> {
+    version: CrateVersion,
+    options: EncodingOptions,
+    reader: R,
+    // buffer used for uncompressing data. This is a tiny optimization
+    // to (potentially) avoid allocation for each (compressed) message
+    uncompressed: Vec<u8>,
+    // internal buffer for unprocessed bytes
+    unprocessed_bytes: BytesMut,
+    // flag to indicate if we're expecting more data to be read.
```
teh-cmc, line 26: No reason not to make these comments actual docstrings so they actually show up in all the tools.

crates/store/re_log_encoding/src/decoder/streaming.rs
```diff
@@ -0,0 +1,371 @@
...
+            uncompressed: Vec::new(),
+            unprocessed_bytes: BytesMut::new(),
+            expect_more_data: false,
+        })
+    }
+
+    fn peek_file_header(data: &[u8]) -> bool {
+        let mut read = std::io::Cursor::new(data);
+        FileHeader::decode(&mut read).is_ok()
+    }
```
teh-cmc, line 54:
```suggestion
    /// Returns true if `data` can be successfully decoded into a `FileHeader`.
    fn peek_file_header(data: &[u8]) -> bool {
        let mut read = std::io::Cursor::new(data);
        FileHeader::decode(&mut read).is_ok()
    }

File: crates/store/re_log_encoding/src/decoder/streaming.rs [hunk 3]
```diff
@@ -0,0 +1,371 @@
...
+                                        DecodeError::MsgPack(err),
+                                    )));
+                                }
+                            }
+                        }
+                        super::MessageHeader::EndOfStream => return std::task::Poll::Ready(None),
+                    }
+                }
+                crate::Serializer::Protobuf => {
+                    let header_size = std::mem::size_of::<file::MessageHeader>();
```
teh-cmc, line 204: I haven't followed the recent developments on the protobuf file format, and this is not new to this PR, but seeing a `mem_size_of` on a struct type that uses the Rust ABI makes me both very uneasy and 99% there's a nasty bug lurking somewhere.

teh-cmc (approved):
I actually kinda love it.

I'm sure we could find myriad of ways to express this using higher level constructs instead, e.g. return a `impl Stream` from a freestanding func that uses using async/await expressions rather than a low-level, manual implementation of `Stream` with `task:Poll` but... I really appreciate the lack of magic here. It's very readable, very straightforward and puts the numerous edge cases in your face.

I also like that this enforces `AsyncBufRead` rather than expect an `AsyncRead` and wrap it in a `BufReader` unconditionally, that way we're sure the caller does the right thing _and_ we won't have multiple layers of buffering going unnoticed.

Sure, it's a pretty big method, but it reads like a book. I'll take that any day vs. a whack-a-mole of async insanity spread over 10 files.

10/10 would pin again

crates/store/re_log_encoding/src/decoder/streaming.rs
jprochazk: In this case it's 2x u64 fields, which is not problematic unless changed. Until this point, the encoding/decoding was fully local to the same file where it's defined, and the compiler would yell at you if you add/remove fields or change their types, pointing out all the places that need to be changed.

crates/store/re_log_encoding/src/codec/file/mod.rs
```diff
@@ -48,6 +48,11 @@ impl MessageHeader {
         let mut buf = [0; std::mem::size_of::<Self>()];
         data.read_exact(&mut buf)?;
 
+        Self::from_bytes(&buf)
+    }
+
+    #[cfg(feature = "decoder")]
+    pub fn from_bytes(buf: &[u8]) -> Result<Self, crate::decoder::DecodeError> {
```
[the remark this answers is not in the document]
zehiko: Agree. Also, thinking about this, I don't think this should be public, ``Decoder`` (and the new ``StreamingDecoder``) should be part of ``file`` module (which should probably be named ``rrd``.  if no one objects, I will do that as a follow up to not make this one to big. 

Will document and add the checks regardless.

crates/store/re_log_encoding/src/lib.rs
```diff
@@ -217,23 +217,28 @@ impl MessageHeader {
...
         read.read_exact(&mut buffer)
             .map_err(decoder::DecodeError::Read)?;
 
-        if u32_from_le_slice(&buffer[0..4]) == 0 && u32_from_le_slice(&buffer[4..]) == 0 {
-            Ok(Self::EndOfStream)
+        Ok(Self::from_bytes(&buffer))
+    }
+
+    #[cfg(feature = "decoder")]
+    pub fn from_bytes(data: &[u8]) -> Self {
```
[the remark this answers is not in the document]
zehiko: agree. the only reason I won't make it unsafe as I plan to move it so it no longer needs to be public. Also, if we're indeed dropping MsgPack support on the floor soon, this one is going away.

PR #8706, opened 2025-01-16
Filter entities in the UI (part 4): Add entity filtering in the blueprint tree

abey79:
### Related

- Part of #8586
- Part of a series of PR:
  - #8645
  - #8652
  - #8654
  - #8672
  - #8706
  - #8728
  - #8795
  - #8863
  - #8886
  - #8933

### What

This PR implements entity filtering in the blueprint tree.

In important design principle applied here is that the filtering applies only on the _visible_ part of items. This has two major consequence:
- The prefix of the data result entity path that matches the view origin is not considered in matching, because it is also not displayed in the tree. (That applies to non-projected data results.)
- The sub-tree defined by the view origin is ignored when matching projected entities, because that sub-tree is not displayed either (`$origin` placeholder in the tree).

Snapshot tests for this is coming in the next PR

https://github.com/user-attachments/assets/42e41fcf-7486-428c-a242-edf3a3b7493f

### DIFF
--- a/crates/viewer/re_context_menu/src/collapse_expand.rs
+++ b/crates/viewer/re_context_menu/src/collapse_expand.rs
@@ -20,7 +20,6 @@ pub fn collapse_expand_container(
                 .container(*container_id)
                 .set_open(ctx.egui_ctx, expand),
 
-            // IMPORTANT: don't call process_view() here, or the scope information would be lost
             Contents::View(view_id) => collapse_expand_view(ctx, view_id, scope, expand),
         }
 


crates/viewer/re_context_menu/src/collapse_expand.rs
```diff
@@ -0,0 +1,88 @@
...
+    scope: CollapseScope,
+    expand: bool,
+) {
+    blueprint.visit_contents_in_container(container_id, &mut |contents, _| {
+        match contents {
+            Contents::Container(container_id) => scope
+                .container(*container_id)
+                .set_open(ctx.egui_ctx, expand),
+
+            // IMPORTANT: don't call process_view() here, or the scope information would be lost
```
Wumpf, line 23: now that this is moved out it's not clear immediately to which `process_view` this refers to

Wumpf (approved):
looking great, love those new tests as well
the nˆ2 behavior in the search is a bit concerning but we talked about this already (more refactors following which may be able to address this along the way)

crates/viewer/re_context_menu/src/collapse_expand.rs
abey79: Dead ~code~ comment from an intermediate refactoring with this pr. Deleted.

PR #8717, opened 2025-01-16
Move the blueprint saving logic to `re_viewer_blueprint`

abey79:
### What

By just reshuffling things around, I _think_ this makes `re_viewer_blueprint` self-contained enough to be used from tests. This will be attempted in #8706


Wumpf (approved):
didn't diff the moved thing but I don't have any grudge against the concept this moving from there to there so lgtm :)

PR #8725, opened 2025-01-17
Port component storage to `arrow-rs`

emilk:
* This is a huge piece of https://github.com/rerun-io/rerun/issues/3741

This makes our store run 100% on arrow-rs.

arrow2 is now relegated to the margins:
* IPC
* `re_types_builder`
* Some legacy methods on `Loggable`, `Archetype` etc


crates/store/re_chunk_store/tests/snapshots/formatting__format_chunk_store.snap
```diff
@@ -1,23 +1,21 @@
...
 expression: "format!(\"{:200}\", store)"
-snapshot_kind: text
 ---
 ChunkStore {
     id: test_id
     config: ChunkStoreConfig { enable_changelog: true, chunk_max_bytes: 393216, chunk_max_rows: 4096, chunk_max_rows_if_unsorted: 1024 }
     stats: {
         num_chunks: 1
-        total_size_bytes: 1.3 KiB
+        total_size_bytes: 1.8 KiB
```
emilk, line 10: This should be fixed when we update arrow and can run `.shrink_to_fit`

crates/store/re_dataframe/src/query.rs
```diff
@@ -1230,7 +1226,21 @@ impl<E: StorageEngineLike> QueryHandle<E> {
     #[inline]
     pub fn next_row_batch(&self) -> Option<ArrowRecordBatch> {
         let row = self.next_row()?;
-        ArrowRecordBatch::try_new(self.schema().clone(), row).ok()
+        match ArrowRecordBatch::try_new_with_options(
+            self.schema().clone(),
+            row,
+            &arrow::array::RecordBatchOptions::new().with_row_count(Some(1)),
```
emilk, line 1232: Explicitly setting row-count to one means it works even when there are no columns (e.g. due to heavy filtering)

crates/store/re_dataframe/src/query.rs
```diff
@@ -1230,7 +1226,21 @@ impl<E: StorageEngineLike> QueryHandle<E> {
...
+        ) {
+            Ok(batch) => Some(batch),
+            Err(err) => {
+                if cfg!(debug_assertions) {
+                    panic!("Failed to create record batch: {err}");
+                } else {
+                    re_log::error_once!("Failed to create record batch: {err}");
+                    None
```
emilk, line 1240: Better than the previous silent error

crates/store/re_types_core/src/loggable_batch.rs
```diff
@@ -43,13 +43,18 @@ fn assert_loggablebatch_object_safe() {
...
-            .map_err(|err| err.into())
+            arrow::buffer::OffsetBuffer::from_lengths(std::iter::repeat(1).take(array.len()));
+        let field = match array.data_type() {
+            arrow::datatypes::DataType::List(field) => field.clone(),
+            _ => unreachable!(
+                "{} should always be a list array",
+                self.descriptor().full_name()
+            ),
+        };
+        ArrowListArray::try_new(field, offsets, array, None).map_err(|err| err.into())
```
emilk, line 57: Is this right? 🤔

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

crates/store/re_chunk_store/tests/snapshots/formatting__format_chunk_store.snap
emilk: Actually, adding `.shrink_to_fit` to `concat` didn't help

crates/store/re_dataframe/src/query.rs
teh-cmc: This comment should be in the code..!

crates/store/re_log_types/src/arrow_msg.rs
```diff
@@ -10,7 +10,9 @@ use arrow::array::RecordBatch as ArrowRecordBatch;
 use crate::TimePoint;
 
 // TODO(#3741): Remove once <https://github.com/apache/arrow-rs/issues/6803> is released
+#[cfg(feature = "serde")]
```
teh-cmc, line 13: What does `serde` has to do with Arrow..? :thinking:

crates/utils/re_arrow_util/src/arrow_util.rs
```diff
@@ -38,6 +32,13 @@ pub fn into_arrow_ref(array: impl Array + 'static) -> ArrayRef {
     std::sync::Arc::new(array)
 }
 
+/// Returns an iterator with the lengths of the offsets
+pub fn offsets_lengths(
+    offsets: &arrow::buffer::OffsetBuffer<i32>,
+) -> impl Iterator<Item = usize> + '_ {
+    offsets.windows(2).map(|w| (w[1] - w[0]) as usize)
```
teh-cmc, line 39: This should be a saturating sub or this will end up crashing the app on malformed inputs.

crates/utils/re_arrow_util/src/arrow_util.rs
```diff
@@ -248,8 +250,9 @@ pub fn new_list_array_of_empties(child_datatype: &DataType, len: usize) -> ListA
 /// Returns an error if the arrays don't share the exact same datatype.
 pub fn concat_arrays(arrays: &[&dyn Array]) -> arrow::error::Result<ArrayRef> {
     #[allow(clippy::disallowed_methods)] // that's the whole point
-    arrow::compute::concat(arrays)
-    // TODO(#3741): call .shrink_to_fit on the result
+    let mut array = arrow::compute::concat(arrays)?;
+    array.shrink_to_fit(); // VERY IMPORTANT! https://github.com/rerun-io/rerun/issues/7222
+    Ok(array)
```
teh-cmc, line 255: Reminder: everything using a compute kernel in this file requires the same treatment.

crates/viewer/re_view_spatial/src/visualizers/arrows2d.rs
```diff
@@ -264,7 +264,9 @@ impl VisualizerSystem for Arrows2DVisualizer {
                                 .map_or(&[], |class_ids| bytemuck::cast_slice(class_ids)),
                             keypoint_ids: keypoint_ids
                                 .map_or(&[], |keypoint_ids| bytemuck::cast_slice(keypoint_ids)),
-                            show_labels: show_labels.unwrap_or_default().get(0).map(Into::into),
+                            show_labels: show_labels
+                                .map(|b| !b.is_empty() && b.value(0))
+                                .map(Into::into),
```
teh-cmc, line 269: I find the way this was expressed before much more readable... is there no fallible `get()` method in arrow1?

teh-cmc (approved):

crates/viewer/re_view_spatial/src/visualizers/arrows2d.rs
emilk: Sadly not. All getters in arrow panics rather than return an `Option`.

crates/store/re_log_types/src/arrow_msg.rs
emilk: Nothing - just a drive-by-fix, similar to this:
* https://github.com/rerun-io/rerun/pull/8734

crates/store/re_dataframe/src/query.rs
emilk: Right you are!

emilk:
@rerun-bot full-check

emilk:
I'm running all the benchmarks locally, comparing with 0.21.0, to see that we don't have any big regressions

emilk:
# Results of benchmarks
Comparing this PR with the 0.21.0 tag (dd025f1), notable changes (red is regression, green improvement):

* Encoding log messages 🔻 20-30%. Likely unrelated to arrow, since we still use arrow2 for this
* Decoding log message 🔻 15% 
* Decode message bundles 🌲 60%
* Generate messages 🔻 25% 
* `arrow_batch_strings2/query` 🔻 10% 
* `single_chunks/split_all` 🌲 20%
* `vec_deque/remove_range/prefilled/middle` 🌲 55%  (maybe because of new std/compiler?)

So, some wins, some losses, but the losses are not huge.

PR #8731, opened 2025-01-19
 Remove use of `into_arrow2_buffer` in generated code

emilk:
* Part of #3741

### DIFF
--- a/crates/build/re_types_builder/src/codegen/rust/serializer.rs
+++ b/crates/build/re_types_builder/src/codegen/rust/serializer.rs
@@ -585,46 +585,40 @@ fn quote_arrow_field_serializer(
         DataType::Utf8 => {
             // NOTE: We need values for all slots, regardless of what the validity says,
             // hence `unwrap_or_default`.
-            let (quoted_transparent_mapping, quoted_transparent_length) =
-                if inner_is_arrow_transparent {
-                    let inner_obj = inner_obj.as_ref().unwrap();
-                    let is_tuple_struct = is_tuple_struct_from_obj(inner_obj);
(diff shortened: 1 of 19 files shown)


emilk:
@rerun-bot full-check

crates/build/re_types_builder/src/codegen/rust/serializer.rs
```diff
@@ -585,66 +585,74 @@ fn quote_arrow_field_serializer(
...
+
+                    let mut buffer_builder = arrow::array::builder::BufferBuilder::<u8>::new(capacity);
                     // NOTE: Flattening to remove the guaranteed layer of nullability: we don't care
                     // about it while building the backing buffer since it's all offsets driven.
-                    let inner_data: arrow::buffer::Buffer =
-                        #data_src.into_iter().flatten() #quoted_transparent_mapping.collect();
+                    for data in #data_src.iter().flatten() {
+                        buffer_builder.append_slice(data #quoted_member_accessor.as_bytes());
+                    }
+                    let inner_data: arrow::buffer::Buffer = buffer_builder.finish();
```
teh-cmc, line 639: I'm surprised we can't just use a `collect()` here and below, `Buffer` seems to implement `FromIterator?

teh-cmc (approved):

crates/build/re_types_builder/src/codegen/rust/serializer.rs
emilk: We can, but I wanted to pre-allocate (reserve) the exact amount of bytes we need.

I'll add a comment explaining this

Issue #8723, opened 2025-01-17
`arkit_scenes` Transforms are broken


jleibs:
![Image](https://github.com/user-attachments/assets/c6f02124-dcef-4f87-8b58-b9f2a0828b47)

Wumpf:
did a lil bit of debugging already, but not all the way yet: It looks like the new `TransformCache` has issues with statically logged transforms!

Wumpf:
have a path towards fixing this, will do that on Monday

PR #8743, opened 2025-01-20
Update wgpu to 24.0.0

Wumpf:
With minimal changes along the way.

* [x] `pixi run js-build-base` still works
    * [x] @jprochazk : We've been through this before, but we didn't fix up the comments on wasm-bindgen: It says "make sure the build script still works". Well it runs, but how do I make sure the output makes sense? I remember us diffing a bunch of things, but not what it was. As-is that comment just makes me nervous without being actionable
* [x] tested a bunch of examples on native MacOS
* [x] tested web

All glitches I found are documented in https://github.com/rerun-io/rerun/issues/8723 and are on `main` as well.


crates/build/re_dev_tools/src/build_web_viewer/lib.rs
```diff
@@ -173,10 +173,7 @@ pub fn build(
         match target {
             Target::Browser => bindgen_cmd.no_modules(true)?.typescript(false),
             Target::Module => bindgen_cmd.no_modules(false)?.typescript(true),
-            Target::NoModulesBase => bindgen_cmd
```
jprochazk, line 176: Why remove `reference_types`?

jprochazk:
> I remember us diffing a bunch of things, but not what it was. As-is that comment just makes me nervous without being actionable

It works if the build script text replacements actually have an effect on the built Wasm. I realise that there's no easy way to check that, ideally it would be a pixi command that builds the Wasm, applies our patch, and diffs the two to confirm the changes were applied.

Something like:
```
mkdir rerun_js/web-viewer/temp
pixi run cargo run -p re_dev_tools -- build-web-viewer --release -g \
  --target no-modules-base \
  --no-default-features \
  --features grpc,map_view \
  -o rerun_js/web-viewer/temp
pixi run js-build-base
git diff --no-index rerun_js/web-viewer/re_viewer.js rerun_js/web-viewer/temp/re_viewer.js
rm -rf rerun_js/web-viewer/temp
```

crates/build/re_dev_tools/src/build_web_viewer/lib.rs
Wumpf: it's deprecated with the explanation that it's now automatically detected

jprochazk (approved):

Issue #6330, opened 2024-05-15
Forbid `unwrap()` in our code


Artxiom:
@Wumpf 
Follow up for #3408.
After the initial refactor there are still some crates left which use unwrap:

- [x] re_data_store
- [x] re_dev_tools
- [x] re_entity_db
- [x] re_format_arrow
- [x] re_log_types
- [x] re_query
- [ ] re_renderer
- [ ] re_renderer_examples/*
- [ ] re_sdk
- [ ] re_space_view_spatial
- [ ] re_space_view_tensor
- [ ] re_space_view_time_series
- [ ] re_time_panel
- [ ] re_types_builder
- [ ] re_types_core
- [ ] re_viewer
- [ ] re_viewport
- [ ] re_ws_comms
- [ ] snippets

They are all marked with:
```
// TODO(#3408): remove unwrap()
#![allow(clippy::unwrap_used)]
```

ilyaaay:
Hello. I want to take this issue (I started with re-renderer)

Artxiom:
@ilyaaay I realized it's not a huge amount of changes yet so I pushed my branch [here](https://github.com/Artxiom/rerun/tree/6330-refactor-unwrap-re_renderer).

Feel free to pick it up - I will be super busy probably until beginning of next year.
Just please `@mention` me somewhere when you do so, so I don't pick it up accidentally again.

riccardocinti:
Hi everyone. I'd like to contribute to this issue and I started working on the top/re_sdk and the store/re_sdk_comms crates.

emilk:
Closed in favor of the original issue:
* https://github.com/rerun-io/rerun/issues/3408

PR #8745, opened 2025-01-20
fix clippy::unwrap_used in re_viewer with Option::expect

brody4hire:
### Related

Part of: #6330

### What

- remove allow directive & TODO comment from `crates/viewer/re_viewer/src/lib.rs`
- use `. unwrap_or_else` with panic in 1 case
- use focused `#[allow(clippy::unwrap_used)]` directives in limited, focused cases
- use `.expect` with descriptive messages in all other cases

(no changelog updates)

__General rationale:__

Most of the cases using `.unwrap` seem to need these to work. I think the best solution is to use `.expect` with a clear message in most of these cases.

__Sanity checks done:__

These succeed for me with no warnings:

- `cargo clippy -p re_viewer --all-features`
- `cargo build -p re_viewer --all-features`
- `cargo fmt` does not introduce inconsistent formatting of the updated code

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -1150,11 +1150,11 @@ impl App {
 
                 self.egui_debug_panel_ui(ui);
 
-                // TODO(andreas): store the re_renderer somewhere else.
-                let egui_renderer = {
-                    let render_state = frame.wgpu_render_state().unwrap();
-                    &mut render_state.renderer.write()
-                };
+                let egui_renderer = &mut frame
+                    .wgpu_render_state()
+                    .expect("Failed to get frame render state")
+                    .renderer
+                    .write();
 
                 if let Some(render_ctx) = egui_renderer
                     .callback_resources
@@ -1836,7 +1836,10 @@ impl eframe::App for App {
         }
 
         // Temporarily take the `StoreHub` out of the Viewer so it doesn't interfere with mutability
-        let mut store_hub = self.store_hub.take().unwrap();
+        let mut store_hub = self
+            .store_hub
+            .take()
+            .expect("Failed to take store hub from the Viewer");
 
         #[cfg(not(target_arch = "wasm32"))]
         if let Some(resolution_in_points) = self.startup_options.resolution_in_points.take() {
@@ -1884,14 +1887,16 @@ impl eframe::App for App {
         let gpu_resource_stats = {
             re_tracing::profile_scope!("gpu_resource_stats");
 
-            let egui_renderer = {
-                let render_state = frame.wgpu_render_state().unwrap();
-                &mut render_state.renderer.read()
-            };
+            let egui_renderer = frame
+                .wgpu_render_state()
+                .expect("Failed to get frame render state")
+                .renderer
+                .read();
+
             let render_ctx = egui_renderer
                 .callback_resources
                 .get::<re_renderer::RenderContext>()
-                .unwrap();
+                .expect("Failed to get render context");
 
             // Query statistics before begin_frame as this might be more accurate if there's resources that we recreate every frame.
             render_ctx.gpu_resources.statistics()
@@ -1910,14 +1915,16 @@ impl eframe::App for App {
         self.purge_memory_if_needed(&mut store_hub);
 
         {
-            let egui_renderer = {
-                let render_state = frame.wgpu_render_state().unwrap();
-                &mut render_state.renderer.read()
-            };
+            let egui_renderer = frame
+                .wgpu_render_state()
+                .expect("Failed to get frame render state")
+                .renderer
+                .read();
+
             let render_ctx = egui_renderer
                 .callback_resources
                 .get::<re_renderer::RenderContext>()
-                .unwrap();
+                .expect("Failed to get render context");
 
             // We haven't called `begin_frame` at this point, so pretend we did and add one to the active frame index.
             let renderer_active_frame_idx = render_ctx.active_frame_idx().wrapping_add(1);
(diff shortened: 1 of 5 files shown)


Wumpf (approved):
thank you!!

crates/viewer/re_viewer/src/app_blueprint.rs
```diff
@@ -219,10 +220,12 @@ impl AppBlueprint<'_> {
 
             let timepoint = store_ctx.blueprint_timepoint_for_writes();
 
+            // Can only fail if we have the wrong number of instances for the component, and we don't
```
brody4hire, line 223: ~~need to clean this comment as well~~

---

I would like to remove this comment & `#[allow(clippy::unwrap_used)]` in favor of using `.expect()` here as well.

crates/viewer/re_viewer/src/app_blueprint.rs
```diff
@@ -197,11 +197,11 @@ pub fn setup_welcome_screen_blueprint(welcome_screen_blueprint: &mut EntityDb) {
         let chunk = Chunk::builder(entity_path)
             .with_component_batches(RowId::new(), timepoint, [&value as &dyn ComponentBatch])
             .build()
-            .unwrap(); // Can only fail if we have the wrong number of instances for the component, and we don't
+            .expect("Failed to build chunk - incorrect number of instances for the component (internal error)");
 
         welcome_screen_blueprint
             .add_chunk(&Arc::new(chunk))
-            .unwrap(); // Can only fail if we have the wrong number of instances for the component, and we don't
+            .expect("Failed to add new chunk for welcome screen - internal error");
```
brody4hire, line 204: nit: improve consistency here

```suggestion
            .expect("Failed to add new chunk for welcome screen (internal error)");
```

brody4hire (commented):
@Wumpf I have shamelessly thrown away your cleanup work in favor of using `.expect("... (internal error)")` calls (I am assuming no offense will be taken). I think this should both reduce the number of lines in most cases and clarify the code a little better.

I think this should be ready now. Please do feel free to apply any other cleanup you think may be needed & be sure to add yourself as a co-author.

crates/viewer/re_viewer/src/ui/welcome_screen/welcome_section.rs
```diff
@@ -56,7 +56,9 @@ pub(super) fn welcome_section_ui(ui: &mut egui::Ui) {
         if ui
             .button(
                 egui::RichText::new("Go to documentation →")
-                    .color(egui::Color32::from_hex("#60A0FF").expect("this color is valid"))
+                    .color(
+                        egui::Color32::from_hex("#60A0FF").expect("color failure (internal error)"),
```
emilk, line 60: As noted in other PR: This should be using the `egui::hex_color!` macro instead: https://docs.rs/egui/latest/egui/macro.hex_color.html

crates/viewer/re_viewer/src/screenshotter.rs
```diff
@@ -89,7 +89,7 @@ impl Screenshotter {
             let h = image.height() as _;
             let image =
                 image::RgbaImage::from_raw(w, h, bytemuck::pod_collect_to_vec(&image.pixels))
-                    .expect("Failed to create image");
+                    .expect("Failed to create image (internal error)");
```
emilk, line 92: I like the "internal error" message for our users, but I think we should add that to the panic handler of `re_crash_handler` instead, since any panic is an internal error.

crates/viewer/re_viewer/src/background_tasks.rs
```diff
@@ -60,7 +60,10 @@ impl BackgroundTasks {
         self.promises
             .remove(name.as_ref())
             .and_then(|promise| match promise.try_take() {
-                Ok(any) => Some(*any.downcast::<T>().unwrap()),
+                Ok(any) => Some(
+                    *any.downcast::<T>()
+                        .unwrap_or_else(|err| panic!("downcast failure: {err:?}")),
```
brody4hire, line 65: I think this is an example where a panic is triggered _not_ by an internal error but by an issue with the calling code.

crates/viewer/re_viewer/src/screenshotter.rs
brody4hire: I would agree in most but not all cases, I think this points to a counter-example as I commented below: https://github.com/rerun-io/rerun/pull/8745/files#r1928954634

crates/viewer/re_viewer/src/background_tasks.rs
brody4hire: I am thinking it may be ideal to update the API to return a promise or result with a promise so that the calling code can poll it without the need for this downcast. We could add a separate API function to do this without breaking any API users. WDYT?

crates/viewer/re_viewer/src/screenshotter.rs
brody4hire: How do you think we should proceed with this?

emilk: A panic _should_ only happen because of a bug in the rerun codebase, which is what I would call an internal error. If there are panics that could happen because of user error, we should fix that separatedly.

crates/viewer/re_viewer/src/app.rs
```diff
@@ -1836,7 +1836,10 @@ impl eframe::App for App {
         }
 
         // Temporarily take the `StoreHub` out of the Viewer so it doesn't interfere with mutability
-        let mut store_hub = self.store_hub.take().unwrap();
+        let mut store_hub = self
+            .store_hub
+            .take()
+            .expect("Failed to take store hub");
```
brody4hire, line 1842: D'oh `cargo fmt` flattens this now that I shortened the expect message, which is not good consistency with the other updates I made and not what I really wanted to happen here - argh

brody4hire:
I have rebased (2x), removed "` (internal error)`" from the end of the expect messages, and updated an error message to keep `cargo fmt` from flatting an update. Please let me know if anything else is needed from me on this.

emilk (approved):
Thank you for this!

PR #8764, opened 2025-01-21
Tagged columnar updates: Rust

### LINKED ISSUE #7167: Support columns with variable partitions in Rust's `send_column` api
Both Python & C++ APIs have ways of partitioning a component column arbitrarily to e.g. log point clouds in a single `send_columns` call.
Rust however, right now only supports partitions of length 1 for each element (i.e. at each timestep you end up with an array of length 1) which is still great for the usecase of sending e.g. `Scalar`, but not the full picture.

Note that the lower level `send_chunk` API that the Rust recording stream comes with has none such restrictions (and is what is used by Python and C++).

### LINKED ISSUE #8753: Tagged columnar updates: Rust
Image

teh-cmc:
I ended up with something slightly different than the original design, because it turned out to really not feel that great in practice.
In particular, it required duplicating _a lot_ of generated code for little (no?) added value.

In short: `.columns` becomes a vanilla method rather than a static method.

```rust 
// Prepare a point cloud that evolves over 5 timesteps, changing the number of points in the process.
let times = TimeColumn::new_seconds("time", 10..15);

#[rustfmt::skip]
let positions = [
    [1.0, 0.0, 1.0], [0.5, 0.5, 2.0],
    [1.5, -0.5, 1.5], [1.0, 1.0, 2.5], [-0.5, 1.5, 1.0], [-1.5, 0.0, 2.0],
    [2.0, 0.0, 2.0], [1.5, -1.5, 3.0], [0.0, -2.0, 2.5], [1.0, -1.0, 3.5],
    [-2.0, 0.0, 2.0], [-1.5, 1.5, 3.0], [-1.0, 1.0, 3.5],
    [1.0, -1.0, 1.0], [2.0, -2.0, 2.0], [3.0, -1.0, 3.0], [2.0, 0.0, 4.0],
];

// At each timestep, all points in the cloud share the same but changing color and radius.
let colors = [0xFF0000FF, 0x00FF00FF, 0x0000FFFF, 0xFFFF00FF, 0x00FFFFFF];
let radii = [0.05, 0.01, 0.2, 0.1, 0.3];

// Partition our data as expected across the 5 timesteps.
let position = rerun::Points3D::update_fields()
    .with_positions(positions)
    .columns([2, 4, 4, 3, 4])?;
let color_and_radius = rerun::Points3D::update_fields()
    .with_colors(colors)
    .with_radii(radii)
    .columns([1, 1, 1, 1, 1])?;

rec.send_columns_v2("points", [times], position.chain(color_and_radius))?;
```

* Fixes #8753 
* Fixes #7167


crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -1030,6 +1030,41 @@ impl RecordingStream {
...
+    ///
+    /// Unlike the regular `log` API, which is row-oriented, this API lets you submit the data
+    /// in a columnar form. The lengths of all of the [`TimeColumn`] and the component columns
+    /// must match. All data that occurs at the same index across the different time and components
+    /// arrays will act as a single logical row.
+    ///
+    /// Note that this API ignores any stateful time set on the log stream via the
+    /// [`Self::set_timepoint`]/[`Self::set_time_nanos`]/etc. APIs.
+    /// Furthermore, this will _not_ inject the default timelines `log_tick` and `log_time` timeline columns.
+    pub fn send_columns_v2(
```
teh-cmc, line 1043: I'll rename this to `send_columns` and remove all the legacy stuff in a follow-up, I don't want the diffs to be unnecessarily hard to follow.

teh-cmc:
Ha, interestingly this won't work with `Scalar` as-is because we need to generate an extra `with_scalars` (plural) method.

I'll auto-generate plural methods for these instances in a follow-up.

teh-cmc:
Related: I also intend to introduce a `unit_columns()` method in a follow-up, for the very common case (i.e. `.columns(std::take::repeat(1).iter(n))`).

crates/store/re_types_core/src/lib.rs
```diff
@@ -187,3 +187,35 @@ pub fn try_serialize_field<C: crate::Component>(
         }
     }
 }
+
+/// Internal serialization helper for code-generated archetypes.
+///
+/// This generates a correctly sized [`SerializedComponentColumn`] for a given indicator, where the
+/// specified `num_rows` value represents the number of rows in the column.
+#[doc(hidden)] // public so we can access it from re_types too
+pub fn indicator_column<A: Archetype>(
```
teh-cmc, line 196: This can be greatly simplified, and much cheaper as well -> indicators are straight null arrays these days, no lists involved.

teh-cmc: That does mean making `SerializedComponentColumn` more complicated just for the sake of indicators, i.e. legacy...

Annoying, but there isn't really a choice here.

Wumpf: hmm yeah we'd have the same problem in C++ as well (in Python all the indicator handling happens on the Rust side anyways iirc)

Is it really worth it?

Wumpf: (re Cpp: `ComponentColumn` there has a `from_indicators` which does pretty much what we're doing here)

teh-cmc: It's more complicated because `SerializedComponentColumn` would have to carry an erased Array instead, and have two paths for everything. I don't like it.

I'm gonna investigate an alternative: have `send_columns_v2` automagically craft indicator columns as required on the fly. I don't like it either but objectively it is far less invasive.

docs/snippets/all/archetypes/points3d_send_columns.rs
```diff
@@ -0,0 +1,37 @@
...
+    let times = TimeColumn::new_seconds("time", 10..15);
+
+    #[rustfmt::skip]
+    let positions = || [
+        vec![[1.0, 0.0, 1.0], [0.5, 0.5, 2.0]],
+        vec![[1.5, -0.5, 1.5], [1.0, 1.0, 2.5], [-0.5, 1.5, 1.0], [-1.5, 0.0, 2.0]],
+        vec![[2.0, 0.0, 2.0], [1.5, -1.5, 3.0], [0.0, -2.0, 2.5], [1.0, -1.0, 3.5]],
+        vec![[-2.0, 0.0, 2.0], [-1.5, 1.5, 3.0], [-1.0, 1.0, 3.5]],
+        vec![[1.0, -1.0, 1.0], [2.0, -2.0, 2.0], [3.0, -1.0, 3.0], [2.0, 0.0, 4.0]],
+    ]
```
Wumpf, line 18: I don't like the way of writing the example this way much. `send_columns` is all about having flat data to begin with and if your data were to look like this then it wouldn't be a good fit.
This is why back then I went with a flat position list in C++ and then provided the partitions separately
https://github.com/rerun-io/rerun/pull/8764/files#diff-4b7c039da83394b4869f80ddbaacbcf555c70653dc990c6a1609ccb8eb3a326cR36

docs/snippets/snippets.toml
```diff
@@ -234,6 +231,10 @@ quick_start = [ # These examples don't have exactly the same implementation.
 "archetypes/points3d_partial_updates" = [ # TODO(cmc): revert once C++ has tagged partial updates
   "cpp",
 ]
+"archetypes/points3d_send_columns" = [
+  "py", # TODO(#8752): needs tagged columnar APIs
+  "cpp", # TODO(#8754): needs tagged columnar APIs
+]
```
Wumpf, line 237: isn't the result already/still the same? How are they differing now?

Wumpf (approved):
very happy you found an ergonomic way without generating tons of `with_` methods again. I like the new-new design you landed at :) 👍

docs/snippets/snippets.toml
teh-cmc: Untagged in py and cpp

crates/store/re_types_core/src/lib.rs
teh-cmc: Oh god, this actually is an already existing problem, indicators are wrapped in a `ListArray` when put into chunks (probably to avoid adding a ton of complexity for a dead feature) :facepalm:.

That means:
* Indicators take 4 bytes per row as of today (the outer list-array's offsets).
* What I did is actually the right thing to do, and it makes me very sad.

Man, indicators really gotta go...

teh-cmc: Goes without saying, this will have to happen a yet another follow up.

teh-cmc: https://github.com/rerun-io/rerun/issues/8768

docs/snippets/all/archetypes/points3d_send_columns.rs
teh-cmc: Ha yeah I didn't notice it was flat, the way this was formatted made me think it wasn't. I'll update.

PR #8765, opened 2025-01-21
Handle static transforms in the transform cache & add tests for these cases

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/8723

### What

The transform cache previously ignored static transforms by accident, thus breaking some of our example scenes.

After some back and forth I decided to handle them transparently: In order to avoid additional lookups, we make sure that we add timepoints for the occurence of static chunks to our cached transform `BTreeMap`s.
The drawback of that is that..
* whenever a static transform changes there's a ripple through all existing cached transforms
* whenever someone queries a timeline we haven't seen yet, we have to present them the static transforms

The logic for all of this isn't terribly complicated, but very easy to get wrong. So this comes with a ton of new tests.

### DIFF
--- a/crates/viewer/re_view_spatial/src/transform_cache.rs
+++ b/crates/viewer/re_view_spatial/src/transform_cache.rs
@@ -39,7 +39,8 @@ pub struct TransformCacheStoreSubscriber {
     /// All components related to pinholes (i.e. [`components::PinholeProjection`] and [`components::ViewCoordinates`]).
     pinhole_components: IntSet<ComponentName>,
 
-    per_timeline: HashMap<Timeline, CachedTransformsPerTimeline>,
+    per_timeline: HashMap<Timeline, CachedTransformsForTimeline>,
+    static_timeline: CachedTransformsForTimeline,
 }
 
 impl Default for TransformCacheStoreSubscriber {
@@ -64,6 +65,10 @@ impl Default for TransformCacheStoreSubscriber {
             .collect(),
 
             per_timeline: Default::default(),
+            static_timeline: CachedTransformsForTimeline {
+                invalidated_transforms: Default::default(),
+                per_entity: Default::default(),
+            },
         }
     }
 }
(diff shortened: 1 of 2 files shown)


crates/viewer/re_view_spatial/src/transform_cache.rs
```diff
@@ -201,55 +256,104 @@ impl TransformCacheStoreSubscriber {
...
+                .static_timeline
+                .per_entity
+                .entry(entity_path.hash())
+                // There have never been any static or non-static transforms for this entity, therefore there's nothing to pass
+                // into that entity for static transforms.
+                .or_insert_with(|| TransformsForEntity::new(None, None));
+
+            // Technically this doesn't query static components but rather just what's at the beginning of the tick timeline,
+            // but it's the most convenient way to the data we want.
+            let query = LatestAtQuery::new(Timeline::log_tick(), TimeInt::MIN);
```
abey79, line 291: Could this break if the user decided to somehow `rr.disable_timeline("log_tick")`?

crates/viewer/re_view_spatial/src/transform_cache.rs
```diff
@@ -281,60 +382,135 @@ impl TransformCacheStoreSubscriber {
...
+        } else {
+            for (timeline, time_column) in event.diff.chunk.timelines() {
+                let per_timeline = self.per_timeline.entry(*timeline).or_insert_with(|| {
+                    CachedTransformsForTimeline::new(timeline, &self.static_timeline)
                 });
+
+                // All of these require complex latest-at queries that would require a lot more context,
+                // are fairly expensive, and may depend on other components that may come in at the same time.
+                // (we could inject that here, but it's not entirely straight forward).
+                // So instead, we note down that the caches is invalidated for the given entity & times.
```
abey79, line 461:
```suggestion
// So instead, we note down that the caches are invalidated for the given entity & times.
```

crates/viewer/re_view_spatial/src/transform_cache.rs
```diff
@@ -281,60 +382,135 @@ impl TransformCacheStoreSubscriber {
...
+
+                // This invalidates any time _after_ the first event in this chunk.
+                // (e.g. if a rotation is added prior to translations later on,
+                // then the resulting transforms at those translations changes as well for latest-at queries)
```
abey79, line 465:
```suggestion
// then the resulting transforms at those translations change as well for latest-at queries)
```

abey79 (approved):
My eyes aren't accustomed to this area of the code but this looks good to me. I really like the test setup too 👍🏻

crates/viewer/re_view_spatial/src/transform_cache.rs
Wumpf: no, if anything it would make this more accurate since then we're sure that this can only yield static components. As is, this could actually give us non-static results. I think what that means is that I shouldn't use `log_tick` but something more made up...

abey79: Makes sense!

Issue #8749, opened 2025-01-21
Make the visualised state of data results more explicit in the UI


abey79:
Consider this view (adapted from the arkit example):

<img width="1744" alt="Image" src="https://github.com/user-attachments/assets/2c096fb3-6d51-4c68-af0f-5bf5f51233dd" />

<br/>

The blueprint tree shows a whole bunch of data results: `/description` and all the `/world/annotations/box-…` ones.

The latter appear because they contain a `Text` component, which is visualisable in the text view:

<img width="532" alt="Image" src="https://github.com/user-attachments/assets/bb627b6b-2698-4c88-aacf-f80ef3542f0e" />

<br/><br/>

Why, then, only the `/description` data result is being displayed? The full story about this is involved, but boils down to: "visualisers for the `box...` entities have not been automatically created, because these entities do not have a `TextDocumentIndicator` logged (or, in the future, a `TextDocument` archetype tag)".

### Background

The determination of a view's content is akin to a funnel:

1) The starting point is the entities present in the store on the current timeline and at the current timepoints.
2) The entity path filter is then applied, which produces an often reduced set of candidate entities. (This is what the "Matches 16 entities" note in the first screenshot refers to.)
3) Then, visualisability is determined, and only **visualisable entities** are kept. An entity is visualisable for a given view type if it contains at least all the _required_ components for one the the archetype visualisers of that view type. In the case of the text view, the only required component of the only visualiser is `Text`.
4) The visualisable entities are then filtered for "applicability". Each view class is allowed to decide which entity are **applicable entities** based on arbitrary criterions. This is rarely used. One example is the bar chart view, for which only 1D tensor are applicable. Entities that are visualisable and applicable become **data result** and are displayed in the blueprint tree. In the text view above, there is no applicability filter, and the screenshot shows the 8 visualisable (and thus applicable) entities.
5) Finally, each of the visualisable entities which _also_ contains a matching indicator component is deemed an **indicated entity**. By default, the viewer spawns visualisers _only_ for indicated entities. In the case of the screenshot above, only the `/description` entity is indicated, which explains why it is the only one that is displayed.

Notes:
- Visualiser can be created and deleted manually. So the _actual_ set of visible entities in a given view can be an arbitrary subset of applicable entities. Only the _default_ set of visible entities corresponds to the indicated entities. For the purpose of this issue, let's call the set of entities for which a visualiser is active the **displayed entities** (but we must find something else for the UI).
[...]

PR #8767, opened 2025-01-21
Rename "applicable" to "maybe visualizable" and adjust various docs accordingly

Wumpf:
### Related

* came up in the same discussion as https://github.com/rerun-io/rerun/issues/8749
    * but isn't _really_ related to that otherwise ;-) 

### What

See title.
Hopefully this makes things less confusing in the future. Went through all the mentions of "applicable" & "applicability" in all *.rs files and tried to set things right.

Small nuance that goes a bit further: the custom bit of "maybe visualizable" set (former applicability) is now called `DataBasedVisualizabilityFilter` - the "maybe" part here is dropped (but documented) since it shouldn't really matter! What it does is what it says now: It's a hermetic function that can only take into account component data but no other context.


abey79 (approved):
Thanks!

Issue #8768, opened 2025-01-21
Log indicators in their own dedicated chunks


teh-cmc:
Indicators are meant to go, but this won't happen overnight. In fact, they will be here for the foreseeable future.
In the meantime, it would be nice to reduce their negative impact, if possible (i.e. with very minimal efforts).

There are two problems with them that I want to focus on in this issue.

First problem: on the standard row-oriented log path, indicators actually consume 4 bytes of space *per row* in each chunk.
The reason for that is that chunks are, at the moment, only capable of carrying `ListArray`s around (we want to be able to use native-typed arrays as columns in the future, but this is not a small change and this won't happen for a long time).
Corollary: each time you log an indicator (i.e. each time you log an archetype), the indicator's `NullArray` gets wrapped in a `ListArray`, and that list must have N offset values where N in the number of rows in the chunk. The null array itself doesn't take space (well it takes a fixed 4 bytes), but the outer list's offsets do.

Second problem: all of that is just as true on the column-oriented path, and becomes even more relevant today with new `send_columns` APIs that actually enforce that the appropriate indicators are logged.

We should investigate the possibility of always logging indicators into their own little chunk, on both the row- and column- oriented paths.
* On the row-oriented path, that means splitting them in a separate chunk if e.g. the main chunk has more than a predefined threshold of rows.
* On the column-oriented path, that means *always* logging them in a separate chunk.

PR #8769, opened 2025-01-21
Make tagged columnar updates work with mono-components too

teh-cmc:
* Generate plural methods even for mono-components, so they can be used with `send_columns`.
* Update `scalar_send_columns` snippet to use new columnar APIs.

```rust
const STEPS: i64 = 64;

let times = TimeColumn::new_sequence("step", 0..STEPS);
let scalars = (0..STEPS).map(|step| (step as f64 / 10.0).sin());

rec.send_columns_v2(
    "scalars",
    [times],
    rerun::Scalar::update_fields()
        .with_many_scalar(scalars)
        .columns(std::iter::repeat(1).take(STEPS as _))?
        .filter(|column| !column.descriptor.component_name.contains("Indicator")),
)?;
```

* Follow-up to #8753

### DIFF
--- a/docs/snippets/all/archetypes/scalar_send_columns.rs
+++ b/docs/snippets/all/archetypes/scalar_send_columns.rs
@@ -1,24 +1,23 @@
 //! Use the `send_columns` API to send scalars over time in a single call.
 
-use rerun::components::Scalar;
 use rerun::TimeColumn;
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let rec = rerun::RecordingStreamBuilder::new("rerun_example_send_columns").spawn()?;
+    let rec = rerun::RecordingStreamBuilder::new("rerun_example_scalar_send_columns").spawn()?;
 
-    // Native time / scalars
-    let timeline_values = (0..64).collect::<Vec<_>>();
+    const STEPS: i64 = 64;
 
-    let scalar_data: Vec<f64> = timeline_values
-        .iter()
-        .map(|step| (*step as f64 / 10.0).sin())
-        .collect();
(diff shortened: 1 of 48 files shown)


docs/snippets/all/archetypes/scalar_send_columns.rs
```diff
@@ -1,24 +1,23 @@
...
-    let scalar_data: Vec<Scalar> = scalar_data.into_iter().map(Into::into).collect();
-
-    rec.send_columns("scalars", [timeline_values], [&scalar_data as _])?;
+    rec.send_columns_v2(
+        "scalars",
+        [times],
+        rerun::Scalar::update_fields()
+            .with_many_scalar(scalars)
+            .columns(std::iter::repeat(1).take(STEPS as _))?
+            .filter(|column| !column.descriptor.component_name.contains("Indicator")),
```
teh-cmc, line 19: The other languages do not log an indicator here, and for good reasons: indicators in a scalar chunk have huge overhead, and are completely useless as far as I can tell (i.e. they aren't needed for heuristics).

In a follow-up, I will introduce a hardcoded exception so that `Scalar::columns` just doesn't generate it.
* See also #8768

teh-cmc:
At least half of the generated code seems to be for blueprint archetypes. We don't need it, I'll add a hardcoded exception to ignore blueprint archetypes tomorrow.

Wumpf (approved):
lgtm!
We should be able to use this on the `image_send_column` example. Can you have a look there as well?

teh-cmc:
> lgtm! We should be able to use this on the `image_send_column` example. Can you have a look there as well?

Ah? will do

PR #8771, opened 2025-01-22
Columnar APIs: do not autogenerate indicator for `Scalar`

teh-cmc:
See https://github.com/rerun-io/rerun/pull/8769#pullrequestreview-2565378138 for rationale:
image

* Related: #8768 
* Follow-up to #8753


Wumpf (approved):
rather brute force, but yeah everything else would be overengineering 🤷

Issue #7081, opened 2024-08-06
`rerun::Collection` borrows data too eagerly, making it very easy to cause segfaults & read of invalid data


facontidavide:
**Describe the bug**

I have a class like this:

```c++
rerun::Boxes2D createBoxes(....) {
  std::vector<rerun::Position2D> centers;
  std::vector<rerun::datatypes::Vec2D> sizes;
  // fill the vectors
  return rerun::Boxes2D::from_centers_and_sizes(centers, sizes);
}
```

Unfortunately, the result seems to be corrupted, ore specifically the centers (half-sizes, seems to be ok).
I believe that the problem is that the `Collection` is borrowing the reference to the centers, instead of taking owership.

The following code **fixes** the issue:

```c++
rerun::Boxes2D createBoxes(....) {
  std::vector<rerun::Position2D> centers;
  std::vector<rerun::datatypes::Vec2D> sizes;
  // fill the vectors
  auto boxes rerun::Boxes2D::from_sizes(sizes);
  boxes.centers = rerun::Collection<rerun::Position2D>::take_ownership(std::move(centers));
  return boxes;
}
```

**To Reproduce**
Try the function mentioned above and then print the centers:

```c++
auto boxes = risc::createBoxes(....);
for (const auto& center : *(boxes.centers)) {
  std::cout << center.x() << " / " << center.y() << "\n";
}
```

**Rerun version**
0.17.0

Wumpf:
This is pretty much by design and document as such (although maybe not easy to find :/).
`rerun::Collection` tries to borrow whenever possible to avoid copies. The way this works is that everything that isn't created from a temporary (r value reference) will go down the `borrow` path of the collection.

That said, I'm starting to regret this design. It's great to have the _option_ of borrowing data, but we do it too aggressively. This isn't Rust where we can do this kind of operation safely, instead this makes everything very easy to fail as you've experienced.
I think we should change `rerun::Container` to borrow only on explicit codepath. This will be a breaking change for the container adapter system which right now is based on the premise of behaving like this with arbitrary types - instead we need to restyle this to have adapters that interact with explicit `borrow` methods.

Somewhat related to this recent change [here](https://github.com/rerun-io/rerun/pull/7055) where I made it easier to do explicit borrow & ownership taking.

facontidavide:
Yes, I can understand that you are trying to import good ideas from Rust, but without the compiler support, you CANNOT do that.

[...]

PR #8727, opened 2025-01-17
Introduce eager serialization & update/clear APIs in the C++ SDK

Wumpf:
### Related

* https://github.com/rerun-io/rerun/issues/3381
* https://github.com/rerun-io/rerun/issues/7245
* once fully done will obsoletes https://github.com/rerun-io/rerun/issues/7081

### What

Introduces both eager serialization (serialization upon Archetype construction) & update/clear APIs to C++ codegen.
Just like on Rust, we're doing a case by case opt-in right now because there's some extensions & tests that need individual changes. Ported so far are:

* all blueprint types (they're unused right now)
* annotation_context
* arrows2d
* arrows3d
* points2d
* points3d
* transform3d

everything else still needs porting over

---
* [x] passes main ci

Issue #6832, opened 2024-07-09
Split Tensor component into several archetypes


Wumpf:
* Split out of https://github.com/rerun-io/rerun/issues/6388

Related to:
* https://github.com/rerun-io/rerun/issues/2341

We generate archetypes and components for all tensor variants (TensorF32, TensorU8, etc) and make sure they share the same Visualizer:

```
archetype TensorU8 {
    buffer: BufferU8,
    
    // One of these
    shape: TensorShape,
    shape: Vec<TensorDimension>,
}

component BufferU8 {
    data: [u8],
}

archetype TensorF32 {
    buffer: BufferF32,

    // One of these
    shape: TensorShape,
    shape: Vec<TensorDimension>,
}

component BufferF32 {
    data: [f32],
}
```

* mechanics of same-visualizer are a bit unclear. Have visualizer just listen to several indicators / archetypes? Breaks 1:1 relationship that we were striving for. Can revisit later?
* this _will_ break some "use this tensor like an image" cases that we allow today. Mitigate only as far as meaningful

Impact on Mesh's texture:
Log an Image archetype at the same spot instead.

Detailed rationale (via @jleibs on https://github.com/rerun-io/rerun/issues/6388#issuecomment-2134003885):

Most of the choices for working with tensors fall into one of 4 categories.

### Typed buffer, multiple data-types (the proposal)

Pros:
 - When processing a chunk the raw arrow data is much easier to work with
 - Opportunity to align with the official arrow spec for tensor representation
 - Aligns with our long-term direction of wanting to have multiple types and datatype conversions

Cons:
 - Multi-datatype representation means we must either proliferate typed components or introduce datatype conversions.

The current hypothesis is that proliferating types is a known challenge and can be mostly automated with a mixture of code-gen and some helper code, whereas datatype conversions is an unknown challenge.

Still this puts us on a pathway where once we support multi-typed components, we mostly delete a bunch of code and everything gets simpler. Any type conversions move from visualizer-space to data-query-space, but the types and arrow representations we work with don't actually need to change.

### Untyped buffer with type-id

Pros
- Avoids arrow unions while maintaining a single datatype.

Cons
- Forces arrow users to do annoying user-space datatype casting.
- Doesn't align with our long-term goals

### Typed buffer with union

Pros
- Status quo. Already works.

Cons
- Forces arrow users to do annoying poorly supported union operations when loading or reading tensors.

emilk:
[...]

PR #8779, opened 2025-01-22
Make all C++ archetypes eager serialized & provide generated update/clear APIs

Wumpf:
### Related

* follow-up of https://github.com/rerun-io/rerun/pull/8727
* finishes C++ portion of https://github.com/rerun-io/rerun/issues/7245
* finishes C++ portion of https://github.com/rerun-io/rerun/issues/3381
* fixes https://github.com/rerun-io/rerun/issues/7081
    * passing an archetype is no longer a problem
* associated codegen attribute removed in https://github.com/rerun-io/rerun/pull/8780 

### What

(see title)
On every archetype in C++ you can now use `update_fields` & `clear_fields`!

Commit by commit recommended - most of them are trivial! Outstanding are:
* `tensor`
     * Because of #6832 `with_dim_names` became quite hard. This is covered by a snippet (which I tested manually) but I also added a test now
* `video_asset` 
    * needed to extract video data & media type from the serialized arrow data for a utility method. Nothing too scary, and no large copies involved
* `clear`
    * needed a new cpp attribute to suppress the `=default` constructor as this one has its own (previous codegen wasn't as rigorous with default ctors!) 

---

* [x] pass complete snippet comparison run locally
* [x] please the different compiler gods and their very individual preferences 
* [x] succeed `main` ci


Wumpf:
nope, only pushing a new commit fixed it.

teh-cmc (approved):

Issue #4027, opened 2023-10-26
GCC false-positives for Werror=maybe_uninitialized


jleibs:
## Overview
We have a handful of places where we need to disabled `-Wmaybe-uninitialized` primarily related to usage of std::optional.

It seems that move-assigning to the std::optional can lead to situations where gcc thinks we are using uninitialized-memory because the memory in the optional hasn't been initialized (nor does it need to be).

This issue is to track those usages and maybe review or better work around them in the future.

### Archetypes
- Because archetypes contain optional ComponentBatches ...

Wumpf:
The problem with `with_component` on archetypes went away with eager serialization. Those suppressions got removed in

* https://github.com/rerun-io/rerun/pull/8780

Others are still around. Might need another pass.

PR #8780, opened 2025-01-22
Remove `attr.cpp.archetype_eager` codegen attribute

Wumpf:
### Related

* follow-up to https://github.com/rerun-io/rerun/pull/8779

### What

Makes eager serialization & update/clear apis the norm in C++ with no opt out.

Also removes no longer needed warning suppression on builder methods!

* [x] said warning suppression is actually no longer needed on any compiler we test


teh-cmc (approved):
So... is https://github.com/rerun-io/rerun/issues/4027 fixed then? :thinking: 
* https://github.com/rerun-io/rerun/issues/4027

Wumpf:
We have unfortunately still other cases of that. I added a few again in the last, but yeah overall a lot less, gonna update that ticket a bit

Issue #8760, opened 2025-01-21
Implement parallel untagged + tagged indexing


teh-cmc:
We need the store to maintain two indices at write time: the legacy untagged one, and a new tagged one.

The legacy untagged index is necessary to avoid all sorts of UB until everything has been ported to tagged data.

The new tagged index will allow us to start porting things incrementally. Of course we also need accompanying tagged query APIs.

teh-cmc:
I knew I was forgetting yet another subtle complication: maintaining an untagged index is not enough, you can still end up in a situation where a single Chunk has both untagged and tagged data for a component, and no index is gonna save you there.
This is what's happening here (see attached screenshot). if i had to guess, this is because a runtime blueprint write ends up compacted in a pre-existing, tagged blueprint chunk, and now the resulting chunk is both tagged and untagged for that component.

Once we're done with all the API updates on the SDK side, it shouldn't ever be possible for a user to end up in that situation *when working in new recording*, so that end is covered.
That leaves A) runtime blueprint writes and B) user writes to a pre-existing, legacy recording. Obviously the correct fix for blueprint writes is to port all of them to tagged APIs, but A) that will not happen for 0.22 and B) that doesn't take care of the other problem.

I see two possible avenues here:
* Modify the compaction logic so that when tagged data is compacted with untagged data of the same component, we merge them together and keep the tags going forward. This should be well-specified with the current data model, and helps with propagating tags going forward.
* Modify the compaction logic so that tagged and untagged data of the same component is never compacted together.

PR #8782, opened 2025-01-22
Forbid compaction of tagged+untagged data for a single component

teh-cmc:
This implements the second solution described in https://github.com/rerun-io/rerun/issues/8760#issuecomment-2607749175:
> I knew I was forgetting yet another subtle complication: maintaining an untagged index is not enough, you can still end up in a situation where a single Chunk has both untagged and tagged data for a component, and no index is gonna save you there.
> This is what's happening here (see attached screenshot). if i had to guess, this is because a runtime blueprint write ends up compacted in a pre-existing, tagged blueprint chunk, and now the resulting chunk is both tagged and untagged for that component.
> 
> Once we're done with all the API updates on the SDK side, it shouldn't ever be possible for a user to end up in that situation *when working in new recording*, so that end is covered.
> That leaves A) runtime blueprint writes and B) user writes to a pre-existing, legacy recording. Obviously the correct fix for blueprint writes is to port all of them to tagged APIs, but A) that will not happen for 0.22 and B) that doesn't take care of the other problem.
> 
> I see two possible avenues here:
> * Modify the compaction logic so that when tagged data is compacted with untagged data of the same component, we merge them together and keep the tags going forward. This should be well-specified with the current data model, and helps with propagating tags going forward.
> * Modify the compaction logic so that tagged and untagged data of the same component is never compacted together.

I can confirm this fixes the "changing view origin crashes in debug" issue.

* Part of #8760

### DIFF
--- a/crates/store/re_chunk/src/merge.rs
+++ b/crates/store/re_chunk/src/merge.rs
@@ -250,6 +250,35 @@ impl Chunk {
             })
     }
 
+    /// Returns `true` if both chunks share the same descriptors for the components that
+    /// _they have in common_.
+    #[inline]
+    pub fn same_descriptors(&self, rhs: &Self) -> bool {
+        self.components.values().all(|lhs_per_desc| {
+            if lhs_per_desc.len() > 1 {
+                // If it's already in UB land, we might as well give up immediately.
+                return false;
+            }
+
+            lhs_per_desc.keys().all(|lhs_desc| {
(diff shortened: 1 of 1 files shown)


teh-cmc:
Before:
image

After:
image
image

teh-cmc:
@rerun-bot full-check

abey79 (approved):
Makse sense to me 👍🏻

Issue #3381, opened 2023-09-20
Allow archetypes to clear and/or update existing components


emilk:
### Backstory
* https://github.com/rerun-io/rerun/issues/3279
* https://github.com/rerun-io/rerun/issues/3341

What should the following code do:

```rs
log("points", Points3D::new(positions).with_colors(…));
…
log("points", Points3D::new(new_positions));
```

Should the new positions inherit the colors of the previous log call?

That would be quite non-intuitive.

Related to this: how does one update just the colors of a point cloud?

### Design
We've all converged on the following API:

``` rs
// We do NOT implement `Default` for archetypes.
// users need to explcitly use `clear` or `update`
struct Points3D {
   positions: Optional<Vec<Vec3>>,
   color: Optional<Vec<Color>>,
   radii: Optional<Vec<f32>>,
}

impl Points3D {
    /// Clear all components of this archetype to the empty list.
    fn clear() -> Self {
        Self {
            positions: Some(vec![]),
            color: Some(vec![]),
            radii: Some(vec![]),
        }
    }
    
    /// Create a new `Point3d` point cloud with some positions,
    /// and with emtpy lists for all other archetypes.
    fn new(positions: Vec<Vec3>) -> Self {
        Self {
            positions: Some(positions),
            ..Self::clear()
        }
    }
    
    /// An empty `Point3d` point cloud, with `None` for each component.
    fn update() -> Self {
        Self {
            positions: None,
            color: None,
            radii: None,
        }
    }
}

fn main() {
    // Log a point cloud:
    log("points", Points3D::new(…).with_colors(…));
    
    log("points", Clear{}); // will clear all components of all archetypes
    log("points", Points3D::clear()); // will clear this archetype of the entity
    log("points", Points3D::new(positions).with_colors(…)); // overwrites everything
    log("points", Points3D::update().with_colors(…)); // overwrites only colors

    // This is also possible, though a somewhat weird way to write it:
    log("points", Points3D::clear().with_colors(…)); // overwrites everything
}
```

Python:
```py
log("points", Clear()); # will clear all components of all archetypes
log("points", Points3D.clear()); # will clear this archetype of the entity
log("points", Points3D(positions, colors=[…], radii=None); # overwrites everything, even radii
log("points", Points3D.update(colors=[…], radii=None)); # overwrites only colors and radii
```

### Necessary changes
* All components are `Options`, even the required ones
* When querying, an empty lists should be treated the same as a missing components
[...]

PR #8785, opened 2025-01-23
Make  `Box`/`AssetVideo`/`ViewCoordinates`/`Asset3D` eager serialized in Rust

Wumpf:
### Related

* Part of
    * https://github.com/rerun-io/rerun/issues/3381
    * https://github.com/rerun-io/rerun/issues/7245 

### What

View coordinates comes with some slight api breakage since we can no longer offer constants for the different orientations.

After this, the only archetypes that haven't migrated are:
* Image
* Mesh3D
* Pinhole

### DIFF
--- a/crates/store/re_types/src/archetypes/asset_video_ext.rs
+++ b/crates/store/re_types/src/archetypes/asset_video_ext.rs
@@ -32,30 +32,52 @@ impl AssetVideo {
     #[inline]
     pub fn from_file_contents(contents: Vec<u8>, media_type: Option<impl Into<MediaType>>) -> Self {
         let media_type = media_type.map(Into::into);
-        let media_type = MediaType::or_guess_from_data(media_type, &contents);
-        Self {
-            blob: contents.into(),
-            media_type,
+        if let Some(media_type) = MediaType::or_guess_from_data(media_type, &contents) {
+            Self::new(contents).with_media_type(media_type)
+        } else {
+            Self::new(contents)
         }
     }
 
     /// Determines the presentation timestamps of all frames inside the video.
     ///
     /// Returned timestamps are in nanoseconds since start and are guaranteed to be monotonically increasing.
+    ///
+    /// Panics if the serialized blob data doesn't have the right datatype.
     #[cfg(feature = "video")]
     pub fn read_frame_timestamps_ns(&self) -> Result<Vec<i64>, re_video::VideoLoadError> {
+        use re_types_core::Loggable;
+
         re_tracing::profile_function!();
+
+        let Some(blob) = self.blob.as_ref() else {
+            return Ok(Vec::new());
+        };
+
+        // Grab blob data without a copy.
+        let blob_list_array = blob
+            .array
+            .as_any()
+            .downcast_ref::<arrow::array::ListArray>()
+            .expect("Video blob data is not a ListArray");
+        let blob_data = blob_list_array.values().to_data();
+        let blob_bytes = blob_data.buffer(0);
 
         let Some(media_type) = self
             .media_type
-            .clone()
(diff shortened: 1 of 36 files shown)


teh-cmc:
@rerun-bot full-check

crates/store/re_types/src/archetypes/asset_video_ext.rs
```diff
@@ -32,30 +32,52 @@ impl AssetVideo {
...
+        let Some(blob) = self.blob.as_ref() else {
+            return Ok(Vec::new());
+        };
+
+        // Grab blob data without a copy.
+        let blob_list_array = blob
+            .array
+            .as_any()
+            .downcast_ref::<arrow::array::ListArray>()
+            .expect("Video blob data is not a ListArray");
```
teh-cmc, line 62: Bit unfortunate to risk a panic in a method that already returns a result anyhow?

crates/viewer/re_view_spatial/src/mesh_loader.rs
```diff
@@ -1,14 +1,18 @@
...
-    archetypes::{Asset3D, Mesh3D},
-    components::{AlbedoFactor, MediaType},
-};
+use re_types::{archetypes::Mesh3D, components::MediaType};
 use re_viewer_context::{gpu_bridge::texture_creation_desc_from_color_image, ImageInfo};
 
 use crate::mesh_cache::AnyMesh;
 
+#[derive(Debug, Clone)]
+pub struct Asset3D<'a> {
```
teh-cmc, line 10: Can we name that `NativeAsset3D` or something? This will be quite confusing in the future.

docs/content/reference/migration/migration-0-22.md
```diff
@@ -25,3 +25,10 @@ It was no longer used for several releases, so we removed it.
 
 **Note**: although `rr.log_components()` is technically a public API, it is undocumented, and we discourage using it.
 For logging custom components, use [`rr.AnyValue`](https://ref.rerun.io/docs/python/main/common/custom_data/#rerun.AnyValues) and [`rr.AnyBatchValue`](https://ref.rerun.io/docs/python/main/common/custom_data/#rerun.AnyBatchValue).
+
+
+### Rust's `ViewCoordinates` archetype now has static methods instead of constants
+
+As part of the switch to "eager archetype serialization" (serialization of archetype components now occurs at time if archetype instantiation rather than logging), we can no longer offer constants
+for the `ViewCoordinates` archetype like `ViewCoordinates::RUB`.
```
teh-cmc, line 33:
```suggestion
As part of the switch to "eager archetype serialization" (serialization of archetype components now occurs at time of archetype instantiation rather than logging), we can no longer offer constants for the `ViewCoordinates` archetype like `ViewCoordinates::RUB`.

**Review by teh-cmc — APPROVED** [review]
neat

File: crates/store/re_types/src/archetypes/asset_video_ext.rs (hunk 1 above)
Comment (Wumpf) [reply crates/store/re_types/src/archetypes/asset_video_ext.rs:62]: agreed, but the problem is that `re_video::VideoLoadError` ofc knows nothing about any kind of arrow shenanigans we might be dealing with, which means I'd need to change the error type here or leak things into video load error

File: crates/viewer/re_view_spatial/src/mesh_loader.rs (hunk 2 above)
Comment (Wumpf) [reply crates/viewer/re_view_spatial/src/mesh_loader.rs:10]: `NativeAsset3D` sounds good to me

File: crates/store/re_types/src/archetypes/asset_video_ext.rs (hunk 1 above)
Comment (teh-cmc) [reply crates/store/re_types/src/archetypes/asset_video_ext.rs:62]: You can always have an escape hatch like:
```rust
    #[error(transparent)]
    Other(#[from] anyhow::Error),
```
:man_shrugging:

PR #8789, opened 2025-01-23
Make `Pinhole` archetype in Rust eager serialized

Wumpf:
### Related

* sister PR to
    * https://github.com/rerun-io/rerun/pull/8785
    * https://github.com/rerun-io/rerun/pull/8793
* Part of
    * https://github.com/rerun-io/rerun/issues/3381
    * https://github.com/rerun-io/rerun/issues/7245 

### What

Deprecates a bunch of methods on `Pinhole`, all of them still functional though through ad-hoc deserialization utilities.

In the Viewer (concrete, `re_spatial_view`) we don't use any of that and have now a new purpose-built type - I considered using the generated (`attr.rust.archetype_native`) type, but a custom built made a lot more sense for untangling the mess of adhoc queries we have around Pinhole. It might not look like it but the aspiration here was to not change how things work by much while having things better decoupled & cleaner.
Bonus: the odd resolution oracle we _already_ used is now visible on the pinhole visualizers's ui

### DIFF
--- a/crates/viewer/re_view_spatial/src/pinhole.rs
+++ b/crates/viewer/re_view_spatial/src/pinhole.rs
@@ -1,0 +1,125 @@
+use re_types::{archetypes, components};
+
+use crate::resolution_of_image_at;
+
+/// A pinhole camera model.
+///
+/// Corresponds roughly to the [`re_types::archetypes::Pinhole`] archetype, but uses `glam` types.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Pinhole {
+    pub image_from_camera: glam::Mat3,
+    pub resolution: glam::Vec2,
+}
+
+impl Pinhole {
+    /// Width/height ratio of the camera sensor.
+    #[inline]
+    pub fn aspect_ratio(&self) -> f32 {
+        self.resolution.x / self.resolution.y
+    }
+
+    /// Principal point of the pinhole camera,
+    /// i.e. the intersection of the optical axis and the image plane.
+    ///
+    /// [see definition of intrinsic matrix](https://en.wikipedia.org/wiki/Camera_resectioning#Intrinsic_parameters)
+    #[inline]
+    pub fn principal_point(&self) -> glam::Vec2 {
+        glam::vec2(
+            self.image_from_camera.col(2)[0],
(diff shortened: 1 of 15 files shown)


crates/viewer/re_view_spatial/src/pinhole.rs
```diff
@@ -0,0 +1,125 @@
...
+///
+/// Fallback provider will be used for everything but the projection itself.
+/// Does NOT take into account blueprint overrides, defaults and fallbacks.
+/// However, it will use the resolution of the image at the entity path if available.
+///
+/// If the projection isn't present, returns `None`.
+// TODO(andreas): Give this another pass and think about how we can remove this.
+// Being disconnected from the blueprint & fallbacks makes this a weird snowflake with unexpected behavior.
+// Also, figure out how this might actually relate to the transform cache.
+pub fn query_pinhole_and_view_coordinates_from_store_without_blueprint(
```
Wumpf, line 89: this is essentially what `query_pinhole_legacy` used to do. But with an EVEN scarier name!

crates/viewer/re_view_spatial/src/ui_2d.rs
```diff
@@ -153,11 +150,15 @@ impl SpatialView2D {
         // Note that we can't rely on the camera being part of scene.space_cameras since that requires
         // the camera to be added to the scene!
         //
-        // TODO(jleibs): Would be nice to use `query_pinhole` here, but we don't have a data-result or the other pieces
-        // necessary to properly handle overrides, defaults, or fallbacks. We don't actually use the image_plane_distance
-        // so it doesnt technically matter.
+        // TODO(jleibs, andreas): We don't have a data-result or the other pieces
+        // necessary to properly handle overrides, defaults, or fallbacks.
```
emilk, line 154: Is there maybe an issue for this? (pinhole-blueprint-overrides)

emilk (approved):
That was a big surgery

crates/viewer/re_view_spatial/src/ui_2d.rs
Wumpf: I'll link the transform override issue!

Issue #8777, opened 2025-01-22
Broken selection panel UI for (at least) custom data?


teh-cmc:
If you look at the recording resulting from this thing:
```py
rr.init("rerun_example_any_values_send_columns", spawn=True)

timestamps = np.arange(0, 64)

# Log two components, named "sin" and "cos", with the corresponding values
values = rr.AnyValues(
    sin=np.sin(timestamps / 10.0),
    cos=np.cos(timestamps / 10.0),
)

rr.send_columns(
    "/",
    times=[rr.TimeSequenceColumn("step", timestamps)],
    components=values.as_component_batches(),
)
```
what [you'll see is that on 0.21 it shows what you'd expect](https://rerun.io/viewer/version/0.21.0?url=https%3A%2F%2Fstatic.rerun.io%2Frrd%2F0.21.0%2Fmuh_floats_2797d2e679f2d2b4d55d0e57a332bef56d412eb2.rrd):
![Image](https://github.com/user-attachments/assets/2ba6f95f-0451-4e89-a6aa-54fed06a460e)

whereas [on latest `main` is not only doesn't show what you'd expect, what it shows is just plain wrong](https://rerun.io/viewer/version/main?url=https%3A%2F%2Fstatic.rerun.io%2Frrd%2F0.21.0%2Fmuh_floats_2797d2e679f2d2b4d55d0e57a332bef56d412eb2.rrd) (that's the size of the whole column, not this specific value at this specific timestamp):
![Image](https://github.com/user-attachments/assets/4bcf9ac1-627e-4742-ab94-eb32617d3cb9)

PR #8790, opened 2025-01-23
Fix display of compacted arrow-data

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/8777

The problem was using the wrong function to calculate the size of the buffer (using the size of the whole buffer, rather than just the part we are slicing).

I decided to also update the docs for `SizeBytes` and improve the size estimation of `Vec` and friends to take capacity into account.


teh-cmc (approved):
n.e.a.t.

Issue #8574, opened 2025-01-03
Links in markdown don't work


Wumpf:
try clicking any of the links e.g. in the description here https://app.rerun.io/?url=https%3A%2F%2Fapp.rerun.io%2Fversion%2F0.21.0%2Fexamples%2Fdetect_and_track_objects.rrd

Doesn't do anything!

Checked on web & native

jprochazk:
This is likely a regression introduced in 0.21, 0.20.3 works: https://app.rerun.io/version/0.20.3/index.html?url=https://app.rerun.io/version/0.20.3/examples/detect_and_track_objects.rrd

PR #8794, opened 2025-01-23
Fix clicking of links in markdown

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/8574
* Bug introduced in https://github.com/rerun-io/rerun/pull/8495

### What
The text document had a `response.interact(Sense::click())` call _on top_ of it, catching all clicks that should have ended up on the hyperlinks. This is now changed into setting `Sense::click` on the underlying `Ui` when creating it, meaning it will catch clicks _behind_ the hyperlinks, i.e. any click that doesn't interact with the document iteself.

### DIFF
--- a/crates/viewer/re_view_text_document/src/view_class.rs
+++ b/crates/viewer/re_view_text_document/src/view_class.rs
@@ -122,64 +122,22 @@ Displays text from a text component, as raw text or markdown."
         let state = state.downcast_mut::<TextDocumentViewState>()?;
         let text_document = system_output.view_systems.get::<TextDocumentSystem>()?;
 
-        let response = egui::Frame {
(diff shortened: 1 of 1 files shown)


abey79:
Also, there is a perma-horizontal scroll bar in the view, at least in the DNA example. This is not new though, same in 0.21.

image

emilk:
> However, clicking in the "void" no longer selects the text view like it use to do

it-works-on-my-computer

It works for me O.o

abey79:
> It works for me O.o

what the heck? I tried TWICE, and now it's working for me too. No idea what happened... I should take a day off 😞

abey79 (approved):

PR #8802, opened 2025-01-24
Post-eager serialization cleanup

Wumpf:
Now that all types are eager serialized, there was some things to clean-up.

* removed `attr.rust.archetype_eager` (always on)
* removed `attr.rust.archetype_native`, we only used it for container blueprint and it wasn't really necessary there (the "native type" was immediately destructured anyways 🤷 )
* some ticket mentions


teh-cmc (approved):
neat

Issue #8662, opened 2025-01-13
Only render events for streams that are visible in the time panel


abey79:
Querying and rendering the event view for streams in the time panel is, despite recent chunkification, quite slow. This is made worse by the "expand all" behaviour of the filter feature (https://github.com/rerun-io/rerun/pull/8654).

A likely optimisation is:
- to avoid rendering evens for everything that is above the visible part of the scroll area
- early out as soon as we're bellow the visible part of the scroll area

abey79:
Maybe something I can sneak in when I do the refactor this cycle.

emilk:
Is this a performance drain when there are many entities? What is the slow part of it? Do you have a profiler screenshot?

abey79:
The time panel is definitely a time hog when everything is uncollapsed. However, contrary to what I initially thought, we _already_ skip drawing the event when not in the visible area.

The "early-out-when-bottom-is reached" optimisation sound a bit stupid (things would get worse upon scrolling the list), it does make sense in the context of filter. Indeed, early during typing the query, everything that matches (which can be a lot) is shown uncollapsed. Early exiting would allow keeping the frame time reasonable during the early moment of the user typing the query.

### Collapse all

<img width="1518" alt="Image" src="https://github.com/user-attachments/assets/c5db8ced-9758-4d86-b06d-5be40ce2a11a" />

### Expand all

<img width="1518" alt="Image" src="https://github.com/user-attachments/assets/24b888cd-223b-4309-83c7-b09e14a67843" />

### Expand all profile

<img width="1950" alt="Image" src="https://github.com/user-attachments/assets/8755ff74-9a0b-45c7-9780-3fb393971677" />

abey79:
This issue is related and likely relevant:
- https://github.com/rerun-io/rerun/issues/8798

Issue #8233, opened 2024-11-27
Project: Performant visualization of scenes with large number of entities


teh-cmc:
## Context

We want the viewer to be able to scale to scenes with large numbers of entities.
This of course means visualizing these scenes, but also ingesting them in the first place.

This is blocked on a number of specific implementation issues, but put broadly: the work the viewer has to do to layout a scene more often than not grows linearly with the number of entities present in the _entire_ dataset.

There are only two ways to combat this:
* Break the linear growth where possible (e.g. only compute the data for the visible part of a list as opposed to the entire list).
* (Incremental) caching where possible (e.g. incrementally cache and update the transform hierarchy).

Of course in many cases, option 1 isn't even an option: if the user wants to visualize all entities in the scene, then somehow we have to make that fast.

Incremental caching of aggregated data (which is what the visualizers work with) is very hard, but will be a must in order to reach our performance goals. 

This issue is *not* about:
* Making the UX work for large number of entities (search, etc).

### Air Traffic example (`2h` dataset)

* Ingestion directly from the SDK should be bottlenecked by the SDK, always.
* Ingestion from file should be order of magnitude faster, in the single digit seconds realm.
* Should be able to visualize, without visible time range, at or close to 60FPS on any decent machine, on the web.
* Should be able to visualize, with infinite visible time range, to run at 60FPS on high-end machines, on native.

### Revy

Revy was infamously bottlenecked by the performance of many entities (game scenes have _a lot_ of them).
This is a good opportunity to revive that project, if we can make it happen.

* Should be able to ingest and visualize Bevy's "Alien Cake Addict" example in real-time, on a decent machine, on the web.

## Relevant material

Writings:
* https://github.com/rerun-io/rerun/issues/5974
* https://github.com/rerun-io/rerun/issues/8221
* TODO(cmc): link to updated range semantics proposal
* TODO(cmc): link to range-zipped caching proposal

PRs:
* https://github.com/rerun-io/rerun/pull/5449
* https://github.com/rerun-io/rerun/pull/8224

## Sub-issues

_To be replaced with actual sub-issues once/if we have access to them. We can also remove the `project-many-entities-perf` label at that point._

* https://github.com/rerun-io/rerun/issues/4298
* https://github.com/rerun-io/rerun/issues/7025
* https://github.com/rerun-io/rerun/issues/7084
* https://github.com/rerun-io/rerun/issues/7604
* https://github.com/rerun-io/rerun/issues/7678
[...]

PR #8808, opened 2025-01-24
Improve performance for Blueprint & Streams Panel for many entities

### LINKED ISSUE #8798: Optimise the blueprint tree panel rendering
The blueprint tree panel can take a sizeable amount of frame time when uncollapsed with lots of entities. For example, with the air traffic data example and 0.21.0 and just the 3D view it takes:
- 0.013 ms when the view is collapsed
- 3ms when the view is fully expanded
(see screenshots below)

One naive optimisation would be to early out of the entire blueprint tree function when we reached the bottom of the screen. However, that would prevent some side effect to happen. For example, we must reach the location of the "scroll to me" `Item` so we may set the scroll offset of the enclosing scroll area.

One promising avenue would be to early out _within_ `ListItem::show_*()` when offscreen. Since the height is fixed, `ListItem` could sub itself for a `ui.add_space(self.height)` (or a slightly more elaborate version of that to provide `egui::Responses`, etc.) instead of actually calling into the `ListItemContent` ui code.

Image

Image

abey79:
### Related

- Part of #8233
- Fixes #8798
- Part of #8662 (or fixes it?)

### What

This PR adds an option to `ListItem` to disable rendering offscreen. In that case, `ListItem` skips some code and in particular does not call `ListItemContent::ui()`. The full sizing computation is still run, so it does not affect layout.

Both the blueprint tree and the time panel now opt in to this optimisation.

DNM: chaineded to blueprint tree refactor to minimise conflicts

### Benchmark

This optimisation is only measurable in "worst case" scenario, such as fully uncollapsed blueprint tree and streams tree in the air traffic data example. In this case, we shave about 13-15ms off the frame time (90% of which in the time panel) on a M4 Max.

Before/after screenshot (with the view hidden for better frame time stability): (61.8 vs. 47.4ms)

image

image

### DIFF
--- a/crates/viewer/re_ui/src/list_item/list_item.rs
+++ b/crates/viewer/re_ui/src/list_item/list_item.rs
@@ -49,6 +49,7 @@ pub struct ListItem {
     force_background: Option<egui::Color32>,
     pub collapse_openness: Option<f32>,
     height: f32,
+    render_offscreen: bool,
 }
 
 impl Default for ListItem {
@@ -62,6 +63,7 @@ impl Default for ListItem {
             force_background: None,
             collapse_openness: None,
             height: DesignTokens::list_item_height(),
+            render_offscreen: true,
         }
     }
 }
@@ -137,6 +139,17 @@ impl ListItem {
     #[inline]
     pub fn with_height(mut self, height: f32) -> Self {
         self.height = height;
+        self
+    }
+
+    /// Controls whether [`Self`] calls [`ListItemContent::ui`] when the item is not currently
+    /// visible.
+    ///
+    /// Skipping rendering can increase performances for long lists that are mostly out of view, but
+    /// this will prevent any side effects from [`ListItemContent::ui`] from occuring. For this
+    /// reason, this is an opt-in optimization.
+    pub fn render_offscreen(mut self, render_offscreen: bool) -> Self {
+        self.render_offscreen = render_offscreen;
         self
     }
 
@@ -274,6 +287,7 @@ impl ListItem {
             force_background,
             collapse_openness,
             height,
+            render_offscreen,
         } = self;
 
         let collapse_extra = if collapse_openness.is_some() {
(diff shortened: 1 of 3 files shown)


crates/viewer/re_ui/src/list_item/list_item.rs
```diff
@@ -140,6 +142,18 @@ impl ListItem {
         self
     }
 
+    /// Controls whether [`Self`] calls [`ListItemContent::ui`] when the item is not currently
+    /// visible.
+    ///
+    /// Skipping rendering can increase performances for long lists that are mostly out of view, but
+    /// this will prevent any side effects from [`ListItemContent::ui`] from occurring. For this
+    /// reason, this is an opt-in optimization.
```
Wumpf, line 150: to clarify, could such a side-effect change the vertical size of a list array? If not, then wouldn't it be reasonable to expect this to be opt-in? I mean we're passing a closure after all for drawing a thing - if the thing isn't on screen, why would you expect your closure to be called; lotsa place don't do that

Wumpf (approved):
yay!

crates/viewer/re_ui/src/list_item/list_item.rs
abey79: No, the vertical size is fully determined by `ListItem::height`. Also, this change only affects the list item itself, not the fact that it might have children in a collapsing section below it. So there cannot be any issue with vertical layout.

And you are probably right that I'm being overly defensive with this opt-in stuff. I really don't have a strong reason to _not_ make it opt-out (or even always on), except of being scared to break things in untested areas 😅

Wumpf: hmmm I'd really like to make it opt-out, but we can also do this maybe just after the next release then if we're too scared of adding regressions

abey79: Reminder: https://github.com/rerun-io/rerun/issues/8815

PR #8810, opened 2025-01-25
Fix CPU spike caused by hanging connection after socket closure (#8806)

### LINKED ISSUE #8806: `serve_web` followed by `disconnect` outside of main thread causes 100% CPU usage
**Description**
`serve_web` followed by `disconnect` in an asyncio executor causes 100% CPU usage. `disconnect` closes the websocket as expected. This issue was encountered while running `serve_web` in a FastAPI request and can be reproduced in an asyncio context.

- Running `rec.inner.flush()` before `disconnect` didn't help.
- Sleeping for some time before `disconnect` to ensure all data is streamed also didn't resolve the issue.

**To Reproduce**
- Minimal Example:
```python
from uuid import uuid4
import rerun as rr
import asyncio

async def stream():
    rec = rr.new_recording(application_id=str(uuid4()), recording_id=uuid4())
    rr.serve_web(open_browser=False, web_port=0, ws_port=0, recording=rec)
    rec.disconnect()


async def main():
    await stream()

    while True:
        print("Running")
        await asyncio.sleep(1)


asyncio.run(main())
```

- Minimal Example with FastAPI:
```python
from uuid import uuid4

from fastapi import FastAPI
import rerun as rr

app = FastAPI()

@app.post("/api/stream")
async def stream():
    rec = rr.new_recording(application_id=str(uuid4()), recording_id=uuid4())
    rr.serve_web(open_browser=False, web_port=0, ws_port=0, recording=rec)
    rec.disconnect()
```

```toml
[project]
name = "2025-01-rerun-fastapi-issue"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
    "fastapi[standard]>=0.115.7",
    "rerun-sdk>=0.21.0",
]
```

```bash
fastapi run main.py
curl -X POST http://localhost:8000/api/stream
```

**Expected behavior**
Calling disconnect should not leave a process running that consumes 100% of CPU resources.

**Desktop:**
 - OS: NixOS 24.04

**Rerun version**
0.21.0

goktug97:
After the `RerunServer` shuts down, the `broadcast_thread_func` in [`crates/store/re_ws_comms/src/server.rs`](https://github.com/rerun-io/rerun/blob/5a9794990c4903c088ad77174e65eb2573162d97/crates/store/re_ws_comms/src/server.rs#L341) could not terminate due to the `log_rx.recv` function entering an infinite loop while attempting to receive data.

To resolve this, `recv` has been replaced with `recv_timeout`, allowing the thread to properly shut down and preventing excessive CPU usage.

### Related

* Closes #8806

### DIFF
--- a/crates/utils/re_smart_channel/src/receive_set.rs
+++ b/crates/utils/re_smart_channel/src/receive_set.rs
@@ -118,23 +118,15 @@ impl<T: Send> ReceiveSet<T> {
 
         let mut rx = self.receivers.lock();
 
-        loop {
-            rx.retain(|r| r.is_connected());
-            if rx.is_empty() {
-                return Err(RecvError);
-            }
-
-            let mut sel = Select::new();
-            for r in rx.iter() {
-                sel.recv(&r.rx);
-            }
-
-            let oper = sel.select();
-            let index = oper.index();
-            if let Ok(msg) = oper.recv(&rx[index].rx) {
-                return Ok(msg);
-            }
+        rx.retain(|r| r.is_connected());
+        let mut sel = Select::new();
+        for r in rx.iter() {
+            sel.recv(&r.rx);
         }
+
+        let oper = sel.select();
+        let index = oper.index();
+        oper.recv(&rx[index].rx).map_err(|_err| RecvError)
     }
 
     /// Returns immediately if there is nothing to receive.


Wumpf (changes requested):
thanks for the contribution! However, I don't think that's the correct fix: Instead of making `broadcast_thread_func` a "slow busyloop", `stop_listener` should wakeup the broadcast thread by hanging up the `ReceiveSet` after setting the shutdown flag. `ReceiveSet::recv` method is documented to return if the set is empty, so calling `clear` on it after setting the shutdown flag should hopefully do it.

Edit: `clear` won't work, that just gets stuck in a lock. Needs some more prodding on how a `ReceiveSet` can be woken up, might be that we need to send a poisoned message of sort

Wumpf:
> `log_rx.recv` function entering an infinite loop while attempting to receive data.

That actually also sounds a bug in `ReceiveSet` then - I'd expect it at most to block indefinitely rather than busy-looping

goktug97:
Makes sense, I checked the `ReceiveSet` and issue is happening because `is_empty()` check is done on `Vec<Receiver>`, not on the `Receiver` itself. Will revert the previous changes and push the fix.

goktug97:
The issue was due to these lines;
```rust
if let Ok(msg) = oper.recv(&rx[index].rx) {
    return Ok(msg);
}
```
It starts returning `RecvError` after disconnecting but that branch is not handled so loop enters infinite loop. Latest push solves the problem and `serve_web` works expected.

Wumpf (approved):
great catch, thank you so much. I agree, that makes more sense now

Wumpf:
Hmm strange, not seeing a report from the Rust CI. Maybe something is slightly broken. Ah well, `main` will run it as well and I'm feeling confident enough

PR #8819, opened 2025-01-27
Low-hanging optimizations of the time panel

abey79:
### Related

* Part of #8233

### What

Two small optimisations which, together, account for 2ms in a worse case scenario.

Benchmark: air traffic data, no views, uncollapsed time panel

image

Result:

image

### DIFF
--- a/crates/viewer/re_data_ui/src/item_ui.rs
+++ b/crates/viewer/re_data_ui/src/item_ui.rs
@@ -185,8 +185,7 @@ pub fn instance_path_icon(
         if db
             .storage_engine()
             .store()
-            .all_components_on_timeline(timeline, &instance_path.entity_path)
-            .is_some()
+            .entity_has_data_on_timeline(timeline, &instance_path.entity_path)
         {
             &icons::ENTITY
         } else {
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
neat!

Wumpf:
but the snapshot tests catch _something_. So maybe that changes appearance in some cases?
Want online diff reports so badly...

abey79:
> but the snapshot tests catch _something_. So maybe that changes appearance in some cases? Want online diff reports so badly...
> 
> edit: this should be unrelated? those tests are about the blueprint panel, not the timepanel

I'm changing a function that is also called in the blueprint tree, and apparently I broke it. Will look into it.

crates/viewer/re_data_ui/src/item_ui.rs
```diff
@@ -185,8 +185,7 @@ pub fn instance_path_icon(
         if db
             .storage_engine()
             .store()
-            .all_components_on_timeline(timeline, &instance_path.entity_path)
-            .is_some()
+            .entity_has_data_on_timeline(timeline, &instance_path.entity_path)
```
abey79, line 188: This behaves slightly differently for reasons I don't really comprehend, at least in some tests.

abey79: According to Clement, one is affected by GC but not the other, which doesn't explain why it break a test where no GC is supposed to take place.

abey79:
Drafting this for now—I'll revisit it after my other PRs have landed.

PR #8820, opened 2025-01-27
`ComponentBatch` doesn't implement `AsComponents` anymore

### LINKED ISSUE #8757: Enforce usage of tagged logging APIs: Rust

teh-cmc:
This makes it impossible to pass something implementing `ComponentBatch` (i.e. some native data that knows how to serialized itself into arrow component data) directly to `RecordingStream::log`.

This forces the caller to either..:
* explicitly opt into the low-level / component-level APIs via `my_data.serialized()`, which expose further custom-tagging facilities, or
* use the archetype-level APIs instead, which handle tagging automatically.

Similarly, this removes all the legacy support infrastructure that was needed to carry `ComponentBatch`es all the way to the logger:
* `log_component_batches`
* `AsComponents::as_component_batches`
* `ComponentBatchCow`, `ComponentBatchCowWithDescriptor`, ...
* etc

I think this is all that's needed in order to..:
* fix #8757

At the very least, it makes sure that existing code that used to log `ComponentBatch`es directly breaks at compile-time.


teh-cmc:
@rerun-bot full-check

teh-cmc:
@rerun-bot full-check

Wumpf (approved):

Wumpf:
Have something very very similar in progress in C++, so looks like we're aligned on how to go about this!

Issue #8744, opened 2025-01-20
Unify `sorbet` and `rerun` arrow metadata


jleibs:
We currently have 2 different arrow-metadata encoding schemas: 
- In query API results we use:
  - `sorbet.path`, `sorbet.semantic_family`, `sorbet.logical_type`, `sorbet.semantic_type`
- In chunk transport we use:
  - `rerun.entity_path`, `rerun.archetype_name`, `rerun.archetype_field_name`, Name of column as component name

Keeping track of which encoding is required at different points in the pipeline is hard and only adds confusion while bringing no meaningful utility.

We should pick a single set of metadata keys and use them consistently as part of our first release of requiring tagged-component metadata to be populated.

jleibs:
Some frameworks have restrictions on what defines a valid column-name. For example: lance doesn't permit top-level field names to contain a ".".

This is a good argument for Rerun being agnostic about the choice of field-name and encoding all metadata as proper metadata fields.

emilk:
We discussed today a bit on how to handle _mono components_, i.e. the common case of single-instances (e.g. scalars). We want to make this as ergonomic as possible.

* We will start supporting RecordBatches with Mono-types in them, and do the list-array wrapping on the way into the chunkstore so that code in viewer-space never has to worry about mono-types.
* Future: we can also introduce a sorbet tag, e.g. `rerun.mono` and use that to unwrap-list arrays when generating query results in order to make these datatypes round-trip properly from the perspective of users.

emilk:
This is the direction I'm heading:

```rs
pub struct ChunkSchema {
    chunk_id: ChunkId,
    entity_path: EntityPath,
    columns: Vec<ComponentColumnDescriptor>,
    …
}
impl From<ChunkSchema> for ArrowSchema { … }
impl TryFrom<ArrowSchema> for ChunkSchema { … }

struct ChunkBatch {
    schema: ChunkSchema,
    batch: ArrowRecordBatch,
}
impl AsRef<ArrowRecordBatch> for ChunkBatch { … }
impl TryFrom<ArrowRecordBatch> for ChunkBatch { … }
```

And similar for dataframes:

``` rs
pub struct DataframeSchema {
    // Each data column is guaranteed to have an entity path
    columns: Vec<ComponentColumnDescriptor>,
    …
}
impl AsRef<ChunkSchema> for ArrowSchema { … }
impl TryFrom<ArrowSchema> for DataframeSchema { … }

struct DataframeBatch {
[...]

Issue #8563, opened 2024-12-26
Use separate response messages in our gRPC API


emilk:
It is [best practice](https://kreya.app/blog/grpc-best-practices/#separate-request-and-response-messages) to use a different response message to each rpc call, as that allows us to add more fields to the response in the future without sacrificing compatibility.

That is, instead of 

```
    rpc QueryCatalog(QueryCatalogRequest) returns (stream DataframePart) {}
    rpc RegisterRecording(RegisterRecordingRequest) returns (DataframePart) {}
```

we should have

```
    rpc QueryCatalog(QueryCatalogRequest) returns (stream QueryCatalogResponse) {}
    rpc RegisterRecording(RegisterRecordingRequest) returns (RegisterRecordingResponse) {}
```

## Related
* https://github.com/rerun-io/rerun/issues/9125

grtlr:
I can do this as part of #8566.

zehiko:
the most recent API changes were made as part of https://github.com/rerun-io/dataplatform/issues/74 (which is good to read for understanding the motivation behind the changes).

PR #8821, opened 2025-01-27
Use Arrow IPC to encode the column schema

emilk:
### Related
* Sibling PR: https://github.com/rerun-io/dataplatform/pull/154


crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -39,7 +39,7 @@ message GetRecordingSchemaRequest {
 }
 
 message GetRecordingSchemaResponse {
-    repeated rerun.common.v0.ColumnDescriptor column_descriptors = 1;
+    bytes arrow_schema_as_ipc = 2;
```
jleibs, line 42: There exists a `message Schema`, which calso contains `bytes arrow_schema = 1` in common.proto. Seems like we should be using that here.

crates/store/re_sorbet/src/ipc.rs
```diff
@@ -0,0 +1,21 @@
...
+    let mut ipc_bytes = Vec::<u8>::new();
+    #[allow(clippy::disallowed_types)] // We only use StreamWriter for the schema, which is fine
+    let mut writer = arrow::ipc::writer::StreamWriter::try_new(&mut ipc_bytes, schema)?;
+    // writer.write(&batch)?;  // Is this needed?
+    writer.finish()?;
+    Ok(ipc_bytes)
+}
+
+/// Decode an arrow schema from IPC bytes.
+pub fn schema_from_ipc(ipc_bytes: &[u8]) -> Result<ArrowSchemaRef, ArrowError> {
```
jleibs, line 17: Would be nice if this were a helper implementing directly on: `re_protos::common::v0::Schema`.

jleibs (approved):

crates/store/re_protos/proto/rerun/v0/remote_store.proto
emilk: There is an advantage to using a different message response type for each rpc though:

https://github.com/rerun-io/rerun/issues/8563

jleibs: Yes; agree with separation of Response types for clarity. What I was suggesting was using the type from common here instead of inlining bare bytes into the response:

```
message GetRecordingSchemaResponse {
  rerun.common.v0.Schema schema = 2;
}
```

emilk: ah yeah, ofc :)

emilk:
@rerun-bot full-check

crates/store/re_sorbet/src/ipc.rs
```diff
@@ -0,0 +1,21 @@
+use arrow::{
+    datatypes::{Schema as ArrowSchema, SchemaRef as ArrowSchemaRef},
+    error::ArrowError,
+};
+
+/// Encode an arrow schema as IPC bytes.
+pub fn ipc_from_schema(schema: &ArrowSchema) -> Result<Vec<u8>, ArrowError> {
```
zehiko, line 7: nit: in ``re_log_types`` we've mostly been using ``encode`` and ``decode`` and their variants, I wonder if for consistency we should do the same here? perhaps ``ipc_encode`` or ``ipc_encode_schema``, rather than "creating" ipc from something.

emilk: Good feedback, but I'll do it in a future PR

Issue #8754, opened 2025-01-21
Tagged columnar updates: C++


teh-cmc:
Likely very close in design to the Rust one (which remains to be seen what the final version looks like in the end).

PR #8828, opened 2025-01-27
C++ `columns` method for convenient `send_columns` call through archetypes

Wumpf:
### Related

* Part of https://github.com/rerun-io/rerun/issues/8754

### What

C++ point `send_columns` example is now a lot shorter:

```cpp
    // Partition our data as expected across the 5 timesteps.
    auto position =
        rerun::Points3D::update_fields().with_positions(positions).columns({2, 4, 4, 3, 4});
    auto color_and_radius =
        rerun::Points3D::update_fields().with_colors(colors).with_radii(radii).columns();

    rec.send_columns2("points", time_column, position, color_and_radius);
```

Left todo among other things:
* update all other examples
    * we'll likely need the `with_many` calls we introduced in Rust as well
* `send_columns2` should become `send_columns`
    * this will be a breaking change -> add migration guide

### DIFF
--- a/rerun_cpp/src/rerun/recording_stream.hpp
+++ b/rerun_cpp/src/rerun/recording_stream.hpp
@@ -757,8 +757,6 @@ namespace rerun {
         /// Note that this API ignores any stateful time set on the log stream via the `RecordingStream::set_time_*` APIs.
         /// Furthermore, this will _not_ inject the default timelines `log_tick` and `log_time` timeline columns.
         ///
-        /// Any failures that may occur during serialization are handled with `Error::handle`.
-        ///
         /// \param entity_path Path to the entity in the space hierarchy.
         /// \param time_columns The time columns to send.
         /// \param component_columns The columns of components to send. Both individual `ComponentColumn`s and `Collection<ComponentColumn>`s are accepted.


rerun_cpp/src/rerun/recording_stream.hpp
```diff
@@ -717,6 +717,77 @@ namespace rerun {
...
+        /// Unlike the regular `log` API, which is row-oriented, this API lets you submit the data
+        /// in a columnar form. Each `TimeColumn` and `ComponentColumn` represents a column of data that will be sent to Rerun.
+        /// The lengths of all of these columns must match, and all
+        /// data that shares the same index across the different columns will act as a single logical row,
+        /// equivalent to a single call to `RecordingStream::log`.
+        ///
+        /// Note that this API ignores any stateful time set on the log stream via the `RecordingStream::set_time_*` APIs.
+        /// Furthermore, this will _not_ inject the default timelines `log_tick` and `log_time` timeline columns.
+        ///
+        /// Any failures that may occur during serialization are handled with `Error::handle`.
```
teh-cmc, line 760: Either this one or the one above has to be wrong about this, right?

teh-cmc (approved):

rerun_cpp/src/rerun/recording_stream.hpp
Wumpf: good catch, c&p mistake!

PR #8829, opened 2025-01-27
initial gRPC spec changes for supporting index creation and querying

zehiko:
### What

Expose creating and querying an index over an entire collection of data. 

We add support for:
* vector index
* inverted index for full text search
* btree index

Query data is always represented with a record batch with just 1 column that carries the query data.


crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -32,6 +42,108 @@ message DataframePart {
     bytes payload = 1000;
 }
 
+// ---------------- IndexCollection ------------------
+
+message IndexCollectionRequest {
```
jleibs, line 47: Naming this something like `CreateIndexRequest` seems more clear.

crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -32,6 +42,108 @@ message DataframePart {
...
+message BTreeIndex {
+    // TODO(zehiko) as properties as needed
+}
+
+message IndexCollectionResponse {}
+
+
+// ---------------- QueryCollectionIndex ------------------
+
+message QueryCollectionIndexRequest {
```
jleibs, line 96: This should take a ComponentColumnDescriptor to look up which index we're using.

crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -32,6 +42,108 @@ message DataframePart {
...
+        InvertedIndexQuery inverted = 1;
+        VectorIndexQuery vector = 2;
+        BTreeIndexQuery btree = 3;
+    }
+}
+
+message InvertedIndexQuery {
+    // Query to execute represented as the arrow data
+    // Query should be a unit RecordBatch with 2 columns:
+    // - 'index' column with the name of the column we want to query
```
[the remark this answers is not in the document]
jleibs: Return an error if it's > 1 row or > 1 column.

crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -9,6 +9,16 @@ service StorageNode {
     rpc Query(QueryRequest) returns (stream DataframePart) {}
     rpc FetchRecording(FetchRecordingRequest) returns (stream rerun.common.v0.RerunChunk) {}
 
+    rpc IndexCollection(IndexCollectionRequest) returns (IndexCollectionResponse) {}
+    // Collection index query response is a RecordBatch with 2 columns:
```
jleibs, line 13: I think we should (at least optionally) include the value of the data column in these results.

crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -9,6 +9,16 @@ service StorageNode {
     rpc Query(QueryRequest) returns (stream DataframePart) {}
     rpc FetchRecording(FetchRecordingRequest) returns (stream rerun.common.v0.RerunChunk) {}
 
+    rpc IndexCollection(IndexCollectionRequest) returns (IndexCollectionResponse) {}
```
grtlr, line 12: Right now everything lives under one `service StorageNode`. I wonder if it is cleaner to create several services that groups similar calls logically? Maybe something like `CatalogService`, `IndexService`, ...? It might be nice in the future to decide on a fine granular level which services should spin up, for example if we want to distribute loads across multiple VMs?

crates/store/re_protos/proto/rerun/v0/remote_store.proto
jleibs: An optional limit probably makes sense here.

crates/store/re_protos/proto/rerun/v0/remote_store.proto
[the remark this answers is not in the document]
jleibs: Tracking which recordings are in the index so we know whether reindexing is necessary would be nice to think about, but not necessary yet.

zehiko: there is still a tight relation between catalog, collection, collection query path, collection index query path. This definitely requires bit of thinking how we split it.  I'll create an issue, but I won't tackle this as part of this story.

crates/store/re_protos/proto/rerun/v0/remote_store.proto
zehiko: returning as part of any query for now, we can easily make it optional if we see it makes more sense.

crates/store/re_protos/proto/rerun/v0/remote_store.proto
zehiko: done

crates/store/re_protos/proto/rerun/v0/remote_store.proto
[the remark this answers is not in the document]
zehiko: https://github.com/rerun-io/dataplatform/issues/156

This will be tackled in a first follow up.

jleibs (approved):

PR #8831, opened 2025-01-27
Explain and test the hashing behavior relied upon in the micro-batcher

teh-cmc:
Title.

* Supersedes #https://github.com/rerun-io/rerun/pull/8788


teh-cmc:
I'm not sensing any objection, and the number in the PR tab is making me very uncomfortable right now :skull: (also who doesn't want more docs and tests :upside_down_face:)

PR #8832, opened 2025-01-27
Add hint for missing web viewer when building `re_web_viewer_server`

oxkitsune:
### What

Tell the user to build the web viewer when building `re_web_viewer_server`, and the required files do not exist.

Could perhaps use the `cargo::error=` directive, which is now respected in 1.84?

### DIFF
--- a/crates/viewer/re_web_viewer_server/build.rs
+++ b/crates/viewer/re_web_viewer_server/build.rs
@@ -1,4 +1,14 @@
+use std::path::Path;
+
 fn main() {
     // https://blog.rust-lang.org/2024/05/06/check-cfg.html
     println!("cargo::rustc-check-cfg=cfg(disable_web_viewer_server)");
+
+    let viewer_js_path = Path::new("./web_viewer/re_viewer.js");
+    let viewer_wasm_path = Path::new("./web_viewer/re_viewer_bg.wasm");
+
+    assert!(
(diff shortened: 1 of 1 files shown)


Wumpf (approved):
thanks, great addition!

crates/viewer/re_web_viewer_server/build.rs
```diff
@@ -1,4 +1,14 @@
...
+
 fn main() {
     // https://blog.rust-lang.org/2024/05/06/check-cfg.html
     println!("cargo::rustc-check-cfg=cfg(disable_web_viewer_server)");
+
+    let viewer_js_path = Path::new("./web_viewer/re_viewer.js");
+    let viewer_wasm_path = Path::new("./web_viewer/re_viewer_bg.wasm");
+
+    assert!(
+        viewer_js_path.exists() && viewer_wasm_path.exists(),
```
emilk, line 11: This should probably have been a `||`

oxkitsune: they both NEED to exist afaik right?

Wumpf: correct https://github.com/rerun-io/rerun/blob/main/crates/viewer/re_web_viewer_server/src/lib.rs#L30

PR #8833, opened 2025-01-27
Always split indicators into their own dedicated chunks

teh-cmc:
* Fixes https://github.com/rerun-io/rerun/issues/8768

 ---

* [x] full check passed


teh-cmc:
@rerun-bot full-check

Wumpf (approved):
that's a lot less involved than I feared it would be, nice

PR #8845, opened 2025-01-28
Fix formatting in Python SDK LineStrip2D/3D

goktug97:
Error message in Python SDK LineStrip2D/3D is missing f-string


teh-cmc:
Thanks

PR #8849, opened 2025-01-29
Python: untagged-resistant APIs

teh-cmc:
It's pretty much just a matter of explicitly asking for `DescribedComponentBatch`es instead of `ComponentBatchLike`s in a couple of places.

The end result is that legacy code still works, but yields type errors.

* Fixes https://github.com/rerun-io/rerun/issues/8756

### DIFF
--- a/rerun_py/rerun_sdk/rerun/_baseclasses.py
+++ b/rerun_py/rerun_sdk/rerun/_baseclasses.py
@@ -172,7 +172,7 @@ class ComponentBatchLike(Protocol):
 class AsComponents(Protocol):
     """Describes interface for interpreting an object as a bundle of Components."""
 
-    def as_component_batches(self) -> Iterable[ComponentBatchLike]:
+    def as_component_batches(self) -> list[DescribedComponentBatch]:
         """
         Returns an iterable of `ComponentBatchLike` objects.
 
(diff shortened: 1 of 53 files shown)


rerun_py/rerun_sdk/rerun/any_value.py
```diff
@@ -213,15 +213,15 @@ def __init__(self, drop_untyped_nones: bool = True, **kwargs: Any) -> None:
         """
         global ANY_VALUE_TYPE_REGISTRY
 
-        self.component_batches = []
+        self.component_batches = list([])
```
abey79, line 216: nit: that's weird. Should be either `list()` or `[]` (i believe both are strictly equivalent).

rerun_py/rerun_sdk/rerun/_baseclasses.py
```diff
@@ -204,24 +204,27 @@ def archetype_name(cls) -> str:
...
-        Check out the `manual_indicator` API example to see what's possible.
         """
         from ._log import IndicatorComponentBatch
 
-        return IndicatorComponentBatch(cls.archetype_name())
+        indicator = IndicatorComponentBatch(cls.archetype_name())
+        return DescribedComponentBatch(indicator, indicator.component_descriptor())
 
-    def as_component_batches(self) -> Iterable[ComponentBatchLike]:
+    def as_component_batches(self, *, include_indicators: bool = True) -> list[DescribedComponentBatch]:
```
abey79, line 218: I don't mind the switch from returning a generator to returning a list, because the performance implications here are negligible, so I'm ok with this change.

For records, if this due to typing issues, I believe it should have been typed `Iterator[...]` instead of `Iterable[...]` (the technically correct type is `Generator[…, …]` but its too verbose for simple generators).

Also, as a rule of thumb, return type annotation should be "the real thing that is actually returned" (so not `....Like` stuff), while function arguments should have the most general thing it's able to accept (so `...Like` are ok).

abey79 (approved):
lgtm

rerun_py/rerun_sdk/rerun/_baseclasses.py
teh-cmc: > I don't mind the switch from returning a generator to returning a list, because the performance implications here are negligible, so I'm ok with this change.

Honestly I wouldn't be surprised if this turns out to be faster... This always returns list with a handful of elements max, I can definitely see the fixed overhead of the generator machinery being bigger than that.

> For records, if this due to typing issues, I believe it should have been typed `Iterator[...]` instead of `Iterable[...]` (the technically correct type is `Generator[…, …]` but its too verbose for simple generators).

This was due to needing to be able to check the `len()` in a bunch of places.

> Also, as a rule of thumb, return type annotation should be "the real thing that is actually returned" (so not `....Like` stuff), while function arguments should have the most general thing it's able to accept (so `...Like` are ok).

Strongly agree. This is one of my major pet peeve with the Python SDK right now -- like half of the functions return protocols and SomethingLike and such...

teh-cmc:
Rebasing + runtime a full roundtrip suite locally before merging :slow_beachball:

Issue #8848, opened 2025-01-29
ffmpeg-cli decoder requires additional frames enqueued for some videos in order to produce frames + spurious errors


Wumpf:
Some video material, for instance https://github.com/user-attachments/assets/726c1803-9b4b-4dfb-a6ee-29a23a2353a9
(source https://huggingface.co/datasets/pollen-robotics/apple_storage) requires us to enqueue up to 16 more video samples than what expect.

As of [TODO PR LINK] we now make sure that we have at least 16 samples enqueued for the ffmpeg cli decoder at any point in time to work around this.

---

Furthermore, even with a more aggressive enqueuing strategy that takes this into account we are seeing spurious errors like these which recover either after rendering more frames.

![Image](https://github.com/user-attachments/assets/251b21c1-0b00-4c7d-b9fd-366dd9e2775d)
![Image](https://github.com/user-attachments/assets/ead82ced-d6f5-4cdf-89d6-93f1a6c5bc1f)

It also has been observed that when viewing several of these videos at once errors only disappear after seeking:

![Image](https://github.com/user-attachments/assets/ee5f781b-3552-4f01-96d4-72d4185d04f1)

---

As of now it's unclear what is different about this video material than others we tried in the past. The video example linked above is an "easy" case: It has only GOPs with length 2, each consisting of an IDR frame (marked as such both in the NAL header & mp4 container) and a P frame. Did not find any special properties in the SPS (also looked into parsing the PPS but didn't succeed).

Already experimented among other things with....
* various flags to ffmpeg to reduce latency further or override timings
* insert `EndStream`/`EndStream` NAL units at the end of GOPs
* inserting sleeps to feed ffmpeg slower
* insert fill data on stdin
* convert to h264 and back to mp4 with ffmpeg - both play fine and no change if done with copy flags to avoid re-encoding (this should exclude most mp4 specific issues, but note that the AVCC box holding PPS & SPS should be completely replicated)
* extract single frames to image with ffmpeg works fine
* other video players handle this just fine (tried Windows Media Player, VLC, Firefox, Chrome)

Things we should still look into:
* try to build our own "problematic video" in order to understand better what causes these issues (notably the video was chunked from a larger video)
* read-up more on WebCodec implementation in Firefox & Chrome -> prototype with directly interfacing libav
* ?

PR #8850, opened 2025-01-29
Fix playback issues with some h264 videos on native & Safari

Wumpf:
### Related

* This solution so far is a workaround, details on what's still bad are in https://github.com/rerun-io/rerun/issues/8848

### What

Works around the issue outlined in https://github.com/rerun-io/rerun/issues/8848
In a nutshell: For certain videos we would only ever receive frames that are 16 frames behind of what we enqueued, completely disregarding the fact that we always enqueue full GOPs (I repeatedly confirmed that this is the case with different tools). This affects only h264 on the ffmpeg-cli & WebCodec @ Safari decoders. The exact circumstances that make some videos require this and other not are unknown at this point.

This is "solved" here by making our frame enqueing process more flexible and allowing decoder implementations to specify a "minimum look ahead" number of frames.
Note that oftentimes we still enqueue much more frames regardless: We still want to enqueue the entire current & next GOP for smooth transitions. Typically two h264 GOPs encompass a lot more frames than requested here (citation needed). However, the tested problematic videos operate with GOPs that contain only two frames (single I & P frame) so there this actually makes a difference.

Testing: Beyond the actual video, I tested various samples form our internal library of "curious videos to test" to ensure that the adjusted enqueuing logic doesn't regress anything. (personally, I think it's actually much nicer independent of the workaround :))


crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -235,49 +236,75 @@ impl VideoPlayer {
...
 
-        self.current_gop_idx = requested_gop_idx;
-        self.current_sample_idx = requested_sample_idx;
+        // Ensure that we have as many GOPs enqueued currently as needed in order to...
+        // * cover the GOP of the requested sample _plus one_ so we can always smoothly transition to the next GOP
+        // * cover at least `min_num_samples_to_enqueue_ahead` samples to work around issues with some decoders
+        //   (note that for large GOPs this is usually irrelevant)
+        //
+        // (potentially related to:) TODO(#7327, #7595): We don't necessarily have to enqueue full GOPs always.
+        // In particularly beyond `requested_gop_idx` this can be overkill.
```
Wumpf, line 278: this is a pre-existing issue, it just wasn't referenced here before

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -223,8 +225,7 @@ impl VideoPlayer {
             //
             // By resetting the current GOP/sample indices, the frame enqueued code below
             // is forced to reset the decoder.
-            self.current_gop_idx = usize::MAX;
-            self.current_sample_idx = usize::MAX;
+            self.last_requested_gop_idx = usize::MAX;
```
Wumpf, line 228: this is a bit wonky imho - why not just `self.reset` directly. A: This would lead to a double reset.
In the interest of keeping the changes here more minimal I'm not touching this piece

crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -190,8 +191,27 @@ impl AsyncDecoder for WebVideoDecoder {
...
+        }
+    }
 }
 
+const IS_SAFARI: Lazy<bool> = Lazy::new(|| {
+    web_sys::window()
+        .and_then(|w| w.navigator().user_agent().ok())
+        .map(|ua| ua.contains("Safari"))
+        .unwrap_or(false)
+});
```
emilk, line 213: A lot of modern browser user agents has the string `Safari` in them.

Look at this instead:
https://github.com/rerun-io/rerun/blob/e0c706b01a52fa463f60b4175c4f3b730dbaa002/crates/viewer/re_viewer_context/src/viewer_context.rs#L254-L271

Wumpf: thanks! 👀 👀

Issue #4413, opened 2023-12-01
Cannot range-select views with shift-click


teh-cmc:
One can select multiple views with ctrl+click in the blueprint panel, but when you have 20 views this quickly become cumbersome... a shift+click range-selection kinda thing would be nice.

abey79:
The v1 of this feature would be to operate within a single hierarchical level. So:

1) keep track of the last selected items within the current (possibly multi-)selection
2) upon shift click of a new, unselected item:
  - if that item is in the same hierarchical level as the last selected item (e.g. a view in the same parent container, or an entity part in the same parent entity path), do the shift select
  - if that item is NOT in the same hierarchical level: behave as with cmd-click, aka just add that one item to the selection.

@gavrelina does that sound like a reasonable plan to you?

PR #8852, opened 2025-01-29
Refactor interaction handling code

abey79:
### Related

* Closes https://github.com/rerun-io/rerun/issues/4413

### What

This PR makes it possible to range select stuff in the blueprint tree by shift-clicking items.

This works as follows:
- Any time an item is clicked without shift, it is remembered as the "last clicked item"
- Any time an item is clicked with shift, the items between the "last clicked item" and the shift-clicked item (inclusive)…
  - …replace the current selection (shift-click)
  - …extend the current selection (cmd/ctrl-shift-click)

https://github.com/user-attachments/assets/b77d8811-7eb1-4ead-b863-66274ea5d5c3


crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
```diff
@@ -576,11 +642,111 @@ impl BlueprintTree {
...
+                if items_iterator.len() > 0 {
+                    if modifiers.command {
+                        ctx.selection_state.extend_selection(items_iterator);
+                    } else {
+                        ctx.selection_state.set_selection(items_iterator);
+                    }
+                }
+            }
+        } else {
+            self.last_clicked_item = Some(item);
```
Wumpf, line 692: I feel like `last_clicked_item` is a bit of a misnomer since after a shift-click it is not the last clicked but more like `start_of_range_selection`, but I figure that's also confusing since when we first set it we don't know yet if that is going to be a range selection. Hrrrrm :(

Maybe `range_selection_anchor` or something?

crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
```diff
@@ -576,11 +642,111 @@ impl BlueprintTree {
...
+                        (
+                            item,
+                            Some(ItemContext::BlueprintTree {
+                                filter_session_id: self.filter_state.session_id(),
+                            }),
+                        )
+                    });
+
+                if items_iterator.len() > 0 {
+                    if modifiers.command {
```
Wumpf, line 684: TIL I learned about this subtle behavior being standard (it truly is, tried it)

Wumpf: since this is going to show up in more places, maybe that would be a good highlevel utility on the selection state? aka `selection_state.shift_select(modifiers.command, items_iterator)` (not a good name)

crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
```diff
@@ -576,11 +642,111 @@ impl BlueprintTree {
...
+        shift_clicked_item: &Item,
+    ) -> Vec<Item> {
+        let mut items_in_range = vec![];
+        let mut found_last_clicked_items = false;
+        let mut found_shift_clicked_items = false;
+
+        blueprint_tree_data.visit(|blueprint_tree_item| {
+            let item = blueprint_tree_item.item();
+
+            if Some(&item) == self.last_clicked_item.as_ref() {
```
Wumpf, line 714: wouldn't it makre more sense to early out on method start if `self.last_clicked_item` is none?

Wumpf (approved):
beautiful! The controlflow utility for recursive walks is marvelous!

no tests for this though? :O
blueprint_tree should be testable by inducing some actions in the test harness, or is this too complicated?

crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
abey79: `range_selection_anchor_item` it is :+1:

abey79: `range_selection_anchor_item` it is :+1:

crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
abey79: I'm going to invoke the rule of 3 here :)

PR #8853, opened 2025-01-29
Require descriptors to be provided on all log calls in C++ (either explicitely or implicitely via archetype)

Wumpf:
### Related

* fixes https://github.com/rerun-io/rerun/issues/8758

### What

Followed the Rust example and opted out of deprecating this, instead making it a breaking change.
Iterated on the migration guide & adjusted snippets.

* [x] pass main ci

### DIFF
--- a/docs/content/reference/migration/migration-0-22.md
+++ b/docs/content/reference/migration/migration-0-22.md
@@ -39,11 +39,29 @@ As part of the switch to "eager archetype serialization" (serialization of arche
 
 However, it is still possible to do so with the `TensorData` component.
 
-### C++ `RecordingStream::send_column` no longer takes raw component collections
+### C++ `RecordingStream::log`/`send_column` no longer takes raw component collections
 
-Previously, `RecordingStream::send_column` accepted raw component collections.
+Previously, both `RecordingStream::log` and `RecordingStream::send_column` were able to
+handle raw component collections which then would be serialized to arrow on the fly.
 
-This is no longer the case and only `rerun::ComponentColumn` and anything else from which
+#### `log`
+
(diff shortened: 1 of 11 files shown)


docs/content/reference/migration/migration-0-22.md
```diff
@@ -39,11 +39,29 @@ As part of the switch to "eager archetype serialization" (serialization of arche
...
+
+Under the hood we allow any type that implements the `AsComponents` trait.
+However, `AsComponents` is no longer implemented for collections of components / implementors of `Loggable`.
+
+Instead, you're encouraged to use archetypes for cases where you'd previously use loose collections of components.
+This is made easier by the fact that archetypes can now be created without specifying required components.
+For example, colors of a point cloud can be logged without position data:
+
+```cpp
+rec.log("points", rerun::Points3D().with_colors(colors));
```
teh-cmc, line 57: Hmmm.. I don't like pushing people towards this ("this" being an implicit partial update, as opposed to `rerun::Points3D::update_fields()`). If somehow executes that thing, they will end up in a very confusing situation where the viewer displays nothing...

teh-cmc (approved):
That predates this PR, but I'm not a fan of the fact that the most basic constructor allows not specifying required components now :confused:.

That means there is no codepath left that pushes the user towards the sensible default, compared to e.g. in Rust where `new()` still enforces requirements.

Wumpf:
@rerun-bot full-check

Wumpf:
that's a valid concern, let's talk about pro/con & solutions before shipping this. But at that point we need to do that as iteration anyways, so merging this as-is

PR #8854, opened 2025-01-29
Make it possible to change the contents of a view multiple times per frame

### LINKED ISSUE #8518: Improve the entity path filter mutation API

abey79:
### Related

* Fixes #8518 

### What

Before, when we would mutate a view entity path filter multiple time in a frame, only the last mutation would be recorded. This could be visible by:
- multi-selecting several view entities in the blueprint tree (within the same view)
- right-click -> Remove

==> only one entity would be removed.

This PR makes it possible to mutate a view's content multiple time per frame, solving this bug—and possibly others.


crates/viewer/re_viewport_blueprint/src/view_contents.rs
```diff
@@ -37,8 +40,17 @@ use crate::{ViewBlueprint, ViewProperty};
...
+
+    /// Deserialized entity path filter.
+    ///
+    /// Consider this read-only.
+    entity_path_filter: ResolvedEntityPathFilter,
+
+    /// Update entity path filter.
+    ///
+    /// Mutations go to this value and should be saved to the blueprint store when they occur.
+    new_entity_path_filter: Arc<Mutex<ResolvedEntityPathFilter>>,
```
emilk, line 53: This is such a common pattern in imgui. We should at some point think of coming up with a higher-level abstraction for it 🤔  (but not now)

emilk (approved):

crates/viewer/re_viewport_blueprint/src/view_contents.rs
abey79: 💯

PR #8861, opened 2025-01-30
Show start of large arrow values instead of just their size

emilk:
This changes the arrow_ui to always show the actual contents of the data, and never to show the summary of "10.2 MiB of data".

The reasons are two-fold:
* A) The size estimation was wrong, so we would show the wrong size, and too often
* B) It is probably nicer to se the contents of the data instead, even if we can only fit the start of it because it is too long (see screenshot test in PR diff)

## Worries
Formatting a huge arrow array as a string may be slow. This is difficult to see in practice, because we special-treat the display of `Blob` in the UI anyways.

## Other changes
I also changed nulls to be explicitly formatted as the string `null`. Before they were omitted.


teh-cmc (approved):

teh-cmc:
Yeah this is sooo much better

teh-cmc:
I'll rebase on https://github.com/rerun-io/rerun/pull/8862 as soon as possible then merge.

emilk:
I'll update the tests

PR #8728, opened 2025-01-17
Filter entities in the UI (part 5): Add snapshot tests for the blueprint tree

abey79:
### Related

- Part of #8586
- Part of a series of PR:
  - #8645
  - #8652
  - #8654
  - #8672
  - #8706
  - #8728
  - …

### What

This PR adds snapshot tests for the blueprint tree. It includes several improvements to the testing rig, such that those test are now quite compact. The graph view test is simplified to benefit from these changes.

PR #8863, opened 2025-01-30
Filter entities in the UI (part 7): Refactor and optimize `re_time_panel` and add more tests

abey79:
### Related

- Part of #8586
- Part of a series of PR:
  - #8645
  - #8652
  - #8654
  - #8672
  - #8706
  - #8728
  - #8795
  - #8863
  - ...

### What

This PR refactors the streams tree code in a similar way as what was done in the blueprint tree in #8795. In addition, it makes it significantly faster.

I've decided *against* listing all components in the data structure and instead keep that at the (on-demand) UI level. The reason for that is that `store.all_components_for_entity()` can be rather expensive as the number of entities grows.

**Note**: This PR does not introduce shift-click range selection (that will be in a follow up PR). Because of the design decision above, it will be slightly more tricky (but worth it perf-wise).

#### Visual changes

Only visual/behavioural change introduced in this PR is that leaf entities are now collapsed by default when the entity filter is active (we're searching for entities, not components, so we don't need to see them).

#### Performance

Like the previous PR, we are now parsing the entire entity tree regardless of whether we're going to use/display it or not. With air traffic data, that's 0.3ms only.

As for the optimisations, it's hard to keep track of everything. But frame time for air traffic data with no view and uncollapsed time panel went from ~44ms (0.21) to ~26ms (this PR).

### DIFF
--- a/crates/viewer/re_time_panel/tests/time_panel_tests.rs
+++ b/crates/viewer/re_time_panel/tests/time_panel_tests.rs
@@ -1,12 +1,14 @@
 use egui::Vec2;
 
 use re_chunk_store::{LatestAtQuery, RowId};
-use re_log_types::build_frame_nr;
-use re_log_types::example_components::MyPoint;
-use re_log_types::external::re_types_core::Component;
+use re_entity_db::InstancePath;
+use re_log_types::{
+    build_frame_nr, example_components::MyPoint, external::re_types_core::Component, EntityPath,
+    TimeInt, TimePoint, TimeType, Timeline,
+};
 use re_time_panel::TimePanel;
-use re_viewer_context::blueprint_timeline;
-use re_viewer_context::test_context::TestContext;
+use re_types::archetypes::Points2D;
+use re_viewer_context::{blueprint_timeline, test_context::TestContext, CollapseScope, TimeView};
 use re_viewport_blueprint::ViewportBlueprint;
 
 #[test]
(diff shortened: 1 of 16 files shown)


crates/viewer/re_time_panel/tests/time_panel_tests.rs
```diff
@@ -137,17 +162,129 @@ pub fn run_time_panel_filter_tests(filter_active: bool, query: &str, snapshot_na
...
+            run_expanded_time_panel_and_save_snapshot(
+                test_context,
+                time_panel,
+                &format!("various_entity_kinds_{timeline}_{time}"),
+            );
+        }
+    }
+}
+
+pub fn log_all_data(test_context: &mut TestContext) {
```
Wumpf, line 206: nit: a bit irritating that at the top there's some tests that log stuff and here we have a "log_all_data" which logs all the data for a _specific_ test

crates/viewer/re_time_panel/tests/time_panel_tests.rs
```diff
@@ -55,27 +53,62 @@ pub fn time_panel_dense_data_should_match_snapshot() {
...
-        .recording_store
-        .add_chunk(&Arc::new(builder.build().unwrap()))
-        .unwrap();
+        builder
+    });
 
     run_time_panel_and_save_snapshot(test_context, TimePanel::default(), "time_panel_dense_data");
 }
 
+fn run_time_panel_and_save_snapshot(
```
Wumpf, line 75: the difference between this an `run_expanded_time_panel_and_save_snapshot` isn't quite clear. both are exptended, one is just larger and expands the items rather than the time panel (as the name suggests). Looks like a single parameterized one would be better

Wumpf (approved):
looking great!

PR #8864, opened 2025-01-30
Clickable URLs in named components / `AnyValues`

### LINKED ISSUE #8837: Detect URLs in strings and render them appropriately
## Prior discussion

* #8567 

## Motivation

Users want to attach URLs to instances to provide further information. In the issue linked above, we explored creating a URI component, but after some discussion settled on a more general solution: simply render strings that look like URIs (or that can be parsed as such) as links. This should work nicely together with `AnyValue` and show up in tooltips and in the selection panel.

grtlr:
### Related

* Supersedes #8567.
* Closes #8837.

### What

Title.

https://github.com/user-attachments/assets/a08fd640-254b-45d1-9ba2-d83802980dd9

Minor problems before this can be merged
* [x] The component name disappears when hovering over the URI entry, and maybe related
* [x] The chunk store browser crashes with a panic.


crates/viewer/re_ui/src/ui_layout.rs
```diff
@@ -95,15 +99,21 @@ impl UiLayout {
         self.data_label_impl(ui, string.as_ref())
     }
 
+    fn decorate_url(ui: &mut egui::Ui, text: &str, galley: Arc<egui::Galley>) -> egui::Response {
+        if url::Url::parse(text).is_ok() {
+            ui.re_hyperlink(text, text)
+        } else {
+            ui.label(galley)
+        }
+    }
```
abey79, line 108: This should be more correct, as it will benefit from the fine-tuning we do in `data_label_impl`.

```suggestion
    fn decorate_url(ui: &mut egui::Ui, text: &str, galley: Arc<egui::Galley>) -> egui::Response {
        if url::Url::parse(text).is_ok() {
            ui.re_hyperlink(galley, text)
        } else {
            ui.label(galley)
        }
    }
```

abey79 (approved):
lgtm. I was initially unhappy to see `re_hyperlink` _not_ truncate in the narrow snapshot, but the last commit updated the snapshot :)

PR #8865, opened 2025-01-30
Update to latest egui

emilk:
## Changes
* When painting a rect stroke, you must chose if the stroke is inside, outside, or on the rectangle.
* TextEdit and ColorButton have been shrunk to fit inside their widget rects

## TODO
* [x] Check that the choices of `StrokeKind` makes sense everywhere


abey79 (approved):
LGTM. Did some cursory testing, didn't spot anything super weird.

PR #8870, opened 2025-01-30
Implement range selection with shift-click in the time panel

abey79:
### What

Same as #8852 but in the streams tree.

This time, the visitor is slightly weird in that it enumerates the component on the fly instead of having them in the data structure. The reason for that is that this is costly and we don't want to _always_ do it even if we dont needed (eg. when the tree is collapsed).

https://github.com/user-attachments/assets/a8358e59-ae29-42a5-b579-218f16e832b3


crates/viewer/re_time_panel/Cargo.toml
```diff
@@ -25,10 +25,11 @@ re_data_ui.workspace = true
 re_entity_db.workspace = true
 re_format.workspace = true
 re_int_histogram.workspace = true
-re_log_types.workspace = true
 re_log.workspace = true
+re_log_types.workspace = true
```
Wumpf, line 29: somewhat curious this needs log types actually and not just types_core, ah well...

crates/viewer/re_time_panel/src/lib.rs
```diff
@@ -148,6 +147,12 @@ pub struct TimePanel {
     /// recording with a different application id.
     #[serde(skip)]
     filter_state_app_id: Option<ApplicationId>,
+
+    /// Last clicked item.
+    ///
+    /// We keep track of it to implement the range selection using shift-click.
+    #[serde(skip)]
+    last_clicked_item: Option<Item>,
```
Wumpf, line 155: same as on the other pr, not happy about the name. But now it's a thing so not feeling strongly about it ;)

crates/viewer/re_time_panel/src/lib.rs
```diff
@@ -840,164 +847,291 @@ impl TimePanel {
...
+            let item = if let Some(component_name) = component_name {
+                Item::ComponentPath(ComponentPath::new(
+                    entity_data.entity_path.clone(),
+                    component_name,
+                ))
+            } else {
+                entity_data.item()
+            };
+
+            if Some(&item) == self.last_clicked_item.as_ref() {
```
Wumpf, line 1100: as in other pr, `last_clicked_item` being some seems like a prerequisite (?)

crates/viewer/re_time_panel/src/lib.rs
abey79, line 155: to be renamed `range_selection_anchor_item`

abey79, line 155: to be renamed `range_selection_anchor_item`

Wumpf (approved):
nice, reading this right after the blueprint tree one this feels like more of the same with situational adaptation :)

Wumpf:
again, surprised this doesn't come with tests

abey79:
> again, surprised this doesn't come with tests

https://github.com/rerun-io/rerun/issues/8871

crates/viewer/re_time_panel/Cargo.toml
abey79: It was there already, was just reordered.

``` rust
use re_log_types::{
    ApplicationId, ComponentPath, EntityPath, ResolvedTimeRange, TimeInt, TimeReal, TimeType,
};
```

Issue #8857, opened 2025-01-29
TransformCache doesn't handle clears


Wumpf:
Fairly sure that's what causes extra transform arrows to show up here (where before they would move the the origin.. which is also questionable, but doesn't look broken like this does) https://rerun.io/viewer/version/main?url=https%3A%2F%2Fstatic.rerun.io%2Frrd%2F0.20.0%2Frevy_breakout_3572dc5d61f77dc4fc9675a85c74035a6ee4b020.rrd

PR #8872, opened 2025-01-30
Fix handling clears in transform cache

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/8857

### What

Testing:
* unit test
* revy breakout looks again as before https://rerun.io/viewer/pr/8872?url=https%3A%2F%2Fstatic.rerun.io%2Frrd%2F0.20.0%2Frevy_breakout_3572dc5d61f77dc4fc9675a85c74035a6ee4b020.rrd

as before, most of this is added test code. Some clutter in the diff because I split `add_chunk` into `add_non_static_chunk` and `add_static_chunk` to make it easier to reason. Also this unfortunately rippled into minor interface change since I had to go with `EntityPath` in one spot where I used `EntityPathHash` before

### DIFF
--- a/crates/viewer/re_view_spatial/src/transform_cache.rs
+++ b/crates/viewer/re_view_spatial/src/transform_cache.rs
@@ -41,6 +41,11 @@ pub struct TransformCacheStoreSubscriber {
 
     per_timeline: HashMap<Timeline, CachedTransformsForTimeline>,
     static_timeline: CachedTransformsForTimeline,
+    // Need to keep track of all recursive clears that ever happened.
+    //
+    // Otherwise, new incoming entities may not correctly change their transform at the time of clear
+    //recursive_clears: HashSet<EntityPath>,
+    // TODO:
 }
 
 impl Default for TransformCacheStoreSubscriber {
@@ -85,6 +90,9 @@ bitflags::bitflags! {
 
         /// The entity has a pinhole projection or view coordinates, i.e. either [`components::PinholeProjection`] or [`components::ViewCoordinates`].
         const PinholeOrViewCoordinates = 1 << 2;
+
+        /// The entity has a clear component.
+        const Clear = 1 << 3;
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/viewer/re_view_spatial/src/transform_cache.rs
```diff
@@ -392,127 +476,173 @@ impl TransformCacheStoreSubscriber {
         }
     }
 
-    fn add_chunk(&mut self, event: &re_chunk_store::ChunkStoreEvent, aspects: TransformAspect) {
+    fn add_non_static_chunk(
```
teh-cmc, line 479: `add_chunk` or `add_temporal_chunk` is what we usual call this

crates/viewer/re_view_spatial/src/transform_cache.rs
```diff
@@ -392,127 +476,173 @@ impl TransformCacheStoreSubscriber {
...
+            // So instead, we note down that the caches are invalidated for the given entity & times.
+
+            // This invalidates any time _after_ the first event in this chunk.
+            // (e.g. if a rotation is added prior to translations later on,
+            // then the resulting transforms at those translations change as well for latest-at queries)
+
+            let mut invalidated_times = Vec::new();
+            let Some(min_time) = time_column.times().min() else {
+                continue;
+            };
```
teh-cmc, line 508: Technically, this is too conservative (i.e. you'd want a bunch of `min_time_for_component` instead).
In practice, this is absolutely fine and is probably not worth the extra complexity.

teh-cmc (approved):
If there's a nasty edge-case lurking somewhere in there, I certainly won't catch it just by reading.
The general idea looks pretty legit to me.

crates/viewer/re_view_spatial/src/transform_cache.rs
Wumpf: much better, thanks

PR #8873, opened 2025-01-30
`re_time_panel`: no longer cram everything in `lib.rs`

abey79:
Purely moving around stuff to no longer have a fat `lib.rs`


Wumpf (approved):
ah nice, finally being able to tell what the crate exports. also have it export a lot less things 👌

PR #8875, opened 2025-01-30
Improve look of bar charts slightly

emilk:
## What
Removes the gap between the bars. The gaps looked uneven because of pixel grid aliasing. Better to have no gaps (it is far less noticeable that some bars are a pixel wider/narrower than others since they are already so wide).

## Before/after
The comparison pictures are taken on a high-DPI display. The difference is much bigger on a low-DPI display.

### Before
Screenshot 2025-01-31 at 05 13 41

### After
image


abey79 (approved):
Yeah, it does look better.

PR #8879, opened 2025-01-31
Remove `unwrap`s in web build of `re_viewer`

emilk:
This should produce better error messages when running Rerun Wasm outside of a browser.

Required because of
* https://github.com/rerun-io/rerun/pull/8745

---
* [x] full check


emilk:
@rerun-bot full-check

Wumpf (approved):

Wumpf:
full check got far enough to only have some wheel based checks, so this can go in

PR #8884, opened 2025-01-31
`AsComponents::serialize` is now `AsComponents::as_batches` and returns `Collection<ComponentBatch>`

Wumpf:
### What

See title. This is a breaking change for custom `AsComponents` implementations. But we had plenty of breaking changes in that area, so now it's the right time.

The rename `serialize` to `as_batches` is motivated by the fact that _previously_ we typically did to-arrow serializations in `AsComponents`. This is no longer the case and it's more of a collecting operation, just as we do in Rust's `AsComponents::as_serialized_batches` - since C++ only has `ComponentBatch` and no `SerializedComponentBatch` it follows that the method is just called `as_batches` now.

The change in return type fixes an old todo (no ticket) for using `rerun::Collection` for more flexibility in this area (pending improvements to `rerun::Collection`) and making our API handling more consistent in that regard.

### DIFF
--- a/rerun_cpp/src/rerun/as_components.hpp
+++ b/rerun_cpp/src/rerun/as_components.hpp
@@ -6,7 +6,7 @@
 #include "loggable.hpp"
 
 namespace rerun {
-    /// The AsComponents trait is used to convert a type into a list of as_batchesd component.
+    /// The AsComponents trait is used to convert a type into a list of component batches.
     ///
     /// It is implemented for various built-in types as well as collections of components.
     /// You can build your own archetypes by implementing this trait.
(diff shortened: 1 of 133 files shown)


rerun_cpp/src/rerun/as_components.hpp
```diff
@@ -6,7 +6,7 @@
 #include "loggable.hpp"
 
 namespace rerun {
-    /// The AsComponents trait is used to convert a type into a list of serialized component.
+    /// The AsComponents trait is used to convert a type into a list of as_batchesd component.
```
teh-cmc, line 9: typo

teh-cmc (approved):

PR #8887, opened 2025-01-31
Clarify use of `std::move(*this)` & similar situations

Wumpf:
Originally I wanted to remove superfluous `std::move` which can prevent return value optimization (see https://stackoverflow.com/questions/14856344/when-should-stdmove-be-used-on-a-function-return-value or https://developers.redhat.com/blog/2019/04/12/understanding-when-not-to-stdmove-in-c) buuuut I ended up backpadelling and documenting why we need `std::move` in these places anyways: `return *this` would cause a copy since `*this` is never a temporary

(C++23 (!!) will allow handling this better, with `Deducing this` which then allows perfect forwarding of `this`, see for instance https://stackoverflow.com/questions/73043766/return-type-for-this-in-a-builder-class-method-lvalue-vs-rvalue-refence)

### DIFF
--- a/CODE_STYLE.md
+++ b/CODE_STYLE.md
@@ -191,7 +191,7 @@ class Rect {
 
     Rect with_color(Color color) && {
         _color = color;
-        return std::move(*this);
+        return std::move(*this); // `*this` is always an lvalue, so we have to move it to avoid a copy.
     }
 }
 ```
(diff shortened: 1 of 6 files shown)


teh-cmc (approved):

teh-cmc:
Piggybacking on this in order to fix `main`.

teh-cmc:
@rerun-bot full-check

PR #8890, opened 2025-02-03
Expose creating and querying index for a ReDap collection in python SDK

zehiko:
### What

Basic plumbing that enables creating and querying vector and FTS index. Joint effort from @jleibs  and me.

### Testing done
Ran the latest ``Indexing`` notebook successfully.


examples/python/remote/metadata.py
```diff
@@ -68,3 +72,8 @@
         schema = conn.get_recording_schema(id)
         for column in schema:
             print(column)
+    elif args.subcommand == "create-index":
```
jleibs, line 75:
```suggestion
    elif args.subcommand == "create-vector-index":
```

examples/python/remote/metadata.py
```diff
@@ -68,3 +72,8 @@
...
+        column = rr.dataframe.ComponentColumnSelector(args.entity_path, args.index_column)
+        index = rr.dataframe.IndexColumnSelector("log_tick")
```
jleibs, line 77: Might as well make this and the collection name configurable as well.

rerun_py/src/remote.rs
```diff
@@ -360,6 +368,336 @@ impl PyStorageNodeClient {
...
+    /// -------
+    /// pa.RecordBatchReader
+    ///     The results of the query.
+    #[pyo3(signature = (
+            collection,
+            query,
+            column,
+            top_k,
+        ))]
+    fn query_vector_index(
```
jleibs, line 541: I'm realizing here and elsewhere that `query` isn't really the right term. Propose we swap this with "search":
https://github.com/rerun-io/dataplatform/issues/185

jleibs (approved):

PR #8892, opened 2025-02-03
Deprecate Python's `log_components`

Wumpf:
This method was already undocumented and it doesn't quite fit in the overall story of how partial updates should be done.

it remains as an implementation detail (as `_log_components`) but is deprecated from the public API

Took the opportunity to make some structure improvements to the migration guide

### DIFF
--- a/rerun_py/rerun_sdk/rerun/_log.py
+++ b/rerun_py/rerun_sdk/rerun/_log.py
@@ -160,7 +160,7 @@ def log(
                 f"but got {type(entity)} instead."
             )
 
-    log_components(
+    _log_components(
         entity_path=entity_path,
         components=components,
         static=static,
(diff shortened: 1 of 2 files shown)


rerun_py/rerun_sdk/rerun/_log.py
```diff
@@ -167,6 +168,10 @@ def log(
     )
 
 
+@deprecated(
+    """Use `log` with partial update APIs instead.
+  See: https://www.rerun.io/docs/reference/migration/migration-0-22 for more details."""
```
abey79, line 173:
```suggestion
  See: https://www.rerun.io/docs/reference/migration/migration-0-22?speculative-link for more details."""
```

abey79 (approved):
thanks

PR #8897, opened 2025-02-03
Fix handling null timestamps in the dataframe

### LINKED ISSUE #8893: Unexpected dataframe for custom timeline
This is either a faulty dataframe being returned, or a faulty handling of it by the dataframe view.

Repro:

- Run script below
- Add dataframe view
- Set dataframe timeline to `log_time`

```python
import rerun as rr

rr.init("rerun_example_df_bug")
rr.connect_tcp()

rr.log("stuff", rr.Points3D([[0, 0, 0], [1, 1, 1], [2, 4, 2]]))

rr.set_time_sequence("some_timeline", 10)

rr.log("other_stuff", rr.Points3D([[10, 0, 0], [1, 11, 1], [2, 2, 12]]))
```

Image

abey79:
### Related

- Closes #8893 

### What

Since #8639, the dataframe view would break with a timeline that contains null, which can easily happen with code as the repro in #8893. Previously, the dataframe was quite possibly displaying garbage instead of null timestamps.

This PR now properly handles nulls in timelines.

### DIFF
--- a/crates/store/re_chunk/src/chunk.rs
+++ b/crates/store/re_chunk/src/chunk.rs
@@ -6,7 +6,7 @@ use arrow::{
         Array as ArrowArray, ArrayRef as ArrowArrayRef, ListArray as ArrowListArray,
         StructArray as ArrowStructArray, UInt64Array as ArrowUInt64Array,
     },
-    buffer::ScalarBuffer as ArrowScalarBuffer,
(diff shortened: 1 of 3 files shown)


Wumpf:
added to changelog with slightly different title since this may fix things for folks even prior to the linked regression

Wumpf (approved):
👍

PR #8900, opened 2025-02-03
Add snapshot test to cover for the case of timeline with nulls

abey79:
Some tests to go along #8893


crates/viewer/re_view_dataframe/tests/snapshots/null_timeline.png
abey79, line 1: arf, I need to add some background there so it looks a bit better

teh-cmc (approved):

PR #8901, opened 2025-02-03
Rename the index APIs from `query` to `search`

jleibs:
Cleaning up the APIs to replace query with search and.

Also replace collection w/ entry where appropriate.

### DIFF
--- a/crates/store/re_protos/proto/rerun/v0/remote_store.proto
+++ b/crates/store/re_protos/proto/rerun/v0/remote_store.proto
@@ -9,16 +9,17 @@ service StorageNode {
     rpc Query(QueryRequest) returns (stream DataframePart) {}
     rpc FetchRecording(FetchRecordingRequest) returns (stream rerun.common.v0.RerunChunk) {}
 
-    rpc CreateCollectionIndex(CreateCollectionIndexRequest) returns (CreateCollectionIndexResponse) {}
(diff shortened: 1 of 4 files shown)


crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -9,16 +9,17 @@ service StorageNode {
...
     // - 'resource_id' column with the id of the resource
     // - timepoint column with the values representing the points in time
     // where index query matches. What time points are matched depends on the type of
     // index that is queried. For example for vector search it might be timepoints where
-    // top-K matches are found within *each* resource in the collection. For inverted index
+    // top-K matches are found within *each* resource in the indexed entry. For inverted index
     // it might be timepoints where the query string is found in the indexed column
     // - 'data' column with the data that is returned for the matched timepoints
-    rpc QueryCollectionIndex(QueryCollectionIndexRequest) returns (stream DataframePart) {}
+    rpc SearchIndex(SearchIndexRequest) returns (stream DataframePart) {}
```
zehiko, line 22: curious why "search" over "query"? although I do kinda like it. 

as a side note I had "collection" in the names as I thought we might have these per resource / entry, but I'm sure we'll figure it out as we go.

zehiko (approved):

Issue #6344, opened 2024-05-16
Include blueprint components in the `check_all_components_ui` checklist


abey79:
Currently it covers only `rr.components`. It should likely cover `rr.blueprint.components` as well.

PR #8905, opened 2025-02-04
Get rid of `check_all_components_ui`

teh-cmc:
Any reason we still have this? Wasn't this fully automated already?

* Closes https://github.com/rerun-io/rerun/issues/6344


abey79 (approved):
Yeah let's remove this!

PR #8908, opened 2025-02-04
Support unknown timeline in dataframe view

### LINKED ISSUE #8712: Dataframe view handling of timeline is broken
The timeline view appears to incorrectly behave when the timeline it's configured with does not exist. Unfortunately, that's a very common situation:

1) run some logging script with a custom timeline
2) configure a dataframe view set on that custom timeline
3) run the logging script again
4) 💥 

What seem to happen is that when the 2nd recording is created but before the first data reaches the viewer, the dataframe view doesn't see the timeline it's configured for and incorrectly handle the situation.

This can be reproduced with the simplest of the script:

```python
import rerun as rr


def main():
    rr.init("rerun_example_bug_dataframe")
    rr.connect_tcp()

    rr.set_time_sequence("another_timeline", 0)

    rr.log("pt", rr.Points3D([0, 0, 0]))



if __name__ == "__main__":
    main()
```

abey79:
### Related

- Fixes #8712 

### What

Initially, the dataframe view was built around the idea that the timeline it's configured for is always valid. It uphold this invariant, it would auto-switch to a known timeline in case the current on is unknown.

This was deeply flawed:
1) Recording has some stuff on timeline "X".
2) User configures a dataframe view with timeline "X"
3) New recording of the same app_id has _nothing_ on timeline "X" (e.g. because some event didn't happen in this run, or just because it takes time for data to be logged on "X")
==> dataframe would auto-switch to another timeline
==> 😤

With this PR, the dataframe now supports unknown timeline, at the cost of:
- displaying a warning in the view UI
- disabling most of the selection panel UI

Note: this PR does NOT introduce a way to specify a yet unknown timeline in the UI. You need to use the blueprint API for that, or set the timeline in a recording of your app id that has that timeline.

image

### DIFF
--- a/crates/viewer/re_view_dataframe/src/view_class.rs
+++ b/crates/viewer/re_view_dataframe/src/view_class.rs
@@ -4,8 +4,9 @@ use re_chunk_store::{ColumnDescriptor, SparseFillStrategy};
 use re_dataframe::QueryEngine;
 use re_log_types::EntityPath;
 use re_types_core::ViewClassIdentifier;
+use re_ui::UiExt;
 use re_viewer_context::{
-    SystemExecutionOutput, ViewClass, ViewClassRegistryError, ViewId, ViewQuery, ViewState,
+    Item, SystemExecutionOutput, ViewClass, ViewClassRegistryError, ViewId, ViewQuery, ViewState,
     ViewStateExt, ViewSystemExecutionError, ViewerContext,
 };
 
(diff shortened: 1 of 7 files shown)


crates/viewer/re_view_dataframe/src/view_class.rs
```diff
@@ -105,12 +106,7 @@ mode sets the default time range to _everything_. You can override this in the s
     ) -> Result<(), ViewSystemExecutionError> {
         let state = state.downcast_mut::<DataframeViewState>()?;
         let view_query = view_query::Query::from_blueprint(ctx, view_id);
-        let Some(view_columns) = &state.view_columns else {
-            // Shouldn't happen, except maybe on the first frame, which is too early
-            // for the user to click the menu anyway.
```
abey79, line 110: that didn't age well

Wumpf (approved):
lgtm!

Issue #8816, opened 2025-01-27
Support loading LeRobot datasets directly in the viewer


nikolausWest:
- We want to make it extremely easy and fast to load and view LeRobot datasets in the Rerun Viewer. 
- We want this to work both for "standard datasets" (video, actions, and state) and any community variations (depth images, point clouds, etc). 
- Prioritize V2, but ideally make it work with as many "in the wild" datasets as possible
- Prioritize putting stuff into viewer, but mainly make it work for the community
   - Could mean you need some python utils
   - Could mean trying to influence LeRobot community / maintainers to add Rerun semantic tags to the data

Something to keep in mind:
- The coordination mechanism between dataloaders is non-existent. All applicable dataloaders always run.
    - This means you should test it with non-LeRobot datasets and how other dataloders react with LeRobot datasest
- Could consider doing this via the dataloder URL

oxkitsune:
Some smaller datasets for easy testing:
- v2 dataset for cooking tasks, contains videos, state, action and rewards: [lerobot/stanford_robocook](https://huggingface.co/datasets/lerobot/stanford_robocook)
- v2 dataset, generic, contains videos, state, action and rewards: [lerobot/pusht](https://huggingface.co/datasets/lerobot/pusht)
- v2 dataset, apple storage robot, contains videos, state, and rewards: [pollen-robotics/apple_storage](https://huggingface.co/datasets/pollen-robotics/apple_storage)
- small v1 dataset: [njkong/se2_gripper_multi_camera](https://huggingface.co/datasets/njkong/se2_gripper_multi_camera)

oxkitsune:
Found a dataset that consists of images instead of videos: [lerobot/pusht_image](https://huggingface.co/datasets/lerobot/pusht_image).

Also would be nice to automatically infer the names of the features for the entities, some datasets contain these.

oxkitsune:
Having #7278 would be really nice for action and observation values, these are stored as single tensors in the LeRobot dataset.

And https://github.com/rerun-io/rerun/issues/6264 would be nice to make navigating the episodes easier

grtlr:
Just a quick heads up: #9249 will allow naming recordings, and I have changed the LeRobot loader to use _Epsiode X_ as a name.

Wumpf:
for all we know we're done. Caveats:
* there might be still datasets with things we can't load
* we eventually want to log more metadata to the recording properties

PR #8909, opened 2025-02-04
Initial LeRobot dataset loader

oxkitsune:
### Related

#8816 

### What

This implements an initial dataloader for LeRobot datasets. This initial version is able to load in basic v2 datasets, with video and observation/action scalars. Each episode is loaded into its own recording under the same application id.

### Still left:

- [x] Load and log images for each frame


crates/store/re_data_loader/src/le_robot/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+            .replace("{episode_chunk:03d}", &format!("{chunk:03}"))
+            .replace("{episode_index:06d}", &format!("{episode_index:06}"))
+            .into())
+    }
+
+    /// Get the path to a video observation for a specific episode index.
+    pub fn video_path(
+        &self,
+        observation_key: &str,
+        episode_index: usize,
```
emilk, line 208: We have both episode indices and chunk indices; both encoded as `usize`, leaving room for mixing them up. It could be worth considering adding strong newtypes for these:

```rs
pub struct EpisodeIndex(pub usize);

pub struct ChunkIndex(pub usize);
```

crates/store/re_data_loader/src/le_robot/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+    Int64,
+}
+
+// TODO(gijsd): Do we want to stream in episodes or tasks?
+#[cfg(not(target_arch = "wasm32"))]
+fn load_jsonl_file<D>(filepath: impl AsRef<Path>) -> Result<Vec<D>, LeRobotError>
+where
+    D: DeserializeOwned,
+{
+    let entries = std::fs::read_to_string(filepath)?
```
emilk, line 256: Let's make sure we include the path in the returned error (using `map_err`)

crates/store/re_data_loader/src/le_robot/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+        .lines()
+        .map(|line| serde_json::from_str(line))
+        .collect::<Result<Vec<D>, _>>()?;
```
emilk, line 259: same

crates/store/re_data_loader/src/le_robot/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    Ok(entries)
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct LeRobotDatasetEpisode {
+    pub episode_index: usize,
+    pub tasks: Vec<String>,
+    pub length: u32,
+}
```
emilk, line 269: docstrngs

crates/store/re_data_loader/src/le_robot/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct LeRobotDatasetTask {
+    pub task_index: usize,
+    pub task: String,
+}
```
emilk, line 275: docstrings

crates/store/re_data_loader/src/le_robot/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+            .replace("{video_key}", observation_key)
+            .into())
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Feature {
+    pub dtype: DType,
+    pub shape: Vec<usize>,
+}
```
emilk, line 237: Docstrings!

(What is the shape of a bool? O.o)

crates/store/re_data_loader/src/loader_lerobot.rs
```diff
@@ -0,0 +1,270 @@
...
+                .ok_or_else(|| anyhow!("failed to get frame index"))?
+                .clone();
+
+            let timeline = re_log_types::Timeline::new_sequence("frame_index");
+            let times: &arrow::buffer::ScalarBuffer<i64> = frame_indices
+                .downcast_array_ref::<Int64Array>()
+                .ok_or_else(|| anyhow!("frame indices are of an unexpected type"))?
+                .values();
+
+            let time_column = re_chunk::TimeColumn::new(Some(true), timeline, times.clone());
```
emilk, line 66: Are we _sure_ it is sorted? How can we be sure?

crates/store/re_data_loader/src/loader_lerobot.rs
```diff
@@ -0,0 +1,270 @@
...
+                            &dataset,
+                            feature_key,
+                            episode_idx,
+                            &timeline,
+                            time_column.clone(),
+                        )?);
+                    }
+                    DType::Image | DType::Int64 | DType::Bool => {
+                        // TODO(gijsd): log images
+                        re_log::warn_once!("Logging for dtype {:?} not implemented", feature.dtype);
```
emilk, line 84:
```suggestion
                        re_log::warn_once!("Loading LeRobot dtype {:?} into Rerun is not yet implemented", feature.dtype);
```

emilk (approved):
Looks like a good start; but is difficult to follow without docstrings

Wumpf (commented):
didn't get far and then got distracted, now coming back it looks like @emilk beat me to it once more 😄. Just a few more comments on structure

crates/store/re_data_loader/Cargo.toml
```diff
@@ -42,10 +44,13 @@ itertools.workspace = true
 notify.workspace = true
 once_cell.workspace = true
 parking_lot.workspace = true
+parquet = { workspace = true, features = ["arrow", "snap"] }
 rayon.workspace = true
 thiserror.workspace = true
 uuid.workspace = true
 walkdir.workspace = true
+serde.workspace = true
+serde_json.workspace = true
```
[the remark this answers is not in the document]
oxkitsune: Mine does too, not sure why it wasn't working 🙃

crates/store/re_data_loader/src/le_robot/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+
+        Ok(std::borrow::Cow::Owned(contents))
+    }
+}
+
+#[derive(Debug)]
+pub struct LeRobotDatasetMetadata {
+    pub info: LeRobotDatasetInfo,
+    pub episodes: Vec<LeRobotDatasetEpisode>,
+    pub tasks: Vec<LeRobotDatasetTask>,
```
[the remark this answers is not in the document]
oxkitsune: These are loaded in from a `jsonl` file, which in an ideal world is ordered by `episode_index` and `task_index`, but since the dataset can be constructed with an arbitrary order I feel a bit uneasy relying on the index in the vec.

crates/store/re_data_loader/src/le_robot/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+            metadata,
+        })
+    }
+
+    pub fn read_episode_data(&self, episode_index: usize) -> Result<RecordBatch, LeRobotError> {
+        if !self
+            .metadata
+            .episodes
+            .iter()
+            .any(|episode| episode.episode_index == episode_index)
```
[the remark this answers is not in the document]
oxkitsune: > These are loaded in from a jsonl file, which in an ideal world is ordered by episode_index and task_index, but since the dataset can be constructed with an arbitrary order I feel a bit uneasy relying on the index in the vec.

That's currently not guaranteed no, however I think I can sort them after loading in to avoid this silly look up

crates/store/re_data_loader/src/lib.rs
```diff
@@ -14,6 +14,11 @@ mod loader_archetype;
 mod loader_directory;
 mod loader_rrd;
 
+#[cfg(not(target_arch = "wasm32"))]
+pub mod le_robot;
```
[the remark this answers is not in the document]
oxkitsune: yeaaa, i was anticipating more files in the module when I made it. Turned it into a single file again (with consistent _ placement)

crates/store/re_data_loader/src/le_robot/mod.rs
oxkitsune: The shape directly correlates to the shape of the pytorch tensor that LeRobot creates when loading the dataset.
So for a bool this would be `[1,]`, for a 224x224 rgb image this would be `[224, 224, 3]`.

crates/store/re_data_loader/src/loader_lerobot.rs
oxkitsune: We're not, and there's no harm in sorting these

PR #8912, opened 2025-02-04
Fix action button showing up for recording in app id tooltip

### LINKED ISSUE #8902: Remove Button Overlaps with Recording Size Units in Recordings Tab
**Describe the bug**
The remove button for recordings under an application id overlaps with the recording size units.

**To Reproduce**
1. Open any recording
2. Hover over the application ID in the recordings tab to get the "Loaded recordings" tooltip.
3. The unit of the recording size will overlap with the button to remove the recording from the app.

**Expected behavior**
It should not overlap.

**Screenshots**
Image

 - OS: macOS Sequoia 15.0.1

**Rerun version**
```
$ rerun --version
0.22.0-alpha.1+dev
```

abey79:
### Related

- Fixes #8902 

### What

The app id tooltip no longer has action buttons on recordings in its tooltip:

image

The button is still there in the selection panel though, which now truncates properly:

image

### DIFF
--- a/crates/viewer/re_data_ui/src/app_id.rs
+++ b/crates/viewer/re_data_ui/src/app_id.rs
@@ -50,7 +50,7 @@ impl crate::DataUi for ApplicationId {
                 ui.add_space(8.0);
                 ui.strong("Loaded recordings for this app");
                 for entity_db in recordings {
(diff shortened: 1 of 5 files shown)


Wumpf (approved):
makes sense

PR #8913, opened 2025-02-04
Fix memory leaks when closing recordings

### LINKED ISSUE #8858: Opening & closing recording leaks a memory every time
Haven't dived in yet how much exactly leaks, but when I open and close arkitscenes on 0.21 repeatedly the memory usage goes up.
The first time I expect some memory to linger because some of the caches aren't associated with recordings (for instance the cpu/gpu belts) but most should be by now. Also, this behavior continuous after many more attempts.

teh-cmc:
All the lazily registered subscribers need to know about this.

* Fixes #8858 

## Test

Load 50 DNA recordings into a viewer, then close them all.

### Results

Before:
image

After:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,6 @@ icd.json
 
 .pixi/*
 justfile
+
+# heaptrack files
+*.zst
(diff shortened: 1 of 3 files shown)


Wumpf (approved):
nice, makes sense!

PR #8914, opened 2025-02-04
Switch .vscode extension to dependi

jleibs:
The `serayuzgur.crates` extension is deprecated and causes a warning in vs code.


Wumpf (approved):
thanks!

PR #8916, opened 2025-02-04
Implement snapshot test for 2D draw order

teh-cmc:
Title.

I confirm that the ordering on the snapshot looks correct (i didnt bother matching exactly the original colors).


crates/viewer/re_view_spatial/tests/draw_order.rs
```diff
@@ -0,0 +1,226 @@
...
+            )
+        });
+    }
+
+    let view_id = setup_blueprint(&mut test_context);
+    run_view_ui_and_save_snapshot(
+        &mut test_context,
+        view_id,
+        "draw_order",
+        egui::vec2(300.0, 150.0) * 2.0,
```
Wumpf, line 142: curious resolution

Wumpf (approved):
great to have this finally automated! \o/

crates/viewer/re_view_spatial/tests/draw_order.rs
teh-cmc: The one I copy pasted felt too small for this... in fact it felt 2x too small :x

Wumpf:
Ci produces slightly different image, might be that we have to ramp up the threshold or do something about re_renderer's anti-aliasing settings during testing 🤔
I'll run on different machines later to see what the diff is (we unfortunately don't get image diffs from ci yet :/)

Wumpf:
oh wait a second, it's not complaining about _this_ test, but about `unknown_timeline_view_ui.diff.png` Oo
merging in main...

teh-cmc:
Hmm, surely these logs have something to do with it :thinking: digging...
image

Wumpf:
that run looks a lot "better" 😄

teh-cmc:
I've reproduced the logs locally by adding `re_log::setup_logging();` to the test itself -- we were probably just lucky to get them on the CI because some other test did.

Wumpf:
Huh I thought I've put this on `TextContext::default` but apparently that never made it in

teh-cmc:
> Huh I thought I've put this on `TextContext::default` but apparently that never made it in

Looks like an attempt has been made for sure:
```rust
        // We rely a lot on logging in the viewer to identify issues.
        // Make sure logging is set up if it hasn't been done yet.
        let _ = env_logger::builder().is_test(true).try_init();
```

Maybe I just replace that with a call to `setup_logging`.

teh-cmc:
> Maybe I just replace that with a call to `setup_logging`.

That seems to work nicely.

Wumpf (approved):
extension trait & log change lgtm

Issue #8894, opened 2025-02-03
Add example/snippets demonstrating `set_time_ctrl`


Wumpf:
we can't really advertise it if we don't demonstrate it

Issue #8874, opened 2025-01-30
Add notebook support for `set_active_recording_id`


jleibs:
- https://github.com/rerun-io/rerun/pull/8809 added support for the time controls

However, would be extremely nice to be able to swap which recording is active as well.

PR #8917, opened 2025-02-04
More Viewer control in notebooks (recording id, panel states)

jprochazk:
### Related

* Closes https://github.com/rerun-io/rerun/issues/8874
* Closes https://github.com/rerun-io/rerun/issues/8894

### What

This PR introduces the following _notebook-only_ APIs:

- `Viewer.set_active_recording`, allowing the user to programmatically choose the active recording in multi-recording scenarios.
- `Viewer.update_panels`, exposing the ability to partially update the state of panels.
- A `use_global_recording` param in the `rerun.notebook.Viewer` constructor, which allows the `Viewer` to be displayed in a "clean" state, without implicitly capturing any data from any recording. Recordings can then be added later using the existing `Viewer.add_recording` method.

I also discovered that our build script was subtly broken for a while now :grimacing: specifically, the `CLOSURE_DTORS` finalizer would throw an exception because `wasm` was `null`. Luckily it didn't cause any issues except for there being some error spam in the console.


rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -190,6 +197,78 @@ def _repr_mimebundle_(self, **kwargs: dict) -> tuple[dict, dict] | None:  # type
     def _repr_keys(self):  # type: ignore[no-untyped-def]
         return self._viewer._repr_keys()
 
+    def update_panels(
```
jprochazk, line 200: A "partial update" API felt the most intuitive to me. This means that we need a way to differentiate between "don't update this panel state" (`None`) and "please reset this panel state" (`default`).

examples/python/notebook/cube.ipynb
jleibs, line 1: It's nice to have the two recordings look different.

Maybe just making the cubes different colors:
```
viewer = rr.notebook.Viewer(use_global_recording=False)
viewer.display()

recordings = [
    rr.new_recording("rerun_example_time_ctrl", recording_id="example_a"),
    rr.new_recording("rerun_example_time_ctrl", recording_id="example_b"),
]

for rec in recordings:
    viewer.add_recording(rec)

rec_colors = {
    'example_a': [0, 255, 0],
    'example_b': [255, 0, 0]
}

STEPS = 100
twists = math.pi * np.sin(np.linspace(0, math.tau, STEPS)) / 4
for t in range(STEPS):
    cube = build_color_grid(10, 10, 10, twist=twists[t])
    for rec in recordings:
        rec.set_time_sequence("step", t)
        rec.log("cube", rr.Points3D(cube.positions, colors=rec_colors[rec.get_recording_id()], radii=0.5))
```

jleibs (approved):
Looks nice!

There does seem to be a lurking race-condition in here. The first recording only gets its time adjusted like half the time for me.

PR #8918, opened 2025-02-04
Set default log level to `warn` now that notification center is in place

### LINKED ISSUE #7815: Rust logs output (at info level) shows up as Red in jupyter notebook
For example:
image

When running in the context of a notebook (or python in general) it could be nice if rust logging data was routed back by way of the python logging framework so that it can be configured via `logging.basicConfig`

oxkitsune:
### Related

Closes #7815 

### What

Set the default log level to `warn`, now that we have the notification center.


Wumpf (approved):
👍

PR #8921, opened 2025-02-05
Add some missing gitignore expressions

jleibs:
### What

No reason to risk accidentally committing these.


.gitignore
```diff
@@ -34,6 +34,7 @@ __pycache__
 *.pyd
 *.so
 **/.pytest_cache
+**/.ipynb_checkpoints/
```
drauschenbach, line 37: I also observed the need for this.

PR #8923, opened 2025-02-05
Don't compare arrows3d_simple C++ snippet b/c floats

Wumpf:
I give up! This previously was thought to be fixed by https://github.com/rerun-io/rerun/pull/8867 which made it work on my local windows machine. But on CI we still have discrepancies to Rust

### DIFF
--- a/docs/snippets/snippets.toml
+++ b/docs/snippets/snippets.toml
@@ -17,35 +17,53 @@
 # Tip: use UTF8 NBSPs (https://unicode-explorer.com/c/00A0) as needed.
 [snippets_ref]
 features = [
-  ["Update rows", [
-    "archetypes/scalar_row_updates",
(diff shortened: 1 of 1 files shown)


docs/snippets/snippets.toml
```diff
@@ -17,35 +17,53 @@
...
+    ],
+  ],
+  [
+    "Query dataframes",
+    [
+      "reference/dataframe_query",
+      "reference/dataframe_view_query",
+    ],
+  ],
 ]
```
Wumpf, line 67: my format-on-save demanded this 🤷

emilk: Why isn't this covered by `pixi run toml-fmt`  ?

emilk (approved):

PR #8926, opened 2025-02-05
Update egui to 0.31

lucasmerlin:
### What

Update egui and the relevant crates


crates/viewer/re_view_map/src/visualizers/geo_line_strings.rs
```diff
@@ -126,7 +126,7 @@ impl GeoLineStringsVisualizer {
                 .iter()
                 .flat_map(|(_, batch)| batch.lines.iter())
                 .flatten()
-                .map(|pos| (pos.lat(), pos.lon())),
+                .map(|pos| (pos.lat(), pos.lng())),
```
lucasmerlin, line 129: lat and lng are now deprecated, I could change them to .y() and .x() but I feel that order would be confusing so I've kept it like this for now

emilk (approved):
LGTM if you've tested the geo/walkers changes

crates/viewer/re_view_map/src/visualizers/geo_line_strings.rs
lucasmerlin: Turns out we disallow deprecated things, so I've changed it to y and x

Issue #8925, opened 2025-02-05
Snapshot test context tries to display vanilla images as depth & segmentation images


teh-cmc:
E.g. this:
```rust
test_context.log_entity("2d_layering/middle_blue".into(), |builder| {
    let mut image = Array::<u8, _>::zeros((192, 192, 3).f());
    image.slice_mut(s![.., .., 2]).fill(255);

    builder.with_archetype(
        RowId::new(),
        TimePoint::default(),
        &re_types::archetypes::Image::from_color_model_and_tensor(
            re_types::datatypes::ColorModel::RGB,
            image,
        )
        .unwrap()
        .with_draw_order(1.1),
    )
});
```
yields:
```
[2025-02-05T10:14:38Z ERROR re_view_spatial::visualizers::utilities::textured_rect] Failed to create texture for "/2d_layering/middle_blue": Depth image does not support the ColorModel RGB
[2025-02-05T10:14:38Z ERROR re_view_spatial::visualizers::utilities::textured_rect] Failed to create texture for "/2d_layering/middle_blue": Segmentation image does not support the ColorModel RGB
```

Running the same code using the normal viewer does not trigger these logs (??).

PR #8927, opened 2025-02-05
Handle indicators, overrides & view states in view tests

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/8925

### What

Main problem was that we ignored indicators previously, meaning we didn't apply proper visualizer selection before.
As a sideeffect we now would also handle overrides and no longer build view states adhoc in tests

draft until https://github.com/rerun-io/rerun/pull/8916 lands (which can also use the changed view state access now)


teh-cmc (approved):

teh-cmc:
This fails on unrelated snapshots -- I'm gonna take a bet.

teh-cmc:
Yeah this passes locally after rebasing main.

PR #8933, opened 2025-02-05
Filter entities in the UI (part 9): Support `/` in queries

### LINKED ISSUE #8903: Properly support `/` in filter queries
We discussed a plan with @Wumpf.

The plan is:

- parse the input query into keyword as below.
- update match logic accordingly
- for the highlight ranges:
  - perform ranges extraction for the full hierarchy in leaf items
  - accumulate (HashSet?) match ranges in the `hierarchy` vector
  - merge ranges after traversing children, based on `hierarchy.last()`

```rust
enum Keyword {
        parts: Vec<&str>,

		/// e.g. if the query is "/a/b/c"
        first_part_must_fully_match: bool,

	    /// e.g. if the query is "a/b/c/"
        last_part_must_fully_match: bool,
}
```

abey79:
### Related

- Closes #8903
- Part of #8586 
- Part of a series of PR:
  - #8645
  - #8652
  - #8654
  - #8672
  - #8706
  - #8728
  - #8795
  - #8863
  - #8886
  - #8933

### What

This PR supports `/` in filter query. Semantics are now as follows:

- Queries are made of whitespace separated keyword.
- Paths are not matched unless they match all keywords individually at least once.
- Keywords may have 1 or more  `/` separated "parts".
- Single part keyword behave as before: paths are matched if they contain that keyword anywhere at least once.
- Single part keyword can now be prefixed and/or postfixed with a slash, eg. "/fir". In that case, the keyword part must be found at the beginning and/or end of at least one path part to match (e.g. "/fir" matches "/my/first/entity" but doesn't match "xxfire/entity").
- Multi-part keyword (e.g. "a/b/c") match paths that contains corresponding, matching sequence of part (e.g. "/xxxxa/b/cxxxxx"). Such keywords can also be prefixed and/or postfixed with a "/", with the same semantics.

image


Wumpf (approved):
pre-approved with comments during a call-based review.

absolutely love this, both result & execution

Issue #8940, opened 2025-02-05
Graph view snapshots are broken


teh-cmc:
`main` is not complaining for no reasons, this is what you see if you try to update the graph view snapshots right now:
![Image](https://github.com/user-attachments/assets/16c44b6d-a68e-4f40-82f0-c717b0b8dc97)

Wumpf:
`main` started complaining by now, phew!
https://github.com/rerun-io/rerun/actions/runs/13161352407/job/36730623545
``` 
Image did not match snapshot. Diff: 115, "/Users/runner/work/rerun/rerun/crates/viewer/re_view_graph/tests/snapshots/coincident_nodes.diff.png". 
```

PR #8941, opened 2025-02-05
Fix graph view test

Wumpf:
* Fixes https://github.com/rerun-io/rerun/issues/8940

By using test-log_entity API with user facing archetypes so proper indicators etc. are set.

This broke in https://github.com/rerun-io/rerun/pull/8927 since this PR makes it so that indicator based visualizer filtering is active - curiously without that we'd just use all the available visualizers which in our tests _so far_ worked out fine.
Still totally unclear why we intermittently didn't see this breaking (it luckily breaks main right now)


abey79 (approved):
Nice!

PR #8942, opened 2025-02-05
Reduce logspam on failed ci tests a bit

Wumpf:
Failed jobs like https://github.com/rerun-io/rerun/actions/runs/13162992946/job/36736204297 are drowning in very very unhelpful
```
 [2025-02-05T17:34:38Z INFO  re_renderer::context] wgpu adapter backend: Vulkan, device_type: Cpu, name: "llvmpipe (LLVM 15.0.7, 256 bits)", driver: "llvmpipe", driver_info: "Mesa 24.2.3 (LLVM 15.0.7)"
[2025-02-05T17:34:38Z WARN  re_renderer::context] Bad software rasterizer detected - expect poor performance and crashes. See: https://www.rerun.io/docs/getting-started/troubleshooting#graphics-issues
```

also the bit about "bad software rasterizer" is clearly false. We love llvmpipe, it's just slow and you better have right version (which one)


abey79 (approved):
Nice!

PR #8944, opened 2025-02-06
Fix nightly ("re_tuid/serde")

abey79:
☝🏻


abey79:
@rerun-bot full-check

Issue #6576, opened 2024-06-14
`Chunk`s should support non-`ListArray` component columns


teh-cmc:
A `Chunk` stores its component columns as `ListArray`s, where each entry in the list-array is a component batch.
```rust
pub struct Chunk {
    // ...
    pub(crate) components: BTreeMap<ComponentName, ArrowListArray<i32>>,
    // ...
}
```

For most cases this is fine, but for small mono-components this is not optimal as we have to spend extra compute and space on list offsets that we don't need.

The worst case, as usual, is `Scalar`s: with a `ListArray<i32>` that's a 50% space overhead we have pay in order to store a... monotonically increasing list of offsets. :unamused:

We are actually not that far from being able to support component columns that are backed directly by their underlying `list.values()` array instead of the list itself, for these cases.
This would allow e.g. the Python SDK to pretty much pass a numpy scalar array as-is, without any `array_to_list_array()` shenanigans.

In effect this means moving to something like this:
```rust
pub struct Chunk {
    // ...
    pub(crate) components: BTreeMap<ComponentName, Either<ArrowListArray<i32>, Box<dyn ArrowArray>>>,
    // ...
}
```
which is really the same as:
```rust
pub struct Chunk {
    // ...
    pub(crate) components: BTreeMap<ComponentName, Box<dyn ArrowArray>>,
    // ...
}
```

There are a few subtleties to get right though, especially when it comes to all the sorting/shuffling/slicing we do, where we basically need to have specialized implementations for `ListArray` vs. everything else.
Nothing unsurmountable though.

- Related: #4184

PR #8945, opened 2025-02-06
Replace `TransportChunk` with `re_sorbet::ChunkBatch`

emilk:
### Related
* Related https://github.com/rerun-io/rerun/issues/6576
* Sibling PR: https://github.com/rerun-io/dataplatform/pull/214

### What
An arrow record batch needs to follow a specific schema to be be compatible with Rerun,
and that schema is defined in `ChunkSchema`. If a record batch matches the schema, it can be converted to a `ChunkBatche.

The schema has:
* One RowId column
* N index (time) columns
* N data (component) columns, all ListArrays

`ChunkBatch::try_from(RecordBatfch)` will automatically wrap data columns in `ListArray` if needed.

```
impl AsRef<ArrowRecordBatch> for ChunkBatch { … }
impl From<ChunkBatch> for ArrowRecordBatch { … }
impl TryFrom<ArrowRecordBatch> for ChunkBatch { … }
```

### TODO
* [x] Fix all TODOs
* [x] Fix rust tests
* [x] Fix python roundtrip tests
* [x] Run full test suite

### For future PRs
* Put the magic string constants into `const`ants
* Introduce `DataframeBatch`
* Rename `ComponentColumnDescriptor` to `DataColumnSchema`, etc
* `ComponentColumnDescriptor` is both used for dataframes and chunks. Resolve somehow


crates/store/re_dataframe/src/snapshots/re_dataframe__query__tests__async_barebones_static.snap
```diff
@@ -12,7 +12,7 @@ expression: "TransportChunk::from(dataframe)"
 │ │ type: "i64"            ┆ type: "Timestamp(ns)"  ┆ type: "List[u32]"            ┆ type: "List[Utf8]"           ┆ type: "List[Struct[2]]"      │ │
 │ │ index_name: "frame_nr" ┆ index_name: "log_time" ┆ component: "example.MyColor" ┆ component: "example.MyLabel" ┆ component: "example.MyPoint" │ │
 │ │ kind: "index"          ┆ kind: "index"          ┆ entity_path: "/this/that"    ┆ entity_path: "/this/that"    ┆ entity_path: "/this/that"    │ │
-│ │                        ┆                        ┆ kind: "data"                 ┆ is_static: "yes"             ┆ kind: "data"                 │ │
+│ │                        ┆                        ┆ kind: "data"                 ┆ is_static: "true"            ┆ kind: "data"                 │ │
```
emilk, line 15: I wanted it to be more JSON-like

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

rerun_py/rerun_sdk/rerun/dataframe.py
```diff
@@ -61,9 +61,9 @@ def __init__(self, metadata: dict[bytes, bytes], col: pa.Array):
     def component_descriptor(self) -> ComponentDescriptor:
         kwargs = {}
         if SORBET_ARCHETYPE_NAME in self.metadata:
-            kwargs["archetype_name"] = "rerun.archetypes" + self.metadata[SORBET_ARCHETYPE_NAME].decode("utf-8")
```
emilk, line 64: Note: this used to be buggy (missing period after `archetypes`)

crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -333,84 +327,32 @@ async fn stream_catalog_async(
 
     re_log::info!("Starting to read...");
     while let Some(result) = resp.next().await {
-        let input = TransportChunk::from(result.map_err(TonicStatusError)?);
-
-        // Catalog received from the ReDap server isn't suitable for direct conversion to a Rerun Chunk:
```
zehiko, line 338: nice to see this gone!

crates/store/re_sorbet/src/chunk_batch.rs
```diff
@@ -51,58 +219,109 @@ impl From<ChunkBatch> for ArrowRecordBatch {
...
+    let mut fields: Vec<ArrowFieldRef> = Vec::with_capacity(num_columns);
+    let mut columns: Vec<ArrowArrayRef> = Vec::with_capacity(num_columns);
 
-    /// The key used to identify a Rerun [`EntityPath`] in the record batch metadata.
-    const CHUNK_METADATA_KEY_ENTITY_PATH: &'static str = "rerun.entity_path";
-}
+    for (field, data) in itertools::izip!(batch.schema().fields(), batch.columns()) {
+        if data
+            .downcast_array_ref::<arrow::array::ListArray>()
+            .is_some()
```
teh-cmc, line 275: This is fairly ambiguous, we do have *components* (e.g. `Blob`) which are themselves lists, but that doesn't necessarily mean they represent a column.

In general, I don't really like the idea of something implicitly casting things to lists like that (for this and other reasons) -- but I'm lacking context, maybe there are good reasons here.

crates/store/re_sorbet/src/chunk_batch.rs
```diff
@@ -51,58 +219,109 @@ impl From<ChunkBatch> for ArrowRecordBatch {
...
+            }
+
+            let data_arrays = sliced.iter().map(|e| Some(e.as_ref())).collect::<Vec<_>>();
+            #[allow(clippy::unwrap_used)] // we know we've given the right field type
+            let list_array: arrow::array::ListArray =
+                re_arrow_util::arrow_util::arrays_to_list_array(
+                    data_field_inner.data_type().clone(),
+                    &data_arrays,
+                )
+                .unwrap();
```
teh-cmc, line 309: I'm surprised this looks this costly/complicated? This should just be a matter of creating a listarray with unit offsets no?
E.g. in pseudo-rust:
```rust
let offsets = std::iter::repeat(1).take(data.len());
let list_array = ListArray::new(data_field, offsets, data, None);
```

crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -333,84 +327,32 @@ async fn stream_catalog_async(
...
-                ArrowField::new("item", field.data_type().clone(), true /* nullable */);
-
-            let data_field = ArrowField::new(
-                field.name().clone(),
-                ArrowDataType::List(Arc::new(data_field_inner.clone())),
-                false, /* not nullable */
-            )
-            .with_metadata(TransportChunk::field_metadata_data_column());
-
-            let mut sliced: Vec<ArrowArrayRef> = Vec::new();
```
teh-cmc, line 379: Ha! So that's where it came from. Yeah we should provide a util in `arrow_util` that shows how to do this much more cheaply.

crates/store/re_sorbet/src/column_schema.rs
```diff
@@ -35,6 +35,16 @@ pub enum ColumnDescriptor {
 }
 
 impl ColumnDescriptor {
+    /// Debug-only sanity check.
```
teh-cmc, line 38: There's nothing in this method that guarantees debug-only behavior though?

crates/top/rerun/src/commands/rrd/filter.rs
```diff
@@ -111,32 +111,40 @@ impl FilterCommand {
                 Ok(msg) => {
                     let msg = match msg {
                         re_log_types::LogMsg::ArrowMsg(store_id, mut msg) => {
-                            if !should_keep_entity_path(&dropped_entity_paths, &msg.batch.schema())
+                            match re_sorbet::ChunkSchema::try_from(msg.batch.schema_ref().as_ref())
```
teh-cmc, line 114: Why is `re_sorbet` being mentioned here? Isn't `re_sorbet` all about Sorbet-formatted chunks, as opposed to this tool which manipulates streams of Rerun-formatted chunks?

crates/store/re_sorbet/src/chunk_batch.rs
```diff
@@ -1,26 +1,194 @@
...
+}
+
+impl MismatchedChunkSchemaError {
+    pub fn custom(s: impl Into<String>) -> Self {
+        Self::Custom(s.into())
+    }
 }
 
-/// The arrow [`ArrowRecordBatch`] representation of a Rerun chunk.
+/// The [`ArrowRecordBatch`] representation of a Rerun chunk.
```
teh-cmc, line 38: I'm getting super confused regarding terminology here: so this is _not_ a Sorbet chunk, this is a Rerun chunk (i.e. single-entity kinda thing), right? Are Sorbet chunks defined somewhere or have we just dropped them for now?

crates/store/re_sorbet/src/lib.rs
```diff
@@ -17,4 +26,14 @@ pub use self::{
         ArrowBatchMetadata, ArrowFieldMetadata, MetadataExt, MissingFieldMetadata,
         MissingMetadataKey,
     },
+    row_id_column_schema::{RowIdColumnDescriptor, WrongDatatypeError},
 };
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum BatchType {
```
teh-cmc, line 33: Haaaaaaaaaaaaaa, so it's _both_ now! okay

crates/store/re_sorbet/src/column_schema.rs
emilk: No guarantees other than the docstring contract

emilk:
@teh-cmc as described in https://github.com/rerun-io/rerun/issues/8744 there are two types of record batches we care about:

* chunks (single-entity)
* dataframes (multi-entity)

This PR implements the first, with the second one coming as a second PR

teh-cmc:
> @teh-cmc as described in #8744 there are two types of record batches we care about:
> 
>     * chunks (single-entity)
> 
>     * dataframes (multi-entity)
> 
> 
> This PR implements the first, with the second one coming as a second PR

I guess most of my confusion comes from the mention of `sorbet` -- all this of this very much looks plain-old Rerun (although I guess plain-old Rerun _can_ do multi-entity records now, which is the new thing).

(That ticket even says: "We will stop using sorbet name until we have cycles to make this a more universal spec.")

crates/store/re_sorbet/src/chunk_batch.rs
emilk: We need to find data columns that should be ListArrays (lists of lists), but aren't. Yes, I think we can do better then this.

crates/store/re_sorbet/src/chunk_batch.rs
emilk: I'll look at it in separate PR (this one is big enough already). But I'll add a TODO.

emilk:
I see "sorbet" as the spec for how we encode Rerun data on-the-wire. As such, it covers both chunks and dataframes.

We don't use `sorbet` as a name in any of the meta tags (yet).

crates/store/re_grpc_client/src/redap/mod.rs
emilk: * https://github.com/rerun-io/rerun/pull/8998

PR #8936, opened 2025-02-05
Move `tests/assets/video` over to git-lfs

teh-cmc:
Title.

PR #8946, opened 2025-02-06
Add `git-lfs` to our CI container

teh-cmc:
Title.

* See it in action in https://github.com/rerun-io/rerun/pull/8936.


emilk:
I think we should wait with messing around with CI until after releasing 0.22

PR #8947, opened 2025-02-06
Improved help view

### LINKED ISSUE #8641: Improve info-hover for views + timeline
## Why
We gathered enough evidences from user questions and chats where they ask for shortkeys or options that are actually present in rerun, but they did not find them. We have also observed that some of them **do** use the ?-panels, but still miss info in them. 
- Main hypothesis is that current info panels are verbose and hard to scan through to quickly find necessary hot keys. 
- inconsistent way of presenting the hotkeys (sometimes in the beginning of the sentence, sometimes in the end)
- Inconsistent description of the same hotkeys for different views
- Inconsistent information, i.e. only hotkeys for 3D View and a very long explanation of options of the view in Dataframe
- Verbose: "right mouse click" could be changed to 1 icon
- Funky spacing and font sizes

## What
Change the way we display the "?-info" on hover for all view types + timeline. 
- 2D View
- 3D View
- Time series view
- Bar chart view
- Map view
- Graph view
- Tensor view
- Text document view
- Dataframe view
- Timeline panel in streams

### What this issue is NOT about:
- Actually changing the hotkeys. That is an orthogonal issue (if an issue at all)
- Other hovers, for example in entity path hovers or selection panel hovers.

lucasmerlin:
### Related
- closes #8641 

### What

Much nicer help popups based on @gavrelina's designs

Screenshot 2025-02-06 at 11 27 28

I've changed some of the texts from the design (in the time series view):
- Place cursor in view -> Move time cursor
- Toggle legend visibility -> Hide/show series

Also, instead of always showing `Cmd/Ctrl` like in the design, I'm showing only the button that is relevant for the users OS

TODO:

- [x] do timeline help


crates/viewer/re_ui/data/icons/shift.png
emilk, line 1: We do have all the modifiers in unicode with the Inter font we use in Rerun; see https://docs.rs/egui/latest/egui/struct.ModifierNames.html#associatedconstant.SYMBOLS

We use then in the Rerun menu: 
image

(even though we should really crank up the contrast on these…)

We should probably think about how we can be consistent with how we show modifiers across the app (but not a blocker for this PR/release imho)

crates/viewer/re_ui/examples/re_ui_example/main.rs
```diff
@@ -479,6 +479,20 @@ impl egui_tiles::Behavior<Tab> for MyTileTreeBehavior {
...
+                        ControlRow::new("Pan", icon_text!(icons::LEFT_MOUSE_CLICK, " + drag")),
+                        ControlRow::new("Zoom", icon_text!("Ctrl / Cmd +", icons::SCROLL)),
+                        ControlRow::new(
+                            "Reset view",
+                            icon_text!("double", icons::LEFT_MOUSE_CLICK),
+                        ),
+                    ])
+                    .ui(ui);
+            });
+
```
lucasmerlin, line 495: Update this to use the Builder Api

crates/viewer/re_view_map/src/map_view.rs
```diff
@@ -102,17 +102,12 @@ impl ViewClass for MapView {
...
-
-- Pan by dragging.
-- Zoom with pinch gesture.
-- Double-click to reset the view."
-            .to_owned()
+    fn help(&self, _egui_ctx: &egui::Context) -> Help<'_> {
+        Help::new("Map view")
+            .docs_link("https://rerun.io/docs/reference/types/views/map_view")
+            .control("Pan", icon_text!(icons::LEFT_MOUSE_CLICK, "+ drag"))
+            .control("Zoom", icon_text!("Ctrl/Cmd +", icons::SCROLL))
```
lucasmerlin, line 109: Modifiers::COMMAND

crates/viewer/re_ui/data/icons/lmc.png
abey79, line 1: All our other icons are fully white, and we taint them in code. I understand these ones are quite specific in how we're going to use them, but I still feel it would be more consistent to do the same (and, yeah, sorry—I should have told you that when we discussed it)

emilk (approved):
Beautiful!

emilk:
Let's merge when green; then ship 0.22

PR #8949, opened 2025-02-06
Checklist for LeRobot dataloader

teh-cmc:
It's a bit of an odd one because of the way the dataloader works, but I would still feel much better with an odd test than no tests at all.

I've uploaded a heavily sampled down `apple_storage` dataset to LFS in order to make this work.

* DNM: requires #8936


teh-cmc:
I'll sample even more aggressively (not sure why I kept 10 rather than.. I dunno, 3 maybe?) then merge this.

PR #8951, opened 2025-02-06
Fix slight parity issue on migration guide

Wumpf:
"before" part was incomplete in some sections, making it look better than it was ;-)


Wumpf (commented):
trust me I know what I'm doing ;-). no point in waiting for review here

PR #8952, opened 2025-02-06
Treat zero-angle rotation as identity when axis is invalid

oxkitsune:
### What

A rotation with a zero angle is always an identity transform, so we can treat it as such


Wumpf (approved):

emilk (approved):

Wumpf:
put this to "exclude from changelog" since this regression didn't show up in any release yet

PR #8956, opened 2025-02-06
Add example for python notebook with partial & columnar updates

Wumpf:
image

So far we crammed everything into `cube.ipynb` so was a bit unsure from where to link it. Not linked anywhere right now but I figure better than not having it.


teh-cmc (approved):
Works on my machine :tm:

PR #8958, opened 2025-02-06
Automated `check_annotations`

teh-cmc:
Title.


crates/viewer/re_view_spatial/tests/snapshots/annotations.png
abey79, line 1: Looks like this one is not actually used in the test? I don't remember if CI covers for that.

teh-cmc: yeah nice catch that's my old test

abey79 (approved):
so good to get rid of these checklist

PR #8960, opened 2025-02-06
Stop forgetting about updating the migration guide link

abey79:
☝🏻


Wumpf (approved):
great catch. Check lgtm, but I read _somewhere_ it doesn't work yet ;)

PR #8962, opened 2025-02-06
Fix broken docs link on text view's help text

Wumpf:
We should add this to a patch release if we do one.


teh-cmc (approved):

abey79:
Duplicates https://github.com/rerun-io/rerun/pull/8963

PR #8965, opened 2025-02-07
DRAFT: Propose new uniform gRPC service for Catalog / StorageNode

jleibs:
This refactors and renames much of the existing gRPC service to bring it inline with the multi-entry/collection catalog model.

The core of the proposal is that the Storage Node will implement a subset of these APIs and act as a single-entry catalog exposing a single collection.

This means the viewer can connect directly to a storage node and still have a reasonable experience that benefits from the catalog-view work-stream.
- For example, we should have a collection view that allows the viewer to filter a view of the PropertyTable and from there navigate to individual recordings.

### DIFF
--- a/crates/store/re_protos/proto/rerun/v0/catalog.proto
+++ b/crates/store/re_protos/proto/rerun/v0/catalog.proto
@@ -1,0 +1,249 @@
+syntax = "proto3";
+
+package rerun.redap.v0;
+
+import "rerun/v0/common.proto";
+
+// The RerunCatalog service is a common API set exposed by:
+// - The Rerun Catalog Server
+// - Any Rerun Storage Node
+//
+// Some Important terminology:
+// - Resource: A resource is any data that can be mapped to a single recording or table in the Rerun data model.
+//   - Examples: A single Rerun recording, an MCAP file, parquet table, etc.
+//   - Every resource across the catalog has a unique identifier.
+// - Collection: A collection contains multiple resources of a similar type.
+//   - Every collection includes a "Property Table" of properties associated with the resources in a collection.
+// - Catalog Entry: An entry within a catalog is a named interface that facilitates access to resources or collections.
+//   - Examples: A collection of recordings, a table of training data, a view created from a query.
+//
+// A Rerun Storage Node generally acts as a single-entry catalog, serving all
+// the recordings stored in the node as a single collection along with a single
+// property-table.  This makes it convenient to directly connect to a Storage Node
+// with the Rerun Viewer in minimal deployments without the need to set up a
+// separate catalog server.
+//
+// This also extends to the Viewer itself, which contains a minimal implementation of a fully
+// in-memory storage node. Each recording sent directly to the viewer from the Rerun SDK will
+// be stored in the viewer's local collection.
+service RerunCatalog {
+    // ---------------- Shared Catalog APIs ------------------
+
+    // List all the Entries in the catalog
+    //
+    // The storage node will only return a single entry for its collection.
+    rpc ListEntries(ListEntriesRequest) returns (ListEntriesResponse) {}
+
+    // ---------------- Catalog-Only Catalog APIs ------------------
+    // These APIs are not implemented in the Storage Node
+
+    // Add a new entry to the catalog
+    rpc AddEntry(AddEntryRequest) returns (AddEntryResponse) {}
+
+    // Remove an entry from the catalog
+    rpc RemoveEntry(RemoveEntryRequest) returns (RemoveEntryResponse) {}
+    // Remove an entry from the catalog
+
+    // ---------------- Shared Collection APIs ------------------
+    // APIs for interacting with catalog entries which are collections
+    //
+    // All of the Request messages include a `catalog_entry` field. This can be omitted
+    // when interacting with a Storage Node.
+
+    // Get the Schema of the property table for a particular catalog entry
+    rpc GetPropertyTableSchema(GetPropertyTableSchemaRequest)
+        returns (GetPropertyTableSchemaResponse) {}
+
+    // Set the properties of one or more resources within a catalog entry
+    //
+    // May fail if the catalog entry is not writable.
+    rpc UpdateResourceProperties(UpdateResourcePropertiesRequest)
+        returns (UpdateResourcePropertiesResponse) {}
+
+    // Retrieve the table of properties for all resources stored in the collection.
+    //
+    // Possibly filtered / column projected.
+    //
+    // The schema of the returned Dataframe should match the results of `PropertyTableSchema`.
+    rpc GetResourceProperties(GetResourcePropertiesRequest)
+        returns (stream rerun.common.v0.DataframePart) {}
+
+    // ---------------- Catalog-Only Collection APIs ------------------
+
+    // Register one or more resources with catalog entry
+    //
+    // This API is only supported for some catalog entries.
+    rpc RegisterResources(RegisterResourcesRequest) returns (rerun.common.v0.DataframePart) {}
+
+    // Remove a resource from the a manually-managed catalog entry
+    //
+    // This API is only supported for some catalog entries.
+    rpc UnregisterResource(UnregisterResourcesRequest) returns (UnregisterResourceResponse) {}
+
+    // Remove all resources from a manually-managed catalog entry
+    //
+    // This API is only supported for some catalog entries.
+    rpc UnregisterAllResources(UnregisterAllResourcesRequest)
+        returns (UnregisterAllResourcesResponse) {}
+
+    // Create an index for a specific column in the catalog entry
+    rpc CreateIndex(CreateIndexRequest) returns (CreateIndexResponse) {}
+
+    // Search over an existing index  in the catalog entry.
+    //
+    // The response to `SearchIndex` a RecordBatch with 3 columns:
+    // - 'resource_id' column with the id of the resource
+    // - timepoint column with the values representing the points in time
+    // where index query matches. What time points are matched depends on the type of
+    // index that is queried. For example for vector search it might be timepoints where
+    // top-K matches are found within *each* resource in the indexed entry. For inverted index
+    // it might be timepoints where the query string is found in the indexed column
+    // - 'data' column with the data that is returned for the matched timepoints
+    rpc SearchIndex(SearchIndexRequest) returns (stream rerun.common.v0.DataframePart) {}
+
+    // ---------------- Resource APIs ------------------
+    // These APIs do not need to be mediated by a catalog entry
+    // The resource-id is sufficient to uniquely locate the resource within the catalog.
+
+    rpc GetResourceSchema(GetResourceSchemaRequest) returns (GetResourceSchemaResponse) {}
+
+    rpc ListChunks(ListChunksRequest) returns (stream ChunkInfo) {}
+
+    rpc GetChunks(stream GetChunkRequest) returns (stream rerun.common.v0.RerunChunk) {}
+
+    // Run a Query against a single resource in the storage node
+    rpc ResourceQuery(ResourceQueryRequest) returns (stream rerun.common.v0.DataframePart) {}
+}
+
+message CatalogEntry {
+    string name = 1;
+}
+
+// ---------------- Catalog APIs ------------------
+
+// TODO
+message ListEntriesRequest {}
+message ListEntriesResponse {}
+
+// TODO
+message AddEntryRequest {}
+message AddEntryResponse {}
+
+// TODO
+message RemoveEntryRequest {}
+message RemoveEntryResponse {}
+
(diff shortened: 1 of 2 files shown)


crates/store/re_protos/proto/rerun/v0/catalog.proto
```diff
@@ -0,0 +1,249 @@
...
+message UpdateResourcePropertiesRequest {
+    CatalogEntry entry = 1;
+
+    // This dataframe must contain a column indicating the resource_id
+    rerun.common.v0.DataframePart properties = 2;
+}
+
+message UpdateResourcePropertiesResponse {}
+
+message GetResourcePropertiesRequest {
```
jleibs, line 156: A collection might contain a huge number of resources. Maybe millions in the case of something like a large-scale simulation.

This API maybe needs to support options for limit/pagination.

crates/store/re_protos/proto/rerun/v0/catalog.proto
```diff
@@ -0,0 +1,249 @@
...
+    // List all the Entries in the catalog
+    //
+    // The storage node will only return a single entry for its collection.
+    rpc ListEntries(ListEntriesRequest) returns (ListEntriesResponse) {}
+
+    // ---------------- Catalog-Only Catalog APIs ------------------
+    // These APIs are not implemented in the Storage Node
+
+    // Add a new entry to the catalog
+    rpc AddEntry(AddEntryRequest) returns (AddEntryResponse) {}
```
grtlr, line 41: Nit: All the `Entry|Entries` endpoints somehow lack descriptiveness.

crates/store/re_protos/proto/rerun/v0/catalog.proto
```diff
@@ -0,0 +1,249 @@
...
+package rerun.redap.v0;
+
+import "rerun/v0/common.proto";
+
+// The RerunCatalog service is a common API set exposed by:
+// - The Rerun Catalog Server
+// - Any Rerun Storage Node
+//
+// Some Important terminology:
+// - Resource: A resource is any data that can be mapped to a single recording or table in the Rerun data model.
```
grtlr, line 12: We'll probably also need concepts similar to _sources_ and _sinks_ to describe the different format conversions/coercions that might happen, e.g sending an `.mcap` file to a `.rrd` sink forces a conversion on write.

crates/store/re_protos/proto/rerun/v0/catalog.proto
```diff
@@ -0,0 +1,249 @@
...
+//   - Examples: A single Rerun recording, an MCAP file, parquet table, etc.
+//   - Every resource across the catalog has a unique identifier.
+// - Collection: A collection contains multiple resources of a similar type.
+//   - Every collection includes a "Property Table" of properties associated with the resources in a collection.
+// - Catalog Entry: An entry within a catalog is a named interface that facilitates access to resources or collections.
+//   - Examples: A collection of recordings, a table of training data, a view created from a query.
```
grtlr, line 18: Is that the same _Examples_ as mentioned above?

crates/store/re_protos/proto/rerun/v0/catalog.proto
```diff
@@ -0,0 +1,249 @@
...
+
+package rerun.redap.v0;
+
+import "rerun/v0/common.proto";
+
+// The RerunCatalog service is a common API set exposed by:
+// - The Rerun Catalog Server
+// - Any Rerun Storage Node
+//
+// Some Important terminology:
```
grtlr, line 11: Maybe add the definition of _Property_ (which I think is what we used to call  _Metadata_, right?).

crates/store/re_protos/proto/rerun/v0/catalog.proto
```diff
@@ -0,0 +1,249 @@
...
+// - Resource: A resource is any data that can be mapped to a single recording or table in the Rerun data model.
+//   - Examples: A single Rerun recording, an MCAP file, parquet table, etc.
+//   - Every resource across the catalog has a unique identifier.
+// - Collection: A collection contains multiple resources of a similar type.
+//   - Every collection includes a "Property Table" of properties associated with the resources in a collection.
```
grtlr, line 16: Maybe clarify: _Properties_ are what used to be called _Metadata_, right?

crates/store/re_protos/proto/rerun/v0/catalog.proto
```diff
@@ -0,0 +1,249 @@
...
+    // Remove an entry from the catalog
+
+    // ---------------- Shared Collection APIs ------------------
+    // APIs for interacting with catalog entries which are collections
+    //
+    // All of the Request messages include a `catalog_entry` field. This can be omitted
+    // when interacting with a Storage Node.
+
+    // Get the Schema of the property table for a particular catalog entry
+    rpc GetPropertyTableSchema(GetPropertyTableSchemaRequest)
```
grtlr, line 54: Nit: Maybe just use `Properties` instead of `PropertyTable`? Or is it important to be explicit about the table aspect here?

crates/store/re_protos/proto/rerun/v0/catalog.proto
```diff
@@ -0,0 +1,249 @@
...
+
+message GetPropertyTableSchemaResponse {
+    // The schema of the property table for the given catalog entry.
+    rerun.common.v0.Schema schema = 1;
+}
+
+message UpdateResourcePropertiesRequest {
+    CatalogEntry entry = 1;
+
+    // This dataframe must contain a column indicating the resource_id
```
grtlr, line 150: It would be really great to have that information as actual types. Conceptually, it would be great to have `struct Dataframe<I: Index>`, where `Index`, in this case, would be something like `ResourceId`. For data frames that don't have an explicit index (the order is the index), we could default to `struct Dataframe<I: Index = ()>`.

Of course `proto` does not give us generics, but maybe we could "monomorphize" the few instances where this is the case. We could then have specialized constructors + deserializers that ensure these invariants.

grtlr (commented):
Thank your for writing this out @jleibs!

I think the conceptual model (_catalog_, _entry_, _storage node_, ...) makes sense—we have refined this quite a bit over the last couple of weeks. We'll probably still find some weird edge cases that we haven't thought about yet.

I would love to make sure that we also spend some time to make the API descriptive and a joy to work, as we'll be spending most of our time with it in the upcoming cycles.

crates/store/re_protos/proto/rerun/v0/catalog.proto
```diff
@@ -0,0 +1,249 @@
...
+// - The Rerun Catalog Server
+// - Any Rerun Storage Node
+//
+// Some Important terminology:
+// - Resource: A resource is any data that can be mapped to a single recording or table in the Rerun data model.
+//   - Examples: A single Rerun recording, an MCAP file, parquet table, etc.
+//   - Every resource across the catalog has a unique identifier.
+// - Collection: A collection contains multiple resources of a similar type.
+//   - Every collection includes a "Property Table" of properties associated with the resources in a collection.
+// - Catalog Entry: An entry within a catalog is a named interface that facilitates access to resources or collections.
```
emilk, line 17: You define what a _collection_ is, but not what a _catalog_ is. Are those the same thing or different things?

jleibs: Tried to make this more clear.

crates/store/re_protos/proto/rerun/v0/redap.proto
```diff
@@ -0,0 +1,275 @@
...
+
+    // Get information about the chunks that make up the resource.
+    //
+    // This includes the ability to filter the chunks by time-range or entity-path.
+    //
+    // ChunkInfo contains metadata about the chunk but not the data itself.
+    rpc GetChunkInfos(GetChunkInfosRequest) returns (stream ChunkInfo) {}
+
+    // Retrieve specific chunks of data from the resource.
+    rpc GetChunks(stream GetChunkRequest) returns (stream rerun.common.v0.RerunChunk) {}
```
jprochazk, line 130:
```suggestion
    rpc GetChunks(GetChunkRequest) returns (stream rerun.common.v0.RerunChunk) {}
```

`stream` in request position doesn't work on web. `GetChunksRequest` would contain `chunk_ids` field with a list of chunk IDs to retrieve. This is how we make it work for web

crates/store/re_protos/proto/rerun/v0/redap.proto
```diff
@@ -0,0 +1,275 @@
...
+
+    // TODO: Chunk-filtering options
+    // Time-range
+    // Entity-paths
+}
+
+message GetChunkRequest {
+    rerun.common.v0.RecordingId recording_id = 1;
+    rerun.common.v0.Tuid chunk_id = 2;
+}
```
jprochazk, line 262:
```suggestion
message GetChunksRequest {
    rerun.common.v0.RecordingId recording_id = 1;
    repeated rerun.common.v0.Tuid chunk_ids = 2;
}
```

crates/store/re_protos/proto/rerun/v0/redap.proto
```diff
@@ -0,0 +1,275 @@
...
+    // top-K matches are found within *each* resource in the indexed entry. For inverted index
+    // it might be timepoints where the query string is found in the indexed column
+    // - 'data' column with the data that is returned for the matched timepoints
+    rpc SearchIndex(SearchIndexRequest) returns (stream rerun.common.v0.DataframePart) {}
+}
+
+// The Resource service is for accessing and querying the contents of individual resources.
+//
+// Each request includes a `resource_id` field, which is used to uniquely identify the resource within the catalog.
+service Resource {
```
jprochazk, line 118: How does an SDK+Viewer pair communicate the "default recording"? One option is to communicate that through a Viewer-control side channel. Another option is to not have a default recording at all, and you'd _always_ need to choose it after spawning the Viewer, which feels like it would quickly become annoying.

jprochazk (approved):

crates/store/re_protos/proto/rerun/v0/redap.proto
jprochazk: We also need a way to stream chunks to the resources to replace `WriteMessages`.

crates/store/re_protos/proto/rerun/v0/redap.proto
jprochazk: We also need a way to stream chunks to the resources to replace `WriteMessages`.

emilk: We should add some sort of lint/CI job to catch this common error

crates/store/re_protos/proto/rerun/v0/redap.proto
```diff
@@ -0,0 +1,275 @@
...
+//
+// In such a minimal deployment, the exposed Catalog might consist of a single Collection while the Add/RemoveEntry
+// APIs are unimplemented.
+
+// The Catalog service is for listing and management of multiple catalog entries.
+service Catalog {
+    // List all the available entries in the Catalog
+    //
+    // This must be implemented by all Catalog-compatible services.
+    rpc ListEntries(ListEntriesRequest) returns (ListEntriesResponse) {}
```
abey79, line 44: Maybe consider the `FindEntry` / `FindEntries` pair of end points as discussed during meet-up?

crates/store/re_protos/proto/rerun/v0/redap.proto
```diff
@@ -0,0 +1,275 @@
...
+    // Remove a resource from the a manually-managed catalog entry
+    //
+    // This API is only supported for some catalog entries.
+    rpc UnregisterResource(UnregisterResourcesRequest) returns (UnregisterResourceResponse) {}
+
+    // Remove all resources from a manually-managed catalog entry
+    //
+    // This API is only supported for some catalog entries.
+    rpc UnregisterAllResources(UnregisterAllResourcesRequest)
+        returns (UnregisterAllResourcesResponse) {}
```
abey79, line 97: According to the "pizza" discussion, these should all move to the top-level, "catalog" scope. Instead, there should be `AddResource(resource_id)` and `RemoveResrouce(resource_id)`, which would just add/remove a pre-existing resource to the collection.

crates/store/re_protos/proto/rerun/v0/redap.proto
abey79, line 118: How would one obtain a (possibly filtered) list of _all_ resources? (According to the pizza discussion, such a list exist at the catalog level, regardless of appartenance to a collection.)

I see 2 ways about this:
- `ListResources` (or rather `FindResource`/`FindResources`?)
- There always exists some `__everything__` meta-collection that always contains all resources, and can be queried with the regular Collection APIs.

abey79:
I believe this is outdated by some eons. Close?

Issue #8924, opened 2025-02-05
Allow disabling antialiasing (and alpha-to-coverage!) for screenshot comparision tests


Wumpf:
Anti-aliasing, especially alpha-to-coverage, varies greatly across devices/drivers/etc.
This is a large source of discrepancy in screenshot testing. We should disable this for almost all image comparison tests
(or rather make it opt-in, some should probably still test this!)

PR #8971, opened 2025-02-07
Disable anti-aliasing in tests

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/8924

### What

Motivation see ticket.
Adds the ability to re_renderer to disable MSAA upon context creation - we don't want this to be fully dynamic since it means recreating a lot of resources, but having this compile time set is bad as well (adds permutation, some tests may want to opt in).

Disables then MSAA for all tests by default, adjusting image threshholds accordingly.

* [x] find new image thresholds that work
* [x] pass main ci (screenshot testing on all platforms!)


Wumpf:
@rerun-bot full-check

crates/viewer/re_renderer/src/context.rs
```diff
@@ -26,7 +26,66 @@ pub enum RenderContextError {
         "The GPU/graphics driver is lacking some abilities: {0}. \
         Check the troubleshooting guide at https://rerun.io/docs/getting-started/troubleshooting and consider updating your graphics driver."
     )]
-    InsufficientDeviceCapabilities(#[from] crate::config::InsufficientDeviceCapabilities),
+    InsufficientDeviceCapabilities(#[from] crate::device_caps::InsufficientDeviceCapabilities),
+}
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum MsaaMode {
```
emilk, line 33:
```suggestion
/// Controls MSAA (Multi-Sampling Anti-Aliasing)
#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
pub enum MsaaMode {
```

crates/viewer/re_renderer/src/context.rs
```diff
@@ -26,7 +26,66 @@ pub enum RenderContextError {
...
+    /// especially in alpha-to-coverage cases.
+    ///
+    /// Note that this doesn't necessarily mean that we never use any multisampled targets,
+    /// merely that the main render target is not multisampled.
+    /// Some renderers/postprocessing effects may still incorporate textures with a sample count higher than 1.
+    Off,
+
+    /// 4x MSAA.
+    ///
+    /// As of writing 4 samples is the only option that works with `WebGPU`, but it guaranteed to be always available.
```
emilk, line 46:
```suggestion
    /// As of writing 4 samples is the only option (other than _Off_) that works with `WebGPU`,
    /// and it is guaranteed to be always available.
```

crates/viewer/re_renderer/src/view_builder.rs
```diff
@@ -359,41 +363,63 @@ impl ViewBuilder {
...
+            depth_or_array_layers: 1,
         };
-        let hdr_render_target_msaa = ctx
-            .gpu_resources
-            .textures
-            .alloc(&ctx.device, &main_target_desc);
-        // Like hdr_render_target, but with MSAA resolved.
-        let main_target_resolved = ctx.gpu_resources.textures.alloc(
+
+        // TODO(andreas): Should tonemapping preferences go here as well? Likely! g
```
emilk, line 374:
```suggestion
        // TODO(andreas): Should tonemapping preferences go here as well? Likely!
```

emilk (approved):
Great!

crates/viewer/re_renderer/src/view_builder.rs
Wumpf: g!

PR #8975, opened 2025-02-09
draft: correct unwanted eye (camera) movement upon firing shortcuts

kailiuca:
### What

This PR addressed a UI glitch that eye could move slightly upon firing up a shortcut, e.g., ctrl + s. As shown in the video below, across all the platforms, due to the view still listening for any keyboard_navigation, there is a small backwards move before (at 0:03 in the video below) and after (0:05) the save dialog.

https://github.com/user-attachments/assets/c4f4e95e-2430-41f9-8d4d-50ed8f0e2b2a

Besides, there is also a platform-dependent issue: when testing in Ubuntu 22.04.5 LTS, after closed the save dialog, the egui input context still registered the ctrl modifier and "S" keys_pressed, causing the eye moved backwards as long as cursor hovers the view (0:14 and 0:19 in the video below).

https://github.com/user-attachments/assets/0f6698aa-08ae-4584-8fc5-4f35358ac9eb

### DIFF
--- a/crates/viewer/re_view_spatial/src/eye.rs
+++ b/crates/viewer/re_view_spatial/src/eye.rs
@@ -445,6 +445,12 @@ impl ViewEye {
             return false; // e.g. we're typing in a TextField
         }
 
+        // Navigation should not be activated if any modifiers is pressed
+        let modifiers_pressed = egui_ctx.input(|input| input.modifiers.any());
+        if modifiers_pressed {
+            return false;
+        }
+
         let mut did_interact = false;
         let mut requires_repaint = false;
 
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -702,6 +702,12 @@ impl App {
                 Some(re_log_types::StoreId::random(StoreKind::Recording))
             });
 
+        // Clear any shortcut input before firing up any UI components
+        egui_ctx.input_mut(|input| {
+            input.modifiers = Default::default();
+            input.keys_down = Default::default();
+        });
+
         match cmd {
             UICommand::SaveRecording => {
                 if let Err(err) = save_recording(self, store_context, None) {


crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -445,6 +445,12 @@ impl ViewEye {
             return false; // e.g. we're typing in a TextField
         }
 
+        // Navigation should not be activated if any modifiers is pressed
+        let modifiers_pressed = egui_ctx.input(|input| input.modifiers.any());
+        if modifiers_pressed {
+            return false;
+        }
```
emilk, line 452: Currently we speed up and slow down the WSAD movement when the user presses shift/ctrl, and it seems like this code would break that?

More likely we want to only return early if any modifiers _other_ than shift/ctrl are pressed.

crates/viewer/re_viewer/src/app.rs
```diff
@@ -702,6 +702,12 @@ impl App {
                 Some(re_log_types::StoreId::random(StoreKind::Recording))
             });
 
+        // Clear any shortcut input before firing up any UI components
```
emilk, line 705: This comment would be more useful if it described _why_ instead of _what_

Wumpf:
@kailiuca are you planning to follow-up on this or do you want someone else to take over?

kailiuca:
> @kailiuca are you planning to follow-up on this or do you want someone else to take over?

Apologize @Wumpf , was super busy in the past few weeks. I've put up an alternative implementation.

crates/viewer/re_view_spatial/src/eye.rs
kailiuca: Thanks @emilk for the feedback. I wasn't aware of this behavior.

In this case, I'm proposing an alternative solution, please see my latest change: https://github.com/rerun-io/rerun/pull/8975/files#diff-daa0aae659935f597d1eb9ce8af241e875d7e61e03cf725151ec26f0c75efd30R462-R463

kailiuca: This approach mainly to stop the shortcut to propagate by clearing the input keys.

@emilk How do you feel about implementing an event propagation control feature for egui? Would this be something against the design of egui?

crates/viewer/re_viewer/src/app.rs
kailiuca: Thanks for the feedback, I'll review if that's still needed (feeling unlikely) if the alternative fix is implemented.

Wumpf:
Didn't notice this was ready for review again. `input.keys_down.remove(&kb_shortcut.logical_key);` seems to work fine. So let's just clean up the commented out code and land it 🤷

Wumpf (approved):

PR #8977, opened 2025-02-09
Save blueprint data as static (websocket server) #8741

### LINKED ISSUE #8741: blueprint `make_active` doesn't work with the web viewer
Blueprints sent via `send_blueprint` arrive just fine but they're not made active, even when 

Short repro:
```py
import rerun as rr
import rerun.blueprint as rrb
import numpy as np


blueprint = rrb.Blueprint(rrb.Spatial2DView(origin=f"image", name="test name"))
rr.init("image_viewer")
rr.serve_web(
    open_browser=True,
    web_port=9072,
    ws_port=9073,
)
rr.send_blueprint(blueprint)

image = np.random.randint(0, 256, (96, 96, 3))
rr.log("image", rr.Image(image))

input("Press Enter to exit...")
```

The viewer evidently received the blueprint and made it the default (`send_column` by makes the blueprint the default and makes it active) but it doesn't switch to this blueprint unless prompted to do so from the ui.
This does *not* happen when spawning the native viewer (`spawn=True` on init).

Curiously, when using `rr.init("image_viewer", default_blueprint=blueprint)`, the blueprint *is* set as default and visible!

See also https://discord.com/channels/1062300748202921994/1330599362774171768/1330599362774171768

DerpDays:
### Related
* Closes #8741

### What
Makes blueprint contents be saved internally as static (along with their activation command), instead of as a normal message.

This fixes the issue with default blueprints not being activated on web clients when a new user connects, since after this change blueprint data is also saved as static, this means the blueprint data is now sent to the client before the activation command (since the activation command is read and sent to the sink after the blueprint data).

More info can be found in the related issue [#8741](https://github.com/rerun-io/rerun/issues/8741#issuecomment-2646156711).

### DIFF
--- a/crates/store/re_ws_comms/src/server.rs
+++ b/crates/store/re_ws_comms/src/server.rs
@@ -19,7 +19,7 @@ use parking_lot::Mutex;
 use polling::{Event, Poller};
 use tungstenite::WebSocket;
 
-use re_log_types::LogMsg;
+use re_log_types::{LogMsg, StoreKind};
 use re_memory::MemoryLimit;
 use re_smart_channel::ReceiveSet;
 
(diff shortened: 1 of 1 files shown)


teh-cmc:
That's exactly the right fix, thanks!

Note for others reading: this does _not_ log the blueprint as static, which would be very scary, it's just more websocket server buffer shenanigans.

teh-cmc (approved):

PR #8981, opened 2025-02-10
Infer entity paths from LeRobot dataset feature metadata 

### LINKED ISSUE #8969: Infer entity names from LeRobot feature metadata
Currently, scalar entities don't use the *optional* feature names from dataset descriptions, e.g.:

```json
"observation.state": {
    "dtype": "float32",
    "shape": [
        2
    ],
    "names": {
        "motors": [
            "motor_0",
            "motor_1"
        ]
    }
}
```

we should use these to infer better names for the scalar entities we log.

oxkitsune:
### Related

Closes #8969 

### What

We now use the feature metadata from the LeRobot dataset to make the entity paths where possible.

image
(motor names for both observation and action values when visualizing [lerobot/pusht](https://huggingface.co/datasets/lerobot/pusht))

### DIFF
--- a/crates/store/re_data_loader/src/lerobot.rs
+++ b/crates/store/re_data_loader/src/lerobot.rs
@@ -352,6 +352,7 @@ impl LeRobotDatasetInfo {
 pub struct Feature {
     pub dtype: DType,
     pub shape: Vec<usize>,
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
lgtm!

PR #8983, opened 2025-02-10
Fix mouse pointers being 

teh-cmc:
The mouse pointers were out of frame after we reduced the image size following the AA fixes.

For that reason, the cam wouldn't move, which meant it was still in follow mode, which meant the results were wrong.

The rest of the diffs is just my vulkan upgrade over the week-end, it seems.


teh-cmc:
@rerun-bot full-check

Issue #8877, opened 2025-01-31
gRPC: download only specific time point or range


emilk:
When looking at a recording, I want to be able to send a link to a colleague pointing them at a specific time point or time range (a point of interest). This should generate a `rerun://` link on a format similar to this: `rerun://ip:port/store/ID?log_time=2025-01-31T02:55:20Z...2025-01-31T02:58:40Z`

* See https://github.com/rerun-io/rerun/issues/8876 for discussions around specific URL format

## gRPC protocol
I suggest we implement this by doing a latest-at or range-query on the server to find all chunk IDs that the viewer would need to show the given time range, and return their IDs. A separate `rpc` call would be used to then fetch the chunks (`rpc get_chunks(StoreId, stream ChunkId) -> stream  Chunk`).

There are two big advantages to this approach
* The server doesn't have to do expensive chunk slicing (saving CPU and memory)
* We can reuse the same `rpc` calls to implement larger-than-RAM (on-demand streaming)

The disadvantage is a potential waste of bandwidth (streaming parts of chunks that are outside the time range). This is mitigated by limiting the sizes of chunks during compaction.

## Viewer
When opening such a URI, we also set the timeline and time cursor according to the query parameter. If it is a single time point, just use it.
If it is a range, use the start of the range, and set the loop-range to the query range, and enable looping. Also zoom the time panel so that the time range fills the width of the screen.

![Image](https://github.com/user-attachments/assets/6ea9053a-dab5-4cff-9f46-a602db6ca06b)

## Related issues
* https://github.com/rerun-io/rerun/issues/8876

jprochazk:
Note that RPCs in the form of
```
rpc F(stream Request) (Response) {}
rpc F(stream Request) (stream Response) {}
```
are not supported by `grpc-web` at the moment, so are a non-starter. The blocker on their side is general availability of the `WebTransport` API... which will take a while.

emilk:
@jprochazk Can you elaborate please? What is not supported? Streaming parameters? Is there some issue or discussion you can link to?

I guess your comment is about my proposed `rpc get_chunks(StoreId, stream ChunkId) -> stream  Chunk`, but this could just as well be a `rpc get_chunk(StoreId, ChunkId) -> Chunk` (which I assume `grpc-web` supports)

grtlr:
This has been implemented both in the dataplatform and in the viewer:

* #9042

PR #8984, opened 2025-02-10
Adapt `re_grpc_client` to new URI scheme

### LINKED ISSUE #8876: Define and implement gRPC URI structure
We want to be able to hand out URIs pointing at speicific things inside a storage node.
We also want to distinguish a link to a storage nide from a `re_grpc_server` (see https://github.com/rerun-io/rerun/issues/8761).

This is not part of the gRPC specification, so we need to define this ourselves.

## URI
Here's a reminder of what the different parts of a URI are called (from [wikipedia](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier)):

Image

### scheme
I think we should use `rerun://` as the scheme, to indicate that is this is something Rerun-specific. This makes it obvious what the URL is about, and we can potentially tell the OS to open such links using the `rerun` binary.

### Host/port
If no port is specified, we should use the default Rerun port.

### Path
Here are some suggestions:

* `/stream`: for connecting to an event stream. This is currently the only mode supprorted by `re_grpc_server`: forwarding all events from a SDK sink
* `/catalog/ID`: display a specific catalog
* `/store/ID`: download a specific recording and/or blueprint (or should we maybe distinguish between the two in the URI?)

#### Recording queries
When downloading a recording, we should be a ble to set specific query, to only download part of the recording. Most importantly, we want to filter on _time_, to either only download the data for a specific time point or a time range.

Suggestion:

* `?log_time=$TIME`: specific time point
* `?log_time=$TIME...$TIME`: time range (inclusive)

Where `$TIME` is one of the following:
* Sequence number on the format `frame_nr=#51` 
* Date-time in ISO 8601 (e.g. `log_time=2025-01-31T02:58:40Z`)
* Time offset/duration in seconds (`sim_time=120.12s`)

The viewer should download only the chunks necessary to show the given timepoint or time range, and then also move the time cursor to match.

* See https://github.com/rerun-io/rerun/issues/8877

## Open questions
* How do we distinguish between HTTP and HTTPS? Do we need to?

## Related issues
* https://github.com/rerun-io/rerun/issues/8761

grtlr:
### Related

* Prepares #8876.
* Does not address the `time_log` aspect: #8877

### What

This implements `rerun://` (an alias for `rerun+https://`) and `rerun+http://`.

Please test with:
* `pixi run rerun rerun://redap.rerun.io/catalog`
* `pixi run rerun rerun+https://redap.rerun.io/catalog`
* and ideally with local Redap build `pixi run rerun rerun+http://localhost:51234/catalog`


crates/store/re_grpc_client/src/address.rs
```diff
@@ -1,45 +1,78 @@
...
+        match self {
+            Self::Rerun => write!(f, "rerun"),
+            Self::RerunHttp => write!(f, "rerun+http"),
+            Self::RerunHttps => write!(f, "rerun+https"),
+        }
+    }
+}
+
+impl Scheme {
+    fn convert(&self) -> &str {
```
jleibs, line 29: This needs a docstring and maybe a clearer name -- convert to what?

Seems more like `to_http_scheme()`

crates/store/re_grpc_client/src/address.rs
```diff
@@ -1,45 +1,78 @@
...
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct Origin {
+    scheme: Scheme,
+    host: url::Host<String>,
+    port: u16,
+}
+
+impl Origin {
+    pub fn convert(&self) -> String {
```
jleibs, line 45: Same as above -- maybe `to_url`?

jleibs: Wondering if it would even make sense to implement `From<Origin> for Url`?

I believe this would let you call:
```
let tonic_client = tonic::transport::Endpoint::new(origin.into())?
```

jleibs (approved):
Thanks! Looks good with a few nits on naming.

PR #8985, opened 2025-02-10
Add reindex capability to remote store protocol

drauschenbach:
### Related

Part of https://github.com/rerun-io/dataplatform/issues/156.

### What

We want to be able to manually request the reindexing of an existing index. This could be used after a recording is added, which leaves the index in a stale state.


zehiko (approved):
LGTM!

(I believe you can turn this from a draft into "ready for review" and merge, that is, if we won't cause too much conflicts for Jeremy and the wider renaming and split changes)

PR #8988, opened 2025-02-11
Document default-log level change

### LINKED ISSUE #8979: [docs] default_log_filter return warn if RUST_LOG not set in 0.22.0
**Describe the bug**
In 0.21.0, default_log_filter will return info if RUST_LOG environment variable is not set.
In 0.22.0, https://github.com/rerun-io/rerun/pull/8918 set the default log filter to warn
we should update docs(https://docs.rs/re_log/0.22.0/re_log/fn.default_log_filter.html) accordingly 
this was not included in the changelog of 0.22.0 as well.

**Rerun version**
0.22.0

oxkitsune:
### Related

Closes #8979

### What

This documents the change to the default log level that landed in `0.22` / #8918


teh-cmc:
Should probably add a very quick note to the migration guide itself while we're here

teh-cmc (approved):
Very nice

Issue #8994, opened 2025-02-11
Remove support for `MsgPack` serializer


jprochazk:
We should completely remove it and all the support code at some point. We have no backwards compatibility guarantee at this moment, but we want to avoid breaking people every single release. ~~So we'll wait until at least `0.24` to do this.~~ We have decided that we'd like to prioritize moving closer to `1.0`, so are making the breaking change for `0.23`.

Issue #8993, opened 2025-02-11
Migrate `rrd` to protobuf


jprochazk:
We have had a `protobuf` encoder/decoder implementation for a while now, but have not actually updated our usage of `rrd` to it. We should do that ASAP.

Issue #8631, opened 2025-01-09
Remove `LogMsg`


jprochazk:
Currently, all communication between SDK and Viewer is done through `LogMsg`:

https://github.com/rerun-io/rerun/blob/3a50bc23ccb9f90083546b69efcb51b063bd830f/crates/store/re_log_types/src/lib.rs#L263-L280

With `LogMsg`, we're mixing a lot of different concerns into one thing. Now that we are migrating our SDK/Viewer comms to gRPC, we can do better. For example:

- An endpoint to acquire the store info for a recording ID, replacing `SetStoreInfo`.
- An endpoint to stream in all chunks from a recording, replacing `ArrowMsg`, and removing the need to pair each chunk with a store ID, because it's implicitly known by the client when opening the stream.
- An endpoint to fetch the blueprint for a recording. This wouldn't have to be streamed at all, so we wouldn't need a `BlueprintActivationCommand` as a "stream end" marker.

On the `rrd` file side, we could:
- Migrate it to only hold one recording, and put `StoreInfo` in the file header, removing the need for `SetStoreInfo`.
- Use a separate file type for blueprints, removing the need for `BlueprintActivationCommand`. We already do this with `.rbl`, but we'd need to actually enforce it.
 
We probably still want multi-recording (including mixed with blueprint) files, but we could wrap these into a "rerun archive" file format, and provide the tools necessary to produce these files, as we currently do with `rerun rrd`.

PR #8995, opened 2025-02-11
Migrate file format to protobuf

jprochazk:
### Related

* Closes https://github.com/rerun-io/rerun/issues/8993
* Closes https://github.com/rerun-io/rerun/issues/8994

### What

* Use `protobuf` encoding for our file format.
* Completely remove `msgpack` code from the repository.

Tested by opening running the `clock` example and attempting to load the data in both the native and web viewers.

There are still some pieces left in the file format that have now been made redundant. One example is the per-recording `compression` field. We don't use that anymore, compression is now specified per `ArrowMsg`. There are a few other things like that, but I believe that refactoring is better left to the time when we decide to [get rid of `LogMsg`](https://github.com/rerun-io/rerun/issues/8631), which would be another breaking change for the file format.


crates/store/re_log_encoding/src/codec/file/mod.rs
```diff
@@ -30,6 +30,15 @@ pub(crate) struct MessageHeader {
 }
 
 impl MessageHeader {
+    /// Size of an encoded message header, in bytes.
+    // NOTE: This is `size_of` on a `repr(Rust)` struct,
+    //       which is fine because we control the layout
+    //       in the definition above, and tests would quickly
+    //       catch any sort of regression.
```
teh-cmc, line 37: Why are we not switching to the C ABI if we need control over layout though?

crates/store/re_log_encoding/src/codec/file/mod.rs
```diff
@@ -30,6 +30,15 @@ pub(crate) struct MessageHeader {
...
+    pub const SIZE_BYTES: usize = std::mem::size_of::<Self>();
+
+    // NOTE: We use little-endian encoding, because we live in
+    //       the 21st century.
```
teh-cmc, line 41: Reading this just after reviewing an arrow PR that encodes stuff as BE to be spec compliant lol

teh-cmc (approved):
Looking good.

We will need to fire a nightly ASAP once this is merged -- this will break all developers' welcome screen for a while. Let's try to do that a not too inconvenient time.

crates/store/re_log_encoding/src/codec/file/mod.rs
jprochazk: This part is our format, and it doesn't go over the network so nobody can claim it should be network-endian! :smile:

crates/store/re_log_encoding/src/codec/file/mod.rs
jprochazk: I thought about this a bit more, and maybe this should just be a literal constant. We need control over the _encoding_, not the layout of the struct

PR #8998, opened 2025-02-11
Simplify and test `wrap_in_list_array`

emilk:
### Related
* Follow-up to https://github.com/rerun-io/rerun/pull/8945

### DIFF
--- a/crates/utils/re_arrow_util/src/lib.rs
+++ b/crates/utils/re_arrow_util/src/lib.rs
@@ -1,5 +1,7 @@
 //! Helpers for working with arrow
 
+use std::sync::Arc;
+
 use arrow::{
     array::{Array, ArrayRef, ArrowPrimitiveType, BooleanArray, ListArray, PrimitiveArray},
     buffer::{NullBuffer, OffsetBuffer},
@@ -35,9 +37,7 @@ pub fn into_arrow_ref(array: impl Array + 'static) -> ArrayRef {
 }
 
 /// Returns an iterator with the lengths of the offsets.
-pub fn offsets_lengths(
-    offsets: &arrow::buffer::OffsetBuffer<i32>,
-) -> impl Iterator<Item = usize> + '_ {
+pub fn offsets_lengths(offsets: &OffsetBuffer<i32>) -> impl Iterator<Item = usize> + '_ {
     // TODO(emilk): remove when we update to Arrow 54 (which has an API for this)
     offsets.windows(2).map(|w| {
         let start = w[0];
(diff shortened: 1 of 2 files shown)


crates/utils/re_arrow_util/src/lib.rs
```diff
@@ -384,3 +384,87 @@ where
...
+/// Convert `[A, B, null, D, …]` into `[[A], [B], null, [D], …]`
+pub fn wrap_in_list_array(field: &Field, array: ArrayRef) -> (Field, ListArray) {
+    re_tracing::profile_function!();
+
+    debug_assert_eq!(field.data_type(), array.data_type());
+
+    let item_field = Arc::new(Field::new(
+        "item",
+        field.data_type().clone(),
+        field.is_nullable(), // TODO(emilk): it would be nice to remove the "inner nullability", and just have outer nullability.
```
emilk, line 399: I'm not sure how to strip the input `array` from its null buffer, and I don't think it matters much (there is no memory overhead)

crates/utils/re_arrow_util/src/lib.rs
```diff
@@ -384,3 +384,87 @@ where
...
+        );
+        assert_eq!(original_array.len(), 4);
+        assert_eq!(original_array.null_count(), 1);
+
+        let (new_field, new_array) =
+            wrap_in_list_array(&original_field, into_arrow_ref(original_array.clone()));
+
+        assert_eq!(new_field.data_type(), new_array.data_type());
+        assert_eq!(new_array.len(), original_array.len());
+        assert_eq!(new_array.null_count(), original_array.null_count());
```
jleibs, line 443: Would be nice to have something like:
```
assert_eq!(original_field.data_type(), new_array.values().data_type());
```

crates/utils/re_arrow_util/src/lib.rs
```diff
@@ -384,3 +384,87 @@ where
...
+        array::{Array as _, AsArray as _, Int32Array},
+        buffer::{NullBuffer, ScalarBuffer},
+        datatypes::{DataType, Int32Type},
+    };
+
+    use super::*;
+
+    #[test]
+    fn test_wrap_in_list_array() {
+        // Convert [42, 1337, null, 69] into [[42], [1337], null, [69]]
```
jleibs, line 429:
```suggestion
        // Convert [42, 69, null, 1337] into [[42], [69], null, [1337]]
```
Inconsistent with the data below.

jleibs (approved):

crates/utils/re_arrow_util/src/lib.rs
emilk: That makes sense - I'll update the comment!

Issue #9002, opened 2025-02-12
`rerun --web-viewer file.rrd` no longer loads rrd


jleibs:
To repro:
```
cargo run -p clock -- --save clock.rrd
pixi run rerun-web clock.rrd
```

Expectation:
- Should load the rrd into the web-viewer

Actual:
- Just get the welcome screen

Confirmed this is a regression from: https://github.com/rerun-io/rerun/pull/8838

PR #9004, opened 2025-02-12
Fix: `--serve-web` should consume data sources

jprochazk:
### Related

* Closes https://github.com/rerun-io/rerun/issues/9002

### What

Previously we would create a `RerunServer` from `rxs` before serving the Web Viewer. The gRPC server that's started by `--serve-web` will now consume `rxs` as well. It can still be sent additional data via `connect_grpc` in our SDKs, even after it consumes its initial data sources.

Tested with (from linked issue):
```
cargo run -p clock -- --save clock.rrd
pixi run --rerun-web clock.rrd
```

### DIFF
--- a/crates/store/re_grpc_server/src/lib.rs
+++ b/crates/store/re_grpc_server/src/lib.rs
@@ -27,6 +27,8 @@ use tower_http::cors::CorsLayer;
 pub const DEFAULT_SERVER_PORT: u16 = 9876;
 pub const DEFAULT_MEMORY_LIMIT: MemoryLimit = MemoryLimit::UNLIMITED;
 
+// TODO(jan): Refactor `serve`/`spawn` variants into a builder?
+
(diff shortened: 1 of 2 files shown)


crates/store/re_grpc_server/src/lib.rs
```diff
@@ -27,6 +27,8 @@ use tower_http::cors::CorsLayer;
 pub const DEFAULT_SERVER_PORT: u16 = 9876;
 pub const DEFAULT_MEMORY_LIMIT: MemoryLimit = MemoryLimit::UNLIMITED;
 
+// TODO(jan): Refactor `serve`/`spawn` variants into a builder?
```
jprochazk, line 30: I keep adding more of these, it's unlikely to increase any further but if it does I'll refactor it into more of a builder.

teh-cmc (approved):

Issue #6819, opened 2024-07-08
Proposal: remove support for nullable components


emilk:
At the moment we don’t ever make use of nullable components.
For instance: we don’t allow setting the radius of only _some_ points in a point-cloud to `null`.

Still, we have full support for it in our `Loggable` trait and in our codegen.
We have a `to_arrow_opt` taking an iterator of optional values, and a `from_arrow_opt` that returns a `Vec<Option<Self>>`, yet these are never called.

This unused support has several downsides:
* Extra complexity in the codegen
* A lot of extra generated code (we generate both `from_arrow_opt` and `from_arrow`)
* Runtime overhead (we codegen `to_arrow_opt`, but only ever call `to_arrow` which wraps each element in `Some(…)`)

Removing support for it would definitely make the transition away from `arrow2` easier. And we can always add it back in the future.

It seems silly to support a feature we _may_ want in the future (YAGNI / ease into complexity).

jleibs:
Just to clarify a bit further, this would make it so that the **inner** values of arrow batches are never null. However the **outer** list-arrays itself would still have nullable arrays if an optional value wasn't provided in a row.

Maybe this should be articulated as: "Remove support for individually nullable items within a batch".

Logically I'm totally onboard, though there may be lurking consequences when it comes to how some range-operations are processed. This is probably fresher in @teh-cmc 's mind.

It could be worth thinking some more about both this and https://github.com/rerun-io/rerun/issues/6602 together in terms of simplifying the things we support.

Also very peripherally related: I remember nullabillity (along with unions) were one of the annoying things that we ran up against when we were previously using polars. A simplification that made it easier to guarantee Rerun data frame results can be loaded into polars without issues would also be a huge win.

emilk:
> Just to clarify a bit further, this would make it so that the inner values of arrow batches are never null. However the outer list-arrays itself would still have nullable arrays if an optional value wasn't provided in a row.

Good clarification. Just to be extra clear:

* We currently support: `Option<Vec<Option<C>>>`
* The proposal is to only support `Option<Vec<C>>`

teh-cmc:
There are two distinct issues here: one is the presence of field-level nullability, another is dealing with that nullability efficiently.

---

We _do_ make use of instance-level nullability, just not at the top layer (because of limitations in the flatbuffers IDL which have thankfully prevented us to do so).

If you look at any datatype that has optional fields, like e.g. the `Material` datatype:
```rust
[...]

PR #9009, opened 2025-02-12
Remove check for inner nullability of component columns

emilk:
### Related
* https://github.com/rerun-io/rerun/pull/8998
* #6819

### What
`pixi run rerun rerun://redap.rerun.io/catalog` failed with `Detected malformed Chunk: The outer array in chunked component batch must be a sparse list, got List(Field { name: "item", data_type: Utf8, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} })`.

The cause was a check in the `Chunk::sanity_check` requiring _inner mutability_, i.e. allowing a single component to be null in a list. The intention of the check was to check for _outer_ nullability, i.e. that any cell in the column can be `null` (or a dense list).

In other words, we want to support `Vec<Option<Vec<T>>`, but NOT require `Vec<Option<Vec<Option<T>>>`.

Why did this trigger now? Because we now allow sending component columns as "mono components", i.e. as `Vec<T>` and that is then automatically changed to `Vec<Vec<T>>`, but wether or not it has interior nullability depends on if the source data had it.

### DIFF
--- a/crates/store/re_chunk/src/chunk.rs
+++ b/crates/store/re_chunk/src/chunk.rs
@@ -1521,24 +1521,19 @@ impl Chunk {
             component_name.sanity_check();
             for (component_desc, list_array) in per_desc {
                 component_desc.component_name.sanity_check();
-                if !matches!(list_array.data_type(), arrow::datatypes::DataType::List(_)) {
+                // Ensure that each cell is a list (we don't support mono-components yet).
+                if let arrow::datatypes::DataType::List(_field) = list_array.data_type() {
+                    // We don't check `field.is_nullable()` here because we support both.
+                    // TODO(#6819): Remove support for inner nullability.
+                } else {
                     return Err(ChunkError::Malformed {
                         reason: format!(
-                            "The outer array in a chunked component batch must be a sparse list, got {:?}",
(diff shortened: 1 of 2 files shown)


emilk:
`cargo nextest run --all-features --all-targets` passes

crates/store/re_chunk/src/chunk.rs
```diff
@@ -1521,24 +1521,19 @@ impl Chunk {
...
-                if !matches!(list_array.data_type(), arrow::datatypes::DataType::List(_)) {
+                // Ensure that each cell is a list (we don't support mono-components yet).
+                if let arrow::datatypes::DataType::List(_field) = list_array.data_type() {
+                    // We don't check `field.is_nullable()` here because we support both.
+                    // TODO(#6819): Remove support for inner nullability.
+                } else {
                     return Err(ChunkError::Malformed {
                         reason: format!(
-                            "The outer array in a chunked component batch must be a sparse list, got {:?}",
+                            "The inner array in a chunked component batch must be a list, got {:?}",
```
teh-cmc, line 1531: This is not the inner array, this is the outer list_array, i.e. the column

teh-cmc: But I already merged, so

PR #9010, opened 2025-02-12
gRPC spec update: Add support for fetching temporal range of recordings

zehiko:
### What

Adding support for fetching chunks that belong to specific range of the recording. This consists of the 2 APIs calls:
* fetching chunk ids for a given range
* fetching chunks for the given chunk ids (which we might want to leverage separately in the future)

As there is a limitation for the client side gRPC streaming on the web (see [here](https://github.com/rerun-io/rerun/issues/8877#issuecomment-2652986044)), we've decided to send a batch of chunk ids to the server. To make the APIs somewhat symmetrical, the 1st API call will stream batches of chunk ids. 

Closes
* https://github.com/rerun-io/rerun/issues/8877

### DIFF
--- a/crates/store/re_protos/proto/rerun/v0/remote_store.proto
+++ b/crates/store/re_protos/proto/rerun/v0/remote_store.proto
@@ -11,6 +11,9 @@ service StorageNode {
 
     rpc CreateIndex(CreateIndexRequest) returns (CreateIndexResponse) {}
 
+    rpc GetChunkIds(GetChunkIdsRequest) returns (stream GetChunkIdsResponse) {}
+    rpc GetChunks(GetChunksRequest) returns (stream rerun.common.v0.RerunChunk) {}
+
     // The response to `SearchIndex` a RecordBatch with 3 columns:
     // - 'resource_id' column with the id of the resource
     // - timepoint column with the values representing the points in time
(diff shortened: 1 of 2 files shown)


crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -12,6 +12,9 @@ service StorageNode {
     rpc CreateIndex(CreateIndexRequest) returns (CreateIndexResponse) {}
     rpc ReIndex(ReIndexRequest) returns (ReIndexResponse) {}
 
+    rpc GetChunkIds(GetChunkIdsRequest) returns (stream GetChunkIdsResponse) {}
+    rpc GetChunks(GetChunksRequest) returns (stream rerun.common.v0.RerunChunk) {}
```
drauschenbach, line 16: This is typically [not advised](https://kreya.app/blog/grpc-best-practices/#separate-request-and-response-messages), and I've personally made the mistake of re-using return types before I learned of this advice.

zehiko: I think there's a topic here, this already came up. Right now we re-use ``DataframePart`` and ``RerunChunk`` across more than a few calls, which was intentional and done as part of https://github.com/rerun-io/dataplatform/issues/74. 
For now I'd keep it, but let's discuss this re-use and get rid or keep all of them, if that makes sense.

emilk: * https://github.com/rerun-io/rerun/issues/8563

drauschenbach: Adding pagination support for `GetChunks` might be a case in point. A typical `GetChunksResponse` message would include a `RerunChunk`, and also possibly a breadcrumb to the next page, or even stats.

Temporary consistency for now sounds good though.

Issue #9003, opened 2025-02-12
Project: first-class integration and UI for Redap catalogs and collections in the viewer (first iteration)


abey79:
## Summary description

So far, when querying a Redap collection, the viewer would transform the resulting table on-the-fly into a regular recording, along with a blueprint displaying it as a dataframe view. This was the quickest way to display something, but comes with restrictions:
- The dataframe view is intended as a generic tool, and is not the right place to have catalog/collection-specific customisation.
- Forking the dataframe view into a dedicated catalog/collection view could help, but would still suffer from the constraints of the existing recording/view system abstractions.
- The current data ingestion pipeline (data sources -> channel of `LogMsg` -> `StoreHub` is not suited for the kind of communication required to continuously sync a catalog/collection display with the remote server.

The goal of this project is to implement an initial version of a dedicated communication pipeline and UI to display catalog and collection, including:
- dedicated ingestion pipeline for catalog data
- dedicated list UI in the left panel for servers/catalog/collection
- dedicated table UI "over"/instead of the viewport when a collection is selected

For each of these, the scope is a MVP approach to reach feature parity with the status quo. This project is complete when the initial implementation is merged on `main`, enabling iterative improvements in smaller PRs.

## Tasks

- [x] Make the async runtime available to the viewer.
- [x] Establish a parallel ingestion path for catalog/collections
- Initial version of `re_catalog_hub`, a crate that implements a UI for redap catalog/collections, as well as the communication with redap server to source the required data.
  - [ ] UI: Left panel "Server" section with a list of server and their catalog/collections.
  - [ ] UI: Central panel collection view, initially identical to what the dataframe currently displays for catalogs.
  - [ ] Comm: ability to fetch collection data (stretch goal: "Refresh" button)

## Design

The new server/catalog/collection is to be accessed using a new "Local"/"Server" tab bar.

<img width="1335" alt="Image" src="https://github.com/user-attachments/assets/270f3282-9b2d-4def-9d24-4c98a53ce23a" />

## Non-goals and won't do

There is no end to this chapter of Rerun's TODO list. This project ends when we can start to tick items off.

PR #9018, opened 2025-02-12
First iteration of a dedicated UI and communication for the Redap server/catalog/collection browser

abey79:
### Related

* Closes: #9003

### What

Implement a whole new Redap server/catalog/collection UI.


crates/store/re_grpc_client/src/redap/address.rs
```diff
@@ -35,12 +35,15 @@ pub enum ConnectionError {
     /// The given url is not a valid Rerun storage node URL.
     #[error("URL {url:?} should follow rerun://host:port/recording/12345 for recording or rerun://host:port/catalog for catalog")]
     InvalidAddress { url: String, msg: String },
+
+    #[error("Catalog URL {origin:?} cannot be loaded as a recording")]
+    CannotLoadCatalogAsRecording { origin: Origin },
```
grtlr, line 40: I think instead of the `Origin` we should pass in the entire recording URL (`String`).

Btw, we you're at it could you please remove `InvalidAddress`? We are not using that anymore. Is there a way to find unused variants in libraries?

crates/viewer/re_chunk_store_ui/src/chunk_store_ui.rs
```diff
@@ -67,13 +67,9 @@ impl Default for DatastoreUi {
...
     /// Show the ui.
-    pub fn ui(
-        &mut self,
-        ctx: &ViewerContext<'_>,
-        ui: &mut egui::Ui,
-        datastore_ui_active: &mut bool,
-        time_zone: TimeZone,
-    ) {
+    pub fn ui(&mut self, ctx: &ViewerContext<'_>, ui: &mut egui::Ui, time_zone: TimeZone) -> bool {
+        let mut datastore_ui_active = true;
```
grtlr, line 71: That code looks a bit suspicious—could we please add a comment to explain what is going on?

crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -366,14 +382,17 @@ impl AppState {
 
         if *show_settings_ui {
             // nothing: this is already handled above
-        } else if *show_datastore_ui {
-            datastore_ui.ui(&ctx, ui, show_datastore_ui, app_options.time_zone);
+        } else if *display_mode == DisplayMode::ChunkStoreBrowser {
+            let datastore_view_active = datastore_ui.ui(&ctx, ui, app_options.time_zone);
+            if !datastore_view_active {
+                *display_mode = DisplayMode::Viewer;
+            }
```
grtlr, line 389: This took me a while to figure out (see comment above)—mostly because of the `&mut bool` in `chunk_store_ui`. Wonder if we should propagate the change of just returning a `bool` to that method as well.

grtlr (approved):
LGTM—only minor comments.

crates/viewer/re_viewer/src/app_state.rs
abey79: I've named/commented stuff to be more explicit.

abey79: This chunk store ui could benefit from a refactor based on a command queue. But that's for another day.

PR #9027, opened 2025-02-13
Load lerobot dataset on separate IO thread

oxkitsune:
### What

This runs the LeRobot dataset loader on a separate IO thread to avoid blocking the ui.


Wumpf (approved):
lgtm!

PR #9028, opened 2025-02-13
Log task descriptions for each LeRobot dataset episode

oxkitsune:
### Related

This is based on #9027, which should be merged first

### What

This adds support for the `DType::String` and logs the task description for an episode to the viewer.


crates/store/re_data_loader/src/load_file.rs
```diff
@@ -23,6 +23,8 @@ pub fn load_from_path(
     // NOTE: This channel must be unbounded since we serialize all operations when running on wasm.
     tx: &Sender<LogMsg>,
 ) -> Result<(), DataLoaderError> {
+    use crate::DataLoaderSettings;
+
```
teh-cmc, line 27: Inline this, this looks weird just standing there on its own (and weird looking makes the reader thinks that weird things are happening).

crates/store/re_data_loader/src/load_file.rs
```diff
@@ -35,6 +37,17 @@ pub fn load_from_path(
 
     re_log::info!("Loading {path:?}…");
 
+    // When loading a LeRobot dataset, avoid sending a `SetStoreInfo` message since the LeRobot
+    // loader handles this automatically.
+    let settings = if crate::lerobot::is_lerobot_dataset(path) {
+        &DataLoaderSettings {
```
teh-cmc, line 43:
```suggestion
        &crate::DataLoaderSettings {

teh-cmc (approved):

crates/store/re_data_loader/src/load_file.rs
oxkitsune: My LSP does weird things sometimes 🙃

Issue #9019, opened 2025-02-12
Make line series visualizer handle vectors of scalars per time index


Wumpf:
* Part of #9005

Things to take into account:
* multidimensional scalars (duh!)
* various properties of the `SeriesLine` - not all equally important
   * names
   * colors
   * stroke width
   * etc. 
* basic ui interaction
* no perf regression (may take opportunity to refresh understanding of existing perf bottlenecks)

out of scope for first iteration:
* complex ui interaction
* visibility handling

PR #9033, opened 2025-02-13
Make `SeriesLine` visualizer work with several scalars per time

Wumpf:
### Related

* Fixes [#9019](https://github.com/rerun-io/rerun/issues/9019)

### What

image

Try the example in the browser: 
https://rerun.io/viewer/pr/9033?url=https://static.rerun.io/rrd/0.22.0/multiplot_0a77742f7ddd55361c4b7bd9976caa1cc45f9c14.rrd

`SeriesLine` visualizer now operates _as-if_...
* scalars was a primary component (not new really actually) which is an array of arbitrary length on each timestamp
* colors, name and stroke-width are components that are splatted upon each scalar

**In other words: Single entity, multiple plots**

To test this out I added a manual "test" for now (to be removed again) which tests both static & varying color & widths as well as gaps.

In a follow-up I'll add automated tests for this.
This PR intentionally does NOT touch any APIs. Logging happens right now via the same generic APIs we use for column batches. That API didn't really make sense before here but works pretty much how one would expect, begging the question if this isn't already good enough 🤔 

Performance before/after was verified to be unchanged using
`cargo run --release --bin plot_dashboard_stress -- --num-plots 16 --num-series-per-plot 32`
(curiously it actually improved a lil bit from 30.1ms 29.6ms overall viewer time but that's probably just measurement fluctuation / background task activity)
after

### DIFF
--- a/crates/viewer/re_view_time_series/src/line_visualizer_system.rs
+++ b/crates/viewer/re_view_time_series/src/line_visualizer_system.rs
@@ -228,14 +228,12 @@ impl SeriesLineSystem {
             // Keep in mind clears here.
             let num_series = all_scalar_chunks
                 .iter()
-                .next()
-                .and_then(|chunk| chunk.iter_slices::<f64>(Scalar::name()).next())
-                .map(|slice| slice.len())
+                .find_map(|chunk| {
+                    chunk
+                        .iter_slices::<f64>(Scalar::name())
+                        .find_map(|slice| (!slice.is_empty()).then_some(slice.len()))
+                })
                 .unwrap_or(1);
-            if num_series == 0 {
-                re_log::warn_once!("Empty scalar array found for {entity_path:?}");
-                return;
-            }
 
             // Allocate all points.
             {
--- a/crates/viewer/re_view_time_series/src/line_visualizer_system.rs
+++ b/crates/viewer/re_view_time_series/src/line_visualizer_system.rs
@@ -10,11 +10,11 @@ use re_types::{
     components::{Color, Name, Scalar, StrokeWidth},
     Archetype as _, Component, Loggable,
 };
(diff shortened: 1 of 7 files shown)


crates/viewer/re_view_time_series/src/line_visualizer_system.rs
```diff
@@ -233,11 +222,37 @@ impl SeriesLineSystem {
...
+            // Keep in mind clears here.
+            let num_series = all_scalar_chunks
+                .iter()
+                .next()
+                .and_then(|chunk| chunk.iter_slices::<f64>(Scalar::name()).next())
+                .map(|slice| slice.len())
+                .unwrap_or(1);
+            if num_series == 0 {
+                re_log::warn_once!("Empty scalar array found for {entity_path:?}");
+                return;
```
Wumpf, line 237: this is actually a bit fishy. What if the series starts with an empty array for whatever reason. Should at least search until we find one that isn't empty.

teh-cmc:
> To test this out I added a manual "test" for now (to be removed again) which tests both static & varying color & widths as well as gaps.

I actually would like to make this part of the Plots example. This is valuable both for our users and us, as something easily accessible from the welcome screen.

docs/snippets/all/archetypes/scalar_simple.rs
teh-cmc, line 1: Revert? (Is the CI passing because this roundtrip is disabled because of floats...)

tests/rust/experimental_multi_scalar/src/main.rs
teh-cmc, line 1: neat

teh-cmc (approved):
Haven't digged into every single detail, but the general approach of having two dedicated code paths makes a lot of sense to me. Let's get this soaked.

Issue #8992, opened 2025-02-11
Encode RowId:s in a sane way in arrow


emilk:
Due to weird historical reasons we encode `RowId`:s as a `struct` of two `UInt64`. This is horrible.

Let's switch to whatever is standard for a UUID. Probably a fixed-size 16 byte.

We should be able to do so in a nice backwards-compatible way with the new `re_sorbet` interface

Should probably wait for
* https://github.com/rerun-io/rerun/issues/8744

jleibs:
https://arrow.apache.org/docs/format/CanonicalExtensions.html#uuid

PR #9038, opened 2025-02-13
Enocde RowId as FixedSizeBinaryArray

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/8992

### What
Change how we encode `RowId` in arrow from a weird `StructArray` to a `FixedSizeBinaryArray` ([same as how UUID:s are usually encoded](https://arrow.apache.org/docs/format/CanonicalExtensions.html#uuid)).

This is a breaking change (a lot of those going around anyways).


emilk:
@rerun-bot full-check

crates/store/re_types_core/src/id.rs
```diff
@@ -111,8 +115,6 @@ impl std::ops::DerefMut for ChunkId {
     }
 }
 
-crate::delegate_arrow_tuid!(ChunkId as "rerun.controls.ChunkId");
```
teh-cmc, line 114: We still need this in some form or another -- ChunkIds as a column are used all over the place in the DP's  indexes

crates/store/re_types_core/src/tuid.rs
```diff
@@ -40,92 +41,37 @@ impl Loggable for Tuid {
...
-        )))
+        let iter = iter.into_iter();
+
+        let mut builder = FixedSizeBinaryBuilder::with_capacity(iter.size_hint().0, BYTE_WIDTH);
+        for tuid in iter {
+            #[allow(clippy::unwrap_used)] // Can't fail because `BYTE_WIDTH` is correct.
+            builder.append_value(tuid.into().as_bytes()).unwrap();
+        }
+
+        Ok(Arc::new(builder.finish()))
```
teh-cmc, line 57: That would probably make it easier on both arrow-rs and rustc:
```suggestion
        FixedSizeBinary::try_from_iter(iter.map(|tuid| tuid.into().as_bytes()).map(Arc::new)

File: crates/utils/re_tuid/src/lib.rs [hunk 3]
```diff
@@ -11,15 +11,18 @@
...
 #[derive(Clone, Copy, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "bytemuck", derive(bytemuck::AnyBitPattern))]
+#[cfg_attr(
+    feature = "bytemuck",
+    derive(bytemuck::AnyBitPattern, bytemuck::NoUninit)
+)]
 pub struct Tuid {
     /// Approximate nanoseconds since epoch.
-    /// A LE u64 encoded as bytes to keep the alignment of `Tuid` to 1.
+    /// A BE u64 encoded as bytes to keep the alignment of `Tuid` to 1.
```
teh-cmc, line 20: A comment explaining the (unsual) choice of big-endianess (+ link to the arrow spec) would be nice

teh-cmc (approved):

crates/store/re_types_core/src/tuid.rs
emilk: Except `FixedSizeBinary::try_from_iter` fails on zero-length input, so would need an `if`-statement

Issue #9029, opened 2025-02-13
Split global context off from `ViewerContext` and make `DataUi` (and friends) "context-agnostic"


abey79:
### Background

The current `ViewerContext` is way too specific in terms of current recording/blueprint and visualisation thereof to be meaningfully used in for the upcoming "Redap browser". However, it does contain stuff of general usefulness that the Redap browser requires, such as the registry of component UI.

### Proposal

Let's split it!

![Image](https://github.com/user-attachments/assets/a9f70a9b-9ac9-4f81-84b6-8eab23d18960)

We currently already have a hierarchy of increasingly specific context structure:

```
ViewerContext -> ViewContext -> QueryContext
```

We propose to extend this hierarchy:

```
GlobalContext -> ViewerContext -> ViewContext -> QueryContext
        |
        v              
RedapBrowserContext
```

Some subsystems, such as `re_data_ui` and `re_component_ui`, should be available globally (because we display stuff everywhere), but benefit from more specific context when available. For example a `ClassId` component might be able to display the corresponding color if, and only if, a datastore and a latest-at query are known, enabling an `AnnotationContext` to be queried. The `ClassId` should _also_ be correctly displayed (albeit without color) if there is no such knowledge.

To address that, we further propose to update these api to accept an enum of any of these contexts:

```rust
pub enum AnyContext<'a> {
    Global(GlobalContext<'a>),
    RedapBrowser(RedapBrowserContext<'a>),
    Viewer(ViewerContext<'a>),
    View(crate::ViewContext<'a>),
    Query(crate::QueryContext<'a>),
}
```

### Some code to get started

```rust
//TODO: Step 2
/// This is what we pass to `DataUi` and friends, which should do their best effort at displaying
/// stuff to the extent permitted by the actually provided context.  
pub enum AnyContext<'a> {
    Global(GlobalContext<'a>),
    RedapBrowser(RedapBrowserContext<'a>),
    Viewer(ViewerContext<'a>),
    View(crate::ViewContext<'a>),
    Query(crate::QueryContext<'a>),
}

//TODO: Step 1
/// This is application-wide context that applies to the whole Rerun viewer app (namely both the
/// local viewer and redap browser).
pub struct GlobalContext<'a> {
    /// Global options for the whole viewer.
    pub app_options: &'a AppOptions,

    /// Runtime info about components and archetypes.
    ///
    /// The component placeholder values for components are to be used when [`crate::ComponentFallbackProvider::try_provide_fallback`]
[...]

PR #9039, opened 2025-02-14
Factor out `GlobalContext` from `ViewerContext`

grtlr:
### Related

* #9029


crates/viewer/re_viewer_context/src/global_context/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+
+pub use app_options::AppOptions;
+
+use crate::ViewClassRegistry;
+pub use command_sender::{
+    command_channel, CommandReceiver, CommandSender, SystemCommand, SystemCommandSender,
+};
+pub use component_ui_registry::{ComponentUiRegistry, ComponentUiTypes};
+pub(crate) use item::resolve_mono_instance_path_item;
+pub use item::Item;
```
Wumpf, line 20: nit: would be to have separate block for local use and exports (`pub use`) so it's easy to see at a glance what we're exporting from this module

crates/viewer/re_viewer_context/src/viewer_context.rs
```diff
@@ -7,17 +7,19 @@ use re_entity_db::entity_db::EntityDb;
...
 use crate::{
     query_context::DataQueryResult, AppOptions, ApplicationSelectionState, CommandSender,
     ComponentUiRegistry, DragAndDropManager, IndicatedEntities, ItemCollection,
     MaybeVisualizableEntities, PerVisualizer, StoreContext, SystemCommandSender as _, TimeControl,
     ViewClassRegistry, ViewId,
 };
 
-/// Common things needed by many parts of the viewer.
-pub struct ViewerContext<'a> {
+impl ViewerContext<'_> {
```
Wumpf, line 18: nit: I appreciate that this keeps the diff smaller, but it's a bit strange to first have the impl and then the struct

Wumpf (approved):
lgtm()

PR #9040, opened 2025-02-14
Fix: Configure max message size for gRPC server

jprochazk:


emilk (approved):
LGTM if tested that it now works to log >4MiB videos

oxkitsune (approved):
both checklists work now, awesome

jprochazk:
3 line PR, not going to waste CI here

PR #9041, opened 2025-02-14
Encode wether or not a chunk is sorted by `RowId` in the column metadata

emilk:

It used to be encoded in the `RecordBatch` metadata. It makes more sense to encode it on the column metadata though, and brings the `RowId` column closer to the time columns (I hope to unify the two in the near future).


crates/store/re_chunk/src/transport.rs
```diff
@@ -224,10 +218,11 @@ impl Chunk {
...
         };
 
+        let is_sorted_by_row_id = batch.chunk_schema().row_id_column().is_sorted;
         let mut res = Self::new(
-            id,
-            entity_path,
-            is_sorted.then_some(true),
+            batch.chunk_id(),
+            batch.entity_path().clone(),
+            is_sorted_by_row_id.then_some(true),
```
teh-cmc, line 225: The doc below says 
```
`false` means either "unsorted" or "unknown".
```
but this seems to say that the mere presence of a value means `true`? which is it?

crates/store/re_sorbet/src/row_id_column_descriptor.rs
```diff
@@ -69,24 +75,28 @@ impl TryFrom<&ArrowField> for RowIdColumnDescriptor {
     type Error = WrongDatatypeError;
 
     fn try_from(field: &ArrowField) -> Result<Self, Self::Error> {
-        Self::try_from(field.data_type())
+        // Self::try_from(field.data_type())
```
teh-cmc, line 78: dead code

crates/store/re_sorbet/src/row_id_column_descriptor.rs
```diff
@@ -69,24 +75,28 @@ impl TryFrom<&ArrowField> for RowIdColumnDescriptor {
...
-    fn try_from(data_type: ArrowDatatype) -> Result<Self, Self::Error> {
-        WrongDatatypeError::compare_expected_actual(&RowId::arrow_datatype(), &data_type)?;
-        Ok(Self {})
-    }
-}
+//     fn try_from(data_type: ArrowDatatype) -> Result<Self, Self::Error> {
+//         WrongDatatypeError::compare_expected_actual(&RowId::arrow_datatype(), &data_type)?;
+//         Ok(Self {})
+//     }
+// }
```
teh-cmc, line 102: What's this for?

teh-cmc (approved):

crates/store/re_chunk/src/transport.rs
emilk: Right now we only encode the knowledge that is it sorted if it is sorted… maybe we should change that though.

PR #9042, opened 2025-02-14
Fetch time range

jprochazk:
Depends on:
* ~~https://github.com/rerun-io/rerun/pull/9044~~

Syntax:
```
rerun+http://addr:port/recording/<recording-id>?time_range=<timeline>@<min>s?..<max>s?
```

The `s` suffix means "seconds", and if present, we assume the given `timeline` is temporal. If it is absent, we assume it is a sequence timeline.

Examples:

```sh
# Temporal timeline
rerun+http://127.0.0.1:51234/recording/c2c6bef6-5bdb-43a8-8563-ce878b238936?time_range=stable_time@1.1198s..3.4998s

# Sequence timeline
rerun+http://127.0.0.1:51234/recording/c2c6bef6-5bdb-43a8-8563-ce878b238936?time_range=log_tick@200..500
```

Also added a new UI command, `CopyTimeRangeLink` (only available from command palette for now), which constructs a redap URL from the active recording ID + loop selection. You need to have a redap recording loaded for that to work.

To test:
1. Build `rgbd` (or any other large-ish interesting recording)
   ```
   pixi run -e wheel-test py-build-common
   pixi run -e wheel-test build-examples rrd --channel release --examples rgbd example_data
   ```
1. Start a catalog server on localhost
1. Register the `rgbd` recording
1. `pixi run rerun-web`
1. Once browser tab opens, change `?url=` to (percent-encoded, e.g. using browser dev tools `encodeURIComponent`) `rerun+http://127.0.0.1:51234/recording/<rgbd-recording-id>`
1. Create a loop selection (shift + left mouse button drag on the timeline)
1. <kbd>CMD + P</kbd> -> `Copy link to selected time range`
1. Open a new tab and paste the link in


crates/top/rerun/src/commands/entrypoint.rs
```diff
@@ -894,6 +912,7 @@ fn run_impl(
                         startup_options,
                         cc,
                         re_viewer::AsyncRuntimeHandle::new_native(tokio_runtime_handle),
+                        (command_sender, command_receiver),
```
Wumpf, line 915: I think we should solve this differently. Instead of passing in channels, why not tell send commands to the app from the outside
`re_viewer::App` already exposes a bunch of things, why not add an ability to send commands to a live app. This is also super useful for anyone hosting rerun in their own rust application and wanting to instrument it

crates/viewer/re_viewer/src/app.rs
```diff
@@ -314,7 +314,7 @@ impl App {
             screenshotter.screenshot_to_path_then_quit(&creation_context.egui_ctx, screenshot_path);
         }
 
-        let (command_sender, command_receiver) = command_channel();
+        let (command_sender, command_receiver) = command_channel;
```
Wumpf, line 317: see `entrypoint.rs` I think this should be done differently

crates/viewer/re_viewer/src/app.rs
```diff
@@ -1098,6 +1128,40 @@ impl App {
...
 
+    fn run_copy_time_range_link_command(&mut self, store_context: Option<&StoreContext<'_>>) {
+        let Some(entity_db) = store_context.as_ref().map(|ctx| ctx.recording) else {
+            return;
+        };
+
+        let base_url = match &entity_db.data_source {
+            Some(SmartChannelSource::RerunGrpcStream { url }) => url,
+            // not a recording endpoint
+            _ => return,
```
Wumpf, line 1139: if we can't hide the action in this case, can we at least log a warning that this doesn't have an effect? I fell for this trying to run the command without having a grpc connection and then nothing happened

Wumpf (changes requested):
awesome feature

the `on_cmd` pattern feels a lil bit hacky to me but I can't think of a cleaner and as flexible alternative right now 🤷 

Think we could do better on the dependency & channel passing (if possible) though

crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -180,21 +190,61 @@ pub async fn stream_recording_async(
...
+                }),
+                chunk_ids,
+            })
+            .await
+            .map_err(TonicStatusError)?
+            .into_inner()
+    } else {
+        client
+            .fetch_recording(FetchRecordingRequest {
+                recording_id: Some(RecordingId {
```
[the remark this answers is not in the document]
jprochazk: Yeah, this will have to change a lot at some point. Right now we fetch all the chunks from the chunk IDs we're given, because we don't have "seek to next part of recording" functionality at all yet.

crates/store/re_grpc_client/src/lib.rs
```diff
@@ -64,6 +64,12 @@ pub enum StreamError {
     InvalidSorbetSchema(#[from] re_sorbet::SorbetError),
 }
 
+// TODO(ab, andreas): This should be replaced by the use of `AsyncRuntimeHandle`. However, this
+// requires:
+// - `AsyncRuntimeHandle` to be moved lower in the crate hierarchy to be available here (unsure
+//   where).
+// - Make sure that all callers of `DataSource::stream` have access to an `AsyncRuntimeHandle`
+//   (maybe it should be in `GlobalContext`?).
```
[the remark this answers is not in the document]
jprochazk: I didn't write this, just moved it. I thought you guys talked about it already 🤔

crates/utils/re_uri/Cargo.toml
```diff
@@ -15,5 +15,8 @@ version.workspace = true
 workspace = true
 
 [dependencies]
+re_log_types.workspace = true
```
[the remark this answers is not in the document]
jprochazk: Yeah, I tried that, and decided to just add the dependency. We could have them duplicated here and do more conversions... but it's not necessary, because there are no circular dependency issues or anything like that. Yet.

crates/viewer/re_viewer_context/src/global_context/command_sender.rs
```diff
@@ -101,6 +101,12 @@ pub enum SystemCommand {
         timeline: re_chunk::Timeline,
     },
 
+    SetLoopSelection {
```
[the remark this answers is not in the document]
jprochazk: I believe if the range is too small or empty, then it just doesn't set the loop selection at all. But I need to test that.

crates/top/rerun/src/commands/entrypoint.rs
jprochazk: What do you mean by "add ability to send commands to a live app"? What would that look like?

I agree that this isn't a great solution. I tried to minimizee the changes to the flow of `entrypoint` here. I would personally much prefer constructing the `App` first, then parsing the data sources, and finally adding them all as receivers, instead of the opposite. Then it would always be possible to give the data sources access to the command sender.

jprochazk: We could have a separate channel for `DataSourceCommand`s, and the receiver gets passed to the Viewer, which would read from it like it does from smart channels. Not sure if that's what you meant.

crates/store/re_grpc_client/src/lib.rs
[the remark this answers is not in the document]
Wumpf: lol then maybe we wrote that in our pairing session earlier this week

crates/top/rerun/src/commands/entrypoint.rs
jprochazk: (we decided to refactor this later)
* https://github.com/rerun-io/rerun/issues/9113

jprochazk:
Other than addressing review, copying link to time range on web will append the recording url to the viewer url, so it's a kind of "one click sharing" feature 🙂.

Wumpf (approved):

Issue #9034, opened 2025-02-13
Add support for `RowIdColumnDescriptor` to  `re_sorbet::ColumnDescriptor`


abey79:
As the title say. It has implications on the dataframe API though, which currently does not expose row id.

PR #9043, opened 2025-02-14
Add `ComponentDescriptorRef` to `re_sorbet`

abey79:
### Related

* Related to #9034
* Required for #9018 

### What

This PR introduce the `ComponentDescriptorRef` type to `re_sorbet`. It's like `ComponentDescriptor` but:
- It has refs instead of owned copies.
- It has a `RowId` enum (which would be difficult to add to `ComponentDescriptor` because of the ramifications on the dataframe API—let's wait for the RowID <-> Index column merge for that).


crates/store/re_sorbet/src/column_descriptor_ref.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub fn short_name(&self) -> String {
+        match self {
+            Self::RowId(descr) => descr.name().to_owned(),
+            Self::Time(descr) => descr.timeline.name().to_string(),
+            Self::Component(descr) => descr.component_name.short_name().to_owned(),
+        }
+    }
+
+    #[inline]
+    pub fn is_static(&self) -> bool {
```
emilk, line 35: I know the existing `ColumnDescriptor` already had these functions, and that they are undocumented. But do we really need them? I prefer to remove them, as I find it hard to explain why the row-id column is not static, for instance.

If we do want them, then let us add docstrings for why their indended use is

crates/store/re_sorbet/src/row_id_column_descriptor.rs
```diff
@@ -34,6 +34,12 @@ impl RowIdColumnDescriptor {
         Self {}
     }
 
+    #[inline]
+    #[expect(clippy::unused_self)]
+    pub fn name(&self) -> &'static str {
```
emilk, line 39: What is the intentional use for this? If this is supposed to be the default column name, then I think we should use either PascalCase or snake_case. If this is supposed to be a human readable string, then I think "Row ID" is better.

emilk (approved):

crates/store/re_sorbet/src/column_descriptor_ref.rs
abey79: I'll remove all the ones I dont need (which is all but `short_name`)

crates/store/re_sorbet/src/row_id_column_descriptor.rs
abey79: Human readable name, so "Row ID" it is.

PR #9044, opened 2025-02-14
Include static chunks in all-component queries

zehiko:
^.

### DIFF
--- a/crates/store/re_chunk_store/src/query.rs
+++ b/crates/store/re_chunk_store/src/query.rs
@@ -743,7 +743,8 @@ impl ChunkStore {
         chunks
     }
 
-    /// Returns the most-relevant _temporal_ chunk(s) for the given [`RangeQuery`].
+    /// Returns the most-relevant chunk(s) for the given [`RangeQuery`]. Optionally, static chunks
+    /// can be included in the results.
     ///
     /// The returned vector is guaranteed free of duplicates, by definition.
     ///
@@ -753,11 +754,12 @@ impl ChunkStore {
     /// The caller should filter the returned chunks further (see [`Chunk::range`]) in order to
     /// determine how exactly each row of data fit with the rest.
     ///
-    /// **This ignores static data.**
+    ///
     pub fn range_relevant_chunks_for_all_components(
         &self,
         query: &RangeQuery,
         entity_path: &EntityPath,
+        include_static: bool,
     ) -> Vec<Arc<Chunk>> {
         re_tracing::profile_function!(format!("{query:?}"));
 
@@ -785,7 +787,23 @@ impl ChunkStore {
 
         debug_assert!(chunks.iter().map(|chunk| chunk.id()).all_unique());
 
-        chunks
+        if include_static {
+            let static_chunks = self
+                .static_chunk_ids_per_entity
+                .get(entity_path)
+                .map(|static_chunks_per_component| {
+                    static_chunks_per_component
+                        .values()
+                        .filter_map(|chunk_id| self.chunks_per_chunk_id.get(chunk_id).cloned())
+                        .collect_vec()
+                })
+                .unwrap_or_default();
+
+            debug_assert!(static_chunks.iter().map(|chunk| chunk.id()).all_unique());
+            static_chunks.into_iter().chain(chunks).collect()
+        } else {
+            chunks
+        }
     }
 
     fn range<'a>(
--- a/crates/store/re_chunk_store/src/query.rs
+++ b/crates/store/re_chunk_store/src/query.rs
@@ -592,6 +592,8 @@ impl ChunkStore {
 
     /// Returns the most-relevant chunk(s) for the given [`LatestAtQuery`].
     ///
+    /// Optionally include static data.
+    ///
     /// The [`ChunkStore`] always work at the [`Chunk`] level (as opposed to the row level): it is
     /// oblivious to the data therein.
     /// For that reason, and because [`Chunk`]s are allowed to temporally overlap, it is possible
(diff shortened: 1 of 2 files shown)


crates/store/re_chunk_store/src/query.rs
```diff
@@ -785,7 +787,23 @@ impl ChunkStore {
...
+                        .filter_map(|chunk_id| self.chunks_per_chunk_id.get(chunk_id).cloned())
+                        .collect_vec()
+                })
+                .unwrap_or_default();
+
+            debug_assert!(static_chunks.iter().map(|chunk| chunk.id()).all_unique());
+            static_chunks.into_iter().chain(chunks).collect()
+        } else {
+            chunks
+        }
```
teh-cmc, line 806: You need to do this first, and then for every component that didn't get any static chunks (and only those), extend the results with the existing temporal logic.

zehiko: that makes sense, tnx

crates/store/re_chunk_store/src/query.rs
```diff
@@ -587,27 +590,106 @@ impl ChunkStore {
...
+    /// The returned vector is guaranteed free of duplicates, by definition.
+    ///
+    /// The [`ChunkStore`] always work at the [`Chunk`] level (as opposed to the row level): it is
+    /// oblivious to the data therein.
+    /// For that reason, and because [`Chunk`]s are allowed to temporally overlap, it is possible
+    /// that a query has more than one relevant chunk.
+    ///
+    /// The caller should filter the returned chunks further (see [`Chunk::latest_at`]) in order to
+    /// determine what exact row contains the final result.
+    pub fn latest_at_relevant_chunks_for_all_components_include_static(
```
teh-cmc, line 673: Is this whole method an oversight?

crates/store/re_chunk_store/src/query.rs
```diff
@@ -587,27 +590,106 @@ impl ChunkStore {
...
     ///
-    /// The returned vector is guaranteed free of duplicates, by definition.
+    /// Optionally include static data.
     ///
     /// The [`ChunkStore`] always work at the [`Chunk`] level (as opposed to the row level): it is
     /// oblivious to the data therein.
     /// For that reason, and because [`Chunk`]s are allowed to temporally overlap, it is possible
     /// that a query has more than one relevant chunk.
     ///
+    /// This is a multi-component query, so it may yield duplicate chunks, but any given chunk
```
teh-cmc, line 602: > This is a multi-component query, so it may yield duplicate chunks

Shouldn't it dedupe them before returning then? I'm not sure what this endpoint is for, but I suspect that every downstream consumer of this will want to dedupe the data, if only for performance reason (e.g. don't try and fetch the same chunk twice from a remote redap store)?

> but any given chunk will never contain both _static_ and _temporal_ data.

I would remove this, as this is nothing specific to this function: that's just how chunks work (and as someone who knows how chunks work, this confuses me into thinking something weird is happening, when it in fact isn't).

crates/store/re_chunk_store/src/query.rs
```diff
@@ -587,27 +590,106 @@ impl ChunkStore {
...
+                .temporal_chunk_ids_per_entity_per_component
+                .get(entity_path)
+                .and_then(|temporal_chunk_ids_per_timeline_per_component| {
+                    temporal_chunk_ids_per_timeline_per_component.get(&query.timeline())
+                })
+                .map(|temporal_chunk_ids_per_component| {
+                    temporal_chunk_ids_per_component
+                        .iter()
+                        .filter(|(component_name, _)| {
+                            static_chunks_per_component.contains_key(component_name)
```
teh-cmc, line 637: I might be misreading this, but isn't that the opposite of what you're trying to achieve? I.e. you want to discard any temporal data that belongs to a component for which you already found static data.

```suggestion
                            !static_chunks_per_component.contains_key(component_name)

File: crates/store/re_chunk_store/src/query.rs [hunk 5]
```diff
@@ -743,26 +831,71 @@ impl ChunkStore {
...
-    ///
-    /// The returned vector is guaranteed free of duplicates, by definition.
+    /// Returns the most-relevant chunk(s) for the given [`RangeQuery`].
     ///
     /// The criteria for returning a chunk is only that it may contain data that overlaps with
-    /// the queried range.
+    /// the queried range, or that it is static.
+    ///
+    /// This is a multi-component query, so it may yield duplicate chunks, but any given chunk
+    /// will never contain both _static_ and _temporal_ data.
```
teh-cmc, line 840: Same remarks

crates/store/re_chunk_store/src/query.rs
```diff
@@ -743,26 +831,71 @@ impl ChunkStore {
...
+                    self.temporal_chunk_ids_per_entity_per_component
+                        .get(entity_path)
+                        .and_then(|temporal_chunk_ids_per_timeline_per_component| {
+                            temporal_chunk_ids_per_timeline_per_component.get(&query.timeline())
+                        })
+                        .map(|temporal_chunk_ids_per_component| {
+                            temporal_chunk_ids_per_component
+                                .iter()
+                                .filter(|(component_name, _)| {
+                                    !static_chunks_per_component.contains_key(component_name)
```
teh-cmc, line 876: Ha!

crates/store/re_chunk_store/src/query.rs
```diff
@@ -743,26 +831,71 @@ impl ChunkStore {
...
+                // need to make sure that the resulting chunks' global time ranges intersect with the
+                // time range of the query itself.
+                .filter(|chunk| {
+                    chunk
+                        .timelines()
+                        .get(&query.timeline())
+                        .is_some_and(|time_column| {
+                            time_column.time_range().intersects(query.range())
+                        })
+                });
```
teh-cmc, line 894: That's non-trivial: I would put that code outside the if statement so both paths share it (use `Either`).

teh-cmc (changes requested):

crates/store/re_chunk_store/src/query.rs
jprochazk: > I'm not sure what this endpoint is for, but I suspect that every downstream consumer of this will want to dedupe the data, if only for performance reason

Right now, it's only for the time range query in redap. I agree that de-duplicating probably makes sense.

crates/store/re_chunk_store/src/query.rs
```diff
@@ -587,40 +587,78 @@ impl ChunkStore {
...
+                })
+                .into_iter()
+                .flatten()
+                .filter_map(|temporal_chunk_ids_per_time| {
+                    self.latest_at(query, temporal_chunk_ids_per_time)
+                })
+                .flatten()
+                // The latest_at queries may yield duplicate chunks, and it's unlikely
+                // anyone will use this without also deduplicating first.
+                .unique_by(|chunk| chunk.id());
```
jprochazk, line 648: Am I wrong in thinking that the only place where I need to de-duplicate is the temporal chunks? They shouldn't overlap with the static chunks at this point. But I added back the debug assert just in case, and it fails.

jprochazk: Ah, nope. I need to dedupe _both_ static and temporal chunks. :facepalm: I'll just do it when they're combined to avoid the extra allocation...

jprochazk:
Dataplatform PR coming right up

PR #9051, opened 2025-02-17
Implement `TryFrom<&str>` for `redap::Origin`

grtlr:
### What

With `impl TryFrom<&str> for redap::Origin`, we can bring the URI conversion logic to `redap-cli`, too.

### Related

Ultimately, we will want to move this entire logic to a dedicated `re_redap_protos` crate:

* https://github.com/rerun-io/dataplatform/issues/60


abey79 (approved):
makes sense 👍🏻

PR #8733, opened 2025-01-20
Update to arrow2 53.4; use it for IPC serialization

emilk:
### Related
* Part of #3741 
* Possible thanks to https://github.com/apache/arrow-rs/issues/6803

PR #9057, opened 2025-02-17
Improve comment regarding arrow1-IPC-serialization

teh-cmc:
The arrow2 serializer drops half of the metadata in the process, which is making it hard to work on manifests...

As far as I know there is no problem with enabling arrow1 on this path (and I can confirm that it fixes my metadata issues).

* Related: https://github.com/rerun-io/rerun/pull/8733

### DIFF
--- a/crates/store/re_log_encoding/src/codec/arrow.rs
+++ b/crates/store/re_log_encoding/src/codec/arrow.rs
@@ -3,7 +3,7 @@ use super::CodecError;
 use arrow::array::RecordBatch as ArrowRecordBatch;
 
 /// TODO(#3741): switch to arrow1 once <https://github.com/apache/arrow-rs/issues/6803> is released
-const SERIALIZE_WITH_ARROW_1: bool = true; // I _think_ we can use arrow1 here, because we don't encounter the above bug in this context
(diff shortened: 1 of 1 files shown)


teh-cmc:
Well, look that this cannot be enabled after all :/

teh-cmc:
Guess I'll keep it disabled and update the comment.

crates/store/re_log_encoding/src/codec/arrow.rs
```diff
@@ -3,7 +3,7 @@ use super::CodecError;
 use arrow::array::RecordBatch as ArrowRecordBatch;
 
 /// TODO(#3741): switch to arrow1 once <https://github.com/apache/arrow-rs/issues/6803> is released
-const SERIALIZE_WITH_ARROW_1: bool = false; // I _think_ we can use arrow1 here, because we don't encounter the above bug in this context
+const SERIALIZE_WITH_ARROW_1: bool = false; // Enabling this will break cross-language roundtrips.
```
emilk, line 6: Maybe add some more context, like how and why?

emilk (approved):

Issue #4974, opened 2024-01-30
Sync egui_plot item visibility with entity visibility


Wumpf:
Annoyingly there's two types of visibility right now in the plot: egui_plot "hidden item" (`PlotMemory::hidden_items`) and our regular visibility flag. The former is set by clicking on the legend in egui_plot, the later via our regular api.

To clean this up the hidden items need to be accessible in egui plot. Note that my last change on egui plots here
* https://github.com/emilk/egui/pull/3920

actually made this a bit of a mess now: This change introduces ids that can be optionally set, but internally plot items are uniquely identified by name! We should use the id everywhere but automatically derive it from the name if not specified otherwise.

PR #9061, opened 2025-02-17
Fix inconsistencies in plot highlights

Wumpf:
### Related

* To fully work needs https://github.com/emilk/egui_plot/pull/81

### What

Noticed these issues while looking into solving
* https://github.com/rerun-io/rerun/issues/4974

I noticed that syncing of highlights is also all over the place. Seemed related enough to sort it out right away!

https://github.com/user-attachments/assets/57ad14f9-a586-4ca9-bb19-b2a2116ebbc7

Drawback is that for some already large markers this may be a bit too noisy:

https://github.com/user-attachments/assets/74c8d3c5-31ee-4bca-8463-4b8659715fa2

But I believe it's the right tradeoff!

Internally it also moves over to `InstancePath` over `EntityPath` for plot association in order to handle multi-scalar cases without surprises.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2166,8 +2166,7 @@ dependencies = [
 [[package]]
 name = "egui_plot"
 version = "0.31.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


abey79 (approved):
Makes sense!

PR #9062, opened 2025-02-17
Introduce a `LogMsg` that keeps track of its origins in bytes

teh-cmc:
This introduces `StreamingLogMsg` (which is a terrible name, holy), which encapsulates a `LogMsg` and tells you exactly where, in terms of byte, you can find that data in the underlying storage.


zehiko (approved):

teh-cmc:
Merging, but the door remains open to anyone that has a better name for this.

PR #9064, opened 2025-02-17
Switch back to the viewer when clicking a recording URI

abey79:
### Related

- Follow-up to #9018

### What

So far, clicking on a recording URI in the redap browser would silently load the recording in the background. With this PR, the UI switches back to viewer mode whenever a new recording is loaded, unless the user cmd/ctrl- or middle-clicked the link, in which case the UI remains in redap browser mode (mimicking the load tab in the background behaviour of browser).

This involved a number of changes:
- `RecordingUri` is now removed, because the default `arrow_ui` handling of links is sufficient.
- `rerun://` links are globally handled by the viewer, taking into account the same-tab/new-tab behaviour in egui `OpenUrl` event.
- `ui.re_hyperlink` can now be specified is the new-tab behaviour is forced or not. It is the case for our documentation link, but not for generic urls (e.g. from `arrow_ui()`
- Some cleanup around `redap::Scheme` happened.


crates/viewer/re_component_ui/src/recording_uri.rs
```diff
@@ -26,7 +26,10 @@ pub fn singleline_view_recording_uri(
             })
             .inner;
 
-        if response.clicked() {
+        if response.clicked() || response.middle_clicked() {
+            let do_no_switch_to_viewer =
+                response.middle_clicked() || ui.input(|i| i.modifiers.command);
```
emilk, line 31: First of all, do not use negations in variable names (https://github.com/rerun-io/rerun/blob/main/CODE_STYLE.md#naming - read [Code Complete](https://en.wikipedia.org/wiki/Code_Complete) for more details 😉).


```rs
let open_in_background =
    response.middle_clicked() || ui.input(|i| i.modifiers.command);
```

----

But if you want to be consistent with egui you can instead:

```
let open_in_background = ui.ctx.output(|o| o.platform_output.commands.iter().any(|c|
   if let OutputCommand::OpenUrl((url) = c {
        url.new_tab
   } else {
       false 
	}));      
```


Yes, that's quite verbose, but it will work consistently with egui (I believe alt-click also opens in a background tab, for instance).

We could also consider having the hyperlink widget return the `OpenUrl` command, if any.

crates/viewer/re_component_ui/src/recording_uri.rs
```diff
@@ -35,7 +38,10 @@ pub fn singleline_view_recording_uri(
             match data_source.stream(None) {
                 Ok(re_data_source::StreamSource::LogMessages(rx)) => _ctx
                     .command_sender()
-                    .send_system(SystemCommand::AddReceiver(rx)),
+                    .send_system(SystemCommand::AddReceiver {
+                        rx,
+                        switch_to_viewer: !do_no_switch_to_viewer,
```
emilk, line 43:
```suggestion
                        switch_to_viewer: !did_open_in_new_tab,
```

crates/viewer/re_component_ui/src/recording_uri.rs
```diff
@@ -35,7 +38,10 @@ pub fn singleline_view_recording_uri(
             match data_source.stream(None) {
                 Ok(re_data_source::StreamSource::LogMessages(rx)) => _ctx
                     .command_sender()
-                    .send_system(SystemCommand::AddReceiver(rx)),
+                    .send_system(SystemCommand::AddReceiver {
```
emilk, line 41: Wait, do we generate both a `egui::OutputCommand::OpenUrl`, AND explicitly handle it here?

So that if somone installs an OS handler for `rerun://` URLs, this will open the urls twice?

An alternative is to catch the `rerun://` URLs here instead:

https://github.com/rerun-io/rerun/blob/8242c88d5511e840710f86bace673158eb83a290/crates/viewer/re_viewer/src/app_state.rs#L773-L787

That way, `rerun://` URLs will work _anywhere in our UI_. Users could even put `rerun://` in their arrow data, and we will show them as URLs, thanks to
https://github.com/rerun-io/rerun/blob/8242c88d5511e840710f86bace673158eb83a290/crates/viewer/re_ui/src/ui_layout.rs#L102-L108

crates/viewer/re_component_ui/src/recording_uri.rs
```diff
@@ -26,7 +26,10 @@ pub fn singleline_view_recording_uri(
             })
             .inner;
 
-        if response.clicked() {
+        if response.clicked() || response.middle_clicked() {
```
emilk, line 29: I see the parent scope is surrounded by `#[cfg(not(target_arch = "wasm32"))]` - why?

abey79: That's a good question, I noticed that too. I think it's unwarranted, but it's currently difficult to test. My next todo is to add the ability to add a new redap server from the UI. When we have that, we can test how stuff behaves with the web viewer and deal with this conditional.

crates/viewer/re_component_ui/src/recording_uri.rs
abey79: > Wait, do we generate both a `egui::OutputCommand::OpenUrl`, AND explicitly handle it here?

It seems so indeed, good catch. We could get rid of the `RecordingUri` component altogether!

abey79: Oh, actually no, this did not double-open the URL, because it's using `Ui::link()`, which doesn't trigger a `OpenUrl` event.

crates/store/re_grpc_client/src/redap/address.rs
```diff
@@ -64,6 +68,40 @@ impl Scheme {
...
+        }
+    }
+}
+
+impl TryFrom<&str> for Scheme {
+    type Error = InvalidScheme;
+
+    fn try_from(url: &str) -> Result<Self, Self::Error> {
+        if url.starts_with("rerun://") {
+            Ok(Self::Rerun)
```
emilk, line 96: Do we need to distinguish `rerun://` from `rerun+https://`, or can we just return `RerunHttps` here and remove the `Rerun` alternative?

crates/store/re_grpc_client/src/redap/address.rs
```diff
@@ -64,6 +68,40 @@ impl Scheme {
             Self::RerunHttp => "http",
         }
     }
+
+    /// Converts a rerun url into a canonical http or https url.
+    fn canonical_url(&self, url: &str) -> String {
```
emilk, line 73: Is this the best name for this? Maybe "web_url" or "http_url"? 🤔

crates/store/re_grpc_client/src/redap/address.rs
```diff
@@ -64,6 +68,40 @@ impl Scheme {
...
+        if url.starts_with("rerun://") {
+            Ok(Self::Rerun)
+        } else if url.starts_with("rerun+http://") {
+            Ok(Self::RerunHttp)
+        } else if url.starts_with("rerun+https://") {
+            Ok(Self::RerunHttps)
+        } else {
+            Err(InvalidScheme)
+        }
+    }
```
emilk, line 104: Looks like `enum Scheme` is not used anywhere outside this module - can we make it non-`pub`?

crates/viewer/re_viewer/src/app.rs
```diff
@@ -571,9 +570,19 @@ impl App {
...
-            SystemCommand::AddReceiver(rx) => {
+            SystemCommand::AddReceiver {
+                rx,
+                switch_to_viewer,
+            } => {
                 re_log::debug!("Received AddReceiver");
                 self.add_receiver(rx);
+                if switch_to_viewer {
+                    self.state.display_mode = DisplayMode::Viewer;
+                }
```
emilk, line 581: This could be its own command

emilk (approved):

crates/viewer/re_ui/src/ui_ext.rs
```diff
@@ -980,7 +986,13 @@ pub trait UiExt {
                 .add(Button::image_and_text(image, text))
                 .on_hover_cursor(egui::CursorIcon::PointingHand);
 
+            // Inspired from `egui::Ui::Hyperlink::ui()`
             if response.clicked() {
+                ui.ctx().open_url(egui::OpenUrl {
+                    url: url.to_string(),
+                    new_tab: always_new_tab || ui.input(|i| i.modifiers.any()),
```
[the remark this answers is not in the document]
abey79: TODO: docstring

crates/store/re_grpc_client/src/redap/address.rs
grtlr: If we want to restore the original URL (for example for debugging) we need to have all variants.

Issue #9021, opened 2025-02-12
Handle visibility of multi-scalar time/point series


Wumpf:
Need to introduce some form of "visibility vector"

To figure out:
* how to model this data
* how is it displayed in override menu
    * this is a common problem here, also with names and colors as all of these become arrays!

May need/want to fix in the process
* https://github.com/rerun-io/rerun/issues/4974

Wumpf:
Override menu doesn't show nicely. Otherwise we're done - data is modelled and usable from the legend menu

Wumpf:
Closing override ui limitations in favor of this more general issue:
* https://github.com/rerun-io/rerun/issues/9331

PR #9067, opened 2025-02-17
Move `rerun.components.blueprint.Visible` to `rerun.components.Visible`

Wumpf:
### Related

* this turned out to be necessary in order to address https://github.com/rerun-io/rerun/issues/9021
   * we want per-series visibility and introducing a new component for that would be very much counter to our vision of handling "scoping" with tags (whereas component names are about (as far as possible) context agnostic semantics) 

### What

Surprisingly little friction on this. Originally wanted to add a migration guide entry for this, but it practically doesn't affect user interface (which is a bad thing! we have an open issue that it's hard/not possible to control visibility in blueprint code!)


emilk (approved):

Wumpf:
Need to have a look at those link errors, they complain about `Visible` component!

PR #9068, opened 2025-02-17
Infer column partition size from shape in `Scalar.columns()`

### LINKED ISSUE #9060: `Scalar.columns` assumes 1 scalar value per row
When logging a numpy array as column of scalar values, currently it assumes 1 scalar value per row and partitions it as such when creating columns. However, numpy arrays contain their shape which can be used to pick the correct partition size.

oxkitsune:
### Related

Closes #9060

### What

We now infer partition size from array-like shape in `send_columns` when logging Scalar values. 

I've also added an Archimedes' spiral example (in Cartesian coordinates 😄) to plots to demonstrate this:

image

### DIFF
--- a/crates/build/re_types_builder/src/codegen/python/mod.rs
+++ b/crates/build/re_types_builder/src/codegen/python/mod.rs
@@ -2488,6 +2488,17 @@ fn quote_clear_methods(obj: &Object) -> String {
     ))
 }
 
+fn quote_kwargs(obj: &Object) -> String {
+    obj.fields
+        .iter()
+        .map(|field| {
+            let field_name = field.snake_case_name();
+            format!("'{field_name}': {field_name}")
+        })
+        .collect_vec()
+        .join(",\n")
+}
+
 fn quote_partial_update_methods(reporter: &Reporter, obj: &Object, objects: &Objects) -> String {
     let name = &obj.name;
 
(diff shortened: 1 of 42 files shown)


Wumpf (changes requested):
this in need of some python unit tests otherwise this feature won't last for more than one release if at all ;-)

I'm quite unhappy with the name guarded exception for scalar here. this means that this falls immediately apart again when we want to log things like arrays of arrays of vec3 for the same purpose.
Isn't the real problem that our arrow serialization flattens the array incorrectly in `len(batches[0]._batch.as_arrow_array())`? Taking the vec3 example on points this works out fine afaik since we actually preserve the batches. The reason the same is not happening with scalars is probably all about the archetype definition still assuming mono component

Happy to discuss details on a call

Wumpf:
addendum: the archetype notion of mono vs "multi" component has no influence on logging. we have to take a closer look at serialization instead

Wumpf:
```py
class Float64Batch(BaseBatch[Float64ArrayLike]):
    _ARROW_DATATYPE = pa.float64()

    @staticmethod
    def _native_to_pa_array(data: Float64ArrayLike, data_type: pa.DataType) -> pa.Array:
        array = np.asarray(data, dtype=np.float64).flatten()
        return pa.array(array, type=data_type)
```

the `.flatten()` piece here is the core of the issue I believe.
Removing this might have some ugly ripple effects that we have to think about

oxkitsune:
Had a call with @Wumpf and we decided to try the following approach:

Instead of partitioning all of columns with the same size like this:
```py
columns = [batch.partition(lengths) for batch in batches]
```

We will instead compute the partition size for each parameter based on whether its a primitive array or not. Which lets us get rid of the special case and the filthy name check.

Wumpf (approved):
awesome! 🥳

PR #9071, opened 2025-02-18
Remove `grpc` as a feature flag (always have it enabled)

emilk:
Let's keep it simple.

If we ever need to build something without gRPC, we can ad back the feature flag again. But likely, YAGNI


emilk:
@rerun-bot full-check

abey79 (approved):

PR #9072, opened 2025-02-18
Better error message for using MsgPack serializer

jprochazk:


crates/store/re_log_encoding/src/lib.rs
```diff
@@ -121,6 +122,9 @@ pub enum OptionsError {
     #[error("Unknown compression: {0}")]
     UnknownCompression(u8),
 
+    #[error("Attempted to use the removed MsgPack serializer, which is no longer supported")]
+    RemovedMsgPackSerializer,
+
```
grtlr, line 127: This is an error that we expect many users to see—should we try to improve the error message (e.g. explicitly mentioning that old `.rrd` files are not longer supported)?

Our users might not even know what a `MsgPack` serializer is.

crates/store/re_log_encoding/src/lib.rs
```diff
@@ -121,6 +122,9 @@ pub enum OptionsError {
     #[error("Unknown compression: {0}")]
     UnknownCompression(u8),
 
+    #[error("Attempted to use the removed MsgPack serializer, which is no longer supported")]
```
grtlr, line 125:
```suggestion
    #[deprecated(since = "1.0.0")]
    #[error("Attempted to use the removed MsgPack serializer, which is no longer supported")]
```

Maybe we should start adding breadcrumbs for ourselves to remind us to remove these things?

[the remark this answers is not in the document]
grtlr: Maybe something along the lines of:

> `.rrd` files that were created before Rerun v0.22 are no longer supported.

Maybe we add a link to the release notes too?

jprochazk: Agreed, but not sure about `deprecated`. That just turns into `#[expect(deprecated)]` wherever it's used, which in this case is the same file. I'll add a comment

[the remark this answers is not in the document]
jprochazk: But that doesn't make sense in the context of where this error is produced. You're not deserializing an RRD, you're only decoding a `FileHeader`. We don't know what your intentions are, and we should be as specific as possible about what the _actual_ issue is. The header contained `0x01` in the `serializer` offset, so we should report _that_ error.

The user will see the following warning:
```
Found log stream with Rerun version 0.22.0, which is incompatible with the local Rerun version 0.23.0. Loading will try to continue, but might fail in subtle ways.
```
It should be much more obvious that they can't load that RRD file in that version of the Viewer.

If we want to be more strict about version mismatches, that's a different topic altogether. I'm just trying to give us (the developers) a clearer message for what went wrong here, because we don't see the version mismatch warning.

crates/store/re_log_encoding/src/lib.rs
grtlr: Thank you for the clarification—I didn't know about the version mismatch warning.

grtlr (approved):

PR #9079, opened 2025-02-18
Plot series visibility separarate from entity visibility, handle multi-series visibility

### LINKED ISSUE #4974: Sync egui_plot item visibility with entity visibility

Wumpf:
### Related

* Requires https://github.com/emilk/egui_plot/pull/82
* Fixes #4974
    * more obsoletes though, details below
* Most of #9021 
    * fixed in the strictest sense, but I still want some better UX in the override menu 

### What

Allows to control the visibility of individual time series on a single entity. There is intentionally no interaction with the overall visible state of the entity.
Demonstration of what this means for single series on an entity:

https://github.com/user-attachments/assets/e0d5a243-98c4-467c-8e83-b6314e23850f

This gets more interesting once we have several series on an entity though. Here, one of the series was disabled to begin with:

https://github.com/user-attachments/assets/e810b13e-2cbd-4b68-8faa-28dce92dcea6

As illustrated, one thing we're still missing is the ability to work with multiple visibility values in the overrides.

Unfortunately, I had to introduce a new component for this rather than (as planned) re-use the `Visible` component: our "regular" entity visibility is stored on the override path and we don't want to conflict with this.
Once we have the full tagging story (tried to not make it worse here) we should be able to just use `Visible` again.
... alternatively we might make `SeriesVisible` an enum in order to explicitly hide things from the legend 🤔


crates/viewer/re_view_time_series/src/point_visualizer_system.rs
```diff
@@ -247,6 +254,17 @@ impl SeriesPointSystem {
                     .map(|index| (index, ()))
             };
 
+            // TODO(#9020): support multiple series.
+            let num_series = 1;
```
abey79, line 258: Is there anything about this that is more complicated than copy/pasting the changes made to `line_visualizer_system`?

abey79 (approved):
Very nice!

crates/viewer/re_view_time_series/src/point_visualizer_system.rs
Wumpf: not really, no. The complication is mostly me not wanting to do a full copy & paste and re-use some stuff better. Maintaining these both in parallel is annoying

PR #9083, opened 2025-02-19
Move URI handling logic to new `re_uri` crate

### LINKED ISSUE #8411: Choose a good `max_decoding_message_size` for gRPC
The default `StorageNodeClient` decoding size is 4 MB, which is too small as some of our messages can be orders of magnitude larger than that. We need to choose a better value, as it's currently set to `usize::MAX` to work around that

- https://github.com/rerun-io/rerun/pull/8347#discussion_r1879868667

grtlr:
### Related

* Cleanup of #9064
* Adds `/proxy` logic from #8876
* Closes #8411

### What

Moves the `address.rs` logic out of `re_grpc_client` into its own crate (because you can never have enough crates 📦).

Additionally, we try to follow the mantra of _parse, don't validate_ here and pass around `CatalogEndpoint` and `RecordingEndpoint` objects, rather than URI `String`s. Doing so gives us stronger type guarantees and avoids having to re-validate in some cases.

### Limitations

_As of now, opening catalog endpoints via the web viewer is not supported, hence the `bail!` in `web_tools.rs`._


ARCHITECTURE.md
abey79, line 1: You need to update the crate schematics as well, as per the comment in that file.

Pro tip: for some reason, I can't always go straight from figma/copy as png to `pixi run upload-image`, so I go to Preview.app and `cmd-N`, `cmd-A`, `cmd-C` in between.

crates/store/re_data_source/src/data_source.rs
```diff
@@ -126,7 +141,7 @@ impl DataSource {
                 follow: false,
             }
         } else {
-            // If this is sometyhing like `foo.com` we can't know what it is until we connect to it.
+            // If this is something like `foo.com` we can't know what it is until we connect to it.
```
abey79, line 144: what even is `typo` doing? 🤦🏻

abey79 (approved):
That's a very nice refactor, thanks.

crates/utils/re_uri/src/lib.rs
```diff
@@ -4,47 +4,19 @@
...
+    endpoints::{catalog::CatalogEndpoint, recording::RecordingEndpoint},
+    error::Error,
+};
 
 /// The different schemes supported by Rerun.
 ///
 /// We support `rerun`, `rerun+http`, and `rerun+https`.
 #[derive(Debug, PartialEq, Eq, Copy, Clone, Hash)]
-pub enum Scheme {
+enum Scheme {
```
abey79, line 19: I can make those changes myself but heads up: I need to build a form-like UI for an `Origin`, with a scheme popup, a host text field, and a port integer text field. This means I'll need:
- `pub` access to `Scheme`
- ability to build an `Origin` from it's parts

abey79 (approved):
lgtm!

crates/utils/re_uri/src/lib.rs
grtlr: That sounds reasonable!

PR #9086, opened 2025-02-19
Persist servers and allow adding/removing them from the UI

abey79:
### What

With this PR, the servers are now persisted across viewer relaunch in the Redap browser, and the can be added and removed from the UI. The handling of the collection query management has been improved, so was its UI.

Note: this makes a pre-existing `crossbeam-channel` dependency explicit.

Add/remove buttons:

image

<br/><br/>

Add server modal:

image


crates/viewer/re_redap_browser/src/requested_object.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// A handle to an object that is requested asynchronously.
+pub enum RequestedObject<T: Send + 'static> {
+    Pending(Receiver<T>),
+    Completed(T),
+}
+
+impl<T: Send + 'static> RequestedObject<T> {
+    /// Create a new [`Self`] with the given future.
+    #[cfg(not(target_arch = "wasm32"))]
```
abey79, line 13: I've add to duplicate these function because native _wants_ `Send`, but wasm doesn't want it at all. I have no idea if there is a better pattern here.

Cargo.toml
```diff
@@ -183,6 +183,7 @@ const_format = "0.2"
 convert_case = "0.6"
 criterion = "0.5"
 crossbeam = "0.8"
+crossbeam-channel = "0.5"
```
abey79, line 186: was already in the dependency tree

crates/viewer/re_redap_browser/src/requested_object.rs
Wumpf: We should do like wgpu does and add this utility https://github.com/wumpf/wgpu/blob/d24dbc9b0ee54256a256287dc053c8e3cc5398a3/wgpu-types/src/lib.rs#L8026
at least I believe we can solve it with that

abey79: worked 🎉

crates/viewer/re_redap_browser/src/requested_object.rs
```diff
@@ -0,0 +1,64 @@
+use crossbeam_channel::{bounded, Receiver};
+
+use re_viewer_context::{AsyncRuntimeHandle, WasmNotSend};
+
+/// A handle to an object that is requested asynchronously.
+pub enum RequestedObject<T: Send + 'static> {
```
Wumpf, line 6: this turned out really nice and short :)

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -1,94 +1,202 @@
...
+impl<'de> serde::Deserialize<'de> for RedapServers {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let origins = Vec::<re_uri::Origin>::deserialize(deserializer)?;
+
+        let servers = Self::default();
+        for origin in origins {
+            let _ = servers.command_sender.send(Command::AddServer(origin));
```
Wumpf, line 105: brief comment on why we have to delay might be nice for posterity

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -104,158 +212,57 @@ impl RedapServers {
...
-        // clear selection upon clicking the empty space
-        if empty_space_response.clicked() {
-            self.command_queue.lock().push(Command::DeselectCollection);
-        }
     }
 
-    pub fn ui(&self, ctx: &ViewerContext<'_>, ui: &mut egui::Ui) {
-        //TODO(ab): we should display something even if no catalog is currently selected.
+    fn add_server_modal_ui(&mut self, ui: &egui::Ui) {
+        //TODO(ab): borrow checker doesn't let me use `with_ctx()` here, I should find a better way
```
Wumpf, line 250: ah that pattern is tricky, don't know either.  Oooooonnneeee day the Polonius borrow checker will solve this, any decade now

Wumpf (approved):
didn't venture too deep into the actual ui code, but data flow turned out nicely now I think :)

crates/viewer/re_redap_browser/src/servers.rs
abey79: in fairness, pulling that off would be pretty great!

PR #9092, opened 2025-02-20
Fix `insta` crate leaking into regular (non-testing) builds

Wumpf:
this meant that even minimal Rust applications building with our SDK dragged in this dependency, adding unnecessary 1.5 cpu seconds on my M1.

To solve this problem "once and for all" we now execute `cargo deny` always with `--exclude-dev`, meaning it no longer checks for any dev dependencies. This way we can flag pure testing crates like insta and criterion as forbidden.
(Note that licenses were already ignored  for dev dependencies, see https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html#the-include-dev-field-optional)

Plus some general updates to our `deny.toml`

### DIFF
--- a/crates/store/re_dataframe/Cargo.toml
+++ b/crates/store/re_dataframe/Cargo.toml
@@ -41,7 +41,6 @@ re_types_core.workspace = true
 # External dependencies:
 anyhow.workspace = true
 arrow.workspace = true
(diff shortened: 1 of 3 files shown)


abey79 (approved):
Makes sense to me at least

PR #9093, opened 2025-02-20
Build Rust snippets first then run them in parallel as a separate step

Wumpf:
This should avoid race conditions we're seeing for instance here https://github.com/rerun-io/rerun/actions/runs/13427212000/job/37513736608 and might be overall faster Also, we still didn't run C++ tests in parallel despite already factoring out the build (running CMake in parallel is known to be racy)

* [x] Tested locally, but want to see main build passing

### DIFF
--- a/docs/snippets/compare_snippet_output.py
+++ b/docs/snippets/compare_snippet_output.py
@@ -71,12 +71,12 @@ def main() -> None:
     parser = argparse.ArgumentParser(description="Run end-to-end cross-language roundtrip tests for all API examples")
     parser.add_argument("--no-py", action="store_true", help="Skip Python tests")
     parser.add_argument("--no-cpp", action="store_true", help="Skip C++ tests")
-    # We don't allow skipping Rust - it is what we compate to at the moment
+    # We don't allow skipping Rust - it is what we compare to at the moment.
     parser.add_argument("--no-py-build", action="store_true", help="Skip building rerun-sdk for Python")
     parser.add_argument(
         "--no-cpp-build",
         action="store_true",
-        help="Skip cmake configure and ahead of time build for rerun_c & rerun_cpp",
+        help="Skip cmake configure and ahead of time build for rerun_c & rerun_prebuilt_cpp",
     )
     parser.add_argument("--full-dump", action="store_true", help="Dump both rrd files as tables")
     parser.add_argument("--release", action="store_true", help="Run cargo invocations with --release")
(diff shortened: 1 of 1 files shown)


Wumpf:
@rerun-bot full-check

docs/snippets/compare_snippet_output.py
```diff
@@ -95,26 +95,20 @@ def main() -> None:
...
-        start_time = time.time()
-        run(["pixi", "run", "-e", "cpp", "cpp-build-snippets"])
-        elapsed = time.time() - start_time
-        print(f"rerun-sdk for C++ built in {elapsed:.1f} seconds")
-        print("")
+        build_cpp()
+
+    build_rust(build_env, args.release, args.target, args.target_dir)
+
+    run(["cargo", "deny", "check"])
```
Wumpf, line 111: oops

Wumpf:
@rerun-bot full-check

docs/snippets/compare_snippet_output.py
```diff
@@ -195,18 +181,57 @@ def main() -> None:
...
+        cmd += ["--release"]
+
+    start_time = time.time()
+    run(cmd, env=build_env, timeout=12000)
+    elapsed = time.time() - start_time
+    print(f"Snippets built in {elapsed:.1f} seconds")
+    print("")
+
+
+def build_python(build_env: dict[str, str]):
```
emilk, line 215:
```suggestion
def build_python_sdk(build_env: dict[str, str]):
```

docs/snippets/compare_snippet_output.py
```diff
@@ -195,18 +181,57 @@ def main() -> None:
...
         check_non_empty_rrd(python_output_path)
     elif language == "rust":
-        rust_output_path = run_rust(example, args.release, args.target, args.target_dir)
+        rust_output_path = run_prebuilt_rust(example, args.release, args.target, args.target_dir)
         check_non_empty_rrd(rust_output_path)
     else:
         assert False, f"Unknown language: {language}"
 
 
+def build_rust(build_env: dict[str, str], release: bool, target: str | None, target_dir: str | None):
```
emilk, line 196:
```suggestion
def build_rust_snippets(build_env: dict[str, str], release: bool, target: str | None, target_dir: str | None):
```

docs/snippets/compare_snippet_output.py
```diff
@@ -195,18 +181,57 @@ def main() -> None:
...
+    print("----------------------------------------------------------")
+    print("Building rerun-sdk for Python…")
+    start_time = time.time()
+    run(["pixi", "run", "py-build", "--quiet"], env=build_env, timeout=12000)
+    elapsed = time.time() - start_time
+    print(f"rerun-sdk for Python built in {elapsed:.1f} seconds")
+    print("")
+
+
+def build_cpp():
```
emilk, line 225:
```suggestion
def build_cpp_snippets():
```

emilk (approved):
Awesome! Small suggestion for name changes, but only if it makes sense

Issue #9108, opened 2025-02-21
Re-enable wheel-tests for `macos-x64`


jleibs:
Our wheel tests currently depend on torch, but torch has dropped support for macos-x64.

We should make it possible to run (most) of our tests without torch and somehow make the dep optional in our pixi environments.

Issue #9105, opened 2025-02-21
Test failures in `test_any_values.py`


GaetanLepage:
**Describe the bug**

Context: updating `rerun` in nixpkgs: https://github.com/NixOS/nixpkgs/pull/379876

Running `pytest` gives us three new failures:
```
=========================== short test summary info ============================
FAILED rerun_py/tests/unit/test_any_values.py::test_any_value - IndexError: list index out of range
FAILED rerun_py/tests/unit/test_any_values.py::test_bad_any_value - assert 0 == 1
FAILED rerun_py/tests/unit/test_any_values.py::test_none_any_value - assert 0 == 1
================== 3 failed, 424 passed, 22 warnings in 3.98s ==================
```

**To Reproduce**
1. Run `pytest`

**Expected behavior**
All tests succeed

**Screenshots**
Not applicable.

**Backtrace**
Here is the log for one of them:
```
________________________________ test_any_value ________________________________

    def test_any_value() -> None:
        values = rr.AnyValues(foo=[1.0, 2.0, 3.0], bar="hello")

        batches = list(values.as_component_batches())

>       foo_batch = batches[0]
E       IndexError: list index out of range

rerun_py/tests/unit/test_any_values.py:14: IndexError
```

**Desktop (please complete the following information):**
 - OS: NixOS

**Rerun version**
0.22.1

Wumpf:
Thanks for reporting! Strange, this works fine on CI and locally. I wonder if there's some dependency shenanigans involved, but it really doesn't look like that.

The test code in question is also a little bit strange here. `as_component_batches()` already returns a `list`, so for some reason it is wrapped in a list here again and then we access element 0. Failing that would imply that the returned list was empty...?

Wumpf:
@GaetanLepage to clarify, what is the setup for running these tests exactly? Our tests generally assume that they run the specified pixi environment, i.e. `pixi run py-test` in this case

abey79:
I could not reproduce with:
```
$ git clone --branch release-0.22.1 https://github.com/rerun-io/rerun

$ cd rerun

$ pixi --version
pixi 0.39.0

$ pixi run py-test
```

jleibs:
> We are indeed not using pixi.

And for a package-manager build using pixi wouldn't make sense as it would provide an alternative self-contained environment using conda, rather than nix for deps. In your case you definitely want to be testing against the other nix packages as you are doing.

Issue #8996, opened 2025-02-11
CI for numpy 2.0 (especially type annotations)


gene-maniro:
Given that numpy 2.0 support was added in https://github.com/rerun-io/rerun/pull/8306, would it be possible to run CI or at least release check that `rerun` is still importable with numpy 2.0?

Example type annotation conflict that would be caught by this import check
`AttributeError: `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead.`

Issue #8486, opened 2024-12-16
`AnyValue` is having issues with numpy 2 (?)


teh-cmc:
This works with numpy < 2:
```sh
$ python tests/python/release_checklist/check_focus.py
# works
```

But fails in an environment with numpy >= 2:
```sh
$ python tests/python/release_checklist/check_focus.py
Traceback (most recent call last):
  File "/home/cmc/dev/rerun-io/rerun/tests/python/release_checklist/check_focus.py", line 57, in <module>
    run(args)
  File "/home/cmc/dev/rerun-io/rerun/tests/python/release_checklist/check_focus.py", line 48, in run
    log_some_views()
  File "/home/cmc/dev/rerun-io/rerun/tests/python/release_checklist/check_focus.py", line 36, in log_some_views
    rr.log(f"a_entity_{i}", rr.AnyValues(empty=0))
                            ^^^^^^^^^^^^^^^^^^^^^
  File "/home/cmc/dev/rerun-io/rerun/venv312/lib/python3.12/site-packages/rerun_sdk/rerun/any_value.py", line 192, in __init__
    batch = AnyBatchValue(name, value, drop_untyped_nones=drop_untyped_nones)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cmc/dev/rerun-io/rerun/venv312/lib/python3.12/site-packages/rerun_sdk/rerun/any_value.py", line 91, in __init__
    np_value = np.atleast_1d(np.array(value, copy=False))
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Unable to avoid copy while creating an array as requested.
If using `np.array(obj, copy=False)` replace it with `np.asarray(obj)` to allow a copy when needed (no behavior change in NumPy 1.x).
For more details, see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword.
```

PR #9109, opened 2025-02-21
Fix support for numpy-2 

jleibs:
### What
- Fixes https://github.com/rerun-io/rerun/issues/9105
- Fixes https://github.com/rerun-io/rerun/issues/8996
- Fixes https://github.com/rerun-io/rerun/issues/8486
- Fixes https://github.com/rerun-io/rerun/issues/6762
- Fixes https://github.com/rerun-io/rerun/issues/4807

In order to move our pixi project to numpy to, we also had to upgrade pytorch to 2.3+ (we might consider jumping all the way to 2.6, but I was trying to find the minimum bump to risk breaking other things).

Unfortunately torch-2.2 is the last version that supported intel macs.

I disabled the wheel-testing on nightly with an issue to re-enable: https://github.com/rerun-io/rerun/issues/9108


emilk:
The last Intel MacBook was released 2021, which is not that long ago (feels like it's been a decade). Still, I feel pretty OK with dropping support for it. I really doubt there are a lot of intel Mac developers out there.

.github/workflows/nightly.yml
```diff
@@ -270,15 +270,16 @@ jobs:
...
-  test-wheel-macos-x64:
-    needs: [checks, build-wheel-macos-x64]
-    name: "macos-x64: Test Wheels"
-    uses: ./.github/workflows/reusable_test_wheels.yml
-    with:
-      CONCURRENCY: nightly-macos-x64
-      PLATFORM: macos-x64
-      WHEEL_ARTIFACT_NAME: macos-x64-wheel
-    secrets: inherit
+  # TODO(#9108): Test macos wheels
```
abey79, line 273: I'm afraid that will backfire, but I don't see a way around it, and this PR answers an actual, pressing need. So yeah, let's take the plunge.

pixi.toml
```diff
@@ -586,18 +589,18 @@ platforms = ["linux-64", "linux-aarch64", "osx-arm64", "osx-64", "win-64"]
...
-rerun-notebook = "==0.17.0"
+rerun-sdk = "==0.22.1"
+rerun-notebook = "==0.22.1"
 
 # EXAMPLES ENVIRONMENT
 [feature.examples-common]
 # Ideally we would could remove`linux-aarch64` from here, but it breaks other things.
 # See: https://github.com/prefix-dev/pixi/issues/1051
-platforms = ["linux-64", "linux-aarch64", "osx-arm64", "osx-64", "win-64"]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
abey79, line 603: While `osx-64` is in definitely in the past, im afraid `linux-aarch64` is in the future. We'll have to deal with that sooner than later.

pixi.toml
```diff
@@ -658,22 +665,6 @@ stdio = { path = "examples/python/stdio", editable = true }
 structure_from_motion = { path = "examples/python/structure_from_motion", editable = true }
 
 # TODO(jleibs): Do this with less duplication
```
abey79, line 667: looks like you just did :)

abey79 (approved):
As noted, there are things that I dont like much here, but then again I like the previous situation even less 😅  Regardless, thanks for the grunt work!

jleibs:
> Still, I feel pretty OK with dropping support for it.

To be clear: we are not dropping support for intel macs. We still build and publish wheels for them. We just can't run our unit-tests until we fix: https://github.com/rerun-io/rerun/issues/9108

Benjamin-Tan:
Is it possible to have numpy1.x compatible api and not have numpy2 as the minimum requirement?

emilk:
I would much prefer not to have to maintain compatibility with both numpy 1 and numpy 2, with all that entails (testing against both, for instance)

Is there anything in particular blocking you?

If it is as simple as changing our code in one or two places so that it works both with numpy 1 and numpy 2, then go ahead an open a PR!

Benjamin-Tan:
Mainly about the challenge of upgrading all the system dependencies to be numpy2 compatible (especially pybind-related components). 

This is my first PR https://github.com/rerun-io/rerun/pull/11100. I tested with `pixi run -e py py-build && pixi run -e py py-test` and it seems to pass all the tests. I'm looking forward to your review and getting this in.

PR #9118, opened 2025-02-24
Make remaining data fetching APIs catalog entry aware

zehiko:
Title.


teh-cmc (approved):
Needs a quick sibling PR on dataplatform, or this will break everything because of the tag re-use.

PR #9120, opened 2025-02-24
Add entity search to the add/remove entity modal

abey79:
### What

Add entity search to the add/remove entity modal, similar to the one in the time panel.

### DIFF
--- a/crates/viewer/re_selection_panel/src/view_entity_picker.rs
+++ b/crates/viewer/re_selection_panel/src/view_entity_picker.rs
@@ -1,9 +1,13 @@
+use std::ops::Range;
+
 use itertools::Itertools;
 use nohash_hasher::IntMap;
+use smallvec::SmallVec;
 
 use re_data_ui::item_ui;
 use re_entity_db::{EntityPath, EntityTree, InstancePath};
 use re_log_types::{ResolvedEntityPathFilter, ResolvedEntityPathRule};
+use re_ui::filter_widget::{format_matching_text, FilterMatcher, FilterState, PathRanges};
 use re_ui::{list_item, UiExt as _};
 use re_viewer_context::{DataQueryResult, ViewId, ViewerContext};
 use re_viewport_blueprint::{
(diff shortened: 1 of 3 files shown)


abey79:
Still need to put the search field out of the scroll area

crates/viewer/re_ui/src/filter_widget.rs
```diff
@@ -172,6 +176,70 @@ impl FilterState {
...
+            stroke
+        };
+
+        egui::Frame::new()
+            .inner_margin(egui::Margin::symmetric(3, 2))
+            .fill(ui.visuals().extreme_bg_color)
+            .stroke(stroke)
+            .corner_radius(visuals.corner_radius)
+            .show(ui, |ui| {
+                egui::Frame::new().show(ui, |ui| {
```
lucasmerlin, line 210: Huh I don't think we need the nested frame right?

crates/viewer/re_ui/src/filter_widget.rs
```diff
@@ -172,6 +176,70 @@ impl FilterState {
...
     }
+
+    /// Display the filter widget as a search field.
+    ///
+    /// In this mode, the filter is active as soon as the query is non-empty. The session remains
+    /// active until the query is cleared.
+    pub fn search_field_ui(&mut self, ui: &mut egui::Ui) {
+        let inner_state = self.inner_state.get_or_insert_with(Default::default);
+
+        let textedit_id = Id::new(ui.id().with("textedit"));
```
lucasmerlin, line 187:
```suggestion
        let textedit_id = ui.id().with("search_field");
```

crates/viewer/re_ui/src/filter_widget.rs
abey79: Oh right, this is a remnant of one of the way I did margins at some point.

crates/viewer/re_selection_panel/src/view_entity_picker.rs
```diff
@@ -246,3 +296,136 @@ fn add_entities_line_ui(
         }
     });
 }
+
+// ---
+
+#[derive(Debug)]
+pub struct EntityData {
```
Wumpf, line 303: Bit wary of this generic name. Can we come up with something more situational? `EntityPickerEntry` or something?

crates/viewer/re_ui/src/filter_widget.rs
```diff
@@ -172,6 +176,68 @@ impl FilterState {
 
         title_response
     }
+
+    /// Display the filter widget as a search field.
+    ///
+    /// In this mode, the filter is active as soon as the query is non-empty. The session remains
+    /// active until the query is cleared.
+    pub fn search_field_ui(&mut self, ui: &mut egui::Ui) {
```
Wumpf, line 184: good distinction, makes sense

Wumpf (approved):
very nice, happy to see this loose end tied and re-using the core of the the filter logic unaltered 👍

Issue #8761, opened 2025-01-21
Use the same `grpc://` client for both Storage Node and re_grpc_server


emilk:
As a user, I want to write `rerun grpc://some-ip` and not care if what I'm talking to is `re_grpc_server` or the storage node.

Right now SN and `re_grpc_server` has no overlap in functionality, but eventually they will (e.g. both should be able to list a catalogue of recordings in it).

This will fix the `temp://` schema thing added in 
* https://github.com/rerun-io/rerun/pull/8730

We could use a REST-like url parsing scheme, e.g. `grpc://ip/recording/recording-id` vs `grpc://ip/events`, and have that decide which `rpc` endpoint is hit.

We should probably unify the two gRPC clients regardless, since eventually there will be overlap in functionality between the two servers.

## Related
* https://github.com/rerun-io/rerun/issues/8876

jleibs:
This issue needs refining -- it currently conflates some storage-node and catalog-server in a way that we're trying to move away from.

We still need to decide if catalog-server and storage-node should share a grpc service definition.

jleibs:
https://github.com/rerun-io/rerun/pull/8965

PR #9129, opened 2025-02-25
Remove `temp://` and some old `MessageProxy` variants

grtlr:
### Related

* Closes: #8761

### What

Removes the temporary `temp://` URI scheme and streamlines handling of URIs.


crates/top/rerun/src/commands/entrypoint.rs
```diff
@@ -862,14 +873,15 @@ fn run_impl(
         Ok(())
     } else if is_another_server_running {
         // Another viewer is already running on the specified address
-        let url = format!("http://{server_addr}")
-            .parse()
-            .expect("should always be valid");
+        let url = format!("http://{server_addr}/proxy");
```
jleibs, line 876: Should this be `rerun+http`?

crates/top/re_sdk/src/web_viewer.rs
```diff
@@ -210,6 +210,12 @@ impl WebViewerConfig {
         };
 
         if let Some(source_url) = source_url {
+            // TODO(jan): remove after we change to `rerun-http`
```
jleibs, line 213: I'm a bit confused by this TODO -- what still needs to change?

jleibs (approved):

crates/top/rerun/src/commands/entrypoint.rs
jprochazk: Yes

crates/top/re_sdk/src/web_viewer.rs
jprochazk: `+` in a query param is treated as whitespace, meaning that to actually connect to a Rerun server from the web viewer, you need to percent-encode the URL first (or at least the `+` character, you can see it all over the place here). We want to change it to `rerun-http`/`rerun-https` instead of `rerun+http`/`rerun+https`, because the `-` character does not have special meaning in any part of the URL.

jleibs:
Looks like we need a missing `/proxy` here:
https://github.com/rerun-io/rerun/blob/8e61d0648f95fa57164c20839fb03abbb7729b2e/crates/top/re_sdk/src/web_viewer.rs#L77

jprochazk:
tested with - opens browser and streams data correctly
```
pixi run -e wheel-test py-build-common -F web_viewer && pixi run -e wheel-test python examples/python/rgbd/rgbd.py --serve
```

jleibs (approved):

Issue #9110, opened 2025-02-22
Add data files to git lfs to test backwards compatibility


emilk:
We want to have automated tests to make sure we have backwards compatible chunk loaders.

We should add files to `git lfs` for all previous versions of all components and archetypes. We should test that
* They load without error
* Produce the same visuals?

### Sourcing the .rrd:s
The output of the snippets is a natural choice.

### Container (file) format
The first thing we want to test is the compatibility of the _arrow_ data. But how do we store the record batches?
`.rrd` is the obvious choice, though that means we also need to commit on the backwards compatibility of that.

emilk:
`compare_snippet_output.py` should compare its output with the `.rrd` files checked in to `git lfs`:

* If missing, it is an error (a new snippet needs to check in an .rrd for its output)
* If it differs, it is an error

### Complications
If we _change_ a snippet (so that it outputs something different), we need to either:
* Update the file on `git lfs`
* Add a _new_ file to `git lfs` alongside the old one
* Add an opt-out flag for comparing to the file on `git lfs` to `snippets.toml`

Since this should be a rather rare occurrence (our snippets tend to stay pretty stable), any one solution will do.

PR #9130, opened 2025-02-25
Fix formatting of empty record batches

emilk:
## What
* Fix formatting of chunks with zero rows
* Include metadata in `rerun rrd print -vv` (same fix as above)
* Simplify formatting of chunks with no metadata

## Related
* Part of https://github.com/rerun-io/rerun/issues/9110


crates/store/re_dataframe/src/snapshots/re_dataframe__query__tests__filtered_index_values.snap
emilk, line 1: I hope you're looking at these diffs in a very wide browser

abey79 (approved):
makes sense

PR #9133, opened 2025-02-25
Test that we don't break backwards compatibility of .rrd files

### LINKED ISSUE #9110: Add data files to git lfs to test backwards compatibility

emilk:
### Related
* Part of #9110

### What
Adds a bunch of `.rrd`:s to `git lfs`, that we should keep there as a test of backwards compatibility.
If any new code breaks loading of these old .rrd files, the CI will complain.

Running locally:
> pixi run check-backwards-compatibility

### What if the CI complains about my PR?
Then your code broke backwards compatibility. Can you make the change so that it doesn't? If not, consult with me! We are not yet promising backwards compatibility, but we should at least make a reasonable effort.

### What does this PR cover?
All Rust snippets are tested, as are the main examples (as of today). This should include most components, but does not cover _everything_. So even if `pixi run check-backwards-compatibility` passes, it is possible that we have broken backwards compatibility in some subtle way. But this is at least a start.

### TODO
* [x] Add a single pixi command to verify the files

### Future work
* Add more files, for improved coverage
* Automatically detect if some snippets/examples are missing from the test dataset


teh-cmc:
@rerun-bot full-check

.gitattributes
```diff
@@ -2,7 +2,7 @@
 *.zip -text diff
 Cargo.lock linguist-generated=false
 pixi.lock linguist-generated=true
-
 # git-lfs stuff
 **/snapshots/**/*.png filter=lfs diff=lfs merge=lfs -text
 tests/assets/video/*.mp4 filter=lfs diff=lfs merge=lfs -text
+*.rrd filter=lfs diff=lfs merge=lfs -text
```
teh-cmc, line 8: I'd feel better if this was `tests/assets/rrd/*.rrd` instead (or a sub `.gitattributes` file in that folder) :grimacing:

tests/assets/rrd/generate-compatibility-rrds.sh
```diff
@@ -0,0 +1,34 @@
...
+
+echo "Generating snippet .rrd files…"
+pixi run -e py docs/snippets/compare_snippet_output.py --no-py --no-cpp
+
+echo "Copying .rrd files to ${DEST_DIR}…"
+find "$SOURCE_DIR" -type f -name "*.rrd" -exec cp {} "${DEST_DIR}" \;
+
+
+echo "Tracking .rrd files with git LFS…"
+git lfs track "*.rrd"
```
teh-cmc, line 30: That shouldn't be here, the .gitattributes rule above is already wildcarded and shouldn't ever change?

teh-cmc (approved):

PR #9136, opened 2025-02-25
Use base32 to format our TUIDs

emilk:
Saves screenspace, e.g. when formatting dataframes

```diff
- 182342300C5F8C327a7b4a6e5a379ac4
+ 30HM4C0CBY634YKV99Q5MDWTRG
```

The old format still parses, so this is NOT a breaking change.

### Related
* Part of https://github.com/rerun-io/dataplatform/issues/250


teh-cmc:
My problem with this is that it's gonna make it very hard to match TUIDs dumped by Rerun-native stuff with TUID dumped by all other external tools and libraries (which will just use the vanilla FixedSizedBytes[16] representation, and are gonna become more and more become).

emilk:
That's a good point 💡

PR #9137, opened 2025-02-25
Add context menu action to copy entity paths to clipboard

Wumpf:
Small improvement to make it easier to extract entity paths for arbitrary purposes:

https://github.com/user-attachments/assets/b1acc95a-c87e-40d9-a521-728dd2440139

### DIFF
--- a/crates/viewer/re_context_menu/src/actions/copy_entity_path.rs
+++ b/crates/viewer/re_context_menu/src/actions/copy_entity_path.rs
@@ -1,0 +1,53 @@
+use itertools::Itertools;
+use re_viewer_context::Item;
+
+use crate::{ContextMenuAction, ContextMenuContext};
+
+pub struct CopyEntityPathToClipboard;
+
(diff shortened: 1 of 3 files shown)


crates/viewer/re_context_menu/src/actions/copy_entity_path.rs
```diff
@@ -0,0 +1,53 @@
...
+            | Item::DataSource(_)
+            | Item::StoreId(_)
+            | Item::Container(_)
+            | Item::View(_) => false,
+            Item::DataResult(..) | Item::InstancePath(_) | Item::ComponentPath(_) => true,
+        }
+    }
+
+    fn label(&self, _ctx: &ContextMenuContext<'_>) -> String {
+        if _ctx.selection.len() == 1 {
```
abey79, line 25:
```suggestion
    fn label(&self, ctx: &ContextMenuContext<'_>) -> String {
        if ctx.selection.len() == 1 {
```

abey79 (approved):
thanks 😍

PR #9145, opened 2025-02-26
Add refresh button to the collection UI

abey79:
title

image

### DIFF
--- a/crates/viewer/re_redap_browser/src/collections.rs
+++ b/crates/viewer/re_redap_browser/src/collections.rs
@@ -34,7 +34,7 @@ pub struct Collections {
 }
 
 impl Collections {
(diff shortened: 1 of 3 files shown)


crates/viewer/re_redap_browser/src/collections.rs
```diff
@@ -34,21 +34,20 @@ pub struct Collections {
 }
 
 impl Collections {
-    pub fn add(
+    pub fn add_or_refresh(
```
jprochazk, line 37: Nit: It now unconditionally re-fetches, so maybe just `fetch`?
```suggestion
    pub fn fetch(
```

jprochazk (approved):

PR #9147, opened 2025-02-26
Add selection change and time(line) change callbacks to Jupyter Notebook and JS APIs

jprochazk:
* Part of https://github.com/rerun-io/rerun/issues/9169

## What

This PR adds basic selection and timeline callbacks to the Viewer:
* `selectionchange`: Fired when an item is selected.
  * We only expose `View`, `Container`, `Entity`, and `Instance`.
* `timelinechange`: Fired when the active timeline changes.
* `timeupdate`: Fired when the time on the active timeline changes.

These callbacks are exposed in the Web Viewer:

```ts
// We use the Node `EventEmitter` pattern in JS:
viewer.on("selectionchange", (selection) => console.log(selection));
```

And in Jupyter Notebooks:

```py
from rerun_notebook import Viewer, ViewerCallbacks

# We (Jan and Gijs) chose to expose this as a class that you inherit from:
class MyCallbacks(ViewerCallbacks):
  def on_selection_change(self, selection):
    print(selection)

# Which is then passed to a `Viewer` instance:
viewer = Viewer()
viewer.register_callbacks(MyCallbacks())
```

Because of how these callbacks pass through the Web Viewer, they are technically also exposed in Rust, via `StartupOptions`. I'm not sure how far we want to take that, so I didn't expose it all the way through the top-level `rerun` crate.


Wumpf:
Part of
* https://github.com/rerun-io/rerun/issues/9169

crates/viewer/re_viewer_context/src/callback.rs
```diff
@@ -0,0 +1,120 @@
...
+// When changing or adding selection items, grep for the following term:
+//   SELECTION ITEM DEFINITION
+// ======================================================================
+
+// SELECTION ITEM DEFINITION
+/// A single item in a selection.
+#[derive(Debug, serde::Serialize)]
+#[serde(tag = "type")]
+#[serde(rename_all = "snake_case")]
+pub enum SelectionItem {
```
Wumpf, line 20: The name makes me a bit unhappy - newcomer to this corner of the codebase will 100% be confused about the relationship between `Item`, which is used for selections, and `SelectionItem` which is used for reporting selections in callbacks :spiral_eyes:

Not entirely obvious how to avoid this _elegantly_. How about not elegant:

```suggestion
/// A single item in a selection as reported in callbacks.
#[derive(Debug, serde::Serialize)]
#[serde(tag = "type")]
#[serde(rename_all = "snake_case")]
pub enum CallbackSelectionItem {
```

I figure for the Rust API (i.e. the `rerun` user facing crate) we could still expose it as `pub type SelectionItem = stuff::CallbackSelectionItem` :thinking:

crates/viewer/re_viewer_context/src/callback.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Fired when the timepoint changes.
+    ///
+    /// Does not fire when `on_seek` is called.
+    pub on_time_update: Rc<dyn Fn(TimeReal)>,
+
+    /// Fired when the timeline is paused.
+    pub on_pause: Rc<dyn Fn()>,
+
+    /// Fired when the timeline is played.
+    pub on_play: Rc<dyn Fn()>,
```
Wumpf, line 92: nice, very happy to see these all in on place now :)

crates/viewer/re_viewer/src/app.rs
```diff
@@ -83,12 +83,8 @@ pub struct StartupOptions {
     /// This also can be changed in the viewer's option menu.
     pub video_decoder_hw_acceleration: Option<re_video::decode::DecodeHardwareAcceleration>,
 
-    /// Interaction between JS and timeline.
-    ///
-    /// This field isn't used directly, but is propagated to all recording configs
-    /// when they are created.
-    #[cfg(target_arch = "wasm32")]
-    pub timeline_options: Option<crate::web::TimelineOptions>,
+    /// Interaction between JS and the viewer.
```
Wumpf, line 86: (you had a comment fix locally for this during our call today that hasn't been pushed yet)

crates/viewer/re_viewer/src/app.rs
```diff
@@ -232,11 +227,10 @@ pub struct App {
 
     reflection: re_types_core::reflection::Reflection,
 
-    /// Interaction between JS and timeline.
+    /// Interaction between JS and the viewer.
```
Wumpf, line 230: same comment fix as above applies here, talking about JS only here is _kinda_ wrong and definitely confusing :)

crates/viewer/re_viewer/src/web.rs
```diff
@@ -676,7 +683,49 @@ fn create_app(
         force_wgpu_backend: render_backend.clone(),
         video_decoder_hw_acceleration,
         hide_welcome_screen: hide_welcome_screen.unwrap_or(false),
-        timeline_options: timeline.clone(),
+
+        // CALLBACK DEFINITION
```
Wumpf, line 687: nit?: I understand why you introduced these markers and I don't have a better idea for x-language things that are hard to unit-test either. But from experience with projects that used those more extensively I think we should avoid them at least where things are statically checked. This is one such place (if you got the web.rs callback struct fine you're good!).

The reason for my hesistation is that they eventually become both infectious _and_ inconsistent: I've seen those move away from the relevant code over time due to happenstance, or being removed because the central explanation hasn't been found, hastening code rot. In other places people add them expecting that to be enough of an explanation despite making code just cryptic. And on top of that newcomers get a false sense of security "I searched for all the markers I should be fine" instead of stop-readthedocs-think-act.
(my 2 cents ;-))

Wumpf:
image
wheeeee so cool :)

rerun_notebook/callbacks.ipynb
```diff
@@ -0,0 +1,128 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8c7dedfe",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%env RERUN_NOTEBOOK_ASSET=inline"
```
Wumpf, line 20: Huh, this makes it nicer for development, I forgot that var and it just set it for me. Question: We can't really set this in an "offical" example like `example.ipynb` because that would mess with using that example with a public Rerun install, right? I.e. when we polish up this sample for public consumption this has to go?

Wumpf (approved):
Looking good!
Feel free to move interface things I pointed out to previously discussed follow-up (in that case plz don't resolve the affected comments so I can look them up later as a reminder)

crates/viewer/re_viewer_context/src/callback.rs
```diff
@@ -0,0 +1,120 @@
...
+#[derive(Debug, serde::Serialize)]
+#[serde(tag = "type")]
+#[serde(rename_all = "snake_case")]
+pub enum SelectionItem {
+    /// Selected a single entity.
+    ///
+    /// Examples:
+    /// * A full point cloud.
+    /// * A mesh.
+    Entity { entity_path: String },
```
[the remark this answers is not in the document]
jprochazk: It serializes as a `Vec<String>`, which is not great for usage from JS/Python, after it's deserialized from JSON. We could do some additional work there to join it into a string, but I'd rather keep the logic for stringifying entity paths on the Rust side.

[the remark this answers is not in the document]
Wumpf: makes sense 👍

pixi.lock
```diff
@@ -11415,6 +11415,19 @@ packages:
   - jinja2 ; extra == 'compiler'
```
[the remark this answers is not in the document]
jprochazk: I'm on pixi `0.41.4`, with merged `main`. Running `pixi run -e py py-build-notebook` does this. Really not sure why, but it doesn't seem harmful, at least.

crates/viewer/re_viewer/src/web.rs
jprochazk: It's just that there is no good way to statically check this everywhere :grimacing:. But I do think as long as whoever is adding or changing callbacks actually tests them everywhere we expose them, then it'll be difficult to miss any part of this, as they won't work at all otherwise.

I'll remove these markers.

rerun_notebook/callbacks.ipynb
jprochazk: https://github.com/rerun-io/rerun/issues/9274

crates/viewer/re_viewer_context/src/callback.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    /// Selected a single instance.
+    ///
+    /// Examples:
+    /// * A single point in a point cloud.
+    Instance {
+        entity_path: String,
+        instance_id: u64,
+    },
```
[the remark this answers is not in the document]
jprochazk: (will do in follow-up)

PR #9148, opened 2025-02-26
Add experimental `send_recording` python api 

### LINKED ISSUE #2044: `RecordingStream` should be able to forward an existing RRD file
This opens two use important use cases:
1. This would allow one to easily stream a data recording rrd file to a remote viewer, something I've wanted to be able to do for a while (e.g. `rerun my_rrd --connect`)
2. This would allow one to save blueprints as rrd and then send them back from their scripts

oxkitsune:
### Related

* Resolves #2044

### What

Adds api to send a recording loaded from RRD to a new recording stream, cloning the data in the process


rerun_py/src/python_bridge.rs
```diff
@@ -79,7 +82,7 @@ type GarbageReceiver = crossbeam::channel::Receiver<ArrowRecordBatch>;
 /// accumulated data for real.
 //
 // NOTE: `crossbeam` rather than `std` because we need a `Send` & `Sync` receiver.
-static GARBAGE_QUEUE: Lazy<(GarbageSender, GarbageReceiver)> =
+pub(crate) static GARBAGE_QUEUE: Lazy<(GarbageSender, GarbageReceiver)> =
```
Wumpf, line 85: nit: here and below, why not just `pub` given that `lib.rs` doesn't do blanket re-export

rerun_py/rerun_sdk/rerun/sinks.py
```diff
@@ -472,6 +473,31 @@ def send_blueprint(
     )
 
 
+def send_recording(recording: Recording, stream: RecordingStream | None = None) -> None:
```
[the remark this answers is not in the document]
Wumpf: Also, we need a snippet using this.
Ideally we'd add it to all languages if only to have the snippet comparision test do something - I figure at least for Rust we could whip up something?

rerun_py/src/python_bridge.rs
```diff
@@ -1275,6 +1279,22 @@ fn send_blueprint(
...
+/// Send a recording to the given recording stream.
+// TODO(gijsd): Figure out naming for the recording stream parameter, it's called `recording` everywhere else.
+#[pyfunction]
+#[pyo3(signature = (recording, stream = None))]
+fn send_recording(recording: &PyRecording, stream: Option<&PyRecordingStream>) {
+    let Some(stream) = get_data_recording(stream) else {
+        return;
+    };
+
+    // TODO(gijsd): Overwrite store info here, since we want to replace it with the info from the recording we're sending
```
Wumpf, line 1291: we should document status quo in the api

rerun_py/src/dataframe.rs
```diff
@@ -1382,6 +1383,36 @@ impl PyRecording {
...
+            GARBAGE_QUEUE.0.send(chunk).ok();
+        };
+        batcher_config.hooks.on_release = Some(on_release.into());
+
+        let application_id = self.application_id()?;
+        let recording_id = self.store.read().id();
+
+        let recording = RecordingStreamBuilder::new(application_id)
+            .batcher_config(batcher_config)
+            .is_official_example(false)
```
Wumpf, line 1399: leaving that to default does the usual detection of when a thing is an example, that should be fine
```suggestion
```

rerun_py/src/dataframe.rs
```diff
@@ -1382,6 +1383,36 @@ impl PyRecording {
...
+            .store_id(recording_id.clone())
+            .store_source(re_log_types::StoreSource::PythonSdk(python_version(py)))
+            .default_enabled(true)
+            .buffered()
+            .map_err(|err| PyRuntimeError::new_err(err.to_string()))?;
+
+        all_recordings().insert(recording_id.clone(), recording.clone());
+
+        for chunk in self.store.read().iter_chunks() {
+            let msg = chunk.to_arrow_msg().expect("oops");
```
Wumpf, line 1409: should never happen but needs proper error handling

rerun_py/src/dataframe.rs
```diff
@@ -1382,6 +1383,36 @@ impl PyRecording {
             .as_str()
             .to_owned())
     }
+
+    fn to_stream(&self, py: Python<'_>) -> PyResult<crate::python_bridge::PyRecordingStream> {
```
Wumpf, line 1387: Needs docs & snippet

Wumpf (changes requested):
Independent on whether we want to solve this differently in the future, the APIs are very valuable to have I believe.
There's some todos left though for documenting this out etc.
Also, I don't think we should close the linked issue until we implement this in all SDKs

rerun_py/src/python_bridge.rs
oxkitsune: i don't need to re-export them even, for the original implementation done at the kitchen table I had to, but not anymore 🕺

rerun_py/rerun_sdk/rerun/sinks.py
teh-cmc, line 476: Do we need this Python function at all? Can this logic be done in the Rust implementation instead? @jleibs has recently implemented a bunch of APIs for the Python SDK that don't have a single line of Python in them (AFAIK?), and I would very much like to keep that trend...

oxkitsune: I looked into it, but it doesn’t seem super straightforward in this case. There are a few things that make it a bit tricky, e.g. `RecordingStream` wrapping `PyRecordingStream`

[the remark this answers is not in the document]
oxkitsune: I've added snippets for both rust/py 👍

abey79: Yes. This can only be done correctly once everything else is done correctly. By "correctly", I mean according to this issue:
- https://github.com/rerun-io/rerun/issues/9187

abey79 (changes requested):

rerun_py/rerun_sdk/rerun/sinks.py
```diff
@@ -488,6 +490,31 @@ def send_blueprint(
     )
 
 
+def send_recording(embedded_recording: Recording, recording: RecordingStream | None = None) -> None:
```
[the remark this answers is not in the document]
abey79: This is relevant: https://github.com/rerun-io/rerun/issues/9187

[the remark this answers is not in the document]
jleibs: I'm late to the party on this one, but we should really think about the flow between the `Recording` and a `RecordingStream` object a bit more.

For example: the `MemoryRecording` would ideally go away and just be a `Recording`. Would be so nice to be able to log to the in-memory recording and then query it without needing to re-load it from an rrd.

Wumpf (approved):
.rrd isn't checked in as a LFS file, should be though

otherwise (and remaining comment) looks all good to me now. I'd be in favor of landing this prior to additional Python refactors, but I'll let you and @abey79 figure that out ;-)

Wumpf (changes requested):
`examples/assets/example.rrd` is already checked in as an lfs file by now, this overrides it with an regular binary

Wumpf (approved):

abey79 (approved):
nice

PR #9149, opened 2025-02-26
Fix multi-value `Scalar` breaking legend disambiguation

abey79:
### Related

* #7140
* #9033

### What

The recently introduced support for multi-value `Scalar` batch in the time-series view broke the way legend items are disambiguated, e.g. when several series share the same last entity part (see #7140). The reason for that is that now the same entity path might be passed multiple time to the utility that actually performs the disambiguation. This PR fixes this issue by ensuring that such duplication doesn't happen.

Before:
image

After:
image

### DIFF
--- a/crates/viewer/re_view_time_series/src/view_class.rs
+++ b/crates/viewer/re_view_time_series/src/view_class.rs
@@ -480,7 +480,9 @@ impl ViewClass for TimeSeriesView {
             state.default_names_for_entities = EntityPath::short_names_with_disambiguation(
                 all_plot_series
                     .iter()
-                    .map(|series| series.instance_path.entity_path.clone()),
+                    .map(|series| series.instance_path.entity_path.clone())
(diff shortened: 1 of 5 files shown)


crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -480,7 +480,9 @@ impl ViewClass for TimeSeriesView {
             state.default_names_for_entities = EntityPath::short_names_with_disambiguation(
                 all_plot_series
                     .iter()
-                    .map(|series| series.instance_path.entity_path.clone()),
+                    .map(|series| series.instance_path.entity_path.clone())
+                    // `short_names_with_disambiguation` expects no duplicate entities
+                    .collect::<nohash_hasher::IntSet<_>>(),
```
Wumpf, line 485: you could also just call `unique` on it I figure 🤷

Wumpf (approved):
magic 😄

PR #9150, opened 2025-02-26
Switch from `cargo-machete` to `cargo-shear`

oxkitsune:
### What

Replaces `cargo-machete` with [cargo-shear](https://github.com/Boshen/cargo-shear) in our CI, as `cargo-machete` had a lot of false negatives.


abey79 (approved):
I was missing some motivation for the change in the PR description. I found it looking at the diff. Crazy how many false negatives we had!

Issue #8557, opened 2024-12-20
Can't set time range on individual entities via blueprint api


Wumpf:
Adding this to the helix demo doesn't have any effect right now:
```py
    blueprint = rrb.Blueprint(
        rrb.Spatial3DView(
            origin="/",
            overrides={
                "helix/structure/scaffolding/beads": [
                    rrb.VisibleTimeRange(
                        "stable_time",
                        start=rrb.TimeRangeBoundary.cursor_relative(seconds=-0.3),
                        end=rrb.TimeRangeBoundary.cursor_relative(seconds=0.3),
                    ),
                ]
            },
        ),
    )
    rr.send_blueprint(blueprint)
```

We store it already in the blueprint but it's stored on "recursive override path" which is a vestigue from back when we tried a model of "indidual & recursive" overrides. We dumped that idea and instead want to make overrides be path-expression enabled in the future. Individua/recursive is really only used for recursive visibility & interactivity today and we should clean up everything else. See:
* https://github.com/rerun-io/rerun/issues/6673

Python blueprint api overrides use only the "individual" override path.

-> The fix MIGHT be as easy to change these two code paths to use "individual" instead. But a bit more looking into is needed
* https://github.com/rerun-io/rerun/blob/19e93a63e56e2a96054b6e899612640c7ea558f1/crates/viewer/re_viewport_blueprint/src/view_contents.rs#L541
* https://github.com/rerun-io/rerun/blob/19e93a63e56e2a96054b6e899612640c7ea558f1/crates/viewer/re_selection_panel/src/visible_time_range_ui.rs#L47

We also should really add a test for this - at least a release_checklist one

PR #9151, opened 2025-02-26
Fix setting visible time range from overrides

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/8557

### What

Overrides from blueprint do what they're supposed to again (now? not sure if this ever worked correctly :/)
image

Todo:
* [ ] port `rerun\tests\python\release_checklist\check_time_range_individual_overrides.py` to a screenshot test and test this
    * @teh-cmc didn't you already start this one and didn't finish it because of the bug fixed here?

### DIFF
--- a/crates/viewer/re_selection_panel/src/visible_time_range_ui.rs
+++ b/crates/viewer/re_selection_panel/src/visible_time_range_ui.rs
@@ -44,12 +44,10 @@ pub fn visible_time_range_ui_for_data_result(
     ui: &mut Ui,
     data_result: &re_viewer_context::DataResult,
 ) {
-    // TODO(#8557): Almost certainly this is wrong and should be regular override path.
-    let override_path = data_result.recursive_override_path();
     let query_range = data_result.property_overrides.query_range.clone();
 
     let is_view = false;
-    visible_time_range_ui(ctx, ui, query_range, override_path, is_view);
+    visible_time_range_ui(ctx, ui, query_range, data_result.override_path(), is_view);
 }
 
 /// Draws ui for a visible time range from a given override path and a resulting query range.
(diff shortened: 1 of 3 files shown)


crates/viewer/re_view_spatial/tests/snapshots/intra_timestamp/46_46.png
teh-cmc, line 1: Wait, where's the second pink point?

crates/viewer/re_view_spatial/tests/visible_time_range.rs
```diff
@@ -344,28 +329,42 @@ fn setup_blueprint(
...
-                .join(&EntityPath::from("ViewContents/recursive_overrides/green"));
-
-            ctx.save_blueprint_component(&property_path, &visible_time_range_list);
+            let visible_time_range_list =
+                re_types::blueprint::archetypes::VisibleTimeRanges::new([green_time_range]);
+            ctx.save_blueprint_archetype(
+                &re_viewport_blueprint::ViewContents::override_path_for_entity(
+                    view_id,
+                    &"green".into(),
+                ),
```
teh-cmc, line 365: So, this tests the individual path, right? Do we have any test for the recursive path?

teh-cmc (approved):
Neat. I'd sleep better if we covered both individual and recursive in our test though (it might already be the case and I just missed it).

Wumpf:
`recursive` is fully defunct for everything but `visible` right now and we don't have it exposed to users

crates/viewer/re_view_spatial/tests/visible_time_range.rs
Wumpf: as per comment below. No, because we don't expose it either. It's a remnant of a discarded design and only used for `visible` on some ui interactions

crates/viewer/re_view_spatial/tests/snapshots/intra_timestamp/46_46.png
Wumpf: well spotted, what did I mess up here 🤔 🔍

Wumpf: a point was misplaced, fixed!

PR #9152, opened 2025-02-27
Fix shader compiluation failure in Chrome 131

### LINKED ISSUE #8377: Error with webviewer
**Describe the bug**
Got error with web viewer:
LinkError: WebAssembly.instantiate(): Import #10 "wbg" "__wbg_structuredClone_046ae3829d60ba4b": function import requires a callable

**To Reproduce**
Built the web viewer with 0.20.3 and tried to run it on of my .rrd files.

**Rerun version**
0.20.3

yakunouyang:
### Related

* Mentioned in #8377

### What

https://github.com/rerun-io/rerun/blob/af3312a21fb157b76dce2c509a1934ad252d1b67/crates/viewer/re_renderer/shader/point_cloud.wgsl#L134

The variable `distance` collide with internal function `distance()`, resulting in wgsl build error on macOS Chrome.

More information in #8377.

Image

### DIFF
--- a/crates/viewer/re_renderer/shader/point_cloud.wgsl
+++ b/crates/viewer/re_renderer/shader/point_cloud.wgsl
@@ -131,9 +131,9 @@ fn vs_main(@builtin(vertex_index) vertex_idx: u32) -> VertexOut {
 /// 2D primitives are always facing the camera - the difference to sphere_quad_coverage is that
 /// perspective projection is not taken into account.
 fn circle_quad_coverage(world_position: vec3f, radius: f32, circle_center: vec3f) -> f32 {
-    let distance = distance(circle_center, world_position);
-    let feathering_radius = fwidth(distance) * 0.5;
-    return smoothstep(radius + feathering_radius, radius - feathering_radius, distance);
+    let circle_distance = distance(circle_center, world_position);
+    let feathering_radius = fwidth(circle_distance) * 0.5;
+    return smoothstep(radius + feathering_radius, radius - feathering_radius, circle_distance);
 }
 
 fn coverage(world_position: vec3f, radius: f32, point_center: vec3f) -> f32 {


emilk (approved):
Thanks!

Wumpf:
does not fix the linked issue, that was originally about something else entirely

Wumpf:
Fix can't hurt, but fwiw I can't repro this on MacOs M1 Chrome `133.0.6943.142` which is surprising since this is almost the exact setup you mentioned in https://github.com/rerun-io/rerun/issues/8377#issuecomment-2687046499
Need to check whether Chrome/Tint or Wgpu/Naga is in spec violation here.

yakunouyang:
> Fix can't hurt, but fwiw I can't repro this on MacOs M1 Chrome `133.0.6943.142` which is surprising since this is almost the exact setup you mentioned in [#8377 (comment)](https://github.com/rerun-io/rerun/issues/8377#issuecomment-2687046499) Need to check whether Chrome/Tint or Wgpu/Naga is in spec violation here.

After some digging, I understand what happened: when I tried to add flag `--enable-dawn-features=dump_shaders,disable_symbol_renaming` to the chrome browser, it auto updates and take me to the `133.0.6943.142` version. I found the version before update ([131.0.6778.0](http://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Mac_Arm/1368535/)), and this bug reproduces, sorry for the confusion.

As for the specs, this issue may help https://github.com/gpuweb/gpuweb/issues/2619, the latest Chrome/Tint accepts such syntax, and the tint ir dumped by latest Chrome also fixes this problem:
image

That being said, I think this bug still relevant, it affects the older browsers, and Metal Shader Language clearly doesn't accept such syntax.

PR #9153, opened 2025-02-27
Show EXIF data for JPEG/TIFF images

emilk:
When selecting an JPEG or TIFF image, we will show its EXIF contents in the selection panel

Screenshot 2025-02-27 at 11 15 46


abey79 (approved):
nice

PR #9157, opened 2025-02-27
Better support for catalog on web

jprochazk:
- `?url` now supports catalog
- Also makes `AddRedapServer` command show the server browser


abey79 (approved):
Nice!

Issue #8635, opened 2025-01-10
Add a timeline type for relative time


emilk:
We currently have two time types:
* Sequence
* Temporal

The problem is that the temporal type is used both for absolute times (2024-01-10 09:44:01Z) and relative times (`+32m 12s`).
Because of this problem, relative times will show up as `1970-01-01 00:32:12Z` in arrow formatting.

Relative times are useful for "time since robot started" or whatever.

I suggest we split `Temporal` into absolute and relative. The arrow datatype would then become:

* `Sequence` - > `DataType::Int64`
* `Duration` - > `DataType::Duration64(TimeUnit::Nanosecond)`
* `Timestamp` - > `DataType::Timestamp(TimeUnit::Nanosecond, None)`

### Should do this first
* https://github.com/rerun-io/rerun/issues/9084

### Related
* https://github.com/rerun-io/rerun/issues/8471
* https://github.com/rerun-io/rerun/issues/9082

### Future work
* Adding even more types of indices, e.g. UUID

## TODO
* [x] Introduce new SDK APIs (see below)
   * [x] Python
     * https://github.com/rerun-io/rerun/pull/9166
     * https://github.com/rerun-io/rerun/pull/9179
     * https://github.com/rerun-io/rerun/pull/9191
	 * https://github.com/rerun-io/rerun/pull/9192
     * https://github.com/rerun-io/rerun/pull/9342
   * https://github.com/rerun-io/rerun/pull/9200
   * [x] Rust
     * https://github.com/rerun-io/rerun/pull/9226
     * https://github.com/rerun-io/rerun/pull/9236
* https://github.com/rerun-io/rerun/pull/9292
* [x] Remove `struct Time`
* [x] Fix all `TODO(#8635)`:s
* [ ] Unify naming
	* [x] Use the same names (`sequence, duration, timestamp`) in all language APIs
	* https://github.com/rerun-io/rerun/pull/9343
	* https://github.com/rerun-io/rerun/pull/9343
	* [x] Revert `index` related names
	  * [x] Rename `IndexCell` -> `TimeCell`
	  * [x] Rename `set_index` -> `set_time`
	  * [x] Rename `nonstatic` -> `temporal`

## Future work
* Rename "timeline" to "index" everywhere
  * Rename `disable_timeline` and `reset_time`

abey79:
Here is a fleshed out version of the proposed Python API that annotates the single-non-`None` arg restriction:

```python
from __future__ import annotations

from typing import overload

[...]

PR #9166, opened 2025-03-03
Add `rr.set_index` to replace `rr.set_time_*`

emilk:
### Related

### Migration / what
See [the new migration guide](https://github.com/rerun-io/rerun/blob/emilk/new-python-index-api/docs/content/reference/migration/migration-0-23.md#-python-replaced-rrset_time_-with-rrset_index)

### TODO
* [x] Replace all uses of the old functions with the new
* [x] Deprecate the old functions
* [x] Add entry to migration guide
* [x] Add a new snippet
* [x] Full check

### Future work
* Replace `TimeSequenceColumn/TimeSecondsColumn/TimeNanosColumn` with `IndexColumn("frame_nr", sequence=…), IndexColumn("capture_time", timestamp=…)`
* C++ and Rust
* Having absolute times and relative times actually work differently
* Rename the other time-related functions?


rerun_py/rerun_sdk/rerun/time.py
```diff
@@ -1,12 +1,137 @@
...
+            date_time = date_time.astimezone(timezone.utc)
+        epoch = datetime(1970, 1, 1, tzinfo=timezone.utc)
+        return int((date_time - epoch).total_seconds() * 1e9)
+    elif isinstance(date_time, np.datetime64):
+        return date_time.astype("int64")  # type: ignore[no-any-return]
+    else:
+        raise TypeError("date_time must be an int, float, datetime, or numpy.datetime64 object")
+
+
+# TODO(#8635): deprecate
```
abey79, line 134: Any reason not to do that directly?


```suggestion
@deprecated(
    """Use `set_index()` instead.
    See: https://www.rerun.io/docs/reference/migration/migration-0-23?speculative-link for more details."""
)
```

abey79: May be worth considering converting them all in this PR? Doesn't sound tooo bad.

Also, ideally, I'd rather have the test converted in this PR, and a new test added to cover for illegal calling patterns.

abey79: (happy to help with that)

emilk: Yeah, feel free to push some commits! 🙏  Let me know what you wanna work on though so we don't overlap

> May be worth considering converting them all in this PR?
Yeah, it's in the TODO of the PR description

emilk:
@rerun-bot full-check

docs/content/reference/migration/migration-0-23.md
```diff
@@ -7,3 +7,47 @@ order: 989
...
+
+`timedelta` must be either:
+* seconds as `int` or `float`
+* `datetime.timedelta`
+* `numpy.timedelta64`.
+
+`datetime` must be either:
+* seconds since unix epoch (1970-01-01) as `int` or `float`
+* `datetime.datetime`
+* `numpy.datetime64`.
```
abey79, line 27:
```suggestion
`timedelta` must be either:
* seconds as `int` or `float`
* [`datetime.timedelta`](https://docs.python.org/3/library/datetime.html#datetime.timedelta)
* [`numpy.timedelta64`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.timedelta64)

`datetime` must be either:
* seconds since unix epoch (1970-01-01) as `int` or `float`
* [`datetime.datetime`](https://docs.python.org/3/library/datetime.html#datetime.datetime)
* [`numpy.datetime64`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.datetime64)
```

docs/content/reference/migration/migration-0-23.md
```diff
@@ -7,3 +7,47 @@ order: 989
...
+##### `rr.set_sequence("foo", 42)`
+New: `rr.set_index("foo", sequence=42)`
+
+##### `rr.set_time_seconds("foo", duration_seconds)`
+When using relative times (durations/timedeltas): `rr.set_index("foo", timedelta=duration_seconds)`
+You can also pass in a `datetime.timedelta` or `numpy.timedelta64` directly.
+
+##### `rr.set_time_seconds("foo", seconds_since_epoch)`
+New: `rr.set_index("foo", datetime=seconds_since_epoch)`
+You can also pass in a `datetime.datetime` or `numpy.datetime64` directly.
```
abey79, line 39:
```suggestion
##### `rr.set_time_seconds("foo", duration_seconds)`
When using relative times (durations/timedeltas): `rr.set_index("foo", timedelta=duration_seconds)`
You can also pass in a [`datetime.timedelta`](https://docs.python.org/3/library/datetime.html#datetime.timedelta) or [`numpy.timedelta64`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.timedelta64) directly.

##### `rr.set_time_seconds("foo", seconds_since_epoch)`
New: `rr.set_index("foo", datetime=seconds_since_epoch)`
You can also pass in a [`datetime.datetime`](https://docs.python.org/3/library/datetime.html#datetime.datetime) or [`numpy.datetime64`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.datetime64) directly.
```

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

docs/snippets/compare_snippet_output.py
```diff
@@ -64,7 +64,7 @@ def __lt__(self, other: Example) -> bool:
         return self.name < other.name
 
     def __repr__(self) -> str:
-        return f"Example(subdir={self.subdir}, name={self.name})"
+        return f"{self.subdir}/{self.name}"
```
abey79, line 67: nit/psa: in the python canon, `__repr__` should ideally return a string that is valid python and interprets as the original object. So the previous version was more correct (though python nerds would rather have `self.__class__.__name__` instead of duplicating the class name). Your new implementation is a better fit for `__str__`, which is the user facing representation of the underlying data (similar to rust's `Display`). In this case, I don't really care, but it is very nice to have a correct repr for any object that has a chance to be used by users in a notebook/REPL context.

abey79: Actually, the correct way is to make `Example` a `@dataclass.dataclass` and all this boilerplate will be automagically generated, including `__init__`, `__eq__`, `__lt__` (though not with the current implementation), `__repr__`, and possibly `__hash__` and `__str__`.

docs/snippets/compare_snippet_output.py
```diff
@@ -286,7 +298,7 @@ def check_non_empty_rrd(path: str) -> None:
     from pathlib import Path
 
     assert Path(path).stat().st_size > 0
-    print(f"Confirmed output written to {Path(path).absolute()}")
+    # print(f"Confirmed output written to {Path(path).absolute()}")
```
abey79, line 301: dead code?

rerun_py/docs/gen_common_index.py
```diff
@@ -108,6 +108,7 @@ class Section:
     Section(
         title="Timeline functions",
         func_list=[
+            "set_index",
             "set_time_sequence",
             "set_time_seconds",
             "set_time_nanos",
```
abey79, line 114: Shouldn't these be removed from the documentation?

rerun_py/rerun_sdk/rerun/__init__.py
```diff
@@ -226,6 +227,7 @@ def _init_recording_stream() -> None:
             notebook_show,
         ]
         + [
+            set_index,
             set_time_sequence,
             set_time_seconds,
             set_time_nanos,
```
abey79, line 233: This makes me think that it might be worth opening a reminder "remove deprecated API" issue scheduled for 0.24 or 0.25, and ideally tag these places (though a global search will probably good enough when we actually do it).

abey79 (approved):
Great stuff. This API is much better imo

docs/snippets/compare_snippet_output.py
emilk: I think I wanted to override `__str__` instead of `__repr__`.

docs/snippets/compare_snippet_output.py
emilk: Yes, but may be useful when debugging later

rerun_py/docs/gen_common_index.py
emilk: I did, but then I got confusing errors 😞

PR #9172, opened 2025-03-03
Copy description of selection (entity path, store id, etc.) via `cmd/ctrl + c`

Wumpf:
### Related

* follow-up to https://github.com/rerun-io/rerun/pull/9137

### What

https://github.com/user-attachments/assets/6399a8f0-3e05-400b-8681-d194ab48d537

For various selections you can now press `ctrl+c` to copy what they're pointing to. This is most useful for entity paths but also allows to copy things like store ids.
The system automatically backs off if any text is selected anywhere in the viewer.


abey79 (approved):
very nice!

Issue #7653, opened 2024-10-09
Ability to jump to a specific frame/timestamp in the timeline view


teh-cmc:
I'm trying to jump to frame `#15000`. The frame label should be editable and let me enter the exact value I want to go to.
![image](https://github.com/user-attachments/assets/bfe343d7-d219-4eaf-a3a9-ae1007708968)

Existing workaround:
1. use the time cursor to get close to the target
2. use arrow keys to reach the target exactly

PR #9174, opened 2025-03-03
Editable time control & time context menus

Wumpf:
### Related

* Fixes [#7653](https://github.com/rerun-io/rerun/issues/7653)

### What

https://github.com/user-attachments/assets/9ea1e396-6ad7-4a27-9f5e-648903eafaee

Time in the timeline panel is now an input field that supports
* raw timestamps
* relative times
* date times

Edits are applied when the field looses its focus (you can also just press `enter` for that).

Furthermore, there's now context menus to copy timeline name & raw timestamps now.

Previously, the time panel was dragable in some situations (time indices not exceeding a certain numerical range). We could bring that back, but I decided to drop it in favor of a simple edit field - dragging would have made this a lot more complicated.


crates/store/re_log_types/src/time_point/mod.rs
```diff
@@ -182,6 +182,35 @@ impl TimeType {
         }
     }
 
+    /// Parses a human-readable time string into a [`TimeInt`].
+    pub fn parse_time(&self, s: &str, time_zone_for_timestamps: TimeZone) -> Option<TimeInt> {
+        match s {
+            "<static>" => Some(TimeInt::STATIC),
+            "−∞" => Some(TimeInt::MIN),
+            "∞" | "+∞" => Some(TimeInt::MAX),
```
abey79, line 190: worth adding `"inf"` here?

abey79: (and, if so, a `.to_lowercase()` somewhere)

abey79 (changes requested):
Great change and lgtm! But... did it break collapsed time panel?

Here is what I see with the web build:

image

vs. the [latest build from main](https://rerun.io/viewer/version/main?url=https%3A%2F%2Fapp.rerun.io%2Fversion%2Fnightly%2Fexamples%2Fdna.rrd):

image

Wumpf:
woah. Totally missed that!

abey79 (approved):
Thanks!

Wumpf:
need to figure out what happened to the screenshot margins

Wumpf:
nooooooo was too slow with merging this

PR #9178, opened 2025-03-04
gRPC spec update: Fetch range of chunks

zehiko:
### What

With the introduction of manifest, ``GetChunkIds`` is no longer needed, hence we get rid of it. We introduce a new call for fetching range of chunks that basically merged query manifest + get chunks into a single call.

In order to support fetching chunk ranges directly from disk we add a small utility method for reading ``.rrd`` file header asynchronously. (I've added a sync counterpart, although it's currently unused)

### Testing done
e2e integration


crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -77,24 +77,20 @@ message DataframePart {
...
+    // which catalog entry do we want to fetch the chunks from
     CatalogEntry entry = 1;
-    // recording id from which we're want to fetch the chunk ids
+    // recording id from which we're want to fetch the chunk
     rerun.common.v0.RecordingId recording_id = 2;
     // timeline for which we specify the time range
     rerun.common.v0.IndexColumnSelector time_index = 3;
-    // time range for which we want to fetch the chunk ids
+    // time range for which we want to fetch the chunks
     rerun.common.v0.TimeRange time_range = 4;
```
jleibs, line 91: Nit: I know there's a big time -> index rename happening. Is this an opportunity to align naming?

zehiko: sure, will update it, tnx.

crates/store/re_protos/proto/rerun/v0/remote_store.proto
```diff
@@ -23,10 +23,10 @@ service StorageNode {
 
     rpc ReIndex(ReIndexRequest) returns (ReIndexResponse) {}
 
-    rpc GetChunkIds(GetChunkIdsRequest) returns (stream GetChunkIdsResponse) {}
-
     rpc GetChunks(GetChunksRequest) returns (stream rerun.common.v0.RerunChunk) {}
 
+    rpc GetChunksRange(GetChunksRangeRequest) returns (stream rerun.common.v0.RerunChunk) {}
```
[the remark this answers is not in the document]
teh-cmc: Generally agree with the direction that @zeljko just described.

Because things are still in a pretty half-assed overall state, I agree it's quite hard to see where we're going though...

Here's the explicit vision I am aiming for (in pseudo-protobuf to make it clearer):
```proto
/// Self-explanatory.
///
/// (In the real world, the request part would be its own dedicated message type).
rpc GetChunksById(ids: [RerunChunkId]) returns (stream RerunChunk) {}

/// Queries the manifest and returns a bunch of chunk-level metadata, including but not limited
/// to Chunk IDs.
///
/// (In the real world, the request part would be its own dedicated message type).
rpc QueryManifest(query: ManifestQuery) returns (stream ManifestChunk) {}

/// This is a helper that combines `GetChunksById` and `QueryManifest`, because that can be done
/// both much more conveniently and more performant by staying on the server-side.
///
/// Clients can achieve the same by manually calling `GetChunksById` and `QueryManifest` if
/// they so wish.
///
/// (In the real world, the request part would be its own dedicated message type).
rpc GetRelevantChunks(query: ManifestQuery) returns (stream RerunChunk) {}

/// All fields can be specified at the same time. For example, it is pretty common to want both
/// a LatestAt and a Range.
message ManifestQuery {
    QueryManifestLatestAtRelevantChunks latest_at = 1;
    QueryManifestRangeRelevantChunks range = 2;
    QueryManifestDataframeRelevantChunks range = 2;
    ScanParameters scan_parameters = 4;
}
```

I'll make a dedicated issue explaining this plan so people can express how wrong I am 🙃

teh-cmc (approved):
I'll open an issue describing what I think is our complete vision for these chunk fetching APIs (see my comment above).

In the meantime, this pushes us into the right direction.

teh-cmc:
Also, reminder that the manifest doesn't even support latest-at right now, so the data returned by this is only partially complete.

I still think we should go forward with this. If anything this gives me all the more reason to prioritize finishing server-side queries.

teh-cmc:
> Also, reminder that the manifest doesn't even support latest-at right now, so the data returned by this is only partially complete.
> 
> I still think we should go forward with this. If anything this gives me all the more reason to prioritize finishing server-side queries.

Corollary: we should probably leave the old endpoints in for now, if we have any customers that rely on them (do we?). At least until we make these new endpoints correct.

zehiko:
> > Also, reminder that the manifest doesn't even support latest-at right now, so the data returned by this is only partially complete.
> > I still think we should go forward with this. If anything this gives me all the more reason to prioritize finishing server-side queries.
> 
> Corollary: we should probably leave the old endpoints in for now, if we have any customers that rely on them (do we?). At least until we make these new endpoints correct.

I'm quite sure there's no active users of "fetch range" endpoints today and the extra logic that would be kept on server side makes me think this is not worth it. I can verify with the customers before proceeding with this update. If anything, it will be simpler for them to use the new single endpoint as opposed to 2 calls.

crates/store/re_protos/proto/rerun/v0/remote_store.proto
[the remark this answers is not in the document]
teh-cmc: * Added some notes for a discussion over at https://github.com/rerun-io/rerun/issues/9198

crates/store/re_protos/proto/rerun/v0/remote_store.proto
zehiko: actually, decided to follow up on this one, I need to clarify with Emil few things.

PR #9182, opened 2025-03-04
Python SDK spring cleaning: 3.9, no more monkey patching, more lints

### LINKED ISSUE #5445: RecordingStream monkey-patched functions result in mypy errors
We money-patch a bunch of helpers into the RecordingStream class, but in practice using these causes mypy to produce a bunch of errors.

For example
```
rerun_py/rerun_sdk/rerun/blueprint/api.py:77: error: "RecordingStream" has no attribute "log"  [attr-defined]
rerun_py/rerun_sdk/rerun/blueprint/api.py:84: error: "RecordingStream" has no attribute "log"  [attr-defined]
rerun_py/rerun_sdk/rerun/blueprint/api.py:184: error: "RecordingStream" has no attribute "log"  [attr-defined]
rerun_py/rerun_sdk/rerun/blueprint/api.py:292: error: "RecordingStream" has no attribute "log"  [attr-defined]
rerun_py/rerun_sdk/rerun/blueprint/api.py:314: error: "RecordingStream" has no attribute "set_time_seconds"  [attr-defined]
rerun_py/rerun_sdk/rerun/blueprint/api.py:318: error: Returning Any from function declared to return "MemoryRecording"  [no-any-return]
rerun_py/rerun_sdk/rerun/blueprint/api.py:318: error: "RecordingStream" has no attribute "memory_recording"  [attr-defined]
```

Now that we've introduced code-gen into other parts of our code-base, we should consider just code-genning these helpers to cut down on runtime shenanigans and bring us proper type signatures.

abey79:
This PR cleans up multiple things:
- Enforce Python 3.9 for the SDK (3.8 was deprecated in 0.20)
- Remove the nasty monkey patching of stuff in `RecordingStream` (mostly by resolving the circular import issues)
- Fix the use of mutable argument default in the `View` class hierarchy
- Removed as many `# type: ignore[....]` comment as possible
  - Unfortunately, I cannot enable the `expect`-like behaviour of mypy without make the codegen way too smart about adding these comments (or fixing the underlying logic to avoid needing these comment in the first place).
- 	Enable the `B` class of lint for Ruff and fix what it caught

Commit-by-commit review suggested:
- First commit has the bulk of the interesting `RecordingStream` stuff
- [791c685](https://github.com/rerun-io/rerun/pull/9182/commits/791c6853a12bc7ed33cd6ae35dbdc72dc5b22ea6) worth taking a look to as manual work is going there
- last two commits are generated from tooling (especially https://github.com/rerun-io/rerun/pull/9182/commits/510c83fa10bc4eb7e807e73e48c4a948aa000ce8) so best filter out that noise

Related:
- Fixes #5445


rerun_py/rerun_sdk/rerun/_unions.py
```diff
@@ -14,25 +14,22 @@ def build_dense_union(data_type: pa.DenseUnionType, discriminant: str, child: pa
...
-
-        return pa.Array.from_buffers(
-            type=data_type,
-            length=len(child),
-            buffers=[None, type_ids.buffers()[1], value_offsets.buffers()[1]],
-            children=children,
-        )
-
-    except ValueError as e:
-        raise ValueError(e.args)
```
abey79, line 38: this whole diff is weird but i basically just removed that weird catch/raise thing

rerun_py/rerun_sdk/rerun/recording_stream.py
abey79, line 1: the interesting stuff is here, which is why github collapses it by default

examples/python/detect_and_track_objects/detect_and_track_objects.py
```diff
@@ -7,9 +7,10 @@
 import json
 import logging
 import os
+from collections.abc import Sequence
```
emilk, line 10: why are some of these moved from `typing` to `collection.abc`? Is there a lint to catch the wrong usage?

examples/python/rgbd/rgbd.py
```diff
@@ -141,13 +141,16 @@ def download_progress(url: str, dst: Path) -> None:
...
+    with (
+        open(dst, "wb") as file,
+        tqdm(
+            desc=dst.name,
+            total=total,
+            unit="iB",
+            unit_scale=True,
+            unit_divisor=1024,
+        ) as bar,
+    ):
```
emilk, line 153: what are the extra () good for ?

rerun_py/pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
 description = "The Rerun Logging SDK"
 keywords = ["computer-vision", "logging", "rerun"]
 name = "rerun-sdk"
-requires-python = ">=3.8"
+requires-python = ">=3.9"
```
emilk, line 25: nice

rerun_py/rerun_sdk/rerun/__init__.py
```diff
@@ -12,18 +11,13 @@
 __version__ = "0.23.0-alpha.1+dev"
 __version_info__ = (0, 23, 0, "alpha.1")
 
-if sys.version_info < (3, 9):
-    warnings.warn(
-        "Python 3.8 is past EOL (https://devguide.python.org/versions/). Rerun version 0.21 will drop support/testing of Python 3.8.",
-        DeprecationWarning,
-    )
```
emilk, line 19: Wat happens if someone tries to run Rerun with python 3.8 now?  Should we still print a warning?

rerun_py/rerun_sdk/rerun/archetypes/depth_image_ext.py
```diff
@@ -115,7 +115,7 @@ def __init__(
         try:
             datatype = ChannelDatatype.from_np_dtype(image.dtype)
         except KeyError:
-            raise ValueError(f"Unsupported dtype {image.dtype} for DepthImage")
+            raise ValueError(f"Unsupported dtype {image.dtype} for DepthImage") from None
```
emilk, line 118: What does `raise X from None` mean? O.o

rerun_py/rerun_sdk/rerun/blueprint/api.py
```diff
@@ -88,9 +89,9 @@ def __init__(
         self.origin = origin
         self.contents = contents
         self.visible = visible
-        self.properties = properties
-        self.defaults = defaults
-        self.overrides = overrides
+        self.properties = properties if properties is not None else {}
+        self.defaults = defaults if defaults is not None else []
+        self.overrides = overrides if overrides is not None else {}
```
emilk, line 94: Sometimes I wish I was writing Lua and could just write `overrides or {}` instead of all this verbosity…

rerun_py/rerun_sdk/rerun/blueprint/api.py
```diff
@@ -129,22 +130,23 @@ def _log_to_stream(self, stream: RecordingStream) -> None:
...
             else:
                 raise ValueError(f"Provided default: {default} is neither a component nor a component batch.")
 
         for path, components in self.overrides.items():
-            stream.log(  # type: ignore[attr-defined]
-                f"{self.blueprint_path()}/ViewContents/individual_overrides/{path}", components, recording=stream
+            stream.log(
+                f"{self.blueprint_path()}/ViewContents/individual_overrides/{path}",
+                components,  # type: ignore[arg-type]
             )
```
emilk, line 150: Not sure what happened here, but I assume you know what you're doing

rerun_py/rerun_sdk/rerun/recording_stream.py
```diff
@@ -291,41 +408,668 @@ def __del__(self):  # type: ignore[no-untyped-def]
...
+        """
+
+        from ._log import log_file_from_path
+
+        log_file_from_path(
+            file_path=file_path,
+            entity_path_prefix=entity_path_prefix,
+            static=static,
+            recording=self,
+        )
```
emilk, line 1078: This is a lot of new code. It looks like it is all just forwarding calls - can we forward the docstring too o we don't need to maintain two sets of them?

emilk (commented):
I'm not a big fan of the 700 lines of new code we need to maintain.

What is the benefit of this duplication?
What are the downsides?
How can we ensure the two APIs don't diverge?

rerun_py/rerun_sdk/rerun/__init__.py
abey79: Good catch! I replaced that by raising an exception, but ruff deleted the entire thing because, well, the `if` is never supposed to be true. And it's right in a sense: no tooling will ever allow installing the `rerun-sdk` package in a <3.9 environment. I'll put it back regardless with a ruff exception, just to be safe.

rerun_py/rerun_sdk/rerun/archetypes/depth_image_ext.py
abey79: This is a bit of an opinionated lint from Ruff, but I like it because it forces developer to learn about that feature ad think of how to use it.

This is the lint description: https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

Whether or not to chain exception is a logical/semantic question and only the dev can make the call, so the lint rejects the implicit behaviour.

rerun_py/rerun_sdk/rerun/blueprint/api.py
abey79: Well you can, but I don't like it much because of how many things are falsy in python.

abey79:
> I'm not a big fan of the 700 lines of new code we need to maintain.
> 
> What is the benefit of this duplication? What are the downsides? How can we ensure the two APIs don't diverge?

The primary benefit of this duplication is to make LSP happy with both the stateful and stateless API. For library code, this is very significant and, IMO, outweighs some of the annoyance required.

These docstring cannot be just dynamically duplicated, because the stateful apis don't have a `recording` argument. This could _maybe_ done by removing the `recording` argument from the stateful API, which I advocate.

I see this PR as a first step toward a saner situation where the stateless API is first class. Here is my plan, which addresses the duplication (somewhat):
- https://github.com/rerun-io/rerun/issues/9187

emilk (approved):

rerun_py/rerun_sdk/rerun/blueprint/api.py
abey79: Yes. This is a first step towards removing all these `# type: ignore`, but there is more work to do.

PR #9191, opened 2025-03-05
Update docstrings for `TimeInt` and `TimeRangeBoundary`

emilk:
### Related

### DIFF
--- a/rerun_py/rerun_sdk/rerun/datatypes/time_int_ext.py
+++ b/rerun_py/rerun_sdk/rerun/datatypes/time_int_ext.py
@@ -10,26 +10,31 @@ class TimeIntExt:
         """
         Create a new instance of the TimeInt datatype.
 
+        Exactly one of `seq`, `seconds`, or `nanos` must be provided.
+
         Parameters
         ----------
         seq:
-            Time as a sequence number. Mutually exclusive with seconds and nanos.
+            Time as a sequence number.
+
         seconds:
-            Time in seconds. Mutually exclusive with seq and nanos.
+            Time in seconds.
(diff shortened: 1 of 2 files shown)


rerun_py/rerun_sdk/rerun/datatypes/time_int_ext.py
```diff
@@ -10,26 +10,31 @@ def __init__(self: Any, *, seq: int | None = None, seconds: float | None = None,
         """
         Create a new instance of the TimeInt datatype.
 
+        Exactly one of `seq`, `seconds`, or `nanos` must be provided.
```
abey79, line 13: I think we should be able to roll the `@overload` stuff on `__init__` here too.

rerun_py/rerun_sdk/rerun/datatypes/time_range_boundary_ext.py
```diff
@@ -76,35 +76,38 @@ def absolute(
         """
         Boundary that is at an absolute time.
 
+        Exactly one of 'time', 'seq', 'seconds', or 'nanos' must be provided.
```
abey79, line 79: ditto. and on `__init__` too.

abey79 (approved):
LGTM, but let's `@overload` all the things!

rerun_py/rerun_sdk/rerun/datatypes/time_int_ext.py
emilk: I'd rather not take the time to do that now. That's an orthogonal improvement

PR #9192, opened 2025-03-05
Python API: Rename `sequence` time parameter to `seq`

emilk:
This makes it the same as the existing APIs for `TimeInt` and `TimeRangeBoundary`.

It is also shorter.

### Related


docs/content/reference/migration/migration-0-17.md
```diff
@@ -53,7 +53,7 @@ Now the override can be specified from the blueprint, removing the need to inclu
 
 ```python
 for t in range(1000):
-    rr.set_index("frame_nr", sequence=t)
+    rr.set_index("frame_nr", seq=t)
```
abey79, line 56: I should have pointed that out in the other PR already, but the 0.17 migration guide should best reflect the 0.17 era situation 😬

emilk: indeed - this was a search-and-replace mistake

emilk:
It'll be easier to do this from scratch then fixing those merge conflicts

PR #9196, opened 2025-03-05
Update test and snippets using `new_recording()` to _actually_ use the stateless API

abey79:
### Related

- Follow up to #9182

### What

Some (too little imo) existing snippets and test explicitly create a recording stream but didn't use its API, relying on the `recording=rec` shenanigans. This PR fixes this.


oxkitsune (approved):
this is great! might make sense to include something similar in #9148

abey79:
> this is great! might make sense to include something similar in #9148

Indeed, that PR needs to be up to new standards. I'll drop a comment or two!

PR #9200, opened 2025-03-05
New C++ API for timestamp/duration indices

emilk:
### Related

### What
Updated the C++ API to distinguish between absolute timestamps and and relative durations.

We've deprecated the following functions in `RecordingStream`, with the following replacements:
* `set_time_sequence` -> `set_index_sequence`
* `set_time` -> `set_index_duration` or `set_index_timestamp`
* `set_time_seconds` -> `set_index_duration_secs` or `set_index_timestamp_seconds_since_epoch`
* `set_time_nanos` -> `set_index_duration_nanos` or `set_index_timestamp_nanos_since_epoch`

`TimeColumn` also has deprecated functions.

### Questions to reviewer
The names for the functions in `TimeColumn` are somewhat inconsistent with those of `RecordingStream`.

### Other notes
I also changed/broke the C API.

### TODO
* [ ] Full check


emilk:
@rerun-bot full-check

docs/content/reference/migration/migration-0-23.md
```diff
@@ -8,7 +8,16 @@ Previously, you could (confusingly) have two timelines with the same name, as lo
...
 Timelines are now uniquely identified by name, and if you use different types on the same timeline, you will get a logged warning, and the _latest_ type will be used to interpret the full set of time data.
 
-## 🐍 Python: replaced `rr.set_time_*` with `rr.set_index`
+## Rename some timeline-related things as "index"
+We're planning on adding support for different types of indices in the future, so to that point we're slowly migrating our API to refer to these things as _indices_ rather than _timelines_.
+
+## Differentiate between timestamps and durations
+We've added a explicit API for setting time, where you need to explicitly specify if a time is either a timestamp (e.g. `2025-03-03T14:34:56.123456789`) or a duration (e.g. `123s`).
+
+Before Rerun would try to guess what you meant (small values were assumed to be durations, and large values were assumes to be durations since the Unix epoch, i.e. timestamps).
```
Wumpf, line 17:
```suggestion
Before, Rerun would try to guess what you meant (small values were assumed to be durations, and large values were assumes to be durations since the Unix epoch, i.e. timestamps).
```

rerun_cpp/src/rerun/c/rerun.h
```diff
@@ -271,11 +271,14 @@ enum {
 typedef uint32_t rr_time_type;
 
 enum {
```
Wumpf, line 273: Leaving a comment here why 0 is unused would be helpful

rerun_cpp/src/rerun/c/rerun.h
```diff
@@ -487,37 +490,16 @@ extern void rr_recording_stream_stdout(rr_recording_stream stream, rr_error* err
...
-///
-/// Used for all subsequent logging performed from this same thread, until the next call
-/// to one of the time setting methods.
-///
-/// For example:
-/// `rr_recording_stream_set_time_nanos(stream, "sim_time", sim_time_nanos, &err)`.
-extern void rr_recording_stream_set_time_nanos(
-    rr_recording_stream stream, rr_string timeline_name, int64_t ns, rr_error* error
+/// `rr_recording_stream_set_time_sequence(stream, "frame_nr", RR_TIME_TYPE_SEQUENCE, frame_nr, &err)`.
+extern void rr_recording_stream_set_index(
```
Wumpf, line 500: nice, makes our C api crispier! <3

rerun_cpp/src/rerun/collection.hpp
```diff
@@ -355,6 +355,15 @@ namespace rerun {
...
         }
 
+        /// Returns a pointer to the first element, if any, else `nullptr`.
+        const TElement* first() const {
+            if (empty()) {
+                return nullptr;
+            } else {
+                return begin();
+            }
+        }
```
Wumpf, line 365: `first` is very unidiomatic in C++, practically unheard of afaik! So I'd rather not add that, seems quite confusing
However, I think we should just make all this safer & better by having `data` return `nullptr` if the collection is empty. That way all of `data`/`begin`/`end` get better.
This also still lines up with what folks would expect from `begin` - the only condition really is that for an empty _thing_ `begin` == `end`. There doesn't seem to be much definition on what it is supposed to be usually

(plz extend existing (?) unit tests for empty collections accordingly)

Wumpf (approved):
Looking great! Feeling somewhat strong about not having `Collection::first` (see comment), but otherwise all good

> The names for the functions in TimeColumn are somewhat inconsistent with those of RecordingStream.

Seems like a good opportunity to make things more similar to Rust? I figure you'll eventually want to move towards `IndexColumn`? 🤔 
Not sure what the exact question to the Reviewer is here 😄. Can we make things less inconsistent?

rerun_cpp/src/rerun/collection.hpp
emilk: In the end I never made use of it, so I'll just delete it.

emilk:
@rerun-bot full-check

PR #9201, opened 2025-03-05
Enable a bunch of `ruff` lints

abey79:
This PR introduce  a whole bunch of `ruff` lints along with the necessary fixes.


examples/python/face_tracking/face_tracking.py
```diff
@@ -258,7 +258,7 @@ def detect_and_log(self, image: cv2.typing.MatLike, frame_time_nano: int) -> Non
             else self._detector.detect(image)
         )
 
-        def is_empty(i):  # type: ignore[no-untyped-def]
+        def is_empty(i: Iterator[Any]) -> Optional[bool]:
             try:
                 next(i)
                 return False
```
timsaucer, line 264: I'm guessing this was auto-added, but I think this should return `bool`, not `Optional[bool]`. I'm guessing it did that because it didn't realize that all possible return paths are covered.

timsaucer (approved):
Very nice.

PR #9203, opened 2025-03-05
Fix nightly build by updating pixi and constraining to `--platform` in `pixi_install_wheel.py`

jleibs:
### What

The fact that `examples-common` does not support linux-arm was still leading to pixi confusion when trying to install the wheel into the environment, even though it should be totally valid.
- This is still related to: https://github.com/prefix-dev/pixi/issues/1051

Just updating the wheel for the current platform we are testing appears to work around the problem.


jleibs:
@rerun-bot full-check

abey79 (approved):
Nicely done! Let's see how CI feels about this.

PR #9204, opened 2025-03-05
Add/cleanup some `@overload`

abey79:
Title. Follow-up to #9191

### DIFF
--- a/rerun_py/tests/unit/test_time_range_boundary.py
+++ b/rerun_py/tests/unit/test_time_range_boundary.py
@@ -7,47 +7,47 @@ import rerun as rr
 def test_time_range_boundary_failure_cases() -> None:
     # Too many arguments for absolute.
     with pytest.raises(ValueError):
-        rr.TimeRangeBoundary.absolute(rr.TimeInt(seq=0), seq=123)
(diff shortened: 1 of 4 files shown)


rerun_py/tests/unit/test_time_range_boundary.py
```diff
@@ -7,47 +7,47 @@
 def test_time_range_boundary_failure_cases() -> None:
     # Too many arguments for absolute.
     with pytest.raises(ValueError):
-        rr.TimeRangeBoundary.absolute(rr.TimeInt(seq=0), seq=123)
+        rr.TimeRangeBoundary.absolute(rr.TimeInt(seq=0), seq=123)  # type: ignore[call-overload]
```
emilk, line 10: Can the `type: ignore[call-overload]` be applied to the entire contents of the function or file instead, to reduce the visual clutter?

emilk (approved):
🙏

rerun_py/tests/unit/test_time_range_boundary.py
abey79: i didn't find a way for that, and tbh i like visibility for that stuff, as its there for a reason (even if it looks ugly)

PR #9205, opened 2025-03-05
Add tests for `VisibleTimeRange` and intra-timestamp data

lucasmerlin:
### What

This takes some parts of our [thumbnail demo](https://github.com/rerun-io/rerun/compare/main...lucas/thumbnails) branch and turns them into the test. 
Also, we realized that this is similar to the check_range_partial_updates release checklist check, so I turned that into a test and removed it from the checklist.
I tried to recreate the checklist recording as faithfully as possible and I think it logically exactly matches what's happening (even though it's now in 2d). @teh-cmc can you review this to confirm that I didn't miss anything from what the checklist check did?

Also it's a bit annoying that the initial points are so small, but I think it's fine?


teh-cmc:
@lucasmerlin 's off so I'll just quickly take this over the finish line

teh-cmc (approved):

PR #9210, opened 2025-03-06
Infer partition size for FixedSizeList-backed components

### LINKED ISSUE #9171: Infer partition size for `FixedSizeList`-backed components
This will make it easier to send columns of points, lines and so on. 

When trying to do this currently it results in something like this:
```
ValueError: Expected either a flat array with a length multiple of 3 elements, or an array with shape (`num_elements`, 3). Shape of passed array was (17004, 21, 3).
```

oxkitsune:
### Related

Closes #9171

### What

This lets the user do:
```py
rr.send_columns(
  "random_3d_points",
  indexes=[times],
  columns=rr.Points3D.columns(positions=np.random.random((len(times), 21, 3))),
)
```


crates/build/re_types_builder/src/codegen/python/mod.rs
```diff
@@ -2649,8 +2649,8 @@ fn quote_columnar_methods(reporter: &Reporter, obj: &Object, objects: &Objects)
             for batch in batches:
                 arrow_array = batch.as_arrow_array()
 
-                # For primitive arrays, we infer partition size from the input shape.
-                if pa.types.is_primitive(arrow_array.type):
+                # For primitive arrays and fixed size list arrays, we infer partition size from the input shape.
+                if pa.types.is_primitive(arrow_array.type) or pa.types.is_fixed_size_list(arrow_array.type):
```
Wumpf, line 2653: it's technically fixed sized lists of primitive arrays, no? 🤔

rerun_py/rerun_sdk/rerun/_validators.py
```diff
@@ -81,10 +81,10 @@ def flat_np_array_from_array_like(array: npt.NDArray[Any], dimension: int) -> np
...
         valid = (array.shape[0] % dimension) == 0
     elif len(array.shape) >= 2:
-        valid = array.shape[1] == dimension
-
-        # Don't care about trailing dimensions if they're all 1.
-        valid = valid and all(d == 1 for d in array.shape[2:])
+        # Get the last dimension that is not 1. If all dimensions are 1, this returns 1.
+        last_non_singleton = next((d for d in reversed(array.shape[1:]) if d != 1), 1)
+        valid = last_non_singleton == dimension
+        print("shape:", array.shape, last_non_singleton, valid)
```
Wumpf, line 87: left over debug print?

rerun_py/rerun_sdk/rerun/_validators.py
```diff
@@ -81,10 +81,10 @@ def flat_np_array_from_array_like(array: npt.NDArray[Any], dimension: int) -> np
     if len(array.shape) == 1:
         valid = (array.shape[0] % dimension) == 0
     elif len(array.shape) >= 2:
-        valid = array.shape[1] == dimension
-
-        # Don't care about trailing dimensions if they're all 1.
-        valid = valid and all(d == 1 for d in array.shape[2:])
+        # Get the last dimension that is not 1. If all dimensions are 1, this returns 1.
+        last_non_singleton = next((d for d in reversed(array.shape[1:]) if d != 1), 1)
```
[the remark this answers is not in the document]
Wumpf: there should be tests for this methods, both direct and indirect 🤔 
you may want to add additional ones to pin down this behavior

Wumpf (changes requested):
pixi.lock change is suspicious, why did it change here?
Cool that this just works :O. But needs a test or snippet using this
Could `docs/snippets/all/archetypes/points3d_column_updates.py` be updated for that?

rerun_py/rerun_sdk/rerun/_validators.py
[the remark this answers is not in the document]
oxkitsune: yea, was thinking the same, also might be good to document the behavior somewhere.

crates/build/re_types_builder/src/codegen/python/mod.rs
oxkitsune: yes. good catch would come back to haunt us I think.

Wumpf:
Or maybe you're running an old pixi. We just updated https://github.com/rerun-io/rerun/pull/9203

oxkitsune:
> Or maybe you're running an old pixi. We just updated #9203

```
❯ pixi --version
pixi 0.41.4
```
I think I'm up to date right?

crates/build/re_types_builder/src/codegen/python/mod.rs
Wumpf: we concluded on a Zoom call that this actually doesn't matter at all here - if we had any fixed sized non primitive arrays it should just work, but alas we don't have any to begin with

Wumpf (commented):
love the new test, should have had this from the start

docs/snippets/all/archetypes/arrows3d_row_updates.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+        std::array<std::array<float, 3>, 5> origin;
+        std::array<std::array<float, 3>, 5> vector;
+        for (size_t j = 0; j < 5; j++) {
+            origin[j] = {xs[j], xs[j], 0.0};
+            vector[j] = {xs[j], xs[j], zs[j]};
+        }
+
+        origins.emplace_back(std::move(origin));
+        vectors.emplace_back(std::move(vector));
```
Wumpf, line 34: move are superfluous for `std::array` (has no effect really here since `std::array` isn't heap allocated)
```suggestion
        origins.emplace_back(origin);
        vectors.emplace_back(vector);
```

docs/snippets/all/archetypes/arrows3d_column_updates.cpp
```diff
@@ -0,0 +1,52 @@
...
+    auto time_column = rerun::TimeColumn::from_durations("time", std::move(times));
+
+    auto arrows =
+        rerun::Arrows3D().with_origins(origins).with_vectors(vectors).columns({5, 5, 5, 5, 5});
+
+    rec.send_columns(
+        "arrows",
+        time_column,
+        arrows,
+        rerun::Arrows3D().with_colors(colors).columns()
```
Wumpf, line 50: doesn't make a difference practically but let's use `update_fields` when that's the intention
```suggestion
        rerun::Arrows3D::update_fields().with_colors(colors).columns()
```

Wumpf (approved):
phew, lotsa snippet work. looking great

PR #9214, opened 2025-03-06
Run `mypy` on the contents of `scripts/`

abey79:
So far, we didn't run mypy on our `scripts/`, and they were riddled with issues. This PR:
- moves the mypy config to `rerun_py/pyproject.toml`
- add `scripts` to the list of places mypy checks
- fixes all the errors that mypy found


grtlr (approved):
Nice!

grtlr (approved):

PR #9216, opened 2025-03-06
Update MSRV to 1.84

emilk:
Best reviewed commit-by-commit.

This PR also opts-in to the MSRV-aware crate resolver in cargo.
See [this link](https://blog.rust-lang.org/2025/01/09/Rust-1.84.0.html#cargo-considers-rust-versions-for-dependency-version-selection) for more.

* Sibling PR: https://github.com/rerun-io/dataplatform/pull/346


emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

crates/build/re_types_builder/src/codegen/rust/blueprint_validation.rs
```diff
@@ -18,7 +18,8 @@ pub(crate) fn generate_blueprint_validation(
 ) {
     let blueprint_scope = Some("blueprint".to_owned());
     let mut code = String::new();
-    code.push_str(&format!("// {}\n", autogen_warning!()));
+    code.push_str(&format!("// {}\n\n", autogen_warning!()));
+    code.push_str("#![expect(clippy::empty_line_after_doc_comments)]\n\n");
```
Wumpf, line 22: a bit strange, why not fix that empty line?

Wumpf (approved):
looks good!

crates/build/re_types_builder/src/codegen/rust/blueprint_validation.rs
emilk: Because I suspect that's gonna be a lot harder

Wumpf: kk, didn't check; may or may not 😄

PR #9226, opened 2025-03-06
Introduce `IndexCell`

emilk:
### Related
* Part of #8635 

### What
Introduces a new type, `IndexCell`:

```rs
pub struct IndexCell {
    pub typ: TimeType,
    pub value: NonMinI64,
}
```

And changes `TimePoint` to use it:

```rs
pub struct TimePoint(BTreeMap<TimelineName, IndexCell>);
```

### Naming
The thought behind the name `IndexCell` is two-fold:

A) It is symmetrical with `DataCell`
B) It will allow us to at some point broaden to more types of indices (e.g. UUID)

I’d like to rename `TimeType` to `IndexType` at some point. I haven’t planned this in detail though; I mostly want to avoid having to rename the _new_ things I add.

### TODO
* [x] Full check

### Future work
There are A LOT of tests using `Timeline` and `TimeInt` that would be better seved by using `TimelineName` and `IndexCell`, but this is too much to clean up in this PR.

There is also A LOT of (old) parameters of the type `impl TryInt<TimeInt>` which in effect means "An i64 or TimeInt that will be converted to a non-static `TimeInt` via saturating cast". It took me a long time to figure out that this is what it meant. We should come up with a nicer pattern for this.

Maybe deprecate `TimePoint::with` and `TimePoint::insert`?


emilk:
@rerun-bot full-check

teh-cmc (approved):
"nonstatic" sounds a bit weird, but I guess we gotta stay away from "temporal" if we want to move towards generic indexes... :shrug:

emilk:
Good point 🤔 
I'll add a TODO about that naming to the parent issue

PR #9229, opened 2025-03-07
Update `ring`

emilk:
### Related
Fix RUSTSEC


Cargo.lock
```diff
@@ -3930,7 +3930,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4979f22fdb869068da03c9f7528f8297c6fd2606bc3a4affe42e6a823fdb8da4"
 dependencies = [
  "cfg-if",
- "windows-targets 0.52.6",
+ "windows-targets 0.48.5",
```
Wumpf, line 3933: this one seems to oscillate, seen that before. Strange.

Wumpf (approved):

Issue #9024, opened 2025-02-12
Logging level too low when running `rerun --serve-web`


jleibs:
0.21:
```
➜  rerun --serve-web
[2025-02-12T18:33:49Z INFO  re_sdk_comms::server] Hosting a SDK server over TCP at 0.0.0.0:9876. Connect with the Rerun logging SDK.
[2025-02-12T18:33:49Z INFO  re_ws_comms::server] Hosting a WebSocket server on ws://localhost:9877. You can connect to this with a native viewer (`rerun ws://localhost:9877`) or the web viewer (with `?url=ws://localhost:9877`).
[2025-02-12T18:33:49Z INFO  re_sdk::web_viewer] Hosting a web-viewer at http://localhost:9090?url=ws://localhost:9877
```

0.22:
```
➜  rerun --serve-web

```
Having the INFO log statements with port number, etc are extremely important in this case.

Wumpf:
this was a deliberate change we made to reduce logspam in the context of the _viewer_. We have to find a compromise of when to show how much log

jleibs:
Yeah, the trade-off is hard.

That said, the real motivation for the change (https://github.com/rerun-io/rerun/issues/7815) wasn't to reduce logspam in the viewer, but logspam via the SDK when using in a notebook context.

I don't actually see an issue with the viewer sticking with INFO by default. It's the SDK info messages that are kind of spammy.

jleibs:
> We need to find a way to not have them show up in the notebook then
The notebook doesn't spawn a viewer instance.

Wumpf:
Sure those particular messages won't show up there, but the logspam issues was in particular about the notebook showing too much log from the viewer independent of what that shows in particular

emilk:
The cause of this problem is:
* https://github.com/rerun-io/rerun/pull/8918
*

Wumpf:
@jleibs and me had a brief call about this. Conclusions
* standalone cli should always show info
* sdk should show warning
* viewer opened from sdk is where things get trickier
    * if we're spawning a separate process (non-notebook) then probably default to what cli does, but we could pick what we want here explicitely
    * if we're running in the same process, do what sdk does (notebook!)

PR #9231, opened 2025-03-07
Revert log level to `info` for everything except Python sdk

emilk:
* Closes https://github.com/rerun-io/rerun/issues/9024
* Reverts https://github.com/rerun-io/rerun/pull/8988
* Reverts https://github.com/rerun-io/rerun/pull/8918

Also, fixes some subtle issues with setting the max log filter on native.

### DIFF
--- a/crates/utils/re_log/src/lib.rs
+++ b/crates/utils/re_log/src/lib.rs
@@ -94,7 +94,8 @@ const CRATES_AT_INFO_LEVEL: &[&str] = &[
 /// Web: `debug` since web console allows arbitrary filtering.
 #[cfg(not(target_arch = "wasm32"))]
 pub fn default_log_filter() -> String {
-    let base_log_filter = if cfg!(debug_assertions) {
+    let base_log_filter = if false {
+        //cfg!(debug_assertions) {
         "debug"
     } else {
         // Important to keep the default at (at least) "info",
@@ -113,7 +114,7 @@ pub fn default_log_filter() -> String {
 /// Web: `debug` since web console allows arbitrary filtering.
 #[cfg(target_arch = "wasm32")]
 pub fn default_log_filter() -> String {
-    "debug".to_string()
+    "debug".to_owned()
 }
 
 /// Determines the log filter from the `RUST_LOG` environment variable or an explicit default.


jleibs:
It doesn't seem like that much more work to just modify this line:
https://github.com/rerun-io/rerun/blob/main/rerun_py/src/python_bridge.rs#L111

To call a tweaked version of  `setup_logging()` like  `setup_logging_with_default("warn")`

Wumpf:
yeah.. let's do that right away, stopping the bleed. Emil is out, I'm on it now

jleibs (approved):
Looks reasonable if it's been tested.

Wumpf:
Checked cli, python sdk on example and notebook. All of then looked correct to me

crates/top/rerun/src/commands/entrypoint.rs
```diff
@@ -809,14 +809,12 @@ fn run_impl(
...
-        #[cfg(not(feature = "server"))]
-        {
+        if !cfg!(feature = "server") {
             _ = (call_source, rxs);
             anyhow::bail!("Can't host server - rerun was not compiled with the 'server' feature");
         }
 
-        #[cfg(not(feature = "web_viewer"))]
-        if args.web_viewer {
+        if !cfg!(feature = "web_viewer") {
```
emilk, line 817: Looks like we can no longer to `--serve` without a web viewer now, which is a regression

crates/utils/re_log/src/lib.rs
```diff
@@ -86,32 +86,78 @@ const CRATES_AT_INFO_LEVEL: &[&str] = &[
...
 pub fn default_log_filter() -> String {
-    let mut rust_log = std::env::var("RUST_LOG").unwrap_or_else(|_| {
-        if cfg!(debug_assertions) {
-            "debug".to_owned()
-        } else {
-            "warn".to_owned()
-        }
-    });
+    let base_log_filter = if false {
+        //cfg!(debug_assertions) {
```
emilk, line 98: Mistake?

emilk: * Fixed in https://github.com/rerun-io/rerun/pull/9290

PR #9233, opened 2025-03-07
Check dataset version when loading LeRobot dataset

### LINKED ISSUE #9177: Explicit version checking for LeRobot datasets
We should explicitly check the version of the dataset when attempting to load it.

oxkitsune:
### Related

* Closes #9177

### What

The loader now explicitly validates the dataset version, preventing unexpected behavior or silent failures due to unsupported dataset versions.

### DIFF
--- a/crates/store/re_data_loader/src/lerobot.rs
+++ b/crates/store/re_data_loader/src/lerobot.rs
@@ -20,8 +20,13 @@ use parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 
-/// Check whether the provided path contains a Le Robot dataset.
+/// Check whether the provided path contains a `LeRobot` dataset.
 pub fn is_lerobot_dataset(path: impl AsRef<Path>) -> bool {
+    is_v1_lerobot_dataset(path.as_ref()) || is_v2_lerobot_dataset(path.as_ref())
+}
+
+/// Check whether the provided path contains a v2 `LeRobot` dataset.
+pub fn is_v2_lerobot_dataset(path: impl AsRef<Path>) -> bool {
     let path = path.as_ref();
 
     if !path.is_dir() {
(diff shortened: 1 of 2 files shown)


crates/store/re_data_loader/src/lerobot.rs
```diff
@@ -35,6 +40,21 @@ pub fn is_lerobot_dataset(path: impl AsRef<Path>) -> bool {
...
 
+/// Check whether the provided path contains a v1 `LeRobot` dataset.
+pub fn is_v1_lerobot_dataset(path: impl AsRef<Path>) -> bool {
+    let path = path.as_ref();
+
+    if !path.is_dir() {
+        return false;
+    }
+
+    ["meta_data", "data"].iter().all(|subdir| {
```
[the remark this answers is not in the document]
oxkitsune: Added a comment.

I can't really think of any other method that doesn't require us to parse the actual dataset info file. The schema is a bit different between v1 and v2 datasets, so we'd have to parse it differently.

crates/store/re_data_loader/src/lerobot.rs
```diff
@@ -20,16 +20,40 @@ use parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
...
+pub fn is_v2_lerobot_dataset(path: impl AsRef<Path>) -> bool {
+    let path = path.as_ref();
+
+    if !path.is_dir() {
+        return false;
+    }
+
+    // v2 `LeRobot` datasets store the metadata in a `meta` directory,
+    // instead of the `meta_data` directory used in v1 datasets.
+    has_sub_directories(&["meta_data", "data"], path)
```
Wumpf, line 38: why is this check the same then as in `is_v1_lerobot_dataset`

Wumpf (changes requested):

crates/store/re_data_loader/src/lerobot.rs
oxkitsune: Wow i missed that 🤦  thanks

Wumpf (approved):

PR #9236, opened 2025-03-07
Add `RecordingStream::set_index`

emilk:
### Related
* Part of https://github.com/rerun-io/rerun/issues/8635

### What
Introduce a new API for setting timeline values.

`rec.set_index(timeline_name, rerun::IndexCell)` 

#### Deprecated: `rec.set_time_seconds` and `rec.set_time_nanos`
Use one of these instead:
* `rec.set_duration_seconds`
* `rec.set_timestamp_seconds_since_epoch`
* `rec.set_index` with `std::time::Duration`
* `rec.set_index` with `std::time::SystemTime`

### TODO
* [ ] Migration guide
* [ ] Full check

PR #9241, opened 2025-03-09
Refactor time code, and change formatting of time ranges in URIs

emilk:
### Related
* Follows https://github.com/rerun-io/rerun/pull/9236
* Part of #8635 

### What
A lot of refactors and cleanups in preparation for the above issue.

`struct Time` is going away (but is not yet gone).
In its place will be `struct Duration` and the new `struct Timestamp`.
I've started moving code out of `Time` and onto these types.
I've also started using [`jiff`](https://crates.io/crates/jiff), which is looking to become the best time/date library.

`IndexCell` now implements `FromStr` and `Display`, using jiff. This is used for the time range parameters for the gRPC URI.
 The format is one of:

* `123` for sequences
* `123.123s` for durations
* `2022-01-01T00:00:03.123456789Z` for timestamps

The duration and timestamps should follow ISO 8601

### Do-no-merge
Because the merge target is
* https://github.com/rerun-io/rerun/pull/9236

### DIFF
--- a/crates/store/re_log_types/src/index/non_min_i64.rs
+++ b/crates/store/re_log_types/src/index/non_min_i64.rs
@@ -208,6 +208,24 @@ impl<'de> serde::Deserialize<'de> for NonMinI64 {
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+#[error("Failed to parse NonMinI64: {0}")]
+pub enum ParseNonMinI64Error {
+    Std(#[from] std::num::ParseIntError),
+
+    #[error("out-of-range")]
+    OutOfRange,
+}
+
+impl std::str::FromStr for NonMinI64 {
+    type Err = ParseNonMinI64Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let int = i64::from_str(s)?;
+        Self::new(int).ok_or(ParseNonMinI64Error::OutOfRange)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/crates/store/re_log_types/src/index/non_min_i64.rs
+++ b/crates/store/re_log_types/src/index/non_min_i64.rs
@@ -213,8 +213,8 @@ impl<'de> serde::Deserialize<'de> for NonMinI64 {
 pub enum ParseNonMinI64Error {
     Std(#[from] std::num::ParseIntError),
 
-    #[error("out-of-range")]
-    OutOfRange,
+    #[error("Value is equal to minimum i64. Every i64 integer *except* the lowest representable number of a signed 64 bit number is valid.")]
+    InvalidValue,
 }
 
 impl std::str::FromStr for NonMinI64 {
@@ -222,7 +222,7 @@ impl std::str::FromStr for NonMinI64 {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let int = i64::from_str(s)?;
-        Self::new(int).ok_or(ParseNonMinI64Error::OutOfRange)
+        Self::new(int).ok_or(ParseNonMinI64Error::InvalidValue)
     }
 }
 
--- a/crates/store/re_log_types/src/index/time.rs
+++ b/crates/store/re_log_types/src/index/time.rs
@@ -154,38 +154,8 @@ impl Time {
 
     /// Best effort parsing of a string into a [`Time`] from a human readable string.
     pub fn parse(s: &str, time_zone_for_timestamps: TimeZone) -> Option<Self> {
-        // Try parsing as a simple relative time with unit suffix (e.g. "1.234s", "1.234ms")
-        let suffixes = [("s", 1e9), ("ms", 1e6), ("us", 1e3), ("ns", 1.0)];
-        for (suffix, to_ns) in suffixes {
-            if let Some(s) = s.strip_suffix(suffix) {
-                if let Ok(value) = s.parse::<f64>() {
-                    return Some(Self::from_ns_since_epoch((value * to_ns).round() as i64));
-                }
-            }
-        }
-
-        // Parse a relative time of day.
-        let time_formats = [
-            // Just time with milliseconds
-            "[hour]:[minute]:[second].[subsecond]",
-            // Just time
-            "[hour]:[minute]:[second]",
-        ];
-        for format in time_formats {
-            let format = time::format_description::parse_borrowed::<2>(format)
-                .expect("Invalid format string");
-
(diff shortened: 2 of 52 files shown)


crates/store/re_log_types/src/index/timeline.rs
```diff
@@ -65,14 +64,30 @@ impl Timeline {
...
+    /// For absolute timestamps.
+    #[inline]
+    pub fn new_timestamp(name: impl Into<TimelineName>) -> Self {
+        Self {
+            name: name.into(),
+            typ: TimeType::Time,
         }
     }
 
+    #[deprecated(note = "Use `Timeline::new_duration` or `new_timestamp` instead")]
```
Wumpf, line 85: here and elsewhere, should fill ou the `since` field so it's easier for us to keep track of when to remove deprecated methods

crates/store/re_log_types/src/index/time.rs
```diff
@@ -55,272 +35,101 @@ impl Time {
...
 
         20 <= years_since_epoch && years_since_epoch <= 150
     }
 
-    /// Returns the absolute datetime if applicable.
-    pub fn to_datetime(self) -> Option<OffsetDateTime> {
-        let ns_since_epoch = self.nanos_since_epoch();
-        if self.is_absolute_date() {
-            OffsetDateTime::from_unix_timestamp_nanos(ns_since_epoch as i128).ok()
+    /// RFC3339
```
Wumpf, line 45:
```suggestion
    /// Formats the time as specified by ISO standard [`RFC3339`](https://www.rfc-editor.org/rfc/rfc3339.html).
```

crates/store/re_log_types/src/index/non_min_i64.rs
```diff
@@ -208,6 +208,24 @@ impl<'de> serde::Deserialize<'de> for NonMinI64 {
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+#[error("Failed to parse NonMinI64: {0}")]
+pub enum ParseNonMinI64Error {
+    Std(#[from] std::num::ParseIntError),
+
+    #[error("out-of-range")]
+    OutOfRange,
```
Wumpf, line 217: out of range seems a bit odd. the range is really "everything but i64::MIN", so it's more like

```suggestion
    #[error("Value is equal to minimum i64. Every i64 integer *except* the lowest representable number of a signed 64 bit number is valid.")]
    InvalidValue,
```

crates/store/re_log_types/src/index/timestamp.rs
```diff
@@ -0,0 +1,428 @@
...
+    type Error = web_time::SystemTimeError;
+
+    fn try_from(time: web_time::SystemTime) -> Result<Self, Self::Error> {
+        time.duration_since(web_time::SystemTime::UNIX_EPOCH)
+            .map(|duration_since_epoch| Self(duration_since_epoch.as_nanos() as _))
+    }
+}
+
+// ------------------------------------------
+// Joff converters
```
Wumpf, line 65:
```suggestion
// `jiff` converters
```

crates/store/re_log_types/src/index/timestamp_format.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    /// Display as UTC.
+    Utc,
+
+    /// Show as seconds since unix epoch
+    UnixEpoch,
+}
+
+impl TimestampFormat {
+    pub fn to_jiff_tz(self) -> jiff::tz::TimeZone {
```
Wumpf, line 16: `tz` is a bit too unclear without having lots of context imho
```suggestion
    pub fn to_jiff_time_zone(self) -> jiff::tz::TimeZone {
```

crates/store/re_log_types/src/index/timestamp.rs
```diff
@@ -0,0 +1,428 @@
...
+}
+
+impl std::fmt::Debug for Timestamp {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.format_iso().fmt(f)
+    }
+}
+
+impl Timestamp {
+    /// RFC3339
```
Wumpf, line 113:
```suggestion
    /// Formats the time as specified by ISO standard [`RFC3339`](https://www.rfc-editor.org/rfc/rfc3339.html).  
```

crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -2071,30 +2110,27 @@ impl RecordingStream {
...
+        self.set_index(timeline, IndexCell::from_sequence(sequence.into()));
+    }
 
-            ThreadInfo::set_thread_time(
-                &inner.info.store_id,
-                timeline.into(),
-                IndexCell::from_sequence(seq),
-            );
-        };
+    /// Short for `set_index(timeline, std::time::Duration::from_secs_f64(secs))`.
```
Wumpf, line 2117: compared to the others a bit short on doc

crates/utils/re_uri/src/lib.rs
```diff
@@ -23,64 +25,45 @@ impl TimeRange {
...
+        let min = min.parse::<IndexCell>().map_err(|err| {
+            Error::InvalidTimeRange(format!("Failed to parse time index '{min}': {err}"))
+        })?;
+        let max = max.parse::<IndexCell>().map_err(|err| {
+            Error::InvalidTimeRange(format!("Failed to parse time index '{max}': {err}"))
+        })?;
+
+        if min.typ() != max.typ() {
+            return Err(Error::InvalidTimeRange(
+                "min/max had differing types".to_owned(),
```
Wumpf, line 61: would be good to message what types were detected in case user doesn't realize what went wrong

crates/viewer/re_viewer_context/src/time_drag_value.rs
```diff
@@ -73,6 +73,33 @@ impl TimeDragValue {
...
+        timestamp_format: TimestampFormat,
+    ) -> Response {
+        match time_type {
+            TimeType::Sequence => {
+                self.sequence_drag_value_ui(ui, time, absolute, low_bound_override)
+            }
+
+            TimeType::Time => {
+                self.temporal_drag_value_ui(
+                    ui,
```
Wumpf, line 92: those inner functions no longer need to be pub I think

Wumpf (approved):
lotsa goodies in here!
What concerns me a bit is that previously all the "time" functions on RecordingStream were neatly bundled - they all started with `set_time_X`
But now it's a bit all over the place with `set_index`, `set_duration` etc. 🤔 
Don't have good idea yet what to do about it

PR #9247, opened 2025-03-10
Stop inspecting `is_official_example` in `StoreInfo`

### LINKED ISSUE #9189: Remove `is_official_example` and rely on `rerun_example_` prefix in analytics

grtlr:
### Related

* Closes #9189.

### What

Title. _This change should be non-breaking._


jprochazk:
We may have some PostHog stuff that depends on it - can you please check? Alternatively, add it back to the event only and always set it to the value of `app_id_starts_with_rerun_example`

jprochazk (approved):

PR #9248, opened 2025-03-10
Remove duplicate `send_columns` docs entry

oxkitsune:
### What

Remove duplicate point cloud `send_columns` docs entry.


Wumpf (approved):
great catch, thanks!

PR #9250, opened 2025-03-10
minor: add python dependencies for examples

timsaucer:
### What

This PR adds two python dependencies for use in the examples. Additionally, it moves the `remote` feature to be used by default.

I chose not to remove `remote` as a feature and just include its dependencies so that users who are extending rerun without remote connection capability can choose to opt out of it.


Wumpf:
`remote` feature got removed in
* https://github.com/rerun-io/rerun/pull/9252

pixi.toml
```diff
@@ -668,6 +668,8 @@ umap-learn = "==0.5.7"
 # External deps
 jupyter = ">=1.0"
 polars = ">=0.12.0"
+datafusion = ">=45.2.0"
+av = "14.2.0"
```
Wumpf, line 672: Emil would be insisting on sorting those alphabetically ;-)

Wumpf (approved):

pixi.lock
emilk, line 1: Unintentionally deleted I assume?

PR #9209, opened 2025-03-06
Archetype based overrides & defaults 

Wumpf:
### Related

* Fixes #9059

### What

Defaults and overrides can now be set using archetypes. This not only leads to a much more ergonomic API with working auto-complete for archetype fields, but also paves the way for handling tagging in blueprint overrides/defaults which in the future will allow more fine grained defaults/overrides using the archetype "context information".

### Examples

Setting default & override for radius

Before:
```py
rrb.Spatial2DView(
    name="Rect 1",
    origin="/",
    contents=["/**"],
    defaults=[rr.components.Radius(2)],
    overrides={"rect/0": [rr.components.Radius(1)]},
)
```
After:
```py
rrb.Spatial2DView(
    name="Rect 1",
    origin="/",
    contents=["/**"],
    defaults=[rr.Boxes2D.from_fields(radii=1)],
    overrides={"rect/0": rr.Boxes2D.from_fields(radii=2)},
)
```

Setting up styles for a plot.

Before:
```py
rrb.TimeSeriesView(
    name="Trig",
    origin="/trig",
    overrides={
        "/trig/sin": [rr.components.Color([255, 0, 0]), rr.components.Name("sin(0.01t)")],
        "/trig/cos": [rr.components.Color([0, 255, 0]), rr.components.Name("cos(0.01t)")],
    },
),
rrb.TimeSeriesView(
    name="Classification",
    origin="/classification",
    overrides={
        "classification/line": [rr.components.Color([255, 255, 0]), rr.components.StrokeWidth(3.0)],
        "classification/samples": [rrb.VisualizerOverrides("SeriesPoint")], # This ensures that the `SeriesPoint` visualizers is used for this entity.
    },
),
```
After:
```py
rrb.TimeSeriesView(
    name="Trig",
    origin="/trig",
    overrides={
        "/trig/sin": rr.SeriesLine.from_fields(color=[255, 0, 0], name="sin(0.01t)"),
        "/trig/cos": rr.SeriesLine.from_fields(color=[0, 255, 0], name="cos(0.01t)"),
    },
),
rrb.TimeSeriesView(
    name="Classification",
    origin="/classification",
    overrides={
        "classification/line": rr.SeriesLine.from_fields(color=[255, 255, 0], width=3.0),
        "classification/samples": rrb.VisualizerOverrides("SeriesPoint"), # This ensures that the `SeriesPoint` visualizers is used for this entity.
    },
),
```

---

### Technical notes on corner cases

There's two special cases that were loose component so far:
[...]

PR #9257, opened 2025-03-11
Improve annotations on `View` classes and add `tests/python` to mypy coverage

abey79:
### Related

* Improves on #9209 

### What

This PR improves the signatures of the `View` python class (and subclasses), and adds `tests/python` in the coverage of `mypy`.


Wumpf (approved):
great stuff, thanks!

Issue #9251, opened 2025-03-10
Introduce archetype(s?) for `Visible` and `Interactive` properties


Wumpf:
We haven't documented this well, but users currently correctly override visible/interactive in their blueprints.
With https://github.com/rerun-io/rerun/pull/9209 landing, we'd break this, so we have to introduce an archetype that allows you to log this

While on it we might as well consider adding recursive visibility to this - visibility is the only hold out for "recursive override"  which is an older design we discarded in favor of building out path-expression-based overrides at some point in the future. Since path-expresion-based overrides are further out, it might make sense to add a `VisibleRecursive` component in the meantime.

Related to 
* https://github.com/rerun-io/rerun/issues/6541

PR #9258, opened 2025-03-11
`EntityProperties` archetype prototype

Wumpf:
* Part of https://github.com/rerun-io/rerun/issues/9251

As layed out in above ticket, we need to put `Interactive` and `Visible` into an entity.

This proposal (once fully propagated) solves everything noted in #9251 and additionally:
* removes the last hold-out of "recursive overrides", allowing faster & better structured blueprint query handling in the near future
* exposes the illustrious recursive visibility to users

Drawbacks:
* Small departure from https://github.com/rerun-io/rerun/issues/8368
     * this archetype will always be "special"
* Walking right into more things that need tagging, adding a little bit technical debt
     * flip side: this _adds_ much wanted tagging to the components at hand   

`EntityProperties` is a pretty poor name since it's overly general and can be confused with _everything else_ that's on an entity. Looking for something better.

----
For discoverability, the central fbs definition file:

```rs
namespace rerun.archetypes;

/// General visualization properties of an entity.
///
/// TODO(#6541): Fields of this archetype currently only have an effect when logged in the blueprint store.
// TODO(#9251): Add example
table EntityProperties (
  "attr.docs.unreleased",
  "attr.docs.category": "General"
) {
    /// Whether the entity can be interacted with.
    ///
    /// Non interactive components may still be still visible, but mouse interactions in the view are disabled.
    ///
    /// Defaults to true.
    interactive: rerun.components.Interactive ("attr.rerun.component_optional", nullable, order: 100);

    /// Whether the entity is visible.
    ///
    /// If this is set, it will take precedence over the `visible_recursive` field
    /// and any `visible_recursive` setting further up in the hierarchy.
    ///
    /// Defaults to true.
    visible: rerun.components.Visible ("attr.rerun.component_optional", nullable, order: 200);

    /// Whether the entity and its children are visible.
    ///
    /// This property is propagated down the entity hierarchy until another child entity
    /// sets `visible_recursive` to a different value at which point propagation continues with that value instead.
    ///
    /// `visible_recursive` is ignored on any individual entity that has the `visible` field set.
    /// (But this does not affect tree propagation of the property)
    ///
    /// Defaults to true.
    // TODO(#6889): Use `Visible` component and rely on field tag instead.
    visible_recursive: rerun.components.VisibleRecursive ("attr.rerun.component_optional", nullable, order: 300);
}
```


Wumpf:
draft review done: proposal got accepted with namechange to `EntityBehavior`
Will come up with the "real" PR at a later point (hopefully soon!)

Issue #9163, opened 2025-02-28
Contributor ci is broken again


Wumpf:
See failures on https://github.com/rerun-io/rerun/pull/9152

PR #9261, opened 2025-03-11
Install nextest in contrib CI

abey79:
Contributor CI fails because we use nextest but dont install it. On regular CI, this is done by setup-rust, which we can't use for contributor CI since it relies on repo's secrets.


Wumpf:
Fixes 
* https://github.com/rerun-io/rerun/issues/9163

Wumpf (approved):

PR #9262, opened 2025-03-11
Fix `columns` for columns of multiple batch-length==1 fixed size arrays

Wumpf:
### Related

* Follow-up to https://github.com/rerun-io/rerun/pull/9210

### What

In other words, the following snippet broke (which got caught by now-red main ci! phew!)

```py
rr.send_columns(
    "box",
    indexes=[rr.IndexColumn("tick", sequence=range(1, 101))],
    columns=rr.Transform3D.columns(
        translation=[[0, 0, t / 10.0] for t in range(100)],
        rotation_axis_angle=[
            rr.RotationAxisAngle(axis=[0.0, 1.0, 0.0], radians=truncated_radians(t * 4)) for t in range(100)
        ],
    ),
)
```

`rotation_axis_angle` worked fine, but `translation` broke because the `columns` call incorrectly assumed that it's 100 columns (correct) each with a batch size of 3 elements (very wrong).

* [x] pass full-ci


Wumpf:
@rerun-bot full-check

Wumpf:
@rerun-bot full-check

abey79 (approved):

PR #9267, opened 2025-03-12
Convert `store_version` and `store_source` to record batches

### LINKED ISSUE #9190: Move `store_source` and `store_version` from `StoreInfo` into `archetypes::RecordingProperties`

grtlr:
### Related

* Closes #9190.

### What

> [!IMPORTANT]
> Sorry for the initial ugliness—but this should serve as a starting point for discussion about how fine granular we want to record information about the source of recordings in the future.

These fields are mainly used to show information in the viewer, as well as provide telemetry information. As such, I think we might be able to replace many of these entries with `Utf8`.


grtlr:
Closing because not actionable right now. The issue linked above provides more context.

PR #9269, opened 2025-03-12
Correct truncation of AnyValues when using strings or bytes

### LINKED ISSUE #8781: PythonAPI: Adding additional AnyValues with lists of strings sometimes give (empty) or truncated text
Could not find similar bug

**Describe the bug**
In PythonAPI: When appending `rr.AnyValues` with a list of strings to `Boxes3D` the values often show as (Empty).
It is a bit inconsistent when it happens, but in my experience it happens mostly when the list is created dynamically and could contain for example `None` values (even when these are converted to string) or when the length of the strings vary.
I have had similar issues with this python API where floats was converted to ints unless I added a 0.0 at the end of the list for example.

Image

**Workaround** I have found that appending one large string value could fix the problem. And depending on the length it could truncate the other strings if it is shorter. 
Example:

Image
From code:

Image

**To Reproduce**

Something like this:

Image

**Expected behavior**
I expected the values to show in the rerun viewer like this:

Image

**Backtrace**
No trace

 - Ubuntu 22.04.4 LTS

**Rerun version**
 0.20.0 and 0.21.0: Python API

**Additional context**
I cannot give my own code, but could try to give a fully reproducible code if there is interest in solving the issue.
My guess is that it is some typing issue (specifically with characters/string and floats) when the python code is sent to Rust.

timsaucer:
### Related

* Closes #8781

### What

This PR adjusts the way we process AnyValues. WIthout this change we have cases where numpy is setting data types to have fixed length based on the first value it receives. This ends up truncating the data.

Additionally, we cannot simply call `pa.array()` because types like strings and bytes are iterable and will get turned into an array of characters or bytes, respectively.

This PR attempts three passes at converting values

- Attempt to call `pa.array()` directly, but with a special case to ignore string and bytes
- Attempt to cast to a pyarrow Scalar and make an array from the scalar
- Fall back to handling with numpy, which handles a wide variety of mixes between lists, tuples, etc.

Added unit test to capture failure mode in attached issue.

* [x] pass full-check ci

### DIFF
--- a/rerun_py/rerun_sdk/rerun/any_value.py
+++ b/rerun_py/rerun_sdk/rerun/any_value.py
@@ -88,9 +88,11 @@ class AnyBatchValue(ComponentBatchLike):
                         except TypeError as e:
                             pass
                     if self.pa_array is None:
-                        pa_scalar = pa.scalar(value, type=pa_type)
-                        if pa_scalar is not None:
+                        try:
+                            pa_scalar = pa.scalar(value, type=pa_type)
                             self.pa_array = pa.array([pa_scalar], type=pa_type)
+                        except TypeError:
+                            pass
                     if self.pa_array is None:
                         # Fall back - use numpy
                         np_value = np.atleast_1d(np.asarray(value, dtype=np_type))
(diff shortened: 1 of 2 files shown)


timsaucer:
@rerun-bot full-check

Wumpf:
@timsaucer the full-check/`main` test suit highlighted an error when running the `structure_from_motion` example

Wumpf (approved):
code changes lgtm! But ci thinks otherwise (haven't digged into what's happening yet)

timsaucer:
I've identified that error and will get the commit updated today

timsaucer:
@rerun-bot full-check

PR #9272, opened 2025-03-13
Ignore `__init__.py` in snippets.toml

abey79:
Adding that `__init__.py` in #9260 made mypy happy but snippets sad. It's now explicitly ignored.


abey79:
@rerun-bot full-check

Wumpf (approved):

Issue #9254, opened 2025-03-11
Setting Visible and Interactive in blueprint overrides disables the corresponding GUI buttons


aecins:
**Describe the bug**
Setting the `Visible` and `Interactive` properties to `True` in blueprint overrides results in the corresponding buttons in the GUI being unmodifiable.

**To Reproduce**
Run the notebook below: https://colab.research.google.com/drive/1TkiWBL-lPCuq6MZ394T803hYuctPFHW5?usp=sharing 

**Expected behavior**

`Visible` and `Interactive` GUI buttons can be toggled on and off regardless of the blueprint override values.

**Rerun version**
rerun-cli 0.22.1 (default grpc map_view nasm native_viewer release web_viewer) [rustc 1.81.0 (eeb90cda1 2024-09-04), LLVM 18.1.7] x86_64-unknown-linux-gnu release-0.22.1 6b5a938, built 2025-02-20T10:00:25Z
Video features: av1 default ffmpeg nasm serde

Wumpf:
Thanks for the repro case!

Issue #7337, opened 2024-09-03
Blueprint API: Visibility per entity basis(python)


NickLaurenson-Visionick:
**Is your feature request related to a problem? Please describe.**

Currently it's not possible to set the visibility per entity in the python blueprint API. 
The current workaround is to use the per view `content` keyword to remove the data from the view.

**Describe the solution you'd like**

Maybe add a visibility keyword in the View API

**Describe alternatives you've considered**

**Additional context**

Mimic this functionnality from the UI:
![image](https://github.com/user-attachments/assets/d85655cc-2de0-42e5-a778-109d690bcea3)

Wumpf:
Associated discord thread https://discord.com/channels/1062300748202921994/1280068537666240584/1280068537666240584


Overrides of (individual) visibility is already possible, but slightly awkward and not documented:

```py
import rerun as rr
import rerun.blueprint as rrb

rr.init(
    "rerun_example_visibility_via_blueprint",
    spawn=True,
    default_blueprint=rrb.Spatial3DView(
        overrides={"points": [rrb.components.Visible(False)]}
    ),
)
rr.log("points", rr.Points3D([[0, 0, 0], [1, 1, 1]]))
```

Should document this and make it a bit more approachable maybe

kabouzeid:
Ideally, we could also set the default visibility during logging. Currently I'm logging many entity paths, and have to manually click 20+ times to hide the stuff I might want to look at on demand, but usually needs to be hidden.

Issue #6541, opened 2024-06-11
Allow logging visible & interative to datastore


Wumpf:
visible & interactive are regular components by now. The only thing prevent a user to log it straight to the data store is the fact that we only query for blueprint overrides now. Afaik all of it goes through the `is_visible` utility on `DataResult`, so it should be easy to fix.

While the interaction with the ui controls might be a bit surprising, it allow you to show/unshow things in a timedependent fashion

ChengHsiuChi:
I am looking for similar features. I want to log objects that are invisible by default unless manually toggled on. Currently, I log many signals into 2D views—while they can be useful together at times, I usually only need one or two key signals, not all at once. So, I’d like them to remain hidden by default.

emilk:
I ran into this when trying to hide the collision meshes of an URDF file

Wumpf:
@emilk how badly do you still need this for URDF or did you decide now to solve it with some kind of draw order?

emilk:
I decided not to solve it for now. It's fine to punt on.

PR #9281, opened 2025-03-13
New `EntityBehavior` archetype for easy `visible`/`interactive` blueprint overrides

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/9254
* Fixes https://github.com/rerun-io/rerun/issues/9251
* Fixes https://github.com/rerun-io/rerun/issues/7337

Related to, but does NOT address
* https://github.com/rerun-io/rerun/issues/6541

### What

Introduces a new archetype, `EntityBehavior`, which bundles `Visible` and `Interactive` component.

Since all overrides are now archetype based, this has been necessary in order to set overrides for these without additional overhead.

Sample snippet:
```py
"""Configure interactivity & visibility of entities."""

import rerun as rr
import rerun.blueprint as rrb

rr.init("rerun_example_entity_behavior", spawn=True)

# Use `EntityBehavior` to override visibility & interactivity of entities in the blueprint.
rr.send_blueprint(
    rrb.Spatial2DView(
        overrides={
            "hidden_subtree": rr.EntityBehavior(visible=False),
            "hidden_subtree/not_hidden": rr.EntityBehavior(visible=True),
            "non_interactive_subtree": rr.EntityBehavior(interactive=False),
        }
    )
)

rr.log("hidden_subtree", rr.Points2D(positions=(0, 0), radii=0.5))
rr.log("hidden_subtree/also_hidden", rr.LineStrips2D(strips=[(-1, 1), (1, -1)]))
rr.log("hidden_subtree/not_hidden", rr.LineStrips2D(strips=[(1, 1), (-1, -1)]))
rr.log("non_interactive_subtree", rr.Boxes2D(centers=(0, 0), half_sizes=(1, 1)))
rr.log("non_interactive_subtree/also_non_interactive", rr.Boxes2D(centers=(0, 0), half_sizes=(0.5, 0.5)))
```
image

Furthermore, `Visible` and `Interactive` components are now defined to always propagate recursively in the tree.

This cleans up an old artifact of blueprint overrides in the viewer: Previously, we would distiniguish "recursive overrides" and "individual overrides". The recursive kind was at this point exclusively used for visible and interactive. More so, the viewer would **never** actually use individual overrides for `visible`/`interactive`. This caused issues when introducing individual overrides via blueprint (see #9254). This PR **removes** recursive overrides entirely, which greatly simplifies the blueprint query logic - likely further optimizations & simplifications are unlocked by this!

Note that since all visibility configuration is recursive, it can be tedious to make only individual non-leave entities visible. However, I intentionally sticked with status quo, just simplifying how its modelled. With this override being much more readily accessible, I think we're in a much better spot to begin with here.

* [x] migration guide
* [x] full-check ci
* [x] move to blueprint namespace


nikolausWest:
What about calling it ViewBehavior to make it even more clear its about how the entity behaves in a view?

Wumpf:
Iirc `ViewBehavior` came up but we (@jleibs and me) discarded it because we felt that this would be a thing on the view itself, i.e. defining how the view as a hole would behave.

nikolausWest:
How about `BehaviorInViewer` to be even more on the nose?

```python
rr.log("/path/to/my/camera", rr.BehaviorInViewer(interactive=False, visible=True))
```

jleibs:
It's not strictly limited to the viewer. Esp if you log it that way... it's in the data. Like all archetypes the viewer does something with that data. We don't write RenderedBox2dInViewer just because that's what the viewer uses it for.

Additionally I like EntityBehavior because it describes the semantics. This visibility applies to *the whole entity* and may impact multiple archetypes even though it doesn't have those archetypes tags.

nikolausWest:
I know this thing ends up stored along with the data but my thinking is that these things are special in that we don't intend them to be used to model the data in any way. They really just describe what behavior the entity should have in the viewer.

nikolausWest:
Another one could be `VisualizerBehavior`: it configures the behavior of any visualizer on this entity. It doesn't hide the entity in general or affect what comes back in a dataframe query. It hides the resulting visualizer

jleibs:
> we don't intend them to be used to model the data in any way

Isn't that for users to decide?

nikolausWest:
> Isn't that for users to decide?

It feels wrong but I guess

jleibs:
The best signal of "we don't want users to think of this for modeling data" would be to stick it in the blueprint namespace instead. 

Introducing viewer-specific concepts into blueprint makes sense since the blueprint is specifically about semantic modeling of the viewer behavior.

nikolausWest:
> The best signal of "we don't want users to think of this for modeling data" would be to stick it in the blueprint namespace instead.

Fair enough, that makes sense.

Wumpf (commented):
notes from PR walkthrough with @abey79

crates/store/re_types/definitions/rerun/blueprint/archetypes/entity_behavior.fbs
```diff
@@ -0,0 +1,28 @@
+namespace rerun.blueprint.archetypes;
+
+/// General visualization behavior of an entity.
+///
+/// TODO(#6541): Fields of this archetype currently only have an effect when logged in the blueprint store.
+///
+/// \example archetypes/entity_behavior missing="cpp,rs"
```
abey79, line 7: I believe you need to slap an `image="https://static.rerun.io/....` argument here for this snippet to be live.

docs/content/reference/migration/migration-0-23.md
```diff
@@ -238,6 +238,45 @@ but for the moment the Viewer is not able to make this distinction.
 For details see [#6889](https://github.com/rerun-io/rerun/issues/6889).
 
 
+### Overriding `Visible` and `Interactive` is now always recursive
+
+Previously, it was possible to override visibility individually, but not recursively.
+Also, Viewer interaction [was hampered](https://github.com/rerun-io/rerun/issues/9254) by this.
+
+Overrides are for these two properties are now always recursive and can be done via the new `EntityBehavior` archetype.
```
abey79, line 246:
```suggestion
Overrides for these two properties are now always recursive, and can be applied using the new `EntityBehavior` archetype.
```

abey79 (approved):
Went over it all a second time and looks great. Such a nice clean-up and improvement!

Wumpf:
@rerun-bot full-check

Wumpf:
@rerun-bot full-check

PR #9282, opened 2025-03-14
Upload snapshot test results as artifact

lucasmerlin:
### What

Makes it easier to debug failing snapshots in CI.

See here for an example of a failed run: https://github.com/rerun-io/rerun/actions/runs/13853649550/job/38765622419?pr=9282

Click the "Upload test results" section to get a download link


Wumpf (approved):
nice!

Can we please also add this to the `contrib_checks.yml`?
Also this is quite hard to find if one isn't aware of it, can we have the error message give instructions on ci where to find these artifacts? 🤔

PR #9291, opened 2025-03-16
Rename `set_index` to `set_time`

emilk:
### Related

### What
* Pure renamings
* All the previous names were introduces since the last release
* Makes the three APIs more consistent with each other

Review commit by commit!

### Why keep "time" instead of "index"?
A) the word "index" name-clashes with another index in DP
B) changing "time" to "index" everywhere is a big breaking change
C) if we're renaming "time" to "index" we should go all the way (rename “timeline” etc too) or not at all
D) we would like to add more index types in the future, but realistically it won’t happen for a long time


docs/content/getting-started/data-out/analyze-and-send.md
```diff
@@ -46,7 +46,7 @@ Here is how to send the data as a scalar:
 ```python
 rr.send_columns(
     "/jaw_open_state",
-    indexes=[rr.IndexColumn("frame_nr", sequence=df["frame_nr"])],
+    times=[rr.TimeColumn("frame_nr", sequence=df["frame_nr"])],
```
grtlr, line 49: In the Rust API, `times` is still called `indexes`, so we should probably unify.

Nit: `times` sounds/reads a bit weird here to me. Maybe `time_columns`, although it's a bit longer?

docs/content/reference/migration/migration-0-17.md
```diff
@@ -41,7 +41,7 @@ For example, setting color and enabling the `SeriesPoint` visualizer was previou
 rr.log("data", rr.SeriesPoint(color=[255, 255, 0]), static=True)
 
 for t in range(1000):
-    rr.set_index("frame_nr", sequence=t)
+    rr.set_time("frame_nr", sequence=t)
```
grtlr, line 44: Probably search-and-replace error, but we should not change old migration guides.

docs/content/reference/migration/migration-0-22.md
```diff
@@ -263,7 +263,7 @@ radii = [0.05, 0.01, 0.2, 0.1, 0.3]
 
 rr.send_columns(
     "points",
-    indexes=[rr.TimeSecondsColumn("time", times)],
+    times=[rr.TimeSecondsColumn("time", times)],
```
grtlr, line 266: same as above

grtlr (changes requested):
Overall, I really like these changes to the API—just a small minor comments. The migration guide should be fixed though, everything else is a suggestion.

docs/content/reference/migration/migration-0-17.md
emilk: Same here - this is reverting a search-and-replace error in an earlier PR 🤦

emilk: Good catch!

emilk:
@rerun-bot full-check

grtlr (approved):

PR #9301, opened 2025-03-17
Initial python wrapper for the new catalog API

abey79:
### Related

- follow-up to #9299
- part of https://github.com/rerun-io/dataplatform/issues/400
- part of [#9360](https://github.com/rerun-io/rerun/issues/9360)

### What

This PR introduces the new catalog client API to the Python SDK. This is very much WIP, with the on-going todo list being captured in #9360. This PR primarily introduce some core patterns:

- How the connection is managed though a `ConnectionHandle`, which is owned by the root `CatalogClient` instance but cloneable at will by related instances and method implementation.
- An `Entry` hierarchy, with a `Dataset` sub-class, to model the object found in the catalog.
- Error management.


rerun_py/src/catalog/catalog_client.rs
```diff
@@ -0,0 +1,150 @@
...
+        let origin = re_uri::Origin::try_from(addr.as_str()).map_err(to_py_err)?;
+
+        let runtime = tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()?;
+
+        let connection = ConnectionHandle::new(origin.clone(), runtime)?;
+
+        Ok(Self { origin, connection })
+    }
```
timsaucer, line 39: Is this a problem if a user creates multiple catalog clients? I can see this happening, especially if someone is working in a notebook env during testing. I don't know if it's the best approach, but we worked around this in datafusion-python by creating a util to get the runtime handle [here](https://github.com/apache/datafusion-python/blob/main/src/utils.rs#L31). Also, `delta-rs` did a more thorough job [here](https://github.com/delta-io/delta-rs/blob/87010461cfe01563d91a4b9cd6fa468e2ad5f283/python/src/utils.rs#L10-L31)

rerun_py/src/catalog/catalog_client.rs
```diff
@@ -0,0 +1,150 @@
...
+            ))?;
+
+        Python::with_gil(|py| {
+            let entry_id = Py::new(
+                py,
+                entry_details
+                    .id
+                    .ok_or(MissingGrpcFieldError::new_err("No id in entry"))
+                    .map(PyEntryId::from)?,
+            )?;
```
timsaucer, line 135: Don't you already have `py` being passed in?

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -0,0 +1,107 @@
...
+/// Connection handle to a catalog service.
+#[derive(Clone)]
+pub struct ConnectionHandle {
+    #[expect(dead_code)]
+    origin: re_uri::Origin,
+
+    /// A tokio runtime for async operations. This connection will currently
+    /// block the Python interpreter while waiting for responses.
+    /// This runtime must be persisted for the lifetime of the connection.
+    runtime: Arc<tokio::runtime::Runtime>,
```
timsaucer, line 29: nit: Can use the tokio runtime handle instead of Arc<Runtime>. TBH I don't think it makes a huge difference but it's their preferred pattern

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -0,0 +1,107 @@
...
+    //TODO(ab): return nicer wrapper object over the gRPC message
+    pub fn find_entries(&mut self, filter: EntryFilter) -> PyResult<Vec<EntryDetails>> {
+        let response =
+            self.runtime
+                .block_on(self.client.find_entries(
+                    re_protos::catalog::v1alpha1::FindEntriesRequest {
+                        filter: Some(filter),
+                    },
+                ))
+                .map_err(to_py_err)?;
```
timsaucer, line 61: We seem to be doing a fair amount of blocking operations. Do we want instead to use async python functions? This is a question - I really don't know which is better for the use case right now.

rerun_py/src/catalog/errors.rs
```diff
@@ -0,0 +1,81 @@
...
+
+// Custom exception classes.
+create_exception!(catalog, MissingGrpcFieldError, PyValueError);
+
+// ---
+
+/// Private error type to server as a bridge between various external error type and the
+/// [`to_py_err`] function.
+#[expect(clippy::enum_variant_names)] // this is by design
+enum Error {
```
timsaucer, line 33: nit: I personally dislike when we just use `Error` and prefer something like `PyRerunError` just because I find it makes debugging messages easier to parse.

rerun_py/src/catalog/errors.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+/// Global mapping of all our internal error to user-facing Python errors.
+///
+/// Use as `.map_err(to_py_err)?`.
+#[expect(private_bounds)] // this is by design
+pub fn to_py_err(err: impl Into<Error>) -> PyErr {
+    match err.into() {
+        Error::ConnectionError(err) => PyConnectionError::new_err(err.to_string()),
```
timsaucer, line 63: If we implement `From<Error> for PyErr` then we should be able have this function return a `Error`. It just splits up the conversion to our `Error` and the mapping into a `PyErr` into two steps. I find it slightly easier to follow and the end effect to the code should be the same. That is, you should still end up with something like `.map_err(to_py_err)?` in the code

timsaucer (approved):
I have some suggestions, but overall I think this code looks very reasonable.

rerun_py/src/catalog/catalog_client.rs
abey79: I dont think this is a problem to have multiple tokio runtimes, but that data fusion approach looks good. I might give it a go.

rerun_py/src/catalog/connection_handle.rs
abey79: I tried and got into deadlock for some reason. I'll give it another go as per your other suggestions.

rerun_py/src/catalog/connection_handle.rs
abey79: I think there is value in having a very simple, friction-less blocking API, which is what this PR is mainly about.

Of course we'll have to sort out our concurrency story, starting with the dataframe transfer part. I can also see a future where we have alternative API such as multithreaded iterators or, indeed, async python. But i'd rather postpone that to after we sort out the basics.

rerun_py/src/catalog/errors.rs
abey79: I renamed it. Note that it should not be user visible though, since it's always mapped to a native python exception or another custom exception class.

PR #9307, opened 2025-03-17
Prevent modals from shrinking / moving around when the content changes

lucasmerlin:
### What

This prevents modals from shrinking if their content size changes, which is annoying when the size rapidly changes, e.g. during autocomplete.

The size will be reset if the screen size changes, so that the modal can adjust accordingly.

Question is wether the prevent_shrinking call should happen inside or outside the frame.

Outside:

https://github.com/user-attachments/assets/36c65743-1968-4a28-a68b-10e72ec73ba9

Inside:

https://github.com/user-attachments/assets/028b2230-7fa2-4561-a6bd-bd2ac8a5d579

As is, the PR implements the Outside behavior (which I think is nicer), switching it would be simple though.

@gavrelina which one do you prefer?


crates/viewer/re_ui/src/modal.rs
```diff
@@ -173,101 +173,105 @@ impl ModalWrapper {
...
-                ui.spacing_mut().item_spacing.y = 0.0;
-
-                if let Some(min_width) = self.min_width {
-                    ui.set_min_width(min_width);
+                prevent_shrinking(ui);
+                egui::Frame {
+                    fill: ctx.style().visuals.panel_fill,
+                    ..Default::default()
                 }
+                .show(ui, |ui| {
```
lucasmerlin, line 184: For the `Outside` behavior I had to draw the frame separately from the modal which is a bit annoying.

grtlr (changes requested):
Something seems to be a bit off (I don't know if this is intended though):

The initial list is quite long, and when filtering, I would expect it to shrink horizontally but keep its position.

Here is a video:

https://github.com/user-attachments/assets/7ed3bbb1-fd62-40b1-b275-1fa006c9ab0a

abey79:
> I would expect it to shrink horizontally while keeping its position

Well that's one of two option we must choose from, see the two videos at the top :)

lucasmerlin:
> The initial list is quite long, and when filtering, I would expect it to shrink horizontally while keeping its position. The way it is right now, there is a lot of wasted space.

Hmm, the initial list really is quite tall though, maybe we should also set a max height? 
Also I guess you mean shrink vertically?

crates/viewer/re_ui/src/modal.rs
```diff
@@ -335,3 +334,28 @@ fn view_padding_frame(params: &ViewPaddingFrameParams) -> egui::Frame {
...
     }
 }
+
+/// Prevent a UI from shrinking if the content size changes.
+///
+/// Should be called at the beginning of the [`egui::Ui`], before any other content is added.
+/// Will reset if the screen size changes.
+pub fn prevent_shrinking(ui: &mut egui::Ui) {
+    // The Uis response at this point will conveniently contain last frame's rect
+    let last_rect = ui.response().rect;
```
abey79, line 344: egui question: would it make sense to have a `last_rect: Option<Rect>` in `Response`?

abey79: Also, would it make sense to somehow have this entire utility function in egui?

abey79 (approved):
Code lgtm. FWIW, I think I prefer the command pallet behaviour vertically (shrinks with content from the bottom), but no shrinking horizontally. (Obviously subjective) rational: a vertical offset of the modal wrt to the outer window looks ok, but a lateral one doesn't.

abey79:
>  I think I prefer the command pallet behaviour vertically

I caught up with slack and saw Katya has the opposite opinion. She obviously wins.

> maybe we should also set a max height?

I think so, at least as a fraction of the screen height.

grtlr:
Yeah, sorry for the ambiguity, I meant shrinking in y-direction.

crates/viewer/re_ui/src/modal.rs
lucasmerlin: Hmm, I'm not sure if last_rect would be generally useful in the response, I could rather see having something like Ui::last_rect -> Option<Rect>, which I think would be much more useful.

> Also, would it make sense to somehow have this entire utility function in egui?

I thought about it but I wasn't sure where a good place would be. Maybe this could be a feature of Modal or Popup or Area, but I think Area is already complex enough, but if you have it on Modal you might want it on Popup and Window as well and then I just decided to add it to rerun for now haha

gavrelina:
OMG! I am so sorry about this ambiguity!! 
I actually agree with @abey79 and @grtlr but I misscalled the name of it in my slack comment (mixed up outside and inside)!

So I DO like when it shrinks TOGETHER with the missing content, and the mouse stays where it is AND with having a minimum height as well (not like in the original video, when with 0 results it completely shrinks vertically)

Sorry @lucasmerlin for this mistake!

lucasmerlin:
@grtlr can you have another look?

Issue #9131, opened 2025-02-25
Heuristically created blueprint overlaps encoded images


Wumpf:
According to [this report](https://discord.com/channels/1062300748202921994/1343727348683636796/1343727348683636796) encoded images are by default put into the same view rather than spawning separate views.

slyfox3:
I'm also seeing the same, in c++. Is this a regression?

slyfox3:
I realized if the image is logged with an pinhole model, the auto-layout puts each image into its own view. This could be an okay workaround, with an optional `with_image_plane_distance(.0f)` trick.

Wumpf:
This is independent of SDK languages as it's a bug in the viewer

PR #9308, opened 2025-03-17
Speed up 2d spawn heuristic for many entities & fix heuristic creating overlapping views for `EncodedImage`

Wumpf:
### Related

* Addresses comment in https://github.com/rerun-io/rerun/issues/8233#issuecomment-2676186007
* Fixes https://github.com/rerun-io/rerun/issues/9131

### What

Fixes excruciating slow heuristic for finding the right view roots for 2d images. This hit very hard for scenes with many entities containing images.

Fixed heuristic incorrectly handling `EncodedImage` in the process.

For testing I applied this fix to 0.22.1 and tried the RRD provided [here](https://github.com/rerun-io/rerun/issues/8233#issuecomment-2684545085) on a local release build. For my M1 Max Mac I get:

Before: `spawn_heuristics` for 2D takes 94ms 😱  
After:`spawn_heuristics` for 2D takes: 0.290ms

(branch with backported fix is `andreas/0.22.1/andreas/fix-encodedimage-spawn-heurist-and-make-image-view-heuristics-faster`)

### DIFF
--- a/crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
+++ b/crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
@@ -20,6 +20,7 @@ bitflags::bitflags! {
         const VIDEO = 0b10000;
     }
 }
+
 #[derive(Debug, Clone, Default)]
 pub struct MaxDimensions {
     pub width: u32,
--- a/crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
+++ b/crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
@@ -4,6 +4,7 @@ use once_cell::sync::OnceCell;
 use re_chunk_store::{ChunkStore, ChunkStoreSubscriberHandle, PerStoreChunkSubscriber};
 use re_log_types::{EntityPath, StoreId};
 use re_types::{
+    archetypes,
     components::{Blob, ImageFormat, MediaType},
     external::image,
     Component as _, Loggable as _,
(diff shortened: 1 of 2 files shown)


crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
```diff
@@ -13,6 +14,11 @@ use re_types::{
 pub struct MaxDimensions {
     pub width: u32,
     pub height: u32,
+
+    pub has_image: bool,
+    pub has_encoded_image: bool,
+    pub has_depth_image: bool,
+    pub has_video: bool,
```
emilk, line 21: segmentation image is conspicuously missing from this list.

crates/viewer/re_view_spatial/src/view_2d.rs
```diff
@@ -273,78 +270,55 @@ impl ViewClass for SpatialView2D {
...
-                    component_name,
-                )
-            })
-            .sum()
-    }
+#[derive(Default, Debug)]
+struct ImageCounts {
+    image: usize,
+    encoded_image: usize,
+    depth: usize,
```
emilk, line 277: Again, why is segmentation image missing?

emilk (approved):
🚀

crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
Wumpf: didn't need it later on (see how this ends up being used)
but I agree that on this thing alone this is just weeeeeird! Let's add it, don't think that hurts!

crates/viewer/re_view_spatial/src/view_2d.rs
Wumpf: leaving a comment on this one

Issue #9310, opened 2025-03-18
add back timezone to timestamp datatype


zehiko:
### What

We use ``IndexColumnDescriptor`` on the redap server side and currently Lance (TODO - issue) doesn't support storing a ``Timestamp`` data type with the timezone, although this is per arrow spec. 

We reverted adding timezone as part of https://github.com/rerun-io/rerun/pull/9311. Now need to add it back as soon Lance fixes this.

PR #9311, opened 2025-03-18
Revert adding timezone to timestamp datatype

zehiko:
### What

Temporary revert adding timezone to timestamp datatype while we don't have support for it on the ReDap server side.

Related to https://github.com/rerun-io/rerun/pull/9292

Follow up issue: https://github.com/rerun-io/rerun/issues/9310


emilk (approved):

emilk:
snapshot tests will fail

zehiko:
>  snapshot tests will fail

ah, right, I see them now in the last PR. fixing...

PR #9314, opened 2025-03-18
Better document required Python version

Wumpf:
Someone asked on Discord today what Python version we support and I noticed its documentation was a) hidden and b) wrong. Making things better here.
Also parts of CI apparently still ran with 3.8 it seems.


jleibs (approved):

Wumpf:
need to manually cherry-pick things without pixi.lock change to docs

PR #9315, opened 2025-03-18
Prevent automatic opening of views for archetypes in `RecordingProperties`

### LINKED ISSUE #9313: Prevent automatic opening of views for archetypes in `RecordingProperties`

grtlr:
### Related

* Closes #9313

### What

This PR adds `- /__partition_properties/**` to the filter list unless overwritten. This unifies our dataframe APIs again and is a first step to logging arbitrary data to properties.

### Test

To test, it's probably best to look at one of the `recording_properties.{cpp|py|rs}` snippets of the follow up PR:

* #9316

Here is a demo of the UI interactions:

https://github.com/user-attachments/assets/6158b77b-6200-453c-b5e8-80b9a854d25c


crates/store/re_log_types/src/path/entity_path_filter.rs
```diff
@@ -89,6 +89,9 @@ impl TryFrom<&str> for EntityPathFilter {
 /// The last rule matching `/world/car/hood` is `- /world/car/**`, so it is excluded.
 /// The last rule matching `/world` is `- /world`, so it is excluded.
 /// The last rule matching `/world/house` is `+ /world/**`, so it is included.
+///
+/// Unless otherwise specified, the [`ResolvedEntityPathFilter`] will filter out entities
+/// at [`EntityPath::partition_properties`].
```
Wumpf, line 94: That's not really a property of the `ResolvedEntityPathFilter` itself: if I were to create one without going through `EntityPathFilter` first (not possible right now iirc) then it would *not*. filter out `partition_properties`
I.e. it instead a property of `EntityPathFilter` to _produced_ such `ResolvedEntityPathFilter`s that it will filter out `partition_properties` if not specified otherwise

crates/store/re_log_types/src/path/entity_path_filter.rs
```diff
@@ -422,7 +429,20 @@ impl EntityPathFilter {
...
+            })
             .collect();
+
+        if !seen_properties {
+            rules.insert(
+                ResolvedEntityPathRule::including_subtree(&EntityPath::partition_properties()),
+                RuleEffect::Exclude,
+            );
+        }
+
```
Wumpf, line 445: so `resolve_strict`doesn't down that? Are we ok with that fact (when do we even use `resolve_strict`)? If so, then we should document that

Wumpf (changes requested):
When we discussed this I totally missed the fact that this would show up in the ui - it makes sense because the ui wants to use the ordering from the resolved filter, so it resolves and then for each resolved rule shows its "source" rule. I thiiiink I'm more happy than sad about this, since it makes things a lot more tractable for the user.
However, it breaks the dialog based ui a little bit: It's not possible to re-add the properties in here
image
in a similar fashion to how deleting the `-` rule for properties has no effect and shows up again. I'm ok with that behavior in the text box, but in the dialog it's a bit weird. Maybe just hide it for now? Otherwise it always shows up on top which isn't all that helpful anyways

crates/store/re_log_types/src/path/entity_path_filter.rs
grtlr: `resolve_strict` seems to be unused. Should I remove it?

grtlr: I've adapted it for now.

crates/viewer/re_selection_panel/src/view_entity_picker.rs
```diff
@@ -239,18 +239,46 @@ fn add_entities_line_ui(
...
 
-            if response.clicked() {
-                view.contents.remove_filter_rule_for(ctx, entity_path);
-            }
+                ui.add_enabled_ui(enabled, |ui| {
+                    let response = ui.small_icon_button(&re_ui::icons::ADD);
+
+                    if response.clicked() {
+                        view.contents.remove_filter_rule_for(ctx, entity_path);
+                        view.contents.raw_add_entity_inclusion(
```
grtlr, line 250: Wonder if we should this operation "atomically" under a single lock.

crates/store/re_log_types/src/path/entity_path_filter.rs
Wumpf: would have removed it 🤷

crates/viewer/re_view_graph/src/view.rs
```diff
@@ -94,17 +94,21 @@ impl ViewClass for GraphView {
         Default::default()
     }
 
-    fn spawn_heuristics(&self, ctx: &ViewerContext<'_>) -> ViewSpawnHeuristics {
+    fn spawn_heuristics(
+        &self,
+        ctx: &ViewerContext<'_>,
+        suggested_filter: &ResolvedEntityPathFilter,
+    ) -> ViewSpawnHeuristics {
+        // TODO(grtlr): Consider using `suggest_view_for_each_entity` here too.
```
Wumpf, line 102: good call not going there now-now :)

crates/viewer/re_view_spatial/src/heuristics.rs
```diff
@@ -28,7 +29,24 @@ pub fn default_visualized_entities_for_visualizer_kind(
...
+                }
+
                 let maybe_visualizable = ctx.maybe_visualizable_entities_per_visualizer.get(&id)?;
+
+                if maybe_visualizable
+                    .iter()
+                    .all(|e| suggested_filter.matches(e))
+                {
+                    return None;
+                }
```
Wumpf, line 48: confused by this. So if either all indicated or all maybe visualizers match are full contained in the `suggested_filter`, then we say no entity is visualized? Sounds rather strange to me, where do I read this wrong? I would have thought that instead we filter down the final flattend result with the `suggested_filter` at the end.

btw. this is the very place that makes me nervous perf wise. We hurl a lot of entities through this place and have one step more there. Well at least it's a bulk operation and the filter not very complex, I have hope that this can be optimized later if it should show up as an issue (plus the more general notion of "don't run all this things every frame 🤨 " which you're right we could use some nudge towards (don't even think it's terribly hard!) 😄

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -213,6 +217,15 @@ impl ViewClass for TimeSeriesView {
                 .extend(maybe_visualizable.iter().cloned());
         }
 
+        // Ensure we don't modify this list anymore before we check the `suggested_filter`.
+        let indicated_entities = indicated_entities;
```
Wumpf, line 221: neat trick, should do that more often

Wumpf (approved):
lgtm minus bugs you mentioned on slack (didn't test myself yet and didn't see anything obviously wrong yet except things commented on)

thought we land the filter change before going heuristics in a separate PR, but sure why not 🤷

crates/viewer/re_view_time_series/src/view_class.rs
grtlr: Ahaha, it's me being lazy and keeping the diff small. Introducing a scope would be even cleaner at the cost of an additional nesting level.

crates/viewer/re_view_spatial/src/heuristics.rs
grtlr: Yeah, that makes sense!

Issue #9199, opened 2025-03-05
Add `set_properties` for arbitrary data to all SDKs


grtlr:
Implementing the feature as is would be quite limited. For this to work more generally, we need either (or all of):

* Logging to sub-entities of `/__partition_properties`, here the question will be how to best visualize the tree.
* Allow multiple components of the same time (e.g. `start_time` & `end_time`) aka #6889
* The archetype pendant to `re_component_ui`, so that we can, for example show geo coordinates, or encoded images.

PR #9316, opened 2025-03-19
Provide APIs to log arbitrary data to recording properties

grtlr:
### Related

* Closes: #9199 

### What

This PR should cover all use-cases that we could come up with so far and be flexible enough to extend into the future without breaking changes. Here is the gist:
* We map a property name to an `EntityPath` that lives at `/__partition_properties/<property-name>`. Builtin recording properties (name, start time) will be at the special path `/__partition_properties/recording`.
  * Nothing lives directly at the root.
  * The current `RecordingProperties` archetype stays as is.
* The value of a property is always an archetype (potentially AnyValues). In the future we might open this up to specify general Arrow values.
* The API will be: `rec.send_property(name: str, archetype: AsComponent | [AsComponent])`
  * Motivation: `send_` signals that the properties are not guaranteed in a distributed setting.
  * `name` is positional, so we can upgrade it to reflect hierarchies later without breaking user code. (For example by renaming to `entity_path`).
* For now we don't expose the fact that this modeling allows for hierarchical grouping of properties.
* We will have to special methods to set the builtin properties:
  * `send_recording_name`
  * `send_recording_start_time_nanos`
* When creating a `RecordingStream` with an explicit recording ID, we don't send the builtin properties automatically, because that means we are usually in "append mode".

Open questions:
* The selection panel currently expects component-level data. Can we share the work for implementing something like re_archetype_ui with the catalog group so we could use that both in the tables and in the selection panel?


rerun_cpp/src/rerun/recording_stream.hpp
```diff
@@ -868,34 +870,87 @@ namespace rerun {
...
+        ///
+        /// \param entity_path The path to the recording property.
+        /// \param properties The properties of the recording.
+        /// \see `set_properties`
+        template <typename... Ts>
+        Error try_set_properties_with_prefix(std::string_view entity_path, const Ts&... properties)
+            const {
+            rr_error status = {};
+            this->log_static(
+                this->PARTITION_PROPERTIES_ENTITY_PATH + std::string(entity_path),
```
Wumpf, line 892: should add a slash here so user can specify `"thing"` as path and not break things

grtlr: Ah good catch!

rerun_cpp/src/rerun/recording_stream.cpp
```diff
@@ -319,19 +317,21 @@ namespace rerun {
         return status;
     }
 
-    Error RecordingStream::try_set_properties(archetypes::RecordingProperties& properties) const {
+    Error RecordingStream::try_set_name(std::string_view name) const {
         rr_error status = {};
-        this->log_static(PARTITION_PROPERTIES_ENTITY_PATH, properties);
+        this->log_static(
```
Wumpf, line 322: nit: `this->` for accessing members is imho not very idiomatic, but what is in C++ these days...

rerun_cpp/src/rerun/recording_stream.hpp
```diff
@@ -62,6 +59,11 @@ namespace rerun {
     /// The data will be timestamped automatically based on the `RecordingStream`'s
     /// internal clock.
     class RecordingStream {
+      private:
+        // TODO(grtlr): Ideally we'd expose more of the `EntityPath` struct to the C++ world so
+        //               that we don't have to hardcode this here.
+        inline static const std::string PARTITION_PROPERTIES_ENTITY_PATH = "__partition_properties";
```
Wumpf, line 65: nit:
```suggestion
       static constexpr const char PARTITION_PROPERTIES_ENTITY_PATH[] = "__partition_properties";
```

_even_ more const, also no startup alloc. That's what we use on string constants in archetypes etc.

rerun_cpp/src/rerun/recording_stream.hpp
```diff
@@ -868,34 +870,87 @@ namespace rerun {
...
+        ///
+        /// \param entity_path The path to the recording property.
+        /// \param properties The properties of the recording.
+        /// \see `set_properties`
+        template <typename... Ts>
+        Error try_set_properties_with_prefix(std::string_view entity_path, const Ts&... properties)
+            const {
+            rr_error status = {};
+            this->log_static(
+                this->PARTITION_PROPERTIES_ENTITY_PATH + "/" + std::string(entity_path),
```
Wumpf, line 892: ah right now with the constexpr suggestion above I broke this 🤦 
Can we just add the `/` so the constant and call it a day?

rerun_cpp/src/rerun/recording_stream.hpp
```diff
@@ -868,34 +870,87 @@ namespace rerun {
...
+        ///
         /// \param properties The properties of the recording.
-        /// \see `try_set_recording_properties`
-        void set_properties(archetypes::RecordingProperties& properties) const {
-            try_set_properties(properties).handle();
+        /// \see `try_set_properties`
+        template <typename... Ts>
+        void set_properties(const Ts&... properties) const {
+            // We need to split this into two arguments so that the compiler can bottom out.
+            try_set_properties_with_prefix("/", properties...).handle();
```
Wumpf, line 907: consequence of above is 
```suggestion
            try_set_properties_with_prefix("", properties...).handle();
```

rerun_cpp/src/rerun/recording_stream.hpp
```diff
@@ -868,34 +870,87 @@ namespace rerun {
...
+        /// This is a shorthand method for logging to the root of the recording properties.
+        ///
         /// \param properties The properties of the recording.
-        /// \see `set_recording_properties`
-        Error try_set_properties(archetypes::RecordingProperties& properties) const;
+        /// \see `set_properties`
+        template <typename... Ts>
+        Error try_set_properties(const Ts&... properties) const {
+            // We need to split this into two arguments so that the compiler can bottom out.
+            try_set_properties_with_prefix("/", properties...);
```
Wumpf, line 919:
```suggestion
            try_set_properties_with_prefix("", properties...);
```

Wumpf (commented):
mechanics lgtm.
I need to think a bit more about naming. Doesn't quite align with new slew of time namings iirc

grtlr:
@rerun-bot full-check

docs/snippets/all/concepts/recording_properties.cpp
```diff
@@ -2,15 +2,59 @@
...
-    auto properties = rerun::archetypes::RecordingProperties().with_start_time(0).with_name(
-        "My recording (initial)"
+    // Overwrites the name from above.
+    rec.send_recording_name("My recording");
+
+    // Start time is set automatically, but we can overwrite it at any time.
+    rec.send_recording_start_time_nanos(1742539110661000000);
+
+    auto points = rerun::Points3D({{1.0, 0.1, 1.0}});
+    rec.send_property("camera_left", points);
```
Wumpf, line 19:
```suggestion

    // Adds a user-defined property to the recording, using an existing Rerun type.
    auto points = rerun::Points3D({{1.0, 0.1, 1.0}});
    rec.send_property("camera_left", points);
```

docs/snippets/all/concepts/recording_properties.cpp
```diff
@@ -2,15 +2,59 @@
...
+
+    std::shared_ptr<arrow::Array> arrow_array;
```
Wumpf, line 21:
```suggestion
    // Adds another property, this time with user-defined data.
    {
          std::shared_ptr<arrow::Array> arrow_array;
          // ...
```

I think this helps guiding through the example a little bit just like you did in Rust & Python :)

docs/snippets/all/concepts/recording_properties.py
```diff
@@ -5,7 +5,27 @@
...
-rec.set_properties(rr.RecordingProperties(name="My recording (initial)", start_time=0))
-
 # Overwrites the name from above.
-rec.set_name("My recording")
+rec.send_recording_name("My recording")
+
+# Start time is set automatically, but we can overwrite it at any time.
+rec.send_recording_start_time_nanos(1742539110661000000)
+
+# Adds a user-defined property to the recording.
```
Wumpf, line 14:
```suggestion
# Adds a user-defined property to the recording, using an existing Rerun type.
```

docs/snippets/all/concepts/recording_properties.rs
```diff
@@ -1,16 +1,47 @@
...
-        &rerun::archetypes::RecordingProperties::new()
-            .with_start_time(0)
-            .with_name("My recording (initial)"),
+    // Recordings can have an optional name.
+    rec.send_recording_name("My recording")?;
+
+    // Start time is set automatically, but we can overwrite it at any time.
+    rec.send_recording_start_time(1742539110661000000)?;
+
+    // Adds a user-defined property to the recording.
```
Wumpf, line 16:
```suggestion
    // Adds a user-defined property to the recording, using an existing Rerun type.
```

rerun_cpp/src/rerun/recording_stream.hpp
```diff
@@ -9,6 +9,7 @@
 #include <vector>
 
 #include "as_components.hpp"
+#include "c/rerun.h"
```
Wumpf, line 12: It's on me to not add a lint for that, but I really don't want to import all the public c methods into the user's namespace (this does just that). This is because it has a lot of free functions that are all kinda hard to use and quite unsafe. May easily confuse a newcomer looking into their autocomplete. So far `#include "c/rerun.h"` is only ever included in cpp files for that reason.
Can you please add this to `rerun_cpp\tests\header_checks.cpp` (ping me on slack if unclear how) and avoid this include?

.. also don't see why its needed in the first place :)

rerun_py/rerun_sdk/rerun/_properties.py
```diff
@@ -45,4 +59,25 @@ def set_name(name: str, recording: RecordingStream | None = None) -> None:
...
+def send_recording_start_time_nanos(nanos: int, recording: RecordingStream | None = None) -> None:
+    """
+    Send the start time of the recording.
+
+    This timestamp is shown in the Rerun Viewer.
+
+    Parameters
+    ----------
+    nanos : int
+        The start time of the recording.
```
Wumpf, line 74: should document what kind of time this is as done on the other languages

crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -1160,20 +1163,32 @@ impl RecordingStream {
...
+    /// Sends the name of the recording.
     #[inline]
-    pub fn set_name(&self, name: impl Into<String>) -> RecordingStreamResult<()> {
+    pub fn send_recording_name(&self, name: impl Into<String>) -> RecordingStreamResult<()> {
         let update = RecordingProperties::update_fields().with_name(name.into());
-        self.set_properties(&update)
+        self.log_static(EntityPath::recording_properties(), &update)
+    }
+
+    /// Sends the start time of the recording.
```
Wumpf, line 1184: since integers implicitely cast into `Timestamp` this should specify what time it is.
We should consider refining this in a follow-up: `set_time` makes the unit fairly explicit by the kinds of conversions it allows

```
rec.set_time("frame_nr", rerun::TimeCell::from_sequence(42));
rec.set_time("duration", std::time::Duration::from_millis(123));
rec.set_time("capture_time", std::time::SystemTime::now());
```
We should be able to have the `Timestamp` type to work similar to that 🤔

Wumpf (approved):
some doc nits, I do feel strongly about `rerun.h` though (see comment). Otherwise all good, pre-approving with those to-be-done fixes!

docs/snippets/all/concepts/recording_properties.cpp
```diff
@@ -15,10 +15,14 @@ arrow::Status run_main() {
...
 
+    // Adds a user-defined property to the recording, using an existing Rerun type.
     auto points = rerun::Points3D({{1.0, 0.1, 1.0}});
     rec.send_property("camera_left", points);
 
-    std::shared_ptr<arrow::Array> arrow_array;
+    // Adds another property, this time with user-defined data.
+    {
+          std::shared_ptr<arrow::Array> arrow_array;
+          // ...
```
Wumpf, line 25: ah that was mean of me wasn't it. not a "real" gh suggestion. Would have acutally done a real one but it doesn't allow over code areas that have both additions and deletions

grtlr:
@rerun-bot full-check

grtlr:
@rerun-bot full-check

grtlr:
@rerun-bot full-check

PR #9320, opened 2025-03-19
DataFusion proof of concept TableProvider for catalog query

timsaucer:
### Related

Closes https://github.com/rerun-io/dataplatform/issues/407

### What

This PR demonstrates conversion of the GRPC queries for our catalog provider into a DataFusion `TableProvider`. With this a user can register this table provider into their DataFusion `SessionContext` and use it to perform dataframe operations.

Currently it does not support push down filtering, but this can be added soon.

Also we want to move this over from the legacy GRPC calls to the new design.


timsaucer:
Closing in favor of #9367

PR #9321, opened 2025-03-19
Fix `to_nanos` incorrectly handling `np.timedelta` 

oxkitsune:
### What

Fixed the following:
- `np.timedelta` has `np.integer` as superclass so we should do the instance check before that happens.
- `pa.timestamp("ns")` expects the values passed in to be of type `np.datetime64[ns]`
- `np.datetime` that contain non-nanosecond timestamps should first be converted to nanosecond scale.

I also added a bunch of unit tests for the `to_nanos` and `to_nanos_since_epoch` methods.

### DIFF
--- a/rerun_py/rerun_sdk/rerun/_send_columns.py
+++ b/rerun_py/rerun_sdk/rerun/_send_columns.py
@@ -102,7 +102,7 @@ class TimeColumn(TimeColumnLike):
         elif timestamp is not None:
             # TODO(zehiko) add back timezone support (#9310)
             self.type = pa.timestamp("ns")
-            self.times = [to_nanos_since_epoch(timestamp) for timestamp in timestamp]
+            self.times = [to_nanos_since_epoch(timestamp).astype("datetime64[ns]") for timestamp in timestamp]
 
     def timeline_name(self) -> str:
         """Returns the name of the timeline."""
--- a/rerun_py/rerun_sdk/rerun/_send_columns.py
+++ b/rerun_py/rerun_sdk/rerun/_send_columns.py
@@ -99,14 +99,14 @@ class TimeColumn(TimeColumnLike):
         elif duration is not None:
             self.type = pa.duration("ns")
             if isinstance(duration, np.ndarray):
-                self.times = duration.astype("timedelta64[ns]").tolist()
+                self.times = duration.astype("timedelta64[ns]").tolist()  # type: ignore[assignment]
             else:
                 self.times = [np.int64(to_nanos(duration)).astype("timedelta64[ns]") for duration in duration]
         elif timestamp is not None:
             # TODO(zehiko) add back timezone support (#9310)
             self.type = pa.timestamp("ns")
             if isinstance(timestamp, np.ndarray):
-                self.times = timestamp.astype("datetime64[ns]").tolist()
+                self.times = timestamp.astype("datetime64[ns]").tolist()  # type: ignore[assignment]
             else:
                 self.times = [
                     np.int64(to_nanos_since_epoch(timestamp)).astype("datetime64[ns]") for timestamp in timestamp
--- a/rerun_py/rerun_sdk/rerun/_send_columns.py
+++ b/rerun_py/rerun_sdk/rerun/_send_columns.py
@@ -99,9 +99,9 @@ class TimeColumn(TimeColumnLike):
         elif duration is not None:
             self.type = pa.duration("ns")
             if isinstance(timestamp, np.ndarray):
-                self.times = timestamp.astype("timedelta64[ns]").astype("int64").tolist()
+                self.times = timestamp.astype("timedelta64[ns]").tolist()
             else:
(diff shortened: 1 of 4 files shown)


rerun_py/rerun_sdk/rerun/time.py
```diff
@@ -108,35 +108,44 @@ def set_time(
...
     elif isinstance(duration, timedelta):
         return round(1e9 * duration.total_seconds())
-    elif isinstance(duration, np.timedelta64):
-        return duration.astype("timedelta64[ns]").astype("int64")  # type: ignore[no-any-return]
+    elif isinstance(duration, (int, np.integer)):
+        return 1_000_000_000 * int(duration)  # Interpret as seconds and convert to nanos
+    elif isinstance(
+        duration,
+        # Only allowing doubles since 32-bit and 16-bit floats lose precision when multiplying at this scale
+        (float, np.float64),
```
emilk, line 121: I think this should be up to the user.

For instance: f32 still has millisecond precision for durations under ~4 hours. Not great; not terrible.

rerun_py/rerun_sdk/rerun/time.py
```diff
@@ -108,35 +108,44 @@ def set_time(
...
+def to_nanos_since_epoch(timestamp: int | np.integer | float | np.float64 | datetime | np.datetime64) -> np.int64:
     if isinstance(timestamp, (int, np.integer)):
-        return 1_000_000_000 * int(timestamp)  # Interpret as seconds and convert to nanos
-    elif isinstance(timestamp, float):
-        return round(1e9 * timestamp)  # Interpret as seconds and convert to nanos
+        return np.int64(1_000_000_000 * int(timestamp))  # Interpret as seconds and convert to nanos
+    elif isinstance(
+        timestamp,
+        # Only allowing doubles since 32-bit and 16-bit floats lose precision when multiplying at this scale
+        (float, np.float64),
```
emilk, line 136:
```suggestion
        # Only allowing f64 since anything less has way too little precision for measuring time since 1970
        (float, np.float64),
```

rerun_py/rerun_sdk/rerun/_send_columns.py
```diff
@@ -102,7 +102,7 @@ def __init__(
         elif timestamp is not None:
             # TODO(zehiko) add back timezone support (#9310)
             self.type = pa.timestamp("ns")
-            self.times = [to_nanos_since_epoch(timestamp) for timestamp in timestamp]
+            self.times = [to_nanos_since_epoch(timestamp).astype("datetime64[ns]") for timestamp in timestamp]
```
emilk: reading this now it strikes me we should check for numpy-arrays here and in those cases do array operations instead of slow-as-turtles looping 😬  Perhaps leave it as a TODO…

emilk (approved):
Thanks for fixing this and adding unit tests 🙏

rerun_py/rerun_sdk/rerun/_send_columns.py
oxkitsune: Eh, can just do that real quick

rerun_py/rerun_sdk/rerun/time.py
```diff
@@ -108,35 +108,44 @@ def set_time(
         )
 
 
-def to_nanos(duration: int | float | timedelta | np.timedelta64) -> int:
-    if isinstance(duration, (int, np.integer)):
-        return 1_000_000_000 * int(duration)  # Interpret as seconds and convert to nanos
-    elif isinstance(duration, float):
-        return round(1e9 * duration)  # Interpret as seconds and convert to nanos
+def to_nanos(duration: int | np.integer | float | np.float64 | timedelta | np.timedelta64) -> int:
```
abey79, line 111: this would really benefit from being renamed `second_to_nanos` (assuming it's not published in the `rerun` namespace)

rerun_py/rerun_sdk/rerun/time.py
```diff
@@ -108,35 +108,44 @@ def set_time(
...
         return 1_000_000_000 * int(timestamp)  # Interpret as seconds and convert to nanos
-    elif isinstance(timestamp, float):
-        return round(1e9 * timestamp)  # Interpret as seconds and convert to nanos
+    elif isinstance(
+        timestamp,
+        # Only allowing f64 since anything less has way too little precision for measuring time since 1970
+        (float, np.float64),
+    ):
+        # Interpret as seconds and convert to nanos
+        return np.round(1e9 * timestamp).astype("int64")  # type: ignore[no-any-return]
```
abey79, line 138: low-key annoyed by these `# type: ignore` things. Did you ask Claude about it?

abey79 (approved):
Looks good to me!

rerun_py/rerun_sdk/rerun/time.py
emilk: it's not only from seconds though, it is also from `timedelta`

PR #9323, opened 2025-03-19
Use `RecordingEndpoint` in `SmartMessageSource::RedapGrpcStream`

lucasmerlin:
### Related
This in preparation for https://github.com/rerun-io/dataplatform/issues/396

### What

Changes `SmartMessageSource::RedapGrpcStream` to hold a RecordingEndpoint, which should make it more typesafe and convenient to use. This also allowed me to clean up the `CopyTimeRangeLink` implementation by using the `RecordingEndpoint` to format the url instead of doing it manually.

### DIFF
--- a/crates/utils/re_smart_channel/src/lib.rs
+++ b/crates/utils/re_smart_channel/src/lib.rs
@@ -50,7 +50,7 @@ pub enum SmartChannelSource {
     Stdin,
 
     /// The data is streaming in directly from a Rerun Data Platform server, over gRPC.
-    RedapGrpcStream { url: String },
+    RedapGrpcStream(re_uri::RecordingEndpoint),
 
     /// The data is streaming in via a message proxy.
     MessageProxy { url: String },
@@ -60,9 +60,8 @@ impl std::fmt::Display for SmartChannelSource {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Self::File(path) => path.display().fmt(f),
-            Self::RrdHttpStream { url, follow: _ }
-            | Self::RedapGrpcStream { url }
-            | Self::MessageProxy { url } => url.fmt(f),
+            Self::RrdHttpStream { url, follow: _ } | Self::MessageProxy { url } => url.fmt(f),
+            Self::RedapGrpcStream(endpoint) => endpoint.fmt(f),
             Self::RrdWebEventListener => "Web event listener".fmt(f),
             Self::JsChannel { channel_name } => write!(f, "Javascript channel: {channel_name}"),
             Self::Sdk => "SDK".fmt(f),
@@ -121,7 +120,7 @@ pub enum SmartMessageSource {
     Stdin,
 
     /// A file on a Rerun Data Platform server, over `rerun://` gRPC interface.
-    RerunGrpcStream { url: String },
+    RedapGrpcStream(re_uri::RecordingEndpoint),
 
     /// A stream of messages over message proxy gRPC interface.
     MessageProxy { url: String },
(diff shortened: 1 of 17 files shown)


crates/utils/re_smart_channel/src/lib.rs
```diff
@@ -121,7 +120,7 @@ pub enum SmartMessageSource {
     Stdin,
 
     /// A file on a Rerun Data Platform server, over `rerun://` gRPC interface.
-    RerunGrpcStream { url: String },
+    RedapGrpcStream(re_uri::RecordingEndpoint),
```
lucasmerlin, line 123: I've renamed this from `RerunGrpcStream` to match the variant in `SmartChannelSource`. Or was this naming intentional?

crates/utils/re_uri/src/endpoints/recording.rs
```diff
@@ -1,6 +1,6 @@
 use crate::{Origin, TimeRange};
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
```
lucasmerlin, line 3: serde is needed so we can use it in SmartChannelSource. 

Instead of deriving, should it maybe be a custom serialization that uses `Display` and `FromStr`?

abey79: I dont think we care about serialised representation.

crates/utils/re_smart_channel/src/lib.rs
abey79: Doubt it was. Thanks for that clean up.

crates/utils/re_uri/src/lib.rs
```diff
@@ -43,7 +43,7 @@ pub use self::{
     error::Error,
 };
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
```
grtlr, line 46: Nit: We usually put `serde` behind a `cfg` feature flag—would that make sense here?

grtlr (approved):
Nice improvement! I wanted to do this after landing the initial `re_uri` changes, but totally forgot. Thank you for picking this up!

crates/utils/re_uri/src/lib.rs
lucasmerlin: I didn't because the rest of re_uri also didn't 😬 I'll update it to be behind a cfg everywhere in the crate and make serde optional

abey79: TBH I'm not sure it's worth it. We'll never use `re_uri` is a serde-less environment.

abey79: (The general trend has been to minimise these cargo features as much as possible lately.)

grtlr: I was thinking about `redap-cli` - in general, on the viewer side yes, not sure about dataplatform. But I don't have any strong opinions (but I do prefer less features).

crates/viewer/re_viewer/src/ui/top_panel.rs
```diff
@@ -222,10 +222,12 @@ fn connection_status_ui(ui: &mut egui::Ui, rx: &ReceiveSet<re_log_types::LogMsg>
             }
             re_smart_channel::SmartChannelSource::Stdin => "Loading stdin…".to_owned(),
             re_smart_channel::SmartChannelSource::RrdHttpStream { url, .. }
-            | re_smart_channel::SmartChannelSource::RedapGrpcStream { url }
             | re_smart_channel::SmartChannelSource::MessageProxy { url } => {
                 format!("Waiting for data on {url}…")
             }
+            re_smart_channel::SmartChannelSource::RedapGrpcStream(endpoint) => {
+                format!("Waiting for data on {endpoint}…")
```
jprochazk, line 229: Just a thought: Do we want to display the full endpoint here, time range and all? We definitely want the protocol+host+path, but query params seem like noise to me. You can always see the full endpoint by selecting the recording.

crates/utils/re_uri/src/lib.rs
lucasmerlin: I've reverted it for the sake of less features

crates/viewer/re_viewer/src/ui/top_panel.rs
lucasmerlin: Good point, I've updated it

PR #9324, opened 2025-03-19
manifest registry service spec update: make partition id a string type

zehiko:
### What

Partition ids can be user defined fields (episode id, recording id, etc). For simplicity we now represents  in the manifest registry service with a  simple string type.  What actual field and type defines the partition id will most likely be defined as part of creating the ``Dataset``. 

This will also simplify the initial integration with ``rrd`` and the ``StoreId::id()``. 

We might revisit how we represent and deal with the partition ids e2e, but it will be simpler to do that as concrete cases and client integration comes along.

### DIFF
--- a/crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
+++ b/crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
@@ -89,7 +89,7 @@ message UnregisterPartitionsRequest {
   rerun.common.v1alpha1.DatasetHandle entry = 1;
 
   // Partitions to remove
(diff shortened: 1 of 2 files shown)


teh-cmc (changes requested):
Even if it's just a string, it deserves a dedicated type (in fact it's probably even more important it does _because_ it's just a string :grimacing:).

I suggest something like:
```proto
message PartitionId {
    optional string id = 1;
}
```

zehiko:
> Even if it's just a string, it deserves a dedicated type
 
ah, I had that initially, but then thought about layers of wrapping of ``Option``...anyways, will add it back.

PR #9325, opened 2025-03-19
Settle for microsecond precision in test

emilk:
This is because Rust's `SystemTime` has system-dependant accuracy:

https://doc.rust-lang.org/stable/std/time/struct.SystemTime.html#platform-specific-behavior


Wumpf (approved):
not using `SystemTime` is not really on the table I figure?
Well whatever, microseconds datetime should be enough for anyone amiright ;-)

emilk:
Since this is an example, I want to show you can use `SystemTime`.

Issue #9020, opened 2025-02-12
Make point series visualizer handle vectors of scalars per time index


Wumpf:
All the same things as
* https://github.com/rerun-io/rerun/issues/9019
but for points.

PR #9327, opened 2025-03-20
Enable scalar arrays per time point for scatter plots

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/9020

### What

Works mostly by ripping out a whole bunch of code from the `SeriesLineVisualizer` and re-use it in the `SeriesPointVisualizer`.

image

There's _so much_ that could be improved there and lots of duplication to be avoided. But I refrained falling down that rabbit hole - also I believe the "zipping up" into an array of structs is premature there to begin with, we should rather keep the arrays separate for longer (but without having tried it that's pure speculation whether this would be actually better since we do have to "zip up" things for egui plot eventually, also aggregation can profit from this setup).

Plots have decent image comparison test coverage already - in fact I already had a test for how arrays of scalars interact with point series, see updated images :)

### DIFF
--- a/crates/viewer/re_view/src/lib.rs
+++ b/crates/viewer/re_view/src/lib.rs
@@ -28,7 +28,8 @@ pub use query::{
     latest_at_with_blueprint_resolved_data, range_with_blueprint_resolved_data, DataResultQuery,
 };
 pub use results_ext::{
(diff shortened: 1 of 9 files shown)


abey79 (approved):
I'm a simple man. Moar pretty pictures, less SLOC => happy :) I agree more could be done, but that's a strict win already (in addition to the actual new feature).

PR #9329, opened 2025-03-20
Fix visibility flags for series line/point not splatting correctly

Wumpf:
### Related

* builds on https://github.com/rerun-io/rerun/pull/9327
* part of https://github.com/rerun-io/rerun/issues/9021

### What

the per series visibility didn't have the usual splatting behavior, making changes in code and ui more confusing than necessary


abey79 (approved):
I'm not 100% sure of the implications but I trust you and code lgtm :)

edit: is that something that should be captured in a test?

Wumpf:
hmm yeah I could add a test for that. If I don't get to it today I'll just merge it because I'm out next week and this is strictly better than without

PR #9333, opened 2025-03-20
Double clicking plots focuses them now in other panels

Wumpf:
Double click on a time series (line or scatter) will now not reset the view but instead perform the viewer's "focus" action - this causes scroll-to in blueprint & streams panels.
Furthermore, since this is now relevant due to multi-scalar per entity, double click selects now the entire entity rather than an instance (similar to double clicking a point cloud point)

https://github.com/user-attachments/assets/da195f8b-4958-4deb-b1a3-4743732cd0b2

### DIFF
--- a/crates/viewer/re_view_time_series/src/view_class.rs
+++ b/crates/viewer/re_view_time_series/src/view_class.rs
@@ -65,6 +65,9 @@ pub struct TimeSeriesViewState {
     /// (e.g. to avoid `hello/x` and `world/x` both being named `x`), and this knowledge must be
     /// forwarded to the default providers.
     pub(crate) default_names_for_entities: HashMap<EntityPath, String>,
+
+    /// Whether to reset the plot bounds next frame.
+    reset_bounds_next_frame: bool,
 }
 
 impl Default for TimeSeriesViewState {
@@ -72,10 +75,16 @@ impl Default for TimeSeriesViewState {
         Self {
             is_dragging_time_cursor: false,
             was_dragging_time_cursor: false,
-            saved_auto_bounds: Default::default(),
+            saved_auto_bounds: egui::Vec2b {
+                // Default x bounds to automatically show all time values.
+                x: true,
+                // Never use y auto bounds: we dictated bounds via blueprint under all circumstances.
+                y: false,
+            },
             scalar_range: [0.0, 0.0].into(),
             time_offset: 0,
             default_names_for_entities: Default::default(),
+            reset_bounds_next_frame: false,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


jprochazk:
Is this the expected behavior? When I zoom out and double-click, it "expands" the plot to fit the zoom level. Resetting to default blueprint doesn't undo this. I don't know if this is the intended behavior, but it doesn't seem right

https://github.com/user-attachments/assets/192ce347-c329-4443-bb16-a590cba254ec

Wumpf:
Good catch, I clearly broke something there. It even shows in the video I posted on the PR, totally missed that 🤦

Wumpf:
fixed & updated description video

jprochazk:
Am I supposed to be able to select a single instance of e.g. the scatter plot in the plots example? I can't seem to be able to do that. But that doesn't work on `main` either. Double-clicking seems to just do the same thing as single-click.

As for the zooming behavior, it seems to work now. There is still one hiccup, but not sure how actionable it is. If I zoom into the scatter plot, and double click without moving my mouse, the scatter plot stretches like in my original video. After moving my mouse, it goes back to normal. This is in the web viewer, I have not tested the native viewer. Zooming out works well.

https://github.com/user-attachments/assets/356697a8-5c76-443a-8913-3a9f2d61db30

jprochazk:
Weirdly enough, the zoom-in into double-click issue is the exact same on `main`, but there it only happens for a single frame.

https://github.com/user-attachments/assets/091089f8-8915-4139-8bab-bcb4faed953e

jprochazk (approved):
Code-wise LGTM, not sure how actionable the other stuff is.

Wumpf:
> Am I supposed to be able to select a single instance (/point) of e.g. the scatter plot in the plots example? I can't seem to be able to do that. But that doesn't work on main either.

Known issue, that's the same problem we have when clicking a point in a point cloud with time series ranges: selections aren't time line aware

> Double-clicking seems to just do the same thing as single-click.

That shouldn't be the case though. Single click shouldn't perform the focus action. Regarding what is selected  they are expected to be the same iff you have a single series. So I think we're talking about the same thing, but the time aspect is indeed confusing 🤔 

> If I zoom into the scatter plot, and double click without moving my mouse, the scatter plot stretches like in my original video. After moving my mouse, it goes back to normal. This may just be a missing request_redraw somewhere.

Uhoh. Want to have a look at this, your assessment sounds about right. Thank you!

Wumpf:
fixed the delayed adjustment bug with another repaint. sucks a bit that that is needed, but this will do for now

PR #9335, opened 2025-03-20
Fix rust analyzer in VSCode not running properly on test files

Wumpf:
Rust analyzer wouldn't check any files in tests directories properly for me and others using VSCode.

Removing the override commands for check (and for build for good measure) fixes the issue for me: local testing showed that explicit target-dir specification there doesn't seem to be necessary (we have a different target dir for rust analyzer so one can run bacon/clippy/builds while rust analyzer is performing checks)

Also I removed the removed `invocationStrategy` options (they no longer had any effect)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,7 +16,7 @@
     // Whether `--workspace` should be passed to `cargo check`. If false, `-p <package>` will be passed instead.
     "rust-analyzer.check.workspace": false,
     "rust-analyzer.cargo.allTargets": true,
-    // --all-features will set the `__ci` feature flag, which stops re_web_viewer_server/build.rs from building the web viewer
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -18,26 +18,10 @@
     "rust-analyzer.cargo.allTargets": true,
     // --all-features will set the `__ci` feature flag, which stops re_web_viewer_server/build.rs from building the web viewer
```
Wumpf, line 19: actually I don't think that's true anymore, need to check on this

grtlr (approved):
This makes the tests work on my side, and so far I have not encountered any downsides.

My main concern is this interfering with codegen, but we add `rust-analyzer.cargo.targetDir": "target_ra` so 🤷.

If you're unsure, maybe tag someone else too.

Wumpf:
hmmm yolo, can't be worse 😄

Issue #7343, opened 2024-09-03
Time Series reset view issue when hiding data via the view ui


NickLaurenson-Visionick:
**Describe the annoyance**

Resseting the view in a Time Series doesn't work when some data are hidden via the View. It behave as if the data is still visible.

When we hide the data via the blueprint panel, it work as espected.

**To Reproduce**

1. Run this script
```python
"""Log a scalar over time."""

from math import cos, sin, tau

import rerun as rr

rr.init("rerun_example_series_line_style", spawn=True)

# Set up plot styling:
# They are logged as static as they don't change over time and apply to all timelines.
# Log two lines series under a shared root so that they show in the same plot by default.
rr.log("trig/sin",
       rr.SeriesLine(color=[255, 0, 0], name="sin(0.01t)", width=2), static=True)
rr.log("trig/cos",
       rr.SeriesLine(color=[0, 255, 0], name="cos(0.01t)", width=4), static=True)

# Log the data on a timeline called "step".
for t in range(0, int(tau * 2 * 100.0)):
    rr.set_time_sequence("step", t)

    rr.log("trig/sin", rr.Scalar((sin(float(t) / 100.0)+100)))
    rr.log("trig/cos", rr.Scalar(cos(float(t) / 100.0)))

```
2. Click here:
![image](https://github.com/user-attachments/assets/36c9c091-799a-4a11-9c34-6c2f1ffe3ab8)

4. double click on the View to reset the camera

**Expected behavior**

**Desktop (please complete the following information):**
 - OS: ubuntu 22.04

Wumpf:
Strongly related to:
* https://github.com/rerun-io/rerun/issues/4974

PR #9353, opened 2025-03-21
Fix time series display range being affected by invisible plots

Wumpf:
### Related

* https://github.com/rerun-io/rerun/issues/7343

### What

Reset plot view before:
image

After:
image


abey79 (approved):
thanks! that behaviour was really annoying

PR #9359, opened 2025-03-24
redap_jobs: introduce grpc definitions

andrea-reale:
### Related

* Part of [redap #276 ](https://github.com/rerun-io/dataplatform/issues/276)

### What

This PR introduces gRPC definitions for the service that will enable interaction with redap persistent jobs. See [redap #276 ](https://github.com/rerun-io/dataplatform/issues/276).


crates/store/re_protos/proto/rerun/v1alpha1/jobs.proto
```diff
@@ -0,0 +1,50 @@
+syntax = "proto3";
+
+package rerun.redap_jobs.v1alpha1;
+
+import "rerun/v1alpha1/common.proto";
+
+// JobsService is the service for submitting and querying persistend redap tasks.
```
teh-cmc, line 7:
```suggestion
// JobsService is the service for submitting and querying persistent redap tasks.
```

crates/store/re_protos/proto/rerun/v1alpha1/jobs.proto
```diff
@@ -0,0 +1,50 @@
...
+service JobsService {
```
teh-cmc, line 8: It seems like the notion of a job doesn't exist at all anymore, should this be a `TasksService`?

crates/store/re_protos/proto/rerun/v1alpha1/jobs.proto
```diff
@@ -0,0 +1,50 @@
...
+  rpc Query(QueryRequest) returns (QueryResponse);
+}
+
+message SubmitTasksRequest {
+  repeated Task tasks = 1;
+}
+
+message SubmitTasksResponse {}
+
+// A task is the unit of work in a tasks.
```
teh-cmc, line 21: ?

crates/store/re_protos/proto/rerun/v1alpha1/jobs.proto
```diff
@@ -0,0 +1,50 @@
...
+message Task {
+  // Unique identifier for the task
+  string task_id = 1;
+  // Type of the task
+  string task_type = 2;
+  // Task-type dependant data necessary de-serialize the task
```
teh-cmc, line 27:
```suggestion
  // Task-type dependent data necessary to de-serialize the task
```

crates/store/re_protos/proto/rerun/v1alpha1/jobs.proto
```diff
@@ -0,0 +1,50 @@
...
+    QueryAllTasksRequest query_all_tasks = 1;
+    // Filter tasks by id
+    QueryFilterTasksRequest query_filter_tasks = 2;
+  }
+}
+
+message QueryAllTasksRequest {}
+message QueryFilterTasksRequest {
+  repeated string task_ids = 1;
+}
```
teh-cmc, line 44: `oneof` can create a lot of usability issues down the line -- my advice is to avoid it when it's not absolutely necessary.

Dropping it will also remove a bunch of typing layers in this instance:
```suggestion
// QueryRequest is the request message for querying tasks status
message QueryRequest {
  // If set, unconditionally fetch all tasks, always.
  bool fetch_all = 1;

  repeated string task_ids = 2;
}
```

At which point I would suggest to simply do this:
```suggestion
// QueryRequest is the request message for querying tasks status
message QueryRequest {
  // Empty queries for all tasks, if the server allows it, otherwise returns an error.
  repeated string task_ids = 1;
}
```

crates/store/re_protos/proto/rerun/v1alpha1/jobs.proto
andrea-reale: sloppy find and replace

PR #9362, opened 2025-03-24
Move opt-in of recording properties rule to `ViewContent`

### LINKED ISSUE #9351: Unify `view_content` behavior across UI and SDK
Currently, we show the explicitly ignored partition properties in the `view_contents` of the UI. This has several problems:

* It looks noisy to have something in red always show up.
* All our usages should be unified, and right now properties show up in the UI, while they are automagically added for things like dataframe queries.

For now, we should just hide (filter out) the recording properties everywhere.

grtlr:
### Related

* Closes #9351 

### What

This moves the opt-in logic for querying recording properties down one level to the `ViewContents`. This should unify our handling of that query in the UI as well as across our different SDKs (dataframe, ...).


grtlr:
Superseded by #9366.

PR #9364, opened 2025-03-24
Change `DeleteDatasetEntry` to `DeleteEntry` in both the protos and the Python SDK

### LINKED ISSUE #9344: Rename `DeleteDatasetEntry` to `DeleteEntry` in protos
No need to have `DeleteDataset`, `DeleteTable`, etc. All of these have an `EntryId`, so they can use a single `DeleteEntry` call.

abey79:
### Related

- Fixes #9344
- Sibling https://github.com/rerun-io/dataplatform/pull/444

### What

All entries should be deletable, so it was a mistake to make it specific to dataset in the first place.

### DIFF
--- a/crates/store/re_protos/proto/rerun/v1alpha1/catalog.proto
+++ b/crates/store/re_protos/proto/rerun/v1alpha1/catalog.proto
@@ -10,9 +10,11 @@ import "rerun/v1alpha1/common.proto";
 
 service CatalogService {
   rpc FindEntries(FindEntriesRequest) returns (FindEntriesResponse) {}
+  rpc DeleteEntry(DeleteEntryRequest) returns (DeleteEntryResponse) {}
+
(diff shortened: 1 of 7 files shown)


crates/store/re_protos/proto/rerun/v1alpha1/catalog.proto
```diff
@@ -27,6 +28,14 @@ message FindEntriesResponse {
   repeated EntryDetails entries = 1;
 }
 
+// DeleteDatasetEntry
+
+message DeleteEntryRequest {
+  rerun.common.v1alpha1.Tuid id = 1;
```
teh-cmc, line 34: I would prefer if we had an actual `EntryId` type. Our total number of ID things has quadrupled in the last weeks, a lot of things are Tuids now...

teh-cmc (approved):

jleibs (approved):
Let's handle the EntryId topic separately. It probably makes sense, but all the nesting is so annoying in practice esp in these cases where it's unambiguous from context.

PR #9365, opened 2025-03-24
introduce the `Frontend` service

teh-cmc:
Introduces the `Frontend` service, which is the public gRPC service for the entire platform.

It does very little beyond basic auth and proxying at the moment, except for one very important thing: it resolves dataset IDs into physical dataset handles for downstream services that don't have this information.
It also acts as the last encapsulation barrier: all the services downstream are effectively private and thus have less sensitive to breaking changes.

On the DP side, this service is implemented by the `redap_frontend` crate, which we expect will do a lot more in the future (e.g. auth, tls termination, rerun-specific load-balancing, even support other protocols such as Arrow Flight or raw HTTP).

* Implementation: https://github.com/rerun-io/dataplatform/pull/445


crates/store/re_protos/proto/rerun/v1alpha1/frontend.proto
```diff
@@ -0,0 +1,69 @@
...
+// --- Manifest Registry ---
+
+message RegisterPartitionsRequest {
+  rerun.common.v1alpha1.Tuid dataset_id = 1;
+
+  // Partitions to add
+  repeated rerun.manifest_registry.v1alpha1.Partition partitions = 2;
+
+  // what to do if partition is already registered
+  // TODO(cmc): why is this in manifest_registry??
```
teh-cmc, line 50: @teh-cmc

crates/store/re_protos/proto/rerun/v1alpha1/frontend.proto
```diff
@@ -0,0 +1,69 @@
...
+
+import "rerun/v1alpha1/catalog.proto";
+import "rerun/v1alpha1/common.proto";
+import "rerun/v1alpha1/manifest_registry.proto";
+
+// TODO(cmc): I've decided to re-use the underlying response types for now, in the name of expediency.
+// This is bad in the long term (breaking changes to the internal APIs should not result in breaking
+// changes to the public API and, in general, the two being coupled doesn't make much sense).
+// It's the very early days though, we know we're going to completely break all of this in the coming
+// weeks anyway.
```
teh-cmc, line 13: Then again, this will be less effort than fighting our lints so... let's do it?

teh-cmc:
Surely the test-wheels failure is unrelated...

andrea-reale (approved):

PR #9367, opened 2025-03-24
DataFusion queries against new catalog

timsaucer:
### Related

Closes https://github.com/rerun-io/dataplatform/issues/407

### What

This PR includes methods for turning both request/response and streaming services from our GRPC definitions into DataFusion table providers. It expands the previous work and supersedes https://github.com/rerun-io/rerun/pull/9320

This code creates two different traits for conversion of either single response or streams of data into table providers.


Wumpf:
@timsaucer I reckon this is outdated by now and should be closed?

PR #9377, opened 2025-03-26
Improve local <> remote flow

lucasmerlin:
### Related

* Part of https://github.com/rerun-io/dataplatform/issues/396

### What

Groups items in the recordings panel by their dataset / locally by app id / examples


crates/viewer/re_viewer/src/ui/recordings_panel.rs
```diff
@@ -152,90 +149,227 @@ fn recording_list_ui(
         );
     }
 
-    for (app_id, entity_dbs) in entity_dbs_map {
-        app_and_its_recordings_ui(ctx, ui, &app_id, entity_dbs);
+    let title =
+        |title| re_ui::list_item::LabelContent::new(egui::RichText::new(title).size(11.0).strong());
```
abey79, line 153: nit: if you turn that into a a fn, it could accept `impl Into<String>` and you could skip some `.to_owned()` later

crates/viewer/re_viewer/src/ui/recordings_panel.rs
```diff
@@ -152,90 +149,227 @@ fn recording_list_ui(
...
     ui: &mut egui::Ui,
-    app_id: &ApplicationId,
+    kind: &DatasetKind,
     mut entity_dbs: Vec<&EntityDb>,
 ) {
     entity_dbs.sort_by_key(|entity_db| entity_db.recording_property::<Timestamp>());
 
-    let app_item = Item::AppId(app_id.clone());
-    let selected = ctx.selection().contains_item(&app_item);
+    // TODO(lucasmerlin): Can we select a dataset?
```
abey79, line 312: Not yet. Soon, the redap browser should make the dataset selectable and show its TOC though.

abey79 (approved):
Nice!

crates/viewer/re_viewer/src/ui/recordings_panel.rs
lucasmerlin: Added this as a helper to LabelContent. Also added ListItem::header which adds a matching y offset

crates/viewer/re_ui/src/list_item/list_item.rs
```diff
@@ -287,10 +307,16 @@ impl ListItem {
...
+            mut height,
+            y_offset,
             render_offscreen,
         } = self;
 
+        if y_offset != 0.0 {
+            ui.add_space(y_offset);
+            height -= y_offset;
+        }
+
```
abey79, line 319: The result of this is the gap, visible eg here:

image

I wonder if things would not look better if we kept the original height for the highlight.

An alternative would be to not have full-width highlight for those titles, but some other kind of highlight (to suggest they are [sometime] clickable). I think i would prefer that, but then it becomes a design issue, so probably out of scope for this pr.

abey79 (approved):

PR #9381, opened 2025-03-26
Implement basic UI to display recording properties

### LINKED ISSUE #9188: Create dedicated UI for recording properties

grtlr:
### Related

* Closes #9188

### What

This crates a basic _Properties_ section in the `re_selection_panel` to show the recording properties.

> [!IMPORTANT]
> The truncation in the screenshot below will be handled by:
>
> * #9391

image


crates/viewer/re_data_ui/src/entity_db.rs
```diff
@@ -227,5 +213,32 @@ impl crate::DataUi for EntityDb {
                 }
             }
         }
+
+        ui.section_collapsing_header("Properties").show(ui, |ui| {
```
abey79, line 217: This has a weird interaction with the top-level `Data` collapsing header:

Collapse properties:
image

Collapse Data:

image

(I'd expect "Properties" to still be visible)

My  suggestion would be to NOT add this here, but instead add it to `re_selection_panel` when the recording is selected.

abey79 (commented):
I strongly suggest to move properties to the selection panel instead of the `EntityDb` data ui.

abey79 (approved):
Looks good, thanks!

PR #9338, opened 2025-03-20
replace panic with proper error messages for some codegen paths

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/9022

### What

* [ ] migration guide
* [ ] describe what's going on here

PR #9384, opened 2025-03-26
Python/CI: Raise exception on warning, e.g. call of @deprecated

emilk:
### Related
* https://github.com/rerun-io/rerun/pull/9338

### What
We do not want to call deprecated functions from our examples and snippets (it sets a bad example!).

Unfortunately there is nothing stopping us from doing so right now.
Until this PR.

With `PYTHONWARNINGS="error"` _any_ Python `warning` results in an exception, which should fail the example/snippet.

### TODO
* [x] `@rerun-bot full-check`

### DIFF
--- a/.github/workflows/reusable_test_wheels.yml
+++ b/.github/workflows/reusable_test_wheels.yml
@@ -195,7 +195,7 @@ jobs:
       - name: Check for release checklist
         if: ${{ inputs.FAST }}
         # Only check that the release checklist executes successfully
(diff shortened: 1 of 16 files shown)


timsaucer (approved):
LGTM so long as CI all passes

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

PR #9388, opened 2025-03-26
fix: typo in apps-and-recordings.md

dslemusp:
### What

Small typo in apps and recordings section md.


dslemusp:
I close this because I thought there was a typo but it turns out the meaning is correct, so my bad!

Issue #9391, opened 2025-03-26
Truncations in `DataUi` for `EntityDb`


grtlr:
* Requires #9381

**Describe the bug**

For some reason, the `DataUi` for `EntityPath` truncates the component names, as shown in the screenshot below.

Note that this only happens when there are multiple entity paths of recording properties.

**To Reproduce**
1. Run the `recording_properties` snippet.
2. Select the recording and go to the selection panel.
3. Truncation does not happen on DNA for instance.

**Expected behavior**

Proper truncation (if required) should be applied

**Screenshots**

<img width="355" alt="image" src="https://github.com/user-attachments/assets/a1d30aa4-8564-4f9c-a4bc-9cb043084d70" />

**Additional context**

Internal discussion: https://rerunio.slack.com/archives/C04MTSM2U91/p1742991946191259

PR #9392, opened 2025-03-26
Ensure that `component_list_ui` uses unique id for its `list_item_scope`

abey79:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/9391
* Exposed by https://github.com/rerun-io/rerun/pull/9381

### What

The new property UI would call multiple `EntityPath::data_ui` sequentially. These would lead to `component_list_ui` to be called, and because they had a hard-coded id, there would be cross-talk between the scopes' underlying egui memory storage. This PR makes the id unique.


grtlr (approved):
Oh, nice catch! 🤩

PR #9400, opened 2025-03-27
Add `detach_process` option to `spawn`

imstevenpmwork:
### Related Discussion  
For more context, see this thread:  
https://discord.com/channels/1062300748202921994/1354497232103145642  
Related to: https://github.com/rerun-io/rerun/commit/1942b43262252e1d0c887623c30451c0cdd6b9ad

### What’s New  
This update introduces a `detach_process` option for the `spawn()` method in the `Viewer` (available in Rust, Python, and C++).  

### Why It Matters  
This gives client code control over whether the `Viewer` runs independently of the parent process or remains tied to it. Keeping it attached ensures the `Viewer` closes when the parent process terminates with `SIGINT`. Detaching (default behavior) allows it to persist if needed.

While this achieves the desired functionality, a future improvement could be adding a proper `close()` method to gracefully shut down the `Viewer` context.  

### Note on Web Support  
This change wasn’t applied to the web API since I'm not sure how process management behaves there.

### How it was tested?
(Python only) Small edit to the `minimal.py` to evaluate the behavior

```python3
#!/usr/bin/env python3
"""Demonstrates the most barebone usage of the Rerun SDK."""

import sys

import numpy as np
import rerun as rr  # pip install rerun-sdk
import time


def main() -> None:
    # sanity-check since all other example scripts take arguments:
    assert len(sys.argv) == 1, f"{sys.argv[0]} does not take any arguments"

    rr.init("rerun_example_minimal")
    rr.spawn(detach_process=False)

    positions = np.vstack([xyz.ravel() for xyz in np.mgrid[3 * [slice(-10, 10, 10j)]]]).T
    colors = np.vstack([rgb.ravel() for rgb in np.mgrid[3 * [slice(0, 255, 10j)]]]).astype(np.uint8).T

    rr.log("my_points", rr.Points3D(positions, colors=colors, radii=0.5))
    time.sleep(10)  # CRTL+C here: Viewer process will exit


if __name__ == "__main__":
    main()
```

As expected, when `detach_process=False` if we send a `SIGINT` to the application, the Viewer exits and the PID assigned to it is retired.


emilk (commented):
Looks great so far!

emilk:
@rerun-bot full-check

emilk:
The CI failures seems unrelated

Issue #9005, opened 2025-02-12
Project: Plotting of "signals" style data


Wumpf:
## Summary description

Allow sending/logging of multi-dimensional data and plot them without additional data transformation by the user.

## Background and problem

Many datasets use state vectors to capture robot state. For example lerobot uses a 7-dimensional vector of motor positions as their state.

Today, users have to break these up into separate columns/entities is somewhat awkward compared to just being able to log and plot using a ListArray (or more direct API) instead.

## Grounding use-cases & evaluation

* Hugging face loader
* IMU data
* other things our design partners may hit?
* TODO: fill in more

## Outline & Vision

Previous proposals were focused around introducing new purpose-built archetypes, see https://github.com/rerun-io/rerun/issues/7278. A number of issues were identified with this line of thinking:
* lots of code duplication, including existing technical debt
* type explosion
* very hard to name - depending on your area this may be about "signals" or "scalars" or "vectors" etc.

Taking a step back, our actual goal is for users to be able to supply arbitrary data and then (explicit via blueprint or by automated heuristic) set up a a visualizer for their data.

In accordance with the vision of a strong archetype-visualizer correlation (see https://github.com/rerun-io/rerun/issues/8368), we therefore should strive for a more flexible model where archetypes describes the visualization, but do not (overly) constrain how you present your data.

Concrete this means that both `SeriesLines` and `SeriesPoints` _visualizers_ should be able to deal with more arbitrary kinds of "scalars" - different dimensions & different types.
They source the data from an appropriately tagged component

Note that we long planned to make component tags overwriteable in the viewer (see general tagged component issue https://github.com/rerun-io/rerun/issues/6889). Fully implemented this should allow us to...
* present the "data"/scalar component as input of the respective visualizers == field on the archetype
* deprecate the existing `Scalar` archetype but (!) still pick its contents up in a `SeriesLines` visualizer by means of heuristically overwriting the tag such that the visualizer consumes the component array
    * open question: how far can this go, can we pick up arbitrary "untagged" data? Likely! But requires more complex heuristic. If this is instead easy to configure (from code & viewer) this should not be necessary.

## Areas & first steps

Thanks to now landed eager serialization, we can easily present the viewer with much more complex data than the API specification presents as "valid". This means in turn that we can make visualizers deal with data that is arbitrarily more complex and deal with API changes completely separately.

Therefore we can split this project into three _mostly_ separate tracks:

TODO: many of these bullet points should be sub-issues

[...]

PR #9402, opened 2025-03-27
Use multi-dimensional scalar values when loading LeRobot dataset

### LINKED ISSUE #9318: Use multi scalar in lerobot
should fit right in! :)

* Blocked on https://github.com/rerun-io/rerun/pull/9338

oxkitsune:
### Related

* Closes #9318 

### What

We now use the multi-scalar support added in #9005 when loading multi-dimensional action/observation values from a LeRobot dataset.
This is the preferred as it aligns with how data is consumed during training, each frame’s actions or observations are expected as a single tensor rather than a collection of separate scalar values. Storing it as one neat array makes it easier to query action/observation values back out of Rerun, as there's now just a single entity path.

### DIFF
--- a/crates/store/re_data_loader/src/loader_lerobot.rs
+++ b/crates/store/re_data_loader/src/loader_lerobot.rs
@@ -571,6 +571,7 @@ fn make_scalar_batch_entity_chunks(
 
     Ok(chunks.into_iter())
 }
+
 fn make_scalar_entity_chunk(
     entity_path: String,
     timelines: &IntMap<TimelineName, TimeColumn>,
--- a/crates/store/re_data_loader/src/loader_lerobot.rs
+++ b/crates/store/re_data_loader/src/loader_lerobot.rs
@@ -511,7 +511,7 @@ fn load_scalar(
                 ))
             })?;
 
-            let sliced = slice_scalar_array(feature_data).with_context(|| {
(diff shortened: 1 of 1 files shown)


crates/store/re_data_loader/src/loader_lerobot.rs
```diff
@@ -485,6 +486,8 @@ fn load_scalar(
             format!("Failed to get field for feature {feature_key} from parquet file")
         })?;
 
+    let entity_path = field.name().to_string();
```
emilk, line 489: I think it would be better to explicit here: do we want slashes in the field name to produce separators in the entity hierarchy?

Chose between `EntityPath::from_single_string` and `EntityPath::parse_forgiving`

crates/store/re_data_loader/src/loader_lerobot.rs
```diff
@@ -528,7 +531,7 @@ fn load_scalar(
 }
 
 fn make_scalar_batch_entity_chunks(
-    field: &Field,
+    entity_path: String,
```
emilk, line 534: strongly typed > stringly typed

```suggestion
    entity_path: EntityPath,
```

emilk (commented):
Nice! Please add little more to the PR description. What does this mean for our users?

emilk (approved):
Nice! Please add little more to the PR description. What does this mean for our users?

PR #9413, opened 2025-03-27
Update redap browser to use the new catalog API

abey79:
Title. This implementation is somewhat quick-and-dirty: the _entire_ list of datasets and their partition table is downloaded at once, without per-dataset granularity. This will have to be improved in the future, but that at least unblocks improving the UI and removing the legacy API.

*Note*: The new url scheme used in the dataset table is not yet supported, so it breaks the ability to load partition to the viewer. (I'll work on that next.)


lucasmerlin (approved):
Awesome!

PR #9415, opened 2025-03-27
Fix optional notebook package import

jprochazk:
`rerun_notebook` is an optional dependency. Right now, we attempt to import `rerun_notebook` unconditionally, and ignore the `ImportError` if there is one. It turns out this is not enough, because the `rerun_notebook` package is attempting to load a file which does not exist, which raises `FileNotFoundError` instead.

We can assume the widget always exists if the package is imported, so I updated the logic to actually check if the package is installed, and only attempt to import it if so.

Testing:
```
rm -rf .pixi/envs/examples
pixi run py-build-examples
pixi run -e examples minimal
```
Running this on main results in a `FileNotFoundError`, which is gone on this PR.


rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -7,6 +7,7 @@
 from typing import TYPE_CHECKING, Any, Literal
 
 import numpy as np
+import pkg_resources
```
abey79, line 10: I think `pkg_resource` is part of `setuptools` on which we don't explicitly depend. Any chance you could do the same with `importlib`?

jprochazk: Seems like `importlib.util.find_spec` does the same thing

PR #9416, opened 2025-03-27
Add new gRPC definitions related to Tables

jleibs:
### What

This is a very minimal generic Table API. It gives us the schema and lets us scan the table.


crates/store/re_protos/proto/rerun/v1alpha1/catalog.proto
```diff
@@ -103,3 +116,30 @@ message DatasetEntry {
...
+  SYSTEM_TABLE_KIND_ENTRIES = 2;
+}
+
+message SystemTable {
+  SystemTableKind kind = 1;
+}
+
+message LanceTable {
+  optional string dataset_url = 1;
+}
```
abey79, line 145: Is this used at all for now? Maybe skip until its used?

abey79 (approved):
nice

PR #9417, opened 2025-03-28
Introduce datafusion bindings

jleibs:
### Related

Depends on:
- https://github.com/rerun-io/rerun/pull/9416

### What

I hijacked @timsaucer 's previous work from:
- https://github.com/rerun-io/rerun/pull/9367

This now goes through a generic TableEntry listing with both schema and table provided by the server.


ARCHITECTURE.md
```diff
@@ -170,6 +170,7 @@ Update instructions:
 | Crate           | Description                                                     |
 | --------------- | --------------------------------------------------------------- |
 | re_dataframe    | The Rerun public data APIs.                                     |
+| re_datafusion   | DataFusion interfaces to Rerun gRPC queries                     |
```
abey79, line 173: The architecture schema needs to be updated as well.

crates/store/re_datafusion/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+    frontend::v1alpha1::frontend_service_client::FrontendServiceClient,
+};
+use table_entry_provider::TableEntryProvider;
+use tonic::transport::Channel;
+
+pub mod grpc_streaming_provider;
+pub mod partition_list;
+pub mod table_entry_provider;
+
+pub struct DataFusionConnector {
```
abey79, line 19: nit: i'm growing increasingly uncomfortable with `lib.rs`/`mod.rs` containing anything else that published symbols.

crates/store/re_datafusion/src/grpc_streaming_provider.rs
```diff
@@ -0,0 +1,166 @@
...
+use futures_util::StreamExt as _;
+use tokio_stream::Stream;
+
+#[async_trait]
+pub trait GrpcStreamToTable:
+    std::fmt::Debug + 'static + Send + Sync + Clone + std::marker::Unpin
+{
+    type GrpcStreamData;
+
+    async fn create_schema(&mut self) -> SchemaRef;
```
abey79, line 25: this need to return a `Result`

abey79 (commented):
What is the status of this PR? I would like it to land but there seems to be some unfinished business. Here are some comments anyways.

emilk:
If you're hitting this error:

> Internal error: Min/Max accumulator not implemented for type Duration(Nanosecond)

Then patch `datafusion` like we do in
* https://github.com/rerun-io/dataplatform/pull/494/

crates/store/re_datafusion/src/datafusion_connector.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use crate::partition_list::PartitionListProvider;
+use crate::table_entry_provider::TableEntryProvider;
+use datafusion::{catalog::TableProvider, error::DataFusionError};
+use re_log_types::external::re_tuid::Tuid;
+use re_protos::{
+    catalog::v1alpha1::{ext::EntryDetails, DatasetEntry, EntryFilter, ReadDatasetEntryRequest},
+    frontend::v1alpha1::frontend_service_client::FrontendServiceClient,
+};
+use tonic::transport::Channel;
```
abey79, line 11: nit

```suggestion
use datafusion::{catalog::TableProvider, error::DataFusionError};
use tonic::transport::Channel;

use re_log_types::external::re_tuid::Tuid;
use re_protos::{
    catalog::v1alpha1::{ext::EntryDetails, DatasetEntry, EntryFilter, ReadDatasetEntryRequest},
    frontend::v1alpha1::frontend_service_client::FrontendServiceClient,
};

use crate::partition_list::PartitionListProvider;
use crate::table_entry_provider::TableEntryProvider;
```

crates/store/re_datafusion/src/partition_list.rs
```diff
@@ -0,0 +1,82 @@
...
+        frontend_service_client::FrontendServiceClient, ScanPartitionTableRequest,
+    },
+    manifest_registry::v1alpha1::ScanPartitionTableResponse,
+};
+use tonic::transport::Channel;
+
+use crate::grpc_streaming_provider::{GrpcStreamProvider, GrpcStreamToTable};
+
+#[derive(Debug, Clone)]
+pub struct PartitionListProvider {
```
abey79, line 26: Shouldn't this be named `PartitionTableProvider`?

crates/store/re_datafusion/src/partition_list.rs
```diff
@@ -0,0 +1,82 @@
...
+        Ok(Arc::new(Schema::new_with_metadata(
+            vec![
+                Field::new("id", Tuid::arrow_datatype(), true),
+                Field::new("name", DataType::Utf8, true),
+                Field::new("entry_type", DataType::Int32, true),
+                Field::new("created_at", DataType::Int64, true),
+                Field::new("updated_at", DataType::Int64, true),
+            ],
+            HashMap::default(),
+        )))
```
abey79, line 56: Would be nice to directly wire that to the existing `GetPartitionTableSchema` rpc (and disable it in favour of the hardcoded schema if that rpc happens to not be implemented yet)

crates/store/re_datafusion/src/table_entry_provider.rs
```diff
@@ -0,0 +1,84 @@
...
+        frontend_service_client::FrontendServiceClient, GetTableSchemaRequest, ScanTableRequest,
+        ScanTableResponse,
+    },
+};
+use tonic::transport::Channel;
+
+use crate::grpc_streaming_provider::{GrpcStreamProvider, GrpcStreamToTable};
+
+#[derive(Debug, Clone)]
+pub struct TableEntryProvider {
```
abey79, line 23: inclined to go verbose and call this `TableEntryTableProvider`. My first reaction was "oh you mean EntryTableProvider" (which is not what this thing actually is).

crates/store/re_datafusion/src/grpc_streaming_provider.rs
```diff
@@ -0,0 +1,166 @@
...
+use futures_util::StreamExt as _;
+use tokio_stream::Stream;
+
+#[async_trait]
+pub trait GrpcStreamToTable:
+    std::fmt::Debug + 'static + Send + Sync + Clone + std::marker::Unpin
+{
+    type GrpcStreamData;
+
+    async fn create_schema(&mut self) -> Result<SchemaRef, DataFusionError>;
```
abey79, line 25: I'm also tripped up by this name. The schema is not created, since it exists _somewhere_ to begin with, right?. So more like `get_schema` or `fetch_schema` (I prefer the latter, indicates that the schema is to be retrieved by the grpc stream originating service).

rerun_py/src/dataframe.rs
```diff
@@ -1462,3 +1468,24 @@ pub fn load_archive(path_to_rrd: std::path::PathBuf) -> PyResult<PyRRDArchive> {
...
+        &self,
+        py: Python<'py>,
+    ) -> PyResult<Bound<'py, PyCapsule>> {
+        let capsule_name = cr"datafusion_table_provider".into();
+
+        let runtime = get_tokio_runtime().handle().clone();
+        let provider = FFI_TableProvider::new(Arc::clone(&self.provider), false, Some(runtime));
+
+        PyCapsule::new(py, provider, Some(capsule_name))
+    }
```
abey79, line 1490: Is there a strong reason not to implement that directly on `PyTable`, so we dont have to go through a `datafusion_provider()` call?

cc @timsaucer

abey79 (approved):
This is looking really good! Hoping to see it landed by (my) tomorrow :)

crates/store/re_datafusion/src/partition_list.rs
jleibs: Still waiting. Added a TODO though.

rerun_py/src/dataframe.rs
jleibs: Made the change -- I like it.

abey79: Nice. Is there a reason to leave `PyDataFusionTable` behind or is that an oversight?

Issue #9336, opened 2025-03-20
Hide the redap browser UI


abey79:
The redap browser is far from being mature enough, so it should be feature-flagged away for the 0.23 release.

PR #9418, opened 2025-03-28
Unify redap browser and recording panel

lucasmerlin:
### Related

* Closes https://github.com/rerun-io/dataplatform/issues/396
* Closes https://github.com/rerun-io/rerun/issues/9336

### What

After the initial implementation wasn't as nice to use as we hoped (the recording panel ui changing when clicking a dataset was disorienting) I've re-done the PR with a more unified implemenation where the redap browser is integrated in the recording panel. 

- redap servers / datasets are now always shown in the recording panel, and open recodings are listed in each dataset
- state of open dataset/server is now stored via DisplayMode
- add new Item::RedapEntry and Item::RedapServer
- don't add a feature flag but instead show a warning in add server modal
- removes server/local tab
- if a server was added, the left panel will be shown when rerun starts

Screenshot 2025-04-02 at 11 17 22

https://github.com/user-attachments/assets/087e3ffa-6dbe-499a-86a6-c5c4da6f2c9f


crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -166,6 +174,26 @@ impl RedapServers {
...
+        for server in self.servers.values() {
+            if let Some(dataset) = server.find_dataset(EntryId { id: entry_id }) {
+                return Some(dataset);
+            }
+        }
+
+        None
+    }
+
+    pub fn select_entry(&self, entry_id: re_tuid::Tuid) {
```
lucasmerlin, line 187: Should this be a Tuid or a re_protos::EntryId (seems weird) or should we create a new EntryId(re_tuid::Tuid) struct somewhere?

abey79: We can use `EntryId` from `rerun.common.v1alpha1.ext` (note the `ext`). It' converts to tuid and the actual proto one as required.

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -69,11 +72,16 @@ impl Server {
     }
 }
 
+pub enum Selection {
+    Dataset(EntryId),
+    Server(re_uri::Origin),
+}
+
```
abey79, line 79: For consistency with naming in the rest of the viewer, I'd call this structure `Item` (similar to `re_viewer_context::Item`). Also, we can use `Entry` here, as tables are coming soon-ish.

```suggestion
pub enum Item {
    Entry(EntryId),
    Server(re_uri::Origin),
}

```

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -128,6 +136,7 @@ impl Default for RedapServers {
 }
 
 pub enum Command {
+    SelectServer(re_uri::Origin),
     SelectEntry(EntryId),
```
abey79, line 140: Might consider merging these into `SelectItem(Item)`

abey79 (approved):
Thanks! Code generally lgtm–some minor comments here and there. Also, as discussed earlier, the workflow still has much room for improvement. Would be great to discuss with @gavrelina the next step for that.

lucasmerlin:
I think we should now be able to get rid of the welcome_screen application id, but I'll do that in a follow up PR

abey79 (approved):
code lgtm. I haven't been able to test is because the server fails to load for me right now, but I dont expect it to be related to this PR.

PR #9419, opened 2025-03-28
Switch to bufbuild instead of vscode-proto3

jleibs:
Our recommended tooling for formatting protobufs is `buf`.

`vscode-proto3` was behaving inconsistently.


emilk (approved):
Nice

PR #9422, opened 2025-03-28
Introduce URI scheme for dataset and add support for opening remote recordings using the new API

abey79:
Introduce the URL scheme for downloading dataset data:

```
$ORIGIN/dataset/$DATASET_ID/data?partition_id=$PARTITION_ID[&time_range=$TIME_RANGE]
```

Relevant slack thread for the backing discussion: https://rerunio.slack.com/archives/C08EWSCS26P/p1743092658188489

For now, we only support fetching a single partition, so the `partition_id` parameter is mandatory. In the future, we will support arbitrary queries.  The `time_range` parameter is optional. If used, the time range will be highlighted, but filtering on actual time range when fetching chunk is not yet supported (https://github.com/rerun-io/dataplatform/issues/474).

NOTES:
- Review commit-by-commit. The first commit is a pure move refactor.
- I've updated the existing `Recording` tests instead of adding new ones, because this stuff will soon be removed anyways.


crates/store/re_data_source/src/data_source.rs
```diff
@@ -229,13 +229,50 @@ impl DataSource {
...
+                        re_log::warn!(
+                            "Error while streaming {endpoint}: {}",
+                            re_error::format_ref(&err)
+                        );
+                    }
+                });
+                Ok(StreamSource::LogMessages(rx))
+            }
+
+            Self::RerunGrpcStream(re_uri::RedapUri::DatasetData(endpoint)) => {
```
abey79, line 275: GitHub diff is misleading here. The actual new code is after 🙄

crates/utils/re_uri/src/endpoints/dataset.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub time_range: Option<TimeRange>,
+}
+
+impl std::fmt::Display for DatasetDataEndpoint {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}/dataset/{}?partition_id={}",
+            self.origin, self.dataset_id, self.partition_id
+        )?;
```
jleibs, line 27: nit: even though `partition_id` is currently required structuring the same was as the time-range portion:
```
{
    write!(f, "?partition_id={}", self.partition_id)?;
}
```
will make it a bit easier when we eventually decide to make `partition_id` optional and retrieve the full dataset.

jleibs (approved):

Issue #9139, opened 2025-02-26
Remove `RemoteRecording` API in favor of an improved `Recording` API


teh-cmc:
## Context

We have two Recording APIs right now: `Recording` and `RemoteRecording`:
* `Recording` is backed by a local `ChunkStore` that is fed from an `rrd` file and exposes the local Dataframe API.
  * All the data is stored locally.
  * All the processing in done locally.
  * The returned data is just pointers into shared memory.
* `RemoteRecording` is backed by a remote dataplatform that exposes the remote Dataframe API.
  * All the data is stored remotely (where it is loaded in its entirety into the memory of the server!).
  * All the processing in done remotely.
  * The returned data is a full blown copy, in order to be sent through the network.

There are a lots of problems with the remote codepath:
1. Every API needs to be reimplemented twice, with all the gRPC glue etc. This is already an issue, `RemoteRecording` only covers a very small portion of `Recording` because we don't have time to implement both.
1. All data needs to be loaded into the server's memory, which is a multi-tenant thing, so that's very problematic.
1. All data needs to be copied, which means that features like sparse-filling that are free locally, become extremely expensive in a remote setup.
1. This goes against our general design of: "the dataplatform returns the relevant chunks for a query, which get fetched by your local chunkstore, which is then in charge of processing your query" (which is, for example, how visualization itself is gonna work!).
1. A lot more that I'm forgetting right now.

We *will* need some form of remote processing at some point (e.g. your query requires 200GiB of RAM to compute, but returns a single row), but that is likely far off into the future, and meanwhile it just creates a lot of problems for us.

## What to do

We need to make it possible to feed the `Recording` API from a remote source, e.g.: `rec = Recording("rerun://myplatform", query)`, where `query` can be any supported manifest query.

Doing so will fetch all the relevant chunks from the dataplatform, and load them into a local `ChunkStore`. All the usual APIs are then available from there.

cc @jleibs  @nikolausWest

nikolausWest:
Sounds great!

PR #9423, opened 2025-03-28
Remove legacy API protos and their use, including the legacy Python API

abey79:
### Related

- Sibling: https://github.com/rerun-io/dataplatform/pull/480

### What

Remove the legacy proto and the legacy server Python wrapper. This means that `PyRecordingHandle` can now only by a local recording (which will have to be fixed with he new api anyways, see https://github.com/rerun-io/rerun/issues/9139)


rerun_py/src/remote.rs
```diff
@@ -1,3 +1,6 @@
+// TODO(rerun-io/dataplatform#405), TODO(rerun-io/dataplatform#400): remove this (now disabled) file
+// when the API migration is completed.
+
```
abey79, line 3: I've excluded this file from the project but I think keping it "as reference" for the duration of the migration is useful.

Issue #9427, opened 2025-03-31
Classify each type as either stable or unstable


emilk:
Each type (archetype, component) should require a `attr.docs.state` that has to have one of the following values:

* `unreleased` (replaces `attr.docs.unreleased`)
* `unstable` (released, but may change significantly or be removed)
* `stable` (can change, but will do so in a way that old data still works)
* `deprecated` (replaces `attr.rerun.deprecated`)

By _requiring_ this attribute, we have to make a conscious choice. It also solves the problem of forgetting `attr.docs.unreleased` (leading to dead links until the next release).

`unstable` will be applied to:
* `ViewCoordinates`
* `AnnotationContext`

PR #9431, opened 2025-03-31
Introduce `attr.docs.state` to our codegen

emilk:
### Related
* Part of https://github.com/rerun-io/rerun/issues/9427

### What
To make it easier to manage the "stabilization" state of our types, I've introduced a new attribute (that replaces several old ones).

My plan is to make this mandatory for all archetypes and components, with the benefits:
A) force us to consider its state when adding a new type
B) make sure we don't forget to set it to `unreleased` (to add `?speculative-link`)

I will make it mandatory in a future PR though, to keep this PR smaller.

### Docs for `attribute "attr.docs.state";`
Must be set on all archetypes and components.

Must have one of the following values:
- `unreleased`
- `experimental`
- `unstable`
- `stable`
- `deprecated`

### `unreleased`
Any external links referencing this datatype will be given a speculative link marker.
Speculative links contain a `?speculative-link` query param. Any such links are ignored by linkinator,
and we check for their absence as the first step in our release process.

### `experimental`
New API that is not yet fully supported, and may change significantly in future versions.
Implies everything that `unstable` implies.

### `unstable`
Used for types that are likely to be removed or changed significantly,
and in a way that the data won't be backwards compatible.

### `stable`
Used for types that are unlikely to be removed or changed significantly.
If they are changed, we will make sure that the old data can still be read.

### `deprecated`
Marks something as deprecated followed by a (mandatory!) migration note.

If specified on an object (struct/enum/union), it becomes deprecated such
that using the object should emit a warning in all target languages.
Furthermore, documentation will mention that the object is deprecated and display
the specified migration note.

If set, you MUST also set `attr.docs.deprecated_since` and `attr.rerun.deprecated_notice` to a non-empty string.


crates/build/re_types_builder/src/lib.rs
```diff
@@ -206,9 +204,11 @@ pub const ATTR_RUST_OVERRIDE_CRATE: &str = "attr.rust.override_crate";
 pub const ATTR_RUST_REPR: &str = "attr.rust.repr";
 pub const ATTR_RUST_TUPLE_STRUCT: &str = "attr.rust.tuple_struct";
 
-pub const ATTR_DOCS_UNRELEASED: &str = "attr.docs.unreleased";
```
abey79, line 209: As noted above, I'd keep that as an independent attribute.

abey79 (approved):
I love this idea, thanks for getting that done! I do have some major-ish semantic-level suggestions though:

- `unreleased` is semantically orthogonal to the "stability" nature of an API. Also, I would much rather have the burden of stability determination be born by the PR author/reviewer than then release manager. So I suggest keeping the `unreleased` attribute separate.
- Scoping stability in the `"attr.docs.*"` namespace feels wrong to me. This is a pretty core attribute of a piece of API, with impacts beyond just documentation. For example, deprecating an API may "crash" user code (with `strict=True`). So I suggest using the `"attr.rerun.*"` namespace instead.
- I have trouble telling `unstable` from `experimental` appart, and fail to come up with a use-case where that distinction is actually useful for the user (ie. the same kind of precaution/risk would apply to both). So I suggest getting rid of `experimental` (we can keep this qualifier for features, e.g. "this [possibly feature-flagged] experimental feature X requires using unstable API Y"). Alternatively, a more objective documentation is required to help devs pick which one to pick.

I feel pretty strongly about the above, but will let you have the last word on it, and the code lgtm, so here is a ✅ 🙂

PR #9435, opened 2025-03-31
Infer image/video channel index from LeRobot metadata

oxkitsune:
### What

Most LeRobot datasets come with a name for each dimension of an `image` or `video` feature. This means we can attempt to infer the index, and do less guessing.

Additionally I rewrote the `Names` parsing to be a bit more general and support even more possible ways to name the dimensions of your feature!


abey79 (approved):
I'm lacking some context to assess the details, but code lgtm and nicely tested 👍🏻

PR #9438, opened 2025-03-31
Align all RPCs with latest spec

teh-cmc:
Title.

* https://github.com/rerun-io/rerun/pull/9438


crates/store/re_protos/proto/rerun/v1alpha1/common.proto
```diff
@@ -2,6 +2,8 @@ syntax = "proto3";
 
 package rerun.common.v1alpha1;
 
+// TODO(cmc): aren't most of the definitions in here dead code?
```
zehiko, line 5: most of things were defined for the very first PR's need for the ``QueryExpression``. Since that is gone, we can probably remove large parts of it.

zehiko (approved):

Issue #9363, opened 2025-03-24
Spawn heuristics open empty `SpatialView2Ds`


grtlr:
**Describe the bug**

There seems to be a problem with the spawn heuristics for `SpatialView2D` recommending empty views. What's interesting is that this even tries to open a view for `/__properties`, which should never happen.

**To Reproduce**
1. Open viewer.
2. Click the "Detect and track objects" example.
3. A couple of wild 2D views appear.

**Expected behavior**

No extra views should appear.

**Screenshots**

<img width="1307" alt="Image" src="https://github.com/user-attachments/assets/04483b80-a358-4e8b-8640-8dddc3c47558" />

**Additional context**

Potentially related PRs (could be a merge race):

* #9308
* #9315 

Note that reverting #9308 seems to "fix" the problem.

PR #9439, opened 2025-03-31
Fix 2D image split heuristic

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/9363
* Regression caused by https://github.com/rerun-io/rerun/pull/9308

### What

As far as I can tell the main breakage occurred because previously we tracked which visualizable entities remain in the subtree during image split computation. That's fixed by stopping image split detection when there's no images anymore (duh! ;-)). There was another issue where we'd count videos and encoded images both as each other additionally.

Adds a simple snapshot test for the spawn heuristic which was confirmed catching this! First of its kind 😱

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7050,6 +7050,7 @@ dependencies = [
  "glam",
  "hexasphere",
  "image",
+ "insta",
  "itertools 0.14.0",
  "ndarray",
  "nohash-hasher",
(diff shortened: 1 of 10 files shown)


abey79 (approved):
Nice! And great to have such a test. Would have come in handy back in the days where breaking heuristics was a daily topic :)

Issue #9432, opened 2025-03-31
Fix message decoding size limits globally


jprochazk:
The gRCP-generated clients/servers have the `message_decoding_size` configured to some value that tends to be too small for us. We should figure out a way to increase those limits globally.

The bandaid fix will be to ensure that every place has it properly configured

Wumpf:
@jprochazk is this easy or hard to do? - I don't know at all what's involved there
Sounds like it would hedge against future bugs

jprochazk:
I don't know either, given that it requires more research I'd consider this hard/high-risk rather than easy. I'd rather remove it from the milestone

PR #9442, opened 2025-04-01
Set `StorageNodeClient` max decoding size

jprochazk:
* Part of https://github.com/rerun-io/rerun/issues/9432


jprochazk:
cargo deny failure is unrelated

emilk (commented):
Has this been tested?

Have you looked for other places where we should increase a similar limit?

jprochazk:
> Has this been tested?

I'm getting reports that this may not even be necessary anymore as the `StorageNodeClient` is going away

> Have you looked for other places where we should increase a similar limit?

Yes, every other place has `max_decoding_message_size` set.

Issue #9404, opened 2025-03-27
Provide Python API to spawn a gRPC server


oxkitsune:

jprochazk:
Current plan is to expose `serve_grpc` from all SDKs.

PR #9447, opened 2025-04-01
Add `serve_grpc` API

jprochazk:
* Closes https://github.com/rerun-io/rerun/issues/9404

This PR adds the `serve_grpc` API to all our SDKs, and the `--serve-grpc` option to the CLI. It can be used to spawn a gRPC server without requiring users to also spawn a Viewer.

All of this is gated behind the `server` feature in both the CLI and our SDKs.

### Connect a native Viewer to a gRPC server hosted by the SDK

Host a gRPC server, log some data to it, and then wait for incoming connections:

<details><summary>Python</summary>

```python
import time
import numpy as np
import rerun as rr

rr.init("rerun_example_serve_grpc")
rr.serve_grpc()

SIZE = 10
pos_grid = np.meshgrid(*[np.linspace(-10, 10, SIZE)] * 3)
positions = np.vstack([d.reshape(-1) for d in pos_grid]).T
col_grid = np.meshgrid(*[np.linspace(0, 255, SIZE)] * 3)
colors = np.vstack([c.reshape(-1) for c in col_grid]).astype(np.uint8).T
rr.log("my_points", rr.Points3D(positions, colors=colors, radii=0.5))

# We have to explicitly wait here, otherwise the server will shutdown immediately.
while True:
  try:
    time.sleep(1)
  except KeyboardInterrupt:
    break
```

</details>

<details><summary>C++</summary>

```cpp
#include <rerun.hpp>
#include <rerun/demo_utils.hpp>
#include <thread>
#include <chrono>

using rerun::demo::grid3d;

int main() {
    // Create a new `RecordingStream` which sends data over gRPC to the viewer process.
    const auto rec = rerun::RecordingStream("rerun_example_serve_grpc_cpp");
    // Try to spawn a new viewer instance.
    rec.serve_grpc().exit_on_failure();

    // Create some data using the `grid` utility function.
    std::vector<rerun::Position3D> points = grid3d<rerun::Position3D, float>(-10.f, 10.f, 10);
    std::vector<rerun::Color> colors = grid3d<rerun::Color, uint8_t>(0, 255, 10);

    // Log the "my_points" entity with our data, using the `Points3D` archetype.
    rec.log("my_points", rerun::Points3D(points).with_colors(colors).with_radii({0.5f}));

    // We have to explicitly wait here, otherwise the server will shutdown immediately.
    while (true) {
        std::this_thread::sleep_for(std::chrono::milliseconds(1000));
    }
}
```

</details>

<details><summary>Rust</summary>

```rust
use rerun::{demo_util::grid, external::glam};

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let rec = rerun::RecordingStreamBuilder::new("rerun_example_serve_grpc_rs")
        .serve_grpc()?;

    let points = grid(glam::Vec3::splat(-10.0), glam::Vec3::splat(10.0), 10);
    let colors = grid(glam::Vec3::ZERO, glam::Vec3::splat(255.0), 10)
        .map(|v| rerun::Color::from_rgb(v.x as u8, v.y as u8, v.z as u8));

    rec.log(
        "my_points",
        &rerun::Points3D::new(points)
            .with_colors(colors)
            .with_radii([0.5]),
    )?;

    // We have to explicitly wait here, otherwise the server will shutdown immediately.
    loop {
        std::thread::sleep(std::time::Duration::from_secs(1));
    }
}
```

</details>

Then connect to the server:
```
rerun --connect
```

### Connect a native Viewer to a gRPC server hosted by the CLI, with data supplied by the Python SDK

First, start the gRPC server:

```
rerun --serve-grpc
```

Then, log some data to it:

```python
import numpy as np
import rerun as rr

rr.init("rerun_example_serve_grpc")
rr.connect_grpc()

SIZE = 10
pos_grid = np.meshgrid(*[np.linspace(-10, 10, SIZE)] * 3)
positions = np.vstack([d.reshape(-1) for d in pos_grid]).T
col_grid = np.meshgrid(*[np.linspace(0, 255, SIZE)] * 3)
colors = np.vstack([c.reshape(-1) for c in col_grid]).astype(np.uint8).T
rr.log("my_points", rr.Points3D(positions, colors=colors, radii=0.5))
```

Finally, connect to the server from a native Viewer:

```
rerun --connect
```


emilk:
```
> rerun --serve-grpc
[… INFO  re_grpc_server] Listening for gRPC connections on 0.0.0.0:9876

> rerun --connect 0.0.0.0:9876
Error: invalid or missing scheme (expected `rerun(+http|+https)://`)
```

Let's improve the first message to

> Listening for gRPC connections on 0.0.0.0:9876. You can connect to this server by running `rerun --connect rerun+http://0.0.0.0:9876/proxy`

The second error message could also be improved with:

> …expected one of:  rerun://  rerun+http://  rerun+https://

---

```
> rerun --connect rerun://0.0.0.0:9876/proxy
[… WARN  re_viewer::app] Data source rerun://0.0.0.0:9876/proxy has left unexpectedly: transport error
```
We should be able to improve this error message too, and hint about what went wrong.

What more common mistakes could the user do, and how can we help them in those situations?

crates/top/re_sdk/src/grpc_server.rs
```diff
@@ -0,0 +1,79 @@
...
+            re_smart_channel::SmartChannelSource::Sdk,
+        );
+        let server_handle = std::thread::Builder::new()
+            .name("message_proxy_server".to_owned())
+            .spawn(move || {
+                let mut builder = tokio::runtime::Builder::new_current_thread();
+                builder.enable_all();
+                let rt = builder.build().expect("failed to build tokio runtime");
+
+                rt.block_on(re_grpc_server::serve_from_channel(
```
emilk, line 42: Will this work even if the user already has `#[tokio::main]` on their `fn main`?

crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -401,6 +405,63 @@ impl RecordingStreamBuilder {
...
+    /// To configure the gRPC server's IP and port, use [`Self::serve_grpc_opts`] instead.
+    ///
+    /// The gRPC server will buffer in memory so that late connecting viewers will still get all the data.
+    /// You can limit the amount of data buffered by the gRPC server using [`Self::serve_grpc_opts`],
+    /// with the `server_memory_limit` argument. Once the memory limit is reached, the earliest logged data
+    /// will be dropped. Static data is never dropped.
+    pub fn serve_grpc(self) -> RecordingStreamResult<RecordingStream> {
+        self.serve_grpc_opts(
+            "0.0.0.0",
+            9876,
```
emilk, line 424:
```suggestion
            DEFAULT_SERVER_PORT,
```

crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -1809,6 +1870,49 @@ impl RecordingStream {
...
+    /// To configure the gRPC server's IP and port, use [`Self::serve_grpc_opts`] instead.
+    ///
+    /// The gRPC server will buffer all log data in memory so that late connecting viewers will get all the data.
+    /// You can limit the amount of data buffered by the gRPC server with the `server_memory_limit` argument.
+    /// Once reached, the earliest logged data will be dropped. Static data is never dropped.
+    pub fn serve_grpc(
+        &self,
+        server_memory_limit: re_memory::MemoryLimit,
+    ) -> RecordingStreamResult<()> {
+        self.serve_grpc_opts("0.0.0.0", 9876, server_memory_limit)
```
emilk, line 1896:
```suggestion
        self.serve_grpc_opts("0.0.0.0", DEFAULT_SERVER_PORT, server_memory_limit)
```

crates/top/rerun/src/commands/entrypoint.rs
```diff
@@ -57,12 +57,18 @@ Examples:
     Open an .rrd file and stream it to a Web Viewer:
         rerun recording.rrd --web-viewer
 
-    Host a Rerun gRPC server which listens for incoming connections from the logging SDK, buffer the log messages, and serves the results:
+    Host a Rerun gRPC server which listens for incoming connections from the logging SDK, buffers the log messages, and serves the results:
```
emilk, line 60: Let's be consistent with the rest of the descriptions and use imperative:

```suggestion
    Host a Rerun gRPC server which listens for incoming connections from the logging SDK, buffer the log messages, and serve the results:
```

rerun_py/src/python_bridge.rs
```diff
@@ -962,6 +963,55 @@ impl PyBinarySinkStorage {
...
+    recording: Option<&PyRecordingStream>,
+) -> PyResult<()> {
+    #[cfg(feature = "server")]
+    {
+        let Some(recording) = get_data_recording(recording) else {
+            return Ok(());
+        };
+
+        if re_sdk::forced_sink_path().is_some() {
+            re_log::debug!("Ignored call to `serve()` since _RERUN_TEST_FORCE_SAVE is set");
```
emilk, line 982:
```suggestion
            re_log::debug!("Ignored call to `serve_grpc()` since _RERUN_TEST_FORCE_SAVE is set");
```

rerun_py/rerun_sdk/rerun/sinks.py
```diff
@@ -211,6 +211,71 @@ def disconnect(recording: RecordingStream | None = None) -> None:
...
+        Specifies the [`rerun.RecordingStream`][] to use.
+        If left unspecified, defaults to the current active data recording, if there is one.
+        See also: [`rerun.init`][], [`rerun.set_global_data_recording`][].
+    server_memory_limit:
+        Maximum amount of memory to use for buffering log data for clients that connect late.
+        This can be a percentage of the total ram (e.g. "50%") or an absolute value (e.g. "4GB").
+
+    """
+    if not is_recording_enabled(recording):
+        logging.warning("Rerun is disabled - serve() call ignored")
```
emilk, line 252:
```suggestion
        logging.warning("Rerun is disabled - serve_grpc() call ignored")
```

rerun_py/rerun_sdk/rerun/sinks.py
```diff
@@ -211,6 +211,71 @@ def disconnect(recording: RecordingStream | None = None) -> None:
...
+    """
+    Serve log-data over gRPC.
+
+    You can connect to this server with the native viewer using `rerun rerun+http://localhost:{grpc_port}/proxy`.
+
+    The gRPC server will buffer all log data in memory so that late connecting viewers will get all the data.
+    You can limit the amount of data buffered by the gRPC server with the `server_memory_limit` argument.
+    Once reached, the earliest logged data will be dropped. Static data is never dropped.
+
+    This function returns immediately. In order to keep the server running, you must keep the process running
```
emilk, line 230:
```suggestion
    This function returns immediately. In order to keep the server running, you must keep the Python process running
```

docs/content/reference/sdk/operating-modes.md
```diff
@@ -48,16 +48,26 @@ You will need to start a stand-alone Viewer first by typing `rerun` in your term
...
 
-This starts the web version of the Rerun Viewer in your browser, and streams data to it in real-time using gRPC.
+There are two kinds of `serve` calls:
+
+* `serve_web`
+* `serve_grpc`
+
+Both will start a Rerun gRPC server in your process, and stream logged data to it.
+
+The `serve_web` call starts an additional server which hosts the assets required to run the Rerun Viewer in your browser.
```
emilk, line 58: This feels like a confusing Venn diagram, with one command being the super-set of the other, and with overlapping parameters.

I suggest we go with one of these alternatives instead:

### Integrate `serve_web` into `serve_grpc`
```py
rr.serve_grpc() # Only starts the gRPC server

rr.serve_grpc(host_web_viewer=True) # Also hosts the web viewer

# Also opens a browser with the web viewer, connected to the gRPC server
rr.serve_grpc(host_web_viewer=True, open_browser=True)
```

### Replace `serve_web` with `serve_web_viewer`
```py
rr.serve_grpc() # Hosts the gRPC server
rr.serve_web_viewer() # Hosts a web viewer
```

The downside is that the `open_browser` param would probably have to go, or blindly guess on the gRPC ip/port

### Integrate both functions into a common `serve` function
```py
rr.serve(grpc=True) # Only host gRPC server
rr.serve(viewer=True) # Only host web viewer
rr.serve(grpc=True, viewer=True) # Host both
rr.serve() # Error: must choose at least one thing to serve
```
---

I think any of these alternatives will be less confusing to our users.

In either case, we keep the old `serve` and `serve_web`, but deprecated.

WDYT?

jprochazk:
> > rerun --connect rerun://0.0.0.0:9876/proxy
> > [… WARN  re_viewer::app] Data source rerun://0.0.0.0:9876/proxy has left unexpectedly: transport error
> 
> We should be able to improve this error message too, and hint about what went wrong.

We have no idea. `tonic` doesn't expose any concrete information about what went wrong, their error is a wrapper over `Box<dyn std::error::Error>`. All we can say is that the connection failed. I'll update the message to say that.

docs/content/reference/sdk/operating-modes.md
jprochazk: I do think our operating modes are pretty confusing, mostly because we want very specific workflows to be simple. For example, the whole "annotate your script with `rr.init("...", spawn=True)`, `rr.log`, and now you have logging". That depends on a whole metric ton of magic.

But with this PR, I'm trying to fill a gap, not change existing APIs. Changing any our operating modes in major ways would be extremely disruptive, so I'm reluctant to touch any of it any more than renaming some functions.

> Integrate serve_web into serve_grpc
> ...
>
> Replace serve_web with serve_web_viewer
> ...
> 
> Integrate both functions into a common serve function
> ...

Removing gRPC from `serve_web` and calling it `serve_web_viewer` is a win in terms of clarity, but not meaningfully different in usage. You still have to be aware of the fact that both of those things need to happen, but now it's two calls instead of one.

We also need to be careful about keeping our SDKs consistent, and some of those options wouldn't work nearly as well in Rust or C++. I don't even know if we can package Wasm into the C++ build at all.

crates/top/re_sdk/src/grpc_server.rs
jprochazk: Yes, we're spawning a new thread, so anything goes.

docs/content/reference/sdk/operating-modes.md
emilk: We can definitely save it for a separate PR. I'll create an issue.

emilk: * https://github.com/rerun-io/rerun/issues/9469

crates/store/re_grpc_server/src/lib.rs
```diff
@@ -73,7 +73,12 @@ async fn serve_impl(
     let incoming =
         TcpIncoming::from_listener(tcp_listener, true, None).expect("failed to init listener");
 
-    re_log::info!("Listening for gRPC connections on {addr}");
+    let connect_addr = if addr.ip().is_loopback() || addr.ip().is_unspecified() {
+        format!("rerun+http://127.0.0.1:{}/proxy", addr.port())
+    } else {
+        format!("rerun+http://{addr}/proxy")
+    };
+    re_log::info!("Listening for gRPC connections on {addr}. Connect by running `rerun --connect {connect_addr}`");
```
emilk, line 81: very nice

emilk (approved):
Nice

PR #9462, opened 2025-04-02
Preparing for re_protos for dataset queries

teh-cmc:
Just clean up all the type shenanigans, add extensions, kill duplicate objects, etc.

* https://github.com/rerun-io/dataplatform/pull/509


zehiko (approved):
nice!

also thanks for moving stuff out of ``types.rs``

PR #9467, opened 2025-04-02
Assume `rerun.kind: data` when `rerun.kind` is missing

abey79:
### What
- default `rerun.kind` to `"data"`

This should solve the problem of the partition table not being Sorbet compliant, by making more things sorbet compliant.

### TODO
* [x] Test with a real partition table

image


emilk:
I'll take this over

jleibs (approved):
Thank you!

PR #9472, opened 2025-04-03
Implement `Dataset.partition_table()` to expose the partition table as a datafusion table provider

abey79:
### Related

* part of https://github.com/rerun-io/dataplatform/issues/405
* closes https://github.com/rerun-io/dataplatform/issues/402

### What

- Fixes `PartitionTableProvider` to use `GetPartitionTableSchema` instead of a hard-coded schema
- Expose it in python as `Dataset.partition_table()`

image


Wumpf (approved):
pretty!

PR #9475, opened 2025-04-03
Impl `QueryDataset`

teh-cmc:
Nothing fancy here, see the sibling.

* https://github.com/rerun-io/dataplatform/pull/515


crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
```diff
@@ -310,13 +310,16 @@ message QueryDatasetRequest {
   // all chunks that match other query parameters will be included.
   repeated rerun.common.v1alpha1.Tuid chunk_ids = 3;
 
+  // Which entity paths are we interested in? Leave empty to query all of them.
+  repeated rerun.common.v1alpha1.EntityPath entity_paths = 4;
```
teh-cmc, line 314: `entity_paths` is now part of the main request, which allows the caller to query for specific entity paths without any particular latest-at and/or range query.

PR #9476, opened 2025-04-03
Refactor list item visuals

lucasmerlin:
### Related

* Continuation of https://github.com/rerun-io/rerun/pull/9418

### What

* Changes the icons for recordings and datasets
* Refactors list item visuals to match @gavrelina s design 
* ListItem now has a active flag, with a nice bg highlight
* Fixes hover interactions for list item buttons
* Don't grey out `Unnamed` LabelContent

https://github.com/user-attachments/assets/e188602a-4e5f-49af-aa04-aceb712e5f80


gavrelina:
Comment about impacting the rest of the lists!

Having icons dimmer than the text will be great for the other spots (blueprint + streams), because we better communicate hierarchy and not overwhelming users with equally high contrast to everything!

This however will make the autogenerated namespace of views and containers look awkward => both fully grey. and that looks like something inactive, or invisible => which is not the case. So we shall color all the names the same and we can still keep _italic_ style for autogenerated ones.

crates/viewer/re_ui/tests/snapshots/list_items.png
abey79, line 1: is it expected that the `active` version has some (light)  background?

crates/viewer/re_ui/src/list_item/label_content.rs
```diff
@@ -14,7 +14,7 @@ pub struct LabelContent<'a> {
     italics: bool,
 
     label_style: LabelStyle,
-    icon_fn: Option<Box<dyn FnOnce(&mut egui::Ui, egui::Rect, egui::style::WidgetVisuals) + 'a>>,
+    icon_fn: Option<Box<dyn FnOnce(&mut egui::Ui, egui::Rect, ListVisuals) + 'a>>,
```
abey79, line 17: I would have expected these changes to apply to `PropertyContent` as well. Any reason not to?

abey79 (approved):
Code lgtm. I trust @gavrelina and you for the actual design :)

crates/viewer/re_ui/tests/snapshots/list_items.png
lucasmerlin: Yes, that was intentional!

crates/viewer/re_ui/src/list_item/label_content.rs
lucasmerlin: Ah, you're right, I missed that

lucasmerlin: I've opened an issue and I'll do it in a separate PR, since I want to focus on the table ui improvements for the demo for now:

https://github.com/rerun-io/rerun/issues/9485

PR #9478, opened 2025-04-03
Support `#datapath` fragment on redap URL

emilk:
This adds support for appending a data path (e.g. `/some/entity[#42]`) to a redap URL, as a `#fragment`.

When clicking a redap URI with such a fragment, we will focus on that entity.
This is untested, however, and depends on us keeping the focus until the entity is loaded.

A `DataPath` is currently defined as

```rs
pub struct DataPath {
    pub entity_path: EntityPath,
    pub instance: Option<Instance>,
    pub component_name: Option<ComponentName>,
}
```

We will wanna extend this in the future to support tagging, and maybe even a timepoint.

### Extending it with time point

It's not clear how to specify _both_ time _and_ a data path in the fragment 🤔  I guess we need to use a special character, like comma:

`#/entity/path,log_tick=42` maybe

…and then comes the trouble of escaping

### DIFF
--- a/crates/utils/re_uri/src/endpoints/dataset.rs
+++ b/crates/utils/re_uri/src/endpoints/dataset.rs
@@ -14,6 +14,7 @@ pub struct DatasetDataEndpoint {
     pub origin: Origin,
     pub dataset_id: re_tuid::Tuid,
 
+    // Query parameters: these affect what data is returned.
     pub partition_id: String,
     pub time_range: Option<TimeRange>,
 }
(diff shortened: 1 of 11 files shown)


crates/utils/re_uri/src/endpoints/dataset.rs
```diff
@@ -7,15 +7,20 @@ use crate::{Origin, RedapUri, TimeRange};
 /// Currently, the following format is supported:
 /// `<origin>/dataset/$DATASET_ID/data?partition_id=$PARTITION_ID&time_range=$TIME_RANGE`
 ///
-/// `partition_id` is mandatory, and `time_range` is optional. In the future, it will be extended to
-/// richer queries.
+/// `partition_id` is currnelty mandatory, and `time_range` is optional.
```
Wumpf, line 10:
```suggestion
/// `partition_id` is currently mandatory, and `time_range` is optional.
```

scripts/lint.py
```diff
@@ -249,11 +249,6 @@ def lint_line(
             if not app_id.startswith("rerun_example_") and not app_id == "<your_app_name>":
                 return f"All examples should have an app_id starting with 'rerun_example_'. Found '{app_id}'"
 
-    # Methods that return Self should usually be marked #[inline] or #[inline(always)] since they indicate a builder.
-    if re.search(r"\(mut self.*-> Self", line):
-        if prev_line_stripped != "#[inline]" and prev_line_stripped != "#[inline(always)]":
-            return "Builder methods impls should be marked #[inline]"
```
Wumpf, line 255: that wasn't a bad lint though, no? could just do `NOLINT` on `without_query_and_fragment` 🤷

Wumpf (approved):
>  depends on us keeping the focus until the entity is loaded.

that's a bit of a problem 🤔
We need to find a way to delay that. Iirc we kill of invalid selections, so the selection will be lost immediately if the target in question isn't there yet

Would have been nice to always support that fragment, even when dealing with non datasetdata urls (i.e. with the "classic" url parameter on the web viewer). But I'm too much out of the loop there to understand how hard that would be.

looks very promising otherwise, let's ship it and iterate as needed

scripts/lint.py
emilk: I added it back before rustc did auto-inlining of trivial functions. It _should_ not be needed anymore

emilk:
* This is improved in https://github.com/rerun-io/rerun/pull/9535

PR #9482, opened 2025-04-03
Automatically add unknown catalog servers

Wumpf:
previously, when clicking a rerun dataplatform url to a catalog you're not connected with, the recording would load fine but not show up on the left hand panel since that depended on the catalog being known.

Alternatives considered:
Show the recording under "disconnected". In some ways this is nicer since nobody asked to connect to the catalog as a whole. But as long as this is light weight in both perf and ux this is probably fine 🤷


abey79 (approved):

abey79:
@emilk sorry for the review request, I missed click 😬

PR #9486, opened 2025-04-04
Show tables under local and use table icon

lucasmerlin:
### What

- Rename "Local recordings" to just "Local"
- in the design it was "Local storage", but I think that is a bit confusing since for now they aren't stored, if you close the viewer it's gone
- Also uses new table icon


grtlr (approved):
Nice!

PR #9491, opened 2025-04-04
Add `Dataset.partition_url()` to produce urls for a given partition

abey79:
### Related

* Part of https://github.com/rerun-io/dataplatform/issues/405

### What

Title.

image


Wumpf (approved):
neat, that's convenient!

PR #9493, opened 2025-04-04
Allow passing url to notebook viewer

Wumpf:
Makes it possible to open viewers directly with URLs in python notebooks:

Example:
```py
viewer = rr.notebook.Viewer(
   width=1100, height=800,
   url=f"rerun+http://127.0.0.1:51234/dataset/{dataset.id}/data?partition_id={first_episode}"
)
viewer.display()
```

### DIFF
--- a/rerun_py/rerun_sdk/rerun/notebook.py
+++ b/rerun_py/rerun_sdk/rerun/notebook.py
@@ -84,6 +84,7 @@ class Viewer:
         *,
         width: int | None = None,
         height: int | None = None,
+        url: str | None = None,
         blueprint: BlueprintLike | None = None,
         recording: RecordingStream | None = None,
         use_global_recording: bool = True,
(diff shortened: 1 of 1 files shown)


rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -144,20 +147,26 @@ def __init__(
...
-                raise ValueError("No recording specified and no active recording found")
-
-            bindings.set_callback_sink(
-                recording=recording.to_native(),
-                callback=self._flush_hook,
-            )
-
-            if blueprint is not None:
-                recording.send_blueprint(blueprint)
+                if url is None:
```
jprochazk, line 156: This all feels very wrong, but not because of what you added. The whole `use_global_recording` thing is not working as advertised at all. I can make a separate PR to fix this afterwards

jprochazk (approved):

rerun_py/rerun_sdk/rerun/notebook.py
Wumpf: ohhhhh seeing it as well now. Yeah that parameter is super broken Oo. Thanks

PR #9494, opened 2025-04-04
Add a dumb cache for schemas

jleibs:
### What

Schema requests are currently very slow. We cache them for now, though this will have issues if you access the schema and then update the dataset.


abey79 (approved):
scary but ok 😅

PR #9496, opened 2025-04-04
Dataframe queries on dataset: silo by partition and inject partition id in the results

abey79:
Title.

### DIFF
--- a/crates/store/re_datafusion/src/dataframe_query_provider.rs
+++ b/crates/store/re_datafusion/src/dataframe_query_provider.rs
@@ -1,35 +1,41 @@
+use std::collections::BTreeMap;
 use std::sync::Arc;
 
-use arrow::datatypes::SchemaRef;
+use arrow::array::{Array, RecordBatch, StringArray};
+use arrow::datatypes::{DataType, Field, Schema, SchemaRef};
 use datafusion::{
     catalog::{streaming::StreamingTable, TableProvider},
     error::DataFusionError,
     execution::SendableRecordBatchStream,
     physical_plan::{stream::RecordBatchStreamAdapter, streaming::PartitionStream},
 };
-use futures_util::StreamExt as _;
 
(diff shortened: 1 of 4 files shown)


crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -1,35 +1,41 @@
...
     pub fn new(
-        query_engine: QueryEngine<StorageEngine>,
+        query_engines: BTreeMap<String, QueryEngine<StorageEngine>>,
         query_expression: QueryExpression,
     ) -> Self {
-        let schema = query_engine
-            .query(query_expression.clone())
-            .schema()
-            .clone();
+        let schema = query_engines
```
jleibs, line 26: Ahhh, I see -- we still need the query engine to get the "real" schema based on the query expression, even though we have a schema.  We'll still want to make this a client-local operation as it should be fully resolvable just based on the actual schema... but much more than I want to tackle right now.

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -72,3 +88,26 @@ impl std::fmt::Debug for DataframeQueryTableProvider {
...
+
+fn prepend_string_column(
+    batch: &RecordBatch,
+    column_name: &str,
+    value: &str,
+) -> Result<RecordBatch, arrow::error::ArrowError> {
+    let row_count = batch.num_rows();
+
+    let new_array =
+        Arc::new(StringArray::from(vec![value.to_owned(); row_count])) as Arc<dyn Array>;
```
jleibs, line 100: TODO: since arrow arrays are cheap to clone while using shared storage, I believe we can actually pre-allocate a largish semi-static arrow array with a repeated partition_id once, and then re-use a possibly sliced views of that buffer for each record-batch.

PR #9499, opened 2025-04-04
Merge schemas when querying across partitions

jleibs:
This fixes two issues introduced by:
- https://github.com/rerun-io/rerun/pull/9496

1) The schema needed the new partition_id column injected into it or we ended up losing a column
2) Not all chunkstores are guaranteed to return the same results, so we do a schema merge operation.


crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -62,14 +59,22 @@ impl PartitionStream for DataframeQueryTableProvider {
         let engines = self.query_engines.clone();
         let query_expression = self.query_expression.clone();
 
+        let target_schema = self.schema.clone();
         let stream = futures_util::stream::iter(engines.into_iter().flat_map(
             move |(partition_id, query_engine)| {
+                let inner_schema = target_schema.clone();
```
timsaucer, line 65: I'm not sure why we need to clone if we're just borrowing later. I'm guessing we needed to in order for the `move` ?

timsaucer (approved):
This all seems reasonable to me.

crates/store/re_datafusion/src/dataframe_query_provider.rs
jleibs: exactly

PR #9500, opened 2025-04-05
Display thumbnail in tables by generating content-based hash for `Blob` components

abey79:
This PR enables thumbnail rendering in tables (dataframe view, redap browser, table entries) by "faking" a `RowId` based on the hash of the data.

### DIFF
--- a/crates/viewer/re_view_dataframe/src/display_record_batch.rs
+++ b/crates/viewer/re_view_dataframe/src/display_record_batch.rs
@@ -151,6 +151,43 @@ impl ComponentData {
                 data
             };
 
+            // TODO(ab): somehow address this hack.
+            //
+            // Background: the `row_id` passed to `ui_raw` is only ever used as a cache key for
+            // images, and we must provide one for images to be displayed. Since in general we don't
+            // have a row id available (e.g. for an arbitrary table entry, or for a dataframe query
+            // result), we just hash the data to get a deterministic row id.
+            let row_id = (component_name.as_str() == "rerun.components.Blob")
+                .then(|| {
+                    re_tracing::profile_scope!("Blob hash");
+
+                    let blob = re_types::components::Blob::from_arrow(&data_to_display).ok()?;
+                    let blob = &blob[0];
+
+                    fn hash_u8_array(data: &[u8]) -> u64 {
+                        use std::hash::{DefaultHasher, Hash as _, Hasher as _};
+
+                        let mut hasher = DefaultHasher::new();
+                        data.hash(&mut hasher);
+                        hasher.finish()
+                    }
+
+                    // According to Claude, hashing the first and last 2kBi of the image buffer
+                    // provides enough entropy.
+                    const MAX_SLICE_SIZE: usize = 2_048 * 1_024;
+
+                    let buffer = blob.as_slice();
+                    let half_len = (buffer.len() / 2).min(MAX_SLICE_SIZE);
+
+                    Some(re_chunk_store::RowId::from_u128(
+                        u128::from(hash_u8_array(&buffer[..half_len]))
+                            | (u128::from(hash_u8_array(
+                                &buffer[buffer.len().saturating_sub(half_len)..],
+                            )) << 64),
+                    ))
+                })
+                .flatten();
+
             ctx.component_ui_registry().ui_raw(
                 ctx,
                 ui,
@@ -159,7 +196,7 @@ impl ComponentData {
                 ctx.recording(),
                 entity_path,
                 component_name,
-                None,
+                row_id,
                 data_to_display.as_ref(),
             );
         } else {


crates/viewer/re_view_dataframe/src/display_record_batch.rs
```diff
@@ -151,6 +151,46 @@ impl ComponentData {
                 data
             };
 
+            // TODO(ab): somehow address this hack.
+            //
+            // Background: the `row_id` passed to `ui_raw` is only ever used as a cache key for
+            // images, and we must provide one for images to be displayed. Since in general we don't
+            // have a row id available (e.g. for an arbitrary table entry, or for a dataframe query
+            // result), we just hash the data to get a deterministic row id.
+            let row_id = (component_name.as_str() == "rerun.components.Blob")
```
emilk, line 160: At this point we should no longer be calling it `row_id`. It should be something like `blob_hash: Hash64` instead.

emilk (changes requested):

emilk:
> It also makes the table noticeably more sluggish to scroll the first time (as the images get cached.

Where is this time spent? You can use the "spike" feature of puffin viewer to figure that out.

* Maybe https://github.com/emilk/egui/pull/5394 will help?

crates/viewer/re_view_dataframe/src/display_record_batch.rs
abey79: I used `row_id` to be consistent with the `DataUi` api, but I agree we should change it all:
- https://github.com/rerun-io/rerun/issues/9514

crates/viewer/re_view_dataframe/src/display_record_batch.rs
```diff
@@ -151,6 +151,46 @@ impl ComponentData {
...
+                        data1.hash(&mut hasher);
+                        data2.hash(&mut hasher);
+                        let hash = hasher.finish();
+
+                        u128::from(hash) | (u128::from(hash) << 64)
+                    }
+
+                    // According to Claude, hashing the first and last 2kBi of the image buffer
+                    // provides enough entropy. For performance reasons, I decided to cut that to
+                    // 2x512KiB.
```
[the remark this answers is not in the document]
abey79: For record, I'll own my inability to spell units and being all over the place with the magnitudes here (my intension was 512B, _not_ 512KiB 🤦🏻). Claude wasn't involved with the actual coding here (it probably would have done a better job).

crates/viewer/re_view_dataframe/src/display_record_batch.rs
emilk: Let's do that issue first then. This is, as far as I know, the only place where we create a `RowId` that isn't actually a row id. It's a lie. Using it to look up a row will always fail. Calling `.nanos_since_epoch()` on it will return nonsense. Let's not break our types intentionally.

abey79: Disagree. The follow-up PR routes row id (if any) all the way to this place. It makes sense to keep this consistent. Also disagree doing the other issue first. It _should_ be easy, but maybe it's not, and this is high priority demo material.

emilk: Passing around a `RowId` that is not a `RowId` is worrying tech debt, and a foot-gun lying around waiting to go off. If we need to move fast for the demo: fine; but then make sure the follow-up issue is added to next release milestone so we don't just accumulate tech debt.

abey79: It was on my short term todo. I've added it to the milestone.

crates/viewer/re_view_dataframe/src/display_record_batch.rs
```diff
@@ -151,6 +151,26 @@ impl ComponentData {
...
+            // Background: the `row_id` passed to `ui_raw` is only ever used as a cache key for
+            // images, and we must provide one for images to be displayed. Since in general we don't
+            // have a row id available (e.g. for an arbitrary table entry, or for a dataframe query
+            // result), we just hash the data to get a deterministic row id.
+            let row_id = (component_name.as_str() == "rerun.components.Blob")
+                .then(|| {
+                    re_tracing::profile_scope!("Blob hash");
+
+                    let blob = re_types::components::Blob::from_arrow(&data_to_display).ok()?;
+                    let buffer = &blob[0].as_slice();
```
emilk, line 165: This will panic if there are zero blobs, let's leave a TODO about that too

emilk (approved):

crates/viewer/re_view_dataframe/src/display_record_batch.rs
```diff
@@ -151,6 +151,26 @@ impl ComponentData {
                 data
             };
 
+            // TODO(ab): somehow address this hack.
```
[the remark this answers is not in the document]
abey79: Well the hack is about needing to do content-based hashing instead of having a hash to begin with. But maybe that hack is actually acceptable and should be implemented directly in `Blob` ui?

PR #9502, opened 2025-04-05
Catalog API improvements and typing stubs

abey79:
### What

- Add (basic) docstring to all APIs (this will need to be further improved)
- Add typing stubs in `rerun_bindings.pyi`
- export everything in the `rr.catalog` namespace (ie. use `rr.catalog.CatalogClient(...)`)
- `get_dataset()` and `get_table()` can be passed either a name or an id as string (both will be tried before raising an error)
- BREAKING: `filter_paritition_id()` accepts one or more partition id as `*arg`: `view.filter_parition_id("part_1", "part_2")`
- default argument for `create_[vector|fts]_index`
- BREAKING: `EntryType` -> `EntryKind` on python side

### DIFF
--- a/rerun_py/rerun_bindings/rerun_bindings.pyi
+++ b/rerun_py/rerun_bindings/rerun_bindings.pyi
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
 import os
-from collections.abc import Iterator, Sequence
+from collections.abc import Iterable, Iterator, Sequence
 from datetime import datetime
 from enum import Enum
-from typing import Any, Callable, Optional, Self, Iterable
+from typing import Any, Callable, Optional, Self
 
 import pyarrow as pa
 
--- a/rerun_py/rerun_bindings/rerun_bindings.pyi
+++ b/rerun_py/rerun_bindings/rerun_bindings.pyi
@@ -257,7 +257,7 @@ class RecordingView:
 
     def filter_range_nanos(self, start: int, end: int) -> RecordingView:
         """
-        Filter the view to only include data between the given index values expressed as seconds.
+        Filter the view to only include data between the given index values expressed as nanoseconds.
 
         This range is inclusive and will contain both the value at the start and the value at the end.
 
--- a/rerun_py/rerun_bindings/rerun_bindings.pyi
+++ b/rerun_py/rerun_bindings/rerun_bindings.pyi
@@ -1122,7 +1122,7 @@ class DataframeQueryView:
 
     def filter_range_nanos(self, start: int, end: int) -> Self:
         """
-        Filter the view to only include data between the given index values expressed as seconds.
+        Filter the view to only include data between the given index values expressed as nanoseconds.
 
         This range is inclusive and will contain both the value at the start and the value at the end.
 
--- a/rerun_py/rerun_bindings/rerun_bindings.pyi
+++ b/rerun_py/rerun_bindings/rerun_bindings.pyi
@@ -935,13 +935,11 @@ def asset_video_read_frame_timestamps_nanos(
 ## CATALOG                                                                                                         ##
 #####################################################################################################################
 
-
 class EntryId:
     """A unique identifier for an entry in the catalog."""
 
     def __str__(self) -> str:
         """Return str(self)."""
-
 
 class EntryKind:
     """The kinds of entries that can be stored in the catalog."""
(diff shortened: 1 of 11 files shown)


rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -927,3 +930,341 @@ def asset_video_read_frame_timestamps_nanos(
...
+        RecordingView
+            A new view containing only the data within the specified range.
+
+            The original view will not be modified.
+
+        """
+
+    def filter_range_nanos(self, start: int, end: int) -> Self:
+        """
+        Filter the view to only include data between the given index values expressed as seconds.
```
Wumpf, line 1125:
```suggestion
        Filter the view to only include data between the given index values expressed as nanoseconds.
```

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -927,3 +930,341 @@ def asset_video_read_frame_timestamps_nanos(
...
+    def filter_index_values(self, values: IndexValuesLike) -> Self:
+        """
+        Filter the view to only include data at the provided index values.
+
+        The index values returned will be the intersection between the provided values and the
+        original index values.
+
+        This requires index values to be a precise match. Index values in Rerun are
+        represented as i64 sequence counts or nanoseconds. This API does not expose an interface
+        in floating point seconds, as the numerical conversion would risk false mismatches.
```
Wumpf, line 1156: that's not _entirely_ true anymore though? we have now three distinct kinds of time one of them being nanoseconds since epoch

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -927,3 +930,341 @@ def asset_video_read_frame_timestamps_nanos(
...
+    def using_index_values(self, values: IndexValuesLike) -> Self:
+        """
+        Replace the index in the view with the provided values.
+
+        The output view will always have the same number of rows as the provided values, even if
+        those rows are empty. Use with [`.fill_latest_at()`][rerun.dataframe.RecordingView.fill_latest_at]
+        to populate these rows with the most recent data.
+
+        This requires index values to be a precise match. Index values in Rerun are
+        represented as i64 sequence counts or nanoseconds. This API does not expose an interface
```
Wumpf, line 1202: as above, I'm confused about the impliations of the recent time refactors here

rerun_py/src/catalog/catalog_client.rs
```diff
@@ -127,10 +134,17 @@ impl PyCatalogClient {
 
     //TODO(#9360): `dataset_from_url()`
 
-    fn get_table(self_: Py<Self>, id: EntryIdLike, py: Python<'_>) -> PyResult<Py<PyTable>> {
+    /// Get a table by name or id.
+    ///
```
Wumpf, line 138:
```suggestion
```

rerun_py/src/catalog/dataset.rs
```diff
@@ -32,13 +32,16 @@ use crate::dataframe::{
...
+/// A dataset entry in the catalog.
 #[pyclass(name = "Dataset", extends=PyEntry)]
 pub struct PyDataset {
     pub dataset_handle: DatasetHandle,
 }
 
 #[pymethods]
 impl PyDataset {
+    /// Return the dataset manifest URL.
+    //TODO(ab): not sure we want this to be public
```
Wumpf, line 44: what are reasons pro/con?

Wumpf (approved):
beautiful! want!

rerun_py/rerun_bindings/rerun_bindings.pyi
abey79: Well, this api takes `i64` and ultimately tests against `TimeInt`. If I understand correctly, the seq/duration/datetime semantic layer is entirely short-circuited. So you could say that the mention of "nanoseconds" accurately "hand waves" the fact that this might be either relative or absolute since epoch.

rerun_py/rerun_bindings/rerun_bindings.pyi
abey79: Basically none? Again, this makes this API relatively low-level and, frankly, improvable. For starters, it should be compatible with the index column with use with `rr.send_column()`.

rerun_py/src/catalog/catalog_client.rs
abey79: Reverting this. this becomes a malformed python docstring per ruff &co.

PR #9505, opened 2025-04-05
Improved automatically adding unknown catalog server

Wumpf:
Like

* https://github.com/rerun-io/rerun/pull/9482

but more general! Previously, automatically opening the server wouldn't work when we got a url from somewhere else, other than clikcing a link in the viewer. This is easy to run into and made even easier now by

* https://github.com/rerun-io/rerun/pull/9493 

This fixes it by checking on new receivers directly.

Also, I made adding preexisting servers a debug message rather than a warning: Today we persist servers, so that means when you run `rerun "rerun+http://localhost:51234"` you get a warning the second time you do so which seems weird.

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -465,6 +465,19 @@ impl App {
         #[cfg(not(target_arch = "wasm32"))]
         let rx = crate::wake_up_ui_thread_on_each_msg(rx, self.egui_ctx.clone());
 
+        // Whenever adding a redap receiver from a server that we don't know about already add it.
+        //
+        // Otherwise we end up in a situation where we have a data from an unknown server,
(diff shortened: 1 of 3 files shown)


crates/viewer/re_viewer/src/app.rs
```diff
@@ -465,6 +465,19 @@ impl App {
         #[cfg(not(target_arch = "wasm32"))]
         let rx = crate::wake_up_ui_thread_on_each_msg(rx, self.egui_ctx.clone());
 
+        // Whenever adding a redap receiver from a server that we don't know about already add it.
```
emilk, line 468: This sentence is very hard to parse O.o

emilk, line 468:
```suggestion
        // Add unknown redap servers.
```

emilk (approved):

Issue #9303, opened 2025-03-17
Create `RecordingStream` from file


Wumpf:
A common pattern when working with dataloaders or existing rrd files is to create recording streams and then immediately log a file to them which may even override the previously set recording id (causing a weird situation where we actually have two recordings when we only want one)

Instead, it would be great if we could create recording streams directly from an existing file

PR #9506, opened 2025-04-05
New API for sending files without a recording stream

Wumpf:
### Related

* Related to #9303 
    * This here is an easy more straight forward alternative 

### What

Often it's useful to open a file without creating a recording. In particular for...
* loading an rbl without yet setting up anything else
* loading an rrd
* oneshot visualization, i.e. anything you'd do with `rerun <my file>`

Today this requires creating a recording stream which _always_ willl create a new recording that clutters the viewer.

Usage example:
```py
viewer = rr.notebook.Viewer(width=1100, height=800, url=dataset.partition_url(first_episode))
viewer.send_file("droid.rbl")
viewer.display()
```
Note that in particular for the notebook this works without any existing sinks.

Draft:
* [ ] check with teammates if we want it in this form - interaction with existing sinks is a bit strange 
* [ ] add example / test
* [ ] expose to regular python api, not just notebook
* [ ] improve docs

### DIFF
--- a/crates/top/re_sdk/src/lib.rs
+++ b/crates/top/re_sdk/src/lib.rs
@@ -21,6 +21,8 @@ mod binary_stream_sink;
 mod global;
 mod log_sink;
 mod recording_stream;
+#[cfg(feature = "data_loaders")]
+mod send_file;
 mod spawn;
 
 // -------------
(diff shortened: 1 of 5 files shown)


Wumpf:
shelving this indefinitely

PR #9508, opened 2025-04-06
Don't show "no recordings loaded" when there's tables or a server connected

Wumpf:
"no recordings loaded" would show even if remote recordings are loaded. Looked weird because once there's a server (with or without recordings) or a table, the list isn't really empty anymore anyways

Examples:
Technically no recording, but writing "no recordings loaded" wouldn't be helpful here (screenshot with fix)
image

With a table loaded this label is rather odd (screenshot without fix)
image

### DIFF
--- a/crates/viewer/re_viewer/src/ui/recordings_panel.rs
+++ b/crates/viewer/re_viewer/src/ui/recordings_panel.rs
@@ -114,7 +114,12 @@ fn recording_list_ui(
 
     servers.server_list_ui(ui, ctx, remote_recordings);
 
-    if local_recordings.is_empty() && welcome_screen_state.hide {
+    // Show placeholder message if there's absolutely nothing else to show.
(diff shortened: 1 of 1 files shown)


crates/viewer/re_viewer/src/ui/recordings_panel.rs
```diff
@@ -114,7 +114,12 @@ fn recording_list_ui(
 
     servers.server_list_ui(ui, ctx, remote_recordings);
 
-    if local_recordings.is_empty() && welcome_screen_state.hide {
+    // Show placeholder message if there's absolutely nothing else to show.
+    if ctx.storage_context.tables.is_empty()
+        && servers.is_empty()
```
Wumpf, line 119: note that due to recent changes presence of a remote recording implies presence of a server

emilk (approved):

PR #9509, opened 2025-04-06
Tweaks for displaymode/active application on incoming servers & applications

Wumpf:
These are all around workflows with incoming "disembodied" blueprint in a readp server context:

(commit by commit)
* don't change display mode aggressively upon seeing a new datasource (this might be just a blueprint and we're looking at a table or a datset table right now)
* if a server was added, show that server (has relatively little effect right now since we don't have a server dataset overview _yet_
* don't set the application id if there's no matching recording for it
    * this is probably the most controversial but also most important one in here: If I send a blueprint to Rerun while I'm looking at something else, I don't want it to switch application id & active entry if we don't have a recording of interest yet.

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -525,6 +525,7 @@ impl App {
     ) {
         match cmd {
             SystemCommand::ActivateApp(app_id) => {
+                self.state.display_mode = DisplayMode::LocalRecordings;
                 store_hub.set_active_app(app_id);
             }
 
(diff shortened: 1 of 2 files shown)


crates/viewer/re_viewer/src/app.rs
```diff
@@ -579,15 +577,23 @@ impl App {
             }
             SystemCommand::AddRedapServer { endpoint } => {
                 let re_uri::CatalogEndpoint { origin } = endpoint;
-                self.state.redap_servers.add_server(origin);
+                self.state.redap_servers.add_server(origin.clone());
+
+                if self
+                    .store_hub
+                    .as_ref()
+                    .map_or(true, |store_hub| store_hub.active_entry().is_none())
```
emilk, line 585: I usually find `is_none_or` and `is_some_and` more readable

```suggestion
                    .is_none_or(|store_hub| store_hub.active_entry().is_none())
```

emilk (approved):

crates/viewer/re_viewer/src/app.rs
grtlr: Wonder why this is not picked up by [`unnecessary_unwrap_or`](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_map_or).

emilk: Good question 🤔

PR #9456, opened 2025-04-02
Add new `ViewerClient::send_table` method to SDK

grtlr:
### Related

* Part of #9455.

### What

Adds a new API to the SDK to send tables directly to the viewer.

### Test

This PR also provides snippets for testing:

```sh
cargo run -p snippets -- send_table
```

```sh
pixi run py-build
pixi run -e py python docs/snippets/all/howto/send_table.py
```

### TODOs & Known Issues

* [x] Move `TableId` to `common` protobuf definition.
* [x] Check overwriting of tables work.
* [x] Better error messages if viewer connection can't be established.
* [x] Deleting a table logs an `re_log::error!`
* [x] Error handling!
* [ ] `failed serving connection: connection error` when sending from Rust snippet

Issue #9465, opened 2025-04-02
Deleting all recordings switches to welcome screen although tables still present


grtlr:
**Describe the bug**
Title.

**To Reproduce**
1. Open recording from welcome screen.
2. Add table entry.
3. Remove above recording
4. Viewer jumps back to the welcome screen

**Additional context**

* Introduced by: #9437

PR #9510, opened 2025-04-07
Implement `send_table` in `re_grpc_server` and `re_viewer`

grtlr:
### Related

* Pulled out from #9456

### What

The server-side and viewer logic for `send_table`. Also contains, the required changes for the notebook.


crates/store/re_grpc_server/src/lib.rs
```diff
@@ -247,28 +255,33 @@ pub fn spawn_with_recv(
...
+    let (channel_log_tx, channel_log_rx) = re_smart_channel::smart_channel(
         re_smart_channel::SmartMessageSource::MessageProxy {
             url: format!("rerun+http://{addr}/proxy"),
         },
         re_smart_channel::SmartChannelSource::MessageProxy {
             url: format!("rerun+http://{addr}/proxy"),
         },
     );
-    let (message_proxy, mut broadcast_rx) = MessageProxy::new_with_recv(memory_limit);
+    let (channel_table_tx, channel_table_rx) = crossbeam::channel::unbounded();
```
emilk, line 270: Why are we using two channels here and elsewhere, instead of just one?

rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -212,6 +215,27 @@ def add_recording(
...
+        # Get current schema
+        schema = record_batch.schema
+        schema = schema.with_metadata({b"__table_id": table_id})
+
+        # Create new record batch with updated schema
+        return pyarrow.RecordBatch.from_arrays(record_batch.columns, schema=schema)
+
+    def send_table(
+        self,
+        id: str,
```
emilk, line 228: What format should this id take? anything-goes?

rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -212,6 +215,27 @@ def add_recording(
...
+        table: RecordBatch,
+    ) -> None:
```
emilk, line 230: We should have a pylint that requires public methods to have docstrings

crates/viewer/re_viewer_context/src/store_hub.rs
```diff
@@ -391,6 +396,7 @@ impl StoreHub {
     }
 
     /// Remove all open recordings and applications, and go to the welcome page.
+    // TODO(grtlr): Make sure we also clear all tables
```
emilk, line 399: Maybe this should link to the issue
* https://github.com/rerun-io/rerun/issues/9465

crates/viewer/re_viewer/src/web.rs
```diff
@@ -22,6 +25,11 @@ use crate::web_tools::{
 static GLOBAL: AccountingAllocator<std::alloc::System> =
     AccountingAllocator::new(std::alloc::System);
 
+struct Channel {
+    log_tx: re_smart_channel::Sender<re_log_types::LogMsg>,
+    table_tx: crossbeam::channel::Sender<re_log_types::TableMsg>,
```
emilk, line 30: I guess a consequence of not putting `TableMsg` as a variant of `LogMsg` is that you cannot store tables in .rrd files. Is this intentional?

grtlr: Yes, we specifically said that tables shouldn't end up in `.rrd` files, because there are much better formats for them.

crates/viewer/re_viewer/src/app.rs
```diff
@@ -171,6 +171,8 @@ struct PendingFilePromise {
     promise: poll_promise::Promise<Vec<re_data_source::FileContents>>,
 }
 
+type ReceiveSetTable = parking_lot::Mutex<Vec<crossbeam::channel::Receiver<TableMsg>>>;
```
emilk, line 174: `ReceiveSetTable` reads weird to me

```suggestion
type TableReceiverSet = parking_lot::Mutex<Vec<crossbeam::channel::Receiver<TableMsg>>>;
```

crates/viewer/re_viewer/src/app.rs
```diff
@@ -460,29 +464,25 @@ impl App {
...
-        // Otherwise we end up in a situation where we have a data from an unknown server,
-        // which is unnecessary and can get us into a strange ui state.
-        if let SmartChannelSource::RedapGrpcStream(endpoint) = rx.source() {
-            if !self.state.redap_servers.has_server(&endpoint.origin) {
-                self.command_sender
-                    .send_system(SystemCommand::AddRedapServer {
-                        endpoint: re_uri::CatalogEndpoint::new(endpoint.origin.clone()),
-                    });
-            }
-        }
```
emilk, line 479: Was deleting this code intentional?

crates/viewer/re_viewer/src/web.rs
emilk: That decision should go into the docstring of `TableMsg` and/or `LogMsg` then

crates/viewer/re_viewer/src/web.rs
```diff
@@ -236,8 +247,12 @@ impl WebHandle {
...
         };
 
-        if let Some(tx) = self.tx_channels.remove(id) {
-            tx.quit(None).warn_on_err_once("Failed to send quit marker");
+        if let Some(channel) = self.tx_channels.remove(id) {
+            channel
+                .log_tx
+                .quit(None)
+                .warn_on_err_once("Failed to send quit marker");
+            drop(channel.table_tx);
```
emilk, line 255: I think we should use the destructor pattern here so we remember to update the code if we add more fields things to the `channel` type. It could also be made a member function

crates/viewer/re_viewer/src/web.rs
```diff
@@ -293,6 +309,41 @@ impl WebHandle {
...
+        let Some(app) = self.runner.app_mut::<crate::App>() else {
+            return;
+        };
+
+        if let Some(channel) = self.tx_channels.get(id) {
+            let tx = channel.table_tx.clone();
+
+            let cursor = std::io::Cursor::new(data);
+            let stream_reader = arrow::ipc::reader::StreamReader::try_new(cursor, None) else {
+                re_log::error_once!("Failed to create cursor");
```
emilk, line 323: Log messages should contain useful information. Imagine you are using some random application and it tells you "Failed to create cursor". What would you do in response? Install new mouse drivers?

Let's make sure all error messages are clear and _actionable_. In this case, we should capture the error message that arrow gives us (NEVER throw away a useful error message) and also give some context.

"Failed to interpret data as IPC-encoded arrow: {err}"

Same later on

emilk: I assume this meant to be `let Ok(…)` btw, otherwise the `else` is never gonna execute

crates/viewer/re_viewer/src/app.rs
grtlr: Oh no, thank you for catching that.

crates/store/re_grpc_server/src/lib.rs
```diff
@@ -294,15 +307,92 @@ pub fn spawn_with_recv(
...
+                Ok(msg) => {
+                    if channel_table_tx.send(msg).is_err() {
+                        re_log::debug!(
+                            "message proxy smart channel receiver closed, closing sender"
+                        );
+                        break;
+                    }
+                }
+                Err(err) => {
+                    re_log::error!("dropping table due to failed decode: {err}");
```
emilk, line 339: For consistency, use upper case initials in your log messages. Remember that INFO, WARN and ERROR are all shown to the user, both in the viewer and in the console.

```suggestion
                    re_log::error!("Dropping table due to failed decode: {err}");
```

crates/store/re_grpc_server/src/lib.rs
```diff
@@ -532,9 +672,54 @@ impl MessageProxy {
 
         Box::pin(history.chain(channel))
     }
+
+    async fn new_client_table_stream(&self) -> ReadTablesStream {
+        let (sender, receiver) = oneshot::channel();
+        if let Err(err) = self.event_tx.send(Event::NewClient(sender)).await {
+            re_log::error!("Error initializing new client: {err}");
```
emilk, line 679:
```suggestion
            re_log::error!("Error accepting new client: {err}");
```

maybe?

crates/store/re_grpc_server/src/lib.rs
```diff
@@ -294,15 +307,92 @@ pub fn spawn_with_recv(
...
+                }),
+                Err(broadcast::error::RecvError::Closed) => {
+                    re_log::debug!("message proxy server shut down, closing receiver");
+                    // We don't have to explicitly call `quit` on crossbeam channels.
+                    break;
+                }
+                Err(broadcast::error::RecvError::Lagged(n)) => {
+                    re_log::debug!(
+                        "message proxy receiver dropped {n} messages due to backpressure"
+                    );
```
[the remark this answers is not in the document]
grtlr: That was copy and paste, I will fix the logging above too.

crates/store/re_grpc_server/src/lib.rs
grtlr: https://github.com/rerun-io/rerun/issues/9516

crates/viewer/re_viewer/src/app.rs
grtlr: I named it this way to make it somewhat symmetric with `ReceiveSet<LogMsg>`.

crates/store/re_grpc_server/src/lib.rs
grtlr: Documented in `LogMsg` and `TableMsg`.

rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -212,6 +215,39 @@ def add_recording(
...
+        """
+        Sends a table in the form of a dataframe to the viewer.
+
+        Parameters
+        ----------
+        id : str
+            The name that uniquely identifies the table in the viewer.
+            This name will also be shown in the recording panel.
+        table : RecordBatch
```
emilk, line 239: no need to repeat the types
```suggestion
        id:
            The name that uniquely identifies the table in the viewer.
            This name will also be shown in the recording panel.
        table:
```

emilk (approved):

PR #9511, opened 2025-04-07
Fix opening partitions from redap origins with port 443 or 80

lucasmerlin:
### What

This fixes opeing recordings from redap.rerun.io

### DIFF
--- a/crates/utils/re_uri/src/origin.rs
+++ b/crates/utils/re_uri/src/origin.rs
@@ -38,7 +38,7 @@ impl Origin {
         // `Origin` in that case.
         let mut http_url = url::Url::parse(&rewritten)?;
 
-        if http_url.port().is_none() {
+        if http_url.port_or_known_default().is_none() {
             // If no port is specified, we assume the default redap port:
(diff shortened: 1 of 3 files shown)


crates/utils/re_uri/src/origin.rs
```diff
@@ -38,7 +38,10 @@ impl Origin {
         // `Origin` in that case.
         let mut http_url = url::Url::parse(&rewritten)?;
 
-        if http_url.port().is_none() {
+        // If we parse a Url from e.g. `https://redap.rerun.io:443`, `port` in the Url struct will
+        // be `None`. So we need to use `port_or_known_default` to get the port back.
+        // See also: https://github.com/servo/rust-url/issues/957
+        if http_url.port_or_known_default().is_none() {
```
emilk, line 44: Well this is very annoying. This means the default port logic doesn't work.

We talked this through in a huddle, and decided we'll leave it as a TODO until https://github.com/servo/rust-url/issues/957 is fixed

emilk (approved):

Issue #9469, opened 2025-04-02
Add dedicated `serve_web_viewer` function


emilk:
We have two function for serving data in our SDK:

* `serve_grpc` (https://github.com/rerun-io/rerun/pull/9447)
* `serve_web`

`serve_grpc` starts a gRPC server and uses it as a log sink. Makes sense.

`serve_web` also starts a gRPC server AND a web server. This makes less sense.

I suggest we go with one of these alternatives instead:

## ❌ Integrate `serve_web` into `serve_grpc`
```py
rr.serve_grpc() # Only starts the gRPC server

rr.serve_grpc(host_web_viewer=True) # Also hosts the web viewer

# Also opens a browser with the web viewer, connected to the gRPC server
rr.serve_grpc(host_web_viewer=True, open_browser=True)
```

PRO:
- one single function
- always a log sink
- can't host a web viewer without a gRPC server

CON:
- can't host a web viewer without a gRPC server
- one function, two features

## ✅ Replace `serve_web` with `serve_web_viewer`  ← **We're going with this one!**
Separate function for orthogonal things:

```py
grpc_uri = rr.serve_grpc() # Hosts the gRPC server
rr.serve_web_viewer(connect_to=grpc_uri) # Hosts a web viewer
```

PRO:
- two functions for two features
- you can access all four quadrants of the cartesian feature product

CON:
- there is little value in just hosting a web viewer without a gRPC server
- confusingly, `serve_grpc` is a log sink but `serve_web_viewer` is not
- `open_browser` param would probably have to go, or blindly guess on the gRPC ip/port.

## ❌ Integrate both functions into a common `serve` function
```py
rr.serve(grpc=True) # Only host gRPC server
rr.serve(viewer=True) # Only host web viewer
rr.serve(grpc=True, viewer=True) # Host both
rr.serve() # Error: must choose at least one thing to serve
```

PRO:
- one single function, with a clear name
- you can access all four quadrants of the cartesian feature product

CON:
- its only a sink of `grpc=True`
- one function for two features
- there is little value in just hosting a web viewer without a gRPC server

---

I think any of these alternatives will be less confusing to our users, but I think the first alternative is probably the best.

In either case, we keep the old `serve_web` (and even older `serve`), but deprecated.

emilk:
Let's keep the `rr.serve_web()` as is for 0.23, but deprecate it for Rerun 0.24:
* https://github.com/rerun-io/rerun/issues/9541

[...]

PR #9512, opened 2025-04-07
Refactor: replace naked `url: String` with `re_uri` type

emilk:
### Related
* Quick refactor on the path to https://github.com/rerun-io/rerun/issues/9469


crates/store/re_grpc_client/src/message_proxy/read.rs
```diff
@@ -10,19 +10,18 @@ use crate::TonicStatusError;
 use crate::MAX_DECODING_MESSAGE_SIZE;
 
 pub fn stream(
-    endpoint: re_uri::ProxyEndpoint,
+    url: re_uri::ProxyEndpoint,
```
jprochazk, line 13: We're now calling it three different things at the same time: `url`, `uri`, and `endpoint`. Can we decide on just one and rename the other parts to that? e.g. If we keep `re_uri`, then:
```suggestion
    uri: re_uri::ProxyUri,
```

crates/store/re_grpc_server/src/lib.rs
```diff
@@ -248,13 +248,17 @@ pub fn spawn_with_recv(
     memory_limit: MemoryLimit,
     shutdown: shutdown::Shutdown,
 ) -> re_smart_channel::Receiver<re_log_types::LogMsg> {
-    let (channel_tx, channel_rx) = re_smart_channel::smart_channel(
-        re_smart_channel::SmartMessageSource::MessageProxy {
-            url: format!("rerun+http://{addr}/proxy"),
-        },
-        re_smart_channel::SmartChannelSource::MessageProxy {
-            url: format!("rerun+http://{addr}/proxy"),
+    let url = re_uri::ProxyEndpoint::new(re_uri::Origin {
```
jprochazk, line 251: This should also use `re_uri::Origin::from_scheme_and_socket_addr`

crates/store/re_grpc_client/src/message_proxy/read.rs
emilk: I agree completely… and really these are URLs, since they include scheme…

emilk: I'll clean this up in a follow-up PR

PR #9513, opened 2025-04-07
fix-ci: update to tokio 1.44.2 to fix RUSTSEC-2025-0023

andrea-reale:
### What

Fixes the cargo-deny workflow

```
 error[unsound]: Broadcast channel calls clone in parallel, but does not require `Sync`
    ┌─ /home/runner/work/rerun/rerun/Cargo.lock:686:1
    │
686 │ tokio 1.44.1 registry+https://github.com/rust-lang/crates.io-index
    │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unsound advisory detected
    │
    ├ ID: RUSTSEC-2025-0023
    ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2025-0023
    ├ The broadcast channel internally calls `clone` on the stored value when
      receiving it, and only requires `T:Send`. This means that using the broadcast
      channel with values that are `Send` but not `Sync` can trigger unsoundness if
      the `clone` implementation makes use of the value being `!Sync`.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10067,9 +10067,9 @@ dependencies = [
 
 [[package]]
 name = "tokio"
-version = "1.44.1"
+version = "1.44.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f382da615b842244d4b8738c82ed1275e6c5dd90c459a30941cd07080b06c91a"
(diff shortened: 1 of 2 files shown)


abey79 (approved):
thanks

Issue #9440, opened 2025-04-01
Can't use arrow datatype `Duration` with `datafusion`


emilk:
## Problem
Datafusion 46 does not support the `Duration` datatype:

> Internal error: Min/Max accumulator not implemented for type Duration(Nanosecond).

This is fixed on Datafusion trunk:

* https://github.com/apache/datafusion/pull/15322

Which will be released maybe mid April (1-3 weeks away):

* https://github.com/apache/datafusion/issues/15072

### Wait
Wait for a new datafusion release:
* https://github.com/apache/datafusion/issues/15072

### Avoid `Duration` datatype
We could avoid using the `Duration` datatype until the next datafusion release. If we do, we should wait with the Rerun 0.23 release.

### Patch `datafusion` in dataplatform
Probably the best solution

PR #9518, opened 2025-04-07
Remove datafusion patch

emilk:
### Related
* https://github.com/rerun-io/dataplatform/pull/494
* https://github.com/rerun-io/rerun/issues/9440
* https://github.com/rerun-io/dataplatform/pull/498

### What
[@teh-cmc reports](https://github.com/rerun-io/dataplatform/pull/494#issuecomment-2769603069) that this patch isn't needed since https://github.com/rerun-io/dataplatform/pull/498, and since having a patch crate blocks the 0.23 release, I'd like to revert it to unblocking making an alpha release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2126,7 +2126,8 @@ dependencies = [
 [[package]]
 name = "datafusion"
 version = "46.0.1"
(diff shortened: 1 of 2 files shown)


emilk:
Let's try it and see what happens 🤷

PR #9519, opened 2025-04-07
Reordering and hiding of table columns

lucasmerlin:
### Related

* Part of https://github.com/rerun-io/dataplatform/issues/397

### What

Adds a `Columns` menu where columns can be reordered and shown/hidden.

https://github.com/user-attachments/assets/8e448e0a-f055-4926-ba3e-96ff001923e5


Cargo.toml
```diff
@@ -158,6 +158,7 @@ egui = { version = "0.31.1", features = [
   "rayon",
 ] }
 egui_commonmark = { version = "0.20.0", default-features = false }
+egui_dnd = { version = "0.12.0" }
```
lucasmerlin, line 161: I used egui_dnd, since it makes this kind of sorting really easy, but I could also try to implement it without it if we don't want another dependency.

crates/viewer/re_redap_browser/src/table_utils.rs
```diff
@@ -65,3 +66,129 @@ pub fn cell_ui<R>(
...
+    ///
+    /// Don't forget to call [`Self::store`] to persist the changes.
+    pub fn get_with_columns(
+        ctx: &Context,
+        persisted_id: Id,
+        columns: impl Iterator<Item = ColumnConfig>,
+    ) -> Self {
+        ctx.data_mut(|data| {
+            let config: &mut Self =
+                data.get_persisted_mut_or_insert_with(persisted_id, || Self::new(persisted_id));
```
abey79, line 115: Is there a reason this takes a `persisted_id` as arg instead of using `self.id`? If so, it should be documented.

crates/viewer/re_redap_browser/src/dataset_ui.rs
```diff
@@ -94,35 +110,40 @@ pub fn dataset_ui(
...
-                .iter()
-                .map(|field| {
-                    egui_table::Column::new(200.0)
-                        .resizable(true)
-                        .id(egui::Id::new(field))
-                })
+            table_delegate
+                .table_config
+                .visible_column_ids()
+                .map(|id| egui_table::Column::new(200.0).resizable(true).id(id))
```
abey79, line 137: This is the kind of delegate <-> table config thing that could be more tightly integrated.

abey79 (approved):
Nice!

crates/viewer/re_redap_browser/src/table_utils.rs
lucasmerlin: There is no `self` at this point. I named it `persisted_id` to make it clear that this Id is used for presistence and should ideally be consistent across versions.

Cargo.lock
```diff
@@ -9129,6 +9175,19 @@ dependencies = [
...
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b6277f0217056f77f1d8f49f2950ac6c278c0d607c45f5ee99328d792ede24ec"
+dependencies = [
+ "ab_glyph",
+ "log",
+ "memmap2 0.9.5",
+ "smithay-client-toolkit",
+ "tiny-skia",
+]
```
emilk, line 9189: Why do we need this?

Cargo.toml
emilk: Depending on dnd is fine, but the added `sctk-adwaita` dependency confuses me 🤔

crates/viewer/re_redap_browser/src/table_utils.rs
abey79: ah right. I missed the fact that `new(id)` is not a public api, and this is. Nvm my comment then.

Cargo.lock
lucasmerlin: I currently have a feature in hello_egui_utils that depends on eframe, I plan on removing that with the next hello_egui release. 
I only use it in my dev dependencies for the examples, I didn't realize it would spam peoples Cargo.lock even if it's never enabled. There is also an issue for it: https://github.com/lucasmerlin/hello_egui/issues/58

Why this causes winit to depend on `sctk-adwaita`, no idea 🤔

lucasmerlin: I'll definitely fix it with the next egui release though.

PR #9529, opened 2025-04-07
Make `PyDataframeQueryView` immutable

abey79:
Title. This makes it more consistent with how DataFusion and others usually work.

Part of:
- https://github.com/rerun-io/dataplatform/issues/405


jleibs (approved):
Thanks!

PR #9530, opened 2025-04-07
 DataSet parition ID to URL user defined function

timsaucer:
### Related

Demo 25-04

### What

This adds a udf that can be used thusly:

```python
df = df.with_column("partition_url", dataset.partition_url_udf(df.col("rerun_partition_id")))
```


abey79 (approved):
Nice!

Wumpf:
@abey79 / @timsaucer any reason not to get this towards landing?

timsaucer:
> @abey79 / @timsaucer any reason not to get this towards landing?

If we need this for a demo, I can get it updated and merged in. Otherwise I think the DF48 version of the UDFs is going to be significantly cleaner.

abey79:
Yeah, let's punt on it. I think it was a proposal to address my need to generate the URL column in the viewer, but I did that with basic DF expressions and it's fine. Eventually it's nice to have this for users, but it should be as streamlined as possible so best wait for DF48.

Wumpf:
iirc we go this addressed by now.
Closing (please re-open if needed)

PR #9507, opened 2025-04-06
`DisplayRecordbatch`: route `RowId` to `ComponentData::data_ui` if available

abey79:
### Related

* Follow up to https://github.com/rerun-io/rerun/pull/9500
* Closes https://github.com/rerun-io/rerun/issues/9501

### What

This PR improves `DisplayRecordBatch` such that it uses row ids if they are available. Unfortunately, this is kind of hard to test as there is no simple way to get dataframe with both row ids and thumbnails.

DNM: chained PR

PR #9532, opened 2025-04-07
DEMO HACK: blob acts as thumbnail

abey79:
DO NOT MERGE.

This commit can be cherry picked to:
1) force Table entry rows to be 100px heigh
2) render `Blob` as thumbnail (instead of the usual component ui)

**NOTE**: make sure #9500 and #9507 are merged beforehand, or merge conflict might arise.

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7968,6 +7968,7 @@ dependencies = [
  "re_arrow_util",
  "re_chunk_store",
  "re_component_ui",
+ "re_data_ui",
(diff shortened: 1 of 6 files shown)


emilk:
What should we do with this PR?

abey79:
> What should we do with this PR?

This is nowhere near landable, so I'm closing it. Here is an issue to capture the intent:

- https://github.com/rerun-io/rerun/issues/9795

Issue #9534, opened 2025-04-07
Refactor `TimeInt` vs `NonMinI64` vs `i64`


emilk:
We store our time values as `i64`, but we reserve `i64::MIN` as a special value that we _sometimes_ use to indicate static data. This has lead us into what feels like over-engineering with a bunch of types:

* `NonMinI64` - the set of valid time values
* `TimeInt` - either the special `STATIC` (`i64::MIN`) or a valid `NonMinI64`
* `i64` - how data is actually stored in arrow

At the same time, we convert between these types so often that we have a bunch of `.into()` that just does saturating casts.

What can we do to simplify this situation?

Should we replace it all with `i64`?

joelreymont:
Is this still on the table? 

I took a stab at it but the refactoring due to conversion to `i64` is massive! 

That said, the changes are largely cosmetic.

PR #9535, opened 2025-04-07
Support setting a time value in uri fragment

emilk:
Actually hooking this up properly will have to wait to a follow-up PR (this one is big enough already)

### DIFF
--- a/crates/store/re_log_types/src/index/time_int.rs
+++ b/crates/store/re_log_types/src/index/time_int.rs
@@ -168,14 +168,23 @@ impl From<NonMinI64> for TimeInt {
     }
 }
 
-impl TryFrom<TimeInt> for NonMinI64 {
-    type Error = TryFromIntError;
-
-    #[inline]
-    fn try_from(t: TimeInt) -> Result<Self, Self::Error> {
-        Self::new(t.as_i64()).ok_or(TryFromIntError)
+impl From<TimeInt> for NonMinI64 {
+    fn from(value: TimeInt) -> Self {
+        match value.0 {
+            Some(value) => value,
+            None => Self::MIN,
+        }
     }
 }
+
+// impl TryFrom<TimeInt> for NonMinI64 {
+//     type Error = TryFromIntError;
+
+//     #[inline]
+//     fn try_from(t: TimeInt) -> Result<Self, Self::Error> {
+//         Self::new(t.as_i64()).ok_or(TryFromIntError)
+//     }
+// }
 
 impl From<TimeInt> for Duration {
     #[inline]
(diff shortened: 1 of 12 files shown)


crates/store/re_log_types/src/index/time_int.rs
```diff
@@ -168,15 +168,24 @@ impl From<NonMinI64> for TimeInt {
...
+impl From<TimeInt> for NonMinI64 {
+    fn from(value: TimeInt) -> Self {
+        match value.0 {
+            Some(value) => value,
+            None => Self::MIN,
+        }
     }
 }
 
+// impl TryFrom<TimeInt> for NonMinI64 {
```
jleibs, line 180: Is this being kept around for any particular reason?

crates/utils/re_uri/src/fragment.rs
```diff
@@ -1,17 +1,37 @@
-use re_log_types::DataPath;
+use re_log_types::{DataPath, TimeCell, TimelineName};
 
 /// We use the `#fragment` of the URI to point to a specific entity.
+///
+/// Format:
+/// * `#/entity/path`
+/// * `#/entity/path[#42]`
+/// * `#/entity/path[#42]&log_tick=32`
+/// * `#/entity/path&log_time=2022-01-01T00:00:03.123456789Z`
```
jleibs, line 9: This is inconsistent with the time_range syntax that uses: `{name}@{min}`. Should this perhaps be:
```
#/entity/time_selection=log_time@2022-01-01T00:00:03.123456789Z
```

crates/utils/re_uri/src/fragment.rs
```diff
@@ -22,18 +42,122 @@ impl Fragment {
...
-            Ok(path) => {
-                data_path = Some(path);
-            }
-            Err(err) => {
-                re_log::warn_once!(
-                    "Failed to parse URL fragment '#{fragment}`: {err} (expected a data path)"
-                );
+        for part in split_on_unescaped_ampersand(fragment) {
+            if let Some((timeline, time)) = part.split_once('=') {
+                let timelinen = TimelineName::from(timeline);
```
jleibs, line 49: Not sure if `timelinen` is short for timeline name or a typo... any reason this can't just be `timeline` or `name`, or even `tn`?

emilk: typo

crates/store/re_log_types/src/index/time_int.rs
emilk: Yes, for https://github.com/rerun-io/rerun/issues/9534 - I'll add a comment

jleibs (approved):
Just tested -- works great. Would still love to unify the syntax for time references but other than that I'm happy with it.

Issue #9455, opened 2025-04-02
Project: Table entries in the viewer in SDKs


grtlr:
## Summary description

Add support for Tables (read dataframes) as first-class entries in the Rerun viewer and Python SDK, enabling direct dataframe visualization and better integration with notebook workflows.

This issue is mostly AI generated 🤖. The original internal proposal was written by @abey79 and @grtlr.

## Background and problem

The current chunk-centric storage/visualization infrastructure, while excellent for unaligned data, introduces unnecessary friction when dealing with simple dataframes:

- Dataframes must be deconstructed to fit the current logging API
- Complex blueprint creation is required to emulate dataframe display
- The "recording" user experience isn't optimal for dataframe visualization

## Grounding use-cases

- Send a dataframe from a notebook/script to the viewer
- Load and display a remote table in the viewer
- Enhanced dataframe display compared to traditional notebook/HTML-based tools
- Better integration with other viewer features

## How to describe, demo and evaluate

Success metrics:

- Implementation of a simplified but functional Table support in the viewer
- Successful integration with Python SDK
- Basic table visualization functionality
- Support for atomic table operations (drop & replace)

## Designs and plans

Key implementation components:

- New RerunClient top-level object for viewer connection
Table storage using Vec
- Integration with existing StoreHub/StoreBundle architecture
- UI updates including renamed "Stores" panel and table-specific viewing mode

## Tasks

- Implement `TableStore` for in-memory storage #9437
- Update UI to support table visualization #9437
- Create new gRPC endpoints
- Add Table support to Python SDK
- Notebook support

## Non-goals and won't do

- Support for tables that don't fit in memory
- Complex table operations beyond drop & replace
- Alternative table visualizations or blueprint support
- Garbage collection for tables
- Reading tables out of the viewer
- Support for column hide/show/reordering, filtering, and searching in this phase

grtlr:
The last bits of this have been implemented as outlined in this issue via #9538.

Issue #8692, opened 2025-01-14
Project: improved viewer support for table-style data


jleibs:
Under development by @jleibs 

## Summary description
Make the viewer work more nicely for table-style data such as data sourced from the recording catalog

## Background and problem

The viewer currently makes fairly deep assumptions that the primary workflow of a user is to understand how 1 or more entities evolve over time.
- By default, views will show all of the entities, but only data for a single time-point for each (unless using range queries). This is efficient when the number of entities is small (see the assorted many-entities performance issues).
- Additionally, the hover/selection model is entity-centric. You can select an inspect an entity. There is no equivalent selection for "index" -- there is always a single time-point implicitly selected by the position of the time cursor.

These assumptions flow all the way into the storage layer -- each entity gets its own parallel data-structure, and our IPC format requires a separate chunk per-entity. Each entity logically exists as one or more columns.

However, this model is at odds with the way users think about some kinds of tabular data. In particular data where we think of the row as "entity-like". For example in a catalog view each row is a "recording".

Currently we have two choices for how we load a table of recordings into the viewer:
1. We can use index such as registration time or even just row-id.
  - This has the benefit of working reasonably for the dataframe view
  - However, the timeline view isn't particularly useful
  - There's also no way to "select" a recording
  - Non-table views generally need to be configured to show VisibleTimeRange to make sense or else they only show data associated with a single recording.
2. We can give each recording its own entity-path
  - This totally falls over after 100 or so entities due to the many-entities problem
  - The dataframe view is effectively useless -- we just end up with a super-wide table and no way to correlated data between recordings
  - But, other views, conceptually work, and you can select a recording and view its user-component-metadata

## Grounding use-cases
 
Assume we have added Lat/Lon metadata for every recording in our catalog.

Send the catalog of recordings to the viewer:
- Initially view it as a table
- Switch to a map view and see all the recordings as points on the map
- Be able to "select" a single recording

## How to describe, demo and evaluate

TODO

## Designs and plans
Two possible initial ideas

### Option 1
Expand the types of indexes we support and remove special-casing about temporal/sequential indexing:
- add an "id" index type
- if the index is "id" type, we could default views to show the full index range (need to rename visible time range)
- make it possible to hover/select select a "id" (index-value)
 
### Option 2
More optimizations around a very specific many-entity workflow:
- Introduce a multi-entity chunk where we store entity as a column. (This could be a configuration option chosen at the top-level of a recording).
[...]

PR #9538, opened 2025-04-08
Add experimental `send_table` to Python SDK

### LINKED ISSUE #8443: Proper viewer support for dataframes
## Context
We now have two main types of data:

### Recording-like
- Recording-like data is our historical Rerun "logged" concept.
- Data in a single recording is divided into "Rerun Chunks".
- Unlike an Arrow-IPC Stream, each Chunk in a recording stream is allowed to have a different schema.
- Rerun chunks depend on the existence of certain required columns related to rows and indexes:
  - -> **Every Rerun Chunk is a valid RecordBatch, but not every RecordBatch is a valid Rerun Cunk** 
- A Rerun ChunkStore indexes these chunks and allows for flexible querying operations that return chunks.

### Table-like
- Table-like data has started showing up in our new APIs where we want to map things to a single data-frame.
  - Query Results
  - Catalog
- This exactly matches the traditional Arrow concept
- Any Table-like data exposed as a user-facing python API should map to a `pa.RecordBatchReader`

## Improved Viewer Support

In principal, any Dataframe can be converted to an equivalent set of Rerun chunks by:
- Injecting a row-id based timeline if one doesn't exist already
- Splitting apart columns that belong to separate entities
- Wrapping any non-list types as arrow list arrays.

However, the question is where we should apply this transformation. Doing this on the viewer-ingest side (rather than send side) would both simplify a lot of logging code as well as data-platform implementation.

## Proposal

For incoming client streams (e.g. TCP, notebook, etc.)
- Some header in the stream (maybe part of `StoreInfo`) should determine whether a stream is a "RerunChunk" stream or a "Dataframe" stream.

For gRPC responses we have the abillity to type these more directly. 

In the short term, if the stream is a dataframe stream, each `DataframePart` read from the stream should be converted to 1 or more Rerun chunks, which are injected into a Store.

Longer-term we might introduce an alternative to the ChunkStore for working with these Dataframe stores more directly.

### LINKED ISSUE #9348: Port `send_dataframe` to Rust and improve usability
I think I will start to port `send_dataframe` over to Rust land. As a second step we could create more convenience functions around it (like automagically creating index columns). 

#### Related

* #4466 
* #8443
* #8593
* #8692
* #9302

Internal discussion on requirements: https://rerunio.slack.com/archives/C041NHU952S/p1742471007654759

grtlr:
### Related

* Supersedes #9456
* Closes #9348
* Closes #8443
* Part of #8692
* Part of #9455

### What

This adds minimal support for `send_table` to `re_grpc_client` and implements a new experimental API in the Python SDK:

```py
client = ViewerClient(addr="rerun+http://0.0.0.0:9876")
client.send_table(
    "Hello from Python",
    pa.RecordBatch.from_pydict({
        "id": [1, 2, 3],
        "url": ["https://www.rerun.io", "https://github.com/rerun-io/rerun", "https://crates.io/crates/rerun"],
    }),
)
```

### Testing

Either via the snippet:

```sh
pixi run py-build
pixi run -e py python docs/snippets/all/howto/send_table.py
```

Or via the notebook that was updated in this PR.


rerun_py/src/viewer.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+/// Connection handle to the message proxy service.
+///
+/// This handle is modelled after [`crate::catalog::ConnectionHandle`] and only concerned with
+/// table-related operations, most importantly `WriteTable`.
+// TODO(grtlr): In the future, we probably want to merge this with the other APIs.
+#[derive(Clone)]
+pub struct TableConnectionHandle {
```
abey79, line 58: Wouldn't `ViewerConnectionHandle` be a better name here?

docs/snippets/all/howto/send_table.py
abey79, line 1: We need to cover this feature in the docs. Can you please open an issue an tag it with the 0.23 milestone?

crates/store/re_grpc_client/src/message_proxy/write_table.rs
```diff
@@ -0,0 +1,13 @@
+use re_protos::sdk_comms::v1alpha1::message_proxy_service_client::MessageProxyServiceClient;
+
+#[cfg(not(target_arch = "wasm32"))]
+pub type TableClient = MessageProxyServiceClient<tonic::transport::Channel>;
+
+#[cfg(not(target_arch = "wasm32"))]
+pub async fn table_client(
```
abey79, line 7: Again, `viewer_client`?

abey79 (approved):
Looks very nice and straighforward!

PR #9540, opened 2025-04-08
Add `rr.serve_web_viewer`

emilk:
### Related
* Part of https://github.com/rerun-io/rerun/issues/9469
* Chained to https://github.com/rerun-io/rerun/pull/9539

### What
This adds `rr.serve_web_viewer` which ONLY hosts the web viewer HTML+JS+WASM.

In contrast to the existing `serve_web`, the new `serve_web_viewer`
* does NOT host a gRPC server
* is NOT a log sink
* is NOT available on `RecordingStream`

### Why?
`rr.serve_web` (and the old `rr.serve`) were confusing users because it did two things:
* Hosted the web viewer over HTTP
* Started a gRPC server (previously: WebSocket server)

The new `rr.serve_web_viewer` which is completely orthogonal to `rr.serve_grpc` (no overlap in functionality).
Thus everything `rr.serve()` did can now be done with two function calls, to make it clear what is happening:

#### OLD
```py
rr.serve_web()
```

#### NEW

```py
url = rr.serve_grpc()
rr.serve_web_viewer(connect_to=url)
```

### DIFF
--- a/crates/viewer/re_web_viewer_server/src/lib.rs
+++ b/crates/viewer/re_web_viewer_server/src/lib.rs
@@ -81,6 +81,7 @@ impl FromStr for WebViewerServerPort {
 
 /// HTTP host for the Rerun Web Viewer application
 /// This serves the HTTP+Wasm+JS files that make up the web-viewer.
+#[must_use = "Dropping this means stopping the server"]
 pub struct WebViewerServer {
     inner: Arc<WebViewerServerInner>,
     thread_handle: Option<std::thread::JoinHandle<()>>,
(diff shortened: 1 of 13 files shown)


crates/viewer/re_web_viewer_server/src/lib.rs
```diff
@@ -169,6 +170,12 @@ impl WebViewerServer {
             thread_handle.join().ok();
         }
     }
+
+    /// Keeps the web viewer running until the parent process shuts down.
+    pub fn detach(self) {
+        #[expect(clippy::mem_forget)]
+        std::mem::forget(self);
```
jprochazk, line 177: I'm pretty sure this will work fine, but if the thread shuts down for some reason, we will leak memory, which is unfortunate. If we want to avoid joining here, we can do this instead:
```suggestion
        let Self { inner: _, thread_handle } = self;
        // dropping the thread_handle detaches the thread.
        drop(thread_handle);
```

jprochazk (approved):

PR #9546, opened 2025-04-08
Improve recording panel ui

lucasmerlin:
### What

- fix left sidebar size changing when selecting table
- fix dataset icon not properly aligned
- clean up table handling in app_state
- remove welcome app id based welcome screen and select examples origin if welcome app id is active in `DisplayMode::LocalRecordings`


abey79 (approved):
Thanks!

Issue #9541, opened 2025-04-08
Deprecate `serve_web`


emilk:
We added `rr.serve_web_viewer` in 0.23:
* https://github.com/rerun-io/rerun/pull/9540

We should deprecate the old (confusing) `rr.serve_web`

Wumpf:
@emilk what is the motivation for doing so? It seems like a good shortcut, see doc:
```
  Calling `serve_web` is equivalent to calling [`rerun.RecordingStream.serve_grpc`][] followed by [`rerun.serve_web_viewer`][].
```

emilk:
The motivation can be found in https://github.com/rerun-io/rerun/pull/9540:

----

`rr.serve_web` were confusing users because it did two things:
* Hosted the web viewer over HTTP
* Started a gRPC server (previously: WebSocket server)

The new `rr.serve_web_viewer` which is completely orthogonal to `rr.serve_grpc` (no overlap in functionality).
Thus everything `rr.serve_web()` did can now be done with two function calls, to make it clear what is happening:

----

TLDR: `serve_web` is badly named (it sounds like it does one thing, but it does two different things). Better to be explicit.

PR #9547, opened 2025-04-08
Make use of the new `serve_web_viewer` function

emilk:
### Related
* Follows https://github.com/rerun-io/rerun/pull/9540
* Closes https://github.com/rerun-io/rerun/issues/9469
* Later: #9541

### What
This makes use of the new `serve_web_viewer` function, and refers to it from docs.

### DIFF
--- a/crates/utils/re_uri/src/origin.rs
+++ b/crates/utils/re_uri/src/origin.rs
@@ -86,15 +86,9 @@ impl Origin {
 
         if !has_port {
             // If no port is specified, we assume the default redap port:
-            http_url.set_port(Some(51234)).ok();
-        }
-
-        // If we parse a Url from e.g. `https://redap.rerun.io:443`, `port` in the Url struct will
-        // be `None`. So we need to use `port_or_known_default` to get the port back.
-        // See also: https://github.com/servo/rust-url/issues/957
-        if http_url.port_or_known_default().is_none() {
-            // If no port is specified, we assume the default redap port:
-            http_url.set_port(Some(51234)).ok();
+            http_url
+                .set_port(Some(crate::DEFAULT_GRPC_SERVER_PORT))
+                .ok();
         }
 
         let url::Origin::Tuple(_, host, port) = http_url.origin() else {
--- a/crates/utils/re_uri/src/origin.rs
+++ b/crates/utils/re_uri/src/origin.rs
@@ -48,7 +48,10 @@ impl Origin {
 
     /// Parses a URL and returns the [`crate::Origin`] and the canonical URL (i.e. one that
     ///  starts with `http://` or `https://`).
-    pub(crate) fn replace_and_parse(input: &str) -> Result<(Self, url::Url), Error> {
+    pub(crate) fn replace_and_parse(
+        input: &str,
+        default_port: Option<u16>,
+    ) -> Result<(Self, url::Url), Error> {
         let scheme: Scheme;
         let rewritten;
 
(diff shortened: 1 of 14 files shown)


emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

zalo:
How difficult would it be for this new architecture to handle multiple servers serving to the same viewer?

emilk:
The viewer on `main` can connect to multiple servers already!

crates/utils/re_uri/src/redap_uri.rs
```diff
@@ -362,7 +362,7 @@ mod tests {
 
     #[test]
     fn test_localhost_url() {
-        let url = "rerun+http://localhost:51234/catalog";
```
abey79, line 365: Although it doesn't change the test, having these recap-centric tests using a non-redap port sends to wrong message. These changes should be reverted imho.

crates/utils/re_uri/src/origin.rs
```diff
@@ -48,25 +48,49 @@ impl Origin {
...
+                    .port()
+                    .is_some();
+            } else {
+                // Should not happen.
+            };
+        }
+
+        if !has_port {
+            // If no port is specified, we assume the default:
+            http_url.set_port(Some(crate::DEFAULT_PROXY_PORT)).ok();
```
grtlr, line 89: Wild! 🤓

grtlr (approved):
Nice!

crates/utils/re_uri/src/redap_uri.rs
```diff
@@ -401,65 +401,101 @@ mod tests {
...
+            (
+                "localhost",
+                RedapUri::Catalog(CatalogUri {
+                    origin: Origin {
+                        scheme: Scheme::RerunHttp,
+                        host: url::Host::Domain("localhost".to_owned()),
+                        port: 9876,
+                    },
+                }),
+            ),
```
abey79, line 471: That's about connecting to a local server right? Is the server switching to 9876 as well?

crates/utils/re_uri/src/redap_uri.rs
```diff
@@ -401,65 +401,101 @@ mod tests {
...
+                "127.0.0.1/proxy",
+                RedapUri::Proxy(ProxyUri {
+                    origin: Origin {
+                        scheme: Scheme::RerunHttp,
+                        host: url::Host::Ipv4(Ipv4Addr::new(127, 0, 0, 1)),
+                        port: 9876,
+                    },
+                }),
+            ),
+        ];
```
abey79, line 492: Shouldn't there be tests for non-localhost server? (http -> 80, https -> 443)

abey79 (commented):
It's probably because I lack the context, but I'm confused as to why this `serve_web` PR changes defaults related to connecting to a remote server.

crates/utils/re_uri/src/origin.rs
emilk: yes, that would be nice, but since `Origin` doesn't know about that it would require more structural changes (e.g. have `Origin` parse port to `None`).

It would be easier to just switch redap to use the same port as OSS

abey79: It doesn't make sense for `Origin` to even be able to produce a default port. Only `RedapUri` has sufficient context for that.

jprochazk: > It would be easier to just switch redap to use the same port as OSS

It can't be on the same port until it's the same service

docs/content/reference/sdk/operating-modes.md
```diff
@@ -46,31 +46,28 @@ You will need to start a stand-alone Viewer first by typing `rerun` in your term
 [`RecordingStream::connect`](https://docs.rs/rerun/latest/rerun/struct.RecordingStream.html#method.connect)
 
 
-### Serve
+### `serve_grpc`
+Calling `serve_grpc` will start a Rerun gRPC server in your process, and stream logged data to it.
+This gRPC server can then be connected to from the Rerun Viewer, e.g. by running `rerun --connect localhost/proxy`.
```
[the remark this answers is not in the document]
abey79: I dont see an alternative to making the port an `Option` in `Origin`, for that.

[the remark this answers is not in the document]
jprochazk: You can do `rerun --connect` (without a URL), and it will default to `rerun+http://localhost:9876/proxy`. We can advertise that as the thing most people should do for the time being (until we have only one service)

crates/utils/re_uri/src/origin.rs
emilk: …or else?

docs/content/reference/sdk/operating-modes.md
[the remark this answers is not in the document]
emilk: I didn't know that - I will update the docs

crates/utils/re_uri/src/origin.rs
jprochazk: > as long as we don't run both at the same time on the same machine

You are doing that every time you run dataplatform locally at the same time as a native viewer, unless the viewer is started with `--connect`. Even if the proxy isn't used, it still starts up.

emilk:
Changes:
* For localhost, guess default port based on endpoint (/proxy or not)
* For non-localhost, use 80 or 443 as default port (depending on shceme)

Issue #6198, opened 2024-05-02
Enable `clippy::iter_over_hash_type` lint


emilk:
Iterating over HashMap and HashSet causes a lot of bugs in our UI.

We should fix it where it makes sense, and loudly comment/name things as `unsorted` or similar when we want the performance (see CODE_STYLE.md).

emilk:
We could start by adding `#![warn(clippy::iter_over_hash_type)] // TODO(#6198): enable everywhere` to the `lib.rs`:es of some of our viewer crates

Ghady-Abboud:
I'm assuming this is fixed, right?
```
> cargo clippy  2>&1 | rg 'iteration over unordered hash-based type' | wc -l
0
```

Wumpf:
No, the issue is about enabling the lint int he first place. But I'll be closing this once https://github.com/rerun-io/rerun/pull/12705 is in (tbh that one fell a bit through the cracks, doing some cleanup on our OSS backlog atm).

Issues like this aren't useful anymore as they cause more churn than we're willing to do reviews on - a mass refactor that like is easily done with an internal contributer and an agent (although that might still be a questionable value proposition). In the past issues like this were nice to give small pieces to people that want to get familiar with the codebase. Not practical anymore in the ages of coding agents and no way to turn back time.

PR #9553, opened 2025-04-09
Order dataset by name in the recording panel

abey79:
Title. The order of dataset was previously unstable.


emilk (approved):
* Another strike for https://github.com/rerun-io/rerun/issues/6198

PR #9557, opened 2025-04-09
Import `rerun.notebook` lazily

jprochazk:
We've had a bunch of issues caused by the `from .notebook import ...` line in `rerun_sdk/__init__.py`. Previously, we'd attempt to ignore the `ImportError` if `rerun_notebook` was not installed. It is much much simpler to remove that line, and require that the user explicitly import the `Viewer` widget from `rerun.notebook`. We no longer need to do any `ImportError`-ignoring hacks or attempt to check if the package is installed or not, at the cost of not being able to import the `Viewer` widget as `rr.notebook.Viewer`.

We can still provide `rr.notebook_show` and `RecordingStream.notebook_show`, as we can import `rerun.notebook` inside the function when it is called, which the user shouldn't do unless they have the `rerun.notebook` package installed.

This PR also sets the `RERUN_NOTEBOOK_ASSET` env to `serve-local` by default in the local workspace. We detect this through an extra env variable that is set by pixi via the activation script.

### DIFF
--- a/examples/python/notebook/cube.ipynb
+++ b/examples/python/notebook/cube.ipynb
@@ -24,8 +24,8 @@
     "\n",
     "import numpy as np\n",
     "import rerun as rr  # pip install rerun-sdk\n",
-    "from rerun.notebook import Viewer  # pip install rerun-notebook\n",
-    "import rerun.blueprint as rrb"
+    "import rerun.blueprint as rrb\n",
+    "from rerun.notebook import Viewer  # pip install rerun-notebook"
    ]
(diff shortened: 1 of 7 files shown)


abey79 (approved):
This change makes sense to me. We follow the same pattern in the new catalog python bindings (let the user try something, and fail if the required package are not installed).

The only thing i'm worried about this implementation is that the error will be somewhat "generic", right? Can we make sure that the error specifies that the `rerun_notebook` package should be installed?

PR #9562, opened 2025-04-10
Exclude `how_to/send_table` from snippets

emilk:
* Follow-up to https://github.com/rerun-io/rerun/pull/9538


emilk:
@rerun-bot full-check

PR #9563, opened 2025-04-10
Include field names in Archetype docs

emilk:
Include field names when listing the components of an archetype.

Field names will become more important in the future, and is already important context.

## Before
image

## After
image


abey79 (approved):
very nice change

PR #9569, opened 2025-04-10
Fix `Mat3x3` shape inference in `.columns` api

### LINKED ISSUE #9489: Inferring shape for `Mat3x3` does not work in `.columns` api
`Mat3x3` is stored as a FixedSizedListArray of length 9 in our arrow format, which means we currently expect the user to pass in a batch of `[N, 9]`. However, it makes more sense to pass in an array of shape `[N, 3, 3]` in this case

oxkitsune:
### Related

* Fixes #9489

### What

When inferring the shape for `Mat3x3` from a numpy array we relied on the shape of the underlying arrow array. However, for `Mat3x3` is backed by a `FixedSizeListArray` with length 9, meaning that shapes such as `[N, 3, 3]` would not be inferred correctly. 

This PR fixes this by inferring the shape based on the product of the last dimensions of the tensor, e.g. `3 * 3 = 9`


Wumpf (approved):
great!

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

PR #9570, opened 2025-04-10
Add `row_/chunk_` prefixes to `RowId/ChunkId`

emilk:
## Related
* Part https://github.com/rerun-io/dataplatform/issues/250

## What
When printing RowId and ChunkId, prefix with `row_` and `chunk_`, e.g. `row_182342300C5F8C327a7b4a6e5a379ac4`.

When parsing, this prefix is optional. But the _wrong_ prefix is a hard error.

This is meant to help us developers more easily understand printouts, urls, etc.

The ids are still stored the same in arrow, though in a future PR I'd like to add some metadata to the column so that we have the namespace there as well.

## TODO
* ~Sibling dataplatform PR~ No PR required (I checked)

### DIFF
--- a/crates/store/re_types_core/src/id.rs
+++ b/crates/store/re_types_core/src/id.rs
@@ -44,15 +44,21 @@ impl std::fmt::Display for ChunkId {
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+#[error("Invalid ChunkId: {0}")]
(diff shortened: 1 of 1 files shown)


teh-cmc (approved):
sweet

emilk:
Unrelated CI failure

PR #9571, opened 2025-04-10
add another (fallible) helper for arrow downcasting 

zehiko:
We often downcast arrow arrays with expectations that the downcast should yield a value, and if it doesn't, we return an error. Adding such helper to ``re_arrow_utils``.


teh-cmc (approved):

emilk (approved):
Nice!

Issue #9555, opened 2025-04-09
Implement a simple catalog UI


abey79:
The catalog server should be selectable and have (initially) the simple UI below (use a generic "Catalog" title since we don't have catalog name yet).

This UI should be shown when the catalog is selected, and that should happen at least:
- when clicked in the left panel
- when a catalog URI (e.g. `rerun://redap.rerun.io`) is received 

![Image](https://github.com/user-attachments/assets/f3dfea15-f0b4-43f5-87a3-f5664f3ff5cb)

PR #9576, opened 2025-04-10
Add basic server ui

abey79:
### Related

* Closes https://github.com/rerun-io/rerun/issues/9555

### What

This PR adds a basic server UI. It also allows selecting severs by clicking on them in the recording panel.

image


Wumpf (approved):
great start, love it! Tried it against our ongoing San Fransisco dataplatform demo and works like a charm

PR #9579, opened 2025-04-10
Replace anyhow in re_types

### LINKED ISSUE #1845: Replace `anyhow` with `thiserror` in libraries
We have been too liberal of our use of `anyhow`. For any `pub` function in our libraries, we should be using `thiserror` instead of `anyhow`, to get more type safety and explicitness of what can go wrong.

Internal use of `anyhow` in a crate is less bad, but would also be good to change to `thiserror`.

`anyhow` should mostly be used in high-level crates (binaries) to easily summarize errors coming from many different places without having to create too many error `enum`s

ishakbhatt:
### Related
* Closes #1845

### What
* replaces `anyhow` usage with `thiserror` and no longer in pub functions for `re_types`
* custom enums for results


grtlr (changes requested):
Thank you very much for your PR! 🙏 

It's best to do crates that have this `TODO` one-by-one, but we should always have a version that compiles on `main`. So in this case, please make changes to the `re_data_loader` crate too.

In particular, can you please adapt the `loader_archetype.rs` file , for example by adding `.map_err(anyhow::Error::msg)` to the `from_file_*` methods?

PS: Also, very sorry that it took us so long to review this PR we are usually much quicker to respond.

Wumpf:
closing PR because abandoned by author

PR #9566, opened 2025-04-10
Prevent switches to welcome screen if tables are present

grtlr:
### Related

* Closes #9465

### What

All of this logic is quite brittle, but it should to unblock the `v0.23` release.

PR #9586, opened 2025-04-11
Introduce `Navigation` in preparation for tabs and simplify welcome screen loading

grtlr:
### Related

* Follow up of #9566.

### What

This introduces a `Navigation` object to manage display modes. It also converts the welcome screen into a display mode.

In follow up PRs I will:

* Create display mode for `DataStore`.
* Get rid of requirement for application id.
* Change the display modes to include `StoreId` (or other relevant ids, where appropriate).

### DIFF
--- a/crates/viewer/re_viewer/src/app_state.rs
+++ b/crates/viewer/re_viewer/src/app_state.rs
@@ -23,6 +23,7 @@ use re_viewport_blueprint::ViewportBlueprint;
 
 use crate::{
     app_blueprint::AppBlueprint,
+    navigation::Navigation,
     ui::{recordings_panel_ui, settings_screen_ui},
 };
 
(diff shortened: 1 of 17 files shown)


lucasmerlin:
Should we think about whether using [egui_router](https://github.com/lucasmerlin/hello_egui/tree/main/crates/egui_router) would make sense? It basically has a similar `Navigation` api but also handles integration with the browser history api and transition between routes. 

Also, I'd like to support tabs with it, since I also want tabs for hellopaint, so it might be a great fit.

crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -177,467 +172,493 @@ impl AppState {
...
+        match (self.navigation.peek(), store_context) {
+            (DisplayMode::WelcomeScreen, _) => {
+                self.welcome_screen.ui(
+                    ui,
+                    command_sender,
+                    welcome_screen_state,
+                    is_history_enabled,
+                );
+            }
+            (DisplayMode::Settings, _) => {
```
abey79, line 187: nit: Let's add some whitespace in that match statement for clarity

```suggestion

            (DisplayMode::Settings, _) => {
```

crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -177,467 +172,493 @@ impl AppState {
...
+                settings_screen_ui(ui, &mut self.app_options, command_sender);
+            }
+            (_, Some(store_context)) => {
+                let blueprint_query = self.blueprint_query_for_viewer(store_context.blueprint);
+                let Self {
```
abey79, line 192: Would be great to move this arm's implementation to a dedicated metod

abey79 (approved):
Thanks! I really like the direction where this is going!

PR #9589, opened 2025-04-11
Document `send_table` in the Python SDK

### LINKED ISSUE #9552: Document `send_table` in the Python SDK
This API is still experimental. We should make that clear from our docs, while providing some guidance on how to use that feature and link to the snippet and notebook for example.

grtlr:
### Related

* Closes #9552 

### What

Title.

Once this API (and our table representation in the viewer) stabilizes, I will also add some screenshots.

### DIFF
--- a/docs/content/howto/logging/send_table.md
+++ b/docs/content/howto/logging/send_table.md
@@ -1,0 +1,128 @@
+---
+title: Send tables to Rerun
+order: 0
+description: Shows how to send tables as dataframes to the Rerun viewer.
+---
+
+> **Note:** The `send_table` API is currently experimental and may change in future releases.
+
+Rerun now supports sending tabular data to the Rerun Viewer! This feature allows you to visualize and interact with dataframes (encoded as Arrow record batches) directly in the Rerun Viewer environment.
+
+## Overview
+
+The `send_table` API provides a straightforward way to send tabular data to the Rerun Viewer. This is particularly useful for:
+
+- Inspecting dataframes alongside other visualizations
+- Debugging data processing pipelines
+- Presenting structured data in a readable format
+
+## References
+
+For complete examples of using `send_table`, please refer to:
+
+- [🐍 Jupyter Notebook](https://github.com/rerun-io/rerun/blob/main/examples/python/notebook/send_table.ipynb)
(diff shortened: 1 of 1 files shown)


emilk:
Rendered: https://github.com/rerun-io/rerun/blob/grtlr/docs-send-table/docs/content/howto/logging/send_table.md

docs/content/howto/logging/send_table.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Basic usage
+
+### Connecting to the Viewer
+
+```python
+from rerun_bindings import ViewerClient
+
+# Connect to a running Rerun Viewer
+client = ViewerClient(addr="rerun+http://0.0.0.0:9876")
```
emilk, line 47: Is there no specific `/endpoint` for this?

docs/content/howto/logging/send_table.md
```diff
@@ -0,0 +1,128 @@
...
+import pandas as pd
+import numpy as np
+import pyarrow as pa
+
+# Create a sample DataFrame
+dates = pd.date_range("20230101", periods=6)
+df = pd.DataFrame(np.random.randn(6, 4), index=dates, columns=list("ABCD"))
+
+# Convert to record batch and send to the viewer.
+client.send_table("Pandas DataFrame", pa.RecordBatch.from_pandas(df))
```
emilk, line 79: Could we do this `pa.RecordBatch.from_pandas` call in `send_table` instead (so the user doesn't have to)? If not, why not? Is it just that we don't want to depend on pandas?

docs/content/howto/logging/send_table.md
```diff
@@ -0,0 +1,128 @@
...
+
+# For inline display
+os.environ["RERUN_NOTEBOOK_ASSET"] = "inline"
+
+# Create and display the viewer
+viewer = rr.notebook.Viewer(width=1024, height=350)
+viewer.display()
+
+# Send table directly to the inline viewer
+viewer.send_table("My Table", pa.RecordBatch.from_pydict({
```
emilk, line 99: Same here - maybe we should just support giving a `dict` to `send_table`?

emilk (approved):

docs/content/howto/logging/send_table.md
grtlr: Yes, it uses `/proxy` (our default endpoint), I will add it for consistency.

PR #9537, opened 2025-04-08
re_protos: use async registration by default (protos)

andrea-reale:
### Related

* Part of [dataplatform#527](https://github.com/issues/assigned?issue=rerun-io%7Cdataplatform%7C527)
* Related [dataplatform PR](https://github.com/rerun-io/dataplatform/pull/543)

### What
Make async dataset and partition registrations the default registration method for redap. Keep supporting synchronous registration via the ManifestRegistry direct endpoint only.

* **This is a breaking change of the redap API**

PR #9590, opened 2025-04-11
rerun_py: adapt the `register` API to the new asynchronous nature of  `register_with_dataset`

andrea-reale:
### Related

* Related to: https://github.com/rerun-io/dataplatform/issues/527
* Follows:
  * https://github.com/rerun-io/rerun/pull/9537
  * https://github.com/rerun-io/dataplatform/pull/543 

### What
In https://github.com/rerun-io/dataplatform/pull/543, we replaced the synchronous register_with_dataset endpoint with an
asynchronous equivalent. The python client, however, expects the
endpoint to be synchronous.

This change is a quick-fix to make the python client block for dataset
registration task completion so that current user should not feel the
difference.

We should improve this in the future by exposing the asynchronous nature
of the endpoint directly to the client.

### DIFF
--- a/rerun_py/src/catalog/connection_handle.rs
+++ b/rerun_py/src/catalog/connection_handle.rs
@@ -6,6 +6,7 @@ use pyo3::{
     create_exception, exceptions::PyConnectionError, exceptions::PyRuntimeError, PyErr, PyResult,
     Python,
 };
+use re_log_encoding::codec::wire::decoder::Decode as _;
 use tokio_stream::StreamExt as _;
 
 use re_chunk::{LatestAtQuery, RangeQuery};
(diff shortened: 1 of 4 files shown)


rerun_py/src/catalog/dataset.rs
```diff
@@ -98,11 +98,19 @@ impl PyDataset {
...
+        // A more permanent solution is to expose an asynchronous register method, and/or
+        // the timeout directly to the caller.
+        const MAX_REGISTER_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(60);
         let super_ = self_.as_super();
         let mut connection = super_.client.borrow(self_.py()).connection().clone();
         let dataset_id = super_.details.id;
 
-        connection.register_with_dataset(self_.py(), dataset_id, recording_uri)
+        let task_id = connection.register_with_dataset(self_.py(), dataset_id, recording_uri)?;
+        connection.wait_for_task(self_.py(), &task_id, MAX_REGISTER_TIMEOUT)
```
abey79, line 113: I think the end game here is to immediately return some kind of `PyRemoteTask` object that would expose some task management APIs, such as `.status()` and `.block(timeout)`. Would be great if you could open an issue about that and add a todo here.

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -165,9 +166,10 @@ impl ConnectionHandle {
         py: Python<'_>,
         dataset_id: EntryId,
         recording_uri: String,
-    ) -> PyResult<()> {
+    ) -> PyResult<String> {
```
abey79, line 169: I would use `TaskId` here.

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -177,9 +179,74 @@ impl ConnectionHandle {
...
+                .ok_or_else(|| PyValueError::new_err("received response without task id"))?;
+
+            Ok(task_id.id)
+        })
+    }
+
+    pub fn wait_for_task(
+        &mut self,
+        py: Python<'_>,
+        task_id: &str,
```
abey79, line 196: ....and here

abey79 (approved):
nice

PR #9704, opened 2025-04-13
Replace uses of `arrow2` in `rerun_c` with `arrow-rs`

emilk:
### Related

### What
After this, the only arrow2 remaining is in `re_types_builder`


emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

abey79 (approved):
nice

PR #9728, opened 2025-04-14
Allow using table name to create a table entry datafusion table provider

abey79:
Improvements made on the way to the datafusion table widget:

- Introduce `EntryIdOrName` (convenient for apis accepting both)
- Make it possible to create a `TableEntryTableProvider` using `EntryIdOrName`
- Standardise our datafusion table provider trait around `DataFusionError`

### DIFF
--- a/crates/store/re_log_types/src/entry_id.rs
+++ b/crates/store/re_log_types/src/entry_id.rs
@@ -40,6 +40,10 @@ impl FromStr for EntryId {
 
 // ---
 
+/// Either an id or a name for an entry.
+///
+/// This helper type should only be used for APIs to offer the convenience to refer to entries by
+/// either name or id. For storage/indexing purposes, use [`EntryId`].
 #[derive(Debug, Clone)]
 pub enum EntryIdOrName {
     Id(EntryId),
--- a/crates/store/re_log_types/src/entry_id.rs
+++ b/crates/store/re_log_types/src/entry_id.rs
@@ -37,3 +37,29 @@ impl FromStr for EntryId {
         re_tuid::Tuid::from_str(s).map(|id| Self { id })
     }
 }
+
+// ---
+
+#[derive(Debug, Clone)]
+pub enum EntryIdOrName {
+    Id(EntryId),
+    Name(String),
+}
+
+impl From<EntryId> for EntryIdOrName {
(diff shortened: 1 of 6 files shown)


crates/store/re_log_types/src/entry_id.rs
```diff
@@ -37,3 +37,29 @@ impl FromStr for EntryId {
         re_tuid::Tuid::from_str(s).map(|id| Self { id })
     }
 }
+
+// ---
+
+#[derive(Debug, Clone)]
+pub enum EntryIdOrName {
+    Id(EntryId),
+    Name(String),
```
grtlr, line 46: Maybe?

```suggestion
    Name(TableId),
```

In general, our types are becoming a bit out of hand. We now have:

* `EntryId`
* `EntryKind` (in `re_redap_browser`)
* `TableId`
* `StoreHubEntry`
* `EntryIdOrName`

And there probably more ad-hoc variants that I don't know about.

I'd really love if we could unify / streamline this. For example could we move `EntryKind` to `re_log_types` too?

If it's not straight forward to unify, maybe we can at least add (better) doc comments that explain the landscape?

grtlr (approved):

crates/store/re_log_types/src/entry_id.rs
abey79: It doesn't really make sense to use `TableId` here. For starters, this is an entry, so either a dataset or a table. Also, it really is about the user provided name, so `String` makes sense. That said, I'm all for standardising types if we can.

Some thoughts:
- I will clarify that `EntryIdOrName` should be used _only_ for APIs, not storage. Storage should always use `EntriyId`.
- The `TableId` situation is not ideal. This is a table name, not a table id (as per the definition used in the server). Note that the server enforces both to be unique, and so should the viewer (since the viewer is to become a [light] server eventually). I think it would make sense to rename `TableId` to `EntryName`. Then it might make sense to use it for `EntryIdOrName`, though what matter for this type is the `impl From` and we'd definitely want to support `String`/`&str` for convenience.
- I understand the purpose of `re_redap_browser::EntryKind`, but its naming is unfortunate. `EntryKind` is a proto-defined enum (currently only codegen'd, would benefit from having a "clean" version in `re_log_types` with `From/TryFrom` conversion).

PR #9729, opened 2025-04-14
Revert `Navigation::push_unqiue` until we have unique entries

grtlr:
### Related

* Regression introduced by #9586 

### What

We're not ready for `push_unique` yet because or `LocalRecordings` variant is not unique for each recording yet. This leads to a bug where we prematurely revert back to the Welcome Screen.


grtlr:
Superseded by #9735.

Issue #9520, opened 2025-04-07
Show everything when the entity filter is empty


emilk:
When you click the filter button it's confusing that it hides all entities. It would be more natural for an empty filter to be mean "no filtering" == "show everything". That's what we do in the command palette.

More motivation (from @gavrelina):
* maintain context
*progressive disclosure: start broad and let user narrow down with the search
* more common (hence familiar) pattern
* could be confused! but maybe not with accidental press but with thinking there is nothing to search for (in a very narrow streams panel for example? but also a bit farfetched :slightly_smiling_face: )

---

![Image](https://github.com/user-attachments/assets/3b85705c-30d1-49ff-954f-ee9afeea4091)

emilk:
Maybe the problem then is that an empty text edit means "show nothing" instead of "show everything". WDYT @gavrelina ?

PR #9734, opened 2025-04-14
Do not hide all entities when the search is active but empty

abey79:
### Related

* closes https://github.com/rerun-io/rerun/issues/9520

### What

This PR changes the behaviours of the entity search filter when the search is activated but the query is still empty. Now, the displayed content is unchanged compared to when the search is inactive. Previously, nothing would be displayed in that situation. This required subtle change of when a session id is returned, such that the collapse state used in the active-but-empty-query state is the same as in the inactive state.


emilk (approved):
Thanks ❤️

Issue #9581, opened 2025-04-10
Use a generic `Event` with payload in callbacks


jprochazk:
We recently introduced Viewer callbacks for a few things, like selection and timeline events. The approach we chose was to expose a method per callback which the user could register. This has a number of issues, the biggest of which is how many layers we have to deal with when introducing new callbacks, or changing existing ones.

We'd also like to be able to make some data available in the payload of _every_ event, such as the active application and recording IDs. Every time we add one of those, we now have to change every event throughout all the layers. That's quite a lot of churn.

It may be a lot simpler to have a single `Event` type defined in Rust code for which we generate serialization code.

```rust
#[non_exhaustive] // prevent future breakage when adding new fields.
struct Event {
  // Fields set for all events
  application_id: ApplicationId,
  recording_id: RecordingId,

  /// Event-specific payload.
  detail: EventDetail,
}

#[non_exhaustive] // prevent future breakage when adding new events.
enum EventDetail {
  /// Fired when the selection changes.
  Selection(Selection),

  /// Fired when the timeline cursor moves.
  TimeUpdate(TimeUpdate),

  /// Fired when a different active timeline is selected.
  TimelineChange(TimelineChange),

  // etc.
}
```

Given something like the above, we would only need a single callback that would never have to change.

To make future changes to the `Event` types far less painful, we can create JS- and Python-specific types to which we map the `Event` type defined in `Rust`. These JS/Python-specific types can then use `wasm-bindgen` and `pyo3` to automatically generate equivalent types in their respective languages.

This means that to introduce a new event, a contributor would:

- Update the `Event` type in Rust
- Update its mappings to JS/Python, also in Rust code

...and that's it. 

It's unclear whether or not we can fully codegen all the JS/Python mappings, it may require writing some boilerplate.

Once https://www.github.com/PyO3/pyo3/issues/2454 lands, we can likely get rid of _all_ layers going into Jupyter/Gradio.

PR #9739, opened 2025-04-14
Unify viewer callbacks into a single `Event` interface

jprochazk:
## Related

* Closes https://github.com/rerun-io/rerun/issues/9581

## What

This is a full rewrite of the callback API.

The callback interface has been unified into a single `on_event` callback. This results in the removal of many layers of boilerplate, but a few still remain:

* `ViewerEvent` definition in Rust in `crates/viewer/re_viewer/event.rs`.
  * This definition uses a few custom serialization/deserialization methods through `serde(with = "...")`, for example to serialize `EventPath` as a string instead of an array of strings. This makes the downstream APIs a lot nicer. All the custom serialization is defined in the same file.
* `ViewerEvent` definition in the Web Viewer API in `rerun_js/web-viewer/index.ts`.
  * Because the Viewer outputs events as JSON, they can be deserialized into JavaScript objects without any extra boilerplate, via `JSON.parse`
* `ViewerEvent` definition in the Python SDK in `rerun_py/rerun_sdk/rerun/event.py`
  * Deserialization is done manually.

I hope there is some way to get rid of the need to manually update type definitions and the Python deserialization code, but I'm leaving that for a future follow-up:

* https://github.com/rerun-io/rerun/issues/9740

I've had to update the remaining layers many times while writing this PR, and I can confirm that overall, it is a huge improvement. Instead of having to remember all the different layers, you only have to grep for `ViewerEvent`. 

### Rust API changes

The Rust API now exposes `on_event` in `StartupOptions`, and `callbacks` has been removed. `on_event` yields instances of `ViewerEvent`. The `ViewerEvent` definition includes shared fields, such as `recording_id`.

Users can `match` on the specific event kind in the callback:

```rust
let startup_options = re_viewer::StartupOptions {
    on_event: Arc::new(move |event: re_viewer::ViewerEvent| {
        event.recording_id;
        match event.kind {
            re_viewer::ViewerEventKind::SelectionChange(selection) => {
                for item in selection.items {
                    // ...
                }
            }
            _ => {}
        }
    });
};
```

### Web Viewer API changes

The Web Viewer API exposes events in the same way as before, using the `EventEmitter` pattern. The events have been renamed to `snake_case`:

* `timeupdate` to `time_update`
* `timelinechange` to `timeline_change`
* `selectionchange` to `selection_change`

The callback now yields subtypes of `ViewerEvent`. Each subtype of `ViewerEvent` inherits from `ViewerEventBase`, which defines shared fields, such as `recording_id`. Concrete event types define the payload inline.

```ts
viewer.on("selection_change", event => console.log(event.recording_id, event.items));
```

### Jupyter Notebook API changes

The `Viewer` class now has an `on_event` method, and `register_callbacks` has been removed. The `on_event` method accepts a single callback instead of a class. The callback is called with instances of `ViewerEvent`.

Just like TS, we expose some shared fields on `ViewerEvent` (by inheriting from a `ViewerEventBase` class), and event-specific payloads are placed inline.

Each event class comes with a `type` property, which may be used to discriminate the event. This form of type discrimination is something which Python type checkers and language servers understand, so it results in type narrowing inside `if/elif/else` and `match`.

```python
def on_viewer_event(event: ViewerEvent):
  if event.type == "selection_change":
    # note: `event` has type `SelectionChangeEvent`
    for item in event.items:
      pass # do something with `item`

viewer.on_event(on_viewer_event)
```


rerun_py/rerun_sdk/rerun/event.py
```diff
@@ -0,0 +1,164 @@
...
+                )
+
+        return SelectionChangeEvent(
+            application_id=app_id,
+            recording_id=recording_id,
+            items=items,
+        )
+
+    else:
+        raise ValueError(f"Unknown event type: {event_type}")
```
emilk, line 164: Pro-tip: always add quotes, or you'll tear your hair out trying to figure out why the string `'container '` doesn't match

```suggestion
        raise ValueError(f"Unknown event type: '{event_type}'")
```

examples/rust/viewer_callbacks/src/main.rs
```diff
@@ -39,31 +39,26 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            Arc::new(move |event: ViewerEvent| match event.kind {
+                ViewerEventKind::Play | ViewerEventKind::Pause => {}
+                ViewerEventKind::TimeUpdate { time } => {
+                    shared_state.lock().current_time = time.as_f64();
+                }
+                ViewerEventKind::TimelineChange {
+                    timeline_name,
+                    time,
+                } => {
+                    let mut shared_state = shared_state.lock();
```
emilk, line 53: nit: this could be moved to before the `match` so that we don't have to lock in each branch

crates/viewer/re_viewer/src/event.rs
```diff
@@ -0,0 +1,381 @@
+//! Viewer event definitions.
+//!
+//! A callback may be registered to the Viewer via [`crate::StartupOptions::on_event`]
+//! which will receive instances of [`ViewerEvent`].
+
+// NOTE: Any changes to the type definitions in this file must be replicated in:
+// - rerun_js/web-viewer/index.ts (ViewerEvent)
+// - rerun_py/rerun_sdk/rerun/event.py (ViewerEvent)
```
emilk, line 8: nice

emilk (approved):
Awesome!

crates/store/re_log_types/src/instance.rs
```diff
@@ -37,16 +37,16 @@ impl Instance {
     ///
     /// The opposite of [`Self::is_specific`].
     #[inline]
-    pub fn is_all(self) -> bool {
-        self == Self::ALL
+    pub fn is_all(&self) -> bool {
```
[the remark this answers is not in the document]
jprochazk: I needed this for `serde(skip_serializing_if)`, but I can probably just make a one-liner function in this module for that.

PR #9745, opened 2025-04-15
Revert "Fix `Mat3x3` shape inference in `.columns` api"

emilk:
* Reverts rerun-io/rerun#9569
* Re-opens https://github.com/rerun-io/rerun/issues/9489

Trying to fix CI…


emilk:
@rerun-bot full-check

oxkitsune:
Fixed in #9746

PR #9746, opened 2025-04-15
Fix shape inference for non fixed size list array types

oxkitsune:
### Related

* Fixes regression from #9569 

### What

The partitioning shape was assumed to be the product of the outer dimensions of the input tensor, however this is incorrect and slipped through. It should default to the second dimension or 1 instead.

### DIFF
--- a/crates/build/re_types_builder/src/codegen/python/mod.rs
+++ b/crates/build/re_types_builder/src/codegen/python/mod.rs
@@ -2691,7 +2691,7 @@ fn quote_columnar_methods(reporter: &Reporter, obj: &Object, objects: &Objects) 
                         # (This should have been already validated by conversion to the arrow_array)
                         batch_length = 1
                     else:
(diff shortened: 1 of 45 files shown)


oxkitsune:
@rerun-bot full-check

crates/build/re_types_builder/src/codegen/python/mod.rs
```diff
@@ -2691,7 +2691,7 @@ fn quote_columnar_methods(reporter: &Reporter, obj: &Object, objects: &Objects)
                         # (This should have been already validated by conversion to the arrow_array)
                         batch_length = 1
                     else:
-                        batch_length = int(np.prod(shape[1:])) if len(shape) > 1 else 1  # type: ignore[redundant-expr,misc]
+                        batch_length = shape[1] if len(shape) > 1 else 1  # type: ignore[redundant-expr,misc]
```
emilk, line 2694: Can we add a test case to cover the thing that failed on CI? 🙏

emilk (approved):

PR #9743, opened 2025-04-15
Move `RequestedObject` and `table_utils` to `re_dataframe_ui`

abey79:
### Related

* Part of https://github.com/rerun-io/rerun/issues/9741

### What

Move `RequestedObject` and `table_utils` to the new `re_dataframe_ui` crate.

PR #9748, opened 2025-04-15
Revert making `RequestedObject` clone

abey79:
#9743 mistakenly made `RequestedObject` clone. This makes for unsound logic though, as it relies on a one-shot channel. This PR removes `Clone` and switched to `std::mpsc` to enforce non-clonability.

Related:
- #9743


lucasmerlin (approved):
Thanks!

crates/viewer/re_dataframe_ui/src/requested_object.rs
```diff
@@ -1,9 +1,9 @@
-use crossbeam_channel::{bounded, Receiver};
+use std::sync::mpsc::{sync_channel, Receiver};
 
 use re_viewer_context::{AsyncRuntimeHandle, WasmNotSend};
 
 /// A handle to an object that is requested asynchronously.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
```
emilk, line 6: Add a comment here explaining why it _shouldn't_ be `Clone`. Or even enforce it in a test.

abey79: I'm not sure how to write that test. Do you have an example for that?

emilk: You can use https://github.com/rerun-io/rerun/pull/9748

abey79: Nice!

abey79:
ci failure is not related

PR #9751, opened 2025-04-15
Compare snippets to legacy rrds

emilk:
## Related
* Requires https://github.com/rerun-io/rerun/pull/9750
* Part of #9110

## What
This restructures the folder of .rrds that we use for backwards-compatibility checks, so that it has a `snippets` folder, matching the structure of our snippets. These are then read by `compare_snippets_output.py` and are compared to the latest output of the snippets. They should be equal to each other (since the old .rrds should be migrated on ingestion).

### Consequences
If we add a new snippet, the CI will fail until the output .rrd of that snippet is checked in to CI (using `compare_snippets_output.py --write-missing-backward-assets`)

If we _change_ a snippet (so that it outputs something different), we need to either:
* Update the file on `git lfs`
* Add a _new_ file to `git lfs` alongside the old one
* Add an opt-out flag for comparing to the file on `git lfs` to `snippets.toml`

Since this should be a rather rare occurrence (our snippets tend to stay pretty stable), any one solution will do, and I have not decided on one yet. Maybe cross that bridge if and when we get to it?

### DIFF
--- a/docs/snippets/compare_snippet_output.py
+++ b/docs/snippets/compare_snippet_output.py
@@ -11,6 +11,7 @@ import os
 import shutil
 import sys
 import time
+import subprocess
 from pathlib import Path
 from typing import Any, cast
 
@@ -198,8 +199,10 @@ def main() -> None:
 
         if args.write_missing_backward_assets:
             if not backwards_path.exists():
+                print(f"Writing new backwards-compatibility file to {backwards_path}…")
                 backwards_path.parent.mkdir(parents=True, exist_ok=True)
                 shutil.copyfile(rust_output_path, backwards_path)
+                subprocess.call(["git", "add", "-f", backwards_path])
         else:
             try:
                 # Compare old snippet files checked in to git lfs, to the newly generated ones.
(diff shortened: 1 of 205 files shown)


emilk:
@rerun-bot full-check

emilk:
I think the CI is failing because of floating point inaccuracies… but I have a draft PR for fixing that:
* https://github.com/rerun-io/rerun/pull/9750
Will try to get that in first :)

docs/snippets/compare_snippet_output.py
```diff
@@ -56,6 +58,15 @@ def extra_args(self) -> list[str]:
...
         return f"docs/snippets/all/{self.subdir}/{self.name}_{language}.rrd"
 
+    def backwards_compatibility_path(self) -> Path:
+        """
+        Files checked in to CI that tests backwards compatibility.
+
+        We use this path as a source for comparison ("are old files correctly migrated?")
+        and as a destination when these files need updating using --write-missing-backward-assets.
+        """
+        return Path(f"tests/assets/rrd/snippets/{self.subdir}/{self.name}.rrd")
```
grtlr, line 68: As discussed internally, maybe add version to the path here.

emilk: Added to the TODO of https://github.com/rerun-io/rerun/issues/9110

PR #9752, opened 2025-04-16
Re-introduce `Navigation` and lean into `DisplayMode`

grtlr:
### Related

* Brings back `Navigation` from #9586,
* which was reverted in #9735

### What

This PR basically brings back a more robust version of the `Navigation` mechanism and leans into the concept of _display modes_ to manage the state of the Rerun viewer.


abey79 (approved):
Much left to do, but this is definitely a step is the right direction. Thanks!

PR #9753, opened 2025-04-16
Update fialing test

emilk:
The test was added in
* https://github.com/rerun-io/rerun/pull/9329
and immediately broke, probably because of
* https://github.com/rerun-io/rerun/pull/9517

Still, the plot before and after shows the same thing: nothing.


emilk:
@rerun-bot full-check

jprochazk (approved):

Issue #9183, opened 2025-03-04
`custom_view` example's visualizer isn't active by default


Wumpf:
This makes the "color" view in that example appear empty!

It is possible to add the visualizer manually from the ui:

![Image](https://github.com/user-attachments/assets/3fbf25e8-b0ab-494d-bdbb-097ec7db9cd0)

Unfortunately, this gives us a close to nonsensical error message:
`Could not find datatype for component InstanceColorIndicator. Using null array as placeholder`
(we don't actually need placeholders for indicator components _ever_. And if we did they are absolutely trivial!)

TODO:
* [ ] Implement `choose_default_visualizers` on the custom view to make the visualizer show if there's colors present
* [x] Fix the error (or just make disappear)

PR #9755, opened 2025-04-17
Fix handling of custom indicator components

Wumpf:
### Related

* part of https://github.com/rerun-io/rerun/issues/9183

### What

User defined indicators like in the custom view example weren't detected as indicators, causing them to show up in edit ui and leading to needless (and confusing!) inclusion in some queries.

Note that indicators should soon ™️  go away as we start reading out component tags (which contain archetype information!) in the viewer.

### DIFF
--- a/crates/store/re_types_core/src/loggable.rs
+++ b/crates/store/re_types_core/src/loggable.rs
@@ -202,18 +202,12 @@ impl ComponentName {
 
     /// Is this an indicator component for an archetype?
     pub fn is_indicator_component(&self) -> bool {
-        (self.starts_with("rerun.components.") || self.starts_with("rerun.blueprint.components."))
-            && self.ends_with("Indicator")
+        self.ends_with("Indicator")
     }
 
     /// If this is an indicator component, for which archetype?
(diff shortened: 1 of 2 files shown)


teh-cmc (approved):

emilk (approved):

crates/store/re_types_core/src/loggable.rs
```diff
@@ -202,18 +202,12 @@ impl ComponentName {
...
 
     /// If this is an indicator component, for which archetype?
     pub fn indicator_component_archetype(&self) -> Option<String> {
-        if let Some(name) = self.strip_prefix("rerun.components.") {
-            if let Some(name) = name.strip_suffix("Indicator") {
-                return Some(name.to_owned());
-            }
-        }
-        None
+        self.strip_suffix("Indicator").map(|name| name.to_owned())
```
emilk, line 210: This changed `indicator_component_archetype` from returning a short name (`Boxed3D`) to returning a fully qualified name (`rerun.archetypes.Boxed3D`), which broke all existing uses of `indicator_component_archetype` :angr

PR #9756, opened 2025-04-17
Update instructions for running the DNA example

vutunganh:
### What
The paths to some dependencies were outdated. I am updating them so that they work with the current version (0.22.1).

### DIFF
--- a/docs/content/getting-started/data-in/python.md
+++ b/docs/content/getting-started/data-in/python.md
@@ -34,8 +34,8 @@ import rerun as rr
 
 from math import tau
 import numpy as np
-from rerun.utilities import build_color_spiral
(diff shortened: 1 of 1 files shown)


abey79:
I don't think our intent was to expose those submodules, but I do see that our re-export do not fully support modern type checking. I'm afraid we do not want to land these changes, but I will clean that up in another PR.

abey79:
This `utilities` sub-package should probably be cleaned away to begin with:
https://github.com/rerun-io/rerun/issues/9758

abey79:
Thanks for bringing this to my attention! I'm closing this in favour of:
- https://github.com/rerun-io/rerun/pull/9759

PR #9757, opened 2025-04-17
Set wheels' min python version to 3.9

abey79:
Currently, our python package is advertised as >=3.9, but our wheels still specify 3.8:

image

With this PR, our wheels should specify 3.9 as well.

related:
* #9182

### DIFF
--- a/rerun_py/Cargo.toml
+++ b/rerun_py/Cargo.toml
@@ -80,9 +80,9 @@ numpy.workspace = true
 once_cell.workspace = true
 parking_lot.workspace = true
 pyo3 = { workspace = true, features = [
-  "abi3-py39",  # Require at least Python 3.9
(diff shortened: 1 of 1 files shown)


rerun_py/Cargo.toml
```diff
@@ -80,7 +80,7 @@ numpy.workspace = true
 once_cell.workspace = true
 parking_lot.workspace = true
 pyo3 = { workspace = true, features = [
-  "abi3-py38",
+  "abi3-py39",
```
emilk, line 83: Let's make it easier to find with a text search once we want to update the minimum supported Python version again:
```suggestion
  "abi3-py39",  # Require at least Python 3.9
```

emilk (approved):

Issue #9741, opened 2025-04-15
Project: DataFusion Table Widget to improve table experiences


abey79:
## Summary description

A full-feature, visually rich, DataFusion-backed dataframe widget.

## Grounding use-cases

To be deployed in the following context:
- Server browser (show entry table, dataset partition table, table entries, etc.)
- Local tables (e.g. from `send_table()`
-  Dataframe API result

## How to describe, demo and evaluate

Goals:
- Displaying a dataframe in the viewer should offer a far better experience than using the standard dataframe visualisation tools in e.g. notebook environments.
- Sufficient styling option to be used instead of the current welcome screen.

## Tasks

- Column visibility (**NOTE**: for now, we keep the `SorbetBatch`-level column visibility handling)
  - [x] Drop down down menu for columns list and visibility overview
  - [ ] Hide from the column header
  - [ ] Select column and bulk hide 
- Column ordering (**NOTE**: for now, we keep the `SorbetBatch`-level column order handling)
  - [x] Reorder from drop down menu
  - [ ] Reorder multiple column at once
  - [ ] Drag and drop column directly (instead of using drop down menu)
- Sorting
  - [x] Sort by one column
  - [ ] Sort by multiple columns
- Filtering ([designs overview](https://rerunio.slack.com/archives/C04MTSM2U91/p1746529626139109))
  - [ ] Filter by column sub-menu
  - [ ] Filter summary in top bar?
  - Types
    - [ ] Text (full text filtering?)
    - [ ] Enum/categorical
    - [ ] Scalar (min/max/etc.)
    - [ ] bool
    - [ ] Time
    - [ ] Date (mini calendar?)
- Alternative cell visualisers
  - [x] https://github.com/rerun-io/rerun/issues/9795
  - [x] Example: show `Blob` as thumbnail
  - [ ] Example: show scalars as sparkline
- Table styling options
  - [ ] Row style: small/medium/large
  - [ ] Table style: table vs. panels (e.g. like the welcome screen
- Navigation and interactivity
  - [ ] Row hover highlight
  - [ ] Select row + side panel
  - [ ] Open one or more linked recording in tab (require tabs)

## Non-goals and won't do

This is related, but independent to the question of table blueprints.

abey79:
Closing this has no longer relevant for organisation. Short-term, actionable projects to be created instead.

PR #9764, opened 2025-04-17
Introduce a DataFusion-backed table widget

abey79:
### Related

* part of https://github.com/rerun-io/rerun/issues/9741

### What

This PR introduces `DataFusionTableWidget`, a widget capable of displaying a table in a datafusion session context, along with the necessary changes to use it for partition table display in the redap browser.

Notes:
- The core piece is the `DataFusionAdapter`, which triggers and track completion of datafusion queries (which are async), and uses egui memory to save state across frames.
- A `TableBlueprint` stuff makes an appearance. This eventually leads to the discussion of manipulating them from code, etc.
- The pre-existing `TableConfig` remains. At first sight, it seems that it should be merged with `DataFusionAdapter`, since both construct implement manipulation of the dataframe being displayed. There are good reason to keep them separate though:
  - It makes sense to keep column visibility/order at the RecordBatch/UI level and outside of the (slower) datafusion loop.
  - Eventually, I would like that logic to be implemented directly into `egui_table`, since it is potentially useful for any table user.
- Likewise, there is a small `Columns` helper. This one will definitely be merged with `TableConfig` as soon as the other use of `TableConfig` are replaced by `DataFusionTableWidget` (namely: the in-viewer table).


crates/store/re_datafusion/src/wasm_wrapper.rs
```diff
@@ -0,0 +1,39 @@
...
+{
+    let (tx, rx) = futures::channel::oneshot::channel();
+
+    wasm_bindgen_futures::spawn_local(async {
+        let _ = tx.send(f.await);
+    });
+
+    rx.then(|result| async {
+        result.unwrap_or_else(|err| Err(DataFusionError::External(err.into())))
+    })
```
lucasmerlin, line 38: I think `map` should work here, maybe it failed when we tried it because the trait wasn't imported. Then you should be able to remove the async {}

https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.map

lucasmerlin (dismissed):
Nice!

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -0,0 +1,459 @@
...
+
+            (Some(Err(err)), _) => {
+                let error = format!("Could not load table: {err}");
+                drop(dataframe);
+
+                ui.horizontal(|ui| {
+                    ui.error_label(error);
+
+                    if ui.small_icon_button(&re_ui::icons::RESET).clicked() {
+                        table_state.clear(ui);
```
abey79, line 204: make this more explicit

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -0,0 +1,459 @@
...
+            &session_ctx,
+            table_ref,
+            id,
+            initial_blueprint,
+            refresh,
+        );
+
+        let dataframe = table_state.dataframe.lock();
+
+        let sorbet_batches = match (dataframe.try_as_ref(), &table_state.last_dataframe) {
```
abey79, line 193: `last_dataframe` is confusion, rename with `record_Batch` or similar

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -0,0 +1,202 @@
...
+
+        Arc::ptr_eq(session_ctx, &other.session_ctx)
+            && table_ref == &other.table_ref
+            && blueprint == &other.blueprint
+    }
+}
+
+type RequestedDataframe = RequestedObject<Result<Vec<SorbetBatch>, DataFusionError>>;
+
+#[derive(Clone)]
```
abey79, line 105: docstring!

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -0,0 +1,202 @@
...
+
+use re_sorbet::{BatchType, SorbetBatch};
+use re_viewer_context::AsyncRuntimeHandle;
+
+use crate::table_blueprint::TableBlueprint;
+use crate::RequestedObject;
+
+/// A table blueprint along with the context required to execute the corresponding datafusion query.
+#[derive(Clone)]
+pub struct DataFusionQuery {
```
abey79, line 17: does this need to be `pub`

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -0,0 +1,202 @@
...
+            table_ref,
+            blueprint,
+        }
+    }
+
+    /// Execute the query and produce a vector of [`SorbetBatch`]s.
+    ///
+    /// Note: the future returned by this function must be `'static`, so it takes `self`. Use
+    /// `clone()` as required.
+    async fn execute(self) -> Result<Vec<SorbetBatch>, DataFusionError> {
```
abey79, line 41: use type alias

crates/viewer/re_redap_browser/src/tables_session_context.rs
```diff
@@ -0,0 +1,157 @@
...
+    ConnectionError(#[from] ConnectionError),
+
+    #[error(transparent)]
+    DataFusionError(#[from] DataFusionError),
+
+    #[error(transparent)]
+    TypeConversionError(#[from] TypeConversionError),
+}
+
+pub struct Table {
```
abey79, line 35: Remove `pub`?

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -88,6 +107,39 @@ impl Server {
...
+            let name = desc.name();
+            name.strip_prefix("rerun_")
+                .unwrap_or(name)
+                .replace('_', " ")
+        })
+        .generate_partition_links(
+            "recording link",
+            DATASET_MANIFEST_ID_FIELD_NAME,
+            self.origin.clone(),
+            EntryId::new(),
```
abey79, line 138: FIXME!!!

lucasmerlin (approved):
Awesome!

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -0,0 +1,203 @@
...
+pub struct DataFusionAdapter {
+    id: egui::Id,
+
+    /// The query used to produce the dataframe.
+    query: DataFusionQuery,
+
+    // Used to have something to display while the new dataframe is being queried.
+    pub last_sorbet_batches: Option<Vec<SorbetBatch>>,
+
+    pub requested_sorbet_batches: Arc<Mutex<RequestedSorbetBatches>>,
```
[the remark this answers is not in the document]
abey79: > This Arc<Mutex<>> is only necessary because the egui memory has a Clone bound right?

Exactly. It's not the first time I'm annoyed by it, so it would sure be great to relax this bound.

crates/viewer/re_dataframe_ui/src/requested_object.rs
```diff
@@ -22,6 +22,7 @@ impl<T: Send + 'static> RequestedObject<T> {
         let handle = Self::Pending(rx);
 
         runtime.spawn_future(async move {
+            //TODO(ab): implement cancellation using another channel
             let result = func.await;
             let _ = tx.send(result);
         });
```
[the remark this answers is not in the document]
abey79: Agreed.

https://github.com/rerun-io/rerun/issues/9836

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -0,0 +1,203 @@
...
+use re_sorbet::{BatchType, SorbetBatch};
+use re_viewer_context::AsyncRuntimeHandle;
+
+use crate::table_blueprint::TableBlueprint;
+use crate::RequestedObject;
+
+/// A table blueprint along with the context required to execute the corresponding datafusion query.
+#[derive(Clone)]
+struct DataFusionQuery {
+    session_ctx: Arc<SessionContext>,
```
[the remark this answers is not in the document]
abey79: I had a cursory look at the impl and usage in the datafusion repo. Seems like using `Arc` is a common pattern, though I agree it seems that cloning would also be ok.

crates/store/re_datafusion/src/search_provider.rs
```diff
@@ -54,25 +55,31 @@ impl GrpcStreamToTable for SearchResultsTableProvider {
...
-            .map_err(|err| DataFusionError::External(Box::new(err)))?
-            .schema();
+        let mut client = self.client.clone();
+
+        let schema = make_future_send(async move {
+            Ok::<_, DataFusionError>(
+                client
+                    .search_dataset(request)
+                    .await
+                    .map_err(|err| DataFusionError::External(Box::new(err)))?
```
timsaucer, line 65: I don't know if we get a lot of value out of carrying around pointers to errors over just converting them to a string. There's a helper function `exec_datafusion_err` that will take a format string. I often do this as `.map_err(|err| exec_datafusion_err!("{err}"))`. Not blocking by any means.

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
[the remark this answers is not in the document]
timsaucer: `Arc` is the common pattern because at some point we might start adding in some custom functions, like table functions or scalar udfs to make our life easier. We'd likely want to register those in the session context, so we wouldn't want to start cloning around.

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        if let Some(sort_by) = sort_by {
+            dataframe = dataframe.sort(vec![
+                col(&sort_by.column).sort(sort_by.direction.is_ascending(), true)
+            ])?;
+        }
+
+        // collect
+        let record_batches = dataframe.collect().await?;
```
timsaucer, line 74: If we want to move to a world where we start displaying data as it comes in, instead of doing a `collect()` call here we would execute the dataframe with a custom execution plan that handles the batches as they stream in. The down side of what we have here is that it will try to collect everything into memory and it may not fit. This is another reason we need to start testing on larger datasets (>1Tb).

crates/viewer/re_redap_browser/src/tables_session_context.rs
```diff
@@ -0,0 +1,160 @@
...
+                name: None,
+                entry_kind: None,
+            }),
+        })
+        .await?
+        .into_inner()
+        .entries
+        .into_iter()
+        .map(TryInto::try_into)
+        .collect::<Result<Vec<EntryDetails>, _>>()?;
```
timsaucer, line 115: Not blocking for this PR but if we implement a datafusion catalog provider that does this + also gets other tables available in the platform it would be a cleaner solution IMO.

timsaucer (commented):
I reviewed the datafusion relevant bits, but not the UI parts that I'm not really qualified to address. Overall I think this is a great addition! No blockers from me.

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
abey79: We brainstormed quite a bit about that with @jleibs last week. Dealing with that properly will need us to introduce some kind of `SparseDataframe` new data structure. As the name implied, this data structure would provide a representation of dataframe that could possibly be only piece-wise loaded in memory, and could possibly have an unknown total number of rows.

At frame drawing time, the ui would:
- render the `SparseDataframe` in its current state (that might include rendering placeholder/loading indicator for the parts that are not currently in memory)
- inform the `SparseDataframe` of likely future needs to be preloaded (e.g. if i'm displaying rows 100 to 120, it might be a good idea to preload rows 121 to 150)

Asynchronously, the `SparseDataframe` would trigger asynchronous tasks to load more of its content in memory, as per e.g. the pre-loading requirements. In that context, a streaming implementation like you suggest would very much make sense.

Interestingly, this exactly mirrors our current ChunkStore setup. At frame drawing time, the store is queried by our visualisers and returns whatever data it contains _at this point in time_. Asynchronously, more data may arrive at any time (including out-of-order), and data may also disappear (due to garbage collection).

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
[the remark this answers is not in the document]
abey79: I added a note/todo

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -0,0 +1,463 @@
...
+
+use crate::datafusion_adapter::DataFusionAdapter;
+use crate::table_blueprint::{PartitionLinksSpec, SortBy, SortDirection, TableBlueprint};
+use crate::table_utils::{
+    apply_table_style_fixes, cell_ui, header_ui, ColumnConfig, TableConfig, CELL_MARGIN,
+};
+use crate::DisplayRecordBatch;
+
+/// Keep track of the columns in a sorbet batch, indexed by id.
+//TODO(ab): merge this into `TableConfig` when table config is no longer used elsewhere.
```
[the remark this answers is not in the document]
abey79: When egui_table implements column visibility and ordering, I can see two level:
- egui_tables only know about column id and would provide that to the delegate (in which case we'll probably continue to maintain an id->desc map)
- egui_tables provides facilities to store per-column references to user data, and directly provide said references to the delegate (which amounts to including a generic id -> `T: 'a` map, aka exactly what this `Columns` helper is about)

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -88,6 +107,39 @@ impl Server {
...
+        re_dataframe_ui::DataFusionTableWidget::new(
+            self.tables_session_ctx.ctx.clone(),
+            egui::Id::new(&self.origin),
+            dataset.name(),
+        )
+        .title(dataset.name())
+        .title_button(ItemActionButton::new(&re_ui::icons::RESET, || {
+            let _ = ctx
+                .command_sender
+                .send(Command::RefreshCollection(self.origin.clone()));
```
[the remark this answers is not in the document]
abey79: cb05b90

Issue #9774, opened 2025-04-22
Heuristics check does not pass


teh-cmc:
![Image](https://github.com/user-attachments/assets/9ff3f829-9770-43cc-8bcc-15d9c2acf340)

emilk:
There is a 2D view in the top-left containing the `3D` entity, which makes no sense

PR #9775, opened 2025-04-22
Bug fix: prevent creating a 2D space view with an ancestor root

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/9774
* Probably caused by https://github.com/rerun-io/rerun/pull/9308


crates/viewer/re_view_spatial/tests/spawn_heuristics.rs
```diff
@@ -1,108 +1,219 @@
...
+                }
+            };
+            builder.with_archetype(row_id, time, component)
+        });
+    }
+
+    test_context
+}
 
+fn run_herustics_snapshot_test(name: &str, test_context: &TestContext) {
```
jprochazk, line 105:
```suggestion
fn run_heuristics_snapshot_test(name: &str, test_context: &TestContext) {
```

crates/viewer/re_view_spatial/src/view_2d.rs
```diff
@@ -275,32 +280,39 @@ impl ViewClass for SpatialView2D {
...
         } = self;
+        color + depth + segmentation
+    }
 
-        *image > 0 || *encoded_image > 0 || *depth > 0 || *video > 0 || *segmentation > 0
+    fn has_any_images(&self) -> bool {
+        self.total() > 0
+    }
+
+    fn count(&mut self, dims: &MaxDimensions) {
```
jprochazk, line 304: `count` sounds like it will return something
```suggestion
    fn update_counts(&mut self, dims: &MaxDimensions) {
```

Issue #9776, opened 2025-04-22
Proxy drops messages due to backpressure


emilk:
This happens when I run

> pixi run -e examples python tests/python/release_checklist/main.py

![Image](https://github.com/user-attachments/assets/6987d078-f2cb-4055-8365-33592e8d7c66)

We need to increase that channel length, or remove its bound

jprochazk:
As a temporary workaround, we increased the queue size in:
* https://github.com/rerun-io/rerun/pull/9777

CodeMan62:
fixed or not??

PR #9777, opened 2025-04-22
Increase grpc server message queue size

jprochazk:
* Part of https://github.com/rerun-io/rerun/issues/9776

### DIFF
--- a/crates/store/re_grpc_server/src/lib.rs
+++ b/crates/store/re_grpc_server/src/lib.rs
@@ -47,7 +47,7 @@ const MAX_ENCODING_MESSAGE_SIZE: usize = MAX_DECODING_MESSAGE_SIZE;
 // Channel capacity is completely arbitrary, e just want something large enough
 // to handle bursts of messages. This is roughly 1 MiB of `Msg` (excluding their contents).
 const MESSAGE_QUEUE_CAPACITY: usize =
-    (1024 * 1024 / std::mem::size_of::<Msg>()).next_power_of_two();
+    (16 * 1024 * 1024 / std::mem::size_of::<Msg>()).next_power_of_two();
 
 // TODO(jan): Refactor `serve`/`spawn` variants into a builder?
 
--- a/crates/store/re_grpc_server/src/lib.rs
+++ b/crates/store/re_grpc_server/src/lib.rs
@@ -45,7 +45,7 @@ const MAX_DECODING_MESSAGE_SIZE: usize = u32::MAX as usize;
 const MAX_ENCODING_MESSAGE_SIZE: usize = MAX_DECODING_MESSAGE_SIZE;
 
 // Channel capacity is completely arbitrary, e just want something large enough
-// to handle bursts of messages. This is roughly 1 MiB of `Msg` (excluding their contents).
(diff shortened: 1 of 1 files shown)


crates/store/re_grpc_server/src/lib.rs
```diff
@@ -44,6 +44,10 @@ pub const DEFAULT_MEMORY_LIMIT: MemoryLimit = MemoryLimit::UNLIMITED;
 const MAX_DECODING_MESSAGE_SIZE: usize = u32::MAX as usize;
 const MAX_ENCODING_MESSAGE_SIZE: usize = MAX_DECODING_MESSAGE_SIZE;
 
+// roughly 1 MiB of messages
+const MESSAGE_QUEUE_CAPACITY: usize =
+    (1024 * 1024 / std::mem::size_of::<Msg>()).next_power_of_two();
```
emilk, line 49: This effectively multiplies the previous value with 8… that seems low.

crates/store/re_grpc_server/src/lib.rs
```diff
@@ -597,9 +601,9 @@ impl MessageProxy {
     ) {
         // Channel capacity is completely arbitrary.
         // We just want something large enough to handle bursts of messages.
```
emilk, line 603: This comment should move now

crates/store/re_grpc_server/src/lib.rs
jprochazk: I can't repro your issue no matter what I try, so I'm not sure what the "proper" fix is. The previous limit was 1024 messages, and the logs you posted show around 50 messages being dropped, so I figured x8 was enough.

emilk: My screenshot show several hundred messages dropped. I rather be safe than sorry here - we can easily afford 16MiB for these pipes.

crates/store/re_grpc_server/src/lib.rs
```diff
@@ -44,6 +44,11 @@ pub const DEFAULT_MEMORY_LIMIT: MemoryLimit = MemoryLimit::UNLIMITED;
 const MAX_DECODING_MESSAGE_SIZE: usize = u32::MAX as usize;
 const MAX_ENCODING_MESSAGE_SIZE: usize = MAX_DECODING_MESSAGE_SIZE;
 
+// Channel capacity is completely arbitrary, e just want something large enough
+// to handle bursts of messages. This is roughly 1 MiB of `Msg` (excluding their contents).
+const MESSAGE_QUEUE_CAPACITY: usize =
+    (1024 * 1024 / std::mem::size_of::<Msg>()).next_power_of_two();
```
jprochazk, line 50:
```suggestion
    (16 * 1024 * 1024 / std::mem::size_of::<Msg>()).next_power_of_two();
```

crates/store/re_grpc_server/src/lib.rs
jprochazk: I'm not going to close the issue, because we should investigate _why_ that is happening instead of only increasing the queue size and hoping that is enough. This should be a temporary bandaid fix

emilk (approved):

PR #9780, opened 2025-04-23
Centralize `RedapClient` configuration/initialization/etc

teh-cmc:
This is basically my telemetry branch, but without the telemetry.

It refactors the Redap client construction to allow for arbitrary chains of middlewares, and more importantly makes sure that this exact client is used _everywhere_. This will become very important as soon as we introduce more middlewares (including auth -- apparently we've never set up auth here!).


teh-cmc:
@rerun-bot full-check

Issue #9781, opened 2025-04-23
View heuristics for graphs are completely broken


emilk:
Repro: open graph example, then run

![Image](https://github.com/user-attachments/assets/ab2ed84e-70ad-49e0-a29a-a3e312f579d4)

Reuslt:

![Image](https://github.com/user-attachments/assets/ba8f9b0d-70ec-47d3-9758-f8bdde694194)

The bug is somewhere in `fn spawn_heuristcs` in `crates/viewer/re_view_graph/src/view.rs`

PR #9785, opened 2025-04-23
Fix graph view heuristic always recommending no views

jprochazk:
* Closes https://github.com/rerun-io/rerun/issues/9781

The filter contains entities which _should not_ create views, but the filter was used to determine which entities _should_ turn into views. It only contains `+ /__properties/**`, so it could never match anything.

I believe `then_some` as a concept is at fault here, this should've been an `if` from the start :grimacing:

### DIFF
--- a/crates/viewer/re_view_graph/src/view.rs
+++ b/crates/viewer/re_view_graph/src/view.rs
@@ -105,9 +105,11 @@ impl ViewClass for GraphView {
             .get(&NodeVisualizer::identifier())
         {
             ViewSpawnHeuristics::new(maybe_visualizable.iter().cloned().filter_map(|entity| {
-                suggested_filter
-                    .matches(&entity)
-                    .then_some(RecommendedView::new_single_entity(entity))
+                if !suggested_filter.matches(&entity) {
+                    Some(RecommendedView::new_single_entity(entity))
+                } else {
+                    None
+                }
             }))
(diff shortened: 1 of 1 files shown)


crates/viewer/re_view_graph/src/view.rs
```diff
@@ -105,9 +105,11 @@ impl ViewClass for GraphView {
...
         {
             ViewSpawnHeuristics::new(maybe_visualizable.iter().cloned().filter_map(|entity| {
-                suggested_filter
-                    .matches(&entity)
-                    .then_some(RecommendedView::new_single_entity(entity))
+                if !suggested_filter.matches(&entity) {
+                    Some(RecommendedView::new_single_entity(entity))
+                } else {
+                    None
+                }
```
emilk, line 112: Wow, that is very confusing naming. `suggested_filter` should be called `ignored_entities`

emilk, line 112: let's at least avoid double negatives

```suggestion
                if suggested_filter.matches(&entity) {
                    None
                } else {
                    Some(RecommendedView::new_single_entity(entity))
                }
```

emilk: Cleaned this up in https://github.com/rerun-io/rerun/pull/9788

Issue #9782, opened 2025-04-23
Can't see a depth image stacked on top of a color image


emilk:
Our heuristics recommend putting a depth image on top of a color image, but when we do, the depth image is fully covered:

![Image](https://github.com/user-attachments/assets/c77ea400-cb4d-4e44-92de-5cc518e97f26)

For segmentation images we set an opacity, so there it works:

![Image](https://github.com/user-attachments/assets/e052f74a-b8fa-488d-bbae-d6335de91819)

So the problem is that when stacking a depth image on top of a RGB image, the depth image should get an opacity of 0.5, but it doesn't

## Related
* https://github.com/rerun-io/rerun/issues/7708

Wumpf:
Fixed in 
* https://github.com/rerun-io/rerun/pull/11014


(or earlier?)

PR #9787, opened 2025-04-23
Improve 2D heuristic fallback values for image opacity

jprochazk:
* https://github.com/rerun-io/rerun/issues/9782

Previously, only segmentation images had an opacity fallback of `0.5` when other _non-segmentation_ images were in the same view. We now keep track of the number of images per image kind, and use that to determine what the opacity should be for all image kinds that support opacity.

The images further to the front will be transparent if there are images possibly behind them.

Assuming the draw order fallback is also used, the draw order will be (front to back):
* Segmentation
* Color
* Depth <sup>(does not support opacity, so we can ignore that)</sup>

Examples of fallback opacity:
* 2x Segmentation:
  * Segmentation (1): 1.0
  * Segmentation (2): 1.0
* 1x Segmentation + 1x Color:
  * Segmentation: 0.5
  * Color: 1.0
* 1x Color + 1x Depth:
  * Color: 0.5
  * Depth: 1.0 <sup>(ignored)</sup>


jprochazk:
So, this is annoying. I noticed some snapshot tests were failing, and after updating them, now they're failing on CI. But looking at the differences, there really aren't any. Not sure what to do here.

Wumpf (approved):
lgtm!

Emil is right that we should also add this to the video visualizer just the same! Please do so before merging, thx!

The image comparision changes are strange indeed! Have you checked the diff images that were changed on ci? Ci generates a zip with before/after/diff images

Is it maybe an LFS issue (before/after not being lfs)? (doesn't look like though, it says 4 lines changed for the diff)

jprochazk:
> Have you checked the diff images that were changed on ci? 

Yes. There is some difference between how it renders on my machine and CI. Is there a chance that it's different llvmpipe versions, or maybe it isn't software rendering on my end?

jprochazk:
> Emil is right that we should also add this to the video visualizer just the same! Please do so before merging, thx!

It looks like video doesn't support `Opacity`. At a glance it seems like it could, but it feels out of scope for this PR

jprochazk:
Opened an issue for that and linked it in a comment:
* https://github.com/rerun-io/rerun/issues/9832

jprochazk:
which is proving to be a mistake, because now I'll have to wait for CI to report all the _sequential_ failures one by one...

jprochazk:
don't merge this yet, i left some temporary code in tests so i could get more failures in parallel

jprochazk:
@Wumpf 
It seems like what you pushed in [`b914586`](https://github.com/rerun-io/rerun/pull/9787/commits/b9145869509bc9e7b3696952df9236be8ee6fd2e) didn't fix the failures :thinking:

Wumpf (approved):

PR #9788, opened 2025-04-23
Refactor entity filtering in heuristics

emilk:
* Follow-up to https://github.com/rerun-io/rerun/pull/9785

I cleaned up the code in several ways:
* More clear naming
* Use a _inclusive_ filter to match stdlib (`retain`, `filter`) and to avoid double negatives
* Use a `dyn Fn` for future flexibility
* Simplified code around uses of `any/all` iterators (and maybe fixed some bugs?)
* Use `ViewSpawnHeuristics::empty` instead of `default` to be more explicit


grtlr (approved):
Nice, that is _much_ more readable!

PR #9794, opened 2025-04-24
Fix doclinks

jprochazk:
Just `scalar` -> `scalars`


jprochazk:
All `rerun.io` link check errors will clear up once this deploys

Issue #9406, opened 2025-03-27
Make C++ build fail if using a deprecated function


emilk:
If we're using a function marked `[[deprecated]]` in any of our examples or snippets, that should turn the CI red.

ishakbhatt:
Taking a look at this!

emilk:
Thank you! I did add a few compiler flags to our top-level `CMakeLists.txt`: https://github.com/rerun-io/rerun/blob/main/CMakeLists.txt#L56 but I'm not sure they catch everything.

In particular, we want `pixi run -e cpp cpp-build-all` to fail if any of our examples or snippets (`docs/snippets/**/*.cpp`) use a `class` or function marked `[[deprecated]]`. Maybe it already works, but needs proper verification.

ishakbhatt:
Status update: 

I attempted to verify that `Werror=deprecated-declarations` fails snippets or examples (I started with `arrows3d_row_updates`) and I am seeing that I can still call deprecated functions without errors.

It seems, after printing compiler flags with what is in `main`, that it's only printing compiler flags from pixi, which is generated in `.pixi` in `.pixi/envs/cpp/include`. I went ahead and changed it to `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wdeprecated -Werror=deprecated-declarations")` in the main `CMakeLists.txt` and it shows up. It is not ideal for users to have to do the same in their projects, but I am also investigating doing the same in the other `CMakeLists.txt`. 

I will make another comment here when it is fully verified! Just wanted to give an update here on what I investigated.

PR #9796, opened 2025-04-24
Correct deprecation warnings in c++ headers when using a deprecated type

timsaucer:
### Related

Part of #9406

### What

We have a mismatched set of push/pop operations for disabling deprecation warnings. We need to keep these warnings in so that we can include our entire C++ header set but still create a warning when one of these deprecated types is used.

This PR addresses the bug of the mismatched push/pop operations that was leading to not generating any deprecation warnings.


timsaucer:
@rerun-bot full-check

timsaucer:
Closing in favor of #9801

PR #9798, opened 2025-04-24
Update MSRV to 1.85

emilk:
### Why?
I want to start using `edition=2024` :)  I'll wait for that for a later PR though.


crates/build/re_build_tools/src/hashing.rs
```diff
@@ -26,7 +26,7 @@ pub fn iter_dir<'a>(
     extensions: Option<&'a [&'a str]>,
 ) -> impl Iterator<Item = PathBuf> + 'a {
     iter_dir_filtered(path, move |entry: &Path| {
-        extensions.map_or(true, |extensions| {
+        extensions.is_none_or(|extensions| {
```
abey79, line 29: those are so much better 😍

abey79 (approved):
thanks!

PR #9804, opened 2025-04-24
Direct API for setting blueprint on a notebook-embedded viewer

jleibs:
### What

This is a more ergonomic approach to using https://github.com/rerun-io/rerun/pull/9506 for blueprints.

This lets you use Blueprint objects directly without needing to save out to `.rbl` files.

Example:
```
blueprint = rrb.Spatial3DView(contents="/**")
viewer = rr.notebook.Viewer(width=1350, height=800, url="rerun+http://localhost:51234")
viewer.set_application_blueprint("DROID", blueprint)
viewer.display()
```


rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -191,6 +191,48 @@ def _add_table_id(self, record_batch: RecordBatch, table_id: str) -> RecordBatch
...
+        Parameters
+        ----------
+        application_id: str
+            The ID of the application to set the blueprint for.
+        blueprint: BlueprintLike
+            The blueprint to set for the application.
+        make_active: bool
+            Whether to make the blueprint active.
+            If `True`, the blueprint will be set as the active blueprint for the application.
+        make_default: bool
```
emilk, line 209: no need to repeat the types, and will lead to bad docs if the types do change

```suggestion
        application_id:
            The ID of the application to set the blueprint for.
        blueprint:
            The blueprint to set for the application.
        make_active:
            Whether to make the blueprint active.
            If `True`, the blueprint will be set as the active blueprint for the application.
        make_default:
```

rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -191,6 +191,48 @@ def _add_table_id(self, record_batch: RecordBatch, table_id: str) -> RecordBatch
         # Create new record batch with updated schema
         return RecordBatch.from_arrays(record_batch.columns, schema=schema)
 
+    def set_application_blueprint(
```
emilk, line 194: We should have an example/test notebook actually use this API

emilk (approved):
Looks good, but I don't want to merge this before we have a test for it (even a manual one)

rerun_py/rerun_sdk/rerun/notebook.py
abey79: added an example notebook

PR #9823, opened 2025-04-28
Remove dead proto objects and their extensions

teh-cmc:
Remove a bunch of dead proto objects that had me very confused.


teh-cmc:
@rerun-bot full-check

Issue #6889, opened 2024-07-15
Tagged components


emilk:
Consider the `Color` component as applied to a `Mesh3D` archetype. Does the `Color` specify the color of faces, edges, or vertices?

To support this, we could introduced _tagged_ components, allowing each entity to have multiple instances of the same component, if they have different tags.

Each `(Tag, ComponentName)` tuple would be its own column of data.

## References
https://www.flecs.dev/flecs/md_docs_2Relationships.html

Wumpf:
An instance of tagged component @jleibs and me discussed recently is fill out and wireframe color in a situation where both solid & fill are supported at the same time (another outcome of the discussion was that we don't want to support it right now, more on that later & elsewhere):
Let's say we have a visualizer that supports doing both at the same time:
* if only one of wireframe & fill is enabled: query _any_ color and use that
* if both are enabled: check for specific color tags and act accordingly
alternatively the tags themselves can also influence the default of what's enabled!

kpreid:
This seems like a potentially lot of added fundamental complexity in implementation and possibly API. I wonder if there's a way to get the same effect, simpler, in addition to wondering exactly what the semantics of this tag system is going to be.

* What if component types as currently known were removed entirely, and Rerun used only pairs of `(Tag, Data)`, that is, components would be like `(SolidColor, Rgba)` and `(Radius, f32)`? That's still a unique static type per component.

* Another, probably more Rust-only frame but this is a brainstorming sort of remark: what if these were just parameterized components? That is, `components::Color<Solid>` vs `components::Color<Edge>`. That way, you avoid code duplication, and don't need to extend the existing architecture of component types (but probably need fancier code generation).

teh-cmc:
We've had a long discussion about leaf (nee "out-of-tree") transforms a couple days back, during which the desire for tagged components came up _a lot_. Again. We keep running into data modeling deadlocks.

In parallel, @gavrelina  has been experimenting with the Dataframe View and has hit more of the same issues.

We keep running into data modeling deadlocks, again and again.

---

My guess is that tagged components will greatly influence the design of our data model and query language, and to an extent even the UX of the viewer itself. As such I think it's really important that we get there sooner rather than later.
It doesn't have to be perfect either: the sooner we get to play with these new tools and refine our data model with them in mind, the sooner we can start stabilizing and ultimately _(gasp)_ standardizing things.

[...]

PR #9824, opened 2025-04-28
Remove component name indirection from `ChunkComponents`

Wumpf:
### Related

* Tiny prep piece for the next step on https://github.com/rerun-io/rerun/issues/6889

### What

After talking with @teh-cmc a bit we decided that we want to do Desc -> data/etc. maps everywhere instead of adding an indirection for component names. The latter has advantages since there are plenty situations where we want to make statements about all columns with a given component name. However, it makes a lot of things a lot more complex and the indirection is likely harmful for small numbers of columns per chunk, i.e. it becomes a bit of a muddy tradeoff.

This also happens to be a great sightseeing task for me to get started with getting queries and the viewer at large over to be working with fully tagged components.

### DIFF
--- a/crates/store/re_chunk/src/chunk.rs
+++ b/crates/store/re_chunk/src/chunk.rs
@@ -66,70 +66,20 @@ pub type ChunkResult<T> = Result<T, ChunkError>;
 // ---
 
 #[derive(Debug, Clone, Default, PartialEq)]
-pub struct ChunkComponents(
-    // TODO(#6576): support non-list based columns?
-    //
-    // NOTE: The extra `ComponentName` layer is needed because it is very common to want to look
-    // for anything matching a `ComponentName`, without any further tags specified.
-    pub IntMap<ComponentName, IntMap<ComponentDescriptor, ArrowListArray>>,
-);
+pub struct ChunkComponents(pub IntMap<ComponentDescriptor, ArrowListArray>);
 
 impl ChunkComponents {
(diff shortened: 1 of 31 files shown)


teh-cmc:
@rerun-bot full-check

crates/store/re_chunk/src/chunk.rs
```diff
@@ -66,70 +66,20 @@ pub type ChunkResult<T> = Result<T, ChunkError>;
...
-    #[inline]
-    pub fn into_iter_flattened(
-        self,
-    ) -> impl Iterator<Item = (ComponentDescriptor, ArrowListArray)> {
-        self.0.into_values().flatten()
+        component_name: ComponentName,
+    ) -> impl Iterator<Item = &ArrowListArray> {
+        self.0.iter().filter_map(move |(desc, array)| {
+            (desc.component_name == component_name).then_some(array)
+        })
```
abey79, line 82: Whenever I see a hash map lookup change to a linear search one, I always wonder if we ever run into cases where this _actually_ becomes slower and not the other way around :)

abey79 (approved):
lgtm

crates/store/re_chunk/src/chunk.rs
Wumpf: yeah so hard to tell :/
Lotsa flux here on the way though. We'll have to do some perf checking after the storm is over.

Issue #9812, opened 2025-04-25
Potential regression of slower speed v0.22.1->v0.23.0


makeecat:
## Description
The overall logging speed seems to be slower, not sure if it is related to the spawn process of grpc.

## Reproduction
```
## Install dependencies
cargo install hyperfine
cargo install rerun@0.22.1
mv ~/.cargo/bin/rerun ~/.cargo/bin/rerun@0.22.1
cargo install rerun@0.23.0
mv ~/.cargo/bin/rerun ~/.cargo/bin/rerun@0.23.0

## build testbed
git clone git@github.com:makeecat/peng.git peng
git clone git@github.com:makeecat/peng.git peng-prev
## set peng-prev before bumping rerun from 0.22.1 to 0.23.0
cd peng-prev
git checkout 74b8727bac5c49b67677c08c3f47bb7b1676e802
cargo build --release
cd ../peng
cargo build --release
cd ..

## launch rerun 0.22.1
rerun@0.22.1
## launch peng-prev
hyperfine --warmup 10 -r 30 ./peng-prev/target/release/peng_quad

## launch rerun 0.23.0
rerun@0.23.0
## launch peng
hyperfine --warmup 10 -r 30 ./peng/target/release/peng_quad
```

## Result
on MacOS Apple M1, Sequoia 15.4.1:

peng-prev: 987.1ms +- 69.0ms
peng:      1.143ms += 52.0ms

on ROG STRIX15 x64, Ubuntu 22.04:

peng-prev: 812.6ms +- 27.2ms
peng:      842.8ms +- 17.7ms

jprochazk:
Hi, I unfortunately can't reproduce the perf regression on any of my machines.

We profiled similar workloads and found some performance regressions in rrd encode due to excessive allocations. I pushed https://github.com/rerun-io/rerun/pull/9826 which does reduces the number of allocations we do to be the same as in 0.22.1, a little better in some cases.

Could you set the rerun dependency in `peng` to point to that git branch, and check if it makes any difference for you?

On my end, even that doesn't yield any real difference in the numbers I'm seeing for your repro:

```
# Mac M2 (Sonoma 14.5)
peng @ 74b8727bac5c49b67677c08c3f47bb7b1676e802 + rerun 0.22.1
  751.5 ms ±  13.6 ms

peng @ main + rerun 0.23.1
  737.1 ms ±  20.5 ms

peng @ main, patched with jan/encode-perf + rerun 0.23.1
[...]

PR #9826, opened 2025-04-28
Improve encode perf

jprochazk:
* Closes https://github.com/rerun-io/rerun/issues/9812

This brings us roughly back to where we were with `0.22.1`. The issue was excessive allocations.

Prost supports zero-copy decoding for byte/string fields, but not for encoding. I had to resort to some unsafe code to remove an extra allocation during serialization to protobuf wire format, which I believe is justified due to the massive difference in maximum memory usage.

### Benchmarks

These are very crude, and the variance is quite high. I chose the worst results out of 5 runs.

<details><summary><code>0.22.1</code></summary>

```
$ git checkout 0.22.1 && git cherry-pick --no-commit 3b6c163cf2764f6dfd78b8e5976fceaed7f7e13f
$ cargo build -p minimal --release && /usr/bin/time -v ./target/release/minimal > /dev/null


User time (seconds): 8.40
System time (seconds): 0.42
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.63
Maximum resident set size (kbytes): 964012
Minor (reclaiming a frame) page faults: 251991
```

</details>

<details><summary><code>0.23.0</code></summary>

```
$ git checkout 0.23.0 && git cherry-pick --no-commit 3b6c163cf2764f6dfd78b8e5976fceaed7f7e13f
$ cargo build -p minimal --release && /usr/bin/time -v ./target/release/minimal > /dev/null

User time (seconds): 8.38
System time (seconds): 1.16
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:05.22
Maximum resident set size (kbytes): 1505304
Minor (reclaiming a frame) page faults: 936506
```

</details>

<details><summary>This branch</summary>

```
$ git checkout jan/encode-perf-benchmark
$ cargo build -p minimal --release && /usr/bin/time -v ./target/release/minimal > /dev/null

User time (seconds): 8.38
System time (seconds): 0.47
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.73
Maximum resident set size (kbytes): 1170748
Minor (reclaiming a frame) page faults: 296538
```

</details>

The most important metrics are wall clock time (how long it took for the program to exit), and max rss (the maximum amount of _physical RAM_ the program used at any single point in time). I'm also including system time and page faults, both of which together provide a hint about how much the program is asking the OS to manage memory for it.

There may be a similar regression in decode perf, but I have not looked into that. It's unlikely to be as severe though. There are definitely improvements to be made there if we feel the need for them, such as decoding from `Bytes` which is specialized to zero-copy in prost.


Wumpf:
@jprochazk can you gauge in what shape this is? Do you think it's worth pursuing shorterm or should we shelve this?

jprochazk:
We should revisit the ideas here now that we have E2E latency measurements, but I don't think it makes sense to finish this PR specifically

PR #9830, opened 2025-04-29
Fix missing feature for `custom_view`

Wumpf:
Caused `cargo check --no-default-features -p custom_view` to fail, regressed in #9825


Wumpf:
@rerun-bot full-check

jprochazk (approved):

Issue #6695, opened 2024-07-01
Do not crop tensor axes to the area where the tensor is visible


teh-cmc:
**edit**: the axis label is cropped to the area where the tensor is actually visible, which leads to unexpected visual glitches for very thin tensors.

<img width="116" alt="Image" src="https://github.com/user-attachments/assets/17151cb6-2316-47b9-b986-d552169169a7" />

---

Data disappears and reappears as you hover it or resize the viewer.

https://github.com/rerun-io/rerun/assets/2910679/c708cada-4512-4a6b-bddc-5279780d0c10

Wumpf:
the description sounds like we're missing a redraw on changing the property, but the capture looks like much more is broken? I can see there's a 1D tensor here, but I don't get that's overdrawn there upon mouse interaction

Not unlikely that aspect ratio is broken in particular for 1D tensor

teh-cmc:
100% repro:
```
git checkout main
pixi run -e examples py-build
pixi run -e examples python tests/python/release_checklist/check_1d_tensor_data.py

Go to the tensor view, switch the dimension mapping so that it goes vertical, see that part of the tensor is somehow missing.
The problem gets worse as your window's height grow larger.
```
![image](https://github.com/rerun-io/rerun/assets/2910679/3ed56f09-ed31-44b9-a63b-e6c210b79f9c)

Wumpf:
Got a very subtle repro now. Looks like this is the on-hover axis labels not fully showing
![image](https://github.com/rerun-io/rerun/assets/1220815/c084d609-0f82-43db-9c09-c772eab6c8a8)

abey79:
What you're seeing is basically the axis label slightly overlapping the tensor.

<img width="373" alt="Image" src="https://github.com/user-attachments/assets/4211bb44-188a-4113-a515-ea98786551c2" />

That label disappears on hover, thus the dynamic behaviour.

The root cause is that this label is cropped to the tensor area, which doesn't make any sense.

<img width="116" alt="Image" src="https://github.com/user-attachments/assets/17151cb6-2316-47b9-b986-d552169169a7" />

PR #9831, opened 2025-04-29
Tensor view improvements

abey79:
### Related

- Fixes https://github.com/rerun-io/rerun/issues/6695

### What

Small improvement to the tensor view:
- do not crop the axes label to the area of the tensor itself
- expand the scroll area to the entire view
- make the view hoverable and clickable

#### Before

image

image

#### After

image

image


Wumpf (approved):
yay, great to be rid of that ancient issue

PR #9839, opened 2025-04-29
Improve `set_time` error handling for large Python integers

### LINKED ISSUE #9822: `set_time` Cannot handle large Python integers
### What

With the new `set_time` APIs in 0.23, the following error became pretty common when using the python API. We should have it implicitly do the required conversions to make the Python API easier to use.

```
set_time_timestamp_nanos_since_epoch(OverflowError: Python int too large to convert to C long
```

oxkitsune:
### Related

* Closes #9822 

### What

Adds a clear error message when `set_time` receives an out-of-range timestamp.

### DIFF
--- a/rerun_py/rerun_sdk/rerun/time.py
+++ b/rerun_py/rerun_sdk/rerun/time.py
@@ -101,11 +101,18 @@ def set_time(
         )
     elif timestamp is not None:
         nanos = to_nanos_since_epoch(timestamp)
-        bindings.set_time_timestamp_nanos_since_epoch(
-            timeline,
-            nanos,
-            recording=recording.to_native() if recording is not None else None,
-        )
+        try:
+            bindings.set_time_timestamp_nanos_since_epoch(
+                timeline,
+                nanos,
+                recording=recording.to_native() if recording is not None else None,
+            )
+        except OverflowError as err:
+            raise ValueError(
+                f"set_time: `timestamp={timestamp!r}` is out of range; "
+                f"expected seconds since Unix epoch, datetime.datetime, or numpy.datetime64 "
+                f"(timeline='{timeline}')"
+            ) from err
 
 
 def to_nanos(duration: int | np.integer | float | np.float64 | timedelta | np.timedelta64) -> int:


rerun_py/rerun_sdk/rerun/time.py
```diff
@@ -101,11 +101,18 @@ def set_time(
...
+                timeline,
+                nanos,
+                recording=recording.to_native() if recording is not None else None,
+            )
+        except OverflowError as err:
+            raise ValueError(
+                f"set_time: `timestamp={timestamp!r}` is out of range; "
+                f"expected seconds since Unix epoch, datetime.datetime, or numpy.datetime64 "
+                f"(timeline='{timeline}')"
+            ) from err
```
emilk, line 115: How large integers were you passing in, and why?

And shouldn't we protect against the same problem for sequence and duration timelines too?

oxkitsune: I was passing a timestamp in milliseconds since unix epoch, e.g. '1_745_943_196_000'. But since we assume seconds, it would then be multiplied by 1e9 to convert to nanoseconds.

I don't think sequence and duration suffer from the same gotcha, so I'd argue it's not necessary.

rerun_py/rerun_sdk/rerun/time.py
```diff
@@ -101,11 +101,18 @@ def set_time(
...
-        )
+        try:
+            bindings.set_time_timestamp_nanos_since_epoch(
+                timeline,
+                nanos,
+                recording=recording.to_native() if recording is not None else None,
+            )
+        except OverflowError as err:
+            raise ValueError(
+                f"set_time: `timestamp={timestamp!r}` is out of range; "
```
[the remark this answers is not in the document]
oxkitsune: It formats the value using `repr()`, instead of `str()`.

```py
import numpy as np

a = np.int64(123)
b = 123

print(f"{a!r} {a}")
print(f"{b!r} {b}")
```

would result in:

```
np.int64(123) 123
123 123
```


Figured that this would be desired to clearly show which type is being used, since we can take in numpy integers, python integers and floats iirc.

rerun_py/rerun_sdk/rerun/time.py
oxkitsune: Yea I like this more as well. However `2_000_000_000_000` is only valid for about 8 years. 

Let's check if timestamp is bigger than `1e11` which will always mean milliseconds instead of seconds 👍

emilk (approved):

rerun_py/rerun_sdk/rerun/time.py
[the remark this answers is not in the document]
emilk: TIL!

Issue #9837, opened 2025-04-29
Search APIs don't properly resolve Component Selector against remote schema


jleibs:
The Vector/FTS Search APIs currently have inconsistent hacks that only sometimes do the right thing.

See, for example:
https://github.com/rerun-io/rerun/blob/bc0e318515768055717ef01e0ef88965ebb7a2e1/rerun_py/src/catalog/dataset.rs#L227-L232

The reasoning behind this is that the server needs to know the exact column to build the index over and search.

Now that we have an API to get the schema, we can do much better.

The logic should now be:
- Get the schema for the dataset
- Scan the schema to find a SINGLE match for the component selector
  - If there are no matches, or multiple matches this should raise an error about an ambiguous component selector.
  - Note: tagged components are going to create situations that will be inherently ambiguous (follow up with @Wumpf to make sure we are considering this in the tagged component workstream -- component Selectors will need to be expanded to be tagged-component aware.)

This logic needs to be replicated in all of:
- `create_fts_index`
- `create_vector_index`
- `fts_search`
- `vector_search`

This probably means writing some helper on dataset like `resolve_component_selector.  There is some similar logic for view contents now in `dataframe_query.rs`.

PR #9841, opened 2025-04-29
feat: add datafusion scalar UDF examples

timsaucer:
**This PR must remain in draft until the upstream datafusion version 48 is released**

# Overview

This PR adds in two DataFusion Scalar UDFs as the first building blocks of our analysis toolkit.

- `BoundedImageExtractionUdf` will take a video blob, frame ID, and 2d bounding boxes and extract the images within the bounding box
- `DepthImageToPointCloudUdf` will take a depth image and pinhole and extract a point cloud

Also provided are the wrapper functions that can be used to take a DataFusion DataFrame and pass user friendly entity paths to perform the required UDF manipulations.

There are two example jupyter notebooks demonstrating this using data from our standard examples.

## Running the examples

First you will need to download the associated RRD for the demo files indicated in each notebook.

This requires a version of `datafusion-python` that is not yet released. You can pull [from this branch](https://github.com/timsaucer/datafusion-python/tree/feat/scalar-udf-metadata) and you will need to perform a local build using

After setting up the `uv` environment per the readme in the repo, build:

```
maturin build --features substrait
```

The in your `rerun` repo you will need 

```
pixi run -e py pip install /path/to/your/datafusion-python/target/wheels/datafusion-46.0.0-cp39-abi3-macosx_11_0_arm64.whl --force-reinstall
```

Your exact .whl file will likely be different.

With that installed, you should be able to run the example jupyter notebooks.


rerun_py/rerun_sdk/rerun/utilities/datafusion/functions.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    prior_columns = df.schema().names
+    temp_path = f"{output_path}_extracted_images"
+    bounded_image_extraction = udf(rerun_bindings.datafusion.bounded_image_extraction_udf())
+
+    df = df.select(
+        *prior_columns,
+        bounded_image_extraction(
+            col(frame_column),
+            col(f"{boxes_2d_path}:Position2D"),
```
jleibs, line 167: Very related to: https://github.com/rerun-io/rerun/issues/9837

I don't think we want to depend on hard-coded component names here.

Would be great if we could build a helper with a signature like:
```
resolve_component_selector(df.schema(), boxes_2d_path, rr.archetypes.Boxes2D.centers)
```

with some magic that uses `rr.archetypes.Boxes2D.centers` to resolve the full triple (archetype, component, field), and then scans the schema to find the column that's the best match

timsaucer:
If you want to test these using the non-released datafusion that supports scalar UDFs, you will need to check out datafusion-python at the commit on [this branch](https://github.com/timsaucer/datafusion-python/tree/feat/scalar-udf-metadata) and build it's wheel.

```
git clone https://github.com/timsaucer/datafusion-python
cd datafusion-python
git checkout feat/scalar-udf-metadata
uv venv
source .venv/bin/activate
uv sync --dev --no-install-package datafusion
maturin build --release
```

This should have a wheel built in `target/wheels` and you can pip install that wheel.

Then also build this branch of rerun and you can run the examples as in `examples/python/notebook/convert_depth_image_to_point_cloud.ipynb`

Wumpf:
that sounds like a great sample to have, but looks like all this is a bit outdated by now. Closing

PR #9842, opened 2025-04-29
Make store aware of full component descriptors

Wumpf:
### Related

* Another shaky step towards #6889 

### What

Makes `re_chunk_store` (and more methods of `re_chunk`) fully aware of `ComponentDescriptor` and have most of its operations operate on `CompontentDesriptor` rather than `ComponentName`. 

For this not to ripple out to every corner I had to patch various places, most notably `re_query` and everything relating dataframe queries, to still work with component names meanwhile. Note that some of this might incur a temporary performance cost!

Some callsites have already been fully ported:
* `re_chunk_store_ui`
* `re_time_panel`, bar its `Item::ComponentPath` links which don't understand descriptors yet

Chunkbrowser after - notice that it lists components with full descriptor now
image


Wumpf:
@rerun-bot full-check

crates/store/re_chunk_store/src/query.rs
```diff
@@ -55,7 +58,7 @@ impl ChunkStore {
     }
 
     /// Retrieve all [`ComponentName`]s in the store.
```
Wumpf, line 60:
```suggestion
    /// Retrieve all [`ComponentDescriptor`]s in the store.
```

crates/store/re_query/src/latest_at.rs
```diff
@@ -39,7 +39,7 @@ fn compare_indices(lhs: (TimeInt, RowId), rhs: (TimeInt, RowId)) -> std::cmp::Or
 }
 
 impl QueryCache {
-    /// Queries for the given `component_names` using latest-at semantics.
+    /// Queries for the given [`ComponentDescriptor`]s using latest-at semantics.
```
Wumpf, line 42: not entiiiiirely true yet. it takes everything that intos into a descriptor and then only uses the name for now

crates/viewer/re_data_ui/src/item_ui.rs
```diff
@@ -457,70 +455,6 @@ fn entity_tree_stats_ui(
     }
 }
 
-/// Show a component path and make it selectable.
```
Wumpf, line 460: turned out nobody used those, so removed instead of patched

Wumpf:
main ci run failed with known issue (someone is on it)

Wumpf:
@rerun-bot full-check

teh-cmc:
Sibling so we can see whether `redap_manifest_registry` still pass:
* https://github.com/rerun-io/dataplatform/pull/662

teh-cmc (approved):
LGTM (if tests on sibling branch pass), except for the explicit descriptor passed around in addition to the compoent `C` type itself everywhere -- what's that about?

Wumpf:
> except for the explicit descriptor passed around in addition to the component C type itself everywhere -- what's that about?

(talked also on Slack about this.) The key problem is that `C::descriptor` doesn't have archetype name and field name set correctly. There are unfortunately a whole bunch of places, expecially in blueprint, that want to operate on the deserialized type, so it's not easy to get rid of those methods.

crates/store/re_query/src/latest_at.rs
```diff
@@ -60,13 +60,18 @@ impl QueryCache {
         // has non-negligible overhead even if the final result ends up being nothing, and our
         // number of queries for a frame grows linearly with the number of entity paths.
         let component_names = component_descrs.into_iter().filter_map(|component_descr| {
-            let component_descr = component_descr.into();
-            store
-                .entity_has_component_on_timeline(
-                    &query.timeline(),
+            // TODO(#6889): As an interim step we ignore the descriptor here for the moment.
+            let component_descr = store
+                .entity_component_descriptors_with_name(
```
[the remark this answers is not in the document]
Wumpf: Yeah I almost made a point of ignoring perf here. This PR gets us 100% in "please please please don't ship this as-is" state.

PR #9844, opened 2025-04-30
Update mac runners to `macos-15` and use clang 18

abey79:
### Related

- Fix CI regression introduced by #9764 

### What

#9764 pulls DataFusion to the wasm target, and old clang version struggle to deal with it (LLVM error: section too large"). In particular, it fails with the mac runners, which default to clang 15. So this PR:
- updates mac runners to 15, which include (optional) llvm 18 (note that `macos-latest` points at `macos-14`, which isn't quite the latest)
- enables llvm 18 (the default is still 15)

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -47,6 +47,7 @@ runs:
       shell: bash
 
     - name: Check clang version
+      if: ${{ runner.os == 'macOS' }}
       run: |
         clang --version
         which clang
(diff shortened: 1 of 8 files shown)


abey79:
@rerun-bot full-check

abey79:
@rerun-bot full-check

abey79:
@rerun-bot full-check

emilk:
> (note that `macos-latest` points at `macos-14`, which isn't quite the latest)

🤦

emilk (approved):

Issue #9845, opened 2025-04-30
Change `flush_blocking` to make flush timeout optional


grtlr:
**Is your feature request related to a problem? Please describe.**

Currently both `flush_blocking` in our SDKs and dropping the gRPC client use `MessageProxyClient::flush()`, which can result in either dataloss (during shutdown), or `flush_blocking` not sending all data.

**Describe the solution you'd like**

We should untangle both mechanisms. Dropping a gRPC client should result in a flush with a sensible default timeout (or specified by the user). Explicitly calling `flush_blocking` should always send all data, unless a timeout is explicitly specified by the user.

**Additional context**

* Internal discussion: https://rerunio.slack.com/archives/C033K5VS2KD/p1745938444003479
* Related issue with repro: #9818

PR #9846, opened 2025-04-30
Improve and mitigate warnings around dataloss when flushing

### LINKED ISSUE #9818: [C++ repro] Logging lots of image looses data
User-provided C++ repro highlight loss of logged data.

Expected: all data is logged, 50 frames on the timeline
Actual: partial data, only ~33 frames

Repro repo: https://github.com/ExpertOfNil/rerun_cpp_mve
Discord thread: https://discord.com/channels/1062300748202921994/1364955449602084874
Slack thread: https://rerunio.slack.com/archives/C041NHU952S/p1745826038111429

grtlr:
### Related

* Closes #9818.

### What

> [!IMPORTANT]
> This PR also changes the way `RecordingStream` is free'd in the C/C++ API. Before we called `stream.disconnect`, which unnecessarily replaced the current sink with a _buffered_ sink that would be immediately dropped afterwards. Not only did this cause spam in the log outputs, it also lead to race conditions upon (log) application shutdown.

This PR makes it more explicit why we drop data during flushing, by bumping the log messages to `warn!`. It also improves the message by pointing the users to `flush_timeout`.

We also bump the default timeout from two seconds to now 3 seconds.

It's worth taking note that explicitly calling `flush_blocking` from our SDKs should be able to opt-out of this timeout, to ensure all data is sent. This will be tracked here:

* #9845.

### DIFF
--- a/crates/top/rerun_c/src/lib.rs
+++ b/crates/top/rerun_c/src/lib.rs
@@ -512,7 +512,7 @@ thread_local! {
 pub extern "C" fn rr_recording_stream_free(id: CRecordingStream) {
     if THREAD_LIFE_TRACKER.try_with(|_v| {}).is_ok() {
         if let Some(stream) = RECORDING_STREAMS.lock().remove(id) {
-            stream.disconnect();
+            drop(stream);
         }
     } else {
         // Yes, at least as of writing we can still log things in this state!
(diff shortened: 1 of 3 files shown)


crates/top/rerun_c/src/lib.rs
```diff
@@ -512,7 +512,7 @@ thread_local! {
 pub extern "C" fn rr_recording_stream_free(id: CRecordingStream) {
     if THREAD_LIFE_TRACKER.try_with(|_v| {}).is_ok() {
         if let Some(stream) = RECORDING_STREAMS.lock().remove(id) {
-            stream.disconnect();
+            drop(stream);
```
abey79, line 515: I'm missing the implications of this change, if any.

abey79 (approved):
Lgtm, thanks.

crates/top/rerun_c/src/lib.rs
grtlr: Added the note from the PR as comment.

PR #9852, opened 2025-04-30
protos: support parallel partition manifest creation during dataset registration

andrea-reale:
Protos change to support parallel partition manifests creation in https://github.com/rerun-io/dataplatform/pull/663

### DIFF
--- a/crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
+++ b/crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
@@ -131,7 +131,7 @@ message RegisterWithDatasetRequest {
 }
 
 message RegisterWithDatasetResponse {
-  rerun.common.v1alpha1.TaskId id = 1;
+  rerun.common.v1alpha1.DataframePart data = 1;
 }
 
 // TODO(andrea): This is a copy of RegisterWithDatasetRequest.
--- a/rerun_py/src/catalog/connection_handle.rs
+++ b/rerun_py/src/catalog/connection_handle.rs
@@ -7,8 +7,10 @@ use pyo3::{
     Python,
 };
 use re_log_encoding::codec::wire::decoder::Decode as _;
+use re_protos::manifest_registry::v1alpha1::RegisterWithDatasetResponse;
 use tokio_stream::StreamExt as _;
 
+use re_arrow_util::ArrowArrayDowncastRef as _;
 use re_chunk::{LatestAtQuery, RangeQuery};
 use re_chunk_store::ChunkStore;
 use re_dataframe::ViewContentsSelector;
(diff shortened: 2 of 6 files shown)


rerun_py/src/catalog/dataset.rs
```diff
@@ -108,8 +109,8 @@ impl PyDataset {
         let mut connection = super_.client.borrow(self_.py()).connection().clone();
         let dataset_id = super_.details.id;
 
-        let task_id = connection.register_with_dataset(self_.py(), dataset_id, recording_uri)?;
-        connection.wait_for_task(self_.py(), &task_id, MAX_REGISTER_TIMEOUT)
+        let task_ids = connection.register_with_dataset(self_.py(), dataset_id, recording_uri)?;
+        connection.wait_for_tasks(self_.py(), &task_ids, MAX_REGISTER_TIMEOUT)
```
teh-cmc, line 113: Why is the timeout mandatory? Can we not make it optional and block indefinitely if it's not specified?

crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
```diff
@@ -131,7 +131,7 @@ message RegisterWithDatasetRequest {
 }
 
 message RegisterWithDatasetResponse {
-  rerun.common.v1alpha1.TaskId id = 1;
+  rerun.common.v1alpha1.DataframePart data = 1;
```
teh-cmc, line 134: Please specify the expected schema for this dataframe as an extension, similar to what we do for `CreatePartitionsManifestResponse`:
https://github.com/rerun-io/rerun/blob/1bdf400d8e0f53df1e851d2557880234b1c6cd49/crates/store/re_protos/src/v1alpha1/rerun.manifest_registry.v1alpha1.ext.rs#L283-L352

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -179,18 +180,32 @@ impl ConnectionHandle {
...
+                    PyValueError::new_err("bug: rerun_task_id column not found in response")
+                })?
+                .try_downcast_array_ref::<arrow::array::StringArray>()
+                .map_err(to_py_err)?
+                .iter()
+                .map(|v| {
+                    v.map(|id| TaskId { id: id.to_owned() })
+                        .ok_or_else(|| PyValueError::new_err("bug: found null task_id"))
+                })
+                .collect::<Result<Vec<_>, _>>()
```
teh-cmc, line 201: Once you have a schema extension defined, you can just check whether `schema()` [contains](https://docs.rs/arrow/latest/arrow/datatypes/struct.Schema.html#method.contains`) the returned dataframe, and omit further checks (i.e. just ok()/flatten() everything as needed).

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -225,23 +240,33 @@ impl ConnectionHandle {
...
-            if exec_status == "success" {
-                Ok(())
-            } else {
-                Err(PyValueError::new_err(format!("task failed: {msgs}",)))
+                .ok_or_else(|| PyValueError::new_err("bug: msgs column not found in response"))?
+                .try_downcast_array_ref::<arrow::array::StringArray>()
+                .map_err(to_py_err)?;
+
+            // report the first non success error
+            for (i, status) in exec_statuses.iter().enumerate() {
```
teh-cmc, line 258: Same here (and everywhere else), this can be simplified to `.iter().flatten()` since you know you don't expect nullability (and will confirm that's the case using `Schema::contains`).

rerun_py/src/catalog/dataset.rs
andrea-reale: Operations with infinite timeout imho create all sorts of horrible leaks and hard to debug problems. We even cap the maximum timeout server-side.

rerun_py/src/catalog/connection_handle.rs
andrea-reale: This is a different API though, so I should probably first define the schema for `QueryTasksOnCompletionResponse` etc.

While that's definitely something to do, If you don't mind I'd leave it out of this PR and tackle it here: https://github.com/rerun-io/dataplatform/issues/687

rerun_py/src/catalog/dataset.rs
```diff
@@ -108,8 +109,9 @@ impl PyDataset {
         let mut connection = super_.client.borrow(self_.py()).connection().clone();
         let dataset_id = super_.details.id;
 
-        let task_id = connection.register_with_dataset(self_.py(), dataset_id, recording_uri)?;
-        connection.wait_for_task(self_.py(), &task_id, MAX_REGISTER_TIMEOUT)
+        let task_ids =
+            connection.register_with_dataset(self_.py(), dataset_id, vec![recording_uri])?;
+        connection.wait_for_tasks(self_.py(), &task_ids, MAX_REGISTER_TIMEOUT)
```
teh-cmc, line 114:
```suggestion
        // TODO(ab): expose the timeout to python (default to MAX_REGISTER_TIMEOUT if unset)
        connection.wait_for_tasks(self_.py(), &task_ids, MAX_REGISTER_TIMEOUT)
```

teh-cmc (approved):

rerun_py/src/catalog/dataset.rs
andrea-reale: There's already a TODO with an open issue a few lines above!

PR #9854, opened 2025-04-30
Properly resolve component selectors in dataset index creation and search APIs

abey79:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/9837
* Further issue to address:
  * https://github.com/rerun-io/rerun/issues/9853
  * #9855 

### What

Initial attempt to formalise component column selector, how they are matched against a schema, and how they are expressed in our Python API. Applied on dataset index creation/search APIs.

TODO:
- [x] use `AnyComponentColumn` in APIs
- [x] cleanup and fix type stubs

PR #9860, opened 2025-05-01
Improve PR number parsing resiliency

jprochazk:
This sort of thing failed before, because there were two PR numbers in the title:
* [Fix lint error from #9854 (#9856)](https://github.com/rerun-io/rerun/commit/fba7990af2f22d98dcbbb9e7a51a6e1af7c0fc8b)

Now it won't

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,7 @@
 rerun_cpp/docs/**/*
 
 rerun_py/docs/templates/**
+rerun_py/site/**
 
 rerun_js/web-viewer/re_viewer_bg.js
 rerun_js/web-viewer/re_viewer.js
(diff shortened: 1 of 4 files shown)


.prettierignore
```diff
@@ -1,6 +1,7 @@
 rerun_cpp/docs/**/*
 
 rerun_py/docs/templates/**
+rerun_py/site/**
```
jprochazk, line 4: Every time i ran `pixi run fmt` locally prettier would complain about invalid HTML files in here, I decided to exclude it

abey79 (approved):
Nice thanks!

abey79 (approved):
Nice thanks!

Issue #9866, opened 2025-05-02
Auto major/minor grid scaling of the ground plane does not work when seen from below


abey79:
Demo:

https://github.com/user-attachments/assets/2c419675-b2a7-41b6-97a5-c51975ecaea5

abey79:
Tagging 0.23.2 because its likely a quick issue @Wumpf will not resist fixing, but not a release blocker.

PR #9867, opened 2025-05-02
Fix world grid not scaling correctly when camera is below the grid

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/9866

### What

Before:
image

After:
image


abey79 (approved):
as it was foretold :)

PR #9870, opened 2025-05-02
Series of build changes for web release to minimize generated wasm

timsaucer:
### Related

The recent addition of datafusion dependency has made our wasm size too large to upload to some places. This PR introduces a series of build options to reduce the blob size.

### What

This PR adds a build profile specific to minimizing file size. Additionally it changes the web build tool to use this profile when the `--release` parameter is set.


timsaucer:
At least part of it is failing, but I need to step away. I'll pick this back up monday morning.

crates/build/re_dev_tools/src/build_web_viewer/lib.rs
```diff
@@ -212,7 +212,9 @@ pub fn build(
             "-O2",
             "--output",
             wasm_path.as_str(),
-            "--enable-reference-types",
+            "--disable-reference-types",
```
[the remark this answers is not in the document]
timsaucer: This made a negligible change in the final tar size, so turned back on

abey79 (approved):
Lgtm. @jprochazk would you think it's worth triggering an alpha to test the build?

jprochazk:
Still need to take a closer look at how this affects the build script, adding DNM for now

crates/build/re_dev_tools/src/build_web_viewer/lib.rs
```diff
@@ -213,6 +213,8 @@ pub fn build(
             "--output",
             wasm_path.as_str(),
             "--enable-reference-types",
+            "--vacuum",
+            "--strip-debug",
```
emilk, line 217: How does this affect the `debug_symbols` option (`-g`) ?

timsaucer: I removed the `-g` option in the pixi.toml file for the release build. Turning that back on and removing the `--strip-debug` adds 1.1 Mb to the zipped file.

I guess the question is how important it is to have those in a release binary. If it's just for debugging/profiling we could add a build that is essentially what we have now that would not get uploaded anywhere besides our own testing.

emilk: For sure - what I would want is that running `pixi rerun-build-web-release` would produce a `.wasm` without debug symbols, but running `pixi rerun-build-web-release -g` would keep them intact. And with the new `--strip-debug` I don't know if the second holds true

jprochazk:
I've got some commits coming soon to fix the build script and pass `-g` through to the build again

jprochazk (approved):
gzip `.wasm` is under 9MB now, so we've got some time before we have to think about this again

PR #9871, opened 2025-05-02
Query cache is now aware of component tags, `latest_at` queries can opt-in to full tags

Wumpf:
### Related

* Another part of #6889 

### What

Next step toward tagged components in the Viewer. Major changes in this iteration:
* query cache internally uses `ComponentDescriptor` rather than `ComponentName` as its key
* latest_at method on the query cache can be used with _either_ component name or descriptor queries. component name queries behave as before and pick the first (think: random, indetermined) matching descriptor whereas exact descriptor based queries are now _exact_. This allows us to opt-in to fully tagged queries bit by bit. **Note:** `ComponentName` behavior is temporary! If we want component name based queries in the future their resulting selection should not be "random" as it is right now!
* latest_at indexes now indexed over `ComponentDescriptor` rather `ComponentName`
* some related tests now use tagged queries. Added a test for "exact" latest_at queries.

note that the less common range queries haven't been adjusted yet, but cache keys are still descriptor based


Wumpf:
@rerun-bot full-check

crates/store/re_query/src/cache.rs
```diff
@@ -337,7 +337,7 @@ impl ChunkStoreSubscriber for QueryCache {
                         let key = QueryCacheKey::new(
                             chunk.entity_path().clone(),
                             timeline,
-                            component_desc.component_name,
+                            component_desc,
```
grtlr, line 340: Very much nitpicking
```suggestion
                            component_descr,
```

grtlr (approved):
LGTM!

PR #9877, opened 2025-05-05
 Basic ability to show components that only differ by archetype/field name on same entity

Wumpf:
### Related

* another piece of #6889 

### What

Or in more technical terms: `ComponentPath` is now uses `ComponentDescriptor` rather than `ComponentName`

This makes this the first PR in the new series of tagged component PRs that has a clear visual & functionality impact on the Viewer (besides ~hidden features like the chunk browser)

image

image


crates/utils/re_smart_channel/src/lib.rs
```diff
@@ -19,6 +19,7 @@ pub use sender::Sender;
 /// Identifies in what context this smart channel was created, and who/what is holding its
 /// receiving end.
 #[derive(Clone, Debug, PartialEq, Eq, Hash, serde::Deserialize, serde::Serialize)]
+#[expect(clippy::large_enum_variant)]
```
grtlr, line 22: Why is this large all of a sudden? Ah, because `re_uri`.

crates/viewer/re_data_ui/src/instance_path.rs
```diff
@@ -199,7 +199,8 @@ fn component_list_ui(
                 }
 
                 // TODO(#6889): `ComponentPath` should always have a descriptor and we should be iterating over descriptors.
```
grtlr, line 201: Do we still need that `TODO`?

crates/viewer/re_selection_panel/src/visualizer_ui.rs
```diff
@@ -279,8 +280,17 @@ fn visualizer_components(
...
                 };
 
+                // TODO(#6889): query results should already operate on full descriptors
+                let component_descr = db
+                    .storage_engine()
+                    .store()
+                    .entity_component_descriptors_with_name(&entity_path, component_name)
+                    .into_iter()
+                    .next()
+                    .unwrap_or_else(|| ComponentDescriptor::new(component_name));
```
grtlr, line 290: Do you think it's worth to `warn!` here?

crates/viewer/re_selection_panel/src/visualizer_ui.rs
```diff
@@ -313,10 +323,23 @@ fn visualizer_components(
...
+                                .recording()
+                                .storage_engine()
+                                .store()
+                                .entity_component_descriptors_with_name(
+                                    &data_result.entity_path,
+                                    component_name,
+                                )
+                                .into_iter()
+                                .next()
+                                .unwrap_or_else(|| ComponentDescriptor::new(component_name));
```
grtlr, line 337: See comment above.

grtlr (approved):
Wonder if it makes sense to add snapshot test that logs the same component multiple times to the same entity in `re_time_panel` already?

Unfortunately, I don't think we have snapshot tests for the selection panel yet.

Other than that, LGTM!

crates/utils/re_smart_channel/src/lib.rs
Wumpf: yeah looks like I tripped a treshhold. Something in there should probably be a separate alloc, but didn't want to go there right now

crates/viewer/re_selection_panel/src/visualizer_ui.rs
Wumpf: yeah let's do that. should never happen and if it does while we're working on this we'd really want to know

crates/viewer/re_data_ui/src/instance_path.rs
Wumpf: good catch!

Issue #9843, opened 2025-04-29
Cannot Run Tensor Example in Python


Charlie-AuZone:
**Describe the bug**

**To Reproduce**
1. Use the following program:
```python
"""Create and log a tensor."""

import numpy as np
import rerun as rr

tensor = np.random.randint(0, 256, (8, 6, 3, 5), dtype=np.uint8)  # 4-dimensional tensor

rr.init("rerun_example_tensor", spawn=True)

# Log the tensor, assigning names to each dimension
rr.log("tensor", rr.Tensor(tensor, dim_names=("width", "height", "channel", "batch")))
```

**Expected behavior**
As described in the documentation: https://rerun.io/docs/reference/types/archetypes/tensor

**Screenshots**

![Image](https://github.com/user-attachments/assets/90c5bbc3-ad0c-4d12-b462-8f63d034670f)

**Backtrace**
```
RerunWarning: TensorDataBatch: AttributeError('pyarrow.lib.StructType' object has no attribute 'fields')
  rr.log("tensor", rr.Tensor(tensor, dim_names=(
```

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro 
 - 10.0.19045 Build 19045

**Rerun version**

Python 3.10
```python
import rerun as rr
print(rr.version())
```
`rerun_py 0.22.0 (extension-module nasm pypi remote web_viewer) [rustc 1.81.0 (eeb90cda1 2024-09-04), LLVM 18.1.7] x86_64-pc-windows-msvc release-0.22.0 007389f, built 2025-02-06T15:02:31Z`

```
rerun --version
rerun-cli 0.22.0-rc.1 (default grpc map_view nasm native_viewer release web_viewer) [rustc 1.81.0 (eeb90cda1 2024-09-04), LLVM 18.1.7] x86_64-pc-windows-msvc release-0.22.0 c771392, built 2025-02-06T11:22:55Z
```

My PyArrow version is 16.1.0

Charlie-AuZone:
I tried upgrading to pyarrow 20.0.0 and it looks like the issue is fixed! Thank you! I guess the advertised `pyarrow>=14.0.2` is too low.

PR #9876, opened 2025-05-04
Update version for pyarrow to 20.0

agavrel:
### Related

* Closes #9843

### What

Bumping version on pyarrow

PR #9878, opened 2025-05-05
Fix incorrectly advertised minimum supported `pyarrow` version (18.0.0 is required)

Wumpf:
### Related

* fixes https://github.com/rerun-io/rerun/issues/9843
* replaces https://github.com/rerun-io/rerun/pull/9876

### What

The rerun python sdk incorrectly advertised being compatible with pyarrow 14.0.2, when in actuality we require 18.0.0.
(This is the oldest version that seems to be working with our current codebase and is what we previously tested locally & ci via the pixi environment setting)


abey79 (approved):
thanks for looking into this!

PR #9883, opened 2025-05-05
Use full component descriptors optionally in ranges queries

Wumpf:
### Related

* Yet another step towards #6889 
* Continuation of https://github.com/rerun-io/rerun/pull/9871 for ranges

### What

Uses the same temporary `MaybeTagged` construct for range queries and makes range query results being indexed over tagged components


grtlr (approved):
🚂 _choo choo_

Wumpf:
should add a test like https://github.com/rerun-io/rerun/blob/main/crates/store/re_query/tests/latest_at.rs#L62 !

crates/store/re_chunk/src/iter.rs
```diff
@@ -103,6 +104,55 @@ impl Chunk {
...
+            let indices = izip!(std::iter::repeat(TimeInt::STATIC), self.row_ids());
+
+            if let Some(validity) = list_array.nulls() {
+                Either::Right(Either::Left(Either::Left(
+                    indices
+                        .enumerate()
+                        .filter_map(|(i, o)| validity.is_valid(i).then_some(o)),
+                )))
+            } else {
+                Either::Right(Either::Left(Either::Right(indices)))
```
grtlr, line 135: 🥁 🎺  _marching band playing_ (code is fine though)

grtlr (approved):

Issue #9829, opened 2025-04-29
More gracefully fail when grpc server binding fails


abey79:
This panic is triggered by our users. We should more gracefully handle this failure mode:

https://github.com/rerun-io/rerun/blob/66de4bba47801f2b7375c349e391723ff6acd453/crates/store/re_grpc_server/src/lib.rs#L85

PR #9884, opened 2025-05-05
Propagate `re_grpc_server` TCP error

jprochazk:
* Closes https://github.com/rerun-io/rerun/issues/9829

Using `anyhow`, because these errors are not inspectable anyway


emilk (approved):
⚠️  We should never have `panic/unwrap/expect` in our code (outside of tests): https://github.com/rerun-io/rerun/blob/main/CODE_STYLE.md#avoid-unwrap-expect-etc

Crashes are to be avoided at all costs

PR #9888, opened 2025-05-06
Use descriptors when accessing components of `ViewProperty`

grtlr:
### Related

* Part of #6889.

### What

This changes the views to query the data using `ComponentDescriptor` instead of `ComponentName`. Important to note: under the hood we still use the `ComponentName` to perform the actual lookup in the blueprint store.

When doing the review, please pay close attention that we always use the correct `blueprint::archetype` to get the descriptor, because this seems quite error prone. 🙏


Wumpf (approved):
great step forward! Couldn't spot any mistakes in the grindy part

PR #9891, opened 2025-05-06
Forbid opening PRs from `main` branch of a fork

emilk:
### Related
* Based on https://github.com/emilk/egui/pull/6899
* Tested successfully in https://github.com/emilk/egui/pull/7028

## What
Fail all PRs that are opened from the `main` branch of the fork.

## Why
PR:s opened from the `main` branch cannot be collaborated on. That is, we maintainers cannot push our own commits to it (e.g. to fix smaller problems with it before merging). This is because `main` is usually protected (and even if it isn't, it probably _should_ be, and we should not push commits directly into the `main` of some fork)

### DIFF
--- a/.github/workflows/enforce_branch_name.yml
+++ b/.github/workflows/enforce_branch_name.yml
@@ -1,0 +1,34 @@
+name: PR Branch Name Check
+
+on:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


grtlr (approved):
Oh that's nice!

PR #9895, opened 2025-05-06
Add `Dataset.register_batch` and wrapper for task ids

abey79:
### Related

* closes https://github.com/rerun-io/dataplatform/issues/689

### What

- Introduces `Dataset.register_batch()` to register multiple RRDs at once. 
- Introduces `Task` and `Tasks`, which are wrapper over one, respectively a bunch of, task ids. For now, they just have a `.wait(timeout_secs)` method.
- Add an optional `timeout_secs` argument to `Dataset.register()`

### DIFF
--- a/rerun_py/rerun_bindings/rerun_bindings.pyi
+++ b/rerun_py/rerun_bindings/rerun_bindings.pyi
@@ -1038,8 +1038,36 @@ class Dataset(Entry):
     def partition_url(self, partition_id: str) -> str:
         """Return the URL for the given partition."""
 
-    def register(self, recording_uri: str) -> None:
-        """Register a RRD URI to the dataset."""
+    def register(self, recording_uri: str, timeout_secs=60) -> None:
+        """
+        Register a RRD URI to the dataset and wait for completion.
+
+        This method registers a single recording to the dataset and blocks until the registration is
+        complete, or after a timeout (in which case, a `TimeoutError` is raised).
+
+        Parameters
+        ----------
+        recording_uri: str
+            The URI of the RRD to register
+
+        timeout_secs: int
+            The timeout after which this method returns.
+
+        """
+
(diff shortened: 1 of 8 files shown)


rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1305,6 +1333,36 @@ class DataFusionTable:
...
+        """The task id."""
+
+    def wait(self, timeout_secs: int) -> None:
+        """
+        Block until the task is completed or the timeout is reached.
+
+        A `TimeoutError` is raised if the timeout is reached.
+        """
+
+class Tasks:
```
jleibs, line 1350: @andrea-reale the design doc refers to a collection of tasks related to a user-visible feature as an "Operation," though I don't see operation exposed as a concept anywhere. It seems like it would be helpful to be able to query the state of an Operation and get back all of the associated tasks.

I wonder if that's a better abstraction here than Tasks, as it's more meaningfully tied to a single originating request.

jleibs (approved):
Definitely looking forward to the polling bits.

rerun_py/rerun_bindings/rerun_bindings.pyi
abey79: I now find myself hashing a `Vec<TaskId>` to "name" a `Tasks` object. That's another indication that an `Operation` object would be indeed convenient :)

abey79: I've opened an issue:
- https://github.com/rerun-io/dataplatform/issues/709

andrea-reale: As commented in the PR, I initially had the concept of Operation in the code, but it turned out to complicate the server-side code without much practical advantage. I think it makes sense though to have it a as a client-side construct.

PR #9899, opened 2025-05-07
Add `Tasks.status_table()` to expose to the tasks status and metadata

abey79:
### Related

- follow-up to #9895

### What

This exposes `tasks.status_table()` to return a datafusion table containing the current status of the underlying tasks. 

Basically so you can do things like this:

image

### DIFF
--- a/rerun_py/src/catalog/task.rs
+++ b/rerun_py/src/catalog/task.rs
@@ -68,6 +68,7 @@ impl PyTasks {
         Ok(())
     }
 
+    /// Return a table with the status of all tasks.
     pub fn status_table(&self, py: Python<'_>) -> PyResult<PyDataFusionTable> {
         let mut connection = self.client.borrow(py).connection().clone();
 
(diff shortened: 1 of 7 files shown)


rerun_py/src/catalog/task.rs
```diff
@@ -63,7 +68,23 @@ impl PyTasks {
         Ok(())
     }
 
-    //TODO(ab): add method to poll about status (how many are done, etc.)
+    pub fn status_table(&self, py: Python<'_>) -> PyResult<PyDataFusionTable> {
+        let mut connection = self.client.borrow(py).connection().clone();
+
+        // TODO(dataplatform/issues#709): we'd use `OperationId` here if we had it.
```
andrea-reale, line 74: Trivia: at the beginning of time I had this concept of `Operation` as a group of tasks implemented in the tasks subsystem too and exposed in the API. 

Eventually I got rid of it because it was noisy and did not add much value: it is extremely interesting to see it pop again on the client side.

andrea-reale (approved):

rerun_py/src/catalog/task.rs
abey79: I mean, maybe it could be a client-side-only construct 🤔

PR #9900, opened 2025-05-07
Save `ViewProperty` components to blueprint store with descriptors

grtlr:
### Related

* Part of #6889.

### What

This ensures that all writes to the blueprint store from `ViewProperties` are tagged.

### DIFF
--- a/crates/viewer/re_view_spatial/src/view_3d.rs
+++ b/crates/viewer/re_view_spatial/src/view_3d.rs
@@ -505,7 +505,11 @@ fn view_property_ui_grid3d(
                         .changed()
                         {
                             let color = re_types::components::Color::from(edit_color);
-                            property.save_blueprint_component(ctx, &[color]);
+                            property.save_blueprint_component(
+                                ctx,
+                                &LineGrid3D::descriptor_color(),
+                                &[color],
+                            );
                         }
                     },
                     None, // No multiline editor.
--- a/crates/viewer/re_viewer_context/src/blueprint_helpers.rs
+++ b/crates/viewer/re_viewer_context/src/blueprint_helpers.rs
@@ -60,6 +60,7 @@ impl ViewerContext<'_> {
             ));
     }
 
+    // TODO(#6889): This saves a component without tags.
     pub fn save_blueprint_component(
         &self,
         entity_path: &EntityPath,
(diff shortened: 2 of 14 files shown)


crates/viewer/re_view_spatial/src/view_3d.rs
```diff
@@ -505,7 +505,11 @@ fn view_property_ui_grid3d(
                         .changed()
                         {
                             let color = re_types::components::Color::from(edit_color);
-                            property.save_blueprint_component(ctx, &[color]);
+                            property.save_blueprint_component(
+                                ctx,
+                                &LineGrid3D::descriptor_color(),
+                                &[color],
```
Wumpf, line 511: nit: this was prexisting, but can't we do this?
```suggestion
                                &color,
```

crates/viewer/re_viewer_context/src/blueprint_helpers.rs
```diff
@@ -60,6 +60,7 @@ impl ViewerContext<'_> {
             ));
     }
 
+    // TODO(#6889): This saves a component without tags.
     pub fn save_blueprint_component(
```
Wumpf, line 64: great callout, easy to miss

crates/viewer/re_viewport_blueprint/src/view_properties.rs
```diff
@@ -190,9 +190,18 @@ impl ViewProperty {
...
     ) {
-        ctx.save_blueprint_component(&self.blueprint_store_path, components);
+        let Some(serialized) = components
+            .serialized()
+            .map(|b| b.with_descriptor_override(component_descr.clone()))
+        else {
+            re_log::warn!("could not serialize components with descriptor `{component_descr}`");
+            return;
+        };
+
```
Wumpf, line 203: suggestion for improvement: Can we do a runtime check here that the descriptor shows up in the property's archetype? I believe we have all we need for that.

I thiiiiink that check should be a panic on debug and a warn_once on release, but I'm not entirely sure about that.


---

Even further, but more controversial: at least right now `ComponentBatch` can still make a statement about its component name. We should probably also runtime check that? That bit is a bit nasty though:
* we don't want the descriptor on `ComponentBatch` because it's usually incomplete as discussed... but the `name` could still be save?
* is there a (near?) future where we _intentionally_ use a physical component type that does not match with the descriptor because we know that their compatible on the datatype/encoding layer? More likely that in those cases we're coming in with raw arrow arrays! But I'm 100% sure about that yet 😬

Wumpf (approved):
great!

Wumpf (approved):

PR #9901, opened 2025-05-07
Store and query overrides & defaults with full `ComponentDescriptor`

Wumpf:
### Related

* tagging business #6889 

### What

Fixes component defaults again (were broken on `main`).

* component ui does read and store now with fully tagged components
* default & override creation now uses fully tagged components
* _some_ operations on `ViewProperty` now store tagged components
* initial query of overrides uses now tags (default already did, causing a regression  in the interim)

Note that since lookups into query results are still not tagged, `default` spill from one tag to the other at this point. This happens because we do a single query for all the defaults which in the latest-at-cache materializes as a single chunk containing all default values for a given view.

https://github.com/user-attachments/assets/2ef257ac-0134-4a47-b77d-c15834131c28

### DIFF
--- a/crates/store/re_types_core/src/reflection.rs
+++ b/crates/store/re_types_core/src/reflection.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 
 use arrow::array::{Array as _, ArrayRef};
 
-use crate::{ArchetypeFieldName, ArchetypeName, ComponentName};
+use crate::{ArchetypeFieldName, ArchetypeName, ComponentDescriptor, ComponentName};
 
 /// A trait for code-generated enums.
 pub trait Enum:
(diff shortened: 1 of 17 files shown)


crates/store/re_types_core/src/reflection.rs
```diff
@@ -304,3 +304,15 @@ pub struct ArchetypeFieldReflection {
...
+    /// Returns the component descriptor for this field.
+    #[inline]
+    pub fn component_descriptor(&self, archetype_name: ArchetypeName) -> ComponentDescriptor {
+        ComponentDescriptor {
+            component_name: self.component_name,
+            archetype_field_name: Some(self.name),
+            archetype_name: Some(archetype_name),
+        }
+    }
+}
```
grtlr, line 318: ❤️

crates/viewer/re_viewport_blueprint/src/view_properties.rs
```diff
@@ -196,33 +196,34 @@ impl ViewProperty {
     }
 
     /// Clears a blueprint component.
-    pub fn clear_blueprint_component<C: re_types::Component>(&self, ctx: &ViewerContext<'_>) {
+    pub fn clear_blueprint_component_by_name<C: re_types::Component>(
```
grtlr, line 199:
```suggestion
    // TODO(#6889): Blueprints should be cleared by descriptor only.
    pub fn clear_blueprint_component_by_name<C: re_types::Component>(
```

grtlr (approved):
Also tested that setting defaults works again.

Nice!

PR #9907, opened 2025-05-07
Introduce `Chunk(s)WithDescriptor` to avoid some descriptor passing duplication

Wumpf:
### Related

* #6889 

### What

It's not a panacea but avoids some really silly repeated passing of descriptors.
(Which makes it harder to pick the wrong one by means of having to pick less often ;-)

### DIFF
--- a/crates/viewer/re_view/src/chunks_with_descriptor.rs
+++ b/crates/viewer/re_view/src/chunks_with_descriptor.rs
@@ -1,0 +1,85 @@
+use std::borrow::Cow;
+
+use re_chunk_store::{
+    external::re_chunk::{ChunkComponentIter, ChunkComponentSlicer},
+    Chunk,
+};
+use re_log_types::{TimeInt, TimePoint, TimelineName};
+use re_types::{Component, ComponentDescriptor, RowId};
+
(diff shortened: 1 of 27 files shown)


Wumpf:
for practical reason this utility is in `re_view` right now. But we can ofc push it further down in the crate hierarchy if it turns out useful outside of views!

crates/viewer/re_view/src/chunks_with_descriptor.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /// See [`Chunk::iter_component_timepoints`].
+    #[inline]
+    pub fn iter_component_timepoints(&self) -> impl Iterator<Item = TimePoint> + 'chunk {
+        self.chunk.iter_component_timepoints(self.descriptor)
+    }
+}
+
+impl<'chunk> std::ops::Deref for ChunkWithDescriptor<'chunk, '_> {
```
Wumpf, line 78: I added that for easy access of non descriptor related chunk things. But I should revisit that decision 🤔

grtlr (approved):
Ah, that's a nice improvement!

PR #9912, opened 2025-05-08
Disable cursor blinking in tests to fix flakeyness

lucasmerlin:
### What
Blinking cursors can cause flakey tests. The kittest Harness disables cursor blinking by default, but since we overwrite the style in `apply_style_and_install_loaders` this got reset. This PR should ensure cursor blinking is disabled in tests.


crates/viewer/re_ui/src/design_tokens.rs
```diff
@@ -85,6 +85,11 @@ impl DesignTokens {
             ..Default::default()
         };
 
+        // We preserve the text cursor blink. It's false by default if we're in a test
+        // and true by default otherwise. We don't want to interfere with this behavior
+        // (Having a blinking cursor in tests can make the tests flaky).
+        egui_style.visuals.text_cursor.blink = ctx.style().visuals.text_cursor.blink;
+
```
emilk, line 92: I think this would be more robust, in case there are (or will be) other things set special for tests (or other things)

```suggestion
let mut egui_style = Arc::unwrap_or_clone(ctx.style_of(egui::Theme::Dark));
```

PR #9913, opened 2025-05-08
Use `MaybeTagged` as an interim solution for porting view results lookups & port map view

Wumpf:
### Related

* Part of #6889 

### What

See title. This allow us to switch lookups one by one and delay porting the chunk iter utilities until we're done on porting views!


grtlr (approved):
👍

Issue #9795, opened 2025-04-24
Support alternate, per column data ui function in `DataFusionTableWidget`


abey79:
- Part of #9741 
- Hacked exemple implementation #9532

The default `component_ui` function is not always the best to use in a given table. The `DataFusionTableWidget` should  have an api to register custom UI function for a given column.

PR #9914, opened 2025-05-08
Introduce `DisplayComponentColumn` to isolate the component column ui logic

abey79:
### Related

* in prep for https://github.com/rerun-io/rerun/issues/9795

### What

Moving code around to isolate the component column ui logic in a new struct, so new features (aka alternate UI method) can be more easily added.


lucasmerlin (approved):
Nice!

PR #9916, opened 2025-05-08
Only trigger a datafusion update when the relevant part of the blueprint changes

abey79:
### Related

* in prep for https://github.com/rerun-io/rerun/issues/9795

### What

Another pure refactor to express the fact that only _part_ of a `TableBlueprint` influences/is handled by the datafusion query. Currently, that "part" is 100% of the table blueprint, but I want to add new stuff to it that does not concern datafusion. This PR makes this separation of concern more explicit.


crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -103,12 +128,12 @@ impl PartialEq for DataFusionQuery {
         let Self {
             session_ctx,
             table_ref,
-            blueprint,
+            query_data: blueprint,
```
emilk, line 131: should we use the same name here?

emilk (approved):

PR #9103, opened 2025-02-21
Update egui to latest master

lucasmerlin:
### What

I was curious how my work on egui [menus](https://github.com/emilk/egui/pull/5716) and [popups](https://github.com/emilk/egui/pull/5713) would work in rerun, so I gave it a try.

### Findings so far

- The new tooltip positioning works well unless we scroll in the scroll area. Maybe we should intersect the widget rect with the clip rect.

https://github.com/user-attachments/assets/54e7ffb1-f0a9-4f2d-8776-e0bea8777903

PR #9923, opened 2025-05-08
Fix notification popup instantly closing

lucasmerlin:
### What

#9103 broke the notification popup, this is the fix. Seems like `view_space_origin_widget_editing_ui` was also broken, but I'm not sure where that ui is.


abey79 (approved):
Thanks for the fix. It sure looks like a footgun that we should consider addressing :)

PR #9924, opened 2025-05-08
Update a typo in docs

vutunganh:
### What

Parameter `degrees` is stated to receive "radians". This fixes it.


Wumpf (approved):
good catch, thanks!

PR #9925, opened 2025-05-08
All queries to the query cache are now guaranteed to use tagged components

Wumpf:
### Related

* #6889 

### What

Major milestone: While not all indexing into chunks and query results is tagged yet, all queries themselves are now guaranteed to be tagged (ignoring some remaining workaround where we gather descriptors/tags in a dubious way).

Unfortunately it escalated a bit all over the codebase and a few places needed surprisingly elaborate fixes.

### DIFF
--- a/crates/store/re_chunk_store/src/query.rs
+++ b/crates/store/re_chunk_store/src/query.rs
@@ -152,7 +152,8 @@ impl ChunkStore {
     /// Static components are always included in the results.
     ///
     /// Returns `None` if the entity doesn't exist at all on this `timeline`.
-    pub fn all_components_on_timeline_sorted(
+    // TODO(#6889): Remove in favor of `all_components_on_timeline_sorted`.
+    pub fn all_components_on_timeline_sorted_by_name(
         &self,
         timeline: &TimelineName,
         entity_path: &EntityPath,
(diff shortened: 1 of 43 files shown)


crates/store/re_chunk_store/src/query.rs
```diff
@@ -195,6 +196,55 @@ impl ChunkStore {
         }
     }
 
+    /// Retrieve all the [`ComponentName`]s that have been written to for a given [`EntityPath`] on
+    /// the specified [`Timeline`].
+    ///
+    /// Static components are always included in the results.
+    ///
+    /// Returns `None` if the entity doesn't exist at all on this `timeline`.
+    pub fn all_components_on_timeline_sorted(
```
Wumpf, line 205: copy of the name version above. introducing this prevented further rippling

crates/viewer/re_data_ui/src/annotation_context.rs
```diff
@@ -94,9 +97,23 @@ fn annotation_info(
 ) -> Option<AnnotationInfo> {
     // TODO(#3168): this needs to use the index of the keypoint to look up the correct
     // class_id. For now we use `latest_at_component_quiet` to avoid the warning spam.
+
+    // TODO(#6889): If there's several class ids we have no idea which one to use.
+    // This code uses the first one that shows up.
+    // We should instead for a class id that is likely a sibling of the keypoint id.
```
grtlr, line 103:
```suggestion
    // We should search instead for a class id that is likely a sibling of the keypoint id.
```

crates/store/re_entity_db/src/entity_db.rs
```diff
@@ -140,10 +140,21 @@ impl EntityDb {
         self.store_info().map(|ri| &ri.application_id)
     }
 
-    pub fn recording_property<C: re_types_core::Component>(&self) -> Option<C> {
+    pub fn recording_property<C: re_types_core::Component>(
```
grtlr, line 143: There are only two properties that we query, maybe it makes sense to add direct accessors using their two descriptors?

grtlr (approved):
LGTM!

Soo nice to get rid of some of the crutches!

crates/store/re_entity_db/src/entity_db.rs
Wumpf: I'd love to but I believe you tried yourself in the past: `re_entity_db` doesn't know `re_types`

PR #9932, opened 2025-05-08
Fix missing reference to `Spatial3DView` in `Pinhole` docs

hu-po:
### What

Removes duplicate bullet point for `Spatial2DView` in `Pinhole` docs, replacing it with `Spatial3DView`.


abey79 (approved):
Thanks for making the change, lgtm!

abey79:
The test failure seem unrelated.

PR #9936, opened 2025-05-09
Add `DataFusionTableWidget::column_visibility` and use it to improve visibility defaults for the partition table

abey79:
### Related

* part of https://github.com/rerun-io/rerun/issues/9741

### What

Title.

Also fixes the column ordering, which was wrongly based on their id. And rename a bunch of things for consistency.

Current default view:

image


crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -57,7 +53,9 @@ impl Columns<'_> {
     }
 }
 
-type ColumnRenamerFn<'a> = Option<Box<dyn Fn(&ColumnDescriptorRef<'_>) -> String + 'a>>;
+pub type ColumnNameFn<'a> = Option<Box<dyn Fn(&ColumnDescriptorRef<'_>) -> String + 'a>>;
+
+pub type ColumnVisibilityFn<'a> = Option<Box<dyn Fn(&ColumnDescriptorRef<'_>) -> bool + 'a>>;
```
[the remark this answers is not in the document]
abey79: These `type`s are basically to reduce the noise in the struct definition (which clippy complained about). They are _not_ public (I removed that erroneous `pub` qualifier btw). The actual public interface is the builder method (which spell the entire signature for clarity, since they are simple enough). So since this whole option-box thing is used twice (here and in the delegate), I'd rather leave it as is.

crates/store/re_log_types/src/path/entity_path_part.rs
```diff
@@ -3,7 +3,7 @@ use re_string_interner::InternedString;
 use crate::PathParseError;
 
 pub const RESERVED_NAMESPACE_PREFIX: &str = "__";
-const PROPERTIES_PART: &str = "__properties";
+pub const PROPERTIES_PART: &str = "__properties";
```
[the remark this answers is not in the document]
abey79: makes sense

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -73,7 +71,10 @@ pub struct DataFusionTableWidget<'a> {
     /// Closure used to determine the display name of the column.
     ///
     /// Defaults to using [`ColumnDescriptorRef::name`].
-    column_renamer: ColumnRenamerFn<'a>,
+    column_name_fn: ColumnNameFn<'a>,
+
+    /// Closure used to determine the visibility of the column
```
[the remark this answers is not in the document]
abey79: default. I updated the method/field names 👍🏻

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -125,16 +129,30 @@ impl Server {
...
+                entity_path.starts_with(&std::iter::once(EntityPathPart::properties()).collect())
+            }) {
+                true
+            } else {
+                matches!(
+                    desc.short_name(),
+                    RECORDING_LINK_FIELD_NAME
+                        | DATASET_MANIFEST_ID_FIELD_NAME
+                        | DATASET_MANIFEST_REGISTRATION_TIME_FIELD_NAME
+                )
```
emilk, line 151: Matching on string here is extremely brittle and is already causing problems with https://github.com/rerun-io/rerun/pull/9983

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -115,6 +117,8 @@ impl Server {
         ui: &mut egui::Ui,
         dataset: &Dataset,
     ) {
+        const RECORDING_LINK_FIELD_NAME: &str = "recording link";
```
emilk, line 120: What is this hard-coded string? Where does it come from?

abey79: This is a column that is live injected using datafusion.

emilk: Where is it live-injected? Who gives it this name? What other piece of code needs to be updated in tandem with this piece of code?

abey79: As per the location of this const suggests, this is all local to this method. `generate_partition_links` is what triggers that column generation (it takes a column name as input), and `default_column_visibility` defines, well, what the default column visibility (and it wants that generated column to be visible by default).

crates/viewer/re_redap_browser/src/servers.rs
abey79: These columns don't have much metadata—basically just a raw record batch built ad hoc on the server. Specifically, they don't have an archetype. Again, we need a much strong system to relate record batch/dataframe column name with column descriptors, which must account for arbitrary tables sent by the server and/or users.

image

emilk: What is `Name` in the above screenshot? Is that `ComponentName`? If so, can we match against that instead of `display_name`?

PR #9937, opened 2025-05-09
Add component type to doc string of `ComponentDescriptor::descriptor_*`

grtlr:
### Related

* Part of #6889.

### What

This is to prevent me from going insane and to limit the amount of bugs we introduce.

#### Rendered

image


Wumpf (approved):
personally I find it a bit excessive, but sure if you think it helps you it will also help other people :)

grtlr:
@rerun-bot full-check

PR #9938, opened 2025-05-09
Use `ComponentDescriptor` in `HybridLatestAtResult`

grtlr:
### Related

* Part of #6889.
* Builds on #9929.
* Requires #9933.

### What

Title.


crates/viewer/re_viewport_blueprint/src/container.rs
```diff
@@ -79,15 +79,15 @@ impl ContainerBlueprint {
         // This is a required component. Note that when loading containers we crawl the subtree and so
         // cleared empty container paths may exist transiently. The fact that they have an empty container_kind
         // is the marker that the have been cleared and not an error.
-        let container_kind = results.component_instance::<ContainerKind>(0)?;
+        let container_kind = results.component_mono::<ContainerKind>()?;
```
grtlr, line 82: Can we use `component_instance` and `component_mono` interchangeably?

crates/viewer/re_view/src/results_ext.rs
```diff
@@ -41,14 +41,19 @@ pub struct HybridRangeResults<'a> {
 impl HybridLatestAtResults<'_> {
     /// Returns the [`UnitChunkShared`] for the specified [`re_types_core::Component`].
     #[inline]
-    pub fn get(&self, component_name: impl Into<ComponentName>) -> Option<&UnitChunkShared> {
+    // TODO(#6889): This method seems to be unused?
```
grtlr, line 44: Should we remove it?

crates/viewer/re_view/src/results_ext.rs
```diff
@@ -41,14 +41,19 @@ pub struct HybridRangeResults<'a> {
...
+        component_name: impl Into<ComponentName>,
+    ) -> Option<&UnitChunkShared> {
         let component_name = component_name.into();
         self.overrides
             .get_by_name(&component_name)
             .or_else(|| self.results.get_by_name(&component_name))
             .or_else(|| self.defaults.get_by_name(&component_name))
     }
 
+    // TODO(#6889): Right now, fallbacks are on a per-component basis, so it's fine to pass the component name here.
```
Wumpf, line 56: added to the list of things to do on that ticket

crates/viewer/re_viewport_blueprint/src/container.rs
Wumpf: IIRC the difference is that some (?) `component_mono` implementations also will check whether a component was truly mono and allow you to issue a warning

Wumpf (approved):

Wumpf:
was curious and poked a lil bit into the ci failure: took me a while to notice but the primary issue presented by the failing images is that it draws lines instead of points.

Which implies an issue with indicators (wtf why is that the only test failing then.. but sure usually there's no choice; have only a vague hunch about that :/). Looking at impl<A: Archetype> crate::ComponentBatch for GenericIndicatorComponentArray<A> indicators are never logged with their full descriptor, but are advertised as such.

So I quickly hacked the generated SeriesPoints::descriptor_indicator to be name only and sure enough that fixed the test.
What do do about this knowledge? Uhm.. yeah... that is definitely a Monday problem :smile:

Issue #9138, opened 2025-02-25
Default draworder set so that bounding boxes don't correctly draw.


pablovela5620:
**Describe the annoyance**

Default draworder when logging a video asset + points2d + bounding box lead to bounding box being drawn behind the frame of the video. 

*what it currently looks like (draw order 10 on bbox which was set by default)*

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/e44c4508-a5a6-43c7-b7b3-77f0df1c86c5" />

*what it should look like (draw order set to 35, 30 is the draw order of the points 2d)*

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/086e67d0-e91d-4086-9697-77d7817e5859" />

**To Reproduce**
This can be reproduced on the [pi0-lerobot repo](https://github.com/rerun-io/pi0-lerobot)
1. Follow [install instructions](https://github.com/rerun-io/pi0-lerobot?tab=readme-ov-file#installation) (git clone + pixi install)
2. Run [pose estimation example](https://github.com/rerun-io/pi0-lerobot?tab=readme-ov-file#human-pose-and-kinematics)
3. On viewer bounding box shows below image

**Expected behavior**

Bounding box should be showing above image instead of under with default draw orders

**Screenshots**

**Desktop (please complete the following information):**
 - OS: 

**Additional context**

pablovela5620:
Following up with a different example, it seems like the default draw order is being affected in this example where the keypoints are hidden behind the image. I included the [rrd here](https://huggingface.co/datasets/pablovela5620/rrd-examples/resolve/main/kpt-draw-order-failure.rrd)

https://github.com/user-attachments/assets/e9864261-dde8-4c38-92ee-7996556c3ae7

PR #9946, opened 2025-05-09
Add draw order to video (-frame-reference), fix 2d objects sometimes being behind videos

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/9138

### What

Adds a draw order to video assets which also fixes issues with bounding boxes sometimes not showing correctly on top of videos.

(Commit-by-commit friendly PR; but most of it is gen'ed, it's a simple one :))

### DIFF
--- a/crates/store/re_types/src/components/draw_order_ext.rs
+++ b/crates/store/re_types/src/components/draw_order_ext.rs
@@ -9,6 +9,9 @@ impl DrawOrder {
     pub const DEFAULT_DEPTH_IMAGE: Self = Self(Float32(-20.0));
 
     /// Draw order used for images if no draw order was specified.
+    pub const DEFAULT_VIDEO_FRAME_REFERENCE: Self = Self(Float32(-15.0));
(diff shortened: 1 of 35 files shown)


crates/store/re_types/src/components/draw_order_ext.rs
```diff
@@ -8,6 +8,9 @@ impl DrawOrder {
     /// Draw order used for depth image when they're not shown as a 3D point cloud and no draw order was specified.
     pub const DEFAULT_DEPTH_IMAGE: Self = Self(Float32(-20.0));
 
+    /// Draw order used for images if no draw order was specified.
+    pub const DEFAULT_VIDEO_FRAME_REFERENCE: Self = Self(Float32(-15.0));
```
Wumpf, line 12: figured between depth and images makes sense - it's still like an image but if there is also an image it should show on top
🤷

emilk (approved):

PR #9950, opened 2025-05-10
Fix redirect to the latest migration guide

abey79:
Title.


Wumpf (approved):
oh, thank you for cleaning after me!

Issue #9954, opened 2025-05-12
LeRobot dataloader uses untagged `Scalar` component


Wumpf:
It should use `Scalars::descriptor_scalars()` as tag instead.
The fact that this still has been around in 0.23, means that we should also add migration code that adds the tag for any untaged `Scalar`.

oxkitsune:
~~This was fixed in #9957~~

Just realized this is about migrator code 😅

PR #9957, opened 2025-05-12
Fix scalar loading for LeRobot

emilk:


Wumpf:
Fixes half of
* https://github.com/rerun-io/rerun/issues/9954

still want to patch old data!

Wumpf (approved):

Issue #3055, opened 2023-08-21
Tracking issue: Improved keyboard support


martenbjork:

abey79:
Related: emilk/egui#3272

PR #9960, opened 2025-05-12
Arrow keys can be now used to navigation and un/collapse items in the blueprint and streams trees

abey79:
### Related

* Part of #3055

### What

Now in both the blueprint tree and streams tree:
- left/right collapses/uncollapses the selected item
- up/down selects the previous/next item

**Notes**:
- This only works with single selection.
- Previously, the left/right arrows were used to move the time cursor. This now requires cmd/alt-left/right. Also, the go to start/end of timeline is now alt-left/right.

### DIFF
--- a/crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
+++ b/crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
@@ -1,5 +1,6 @@
 use egui::{Response, Ui};
 use smallvec::SmallVec;
+use std::ops::ControlFlow;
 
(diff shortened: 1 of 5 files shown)


Wumpf (approved):
loving this so much. Just wish the impl was shorter 😅

amfaber:
I am using rerun to inspect intermediate results during computer vision algorithm development - using the bare left and right keys for timeline navigation was a major feature for me, and being forced to now hold ctrl is pretty awkward. Would it be possible to add a setting that can perhaps invert these two key mappings? I.e. bare arrow keys for timeline and ctrl + keys to navigate the UI?

Issue #9270, opened 2025-03-12
Viewing sparsely logged scalars on a Visible time range relative to current time is glitchy


simon-wh:
I usually am viewing scalars being logged with the Visible Time range overridden to be `current_time -5 -> current_time -0`. When I have data that is getting continuously logged on the timeline, the viewing experience of the plot is good and very smooth. When there is data that gets logged infrequently, then the viewport is extremely glitchy. It makes things very frustrating to view.

Here's a video I took of it happening: https://drive.google.com/file/d/1-e4LdgROvE5Lv401vd5MTBLmGXlXQ5DV/view?usp=sharing

Here's a recording (.rrd) that I made to reproduce this, the video was taken on https://rerun.io/viewer using this rrd:
https://drive.google.com/file/d/1DJ5EgPQC7BdMk8IlVbFzN9fiCUGjPVWB/view?usp=sharing
[Recorded on `v0.22.1`]

The only change needed to be made from the default blueprint is to override the Visible time range like the image:
![Image](https://github.com/user-attachments/assets/ddd7189f-2e93-4254-8e17-1f7dd5d31a69)

I didn't experience this in older rerun versions ~0.15 era, been seeing this at least on >=0.20 iirc 

**Desktop (please complete the following information):**
macOS Sonoma 14.7.4
desktop + web builds

**Rerun version**

0.20.3
0.22.1

simon-wh:
This issue is very frustrating for me and I encounter it all the time. Any pointers on how to push this forward? I don't mind trying to fix it myself if it'll get it resolved soon 🙂

Wumpf:
@simon-wh  I checked your recording again and realized that the issue is orthogonal and way worse than #6105! Diving a bit into the code it looks like ever since we actually cut off the queries correctly (in https://github.com/rerun-io/rerun/pull/7133) we queried extra scalars at the start and end of the range. For points that obviously wrong, for lines we have to cut the lines to the range.
I fixed the point case right away [over here](https://github.com/rerun-io/rerun/pull/9963), that should fix the issue at hand for you - I tried your rrds with point visualizers with the fix backported to 0.23.2.

Apologies this took so long for a proper follow-up

simon-wh:
@Wumpf Thanks for looking into it! No problem, I'm just happy it's going to be fixed 😍

Wumpf:
[...]

PR #9963, opened 2025-05-13
Fix time series points querying extra points at start and end of time range (causing stuttering on playback in some scenes)

Wumpf:
### Related

* Fixes half of #9270

### What

Ever since we actually cut off the range queries correctly (in https://github.com/rerun-io/rerun/pull/7133) we queried an extra scalars at the start and end of the range. For points that obviously wrong, for lines we have to cut the lines to the range - this caused the stuttering described & demonstrated in #9270

I fixed this for points only here since it's trivial.


grtlr (approved):
LGTM!

PR #9965, opened 2025-05-13
Fix panel hide buttons saving untagged blueprint

Wumpf:
* #6889 

broke as part of tagged component effort - related to the issue of component batch pretending to know its descriptor when it actually just uses `C::descriptor` which is almost always incorrect.


emilk:
Fixes the panel hide/show buttons, which were broken

emilk (approved):

Issue #9967, opened 2025-05-14
Select multiple recordings


emilk:
Currently it is not possible to select multiple recordings in the UI using e.g. shift-click.

PR #9968, opened 2025-05-14
Save all selected recordings

emilk:
### Related
* https://github.com/rerun-io/rerun/issues/9967

### What
Select an `AppId`, then hit `Save`, and you will be asked for a folder into which all recordings of that application will be saved.


Wumpf (approved):
lgtm!

PR #9976, opened 2025-05-15
Enable debug symbols for workspace crates

teh-cmc:
I'm very scared, but it's worth trying...


teh-cmc:
I'll do the same on the DP side to keep the DX somewhat consistent

teh-cmc:
* https://github.com/rerun-io/dataplatform/pull/736

Wumpf (approved):
thank you!

PR #9979, opened 2025-05-15
Upload `.rrd` to dataplatform

grtlr:


Wumpf:
with so many things moving around in the dataplatform's interfaces we'll have to rewrite this. The ui work I did in there could be salvaged but it was quite sloppy.
-> Closing this POC

Issue #9978, opened 2025-05-15
Rename fields on ComponentDescriptor & make fields unique identifiers


emilk:
Links

* [https://rerunio.slack.com/archives/C07T0LFT5BQ/p1744384508241169](https://rerunio.slack.com/archives/C07T0LFT5BQ/p1744384508241169)
* [https://rerunio.slack.com/archives/C033K5VS2KD/p1746718029104749](https://rerunio.slack.com/archives/C033K5VS2KD/p1746718029104749)

## Surface Area

* Descriptor changes -> also isolated straight forward (in code at least)
  * Probably not to bad -> similar to recent descriptor changes
* Renaming `fbs` namespaces / codegen'ed namespaces
* Sorbet column metadata changes (requires migration code)
  * \-> Isolated; worth doing now
  * fields `rerun.archetype_field`, `rerun.archetype`, `rerun.component_name`
    * Use colons for separation (similar to Arrow's extension metadata)
    * ~~Do we add versioning information?~~
  * values `rerun:Points3D` (from `rerun.archetypes.Points3D`)
* API / SDK changes changes
* Doc changes (derived from other)
  * Manual
  * generated

## Current Plan

In Sorbet metadata keys:

* We keep the `rerun.archetype` metadata key as `rerun:archetype` \[pro/con see discussion notes below\]
* Combine `rerun.archetype` and `rerun.archetype_field` and name this `rerun:component`
  * values are the way *Rerun* standardizes archetype name + field name syntax: `Points3D:positions` (note no `rerun` prefix, which is implied by `rerun:component`; run this by Niko)
* `rerun.component` -> `rerun:component_type`
  * contents do not change? Maybe also go `rerun:Position3D`

Descriptor:

```rust
struct ComponentDescriptor {
    archetype: Option<ArchetypeName>,      // archetype_name: Option<ArchetypeName>,
    // "positions" -> "rerun:Position3D:positions", suddenly unique! Punt on the sideeffects though other than ui.
    component: ComponentIdentifier,        // former archetype_field_name: ArchetypeFieldName
    component_type: Option<ComponentType>, // former component_name: Option<ComponentName>
}
```

Corollary for **future** internal/SDK representation around how columns are identified:

* **For keying in the store & viewer we come full circle: There's a single string, just like in the ComponentName days before ComponentDescriptor**
  * \-> most places that call `x_descriptor()` will only go `x_component()`
[...]

PR #9983, opened 2025-05-15
Use archetype field name in ui instead of component name

emilk:
## Related
* Part of https://github.com/rerun-io/rerun/issues/9978

## What
Show the archetype field name instead of the component name in most places of the UI

image
image
Screenshot 2025-05-15 at 15 56 02

### DIFF
--- a/rerun_py/tests/unit/test_dataframe.py
+++ b/rerun_py/tests/unit/test_dataframe.py
@@ -254,8 +254,12 @@ class TestDataframe:
                 print(f"table.schema: {table.schema}")
                 assert table.column("my_index")[0].equals(self.expected_index0[0]), f"col_selector: {col_selector}"
                 assert table.column("my_index")[1].equals(self.expected_index1[0]), f"col_selector: {col_selector}"
-                assert table.column("/points:Points3D.positions")[0].values.equals(self.expected_pos0), f"col_selector: {col_selector}"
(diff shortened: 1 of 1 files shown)


crates/store/re_types_core/src/archetype.rs
```diff
@@ -182,6 +182,16 @@ impl ArchetypeName {
             full_name
         }
     }
+
+    /// Url to the rerun docs for this Rerun archetype.
+    pub fn doc_url(&self) -> Option<String> {
+        // This code should be correct as long as this url passes our link checker:
+        // https://rerun.io/docs/reference/types/archetypes/line_strips3d
```
Wumpf, line 189: neat "trick" to get link checking 😄

crates/viewer/re_selection_panel/src/selection_panel.rs
```diff
@@ -185,6 +186,59 @@ impl SelectionPanel {
                     }),
                 );
 
+                let ComponentDescriptor {
+                    archetype_name,
+                    archetype_field_name,
+                    component_name,
+                } = component_descriptor;
+
+                if let Some(archetype_name) = archetype_name {
```
Wumpf, line 195: Wondering if it's better to explicitly call out the archetype / field being `None` when they're none since it's quite unusual 🤔

Wumpf (approved):
great step into the right direction!
appreciate the sorting out of "column name" vs "display name" etc.

PR #9990, opened 2025-05-19
Use weaker version of `ComponentDescriptor` for indicators

grtlr:
### Related

* Parto of #6889.
* From https://github.com/rerun-io/rerun/pull/9938#issuecomment-2888808593:

> was curious and poked a lil bit into the ci failure: took me a while to notice but the primary issue presented by the failing images is that it draws lines instead of points.
> 
> Which implies an issue with indicators (wtf why is that the only test failing then.. but sure usually there's no choice; have only a vague hunch about that :/). Looking at `impl<A: Archetype> crate::ComponentBatch for GenericIndicatorComponentArray<A>` indicators are never logged with their full descriptor, but are advertised as such.
> 
> So I quickly hacked the generated `SeriesPoints::descriptor_indicator` to be name only and sure enough that fixed the test. What to do about this knowledge? Uhm.. yeah... that is definitely a Monday problem 😄

### What

Because indicators are about to go away soon (™️) anyways, making them weaker (qualification-wise) is probably the easiest choice.

Doing this in a separate PR to help with potential `git bisect` in the future.

### DIFF
--- a/crates/build/re_types_builder/src/codegen/rust/api.rs
+++ b/crates/build/re_types_builder/src/codegen/rust/api.rs
@@ -1144,7 +1144,7 @@ fn quote_trait_impls_for_archetype(reporter: &Reporter, obj: &Object) -> TokenSt
                 #[inline]
                 pub fn descriptor_indicator() -> ComponentDescriptor {
                     ComponentDescriptor {
-                        archetype_name: Some(#archetype_name.into()),
+                        archetype_name: None,
                         component_name: #indicator_component_name.into(),
                         archetype_field_name: None,
                     }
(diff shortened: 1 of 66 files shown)


Wumpf (changes requested):
actually, let's be a little bit more thorough to avoid surprises during this week - we should do the same in C++ and Python

grtlr:
It looks like C++ is only using the `component_name` for indicators already. I think the Python code, will also do the correct thing via the Rust-bindings.

crates/store/re_chunk/src/migration.rs
```diff
@@ -9,6 +9,29 @@ use re_types_core::arrow_helpers::as_array_ref;
...
+            .components
+            .0
+            .drain()
+            .map(|(mut descriptor, list_array)| {
+                if descriptor.component_name.is_indicator_component() {
+                    descriptor.archetype_name = None;
+                }
+                (descriptor, list_array)
+            })
+            .collect();
```
Wumpf, line 30: can't you do an in-place modification instead of drain + collect? Should be a lil bit more efficient.
(non-blocking comment)

Wumpf (approved):

PR #9994, opened 2025-05-19
Improve custom visualizer example

Wumpf:
### Related

* Part of #6889 

### What

Improves the custom visualizer example a bit to play nice with the new world of tagged components and improve its docs overall.

Main change is that it now only visualizes colors of 3D points. The previous behavior of picking up any color is still possible but we don't have a good way to express that in the UI, also turns out this is a lot more straight forward for an example anyways.

On top of that I fixed interaction behavior in the custom view and better unified double-clicking behavior for instances across all views.

screenshot


crates/viewer/re_viewer_context/src/viewer_context.rs
```diff
@@ -228,6 +229,18 @@ impl ViewerContext<'_> {
...
             if response.double_clicked() {
                 if let Some(item) = interacted_items.first_item() {
+                    // Double click always selects the whole instance and nothing else.
+                    let item = if let Item::DataResult(view_id, instance) = item {
+                        interacted_items = Item::DataResult(
+                            *view_id,
+                            InstancePath::entity_all(instance.entity_path.clone()),
+                        )
+                        .into();
+                        interacted_items.first_item().unwrap() // Just set it
```
abey79, line 239:
```suggestion
                        interacted_items.first_item().expect("That item was just added")
```

examples/rust/custom_view/src/points3d_color_visualizer.rs
```diff
@@ -0,0 +1,106 @@
...
+    ) -> Result<Vec<re_renderer::QueueableDrawData>, ViewSystemExecutionError> {
+        // For each entity in the view that should be displayed with the `InstanceColorSystem`…
+        for data_result in query.iter_visible_data_results(Self::identifier()) {
+            // Query components while taking into account blueprint overrides
+            // and visible history if enabled.
+            let results = data_result.query_components_with_history(
+                ctx,
+                query,
+                [&rerun::Points3D::descriptor_colors()],
+            );
```
abey79, line 61: This is actually looking pretty good now! 🤩

abey79 (approved):
Nice!

PR #9996, opened 2025-05-19
Adapt `WriteChunks` definitions

grtlr:
### Related

* Part of #9981.
* Might be superseded by #9991 (leaving the order for @abey79 to decide).

### What

This changes the definitions of `WriteChunks` to what we have discussed / demoed during the meetup. Note that merging this PR will break the dataplatform build without a sibling PR.


abey79:
This is covered in #9991

PR #9997, opened 2025-05-19
Replace `tinystl` with `stl_io` to support more stl files

oxkitsune:
### What

This replaces `tinystl` with `stl_io` in our stl parser, because some binary stl files (e.g. g1 [right_wrist_yaw_link.STL](https://github.com/unitreerobotics/unitree_ros/blob/master/robots/g1_description/meshes/right_wrist_yaw_link.STL)) could not be parsed, resulting in a warning:

```
 Failed to load mesh "../unitree_ros/robots/g1_description/meshes/right_wrist_yaw_link.STL": Error loading STL mesh: Error { kind: InvalidData, message: "stream did not contain valid UTF-8" }
```

Using `stl_io` this mesh is parsed correctly.

TODO:
- [x] Currently this uses an empty string for the mesh name as debug label, as `stl_io` does not expose this field. This is in line with the behavior of `tinystl` for meshes without a name. I opened a PR that exposes the mesh name: https://github.com/hmeyer/stl_io/pull/26

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -296,6 +296,7 @@ similar-asserts = "1.4.2"
 slotmap = { version = "1.0.6", features = ["serde"] }
 smallvec = { version = "1.0", features = ["const_generics", "union"] }
 static_assertions = "1.1"
+stl_io = "0.8.5"
 strum = { version = "0.26", features = ["derive"] }
 strum_macros = "0.26"
 sublime_fuzzy = "0.7"
@@ -310,7 +311,6 @@ time = { version = "0.3.36", default-features = false, features = [
   "wasm-bindgen",
 ] }
 tiny_http = { version = "0.12", default-features = false }
-tinystl = { version = "0.0.3", default-features = false }
 tobj = "4.0"
 tokio = { version = "1.44.2", default-features = false }
 tokio-stream = "0.1.16"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -296,7 +296,7 @@ similar-asserts = "1.4.2"
 slotmap = { version = "1.0.6", features = ["serde"] }
 smallvec = { version = "1.0", features = ["const_generics", "union"] }
 static_assertions = "1.1"
-stl_io = "0.8.5"
+stl_io = { git = "https://github.com/oxkitsune/stl_io", branch = "gijs/mesh-name" }
 strum = { version = "0.26", features = ["derive"] }
 strum_macros = "0.26"
 sublime_fuzzy = "0.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -296,7 +296,7 @@ similar-asserts = "1.4.2"
 slotmap = { version = "1.0.6", features = ["serde"] }
 smallvec = { version = "1.0", features = ["const_generics", "union"] }
 static_assertions = "1.1"
-stl_io = { git = "https://github.com/oxkitsune/stl_io", branch = "gijs/mesh-name" }
+stl_io = "0.8.5"
 strum = { version = "0.26", features = ["derive"] }
 strum_macros = "0.26"
 sublime_fuzzy = "0.7"
--- a/crates/viewer/re_renderer/src/importer/stl.rs
+++ b/crates/viewer/re_renderer/src/importer/stl.rs
@@ -1,6 +1,5 @@
 use itertools::Itertools as _;
 use smallvec::smallvec;
-use stl_io::IndexedMesh;
 
 use crate::{CpuModel, RenderContext, mesh};
 
--- a/crates/viewer/re_renderer/src/importer/stl.rs
+++ b/crates/viewer/re_renderer/src/importer/stl.rs
@@ -1,13 +1,13 @@
 use itertools::Itertools as _;
 use smallvec::smallvec;
-use tinystl::StlData;
+use stl_io::IndexedMesh;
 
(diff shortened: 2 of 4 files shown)


crates/viewer/re_renderer/src/importer/stl.rs
```diff
@@ -43,14 +42,24 @@ pub fn load_stl_from_buffer(
             .tuples::<(_, _, _)>()
             .map(glam::UVec3::from)
             .collect::<Vec<_>>(),
-        vertex_positions: bytemuck::cast_slice(&triangles).to_vec(),
+
+        vertex_positions: faces
+            .iter()
+            .flat_map(|f| {
```
Wumpf, line 48: bytemucking this no longer works here? :(
Why? Looks bytemuckable

Wumpf:
please make sure to test this with a wide variety of STL files

crates/viewer/re_renderer/src/importer/stl.rs
oxkitsune: Sadly not, because we need to grab the vertex positions from `vertices` :(

But I am not very familiar with bytemuck, so maybe there's some other secret technique.

Wumpf: ohhh wait a sec, I wasn't reading any of this properly  before. This thing has indices?! That's.. kind of bs on the surface since STL doesn't even support vertex indices (it's like the least complex format you can imagine). So I got curious and checked the loader: it _creates_ indices whenever it sees two vertices that are exactly the same.. which happens very very often in STL ofc 🤦 . It's not pretty how it's done imho, but works
https://docs.rs/stl_io/latest/src/stl_io/lib.rs.html#69

.. so what you should be really doing here is to bytemuck the face array into `triangle_indices` and then bytemuck the vertices into `vertex_positions`. And hope that this still works 😄

crates/viewer/re_renderer/src/importer/stl.rs
```diff
@@ -20,18 +20,17 @@ pub fn load_stl_from_buffer(
...
-    } = StlData::read_buffer(std::io::BufReader::new(cursor)).map_err(StlImportError::TinyStl)?;
+    let mut cursor = std::io::Cursor::new(buffer);
+    let IndexedMesh { vertices, faces } =
+        stl_io::read_stl(&mut cursor).map_err(StlImportError::StlIoError)?;
+    let num_vertices = faces.len() * 3;
 
-    let num_vertices = triangles.len() * 3;
+    // TODO(gijsd): parse name from ASCII?
+    // https://github.com/hmeyer/stl_io/pull/26
+    let name = "";
```
Wumpf, line 30: bit of a regression. We can live with that but then let's patch the whole thing up to require a `DebugLabel` being passed in.

Cargo.toml
```diff
@@ -296,6 +296,7 @@ similar-asserts = "1.4.2"
 slotmap = { version = "1.0.6", features = ["serde"] }
 smallvec = { version = "1.0", features = ["const_generics", "union"] }
 static_assertions = "1.1"
+stl_io = { git = "https://github.com/oxkitsune/stl_io", branch = "gijs/mesh-name" }
```
Wumpf, line 299: given activity on that crate, do you think this is going to land and be released soon? Otherwise we have to get it out ourselves as it's blocking our own release

crates/viewer/re_renderer/src/importer/stl.rs
```diff
@@ -20,15 +19,28 @@ pub fn load_stl_from_buffer(
...
-        name,
-        triangles,
-        normals,
-        ..
-    } = StlData::read_buffer(std::io::BufReader::new(cursor)).map_err(StlImportError::TinyStl)?;
+    let mut cursor = std::io::Cursor::new(buffer);
+    let reader = stl_io::create_stl_reader(&mut cursor).map_err(StlImportError::StlIoError)?;
+    // TODO(hmeyer/stl_io#26): parse name from ASCII?
+    // https://github.com/hmeyer/stl_io/pull/26
+    let name = reader.name().cloned().unwrap_or_default();
```
Wumpf, line 26: afaik we use this only for debug & error messages. So if we can't get a name we can force callers to provide a debug name.

Comment is confusing now since you do have a name in here due to the patched library version ;-)

crates/viewer/re_renderer/src/importer/stl.rs
```diff
@@ -20,15 +19,28 @@ pub fn load_stl_from_buffer(
...
 
-    let num_vertices = triangles.len() * 3;
+    let (normals, vertices): (Vec<_>, Vec<_>) = reader
```
Wumpf: this means we're collecting the normal array twice - here and further down when assigning `vertex_normal`, that's not great given that this can be _a lot_ of. Might as well create the final one here (bytemuck cast for vertices is cheap though)

Wumpf (changes requested):
This library makes me unhappy since I don't think its datastructures make any sense in the context of STL. But whatever, if we ever care deeply about loading performance here, we can fork it proper later. It loading more files is obviously more important!

crates/viewer/re_renderer/src/importer/stl.rs
Wumpf: can probably also just bytemuck down the line?

Wumpf (approved):

PR #9999, opened 2025-05-19
`LatestAtResults::component_mono_quiet` now requires `ComponentDescriptor`

### LINKED ISSUE #9917: `ViewCoordinates` component can be tagged with `Pinhole` or `ViewCoordinates` archetype, causing confusion
In both code and ux: It's unclear when which one is used. Code resolves this right now by querying both and giving precedence to the one on `Pinhole`

Similar to
* https://github.com/rerun-io/rerun/issues/9889

grtlr:
### Related

* Part of #6889.
* Closes #9917.

### What

Title. @Wumpf please double check if this is the logic that you had in mind for #9917. 🙏

### DIFF
--- a/crates/viewer/re_view_spatial/src/pinhole.rs
+++ b/crates/viewer/re_view_spatial/src/pinhole.rs
@@ -99,7 +99,12 @@ pub fn query_pinhole_and_view_coordinates_from_store_without_blueprint(
         [
             &archetypes::Pinhole::descriptor_image_from_camera(),
             &archetypes::Pinhole::descriptor_resolution(),
+            // Note that `components::ViewCoordinates` is somewhat special, in that for convenience it can
+            // be specified in multiple places (i.e. `archetypes`). This used to be fine, but got quite a
+            // bit more cumbersome with fully-qualified component descriptors. Because of this, we now have
+            // to query using descriptors from a "secondary" archetype.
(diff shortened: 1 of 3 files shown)


crates/viewer/re_view_spatial/src/pinhole.rs
```diff
@@ -103,16 +103,22 @@ pub fn query_pinhole_and_view_coordinates_from_store_without_blueprint(
...
         .unwrap_or_else(|| {
             resolution_of_image_at(ctx, query, entity_path).unwrap_or([100.0, 100.0].into())
         });
-    let camera_xyz = query_results
-        .component_mono_quiet::<components::ViewCoordinates>()
+    let camera_xyz: components::ViewCoordinates = query_results
+        .component_mono_quiet(&archetypes::Pinhole::descriptor_camera_xyz())
+        .or_else(|| {
+            query_results.component_mono_quiet(&archetypes::ViewCoordinates::descriptor_xyz())
+        })
```
Wumpf, line 121: try using the utility `query_view_coordinates` and (!!) make sure that the query contains the view coordinates archetype descriptor

Wumpf: correction: you can't do the former since that thing does new queries. Your code here is a lot better, but still needs the right components on query_results. Not sure what's the best way to proceeed otherwise, quite messy :)

PR #10001, opened 2025-05-19
Add `/env/` to `.gitignore`

grtlr:
Some of our tooling (for example `pixi run lint-rerun`) relies on `.gitignore`. By adding `/env/` here we prevent runing unneeded lints on these files.


Wumpf (approved):
never used `./env/` for my env, but sure why not

PR #10003, opened 2025-05-19
Allow editing existing recording properties, like name

emilk:
### Related
* Pulled out from https://github.com/rerun-io/rerun/pull/9970

### What
If a recording has a name, you can now edit it in the selection panel.

If a recording has no name, there is no way to add one. That would require design, and thus a lot more work, and is out-of-scope for this cycle.

### DIFF
--- a/crates/viewer/re_data_ui/src/instance_path.rs
+++ b/crates/viewer/re_data_ui/src/instance_path.rs
@@ -32,17 +32,23 @@ impl DataUi for InstancePath {
             instance,
         } = self;
 
-        if !db.is_known_entity(entity_path) {
+        // NOTE: the passed in `db` is usually the recording; NOT the blueprint.
+        let component = if ctx.recording().is_known_entity(entity_path) {
+            // We are looking at an entity in the recording
+            ctx.recording_engine()
+                .store()
+                .all_components_on_timeline(&query.timeline(), entity_path)
+        } else if ctx.blueprint_db().is_known_entity(entity_path) {
+            // We are looking at an entity in the blueprint
+            ctx.blueprint_db()
+                .storage_engine()
+                .store()
+                .all_components_on_timeline(&query.timeline(), entity_path)
+        } else {
             ui.error_label(format!("Unknown entity: {entity_path:?}"));
             return;
-        }
-
-        let component_descriptors = db
-            .storage_engine()
-            .store()
-            .all_components_on_timeline(&query.timeline(), entity_path);
-
-        let Some(component_descriptors) = component_descriptors else {
+        };
+        let Some(components) = component else {
             // This is fine - e.g. we're looking at `/world` and the user has only logged to `/world/car`.
             ui_layout.label(
                 ui,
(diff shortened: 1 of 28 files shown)


crates/viewer/re_viewer_context/src/test_context.rs
```diff
@@ -474,13 +474,21 @@ impl TestContext {
...
+                                .expect("Updating the recording chunk store failed");
+                        }
+                    } else if store_id == self.blueprint_store.store_id() {
+                        for chunk in chunks {
+                            self.blueprint_store
+                                .add_chunk(&Arc::new(chunk))
+                                .expect("Updating the blueprint chunk store failed");
+                        }
+                    } else {
+                        panic!("Unknown store id: {store_id:?}");
```
Wumpf, line 491: I think the assert that this panic replaces was wrong: system commands come in with a delay and _who knows_ maybe the store was removed by now! (sounds like a recipe for future mystery bug!). I'm ok with a debug-only panic (:

crates/viewer/re_viewer/src/app.rs
```diff
@@ -565,29 +565,36 @@ impl App {
...
                 egui_ctx.request_repaint(); // Many changes take a frame delay to show up.
             }
-            SystemCommand::UpdateBlueprint(blueprint_id, chunks) => {
+
+            SystemCommand::AppendToStore(store_id, chunks) => {
                 re_log::trace!(
-                    "Update blueprint entities: {}",
+                    "Update entities: {}",
                     chunks.iter().map(|c| c.entity_path()).join(", ")
                 );
```
Wumpf, line 573: before it was clear where we'd write to, now it would be nice to add which store we're writing to the trace

Wumpf (approved):
code looks good!

I'm lil bit nervous about actually landing edits to the store, but this is very isolated so whatcouldpossiblygowrong.

Should we disallow edits for stores that are "on" data platform stores? The problem is that the ui suggestions that one is editing properties of a recording that lives on a remote server but in actuality we can't persist any of this yet other than saving it to an rrd.

crates/viewer/re_data_ui/src/instance_path.rs
```diff
@@ -32,22 +32,17 @@ impl DataUi for InstancePath {
...
-                .store()
-                .all_components_on_timeline(&query.timeline(), entity_path)
-        } else if ctx.blueprint_db().is_known_entity(entity_path) {
-            // We are looking at an entity in the blueprint
-            ctx.blueprint_db()
-                .storage_engine()
-                .store()
-                .all_components_on_timeline(&query.timeline(), entity_path)
-        } else {
+        if !db.is_known_entity(entity_path) {
```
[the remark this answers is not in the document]
emilk: No indeed it does not. The passed in recording seems to always the same as `ctx.recording()`… which makes me wonder why we have both

crates/viewer/re_viewer_context/src/test_context.rs
Wumpf: ah, missed that

PR #10004, opened 2025-05-19
Allow selecting multiple recordings

emilk:
### What
Allow selecting multiple recording with cmd-click.


Wumpf:
I think this makes the behavior on what recording is activated a bit weird: you can now activate a recording by deselecting it:


https://github.com/user-attachments/assets/723c6543-6ec9-41c4-8a8f-1e44e3574c90


Not a huge deal I guess, but feels awkward to me

emilk:
Not great; not terrible

Wumpf (approved):
ah well, the more things run through `handle_select_hover_drag_interactions` the better 😄

Issue #9889, opened 2025-05-06
Resolve tag component issues with `Boxes3D`/`Ellipsoids3D`/`Capsules3D`


Wumpf:
`Boxes3D`, `Ellipsoids3D` and `Capsules3D` all repeat the components `PoseTranslation3D`, `PoseRotationAxisAngle` and `PoseRotationQuat` in their archetype definition. It’s documented as
```/// Note that orienting and placing the box is handled via `[archetypes.InstancePoses3D]`.
/// Some of its component are repeated here for convenience.```

This was already dubious starting in 0.22 since you couldn’t change a previously logged `Boxes3D::centers` with a later logged `InstancePoses3D::translations` (well actually it was undetermined which one you’d see in the viewer..)
Now with fully tagged component awareness this whole constructs falls over completely: those components no longer match up at all and won’t be picked up by our transform processing since that one looks for components that are tagged as being part of `InstancePoses3D` :collision:

I’m not entirely sure how to design around this. My most preferred variant is to deprecate those fields and patch them up on load. Any utility method we have of want to add can use the instance pose tags.

Looking for opinions & ideas!

Wumpf:
Internal slack thread about the topic https://rerunio.slack.com/archives/C033K5VS2KD/p1746519651134839

Wumpf:
worth checking how many ros message types come with orientation like properties

PR #10010, opened 2025-05-19
Fix issues when combining `InstancePoses3D` with transform properties of `Boxes3D`/`Ellipsoids3D`/`Capsules3D`

Wumpf:
### Related

* part of #6889 
* fixes https://github.com/rerun-io/rerun/issues/9889 

### What

Previously, when logging both translations (or any shared component) on `InstancePoses3D` and on `Boxes3D` we wouldn't correctly resolve this.
With tag aware component aware queries, the problem got worse: we now have to query all types of pose transforms and combine them. This also implies that you have have several differently transformed _archetypes_ on the same entity.

In the future we may decide to remove `InstancePoses3D` and instead express these leaf transforms as an "extension" of all archetypes in the spatial view. To that end, I refrained from as initially planned to ad-hoc apply these transforms on visualizers, and instead integrated "archetype sensitive" transforms into `TransformCache` and `TransformTreeContext`. 

The result is unfortunately a little bit bulky but workable in its final form. It stands to reason that this may be a source of performance regressions, we have to check if this affects critical paths too much (added a note to the tagged component ticket to that end)
There's clearly lots of room for improvement here!


crates/viewer/re_view_spatial/src/transform_cache.rs
```diff
@@ -850,24 +924,84 @@ fn query_and_resolve_instance_poses_at_entity(
...
+        entity_db,
+        query,
+        archetypes::InstancePoses3D::name(),
+        &archetypes::InstancePoses3D::descriptor_translations(),
+    );
+
+    // Some archetypes support their own instance poses.
+    // TODO(andreas): can we quickly determine whether this is necessary for any given archetype?
+    let mut per_archetype = IntMap::default();
+    for (archetype_name, descriptor_translations) in
```
Wumpf, line 939: let's try to improve this a bit by doing a single large query and then operate on the query result instead!

crates/viewer/re_view_spatial/src/transform_cache.rs
```diff
@@ -841,6 +895,26 @@ fn query_and_resolve_tree_transform_at_entity(
     Some(transform)
 }
 
+/// Lists all archetypes except [`archetypes::InstancePoses3D`] that have their own instance poses.
+// TODO(andreas, jleibs): Model this out as a generic extension mechanism.
```
Wumpf, line 899: find or create ticket

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -473,20 +502,22 @@ fn transform_info_for_downward_propagation(
             reference_from_pinhole_entity: reference_from_entity,
         });
         reference_from_entity *= entity_from_2d_pinhole_content;
```
Wumpf, line 504: full disclosure: I have no idea why below re-transform of the instance poses was there when the actual collection happened earlier already. Looks 100% wrong to me now. I bet we just never ran into this because nobody reprojects things with instance poses

begs the question why there's no test. Should add one if it's not too much work :/

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -473,20 +502,22 @@ fn transform_info_for_downward_propagation(
...
-
-        // Need to update per instance transforms as well if there are poses!
-        if has_instance_transforms {
-            *reference_from_instances.first_mut() = reference_from_entity;
-        } else {
-            for reference_from_instance in &mut reference_from_instances {
-                *reference_from_instance *= entity_from_2d_pinhole_content;
-            }
-        }
```
Wumpf, line 484: full disclosure: I have no idea why this re-transform of the instance poses was there when the actual collection happened earlier already. Looks 100% wrong to me now. I bet we just never ran into this because nobody reprojects things with instance poses

begs the question why there's no test. Should add one if it's not too much work :/

crates/viewer/re_view_spatial/src/transform_cache.rs
```diff
@@ -850,24 +927,94 @@ fn query_and_resolve_instance_poses_at_entity(
...
+    );
+
+    // Some archetypes support their own instance poses.
+    // TODO(andreas): can we quickly determine whether this is necessary for any given archetype?
+    let mut instance_from_archetype_poses_per_archetype = IntMap::default();
+    for (archetype_name, descriptor_translations) in
+        archetypes_with_instance_pose_transforms_and_translation_descriptor()
+    {
+        if let Ok(mut instance_from_archetype_poses) =
+            SmallVec1::try_from_vec(query_and_resolve_instance_from_pose_for_achetype_name(
```
Wumpf, line 946: typo in method name

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -395,35 +449,25 @@ fn transform_info_for_upward_propagation(
...
-            (SmallVec1::new(reference_from_entity), false)
-        };
-
     // Apply tree transform.
     reference_from_entity *= transforms_at_entity
         .parent_from_entity_tree_transform
         .inverse();
-    if has_instance_transforms {
-        for reference_from_instance in &mut reference_from_instances {
-            *reference_from_instance = reference_from_entity * (*reference_from_instance);
```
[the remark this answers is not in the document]
Wumpf: Added a test: https://github.com/rerun-io/rerun/pull/10024
Again same before/after. It's fine 🤷

crates/viewer/re_view_spatial/src/transform_cache.rs
```diff
@@ -1705,16 +1874,161 @@ mod tests {
...
+            );
+
+            // More readable sanity check on translations which aren't affected by the rotation.
+            assert_eq!(
+                query_result[0].translation,
+                glam::Vec3A::new(11.0, 2.0, 3.0)
+            );
+            // Since rotation isn't 100% accurate, we need to check for equality with a small tolerance.
+            let eps = 0.000001;
+            // Rotation on the first box affects all insteances since it's splatted.
```
grtlr, line 2000:
```suggestion
            // Rotation on the first box affects all instances since it's splatted.
```
Why is our typo checker not finding these 🤔?

grtlr (approved):
Thank you again for the walkthrough, that was very helpful for understanding this PR!

PR #10015, opened 2025-05-20
tasks: wait_for_tasks can wait for streaming responses

andrea-reale:
### Related
* Required by: https://github.com/rerun-io/dataplatform/pull/762

### What

In https://github.com/rerun-io/dataplatform/pull/762 we made the tasks-completion streaming endpoitn a real streaming endpoint; before it was a stream with one item.

This PR adusts the `wait_for_tasks` endpoint so that it does no longer make the assumption there is only one item ever in the stream.

### What next
We could go further and improve the API adding a different `subscribe-tasks` method that yields results as tasks complete. This would let higher level utils like the dataset-util do a better job at a progress bar.  However, I did not adventure that far, as this was for me just a way to unblock 762.


rerun_py/src/catalog/connection_handle.rs
```diff
@@ -269,48 +269,62 @@ impl ConnectionHandle {
...
 
-                if status != "success" {
-                    return Err(PyValueError::new_err(format!(
-                        "task failed: {}",
-                        msgs.value(i)
-                    )));
+                // TODO(andrea): all this column unrwapping is a bit hideous. Maybe the idea of returning a dataframe rather
+                // than a nicely typed object should be revisited.
+                let schema = item.schema();
+                let col_indices = ["task_id", "exec_status", "msgs"]
```
abey79, line 287: pretty please extend `QueryTasksOnCompletionResponse` on the model of e.g. `CreatePartitionManifestsResponse`, with `const`s for field names and, ideally, a `schema()` method 🙏🏻

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -269,48 +269,62 @@ impl ConnectionHandle {
...
+                            .map_err(to_py_err)?,
+                    )
+                };
+                for i in 0..projected.num_rows() {
+                    if statuses.value(i) != "success" {
+                        let err = format!("task {}: {}", task_ids.value(i), msgs.value(i));
+                        errors.push(err);
+                    }
                 }
             }
```
abey79, line 317: Stretch goal: it would be very nice to have something like this:

```rust
impl QueryTasksOnCompletionResponse {
   pub fn iter(&self) -> impl Iterator<Item = (TaskId, TaskStatus, String)> { /* .... */ }
}
```

(with some `TypeConversionError` thrown in the mix)

abey79 (approved):
Lgtm. As always there is room for grpc message ergonomic improvement. Would be great to  at least do the minimum that's needed to document the opaque dataframe response.

rerun_py/src/catalog/connection_handle.rs
andrea-reale: Yesterday I have had a time-boxed attempt to move all the proper datamodel from `redap_tasks` in dataplatform to the `ext` thing, but the result was a bit awful, because the datamodel is relatively complex and the `ext` file became huge.

I will fight this battle another day, but I captured my notes in https://github.com/rerun-io/dataplatform/issues/811

crates/store/re_protos/src/v1alpha1/rerun.redap_tasks.v1alpha1.ext.rs
```diff
@@ -1,12 +1,56 @@
+use arrow::datatypes::{DataType, Field, Schema, TimeUnit};
+
 use crate::common::v1alpha1::DataframePart;
 use crate::v1alpha1::rerun_redap_tasks_v1alpha1::QueryTasksResponse;
 use crate::{TypeConversionError, missing_field};
 
+// TODO(dataplatform#811): improve converter methods
+
 impl QueryTasksResponse {
+    pub const TASK_ID: &str = "task_id";
```
zehiko, line 10: nit: I already saw repetition of some of these constants, would it make sense to move them at the module level (the ones that are shared, ofc). 
We now have task_id, kind, data in more than a few places...

zehiko (approved):
LGTM!

PR #10018, opened 2025-05-20
Mention URL scheme in `connect_grpc` docs

jprochazk:
We had a few places that mention connecting to a URL, but do not specify what that URL should look like.


abey79 (approved):
nice, thanks!

PR #10021, opened 2025-05-20
Use BuildJet runners with more CPUs and RAM

abey79:
### Related

* attempt at fixing these err 137 ci failure: https://github.com/rerun-io/rerun/actions/runs/15128223538/job/42525117004

### What

This PR bumps our BuildJet runners to 16vcpu/24GB, up from 8vcpu/12GB.

In the worst case, this should double our monthly costs, which has recently averaged at around ~35USD/month. Hopefully, faster workflow should mean less cost increase.

### DIFF
--- a/.github/workflows/reusable_build_and_upload_rerun_c.yml
+++ b/.github/workflows/reusable_build_and_upload_rerun_c.yml
@@ -87,7 +87,7 @@ jobs:
         run: |
           case "${{ inputs.PLATFORM }}" in
             linux-arm64)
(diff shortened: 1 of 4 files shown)


abey79:
@rerun-bot full-check

abey79:
> Started a full build: [rerun-io/rerun/actions/runs/15140638061](https://github.com/rerun-io/rerun/actions/runs/15140638061)

I don't think this uses any BuildJet runners, so we'll probably have to wait for the next nightly.

Wumpf (approved):
let's try 🤷

PR #10022, opened 2025-05-20
Do not silently drop arrow metadata when converting to/from column descriptors

abey79:
### Related

* slack thread: https://rerunio.slack.com/archives/C04DLS7CCE4/p1747756235011719
 
### What

We currently drop any arrow metadata that is not explicitly supported when converting from/to `SorbetBatch`. This PR attempts to rectify that.

TODO:
- [ ] do the same for `RowIdColumnDescriptor`
- [ ] do the same for `IndexColumnDescriptor`
- [ ] update table header UI accordingly


abey79:
outdated

PR #10023, opened 2025-05-20
No longer require graphics API validation layer for image comparision tests

Wumpf:
Tests recently started failing on ci with `InstanceFlags::VALIDATION requested, but unable to find layer: VK_LAYER_KHRONOS_validation`
Unclear why that happens. Did some image suddenly update? Attempting to work around this by disabling validation.


Wumpf:
@rerun-bot full-check

PR #10026, opened 2025-05-21
Require `ComponentDescriptor` in `ChunkBuilder::with_component_batch*` methods

grtlr:
### Related

* Part of #6889.

### What

Ultimately, we won't to get rid of `fn Component::descriptor()`, as it will always lead to untagged components, which most likely is a bug.

This PR narrows down the usage of that trait method so that it can be removed (or disallowed eventually). We have also decided that we will enforce fully-specified descriptors from our SDKs, which is why the `ChunkBuilder::with_component_batch*` methods now expect a corresponding descriptor too. In general `ChunkBuilder::with_archetype` should be preferred though.


Wumpf (approved):
oof, that's a lot. Gave it a rough browse but no thorough review. Thanks for biting this bullet!

PR #10029, opened 2025-05-21
Remove `data_based_visualizability_filter` from `BarChart` and `EncodedImage`

grtlr:
### Related

* Part of #6889.

### What

Part of long tail of tagged component changes and effort to remove `Component::descriptor`.


Wumpf (approved):
thanks!

PR #10032, opened 2025-05-21
moving partition type and storage url to partition manifest

zehiko:
As part of append workflow changes, we're moving type and storage url to partition manifest. This is the necessary rerun side update.

### DIFF
--- a/crates/store/re_protos/src/lib.rs
+++ b/crates/store/re_protos/src/lib.rs
@@ -83,12 +83,13 @@ pub mod manifest_registry {
         /// `DatasetManifest` mandatory field names. All mandatory metadata fields are prefixed
         /// with "rerun_" to avoid conflicts with user-defined fields.
         pub const DATASET_MANIFEST_ID_FIELD_NAME: &str = "rerun_partition_id";
+        pub const DATASET_MANIFEST_CREATED_AT_FIELD_NAME: &str = "rerun_created_at";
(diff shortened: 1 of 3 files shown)


crates/store/re_protos/src/lib.rs
```diff
@@ -83,12 +83,13 @@ pub mod manifest_registry {
...
         pub const DATASET_MANIFEST_PARTITION_MANIFEST_UPDATED_AT_FIELD_NAME: &str = "rerun_partition_manifest_updated_at";
         pub const DATASET_MANIFEST_PARTITION_MANIFEST_URL_FIELD_NAME: &str = "rerun_partition_manifest_url";
-        pub const DATASET_MANIFEST_RECORDING_TYPE_FIELD_NAME: &str = "rerun_partition_type";
-        pub const DATASET_MANIFEST_REGISTRATION_TIME_FIELD_NAME: &str = "rerun_registration_time";
-        pub const DATASET_MANIFEST_START_TIME_FIELD_NAME: &str = "rerun_start_time";
-        pub const DATASET_MANIFEST_STORAGE_URL_FIELD_NAME: &str = "rerun_storage_url";
+
+        /// Partition manifest mandatory field names.
+        pub const PARTITION_MANIFEST_PARTITION_TYPE_FIELD_NAME: &str = "rerun_partition_type";
+        pub const PARTITION_MANIFEST_STORAGE_URL_FIELD_NAME: &str = "rerun_storage_url";
```
abey79, line 92: What's the purpose of these constant? Are they returned by other endpoints? If so, would it be possibly to add the `schema()` method to the corresponding responses?

abey79 (approved):

Issue #9906, opened 2025-05-07
`AnyValues` should use argument names to set `archetype_field_name` metadata


jleibs:
I believe this will make AnyValues conform to the implicit pattern of other ArchetypeConstructors where the parameter-names of the constructors define fields, rather than component-names.

PR #10038, opened 2025-05-21
Remove untagged `Component::descriptor()`

grtlr:
### Related

* Part of #6889.

### What

Another big step towards having full tagging support in the viewer. By removing `Component::descriptor()` we make it harder to accidentally use untagged descriptors and therefore reduce bugs in our code, and in our users code.

This refactor goes right up to the point where we have to change our SDKs, so there is more design work to be made (via new `AnyValues`).

To achieve this we still rely on untagged components when implementing `ComponentBatch::descriptor` (left todos there).

Next steps are:
* Remove remaining untagged methods (track usage of `Component::name()`)
* Designing new `AnyValues` #9906 for all SDKs.
* Making `ComponentDescriptor::new` more explicit when using untagged components or prevent untagged components all together by removing the optionals.
* Optionally getting rid of `ComponentBatch` (Maybe we wait for the removal of indicator components).

## Dependency

* Requires: https://github.com/rerun-io/dataplatform/pull/815

### DIFF
--- a/crates/build/re_types_builder/src/codegen/rust/api.rs
+++ b/crates/build/re_types_builder/src/codegen/rust/api.rs
@@ -1016,8 +1016,8 @@ fn quote_trait_impls_for_datatype_or_component(
         quote! {
             impl ::re_types_core::Component for #name {
                 #[inline]
-                fn descriptor() -> ComponentDescriptor {
-                    ComponentDescriptor::new(#fqname)
+                fn name() -> ComponentName {
+                    #fqname.into()
                 }
             }
         }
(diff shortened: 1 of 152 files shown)


Wumpf (approved):
looks good!
interestingly this now for the first time makes "untagged" a concept in the codebase, but I figure that's how we used the term by now anyways

Wumpf:
I'm nervous about "untagged" because of the rerun-historical connotation: There was a world where we had only component name so this was truly untagged. But since 0.22 everything always has a tag, just that that tag is sometimes component-name-only. Calling component-name-only "untagged" rubs me a little bit the wrong way because compared to 0.21 it's still technically tagged. But a) that's ancient past now and b) I don't have any better idea what to call this either and it's good to give it a name!

grtlr:
We have decided on calling these type of descriptors _partial_ instead of _untagged_.

PR #10041, opened 2025-05-21
Upload NPM package as tarball to GCS

jprochazk:
We now use `yarn pack` to produce a tarball of each package, and upload them to GCS. They'll be available at `https://build.rerun.io/{spec}/rerun_js/{package-name}.tar.gz`, where `spec` is the version/commit, and `package-name` is what's listed in `package.json`.

For example, the `web-viewer` package built for commit c3a864f is hosted publicly at `https://build.rerun.io/commit/c3a864f/rerun_js/web-viewer.tar.gz`. It is possible to install this package via NPM directly:

```
npm install https://build.rerun.io/commit/c3a864f/rerun_js/web-viewer.tar.gz
```

We build these on PRs (using the same filter as web viewer build), on `main`, and for `nightly`.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -83,6 +83,22 @@ jobs:
       NIGHTLY: true
     secrets: inherit
 
+  build-js:
+    name: "Build JS"
+    uses: ./.github/workflows/reusable_build_js.yml
+    with:
+      CONCURRENCY: nightly
+    secrets: inherit
+
+  upload-js:
+    name: "Upload JS"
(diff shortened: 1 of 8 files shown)


abey79 (approved):
Well done! Do I understand correctly that this does not push the tarballs to the GH releases (dev and actual releases)? Would this be easy to add?

jprochazk:
Yeah, it doesn't, I forgot about that. I can add it

jprochazk:
(test failure unrelated, merging)

PR #10043, opened 2025-05-21
Expose LineBatchBuilder

Vrixyz:
`LineBatchBuilder` was not public, resulting in missing documentation and inability to use it in user code explicitly.


Wumpf (approved):
👍

PR #10044, opened 2025-05-22
Remove `component_mono_with_log_level_by_name()` from `LatestAtResults`

grtlr:
### Related

* Part of #6889.

### What

This is one of the two remaining `*_by_name()` methods in `LatestAtResults` of `re_query`. The other one is a bit more involved, so I'm splitting things up.


Wumpf (approved):
thanks for cleaning up!

PR #10045, opened 2025-05-22
All methods of `re_query::LatestAtResults` use `ComponentDescriptor`

grtlr:
### Related

* Part of #6889.

### What

This finally should make all queries fully tag aware.


Wumpf (approved):
wohooo! \o/

PR #10050, opened 2025-05-22
Split off `GlobalContext` to its own crate

abey79:
### What

This is the rabbit hole resulting from the realisation that `rerun-cli` (without any viewer feature) would depend on datafusion (which is a _large_ dependency sub-tree). The root cause was that some CLI stuff relied on the `SystemCommand` stuff, which pulled `re_viewer_context`, which pulled way to many things.

This PR splits of `GlobalContext`, which was already a sub-part of `ViewerContext`. So it's the continuation of this previous work:
- https://github.com/rerun-io/rerun/pull/9039

The intended goal is achieved:

```
❯ cargo tree -i datafusion -p rerun-cli --no-default-features
warning: nothing to print.
```

(Note that the default features include the web and native viewer, so it obviously still depend on datafusion.)

Implementation note:
- I had to move the component UI and view class registries from `GlobalContext` back to `ViewerContext`. It's a bit sad, but does make sense (these are _definitely_ viewer things).
- I opted to blanket re-export everything from `re_global_context` into `re_viewer_context`'s namespace, to minimise the diff. (But I'm happy deal with the consequence of not doing that if we prefer.)
- The orphan rule got in the way, so `ItemCollection` is (very slightly) more verbose to build in a few places.


crates/viewer/re_viewer_context/src/selection_state.rs
```diff
@@ -115,13 +117,23 @@ impl From<Item> for ItemCollection {
...
 }
 
-impl<T> From<T> for ItemCollection
-where
-    T: Iterator<Item = (Item, Option<ItemContext>)>,
-{
-    #[inline]
-    fn from(value: T) -> Self {
-        Self(value.collect())
+// impl<T> From<T> for ItemCollection
```
jleibs, line 120: Any particular reason this is hanging around?

jleibs (approved):

crates/viewer/re_viewer_context/src/selection_state.rs
abey79: Remnants of a short—and lost—battle against the orphan rule.

PR #10051, opened 2025-05-22
Fix rerun js package upload paths

jprochazk:
It's uploading `rerun_js` instead of the artifact (`rerun_js_package`)


jprochazk:
yolo merging as this was just a typo

PR #10054, opened 2025-05-23
Fix column names and hide indicator components in the partition table

abey79:
Title.

Part of:
- https://github.com/rerun-io/dataplatform/issues/750

Default view now looks like this:

image

### DIFF
--- a/crates/viewer/re_redap_browser/src/servers.rs
+++ b/crates/viewer/re_redap_browser/src/servers.rs
@@ -144,7 +144,7 @@ impl Server {
                 entity_path.starts_with(&std::iter::once(EntityPathPart::properties()).collect())
             }) {
                 // Property column, just hide indicator components
-                //TODO(ab): remove this when we no longer have indicator components
+                //TODO(#8129): remove this when we no longer have indicator components
                 !desc
                     .column_name(BatchType::Dataframe)
                     .ends_with("Indicator")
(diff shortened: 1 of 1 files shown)


crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -130,16 +131,23 @@ impl Server {
...
-                .unwrap_or(name.as_ref())
-                .replace('_', " ");
+            let mut name = default_display_name_for_column(desc);
+
+            if name.starts_with("rerun_") {
+                name = name
+                    .strip_prefix("rerun_")
+                    .unwrap_or(name.as_ref())
+                    .replace('_', " ");
+            }
```
Wumpf, line 141: nit:
```suggestion
            let name = default_display_name_for_column(desc);
            let name = name.strip_prefix("rerun_").map_or(name, |name| name.replace('_', " "));
```
(or whatever exactly compiles)

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -130,16 +131,23 @@ impl Server {
...
 
             let default_visible = if desc.entity_path().is_some_and(|entity_path| {
                 entity_path.starts_with(&std::iter::once(EntityPathPart::properties()).collect())
             }) {
-                true
+                // Property column, just hide indicator components
+                //TODO(ab): remove this when we no longer have indicator components
```
Wumpf, line 147:
```suggestion
                //TODO(#8129): remove this when we no longer have indicator components
```

makes it easier to find

Wumpf (approved):

Issue #2960, opened 2023-08-11
Future strategy for icon management


abey79:
(forked from discussion in #2920)

We currently use 24x24px png exported from Figma.

This approach has two shortcomings:
- it doesn't scale well for high-dpi screen.
- it's inadequate for (at least) collapsing triangles which must be animated in rotation (see #2738, #2920)

Moving forward, our strategy for icons should enable:
- Ease of changing out icons (e.g just drop a .png or .svg in folder)
- Low code bloat (don't increase .wasm size too much)
- Visual quality (icons shouldn't be blurry, no matter what dpi scaling the user has)

cc @martenbjork @emilk

abey79:
[Comment](https://github.com/rerun-io/rerun/pull/2920#issuecomment-1674388236) from @emilk:

> We should create a proper issue about how to do icons. Some quick, random thoughts:
> 
> There are at least three things to consider:
> 
> * Ease of changing out icons (e.g just drop a .png or .svg in folder)
> * Low code bloat (don't increase .wasm size too much)
> * Visual quality (icons shouldn't be blurry, no matter what dpi scaling the user has)
> 
> There are two major ways to go.
> 
> ### Imagers (e.g. .png)
> * very easy to create and use
> * but one imager will never be optimal for all resolution scalings.
>   
>   * we could export one high-res .png and dynamically rescale it to fit the users resolution. Pretty easy to implement, and low code bloat.
> 
> ### Vector graphics
> Using `epaint` vector rendering (like this PR) works great (proper anti-aliasing for all resolutions), but only for convex polygons.
> 
> Dor concave polygons we either need to implement something new in epaint, or use some 3rd party rendering tool (more code bloat)
> 
> * `egui_extras` uses a lot of dependencies for this, see https://docs.rs/egui_extras/latest/egui_extras/image/struct.RetainedImage.html#method.from_svg_bytes
> 
> `.svg` is a horribly complex format. tinyvg is an interesting new alternative, but still requires us to ship code for parsing and rasterizing, increasing .wasm size.
> 
> ### Conclusions?
> I'm leaning towards using a single high-res .png for each icon, then resize it using some high-quality algo (e.g. Lanczos filtering) to create a crisp icon for the current resolution scaling.
> 
> Easy to use, easy to implement, high-quality, low code bloat.

abey79:
My two cents:

[...]

PR #10055, opened 2025-05-23
Switch icons to .svg

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/2960

### What
Replaces all .png icons for .svg

The apparent differences are pretty minor, but in some situations the results should be slightly less blurry.

### TODO
* [x] Fix the icons in Figma so that they are 14x14 when exported, instead of 14x15 (@gavrelina 🙏 )
* [x] Figure out why they appear more blurry than the .png:s 😬  (maybe it's the above thing?)
* Add missing SVG:s not found in the "Latest icons" section on Figma (@gavrelina 🙏)
	* [x] `bottom_panel_toggle`
	* [x] `breadcrumbs_separator`
	* [x] `data_source`
	* [x] `discord`
	* [x] `dnd_handle`
	* [x] `left_panel_toggle`
	* [x] `notification`
	* [x] `rerun_io`
	* [x] `rerun_menu`
	* [x] `right_panel_toggle`
	* [x] `settings`
	* [x] `video_error`


Wumpf:
difference is huge if you set your viewer to huge! Enabling changelog on this!

image

image


One interesting detail here: was the rerun menu icon supposed to be bigger?

Wumpf (changes requested):
yeah actually that menu regressed quite a bit, misses the arrow as well! Should fix that before landing. otherwise great!

before:
image

after:
image

emilk:
Fixed (by reverting the icon):
image

Wumpf (approved):

Issue #9340, opened 2025-03-21
Add `AnyValues` wrapper to Rust and C++ APIs


grtlr:
We want to make it simple to log arbitrary user data. Currently, adding `AnyValues` in Rust and in C++ is very annoying. Now that we have eager-serialization this should be a lot simpler. 

See also our docs: https://rerun.io/docs/concepts/entity-component#adding-custom-data

Supporting this use-case will likely become more important once we have recording properties with arbitrary data (#9199). Here is a comparison, with Python being the clear winner:

### Rust 🦀 
```rs
let confidences = rerun::SerializedComponentBatch::new(
    Arc::new(arrow::array::Float64Array::from(vec![0.3, 0.4, 0.5, 0.6])),
    rerun::ComponentDescriptor::new("confidences"),
);

let traffic = rerun::SerializedComponentBatch::new(
    Arc::new(arrow::StringArray::from(vec!["low"])),
    rerun::ComponentDescriptor::new("traffic"),
);

let weather = rerun::SerializedComponentBatch::new(
    Arc::new(arrow::StringArray::from(vec!["sunny"])),
    rerun::ComponentDescriptor::new("weather"),
);

rec.send_property(
    "situation",
    &[&confidence as &dyn rerun::AsComponents, &traffic, &weather],
)?;

```

### Python 🐍 

```py
rec.send_property(
    "situation",
    rr.AnyValues(
        confidences=[0.3, 0.4, 0.5, 0.6],
        traffic="low",
        weather="sunny",
    ),
)

```

### C++ 🌊 

```cpp
std::shared_ptr<arrow::Array> arrow_array;

arrow::DoubleBuilder confidences_builder;
ARROW_RETURN_NOT_OK(confidences_builder.AppendValues({1.2, 3.4, 5.6}));
ARROW_RETURN_NOT_OK(confidences_builder.Finish(&arrow_array));
auto confidences =
    rerun::ComponentBatch::from_arrow_array(std::move(arrow_array), "confidence");

arrow::StringBuilder traffic_builder;
ARROW_RETURN_NOT_OK(traffic_builder.Append("low"));
ARROW_RETURN_NOT_OK(traffic_builder.Finish(&arrow_array));
auto traffic =
    rerun::ComponentBatch::from_arrow_array(std::move(arrow_array), "traffic");

arrow::StringBuilder weather_builder;
ARROW_RETURN_NOT_OK(weather_builder.Append("sunny"));
[...]

PR #10074, opened 2025-05-26
Provide `AnyValues` helpers in Rust SDK

### LINKED ISSUE #9906: `AnyValues` should use argument names to set `archetype_field_name` metadata

grtlr:
### Related

* Part of #6889.
* Closes #9906.
* Part of #9340.

### What

This PR does multiple things:

* Works towards providing a common `AnyValues` interface that is similar across Python 🐍 , C++ 🌊 , and Rust 🦀 SDKs (so far we only have `AnyValues` in Python).
* By using the new `AnyValues` we ensure that more of the logged data is actually tagged.

### DIFF
--- a/crates/store/re_types/src/any_values.rs
+++ b/crates/store/re_types/src/any_values.rs
@@ -1,0 +1,108 @@
+//! Utilities to log arbitrary data to Rerun.
+
+use nohash_hasher::IntMap;
+
+use crate::{
+    ArchetypeFieldName, ArchetypeName, Component, ComponentDescriptor, SerializedComponentBatch,
+};
+use re_types_core::{try_serialize_field, AsComponents, ComponentName, Loggable};
+
+/// A helper for logging arbitrary data to Rerun.
+#[derive(Default)]
+pub struct AnyValues {
+    archetype_name: Option<ArchetypeName>,
+    batches: IntMap<ArchetypeFieldName, SerializedComponentBatch>,
+}
+
+impl AnyValues {
+    /// Assigns an (archetype) name to this set of any values.
+    #[inline]
+    pub fn new(archetype_name: impl Into<ArchetypeName>) -> Self {
+        Self {
+            archetype_name: Some(archetype_name.into()),
+            batches: Default::default(),
+        }
+    }
+
+    /// Adds a field of arbitrary data to this archetype.
+    ///
+    /// In many cases, it might be more convenient to use [`Self::with_component`] to log an existing Rerun component instead.
+    #[inline]
+    pub fn with_field(
+        mut self,
+        archetype_field_name: impl Into<ArchetypeFieldName>,
+        array: arrow::array::ArrayRef,
+    ) -> Self {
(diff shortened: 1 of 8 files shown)


docs/snippets/all/tutorials/any_values.rs
```diff
@@ -7,35 +7,25 @@ use rerun::external::arrow;
...
+            "homepage",
+            Arc::new(arrow::array::StringArray::from(vec![
+                "https://www.rerun.io",
+            ])),
+        )
+        .with_field(
+            "repository",
+            Arc::new(arrow::array::StringArray::from(vec![
+                "https://github.com/rerun-io/rerun",
+            ])),
```
emilk, line 25: If we use `impl Into<ArrayRef>` as the argument type, maybe we can get rid of the `Arc::new` here?

grtlr: I tried this, but unfortunately could not get the compiler to comply.

docs/snippets/all/concepts/recording_properties.rs
```diff
@@ -19,26 +19,22 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            "weather",
+            Arc::new(arrow::array::StringArray::from(vec!["sunny"])),
+        );
 
     // Adds another property, this time with user-defined data.
-    rec.send_property(
-        "situation",
-        &[&confidences as &dyn rerun::AsComponents, &traffic, &weather],
-    )?;
+    rec.send_property("situation", &other)?;
```
emilk, line 37: This became _a lot_ nice ⭐

crates/store/re_types/src/any_values.rs
```diff
@@ -0,0 +1,108 @@
...
+#[derive(Default)]
+pub struct AnyValues {
+    archetype_name: Option<ArchetypeName>,
+    batches: IntMap<ArchetypeFieldName, SerializedComponentBatch>,
+}
+
+impl AnyValues {
+    /// Assigns an (archetype) name to this set of any values.
+    #[inline]
+    pub fn new(archetype_name: impl Into<ArchetypeName>) -> Self {
```
emilk, line 20: I think most users will first grab for `AnyValues::new()`, and maybe not even realize there is a `AnyValues::default()`, and so they will wonder what they should pick for archetype name.

So I suggest we do builder pattern all the way instead:

```rs
AnyValues::new().with_field(…)
// vs
AnyValues::new().with_archtype_name(…).with_field(…)
```

or alternatively use a more verbose ctor name:


```rs
AnyValues::new().with_field(…)
// vs
AnyValues::from_archtype_name(…).with_field(…)
```

grtlr: I deliberately made that choice to nudge users to provide an archetype name, which should be good practice going forward as it helps with uniquely identifying the data going forward.

PR #10075, opened 2025-05-26
Fix incorrect 'structfields' doc links

Vrixyz:
rust analyzer wasn't interpreting those. to be noted those types are not part of public API (thus not appearing in rustdoc).


Wumpf (approved):
thx! must have been a copy past error from wgpu

PR #9820, opened 2025-04-28
Allow for custom visualizer with custom shader that integrates into existing view

Wumpf:
### Related

* TODO: there gotta be some related ticket!

### What

Previously, it was already possible to add fully custom views to the Viewer. However, it was pretty much impossible to add a visualizer to an existing View, e.g. to add a new kind of visualization to the 3D view. This PR enables exactly that!

Most of this PR is an example demonstrating such a custom visualizer extending the 3D view with a new visualization. However, there's plenty of fixes and  function/struct/trait exports that had to happen to make this possible!

DRAFT TODO:
* [ ] split out general improvements that also apply to custom views into a separate PR?
* [ ] visualize something interesting
* [ ] make custom visualizer pickable
* [ ] give custom visualizer outlines
* [ ] add to viewer customization doc page (can be short for starters)

PR #10079, opened 2025-05-26
Expose `DrawPhase`

ThierryBerger:
If one wants to make a custom renderer, `DrawPhase` not being exposed is getting in the way, as it's a public interface of the `Renderer` trait, especially for [`participated_phases()`](https://docs.rs/re_renderer/latest/re_renderer/renderer/trait.Renderer.html#tymethod.participated_phases)


Wumpf (approved):
makes sense. Great to have someone actually try this stuff!

I figure I should split out the stuff I exposed in https://github.com/rerun-io/rerun/pull/9820 as well as a separate pr

Wumpf:
there we go! @Vrixyz you might find this helpful https://github.com/rerun-io/rerun/pull/10081

Issue #10067, opened 2025-05-23
Command-clicking on partition link should not immediately switch to that recording


abey79:
In the partition table, the "Open" link detects cmd/ctrl and use `OpenUrl::new_tab()` when it's pressed. However, something is broken with how we handle these events, because the viewer switches to the recording regardless.

- Adjacent to: https://github.com/rerun-io/rerun/issues/10066

Issue #10066, opened 2025-05-23
Improve the UX of open remote recordings


abey79:
Currently, the UX of loading remote recordings is pretty bad:

1) click the `Open` button
2) the data source appears but we're presented with the welcome screen
3) the recording finally appears and is displayed

We need to improve on that, by either:
- displaying something more reasonable than the welcome screen when we have a datasource but no recording
- immediately spewing an empty recording so we can switch to it

At first sight, the latter is the preferred approach.

- Part of: https://github.com/rerun-io/dataplatform/issues/823
- Adjacent to: https://github.com/rerun-io/rerun/issues/10067

PR #10080, opened 2025-05-26
Improve UX of opening recordings in catalog browser

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/10067
* Closes https://github.com/rerun-io/rerun/issues/10066

#### Fix "open in background" when clicking redap recording links
This adds a `bool` flag to the `re_smart_channel` data srouce to indicate wether or not the user intended to switch to this recording once it has loaded.

Putting it in the smart channel was not the easiest decision, but I think it was the least bad.

For one, the smart channel source already keeps track of things like the URI fragment which already contains instructions on what to do next, once the URI is loaded (e.g. go to a specific timeline).

But also, I found no better place for this data.

#### Show "Switch to" button if recording is already open
cmd-clicking a few links:
open-in-background

### DIFF
--- a/crates/viewer/re_component_ui/src/variant_uis/redap_uri_button.rs
+++ b/crates/viewer/re_component_ui/src/variant_uis/redap_uri_button.rs
@@ -52,10 +52,6 @@ pub fn redap_uri_button(
         if response.clicked() {
             // Show it:
             ctx.command_sender()
-                .send_system(SystemCommand::ChangeDisplayMode(
-                    re_viewer_context::DisplayMode::RedapServer(uri.origin().clone()),
-                ));
-            ctx.command_sender()
                 .send_system(SystemCommand::SetSelection(
                     re_viewer_context::Item::StoreId(loaded_recording_id),
                 ));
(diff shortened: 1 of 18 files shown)


crates/viewer/re_component_ui/src/variant_uis/redap_uri_button.rs
```diff
@@ -32,21 +32,46 @@ pub fn redap_uri_button(
...
+        } else if response.clicked() {
+            let url = if ui.input(|i| i.modifiers.any()) {
+                egui::OpenUrl::new_tab(uri)
+            } else {
+                egui::OpenUrl::same_tab(uri)
+            };
 
-        ui.ctx().open_url(url);
+            ui.ctx().open_url(url);
+        }
```
Wumpf, line 74: there gotta be a utility for this in egui, right..?

Wumpf (approved):
> For one, the smart channel source already keeps track of things like the URI fragment which already contains instructions on what to do next, once the URI is loaded (e.g. go to a specific timeline).

annoying yeah. But agreed, it's not that much different from the url telling what to do 🤷 


The result is great, so much better <3 and the rest of the impl is fairly clear as well!

crates/viewer/re_component_ui/src/variant_uis/redap_uri_button.rs
emilk: it will soon: https://github.com/emilk/egui/pull/7093 😆

Issue #10065, opened 2025-05-23
Tagged components milestone 4: descriptor-driven dataframe queries and views


grtlr:
Right now neither the dataframe queries nor the dataframe view are descriptor-aware, which needs to change in the future to give our users more flexibility.

In cases of ambiguity, we should probably use (`archetype_name`, `archetype_field_name`), similar to how we present components in the UI (rerun-io/rerun#9983) and how we are about to change `AnyValues` (#9906).

Issue #8129, opened 2024-11-13
Tagged components milestone 2: no more indicators


teh-cmc:
The second milestone for tagged components is to remove the concept of indicator components **entirely**.
The semantics of the viewer must not be impacted in any way by that change.

* Remove all codegen related to indicator components.
* Replace all logic that relies on querying for indicators with new descriptor-based logic ("is there anything on that path that belongs to the `Points3D` archetype?").

---

* Part of https://github.com/rerun-io/rerun/issues/6889
* Requires #7948

Issue #7699, opened 2024-10-11
Make it possible to look up ComponentName without interning the string


jleibs:
In theory if I have a string I should be able to hash it and find the corresponding value in a `BTreeSet<ComponentName>`, however in practice to find the value in the set I have to create a ComponentName which does the interning.

kpreid:
At a glance, I think that to make this possible, what you need is:

* `pub struct UninternedString(str);`
* `impl Hash for UninternedString { /* mimic impl Hash for InternedString */ }`
* `impl Borrow<UninternedString> for InternedString`
* `impl AsRef<UninternedString> for &str` (must not be `Borrow` since it hashes differently)

PR #10082, opened 2025-05-27
Make `archetype_field_name` mandatory in `ComponentDescriptor`

### LINKED ISSUE #9906: `AnyValues` should use argument names to set `archetype_field_name` metadata

### LINKED ISSUE #10152: Write `schema` test cases for `re_chunk_store`
With #6889, the types of data that lands in the chunk store can be much more complex. Also, with #10082 we cannot rely on the component name by itself anymore for retrieving information about a column.

An example for this is are two different `arrow::DataType` that are logged to the same `archetype_field_name` in different entities. #10082 should handle this already, but we need more thorough test cases.

Similarly, we should write test cases for the `archetype_field_name` of an entity changing its datatype.

grtlr:
### Related

* Part of #6889.
* Part of #9978.
* Part of #10065.
* Will make #8129 a lot easier.
* Closes #9906.
* Closes https://github.com/rerun-io/dataplatform/issues/841.
* Closes #10152.

### What

This PR makes big changes to `ComponentDescriptor`:

* `archetype_field_name` is now mandatory.
* `component_name` is now optional.

It also sprinkles some ✨`TODO(#6889)`✨ comments in places that still need some tagging love 🫶.

### Big (breaking) changes

Split up by crate to make things easier to review. We use the following symbols:

* ♻️ means we require migration code.
* 🪶 means we need to point this out in the migration guide.

#### `re_types_core`

* 🪶 The obvious `ComponentDescriptor` changes.

#### `re_types`

* Indicator components now become indicator fields. This is only temporary until we get rid of indicators.

#### `re_log_types`

* 🪶 Removed auto-filling of `rerun.components.` to `ComponentName` in path parsing.
* ♻️ Swapped the `ComponentName` and `ArchetypeFieldName` location in path parsing.

#### `re_sorbet`

* ♻️ Changed `ColumnDescriptor` to reflect new `ComponentDescriptor`.
* 🪶 Lookup `ColumnDescriptor` by `ArchetypeFieldName` instead of `ComponentName`.
* Changed `ComponentColumnSelector`.
* 🪶 Changed `ComponentColumnDescriptor::column_name` to use fully-qualified column names.

#### `re_chunk_store`

* 🪶 Changed `per_metadata_column` to be descriptor-aware. Searching by component name is less performant now because it uses a scan over this index.
* 🪶 Introduced `ColumnIdentifier` as a way to specify columns in `ViewContents`

#### `re_viewer_context`

* Emit warnings when trying to show a component UI for a descriptor without component name.

#### `rerun_py`

* 🪶 Dataframe queries now use the new string representation everywhere. It's not allowed to pass in components anymore.
* 🪶 `resolve_component_column_selector` now returns an `Option`.

### Important areas for review

* [x] Testcases in `parse_path.rs`
* [x] Removal of `same_descriptor` check from `re_chunk`
* [x] Migration in `re_sorbet`
* [x] Changes to `per_column_metadata`, `lookup_datatpe`, and `resolve_component_selector`.
* [x] Changes to `query.rs` in `re_dataframe`.
* [x] Changes to `dataframe_query.rs` + `dataframe.rs` in `rerun_py`.
* [x] URDF loader changes in `re_data_loader` @emilk.
* [x] Changes to `re_spatial_view` (`filter_map`) @Wumpf.
* [x] Python dataframe test cases @jleibs.

### Outstanding

* #10065
* #10211 
* #10212

### DIFF
--- a/crates/viewer/re_time_panel/src/time_panel.rs
+++ b/crates/viewer/re_time_panel/src/time_panel.rs
@@ -16,10 +16,7 @@ use re_log_types::{
 };
 use re_types::blueprint::components::PanelState;
 use re_types_core::ComponentDescriptor;
-use re_ui::{
-    ContextExt as _, Help, SyntaxHighlighting as _, UiExt as _, filter_widget, icon_text, icons,
-    list_item,
-};
+use re_ui::{ContextExt as _, Help, UiExt as _, filter_widget, icon_text, icons, list_item};
 use re_ui::{IconText, filter_widget::format_matching_text};
 use re_viewer_context::{
     CollapseScope, HoverHighlight, Item, ItemCollection, ItemContext, RecordingConfig, TimeControl,
@@ -872,15 +869,13 @@ impl TimePanel {
                     )
                     .show_hierarchical(
                         ui,
-                        list_item::LabelContent::new(
-                            component_descr.syntax_highlighted(ui.style()),
-                        )
-                        .with_icon(if is_static {
-                            &re_ui::icons::COMPONENT_STATIC
-                        } else {
-                            &re_ui::icons::COMPONENT_TEMPORAL
-                        })
-                        .truncate(false),
+                        list_item::LabelContent::new(component_descr.archetype_field_name.as_str())
+                            .with_icon(if is_static {
+                                &re_ui::icons::COMPONENT_STATIC
+                            } else {
+                                &re_ui::icons::COMPONENT_TEMPORAL
+                            })
+                            .truncate(false),
                     );
 
                 self.handle_interactions_for_item(
--- a/crates/viewer/re_view_dataframe/src/dataframe_ui.rs
+++ b/crates/viewer/re_view_dataframe/src/dataframe_ui.rs
@@ -318,9 +318,15 @@ impl egui_table::TableDelegate for DataframeTableDelegate<'_> {
 
             header_ui(ui, connected_to_next_cell, |ui| {
                 let header_content = |ui: &mut egui::Ui| {
-                    let text = egui::RichText::new(column.display_name())
-                        .strong()
-                        .monospace();
+                    let text = egui::RichText::new(
+                        if let ColumnDescriptor::Component(component) = column {
+                            component.archetype_field_name.to_string()
+                        } else {
+                            column.display_name()
+                        },
+                    )
+                    .strong()
+                    .monospace();
                     let archetype = column.archetype_name().map_or("", |a| a.short_name());
 
                     let is_selected = match column {


grtlr:
@rerun-bot full-check

crates/store/re_chunk/src/chunk.rs
```diff
@@ -1227,7 +1213,7 @@ impl Chunk {
     pub fn component_names(&self) -> impl Iterator<Item = ComponentName> + '_ {
         self.components
             .keys()
-            .map(|desc| desc.component_name)
+            .filter_map(|desc| desc.component_name)
```
jleibs: This edit makes me wonder if there should also be an `archetype_field_names()` helper -- do we not need it? What does `component_names()` get used for at this point?

crates/store/re_chunk/src/merge.rs
```diff
@@ -254,10 +238,7 @@ impl Chunk {
     /// * Use the same datatypes for the components they have in common.
     #[inline]
     pub fn concatenable(&self, rhs: &Self) -> bool {
-        self.same_entity_paths(rhs)
-            && self.same_timelines(rhs)
-            && self.same_datatypes(rhs)
-            && self.same_descriptors(rhs)
+        self.same_entity_paths(rhs) && self.same_timelines(rhs) && self.same_datatypes(rhs)
```
jleibs, line 241: This looks suspcious to me. Some form of `same_descriptors()` still important for being able to concatenate or else we risk losing metadata.

crates/store/re_chunk/src/transport.rs
jleibs, line 1: This change again looks somewhat problematic.

The usage of: `components.iter()` on line 84 means that the resultant schema will only contain schema entries for columns that contain components, which is no longer guaranteed.

crates/store/re_chunk_store/src/dataframe.rs
```diff
@@ -106,7 +107,7 @@ pub type IndexRange = ResolvedTimeRange;
 /// ## Terminology: view vs. selection vs. filtering vs. sampling
 ///
 /// * The view contents specify which subset of the database (i.e., which columns) the query runs on,
-///   expressed as a set of [`EntityPath`]s and their associated [`ComponentName`]s.
+///   expressed as a set of [`EntityPath`]s and their associated [`re_types_core::ComponentName`]s.
```
jleibs: ComponentName should be archetype field (unless we're going to eventually rename archetype field back to component).

crates/store/re_chunk_store/src/query.rs
```diff
@@ -272,7 +272,60 @@ impl ChunkStore {
         }
     }
 
+    /// Retrieves the [`ComponentDescriptor`] at a given [`EntityPath`] that has a certain [`ArchetypeName`]
```
jleibs, line 275: I know you wanted to save the other renames for the future, but it's super super hard to keep ComponentDescriptor vs ColumnDescriptor straight in my head at this point.

crates/store/re_chunk_store/src/store.rs
```diff
@@ -391,6 +391,41 @@ impl ChunkStoreHandle {
     }
 }
 
+/// A column in a given [`EntityPath`] is uniquely identified by its [`ArchetypeName`] (which can be optional),
+/// and its [`ArchetypeFieldName`].
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct ColumnIdentifier {
```
jleibs, line 397: This seems like an anti-concept to me. In my mind a "column" exists in the chunk store and also includes the entity-path. Do we need an entity-free column?

jleibs (commented):
Need to continue reviewing later, but wanted to save some initial comments.

crates/store/re_chunk/src/chunk.rs
grtlr: Good catch!

I have resolved this—we already have `.component_descriptors()`, which I think should be enough.

crates/store/re_chunk/src/merge.rs
grtlr: @wumpf and I looked at this in detail. Our
 conclusion was that `same_datatypes` was already handling all cases of `same_descriptors`, which is why I removed it.

crates/store/re_chunk_store/src/query.rs
grtlr: Yeah, I have been struggling with that too. Too m `ColumnDescriptor` is more like `ColumnInfo`. I don't know what better names are for these two types in particular, though.

crates/store/re_chunk/src/transport.rs
grtlr: Disclaimer: I'm very sorry, I somehow botched this comment using a rebase onto `main`, which is weird because `transport.rs` has not changed.

I vaguely remember this being about the `c.sanity_check()` line? In this case, the outcome of the sanity check is merely a warning, that does not change building of the actual chunk.

crates/store/re_chunk_store/src/store.rs
grtlr: I agree, `ColumnIdentifier` is somewhat of an implementation detail. We use it in places where we the `EntityPath` is part of the context (`ChunkStore::per_column_metadata`, `ViewContentsSelector`, ...).

The main problem is that `ViewContentSelector` leaks its internal representation, which sadly forces `ColumnIdentifier` into the public interface too.

crates/store/re_chunk/src/migration.rs
```diff
@@ -59,8 +36,11 @@ impl Chunk {
                     patched_descriptor.archetype_name =
                         Some(archetype_name.as_str().replace(from, to).into());
                 }
-                patched_descriptor.component_name =
-                    descriptor.component_name.replace(from, to).into();
+
```
Wumpf: It's probably time to remove this method, `patched_for_blueprint_021_compat`. Just so hard to reason about this old migration code now. This one is even a couple of versions old and we said we only fully support version to version compatibility :)

crates/store/re_chunk_store/src/dataframe.rs
```diff
@@ -395,87 +397,47 @@ impl ChunkStore {
...
-            .and_then(|per_descr| per_descr.iter().next());
-
-        let component_descr = column_info.map_or(
-            ComponentDescriptor::new(selected_component_name),
-            |(descr, _metadata)| descr.clone(),
-        );
-        component_descr.component_name.sanity_check();
+        // Unfortunately, we can't return an error here, so we craft a default descriptor and
+        // add information to it that we find.
+        let mut result = ComponentColumnDescriptor {
```
Wumpf, line 405: the note about `// TODO(#7699)` regarding interning is still valid though?

crates/store/re_chunk_store/src/query.rs
```diff
@@ -272,7 +272,60 @@ impl ChunkStore {
...
+                && descr.archetype_name == archetype_name
+        };
+
+        let static_chunks = self.static_chunk_ids_per_entity.get(entity_path);
+        let static_components_descr =
+            static_chunks
+                .iter()
+                .flat_map(|static_chunks_per_component| {
+                    static_chunks_per_component.keys().filter(matches)
+                });
```
Wumpf, line 294: nit: instead of chaining we could early out if we already found it here.
I.e. something like

```rs
if let Some(descr) = static_components_descr.next() {
    return descr;
}
```

drawback is that you can't do the sanity check anymore then

also, I'd check temporal chunks here first because given typical data it's more likely to find things there

crates/store/re_chunk_store/src/writes.rs
```diff
@@ -368,30 +358,44 @@ impl ChunkStore {
...
+                .or_insert((
+                    component_descr.clone(),
+                    ColumnMetadataState {
+                        is_semantically_empty: true,
+                    },
+                    list_array.value_type().clone(),
+                ));
             {
+                if *datatype != list_array.value_type() {
+                    // TODO(grtlr): we might even need to make the decision to drop that chunk entirely?
```
Wumpf: do you know what practically is gonna happen down the line? 🤔 
dropping is probably best since lots of systems assume this to never ever happen. Do we spec this stuff somewhere properly? :/

crates/store/re_chunk_store/tests/dataframe.rs
```diff
@@ -0,0 +1,136 @@
...
+};
+use re_log_types::{
+    EntityPath, build_frame_nr,
+    example_components::{MyColor, MyLabel, MyPoint, MyPoints},
+};
+use re_sorbet::ChunkColumnDescriptors;
+use re_types::Archetype as _;
+
+#[test]
+fn schema() -> anyhow::Result<()> {
```
Wumpf: I'm slow, this could have helped :)
```suggestion
/// Tests whether the store has the expected schema after populating it with a chunk.
#[test]
fn schema() -> anyhow::Result<()> {
```

crates/store/re_chunk_store/tests/dataframe.rs
```diff
@@ -0,0 +1,136 @@
...
+            MyPoints::descriptor_points(),
+            MyPoints::descriptor_indicator(),
+        ]
+    );
+
+    Ok(())
+}
+
+#[test]
+fn schema_for_query() -> anyhow::Result<()> {
```
Wumpf:
```suggestion
/// Tests whether the `schema_for_query` for a given query has the expected contents.
#[test]
fn schema_for_query() -> anyhow::Result<()> {
```

rerun_py/rerun_sdk/rerun/_log.py
```diff
@@ -218,6 +218,8 @@ def _log_components(
     instanced: dict[ComponentDescriptor, pa.Array] = {}
 
     descriptors = [comp.component_descriptor() for comp in components]
+    print(descriptors[0])
+    print(descriptors[1])
```
Wumpf: left over debug prints!

rerun_py/rerun_sdk/rerun/_baseclasses.py
```diff
@@ -85,37 +85,35 @@ def __str__(self) -> str:
...
+            return f"{archetype_field_name}#{component_name}"
+        elif archetype_name is not None and component_name is not None:
+            return f"{archetype_name}:{archetype_field_name}#{component_name}"
 
-        return component_name
+        return archetype_field_name
 
-    def with_overrides(self, *, archetype_name: str | None, archetype_field_name: str | None) -> ComponentDescriptor:
+    def with_overrides(self, *, archetype_name: str | None, component_name: str | None) -> ComponentDescriptor:
         """Unconditionally sets `archetype_name` & `archetype_field_name` to the given ones (if specified)."""
```
Wumpf:
```suggestion
        """Unconditionally sets `archetype_name` & `component_name` to the given ones (if specified)."""
```

rerun_py/rerun_sdk/rerun/_baseclasses.py
```diff
@@ -85,37 +85,35 @@ def __str__(self) -> str:
...
-            component_name,
+            archetype_field_name,
+            component_name=component_name,
             archetype_name=archetype_name,
-            archetype_field_name=archetype_field_name,
         )
 
-    def or_with_overrides(self, *, archetype_name: str | None, archetype_field_name: str | None) -> ComponentDescriptor:
+    def or_with_overrides(self, *, archetype_name: str | None, component_name: str | None) -> ComponentDescriptor:
         """Sets `archetype_name` & `archetype_field_name` to the given one iff it's not already set."""
```
Wumpf:
```suggestion
        """Sets `archetype_name` & `component_name` to the given one iff it's not already set."""
```

crates/viewer/re_viewer_context/src/component_ui_registry.rs
```diff
@@ -422,17 +422,24 @@ impl ComponentUiRegistry {
             return;
         }
 
+        let Some(component_name) = component_descr.component_name else {
+            re_log::warn_once!(
+                "Cannot show raw component ui for descriptors without component name: {component_descr}"
+            );
+            return;
+        };
```
Wumpf, line 430: is that true? We should just always use the fallback one then I reckon. That's also gonna be the case for future-AnyValues.
(Or is it because we do a datatype lookup somewhere that we can't do otherwise? (yet?))

applies to occurences below as well

crates/viewer/re_viewer_context/src/viewer_context.rs
```diff
@@ -297,12 +297,7 @@ impl ViewerContext<'_> {
                 .store()
                 .lookup_datatype(&component)
                 .or_else(|| self.blueprint_engine().store().lookup_datatype(&component))
-                .unwrap_or_else(|| {
-                    if !component.is_indicator_component() {
-                        re_log::error_once!("Could not find datatype for component {component}. Using null array as placeholder.");
-                    }
-                    arrow::datatypes::DataType::Null
-                })
+                .unwrap_or(arrow::datatypes::DataType::Null)
```
Wumpf: Hitting this is still an error if there's a component name, no? :thinking:

crates/viewer/re_time_panel/src/time_panel.rs
```diff
@@ -933,8 +933,7 @@ impl TimePanel {
                                 .show_flat(
                                     ui,
                                     list_item::LabelContent::new(format!(
-                                        "{kind} {} component, logged {num_messages}",
-                                        component_descr.component_name.short_name()
+                                        "{kind} {component_descr} component, logged {num_messages}",
```
Wumpf, line 931: is this using the display name that leaves out component name by default?

crates/viewer/re_selection_panel/src/visualizer_ui.rs
```diff
@@ -216,9 +217,15 @@ fn visualizer_components(
         let result_default = query_result.defaults.get(component_descr);
         let raw_default = non_empty_component_batch_raw(result_default, component_descr);
 
-        let raw_fallback = visualizer
-            .fallback_provider()
-            .fallback_for(&query_ctx, component_descr.component_name);
+        // If we don't have a component name, we don't have a way to retrieve a fallback. Therefore, we return a `NullArray` as a dummy.
```
Wumpf, line 219: maybe better to push that down into `fallback_for`
will be easier when we then later refactor fallback system to be a registry of simple descriptor rules (I still very much intend to go there :))

crates/viewer/re_chunk_store_ui/src/chunk_ui.rs
```diff
@@ -103,7 +103,7 @@ impl ChunkUi {
                     ui.style_mut().wrap_mode = Some(egui::TextWrapMode::Truncate);
 
                     let response = ui
-                        .button(component_desc.component_name.short_name())
+                        .button(component_desc.archetype_field_name.to_string())
```
Wumpf: nit: 
```suggestion
                        .button(component_desc.archetype_field_name.as_str())
```

crates/viewer/re_time_panel/src/time_panel.rs
grtlr: It should, yes.

crates/store/re_chunk_store/src/writes.rs
grtlr: Unfortunately, I don't know. @jleibs or @teh-cmc might have more knowledge / opinions?

jleibs: Ideally, we should never drop data. Even if users do awful things and change the datatype.

What is required to happen is that we split the chunks because all data for a column must have the same type in any given chunk.

Beyond that we eventually need pathways to do one of:
1. Coerce the column to a uniform type across all chunks
2. Split/quarantine the column into a separate type-scoped column

crates/store/re_chunk_store/src/store.rs
jleibs: Got it. The name is the bigger problem in that case then. It's actually something like `ArchetypeFieldIdentifier`, where `Entity` + `ArchetypeField` = `Column`

PR #10084, opened 2025-05-27
Implement hot-reloading of design tokens / theme

emilk:
Just enabled in debug builds for us developers, while in the workspace


Wumpf (approved):
bit annoying to spawn a thread for that, but at least it's neatly self-contained this way

Issue #9433, opened 2025-03-31
First person view 0 movement speed


szaboboldizsarmark:
**Unable to move in first person (orbit camera can move) when the scene is infinitely small (e.g. only consists of a single point**

---

**The issue**
This is an oversight, rerun calculates the orbit camera movement speed based on the zoom level, but since the first person view camera doesn't zoom in or out, a different approach had to be taken, where the movement speed of the camera is based on the scene size:

Small scene size -> small movement speed
Big scene size -> a lot of movement speed

When the scene size is infinitely small (e.g. scene that only has a single Points3D point), the movement speed of the first person camera will be 0, thus you are unable to move. Having your scene consisting of a single point is actually something that you might want to do, if you are just visualizing a position for example (this is how I discovered the bug).

---

**Solution**
Calculating the movement speed based off of scene size is a terrible idea in general, what if you had a huge scene but wanted the camera precisely (slowly)? You can't do that, since movement speed is fixed.

If we look back how the orbit camera handles the speed calculation we can see that we use the mouse wheel to adjust the zoom which intern adjusts the movement speed. When using the first person camera the mouse wheel actually doesn't do anything (since zooming in is only relevant in orbit camera), why not just let the user adjust the movement speed with the mouse wheel? This would solve all of our issues.

Now we can:
- Fine tune our movement speed to enable slow and precise movements (just like with the orbit camera), while still letting us increase it to move great distances when our scene is large
- Completely solves the 0 movement speed bug when the scene is only a single point.

Note:
Alongside making the movement speed adjustable with the mouse wheel, I would also keep the current calculation, but instead use it to figure out an initial movement speed, and then the user can adjust if they feel like it. I would also clamp this initial movement speed to a non zero value, to prevent the initial movement speed being zero when the scene is infinitely small.

A lot of video games let you use this type of camera control (adjustable speed) when in spectator mode (e.g. Minecraft, Space Engineers, etc...), so implementing this would make a lot of sense.

---

**To Reproduce**
1. Add a single Points3D point to a newly created scene with nothing else in it
2. Switch to the first person camera view
3. Experience not being able to move

**Desktop (please complete the following information):**
 - OS: 

**Rerun version**

**Additional context**

Wumpf:
>  why not just let the user adjust the movement speed with the mouse wheel

part of blueprint configurable camera, see 
* https://github.com/rerun-io/rerun/issues/6757

[...]

Issue #6825, opened 2024-07-09
Store (eye) camera properties of 3D View in Blueprint


Wumpf:
This makes it
* configurable via blueprint
* persisted in blueprint files
* visible from ui

Pre-requisite for pretty much all future camera improvements since it makes it a lot easier to debug things

michel-ristau-st:
Could you please check my clarification in #11297 (which is marked as duplicate)? It explains that I'm referring to API/script-based camera control, not to control by the GUI or Blueprints. @Wumpf

Wumpf:
replied here https://github.com/rerun-io/rerun/issues/11297#issuecomment-3442055389

Issue #6757, opened 2024-07-04
Make (eye) camera speed configurable


Wumpf:
One of the many things that should go into the camera's blueprint!

See also
* https://github.com/rerun-io/rerun/issues/6013
* https://github.com/rerun-io/rerun/issues/5451

Parskatt:
Imo, it would be nice to just single press shift to increase and ctrl to decrease speed, and make them accumulate.

Parskatt:
But also setting via blueprint is nice, I have somewhat weird scenes where the autosetting is way too fast in general.

Wumpf:
Thanks to @jccampagne  speed is now configurable as a ephemeral view state in the view selection!
Keeping this open since we'd still want to persist things in the blueprint together with other camera properties

Wumpf:
* https://github.com/rerun-io/rerun/pull/10445

added it to the blueprint!

PR #10085, opened 2025-05-27
Add a UI Slider to change speed of eye camera in 3D Views

jccampagne:
### Related

* #9433 
* #6757

### What

- Add a UI slider to change the speed of the eye in the view (for "First Person" and "Orbital").
- Update `state_3d.last_eye_interaction` when slider is changed, otherwise changing the speed will not work unless the user interacts with the view first.

Caveat:
* The values for default, min and max are hardcoded. I was thinking of using the bounding box of the scene (not sure how yet) to set the default value to 1/10th of the largest dimension.
* Still figuring out how this would relates with #6825

### DIFF
--- a/crates/viewer/re_view_spatial/src/ui.rs
+++ b/crates/viewer/re_view_spatial/src/ui.rs
@@ -1,4 +1,5 @@
 use egui::{DragValue, NumExt as _, WidgetText, emath::OrderedFloat, text::TextWrapping};
+use web_time::Instant;
 
 use macaw::BoundingBox;
 use re_format::format_f32;
@@ -170,6 +171,10 @@ impl SpatialViewState {
                 {
                     eye.set_speed(None);
                 } else if previous_speed != speed {
+                    // Make sure the camera is marked as interacted with.
+                    // Otherwise, it will continuously be reset to the default camera.
+                    // (once the speed property is part of the blueprint this concern will go away since all "non-fallback" state then lives in the blueprint store)
+                    self.state_3d.last_eye_interaction = Some(Instant::now());
                     eye.set_speed(Some(speed));
                 }
             });


crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -359,11 +372,7 @@ impl ViewEye {
         drag_threshold: f32,
         bounding_boxes: &SceneBoundingBoxes,
     ) -> bool {
-        let mut speed = match self.mode {
-            EyeMode::FirstPerson => 0.1 * bounding_boxes.current.size().length(), // TODO(emilk): user controlled speed
-            EyeMode::Orbital => self.orbit_radius,
-        };
```
emilk, line 365: Previously the speed was chosen to adapt to the scene size. Small scenes: slow camera, big scenes: fast camera. This means the camera was always at least "okish".

The new code discards all that, and the new default speed of `1` may work for some scenes, but completely fail for others.

I'd like to have the default camera speed still be approximately the same as before. I see a few different ways to approach this:

A) The camera speed the user picks is relative to the old default (e.g. a multiplier, rather than an absolute speed)
B) The default camera speed is based on the old heuristic (instead of being 1)
C) The camera speed is the old heuristic until the user changes it (e.g. use `speed_override: Option<f32>`)

WDYT?

jccampagne:
Additional questions:

### Speed value and user changes

What should happen if:
- a user views a scene in a specific mode (default mode is `EyeMode::Orbital`) with a certain speed;
- the user changes the speed (via the slider),
- and then changes the `mode` of the eye?

Currently the default if `EyeMode::Orbital`, so the default speed would be `self.orbit_radius`; what would happen if the users changes the speed via the slider,  and then changes to `EyeMode::FirstPerson`.


Possible approaches:
- [approch 1.x] only 1 speed per eye and, when the mode changes:
   - [approach 1.1] the speed resets to the default (similar to old code);
   - [approach 1.2] or the speed value stays the same when the mode changes;
- [approach 2] different speed per mode (1 speed for orbital, 1 speed of first person).


### Default speed and scene updates

I also noted that currently (code in main branch) the "default" speed changes as the scene is loaded/updated, as the bounding boxes changes. For example, in the air traffic example, the speed changes from 377673.75 to 882645.75 for orbital mode.

Wumpf:
@jccampagne sorry for the long delay!

I think there should be only a single value for translation speed in orbit eye & ego camera. I mean the basic translation movement is pretty comparable and independent of any other values 🤷. I.e. approach 1.2. This also meshes well with moving this into blueprint:

> Still figuring out how this would relates with https://github.com/rerun-io/rerun/issues/6825

We have to introduce a view property archetype that controls the (eye) camera as a whole. For starters that would include the camera mode and camera speed settings you're introducing here
I.e. we want to move more and more out of the `view state` and into blueprint.

> I also noted that currently (code in main branch) the "default" speed changes as the scene is loaded/updated, as the bounding boxes changes. For example, in the air traffic example, the speed changes from 377673.75 to 882645.75 for orbital mode.

Yes, this is intentional. The problem is that at no point in time we can be certain that the scene is "complete" - maybe all data we got so far was just preperatory and there's still major elements missing in the scene. A bit of a conundrum with all these scene size dependent heuristics.

jccampagne:
Hi @Wumpf,

no worries! I was also away for a few days, just got back.

Thanks for the comments. I just had another look.
I have added the previous heureustic back in function `compute_speed_for_mode()`.

In function `lerp()`, I had to pass `bounding_boxes` to compute the interpolation in the case of the speed being calculated automatically (see `compute_speed_lerp()`). If you can think of a more elegant solution…?


I will have a look at the blueprints; but I thought that was orthogonal to the actual speed calculation and UI update. So, that could be handled in a later step, right?

Also I saw this draft PR: https://github.com/rerun-io/rerun/pull/8211
Is it still current/relevant?

crates/viewer/re_view_spatial/src/eye.rs
jccampagne: Option C was implemented in `speed()` using the `SpeedControl` enum.

see https://github.com/rerun-io/rerun/pull/10085/files#diff-242e5fb1310bf1a865d6c25402303099d00d6a81e76137ccfaaf36db8987b24bR299

Wumpf:
> I will have a look at the blueprints; but I thought that was orthogonal to the actual speed calculation and UI update. So, that could be handled in a later step, right?

yes and no :). The camera and all its properties are right now completely stored in the ephemeral `ViewState`, but really it should be part of the blueprint which means that it is persisted to disk and can be set from the blueprint api.

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -276,6 +287,27 @@ impl ViewEye {
         }
     }
 
+    /// Compute the actual speed depending on the [`EyeMode`].
+    fn compute_speed_for_mode(&self, bounding_boxes: &SceneBoundingBoxes) -> f32 {
```
Wumpf: going with naming scheme we have for blueprint properties (and this should be one eventually) this would be `fallback_speed_for_mode` since "fallback" is the term we use for the dynamically calculated values we use when a blueprint property wasn't specified

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -311,10 +343,27 @@ impl ViewEye {
...
+    fn compute_speed_lerp(
+        &self,
+        other: &Self,
+        t: f32,
+        bounding_boxes: &SceneBoundingBoxes,
+    ) -> SpeedControl {
+        let self_speed = match self.speed {
+            SpeedControl::Auto => self.compute_speed_for_mode(bounding_boxes),
+            SpeedControl::Override(s) => s,
+        };
```
Wumpf: why not
```suggestion
        let self_speed = self.speed(bounding_boxes);
```

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -330,6 +379,7 @@ impl ViewEye {
                 // matters if the user starts interacting half-way through the lerp,
                 // and even then it's not a big deal.
                 eye_up: self.eye_up.lerp(other.eye_up, t).normalize_or_zero(),
+                speed: self.compute_speed_lerp(other, t, bounding_boxes),
```
Wumpf: yeah this is a bit of a weird situation. It's just as broken as what it's beeing done with `mode`. Might as well do the same and just take over `other.speed` here
iirc practically this is used for transitioning on camera double clicks. So I think it's fine to just simplify this

Wumpf:
to clarify on the whole blueprint thing: I would love to have this in the blueprint from the start but I don't want perfect to e in the way of good. I'd rather land this simpler pr first than not having anything for a long while :)
Putting things into blueprint is _probably_ (!) not terribly hard either but has a whole rat-tail of starting a new cross-language type definition, potentially extending examples, writing public api docs, figuring out how the data from the blueprint store is piped through here.. etc.

crates/viewer/re_view_spatial/src/eye.rs
jccampagne: Thanks for catching this.
Fixed in the next commit.

crates/viewer/re_view_spatial/src/eye.rs
jccampagne: Renamed to `fallback_speed_for_mode()`

crates/viewer/re_view_spatial/src/eye.rs
jccampagne: Ok. Simplified in the next commit (revert the logic, deleted `compute_speed_lerp()`).

jccampagne:
> to clarify on the whole blueprint thing: I would love to have this in the blueprint from the start but I don't want perfect to e in the way of good. I'd rather land this simpler pr first than not having anything for a long while :)

I agree - but am OK either way :)

> Putting things into blueprint is _probably_ (!) not terribly hard either but has a whole rat-tail of starting a new cross-language type definition, potentially extending examples, writing public api docs, figuring out how the data from the blueprint store is piped through here.. etc.

Will have a look.


While playing with the app, I thought wouldn't it be great to be able to control the rotational speed as well? But that'll be done with the blueprints :-)

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -159,6 +159,15 @@ pub enum EyeMode {
     Orbital,
 }
 
+/// The speed of a [`ViewEye`] can be computed automatically or set manually.
+#[derive(Clone, Copy, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+enum SpeedControl {
```
jccampagne: Happy with the naming of the enums and type?
I am still not familiar with the naming scheme in rerun.
Happy change to align with the rest (as for `fallback_speed_for_mode`)

Wumpf (changes requested):
ah sorry I think we're still not quite done here. I checked it out now and ran it: with the default selection panel width the speed field is cut off
image

besides I don't think a slider makes sense here because there's no clearly defined limit. Let's used a drag box instead. That should fit even without refactoring the view further (which probably should happen anyways - all the blueprint driven properties below are expanding and those "ad hoc" options are not, quite inconsistent; but one thing at a time)

crates/viewer/re_view_spatial/src/eye.rs
Wumpf: no strong opinion on this 🤷 Would have maybe called it `CameraTranslationSpeed`

ci is picky though about newline between enum variants ;)

jccampagne:
> ah sorry I think we're still not quite done here. I checked it out now and ran it: with the default selection panel width the speed field is cut off...
> besides I don't think a slider makes sense here because there's no clearly defined limit. Let's used a drag box instead. 

I initially used `DragValue`, but I could not find an option to make the scale logarithmic. Using a linear scale in the air traffic example is inconvenient given the size of the world. The user may want to rapidly move around (order of kilometers), and then change the speed granularly to position the camera finely (order of a few meters). In the ARKit example, the user may want to move to the other side of the room (meters) and then align the camera precisely (centimetres).


Right now, the lower and upper bounds are fixed, it's true it's not ideal (it was a TODO actually):

```
Slider::new(&mut speed, 0.001..=1_000_000.0).logarithmic(true)
```

A better alternative could be:
- setting the lower limit to zero (negative speed is not really meaningful);
- and the make the upper limit could depend on the bounding box (eg: rounded to the order of magnitude of the bound size, or reuse the fallback heuristic). While it's possible to set +infinity for the upper bound, it causes translation issues and it's not useable.

(Note about the heuristic: today it's a fraction (0.1) of the length of the velocity - using a rounding function based on magnitude, can compute nice round values, that would look nicer in the UI).

If a slider is still not desirable but a logarithmic scale is, maybe the scaling logic can be implemented in a wrapper around DragValue. I saw there's a [from_get_set](https://docs.rs/egui/latest/egui/widgets/struct.DragValue.html#method.from_get_set) method - maybe that can be used? (not sure what it does exactly yet.

Another approach is to have egui's `DragValue` offer the log scale option like `Slider`?

What do you think?


> That should fit even without refactoring the view further (which probably should happen anyways - all the blueprint driven properties below are expanding and those "ad hoc" options are not, quite inconsistent; but one thing at a time)

Ok.

jccampagne:
Sorry about the "was requested for review" noise. Didn't intend to click on that button (no way to undo). There's still things to do (around the slider).

Wumpf:
We use relative speed like this a lot https://github.com/rerun-io/rerun/blob/278f677c565b5ece2922d422d422d1a3c64b9706/crates/viewer/re_component_ui/src/datatype_uis/float_drag.rs#L146
that should give you pretty much the desired effect

jccampagne:
Using `DragValue` now. I kept it simple for now.
Looking at Blueprints...

Wumpf (approved):
thanks!
I just quickly added reset on double click while I had the pr checked out for testing anyways

jccampagne:
Cool.

Looks like `self.state_3d.last_eye_interaction = Some(Instant::now());` is needed though?
UI is not responding well without it during testing?

jccampagne:
To reproduce:

Without `self.state_3d.last_eye_interaction = Some(Instant::now());` in the code.

1. start Rerun
2. load an example
3. click on the 3D view (**without** dragging the mouse - otherwise dragging it will update `self.state_3d.last_eye_interaction`)
4. try to drag the Translation Speed DragValue - nothing will happen.


Dragging the mouse in the 3D view will update `self.state_3d.last_eye_interaction` and then the `DragValue` will work.

Wumpf:
yikes. Thanks, I didn't notice that! 🙇 
Setting `last_eye_interaction` is a bit of a hacky fix for that, but yeah after reading the code a bit more it unfortunately makes sense, bringing it back with a comment

Issue #10042, opened 2025-05-21
"Current decoded frame" in selection panel shows a playing video


nikolausWest:
Selecting a video while paused and scrolling down to the "Current decoded frame" in the in selection panel shows a playing video rather than the current frame

re_viewer 0.24.0-alpha.1 (c593098)

https://github.com/user-attachments/assets/21257ea4-20fd-4272-9abd-4679f3f9b974

Wumpf:
Does not happen on 0.23.x

See for example https://rerun.io/viewer?url=https%3A%2F%2Fapp.rerun.io%2Fversion%2F0.23.2%2Fexamples%2Fdetect_and_track_objects.rrd
Versus it being broken https://rerun.io/viewer/pr/10040?url=https%3A%2F%2Fapp.rerun.io%2Fversion%2Fnightly%2Fexamples%2Fdetect_and_track_objects.rrd

likely a tagged component related regression.

PR #10091, opened 2025-05-27
Fix video frame reference ui playing video

Wumpf:
* Fixes https://github.com/rerun-io/rerun/issues/10042

This was a recent regression caused by the tagged component effort.

### DIFF
--- a/crates/viewer/re_data_ui/src/instance_path.rs
+++ b/crates/viewer/re_data_ui/src/instance_path.rs
@@ -4,7 +4,7 @@ use re_chunk_store::UnitChunkShared;
 use re_entity_db::InstancePath;
 use re_log_types::ComponentPath;
 use re_types::{
-    ArchetypeName, Component, ComponentDescriptor, components,
+    ArchetypeName, Component, ComponentDescriptor, archetypes, components,
     datatypes::{ChannelDatatype, ColorModel},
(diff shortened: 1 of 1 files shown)


crates/viewer/re_data_ui/src/instance_path.rs
```diff
@@ -524,16 +524,25 @@ fn preview_single_blob(
...
         .ok()?;
 
+    // Media type comes typically alongside the blob.
     let media_type = find_and_deserialize_archetype_mono_component::<components::MediaType>(
         components,
         blob_descr.archetype_name,
     )
     .or_else(|| components::MediaType::guess_from_data(&blob));
 
-    let video_timestamp = find_and_deserialize_archetype_mono_component::<components::VideoTimestamp>(
```
emilk, line 533: Why does `find_and_deserialize_archetype_mono_component` work for `MediaType` but not for `VideoTimestamp`?

Wumpf: tried to capture this weird bit in those comments sounds like that worked only partially :)

`MediaType` is actually associated with _several_ archetypes, in fact we don't have a `Blob` archetype but we're interested in the `MediaType` that was in the same archetype as the blob. Lemme make this a lil bit more clear there...

PR #10094, opened 2025-05-28
Make icons sharper on low-DPI screens

emilk:
### What
A lot of icons have a vertical or horizontal center-line. By using an odd icon size, we make it _a lot_ sharper on standard-DPI screens. See example below!

Only possible thanks to the icons now being SVG

#### Before
list_items old

Screenshot 2025-05-28 at 09 41 21

#### After
Screenshot 2025-05-28 at 09 41 27
list_items


emilk:
@gavrelina says she has a better plan!

PR #10096, opened 2025-05-28
Fix nightly asset sync

jprochazk:
Don't gzip already gzipped files


jprochazk:
I'll merge this and run the nightly on `main`

Issue #9849, opened 2025-04-30
Fix wasm compilation of Wasm target on macOS


abey79:
Currently, a newer version of llvm/clang must be installed on mac, because the Apple-provided one fails building wasm with:

```
failed to build archive at /Users/tsaucer/src/rerun/target_wasm/wasm32-unknown-unknown/release/deps/libpsm-43030b67c35315ca.rlib: LLVM error: section too large
```

We should figure out a way to either fix the build, or make it so that using pixi solves the issue.

emilk:
The PR that first caused this problem was:
* https://github.com/rerun-io/rerun/pull/9764

Wumpf:
To get it to work I had to specify homebrews llvm installation directly. I.e. `PATH="/opt/homebrew/opt/llvm/bin/:$PATH"`

Wumpf:
should check whether 
* https://github.com/rerun-io/rerun/pull/10214

had any effect

ntjohnson1:
Confirmed with git bisect that this has been the case since datafusion was added.

This PR under datafusion appears to work around the issue https://github.com/apache/datafusion/pull/16418

However, IIUC rerun->datafusion->a few layers->[this assembly](https://github.com/rust-lang/stacker/blob/master/psm/src/arch/wasm32.s) -> [this object](http://github.com/rust-lang/stacker/blob/master/psm/src/arch/wasm32.o). Didn't get a chance to go much deeper than that but I wonder if we can assemble that on mac to fix the root cause (since we know things play nice with latest llvm).

ntjohnson1:
I'll be working on this intermittently, so trying to capture in case someone else wants to pick this up to push through faster

Potential next steps (in decreasing preference) :
* Try to isolate build to pixi environment (more or less debugging build inside of conda from [this](https://github.com/rerun-io/rerun/pull/10216))
   * NOTE: On [this issue](https://github.com/rustwasm/wasm-pack/issues/1381) someone was able to create a minimum repro just around psm, that is probably a tighter iteration loop to validate against
* Update our datafusion version to a fork of datafusion that includes the PR above or wait for the next release and update to that
* Try to see if we can generate the object from the assembly included in the psm crate source. Can we debug/modify that code to be compatible with an older version of llvm/apple clang? The fact it works with newer llvm and the source code comment makes me think there is just a compatibility issue in that crate rather than the crate not supporting wasm correctly.

ntjohnson1:
[...]

PR #10110, opened 2025-06-02
Clang 18 for everyone and everything

teh-cmc:
Switch all pixi things to clang 18, which I've confirmed is capable of linking latest `mimalloc` on all platforms (as opposed to the current clang 16 we're using).

* Should also fix https://github.com/rerun-io/rerun/issues/9849


teh-cmc:
Will undraft once (if) we pass full CI.

teh-cmc:
@rerun-bot full-check

Wumpf:
note that on MacOS we still have to use system-clang for _some_ thing - very annoying but couldn't figure out how to get things to link otherwise, see https://github.com/rerun-io/rerun/pull/10214

Wumpf:
happy to take this over when I have a minute

Wumpf:
@Wumpf apparently never had a minute...


Also, at this point the macos clang situation has moved on rather weirdly.

Issue #9886, opened 2025-05-05
Eliminate notebook widget timeouts


jprochazk:
The way forward here is to not report a warning at all. We initially added the timeout so users would know that they need to change the env var if the Viewer doesn't show up due to the widget not being accessible, because there is no other indication that it won't, and you'd end up sitting there waiting for a while. That was a really bad UX.

I think we're okay with it just taking a while, even upwards of half a minute depending on your environment... Some notebook environments are just really slow.

We'll instead check if the widget asset is accessible by sending a HEAD request, and fail early if it isn't. If we're running with `inline` or `serve-local`, we won't check for the asset at all. Then we can disable the `block_until_ready` mechanism by default. We already buffer data on the Python side before the Viewer is ready, so that is not going to be an issue.

Wumpf:
Related to
* https://github.com/rerun-io/rerun/issues/9863

PR #10111, opened 2025-06-02
Improve `rerun_notebook` startup times

jprochazk:
### Related

* Closes https://github.com/rerun-io/rerun/issues/9886

### What

We want to store the Wasm binary as a separate file, and let the Viewer load it via `fetch` at runtime. This leads to the best outcomes in terms of startup times.

#### Build changes

The `rerun_notebook` build has been updated to:
* Copy the Wasm binary from node_modules to `/static`
* Preserve special inline marker comments (`//!<INLINE-MARKER>` at the time of writing)

We manually copy the Wasm binary to the output directory, because any of my attempts to do this via `esbuild` failed. We just need the file there, so we do it manually.

We need to mark the start and end of a block of code that should be replaced with the inlined Wasm binary + the glue that loads it from a base64-encoded string. We use `esbuild` to bundle the notebook widget, and `esbuild` does not preserve regular comments. There are also no options to enable this behavior.
The exception is "legal comments" which are normally used for e.g. copyright information. The marker comments just happen to be something that `esbuild` considers a "legal comment".

The `rerun-js/web-viewer` build no longer combines the Wasm and JS into a single file - this is instead done at _runtime_, by the `rerun_notebook` package, becuase we don't want to publish two copies of the Wasm binary (one as its own file, one inlined into the JS widget) when publishing the package. It has a negligible startup cost, which is paid once at time of `import rerun.notebook`, and only if `RERUN_NOTEBOOK_ASSET` is set to `inline`.

The asset server has been updated to also serve the `re_viewer_bg.wasm` file. It serves it gzipped, with the right `content-type` so that it can be stream-compiled.

#### Check for assets

We now send a `HEAD` request to check if the assets we need are accessible on the browser side, by injecting a small JS snippet in parallel with the regular Viewer widget loading. This injection is done via a combination of `ipywidgets.HTML` and a base64-encoded javascript file inserted into a `style` tag's `onload` property.

It's done in this convoluted way because `ipywidgets.HTML` uses `innerHTML` to set the content of the element it creates. JS is not executed by running `innerHTML` (see https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations)... except if it's in special properties which are treated as event listeners. It's base64-encoded so that the javascript we inject does not need to avoid using quotes or other symbols which may cause the resulting HTML string to not parse properly.

If the widget isn't accessible, we let the user know with a clear error message. 

Additionally, the `rerun_notebook` import has been moved from inside the `Viewer` constructor to the top-level again. Despite `rerun_notebook` being an optional package, we've been able to do this without causing any issues for a while, because we no longer import the `rerun.notebook` package anywhere inside `rerun-sdk` eagerly. Something like blueprint `_ipython_display_` impls import it inside the function body, at which point we expect the user to have the `rerun_notebook` package installed, because we expect `_ipython_display_` to be called from the context of a notebook. If someone tries to call it outside of that context, that's their problem ¯\\\_(ツ)\_/¯.

#### `block_until_ready=False` by default

If the widget asset was not accessible for whatever reason, the widget would never finish loading without any error messages, unless you opened the browser devtools console. That was a super frustrating experience, so we instead set a hard limit (10 seconds) for how long the Viewer should take to load, and blocked cell execution until it did. Those 10 second turn out to just not be enough in many environments.

Because we now check if the required assets are accessible in the browser, we no longer need this, as we won't ever timeout in the same way. We already buffer all input data before the widget has loaded, so there is no issue with letting cell execution continue while the widget is loading.


jprochazk:
Current status: Investigating why "restart kernel and run all cells" does not work. Viewers show up, but receive no data. Others don't show up at all. At least the notebook doesn't get stuck like on `main`

jprochazk:
Decided to leave that as a separate task, because for me it's also broken on main.

jprochazk:
Opened an issue for the 2nd part:
- https://github.com/rerun-io/rerun/issues/10176

rerun_notebook/src/rerun_notebook/__init__.py
```diff
@@ -36,24 +41,123 @@
...
+    with ThreadPoolExecutor(max_workers=2) as e:
+        results = list(e.map(_is_url_accessible, urls_to_check))
+
+    success = True
+    for url, exists in results:
+        if not exists:
+            success = False
+            print(f'"{url.geturl()}" is not accessible')
+    if not success:
+        raise ValueError(
```
jleibs: I tried this out by overriding RERUN_NOTEBOOK_ASSET.

Getting a `ValueError` raised at-import-time seems like a fairly poor experience:
image

My suggestion for a compromise would be to print a warning message and store a global. Then check the state of the global and raise at usage time.

rerun_notebook/src/rerun_notebook/__init__.py
```diff
@@ -36,24 +41,123 @@
...
+    ]
+
+    with ThreadPoolExecutor(max_workers=2) as e:
+        results = list(e.map(_is_url_accessible, urls_to_check))
+
+    success = True
+    for url, exists in results:
+        if not exists:
+            success = False
+            print(f'"{url.geturl()}" is not accessible')
```
jleibs: We haven't been super consistent about this, but this is one that seems like it should go to stderr.

rerun_notebook/src/rerun_notebook/__init__.py
jleibs: Reading on I see other bad values of `RERUN_NOTEBOOK_ASSET` do raise value-errors at import-time. They do so deterministically so I guess this is consistent with that if we want to leave it.

rerun_notebook/src/rerun_notebook/__init__.py
```diff
@@ -36,24 +41,123 @@
...
+    ESM_MOD = _inline_widget()
+else:  # remote widget
     ESM_MOD = ASSET_ENV
+    # note that the JS expects the Wasm binary to exist at the same path as itself
     if not (ASSET_ENV.startswith(("http://", "https://"))):
         raise ValueError(f"RERUN_NOTEBOOK_ASSET should be a URL starting with http or https. Found: {ASSET_ENV}")
+    if not (ASSET_ENV.endswith("widget.js")):
+        raise ValueError(f"RERUN_NOTEBOOK_ASSET should be a URL pointing to a `widget.js` file. Found: {ASSET_ENV}")
+
+    _check_if_assets_accessible(ASSET_ENV)
```
jleibs: I'm realizing one fundamental (and unavoidable) problem with this is that this check is from from the python kernel -- executing on the server machine hosting the notebook.

But the actual future access will happen from the users browser, which could be a different machine with different networking constraints.

jprochazk: That's true. If the user is running some setup where their Jupyter server is on a cloud machine somewhere, and they can't access the widget but the server can for whatever reason, they have the option of using `inline` in that case. But they won't see any error messages, things will just silently fail.

I'm not really sure how to surface an error message in that case. Because the server can access the widget, it won't raise that `ValueError`. So it must be done by the client. But the client can't start, because it can't be downloaded... Maybe the anywidget glue for starting the JS widget should handle that. That sounds like something we could contribute to anywidget.

jprochazk:
image

rerun_js/web-viewer/index.ts
```diff
@@ -3,23 +3,27 @@ import type { WebHandle, wasm_bindgen } from "./re_viewer";
...
-async function fetch_viewer_wasm() {
-  return fetch(new URL("./re_viewer_bg.wasm", import.meta.url));
+async function fetch_viewer_wasm(base_url?: string): Promise<Response> {
+  //!<INLINE-MARKER>
+  if (base_url) {
+    return fetch(new URL("./re_viewer_bg.wasm", base_url))
+  } else {
+    return fetch(new URL("./re_viewer_bg.wasm", import.meta.url));
+  }
+  //!<INLINE-MARKER>
```
jleibs: Nit: feel free to do this in another PR (or not at all) since it was already like this.

I've been bitten before by symmetric markers in the past. Being explicit about `INLINE-MARKER-OPEN` and `INLINE-MARKET-CLOSE` might be marginally more robust.

rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -275,13 +276,14 @@ def display(self, block_until_ready: bool = True) -> None:
...
 
+        display(self._error_widget)
         display(self._viewer)
 
         if block_until_ready:
             self._viewer.block_until_ready()
 
     def _ipython_display_(self) -> None:
-        self.display(block_until_ready=True)
+        self.display()
```
jleibs, line 286: I don't remember why this was `True` before. I assume you have some context about why we changed it, so maybe an explicit value / comment here is still warranted.

rerun_notebook/src/rerun_notebook/__init__.py
```diff
@@ -36,24 +42,101 @@
...
-else:
+elif ASSET_ENV == ASSET_MAGIC_INLINE:  # inline widget
+    # in this case, `ESM_MOD` is the contents of a file, not its path.
+    ESM_MOD = _inline_widget()
+else:  # remote widget
     ESM_MOD = ASSET_ENV
+    # note that the JS expects the Wasm binary to exist at the same path as itself
     if not (ASSET_ENV.startswith(("http://", "https://"))):
         raise ValueError(f"RERUN_NOTEBOOK_ASSET should be a URL starting with http or https. Found: {ASSET_ENV}")
+    if not (ASSET_ENV.endswith("widget.js")):
```
jleibs, line 117: My read is that we still explicitly point at `widget.js`, but also expect to find an adjacent `re_viewer_bg.wasm`?

There are probably some docs somewhere that should be updated to mention this where we talk about self-hosting your own asset.

emilk:
@jleibs do you still intend to review this?

jleibs (approved):

jleibs:
> @jleibs do you still intend to review this?

lol, looks like I was actively re-reviewing when you asked.

rerun_py/rerun_sdk/rerun/notebook.py
jprochazk: I changed this from `_repr_mimebundle_` + `_repr_keys` to `_ipython_display`. It was likely because I wanted to call a blocking `display` inside `_ipython_display_`, and that didn't make sense inside the `repr` methods.

As for why this is explicitly set to `True`, I don't really know. I probably just forgot :grimacing:... I suspect at some point `display` had `block_until_ready` set to `False`, and I had to explicitly set it to `True`, then later changed the default to `True` and didn't remove this explicit kw param.

PR #10113, opened 2025-06-02
Make ViewContext more useful and use it more aggressively, remove slow fallback provider path

Wumpf:
* `QueryContext` previously had an optional view context which made it pretty hard to use and broke the neat hierarchy of contexts
* `ViewContext` didn't know the class id which is a a very useful thing to know about a view (and what triggered this pr..)
* `ViewContext` no longer contains a collection of all visualizers. The previous collections were actually potentially outdated as there was a hacky ad-hoc cache there but visualizers may be added at runtime. In other situations we had "processed" and freshly initialized visualizers next to each other which was very confusing.
      * this was used for some ui aspects (not perf critical) and for a super slow path for fallback provisioning - so I removed this slow path!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7789,7 +7789,6 @@ name = "re_view"
 version = "0.24.0-alpha.1+dev"
 dependencies = [
  "ahash",
- "arrow",
  "egui",
  "glam",
  "itertools 0.14.0",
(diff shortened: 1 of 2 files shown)


emilk (approved):
Thanks for cleaning this up 🙏

PR #10119, opened 2025-06-02
Update glam to 0.30, and replace `re_math` with `macaw`

emilk:
We only published `re_math` because `macaw` wasn't on crates.io. Now it is. So let's use it. It's even up-to-date with glam! And now so are we!

Best to review the commits individually!


abey79 (approved):
Nice!

PR #10121, opened 2025-06-03
Rename `StoreHubEntry` to `RecordingOrTable`

abey79:
### Related

* Part of https://github.com/rerun-io/dataplatform/issues/777

### What

Pure refactor: let's not name entry things that are not (always) entries. It got me badly confused.

As an aside, the relationship between entries (aka Dataset/Table/etc.) and the recording panel/what-is-currently-displayed is cursed:
- A recording (leaf in recording panel) is no an entry, but _may_ be a partition of a (dataset) entry.
- Application ID are technically not entries, but are presented in a similar way to datasets (aka a "folder" that contains recording/partition), which are entries.
- Tables are entries, but are leaf item in the recording panel

### DIFF
--- a/crates/viewer/re_data_ui/src/item_ui.rs
+++ b/crates/viewer/re_data_ui/src/item_ui.rs
@@ -736,7 +736,9 @@ pub fn entity_db_button_ui(
                     });
             if resp.clicked() {
                 ctx.command_sender()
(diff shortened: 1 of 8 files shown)


Wumpf (approved):
👍

Issue #7484, opened 2024-09-24
Support for streaming encoded video from continuous log/send calls


Wumpf:
We want to be able to support streaming video in via continuous log/send calls.

We want to be overall compatible with compressed video as it shows up in mcap, see https://github.com/foxglove/foxglove-sdk/blob/main/schemas/ros2/CompressedVideo.msg, and stick otherwise closely to WebCodec's specification https://www.w3.org/TR/webcodecs

------------------

The plan so far.

* [x] Milestone 1: Load pre-saved demuxed video streams to Rerun
  * supports all codecs we have today
  * can append more frames over time, but experience in Rerun doesn't need to be a good experience yet, accept stutters etc.
  * basic Rerun archetype & components in place
  * Sub milestones / shippables:
    * [x] 1.1: h264 only. No web support yet, lacks ui inspection on panel
    * [x] 1.2: Web support
    * [ ] ~1.3: Support all currently supported codecs on web/native respectively~
         * we'll do this at a later point. created subissues for each codec
    * [x] 1.4: UI experience of video stream vs mp4 on par
* [x] Milestone 2: Stream video streams to Rerun in realtime (but allowing for quite some latency)
    * comes with example app that continuously streams data to Rerun
    * experience of looking at video in the Viewer is smooth all around. Latency (from logging application) is around <100ms for local video
    * should play nice with GC as well - should not break down when loosing video chunks from the start
    * may need to implement a "backpressure" mechanism on the timeline cursor so the decoder has time to decode before we move on. 
* Related but not (strictly in-scope for this project:
   * Cover more codecs in native, improve decoder performance, mitigate existing glitches
      * improve frame enqueuing logic, in particular on web
      * add new decoder types to Rerun native. Candidates to (re-)evaluate:
        * gstreamer-rs
        * ffmpeg via bindings
        * [vk-video](https://github.com/software-mansion/smelter/tree/master/vk-video) linux/windows only (but potentially very high performance!)
        * openh264 (narrow scope but potentially more stable than what we have on ffmpeg)
    * BFrame support in demuxed data
      * trivial for formats other than h264/h265 since there's no decoder/presentation reordering? May need to extend sample with optional decode timestamps
      * https://github.com/rerun-io/rerun/issues/10090

------------------

## Archived: Initial sketch & notes

The rough sketch for this feature would be to introduce an `VideoStream` archetype to which users continuously add `VideoSample` components (or in manual batches via `send_column`).

[...]

PR #10123, opened 2025-06-03
Prepare `re_video` and `re_renderer::VideoPlayer` for streaming video

Wumpf:
### Related

* #7484
* Directly followed by https://github.com/rerun-io/rerun/pull/10125

### What

First PR in a series of PRs enabling ✨ video streaming ✨ 

Changes to how we capture the metadata of a video:
* Introduce new `StableIndexDeque` and use it for everything in `re_video` that should support dropping & appending
* `VideoData` -> `VideoDataDescription`
* remove `re_video::Config`
* make `StsdBox` optional on `VideoDataDescription` but add optional codec & `coded_dimensions` instead
* duration of video samples and video as a whole is now optional (since video streams are open ended!)

Changes to the video player (re_renderer):
* allocate video frame texture lazily (we may not know the size ahead of time!)
* `VideoChunkDecoder` -> `VideoSampleDecoder` for more consistent naming
* video data description is no longer held via `Arc` but instead owned. This allows for editing it later on
* handle some GC related cornercases to avoid out of bounds access notably still missing: be aware of dynamically growing GOPs


crates/utils/re_video/src/demux/mod.rs
```diff
@@ -60,12 +109,20 @@ pub struct VideoData {
     /// Samples must be decoded in decode-timestamp order,
     /// and should be presented in composition-timestamp order.
     ///
-    /// In MP4, one sample is one frame.
-    pub samples: Vec<Sample>,
+    /// We assume one sample yields exactly one frame from the decoder.
+    ///
+    /// To facilitate streaming, samples may be removed from the beginning and added at the end,
+    /// but individual samples are never supposed to change.
+    pub samples: StableIndexDeque<Sample>,
```
Wumpf, line 135: Maybe rename `Sample` to `SampleMetadata` or something

crates/utils/re_video/src/stable_index_deque.rs
```diff
@@ -0,0 +1,322 @@
...
+        &self.vec[index - self.index_offset]
+    }
+}
+
+impl<T> std::ops::IndexMut<usize> for StableIndexDeque<T> {
+    #[inline]
+    fn index_mut(&mut self, index: usize) -> &mut Self::Output {
+        &mut self.vec[index - self.index_offset]
+    }
+}
```
Wumpf, line 346: strongly consider removing those! (try at least :))

crates/utils/re_video/src/decode/av1.rs
```diff
@@ -258,12 +258,15 @@ fn create_frame(debug_name: &str, picture: &dav1d::Picture) -> Result<Frame> {
...
-            duration: Time(picture.duration()),
+            // TODO(andreas): dav1d has a user-data field that isn't exposed yet.
+            // We should us that to populate these fields.
+            is_sync: None,
+            sample_idx: None,
+            frame_nr: None,
             latest_decode_timestamp: None,
+
+            presentation_timestamp: Time(picture.timestamp().unwrap_or(0)),
+            duration: (picture.duration() == 0).then_some(Time(picture.duration())),
```
emilk: Is this really correct? Looks like it should be 

```suggestion
            duration: (picture.duration() != 0).then_some(Time(picture.duration())),
```

Has this been tested? :)

crates/viewer/re_view_spatial/src/visualizers/videos.rs
```diff
@@ -188,13 +188,16 @@ impl VideoFrameReferenceVisualizer {
...
+                        video_resolution =
+                            glam::vec2(coded_dimensions[0] as _, coded_dimensions[1] as _);
+                    }
 
                     match video.frame_at(
                         ctx.render_ctx(),
                         player_stream_id,
                         video_timestamp.as_secs(),
-                        &video_data,
+                        &std::iter::once(video_data.as_ref()).collect(),
```
emilk: I wish there was a nicer way to do this, but I don't know what that would be

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -122,37 +119,50 @@ impl VideoPlayer {
...
+    ///
+    /// The video data description may change over time by adding and removing samples and GOPs,
+    /// but other properties are expected to be stable.
     pub fn frame_at(
         &mut self,
         render_ctx: &RenderContext,
         time_since_video_start_in_secs: f64,
-        video_data: &[u8],
+        video_description: &re_video::VideoDataDescription,
+        video_data: &StableIndexDeque<&[u8]>,
```
emilk: What is the (outer) index of this? Is it `re_video::Sample::buffer_index`? If so, I suggest renaming this to

```suggestion
        video_buffers: &StableIndexDeque<&[u8]>,
```

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -248,14 +254,24 @@ impl VideoPlayer {
...
+            let current_pts = requested_sample
+                .map(|s| s.presentation_timestamp)
+                .unwrap_or(Time::MIN);
+
+            let requested_sample = video_description.samples.get(requested_sample_idx);
+            let requested_sample_pts = requested_sample
+                .map(|s| s.presentation_timestamp)
+                .unwrap_or(Time::MIN);
+
+            if requested_sample_pts < current_pts {
```
emilk, line 269: There's a bunch of things that can become `None` above.

If they become `None`, should we or should we not enter the body of this `if` here?

It seems to me the readability would go up a lot of this was made explicit by not using `.unwrap_or(Time::MIN)` but instead propagating `None` so it is clear what happens in that case.

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -311,69 +327,40 @@ impl VideoPlayer {
...
     /// Enqueue all samples in the given GOP.
     ///
     /// Does nothing if the index is out of bounds.
-    fn enqueue_gop(&mut self, gop_idx: usize, video_data: &[u8]) -> Result<(), VideoPlayerError> {
-        let Some(gop) = self.data.gops.get(gop_idx) else {
+    fn enqueue_gop(
+        &mut self,
+        video_description: &re_video::VideoDataDescription,
+        gop_idx: usize,
+        video_data: &StableIndexDeque<&[u8]>,
```
```suggestion
        video_buffers: &StableIndexDeque<&[u8]>,
```

crates/viewer/re_data_ui/src/video.rs
```diff
@@ -48,15 +49,19 @@ pub fn video_result_ui(
     }
 }
 
-fn video_data_ui(ui: &mut egui::Ui, ui_layout: UiLayout, video_data: &VideoData) {
+fn video_data_ui(ui: &mut egui::Ui, ui_layout: UiLayout, video_data: &VideoDataDescription) {
```
emilk:
```suggestion
fn video_data_ui(ui: &mut egui::Ui, ui_layout: UiLayout, video_descr: &VideoDataDescription) {
```

would be a better name, now that we've made it clear it doesn't actually contain the video data

emilk: here and everywhere

crates/viewer/re_data_ui/src/video.rs
```diff
@@ -257,13 +270,23 @@ pub fn show_decoded_frame_info(
         // but the point here is not to have a nice viewer,
         // but to show the user what they have selected
         ui.ctx().request_repaint(); // TODO(emilk): schedule a repaint just in time for the next frame of video
-        ui.input(|i| i.time) % video.data().duration().as_secs_f64()
+        let time = ui.input(|i| i.time);
+        if let Some(duration) = video.data().duration() {
+            time % duration.as_secs_f64()
+        } else {
+            time
```
emilk: Using naked `time` here makes no sense - it has no relation to the timestamp of the video.

It would make more sense to return `f32::INFINITY` to always show the last frame of the video, but you should at least add a TODO here to fix this once you start working on the UI for streaming video

crates/utils/re_video/src/stable_index_deque.rs
Wumpf: used too often quite legimately

crates/viewer/re_data_ui/src/video.rs
```diff
@@ -297,23 +320,33 @@ pub fn show_decoded_frame_info(
...
-            );
+            let response = texture.map(|texture| {
+                crate::image::texture_preview_ui(
+                    render_ctx,
+                    ui,
+                    ui_layout,
+                    "video_preview",
+                    re_renderer::renderer::ColormappedTexture::from_unorm_rgba(texture),
+                )
+            });
```
emilk: Do an if-else here instead with `else { ui.allocate_space(egui::Vec2::splat(ui.available_width().at_most(128.0))) }` so you always get a response. Then you can avoid the (very) ugly code below

crates/viewer/re_data_ui/src/video.rs
```diff
@@ -297,23 +320,33 @@ pub fn show_decoded_frame_info(
...
+                    || {
+                        ui.allocate_space(egui::Vec2::splat(ui.available_width().at_most(128.0)))
+                            .1
+                    },
+                    |response| {
+                        egui::Rect::from_center_size(
+                            response.rect.center(),
+                            0.75 * response.rect.size(),
+                        )
+                    },
```
emilk: please revert

crates/viewer/re_renderer/src/video/chunk_decoder.rs
```diff
@@ -137,29 +137,33 @@ impl VideoChunkDecoder {
...
+                frame.content.width(),
+                frame.content.height(),
+            )
+        });
 
         let is_up_to_date = video_texture
             .frame_info
             .as_ref()
-            .is_some_and(|info| info.presentation_time_range() == frame_time_range);
+            // We assume that every frame has a unique presentation timestamp.
```
emilk: I don't understand this. There is only one `frame.info.presentation_timestamp` and it is not a collection… so it's not much of an assumtion?

crates/utils/re_video/src/demux/mod.rs
Wumpf: done in #10126

crates/utils/re_video/src/stable_index_deque.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// Creates a new deque from an iterator and an index offset.
+    ///
+    /// ```
+    /// # use re_video::StableIndexDeque;
+    /// let v = StableIndexDeque::from_iter_with_offset(0..2, 1);
+    /// assert_eq!(v.get(0), None);
+    /// assert_eq!(v.get(1), Some(&0));
+    /// assert_eq!(v.get(2), Some(&1));
+    /// ```
+    pub fn from_iter_with_offset(iter: impl IntoIterator<Item = T>, index_offset: usize) -> Self {
```
emilk: This may be just me, but I feel the offset should come _first_, because it indicates the elements that are missing in the front:

`{2 ellided} + [a, b, c]`  ==  `[-, -, a, b, c]`

crates/utils/re_video/src/stable_index_deque.rs
```diff
@@ -0,0 +1,322 @@
...
+    #[inline]
+    pub fn iter(&self) -> impl Iterator<Item = &T> {
+        self.vec.iter()
+    }
+
+    /// See [`VecDeque::iter_mut`].
+    #[inline]
+    pub fn iter_mut(&mut self) -> impl Iterator<Item = &mut T> {
+        self.vec.iter_mut()
+    }
```
emilk: We should warn off users from doing `.iter().enumerate()` and maybe provide a helper (`.iter_indexed()` or similar) that takes the offset into account

crates/utils/re_video/src/stable_index_deque.rs
```diff
@@ -0,0 +1,322 @@
+use std::collections::VecDeque;
+
+/// A deque with an offset that allows indices into it to stay valid.
```
emilk, line 3:
```suggestion
/// A deque with an offset that allows indices into it to stay valid.
///
/// Maps each integer in a user-defined range to individual elements.
```

crates/utils/re_video/src/stable_index_deque.rs
```diff
@@ -0,0 +1,322 @@
...
+///
+/// // Indices for `StableIndexDeque` are stable:
+/// # use re_video::StableIndexDeque;
+/// let mut v = (0..2).collect::<StableIndexDeque<i32>>();
+/// assert_eq!(v.get(1), Some(&1));
+/// v.pop_front();
+/// assert_eq!(v.get(1), Some(&1));
+/// ```
+#[derive(Default, Clone, Debug)]
+pub struct StableIndexDeque<T> {
```
emilk, line 26: Maybe `RangeDeque` would be a better name?

crates/utils/re_video/src/stable_index_deque.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// let mut v = (0..4).collect::<StableIndexDeque<i32>>();
+    /// v.pop_front();
+    /// v.truncate_to_index(2);
+    /// assert_eq!(v.num_elements(), 1);
+    /// assert_eq!(v.get(0), None);
+    /// assert_eq!(v.get(1), Some(&1));
+    /// assert_eq!(v.get(2), None);
+    /// assert_eq!(v.get(3), None);
+    /// ```
+    pub fn truncate_to_index(&mut self, first_index_not_contained: usize) {
```
emilk: We need this function on both ends of the `StableIndexDeque` (as discussed over voice), so perhaps:

```rs
/// Remove the first N elements.
///
/// N must be <= [`Self::num_elements`].
fn drop_first_n(&mut self, num_elements_to_remove: usize) { … }

/// Remove the last N elements.
///
/// N must be <= [`Self::num_elements`].
fn remove_last_n(&mut self, num_elements_to_remove: usize) { … }
```

crates/utils/re_video/src/stable_index_deque.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// ```
+    /// # use re_video::StableIndexDeque;
+    /// let mut v = (0..5).collect::<StableIndexDeque<i32>>();
+    /// v.pop_front();
+    /// assert_eq!(v.iter_index_range(&(0..5)).cloned().collect::<Vec<_>>(), vec![1, 2, 3, 4]);
+    /// assert_eq!(v.iter_index_range(&(2..4)).cloned().collect::<Vec<_>>(), vec![2, 3]);
+    /// assert_eq!(v.iter_index_range(&(3..5)).cloned().collect::<Vec<_>>(), vec![3, 4]);
+    /// ```
+    #[inline]
+    pub fn iter_index_range(&self, range: &std::ops::Range<usize>) -> impl Iterator<Item = &T> {
```
emilk: I think the name should explain that this is clamped (and thus always safe to call)

crates/utils/re_video/src/stable_index_deque.rs
emilk: yes!

crates/viewer/re_renderer/src/video/mod.rs
```diff
@@ -94,33 +97,40 @@ impl Video {
     ///
     /// Currently supports the following media types:
     /// - `video/mp4`
-    pub fn load(debug_name: String, data: Arc<VideoData>, decode_settings: DecodeSettings) -> Self {
+    pub fn load(
+        debug_name: String,
+        data: VideoDataDescription,
```
emilk:
```suggestion
        video_description: VideoDataDescription,
```

crates/viewer/re_renderer/src/video/mod.rs
```diff
@@ -94,33 +97,40 @@ impl Video {
...
             debug_name,
-            data,
+            video_description: data,
             players,
             decode_settings,
         }
     }
 
-    /// The video data
+    /// The video data.
```
emilk: not really

crates/viewer/re_renderer/src/video/mod.rs
```diff
@@ -94,33 +97,40 @@ impl Video {
...
     #[inline]
-    pub fn data(&self) -> &Arc<re_video::VideoData> {
-        &self.data
+    pub fn data(&self) -> &re_video::VideoDataDescription {
```
emilk:
```suggestion
    pub fn data_descr(&self) -> &re_video::VideoDataDescription {
```

crates/viewer/re_renderer/src/video/mod.rs
```diff
@@ -94,33 +97,40 @@ impl Video {
...
 
-    /// Natural width of the video.
+    /// Mutable access to the video data.
+    ///
+    /// Use with care. It's valid to add samples and groups of pictures, but arbitrary
+    /// changes may interfere with subsequent decoding on existing video streams.
     #[inline]
-    pub fn width(&self) -> u32 {
-        self.data.width()
+    pub fn data_mut(&mut self) -> &mut re_video::VideoDataDescription {
```
emilk:
```suggestion
    pub fn data_descr_mut(&mut self) -> &mut re_video::VideoDataDescription {
```

crates/viewer/re_renderer/src/video/mod.rs
```diff
@@ -137,7 +147,7 @@ impl Video {
         render_context: &RenderContext,
         player_stream_id: VideoPlayerStreamId,
         time_since_video_start_in_secs: f64,
-        video_data: &[u8],
+        video_data: &StableIndexDeque<&[u8]>,
```
emilk:
```suggestion
        video_buffers: &StableIndexDeque<&[u8]>,
```

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -38,20 +38,69 @@ impl std::fmt::Display for ChromaSubsamplingModes {
...
 
-/// Decoded video data.
+/// The basic codec family used to encode the video.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum VideoCodec {
+    Av1,
+    H264,
+    H265,
+    Vp8,
+    Vp9,
```
emilk: Docstrings with their aliases (`AVC` etc)

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -140,63 +209,59 @@ impl VideoData {
...
-        self.samples.len()
+        self.samples.num_elements()
     }
 
     /// Returns the subsampling mode of the video.
     ///
     /// Returns None if not detected or unknown.
     pub fn subsampling_mode(&self) -> Option<ChromaSubsamplingModes> {
-        match &self.config.stsd.contents {
+        match &self.stsd.as_ref()?.contents {
```
emilk, line 283: we could shorten this code (here and below) with a helper function 🤷 

```suggestion
        match &self.stsd()?.contents {
```

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -481,18 +540,22 @@ pub struct Sample {
...
+    /// from which the [`VideoDataDescription`] was loaded.
+    /// For video streams, `data` refers to the currently available data
+    /// which is described by the [`VideoDataDescription`].
     ///
     /// Returns `None` if the sample is out of bounds, which can only happen
     /// if `data` is not the original video data.
-    pub fn get(&self, data: &[u8]) -> Option<Chunk> {
-        let data = data
+    pub fn get(&self, data: &StableIndexDeque<&[u8]>, sample_idx: usize) -> Option<Chunk> {
+        let buffer = *data.get(self.buffer_index)?;
```
emilk:
```suggestion
    pub fn get(&self, buffers: &StableIndexDeque<&[u8]>, sample_idx: usize) -> Option<Chunk> {
        let buffer = *buffers.get(self.buffer_index)?;
```

"data" is an empty name

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -581,16 +620,20 @@ impl std::fmt::Debug for VideoData {
 /// - The index of `needle` in `v`, if it exists
 /// - The index of the first element in `v` that is lesser than `needle`, if it exists
 /// - `None`, if `v` is empty OR `needle` is greater than all elements in `v`
-pub fn latest_at_idx<T, K: Ord>(v: &[T], key: impl Fn(&T) -> K, needle: &K) -> Option<usize> {
+fn latest_at_idx<T, K: Ord>(
+    v: &StableIndexDeque<T>,
+    key: impl Fn(&T) -> K,
+    needle: &K,
+) -> Option<usize> {
```
emilk: Could be a member function of `StableIndexDeque` maybe?

crates/utils/re_video/src/demux/mp4.rs
```diff
@@ -109,9 +98,9 @@ impl VideoData {
         {
             re_tracing::profile_scope!("Sanity-check samples");
             let mut samples_are_in_decode_order = true;
-            for window in samples.windows(2) {
+            for window in samples.iter().tuple_windows::<(&Sample, &Sample)>() {
```
emilk:
```suggestion
            for (a, b) in samples.iter().tuple_windows::<(&Sample, &Sample)>() {
```

crates/utils/re_video/src/demux/mp4.rs
```diff
@@ -109,9 +98,9 @@ impl VideoData {
...
                 samples_are_in_decode_order &=
-                    window[0].decode_timestamp <= window[1].decode_timestamp;
+                    window.0.decode_timestamp <= window.1.decode_timestamp;
```
emilk:
```suggestion
                    a.decode_timestamp <= b.decode_timestamp;
```

crates/utils/re_video/src/demux/mp4.rs
```diff
@@ -63,13 +51,14 @@ impl VideoData {
                 let byte_offset = sample.offset as u32;
                 let byte_length = sample.size as u32;
 
-                samples.push(Sample {
+                samples.push_back(Sample {
                     is_sync: sample.is_sync,
-                    sample_idx,
```
emilk, line 68: Nice that the sample no longer need to know their index

emilk (approved):
Very nice!

crates/viewer/re_renderer/src/video/player.rs
emilk: Here and everywhere

crates/utils/re_video/src/decode/av1.rs
Wumpf: great catch

crates/utils/re_video/src/demux/mod.rs
Wumpf: seems overkill to me :shrug:

crates/utils/re_video/src/stable_index_deque.rs
Wumpf: at no point the user defines a range though? You may construct it from an offset, but there's no ranges involved directly anywhere :thinking:

crates/utils/re_video/src/stable_index_deque.rs
Wumpf: I'm not happy with the current name either. But not gonna apply this since I think that's less descriptive as we're not dealing with ranges anywhere. RangeDeque sounds like a datastructure that allows pushing and popping in ranges and maybe has special handling for gaps in there.

crates/utils/re_video/src/stable_index_deque.rs
Wumpf: yeah that makes sense! :)

crates/utils/re_video/src/stable_index_deque.rs
Wumpf: oh yes this would actually be really useful in a bunch of places! Adding it!

crates/utils/re_video/src/stable_index_deque.rs
Wumpf: why not `remove` on both sides though?

Wumpf: this is not what this method does though - it truncates to a certain _index_. I.e. it makes sure that everything past a given index is no longer in the queue.
But we can work with that. Let's be super explicit and try `remove_all_with_index_larger_equal` and `remove_all_with_index_smaller`

crates/viewer/re_renderer/src/video/chunk_decoder.rs
Wumpf: should say `uniquely identified` :lightbulb:

crates/viewer/re_renderer/src/video/player.rs
Wumpf: agreed, this is _meh_. I'll revisit this when looking into smooth stream playback - setting a reminder

Issue #10030, opened 2025-05-21
Tagged component Viewer-UI/UX implications


gavrelina:
## Context
We introduce tagged components and this shall be reflected throughout UI/UX of our Viewer. 
Before: one entity path could have had a several `Color` components
Now: thanks to tagged components we are now aware of which archetype each of these `Color` components belongs to. Also we have more semantic names instead of "data type" names. 

## Designs
Approach: group by archetypes. Use very similar styling for "marking" archetypes, in thie examples below its always a bit grey and smaller font, as it's more of "an indicator" rather than the data itself. 
[Figma section](https://www.figma.com/design/72AuvfbssFJeqqnQoJscYT/Katya-s-playground?node-id=5723-4487&t=Yt0vSGvMfcjfQlEo-1) **_with more detailed instructions_**

example of spec:
<img width="1019" alt="Image" src="https://github.com/user-attachments/assets/9a8e426c-680e-4ed0-b636-7b18319c0abf" />

### Streams panel
<img src="https://github.com/user-attachments/assets/8f25a5e6-10ca-4ac2-ab22-a7880b33b1eb" width="200">

### Selection panel
<img src="https://github.com/user-attachments/assets/eb15d9f0-c7c6-45d5-9f95-6860ff579fd1" width="200">. <img src="https://github.com/user-attachments/assets/982d8db6-9bf6-413b-a1cf-068f27b734e1" width="200">

### Dataframe view
<img src="https://github.com/user-attachments/assets/014dd81d-e922-4edb-b52a-5557ccfa0bcb" width="800">

**Additional context**
Related to https://github.com/rerun-io/rerun/issues/6889
Related to https://github.com/rerun-io/rerun/issues/8368
Related to https://github.com/rerun-io/rerun/issues/3231

nikolausWest:
Very nice in general. Nit-y question: in the panels you write out "Pinhole Archetype" but in the Dataframe View you just write "Pinhole". Is there any thinking behind the difference? My gut feeling is I don't like exposing the Archetype word so prominently since it isn't really a concept you have to understand until you start going deeper but this is a very blurry line and I haven't thought deeply about this

gavrelina:
[...]

PR #10124, opened 2025-06-03
Group components by archetype in the streams panel

lucasmerlin:
### Related

* part of https://github.com/rerun-io/rerun/issues/10030

### What

Groups components in the streams panel by archetype:

image

### DIFF
--- a/crates/viewer/re_data_ui/src/lib.rs
+++ b/crates/viewer/re_data_ui/src/lib.rs
@@ -5,6 +5,7 @@
 use re_log_types::EntityPath;
 use re_types::{ComponentDescriptor, RowId};
 use re_viewer_context::{UiLayout, ViewerContext};
+use std::collections::BTreeMap;
 
 mod annotation_context;
 mod app_id;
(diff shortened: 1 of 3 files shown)


crates/viewer/re_data_ui/src/lib.rs
```diff
@@ -41,6 +43,30 @@ pub fn sorted_component_list_for_ui<'a>(
...
+        .fold(BTreeMap::default(), |mut acc, descriptor| {
+            acc.entry(descriptor.archetype_name.clone())
+                .or_insert_with(Vec::new)
+                .push(descriptor.clone());
+            acc
+        });
+
+    for (_, components) in map.iter_mut() {
+        // Sort by the short name, as that is what is shown in the UI.
+        components.sort_by_key(|c| c.display_name());
```
lucasmerlin, line 64: @Wumpf you said these should instead be sorted by importance, how would I do that?

Wumpf: global context / viewer context hold on to `re_types_core::reflection::Reflection` which has the information you're looking for. Might not be entirely straight forward to pipe this through - happy to solve this in pair programming this afternoon

gavrelina:
nice! 1 nit: it shall not have a hover state, as it's more like a header rather than smth to click on

lucasmerlin:
> nice! 1 nit: it shall not have a hover state, as it's more like a header rather than smth to click on

Fixed!

PR #10125, opened 2025-06-03
Prepare video visualizer for upcoming stream visualizer 

Wumpf:
### Related

* https://github.com/rerun-io/rerun/issues/7484
* Follow-up to https://github.com/rerun-io/rerun/pull/10123
* Followed by https://github.com/rerun-io/rerun/pull/10126

### What

* rename `VideoCache` to `VideoAssetCache`
* move out utilities from video frame reference visualizer to be used in video stream
* handle pickable rect bounding boxes more uniformly

video visualizer looks like a rewrite but it's mostly moving out + rename


crates/viewer/re_view_spatial/src/visualizers/utilities/spatial_view_visualizer.rs
```diff
@@ -62,3 +92,21 @@ impl SpatialViewVisualizerData {
         self
     }
 }
+
+fn bounding_box_for_textured_rect(
+    textured_rect: &re_renderer::renderer::TexturedRect,
+) -> macaw::BoundingBox {
```
emilk: This could also be a member-method on `TexturedRect` 🤷 
Makes it easier to find

Issue #8748, opened 2025-01-21
Add versioning info in the arrow metadata


emilk:
We want to version either component (columns), or chunks (record batches), or both. This is the first step in providing backwards compatibility for our data.

PR #10126, opened 2025-06-03
New `VideoStream` archetype for loose video samples

Wumpf:
### Related

* https://github.com/rerun-io/rerun/issues/7484
* Follow-up to https://github.com/rerun-io/rerun/pull/10125

### What

Introduces & implements a new `VideoStream` archetype.
Like `VideoFrameReference` this is backed by a specialized cache. However, the `VideoStreamCache` handles chunk addition and removal (!) for all its monitored video streams. To feed the video player properly it also features on-the-fly GOP detection!

This marks the first piece of towards full streaming support. Notably still missing:
* better unit testing for all of this. We should be able to have some unit tests for the cache and some image comparision tests for video in general
* examples & snippets - this PR comes with "tests" that aren't fully formulated examples yet
* wider codec support
* web player support
* smooth streaming (among other things video player isn't aware of gops growing)
* ... and more.

To run streaming example:
```
pixi run py-build
pixi run -e py python ./tests/python/video_ffmpeg/main.py
```

To limit memory in order to test GC behavior try:
```
pixi run rerun --memory-limit 200MB
pixi run -e py python ./tests/python/video_ffmpeg/main.py
```

### DIFF
--- a/crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_narrow/VideoSample_placeholder.png
+++ b/crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_narrow/VideoSample_placeholder.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:8da24feca371287abb82723e43a3a3bf628d94af5689824105afd567cd6010d6
-size 2841
+oid sha256:b052a99437a77e4c3ddb6613b3d944fe90e872b50b2be52d0c1a7041ee926b35
+size 2843
--- a/crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_wide/VideoSample_placeholder.png
+++ b/crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_wide/VideoSample_placeholder.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:b0091979279da95744473118cdd7aa472a67b8906dbe49ccb107f77d23de4993
-size 3156
+oid sha256:e2dfb7e6848faae48ec178407bcc37524bfc98a03ce4c5b6a44a45616db27f83
+size 3158


crates/build/re_types_builder/src/codegen/rust/reflection.rs
```diff
@@ -126,11 +126,16 @@ fn generate_component_reflection(
...
+        let is_enum_with_default = obj.is_enum()
+            && obj
+                .fields
+                .iter()
+                .any(|field| field.attrs.has(crate::ATTR_DEFAULT));
+        let has_default_attr = obj
+            .try_get_attr::<String>(ATTR_RUST_DERIVE_ONLY)
+            .or_else(|| obj.try_get_attr::<String>(ATTR_RUST_DERIVE))
+            .is_some_and(|derives| derives.contains("Default"));
+        let auto_derive_default = is_enum_with_default || has_default_attr;
```
emilk, line 138: I think explicit numbering of enum variants could be considered a feature. You really don't want to change the numbering, because it will break backwards compatibility

crates/store/re_types/definitions/rerun/components/video_codec.fbs
```diff
@@ -0,0 +1,65 @@
...
+//     /// Enum value is the fourcc for 'hev1' (the WebCodec string assigned to this) in big endian.
+//     H265 = 0x68657631, // b'hev1'.hex()
+//     /// Enum value is the fourcc for 'av01' (the WebCodec string assigned to this) in big endian.
+//     AV1 = 0x61763031, // b'av01'.hex()
+//     /// Enum value for 'vp80' which is the most common fourcc for VP8.
+//     /// Unlike with other codecs, WebCodec standard assigns only 3 letters ('vp8') to VP8.
+//     VP8 = 0x76703830, // b'vp80'.hex()
+//     /// Enum value is the fourcc for 'vp09' (the WebCodec string assigned to this) in big endian.
+//     VP9 = 0x76703039, // b'vp09'.hex()
+// }
```
emilk, line 68: We should consider doing this right now to avoid having to migrate this later.

Not needed if you mark this component with `"attr.rerun.state": "unstable"` and add a TODO

crates/store/re_types/definitions/rerun/components/video_codec.fbs
```diff
@@ -0,0 +1,65 @@
+namespace rerun.components;
+
+/// The codec used to encode video stored in [components.VideoSample].
+///
+/// Support of these codecs by the Rerun Viewer is platform dependent.
+/// For more details see check the [video reference](https://rerun.io/docs/reference/video).
+enum VideoCodec: ubyte ("attr.docs.unreleased") {
```
emilk:
```suggestion
enum VideoCodec: ubyte ("attr.docs.unreleased", "attr.rerun.state": "unstable") {
```

crates/store/re_types/definitions/rerun/archetypes/video_stream.fbs
```diff
@@ -0,0 +1,53 @@
...
+///
+/// All components except `sample` are typically logged statically once per entity.
+/// `sample` is then logged repeatedly for each frame on the timeline.
+///
+/// TODO(#7484): Add snippet.
+table VideoStream (
+  "attr.docs.unreleased",
+  "attr.docs.category": "Video",
+  "attr.docs.view_types": "Spatial2DView, Spatial3DView: if logged under a projection",
+  "attr.rerun.state": "stable"
```
emilk: Do you really want people using this right away? 😬 
I think it should at least have a snippet and an example before it is marked stable.

```suggestion
  "attr.rerun.state": "unstable"
```

crates/store/re_types/definitions/rerun/archetypes/video_stream.fbs
```diff
@@ -0,0 +1,53 @@
...
+    /// Each video sample must contain enough data for exactly one video frame
+    /// (this restriction may be relaxed in the future for some codecs).
+    ///
+    /// See [components.VideoCodec] for codec specific requirements.
+    sample: rerun.components.VideoSample ("attr.rerun.component_required", required, order: 1000);
+
+    /// The codec used to encode the video chunks.
+    ///
+    /// This property is expected to be constant over time and is ideally logged statically once per stream.
+    codec: rerun.components.VideoCodec ("attr.rerun.component_required", order: 1100);
```
emilk, line 42: I still don't understand why have have all of

* `"attr.rerun.component_required"` vs `"attr.rerun.component_optional"`
* `required` vs `nullable` vs ``

Does all six combinations of these produce six different things? O.o

crates/store/re_types/definitions/rerun/archetypes/video_stream.fbs
```diff
@@ -0,0 +1,53 @@
...
+    ///
+    /// The current timestamp is used as presentation timestamp (PTS) for all data in this sample.
+    /// There is currently no way to log differing decoding timestamps, meaning
+    /// that there is no support for B-frames.
+    /// See <https://github.com/rerun-io/rerun/issues/10090> for more details.
+    // TODO(#10090): See above.
+    ///
+    /// Unlike any other data in Rerun, video samples are not allowed to be logged out of order,
+    /// as this may break live video playback.
+    /// I.e. any appended sample should have a timestamp greater than all previously logged samples.
```
emilk, line 30: Does this really mean you need to log them in order? They will be sorted by the chunk-store anyway? Is the problem the gaps out-of-order can produce?

crates/utils/re_video/src/decode/gop_detection.rs
```diff
@@ -0,0 +1,56 @@
...
+        crate::VideoCodec::H265 => Err(StartOfGopDetectionFailure::UnsupportedCodec(codec)),
+        crate::VideoCodec::Vp8 => Err(StartOfGopDetectionFailure::UnsupportedCodec(codec)),
+        crate::VideoCodec::Vp9 => Err(StartOfGopDetectionFailure::UnsupportedCodec(codec)),
+    }
+}
+
+/// Try to determine whether a frame chunk is the start of a closed GOP.
+///
+/// Expects Annex B encoded frame.
+fn is_annexb_sample_start_of_gop(sample_data: &[u8]) -> bool {
```
emilk, line 33: Remind me - is start_of_gop same as sync=keyframe? or only sometimes?

crates/utils/re_video/src/decode/nalu.rs
```diff
@@ -0,0 +1,187 @@
+/// In Annex-B before every NAL unit is a nal start code.
```
emilk:
```suggestion
/// In Annex-B before every NAL unit is a NAL start code.
```

crates/utils/re_video/src/decode/nalu.rs
```diff
@@ -0,0 +1,187 @@
...
+struct AnnexBSeparatorPos {
+    separator_pos: usize,
+    separator_len: usize,
+}
+
+fn annex_b_next_nal_unit_start_pos(sample_data: &[u8]) -> Option<AnnexBSeparatorPos> {
+    let mut pos = 0;
+
+    while pos + NAL_START_CODE_SHORT.len() <= sample_data.len() {
+        if &sample_data[pos..pos + NAL_START_CODE_SHORT.len()] == NAL_START_CODE_SHORT {
```
emilk: This would be simpler with `.strip_prefix` or `.starts_with`

crates/utils/re_video/src/decode/nalu.rs
```diff
@@ -0,0 +1,187 @@
...
+            return Some(AnnexBSeparatorPos {
+                separator_pos: pos,
+                separator_len: NAL_START_CODE_SHORT.len(),
+            });
+        }
+        if pos + NAL_START_CODE.len() < sample_data.len()
+            && &sample_data[pos..pos + NAL_START_CODE.len()] == NAL_START_CODE
```
emilk: This too

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+            &mut video_sample_buffers,
+            &mut samples,
+            &mut gops,
+        )?;
+    }
+
+    Ok((
+        re_video::VideoDataDescription {
+            codec,
+            stsd: None,                               // Only on mp4s have this.
```
emilk:
```suggestion
            stsd: None,                               // Only mp4s have this.
```

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+            samples_statistics: re_video::SamplesStatistics::NO_BFRAMES, // TODO(#10090): No b-frames for now.
+            mp4_tracks: std::iter::once((0, Some(re_video::TrackKind::Video))).collect(),
+        },
+        video_sample_buffers,
+    ))
+}
+
+/// Reads all samples from a chunk and appends them to a list.
+///
+/// Rejects out of order samples - new samples must have a higher timestamp than the previous ones.
```
emilk, line 244: How can they be out-of-order? We read them chronologically from the store, no?

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+    let values = inner_list_array
+        .values()
+        .downcast_array_ref::<arrow::array::PrimitiveArray<arrow::array::types::UInt8Type>>()
+        .ok_or(VideoStreamProcessingError::InvalidVideoSampleType(
+            raw_array.data_type().clone(),
+        ))?;
+    let values = values.values().inner();
+
+    let offsets = inner_list_array.offsets();
+    let lengths = re_arrow_util::offsets_lengths(inner_list_array.offsets()).collect::<Vec<_>>();
```
emilk: Looks like this have been added to arrow: https://docs.rs/arrow/latest/arrow/buffer/struct.OffsetBuffer.html#method.lengths

crates/utils/re_video/src/time.rs
```diff
@@ -3,6 +3,8 @@
 pub struct Timescale(u64);
 
 impl Timescale {
+    pub const IDENTITY: Self = Self(1);
```
emilk: The docs says `Timescale` represent number of time units per second. Having one per second seems extremely coarse.

Looking at later code, it looks like you are actually using nanoseconds as `Time`, which would mean you want this:

```suggestion
    /// Each [`Time`] tick is one nanosecond.
    pub const NANOSECOND: Self = Self(1_000_000_000);
```

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+            &mut gops,
+        )?;
+    }
+
+    Ok((
+        re_video::VideoDataDescription {
+            codec,
+            stsd: None,                               // Only on mp4s have this.
+            coded_dimensions: None,                   // Unknown so far.
+            timescale: re_video::Timescale::IDENTITY, // We use raw rerun timestamps, so we don't have to scale time.
```
emilk: I'm surprised this is not causing massive bugs, at least with the UI. See my previous comments about the definition of `Timescale`.

We use `Timescale` to map `re_video::Time` values into actual time units (e.g. seconds). For temporal timelines we can do so by using the timescale 1e9, and everything is fine.

For sequential timelines we're not so lucky. I see two common cases: `log_tick` and `frame_idx`. In the first case there is no relationship between the frame number and actual durations. And the second case - well, we don't know if we are in the second case. And even if we did, we don't know the FPS of the video here (I think?).

```suggestion
            timescale: match timeline.typ() {
                TimeType::Sequence => None, // Can't translate the sequence time to real durations
                TimeType::DurationNs | TimeType::TimestampNs => Some(Timescale::NANOSECONDS),
            }
```

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+        chunk
+            .iter_component_offsets(frame_chunk_descr)
+            .zip(chunk.iter_component_indices(&timeline, frame_chunk_descr))
+            .filter_map(move |((idx, len), (time, _row_id))| {
+                debug_assert_eq!(len, 1, "Expected only a single video sample per timestep");
+
+                let sample_idx = sample_base_idx + idx;
+                let byte_offset = &offsets[idx..idx + len];
+                let byte_length = &lengths[idx..idx + len];
+                let decode_timestamp = re_video::Time(time.as_i64());
```
emilk, line 342: This is worth a comment:

For temporal timelines we assign the nanosecond to the `Time`, and use a `Timescale` of `1e9`. Great.
For sequential timelines we assign the sequence number to the time (likely the camera frame number, or maybe `log_tick`).

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+
+                // Make sure samples are in order.
+                // Within a chunk we know that, but we want to check whether the *previous* samples
+                // (from prior calls to this method) are also in order with the new one.
+                if previous_max_presentation_timestamp > decode_timestamp {
+                    re_log::warn_once!("Out of order insertion into video streams is not supported. Ignoring any out of order samples.");
+                    return None;
```
emilk: Seems like this should be a `debug_assert` since it is looking for bugs in our code, not in the user data (see previous comment)

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+
+                // Only a single sample/blob is expected per timestamp.
+                // It should not be possible to log more than one.
+                debug_assert_eq!(byte_offset.len(), 1);
+                debug_assert_eq!(byte_length.len(), 1);
+                let byte_offset = byte_offset[0] as u32;
+                let byte_length = byte_length[0] as u32;
+
+                let is_sync_result = re_video::is_sample_start_of_gop(
+                    &values[byte_offset as usize..(byte_offset + byte_length) as usize],
```
emilk:
```suggestion
                let byte_offset = byte_offset[0];
                let byte_length = byte_length[0];

                let is_sync_result = re_video::is_sample_start_of_gop(
                    &values[byte_offset..(byte_offset + byte_length)],
```

seems more readable 🤷

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+    let mut previous_max_presentation_timestamp = samples
+        .back()
+        .map_or(re_video::Time::MIN, |s| s.presentation_timestamp);
+
+    // Extract sample metadata.
+    samples.extend(
+        chunk
+            .iter_component_offsets(frame_chunk_descr)
+            .zip(chunk.iter_component_indices(&timeline, frame_chunk_descr))
+            .filter_map(move |((idx, len), (time, _row_id))| {
```
emilk, line 324: I think it's about time we define a `struct PositiveRange<Idx> { pub start: Idx, len: Idx }` somewhere instead of using two struct fields in some places (`Sample::byte_offset/byte_length`) and naked tuples in other (`iter_component_offsets`), both with their own drawbacks

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+                debug_assert_eq!(len, 1, "Expected only a single video sample per timestep");
```
emilk:
```suggestion
                debug_assert_eq!(len, 1, "Expected only a single VideoSample per row (it is a mono-component)");
```

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+
+    let mut previous_max_presentation_timestamp = samples
+        .back()
+        .map_or(re_video::Time::MIN, |s| s.presentation_timestamp);
+
+    // Extract sample metadata.
+    samples.extend(
+        chunk
+            .iter_component_offsets(frame_chunk_descr)
+            .zip(chunk.iter_component_indices(&timeline, frame_chunk_descr))
```
emilk: style: I find `zip!` to be strictly more readable, but YMMV

```suggestion
        zip!(
            chunk.iter_component_offsets(frame_chunk_descr),
            chunk.iter_component_indices(&timeline, frame_chunk_descr))
```

emilk (commented):
I have reviewed everything except the (very complicated) `video_stream_cache.rs`

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+        // As of writing, in a debug wasm build with Chrome loading a 600MiB 1h video
+        // this assumption holds up fine: There is a (sufferable) delay,
+        // but it's almost entirely due to the decoder trying to retrieve a frame.
+    }
+
+    /// Keep existing cache entries up to date with new and removed video data.
+    fn on_store_events(&mut self, events: &[ChunkStoreEvent]) {
+        re_tracing::profile_function!();
+
+        let frame_chunk_descr = VideoStream::descriptor_sample();
```
emilk: `sample_descr` is probably a better name

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+    (
+        re_video::VideoDataDescription,
+        StableIndexDeque<SampleBuffer>,
+    ),
+    VideoStreamProcessingError,
+> {
+    re_tracing::profile_function!();
+
+    let frame_sample_descr = VideoStream::descriptor_sample();
+    let codec_chunk_descr = VideoStream::descriptor_codec();
```
emilk:
```suggestion
    let sample_descr = VideoStream::descriptor_sample();
    let codec_descr = VideoStream::descriptor_codec();
```

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+    // TODO(andreas): Can we be more clever about the chunk range here and build up only what we need?
+    // Kinda tricky since we need to know how far back (and ahead for b-frames) we have to look.
+    let entire_timeline_query =
+        re_chunk::RangeQuery::new(timeline, re_log_types::ResolvedTimeRange::EVERYTHING);
+    let query_results = store.storage_engine().cache().range(
+        &entire_timeline_query,
+        entity_path,
+        &[frame_sample_descr.clone(), codec_chunk_descr.clone()],
+    );
+    let video_chunks = query_results
```
emilk:
```suggestion
    let sample_chunks = query_results
```

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+        video_sample_buffers,
+    ))
+}
+
+/// Reads all samples from a chunk and appends them to a list.
+///
+/// Rejects out of order samples - new samples must have a higher timestamp than the previous ones.
+fn read_samples_from_chunk(
+    timeline: TimelineName,
+    frame_chunk_descr: &re_types::ComponentDescriptor,
```
emilk: Do we really need to pass this by argument? It's just `VideoStream::descriptor_sample()` anyways?

emilk: also: name

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+                    return None;
+                }
+                previous_max_presentation_timestamp = decode_timestamp;
+
+                // Only a single sample/blob is expected per timestamp.
+                // It should not be possible to log more than one.
+                debug_assert_eq!(byte_offset.len(), 1);
+                debug_assert_eq!(byte_length.len(), 1);
+                let byte_offset = byte_offset[0] as u32;
+                let byte_length = byte_length[0] as u32;
```
emilk: This will panic if anyone logs an empty list of samples for some timestamp

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+    gops: &mut StableIndexDeque<re_video::demux::GroupOfPictures>,
+) -> Result<(), VideoStreamProcessingError> {
+    re_tracing::profile_function!();
+
+    let Some(raw_array) = chunk.raw_component_array(frame_chunk_descr) else {
+        // This chunk doesn't have any video chunks.
+        return Ok(());
+    };
+
+    let inner_list_array = raw_array
```
emilk, line 300: This function badly need a long explanation and/or a diagram. Here's a start:

The `raw_array` above is logically a `Vec<Vec<Blob>>`, where the inner `Vec` always has a len=1, because we're dealing with a "mono-component" (each `VideoStream` has exactly one `VideoSample` instance per time)`.

Because of how arrow works, the bytes of all the blobs are actually sequential in memory (yay!) in a single buffer, what you call `values` below (could use a better name btw).

We want to figure out the byte offsets of each blob within the arrow buffer that holds all the blobs, i.e. get out a `Vec<ByteRange>`.

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
emilk: It would be much better to error here if `len != 1`. Right now the code may crash if this invariant isn't held, even in release builds.

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+    /// it creates all the necessary metadata.
+    /// For any stream in the cache, metadata will be kept automatically up to date for incoming
+    /// and removed video frames chunks.
+    pub fn entry(
+        &mut self,
+        store: &re_entity_db::EntityDb,
+        entity_path: &EntityPath,
+        timeline: TimelineName,
+        decode_settings: DecodeSettings,
+    ) -> Result<Arc<RwLock<PlayableStoreVideoStream>>, VideoStreamProcessingError> {
```
emilk: maybe something like this would be helpful

```rs
type SharablePlayableStoreVideoStream = Arc<RwLock<PlayableStoreVideoStream>>;
```

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,555 @@
...
+    sample_index_range: std::ops::Range<re_video::SampleIndex>,
+}
+
+/// Video stream from the store, ready for playback.
+///
+/// This is compromised of:
+/// * raw video stream data (pointers into all live rerun-chunks holding video frame data)
+/// * metadata with that we know about the stream (where are I-frames etc.)
+/// * active players for this stream and their state
+pub struct PlayableStoreVideoStream {
```
emilk: I find the "Store" part of this name more confusing than useful

crates/build/re_types_builder/src/codegen/rust/reflection.rs
Wumpf: not following - did I change the numbering of enums? What I'm doing here is that not all enums in Rust have to have a default value, thus Rerun type reflection doesn't always provide a default method for them now

crates/store/re_types/definitions/rerun/archetypes/video_stream.fbs
Wumpf: makes sense, but what are we using as criterium? I figured this is about data stability and I'm somewhat confident that this is a format we'll support :shrug:
but yeah, seems premature!

crates/store/re_types/definitions/rerun/archetypes/video_stream.fbs
Wumpf: tbh don't know either, would need to dig it up
can't do `requried` here it tells me because fbs doesn't support this on this kind of field. But the `component_required` does what I want

crates/store/re_types/definitions/rerun/archetypes/video_stream.fbs
Wumpf: They are sorted within individual chunks.
But if you don't log them in order we may also get chunks with samples out of order and _that_ is what we can't handle in the `VideoStreamCache` - we're not able to handle insertions there (yet?). Handling insertions is a bit of a headache because it means that we can no longer use a deque for sample-metadata/gops/sample-buffers and it makes the video player logic a lot more complicated since it may get told at any point that the samples it just enqueued into the decoder were the wrong ones for a given timestamp that _looked_ like it was sandwiched between two samples.

crates/store/re_types/definitions/rerun/components/video_codec.fbs
Wumpf: you're the expert on enums, do you have an intuition on how hard could this turn out to be? Sounds like it may break assumptions in a whole lot of places?

Wumpf: if we had version numbers on chunks this would be trivial to migrate at any point :thinking:

crates/utils/re_video/src/decode/gop_detection.rs
Wumpf: correct. `start_of_gop` is always `is_sync` throughout our code

```
pub struct SampleMetadata {
    /// Is this the start of a new (closed) [`GroupOfPictures`]?
    ///
    /// What this means in detail is dependent on the codec but they are generally
    /// at least I(DR)-frames and often have additional metadata such that
    /// a decoder can restart at this frame.
    pub is_sync: bool,
```

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: Guess we didn't notice any issues with this so far simply because I'm using raw time ints consistently and there's no ui around those new samples yet (might not show up anywhere either way? List of samples as we have on video asset doesn't make sense)

Yes, we don't know FPS. We could probably know it in some cases with a lot of effort, but modern codecs (depending on config) don't have stable framerates so the point is moot anyways.

`None` as a timescale just causes issues elsewhere. Let's just use `IDENTITY` there then and be mindful about this when we end up showing values from `SampleMetadat` :/

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: Chunks may be inserted between two existing chunks. will clarify comment
Also, if I recall correctly, data within chunks is ordered. But there's no guarantee that consecutive chunks don't overlap, right?

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: find it always overkill to introduce type aliases that are used only once. but sure :shrug:

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: Guess I've been working with this stuff so long that I'm blind to these kind of complications, it just feeds in a chunk to the datastructures that are described elsewhere already :shrug:
Writing the stuff you mentioned but no idea how to draw a diagram for this

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: nice, but we don't have it yet. And I'm not going to update arrow now ;-)

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: we could actually also zip in offsets and lengths here

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: agree, but don't want to do that in this PR

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: As per pervious comments that's not accurate: we may receive out of order chunks and those are seen as user errors as documented. But what we should do is to do this as a separate check

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: nice try, but byte_offset is an i32

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: don't like it because `izip!` is not cargo fmt friendly

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: nice catch

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
emilk: Right - data within a single chunk is not necessarily sorted

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
emilk: Mindful how? To never use it when it is IDENTITY? Sounds like the job of an `Option` 😬

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
emilk: Yup, my bad

crates/store/re_types/definitions/rerun/archetypes/video_stream.fbs
emilk: Yeah, my bad

crates/store/re_types/definitions/rerun/components/video_codec.fbs
emilk: * Yup: https://github.com/rerun-io/rerun/issues/8748

crates/store/re_types/definitions/rerun/components/video_codec.fbs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// Advanced Video Coding (AVC/H.264)
+    ///
+    /// See <https://en.wikipedia.org/wiki/Advanced_Video_Coding>
+    ///
+    /// [components.VideoSample]s using this codec should be formatted according to Annex B specification.
+    /// (Note that this is different from AVCC format found in MP4 files.
+    /// To learn more about Annex B, check for instance <https://membrane.stream/learn/h264/3>)
+    /// Key frames (IDR) require inclusion of a SPS (Sequence Parameter Set)
+    H264,
```
emilk: This is what I think should be explicit (e.g. `=1`)

crates/build/re_types_builder/src/codegen/rust/reflection.rs
emilk: I saw an enum variant in your codec without a value assigned to it, which I thought was forbidden, and so I thought this code change was to blame for allowing it… but I see now I was wrong 🤦

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
emilk: For sure

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,587 @@
...
+        },
+        video_sample_buffers,
+    ))
+}
+
+/// Reads all video samples from a chunk into an existing video description.
+///
+/// Rejects out of order samples - new samples must have a higher timestamp than the previous ones.
+/// Since samples within a chunk are guaranteed to be ordered, this can only happen if a new chunk
+/// is inserted that has is timestamped to be older than the data in the last added chunk.
```
emilk, line 246: I tink this is wrong. Data within a chunk is sorted by `RowId` which may not be aligned with whatever timeline we are on. In other words: data within a rerun chunk is NOT guaranteed to be sorted. Furthermore, chunks can cover have overlapping time ranges.

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,587 @@
...
+                let sample_idx = sample_base_idx + idx;
+                let byte_offset = offsets[idx] as usize;
+                let byte_length = lengths[idx];
+
+                // Note that the conversion of this time value is already handled by `VideoDataDescription::timescale`:
+                // For sequence time we use a scale of 1, for nanoseconds time we use a scale of 1_000_000_000.
+                let decode_timestamp = re_video::Time(time.as_i64());
+
+                // Samples within a chunk are expected to be always in order.
+                debug_assert!(decode_timestamp > previous_max_presentation_timestamp);
```
emilk: I think this is a bad assumption. As noted above, chunks are sorted by `RowId`, but cannot be sorted on all timelines. We should have a test that actively log a video backwards and in a random order so make sure we don't crash in those cases

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,587 @@
...
+    }
+
+    // Fill out durations for all new samples plus the first existing sample for which we didn't know the duration yet.
+    // (We set the duration for the last sample to `None` since we don't know how long it will last.)
+    // (Note that we can't use tuple_windows here because it can't handle mutable references)
+    for sample in sample_base_idx.saturating_sub(1)..samples.next_index() - 1 {
+        samples[sample].duration = Some(
+            samples[sample + 1].presentation_timestamp - samples[sample].presentation_timestamp,
+        );
+    }
```
emilk: This looks like it would crash if `sample_base_idx - 1` has been dropped from `samples`

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,587 @@
...
+
+    // Sanity checks on chunk buffers.
+    if let Some(last_buffer) = chunk_buffers.back() {
+        debug_assert!(
+            last_buffer.sample_index_range.end == sample_base_idx,
+            "Sample range of each chunk buffer must be non-overlapping and without gaps. \
+                (last_buffer.sample_range.end: {}, sample_base_idx: {})",
+            last_buffer.sample_index_range.end,
+            sample_base_idx
```
emilk:
```suggestion
        debug_assert_eq!(
            last_buffer.sample_index_range.end, sample_base_idx,
            "Sample range of each chunk buffer must be non-overlapping and without gaps."
```

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -0,0 +1,587 @@
...
+}
+
+/// Adjust GOPs for removed samples at the back of the sample list.
+fn adjust_gops_for_removed_samples_back(video_data: &mut re_video::VideoDataDescription) {
+    let max_sample_index = video_data.samples.next_index();
+    while let Some(gop) = video_data.gops.back_mut() {
+        if gop.sample_range.start >= max_sample_index {
+            video_data.gops.pop_back();
+        } else {
+            gop.sample_range.end = max_sample_index;
```
emilk: nit: "max" is inclusive ("the maximum value")

```suggestion
    let end_sample_index = video_data.samples.next_index();
    while let Some(gop) = video_data.gops.back_mut() {
        if gop.sample_range.start >= end_sample_index {
            video_data.gops.pop_back();
        } else {
            gop.sample_range.end = end_sample_index;
```

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: it's such a weird corner case T_T
I can explore the option, but it's just gonna end up applying scale 1 in lots of places. I think our main interaction point with time conversion other than ui is when we ask the player to give us the frame for a given time which is expressed in seconds right now. So technically that's also wrong 🤔

crates/store/re_types/definitions/rerun/archetypes/video_stream.fbs
emilk: This is the docs we have:
https://github.com/rerun-io/rerun/blob/804028ca16cbf4cfcd1d5c6e1705f9b926b1639f/crates/store/re_types/definitions/rerun/attributes.fbs#L53-L58

So yeah, this goes under "experimental".

We should strive to always have `main` in a releasable state, so the question to ask is "Is this ready to be released as part of our public APIs?"

crates/store/re_types/definitions/rerun/components/video_codec.fbs
Wumpf: ah thx, sure!

emilk (approved):
Let's mark `VideoStream` as `"attr.rerun.state": "unstable"`, then LGTM!

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: but we are iterating on them in order with those iterators above? 🤔
sounds like we should huddle about this, but going to be hard to squeeze that in. I can at the _very least_ adjust those comments to not lie about chunks!

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
emilk: The whole point of `TimeScale` is to convert between real durations and `Time` values, so ALL uses of it WILL be wrong (except in cases where we both multiply and divide with it, which cancels out - but we should probably find those places and not do that, since it looses precision anyways)

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: We talked about this again on a huddle briefly and concluded that we have to look this up and document accordingly.

Now eventually looking at it the answer is clear: We want `chunk.sorted_by_timeline_if_unsorted` to be sure! :lightbulb:

PR #10128, opened 2025-06-03
Add support for remote dataset blueprints

abey79:
### Related

* Part of https://github.com/rerun-io/dataplatform/issues/777
* Sibling TODO

### What

With this PR, the viewer now checks if a dataset blueprint exists on the server, and, if so, it (always) downloads it before downloading a partition. The downloaded blueprint is set as default.

This PR also includes the proto updates necessary for both client and server implementation of this feature.

Note: there is currently no way to _set_ a partition blueprint. This must be done by manually uploading a `.rbl` with the CLI.


crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -223,56 +227,136 @@ pub fn get_chunks_response_to_chunk_and_partition_id(
...
         .into_inner()
         .try_into()?;
 
-    let dataset_name = read_dataset_response.dataset_entry.details.name;
+    let blueprint_dataset = response.dataset_entry.blueprint_dataset;
+
+    if let Some(blueprint_dataset) = blueprint_dataset {
+        re_log::debug!("Streaming blueprint dataset {blueprint_dataset}");
+
+        let blueprint_store_id = StoreId::random(StoreKind::Blueprint);
```
jleibs: Why can't this use the partition_id from the blueprint?  In the future we should be looking that up based on a query, but even now, we know it to be fixed? A comment here for posterity would at least help keep track of this.

crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -223,56 +227,136 @@ pub fn get_chunks_response_to_chunk_and_partition_id(
...
+            store_source: StoreSource::Unknown,
+            store_version: None,
+        };
+
+        stream_partition_from_server(
+            &mut client,
+            blueprint_store_info,
+            &tx,
+            blueprint_dataset,
+            None,
```
jleibs: I'm similarly surprised that we're requesting the whole dataset and not a single partition.

crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -223,56 +227,136 @@ pub fn get_chunks_response_to_chunk_and_partition_id(
...
+            &on_cmd,
+            on_msg.as_deref(),
+        )
+        .await?;
+
+        if tx
+            .send(LogMsg::BlueprintActivationCommand(
+                BlueprintActivationCommand {
+                    blueprint_id: blueprint_store_id,
+                    make_active: false,
```
jleibs: Years later, the question of whether users expect this to be `true` or `false` continues to be elusive :disappointed:

jleibs (approved):

crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -223,56 +227,136 @@ pub fn get_chunks_response_to_chunk_and_partition_id(
...
+
+        let blueprint_store_info = StoreInfo {
+            application_id: uri.dataset_id.to_string().into(),
+            store_id: blueprint_store_id.clone(),
+            cloned_from: None,
+            store_source: StoreSource::Unknown,
+            store_version: None,
+        };
+
+        stream_partition_from_server(
```
jleibs: My read of the server code has me realizing that it will actually be fairly common for the dataset to exist but to have no partitions. In that case, we are still going to stream, and then activate, an empty blueprint, which I don't believe we want. We likely need to add a check in here that there exists at least 1 partition.

crates/store/re_grpc_client/src/redap/mod.rs
abey79: The partition ID is not unique. As a matter of fact, it could always be e.g. `__default`. So we'd get store ID clashes.

Now that I think of it, we have this problem with the recording store ID too. I should probably randomise it as well.

crates/store/re_grpc_client/src/redap/mod.rs
abey79: :lolcry:

crates/store/re_grpc_client/src/redap/mod.rs
abey79: good catch!

crates/store/re_grpc_client/src/redap/mod.rs
abey79: Yeah, we don't currently have the ability to override the partition id upon registration, so I can make no assumption on the partition id for now. So we can:
1) download everything and hope there is only one blueprint in there (current implementation, unblocks sandbox)
2) add the ability to override partition id to `redap register ...` and use a known partition id such as `__default` (my preferred way, eventually)
3) list partition id and pick e.g the first

abey79: For now I'm going with (1) and leave a comment + open an issue.

abey79: https://github.com/rerun-io/dataplatform/issues/858

PR #10132, opened 2025-06-03
WIP implementation of split query

jleibs:
### What

Rather than using the inline query -> chunk API, fetch the list of chunk-ids first, and then fetch the chunks.

This surfaces a fairly significant design problem with the chunk-id-based get_chunks API.


Wumpf:
@jleibs I reckon this experiment is no longer applicable. Can we close?

Wumpf:
I take this as a "yes plz close". Especially given that by now we have split fetch+get implemented on `main` and there's no more `get_chunks` in the api

PR #10134, opened 2025-06-04
Fix errors in getting-started/data-in documentation

wyfo:
Both Rust and CPP didn't compile because of typos/out-of-date API.

`set_duration_secs`/`set_time_duration` were also missing in the structure animation loop.

Last but not least, Rust dependencies have been updated.


abey79 (approved):
Thanks. (We really need to turn these into snippets, so they are properly run/tested by CI.)

Issue #2628, opened 2023-07-07
Python SDK incompatible with old GLIBC


emilk:
# In short
The Rerun Python SDK is incompatible with GLIBC 2.17, and therefore CentOS 7.

I don't know what is our minimum required GLIBC version.

# Problem
[A user reported](https://discord.com/channels/1062300748202921994/1126231826231480483) trying to use the Rerun Python SDK in a university cluster running CentOS 7, which uses GLIBC 2.17 (released 2012-12-25). The result: 

> ERROR: Could not find a version that satisfies the requirement rerun-sdk (from versions: none)

The user further reports:

> I see that the wheel file on PyPy specifies the platform tag `manylinux_2_31` which  is incompatible with my GLIBC version (2.17)

# Solution
I think the solution is building the wheels on a machine with an old glibc version.

There are some clues here:
* https://kobzol.github.io/rust/ci/2021/05/07/building-rust-binaries-in-ci-that-work-with-older-glibc.html
* https://github.com/explosion/tokenizations/pull/7

We should at least try, and we should also figure out what is the exact minimum GLIBC version required by the Rerun Python SDK.

nchodosh:
Hi,

I'm the user who originally created this issue. 

My university cluster was updated to CentOS 8 finally but our GLIBC is still only 2.28. I'd really like to try out your tool so I'm wondering if there's any progress on this?

jleibs:
@nchodosh when we tried to do this before we ran into some issues with build environments that were old enough to include the older glibc not having access to a recent enough libgtk3.

However, I just found out that the packages we build for `conda-forge` are built using an older glibc that should be compatible with centos8.

Is there any possibility you could use the Conda package instead of the pypi wheel?

Parskatt:
Ubuntu 18 is still stuck at glibc 2.28 (2.27?), can we get this on pip? Don't like conda installs.

ducha-aiki:
>My university cluster was updated to CentOS 8 finally but our GLIBC is still only 2.28. I

Exactly my situation as well.

alexanderswerdlow:
Same here; our cluster is still on CentOS 7.

I'd imagine that most users on very old glibc versions (maybe not Ubuntu 18, but likely CentOS 7) care less about display support, especially given the excellent support for streaming w/re-run, if that makes things any easier!

jleibs:
[...]

PR #10136, opened 2025-06-04
Build `manylinux_2_28` wheels (down from `manylinux_2_31`)

abey79:
### Related

* part of https://github.com/rerun-io/rerun/issues/2628

### What

title


abey79:
@rerun-bot full-check

abey79:
A better fix is underway

PR #10137, opened 2025-06-04
Make Rust getting-started example more idiomatic

wyfo:
Use `Iterator::zip`.

### DIFF
--- a/docs/content/getting-started/data-in/rust.md
+++ b/docs/content/getting-started/data-in/rust.md
@@ -27,7 +27,7 @@ For this example in particular, we're going to need all of these:
 [dependencies]
 rerun = "0.23"
 itertools = "0.14"
-rand = "0.9"
+rand = "0.8"
 ```
 
 While we're at it, let's get imports out of the way:
@@ -179,8 +179,8 @@ rec.log(
 Which only leaves the beads:
 
 ```rust
-let mut rng = rand::rng();
-let offsets = (0..NUM_POINTS).map(|_| rng.random::<f32>()).collect_vec();
+let mut rng = rand::thread_rng();
+let offsets = (0..NUM_POINTS).map(|_| rng.r#gen::<f32>()).collect_vec();
 
 let beads = lines
     .iter()


wyfo:
In `rerun::Position3D::from(bounce_lerp(p1, p2, offset))`, `rerun::Position3D::from` can in fact be removed, as `rerun::Points3D::new` already does the conversion.
I didn't remove it because one might prefer the conversion to be explicit, but I can do it if you judge that it's better without it — I prefer without it.

Wumpf (changes requested):
neat, looks better indeed. Can you please also copy this over to its source on https://github.com/rerun-io/rerun/blob/main/examples/rust/dna/src/main.rs?

> rerun::Position3D::from can in fact be removed, as rerun::Points3D::new already does the conversion.

yes, relying on `new` would be better. This code likely predates some prettification passes we did on the Rust SDK

examples/rust/dna/Cargo.toml
```diff
@@ -10,4 +10,4 @@ publish = false
 rerun = { path = "../../../crates/top/rerun" }
 
 itertools = "0.14"
-rand = "0.8"
+rand = "0.9"
```
Wumpf: Please revert this - this leads to crate duplication within the repository (see added crates on Cargo.lock) while not providing any concrete value. Instead it should be updated everywhere, but that's out of scope for this PR

wyfo: Do you want to also roll back the update of `rand` I made in #10134?

Wumpf: oh, yes that would be good, didn't see that PR.

cc/FYI: @abey79 I'd really want those snippets from .md  files and the respective "full examples" to be 100% in sync. It's unfortunate that we can't embedd directly there.

-> We should also sync back the cpp file, although I reckon this has been done already given that this fixed a compilation issue

Wumpf (approved):

Wumpf:
Rust tests failed due to spurious image comparision test, seen this one before, in any case nothing to do with this PR
Python test failure is interesting, but doesn't happen on `main`; failed in active area of development so I figure this has been fixed already (or may haunt us on nightly builds, who knows)

PR #10140, opened 2025-06-04
Group components by archetype in selection panel

lucasmerlin:
### Related

* part of https://github.com/rerun-io/rerun/issues/10030

### What
Group components by archetype in the selection panel (data ui and visualizers). Also changes components in popups.

I've also updated the text colors and image component button icons as described in the issue.

Screenshot 2025-06-04 at 12 35 42

### DIFF
--- a/crates/viewer/re_data_ui/src/lib.rs
+++ b/crates/viewer/re_data_ui/src/lib.rs
@@ -58,7 +58,7 @@ pub fn sorted_component_list_by_archetype_for_ui<'a>(
                 reflection
                     .fields
                     .iter()
-                    .position(|field| field.component_name == c.component_name)
+                    .position(|field| Some(field.name) == c.archetype_field_name)
                     .unwrap_or(usize::MAX)
             });
         } else {


crates/viewer/re_data_ui/src/instance_path.rs
```diff
@@ -91,12 +100,21 @@ impl DataUi for InstancePath {
...
         if ui_layout.is_single_line() {
+            let indicator_count = unordered_components
+                .iter()
+                .filter(|c| c.component_name.is_indicator_component())
+                .count();
+            // TODO: Change this text to somehting like "2 archetypes with 7 components total"?
             ui_layout.label(
                 ui,
                 format!(
                     "{} component{} (including {} indicator component{})",
```
lucasmerlin: Should I change this text to somehting like "2 archetypes with 7 components total"?

Wumpf: yes that would be nice! :)

lucasmerlin: Done!

crates/viewer/re_data_ui/src/lib.rs
```diff
@@ -59,9 +48,23 @@ pub fn sorted_component_list_by_archetype_for_ui<'a>(
...
+        if let Some(reflection) = archetype
+            .as_ref()
+            .and_then(|a| reflection.archetypes.get(a))
+        {
+            // Sort components by their importance
+            components.sort_by_key(|c| {
+                reflection
+                    .fields
+                    .iter()
+                    .position(|field| field.component_name == c.component_name)
```
Wumpf: this is gonna break down soon! right now the component_name is still unique within an archetype but soon we'll have several fields with the same component name == component type in archetypes (e.g. node  & edge color on graphs).
Instead you should match by field name

Wumpf (approved):
gorgeous! good to go once the comment on sorting is addressed!

Love the look overall but I wonder if the font isn't a bit too small. On low-dpi screen its on the edge of what's readable imho
image

Wumpf:
cc: @gavrelina

PR #10141, opened 2025-06-04
CMake improvement to download arrow inside the rerun build folder

jzubizarreta:
Small improvement to make sure the arrow code is downloaded inside the rerun build folder. This avoids polluting custom project build folders when using CMake and FetchContent


jzubizarreta:
It seems like I cannot add any labels to the issue :(

Wumpf (approved):
ah neat that makes sense! Thank you!

Issue #7931, opened 2024-10-29
Sync time axis for multiple TimeSeriesView


jviereck:
In my setup I am displaying data in two TimeSeriesView. The TimeSeriesView are positioned in a vertical viewport above each other. To compare the data from multiple axis, I have to manually position the x-axis (time in this case) to align well. It would be great if there would be a way to keep the two time-/x-axis in sync between the two views.

I was wondering if there could be a flag on the vertical viewport to sync the x-axis for contained TimeSeriesViews.

If someone could provide me with some pointers on how to implement this feature, I am more than happy to work on a pull request.

Wumpf:
> If someone could provide me with some pointers on how to implement this feature, I am more than happy to work on a pull request.

I think this boils down to being able to connect blueprint properties of views with each other, so this might quickly turn into a more general infrastructure task.
Essentially [this property](https://github.com/rerun-io/rerun/blob/main/crates/store/re_types/definitions/rerun/blueprint/views/time_series.fbs#L19) should be automatically sourced from a different view _or_ by implementing an inheritance scheme for containers (that would be closer to your idea of putting it on a container).

A more short-term solution that's worth experimenting with would be to add an optional view id to that property archetype specifically.

I'm surprised we didn't have an issue for this already (couldn't fine any at least)

teh-cmc:
> In my setup I am displaying data in two TimeSeriesView. The TimeSeriesView are positioned in a vertical viewport above each other. To compare the data from multiple axis, I have to manually position the x-axis (time in this case) to align well. It would be great if there would be a way to keep the two time-/x-axis in sync between the two views.

Is there anything in particular in your use case that is preventing you from displaying the two timeseries in the same view?

jviereck:
> Is there anything in particular in your use case that is preventing you from displaying the two timeseries in the same view?

The y-scale of the data is very different. This makes it hard to show the data in the same TimerSeriesView. For this task, it would be possible to have to Y-axis plotted in one TimeSeriesView. However, this works only for two y-axis and doesn't scale for more Y-axis or separating out the data into multiple TimeSeriesViews (e.g. one TimeSeriesView for forces, one for centroidal position, one for torques of a robot etc).

jviereck:
@Wumpf @teh-cmc Do you have a pointer for me how to update the global state to keep track from a shared x-axis range from the TimeSeriesViews?

Wumpf:
> Is there a way to pass down a shared state through all the components to all the views?

[...]

Issue #9789, opened 2025-04-23
Sub-plots within the TimeSeries view


abey79:
## Background

Axis-aligned stacks of plot is a widely requested feature for which there is exists no convenient work-around. #7931 proposes a somewhat generic mechanism to "synchronise" axes between multiple views, in hope to benefit from the layout flexibility of our view system. The present issue is a counter-proposal based on the (admittedly somewhat subjective) observation that the view system is too rigid and abstract for this purpose, and that a pragmatic, "sub-plots-within-a-view" approach would provide the same/similar results with lower complexity, both in UX design and code.

## Proposal

The core proposal is to allow multiple plots in a grid _within the same timeseries view_, à la matplotlib's subplots. Currently, a view can only contain a single plot. Also, within these sub-plots, any align axis would automatically be linked and aligned, without a possibility to opt-out (rationale: limit the number of configuration option + use the view system if you want loosely coupled plots).

## Challenges

- Design of the UI to configure the plot view (see proposal below)
- Design of the blueprint APIs (in particular, how the "entity path filter" mini-language fits in the picture
- Implementation:
  - egui_plot needs quite a bit of cleanup work
  - can the multi-plot layout—and alignement!—be done outside of egui_plots?

## UX proposal

The core proposal consist of a "mini-map" of the plot layout where entities can be dragged and dropped to the desired sub-plot.

By default, there is a single subplot, and all entities produced by the entity path filter are assigned to it:

<img src="https://github.com/user-attachments/assets/7cfd2fb4-a0c8-4a63-b8fd-544aebd84b87" width="550px">

An entity would be removed by filtering it out from the entity path filter.

The user would click on the + icons to create further rows/column of subplots, and then drag-and-drop entity to dispatch them to the desired sub-plot:

<img src="https://github.com/user-attachments/assets/4444180e-0760-45b9-b355-89010816ba9a" width="550px">

### Stretch goal 1: assign "weight" to columns or row

<img src="https://github.com/user-attachments/assets/8e20131b-918d-4aa3-b54a-3a01a128d043" width="550px">

### Stretch goal 2: optional assign an entity one of the X axes

An area would be displayed at the bottom of each column, in which case that entity would be used as abscissa instead of time values. This of course would require some time alignement for all scalars involved, but this is typically the case anyways. For such plot, the time cursor would be displayed as a marker highlighting the corresponding points, as opposed to a vertical line.

[...]

PR #10146, opened 2025-06-04
Add option to link the X-axis of multiple time plots

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/7931
* https://github.com/rerun-io/rerun/issues/9789

### What
This adds a blueprint option to link the horizontal/time/X axes of multiple plots together.

It's very simple, but delivers a lot of value for the amount of time it took to implement (<2h).

https://github.com/user-attachments/assets/19d77193-e397-4254-bfec-50cc12ccf908


crates/viewer/re_view/src/view_property_ui.rs
```diff
@@ -36,42 +36,29 @@ fn view_property_ui_impl(
     };
 
     let query_ctx = property.query_context(ctx);
-    // If the property archetype only has a single component, don't show an additional hierarchy level!
-    if reflection.fields.len() == 1 {
```
emilk, line 40: The old code caused a UI showing "Time axis" followed by a checkbox, omiting the crucial "Link" title.

In general, I think we should avoid special cases like this in the UI. Better to be consistent, if sometimes extra verbose.

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -474,6 +484,17 @@ impl ViewClass for TimeSeriesView {
...
+        if state.reset_bounds_next_frame {
+            plot = plot.reset();
+        }
+
+        if link_x_axis {
+            plot = plot.link_axis(timeline.name().as_str(), [true, false]);
+        }
+
+        // Sharing the same cursor is always nice:
+        plot = plot.link_cursor(timeline.name().as_str(), [true, false]);
```
emilk, line 498: Bonus feature! Hover one plot, see a vertical line in all plots.

HALtheWise:
This is really cool! Thanks @emilk 

Bug report: It appears that double-clicking a synced plot only zooms-to-fit on y, not x.

crates/store/re_types/definitions/rerun/blueprint/archetypes/time_axis.fbs
```diff
@@ -0,0 +1,15 @@
...
+
+/// Configuration for the time (Y) axis of a plot.
+table TimeAxis (
+    "attr.rerun.scope": "blueprint",
+    "attr.rust.derive": "Default"
+) {
+    // --- Optional ---
+
+    /// If true, link the X/time axis to all other plots where this is enabled.
+    link: rerun.blueprint.components.LinkAxis ("attr.rerun.component_optional", nullable, order: 2200);
```
Wumpf: @nikolausWest and I discussed for a while how we can have this feature without adding too much "conceptual debt", i.e. adding snowflake concepts that don't extend well with the rest of the Viewer. We didn't draw terribly good conclusions, but something we got particular wary of is to call this just "link" when it's more of a global state sharing kind of thing. We'd therefore advocate making this more explicit e.g. by making this an enum/radiobox choice between `Local` and `Link to global` (the global part here is important). Niko also proposed to link to the time range on the timeline since this is something that the viewer already has and is trivial to inspect for the user (as opposed to this more ellusive global state), but I worry that this is too limiting.
What do you think?

crates/store/re_types/definitions/rerun/blueprint/components/link_axis.fbs
```diff
@@ -0,0 +1,13 @@
+namespace rerun.blueprint.components;
+
+// ---
+
+    /// If true, link the X/time axis to all other plots where this is enabled.
```
Wumpf:
```suggestion
/// If true, link the X/time axis to all other plots where this is enabled.
```

crates/viewer/re_view/src/view_property_ui.rs
Wumpf: I get the motivation, but this is a pretty ugly regression here imho:

Before:
image


After:
image

The concrete problem on the time series view would also go away if we just add another property to the time axis 😁. Maybe let's add the lock x axis option like we have on the y axis, that might be really nice and not hard to do!

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -474,6 +484,17 @@ impl ViewClass for TimeSeriesView {
                 }
             });
 
+        if state.reset_bounds_next_frame {
+            plot = plot.reset();
+        }
+
+        if link_x_axis {
+            plot = plot.link_axis(timeline.name().as_str(), [true, false]);
```
Wumpf: convenient, didn't know egui plot had this option already!

crates/viewer/re_view/src/view_property_ui.rs
emilk: Maybe we can omit it only if the name of the archetype and the component is identical ("Near clip plane > Near clip plane")?

> Maybe let's add the lock x axis option like we have on the y axis, that might be really nice and not hard to do!

I rather not, at least right now. The X axis is all the same for all the plots, but the Y axis may be very different, so on the Y axis it would make more sense to have multiple link groups.

crates/store/re_types/definitions/rerun/blueprint/archetypes/time_axis.fbs
emilk: "Link to global" is probably better copy, and changing the ui to radio-buttons may be better UX (@gavrelina).

But I'd prefer to keep the feature itself as is for now. Sure we could link it to the visible time range on the streams panel, or the selected time range on the streams panel… but would that be better, or worse? The only thing that's certain is that it would definitely be a lot more complicated to implement!

crates/viewer/re_view_time_series/src/view_class.rs
emilk: Check out the egui_plot demo for more: https://emilk.github.io/egui_plot/ :)

crates/viewer/re_view/src/view_property_ui.rs
Wumpf: sounds good to me!

crates/store/re_types/definitions/rerun/blueprint/archetypes/time_axis.fbs
Wumpf: yeah agreed, the feature and its implementation can stay for as-is for now, not convinced either what's better!

crates/viewer/re_view/src/view_property_ui.rs
emilk: I did something similar:

Screenshot 2025-06-06 at 20 06 29

emilk:
Newest iteration of the UI:

image

emilk:
Screenshot 2025-06-07 at 07 58 52

PR #10147, opened 2025-06-05
Remove `_COMPONENT_DESCRIPTOR` from components in Python

grtlr:
### Related

* Closes #10058.
* Precursor to #10082.

### What

Analogous to the Rust API, we guarantee that all batches have a full descriptor.

PR #10150, opened 2025-06-05
Fix typo in `descr_custom_archetype.py` snippet

grtlr:
### Related

* This was introduced by #10147.

### What

Title.


grtlr:
Tested locally, so YOLO.

PR #10145, opened 2025-06-04
Add `UpdateDatasetEntry` rpc and update `DatasetEntry`

abey79:
### Related

* TODO

### What

TODO

PR #10156, opened 2025-06-06
Add APIs to Dataset to query and update the associated blueprint

abey79:
### Related

* Follow-up to https://github.com/rerun-io/rerun/pull/10145

### What

This PR introduces the following APIs to the Python SDK, related to the querying and updating of the associated blueprint dataset and default blueprint partition ID:

```python
class Dataset:
    # …
    
    @property
    def blueprint_dataset_id(self) -> EntryId | None:
        """The ID of the associated blueprint dataset, if any."""

    def blueprint_dataset(self) -> Dataset | None:
        """The associated blueprint dataset, if any."""

    @property
    def default_blueprint_partition_id(self) -> str | None:
        """The default blueprint partition ID for this dataset, if any."""

    def set_default_blueprint_partition_id(self, partition_id: str) -> None:
        """
        Set the default blueprint partition ID for this dataset.
        This fails if the change cannot be made to the remote server.
        """

    def partition_ids(self) -> list[str]:
        """Returns a list of partitions IDs for the dataset."""
```

The `partition_ids()` helper comes as a bonus, it proved useful for the dataset loader script.

DNM: chained


rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1034,6 +1034,27 @@ class Dataset(Entry):
     def arrow_schema(self) -> pa.Schema:
         """Return the Arrow schema of the data contained in the dataset."""
 
+    @property
```
jleibs: nit: I've been finding usage of properties here to be add more uncertainty than clarity. I'm sort of left guessing when I should be including a `()` if I'm just looking at bare code.

jleibs (approved):

rerun_py/rerun_bindings/rerun_bindings.pyi
abey79: Yeah I agree, the asymmetry of it all is annoying. I'll switch to pure functions.

Issue #5470, opened 2024-03-12
Design `tee` (multi-sink) API for Python


emilk:
We are currently redesigning how `init` works in Python as part of our ongoing Blueprint work.

It would be useful to come up with a design for multi-sink ("tee") so we don't have to change the API again in a few weeks.

We don't need to implement multi-sink, just decide and implement its API.

* This is part of https://github.com/rerun-io/rerun/issues/1998

Issue #1998, opened 2023-04-28
Support multiple sinks at once (`tee` logging)


teh-cmc:
Something along the lines of this pseudo code:
```rust
struct LogSinkBundle(Vec<Box<dyn LogSink>>)

impl LogSink for LogSinkBundle { ... }
```

also need to update CLI args helpers for both Rust and Python to support mixing `--save` with anything else.

emilk:
It is very useful to be able to dynamically add and remove sinks at runtime.

For instance:
* you start rerun with no sinks
* something happens, so you now start dumping to an .rrd on disk
* you hit a rare corner case happens, so you call `spawn` and start streaming to the viewer, then stop that stream once you reached the end of the corner case

This mean we need an API along the following lines:

```py
rerun = rr.new_recording(app_id=…, …)
rerun.add_sink(file_path="foo.rrd")
viewer_sink_id = rerun.add_sink(spawn=True)
…
rerun.remove_sink(viewer_sink_id)
```

When no sinks are attached, `rerun.log(…)` should be a no-op

emilk:
An interesting case to consider is streaming to an .rrd and calling `spawn`. We could set up the native viewer to be able to stream-load an .rrd file that is being written to, requiring no extra WebSocket connection. That would also mean the `rerun` library without the `WebSocket` feature (and huge dependency tree) could still support `spawn` (see https://github.com/rerun-io/rerun/issues/4788)

vmayoral:
Adding to the thread another potential use case of interest: https://github.com/rerun-io/rerun/issues/6685

In a nutshell: Periodic calls to `rr.save()` to generate smaller/manageable log files. This is particular relevant in robotic systems generating tons of data from depth sensors and multiple cameras.

vmayoral:
Hello all, any updates on this topic? Rationale at https://github.com/rerun-io/rerun/issues/1998#issuecomment-1887409983 makes a lot of sense to me as well. Thanks!

spewil:
Tracking this as well, having the ability to tee at the SDK would be great! I'd like to be able to persist logs as well as "check" on them as I wish. If I can help push this along, let me know!

bjin-bdai:
Came across this as well while trying to simultaneously stream data to the viewer while log to disk. I think this is a pretty strong feature so that we can safely have our data on disk in case the viewer crashes mid-session before we're able to save anything. 

[...]

PR #10158, opened 2025-06-07
Multi-sink / `tee` Python API

jprochazk:
* Part of https://github.com/rerun-io/rerun/issues/1998#issuecomment-1887409983
* Closes https://github.com/rerun-io/rerun/issues/5470

### What

We want some kind of "log to multiple sinks" API, which would alleviate some long-standing issues:
- Saving data to a file in case the Viewer crashes.
- Saving a larger-than-RAM recording to a file that's also being streamed to a Viewer, which will GC and drop large portions of it in an unrecoverable way.

#### API Additions

The entrypoint for this in the Python SDK would be `tee` on `RecordingStream`:

```py
import rerun as rr
from rerun.utilities import build_color_grid

rec = rr.RecordingStream("rerun_example_tee")
rec.tee(rr.FileSink("data.rrd"), rr.GrpcSink())

grid = build_color_grid()
rec.log("points", rr.Points3D(positions=grid.positions, colors=grid.colors))
```

The above snippets are functional on this PR. You must start a Viewer (or gRPC server) to connect to before running it. It logs the data to both the server and a `data.rrd` file.

The exposed `FileSink` and `GrpcSink` are simple dataclasses that temporarily hold information for how to setup the sink:
- `FileSink` stores a `path`
- `GrpcSink` stores a `url`, along with a flush timeout

The `tee` call then converts these sinks into their Rust equivalents, wraps them in a `MultiSink`, and calls `set_sink` on that.

Calling `set_sink` ensures the usual "dump backlog into the new sink" behavior is preserved, so that the initial `SetStoreInfo` logged by `rr.init` is sent through to every sink correctly.

The sink that's being set is the new `MultiSink`, a wrapper over `Vec<Box<dyn LogSink>>` which clones all received messages to every stored sink.

Currently only `FileSink` and `GrpcSink` are available. A user may use the same kind of sink multiple times, but there must not be any duplicates. For example, two `GrpcSink` with the same `url` would count as duplicates.

### Open questions

- What other sinks should be exposed?
  - `BufferedSink` doesn't make that much sense. It's something we use to temporarily hold data in memory (specifically only a `SetStoreInfo` message) before the user sets up a real "endpoint" sink, like GRPC or a file.
  - `MemorySink` and `CallbackSink` are important, as they interact with Notebook/Gradio/JS integration. We should allow for example sending data to a Viewer embedded in a notebook, and saving the same data to a file at the same time.
- Needs name bikeshedding, `tee` _makes sense_, but isn't very discoverable.
- Can we avoid the extra overhead of cloning all messages to every sink?
  - The current `LogSink` trait makes this impossible, as it requires that every stored sink be passed an owned `LogMsg`.


rerun_py/rerun_sdk/rerun/sinks.py
```diff
@@ -25,6 +26,89 @@ def is_recording_enabled(recording: RecordingStream | None) -> bool:
...
+            flush_timeout_sec=self.flush_timeout_sec,
+        )
+
+
+@dataclass(frozen=True)
+class FileSink:
+    path: str
+
+    def _to_native(self) -> bindings.PyFileSink:
+        return bindings.PyFileSink(path=self.path)
```
jleibs: I don't see much reason for these to include python-native wrappers. Vs exposing the rust objects directly. See, for example: `IndexColumnSelector` / `ComponentColumnDescriptor` as examples of making rust-defined objects directly accessible.

rerun_py/src/python_bridge.rs
```diff
@@ -331,7 +334,7 @@ fn shutdown(py: Python<'_>) {
     re_log::debug!("Shutting down the Rerun SDK");
     // Release the GIL in case any flushing behavior needs to cleanup a python object.
     py.allow_threads(|| {
-        // NOTE: Do **NOT** try and drain() `all_recordings` here.
+        // NOTE: Do **NOT* try and drain() `all_recordings` here.
```
jleibs: Looks like an unintentional change?

jleibs (commented):
Clever solution.  

Not having sink-objects was the big issue we ran into before, and you kind of side-stepped that by not making, e.g. GrpcSink a first-class wrapper on top of the rust-side sinks. I don't remember why that was hard before, or if it was just work.

This could still serve as a stepping stone in that direction though.

nikolausWest:
In general I think this looks great but I think we should skip the `rr.tee()` api and stick to the api on the `rr.RecordingStream` object. Other than than let's land it!

rerun_py/rerun_sdk/rerun/recording_stream.py
```diff
@@ -485,6 +486,22 @@ def __del__(self) -> None:  # type: ignore[no-untyped-def]
...
+        self,
+        *sinks: LogSinkLike,
+        default_blueprint: BlueprintLike | None = None,
+    ) -> None:
+        """
+        Stream data to multiple different sinks.
+
+        Duplicate sinks are not allowed. For example, two [`GrpcSink`]s that
+        use the same `url`.
+        """
```
nikolausWest, line 523: Example code might be useful here

nikolausWest:
This feature will also require a how-to guide and perhaps a look over our docs in general to make sure it's discoverable. That means it should be referenced to in the right places where one might wonder about it (e.g. in the rr.init docs) fyi @pweids

Wumpf:
@pweids / @nikolausWest / @jprochazk are we still taking this on for 0.24 or do we think it's too much too soon?
(I know we'd all love to have it, but seems like yet-another-feature?)

jprochazk:
It's not far from completion, I think we should land it for 0.24

crates/top/re_sdk/src/log_sink.rs
```diff
@@ -74,6 +74,52 @@ pub trait LogSink: Send + Sync + 'static {
...
+        }
+    }
+
+    // NOTE: this is only really used for BufferedSink,
+    //       and by the time you `set_sink` you probably don't have
+    //       a buffered sink anymore
+    #[inline]
+    fn drain_backlog(&self) -> Vec<LogMsg> {
+        Vec::new()
+    }
```
abey79, line 118: Trusting you here—I don't have a deep enough understanding of that stuff to make a call.

rerun_py/src/python_bridge.rs
```diff
@@ -618,6 +621,114 @@ fn spawn(
...
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.uri.hash(state);
+    }
+}
+
+#[pymethods]
+impl PyGrpcSink {
+    #[new]
+    #[pyo3(signature = (url=None, flush_timeout_sec=re_sdk::default_flush_timeout().expect("always Some()").as_secs_f32()))]
+    #[pyo3(text_signature = "(self, url=None, flush_timeout_sec=None)")]
```
abey79, line 647: I'm look surprise to see both here. Aren't they redundant?

rerun_py/src/python_bridge.rs
```diff
@@ -618,6 +621,114 @@ fn spawn(
...
+#[derive(PartialEq, Hash)]
+struct PyFileSink {
+    path: String,
+}
+
+#[pymethods]
+impl PyFileSink {
+    #[new]
+    #[pyo3(signature = (path))]
+    #[pyo3(text_signature = "(self, path)")]
```
abey79, line 671: ditto

rerun_py/src/python_bridge.rs
```diff
@@ -618,6 +621,114 @@ fn spawn(
...
+    fn new(path: String) -> Self {
```
abey79: You should use `PathBuf` here, and annotate `os.PathLike`.

See https://pyo3.rs/v0.25.1/conversions/tables.html?highlight=PathBuf#argument-types

rerun_py/src/python_bridge.rs
```diff
@@ -672,7 +783,7 @@ fn connect_grpc_blueprint(
 ) -> PyResult<()> {
     let url = url.unwrap_or_else(|| re_sdk::DEFAULT_CONNECT_URL.to_owned());
 
-    if let Some(blueprint_id) = (*blueprint_stream).store_info().map(|info| info.store_id) {
+    if let Some(blueprint_id) = (blueprint_stream).store_info().map(|info| info.store_id) {
```
abey79:
```suggestion
    if let Some(blueprint_id) = blueprint_stream.store_info().map(|info| info.store_id) {
```

rerun_py/rerun_sdk/rerun/recording_stream.py
```diff
@@ -485,6 +486,46 @@ def __del__(self) -> None:  # type: ignore[no-untyped-def]
     get_recording_id = get_recording_id
     is_enabled = is_enabled
 
+    def tee(
```
abey79, line 489: Has there been discussion about this choice of name?

I suggest using `set_sinks()` instead, because it's more explicit, and it is also more future-proof (ie. I would like to drop all these weird `save()`, etc. APIs in favour of exposing all sinks and being explicit with plugging a stream onto them).

rerun_py/src/python_bridge.rs
jprochazk: No, `signature` is what gets pasted into the code on the Rust side. Normally that also generates a `text_signature`, but not for `new`. We instead have a separate `text_signature` which our runtime vs stub type checking script reads.

abey79: TIL 👍🏻

rerun_py/rerun_sdk/rerun/recording_stream.py
jprochazk: I agree that `tee` probably isn't the best name, but I like it more than `set_sinks`. Another followup to the list, properly document this everywhere we talk about operating modes.

I'm going to start a bikeshedding thread in Slack, since we have some time until 0.24 to change it, but leave it as is for now and merge

jprochazk: https://rerunio.slack.com/archives/C033K5VS2KD/p1750343062698689

crates/top/re_sdk/src/log_sink.rs
jprochazk: Another thing which I didn't write here: which sink do you pick to call `drain_backlog` on? It's not clear to me at all, but luckily we probably don't need to worry about that, since you can't insert a `BufferedSink` into a `MultiSink` as a user

rerun_py/rerun_sdk/rerun/recording_stream.py
thz: The name `tee` suggests two sinks, which seems arbitrary to me. Wouldn't it be cleaner to have 2 (or more) calls to something like `add_sink`?

jprochazk: We decided on `.attach_sinks` in the slack thread. I'll merge this as-is (with review other comments addressed) and do that in a follow-up PR

jprochazk: > Wouldn't it be cleaner to have 2 (or more) calls to something like `add_sink`?

The way this works under the hood would make any kind of "add sink" operation a bit more complicated. The presence of that initial `SetStoreInfo` message means we _must_ set all the sinks "at once"

PR #10160, opened 2025-06-07
Show help clicking the `?` as well as by just hovering it

emilk:
This makes it easier to discover

help-button


emilk:
Ine of our snapshot tests fails because I changed the “❓” text for a proper svg icon:

> harness.get_by_label(“:question:”).hover();

https://github.com/rerun-io/rerun/blob/38bf741a0da02de3ca47e4b2357d7ad595c509d2/crates/viewer/re_ui/tests/help_ui_test.rs#L57

@lucasmerlin any idea how we best fix this?

crates/viewer/re_ui/src/ui_ext.rs
```diff
@@ -1068,12 +1067,28 @@ pub trait UiExt {
...
+            .ui(ui, |ui| {
+                if let Some(help_ui) = help_ui.take() {
+                    help_ui(ui);
+                }
+            })
+            .0;
+
+        if let Some(help_ui) = help_ui.take() {
+            button_response.on_hover_ui(help_ui)
+        } else {
```
lucasmerlin, line 1098: It would be nice if there was some way to force a tooltip open

emilk: yes - that would be a nicer solution here I think… but this will have to do for now.

emilk:
I added an invisible "alt_text" to all our icon buttons (per Lucas' suggestion). It can be used by kittest and for accessibility (e.g. screen readers), but I also find it improves our code readability by forcing us to document what an icon click actually does

lucasmerlin:
Awesome!

Issue #10167, opened 2025-06-08
rr.TimeColumn doesn't support numpy array with epoch second timestamps


MichaelGrupp:
I tried to log [a batch of data](https://github.com/MichaelGrupp/evo/tree/master/contrib/rerun_example) with timestamps that are stored as plain seconds since Unix epoch in a numpy array.
According to the `TimeColumn` documentation, I expected that this should be supported:
```
        timestamp:
            Used for absolute time indices, like `capture_time`.
            Must either be in seconds since Unix epoch, [`datetime.datetime`][], or [`numpy.datetime64`][].
```

Logging succeeded but the timeline in the viewer was completely wrong.

[Looking at the code](https://github.com/rerun-io/rerun/blob/48974373a6efe724a0203ea0f8ee56116ff4c74c/rerun_py/rerun_sdk/rerun/_send_columns.py#L114), it turns out that `TimeColumn.__init__()` implicitly assumes any np.ndarray passed as `timestamp` to have a datetime dtype. This effectively means your seconds array will be interpreted as nanoseconds.

> Note: the `duration` kwarg works with numeric arrays as advertised, there is a conversion.

**Workaround:** convert numpy array to list 
```python
rr.TimeColumn("time", timestamp=list(timestamps)))
```

**Proposed fix:** convert the array, similar to what is done for `duration`.

PR #10168, opened 2025-06-08
Fix support for seconds-since-Epoch numpy arrays in TimeColumn.__init…

MichaelGrupp:
Apply the equivalent of the conversion that is done for the `duration` kwarg.

Fixes accidental interpretation of seconds as nanoseconds.

Closes: #10167

### DIFF
--- a/rerun_py/rerun_sdk/rerun/_send_columns.py
+++ b/rerun_py/rerun_sdk/rerun/_send_columns.py
@@ -112,7 +112,14 @@ class TimeColumn(TimeColumnLike):
         elif timestamp is not None:
             # TODO(zehiko) add back timezone support (#9310)
             if isinstance(timestamp, np.ndarray):
-                self.times = pa.array(timestamp.astype("datetime64[ns]"), pa.timestamp("ns"))
+                if np.issubdtype(timestamp.dtype, np.datetime64):
+                    # Already a datetime array, just ensure it's in nanoseconds
+                    self.times = pa.array(timestamp.astype("datetime64[ns]"), pa.timestamp("ns"))
+                elif np.issubdtype(timestamp.dtype, np.number):
+                    # Numeric array that needs conversion to nanoseconds
+                    self.times = pa.array((timestamp * 1e9).astype("datetime64[ns]"), pa.timestamp("ns"))
+                else:
+                    raise TypeError(f"Unsupported numpy array dtype: {timestamp.dtype}")
             else:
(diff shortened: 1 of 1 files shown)


rerun_py/rerun_sdk/rerun/_send_columns.py
```diff
@@ -112,7 +112,14 @@ def __init__(
...
             if isinstance(timestamp, np.ndarray):
-                self.times = pa.array(timestamp.astype("datetime64[ns]"), pa.timestamp("ns"))
+                if np.issubdtype(timestamp.dtype, np.datetime64):
+                    # Already a datetime array, just ensure it's in nanoseconds
+                    self.times = pa.array(timestamp.astype("datetime64[ns]"), pa.timestamp("ns"))
+                elif np.issubdtype(timestamp.dtype, np.number):
+                    # Numeric array that needs conversion to nanoseconds
+                    self.times = pa.array((timestamp * 1e9).astype("datetime64[ns]"), pa.timestamp("ns"))
+                else:
+                    raise TypeError(f"Unsupported numpy array dtype: {timestamp.dtype}")
```
MichaelGrupp, line 122: Almost the same as the equivalent in the other if-branch a few lines above. Let me know if duplication should be reduced here somehow.

_Conceptually_, it should be also possible to simplify this to a single list comprehension that covers all cases like below, since numpy arrays are iterable (with a slight (neglectible?) performance loss, but less bug prone code).

Wumpf (changes requested):
thank you for looking into this!
Change looks good to me as is, but there should be test cases added to `rerun_py/tests/unit/test_time_column.py`

rerun_py/tests/unit/test_time_column.py
```diff
@@ -86,6 +86,30 @@ def test_duration_column(
...
+        pa.array(
+            [
+                np.datetime64("1970-01-01T00:00:00", "ns"),
+                np.datetime64("1970-01-01T00:00:01.50", "ns"),
+                np.datetime64("1970-01-01T00:00:02.25", "ns"),
+                np.datetime64("1970-01-01T00:00:03", "ns"),
+            ],
+            type=pa.timestamp("ns"),
+        ),
+    ),
```
MichaelGrupp, line 112: Same test fixture as the one for plain lists above, but with np.array.

Wumpf (approved):
neat, thx!

Issue #10170, opened 2025-06-08
Python lists/arrays of RGB(A) colors can't be sent as columns


MichaelGrupp:
**To Reproduce**

Try to send columns with color lists specified as list/array of RGB or RGBA values, not as packed ints.

You will get an error like:
```
RerunWarning: send_columns: ArrowInvalid(Length spanned by list offsets (20) larger than values array (length 5))
```

E.g. pick one of the examples from https://rerun.io/docs/howto/logging/send-columns and modify it to have RGBA tuples instead of ints as colors.

<details> <summary> Click to show example </summary>

```python
from __future__ import annotations

import numpy as np
import rerun as rr

rr.init("rerun_example_points3d_column_updates", spawn=True)

# Prepare a point cloud that evolves over 5 timesteps, changing the number of points in the process.
times = np.arange(10, 15, 1.0)
# fmt: off
positions = [
    [1.0, 0.0, 1.0], [0.5, 0.5, 2.0],
    [1.5, -0.5, 1.5], [1.0, 1.0, 2.5], [-0.5, 1.5, 1.0], [-1.5, 0.0, 2.0],
    [2.0, 0.0, 2.0], [1.5, -1.5, 3.0], [0.0, -2.0, 2.5], [1.0, -1.0, 3.5],
    [-2.0, 0.0, 2.0], [-1.5, 1.5, 3.0], [-1.0, 1.0, 3.5],
    [1.0, -1.0, 1.0], [2.0, -2.0, 2.0], [3.0, -1.0, 3.0], [2.0, 0.0, 4.0],
]
# fmt: on

# At each timestep, all points in the cloud share the same but changing color and radius.
colors = [
    (255, 0, 0, 255),
    (0, 255, 0, 255),
    (0, 0, 255, 255),
    (255, 255, 0, 255),
    (0, 255, 255, 255),
]
radii = [0.05, 0.01, 0.2, 0.1, 0.3]

rr.send_columns(
    "points",
    indexes=[rr.TimeColumn("time", duration=times)],
    columns=[
        *rr.Points3D.columns(positions=positions).partition(lengths=[2, 4, 4, 3, 4]),
        *rr.Points3D.columns(colors=colors, radii=radii),
    ],
)
```

</details>

**Expected behavior**

All `Rgba32ArrayLike` input works (preferred), or document & check that only 1D lists of int colors are supported.

**Desktop (please complete the following information):**
 - OS: 
macOS Sequoia 15.5

**Rerun version**

```
[...]

PR #10171, opened 2025-06-08
Fix wrong batch_length of colors column in Python API

MichaelGrupp:
The batch_length must be always 1 because this is the length that the ColorsBatch has after conversion of the user input.

Fixes: #10170

---

Note: there might be a more stable solution, but this seems to be at least viable as a hotfix.


Wumpf (changes requested):
Thanks for trying to address this! However, matching indiscriminately by field name name and from there assuming type/shape of the data breaks way too easily. We have to come up with something better here. Closing the PR since the approach won't work (happy about another attempt though :)

I'm also not sure we can solve this for the general case anyways. I'll write more notes on the ticket

Issue #9973, opened 2025-05-14
Investigate e2e performance of `rr.log(…, rr.Image(numpy))`


emilk:
While playing with the LeRobot Rerun visualizer we noticed a big (and ever-increasing) latency when logging raw images from Python to Rerun over gRPC.

I want to investigate this.

## Repro
Log images from TWO cameras, each at 640x480 RGB.
The images are to be logged at 180 Hz (this is a bug in LeRobot - see https://github.com/huggingface/lerobot/pull/903/files#r2089058089)

Call:
`rr.log(entity_path, rr.Image(foo.numpy()), static=True)`

## Observed
After a few seconds, we have a few seconds of latency, and it keeps increasing.
Turn on the Rerun performance metrics to see the latency in the UI.

## Wanted
As a user, I want to understand _where_ the latency is coming from. Is it in the SDK or the viewer?

I also want some way to fix it. There should be a `rr.flush()` command (there is only `rec.flush`).

There should also be a maximum length on queues so that we get some back-pressure, or at least an option to turn that on. And if any queue grows too long (in seconds) we should log a warning so we understand which queue.

Issue #1891, opened 2023-04-18
Drop `log_time`


teh-cmc:

emilk:
This sounds like two separate issues to me.

Getting rid of `log_time` is a space-optimization, but adds complexity.

Adding `ingest_time` is a new feature.

Wumpf:
Closing this: We have ingest_time now, see [#10173](https://github.com/rerun-io/rerun/pull/10173) and there's this issue for optional log time/tick:
* #10243

PR #10173, opened 2025-06-09
Add detailed view of latency

emilk:
## Related
* Investigates https://github.com/rerun-io/rerun/issues/9812
* Closes https://github.com/rerun-io/rerun/issues/9973

## What
This adds a breakdown of where the end-to-end latency is coming from. It automatically appears in the top panel when latency is high. Hover it to see details:

image

## Notable changes
The latency metrics are no longer opt-in. We always show them now if they get surprisingly high.

The measurements are made by adding extra metadata to the record batches at IPC encode/decode time.

We already have accurate timestamps at log-time (`RowId`) and batch-time (`ChunkId`).

## Testing

Apply this diff:

```
diff --git a/tests/rust/log_benchmark/src/image.rs b/tests/rust/log_benchmark/src/image.rs
index e36478262a1..877c3e0bce4 100644
--- a/tests/rust/log_benchmark/src/image.rs
+++ b/tests/rust/log_benchmark/src/image.rs
@@ -26,8 +26,7 @@ fn prepare() -> Vec<u8> {
 fn execute(mut raw_image_data: Vec<u8>) -> anyhow::Result<()> {
 
-    let (rec, _storage) =
-        rerun::RecordingStreamBuilder::new("rerun_example_benchmark_").memory()?;
+    let rec = rerun::RecordingStreamBuilder::new("rerun_example_benchmark_").connect_grpc()?;
 
     for i in 0..NUM_LOG_CALLS {
         raw_image_data[i] += 1; // Change a single pixel of the image data, just to make sure we transmit something different each time.
```

The run this in one terminal:

```sh
pixi run rerun-release 
```

and in another:

```sh
cargo run -p log_benchmark --release -- --benchmarks image --profile
```

## Findings
In a test where I encode a bunch of uncompressed 1024² RGBA images (`tests/rust/log_benchmark/src/image.rs`).
I found that by far the slowest part is the IPC encoding of arrow data (yes, this is a release build):

Image

zooming in:

Image

This suggests that if we simply encode the chunks in parallel instead on N cores, we will get a N-times speedup (when the actual logging thread is not the bottleneck).

This also gives a likely reason for the recent regression: we switched from `arrow2` to `arrow-rs`.

Perhaps we could also spend some time optimizing `arrow::ipc::writer::StreamWriter`


nikolausWest:
Really great stuff! So great to see this breakdown so easily.

> This suggests that if we simply encode the chunks in parallel instead on N cores, we will get a N-times speedup (when the actual logging thread is not the bottleneck).

We should keep in mind here though that in the user's system, it may not be desirable for the logging to take up all the resources. Don't know what the answer is but this might need to be a setting (using many threads to improve latency is obviously great ion an off-device bridge but perhaps not on device in production). 

> Perhaps we could also spend some time optimizing arrow::ipc::writer::StreamWriter

Any improvements we can do to the single threaded case would be great in both cases =)

jprochazk:
I wonder how the changes from this draft affect the measurements here:

- https://github.com/rerun-io/rerun/pull/9826

Will try it myself soonish

jprochazk:
> The latency metrics are no longer opt-in. We always show them now if they get surprisingly high.
> The measurements are made by adding extra metadata to the record batches at IPC encode/decode time.

I would expect this to be part of the "show performance metrics" setting. How much overhead does this add to encode/decode time?

emilk:
> > The latency metrics are no longer opt-in. We always show them now if they get surprisingly high.
> > The measurements are made by adding extra metadata to the record batches at IPC encode/decode time.
> 
> I would expect this to be part of the "show performance metrics" setting. How much overhead does this add to encode/decode time?

We need to insert the measurement on the SDK side, without knowing the settings for the viewer. But this is only a single time measurement per chunk, so the added overhead is minimal.

crates/viewer/re_viewer/src/ui/top_panel.rs
```diff
@@ -426,103 +448,75 @@ fn memory_use_label_ui(ui: &mut egui::Ui, gpu_resource_stats: &WgpuResourcePoolS
...
-                    "Queue latency: {}, length: {}",
-                    latency_text(latency_sec),
-                    format_uint(queue_len),
-                ));
-
-                ui.label(
-                    "When more data is arriving over network than the Rerun Viewer can ingest, a queue starts building up, leading to latency and increased RAM use.\n\
-                         We call this the queue latency.");
-            });
-        }
```
emilk, line 450: I removed the old "queue latency", which was just measuring the latency from receiving data until ingesting it; which we now cover in another way with the new code

nikolausWest:
Perhaps a bit silly but did a Chatgpt deep research to look for opportunities to improve performance. May be something in there of interest

[Performance of Arrow StreamWriter in Rerun.pdf](https://github.com/user-attachments/files/20657932/Performance.of.Arrow.StreamWriter.in.Rerun.pdf)

crates/viewer/re_viewer/src/ui/top_panel.rs
```diff
@@ -426,103 +448,75 @@ fn memory_use_label_ui(ui: &mut egui::Ui, gpu_resource_stats: &WgpuResourcePoolS
...
+                }
+                if let Some(transmission) = transmission {
+                    ui.label("encode -> decode (transmission)");
+                    ui.monospace(latency_text(transmission));
+                    ui.end_row();
+                }
+                if let Some(decode2ingest) = decode2ingest {
+                    ui.label("decode -> ingest");
+                    ui.monospace(latency_text(decode2ingest));
+                    ui.end_row();
```
emilk: Any suggestions on how we make this more understandable?

I was thinking maybe having it be like a chain instead:


```
  log
    ↓
  2.1ms
    ↓
  batch
    ↓
  4.5ms
    ↓
SDK send
    ↓
  32.2ms
    ↓
Viewer receive
    ↓
  12.2ms
    ↓
 ingest
```
  
Would that make it clear that the measurement is the delay of the gap between each event?

Or should we name the gaps (e.g. "batching", "transmission", "input processing)?

Maybe @gavrelina has some thoughts

crates/viewer/re_global_context/src/app_options.rs
```diff
@@ -67,8 +67,7 @@ pub struct AppOptions {
 impl Default for AppOptions {
     fn default() -> Self {
         Self {
-            low_latency: 0.100,
-            warn_latency: 0.200,
+            warn_e2e_latency: 1.0,
```
emilk, line 70: I also renamed it so that the old value in peoples configs wouldn't be transferred to the new option

crates/store/re_sorbet/src/timing_metadata.rs
```diff
@@ -0,0 +1,102 @@
...
+
+use crate::ArrowBatchMetadata;
+
+/// When was this batch last encoded into IPC bytes?
+/// Usually this is when the batch was last sent over gRPC.
+pub const LAST_ENCODED_AT: &str = "sorbet.encoded_at";
+
+/// When was this batch last decoded from IPC bytes?
+/// Usually this is when the batch was last received over gRPC.
+pub const LAST_DECODED_AT: &str = "sorbet.decoded_at";
```
emilk: Note to self: these could use better names (e.g. `last_ipc_encoded_at`)

crates/store/re_log_encoding/src/protobuf_conversions.rs
```diff
@@ -51,12 +51,15 @@ pub fn log_msg_from_proto(
...
-                store_id,
-                chunk.to_arrow_msg()?,
-            ))
+            let arrow_msg = re_log_types::ArrowMsg {
+                chunk_id: sorbet_schema.chunk_id().as_tuid(),
+                batch,
+                on_release: None,
+            };
+
+            Ok(re_log_types::LogMsg::ArrowMsg(store_id, arrow_msg))
```
emilk, line 62: This is just a drive-by optimization - no need to roundtrip to a whole Chunk

crates/store/re_log_encoding/src/codec/file/decoder.rs
```diff
@@ -49,9 +49,15 @@ pub fn decode_bytes(message_kind: MessageKind, buf: &[u8]) -> Result<Option<LogM
...
+            let sorbet_schema = re_sorbet::ChunkSchema::try_from(batch.schema_ref().as_ref())?;
 
-            Some(LogMsg::ArrowMsg(store_id, chunk.to_arrow_msg()?))
+            let arrow_msg = re_log_types::ArrowMsg {
+                chunk_id: sorbet_schema.chunk_id().as_tuid(),
+                batch,
+                on_release: None,
+            };
+
+            Some(LogMsg::ArrowMsg(store_id, arrow_msg))
```
emilk, line 61: Same drive-by optimization here

crates/store/re_log_encoding/src/codec/arrow.rs
```diff
@@ -8,7 +8,18 @@ pub(crate) fn write_arrow_to_bytes<W: std::io::Write>(
...
+    re_tracing::profile_function!();
+
+    let mut schema = std::sync::Arc::unwrap_or_clone(batch.schema().clone());
+
+    // Remember when the encoding took place,
+    // for ~accurate latency measurement of e.g. gRPC calls.
+    schema.metadata.insert(
+        re_sorbet::timing_metadata::LAST_ENCODED_AT.to_owned(),
+        re_sorbet::timing_metadata::now_timestamp(),
+    );
```
emilk: I should maybe move these to `re_grpc_server/client` instead so we don't add them when writing to file… but then again, it's nice to always have these latency measurements, e.g. when e.g. piping data via stdin/stdout, or when using redap (when it supports streaming, in some future)

Wumpf:
> Perhaps a bit silly but did a Chatgpt deep research to look for opportunities to improve performance. May be something in there of interest
> 
> [Performance of Arrow StreamWriter in Rerun.pdf](https://github.com/user-attachments/files/20657932/Performance.of.Arrow.StreamWriter.in.Rerun.pdf)

quite silly, yes. Most of it is just a wall of text describing Rerun with lots of weird half truths, followed by generalisms on how this sort of thing is approached and some random bs (like "use an async streamwriter" (wat?!), "upgrade to arrow 54". why? doesn't say)

crates/store/re_entity_db/src/ingestion_statistics.rs
```diff
@@ -0,0 +1,216 @@
...
+/// Statistics about the latency of incoming data to a store.
+pub struct IngestionStatistics {
+    store_id: StoreId,
+    stats: Mutex<LatencyStats>,
+}
+
+impl IngestionStatistics {
+    #[inline]
+    pub fn on_events(&self, timestamps: &TimestampMetadata, events: &[ChunkStoreEvent]) {
+        if let Some(nanos_since_epoch) = nanos_since_epoch() {
```
Wumpf: if we're not able to get nanos_since_epoch, aren't the other timestamps still valid?

crates/store/re_entity_db/src/ingestion_statistics.rs
```diff
@@ -0,0 +1,216 @@
...
+            let mut stats = self.stats.lock();
```
Wumpf: nit: too eager locking? many calls to `on_events` won't have anything with this story id :thinking:

crates/store/re_entity_db/src/ingestion_statistics.rs
```diff
@@ -0,0 +1,216 @@
...
+    transmission: History<f32>,
+
+    /// Time from the incoming IPC data being decoded to it being ingested into the store.
+    decode2ingest: History<f32>,
+}
+
+impl Default for LatencyStats {
+    fn default() -> Self {
+        let min_samples = 0; // 0: we stop displaying e2e latency if input stops
+        let max_samples = 32 * 1024; // don't waste too much memory on this - we just need enough to get a good average
```
Wumpf: 32 * 1024 * sizeof(f32) * 5 == 0.625MiB per stats object (i.e. per store). Acceptable, but not that little actually hmmmmm

crates/store/re_entity_db/src/ingestion_statistics.rs
```diff
@@ -0,0 +1,216 @@
...
+        let TimestampMetadata {
+            last_encoded_at,
+            last_decoded_at,
+        } = timestamps;
+
+        let last_encoded_at_nanos = last_encoded_at.and_then(system_time_to_nanos);
+        let last_decoded_at_nanos = last_decoded_at.and_then(system_time_to_nanos);
+
+        for row_id in chunk.row_ids() {
+            let row_creation_nanos = row_id.nanos_since_epoch() as i64;
```
Wumpf, line 100: question for outside of this PR: doe you have an intuition of how eagerly we're creating row ids? I.e. when user data comes in ideally the first thing we should do is to create this TUID, best even before arrow serialization :lightbulb:

crates/store/re_entity_db/src/ingestion_statistics.rs
```diff
@@ -0,0 +1,216 @@
...
+            e2e.flush(now);
+            log2chunk.flush(now);
+            chunk2encode.flush(now);
+            transmission.flush(now);
+            decode2ingest.flush(now);
+        }
+
+        Some(LatencySnapshot {
+            e2e: e2e.average()?,
+            log2chunk: log2chunk.average()?,
```
Wumpf: related to handling `nanos_since_epoch` failing: if we don't have e2e time, aren't the other timings still interesting? Sure it complicates the logic a bit, but that way we'd be able to still show a lot of timings even in the web viewer

crates/store/re_log_encoding/src/codec/arrow.rs
```diff
@@ -8,7 +8,18 @@ pub(crate) fn write_arrow_to_bytes<W: std::io::Write>(
     writer: &mut W,
     batch: &ArrowRecordBatch,
 ) -> Result<(), CodecError> {
-    let mut sw = arrow::ipc::writer::StreamWriter::try_new(writer, batch.schema_ref())
+    re_tracing::profile_function!();
+
+    let mut schema = std::sync::Arc::unwrap_or_clone(batch.schema().clone());
```
Wumpf: `unwrap_or_clone` on a thing you just cloned doesn't make sense. _obviously_ `Arc::unwrap_or_clone` has to clone the thing now, only that we needlessly first bumped the ref count and then called `Arc::try_unwrap`

Can as well just clone the thing immediately and save on a lot of atomic operations (which are hard for compilers to optimize)

```suggestion
    let mut schema = (*batch.schema()).clone();
```

crates/store/re_log_encoding/src/codec/arrow.rs
Wumpf: agreed, might as well keep them always. It's not adding all that much overhead to the file I reckon anyways
Can you please add a comment here to document this rationale? Lest we visit this later and wonder why

nikolausWest:
> quite silly, yes. Most of it is just a wall of text describing Rerun with lots of weird half truths, followed by generalisms on how this sort of thing is approached and some random bs (like "use an async streamwriter" (wat?!), "upgrade to arrow 54". why? doesn't say)

Yeah I only skimmed it before sharing and it seemed interesting. After reading it turns out it was mostly AI slop 🤷‍♂️. Some day it will work!

crates/store/re_log_encoding/src/protobuf_conversions.rs
Wumpf: if you would have put those into a separate PR you could have had a nice changelog entry on optimizations... but I reckon you didn't have the time to check whether that's even a meaningful immpact anyways (?) :smile:

crates/store/re_sorbet/src/lib.rs
```diff
@@ -28,6 +28,7 @@ mod selectors;
 mod sorbet_batch;
 mod sorbet_columns;
 mod sorbet_schema;
+pub mod timing_metadata;
```
Wumpf: nit: this was such a nice flat interface until now ;-)

crates/store/re_sorbet/src/sorbet_schema.rs
```diff
@@ -27,6 +29,9 @@ pub struct SorbetSchema {
 
     /// The heap size of this batch in bytes, if known.
     pub heap_size_bytes: Option<u64>,
+
+    /// Statistics
```
Wumpf: nit^2
```suggestion
    /// Timing statistics.
```

crates/store/re_sorbet/src/timing_metadata.rs
Wumpf: How about:

```suggestion

/// When was this batch last encoded into IPC bytes?
/// Usually this is when the batch was last sent over gRPC.
pub const KEY_TIMESTAMP_IPC_ENCODED: &str = "sorbet.timestamp_ipc_encoded";

/// When was this batch last decoded from IPC bytes?
/// Usually this is when the batch was last received over gRPC.
pub const KEY_TIMESTAMP_IPC_DECODED: &str = "sorbet.timestamp_ipc_decoded";
```

I realize the `last` portion is important for accuracy, but I figure it's not often relevant..?

crates/store/re_sorbet/src/timing_metadata.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+#[test]
+fn test_timestamp_encoding() {
+    let now = web_time::SystemTime::now();
+    let encoded = encode_timestamp(now);
+    assert_eq!(encoded.len(), 20); // e.g. "1700000000.012345678"
+    let parsed = parse_timestamp(&encoded).unwrap();
+    assert_eq!(parsed, now);
+}
```
Wumpf: wasn't there a clippy lint for tests being at the bottom always? Thought we had that enabled

crates/utils/re_arrow_util/src/batches.rs
```diff
@@ -78,6 +78,21 @@ pub fn concat_polymorphic_batches(batches: &[RecordBatch]) -> arrow::error::Resu
...
+/// Add a new key/value pair to the metadata of a [`RecordBatch`],
+/// replacing any existing value for that key.
+// TODO(apache/arrow-rs#7628): this should be built into Arrow, but it isn't yet.
+#[must_use]
+pub fn insert_metadata(
+    record_batch: RecordBatch,
+    key: impl Into<String>,
+    value: impl Into<String>,
+) -> RecordBatch {
+    let mut new_schema = std::sync::Arc::unwrap_or_clone(record_batch.schema());
```
Wumpf, line 90: We own `record_batch` so in theory the `unwrap_or_clone` coooouuuuuld make sense. But since `schema()` clones out an `Arc` this `unwrap_or_clone` will also always clone. We'd need a way to deconstruct the record_batch to do this efficiently... might actually be worth it if this utility is called often. Meanwhile this less fancy version is actually faster:

```suggestion
    let mut new_schema = (*record_batch.schema_ref()).clone();
```

crates/utils/re_arrow_util/src/batches.rs
```diff
@@ -78,6 +78,21 @@ pub fn concat_polymorphic_batches(batches: &[RecordBatch]) -> arrow::error::Resu
...
+    new_schema.metadata.insert(key.into(), value.into());
+    #[allow(clippy::unwrap_used)] // cannot fail
```
Wumpf: Well, I didn't know the unwrap failure conditions until I looked ;)

```suggestion
    #[allow(clippy::unwrap_used)] // cannot fail because the new schema is always a superset of the old
```

crates/viewer/re_viewer/src/ui/top_panel.rs
Wumpf: I think that would be pretty nice the way you describe it. Main advantage with naming the gaps is brevity, but it makes it less clear how things are connected

But I trust @gavrelina's judgement more than my own here if she has more suggestions!

tests/rust/log_benchmark/src/image.rs
```diff
@@ -1,12 +1,11 @@
...
 
-// About 1gb of image data.
-const IMAGE_DIMENSION: u64 = 16_384;
+const IMAGE_DIMENSION: u64 = 1024;
 const IMAGE_CHANNELS: u64 = 4;
 
 // How many times we log the image.
 // Each time with a single pixel changed.
-const NUM_LOG_CALLS: usize = 4;
+const NUM_LOG_CALLS: usize = 20_000;
```
Wumpf, line 8: this benchmark exists also for C++ and Python and was meant to compare those. If you change this one, you should also change the others!

Wumpf:
* Fixes https://github.com/rerun-io/rerun/issues/1891

Well, actually only the optional part of it. But removing `log_time` isn't really on the table without deep consideration and strong motivation as it permeates a lot of things.

crates/store/re_entity_db/src/ingestion_statistics.rs
emilk: They all will, because this is not a real store subscriber. I'll rearrange the code to make that obvious

crates/store/re_entity_db/src/ingestion_statistics.rs
emilk: If we can't get the current time of day… then all bets are off

crates/store/re_entity_db/src/ingestion_statistics.rs
emilk: True - I can lower this a bit

crates/store/re_entity_db/src/ingestion_statistics.rs
emilk: It's early, but not that early:

https://github.com/rerun-io/rerun/blob/0b182fcb3b3545eafde2968605b96291c0193079/crates/top/rerun_c/src/lib.rs#L804-L806

https://github.com/rerun-io/rerun/blob/0b182fcb3b3545eafde2968605b96291c0193079/rerun_py/src/arrow.rs#L69-L72

https://github.com/rerun-io/rerun/blob/0b182fcb3b3545eafde2968605b96291c0193079/crates/top/re_sdk/src/recording_stream.rs#L1197

https://github.com/rerun-io/rerun/blob/0b182fcb3b3545eafde2968605b96291c0193079/crates/top/re_sdk/src/recording_stream.rs#L1237

crates/store/re_sorbet/src/timing_metadata.rs
emilk: Not to my knowledge. I think it makes sense to put unit-tests next to the functions they test 🤷

emilk:
image

crates/viewer/re_viewer/src/ui/top_panel.rs
emilk: image

crates/utils/re_arrow_util/src/batches.rs
Wumpf: didn't like that suggestion or doesn't compile?

Wumpf (approved):

PR #10174, opened 2025-06-09
LeRobot Dataloader: Mark robot type as optional

oxkitsune:
### What

This marks the `robot_type` as an `Option`, because it can be null


Wumpf (approved):
lgtm, but isn't this changelog material as it may fix lerobot loading bugs?

PR #10178, opened 2025-06-09
Replace our own annexb/nal/sps parsing with `h264reader` crate

Wumpf:
### Related

* loosely related to #7484

### What

We'll soon want to read out more data from SPS & PPS. Time to kick that handwritten code out and replace it with an off the shelf crate so we have to worry less about this.

Tested with video stream prototype and some test mp4s I had around.


crates/utils/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
```diff
@@ -95,6 +93,14 @@ impl From<Error> for crate::decode::Error {
     }
 }
 
+/// In Annex-B before every NAL unit is a NAL start code.
+///
+/// Can also be 3 bytes of 0x00.
```
emilk: Is this really correct? Isn't it `[0, 0, 1]`?

How does a parser know the difference between `[0, 0, 0] + [data starting on 1]` and `[0, 0, 0, 1] + [data]`?

emilk (approved):
less code!

crates/utils/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
Wumpf: > Is this really correct? Isn't it [0, 0, 1]?

oh you're right, this was already wrong in the previous file.

PR #10179, opened 2025-06-09
Fix video frame reference preview selection preview (again)

Wumpf:
Fixed this recently, but unfortunately regressed again: when clicking on a video with video frame reference it would just play automatically.

These admittedly quite hacky special case data uis are a bit hard to unit test since providing data for it is not entirely straight forward. But we really should :/
I wish this was a clean extension mechanism to our component ui system to cover this in an extensible way.


Wumpf:
merging this without review. it's not a big change and kinda deep in the woods

PR #10180, opened 2025-06-09
Update selection UI snapshot images

emilk:
### Related
* Following https://github.com/rerun-io/rerun/pull/10146
* which merge-raced ahead of https://github.com/rerun-io/rerun/pull/10161


emilk:
Yolo-merging to make main green

PR #10183, opened 2025-06-10
Introduce `ConnectionClient`, an ergonomic wrapper over the frontend gRPC client

abey79:
### What

This PR builds on the various `ext` grpc "clean types" and the Python SDK's `ConnectionHandle` proto-abstraction to introduce `ConnectionClient`, a wrapper over the frontend client that provides ergonomic wrapper over the various end-points using the `ext` types.

Intended uses:
- The redap browser + other viewer facilities (e.g. ingestion of remote recordings, etc.)
- The Python SDK. In this context, the existing `ConnectionHandle` becomes a very thin wrapper over `ConnectionClient` which blocks and convert to SDK types.
- The redap CLI.
- Others?

This PR is an initial take which:
- establishes the pattern
	- in particular, the wrapper is actually generic over the actual `FrontendServiceClient`, such that it can be used regardless of the exact type (which depends on the instrumentation, of which there is more in the redap CLI).
- implements a few endpoints

### Future work
- https://github.com/rerun-io/rerun/issues/10188
- use `ConnectionClient` in the redap cli
- related: port all table providers to use `ConnectionRegistryHandle`

### DIFF
--- a/crates/store/re_grpc_client/src/connection_client.rs
+++ b/crates/store/re_grpc_client/src/connection_client.rs
@@ -145,6 +145,7 @@ where
     }
 
     /// Get a list of partition IDs for the given dataset entry ID.
+    //TODO(ab): is there a way—and a reason—to not collect and instead return a stream?
     pub async fn get_dataset_partition_ids(
         &mut self,
         entry_id: EntryId,


crates/store/re_datafusion/src/datafusion_connector.rs
```diff
@@ -12,11 +12,12 @@ use crate::partition_table::PartitionTableProvider;
 use crate::table_entry_provider::TableEntryTableProvider;
 
 pub struct DataFusionConnector {
-    client: RedapClient,
+    //TODO(ab): this should hold on to a `ConnectionRegistryHandle` instead of a `ConnectionClient`
```
timsaucer: Is this for a later PR or something you want to do before merging?

crates/store/re_datafusion/src/partition_table.rs
```diff
@@ -21,12 +21,13 @@ use crate::wasm_compat::make_future_send;
 
 #[derive(Debug, Clone)]
 pub struct PartitionTableProvider {
-    client: RedapClient,
+    //TODO(ab): this should use a `ConnectionRegistryHandle` instead
```
timsaucer: Same Q as before

crates/store/re_grpc_client/src/connection_client.rs
```diff
@@ -0,0 +1,183 @@
...
+};
+use re_protos::catalog::v1alpha1::{
+    CreateDatasetEntryRequest, DeleteEntryRequest, EntryFilter, FindEntriesRequest,
+    ReadDatasetEntryRequest,
+};
+use re_protos::common::v1alpha1::ext::{IfMissingBehavior, PartitionId, ScanParameters};
+use re_protos::external::prost::bytes::Bytes;
+use re_protos::frontend::v1alpha1::ext::ScanPartitionTableRequest;
+use re_protos::frontend::v1alpha1::frontend_service_client::FrontendServiceClient;
+use re_protos::manifest_registry::v1alpha1::ScanPartitionTableResponse;
```
Wumpf: ni: how about some `use` hygenie and merge/group things?

crates/store/re_grpc_client/src/connection_client.rs
```diff
@@ -0,0 +1,183 @@
...
+
+use crate::StreamError;
+
+/// Expose an ergonomic API over the gRPC redap client.
+///
+/// Implementation note: this type is generic so that it can be used with several client types. This
+/// is useful for e.g. the redap cli, which has a different instrumented type. For the viewer,
```
Wumpf:
```suggestion
/// is useful for e.g. the closed source redap cli, which has a different instrumented type. For the viewer,
```

Is that how we do this? I'm not sure :thinking:

timsaucer (approved):
Overall this is very straightforward and a nice cleanup. There are a bunch of TODO statements, but it looks like those are for a follow on PR.

crates/store/re_grpc_client/src/connection_client.rs
```diff
@@ -0,0 +1,183 @@
...
+    //TODO(#10188): this should disappear once we have wrapper for all endpoints and the client code
+    //is using them.
+    pub fn inner(&mut self) -> &mut FrontendServiceClient<T> {
+        &mut self.0
+    }
+}
+
+// ---
+
+impl<T> GenericConnectionClient<T>
```
Wumpf, line 55: should I be worried that none of its methods are documented?

crates/store/re_datafusion/src/datafusion_connector.rs
abey79: Later PR, but I'm gonna capture this with a proper issue.

crates/store/re_grpc_client/src/connection_client.rs
```diff
@@ -0,0 +1,183 @@
...
+            let record_batch = resp?.data()?.decode()?;
+
+            let partition_id_col = record_batch
+                .column_by_name(COLUMN_NAME)
+                .ok_or_else(|| StreamError::MissingDataframeColumn(COLUMN_NAME.to_owned()))?;
+
+            let partition_id_array =
+                partition_id_col.try_downcast_array_ref::<arrow::array::StringArray>()?;
+
+            partition_ids.extend(
```
Wumpf, line 183: collecting this to a vec means that we're blocking until we're through with gathering all the partitions. Not a problem now, but sounds like something we'd have to fix up later. Any chance we can return an async iterator here?

crates/store/re_datafusion/src/datafusion_connector.rs
abey79: https://github.com/rerun-io/rerun/issues/10191

crates/store/re_grpc_client/src/connection_client.rs
abey79: i dodged it

crates/store/re_grpc_client/src/connection_client.rs
abey79: Yeah we suck at this. Basically it's DIY documented by `frontend.proto`, which is... is very partially documented as well 😅

crates/store/re_grpc_client/src/connection_client.rs
abey79: I tried, but:
1) It's rather obnoxious, with a `Result<impl Stream<Item=Result<PartitionId, StreamError>>, StreamError>` return type
2) The calling code is collecting/blocking anyways.
3) The actual data (just partitions IDs) does not warrant that much effort.

I've added a TODO and intend to punt on this. I fully expect this topic to be revisited as I port more ambitious endpoint (e.g. `get_chunk`), where we'll most definitely want to support streaming.

PR #10189, opened 2025-06-10
Enable `VideoStream` H264 on the web

Wumpf:
### Related

* important piece of https://github.com/rerun-io/rerun/issues/7484

### What

In order to enable streaming on the web we have to extract a fully qualified codec descriptor string, as [specified by the webcodecs codec registry](https://www.w3.org/TR/webcodecs-codec-registry/#video-codec-registry)

Doing so means that we have to even deeper inspect the video bistream and extract this information on-the-fly.
Setting the stage:
* for mp4, this information is readily available as part of the STDS box which in turn for HEVC/AVC contains the AVCC box
    * presence of AVCC box implies that the bitstream is in AVCC (also known as HVCC  for HEVC, but essentially the same) rather than Annex B
         * WebCodec can handle both, phew!
* we _already_ startd doing some inspection of the bitstream, namely we extract group-of-pictures (GOP)
    * for AVC/h264 every GOP starting sample (== `is_sync`) is required to contain an IDR frame AND an  Sequence Parameter Set (SPS)
    * SPS is sufficient for getting the codec string and has other juicy information we'd like to have around!

Therefore, this PR:
* ...further evolves `VideoDataDescription` to not be as little reliant on AVCC, splitting out the optional `VideoEncodingDetails` (Q: better name? it's not like we're encoding something, it's more like some loosy goose information on _what_ got encoded _how_)
* ...enhances GOP detection to produce `VideoEncodingDetails`
    * a side effect is that we have now more powerful parsing for  `VideoEncodingDetails` and have more things available more often \o/. Let's make use of that in to-be-added ui!
* ...resets decoders whenever `VideoEncodingDetails` changes
    * we regard changes to `VideoEncodingDetails` over time as a bug and will always use the latest found `VideoEncodingDetails`

All this causes some shuffling around. The fact that I changed namespaces a bit didn't help either, but the arbitrary split between `decode` and `demux` didn't age well. I'll clean this up further in a follow-up.

---

Sideffect, more information from MP4 with H.264:

Before:
image

After:
image

### DIFF
--- a/crates/store/re_types/definitions/rerun/components/video_codec.fbs
+++ b/crates/store/re_types/definitions/rerun/components/video_codec.fbs
@@ -31,21 +31,20 @@ enum VideoCodec: ubyte ( // TODO(#7484): use fourcc values. Requires us to suppo
     /// Key frames (IRAP) require inclusion of a SPS (Sequence Parameter Set)
     // TODO(andreas): foxglove `CompressedVideo` mentions PPS (picture parameter set) and VPS (video parameter set) being required for IRAP as well.
     // need to confirm if we need those as well (they exist in h264 as well but we didn't need them so far?)
+    // TODO(#10185): Add support for H265.
     //H265,
 
     /// AOMedia Video 1 (AV1)
     ///
     /// See <https://en.wikipedia.org/wiki/AV1>
+    // TODO(#10184): Add support for AV1.
     //AV1,
 
-    /// VP8
-    ///
-    /// See <https://en.wikipedia.org/wiki/VP8>
-    //VP8,
 
     /// VP9
     ///
     /// See <https://en.wikipedia.org/wiki/VP9>
+    // TODO(#10186): Add support for VP8.
     //VP9,
 }
 
--- a/crates/utils/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
+++ b/crates/utils/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
@@ -15,16 +15,16 @@ use ffmpeg_sidecar::{
     command::FfmpegCommand,
     event::{FfmpegEvent, LogLevel},
 };
-use h264_reader::nal::{UnitType, sps};
+use h264_reader::nal::UnitType;
 use parking_lot::Mutex;
 
 use crate::{
-    PixelFormat, Time,
+    PixelFormat, Time, VideoEncodingDetails,
     decode::{
         AsyncDecoder, Chunk, DecodeError, Frame, FrameContent, FrameInfo, OutputCallback,
-        YuvPixelLayout,
         ffmpeg_h264::{FFMPEG_MINIMUM_VERSION_MAJOR, FFMPEG_MINIMUM_VERSION_MINOR, FFmpegVersion},
     },
+    demux::ChromaSubsamplingModes,
 };
 
 use super::version::FFmpegVersionParseError;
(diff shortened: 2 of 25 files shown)


crates/store/re_types/definitions/rerun/components/video_codec.fbs
```diff
@@ -31,21 +31,20 @@ enum VideoCodec: ubyte ( // TODO(#7484): use fourcc values. Requires us to suppo
...
+    // TODO(#10185): Add support for H265.
     //H265,
 
     /// AOMedia Video 1 (AV1)
     ///
     /// See <https://en.wikipedia.org/wiki/AV1>
+    // TODO(#10184): Add support for AV1.
     //AV1,
 
-    /// VP8
```
Wumpf, line 41: don't think it makes sense to support VP8, which is why it didn't get a ticket

crates/utils/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
```diff
@@ -191,62 +192,34 @@ struct FFmpegProcessAndListener {
     on_output: Arc<Mutex<Option<Arc<OutputCallback>>>>,
 }
 
-fn try_get_sps_from_avcc(avcc: &re_mp4::Avc1Box, debug_name: &str) -> Option<sps::SeqParameterSet> {
```
Wumpf, line 194: no need to look at the avcc in detail here anymore - we do so now agnostically by having a look at the new `VideoEncodingDetails`

crates/utils/re_video/src/decode/gop_detection.rs
```diff
@@ -4,38 +4,78 @@ use h264_reader::{
...
+
+impl Eq for DetectGopStartError {}
+
+/// Result of a successful GOP detection.
+///
+/// I.e. whether a sample is the start of a GOP and if so, encoding details we were able to extract from it.
+#[derive(Default, PartialEq, Eq, Debug)]
+pub enum GopStartDetection {
+    /// The sample is the start of a GOP and encoding details have been extracted.
+    StartOfGop(VideoEncodingDetails),
```
Wumpf, line 43: in theory we can also detect a gop start but fail to extract those details. Went there at first but got needlessly complex for what effectively has to be a broken gop

crates/utils/re_video/src/decode/gop_detection.rs
```diff
@@ -75,22 +129,43 @@ fn is_annexb_sample_start_of_gop(mut sample_data: &[u8]) -> bool {
 
         reader.push(&sample_data[..chunk_size]);
 
-        if reader.nal_handler_ref().detected_gop() {
-            return true;
+        // In case of SPS parsing failure keep going.
+        // It's unlikely, but maybe there's another SPS in the chunk that succeeds parsing.
```
Wumpf, line 133: ... tbh I don't even now what that means if that would happen. But with everything video I generously assume everything can happen...

crates/utils/re_video/src/decode/gop_detection.rs
```diff
@@ -100,21 +175,47 @@ mod test {
             0x00, 0x00, 0x00, 0x01, 0x65, 0x88, 0x84, 0x21, 0x43, 0x02, 0x4C, 0x82, 0x54, 0x2B,
             0x8F, 0x2C, 0x8C, 0x54, 0x4A, 0x92, 0x54, 0x2B, 0x8F, 0x2C, 0x8C, 0x54, 0x4A, 0x92,
         ];
-        let result = is_annexb_sample_start_of_gop(sample_data);
-        assert!(result);
+        let result = detect_h264_annexb_gop(sample_data);
+        assert_eq!(
+            result,
+            Ok(GopStartDetection::StartOfGop(VideoEncodingDetails {
```
Wumpf, line 181: the fact that this above ai generated bytes actually generated something we can parse makes me really curious about that IDR frame in there 😆

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -16,16 +16,27 @@ use super::{Time, Timescale};
...
     Yuv444,
 
     /// Subsampling in X only.
     Yuv422,
 
     /// Subsampling in both X and Y.
     Yuv420,
+
+    /// No subsampling at all, since the format is monochrome.
+    Monochrome,
```
Wumpf: merging `monochrome` into the `ChromaSubsamplingModes` made a bunch more things saner!

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -109,8 +121,13 @@ pub struct VideoDataDescription {
     /// The codec used to encode the video.
     pub codec: VideoCodec,
 
-    /// Encoded width & height if known.
-    pub coded_dimensions: Option<[u16; 2]>,
+    /// Various information about how the video was encoded.
+    ///
+    /// Should any of this change during the lifetime of a decoder, it has to be reset.
```
Wumpf, line 127: another piece of loose contract on the mutability of `VideoDataDescription`. Makes me rather uneasy, but it's workable

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -287,93 +333,6 @@ impl VideoDataDescription {
         self.samples.num_elements()
     }
 
-    /// Returns the subsampling mode of the video.
-    ///
-    /// Returns None if not detected or unknown.
-    pub fn subsampling_mode(&self) -> Option<ChromaSubsamplingModes> {
```
Wumpf, line 293: this parsing code moved into `mp4.rs` where it belongs

crates/utils/re_video/src/demux/mod.rs
emilk: nit/style: I always put the simplest state ("off/none/monochrome") at the top of my enums

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -34,6 +45,7 @@ impl std::fmt::Display for ChromaSubsamplingModes {
             Self::Yuv444 => write!(f, "4:4:4"),
             Self::Yuv422 => write!(f, "4:2:2"),
             Self::Yuv420 => write!(f, "4:2:0"),
+            Self::Monochrome => write!(f, "monochrome"),
```
emilk: To be clear - is this different from 4:0:0?

crates/utils/re_video/src/demux/mp4.rs
```diff
@@ -158,3 +160,102 @@ fn unknown_codec_fourcc(mp4: &re_mp4::Mp4, track: &re_mp4::Track) -> re_mp4::Fou
...
+                match (
+                    av01_box.av1c.chroma_subsampling_x != 0,
+                    av01_box.av1c.chroma_subsampling_y != 0,
+                ) {
+                    (true, true) => Some(ChromaSubsamplingModes::Yuv420), // May also be monochrome.
+                    (true, false) => Some(ChromaSubsamplingModes::Yuv422),
+                    (false, true) => None, // Downsampling in Y but not in X is unheard of!
+                    // Either that or monochrome.
+                    // See https://aomediacodec.github.io/av1-spec/av1-spec.pdf#page=131
+                    (false, false) => Some(ChromaSubsamplingModes::Yuv444),
```
emilk: "May also be monochrome" etc - do we still end up in this branch in that case, or is that already handled by `if av01_box.av1c.monochrome {`?

crates/utils/re_video/src/demux/mod.rs
Wumpf: No, it is 4:0:0. I'll add a comment to clarify
Figured calling it 4:0:0 might be confusing

crates/utils/re_video/src/demux/mp4.rs
Wumpf: Yeah we check monochrome earlier because of that. Clarifying - practically a c&p error.

PR #10196, opened 2025-06-10
Fix URDF join angles

emilk:
This works better for a few new URDF files I found. It also follows what the URDF docs say.

### DIFF
--- a/crates/store/re_data_loader/src/loader_urdf.rs
+++ b/crates/store/re_data_loader/src/loader_urdf.rs
@@ -14,6 +14,7 @@ use re_types::{
     AsComponents, Component as _, ComponentDescriptor, SerializedComponentBatch,
     archetypes::{Asset3D, Transform3D},
     datatypes::Vec3D,
+    external::glam,
 };
 
 use crate::{DataLoader, DataLoaderError, LoadedData};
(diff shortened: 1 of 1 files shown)


crates/store/re_data_loader/src/loader_urdf.rs
```diff
@@ -549,21 +550,5 @@ fn log_geometry(
...
-    let z = cr * cp * sy + sr * sp * cy;
-    let w = cr * cp * cy - sr * sp * sy;
-
-    let norm = (x * x + y * y + z * z + w * w).sqrt();
-    if norm > 0.0 {
-        [x / norm, y / norm, z / norm, w / norm]
-    } else {
-        [0.0, 0.0, 0.0, 1.0]
-    }
+    glam::Quat::from_euler(glam::EulerRot::ZYX, yaw, pitch, roll).to_array()
```
oxkitsune, line 553: The urdf docs say: 

> Represents the rotation around fixed axis: first roll around x, then pitch around y and finally yaw around z. All angles are specified in radians.

The `EulerRot::ZYX` seems a bit weird then right?

oxkitsune (changes requested):
(edit)

crates/store/re_data_loader/src/loader_urdf.rs
emilk: I pass in ZYX and yaw,pitch,roll, and it's applied left-to-right, so that all matches (Z=way, Y=pitch, X=roll)

PR #10187, opened 2025-06-10
Fix panic due to using blocking lock on async executor thread

abey79:
### Related

* Fixes #10166
* Follow up to https://github.com/rerun-io/rerun/pull/10078

### What

Using Tokio's Mutex/RwLock's `blocking_read|write` panics if called within an asynchronous execution context, which happens in some of our viewer customisation example. This PR wraps those calls as per tokio's [doc](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html#panics-1).

PR #10199, opened 2025-06-11
Fix `cargo build -p re_grpc_client` (and main CI)

emilk:
### Related
* Introduced in https://github.com/rerun-io/rerun/pull/10187


emilk:
@rerun-bot full-check

Issue #2116, opened 2023-05-15
Dropped unused python recordings


teh-cmc:
`RecordingStream`s created from Python have their lifetimes managed by the Rust SDK: this is important as we cannot have `Drop` implementations running at arbitrary points in time depending on what the Python GC happens to be doing.

The one downside to this is that these recordings are never really dropped, they are merely disconnected, which means e.g. their batching thread is still running for no reason.

In 99.9% of cases this is completely irrelevant since all these threads do is sleeping, but still it'd be nice to clean up at some point.

PR #10201, opened 2025-06-11
Fix initializing two recordings with the same recording id causing SDK hangs

### LINKED ISSUE #9948: Initializing twice with same recording id hangs the SDK
Repeated `init`calls will cause the SDK to hang on exit:
```py
import rerun as rr

rr.init("app1", spawn=True)
rr.log("point", rr.Points3D([1, 2, 3], colors=[255, 0, 0]))

rr.init("app1", spawn=True)
rr.log("point", rr.Points3D([1, 2, 3], colors=[0, 255, 0]))
```

Note that as per documentation this should log to the same recording stream since the recording id is generated once. This works as documented with explicit recording streams:

```py
import rerun as rr

with rr.RecordingStream("app1") as rec:
    rec.spawn()
    rec.log("point", rr.Points3D([1, 2, 3], colors=[255, 0, 0]))

with rr.RecordingStream("app1") as rec:
    rec.spawn()
    rec.log("point", rr.Points3D([1, 2, 3], colors=[0, 255, 0]))
```

Curiously, when using `rr.init` more "realistically" with different recording ids, we don't get a hang on shutdown:
```py
from uuid import uuid4
import rerun as rr

rr.init("app1", recording_id=uuid4(), spawn=True)
rr.log("point", rr.Points3D([1, 2, 3], colors=[255, 0, 0]))

rr.init("app1", recording_id=uuid4(), spawn=True)
rr.log("point", rr.Points3D([1, 2, 3], colors=[0, 255, 0]))
```

Tested on 0.23.2

AhmedMousa-ag:
### Related

* Closes #9948

### What

* The hang in the issue #9948 is caused when trying to insert duplicated recording, and since we don't need this behavior, just avoid inserting the new recording. The real reason for hang not known yet, but it's not replicated in PyTests.

* Moved the py_rerun_warn function from `rerun_py/src/dataframe.rs` to `rerun_py/src/utils.rs`  for readability.

* Added a new function that does both logging the warning using `re_log::warn` macro and `PyErr::warn` to detect the warning in python as using `re_log::warn` macro alone isn't enough to detect the warning in python, and using PyErr::warn doesn't log the warning as per your code convention. I suggest to use this function for other warnings in the python bridge

* Added a unit test for detecting the warning of duplicated recording stream. Unforntually the hanging behavior can't be replicated in PyTests, still not sure why.


rerun_py/src/python_bridge.rs
```diff
@@ -238,6 +238,18 @@ fn new_recording(
         default_store_id(py, StoreKind::Recording, &application_id)
     };
 
+    // NOTE: The Rust-side of the bindings must be in control of the lifetimes of the recordings!
+    // Check if recording already exists first.
+    // Put it in this scop to drop the mutex as early as possible as all_recordings function returns a mutex guard, and will be called again at the end of this function.
```
Wumpf:
```suggestion
    // Note that this is scoped in order to release the Mutex locked by `all_recordings` as soon as possible
```

rerun_py/src/python_bridge.rs
```diff
@@ -238,6 +238,18 @@ fn new_recording(
...
+    if let Some(existing_recording) = all_recordings().get(&recording_id) {
+        let _=utils::py_log_warn(
+            format!("Recording with id: {} already exists, will ignore creation and return existing recording.",
+            &recording_id).as_str()
+        );
+        // py_rerun_warn(c"Recording Stream already exists, will ignore creation and return existing recording.");
```
Wumpf: redundant?
```suggestion

```

rerun_py/src/python_bridge.rs
```diff
@@ -238,6 +238,18 @@ fn new_recording(
...
     };
 
+    // NOTE: The Rust-side of the bindings must be in control of the lifetimes of the recordings!
+    // Check if recording already exists first.
+    // Put it in this scop to drop the mutex as early as possible as all_recordings function returns a mutex guard, and will be called again at the end of this function.
+    if let Some(existing_recording) = all_recordings().get(&recording_id) {
+        let _=utils::py_log_warn(
+            format!("Recording with id: {} already exists, will ignore creation and return existing recording.",
+            &recording_id).as_str()
+        );
```
Wumpf: failure to emit the warning actually hints at a serious problem, should not ignore this result

```suggestion
        utils::py_log_warn(
            format!("Recording with id: {} already exists, will ignore creation and return existing recording.",
            &recording_id).as_str()
        )?;
```

rerun_py/src/utils.rs
```diff
@@ -23,3 +23,26 @@ where
...
+
+pub fn py_log_warn(msg: &str) -> PyResult<()> {
+    re_log::warn!(msg);
+    let csmsg;
+    let cmsg = match CString::new(msg) {
+        Ok(csm_temp) => {
+            csmsg = csm_temp;
+            csmsg.as_c_str()
+        }
+        _ => c"Warning: :Failed to get warning message.",
```
Wumpf: if we'd ever hit this this is very unlikely to be caught. I think it would be better to just map the error to `PyResult` and return

rerun_py/src/utils.rs
```diff
@@ -23,3 +23,26 @@ where
...
+    Python::with_gil(|py| {
+        let warning_type = PyModule::import(py, "rerun")?
+            .getattr("error_utils")?
+            .getattr("RerunWarning")?;
+        PyErr::warn(py, &warning_type, msg, 0)?;
+        Ok(())
+    })
+}
+
+pub fn py_log_warn(msg: &str) -> PyResult<()> {
```
Wumpf: the naming distinction of those two functions is totally intransparent to me. Let's call them both `py_rerun_warn` and rename the cstr one to `py_rerun_warn_cstr`

rerun_py/tests/unit/test_expected_warnings.py
```diff
@@ -50,3 +50,14 @@ def test_expected_warnings() -> None:
...
         "Must provide one of principal_point, resolution, or width/height)",
     )
+    expect_warning(
+        _duplicated_recording,
+        "already exists, will ignore creation and return existing recording.",
+    )
+
+
+def _duplicated_recording() -> None:
+    """A special case of creating duplicated streams and logging information, it causes a hang, this test is for this case."""
```
Wumpf:
```suggestion
    """Regression test for #9948: creating the same recording twice caused hangs in the past (should instead warn)"""
```

rerun_py/tests/unit/test_expected_warnings.py
```diff
@@ -50,3 +50,14 @@ def test_expected_warnings() -> None:
...
+    application_id = "test_app_id"
+    rr.init(application_id)
+    rr.init(application_id)
```
Wumpf: should inspect the recording before/after and verify that it got the same id

rerun_py/tests/unit/test_expected_warnings.py
```diff
@@ -50,3 +50,14 @@ def test_expected_warnings() -> None:
         lambda: rr.log("world/image", rr.Pinhole(focal_length=3)),
         "Must provide one of principal_point, resolution, or width/height)",
     )
+    expect_warning(
+        _duplicated_recording,
+        "already exists, will ignore creation and return existing recording.",
+    )
+
+
+def _duplicated_recording() -> None:
```
Wumpf:
```suggestion
def test_init_twice() -> None:
```

rerun_py/tests/unit/test_expected_warnings.py
Wumpf: should test the actually expected warning here like the other tests in this file, otherwise this is not testing anything

Wumpf (changes requested):
thanks for looking into this! Needs some more polish though

rerun_py/src/python_bridge.rs
AhmedMousa-ag: The reason for possible error is that according to `pyo3` docs that "May return an Err(PyErr) if warnings-as-errors is enabled". I assumed you wouldn't like to stop the functionality or rerun because of a failure of ommiting warnings.
Noted.

rerun_py/src/python_bridge.rs
AhmedMousa-ag: True, sorry missed removing the comment.

rerun_py/src/utils.rs
AhmedMousa-ag: `py_rerun_warn` is function was already there, I just moved it to utils.
Noted will rename.

rerun_py/src/utils.rs
AhmedMousa-ag: Noted

rerun_py/tests/unit/test_expected_warnings.py
AhmedMousa-ag: I followed the same coding pattern in this file by putting all warning tests at function `test_expected_warnings`. 
Will change it as per suggestion into a seperate function.

rerun_py/src/utils.rs
AhmedMousa-ag: There's also a reason for `py_log_warn`, as it logs the warning into the rerun logging system using `re_log::warn` , without it the warning message isn't logged, also using it alone is not enough to ommit the warning on the python run time. therefore I needed both.  Hence the naming for `log` word

AhmedMousa-ag: I will add a line documenting their purposes to avoid confusion

Wumpf: doesn't that mean the user sees the warning twice? sounds pretty bad

AhmedMousa-ag: Only once, unless the warnings are catched and displayed on the python side.
I believe you might want to inititate a discussion on the practices of this area, the code doesn't have a conistancy of dealing with it and there're around 5 warnings on the rust side only one of them can be catched on the python side. All other warnings are on the python side (I didn't check if the python warnings follow the same pattern of the rust warnings or not).

rerun_py/tests/unit/test_expected_warnings.py
Wumpf: the old method was all about expected warnings from log. the new one is about init. which is arguably means that it should go elsewhere, but whatever

but you didn't actually follow the pattern here either way - the pattern is using a utility method to check for the expected warning. Don't understand why you're not using that here as well?

you're now testing `rr.init` with explicit recording ids. Which is also interesting to test, but would be nice to still test the default parameter as well. I think there's a getter on the recording for retrieving it?

AhmedMousa-ag: `rr.init` is function and returns None, in order to get the returned id, would have to return the Recording at the end of `init` function, I also believe it will be deprecated soon, not sure.
[It's in this file](https://github.com/rerun-io/rerun/blob/f2ff108a397cff77e049ecbe259b9d6534dc9fdd/rerun_py/rerun_sdk/rerun/__init__.py#L314)

rerun_py/src/utils.rs
Wumpf: tbh no idea about the rest, and yes that's out of scope. But practically speaking if I'm in the SDK and make the dupl-init mistake do I see the warning pop up twice on the console? If no then that's fine, but I'm very much confused to why not

rerun_py/tests/unit/test_expected_warnings.py
Wumpf: I haven't tried myself tbh, so maybe this leads nowhere, but I thought something like `get_global_data_recording().get_recording_id()` should work

Wumpf: I can also take this over, it's a lot of context and I don't mean to push you from one detail to the next. Thank you for the work you're putting in already!

AhmedMousa-ag: I think your suggestion would work, I didn't pay attention to their existance

AhmedMousa-ag: But it doesn't check for the returned recording.

AhmedMousa-ag: Done

AhmedMousa-ag:
Which pixi version are you using? it says in the toml file you're using `0.34.0` but the pixi.lock refer to a higher version.

 
> Running python tests with `pixi run py-test` shows that a lot of tests are now broken because of the emitted warning. Haven't looked through the exact cause on all of them whether the warning should be handled, or strict mode is leaking etc. but this clearly needs some attention

AhmedMousa-ag:
I found that returning the warning result to python caused multipile errors on the python side, I will ignore it as it was. I believe it's enough to see the warning on in terminal, logs and be known in python side. instead of raising as an error and hault the whole process.

Wumpf:
I really don't want to create precedence in ignoring warnings. This means that `strict_mode` no longer works as expected everywhere.
I'm looking into fixing up those tests instead

rerun_py/src/utils.rs
Wumpf: I checked and it _does_ show up twice, so let's not do that (I'm taking care of it together with other things)

Wumpf:
yikes. Turns out we never actually delete rust sided recordings ever. I.e. this:
```py
        with rr.RecordingStream("rerun_example_test_recording") as rec:
            pass
```
will leak a recording on the rust side 😱 

Looking into this as well now separately

Wumpf:
the leaking recordings is actually a known issue 😬 
* https://github.com/rerun-io/rerun/issues/2116

tried to tackle it, but I need help from @abey79 for that. Wasn't too hard to get things to work out without. So this should do now if ci passes.

Wumpf (approved):

AhmedMousa-ag:
Why not to delete the recording on `__del__()` destructor when the recording stream is deleted? [Here](https://github.com/rerun-io/rerun/blob/ae4c7e2965c353ddd495fcdf6feb92e83440f6a3/rerun_py/rerun_sdk/rerun/recording_stream.py#L472) 

The implementation in rust and in python should be straight forward.

```rust
/// Delete a recording stream by id.
#[allow(clippy::fn_params_excessive_bools)]
#[allow(clippy::struct_excessive_bools)]
#[pyfunction]
#[pyo3(signature = (
    recording_id,
))]
fn delete_recording(_py: Python<'_>, recording_id: String) -> PyResult<()> {
    if all_recordings()
        .remove(&StoreId::from_string(
            StoreKind::Recording,
            recording_id.clone(),
        ))
        .is_none()
    {
        utils::py_rerun_warn(format!("Recording id: {recording_id} not found.").as_str())?;
    };
    Ok(())
}
```


> the leaking recordings is actually a known issue 😬
> 
> * [Drop unused python recording streams #2116](https://github.com/rerun-io/rerun/issues/2116)
> 
> tried to tackle it, but I need help from @abey79 for that. Wasn't too hard to get things to work out without. So this should do now if ci passes.

Wumpf:
I think that's what we going end up doing in some form or capactiy but I found untangling the ramifications rather complicated. See:

* https://github.com/rerun-io/rerun/blob/ae4c7e2965c353ddd495fcdf6feb92e83440f6a3/rerun_py/src/python_bridge.rs#L47
* https://github.com/rerun-io/rerun/blob/ae4c7e2965c353ddd495fcdf6feb92e83440f6a3/rerun_py/src/python_bridge.rs#L334
* https://github.com/rerun-io/rerun/blob/ae4c7e2965c353ddd495fcdf6feb92e83440f6a3/rerun_py/src/python_bridge.rs#L465

Recording streams are internally already refcounted but we have this `all_recordings` registry on top and the Python object that wraps the recording is rather confusing as well: removing from the `all_recordings` registry doesn't do the actual drop which (see above comment links) is "dangerous" if done in the wrong spot, therefore the "real" drop of the inner recording stream (the one wrapped by the recording) happens at another point in time we don't control 😵‍💫 
We might first need to make the python objects hold handles rather than "real" recording streams

AhmedMousa-ag:
Ok, so we don't know what the grabage collector doing at any time, but can't we invoke the garabge collection in the destructor? `gc.collect()`.
Usually invoking the garbage collection should be avoided, but is it worth it? will it solve the issue?

Wumpf:
There's no gurarantees what-so-ever on what `gc.collect` exactly does, it's more of a hint

PR #10206, opened 2025-06-11
Add missing word in entity path docs page

vegarsti:
Noticed this while reading the docs. I'm not sure this verb is exactly right, feel free to change. Maybe it should be "escape"?


Wumpf (approved):
ah nice catch, thank you

PR #10209, opened 2025-06-11
Tsaucer/bump tower http

timsaucer:
This PR updates a few dependencies, `tonic` and `tower-http`. It requires regenerating the proto files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -990,11 +990,10 @@ dependencies = [
 
 [[package]]
 name = "axum"
-version = "0.7.7"
+version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "504e3947307ac8326a5437504c517c4b56716c9d98fac0028c2acc7ca47d70ae"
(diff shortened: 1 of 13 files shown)


Cargo.lock
```diff
@@ -10328,26 +10321,6 @@ dependencies = [
  "web-sys",
 ]
 
-[[package]]
-name = "tower"
-version = "0.4.13"
```
emilk, line 10333: Looks like we can finally get rid of this awful thing:

https://github.com/rerun-io/rerun/blob/eb30cfb1cd4b1873e3722bd16ae1c756e7d7e072/deny.toml#L72

deny.toml
```diff
@@ -73,6 +73,7 @@ skip-tree = [
   { name = "walkers" },   # TODO(#7876): suppress that when `walkers` is updated
   { name = "windows" },   # Latest sysinfo uses 0.57 max, latest wgpu uses 0.58. See also https://github.com/GuillaumeGomez/sysinfo/pull/1338.
   { name = "zbus" },      # Old version via rfd
+  { name = "rand" },      # DF and object store are on different versions
```
emilk: nit: sort 🙏

PR #10213, opened 2025-06-12
Ensure LeRobot Dataloader logs tagged components

### LINKED ISSUE #9954: LeRobot dataloader uses untagged `Scalar` component

### LINKED ISSUE #10031: Use `ChunkBuilder::with_archetype` in LeRobot dataset loader and write migration guide
The LeRobot loader logged untagged components. These won't be picked up by views anymore in the future.

We need to add a section to the migration guide that explains that people should re-import their LeRobot datasets.

From our discussion: https://rerunio.slack.com/archives/C033K5VS2KD/p1747830851731179?thread_ts=1747826887.692779&cid=C033K5VS2KD

> We have decided not to write migration code for LeRobot datasets, because we cannot tell if a particular chunk is actually from the LeRobot loader, which in turn could break other user's chunks (even in the future).
>
> Our rationale is that when people have a LeRobot dataset, they can just reload it again (the amount of users that manually saved each episode of a dataset to .rrds is probably small).

> [!IMPORTANT]
> We also should ensure that the LeRobot loader properly logs the archetype for each chunk.

oxkitsune:
### Related

* Closes #9954 
* Closes #10031

### What

* Adds a note in the migration guide about `.rrd` files created with the LeRobot dataloader in Rerun versions before `0.24`.
* Sends the `VideoFrameReference` component chunks by constructing an Archetype using the columnar api instead of building the chunk manually.

### DIFF
--- a/crates/store/re_data_loader/src/loader_lerobot.rs
+++ b/crates/store/re_data_loader/src/loader_lerobot.rs
@@ -21,7 +21,6 @@ use re_chunk::{
 };
 use re_log_types::{ApplicationId, StoreId};
 use re_types::{
-    Archetype, ComponentBatch,
     archetypes::{
         self, AssetVideo, DepthImage, EncodedImage, Scalars, TextDocument, VideoFrameReference,
     },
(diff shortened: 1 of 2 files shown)


grtlr (approved):
Thanks! LGTM

PR #10215, opened 2025-06-12
Use new column name in `blueprint::datatypes::ComponentColumnSelector`

### LINKED ISSUE #10129: Change `blueprint.datatypes.ComponentColumnSelector` to use column name
Once we have settled on a specification for column names, we need to update our blueprints to use this new format.

grtlr:
### Related

* Closes #10129.
* Part of #6889.
* Part of #10065.

### What

This changes the blueprint logic to use the new `<entity_path>:[<archetype_name>]:<archetype_field_name>` representation.


crates/store/re_types/src/blueprint/components/component_column_selector_ext.rs
```diff
@@ -2,7 +2,7 @@ use re_log_types::EntityPath;
 
 impl super::ComponentColumnSelector {
     /// Create a [`Self`] from an [`EntityPath`] and a [`re_types_core::ComponentName`] expressed as string.
-    pub fn new(entity_path: &EntityPath, component_name: &str) -> Self {
+    pub fn new(entity_path: &EntityPath, component_name: String) -> Self {
```
Wumpf, line 5: small question & suggestion: do we want to the parameters to be like this in the future? (fwiw `blueprint::datatypes::ComponentColumnSelector`'s field is called `component` already)
```suggestion
    pub fn new(entity_path: &EntityPath, component: String) -> Self {
        crate::blueprint::datatypes::ComponentColumnSelector::new(entity_path, component)
```

arguably in this case `column_name` would be even more more accurate?

crates/store/re_types/definitions/rerun/blueprint/datatypes/component_column_selector.fbs
```diff
@@ -14,6 +13,10 @@ table ComponentColumnSelector (
   entity_path: rerun.datatypes.EntityPath (order: 100);
 
   /// The name of the component.
+  ///
+  /// If the component does not have an associated `ArchetypeName`, this is just a plain field name. Otherwise,
+  /// this is a combination of `<ArchetypeName>:<ArchetypeFieldName>`. An example for this would be
+  /// `Points3D:positions`, for the `positions` field in the `Points3D` archetype.
   // NOTE: this field cannot be named `component_name`, because it would then clash with
   // `ComponentMixin.component_name()` in the Python API.
   component: rerun.datatypes.Utf8 (order: 200);
```
Wumpf: As per our discussion with Jeremy this is true for Rerun builtins, but more general this is just the column name.

Would it be better to write it like this:


```suggestion
  /// The name of the component.
  ///
  /// This acts as the column name in the context of a given `entity_path`
  /// By convention, for all Rerun typtes this is a combination of `<ArchetypeName>:<ArchetypeFieldName>`.
  /// (if there's no `ArchetypeName` it is just the field name).
  /// An example for this would be `Points3D:positions`, for the `positions` field in the `Points3D` archetype.
  // NOTE: this field cannot be named `component_name`, because it would then clash with
  // `ComponentMixin.component_name()` in the Python API.
  component: rerun.datatypes.Utf8 (order: 200);
```

crates/store/re_types/src/blueprint/datatypes/component_column_selector_ext.rs
```diff
@@ -1,19 +1,34 @@
...
 
 impl super::ComponentColumnSelector {
-    /// Create a [`Self`] from an [`EntityPath`] and a [`re_types_core::ComponentName`] expressed as string.
-    pub fn new(entity_path: &EntityPath, component_name: &str) -> Self {
+    /// Create a [`Self`] from an [`EntityPath`] and a column name.
+    pub fn new(entity_path: &EntityPath, column_name: String) -> Self {
         Self {
             entity_path: entity_path.into(),
-            component: component_name.into(),
+            component: column_name.into(),
```
Wumpf: curious inconsistency. Should we rename the field to `column_name`? related to comment on fbs

docs/content/reference/migration/migration-0-24.md
```diff
@@ -63,3 +63,18 @@ Now instead they are scaled relative to the entity's center.
...
+```
+<entity_path>:[<archetype_name>]:<archetype_field_name>
+```
+
+Note that the `archetype_name` section is optional, because components can also be logged as plain fields.
+
+### Blueprints
+
+We have updated our logic so that the `component` field of `blueprint.datatypes.ComponentColumnSelector` follows the same schema.
+
```
Wumpf: nice!

Wumpf (approved):
some questions around naming metaphysics, but otherwise great small step!

crates/store/re_types/src/blueprint/components/component_column_selector_ext.rs
grtlr: Ah good catch, guess I should stop and think from time to time when doing a mechanic refactor 😅.

grtlr:
Merging because the only failing tests are unrelated link checks.

PR #10217, opened 2025-06-12
Wrap `PyCatalogClient` in a Python `CatalogClient`

abey79:
### What

This establishes the datafusion pattern of wrapping "internal" rust-implemented objects with a pure-python one, to cleanly expose the API and facilitate the implementation of pythonic API.

To make all the linting happy, I opted to triplicate the object definition:
- Once in rust, for the internal object.
- `.pyi` stubs for the rust internal object.
- Python implementation of the public object.

This is not ideal, but will hopefully be solved by auto-generating the stubs. Also, I made the process 83% less obnoxious by _not_ enforcing docstring for internal object.


rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -11,3 +12,55 @@
...
+
+    def get_table(self, *, id: EntryId | str | None = None, name: str | None = None) -> Table:
+        """Returns a table by its ID or name."""
+
+        return self._raw_client.get_table(self._resolve_name_or_id(id, name))
+
+    def entries_table(self) -> Table:
+        """Returns a table containing all entries in the catalog."""
+
+        return self.get_table(name="__entries")
```
abey79, line 46: We have a winner here.

rerun_py/src/catalog/catalog_client.rs
```diff
@@ -194,59 +194,3 @@ impl PyCatalogClient {
...
-                    py,
-                    PyEntryId {
-                        id: entry_details[0].id,
-                    },
-                )
-            }
-            Self::Id(id) => Ok(id),
-        }
-    }
-}
```
abey79, line 252: Another winner.

rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -11,3 +13,59 @@
...
+    def _resolve_name_or_id(self, id: EntryId | str | None = None, name: str | None = None) -> EntryId:
+        """Helper method to resolve either ID or name. Returns the id or throw an error."""
+
+        # TODO(ab): this screams for a `match` statement in Python 3.10+
+        if id is not None and name is not None:
+            raise ValueError("Only one of 'id' or 'name' must be provided.")
+        elif id is not None:
+            if isinstance(id, str):
+                id = EntryId(id)
+            return id
```
jleibs: This is the exact kind of grey-area code that made me want to write less stuff in python and more stuff in rust.

The type-hinting will suggest `id: EntryId | str`, but it's python so :man_shrugging: 

So any non-str type passed into resolve will come out again. I guess that's fine in this case since the helper class should now (hopefully) throw an exception if the input isn't actually an EntryId, but I have to go check all the users of `_resolve_name_or_id` to be sure.

rerun_py/src/catalog/catalog_client.rs
```diff
@@ -1,19 +1,16 @@
...
 };
 
-use re_log_types::EntryId;
 use re_protos::catalog::v1alpha1::EntryFilter;
 
 use crate::catalog::{ConnectionHandle, PyDataset, PyEntry, PyEntryId, PyTable, to_py_err};
 
 /// Client for a remote Rerun catalog server.
-#[pyclass(name = "CatalogClient")]
+#[pyclass(name = "CatalogClientInternal")]
```
jleibs, line 13: Internal on the pyclass helps, but we should probably also name the struct types differently to make Public vs Non-public types clear to developers..

rerun_py/rerun_sdk/rerun/catalog.py
abey79: This was some bad python—I fixed it. Also, i don't think rust helps a lot with these dynamic typing shenanigans either. You'll get an exception regardless.

abey79:
> I'm still not totally sold

Wait for the follow-up PR :) This is not coming out of the blue. It addresses real problems.

rerun_py/src/catalog/catalog_client.rs
abey79: Yeah I've been hesitating. Datafusion is not doing that, eg. they have `PyThing` in rust, and `Thing` in python delegating to `ThingInternal`. But they do have wrapper for every  single object.

rerun_py/rerun_sdk/rerun/catalog.py
jleibs: > i don't think rust helps a lot with these dynamic typing shenanigans either.

Yeah, it's not magic, but rust still gives me higher confidence that an error will be handled deterministically and in a way that someone has thought about.

> You'll get an exception regardless.

Yeah, this particular case works out, but other cases could be the difference between an exception getting thrown or handled correctly.

jleibs:
> Wait for the follow-up PR :) 

I agree -- the rust-calling-python stuff is gross and we need a solution for that regardless.  I'm sure we'll work our way to a pattern and some helpers that find a nice balance in this next phase of our rust+python journey.

rerun_py/src/catalog/catalog_client.rs
jleibs: >they do have wrapper for every single object.

I think that's the main difference. If we're going to mix wrapped and unwrapped types I want to know which is which easily. We may even want to split them into separate `.rs` files.

rerun_py/rerun_sdk/rerun/catalog.py
abey79: `thiserror` and `to_py_err` would like to have a word with you :)

Seriously though, that particular situation is far from ideal on the rust side. It takes nearly as much effort to single out specific errors and make sure they are handled/documented "correctly". I'm as guilty as the next guy tbh. 

The only recent instance I recall thinking about it seriously is `Dataset.register`: I really wanted a `TimeoutError` to be raised (and documented it). The actual implementation is essentially implicit (I "trust" `to_py_err`). This could maybe be improved, but end of the day, the only sane way to go about this is testing the interfaces, which applies equally well to rust and python impl.

rerun_py/src/catalog/catalog_client.rs
abey79: Renamed to `PyCatalogClientInternal`.

PR #10218, opened 2025-06-12
Smooth(er) video stream playback: make video player aware of growing group of pictures

Wumpf:
### Related

* Part of https://github.com/rerun-io/rerun/issues/7484

### What

Makes the video player aware of the fact that groups of pictures may grow.
For 1.5k*1.5k MacOS webcam this gives me very smooth video playback with decent latency in the browser in the ffmpeg streaming test case - no delayed playback needed. Due to the way we implement h264 decoding on native it's not perfectly smooth there, but bearable.

More improvements still to follow but this puts us into a decent spot!

commit by commit!


crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -223,51 +230,125 @@ impl VideoPlayer {
...
+
+        let min_end_sample_idx =
+            requested_sample_idx + self.sample_decoder.min_num_samples_to_enqueue_ahead();
+        loop {
+            let last_enqueued = self
+                .last_enqueued
+                .expect("We ensured that at least one GOP was enqueued.");
+
+            // Enqueued enough samples as described above?
+            if last_enqueued.gop_idx > requested_gop_idx
```
```suggestion
            if last_enqueued.gop_idx >= requested_gop_idx
```

Looks like it should be this, no? If not, why not? And why is the row below different?

Wumpf: as per logic above, we want to be one gop ahead.
But the `end` in `min_end_sample_idx` isn't correct. `end` is exclusive, not what we want here; should be `min_last_sample_idx`

PR #10221, opened 2025-06-13
Major overhaul of `CatalogClient` API

abey79:
### Related

* builds on https://github.com/rerun-io/rerun/pull/10217
* part of https://github.com/rerun-io/dataplatform/issues/750
* part of https://github.com/rerun-io/dataplatform/issues/871
* relevant thread https://linear.app/rerun/issue/DPF-1737/sandbox-datafusion-interfaces#comment-54e705a3
* sibling https://github.com/rerun-io/dataplatform/pull/948

#### Background

We want to address two persona:

- the data person, who wants to go straight to dataframes to get (data) work done;
- the integration/db eng/whatnot person, who is more interested in the overall structure and rarely care about the actual data content.

Our python entry wrapper objects cater to the latter, but are an obstacle to the former.

To address this distinction:
 - The use of the _entry/_entries suffix indicates that we return the Python wrapper object
- Not using that suffix indicates that we directly return (a lazy handle to) the data.
- We currently focus on tables and (non-blueprint) datasets.
- You can still see/access blueprint datasets via the generic entries APIs.

#### New API overview

```python
class CatalogClient:
  def __init__(self, global_ctx: datafusion.SessionContext | None = None):
    self._global_ctx = global_ctx or datafusion.GetGlobalContext()  # or whatever it's called

  # Get materialized lists of Python entry wrapper objects
  def all_entries(self) -> list[Entry]: ...
  def dataset_entries(self) -> list[DatasetEntry]: ...
  def table_entries(self) -> list[TableEntry]: ...

  # Get materilized list of entry names
  def entry_names(self) -> list[str]: ...
  def dataset_names(self) -> list[str]: ...
  def table_names(self) -> list[str]: ...

  # Returns DF dataframes (reminder: these are lazy)
  # Implies the registration of a table provider for `__entries` 
  def entries(self) -> datafusion.DataFrame: ...
  def datasets(self) -> datafusion.DataFrame: ...
  def tables(self) -> datafusion.DataFrame: ...

  # Get single Python entry wrapper object
  def get_dataset_entry(self, name_or_id) -> DatasetEntry: ...
  def get_table_entry(self, name_or_id) -> TableEntry: ... # This also acts as table provider

  # Get the (data) content of an entry
  def get_dataset(self, name_or_id) -> DatasetEntry: ... # TODO: should be a new `Dataset`
  def get_table(self, name_or_id) -> datafusion.DataFrame: ...
```

**IMPORTANT**: this PR severely breaks our current implementation in many ways!

#### Future work

- Tables have a Table object as Python wrapper, and datafusion.DataFrame as data counterpart. This distinction doesn't really exists for datasets, which breaks some of the symmetry. We could split the current Dataset object into a DatasetEntry for the entry stuff (currently: reading/writing blueprint related properties) and a DatasetData for the data stuff (schema, dataframe query, etc.). World balance would ensue.
- We will probably need more facilities for using custom session context, or registering stuff to arbitrary session context. This could be tackled by register_to_context method(s) that we would add as the need arises.


rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -1,60 +1,137 @@
...
+        Note: This is currently an alias for `get_dataset_entry`. In the future, it will return a data-oriented dataset
+        object instead.
         """
+        return self.get_dataset_entry(id=id, name=name)
+
+    def get_table(self, *, id: EntryId | str | None = None, name: str | None = None) -> datafusion.DataFrame:
+        """
+        Returns a table by its ID or name.
+
+        Note: This is currently an alias for `get_table_entry`. In the future, it will return a data-oriented table
```
jleibs:
```suggestion
        Note: This is an alias for `get_table_entry().df()`.
```

rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -1,60 +1,137 @@
...
+    # ---
+
+    def create_dataset(self, name: str) -> DatasetEntry:
+        """Creates a new dataset with the given name."""
+        return self._raw_client.create_dataset(name)
+
+    def entries_table(self) -> TableEntry:
+        """Returns a table containing all entries in the catalog."""
+
+        return self.get_table_entry(name="__entries")
```
jleibs: This seems redundant with `entries()` can this go away?

rerun_py/src/catalog/catalog_client.rs
```diff
@@ -178,11 +225,35 @@ impl PyCatalogClientInternal {
             details: dataset_entry.details,
         };
 
-        let table = PyTable::default();
+        let table = PyTableEntry::default();
```
jleibs, line 228: I can't comment on it but I noticed a pre-existing typo 8 lines above. `dataset_entry` should be `table_entry` there.

jleibs (approved):

rerun_py/rerun_sdk/rerun/catalog.py
abey79: 🥌

PR #10224, opened 2025-06-13
Use egui atoms in help view

lucasmerlin:
### Related

- atomics landed in egui (https://github.com/emilk/egui/pull/5830)

### What

Updates egui to latest master and uses atoms in the help view, removing the icon_text macro since it was basically atomics lite

Seems like the snapshots the only thing that changed is the three dot icon slightly moved. But it seems a bit clearer now, so that's a win.


Wumpf:
Superseeded by
* https://github.com/rerun-io/rerun/pull/10232
?

lucasmerlin:
> Superseeded by
> 
>     * [Fix some actions happening on key _up_ instead of key _down_ #10232](https://github.com/rerun-io/rerun/pull/10232)
>       ?

Yeah, but I did some atomics related changes here, so this should still be merged

PR #10227, opened 2025-06-13
Fix raw component errors due to missing component name

grtlr:
### Related

* Part of #6889.

### What

Tables should work again 😅.


abey79 (approved):
Tested to work, thanks!

PR #10222, opened 2025-06-13
Disable reading/writing timing metadata to/from encoded record batches

Wumpf:
### Related

* Fixes test regression introduced in https://github.com/rerun-io/rerun/pull/10173

### What

Related internal slack discussion
https://rerunio.slack.com/archives/C07T0LFT5BQ/p1749799544492489

* [x] pass full check ci

cc: @emilk

PR #10228, opened 2025-06-13
Re-enable latency measurements

emilk:
### Related
* https://github.com/rerun-io/rerun/pull/10222
* https://github.com/rerun-io/rerun/pull/10173
* https://github.com/rerun-io/rerun/issues/9812

### What
My previous PR added timestamps to the arrow metadata whenever we encoded or decoded arrow IPC. This caused all sorts of test failures.

In this PR, we instead only insert it in two places:
* The gRPC log sink, used by the SDK
* The gRPC client used by the viewer

### How to test
```sh
pixi run rerun-release
```
```sh
cargo run -p log_benchmark --release -- --benchmarks image --connect
```

### TODO
* [x] Run `main` CI

### Future work
* Add timestamp for when the messages hits the proxy server (`re_grpc_server`)
* Or better yet: https://github.com/rerun-io/rerun/issues/10229 🤦


emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

tests/rust/log_benchmark/src/main.rs
```diff
@@ -72,6 +72,11 @@ struct Args {
     /// If enabled, brings up the puffin profiler on startup.
     #[clap(long, default_value = "false")]
     profile: bool,
+
+    /// If true, connect to a running Rerun viewer
+    /// instead of writing to a memory buffer.
+    #[clap(long, default_value = "false")]
+    connect: bool,
```
emilk, line 79: This makes everything a lot easier to test

crates/viewer/re_view_spatial/tests/transform_clamping.rs
```diff
@@ -254,7 +254,7 @@ fn run_view_ui_and_save_snapshot(
                 modifiers: egui::Modifiers::default(),
             });
             harness.run_steps(10);
-            let broken_percent_threshold = 0.0039;
+            let broken_percent_threshold = 0.0045;
```
emilk, line 257: Unrelated flaky test

Wumpf: unexpectedly high, makes me nervous. But I saw the same test just barely passing or failing. I do think there's a real problem though where sometimes the camera center thingy shows and sometimes not
don't want to bother with this either atm though :/. lgtm

teh-cmc (approved):
The general idea of isolating this to the gRPC proxy server and client seems sound 👍

I'm not familiar enough with `re_grpc_client` to be able to tell for certain whether any metadata might possibly end up leaking into Redap and/or RRD files or not "because reasons".
The only way to be sure is to go the Redap repo, depend on this branch, and run the registry tests: `cargo t --all-features -p redap_manifest_registry`. If the snapshots are flaky, then you know something is wrong.
Approved if these tests pass (I would do it myself, but I don't have enough spare disk space to change branches right now 🙃).

emilk:
`cargo t --all-features -p redap_manifest_registry` still passes when pointed at this branch (I ran it twice!)

PR #10230, opened 2025-06-15
Remove usage of `unwrap()` in `re_sdk`

lgingerich:
### Related

* Part of https://github.com/rerun-io/rerun/issues/6330

### What

- Remove allow directive & TODO comment from re_sdk/src/lib.rs
- Remove usage of `unwrap()` and replace with `expect()` and a descriptive message

`cargo clippy -p re_sdk` and `cargo build -p re_sdk` both pass with no warnings

### DIFF
--- a/crates/top/re_sdk/src/spawn.rs
+++ b/crates/top/re_sdk/src/spawn.rs
@@ -77,12 +77,18 @@ impl Default for SpawnOptions {
 impl SpawnOptions {
     /// Resolves the final connect address value.
     pub fn connect_addr(&self) -> std::net::SocketAddr {
-        std::net::SocketAddr::new("127.0.0.1".parse().unwrap(), self.port)
+        std::net::SocketAddr::new(
+            "127.0.0.1".parse().expect("Failed to parse ip address"),
+            self.port,
+        )
     }
 
     /// Resolves the final listen address value.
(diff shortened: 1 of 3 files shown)


crates/top/re_sdk/src/spawn.rs
```diff
@@ -77,12 +77,18 @@ impl Default for SpawnOptions {
 impl SpawnOptions {
     /// Resolves the final connect address value.
     pub fn connect_addr(&self) -> std::net::SocketAddr {
-        std::net::SocketAddr::new("127.0.0.1".parse().unwrap(), self.port)
+        std::net::SocketAddr::new(
+            "127.0.0.1".parse().expect("Failed to parse ip address"),
```
emilk: Using `expect` here is a bit silly - these _can't_ fail, so we're only making the binary bigger unnecessarily. Better to add `#[allow(clippy::unwrap_used)]` selectively here, or use a constructor that cannot fail (e.g. `Ipv4Addr::new(127, 0, 0, 1)`)

emilk (approved):
Thanks!

crates/top/re_sdk/src/spawn.rs
lgingerich: Good feedback — I changed to use the constructor option

Issue #5251, opened 2024-02-22
Support for collada / .dae files


roym899:
**Is your feature request related to a problem? Please describe.**
While otherwise relatively uncommon, COLLADA / `.dae` files are often used for ROS packages, since it's the [recommended format according to ROS 1 docs](https://wiki.ros.org/urdf/XML/link#:~:text=The%20recommended%20format%20for%20best%20texture%20and%20color%20support%20is%20Collada%20.dae%20files.), and still [mentioned in ROS 2 docs](https://docs.ros.org/en/iron/Tutorials/Intermediate/URDF/Building-a-Visual-Robot-Model-with-URDF-from-Scratch.html#:~:text=STL%20is%20fairly%20common%2C%20but%20the%20engine%20also%20supports%20DAE%2C%20which%20can%20have%20its%20own%20color%20data%2C%20meaning%20you%20don%E2%80%99t%20have%20to%20specify%20the%20color/material).

Here are a few examples of robot descriptions that use .dae files: [PR2](https://github.com/PR2/pr2_common/tree/melodic-devel/pr2_description/meshes), [TurtleBot](https://github.com/turtlebot/turtlebot/tree/melodic/turtlebot_description/meshes), [Nao](https://github.com/ros-naoqi/nao_meshes/tree/master), [Husky](https://github.com/husky/husky/tree/noetic-devel/husky_description/meshes), [Spot](https://github.com/heuristicus/spot_ros/tree/master/spot_description/meshes)

**Describe the solution you'd like**
Add `.dae` support to Asset3D.

**Describe alternatives you've considered**
Current workaround is to use another library to load the file and log via Mesh3D.

patrickelectric:
That would be great!

PR #10231, opened 2025-06-15
Support loading ROS paths in URDF loader

### LINKED ISSUE #10182: Support loading ROS paths in URDF loader
### What

The URDF loader introduced in #10105 does not support loading assets from ROS paths yet. 
Ideally we support this without including ROS as a dependency.

oxkitsune:
### Related

* Closes #10182 

### What

This attempts to resolve an absolute path to ROS resources by parsing the `package://` uri and using the available ROS environment variables.

Took a while to set up a ROS environment on my my mac, but managed to get it working, and besides the missing collada support (#5251), this resolves all `.stl` files correctly:
2025-06-15_19-24-26@2x

### DIFF
--- a/crates/store/re_data_loader/src/loader_urdf.rs
+++ b/crates/store/re_data_loader/src/loader_urdf.rs
@@ -1,4 +1,5 @@
 use std::{
+    env,
     path::{Path, PathBuf},
     sync::{Arc, mpsc::Sender},
 };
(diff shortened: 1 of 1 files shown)


crates/store/re_data_loader/src/loader_urdf.rs
```diff
@@ -620,3 +618,80 @@ fn log_geometry(
...
+                resolved_path.display()
+            )
+        })
+    } else {
+        root_dir
+            .map(|root| root.join(resource_path))
+            .and_then(|full_path| {
+                std::fs::read(&full_path)
+                    .with_context(|| format!("Failed to read file: {}", full_path.display()))
+                    .ok()
```
emilk: ⚠️  `.ok()` here throws away the nice error message above!

Usually I find `if let … else` more readable anyhow

emilk (changes requested):
Great! But you messed up the error handling a bit

crates/store/re_data_loader/src/loader_urdf.rs
oxkitsune: Originally this method returned an `Option`, and I forgot this part. Fixed it now :)

Issue #10165, opened 2025-06-08
No repaint on key press?


emilk:
It seems like egui is not repainting when pressing keys.

## Symptom: unresponsive key input
First: Make sure nothing is animating!

Some of our shortcuts only trigger on key _release_ lately. I suspect recent egui activity.

For instance: press ⌘+P for the command palette and notice it only appears when you let go of the ⌘ key, but only on web builds(!). Ctrl-P seems unaffected on web though.

Or select something in the blueprint panel and press down to select the next item: it is selected on key release.
Do the same in the blueprint panel: and it responds on key press.

emilk:
Adding an extra call to `request_repaint` after any (successful) call to `consume_shortcut` seems to resolve the issue, so I suspect the problem is this:

Whenever a key is pressed, egui _should_ repaint _twice_. The first repaint is so that the application can notice the key press (e.g. via `consume_shortcut`), and the second repaint is so that any effect with a frame-delay have a chance to show up. Here is the code in egui that handles this:

https://github.com/emilk/egui/blob/bdbe6558527899331440dd11024c07bcebb3c876/crates/egui/src/context.rs#L178-L180

…but obviously this is somehow broken.

Note that this bug only affects things that appear with a frame-delay, which is why you won't notice this problem everywhere. Tat is: it only affects code on form of "if consume_shortcut(…) { show_something_next_frame }`.

I wonder if this broke with multi-pass, and we didn't notice until now? 🤔  But lowering `max_passes` to 1 does not resolve the problem.

PR #10232, opened 2025-06-15
Fix some actions happening on key _up_ instead of key _down_

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/10165
* Includes https://github.com/emilk/egui/pull/7134
* All changes: https://github.com/emilk/egui/compare/f0abce9bb8591466ce01f741eced5b271d3a4dc1...54fded362dfc82f0ea188a110fc950d9adbb2f13

### Updated snapshot images
For some reason, the `…` more buttons has become slightly sharper. Maybe related to the recent atomics work?


lucasmerlin (approved):
Nice!

I think only the more icon changed because it's not a perfect square: `width="15" height="14"`

PR #10233, opened 2025-06-16
Move the default plot legend to bottom left instead of right

emilk:
I noticed that the labels often cover the new data streaming in from the right, which is usually the most interesting data.

Screenshot 2025-06-15 at 17 43 48


Wumpf (approved):
makes sense

Issue #5589, opened 2024-03-20
The viewer wastes too much vertical space


teh-cmc:
In general, I find that:
- The top bar is way too tall, with an information density dangerously approaching zero.
- The tab bars are also way too tall.
- The collapsed time panel is, once again, also way too tall.
- And of course the collapsed time panel should never turn into two things like it does in evidence 1 (first time I see that one to be fair).

As someone with a long-love for Vim in part because of how much it respects the user's screen estate, this is a really painful one...

Of course all of these get worse on small screens, i.e. iframes and mobiles.

My proposal would be to A) make all these bars much thinner and maybe B) make them responsive based on the current height of the app, with some fixed minimum and maximum bounds.

Evidence 1:
![image](https://github.com/rerun-io/rerun/assets/2910679/d1ad2b06-3d0d-4f81-8b95-47dfaf61c4f0)

Evidence 2:
![image](https://github.com/rerun-io/rerun/assets/2910679/e92f3f85-74e5-4d98-a019-8e62abfc4e9c)

Evidence 3:
![image](https://github.com/rerun-io/rerun/assets/2910679/3273401c-01fe-4a28-894c-9e02ca998dd0)

teh-cmc:
Extra datapoint, a 3x3 grid in vim:
![image](https://github.com/rerun-io/rerun/assets/2910679/1fa94413-86f5-4f3f-98e0-8f41f9be2fd7)

Vim famously strips away all tab bars within the grid itself to save on screen space and information density, with the assumption that you already know what is what just by the looking at the contents anyway.
In doubt, more info about the currently selected cell can be found in the one single status bar at the bottom.

abey79:
#### Title bar

Specifically on mac, the title bar takes _exactly_ the right vertical space in windowed mode, which is the same as any other (standard) window (though they tend to becoming increasingly rare). As soon as it goes full screen, then it's back to using too much space as per your screenshot.

<img width="1284" alt="image" src="https://github.com/rerun-io/rerun/assets/49431240/686d4c0d-0082-4b8d-b3b3-c5a6e4205b50">

#### Tab bar

For the tab bar, I agree, but it's not very many pixels. They can't be thiner than e.g. the `ListItem`. Exhibit:

<img width="518" alt="image" src="https://github.com/rerun-io/rerun/assets/49431240/ce13b4f6-33ed-4208-b856-a7db7d00c0f1">

[...]

PR #10234, opened 2025-06-16
Make plot axis thinner

emilk:
### Related
* https://github.com/emilk/egui_plot/pull/113
* #5589 

### What
We were wasting a lot of vertical (and horizontal!) space. Now we don't.

Also new feature: drag the axis values to zoom in on the axis.

## Before/after
Screenshot 2025-06-15 at 19 19 18

## Before
Screenshot 2025-06-15 at 20 29 17

## After
Screenshot 2025-06-15 at 20 29 40


grtlr (approved):
Nice improvement!

A somewhat related thing I noticed: The horizontal padding between the axis ticks (i.e. the "0" in the snapshot tests) could be a tad larger. Right now it's ~1 pixel with the number being crammed right next to the plot.

The vertical space between the x-axis and the plot looks more like 2-3 pixels.

abey79:
> Nice improvement!
> 
> A somewhat related thing I noticed: The horizontal padding between the axis ticks (i.e. the "0" in the snapshot tests) could be a tad larger. Right now it's ~1 pixel with the number being crammed right next to the plot.
> 
> The vertical space between the x-axis and the plot looks more like 2-3 pixels.

I agree with that. Ideally, the plot content should be left-aligned with view icon (in the view title bar).

emilk:
I'll keep improving those margins today :)

Issue #8538, opened 2024-12-19
Ability to show edges grid on 3D mesh


dimentary:
**Is your feature request related to a problem? Please describe.**
The ability to show edges grid on top of the mesh, currently you could only show vertices on top of the mesh with the "Points3D" visualiser, which is not very informative (also you can't change points color separately from the mesh color):
<img width="607" alt="Image" src="https://github.com/user-attachments/assets/0fc01d59-6da6-4a38-9fc9-0bf452ffb419" />

**Describe the solution you'd like**
A simple toggle that once turned on shows edges on top of the mesh surface, the result should look something like this:
![Image](https://github.com/user-attachments/assets/f18a4d5d-8642-4774-a25e-91fe3006ad17)

Wumpf:
> also you can't change points color separately from the mesh color

this will be soon addressed by tagged components, see also https://github.com/rerun-io/rerun/issues/6889

---

some notes on implementation:

**Datastructures:**
tbd if this should be a separate visualizer/archetype or just be a display option on `Asset3D` & `Mesh3D`. Leaning towards the later for ease of use (also the intricacies of "archetype inheritance", see https://github.com/rerun-io/rerun/issues/8368 might be be out too long).

**Rendering:**
Implementation-wise we won't be able to just simply draw the triangle mesh as wireframe since that's unfortunately not fully portable (see https://wgpu.rs/doc/wgpu/struct.Features.html#associatedconstant.POLYGON_MODE_LINE - this is a native-only wgpu feature). Using our line renderer is an option, but has too much overhead, we'd really like to use the data we already have.
Drawing from vertex+index buffer using vertex pulling (doing random buffer access in vertex & index buffer) is actually quite doable and that would allow us to render the lines _on_ the solid mesh, avoiding issues with lines intersecting the solid mesh when both solid & wire is drawn.
https://webgpu.github.io/webgpu-samples/?sample=wireframe is a nice implementation of just that and allows to switch between two variants, one with separate line geometry (note how this requires a depth bias which can be fiddly to configure) and one using barycentric coordinates.
**But** we can't do that on WebGL - let's check for alternatives & then decide whether it's worth feature-gating this.

julianschoep:
As a temporary work-around you can visualize the mesh connectivity by plotting line strips like so:

```python
rr.log(
    "mesh_grid",
    rr.LineStrips3D(
        mesh.vertices[mesh.faces],
        colors=[0, 0, 0],
        radii=0.0005,
[...]

PR #10238, opened 2025-06-16
Add `FillMode` to `Capsules3D`

oxkitsune:
### Related

* Part of #1361
* Part of #8538 

### What

Adds the `FillMode` component to the `Capsules3D` archetype, now that we can duplicate the `Radius` component for `line_radii`

### DIFF
--- a/crates/store/re_types/definitions/rerun/archetypes/capsules3d.fbs
+++ b/crates/store/re_types/definitions/rerun/archetypes/capsules3d.fbs
@@ -51,20 +51,23 @@ table Capsules3D (
 
   // --- Optional ---
 
-  // TODO(#1361): Add fill_mode component, or whatever succeeds it in case a wireframe
-  // or other alternate rendering is wanted.
+  /// Optional radii for the lines used when the cylinder is rendered as a wireframe.
+  line_radii: [rerun.components.Radius] ("attr.rerun.component_optional", nullable, order: 3000);
+
+  /// Optionally choose whether the cylinders are drawn with lines or solid.
+  fill_mode: rerun.components.FillMode ("attr.rerun.component_optional", nullable, order: 3100);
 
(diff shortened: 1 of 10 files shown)


crates/store/re_types/definitions/rerun/archetypes/capsules3d.fbs
```diff
@@ -51,20 +51,23 @@ table Capsules3D (
 
   // --- Optional ---
 
-  // TODO(#1361): Add fill_mode component, or whatever succeeds it in case a wireframe
-  // or other alternate rendering is wanted.
+  /// Optional radii for the lines used when the cylinder is rendered as a wireframe.
+  line_radii: [rerun.components.Radius] ("attr.rerun.component_optional", nullable, order: 3000);
```
Wumpf, line 55: so good that we can do this now 😄

crates/viewer/re_view_spatial/src/proc_mesh.rs
```diff
@@ -65,6 +65,9 @@ pub enum ProcMeshKey {
         /// The cylinder part of the capsule is approximated as a mesh with (N + 1) × 4
         /// flat faces.
         subdivisions: usize,
+
+        /// If true, then when a wireframe mesh is generated
```
Wumpf:
```suggestion
        /// If true, then when a reduced wireframe mesh, showing the outlines of the shape, is generated.
```
here and below.

Wumpf (approved):
nice! code lgtm

crates/viewer/re_view_spatial/src/proc_mesh.rs
oxkitsune: Copy pasta gone wrong, changed it to:

```
/// If true, wireframe meshes are generated with reduced complexity,
/// showing a minimal set of lines that outline the shape.
```

PR #10239, opened 2025-06-16
Add "deep inspection" selection ui for the new `VideoStream` archetype

Wumpf:
### Related

* part of #7484

### What

Adds UI for `VideoStream`, sharing almost all its code with `VideoAsset` & `VideoFrameReference` :)
image

Note that video asset's properties are also collapsible now:
image

### DIFF
--- a/crates/viewer/re_viewer_context/src/utils/video.rs
+++ b/crates/viewer/re_viewer_context/src/utils/video.rs
@@ -15,9 +15,11 @@ pub fn video_timestamp_component_to_video_time(
     }
 }
 
-pub fn video_time_from_query(query: &re_chunk_store::LatestAtQuery) -> re_video::Time {
-    // Video streams are handled like "infinite" videos both forward and backwards in time.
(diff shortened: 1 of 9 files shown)


crates/viewer/re_viewer_context/src/utils/video.rs
```diff
@@ -14,3 +14,11 @@ pub fn video_timestamp_component_to_video_time(
         re_video::Time((video_timestamp.0.0 as f64 * fps as f64) as i64)
     }
 }
+
+pub fn video_time_from_query(query: &re_chunk_store::LatestAtQuery) -> re_video::Time {
```
emilk: I think we should clairfy in name and/or docstrings that this is only for video streams, not video frame references (I think?)

emilk (approved):
Nice!

crates/viewer/re_viewer_context/src/utils/video.rs
Wumpf: you're right, good point

PR #10245, opened 2025-06-16
Rename `ComponentDescriptor` fields and use colons in Sorbet metadata

### LINKED ISSUE #9978: Rename fields on ComponentDescriptor & make fields unique identifiers

### LINKED ISSUE #10065: Tagged components milestone 4: descriptor-driven dataframe queries and views

grtlr:
### Related

* Part of #6889.
* Closes #9978.
* Closes #10065. 

### What

This PR updates our Sorbet data model, and our code around `ComponentDescriptors` to the decisions we made in #9978.

Most notably are the changes to `rerun.component` and `rerun.component_type`.

#### Mostly mechanical renames:

Code

* `ComponentName` to `ComponentType`
* `component_name` to `component_type`
* "component name" to "component type"
* `archetype_field_name` to `component`
* `archetype_name` to `archetype`
* Changes the `ComponentDescriptor` accordingly

Sorbet

* Changed all metadata fields to use `:` for name spacing.
* `rerun.component` -> `rerun:component_type` 
* `rerun.archetype_field_name` -> `rerun:component` now with new semantics (`ComponentIdentifier`).

#### Logic changes

* We now fulfill dataframe queries directly using the `ComponentIdentifier`.
* Removed remainders of `ComponentLike` interface.
* Migration code in `re_sorbet`.

#### Open questions:

* [x] @abey79 What is `entry_kind` in the Redap browser and does this clash?

### TODO

* [x] full build
* [x] check dataplatform https://github.com/rerun-io/dataplatform/pull/1004


grtlr:
@rerun-bot full-check

crates/store/re_chunk/src/chunk.rs
```diff
@@ -74,12 +74,12 @@ impl ChunkComponents {
     ///
     /// I.e semantically equivalent to `get("MyComponent:*.*")`
     #[inline]
-    pub fn get_by_component_name(
+    pub fn get_by_component_type(
```
Wumpf, line 77: above comment is outdated
might be worth grepping for `component name` (haven't tried myself yet)

Wumpf (approved):

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -914,7 +914,7 @@ def send_arrow_chunk(
     timelines: `Dict[str, arrow::Int64Array]`
         A dictionary mapping timeline names to their values.
     components: `Dict[str, arrow::ListArray]`
-        A dictionary mapping component names to their values.
+        A dictionary mapping component types to their values.
```
Wumpf, line 968: digged into this method and looks like this isn't actually true. It ends up creating descriptors https://github.com/rerun-io/rerun/blob/2f7d27ab9900494ea7d780efd967569038feccde/rerun_py/src/arrow.rs#L137 from fully formed python objects https://github.com/rerun-io/rerun/blob/2f7d27ab9900494ea7d780efd967569038feccde/rerun_py/src/arrow.rs#L27
which implies that the type annotation is also wrong? This is used as a utility for `AnyValues`

so in our

bacon.toml
```diff
@@ -51,9 +51,17 @@ need_stdout = true
...
+  "cargo",
+  "nextest",
+  "run",
+  "--hide-progress-bar",
+  "--all-features",
+  "--failure-output",
+  "final",
+]
 need_stdout = true
-watch = ["tests"]
```
jleibs, line 56: Is this change actually related?

crates/build/re_types_builder/src/codegen/rust/api.rs
```diff
@@ -1032,7 +1032,7 @@ fn quote_trait_impls_for_datatype_or_component(
         quote! {
             impl ::re_types_core::Component for #name {
                 #[inline]
-                fn name() -> ComponentName {
+                fn name() -> ComponentType {
```
jleibs, line 1035: Future PR but `name() -> ComponentType` looks weird. Maybe `type_name()`?

crates/store/re_chunk/src/builder.rs
```diff
@@ -79,9 +79,9 @@ impl ChunkBuilder {
                 .with_row(cell.value);
         }
 
-        for (component_name, array) in components {
+        for (component_type, array) in components {
             self.components
-                .entry(component_name)
+                .entry(component_type)
```
jleibs: This looks like a search/replace error from a prior misnamed thing. `self.components` is indexed by descriptor, not by name or type.

crates/store/re_chunk/src/chunk.rs
```diff
@@ -1162,7 +1162,7 @@ impl Chunk {
     /// Returns an iterator over the [`RowId`]s of a [`Chunk`], for a given component.
     ///
     /// This is different than [`Self::row_ids`]: it will only yield `RowId`s for rows at which
-    /// there is data for the specified `component_name`.
+    /// there is data for the specified `component_type`.
```
jleibs: Another case that I think is descriptor, not type, correct?

crates/store/re_chunk/src/helpers.rs
```diff
@@ -281,13 +281,13 @@ impl UnitChunkShared {
...
     /// Returns an error if the data cannot be deserialized.
-    /// In debug builds, panics if the descriptor doesn't have the same component name as the component type.
+    /// In debug builds, panics if the descriptor doesn't have the same type as the component type.
     #[inline]
     pub fn component_batch<C: Component>(
         &self,
         component_descr: &ComponentDescriptor,
     ) -> Option<ChunkResult<Vec<C>>> {
-        debug_assert_eq!(Some(C::name()), component_descr.component_name);
+        debug_assert_eq!(Some(C::name()), component_descr.component_type);
```
jleibs, line 290: This reinforces my above assertion that `::name()` should be `::type()` or `::type_name()`

crates/store/re_chunk/src/iter.rs
```diff
@@ -72,7 +72,7 @@ impl Chunk {
     /// If the chunk is static, `timeline` will be ignored.
     ///
     /// This is different than [`Self::iter_indices`] in that it will only yield indices for rows
-    /// at which there is data for the specified `component_name`.
+    /// at which there is data for the specified `component_type`.
```
jleibs:
```suggestion
    /// at which there is data for the specified `component` descriptor.
```

crates/store/re_chunk/src/iter.rs
```diff
@@ -139,7 +139,7 @@ impl Chunk {
     /// Returns an iterator over the [`TimePoint`]s of a [`Chunk`], for a given component.
     ///
     /// This is different than [`Self::iter_timepoints`] in that it will only yield timepoints for rows
-    /// at which there is data for the specified `component_name`.
+    /// at which there is data for the specified `component_type`.
```
jleibs: Again

crates/store/re_chunk/src/slice.rs
```diff
@@ -16,7 +16,7 @@ use crate::{Chunk, RowId, TimeColumn};
 // Most of them are indirectly stressed by our higher-level query tests anyhow.
 
 impl Chunk {
-    /// Returns the cell corresponding to the specified [`RowId`] for a given [`re_types_core::ComponentName`].
+    /// Returns the cell corresponding to the specified [`RowId`] for a given [`re_types_core::ComponentType`].
```
jleibs:
```suggestion
    /// Returns the cell corresponding to the specified [`RowId`] for a given [`re_types_core::ComponentDescriptor`].
```

crates/store/re_chunk/src/slice.rs
```diff
@@ -1181,11 +1181,11 @@ mod tests {
                 (row_id5, MyPoints::descriptor_labels(), Some(labels5 as _)),
             ];
 
-            for (row_id, component_name, expected) in expectations {
+            for (row_id, component_type, expected) in expectations {
```
jleibs: All of these are descriptors, too.

crates/store/re_chunk_store/tests/dataframe.rs
```diff
@@ -52,10 +51,10 @@ fn schema() -> anyhow::Result<()> {
             .collect::<Vec<_>>(),
         // The following should be in lexicographical order!
         vec![
+            MyPoints::descriptor_indicator(),
```
jleibs, line 59: Both the previous and and the new version are in violation of the comment 2 lines above.

crates/store/re_chunk_store/tests/reads.rs
```diff
@@ -56,16 +56,16 @@ fn all_components() -> anyhow::Result<()> {
         |store: &ChunkStore, entity_path: &EntityPath, expected: Option<&[ComponentDescriptor]>| {
             let timeline = TimelineName::new("frame_nr");
 
-            let component_names = store.all_components_on_timeline_sorted(&timeline, entity_path);
+            let component_types = store.all_components_on_timeline_sorted(&timeline, entity_path);
 
-            let expected_component_names = expected.map(|expected| {
+            let expected_component_types = expected.map(|expected| {
```
jleibs: Another instances of types where we mean descriptors.

crates/store/re_chunk_store/tests/snapshots/formatting__format_chunk_store.snap
```diff
@@ -12,25 +12,25 @@ ChunkStore {
...
-        │ │ row_0000000067816A6Bb4b8c1254d40007b        ┆ 1                    ┆ 2025-01-10T18:43:42.123456789 ┆ [0, 1, 2]                  ┆ [0, 1, 2]                   │ │
-        │ └─────────────────────────────────────────────┴──────────────────────┴───────────────────────────────┴────────────────────────────┴─────────────────────────────┘ │
-        └───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+        ┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+        │ METADATA:                                                                                                                                                                     │
+        │ * rerun.entity_path: /this/that                                                                                                                                               │
```
jleibs: Does this need to be `rerun:` somewhere still?

crates/store/re_dataframe/src/snapshots/re_dataframe__query__tests__async_barebones_static.snap
```diff
@@ -7,8 +7,8 @@ expression: DisplayRB(dataframe)
 │ ---                  ┆ ---                  ┆ ---                                ┆ ---                                ┆ ---                                │
 │ type: i64            ┆ type: Timestamp(ns)  ┆ type: List[u32]                    ┆ type: List[Utf8]                   ┆ type: List[Struct[2]]              │
 │ index_name: frame_nr ┆ index_name: log_time ┆ archetype: example.MyPoints        ┆ archetype: example.MyPoints        ┆ archetype: example.MyPoints        │
-│ kind: index          ┆ kind: index          ┆ archetype_field: colors            ┆ archetype_field: labels            ┆ archetype_field: points            │
-│                      ┆                      ┆ component: example.MyColor         ┆ component: example.MyLabel         ┆ component: example.MyPoint         │
+│ kind: index          ┆ kind: index          ┆ component: example.MyPoints:colors ┆ component: example.MyPoints:labels ┆ component: example.MyPoints:points │
```
jleibs, line 10: Can we update this example to use `example:MyPoints`?

crates/store/re_dataframe/src/query.rs
```diff
@@ -343,11 +343,11 @@ impl<E: StorageEngineLike> QueryHandle<E> {
                         let query =
                             re_chunk::LatestAtQuery::new(TimelineName::new(""), TimeInt::STATIC);
 
-                        // TODO(#6889): We don't allow passing in full descriptors to dataframe queries yet. Everything works via component name.
+                        // TODO(#6889): We don't allow passing in full descriptors to dataframe queries yet. Everything works via component type.
```
jleibs: Is this true? I thought queries were now specified in terms of `component` or descriptor now.

I see:
```
view_contents: ChunkColumnDescriptors,
```

and

```
pub struct ChunkColumnDescriptors {
    /// The primary row id column.
    pub row_id: RowIdColumnDescriptor,

    /// Index columns (timelines).
    pub indices: Vec<IndexColumnDescriptor>,

    /// The actual component data
    pub components: Vec<ComponentColumnDescriptor>,
}
```

crates/store/re_dataframe/src/query.rs
```diff
@@ -343,11 +343,11 @@ impl<E: StorageEngineLike> QueryHandle<E> {
...
                         let component_descriptor = store
-                            .entity_component_descriptors_with_name(
+                            .entity_component_descriptors_with_type(
```
jleibs: This actually seems like a bug. Can't we just directly use the descriptor we already have?

If not it seems like we should have an:
```
entity_component_descriptors_with_component
```

crates/store/re_dataframe/src/query.rs
```diff
@@ -629,15 +626,15 @@ impl<E: StorageEngineLike> QueryHandle<E> {
             .components
             .into_iter()
             .next()
-            .map(|(_component_name, chunks)| {
+            .map(|(_component_type, chunks)| {
```
jleibs:
```suggestion
            .map(|(_descriptor, chunks)| {
```

crates/store/re_dataframe/src/snapshots/re_dataframe__query__tests__selection-3.snap
jleibs, line 1: :warning: this looks like a test regression! :warning: 

Column 6 is now a duplicate of column 2. This seems like we're probably matching the wrong descriptor for an input?

crates/store/re_log_types/src/example_components.rs
```diff
@@ -31,34 +31,37 @@ impl MyPoints {
     }
 
     pub fn descriptor_points() -> ComponentDescriptor {
+        let archetype_name = "example.MyPoints".into();
```
jleibs, line 34:
```suggestion
        let archetype_name = "example:MyPoints".into();
```

crates/store/re_log_types/src/path/component_path.rs
```diff
@@ -2,16 +2,16 @@ use re_types_core::ComponentDescriptor;
...
-/// Example: `camera/left/points:Color`
+/// Example: `camera/left/points:Points3D:color#Color`
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub struct ComponentPath {
     /// `camera / "left" / points / #42`
     pub entity_path: EntityPath,
 
-    /// e.g. `Points3D:Color#color`
+    /// e.g. `Points3D:color#Color`
```
jleibs: The `#Color` bit shouldn't be here any longer, right?

jleibs (commented):
:sweat: lots to go still bit leaving first batch of components since there's a good bit to fix here.

bacon.toml
grtlr: Not really, it's a drive-by fix that I used to test. Happy to pull it out if you think it should have its own commit.

crates/store/re_dataframe/src/snapshots/re_dataframe__query__tests__async_barebones_static.snap
grtlr: I think we should do this when we rename the _values_ of our metadata fields.

crates/store/re_dataframe/src/query.rs
grtlr: Oh good catch, fixed it!

crates/store/re_chunk_store/tests/dataframe.rs
grtlr: Improved the comment to clarify.

crates/store/re_dataframe/src/snapshots/re_dataframe__query__tests__selection-3.snap
grtlr: Ah, the logic was correct, but the test case was stupid. I've improved it

crates/store/re_protos/proto/rerun/v1alpha1/common.proto
```diff
@@ -190,14 +190,23 @@ message PartitionId {
...
+  reserved 2;
+  reserved "archetype_field_name";
+
+  reserved 3;
+  reserved "component_name";
 
-  // Optional name of the field within `Archetype` associated with this data.
-  optional string archetype_field_name = 2;
+  // Optional semantic name associated with this data.
+  optional string component_type = 4;
```
jleibs: Nit: why put component-type first?

This order makes sense to me from a required vs optional perspective
- component (required)
- archetype (optional)
- component_type (optional)

This order makes sense to be strictly lexicographically:
- archetype (optional)
- component (required)
- component_type (optional)

crates/store/re_chunk_store/tests/dataframe.rs
jleibs: This comment still isn't clarified.

Should include a note that indicator doesn't include an archetype, so it sorts first (and will go away).
Also a note about why this ordering is important. Presumably to match some other ordering expectation elsewhere in the test.

crates/store/re_query/src/range.rs
```diff
@@ -15,7 +15,7 @@ use crate::{QueryCache, QueryCacheKey, QueryError};
 // --- Public API ---
 
 impl QueryCache {
-    /// Queries for the given `component_names` using range semantics.
+    /// Queries for the given `component_types` using range semantics.
```
jleibs:
```suggestion
    /// Queries for the given `ComponentDescriptor`s using range semantics.
```

crates/store/re_query/src/range.rs
```diff
@@ -93,7 +93,7 @@ impl RangeResults {
         }
     }
 
-    /// Returns the [`Chunk`]s for the specified `component_name`.
+    /// Returns the [`Chunk`]s for the specified `component_type`.
```
jleibs:
```suggestion
    /// Returns the [`Chunk`]s for the specified `ComponentDescriptor`.
```

crates/store/re_sorbet/src/migrations/v0_23.rs
```diff
@@ -293,76 +351,7 @@ pub fn reorder_columns(batch: &ArrowRecordBatch) -> ArrowRecordBatch {
...
-        &RecordBatchOptions::default().with_row_count(Some(batch.num_rows())),
-    )
-    .ok_or_log_error()
-    .unwrap_or_else(|| ArrowRecordBatch::new_empty(schema))
-}
-
-/// Move indicator component to archetype field name.
-// TODO(#8129): For now, this renames the indicator column metadata. Eventually, we want to remove the column altogether.
-#[tracing::instrument(level = "trace", skip_all)]
-pub fn rewire_indicator_components(batch: &ArrowRecordBatch) -> ArrowRecordBatch {
```
jleibs, line 305: Just clarifying: this behavior is now covered by the new: rewire_tagged_components?

The logic seems a bit different so I want to make sure we're not dropping something here.

crates/store/re_sorbet/src/sorbet_batch.rs
```diff
@@ -169,9 +169,23 @@ impl SorbetBatch {
...
+            crate::migrations::v0_23::migrate_record_batch(&batch)
+        } else {
+            batch.clone()
+        };
+
+        // Migrations to `0.24` below.
+        let batch = make_all_data_columns_list_arrays(
+            &batch,
+            crate::migrations::v0_24::is_component_column,
+        );
```
jleibs: This whole section is very hard to follow. The meaning of "to" and "below" here are both confusing to me.

This reads as: If the schema doesn't match 0.24, then apply these migrations 0.23.  But after applying these migrations, aren't we then at 0.24?

But then what are the 0.24 migrations doing? It seems we hit this code-path unconditionally, even if we already match 0_24 schema?

crates/store/re_sorbet/src/migrations/v0_23.rs
jleibs, line 1: The naming of this file is confusing as things start to acculate. There is an ambiguity as to whether this file migrates FROM 0.22 -> 0.23 or FROM 0.23 -> 0.24.

I believe it's the former as in "In Rerun 0.23 we apply these migrations to bring 0.22 UP to 0.23", but we really need a comment to make it clear.

design/batching.md
```diff
@@ -429,13 +429,13 @@ impl DataCell {
 
     /// Builds a new `DataCell` from an arrow array.
     //
-    // TODO(cmc): We shouldn't have to specify the component name separately, this should be part of the
+    // TODO(cmc): We shouldn't have to specify the component type separately, this should be part of the
```
jleibs: This bit below looks somewhat obsolete given ComponentName no longer exists. I think trait has changed more than just the swap though.

jleibs (approved):
Oof. Ok, I tried but at some point my brain melted and there's no way I caught everything. I think the only real option is to merge and do very very thorough testing of 0.24.

crates/store/re_sorbet/src/migrations/v0_23.rs
grtlr: Yes, this was superseded by `rewire_tagged_components`, it's just that I introduced the changes gradually.

crates/store/re_protos/proto/rerun/v1alpha1/common.proto
grtlr: The order was by accident, it just happened to be the order of my refactor. Fixing it.

design/batching.md
grtlr: Yeah, I've reverted the doc to its original state.

crates/store/re_sorbet/src/sorbet_batch.rs
grtlr: I've updated the comments.

crates/store/re_sorbet/src/migrations/v0_23.rs
grtlr: Added a doc comment to `migrations/mod.rs`

grtlr:
@rerun-bot full-check

grtlr:
@rerun-bot full-check

PR #10252, opened 2025-06-16
Support u16/u32/u64 wide Rerun `enum` types in data definition

Wumpf:
### Related

* part of https://github.com/rerun-io/rerun/issues/7484

### What

Some internal docs talked about how our enums are sparse unions and thus use 8bit variant indices. But that's already not the case - we always serialize/deserialize enums as primitive integer arrays as easily seen by the changes for generated files

Used now in new `VideoCodec` to enable FOURCC values there
Tested via video stream example from python (which implies working serialize in python & deserialize on Rust. Once we add archetype snippets for `VideoStream` we'll have serialization testing in all languages)

### DIFF
--- a/crates/build/re_types_builder/src/objects.rs
+++ b/crates/build/re_types_builder/src/objects.rs
@@ -849,10 +849,10 @@ impl EnumIntegerType {
 
     pub fn format_value(&self, value: u64) -> String {
         match self {
-            EnumIntegerType::U8 => format!("{}", value),
-            EnumIntegerType::U16 => format!("0x{:0X}", value as u16),
-            EnumIntegerType::U32 => format!("0x{:0X}", value as u32),
-            EnumIntegerType::U64 => format!("0x{:0X}", value),
+            Self::U8 => format!("{value}"),
+            Self::U16 => format!("0x{:0X}", value as u16),
+            Self::U32 => format!("0x{:0X}", value as u32),
+            Self::U64 => format!("0x{value:0X}"),
         }
     }
 }


crates/build/re_types_builder/src/codegen/docs/website.rs
```diff
@@ -460,8 +460,12 @@ fn write_fields(reporter: &Reporter, objects: &Objects, o: &mut String, object:
     for field in &object.fields {
         let mut field_string = format!("#### `{}`", field.name);
 
-        if let Some(enum_value) = field.enum_value {
-            field_string.push_str(&format!(" = {enum_value}"));
+        if let Some(enum_or_union_variant_value) = field.enum_or_union_variant_value {
+            if object.is_enum() && object.enum_integer_type() != Some(EnumIntegerType::U8) {
+                field_string.push_str(&format!(" = 0x{enum_or_union_variant_value:X}"));
```
emilk: Let's include leading zeroes too

```suggestion
                field_string.push_str(&format!(" = 0x{enum_or_union_variant_value:0X}"));
```

crates/store/re_types/src/components/video_codec.rs
```diff
@@ -37,7 +37,9 @@ pub enum VideoCodec {
     /// (Note that this is different from AVCC format found in MP4 files.
     /// To learn more about Annex B, check for instance <https://membrane.stream/learn/h264/3>)
     /// Key frames (IDR) require inclusion of a SPS (Sequence Parameter Set)
-    H264 = 1,
+    ///
+    /// Enum value is the fourcc for 'avc1' (the WebCodec string assigned to this codec) in big endian.
+    H264 = 1635148593,
```
emilk: Can we use hex formatting here instead? 🙏

rerun_cpp/src/rerun/components/video_codec.hpp
```diff
@@ -37,7 +37,9 @@ namespace rerun::components {
         /// (Note that this is different from AVCC format found in MP4 files.
         /// To learn more about Annex B, check for instance <https://membrane.stream/learn/h264/3>)
         /// Key frames (IDR) require inclusion of a SPS (Sequence Parameter Set)
-        H264 = 1,
+        ///
+        /// Enum value is the fourcc for 'avc1' (the WebCodec string assigned to this codec) in big endian.
+        H264 = 1635148593,
```
emilk: and here!

rerun_py/rerun_sdk/rerun/components/video_codec.py
```diff
@@ -31,7 +31,7 @@ class VideoCodec(Enum):
     ⚠️ **This type is _unstable_ and may change significantly in a way that the data won't be backwards compatible.**
     """
 
-    H264 = 1
+    H264 = 1635148593
```
emilk: And here!

crates/build/re_types_builder/src/codegen/docs/website.rs
Wumpf: `enum_or_union_variant_value` is always 64bit, so this should use different variants then

crates/store/re_types/src/components/video_codec.rs
Wumpf: guess we should do hex formatting for anything larger than u8. for u8 it's debatable, so far we were fine without

jleibs (approved):
Looks great. I think we use went with u8 before because we were migrating from union which was only i8 and didn't want to tackle making it more generic like this at the time. Nice work.

PR #10253, opened 2025-06-16
Improve catalog table ux

lucasmerlin:
### Related

* closes https://github.com/rerun-io/dataplatform/issues/954

### What

- hide registration time by default
- make recording link second column
- make recording link a link and make it clickable in the whole cell
- truncate `/__properties/recording:RecordingProperties:` from the header name
- Fix TableConfig not being remembered
- properly center text in the header row

After:

https://github.com/user-attachments/assets/8618cc46-237a-410c-9a12-d7a23102dc86

Before:

Screenshot 2025-06-16 at 18 24 11


Wumpf:
screenshot before/after plz if possible :)

crates/viewer/re_dataframe_ui/src/table_blueprint.rs
```diff
@@ -123,20 +125,34 @@ impl ColumnBlueprint {
...
                 // noise in column names.
                 desc.column_name(BatchType::Chunk)
             } else {
                 desc.column_name(BatchType::Dataframe)
-            }
+            };
+
+            name.strip_prefix("/__properties/recording:RecordingProperties:")
+                .map(|s| s.to_owned())
+                .unwrap_or(name)
```
lucasmerlin: Is this fine or is there a better way?

emilk: I definitely think we should do this properly

crates/store/re_sorbet/src/component_column_descriptor.rs
```diff
@@ -276,16 +281,25 @@ impl ComponentColumnDescriptor {
...
+                                self.entity_path,
+                                if short_archetype {
+                                    archetype_name.short_name()
+                                } else {
+                                    archetype_name.full_name()
+                                },
+                                self.archetype_field_name
+                            )
+                        }
+                    }
```
lucasmerlin: @emilk Gave it another go, is this better? Not sure if I'm missing some helpers / constants I could be using to make this cleaner. Seems like the constant for "rerun.archetypes.RecordingProperties" is in re_types which would cause a cyclic dependency with re_sorbet so I can't use that

emilk: Hum… I was hoping that `re_sorbet` could depend on `re_types` as I suspect we might need that for future migrations anyway 🤔 

The only thing that `re_types` uses is `re_sorbet::ComponentColumnSelector` (and only since yesterday). Maybe we could just move that?

emilk: But actually - should this even be in `re_sorbet`? Seems like this is something the caller (`re_dataframe_ui`) could strip out instead?

crates/store/re_sorbet/src/component_column_descriptor.rs
```diff
@@ -294,12 +308,16 @@ impl ComponentColumnDescriptor {
...
+        self.column_name_impl(batch_type, true, false)
     }
 
     /// Uses fully-qualified [`ArchetypeName`]s, if present.
     pub fn column_name_qualified(&self, batch_type: BatchType) -> String {
-        self.column_name_impl(batch_type, false)
+        self.column_name_impl(batch_type, false, false)
+    }
+
+    pub fn column_name_without_recording_properties_prefix(&self, batch_type: BatchType) -> String {
```
emilk: Docstring

emilk (approved):
I think its time to add a screenshot test for this!

PR #10254, opened 2025-06-16
Remove perf regression alerts

jprochazk:
These things don't work if a `gh-pages` branch doesn't exist, and I don't want to keep an empty one around. A GH action should not need 10 different configuration options just to say "don't use git". We need a different solution for this.

Besides, I don't think I've actually seen a size/perf regression alert in any PR for a very long time. Typically the changes are so small that it won't show up, but the changes are still there. It's a lot more productive to monitor our graphs periodically:

https://build.rerun.io


Wumpf (approved):
agreeing with your assessment!

PR #10257, opened 2025-06-16
Make it possible to click the barchart view to select it

emilk:


Wumpf (approved):
neat, thanks!

Wumpf:
I see, we have different ideas what goes to the changelog and what doesn't :)
Very much in favor for small Qol improvements to show up, but always debatable what spams it 🤷

PR #10261, opened 2025-06-17
New dataset API for just retrieving chunk_ids assocaited with a query

jleibs:
This makes it easier to benchmark the query vs chunk fetching portions of the server APIs.

### DIFF
--- a/rerun_py/src/catalog/connection_handle.rs
+++ b/rerun_py/src/catalog/connection_handle.rs
@@ -1,11 +1,13 @@
-use arrow::array::RecordBatch;
+use arrow::array::{RecordBatch, RecordBatchIterator, RecordBatchReader};
 use arrow::datatypes::Schema as ArrowSchema;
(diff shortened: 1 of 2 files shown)


rerun_py/src/catalog/connection_handle.rs
```diff
@@ -456,4 +460,96 @@ impl ConnectionHandle {
 
         Ok(stores)
     }
+
+    pub fn get_chunk_ids_for_dataframe_query(
```
abey79, line 465: I'd rather have any new such helper delegate to `ConnectionClient`, but let's at least have a todo here.

abey79 (approved):

rerun_py/src/catalog/connection_handle.rs
jleibs: I'm leaving as is for consistency, but added the comment and we can tackle them all at the same time in a separate PR.

PR #10263, opened 2025-06-17
Fix:type hint for 3.8

junlinp:
### What
Fix type hint in 3.8.
image

Enviroment: ubuntu 20.04   python 3.8.10
simple code for reproduce:
```
import rerun as rr
if __name__ == "__main__":
    rr.connect_grpc("localhost:9876")
```


Wumpf:
Thanks for wanting to contribute! However, rerun requires Python 3.9+. In fact this breaks a CI job ensuring the more modern type lint syntax

PR #10266, opened 2025-06-17
Make `VideoStream::sample` optional so it's easy to log codec only once

Wumpf:
### Related

* part of #7484

### What

found interface restrictions so far too limiting when doing proper examples

### DIFF
--- a/tests/python/video_ffmpeg/main.py
+++ b/tests/python/video_ffmpeg/main.py
@@ -16,6 +16,9 @@ def capture_webcam_h264() -> None:
     rr.spawn()
     # rr.serve_web()
 
+    # Log the stream's codec once. It never changes to mark it as static.
+    rr.log("video_stream", rr.VideoStream(rr.components.VideoCodec.H264), static=True)
+
     # Target FPS for timing calculations
     target_fps = 30
     frame_duration = 1.0 / target_fps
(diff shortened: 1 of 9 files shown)


tests/python/video_ffmpeg/main.py
```diff
@@ -16,6 +16,9 @@ def capture_webcam_h264() -> None:
     rr.spawn()
     # rr.serve_web()
 
+    # Log the stream's codec once. It never changes to mark it as static.
```
oxkitsune:
```suggestion
    # Log the stream's codec once. It never changes, so mark it as static.
```

tests/python/video_ffmpeg/main.py
```diff
@@ -135,9 +138,7 @@ def is_frame_chunk_start(nal_type: int) -> bool:
                     # We have a complete frame, send it to Rerun
                     frame_time += frame_duration
                     rr.set_time("video_time", duration=frame_time)
-                    rr.log(
-                        "video_stream", rr.VideoStream(sample=current_frame_data, codec=rr.components.VideoCodec.H264)
-                    )
+                    rr.log("video_stream", rr.VideoStream.from_fields(sample=current_frame_data))
```
jprochazk, line 141: Huh, didn't know you could do this. So a required archetype means "it must appear _somewhere_", but it doesn't actually have to be logged with the rest of the archetype :thinking:

oxkitsune (approved):
awesome!

tests/python/video_ffmpeg/main.py
Wumpf: Yeah it's a 0.23 (or was it 0.22?) feature that makes partial updates not suck! :)
Check https://rerun.io/docs/howto/logging/send-partial-updates

Issue #9046, opened 2025-02-14
Add `EncodedDepthImage` archetype


oxkitsune:
`DepthImage`s are currently handled as raw `ImageBuffer` inside, but there are use cases where depth images are stored in encoded formats, such as TIFF (e.g., LeRobot datasets).

Since TIFF decoding is implemented in #8970, this functionality can be reused to support encoded DepthImages.

pablovela5620:
would something like this make it easier to compress depth maps so they don't take up as much memory (like rr.Image.compress() does)

makeecat:
I have made a prototype working with re_mcap to visualize RVL encoded depth image from ros2 mcap bag.

The prototype has the following changes:
1. Implemented the EncodedDepthImage archetype
2. Introduced a new utility crate re_depth_compression to host rvl codecs implementation
3. Integrated EncodedDepthImage archetype into viewer
4. Supported the EncodedDepthImage  archetype in re_mcap.

I would like to get advice from rerun team @oxkitsune on how to proceed merging this new feature into the codebase.

Here is my initial thought:

1. Submit a draft PR with a sample bag to show proof of concept so that rerun team can run the code and get a sense.
2. Get feedback on the general archetype design
3. Split the draft PR into four components: archetype, re_depth_completion crate, viewer integration, re_mcap integration

In the next step, we can implement or wrap more codecs in the re_depth_compression crate.

Please let me know your thoughts.

oxkitsune:
> I have made a prototype working with re_mcap to visualize RVL encoded depth image from ros2 mcap bag.

That is seriously awesome! Looking forward to it!



> I would like to get advice from rerun team @oxkitsune on how to proceed merging this new feature into the codebase.

I think your suggestion of opening a draft PR is a good step, but I'd say that you can open the PR with the new archetype directly alongside it.

The other features (`re_mcap` etc) should be follow up PRs, agreed.

makeecat:
The draft PR is proposed in https://github.com/rerun-io/rerun/pull/11877. The commits of PRs have separated each functionality

PR #10268, opened 2025-06-17
Max image dimension subscriber (used for heuristics) now fully supports `VideoStream`

Wumpf:
### Related

* part of polishing for https://github.com/rerun-io/rerun/issues/7484

### What

This solves issues when view spawning with video stream would have the wrong dimensions.
Plus general improvements to how it works in an all-tagged world.

### DIFF
--- a/crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
+++ b/crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
@@ -2,21 +2,21 @@ use nohash_hasher::IntMap;
 use once_cell::sync::OnceCell;
 
 use re_chunk_store::{ChunkStore, ChunkStoreSubscriberHandle, PerStoreChunkSubscriber};
-use re_log_types::{EntityPath, StoreId};
+use re_log_types::{EntityPath, EntityPathHash, StoreId};
 use re_types::{
-    Archetype as _, Component as _, Loggable as _, archetypes,
-    components::{Blob, ImageFormat, MediaType},
+    Archetype as _, ArchetypeName, Component as _, Loggable as _, archetypes, components,
     external::image,
 };
 
 bitflags::bitflags! {
     #[derive(Debug, Clone, Copy, Default)]
     pub struct ImageTypes: u8 {
-        const IMAGE = 0b0001;
-        const ENCODED_IMAGE = 0b0010;
-        const SEGMENTATION_IMAGE = 0b0100;
+        const IMAGE = 0b1;
+        const ENCODED_IMAGE = 0b10;
+        const SEGMENTATION_IMAGE = 0b100;
         const DEPTH_IMAGE = 0b1000;
-        const VIDEO = 0b10000;
+        const VIDEO_ASSET = 0b10000;
+        const VIDEO_STREAM = 0b100000;
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
```diff
@@ -72,96 +81,129 @@ impl PerStoreChunkSubscriber for MaxImageDimensionsStoreSubscriber {
...
+                ) {
+                    let Some(codec) = codec.first() else {
+                        continue;
+                    }; // Ignore both empty arrays and multiple codecs per row.
+                    if let Some(existing_codec) =
+                        self.video_codecs.insert(entity_path.hash(), *codec)
+                    {
+                        if existing_codec != *codec {
+                            re_log::warn!(
+                                "Overwriting existing video codec for entity path {:?} from {:?} to {:?}. Video codec per entity is expected to never change.",
```
oxkitsune:
```suggestion
                                "Changing video codec for entity path {:?} from {:?} to {:?}. This is unexpected, video codecs should remain constant per entity.",
```
This might make it more clear that this is user error?

crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
```diff
@@ -72,96 +81,129 @@ impl PerStoreChunkSubscriber for MaxImageDimensionsStoreSubscriber {
...
+                        chunk
+                            .iter_slices::<String>(media_type_descr.clone())
+                            .collect()
+                    });
+                    for (blob, media_type) in itertools::izip!(
+                        blobs,
+                        media_types
+                            .into_iter()
+                            .map(Some)
+                            .chain(std::iter::repeat(None))
```
oxkitsune, line 173: neat

crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
```diff
@@ -170,39 +212,54 @@ impl PerStoreChunkSubscriber for MaxImageDimensionsStoreSubscriber {
...
-    media_type: &MediaType,
-    entity_path: &EntityPath,
+    media_type: Option<components::MediaType>,
+    archetype_name: ArchetypeName,
+    debug_name: &str,
 ) -> Option<[u32; 2]> {
     re_tracing::profile_function!();
 
-    if media_type.is_image() {
+    if archetype_name == archetypes::EncodedImage::name() {
```
oxkitsune: This will work until we add `EncodedDepthImage` (#9046). Might be good to test for the `media_type` as well, as this seems like a place to break without going noticed.

oxkitsune (approved):

crates/viewer/re_view_spatial/src/max_image_dimension_subscriber.rs
Wumpf: hmm yeah not sure how to tackle this best. Media type is optional on all of these. And yeah this whole file doesn't scale all that well with any kind of image type :/

Wumpf: I leave a todo with that link so at least the todo zombie check will eventually find it

PR #10259, opened 2025-06-16
Update `egui_plot` and `egui`

emilk:
This adds some margin to the Y-axis axis tickmark labels:

Screenshot 2025-06-16 at 13 27 19

PR #10270, opened 2025-06-17
Update egui (new egui_kittest node)

lucasmerlin:
### Related

* should be merged after https://github.com/rerun-io/rerun/pull/10259

### What

Updates egui_kittest which allows us to use egui::Modifiers in tests


crates/viewer/re_blueprint_tree/tests/range_selection_test.rs
```diff
@@ -83,23 +83,15 @@ fn test_range_selection_in_blueprint_tree() {
 
     harness.run();
     let node2 = harness.get_by_label("entity2");
-    node2.key_down(Key::Shift);
-    node2.click();
-    node2.key_up(Key::Shift);
+    node2.click_modifiers(Modifiers::SHIFT);
```
oxkitsune, line 86: I'm guessing this is doing `click_with_modifiers` right?

oxkitsune (approved):
neat!

crates/viewer/re_blueprint_tree/tests/range_selection_test.rs
lucasmerlin: Uh yeah, should I rename it? I agree that click_modifiers might be a bit confusing

oxkitsune: I think `click_with_modifiers` makes more sense here yes

lucasmerlin: I'll merge this anyway since this will require another egui PR

PR #10275, opened 2025-06-17
Fix fallback  for generic list arrays, 

Wumpf:


emilk (approved):
Nice!

Should we lint `ListArray::values` in `clippy.toml`, with an explanation why? We can always circumvent the lint for when we know what we're doing

Wumpf:
We use them legitimately a lot actually, pretty much whenever we access arrow data in a list array using the list array's offsets etc.

PR #10278, opened 2025-06-17
Build wheel index on main

abey79:
Attempt to build wheel index in push to `main`.

#### Background

We have a wheel index for our nightly and releases on our build server. They offer a facility to install wheels for specific commit without knowing the full path to said wheels. We currently dont have that for push to `main`. We currently only build notebook and linux-x86 on main, but that exactly what we need for dataplatform deployment. Having a wheel index would enable deploying from any ✅  `main` commit.


jleibs (approved):
That looks right. Seems low risk to merge.

abey79:
Failed :( The script wants all wheels. Need to fix that too :(

https://github.com/rerun-io/rerun/actions/runs/15714701744/job/44282665084

PR #7155, opened 2024-08-12
Fix python's `send_columns` failing to convert correctly from `ComponentBatch` to `ComponentColumn` in some cases

Wumpf:
### What

* Fixes  #6592
    * technically not for tensors which is covered by it's union-avoidance-ticket, but this PR adds a note about this
    
Makes it a lot easier to log batches of images in python:
```python
from __future__ import annotations

import numpy as np
import rerun as rr

rr.init("rerun_example_send_columns", spawn=True)

COUNT = 64
WIDTH = 100
HEIGHT = 50
CHANNELS = 3

# Create our time
times = np.arange(0, COUNT)

# Create a batch of images
rng = np.random.default_rng(12345)
image_batch = rng.uniform(0, 255, size=[COUNT, HEIGHT, WIDTH, CHANNELS]).astype(
    dtype=np.uint8
)

# Log the ImageFormat and indicator once, as static.
format_static = rr.components.ImageFormat(
    width=WIDTH, height=HEIGHT, color_model="RGB", channel_datatype="U8"
)
rr.log("images", [format_static, rr.Image.indicator()], static=True)

# Reshape the images so `ImageBufferBatch` can tell that this is several blobs.
rr.send_columns(
    "images",
    times=[rr.TimeSequenceColumn("step", times)],
    components=[rr.components.ImageBufferBatch(image_batch.reshape(COUNT, -1))],
)
```
cc: @Famok

Related things that went into this PR:
* add a note to tensors not supporting batching right now
* support arrays of bytes for blob arrays

Manually tested the snippets for `send_columns` to make sure nothing else broke.
Considered adding the above as a snippet, but I'm still not sure if we want to advertise this prominently. There's some situations where this is useful, but generally we don't have a great usecase for it. Not sure about this 🤔 

### Checklist
* [x] I have read and agree to [Contributor Guide](https://github.com/rerun-io/rerun/blob/main/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/rerun-io/rerun/blob/main/CODE_OF_CONDUCT.md)
* [x] I've included a screenshot or gif (if applicable)
* [x] I have tested the web demo (if applicable):
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/7155?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
[...]

PR #10279, opened 2025-06-17
Add Python example snippet for `VideoStream` using `av` to encode numpy images

Wumpf:
### Related

* part of #7155

### What

Tested on Mac & Windows.
Unlikely I can easily add the same to C++ & Rust even though everything should work just the same. Might warrant doing a "classic" roundtrip test just to be sure this works out alright.

Looks like so:

gh-please-show-this-image-cmon

Rendered https://landing-gq86aduvp-rerun.vercel.app/docs/reference/types/archetypes/video_stream

how to run:
```sh
pixi run py-build
pixi run -e py python ./docs/snippets/all/archetypes/video_stream_synthetic.py
```

TODO:
* [x] test on Linux
* [x] full check to ensure that snippet things run

want to "modernize" the web-cam streaming example into a full snippet example as well - considered it not being a snippet, but it probalby doesn't have enough going for a "full" example. More on that in a follow-up!


Wumpf:
@rerun-bot full-check

grtlr (approved):
Tested locally, runs like a charm 🤩!

_(Well, expect for the loading spinner before clicking play, but that's unrelated I guess.)_

Wumpf:
@rerun-bot full-check

Wumpf:
tested on WSL. not exactly linux but worked there

PR #10284, opened 2025-06-18
Add duration field to the recording UI

pweids:
This is just a small QoL improvement. I found myself looking for this many times when comparing recordings. In particular, I wanted to see the duration of a recording to help compare the sizes between different recordings (I'm evaluating the storage saved by using video).

Take this PR as more of a suggestion, both for the feature and for the actual implementation, as I'm a rust novice and unaware of the general PR process.

Here's a video showing how it works. I skip ahead to show how it ticks over 1 minute.
https://drive.google.com/file/d/1sEGz_6GoJGdZ6XeJ-5VwZh-jZshLtLy0/view?usp=sharing

And a screenshot showing the duration field:
image
image

### DIFF
--- a/crates/viewer/re_data_ui/src/entity_db.rs
+++ b/crates/viewer/re_data_ui/src/entity_db.rs
@@ -1,3 +1,6 @@
+use jiff::SignedDuration;
+use jiff::fmt::friendly::{FractionalUnit, SpanPrinter};
+
 use re_byte_size::SizeBytes as _;
 use re_chunk_store::ChunkStoreConfig;
 use re_entity_db::EntityDb;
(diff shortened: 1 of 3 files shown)


crates/viewer/re_data_ui/src/entity_db.rs
```diff
@@ -85,6 +85,32 @@ impl crate::DataUi for EntityDb {
...
+                        let minutes = (duration.as_secs() % 3600) / 60;
+                        let seconds = duration.as_secs() % 60;
+                        let millis = duration.subsec_millis();
+                        let duration_str = if hours > 0 {
+                            format!("{}h {}m {:02}.{:03}s", hours, minutes, seconds, millis)
+                        } else if minutes > 0 {
+                            format!("{}m {:02}.{:03}s", minutes, seconds, millis)
+                        } else {
+                            format!("{:02}.{:03}s", seconds, millis)
+                        };
```
grtlr: We've been starting to use the `jiff` crate for more and more of these conversions.

It has a [`SpanPrinter`](https://docs.rs/jiff/latest/jiff/fmt/friendly/struct.SpanPrinter.html) that we could use here, but the default `Display` of `jiff::Span` might already work.

grtlr (approved):
I thought this had to do with recording properties so I had to take look 😅.

LGTM, but we could maybe make this a bit more robust using the `jiff` crate.

PS: I merged `main` into this branch in the hope of making the diff nicer, which did not work.

Wumpf:
> PS: I merged main into this branch in the hope of making the diff nicer, which did not work.

this PR is apparently accidentally based on my [unmerged branch](https://github.com/rerun-io/rerun/pull/10279) and can't be merged yet
@pweids to avoid confusion please in the future either..
* simply rebase on top of main so the changes aren't there
* change the target branch from `main` to my branch and then mark the PR as not being mergable (this has complications sometimes though if the target branch keeps moving)
* open this as a draft

additionally, we have the `do-not-merge` tag to signal that nobody should merge a given branch - our CI picks this up and will fail the run.
Thank you! :)
(easily enough caught here, so no worries! Just for future reference)

pweids:
Ah completely forgot I was on your branch, @Wumpf. I'll clean this up. Thanks

crates/viewer/re_data_ui/src/entity_db.rs
pweids: took your suggestion. Much prefer to use a lib that handles all of this safely. Thanks!

pweids:
all fixed up. here's what it looks like from one of the examples

image

crates/viewer/re_data_ui/src/entity_db.rs
grtlr: Thanks, it looks super clean now!

PR #10287, opened 2025-06-18
Stop importing `Self` from `typings`

### LINKED ISSUE #10272: Do not `from typing import Self`
For some reason, such an import made its way to the code base. It breaks because `Self` was introduced only in 3.11.

Should be fixable by using:

```python
if TYPE_CHECKING:
	from typing import Self
```

grtlr:
### Related

* Closes #10272.

### What

TIL.


abey79 (commented):
edit: nvm wrong click. review incoming

abey79 (approved):
Nice!

PR #10290, opened 2025-06-18
Add catalog entry table

lucasmerlin:
### Related

* part of https://github.com/rerun-io/dataplatform/issues/872

### What

Uses the datafusion table to show a table of entries for the whole catalog.

Screenshot 2025-06-18 at 15 00 03

### DIFF
--- a/crates/viewer/re_component_ui/src/variant_uis/redap_entry_kind.rs
+++ b/crates/viewer/re_component_ui/src/variant_uis/redap_entry_kind.rs
@@ -3,9 +3,9 @@ use re_types_core::{ComponentDescriptor, RowId};
 use re_viewer_context::ViewerContext;
 use std::error::Error;
 
-/// Parse an Int32Array as an EntryKind and display it.
+/// Parse an `Int32Array` as an `EntryKind` and display it.
 pub fn redap_entry_kind(
-    ctx: &ViewerContext<'_>,
+    _ctx: &ViewerContext<'_>,
     ui: &mut egui::Ui,
     _component_descriptor: &ComponentDescriptor,
     _row_id: Option<RowId>,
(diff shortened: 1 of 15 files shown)


crates/viewer/re_redap_browser/src/add_server_modal.rs
```diff
@@ -117,7 +118,12 @@ impl AddServerModal {
                                 connection_registry.set_token(&origin, token);
                             }
 
-                            ctx.command_sender.send(Command::AddServer(origin)).ok();
+                            ctx.command_sender
+                                .send(Command::AddServer(origin.clone()))
+                                .ok();
+                            global_ctx.command_sender.send_system(
+                                SystemCommand::ChangeDisplayMode(DisplayMode::RedapServer(origin)),
+                            );
```
Wumpf, line 126: nice!

Wumpf, line 126: nice!

crates/viewer/re_component_ui/src/variant_uis/redap_entry_kind.rs
```diff
@@ -0,0 +1,36 @@
...
+use std::error::Error;
+
+/// Parse an `Int32Array` as an `EntryKind` and display it.
+pub fn redap_entry_kind(
+    _ctx: &ViewerContext<'_>,
+    ui: &mut egui::Ui,
+    _component_descriptor: &ComponentDescriptor,
+    _row_id: Option<RowId>,
+    data: &dyn arrow::array::Array,
+) -> Result<(), Box<dyn Error>> {
```
Wumpf, line 13: out of scope of this pr, but annoys me that this is boxed error. might just as well be anyhow but I bet it's not that diverse in regards of error types and could be thiserror

abey79:
> understanding a bit too little about the envisioned architecture here to gauge whether it's the right call to extend the table blueprint and the data fusion query data like this, but looks great otherwise! :)

That's kind of what I alluded at during yesterday's standup. We need table-blueprint-y stuff but we don't yet have a well established frame for building this. So it happens organically and is occasionally weird. Another way to put it is that we're building hindsight that will help future designs 🤷🏻

PR #9847, opened 2025-04-30
Hacky branch for those who want to experiment with end-to-end tracing

teh-cmc:
This un-mergeable branch makes end-to-end tracing a reality.

To use:
* Start the compose stack in `rerun-io/dataplatform`: `pixi run compose-dev`
* Start your Redap server
* Make sure to rebuild the SDK from this branch: `pixi run py-build-examples`
* Run the test script: `pixi run -e examples python test.py`
* Go to Jaeger: http://localhost:16686/search
  * Select the `rerun-py` service
  * Check out some traces
    image
    image

Issue #9497, opened 2025-04-04
Add `partition_id` field to `Chunk`


abey79:
On the dataplatform, chunks are segregated by partition id, and our query semantics should deal with this boundary in one way or another. (For example, with dataframe queries, the partition id should act as a top-level index.)

Problem: our `Chunk` currently do not store the partition id they belong to. This leads to hacks such as https://github.com/rerun-io/rerun/pull/9496 to keep track of the partition id in parallel to the chunks.

Solution: They should instead be properly tagged with that information.

teh-cmc:
* Superseded by out-of-band metadata: https://linear.app/rerun/issue/DPF-1775

PR #10293, opened 2025-06-18
IO-bound Redap

teh-cmc:
What this does:
* Configures `prost` so that all our `bytes` fields use `bytes::Bytes` instead of `Vec<u8>`
    * This makes our decoded Protobuf messages cheap to clone
* Adds a lot of `tracing` instrumentation spans all over the place
    * These will be no-op on `main` because we don't have any trace subscribers set up there
    * For now you can use the following branch to access all that telemetry: https://github.com/rerun-io/rerun/pull/9847
* Makes `GetChunksResponse` return encoded Chunks (`ArrowMsg`s, to be exact) instead of decoded Chunks
    * On the server-side, this results in the removal of all encoding/decoding from the `Query`/`GetChunks` paths
* Makes `GetChunksResponse` batched (i.e. it's a stream of batches now)
* ⚠️Adds a `chunk_id` field to the `ArrowMsg` Protobuf message
    * This is effectively a first small step towards [out-of-band metadata](https://linear.app/rerun/issue/DPF-1775)
* A lot of changes to the encoding/decoding stack to give the caller more fine-grained control over the different layers
    * In particular, it is crucial to be able to decide when and where to decode transport-level data (Protobuf, cheap) vs. application-level data (Arrow IPC, costly)
    * That whole stack is really starting to suffer from organic growth, but I reckon it'll do for now
* Adds a bunch of early-exit checks to all the Sorbet-migration related tools
* Lots of optmizations around Tokio and IO handling
* Global allocator shenanigans

---

TODO:
* [x] Make sure `ArrowMsg.chunk_id` gets encoded now

---

* Sibling: https://github.com/rerun-io/dataplatform/pull/976
  * See over there for benchmark numbers


teh-cmc:
@rerun-bot full-check

teh-cmc:
AFAIK, the failure during full-check is unrelated, in which case I have no clue why `main` is green :shrug: See https://rerunio.slack.com/archives/C033K5VS2KD/p1750338858537879?thread_ts=1750337545.937309&cid=C033K5VS2KD

crates/store/re_grpc_client/src/redap/mod.rs
teh-cmc, line 1: Two important changes happening here:
* `GetChunksResponse` is now batched (i.e. it's a stream of batches now)
* `GetChunksResponse` returns encoded Chunks (`ArrowMsg`s) instead of decoded Chunks previously

---

Might be of interest to:
* @andrea-reale 
* @abey79        
* @emilk 
* @jprochazk 
* @jleibs 
* @timsaucer 
* @zehiko

crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -216,17 +218,44 @@ pub(crate) async fn client(
...
+                    let _span = tracing::trace_span!(
+                        "get_chunks::batch_decode",
+                        num_chunks = r.chunks.len()
+                    )
+                    .entered();
+
+                    r.chunks
+                        .into_iter()
+                        .map(|arrow_msg| {
+                            let partition_id = arrow_msg.store_id.clone().map(|id| id.id);
```
teh-cmc, line 246: > [!NOTE]
> Note how the `chunk_id` is coming from the transport/protobuf layer instead coming from Arrow metadata! This is possible because a `chunk_id` field was added to `ArrowMsg`.

crates/store/re_log_encoding/src/codec/file/decoder.rs
teh-cmc, line 1: The actual code didn't change one bit here: it just got split into multiple layered functions to give more fine-grained control to the caller.

In particular, it is now possible to only decode up to the transport-layer (Protobuf, cheap), instead of always having to decode all the way to the application-layer (Arrow IPC, costly).

---

Might be of interest to:
@Wumpf         
@emilk 
@jprochazk

crates/store/re_log_encoding/src/decoder/streaming.rs
teh-cmc, line 1: The major change here is the introduction of `StreamingDecoderOptions`:
```rust
pub struct StreamingDecoderOptions {
    pub keep_encoded_protobuf: bool,
    pub keep_decoded_protobuf: bool,
}
```

This gives the caller fine-grained control about how far decoding goes:
* If both are false, only message headers are parsed, but no data is kept (kind of a validation tool)
* If `keep_encoded_protobuf` is set: the decoder will copy the raw protobuf data and keep it around
* If `keep_decoded_protobuf` is set: the decoder decodes the protobuf data and keep it around

The `StreamingDecoder` **never** decodes beyond the transport-layer. It never does anything related to Arrow.

---

Might be of interest to:
@andrea-reale 
@jleibs 
@zehiko

crates/store/re_log_encoding/src/protobuf_conversions.rs
teh-cmc, line 1: The actual code didn't change one bit here: it just got split into multiple layered functions to give more fine-grained control to the caller.

In particular, it is now possible to only decode up to the transport-layer (Protobuf, cheap), instead of always having to decode all the way to the application-layer (Arrow IPC, costly).

---

Might be of interest to:
@Wumpf         
@emilk 
@jprochazk

crates/store/re_protos/proto/rerun/v1alpha1/log_msg.proto
teh-cmc, line 1: :warning: The one single line that makes it all possible :warning: 

This adds a `chunk_id` field to the `ArrowMsg`, which is effectively a first small step towards [out-of-band metadata](https://linear.app/rerun/issue/DPF-1775).

Because this is Protobuf, this is easy to add now and should just be just as easy to remove in the future (e.g. so we can introduce a real, actual dedicated object for metadata).
Since the presence of this field is only relied on by Redap-related code paths, this shouldn't cause backwards compatibility issues.

---

Might be of interest to:
@andrea-reale 
@emilk 
@jprochazk 
@jleibs 
@zehiko

crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
teh-cmc, line 1: Another important performance change, although this one is very localized and barely have any ripple effects.

Might be of interest to:
@andrea-reale 
@jleibs 
@zehiko

crates/store/re_sorbet/src/migration.rs
teh-cmc, line 1: I have added early-exit checks to all migration functions: they yield pretty significant wins since the unhappy paths are pretty costly.

Please review carefully, that's exactly the kind of thing that I tend to get wrong :)

Note that in the future, once we have some form of localized schema caching in place, these paths will become less hot by design. Still, a win's a win.

---

Might be of interest to:
@Wumpf         
@emilk 
@jprochazk 
@grtlr

crates/store/re_sorbet/src/sorbet_batch.rs
teh-cmc, line 1: I have added early-exit checks to all migration functions: they yield pretty significant wins since the unhappy paths are pretty costly.

Please review carefully, that's exactly the kind of thing that I tend to get wrong :)

Note that in the future, once we have some form of localized schema caching in place, these paths will become less hot by design. Still, a win's a win.

---

Might be of interest to:
@Wumpf         
@emilk 
@jprochazk 
@grtlr

rerun_py/src/catalog/connection_handle.rs
teh-cmc, line 1: Nothing crazy happening here, just a lot of optimizations around Tokio.
If you don't consider yourself a Tokio expert, then there may be some interesting learnings for you in there.

---

Might be of interest to:
@andrea-reale 
@abey79        
@emilk 
@jprochazk 
@jleibs 
@timsaucer 
@zehiko

rerun_py/src/lib.rs
teh-cmc, line 1: This makes a major difference, because the client is completely bottlenecked on allocations, like the server before.
Also I don't think we can visualize the result of that accounting anywhere anyhow.

---

Might be of interest to:
@Wumpf         
@emilk

crates/store/re_protos/proto/rerun/v1alpha1/log_msg.proto
abey79: > Since the presence of this field is only relied on by Redap-related code paths, this shouldn't cause backwards compatibility issues.

Those that mean we would be well advised to have a clippy deny on using this attribute on the viewer repo?

crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -115,6 +114,9 @@ pub async fn channel(origin: Origin) -> Result<tonic::transport::Channel, Connec
                 .with_enabled_roots()
                 .assume_http2(true),
         )?
+        // NOTE: Tried it, had no noticeable effects in any of my benchmarks.
+        // .initial_stream_window_size(Some(4 * 1024 * 1024))
+        // .initial_connection_window_size(Some(16 * 1024 * 1024))
```
jleibs: I'm torn between not wanting dead commented code and the bread-crumb that we tried doing this and it didn't improve things.

Regardless, if we're going to add these back, would be nice to start capturing all these tuning/config changes in one place rather than hard-coding.

crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -216,17 +218,44 @@ pub(crate) async fn client(
...
-) -> impl Stream<Item = Result<(Chunk, Option<String>), StreamError>> {
-    response.map(|resp| {
-        resp.map_err(Into::into).and_then(|r| {
-            let batch = r.chunk.ok_or(StreamError::MissingChunkData)?.decode()?;
-
-            let partition_id = batch.schema().metadata().get("rerun.partition_id").cloned();
-            let chunk = Chunk::from_record_batch(&batch).map_err(Into::<StreamError>::into)?;
-
-            Ok((chunk, partition_id))
+) -> impl Stream<Item = Result<Vec<(Chunk, Option<String>)>, StreamError>> {
```
jleibs, line 229: The TODOs on the lines above, in particular https://github.com/rerun-io/rerun/issues/9497 seems obsoleted by this direction?

jleibs: Thinking through future opportunities for access patterns and parallelism, what if we provide the guarantee that *batches* at this level always belong to a single partition.

This would make the signature something like:
```
impl Stream<Item = Result<(String, Vec<Chunk>)>, StreamError>>
```

I believe this should generally be how the data is organized on retrieval anyways.

In turn this means that we can route to local chunk-store handlers at the stream level instead of needing to handle each vec element separately.

crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -216,17 +218,44 @@ pub(crate) async fn client(
...
+    response
+        .then(|resp| {
+            // We want to make sure to offload that compute-heavy work to the compute worker pool: it's
+            // not going to make this one single pipeline any faster, but it will prevent starvation of
+            // the Tokio runtime (which would slow down every other futures currently scheduled!).
+            tokio::task::spawn_blocking(move || {
```
abey79, line 235: As discussed live, this breaks wasm build (in a nasty way: it panics).

Let's add these `spawn_blocking` and similar to `scripts/clippy_wasm/clippy.toml` 🥺

crates/store/re_log_encoding/src/codec/arrow.rs
```diff
@@ -76,16 +79,41 @@ pub(crate) fn decode_arrow(
...
+                let _span = tracing::trace_span!("lz4::decompress").entered();
+                uncompressed.resize(uncompressed_size, 0);
+                lz4_flex::block::decompress_into(data, &mut uncompressed)?;
+                uncompressed.as_slice()
+            }
+        };
+
+        Ok(read_arrow_from_bytes(&mut &data[..])?)
+    } else {
+        // Zero-alloc path: not used today because allocations are bottlenecked on other things,
```
jleibs, line 100: Does this have down-sides? Why not just make this the default now.

That said, `read_arrow_from_bytes` is going to still do a comparable alloc for the arrow data buffers under the hood. The real question here is whether we can decompress directly into a final buffer that becomes the storage used by arrow.

That would both save us the alloc as well as a memcpy.

crates/store/re_log_encoding/src/decoder/streaming.rs
```diff
@@ -34,25 +48,82 @@ pub struct StreamingLogMsg {
...
-impl std::ops::DerefMut for StreamingLogMsg {
-    #[inline]
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.inner
+    /// Returns the decoded transport-level `LogMsg`.
+    ///
+    /// This will only decode at the transport-layer (Protobuf), Arrow data is left untouched.
+    ///
+    /// Compute cost:
+    /// * If [`StreamingDecoderOptions::keep_decoded_protobuf`] was set, this is free.
```
jleibs: This was a bit hard to follow since `StreamingDecoderOptions` isn't visible as an argument or member of this struct. 
```suggestion
    /// * If [`StreamingDecoderOptions::keep_decoded_protobuf`] was set on the [`StreamingDecoder`] when producing the `StreamingLogMsg`, this is free.
```

crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
```diff
@@ -449,7 +450,9 @@ message GetChunksRequest {
 }
 
 message GetChunksResponse {
-  rerun.common.v1alpha1.RerunChunk chunk = 1;
+  // Every gRPC response, even within the confines of a stream, involves HTTP2 overhead, which isn't
+  // cheap by any means, which is why we're returning a batch of `ArrowMsg` rather than a single one.
+  repeated rerun.log_msg.v1alpha1.ArrowMsg chunks = 1;
```
jleibs, line 455: As mentioned in the other comment, I would be very tempted to add an:
```
optional rerun.common.v1alpha1.StoreId store_id = 2;
```
here.

If it's set, then it would considered an error for any chunk to have a mismatching StoreId.

crates/store/re_sorbet/src/migration.rs
```diff
@@ -18,9 +18,18 @@ use re_types_core::{Loggable as _, arrow_helpers::as_array_ref};
...
+#[tracing::instrument(level = "trace", skip_all)]
 pub fn migrate_tuids(batch: &ArrowRecordBatch) -> ArrowRecordBatch {
     re_tracing::profile_function!();
 
+    let needs_migration = batch.schema_ref().fields().iter().any(|field| {
+        field.extension_type_name() == Some("rerun.datatypes.TUID")
+            || field.name() == "rerun.controls.RowId"
+    });
+    if !needs_migration {
+        return batch.clone();
```
jleibs, line 30: I know these clones are fairly cheap, but would it be better for this to to take ownership?
```
pub fn migrate_tuids(batch: ArrowRecordBatch) -> ArrowRecordBatch {
    if !needs_migration {
        return batch;
```
I can't imagine us doing anything with the un-migrated batch.

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -401,62 +405,188 @@ impl ConnectionHandle {
...
-                        store_version: None,
-                    };
+                let mut stores = BTreeMap::default();
+
+                while let Some(chunks_and_partition_ids) = rx.recv().await {
+                    // We want to make sure to offload that compute-heavy work to the compute worker pool: it's
+                    // not going to make this one single pipeline any faster, but it will prevent starvation of
+                    // the Tokio runtime (which would slow down every other futures currently scheduled!).
+                    stores = tokio::task::spawn_blocking({
+                        let mut stores = stores.clone(); // just cloning handles
```
jleibs: I missed the leading `stores =` on first read and thought I was losing my mind for a bit.

```suggestion
                    stores = tokio::task::spawn_blocking({
                        // Clone the stores for mutabillity within the spawned task.
                        // Note at the end of this task we return the mutated stores and assign
                        // it back to the outer stores variable.
                        let mut stores = stores.clone(); 
```

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -401,62 +405,188 @@ impl ConnectionHandle {
...
+                                            store_version: None,
+                                        };
+
+                                        let mut store = ChunkStore::new(
+                                            store_info.store_id.clone(),
+                                            Default::default(),
+                                        );
+                                        store.set_info(store_info);
+                                        ChunkStoreHandle::new(store)
+                                    });
```
jleibs, line 526: Single store per batch would let us move this outside of the `spawn_blocking`. I think that might even let us make it a `spawn()` then which would give us parallel ingestion?

rerun_py/src/catalog/dataset_entry.rs
```diff
@@ -263,11 +267,13 @@ impl PyDatasetEntry {
...
                 get_chunks_response_to_chunk_and_partition_id(catalog_chunk_stream);
 
-            while let Some(chunk) = chunk_stream.next().await {
-                let (chunk, _partition_id) = chunk.map_err(to_py_err)?;
-                store
-                    .insert_chunk(&std::sync::Arc::new(chunk))
-                    .map_err(to_py_err)?;
+            while let Some(chunks) = chunk_stream.next().await {
+                for chunk in chunks.map_err(to_py_err)? {
+                    let (chunk, _partition_id) = chunk;
```
jleibs: This should probably error if _partition_id != partition_id?

jleibs (approved):
Awesome! That was actually much less involved than I was fearing.

crates/store/re_log_encoding/src/codec/file/decoder.rs
```diff
@@ -13,28 +17,83 @@ pub(crate) fn decode(data: &mut impl std::io::Read) -> Result<(u64, Option<LogMs
...
-            Some(LogMsg::SetStoreInfo(set_store_info.try_into()?))
+            let msg = SetStoreInfo::decode(buf)?;
+            Some(re_protos::log_msg::v1alpha1::log_msg::Msg::SetStoreInfo(
+                msg,
+            ))
         }
+
         MessageKind::ArrowMsg => {
-            let arrow_msg = ArrowMsg::decode(buf)?;
+            let msg = ArrowMsg::decode(buf)?;
```
jleibs, line 66: I just realized that for existing RRDs, this will be missing the newly added `chunk_id` field. It's not totally clear to me where that is actually used at this point if at all, but this seems like something that needs some comments and a forward/backward compatibility story.

crates/store/re_sorbet/src/migration.rs
```diff
@@ -231,9 +301,20 @@ pub fn reorder_columns(batch: &ArrowRecordBatch) -> ArrowRecordBatch {
...
 // TODO(#8129): For now, this renames the indicator column metadata. Eventually, we want to remove the column altogether.
+#[tracing::instrument(level = "trace", skip_all)]
 pub fn rewire_indicator_components(batch: &ArrowRecordBatch) -> ArrowRecordBatch {
     re_tracing::profile_function!();
 
+    let needs_rewiring = batch.schema_ref().fields().iter().any(|field| {
+        field
+            .metadata()
+            .get("rerun.component")
+            .is_some_and(|value| value.ends_with("Indicator"))
```
grtlr, line 312: The decision on when to run the migration code is a bit more involved.

You also need to check if the metadata does not contain `rerun.archetype` and `rerun.archetype_field`.

https://github.com/rerun-io/rerun/blob/main/crates/store/re_sorbet/src/migration.rs#L258-L260

Also, this will change again with https://github.com/rerun-io/rerun/pull/10245, but I will take care of the early exit checks there 👍.

grtlr (changes requested):

teh-cmc:
@rerun-bot full-check

crates/store/re_log_encoding/src/codec/file/decoder.rs
teh-cmc: The new `chunk_id` field is only necessary for the chunk cache on the server, which will only work if the data was generated with out-of-band chunk IDs in the first place.

Now the 1 million dollar question is... did I remember to update the encoder so that it writes those out-of-band chunk IDs from now :melting_face: 

In any case I want to make a topic out of this today: there's an interesting BW compat dilemma here.

rerun_py/src/catalog/dataset_entry.rs
teh-cmc: I guess I'll add a filter with a log message

crates/store/re_log_encoding/src/decoder/streaming.rs
```diff
@@ -1,31 +1,45 @@
...
-    pub inner: LogMsg,
+    pub kind: MessageKind,
+
+    /// The [`CrateVersion`] of the RRD stream from which this `LogMsg` was taken from.
+    pub version: CrateVersion,
+
+    /// The original Protobuf-encoded bytes of this `LogMsg`, *including headers*.
+    ///
+    /// Only set if [`StreamingDecoderOptions::keep_encoded_protobuf`] is true.
+    pub encoded: Option<bytes::Bytes>,
```
andrea-reale, line 30: Question and nit.

Question for my understanding: why do we want keep both the encoded and decoded version at the same time?

nit: it'd be clearer to me to read this structure if it were an enum

```rust
enum MsgInner { // or better name
  Encoded(bytes::Bytes),
  Decoded(log_msg::Msg),
  Both(bytes::Bytes, log_msg::Msg),
  Discarded,
}
```

crates/store/re_grpc_client/src/redap/mod.rs
teh-cmc: My experience from the past 3 years is that these breadcrumbs have been super valuable, but I agree comments suck. I'll see if I can at least promote that to a `if false {}` block.

crates/store/re_grpc_client/src/redap/mod.rs
teh-cmc: > The TODOs on the lines above, in particular #9497 seems obsoleted by this direction?

Yeah I think it's fair to say that #9497 is superseded by the general issue of out-of-band metadata in any case.

teh-cmc: > Thinking through future opportunities for access patterns and parallelism, what if we provide the guarantee that _batches_ at this level always belong to a single partition.
> 
> This would make the signature something like:
> 
> ```
> impl Stream<Item = Result<(String, Vec<Chunk>)>, StreamError>>
> ```
> 
> I believe this should generally be how the data is organized on retrieval anyways.
> 
> In turn this means that we can route to local chunk-store handlers at the stream level instead of needing to handle each vec element separately.

That makes sense to me. In general there are a lot of end-to-end opportunities to make the batching more aligned with Rerun semantics all the way to the server (e.g. a single `GetChunksResponse` could be scoped to a partition_id from the beginning).

I'll add a TODO for now and then we can compute contiguous runs in a follow-up.

andrea-reale (commented):
I would lie if I say I understand more than 30% of this without spending a full day on it. But lots of interesting learnings, thanks!

crates/store/re_log_encoding/src/codec/arrow.rs
teh-cmc: > Does this have down-sides? Why not just make this the default now.

It basically leaks some memory per thread, where "some" corresponds to the largest message seen by this thread :shrug: 

> That said, read_arrow_from_bytes is going to still do a comparable alloc for the arrow data buffers under the hood. The real question here is whether we can decompress directly into a final buffer that becomes the storage used by arrow.

Yes, this annoys me greatly :)

I think there might be a way, using the File-oriented APIs: https://github.com/apache/arrow-rs/blob/main/arrow/examples/zero_copy_ipc.rs

I'll add a TODO -- might want to look into this when I get a minute.

crates/store/re_sorbet/src/migration.rs
teh-cmc: It would definitely be slightly better, but the original reference is coming from so deep, I don't want to propagate such a change all over the place just for this.

crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
teh-cmc: Ha! So we did come to the same conclusion. Yeah I agree there are opportunities here, but let's keep that to a TODO / follow-up.

rerun_py/src/catalog/connection_handle.rs
teh-cmc: See other comments regarding partition-based batching.

> I think that might even let us make it a spawn() then which would give us parallel ingestion?

We still need `spawn_blocking` either way, because inserting into the store is super compute intensive (just-in-time compaction).

teh-cmc:
:partying_face: 
image

crates/store/re_protos/proto/rerun/v1alpha1/log_msg.proto
teh-cmc: I think the only real solution here is addressing https://linear.app/rerun/issue/DPF-1775 ASAP. You can't really use that field by accident though, it's only available deep in the decoding layers.

crates/store/re_log_encoding/src/decoder/streaming.rs
teh-cmc: > Question for my understanding: why do we want keep both the encoded and decoded version at the same time?

There are a few scenarios where that could be useful, one of those is any kind of "smart-proxy": you want to be able to both forward the original encoded data as-is, without needlessly re-encoding it all the time, but at the same time you also need some of the decoded information to perform your proxy logic in the first place.

> nit: it'd be clearer to me to read this structure if it were an enum

IMO/IME any enum with a `Both` (or `All`) variant should be treated with caution:
* Mechanically, it's a lot of maintenance pain for generally little benefit.
* Semantically, this isn't an enumeration of the different states possible for the system, it really is more of a bitmask of attributes that affect how the data gets transformed. E.g. it's more like a cache, or a `is_sorted` boolean, that kind of thing.

crates/store/re_log_encoding/src/codec/file/encoder.rs
```diff
@@ -28,13 +28,14 @@ pub(crate) fn encode(
             let payload = encode_arrow(&arrow_msg.batch, compression)?;
             let arrow_msg = ArrowMsg {
                 store_id: Some(store_id.clone().into()),
+                chunk_id: Some(arrow_msg.chunk_id.into()),
```
teh-cmc, line 31: This is what will make chunk caching work on the server! cc @jleibs

teh-cmc: Ooooh but now I look at this and I think it doesn't do what I think it does :eyes:

teh-cmc: Nevermind, it does what I think it does :smile:

Issue #10155, opened 2025-06-06
Notifications in light mode have very little contrast


grtlr:
@gavrelina is probably on this already with her design token changes, but right now, notifications have very little contrast. Here is an example:

<img width="439" alt="Image" src="https://github.com/user-attachments/assets/6e8dafe3-4ea9-4cc5-8a30-ad988b6c1a98" />

gavrelina:
also in dark mode btw, 
assumption: the styling that is applied to those is also applied to smth else. 
initially it was ok, and then it broke along the way

PR #10107, opened 2025-06-01
Perfecting light mode design

gavrelina:
WIP

notes: 

- [x] notifications 
- [x] bar in barchart

- ? shall be an icon and not a font

- backgrounds of all views (except 3D) shall be 1 alias

Fancy triangles icons: needs resizing. tried, but it was magic math beyond my patience.
image
 
separate font style from default and on-surface
now both `Gyroscope` and `3d view` defined by the same alias
image

... many more to copy here from notes

PR #10296, opened 2025-06-18
fix notification text color

gavrelina:
before 

image
image

after 
image
image

It's still okay to have time stamp weak and 0 notifications state weak. 
To improve later => assign a color token instead of using default egui. 

closes https://github.com/rerun-io/rerun/issues/10155
relates to https://github.com/rerun-io/rerun/pull/10107


grtlr (approved):
Much better! Thanks!

PR #10297, opened 2025-06-18
Standalone example for compressed video streams

Wumpf:
### Related

* Part of github.com/rerun-io/rerun/issues/7484

### What

See Readme for all the details.

pic

TODO:
* [x] test on Windows
* [x] test on MacOS
* [x] test on linux - couldn't test this one on WSL
* [x] pass full-ci check


examples/python/camera_video_stream/camera_video_stream.py
```diff
@@ -0,0 +1,123 @@
...
+                # If we pass this just on as-is, then we get an encoded video stream that
+                # just consists entirely of I-frames, thus having very poor compression!
+                # Instead, we want the encoder to decide when to use P & I frames.
+                frame.pict_type = av.video.frame.PictureType.NONE
+
+                for packet in output.encode(frame):
+                    if packet.pts is None:
+                        continue
+                    rr.set_time("video_stream", duration=float(packet.pts * packet.time_base))
+                    rr.log("video_stream", rr.VideoStream.from_fields(sample=bytes(packet)))
```
nikolausWest: Should both of these `"video_stream"` be the same? I find it a little confusing. Because the strings are the same it makes me (a casual reader of Rerun SDK usage) assume they need to be the same or that that has some important meaning

nikolausWest: Also, wow, the Rerun portion of this is example is so minimal 👨‍🍳👌

Wumpf: oh you're right that's very confusing! Same is true for the snippet in https://github.com/rerun-io/rerun/pull/10279
need to fix both

Wumpf: also now that I see it again `set_time` should show up in the readme explain there how it's important and what this PTS business means

Wumpf:
We should put in a better thumbnail - maybe a FPV of a gripper robot picking up something.
May need to wait until next Tuesday as I'm out of office for a few days

examples/python/camera_video_stream/README.md
```diff
@@ -0,0 +1,58 @@
...
+```py
+rr.log("video_stream", rr.VideoStream(codec=rr.VideoCodec.H264), static=True)
+```
+H264 is a very well established codec and well suited for streaming.
+Note that in `setup_output_stream` we specicifially configure `pyAV` to use low latency encoding.
+
+⚠️ Latency is expected to be well below a second, but as of writing Rerun is not well
+optimized for low latency video transmission as used for teleoperations.
+If you need lower latency you should consider tweaking the encoding settings further and
+configure Rerun's [Micro Batching](https://rerun.io/docs/reference/sdk/micro-batching).
```
oxkitsune, line 35: I almost think it would be nice to suggest some values here, but it's kind of difficult to do that without knowing the specific setup people will be running...

examples/python/camera_video_stream/camera_video_stream.py
```diff
@@ -0,0 +1,123 @@
...
+    elif platform.system() == "Windows":
+        if video_device is None:
+            # On windows we *have* to know the device name. Query ffmpeg for it.
+            list_devices_cmd = ["ffmpeg", "-f", "dshow", "-list_devices", "true", "-i", "dummy", "-hide_banner"]
+            devices_output = subprocess.check_output(list_devices_cmd, stderr=subprocess.STDOUT).decode()
+
+            # Extract video device names
+            video_devices = []
+            for line in devices_output.split("\n"):
+                if '"' in line and "Alternative name" not in line:
```
oxkitsune: When I tested this on my Windows machine it got confused with my headphones:

```
Using video device: Microphone (HyperX Cloud II Wireless)
Could not enumerate video devices (or none found).

av.error.OSError: [Errno 5] I/O error: 'video=Microphone (HyperX Cloud II Wireless)'; last error log: [dshow] Could not enumerate video devices (or none found).
```

I tried sidestepping the whole `ffmpeg` shenanigans, but I don't think it's worth more time 😬

To fix it, I suggest we do:

```suggestion
                # FFmpeg will return both `(audio)` and `(video)` devices
                if "(video)" in line and "Alternative name" not in line:
```

oxkitsune (approved):

examples/python/camera_video_stream/README.md
Wumpf: yeah, I'm struggling here as well :/
I think it's fine to leave it be for now, but we may want to revisit it

examples/python/camera_video_stream/camera_video_stream.py
Wumpf: thanks, works on my windows as well!

Wumpf:
@rerun-bot full-check

Wumpf:
full build failed with known issue. Unfortunately, that's part of the testsuite I'm intersted in. Gotta run again after merge

Wumpf:
@rerun-bot full-check

examples/python/camera_video_stream/README.md
emilk: The default batch max-time is 8ms - the end2end latency won't improve anymore than that no matter how much the user tweaks the micro batcher.

Have you looked at the new latency view btw to get a sense of where the latency comes from? 1s is massive

emilk: This would be helpful for you:
* https://github.com/rerun-io/rerun/pull/10402

Issue #10247, opened 2025-06-16
Make it possible to edit an existing server in the redap browser


abey79:
Currently, one must delete/re-create a server to edit some of its parameter—in particular the auth token. We should have an "edit server" feature, which would basically bring back the modal.

the minimum: 
introduce contextual menu for the server entry
<img width="940" alt="Image" src="https://github.com/user-attachments/assets/6f5c4a37-f282-469d-8666-0e97ddc9fd86" />

- related https://github.com/rerun-io/rerun/issues/10246
- part of https://github.com/rerun-io/dataplatform/issues/958
- part of https://linear.app/rerun/issue/DPF-1779/improve-the-ux-of-auth-failure-in-the-viewer

PR #10300, opened 2025-06-19
Allow editing connection settings for an existing redap server

lucasmerlin:
### Related

* closes https://github.com/rerun-io/rerun/issues/10247

### What

Makes it possible to edit an existing redap server


crates/store/re_grpc_client/src/connection_registry.rs
```diff
@@ -65,6 +65,13 @@ impl ConnectionRegistryHandle {
         });
     }
 
+    pub fn get_token(&self, origin: &re_uri::Origin) -> Option<Jwt> {
```
emilk: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

crates/viewer/re_redap_browser/src/add_server_modal.rs
```diff
@@ -6,19 +6,26 @@ use re_ui::modal::{ModalHandler, ModalWrapper};
 use re_uri::Scheme;
 use re_viewer_context::{DisplayMode, GlobalContext, SystemCommand, SystemCommandSender as _};
 
-pub struct AddServerModal {
+pub enum ServerModalMode {
+    Add,
+    Edit(re_uri::Origin),
+}
```
emilk: Docstrings

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -225,19 +226,39 @@ impl Server {
...
+                        .as_button_with_label(ui.tokens(), "Edit server")
+                        .ui(ui)
+                        .clicked()
+                    {
+                        ctx.command_sender
+                            .send(Command::OpenEditServerModal(self.origin.clone()))
+                            .ok();
+                    }
+                    if icons::REMOVE
+                        .as_button_with_label(ui.tokens(), "Remove")
```
emilk, line 257: Inconsistent: every other option ends with " server". I suggest removing the suffix from all of them.

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -381,7 +407,29 @@ impl RedapServers {
...
+                self.server_modal_ui
+                    .open(ServerModalMode::Edit(origin), connection_registry);
+            }
+
+            Command::UpdateServer {
+                previous_origin,
+                new_origin,
+            } => {
+                self.servers.remove(&previous_origin);
+                self.servers.insert(
```
emilk: Why have `UpdateServer` when you can just `RemoveServer` + `AddServer`?

crates/viewer/re_redap_browser/src/add_server_modal.rs
```diff
@@ -6,8 +6,14 @@ use re_ui::modal::{ModalHandler, ModalWrapper};
...
 
+use crate::{context::Context, servers::Command};
+
+/// Should the modal edit an existing server or add a new one?
 pub enum ServerModalMode {
+    /// Show an empty modal to add a new server.
     Add,
+
+    /// Show a modal to edit an existing server.
     Edit(re_uri::Origin),
```
emilk, line 20: What happens if it doesn't already exist? Will it act like `Add` (but with a default value)?

crates/viewer/re_redap_browser/src/add_server_modal.rs
```diff
@@ -48,7 +75,15 @@ impl AddServerModal {
     ) {
         self.modal.ui(
             ui.ctx(),
-            || ModalWrapper::new("Add Server"),
+            || {
+                let title = match &self.mode {
+                    ServerModalMode::Add => "Add Server".to_owned(),
+                    ServerModalMode::Edit(origin) => {
+                        format!("Edit Server: {}", origin.host)
```
emilk: https://github.com/rerun-io/rerun/blob/65bf078f22592b88a918456905382ac19eab7bf0/DESIGN.md#L10-L14

```suggestion
                    ServerModalMode::Add => "Add server".to_owned(),
                    ServerModalMode::Edit(origin) => {
                        format!("Edit server: {}", origin.host)
```

emilk (approved):

gavrelina:
visual nit: give some air to breath! 
see how it takes same amount of space, but feels way nicer:
image

Some more details 
image

emilk:
Looks like we should set `egui::Spacing::window_margin` to 4.0 then in `design_tokens.rs`

crates/viewer/re_redap_browser/src/add_server_modal.rs
lucasmerlin: Yeah, but you shouldn't do that since it'd be confusing (The modal will show `Edit server` but for the user it's a new server). Updated the comment.

lucasmerlin:
> visual nit: give some air to breath!
> see how it takes same amount of space, but feels way nicer:

Your design looks way nicer but instead of doing something custom here we should update the menu styling so that all rerun menus look nice. This should go in a separate PR though

gavrelina:
>Your design looks way nicer but instead of doing something custom here we should update the menu styling so that all rerun menus look nice. This should go in a separate PR though

Works for me unless we abandon it ;) 
Created this
- https://github.com/rerun-io/rerun/issues/10354

PR #10301, opened 2025-06-19
Blink help buttons until the user clicks one of them

emilk:
### What
I've noticed a lot of our users have missed the `?` help icon in our views.

This PR attempts to remedy this by sotly blinking the icons occasionally, until the user clicks at least one of them, thus teaching the user that there is help to find, and where to find it.

Once one help button has been clicked, no help button will ever blink again (unless the user runs `rerun reset`).

### DIFF
--- a/crates/viewer/re_ui/src/ui_ext.rs
+++ b/crates/viewer/re_ui/src/ui_ext.rs
@@ -1076,28 +1076,56 @@ pub trait UiExt {
         }
     }
 
+    /// The help menu appears when clicked and/or hovered
     fn help_button(&mut self, help_ui: impl FnOnce(&mut egui::Ui)) -> egui::Response {
-        // The help menu appears when clicked and/or hovered
         let mut help_ui: Option<_> = Some(help_ui);
 
         let ui = self.ui_mut();
 
-        let menu_button = egui::containers::menu::MenuButton::from_button(
-            ui.small_icon_button_widget(&icons::HELP, "Help"),
-        );
-        let button_response = menu_button
-            .ui(ui, |ui| {
-                if let Some(help_ui) = help_ui.take() {
-                    help_ui(ui);
-                }
-            })
-            .0;
+        // If we never showed the help before, we want to animate it to draw attention:
+        let has_shown_help_id = egui::Id::new("has_shown_help");
+        let has_been_shown: bool =
+            ui.data_mut(|d| *d.get_persisted_mut_or_default(has_shown_help_id));
 
-        if let Some(help_ui) = help_ui.take() {
-            button_response.on_hover_ui(help_ui)
-        } else {
-            button_response
-        }
+        ui.scope(|ui| {
+            if !has_been_shown {
+                blink_inactive_buttons(ui);
+            }
+
+            let menu_button = egui::containers::menu::MenuButton::from_button(
+                ui.small_icon_button_widget(&icons::HELP, "Help"),
+            );
+            let button_response = menu_button
+                .ui(ui, |ui| {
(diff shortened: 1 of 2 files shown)


lucasmerlin:
This seems to break showing the help ui on hover, I have to click now. I guess it's intentional?

emilk:
> This seems to break showing the help ui on hover, I have to click now. I guess it's intentional?

Yes - clicking it is more consistent with every other menu with have (e.g. you don't see the Rerun menu contents by just hovering it).

emilk:
Katya suggested this design:

https://github.com/user-attachments/assets/d400a2fd-6583-4e6e-a708-c81f296c7beb

Again, only until the user actually clicks one help button at some point.

gavrelina:
@emilk I have changes the green color (to match our blue shades) and also from circle to rect. 
Didn't know how to fix the one in the timepanel 
Screenshot 2025-06-24 at 10 07 54

and another suggestion: 
because now we show help on click (which is nice) I think we shall still keep the same help on hover. Otherwise it's quite silly that we "blink" the help icon and when user hover on it it says "click on me!" instead of directly showing them help information. Maybe from dev perspective it's weird to copy paste code (or maybe there is another way) but from interaction I think that's better!

crates/viewer/re_ui/src/ui_ext.rs
```diff
@@ -1076,28 +1076,64 @@ pub trait UiExt {
...
+            let menu_button = egui::containers::menu::MenuButton::from_button(
+                ui.small_icon_button_widget(&icons::HELP, "Help"),
+            );
+
+            let response = menu_button
+                .ui(ui, |ui| {
                     help_ui(ui);
+                    if !user_has_clicked_any_help_button {
+                        // Remember that the user has found and used the help button at least once,
+                        // to stop it from animating in the future:
```
emilk: TODO: send an analytics event here, once https://github.com/rerun-io/rerun/pull/10333 is merged
* https://github.com/rerun-io/rerun/pull/10333

crates/viewer/re_ui/src/ui_ext.rs
```diff
@@ -1076,28 +1076,76 @@ pub trait UiExt {
...
+            let where_to_paint_background =
+                if !user_has_clicked_any_help_button && is_hovering_container {
+                    Some(ui.painter().add(egui::Shape::Noop))
+                } else {
+                    None
+                };
+
+            let menu_button = egui::containers::menu::MenuButton::from_button(
+                ui.small_icon_button_widget(&icons::HELP, "Help"),
+            );
```
lucasmerlin, line 1108: You could just do something like 
```rust
if !user_has_clicked_any_help_button && is_hovering_container {
    ui.visuals_mut().inactive.bg_fill = ui.tokens().highlight_color
}

```
 then it's guaranteed that the help highlight matches the hover highlight and you won't need the where_to_put_background and the sizing hack

emilk: Yeah, I did that initially, but then I wanted a circular background… and then Katya wanted a square background again, but with a custom expand and a custom rounding… so I think the code that stands is still simpler.

lucasmerlin: If it's supposed to be different expand / rounding it's not showing correctly:

https://github.com/user-attachments/assets/565fb6ff-4614-4bda-b4e8-ee73910c7552

gavrelina: It doesn't have to be different rounding! I was trying to match it to the existing one, so it's actually perfect! :) 
sorry for the mess

PR #10303, opened 2025-06-19
Add `Span` with guaranteed non-negative length

emilk:
### What
The built-in `Range` type(s) in `std` all support negative ranges, e.g. `7..3`. In most cases, this is annoying and avoidable. The way we've been avoiding them is by having unsigned `offset/length` pairs.

I thought it would be nice to combine them into one `Span`:

```rs
pub struct Span<Idx: Unsigned> {
    pub start: Idx,
    pub len: Idx,
}
```

In particular, this avoid a bunch of places where we use `(usize, usize)` as the range type, which feels like a footgun waiting to happen.

Moreover, a dedicated struct allows for nice helpers such as multiplying the range (e.g. to convert an element range to a byte range by multiplying by element size).

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -98,11 +98,11 @@ Of course, this will only take us so far. In the future we plan on caching queri
 Here is an overview of the crates included in the project:
 
 <picture>
-  <img src="https://static.rerun.io/crates/bcb1f21ad8a038212c2be5d9101fca3822d7ab6c/full.png" alt="">
-  <source media="(max-width: 480px)" srcset="https://static.rerun.io/crates/bcb1f21ad8a038212c2be5d9101fca3822d7ab6c/480w.png">
-  <source media="(max-width: 768px)" srcset="https://static.rerun.io/crates/bcb1f21ad8a038212c2be5d9101fca3822d7ab6c/768w.png">
-  <source media="(max-width: 1024px)" srcset="https://static.rerun.io/crates/bcb1f21ad8a038212c2be5d9101fca3822d7ab6c/1024w.png">
-  <source media="(max-width: 1200px)" srcset="https://static.rerun.io/crates/bcb1f21ad8a038212c2be5d9101fca3822d7ab6c/1200w.png">
+  <img src="https://static.rerun.io/crates/087dcf1003897305755afe633cf304d20d6437c2/full.png" alt="">
+  <source media="(max-width: 480px)" srcset="https://static.rerun.io/crates/087dcf1003897305755afe633cf304d20d6437c2/480w.png">
+  <source media="(max-width: 768px)" srcset="https://static.rerun.io/crates/087dcf1003897305755afe633cf304d20d6437c2/768w.png">
+  <source media="(max-width: 1024px)" srcset="https://static.rerun.io/crates/087dcf1003897305755afe633cf304d20d6437c2/1024w.png">
+  <source media="(max-width: 1200px)" srcset="https://static.rerun.io/crates/087dcf1003897305755afe633cf304d20d6437c2/1200w.png">
 </picture>
 
 <!-- !!! IMPORTANT!!!


Wumpf:
concept of a positive range makes sense to me but the naming is a little bit off the mark imho: it's not a `Range` in the sense of the std `Range` because it's not start + end, but offset + len.
Maybe more of a `USpan`? Not as easy to discover though

emilk:
There are several ranges in std, only one which is start + end:

Screenshot 2025-06-19 at 12 27 55

(<https://doc.rust-lang.org/std/ops/index.html#structs>)

So I think the word "range" makes sense.

But I'm not fully sold on the `U` part of the name. `LenRange`?

Wumpf:
> There are several ranges in std, only one which is start + end:

ah but they all still have starts & ends in there, no? But yeah I didn't realize there's a whole zoo, this still makes me drop my argument (:
I was skeptical about the U part as well first, but it _does_ restrict its types to unsigned, so it's not wrong as such.
If we stick with the range naming that std establishes here it should be rather `RangeLen` or `RangePositive` 🤔

abey79:
`SpanRange` (or `RangeSpan` to keep to `std` naming scheme)?

emilk:
Maybe `Span` is good after all… at least it's nice and short

ARCHITECTURE.md
```diff
@@ -227,6 +227,7 @@ Update instructions:
 | re_log             | Helpers for setting up and doing text logging in the Rerun crates.                   |
 | re_memory          | Run-time memory tracking and profiling.                                              |
 | re_smart_channel   | A channel that keeps track of latency and queue length.                              |
+| re_span            | An integer range that always has a non-negative length                               |
```
abey79, line 230: The figure needs updating as well.

crates/store/re_chunk/src/lib.rs
```diff
@@ -37,11 +37,12 @@ pub use self::batcher::{
...
 #[doc(no_inline)]
-pub use arrow::array::Array as ArrowArray;
-#[doc(no_inline)]
-pub use re_log_types::{EntityPath, TimeInt, TimePoint, Timeline, TimelineName};
-#[doc(no_inline)]
-pub use re_types_core::{ArchetypeFieldName, ArchetypeName, ChunkId, ComponentName, RowId};
+pub use {
+    arrow::array::Array as ArrowArray,
+    re_log_types::{EntityPath, TimeInt, TimePoint, Timeline, TimelineName},
+    re_span::Span,
```
abey79, line 43: I'm not fan of this re-export. It has the potential to confuse the LSP (and the dev) with a toss of coins whether `re_span::Span` or `re_chunk::Span` gets included. I'd much rather have `re_chunk` re-export `external::re_span` instead.

crates/store/re_log_encoding/src/decoder/streaming.rs
```diff
@@ -5,6 +5,7 @@ use tokio::io::{AsyncBufRead, AsyncReadExt as _};
 use tokio_stream::Stream;
 
 use re_build_info::CrateVersion;
+use re_chunk::Span;
```
abey79, line 8: As per my previous comment, this gave me pause. This use of `Span` hasn't directly to do with chunks, so this `use` is confusing.

crates/utils/re_span/README.md
```diff
@@ -0,0 +1,15 @@
...
+Part of the [`rerun`](https://github.com/rerun-io/rerun) family of crates.
+
+[![Latest version](https://img.shields.io/crates/v/re_span.svg?sepculative-link)](https://crates.io/crates/re_span?sepculative-link)
+[![Documentation](https://docs.rs/re_span/badge.svg)](https://docs.rs/re_span?sepculative-link)
+![MIT](https://img.shields.io/badge/license-MIT-blue.svg)
+![Apache](https://img.shields.io/badge/license-Apache-blue.svg)
+
+An integer range that always has a non-negative length.
+
+The standard `std::ops::Range` can have `start > end`.
```
abey79, line 12: [troll] Though I understand the logic, I'm often confused by your insistence of using `<` (e.g. in `if 10 < my_var {...}`), and here you managed to give me pause by using `>` 😅

crates/store/re_chunk_store/src/lib.rs
```diff
@@ -43,14 +43,14 @@ pub(crate) use self::store::ColumnMetadataState;
...
 #[doc(no_inline)]
-pub use re_chunk::{
-    Chunk, ChunkId, ChunkShared, LatestAtQuery, RangeQuery, RangeQueryOptions, RowId,
-    UnitChunkShared,
+pub use {
+    re_chunk::{
+        Chunk, ChunkId, ChunkShared, LatestAtQuery, RangeQuery, RangeQueryOptions, RowId, Span,
+        UnitChunkShared,
+    },
+    re_log_types::{ResolvedTimeRange, TimeInt, TimeType, Timeline},
```
abey79, line 51: Same comment as above re re-export, though I see we have an established pattern of doing that 😓 So this might be a taste issue....

abey79 (approved):
Very useful, I've certainly be annoyed with `Range`'s edge cases before. Pity we dont have this is `core` to begin with.

Issue #10280, opened 2025-06-17
Can't click top-bar buttons when fullscreen on macOS


emilk:
When going fullscreen on macOS, the buttons on the top bar (e.g. panel toggles) _sometimes_ aren't clickable.

Wumpf:
Kinda tricky to act upon this without a repro case. Do you have regular access to a setup that has this issue? Otherwise there's nothing we can do here to confirm or deny the existence of this issue in future versions

yash-s20:
can you try to reproduce this on an external monitor? that's where it fails for us. on the built-in display it works

yash-s20:
here's what it looks like on the welcome screen. I hope this helps

https://github.com/user-attachments/assets/815a8305-e014-4a33-a356-46395eec4439

yash-s20:
also, this is not run from inside a docker container.

Wumpf:
thanks, I was able to repro it now! Kinda strange because I'm always on external monitors, but maybe when I checked last on this I wasn't

Wumpf:
@emilk @lucasmerlin I strongly suspect this has to do with the menu bar being integrated in the top bar on Mac?
In fact, all menu elements in the top bar don't work and the menu button in particular has extra weird behavior

https://github.com/user-attachments/assets/f1caaf9d-5ed0-4922-b121-ba93d97da03f

emilk:
@Wumpf I think so to. Try setting `FULLSIZE_CONTENT = false` and see if that helps, just to confirm

https://github.com/rerun-io/rerun/blob/2a89e2222d973903ccf6bbe8c4e1a6d518d54469/crates/viewer/re_ui/src/lib.rs#L51-L53

This might be a bug in `winit`

Wumpf:
curious, `FULLSIZE_CONTENT = false` does still exhibit this behavior!

lucasmerlin:
I can reproduce this in the egui demo app, so I don't think this is a rerun only issue. I'll investigate!

https://github.com/user-attachments/assets/e5ba88ff-5ecb-4652-af1a-a0bb929a58eb

Clicking does funny things to the coordinates

lucasmerlin:
After a git bisect, this has apparently been broken since/by https://github.com/emilk/egui/pull/2279 or more recently https://github.com/emilk/egui/pull/3631/files
But looking at this PR it seems really weird that this change would introduce such bug 😅 Maybe some recent macos update triggers this now?

[...]

Issue #7324, opened 2024-08-31
docker container does not run on GPU


DiTo97:
**Describe the bug**
we have built a docker container pulling from the official NVIDA repository hub:

```bash
docker pull nvidia/cuda:11.8.0-devel-ubuntu22.04
```

and ran the procedure to install rerun following your documentation.

rerun works (ran a few examples), but detects only the CPU as rasterizing device. The container easily sees the GPU (NVIDIA A10) well, as we have other workloads in that container that run on the GPU with no issues at all when invoked.

we have also run rerun in the host machine (AWS EC2) outside the container and it works on the GPU just fine.

jleibs:
Can you see if the instructions from:
- https://github.com/rerun-io/rerun/issues/6835

address your issue?

DiTo97:
@jleibs, what's the roadmap for official docker support?

jleibs:
> @jleibs, what's the roadmap for official docker support?

This is not an item on our roadmap at the moment.

"Docker", unfortunately creates a combinatorial explosion of complex host/container environments. In my experience 95% of the issues we have seen have been related to configuration of the Docker or host environment rather than an actual issue with the Rerun view.

We do know that Docker deployments can be made to work.

The instructions from https://github.com/rerun-io/rerun/issues/6835 continue to work for me.

If it is helpful to your debugging, I am running on Arch Linux, using wayland via hyprland, and an RTX 4070 
```
>  nvidia-smi 
Wed Sep  4 17:22:59 2024       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 560.35.03              Driver Version: 560.35.03      CUDA Version: 12.6     |
|-----------------------------------------+------------------------+----------------------+
```

My docker, and nvidia-container package versions are:
```
docker 1:27.2.0-1
libnvidia-container 1.16.1-1
```

If you have followed the linked instructions verbatim, I would deduce the issue may have something to do with the host nvidia driver version, or even NVIDIA A10 support within driver or container runtime support version.

> we have other workloads in that container that run on the GPU

Can you confirm that these are vulkan applications that render to a display?  My go-to test application is typicall `vkcube`. If this finds your GPU but Rerun doesn't, that might help narrow in on the nature of the problem.

All that said, if you are unable to get Rerun to work inside the docker container, I would recommend using the viewer on one of the many supported native environments, or via the web, and just remotely access the data from the docker container without using that environment to do the rendering itself.

Issue #6835, opened 2024-07-09
Add troubleshooting or how-to guide related to running Rerun in a Docker container


jleibs:
This keeps coming up and even if we don't officially support it, would be good to set expectations and point people in the right direction.

At time of writing, with 0.17 the following works.

Install `nvidia-container-runtime`.

Make sure `/etc/docker/daemon.json` contains:
```
{
  "runtimes": {
    "nvidia": {
      "args": [],
      "path": "nvidia-container-runtime"
    }
  }
}
```

Make sure your docker image includes at least: `libgtk-3-dev libxkbcommon-x11-0 vulkan-tools`

Example `Dockerfile` for an ubuntu image.
```
FROM ubuntu:22.04

RUN apt-get update && apt-get install -y python3-pip libgtk-3-dev libxkbcommon-x11-0 vulkan-tools
RUN python3 -m pip install rerun-sdk==0.17.0
```

Jump through all the hoops necessary for GPU access, X authentication, etc.

In particular the following two seem to be crucial:
- `--runtime=nvidia`
- `-e NVIDIA_DRIVER_CAPABILITIES=all`

```
XSOCK=/tmp/.X11-unix
XAUTH=/tmp/.docker.xauth
xauth nlist $DISPLAY | sed -e 's/^..../ffff/' | xauth -f $XAUTH nmerge -
chmod 777 $XAUTH
docker run --runtime=nvidia --rm --gpus all -it --privileged --network=host -e NVIDIA_DRIVER_CAPABILITIES=all -e DISPLAY=$DISPLAY -v $XSOCK:$XSOCK -v $XAUTH:$XAUTH -e XAUTHORITY=$XAUTH rerun:0.17.0 rerun
```

Hillowwold:
Hi! I am able to display the Rerun viewer when I run this code on my ssh server, but the viewer is extremely slow and reaches the point where it is not possible. Is there any way to improve this please?

jleibs:
> Hi! I am able to display the Rerun viewer when I run this code on my ssh server, but the viewer is extremely slow and reaches the point where it is not possible. Is there any way to improve this please?

Can you clarify, are you trying to run the viewer on a remote machine, and then tunnel the viewer over SSH, such as via X-forwarding?  I would not expect this to work. The viewer is fairly graphics intensive and needs access to the local GPU.

[...]

PR #10306, opened 2025-06-19
Log a warning when starting the viewer from inside a Docker container

emilk:
Running the Rerun Viewer from inside a docker container has several problems:

* File dialogs may not work
* Clipboard may not work
* You may get crashes
* You may see graphics bugs
* You may have performance issues

So let's at least warn the user that this is a bad idea!

I also added `is_docker` to our analytics events so we can see how common this is amongst our users.

## Related
* https://github.com/rerun-io/rerun/issues/6835
* https://github.com/rerun-io/rerun/issues/7324
* https://github.com/rerun-io/rerun/issues/10280

## TODO
* [ ] Test running this from inside a docker container to see that the warning actually show up :)

### DIFF
--- a/crates/viewer/re_viewer/src/docker_detection.rs
+++ b/crates/viewer/re_viewer/src/docker_detection.rs
@@ -1,0 +1,27 @@
+use std::fs;
+use std::path::Path;
+
+/// Detect if the application is running inside a Docker container
+pub fn is_docker() -> bool {
+    is_dockerenv_present() || is_docker_in_cgroup()
+}
+
+/// Check for the presence of /.dockerenv file (most reliable method)
+fn is_dockerenv_present() -> bool {
+    Path::new("/.dockerenv").exists()
+}
+
+/// Check if 'docker' appears in cgroup information
+fn is_docker_in_cgroup() -> bool {
+    // Try multiple cgroup paths
+    let cgroup_paths = ["/proc/1/cgroup", "/proc/self/cgroup"];
+
+    for path in &cgroup_paths {
+        if let Ok(contents) = fs::read_to_string(path) {
+            if contents.contains("docker") {
+                return true;
+            }
+        }
+    }
+    false
+}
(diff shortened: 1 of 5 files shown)


crates/viewer/re_viewer/src/docker_detection.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    for path in &cgroup_paths {
+        if let Ok(contents) = fs::read_to_string(path) {
+            if contents.contains("docker") {
+                return true;
+            }
+        }
+    }
+    false
+}
```
abey79: It occurs to me that present and/or future OS might dislike an application trying to access files in "weird" locations. I'm 42.5% sure that if rerun was packaged as a proper mac application, this would have at least prompted a security dialog. I suggest we no-op everything to `false` on all targets but linux.

emilk: Oh yeah, I meant to do that but forgot - good catch!

Wumpf:
@abey79 this is good for landing but simply nobody tested it?

abey79:
> @abey79 this is good for landing but simply nobody tested it?

I think so yes? I sure didn't test it.

Wumpf:
dedusted my docker windows install, having it run in the background and do some building

Wumpf:
okay apparently running a graphical linux container on a windows host is pita. Someone with native Linux should try it

Wumpf:
@jleibs volunteered to give it a spin

Wumpf (approved):
~unfortunately nobody on the team who tried has been able to actually run Rerun in their Docker environments so far (not that we tried _a lot_)~
Looking over the changes this looks benign enough to ship it anyways. Don't see how this could cause issues and even if there were false positives it's not a problem

jleibs:
I can confirm this produces a warning on the CLI when running:
```
root@64ae1f4b82e9:/# rerun
[2025-07-15T19:00:09Z WARN  re_viewer::native] It looks like you are running the Rerun Viewer inside a Docker container. This is not officially supported, and may lead to performance issues and bugs. See https://github.com/rerun-io/rerun/issues/6835 for more.
```

However, that warning does not show up as a notification in the viewer itself.

image

emilk:
> However, that warning does not show up as a notification in the viewer itself.

* Fix in: https://github.com/rerun-io/rerun/pull/10786

PR #10310, opened 2025-06-19
Fix integer overflow in `TimeSeriesView` for too large timestamps

oxkitsune:
### What

This avoids overflow of timestamps in the `TimeSeriesView`


Wumpf (approved):
lgmtwhatcouldpossiblygowrong :)

Issue #10246, opened 2025-06-16
Improve the display of authentication errors in redap browser


abey79:
Currently, auth error appear as generic, badly explained error:

![Image](https://github.com/user-attachments/assets/baacc01e-371d-422a-bb99-37aaaabeb6e7)

We should make it explicit to the user that this is a auth failure.

MVP: same UI, juste update copy to make it clear that this is a auth failure
Stretch goal: maybe @gavrelina has input on how to improve the design.

abey79:
We rediscussed this again, and it would be pretty nice if we could handle auth error in such a way that:
1) make it more informative than scary
2) have a field to entre the token directly on the server central panel UI

This way, it would look somewhat similar to a logging prompt when you go to some web viewer that has the server pre-added but without a token.

PR #10311, opened 2025-06-19
Improved redap auth error Alert

lucasmerlin:
### Related

* closes https://github.com/rerun-io/rerun/issues/10246
* needs https://github.com/rerun-io/rerun/pull/10300

### What

Error for invalid token:

Screenshot 2025-06-19 at 18 36 49

Error for missing token:

I'd put a screenshot here but the sandbox auth doesn't seem to work right now?

### DIFF
--- a/crates/viewer/re_ui/data/light_theme.ron
+++ b/crates/viewer/re_ui/data/light_theme.ron
@@ -46,6 +46,51 @@
       "color": "{Blue.300}"
     },
 
+    "surface_success":{
+        "color": "{Green.500}",
+        "alpha": 25
+    },
+    "surface_info":{
+        "color": "{Blue.500}",
+        "alpha": 25
+    },
+    "surface_error":{
+        "color": "{Red.500}",
+        "alpha": 25
+    },
+    "surface_warning":{
+        "color": "#DF7700",
+        "alpha": 25
+    },
+    "border_success":{
+        "color": "{Green.500}",
+        "alpha": 75
+    },
+    "border_info":{
+        "color": "{Blue.500}",
+        "alpha": 75
+    },
+    "border_error":{
+        "color": "{Red.500}",
+        "alpha": 75
+    },
+    "border_warning":{
+        "color": "#DF7700",
+        "alpha": 75
+    },
+    "icon_content_success":{
+        "color": "{Green.500}"
+    },
+    "icon_content_info":{
+        "color": "{Blue.500}"
+    },
+    "icon_content_error":{
+        "color": "{Red.500}"
+    },
+    "icon_content_warning":{
+        "color": "#DF7700"
+    },
+
     "label_button_icon_color": {
       "color": "{Gray.550}"
     },
--- a/crates/viewer/re_ui/src/alert.rs
+++ b/crates/viewer/re_ui/src/alert.rs
@@ -1,0 +1,120 @@
+use crate::{DesignTokens, Icon, UiExt, icons};
+use egui::{Color32, InnerResponse, Response, Ui, Vec2};
+
+enum AlertKind {
+    Info,
+    Success,
+    Warning,
+    Error,
+}
+
+impl AlertKind {
+    fn stroke_color(&self, ui: &Ui) -> Color32 {
+        match self {
+            AlertKind::Info => ui.tokens().border_info,
+            AlertKind::Success => ui.tokens().border_success,
+            AlertKind::Warning => ui.tokens().border_warning,
+            AlertKind::Error => ui.tokens().border_error,
+        }
+    }
+
+    fn fill_color(&self, ui: &Ui) -> Color32 {
+        match self {
+            AlertKind::Info => ui.tokens().surface_info,
+            AlertKind::Success => ui.tokens().surface_success,
+            AlertKind::Warning => ui.tokens().surface_warning,
+            AlertKind::Error => ui.tokens().surface_error,
+        }
+    }
+
+    fn icon_color(&self, ui: &Ui) -> Color32 {
+        match self {
+            AlertKind::Info => ui.tokens().icon_content_info,
+            AlertKind::Success => ui.tokens().icon_content_success,
+            AlertKind::Warning => ui.tokens().icon_content_warning,
+            AlertKind::Error => ui.tokens().icon_content_error,
+        }
+    }
+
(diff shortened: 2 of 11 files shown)


crates/viewer/re_ui/src/design_tokens.rs
```diff
@@ -765,7 +797,7 @@ fn color_from_json(color_table: &ColorTable, color_alias: &ron::Value) -> anyhow
             let alpha = alpha
                 .as_u8()
                 .ok_or_else(|| anyhow::anyhow!("alpha should be an integer 0-255"))?;
-            color = color.gamma_multiply_u8(alpha);
+            color = color.linear_multiply(alpha as f32 / 255.0);
```
emilk: I'm not sure about this 🤔 
We should add a proper test to `ecolor`. It might be both `gamma_multiply_u8` and `linear_multiply` actually does the wrong thing compared to what is to be expected from normal CSS alpha behavior, i.e. test that `Color32::from_rgba_unmultiplied(r,g,b,a,) == Color32::from_rgb(r,g,b).multiply(a)` and that actually looks like `#RRGGBBAA` does in a browser when we render it (`crates/egui_demo_lib/src/rendering_test.rs`). I've messed this up so many times before.

crates/viewer/re_ui/data/light_theme.ron
```diff
@@ -46,6 +46,51 @@
...
+        "alpha": 100
+    },
+    "border_error":{
+        "color": "{Red.500}",
+        "alpha": 100
+    },
+    "border_warning":{
+        "color": "#DF7700",
+        "alpha": 100
+    },
```
emilk: It would be nice to add a struct for this instead, both here and in `struct DesignTokens`, so we could write this in our ron files:

```
alert_success: Alert {
    surface_color: {
        color: "{Green.500}",
        alpha: 25
    },
    border_color: "{Green.100}",
    icon_color: "{Green.500}",
}
```

crates/viewer/re_ui/Cargo.toml
```diff
@@ -78,7 +78,7 @@ time = { workspace = true, features = [
 
 [dev-dependencies]
 egui_kittest.workspace = true
-rand.workspace = true
+rand = { workspace = true, default-features = true }
```
emilk: do we need all the default features?

what do we need them for? I don't see anything else about `rand` in this PR?

lucasmerlin: It doesn't compile for me without rand default features (at least when I launch it from RustRover, from terminal it's fine for some reason). I assumed since this is dev dependencies, default features wouldn't hurt, but I can also look for the exact feature that's missing.

crates/viewer/re_ui/src/design_tokens.rs
lucasmerlin: Ooops, I actually wanted to revert this. I changed this because the colors didn't match figma but turns out we swapped the alpha values somewhere. 

I think gamma_multiply_u8 should be correct here, but interestingly changing this doesn't affect a single snapshot test 🤔

gavrelina:
I think the new style breaks some alignment in few places 😳, so far found just this one, I'd either center it or move to the next row:
image

Also this video error is so sad .. but it's a very specific context, so I would not change the styling of all the other errors only because this one is an ugly child. It should be a separate issue for me to introduce something nicer...
image

lucasmerlin:
> I think the new style breaks some alignment in few places 

Fixed!

crates/viewer/re_ui/src/alert.rs
```diff
@@ -0,0 +1,102 @@
...
+        match self {
+            Self::Info => icons::INFO,
+            Self::Success => icons::SUCCESS,
+            Self::Warning => icons::WARNING,
+            Self::Error => icons::ERROR,
+        }
+    }
+}
+
+pub struct Alert {
```
emilk, line 32: docstring

crates/viewer/re_ui/src/alert.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    fn frame(&self, ui: &Ui) -> egui::Frame {
+        let colors = self.kind.colors(ui);
+
+        egui::Frame::new()
+            .stroke((1.0, colors.stroke))
+            .fill(colors.fill)
+            .corner_radius(6)
+            .inner_margin(6.0)
+            .outer_margin(1.0) // Needed because we set clip_rect_margin. TODO(emilk): https://github.com/emilk/egui/issues/4019
```
emilk, line 65:
```suggestion
            .outer_margin(1.0) // Needed because we set clip_rect_margin. TODO(emilk/egui#4019): remove clip_rect_margin
```

crates/viewer/re_ui/src/design_tokens.rs
```diff
@@ -746,8 +779,15 @@ fn color_from_json(color_table: &ColorTable, color_alias: &ron::Value) -> anyhow
...
-        Ok(Color32::from_hex(color)
-            .map_err(|color_error| anyhow::anyhow!("Invalid hex color: {color_error:?}"))?)
+        let mut color = Color32::from_hex(color)
+            .map_err(|color_error| anyhow::anyhow!("Invalid hex color: {color_error:?}"))?;
+        if let Ok(alpha) = color_alias.get("alpha") {
+            let alpha = alpha
+                .as_u8()
+                .ok_or_else(|| anyhow::anyhow!("alpha should be an integer 0-255"))?;
+            color = color.gamma_multiply_u8(alpha);
+        }
```
emilk, line 789: duplicated code: this can be moved to after the `if-else` block to unify the handling

crates/viewer/re_view_dataframe/tests/snapshots/unknown_timeline_view_ui.png
emilk, line 1: I assume this change in color is intentional?

lucasmerlin: Yes, it's based on @gavrelina's design

gavrelina: yes! before it was a bit random, now I added logic to it 🙃

PR #10312, opened 2025-06-20
Rename to `tee` to `attach_sinks`

jprochazk:
Changing the name discussed in https://github.com/rerun-io/rerun/pull/10158#discussion_r2157086696

We also decided to add back the stateful API.


rerun_py/rerun_sdk/rerun/sinks.py
```diff
@@ -32,16 +32,44 @@ def is_recording_enabled(recording: RecordingStream | None) -> bool:
...
     *sinks: LogSinkLike,
     default_blueprint: BlueprintLike | None = None,
     recording: RecordingStream | None = None,
 ) -> None:
     """
     Stream data to multiple different sinks.
 
-    Duplicate sinks are not allowed. For example, two [`GrpcSink`]s that
+    Duplicate sinks are not allowed. For example, two [`rerun.GrpcSink`][]s that
     use the same `url`.
```
Wumpf: here and on `recording_stream`:

what happens when I do anyways? (docstring should tell ideally)

rerun_py/rerun_sdk/rerun/sinks.py
```diff
@@ -32,16 +32,44 @@ def is_recording_enabled(recording: RecordingStream | None) -> bool:
...
 
 
-def _tee(
+def attach_sinks(
     *sinks: LogSinkLike,
     default_blueprint: BlueprintLike | None = None,
     recording: RecordingStream | None = None,
 ) -> None:
     """
     Stream data to multiple different sinks.
```
Wumpf, line 41: here and on `recording_stream`:

what happens to any data logged so far, will those new sinks catch up (I reckon no).
I think this is important to clarify.

the interaction of `spawn=True` and `attach_sinks` is also worth pointing out on `rr.init`'s documentation

rerun_py/rerun_sdk/rerun/sinks.py
jprochazk: it throws a `ValueError`, added to docstrings

rerun_py/rerun_sdk/rerun/sinks.py
jprochazk: words are hard. it still seems a bit complicated but i've probably spent enough time trying to reword it, so i'll leave it as is in the latest commit and merge. the fact that it's not "easily" documentable suggests to me that what we have is overly complex, and we should head towards "first class sinks" as the only thing asap.

Issue #10298, opened 2025-06-18
Allow starting gRPC server without full message history


jprochazk:
There's a bunch of workflows where you don't need the gRPC server doubling your memory usage.

For example, using `rr.spawn` and the native Viewer. The Viewer spawns a server, and by the time that server is started, the Viewer is already guaranteed to be able to receive messages. The server and Viewer are in the same process, and the Viewer is not actually connected to it, instead it receives messages via an in-memory channel. We did that to remove the overhead of the gRPC protocol from something that should never have to go through a loopback socket interface.

The data received by the server is immediately ingested into the Viewer's chunk stores, but also kept around in history... That means we're currently doubling memory usage only for the relatively niche case where you may want to also connect another (either web or native) Viewer to that same server.

Wumpf:
It's quite shocking that this issue slipped our collective attention. Unless this turns out to be very complex, let's make this a shipblocker for 0.24 and in case it's possible backport to 0.23

We need to get a lot better at tracking end-to-end memory usage!

jprochazk:
Right now, if you set `--server-memory-limit=0B`, then you already get no history. The problem is we're still wasting a bit of compute every time a message is received. It's stored in history, then immediately GC'd out of it, because we hit the memory limit. We should not store messages at all if the memory limit is `0`.

jprochazk:
cc @jleibs 

The places where we need `--server-memory-limit=0B` by default are:
- `rerun` CLI spawning a native Viewer
- `spawn` in SDK

The following will continue to have the current non-zero default:
- Spawning a web viewer
- Running "headless" (server-only)
- `serve_grpc` in SDK

PR #10314, opened 2025-06-20
Disable gRPC server history for Spawn/CLI

jprochazk:
### Related

* Closes https://github.com/rerun-io/rerun/issues/10298#issuecomment-2987837283

### What

* Changes the default server memory limit to `0B` in:
  * Python, C++, Rust `spawn`
  * CLI


crates/store/re_grpc_server/src/lib.rs
```diff
@@ -484,6 +484,11 @@ impl EventLoop {
     fn handle_msg(&mut self, msg: LogMsgProto) {
         self.broadcast_log_tx.send(msg.clone()).ok();
 
+        if self.is_history_disabled() {
```
Wumpf, line 487: oh neat this part turned out way simpler than I thought it would

Issue #10315, opened 2025-06-20
`ComponentColumnDescriptor::is_static` and other metadata flags can be wrong


abey79:
Given a single chunk, it's impossible to know if a column is truly static, or semantically empty. That can only be determined by considering the entire set of chunks comprising a store/recording/partition. Our current implementation of `ComponentColumnDescriptor` does NOT reflect this. As a result, it's `is_static` field (and other metadata flag) can be wrong, depending on how the schema was computed.

- `ChunkBatch::chunk_schema()` -> WRONG (potentially)
- `ChunkStore::schema() -> correct
- `re_chunk_store::store_schema_from_chunk_stream()` -> correct

This discrepancy should absolutely be captured by our type system.

abey79:
Similar to this issue, a chunk's column being `is_sorted` is no indication that the recording/partition it belongs to has this column actually sorted. As a result, the `is_shorted` metadata should be dropped each time metadata is aggregated.

PR #10316, opened 2025-06-20
Introduce a way to accurately determine a store schema based on a stream of chunks

abey79:
### Related

- related to https://github.com/rerun-io/rerun/issues/10315
- sibling https://github.com/rerun-io/dataplatform/pull/990

### What

This PR introduces `re_sorbet::SchemaBuilder()` to accurately compute store schema based on a stream of chunks.

Background:
- Store schema are needed in the context of the dataframe API
- Previously, we used `ChunkStore` facilities to compute the schema (which was accurate)
- Because `redap` operates on a `Stream` of chunk, it instead used `SchemaBuilder:try_merge` on per-chunk schema.
- This approach yields *wrong* result, see #10315.
- This PR introduces the required tool to fix redap.

### DIFF
--- a/crates/store/re_chunk_store/src/lib.rs
+++ b/crates/store/re_chunk_store/src/lib.rs
@@ -19,7 +19,6 @@ mod drop_time_range;
 mod events;
 mod gc;
 mod query;
-mod schema_builder;
 mod stats;
 mod store;
 mod subscribers;
@@ -31,7 +30,6 @@ pub use self::{
     },
     events::{ChunkCompactionReport, ChunkStoreDiff, ChunkStoreDiffKind, ChunkStoreEvent},
     gc::{GarbageCollectionOptions, GarbageCollectionTarget},
-    schema_builder::SchemaBuilder,
     stats::{ChunkStoreChunkStats, ChunkStoreStats},
     store::{
         ChunkStore, ChunkStoreConfig, ChunkStoreGeneration, ChunkStoreHandle, ColumnIdentifier,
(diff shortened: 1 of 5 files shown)


crates/store/re_chunk_store/src/schema_builder.rs
teh-cmc, line 1: Feels like this belong in `re_sorbet` more than `re_chunk_store`? It's about schema semantics, not storage.

crates/store/re_chunk_store/src/schema_builder.rs
```diff
@@ -0,0 +1,112 @@
...
+    columns: HashMap<ColumnDescriptor, ColumnMetadata>,
+}
+
+impl SchemaBuilder {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Add a chunk to the builder.
+    pub fn add_chunk(&mut self, chunk: &Chunk) -> Result<(), re_chunk::ChunkError> {
```
teh-cmc: This won't be cheap is my guess
```suggestion
    #[tracing::instrument(level = "trace")]
    pub fn add_chunk(&mut self, chunk: &Chunk) -> Result<(), re_chunk::ChunkError> {
```

crates/store/re_chunk_store/src/schema_builder.rs
abey79: Agreed 100%, but `re_chunk` and `re_chunk_store` _depend_ on `re_sorbet`, so I cannot use them there, which I need to. `re_chunk_store` was my next best pick.

abey79: I could move it `re_sorbet` by:
- accepting `ChunkBatch` instead of `Chunk` (which is fine—we do that anyways)
- replacing `chunk.is_static()` by `chunk_batch.schema.chunk_columns.indices.len() == 0` (wrapped in an helper)

abey79: done

teh-cmc (approved):

PR #10324, opened 2025-06-23
Fix pypi notebook publish

jprochazk:
Not sure how this ever worked. Wish the GCP libraries had type annotations so we could've caught this earlier :grimacing:


jprochazk:
I'm gonna yolo merge this one, it's only a typo

Issue #10327, opened 2025-06-23
`AnyValues` breaks the dataframe API


abey79:
Static only queries do not return data logged (statically) with `rr.AnyValues)`.

### Repro

This script repros the issue *IF AND ONLY IF* the workaround listed below is disabled. This repro exists as test (`test_dataframe_static`), which passes thanks to the workaround.

```python
import rerun as rr
import pytest

from .test_dataframe import APP_ID


@pytest.fixture
def any_value_property(tmp_path) -> rr.dataframe.Recording:
    rrd_path = tmp_path / "tmp.rrd"

    with rr.RecordingStream(APP_ID) as rec:
        rec.save(rrd_path)
        rec.log("/test", rr.AnyValues(yak="yuk", foo="bar", baz=42), static=True)

    # Use this to exfiltrate the RRD file for debugging purposes:
    # import shutil
    # shutil.copy(rrd_path, "/tmp/exfiltrated.rrd")

    recording = rr.dataframe.load_recording(rrd_path)
    assert recording is not None

    return recording


def test_dataframe_static(any_value_property):
    recording = any_value_property

    view = recording.view(
        index=None  # static only
        contents="/**",
    )

    # Uncommenting this makes the test pass
    # view = view.fill_latest_at()

    table = view.select_static().read_all()

    # print(table)
    #
    # pyarrow.Table
    # /test:baz: list<item: int64>
    #   child 0, item: int64
    # /test:foo: list<item: string>
    #   child 0, item: string
    # /test:yak: list<item: string>
    #   child 0, item: string
    # ----
    # /test:baz: [[null]]
    # /test:foo: [[null]]
    # /test:yak: [[null]]

    assert table.column(0).to_pylist()[0] is not None
    assert table.column(1).to_pylist()[0] is not None
    assert table.column(1).to_pylist()[0] is not None
```

### Workaround

This is currently worked around with:

https://github.com/rerun-io/rerun/blob/c1b360ec4658cc8bfc4f2f3ec44a21272a89c2b7/rerun_py/src/catalog/dataframe_query.rs#L83-L88

(The same is done with the local API as well.)

abey79:
Dump of the faulty rrd:

```
StoreInfo {
    application_id: ApplicationId(
        "rerun_example_test_recording",
    ),
    store_id: StoreId {
        kind: Recording,
        id: "1df9972800f548908de20251f1bfcfec",
    },
    cloned_from: None,
    store_source: PythonSdk(
[...]

Issue #10193, opened 2025-06-10
Reconcile the legacy and dataset Dataframe Query API


abey79:
### Summary

The Dataframe Query API currently exists in two blends:

* the legacy API, which works on local recordings
* the new API, exposed by `Dataset`, which works on remote recordings

To the maximum extent possible, these API should have the same feature set and the same signatures. Ideally, they should also share as much implementation as possible.

### Local API synopsis

```python
recording = rr.dataframe.load_recording(path_to_some_rrd)
schema = recording.schema()

view: rr.dataframe.RecordingView = recording.view(index="timeline", contents: ...) # ViewContentsLike

view_schema = view.schema()

view = (
	view
		.filter_range_sequence(0, 100)
		.filter_is_not_null("/path:Scalar")
		.fill_latest_at()
)

record_batch_reader = view.select(columns=[…])  # AnyColumn

# alternatively, this forces one row even if all columns are static
record_batch_reader = view.select_static(columns=[...])
```

### Remote API synopsis

```python
client = rr.catalog.CatalogClient("https://redap.rerun.io")
dataset = client.get_dataset("my_dataset")

dataset_schema = dataset.schema()

view: rr.catalog.DataframeQueryView = dataset.dataframe_query_view(index="timeline", contents: ....) # Any

view = (
    view
		.filter_partition_id("episode 1", "episode 2", ....)
		.filter_range_sequence(0, 100)
		.filter_is_not_null("/path:Scalar")
		.fill_latest_at()
)

# return a datafusion dataframe registered in the `client`'s session context
df = view.df()

# or:
ctx = datafusion.SessionContext()
ctx.register_table_provider("my_table", view)

ctx.sql("SELECT * FROM my_table")
```

### Feature mismatch

Features only in legacy API:

* `view.schema()`
* `select_static()`
* provide result as `pa.RecordBatchReader`
* The list of column to select (passed to `select()`

Features only in new API:

* Ability to work on multiple recordings/partition.
* provide result as DataFusion table provider

Other differences:

* `view()` vs. `dataframe_query_view()`
* type of the `contents` argument (`ViewContentsLike ` vs. `Any`)

### Proposed changes

* `TODO`

### Related

* [https://github.com/rerun-io/dataplatform/issues/977](https://github.com/rerun-io/dataplatform/issues/977) ([DPF-1786](https://linear.app/rerun/issue/DPF-1786/api-and-contract-updates-for-static-data))
[...]

PR #10332, opened 2025-06-24
Add support for `index=None` to local and remote dataframe APIs and deprecate `select_static`

abey79:
### Related

- Part of https://github.com/rerun-io/rerun/issues/10193
- Closes https://github.com/rerun-io/dataplatform/issues/977
- Related https://github.com/rerun-io/rerun/issues/10327
- Related #10335

### What

This PR adds support for `recording.view(index=None, ...)` and `dataset.dataframe_query_view(index=None, ...)` to run static-only queries.

When `index=None`, the view will:
- have only static column (among the ones selected with `content`)
- yield a single row per recording/partition, containing the static data

This PR also deprecates `select_static()` which was the former way to achieve the same with the local dataframe API.


rerun_py/tests/unit/test_dataframe.py
```diff
@@ -459,7 +459,12 @@ def any_value_static_recording(tmp_path: pathlib.Path) -> rr.dataframe.Recording
 
     with rr.RecordingStream(APP_ID, recording_id=uuid.uuid4()) as rec:
         rec.save(rrd_path)
-        rec.log("/test", rr.AnyValues(yak="yuk", foo="bar", baz=42), static=True)
+        rec.log(
+            "/test",
+            # Note: defensive parameter names to avoid ANY_VALUE_TYPE_REGISTRY collisions
```
emilk: This comment will become very hard to understand if we ever change the name of `ANY_VALUE_TYPE_REGISTRY`.

Better to explain "We use long, explicit names here because `AnyValue` will remember and check that future calls use the same type for the same name" or similar

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1170,13 +1179,60 @@ class DatasetEntry(Entry):
...
+        included in the view, as determined by the `row_id` column. This will
+        generally be the last value logged, as row_ids are guaranteed to be
+        monotonically increasing when data is sent from a single process.
+
+        If `None` is passed as the index, the view will contain only static columns (among those
+        specified) and no index columns. It will also contain a single row per partition.
+
+        Parameters
+        ----------
+        index : str, optional
```
oxkitsune:
```suggestion
        index : str
```

This might just be convention, but saying this is an optional parameter is a bit confusing. It is NOT optional, you can just pass `None` to it right?

rerun_py/src/catalog/dataset_entry.rs
```diff
@@ -298,7 +298,52 @@ impl PyDatasetEntry {
...
+    /// included in the view, as determined by the `row_id` column. This will
+    /// generally be the last value logged, as row_ids are guaranteed to be
+    /// monotonically increasing when data is sent from a single process.
+    ///
+    /// If `None` is passed as the index, the view will contain only static columns (among those
+    /// specified) and no index columns. It will also contain a single row per partition.
+    ///
+    /// Parameters
+    /// ----------
+    /// index : str, optional
```
oxkitsune: same as above

rerun_py/src/dataframe.rs
```diff
@@ -1270,10 +1277,13 @@ impl PyRecording {
...
     /// monotonically increasing when data is sent from a single process.
     ///
+    /// If `None` is passed as the index, the view will contain only static columns (among those
+    /// specified) and no index columns. It will also contain a single row per partition.
+    ///
     /// Parameters
     /// ----------
-    /// index : str
-    ///     The index to use for the view. This is typically a timeline name.
+    /// index : str, optional
```
oxkitsune: same as above

rerun_py/rerun_bindings/rerun_bindings.pyi
abey79: ah, you're right!

PR #10333, opened 2025-06-24
Create a globally accessible analytics object

emilk:
### Related
* https://github.com/rerun-io/rerun/pull/10301

### What
In https://github.com/rerun-io/rerun/pull/10301 I want to send an analytics event the first time a user clicks a help-button… but there is no way to access the `Analytics` object deep down in `re_ui` without a lot of ugly plumbing. So this PR makes `Analytics` globally accessible so that we can send analytics from anywhere.

In the crash handler we now also use the global analytics object, instead of creating a new one (unless a new one is needed). This makes the analytics handling in the crash handler slightly cleaner and faster.

### DIFF
--- a/scripts/clippy_wasm/clippy.toml
+++ b/scripts/clippy_wasm/clippy.toml
@@ -86,6 +86,7 @@ disallowed-methods = [
   { path = "std::sync::mpsc::Receiver::iter", reason = "Cannot block on Web" },
   { path = "std::sync::mpsc::Receiver::recv_timeout", reason = "Cannot block on Web" },
   { path = "std::sync::mpsc::Receiver::recv", reason = "Cannot block on Web" },
+  { path = "std::thread::sleep", reason = "Can't block on web" },
   { path = "std::thread::spawn", reason = "Cannot spawn threads on wasm" },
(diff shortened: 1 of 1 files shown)


crates/utils/re_analytics/src/native/pipeline.rs
```diff
@@ -264,20 +279,26 @@ fn realtime_pipeline(
...
+            recv(ticker_rx) -> _elapsed => on_flush(&mut session_file),
             recv(event_rx) -> event => {
                 let Ok(event) = event else { break };
-                let Ok(event) = event else { break };
-                on_event(&mut session_file, event);
+                match event {
+                    PipelineEvent::Analytics(event) => on_event(&mut session_file, event),
+                    PipelineEvent::Flush => {
+                        on_flush(&mut session_file);
+                        flush_done_tx.send(()).ok();
```
lucasmerlin, line 295: Wouldn't this block forever on the next event if the flush is triggered without reading the event from the `flush_done_rx` receiver (e.g. by the tick channel)?

lucasmerlin: Ah of course it's only sent on a `PipelineEvent::Flush` and since that is only sent in flush_blocking, I guess this is fine

Issue #6487, opened 2024-06-03
Intermittent SIGBUS when spawning viewer from python


jleibs:
A user reported an intermittent crash when spawning the viewer from python on 0.16:
```
Rerun caught a signal: SIGBUS
Troubleshooting Rerun: https://www.rerun.io/docs/getting-started/troubleshooting 
Report bugs: https://github.com/rerun-io/rerun/issues 

[2024-06-02T22:05:44Z INFO  re_sdk_comms::server] Hosting a SDK server over TCP at 0.0.0.0:9876. Connect with the Rerun logging SDK.
[2024-06-02T22:05:44Z INFO  re_sdk_comms::server] New SDK client connected: 127.0.0.1:58043
[2024-06-02T22:05:44Z INFO  re_sdk_comms::server] New SDK client connected: 127.0.0.1:58044
   3: re_crash_handler::install_signal_handler::signal_handler
   4: __simple_esappend
   5: std::sys::common::thread_local::fast_local::Key<T>::try_initialize
   6: crossbeam_deque::deque::Stealer<T>::steal
   7: rayon_core::registry::WorkerThread::wait_until_cold
   8: rayon_core::registry::ThreadBuilder::run


Rerun caught a signal: SIGBUS
Troubleshooting Rerun: https://www.rerun.io/docs/getting-started/troubleshooting 
Report bugs: https://github.com/rerun-io/rerun/issues 
```

weidenka:
Some here on a Silicon Mac with MacOS 14.5 and `rerun-sdk-0.16.1`.
 
```shell
Rerun caught a signal: SIGBUS
Troubleshooting Rerun: https://www.rerun.io/docs/getting-started/troubleshooting
Report bugs: https://github.com/rerun-io/rerun/issues

   3: re_crash_handler::install_signal_handler::signal_handler
   4: __simple_esappend
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: <unknown>
  12: <unknown>
  13: <unknown>
  14: <unknown>
  15: <unknown>
  16: eframe::native::app_icon::AppTitleIconSetter::update
  17: <eframe::native::wgpu_integration::WgpuWinitApp as eframe::native::winit_integration::WinitApp>::run_ui_and_paint
  18: eframe::native::run::run_and_return::{{closure}}


Rerun caught a signal: SIGBUS
[...]

PR #10334, opened 2025-06-24
Fix `SIGBUS` startup crash on some Mac systems

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/6487
    * this issue discusses different SIGBUS crashes, but it fixes the most commonly discussed one - for everything else we don't know if they're still around and have no hold
* Gets this into rerun: https://github.com/emilk/egui/pull/7252


Wumpf:
nothing to review here, merging as is

PR #10336, opened 2025-06-24
Add `set_sinks` to Rust `RecordingStream`

jprochazk:
* Blocked by https://github.com/rerun-io/rerun/pull/10312

A "variadic" API using a trait implement for tuples up to some fixed size seemed like the right approach here. We want the external API to be reasonably ergonomic.

The API directly accepts `LogSink` implementors instead of some "descriptor" of a log sink. This is probably fine, because we already expose these in our public API anyway.

* [x] pass `main` ci


crates/top/re_sdk/src/log_sink.rs
```diff
@@ -408,6 +453,17 @@ impl GrpcSink {
     }
 }
 
+impl Default for GrpcSink {
+    fn default() -> Self {
+        use std::str::FromStr as _;
+        Self::new(
+            re_uri::ProxyUri::from_str("rerun+http://127.0.0.1:9876/proxy")
```
Wumpf: can we use `re_uri::DEFAULT_PROXY_PORT` here instead of _again_ hardcoding the port? 🤔

crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -345,6 +345,30 @@ impl RecordingStreamBuilder {
         Ok((rec, storage))
     }
 
+    /// Stream data to multiple different sinks.
+    ///
+    /// Duplicate sinks are not allowed. For example, two [`GrpcSink`][grpc_sink]s that
+    /// use the same `url`.
```
Wumpf: as on other PR, things I'd love to have documented:
* what does happen when I have dupl sinks
* what happens to previous data / how does it interact with existing sinks

Wumpf (approved):
nice! great to have this in Rust from the start

surely snippet-tests/`snippets.toml` is gonna complain about this snippet?
Only realized now that `attach_sinks` replaces all existing sinks. Name isn't exactly intuitive for that. Worth discussing

crates/top/re_sdk/src/log_sink.rs
```diff
@@ -118,6 +118,51 @@ impl LogSink for MultiSink {
     }
 }
 
+mod private {
+    pub trait Sealed {}
+}
+
```
Wumpf, line 124: quite nifty this whole thing there :)

jprochazk: Rust's type checker even recognizes subtraits of this as "sealed trait" in error messages

Wumpf: wow, didn't know!

crates/top/re_sdk/src/log_sink.rs
jprochazk: Oops, we already had `DEFAULT_CONNECT_URL`, can just use that.

crates/top/re_sdk/src/recording_stream.rs
jprochazk: Tried to document this a bit better. Calling it on `RecordingStream` is the same as calling `set_sink`, for `RecordingStreamBuilder` the semantics are more straightforward

jprochazk:
@rerun-bot full-check

crates/top/re_sdk/src/log_sink.rs
Wumpf: nice, even better

PR #10345, opened 2025-06-24
Add support for querying static chunks only

abey79:
### Related

- Follow up to #10332 
- Sibling https://github.com/rerun-io/dataplatform/pull/1032
### What

This PR adds support for querying stats chunks only in the grpc supporting structure, and use it for the dataframe query.

DNM:
- [x] chained
- [x] dataplatform update needed

### DIFF
--- a/crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
+++ b/crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
@@ -387,7 +387,7 @@ message QueryLatestAt {
   // Which index column should we perform the query on? E.g. `log_time`.
   //
   // Leave this empty to query for static data.
-  optional rerun.common.v1alpha1.IndexColumnSelector index = 1;
+  rerun.common.v1alpha1.IndexColumnSelector index = 1;
 
   // What index value are we looking for?
   //
--- a/crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
+++ b/crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
@@ -385,9 +385,13 @@ message Query {
 // This has the exact same semantics as the query of the same name on our `ChunkStore`.
 message QueryLatestAt {
   // Which index column should we perform the query on? E.g. `log_time`.
-  rerun.common.v1alpha1.IndexColumnSelector index = 1;
+  //
+  // Leave this empty to query for static data.
+  optional rerun.common.v1alpha1.IndexColumnSelector index = 1;
 
   // What index value are we looking for?
+  //
+  // Leave this empty to query for static data.
   optional int64 at = 2;
 
   // Which components are we interested in?
--- a/crates/store/re_protos/src/v1alpha1/rerun.manifest_registry.v1alpha1.ext.rs
+++ b/crates/store/re_protos/src/v1alpha1/rerun.manifest_registry.v1alpha1.ext.rs
@@ -259,6 +259,7 @@ pub struct QueryLatestAt {
     ///
     /// Use `None` for static only data.
     pub index: Option<String>,
+
     /// The timestamp to query at.
     ///
     /// Use `TimeInt::STATIC` to query for static only data.
--- a/crates/store/re_protos/src/v1alpha1/rerun.manifest_registry.v1alpha1.ext.rs
+++ b/crates/store/re_protos/src/v1alpha1/rerun.manifest_registry.v1alpha1.ext.rs
@@ -169,20 +169,7 @@ impl TryFrom<crate::manifest_registry::v1alpha1::Query> for Query {
 impl From<Query> for crate::manifest_registry::v1alpha1::Query {
     fn from(value: Query) -> Self {
         crate::manifest_registry::v1alpha1::Query {
-            latest_at: value.latest_at.map(|latest_at| {
-                crate::manifest_registry::v1alpha1::QueryLatestAt {
-                    index: latest_at.index.map(|index| {
-                        let timeline: TimelineName = index.into();
-                        timeline.into()
-                    }),
-                    at: if latest_at.at.is_static() {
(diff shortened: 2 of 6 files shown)


crates/store/re_protos/src/v1alpha1/rerun.manifest_registry.v1alpha1.ext.rs
```diff
@@ -266,7 +263,7 @@ pub struct FuzzyComponentDescriptor {
 
 #[derive(Debug, Clone)]
 pub struct QueryLatestAt {
-    pub index: String,
+    pub index: Option<String>,
     pub at: i64,
```
abey79: I'm not sure what to do with `at` here.

`TimeInt::STATIC` is `TimeInt(None)`, but serialise to `i64::MIN`. The grpc version of `at` was ALREADY `optional` for some reason. Was it meant to be for static though? Any preference on how to go about this?

crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
```diff
@@ -385,9 +385,13 @@ message Query {
...
   // Which index column should we perform the query on? E.g. `log_time`.
-  rerun.common.v1alpha1.IndexColumnSelector index = 1;
+  //
+  // Leave this empty to query for static data.
+  optional rerun.common.v1alpha1.IndexColumnSelector index = 1;
 
   // What index value are we looking for?
+  //
+  // Leave this empty to query for static data.
   optional int64 at = 2;
```
abey79: Related to my previous comment, I'm low-key surprised that this is already optional.

crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
```diff
@@ -385,9 +385,13 @@ message Query {
 // This has the exact same semantics as the query of the same name on our `ChunkStore`.
 message QueryLatestAt {
   // Which index column should we perform the query on? E.g. `log_time`.
-  rerun.common.v1alpha1.IndexColumnSelector index = 1;
+  //
+  // Leave this empty to query for static data.
+  optional rerun.common.v1alpha1.IndexColumnSelector index = 1;
```
teh-cmc: `message` types always use explicit presence tracking in proto3, so this `optional` has no effect.

crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
teh-cmc: Everything is always optional in protobuf: `optional` just makes it so that the serialized representation will explicitly keep track of whether the user explicitly made that thing `null`, or just didn't bother to specify it at all. This is very important in this instance so that the other end doesn't automatically deserialize a missing `at` as a `0`.

See https://protobuf.dev/programming-guides/field_presence/ for the complete background.

crates/store/re_protos/src/v1alpha1/rerun.manifest_registry.v1alpha1.ext.rs
teh-cmc: I need to think about this a bit more, but starting with `None` for static at the Protobuf level seems fine, and it's trivial to change course if it turns out it's not.

crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
abey79: Yeah, I probably didn't intend to leave that particular change.

crates/store/re_protos/src/v1alpha1/rerun.manifest_registry.v1alpha1.ext.rs
```diff
@@ -266,13 +255,51 @@ pub struct FuzzyComponentDescriptor {
 
 #[derive(Debug, Clone)]
 pub struct QueryLatestAt {
-    pub index: String,
-    pub at: i64,
+    /// Index name (timeline) to query.
+    ///
+    /// Use `None` for static only data.
+    pub index: Option<String>,
+    /// The timestamp to query at.
```
teh-cmc:
```suggestion
    pub index: Option<String>,

    /// The timestamp to query at.
```

crates/store/re_protos/src/v1alpha1/rerun.manifest_registry.v1alpha1.ext.rs
```diff
@@ -266,13 +255,51 @@ pub struct FuzzyComponentDescriptor {
...
+        crate::manifest_registry::v1alpha1::QueryLatestAt {
+            index: value.index.map(|index| {
+                let timeline: TimelineName = index.into();
+                timeline.into()
+            }),
+            at: if value.at.is_static() {
+                None
+            } else {
+                Some(value.at.as_i64())
+            },
```
teh-cmc: Same here, if there's no index, then the value of `at` shouldn't even matter.

crates/store/re_protos/src/v1alpha1/rerun.manifest_registry.v1alpha1.ext.rs
```diff
@@ -266,13 +255,51 @@ pub struct FuzzyComponentDescriptor {
...
+    pub fn new_static() -> Self {
+        Self {
+            index: None,
+            at: TimeInt::STATIC,
+            fuzzy_descriptors: Vec::new(),
+        }
+    }
+
+    pub fn is_static(&self) -> bool {
+        self.index.is_none() && self.at.is_static()
```
teh-cmc: I'm not a big fan of the fact that the value of `self.at` even matters at all here. A lack of index should mean static, always, period.
(I wouldn't consider this to be a nit: these kinds of multi-field invariants open so many classes of bugs...)

teh-cmc (approved):

Issue #10273, opened 2025-06-17
0.23 archetype names & field names are not correctly ported


Wumpf:
It seems that 0.23 rrds are currently loading without any problems but the viewer subsequently doesn't read out any archetype (field) name data, i.e. not migrating it correctly

See internal Slack thread for example data https://rerunio.slack.com/archives/C033K5VS2KD/p1750119502077239

emilk:
We should figure out why this wasn't caught by our migrations tests (`pixi run check-backwards-compatibility`), and how we can fix that

emilk:
Fixed by releasing 0.23.4, and future problems will be solved by
* #10381

PR #10350, opened 2025-06-24
Release 0.23.4: Fix migration of old 0.22 rrd files

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/10273

### What
NOTE! This is a 0.23.4 release branch, as I want to get these fixes out ASAP.

#### Problem 1: sorbet meta key name confusion
At some point we renamed the keys of our meta-data, and forgot to update our code:

* `rerun.archetype` vs `rerun.archetype_name`
* `rerun.archetype_field` vs `rerun.archetype_field_name`

#### Problem 2: not writing new meta keys on migrate
Our migration code erroneously would keep the old metadata of all our fields.

The new code in this PR not only fixes this, but also retains all non-rerun metadata, because it's the right thing to do.

### TODO
* [x] Test that migrated files work on `main`
* [x] Port the second commit to `main`
* [ ] Port changelog to `main`


crates/store/re_sorbet/src/sorbet_batch.rs
```diff
@@ -160,6 +166,9 @@ impl SorbetBatch {
     /// Will automatically wrap data columns in `ListArrays` if they are not already.
     ///
     /// Will also migrate old types to new types.
+    ///
+    /// Non-Rerun metadata will be preserved (both at batch-level and column-level).
+    /// Rerun-metadata will be updated and added to the batch if needed.
```
grtlr: GitHub removed the suggestion feature for me 😠.

Should be "Rerun metadata"

grtlr (approved):
Ah nice, how subtle!

crates/store/re_sorbet/src/sorbet_batch.rs
```diff
@@ -173,20 +182,27 @@ impl SorbetBatch {
...
             batch.columns()
-        ) {
-            debug_assert_eq!(field.data_type(), column.data_type());
-        }
+        )
+        .map(|(old_field, mut new_field, column)| {
+            debug_assert_eq!(new_field.data_type(), column.data_type());
+
+            let mut metadata = old_field.metadata().clone();
+            metadata.extend(new_field.metadata().clone()); // overwrite old with new
```
grtlr, line 194: Would that work in a setting where we change keys as part of our migration? Because that currently happens on `main`.

emilk: It would, if we would call `rewire_tagged_components` at the top of `SorbetBatch::try_from_record_batch`… unfortunately we call it only in `ChunkBatch::try_from`, which I think is a separate bug (we should always rewire tagged components)

emilk: I'll make sure this is correct when I cherry-pick the commit into `main`

jprochazk (dismissed):
@emilk 
This PR needs to be based off of `release-0.23.4-alpha.2` :grimacing: it contains some important changes to the build process, but it was never fully released as 0.23.4

I think those changes can also just be cherry-picked in case you don't want to rebase

https://github.com/rerun-io/rerun/tree/release-0.23.4-alpha.2

emilk:
Good catch - I'll rebase

emilk:
RC job: https://github.com/rerun-io/rerun/actions/runs/15875469356

emilk:
Release job failed:

image

ci_docker/Dockerfile
```diff
@@ -1,42 +1,37 @@
...
+    glib2-devel \
+    gtk3-devel \
+    openssl-devel \
+    xcb-util-renderutil-devel \
+    xcb-util-devel \
+    xcb-util-wm-devel \
+    libxkbcommon-devel \
     python3-pip \
-    software-properties-common \
     sudo; \
```
emilk, line 33: nit: it would be easier to merge changes to this list if it was sorted

emilk: Also a bit weird that some things are grouped (atk, make/automake) and others not (xcb) 🤔

PR #10355, opened 2025-06-25
Fix exclusion of recording properties in `ResolvedEntityPathFilter::unresolved`

### LINKED ISSUE #10304: Undo is broken
Repro: open an example and close/hide/maximize a view. Then hit undo. Nothing happens.

Seems like we're writing to the blueprint store all the time, which break undo. We may only write on change.

grtlr:
### Related

* Closes #10304.

### What

Due to the exclusion of the recording properties in the `unresolved` rules of an `ResolvedEntityPathFilter`, a (non-existing) change of the entity path filters was emitted, which caused spamming the blueprint store.

This PR fixes this by removing the recording properties exclusion rule.

### DIFF
--- a/crates/store/re_log_types/src/path/entity_path_filter.rs
+++ b/crates/store/re_log_types/src/path/entity_path_filter.rs
@@ -508,6 +508,8 @@ impl EntityPathFilter {
 impl ResolvedEntityPathFilter {
     /// Turns the resolved filter back into an unresolved filter.
     ///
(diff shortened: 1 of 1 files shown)


Wumpf:
I gonna be _that guy_ and kindly ask for sprinkling that into the existing unit tests at the bottom of that file :)

grtlr:
I'm glad you're *that guy* 😊 — will do

Wumpf (approved):
nice, thanks for adding those tests!

PR #10358, opened 2025-06-25
Add "refresh" copy on hover

gavrelina:
Following Niko's feedback for day 3: https://linear.app/rerun/issue/DPF-1796/niko-feedback-triage

it was not clear what does refresh icon do. Adding hover. 
>UX: want on hover description of what this action does

Even better solution to come later: 
- https://linear.app/rerun/issue/DPF-1801/bottom-panel


emilk:
You need to run `cargo fmt`. Again, we should see if we can configure your editor to do so on save, because that's very convenient

PR #10359, opened 2025-06-25
Add `set_sinks` to C++ `RecordingStream`

jprochazk:
Adds `set_sinks` to the C++ API:

```cpp
rerun::RecordingStream stream("test");
stream.set_sinks(rerun::GrpcSink(), rerun::FileSink("data.rrd"));
```

### DIFF
--- a/crates/top/rerun_c/src/lib.rs
+++ b/crates/top/rerun_c/src/lib.rs
@@ -12,7 +12,7 @@ mod ptr;
 mod recording_streams;
 mod video;
 
-use std::ffi::{CString, c_char, c_uchar};
+use std::ffi::{CString, c_char, c_float, c_uchar};
 
 use arrow::{
     array::{ArrayRef as ArrowArrayRef, ListArray as ArrowListArray},
@@ -278,6 +278,48 @@ pub struct CTimeColumn {
 
     /// The sorting order of the times array.
     pub sorting_status: CSortingStatus,
+}
+
+/// Log sink which streams messages to a gRPC server.
+///
+/// The behavior of this sink is the same as the one set by `rr_recording_stream_connect_grpc`.
+///
+/// See `rr_grpc_sink` in the C header.
+#[repr(C)]
+pub struct CGrpcSink {
+    /// A Rerun gRPC URL
+    ///
+    /// Default is `rerun+http://127.0.0.1:9876/proxy`.
+    pub url: CStringView,
+
+    /// The minimum time the SDK will wait during a flush before potentially
+    /// dropping data if progress is not being made. Passing a negative value indicates no timeout,
+    /// and can cause a call to `flush` to block indefinitely.
+    pub flush_timeout_sec: c_float,
+}
+
+/// Log sink which writes messages to a file.
+///
+/// See `rr_file_sink` in the C header.
+#[repr(C)]
+pub struct CFileSink {
+    /// Path to the output file.
+    pub path: CStringView,
+}
+
+/// A sink for log messages.
+///
+/// See specific log sink types for more information:
+/// * [`CGrpcSink`]
+/// * [`CFileSink`]
+///
+/// See `rr_log_sink` and `RR_LOG_SINK_KIND` enum values in the C header.
+///
+/// Layout is defined in [the Rust reference](https://doc.rust-lang.org/stable/reference/type-layout.html#reprc-enums-with-fields).
+#[repr(u8)] // Same as `repr(C)`, but forced to an 8-bit tag.
+pub enum CLogSink {
+    GrpcSink { grpc: CGrpcSink } = 0,
+    FileSink { file: CFileSink } = 1,
 }
 
 #[repr(u32)]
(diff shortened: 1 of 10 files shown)


crates/top/rerun_c/src/lib.rs
```diff
@@ -280,6 +280,48 @@ pub struct CTimeColumn {
...
+/// A sink for log messages.
+///
+/// See specific log sink types for more information:
+/// * [`CGrpcSink`]
+/// * [`CFileSink`]
+///
+/// See `rr_log_sink` and `RR_LOG_SINK_KIND` enum values in the C header.
+///
+/// Layout is defined in [the Rust reference](https://doc.rust-lang.org/stable/reference/type-layout.html#reprc-enums-with-fields).
+#[repr(u8)] // Same as `repr(C)`, but forced to an 8-bit tag.
```
Wumpf: I'm a bit confused about the difference between `#[repr(C, u8)]` and `#[repr(u8)]`. What's written in the reference about it isn't all that helpful to me. Seems to be the same if all variants are `repr(C)` structs?
If so I like `#[repr(C, u8)]` better because it tells the full story, but guess it doesn't matter?

rerun_cpp/src/rerun/log_sink.hpp
```diff
@@ -0,0 +1,74 @@
...
+        union {
+            GrpcSink grpc;
+            FileSink file;
+        };
+    };
+
+    inline GrpcSink::operator LogSink() const {
+        LogSink sink{};
+        sink.kind = LogSink::Kind::Grpc;
+        sink.grpc = GrpcSink{url, flush_timeout_sec};
```
Wumpf:
```suggestion
        sink.grpc = *this;
```

at least I think that should work since auto-generated copy assignment is present

rerun_cpp/src/rerun/log_sink.hpp
```diff
@@ -0,0 +1,74 @@
...
+        return sink;
+    }
+
+    inline FileSink::operator LogSink() const {
+        LogSink sink{};
+        sink.kind = LogSink::Kind::File;
+        sink.file = FileSink{path};
```
Wumpf:
```suggestion
        sink.file = *this;
```

rerun_cpp/src/rerun/recording_stream.hpp
```diff
@@ -139,6 +140,29 @@ namespace rerun {
...
+        /// and the pathname must be `/proxy`.
+        ///
+        /// The default is `rerun+http://127.0.0.1:9876/proxy`.
+        ///
+        /// flush_timeout_sec:
+        /// The minimum time the SDK will wait during a flush before potentially
+        /// dropping data if progress is not being made. Passing a negative value indicates no
+        /// timeout, and can cause a call to `flush` to block indefinitely.
+        ///
+        /// This function returns immediately.
```
Wumpf: We use doxygen syntax for documenting parameters (because C++ docs look nice then!)

```suggestion
        /// Requires that you first start a Rerun Viewer by typing 'rerun' in a terminal.
        ///
        /// \param url
        /// The scheme must be one of `rerun://`, `rerun+http://`, or `rerun+https://`,
        /// and the pathname must be `/proxy`.
        ///
        /// The default is `rerun+http://127.0.0.1:9876/proxy`.
        ///
        /// \param flush_timeout_sec
        /// The minimum time the SDK will wait during a flush before potentially
        /// dropping data if progress is not being made. Passing a negative value indicates no
        /// timeout, and can cause a call to `flush` to block indefinitely.
        ///
        /// This function returns immediately.
```

Turns out this is also wrong on `connect_grpc` and on `spawn`, guess I wasn't around for those reviews or missed it. Easy c&p mistake to make! Would be nice if you could drive-by fix those as well, thank you

rerun_cpp/src/rerun/recording_stream.hpp
```diff
@@ -139,6 +140,29 @@ namespace rerun {
...
+        template <typename... Ts>
+        Error set_sinks(const Ts&... sinks) const {
+            LogSink out_sinks[] = {sinks...};
+            uint32_t num_sinks = sizeof...(Ts);
+            return try_set_sinks(out_sinks, num_sinks);
```
Wumpf, line 152: the more idomatic way would actually to call both methods `set_sinks` and not `try_set_sinks`, relying on C++'s crazy overloading resolution. But since this isn't exposed right now it's up to you 🤷 
(also maybe there's a corner case I'm not seeing atm)

rerun_cpp/tests/recording_stream.cpp
```diff
@@ -336,6 +336,59 @@ void test_logging_to_grpc_connection(const char* url, const rerun::RecordingStre
     }
 }
 
+SCENARIO("RecordingStream can construct LogSinks", TEST_TAG) {
```
Wumpf, line 339: nice, thanks for adding this!

scripts/lint.py
```diff
@@ -29,7 +29,7 @@
 else_return = re.compile(r"else\s*{\s*return;?\s*};")
 explicit_quotes = re.compile(r'[^(]\\"\{\w*\}\\"')  # looks for: \"{foo}\"
 ellipsis = re.compile(r"[^.]\.\.\.([^\-.0-9a-zA-Z]|$)")
-ellipsis_expression = re.compile(r"[\[(<].*\.\.\..*[\])>]")
+ellipsis_expression = re.compile(r"[\[\]\(\)<>\{\}]?.*\.\.\..*[\[\]\(\)<>\{\}]")
```
Wumpf, line 32: 😵‍💫

Wumpf (approved):
always impressive how much more wrangling goes into C++. Looking great though, nice work!

scripts/lint.py
jprochazk: Starting to get real tired of the ellipsis lint :sweat_smile:

Wumpf: m2, but I let Emil have it because it means I can easier argue for weird stuff I insist on for no reason once it comes to it 😁

crates/top/rerun_c/src/lib.rs
jprochazk: If I understand the reference correctly, the exact layout is:

```c
// repr(C)
struct Enum {
  u8 tag;
  union {
    f64 a;
    u64 b;
  }
}

// repr(u8)
union Enum {
  struct { u8 tag; f64 a; } a;
  struct { u8 tag; u64 b; } b;
}
```

Which is technically the same? But I'm surprised that the tag is _inside_ the union for primitive repr enums, and _outside_ for C repr enums...

I'll change this to `repr(C, u8)`

rerun_cpp/src/rerun/recording_stream.hpp
jprochazk: Ah yeah, I forgot that you can do that in C++. Too Rust-brained. I think I'd prefer to leave it as-is though :thinking:

crates/top/rerun_c/src/lib.rs
Wumpf: > Which is technically the same?

might cause differences for padding? wasn't there something about that on the RFC 🤔

PR #10364, opened 2025-06-25
Remove unneeded edit ui warning

grtlr:
### What

The warning here is unnecessary. As the comment above this check states it's not a problem to opt-out of showing an edit UI in this place.


Wumpf:
(merged without waiting for ci, no point torturing it for this)

Issue #10343, opened 2025-06-24
`ArrowMsg::batch` should be of type `re_sorbet::ChunkBatch`


grtlr:
Follow-up issue of [#10245](https://github.com/rerun-io/rerun/issues/10245)

Investigate if we can pass the parsed `ChunkBatch` around in `ArrowMsg` directly to avoid triggering migration code again, and to get rid of the `chunk_id` field on `ArrowMsg`.

The problem is that both `enum LogMsg` and `struct ArrowMsg` lives in `re_log_types`, and `re_sorbet` depends on `re_log_types`.

grtlr:
Removed this from the `0.24` project, because it's

* Not strictly required, but more importantly
* It's a lot of work (we need to change the dependency order between—currently `re_sorbet` depends on `re_log_types` and we would need it the other way around).

emilk:
Alternative suggestion: we change `struct ArrowMsg` to contain the unmigrated record batch, which would also allow us to print the raw (unmigrated) contents of an rrd using `rerun rrd print`

Issue #10322, opened 2025-06-23
Update `SorbetSchema::version` 0.24 to help with future migrations


grtlr:
@jochen said:

> @emilk In `re_sorbet`, we have the `rerun.version` field, which is currently set to "1". Should we start setting this to 0.24 and then change it to the current Rerun version, whenever we have migrations?

emilk:
* Related: https://github.com/rerun-io/rerun/issues/8748

PR #10365, opened 2025-06-25
Improve migration code

emilk:
This cleans up the sorbet migration code in several ways (see commit messages).

It also fixes a footgun: we would sometimes try to parse just the batch schema, without for migrating it. This is now forbidden. All parsing of sorbet data must be done on the batch-level, so that it is properly migrated.

If this becomes a problem, let me know.

### TODO
* [x] `@rerun-bot full-check` (failed with the same error as `main`)
* [x] Check dataplatform
   * [x] still compiles
   * [x] Passes `cargo test`

### DIFF
--- a/crates/store/re_sorbet/src/migrations/v0_24.rs
+++ b/crates/store/re_sorbet/src/migrations/v0_24.rs
@@ -121,7 +121,6 @@ pub fn rewire_tagged_components(batch: &ArrowRecordBatch) -> ArrowRecordBatch {
                 }
             }
 
-
             for (key, value) in &metadata {
                 debug_assert!(!key.starts_with("rerun."), "Metadata `{key}` (with value `{value}`) was not migrated to colon syntax.");
             }


emilk:
@rerun-bot full-check

crates/store/re_log_encoding/src/protobuf_conversions.rs
```diff
@@ -73,10 +73,10 @@ pub fn arrow_msg_from_proto(
         arrow_msg.compression().into(),
     )?;
 
-    let sorbet_schema = re_sorbet::ChunkSchema::try_from(batch.schema_ref().as_ref())?;
+    let chunk_id = re_sorbet::chunk_id_of_schema(batch.schema_ref())?.as_tuid();
```
emilk, line 76: The old code would try to parse a non-migrated schema, causing crashes when using old data.

The new code is maybe less elegant, but should be both be more robust and faster.

crates/store/re_sorbet/src/chunk_columns.rs
```diff
@@ -141,7 +141,10 @@ impl TryFrom<SorbetColumnDescriptors> for ChunkColumnDescriptors {
             return Err(SorbetError::MultipleRowIdColumns(row_ids.len()));
         }
 
-        let row_id = row_ids.pop().ok_or(SorbetError::MissingRowIdColumn)?;
+        let row_id = row_ids.pop().ok_or_else(|| {
+            re_log::debug!("Missing RowId column, but had these columns: {columns:#?}");
```
emilk, line 145: Just improving our chances of fixing this quicker next time

crates/store/re_sorbet/src/lib.rs
```diff
@@ -66,3 +66,28 @@ pub enum BatchType {
...
 }
+
+/// Get the chunk ID from the metadata of the Arrow schema
+/// of a record batch containing a sorbet chunk.
+///
+/// Returns one of:
+/// * `Ok`
+/// * [`SorbetError::MissingChunkId`]
+/// * [`SorbetError::ChunkIdDeserializationError`]
+pub fn chunk_id_of_schema(
```
emilk, line 78:
```suggestion
// TODO(#10343): remove this
pub fn chunk_id_of_schema(
```

Temporary until this is fixed:
* See https://github.com/rerun-io/rerun/issues/10343

crates/store/re_sorbet/src/migrations/make_list_arrays.rs
```diff
@@ -0,0 +1,61 @@
...
+    },
+    datatypes::{FieldRef as ArrowFieldRef, Schema as ArrowSchema},
+};
+
+use re_arrow_util::{ArrowArrayDowncastRef as _, into_arrow_ref};
+use re_log::ResultExt as _;
+
+/// Make sure all data columns are `ListArrays`.
+#[tracing::instrument(level = "trace", skip_all)]
+pub fn make_all_data_columns_list_arrays(
```
emilk, line 16: Old code; just moved

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -3,9 +3,36 @@
...
+    re_tracing::profile_function!();
+
+    if v0_23::matches_schema(&batch) {
+        // Migrations from pre-`v0.23` to `v0.23`:
+        batch = v0_23::reorder_columns(&batch);
+        batch = v0_23::migrate_tuids(&batch);
+        batch = v0_23::migrate_record_batch(&batch);
+    }
+
+    if true {
```
emilk, line 30:
```suggestion
    if true {
        // TODO(#10322): only do this if needed
```

crates/store/re_sorbet/src/migrations/v0_24.rs
```diff
@@ -28,7 +28,12 @@ pub fn rewire_tagged_components(batch: &ArrowRecordBatch) -> ArrowRecordBatch {
         .schema()
         .metadata()
         .keys()
-        .any(|key| key.starts_with("rerun."));
+        .any(|key| key.starts_with("rerun."))
+        || batch
+            .schema()
+            .fields()
+            .iter()
+            .any(|field| field.metadata().keys().any(|key| key.starts_with("rerun.")));
```
emilk, line 36: This can be removed if we do
* https://github.com/rerun-io/rerun/issues/10322

crates/store/re_sorbet/src/migrations/v0_24.rs
```diff
@@ -109,6 +121,7 @@ pub fn rewire_tagged_components(batch: &ArrowRecordBatch) -> ArrowRecordBatch {
                 }
             }
 
+
```
emilk:
```suggestion
```

crates/utils/re_auth/src/service/client.rs
```diff
@@ -20,12 +20,14 @@ impl AuthDecorator {
...
-                },
-            )?;
+            let token = jwt.0.trim();
+            let token =
+                format!("{TOKEN_PREFIX}{token}")
+                    .parse()
+                    .map_err(|err: InvalidMetadataValue| {
+                        error!("malformed token '{token}': {err}");
+                        Status::invalid_argument("malformed token")
+                    })?;
```
emilk, line 30: Kind of unrelated, but I couldn't progress without also fixing this (my token somehow ended up with a `\n` in it)

crates/store/re_sorbet/src/lib.rs
```diff
@@ -66,3 +66,29 @@ pub enum BatchType {
...
+/// * [`SorbetError::MissingChunkId`]
+/// * [`SorbetError::ChunkIdDeserializationError`]
+// TODO(#10343): remove this
+pub fn chunk_id_of_schema(
+    schema: &arrow::datatypes::Schema,
+) -> Result<re_types_core::ChunkId, SorbetError> {
+    let metadata = schema.metadata();
+    if let Some(chunk_id_str) = metadata
+        .get("rerun:id")
+        .or_else(|| metadata.get("rerun.id"))
```
grtlr, line 84: Nit: maybe create a helper method in `v0_23`? That way we won't forget to remove this check someday.

I did the same thing with `is_component_column`.

crates/store/re_sorbet/src/sorbet_schema.rs
```diff
@@ -111,11 +111,17 @@ impl SorbetSchema {
...
-
+impl SorbetSchema {
+    /// Parse an already migrated Arrow schema.
     #[tracing::instrument(level = "trace", skip_all)]
-    fn try_from(arrow_schema: &ArrowSchema) -> Result<Self, Self::Error> {
+    pub(crate) fn try_from_migrated_arrow_schema(
+        arrow_schema: &ArrowSchema,
+    ) -> Result<Self, SorbetError> {
+        debug_assert!(
+            !arrow_schema.metadata.contains_key("rerun.id"),
```
grtlr, line 121: Maybe consolidate witch `matches_schema` (same motivation as above)

grtlr (approved):
LGTM! Just some comments out of personal preference.

Thank you for cleaning this up!

abey79:
This broke the `DatafusionTableWidget`.

There, the flow goes as follows:

1) Get `RecordBatch`es from eg the server.
2) Register said `RecordBatch`es as a table to datafusion.
3) Make some query to datafusion, some of which may reference specific column names.
4) Get `RecordBatch`es as a result of said query.
5) Convert them to `SorbetBatch`es.
6) Display them in the UI.
7) Upon user interaction (e.g. sort criterion), goto (3) and update the query (again, possibly referencing a specific column name, such as the sort column).

The problem is that the physical column names seen in (6) and (7)  are now *different* from the ones seen in (3), because we're tempering with the underlying arrow schema (which is the only way to know of column physical names).

As a result, sort (and possibly other things) is broken on most columns:

image

<br/>

Possible work around:

A) insert step (1b): convert `RecordBatch` to `SorbetBatch` and back, to normalise the column names
B) at step (4), save a copy of the `Fields` and carry them over all the way to (7) to use them as physical names

I like neither. I will attempt (B) as a quick fix.

Issue #10339, opened 2025-06-24
`debug_assert!` sometimes triggered by "Switch to" to an already open partition


abey79:
#### Repro

Open 1 or 2 remote partition, then go back to the partition table and click "Switch to". In some case (maybe depending on the loading state of the recording?), it triggers this behaviour:
- On release, a blueprint store is displayed (noticeable with the "blueprint" timeline
- On debug, debug assert.

#### Trace
```
thread 'main' panicked at 'assertion `left == right` failed
  left: Blueprint
 right: Recording'
re_viewer_context/src/store_hub.rs:502
stack backtrace:
   8: core::panicking::panic_fmt
             at core/src/panicking.rs:75:14
   9: core::panicking::assert_failed_inner
  10: core::panicking::assert_failed
             at core/src/panicking.rs:364:5
  11: re_viewer_context::store_hub::StoreHub::set_active_recording_id
             at re_viewer_context/src/store_hub.rs:502:9
  12: re_viewer::app::App::run_system_command
             at re_viewer/src/app.rs:672:25
  13: re_viewer::app::App::run_pending_system_commands
             at re_viewer/src/app.rs:434:13
      <re_viewer::app::App as eframe::epi::App>::update
             at re_viewer/src/app.rs:2422:9
```

#### Related

- Possibly related to https://github.com/rerun-io/rerun/issues/9848

PR #10367, opened 2025-06-25
Fix switch to link showing when recording isn't loaded yet

lucasmerlin:
### Related

- closes https://github.com/rerun-io/rerun/issues/10339
### What

This ensures that the store id is not a `StoreKind:Blueprint` when searching the store for the loaded recording.


abey79 (approved):
Let's land this asap because we want to fix the next deployment, but please consider a follow-up pr with:
- the lookup logic factored as a well documented helper on `StoreBundle`
- add a "Loading" state by checking the existence of a `SmartReceiver` with the matching datasource (unsure if that info is available in this context though).

Issue #10335, opened 2025-06-24
Remote deprecated `rerun.dataframe.View.select_static` method


abey79:
The `select_static` method is deprecated in favour of using `STATIC_INDEX`. It should be removed after a release or two.

Related:
- https://github.com/rerun-io/rerun/pull/10332

abey79:
Closing this as the entire legacy dataframe API will be deprecated.

Issue #10212, opened 2025-06-12
Fix filtering in the dataframe view


grtlr:
Leftover from #10082.

Currently, the filtering in the dataframe query view seems to be broken.

grtlr:
I've added a TODO in `re_view_dataframe/src/view_query/ui.rs`.

PR #10369, opened 2025-06-25
Add additional Rust test cases for (static) dataframe queries

### LINKED ISSUE #10327: `AnyValues` breaks the dataframe API

grtlr:
### Related

* Part of #10212 
* Related #10332, and transitively
* Related #10193
* Related #10335
* Closes #10327

### What

Additional dataframe query tests in Rust. I had these laying around and I didn't want them to bit rot.

These new test cases are aimed particularly at #10327, which does not seem to need the workaround anymore.

### DIFF
--- a/crates/store/re_chunk_store/tests/dataframe.rs
+++ b/crates/store/re_chunk_store/tests/dataframe.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
 use arrow::array::{StringArray, UInt32Array};
+
 use re_chunk::{Chunk, RowId, TimePoint, TimelineName};
 use re_chunk_store::{
     ChunkStore, ChunkStoreConfig, QueryExpression, StaticColumnSelection, TimeInt,
@@ -11,8 +12,8 @@ use re_log_types::{
     example_components::{MyColor, MyLabel, MyPoint, MyPoints},
 };
 use re_sorbet::ChunkColumnDescriptors;
+use re_types::AnyValues;
 use re_types::AsComponents as _;
-use re_types::{AnyValues, ComponentDescriptor};
 
 #[test]
 /// Tests whether the store has the expected schema after populating it with a chunk.
@@ -177,8 +178,6 @@ fn schema_static_columns() -> anyhow::Result<()> {
 
     store.insert_chunk(&Arc::new(chunk1))?;
 
-    // Both
-
     let query = QueryExpression {
         view_contents: None,
         include_semantically_empty_columns: false,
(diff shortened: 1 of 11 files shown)


crates/store/re_chunk_store/tests/dataframe.rs
```diff
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
-use re_chunk::{Chunk, RowId, TimelineName};
+use arrow::array::{StringArray, UInt32Array};
+use re_chunk::{Chunk, RowId, TimePoint, TimelineName};
```
abey79: nit
```suggestion
use arrow::array::{StringArray, UInt32Array};

use re_chunk::{Chunk, RowId, TimePoint, TimelineName};
```

crates/store/re_chunk_store/tests/dataframe.rs
```diff
@@ -136,3 +139,113 @@ fn schema_for_query() -> anyhow::Result<()> {
...
+            .iter()
+            .map(|column| (column.component_descriptor(), column.is_static))
+            .collect::<Vec<_>>(),
+        vec![
+            (ComponentDescriptor::partial("baz"), true),
+            (ComponentDescriptor::partial("foo"), true),
+            (ComponentDescriptor::partial("yak"), true),
+            (MyPoints::descriptor_points(), false),
+        ]
+    );
```
abey79: this looks fine as is, but I think insta would have been a good match here, especially if it allows testing more of the `ChunkColumnDescriptors` structure. It does require a sane `Debug` impl though.

crates/store/re_dataframe/tests/static.rs
```diff
@@ -0,0 +1,109 @@
...
+            (ComponentDescriptor::partial("foo"), true),
+            (ComponentDescriptor::partial("yak"), true),
+        ]
+    );
+
+    let schema = query_handle.schema().clone();
+
+    let mut reader = RecordBatchIterator::new(query_handle.into_batch_iter().map(Ok), schema);
+
+    let batch = reader.next().expect("there should be at least one batch")?;
```
abey79, line 74: Using insta is even more advisable here, because of how annoying and partial the following boilerplate is. See e.g. `dataplatform/redap_manifest_registry/tests/` where it's used a lot.

abey79 (approved):
Nice thanks. Added some comments that are fine to ignore if you have more urgent stuff on your plate.

crates/store/re_dataframe/tests/static.rs
teh-cmc: Yeah I highly recommend using snapshots for query tests. Most of the query tests in the OSS repo predate our use of snapshots, and they are an absolute nightmare to maintain. All the new snapshot-based ones in Redap are a breeze though :lotus_position:

abey79 (approved):
Seems like #10327 is fixed indeed 🤷🏻 Well done with the new tests.

PR #10371, opened 2025-06-25
Fix C++ crash due to FFI struct mismatch

Wumpf:
### Related

* follow-up to https://github.com/rerun-io/rerun/pull/10314

### What

Luckily, this made `main` ci unhappy!

* [x] pass main ci


Wumpf:
@rerun-bot full-check

crates/top/rerun_c/src/lib.rs
```diff
@@ -118,6 +119,12 @@ impl CSpawnOptions {
         if !self.memory_limit.is_empty() {
             spawn_opts.memory_limit = self.memory_limit.as_str("memory_limit")?.to_owned();
         }
+        if !self.server_memory_limit.is_empty() {
+            spawn_opts.server_memory_limit = self
```
emilk, line 133: This function could use the destructor pattern so that we don't forget to assign to/read from any of the fields 🤷

PR #10375, opened 2025-06-26
Fix migration of indicators from column fields

grtlr:
### What

Indicators that are present on the `field.name()` were not properly migrated to `rerun:component` as they were missing the `rerun.components.` prefix. (💩)

It was not clear from the output of our compare command because all sorts of prefixes were trimmed. This PR also adds options to not trim anything to our chunk formatter. So, hopefully, no one will ever have to go through this again.

* [x] full build

### DIFF
--- a/crates/store/re_format_arrow/src/lib.rs
+++ b/crates/store/re_format_arrow/src/lib.rs
@@ -148,13 +148,13 @@ pub struct RecordBatchFormatOpts {
     /// If `true`, displays the individual columns' metadata too.
     pub include_column_metadata: bool,
 
-    /// If `true`, trims the Rerun prefixes from field names
+    /// If `true`, trims the Rerun prefixes from field names.
     pub trim_field_names: bool,
 
-    /// If `true`, trims the `rerun:` prefix from metadata values
+    /// If `true`, trims the `rerun:` prefix from metadata values.
     pub trim_metadata_keys: bool,
 
-    /// If `true`, trims known Rerun prefixes from metadata values
+    /// If `true`, trims known Rerun prefixes from metadata values.
     pub trim_metadata_values: bool,
 }
 


grtlr:
@rerun-bot full-check

crates/store/re_sorbet/src/migrations/v0_24.rs
```diff
@@ -65,10 +65,14 @@ pub fn rewire_tagged_components(batch: &ArrowRecordBatch) -> ArrowRecordBatch {
...
 
             if field.name().ends_with("Indicator") {
+                let field_name = field.name();
                 // TODO(#8129): Remove indicator components
-                if let Some(archetype) = metadata.remove("rerun.archetype") {
-                    metadata.insert("rerun:component".to_owned(), format!("{archetype}Indicator"));
-                }
+                re_log::debug_once!(
+                    "Moving indicator from field to component metadata field: {field_name}"
+                );
```
emilk, line 72: Doesn't the field name start with `rerun.components` already? Better safe than sorry
```suggestion
                );
                debug_assert!(!field_name.starts_with("rerun"), "We expected a short name here, but got {field_name:?}");
```

crates/top/rerun/src/commands/rrd/compare.rs
```diff
@@ -101,6 +101,8 @@ impl CompareCommand {
                     width: Some(800),
                     include_metadata: true,
                     include_column_metadata: true,
+                    trim_metadata_keys: false,
+                    trim_metadata_values: false,
```
emilk, line 106: nice

emilk:
@rerun-bot full-check (no need to wait for this to complete - I'm just curious)

grtlr:
@rerun-bot full-check

crates/store/re_format_arrow/src/lib.rs
```diff
@@ -96,15 +96,13 @@ impl std::fmt::Display for DisplayMetadata {
...
-                            value
-                        },
-                    )
+                    let key = if *trim_keys { trim_key(key) } else { key };
+                    let value = if *trim_values {
+                        trim_name(value)
+                    } else {
+                        value
+                    };
+                    format!("{prefix}{key}: {value}",)
```
emilk, line 105: Style: Instead of using `let` bindings, you could also use named parameters to the `format! macro:



```rs
format!("{prefix}{key}: {value}",
    key = …,
    value = …,
)
```

emilk (approved):

PR #10376, opened 2025-06-26
Small `Alert` improvements

lucasmerlin:
### Related

- follow up of #10311 

### What

Adds a test and does some slight improvements


crates/viewer/re_ui/tests/snapshots/alert.png
emilk, line 1: Very nice

emilk (approved):

PR #10380, opened 2025-06-26
Fix null-filtering UI in dataframe view

### LINKED ISSUE #10212: Fix filtering in the dataframe view

grtlr:
### Related

* Left-over from #6889.
* Fixes #10212.

### What

When switching to tagged components, we hand-waived over the null filtering UI. This PR fixes this and brings the UI into a tagged component world. The indicators still show up, but you know the drill (™️).

### Reviewing

It's probably best to look at the diff from before our refactoring:

To test, best to compare with stable web viewer: https://app.rerun.io/

### DIFF
--- a/crates/viewer/re_view_dataframe/src/view_query/ui.rs
+++ b/crates/viewer/re_view_dataframe/src/view_query/ui.rs
@@ -5,13 +5,14 @@ use re_chunk_store::ColumnDescriptor;
 use re_log_types::{
     EntityPath, ResolvedTimeRange, TimeInt, TimeType, Timeline, TimelineName, TimestampFormat,
 };
-use re_sorbet::{ColumnSelector, ComponentColumnSelector};
+use re_sorbet::ColumnSelector;
+use re_types::ComponentIdentifier;
 use re_types::blueprint::components;
 use re_ui::{TimeDragValue, UiExt as _, list_item};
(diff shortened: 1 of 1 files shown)


Wumpf:
>  The indicators still show up, but you know the drill (™️).

I do 😄. But since they're hidden elsewhere now we might as well hide them here consequently? 🤔

crates/viewer/re_view_dataframe/src/view_query/ui.rs
```diff
@@ -310,17 +318,11 @@ impl Query {
...
-            let filter_is_not_null = components::FilterIsNotNull::new(
-                active,
-                &filter_entity,
-                filter_component.component,
-            );
-
-            if original_filter_is_not_null.as_ref() != Some(&filter_is_not_null) {
-                self.save_filter_is_not_null(ctx, &filter_is_not_null);
-            }
+        if original_filter_is_not_null.as_ref() != Some(&filter_is_not_null) {
```
Wumpf: doesn't that mean that if the original filter was `None` because none was set in the blueprint, we're now saving something to the blueprint always? Afaik there's no directly visible negative effect on that, but it does mess with the undo stack and needlessly changes the blueprint

grtlr: Ah good catch, I just confirmed and fixed.

PR #10381, opened 2025-06-26
Add test that the right batch metadata is added/removed/retained

emilk:
### Related
* Part of https://github.com/rerun-io/rerun/issues/10273
* Cherry-picked from https://github.com/rerun-io/rerun/pull/10350

### What
Ensure that all the Rerun-specific metadata that is added to a record batch actually makes it into the record batch

* [x] full-check

### DIFF
--- a/crates/store/re_sorbet/src/sorbet_batch.rs
+++ b/crates/store/re_sorbet/src/sorbet_batch.rs
@@ -297,7 +297,7 @@ mod tests {
             "This should remain"
         );
         assert!(
-            ret.schema().metadata().contains_key("rerun:version"),
+            ret.schema().metadata().contains_key("sorbet:version"),
             "This should have been added"
         );
 


crates/store/re_sorbet/src/migrations/v0_23.rs
```diff
@@ -268,56 +268,13 @@ pub fn migrate_record_batch(batch: &ArrowRecordBatch) -> ArrowRecordBatch {
...
-            }
-        }
-
-        false
-    };
-
-    if !needs_reordering {
-        return batch.clone();
-    }
-
```
emilk: Not 100% sure what the bug was in the old code, but I hit a `debug_assert` when I shouldn't have. I opted to just do something a lot simpler (at the cost of two `Vec`-collects)

crates/store/re_sorbet/src/sorbet_batch.rs
```diff
@@ -203,3 +219,125 @@ impl SorbetBatch {
...
+        let old_metadata = add_metadata(old_metadata, (key, value));
+        field.with_metadata(old_metadata)
+    }
+
+    // TODO(apache/arrow-rs#7628): this will be easier soon
+    fn remove_metadata_from_field(to_arrow_field: ArrowField, arg: &str) -> ArrowField {
+        let mut metadata = to_arrow_field.metadata().clone();
+        metadata.remove(arg);
+        to_arrow_field.with_metadata(metadata)
+    }
```
grtlr: That's on main now :tada:.

crates/store/re_sorbet/src/migrations/v0_23.rs
grtlr: Clement tried to optimize the code for early outs. With the new migration code, we shouldn't really hit this code path all too often anymore, so 👍!

grtlr (approved):
Nice!

grtlr:
@rerun-bot full-check

emilk:
@rerun-bot full-check

PR #10382, opened 2025-06-26
Fix `--serve-web <file>`

jprochazk:
Follow-up to:
* https://github.com/rerun-io/rerun/pull/10314

The memory limit for web viewer spawned via CLI was `0B`, which is wrong. We now set the memory limit to `25%` by default if `--serve-web` is set and no explicit `--server-memory-limit` was set either. This means `--serve-web <file>` now works again.

### DIFF
--- a/crates/top/rerun/src/commands/entrypoint.rs
+++ b/crates/top/rerun/src/commands/entrypoint.rs
@@ -121,7 +121,8 @@ Example: `16GB` or `50%` (of system total)."
         long_help = r"An upper limit on how much memory the gRPC server (`--serve-web`) should use.
 The server buffers log messages for the benefit of late-arriving viewers.
 When this limit is reached, Rerun will drop the oldest data.
-Example: `16GB` or `50%` (of system total)."
+Example: `16GB` or `50%` (of system total).
+Default is `0B`, or `25%` if any of the `--serve-*` flags are set."
     )]
     server_memory_limit: Option<String>,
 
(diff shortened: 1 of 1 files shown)


crates/top/rerun/src/commands/entrypoint.rs
```diff
@@ -720,7 +720,18 @@ fn run_impl(
...
+            Some(v) => v.as_str(),
+            None => {
+                if args.serve_web {
+                    "25%"
+                } else {
+                    "0B"
+                }
+            }
+        };
+        re_log::info!("actual memory limit: {value}");
```
Wumpf: With info this always shows on cli, I think that is a bit confusing as is.
Not sure we _do_ always want to show it but either way we should make it clearer
```suggestion
        re_log::debug!("Setting server memory limit to: {value}");
```
The zero case is kinda special though, so that's _extra_ confusing if you're new to Rerun: what server? if it limits to zero bytes then why is it taking space? what does that even do?

crates/top/rerun/src/commands/entrypoint.rs
```diff
@@ -117,13 +117,13 @@ Example: `16GB` or `50%` (of system total)."
 
     #[clap(
         long,
-        default_value = "0B",
+        default_value = None,
```
Wumpf, line 120: I think it's worth documenting the default for `--serve-web`, might be surprising otherwise 🤔

crates/top/rerun/src/commands/entrypoint.rs
jprochazk: This was supposed to a debug log, I'll remove it. Or maybe leave it and set it to debug level

jprochazk:
That condition was also missing `--serve` (which despite being deprecated is still relevant) and `--serve-grpc`, so I added them in

PR #10385, opened 2025-06-26
Add bottom bar to data fusion table

lucasmerlin:
### Related

* closes https://github.com/rerun-io/dataplatform/issues/1011
* related https://github.com/rerun-io/dataplatform/issues/1070 (the refresh button will now only refresh the table, not the whole origin)
### What

Adds a bottom bar to the table:

https://github.com/user-attachments/assets/da685726-d31a-42b1-914e-fe91478c4b44

### DIFF
--- a/crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
+++ b/crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
@@ -11,7 +11,6 @@ use nohash_hasher::IntMap;
 use re_format::format_uint;
 use re_log_types::{EntryId, TimelineName, Timestamp};
 use re_sorbet::{ColumnDescriptorRef, SorbetSchema};
-use re_ui::list_item::ItemButton as _;
 use re_ui::menu::menu_style;
 use re_ui::{ContextExt as _, UiExt as _, icons};
 use re_viewer_context::{AsyncRuntimeHandle, ViewerContext};


crates/viewer/re_ui/src/time.rs
```diff
@@ -0,0 +1,54 @@
...
+        "less than a minute ago".to_owned()
+    } else if seconds < 3600 {
+        let minutes = seconds / 60;
+        format_plural(minutes, "minute")
+    } else if seconds < 24 * 3600 {
+        let hours = seconds / 3600;
+        format_plural(hours, "hour")
+    } else if seconds < 7 * 24 * 3600 {
+        let days = seconds / 86400;
+        format_plural(days, "day")
```
emilk, line 36: The meat of this would maybe fit better in `re_format/src/time.rs` as `format_duration_approximate` or whatever (taking either seconds as f64, or `std::time::Duration`)

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -349,24 +338,115 @@ impl<'a> DataFusionTableWidget<'a> {
...
+                        ui.tokens().text_subdued;
+                    ui.visuals_mut().widgets.active.fg_stroke.color = ui.tokens().text_default;
+
+                    egui::Sides::new().show(
+                        ui,
+                        |ui| {
+                            ui.set_height(height);
+
+                            ui.label("rows:");
+                            ui.strong(format_int(total_rows as i64));
```
emilk:
```suggestion
                            ui.strong(format_uint(total_rows));
```

here and below

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -349,24 +338,115 @@ impl<'a> DataFusionTableWidget<'a> {
...
+                total_columns,
+                visible_columns,
+                table_state.queried_at,
+            );
+
+            if refresh {
+                Self::refresh(ui.ctx(), &session_ctx, table_ref);
+            }
+
+            ui.vertical(|ui| {
```
emilk: An alternative here is to use `TopBottomBar::bottom::show_inside`. It has several benefits:

A) less code
B) more self-explanatory (it explains that it is a bottom-bar)
C) no need for the extra nesting (`show_inside` will eat up some space from the parent, then you can just keep adding to parent)

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -113,14 +111,13 @@ pub struct DataFusionTableWidget<'a> {
...
     /// Clears all caches related to this session context and table reference.
-    pub fn clear_state(
+    pub fn refresh(
         egui_ctx: &egui::Context,
         session_ctx: &SessionContext,
         table_ref: impl Into<TableReference>,
     ) {
         let id = id_from_session_context_and_table(session_ctx, &table_ref.into());
 
-        TableConfig::clear_state(egui_ctx, id);
```
emilk, line 123: Is this something we no longer want to do?

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -349,24 +338,115 @@ impl<'a> DataFusionTableWidget<'a> {
...
+            table_state.update_query(runtime, ui, new_blueprint);
+        }
+    }
+
+    fn bottom_bar_ui(
+        ui: &mut Ui,
+        ctx: &ViewerContext<'_>,
+        total_rows: u64,
+        total_columns: usize,
+        visible_columns: usize,
```
emilk: Nit: we're showing `{visible} out of {total}", but here the order is reversed

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -349,24 +338,115 @@ impl<'a> DataFusionTableWidget<'a> {
...
+        queried_at: Timestamp,
+    ) -> bool {
```
emilk: It's not clear from the signature what this `bool` means. Consider returning an `enum Action` instead (it's more future-proof anyways)

emilk (changes requested):
Please try it with a panel instead - I think the code will become much simpler

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
lucasmerlin: I think so, if I'd configured my columns in the way I like and then when refreshing the table it's all reset I'd be annoyed. But maybe we should add a reset button to the columns menu that clears the table config. @gavrelina?

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
lucasmerlin: Nice, much better indeed. Didn't occur to me to use the Panels for layouting like this.

crates/viewer/re_ui/src/time.rs
lucasmerlin: The Cargo.toml says `# Do NOT add any complex dependencies here. Keep this crate lightweight!` and I'd need the re_log_types crate to properly format the timestamp for > 7 days. I guess I could make it return option instead or extend it to cover months and years.

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -349,24 +337,117 @@ impl<'a> DataFusionTableWidget<'a> {
...
+        );
+
+        match action {
+            Some(BottomBarAction::Refresh) => {
+                Self::refresh(ui.ctx(), &session_ctx, table_ref);
+            }
+            None => {}
+        }
+
+        ui.vertical(|ui| {
```
emilk: Why does this need wrapping in `vertical` now?

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -349,24 +337,117 @@ impl<'a> DataFusionTableWidget<'a> {
...
+                    );
+                });
+            })
+            .response;
+
+        ui.painter().hline(
+            response.rect.x_range(),
+            response.rect.top(),
+            Stroke::new(1.0, ui.tokens().table_header_stroke_color),
+        );
```
emilk: Do we still need an explicit separator? The panel should have one built-in

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -349,24 +337,117 @@ impl<'a> DataFusionTableWidget<'a> {
...
+                    tokens.table_header_height(),
+                )])
+                .num_rows(num_rows)
+                .show(ui, &mut table_delegate);
+        });
 
         table_delegate.table_config.store(ui.ctx());
         drop(requested_sorbet_batches);
-        table_state.update_query(runtime, ui, new_blueprint);
+        if table_state.blueprint() != &new_blueprint {
```
emilk, line 377: Is this a separate bug fix, for stopping spamming to blueprint store?


@grtlr was this something you found with your undo-spam-detector?

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
lucasmerlin: Screenshot 2025-06-26 at 16 52 42

Without it I don't see one (even when I change the color via ui.style_mut 🤔)

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
lucasmerlin: I need this change so I could properly reset the state for the refresh. I could also have moved the reset past this, but I think this is cleaner anyways.

lucasmerlin: I doubt this affects the blueprint store though, the table blueprint is separate from that (which unfortunately also means there is no undo / redo for table actions yet)

PR #10389, opened 2025-06-26
Improve table column names

emilk:
### Related
* Closes https://linear.app/rerun/issue/OSS-1405/partition-table-view-ignores-column-names-from-server

### What
Adds special handling of column names for recording properties (both built-in and user ones).

Has a test now!

### Future work
* https://github.com/rerun-io/rerun/issues/10391

### DIFF
--- a/crates/store/re_sorbet/src/component_column_descriptor.rs
+++ b/crates/store/re_sorbet/src/component_column_descriptor.rs
@@ -273,24 +273,11 @@ impl ComponentColumnDescriptor {
             BatchType::Dataframe => {
                 let prefix = if let Some(suffix) = self
                     .entity_path
-                    .strip_prefix(&EntityPath::recording_properties())
+                    .strip_prefix(&EntityPath::recording_properties()) // TODO(#10318): remove
+                    .or_else(|| self.entity_path.strip_prefix(&EntityPath::properties()))
                 {
                     if suffix.is_root() {
-                        // Built-in recording properties, like `start_time`:
-                        return self
-                            .component_descriptor()
-                            .archetype_field_name()
-                            .to_owned();
-                    } else {
-                        // Something weird
-                        self.entity_path.to_string()
-                    }
-                } else if let Some(suffix) =
-                    self.entity_path.strip_prefix(&EntityPath::properties())
-                {
-                    // User-defined property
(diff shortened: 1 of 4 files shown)


crates/viewer/re_dataframe_ui/src/table_blueprint.rs
```diff
@@ -158,32 +157,54 @@ pub fn default_display_name_for_column(desc: &ColumnDescriptorRef<'_>) -> String
         ColumnDescriptorRef::RowId(_) | ColumnDescriptorRef::Time(_) => desc.display_name(),
```
emilk, line 157: I'm not sure why this is using `display_name` and the code below uses `column_name`… but that's for another day

crates/viewer/re_dataframe_ui/src/table_blueprint.rs
```diff
@@ -158,32 +157,54 @@ pub fn default_display_name_for_column(desc: &ColumnDescriptorRef<'_>) -> String
...
-            strip_recording_properties_prefix(&name)
-                .map(ToOwned::to_owned)
-                .unwrap_or(name)
+            }
         }
     }
 }
 
-/// Strip the `/__properties/recording:RecordingProperties:` prefix from a column name.
-fn strip_recording_properties_prefix(name: &str) -> Option<&str> {
```
emilk, line 177: Moved into `re_sorbet` so that
A) it is done everywhere
B) it is not stringly typed

crates/store/re_sorbet/src/component_column_descriptor.rs
```diff
@@ -271,19 +271,47 @@ impl ComponentColumnDescriptor {
...
+                    if suffix.is_root() {
+                        // Built-in recording properties, like `start_time`:
+                        return self
+                            .component_descriptor()
+                            .archetype_field_name()
+                            .to_owned();
+                    } else {
+                        // Something weird
+                        self.entity_path.to_string()
+                    }
```
jleibs: I think we can get rid of this code path now so that column names match what users will use.

`property:RecordingInfo:start_time` makes it clear what start_time this is and distinguish it from: `property:action:start_time` or something else the user might add.

PR #10390, opened 2025-06-26
Add `DataframeQueryView.to_arrow_reader()`

abey79:
### Related

- part of https://github.com/rerun-io/rerun/issues/10193 (https://linear.app/rerun/issue/OSS-1334/reconcile-the-legacy-and-dataset-dataframe-query-api)

### What

This adds (a prototype of) `DataframeQueryView.to_arrow_reader()`, which is implemented using a throw-aways datafusion session context and our existing table provider. I don't know if we're happy with this design in the long run, but that should allow us to test our queries using a rust-side, instrumented version of datafusion.

One big caveat: it collects all record batch before returning the reader. Reason: datafusion gives me an async stream and the reader is sync.

**Open questions**
- Is the caveat above acceptable? (My take: it is. This API is about quickly getting the entire thing, eg. for Pandas/Polars. If you are serious, go the DataFusion way instead.)
- ~Are we happy with the method name (`to_reader`)? Alternatives: `to_record_batches()`, `to_record_batches_reader()`~ (renamed to `to_arrow_reader()`
- Still TODO, but the intention is to replace `.select(columns)` by this in the local API. This will be technically a regression (no longer possible to choose which columns are in the record batches), albeit one that is easily worked around since the local API will also have the datafusion interface. Still, I'd like to confirm we're happy with this.

Convenient test script I used to tried it out:
```python

import uuid
from pathlib import Path

import numpy as np
import rerun as rr
import tqdm
from rerun.catalog import CatalogClient


def write_rrd(path: Path) -> str:
    partition_id = str(uuid.uuid4())
    with rr.RecordingStream("rerun_example_sandbox", recording_id=partition_id) as rec:
        rec.save(path)

        rec.log("/test", rr.Points2D.columns(positions=np.random.rand(100, 2), radii=5), static=True)

        timestamp = rr.TimeColumn("timeline", sequence=list(range(10000)))
        rec.send_columns("/test", [timestamp], rr.Points3D.columns(positions=np.random.rand(10000, 3)))

        timestamp = rr.TimeColumn("timeline", sequence=list(range(1000)))
        rec.send_columns("/test2", [timestamp], rr.Points3D.columns(positions=np.random.rand(1000, 3)))

        rec.log("/not_relevant", rr.AnyValues(yak="yuk", foo="bar", baz=42), static=True)
        rec.log("/not_relevant", rr.Points2D.columns(positions=np.random.rand(100, 2), radii=5), static=True)

        timestamp = rr.TimeColumn("timeline", sequence=list(range(10000)))
        rec.send_columns("/not_relevant", [timestamp], rr.Points3D.columns(positions=np.random.rand(10000, 3)))

        rec.flush()

    return partition_id


uris = []
partition_ids = []
for i in range(10):
    path = Path(f"/tmp/sandbox{i}.rrd")
    partition_ids.append(write_rrd(path))
    uris.append(f"file://{path.as_posix()}")


c = CatalogClient("rerun+http://127.0.0.1:51234")

try:
    ds = c.get_dataset(name="sandbox")
    ds.delete()
except Exception:
    pass

ds = c.create_dataset(name="sandbox")

for uri in tqdm.tqdm(uris):
    ds.register(uri)

view = ds.dataframe_query_view(index=None, contents={"/test": ["baz"]})
view = view.filter_partition_id(*partition_ids[:3])

print(view.to_reader().read_all())
```


rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1503,6 +1503,9 @@ class DataframeQueryView:
     def df(self) -> Any:
         """Register this view to the global DataFusion context and return a DataFrame."""
 
+    def to_reader(self) -> pa.RecordBatchReader:
```
jleibs: Proposed: `to_arrow_reader()` to make this a bit more explicit that we are doing a sort of ecosystem conversion. After making this call the returned object is just a pyarrow object and can't be used directly with Rerun APIs.

jleibs (approved):
This seems like a pretty useful escape hatch.

PR #10392, opened 2025-06-26
Replace `rerun:version` with `sorbet:version` to help with future migrations

### LINKED ISSUE #10322: Update `SorbetSchema::version` 0.24 to help with future migrations

grtlr:
### Related

* Part of #8748.
* Closes #10322.

### What

This PR replaces the `rerun:version` chunk metadata with a `sorbet:version` metadata that can be used to inform migrations, but crucially is decoupled from the Rerun version to avoid confusion.

The Sorbet version should follow semantic versioning and will be used to determine which migrations to apply.

We start with `sorbet:version: 0.1.0` until we hit a stable state. Previous migrations are mapped to this scheme (`0.0.1` and `0.0.2`) for consistency.

### Chunk store

image

### TODO

* [x] full check
* [x] Dataplatform companion PR: https://github.com/rerun-io/dataplatform/pull/1059


grtlr:
@rerun-bot full-check

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -1,39 +1,63 @@
...
-        // TODO(#10322): only do this if needed
-        // Migrations from `v0.23` to `v0.24`:
-        batch = v0_24::rewire_tagged_components(&batch);
+    if needs_migration_to_version(&batch, "0.1.0") {
+        // Corresponds to migrations from `v0.23` to `v0.24`:
+        batch = v0_0_2__to__v0_1_0::rewire_tagged_components(&batch);
     }
 
-    batch = make_all_data_columns_list_arrays(&batch, v0_24::is_component_column);
+    batch = make_all_data_columns_list_arrays(&batch, v0_0_2__to__v0_1_0::is_component_column);
```
emilk: unrelated really, but this check should always just be on latest version

```suggestion
    batch = make_all_data_columns_list_arrays(&batch, super::is_component_column);
```

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -1,39 +1,63 @@
...
+        batch = v0_0_1__to__v0_0_2::reorder_columns(&batch);
+        batch = v0_0_1__to__v0_0_2::migrate_tuids(&batch);
+        batch = v0_0_1__to__v0_0_2::migrate_record_batch(&batch);
     }
 
-    if true {
-        // TODO(#10322): only do this if needed
-        // Migrations from `v0.23` to `v0.24`:
-        batch = v0_24::rewire_tagged_components(&batch);
+    if needs_migration_to_version(&batch, "0.1.0") {
```
emilk: I suggest you write a helper function instead:

```
let sorbet_version = get_or_guess_sorbet_version(&batch);

if sorbet_version <= Version::new(0, 0, 2) {
    batch = migrate_to_v0_0_2(&batch);
}
if sorbet_version <= Version::new(0, 1, 0) {
    batch = migrate_to_v0_1_0(&batch);
}
```

crates/store/re_sorbet/src/sorbet_schema.rs
```diff
@@ -35,10 +35,14 @@ pub struct SorbetSchema {
...
-    /// The version of the Rerun schema.
-    const METADATA_VERSION: &'static str = "1";
+    const METADATA_KEY_VERSION: &'static str = "sorbet:version";
+
+    /// The version of the Sorbet schema.
+    ///
+    /// This is bumped everytime we require a migration, but notable it is
+    /// decoupled from the Rerun version to avoid confusion as there will not
+    /// be a new Sorbet version for each Rerun version.
+    const METADATA_VERSION: &'static str = "0.1.0";
```
emilk: This could be 

```suggestion
    const METADATA_VERSION: semver::Version = semver::Version::new(0, 1, 0);
```

for more type strength

emilk (changes requested):

crates/store/re_sorbet/src/sorbet_schema.rs
grtlr: Yes!

grtlr:
@rerun-bot full-check

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -1,17 +1,81 @@
-//! These are the migrations that were introduced for each version.
-//!
-//! For example [`v0_24`] contains the migrations that are needed
-//! to go from [`v0_23`] to [`v0_24`].
+#![allow(non_snake_case)]
```
emilk:
```suggestion
#![expect(non_snake_case)]
```

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -1,17 +1,81 @@
...
+        else {
+            re_log::debug_once!("Encountered batch without 'sorbet:version' metadata.");
+            // We still do our best effort and try to perform the migration.
+            return semver::Version::new(0, 0, 0);
+        };
+
+        match semver::Version::parse(version_found) {
+            Ok(version_found) => version_found,
+            Err(err) => {
+                re_log::error_once!("Could not parse 'sorbet:version': {err}");
```
emilk: This should include the version string

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -1,17 +1,81 @@
...
+mod v0_0_1__to__v0_0_2;
+mod v0_0_2__to__v0_1_0;
+
+/// This trait needs to be implemented by any new migrations. It ensures that
+/// all migrations adhere to the same contract.
+trait Migration {
+    const TARGET_VERSION: semver::Version;
+
+    /// The Sorbet version that corresponds to this record batch.
+    fn version(batch: &RecordBatch) -> semver::Version {
```
emilk: Why is this a trait method? We don't expect implementors to change its meaning, do we?

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -1,17 +1,81 @@
...
+        let Some(version_found) = batch
+            .schema_ref()
+            .metadata()
+            .get(SorbetSchema::METADATA_KEY_VERSION)
+        else {
+            re_log::debug_once!("Encountered batch without 'sorbet:version' metadata.");
+            // We still do our best effort and try to perform the migration.
+            return semver::Version::new(0, 0, 0);
```
emilk: Can we guess between 0.01 and 0.0.2 here based on existence of e.g. "rerun.version" vs "rerun:version" ?

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -1,17 +1,81 @@
...
+            }
+        }
+    }
+
+    /// Migrates a record batch from one Sorbet version to the next.
+    fn migrate(batch: RecordBatch) -> RecordBatch;
+}
+
+fn maybe_apply<M: Migration>(mut batch: RecordBatch) -> RecordBatch {
+    let version = M::version(&batch);
```
emilk: Since `maybe_apply` we will parse the version multiple times, which is unnecessary

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -1,17 +1,81 @@
...
+    if version < M::TARGET_VERSION {
+        re_log::debug!("Migrating from `v{version}` to `v{}`", M::TARGET_VERSION);
+        batch = M::migrate(batch);
+
+        let mut metadata = batch.schema().metadata().clone();
+        metadata.insert("sorbet:version".to_owned(), M::TARGET_VERSION.to_string());
+        let schema = Arc::new(arrow::datatypes::Schema::new_with_metadata(
+            batch.schema().fields.clone(),
+            metadata,
+        ));
```
emilk: This can use https://docs.rs/arrow/latest/arrow/array/struct.RecordBatch.html#method.schema_metadata_mut instead (it landed 3 hours ago!) - or add a TODO so we remember to update later (`// TODO(apache/arrow-rs#7628)`)

emilk: You can also use `re_arrow_util::insert_metadata`

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -1,17 +1,81 @@
...
+
+        // TODO(grtlr): clean this up when we have mutable record batches in `arrow-rs`.
```
emilk: …ah. :)

crates/store/re_sorbet/src/migrations/v0_0_1__to__v0_0_2.rs
```diff
@@ -14,6 +14,28 @@ use re_types_core::arrow_helpers::as_array_ref;
...
+
+impl super::Migration for Migration {
+    const TARGET_VERSION: semver::Version = semver::Version::new(0, 0, 2);
+
+    fn version(batch: &ArrowRecordBatch) -> semver::Version {
+        if matches_schema(batch) {
+            // By returning `v0.0.1`, we ensure that the migration always runs.
+            semver::Version::new(0, 0, 1)
+        } else {
+            Self::TARGET_VERSION
```
emilk: This is just super weird. This will classify everything as either 0.0.1 or 0.0.2. Better to just have a single `version` function that uses heuristics for older data

grtlr: Yeah, I was way overthinking this. Thanks for pointing that out!

crates/store/re_sorbet/src/migrations/mod.rs
grtlr: Yeah, I'm doing that now.

crates/store/re_sorbet/src/migrations/mod.rs
emilk: Coming VERY soon:
* https://github.com/rerun-io/rerun/pull/10399 :)

grtlr: Merged!

grtlr:
@rerun-bot full-check

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -1,17 +1,98 @@
...
+            "Migrating record batch from Sorbet 'v{source_version}' to 'v{}'.",
+            M::TARGET_VERSION
+        );
+        batch = M::migrate(batch);
+        batch
+            .schema_metadata_mut()
+            .insert("sorbet:version".to_owned(), M::TARGET_VERSION.to_string());
+        batch
+    } else {
+        batch
```
emilk, line 96: Let's also log a "Trying to load data from a newer version of Rerun…" here when `source_version > TARGET_VERSION`

emilk (approved):
Great - thanks!

crates/store/re_sorbet/src/migrations/mod.rs
grtlr: I'm not sure about that, it would output this multiples times (for each migration) every time we have up-to-date data.

grtlr: Ah, it would make sense when we have an early exit check for the current version maybe (before all the `maybe_apply`).

emilk: If you use `re_log::warn_once` it will only warn… once

emilk:
It would be nice to also land this test of the migration code:
* https://github.com/rerun-io/rerun/pull/10381

emilk (approved):

grtlr:
@rerun-bot full-check

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -20,20 +104,42 @@ pub fn migrate_record_batch(mut batch: RecordBatch) -> RecordBatch {
...
+                    re_log::trace!("Performing migrations...");
+                    batch = maybe_apply::<v0_0_1__to__v0_0_2::Migration>(&batch_version, batch);
+                    batch = maybe_apply::<v0_0_2__to__v0_1_0::Migration>(&batch_version, batch);
+                    batch
+                }
+            }
+            Ordering::Greater => {
+                re_log::warn!(
+                    "Found Sorbet version 'v{batch_version}' that is newer then current supported version 'v{}'. Try using a newer version of Rerun.",
+                    SorbetSchema::METADATA_VERSION
```
emilk:
```suggestion
                re_log::warn_once!(
                    "Found Sorbet version 'v{batch_version}' that is newer then current supported version 'v{}'. Consider updating Rerun!",
                    SorbetSchema::METADATA_VERSION
```

crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -20,20 +104,42 @@ pub fn migrate_record_batch(mut batch: RecordBatch) -> RecordBatch {
...
+                );
+                batch
+            }
+        },
+        Err(err) => {
+            re_log::error!("Skipping migrations due to error: {err}");
```
emilk:
```suggestion
            re_log::error_once!("Skipping migrations due to error: {err}");
```

PR #10399, opened 2025-06-26
Update to arrow 55.2

emilk:
### Related
* Contains https://github.com/apache/arrow-rs/pull/7664

### What
Full release log: https://github.com/apache/arrow-rs/releases/tag/55.2.0


grtlr (approved):
LGTM, and it's oh so nice!

Update should also be safe on the dataplatform side:

* https://github.com/rerun-io/dataplatform/pull/1075

PR #10402, opened 2025-06-27
Fix broken latency measurement

emilk:
## What
The decode timestamp was missing in some circumstances. Now it works again.

Test with:

```
pixi run rerun
```

```
cargo run -p log_benchmark --release -- --benchmarks image --connect
```

## Other
* Always show latency metrics when metrics are enabled (and we have a latency measurement)

### DIFF
--- a/crates/store/re_grpc_server/src/lib.rs
+++ b/crates/store/re_grpc_server/src/lib.rs
@@ -298,8 +298,16 @@ pub fn spawn_with_recv(
                 }
             };
             match msg {
-                Ok(msg) => {
-                    if channel_log_tx.send(msg).is_err() {
+                Ok(mut log_msg) => {
+                    // Insert the timestamp metadata into the Arrow message for accurate e2e latency measurements.
+                    // Note that this function is only called by the viewer
+                    // (that's what the message-receiver is connected to).
+                    log_msg.insert_arrow_record_batch_metadata(
+                        re_sorbet::timestamp_metadata::KEY_TIMESTAMP_VIEWER_IPC_DECODED.to_owned(),
+                        re_sorbet::timestamp_metadata::now_timestamp(),
+                    );
+
+                    if channel_log_tx.send(log_msg).is_err() {
                         re_log::debug!(
                             "message proxy smart channel receiver closed, closing sender"
                         );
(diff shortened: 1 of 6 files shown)


crates/store/re_grpc_server/src/lib.rs
```diff
@@ -298,8 +298,16 @@ pub fn spawn_with_recv(
                 }
             };
             match msg {
-                Ok(msg) => {
-                    if channel_log_tx.send(msg).is_err() {
+                Ok(mut log_msg) => {
+                    // Insert the timestamp metadata into the Arrow message for accurate e2e latency measurements.
+                    // Note that this function is only called by the viewer
```
Wumpf, line 303: tbh I'm too confused about our stack to be confident on this statement, but I trust you thought long and hard about it 😅

crates/store/re_sorbet/src/timestamp_metadata.rs
```diff
@@ -5,10 +5,10 @@
 use crate::ArrowBatchMetadata;
 
 /// When was this batch sent by the SDK gRPC log sink?
-pub const KEY_TIMESTAMP_SDK_IPC_ENCODE: &str = "sorbet.timestamp_sdk_ipc_encoded";
+pub const KEY_TIMESTAMP_SDK_IPC_ENCODE: &str = "rerun:timestamp_sdk_ipc_encoded";
```
Wumpf, line 8: good call!

crates/store/re_sorbet/src/timestamp_metadata.rs
```diff
@@ -63,10 +63,25 @@ impl TimestampMetadata {
         let grpc_encoded_at = metadata
             .get(KEY_TIMESTAMP_SDK_IPC_ENCODE)
             .and_then(|s| parse_timestamp(s.as_str()));
+
         let grpc_decoded_at = metadata
             .get(KEY_TIMESTAMP_VIEWER_IPC_DECODED)
             .and_then(|s| parse_timestamp(s.as_str()));
 
+        if cfg!(debug_assertions) {
+            // Missing both happens all the time - but missing just one is suspicious.
```
Wumpf, line 72: nice!

Wumpf (approved):
lgtm!

Issue #10396, opened 2025-06-26
Route PartitionID through event callbacks


jleibs:
The description of [OSS-1397](https://linear.app/rerun/issue/OSS-1397/internal-recording-id-in-viewer-does-not-match-partition-id) was apparently not explicit enough.

Although the fix from [https://github.com/rerun-io/rerun/pull/10387](https://github.com/rerun-io/rerun/pull/10387) makes the PartitionId visible in the selection panel it does not fix the actual problem — the recording_id that we expose through the JS callbacks is useless for doing the filtering operation we want to do, which requires the PartitionId.<br><br>See:

<img src="https://uploads.linear.app/93919c5e-abe4-4cd5-b77d-7e3eafe3e62d/d8b0e700-e179-42a7-afe4-c860a3c271b2/d433a663-378a-4689-9615-e4364fd828df?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzkzOTE5YzVlLWFiZTQtNGNkNS1iNzdkLTdlM2VhZmUzZTYyZC9kOGIwZTcwMC1lMTc5LTQyYTctYWZlNC1jODYwYTNjMjcxYjIvZDQzM2E2NjMtMzc4YS00Njg5LTk2MTUtZTQzNjRmZDgyOGRmIiwiaWF0IjoxNzUwOTYxNzkwLCJleHAiOjMzMzIxNTIxNzkwfQ.hhH4E14xoRRPMfm4QYZWySzBCuVObVQW_k1SSRK91JA " alt="image.png" width="1481" data-linear-height="1450" />

Done when `captured_event` contains a member that specifies the partition id.

jleibs:
I think it's still very nice to have, but I was able to speed up the motivating query enough that we can reduce the urgency on this one.

PR #10403, opened 2025-06-27
Add partition ID to `TimeUpdateEvent`

abey79:
### Related

- closes https://github.com/rerun-io/rerun/issues/10396

### What

This add a `partition_id` field to the `TimeUpdateEvent` send from the viewer to the notebook. It is set to the current partition id for a remote partition, and `None` otherwise.

### DIFF
--- a/rerun_js/web-viewer/index.ts
+++ b/rerun_js/web-viewer/index.ts
@@ -135,6 +135,7 @@ export type ViewerEvent =
 type ViewerEventBase = {
   application_id: string;
   recording_id: string;
+  partition_id?: string;
 }
 
 export type PlayEvent = ViewerEventBase & {


jprochazk (changes requested):
See the comment at the top of this file:

https://github.com/rerun-io/rerun/blob/f8a3c2408ba4bd5a4af8ac659e8f19f92c264ba7/crates/viewer/re_viewer/src/event.rs#L6-L8

Both the JS and Python callbacks need to be updated. The JS stuff is here:

https://github.com/rerun-io/rerun/blob/f8a3c2408ba4bd5a4af8ac659e8f19f92c264ba7/rerun_js/web-viewer/index.ts#L135-L138

jprochazk (approved):

Issue #10294, opened 2025-06-18
Video seeking backwards causes visible replay from GOP start


Wumpf:
internally the decoder has to start at the GOP start, but we shouldn't bore the user with playing all that back rapidly ofc (also wastes a lot of time)

afaik this also regressed for AssetVideo, not just VideoStream and worked fine in 0.23

PR #10404, opened 2025-06-27
Delay showing video wait-for-decoder spinner on seeks

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/10294

### What

Compared to 0.24 this refines the behavior on seek: we no longer show the spinner without delay on backwards and cross-gop seeks... which is most seeks
I think there's other more subtle behavior changes there, but when testing before/after that's the one thing that stood out.

Compared to `main` this (additionally) fixes a regression where we'd visibly replay from the start of a GOP on every 
(title and tags are for the 0.24 changelog purpose).

⚠️ This PR alone causes debug_assertion to trigger for all videos with b-frames in Firefox & Safari as well as very occasionally on Chrome. Turns out we stumbled into the cause of github.com/rerun-io/rerun/issues/7961
Separate fix for this is coming right after!

---

After trying more incremental approaches I ended up simplifying & condensing the logic around:
* should the video texture be updated with the incoming frame
* should we show a spinner
* should we wait for more up-to-date frames

all of these questions are related and I found it increasingly hard to reason with what we had in place before. I hope the new code & commentary is not just more robust but also easier to read!

### DIFF
--- a/crates/utils/re_video/src/time.rs
+++ b/crates/utils/re_video/src/time.rs
@@ -26,23 +26,6 @@ const fn const_round_f64(v: f64) -> i64 {
     }
 }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn test_const_round_f64() {
-        assert_eq!(const_round_f64(1.5), 2);
-        assert_eq!(const_round_f64(2.5), 3);
-        assert_eq!(const_round_f64(1.499999999), 1);
-        assert_eq!(const_round_f64(2.499999999), 2);
-        assert_eq!(const_round_f64(-1.5), -2);
-        assert_eq!(const_round_f64(-2.5), -3);
-        assert_eq!(const_round_f64(-1.499999999), -1);
-        assert_eq!(const_round_f64(-2.499999999), -2);
-    }
-}
-
 impl Time {
     pub const ZERO: Self = Self(0);
     pub const MAX: Self = Self(i64::MAX);
@@ -127,5 +110,22 @@ impl std::ops::Sub for Time {
     #[inline]
     fn sub(self, rhs: Self) -> Self::Output {
         Self(self.0.saturating_sub(rhs.0))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_const_round_f64() {
+        assert_eq!(const_round_f64(1.5), 2);
+        assert_eq!(const_round_f64(2.5), 3);
+        assert_eq!(const_round_f64(1.499999999), 1);
+        assert_eq!(const_round_f64(2.499999999), 2);
+        assert_eq!(const_round_f64(-1.5), -2);
+        assert_eq!(const_round_f64(-2.5), -3);
+        assert_eq!(const_round_f64(-1.499999999), -1);
+        assert_eq!(const_round_f64(-2.499999999), -2);
     }
 }
--- a/crates/utils/re_video/src/time.rs
+++ b/crates/utils/re_video/src/time.rs
@@ -14,6 +14,35 @@ impl Timescale {
 #[derive(Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Time(pub i64);
 
+/// Round a `f64` to the nearest `i64`.
+///
+/// Does not have exactly the same result as `round`, don't use in contexts where you care!
+/// Workaround for `f64::round` not being `const`.
+const fn const_round_f64(v: f64) -> i64 {
+    if v > 0.0 {
+        (v + 0.5) as i64
+    } else {
+        (v - 0.5) as i64
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_const_round_f64() {
+        assert_eq!(const_round_f64(1.5), 2);
+        assert_eq!(const_round_f64(2.5), 3);
+        assert_eq!(const_round_f64(1.499999999), 1);
+        assert_eq!(const_round_f64(2.499999999), 2);
+        assert_eq!(const_round_f64(-1.5), -2);
+        assert_eq!(const_round_f64(-2.5), -3);
+        assert_eq!(const_round_f64(-1.499999999), -1);
+        assert_eq!(const_round_f64(-2.499999999), -2);
+    }
+}
+
 impl Time {
     pub const ZERO: Self = Self(0);
     pub const MAX: Self = Self(i64::MAX);
@@ -32,7 +61,7 @@ impl Time {
 
     #[inline]
     pub const fn from_secs(secs_since_start: f64, timescale: Timescale) -> Self {
-        Self((secs_since_start * timescale.0 as f64 + 0.5) as i64)
+        Self(const_round_f64(secs_since_start * timescale.0 as f64))
     }
 
     #[inline]
--- a/crates/utils/re_video/src/time.rs
+++ b/crates/utils/re_video/src/time.rs
@@ -5,7 +5,7 @@ pub struct Timescale(u64);
 impl Timescale {
     pub const NANOSECOND: Self = Self(1_000_000_000);
 
-    pub fn new(v: u64) -> Self {
+    pub const fn new(v: u64) -> Self {
         Self(v)
     }
 }
@@ -31,22 +31,22 @@ impl Time {
     }
 
     #[inline]
-    pub fn from_secs(secs_since_start: f64, timescale: Timescale) -> Self {
-        Self((secs_since_start * timescale.0 as f64).round() as i64)
+    pub const fn from_secs(secs_since_start: f64, timescale: Timescale) -> Self {
+        Self((secs_since_start * timescale.0 as f64 + 0.5) as i64)
     }
 
     #[inline]
-    pub fn from_millis(millis_since_start: f64, timescale: Timescale) -> Self {
+    pub const fn from_millis(millis_since_start: f64, timescale: Timescale) -> Self {
         Self::from_secs(millis_since_start / 1e3, timescale)
     }
 
     #[inline]
-    pub fn from_micros(micros_since_start: f64, timescale: Timescale) -> Self {
+    pub const fn from_micros(micros_since_start: f64, timescale: Timescale) -> Self {
         Self::from_secs(micros_since_start / 1e6, timescale)
     }
 
     #[inline]
-    pub fn from_nanos(nanos_since_start: i64, timescale: Timescale) -> Self {
+    pub const fn from_nanos(nanos_since_start: i64, timescale: Timescale) -> Self {
         Self::from_secs(nanos_since_start as f64 / 1e9, timescale)
     }
 
--- a/crates/viewer/re_renderer/src/video/chunk_decoder.rs
+++ b/crates/viewer/re_renderer/src/video/chunk_decoder.rs
@@ -175,16 +175,7 @@ pub fn update_video_texture_with_frame(
         )
     });
 
-    let format = {
-        #[cfg(target_arch = "wasm32")]
-        {
-            copy_web_video_frame_to_texture(render_ctx, source_content, gpu_texture)
-        }
-        #[cfg(not(target_arch = "wasm32"))]
-        {
-            copy_native_video_frame_to_texture(render_ctx, source_content, gpu_texture)
-        }
-    }?;
+    let format = copy_frame_to_texture(render_ctx, source_content, gpu_texture)?;
 
     target_video_texture.source_pixel_format = format;
     target_video_texture.frame_info = Some(source_info.clone());
--- a/crates/viewer/re_renderer/src/video/chunk_decoder.rs
+++ b/crates/viewer/re_renderer/src/video/chunk_decoder.rs
@@ -18,6 +18,7 @@ use crate::{
 
 #[derive(Default)]
 struct DecoderOutput {
+    /// Frames sorted by PTS.
     frames: Vec<Frame>,
 
     /// Set on error; reset on success.
@@ -25,7 +26,8 @@ struct DecoderOutput {
 }
 
 /// Internal implementation detail of the [`super::player::VideoPlayer`].
-// TODO(andreas): Meld this into `super::player::VideoPlayer`.
+///
+/// Expected to be reset upon backwards seeking.
 pub struct VideoSampleDecoder {
     decoder: Box<dyn re_video::AsyncDecoder>,
     decoder_output: Arc<Mutex<DecoderOutput>>,
@@ -51,6 +53,15 @@ impl VideoSampleDecoder {
                         frame.info.presentation_timestamp
                     );
                     let mut output = decoder_output.lock();
+
+                    if let Some(last_frame) = output.frames.last() {
+                        debug_assert!(
+                            last_frame.info.presentation_timestamp
+                                < frame.info.presentation_timestamp,
+                            "Expect new incoming frames to be in increasing PTS order. We expect the sample decoder to be reset upon backwards seeking."
+                        );
+                    }
+
                     output.frames.push(frame);
                     output.error = None; // We successfully decoded a frame, reset the error state.
                 }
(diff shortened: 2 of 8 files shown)


crates/utils/re_video/src/time.rs
```diff
@@ -31,22 +31,22 @@ impl Time {
     }
 
     #[inline]
-    pub fn from_secs(secs_since_start: f64, timescale: Timescale) -> Self {
-        Self((secs_since_start * timescale.0 as f64).round() as i64)
+    pub const fn from_secs(secs_since_start: f64, timescale: Timescale) -> Self {
+        Self((secs_since_start * timescale.0 as f64 + 0.5) as i64)
```
emilk: nit: `as i64` (IIRC) round towards zero, so this this is wrong for negative `secs_since_start`

crates/viewer/re_renderer/src/video/chunk_decoder.rs
```diff
@@ -188,6 +155,43 @@ impl VideoSampleDecoder {
...
+    let format = {
+        #[cfg(target_arch = "wasm32")]
+        {
+            copy_web_video_frame_to_texture(render_ctx, source_content, gpu_texture)
+        }
+        #[cfg(not(target_arch = "wasm32"))]
+        {
+            copy_native_video_frame_to_texture(render_ctx, source_content, gpu_texture)
+        }
+    }?;
```
emilk: This could use the `copy_frame_to_texture` helper

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -2,22 +2,33 @@ use std::time::Duration;
...
 /// Same with showing a spinner: if we show it too fast, it is annoying.
-const DECODING_GRACE_DELAY: Duration = Duration::from_millis(400);
+///
+/// This is wallclock time and independent of how fast a video is being played back.
+const DECODING_GRACE_DELAY_BEFORE_REPORTING: Duration = Duration::from_millis(400);
+
+/// Video time duration we allow to lag behind before no longer updating the output texture.
+///
+/// Note that since this is video time based, not wallclock time.
+/// We can't do wallclock time here since this would because we don't know the playback speed.
```
emilk: broken sentence

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -129,64 +145,90 @@ impl VideoPlayer {
...
+            // * we're not already up to date.
+            let current_frame_info = self.video_texture.frame_info.as_ref();
+            let outdated_frame =
+                current_frame_info.is_none_or(|info| info.presentation_timestamp != requested_pts);
+
+            // * the decoded frame isn't too far behind.
+            //   This happens when catching up on a seek, very rapid playback or simply too slow decoding.
+            //   Without this tolerance, we'd get a "rubber-band effect" in playback where we show a lot of outdated frames before (hopefully) catching up.
+            //   This is especially common when seeking to the end of a large GOP since the decoder has to start from its beginning.
+            let timescale = video_description.timescale.unwrap_or(Timescale::new(30)); // Assume 30 time units per second if there's no scale. As good as any guess!;
```
emilk: Does this mean all guesses are equally bad? What if it is off by a few order of magnitude?

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -129,64 +145,90 @@ impl VideoPlayer {
...
+            let pts_diff = decoded_frame.info.presentation_timestamp - requested_pts;
+            let not_too_far_behind = pts_diff.duration(timescale) <= TOLLERATED_OUTPUT_DELAY;
```
emilk: Is there a risk `not_too_far_behind` will always be `false` if e.g. the software decoder takes more than TOLLERATED_OUTPUT_DELAY to decode a big 8k frame, or whatever?

emilk: Also: please avoid negations in names. `if !not_too_far_behind` hurts my tiny brain. `is_up_to_date` or similar is better

crates/utils/re_video/src/time.rs
Wumpf: yeah.. but `round()` is not const :(

crates/viewer/re_renderer/src/video/player.rs
Wumpf: yes that's correct. With this we will never show an updated picture if we're hanging further behind than the tolerance.
That is pretty bad, but I really don't know what's the alternative here :(

I considered for a while to do something special on user interaction driven seek operations, but if you seek and play then we're just the same spot again.

Wumpf: Happy to discuss this further, ideally on a call

crates/viewer/re_renderer/src/video/player.rs
Wumpf: Then we get the tolerances wrong, not more not less. We could pass in a fallback timescale for this (which is the timeline viewer) but I considered this overly complex.
Reminder: this is only relevant today for video stream data that is logged on a non temporal timeline. For video assets we always have a timescale and for video stream data on a temporal timeline we always set timescale to 1

crates/utils/re_video/src/time.rs
Wumpf: I guess I hack a good-enough `const round` in here (doesn't have to be as rigorous as the real `round`)

crates/viewer/re_renderer/src/video/chunk_decoder.rs
Wumpf: oops. sloppy!

crates/viewer/re_renderer/src/video/player.rs
Wumpf: had a long conversation about this. TODO:
* make this stateful - enter "outdated state" on tolerance exceeding, exit it when caught up (fully!)
* comment things:
    * suggest adding wallclock timers for tracking the time it takes for decoding individual frames (for debugging)
    * .. some stuff we don't do and why
* can we do this is number of frames? that seems to be more wayyy more robust!

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -8,16 +8,31 @@ use super::{VideoFrameTexture, chunk_decoder::VideoSampleDecoder};
...
+///
+/// This a number of frames on the presentation timeline and independent of
+/// sample order for decoding purposes.
+///
+/// Discarded alternatives:
+/// * use video time based tolerance:
+///   -> makes it depend on playback speed whether we hit the threshold or not
+/// * use a wall clock time based tolerance:
+///   -> any seek operation that leads to waiting for the decoder to catch up,
+///      would cause us to show in-progress frames until the tolerance is hit
```
emilk, line 34: ⭐

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -66,6 +81,12 @@ pub struct VideoPlayer {
     ///
     /// Only fully reset after a successful decode.
     last_error: Option<TimedDecodingError>,
+
+    /// The last time the decoder fully caught up with the frame we want to show, if ever.
+    last_time_caught_up: Option<Instant>,
+
+    /// If true, we won't show any frames until the decoder has caught up with the requested frame time.
+    dont_update_texture_until_caught_up: bool,
```
emilk: avoid negatives in names

```suggestion
    hide_texture_until_caught_up: bool,
```

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -66,6 +81,12 @@ pub struct VideoPlayer {
     ///
     /// Only fully reset after a successful decode.
     last_error: Option<TimedDecodingError>,
+
+    /// The last time the decoder fully caught up with the frame we want to show, if ever.
+    last_time_caught_up: Option<Instant>,
+
+    /// If true, we won't show any frames until the decoder has caught up with the requested frame time.
```
emilk: Why don't we want this to always be `true`?

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -439,71 +481,68 @@ impl VideoPlayer {
...
-        Ok(()) => {
-            *last_error = None;
-            Ok(())
-        }
-        Err(err) => {
-            if matches!(err, VideoPlayerError::EmptyBuffer) {
-                // No buffered frames
+/// Determine whether we should update the video texture with the latest decoder result.
+fn should_update_video_texture(
+    dont_update_texture_until_caught_up: &mut bool,
```
emilk: do not prevent yourself from avoiding unnecessary negations

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -439,71 +481,68 @@ impl VideoPlayer {
...
+            false
+        } else {
+            // Figure out how many frames we're behind. If this is higher than a certain tolerance, don't update the texture.
+            //
+            // Since frames aren't in presentation time order and may have varying durations (i.e. the video has variable frame rate),
+            // we have to successively use `latest_sample_index_at_presentation_timestamp`:
+            // Start at the desired sample and walk backwards from there until we find the sample for the actually produced frame.
+            let mut num_frames_behind = 0;
+            let mut sample = requested_sample;
+            loop {
```
emilk: Can this be extracted to its own function?

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -439,71 +481,68 @@ impl VideoPlayer {
...
+                let Some(current_sample) = sample else {
+                    // Sample doesn't exist anymore. Nothing we can do!
+                    break false;
```
emilk: style: using `break` here when it is equivalent to `return` is just adding confusion imho

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -439,71 +481,68 @@ impl VideoPlayer {
...
-                    if DECODING_GRACE_DELAY <= timed_error.time_of_first_error.elapsed() {
-                        // Report the error only if we have been in an error state for a certain amount of time.
-                        // Don't immediately report the error, since we might immediately recover from it.
-                        // Otherwise, this would cause aggressive flickering!
-                        return Err(timed_error.latest_error.clone());
-                    }
+                num_frames_behind += 1;
+                if num_frames_behind > TOLERATED_OUTPUT_DELAY_IN_NUM_FRAMES {
+                    // Too far behind. Add hysteresis to avoid flipping in and out of this state.
+                    *dont_update_texture_until_caught_up = true;
```
emilk: Maybe this `bool` should be called `is_catching_up` instead. How we react to it is a separate concern

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -439,71 +481,68 @@ impl VideoPlayer {
...
+                    break false;
```
emilk: This function is effectively setting two bools (`dont_update_texture_until_caught_up` and the `return` value). 

Do we really need the return value? Let's rename the state from `dont_update_texture_until_caught_up` to `is_catching_up`.

* If we are caught up, `is_catching_up` is `false` and we should update the texture
* If we're far behind, `is_catching_up` is `true` and we should not update the texture
* If we are only a bit behind, then we should only update the texture if `is_catching_up == false`.

So the decision on updating the texture is exactly the the state of `is_catching_up`?

The only corner cases I see is when a sample is missing or similar.
I think those cases can still fit under a single `bool` state (with an appropriate name) or a tri-state enum.

crates/viewer/re_renderer/src/video/player.rs
emilk: Reading later code I now understand this is part of the hysteresis state machine. That would be a lot more clear with an enum:

```rs
enum State {
    /// The decoder is caught up (to within [`TOLERATED_OUTPUT_DELAY_IN_NUM_FRAMES`] frames)
    Current,
    
    /// The decoder is catching up after a long seek
    Behind,
}
 ```

**Review by emilk — CHANGES_REQUESTED** [review]

File: crates/viewer/re_renderer/src/video/player.rs (hunk 12 above)
Comment (emilk) [reply crates/viewer/re_renderer/src/video/player.rs]: On second (or third) thought: let's make it an enum instead

File: crates/viewer/re_renderer/src/video/player.rs (hunk 10 above)
Comment (Wumpf) [reply crates/viewer/re_renderer/src/video/player.rs]: tried it a bit, but I found it's a bit of a weird concept standalone. I think with making this whole thing `is_decoder_catching_up` it's good enough

File: crates/viewer/re_renderer/src/video/player.rs [hunk 14]
```diff
@@ -129,69 +153,101 @@ impl VideoPlayer {
...
+                    &decoded_frame,
+                )?; // Update texture errors are very unusual, error out on those immediately.
+                self.video_texture.frame_info = Some(decoded_frame.info.clone());
+            }
+
+            // We apparently recovered from any errors we had previously!
+            // (otherwise we wouldn't have received a frame from the decoder)
+            self.last_error = None;
+        } else {
+            // If there was no on the sample decoder, the last decoded frame is what we previously reported (if any).
```
emilk: My (very) tired brain has problems understanding this

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -439,71 +490,98 @@ impl VideoPlayer {
...
-                // so it doesn't make a lot of difference.
-                if let Some(timed_error) = last_error {
-                    if DECODING_GRACE_DELAY <= timed_error.time_of_first_error.elapsed() {
-                        // Report the error only if we have been in an error state for a certain amount of time.
-                        // Don't immediately report the error, since we might immediately recover from it.
-                        // Otherwise, this would cause aggressive flickering!
-                        return Err(timed_error.latest_error.clone());
-                    }
+/// Given the current decoder delay state, update it based on the new requested frame and the last decoded frame.
+fn update_decoder_delay_state(
```
emilk, line 496:
```suggestion
#[must_use]
fn update_decoder_delay_state(
```

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -439,71 +490,98 @@ impl VideoPlayer {
...
-        });
+/// Determine whether the decoder is catching up with the requested frame within a certain tolerance.
+fn is_decoder_catching_up(
+    video_description: &re_video::VideoDataDescription,
+    requested_sample: Option<&re_video::SampleMetadata>,
+    decoded_frame_pts: Time,
+) -> bool {
+    let Some(requested_pts) = requested_sample.map(|s| s.presentation_timestamp) else {
+        // Desired sample doesn't exist. We're technically not catching up, but we may as well behave as if we are.
+        return true;
```
emilk: When would we arrive here?

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -439,71 +490,98 @@ impl VideoPlayer {
...
+    //
+    // Since frames aren't in presentation time order and may have varying durations (i.e. the video has variable frame rate),
+    // we have to successively use `latest_sample_index_at_presentation_timestamp`:
+    // Start at the desired sample and walk backwards from there until we find the sample for the actually produced frame.
+    let mut num_frames_behind = 0;
+    let mut sample = requested_sample;
+    loop {
+        let Some(current_sample) = sample else {
+            // Sample doesn't exist anymore. We're technically not catching up, but we may as well behave as if we are.
+            return true;
```
emilk: When does this happen? When we've GC:ed the oldest frames of a video?

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -439,71 +490,98 @@ impl VideoPlayer {
...
-                resolve_target: None,
-                ops: wgpu::Operations::<wgpu::Color> {
-                    load: wgpu::LoadOp::Clear(wgpu::Color::BLACK),
-                    store: wgpu::StoreOp::Store,
-                },
-            })],
-            ..Default::default()
-        });
+/// Determine whether the decoder is catching up with the requested frame within a certain tolerance.
+fn is_decoder_catching_up(
```
emilk: nit: could probably be named slightly better, e.g. `fn is_signfificantly_behind`

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -439,71 +490,98 @@ impl VideoPlayer {
...
+        if num_frames_behind > TOLERATED_OUTPUT_DELAY_IN_NUM_FRAMES {
+            return true;
+        }
+
+        // Check the sample prior to this one.
+        sample = video_description
+            .latest_sample_index_at_presentation_timestamp(
+                current_sample.presentation_timestamp - Time::new(1),
+            )
+            .and_then(|sidx| video_description.samples.get(sidx));
```
emilk: Could be a helper function for even better readability 🤷 

```suggestion
        sample = video_description.previous_sample(current_sample);
```

crates/viewer/re_renderer/src/video/player.rs
Wumpf: My tired brain kept this around when it shouldn't have

crates/viewer/re_renderer/src/video/player.rs
Wumpf: correct.y

PR #10406, opened 2025-06-27
Fix flickering spinner/error & bad error icon for video

Wumpf:
### What

Commit by commit if you want :)

I _think_ all of those are regressions, but so much moved here I'm not entirely sure tbh


crates/viewer/re_view_spatial/src/visualizers/video/mod.rs
```diff
@@ -102,6 +113,17 @@ fn show_video_error(
     video_size: glam::Vec2,
     entity_path: &EntityPath,
 ) {
+    // Register the full video bounds regardless for more stable default view extents for when the error
+    // goes in and out of existance.
+    // The size of the error icon depends on the bounds in turn, making this extra important!
```
emilk: Does it have to though? Why not always have the error have the same size as the video (if known), or a fixed size if not known?

crates/viewer/re_view_spatial/src/visualizers/video/mod.rs
```diff
@@ -161,6 +183,24 @@ fn show_video_error(
...
+                        glam::Vec3::from(world_from_entity.translation),
+                        glam::Vec3::from(world_from_entity.translation)
+                            + glam::Vec3::new(video_size.x, video_size.y, 0.0),
+                    ]
+                    .iter()
+                    .copied(),
+                ),
+                world_from_entity,
+            );
+        }
```
Wumpf: oops. this can go away! covered by `register_video_bounds_with_bounding_box` above

crates/viewer/re_view_spatial/src/visualizers/video/mod.rs
```diff
@@ -206,3 +246,32 @@ fn show_video_error(
...
+    // Only update the bounding box if this is a 2D view.
+    // This is avoids a cyclic relationship where the image plane grows
+    // the bounds which in turn influence the size of the image plane.
+    // See: https://github.com/rerun-io/rerun/issues/3728
+    if class_identifier != SpatialView2D::identifier() {
+        return;
+    }
+
+    // To avoid circular dependency of the bounds and the error rectangle (causing flickering),
+    // make sure the bounding box contains the entire (speculative) video rectangle.
```
Wumpf: comment only applicable to `show_video_error` above

crates/viewer/re_view_spatial/src/visualizers/video/mod.rs
Wumpf: I think we used to do that a while ago but one of us two decided that it was nicer for varying 2d view sizes, see comment a but further below

```rs
    // If we're in a 2D view, make the error rect take a fixed amount of view space.
    // This makes it look a lot nicer for very small & very large videos.
    if let Some(state) = ctx.view_state.as_any().downcast_ref::<SpatialViewState>() {
        if let Some(bounds) = state.visual_bounds_2d {
            // Aim for 1/8 of the larger visual bounds axis.
```

Wumpf: either way "registering" the full bounds here is a good idea because the video may show up eventually :)

Wumpf: talked about it: problem with the current approach is also that icons may cover other videos in the same view. bad!

PR #10408, opened 2025-06-27
Fetch static data first when loading Redap recording

teh-cmc:
* Sibling: https://github.com/rerun-io/dataplatform/pull/1079

### DIFF
--- a/crates/store/re_grpc_client/src/redap/mod.rs
+++ b/crates/store/re_grpc_client/src/redap/mod.rs
@@ -416,18 +416,39 @@ async fn stream_partition_from_server(
     on_cmd: &(dyn Fn(Command) + Send + Sync),
     on_msg: Option<&(dyn Fn() + Send + Sync)>,
 ) -> Result<(), StreamError> {
-    let catalog_chunk_stream = client
-        .inner()
-        .get_chunks(GetChunksRequest {
-            dataset_id: Some(dataset_id.into()),
-            partition_ids: vec![partition_id.into()],
-            chunk_ids: vec![],
-            entity_paths: vec![],
-            select_all_entity_paths: true,
-            query: None,
-        })
-        .await?
-        .into_inner();
+    let static_chunk_stream = {
+        client
+            .inner()
+            .get_chunks(GetChunksRequest {
+                dataset_id: Some(dataset_id.into()),
+                partition_ids: vec![partition_id.clone().into()],
+                chunk_ids: vec![],
+                entity_paths: vec![],
+                select_all_entity_paths: true,
+                exclude_static_data: false,
+                exclude_temporal_data: true,
+                query: None,
+            })
+            .await?
+            .into_inner()
+    };
+
+    let temporal_chunk_stream = {
+        client
+            .inner()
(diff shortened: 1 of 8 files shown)


abey79 (approved):
Code lgtm.

I have two concerns though, none of which needs addressing here (aka lets merge and deploy):
- The dual `exclude_{static|temporal}_chunks` flag is a very user visible and "noisy" feature to ultimately address something that could be entirely implemented server-side by a `sort_by("chunk_is_static")`. Do we have clear indication that the (assumed) other benefits of these flags outweigh the cognitive cost?
- At the grpc layer, there is now an overlap between the `index=None` semantics of the query, and the newly introduced flags. Does it even make sense to keep that? Or should we drop the former for the latter. (Obviously depends on the resolution of the first point.)

nikolausWest:
Did this improve the experience of loading data from the platform?

teh-cmc:
> - The dual `exclude_{static|temporal}_chunks` flag is a very user visible and "noisy" feature to ultimately address something that could be entirely implemented server-side by a `sort_by("chunk_is_static")`. Do we have clear indication that the (assumed) other benefits of these flags outweigh the cognitive cost?

* I don't think there's any extra cognitive cost here: the default is `false` for both, which behaves exactly like it always has (i.e. you could literally remove these fields from an hypothetical "public version" of the proto definition, and users wouldn't be able to tell the difference).
* Sorting is not an option: everything is both computed and fetched in parallel on the server, so chunks will arrive in the client's stream in whatever order happens to be the fastest. Splitting the streams guarantee that static chunks will be given full priority.

> - At the grpc layer, there is now an overlap between the `index=None` semantics of the query, and the newly introduced flags. Does it even make sense to keep that? Or should we drop the former for the latter. (Obviously depends on the resolution of the first point.)

I think it makes sense to remove this once this is in yep.

> Did this improve the experience of loading data from the platform?

We shall know in... 10 minutes -ish.

abey79:
> I don't think there's any extra cognitive cost here

I obviously disagree with that. Any piece of API. Writing documentation, reading documentation, backward compatibility, tests, etc. Is it huge, no, certainly not in the absolute sense. But relative to the alternative of always sorting by "chunk_is_static"? I think so. (Of course unless there are other benefits, which I'm currently not seeing.)

>  you could literally remove these fields from an hypothetical "public version" of the proto definition, and users wouldn't be able to tell the difference

That's my point. We don't. And if we did, the load would fall on the devs.

crates/store/re_grpc_client/src/redap/mod.rs
```diff
@@ -416,18 +416,39 @@ async fn stream_partition_from_server(
...
-            dataset_id: Some(dataset_id.into()),
-            partition_ids: vec![partition_id.into()],
-            chunk_ids: vec![],
-            entity_paths: vec![],
-            select_all_entity_paths: true,
-            query: None,
-        })
-        .await?
-        .into_inner();
+    let static_chunk_stream = {
```
emilk, line 419: It would be great if the code explained _why_ it fetches static data seperatedly

PR #10409, opened 2025-06-27
Preserve pathname prefix in re uri

### LINKED ISSUE #10373: Preserve pathname prefix in `re_uri`
We currently fail to parse URLs like `rerun+http://13.31.13.31/rerun/proxy`. We probably should be more lenient and allow this, because we can always tack more pathname segments at the end.

After parsing, we should keep the original pathname around minus the endpoint, e.g. the above will become `http://13.31.13.31/rerun`. An example gRPC client request to that URL is `POST http://13.31.13.31/rerun/rerun.v1alpha1.sdk_comms.MessageProxyService/ReadMessages`. As long as `/rerun/*` is proxied to the gRPC server with the path rewritten to `/rerun.v1alpha1.sdk_comms.MessageProxyService/ReadMessages`, that should be enough.

stefanwagnerdev:
### Related

* Closes #10373 

### What

This PR implements support for parsing proxy endpoints with arbitrary pathname prefixes in the RedapUri parser, addressing the issue where URLs like rerun+http://host/prefix/proxy would fail to parse

### DIFF
--- a/crates/utils/re_uri/src/endpoints/proxy.rs
+++ b/crates/utils/re_uri/src/endpoints/proxy.rs
@@ -34,6 +34,21 @@ impl ProxyUri {
             prefix_segments: None,
         }
     }
+
+    /// Returns the HTTP URL for connecting to this proxy endpoint.
+    /// This includes the path prefix if present.
+    pub fn endpoint_url(&self) -> String {
+        match &self.prefix_segments {
+            None => self.origin.as_url(),
+            Some(segments) => {
+                if segments.is_empty() {
+                    self.origin.as_url()
+                } else {
+                    format!("{}/{}", self.origin.as_url(), segments.join("/"))
+                }
+            }
+        }
+    }
 }
 
 impl std::str::FromStr for ProxyUri {
--- a/crates/utils/re_uri/src/endpoints/proxy.rs
+++ b/crates/utils/re_uri/src/endpoints/proxy.rs
@@ -3,15 +3,26 @@ use crate::{Origin, RedapUri};
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct ProxyUri {
     pub origin: Origin,
-    pub path_prefix: String,
+
+    /// Path segments before the `/proxy` endpoint.
+    ///
+    /// - `None` for URLs like `rerun://host/proxy` (no prefix)
+    /// - `Some(vec!["prefix"])` for URLs like `rerun://host/prefix/proxy`
+    /// - `Some(vec!["a", "b"])` for URLs like `rerun://host/a/b/proxy`
+    pub prefix_segments: Option<Vec<String>>,
 }
 
 impl std::fmt::Display for ProxyUri {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        if self.path_prefix.is_empty() {
-            write!(f, "{}/proxy", self.origin)
-        } else {
-            write!(f, "{}{}/proxy", self.origin, self.path_prefix)
+        match &self.prefix_segments {
+            None => write!(f, "{}/proxy", self.origin),
+            Some(segments) => {
+                if segments.is_empty() {
+                    write!(f, "{}/proxy", self.origin)
+                } else {
+                    write!(f, "{}/{}/proxy", self.origin, segments.join("/"))
+                }
+            }
         }
     }
 }
@@ -20,7 +31,7 @@ impl ProxyUri {
     pub fn new(origin: Origin) -> Self {
         Self {
             origin,
-            path_prefix: String::new(),
+            prefix_segments: None,
         }
     }
 }
--- a/crates/utils/re_uri/src/endpoints/proxy.rs
+++ b/crates/utils/re_uri/src/endpoints/proxy.rs
@@ -3,17 +3,25 @@ use crate::{Origin, RedapUri};
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct ProxyUri {
     pub origin: Origin,
+    pub path_prefix: String,
 }
 
 impl std::fmt::Display for ProxyUri {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}/proxy", self.origin)
+        if self.path_prefix.is_empty() {
+            write!(f, "{}/proxy", self.origin)
+        } else {
+            write!(f, "{}{}/proxy", self.origin, self.path_prefix)
+        }
     }
 }
 
 impl ProxyUri {
     pub fn new(origin: Origin) -> Self {
-        Self { origin }
+        Self {
+            origin,
+            path_prefix: String::new(),
+        }
     }
 }
 
--- a/crates/utils/re_uri/src/redap_uri.rs
+++ b/crates/utils/re_uri/src/redap_uri.rs
@@ -83,7 +83,7 @@ impl std::str::FromStr for RedapUri {
         let segments = http_url
             .path_segments()
             .ok_or_else(|| Error::UnexpectedBaseUrl(value.to_owned()))?
-            .take(10) // limit to 10 segments
+            .take(10) // increased to handle arbitrary proxy prefixes
             .filter(|s| !s.is_empty()) // handle trailing slashes
             .collect::<Vec<_>>();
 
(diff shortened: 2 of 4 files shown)


crates/utils/re_uri/src/redap_uri.rs
```diff
@@ -83,13 +83,25 @@ impl std::str::FromStr for RedapUri {
         let segments = http_url
             .path_segments()
             .ok_or_else(|| Error::UnexpectedBaseUrl(value.to_owned()))?
-            .take(2)
+            .take(10) // increased to handle arbitrary proxy prefixes
```
jprochazk: This won't handle arbitrary many prefix path segments. 10 is quite a few path segments, and definitely better than 2, but we shouldn't hardcode this.

The comment also doesn't make sense by itself, increased from what?

crates/utils/re_uri/src/redap_uri.rs
```diff
@@ -83,13 +83,25 @@ impl std::str::FromStr for RedapUri {
...
             .filter(|s| !s.is_empty()) // handle trailing slashes
             .collect::<Vec<_>>();
 
         match segments.as_slice() {
             ["proxy"] => Ok(Self::Proxy(ProxyUri::new(origin))),
 
+            // Handle proxy endpoint with pathname prefix - fix for issue #10373
+            segments if segments.last() == Some(&"proxy") => {
```
jprochazk: This general approach is sound, but it should apply to all endpoints, not just `/proxy`. One possible solution is to check the last N segments where N is the length of the endpoint, including any possible dynamic path segments.

For example, for `/proxy` we just need to check the last path segment. But for `/dataset/{dataset_id}`, we need to check the last 2.

crates/utils/re_uri/src/endpoints/proxy.rs
```diff
@@ -3,17 +3,25 @@ use crate::{Origin, RedapUri};
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct ProxyUri {
     pub origin: Origin,
+    pub path_prefix: String,
```
jprochazk: This should probably be a `Option<Vec<String>>`. That does two things:
- Makes it clear that this is _optional_, and
- Ensures that no segment contains a `/`, as that could break the `Display` impl.

Also needs some docs, e.g. `stores all URL path segments before the first endpoint path segment`, or something along those lines. `path_prefix` is a bit too vague

crates/utils/re_uri/src/redap_uri.rs
```diff
@@ -98,7 +110,7 @@ impl std::str::FromStr for RedapUri {
                 Ok(Self::Entry(EntryUri::new(origin, entry_id)))
             }
 
-            ["dataset", dataset_id] => {
+            ["dataset", dataset_id, "data"] => {
```
jprochazk: This is an oversight on our side, and is actually correct according to the comment in `dataset.rs`. The problem is that we may be using `dataset/{dataset_id}` without the `/data` suffix somewhere, including outside of this repo. Changing this may break other code, so please revert. We can fix this separately some other time.

crates/utils/re_uri/src/redap_uri.rs
```diff
@@ -467,6 +480,36 @@ mod tests {
         let address: Result<RedapUri, _> = url.parse();
 
         assert_eq!(address.unwrap(), expected);
+
+        // Test the specific case mentioned in issue #10373
+        let url = "rerun+http://13.31.13.31/rerun/proxy";
```
jprochazk, line 539: We'll need to test other endpoints as well

crates/utils/re_uri/src/redap_uri.rs
```diff
@@ -83,13 +83,25 @@ impl std::str::FromStr for RedapUri {
...
             .ok_or_else(|| Error::UnexpectedBaseUrl(value.to_owned()))?
-            .take(2)
+            .take(10) // increased to handle arbitrary proxy prefixes
             .filter(|s| !s.is_empty()) // handle trailing slashes
             .collect::<Vec<_>>();
 
         match segments.as_slice() {
             ["proxy"] => Ok(Self::Proxy(ProxyUri::new(origin))),
 
+            // Handle proxy endpoint with pathname prefix - fix for issue #10373
```
jprochazk: We typically don't mention issue numbers in regular code paths like this, unless it's a `TODO`. It can and should be mentioned in a regression test.

jprochazk (changes requested):
Hi, thanks for trying to address this! Overall, this needs some more work.

We've got many places in the code that are currently using `re_uri::Origin` directly instead of a full `ProxyUri`. That means they wouldn't handle path prefixes correctly. We should instead change `Origin` to handle path prefixes.

We also need to make sure that we reserve all paths which _don't_ end in one of our endpoints, that way we can continue to add them in the future. For example, `rerun+http://127.0.0.1:9876/rerun/something` should be rejected completely, as neither `rerun/something` nor `something` is a valid endpoint.

Lastly, we'll need some updated docs. We need to be clear about:
- Endpoint paths may contain an arbitrary amount of prefix path segments
- We reserve the right to add other endpoints in the future, and will reject anything that isn't currently a valid endpoint.

This makes me think we should also preserve query params which we aren't using... 🤔 But you don't need to worry about that, it can be addressed another time.

Again, thanks for taking the time to look at this!

crates/utils/re_uri/src/redap_uri.rs
stefanwagnerdev: I removed the take completly

crates/utils/re_uri/src/redap_uri.rs
stefanwagnerdev: It now handles the other endpoints as well

crates/utils/re_uri/src/endpoints/proxy.rs
stefanwagnerdev: added the type and some doc string

crates/utils/re_uri/src/redap_uri.rs
stefanwagnerdev: I added more tests for the other endpoints

crates/utils/re_uri/src/redap_uri.rs
stefanwagnerdev: removed

crates/utils/re_uri/src/redap_uri.rs
stefanwagnerdev: I'm not sure, the tests fail otherwise?

jprochazk: Right, because you're now reading segments from the end. This is resolved, then

jprochazk:
This still won't work as-is. If I add a log to tell me the actual HTTP URL which the gRPC client will attempt to connect to, and I run `pixi run rerun --connect rerun+http://localhost/test/proxy`, I get:

```
[2025-06-30T13:58:27Z DEBUG re_grpc_client::message_proxy::read] Loading rerun+http://localhost:9876/test/proxy via gRPC…
[2025-06-30T13:58:27Z INFO  re_grpc_client::message_proxy::read] actual URL: http://localhost:9876
```

Which is missing the path prefix. This needs to go through `re_uri::Origin`, not the endpoint struct. Here's a test that should pass if path segments are properly supported:

```rust
#[test]
fn resolved_endpoints_with_prefix() {
    let cases = [
        // HTTP
        (
            "rerun+http://localhost/a/b/proxy",
            "http://localhost:9876/a/b",
        ),
        (
            "rerun+http://localhost/a/b/catalog",
            "http://localhost:51234/a/b",
        ),
        (
            "rerun+http://localhost/a/b/entry/1830B33B45B963E7774455beb91701ae",
            "http://localhost:51234/a/b",
        ),
        (
            "rerun+http://localhost/a/b/dataset/1830B33B45B963E7774455beb91701ae/data?partition_id=pid",
            "http://localhost:51234/a/b",
        ),
        // HTTPS
        (
            "rerun://example.com/foo/bar/proxy",
            "https://example.com:443/foo/bar",
        ),
        (
            "rerun://example.com/foo/bar/catalog",
            "https://example.com:443/foo/bar",
        ),
        (
            "rerun://example.com/foo/bar/entry/1830B33B45B963E7774455beb91701ae",
            "https://example.com:443/foo/bar",
        ),
        (
            "rerun://example.com/foo/bar/dataset/1830B33B45B963E7774455beb91701ae/data?partition_id=pid",
            "https://example.com:443/foo/bar",
        ),
    ];

    for (url, expected) in cases {
        let result: RedapUri = url.parse().expect("failed to parse proxy URL");
        assert_eq!(
            expected,
            result.origin().as_url(),
            "failed to resolve {url}"
        );
    }
}
```

jprochazk:
It's also important that you actually test this for your own use case, a locally-running `rerun --serve-web` behind an nginx proxy which you attempt to connect to in a web browser.

Wumpf:
looks abandoned. Closing for book-keeping purposes. Please re-open if there's new movement here

Issue #10366, opened 2025-06-25
Publish JS package documentation


jprochazk:
We publish docs for our SDKs:
* C++ via Doxygen
* Python via mkdocs
* Rust via rustdoc (though we let https://docs.rs/rerun build that for us now, as we abandoned hosting our own)

We should publish docs for our JS packages as well. We haven't done so until now because they've been very simple, but they are gradually becoming more complex. Most recently we've added callbacks which aren't exactly self-explanatory. Also, eventually we may want a full JS SDK.

I've previously used [TypeDoc](https://typedoc.org/) for this, and it will probably work quite well for us as well.

As part of this issue, we should also make sure that the published results are good by doing a full pass over all of our doc comments in JS code.

PR #10410, opened 2025-06-27
Publish JS docs

jprochazk:
* Closes https://github.com/rerun-io/rerun/issues/10366


jprochazk:
@rerun-bot full-check

jprochazk:
@rerun-bot full-check

jprochazk:
@rerun-bot full-check

rerun_js/web-viewer/README.md
```diff
@@ -16,16 +16,19 @@ This package is framework-agnostic. A React wrapper is available at <https://www
 
 ## Install
 
-```
-$ npm i @rerun-io/web-viewer
+```txt
```
Wumpf: nit:
```suggestion
```sh
```

rerun_js/scripts/version.mjs
```diff
@@ -56,8 +58,18 @@ for (const pkg_path of root_package_json.workspaces) {
       /<https:\/\/app\.rerun\.io\/.*\/examples\/dna\.rrd>/,
       `<https://app.rerun.io/version/${version}/examples/dna.rrd>`,
     );
+    if (index_ts) {
```
Wumpf, line 65: don't understand what this is all about. We want to load the ts file because we need it for generating docs but it has a dna.rrd path in there that we have to reformat?

.github/workflows/reusable_deploy_docs.yml
```diff
@@ -92,14 +97,29 @@ jobs:
...
+              window.location.replace("../${{ inputs.PY_DOCS_VERSION_NAME }}/" + window.location.hash);
+            </script>
+          </head>
+          <body>
+            Redirecting to <a href="../${{ inputs.PY_DOCS_VERSION_NAME }}/">../${{ inputs.PY_DOCS_VERSION_NAME }}/</a>
+          </body>
+          </html>
+          EOF
+
+          gsutil cp ./index.html gs://rerun-docs/docs/python/stable/
```
Wumpf, line 122: this python doc stuff seems unrelated to js docs. Do I read this correctly that you're fixing the latest redirect here instead of reuploading the python docs yet-another-time?

jprochazk: Yes, this is how it used to work when we used `mike`. This is a drive-by fix

rerun_js/scripts/version.mjs
jprochazk: I want the versioned link to the DNA example in the `index.ts` file to be updated automatically when our release workflow versions JS stuff

jprochazk:
> Can I have a preview link?

https://ref.rerun.io/docs/js

To be honest, I don't really like the theme at all, it looks pretty bad... but every other theme has some issue with it, like links not being highlighted properly, syntax highlighting not working, or it being incompatible with the latest version of typedoc, among other things.
The default theme is the only one that actually works properly, and I don't think it makes sense to invest time into forking one of the themes to make this look better right now. At least it's functional

rerun_js/web-viewer/README.md
jprochazk: (also applied to web-viewer-react)

jprochazk:
(link checker failure is unrelated :tada:)

Wumpf:
> To be honest, I don't really like the theme at all, it looks pretty bad

I actually like it :)

jprochazk:
> I actually like it :)

Even the rounded corners on backtick text which clip into the text? :smile:

Wumpf:
😱

PR #10412, opened 2025-06-27
Better thumbnail display in table

lucasmerlin:
### Related
- closes https://github.com/rerun-io/dataplatform/issues/1013

### What

Adds a new thumbnail variant ui that doesn't display the image metadata and takes the full available height. 
If the first column is a blob, the row height is increased so the thumbnail has more space.

Also fixes the `Open` link not being centered (that became really annoying now with the increased row height)

https://github.com/user-attachments/assets/c488e84f-bd82-46cf-905f-2b490084899e

For testing, here's a script that creates the table seen in the video (thanks @abey79):
<details><summary>Details</summary>
<p>

```py

import timeit
import uuid
from pathlib import Path

import numpy as np
import tqdm
import pyarrow as pa
import rerun_sdk.rerun as rr
from rerun_sdk.rerun.catalog import CatalogClient


image_file_path = Path(__file__).parent / "ferris.png"


def write_rrd(path: Path) -> str:
    partition_id = str(uuid.uuid4())
    with rr.RecordingStream("rerun_example_sandbox", recording_id=partition_id) as rec:
        rec.save(path)

        rec.log("/test", rr.Points2D.columns(positions=np.random.rand(100, 2), radii=5), static=True)

        timestamp = rr.TimeColumn("timeline", sequence=list(range(10000)))
        rec.send_columns("/test", [timestamp], rr.Points3D.columns(positions=np.random.rand(10000, 3)))

        timestamp = rr.TimeColumn("timeline", sequence=list(range(1000)))
        rec.send_columns("/test2", [timestamp], rr.Points3D.columns(positions=np.random.rand(1000, 3)))

        rec.log("/not_relevant", rr.AnyValues(yak="yuk", foo="bar", baz=42), static=True)
        rec.log("/not_relevant", rr.Points2D.columns(positions=np.random.rand(100, 2), radii=5), static=True)

        timestamp = rr.TimeColumn("timeline", sequence=list(range(10000)))
        rec.send_columns("/not_relevant", [timestamp], rr.Points3D.columns(positions=np.random.rand(10000, 3)))


        rec.send_property(
            "thumbnail",
            rr.EncodedImage(path=image_file_path),
        )

        rec.flush()

    return partition_id


uris = []
partition_ids = []
for i in range(10):
    path = Path(f"/tmp/sandbox{i}.rrd")
    partition_ids.append(write_rrd(path))
    uris.append(f"file://{path.as_posix()}")


c = CatalogClient("rerun+http://127.0.0.1:51234")

try:
    ds = c.get_dataset(name="sandbox")
    ds.delete()
except Exception:
    pass

ds = c.create_dataset(name="sandbox")

print(ds.manifest_url)

for uri in tqdm.tqdm(uris):
    ds.register(uri)

view = ds.dataframe_query_view(index="timeline", contents="/**")
# view = ds.dataframe_query_view(index=None, contents={"/test": ["baz"]})
view = view.filter_partition_id(*partition_ids[:3])

print(pa.Table.from_batches(view.df().collect()))

``` 

</p>
</details>


crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -327,6 +328,20 @@ impl<'a> DataFusionTableWidget<'a> {
 
         let mut new_blueprint = table_state.blueprint().clone();
 
+        let mut row_height = viewer_ctx.tokens().table_line_height();
+
+        // If the first column is a blob, we treat it as a thumbnail and increase the row height.
```
oxkitsune, line 333: should this not be a bit more specific? Especially now that we have `VideoStream`. I'm definitely missing context here, but it'd be bad to try and draw a single h264 blob as image right?

crates/viewer/re_data_ui/src/image.rs
```diff
@@ -66,7 +66,11 @@ pub fn texture_preview_ui(
     texture: ColormappedTexture,
 ) -> egui::Response {
     if ui_layout.is_single_line() {
-        let preview_size = Vec2::splat(ui.available_height());
+        let height = ui.available_height();
+        let [texture_width, texture_height] = texture.width_height();
+        let width =
+            (height * texture_width as f32 / texture_height as f32).at_most(ui.available_width());
+        let preview_size = Vec2::new(width, height);
```
oxkitsune, line 73: This will break the aspect ratio if the `available_width()` is less than the required width. 
I think we should aim to maintain aspect ratio by computing a scale factor for both width and height and picking the minimum value so we don't overflow either dimension, e.g.
```suggestion
        let [texture_width, texture_height] = texture.width_height();

        let available_height = ui.available_height();
        let available_width = ui.available_width();

        let scale_height = available_height / texture_height as f32;
        let scale_width = available_width / texture_width as f32;

        let scale = scale_height.min(scale_width);

        let preview_size = Vec2::new(texture_width as f32 * scale, texture_height as f32 * scale);
```

oxkitsune (changes requested):

crates/viewer/re_data_ui/src/image.rs
lucasmerlin: egui will handle the aspect ratio, so this should be fine

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
lucasmerlin: Ideally this would be handled via the (not yet existing) table blueprint api. Since we don't have that yet, this is a band-aid fix to make the thumbnails in the sandbox more appealing. I'll add a check if this is actually an image though 👍🏻

crates/viewer/re_dataframe_ui/src/display_record_batch.rs
```diff
@@ -232,10 +251,12 @@ impl DisplayComponentColumn {
                     row_id = Some(RowId::from_u128(
                         quick_partial_hash(buffer, SECTION_LENGTH).hash64() as _,
                     ));
+
+                    variant_name = Some(VariantName::from(REDAP_THUMBNAIL_VARIANT));
```
abey79, line 245: Maybe this entire block should be conditioned on the new new `is_image()`. Also, setting the `variant_name` should not be conditioned on the `row_id` being initially `None`.

abey79 (approved):
Nice! See my comment for a small change which would be nice to have.

PR #10413, opened 2025-06-27
Trigger C++ ci on rerun_c changes, trigger Rust ci on `re_ui` data changes

Wumpf:
some missing triggers that tripped folks recently


grtlr (approved):
One small step... LGTM!

PR #10416, opened 2025-06-27
Unify component information in selection panel and header popups

grtlr:
### Related

* Follow up of #10245

### What

Discussion: https://rerunio.slack.com/archives/C033K5VS2KD/p1751035525843599

This unifies the Datafusion widget column header popup and the selection panel to use the new Component descriptor nomenclature.

### Screenshots

| Column header popup | Selection panel |
|--------|--------|
| image | image |

### DIFF
--- a/crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
+++ b/crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
@@ -633,12 +633,7 @@ fn column_descriptor_ui(ui: &mut egui::Ui, column: &ColumnDescriptorRef<'_>) {
                 is_semantically_empty,
             } = desc;
 
-            header_property_ui(ui, "Type", "component");
-            header_property_ui(
-                ui,
-                "Component type",
-                component_type.map(|a| a.as_str()).unwrap_or("-"),
-            );
+            header_property_ui(ui, "Column type", "Component");
             header_property_ui(ui, "Entity path", entity_path.to_string());
             datatype_ui(ui, &column.display_name(), store_datatype);
             header_property_ui(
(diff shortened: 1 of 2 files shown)


crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -633,23 +633,19 @@ fn column_descriptor_ui(ui: &mut egui::Ui, column: &ColumnDescriptorRef<'_>) {
...
-            );
+            header_property_ui(ui, "Column type", "Component");
             header_property_ui(ui, "Entity path", entity_path.to_string());
             datatype_ui(ui, &column.display_name(), store_datatype);
             header_property_ui(
                 ui,
                 "Archetype",
                 archetype.map(|a| a.full_name()).unwrap_or("-"),
             );
+            header_property_ui(ui, "Component", desc.component_descriptor().component);
```
emilk:
```suggestion
            header_property_ui(ui, "Component", component);
```

and then remove the debug-only printing of exactly the same thing (see your own screenshot)

crates/viewer/re_selection_panel/src/selection_panel.rs
```diff
@@ -211,8 +211,8 @@ impl SelectionPanel {
                 }
 
                 ui.list_item_flat_noninteractive(
-                    PropertyContent::new("Archetype field")
-                        .value_text(component_descriptor.archetype_field_name()),
+                    PropertyContent::new("Component")
+                        .value_text(component_descriptor.component.as_str()),
```
emilk: We already does a desctruction of `ComponentDescriptor` and intentionally ignore `component` in it, and then extract it explicitly here anyways… do one or the other

emilk (changes requested):

Issue #10295, opened 2025-06-18
Video asset/stream cache is discarded too often


Wumpf:
Various user interactions cause the video cache to be falsely discarded, causing wait time (load spinner shows up)

ntjohnson1:
Looks like you were already aware. I recorded a video to create an issue for this. Sharing video in case it is helpful. It seems somewhat non-deterministic to particular actions/times.

https://github.com/user-attachments/assets/d5dbd957-2a00-4dde-9d0c-3b0a95b82a56

Wumpf:
the issue your video shows is actually something else, that's
* https://github.com/rerun-io/rerun/issues/10294

PR #10420, opened 2025-06-27
Fix videos re-decoding on some ui interactions

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/10295

### What
This properly evicts videos only if they weren't in used in _any pass_ during the previous _frame_.

In egui, each frame is _usually_ one pass, but can be two (if something calls `egui::Context::request_discard`). In that case, the video may be invisible one pass (because we early-out), but visible the next. Before that lead to an unecessary eviction. The new code solves this.

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -1500,7 +1500,7 @@ impl App {
                         #[cfg(not(target_arch = "wasm32"))]
                         let is_history_enabled = false;
 
-                        render_ctx.begin_frame();
+                        render_ctx.begin_frame(); // This may actually be called multiple times per egui frame, if we have a multi-pass layout frame.
 
                         // In some (rare) circumstances we run two egui passes in a single frame.
                         // This happens on call to `egui::Context::request_discard`.


crates/viewer/re_viewer/src/app.rs
```diff
@@ -1507,6 +1507,13 @@ impl App {
                         );
 
                         render_ctx.begin_frame();
+
+                        // Call begin_frame on caches only after starting the re_renderer frame
+                        // since we rely on the renderer frame index for some tracking of outdated resources.
+                        if let Some(store_hub) = &mut self.store_hub {
+                            store_hub.begin_frame_caches(render_ctx.active_frame_idx());
```
emilk: I think we should only call this on the first pass of each frame, in case there are several. Otherwise the "used this frame" is wrong, since it really means "used this pass".

Reminder: an egui frame usually consist of a single pass, but sometimes two.

crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -161,6 +161,7 @@ pub fn blob_preview_and_save_ui(
                             blob,
                             media_type,
                             ctx.app_options().video_decoder_settings(),
+                            ctx.render_ctx().active_frame_idx(),
```
emilk: We should check and document wether or not this increments each _frame_ or each _pass_

Wumpf: this increments each time we call `begin_frame` on `re_renderer`. Which I think we do on all egui passes...? A bit concerning as that's quite wasteful :/

crates/viewer/re_viewer/src/app.rs
Wumpf: yeah those "this frame" things tripped which is why I removed those used-this-frame booleans.
We can do the is_sizing_pass checks here to not call this in the first place, but it immediately makes me wonder what else is affected by this.
For instance re_renderer also ignores everything in sizing commands 😬

Wumpf: I'll try only-first-pass-each-frame on Monday and test it a bit. I should also have a look at other caches like mesh cache and image decode cache and ensure they still work as expected!

is the right way to check for sizing pass on the `ui` object or is there some more elaborate thing I need to do?

crates/viewer/re_data_ui/src/blob.rs
emilk: We only run a second pass if `request_discard` is called, which is very rare. It currently only happens when a new ListItem scope appears for the first time, and on some interactions with `egui_table` (e.g. when auto-sizing it).

emilk: We could add a little marker (e.g. an orange dot) to the top bar metrics (in debug builds) whenever we run a second pass. That could be helpful for us developers.

crates/viewer/re_viewer/src/app.rs
```diff
@@ -1500,13 +1500,20 @@ impl App {
...
-                            &self.async_runtime,
-                            &self.egui_ctx,
-                        );
+                        render_ctx.begin_frame(); // This may actually be called multiple times per egui frame, if we have a multi-pass layout frame.
+
+                        // In some (rare) circumstances we run two egui passes in a single frame.
+                        // This happens on call to `egui::Context::request_discard`.
+                        let is_start_of_new_frame = egui_ctx.current_pass_index() == 0;
+
+                        if is_start_of_new_frame {
```
emilk, line 1509: Unrelated, but 🤷

Wumpf (commented):
👍 
I do prefer frame counter tracking over boolean since it's nicer for debugging, but let's not do this until we get a better grip on re_renderer's frame counter 👍

emilk:
We can pass in the egui frame count instead if we want 🤷

Issue #9457, opened 2025-04-02
Make `PySchema` a (rich) wrapper over `ArrowSchema`/`SorbetSchema`


abey79:
The Python object `PySchema` was introduced with the datafarme API and currently contains only a list of column descriptor. In the context of the dataplatform's dataset and Sorbet, we must extend `PySchema` to:

- better integrate with the arrow ecosystem (so it should implement the arrow schema pycapsule)
- contain a full schema, not just a list of field (aka also include the schema metadata)
- be used both for the dataframe API and the catalog/dataset API
- eventually, expose all sorts of method to display/navigate the schema in a "sorbet-friendly" way

abey79:
closing as no longer relevant

PR #10428, opened 2025-06-27
Add formatted str representation for PySchema

ntjohnson1:
### Related
Related to #9457 but dramatically less ambitious. Right now we have no repr or str so it doesn't print out anything useful.

### What
This adds a basic __str__ representation. Since this is often getting used in notebooks I don't know if we have a hardline policy on repr vs str. Repr should allow us to recreate the object with the information, but that is what gets displayed if the variable name is typed in a standalone cell in Jupyter. Str is allowed to be more pretty print. Since we didn't have anything before I did a pretty print and set the repr to that.

Example output
```python
>>> dataset.schema()
Column name: /action/gripper_position:Scalars:scalars
                  Entity path: /action/gripper_position
                  Archetype type: rerun.archetypes.Scalars
                  Component type: rerun.components.Scalar
<repeat for columns, not indicator or link>
```

### DIFF
--- a/rerun_py/src/dataframe.rs
+++ b/rerun_py/src/dataframe.rs
@@ -165,6 +165,21 @@ impl PyComponentColumnDescriptor {
         self.0.column_name(re_sorbet::BatchType::Dataframe)
     }
 
+    fn __str__(&self) -> String {
+        format!(
+            "Column name: {}",
+            self.0.column_name(re_sorbet::BatchType::Dataframe)
+        )
+        .to_owned()
+            + format!("\n\tEntity path: {}", self.entity_path()).as_str()
+            + format!("\n\tArchetype type: {}", self.archetype().unwrap_or("None")).as_str()
+            + format!(
+                "\n\tComponent type: {}",
+                self.component_type().unwrap_or("")
+            )
+            .as_str()
+    }
+
     fn __eq__(&self, other: &Self) -> bool {
         self.0 == other.0
     }
@@ -469,6 +484,23 @@ pub struct PySchema {
 /// [`RecordingView.schema()`][rerun.dataframe.RecordingView.schema].
 #[pymethods]
 impl PySchema {
+    fn __str__(&self) -> String {
+        self.component_columns()
+            .iter()
+            .filter(|field| {
+                !field.entity_path().contains("link") && !field.component().contains("Indicator")
+            })
+            .map(|col| col.__str__())
+            .collect::<Vec<_>>()
+            .join("\n")
+    }
+
+    // Technically this violates python's `__repr__` definition
+    // but since we previously didn't have anything that's ok for now.
+    fn __repr__(&self) -> String {
+        self.__str__()
+    }
+
     /// Iterate over all the column descriptors in the schema, ignoring `RowId`.
     fn __iter__(slf: PyRef<'_, Self>) -> PyResult<Py<SchemaIterator>> {
         let py = slf.py();


ntjohnson1:
cc: @pweids

rerun_py/src/dataframe.rs
```diff
@@ -469,6 +489,23 @@ pub struct PySchema {
...
+            .iter()
+            .filter(|field| {
+                !field.entity_path().contains("link") && !field.component().contains("Indicator")
+            })
+            .map(|col| col.__str__())
+            .collect::<Vec<_>>()
+            .join("\n")
+    }
+
+    // Technically this violates python's `__repr__` definition
```
oxkitsune: I think it would be slightly more valid if we implement the contents of `__str__` in `__repr__` and then call `self.__repr__()` from `__str__`

rerun_py/src/dataframe.rs
```diff
@@ -469,6 +489,23 @@ pub struct PySchema {
 /// [`RecordingView.schema()`][rerun.dataframe.RecordingView.schema].
 #[pymethods]
 impl PySchema {
+    fn __str__(&self) -> String {
+        self.component_columns()
+            .iter()
+            .filter(|field| {
+                !field.entity_path().contains("link") && !field.component().contains("Indicator")
+            })
```
oxkitsune:
```suggestion
            .filter(|column| !column.is_indicator)
```
There's a neat utility for checking whether a column contains indicator components, also makes @grtlr's indicator purge easier.

I don't think we should filter out entity paths that contain the word `link`, even if the output contains a full URDF tree with lots of `**/link*/**` entities, these are valid columns that can contain valid data.

rerun_py/src/dataframe.rs
```diff
@@ -165,6 +165,26 @@ impl PyComponentColumnDescriptor {
...
+        );
+        pretty_str.push_str(
+            format!(
+                "\n\tComponent type: {}",
+                self.component_type().unwrap_or("")
+            )
+            .as_str(),
+        );
+        pretty_str.push_str(format!("\n\tComponent: {}", self.component()).as_str());
+        pretty_str
```
oxkitsune:
```suggestion
        format!(
            "Column name: {col}\n\
             \tEntity path: {path}\n\
             \tArchetype: {arch}\n\
             \tComponent type: {ctype}\n\
             \tComponent: {comp}",
            col   = self.0.column_name(re_sorbet::BatchType::Dataframe),
            path  = self.entity_path(),
            arch  = self.archetype().unwrap_or("None"),
            ctype = self.component_type().unwrap_or(""),
            comp  = self.component(),
        )
```
We can use a single `format!` invocation here

rerun_py/src/dataframe.rs
```diff
@@ -469,6 +489,23 @@ pub struct PySchema {
...
 impl PySchema {
+    fn __str__(&self) -> String {
+        self.component_columns()
+            .iter()
+            .filter(|field| {
+                !field.entity_path().contains("link") && !field.component().contains("Indicator")
+            })
+            .map(|col| col.__str__())
+            .collect::<Vec<_>>()
+            .join("\n")
```
oxkitsune:
```suggestion
            .join("\n")
```

We can use `itertools` to avoid the temporary allocation here.

oxkitsune (changes requested):

rerun_py/src/dataframe.rs
ntjohnson1: I'll look for the column indicator components utility.

As for filtering the `link` items I think that's why I added the note in code about the `__str__` being fine but the filtering basically making `__repr__` non-compliant. I was motivated by Paul's example where he did the filtering and pretty printing. Right now the arrow schema uses the repr that prints everything and there are so many links its a little more annoying to use.

I'm not up to speed enough yet to know if we have a nice pattern for our python bindings across the repo but one thought would be to add a `pretty_print` method with some kind of filter.
```
fn __str__(&self) -> String {
   self.pretty_print(<filter link>)
}
```

I guess alternatively with updating the repr for the columns I've at least simplified Paul's filtering code to which might be nicer than adding a filter internally.

```
for field in pyschema_instance.component_columns():
    if "link" not in field.entity_path and "Indicator" not in field.component:
       print(field)
```

oxkitsune: I like the `pretty_print` idea. However, I don't think we should filter anything out (besides indicators) in both `__str__` and `__repr__`.

The problem with filtering out `"link"` at this level, this will 100% lead to data being filtered out, which you don't want being filtered out.

I think we should definitely hide them in Paul's example, but not in the library code.

For Paul's filtering code, also don't think we should be checking for the string "Indicator" in the component name, we should maybe add a temporary getter for the `is_indicator` property. But there's #10315 and Indicators will be dropped (hopefully) soon anyways.

rerun_py/src/dataframe.rs
ntjohnson1: I think python implicitly uses repr is str isn't defined. Removing the str and just implementing repr will continue discussion on other comment.

rerun_py/src/dataframe.rs
ntjohnson1: Ok, then to keep this moderate in scope I plan to remove my explicit `__str__` and just implement `__repr__` with no filtering. I think the `pretty_print` is nice but complicates the interface if we're not doing that elsewhere so maybe punt that to the larger scope issue linked in the description.

As for #10315 I added `is_indicator_component` which basically ends up checking for the `indicator` string but through a more centralized location which will be cleaned up when indicators are removed.

pweids: I agree with @oxkitsune. I filtered out the links in my example because there are tons of them and it's kind of noise for our purposes. But in general, others may need/want to see this column information.

I would like to see something where we can filter the schema itself, like `print([col for col in dataset.schema() if 'link' not in col.name.lower()])`, and these columns pretty-print in the same fashion

ntjohnson1: Hmm is your suggestion different than the updated usage I put in the description? I think I capture that workflow with the updated pretty print for the fields, but maybe I'm missing the nuance.

The schema already has an iterator.
```
>>> for field in dataset.schema():
...if "link" not in field.entity_path and not in field.component.is_indicator():
......print(field)
Column name: /action/gripper_position:Scalars:scalars
                  Entity path: /action/gripper_position
                  Archetype type: rerun.archetypes.Scalars
                  Component type: rerun.components.Scalar
<repeat for columns not link or indicator>
```

pweids: oh I missed the 2nd half of your example above. Nice! Sounds like we are on the same page then. Disregard

PR #10429, opened 2025-06-27
Redap browser: (attempt at) snappier `Open`

teh-cmc:
* Make sure to send `StoreInfo` early
* Make sure to query temporal chunks late

Try it with the auto-generated viewer from this PR, see if you like it :shrug: It not, feel free to close this.


teh-cmc:
Now that we have a spinner, I think I actually prefer the version on `main` where we linger a bit longer on the Redap browser view, and only swap the view in at the last minute.

PR #10411, opened 2025-06-27
Rename `RecordingProperties` to `RecordingInfo`

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/10318
* Does most of https://github.com/rerun-io/rerun/issues/10226
* Follow-up: https://github.com/rerun-io/rerun/pull/10432
* Sibling PR: https://github.com/rerun-io/dataplatform/pull/1104

### What
* Rename `RecordingProperties` to `RecordingInfo`
* Move `/__properties/recording` to `/__properties`
* Add migration code for the above
* Tweak the UI

The user-facing change is minimal: users don't use the code-generated archetype directly, and we hide its path in the UI in most places. Only very advanced users are affected.

### Screenshot
Screenshot 2025-06-29 at 20 05 49

### TODO
* [x] Full check

Issue #10226, opened 2025-06-13
Clean up recording properties UI in selection panel


grtlr:
**Describe the bug**

The display of the recording properties in the selection panel could be improved now that we have the nice per-archetype UI.

**To Reproduce**
1. Go to `__properties` in the streams panel
2. Look at the selection panel

**Expected**

Currently, the _Recording_ label comes from the parsed entity path that we use for the recording properties. We should figure out a way to communicate that fact to the user.

**Screenshots**

<img width="342" alt="Image" src="https://github.com/user-attachments/assets/d9d23a77-9207-47ec-b89a-972560da6523" />

## Proposal to improve: 
- align spacing
- a tiny increase of spacing after each :scope: (which are `recording` `episode` `robot` and `user`)
- Remove "without archetype" for when there is 1 archetype per entity path (in majority case it's all of them)
- Look for design [here](https://www.figma.com/design/04eHlTWW361rIs3YesfTJo/Data-platform?node-id=1438-1976&t=zk48bmHbx4zKwwBR-1)
<img width="471" alt="Image" src="https://github.com/user-attachments/assets/13bf58c2-52f8-4003-b298-16db3b42494c" />

To not miss: 
the rule to display "without archetype" when there is more than 1 archetype per entity oath is especially important for when user clicks on an entity path at a time frame and that entity path has more than 1 archetype. See example: 

<img width="379" alt="Image" src="https://github.com/user-attachments/assets/e23f5bb1-ca54-4ba3-873c-353d412eaf69" />

Compare the edge case for if we do not display it: 

<img width="847" alt="Image" src="https://github.com/user-attachments/assets/f3293d49-635a-4d71-a51a-5bdfb68919b4" />

emilk:
* Closed by https://github.com/rerun-io/rerun/pull/10411

PR #10432, opened 2025-06-29
Add snapshot test for recording properties

emilk:
### Related
* Tests https://github.com/rerun-io/rerun/pull/10411
* Part of https://github.com/rerun-io/rerun/issues/10226

### What
* Add some properties to the default test recording
* Adds a snapshot test of the selection panel when the recording is selected

### Result
selection_panel_recording

### DIFF
--- a/crates/viewer/re_time_panel/tests/snapshots/various_entity_kinds_timeline_a_0.png
+++ b/crates/viewer/re_time_panel/tests/snapshots/various_entity_kinds_timeline_a_0.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:c97a085ec189a9169fe0f14142ad15d00d54163a65eec77ec95f50c65733c48a
-size 128944
+oid sha256:8b7847eb5f0ee47bd90c926a97b97fe28ae0730aaf70d2e96663ba15301a0e9e
+size 129018
--- a/crates/viewer/re_time_panel/tests/snapshots/various_entity_kinds_timeline_a_5.png
+++ b/crates/viewer/re_time_panel/tests/snapshots/various_entity_kinds_timeline_a_5.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:be24b194d139deb958390f58caff01f0276521a8390579e49465018692dcca68
-size 129128
+oid sha256:d762a83620658bccb97b1b49feb0fec1f1ffa90b3dc9417c8dc1c03aa5efbfbe
+size 129202
--- a/crates/viewer/re_time_panel/tests/snapshots/various_entity_kinds_timeline_a_9223372036854775807.png
+++ b/crates/viewer/re_time_panel/tests/snapshots/various_entity_kinds_timeline_a_9223372036854775807.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:ee403ecccb568e5460ed43472718a6414e9e5a895386a89c13ff65f480cb62a0
-size 130030
+oid sha256:2a08a89beae63dc89c2b9e29709b1b3ee83c1ee0c9d7aa93c7e6e35ddc18f473
+size 130049
--- a/crates/viewer/re_time_panel/tests/snapshots/various_entity_kinds_timeline_b_0.png
+++ b/crates/viewer/re_time_panel/tests/snapshots/various_entity_kinds_timeline_b_0.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:44dfe445cf0b06130ae9efebd7df842ecbfd8a2db5c3a8c1cfabf0ac58ede654
-size 126306
+oid sha256:acf58c0818471a347c6444e9c6b7e8e14dc08ae3bdb574b3f575c534a0494a55
+size 126376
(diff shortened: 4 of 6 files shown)


crates/store/re_entity_db/src/entity_db.rs
```diff
@@ -200,11 +204,13 @@ impl EntityDb {
...
     }
 
-    pub fn recording_property<C: re_types_core::Component>(
+    /// Read one of the built-in `RecordingInfo` properties.
+    pub fn recording_info_property<C: re_types_core::Component>(
         &self,
         component_descr: &re_types_core::ComponentDescriptor,
     ) -> Option<C> {
         debug_assert_eq!(component_descr.component_type, Some(C::name()));
+        debug_assert!(component_descr.archetype == Some("rerun.archetypes.RecordingInfo".into()));
```
Wumpf: I though this is wrong at first because this could also be used for custom properties. But the debug_assert message further down cleared this up `"RecordingInfo should be logged at {}. Custom properties should be under a child entity",`. Can we have this here as well? Thanks!

(i.e. I wasn't aware that custom properties aren't allowed at `EntityPath::properties()`)

crates/viewer/re_data_ui/Cargo.toml
```diff
@@ -18,6 +18,12 @@ workspace = true
 [package.metadata.docs.rs]
 all-features = true
 
+
+[features]
+## Hide non-determinsitic parts of the UI for snapshot testing.
```
Wumpf: we have non-deterministic ui? 😱 
worth elaborating (at this point in the review I have now clue!)

Edit: Later-me sees that this makes sense. How about

```suggestion
## Hide non-deterministic parts of the UI for snapshot testing (e.g. displaying time in screenshot tests).
```

Wumpf: Actually, can we please try to do this without feature flag? Doing a runtime check on the context (or some global) whether we're testing would be better a build permutation!

crates/viewer/re_data_ui/src/entity_db.rs
```diff
@@ -84,14 +84,17 @@ impl crate::DataUi for EntityDb {
...
 
-            if let Some(latest_row_id) = self.latest_row_id() {
-                if let Ok(nanos_since_epoch) =
-                    i64::try_from(latest_row_id.nanos_since_epoch())
-                {
-                    let time = re_log_types::Timestamp::from_nanos_since_epoch(nanos_since_epoch);
-                    ui.grid_left_hand_label("Modified");
-                    ui.label(time.format(ctx.app_options().timestamp_format));
-                    ui.end_row();
+            let show_last_modified_time = !cfg!(feature = "testing"); // Hide in tests because it is non-deterministic.
```
Wumpf: as above, can we do this with a runtime check rather than build permutation?

Wumpf: reading further in the review I'm confused by this: the test harness correctly fixes the time with a deterministic value, why is this needed then?

crates/viewer/re_viewer_context/src/test_context.rs
```diff
@@ -171,7 +185,67 @@ impl Default for TestContext {
...
+            // Set some custom recording properties:
+            recording_store
+                .set_recording_property(
+                    EntityPath::properties() / EntityPathPart::from("episode"),
+                    ComponentDescriptor {
+                        archetype: None,
+                        component: "building".into(),
+                        component_type: Some(re_types::components::Text::name()),
+                    },
+                    &re_types::components::Text::from("Warehuse B"),
```
Wumpf:
```suggestion
                    &re_types::components::Text::from("Warehouse B"),
```

https://warehouseb.com/ 🤔  Maybe you wanna pick something else from https://en.wikipedia.org/wiki/Category:Fictional_buildings_and_structures 😄

crates/viewer/re_viewport_blueprint/src/viewport_blueprint.rs
```diff
@@ -83,7 +83,7 @@ pub struct ViewportBlueprint {
 
 impl ViewportBlueprint {
     /// Attempt to load a [`ViewBlueprint`] from the blueprint store.
-    pub fn try_from_db(blueprint_db: &re_entity_db::EntityDb, query: &LatestAtQuery) -> Self {
+    pub fn from_db(blueprint_db: &re_entity_db::EntityDb, query: &LatestAtQuery) -> Self {
```
Wumpf: nice catch, but there's no attempting here, comment needs update as well
(gh doesn't let me do suggestion here for some reason)

Wumpf (approved):
tentatively approved - I'd really like to avoid more usages of `testing` feature flag if possible (build permutations are bad ™️  and iirc `testing` caused us some grief in the past) 

curious, do you know why it added `snapshot_kind: text` everywhere?

crates/viewer/re_data_ui/src/entity_db.rs
emilk: This is based on `RowId` timestamp

crates/viewer/re_data_ui/Cargo.toml
emilk: Good idea

crates/viewer/re_data_ui/src/entity_db.rs
Wumpf: ahhh, makes sense. Too bad

teh-cmc:
> curious, do you know why it added `snapshot_kind: text` everywhere?

Older versions of `insta` used to do that (or is it newer versions? I think it's older :melting_face:)

Issue #2353, opened 2023-06-09
Expose connection status in the Python API


abey79:
The Python API should expose the connection status and/or a disconnect event when streaming and the remote viewer is closed. For live streaming applications (e.g. based on video capture, etc.), this would enable the script to automatically stop streaming when the viewer is closed by the user.

Wumpf:
The underlying `LogSink` traits should have ways to report their status and optionally (?) have hook for waiting status change.

For example something that is very confusing today is that `connect_grpc` for instance is not blocking until a connection has been established and will rather continuously try to establish one. Neither is there an indication when it is done, nor is there a way to block on it nor is there a way to tell when the viewer is disconnected.

fortminors:
@Wumpf @jleibs  @abey79 Are there any plans on implementing this feature?
And are there any temporary workarounds until this is implemented?

Wumpf:
@fortminors we unfortunately don't have any short term plans with it. But since you mentioned on the other issue that you need this from Rust I'm wondering if we could do some "expert-level" api quickly before designing something out that works for all SDK languages: recording streams have different types of sinks (with the upcoming `set_sinks` api we're slowly getting more explicit about them), so if a user can read that out and do a dynamic cast to a `GrpcSink` struct (which is already part of the crate's api iirc!) we could expose a status enum on that 🤔

fortminors:
@Wumpf if you could implement any working option, that would be very useful! Looking forward to it.
The proposal looks promising, but will the status enumerate be updated if the remote rerun is disconnected? Or perhaps there should be some method that retrieves the status on demand?

Wumpf:
@fortminors I tried giving it a "quick" spin, was a bit more involved than I thought it would be.
Would be awesome if you could have a look and check if that's useful for you https://github.com/rerun-io/rerun/pull/10433

fortminors:
@Wumpf Thank you! I have checked it and it works - I am able to intercept the connection loss, however rerun itself stopped working... :)
I continuously get errors in rerun, saying 
```
ERROR re_grpc_server] dropping LogMsg due to failed decode: Failed to read chunk: Bad chunk schema: Missing row_id column
```
It seems that it's a different issue because the rest of the code is left unchanged. I was using rerun `0.23.2` before and now I am trying a fresh one on your branch

fortminors:
[...]

PR #10433, opened 2025-06-29
Experimental connection status inspection for Rust gRPC connections

Wumpf:
### Related

* Part of https://github.com/rerun-io/rerun/issues/2353

### What

Experimental (!!) API for checking the connection status of `connect_grpc`. Comes with a small example snippet.

I was wondering how hard this is to do.. answer is: so-so! What makes this tricky is that today the recording sink is fully owned by the "forwarding thread" which makes sense if you zoom out a bit, but makes it hard to synchronously inspect a sink.
We could make that possible by putting the sink behind an `Arc<Mutex<the_sink>>` but that ofc means that we then have to lock a mutex instead which isn't great either.

Anyways, I think what I put up here is still not terrible and usable for Rust experts. Maybe there's more we should do to mark this as proper experimental until we come up with a proper x-language design? 🤔

### DIFF
--- a/crates/top/re_sdk/src/recording_stream.rs
+++ b/crates/top/re_sdk/src/recording_stream.rs
@@ -982,6 +982,7 @@ impl RecordingStreamInner {
 enum Command {
     RecordMsg(LogMsg),
     SwapSink(Box<dyn LogSink>),
+    InspectSink(Box<dyn FnOnce(&dyn LogSink) + Send + 'static>),
     Flush(Sender<()>),
     PopPendingChunks,
     Shutdown,
@@ -1479,6 +1480,9 @@ fn forwarding_thread(
 
                 *sink = new_sink;
             }
+            Command::InspectSink(f) => {
+                f(sink.as_ref());
+            }
             Command::Flush(oneshot) => {
                 re_log::trace!("Flushing…");
                 // Flush the underlying sink if possible.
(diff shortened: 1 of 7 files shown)


crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -982,6 +982,7 @@ impl RecordingStreamInner {
 enum Command {
     RecordMsg(LogMsg),
     SwapSink(Box<dyn LogSink>),
+    InspectSink(Box<dyn FnOnce(&dyn LogSink) + Send + 'static>),
```
jprochazk: This doesn't seem great... If we had explicit sinks, then the gRPC sink could hold a reference to the status directly, and none of this dance around the forwarding thread would be necessary.

I wrote up an issue about explicit sinks: https://github.com/rerun-io/rerun/issues/10444 - could you link it here with a TODO to make it clear that we'd like to get rid of it when we're living in that shiny future?

docs/snippets/all/howto/check_connection_status.rs
```diff
@@ -0,0 +1,40 @@
...
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let rec = rerun::RecordingStreamBuilder::new("rerun_example_check_connection_status")
+        .connect_grpc()?;
+
+    let (tx, rx) = std::sync::mpsc::channel();
+
+    loop {
+        let tx = tx.clone();
+        rec.inspect_sink(move |sink| {
```
jprochazk, line 13: As they say, nothing more permanent than a temporary solution 😄

jprochazk (approved):
I can't really think of anything else that wouldn't be just as awkward right now, so let's go with this

crates/top/re_sdk/src/recording_stream.rs
Wumpf: agreed. Thank you for the issue, will link that here!

docs/snippets/all/howto/check_connection_status.rs
Wumpf: admittedly that part makes me mega nervous. This solution is _definitely_ not great and I'd hope it will fold into something much nicer in the future, but maybe I just made the nicer future harder still? :/

jprochazk: :shrug: this seems easy to rip out, so I'm not too worried

PR #10435, opened 2025-06-30
Section on component descriptors in `v0.24` migration guide

### LINKED ISSUE #10059: Write migration guide for `ComponentDescriptor` changes
Should contain the following sections

- [ ] Changes to the viewer (only visualize tagged components)
- [ ] Changes to the SDKs
- [ ] Reloading of `LeRobot` datasets due to untagged logging
  - [ ] Depends on #10031

grtlr:
### Related

* Part of #6889.
* Closes #10059.

### What

Title.

### DIFF
--- a/docs/content/reference/migration/migration-0-24.md
+++ b/docs/content/reference/migration/migration-0-24.md
@@ -45,50 +45,44 @@ snippet: howto/serve_web_viewer
 
 Rust's even older `serve` (deprecated since 0.20) has been removed entirely now.
 
-## Tagged components
+## Component descriptors
 
-<!-- TODO(grtlr): These are ad-hoc notes from https://github.com/rerun-io/rerun/pull/10082 and need to be cleaned up! -->
+One limitation that we previously had with our data model was that it was only possible to use any `Component` once per entity path.
+This severely effected the design and flexbility of our archetypes.
+The underlying reason for this was that the our internal datastructures used [`Component`] to index into the data.
+This release changes how components are identified within the viewer and within our APIs:
+Instead of specifying the [`Component`] name, components are now referenced by a new syntax that consists of the (short) archetype name + the archetype field name separated by a colon (`:`).
 
-#### `re_types_core`
+As an example, `Points3D:positions` refers to the `positions` component (`rerun.components.Position3D`) in the `rerun.archetypes.Points3D` archetype.
+For custom data, such as `AnyValues`, it is possible to omit the archetype part of this syntax and only specify the fied name.
 
-* The obvious `ComponentDescriptor` changes.
+Internally we use, what we call a `ComponentDescriptor` to uniquely identify a component.
+Conceptually, it looks like the following:
 
-#### `re_types`
-
-* Indicator components now become indicator fields. This is only temporary until we get rid of indicators.
-
-#### `re_log_types`
-
-* Removed auto-filling of `rerun.components.` to `ComponentType` in path parsing.
-
-#### `re_sorbet`
-
-* Lookup `ColumnDescriptor` by `ComponentIdentifier` instead of `ComponentType`.
-* Changed `ComponentColumnSelector`.
-* Changed `ComponentColumnDescriptor::column_name` to use fully-qualified column names.
-
-#### `rerun_py`
-
-* Dataframe queries now use the new string representation everywhere. It's not allowed to pass in components anymore.
-* `resolve_component_column_selector` now returns an `Option`.
-
-With tagged components, our data model becomes much more flexible.
-We can now have multiple components of the same type living on the same archetype, so we also need a new way to identify columns in a recording.
-For this we introduce the following new string-based representation:
-
-```
-<entity_path>:[<archetype_name>]:<component>
+```rs
+struct ComponentDescriptor {
+  archetype:      Option<String>, // e.g. `rerun.archetypes.Points3D
+  component:      String,         // e.g. `Points3D:positions`
+  component_type: Option<String>, // e.g. `rerun.components.Position3D
+}
 ```
 
-Note that the `archetype_name` section is optional, because components can also be logged as plain fields.
+### Changes
 
-### Blueprints
+When logging data to Rerun using the builtin archetypes no changes to user code should be required.
+There is also migration code in place so that you can open `.rrd` files that were created with `v0.23`.
 
-We have updated our logic so that the `component` field of `blueprint.datatypes.ComponentColumnSelector` follows the same schema.
+These changes are reflected in various parts of Rerun:
 
-#### LeRobot dataloader
+* The selection panel UI comes with a revamped display of archetypes that uses the new syntax to show the `ComponentDescriptor` for each component.
+* The new `:`-based syntax needs to be used when referring to components in the dataframe API and in the dataframe view.
+* Custom data can still use simple field names such as `confidences`, but it is advised to supply a full `ComponentDescriptor`, if possible.
 
-* Fixed an issue where the LeRobot dataloader logged untagged `Name` components for robot observations and actions, `.rrd` files created before `0.24` may include these untagged entries. To fix this, load the dataset in `0.24.0` and resave your episodes to `.rrd` (`0.24.0` now supports saving all selected recordings).
+### Limitations
+
+* In rare cases, it is not possible to migrate previous blueprints, _but only if they were saved from the viewer via the UI.
+* Currently, only Rerun-builtin components are picked up by the visualizers and therefore shown in the views (except for the dataframe view which shows all components).
+* In `v0.23`, the LeRobot dataloader logged incomplete `ComponentDescriptors` for robot observations and actions. To fix this, load the dataset in `v0.24` and resave your episodes to `.rrd` (`v0.24` now supports saving all selected recordings).
 
 ## Dataframe API: `View.select_static` is deprecated
 
--- a/docs/content/reference/migration/migration-0-24.md
+++ b/docs/content/reference/migration/migration-0-24.md
@@ -49,21 +49,25 @@ Rust's even older `serve` (deprecated since 0.20) has been removed entirely now.
 
 One limitation that we previously had with our data model was that it was only possible to use any `Component` once per entity path.
 This severely effected the design and flexbility of our archetypes.
-The underlying reason for this was that the our internal datastructures used [`Component`] to index into the data.
+The underlying reason for this was that the our internal datastructures used the component's type names (previously also referred to as `ComponentName`) to index into the data.
 This release changes how components are identified within the viewer and within our APIs:
-Instead of specifying the [`Component`] name, components are now referenced by a new syntax that consists of the (short) archetype name + the archetype field name separated by a colon (`:`).
+Instead of specifying the component's type name, components are now referenced by a new syntax that consists of the (short) archetype name + the archetype field name separated by a colon (`:`).
 
-As an example, `Points3D:positions` refers to the `positions` component (`rerun.components.Position3D`) in the `rerun.archetypes.Points3D` archetype.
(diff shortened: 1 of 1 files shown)


docs/content/reference/migration/migration-0-24.md
```diff
@@ -45,50 +45,52 @@ snippet: howto/serve_web_viewer
 
 Rust's even older `serve` (deprecated since 0.20) has been removed entirely now.
 
-## Tagged components
+## Component descriptors
 
-<!-- TODO(grtlr): These are ad-hoc notes from https://github.com/rerun-io/rerun/pull/10082 and need to be cleaned up! -->
+One limitation that we previously had with our data model was that it was only possible to use any `Component` once per entity path.
+This severely effected the design and flexbility of our archetypes.
+The underlying reason for this was that the our internal datastructures used [`Component`] to index into the data.
```
Wumpf: are you referring to the component rust struct? Those brackets also not doing anything.

Maybe instead
```suggestion
The underlying reason for this was that the our internal datastructures used the component's type names (previously also referred to as `ComponentName`) to index into the data.
```

not sure if mentioning `ComponentName` is just more confusing

docs/content/reference/migration/migration-0-24.md
```diff
@@ -45,50 +45,52 @@ snippet: howto/serve_web_viewer
...
+This release changes how components are identified within the viewer and within our APIs:
+Instead of specifying the [`Component`] name, components are now referenced by a new syntax that consists of the (short) archetype name + the archetype field name separated by a colon (`:`).
 
-#### `re_types_core`
+As an example, `Points3D:positions` refers to the `positions` component (`rerun.components.Position3D`) in the `rerun.archetypes.Points3D` archetype.
```
Wumpf: the `rerun.components.Position3D` bit is actually confusing since it doesn't show up here anymore at all. Maybe

```suggestion
As an example, `Points3D:positions` refers to the `positions` component in the `rerun.archetypes.Points3D` archetype.
(Previously, it would only be identified by its component type, i.e. `rerun.components.Position3D`)
```

docs/content/reference/migration/migration-0-24.md
```diff
@@ -45,50 +45,52 @@ snippet: howto/serve_web_viewer
...
 
-## Tagged components
+## Component descriptors
 
-<!-- TODO(grtlr): These are ad-hoc notes from https://github.com/rerun-io/rerun/pull/10082 and need to be cleaned up! -->
+One limitation that we previously had with our data model was that it was only possible to use any `Component` once per entity path.
+This severely effected the design and flexbility of our archetypes.
+The underlying reason for this was that the our internal datastructures used [`Component`] to index into the data.
+This release changes how components are identified within the viewer and within our APIs:
+Instead of specifying the [`Component`] name, components are now referenced by a new syntax that consists of the (short) archetype name + the archetype field name separated by a colon (`:`).
```
Wumpf: as above the brackets and component name thing doesn't make sense in this md file I think

```suggestion
Instead of specifying component's type name, components are now referenced by a new syntax that consists of the (short) archetype name + the archetype field name separated by a colon (`:`).
```

docs/content/reference/migration/migration-0-24.md
```diff
@@ -45,50 +45,52 @@ snippet: howto/serve_web_viewer
...
 
-#### `re_types_core`
+As an example, `Points3D:positions` refers to the `positions` component (`rerun.components.Position3D`) in the `rerun.archetypes.Points3D` archetype.
+For custom data, such as `AnyValues`, it is possible to omit the archetype part of this syntax and only specify the field name.
+To uniqule identify a component in a recording we can prefix the entity path, for example: `/helix/structure/left:Points3D:positions`.
```
Wumpf:
```suggestion
To uniquely identify a component in a recording ww prefix the entity path, for example: `/helix/structure/left:Points3D:positions`.
```

not sure if confusing. we _already_ used the entity path, it's just that now we have a syntax for putting everything into a single specifier. Which is relevant for data queries but not for putting data in or the Viewer which is the stuff that we talked about so far

docs/content/reference/migration/migration-0-24.md
```diff
@@ -45,50 +45,52 @@ snippet: howto/serve_web_viewer
...
+In general, the syntax for uniquely identifying a component in recording then becomes:
```
Wumpf, line 62: related to above: as a user should I care where do identify things this way?

docs/content/reference/migration/migration-0-24.md
```diff
@@ -45,50 +45,52 @@ snippet: howto/serve_web_viewer
...
+<entity_path>:[<archetype>:]<field>
+```
 
-* Dataframe queries now use the new string representation everywhere. It's not allowed to pass in components anymore.
-* `resolve_component_column_selector` now returns an `Option`.
 
-With tagged components, our data model becomes much more flexible.
-We can now have multiple components of the same type living on the same archetype, so we also need a new way to identify columns in a recording.
-For this we introduce the following new string-based representation:
+Internally we use, what we call a `ComponentDescriptor` to uniquely identify a component.
```
Wumpf: that's semi-wrong though? Only `component` is now used for _identifying_ the component. `component_type` and `archetype` are just optional metadata at this point. It _happens_ that the viewer right now uses the full `ComponentDescriptor` for hashmaps, but in my understanding that is just because we haven't gotten around to update things again 🤔 

Also confusing because above you talk about how the entity path is part of how things are identified.

All in all maybe we should just not talk about component descriptor at all then or rather only mention that archetype and component_type is metadata we track for each component that we expect to be directly tied to `component` on a given entity.
(as long as that assumption is correct, the viewer using full `ComponentDescriptor` as key doesn't do anything wrong, it's just wasteful which is why we get away with it right now :))

plz correct me if I got something wrong!

docs/content/reference/migration/migration-0-24.md
```diff
@@ -45,50 +45,52 @@ snippet: howto/serve_web_viewer
...
+}
 ```
-<entity_path>:[<archetype_name>]:<component>
-```
 
-Note that the `archetype_name` section is optional, because components can also be logged as plain fields.
+### Changes
+
+When logging data to Rerun using the builtin archetypes no changes to user code should be required.
+There is also migration code in place so that you can open `.rrd` files that were created with `v0.23`.
```
Wumpf, line 87: data from 0.22 should also work if you go through the migration steps since starting there we got the necessary information. Before that we don't know about field names and archetypes. I think that's worth pointing out

docs/content/reference/migration/migration-0-24.md
```diff
@@ -45,50 +45,52 @@ snippet: howto/serve_web_viewer
...
 
-### Blueprints
+These changes are reflected in various parts of Rerun:
 
-We have updated our logic so that the `component` field of `blueprint.datatypes.ComponentColumnSelector` follows the same schema.
+* The selection panel UI comes with a revamped display of archetypes that uses the new syntax to show the `ComponentDescriptor` for each component.
+* The new `:`-based syntax needs to be used when referring to components in the dataframe API and in the dataframe view.
+* Custom data can still use simple field names such as `confidences`, but it is advised to supply a full `ComponentDescriptor`, if possible.
+* Changed the interpretation of `blueprint.datatypes.ComponentColumnSelector` to use the new component identifier.
```
Wumpf, line 96: worth repeating that you can now log arbitrarily overlapping archetypes on a single entity path. That's a killer feature we can't mention often enough!

docs/content/reference/migration/migration-0-24.md
```diff
@@ -45,50 +45,52 @@ snippet: howto/serve_web_viewer
...
 
-#### LeRobot dataloader
+### Limitations
 
-* Fixed an issue where the LeRobot dataloader logged untagged `Name` components for robot observations and actions, `.rrd` files created before `0.24` may include these untagged entries. To fix this, load the dataset in `0.24.0` and resave your episodes to `.rrd` (`0.24.0` now supports saving all selected recordings).
+* In rare cases, it is not possible to migrate previous blueprints, _but only if they were saved from the viewer via the UI.
```
Wumpf:
```suggestion
* In some cases, it is not possible to migrate previous blueprints files, _but_ only if they were saved from the viewer via the UI.
```

don't actually think it's _that_ rare ;-)

docs/content/reference/migration/migration-0-24.md
```diff
@@ -45,50 +45,52 @@ snippet: howto/serve_web_viewer
...
+* Currently, only Rerun-builtin components are picked up by the visualizers and therefore shown in the views (except for the dataframe view which shows all components).
+* In `v0.23`, the LeRobot dataloader logged incomplete `ComponentDescriptors` for robot observations and actions. To fix this, load the dataset in `v0.24` and resave your episodes to `.rrd` (`v0.24` now supports saving all selected recordings).
```
Wumpf, line 103: Something like: 

* overriding visualizers to reinterpret data (e.g. show a point-cloud for mesh vertices) is no longer possible, since visualizers now match for `<archetype>:<field>` instead of component type name. This will be addressed in the future with blueprint driven overrides that will allow to remap data to arbitrary 

(avoiding the outdated term "descriptor overrides here)

docs/content/reference/migration/migration-0-24.md
grtlr: Sorry, muscle memory

docs/content/reference/migration/migration-0-24.md
grtlr: You're right, but I wanted to get `ComponentDescriptor`s in there for the `AnyValues` story.

I now made this into a new section.

docs/content/reference/migration/migration-0-24.md
grtlr: 👍

grtlr:
Thank you for the feedback, a new version is up!

docs/content/reference/migration/migration-0-24.md
```diff
@@ -45,50 +45,60 @@ snippet: howto/serve_web_viewer
...
 
-* Fixed an issue where the LeRobot dataloader logged untagged `Name` components for robot observations and actions, `.rrd` files created before `0.24` may include these untagged entries. To fix this, load the dataset in `0.24.0` and resave your episodes to `.rrd` (`0.24.0` now supports saving all selected recordings).
+* In some cases, it is not possible to migrate previous blueprints, _but only if they were saved from the viewer via the UI_.
+* Currently, only Rerun-builtin components are picked up by the visualizers and therefore shown in the views (except for the dataframe view which shows all components).
+* In `v0.23`, the LeRobot dataloader logged incomplete `ComponentDescriptors` for robot observations and actions. To fix this, load the dataset in `v0.24` and resave your episodes to `.rrd` (`v0.24` now supports saving all selected recordings).
+* Overriding visualizers to reinterpret data (e.g. show a point-cloud for mesh vertices) is no longer possible, since visualizers now match for <archetype>:<field> instead of component type name. This will be addressed in the future with blueprint-driven overrides that will allow to remap data to arbitrary archtypes.
```
Wumpf:
```suggestion
* Overriding visualizers to reinterpret data (e.g. show a point-cloud for mesh vertices) is no longer possible, since visualizers now match for <archetype>:<field> instead of component type name. This will be addressed in the future with blueprint-driven overrides that will allow to remap data to arbitrary archetypes.
```

Wumpf:
> worth repeating that you can now log arbitrarily overlapping archetypes on a single entity path. That's a killer feature we can't mention often enough!

Maybe this isn't so much migration guide territory, but I think this should still be in there (and if I didn't read to hastily this didn't show up yet?)

grtlr:
You're right, I've overlooked at when resolving and then forgot about it.

PR #10436, opened 2025-06-30
Add `ar` notice to BUILD.md for Windows

oxkitsune:
### What

Adds a notice that some sort of `ar` is required to build the Web Viewer on Windows.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -140,11 +140,28 @@ This is automatically done by this shortcut which builds & runs the web viewer:
 pixi run rerun-web
 ```
 
+### Windows
+
 If you're on Windows you have to make sure that your git client creates symlinks,
 otherwise you may get errors during the build.
 Run `git config --show-scope --show-origin core.symlinks` to check if symlinks are enabled.
 You may need to turn on Windows developer mode in order to give the `mklink` command sufficient permissions.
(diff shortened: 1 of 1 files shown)


oxkitsune:
I think we should be able to. I'm not able to test that on my Windows machine currently, but can have a go at it tonight. 
I think I can omit the `config.toml` stuff for sure, however I made so many changes last thursday that I wasn't sure whether I did end up needing that.

Wumpf:
let's coordinate on company slack when we get to it respectively!

Wumpf:
Can't find said Slack discussion. Did we ever talk about this..?

Either way, I don't seem to have `ar` on PATH on Windows but `pixi run rerun-web` works just fine for me off `main`. Do you remember what error you hit back then?

oxkitsune:
I don't know the exact error I got, but it was something along the lines of:
```
error: failed to build archive: no such file or directory
```

I don't have access to my windows machine at the moment, so I cannot try to get a more specific diagnosis.

Wumpf:
Let's revisit then at some point in the future when you do - not closing this off because I don't want to sweep workflow issues under the rug

emilk:
Related:
* https://github.com/rerun-io/rerun/issues/9849
* https://github.com/rerun-io/rerun/pull/10828

Wumpf:
No more repro for anyone 🤷

PR #10438, opened 2025-06-30
Unify `ComponentDescriptor` handling across all SDKs

grtlr:
### Related

* Follow-up of #10245.

### What

* Refactored viewer conventions into `ComponentDescriptorExt`
* This makes the docs on `ComponentDescriptor` much cleaner.
* Identified places in our code that needed some more love.
* Copied these docs to C++ and Python SDKs
* Unified the component descriptors in snippets.
* Removed fuzzy handling of `ComponentIdentifier` (used to accept both short and full-qualified identifiers)

### TODO

* [x] full check
* [x] `.pyi` bindings


grtlr:
@rerun-bot full-check

crates/store/re_types_core/src/component_descriptor.rs
```diff
@@ -15,7 +15,7 @@ pub struct ComponentDescriptor {
     /// Example: `rerun.archetypes.Points3D`.
     pub archetype: Option<ArchetypeName>,
 
-    /// Name of the component associated with this data.
+    /// Identifier of the component associated with this data.
     ///
     /// If `archetype` is `None`, this will be a simple field name.
     /// [`ArchetypeName::with_field`] is a convenient method to create a [`ComponentIdentifier`].
```
Wumpf: while we're here: I think we should point out that this is true for builtin types but generally so it's nothing we should rely upon
(...right?)

rerun_py/src/python_bridge.rs
```diff
@@ -1325,6 +1327,109 @@ fn flush(py: Python<'_>, blocking: bool, recording: Option<&PyRecordingStream>)
...
+    ///
+    /// Example: `rerun.components.Position3D`.
+    #[getter]
+    fn component_type(&self) -> Option<String> {
+        self.0.component_type.map(|a| a.to_string())
+    }
+
+    /// Unconditionally sets `archetype` & `component_type` to the given ones (if specified).
+    ///
+    /// This also changes the archetype part of the descriptor.
```
Wumpf:
```suggestion
    /// This also changes the archetype part of the component field.
```

rerun_py/src/python_bridge.rs
```diff
@@ -1325,6 +1327,109 @@ fn flush(py: Python<'_>, blocking: bool, recording: Option<&PyRecordingStream>)
...
+        }
+        if let Some(component_type) = component_type {
+            cloned = cloned.with_component_type(component_type.into());
+        }
+        Self(cloned)
+    }
+
+    /// Sets `archetype` & `component_type` to the given one iff it's not already set.
+    ///
+    /// This also changes the archetype part of the descriptor.
```
Wumpf:
```suggestion
    /// This also changes the archetype part of the component field.
```

rerun_py/src/python_bridge.rs
```diff
@@ -1325,6 +1327,109 @@ fn flush(py: Python<'_>, blocking: bool, recording: Option<&PyRecordingStream>)
...
+    ///
+    /// Example: `Points3D:positions`. Warning: Never parse this string to retrieve an archetype!
+    #[getter]
+    fn component(&self) -> String {
+        self.0.component.to_string()
+    }
+
+    /// Optional, semantic type associated with this data.
+    ///
+    /// This is fully implied by the `component`, but included for semantic convenience.
```
Wumpf: I'm unsure about the "implied". I mean yes you're not supposed to change it for a given `component` therefore there's a stable relationship. But for custom (!) types we can't look it up, so it's not so strongly implied as that we could infer it. Therefore, the information in here may still be valuable for whatever we're doing, beyond just convenience 🤔 
(may also apply to existing code comments elsewhere)

rerun_py/src/python_bridge.rs
```diff
@@ -1325,6 +1327,109 @@ fn flush(py: Python<'_>, blocking: bool, recording: Option<&PyRecordingStream>)
     });
 }
 
+// --- Components ---
+
+/// A `ComponentDescriptor` fully describes the semantics of a column of data.
+///
+/// Every component is uniquely identified by its `ComponentDescriptor`.
```
Wumpf: similar to a comment I made on the migration guide (you fixed it there :)): `ComponentDescriptor` isn't meant to be the uniquely identifying thing, that's `ComponentDescriptor::component`. Also worth pointing out that this is regardless only true within a given entity path.
we don't ever expect the other semantics to change for a given `ComponentDescriptor::component`, but they (should) not be part of the identifying thing

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -749,6 +749,65 @@ def set_thread_local_blueprint_recording(
     Returns the previous one, if any.
     """
 
+#
+# component descriptor
+#
+
+class ComponentDescriptor:
```
Wumpf, line 756: [code comments here need updating depending on comments below]

grtlr:
@rerun-bot full-check

grtlr:
@rerun-bot full-check

crates/store/re_log_types/src/path/parse_path.rs
```diff
@@ -92,29 +89,6 @@ impl std::str::FromStr for DataPath {
                 return Err(PathParseError::TrailingColon);
             }
```
Wumpf, line 91: worth leaving a note why we're not parsing out the archetype information here
... I know @jleibs wants it to be this way here because archetype should be separate optional information, but it actually makes me a bit uneasy. We're still reading out the component type here though, so I'm not sure whether that's correct. Isn't the whole syntax used here now outdated, as we no longer deliminate field names with # ?

crates/store/re_types_core/src/reflection.rs
```diff
@@ -285,3 +285,96 @@ impl ArchetypeFieldReflection {
...
+/// Note that in general, [`ComponentDescriptor::archetype`] does not place any
+/// constraints on the [`ComponentDescriptor::component`] field, which can be arbitrary.
+///
+/// Components that are built into the Rerun viewer follow the convention that
+/// the `component` field starts with the short name of its archetype.
+pub trait ComponentDescriptorExt {
+    /// Returns the field name of a Rerun-builtin type.
+    ///
+    /// This is the result of stripping the Rerun-builtin [`ArchetypeName`]
+    /// from [`ComponentDescriptor::component`].
```
Wumpf, line 301: but for builtin types we expect the archetype name to be present separately, no? So why would we then go to the `component`? 🤔

grtlr:
@rerun-bot full-check

crates/store/re_types_core/src/reflection.rs
grtlr: After the renames, `archetype_field_name` refers to the field part of a `component` to show in the UI.

Let's say you have a `component` (`ComponentIdentifier`) `Points3D.positions`. This method returns the `positions`. This is used for example to show such a UI:

image

I could also be misunderstanding your question 🙈.

crates/store/re_log_types/src/path/parse_path.rs
grtlr: Yeah, there is some discussion to be had around if we still need this. In fact, iirc, we never parse `ComponentPath` in our code anymore.

The reason that I left the `#` syntax  for now, is that I think it would be helpful later on to do _casts_, i.e. give me this component `as` this `component_type` (if such a coercion is supported).

crates/store/re_types_core/src/reflection.rs
Wumpf: I was reading it wrong, nvm makes all sense!

Wumpf:
@rerun-bot full-check

PR #10442, opened 2025-06-30
Generate one partition stream per partition ID in query provider

timsaucer:
### Related

https://linear.app/rerun/issue/DPF-1796/niko-feedback-triage-sandbox-todo

### What

This PR changes the way we handle partition streams in a query provider. Now we create one `PartitionStream` per partition ID in the data platform. This gives a performance improvement because (a) we no longer have to do a flat map from partition + chunk -> single iterator and (b) allows for greater splitting up of the data processing in datafusion since we now have more than one output stream.

Prior to this change, running the benchmarks locally against `droid:raw`:

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━┓
┃ Benchmark Group/Function     ┃ droid:raw ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━┩
│ Droid                        │           │
│   droid:count_gripper_closes │ 3220.73ms │
└──────────────────────────────┴───────────┘
```

With this change:

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━┓
┃ Benchmark Group/Function     ┃ droid:raw ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━┩
│ Droid                        │           │
│   droid:count_gripper_closes │ 656.21ms  │
└──────────────────────────────┴───────────┘
```

I have also validated that the dataframes returned from this call is unchanged.

### DIFF
--- a/crates/store/re_datafusion/src/dataframe_query_provider.rs
+++ b/crates/store/re_datafusion/src/dataframe_query_provider.rs
@@ -13,10 +13,9 @@ use itertools::Itertools as _;
 use re_dataframe::{QueryEngine, QueryExpression, StorageEngine};
 use re_protos::manifest_registry::v1alpha1::DATASET_MANIFEST_ID_FIELD_NAME;
 
+#[derive(Debug)]
 pub struct DataframeQueryTableProvider {
     pub schema: SchemaRef,
-    query_expression: QueryExpression,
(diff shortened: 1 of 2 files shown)


jleibs (approved):
nice!

PR #10445, opened 2025-06-30
WIP Add Eye3D to Spatial3DView - First Draft

jccampagne:
This is WIP (work in progress), the PR is to get early feedback on the general approach for https://github.com/rerun-io/rerun/issues/6825 - Still more to do (definitions are incomplete, not everything is plugged in the UI, etc.)

Seeking feedback on any level (general approach, naming scheme, ...), especially around how data (property) is retrieve and converted (see `fn extract_eye_spin_speed(eye_property: &ViewProperty)` and `eye_property = ViewProperty::from_archetype::<Eye3D>(...)`).

For now, the only property of the 3D view eye that one can interact with is the spin speed (of the eye). It is a "new" property, it was chosen as it was the simplest to test initially, we don't have to keep it; although it's a bit more flexible that a boolean flag for spin.

Definition in [eye3d.fbs](https://github.com/rerun-io/rerun/compare/main...jccampagne:rerun:eye_properties?expand=1#diff-63f8e7eaa9245c27e10d86414f212bff98116e59ad1f397beff9033659216e59) is tentative:
- Properties were marked `attr.rerun.component_required`, but some would probably be optional?
- Attributes used in code generation are mostly missing.
- properties such as `position`, `target` are placeholders - not used; happy to remove them or change the type, etc.

# UI Preview

The eye properties, as defined, show up in the UI:

image

# Python bindings:

Python bindings have been tested with the following code. The eye spins (the res
```
    view3d = rrb.Spatial3DView(
        origin="/",
        name="3D Scene",
        eye_3d=rrb.Eye3D(
            position=rr.components.Position3D([1, 2, 2]),
            kind=2,
            speed=12345.0,
            spin_speed=15.0, # <--- this works!
            target=rr.components.Position3D(
                rr.datatypes.Vec3D([0, 0, 0]),
            ),
        ),
    )
    blueprint = rrb.Blueprint(
        view3d,
        collapse_panels=False,
    )
    rr.send_blueprint(blueprint)
```

Any comments or suggestions are welcome...


Wumpf:
Thanks for tackling this and happy to see you found your way around our whacky codegen 😄

I think though  we should take this step by step and first only do the two trivially portable ones: speed and camera type.
Everything else likely means also changing how the camera internally works since data has to flow back and forth from the blueprint store continuously. Also, we need to try to be a bit forward looking for different camera models that we'll sill want to add - lots of folks are very unhappy with our default orbit camera.
Yes we absolutely want things like camera position and direction in the blueprint, but that's gonna be trickier and it's nice to separate that into a follow-up iteration.

I think for speed we'd actually like to introduce a new component type since it has a different semantic than `Scalar`. I think a component type `Speed` which is documented as "units per second" could come in handy elsewhere as well.

> Properties were marked attr.rerun.component_required, but some would probably be optional?

Yeah I think everything in here should be `"attr.rerun.component_optional", nullable`. Whenever a value is null in the blueprint we should fall back to current default behavior.
Guess from an API standpoint it's debatable whehter the camera type should always be specified if one specifies the eye camera. But in the future people may not care for scenes where they place the camera once and never move it.

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -339,6 +345,25 @@ impl View3DState {
     }
 }
 
+fn extract_eye_spin_speed(eye_property: &ViewProperty) -> Option<f32> {
+    let eye_spin_comp_descr = re_types::ComponentDescriptor {
+        archetype: Some("rerun.blueprint.archetypes.Eye3D".into()),
+        component: "Eye3D:spin_speed".into(),
+        component_type: Some("rerun.components.Scalar".into()),
+    };
```
Wumpf: you never need to do this yourself, you can always query the generated onces

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -339,6 +345,25 @@ impl View3DState {
...
+    if let Ok(Some(speed_arr)) = eye_spin_component {
+        if let Some(Scalar(re_types::datatypes::Float64(speed))) = speed_arr.first() {
+            if *speed > 0.0 {
+                // FIXME: see how conversion is done in the rest of the code base
+                let speed = *speed as f32;
+                return Some(speed);
+            }
+        }
+    }
+    None
```
Wumpf: we know that we want a single component here, so you can just use `component_or_empty`
or better still use `component_or_fallback` and implement the necessary fallback provider on the view
The other fallback providers for the 3d view are in `view_3d_properties.rs`

this way we have access to fallbacks from everywhere in the viewer - IIRC this isn't used for view properties much yet but it means that in the future we can tell users about the fallback values in the ui for individual properties :)

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -411,6 +419,19 @@ impl ViewEye {
             }
         });
 
+        let kind = eye_property.component_or_fallback::<Eye3DKind>(
+            view_ctx,
+            self,
+            &Eye3D::descriptor_kind(),
+        );
+        match kind {
```
jccampagne, line 383: I suppose we will "merge" the enums `EyeMode` and `Eye3DKind` (either replace or rename) and then this `match` will go away.

crates/viewer/re_view_spatial/src/ui_3d.rs
jccampagne: thanks - using `Eye3D::descriptor_translation_speed()` now

crates/viewer/re_view_spatial/src/ui_3d.rs
jccampagne: using `eye_property.component_or_fallback::<LinearSpeed>()` now

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -472,6 +493,26 @@ impl ViewEye {
...
+        eye_property: &ViewProperty,
+    ) -> f32 {
+        let eye_linear_speed = eye_property.component_or_fallback::<LinearSpeed>(
+            view_ctx,
+            self,
+            &Eye3D::descriptor_translation_speed(),
+        );
+        match eye_linear_speed {
+            Ok(linear_speed) => **linear_speed as f32,
+            Err(err) => {
```
jccampagne, line 509: Not sure when this can happen?

jccampagne:
PR is not ready to merge - still not finished, see below.

## Definitions

>I think though we should take this step by step and first only do the two trivially portable ones: speed and camera type.


* Updated the definitions (`.fbs`) accordingly. Eye3D has 2 attributes now:
  * `Eye3DKind` - we can change the name "kind".
  * `Translation Speed`:  more explicit than just `Speed` - to differentiate from angular or rotational speed
* create a new component: `Linear Speed`

The "old" code (`EyeMode` and previous UI) has been removed. ~~has not been removed yet, I left it for testing; but it now probably gets in the way. This [match](https://github.com/rerun-io/rerun/pull/10445/files#diff-242e5fb1310bf1a865d6c25402303099d00d6a81e76137ccfaaf36db8987b24bR427) should eventually go away.~~

## Float types

Previous speed related code used `f32`.
The new definition, inspired by existing definition uses `f64` or `Float64`.
Should it be aligned, or if heterogeneous at what boundary should the precision conversion happen (UI/code)? Is code in the current PR Ok about this?

~~## Time of last update~~

~~The check on `last_eye_interaction.is_none()` was disabled as it conflicts with the new approach (see https://github.com/rerun-io/rerun/pull/10445/files#diff-6cdab04b74cba7398db1581a0db86cb74529e420ad8a2587c91ce43a5c71cd2dR128-R133)~~

~~I tried to emulate this check by looking at if `eye_property.query_results.compound_index.is_static()`, but not sure it works as intended... (see commented code 
https://github.com/rerun-io/rerun/pull/10445/files#diff-6cdab04b74cba7398db1581a0db86cb74529e420ad8a2587c91ce43a5c71cd2dR135-R141 )~~

## Python code

it looks like this now:

```
def run():
    rr.init("test 1", spawn=True)
    positions = [[0, 0, 0], [5, 0, 0]]
    colors = [[200, 0, 0], [0, 200, 0]]
    radii = [1, 2]

    rr.log("points", rr.Points3D(positions, colors=colors, radii=radii))

    # Create a Spatial3D view to display the points.
    view3d = rrb.Spatial3DView(
        origin="/",
        name="3D Scene",
        eye_3d=rrb.Eye3D(
            # kind=rr.blueprint.components.Eye3DKind.FirstPerson,
            kind="FirstPerson",
            translation_speed=12345.0,
        ),
    )

    blueprint = rrb.Blueprint(
        view3d,
        collapse_panels=False,
    )
    rr.send_blueprint(blueprint)
```

crates/store/re_types/definitions/rerun/blueprint/datatypes/component_column_selector.fbs
```diff
@@ -15,7 +15,7 @@ table ComponentColumnSelector (
   /// The name of the component.
   ///
   /// This acts as the component name in the context of a given `entity_path`
-  /// An example for this would be `Points3D:positions`, for the `positions` field in the `Points3D` archetype.
+  /// An example for this would be `Points3D:positions`, for the `positions` field in the [archetypes.Points3D] archetype.
```
Wumpf: ah thx, ci really ought to do warnings-as-errors on this. I think I filed a ticker for that somewhere :)

Wumpf:
sounds like this is shaping up really well! :)

> I tried to emulate this check by looking at if eye_property.query_results.compound_index.is_static(), but not sure it works as intended... (see commented code
[#10445 (files)](https://github.com/rerun-io/rerun/pull/10445/files#diff-6cdab04b74cba7398db1581a0db86cb74529e420ad8a2587c91ce43a5c71cd2dR135-R141) )

that whole thing was a bit weird. I can help testing & debugging it when you're ready :)

jccampagne:
On first display, and before interaction, the UI element for the `Linear Speed` of the eye does not show the correct value; and it is not updating as the scene is loading... Looking into it...
The fallback should be called, but it isn't.

jccampagne:
Had to implement:
- `impl TypedComponentFallbackProvider<LinearSpeed> for SpatialView3D`
- `impl TypedComponentFallbackProvider<Eye3DKind> for SpatialView3D`

Hope this is the right approach.

Looks like it works: UI is updating the linear speed accordingly - it's quite apparent on the air control example, as it takes a while to load and the bounding boxes change.

https://github.com/user-attachments/assets/151ae300-77ef-48cc-b555-d1b1ff249e3b

crates/store/re_types/definitions/rerun/blueprint/archetypes/eye3d.fbs
```diff
@@ -0,0 +1,14 @@
+namespace rerun.blueprint.archetypes;
+
+/// 3D Eye in a spatial 3D view
+table Eye3D (
+  "attr.rerun.scope": "blueprint"
+) {
+
+  /// Eye kind
```
Wumpf: should be more descriptive of what this is good for.

crates/store/re_types/definitions/rerun/blueprint/components/eye_3d_kind.fbs
```diff
@@ -0,0 +1,18 @@
...
+    "attr.rerun.scope": "blueprint"
+) {
+    /// Invalid value. Won't show up in generated types.
+    Invalid = 0,
+
+    /// First person eye
+    FirstPerson,
+
+    /// Orbital eye
+    Orbital (default),
```
Wumpf: should be more descriptive of what these are exactly and how they're controlled

crates/store/re_types/definitions/rerun/blueprint/components/eye_3d_kind.fbs
```diff
@@ -0,0 +1,18 @@
+namespace rerun.blueprint.components;
+
+
+/// The type of the 3D eye
```
Wumpf: As above, this is a bit thin; as a new comer I wouldn't know how this is useful

crates/store/re_types/definitions/rerun/blueprint/views/spatial3d.fbs
```diff
@@ -12,6 +12,9 @@ table Spatial3DView (
     /// Configuration for the 3D line grid.
     line_grid: rerun.blueprint.archetypes.LineGrid3D (order: 2000);
 
+    /// Configuration for the 3D eye
```
Wumpf, line 15: should describe what eye means here, might not be clear to everyone

crates/store/re_types/definitions/rerun/blueprint/views/spatial3d.fbs
```diff
@@ -12,6 +12,9 @@ table Spatial3DView (
...
+    eye_3d: rerun.blueprint.archetypes.Eye3D (order: 3000);
```
Wumpf: the 3d part on the field name (not on the type so much) is redundant, everything in here is about 3D
Still pondering whether we should call it 

```suggestion
    eye_camera: rerun.blueprint.archetypes.EyeCamera (order: 3000);
```

but I don't like it particular 🤔 . I'll ask around a bit

crates/viewer/re_component_ui/src/lib.rs
```diff
@@ -83,9 +84,11 @@ pub fn create_component_ui_registry() -> re_viewer_context::ComponentUiRegistry
     registry.add_singleline_edit_or_view::<GammaCorrection>(edit_f32_zero_to_max);
     registry.add_singleline_edit_or_view::<GridSpacing>(edit_f32_zero_to_max);
     registry.add_singleline_edit_or_view::<ImagePlaneDistance>(edit_f32_zero_to_max);
+    registry.add_singleline_edit_or_view::<LinearSpeed>(edit_f64_zero_to_max);
     registry.add_singleline_edit_or_view::<MarkerSize>(edit_ui_points);
-    registry.add_singleline_edit_or_view::<StrokeWidth>(edit_ui_points);
     registry.add_singleline_edit_or_view::<NearClipPlane>(edit_f32_zero_to_max);
+    registry.add_singleline_edit_or_view::<Scalar>(edit_f64_zero_to_max);
```
Wumpf: let's not add `Scalar` here - I'm not sure if it might show up in places where we don't want it, no reason to risk that

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -273,44 +255,26 @@ impl ViewEye {
...
-
-    /// Set the speed to a specific value set by the user via the UI.
-    ///
-    /// `None` resets speed to the default.
-    pub fn set_speed(&mut self, new_speed: Option<f32>) {
-        match new_speed {
-            Some(speed) => self.speed = CameraTranslationSpeed::Override(speed),
-            None => self.speed = CameraTranslationSpeed::Auto,
+    /// Compute the actual speed depending on the [`Eye3DKind`].
+    pub fn fallback_speed_for_mode(&self, bounding_boxes: &SceneBoundingBoxes) -> f32 {
```
```suggestion
    pub fn fallback_speed_for_kind(&self, bounding_boxes: &SceneBoundingBoxes) -> f32 {
```

let's be consequent :)

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -411,6 +375,19 @@ impl ViewEye {
...
+        let kind = eye_property.component_or_fallback::<Eye3DKind>(
+            view_ctx,
+            self,
+            &Eye3D::descriptor_kind(),
+        );
+        match kind {
+            Ok(Eye3DKind::FirstPerson) => self.set_kind(Eye3DKind::FirstPerson),
+            Ok(Eye3DKind::Orbital) => self.set_kind(Eye3DKind::Orbital),
+            Err(err) => {
+                re_log::error!("error while getting eye 3d kind: {}", err);
```
```suggestion
                re_log::error_once!("error while getting eye 3d kind: {}", err);
```

don't think we'll ever hit it, but might get spammy quickly if we do

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -472,6 +449,26 @@ impl ViewEye {
...
+    ) -> f32 {
+        let eye_linear_speed = eye_property.component_or_fallback::<LinearSpeed>(
+            view_ctx,
+            self,
+            &Eye3D::descriptor_translation_speed(),
+        );
+        match eye_linear_speed {
+            Ok(linear_speed) => **linear_speed as f32,
+            Err(err) => {
+                re_log::error!("Error while getting linear speed for eye {}", err);
```
Wumpf: as above, just as a precaution
```suggestion
                re_log::error_once!("Error while getting linear speed for eye {}", err);
```

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -574,3 +571,31 @@ impl ViewEye {
         self.center += translate;
     }
 }
+
+// Logic should be similar to `impl TypedComponentFallbackProvider<LinearSpeed> for SpatialView3D`
+impl TypedComponentFallbackProvider<LinearSpeed> for ViewEye {
```
Wumpf, line 555: Interesting idea making the ViewEye the fallback provider, didn't even occur to me but makes sense logically, that way you don't have to pass in the view for using the fallback path. It is a bit weird though because the duplication of those in the view is absolutely not worth it.
It would be better to pass the &dyn fallback provider alongside the property where needed.

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -574,3 +571,31 @@ impl ViewEye {
...
+        let maybe_state = re_viewer_context::ViewStateExt::downcast_ref::<crate::SpatialViewState>(
+            ctx.view_ctx.view_state,
+        );
+        let speed = match maybe_state {
+            Ok(spatial_view_state) => {
+                let bounding_boxes = &spatial_view_state.bounding_boxes;
+                self.fallback_speed_for_mode(bounding_boxes) as f64
+            }
+            Err(view_system_execution_error) => {
+                re_log::error!("Error while downcasting {}", view_system_execution_error);
```
```suggestion
                re_log::error_once!("Error while downcasting {}", view_system_execution_error);
```

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -574,3 +571,31 @@ impl ViewEye {
...
+        };
+        LinearSpeed(re_types::datatypes::Float64(speed))
+    }
+}
+
+impl TypedComponentFallbackProvider<Eye3DKind> for ViewEye {
+    fn fallback_for(&self, _ctx: &re_viewer_context::QueryContext<'_>) -> Eye3DKind {
+        Eye3DKind::default()
+    }
+}
```
Wumpf: just fyi: I think it's fine to keep this here but note that this is technically not needed since iff a default is available on a given component type, it will automatically pick that if there's no fallback provider
(very non-obvious, I know. Need to document this mechanic better)

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -116,10 +116,10 @@ impl View3DState {
         bounding_boxes: &SceneBoundingBoxes,
         space_cameras: &[SpaceCamera3D],
         scene_view_coordinates: Option<ViewCoordinates>,
+        view_ctx: &ViewContext<'_>,
+        eye_property: &ViewProperty,
     ) -> ViewEye {
-        // If the user has not interacted with the eye-camera yet, continue to
-        // interpolate to the new default eye. This gives much better robustness
-        // with scenes that change over time.
+        // Has the user interact with the eye?
```
Wumpf: nothing wrong with the old comment though, no? Or did you find it factually incorrect?

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -443,11 +449,21 @@ impl SpatialView3D {
             return Ok(()); // protect against problems with zero-sized views
         }
 
+        let eye_property = ViewProperty::from_archetype::<Eye3D>(
+            ctx.blueprint_db(),
+            ctx.blueprint_query,
+            query.view_id,
+        );
+        let state1 = state.clone();
```
Wumpf: why is this needed? if it is needs an explanation and better name

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -411,6 +375,19 @@ impl ViewEye {
...
         });
 
+        let kind = eye_property.component_or_fallback::<Eye3DKind>(
+            view_ctx,
+            self,
+            &Eye3D::descriptor_kind(),
+        );
+        match kind {
+            Ok(Eye3DKind::FirstPerson) => self.set_kind(Eye3DKind::FirstPerson),
+            Ok(Eye3DKind::Orbital) => self.set_kind(Eye3DKind::Orbital),
```
```suggestion
            Ok(kind) => self.set_kind(kind),
```

crates/viewer/re_view_spatial/src/view_3d_properties.rs
```diff
@@ -53,5 +53,35 @@ impl TypedComponentFallbackProvider<Plane3D> for SpatialView3D {
...
+    }
+}
+
+impl TypedComponentFallbackProvider<Eye3DKind> for SpatialView3D {
+    fn fallback_for(&self, _ctx: &re_viewer_context::QueryContext<'_>) -> Eye3DKind {
+        Eye3DKind::default()
+    }
+}
+
+re_viewer_context::impl_component_fallback_provider!(SpatialView3D => [BackgroundKind, Color, StrokeWidth, Plane3D, LinearSpeed, Eye3DKind]);
```
Wumpf: as per note on ViewEye, let's not duplicate this code and rather pass in the fallback provider to the ViewEye. Avoiding this kind of oracle code duplication is the very reason those fallback providers are there in the fist place - they provide a mechanism to share fallback code with specifc parts of the code (the camera here) and generic parts (the ui).

Wumpf:
on and would be great if you could update the PR description with the new more limited scope - PR links show up in the changelog, so it's nice when they show the thing that actually landed :)

Wumpf:
Talked a bit with @jleibs about it and we think right now that `eye_controls: EyeControls3D` might be a nicer naming :)

crates/store/re_types/definitions/rerun/blueprint/views/spatial3d.fbs
jccampagne: I'll change it to 

`eye_controls: rerun.blueprint.archetypes.EyeControls3D (order: 3000);`

as suggested below.

crates/viewer/re_view_spatial/src/eye.rs
jccampagne: This is where I may have missed something.
I thought this was needed to set the widget to the right value before any user editing of the value (load a scene, click on the background of the spatial 3d view, and observe the value of the widget for the linear speed; without that, the widget would show the wrong value (zero) and not call the fallback.) This is what I meant in this comment: https://github.com/rerun-io/rerun/pull/10445#issuecomment-3036483068

crates/viewer/re_view_spatial/src/eye.rs
jccampagne: Happy to remove…

jccampagne:
> Talked a bit with @jleibs about it and we think right now that `eye_controls: EyeControls3D` might be a nicer naming :)

Done - see new commits.

jccampagne:
> on and would be great if you could update the PR description with the new more limited scope - PR links show up in the changelog, so it's nice when they show the thing that actually landed :)

I changed the issue link to https://github.com/rerun-io/rerun/issues/6757 (speed).
But didn't find a link for the kind/type of eye (orbital/firstperson).

Wumpf:
I meant actually mostly the description and screenshots itself
We don't have a ticket for the type thing, no need to create one

crates/viewer/re_view_spatial/src/view_3d_properties.rs
jccampagne: Both functions are similar, but not identical: there is an additional level of indirection in the case of the `SpatialView3D` compared to the `ViewEye`: to get from `spatial_view_state` to `eye` in order to call `eye.fallback_speed_for_mode()`.

Maybe I am missing some context, and need to step back a bit; but I could not find an elegant way to make one call the other (avoiding the downcasting and thus doing it twice in the end).

Would you mind detailing what you had in mind…?

Wumpf: I'd either have one forward to the other or just pass in the fallback provider to the camera.. although I guess that might cause a borrow problem 🤔 
I can have a look if you're done with everything else (?) :)

crates/viewer/re_view_spatial/src/ui_3d.rs
jccampagne: The `clone()`? It's because in the next lines, and a bit further down, `state` is mutably-borrowed in the `state.state_3d.update_eye()` call…
True, I didn't spend a lot of time trying to see how this clone could be avoided without rewriting a few more functions signatures and untangle the values…

crates/viewer/re_view_spatial/src/view_3d_properties.rs
jccampagne: sure - have a look if you like.
There's just the `clone` comment above; and just a general question about `f32` and `f64` conversion back and forth (that a brought up a while a go in https://github.com/rerun-io/rerun/pull/10445#issuecomment-3026695604 )

Wumpf:
awesome thanks for all the work you put into this! I did a lil bit of cleanup pass on the docs and added it to the python blueprint sample for 3D views. Once ci goes green I think we're good to merge

crates/viewer/re_view_spatial/src/view_3d_properties.rs
Wumpf: made the forwarding more straight forward, but yeah you were right there's only so much that can be done, I was wrong to call it duplicated code
[`20c4e27` (#10445)](https://github.com/rerun-io/rerun/pull/10445/commits/20c4e27d3d90ecba2038b841c5cb66a12d4906c2)

Wumpf (approved):

Wumpf:
last minute change: when testing it some more I noticed that  `translation speed` is a bit lengthy in the ui. I think calling it just `speed` is fine in that context (I'll rename)
image

jccampagne:
Great! Thanks for the review and changes. It was fun.

PR #10446, opened 2025-06-30
reduce the need for .into on EntityPath in tests

Wumpf:
addressing two annoyances:
* excessive amount of `.into` for `EntityPath` handling
* making it easier to set up single view blueprints

### DIFF
--- a/crates/store/re_log_types/src/path/entity_path.rs
+++ b/crates/store/re_log_types/src/path/entity_path.rs
@@ -491,6 +491,24 @@ impl std::ops::Div<EntityPathPart> for &EntityPath {
     }
 }
 
+impl std::ops::Div<&str> for &EntityPath {
+    type Output = EntityPath;
+
+    #[inline]
+    fn div(self, other: &str) -> Self::Output {
+        self.join(&EntityPath::from(other))
+    }
+}
+
+impl std::ops::Div<&str> for EntityPath {
+    type Output = Self;
+
+    #[inline]
+    fn div(self, other: &str) -> Self::Output {
+        self.join(&Self::from(other))
+    }
+}
+
 // ----------------------------------------------------------------------------
 
 use re_types_core::Loggable;
--- a/crates/store/re_log_types/src/path/entity_path.rs
+++ b/crates/store/re_log_types/src/path/entity_path.rs
@@ -761,7 +761,7 @@ mod tests {
 
     #[test]
     fn test_strip_prefix() {
-        let entity_path = EntityPath::properties() / EntityPathPart::from("episode");
+        let entity_path = EntityPath::properties() / "episode";
         assert_eq!(entity_path.to_string(), "/__properties/episode");
         assert_eq!(
             entity_path.strip_prefix(&EntityPath::properties()),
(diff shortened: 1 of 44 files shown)


crates/store/re_log_types/src/path/entity_path.rs
```diff
@@ -491,6 +491,24 @@ impl std::ops::Div<EntityPathPart> for &EntityPath {
     }
 }
 
+impl std::ops::Div<&str> for &EntityPath {
+    type Output = EntityPath;
+
+    #[inline]
+    fn div(self, other: &str) -> Self::Output {
```
emilk: There is a foot-gun here: `EntityPath::from("/foo/bar") / "mushroom/snake"` -> `/foo/bar/mushroom\/snake`

I suggest we do this:

```suggestion
    fn div(self, other: &'static str) -> Self::Output {
        debug_assert!(!other.contains('/'), "Helpful error message");
```

crates/store/re_log_types/src/path/entity_path.rs
```diff
@@ -491,6 +491,24 @@ impl std::ops::Div<EntityPathPart> for &EntityPath {
...
+        self.join(&EntityPath::from(other))
+    }
+}
+
+impl std::ops::Div<&str> for EntityPath {
+    type Output = Self;
+
+    #[inline]
+    fn div(self, other: &str) -> Self::Output {
```
emilk: Same here

crates/viewer/re_viewer_context/src/test_context.rs
```diff
@@ -344,10 +344,10 @@ impl TestContext {
     /// The provided closure should add content using the [`ChunkBuilder`] passed as argument.
     pub fn log_entity(
         &mut self,
-        entity_path: EntityPath,
+        entity_path: impl Into<EntityPath>,
         build_chunk: impl FnOnce(ChunkBuilder) -> ChunkBuilder,
     ) {
-        let builder = build_chunk(Chunk::builder(entity_path));
+        let builder = build_chunk(Chunk::builder(entity_path.into()));
```
emilk: `Chunk::builder` could also take `impl Into<EntityPath>`

crates/store/re_log_types/src/path/entity_path.rs
```diff
@@ -491,6 +491,24 @@ impl std::ops::Div<EntityPathPart> for &EntityPath {
     }
 }
 
+impl std::ops::Div<&str> for &EntityPath {
```
emilk: You can look for ` / EntityPathPart` in our code to see if there are more places that can be shortened

crates/store/re_log_types/src/path/entity_path.rs
Wumpf: I'm torn whether this is really a bad thing. Almost always that's what I want when concatenating paths like this.
If this was file paths I wouldn't think twice about that behavior

Wumpf: semi-independent of that, why `'static`? Doesn't have to be a string literal, also works fine if I have an existing string?

emilk: Oops - I misread this.

Wumpf: resolved on quick huddle: Emil misread my code and I misread his comment 😅 . This does the right thing and not interpret things as a path part. but let's add a unit test and make it `'static` so this can't be used by accident and cause issues

PR #10448, opened 2025-06-30
Creating custom stream instead of relying on existing partition stream

timsaucer:
### Related

https://linear.app/rerun/issue/DPF-1796/niko-feedback-triage-sandbox-todo

### What

This builds on top of https://github.com/rerun-io/rerun/pull/10442 and switches to a custom stream provider. This stream provider will accumulate record batch rows until it has a reasonable amount and then send a larger batch out to datafusion.

### DIFF
--- a/crates/store/re_datafusion/src/dataframe_query_provider.rs
+++ b/crates/store/re_datafusion/src/dataframe_query_provider.rs
@@ -1,33 +1,63 @@
-use std::collections::{BTreeMap, HashMap};
-use std::sync::Arc;
-
 use arrow::array::{Array, ArrayRef, RecordBatch, StringArray, new_null_array};
+use arrow::compute::SortOptions;
 use arrow::datatypes::{DataType, Field, Schema, SchemaRef};
-use datafusion::{
-    catalog::{TableProvider, streaming::StreamingTable},
-    error::DataFusionError,
-    execution::SendableRecordBatchStream,
-    physical_plan::{stream::RecordBatchStreamAdapter, streaming::PartitionStream},
+use arrow::error::ArrowError;
+use async_trait::async_trait;
+use datafusion::catalog::Session;
+use datafusion::common::{plan_datafusion_err, plan_err};
+use datafusion::datasource::TableType;
+use datafusion::execution::{RecordBatchStream, TaskContext};
+use datafusion::logical_expr::Expr;
+use datafusion::physical_expr::expressions::Column;
+use datafusion::physical_expr::{
+    EquivalenceProperties, LexOrdering, Partitioning, PhysicalExpr, PhysicalSortExpr,
 };
+use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
+use datafusion::physical_plan::{DisplayAs, DisplayFormatType, ExecutionPlan, PlanProperties};
+use datafusion::{
+    catalog::TableProvider, error::DataFusionError, execution::SendableRecordBatchStream,
+};
+use futures_util::Stream;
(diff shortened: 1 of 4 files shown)


crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -181,3 +260,141 @@ pub fn align_record_batch_to_schema(
...
+                    .map(|(index_col, _)| Column::new(index_name, index_col))
+            })
+            .map(|expr| Arc::new(expr) as Arc<dyn PhysicalExpr>);
+        let orderings = match order_col {
+            Some(col_expr) => vec![LexOrdering::new(vec![PhysicalSortExpr::new(
+                col_expr,
+                SortOptions::new(false, true),
+            )])],
+            None => vec![],
+        };
```
abey79, line 254: @jleibs would you expect that `rerun:is_sorted` metadata should propagate all the way to here?

jleibs: We have `query_engine`s here which contain original rerun chunks -- those should have valid `rerun:is_sorted` metadata on them. However, when executing the query via the engine I believe it will re-sort the chunks on the fly as necessary.

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -1,29 +1,64 @@
...
+/// rows with 32b of data. We are setting this lower as a reasonable first guess to avoid
+/// the pitfall of executing a single row at a time, but we will likely want to consider
+/// at some point moving to a dynamic sizing.
+const DEFAULT_BATCH_SIZE: usize = 2048;
 
-#[derive(Debug)]
 pub struct DataframeQueryTableProvider {
     pub schema: SchemaRef,
-    partition_streams: Vec<Arc<DataframePartitionStream>>,
+    // partition_streams: Vec<Arc<DataframePartitionStream>>,
```
teh-cmc:
```suggestion
```

crates/store/re_datafusion/src/dataframe_query_provider.rs
teh-cmc: :point_up_2:

abey79: If we resort things, would it make sense to then tag it with `rerun:is_sorted` and let datafusion know about it?

abey79: Or do we never actually know of anything being _globally_ sorted but the query index?

teh-cmc: I didn't quite understand all the questions but:

> If we resort things, would it make sense to then tag it with `rerun:is_sorted` and let datafusion know about it?

* `re_dataframe` gets its chunks from the query cache, which sorts them just-in-time
* once a chunk is sorted, its `rerun:is_sorted` flag is set, always
* I'm not sure what it means to "let datafusion know about it"

> Or do we never actually know of anything being _globally_ sorted but the query index?

Not sure what that means. Chunks are never globally sorted since we allow interval overlaps. The output of the `re_dataframe` is fully sorted though, because we shuffle the individual rows just-in-time, accounting for overlaps.

timsaucer:
Marking as do-not-merge for the moment just until I can get some more testing time with the batch size and larger dataframes like images or video blobs

zehiko (approved):
(I'm late to the party)

leveraging ``CoalesceBatchesExec`` was a really nice simplification.

Issue #10451, opened 2025-07-01
"<entity> has no components on timeline" when sub-entities have components


nikolausWest:
<img width="813" alt="Image" src="https://github.com/user-attachments/assets/d906b3f0-fea9-4a6c-9509-af9c0593e932" />

This makes things feels broken. Perhaps we could refrain from showing this message when sub-entites have data on the timeline and the tree is collapsed?

PR #10452, opened 2025-07-01
Adjust wording when hovering empty parent entity in timeline

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/10451

### What

ticket suggested dropping the message, but given that hovering those dots usually shows something I decided to refrain from it and instead change the wording:

image

### DIFF
--- a/crates/viewer/re_data_ui/src/instance_path.rs
+++ b/crates/viewer/re_data_ui/src/instance_path.rs
@@ -59,7 +59,7 @@ impl DataUi for InstancePath {
             ui_layout.label(
                 ui,
                 format!(
-                    "only subpaths of {self} have components on timeline {:?}",
+                    "{self} has no own components on timeline {:?}, but its children do",
                     query.timeline()
                 ),
             );


crates/viewer/re_data_ui/src/instance_path.rs
```diff
@@ -59,7 +59,7 @@ impl DataUi for InstancePath {
             ui_layout.label(
                 ui,
                 format!(
-                    "{self} has no components on timeline {:?}",
+                    "only subpaths of {self} have components on timeline {:?}",
```
emilk:
```suggestion
                    "Only subpaths of {self} have components on timeline {:?}",
```

https://github.com/rerun-io/rerun/blob/main/DESIGN.md#casing

emilk: Or maybe

```suggestion
                    "{self} has no own components on timeline {:?}, but its children might",
```

Wumpf: I think it's not a `might` in this case, otherwise there wouldn't be anything to hover

PR #10464, opened 2025-07-01
Add a benchmark for logging many `Scalars`

emilk:


Wumpf:
seems somewhat redundant to `plot_dashboard_stress`. Don't mind that though since it's built differently, but worth linking to that

Wumpf (approved):
lgtm!

PR #10469, opened 2025-07-01
Automate `check_blueprint_overrides` with image comparision test 

Wumpf:
one manual test item less at a time!


Wumpf:
because it is! 😁

PR #10472, opened 2025-07-01
`DatasetEntry.register` now returns the created partition id

abey79:
Title. Useful for `dataset_util.py` when registering/updating blueprints.


rerun_py/src/catalog/connection_handle.rs
```diff
@@ -187,7 +187,7 @@ impl ConnectionHandle {
         py: Python<'_>,
         dataset_id: EntryId,
         recording_uris: Vec<String>,
-    ) -> PyResult<Vec<TaskId>> {
+    ) -> PyResult<Vec<(TaskId, PartitionId)>> {
```
jleibs: `Vec<(TaskId, PartitionId)>` doesn't feel great, esp given the need to do:
```
results.into_iter().map(|(task_id, _)| task_id)
```
below.

Should we just pass through the raw Dataframe instead of restructuring here?

Alternatively some kind of `TaskResult` struct would at least be more explicit.

abey79: I think the right thing to do here is to use an explicit grpc message instead of a dataframe. That's basically what you propose with `TaskResults` (which technically is `RegisterWithDatasetResult` in this case).

I dislike the idea of returning a dataframe, because it means duplicating the annoying column extraction code.

As a stop gap, I'll add a "fake" `RegisterWIthDatasetResult` in `re_proto` that's deserialised from the dataframe.

abey79: I completely reworked this by:
- Moving `register_with_dataset()` to `ConnectionClient`
- Introducing `RegisterWithDatasetTaskDescriptor` (which I strongly advocate for being an actual grpc message)


Tested by registering `droid:sample500` locally.

PR #10473, opened 2025-07-01
Add --connect documentation to the CLI reference

benliepert:
### What

This is a small change so instead of creating an issue and doing some back and forth, you all can decide if it's necessary 😁 

#7149 added the CLI manual and I didn't see why `--connect` would have been intentionally omitted.
So I added its documentation from the [entrypoint](https://github.com/rerun-io/rerun/blob/1438dd212900b95a40d8d7d585f127dd11e4fb2c/crates/top/rerun/src/commands/entrypoint.rs#L189) as other options here look to have done.


Wumpf (approved):
good catch thanks!

There's actually more changes for 0.24 that should be in here already. But this applies to 0.23 to I'll tell the machinery to backport this to the 0.23/current docs!

PR #10476, opened 2025-07-02
Populate fuzzy descriptors

jleibs:
Without fuzzy descriptors populated, we always include every component for an entity path.

### DIFF
--- a/rerun_py/src/catalog/connection_handle.rs
+++ b/rerun_py/src/catalog/connection_handle.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashSet};
 
 use arrow::array::{RecordBatch, RecordBatchIterator, RecordBatchReader};
 use arrow::datatypes::Schema as ArrowSchema;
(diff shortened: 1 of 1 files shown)


crates/store/re_chunk_store/src/dataframe.rs
```diff
@@ -314,7 +314,7 @@ impl QueryExpression {
             return Some(RangeQuery::new(index, *filtered_index_range));
         }
 
-        None
```
jleibs: @teh-cmc I'm not totally confident this is the right thing to do, but without it we would never send the fuzzy descriptors in the first place since the None would mean we don't even output a range-query:
https://github.com/rerun-io/rerun/blob/446854fccb838da717ab23ce8bf7f3423beba5a4/rerun_py/src/catalog/connection_handle.rs#L699-L700

teh-cmc: Given the current implementation it's the only thing you can do, but that would be very unfortunate since `None` has no compute cost while `Some(-inf, +inf)` does.

Let me see if I can get the `fuzzy_descriptors` out of the `Query` layer, they really have nothing to do there.

teh-cmc:
* Okay I'll rebase this on top of https://github.com/rerun-io/rerun/pull/10482 now

crates/store/re_chunk_store/src/dataframe.rs
teh-cmc: That's one hack gone :lotus_position:

teh-cmc:
(Drafted because this doesn't fix anything anymore at the moment due to server changes... wait for it!)

teh-cmc:
* This now works properly in all cases (temporal-only descriptor and static-only descriptor) when running with https://github.com/rerun-io/dataplatform/pull/1131

teh-cmc (approved):

PR #10477, opened 2025-07-02
Reduce log spam from ffmpeg

emilk:
We get a lot of ffmpeg errors that are duplicated, safe for a single hex-value

> [vist#0:0/h264 @ 0x143006b60] [dec:h264 @ 0x143007a90] [error] Decoding error: Invalid data found when processing input
> [vist#0:0/h264 @ 0x143007a90] [dec:h264 @ 0x1733ab230] [error] Decoding error: Invalid data found when processing input

We already had code to remove some of these pointer values, but now that code has been improved.


Wumpf (approved):
neat, much more version robust

PR #10479, opened 2025-07-02
Show average FPS of selected video

emilk:
image

We're hedging here by calling it the _average_ FPS. In case of variable frame rate videos, this is still true.

Most videos has a fixed framerate anyhow, and so knowing its FPS is very useful.

### DIFF
--- a/crates/utils/re_video/src/demux/mod.rs
+++ b/crates/utils/re_video/src/demux/mod.rs
@@ -434,6 +434,8 @@ impl VideoDataDescription {
 
     /// Length of the video if known.
     ///
+    /// NOTE: This includes the duration of the final frame too!
+    ///
     /// For video streams (as opposed to video files) this is generally unknown.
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
lgtm!

can land as is but we could do even better and check the mp4's fps and VFR (variable frame rate) properties if present. Iirc this is all on the SPS for h264 & h265... but we inevitably get into waaaay more complex code :/

An interesting aspect of this is that streamed video may specify fps in the SPS but then not uphold it due to packet loss & varying details etc.

.. makes me all feel like this is way too simplistic, but then again it's useful metric as is I figure?

Wumpf:
taking a step back my "UHM AKTUALLY 🤓" stance here is slowly becoming ridiculous. None of the words I utter there make sense to the average video user but everyone knows what a framerate is and would expect that information. 
shipshipship

Issue #10475, opened 2025-07-01
Video hover flickers terribly on hover


jleibs:
**Describe the bug**
During mouse-over of a video element we get a terrible flicker

**To Reproduce**
Open droid dataset, mouse over the video

**Screenshots**

https://github.com/user-attachments/assets/2d59d125-8beb-406b-b339-0287cc0d9a70

**Desktop (please complete the following information):**
 - OS: Linux on Web

**Rerun version**
`dd70aa7`

**Additional context**

Native works fine

emilk:
Ooh it's this bug again!
* https://github.com/rerun-io/rerun/issues/9969

emilk:
Maybe the logic for the tooltip should be to keep the pivot from the previous frame if it still works 🤔

lucasmerlin:
Here I understand why it happens, the color keeps changing erratically, changing the popups size, making it fit or not fit so it switches back and forth like crazy. I don't think the Popup needs to handle an area that changes size erratically each frame correctly. We should instead investigate why the color sometimes shows `No value` and always allocate enough space so the full color value fits in the popup.

Wumpf:
There might yet be a video specific problem here: we read the color dynamically out of the gpu texture and if you step Jeremy's video you see that we occastionally get `no value`
All while this is a still frame

Wumpf:
`OS: Linux on Web` probably means webgl

I'll look into the no-value part of the issue!

Wumpf:
For reference: no value
![Image](https://github.com/user-attachments/assets/437fa5f8-5ccf-42ea-9115-16bbc3383567)

Wumpf:
confirmed that it happens only with WebGL

Wumpf:
Fixed by combination of
*  #10481
*  #10483

PR #10481, opened 2025-07-02
Fix pixel picking ui width changing rapidly by padding values

lucasmerlin:
### Related

* related https://github.com/rerun-io/rerun/issues/10475

### What

This ensures that the color value label always has the same width by padding the values with leading spaces. This should also improve ux by making changing color values more readable. The values will now also be properly formatted using re_format.

I first had a more complicated fix by calculating the max width the label would have and always allocating at least this much space, even made a nice helper for this, but I think this is the better solution.

To properly fix the issue we also need the fix that some frames don't receive color values sometimes, @Wumpf is working on that.

https://github.com/user-attachments/assets/6f8f8881-51fa-4399-b549-1260f51f6b80

### DIFF
--- a/crates/store/re_types/Cargo.toml
+++ b/crates/store/re_types/Cargo.toml
@@ -101,8 +101,8 @@ serde = { workspace = true, optional = true, features = ["derive", "rc"] }
 glam.workspace = true
 itertools.workspace = true
 mint.workspace = true
-similar-asserts.workspace = true
 rand = { workspace = true, features = ["std", "std_rng"] }
+similar-asserts.workspace = true
 
 [build-dependencies]
 


crates/store/re_types/src/tensor_data.rs
```diff
@@ -353,6 +353,72 @@ impl TensorElement {
...
+        let min_val_len = self.min_value().format().len();
+        let max_val_len = self.max_value().format().len();
+        let value_str = self.format();
+        if value_str.len() < min_val_len || value_str.len() < max_val_len {
+            let max_len = usize::max(min_val_len, max_val_len);
+            format!("{value_str:>max_len$}")
+        } else {
+            value_str
+        }
+    }
```
lucasmerlin: I'm assuming that f32/f64 :: MAX/MIN is the logest value there can be with re_format, is that correct?

emilk: Not sure about that 🤔  something like `-1.0/3.0` could be pretty long too

crates/store/re_types/src/tensor_data.rs
```diff
@@ -353,6 +353,74 @@ impl TensorElement {
...
+
+    /// Formats the element as a string, padded to the width of the largest possible value.
+    pub fn format_padded(&self) -> String {
+        let min_val_len = self.min_value().format().len();
+        let max_val_len = self.max_value().format().len();
+        let value_str = self.format();
+        if value_str.len() < min_val_len || value_str.len() < max_val_len {
+            let max_len = usize::max(min_val_len, max_val_len);
```
emilk:
```suggestion
        let min_val_len = self.min_value().format().len();
        let max_val_len = self.max_value().format().len();
        let max_len = usize::max(min_val_len, max_val_len); // likely at least, e.g. 0/255 for u8. Maybe off for floats.
        let value_str = self.format();
        if value_str.len() < max_len {
```

crates/store/re_types/src/tensor_data.rs
lucasmerlin: Wrote some tests and it's way off 😅 Seems like the longest values can be 22chars long

crates/viewer/re_view_spatial/tests/snapshots/annotations_hover_rect_red.png
lucasmerlin, line 1: This looks a bit awkward now since the number is basically right aligned. I also tried right aligning all the values in the grid, but then it would somehow trim the first space which brought back the flicker, but even worse since everything is right aligned now. But I guess it's not worse than before.

crates/store/re_types/Cargo.toml
```diff
@@ -102,7 +102,7 @@ glam.workspace = true
 itertools.workspace = true
 mint.workspace = true
 similar-asserts.workspace = true
-
+rand = { workspace = true, features = ["std", "std_rng"] }
```
emilk, line 104: nit: sort

PR #10489, opened 2025-07-02
Try fixing webgl spurious stuckage

Wumpf:
Trying out 
* https://github.com/emilk/egui/pull/7291


Wumpf:
it worked!! Let's do a proper egui update now that it's landed to egui's main

Issue #10427, opened 2025-06-27
Attempt to sort by non-sorbet columns Fails


jleibs:
Any column missing rerun-metadata doesn't round-trip properly. Demonstrated here with link-name, but we see the same problem with partition_id.

<img src="https://uploads.linear.app/93919c5e-abe4-4cd5-b77d-7e3eafe3e62d/32dd6b1a-9b8f-4559-876b-b27813454e2b/033d2a58-247f-44a6-97c2-1ec40f92b61f?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzkzOTE5YzVlLWFiZTQtNGNkNS1iNzdkLTdlM2VhZmUzZTYyZC8zMmRkNmIxYS05YjhmLTQ1NTktODc2Yi1iMjc4MTM0NTRlMmIvMDMzZDJhNTgtMjQ3Zi00NGE2LTk3YzItMWVjNDBmOTJiNjFmIiwiaWF0IjoxNzUxMDU1NDE1LCJleHAiOjMzMzIxNjE1NDE1fQ._PVY-45Y7C1qleBH08ocIjNLvkZU9GnoPVu7N8gGvJQ " alt="image.png" width="742" data-linear-height="406" />

<img src="https://uploads.linear.app/93919c5e-abe4-4cd5-b77d-7e3eafe3e62d/6d3af43f-ad90-412d-b910-c8b24ac71003/2d7cee54-538f-4f30-b7a3-f130b8194a55?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzkzOTE5YzVlLWFiZTQtNGNkNS1iNzdkLTdlM2VhZmUzZTYyZC82ZDNhZjQzZi1hZDkwLTQxMmQtYjkxMC1jOGIyNGFjNzEwMDMvMmQ3Y2VlNTQtNTM4Zi00ZjMwLWI3YTMtZjEzMGI4MTk0YTU1IiwiaWF0IjoxNzUxMDU1NDE1LCJleHAiOjMzMzIxNjE1NDE1fQ.cGela-TjLvBGTMIH42BKrRslO3UcpFoI7lR7fOEC368 " alt="image.png" width="1692" data-linear-height="360" />

Issue #10421, opened 2025-06-27
Inconsistent definition of a Sorbet batch


grtlr:
### Problem

*(with quite some help from Claude)*

The Rerun viewer currently requires `sorbet:version` metadata for all record batches, but this conflicts with its stated goal of handling arbitrary record batches. The DataFusion widget implementation assumes every record batch to  SorbetBatch compatibility, creating errors when processing general record batches lacking Sorbet metadata.

This became apparent via:

* #10417 and the corresponding "fix:
* #10418

We should get rid of the inconsistencies and decide what exactly constitutes a SorbetBatch and a ChunkBatch.

**Core Tension:**

* **Intended behavior**: Viewer should accept mixed Sorbet/non-Sorbet data
* **Current implementation**: Strict Sorbet requirements cause migration errors
* **Architectural assumption**: SorbetBatch was designed as lenient "RecordBatch-with-optional-benefits"

**Key Design Questions:**

1. Should `sorbet:version` be batch-level or field-level metadata?
2. How strict should SorbetBatch validation be vs. ChunkBatch?
3. Should migration logic apply to component descriptors in mixed-metadata scenarios?

**Alternative Framings:**

* **Metadata granularity**: Move from batch-level to column-level Sorbet versioning for true mixed-mode support
* **Batch typing**: Establish clear SorbetBatch vs. general RecordBatch handling paths
* **Migration strategy**: Implement best-effort, column-by-column migration rather than batch-level requirements

**Failure Modes:**

* Customer datasets merging different Sorbet versions into single tables
* Performance degradation from redundant validation
* Silent data corruption if migration logic becomes too permissive

**Trade-offs:** Relaxing requirements improves compatibility but may mask legitimate data quality issues. Maintaining strict validation ensures data integrity but limits viewer utility for heterogeneous datasets.

abey79:
With [this recent regression](https://github.com/rerun-io/rerun/pull/10365#issuecomment-3028018582) in mind...

My high-level view of the situation is that `SorbetBatch` is trying to do way too many things at the same time, including:
1) An important cog in the `RecordBatch` <-> `Chunk` serialisation and deserialisation paths, recently adding migration support.
2) A wrapper over user data that applies sorbet semantics on the (very much optional) metadata.

This conflicts in many ways, including:
- (1) needs to fail on error, (2) must only fail in presence of wrong rerun-specific metadata.
- (1) is faithful to the underlying chunk data semantics, (2) must never temper with user data
[...]

PR #10491, opened 2025-07-02
Source physical column names from the original `RecordBatch`es

abey79:
### Related

* closes https://github.com/rerun-io/rerun/issues/10427
* fix bug introduced by https://github.com/rerun-io/rerun/pull/10365
* detailed explanation https://github.com/rerun-io/rerun/pull/10365#issuecomment-3028018582
* related https://github.com/rerun-io/rerun/issues/10421

### What

The `DatafusionTableWidget` needs the physical names of the underlying columns in order to emit column references that DataFusion will understand. Previously, we'd extract `arrow::datatypes::Fields` from the `SorbetBatch`es. However, since #10365, the `Fields` are tempered with in the migration process, making them no longer suitable to extract column physical names.

As a work-around, this PR sources the `Fields` from the original `RecordBatch`es, _before_ conversion to `SorbetBatch`es.

### DIFF
--- a/crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
+++ b/crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
@@ -1,15 +1,18 @@
-use crate::RequestedObject;
-use crate::table_blueprint::{EntryLinksSpec, PartitionLinksSpec, SortBy, TableBlueprint};
-use arrow::datatypes::DataType;
+use std::sync::Arc;
+
+use arrow::datatypes::{DataType, Fields};
 use datafusion::common::{DataFusionError, TableReference};
 use datafusion::functions::expr_fn::concat;
 use datafusion::logical_expr::{col as datafusion_col, lit};
 use datafusion::prelude::{SessionContext, cast, encode};
 use parking_lot::Mutex;
+
 use re_log_types::Timestamp;
 use re_sorbet::{BatchType, SorbetBatch};
 use re_viewer_context::AsyncRuntimeHandle;
-use std::sync::Arc;
+
+use crate::RequestedObject;
+use crate::table_blueprint::{EntryLinksSpec, PartitionLinksSpec, SortBy, TableBlueprint};
 
 /// Make sure we escape column names correctly for datafusion.
 ///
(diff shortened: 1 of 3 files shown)


crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -127,13 +131,21 @@ impl DataFusionQuery {
...
+                col(&sort_by.column_physical_name).sort(sort_by.direction.is_ascending(), true),
             ])?;
         }
 
         // collect
         let record_batches = dataframe.collect().await?;
 
+        // IMPORTANT: fields must be copied here *before* converting to `SorbetBatch`, because that
+        // conversion modifies the field names. As a result, the schema contained in a `SorbetBatch`
+        // cannot be used to derive the physical column names as seen by DataFusion.
```
grtlr:
```suggestion
        // TODO(#10421) IMPORTANT: fields must be copied here *before* converting to `SorbetBatch`, because that
        // conversion modifies the field names. As a result, the schema contained in a `SorbetBatch`
        // cannot be used to derive the physical column names as seen by DataFusion.
```

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -143,7 +155,7 @@ impl DataFusionQuery {
             .collect::<Result<Vec<_>, _>>()
             .map_err(|err| DataFusionError::External(err.into()))?;
 
-        Ok(sorbet_batches)
+        Ok((sorbet_batches, fields))
```
jleibs, line 158: The fact that these must be in sync is a risk but seems reasonable. How does our column-reordering work? Any chance re-ordering / hiding columns causes confusion here?

jleibs (approved):
This look

grtlr (approved):
Thanks @abey79—and sorry for the regression!

:white_checkmark: Tested on native viewer.

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
grtlr: I tried hiding and moving columns and couldn't see any problems. Reloading the table via the bottom bar removes the sorting, but I guess that's expected?

abey79: Yeah we had column index tracking already. This PR doesn't introduce a change of that at all. It basically just source the `Fields` from somewhere else.

PR #10506, opened 2025-07-03
Python SDK: split `dataframe.rs` in multiple files in 

abey79:
### Related

* prep for https://github.com/rerun-io/rerun/issues/10193

### What

Title. Pure refactor, moving things around.


rerun_py/src/dataframe/mod.rs
```diff
@@ -0,0 +1,46 @@
+#![allow(clippy::borrow_deref_ref)] // False positive due to #[pyfunction] macro
+#![allow(clippy::needless_pass_by_value)] // A lot of arguments to #[pyfunction] need to be by value
+#![allow(deprecated)] // False positive due to macro
+#![allow(unsafe_op_in_unsafe_fn)] // False positive due to #[pyfunction] macro
```
Wumpf: let's make those `expect`, lest they stay forever. Already skeptical whether all of those are needed?

Wumpf (approved):
More files and less `pub(crate)` how could I not approve of this 😄 

blindly trusting that this only splits things ;-)

rerun_py/src/dataframe/mod.rs
abey79: good call, I could 🥌 two and move another closer to the source

Issue #10512, opened 2025-07-04
Consider logging a "marker" component for empty archetypes


grtlr:
"Marker" archetypes such as `SeriesPoints` and `SeriesLines` without any values set (for example when using `Self::new()`) will be empty and therefore not show up in the viewer unless specifically mentioned.

In these cases, it might make sense to log a "marker" component to represent that a component was logged. This component could then also be used for those archetypes in the migration logic.

This is a bit of a controversial topic so I'm cc'ing: @andreas and @jleibs (how came up with the idea).

emilk:
Did you come up with a plan for this one?

grtlr:
Yeah, we are going to do

* #10572 

instead for now. I'm removing the issue from the milestone.

grtlr:
We have been fine without markers, which is why I'm closing this issue.

ntjohnson1:
There's a test that refers to this that is now a zombie todo.

grtlr:
@copilot Remove the zombie TODO please.

grtlr:
@claude Remove the zombie TODO please.

grtlr:
_"No one likes to work anymore these days!"_ 😉

* #12585

PR #10513, opened 2025-07-04
Switch from `indicators` to `relevant_archetypes`

grtlr:
### Related

* Part of #6889.
* Part of #8129.
* Potential follow-up: #10512

### What

This PR changes the way entities are "indicated" from looking at `indicators` to looking at the `relevant_archetypes` names. This change also surface one change in our behavior:

"Marker" archetypes such as `SeriesPoints` and `SeriesLines` without any values set (for example when using `Self::new()`) will be empty and therefore not show up in the viewer unless specifically mentioned. For now there are several workarounds:
* explicitly set a field
* create a visualizer override in a blueprint
* implement "marker" components: #10512

### DIFF
--- a/crates/viewer/re_viewer_context/src/typed_entity_collections.rs
+++ b/crates/viewer/re_viewer_context/src/typed_entity_collections.rs
@@ -24,6 +24,7 @@ impl std::ops::Deref for MaybeVisualizableEntities {
     }
 }
 
+// TODO(#8129): Rename to `RelevantEntities`?
 /// List of entities that match the indicator components of a visualizer.
 ///
 /// In order to be a match the entity must have at some point in time on any timeline had any of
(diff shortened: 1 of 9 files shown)


crates/viewer/re_viewer_context/src/typed_entity_collections.rs
```diff
@@ -24,6 +24,7 @@ impl std::ops::Deref for MaybeVisualizableEntities {
     }
 }
 
+// TODO(#8129): Rename to `RelevantEntities`?
```
Wumpf: at the very least the doc already needs updating imho - it still talks about indicator components but they are no longer what drives this type!

crates/viewer/re_viewer_context/src/typed_entity_collections.rs
```diff
@@ -24,11 +24,10 @@ impl std::ops::Deref for MaybeVisualizableEntities {
     }
 }
 
-// TODO(#8129): Rename to `RelevantEntities`?
-/// List of entities that match the indicator components of a visualizer.
+/// List of entities that contain contain archetypes that are relvant for a visualizer.
```
Wumpf: the the entities contain contain relvant!

PR #10514, opened 2025-07-04
Fix: don't always pick `/` as the origin when creating new views

emilk:
### What
When creating a new view from a selection of entities (e.g. by right-clicking them and selecting "Add to new view") the origin would always be set to `/` (except for 2D and 3D views). This was quite confusing, at it also meant the new view would be given the name `/`.

Review commit-by-commit!
The first commit is the fix, the second just a naming change.

image


Wumpf (approved):
nice catch, thanks to k-scale for pointing out this issue! :)

sidenote: interesting how everything acts like the "origin" is the only way to determine the contents of a view whereas it's just a shortcut as long as the default contents of `$origin/**` are used. Seen users being confused by this. The renames here make me wonder again if we could do better

PR #10515, opened 2025-07-04
Make `DisplayRB` available to all crates via `re_format_arrow`

grtlr:
### What

Title. I think we should use this unified and in more places together with `insta::assert_snapshot`.


grtlr:
Took a different approach

Issue #10495, opened 2025-07-02
Bring React package to feature parity with JS package


jprochazk:
Right now, the only thing the React package can do is load up an RRD file. We should add support for more of the JS package features, such as callbacks and recording/app ID shenanigans. Potentially also things like setting time and streaming data into it, but those seem more complex.

The React component as to be resistant to re-renders, meaning the running Viewer should not be restarted each time the component is re-rendered. That's already the case, but it requires some careful management of state inside the component, and manually diffing props.

Wumpf:
@jprochazk are you planning more to do on this front for 0.24?
Unlike there's someone depending on this who I'm not aware of think we can mark this as nice-to-have and remove it from the milestone (please re-add otherwise)

jprochazk:
Adding callbacks got us most of the way there, and it's not clear there's value in trying to do the rest of this right now. It's at 100% feature parity, but I'll close this for now.

PR #10518, opened 2025-07-04
Support callbacks in React

jprochazk:
* Part of https://github.com/rerun-io/rerun/issues/10495

```
pixi run js-build-all
pixi run yarn --cwd rerun_js workspaces run pack
```
Produces `rerun_js/web-viewer/web-viewer.tar.gz` and `rerun_js/web-viewer-react/web-viewer-react.tar.gz`

These can be installed into e.g. https://github.com/rerun-io/web-viewer-react-example by running:
```
npm i # base dependencies
npm i ../rerun/rerun_js/web-viewer/web-viewer.tar.gz
npm i ../rerun/rerun_js/web-viewer-react/web-viewer-react.tar.gz
```

Afterwards run the example as normal. You can find the new callback props available on the `WebViewer` component, e.g. `onPlay`, `onTimeUpdate`, etc.

## Implementation details

I felt bad about adding another layer of stuff that anyone working on events has to deal with, so this requires no effort whatsoever to maintain.

The callback prop types are generated automatically from `@rerun-io/web-viewer` event types. When a callback prop is registered, the prop name is transformed to the event name (remove `on` prefix, snake_case it), which is then directly registered on the underlying `WebViewer` instance. This way we don't need to maintain a list of events on the React side, it stays up to date automatically as new events are added in the non-React package.

### DIFF
--- a/rerun_js/web-viewer-react/index.js
+++ b/rerun_js/web-viewer-react/index.js
@@ -15,6 +15,7 @@ import * as rerun from "@rerun-io/web-viewer";
  * @typedef {(
  *   Omit<import("@rerun-io/web-viewer").WebViewerOptions, "allow_fullscreen" | "enable_history">
  *   & BaseProps
+ *   & import("./types.d.ts").ViewerEvents
  * )} Props
  */
 
@@ -43,7 +44,7 @@ export default class WebViewer extends React.Component {
     startViewer(
       this.#handle,
       /** @type {HTMLDivElement} */ (this.#parent.current),
-      this.props,
+      () => this.props,
     );
   }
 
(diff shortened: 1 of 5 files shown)


jprochazk:
I thought there was some kind of performance regression in the React package, but I probably just built it in debug mode :grimacing:. Can't repro anymore, so this is ready

rerun_js/web-viewer-react/index.js
```diff
@@ -108,6 +130,17 @@ function startViewer(handle, parent, props) {
     width: "100%",
     height: "100%",
   });
+
+  for (const key of Object.keys(props)) {
+    if (key.startsWith("on")) {
+      /** @type {any} */
+      const event = pascalToSnake(key.slice(2));
+      console.log(key, event);
```
Wumpf: isn't that a bit spammy?

jprochazk: whoops, leftover from dev

jprochazk:
> this prop name to event name transformation could be a bit brittle though, no?

Yes, should add a note about making sure the event names are `snake_case`d versions of the same events on the Rust side

Issue #10516, opened 2025-07-04
`TextEdit` for `token` invisible in light mode


emilk:
I'm not sure why "Host name" is visible, but "Token" is not O.o

![Image](https://github.com/user-attachments/assets/9094e6a0-2e86-4875-9647-3b81146e38c6)

PR #10519, opened 2025-07-04
Add contrast between TextEdit and panels in Light Mode

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/10516

I also changed the default panel background from being pure white, to being _almost_ pure white. I suspect this was a mistake on my part at some point, as pure whites and pure blacks are usually reserved for center-stage things, and not panel backgrounds. But Katya is gonna take a big pass over all these colors before 0.25 anyways.

### DIFF
--- a/crates/viewer/re_ui/data/light_theme.ron
+++ b/crates/viewer/re_ui/data/light_theme.ron
@@ -92,7 +92,7 @@
     },
 
     "panel_bg_color": {
-      "color": "{Gray.950}"
+      "color": "{Gray.975}"
     },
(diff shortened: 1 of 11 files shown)


oxkitsune (commented):
With this change I feel like normal text is a bit harder to read due to the reduced contrast.


2025-07-04_14-27-12@2x

2025-07-04_14-26-34@2x

emilk:
Maybe I'll make it a bit brighter again then

oxkitsune (approved):
Perfect!

PR #10520, opened 2025-07-04
Add tab-bar button to hide a view

emilk:
I find myself often wanting to hide a specific view or two, and this change makes it easier.

https://github.com/user-attachments/assets/a56fc8f8-c4b3-41f6-ae94-c2ef7830e57d

### Why does the hide-button belong here?

I think it belongs there for the same reason the Maximize-button belongs there: it controls the contents of the viewport.
Same as each application window in Mac or Windows has both a maximize and minimize (=hide) button.

We could hide it behind a `…` menu at some point, if we get too many buttons there.


emilk:
I should use this icon instead:
Screenshot 2025-07-04 at 14 34 41

emilk:
Remove is too hard to undo. Un-hide is just a click away.

PR #10521, opened 2025-07-04
Stop logging indicators and drop them during migration

grtlr:
### Related

* Part of #8129
* Part of #6889
* Requires #10530

> [!IMPORTANT]
> This PR does not remove all occurrences of indicators from our code base, but I want to keep PRs small. Because of this I'm not closing #8129 yet.

### What

This PR does several things, it:
* stops indicator from being logged in all SDKs
* drops indicators (and columns with `rerun:is_indicator`) during migration
* adds a note to the `multiple_scalars` snippet 
  * ⚠️ this is a change that is breaking BW-compatibility
  * Follow-up issue: #10512
* adds a `rerun rrd compare --ignore-chunks-without-components` option for snippet comparison.
  * Please review, or better test carefully!

### TODO

* [x] carefully think through sandbox implications
  * For now, we don't handle any logic changes for indicators, this will be done in a follow up PR.
* [x] dataplatform companion PR
* [x] full-check (unrelated error via @Wumpf)


crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -122,6 +123,7 @@ pub fn migrate_record_batch(mut batch: RecordBatch) -> RecordBatch {
                     re_log::trace!("Performing migrations...");
                     batch = maybe_apply::<v0_0_1__to__v0_0_2::Migration>(&batch_version, batch);
                     batch = maybe_apply::<v0_0_2__to__v0_1_0::Migration>(&batch_version, batch);
+                    batch = maybe_apply::<v0_1_0__to__v0_1_1::Migration>(&batch_version, batch);
```
Wumpf, line 126: so is that an intra 0.24alpha port then? doesn't seem to hurt, indeed very nice to have these smaller, but I figure but maybe worth pointing out somewhere (unless I missunderstand)

crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -2575,10 +2575,26 @@ impl RecordingStream {
 
 #[cfg(test)]
 mod tests {
+    use insta::assert_debug_snapshot;
     use re_log_types::example_components::{MyLabel, MyPoints};
 
     use super::*;
 
+    struct DisplayDescrs(Chunk);
```
Wumpf, line 2583: nice, cool construct for snapshot testing. Wasn't familiar with that pattern

Wumpf (approved):
oh-my-god-its-happening

awesome, and it's just the beginning 😄

crates/store/re_sorbet/src/migrations/mod.rs
grtlr: Yeah, we have substantial amount of data out there that still uses indicators but is on `0.1.0`. Without the bump we could not migrate it with our current logic.

grtlr:
@rerun-bot full-check

grtlr:
@rerun-bot full-check

PR #10524, opened 2025-07-04
Use built-in egui_kittest broken-pixel treshold utilities instead of custom extensions

Wumpf:
As well as:
* fix a transform test that ignored everything but the last result 😱 
* adjust treshholds to pass tests on my windows machine where images look correct as-is
* update images that should have failed already but only did so on my machine (they just about passed it seems)😬


crates/viewer/re_blueprint_tree/tests/range_selection_test.rs
```diff
@@ -87,7 +87,14 @@ fn test_range_selection_in_blueprint_tree() {
 
     harness.run();
 
-    harness.snapshot("range_selection_in_blueprint_tree");
+    harness.snapshot_options(
+        "range_selection_in_blueprint_tree",
+        // @wumpf's Windows machine needs a bit of a higher threshold to pass this test due to discrepancies in text rendering.
+        // (Software Rasterizer on CI seems fine with the default).
+        &SnapshotOptions::new()
+            .threshold(OsThreshold::new(SnapshotOptions::default().threshold).windows(0.8))
```
emilk, line 95: Seeing this I think we should change `OsThreshold` to hold `Option`:s instead, so that the above is equivalent to 

```suggestion
            .threshold(OsThreshold::default().windows(0.8))
```

emilk (approved):
Nice

crates/viewer/re_blueprint_tree/tests/range_selection_test.rs
Wumpf: agreed, that would be much nicer!

PR #10526, opened 2025-07-07
Show arrow data in header tooltips and ever more data when alt is down

abey79:
### What

Show more data with alt down (see snapshot for an idea).

Side quests:
- moved tooltip code in its own file
- expanded `generic_placeholder_for_datatype`


crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -616,7 +618,39 @@ impl egui_table::TableDelegate for DataFusionTableDelegate<'_> {
...
+                            &column.desc,
+                            column_field.as_ref(),
+                            migrated_column_field.as_ref(),
+                            advanced_view,
+                        );
+                    });
+
+                    if content_changed {
+                        ui.ctx().request_repaint();
+                    }
```
abey79: It seems there should be an easier way to achieve this

crates/viewer/re_ui/src/context_ext.rs
```diff
@@ -151,6 +151,13 @@ pub trait ContextExt {
             self.ctx().debug_painter().add(Shape::mesh(mesh));
         }
     }
+
+    /// Whether to show extra information in the UI, e.g. in tooltips.
+    ///
+    /// This is controlled by holding the `Alt` key down.
+    fn show_extras(&self) -> bool {
+        self.ctx().input(|input| input.modifiers.alt)
+    }
```
grtlr, line 160: Maybe use a more general (and explicit) `.input()` method instead of `show_extras`?

This might interfere with `alt` being used in other places (it can mean different things than "show extras"). In those cases, being verbose might help.

abey79: The idea here was to establish a kind of pattern (see [this thread](https://rerunio.slack.com/archives/C08TEFXNHB2/p1751877552878009?thread_ts=1751875717.387879&cid=C08TEFXNHB2)) but yeah up to discussion.

Issue #9969, opened 2025-05-14
Tooltip jumps around


emilk:
![Image](https://github.com/user-attachments/assets/8c0d42a7-ca00-4564-b2a2-8b97829babc5)

Likely because the recent egui update

lucasmerlin:
Weird, I can't reproduce it. It does jump to the other side if you get too close to the right window border, but it's not jumping around randomly like in your video. Tried with the latest nightly. I also tried different screen resolutions. Does it still happen to you?

emilk:
No, I can't reproduce it either now. 🤷

emilk:
The problem is real:

![Image](https://github.com/user-attachments/assets/2bdece6e-6817-41c9-9b7e-b44ecf3d3c7f)

emilk:
I wonder why the tooltip goes to a new position each frame, even though there is obviously room in the preferred top-right position.

Is it maybe trying to avoid overlapping itself?
As in, each frame it looks for a good place to put itself, where it doesn't cover any other areas. And maybe it gets confused, and thinks the same tooltip from last frame is a _different_ area (maybe because of changing `Id`), and so it find a new position where it doesn't overlap where it was last frame.

In other words: the tooltip is affraid of its own shadow.

Maybe.

emilk:
* Found it: https://github.com/emilk/egui/pull/7304

PR #10527, opened 2025-07-07
Fix tooltips sometimes jumping around on screen

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/9969
* Updates to egui using https://github.com/emilk/egui/pull/7304


Wumpf:
I'm curious about the egui fix, but you accidentally linked the issue twice instead :)

emilk:
Oops! Here is the egui fix:
* https://github.com/emilk/egui/pull/7304

Wumpf (approved):
seems to work for me. Hopefully that fixes it once and for all 🤞🤞🤞

Issue #10291, opened 2025-06-18
Add warnings-as-errors flag for codegen and use it on CI


Wumpf:
We sometimes slip in warnings on our codegen. We should address those before landing things to `main`.
(not being hard error can be nice for development process, but we want to strive for a cleaner `main` branch!)

PR #10528, opened 2025-07-07
Add `warnings-as-errors` to codegen and use it on CI

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/10291

### What

Example error (after edit to `asset_video.fbs`)
```
Error (warning as error enabled):  /Users/andreas/dev/rerun-io/rerun2/crates/store/re_types/definitions/rerun/archetypes/asset_video.fbs rerun.archetypes.AssetVideo: `VideoFrameReference` can be written as a doclink, e.g. [archetypes.VideoFrameReference] in comment: /// In order to display a video, you also need to log a `VideoFrameReference` for each frame.
```


emilk (approved):
Nice

PR #10532, opened 2025-07-07
expose dataset maintenance endpoint on the frontend

zehiko:
As the title says, we're exposing ``DoMaintenance`` that serves the purpose of doing appropriate dataset maintenance on the Redap server.


rerun_py/src/catalog/connection_handle.rs
```diff
@@ -248,6 +248,22 @@ impl ConnectionHandle {
...
+        build_scalar_indexes: bool,
+        compact_fragments: bool,
+    ) -> PyResult<()> {
+        wait_for_future(py, async {
+            self.client()
+                .await?
+                .do_maintenance(dataset_id, build_scalar_indexes, compact_fragments)
+                .await
+                .map_err(to_py_err)
+        })
```
teh-cmc: .in_current_span()

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -248,6 +248,22 @@ impl ConnectionHandle {
         )
     }
 
+    pub fn do_maintenance(
```
teh-cmc, line 252: I'd rather we instrument this one, for consistency

rerun_py/src/catalog/connection_handle.rs
zehiko: ah, right, thanks!

rerun_py/src/catalog/connection_handle.rs
zehiko: makes sense, done

PR #10539, opened 2025-07-08
Fix spurious decode errors with H.264 on native Viewer

Wumpf:
Removes a previous attempt for lower latency that caused spurios decode errors, observed with ffmpeg 7.1 on Mac & Windows.

Also, fix retry-on-error condition: we were checking for the wrong kind of error in `VideoPlayerError::should_request_more_frames`.
This improves pausing on corrupt video data which sometimes may recover on decoder restarts.


Wumpf:
the new tests are very unhappy about this? Needs a closer look.

Wumpf:
ah ofc, the ones where we don't have a decoder bail now. Need to make the redraw condition more fine grained!

good test! :)

Issue #10496, opened 2025-07-03
Memory limit not honored by viewer caches (regression in 0.23)


ben-z:
**Describe the bug**
Garbage collection (and `rr.log(..., static=True)`) is broken in rerun 0.23.

**To Reproduce**
1. Try any `rr.log(..., static=True)` in a loop, and see that garbage collection fails to free up memory after memory limit is reached. 

**Expected behavior**

Garbage collection should work after memory limit is reached.

**Screenshots**

Comparison with memory limit set to `100MiB` (running in the [lerobot](https://github.com/huggingface/lerobot) repo):

```sh
export RUST_LOG=debug LEROBOT_RERUN_MEMORY_LIMIT=100MiB; python -m lerobot.teleoperate --robot.type=so101_follower --robot.port=$F1_PORT --robot.id=f1 --robot.cameras="${CAMERA_CONFIG}" --teleop.type=so101_leader --teleop.port=$L1_PORT --teleop.id=l1 --display_data=true
```

v0.22.1
<img width="2012" alt="image" src="https://github.com/user-attachments/assets/c9807a90-b8a5-44ea-84b2-efb79ef87837" />

v0.23.0 (same behavior in v0.23.4)
<img width="2012" alt="image" src="https://github.com/user-attachments/assets/3e3e82d5-4828-43e0-b762-4098aaf0a83f" />

**Backtrace**
N/A

**Desktop (please complete the following information):**
 - OS: macOS Sequoia 15.5 (24F74)

**Rerun version**

Not working:
```
rerun-cli 0.23.4 (default map_view nasm native_viewer release web_viewer) [rustc 1.84.0 (9fc6b4312 2025-01-07), LLVM 19.1.5] aarch64-apple-darwin release-0.23.4 b411df2, built 2025-06-26T13:24:43Z
Video features: av1 default ffmpeg nasm serde
```

Working:
```
rerun-cli 0.22.1 (default grpc map_view nasm native_viewer release web_viewer) [rustc 1.81.0 (eeb90cda1 2024-09-04), LLVM 18.1.7] aarch64-apple-darwin release-0.22.1 6b5a938, built 2025-02-20T10:18:45Z
Video features: av1 default ffmpeg nasm serde
```

**Additional context**

Wumpf:
Thanks for the report!
[...]

PR #10543, opened 2025-07-08
Improve coverage of memory panel

emilk:
### Related
* Part of investigating https://github.com/rerun-io/rerun/issues/10496

### What
This improves the memory panel by counting ALL stores (not just the active ones), as well as the _viewer_ caches (e.g. video caches etc).

image

### DIFF
--- a/crates/viewer/re_viewer/src/ui/memory_panel.rs
+++ b/crates/viewer/re_viewer/src/ui/memory_panel.rs
@@ -424,9 +424,7 @@ impl MemoryPanel {
             // TODO(emilk): turn off plot interaction, and always do auto-sizing
             .show(ui, |plot_ui| {
                 if let Some(max_bytes) = limit.max_bytes {
-                    plot_ui.hline(
-                        egui_plot::HLine::new("Limit (counted)", max_bytes as f64).width(2.0),
-                    );
+                    plot_ui.hline(egui_plot::HLine::new("Limit", max_bytes as f64).width(2.0));
                 }
 
                 for &time in &self.memory_purge_times {


crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -18,6 +18,12 @@ use crate::{Time, Timescale, VideoCodec, VideoDataDescription, VideoEncodingDeta
 #[repr(transparent)]
 pub struct WebVideoFrame(web_sys::VideoFrame);
 
+impl re_byte_size::SizeBytes for WebVideoFrame {
+    fn heap_size_bytes(&self) -> u64 {
+        0 // unknown, and hopefully VRAM
```
Wumpf: I guess even if we knew, we wouldn't care about it because that's part of the browser process and not on our wasm heap :)

-> nit:
```suggestion
        0 // Part of Browser's memory, not wasm heap.
```

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -366,6 +386,18 @@ pub struct SamplesStatistics {
...
 
+impl re_byte_size::SizeBytes for SamplesStatistics {
+    fn heap_size_bytes(&self) -> u64 {
+        let Self {
+            dts_always_equal_pts: _,
+            has_sample_highest_pts_so_far,
+        } = self;
+        has_sample_highest_pts_so_far
+            .as_ref()
+            .map_or(0, |bitvec| bitvec.len() as u64 / 8)
```
Wumpf: shouldn't we check the capacity instead?

crates/viewer/re_view_spatial/src/mesh_cache.rs
```diff
@@ -85,6 +86,22 @@ impl Cache for MeshCache {
...
+                std::mem::size_of_val(key) as u64
+                    + map
+                        .iter()
+                        .map(|(key, mesh)| {
+                            std::mem::size_of_val(key) as u64
+                                + mesh.as_deref().map_or(0, |mesh| mesh.total_size_bytes())
+                        })
+                        .sum::<u64>()
+            })
+            .sum::<u64>()
```
Wumpf: why is there no implementation of `re_byte_size::SizeBytes` for `HashMap`  that we can call instead?

crates/viewer/re_view_spatial/src/proc_mesh.rs
```diff
@@ -193,6 +215,13 @@ impl Cache for WireframeCache {
         self.0.clear();
     }
 
+    fn bytes_used(&self) -> u64 {
+        self.0
```
Wumpf: again, would be nice if we just had `heap_size_bytes` on hashmaps!
(refraining from commenting this on the rest of the stuff in here, maybe you had a good reason?)

crates/viewer/re_viewer_context/src/cache/image_decode_cache.rs
```diff
@@ -134,6 +134,11 @@ impl Cache for ImageDecodeCache {
         self.generation += 1;
     }
 
+    /// Total memory used by this cache, in bytes.
+    fn bytes_used(&self) -> u64 {
+        self.memory_used
```
Wumpf: huh, a bit weird that we're not using `re_byte_size::SizeBytes` for this at any point. But I figure this cache has to keep track of its own memory because that's part of its purging strategy

crates/viewer/re_viewer_context/src/cache/image_stats_cache.rs
```diff
@@ -24,6 +25,11 @@ impl Cache for ImageStatsCache {
         // Purging the image stats is not worth it - these are very small objects!
     }
 
+    /// Total memory used by this cache, in bytes.
+    fn bytes_used(&self) -> u64 {
+        self.0.total_size_bytes()
```
Wumpf, line 30: oh, so we do have re_byte_size::SizeBytes on HashMaps, but we weren't able to call it on some of the other types?

crates/viewer/re_viewer_context/src/cache/video_asset_cache.rs
```diff
@@ -23,6 +24,19 @@ struct Entry {
...
 
+impl re_byte_size::SizeBytes for Entry {
+    fn heap_size_bytes(&self) -> u64 {
+        let Self {
+            used_this_frame: _,
+            video,
+        } = self;
+        match video.as_ref() {
+            Ok(video) => video.heap_size_bytes(),
+            Err(_) => 100, // close enough
```
Wumpf, line 35: a bit crude :/
But yeah I can see that `VideoLoadError` is very annoying here

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -22,12 +23,23 @@ use crate::Cache;
...
+    source_chunk_id: ChunkId,
 
     /// Indexes into [`re_video::VideoDataDescription::samples`] that this buffer contains.
     sample_index_range: std::ops::Range<re_video::SampleIndex>,
 }
 
+impl re_byte_size::SizeBytes for SampleBuffer {
+    fn heap_size_bytes(&self) -> u64 {
+        let Self {
+            buffer: _, // ref-counted - already counted in the store
```
Wumpf, line 35: Could we check if we're the last ones holding a reference to this buffer?
That would likely hint at a bug, but I also wouldn't be shocked

crates/viewer/re_view_spatial/src/mesh_cache.rs
emilk: There is, but the default implementation for `Arc` is to count iself as zero:

https://github.com/rerun-io/rerun/blob/e7468460e8884a1878e0693cca6d164b58ce3ab8/crates/utils/re_byte_size/src/std_sizes.rs#L119-L124

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
emilk: Good idea - but doesn't seem like we can: https://docs.rs/arrow/latest/arrow/buffer/struct.Buffer.html

crates/viewer/re_view_spatial/src/mesh_cache.rs
emilk: But if I change the definition of `Arc` to this:

https://github.com/rerun-io/rerun/blob/278fb1aeed909db18fbeb0a2f346c19191db412d/crates/utils/re_byte_size/src/std_sizes.rs#L119-L128

…it should work nicely. 


Or I could divide by the ref count 🤔

teh-cmc: I like the sound of either of those :eyes:

PR #10545, opened 2025-07-08
Fix H.264 with b-frames not decoding last few frames on native

Wumpf:
### Related

* Closes almost #8073
    * technically this closes the original issue, but `VideoStream` still has still problem for incoming data as we never call `end_of_video` which for ffmpeg specifically means we never receive the last few frames. A follow-up PR will address those peculiarities

### What

The issue came down to problems in our frame-info to frame assignment: we get a stream of presentation frame data from ffmpeg without any metadata. Therefore, we have to figure out which frame belongs to which metadata.
We tried to account for DTS reordering by looking at the DTS/PTS relationship, but this turned out to be error prone since we falsely assumed that DTS catching up with PTS is the significant event for synchronization. However, by now we already have frame numbers in PTS order, so instead basing our assignment logic on that fixes the issue and has much simpler logic.

Tested with various H.264 videos on Windows & Mac

### DIFF
--- a/crates/utils/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
+++ b/crates/utils/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
@@ -194,8 +194,8 @@ struct FFmpegProcessAndListener {
 
     /// Number of samples submitted to ffmpeg that has not yet been outputted by ffmpeg.
     ///
-    /// This is a counter for debugging purposes only.
-    outstanding_frames: Arc<AtomicI32>,
+    /// This counter is for debugging purposes only.
+    num_outstanding_frames: Arc<AtomicI32>,
 
     /// If true, the write thread will not report errors. Used upon exit, so the write thread won't log spam on the hung up stdin.
     stdin_shutdown: Arc<AtomicBool>,
@@ -346,7 +346,7 @@ impl FFmpegProcessAndListener {
 
         Ok(Self {
             ffmpeg,
-            outstanding_frames,
+            num_outstanding_frames: outstanding_frames,
             frame_info_tx,
             frame_data_tx,
             listen_thread: Some(listen_thread),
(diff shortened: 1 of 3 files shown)


Wumpf:
Wait a sec, I should be able to enable tests for end of video on H.264 now!

crates/utils/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
```diff
@@ -344,7 +346,7 @@ impl FFmpegProcessAndListener {
 
         Ok(Self {
             ffmpeg,
-            outstanding_frames,
+            num_outstanding_frames: outstanding_frames,
```
emilk: let's use the same name everywhere

crates/utils/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
```diff
@@ -516,68 +518,66 @@ fn write_ffmpeg_input(
 struct FrameBuffer {
     /// Received frame-infos, waiting to be matched to output frames.
     ///
-    /// Sorted by their presentation timestamp.
-    pending: BTreeMap<Time, FFmpegFrameInfo>,
+    /// Sorted by their frame number.
```
emilk: …in presentation order?

crates/utils/re_video/src/decode/ffmpeg_h264/ffmpeg.rs
```diff
@@ -516,68 +518,66 @@ fn write_ffmpeg_input(
...
+    pending: BTreeMap<u32, FFmpegFrameInfo>,
 
-    /// Hightess DTC (Decoder timestamp) encountered so far.
-    highest_dts: Time,
+    /// The frame number of the next frame if we had any so far.
+    next_frame_nr: Option<u32>,
```
emilk: Why is this an option? Is `Some(0)` even a valid state?

Wouldn't a normal `u32` starting at zero make more sense?

emilk: Ah right, we may not be playing from the start. Might be worth a comment.

emilk (approved):
Finally 🥳

emilk (approved):
Finally 🥳

PR #10546, opened 2025-07-08
Fix a few problems with iterating over hash types

emilk:
### Related
* Part of #6198

### DIFF
--- a/crates/store/re_chunk_store/src/query.rs
+++ b/crates/store/re_chunk_store/src/query.rs
@@ -59,7 +59,9 @@ impl ChunkStore {
     }
 
     /// Retrieve all [`ComponentDescriptor`]s in the store.
-    pub fn all_components(&self) -> BTreeSet<ComponentDescriptor> {
+    ///
+    /// See also [`Self::all_components_sorted`].
+    pub fn all_components(&self) -> IntSet<ComponentDescriptor> {
         self.static_chunk_ids_per_entity
             .values()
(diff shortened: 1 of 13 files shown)


crates/store/re_chunk_store/src/query.rs
```diff
@@ -59,7 +59,7 @@ impl ChunkStore {
     }
 
     /// Retrieve all [`ComponentDescriptor`]s in the store.
-    pub fn all_components(&self) -> IntSet<ComponentDescriptor> {
+    pub fn all_components(&self) -> BTreeSet<ComponentDescriptor> {
```
teh-cmc: This is unsorted (and more importantly: un-hashed) on purpose, every bit of perf matters here.

There is already a sorted version of this method available just below.

crates/viewer/re_viewer/src/app.rs
```diff
@@ -1793,7 +1793,7 @@ impl App {
 
         for event in store_events {
             let chunk = &event.diff.chunk;
-            for component_descr in chunk.components().keys() {
+            for component_descr in chunk.components().keys().sorted() {
```
teh-cmc: There is no reason to sort here? We fire a lot of store events, that's not cheap.

crates/store/re_chunk_store/src/query.rs
emilk: There was only one caller of this function - but yeah, it should use `all_components_sorted` instead

teh-cmc (approved):

PR #10547, opened 2025-07-08
Fix `set_time_ctrl` not doing anything when called twice

jprochazk:
To repro the original issue on `main`:
1. call `viewer.set_time_ctrl`
2. move the time cursor in the viewer somewhere else
3. call `viewer_set_time_ctrl` again with _the same arguments_ (same time/timeline)

(3) would not adjust the time cursor like in step (1). This is due to `_time_ctrl` being _stateful_, but not bidirectionally synchronized between the Viewer and SDK. Moving the time cursor doesn't update the `_time_ctrl` value back in Python land, and so the `_time_ctrl` value in (3) isn't changed from (1), resulting in no state update for the Viewer.

For these kinds of things it makes more sense to use custom messages for SDK->Viewer uni-directional things. For Viewer->SDK, we have events exposed for timeline state, which are correctly fired whenever a user input in the Viewer results in the state changing.

### DIFF
--- a/rerun_notebook/src/js/widget.ts
+++ b/rerun_notebook/src/js/widget.ts
@@ -48,11 +48,6 @@ class ViewerWidget {
 
     model.on("msg:custom", this.on_custom_message);
 
-    model.on("change:_time_ctrl", (_, [timeline, time, play]) =>
-      this.on_time_ctrl(null, timeline, time, play),
-    );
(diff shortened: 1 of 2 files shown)


ntjohnson1 (approved):
LGTM. I'm going to pull this into my branch and apply the same change for the partition_url update.

PR #10550, opened 2025-07-08
150ms flush tick by default

teh-cmc:
Title.

### DIFF
--- a/crates/store/re_chunk/src/batcher.rs
+++ b/crates/store/re_chunk/src/batcher.rs
@@ -146,8 +146,8 @@ impl Default for ChunkBatcherConfig {
 impl ChunkBatcherConfig {
     /// Default configuration, applicable to most use cases.
     pub const DEFAULT: Self = Self {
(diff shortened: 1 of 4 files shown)


teh-cmc:
Changed to 200ms, so that it maps to an integer frequency

teh-cmc:
Camera examples are snappy once again

emilk (approved):
Nice!

emilk:
@Wumpf maybe you want to update the live video stream example too?

emilk:
Let's hold off on merging until we are sure that this is the right approach.

Discussion here: https://rerunio.slack.com/archives/C033K5VS2KD/p1752057267261589

Ideally, the default for gRPC sinks would remain at <20ms, while the default for files can be much higher (even >1s).

emilk:
@Wumpf is taking a look to see if we can easily have different defaults for file sinks vs gRPC sinks

teh-cmc:
* Superseded by https://github.com/rerun-io/rerun/pull/10620

Issue #10449, opened 2025-06-30
Notebook widget should dynamically size to the width of the notebook


jleibs:
Here's an example of a super irritating situation:

https://github.com/user-attachments/assets/a6728616-ade5-43d4-ae8f-9b12c5f3c710

My content would otherwise fit just fine, but the viewer is being arbitrarily scaled to a fixed with which was too small for my screen and now I have to scroll.

jprochazk:
A good starting point would be to add a `width=AUTO` mode, where the viewer would take up 100% of the output cell's width. We'd also need to handle height properly. If `width=AUTO` and `height` is not set explicitly, then `height=AUTO` as well. `height=AUTO` doesn't mean it takes up 100% of anything, it just adjusts the height to scale with the width according to some reasonable aspect ratio (e.g. 4:3 or 16:9)

PR #10554, opened 2025-07-08
Notebook auto-sizing

jprochazk:
* Closes https://github.com/rerun-io/rerun/issues/10449

`width` and `height` can now be set to `auto`.
- for `width`, it sets the resulting Viewer's width to 100% of the notebook cell's width
- for `height`, it matches `width` using a 16:9 aspect ratio.

This works the same way even if using numeric for one of width/height and `auto` for the other.

The default width/height is unchanged.

```py
from rerun.notebook import Viewer

Viewer(width="auto", height="auto")
```

### DIFF
--- a/rerun_py/rerun_sdk/rerun/notebook.py
+++ b/rerun_py/rerun_sdk/rerun/notebook.py
@@ -83,10 +83,14 @@ class Viewer:
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


emilk:
Can we have a video please? Would be nice for the changelog!

PR #10555, opened 2025-07-08
Fix issues with end of `VideoStream` & live `VideoStream`

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/8073

### What

This PR consists of two distinct fixes:
* call `end_of_video` after a while for live streams
    * usually we do this when reaching the end of a video. But if a video doesn't have a known duration, we would otherwise not do this. Now there's a wall-clock timer tracking the last addition to the stream that will trigger this 
    * Note that this fix alone unlocks the image comparison test which operates on static video
* have a special tolerance for playing live video
    * today's ffmpeg-exe decoder is not able to emit all frames without decoding delay, so we have to be more tolerant
    * however, the fix is written in a way that it a) gives better experience by not showing a spinner on incoming video b) adjusts to a decoder's need - we already encapsulated the ffmpeg pecularities in `min_num_samples_to_enqueue_ahead`
    * this unearthed a subtle bug in our latest-at frame query 😬 (that's the part that took me longest to discover)

Commit by commit!

Despite our automated tests, let's make manually make sure we didn't regress anything!
(This is a platform independent change on the player, but to be on the save side we should test web as well)
* [x] test live `VideoStream` on Mac
* [x] test static `VideoStream` on Mac
* [x] test `VideoAsset` on Mac
* [ ] test live `VideoStream` on Windows
* [ ] test static `VideoStream` on Windows
* [ ] test `VideoAsset` on Windows
* [ ] test live `VideoStream` on Browser
* [ ] test static `VideoStream` on Browser
* [ ] test `VideoAsset` on Browser


crates/viewer/re_renderer/src/video/chunk_decoder.rs
```diff
@@ -152,7 +152,26 @@ impl VideoSampleDecoder {
...
+            .map_or(true, |(_, frame)| frame.info.presentation_timestamp > pts)
+        {
+            if let Some((next_older_pts, next_older_frame)) = frames_before_pts.pop_last() {
+                debug_assert!(next_older_pts < pts);
+
+                decoder_output
+                    .frames_by_pts
+                    .insert(next_older_pts, next_older_frame);
+            }
+        }
```
Wumpf: this doesn't feel very elegant at all. There gotta be a nicer way to write this.
Also took me ages to spot the bug.

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -505,42 +545,69 @@ impl VideoPlayer {
...
+                &(sample_idx_end.saturating_sub(allowed_delay)..sample_idx_end),
             ) {
-                DecoderDelayState::Behind
-            } else {
-                DecoderDelayState::UpToDateWithinTolerance
+                if sample.presentation_timestamp == last_decoded_frame_pts {
+                    return DecoderDelayState::UpToDateToleratedEdgeOfLiveStream;
+                }
             }
         }
```
Wumpf: this block is new, everything else in this method is just moving around stuff and shouldn't affect behavior
(I did in fact try a bit with this block commented out to check if that's sane as expected)

crates/viewer/re_renderer/src/video/mod.rs
```diff
@@ -68,6 +68,12 @@ pub enum DecoderDelayState {
     /// The decoder is caught up with the most recent requested frame.
     UpToDate,
 
+    /// We're not up to date, but we're close enough to the newest content of a live stream that we're ok.
```
emilk, line 73: What is the tolerance value? Can we link to it from the docstring?

crates/viewer/re_renderer/src/video/mod.rs
```diff
@@ -68,6 +68,12 @@ pub enum DecoderDelayState {
...
+    ///
```
emilk, line 74: We should explain here why we treat live-streams differently from assets

crates/viewer/re_renderer/src/video/mod.rs
```diff
@@ -86,7 +92,9 @@ impl DecoderDelayState {
     pub fn should_request_more_frames(&self) -> bool {
         match self {
             Self::UpToDate => false,
-            Self::UpToDateWithinTolerance | Self::Behind => true,
+            Self::UpToDateWithinTolerance
+            | Self::Behind
+            | Self::UpToDateToleratedEdgeOfLiveStream => true,
```
emilk, line 115: Does this mean egui will keep repainting if we pause towards the end of a live video stream? Or will we eventually flush it and end up in `UpToDate`?

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -34,6 +34,10 @@ const DECODING_GRACE_DELAY_BEFORE_REPORTING: Duration = Duration::from_millis(40
 ///      would cause us to show in-progress frames until the tolerance is hit
 const TOLERATED_OUTPUT_DELAY_IN_NUM_FRAMES: usize = 3;
 
+/// If we haven't seen new samples in this amount of time, we assume the video has ended
+/// and signal the end of the video to the decoder.
+const TIME_UNTIL_VIDEO_ASSUMED_ENDED: Duration = Duration::from_millis(250);
```
```suggestion
/// If we haven't seen new samples in this amount of time, we assume the video has ended
/// and signal the end of the video to the decoder.
///
/// A high value here will result in…
/// A low value here will result in …
const TIME_UNTIL_VIDEO_ASSUMED_ENDED: Duration = Duration::from_millis(250);
```

We should also consider moving all these options into a proper `struct` that we can tweak in some advanced video settings (i.e. add to `AppOptions`).

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -488,15 +523,20 @@ impl VideoPlayer {
...
-            self.sample_decoder.end_of_video()?;
+            self.signal_end_of_video_if_not_already_signaled()?;
         }
 
         Ok(())
     }
 
+    fn signal_end_of_video_if_not_already_signaled(&mut self) -> Result<(), VideoPlayerError> {
+        if !self.signaled_end_of_video {
+            self.sample_decoder.end_of_video()?;
```
emilk, line 551: Why do we need a bool to protect this call? Can we make the call idempotent instead?
We have so much state in this module already

crates/viewer/re_renderer/src/video/chunk_decoder.rs
```diff
@@ -152,7 +152,26 @@ impl VideoSampleDecoder {
...
+            .is_none_or(|(_, frame)| frame.info.presentation_timestamp > pts)
+        {
+            if let Some((next_older_pts, next_older_frame)) = frames_before_pts.pop_last() {
+                debug_assert!(next_older_pts < pts);
+
+                decoder_output
+                    .frames_by_pts
+                    .insert(next_older_pts, next_older_frame);
+            }
+        }
```
emilk: This would be a lot shorter written as latest-at query followed by a split:

```suggestion
let latest_at_pts =
	 decoder_output.frames_by_pts
	 .range(..=pts)
	 .next_back()
	 .map_or(pts, |(k, v)| *k);

decoder_output.frames_by_pts = decoder_output.frames_by_pts.split_off(&latest_at_pts);
```

crates/viewer/re_renderer/src/video/mod.rs
Wumpf: we keep repainting until we hit `UpToDate`. If we pause at the end of a live video stream we should eventually hit `UpToDate` - all other states mean that we're still wait on more results

sounds like I should clarify that in the doc comments! Or maybe `UpToDate*Tolerance*` is the wrong naming!

crates/viewer/re_renderer/src/video/player.rs
Wumpf: I don't want to call `end_of_video` whenever a static `VideoStream` is noticed to be old.

This conveniently _also_ protects from calling `end_of_video` if we _both_ enqueue the last gop and notice that a `VideoStream` is not being updated anymore.

Seemed like the easiest solution to catch all of these 🤷

emilk:
The tests fail locally on my mac, even on main :/

emilk:
@rerun-bot full-check

Wumpf:
Mac CI was happy with those though: https://github.com/rerun-io/rerun/actions/runs/16177965355/job/45667554861
Had it succeed many times on my Mac, giving it another spin post your merge now just to be sure

@emilk What ffmpeg version are you running with?

emilk:
```
❯ ffmpeg --version
ffmpeg version 7.1.1 Copyright (c) 2000-2025 the FFmpeg developers
  built with Apple clang version 16.0.0 (clang-1600.0.26.6)
  configuration: --prefix=/opt/homebrew/Cellar/ffmpeg/7.1.1_2 --enable-shared --enable-pthreads --enable-version3 --cc=clang --host-cflags= --host-ldflags='-Wl,-ld_classic' --enable-ffplay --enable-gnutls --enable-gpl --enable-libaom --enable-libaribb24 --enable-libbluray --enable-libdav1d --enable-libharfbuzz --enable-libjxl --enable-libmp3lame --enable-libopus --enable-librav1e --enable-librist --enable-librubberband --enable-libsnappy --enable-libsrt --enable-libssh --enable-libsvtav1 --enable-libtesseract --enable-libtheora --enable-libvidstab --enable-libvmaf --enable-libvorbis --enable-libvpx --enable-libwebp --enable-libx264 --enable-libx265 --enable-libxml2 --enable-libxvid --enable-lzma --enable-libfontconfig --enable-libfreetype --enable-frei0r --enable-libass --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libopenjpeg --enable-libspeex --enable-libsoxr --enable-libzmq --enable-libzimg --disable-libjack --disable-indev=jack --enable-videotoolbox --enable-audiotoolbox --enable-neon
  libavutil      59. 39.100 / 59. 39.100
  libavcodec     61. 19.101 / 61. 19.101
  libavformat    61.  7.100 / 61.  7.100
  libavdevice    61.  3.100 / 61.  3.100
  libavfilter    10.  4.100 / 10.  4.100
  libswscale      8.  3.100 /  8.  3.100
  libswresample   5.  3.100 /  5.  3.100
  libpostproc    58.  3.100 / 58.  3.100
Unrecognized option '-version'.
Error splitting the argument list: Option not found
```

Wumpf:
Running with the same version on my M1 (slightly different build, but shouldn't matter)
Ran `cargo nextest run --all-features --all-targets` and `cargo nextest run --all-features --all-targets video` (to get some scheduling variation) both succeeding on my machine :/

crates/viewer/re_renderer/src/video/player.rs
Wumpf: done that now, looks a lil bit nicer but oh my I really don't want to expose any of these, every single one of them is just insane in what it does >.<

crates/viewer/re_renderer/src/video/mod.rs
Wumpf: we haven't done that on the others either, but sure why not

Wumpf:
addressed all review comments. Time for another round of testing (both manual & automated). Unfortunately I don't have a lead on Emil's issue yet :/

* [x] regular ci
* [ ] full ci
* [x] repeat all manual tests listed in descr - skipped Windows right now, Mac & Mac Web should be enough

Wumpf:
@rerun-bot full-check

emilk:
My issues exist on `main`, so should not hold up this PR

PR #10557, opened 2025-07-08
Dynamically Set Active Partition URL and Select Time Window

ntjohnson1:
### What
1. This adds a public method `set_active_partition_url` on the viewer aligned with `set_active_recording_id` to dynamically toggle that
2. This updates `dataset.partition_url` to take in arguments that can constrain the observed time window
3. This fixes a variety of paper cuts related to `pixi run fast-lint` apparently linux users haven't seen these issues even though I think they are actual bugs

### DIFF
--- a/rerun_notebook/src/rerun_notebook/__init__.py
+++ b/rerun_notebook/src/rerun_notebook/__init__.py
@@ -261,8 +261,8 @@ class Viewer(anywidget.AnyWidget):  # type: ignore[misc]
     def set_active_recording(self, recording_id: str) -> None:
         self.send({"type": "recording_id", "recording_id": recording_id})
 
-    def set_active_partition_uri(self, partition_uri: str) -> None:
-        self._partition_uri = partition_uri
+    def set_active_partition_url(self, partition_url: str) -> None:
+        self._partition_url = partition_url
 
     def _on_raw_event(self, callback: Callable[[str], None]) -> None:
         """Register a set of callbacks with this instance of the Viewer."""
--- a/rerun_notebook/src/rerun_notebook/__init__.py
+++ b/rerun_notebook/src/rerun_notebook/__init__.py
@@ -261,6 +261,9 @@ class Viewer(anywidget.AnyWidget):  # type: ignore[misc]
     def set_active_recording(self, recording_id: str) -> None:
         self.send({"type": "recording_id", "recording_id": recording_id})
 
+    def set_active_partition_uri(self, partition_uri: str) -> None:
+        self._partition_uri = partition_uri
+
     def _on_raw_event(self, callback: Callable[[str], None]) -> None:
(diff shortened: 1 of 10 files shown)


rerun_notebook/src/rerun_notebook/__init__.py
```diff
@@ -154,6 +154,8 @@ class Viewer(anywidget.AnyWidget):  # type: ignore[misc]
     _width = traitlets.Int(allow_none=True).tag(sync=True)
     _height = traitlets.Int(allow_none=True).tag(sync=True)
 
+    # TODO(nick): This traitlet is only used for initialization
```
jprochazk, line 157: We could achieve this by calling `open_url` in the constructor. We'd need to ensure that events sent before the Viewer is ready are buffered like data with `_data_queue`.

At that point having another queue doesn't seem great, because data, tables, and other events all go through the same interface. We should have a single `_event_queue` and a helper (`_send_buffered` or something) to push data into it, or send it directly depending on if the Viewer is ready yet.

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -606,8 +606,7 @@ def load_archive(path_to_rrd: str | os.PathLike[str]) -> RRDArchive:
     """
 
 # AI generated stubs for `PyRecordingStream` related class and functions
-# TODO(#9187): this will be entirely replaced with `RecordingStream` is itself written in Rust
-
+# TODO(#9187): this will be entirely replaced with `RecordingStream` is qitself written in Rust
```
jprochazk: Typo here:

```suggestion
# TODO(#9187): this will be entirely replaced when `RecordingStream` is itself written in Rust
```

(also `with` -> `when` while we're already at it)

rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -360,6 +360,25 @@ def set_active_recording(
...
 
+    def set_active_partition_url(
+        self,
+        *,
+        url: str,
+    ) -> None:
+        """
+        Set the active partition url for the viewer.
+
+        This is equivalent to clicking on the partition in the dataset table.
```
jprochazk: Does clicking on the partition in the dataset table close the previously opened partition? If not, then this line seems misleading 🤔

In general, this would probably be better served by the `open_url` API, because you should be able to have more than one partition open at a time, right?

ntjohnson1: Removed the line. I guess I don't fully know how we are managing `open` urls vs the single active one.

ntjohnson1:
As you can see I ran into a bit of a headache with CI. I'm punting on open_url for follow up. If you have bandwidth to pick it up that'd be great. Otherwise we can probably check in tomorrow.

Issue #10561, opened 2025-07-08
Running `rr.init()` no longer resets notebook state correctly


jleibs:
I believe this PR broke a normal notebook workflow:
https://github.com/rerun-io/rerun/pull/10201

See: https://github.com/rerun-io/rerun/blob/main/examples/python/notebook/cube.ipynb

It used to be normal/expected to call `rr.init()` multiple times in the notebook to reset the context of the recording.

Now this produces an error:

<img width="1565" height="1363" alt="Image" src="https://github.com/user-attachments/assets/4484586b-6a00-4985-809c-ff0dcaddf461" />

Issue #10562, opened 2025-07-08
`RecordingStream` objects are no longer isolated


jleibs:
Simple repro:
```
import rerun as rr

rec1 = rr.RecordingStream("rerun_example", recording_id="episode42")
rec1.log("/small_data", rr.TextLog("Small data -- send it to the viewer"))
rec1.spawn()

rec2 = rr.RecordingStream("rerun_example", recording_id="episode42")
rec2.log("/big_data", rr.TextLog("Big data -- send it to a file"))
rec2.save("big_data.rrd")
```
## Expected Behavior
- `/small_data` should end up in the viewer
- `/big_data` should end up in the `.rrd` file

When opening `big_data.rrd` later, it is merged into the recording, correctly.

This works as expected in 0.23.

## Regression
- All data is sent to the viewer

<img width="1641" height="1053" alt="Image" src="https://github.com/user-attachments/assets/a80183fc-2c04-47fc-bd4a-9419c2061ce0" />

PR #10563, opened 2025-07-08
Fix deadlock while preserving recording isolation

jleibs:
### Related

- Resoles: https://github.com/rerun-io/rerun/issues/10562
- Resolves: https://github.com/rerun-io/rerun/issues/10561
- While fixing the dead-lock originally addressed in: https://github.com/rerun-io/rerun/pull/10201

### Also
In order to preserve backwards compatibility in environments where `rr.init()` was depended upon to clean up existing resources semi-deterministically due to creation of a new stream, this now does a cleanup sweep of any recordings that only exist in the all_recordings list.

### DIFF
--- a/crates/top/re_sdk/src/recording_stream.rs
+++ b/crates/top/re_sdk/src/recording_stream.rs
@@ -816,11 +816,7 @@ impl RecordingStream {
     pub fn ref_count(&self) -> usize {
         match &self.inner {
             Either::Left(strong) => Arc::strong_count(strong),
-            Either::Right(weak) => weak
-                .upgrade()
-                .map_or(0, |this: Arc<Option<RecordingStreamInner>>| {
-                    Arc::strong_count(&this)
-                }),
+            Either::Right(weak) => weak.strong_count(),
         }
     }
 }
--- a/crates/top/re_sdk/src/recording_stream.rs
+++ b/crates/top/re_sdk/src/recording_stream.rs
@@ -811,6 +811,18 @@ impl RecordingStream {
             },
         }
     }
+
+    /// Returns the current reference count of the [`RecordingStream`].
+    pub fn ref_count(&self) -> usize {
+        match &self.inner {
+            Either::Left(strong) => Arc::strong_count(strong),
+            Either::Right(weak) => weak
+                .upgrade()
+                .map_or(0, |this: Arc<Option<RecordingStreamInner>>| {
+                    Arc::strong_count(&this)
+                }),
+        }
+    }
 }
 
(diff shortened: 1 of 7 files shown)


crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -811,6 +811,18 @@ impl RecordingStream {
...
     }
+
+    /// Returns the current reference count of the [`RecordingStream`].
+    pub fn ref_count(&self) -> usize {
+        match &self.inner {
+            Either::Left(strong) => Arc::strong_count(strong),
+            Either::Right(weak) => weak
+                .upgrade()
+                .map_or(0, |this: Arc<Option<RecordingStreamInner>>| {
+                    Arc::strong_count(&this)
```
teh-cmc: Right?
```suggestion
                    Arc::strong_count(&this).saturating_sub(1)
```

teh-cmc: Or actually, why not just [`weak.strong_count()`](https://doc.rust-lang.org/std/sync/struct.Weak.html#method.strong_count)?

rerun_py/src/python_bridge.rs
```diff
@@ -254,6 +253,28 @@ fn rerun_bindings(py: Python<'_>, m: &Bound<'_, PyModule>) -> PyResult<()> {
...
+            recording.flush_blocking();
+        }
+
+        // Flush the garbage queue.
+        flush_garbage_queue();
+
+        // Finally remove any recordings that have a refcount of 1, which means they are ONLY
+        // referenced by the `all_recordings` list and thus can't be referred to by the Python SDK.
+        all_recordings().retain(|recording| recording.ref_count() > 1);
+    });
```
teh-cmc, line 276: I think if we were to run this logic when initializing a `rr.RecordingStream`, we might be close to solving:
* https://github.com/rerun-io/rerun/issues/2116

teh-cmc (approved):
sweet

jprochazk:
This should also close https://github.com/rerun-io/rerun/issues/10561, right?

crates/top/re_sdk/src/recording_stream.rs
jleibs: Thank you. Claude suggestions making me a worse coder.

rerun_py/src/python_bridge.rs
jleibs: I was thinking about that but figured there are more weird cases where we create recording_streams than there should be where we call rr.init().

Issue #10540, opened 2025-07-08
Port our use of old `memory.popup` API with new `egui::Popup`


emilk:
This is part of making sure the new API is good enough to deprecate the old API.

PR #10565, opened 2025-07-09
Use `egui::Popup` for notification panel

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/10540
* Which in turn is to unblock further work on https://github.com/emilk/egui/issues/7037
* Which in turn is to unblock the next egui release
* Which in turn is to unblock the next Rerun release :)
* Even nicer API coming in https://github.com/emilk/egui/pull/7315

### What
No functional changes - just using the new API.

In other words, mostly moving code around.

### DIFF
--- a/crates/viewer/re_selection_panel/src/view_space_origin_ui.rs
+++ b/crates/viewer/re_selection_panel/src/view_space_origin_ui.rs
@@ -208,7 +208,7 @@ fn view_space_origin_widget_editing_ui(
         }
     };
 
-    ui.list_item_popup(popup_id, &output.response, 4.0, suggestions_ui);
+    ui.list_item_popup(popup_id, &output.response, suggestions_ui);
 
     if control_flow.is_continue() && !ui.memory(|mem| mem.is_popup_open(popup_id)) {
         control_flow = ControlFlow::Break(None);
--- a/crates/viewer/re_ui/src/ui_ext.rs
+++ b/crates/viewer/re_ui/src/ui_ext.rs
@@ -1,7 +1,7 @@
 use std::hash::Hash;
 
 use egui::{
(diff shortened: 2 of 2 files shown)


crates/viewer/re_viewer/src/app.rs
```diff
@@ -1575,6 +1575,8 @@ impl App {
                     self.show_text_logs_as_notifications();
                 }
             });
+
+        self.notifications.show_toasts(egui_ctx);
```
emilk, line 1579: This was previously called from the same function that showed the notification button, which I think was a mistake. Even if the top panel is hidden we should still show the toast notifications.

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -232,57 +212,46 @@ impl NotificationPanel {
...
+            if !notifications.is_empty() {
+                ui.label(format!("Notifications ({})", notifications.len()));
+            } else {
+                ui.label("Notifications");
+            }
+            ui.with_layout(egui::Layout::top_down(egui::Align::Max), |ui| {
+                if ui.small_icon_button(&icons::CLOSE, "Close").clicked() {
+                    ui.memory_mut(|mem| {
+                        mem.close_popup(notification_panel_popup_id());
                     });
```
emilk:
```suggestion
                    ui.close();
```

lucasmerlin (approved):
Nice! I think the popup being all the way in the right corner feels odd though, I think it feels more natural if it's aligned with the button.

Screenshot 2025-07-09 at 11 04 42

Other apps usually also have it aligned with the button:

Screenshot 2025-07-09 at 11 06 42

emilk:
Some do, but not all. For instance, this is the control panel in macOS:

Screenshot 2025-07-09 at 11 45 15

I'd rather not change katyas design here unnecessarily

PR #10538, opened 2025-07-08
Add `RegisterTable` rpc, `LanceTable` message, and related proto changes

abey79:
### Related

- part of https://linear.app/rerun/project/add-the-abillity-to-register-a-lance-table-with-the-catalog-0e23e1128967/overview
- part of https://github.com/rerun-io/dataplatform/issues/720
- sibling https://github.com/rerun-io/dataplatform/pull/1197
- closes https://github.com/rerun-io/dataplatform/issues/1173

### What

Add support for `RegisterTable`, to register a foreign lance table to a redap server as a new table entry. Includes:

- Add `RegisterTable` rpc and associated messages
- Remove `Schema` from `TableEntry`
- Add `CatalogClient.register_table` to the Python SDK

Not included in this PR but obvious next step: add support for remote tables in the redap browser.

PR #10569, opened 2025-07-09
Add support for table entries in the redap browser

abey79:
### Related

- follow-up to https://github.com/rerun-io/rerun/pull/10538
- part of https://linear.app/rerun/project/add-the-abillity-to-register-a-lance-table-with-the-catalog-0e23e1128967/overview
- part of https://github.com/rerun-io/dataplatform/issues/720

### What

This PR introduces support for table entries in the redap browser. The tables now appear sorted by name, after the datasets. System tables are filtered out for now (currently this is the `__entries` table).

Side quests:
- some cleanup of connection client code
- renamed `EntryKind` to `DatasetKind` as it's a much more accurate description of what this thing is about
- I initially considered splitting `Item::RedapEntry` into two, one for dataset and one for tables, but I backtracked from that as its not really required.

The latter to points are clearly part of the debt we need to repay: https://linear.app/rerun/project/repay-debt-around-re-viewe-storehub-displaymode-and-the-recording-145b17660709/overview

### DIFF
--- a/crates/viewer/re_context_menu/src/lib.rs
+++ b/crates/viewer/re_context_menu/src/lib.rs
@@ -382,10 +382,9 @@ trait ContextMenuAction {
                 }
                 Item::Container(container_id) => self.process_container(ctx, container_id),
                 Item::RedapServer(origin) => self.process_redap_server(ctx, origin),
-                Item::RedapDatasetEntry(entry_id) => {
-                    self.process_redap_dataset_entry(ctx, entry_id);
+                Item::RedapEntry(entry_id) => {
+                    self.process_redap_entry(ctx, entry_id);
                 }
-                Item::RedapTableEntry(entry_id) => self.process_redap_table_entry(ctx, entry_id),
             }
         }
     }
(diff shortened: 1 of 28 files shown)


crates/viewer/re_context_menu/src/lib.rs
```diff
@@ -409,7 +411,7 @@ trait ContextMenuAction {
     /// Process a single redap server.
     fn process_redap_server(&self, _ctx: &ContextMenuContext<'_>, _origin: &re_uri::Origin) {}
 
-    /// Process a single redap entry.
+    /// Process a single reda entry (dataset or table).
```
emilk:
```suggestion
    /// Process a single redap entry (dataset or table).
```

crates/viewer/re_redap_browser/src/entries.rs
```diff
@@ -100,21 +118,38 @@ impl Entries {
         egui_ctx: &egui::Context,
         origin: re_uri::Origin,
     ) -> Self {
-        let datasets = fetch_dataset_entries(connection_registry, origin);
+        let datasets_fut = fetch_dataset_entries(connection_registry.clone(), origin.clone());
```
emilk, line 121: what is a "fut"?

emilk: oh, a future

abey79: Yep, and a bright one, fully of entries 🎉

emilk: image

PR #10573, opened 2025-07-09
Improve table line height logic

emilk:
### Related
* Part of https://github.com/rerun-io/rerun/issues/5589

### What
This defines two table styles, `Dense` and `Spacious`. The result is that we fit more on the screen in the places where it matters:

image

But still have nice spacing in places where we want to:

image


emilk:
Future work:
* https://github.com/rerun-io/rerun/issues/10575

crates/viewer/re_ui/src/design_tokens.rs
```diff
@@ -33,6 +33,19 @@ impl AlertVisuals {
...
+pub enum TableStyle {
+    /// Used for presenting a lot of information to the user
+    /// without wasting vertical space,
+    /// like when showing log output.
+    #[default]
+    Dense,
+
+    /// Used when we want more margins and spacing,
+    /// e.g. in the catalog view.
+    Spacious,
```
Wumpf: maybe worth calling out that a prime motivation for using this is that there may be additional buttons etc.

Wumpf (approved):
tight! :)

PR #10579, opened 2025-07-10
Remove unused `ViewContextSystem::compatible_component_set`

grtlr:
### Related

* Part of: #8129 
* Part of: #6889 

### What

Turns out the `EntityDepthOffsets` system was still indicators driven. This replaces the use of indicators with `required_components`.


Wumpf:
how did this not fail the depth image comparision test Oo
I'll have a deeper look what's going on there

Wumpf:
I think we should kill this method. If someone really wants to do the optimization be my guest but it doesn't sound all that meaningful at that point
also context systems aren't picky about entities at all more or less by design

Issue #10190, opened 2025-06-10
Error when trying to build the Python SDK from source (MacOS)


pweids:
From a relatively clean MacOS install (Python v3.13.4, rust version v1.85, pixi v0.48.0, Node v24.2.0), I followed the instructions in BUILD.md

```
git lfs install
pixi run check-env
pixi run rerun # worked fine
pixi run py-build --release
```

The error I get is:

```
% pixi run py-build --release
✨ Pixi task (py-build in py): pixi run -e py py-build-common --release
Error:   × Failed to update PyPI packages for environment 'py'
  ├─▶ Failed to prepare distributions
  ├─▶ Failed to build `rerun-notebook @ file:///Users/paul/Development/rerun/rerun_notebook`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `hatchling.build.build_editable` failed (exit status: 1)

      [stdout]

      up to date, audited 87 packages in 527ms

      10 packages are looking for funding
        run `npm fund` for details

      3 vulnerabilities (2 moderate, 1 high)

      To address issues that do not require attention, run:
        npm audit fix

      To address all issues (including breaking changes), run:
        npm audit fix --force

      Run `npm audit` for details.

      > build
      > node build.mjs


      [stderr]
      INFO:hatch_jupyter_builder.utils:Running jupyter-builder
      INFO:hatch_jupyter_builder.utils:Building with hatch_jupyter_builder.npm_builder
      INFO:hatch_jupyter_builder.utils:With kwargs: {'npm': 'npm', 'build_cmd': 'build'}
      INFO:hatch_jupyter_builder.utils:Installing build dependencies with npm.  This may take a while...
      INFO:hatch_jupyter_builder.utils:> /Users/paul/Development/rerun/.pixi/envs/py/bin/npm install
      INFO:hatch_jupyter_builder.utils:> /Users/paul/Development/rerun/.pixi/envs/py/bin/npm run build
      ✘ [ERROR] Could not resolve "@rerun-io/web-viewer/inlined.js"

          src/js/widget.ts:7:7:
            7 │ } from "@rerun-io/web-viewer/inlined.js";
              ╵        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

        The module "./inlined.js" was not found on the file system:

          node_modules/@rerun-io/web-viewer/package.json:45:16:
            45 │       "import": "./inlined.js"
               ╵                 ~~~~~~~~~~~~~~

[...]

PR #10580, opened 2025-07-10
Fix pixi dev environment bootstrapping

Wumpf:
* Fix https://github.com/rerun-io/rerun/issues/10190

On fresh checkouts I always had to disable `rerun-notebook`, then run `js-build-base` to get the environment bootstrapped - `js-build-base` will cause a notebook build and a notebook build needs `js-build-base`
More often than not I just ended up with a changed pixi.toml that I'd always had to be careful not to check in

Ideally the pixi-built versions of those packages would just be stubs until we turn py-build equivalents, which would improve a bunch of other stuff in general.

This breaks running jupyter notebooks directly from pixi directly: i.e. `pixi run -e py jupyter notebook examples/notebook/cube.ipynb` won't run out of the box. But still worked fine for me  after running `pixi run py-build && pixi run py-build-notebook`.

(see also internal slack thread https://rerunio.slack.com/archives/C045J1Z7DU7/p1751226966091869)

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -598,7 +598,7 @@ typos = ">=1.16.20"
 cryptography = "==38.0.4"        # For `upload_image.py`
 google-cloud-storage = "==2.9.0" # For `upload_image.py`
 hatch = "==1.12.0"               # For `rerun_notebook`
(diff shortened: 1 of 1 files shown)


Wumpf:
drafted: trying to get rerun notebook from dev environment to run on my windows at all

abey79 (approved):
makes sense

PR #10582, opened 2025-07-10
Update notebook examples

Wumpf:
Ran through our notebook examples and did some touch up.
Made sure they all work with the build setup as in 
* https://github.com/rerun-io/rerun/pull/10580

### DIFF
--- a/examples/python/notebook/send_table.ipynb
+++ b/examples/python/notebook/send_table.ipynb
@@ -8,7 +8,7 @@
     "### How to run\n",
     "\n",
     "```sh\n",
-    "pixi run -e examples py-build\n",
+    "pixi run -e examples py-build-common\n",
     "pixi run -e examples py-build-notebook\n",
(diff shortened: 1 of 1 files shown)


examples/python/notebook/send_table.ipynb
```diff
@@ -8,9 +8,9 @@
     "### How to run\n",
     "\n",
     "```sh\n",
-    "# pixi run py-build-examples\n",
-    "# pixi run -e examples py-build-notebook\n",
-    "# pixi run -e examples jupyter notebook examples/python/notebook/send_table.ipynb\n",
+    "pixi run -e examples py-build\n",
```
jprochazk: `py-build` isn't a valid command in `-e examples`

Wumpf: oh you're right, I used `pixi run -e examples py-build-common` here and forgot to fix that

PR #10583, opened 2025-07-10
Remove `split_indicators` from `ChunkBatcher`

grtlr:
### Related

* Part of #6889.
* Part of #8129.

### What

Removes unnecessary logic that would split a chunk into two if indicators were present. Also adds a check to `rerun rrd verify` for indicator components.


Wumpf (approved):
lgtm!

PR #10581, opened 2025-07-10
Remove indicators from codegen, `rerun_py`, and `rerun_cpp`

grtlr:
### Related

* Part of #6889
* Part of #8129
* [x] Merge after #10572.

### What

Title. 🎉 

### TODO

* [x] full-check
* [x] No changes required on dataplatform side.

PR #10584, opened 2025-07-10
Removes remnants of indicators from dataframe queries and viewer

### LINKED ISSUE #8129: Tagged components milestone 2: no more indicators

grtlr:
### Related

* Part of #6889.
* Closes #8129.
* [x] Merge after #10583
* [x] Merge after #10581

### What

This removes the remaining reference to indicators from our codebase. Most notably this removes indicators from the dataframe queries and the UI.

This should be fine because:
* We drop indicator columns during Sorbet migration
* By default indicators where not shown anyways
* And they were omitted in queries by default too.

> [!IMPORTANT]
> There is some risk that this PR will make indicators appear again, so we should choose a good time to merge this PR. If they show up again then let's fix the root cause (i.e. migration logic).

### TODO

* [x] full-check
* [x] Dataplatform companion PR: https://github.com/rerun-io/dataplatform/pull/1234

### DIFF
--- a/crates/store/re_chunk/src/transport.rs
+++ b/crates/store/re_chunk/src/transport.rs
@@ -107,7 +107,6 @@ impl Chunk {
                         // Setting them all to `false` at least ensures they aren't written to the arrow metadata:
                         // TODO(#8744): figure out what to do here
                         is_static: false,
-                        is_indicator: false,
                         is_tombstone: false,
                         is_semantically_empty: false,
                     };
--- a/crates/store/re_chunk_store/src/dataframe.rs
+++ b/crates/store/re_chunk_store/src/dataframe.rs
@@ -163,14 +163,6 @@ pub struct QueryExpression {
     /// `view_contents`: [`QueryExpression::view_contents`]
     pub include_semantically_empty_columns: bool,
 
-    /// Whether the `view_contents` should ignore columns corresponding to indicator components.
-    ///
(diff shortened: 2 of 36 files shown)


crates/viewer/re_ui/src/design_tokens.rs
```diff
@@ -610,6 +610,10 @@ impl DesignTokens {
         self.table_row_height(table_style) - self.table_cell_margin(table_style).sum().y
     }
 
+    pub fn header_cell_margin(&self, _table_style: TableStyle) -> Margin {
+        Margin::symmetric(8, 6)
+    }
+
```
abey79: Is this an easter egg to test the reviewer? :)

grtlr: Ahaha, nah just a bad rebase. The diff should be much nicer now that https://github.com/rerun-io/rerun/pull/10581 is merged.

grtlr:
@rerun-bot full-check

Issue #10574, opened 2025-07-09
When opening viewer using a url it's not clear somethign is happening


jleibs:
https://github.com/user-attachments/assets/8ea9e31f-98d5-4e8c-acab-5f179b468cf4

We need some kind of loading page.

Wumpf:
Interestingly enough I _sometimes_ get first a load table screen. But then when rerunning it repros the behavior as demonstrated
Maybe due to prior selection

PR #10587, opened 2025-07-10
Show loading screen when starting viewer with a URL

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/10574 

### What

Previously, when starting the viewer with a data source attached we'd show the welcome screen (or data ui respectively in the notebook). This is irritating because it doesn't give any feedback the ongoing loading process.

https://github.com/user-attachments/assets/aba635c0-9807-456d-920b-b942a146ca4c

Design isn't great, might need some @gavrelina love later on.

* [x] test behavior from url locally
* [x] test for no change in behavior without a url
* [x] test for no behavior change when using spawn from sdk
* [x] test behavior in a notebook

### DIFF
--- a/crates/viewer/re_viewer/src/ui/welcome_screen/loading_data_ui.rs
+++ b/crates/viewer/re_viewer/src/ui/welcome_screen/loading_data_ui.rs
@@ -37,7 +37,11 @@ pub fn loading_text_for_data_sources(log_sources: &[Arc<SmartChannelSource>]) ->
                 return Some(format!("Connecting to {url} …"));
             }
 
-            SmartChannelSource::RrdWebEventListener | SmartChannelSource::JsChannel { .. } => {
+            SmartChannelSource::RrdWebEventListener => {
+                // Waiting on notebook data.
+            }
+
+            SmartChannelSource::JsChannel { .. } => {
                 return Some("Loading…".to_owned());
             }
 
--- a/crates/viewer/re_viewer/src/ui/welcome_screen/loading_data_ui.rs
+++ b/crates/viewer/re_viewer/src/ui/welcome_screen/loading_data_ui.rs
@@ -1,0 +1,56 @@
+use std::sync::Arc;
+
+use re_smart_channel::SmartChannelSource;
+use re_ui::{DesignTokens, UiExt as _};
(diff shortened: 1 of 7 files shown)


crates/viewer/re_viewer/src/ui/welcome_screen/loading_data_ui.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use egui::Widget as _;
+use re_smart_channel::SmartChannelSource;
+use re_ui::{DesignTokens, UiExt as _};
+
+/// Show a loading screen for when we are connecting to a data source.
+pub fn loading_data_ui(ui: &mut egui::Ui, loading_text: &str) {
+    ui.center("loading_data_ui_contents", |ui| {
+        ui.vertical_centered(|ui| {
+            egui::Spinner::new().size(100.0).ui(ui);
```
emilk: same thing, but more idiomatic

```suggestion
            ui.add(egui::Spinner::new().size(100.0));
```

crates/viewer/re_viewer/src/ui/welcome_screen/loading_data_ui.rs
```diff
@@ -0,0 +1,59 @@
+use std::sync::Arc;
```
emilk, line 1: We should maybe move these function into this file, or combine them, or something:

https://github.com/rerun-io/rerun/blob/a5cfa2afce2dab4f28d691385884bca969aaf2f5/crates/viewer/re_viewer/src/ui/top_panel.rs#L304-L331

https://github.com/rerun-io/rerun/blob/a5cfa2afce2dab4f28d691385884bca969aaf2f5/crates/viewer/re_viewer/src/ui/recordings_panel.rs#L51-L99

jleibs:
@Wumpf how are you testing this in the notebook? It has not been working locally for me.

Wumpf:
Works with the same invocation shared on Slack. The problem is that right now the selection state is serialized out which sometimes causes a dataplaform server to be selected on startup which suppresses the loading screen then.
https://github.com/rerun-io/rerun/pull/10591

PR #10590, opened 2025-07-10
Fix `VideoStream` stuttering on web

### LINKED ISSUE #10564: `VideoStream` stutters on web
I'm sure at some point on this journey this worked fine.

It may be caused by WebCodecs api giving us presentation time stamps that we can't match with the ones we have samples for - we more often do direct comparison of requested sample PTS than we used to, so kinda makes sense that we didn't have this issue before. I tried pinpointing when this first happened but it seems to be a while now; have to do python & web builds for each step which takes ages, so didn't fully bisect.
WebCodecs uses microsecond timestamps throughout the api, so I figured I could to a lookup of what we put in. But, fun fact, you can't: returned frames may have PTS for which we never put in any chunks. And as I figured out it's not just a matter of rounding to integers
I think what's happening is that the PTS information may only be used for key frames is then codec depended extrapolated since chunks may be arbitrary bundles of data for which a decoder may generate arbitrary number of frames if it pleases so

Wumpf:
### Related

* Fixes #10564

### What

Fix stutter issues that are caused by unexpected presentation timestamp mismatch.
The main source for this was issues with our microsecond conversion: floats are cut off to integers when going through WebCodec and we blindly trusted the values on the other end and we really need the original PTS in our player.
When addressing this by pushing frame infos to the output callback, I originally didn't consider that:
* decoder reset may not recreate the output callback, therefore causing spurious old frame infos to be in the callback's frame-info queue. It's important to reset that queue as well
* either due to user input error or rounding issues, we may have several frame infos with the same micro-second timestamp. We have to handle this gracefully

Fixes end-of-video issues when continuing a video after all
* WebCodec insists that the next frame is a key frame, if not things go ugly
* our ffmpeg decoder was just broken for a frame sometimes before auto-resetting

Fixes unhandled WebCodec abort exception handling for when `flush` is followed by `reset`.

---

* [x] full check
* [x] round of testing on Mac, including Safari

---

### Earlier misconceptions

For quite a while I thought that this part of the spec was violated by browsers
> Let timestamp and duration be the timestamp and duration from the EncodedVideoChunk associated with output.
(https://www.w3.org/TR/webcodecs/#output-videoframes)
However, that is *not* the case for all I can tell now. All browsers do pass the _integer part_ of the timestamp faithfully through. The other errors we had on our side caused some confusion, but with more debug output I cleared out those false conclusions.

### Can we detect number of frames in packages?

This is a potential error mode that I thought might be happening for a while, but didn't. Still it would be interesting to detect this:

I tried to evolve our GOP detection code to more arbitrary package inspection and then warn about frame counts - code [here](https://github.com/rerun-io/rerun/commit/7a6e81d7b946cba0f1149384ec599bde8a9abe3c).
Unfortunately, my "detection" says there's 5 frames in each package of `video_stream_synthetic.py`. We don't get that many frames back from any decoder and we also also logged the exact number of frame-samples we expected (120).
The problem seems to be that the packages are sliced internally - the nal type I checked is indeed dubbed `SliceLayerWithoutPartitioningNonIdr`. Curiously `SliceLayerWithoutPartitioningIdr` on the other hand only ever shows once :face_with_raised_eyebrow: 
I double checked by hexdumping the last frame of `video_stream_synthetic.py` - it's just 378 bytes and indeed we have 5x `00 00 01 41` which each indicates an non-idr slice. I suspect that depending on codec metadata varying number of slices are expected per frame.

### DIFF
--- a/crates/utils/re_video/src/decode/mod.rs
+++ b/crates/utils/re_video/src/decode/mod.rs
@@ -172,7 +172,8 @@ pub trait AsyncDecoder: Send + Sync {
     /// Called after submitting the last chunk.
     ///
     /// Should flush all pending frames.
-    /// If new chunks are submitted after this, the decoder has to be reset.
+    /// If you plan on sending more chunks after calling `end_of_video`,
+    /// you MUST call [`Self::reset`] FIRST.
     ///
     /// Implementation note:
     /// As of writing there's two decoders that have requirements on what happens for new frames after `end_of_video`
--- a/crates/utils/re_video/src/decode/mod.rs
+++ b/crates/utils/re_video/src/decode/mod.rs
@@ -172,7 +172,12 @@ pub trait AsyncDecoder: Send + Sync {
     /// Called after submitting the last chunk.
     ///
     /// Should flush all pending frames.
-    /// The next submitted chunk is guaranteed to be a key frame.
+    /// If new chunks are submitted after this, the decoder has to be reset.
+    ///
+    /// Implementation note:
+    /// As of writing there's two decoders that have requirements on what happens for new frames after `end_of_video`
+    /// * WebCodec: The next submitted chunk has to be a key frame.
+    /// * FFmpeg-executable: We've shut down stdin, thus we need to restart the process. Doing this without the full context of `reset` is not possible right now.
     fn end_of_video(&mut self) -> Result<()> {
         Ok(())
     }
--- a/crates/utils/re_video/src/decode/mod.rs
+++ b/crates/utils/re_video/src/decode/mod.rs
@@ -172,6 +172,7 @@ pub trait AsyncDecoder: Send + Sync {
     /// Called after submitting the last chunk.
     ///
     /// Should flush all pending frames.
+    /// The next submitted chunk is guaranteed to be a key frame.
     fn end_of_video(&mut self) -> Result<()> {
         Ok(())
     }
@@ -391,9 +392,11 @@ pub struct FrameInfo {
     /// None = unknown.
     pub frame_nr: Option<u32>,
 
-    /// Time at which this sample appears in the frame stream, in time units.
+    /// Time at which this frame appears in the frame stream, in time units.
     ///
     /// The frame should be shown at this time.
+    /// We expect this timestamp to be identical with a the presentation timestamp of the [`crate::Chunk`]
+    /// which is associated with this frame.
     /// Often synonymous with `composition_timestamp`.
     ///
     /// `decode_timestamp <= presentation_timestamp`
--- a/crates/utils/re_video/src/decode/webcodecs.rs
+++ b/crates/utils/re_video/src/decode/webcodecs.rs
@@ -376,8 +376,8 @@ fn init_video_decoder(
                     (None, None) => {
                         // This unfortunately happens every now and then for our video stream demo.
                         re_log::debug_once!("WebCodec decoded video frame for which we didn't have any corresponding frame info.
-                                This happens when the decoder emits more frames than we expected.
-                                This is due to interpolation in the decoder or if a video sample corresponds to more than one video frame which Rerun doesn't support."
+This happens when the decoder emits more frames than we expected which can occur if the sample
+contains more frames than expected (invalid input data) or the decoder interpolates frames unexpectedly."
                             );
                         return;
                     }
--- a/crates/utils/re_video/src/decode/webcodecs.rs
+++ b/crates/utils/re_video/src/decode/webcodecs.rs
@@ -181,11 +181,14 @@ impl AsyncDecoder for WebVideoDecoder {
         // > Let timestamp and duration be the timestamp and duration from the EncodedVideoChunk associated with output.
         //
         // However, in practice, the timestamps received are often off by more than can be just explained with
-        // the wasm->js->wasm transition (which is bound to mess with accuracy) and the fact that
+        // the wasm->js->wasm transition (which is bound to mess with accuracy a little bit) and the fact that
         // WebCodec specifies timestamps to be internally represented as i64.
         // In chrome this is likely related to this bug:
         // https://issues.chromium.org/issues/40925070
         // (comments describe that decoders may mess with the timestamps)
+        //
+        // Timestamps are never fully made up: if we add a large offset and remove it on receive again,
+        // we will never fall below that offset, i.e. it is passed through.
         //
         // Therefore we make a best-effort attempt:
         // * use whole numbered microseconds since first frame
--- a/crates/utils/re_video/src/decode/webcodecs.rs
+++ b/crates/utils/re_video/src/decode/webcodecs.rs
@@ -359,9 +359,7 @@ fn init_video_decoder(
                         web_timestamp_us,
                         frame_info,
                     }) => {
-                        let infos = pending_frame_infos
-                            .entry(web_timestamp_us)
-                            .or_insert(smallvec::SmallVec::default());
+                        let infos = pending_frame_infos.entry(web_timestamp_us).or_default();
                         infos.push(frame_info);
                     }
 
--- a/crates/utils/re_video/src/decode/webcodecs.rs
+++ b/crates/utils/re_video/src/decode/webcodecs.rs
@@ -220,8 +220,6 @@ impl AsyncDecoder for WebVideoDecoder {
         if let Some(duration) = video_chunk.duration {
             let duration_micros = 1e-3 * duration.duration(self.timescale).as_nanos() as f64;
             web_chunk.set_duration(duration_micros);
-        } else {
-            web_chunk.set_duration(1.0 / 30.0 * 1000.0 * 1000.0);
         }
 
         let web_chunk = EncodedVideoChunk::new(&web_chunk)
--- a/crates/utils/re_video/src/decode/webcodecs.rs
+++ b/crates/utils/re_video/src/decode/webcodecs.rs
@@ -1,9 +1,11 @@
-use std::{collections::BTreeMap, sync::Arc};
+use std::{collections::hash_map::Entry, sync::Arc};
 
+use ahash::HashMap;
 use crossbeam::channel::{Receiver, Sender};
 use js_sys::{Function, Uint8Array};
 use once_cell::sync::Lazy;
 use re_mp4::StsdBoxContent;
+use smallvec::SmallVec;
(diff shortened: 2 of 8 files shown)


crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -669,7 +678,18 @@ fn is_significantly_behind(
             // We're technically not catching up, but we may as well behave as if we are.
             return true;
         };
-        if current_sample.presentation_timestamp == decoded_frame_pts {
+        if current_sample.presentation_timestamp <= decoded_frame_pts {
```
Wumpf, line 683: this is for robustness only and shouldn't make any difference

crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -147,17 +167,61 @@ impl AsyncDecoder for WebVideoDecoder {
...
+        // The spec also marks it as entirely optional, but does not specify whether it may affect the decoding itself.
+        // Given that we err on the side of providing too much than too little information.
         if let Some(duration) = video_chunk.duration {
-            let duration_millis = 1e-3 * duration.duration(self.timescale).as_nanos() as f64;
-            web_chunk.set_duration(duration_millis);
+            let duration_micros = 1e-3 * duration.duration(self.timescale).as_nanos() as f64;
+            web_chunk.set_duration(duration_micros);
+        } else {
+            web_chunk.set_duration(1.0 / 30.0 * 1000.0 * 1000.0);
         }
```
Wumpf: remove this else branch again, that was experimental

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -452,6 +455,15 @@ impl VideoPlayer {
         else if requested.gop_idx < last_requested.gop_idx {
             self.reset(video_description)?;
         }
+        // Previously signaled the end of the video, but encountering frames that are newer than the last enqueued.
+        else if self.signaled_end_of_video
```
Wumpf, line 457: note this is two changes:
* moved up so it will be hit even if below `else if` is also true (which then may or may not reset)
* check last enqueued against last sample in the description

crates/utils/re_video/src/decode/mod.rs
```diff
@@ -172,6 +172,12 @@ pub trait AsyncDecoder: Send + Sync {
     /// Called after submitting the last chunk.
     ///
     /// Should flush all pending frames.
+    /// If new chunks are submitted after this, the decoder has to be reset.
```
emilk:
```suggestion
    /// If you plan on sending more chunks after calling `end_of_video`,
    /// you MUST call [`Self::reset`] FIRST.
```

crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -41,10 +42,15 @@ impl std::ops::Deref for WebVideoFrame {
 
 pub struct WebVideoDecoder {
     codec: VideoCodec,
+
     timescale: Timescale,
+    first_frame_pts: Time,
```
emilk, line 58: Should we just store the `VideoDataDescription` here instead? Or will that become out-dated in case of streams?

crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -234,36 +303,94 @@ static IS_SAFARI: Lazy<bool> = Lazy::new(|| {
...
+                            bigger
+                        }
+                    }
+                    (Some(smaller), None) => smaller,
+                    (None, Some(bigger)) => bigger,
+                    (None, None) => {
+                        // This unfortunately happens every now and then for our video stream demo.
+                        re_log::debug_once!("WebCodec decoded video frame for which we didn't have any corresponding frame info.
+                                This happens when the decoder emits more frames than we expected.
+                                This is due to interpolation in the decoder or if a video sample corresponds to more than one video frame which Rerun doesn't support."
```
Wumpf: broken whitespace

crates/utils/re_video/src/decode/webcodecs.rs
Wumpf: yeah you guessed right, it will outdate. Also it's a fairly heavy object with all the sample metadata

crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -147,17 +167,62 @@ impl AsyncDecoder for WebVideoDecoder {
...
+        let web_chunk = EncodedVideoChunkInit::new(&data, web_timestamp_us as _, type_);
+
+        // Empirically, decoders don't care about the presence of the duration field.
+        // The spec also marks it as entirely optional, but does not specify whether it may affect the decoding itself.
+        // Given that we err on the side of providing too much than too little information.
         if let Some(duration) = video_chunk.duration {
-            let duration_millis = 1e-3 * duration.duration(self.timescale).as_nanos() as f64;
-            web_chunk.set_duration(duration_millis);
+            let duration_micros = 1e-3 * duration.duration(self.timescale).as_nanos() as f64;
+            web_chunk.set_duration(duration_micros);
```
emilk, line 230: Maybe round here too for consistency?
```suggestion
            let duration_micros = duration.duration(self.timescale).as_micros() as f64;
            web_chunk.set_duration(duration_micros.round());
```

crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -234,36 +322,94 @@ static IS_SAFARI: Lazy<bool> = Lazy::new(|| {
...
         Closure::wrap(Box::new(move |frame: web_sys::VideoFrame| {
-            // We assume that the timestamp returned by the decoder is in time since start,
-            // and does not represent demuxed "raw" presentation timestamps.
-            let presentation_timestamp =
-                Time::from_micros(frame.timestamp().unwrap_or(0.0), timescale);
-            let duration = frame.duration().map(|d| Time::from_micros(d, timescale));
+            loop {
+                match frame_info_tx.try_recv() {
+                    Ok((web_timestamp_us, frame_info)) => {
+                        pending_frame_info.insert(web_timestamp_us, frame_info);
```
emilk: You should check that you're not overwriting something in `pending_frame_info` that has the same `web_timestamp_us`

crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -234,36 +322,94 @@ static IS_SAFARI: Lazy<bool> = Lazy::new(|| {
...
+                match (timestamp_smaller_equal, timestamp_bigger) {
+                    (Some(smaller), Some(bigger)) => {
+                        if web_timestamp_us - smaller < bigger - web_timestamp_us {
+                            smaller
+                        } else {
+                            bigger
+                        }
+                    }
+                    (Some(smaller), None) => smaller,
+                    (None, Some(bigger)) => bigger,
```
emilk: Would be interesting to add debug-logging to all these branches

crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -234,36 +322,94 @@ static IS_SAFARI: Lazy<bool> = Lazy::new(|| {
...
+                    .next()
+                    .map(|(k, _)| *k);
+
+                match (timestamp_smaller_equal, timestamp_bigger) {
+                    (Some(smaller), Some(bigger)) => {
+                        if web_timestamp_us - smaller < bigger - web_timestamp_us {
+                            smaller
+                        } else {
+                            bigger
+                        }
```
emilk: We talked about this a bit, but for posterity:

On theory is that the decoder uses smearing to cover up hickups. So if a frame gets decoded late, all PTS:es for it and subsequent frames are increased to produce a smooth playback at a higher rate to catch up, but without showing all frames at the same time.

Example: we're decoding a 30 FPS video stream and for some reason we fall behind 0.5 seconds (15 frames). What to do?
* Skip 15 frames? But the user might miss something important!
* Play back the 15 frames extremely quickly? It will look unnatural

So the browser may decide to play back the 15 frames plus the next 45 frames at 1.33x speed. So for two seconds the video moves slightly faster. Good compromise.

If this theory hold, then the PTS:ed should only be wrong in one direction: the should come out greater (or equal) to the ones inserted.

crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -234,36 +322,94 @@ static IS_SAFARI: Lazy<bool> = Lazy::new(|| {
...
+                            );
+                        return;
+                    }
+                }
+            };
+
+            if lookup_timestamp_us as f64 != web_timestamp_us_raw {
+                re_log::trace!(
+                    "WebCodec's decoded frame hat a PTS of {web_timestamp_us_raw}us, but we the closest we had was {lookup_timestamp_us}us. Abs difference: {}us",
+                    (web_timestamp_us_raw - lookup_timestamp_us as f64).abs()
```
emilk: signed difference is much more interesting (especially: do they have the same sign?)

crates/utils/re_video/src/decode/webcodecs.rs
Wumpf: as_micros returns integers already, I'll check if it does correct rounding

Wumpf: it doesn't :/

crates/utils/re_video/src/decode/webcodecs.rs
Wumpf: this turned out critical and caused a lot of confusion on my side

Wumpf:
@rerun-bot full-check

crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -234,36 +333,93 @@ static IS_SAFARI: Lazy<bool> = Lazy::new(|| {
...
+                return;
+            };
+            // WebCodec timestamps are internally represented as i64 according to the spec.
+            // Any floating point part would be a violation of the spec.
+            let web_timestamp_us = web_timestamp_us_raw as u64;
+
+            match pending_frame_infos.entry(web_timestamp_us) {
+                Entry::Occupied(mut entry) => {
+                    let infos = entry.get_mut();
+                    debug_assert!(!infos.is_empty(), "Frame info lists should never be empty.");
```
Wumpf: handle this gracefully - right now remove(0) just panics. Bad!

jprochazk (approved):
Pair-reviewed on Slack, LGTM

PR #10596, opened 2025-07-11
Add `rerun rrd concat` command to concatenate chunks into new recording

### LINKED ISSUE #8860: It should be possible to force-merge two recordings even if they have different recording_ids
Currently if two recordings have different recording ids there is no way to merge the data from them together into a single view.

At a minimum this seems like a a useful flag for the `rerun rrd merge` command.
Would also be nice if the viewer had a way to select multiple recordings and then right-click -> merge.

grtlr:
### Related

* Closes #8860.

### What

This merges the chunks of several `.rrd/.rbl`s into a single `.rrd` recording.

### TODO

* [ ] Consider sharing code with `merge_concat.rs`
* [ ] Figure out blueprint handling.
  * Do we omit blueprints?
  * Only take the first one?
  * Merge them all using heuristic / "deep packet inspection"?

### DIFF
--- a/crates/top/rerun/src/commands/rrd/concat.rs
+++ b/crates/top/rerun/src/commands/rrd/concat.rs
@@ -1,0 +1,209 @@
+use std::io::{IsTerminal as _, Write as _};
+
+use anyhow::Context as _;
+use itertools::Either;
+use re_chunk_store::ChunkStoreConfig;
+use re_entity_db::EntityDb;
+use re_sdk::ApplicationId;
+
+use crate::commands::read_rrd_streams_from_file_or_stdin;
+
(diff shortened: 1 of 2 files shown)


Wumpf:
Why did you add a new command rather than extending `rerun rrd merge`?

grtlr:
Copied from Slack:

> The reason I think a different command makes sense is that we probably want to provide options for setting the application id, recording id, and blueprint handling. All of these don't make sense in the context of `merge`. We could introduce a subcommand though.
>
> I also think what we're doing in this PR is more aptly named `rerun rrd merge`. With the previous functionality being more like `append / concat` (so basically the other way around).

grtlr:
Superseded by

* #10607.

PR #10594, opened 2025-07-10
Compaction: performance & quality improvements

teh-cmc:
Title.

In particular this:
* Fixes a nasty `O(n**2)` that slipped in, bringing back performance to sub-linear levels.
* Introduction compaction passes in the CLI, which is super important to get the most out of compaction in offline scenarios.

```sh
$ pixi run rerun-release rrd compact problematic.rrd -o compacted.rrd --max-rows 999999999 --max-bytes 1048576 --num-pass 50

# [...]

[2025-07-10T17:44:58Z INFO  rerun::commands::rrd::merge_compact] processed 1289999 messages so far, current speed is 19312.92 msg/s
[2025-07-10T17:44:58Z INFO  rerun::commands::rrd::merge_compact] processed 1299999 messages so far, current speed is 19985.97 msg/s
[2025-07-10T17:44:59Z INFO  rerun::commands::rrd::merge_compact] processed 1309999 messages so far, current speed is 19466.20 msg/s
[2025-07-10T17:44:59Z INFO  rerun::commands::rrd::merge_compact] processed 1319999 messages so far, current speed is 18687.46 msg/s
[2025-07-10T17:45:00Z INFO  rerun::commands::rrd::merge_compact] processed 1329999 messages so far, current speed is 17226.69 msg/s
[2025-07-10T17:45:00Z INFO  rerun::commands::rrd::merge_compact] processed 1339999 messages so far, current speed is 18906.00 msg/s
[2025-07-10T17:45:01Z INFO  rerun::commands::rrd::merge_compact] processed 1349999 messages so far, current speed is 17622.72 msg/s
[2025-07-10T17:45:01Z INFO  rerun::commands::rrd::merge_compact] processed 1359999 messages so far, current speed is 18320.93 msg/s
[2025-07-10T17:45:02Z INFO  rerun::commands::rrd::merge_compact] processed 1369999 messages so far, current speed is 19278.27 msg/s
[2025-07-10T17:45:03Z INFO  rerun::commands::rrd::merge_compact] processed 1379999 messages so far, current speed is 19101.29 msg/s
[2025-07-10T17:45:03Z INFO  rerun::commands::rrd::merge_compact] processed 1389999 messages so far, current speed is 20092.77 msg/s
[2025-07-10T17:45:04Z INFO  rerun::commands::rrd::merge_compact] processed 1399999 messages so far, current speed is 19879.84 msg/s
[...]

PR #10600, opened 2025-07-11
Remove hacks to half-buggily-support undef behavior (duped Row IDs)

teh-cmc:
See https://github.com/rerun-io/rerun/pull/10594#discussion_r2198334520:
> Note: this is technically incorrect because it doesn't account for the vector within `chunk_ids_per_min_row_id`, except that vector doesn't make any sense to begin with. I'm going to remove it entirely tomorrow: going out of our way to pseudo-support undefined behavior (duplicated Row IDs) is unproductive at best, non-sensical at worst.

### DIFF
--- a/crates/store/re_chunk_store/src/writes.rs
+++ b/crates/store/re_chunk_store/src/writes.rs
@@ -357,9 +357,7 @@ impl ChunkStore {
 
         self.chunks_per_chunk_id.insert(chunk.id(), chunk.clone());
         self.chunk_ids_per_min_row_id
-            .entry(row_id_range.0)
-            .or_default()
-            .push(chunk.id());
+            .insert(row_id_range.0, chunk.id());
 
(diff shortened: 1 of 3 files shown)


crates/store/re_chunk_store/src/writes.rs
```diff
@@ -357,9 +357,7 @@ impl ChunkStore {
 
         self.chunks_per_chunk_id.insert(chunk.id(), chunk.clone());
         self.chunk_ids_per_min_row_id
```
Wumpf: hmm should we warn if something is already in there? 🤔

Wumpf (approved):
👍

crates/store/re_chunk_store/src/writes.rs
teh-cmc: eh, why not

PR #10602, opened 2025-07-11
Add `open_url`/`close_url` notebook APIs

jprochazk:
cc @ntjohnson1 as this replaces the `set_active_partition_url` API

### DIFF
--- a/crates/viewer/re_viewer/src/web.rs
+++ b/crates/viewer/re_viewer/src/web.rs
@@ -209,6 +209,8 @@ impl WebHandle {
             app.command_sender.clone(),
         ) {
             app.add_log_receiver(rx);
+            app.egui_ctx
+                .request_repaint_after(std::time::Duration::from_millis(10));
         }
     }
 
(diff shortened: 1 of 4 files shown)


ntjohnson1 (commented):
Can you add the workflow in the PR description.

If we do
```python
url_a = <>
url_b = <>
viewer = Viewer(<>)
viewer.open_url(url_a)
# wait to show a
time.sleep(<>)
viewer.open_url(url_b)
# wait to show b
viewer.open_url(url_a)
```

Does this do the right thing by showing a, then b, then a? I don't know the next level deeper enough to know how opening the same url twice behaves. If we can have a bunch of things open my first attempt in updating my code would be to open ALL my urls, then call open a second time to bring a preloaded url to the main view.

jprochazk:
> Does this do the right thing by showing a, then b, then a?

There are no uniqueness constraints on data sources, so this will open A, then B, then A again, resulting in 3 open recordings. It will fetch A's data twice.

> If we can have a bunch of things open my first attempt in updating my code would be to open ALL my urls, then call open a second time to bring a preloaded url to the main view.

We don't have a way to activate a recording by its URL. If you want the exact same behavior as before, you'd do

```python
viewer = Viewer()
current_url = None

# ...
current_url = dataset.partition_url(...)
viewer.open_url(current_url)

# ... later
viewer.close_url(current_url)
current_url = dataset.partition_url(...)
viewer.open_url(current_url)
```

But there should be no harm in leaving the older partitions open, no? So you'd have this instead:

```python
viewer = Viewer()

viewer.open_url(dataset.partition_url(...))

# later, open another one
viewer.open_url(dataset.partition_url(...))
```

PR #10605, opened 2025-07-11
`rrd stats`: add transport-layer only fast path

teh-cmc:
Mostly useful for folks looking into storage-related matters: much much faster to compute, but less high-level information.


Wumpf (approved):
makes sense! 👍

PR #10607, opened 2025-07-11
`rerun rrd route` command to manipulate `.rrd`s on transport level

grtlr:
### Related

* CLI part of #8860.
* Supersedes #10596.

### What

This adds a new `rerun rrd route` command that can be used to route one or multiple `.rrd`s to a new one (via `--output`) or to stdout.

This provides a concise way to merge recordings using the CLI. In future versions, we also want to add that functionality to the UI.

> [!IMPORTANT]
> The `route` command will never decode the Arrow data, therefore no migration or verification will happen, which is good for performance. If you still want migration, merging, and compaction, the result can be piped into `rerun rrd merge`.

Here is an example of the interface, which pipes directly to the viewer:

```sh
rerun rrd route \
  --application-id <new-app-id> \
  --recording-id <new-recording-id> \
  a.rrd b.rrd | rerun -
```

Or you can write the `.rrd` back out:

```sh
rerun rrd route \
  --application-id <new-app-id> \
  --recording-id <new-recording-id> \
  --output <result.rrd> \
  a.rrd b.rrd
```


crates/top/rerun/src/commands/rrd/route.rs
```diff
@@ -0,0 +1,186 @@
...
+    continue_on_error: bool,
+    writer: W,
+    receiver: &Receiver<(InputSource, anyhow::Result<Msg>)>,
+) -> anyhow::Result<()> {
+    re_log::info!("processing input…");
+    let mut num_total_msgs = 0;
+    let mut num_unexpected_msgs = 0;
+    let mut num_blueprints_msgs = 0;
+
+    // TODO(grtlr): encoding should match the original
```
Wumpf: can we though for the general case? It could be that we are able to read an old version but no longer write it

crates/top/rerun/src/commands/rrd/route.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    encoder.finish()?;
+
+    re_log::info_once!(
+        "Processed {num_total_msgs} messages, dropped {num_blueprints_msgs} blueprint messages, and encountered {num_unexpected_msgs} unexpected messages."
+    );
+    Ok(())
+}
+
+fn is_blueprint(msg: &Msg) -> bool {
```
Wumpf: more like `is_blueprint_command`? blueprint data is just fine

crates/top/rerun/src/commands/rrd/route.rs
```diff
@@ -0,0 +1,186 @@
...
+        let mut is_success = true;
+
+        match res {
+            Ok(mut msg) => {
+                num_total_msgs += 1;
+
+                if is_blueprint(&msg) {
+                    num_blueprints_msgs += 1;
+                    continue;
+                }
```
Wumpf: I think we should write something about this behavior in the command's doc

Wumpf (approved):
neat

Wumpf:
cli.md needs updating

crates/top/rerun/src/commands/rrd/route.rs
grtlr: This TODO is inherited from @teh-cmc's `rrd stats` work. I guess the problem is that `read_raw_rrd_streams_from_file_or_stdin` does not really expose that functionality yet.

Wumpf (approved):

PR #10611, opened 2025-07-12
Fix crate publishing caused by `include_bytes` outside of crate

Wumpf:
See build failure here https://github.com/rerun-io/rerun/actions/runs/16236442156/job/45847267954
We accidentally include bytes from outside of the crate which makes it fail to build upon publish.

Removing `RERUN_EXPERIMENTAL_TABLE` altogether, no longer needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8748,7 +8748,6 @@ dependencies = [
  "re_query",
  "re_renderer",
  "re_smart_channel",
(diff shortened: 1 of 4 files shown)


Wumpf:
it compiles! Don't think anything about this is controversial, so pushing to main so I can do another alpha :)

Issue #10255, opened 2025-06-16
Fix misleading GUI hint regarding inheritance of visible time range from parent entities


leonmkim:
**Describe the bug**

![Image](https://github.com/user-attachments/assets/091a0629-5004-4a0c-8134-079678cdef81)

GUI hint for visible time range suggests that "settings are inherited from the parent entity" which as of `0.23.3` is no longer the case. Only the enclosing view properties are propagated. 

**To Reproduce**
1. Given a view of logs of at least one parent-child entity:
2. Override the visible time range of the parent entity 
3. Child entity of parent does not inherit the overriden visible time range

**Expected behavior**
In step 3. of above, I expected all children of the parent to inherit the overriden visible time range settings.

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04

**Rerun version**
`0.23.3`

**Additional context**
Besides fixing the GUI hint, it would be great if parent-child inheritance of visible time range settings could be supported again!

PR #10618, opened 2025-07-13
Fix visible time range documentation on property inheritance

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/10255

### What

image

image


Wumpf:
Used to be correct at some point, but we changed behavior (don't remember when and if it ever hit a stable release)

PR #10619, opened 2025-07-14
Fix `FilterIsNotNull` toggle in dataframe view selection panel

### LINKED ISSUE #10498: Cannot activate "Filter rows where column is not null"
In the dataframe view, this option is perma-disabled, even when clicking on it:

Image

grtlr:
### Related

* Closes #10498.

### What

This PR fixes some problems with the `FilterIsNull` configuration via the selection panel. There were two issues:

* The `active`/`non-active` change was not properly tracked to trigger blueprint changes.
* The combo box did not really account for `None` values, which could lead to writing the UI placeholder `"-"` as a component.

### DIFF
--- a/crates/viewer/re_view_dataframe/src/view_query/ui.rs
+++ b/crates/viewer/re_view_dataframe/src/view_query/ui.rs
@@ -6,13 +6,10 @@ use re_log_types::{
     EntityPath, ResolvedTimeRange, TimeInt, TimeType, Timeline, TimelineName, TimestampFormat,
 };
 use re_sorbet::ColumnSelector;
-use re_types::ComponentIdentifier;
 use re_types::blueprint::components;
 use re_ui::{TimeDragValue, UiExt as _, list_item};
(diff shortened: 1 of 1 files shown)


Wumpf (approved):
lgtm! haven't tested it myself yet (because web build isn't ready yet 😄 ) but looks straight forward!

Wumpf (changes requested):
there's still something wrong here
video in a sec:

Wumpf:
https://github.com/user-attachments/assets/859f1b4d-3acd-491b-9468-ab2d1d25252a

grtlr:
Nice catch! There were two things at play here:
* A bug where I didn't check for changed entity path in `FilterIsNot`.
* The (now) missing indicator on `Transform3D` always leading to a placeholder `"-"`, although data being present at an entity path.

Every time there is a placeholder the UI reacts in the way shown in your video, which now only happens if there is no data on an entity (and with my changes that only happens for a bad view content selection), just as in `v0.23`.

Issue #10542, opened 2025-07-08
Adjust default micro-batcher settings


jleibs:
Our default micro-batcher settings are creating pathological RRDs, which makes things much harder for us than they need to me.

Exa: a 10GB RRD provided by a design partner contained 1.5M chunks

We should pick new settings in a principled way, which means actually doing some proper analysis around trade-offs for end-to-end latency vs performance.

We may need to make the settings more discoverable for users that want to return back to a lower-latency paradigm.

PR #10620, opened 2025-07-14
Source default batcher settings from sink

Wumpf:
### Related

* Closes https://github.com/rerun-io/rerun/issues/10542
* Alternative to https://github.com/rerun-io/rerun/pull/10550

### What

Makes the default flush tick 200ms for file & memory sinks and leaves it at 8ms for grpc sinks.

To facilitate this, I introduced per sink batcher config defaults and had to make the batcher configuration updatable (since Python always first creates a recording stream with a memory sink and then patches it in later). Unfortunately not all config optionscan be updated after the fact, but I think the solution here is good enough for what is practically needed.

### DIFF
--- a/crates/store/re_chunk/src/batcher.rs
+++ b/crates/store/re_chunk/src/batcher.rs
@@ -55,7 +55,8 @@ pub struct BatcherHooks {
     ///
     /// See [`re_log_types::ArrowRecordBatchReleaseCallback`] for more information.
     //
-    // TODO(#6412): probably don't need this anymore.
+    // As of writing this is used to handle memory that's shared from Python with the
+    // Batcher thread - we need special handling for release of potentially Python owned memory.
     pub on_release: Option<re_log_types::ArrowRecordBatchReleaseCallback>,
 }
 
@@ -146,13 +147,19 @@ impl Default for ChunkBatcherConfig {
 impl ChunkBatcherConfig {
     /// Default configuration, applicable to most use cases.
     pub const DEFAULT: Self = Self {
-        flush_tick: Duration::from_millis(8), // We want it fast enough for 60 Hz for real time camera feel
-        flush_num_bytes: 1024 * 1024,         // 1 MiB
+        flush_tick: Duration::from_millis(200),
+        flush_num_bytes: 1024 * 1024, // 1 MiB
         flush_num_rows: u64::MAX,
         chunk_max_rows_if_unsorted: 256,
         max_commands_in_flight: None,
         max_chunks_in_flight: None,
         hooks: BatcherHooks::NONE,
+    };
+
+    /// Low-latency configuration, preferred when streaming directly to a viewer.
+    pub const LOW_LATENCY: Self = Self {
+        flush_tick: Duration::from_millis(8), // We want it fast enough for 60 Hz for real time camera feel
+        ..Self::DEFAULT
     };
 
     /// Always flushes ASAP.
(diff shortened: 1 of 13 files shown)


crates/store/re_chunk/src/batcher.rs
```diff
@@ -55,7 +55,8 @@ pub struct BatcherHooks {
     ///
     /// See [`re_log_types::ArrowRecordBatchReleaseCallback`] for more information.
     //
-    // TODO(#6412): probably don't need this anymore.
+    // As of writing this is used to handle memory that's shared from Python with the
+    // Batcher thread - we need special handling for release of potentially Python owned memory.
```
teh-cmc: > we need special handling for release of potentially Python owned memory

We already have that: look for `ALL_RECORDINGS` in the python bridge, it takes care of all the Python shared memory shenanigans, hence this TODO.

```suggestion
    // TODO(#6412): probably don't need this anymore.
```

crates/store/re_log_types/src/arrow_msg.rs
```diff
@@ -14,7 +14,8 @@ use arrow::array::RecordBatch as ArrowRecordBatch;
 /// every instance.
 /// It is up to the callback implementer to handle this, if needed.
 //
-// TODO(#6412): probably don't need this anymore.
+// As of writing this is used to handle memory that's shared from Python with the
+// Batcher thread - we need special handling for release of potentially Python owned memory.
```
teh-cmc: ditto

teh-cmc:
```suggestion
// TODO(#6412): probably don't need this anymore.
```

teh-cmc (approved):
Neat.

That's both pretty tricky and very sensitive to regressions though, a few tests would be welcome:
* Make sure setting the env does override the per-sink logic and vice-versa
* Make sure updating the sink during logging does update the conf properly etc

crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -3069,4 +3069,149 @@ mod tests {
...
+            unsafe {
+                std::env::remove_var(self.key);
+            }
+        }
+    }
+
+    const CONFIG_CHANGE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
+
+    #[test]
+
```
teh-cmc:
```suggestion
```

crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -3069,4 +3069,149 @@ mod tests {
...
+            ChunkBatcherConfig {
+                flush_num_bytes: 456,
+                ..injected_config
+            },
+        );
+    }
+
+    #[test]
+    fn test_explicit_batcher_config() {
+        // This environment variable should still override the explicit config.
```
teh-cmc: Hmmm, that feels very weird to me. The expectation (ingrained from every other piece of software) is that explicit config in code overrides explicit config in environment which in turn overrides implicit defaults.

I.e. `RecordingStream::batcher_config` > env > `LogSink::default_batcher_config`

teh-cmc: (as an example, we have a lot of tests that hardcode micro-batching settings because they expect some very specific behavior, and would otherwise fail if that behavior would be changed from the environment)

Wumpf: I modeled this after our previous behavior. I'm in favor of changing it but I'm a bit reluctant at this point: python relied on this behavior as we'd always have an explicit config there so we could set the hooks. The hooks are now separate, so we should be able to change behavior as you describe, but I fear I'm missing some place where we set something

teh-cmc: If it's not a regression then it's fine

Wumpf: no, I'm wrong. We always did unwrap or from_env
Fixing this!

Wumpf:
@rerun-bot full-check

PR #10621, opened 2025-07-14
Make it easier to disable the 3D line grid in Python (blueprint)

Wumpf:
I wanted to turn off the 3D grid lines for one of our examples and found it a bit annoying to do. This makes it a lot easier:

Before:
```py
rrb.Spatial3DView(line_grid=rrb.archetypes.LineGrid3D(line_grid=False)),
```

After:
```py
rrb.Spatial3DView(line_grid=False),
```


rerun_py/rerun_sdk/rerun/blueprint/archetypes/line_grid3d_ext.py
```diff
@@ -0,0 +1,55 @@
...
+from ... import datatypes
+from ...error_utils import catch_and_log_exceptions
+
+
+class LineGrid3DExt:
+    """Extension for [LineGrid3D][rerun.blueprint.archetypes.LineGrid3D]."""
+
+    def __init__(
+        self: Any,
+        visible: datatypes.BoolLike | None = None,
```
ntjohnson1, line 14:
```suggestion
        visible: datatypes.BoolLike = True,
```

Any reason not to just set it rue and remove the note in parameters?

rerun_py/rerun_sdk/rerun/blueprint/archetypes/line_grid3d_ext.py
```diff
@@ -0,0 +1,55 @@
...
+        *,
```
ntjohnson1, line 15: From a python perspective kwarg, *, more_kwargs looks a little weird but it looks like it works and wouldn't really break anything.

rerun_py/rerun_sdk/rerun/blueprint/archetypes/line_grid3d.py
```diff
@@ -15,62 +13,20 @@
...
-            Defaults to a slightly transparent light gray.
-
-        """
-
-        # You can define your own __init__ function as a member of LineGrid3DExt in line_grid3d_ext.py
-        with catch_and_log_exceptions(context=self.__class__.__name__):
-            self.__attrs_init__(visible=visible, spacing=spacing, plane=plane, stroke_width=stroke_width, color=color)
-            return
-        self.__attrs_clear__()
+    # __init__ can be found in line_grid3d_ext.py
```
ntjohnson1, line 29: If you have the branch handy and can double check that `help(LineGrid3D)` shows the inherited init properly that would be great. Weird docstrings on inherited methods are unfortunate.

ntjohnson1 (approved):
Two small nits. Otherwise looks great

rerun_py/rerun_sdk/rerun/blueprint/archetypes/line_grid3d_ext.py
Wumpf: yes, this has a different semantic: if you set it to none, then the Viewer decides what to put here. Which _happens_ to be `True` 😅. But in many cases the behavior the viewer uses (we call it "fallback") can be complex guess work

(note that this code is auto generated)

rerun_py/rerun_sdk/rerun/blueprint/archetypes/line_grid3d.py
Wumpf: I have a look. If this is broken it affects like half of all archetypes since this is an extremely common pattern 😅 
(this file is generated by our codegen)

rerun_py/rerun_sdk/rerun/blueprint/archetypes/line_grid3d_ext.py
Wumpf: My python isn't that amazing, how else would you do it?

ntjohnson1: Generally in python we don't include the * and let people use positional or keywords args. We seem to use this pattern here rejecting positional args for convenience in wrapping things.

Short answer I'm not sure if we actually need the `*` to make the extensions work. If we do then this is totally reasonable with that context but a little weird since python typically doesn't allow positional arguments after keyword arguments.

rerun_py/rerun_sdk/rerun/blueprint/archetypes/line_grid3d_ext.py
ntjohnson1: Ya, that makes sense and follows general python semantics however for bools its usually reasonable to just set the fallback in the api. If its autogenerated then we can leave it alone.

rerun_py/rerun_sdk/rerun/blueprint/archetypes/line_grid3d.py
ntjohnson1: 🤞

Wumpf: it does:
```
Help on class LineGrid3D in module rerun.blueprint.archetypes.line_grid3d:

class LineGrid3D(rerun.blueprint.archetypes.line_grid3d_ext.LineGrid3DExt, rerun._baseclasses.Archetype)
 |  LineGrid3D(visible: 'datatypes.BoolLike | None' = None, *, spacing: 'datatypes.Float32Like | None' = None, plane: 'datatypes.Plane3DLike | None' = None, stroke_width: 'datatypes.Float32Like | None' = None, color: 'datatypes.Rgba32Like | None' = None) -> 'None'
```

ntjohnson1: Note that that only shows our API but doesn't appear to show the nice doc string with the parameters etc. Maybe you just snipped it.

Wumpf: aahh you're right, it does not forward the doc string. My language server does on good days.
This is indeed a problem in large parts of our api

ntjohnson1: Actually it looks like it is fine `help(LineGrid3D.__init__)` worked for me. The ordering is just weird based on inheritance.

Issue #10454, opened 2025-07-01
Component defaults ui shows indicators and displays archetype categories poorly


Wumpf:
Screenshot when running `pixi run -e py python ./tests/python/release_checklist/check_blueprint_overrides.py` and selecting the time series view

<img width="441" alt="Image" src="https://github.com/user-attachments/assets/310162af-ce04-4efa-b147-3345391a4a4f" />

Categories should look like they everywhere else now and we should ignore indicators here

Wumpf:
indicators are gone! but we still want the nice categorization we have everywhere else

PR #10623, opened 2025-07-14
Use new archetype name layout in defaults ui

lucasmerlin:
### Related

- closes https://github.com/rerun-io/rerun/issues/10454

### What

Makes the defaults ui look nicer and more in line:
Screenshot 2025-07-14 at 12 25 59

### DIFF
--- a/crates/viewer/re_data_ui/src/instance_path.rs
+++ b/crates/viewer/re_data_ui/src/instance_path.rs
@@ -190,7 +190,7 @@ fn component_list_ui(
                 if archetype.is_none() && components_by_archetype.len() == 1 {
                     // They are all without archetype, so we can skip the label.
                 } else {
-                    archetype_label_ui(ui, archetype);
+                    archetype_label_lit_item_ui(ui, archetype);
                 }
 
                 for (component_descr, unit) in components {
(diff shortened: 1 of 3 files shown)


crates/viewer/re_data_ui/src/instance_path.rs
```diff
@@ -190,7 +190,7 @@ fn component_list_ui(
                 if archetype.is_none() && components_by_archetype.len() == 1 {
                     // They are all without archetype, so we can skip the label.
                 } else {
-                    archetype_label_ui(ui, archetype);
+                    archetype_label_lit_item_ui(ui, archetype);
```
Wumpf: what does the `lit` stand for?

Wumpf (approved):
nice! While we're here, can we also fix up the menu to not show redundant archetype names?
image

crates/viewer/re_data_ui/src/instance_path.rs
lucasmerlin: Ooops, was supposed to be list

lucasmerlin:
Screenshot 2025-07-14 at 12 45 13

Updated!

Issue #10632, opened 2025-07-15
Remove redundant `component_types` and simplify data model


grtlr:
## Related

- [ ] Follow-up of #6889.

## What

Remove redundant component types (⚠️ all of these need (trivial) migration code)
  - [x] make Transform3D and PoseInstance3D use the same components
  - [ ] Force parameters on GraphView
  - [x] series visibility
  - [ ] new property for recordings: end time
  - …more!

Issue #10631, opened 2025-07-15
Project: Override component descriptors in blueprints and UI


grtlr:
With the introduction of tagged components (#6889), we now attach fine-grained semantic information to the logged data. However, this currently comes with the cost that visualizers are much more narrow in what components they can visualize.

The reason for this is that visualizers only know about components that match their built-in types verbatim.

Let's look at some examples!

### User-defined data

`builtin_component` can be visualized by `PointsSeries`, but the more descriptive `user_component` cannot.

```rs
let builtin_cmoponent = ComponentDescriptor {
  archetype: Some("rerun.archetypes.Scalars".into()),
  component: "Scalars:scalars".into(),
  component_type: Some("rerun.components.Scalar".into(),
};

let user_component = ComponentDescriptor {
  archetype: Some("user.Predications".into()),
  component: "Predictions:confidences".into(),
  component_type: Some("rerun.components.Scalar".into(),
};
```

> \[!WARNING\]<br>This came up almost 1:1 in a [Discord question](<https://discord.com/channels/1062300748202921994/1075873257124810852/1394599491667099699>). The solution of the user is to *duplicate* the data, logging a second component that contains the semantic information for usage in dataframe views.

The whole situation gets even worse because our `AnyValue` APIs make logging partially tagged data—none of which can be visualized—very easy.

### Component re-use

It used to be possible to interpret `vertex_positions` of Mesh3D as the `positions` of Points3D, to make the vertices stand out more. This was achieved with visualizer overrides (which can take a list of visualizers, in this case adding `Points3D` and `Mesh3D`.

> \[!WARNING\]<br>It's still possible to specify this, but the `Points3D` visualizer will not "see" the `vertex_positions`, because it's too narrowly defined.

## Desired situation

Visualizers should be able to interpret components from an archetype that's unknown to it as one of its builtin fields. At first, this could be limited to components that share the expected `component_type`. Conceptually, this means overriding a component descriptor in the context of a given visualizer.

These overrides should be configurable via the Viewer UI, as well as from blueprints. One concept that keeps floating around is that each visualizer has pre-defined slots that we can connect to arbitrary components. This notion is particularly natural when there is a one-to-one mapping between archetypes and visualizers. *This is currently not the case, for all our visualizers.*

## Related

* #9334

## Implementation

There still needs to be an investigation on how to best implement this but it could roughly look like this:

* Visualizers can (optionally) expose the component slots that they're interested in.
[...]

PR #10633, opened 2025-07-15
Cleanup after tagged components

### LINKED ISSUE #6889: Tagged components

grtlr:
### Related

* :tada: Closes #6889

### What

This removes the final `TODO(#6889)` from the code (many of them were outdated). I have created two follow-up issues to track future work that is enabled by having component descriptors everywhere:

* #10631
* #10632


crates/viewer/re_viewport_blueprint/src/view_properties.rs
```diff
@@ -103,8 +103,6 @@ impl ViewProperty {
             .ok_or(ComponentFallbackError::UnexpectedEmptyFallback.into())
     }
 
-    // TODO(#6889): It's a bit sad that we can't derive the type of `C` from the `ComponentDescriptor`, which means
```
Wumpf, line 106: guess we all just grew beyond this sadness and started embracing our type erased reality 😄

Wumpf (approved):
niiiiice! Thanks!

PR #10637, opened 2025-07-15
Add juxtaposition of `SeriesLines` and `SeriesPoints` to `visualizer-overrides` snippet

grtlr:
Question was asked on Discord.


Wumpf (changes requested):
I think you've overdone it with the zooming down. Doesn't look good on my lowdpi firefox
image

grtlr:
I removed that commit again.

It would be nice the enforce consistent zoom levels via automatic image capture à la `TestContext`.

Wumpf:
> It would be nice the enforce consistent zoom levels via automatic image capture à la TestContext

The screenshot command of the command palette does that actually
but ofc means you have to know that, think about it, and then do cutting as a separate step...

Wumpf (approved):

Issue #10634, opened 2025-07-15
rerun menu is stuck when the window height is lower than the one of the menu


gavrelina:
**Describe the bug**
rerun menu is stuck when the window height is lower than the one of the menu. And it only starts to flicker second time. 

on Mac

**To Reproduce**
1. make window height less than the height of the menu
2. open menu
3. it opens first time
4. try opening again
5. it flickers 
6. if you make window higher it will be possible to open the menu again

https://github.com/user-attachments/assets/0a3ca75b-c1bd-480b-bd48-c3bbb4a4331f

PR #10642, opened 2025-07-15
Make rerun menu scrollable

lucasmerlin:
### Related

* closes https://github.com/rerun-io/rerun/issues/10634

### What
Makes the menu scrollable if there is not enough space

https://github.com/user-attachments/assets/1f030286-408b-4b4b-a1f7-216f3c17b188


Wumpf:
on second thought, why aren't all egui menus like that out of the box?

Issue #10344, opened 2025-06-24
Improve error message with `AnyValues` type conflicts


abey79:
`rr.AnyValues` keeps a registry for arrow datatypes per component name, such that this code fails:

```python
rr.set_strict_mode(True)

rr.log("entity", rr.AnyValues(value=10))
rr.log("entity", rr.AnyValues(value="ten"))
```

When this happens, the error message is very obscure though, such as:

```
Traceback (most recent call last):
  File "/Users/hhip/src/rerun/dataframe_sandbox_local.py", line 14, in <module>
    rec.log("/test", rr.AnyValues(yak="yuk", foo="bar", baz="bar"), static=True)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/hhip/src/rerun/rerun_py/rerun_sdk/rerun/any_value.py", line 260, in __init__
    batch = AnyBatchValue(name, value, drop_untyped_nones=drop_untyped_nones)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/hhip/src/rerun/rerun_py/rerun_sdk/rerun/any_value.py", line 100, in __init__
    self.pa_array = pa.array(np_value, type=pa_type)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pyarrow/array.pxi", line 362, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 87, in pyarrow.lib._ndarray_to_array
  File "pyarrow/error.pxi", line 92, in pyarrow.lib.check_status
pyarrow.lib.ArrowNotImplementedError: NumPy type not implemented: unrecognized type (19) in GetNumPyTypeName
```

We should improve this error message.

ntjohnson1:
Closing with #10647 but please re-open if the error message is still opaque.

Issue #4572, opened 2023-12-18
AnyValues should support ND arrays


roym899:
Logging higher-dimensional arrays with AnyValues does not work, only scalar values per-instance seem to be supported right now.

**Is your feature request related to a problem? Please describe.**
Logging the associated ND embedding of a derived 3D embedding, i.e., I would like to inspect the raw, high-dimensional embedding associated with a visualizable low-dimensional embedding.

**Describe the solution you'd like**
Any values should support lists of arrays, lists of lists, etc.

PR #10647, opened 2025-07-15
Nick/anyvalue

### LINKED ISSUE #10556: AnyValues should handle torch tensors with optimal performance
According to a user, changing `rr.AnyValues(embeddings=embeddings),` to `rr.AnyValues(embeddings=embeddings.numpy()),`  where `embeddings` is a torch CPU tensor, tripled the speed.

ntjohnson1:
### Related
Closes #10556 and maybe #10344

### What
* Adds support for the [`__dlpack__` ](https://dmlc.github.io/dlpack/latest/) interface so we can optimally convert into 1d arrays (ND-Array support still not available and probably can't do it no copy #4572). This is mostly to support `torch` but should work for a variety of things.

* Catches the edge cases:
   * Calling AnyValues without a kwarg should say SOMETHING, since the positional only value doesn't generate content for the AnyValue
   * Clarify the error message when we hit the fall through for numpy. Raises from the exception from arrow so we have the full traceback but tries to provide more guidance when we expect a value but get another.
   
 ### Performance comparison
This is not a strict benchmark but should be sufficiently convincing things work nice.
 ```python
>>> pixi run -e py python sample.py raw
Loop 100
1
raw:
        0.0024s
>>> pixi run -e py python sample.py to_numpy
Loop 100
1
to_numpy:
        0.0024s
>>> pixi run -e py python sample.py raw_numpy
Loop 100
1
to_numpy:
        0.0025s
```

### CODE REVIEW

**Review by nikolausWest — APPROVED** [review]
Nice!

**Review by nikolausWest — COMMENTED** [review]
Makes me wonder if we shouldn't do this in more places? Could also be worth checking if it works for Jax and TF in that case. If we make it work for all of those and can make it work more broadly that would be very worth documenting and bragging about!

**Comment by ntjohnson1** [comment]
> Makes me wonder if we shouldn't do this in more places? Could also be worth checking if it works for Jax and TF in that case. If we make it work for all of those and can make it work more broadly that would be very worth documenting and bragging about!

By `do this` you mean leverage dlpack?

I can manually check if this works for JAX and TF (it would be weird if it didn't) but adding the unit test requires a bit more infra since we don't have a separate test env and those are heavy packages to import

**Comment by nikolausWest** [comment]
> By do this you mean leverage dlpack?

Yes, I mean have the ability to convert anything with a `__dlpack__` attribute to an arrow array. I think I may have added some numpy conversion logic that was supposed to work with pytorch earler but it may have been removed since. 

If I don't miss-remember there should also be some way to appease mypy and friends by annotating a type hint of the style "anything that has this attribute" but not sure. 

> I can manually check if this works for JAX and TF (it would be weird if it didn't) but adding the unit test requires a bit more infra since we don't have a separate test env and those are heavy packages to import

Yeah manual check seems fine.

**Comment by ntjohnson1** [comment]
Confirmed it works for jax:
```python
    elif version == "raw_jax":
        timing = simple_loop(shape, raw_any, partial(jax.random.uniform, jax.random.key(123)))
```
```bash
>>>python sample.py raw_jax
Loop 100
1
raw_jax:
        0.0025s
```

ntjohnson1:
> Yes, I mean have the ability to convert anything with a __dlpack__ attribute to an arrow array. I think I may have added some numpy conversion logic that was supposed to work with pytorch earler but it may have been removed since.

IIUC we can convert any 1d object to a pyarrow array with this. Anything greater than 1d either needs to be flattened which could still be fast or made a nested list of lists as a pyarrow limitation

> If I don't miss-remember there should also be some way to appease mypy and friends by annotating a type hint of the style "anything that has this attribute" but not sure.

Yes there is. We should be able to use protocols for that

ntjohnson1:
Tensorflow doesn't work yet because they haven't fully committed to the standard yet
https://www.tensorflow.org/api_docs/python/tf/experimental/dlpack/to_dlpack

But on main it appears they might support `__dlpack__` which landed 2 days after the ~next~ last release. So on the next release it hopefully works. https://github.com/tensorflow/tensorflow/pull/89079

PR #10654, opened 2025-07-15
Introduce `RecordingOpen` viewer event

jprochazk:
```py
def event_handler(event):
  if event.type == "recording_open":
    print(recording_id)

viewer = Viewer()
viewer.on_event(event_handler)
```

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -1662,7 +1662,7 @@ impl App {
             // does the job for now.
 
             let msg_will_add_new_store = matches!(&msg, LogMsg::SetStoreInfo(..))
-                && !store_hub.store_bundle().contains(&store_id);
+                && !store_hub.store_bundle().contains(store_id);
 
             let was_empty = {
                 let entity_db = store_hub.entity_db_mut(store_id);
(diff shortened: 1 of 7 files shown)


crates/viewer/re_viewer/src/app.rs
```diff
@@ -1773,13 +1776,15 @@ impl App {
...
 
-            // Do analytics after ingesting the new message,
-            // because that's when the `entity_db.store_info` is set,
-            // which we use in the analytics call.
+            // Do analytics/events after ingesting the new message,
+            // because `entity_db.store_info` needs to be set.
             let entity_db = store_hub.entity_db_mut(store_id);
-            let is_new_store = matches!(&msg, LogMsg::SetStoreInfo(_msg));
-            if is_new_store && entity_db.store_kind() == StoreKind::Recording {
+            if msg_will_add_new_store && entity_db.store_kind() == StoreKind::Recording {
```
jprochazk, line 1782: Before, this was firing off the event twice... Which is really strange. I didn't investigate too deeply, I changed it to check not only that it's a `SetStoreInfo` message, but also that the store actually doesn't exist yet. That fixed it for me. I assume the root cause is that we are sending multiple store info messages somewhere.

Wumpf: huh, that's a bit worrisome 🤔
but your check makes sense regardless

PR #10657, opened 2025-07-15
Make it possible to configure batcher explicitly via python APIs

jleibs:
### What
We changed the behavior of the batcher in: https://github.com/rerun-io/rerun/pull/10620

However, the only way to override this behavior was to set environment variables, which can be tricky/annoying in some circumstances. This plumbs through the configuration settings so that this can be specified explicitly now.

### DIFF
--- a/rerun_py/src/python_bridge.rs
+++ b/rerun_py/src/python_bridge.rs
@@ -284,11 +284,28 @@ fn flush_and_cleanup_orphaned_recordings(py: Python<'_>) {
 #[derive(Clone)]
 pub struct PyChunkBatcherConfig(ChunkBatcherConfig);
 
+#[derive(FromPyObject)]
+enum DurationLike {
+    Int(i64),
+    Float(f64),
+    Duration(time::Duration),
+}
+
+impl DurationLike {
+    fn into_duration(self) -> time::Duration {
+        match self {
+            DurationLike::Int(i) => time::Duration::from_secs(i as u64),
+            DurationLike::Float(f) => time::Duration::from_secs_f64(f),
+            DurationLike::Duration(d) => d,
+        }
+    }
+}
+
 #[pymethods]
 impl PyChunkBatcherConfig {
     #[new]
     fn new(
-        flush_tick: Option<time::Duration>,
+        flush_tick: Option<DurationLike>,
         flush_num_bytes: Option<u64>,
         flush_num_rows: Option<u64>,
         chunk_max_rows_if_unsorted: Option<u64>,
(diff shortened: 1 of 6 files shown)


Wumpf:
the micro batching page should point out the config hatches for Rust & Python. Let's add a snippet for that

rerun_py/src/python_bridge.rs
```diff
@@ -276,6 +279,167 @@ fn flush_and_cleanup_orphaned_recordings(py: Python<'_>) {
...
+    ///     Split a chunk if it contains >= rows than this threshold and one or more of its timelines are unsorted,
+    ///     by default `None`.
+    ///     Equivalent to setting: `RERUN_CHUNK_MAX_ROWS_IF_UNSORTED` environment variable.
+    fn new(
+        flush_tick: Option<DurationLike>,
+        flush_num_bytes: Option<u64>,
+        flush_num_rows: Option<u64>,
+        chunk_max_rows_if_unsorted: Option<u64>,
+    ) -> Self {
+        let default = ChunkBatcherConfig::DEFAULT;
```
Wumpf: this should be `from_env`
Otherwise setting one of these parameters means that all the other environment variables are no longer honored!
This also has the advantage that a user can retrieve the environments' setting :)

Wumpf (changes requested):
need to fix the env var aspect & update docs a bit. But otherwise looks great! Given that we messed with this very stuff I really want this on 0.24

rerun_py/src/python_bridge.rs
teh-cmc: :yes:

teh-cmc:
Hm, I would have expected this to work, looking at the code :thinking: 
```py
>           config = rr.ChunkBatcherConfig(
                flush_num_bytes=123,
                chunk_max_rows_if_unsorted=789,
            )
E           TypeError: ChunkBatcherConfig.__new__() missing 2 required positional arguments: 'flush_tick' and 'flush_num_rows'
```

EDIT: Okay I see the issue

EDIT2:
```diff
commit 1fd54199f976a54262707615c2f554ed04f7a07f
Author: Clement Rey <cr.rey.clement@gmail.com>
Date:   Wed Jul 16 10:52:32 2025 +0200

    fix pyo3 not injecting default values

diff --git a/rerun_py/src/python_bridge.rs b/rerun_py/src/python_bridge.rs
index 846c17a54f9..f343c42c2df 100644
--- a/rerun_py/src/python_bridge.rs
+++ b/rerun_py/src/python_bridge.rs
@@ -304,6 +304,7 @@ pub struct PyChunkBatcherConfig(ChunkBatcherConfig);
 #[pymethods]
 impl PyChunkBatcherConfig {
     #[new]
+    #[pyo3(signature = (flush_tick=None, flush_num_bytes=None, flush_num_rows=None, chunk_max_rows_if_unsorted=None))]
     #[pyo3(
         text_signature = "(self, flush_tick=None, flush_num_bytes=None, flush_num_rows=None, chunk_max_rows_if_unsorted=None)"
     )]
```

teh-cmc:
Extra snippets and optimization guides will come in follow ups

rerun_py/tests/unit/test_batcher_config.py
```diff
@@ -50,6 +50,27 @@ def test_getter_setter() -> None:
     assert config.chunk_max_rows_if_unsorted == 987
 
 
+def test_partial_overrides() -> None:
```
Wumpf, line 53: niiiiice!

Issue #10586, opened 2025-07-10
Add snapshot tests for all our view types


emilk:
We want kittest screenshot tests covering all our View types (BarChart, Tensor, TextLog, …) with some semi-realistic data.

Ideally these should cover both the view itself, and the selection panel settings for that view!

PR #10659, opened 2025-07-16
Automate `check_1d_tensor_data.py`

grtlr:
### What

It's not much but it's honest work. I've also added a `TestContext::save_recording_to_file` function, which helped me a lot to compare my test data to the original recording produced by the check list.

Note: The original check was _indirectly_ testing `auto_layout` and `auto_views`. That part is gone now.


crates/viewer/re_view_spatial/tests/tensor_1d.rs
```diff
@@ -0,0 +1,60 @@
...
+use re_viewer_context::{ViewClass as _, ViewId, test_context::TestContext};
+use re_viewport::test_context_ext::TestContextExt as _;
+use re_viewport_blueprint::ViewBlueprint;
+
+fn x() -> Vec<f64> {
+    (0..100).map(|i| i as f64 * 100.0 / 99.0).collect()
+}
+
+#[test]
+fn test_tensor() -> anyhow::Result<()> {
```
Wumpf: This effectively tests a 1D image that is created froma tensordata struct. Not sure that makes it tensor enough to call it tensor, I think I'd dub this (test & mod) `image_1d` to avoid confusion

crates/viewer/re_view_spatial/tests/tensor_1d.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let tensor_data =
+        datatypes::TensorData::new(vec![1, 100], datatypes::TensorBuffer::F64(x().into()));
+    let image =
+        archetypes::Image::from_color_model_and_tensor(datatypes::ColorModel::L, tensor_data)?;
+    test_context.log_entity("tensor", |builder| {
+        builder.with_archetype(RowId::new(), TimePoint::STATIC, &image)
+    });
+
+    test_context.save_recording_to_file("tensor_1d_image.rrd")?;
```
Wumpf: ignoring the added dependencies on test_context that's neat that you can do this now, but what is this used for?

those added dependencies do give me pause - I often run `cargo run nextest --all-features -p some_specific_crate` and all `some_specific_crate` that are view crates now have to build a bit more. It's probably not much, but worries me. Would be great to have this only where needed 🤔 (but I also really don't want to add even more feature flags :/)

crates/viewer/re_view_spatial/tests/tensor_1d.rs
```diff
@@ -0,0 +1,60 @@
...
+) {
+    let mut harness = test_context
+        .setup_kittest_for_rendering()
+        .with_size(size)
+        .build(|ctx| {
+            test_context.run_with_single_view(ctx, view_id);
+        });
+    harness.run();
+    harness.snapshot(name);
+}
```
Wumpf: this exact piece of code is now so common, we should add it as one of the boilerplate reducing utilities!
can go in separate PR

crates/viewer/re_view_spatial/tests/tensor_1d.rs
```diff
@@ -0,0 +1,60 @@
...
+    Ok(())
+}
+
+fn setup_blueprint(test_context: &mut TestContext) -> ViewId {
+    test_context.setup_viewport_blueprint(|_ctx, blueprint| {
+        blueprint.add_view_at_root(ViewBlueprint::new_with_root_wildcard(
+            SpatialView2D::identifier(),
+        ))
+    })
+}
```
Wumpf: related to comment below: starting to think we should add `test_context.setup_blueprint_with::<SpatialView2D>`

Wumpf (changes requested):
This almost resolves 
* https://github.com/rerun-io/rerun/issues/10586 

\o/

very nice test additions. Let's get on the same page about `save_recording_to_file` first though

crates/viewer/re_view_spatial/tests/tensor_1d.rs
grtlr: So my use case of adding it was:

1. Start writing tests
2. See black rectangle in .png
3. Wonder what would be shown in the viewer
4. `test_context.save_recording_to_file`
5. Open result in the viewer
6. Check if everything looks correct and potentially derive some meaningful values for `EyeControls3D` or VisualBounds.
7. Add these back into the tests.

But I can see your point too and given our compile times I'm happy to remove the method + deps again.

crates/viewer/re_view_spatial/tests/tensor_1d.rs
grtlr: Yes! I have some ideas there already will do!

crates/viewer/re_view_spatial/tests/tensor_1d.rs
grtlr: I was thinking about wrapping the `TestContext` in a wrapper struct:

```rs
struct SingleViewTestContext<SpatialView2D> {
  inner: TestContext
  _marker: PhantomData<SpatialView2D>
}
```

Then we could define these convenience functions on that new struct and it would not bloat the original `TestContext`.

crates/viewer/re_view_spatial/tests/tensor_1d.rs
Wumpf: this makes a ton of sense. We _have_ to make that debugging this stuff easier 🤔 

I'm inclined to just go with what you have because the alternatives I can think of are so much more complicated.
But to be clear, that particular line being left is then just a testing leftover?

crates/viewer/re_view_spatial/tests/tensor_1d.rs
Wumpf: ohhh that's great. Then we can also do the view registration easily. Love it!

crates/viewer/re_view_spatial/tests/tensor_1d.rs
grtlr: Oh yeah, that line has to go

PR #10664, opened 2025-07-16
Automate `check_static_override.py`

grtlr:
### What

More automation! A lot of the overriding and defaults have been covered in other tests, but this adds logging twice to the same entity as a twist. Might be worth keeping, especially because this uses `SpatialView3D` together with overrides and defaults giving some more coverage.

PR #10671, opened 2025-07-16
Add `SingleViewTestContext` convenience wrapper

grtlr:
### Related

* #10659
* #10664 

### What

This should make writing single view tests more convenient over time.


grtlr:
I've rejected that approach.

Issue #7346, opened 2024-09-03
Enhance Support for Custom Axes and 2D Plotting


pperanich:
**Is your feature request related to a problem? Please describe.**

I'm currently using the rerun viewer to visualize time series data. However, I'm facing challenges when trying to visualize the frequency spectrum of a time series signal. The `SeriesLine` plot is limited by requiring the x-axis to be a time axis, which isn't suitable for frequency spectrum data where the x-axis should represent frequency. Additionally, the `LineStrip2D` plot lacks axes and grids, and the `BarChart` lacks support for custom x-values, making these options less ideal for frequency spectrum visualization.

**Describe the solution you'd like**

I propose the introduction of a new plot type or an enhancement to the existing `SeriesLine` plot that allows for frequency spectrum visualization. Key features should include:
- Support for custom x-values, enabling the x-axis to represent frequency bins instead of time.
- Axes and grid support.
- The ability to plot the data as a continuous line strip or as points (e.g., using `SeriesPoint`), with customizable markers.

**Describe alternatives you've considered**

- Using the `BarChart` with custom x-values (once available) to represent the frequency spectrum.
- Using `LineStrip2D` without any axes information.

**Additional context**

The ability to visualize frequency spectra directly within rerun would significantly enhance its utility for multi-modal time series analysis.

Wumpf:
Related to:
* https://github.com/rerun-io/rerun/issues/5622
* https://github.com/rerun-io/rerun/issues/1387


I was sure there was already a ticket on general 2D plotting itself, but couldn't find it, so might as well be the first! :)

abey79:
I believe this proposal would fit this particular need:
- https://github.com/rerun-io/rerun/issues/8286

Xiao-Chenguang:
Same issue around custom x-axis in BarChart. Internally, rerun use `egui_plot` which support custom x-axis to draw BarChart. I tried to enable feature as @pperanich suggested:
>     * Using the `BarChart` with custom x-values (once available) to represent the frequency spectrum.

Check the [fork](https://github.com/Xiao-Chenguang/rerun) for more detail.

Example code in rust:
```rust
fn main() -> Result<(), Box<dyn std::error::Error>> {
    let rec = rerun::RecordingStreamBuilder::new("custom_index_barchart").spawn()?;
    let x = vec![1, 2, 4, 8];
    let y = vec![4, 5, 3, 6];
    rec.log(
        "bars",
[...]

Issue #5622, opened 2024-03-21
Introducing `ScalarChart` in place of `BarChart`?


teh-cmc:
## Context

We have the `Scalar` archetype that makes it possible to log and visualize scalar timeseries (in the most literal sense: the X axis is literally the recording's clock):
```python
for step in range(0, 64):
    rr.set_time_sequence("step", step)
    rr.log("scalar", rr.Scalar(math.sin(step / 10.0)))
```

This is good in that this allows users to just log their scalar data as it comes without having to manually keep track of any kind of state.

This is bad because it means each scalar has to be its own `DataRow` (1 row == 1 timepoint), which leads to performance issues if you just want to log a huge timeseries for which you already have all the data needed in one place.

These performance issues come in two forms:
* On the client-side, logging will be very slow due to the cost of crafting and serializing `DataRow`s for every scalar.
  We do have [a long-term plan](https://github.com/rerun-io/rerun/issues/4184) that would allow users to log of "temporal batches", i.e. multiple timestamps worth of data in a single log call.
  But A) it will be a while before this is implemented and B) it doesn't solve the second form of performance issue, discussed below.
* On the viewer-side, ingestion will be very slow due to the cost of having to index many `DataRow`s.
  Indexing a row is a costly operation: it not only has to run all the datastore logic (indexing all the individual cells etc), but it also triggers a chain of events that need to propagate and update all downstream subscribers (datastore views, time panel, heuristics, clear cascades...).
  AFAICT, batching row ingestion is much harder problem that batching on the logging side.
  And even if we make it past the ingestion, rendering a time panel with a few million entries is probably still no cheap task (?).

Interestingly, we also have the `BarChart` archetype, which already has the nice property of accepting a batch of scalars all at once:
```python
rr.log("bar_chart", rr.BarChart([8, 4, 0, 9, 1, 4, 1, 6, 9, 0]))
```

The one downside is that this doesn't integrate at all with the time cursor, since the barchart as a whole is its own entity.

But, for many cases, this can still be a very useful tool in real world scenarios, especially when combined with timeless/static.

## Proposal

Retire the `BarChart` archetype in favor of a new generic `ScalarChart` archetype:
```fbs
table ScalarChart {
  /// The values. Should always be a rank-1 tensor.
  values: rerun.components.TensorData ("attr.rerun.component_required", order: 1000);

  /// The optional indices. Should always be a rank-1 tensor with the same length as `values`.
  ///
[...]

Issue #1387, opened 2023-02-23
view_coordinates for 2D views


jleibs:
View_coordinates are currently ignored for 2d views. Should be possible to use these to change the way x and y are mapped to up / right on the screen.

Related to: https://github.com/rerun-io/rerun/issues/349

emilk:
Blocked on replacing our `egui::Image` with re_renderer

Wumpf:
I took a quite successful stab at #1581 which just like this ticket requires touching the camera we use for 2D views, making this related

emilk:
This is pretty important when plotting something arbitrary and you want Y=up

patrickelectric:
Any plans to have a open street map visualizer or something like it ?

Wumpf:
@patrickelectric we're hoping it's soon possible to log all the necessary 2d primitives to rerun to make it possible to write an application that puts openstreet map data into Rerun, but we haven't planned to do something like that third party. That's offtopic though - this issue is about a very particular aspect of the 2d view

emilk:
We should do this first:
* https://github.com/rerun-io/rerun/issues/2663

adamconkey:
Interested in this for spatial 2D view, I'm visualizing some polygons and assumed +X would be up and +Y right, but the viewer seems to be +X is down so my polygons are effectively flipped vertically.

PR #10675, opened 2025-07-16
Support flexiable x-axis for bars in BarChart

Xiao-Chenguang:
### Related

* Part of #7346
* Part of #5622
* Part of #1387

### What

**Support customise x-axis (index) for BarChart**

This draft PR include changes that enable index customization of BarChart.
The index of BarChart has to be 0-indexed continus intagers originally, while a lot of use case require a customized index as in related issues.

This PR include changes to both the BarChart data type as view system. As rerun use BarChart from egui_plot where index is supported, an extra field _indexes_ is add to rerun.BarChart.

### DIFF
--- a/crates/store/re_types/definitions/rerun/archetypes/bar_chart.fbs
+++ b/crates/store/re_types/definitions/rerun/archetypes/bar_chart.fbs
@@ -22,4 +22,7 @@ table BarChart (
 
   /// The color of the bar chart
   color: rerun.components.Color ("attr.rerun.component_optional", nullable, order: 2000);
+
+  /// The indexes. Should always be a 1-dimensional tensor (i.e. a vector).
+  indexes: rerun.components.TensorData ("attr.rerun.component_optional", nullable, order: 3000);
 }
--- a/crates/viewer/re_view_bar_chart/Cargo.toml
+++ b/crates/viewer/re_view_bar_chart/Cargo.toml
@@ -27,11 +27,12 @@ re_view.workspace = true
 re_tracing.workspace = true
 re_types = { workspace = true, features = ["egui_plot"] }
 re_ui.workspace = true
-re_viewer_context.workspace = true
+re_viewer_context = { workspace = true }
 re_viewport_blueprint.workspace = true
 
 egui_plot.workspace = true
 egui.workspace = true
+arrow.workspace = true
 
 [dev-dependencies]
 re_viewport = { workspace = true, features = ["testing"] }
--- a/crates/viewer/re_view_bar_chart/src/view_class.rs
+++ b/crates/viewer/re_view_bar_chart/src/view_class.rs
@@ -200,6 +200,7 @@ impl ViewClass for BarChartView {
             } = plot.show(ui, |plot_ui| {
                 fn create_bar_chart<N: Into<f64>>(
                     ent_path: &EntityPath,
+                    indexes: impl Iterator<Item = f64>,
                     values: impl Iterator<Item = N>,
                     color: &re_types::components::Color,
                     theme: egui::Theme,
@@ -214,9 +215,10 @@ impl ViewClass for BarChartView {
                     BarChart::new(
                         "bar_chart",
                         values
+                            .zip(indexes)
                             .enumerate()
-                            .map(|(i, value)| {
-                                Bar::new(i as f64 + 0.5, value.into())
+                            .map(|(i, (value, index))| {
+                                Bar::new(index.into(), value.into())
                                     .width(1.0) // No gaps
                                     .name(format!("{ent_path} #{i}"))
                                     .fill(fill)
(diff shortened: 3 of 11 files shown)


grtlr:
Thank you for your contribution! What's the status of your work, do you plan on finishing this up or is it ready for review yet?

Xiao-Chenguang:
> Thank you for your contribution! What's the status of your work, do you plan on finishing this up or is it ready for review yet?

Hi grtlr, It's my first PR to rerun. So I make it a draft PR as requested. It's ready for review now except the failed RP label check.

grtlr:
Thank you for the clarification! We'll put it onto our review queue!

crates/store/re_types/definitions/rerun/archetypes/bar_chart.fbs
```diff
@@ -22,4 +22,7 @@ table BarChart (
 
   /// The color of the bar chart
   color: rerun.components.Color ("attr.rerun.component_optional", nullable, order: 2000);
+
+  /// The indexes. Should always be a 1-dimensional tensor (i.e. a vector).
+  indexes: rerun.components.TensorData ("attr.rerun.component_optional", nullable, order: 3000);
```
Wumpf: I think the name `indexes` is rather misleading since its not indexing anything. The technical term for this is `abscissa` (see https://en.wikipedia.org/wiki/Abscissa_and_ordinate), I'm inclined to go with that instead.

Documentation should explain what these are actually doing, that's not clear from reading it right now. Also, it should mention what happens in absence of it.

Top level comment `/// The x values will be the indices of the array, and the bar heights will be the provided values.` is now also outdated and needs updating.

crates/viewer/re_view_bar_chart/Cargo.toml
```diff
@@ -27,11 +27,12 @@ re_view.workspace = true
 re_tracing.workspace = true
 re_types = { workspace = true, features = ["egui_plot"] }
 re_ui.workspace = true
-re_viewer_context.workspace = true
+re_viewer_context = { workspace = true }
```
Wumpf: needless change

crates/viewer/re_view_bar_chart/src/visualizer_system.rs
```diff
@@ -17,7 +17,14 @@ use re_viewer_context::{
...
 #[derive(Default)]
 pub struct BarChartVisualizerSystem {
-    pub charts: BTreeMap<EntityPath, (datatypes::TensorData, components::Color)>,
+    pub charts: BTreeMap<
+        EntityPath,
+        (
+            datatypes::TensorData,
+            datatypes::TensorData,
+            components::Color,
+        ),
```
Wumpf: this is getting confusing, let's make that a separate struct so we can name those fields

crates/viewer/re_view_bar_chart/Cargo.toml
```diff
@@ -27,11 +27,12 @@ re_view.workspace = true
...
 re_viewport_blueprint.workspace = true
 
 egui_plot.workspace = true
 egui.workspace = true
+arrow.workspace = true
```
Wumpf: please sort alphabetically

crates/viewer/re_view_bar_chart/src/visualizer_system.rs
```diff
@@ -74,4 +85,39 @@ impl TypedComponentFallbackProvider<components::Color> for BarChartVisualizerSys
     }
 }
 
-re_viewer_context::impl_component_fallback_provider!(BarChartVisualizerSystem => [components::Color]);
+impl TypedComponentFallbackProvider<components::TensorData> for BarChartVisualizerSystem {
+    fn fallback_for(&self, ctx: &QueryContext<'_>) -> components::TensorData {
```
Wumpf, line 93: nice!

Wumpf (changes requested):
thanks for your contribution! I think with some iteration on naming and documentation we can take this as-is :)

Please add an additional snippet, similar to `docs/snippets/all/archetypes/tensor_simple.py`/cpp/rs demonstrating the new feature. Thank you!

crates/store/re_types/definitions/rerun/archetypes/bar_chart.fbs
Xiao-Chenguang: Yes, abscissa makes more sense. Another choices is `arguments` as used in `egui_plot::BarChart` in https://github.com/emilk/egui_plot/blob/875477028270faf0f1d60c34496c4e4f619a005b/egui_plot/src/items/bar.rs#L10C1-L21C20. This is more consistant with `egui_plot` while not intuitive enough.

The top level comment may be updated as `/// The bar heights will be the provided values, and the bar abscissa will be the provided arguments or default to the index of the provided values.`

crates/viewer/re_view_bar_chart/src/visualizer_system.rs
Xiao-Chenguang: That's for minimum changes to the code, while it does look confusing. Any suggestion or reference on this data structure?

crates/store/re_types/definitions/rerun/archetypes/bar_chart.fbs
Xiao-Chenguang: Let's use abscissa.

crates/viewer/re_view_bar_chart/Cargo.toml
Xiao-Chenguang: reverted

crates/viewer/re_view_bar_chart/src/visualizer_system.rs
Xiao-Chenguang: A BarChartData structure is created for this.

crates/viewer/re_view_bar_chart/Cargo.toml
Xiao-Chenguang: sorted now

Wumpf:
interesting, the image comparision test found that this actually breaks layout a bit

before:
tensor_1d


after:
tensor_1d new

crates/viewer/re_view_bar_chart/src/view_class.rs
```diff
@@ -214,9 +215,10 @@ impl ViewClass for BarChartView {
                     BarChart::new(
                         "bar_chart",
                         values
+                            .zip(indexes)
                             .enumerate()
-                            .map(|(i, value)| {
-                                Bar::new(i as f64 + 0.5, value.into())
+                            .map(|(i, (value, index))| {
+                                Bar::new(index.into(), value.into())
```
Wumpf: bars moved because of this. let's try to preserve the old behavior for integer indices

crates/viewer/re_view_bar_chart/src/view_class.rs
```diff
@@ -240,14 +240,14 @@ impl ViewClass for BarChartView {
...
+                        TensorBuffer::U16(data) => data.iter().map(|v| *v as f64 + 0.5).collect(),
+                        TensorBuffer::U32(data) => data.iter().map(|v| *v as f64 + 0.5).collect(),
+                        TensorBuffer::U64(data) => data.iter().map(|v| *v as f64 + 0.5).collect(),
+                        TensorBuffer::I8(data) => data.iter().map(|v| *v as f64 + 0.5).collect(),
+                        TensorBuffer::I16(data) => data.iter().map(|v| *v as f64 + 0.5).collect(),
+                        TensorBuffer::I32(data) => data.iter().map(|v| *v as f64 + 0.5).collect(),
+                        TensorBuffer::I64(data) => data.iter().map(|v| *v as f64 + 0.5).collect(),
                         TensorBuffer::F16(data) => data.iter().map(|v| f64::from(*v)).collect(),
                         TensorBuffer::F32(data) => data.iter().map(|v| *v as f64).collect(),
                         TensorBuffer::F64(data) => data.iter().copied().collect(),
```
Xiao-Chenguang: @Wumpf The shift of 0.5 should be reverted. I am confused why we skip the shift for float data? This change may lead to discrepency when we supply the same data with different data type. Maybe add `+0.5` back when create `Bar`s is better.

Wumpf: previously, we placed bars in the middle of their index. That's why the image comparison test broke. But yeah you're right, doing this only for one type of data is definitely wrong. Thank for stopping me there

Wumpf: also, I triggered clippy

Wumpf: fixed

Wumpf:
@Xiao-Chenguang maybe you want to have a last look before merging given that I messed with it so much now :)

Xiao-Chenguang:
> @Xiao-Chenguang maybe you want to have a last look before merging given that I messed with it so much now :)

it looks good now. let's go ahead

PR #10676, opened 2025-07-16
Don't style (&describe) missing video samples like an error

Wumpf:
Before:
466967029-79f232c8-eac4-41a6-aadb-dc0d81456327

After:
image

Issue is that missing video samples may be entirely transitional, so not actually an error. Making look like an error is irritating at best

### DIFF
--- a/crates/viewer/re_view_spatial/src/pickable_textured_rect.rs
+++ b/crates/viewer/re_view_spatial/src/pickable_textured_rect.rs
@@ -14,8 +14,8 @@ pub enum PickableRectSourceData {
     /// The rectangle is a frame in a video.
     Video,
 
-    /// The rectangle represents an error icon.
-    ErrorPlaceholder,
+    /// The rectangle represents an placeholder icon.
+    Placeholder,
 }
 
 /// Image rectangle that can be picked in the view.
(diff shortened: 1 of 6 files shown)


crates/viewer/re_view_spatial/src/pickable_textured_rect.rs
```diff
@@ -14,8 +14,8 @@ pub enum PickableRectSourceData {
     /// The rectangle is a frame in a video.
     Video,
 
-    /// The rectangle represents an error icon.
-    ErrorPlaceholder,
+    /// The rectangle represents an placeholder icon.
```
Wumpf:
```suggestion
    /// The rectangle represents a placeholder icon.
```

crates/viewer/re_view_spatial/src/picking_ui.rs
```diff
@@ -256,7 +256,7 @@ fn get_pixel_picking_info(
             .and_then(|picked_rect| {
                 if matches!(
                     picked_rect.source_data,
-                    PickableRectSourceData::ErrorPlaceholder
+                    PickableRectSourceData::Placeholder { .. }
```
Wumpf: oh interesting, clippy doesn't complain about those superfluous curlies. Those are there because originally I had this convey whether it's an error or not

Wumpf:
hmm I could give you some superfluous curly braces that you can remove again, would that help?

PR #10678, opened 2025-07-16
Return dataframe types instead of polluting with any

ntjohnson1:
### What
Right now we return our dataframes as Any which basically makes anything they touch Any and that makes thing less helpful inside the IDE. Exploring making them not Any but actual dataframes.

We know its a DataFrame so we label it as such even though rust only can infer PyAny.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -21,12 +21,6 @@ IS_IN_RERUN_WORKSPACE = "yes"
 [target.wasm32-unknown-unknown]
 rustflags = ['--cfg=web_sys_unstable_apis', '--cfg=getrandom_backend="wasm_js"']
 
-[target.x86_64-apple-darwin]
-rustflags = ["-C", "link-arg=-undefined", "-C", "link-arg=dynamic_lookup"]
-
-[target.aarch64-apple-darwin]
-rustflags = ["-C", "link-arg=-undefined", "-C", "link-arg=dynamic_lookup"]
-
 # TODO(emilk): This should be unnecessary when we update to the 2024 edition in Rust 1.85.
 # See https://blog.rust-lang.org/2025/01/09/Rust-1.84.0.html#cargo-considers-rust-versions-for-dependency-version-selection
 [resolver]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -21,6 +21,12 @@ IS_IN_RERUN_WORKSPACE = "yes"
 [target.wasm32-unknown-unknown]
 rustflags = ['--cfg=web_sys_unstable_apis', '--cfg=getrandom_backend="wasm_js"']
 
+[target.x86_64-apple-darwin]
+rustflags = ["-C", "link-arg=-undefined", "-C", "link-arg=dynamic_lookup"]
+
+[target.aarch64-apple-darwin]
+rustflags = ["-C", "link-arg=-undefined", "-C", "link-arg=dynamic_lookup"]
+
 # TODO(emilk): This should be unnecessary when we update to the 2024 edition in Rust 1.85.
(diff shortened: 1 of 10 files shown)


.cargo/config.toml
```diff
@@ -21,6 +21,12 @@ IS_IN_RERUN_WORKSPACE = "yes"
 [target.wasm32-unknown-unknown]
 rustflags = ['--cfg=web_sys_unstable_apis', '--cfg=getrandom_backend="wasm_js"']
 
+[target.x86_64-apple-darwin]
+rustflags = ["-C", "link-arg=-undefined", "-C", "link-arg=dynamic_lookup"]
+
+[target.aarch64-apple-darwin]
+rustflags = ["-C", "link-arg=-undefined", "-C", "link-arg=dynamic_lookup"]
```
Wumpf: what is this, what does it and why is it needed? ;)

deny.toml
```diff
@@ -66,6 +66,7 @@ skip = [
   { name = "pulldown-cmark" },   # Build-dependency via `ply-rs` (!). TODO(emilk): use a better crate for .ply parsing
   { name = "redox_syscall" },    # Plenty of versions in the wild
   { name = "rustc-hash" },       # numpy with compatible pyo3 requires different version than wgpu
+  { name = "bzip2" },            # Datafusion-avro has us indirectly pinned to this
```
Wumpf: where does the conflict stem though from? Do we oureslves just have to update bzip2 or is it another package that wants a different version?

rerun_py/Cargo.toml
```diff
@@ -81,6 +81,7 @@ comfy-table.workspace = true
 crossbeam.workspace = true
 datafusion.workspace = true
 datafusion-ffi.workspace = true
+datafusion-python.workspace = true
```
Wumpf: this seems to be quite a large dependency for what's mostly a type annotation change? I figure us taking this on is inevitable though anyways?

Leaving this to @timsaucer how's literally the expert on this :)

Wumpf (changes requested):
actual py/rs changes look nice! but there's some comments that need adressing

also would be nice if @timsaucer could look over this :)

rerun_py/src/catalog/table_entry.rs
```diff
@@ -51,23 +51,31 @@ impl PyTableEntry {
...
         // We're fine with this failing.
-        ctx.call_method1("deregister_table", (table_name.clone(),))?;
+        PySessionContext::deregister_table(&mut py_session_ctx, &table_name)?;
 
-        ctx.call_method1("register_table_provider", (table_name.clone(), self_))?;
+        PySessionContext::register_table_provider(
+            &mut py_session_ctx,
+            &table_name,
+            self_.into_bound_py_any(py)?,
+        )?;
```
timsaucer: We definitely do not want to do this. I'll go into more detail later today, but here is some background reading that explains why I hope: https://datafusion.apache.org/python/contributor-guide/ffi.html

timsaucer (changes requested):

.cargo/config.toml
teh-cmc: We definitely should not set _native_ RUSTFLAGS here, this will cause all sorts of nasty issues both for developers and CIs alike.

timsaucer:
This is just for the type hinting, right? Can't we just make the changes in the pyi files and not touch the rust code?

ntjohnson1:
> This is just for the type hinting, right? Can't we just make the changes in the pyi files and not touch the rust code?

That's what I'm staging now. I'm not sure if our signature checking allows type narrowing but I'll test that locally.

Mostly I couldn't find a good reference poking through the commits for why we were following the pattern we were for datafusion where for arrow we were able to directly return their specified types. Not dynamically importing the python inside of rust seemed like a nice side effect but from the link above not currently feasible.

PR #10680, opened 2025-07-17
Stop using unnecessary CentralPanel in most image comparison tests

Wumpf:
We often used a separate `CentralPanel` instead of just calling `build_ui` on the egui_kittest testharness.
This meant that the extra guardband the harness added was black rather transparent

* [x] full ci to make sure that image comp doesn't fail on mac/windows

PR #10682, opened 2025-07-17
Automate `check_latest_at_partial_updates.py`

grtlr:
### Related

* [x] Potential merge race with #10680.

### What

Moar automation.


Wumpf (approved):
excellent. that's a really good one to have automated, thanks for taking care that the points are very visible on the snapshots

PR #10684, opened 2025-07-17
Use `warning_label` and automate  `check_static_components_ui.py`

grtlr:
### Related

* [ ] Potential merge race with #10680.

### What

Moar automation. Found a small inconsistency, where we did not use `warning_label`.


crates/viewer/re_selection_panel/src/selection_panel.rs
```diff
@@ -1134,4 +1141,184 @@ mod tests {
...
+                    );
+                });
+                test_context.handle_system_commands();
+            });
+
+        harness.run();
+        harness.snapshot("selection_panel_component_static_overwrite");
+    }
+
+    /// Snapshot test for the selection panel when a static component with additional time information is selected.
```
Wumpf:
```suggestion
    /// Snapshot test for the selection panel when a static component with additional temporal chunks is selected (user error).
```

PR #10686, opened 2025-07-17
attempt fix doc deployment

Wumpf:
use the redirect instead of direct link which is  used elsewhere, maybe that fixes deployment?!


Wumpf:
well, preview is happy with it

Issue #10689, opened 2025-07-17
Datafusion UDFs depend on arrow.compute being enabled/injected by `.filter`


jleibs:
This is totally bizarre.<br><br>This fails:

```python
import datafusion as dfn
import rerun as rr
from rerun.utilities.datafusion.functions.url_generation import partition_url
client = rr.catalog.CatalogClient("rerun://sandbox.redap.rerun.io")
dataset = client.get_dataset(name="droid:raw")
dataset.partition_table().df().with_column("url", partition_url(dataset))
```

with the exception:

```
Exception: DataFusion error: Execution("PyErr { type: <class 'AttributeError'>, value: AttributeError(\"module 'pyarrow' has no attribute 'compute'\"
```

However, if one includes a filter operation, then everything works:

```
dataset.partition_table().df().filter(dfn.lit(True)).with_column("url", partition_url(dataset))
```

PR #10690, opened 2025-07-17
Explicitly import pyarrow.compute in url_generation.py

jleibs:
### What
- Resolves: https://github.com/rerun-io/rerun/issues/10689


ntjohnson1 (approved):
Can confirm we use pyarrow.compute so importing it makes sense. Why did none of our type checkers/linters etc save us?

ntjohnson1:
FYI nothing saved us because pyarrow isn't typed 
https://github.com/apache/arrow/issues/32609
(~I also got a no types warning from datafusion too so that might not be packaged correctly or that was just my local setup~). We should consider using the 3rd party stubs

Note the datafusion error I saw was becuse rerun imports datafusion but we don't specify it in our dependencies 💥

PR #10696, opened 2025-07-18
Require Datafusion Since We Import It

### LINKED ISSUE #10695: Make Datafusion dependency explicit
**Describe the bug**
Rerun imports from datafusion in python but doesn't specify it as a requirement.

**To Reproduce**
```console
uv venv --python=3.12
source .venv/bin/activate
python
>>> from rerun.utilities.datafusion.functions import url_generation
  File "<stdin>", line 1, in <module>
  File "/Users/nick/repos/rerun/.venv/lib/python3.12/site-packages/rerun_sdk/rerun/utilities/datafusion/functions/url_generation.py", line 4, in <module>
    from datafusion import Expr, ScalarUDF, col, udf
ModuleNotFoundError: No module named 'datafusion'
```

**Expected behavior**
We should be able to import all of the code included in our python package.

 - OS: macOS
 - 
**Rerun version**
'0.24.0'

ntjohnson1:
### Related
Closes #10695 

### What
This add the datafusion version we specify in our examples as a lower bound to make sure datafusion gets install with rerun before we use it. Added a simple smoke test to repro the issue then verified it passed with the install specification.

### DIFF
--- a/rerun_py/pyproject.toml
+++ b/rerun_py/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
   "pillow>=8.0.0",          # Used for JPEG encoding. 8.0.0 added the `format` arguments to `Image.open`
   "pyarrow>=18.0.0",
   "typing_extensions>=4.5", # Used for PEP-702 deprecated decorator
+  "datafusion>=45.0.0",
 ]
 description = "The Rerun Logging SDK"
 keywords = ["computer-vision", "logging", "rerun"]
--- a/pixi.lock
+++ b/pixi.lock
@@ -4117,7 +4117,7 @@ environments:
       - pypi: ./examples/python/shared_recording
       - pypi: ./examples/python/stdio
       - pypi: ./examples/python/structure_from_motion
-      - pypi: rerun_py
+      - pypi: ./rerun_py
       osx-arm64:
       - conda: https://conda.anaconda.org/conda-forge/osx-arm64/aom-3.9.1-h7bae524_0.conda
       - conda: https://conda.anaconda.org/conda-forge/osx-arm64/aws-c-auth-0.7.31-hee1f4ab_5.conda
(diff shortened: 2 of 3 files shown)


rerun_py/pyproject.toml
```diff
@@ -20,6 +20,7 @@ dependencies = [
   "pillow>=8.0.0",          # Used for JPEG encoding. 8.0.0 added the `format` arguments to `Image.open`
   "pyarrow>=18.0.0",
   "typing_extensions>=4.5", # Used for PEP-702 deprecated decorator
+  "datafusion>=45.0.0",
```
timsaucer: For now the best practice is going to be to keep the datafusion python version the same as the datafusion rust version we use for it's FFI crate. They are currently cross-version compatible, but there have been some breaking changes in the most recent release.

That does mean we'd want users to be on 47.0.0.

Also, more generally, I think we don't want this as a dependency in rerun, because it means any of our open source viewer users who are *not* using datafusion now have to bring in this fairly large package. This is why it hasn't been put in as dependency before.

I know some packages do things like `pip install rerun-sdk[datafusion]` type things to add in additional dependencies, but I am not sure how that is done.

ntjohnson1: Gotcha I'll take a look at adding it as optional and hiding the component without that. I'm sure its just a maturing flag. We have a similar issue in our notebook apparently so I'm doing additional larger cleanup for some of the things this kicked up.

rerun_py/rerun_sdk/rerun/utilities/datafusion/functions/url_generation.py
```diff
@@ -2,9 +2,16 @@
...
 from rerun_bindings import DatasetEntry
 
+from rerun.error_utils import RerunOptionalDependencyError
+
+HAS_DATAFUSION = True
+try:
+    from datafusion import Expr, ScalarUDF, col, udf
+except ModuleNotFoundError:
+    HAS_DATAFUSION = False
+
```
abey79, line 14: Is the type checker able to detect if a new method accesses eg `Expr` without first testing `HAS_DATAFUSION`?

ntjohnson1: Good question. Our type checking is SO BROKEN right now https://github.com/rerun-io/rerun/pull/10704. Maybe I merge this and then pull main into that branch to check it out.

PR #10697, opened 2025-07-18
Use `ui.warning_label()` in more places

grtlr:
### What

We should use `ui.warning_label` whenever possible. It looks nicer and works better in light mode.


grtlr:
I had a brief look and was not sure about the other places that use it.

Issue #10639, opened 2025-07-15
Datafusion Table Provider ties up memory in the shared ctx


jleibs:
Every query that returns a `TableProvider` registers it with the shared session context that's part of the `CatalogClient` . <br><br>These TableProviders are still fairly heavy, internally holding a ChunkStore for each queried partition.

This is particularly pathological fore notebooks where we may iterate and run the same query multiple times.

PR #10698, opened 2025-07-18
DataFrame query streaming provider without memory limit

timsaucer:
### Related

Closes https://github.com/rerun-io/rerun/issues/10639

### What

This PR switches us from collecting the entire dataset into memory to collecting a single partition into memory and then streaming out the results.  The remaining issues to address prior to merging are:

- [x] ~~We're getting "Data truncated due to size" even when showing a tiny dataframe, just id and time. It should have many more rows.~~ This was a configuration issue
- [x] I need to make sure I do some static queries also because they may break the time based portions
- [x] Verify this works for ALL different time indices - time stamps vs i64 (u64?)
- [x] Validate sorting is as expected - add unit tests
- [x] Add tracing spans back in


timsaucer:
How I tested this to validate the sorting and work with different types of time indices:

I generated a bunch of `.rrd` files with this:

```python
import rerun as rr
import numpy as np

def store_unsorted_data(index: int):
    filename = f"unsorted_data_{index}"
    rr.init(filename, spawn=True)
    rr.save(f"{filename}.rrd")
    
    red_colors = np.zeros((10,3), dtype=np.uint8)
    red_colors[:,0] = np.linspace(0,255,10)
    blue_colors = np.zeros((10,3), dtype=np.uint8)
    blue_colors[:,2] = np.linspace(0,255,10)

    # Intentionally add data to the RRD out of order
    time_ordering = [0, 1, 2, -3, -4, -5, 6, -7, 8, 9]
    
    from time import time
    now = time()

    rr.log("/text1", rr.TextDocument("Before text"), static=True)
    
    for dt in time_ordering:
        rr.reset_time()
        rr.set_time("my_timestamp", timestamp=now + dt)

        # We want some nulls in time
        if dt % 3 == 0:
            rr.set_time("my_sequence", sequence=dt)
        rr.set_time("my_duration", duration=dt)
    
        positions = np.zeros((10, 3))
        positions[:,0] = np.linspace(-10 + dt,10 + dt,10)

        # We need two entities so that we get more than one chunk
        # per partition
        rr.log(
            "/obj1",
            rr.Points3D(positions, colors=red_colors, radii=0.5)
        )

        # We want nulls in some of the entities
        if dt % 5 == 0:
            positions[:,1] = np.linspace(1,1,10)
            rr.log(
                "/obj2",
                rr.Points3D(positions, colors=blue_colors, radii=0.5)
            )

    rr.log("/text2", rr.TextDocument("After text"), static=True)

# Intentionally call these out of order so that we get different
# timestamps. We will register them in order.

store_unsorted_data(4)
store_unsorted_data(0)
store_unsorted_data(3)
store_unsorted_data(1)
store_unsorted_data(2)
store_unsorted_data(7)
store_unsorted_data(8)
store_unsorted_data(5)
store_unsorted_data(9)
store_unsorted_data(6)

for idx in range(10, 20):
    store_unsorted_data(idx)
```

And then I registered these with my local instance with:

```shell
for file in ~/working/redap_data/unsorted_simple_data/*.rrd; do
    pixi run redap register --id 18575B0F0DA7EB350b164f8fe6e4d812 --rrd-url "file://$file"
done
```

And finally I ran this test:

```python
import rerun as rr
from datafusion import col, functions as F

CATALOG_URL = "rerun+http://localhost:51234"
DATASET = "unsorted_test_data"

client = rr.catalog.CatalogClient(CATALOG_URL)
dataset = client.get_dataset(name=DATASET)

robot = dataset.dataframe_query_view(index="my_timestamp", contents="/**")

streams = (
    robot.fill_latest_at()
    .df()
    .execute_stream_partitioned()
)


prior_partition_ids = set()
for rb_reader in streams:

    prior_partition = ""
    prior_timestamp = 0
    for rb in iter(rb_reader):

        rb = rb.to_pyarrow()
        for idx in range(rb.num_rows):
            partition = rb[0][idx].as_py()
            timestamp = rb[1][idx].value

            if partition < prior_partition:
                print(f"Invalid from {prior_partition} to {partition}")
            assert partition >= prior_partition
            if partition == prior_partition:
                assert timestamp >= prior_timestamp
            else:
                assert partition not in prior_partition_ids
                prior_partition_ids.add(partition)
        
            prior_partition = partition
            prior_timestamp = prior_timestamp
```

I would want some version of this in our tests that we've been discussing adding, but that may be a separate effort.

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -186,3 +438,621 @@ pub fn align_record_batch_to_schema(
...
     )?)
 }
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+struct ChunkInfo {
+    pub start_time: i64,
+    pub end_time: i64,
+    pub chunk_id: ChunkId,
+    pub byte_len: u64,
+}
```
jleibs: This is a sufficiently universally useful concept that it probably belongs somewhere higher up.

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -186,3 +438,621 @@ pub fn align_record_batch_to_schema(
...
+    let mut results = BTreeMap::new();
+
+    for batch in chunk_info_batches.as_ref() {
+        let schema = batch.schema();
+        let end_time_col = schema
+            .fields()
+            .iter()
+            .map(|f| f.name())
+            .find(|name| name.ends_with(":end"))
+            .ok_or(exec_datafusion_err!("Unable to identify time index"))?;
```
jleibs: Nit: this seems like a helper or extension we should add to Schema.

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -186,3 +438,621 @@ pub fn align_record_batch_to_schema(
...
+/// up by the execute fn of the physical plan, so we will start one per output partition,
+/// which is different from the partition_id.
+#[tracing::instrument(level = "trace", skip_all)]
+async fn chunk_stream_io_loop(
+    mut client: ConnectionClient,
+    base_request: GetChunksRequest,
+    mut rerun_partition_ids: Vec<String>,
+    output_channel: Sender<ChunksWithPartition>,
+) -> Result<(), DataFusionError> {
+    rerun_partition_ids.sort();
```
jleibs, line 414: partition-ordering, how it relates to temporal ordering of the data stored within the partitions, and whether partitions are ordered/overlapping  is an interesting area of topics. This would be a helpful place to have a comment explaining any assumptions (or non-assupmtions) and possible TODOs.

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -186,3 +438,621 @@ pub fn align_record_batch_to_schema(
...
+        plan_err!("PartitionStreamExec does not support children")
+    }
+
+    #[tracing::instrument(level = "info", skip_all)]
+    fn execute(
+        &self,
+        partition: usize,
+        _context: Arc<TaskContext>,
+    ) -> datafusion::common::Result<SendableRecordBatchStream> {
+        let (chunk_tx, chunk_rx) = tokio::sync::mpsc::channel(32); // 32 batches of chunks, not 32 chunks
```
jleibs: We hard-coded 32 here, but at the top of the file we have a const for num_ouput_partitions. Is there an optimal relationship betwen them?

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -186,3 +438,621 @@ pub fn align_record_batch_to_schema(
...
+    /// When thread is active, is Some
+    thread_join_handle: Option<std::thread::JoinHandle<()>>,
+}
+
+impl Drop for CpuRuntime {
+    fn drop(&mut self) {
+        // Notify the thread to shutdown.
+        self.notify_shutdown.notify_one();
+        // In a production system you also need to ensure your code stops adding
+        // new tasks to the underlying runtime after this point to allow the
```
jleibs: This is a production system. If there's something actionable, we should probably make it a TODO.

crates/store/re_datafusion/src/dataframe_query_provider_wasm.rs
jleibs, line 1: This whole file warrants a top-level explanation of why it exists.  In particular, I'm curious what the blockers are to using something like: https://github.com/rerun-io/rerun/blob/fa26f30a59366790925b14b47a66c834bd6d43d0/crates/viewer/re_viewer/src/background_tasks.rs#L15-L31

Would be nice to just maintain one implementation.

rerun_py/src/catalog/dataframe_query.rs
```diff
@@ -486,29 +490,40 @@ impl PyDataframeQueryView {
...
+        //     .into_iter()
+        //     .map(|(partition_id, store_handle)| {
+        //         let query_engine = QueryEngine::new(
+        //             store_handle.clone(),
+        //             QueryCache::new_handle(store_handle.clone()),
+        //         );
+        //
+        //         (partition_id, query_engine)
+        //     })
+        //     .collect();
```
jleibs: This is dead code now, right?

crates/store/re_datafusion/src/dataframe_query_provider.rs
timsaucer: Updated documentation.

crates/store/re_datafusion/src/dataframe_query_provider.rs
timsaucer: Carried over from previous implementation but I added text and a const at the top describing how this controls backpressure.

rerun_py/src/catalog/dataframe_query.rs
timsaucer: removed

crates/store/re_chunk_store/Cargo.toml
```diff
@@ -34,6 +34,7 @@ re_format.workspace = true
 re_log = { workspace = true, features = ["setup"] }
 re_log_encoding = { workspace = true, features = ["decoder"] }
 re_log_types.workspace = true
+re_protos.workspace = true
```
grtlr: Maybe we can get rid of this dependency by introducing a wrapper struct?

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -1,188 +1,628 @@
...
-            .into_iter()
-            .map(|(partition_id, query_engine)| DataframePartitionStream {
-                schema: Arc::clone(&schema),
-                query_expression: query_expression.clone(),
-                query_engine,
-                partition_id,
-            })
-            .map(Arc::new)
-            .collect();
+            // let cpu_join_result = cpu_join_result.map_err(|err| exec_datafusion_err!("{err}"))
```
grtlr: Maybe remove the commented out version?

```suggestion
```

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -1,188 +1,628 @@
...
+        if this
+            .cpu_join_handle
+            .as_ref()
+            .map(|h| h.is_finished())
+            .unwrap_or(false)
+        {
+            let join_handle = this
+                .cpu_join_handle
+                .take()
+                .expect("cpu join handle is None");
```
grtlr: nit: trying to avoid `expect`, is this equivalent?

```suggestion
        // If we have any errors on the worker thread, we want to ensure we pass them up
        // through the stream.
        if let Some(join_handle) = this
            .cpu_join_handle
            .as_ref()
            .filter(|h| h.is_finished())
        {
```

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -1,188 +1,628 @@
...
+) -> Result<(), DataFusionError> {
+    let mut current_stores: Option<(
+        String,
+        ChunkStoreHandle,
+        QueryHandle<StorageEngine>,
+        Vec<ChunkInfo>,
+    )> = None;
+    while let Some(chunks_and_partition_ids) = input_channel.recv().await {
+        for chunk_and_partition_id in chunks_and_partition_ids {
+            let (chunk, partition_id) = chunk_and_partition_id;
```
grtlr: nit

```suggestion
        for (chunk, partition_id) in chunks_and_partition_ids {
```

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -1,188 +1,628 @@
...
-        let query_engine = self.query_engine.clone();
-        let query_expression = self.query_expression.clone();
+    let pid_array = Arc::new(StringArray::from(vec![
+        partition_id.to_owned();
+        next_row[0].len()
+    ])) as Arc<dyn Array>;
+
+    let mut arrays = Vec::with_capacity(num_fields + 1);
+    arrays.push(pid_array);
+    arrays.extend(next_row);
```
grtlr: nit: This seems like a hot path. Mabye `next_row.insert(0, pid_array)` is sometimes faster?

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -779,31 +781,3 @@ impl ConnectionHandle {
         )
     }
 }
-
```
grtlr, line 799: I don't think we should introduce a `re_protos` dependency to quite low-level `re_chunk_store` just to remove that function.

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -1,188 +1,628 @@
...
+                    .clone();
+                chunks_to_receive.sort();
+
+                current_stores =
+                    Some((partition_id.clone(), store, query_handle, chunks_to_receive));
+            };
+
+            let (_, store, _, remaining_chunks) = current_stores
+                .as_mut()
+                .expect("current_stores should be set");
```
grtlr: nit: eliminate `expect` by replacing `current_store.is_none` above with https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert_with

crates/store/re_datafusion/src/dataframe_query_provider_wasm.rs
```diff
@@ -0,0 +1,482 @@
...
+
+                if query.num_rows() > 0 {
+                    this.current_query = Some((partition_id, query));
+                }
+            }
+
+            let (partition_id, query) = this
+                .current_query
+                .as_mut()
+                .expect("current_query should be Some");
```
grtlr, line 229: nit: eliminate this `expect` by turning the `while` into a `loop` with `break Some((partition_id, query))`

grtlr (changes requested):
I tried my best to review this, but tbh there is quite a bit of things that I just glanced over because I don't know the details 😬.

I will do a round of testing natively and with the webviewer now though.

I think we should revert `re_protos` dependency in `rec_chunk_store`.

Also, I think we should get rid of some of the `expect` calls. From what I can tell they were all fine, but every `expect` that we can avoid is one less potential viewer crash.

crates/store/re_datafusion/src/dataframe_query_provider.rs
timsaucer: These are not equivalent. We need to do a `take()` so we can take ownership of then handle, but I'll update it with code to remove the `expect` call.

crates/store/re_chunk_store/Cargo.toml
timsaucer: Moved over to re_dataframe

crates/store/re_datafusion/src/dataframe_query_provider_wasm.rs
timsaucer: I added in all the other changes, but I think this one is more readable as is.

nikolausWest:
Amazing getting this in!

Could you share a couple words on how this fixed #10639? I was under the impression that we had two problems: that we collected all data in memory before executing a query, and that the contexts would stick around in a notebook and effectively leak memory. I understand how this PR fixes the first issue but not the second (I have only read the PR description). In which ways am I misunderstanding things here?

timsaucer:
Your understanding is correct. The two issues are related. Because we were collecting it all in memory when creating the table provider and that table provider gets registered with the session context, the collection in memory never goes away until the table is deregistered.

With this change we are collecting per-partition into memory (upcoming PR to send intermediate results) and this happens during the stream *not* during the creation of the table provider. This makes the work far more lazy. We will still have a smaller collection stored in memory that contains the chunk info but this is typically very small. It's just the partition IDs, entity path, chunk ID, and a few extra bytes per chunk.

Does that answer the question?

nikolausWest:
> Does that answer the question?

Yes, thanks!

PR #10700, opened 2025-07-18
Fix C++ docs upload path

jprochazk:
It uploaded everything to `docs/cpp/{version}/html` instead of `docs/cpp/{version}/`


Wumpf (approved):
thank you!

PR #10704, opened 2025-07-18
Try to Fix Python Typing

ntjohnson1:
### Related
Our typing doesn't do what we want. We are running mypy in our base environment that doesn't have `rerun` installed so we can't properly check all of our transitive types. See https://github.com/rerun-io/rerun/pull/10690 and https://github.com/rerun-io/rerun/issues/10695 for a few footguns.

### What
Updates `py-lint` to be two targets`py-lint-non-sdk` which is now a separate `.ini` file since we care about that differently, and `py-lint-rerun` we still run in default environment so imperfect coverage but we catch more now.

I also went through and "fixed" the broken pieces. Somewhat haphazard combination of casts, bug fixes, and typing updates.

NOTES:
1. `pixi run -e py py-lint-rerun` isn't integrated anywhere yet. Unclear where in our CI stack this goes.
2. This will help here but is not a pre-requisite https://github.com/rerun-io/rerun/issues/10695 (however the combination of both may complicate things if we go the route of optional dependencies and testing)
3. General question around BoolArrayLike, FloatArrayLike, is there a reason for a single value not wrapped in array we don't take the python equivalent too?


rerun_py/pyproject.toml
```diff
@@ -19,7 +19,8 @@ dependencies = [
   "numpy>=2",
   "pillow>=8.0.0",          # Used for JPEG encoding. 8.0.0 added the `format` arguments to `Image.open`
   "pyarrow>=18.0.0",
-  "typing_extensions>=4.5", # Used for PEP-702 deprecated decorator
+  "typing_extensions>=4.5",
+  "rerun_notebook",         # I'm not sure if we really need to pin this at all today
```
ntjohnson1: This should get removed and just use the install `rerun-sdk[all]` from the other PR https://github.com/rerun-io/rerun/pull/10696

rerun_py/rerun_sdk/rerun/archetypes/depth_image_ext.py
```diff
@@ -23,6 +23,7 @@
         npt.NDArray[np.uint32],
         npt.NDArray[np.uint64],
         npt.NDArray[np.uint8],
+        np.ndarray[Any, np.dtype[np.floating | np.integer]],
```
ntjohnson1, line 26: This basically handles cv2 images without depending on it

pixi.toml
```diff
@@ -313,8 +313,13 @@ misc-fmt-check = "prettier --check '**/*.{yml,yaml,js,css,html}'"
 # Run first ruff fix, then ruff format, order is important see also https://twitter.com/charliermarsh/status/1717229721954799727
 py-fmt = "ruff check --fix --config rerun_py/pyproject.toml . && ruff format --config rerun_py/pyproject.toml ."
 py-fmt-check = "ruff check --config rerun_py/pyproject.toml . && ruff format --check --config rerun_py/pyproject.toml"
-py-lint = "mypy --config-file rerun_py/pyproject.toml --install-types --non-interactive --no-warn-unused-ignore"
-
+# Get non-internal things that are ok to not check rerun on
+py-lint-non-sdk = "mypy --config-file rerun_py/.non_sdk_mypy.ini --install-types --non-interactive --no-warn-unused-ignore"
+# Need to run this in env with rerun installed
+py-lint-rerun = "mypy --config-file rerun_py/pyproject.toml --install-types --non-interactive --no-warn-unused-ignore"
```
jleibs, line 319: In principle, if you put this under the `python-tasks` feature, then it should only be runnable in that context.

rerun_py/.non_sdk_mypy.ini
```diff
@@ -0,0 +1,25 @@
+[mypy]
+files = examples/python, docs/snippets, scripts
```
jleibs, line 2: I believe we want `examples/python` to consider the `rerun-sdk` itself when doing linting -- is that the case here if we have `ignore_missing_imports = True`?

pixi.toml
ntjohnson1: Gotcha. I forgot/didn't double check the task dependency bit. I'd like to at some point tie this in so a general `pixi run py-lint` would run the non-sdk in default and run this in the py env. I can move it for now though. I think newer pixi versions allow this task dependency specification a little more flexibly.

rerun_py/.non_sdk_mypy.ini
ntjohnson1: Right now we globally have `ignore_missing_imports` so this should be a non-change. If we want to default our non_sdk stuff to `ignore_missing_imports` then we need to specify ALL allowed packages with missing imports, which was a moderately sized list when I checked. If this is run in the base env when rerun-sdk isn't installed then this will ignore it and not complain

ntjohnson1: Ok I set things up so we can run `pixi py-lint` and it will check the sdk. However, that requires maintaining an opt our for every python package not installed in the base environment (see ini) and (pyproject). mypy doesn't seem to allow a per module or per file opt out configuration.

jleibs (approved):
Nice -- the explicit opt-outs is a little annoying but doesn't seem too bad.

Tried testing things locally and it seems to work.

PR #10706, opened 2025-07-18
fix breaking example  in rerun_notebook/example.ipynb

glk0:
### Related
Old signature of inialization of `Viewer` class:
https://github.com/rerun-io/rerun/blob/82d45ce7de4334a64c52f14bff61fe030aa664b1/rerun_notebook/src/rerun_notebook/__init__.py#L177-L185
The new signature
https://github.com/rerun-io/rerun/blob/a1bce5156d46c9d4c9e5610c46bcae47fd09acf4/rerun_notebook/src/rerun_notebook/__init__.py#L179-L188
### What
Fixes `rerun_notebook/example.ipynb` breaking due to reliance on older signature.


ntjohnson1 (approved):
Thanks for contributing for the first time and for raising the other issue!

PR #10713, opened 2025-07-21
Always bootstrap secondaries in plot view

teh-cmc:
The plot view visualizes scalar data within a specific time range, without any kind of time-alignment / bootstrapping behavior:
* For the scalar themselves, this is what you want: if you're trying to plot some data between t=100 and t=200, you don't want to display a point from t=20 (and _extended bounds_ will take care of lines crossing the limit).
* For the secondary components (colors, radii, names, etc), this is a problem though: you don't want your plot to change color depending on what the currently visible time range is! Secondary components have to be bootstrapped.

Before:

https://github.com/user-attachments/assets/4ad5e448-e6a9-46ed-bb02-7c88cb8cd3d1

After:

https://github.com/user-attachments/assets/03030ff5-d836-4e9f-9e70-2871c7afbdcc

### DIFF
--- a/crates/viewer/re_view_time_series/tests/basic.rs
+++ b/crates/viewer/re_view_time_series/tests/basic.rs
@@ -1,10 +1,11 @@
-use re_chunk_store::RowId;
-use re_log_types::TimePoint;
+use re_chunk_store::{RowId, external::re_chunk::ChunkBuilder};
+use re_log_types::{EntityPath, TimePoint, Timeline};
 use re_test_context::{TestContext, external::egui_kittest::SnapshotOptions};
 use re_test_viewport::TestContextExt as _;
+use re_types::{Archetype as _, blueprint::components::Corner2D};
 use re_view_time_series::TimeSeriesView;
 use re_viewer_context::{ViewClass as _, ViewId};
-use re_viewport_blueprint::ViewBlueprint;
+use re_viewport_blueprint::{ViewBlueprint, ViewContents};
 
 fn color_gradient0(step: i64) -> re_types::components::Color {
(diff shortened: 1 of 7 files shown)


grtlr (changes requested):
Nice!

This would be very valuable to have as a snapshot test, maybe based-off @jleibs's issue?

The scaffolding for this is quite minimal, here is an example: https://github.com/rerun-io/rerun/blob/main/crates/viewer/re_view_time_series/tests/basic.rs

grtlr (commented):
Thank you for adding the test!

Did we end up not needing the `Corner2D` after all to remove the legend? Speaking of which, the color change was probably what caused the initial problems with the legend.

Let me quickly try what happens when we add it again.

grtlr (approved):

crates/viewer/re_view_time_series/tests/basic.rs
```diff
@@ -358,3 +358,84 @@ fn run_view_ui_and_save_snapshot(
...
+                    .with_colors([re_types::components::Color::from_rgb(0, 255, 255)])
+                    .with_names(["muh_scalars_from_0"]),
+            )
+            .with_archetype(
+                RowId::new(),
+                TimePoint::from([(Timeline::log_tick(), 45)]),
+                &re_types::archetypes::SeriesLines::new()
+                    .with_widths([5.0])
+                    .with_colors([re_types::components::Color::from_rgb(255, 0, 255)])
+                    .with_names(["muh_scalars_from_45"]),
```
emilk, line 396: My `typos` complain about "muh". What does it mean?

PR #10721, opened 2025-07-22
Integrate basic MCAP loader from `rerun-io/rerun-mcap`

grtlr:
### What

This brings the basic MCAP support that previously lived in https://github.com/rerun-io/rerun-mcap over to the viewer.

### Open question

* Would it be possible to support `wasm32` in the future?

### TODO

* [x] Update `ARCHITECTURE.md` and diagram


grtlr:
Link checks are unrelated.

Issue #10710, opened 2025-07-20
Resolve WASM Duplicate Build Clash


ntjohnson1:
For our last release we had wasm for the viewer and the notebook plugin that were different but trying to pull form the same source. We should resolve this prior to our next release. It's also interesting we didn't have this in any of our alphas which might be related to our release running twice and not cleaning up properly. See here for more details
https://github.com/rerun-io/rerun/issues/10708

PR #10722, opened 2025-07-22
Upload notebook assets to `{version}/notebook/{file}`

jprochazk:
* Closes https://github.com/rerun-io/rerun/issues/10710


scripts/ci/publish_wheels.py
```diff
@@ -58,8 +58,8 @@ def publish_notebook_asset() -> None:
                 blob_base = f"version/{wheel_version}"
                 file_base = "extracted/rerun_notebook/static"
                 blobs = [
-                    (bucket.blob(f"{blob_base}/widget.js"), f"{file_base}/widget.js"),
-                    (bucket.blob(f"{blob_base}/re_viewer_bg.wasm"), f"{file_base}/re_viewer_bg.wasm"),
+                    (bucket.blob(f"{blob_base}/notebook/widget.js"), f"{file_base}/widget.js"),
+                    (bucket.blob(f"{blob_base}/notebook/re_viewer_bg.wasm"), f"{file_base}/re_viewer_bg.wasm"),
```
jprochazk, line 62: When adding this, I forgot that the file will be overwritten by the one generated for the `app.rerun.io` build... which doesn't build with the same flags as the one for the web viewer package and notebook assets.

I think it's a lucky coincidence that this worked on any of the pre-release versions. My solution here is to store the wasm file for the notebook in a separate place.

jprochazk: This doesn't break redap, because it hosts its own version of the notebook asset, which is extracted from the wheel. I tried grepping for `widget.js` in other places and didn't find anything else that needed to be changed

jleibs (approved):
Makes sense.

scripts/ci/publish_wheels.py
jleibs: Only thought I just had is whether should should consider naming the wasm itself differently for clarity? I'm not sure where this particular name comes from or why it was chosen.

jprochazk: I would prefer not to do that, just because of the number of other places where we hardcode this 😬...

PR #10723, opened 2025-07-22
Upload npm package to GCS on release

jprochazk:


rerun_js/scripts/publish.mjs
```diff
@@ -61,3 +61,19 @@ for (const pkg of unpublished) {
   const tag = inferTag(pkg.version);
   $(`npm publish --tag ${tag}`, { cwd: pkg.dir });
 }
+
+const tarballs = [];
```
jprochazk, line 65: Tested this manually by disabling the `npm publish` command invocation above, and checking that the resulting `rerun_js_package` directory is the same as the one produced by the `build_js` job

jleibs (approved):

Issue #10731, opened 2025-07-22
Insufficiently strict version constraint on `parking_lot`


sdd:
**Describe the bug**

re_chunk_store depends on `parking_lot::ArcRwLockReadGuard` which was only added in `parking_lot` 0.12.3, but the workspace Cargo.toml specifies the version dependency of `parking_lot` as 0.12 (https://github.com/rerun-io/rerun/blob/c6cde9ae0fa835faa01795a2b9b3f398041330ff/Cargo.toml#L267)

This caused my attempt to `cargo add rerun` to a project that already had parking_lot 0.12.0 to fail with `error[E0412]: cannot find type `ArcRwLockReadGuard` in crate `parking_lot``.

PR incoming.

PR #10732, opened 2025-07-22
make parking_lot dependency version stricter to prevent import failure

sdd:
### Related

Fixes https://github.com/rerun-io/rerun/issues/10731

### What

re_chunk_store depends on `parking_lot::ArcRwLockReadGuard` which was only added in `parking_lot` 0.12.3, but the workspace Cargo.toml specifies the version dependency of `parking_lot` as 0.12


grtlr:
Ah, great catch—sorry for this! Thank you for your issue and PR 🤩

grtlr (approved):

Issue #10509, opened 2025-07-03
Move `ApplicationId` to `StoreId` (previously in `StoreInfo`)


abey79:
### Situation

Currently, the recording id is interpreted in two different ways:

* as a universally unique address for a "bucket of chunk" (aka two logging processes targeting the same recording id will end up in the same  viewer recording)
* as a unique-within-dataset id for a partition

In both case, these are currently user-provided `str`.

In technical terms:

* The recording id is embedded in the `StoreId`, which appears first in the `StoreInfo`.
* The partition id is copied from there upon ingestion in the server

### Problems

* The two interpretations above are incompatible. For example, it's conceivable that many partition of separate datasets might be called "Episode 1", which would then not be universally unique.
* Because of that, we must use problematic workarounds: when downloading a remote partition, we randomise the `StoreId` instead of using the partition id (lest bad things happen with the "Episode 1" example). This means that in turn, should the locally downloaded partition be saved as a RRD, it will have a RecordingId/PartitionId that is *not* the one it has on the server, making round-tripping impossible.
  * This line of thought occurred to me while working on "local datasets" for the local dataframe API. It will need to extract a partition id to replicate the remote dataframe API.

### Proposal

The core problem is our indexing over recording id only, instead of `(app id, recording id)` (or `(dataset id, partition id)` in redap parlance). Concretely, we index a lot of stuff over `StoreId`, which contains the recording id. The proposal is thus to move the app id from `StoreInfo` to `StoreId` (noting that the former contains the latter). This has the potential to "magically" solve a lot of things with minimal impact (narrator: lol) on the code base.

### Original proposal

**Note**: this was my original idea. The proposal above stemmed from discussion and is better.

<details><summary>original proposal</summary>

TL;DR: Separate the recording id from the partition id.

* Add `Option<PartitionId>` is `StoreInfo` (we can default to considering `StoreInfo.store_id.id`.
* Update the python api accordingly: the user can set a partition id
  * The user should *also* be able to set the recording id, because the rrd merging semantics is useful for advanced cases. In that case, it should be *strongly* advised to use something that has a chance to be unique, e.g. a hash of whatever "experiment id" they have on their end.
* Update `dataplatform` ingestion accordingly

</details>

#### Requires a SDK-viewer grpc protocol update

[...]

Issue #10737, opened 2025-07-23
Update the JS/notebook APIs to use `StoreId` instead of recording id


abey79:
- follow-up to: https://github.com/rerun-io/rerun/pull/10742

`StoreId`, which uniquely identify logical recordings are displayed by the viewer, now contains both the application id and the recording id. However, the whole `WebHandle` JS api refers to recording by `recording_id: &str` only. This is both fragile and inconsistent, and needs to be updated.

This needs some design work though. A naive update could be to add `application_id: &str` arguments everywhere, or introduce a `RecordingStoreId { application_id: String, recording_id : String }` construct.

However:
- In the remote dataset context, we've now turned to using URI to refer to recording, including in notebook context when interacting with the viewer.
- As we want to narrow down the scope of our specifier ("focus on this view and this time range of this recording, using this blueprint") URIs scale better than hard-coded APIs.

In view of that, it would appear to make sense to more generally migrate to a URI-based approach to referring to recordings. (This is also related to the goal of better exposing the viewer's memory store as a server.)

Issue #10730, opened 2025-07-22
When the time comes, remove backward compatibility for 0.24


abey79:
Between 0.24 and 0.25, `StoreId` was changed to include the application id (see https://github.com/rerun-io/rerun/issues/10509). At the grpc level, this meant that:
- `StoreInfo` should no longer have an `application_id` (it moved to the `StoreId` it contains)
- legacy `ArrowMessage` and `BlueprintActivationMessage` are tricky to migrate because they only contain a `StoreId`, which used to not have an application id.

As a result, we introduced a number of mechanisms for migration, including:
- backward compatible grpc schema update
- "smart" grpc message to native object migration
- a cache system in the decoding streams

If/when we decided to drop 0.24 rrd compatibility (which could happen at the earliest for the 0.26 release), we should cleanup the code from all these mechanisms. Look for TODOs marked with this issue number.

abey79:
Tagged 0.26 for consideration.

nikolausWest:
For what it's worth, even though we only advertise version to version stability, I'd much prefer if we didn't break things that fast by a long shot and in general should consider that as an expensive move. When we do break older rrds I think we should also make sure to provide migration tools that are easy to use and work

abey79:
Fair. I don't have a strong opinion on when this needs to happen, if at all. I just wanted to leave breadcrumbs on the places where that clean-up would need to happen.

Also, now that you mention, it would appear that `rrd migrate` should do the right thing after my WIP PR, but could regress if the present issue is implemented naively. I'll update the description and make sure to leave some more breadcrumbs there as well.

(Even then, a clunky but feasible migration path would exist by chaining `rrd migrate` with multiple versions of the rerun CLI.)

Issue #9989, opened 2025-05-16
Identify recordings not just by `StoreId` but source + dataset/app-id + `StoreId`


Wumpf:
Today the Viewer identifies `EntityDb`s only by store id. This causes issues in the ui when trying to distinguish recodings that are dupliated to different servers and/or datasets, potentially with modifications.
Suggestion is to add a `StoreDescriptor` that contains all the necessary identifiers to distinguish recordings in all places

Sidenote: it would be nice to separate blueprint and recording store bookkeeping while we're on it here. We have a lot of places where we filter over and over again

PR #10742, opened 2025-07-23
Add `ApplicationId` in `StoreId`

abey79:
### The problem

For the longest time, "recording id" have been thought of as a unique identifier for a "logical recording" (aka a bucket of data that the viewer accumulates and present as a single in-viewer recording). This couldn't be further from the truth though, and only worked because the rerun sdk randomises the recording id by default.

This has been made glaringly obvious by the rerun server, where the partition id (which is unique to its enclosing dataset only) become the closest semantic match to pre-existing recording ids. As a matter of fact, a RRD's recording id is used as partition id when ingested in a server.

In code, this problem is apparent by `StoreId` (which we use to index stores) being essentially equivalent to a "recording id". This lead to all manners of conceptual and code issues, including the so-called `StoreHub` hell. This PR one step towards fixing this.

### A story of buckets

At the highest level, the rerun data model can be thought of as a two-tier hierarchy of buckets with "auto-grouping" semantics.

The first level is what we've called so far "application" on the SDK side (overdue for renaming), or datasets on redap side. Things that belong to a given application/dataset are presented by the viewer grouped as such. This is visibly the recording panel hierarchy.

The second level is the actual bucket of data (which contains chunks). In code, this is an `EntityDb`, which contains a `ChunkStore`, which contains a bunch of chunks. On the server, this a partition within a dataset. Again, these things are hierarchically presented as belonging to a top-level "group"/application/dataset.

By "auto-grouping" semantics, I mean that:
- Regardless of their source or the time they are ingested, two streams of data marked as belonging to the same "application"/dataset will end up within the same first-level bucket.
- Regardless of their source or the time they are ingested, two streams of data marked as belonging to the same "application"/dataset AND sharing the same recording id/partition id will end up in the exact same leaf bucket.

This auto-grouping properties are desirable:
- It's nice that distinct run of the same process are grouped in the same "application".
- It's nice to have the ability to log two distinct pieces of data into the same logical recording. For example, for each episode, your robot might produce 3 files: a video dump, a high-level log files (~MCAP), a low-level log file (some serial output of some motor control MCU). You might have 3 distinct loaders to load each of these into the same logical recording corresponding to the episode.

The former is a very common rerun workflow that exists since the beginning. The latter is a rarer, more advanced workflow, that is still desirable (and uncannily resembling the server-side "append workflow").

### A brave new `StoreId`

What this PR does is basically to update `StoreId` to reflect the two tier data architecture described above, such that using `StoreId` to index store is actually meaningful.

Here is the new `StoreId`:

```rust
/// A unique id per store.
///
/// The kind of store is part of the id, and can be either a [`StoreKind::Recording`] or a
/// [`StoreKind::Blueprint`].
///
/// ## Uniqueness
///
/// All data associated with a given `StoreId`, regardless of its source, is considered to belong
/// to the same logical recording. As such, when ingested in the viewer, they will be pooled into
/// a single store and displayed as a single, in-viewer recording. This can be very confusing if it
/// happens inadvertently. However, the ability to pool data sharing the same `StoreId` can be very
/// powerful. As a result, we do not want to _enforce_ uniqueness. Still, it is important to
/// understand what a `StoreId` is, and, to avoid footguns, default to make them unique.
///
/// In the context of the logging SDK, the application id is a mandatory, user-defined string. By
/// default, the recording id is a UUID, which ensures unique-by-default behavior. The user
/// can override the recording id though. In that case, the user is responsible for making the
/// application id/recording id pair unique or not, based on their needs.
///
/// In the context of remote recordings (aka a dataset's partition), the application id is the
/// dataset entry id, and the recording id is the partition id. The former is a UUID, and the latter
/// is, by definition, unique within the dataset entry. As a result, the uniqueness of the `StoreId`
/// is always guaranteed in this case.
#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
pub struct StoreId {
    kind: StoreKind,
    recording_id: RecordingId,
    application_id: ApplicationId,
}
```

### Aftermath

This PR deals with the following consequences of adding an application id inside `StoreId`. I'm tagging people who might be interested to check what I did.

- Data ingestion must implement backward compatibility, since data source (SDK, older RRD) will have `StoreId` missing the application id. This is done by looking at `SetStoreInfo` message (which did already contain an application id), caching them, and injecting the application id in subsequent messages. This requires `SetStoreInfo` message to arrive first, which is not guaranteed be always the case in practice. If not, messages will be dropped with a warning. cc @jprochazk @teh-cmc 
- Changes have been made to the data loader APIs to reflect the `StoreId` change. There should be no semantic change though, and this is Rust only (aka the CLI api hasn't been modified). The MCAP loader has been slightly updated for that. cc @grtlr @oxkitsune 
- `EntityDb` now caches a `StoreId` at the top-level for convenience (it used to be buried in the nested `ChunkStore`). cc @teh-cmc 
- Remote recording now have their `RecordingId` properly set to the partition id, which makes displaying the partition id slightly weird. cc @nikolausWest 
- I've disallowed making fully random `StoreId` because it's a potential footgun (I caught a bug while doing so). This however makes many test a tiny bit more noisy. cc @teh-cmc 
- I've shuffled a bit the `Arc` in the `StoreId`/`ApplicationId` types and introduced a new `RecordingId` type. cc @grtlr @teh-cmc 

To ~avoid~ put an end to  scope creep, the following things are post-poned for follow-ups:
- The JS/notebook API has not been changed for now, and everything is still referred to using `recording_id`. Finding the right recording is now best effort, and may lead to finding the wrong recording if several have the same `RecordingId` (which is now possible). This really needs fixing, but there are design considerations. cc @nikolausWest 
  - follow-up issue: https://github.com/rerun-io/rerun/issues/10737
- `StoreId`'s `Display` implementation must be removed as it can no longer be made to be universally looking good for UX. Also, the `Debug` impl could be made more compact. This means that all debugging code should use `Debug` instead of `Display`, and actual user-facing display must be fixed.
  - follow-up issue: https://github.com/rerun-io/rerun/issues/10746

### TODO

- [x] ~migration guide~ no need actually, no user facing change yet
- [x] redap sister-pr
- [x] rename `id` -> `recording_id` in `StoreId` proto (should be backward compatible)
- [x] run full check
- [x] make `StoreId` fields not `pub`
- [x] open an issue and add todo about `StoreId`'s `Display`

### Related

- closes: https://github.com/rerun-io/rerun/issues/10509
- follow-up: https://github.com/rerun-io/rerun/issues/10737
- follow-up: https://github.com/rerun-io/rerun/issues/10746
- breadcrumbs for 0.24 backward compatibility removal (if ever): #10730 
- part of: https://github.com/rerun-io/rerun/issues/9989
- sibling: https://github.com/rerun-io/dataplatform/pull/1357

### DIFF
--- a/crates/store/re_log_encoding/src/protobuf_conversions.rs
+++ b/crates/store/re_log_encoding/src/protobuf_conversions.rs
@@ -29,7 +29,7 @@ impl From<crate::Compression> for re_protos::log_msg::v1alpha1::Compression {
 /// [`crate::decoder::DecodeError::StoreIdMissingApplicationId`] if a message arrives before the
 /// matching `SetStoreInfo` message.
 ///
-/// The provided [`ApplicationIdCache`] must be shared across all calls for the same stream.
+/// The provided [`ApplicationIdInjector`] must be shared across all calls for the same stream.
 #[cfg(feature = "decoder")]
 #[tracing::instrument(level = "trace", skip_all)]
 pub fn log_msg_from_proto(


crates/top/rerun/src/commands/rrd/migrate.rs
```diff
@@ -107,6 +107,12 @@ fn migrate_from_to(from_path: &Utf8PathBuf, to_path: &Utf8PathBuf) -> anyhow::Re
 
     let mut errors = indexmap::IndexSet::new();
 
+    // TODO(#10730): if the legacy `StoreId` migration is removed from `Decoder`, this would break
+    // the ability of `rrd migrate` as well. If we want to keep the ability to migrate here, then
+    // the pre-#10730 app id caching mechanism must somehow be ported here.
+    // TODO(ab): The `StoreId` migration in `Decoder` requires `SetStoreInfo` to arrive before the
+    // corresponding `ArrowMsg`. Ideally this tool would cache orphan `ArrowMsg` until a matching
+    // `SetStoreInfo` arrives.
```
grtlr: The same applies for the other commands, right?

In particular, I think `rrd compare` will break and therefore our backwards-compatible CI will not pass.

abey79:
@rerun-bot full-check

crates/top/rerun/src/commands/rrd/route.rs
```diff
@@ -54,10 +54,7 @@ impl RouteCommand {
...
                 .as_ref()
                 .map(|id| ApplicationId { id: id.clone() }),
-            store_id: recording_id.as_ref().map(|id| StoreId {
-                id: id.clone(),
-                kind: StoreKind::Recording.into(),
-            }),
+            recording_id: recording_id.clone(),
         };
 
         let (rx, _) = read_raw_rrd_streams_from_file_or_stdin(path_to_input_rrds);
```
grtlr, line 62: I wonder if we should add the patching logic to the this function too?

Or do we assume that `rerun rrd route` and `rerun rrd stats` always use up-to-data protobuf versions?

crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -155,7 +155,30 @@ impl RecordingStreamBuilder {
...
+
+    /// Create a new [`RecordingStreamBuilder`] with the given [`StoreId`].
+    //
+    // NOTE: track_caller so that we can see if we are being called from an official example.
+    #[track_caller]
+    pub fn from_store_id(store_id: &StoreId) -> Self {
+        Self {
+            application_id: store_id.application_id().clone(),
+            store_kind: store_id.kind(),
+            recording_id: Some(store_id.recording_id().clone()),
```
grtlr, line 181: That's a weird diff, thanks GitHub 🙈 — looks fine though.

crates/store/re_log_encoding/src/decoder/stream.rs
```diff
@@ -73,14 +76,32 @@ impl StreamDecoder {
...
 
+    /// Read the next message in the stream, dropping messages missing application id that cannot
+    /// be migrated (because they arrived before `SetStoreInfo`).
     pub fn try_read(&mut self) -> Result<Option<LogMsg>, DecodeError> {
+        //TODO(#10730): remove this if/when we remove the legacy `StoreId` migration.
+        loop {
+            let result = self.try_read_impl();
+            if matches!(result, Err(DecodeError::StoreIdMissingApplicationId)) {
+                re_log::warn_once!(
+                    "Dropping message without application id which arrived before `SetStoreInfo`."
```
grtlr: Maybe output the recording id as part of the error?

crates/store/re_log_encoding/src/decoder/mod.rs
```diff
@@ -467,6 +590,103 @@ mod tests {
         }
     }
 
+    /// Test that legacy messages (aka `StoreId` without an application id) are properly decoded.
+    #[test]
+    fn test_decode_legacy() {
```
grtlr, line 618: Highly appreciate this test!

If we want to support Decoder-less patching (for protobuf only use cases like `rrd stats` and `rrd route`) could we please add a test for this too?

crates/store/re_log_types/src/lib.rs
```diff
@@ -101,83 +98,177 @@ impl std::fmt::Display for StoreKind {
...
 
-/// The user-chosen name of the application doing the logging.
-///
-/// Used to categorize recordings.
+/// The user-chosen name of the application doing the logging. In the context of a remote recording,
+/// this is the dataset entry id.
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
-pub struct ApplicationId(pub String);
+pub struct ApplicationId(Arc<String>);
```
grtlr, line 254: What's the reasoning behind the `Arc` here? Avoiding cloning of Strings? Have you considered interning those instead? Don't we look up by `RecordingId` and `ApplicationId` quite a bit?

crates/store/re_protos/src/lib.rs
```diff
@@ -160,6 +163,13 @@ pub enum TypeConversionError {
 
     #[error("could not parse url: {0}")]
     UrlParseError(#[from] url::ParseError),
+
+    #[error("internal error: {0}")]
+    InternalError(String),
+
+    //TODO(#10730): delete when removing 0.24 back compat
```
grtlr, line 170: Thank you for leaving these TODOs as breadcrumbs.

crates/store/re_log_types/src/lib.rs
abey79: We clone these quite a bit, directly and/or indirectly (via cloning of `StoreId`). My reasoning is that this is a good middle-ground between keeping the allocs in check, and leaking user data in interning (might be an issue in redap).

Also, I'm not too worried about performance. Remember that this stuff is used in `O(number of store)`context, which is much lower than the `O(entity)` or `O(component)` you have dealt with recently.

grtlr: I'm not sure the dataplatform even decodes to those types. IIRC they work on the protobuf-level (but I have been wrong with my assumptions recently).

Spoiler alert (by looking at your sibling PR): It does.

crates/top/rerun/src/commands/rrd/migrate.rs
abey79: I have updated and copied that comment throughout the `rerun rrd` impl. I've also update the relevant issue to clarify the impact on `rerun rrd *`.

crates/top/rerun/src/commands/rrd/route.rs
abey79: `read_raw_rrd_streams_from_file_or_stdin` already uses `Decoder` stuff, which does the patching logic. So we're good. The roll back discussion above applies though, and this is one of the place where I've copied that comment of mine.

crates/store/re_log_encoding/src/decoder/mod.rs
abey79: Again, `read_raw_rrd_streams_from_file_or_stdin` (which `rrd stats` and `rrd route` use) is based on `Decoder`, so we should be good. Or am I missing something?

grtlr: Yeah, you're right

abey79:
These protobuf breaking change warnings are produced:

```
{"path":"crates/store/re_protos/proto/rerun/v1alpha1/common.proto","start_line":80,"start_column":3,"end_line":80,"end_column":27,"type":"FIELD_SAME_JSON_NAME","message":"Field \"2\" with name \"recording_id\" on message \"StoreId\" changed option \"json_name\" from \"id\" to \"recordingId\"."}
{"path":"crates/store/re_protos/proto/rerun/v1alpha1/common.proto","start_line":80,"start_column":10,"end_line":80,"end_column":22,"type":"FIELD_SAME_NAME","message":"Field \"2\" on message \"StoreId\" changed name from \"id\" to \"recording_id\"."}
```

This is deemed acceptable because:
- the code breaking change have obviously been taken care of
- we don't use textual representation of proto messages

abey79:
@rerun-bot full-check

jleibs (approved):
I fear of edge-cases is of course concerning -- do we know of any real use-cases that result in deferred SetStoreInfo? As long as this is rare, this all seems reasonable. Really wishing we had prioritized moving to a stream with a header instead of the async setstoreinfo sooner.

abey79:
> do we know of any real use-cases that result in deferred SetStoreInfo?

I don't, but I've not worked on this kind of stuff a lot.

> Really wishing we had prioritized moving to a stream with a header instead of the async setstoreinfo sooner.

I obviously agree after this work 😅

PR #10744, opened 2025-07-23
Add `rerun auth` commands

jprochazk:
Adds the ability to authenticate via WorkOS to the Rerun CLI.

* Sibling PR for landing: https://github.com/rerun-io/landing/pull/1067

## TODO

- [x] Add `rerun auth login`
  - [ ] Store tokens locally somewhere
- [ ] Add `rerun auth revoke` (logout?)
- [ ] Add `rerun auth check`
- [ ] Add `rerun auth details`

## Open questions

- Q: How should we store these tokens? (Is plaintext in `.config/rerun` OK?)

### DIFF
--- a/crates/utils/re_auth/src/cli.rs
+++ b/crates/utils/re_auth/src/cli.rs
@@ -1,0 +1,312 @@
+use std::{
+    collections::HashMap,
+    io::Cursor,
+    sync::mpsc,
+    time::{Duration, Instant},
+};
+
+use base64::prelude::*;
+
+use crate::{Jwt, TokenError, workos};
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("failed to bind listener: {0}")]
+    BindError(std::io::Error),
+
+    #[error("HTTP server error: {0}")]
+    HttpError(std::io::Error),
+
+    #[error("failed to open browser: {0}")]
+    WebBrowser(std::io::Error),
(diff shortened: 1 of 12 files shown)


jprochazk:
@rerun-bot full-check

crates/utils/re_auth/src/cli.rs
```diff
@@ -0,0 +1,371 @@
...
+        webbrowser::open(&login_url).map_err(Error::WebBrowser)?;
+    } else {
+        p.println("Open the following page in your browser:");
+        p.println(&login_url);
+    }
+    p.inc(1);
+
+    // 3. Wait for callback
+    p.set_message("Waiting for browser...");
+    let auth = loop {
```
andrea-reale: probably it's very subjective, but I find the encoding of this part of the state machine (can we call it the server loop?) a bit hard to read. If you have time, are there ways you could extract out some logic to make the loop smaller (to read)?

andrea-reale (approved):
LGTM. Should we document more the auth commands? 

For someone without context it's not necessarily obvious what they are for, I guess.

crates/utils/re_auth/src/cli.rs
jprochazk: I agree, this whole loop can probably live in its own function, some parts of it can be further factored out. I'll try to simplify it a bit

PR #10745, opened 2025-07-23
Fix CPP Interface to VideoStream

ntjohnson1:
### What
Our code gen treats all enums as uint8_t however they can be other types. In this instance we have a uint32_t and by shaving off that precision we get the fairly ambiguous `missing_union_arm` error.  Reported on discord [here](https://discord.com/channels/1062300748202921994/1397335975364919392/1397335975364919392) but  it means the VideoStream isn't available on C++ right now.

```console
re_viewer_context::component_ui_registry] Failed to deserialize component of type rerun.components.VideoCodec: Failed to deserialize "rerun.components.VideoCodec"
```

See my first commit for the point fix then my second commit resolves this in code gen.


rerun_cpp/rerun_sdkConfig.cmake
jleibs, line 1: This file looks like maybe it was supposed to be git-ignored?

jleibs (approved):
I was going to ask you to make a change and then realized I should make sure the change was possible, and in doing so wrote the change, so I pushed it :laughing: 

Looks good.

ntjohnson1:
> I was going to ask you to make a change and then realized I should make sure the change was possible, and in doing so wrote the change, so I pushed it 😆
> 
> Looks good.

I can't approve my own PR but your changes LGTM as well.

PR #10751, opened 2025-07-24
Get rid of `CreatePartitionManifests` dev API

teh-cmc:
This hasn't been used in ages: PM creation was automated and asynchronized a millions years ago. And now that I want to make changes to `DataSource`, this will annoy me for no reason.


teh-cmc:
Oh but I think that's what the CreatePartitionTask uses under the hood :eyes: In which case this isn't a dev API at all

teh-cmc:
> Oh but I think that's what the CreatePartitionTask uses under the hood 👀 In which case this isn't a dev API at all

Nope, it's completely unused, we're good.

PR #10754, opened 2025-07-25
Ability to create dataset with specified ID

teh-cmc:
Only meant to revive datasets from the CLI at the moment.

* Sibling: https://github.com/rerun-io/dataplatform/pull/1364

### DIFF
--- a/crates/store/re_grpc_client/src/connection_client.rs
+++ b/crates/store/re_grpc_client/src/connection_client.rs
@@ -105,10 +105,14 @@ where
     pub async fn create_dataset_entry(
         &mut self,
         name: String,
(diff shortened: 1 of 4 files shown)


crates/store/re_grpc_client/src/connection_client.rs
```diff
@@ -105,10 +105,14 @@ where
     pub async fn create_dataset_entry(
         &mut self,
         name: String,
+        entry_id: Option<EntryId>,
```
abey79, line 108: tbh i'd rather not have this here and hard-code `None`, maybe with a todo to expose this in a sane way.

Issue #10753, opened 2025-07-24
Viewer should gracefully handle "bad datasets"


jleibs:
The first time the viewer hits a bad dataset in the catalog it fails.

This prevents good datasets from loading:

<img src="https://uploads.linear.app/93919c5e-abe4-4cd5-b77d-7e3eafe3e62d/d574be14-d0d1-42b9-aaf2-c847ee7e8114/01638c27-30b7-4bfe-a89d-6a9be2cc7816?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzkzOTE5YzVlLWFiZTQtNGNkNS1iNzdkLTdlM2VhZmUzZTYyZC9kNTc0YmUxNC1kMGQxLTQyYjktYWFmMi1jODQ3ZWU3ZTgxMTQvMDE2MzhjMjctMzBiNy00YmZlLWE4OWQtNmE5YmUyY2M3ODE2IiwiaWF0IjoxNzUzMzc0NjU3LCJleHAiOjMzMzIzOTM0NjU3fQ.hfWUZBx8K5aouOI_ROKoiLSdvi7z4t1JMrSk_optY-U " alt="image.png" width="1201" data-linear-height="815" />

PR #10755, opened 2025-07-25
Handle errors in redap browser more gracefully

lucasmerlin:
### Related

* closes https://github.com/rerun-io/rerun/issues/10753

### What

With this, a single table failing to load should not break the whole origin.

Instead of doing three request, one loading all datasets to register the tables, one loading the table details and one loading the dataset details, now we only do a single request to load all entries and a entry will be `Err` when loading the details _or_ loading the `TableProvider` fails. 

Screenshot 2025-07-28 at 11 46 23

Screenshot 2025-07-28 at 11 46 56


crates/viewer/re_redap_browser/src/tables_session_context.rs
```diff
@@ -149,13 +149,19 @@ async fn register_all_table_entries(
     }
```
lucasmerlin: I feel like the whole register_all_table_entries is kind of redundant with fetch_dataset_entries and fetch_table_entries 🤔

Wouldn't it be nicer if tables would be registered within those methods, since they are already loading all relevant datasets?

lucasmerlin: Refactored since this makes error handling a lot simpler

lucasmerlin:
Screenshot 2025-07-28 at 11 47 21

These errors still show up in the notification panels, should I remove them?

teh-cmc:
> These errors still show up in the notification panels, should I remove them?

I would keep them, but only print the contents of the `message` field. Is that easily achievable?

teh-cmc (approved):
This is fantastic and I want it

PR #8991, opened 2025-02-11
Fix `cpp` linking error on macOS due to missing `framework`s

grtlr:
### Related

* https://github.com/rerun-io/rerun/actions/runs/13260534162

### What

It looks like we were missing some macOS `framework`s that are required for linking with native certificates.

PR #10756, opened 2025-07-25
Fix rerun-cpp cmake export for OSX

reinar:
### Related

No related issues.

### What

Follow-up on PR #8991. In-tree cmake target was fixed, but corresponding cmake export config wasn't, this leads to link errors if rerun is imported through cmake config.


reinar:
While we at it, synchronized windows export target to what was added in #9447, otherwise it will likely fail as well.

grtlr (approved):
Ah good catch, and thank you for the PR!

reinar:
@grtlr branch is updated and can be merged with main, thank you!

grtlr:
Thank you, will merge when CI is green!

PR #10760, opened 2025-07-27
Quote string literals properly in `entity-queries.md` python snippets

kpbaks:
### What

Noticed, while skimming the docs, that some strings in the python code fence were not quoted properly.

image


teh-cmc (approved):
thanks

PR #10765, opened 2025-07-28
Set application id to file path when loading file

### LINKED ISSUE #10757: Set meaningful application ID and recording ID in dataloaders
It looks like we regressed on setting proper ids in our dataloaders. This can be observed, for example, when loading an MCAP file.

We should check all our loaders and see if they work as expected.

Where possible, we also should try to set meaningful recording properties.

oxkitsune:
### Related

* Closes #10757 
 
### What

Title

### DIFF
--- a/crates/store/re_data_loader/src/load_file.rs
+++ b/crates/store/re_data_loader/src/load_file.rs
@@ -37,18 +37,12 @@ pub fn load_from_path(
 
     re_log::info!("Loading {path:?}…");
 
-    // When loading a LeRobot dataset, avoid sending a `SetStoreInfo` message since the LeRobot
-    // loader handles this automatically.
-    let settings = if crate::lerobot::is_lerobot_dataset(path) {
-        &crate::DataLoaderSettings {
-            force_store_info: false,
-            ..settings.clone()
-        }
-    } else {
-        &crate::DataLoaderSettings {
-            application_id: Some(ApplicationId::from(path.display().to_string())),
-            ..settings.clone()
-        }
+    let application_id = ApplicationId::from(path.display().to_string());
+    let settings = &crate::DataLoaderSettings {
+        application_id: Some(application_id),
+        // When loading a LeRobot dataset, avoid sending a `SetStoreInfo` message since the LeRobot loader handles this automatically.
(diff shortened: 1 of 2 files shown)


crates/store/re_data_loader/src/load_file.rs
```diff
@@ -33,7 +35,7 @@ pub fn load_from_path(
         .into());
     }
 
-    re_log::info!("Loading {path:?}…");
+    re_log::info!("Loading {path:?}...");
```
grtlr: We usually prefer the ellipsis, is there a particular reason why you removed it?

crates/store/re_data_loader/src/load_file.rs
```diff
@@ -33,7 +35,7 @@ pub fn load_from_path(
...
 
     // When loading a LeRobot dataset, avoid sending a `SetStoreInfo` message since the LeRobot
     // loader handles this automatically.
```
grtlr: I think it still might be nice conceptually to set the `ApplicationId` to the folder name here. Or is there any other reason to not do this?

grtlr (commented):
Nice! Just some quick questions

crates/store/re_data_loader/src/load_file.rs
oxkitsune: My editor put a yellow box around it, will revert

crates/store/re_data_loader/src/load_file.rs
oxkitsune: This comment specifically refers to the fact we need to avoid sending the `SetStoreInfo` message because the loader sends in episode information for all episodes.

This allows us to load episodes in parallel while the viewer can be used.

grtlr: Oh, sorry GitHub didn't let me reference the lines below, I guess I should have been more clear:

Should we pass the `application_id` for the LeRobot loader here? It would be nice to always have that logic in the same place.

oxkitsune: Because the LeRobot loader manually sends the `SetStoreInfo` message, currently it doesn't use the settings for that, which I will change 👍

grtlr (approved):
Nice! Thank you for the changes

PR #10771, opened 2025-07-29
Fix unused imports due to `decoder` feature flag

grtlr:
CI is failing, features are responsible. Again 😞.

Also drive-by fix of some dead links.

* [ ] nightly


jprochazk:
`wasm-bindgen` was moved into its own org: https://github.com/wasm-bindgen/wasm-bindgen/issues/3359

grtlr:
Changed the link accordingly, the book still seems to be under the old URL though.

jprochazk (changes requested):
I think the link to `twiggy-paths` should be removed, and whatever we needed out of it should be inlined: https://blog.rust-lang.org/inside-rust/2025/07/21/sunsetting-the-rustwasm-github-org/

The link to the rustwasm issue can be retargeted to the new `wasm-bindgen` org

PR #10773, opened 2025-07-29
Use release version as JS docs version name

jprochazk:
`stable` is just a redirect, the actual docs are stored under `docs/js/{version}`. Previously it would both redirect to `docs/js/stable` and also upload docs to `docs/js/stable`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -241,7 +241,7 @@ jobs:
       CONCURRENCY: ${{ github.ref_name }}
       PY_DOCS_VERSION_NAME: ${{ inputs.release-type == 'final' && needs.version.outputs.final || 'dev' }}
       CPP_DOCS_VERSION_NAME: ${{ inputs.release-type == 'final' && 'stable' || 'dev' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -241,7 +241,7 @@ jobs:
       CONCURRENCY: ${{ github.ref_name }}
       PY_DOCS_VERSION_NAME: ${{ inputs.release-type == 'final' && needs.version.outputs.final || 'dev' }}
       CPP_DOCS_VERSION_NAME: ${{ inputs.release-type == 'final' && 'stable' || 'dev' }}
```
grtlr, line 243: Does it make sense to change CPP too?

grtlr (approved):

.github/workflows/release.yml
jprochazk: For C++ we upload the docs twice, once to named (the `stable`/`dev` one) and once to versioned (`0.24.0`). I don't know why we do it that way, I'm pretty sure I just replicated how it used to be on gh-pages :thinking: either way it works so I think it's fine to leave it as is:
https://ref.rerun.io/docs/cpp/stable/structrerun_1_1archetypes_1_1VideoStream.html

Issue #10761, opened 2025-07-28
Scale3D mono index out of bounds


CameronMatthew:
**Describe the bug**

`Scale3D` throws a warning instead of doing scaling.

```python
import rerun as rr

print(rr.__version__)
# 0.23.4

rr.init("scale_example", spawn=True)

rr.log("example_a", rr.Transform3D(scale=rr.Scale3D(2.0)), static=True)
# Couldn't read /example_a:rerun.components.Scale3D @ ((TimeInt::STATIC, RowId(185663C1D827AF633839bc73d55cfe71))): mono index out of bounds: 0 (len=3)

rr.log("example_b", rr.Transform3D(scale=rr.Scale3D([2.0, 2.0, 2.0])), static=True)
# Couldn't read /example_b:rerun.components.Scale3D @ ((TimeInt::STATIC, RowId(185663C1D82971E93839bc73d55cfe73))): mono index out of bounds: 0 (len=3)
```

**To Reproduce**
Run the minimal reproducible example above^.

**Expected behavior**

My transformation includes a scaling factor of 2.0.

**Backtrace**

```
Couldn't read /example_a:rerun.components.Scale3D @ ((TimeInt::STATIC, RowId(18566434E368E7AE1FDB8E2BF5CFDBA6))): mono index out of bounds: 0 (len=3)
```

**Desktop (please complete the following information):**
 - OS: macOS Sequoia 15.5

**Rerun version**

- Originally discovered on 0.23.4. I was able to reproduce from 0.22.0 up to 0.23.4.

grtlr:
Thank you for reporting! This looks like a problem with our Python SDK. As a quick fix, you could try omitting the `rr.Scale` part and only write `scale=2.0`.

ntjohnson1:
Our Scale3D type doesn't have a length attribute, which means there are a few different areas where we check for a length attribute assuming it is a scalar otherwise. That means we end up accidentally generating a Scale3D from a Scale3D as a matrix of values
```python
scale = rr.Scale3D([2.0, 2.0, 2.0])
scale_scale = rr.Scale3D(scale)
scale_scale.xyz
[[2.0, 2.0, 2.0],
[2.0, 2.0, 2.0],
[2.0, 2.0, 2.0]]
```

Simple fix is to just add __len__ but I need to see if I can route this through our code gen to solve more generally.

grtlr:
Aha, so we would implement something like `Vec3DLike` for the `Scale3D` component?

[...]

PR #10774, opened 2025-07-29
Add `__len__` to codegened Python datatypes

ntjohnson1:
### Related

* closes https://github.com/rerun-io/rerun/issues/10761

### What
Right now flatbuffers doesn't expose `__len__` and our codegen doesn't expose it either. A variety of our extensions check for the `__len__` attr in order to determine if something is a scalar or sized. Follow the `__array__` pattern in order to determine when we can generate a `__len__` automatically.


grtlr (approved):
Nit: I wonder if we should write a test case based on the user's repro?

LGTM, neat!

ntjohnson1:
> Nit: I wonder if we should write a test case based on the user's repro?

Did you mean something in addition to this?
56dfa579d344936f7a1da708f8ba1b79536ca5a4

grtlr:
Ah yes, sorry I was on my phone and didn't see the test

Issue #2499, opened 2023-06-21
Add support for Brown-Conrady Lens distortion


nikolausWest:
Most real world 3D computer vision applications need to make use of some kind of lens distortion model. The most common for a pinhole camera is Brown-Conrady lens distortion.

See OpenCV for the most common parametrization [here](https://docs.opencv.org/3.4/da/d54/group__imgproc__transform.html#ga7dfb72c9cf9780a347fbe3d1c47e5d5a)

And this tutorial from Tangram on distortion models: https://www.tangramvision.com/blog/camera-modeling-exploring-distortion-and-distortion-models-part-ii

- [ ] Users can log Brown-Conrady lens distorion models (at least radial + tangential)
- [ ] When combined with a logged pinhole model, the lens distortion is automatically applied both between 3D and 2D views (spaces), as well as between 2D distorted and 2D undistorted views (spaces)
- [ ] The presence and application of a lens distortion model is somehow indicated in the 3D view. Ideally these is an element in the 3D view that can be clicked such that the distortion model becomes selected

emilk:
Some of my thoughts about non-linear lens distortion in general

### Nomenclature
rectify = undistort, i.e. undoing the radial distortion

# Implementation
Implementing this will touch all of the following, and probably more:

* Ray projection/unprojection
* Rendering of the camera frustum (we should render it with curved edges)
* Depth cloud projection
* 2D-in-3D
* 3D-in-2D

We probably also want to support users logging things both to the distorted and rectified images. It is common for users to do their own rectifying and they might do object detection in either the distorted or undistorted image space. This suggests the radial distortion is a new type of transform component (alongside `Transform` and `Pinhole`. For instance:

```py
rr.log_bbox("world", object)
rr.log_transform("world/camera", extrinsics)
rr.log_pinhole("world/camera/pinhole", …)
rr.log_radial_distortion("world/camera/pinhole/distorted", …)
```

# Uses often do their own rectification:
rectified_image = cv.rectify(raw_image, distortion)

# Ideally you only need to log _one_ of the following, and Rerun will infer the other based on the logged distortion:
rr.log_image("world/camera/pinhole", rectified_image)
rr.log_image("world/camera/pinhole/distorted", raw_image)

# Maybe you do your object detection in the raw/distorted image:
rr.log_bbox("world/camera/pinhole/distorted/object", detect_object(raw_image))

# Or you do it in the rectified image:
[...]

PR #10779, opened 2025-08-01
More ROS2 message parsers

oxkitsune:
### What

Adds schema plugins for:
- `sensor_msgs/msg/CameraInfo`
- `sensor_msgs/msg/JointState`
- `std_msgs/msg/String`

### DIFF
--- a/crates/store/re_data_loader/src/mcap/decode.rs
+++ b/crates/store/re_data_loader/src/mcap/decode.rs
@@ -63,7 +63,6 @@ pub trait SchemaPlugin {
     /// # use anyhow::Error;
     /// # use mcap::Channel;
     /// # use re_chunk::Chunk;
-    /// # use re_sorbet::SorbetSchema;
     /// # use re_data_loader::mcap::decode::{
     /// #     McapMessageParser, ParserContext, PluginError, SchemaName, SchemaPlugin,
     /// # };


crates/store/re_data_loader/src/mcap/schema/sensor_msgs/camera_info.rs
```diff
@@ -0,0 +1,214 @@
...
+            width,
+            height,
+            distortion_model,
+            d,
+            k,
+            r: _,
+            p: _,
+            binning_x,
+            binning_y,
+            roi: _,
```
emilk: Instead of ignoring some fields (=throwing away user data), let's log them as custom components (like we already do for `distortion_model` etc).

We can also add a `TODO` to the relevant issue (maybe https://github.com/rerun-io/rerun/issues/2499 in this case?).

crates/store/re_data_loader/src/mcap/schema/sensor_msgs/camera_info.rs
```diff
@@ -0,0 +1,214 @@
...
+#[derive(Default)]
+pub struct CameraInfoSchemaPlugin;
+
+impl SchemaPlugin for CameraInfoSchemaPlugin {
+    fn name(&self) -> SchemaName {
+        "sensor_msgs/msg/CameraInfo".into()
+    }
+
+    fn parse_schema(&self, _channel: &mcap::Channel<'_>) -> Result<SorbetSchema, PluginError> {
+        Err(PluginError::Other(anyhow::anyhow!("todo")))
```
emilk: This is NOT a good error message

crates/store/re_data_loader/src/mcap/schema/sensor_msgs/joint_state.rs
```diff
@@ -0,0 +1,144 @@
...
+            efforts: ListBuilder::with_capacity(Float64Builder::new(), num_rows),
+        }
+    }
+}
+
+impl McapMessageParser for JointStateMessageParser {
+    fn append(&mut self, ctx: &mut ParserContext, msg: &mcap::Message<'_>) -> anyhow::Result<()> {
+        let joint_state = cdr::try_decode_message::<sensor_msgs::JointState>(msg.data.as_ref())
+            .map_err(|err| PluginError::Other(anyhow::anyhow!(err)))?;
+
```
emilk: use the destructor pattern to make sure we don't forget any field in `JointState`

crates/store/re_data_loader/src/mcap/schema/std_msgs/string.rs
```diff
@@ -0,0 +1,71 @@
...
+#[derive(Default)]
+pub struct StringSchemaPlugin;
+
+impl SchemaPlugin for StringSchemaPlugin {
+    fn name(&self) -> SchemaName {
+        "std_msgs/msg/String".into()
+    }
+
+    fn parse_schema(&self, _channel: &mcap::Channel<'_>) -> Result<SorbetSchema, PluginError> {
+        Err(PluginError::Other(anyhow::anyhow!("todo")))
```
emilk: Again, this is not a useful error message

emilk (changes requested):

crates/store/re_data_loader/src/mcap/schema/sensor_msgs/camera_info.rs
oxkitsune: Good catch. Had a bit of fun with arrow structs and managed to get everything in 👍

crates/store/re_data_loader/src/mcap/schema/std_msgs/string.rs
```diff
@@ -0,0 +1,66 @@
...
+        Self {
+            texts: Vec::with_capacity(num_rows),
+        }
+    }
+}
+
+impl McapMessageParser for StringMessageParser {
+    fn append(&mut self, _ctx: &mut ParserContext, msg: &mcap::Message<'_>) -> anyhow::Result<()> {
+        let string_msg = cdr::try_decode_message::<std_msgs::StringMessage>(&msg.data)?;
+        self.texts.push(string_msg.data);
```
emilk: You can use the destructor pattern here too

emilk (approved):
Nice!

Issue #10782, opened 2025-08-01
First person eye control doesn't support scroll-to-zoom


MichaelGrupp:
**Describe the annoyance**

The first person eye control in 3D views is nice, but I found it annoying that mouse scrolling has no effect there (as opposed to the orbit mode and the help menu that says that scrolling zooms).

**To Reproduce**

Open a 3D spatial view, move the mouse inside and scroll / pinch-to-zoom.
The behavior is different depending on the eye type...

* `Orbital`: zooms
* `First person`: nothing

**Expected behavior**

Scrolling does something also in first person mode.

**Desktop (please complete the following information):**
 - OS: macOS & Ubuntu
 - rerun 0.24 / main

---

**Proposal**

I would just move the eye forward/backward along the optical axis in this case. I patched rerun locally with this behavior and found it quite satisfying to use, as you can then navigate through a scene also only using the mouse (left-click head rotation, right-click pan, scroll zoom). PR: https://github.com/rerun-io/rerun/pull/10783

Here's a demo where I use only the mouse with that patch:

https://github.com/user-attachments/assets/64edac62-cd08-4818-8541-3e0e838afd52

PR #10783, opened 2025-08-01
Support scroll-/pinch-to-zoom with first person eye control

MichaelGrupp:
Move the camera along the forward axis of the eye on scroll/ pinch-to-zoom user input. This allows nicer navigation using only the mouse and is more consistent with the Orbital mode.

Closes: #10782


MichaelGrupp:
See ticket for a demo video.

PR #10786, opened 2025-08-04
Show warning about running inside of Docker in top panel

emilk:
### Related
* Follow-up to https://github.com/rerun-io/rerun/pull/10306

### What
We would already print a warning to the terminal, but now we also show it in the notification panel (and as a toast/popup):

Screenshot 2025-08-04 at 11 03 47


crates/viewer/re_viewer/src/lib.rs
```diff
@@ -346,3 +346,16 @@ pub fn reset_viewer_persistence() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// Hook into [`re_log`] to receive copies of text log messages on a channel,
+/// which we will then show in the notification panel.
+pub fn register_text_log_receiver() -> std::sync::mpsc::Receiver<re_log::LogMsg> {
+    let (logger, text_log_rx) = re_log::ChannelLogger::new(re_log::LevelFilter::Info);
+    if re_log::add_boxed_logger(Box::new(logger)).is_err() {
+        // This can happen when `rerun` crate users call `spawn`. TODO(emilk): make `spawn` spawn a new process.
```
Wumpf: this moved comment seems outdated: `spawn` _does_ spawn a new process, even in Rust.
What doesn't is `run` and that's very much intentional

Wumpf (approved):
lgtm, surely a good general fix! But I think we should show the warning permanently in the status bar like we do for software rasterizers

emilk:
> But I think we should show the warning permanently in the status bar like we do for software rasterizers

I think we should move the software rasterizer warning to the notification panel

Wumpf:
it is nice though that it is sticky and included on screenshots. It helps us with bug reports

emilk:
Good point 🤔

emilk:
I added a warning label to the top panel now, per Andreas' recommendation.

image

emilk:
I should make sure `is_docker()` is memoized though… hold on

Wumpf (approved):
nice, thanks for the follow-up! lgtm

PR #10791, opened 2025-08-04
Low-level Arrow conversion of `protobuf`-encoded MCAP messages

grtlr:
### Related

* https://github.com/apache/arrow-rs/issues/8033

### What

> [!IMPORTANT]
> Admittedly, the way this code is called is still a mess. My next steps will be cleaning up the general MCAP data loader architecture, but I wanted to get the `protobuf` decoding out separately.

This PR implements low-level Arrow conversion of `protobuf`-encoded MCAP messages using runtime reflection from the `prost_reflect` crate. Given a `protobuf`-based MCAP schema a dynamic set of Arrow builders is created. Each message is then encoded recursively using these builders.

This PR also contains a drive-by fix for a missing field in `PointCloud2`, happy to pull that out if requested!

### Todos

There is some performance concern with the current implentation, namely that we re-decode the schema once for every MCAP chunk. That will require re-architecting some of the higher-level code, which I will be doing in a separate PR.

* Follow-up: #10807


crates/store/re_data_loader/src/mcap/decode.rs
```diff
@@ -1,9 +1,12 @@
...
 
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{HashMap, hash_map::Entry},
+    sync::Arc,
+};
 
 use anyhow::Context as _;
-use mcap::Message;
+use mcap::{Message, Schema};
```
jleibs: Nit: given how many Schemas we have flying around this might warrant an alias like McapSchema

crates/store/re_data_loader/src/loader_mcap.rs
```diff
@@ -145,13 +146,11 @@ fn from_schema(schema: &Arc<::mcap::Schema<'_>>) -> AnyValues {
...
     } = schema.as_ref();
 
-    let blob = components::Blob(data.clone().into_owned().into());
-
     // Adds a field of arbitrary data to this archetype.
     AnyValues::new("rerun.mcap.Schema")
         .with_field("id", Arc::new(UInt16Array::from(vec![*id])))
         .with_field("name", Arc::new(StringArray::from(vec![name.clone()])))
-        .with_component::<components::Blob>("data", vec![blob])
+        .with_field("data", Arc::new(BinaryArray::from(vec![data.as_ref()])))
```
jleibs, line 153: While I'm happy that we're storing this, the fact that it lives in separate static component is kind of annoying for some practical use-cases since it means the schema and the binary payloads will live in separate chunks that need to be tracked together.

I'm wondering if it's worth actually adding this schema-data to the Field-metadata associate with the "Binary Message Payload" column itself (if and when we create it).  This would make that column totally self-describing and able to be decoded viewer-side using the same reflection techniques you've implemented here.

jleibs (approved):
Nice! This is quite elegant.

crates/store/re_data_loader/src/mcap/schema/protobuf.rs
```diff
@@ -0,0 +1,327 @@
...
+        Kind::Bytes => DataType::Binary,
+        Kind::Message(message_descriptor) => {
+            let fields = message_descriptor
+                .fields()
+                .map(|f| arrow_field_from(&f))
+                .collect::<Fields>();
+            DataType::Struct(fields)
+        }
+        Kind::Enum(_) => {
+            // TODO(grtlr): The enum support in Arrow is currently limited:
```
grtlr: Use `TODO()` that's picked up by zombie issues task.

crates/store/re_data_loader/src/loader_mcap.rs
grtlr: Yeah, that makes sense! From a viewer perspective adding this as a static chunk makes sense though, so I could see us supporting both ways going forward.

Not sure if Arrow field metadata is the right place though, or if this will end up in the promised out-of-band metadata (cc @teh-cmc).

PR #10795, opened 2025-08-04
Add support for layers to the python registration APIs

teh-cmc:
The gRPC API was already exposed, so this is fairly light.

I went with a clamping-semantics-like approach for the batched registration API, i.e. we will either truncate or extend the list of layers automatically as needed. As I implemented a layer-aware `dataset-util`, this really felt like the most pleasant API to work with.

* Said `dataset-util` implementation: https://github.com/rerun-io/dataplatform/pull/1411

### DIFF
--- a/rerun_py/rerun_bindings/rerun_bindings.pyi
+++ b/rerun_py/rerun_bindings/rerun_bindings.pyi
@@ -1367,7 +1367,7 @@ class DatasetEntry(Entry):
         Parameters
         ----------
         recording_uri: str
-            The URI of the RRD to register
+            The URI of the RRD to register.
 
         recording_layer: str
             The layer to which the recording will be registered to.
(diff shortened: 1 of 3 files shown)


rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1357,7 +1357,7 @@ class DatasetEntry(Entry):
 
         """
 
-    def register(self, recording_uri: str, timeout_secs: int = 60) -> str:
+    def register(self, recording_uri: str, recording_layer: str = "base", timeout_secs: int = 60) -> str:
```
teh-cmc: neither timeout nor recording_layer should be positional though, right?

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1379,7 +1382,7 @@ class DatasetEntry(Entry):
 
         """
 
-    def register_batch(self, recording_uris: list[str]) -> Tasks:
+    def register_batch(self, recording_uris: list[str], recording_layers: list[str] = []) -> Tasks:
```
teh-cmc: same here would be my guess

PR #10799, opened 2025-08-05
Resolve CPP Arrow Build

### LINKED ISSUE #10796: C++ SDK compilation fails with CMake > 4.0
**Describe the bug**

`rerun_cpp_sdk` fails to compile on macOS with CMake > 4.0.

Since CMake 4.0, compatibility with CMake < 3.5 has been removed. That means that any project with `cmake_minimum_required(<3.5)` will fail to compile. That is the case for some dependencies of arrow (as tracked [here](https://github.com/apache/arrow/issues/45985)) and consequently makes Rerun fail to compile.

**Workaround**

Until arrow updates its dependencies, Rerun can be compiled by setting the environment variable `export CMAKE_POLICY_VERSION_MINIMUM=3.5`. It would be nice to be able to pass it directly as a CMake flag via `-DCMAKE_POLICY_VERSION_MINIMUM=3.5`, but that is somehow not forwarded by Rerun (as mentioned briefly on [Discord](https://discord.com/channels/1062300748202921994/1075873257124810852/1401910471656734752)).

**To Reproduce**

1. Download and extract the artifact: [sdk](https://github.com/rerun-io/rerun/releases/download/0.24.0/rerun_cpp_sdk.zip).
2. Compile manually with: `cmake -Bbuild` and `cmake --build build`.
3. Compilation fails when compiling arrow (see screenshot and backtrace for the error).

The same goes if you try to use Rerun via `FetchContent` (for example when compiling the dna example).

**Expected behavior**

`rerun_cpp_sdk` should compile.

**Screenshots**

Image

**Backtrace**

[arrow_cpp-build-out.log](https://github.com/user-attachments/files/21591403/arrow_cpp-build-out.log)

Relevant log:

```log
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

**Desktop:**

 - OS: macOS Sequoia 15.5
 - CMake: 4.0.3

**Rerun version**

0.24.0, taken directly from the release page

**Additional context**

I have some colleagues reporting this on Rerun 0.22.0 as well. It probably affects all Rerun versions.

ntjohnson1:
### Related
* closes #10796 

### What
This checks for newer CMake and automatically applies an automatic workaround.
If a user explicitly provides a `CMAKE_POLICY_VERSION_MINIMUM` we defer to that.

Verified by running the cmake commands from `cpp-compile-all`
```console
cmake -G 'Ninja' -B build/debug -S . -DCMAKE_BUILD_TYPE=Debug
cmake --build build/debug --config Debug --target ALL
```
This works

```console
cmake -G 'Ninja' -B build/debug -S . -DCMAKE_BUILD_TYPE=Debug -DCMAKE_POLICY_VERSION_MINIMUM=4.0
cmake --build build/debug --config Debug --target ALL
```
This intentionall fails

### DIFF
--- a/rerun_cpp/download_and_build_arrow.cmake
+++ b/rerun_cpp/download_and_build_arrow.cmake
@@ -69,6 +69,8 @@ function(download_and_build_arrow)
 
     if(CMAKE_VERSION VERSION_GREATER_EQUAL "4.0")
         # TODO(#10798): Arrow can't support CMake 4.0 yet
+        # See arrow issue linked from our internal issue above for more details.
+        # See here for what this does https://cmake.org/cmake/help/latest/variable/CMAKE_POLICY_VERSION_MINIMUM.html
         set(VERSION_PATCH -DCMAKE_POLICY_VERSION_MINIMUM=3.5)
     else()
         set(VERSION_PATCH "")
--- a/rerun_cpp/download_and_build_arrow.cmake
+++ b/rerun_cpp/download_and_build_arrow.cmake
@@ -68,7 +68,7 @@ function(download_and_build_arrow)
     endif()
 
     if(CMAKE_VERSION VERSION_GREATER_EQUAL "4.0")
-        # TODO(nick): Remove when resolved  # See https://github.com/apache/arrow/issues/45985
+        # TODO(#10798): Arrow can't support CMake 4.0 yet
         set(VERSION_PATCH -DCMAKE_POLICY_VERSION_MINIMUM=3.5)
     else()
         set(VERSION_PATCH "")
--- a/rerun_cpp/download_and_build_arrow.cmake
+++ b/rerun_cpp/download_and_build_arrow.cmake
@@ -74,6 +74,11 @@ function(download_and_build_arrow)
         set(VERSION_PATCH "")
     endif()
 
+    # Allow for CMAKE_POLICY Override
+    if(CMAKE_POLICY_VERSION_MINIMUM)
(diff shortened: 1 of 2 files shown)


rerun_cpp/download_and_build_arrow.cmake
```diff
@@ -67,6 +67,18 @@ function(download_and_build_arrow)
         set(ARROW_CMAKE_PRESET ninja-release-minimal)
     endif()
 
+    if(CMAKE_VERSION VERSION_GREATER_EQUAL "4.0")
+        # TODO(#10798): Arrow can't support CMake 4.0 yet
```
Wumpf, line 71:
```suggestion
        # TODO(apache/arrow#45985): Arrow can't support CMake 4.0 yet
```

edit: doesn't work after all

Wumpf: Imho this should link to https://cmake.org/cmake/help/latest/variable/CMAKE_POLICY_VERSION_MINIMUM.html and try to explain a bit more what's going on; or maybe just link to the origin issue to add context. Had no idea about this stuff and I doubt it's all that common knowledge :)

ntjohnson1: I can add a link to the cmake doc. The arrow issue itself actually discusses it a bit and lists the workaround used here, so it felt like that was the best link. Basically we depend on arrow who depends on other things who don't work with cmake > 4.0

Issue #10800, opened 2025-08-05
Rust SDK depends on rfd (file dialog library)


alokpr:
Rust SDK for sending data to the viewer should not depend on a UI library. Is this intended?

emilk:
Thanks for reporting this - this is definitely a mistake somewhere. We'll take a look!

emilk:
Maybe we can do a `cargo tree -p re_sdk` and make sure there are no `rfd`, `egui`, `cocoa`, `winit` etc in that list (on CI I mean) 🤔

alokpr:
@emilk @Wumpf Thanks for the quick fix! Any estimate on when `0.24.1` will be released?

Wumpf:
We want to release it fairly timely, likely at some point next week, maybe already this one

PR #10802, opened 2025-08-05
Fix unnecessary `rfd` and wayland dependencies in SDK crate

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/10800

### What

Adds a test to detect this kind of thing. The test is fairly quick to run and is part of all CI subsets that run Rust checks.

### DIFF
--- a/scripts/ci/rust_checks.py
+++ b/scripts/ci/rust_checks.py
@@ -248,13 +248,16 @@ def denied_sdk_deps(results: list[Result]) -> None:
     results.append(run_cargo("install", "--locked cargo-tree@^0.29.0"))
 
     # Sampling of dependencies that should never show up in the SDK, unless the viewer is enabled.
+    # They are ordered from "big to small" to make sure the bigger leaks are caught & reported first.
+    # (e.g. `re_viewer` depends on `rfd` which is also disallowed, but if re_viewer is leaking, only report `re_viewer`)
     disallowed_dependencies = [
         "re_viewer",
         "wgpu",
         "egui",
         "winit",
         "rfd",  # File dialog library.
-        "objc2-ui-kit",  # MacOS system libraries.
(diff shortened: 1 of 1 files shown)


scripts/ci/rust_checks.py
```diff
@@ -197,44 +214,69 @@ def sdk_variations(results: list[Result]) -> None:
...
+
+    # Sampling of dependencies that should never show up in the SDK, unless the viewer is enabled.
+    disallowed_dependencies = [
+        "re_viewer",
+        "wgpu",
+        "egui",
+        "winit",
+        "rfd",  # File dialog library.
+        "objc2-ui-kit",  # MacOS system libraries.
+        "wayland-sys",  # Linux windowing.
```
emilk: We could add "cocoa" to this list: https://crates.io/crates/cocoa

Also sort the list :)

emilk (approved):
Nice!

It would be nice to see a red CI on the first build step before you committed the fix, but I assume you at least ran the new test locally first?

scripts/ci/rust_checks.py
Wumpf: > We could add "cocoa" to this list: [crates.io/crates/cocoa](https://crates.io/crates/cocoa)

sure why not, but it actually doesn't show up in our cargo.lock 🤷 
(good! because it is indeed deprecated in favor of objc2!)

The sorting of the list is rather deliberate from "big to small" - `re_viewer` will obviously draw in `rfd`, so if `re_viewer` leaked in I want an error about that rather than `rfd`.
I figure the better way to go about that might be to collect all disallowed crates 🤔

emilk: Ah, that ordering makes perfect sense! Please add a comment to that effect :)

Wumpf:
> but I assume you at least ran the new test locally first?

yup!

PR #10803, opened 2025-08-05
Fix disallowed color method declarations in `clippy.toml`

Wumpf:
It checked for `egui` colors rather than `ecolor` which missed a few cases and caused false warnings for violated `expect` when running `cargo clippy -p rerun`.


emilk (approved):
Nice

PR #10804, opened 2025-08-05
Optimize memory usage in `StreamingDecoder`

teh-cmc:
TL;DR: Don't poll on the underlying reader more than we need to, it triggers a cascade of failures that prevent all memory optimizations from kicking in.

See inline comments for details.

* In action: https://github.com/rerun-io/dataplatform/pull/1418


teh-cmc:
Getting this in so I get a chance to deploy and test at scale this morning.

PR #10806, opened 2025-08-05
Update/loguru, cmakelists toolchain pathing and readme

flightwave-maxwell:
### What

CMakeLists.txt has a pathing issue to the conan toolchain file.
* Fixed

Conan file was missing loguru
* Package added

Readme in rerun_cpp to address installing for local conan repo usage
* Updated


rerun_cpp/CMakeLists.txt
```diff
@@ -135,6 +135,17 @@ endif()
...
+# Loguru dependency.
+# Try to find loguru package (e.g., from Conan), otherwise it should be provided by the parent project.
+find_package(loguru QUIET)
+if(loguru_FOUND)
+    message(STATUS "Found loguru package")
+    target_link_libraries(rerun_sdk PRIVATE loguru::loguru)
+else()
+    message(STATUS "loguru package not found - assuming it will be provided by parent project")
+endif()
+
```
Wumpf, line 148: The Rerun C++ SDK isn't dependent on loguru so this shouldn't go here. It needs to be isolated to the examples that need it

rerun_cpp/download_and_build_arrow.cmake
```diff
@@ -4,6 +4,13 @@
 # Tries to build an as small as possible version of Arrow that is compatible with the Rerun C++ SDK.
 function(download_and_build_arrow)
     include(ExternalProject)
+    
+    # Convert CMAKE_TOOLCHAIN_FILE to absolute path if it's set
+    if(CMAKE_TOOLCHAIN_FILE)
+        get_filename_component(ARROW_TOOLCHAIN_FILE "${CMAKE_TOOLCHAIN_FILE}" ABSOLUTE)
+    else()
+        set(ARROW_TOOLCHAIN_FILE "")
+    endif()
```
Wumpf, line 13: why is this needed?

rerun_cpp/test_package/README.md
```diff
@@ -0,0 +1,24 @@
+# Test Package for Rerun SDK
```
Wumpf, line 1: I'd argue this is more of an example than a test. Either way calling this `test_package` is not useful since it doesn't say what it tests and how it relates to other tests we have.
I'd suggest instead putting this in `examples/cpp` and call it something like `conan_build`.

Rest of the readme should also present it as an example instead

rerun_cpp/README.md
```diff
@@ -156,3 +156,84 @@ As of writing we tested the SDK against:
 * Apple Clang 14, 15
 * GCC 9, 10, 12
 * Visual Studio 2022
+
+## Building with Conan
```
Wumpf, line 160: This section should go under `## Build & distribution`

rerun_cpp/README.md
```diff
@@ -156,3 +156,84 @@ As of writing we tested the SDK against:
...
+
+The Rerun C++ SDK can be built and distributed using [Conan](https://conan.io/), which provides better dependency management and easier integration into existing C++ projects.
```
Wumpf, line 162: I don't think the project should be playing favorites with any of the myriad of build systems out there
```suggestion
The Rerun C++ SDK can be built and distributed using [Conan](https://conan.io/).
```

rerun_cpp/README.md
```diff
@@ -156,3 +156,84 @@ As of writing we tested the SDK against:
...
+
+2. **Verify installation**:
+   ```bash
+   conan --version
+   ```
+
+3. **Configure Conan profile** (first time setup):
+   ```bash
+   conan profile detect --force
+   ```
```
Wumpf, line 179: here and elsewhere: style deviates from the rest of the doc for no reason: don't indent and enumerate when if it isn't done anywhere else in here and doesn't add value

rerun_cpp/README.md
```diff
@@ -156,3 +156,84 @@ As of writing we tested the SDK against:
...
+
+### Building the SDK with Conan
+
+1. **Install dependencies**:
```
Wumpf, line 183: would it be also possible to do this with the buildzip? having to clone the rerun repo might be a bit tedious since it's fairly big and also means that building from there the user needs a rust compiler (text should mention that as well!). Also, as-is this needlessly encourages using unstable versions - if done via repo, we should encourage users to check out `latest`

rerun_cpp/README.md
```diff
@@ -156,3 +156,84 @@ As of writing we tested the SDK against:
...
+   ```bash
+   cd rerun_cpp
+   conan install . --build=missing -s build_type=Release
+   ```
+
+2. **Configure and build**:
```
Wumpf, line 189: confused: this isn't actually building with conan, the build is still with CMake? What does the previous step actually install?

rerun_cpp/README.md
```diff
@@ -156,3 +156,84 @@ As of writing we tested the SDK against:
...
+   ```
+
+### Creating a Local Conan Package
+
+To use the Rerun C++ SDK as a dependency in other projects:
+
+1. **Export the recipe to your local Conan cache**:
+   ```bash
+   cd rerun_cpp
+   conan export . --name=rerun_cpp_sdk --version=0.25.0
```
Wumpf, line 203: can the version be set to latest? I want to avoid mentioning outdated versions (in this case this isn't even released yet)

rerun_cpp/README.md
```diff
@@ -156,3 +156,84 @@ As of writing we tested the SDK against:
...
+* **Apache Arrow** - For data serialization
+* **loguru** - For logging support (optional)
+* **Boost** - Various utilities
+* **Other system dependencies** - Automatically resolved
+
+### Troubleshooting Conan Build
+
+* **Missing loguru**: The build will warn if loguru is not found but will continue without logging support
+* **Arrow build issues**: Make sure you have sufficient disk space and memory for the Arrow build
+* **Toolchain issues**: Ensure you're using the Conan-generated toolchain file in your CMake configuration
```
Wumpf, line 239: this whole section sounds super generic and isn't actually useful

rerun_cpp/test_package/README.md
Wumpf: also, please explain the interplay of cmake & conan file in here - I've never used conan before and don't know why they're needed respectively

rerun_cpp/tests/CMakeLists.txt
```diff
@@ -31,7 +31,16 @@ list(APPEND rerun_sdk_tests_SRC "../docs/readme_snippets.cpp")
...
 
 # Include arrow explicitly again, otherwise the arrow headers won't be found.
-target_link_libraries(rerun_sdk_tests PRIVATE loguru::loguru Catch2::Catch2 rerun_sdk rerun_arrow_target)
+# Link loguru if available, otherwise proceed without it
+if(TARGET loguru::loguru)
+    target_link_libraries(rerun_sdk_tests PRIVATE loguru::loguru Catch2::Catch2 rerun_sdk rerun_arrow_target)
+else()
+    message(WARNING "loguru::loguru target not found. Tests will be built without loguru logging support.")
+    target_link_libraries(rerun_sdk_tests PRIVATE Catch2::Catch2 rerun_sdk rerun_arrow_target)
+endif()
```
Wumpf, line 46: That doesn't make sense. The tests don't work without loguru. Warning here just leads to compile error later

Wumpf (changes requested):
thanks for your contribution!

I'm open to the idea of adding more examples & docs for using other build systems if we can verify these on CI (that's something that needs to be added to the PR)

Please improve description & title to be more useful: the real thing that's added here is a readme entry & an example for using Conan. Right now both description & title is very misleading which is a problem as they show up in our changelog (and as a reviewer I had to guess what's going on)

Overall I think there's still quite a few improvements to be made to make this a useful example/readme-entry for new-comers to the project and to make it something we can maintain going forward.
The readme section in particular looks very generic right now and imho isn't helpful enough (I have no issue using AI for these things, but the result gotta be right :))

flightwave-maxwell:
Apologies, this was cloned from a forked repo and the upstream ended up being the original repo and not the forked repo. Thanks for taking time to go through the PR, it was helpful.

PR #10810, opened 2025-08-06
Fix flaky batcher-config test

emilk:
### Related
* Introduced in https://github.com/rerun-io/rerun/pull/10620

### What
Environment variables are annoying

### DIFF
--- a/crates/top/re_sdk/src/recording_stream.rs
+++ b/crates/top/re_sdk/src/recording_stream.rs
@@ -3126,8 +3126,22 @@ mod tests {
 
     const CONFIG_CHANGE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
 
+    fn clear_environment() {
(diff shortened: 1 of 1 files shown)


crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -3143,14 +3157,18 @@ mod tests {
...
+            "Buffered sink should uses the config from the environment"
+        );
 
         // Change sink to our custom sink. Will it take over the setting?
         let injected_config = ChunkBatcherConfig {
             flush_tick: std::time::Duration::from_secs(123),
             flush_num_bytes: 123,
             flush_num_rows: 123,
-            ..ChunkBatcherConfig::DEFAULT
+            ..new_config
```
Wumpf, line 3171: semantically wrong imho, this really should have everything else on default (but the assert above pretty much guarantees that this is the case)

PR #10811, opened 2025-08-06
Make pb-breaking run against `latest` branch rather than `main`

Wumpf:
Fix needed for patch releases


Wumpf:
the idea is that renaming is a wire compatible change and therefore we ignore it manually and then have the next push be successful. See https://rerunio.slack.com/archives/C033K5VS2KD/p1754496759897239

need to do something else to deal with patch releases 🤔

PR #10813, opened 2025-08-06
Upgrade Zombies Script: Don't Query All Issues/ Check External Issues

### LINKED ISSUE #10798: Remove Cmake 4.0 Workarounds
Arrow has issues with Cmake >= 4.0 https://github.com/apache/arrow/issues/45985

This is a mirroring issue to link against for zombie TODOs. Once arrow's issue is resolved we can close this. Alternatively we could update our zombie todo script to scan external issue boards but that feels like a rabbithole.

### LINKED ISSUE #10801: Update Zombie_todos to check against external repos
Right now we only check TODOs against internal issues but it would be nice if we checked externally as well. See #10798

ntjohnson1:
### Related
Closes #10801 and closes #10798 by making it unnecessary with zombie support

### What
This is basically just a full refactor of the zombie script. I asked Claude to add external repo support then I deduped code, removed the need to scan all issues on all repos, progress bar etc. This could probably still be cleaned up a little more if someone felt strongly but we had globals littered around before.

1. Go through all files
2. Try to match every line to internal or external issue format
3. Check if that issue is closed by directly querying github for it (cache results)
4. Print everything out at the end

The one annoying thing I couldn't resolve because I think it is a github limitation is being able to check private repos under the rerun org with a fine grained token. A classic token with repo scope allows you to check all private repos you have access to but a fine grained doesn't unless generated under the organization itself which requires an admin to approve. Note that this should work without a token at all but with a lower rate limit/no private repo access.

Before
```
...snip...
done fetching issues
./crates/viewer/re_renderer/src/device_caps.rs:478: wgpu::Backends::GL // TODO(#8559): Fix WebGPU on Safari
  File "/Users/nick/repos/rerun/scripts/zombie_todos.py", line 258, in <module>
    main()
  File "/Users/nick/repos/rerun/scripts/zombie_todos.py", line 254, in main
    raise ValueError("Clean your zombies!")
ValueError: Clean your zombies!

real    1m43.139s
user    0m9.786s
sys     0m6.694s
```

After
```
time pixi run python scripts/zombie_todos.py  --token <token>
Processing 8499 files...
Using 16 worker threads...
Processing files: 100%|███████████████████████████████████████████████████████████████████████████████████████████████| 8499/8499 [00:21<00:00, 397.97file/s]
./crates/viewer/re_ui/src/alert.rs:65: .outer_margin(1.0) // Needed because we set clip_rect_margin. TODO(emilk/egui#4019): remove clip_rect_margin
./crates/viewer/re_renderer/src/device_caps.rs:478: wgpu::Backends::GL // TODO(#8559): Fix WebGPU on Safari
./crates/viewer/re_renderer/src/view_builder.rs:617: // TODO(gfx-rs/wgpu#1453): Note that this is a limitation that will be lifted in future versions of wgpu.
./crates/store/re_types/src/testing/datatypes/affix_fuzzer21_ext.rs:3: // TODO(apache/arrow-rs#7411)
./rerun_py/src/dataframe/recording.rs:36: // TODO(rerun-io/dataplatform#405): interface with remote data needs to be reimplemented
./examples/python/open_photogrammetry_format/open_photogrammetry_format.py:94: # TODO(Pix4D/pyopf#6): https://github.com/Pix4D/pyopf/issues/6


Cache statistics:
  Total unique issues checked: 208
  Closed issues found: 6
  Open/unknown issues: 202
  Cache hits: 321, Cache misses: 266
  File "/Users/nick/repos/rerun/scripts/zombie_todos.py", line 338, in <module>
    main()
  File "/Users/nick/repos/rerun/scripts/zombie_todos.py", line 334, in main
    raise ValueError("Clean your zombies!")
ValueError: Clean your zombies!

real    0m29.186s
user    0m5.713s
sys     0m4.067s
```

### DIFF
--- a/scripts/zombie_todos.py
+++ b/scripts/zombie_todos.py
@@ -69,7 +69,7 @@ def check_issue_closed(repo_owner: str, repo_name: str, issue_number: int) -> tu
     global issue_cache, cache_hit, cache_miss
     cache_key = (repo_owner, repo_name, issue_number)
 
-    # Check if we already have this result cached (thread-safe)
+    # Check if we already have this result cached
     with cache_lock:
         if cache_key in issue_cache:
             cache_hit += 1
@@ -78,7 +78,7 @@ def check_issue_closed(repo_owner: str, repo_name: str, issue_number: int) -> tu
     # Fetch the result and cache it
     result = check_issue_closed_api(repo_owner, repo_name, issue_number)
 
-    # Store in cache (thread-safe)
+    # Store in cache
     with cache_lock:
         issue_cache[cache_key] = result
 
@@ -277,6 +277,7 @@ def main() -> None:
 
     # Collect all files to process
     files_to_process = []
+    print("Collecting files to process...")
     for root, dirs, files in os.walk(".", topdown=True):
         dirs[:] = [d for d in dirs if not should_ignore(d)]
 


scripts/zombie_todos.py
```diff
@@ -2,109 +2,103 @@
...
-        # Extract issue numbers and authors
-        for issue in data:
-            issues.append(issue["number"])
-            authors[issue["number"]] = issue["user"]["login"]
+def check_issue_closed(repo_owner: str, repo_name: str, issue_number: int) -> tuple[bool, str]:
+    """
+    Check if a specific issue is closed and get its author.
 
-        print(f"done fetching issues - total: {len(issues)}")
+    Uses caching to avoid repeated API calls for the same issue.
```
Wumpf, line 63: nice, as our issue list grows that's clearly the better approach!

scripts/zombie_todos.py
```diff
@@ -248,7 +287,49 @@ def main() -> None:
...
+            for future in as_completed(future_to_file):
+                filepath = future_to_file[future]
+                try:
+                    file_ok, file_display = future.result()
+                    ok &= file_ok
+                    completed_files += 1
+                    display += file_display
+                    pbar.update(1)
+                    # if completed_files % 10 == 0:  # Progress update every 10 files
+                    #     print(f"  Processed {completed_files}/{len(files_to_process)} files...")
```
Wumpf: dead code?

scripts/zombie_todos.py
```diff
@@ -237,7 +275,8 @@ def main() -> None:
 
     should_ignore = parse_gitignore(".gitignore")  # TODO(emilk): parse all .gitignore files, not just top-level
 
-    ok = True
+    # Collect all files to process
+    files_to_process = []
```
Wumpf, line 279: printing something while/before doing that would be nice. Right now when the script starts nothing happens for a while until I see the progress bar

scripts/zombie_todos.py
```diff
@@ -2,109 +2,103 @@
...
-    except json.JSONDecodeError as e:
-        print(f"Error: Failed to parse JSON response: {e}")
+    Returns:
+        (is_closed, author) tuple
+
+    """
+    global issue_cache, cache_hit, cache_miss
+    cache_key = (repo_owner, repo_name, issue_number)
+
+    # Check if we already have this result cached (thread-safe)
```
Wumpf: finding these (probably generated?) "(thread-safe)" comments rather amusing. I mean it's just "thread safe" because python code itself doesn't run in parallel, only the actual requests do

scripts/zombie_todos.py
ntjohnson1: Yes, as our TODO list grows too long we'll probably need to look if we can group the queries instead of issuing an API request once per unique issue to avoid API limits. But that's a future optimization.

scripts/zombie_todos.py
ntjohnson1: Yes generated and will remove since the lock should make that clear. IIUC when one "thread" is making an IO call the other can begin working. I think technically without the thread lock you do have race conditions. So while not parallel the partial concurrency needs the lock.

scripts/zombie_todos.py
ntjohnson1: Hmm resolving that todo above to respect all .gitignore files would probably help here. Collecting all the files shouldn't take that long. In the short term I just added a print out that we are collecting files.

PR #10820, opened 2025-08-07
Crate publishing: after publishing a crate, wait for it to be available

emilk:
I had this problem:
image

So `re_build_info` was successfully published, but when trying to publish something that was using it, we still failed.
I had a similar problem earlier, but with `re_log`

So I added a loop to wait for the crate to become "visible". I hope that will help.

### DIFF
--- a/scripts/ci/crates.py
+++ b/scripts/ci/crates.py
@@ -444,7 +444,21 @@ def publish_crate(crate: Crate, token: str, version: str, env: dict[str, Any]) -
                 dry_run=False,
                 capture=True,
             )
+
+            if not is_already_published(version, crate):
+                # Theoretically this shouldn't be needed… but sometimes it is.
+                print(f"{R}Waiting for {name} to become available…")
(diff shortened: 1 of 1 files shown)


Wumpf (approved):
has a curious relationship with the outer retry loop which afaik was introduces because of the same problem. Bit confusing to read now with both nested, but the approach makes sense and should be more robust than what we have 🤞

Wumpf (approved):
has a curious relationship with the outer retry loop which afaik was introduces because of the same problem. Bit confusing to read now with both nested, but the approach makes sense and should be more robust than what we have 🤞

emilk:
I'll merge this once we've made the next release, so that we actually run the code before we merge it

EDIT: is being tested in https://github.com/rerun-io/rerun/actions/runs/16800914352
EDIT 2: we didn't need the code in the above run, but at least it didn't break anything

Maybe this build will hit it: https://github.com/rerun-io/rerun/actions/runs/16806796956/job/47601393044

EDIT 3: neither build hit the new path. Oh well.

PR #10825, opened 2025-08-07
Improve command palette in light mode

emilk:
### Before
This looked really bad in light mode
comand-palette-before

### After
comand-palette-after

image


lucasmerlin (approved):
Nice!

PR #10829, opened 2025-08-07
Implement row filtering on client side

timsaucer:
### Related

Closes https://github.com/rerun-io/dataplatform/issues/851

### What

This PR identifies when the push down filters provided to the table provider are checking for a component that is not null. If so it pushes this filter down into the chunk store query. This reduces the amount of data coming out of the cpu worker thread and passed into the rest of the datafusion engine.

These are equivalent

- column is not null
- not (column is null)
- column != lit(null)

### DIFF
--- a/crates/store/re_datafusion/src/dataframe_query_common.rs
+++ b/crates/store/re_datafusion/src/dataframe_query_common.rs
@@ -6,9 +6,9 @@ use arrow::array::{
 use arrow::datatypes::{DataType, Field, Int64Type, Schema, SchemaRef, TimeUnit};
 use async_trait::async_trait;
 use datafusion::catalog::{Session, TableProvider};
-use datafusion::common::{DataFusionError, downcast_value, exec_datafusion_err};
+use datafusion::common::{Column, DataFusionError, downcast_value, exec_datafusion_err};
 use datafusion::datasource::TableType;
-use datafusion::logical_expr::Expr;
+use datafusion::logical_expr::{Expr, Operator, TableProviderFilterPushDown};
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::coalesce_batches::CoalesceBatchesExec;
 use re_dataframe::external::re_chunk::ChunkId;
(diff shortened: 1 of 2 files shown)


crates/store/re_datafusion/src/dataframe_query_common.rs
```diff
@@ -190,6 +191,49 @@ impl DataframeQueryTableProvider {
             chunk_request,
         })
     }
+
+    fn column_to_selector(column: &Column) -> Option<ComponentColumnSelector> {
```
emilk: Nit/style: we usually prefer `selector_from_column` (see [`CODE_STYLE.md`](https://github.com/rerun-io/rerun/blob/main/CODE_STYLE.md#matrices)).

Mostly for consistency with matrices, but also writing it `let selector = selector_from_column(column);` means the names in the function title is next to the types of the input/output variables (instead of being swizzled)

crates/store/re_datafusion/src/dataframe_query_common.rs
```diff
@@ -190,6 +191,49 @@ impl DataframeQueryTableProvider {
...
+        ComponentColumnSelector::from_str(column.name()).ok()
+    }
+
+    fn compute_column_is_not_null_filter(
```
emilk: This could use a docstring to explain that what this checks for, especially since the function name can be read two ways (does it return true for the filter `foo == 42` since that is not a null filter?)

emilk: Maybe this would read slightly better?

```suggestion
    fn compute_column_is_neq_null_filter(
```

crates/store/re_datafusion/src/dataframe_query_common.rs
```diff
@@ -190,6 +191,49 @@ impl DataframeQueryTableProvider {
...
+                                if sv.is_null() {
+                                    return Ok(Self::column_to_selector(col));
+                                }
+                            }
+                        }
+                    }
+                    _ => {}
+                }
+
+                Ok(None)
```
emilk: Nit: this would read nicer with a helper function

```suggestion
                if is_neq_null(expr) {
                    Ok(Self::column_to_selector(col))
                } else {
                    Ok(None)
                }
```

That would also allow us to unit-test `is_neq_null`

crates/store/re_datafusion/src/dataframe_query_common.rs
```diff
@@ -190,6 +191,49 @@ impl DataframeQueryTableProvider {
...
+            .map(|expr| {
+                match expr {
+                    Expr::IsNotNull(inner) => {
+                        if let Expr::Column(col) = inner.as_ref() {
+                            return Ok(Self::column_to_selector(col));
+                        }
+                    }
+                    Expr::Not(inner) => {
+                        if let Expr::IsNull(col_expr) = inner.as_ref() {
+                            if let Expr::Column(col) = col_expr.as_ref() {
```
emilk: can use let-chaining

crates/store/re_datafusion/src/dataframe_query_common.rs
```diff
@@ -190,6 +191,49 @@ impl DataframeQueryTableProvider {
...
+                            }
+                        }
+                    }
+                    Expr::BinaryExpr(binary) => {
+                        if binary.op == Operator::NotEq {
+                            if let (Expr::Column(col), Expr::Literal(sv))
+                            | (Expr::Literal(sv), Expr::Column(col)) =
+                                (binary.left.as_ref(), binary.right.as_ref())
+                            {
+                                if sv.is_null() {
```
emilk: can use let-chaining

emilk (approved):
Can't say I follow everything happening here, but… looks good :)

Is there a way we can add a test for this without too much hassle?

PR #10830, opened 2025-08-07
Clean up time range definitions a bit

Wumpf:
* `ResolvedTimeRange` -> `AbsoluteTimeRange` + docs
* `re_uri::TimeRange` -> `re_uri::TimeSelection`, using `AbsoluteTimeRange`

### DIFF
--- a/crates/store/re_chunk/src/batcher.rs
+++ b/crates/store/re_chunk/src/batcher.rs
@@ -11,7 +11,7 @@ use nohash_hasher::IntMap;
 
 use re_arrow_util::arrays_to_list_array_opt;
 use re_byte_size::SizeBytes as _;
-use re_log_types::{EntityPath, AbsoluteTimeRange, TimeInt, TimePoint, Timeline, TimelineName};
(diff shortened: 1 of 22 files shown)


crates/utils/re_uri/src/uri_time_range.rs
```diff
@@ -1,43 +1,40 @@
-use re_log_types::{NonMinI64, ResolvedTimeRangeF, TimeCell};
+use re_log_types::{AbsoluteTimeRange, AbsoluteTimeRangeF, TimeCell};
 
 use crate::Error;
 
+/// A time range as used in URIs, qualified with a timeline.
 #[derive(Debug, Clone, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
-pub struct TimeRange {
+pub struct UriTimeRange {
```
jleibs: We need a better word for Timeline + TimeRange?

Something like TimeSelection maybe?

jleibs (approved):

crates/utils/re_uri/src/uri_time_range.rs
Wumpf: `TimeSelection` sounds nice, let's go with that

PR #10832, opened 2025-08-07
Update MSRV to 1.88

emilk:
This let's us start using `let` chains :)

Best reviewed commit-by-commit

### DIFF
--- a/crates/store/re_grpc_client/src/redap/mod.rs
+++ b/crates/store/re_grpc_client/src/redap/mod.rs
@@ -283,7 +283,8 @@ pub fn get_chunks_response_to_chunk_and_partition_id(
         let resp = resp?;
 
         let _span =
-            tracing::trace_span!("get_chunks::batch_decode", num_chunks = resp.chunks.len()).entered();
+            tracing::trace_span!("get_chunks::batch_decode", num_chunks = resp.chunks.len())
+                .entered();
 
         resp.chunks
             .into_iter()


crates/build/re_types_builder/src/codegen/rust/api.rs
```diff
@@ -1210,17 +1210,17 @@ fn quote_trait_impls_for_archetype(reporter: &Reporter, obj: &Object) -> TokenSt
             #(#all_descriptor_methods)*
         }
 
-        static REQUIRED_COMPONENTS: once_cell::sync::Lazy<[ComponentDescriptor; #num_required_descriptors]> =
-            once_cell::sync::Lazy::new(|| {[#required_descriptors]});
+        static REQUIRED_COMPONENTS: std::sync::LazyLock<[ComponentDescriptor; #num_required_descriptors]> =
```
Wumpf, line 1213: oh, nice. For what do we still need `once_cell` now?

crates/build/re_types_builder/src/codegen/rust/util.rs
```diff
@@ -337,15 +337,13 @@ pub fn doc_as_lines(
...
-                    );
-                    continue;
-                }
+            if example.lines.iter().any(|line| line.contains("```")) {
+                reporter.error(
+                    virtpath,
+                    fqname,
+                    format!("Example {path:?} contains ``` in it, so we can't embed it in the Rust API docs."),
+                );
+                continue;
```
Wumpf, line 346: I think this is wrong: before we'd report all errors and then proceed. The `continue` was just superfluous though.
Now we report one error and if we find one we exit the outer loop. 

the previous behavior was preferable

crates/store/re_chunk_store/src/subscribers.rs
```diff
@@ -194,13 +191,12 @@ impl ChunkStore {
...
-                .as_any()
-                .downcast_ref::<PerStoreStoreSubscriberWrapper<S>>()
-                .and_then(|wrapper| wrapper.get(store_id).map(&mut f))
-        })
+        let subscriber = subscribers.get(handle as usize)?;
+        let subscriber = subscriber.read();
+        subscriber
+            .as_any()
+            .downcast_ref::<PerStoreStoreSubscriberWrapper<S>>()
+            .and_then(|wrapper| wrapper.get(store_id).map(&mut f))
```
Wumpf: curious, why is `?` not suggested here over `and_then`?

crates/store/re_log_encoding/src/stream_rrd_from_http.rs
```diff
@@ -205,9 +209,11 @@ pub mod web_decode {
                 }
             }
             Err(err) => {
-                on_msg(HttpMessage::Failure(
+                if on_msg(HttpMessage::Failure(
```
Wumpf: why is this?

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -219,7 +219,7 @@ impl DensityGraph {
             max: max_y,
         } = y_range;
 
-        let center_y = (min_y + max_y) / 2.0;
+        let center_y = f32::midpoint(min_y, max_y);
```
Wumpf: TIL https://doc.rust-lang.org/src/core/num/f32.rs.html#992

data density graph is perf sensitive though, this looks like it would make it worse though? For all platforms we care about its the same but with f64 casts :/
I don't think we care about the added precision here

crates/viewer/re_view/src/results_ext.rs
```diff
@@ -111,7 +111,7 @@ pub enum HybridResults<'a> {
     LatestAt(LatestAtQuery, HybridLatestAtResults<'a>),
 
     // Boxed because of size difference between variants
-    Range(RangeQuery, Box<HybridRangeResults<'a>>),
+    Range(RangeQuery, HybridRangeResults<'a>),
```
Wumpf, line 112: outdated comment

I see you're trying to save karma :D

Wumpf (approved):
this `clippy::return_and_then` is quite impactful, huh. I like how it reduces right shift in some places!

What's the difference now between `LazyLock` and `once_cell::sync::Lazy`? Doesn't look like this has changed recently? Couldn't find anything about it with a quick search

looks all great except for that issue on codegen I found.
And maybe that midpoint thing - can't see which lint causes this, but I'm not convinced yet it's universally a good idea!

crates/build/re_types_builder/src/codegen/rust/util.rs
emilk: Previously it would report multiple errors for each example, which was unnecessary. If there is one ```, there is almost certainly another, and we don't need to report all of them

Wumpf: ah I see, makes sense then

crates/build/re_types_builder/src/codegen/rust/api.rs
emilk: Apparently only `once_cell::sync::OnceCell` which I now have replaced with `std::sync::OnceLock`

emilk:
> What's the difference now between `LazyLock` and `once_cell::sync::Lazy`? Doesn't look like this has changed recently? Couldn't find anything about it with a quick search

`LazyLock` is the std replacement for `once_cell::sync::Lazy`, introduced in 1.80

crates/viewer/re_time_panel/src/data_density_graph.rs
emilk: Good catch… yeah, I think I'll revert some/most of these

emilk: https://rust-lang.github.io/rust-clippy/master/index.html#/midpoint is the lint btw. For integers it's very good to have in case one is writing a binary search or similar. For floats… less so.

emilk: * Also: https://github.com/emilk/egui/pull/7435

PR #10833, opened 2025-08-07
Fix Mac CI By Updating Default Env

ntjohnson1:
### Related
Broke CI because the task in default depended on cpp env.

### What
Adds llvm tools which includes ar into the default just for mac. Only 20 MB so not too bad in size.


emilk:
@rerun-bot full-check

ntjohnson1:
Bug fix related to https://github.com/rerun-io/rerun/issues/9849

PR #10845, opened 2025-08-08
Shrink many of our error types

emilk:
Large errors leads to large `Result`s, which may lead to a lot of `memcpy`:s when returning a `Result` (even an `Ok` one), which can hurt performance.

Some of our error types were over 100 bytes, even 200 bytes.

Many of our errors are still pretty huge (64+ bytes), but this is a start.

### DIFF
--- a/crates/store/re_chunk/src/chunk.rs
+++ b/crates/store/re_chunk/src/chunk.rs
@@ -62,6 +62,15 @@ pub enum ChunkError {
     InvalidSorbetSchema(#[from] re_sorbet::SorbetError),
 }
 
+#[test]
+fn test_error_size() {
+    assert!(
+        std::mem::size_of::<ChunkError>() <= 72,
+        "Size of error is {} bytes. Let's try to keep errors small.",
+        std::mem::size_of::<ChunkError>()
+    );
+}
+
 pub type ChunkResult<T> = Result<T, ChunkError>;
 
 // ---
--- a/crates/store/re_entity_db/src/instance_path.rs
+++ b/crates/store/re_entity_db/src/instance_path.rs
@@ -105,7 +105,7 @@ impl FromStr for InstancePath {
 
         if let Some(component_descriptor) = component_descriptor {
             return Err(PathParseError::UnexpectedComponentDescriptor(
-                component_descriptor,
+                component_descriptor.into(),
             ));
         }
 
--- a/crates/store/re_grpc_client/src/lib.rs
+++ b/crates/store/re_grpc_client/src/lib.rs
@@ -19,7 +19,29 @@ const MAX_DECODING_MESSAGE_SIZE: usize = u32::MAX as usize;
 
 /// Wrapper with a nicer error message
 #[derive(Debug)]
-pub struct TonicStatusError(pub tonic::Status);
+pub struct TonicStatusError(Box<tonic::Status>);
+
+#[test]
+fn test_error_size() {
+    assert!(
+        std::mem::size_of::<TonicStatusError>() <= 32,
+        "Size of error is {} bytes. Let's try to keep errors small.",
+        std::mem::size_of::<TonicStatusError>()
+    );
+}
+
+impl AsRef<tonic::Status> for TonicStatusError {
+    fn as_ref(&self) -> &tonic::Status {
+        &self.0
+    }
+}
+
+impl TonicStatusError {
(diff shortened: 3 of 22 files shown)


crates/store/re_chunk/src/chunk.rs
```diff
@@ -62,6 +62,15 @@ pub enum ChunkError {
...
 }
 
+#[test]
+fn test_error_size() {
+    assert!(
+        std::mem::size_of::<ChunkError>() <= 72,
+        "Size of error is {} bytes. Let's try to keep errors small.",
+        std::mem::size_of::<ChunkError>()
+    );
+}
```
jprochazk, line 72: All of these can be `const` assertions instead, which will fail these earlier, at compile time.

```suggestion
const _: () = assert!(
    std::mem::size_of::<ChunkError>() <= 72,
    "Error type is too large. Try to reduce its size by boxing some of its variants.",
);
```

Playground link: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=9c0bead836c6b305b767f4e6f9ed4b6b

emilk: My test prints the size of the struct, which is very useful. This can't be done statically afaict.

Also: they don't need to fail on compile-time. This is a performance problem, but you should be able to test and run your new code without having to care about that.

jprochazk: > My test prints the size of the struct, which is very useful. This can't be done statically afaict.

No, [this is possible](https://docs.rs/const_format/latest/const_format/macro.assertcp.html).

But bringing in a dependency for that seems overkill. Why do you need it to print the size of the type? R-A will report the failure at the assertion site, which is in the same file as the type. You can hover over it in your IDE when the assertion fails at compile time.

> Also: they don't need to fail on compile-time. This is a performance problem, but you should be able to test and run your new code without having to care about that.

It's a performance problem, but one that you are now forcing everyone to be aware of and always solve before merging a PR. The sooner they find out about it, the better.

Scenario: You add an error variant, the size now exceeds some arbitrary threshold. You don't notice this. Then you write some code using that error variant, constructing it in different places, mapping to it from different errors, etc. Then you run your tests-- only NOW you find out about the issue. So you have to go back and fix not only the error type, but also every place you construct it (probably inserting a bunch of `Box::new`).

Hopefully you've only added one error variant, otherwise you have to also figure out which of the variants you just added is causing the error to go over the size limit.

emilk: Fair points, but the added code is quite minimal (just adding `Box` and a new `From` implementation), and yes - printing the size is very useful. You wouldn't want to tell clement to just hover the type in vim (especially since it doesn't even work reliably in VSCode)

jprochazk: > Fair points, but the added code is quite minimal (just adding Box and a new From implementation)

I think you've underestimating how annoying it is for this to be a test. Two reviewers on this PR asked you to change it... We have precedent for types being checked against some constraints at compile-time, or code not compiling if you're not handling all struct fields/enum variants somewhere. With tests, at least personally I'll only find out about this issue after CI runs on my PR, adding another thing to do just to "make CI green".

> You wouldn't want to tell clement to just hover the type in vim

I would, because he uses rust analyzer, like everyone else in this company. Hover information is not a VSCode specific thing.

> (especially since it doesn't even work reliably in VSCode)

It is reliable. The only time it doesn't show the size is if the type does not have a size yet. None of these errors have type parameters, so they show the size just fine:

image

Here's what a generic type looks like:
image

And if you fill in the type:
image

jprochazk: We do already depend on `const_format` in the repository, so using it for this wouldn't be adding a new dependency.

Wumpf: I agree with Jan! rustanalyzer struct size reports have been reliable for me as well and anything that puts a failure from compile time to test time is a potential productivity loss

emilk: ok

emilk: * https://github.com/rerun-io/rerun/pull/10925

Wumpf: thank you <3

PR #10846, opened 2025-08-08
Use nightly clippy to automatically convert some code to use let-chains

emilk:


Wumpf (approved):
so beautiful! T_T

timsaucer (approved):
very nice!

PR #10847, opened 2025-08-08
Small nightly clippy fixes

emilk:


Wumpf (approved):
you just can't get enough? 😄

PR #10851, opened 2025-08-11
Fix Mac CI running out of disk space

emilk:
We got "ERROR: IO failure on output stream: No space left on device" here: https://github.com/rerun-io/rerun/pull/10847

I tried switching to a large runner, but it still runs out (apparently the max is 14 GB: https://docs.github.com/en/actions/reference/runners/larger-runners).

But, I discovered we unnecessarily built the web viewer, which takes 15 minutes and lord knows how much disk space.
However, in order to accomplish that I had to refactor how we disable the web viewer.

### DIFF
--- a/crates/viewer/re_view_spatial/tests/draw_order.rs
+++ b/crates/viewer/re_view_spatial/tests/draw_order.rs
@@ -1,6 +1,9 @@
 use re_chunk_store::RowId;
 use re_log_types::TimePoint;
-use re_test_context::TestContext;
+use re_test_context::{
+    TestContext,
+    external::egui_kittest::{OsThreshold, SnapshotOptions},
+};
 use re_test_viewport::TestContextExt as _;
 use re_viewer_context::{ViewClass as _, ViewId};
 use re_viewport_blueprint::ViewBlueprint;
@@ -167,5 +170,8 @@ fn run_view_ui_and_save_snapshot(
         });
 
     harness.run();
-    harness.snapshot(name);
+    harness.snapshot_options(
+        name,
+        &SnapshotOptions::new().failed_pixel_count_threshold(OsThreshold::new(0).macos(40)),
+    );
 }


emilk:
@rerun-bot full-check

Wumpf (approved):
👍

emilk:
Nope, that was not enough. Apparently the large Mac runners still max out at 14 GB: https://docs.github.com/en/actions/reference/runners/larger-runners

I'll try cut out the web build instead

emilk:
@rerun-bot full-check

emilk:
```
  Tracy Profiler initialization failure: CPU doesn't support invariant TSC.
  Define TRACY_NO_INVARIANT_CHECK=1 to ignore this error, *if you know what you are doing*.
  Alternatively you may rebuild the application with the TRACY_TIMER_FALLBACK define to use lower resolution timer.
```

not sure if this is causing the failure… but I'll look into it

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

pixi.toml
```diff
@@ -240,9 +240,7 @@ rerun-web-release = { cmd = "cargo run --package rerun-cli --no-default-features
 # (this looks heavy but takes typically below 0.1s!)
 rerun-build-web-release = "rustup target add wasm32-unknown-unknown && cargo run --quiet -p re_dev_tools -- build-web-viewer --no-default-features --features analytics,map_view --release"
 
-rs-check = { cmd = "rustup target add wasm32-unknown-unknown && python scripts/ci/rust_checks.py", depends-on = [
-  "rerun-build-web", # The checks require the web viewer wasm to be around.
```
emilk, line 244: Yay!

scripts/ci/rust_checks.py
```diff
@@ -69,6 +69,17 @@ def run_cargo(
 
     additional_env_vars["RUSTFLAGS"] = f"{extra_cfgs} {'--deny warnings' if deny_warnings else ''}"
     additional_env_vars["RUSTDOCFLAGS"] = f"{extra_cfgs} {'--deny warnings' if deny_warnings else ''}"
+
+    # We shouldn't require the web viewer .wasm to exist before running clippy, unit tests, etc:
+    additional_env_vars["RERUN_DISABLE_WEB_VIEWER_SERVER"] = "1"
+
+    # Explicitly TRACY to avoid macOS failure on CI, that looks like this:
```
emilk:
```suggestion
    # Disable TRACY to avoid macOS failure on CI, that looks like this:
```

emilk:
@rerun-bot full-check

crates/viewer/re_view_spatial/tests/draw_order.rs
```diff
@@ -167,5 +170,8 @@ fn run_view_ui_and_save_snapshot(
         });
 
     harness.run();
-    harness.snapshot(name);
+    harness.snapshot_options(
+        name,
+        &SnapshotOptions::new().failed_pixel_count_threshold(OsThreshold::new(0).macos(40)),
```
emilk, line 175: I have no idea why I had to adjust this now 🤷

Wumpf: huh, strange indeed. Curious, did it fail on ci or locally or both?

pixi.toml
Wumpf: nice, this was soooo annoying before

crates/viewer/re_view_spatial/tests/draw_order.rs
emilk: It failed on CI. I'm unsure if I tried locally.

PR #10852, opened 2025-08-11
Introduce ChunkKey and FetchChunks API

zehiko:
### What

This PR introduces a ``ChunkKey`` that uniquely identifies a chunk location. It is to be used in a 2-phase query path in which result of the 1st phase is a stream of ``ChunkKey``s that a caller can use in a 2nd phase to actually fetch the chunk data. 
We also introduce ``FetchChunks`` API that uses ``ChunkKey`` and will soon enable 2-phase query. 

There are some important design considerations (that are up for discussion and changes):
* ``ChunkKey`` is (as defined in the design doc) opaque and can be decoded by an appropriate driver/mapper. In order to support it, I've for opted to use ``google.protobuf.Any`` and follow basically the same pattern we've used for ``ProviderDetails`` in the catalog service. I went back and forth on this one, considering ``oneof`` as well, but since there will be a lot of changes and different drivers introduced very soon, I thought we might appreciate the flexibility of it. It obviously comes with downsides (no type safety, runtime errors, small performance overhead) 
* ``ChunkKey`` includes location (storage) url and (opaque) location details.  ~~Open question: should it include ``ChunkId``, will we have a ``ChunkId`` always available for non-rrd chunks? Including it right now would probably temporarily simplify integration with existing local chunk caching support on the Redap side.~~ Yes, we're including ``ChunkId`` in the key.
* ``FetchChunksRequest`` includes partition ids and layers. I wanted to keep these separate from the ``ChunkKey`` as I believe they are current implementation detail where we need to move the metadata back and forth. This is also one of the reasons I didn't use ``DataSource`` in the ``ChunkKey`` as I was thinking that partition layer or kind should not be part of it (but happy to discuss alternative)

### DIFF
--- a/crates/store/re_protos/proto/rerun/v1alpha1/common.proto
+++ b/crates/store/re_protos/proto/rerun/v1alpha1/common.proto
@@ -226,16 +226,16 @@ message TaskId {
 
 // `ChunkKey` provides chunk location details in the data store.
 message ChunkKey {
-  // where is the chunk stored (S3, local file, etc)
-  optional string location_url = 2;
+  // Chunk unique identifier
+  rerun.common.v1alpha1.Tuid chunk_id = 1;
 
-  // additional details about the chunk's location that are specific to the underlying data source (partition type)
-  google.protobuf.Any location_details = 3;
+  // details about the chunk's location that are specific to the underlying data source (partition type)
+  google.protobuf.Any location = 3;
 }
 
 message RrdChunkLocationDetails {
-  // Chunk unique identifier
-  rerun.common.v1alpha1.Tuid chunk_id = 1;
+  // where is the chunk stored (S3, local file, etc)
+  optional string url = 1;
 
   // byte offset of the chunk within the data source
   optional uint64 offset = 2;
--- a/crates/store/re_protos/src/v1alpha1/rerun.common.v1alpha1.ext.rs
+++ b/crates/store/re_protos/src/v1alpha1/rerun.common.v1alpha1.ext.rs
@@ -658,8 +658,8 @@ impl TryFrom<crate::common::v1alpha1::ComponentDescriptor> for ComponentDescript
 
 #[derive(Debug, Clone)]
 pub struct ChunkKey {
-    pub location_url: url::Url,
-    pub location_details: prost_types::Any,
+    pub chunk_id: re_chunk::ChunkId,
+    pub location: prost_types::Any,
 }
 
 impl ChunkKey {
@@ -675,23 +675,19 @@ impl TryFrom<crate::common::v1alpha1::ChunkKey> for ChunkKey {
     type Error = TypeConversionError;
 
     fn try_from(value: crate::common::v1alpha1::ChunkKey) -> Result<Self, Self::Error> {
-        let location_url = value
-            .location_url
-            .ok_or(missing_field!(
-                crate::common::v1alpha1::ChunkKey,
-                "location_url"
-            ))?
-            .parse()?;
+        let tuid = value.chunk_id.ok_or(missing_field!(
+            crate::common::v1alpha1::RrdChunkLocationDetails,
+            "chunk_id"
+        ))?;
+        let id: re_tuid::Tuid = tuid.try_into()?;
+        let chunk_id = re_chunk::ChunkId::from_u128(id.as_u128());
 
-        let location_details = value.location_details.ok_or(missing_field!(
+        let location = value.location.ok_or(missing_field!(
             crate::common::v1alpha1::ChunkKey,
-            "location_details"
+            "location"
         ))?;
 
-        Ok(Self {
-            location_url,
-            location_details,
-        })
+        Ok(Self { chunk_id, location })
     }
 }
 
@@ -710,16 +706,18 @@ impl TryFrom<&[u8]> for ChunkKey {
 
 impl From<ChunkKey> for crate::common::v1alpha1::ChunkKey {
     fn from(value: ChunkKey) -> Self {
+        let tuid =
+            crate::common::v1alpha1::Tuid::from(re_tuid::Tuid::from_u128(value.chunk_id.as_u128()));
         Self {
-            location_url: Some(value.location_url.to_string()),
-            location_details: Some(value.location_details),
+            chunk_id: Some(tuid),
+            location: Some(value.location),
         }
     }
 }
 
 #[derive(Debug, Clone)]
 pub struct RrdChunkLocationDetails {
-    pub chunk_id: re_chunk::ChunkId,
+    pub url: url::Url,
     pub offset: u64,
     pub length: u64,
 }
@@ -730,12 +728,20 @@ impl TryFrom<crate::common::v1alpha1::RrdChunkLocationDetails> for RrdChunkLocat
     fn try_from(
         value: crate::common::v1alpha1::RrdChunkLocationDetails,
     ) -> Result<Self, Self::Error> {
-        let tuid = value.chunk_id.ok_or(missing_field!(
-            crate::common::v1alpha1::RrdChunkLocationDetails,
-            "chunk_id"
-        ))?;
-        let id: re_tuid::Tuid = tuid.try_into()?;
-        let chunk_id = re_chunk::ChunkId::from_u128(id.as_u128());
+        let url = value
+            .url
+            .ok_or(missing_field!(
+                crate::common::v1alpha1::RrdChunkLocationDetails,
+                "location_url"
+            ))?
+            .parse()
+            .map_err(|err| {
+                invalid_field!(
+                    crate::common::v1alpha1::RrdChunkLocationDetails,
+                    "location_url",
+                    err
+                )
+            })?;
 
         let offset = value.offset.ok_or(missing_field!(
             crate::common::v1alpha1::RrdChunkLocationDetails,
@@ -748,7 +754,7 @@ impl TryFrom<crate::common::v1alpha1::RrdChunkLocationDetails> for RrdChunkLocat
         ))?;
 
         Ok(RrdChunkLocationDetails {
-            chunk_id,
+            url,
             offset,
             length,
         })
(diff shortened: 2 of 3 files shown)


crates/store/re_protos/proto/rerun/v1alpha1/common.proto
```diff
@@ -222,6 +224,26 @@ message TaskId {
   string id = 1;
 }
 
+// `ChunkKey` provides chunk location details in the data store.
+message ChunkKey {
+  // where is the chunk stored (S3, local file, etc)
+  optional string location_url = 2;
```
zehiko: one problem that with having location url part of the definition is that in real usage we will want to group chunk fetching by the underlying location url, so having a list of ``ChunkKey``s where we know they all belong to the same url makes the url redundant i.e. we have duplication.

That said - it's nice to have ``ChunkKey`` be self-descriptive i.e. just based on the information inside this struct, we can fetch the actual chunk.

zehiko: (on the other hand, if we start supporting a single partition being spread across multiple files / data sources, then each chunk indeed needs the location url)

crates/store/re_protos/proto/rerun/v1alpha1/frontend.proto
```diff
@@ -306,6 +310,21 @@ message GetChunksRequest {
...
+  // Same as with partition ids, we need to know the layer for each chunk
+  repeated string partition_layers = 2;
+
+  // The location of the chunks to fetch. Note that we don't support client side streaming of chunks
+  // as there's no plan to support it on grpc-web at the moment.
+  //
+  // Chunks to fetch (described by their `ChunkKey`)
+  repeated rerun.common.v1alpha1.ChunkKey chunk_keys = 3;
+}
+
```
abey79, line 329: Are the 3 `repeated` arrays supposed to have the same length here? If yes, it feels there should be a single `repeated` and a sub-message. If not, then the semantics should be better explained in comments.

abey79:
Side note: when practical, it would be very nice to also include related methods in `GenericConnectionClient` when introducing new frontend RPCs. I think the cost of doing so should be relatively small if you already go through the hassle of making `ext` helper objects.

crates/store/re_protos/proto/rerun/v1alpha1/common.proto
```diff
@@ -222,6 +224,26 @@ message TaskId {
...
+
+  // additional details about the chunk's location that are specific to the underlying data source (partition type)
+  google.protobuf.Any location_details = 3;
+}
+
+message RrdLocationDetails {
```
abey79: I'm slightly confused by the naming here. Do I understand correctly that this is "the location of a given chunk within a RRD"? If so, how about `ChunkLocationDetails`? Or if it is important to specify that this is a chunk to be found in a rrd, maybe `RrdChunkLocationDetails`?

zehiko: totally agree, this is a specific version of chunk location details, so should probably be ``RrdChunkLocationDetails``

crates/store/re_protos/proto/rerun/v1alpha1/frontend.proto
zehiko: yes, agree at least a comment is required.

But as I read your comment I remembered how @jleibs  defined the ``GetChunksRequest`` for the Chunk scanner service and I wonder if these 3 fields should in a ``DataframePart`` as 3 columns.

zehiko:
> There are multiple things here that strike me as sub-optimal:

> Every chunk key contains some location url, but chunk key are always used in a context where partition id and partition layers must also be provided. This seems unnecessarily complex.

the need to pass partition id and layer comes from the fact we still have to enrich each ``StreamingChunk`` on the backend with that metadata, so unfortunately we have to move that info back and forth. This is why it's separate from the ``ChunkKey`` as it does not describe it's location. I believe we did plan to enrich ``Chunk``s with this metadata to make them self-sufficient in that regard which would make passing this information back and forth obsolete. 

> I dislike the "structure of 3 vecs" (that must have the same length), vs. a vec of structs. Error prone and requires more sanity checking, etc.

This is a pattern we already have in a few places, for example ``CreateIndexRequest`` and I was aiming for consistency. But again, I'm hoping that eventually partition ids and layers goes away and we're left with just ``ChunkKey``s.  But I've intentionally left your comment unresolved as I'd like Jeremy's take on it as he suggested in the design doc using a dataframe as input which would guarantee the lengths (but that also has its downsides).

> IIUC, FetchChunks is supposed to be used in pair with QueryDataset. Yet both rpc have wildly different API.

``QueryDataset`` is right now generic in terms of its response, you can chose to ask for ``chunk_key``, ``partition_id``, ``layer`` and you can then call ``FetchChunks``, but you can also ask for all other information about each chunk that we store in the partition manifest. I'd prefer symetry as well as that'd be even more aligned with Arrow Flight's 2 phase query pattern, but I don't think it's a huge leap to go from currrent ``QueryDatasetResponse`` to that symmetric response. I'd say it' an evolving process that is moving us closer to Flight's protocol. There is an important difference right now, which is that the client (our streaming table provider) is in control of choosing how to do partitioning, while in Flight's protocol server is in control of the distribution. We can still keep that control on the Frontend and chose how we route per partition requests (or we move towards Flight's approach, but incrementally, not immediately). 


> My feeling is that this API is key enough that it should deserve two separate design sub-discussion:

> What should be the frontend API (with a focus on how the user-side will use these API).
> What should be the manifest registry API (with a focus on an efficient implementation).

Introduction of ``FetchChunks`` endpoint is tightly related to introduction of Chunk scanner service and I don't think we can separate these. Frontend will be used as a router for the Chunk scanner service and same information will need to be passed to the Chunk scanner service. i.e. this PR is step in the direction of having the 2nd phase request (fetching chunks) be served by a separate service (so I don't think we will 2 different APIs).

p.s. I think it's impractical to make a PR a design discussion, so I suggest you, Jeremy and me have a discussion on this topic.

crates/store/re_protos/proto/rerun/v1alpha1/common.proto
jleibs: What's the rationale for this being outside of, rather than inside, the `RrdChunkLocationDetails`. It's not obvious to me there's a reason to make this non-opaque.

crates/store/re_protos/proto/rerun/v1alpha1/common.proto
```diff
@@ -222,6 +224,26 @@ message TaskId {
...
+  // where is the chunk stored (S3, local file, etc)
+  optional string location_url = 2;
+
+  // additional details about the chunk's location that are specific to the underlying data source (partition type)
+  google.protobuf.Any location_details = 3;
+}
+
+message RrdChunkLocationDetails {
+  // Chunk unique identifier
+  rerun.common.v1alpha1.Tuid chunk_id = 1;
```
jleibs: Conversely, chunk_id is the one thing that seems like it could make sense to move out the specific `RrdChunkLocationDetails` and into the public `ChunkKey` context. It would make sense to me that given a ChunkKey, a user should be able to to what `ChunkId` that key will resolve to.

crates/store/re_protos/proto/rerun/v1alpha1/frontend.proto
jleibs: My main motivation for using DataframePart here is I believe that's already the structure of the QueryResponse -- so we already have a literal arrow-array of ChunkKeys. Converting back to a repeated protobuf seems arbitrary.

Likewise, partition_id /  and partition_layer columns must have been populated in that original query response as well or we wouldn't have the data to craft this response.

So yeah, my vote would be make this some form of arrow dataframe payload, but expand my original proposal to specify that it takes a payload of up to 3 columns.

crates/store/re_protos/proto/rerun/v1alpha1/manifest_registry.proto
```diff
@@ -495,6 +498,27 @@ message GetChunksResponse {
   repeated rerun.log_msg.v1alpha1.ArrowMsg chunks = 1;
 }
 
+message FetchChunksRequest {
```
jleibs, line 501: Same comment as above.

jleibs (changes requested):

jleibs:
> What should be the frontend API (with a focus on how the user-side will use these API).

My TLDR is:
- Query returns a dataframe
- User filters the dataframe and selects the 1-3 required columns
  - Partition/layer are only necessary if they actually need those re-injected into the chunks as part of a dataframe-query that uses those columns.
- User submits that filtered dataframe back to cloud
- Cloud returns stream of chunks

crates/store/re_protos/proto/rerun/v1alpha1/frontend.proto
abey79: One difficulty of using dataframe is the polymorphic nature of `location_details`. I can think of a situation where we have a base layer that uses mcap, and additional embedding layers that might use rrd. This would make common for queries to return chunks of heterogenous provenance, leading to a space dataframe whose schema is the union of "location types".

abey79:
> > What should be the frontend API (with a focus on how the user-side will use these API).
> 
> My TLDR is:
> 
> * Query returns a dataframe
> * User filters the dataframe and selects the 1-3 required columns
>   
>   * Partition/layer are only necessary if they actually need those re-injected into the chunks as part of a dataframe-query that uses those columns.
> * User submits that filtered dataframe back to cloud
> * Cloud returns stream of chunks

What is the motivation for using a dataframe here, instead of a repeating 1-to-3-field protobuf struct? My hunch is that for the purpose of chunk caching, using native protobuf is more convenient for most languages, including Rust (which is the only one we care about for that). Are there other intended workflows where the dataframe format makes sense?

More specifically, I fail to see how a dataframe of essentially opaque IDs can be useful in user land, we dataframe prevails.

crates/store/re_protos/proto/rerun/v1alpha1/common.proto
zehiko: was going back and forth on it, mostly wondering if we'll have a ``ChunkId`` at this point for other formats. (but once we have more than a single supported format, we can adjust the abstractions). Makes sense to change now.

crates/store/re_protos/proto/rerun/v1alpha1/common.proto
zehiko: my rationale was that every type of location will need to be identified by a url. But as I write that I realize that for in-memory implementation that would not make sense, hence for now I'll move it into the opaque ``location_details`` (which I guess can be renamed to ``location``.

zehiko:
Closing in favour of https://github.com/rerun-io/rerun/pull/11054

PR #10853, opened 2025-08-11
Add automatic undo-bug-detection code

emilk:
### Related
* https://github.com/rerun-io/rerun/issues/10304
* https://github.com/rerun-io/rerun/pull/10355

### What
We previously had a bug where we would write to the blueprint store every frame, which in turn made undo break, silently.

This PR makes the breakage loud (warn-log SPAM) in debug builds, which should allow us to catch this bug much earlier next time.

To test this I reverted the fix for the previous bug (https://github.com/rerun-io/rerun/pull/10355) and indeed, the detector works.

I've done a bit of testing of our UI, and I've found nothing on `main` that triggers the new warning.

### DIFF
--- a/crates/viewer/re_viewer_context/src/undo.rs
+++ b/crates/viewer/re_viewer_context/src/undo.rs
@@ -145,7 +145,7 @@ impl BlueprintUndoState {
         // then you don't want an undo-point each time you pause the mouse - only on mouse-up!
         // So we still need a call to `is_interacting`, no matter what.
         // We must also make sure that this doesn't ignore actual bugs
-        // (writing spurious data to the bluprint store each frame -
+        // (writing spurious data to the blueprint store each frame -
         // see https://github.com/rerun-io/rerun/issues/10304 and the comment below for more info).
 
         // Don't store too many undo-points:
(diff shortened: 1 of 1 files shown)


crates/viewer/re_viewer_context/src/undo.rs
```diff
@@ -114,30 +124,66 @@ impl BlueprintUndoState {
...
         }
 
         // TODO(emilk): we should _also_ look for long streaks of changes (changes every frame)
         // and disregard those, in case we miss something in `is_interacting`.
-        // Note that this on its own won't enough though - if you drag a slider,
+        // Note that this on its own isn't enough: if you drag a slider,
         // then you don't want an undo-point each time you pause the mouse - only on mouse-up!
+        // So we still need a call to `is_interacting`, no matter what.
+        // We must also make sure that this doesn't ignore actual bugs
+        // (writing spurious data to the bluprint store each frame -
```
grtlr:
```suggestion
        // (writing spurious data to the blueprint store each frame -
```

crates/viewer/re_viewer_context/src/undo.rs
```diff
@@ -25,15 +25,19 @@ pub struct BlueprintUndoState {
     /// If `None`, use the max time of the blueprint timeline.
     current_time: Option<TimeInt>,
 
-    /// Interesting times to undo/redo to.
+    /// The keys forms a set of interesting times to undo/redo to.
```
grtlr:
```suggestion
    /// The keys form a set of interesting times to undo/redo to.
```

PR #10849, opened 2025-08-08
Fix warnings produced by Python examples

emilk:
By default, we treat warnings as errors with `PYTHONWARNINGS=error`.

This means our examples fail to build if our code, or one of our dependencies, produces a logged warning.

This PR fixes a few such warnings, but for `detect_and_track_objects` the problem in in a 3rd party library, so there I added a mechanism for ignoring warnings.

Now, finally, `pixi run -e wheel-test build-examples rrd --channel "nightly"  web_viewer/examples` works again on my machine. Hopefully it will work on CI too!

## TODO
* [x] `@rerun-bot full-check`

PR #10854, opened 2025-08-11
Add ruff type checking rules and run them for auto fix

ntjohnson1:
### Related
Fix the bug found here  https://github.com/rerun-io/rerun/pull/10849/commits/3e4fb923bde51876fc323abd76b242b4b0e80f3a

### What
Adds this ruleset https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc reading over the rules they all seem reasonable and not too cumbersome.

My summary (expect it to be lossy):
1. If a type is ONLY used for type checking it adds it to a TYPE_CHECKING block
2. If a type is only imported for TYPE_CHECKING block but is used elsewhere then it fixes it
3. Some other small formatting details for consistency


rerun_py/tests/test_types/components/affix_fuzzer16.py
```diff
@@ -6,7 +6,7 @@
 from __future__ import annotations
 
 from collections.abc import Sequence
-from typing import Any, Union
+from typing import TYPE_CHECKING, Any, Union
```
ntjohnson1, line 9: Ugh this is an autogenerated file so I need to look at the code gen stuff

ntjohnson1: Hmm, I see a comment about run in the actual code gen source. Is the expected workflow to run codegen then run ruff so this change is ok? I don't think it would be trivial to insert this into our codegen so a two pass approach doesn't seem unreasonable.

ntjohnson1:
I branched off not latest main so it was cleaner with all the changes to rebase. That shouldn't mess up a review in progress. My one question is about if we are ok running codegen then running py-fmt and having that fix the codegen code. If that's ok then this should be set.

I guess one other broad question is: are people actually doing real conflict resolution on our pixi lock file? I feel like that's cumbersome and usually just regenerate a fresh lockfile off the latest to resolve the collision but maybe that's making things worse by increasing the churn.

ntjohnson1:
NOTE: My last commit brings up yet another but slightly more narrow edge case.

In `_log` we use `AsComponents` as a type hint and import it from `_baseclasses` but ALSO import `AsComponents` from `_log` in a bunch of other files. I'm not sure if there is a lint that handles these re-exported symbols situations.

emilk:
We already run `cargo fmt` in our codegen (`re_types_builder`) to format the generated Rust code, so funning python fmt/lint in codegen sounds like a good idea, and I wonder if we don't already do that 🤔 

EDIT: we do: `crates/build/re_types_builder/src/format/python.rs`

ntjohnson1:
> EDIT: we do: crates/build/re_types_builder/src/format/python.rs

Hmm I thought I had to run the formatter after I ran codegen but maybe I did `py-fmt` then `codegen` and the follow up codegen was fine. I guess I'm not concerned then.

emilk:
The CI for each PR will check that the files you check in are the same as what codegen would produce, so if the CI is happy, I'm happy :)

emilk:
@rerun-bot full-check

Running this just to be extra sure

PR #10858, opened 2025-08-11
Fix loading MCAP on wasm

emilk:
Also avoid running external data loaders when loading an .mcap, or a .png, or any other file type that Rerun natively supports.

### DIFF
--- a/crates/store/re_data_loader/src/lib.rs
+++ b/crates/store/re_data_loader/src/lib.rs
@@ -481,10 +481,8 @@ pub const SUPPORTED_POINT_CLOUD_EXTENSIONS: &[&str] = &["ply"];
 
 pub const SUPPORTED_RERUN_EXTENSIONS: &[&str] = &["rbl", "rrd"];
 
-#[cfg(not(target_arch = "wasm32"))]
+/// 3rd party formats with built-in support.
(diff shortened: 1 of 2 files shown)


crates/store/re_data_loader/src/lib.rs
```diff
@@ -504,10 +502,17 @@ pub fn supported_extensions() -> impl Iterator<Item = &'static str> {
...
 pub fn is_supported_file_extension(extension: &str) -> bool {
-    SUPPORTED_IMAGE_EXTENSIONS.contains(&extension)
-        || SUPPORTED_VIDEO_EXTENSIONS.contains(&extension)
-        || SUPPORTED_MESH_EXTENSIONS.contains(&extension)
-        || SUPPORTED_POINT_CLOUD_EXTENSIONS.contains(&extension)
-        || SUPPORTED_RERUN_EXTENSIONS.contains(&extension)
-        || SUPPORTED_TEXT_EXTENSIONS.contains(&extension)
+    debug_assert!(
+        !extension.starts_with('.'),
+        "Expected extion without period, but got {extension:?}"
```
grtlr:
```suggestion
        "Expected extension without period, but got {extension:?}"
```

grtlr (approved):
Thank you for cleaning this up!

PS: something still seems wrong with our typo checking.

Issue #10822, opened 2025-08-07
Automatically update address bar with dataplatform links


Wumpf:
Address bar should always be a useful shareable link that can be copied

Issue #10242, opened 2025-06-16
Use URLs to make referring to data simpler


nikolausWest:
if I find something strange in the data I want to easily be able to save a link to it for later or share with a colleague. The simplest way for me to do that is to copy the URL in the URL bar. To do that, navigating the catalog needs to update the URL bar.

Web viewer:

<img src="https://uploads.linear.app/93919c5e-abe4-4cd5-b77d-7e3eafe3e62d/a06713d9-99b9-4ab1-b118-9978a66d714e/84fee409-4c8f-4239-8d40-7e4bbeaf0df9?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzkzOTE5YzVlLWFiZTQtNGNkNS1iNzdkLTdlM2VhZmUzZTYyZC9hMDY3MTNkOS05OWI5LTRhYjEtYjExOC05OTc4YTY2ZDcxNGUvODRmZWU0MDktNGM4Zi00MjM5LThkNDAtN2U0YmJlYWYwZGY5IiwiaWF0IjoxNzUwMDY2NjY2LCJleHAiOjMzMzIwNjI2NjY2fQ.7C9b3RyMGuALe91jQgKeJgZyyDityUMG1Ni17pUeC4E " alt="Screenshot 2025-06-16 at 11.23.38.png" width="619" data-linear-height="298" />

[...]

PR #10863, opened 2025-08-12
Improve browser navigation for http & Rerun dataplatform links

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/10822
    * I'd like to expand it a bit more though, created a follow-up ticket here https://github.com/rerun-io/rerun/issues/10866
* Fixes https://github.com/rerun-io/rerun/issues/10242
    * again there's more things to be done here to make this even better, but the basic feature is there 

### What

https://github.com/user-attachments/assets/f5f7b02f-f5e2-48a2-bbab-3f1551597599

Most of the improvements are for the dataplatform where you can now navigate between recordings and server selections (table/entry selections aren't available _yet_). However, there's also plenty of improvements for using OSS http links: previously, there was some very limited browser history/navigation for `url=http://something/some.rrd` which was used for going back and forth from the example screen, but there were issues with it: you couldn't navigate between recordings and there was a lot of needless reload happening.

Under the hood this is actually a more impactful change still as this unifies & simplifies various URL opening code paths. The idea being that almost always you should be fine to just go with egui's url opening (`ui.ctx().open_url(egui::OpenUrl::same_tab(uri))`) and the desired & reasonable result _on both native and web_ - this change opens the path for more "import from link"-like functionality that will work just the same on native.

All in all, this PR ends up doing several intertwined things at once:

* Improve browser history/navigation (history.rs) to deal with more different kind of changes
* Simplify & unify link opening code paths
    * Removes some web-only constructs in the process 
* adding a DataSource (which is what most links do) will no longer load data that's already loaded

Unfortunately, there's still various quirks in the code since sometimes we have hacked-in additional information what should happen after a link is opened. It's unfortuantely not always easy to follow all these rules. I did testing with a standalone served viewer and ran with our nascent internal dataplatform notebooks to make sure nothing important broke.


crates/store/re_data_source/src/data_source.rs
```diff
@@ -70,24 +69,26 @@ impl DataSource {
...
-                    false // Too many dots; assume an url
+                    // We use a simple heuristic here: if there are multiple dots, it is likely an url,
+                    // like "example.com/foo.zip".
+                    // If there is only one dot, we treat it as an extension and look it up in a list of common
+                    // file extensions:
+
+                    let parts = uri.split('.').collect_vec();
+                    if parts.len() == 2 {
+                        // Extension or `.com` etc?
+                        re_data_loader::is_supported_file_extension(parts[1])
```
abey79: This is not new but I'm slightly surprised that here we check for supported file extension whereas we unconditionally return `true` in the previous cases.

crates/store/re_data_source/src/data_source.rs
```diff
@@ -97,35 +98,39 @@ impl DataSource {
...
                 uri,
                 select_when_loaded: true,
-            };
+            })
+        } else if url.ends_with(".rrd") || url.ends_with(".rbl") {
+            Some(Self::RrdHttpUrl {
+                url: url.to_owned(),
+                follow: false,
+            })
+        } else {
```
abey79: Isn't this brittle against `?key=val&...` stuff that we (or $DEITY knows what other process) might happen in the future? Maybe worth parsing the string with the `url` crate and check the actual address?

crates/store/re_data_source/src/data_source.rs
```diff
@@ -97,35 +98,39 @@ impl DataSource {
...
         }
-
-        // by default, we just assume an rrd over http
-        Self::RrdHttpUrl { uri, follow: false }
     }
 
     pub fn file_name(&self) -> Option<String> {
         match self {
-            Self::RrdHttpUrl { uri: url, .. } => url.split('/').next_back().map(|r| r.to_owned()),
+            Self::RrdHttpUrl { url, .. } => url.split('/').next_back().map(|r| r.to_owned()),
```
abey79, line 142: This could break for the same reason as previous comment, but would not if the url was sanitised then.

crates/store/re_data_source/src/data_source.rs
```diff
@@ -334,69 +339,74 @@ where
...
+            "foo.png",
+            "/foo/bar/baz",
+            "D:/file",
+        ];
+        let http = [
+            "http://foo.rrd",
+            "example.zip/foo.rrd",
+            "www.foo.zip/foo.rrd",
+            "www.foo.zip/blueprint.rbl",
+        ];
```
abey79:
```suggestion
        let http = [
            "http://foo.rrd",
            "http://foo.rrd?useless_param=1",
            "example.zip/foo.rrd",
            "www.foo.zip/foo.rrd",
            "www.foo.zip/blueprint.rbl",
        ];
```

abey79: (lol I'm only allowed being an annoying prick repeating the same thing a third time here because of the `mod test` induced whitespace change 😅)

crates/utils/re_smart_channel/src/lib.rs
```diff
@@ -29,6 +29,7 @@ pub enum SmartChannelSource {
     /// The channel was created in the context of loading an `.rrd` file over http.
     ///
     /// The `follow` flag indicates whether the viewer should open the stream in `Following` mode rather than `Playing` mode.
+    // TODO(andreas): having follow in here is a bit weird. This should be part of the link fragments instead.
```
abey79, line 32: Having a properly parsed `RrdUri` (in the `RedapUri` style) would address this (and all my previous comments).

crates/viewer/re_global_context/src/command_sender.rs
```diff
@@ -18,12 +18,11 @@ pub enum SystemCommand {
...
 
-    /// Load some data.
+    /// Load data from a given data source.
+    ///
+    /// Will not load any new data if the source is already one of the active data sources.
     LoadDataSource(DataSource),
 
-    /// Clear everything that came from this source, and close the source.
-    ClearSourceAndItsStores(re_smart_channel::SmartChannelSource),
-
```
abey79, line 26: ❤️

crates/viewer/re_viewer/src/app.rs
```diff
@@ -779,6 +772,176 @@ impl App {
...
+                            let time_ctrl = rec_cfg.time_ctrl.get_mut();
+                            time_ctrl.set_play_state(
+                                entity_db.times_per_timeline(),
+                                PlayState::Following,
+                            );
+                        }
+
+                        let store_id = entity_db.store_id().clone();
+                        drop(all_sources);
+                        self.make_store_active_and_highlight(store_hub, egui_ctx, &store_id);
```
abey79: Technically this could be a blueprint right? Should it then have the same `if store_id.is_recording()` check as the `DataSource::FilePath` branch below?

crates/viewer/re_viewer/src/app.rs
```diff
@@ -779,6 +772,176 @@ impl App {
...
+            }
+
+            #[cfg(not(target_arch = "wasm32"))]
+            DataSource::Stdin => {
+                let new_source = SmartChannelSource::Stdin;
+                if all_sources.any(|source| source.is_same_ignoring_uri_fragments(&new_source)) {
+                    if let Some(entity_db) = store_hub.find_recording_store_by_source(&new_source) {
+                        let store_id = entity_db.store_id().clone();
+                        drop(all_sources);
+                        self.make_store_active_and_highlight(store_hub, egui_ctx, &store_id);
```
abey79: same? (maybe this bit of code could go to a local fn or closure to avoid duplication?)

abey79 (approved):
Great work! My comments are a bit obnoxious on the url parsing stuff. Though I do think that an explicit `RrdUri` parsed-url type could make stuff a bit cleaner/more robust, I dont feel _that_ strongly about it (and can be left out of this PR either way).

crates/viewer/re_viewer/src/open_url.rs
```diff
@@ -0,0 +1,195 @@
...
+    re_log::debug!("Opening URL: {url:?}");
+
+    if let Some(mut data_source) = DataSource::from_uri(re_log_types::FileSource::Uri, url) {
+        if let DataSource::RerunGrpcStream {
+            select_when_loaded, ..
+        } = &mut data_source
+        {
+            *select_when_loaded = select_redap_source_when_loaded;
+        } else if let DataSource::RrdHttpUrl { follow, .. } = &mut data_source {
+            *follow = follow_if_http;
```
jprochazk, line 47: This feels like it should be
```suggestion
            *follow = follow_if_http || *follow;
```

Same above with `select_when_loaded`, we shouldn't discard the requested value from the data source, right?

crates/viewer/re_viewer/src/app.rs
```diff
@@ -779,6 +772,176 @@ impl App {
...
+    /// Loads a data source into the viewer.
+    ///
+    /// Tries to detect whether the datasource is already present (either still streaming in or already loaded),
+    /// and if so, will not load the data again.
+    /// Instead, it will only perform any kind of selection/mode-switching operations associated with loading the given data source.
+    ///
+    /// Note that we *do not* change the display mode here _unconditionally_.
+    /// For instance if the datasource is a blueprint for a dataset that may be loaded later,
+    /// we don't want to switch out to it while the user browses a server.
+    fn load_datasource(
```
jprochazk: Nit
```suggestion
    fn load_data_source(
```

crates/viewer/re_viewer/src/web.rs
```diff
@@ -197,20 +198,23 @@ impl WebHandle {
...
             return;
         };
+
+        // TODO(andreas): should follow_if_http be part of the fragments?
         let follow_if_http = follow_if_http.unwrap_or(false);
-        if let Some(rx) = url_to_receiver(
-            &self.connection_registry,
-            app.egui_ctx.clone(),
+        let select_redap_source_when_loaded = true;
+        if try_open_url_in_viewer(
```
jprochazk, line 208: This used to repaint, but it no longer does. Should it?

crates/viewer/re_viewer/src/web_tools.rs
```diff
@@ -80,129 +76,6 @@ pub fn window() -> Result<Window, JsValue> {
     web_sys::window().ok_or_else(|| js_error("failed to get window object"))
 }
```
jprochazk, line 78: Good riddance 😄

crates/viewer/re_viewer/src/startup_options.rs
```diff
@@ -75,11 +75,27 @@ pub struct StartupOptions {
...
-    /// open example recording, see [`crate::history`].
+    /// open example or redap recording, see [`crate::history`].
     #[cfg(target_arch = "wasm32")]
     pub enable_history: bool,
 }
 
+impl StartupOptions {
+    /// Returns `StartupOptions::enable_history` on web, and `false` on native.
+    #[allow(clippy::unused_self)] // Only used on web.
+    pub fn enable_web_history(&self) -> bool {
```
jprochazk: Nit: This reads like a verb, but doesn't take `&mut self`, which confused me for a bit.
```suggestion
    pub fn web_history_enabled(&self) -> bool {
```

(The field comes from `StartupOptions`, where it sort of makes sense for it to be a verb. You are enabling it through the action of setting that field to `true`)

crates/viewer/re_viewer/src/history.rs
```diff
@@ -277,6 +273,11 @@ impl private::Sealed for History {}
 
 impl HistoryExt for History {
     fn push_entry(&self, entry: HistoryEntry) -> Result<(), JsValue> {
+        // Check if this is the exact same entry as before, if so don't do anything.
+        if self.current_entry()?.unwrap_or_default() == entry {
```
jprochazk, line 277: We have a handler for `popstate` which does non-idempotent things. Are we sure this is the right behavior? In what case can you end up with duplicate entries, and why would that be a problem requiring a solution here?

crates/store/re_data_source/src/data_source.rs
```diff
@@ -97,35 +98,39 @@ impl DataSource {
...
                 select_when_loaded: true,
-            };
+            })
+        } else if url.ends_with(".rrd") || url.ends_with(".rbl") {
+            Some(Self::RrdHttpUrl {
+                url: url.to_owned(),
+                follow: false,
+            })
+        } else {
+            None
```
jprochazk: This seems like a welcome change, but I sure do hope nobody depends on our previous assumption here

jprochazk (commented):
The changes in general seem sound to me, I have a few nits and a few questions.

Most importantly we have to test this in every place we put the web viewer. That's at least:
- [x] `app.rerun.io` (I believe you already tested this one)
- [ ] Notebooks - `viewer.open_url`
- [x] `rerun.io` landing page -> Viewer example navigation
- [ ] `rerun.io/docs`

I can help with testing the `rerun.io` stuff tomorrow. I don't want that to block you from merging, so if you feel like it's in a good state, feel free to do that. We can follow up on it later if anything currently untested is broken.

crates/store/re_data_source/src/data_source.rs
Wumpf: oh that's weird indeed.
This heuristic is also weird because it doesn't handle relative file paths either

crates/store/re_data_source/src/data_source.rs
Wumpf: yeah, I hope we didn't depend on this too much. It could be that this breaks links to non-rrd files that we previously would just accept silently. But in that case I'd like to handle this slightly differently anyways

crates/store/re_data_source/src/data_source.rs
Wumpf: it's a really good comment though. Illustrates that we shouldn't have a ton of different places for this kind of parsing!

crates/utils/re_smart_channel/src/lib.rs
Wumpf: agreed, let's shoot for that at a later point. Don't want to spend too much time on "legacy"ish http urls right now

crates/viewer/re_viewer/src/app.rs
Wumpf: it can't be because I made `find_recording_store_by_source` filter by that. I actually ran into that bug previously
Leaving a comment to stress it

.. and an assert for good measure

crates/viewer/re_viewer/src/history.rs
Wumpf: this would happen for instance if a recording gets activated that's already active. Maybe this should be a separate method for "push only if not new" of sorts?
I feel like I'm not understanding the `popstate` ramifications well enough, let's chat about it to make sure I'm not doing something stupid

crates/viewer/re_viewer/src/open_url.rs
Wumpf: Hmmm that's a really good point, but the problem is that `DataSource` has its own opinionated setting for both of these parameters. Ideally they'd just be part of the URL in both cases, I really dislike that they're not. I'll add some commentary

crates/viewer/re_viewer/src/web.rs
Wumpf: I think it should just to be sure, good catch. I used to have repaint direclty inside of `try_open_url_in_viewer`, but there's a case where this causes a deadlock since we're already inside a `egui_ctx` capture and this was kinda hard to catch for me

Wumpf:
great reviews. Thank you guys!
Will do some more testing as recommended. Also, let's follow-up on `popstate`

crates/viewer/re_viewer/src/history.rs
jprochazk: On second thought this check is unnecessary. We already check for duplicate entries during `popstate`: https://github.com/rerun-io/rerun/blob/andreas/dataplatform-url-update/crates/viewer/re_viewer/src/history.rs#L142

Wumpf: but it already takes up "space" in the back button then, no? I'll try it

Wumpf: yeah, if I don't do this we accumulate too many entries here - e.g. switching between the application id view & the recording view generates new entries that we can't actually distininguish.

All of these here except the last are the same page:
image
Guess we should look into giving those better names... later :)

jprochazk:
Can confirm the landing page -> "try in browser" path still works. The recordings aren't being closed, which may be problematic as you could run out of memory. :shrug:

jprochazk:
`rerun.io/docs` looks unaffected as well

Wumpf:
tested `open_url` on dataplatform welcome notebook (hacking some additional open_url calls in)

Wumpf:
> The recordings aren't being closed, which may be problematic as you could run out of memory. 🤷

Makes sense though, in line with the changes I did. I prefer that behavior here. GC (or user) should take care of closing imho

crates/viewer/re_viewer/src/history.rs
jprochazk: I believe changing the names of history entries is impossible. There's a `title` parameter in `replaceEntry` that no browser currently does anything with. MDN call it `unused`: https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState
Issue for context: [whatwg/html#2174](https://www.github.com/whatwg/html/issues/2174)

jprochazk: I looked through the `popstate` code a bit more. Let's leave this check in there. The proper thing to do would be to not store duplicate entries in the first place, but this seems like a reasonable solution.

PR #10856, opened 2025-08-11
Introduce MCAP layers and `rerun mcap` CLI tools

grtlr:
### Motivation

The previous plugin system needed to be reworked because it did not readily support the following requirements:

* No grouping of message types (would be cumbersome to specify message types in CLI).
* Only "per-message" plugins, no global (static) plugins.
  * Emitting `TextLog` warnings not tweakable based on selected plugins. 
* Abstraction for emitting chunks (many `Vec<>` allocations).

### What

* Deprecates old plugin system and replaces that with new layer system.
* Tier-ed trait approach: `trait Layer` and specialized `trait MessageLayer`.
* Groups all `ros2msg` under a plugin
* Basic proof of concept of a CLI command that allows filtering of layers.

PR #10864, opened 2025-08-12
Check if our CLI docs match `rerun man`

### LINKED ISSUE #10640: Automate `pixi run rerun man > cli.md`
Check for unstaged changes on CI etc

grtlr:
### Related

* #10641
* Closes #10640 

### What

This adds a new script that checks if the CLI docs are up to date with the output of `rerun man` (which we use to generate the docs).

You can try it out via:

```sh
pixi run python scripts/ci/check-cli-docs.py
```

The script (written by Claude) accepts a command to run as an optional argument so that we can run it in CI with minimal impact (no rebuilding).

### TODO:

* [x] Based-on: #10856 (which should introduce an error).


scripts/ci/check-cli-docs.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
+"""Check if CLI documentation is up-to-date."""
```
jprochazk: Everything else uses underscores, please rename the file to `check_cli_docs` 😅

scripts/ci/check-cli-docs.py
```diff
@@ -0,0 +1,78 @@
...
+from __future__ import annotations
+
+import argparse
+import difflib
+import subprocess
+import sys
+from pathlib import Path
+
+# TODO(grtlr): Is there a better way to call this?
+default_command = ["cargo", "run", "--features", "release", "--", "man"]
```
jprochazk: The correct command is `pixi run man`

scripts/ci/check-cli-docs.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Check if CLI documentation is up-to-date")
+    parser.add_argument(
```
jprochazk: Why is this an argument?

jprochazk (changes requested):

scripts/ci/check-cli-docs.py
grtlr: In CI, I want to run against the actual binary that was built to avoid any additional work: https://github.com/rerun-io/rerun/pull/10864/files#diff-63182be3cf3f909cddc762c938835d1a29873b1f35b30f375fe2d0c9a7fb8abbR191

jprochazk: I see now. That probably won't work. `pixi run man` uses `--all-features`, and we need that to ensure `man` is complete. The one that's built for Python wheels does not, iirc.

I would move this check entirely out of `test_wheels` and into `reusable_checks.yml`, that way it can run on a separate CI machine

scripts/ci/check-cli-docs.py
grtlr: Fair!

scripts/ci/check-cli-docs.py
grtlr: Thanks! That command already updates the docs. I've added it to the hint output and used its definition in `default_command`.

scripts/ci/check-cli-docs.py
grtlr: I confirmed that the output is the same, but I'm happy to move things to a different check.

jprochazk: We definitely want to generate the manual with the same command always, and that command must include `--all-features`. Moving it to a different workflow would be better here, because it's not a "python" thing. A separate job like the one for markdown lint in `reusable_check.yml` won't slow down the CI too much, since it runs in parallel.

grtlr:
@jprochazk As per your comment, I've moved the checks to the `reusable_checks_rust`  file and this PR is ready for review again.

I'm not entirely sure about the approach, because this means we now have to do a full build (including WASM; the CLI basically contains everything) at that stage, which adds around ~5:30min (but it happens in parallel ofc).

jprochazk (approved):
We need a Build System ™️

PR #10870, opened 2025-08-12
More aggressive linting of `...` vs `…`

emilk:
Alternative PR titles:
* "How to lose friends and alienate people"
* "I'm sorry"
* "Undiagnosed, but everyone's pretty sure"

## Motivation
We lint against writing ellipsis using three dots for the sake of our UI:
* We want it consistent
* We want it beautiful (`…` looks different from `...`)
* We don't want linebreaks in the middle of an ellipsis

This lint is therefore most important in user-facing code, such as the UI, but we also care about beautiful docs.

I've configured my OS to write an ellipsis character with `⌥;`


emilk:
@rerun-bot full-check

grtlr:
I restarted the failed C++ job in full CI

Issue #9144, opened 2025-02-26
Make the viewer robust with datatype generics


emilk:
If the viewer can handle a more broad set of arrow datatypes, then it is easier to maintain forward and backward data compatibility.

### Related

* For datatype generics: [docs.rs/narrow/latest/narrow](<https://docs.rs/narrow/latest/narrow/>)

Wumpf:
Related among others to
* https://github.com/rerun-io/rerun/issues/7049

PR #10777, opened 2025-07-30
Add arrow data inspector in seleciton panel

lucasmerlin:
### Related

* closes https://github.com/rerun-io/rerun/issues/10750

### What

Instead of formatting arrow data as a String and displaying that, this adds an interactive arrow data explorer:

https://github.com/user-attachments/assets/84265daf-e829-4cb3-96cc-bf24bf581a0c

It's still usable if data is rapidly changing:
https://github.com/user-attachments/assets/4b604967-c7f5-472f-8262-a1ca7fbfd121

And it even handles large arrays:
https://github.com/user-attachments/assets/3487d7b6-4077-415f-9ab0-c5863c2c93e8

PR #10875, opened 2025-08-12
Support 2GiB+ videos: Encode `Blob` as arrow `Binary`

emilk:
This changes the encoding of the component column containing bobs from `List<List<u8>>` to `List<Binary>`, i.e. `BinaryArray` or `LargeBinaryArray` (both are supported!).

Old data is migrated on load.

This adds support for blobs larger than 2 GiB, e.g. large video files.

In the viewer we now support both 32-bit and 64-bit offsets for the binary data (`BinaryArray vs LargeBinaryArray`).
This is _technically_ a type of datatype generics. Sort of. See https://github.com/rerun-io/rerun/issues/9144

* The SDK code (Rust, Python, C++) always encoded as 64-bit (supporting >2GiB videos).
* Legacy blobs are converted from 32-bit List<u8> to 64-bit binary

## TODO
* [x] Fix python
* [x] Merge https://github.com/rerun-io/rerun/pull/10899
* [x] Merge https://github.com/rerun-io/rerun/pull/10898
* [x] Use 63-bit `LargeBinary` instead of 31-bit `Binary`
  * Closes https://github.com/rerun-io/rerun/issues/10929
  * Closes https://github.com/rerun-io/rerun/issues/10973
* [x] `@rerun-bot full-check`


crates/store/re_types/definitions/rerun/datatypes/blob.fbs
```diff
@@ -13,5 +13,5 @@ table Blob (
   "attr.rust.repr": "transparent",
   "attr.rust.tuple_struct"
 ) {
-  data: [ubyte] (order: 100);
+  data: [ubyte] (order: 100, "attr.rerun.override_type": "binary");
```
emilk, line 16: This is the key change

emilk:
@rerun-bot full-check

crates/viewer/re_ui/tests/snapshots/arrow_ui.png
emilk, line 1: The new arrow UI never shows the contents of blob, just their size

emilk:
@rerun-bot full-check

emilk:
This is a big PR. I'll split it in two.

emilk:
@rerun-bot full-check

crates/store/re_types/src/archetypes/asset_video_ext.rs
```diff
@@ -52,7 +52,7 @@ impl AssetVideo {
         re_tracing::profile_function!();
 
         let Some(blob_bytes) = self.blob.as_ref().and_then(Blob::serialized_blob_as_slice) else {
-            return Ok(Vec::new());
+            return Err(re_video::VideoLoadError::NoVideoTrack); // Error type is close enough
```
emilk, line 55: drive-by bug fix: previously we would silently ignore this error, which cost me a few minutes of hair-pulling

crates/build/re_types_builder/src/codegen/rust/deserializer.rs
```diff
@@ -824,7 +834,7 @@ fn quote_arrow_field_deserializer(
             quote!(#fqname_use::from_arrow_opt(#data_src).with_context(#obj_field_fqname)?.into_iter())
         }
 
-        _ => unimplemented!("{datatype:#?}"),
+        DataType::Object { .. } => unimplemented!("{datatype:#?}"),
```
grtlr, line 892: Nice!

crates/build/re_types_builder/src/data_type.rs
```diff
@@ -153,8 +153,10 @@ impl DataType {
 pub enum LazyDatatype {
     Atomic(AtomicDataType),
 
+    /// A list of bytes of arbitrary length.
```
grtlr, line 157: Just for visibility (you don't have to change anything): The encoding is not just a list of bytes, but a [prefixed list of bytes](https://arrow.apache.org/docs/cpp/api/datatype.html):

> Bytes view type with 4-byte prefix and inline small string optimization.

Might only ever be important in cases were we want to memory map something.

grtlr (changes requested):
The overall changes look god to me!

I think we should update some of our displays in `re_component_ui` and update the test cases.

We still use the "123 MB" representation for `UInt8Array`s, which should probably be changed back to the standard list rendering.

Test cases that I have stumbled upon

* `custom_empty_array_*` -> List rendering
* `custom_small_array_*` -> List rendering
* `custom_large_blob_*` -> Change to `BinaryArray`
* Maybe add a missing `custom_large_array`

emilk:
> The overall changes look good to me!
> 
> I think we should update some of our displays in `re_component_ui` and update the test cases.
> 
> We still use the "123 MB" representation for `UInt8Array`s, which should probably be changed back to the standard list rendering.

Good catch - fixed. Though at this point I'm only creating merge conflicts for https://github.com/rerun-io/rerun/pull/10777 😬

crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_narrow/custom_large_blob_any_value_large_blob.png
grtlr, line 1: This test should use `BinaryArray` under the hood and therefore show 2.9 KiB.

grtlr (changes requested):
Yeah, it might create a conflict for @lucasmerlin (sorry for that!), but the snapshots should remain stable, so it's good to check if we don't regress again (for example via merge races).

emilk:
> but the snapshots should remain stable, so it's good to check if we don't regress again (for example via merge races).

https://github.com/rerun-io/rerun/pull/10777 has a lot of changed snapshot images… more than GitHub want to even load by default

crates/build/re_types_builder/src/data_type.rs
emilk: You're reading the docs for "BinaryView", which is something else

crates/build/re_types_builder/src/objects.rs
```diff
@@ -1236,14 +1243,28 @@ impl Type {
...
         if let Some(type_override) = attrs.try_get::<String>(fqname, ATTR_RERUN_OVERRIDE_TYPE) {
-            match (typ, type_override.as_str()) {
-                (FbsBaseType::UShort, "float16") => {
-                    return Self::Float16;
+            match type_override.as_str() {
+                "binary" => {
+                    if typ == FbsBaseType::Vector && field_type.element() == FbsBaseType::UByte {
+                        return Self::Binary;
+                    } else {
+                        panic!("{fqname}: 'binary' can only be used on '[ubyte]', git {typ:?}")
```
aedm: Typo: `git` -> `got`

emilk:
```suggestion
                        panic!("{fqname}: 'binary' can only be used on '[ubyte]', got {typ:?}")
```

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

PR #10876, opened 2025-08-12
Add CI tests for gRPC calls to OSS server

timsaucer:
This PR adds CI tests for the interface layer between rerun datafusion queries and the rerun server. It uses the in-memory simple server with generated data. The data files for this were generated with the following code:

```python
from datetime import datetime, timedelta
import random
import rerun as rr

# We want to create data that has three timelines that are not in order
# but when we sort according to each timeline, then the data in the
# columns with the similar names are in order


def maybe_val(val):
    if random.random() > 0.2:
        return val
    else:
        return None

n_rows = 10

def generate_data(filename, n_rows):
    base_time = datetime(2024, 1, 1)
    timestamps = [maybe_val(base_time + timedelta(minutes=i*2)) for i in range(n_rows)]
    
    # Generate durations in minutes
    durations = [maybe_val(timedelta(minutes=30 + i*5)) for i in range(n_rows)]
    
    # Generate sequence numbers
    sequence_numbers = list(maybe_val(x) for x in range(1, n_rows + 1))
    
    obj_x1 = list(maybe_val(float(x)) for x in range(1, n_rows + 1))
    obj_x2 = list(maybe_val(float(x)) for x in range(1, n_rows + 1))
    obj_x3 = list(maybe_val(float(x)) for x in range(1, n_rows + 1))
    
    obj1_indices = list(range(n_rows))
    random.shuffle(obj1_indices)
    
    obj2_indices = list(range(n_rows))
    random.shuffle(obj2_indices)
    
    obj3_indices = list(range(n_rows))
    random.shuffle(obj3_indices)

    rr.init(filename, spawn=True)
    rr.save(f"{filename}.rrd")

    rr.log("/text1", rr.TextDocument("Before text"), static=True)

    for idx in range(0, n_rows):

        rr.reset_time()

        timestamp = timestamps[obj1_indices[idx]]
        duration = durations[obj2_indices[idx]]
        sequence = sequence_numbers[obj3_indices[idx]]

        obj1_pos = obj_x1[obj1_indices[idx]]
        obj2_pos = obj_x2[obj2_indices[idx]]
        obj3_pos = obj_x3[obj3_indices[idx]]

        if timestamp is not None:
            rr.set_time("time_1", timestamp=timestamp)

        if duration is not None:
            rr.set_time("time_2", duration=duration)

        if sequence is not None:
            rr.set_time("time_3", sequence=sequence)
        
        if obj1_pos is not None:
            rr.log("/obj1", rr.Points3D([[obj1_pos, 0.0, 0.0]]))

        if obj2_pos is not None:
            rr.log("/obj2", rr.Points3D([[obj2_pos, 1.0, 0.0]]))
        
        if obj3_pos is not None:
            rr.log("/obj3", rr.Points3D([[obj3_pos, 2.0, 0.0]]))

    rr.log("/text2", rr.TextDocument("After text"), static=True)

generate_data("file1", 25)
generate_data("file2", 50)
```

## TODO
* [x] `@rerun-bot full-check`


timsaucer:
@rerun-bot full-check

abey79 (commented):
I haven't yet looked at this properly, but my knee jerk reaction is: could we factor the server as a pytest fixture and just use our existing test setup? It's actually a rhetoric question. I'm pretty sure we can, the question is at what cost?

timsaucer:
> I haven't yet looked at this properly, but my knee jerk reaction is: could we factor the server setup stuff as a pytest fixture and just use our existing test setup? It's actually a rhetoric question. I'm pretty sure we can, the question is at what cost?

That was my first thought as well, but I needed some extra infra in place to start the server up. I do think I can switch over a portion of it to pytest, but I'll still need these bits to start the server and shut it down. I basically copied what existed for the e2e tests and adjusted it for my use case.

abey79:
> I'll still need these bits to start the server and shut it down

As far as I can tell, this is Python code right? So it could be factored as a fixture: https://docs.pytest.org/en/stable/explanation/fixtures.html

Or am I missing something?

emilk:
I just found this PR - there is a lot of overlap between this and https://linear.app/rerun/project/server-viewer-test-089c611008e2/overview

* https://github.com/rerun-io/rerun/pull/10953
* https://github.com/rerun-io/rerun/pull/10956
* …and others

emilk:
@rerun-bot full-check

abey79:
> I just found this PR - there is a lot of overlap between this and [linear.app/rerun/project/server-viewer-test-089c611008e2/overview](https://linear.app/rerun/project/server-viewer-test-089c611008e2/overview)
> 
> * [Add e2e integration test using memory server #10953](https://github.com/rerun-io/rerun/pull/10953)
> * [Adds UI-based E2E tests #10956](https://github.com/rerun-io/rerun/pull/10956)
> * …and others

Yeah and this one that I recently wrote and captures a whole bunch of requirements that the redap folks have:

https://linear.app/rerun/project/harness-for-sdk-server-e2e-tests-and-benchmarks-13c52dc19a87/overview

emilk:
Screenshot 2025-08-22 at 16 29 43

timsaucer:
Moving to draft while I continue to debug why CI is failing. All tests pass locally.

timsaucer:
Ok, the minimal wheel we're building doesn't include the server. I'm thinking about if I want to keep down the route I have in this PR or do it like in #10956

emilk:
Is this part of some existing Linear issue?

timsaucer:
> Is this part of some existing Linear issue?

Yes. I added a link but it's in the DPF workspace so I don't think it's connecting properly here.

tests/assets/rrd/dataset/file1.rrd
abey79, line 1: In the event I would like to modify this file (because we somehow break our migration code, or because I want to improve the test), how would I go about it?

I don't have a good answer to that. If we check the generation code in, we need to maintain it and make sure the RRD is in sync. If we generate the RRD on the fly, it makes the harness more complicated. Maybe we should leave a "paper trail" that we accept might rot, but might still be useful in the future (e.g. a link to a gist?).

timsaucer: Very good point. Instead of a gist which is associated with a single user rather than organization, I added a readme to the folder that contains the files. The server ignores this file when scanning the directory, so it seems like a good place where someone would look.

PR #10881, opened 2025-08-13
Improve `rerun rrd print`

emilk:
### What
This adds some new options to `rerun rrd print`:

- `--full-metadata`: If true, includes `rerun.` prefixes on keys
- `--transposed`: print the record batches transposed
- `--entity`: print only chunks pertaining to a specific entity

I tried to also add support for printing non-migrated chunks, but that is blocked on https://github.com/rerun-io/rerun/issues/10343

### DIFF
--- a/crates/top/rerun/src/commands/rrd/print.rs
+++ b/crates/top/rerun/src/commands/rrd/print.rs
@@ -1,8 +1,10 @@
 use anyhow::Context as _;
+use arrow::array::RecordBatch;
 use itertools::Itertools as _;
 
 use re_byte_size::SizeBytes as _;
 use re_log_types::{LogMsg, SetStoreInfo};
+use re_sdk::EntityPath;
 
 use crate::commands::read_rrd_streams_from_file_or_stdin;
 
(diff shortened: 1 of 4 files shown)


grtlr:
Potential merge race (CI failures) with:

* https://github.com/rerun-io/rerun/pull/10864

crates/top/rerun/src/commands/rrd/print.rs
```diff
@@ -27,27 +29,72 @@ pub struct PrintCommand {
     #[clap(long, short, action = clap::ArgAction::Count)]
     verbose: u8,
 
+    // NOTE: we use a hack to allow specifying `=false` or `=true` in CLI. See https://github.com/clap-rs/clap/issues/1649#issuecomment-2144932113
```
grtlr, line 32:
```suggestion
    // TODO(clap-rs/clap#1649): we use a hack to allow specifying `=false` or `=true` in CLI. See https://github.com/clap-rs/clap/issues/1649#issuecomment-2144932113
```

This should be picked up by our zombie todo script by now.

crates/top/rerun/src/commands/rrd/print.rs
```diff
@@ -70,58 +117,114 @@ impl PrintCommand {
...
+                    migrared_chunk.num_rows(),
+                    re_format::format_bytes(migrared_chunk.total_size_bytes() as _),
+                    migrared_chunk.entity_path(),
                 );
+
+                match options.verbose {
+                    0 => {
+                        let column_names = migrared_chunk
+                            .component_columns()
+                            .map(|(descr, _)| descr.column_name(re_sorbet::BatchType::Chunk)) // short coulmn name without entity-path prefix
```
grtlr:
```suggestion
                            .map(|(descr, _)| descr.column_name(re_sorbet::BatchType::Chunk)) // short column name without entity-path prefix
```

crates/top/rerun/src/commands/rrd/print.rs
emilk: The clap issue is closed, but there is no nice way other than the workaround

grtlr: Hmm, that's too bad.

PR #10882, opened 2025-08-13
Improve formatting of arrow datatypes: include nullability

emilk:
This improves our printing of arrow datatypes to also include the nullability of lists etc.


crates/store/re_dataframe/src/snapshots/re_dataframe__query__tests__view_contents-2.snap
```diff
@@ -1,12 +1,11 @@
 ---
 source: crates/store/re_dataframe/src/query.rs
 expression: DisplayRB(dataframe)
-snapshot_kind: text
 ---
 ┌──────────────────────┬──────────────────────┬────────────────────────────────────┬────────────────────────────────────┐
 │ frame_nr             ┆ log_time             ┆ /this/that:example.MyPoints:colors ┆ /this/that:example.MyPoints:labels │
 │ ---                  ┆ ---                  ┆ ---                                ┆ ---                                │
-│ type: i64            ┆ type: Timestamp(ns)  ┆ type: List[u32]                    ┆ type: List[Utf8]                   │
+│ type: i64            ┆ type: Timestamp(ns)  ┆ type: List[nullable u32]           ┆ type: List[nullable Utf8]          │
```
emilk, line 8: …alternatively we print `nonullable` when its not nullable, and omit it when it is 🤔

Wumpf: well, which one is the odd one out? 🤔 
To confirm understanding: if the _schema_ says that the field is nullable, the list is a allowed to have a validity map, otherwise it isn't? Bit confused about the implications atm

PR #10891, opened 2025-08-13
Move in-memory server into rerun subcommand

timsaucer:
This PR moves over `rerun_server` to be a subcommand within the main `rerun` binary. No other changes are included.

You can run it with this command:

```shell
pixi run rerun server -d /path/to/dataset
```

But you need to enable the `oss_feature` flag first.

## TODO
* [x] `@rerun-bot full-check`


timsaucer:
@rerun-bot full-check

crates/top/rerun/src/commands/entrypoint.rs
```diff
@@ -562,6 +562,10 @@ enum Command {
     #[cfg(feature = "auth")]
     #[command(subcommand)]
     Auth(AuthCommands),
+
+    /// In memory rerun data server
```
grtlr:
```suggestion
    /// In-memory Rerun data server
```

grtlr: `pixi run man`

crates/store/re_server/README.md
```diff
@@ -8,9 +8,3 @@ Part of the [`rerun`](https://github.com/rerun-io/rerun) family of crates.
 ![Apache](https://img.shields.io/badge/license-Apache-blue.svg)
 
 In-memory opensource implementation of the Redap server.
```
grtlr:
```suggestion
In-memory opensource implementation of the Rerun server.
```

crates/top/rerun/src/commands/entrypoint.rs
```diff
@@ -562,6 +562,10 @@ enum Command {
...
+    #[command(name = "server")]
```
grtlr, line 565: @emilk Do you think we need to put the server behind a feature flag too (like we do with `data_loaders`)?

grtlr (approved):
Apart from the feature flag question and minor comments LGTM.

So nice to finally have this!

docs/content/reference/cli.md
```diff
@@ -22,6 +22,7 @@ The Rerun command-line interface:
 * `reset`: Reset the memory of the Rerun Viewer.
 * `man`: Generates the Rerun CLI manual (markdown).
 * `auth`: Authentication with the redap.
+* `server`: In memory rerun data server.
```
emilk: We should sort this list alphabetically. Having them chronologically ordered by the time they were added makes no sense for a user

crates/top/rerun/src/commands/entrypoint.rs
emilk: Yes, at least assuming it adds a non-negible amount of dependencies/compilation time.

`rerun` is the entry-point for just using Rerun for logging, and we want to try to keep that lean.

ARCHITECTURE.md
```diff
@@ -179,6 +178,7 @@ Update instructions:
 | re_log_encoding | Helpers for encoding and transporting Rerun log messages        |
 | re_protos       | Rerun remote store gRPC API types                               |
 | re_query        | Querying data in the re_chunk_store                             |
+| re_server       | In-memory data server                                           |
```
emilk, line 181: Need to update the png (see above in ARCHITECTURE.md)

crates/store/re_server/README.md
emilk: We need to have some expectation management in this Readme!

* What is it good for?
* What isn't it good for?
* How is this different from running `rerun --serve-grpc`?
* What will be included in this crate eventually?
* What will _never_ be included in this crate?
* Where should people go if they want more?

crates/top/rerun/src/commands/entrypoint.rs
emilk: `scripts/ci/count_dependencies.py -p rerun --no-default-features` 262 -> 296 for this PR, so yes - let's add a feature flag!

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

timsaucer:
Thank you for pulling this over the line while I was out!

PR #10892, opened 2025-08-13
Improve token validation and forget bad tokens

abey79:
### What

This PR improves the token validation workflow as follows:
- calls into `GetVersion` upon establishing a new connection
- checks the result to explicitly single out missing token and bad token failure modes
- in case of bad token, forget about that token if it was saved

This should (partially) address the failure mode where a saved token is no longer valid (though it still requires the user to "refresh" once after the token is forgotten).

Note: calling into `GetVersion` is required because unauthenticated errors are not triggered by "just" establishing the connection


crates/store/re_grpc_client/src/connection_registry.rs
```diff
@@ -127,14 +160,59 @@ impl ConnectionRegistryHandle {
...
 
-        let client = crate::redap::client(origin.clone(), token).await;
-        match client {
+        let client = crate::redap::client(origin.clone(), token.clone()).await;
+        let mut client = match client {
             Ok(client) => {
                 let mut inner = self.inner.write().await;
-                inner.clients.insert(origin, client.clone());
-                Ok(ConnectionClient::new(client))
+                inner.clients.insert(origin.clone(), client.clone());
```
emilk: It's probably worth checking the return value of this (i.e. does this origin already exist in `inner.clients`?)

crates/store/re_grpc_client/src/connection_registry.rs
```diff
@@ -48,6 +49,38 @@ impl ConnectionRegistry {
...
+    Tonic(#[from] tonic::transport::Error),
+
+    #[error("server is expecting an unencrypted connection (try `rerun+http://` if you are sure)")]
+    UnencryptedServer,
+
+    #[error("the server requires an authentication token, but none was provided: {0}")]
+    UnauthenticatedMissingToken(tonic::Status),
+
+    #[error("the server rejected the provided authentication token: {0}")]
+    UnauthenticatedBadToken(tonic::Status),
```
emilk: Tonic has super-verbose error messages. We have a wrapper to fix that though: https://github.com/rerun-io/rerun/blob/045ca0e7f6344f63ab2021f3e8b26186dd364c61/crates/store/re_grpc_client/src/lib.rs#L20-L65

We should probably upstream those improvements at some point, but let's use the wrapper until then

crates/store/re_grpc_client/src/connection_registry.rs
abey79: I'm not sure what this matters? At this point of the code, I already decided to create a new client, so I don't really care if one pre-existed. Currently, that decision is simplistic: if none existed, create one. So it would be a "bug" if there was a client in the map. In the future, that might happen though, for example when/if we cycle through short-lived token. In either case, I don't care much about the old client, or if it exists at all.

crates/store/re_grpc_client/src/connection_registry.rs
emilk: * https://github.com/hyperium/tonic/pull/2403

crates/store/re_grpc_client/src/connection_registry.rs
```diff
@@ -48,6 +50,38 @@ impl ConnectionRegistry {
     }
 }
 
+/// Possible errors when creating a connection.
+#[derive(Debug, thiserror::Error)]
+pub enum ClientConnectionError {
+    /// Native connection error
+    #[cfg(not(target_arch = "wasm32"))]
+    #[error("Connection error: {0}")]
+    Tonic(#[from] tonic::transport::Error),
```
emilk, line 62: Have you checked the formatting of this one?

crates/store/re_grpc_client/src/connection_registry.rs
emilk: So it can happen if two threads calls the parent function at the same time, right (with the same `origin`)? In that case we could have two clients be created at the same time, the first get added to the `inner.clients`, and then later the second one replacing it.

Perhaps that is exactly what we want, but if so I think it's worth a comment to that effect.

Alternatively, wouldn't it be better if the second thread blocked waiting for the first thread, so that only one client was ever created?

crates/store/re_grpc_client/src/connection_registry.rs
```diff
@@ -123,14 +157,61 @@ impl ConnectionRegistryHandle {
...
+                return Err(err.into());
             }
-            Err(err) => Err(err),
+        };
+
+        // Call the version endpoint to check that authentication is successful. It's ok to do this
+        // since we're caching the client, so we're not spamming such request unnecessarily.
+        let request_result = client
+            .inner()
+            .version(re_protos::frontend::v1alpha1::VersionRequest {})
```
emilk: Discussed on slack: the version endpoint doesn't actually check for authentication. We need a `/whoami` endpoint for that:

https://rerunio.slack.com/archives/C07T0LFT5BQ/p1756219199757459

emilk (approved):

crates/store/re_grpc_client/src/connection_registry.rs
abey79: Indeed. For this PR, I'm going to use another end point with a todo to switch to `whoami` when it's available

crates/store/re_grpc_client/src/connection_registry.rs
abey79: You're right. Acquiring the lock before creating the client seems to be the right thing to do 👍🏻

crates/store/re_grpc_client/src/connection_registry.rs
abey79: No, but it appears to be not too crazy.

Image

crates/store/re_grpc_client/src/connection_registry.rs
abey79: Actually, I'm changing this function so we cache the client _only_ if the auth test passes (no need to keep a client around that is not authenticated—also this caused UI issues). So now I'm acquiring the lock a single time for the entire function duration, so concurrent call ends up either:
- second thread using the happy path
- both threads trying and failing to produce a properly authenticated client

abey79: Actually actually, this might be a bad idea. Andreas specifically introduced an intermediate lock release because the client creation process is long-ish. We might have to accept that two thread concurrently create and test a client, and the latest "wins" at caching it.

abey79:
@emilk I completely rewrote the client creation/caching/token mgmt logic. We had a good look with Andreas, but maybe you want to have a look too.

crates/store/re_grpc_client/src/connection_registry.rs
abey79: I ended up rewriting the whole thing. There _is_ a possibility for two thread to race and create a client twice, but this should have no significant impact.

PR #10893, opened 2025-08-13
Put `rerun mcap` behind the `data_loaders` feature flag

grtlr:
This is required for the crates to compile individually with minimal feature sets.


Wumpf (approved):
I wonder why we have this feature flag in the first place. Some data loader that's too heavy to include it by default? 🤔

Wumpf:
image

noted.

grtlr:
> I wonder why we have this feature flag in the first place. Some data loader that's too heavy to include it by default? 🤔

I have no idea, and I already want to get rid of it.

Wumpf:
all things equal I'm encouraging you (or others) to try in a separate iteration!

Issue #10894, opened 2025-08-13
`.df().count()` always returns 0


jleibs:
Example (using droid:sample50)

```
dataset.dataframe_query_view(index="real_time",
                             contents="/action/gripper_position").df().count()

-> 0r
```

However:

```
dataset.dataframe_query_view(index="real_time",
                             contents="/action/gripper_position").df().to_arrow_table().num_rows

-> 13894
```

timsaucer:
Since count_star passes through zero column selection we're getting all data filtered out. I think I'll need to special case this somehow.

timsaucer:
This is evident in that the logical plan shows no projection to the table provider. If you run an aggregate with *any* selection it does work. `df.aggregate([col("rerun_partition_id")], f.count())`

timsaucer:
The bug is corrected, but we are leaving his issue open until we have integration tests in place.

pweids:
@timsaucer are we still experiencing this?

timsaucer:
I thought this was answered in the last comment, but the problem is corrected. I don't want to close the issue until we have test coverage in place. I expect to merge our test harness this morning as soon as CI finishes, so I should be able to add this test today.

PR #10895, opened 2025-08-13
Fix bug with df.count()

timsaucer:
This addresses the issue in  https://github.com/rerun-io/rerun/issues/10894

Do not close that issue until test coverage is complete.

This addresses the corner case where no columns are selected, such as doing count_star with no other operations.


crates/store/re_datafusion/src/dataframe_query_common.rs
```diff
@@ -406,9 +407,10 @@ pub fn align_record_batch_to_schema(
         }
     }
 
-    Ok(RecordBatch::try_new(
```
emilk, line 409: Let's ban `RecordBatch::try_new(` in `clippy.toml` to force us to fix this corner-case everywhere

timsaucer:
@rerun-bot full-check

emilk (approved):

PR #10896, opened 2025-08-14
Unify MCAP parsing into `re_mcap`

grtlr:
### What

This moves most of the MCAP-related stuff into a new `re_mcap` crate that supersedes `re_mcap_ros2`. Most importantly, it hides almost all implementation details and only makes `pub` the necessary types and traits for users to implement their own `Layer`s.

> [!IMPORTANT]
> We still need to figure out how users can customize the `LayerRegistry` from their code.

### Additional

* Renamed `McapMessageParser` to `MessageParser`
* Streamlined some of the error handling

## TODO
* [x] `@rerun-bot full-check`

Issue #10879, opened 2025-08-12
Dataframe Column Names Don't Consider Archetypes Leading to Collisions


ntjohnson1:
**Describe the bug**
Related to what caused me to see https://github.com/rerun-io/rerun/issues/10877 but when I tried to repro something closer to my real use case it felt like two separate bugs.

If I have this in an rrd the dataframe api cannot differentiate the two `data` entries. Even when one is static and the other isn't
<img width="185" height="260" alt="Image" src="https://github.com/user-attachments/assets/408fe880-7a2e-4578-a0a1-6cd5eb85f216" />

**To Reproduce**
```python
import rerun as rr

rr.init("duplicate_tagged_components", spawn=False)
rr.save("sample_any.rrd")
# These are two groupings of content with a variety of fields but at least one intersects
schema = rr.AnyValues().with_field(rr.ComponentDescriptor("data", "mcap.Schema"), 1.0)
image = rr.AnyValues().with_field(rr.ComponentDescriptor("data", "SomeCustomImage"), 2.0)
rr.log("/camera/data", schema, static=True)
rr.log("/camera/data", image)
```

```python
>>> import rerun as rr
>>> rec = rr.dataframe.load_recording("sample_any.rrd")
>>> rec.schema()
Column name: /camera/data:data
        Entity path: /camera/data
        Archetype: SomeCustomImage
        Component type: 
        Component: data
Column name: property:RecordingInfo:start_time
        Entity path: /__properties
        Archetype: rerun.archetypes.RecordingInfo
        Component type: rerun.components.Timestamp
        Component: RecordingInfo:start_time
>>>batches = rec.view(index=None, contents="/camera/**").select()
print(next(iter(batches)))
pyarrow.RecordBatch
log_tick: int64
log_time: timestamp[ns]
/camera/data:data: list<item: double>
  child 0, item: double
----
log_tick: [null]
log_time: [null]
/camera/data:data: [[1]]

>>>batches = rec.view(index="log_time", contents="/camera/**").select()
<stdin>:1: RerunWarning: RecordingView::select: tried to select static data, but no non-static contents generated an index value on this timeline. No results will be returned. Either include non-static data or consider using `select_static()` instead.
```

**Expected behavior**
Internally we disambiguate these. So we should be able to expose that to the dataframe API.
Note that this has a python example but we see the same bug if we use a similar behavior in rust.

**Rerun version**
[...]

PR #10900, opened 2025-08-14
Grtlr/properly tagged descriptors

grtlr:
### Related

* Surfaced by #10879.
* [ ] Merge after #10896.

### What

This PR does two things:

* Only emit unknown schema warnings for `ros2msg` for messages that have a `ros2msg` schema.
* Make all of the `ComponentIdentifier`s unique, by following the Rerun-builtin archetype conventions (i.e. a `ComponentIdentifier` is constructed as follows: `<ArchetypeName>:<field>`).

The latter is somewhat of a workaround until we have better APIs for tagging dynamic archetypes.


crates/utils/re_mcap/src/layers/ros2.rs
```diff
@@ -28,15 +28,19 @@ impl MessageLayer for McapRos2Layer {
...
             re_log::warn_once!(
                 "Encountered ROS2 message without schema in channel {:?}",
                 channel.topic
             );
             return None;
         };
 
-        Some(match name {
+        if schema.encoding.as_str() != "ros2msg" {
+            return None;
```
Wumpf, line 40: should there be a warning for unsupported encoding or did that already happen elsewhere?

grtlr: Yeah, the warning happens one level above this because the individual layers can't know which other layers are active / supported.

PR #10905, opened 2025-08-14
Remove `debug_assert` for `field.name()` in Arrow formatting

grtlr:
### What

This `debug_assert_eq` would trigger when encountering `MapArray`s (which are totally fine in this context):

```sh
# (Triggered by hovering over a Map field).
thread 'main' panicked at 'assertion `left == right` failed
  left: "entries"
 right: "item"'
re_arrow_util/src/format_data_type.rs:130
```


Wumpf (approved):
hmm I wonder where this came from, probably someone (maybe even me) thought that all inner fields are always called `item` and wanted to make sure of that

PR #10907, opened 2025-08-14
Reduce impact of data spill on dataframe queries

timsaucer:
### Related

This is a temporary work around for https://github.com/rerun-io/dataplatform/issues/1494

### What

When the record batch stream is exhausted, clear out all member variables by using an inner struct. This will at least reduce the impact of memory leakage from the above mentioned issue.

### DIFF
--- a/crates/store/re_datafusion/src/dataframe_query_provider.rs
+++ b/crates/store/re_datafusion/src/dataframe_query_provider.rs
@@ -66,7 +66,7 @@ pub(crate) struct PartitionStreamExec {
 
 type ChunksWithPartition = Vec<(Chunk, Option<String>)>;
 
-pub struct DataframePartitionStream {
+pub struct DataframePartitionStreamInner {
     projected_schema: SchemaRef,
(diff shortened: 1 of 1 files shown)


crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -127,15 +137,25 @@ impl Stream for DataframePartitionStream {
...
 
-        this.store_output_channel
+        let result = this
+            .store_output_channel
             .poll_recv(cx)
-            .map(|result| Ok(result).transpose())
+            .map(|result| Ok(result).transpose());
+
+        if let Poll::Ready(None) = &result {
+            this_outer.inner = None;
```
Wumpf, line 151: maybe that's overzealous, but imho it's worth pointing out in the comments that this is the core of the mem-leak-reduction workaround. Almost missed it and was wondering where the fix is

Wumpf (approved):
lgtm!

PR #10917, opened 2025-08-14
Deprecate AnyValues With Archetype

ntjohnson1:
### Related
Closes #https://github.com/rerun-io/rerun/issues/10879

### What
This actually takes two different approaches here we might want to wait to resolve this until we discuss what we want from anyvalues/dynamic archetype builder
* For python deprecates `with_field` since that is the only way to include the archetype
* For rust deprecates new since that is how we include the archetype

[This](https://github.com/rerun-io/rerun/pull/10900) changes the formatting which again sort of resolves the issue.

### DIFF
--- a/crates/store/re_types/src/any_values.rs
+++ b/crates/store/re_types/src/any_values.rs
@@ -15,7 +15,12 @@ pub struct AnyValues {
 }
 
 impl AnyValues {
+    //TODO(#10879): Prune this method in 0.26.0
     /// Assigns an (archetype) name to this set of any values.
+    #[deprecated(
+        since = "0.24.0",
+        note = "Use of archetype leads to subtle errors. Use default instead."
+    )]
     #[inline]
     pub fn new(archetype_name: impl Into<ArchetypeName>) -> Self {
         Self {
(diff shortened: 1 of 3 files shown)


crates/store/re_types/src/any_values.rs
```diff
@@ -15,7 +15,12 @@ pub struct AnyValues {
 }
 
 impl AnyValues {
+    //TODO(#10879): Prune this method in 0.26.0
     /// Assigns an (archetype) name to this set of any values.
+    #[deprecated(
+        since = "0.24.0",
+        note = "Use of archetype leads to subtle errors. Use default instead."
```
emilk, line 22: What subtle errors? Can we explain it to the user, e.g. by linking to the migration guide?

We should also help the user with what to do instead. Are we removing support for grouping anyvalues by archetype name? What should they use instead if they want to group custom values?

ntjohnson1: I should have marked this do-not-merge. This was mostly to spark discussion around different approaches here. More context in the issue. I'm closing this for now and will provide more context into the real solution.

PR #10918, opened 2025-08-14
Updated `.rrd` compatibility note

YohDeadfall:
### What

There's [a blog post](https://rerun.io/blog/release-0.23) by Emil where he states that Rerun promises very limited version compatibility for opening `*.rrd` files:

> There are some limitations. As a first step, we’re only promising that Rerun can load the data from the previous Rerun version. This means you will be able to open an 0.23 .rrd file in Rerun 0.24, but maybe not in Rerun 0.31.

Therefore, it's worth mentioning in the architecture overview.


ARCHITECTURE.md
```diff
@@ -41,7 +41,7 @@ The web viewer can load `.rrd` files (just drag-drop them into the browser), or
 
 `.rrd` ("**R**e**r**un **D**ata") is just a bunch of log messages appended one after the other to a file.
 
-NOTE: `.rrd` files do not yet guarantee any backwards or forwards compatibility. One version of Rerun will likely not be able to open an `.rrd` file generated by another Rerun version.
+NOTE: `.rrd` files do not yet guarantee any backwards or forwards compatibility. One version of Rerun is able to open an `.rrd` file generated by *previous* Rerun version, but it will likely not work for older versions.
```
grtlr:
```suggestion
NOTE: `.rrd` files do not yet offer full backwards or forwards compatibility. However, the current version of Rerun will always be able to open `.rrd` files generated by the *previous* version. This is not guaranteed to work for files generated with older versions.
```

grtlr (commented):
Nice catch, and thank you for pointing this out!

Here are some minor suggestions.

PR #10920, opened 2025-08-14
Fixed highlighting of Python code in docs

YohDeadfall:
### What

Unfortunately, I don't know how your site is compiled, so cannot change it to treat `py` as `python` as it happens on GitHub in markdown. Therefore, I changed all occurrences of `py` to be `python` for proper code rendering.


abey79 (approved):
Thanks! Not sure how we manage to overlook that for so long 😅

PR #10930, opened 2025-08-18
Add Claude.md

bllchmbrs:
Add `claude.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,144 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## About Rerun
+
+Rerun is a multimodal data stack and visualization platform for robotics, spatial AI, and computer vision. It provides SDKs for Python, Rust, and C++ to log time-aware data like images, point clouds, tensors, and text, with a native/web viewer for visualization and debugging.
+
+## Essential Development Commands
+
+### Building and Running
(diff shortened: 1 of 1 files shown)


emilk:
There are some lint errors you can fix with `scripts/lint.py --fix`.

You should also configure your editor to automatically trim trailing whitespace :]

CLAUDE.md
emilk, line 1: Does claude code look for this file auotmatically?

Is there some other name we can give it that makes it usable for other LLMs?

Can we prefix the name with `.` or move it into a sub-repo in order to de-emphasise it for humans?

Wumpf:
would it make sense to combine this with our `building.md` and give that to llms instead? seems a bit nasty to have all this stuff in several places.
(Even better would ofc to feed it pixi.toml, but I guess that's way too raw)

abey79:
I wonder if we should add something in `CONTRIBUTING.md` about AI slop PRs. Maybe that effort would be vain?

Wumpf:
Practically all we can write is that everybody should make sure quality & functionallity is up to speed and that they tested the thing.
We don't really care if the whole thing was generated and I don't think a disclaimer helps all the much. The problems only arise when people create blind PRs. And we haven't seen _much_ of that _yet_.

More interesting to advise AIs in files like claude.md to read our contributing guide, heh.

Either way @bllchmbrs do you still want to go forward with this? As-is I think this file is too brittle and too much maintenance, but if we want to add something to codegen or just simply direct claude/other-llms to our existing docs I'm all ears

PR #10932, opened 2025-08-18
Fix doc link not clickable in view help

abey79:
The "Docs" link text could not be clicked because it's selectable. This PR fixes this.


crates/viewer/re_ui/src/help.rs
```diff
@@ -157,7 +157,10 @@ impl Help {
                                     ui.visuals().widgets.inactive.text_color()
                                 };
 
-                                ui.label(RichText::new("Docs").color(tint).size(11.0));
+                                ui.add(
+                                    egui::Label::new(RichText::new("Docs").color(tint).size(11.0))
+                                        .selectable(false),
+                                );
 
                                 ui.small_icon(&icons::EXTERNAL_LINK, Some(tint));
```
emilk: We should use atomics for all this instead

emilk:
:atom:

PR #10933, opened 2025-08-18
Add `re_workspace_hack` to avoid rebuilding dependencies

emilk:
## Problem
We have many crates.
They depend on 3rd party crates, with different combination of feature flags.
Depending on what target we build for (which binary, or test, etc) different feature flags gets enabled for those 3rd party crates. If they are not EXACTLY THE SAME as THE LAST TIME you compiled, they will be recompiled from scratch (I'm looking at you, `datafusion`).

## The solution
Enter `re_workspace_hack`:
* It depends on ALL our 3rd party crates with all features we need _anywhere_ enabled
* All our crates depend on `re_workspace_hack`
* We remove the dependencies on `re_workspace_hack` just before `cargo publish`

Maintaining this would be hell… except we have https://crates.io/crates/cargo-hakari

## Related
* Part of https://linear.app/rerun/issue/RR-2480/speed-up-buildsrebuilds

## TODO
* [ ] Disable `re_workspace_hack` in `rust_checks.py`
* [ ] `@rerun-bot full-check`


Wumpf:
archiving this.

PR #10938, opened 2025-08-19
Add keyboard shortcut to copy entity hierarchy

oxkitsune:
### What

Adds a keyboard shortcut (<kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>E</kbd>) to copy the entity hierarchy including component types  to the user's clipboard.

### DIFF
--- a/crates/store/re_entity_db/src/entity_db.rs
+++ b/crates/store/re_entity_db/src/entity_db.rs
@@ -1026,13 +1026,10 @@ mod tests {
             db.add_chunk(&Arc::new(chunk))?;
         }
 
-        // Test formatting
-        let formatted = db.format_with_components();
-        assert!(formatted.contains("parent"));
-        assert!(formatted.contains("parent/child1"));
-        assert!(formatted.contains("parent/child1/grandchild"));
-
-        assert!(formatted.contains("example.MyPoint: Struct[2]"));
+        assert_eq!(
+            db.format_with_components(),
+            "/parent\n  /parent/child1\n    /parent/child1/grandchild\n      example.MyPoint: Struct[2]\n"
+        );
 
         Ok(())
     }


crates/viewer/re_viewer/src/app.rs
```diff
@@ -1589,6 +1593,97 @@ impl App {
             .success(format!("Copied {url:?} to clipboard"));
     }
 
+    fn copy_entity_hierarchy_to_clipboard(
```
Wumpf, line 1637: let's not dump all of this into app.rs and split this up instead. The entity tree itself could own `format_entity_tree` which can also be unit tested there then

crates/viewer/re_ui/src/command.rs
```diff
@@ -293,6 +295,11 @@ impl UICommand {
                 "Copy a link to the part of the active recording within the loop selection bounds.",
             ),
 
+            Self::CopyEntityHierarchy => (
+                "Copy entity hierarchy",
+                "Copy the complete entity hierarchy tree to the clipboard",
```
Wumpf:
```suggestion
                "Copy the complete entity hierarchy tree of the currently active recording to the clipboard",
```

Wumpf (changes requested):

crates/viewer/re_viewer/src/app.rs
Wumpf: this then can also be part of the dbg output of the entity tree

crates/store/re_entity_db/src/entity_db.rs
```diff
@@ -939,3 +986,54 @@ impl re_byte_size::SizeBytes for EntityDb {
...
+
+        // Test formatting
+        let formatted = db.format_with_components();
+        assert!(formatted.contains("parent"));
+        assert!(formatted.contains("parent/child1"));
+        assert!(formatted.contains("parent/child1/grandchild"));
+
+        assert!(formatted.contains("example.MyPoint: Struct[2]"));
+
+        Ok(())
```
Wumpf: why not pin the entire string and check it with a single assert? 🤷

PR #10940, opened 2025-08-19
Add snapshot test for full app's setting screen

MichaelGrupp:
Spins up the whole app and navigates to the settings screen.

First step towards more full app integration tests.


crates/utils/re_capabilities/src/main_thread_token.rs
```diff
@@ -45,6 +45,13 @@ impl MainThreadToken {
         }
     }
 
+    /// Only call this from tests, where the main thread would have a different name than "main".
```
lucasmerlin: I think the comment is wrong, tests usually aren't run on the main thread (unless you are using nextest I believe)

crates/viewer/re_viewer/src/ui/rerun_menu.rs
```diff
@@ -32,7 +32,8 @@ impl App {
         let image = re_ui::icons::RERUN_MENU
             .as_image()
             .max_height(desired_icon_height)
-            .tint(ui.tokens().strong_fg_color);
+            .tint(ui.tokens().strong_fg_color)
+            .alt_text("menu");
```
lucasmerlin: Nit: It's not a visible string but it should still match the casing of our other Ui strings (so "Menu")

MichaelGrupp:
...almost sweet... need to get rid of this in the Linux CI:

Bildschirmfoto 2025-08-19 um 20 31 50

Bildschirmfoto 2025-08-19 um 20 28 36

MichaelGrupp:
Should be fixed now™️ with the latest commit, can you retrigger please? :

PR #10953, opened 2025-08-20
Add e2e integration test using memory server

emilk:
This adds a e2e test using the full `rerun server` command and the python api. 

We register an example recording, and query it using rerun cli. Then we insta snapshot the output.

This could be extended to do a full ui e2e test using the viewer
 - see #10956


tests/rust/re_integration_test/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+pub struct TestServer {
+    server: Child,
+}
+
+impl TestServer {
+    pub fn spawn() -> Self {
+        let mut build = Command::new("pixi");
+        build.args(["run", "rerun", "--help"]);
+        build.stdout(Stdio::null());
+        build.spawn().unwrap().wait_for_success();
```
emilk: Probably more clear:

```suggestion
        // First build the binary:
        let mut build = Command::new("pixi");
        build.args(["run", "rerun-build"]);
        build.stdout(Stdio::null());
        build.spawn().unwrap().wait_for_success();
```

Though we need to add the `oss_server` feature to https://github.com/rerun-io/rerun/blob/1055e0026c71014a5d28c8092a7b2d0c5329cdda/pixi.toml#L186

tests/rust/re_integration_test/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+            .spawn()
+            .expect("Failed to start pixi")
+            .wait_for_success();
+
+        let mut server = Command::new("../../../target_pixi/debug/rerun");
+        server.args(["server", "--port", &port.to_string()]);
+        let server = server.spawn().expect("Failed to start rerun server");
+
+        let mut success = false;
+        for _ in 0..1000 {
```
lucasmerlin:
```suggestion
        for _ in 0..50 {
```

PR #10954, opened 2025-08-20
Make viewer test utils reusable

MichaelGrupp:
Introduce `re_viewer::viewer_test_utils` to be usable by other crates (behind `testing` feature flag).

### DIFF
--- a/crates/viewer/re_viewer/src/lib.rs
+++ b/crates/viewer/re_viewer/src/lib.rs
@@ -24,6 +24,7 @@ mod ui;
 mod viewer_analytics;
 
 #[cfg(feature = "testing")]
+#[cfg(not(target_arch = "wasm32"))]
 pub mod viewer_test_utils;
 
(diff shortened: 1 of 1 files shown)


crates/viewer/re_viewer/Cargo.toml
```diff
@@ -103,6 +105,10 @@ re_viewport.workspace = true
 re_analytics = { workspace = true, optional = true }
 re_view_map = { workspace = true, optional = true }
 
+# Test dependencies (optional):
+egui_kittest = { workspace = true, optional = true }
+tokio = { workspace = true, optional = true }
+
```
MichaelGrupp, line 111: Not sure if they still need to be also in dev-dependencies? (see end of the file here)

emilk: No, this is correct.The test has `#![cfg(feature = "testing")]` as it should

Issue #9848, opened 2025-04-30
`debug_assert` triggers when closing partially loaded recording


lucasmerlin:
**Describe the bug**
Opening the air traffic example on main and quickly closing it (while it's still loading) triggers a debug_assert:

**To Reproduce**
1. open air traffic example
2. quickly close it
3. should crash

**Expected behavior**
No crash

**Backtrace**

```
thread 'main' panicked at 'active recordings should always have associated caches'
re_viewer_context/src/store_hub.rs:525
stack backtrace:
   8: core::panicking::panic_fmt
   9: re_viewer::app::App::receive_messages
  10: <re_viewer::app::App as eframe::epi::App>::update
  11: egui::context::Context::run
  12: eframe::native::epi_integration::EpiIntegration::update
[...]
```

abey79:
Possibly easier repro:

1) open two recordings from a remote server (I couldn't trigger the crash with a single recording open)
2) go back to the server screen (I couldn't trigger the crash when remaining on the recording view)
3) close one of the recording from the recording panel
4) :boom:

jccampagne:
I stumbled upon this issue as well yesterday. I had a quick look into it.

## In short

The app panics because it still receives messages, even though the app (air traffic example) has closed already.
This does not happen when all the data has loaded (in the air traffic example).

This is reproducible in the ARKit example as well (slowed down a bit by a debugging session, allowing one to click close while messages are still being processed).

## More details (in chronological order)

### App closing

The user closes the app while it is still loading:

`App.run_system_command()` is called:

and this statement is run:

```rust
fn run_system_command() {
...
            SystemCommand::CloseApp(app_id) => {
                store_hub.close_app(&app_id);
            }
...

}
```

This calls `StoreHub.close_app()`

```rust
        self.caches_per_recording
            .retain(|store_id, _| !store_ids_removed.contains(store_id));
```


### Messages being received

In the next cycle, messages are still arriving. They are processed in `App.receive_messages()`.

```rust
impl App {

    fn receive_messages(&self, store_hub: &mut StoreHub, egui_ctx: &egui::Context) {

....
            match store_hub.entity_db_mut(store_id).add(&msg) {
                Ok(store_events) => {
                    if let Some(caches) = store_hub.active_caches() {
                        caches.on_store_events(&store_events);
                    }
                    self.validate_loaded_events(&store_events);
                }
....
}
```

But, by then, the caches have already been purged, hence the panic on

```rust
impl StoreHub {

[...]

PR #10963, opened 2025-08-22
Fix closing recordings/tables that are still being loaded showing up again

jprochazk:
* Fixes https://github.com/rerun-io/rerun/issues/10965
* Fixes https://github.com/rerun-io/rerun/issues/9848

To reproduce the issue, open a large partition or throttle your network speed. As it's loading, close it in the recording panel. It actually stays there and continues downloading.

The fix is to drop the receiver as well.

I also noticed that dropping the receiver doesn't immediately cancel the download, the `GetChunks` continues in the background according to the console network tab. We should probably look into making those requests fully cancel safe

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -558,7 +558,18 @@ impl App {
 
             SystemCommand::CloseRecordingOrTable(entry) => {
                 // TODO(#9464): Find a better successor here.
+
+                let data_source = match &entry {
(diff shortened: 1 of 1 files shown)


Wumpf (approved):
using the flight data nightly example I confirmed that it fixes the assert 👍  (added ticket to the close list)

This fix is relevant for OSS & Cloud the same way as, adding to changelog accordingly :)

jprochazk:
Created a follow-up issue about request cancellation: https://github.com/rerun-io/rerun/issues/10967

PR #10969, opened 2025-08-22
Fix closing SDK connections

Wumpf:
### Related

* Follow-up to https://github.com/rerun-io/rerun/pull/10963

### What

previous PR was too aggressive: made the viewer "deaf" to anything from the SDK after closing one SDK originated recording


jprochazk:
I think this one is still too aggressive, the only thing we should not retain are redap connections. Nothing else is 1 connection == 1 recording

crates/viewer/re_viewer/src/app.rs
```diff
@@ -565,8 +565,32 @@ impl App {
...
+                        // TODO(#10967): We still stream in some data after that.
+                        SmartChannelSource::RedapGrpcStream { .. } => true,
+
+                        // Don't close generic connections (like to an SDK) that may feed in different recordings over time.
+                        SmartChannelSource::RrdWebEventListener
+                        | SmartChannelSource::JsChannel { .. }
+                        | SmartChannelSource::Sdk
+                        | SmartChannelSource::Stdin
+                        | SmartChannelSource::MessageProxy(_) => false,
+                    };
```
jprochazk, line 587:
```suggestion
                    let should_close = matches!(data_source, SmartChannelSource::RedapGrpcStream { .. });
```

Wumpf:
> Nothing else is 1 connection == 1 recording

Why would it not be the case for http and files?
Http is the reason why we had the exact same failure (plus the assert issue I previously linked) easily repro on the flight data example

crates/viewer/re_viewer/src/app.rs
Wumpf: don't want to use matches since it doesn't warn on new sources

jprochazk:
> Why would it not be the case for http and files?

Because the `.rrd` format supports concatenating (not merging) streams into a single file. So in both of those cases, you could be served an `.rrd` file which contains 2 or more logical recordings. Nothing is stopping an `.rrd` file from having more than one `SetStoreInfo` message. That means if you're streaming that file over HTTP or reading it from disk, and you close it, then it's possible that at some point in the stream it creates and begins populating a new store. So the reasoning for SDK, stdin, messageproxy, etc. _should_ be the same as for http and files

jprochazk:
I haven't tried to understand what is up with the other issue the previous PR fixed yet, maybe there's a different way to fix it? Also, hint hint nudge nudge *cough* https://github.com/rerun-io/rerun/issues/8631 *cough*

Wumpf:
in a nutshell we have the exact same issue you fixed for the grpc entries also with streaming from http: if you close a recording that streams from file or http, we also have to close the data sources otherwise recordings will show up again. In Debug this causes a debug assertion to trigger on top of that which your original PR fixes (and this PR tries to preserve by means of still removing those sources)

> Also, hint hint nudge nudge cough https://github.com/rerun-io/rerun/issues/8631 cough

lalalala I can't hear you ;-) (in this pr at least 😄 )

Wumpf:
> [....] So the reasoning for SDK, stdin, messageproxy, etc. should be the same as for http and files

that does make sense :/ except that it's quite the corner case. I think it's fair to say that the close on those recordings also closes the connection? It's a bit odd that we have this as a separate concept internally that we don't expose :(
I.e. I think the bug of not receiving more recordings is better than the ghost recordings that can come into existance right now (plus the annoying debug assertion described in https://github.com/rerun-io/rerun/issues/9848)

jprochazk:
> I think the bug of not receiving more recordings is better than the ghost recordings that can come into existance right now

I think I agree, and it's an edge case that should eventually go away. Seems acceptable to tell people "don't remove the data source if you have more than one logical recording in it and you want to see both". It seems unlikely for people to even do that in the first place

Wumpf:
👍 but good you brought up this concern - I didn't really consider the possibility. We're clearly not "done" here!

PR #10970, opened 2025-08-22
Persist fallback token

jprochazk:
To reproduce the issue:

- Navigate to an instance with auth enabled, putting in instance URL and token in query params
  - `http://localhost:9090/?url=rerun%2Bhttp%3A%2F%2F127.0.0.1%3A51234%2Fcatalog&token=<TOKEN>`
- The server browser should activate and load up fine
- Reload the page
- The server browser no longer loads, with an error message like "missing credentials".

The token is removed from the URL by the Viewer, but it is also not persisted. That means it's just gone upon reload. Very annoying for the user, so we now persist the fallback token. If a new one is set via `?token=`, then it takes precedence over the stored one.


abey79 (approved):
That looks reasonable to me. Please remove the now updated comment in the `client()` docstring.

PR #10972, opened 2025-08-22
In-process gRPC server for e2e tests

aedm:

### DIFF
--- a/tests/rust/re_integration_test/Cargo.toml
+++ b/tests/rust/re_integration_test/Cargo.toml
@@ -13,13 +13,14 @@ publish = false
 [dependencies]
 nix = { workspace = true, features = ["signal"] }
 ureq.workspace = true
+re_server.workspace = true
+tokio.workspace = true
 
 [dev-dependencies]
 insta.workspace = true
 re_viewer = { workspace = true, features = ["testing"] }
 egui_kittest.workspace = true
 egui.workspace = true
-tokio.workspace = true
 
 [lints]
 workspace = true
--- a/tests/rust/re_integration_test/Cargo.toml
+++ b/tests/rust/re_integration_test/Cargo.toml
@@ -15,6 +15,8 @@ nix = { workspace = true, features = ["signal"] }
 ureq.workspace = true
 re_server.workspace = true
 tokio.workspace = true
+re_protos.workspace = true
+re_grpc_server.workspace = true
 
 [dev-dependencies]
 insta.workspace = true
--- a/tests/rust/re_integration_test/src/lib.rs
+++ b/tests/rust/re_integration_test/src/lib.rs
@@ -118,9 +118,11 @@ pub async fn load_test_data(port: u16) -> String {
         ]);
         let output = script
             .output()
-            .expect("Failed to run re_integration.py script")
-            .stdout;
-        String::from_utf8(output).expect("Failed to convert output to string")
+            .expect("Failed to run re_integration.py script");
+        let stderr = String::from_utf8(output.stderr).expect("Failed to convert stderr to string");
+        let stdout = String::from_utf8(output.stdout).expect("Failed to convert output to string");
+
+        format!("{}\n\n{}", stdout.trim(), stderr.trim())
     })
     .await
     .expect("Failed to run re_integration.py script")
--- a/tests/rust/re_integration_test/src/lib.rs
+++ b/tests/rust/re_integration_test/src/lib.rs
@@ -1,38 +1,60 @@
 //! Integration tests for rerun and the in memory server.
 
-use std::net::TcpListener;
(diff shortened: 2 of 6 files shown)


aedm:
@rerun-bot full-check

aedm:
@rerun-bot full-check

emilk:
Let's get this in and get CI green asap :]

tests/rust/re_integration_test/Cargo.toml
```diff
@@ -15,15 +15,14 @@ re_grpc_client.workspace = true
 re_protos.workspace = true
 re_sdk.workspace = true
 re_uri.workspace = true
-
-ureq.workspace = true
+re_server.workspace = true
```
Wumpf: emil wants these sorted 🙄

tests/rust/re_integration_test/src/lib.rs
```diff
@@ -2,53 +2,38 @@
...
 use re_uri::external::url::Host;
-use ureq::OrAnyStatus as _;
+use std::net::TcpListener;
 
 pub struct TestServer {
-    server: Child,
+    /// The server needs an async function to stop properly but `AsyncDrop` still is experimental.
+    /// Tracking issue: <https://github.com/rust-lang/rust/issues/126482>
+    /// These tests just do the simple thing and drop the server handle hoping for the best.
+    _server_handle: ServerHandle,
```
emilk: We should be able to do better than "hope for the best" :)

Either:
* Use `tokio::blocking` to call an async shutdown in the destructor
* Add an explicit `server.shutdown().await;` and panic in the destructor if it hasn't been called

aedm: I could use some help here. :) I'm not really sure what you mean by `tokio::blocking`, I don't think Tokio has a module like that. If you mean `tokio::spawn_blocking` it does the opposite: calls a blocking function from an async context. Here we need to call async from the synchronous `drop` function. As far as I can tell, Tokio provides no mechanism for synchronously waiting for a single async task to finish.

I can add an explicit shutdown function to TestServer and an assertion in the drop function, but it needs to be added to each test manually.

Alternatively, I can add a `tokio::spawn` call to the TestServer's `drop` function which schedules the server shutdown but doesn't wait for it, so it's somewhat similar to "hoping for the best".

I'll go with the explicit shutdown, it's uncomfortable but it works.

emilk: Sorry, I meant `tokio::task::block_in_place`

aedm: Great! I did not know about that. :)

tests/rust/re_integration_test/Cargo.toml
emilk: Adding new items last to a list is the best way to _guarantee_ a merge conflict when two people add to the same list.

You also have to ask yourself: as reader of the code, what am I most interested in:
* A chronological history of the code base?
* A clean, readable list where I can binary-search with my eyes to find what I want?

Wumpf: that said, we really should have a check for that. If there's no way to tell the toml formatter about it then we need at least something crude in `lints.py`

aedm:
@rerun-bot full-check

tests/rust/re_integration_test/src/lib.rs
aedm: @emilk Fixed, please take a look

Wumpf (approved):

tests/rust/re_integration_test/tests/re_integration.rs
```diff
@@ -3,9 +3,9 @@ use egui_kittest::kittest::Queryable as _;
 use re_integration_test::TestServer;
 use re_viewer::viewer_test_utils;
 
-#[tokio::test]
+#[tokio::test(flavor = "multi_thread")]
```
Wumpf, line 6: didn't know about this! So without it it's using just a single thread instead of spawning worker threads as it would do in regular applications...?

tests/rust/re_integration_test/src/lib.rs
Wumpf: Emil is out because he's not feeling well, so I had a look and looks fine to me 👍

aedm:
One of the snapshots was flaky on CI due to the welcome screen.
Removed the snapshot and added a ticket: https://github.com/rerun-io/rerun/issues/10989

emilk:
@rerun-bot full-check

tests/rust/re_integration_test/tests/re_integration.rs
aedm: `tokio::task::block_in_place` only works with `multi_thread` enabled. :-(

PR #10975, opened 2025-08-23
gitignore `target_pixi_wasm`

jprochazk:


Wumpf:
can't have enough target folders T_T

PR #10982, opened 2025-08-25
Always put the pixi target directory in the root of the repository

emilk:


emilk:
@rerun-bot full-check

.gitignore
```diff
@@ -19,12 +19,13 @@ _deps
...
-**/target_pixi_wasm
-**/target_ra
-**/target_wasm
+# Rust compile target directories.
+# These should only be found in the repository root, so prefix them all with slashes:
+/target
+/target_pixi
+/target_pixi_wasm
+/target_ra
+/target_wasm
```
Wumpf, line 28: clever! :)

Wumpf (changes requested):
yeah good I tested this, on Windows it doesn't resolve the env var and instead creates a directory called `$PIXI_PROJECT_ROOT`

Wumpf:
taking this over and looking into fixing

pixi.toml
```diff
@@ -45,20 +45,25 @@ libc = "2.28"
...
 [target.unix.activation.env]
 # The executable extension for binaries on the current platform.
 EXECUTABLE_EXTENSION = ""
 
+# Put pixi builds in its own build dir, otherwise `pixi run cargo build`
+# conflicts with vanilla`cargo build`, and running one invalidates (parts) of the other.
+# See https://github.com/rerun-io/rerun/issues/10934 for more.
+#
+# As of pixi 0.53.0, environment variable resolving doesn't work with this on Windows,
+# which is why we have to use the `%` syntax for Windows instead of `$`.
```
emilk, line 57: Sounds like a bug in pixi? Is there a pixi issue we can link to here from a TODO?

Wumpf: Looked a bit but couldn't find anything. ~I'll ask on their discord~ Creating a bug

Wumpf: https://github.com/prefix-dev/pixi/issues/4431

emilk:
@rerun-bot full-check

Wumpf:
I think the integration test needs additional love `"../../../C:\\a\\rerun\\rerun/target_pixi/debug/rerun"` looks very much like a failure caused by this!

To be on the save side let's wait until https://github.com/rerun-io/rerun/pull/10972#pullrequestreview-3150860868  is in

Wumpf:
Regular CI looks good now thanks to Gabor's changes. Let's go all in just in case.

@rerun-bot full-check

PR #10985, opened 2025-08-25
Fix `ReceiveSet::recv` panicing without receivers

Wumpf:
### Related

* Followup to https://github.com/rerun-io/rerun/pull/10944

### What
Caused spawning web server without anything connected to crash (`pixi run rerun-web`)

### DIFF
--- a/crates/utils/re_smart_channel/src/receive_set.rs
+++ b/crates/utils/re_smart_channel/src/receive_set.rs
@@ -116,6 +116,11 @@ impl<T: Send> ReceiveSet<T> {
         let mut rx = self.receivers.lock();
 
         rx.retain(|r| r.is_connected());
+        if rx.is_empty() {
+            // Have to early out here, because `Select::select` will panic if there are no channels to select from.
+            return Err(RecvError);
+        }
+
         let mut sel = Select::new();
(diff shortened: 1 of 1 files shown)


emilk (approved):

crates/utils/re_smart_channel/src/receive_set.rs
```diff
@@ -116,6 +116,11 @@ impl<T: Send> ReceiveSet<T> {
         let mut rx = self.receivers.lock();
 
         rx.retain(|r| r.is_connected());
+        if rx.is_empty() {
```
jprochazk, line 119: This calls `retain` again internally - should we check `self.receivers.lock()` directly?

emilk: I think you might have misread this as `self.is_emtpy()`?

jprochazk: i have indeed, yes :grimacing: feel free to ignore

Wumpf: yes, it's doing what Jan is asking for :)

PR #10987, opened 2025-08-25
rerun_cpp: Fix useless semicolon warnings

MichaelGrupp:
This shows up when user code is compiled with `-Wpedantic`.


MichaelGrupp:
I see that pedantic warnings are already on, maybe this also needs to be handled stricter as part of a linter check in CI somehow? https://github.com/rerun-io/rerun/blob/bdbcfffc07df82b3a5f35c5478cc7f981b7baf44/CMakeLists.txt#L62

(could be a follow-up)

Issue #10945, opened 2025-08-20
Try using more buildjet runners


Wumpf:
how much faster would it be to swap out some runners?

Wumpf:
no Mac & Windows runners there and that's what we're bottlenecked on

PR #10990, opened 2025-08-25
Try bigger images for Rust tests

Wumpf:
### Related

* Part of  https://github.com/rerun-io/rerun/issues/10945

### What

Buildjet doesn't have Mac M-series runners unfortunately. But there's a bigger tier still on Github. Let's see how it performs.

TODO:
* [ ] Note down before after numbers

### DIFF
--- a/.github/workflows/contrib_checks.yml
+++ b/.github/workflows/contrib_checks.yml
@@ -97,7 +97,7 @@ jobs:
   rs-lints:
     name: Rust lints (fmt, check, clippy, tests, doc)
     # TODO(andreas): setup-vulkan doesn't work on 24.4 right now due to missing .so
-    runs-on: ubuntu-22.04-large
+    runs-on: buildjet-16vcpu-ubuntu-2204
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/contrib_checks.yml
```diff
@@ -97,7 +97,7 @@ jobs:
   rs-lints:
     name: Rust lints (fmt, check, clippy, tests, doc)
     # TODO(andreas): setup-vulkan doesn't work on 24.4 right now due to missing .so
-    runs-on: ubuntu-22.04-large
+    runs-on: buildjet-16vcpu-ubuntu-2204
```
emilk, line 100: Are we sure we can run contributor checks on buildjet?

emilk (approved):

Wumpf:
looks like there's no windows runners either. Bit of a dead end

.github/workflows/contrib_checks.yml
Wumpf: no idea, but why not?

Issue #10958, opened 2025-08-21
Deprecate Python 3.9


ntjohnson1:
Python 3.9 reaches EOL October 2025. We should get in a habit of announcing when we remove support (and maybe have details for our support approach in general).

----

* [x] add to changelog
* [x] add to migration guide
* [x] edit python version notes support in docs
* [x] print a notice on startup if an old python is used

see also https://github.com/rerun-io/rerun/pull/7933/files

ntjohnson1:
I think we might be able to just add something to the release notes. I can look closer at how some more popular packages manage it.

ntjohnson1:
We should add a note the 3.9 is no longer supported on conda as of 0.25. maybe we take that into consideration for our python version support

jleibs:
Here's what we did last time:
https://github.com/rerun-io/rerun/pull/7933/files

PR #10996, opened 2025-08-25
Upgrade Ruff to 0.12.10

ntjohnson1:
Bumps ruff to latest. I'm not sure if we have specific concerns in which case we can pin lower. Right now we are just behind a few.

### What
Upgrades ruff and runs `py-fmt` which handles almost all the fixes. A few minor cleanups for the remaining.
Some of the biggest pieces are:
* [TC006](https://docs.astral.sh/ruff/rules/runtime-cast-value/)
* [UP045](https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/)
   * This was the motivating factor here since it kicked in for a new repo but I knew we had inconsistencies here
* [C420](https://docs.astral.sh/ruff/rules/unnecessary-dict-comprehension-for-iterable/)
* Improved [PLW1514](https://docs.astral.sh/ruff/rules/unspecified-encoding/)

### TODO
* ~https://github.com/rerun-io/rerun/issues/10958~ apparently this PR still works on 3.9


rerun_py/tests/unit/test_arrows3d.py
```diff
@@ -47,12 +49,12 @@ def test_arrows3d() -> None:
...
-        radii = cast(Optional[Float32ArrayLike], radii)
-        colors = cast(Optional[Rgba32ArrayLike], colors)
-        labels = cast(Optional[Utf8ArrayLike], labels)
-        class_ids = cast(Optional[ClassIdArrayLike], class_ids)
+        vectors = cast("Vec3DArrayLike", vectors)
+        origins = cast("Optional[Vec3DArrayLike]", origins)
+        radii = cast("Optional[Float32ArrayLike]", radii)
+        colors = cast("Optional[Rgba32ArrayLike]", colors)
+        labels = cast("Optional[Utf8ArrayLike]", labels)
+        class_ids = cast("Optional[ClassIdArrayLike]", class_ids)
```
abey79, line 57: somehow these `Optional`s were allowed to survive? :)

rerun_py/tests/unit/test_box2d.py
```diff
@@ -47,13 +49,13 @@ def test_boxes2d() -> None:
...
-        labels = cast(Optional[Utf8ArrayLike], labels)
-        draw_order = cast(Optional[Float32ArrayLike], draw_order)
-        class_ids = cast(Optional[ClassIdArrayLike], class_ids)
+        half_sizes = cast("Vec2DArrayLike", half_sizes)
+        centers = cast("Vec2DArrayLike", centers)
+        radii = cast("Optional[Float32ArrayLike]", radii)
+        colors = cast("Optional[Rgba32ArrayLike]", colors)
+        labels = cast("Optional[Utf8ArrayLike]", labels)
+        draw_order = cast("Optional[Float32ArrayLike]", draw_order)
+        class_ids = cast("Optional[ClassIdArrayLike]", class_ids)
```
abey79, line 58: same

rerun_py/tests/unit/test_box3d.py
```diff
@@ -69,15 +71,15 @@ def test_boxes3d() -> None:
...
-        fill_mode = cast(Optional[rr.components.FillMode], fill_mode)
+        half_sizes = cast("Vec3DArrayLike", half_sizes)
+        centers = cast("Vec3DArrayLike", centers)
+        rotation_axis_angles = cast("RotationAxisAngleArrayLike", rotation_axis_angles)
+        quaternions = cast("QuaternionArrayLike", quaternions)
+        radii = cast("Optional[Float32ArrayLike]", radii)
+        colors = cast("Optional[Rgba32ArrayLike]", colors)
+        labels = cast("Optional[Utf8ArrayLike]", labels)
+        class_ids = cast("Optional[ClassIdArrayLike]", class_ids)
+        fill_mode = cast("Optional[rr.components.FillMode]", fill_mode)
```
abey79, line 82: same (+ many more....)

abey79: Side thought: we should figure out a better way for these archetype tests, it's annoying to have to create/maintain them manually, and I don't like the implementation (e.g. these annoying casts).

abey79 (approved):
Nice, thanks.

examples/python/arkit_scenes/arkit_scenes/download_dataset.py
```diff
@@ -196,7 +196,7 @@ def download_laser_scanner_point_clouds(laser_scanner_point_cloud_id: str, visit
         download_file(file_url, filename, laser_scanner_point_clouds_folder_path)
 
 
-def get_metadata(dataset: str, download_dir: Path) -> Optional[pd.DataFrame]:
```
emilk, line 199: I think this will break `Python 3.9`, so we should do this first: 
* https://github.com/rerun-io/rerun/issues/10958

abey79: Oh, you're right. For a moment I thought python 3.9 did support those (I mixed up with `ThingLike = Thing | Stuff` type variable). ruff should know better though. There must be something wrong in the configuration file.

That said, the earlier we get rid of 3.9 the better in my book, because it gets us ✨`match` ✨

ntjohnson1: Conda just dropped support for new packages with python 3.9. I'm open to aligning with conda but we don't currently have a support policy or announcement in place.

As for this breaking python 3.9 ruff [states](https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/) that this is safe because we import annotations from future: 

With this quick hack (downgrade python and ignore a few packages that require >= 3.10) I was able to run `pixi run py-test` and everything passed.
```diff
diff --git a/pixi.toml b/pixi.toml
index 0f0e7e9d8..2d5a1fb98 100644
--- a/pixi.toml
+++ b/pixi.toml
@@ -610,7 +610,7 @@ prettier = ">=3.6"
 pyarrow = "18.0.0.*"              # Whenever upgrading here, also make sure to upgrade in `rerun_py/pyproject.toml`
 pytest = ">=8.4.1"
 pytest-benchmark = ">=4.0.0,<4.1"
-python = "=3.11"                  # We use the latest Python version here, so we get the latest mypy etc, EXCEPT 3.12 is too new for some of our examples. We run our CI tests on ALL supported versions though.
+python = "=3.9"                  # We use the latest Python version here, so we get the latest mypy etc, EXCEPT 3.12 is too new for some of our examples. We run our CI tests on ALL supported versions though.
 ruff = "0.12.10.*"
 semver = ">=3.0,<3.1"
 taplo = "=0.9.1"
@@ -765,10 +765,10 @@ platforms = ["linux-64", "osx-arm64", "win-64"]
 macos = "12.0" # needed for scipy
 
 [feature.examples-common.dependencies]
-python = "=3.11"
+python = "=3.9"
 
 # TODO(ab): otherwise pixi will try to build from source and fail, not sure why (used by air_traffic_data)
-pyproj = ">=3.7.0"
+#pyproj = ">=3.7.0"
 
 # Need to use the conda version of umap-learn to be compatible with numpy-2
 umap-learn = "==0.5.7"
@@ -790,7 +790,7 @@ blueprint = { path = "examples/python/blueprint", editable = true }
 blueprint_stocks = { path = "examples/python/blueprint_stocks", editable = true }
 camera_video_stream = { path = "examples/python/camera_video_stream", editable = true }
 clock = { path = "examples/python/clock", editable = true }
-controlnet = { path = "examples/python/controlnet", editable = true }
+#controlnet = { path = "examples/python/controlnet", editable = true }
 dataframe_query = { path = "examples/python/dataframe_query", editable = true }
 detect_and_track_objects = { path = "examples/python/detect_and_track_objects", editable = true }
 dicom_mri = { path = "examples/python/dicom_mri", editable = true }
@@ -813,7 +813,7 @@ graphs = { path = "examples/python/graphs", editable = true }
 nuscenes_dataset = { path = "examples/python/nuscenes_dataset", editable = true }
 nv12 = { path = "examples/python/nv12", editable = true }
 objectron = { path = "examples/python/objectron", editable = true }
-open_photogrammetry_format = { path = "examples/python/open_photogrammetry_format", editable = true }
+#open_photogrammetry_format = { path = "examples/python/open_photogrammetry_format", editable = true }
 openstreetmap_data = { path = "examples/python/openstreetmap_data", editable = true }
 plots = { path = "examples/python/plots", editable = true }
 raw_mesh = { path = "examples/python/raw_mesh", editable = true }
```

rerun_py/tests/unit/test_box3d.py
ntjohnson1: I just commented out all these casts and mypy still succeeded. I think we just need to add `--warn-redundant-casts` to our mypy incantation in pixi so we get warned when we are doing useless extra casting work.

EDIT: I added the warn and it didn't flag this even though they seem unnecessary.

rerun_py/tests/unit/test_arrows3d.py
ntjohnson1: Opened an issue for this https://github.com/astral-sh/ruff/issues/20096

ntjohnson1: Echoing from a comment on the issue. It looks like this bug is specific to 3.9 so when we drop it for 3.10 we should get these cleaned up even if the bug isn't fixed.

emilk:
Is this ready to be merged?

ntjohnson1:
> Is this ready to be merged?

Ya, I wanted to wait to do it until late CET so I could merge it and post to slack to try and minimize some conflicts/ CI failures (since I know not everyone has commit hooks setup). I just pulled main and will merge when green

PR #10998, opened 2025-08-25
Correct 'altitude' to 'latitude' in documentation

skalldri:
### Related

N/A

### What

Correct a typo in the documentation that suggests the `GeoLineString` accepts an "altitude" value, when in reality it expects a "latitude" value.


abey79 (approved):
thanks!

abey79:
My bad, fixed in https://github.com/rerun-io/rerun/pull/10999

PR #10999, opened 2025-08-26
Fix documentation for `GeoLineStrings` archetype

abey79:
### Related

* related #10998

### What

I clicked the merge button too quick with #10998. This PR implements the fix correctly.


abey79:
Link failure is a dud.

Wumpf:
Link failure is an error to handle relative links in the new diff based link checking. Looking into it!

PR #10949, opened 2025-08-20
test link checker

grtlr:

PR #11001, opened 2025-08-26
PR link checker improvements/fixes

Wumpf:
### Related

* follow-up to https://github.com/rerun-io/rerun/pull/10949

### What

* update to latest `lychee`
* run PR link checker individually for each changed file so that we can take relative links into account
* better error output
* fix running locally on windows
* look at all markdown links
* parameter for not deleting the temp files (nice for debugging the link checker itself)
* new link check exception for lib.rs since we get `forbidden` from it all the time

Run with some relative links: https://github.com/rerun-io/rerun/actions/runs/17233719945/job/48893553095?pr=11001

---
* [x] Remove dummy file changes again

### DIFF
--- a/scripts/ci/pr_link_checker.py
+++ b/scripts/ci/pr_link_checker.py
@@ -15,6 +15,12 @@ import subprocess
 import sys
 import tempfile
 from pathlib import Path
+from typing import Any
+
+
+def eprint(*args: Any, **kwargs: Any) -> None:
+    """Prints a message to stderr."""
+    print(*args, file=sys.stderr, **kwargs)
 
 
 def get_added_lines_with_links(base_ref: str = "origin/main") -> dict[str, list[str]]:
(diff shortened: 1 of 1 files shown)


scripts/ci/pr_link_checker.py
```diff
@@ -45,105 +52,123 @@ def get_added_lines_with_links(base_ref: str = "origin/main") -> dict[str, list[
...
-        elif line.startswith("+") and not line.startswith("+++") and current_ext:
+        elif line.startswith("+") and not line.startswith("+++") and current_file:
             # This is an added line, check if it contains URLs
             line_content = line[1:]  # Remove the '+' prefix
             if (
                 "http://" in line_content
                 or "https://" in line_content
                 or "ftp://" in line_content
                 or "file://" in line_content
+                or re.search(r"\[.+\]\(.+\)", line_content)  # Markdown links [text](url)
```
abey79, line 78: This has high chances to trigger a false positive due to some language construct (e.g. python list of callable: `funcs[i](args)`. How problematic would that be?

abey79 (approved):
Thanks for improving this!

scripts/ci/pr_link_checker.py
Wumpf: You're right. Shouldn't be problematic at all. I actually considered instead of doing this line extraction logic to mark whole file as to be link checked. But Jochen/Claude already put so much effort into extracting those lines, so I figured I keep it

PR #11002, opened 2025-08-26
Prevent MCAP loader trigger for non-MCAP files on web

### LINKED ISSUE #10995: Drag and dropping images or videos on the web creates a second broken recording due to mcap loader
* drag in a jpg or mp4
* get `Failed to load data from file path="file.mp4" loader="McapLoader" err=Bad magic number`
* two recordings: one broken one and one with the desired data
    * both have an awful name 

Image

Curiously I can't reproduce this issue on native, but maybe that's just because some build flag was different.

oxkitsune:
### Related

* Fixes #10995 

### What

We did not check whether the uploaded file is actually a `.mcap` file on web builds 😳

### DIFF
--- a/crates/store/re_data_loader/src/loader_mcap.rs
+++ b/crates/store/re_data_loader/src/loader_mcap.rs
@@ -1,6 +1,6 @@
 //! Rerun dataloader for MCAP files.
 
-use std::{io::Cursor, sync::mpsc::Sender};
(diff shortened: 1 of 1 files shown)


oxkitsune:
> It's interesting how that fixes also the application id to be not broken

If two dataloaders trigger, they both pick a random application id it seems.

PR #11004, opened 2025-08-26
Fix flaky integration test due to multi-pass warning dot

Wumpf:
Another thing that may or may not show that throws off our integration tests!

### DIFF
--- a/crates/viewer/re_global_context/src/app_options.rs
+++ b/crates/viewer/re_global_context/src/app_options.rs
@@ -16,6 +16,9 @@ pub struct AppOptions {
     /// Show a warning if a software rasterizer is being used (i.e. no GPU).
     pub show_software_rasterizer_warning: bool,
 
+    /// Show an orange dot to warn about multi-pass layout in egui..
+    pub show_multi_pass_warning: bool,
+
     /// Include the "Welcome screen" application in the recordings panel?
     #[serde(alias = "include_welcome_screen_button_in_recordings_panel")]
     pub include_rerun_examples_button_in_recordings_panel: bool,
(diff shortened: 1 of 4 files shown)


crates/viewer/re_global_context/src/app_options.rs
```diff
@@ -16,6 +16,9 @@ pub struct AppOptions {
     /// Show a warning if a software rasterizer is being used (i.e. no GPU).
     pub show_software_rasterizer_warning: bool,
 
+    /// Show an orange dot to warn about multi-pass layout in egui..
+    pub show_multi_pass_warning: bool,
```
emilk: These are serialized/persistent options. Initializing it to `cfg!(debug_assertions)` will still keep it true when you then later run a release build. I don't think that is what we want.

We can mark it as `#[serde(skip)]` but I'm not 100% the option even belongs here. I think we should keep these to user-facing options, in most cases.

Wumpf: I figure in the same vain we should also remove `show_software_rasterizer_warning` then?

crates/viewer/re_viewer/src/ui/top_panel.rs
```diff
@@ -127,11 +127,11 @@ fn top_bar_ui(
             }
         }
 
-        if cfg!(debug_assertions) {
+        if !app.app_env().is_test() {
```
emilk:
```suggestion
        if cfg!(debug_assertions) && !app.app_env().is_test() {
```

emilk (approved):

PR #11017, opened 2025-08-27
[ci] Re-use built CLI for man page check

Wumpf:
We recently introduced a check for the cli's doc page. But we built the cli from scratch for that (plus some sccache caching which doesn't work _that_ well) taking around 7min 😱 . We already build the CLI on *every* CI run, so let's re-use that!


abey79 (approved):
noice!

PR #11019, opened 2025-08-27
C++ build now aggressively picks up sccache

Wumpf:
Previously, sccache got picked up by Windows & Linux, but Mac didn't for some reason. This PR handles this a bit more aggressively.

Note that by changing our top level CMake file this does **not** affect our C++ shippable artifact.

Before caching / before this / after:
* [ ] TODO


Wumpf:
False alarm. C++ tests don't run as often and what I looked at is likely just trashed the cache because of the other changes - on main ci we already got cache hits aplenty, see https://github.com/rerun-io/rerun/actions/runs/17266629749/job/49000204641

PR #11020, opened 2025-08-27
Add flag that allows cleanup of more recent Lance files in the maintenance

zehiko:
### What

Introduce a flag in the ``DoMaintenance`` API call that allows overriding default platform behaviour that prevents cleanup of Lance dataset artifacts older than 1 hour. If ``allow_recent_cleanup`` is set, ``cleanup_before`` will be respected, regardless of its value.

### DIFF
--- a/crates/store/re_protos/proto/rerun/v1alpha1/frontend.proto
+++ b/crates/store/re_protos/proto/rerun/v1alpha1/frontend.proto
@@ -352,6 +352,6 @@ message DoMaintenanceRequest {
 
   // Override default platform behavior and allow cleanup of recent files. This will respect
   // the value of `cleanup_before` timestamp even if it's more recent than 1 hour.
-  // Warning - trying to cleanup more recent Lance artifacts may cause Dataset update issues.
-  bool allow_recent_cleanup = 5;
+  // ⚠️ Do not ever use this unless you know exactly what you're doing. Improper use will lead to data loss.
+  bool unsafe_allow_recent_cleanup = 5;
 }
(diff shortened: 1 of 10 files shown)


crates/store/re_protos/proto/rerun/v1alpha1/frontend.proto
```diff
@@ -343,5 +343,15 @@ message DoMaintenanceRequest {
...
+  // timestamp will be used. This is to prevent still used files (like recent transaction files)
+  // to be removed and cause Lance Dataset update issues.
+  // See https://docs.rs/lance/latest/lance/dataset/cleanup/index.html
+  // and https://docs.rs/lance/latest/lance/dataset/cleanup/fn.cleanup_old_versions.html
   google.protobuf.Timestamp cleanup_before = 4;
+
+  // Override default platform behavior and allow cleanup of recent files. This will respect
+  // the value of `cleanup_before` timestamp even if it's more recent than 1 hour.
+  // Warning - trying to cleanup more recent Lance artifacts may cause Dataset update issues.
+  bool allow_recent_cleanup = 5;
```
teh-cmc: I would go so far as to call this `unsafe_allow_recent_cleanup`. And use scarier words too. The people need to be scared.

```suggestion
  // Override default platform behavior and allow cleanup of recent files. This will respect
  // the value of `cleanup_before` timestamp even if it's more recent than 1 hour.
  //
  // ⚠️ Do not ever use this unless you know exactly what you're doing. Improper use will lead to data loss.
  bool unsafe_allow_recent_cleanup = 5;
```

zehiko: agree :)

teh-cmc:
Is this still a draft?

zehiko:
>  Is this still a draft?

no, my bad 🤦‍♂️

PR #11021, opened 2025-08-27
[POC] Add support for converting tables to recordings

abey79:
### Related

* related to https://linear.app/rerun/project/visualize-tables-in-the-viewer-draft-056c36c5097e/overview

### What

TBD.

Look for the `Convert to recording` button in the table footer

### TODO

- [ ] write tests for the conversion code
- [ ] move conversion code elsewhere?
- [ ] UI of converted recording: app id? etc.?
- [ ] address `/__properties/` clash when converting a partition table to a recording


ntjohnson1:
Using this draft for discussion. This probably needs to be rebased for things to work. I had a question about how we support static data/AssertVideos more generally.

Here is an RRD that has asset videos:  https://drive.google.com/file/d/1bib3X-5XZ4WbN9qVgPgpllZ4A1pqzpsA/view?usp=drive_link

If you load it in the OSS server then

```python

import os

CATALOG_URL = "rerun+http://localhost:51234"
os.environ["RERUN_NOTEBOOK_ASSET"] = "serve-local"
import rerun as rr
client = rr.catalog.CatalogClient(CATALOG_URL)
all_entries = client.all_entries()
first_entry = all_entries[0]
dataset: rr.catalog.DatasetEntry = client.get_dataset_entry(name=first_entry.name)
top_view_video = dataset.dataframe_query_view(index=None, contents={"/observation.images.Lwebcam": ["AssetVideo:blob", "VideoFrameReference:timestamp"]}).df()

viewer = Viewer(width=WIDTH, height=HEIGHT, url=CATALOG_URL)
viewer.send_table("video", pa.concat_batches(top_view_video.limit(1).drop("frame_index").collect()))
viewer.send_table("video_index", pa.concat_batches(top_view_video.limit(1).collect()))
viewer
```

If I drop the frame_index (which is none since static) then nothing shows up under streams.
If I don't drop the frame_index an index is added (which is wrong since static).

The other confounding factor is that I probably want to include our VideoFrameReference with this but that's not static so I'm not sure how to join our static and non-static columns into a table (maybe a hyper-table question).

Wumpf:
still curious about getting this somewhere but seems like there's no intention of moving it forward atm, so archiving it.

PR #11022, opened 2025-08-27
use time ceiling instead of cast in UDF

timsaucer:
When using nanosecond time stamps that are not easily coercible into microseconds, the `partition_url_with_timeref_udf` function will throw an exception. Instead we use the ceiling to convert to nearest microseconds.

The rest of this PR is improving the unit tests. Specifically:

- Add 20 files to the dataset instead of 2. This tests datafusion's partitioning scheme better because even with the default 14 threads we are guaranteed to get more than one partition in a single thread.
- Guarantees we are using timestamps that are not coercible to microseconds.

### DIFF
--- a/rerun_py/rerun_sdk/rerun/utilities/datafusion/functions/url_generation.py
+++ b/rerun_py/rerun_sdk/rerun/utilities/datafusion/functions/url_generation.py
@@ -100,6 +100,10 @@ def partition_url_with_timeref_udf(dataset: DatasetEntry, timeline_name: str) ->
         raise RerunOptionalDependencyError("datafusion", "datafusion")
 
     def inner_udf(partition_id_arr: pa.Array, timestamp_arr: pa.Array) -> pa.Array:
+        # The choice of `ceil_temporal` is important since this timestamp drives a cursor
+        # selection. Due to Rerun latest-at semantics, in order for data from the provided
+        # timestamp to be visible, the cursor must be set to a point in time which is
+        # greater than or equal to the target.
         timestamp_us = pa.compute.ceil_temporal(timestamp_arr, unit="microsecond")
 
         timestamp_us = pa.compute.strftime(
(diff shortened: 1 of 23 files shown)


rerun_py/rerun_sdk/rerun/utilities/datafusion/functions/url_generation.py
```diff
@@ -100,7 +100,7 @@ def partition_url_with_timeref_udf(dataset: DatasetEntry, timeline_name: str) ->
         raise RerunOptionalDependencyError("datafusion", "datafusion")
 
     def inner_udf(partition_id_arr: pa.Array, timestamp_arr: pa.Array) -> pa.Array:
-        timestamp_us = pa.compute.cast(timestamp_arr, pa.timestamp("us"))
+        timestamp_us = pa.compute.ceil_temporal(timestamp_arr, unit="microsecond")
```
jleibs, line 107:
```suggestion
        // The choice of `ceil_temporal` is important since this timestamp drives a cursor selection. Due to Rerun latest-at semantics, in order for data from the provided timestamp to be visible, the cursor must be set to a point in time which is greater than or equal to the target.
        timestamp_us = pa.compute.ceil_temporal(timestamp_arr, unit="microsecond")
```

jleibs (approved):

rerun_py/tests/unit/test_datafusion_tables.py
```diff
@@ -123,7 +123,8 @@ def test_df_count(server_instance: tuple[subprocess.Popen[str], DatasetEntry]) -
 
     count = dataset.dataframe_query_view(index="time_1", contents="/**").df().count()
 
-    assert count == 64
+    # We will need to update this if the underlying files are regenerated
+    assert count == 613
```
emilk: Maybe `assert count > 0` is better?

timsaucer: Yes, good point

rerun_py/rerun_sdk/rerun/utilities/datafusion/functions/url_generation.py
timsaucer: Added but with `#` not `//` ;)

PR #11023, opened 2025-08-27
Respect `CMAKE_BUILD_TYPE` when building the Rust code

emilk:
We previously would always compile the Rust-side of our C++ SDK in release mode. This meant that things like `debug_assert!` wasn't active in e.g. our C++ unit tests.

Our `pixi.toml` sets `--config Debug` and `--config Release` in the right places, and I've checked that those are indeed respected.

## Other smaller improvements
* Better error messages around string arguments
* Better error messages if there is a panic or SIGSEGV in the Rust code


Wumpf (approved):
nice!

Issue #11024, opened 2025-08-28
Apply backpressure when viewer cannot ingest data fast enough


emilk:
If the viewer cannot ingest data fast enough we currently end up in a situation where the viewer just grows its memory. This is particularly bad on Web, where we have limited memory.

## Plan
Make `re_smart_channel` smart enough to keep track of the size (in bytes!) of the data on its channel.
Then gRPC backpressure (i.e. throttle) if the channel grows "too heavy".

PR #11025, opened 2025-08-28
Remove `--drop-at-latency`

emilk:
### Related
* https://github.com/rerun-io/rerun/issues/11024

### What
This option has been broken for a while (a no-op).

A better solution is
* https://github.com/rerun-io/rerun/issues/11024


Wumpf (changes requested):
Despite being broken to begin with, I think this warrants a migration note. Otherwise lgtm!

Wumpf:
Huh. Interesting unrelated ci failure!

PR #11026, opened 2025-08-28
Decouple public and private gRPC APIs

teh-cmc:
The Rerun Cloud API is now completely standalone: it does not refer to any internal packages.

Users of the Rerun Cloud API will now only interact with a single package (`rerun.frontend.v1alpha1`) and API (`rerun.frontend.v1alpha.FrontendService`).

---

* Sibling: https://github.com/rerun-io/dataplatform/pull/1599


crates/store/re_protos/proto/rerun/v1alpha1/frontend.proto
```diff
@@ -355,3 +582,276 @@ message DoMaintenanceRequest {
...
+message Task {
+  // Unique identifier for the task
+  rerun.common.v1alpha1.TaskId id = 1;
+  // Type of the task
+  string task_type = 2;
+  // Task-type dependant data necessary to de-serialize the task
+  bytes task_data = 3;
+}
+
+// `SubmitTasksRequest` is the request message for submitting tasks
```
zehiko, line 602: what happens with Task endpoints?

teh-cmc: Huh.. nothing in particular? what do you mean?

zehiko: ah nevermind, as Andrea just clarified, I guess client still depends on some of these messages even though the Task API is gone from Frontend.

teh-cmc: The Task APIs are still there tho?

PR #11029, opened 2025-08-28
Add `tzdata` dependency to fix Windows CI

timsaucer:
This is related to the following CI failure: https://github.com/rerun-io/rerun/actions/runs/17284887058/job/49062341297

This adds tzdata, a 120kb dependency. With this users will have time zones support that is needed by pyarrow. There are other ways to get the time zone support, which is why it is not a required dependency.

### DIFF
--- a/rerun_py/tests/unit/test_datafusion_tables.py
+++ b/rerun_py/tests/unit/test_datafusion_tables.py
@@ -28,6 +28,14 @@ DATASET_FILEPATH = pathlib.Path(__file__).parent.parent.parent.parent / "tests" 
 
 @pytest.fixture(scope="session", autouse=True)
 def setup_windows_tzdata() -> None:
+    """
+    Adds timezone data on Windows machines.
+
+    Pyarrow requires timezone data to handle timestamps properly.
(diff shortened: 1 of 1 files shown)


timsaucer:
@rerun-bot full-check

timsaucer:
Demonstration this fixes nightly: https://github.com/rerun-io/rerun/actions/runs/17401565322/job/49398805978

The other failures in that nightly test are unrelated. I didn't touch any rust files, so this push could not possibly have broken those cargo test failures.

rerun_py/tests/unit/test_datafusion_tables.py
```diff
@@ -25,6 +26,12 @@
 DATASET_FILEPATH = pathlib.Path(__file__).parent.parent.parent.parent / "tests" / "assets" / "rrd" / "dataset"
 
 
+@pytest.fixture(scope="session", autouse=True)
+def setup_windows_tzdata() -> None:
+    if platform.system() == "Windows":
+        pa.util.download_tzdata_on_windows()
```
emilk: This should have a comment explaining why we need this on windows and not elsewhere

PR #11030, opened 2025-08-28
rerun.frontend[.FrontendService] -> rerun.cloud[.RerunCloudService]

teh-cmc:
Title.

This will break all the HTTP paths, and therefore all clients.

* Sibling: https://github.com/rerun-io/dataplatform/pull/1604


teh-cmc:
> `rerun.cloud.Service` would sound to me less redundant than `rerun.cloud.RerunCloudService`

IME a lot of tools won't show the fully-qualified name by default, so better be explicit :shrug:

PR #11027, opened 2025-08-28
Add snippet on loading MCAP

oxkitsune:
### Related

* Part of #11011 

### What

This adds a snippet that loads a new example mcap file, this will be used in the upcoming docs.

PR #11031, opened 2025-08-28
Use `opened_store_id_or_recommended()` in MCAP loader

oxkitsune:
### What

Update MCAP loader to prefer the opened store when available, falling back to the recommended store ID if none is opened.

This was causing the recording to split in two when loading an MCAP file using the SDK.

### DIFF
--- a/crates/store/re_data_loader/src/loader_mcap.rs
+++ b/crates/store/re_data_loader/src/loader_mcap.rs
@@ -162,7 +162,7 @@ fn load_mcap(
 ) -> Result<(), DataLoaderError> {
     re_tracing::profile_function!();
 
-    let store_id = settings.recommended_store_id();
+    let store_id = settings.opened_store_id_or_recommended();
 
     if tx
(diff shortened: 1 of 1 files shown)


oxkitsune:
> is that correct though? Sounds like with that everything becomes an import and we never open a new recording. Or does something already switch to a new recording for mcap? In that case plz document that & why at the changed callsite

This avoids everything being a new recording. When loading in a file programmatically (e.g. using `load_file_from_path` like in the snippet in #11027), we expect it to load the mcap file into the recording that we've created using `rr.init`. 

This changes the behavior slightly;
- when dragging an mcap file into the viewer, it will create a _new_ recording for that file, unless there's a recording that's already open.
- when loading an mcap file using the cli, it will create a _new_ recording regardless of whether there's one open already.
- when calling `load_file_from_path` it will respect the open recording instead of always creating a new recording.

Wumpf:
thank, I think that makes sense

Issue #11028, opened 2025-08-28
LargeBinary Migration Doesn't Update Datafusion Schema


ntjohnson1:
**Describe the bug**
If we've already registered a dataset on the cloud with `list<list<uint8>>` in the schema then when we try to access that column the schema expects the original type but we are returning `list<large_binary>`

**To Reproduce**
Simpler repro:
1. Using dataplatform update to point to local rerun version
2. `pixi run dev`
3. enter a pixi shell with rerun installed `pixi run -e examples`
```python
import rerun as rr
import os
CATALOG_URL = os.environ["REDAP_URI"]

client = rr.catalog.CatalogClient(CATALOG_URL, token=os.getenv('REDAP_TOKEN'))
dataset: rr.catalog.DatasetEntry = client.get_dataset_entry(name="droid:raw")

df = dataset.dataframe_query_view(index="real_time", contents="/thumbnail/camera/wrist").df()
print(df.schema())
bad_result = df.limit(1)
print(bad_result)
```

1. Launch local notebook from cloud repo and point to existing dataset
2. Query a column with `blob` type
```python
>> dataset.dataframe_query_view(index="real_time", contents="/thumbnail/camera/wrist").df().limit(1)
Exception: DataFusion error: Execution("Arrow error: Invalid argument error: column types must match schema types, expected List(Field { name: \"item\", data_type: List(Field { name: \"item\", data_type: UInt8, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }) but found List(Field { name: \"item\", data_type: LargeBinary, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }) at column index 4")
```

**Expected behavior**
We can query data even if migrating the format.

**Rerun version**
```console
rerun-cli 0.25.0-alpha.1+dev (base map_view nasm native_viewer oss_server release_no_web_viewer) [rustc 1.88.0 (6b00bc388 2025-06-23), LLVM 20.1.5] aarch64-apple-darwin (debug)
Video features: av1 default ffmpeg nasm serde
```
commit: 241df80b8d97f144aaefde067656257b712a9859

## Related
* https://github.com/rerun-io/rerun/pull/10875

ntjohnson1:
[...]

PR #11032, opened 2025-08-28
Migrate arrow schemas right after decoding them from protobuff

emilk:
### Related
* https://linear.app/rerun/project/cloud-support-for-evolution-of-data-schemas-59fb0b1a81ee/overview
* Solves https://github.com/rerun-io/rerun/issues/11028
* https://github.com/rerun-io/dataplatform/issues/1605
* Follows https://github.com/rerun-io/rerun/pull/10875

### What
This migrates all arrow schemas to the latest sorbet version right after decoding them off the gRPC wire


crates/store/re_sorbet/src/migrations/mod.rs
```diff
@@ -79,7 +76,11 @@ fn get_or_guess_version(batch: &RecordBatch) -> Result<semver::Version, Error> {
             // The migration code from `v0.0.2` to `v0.1.0` should be able handle this.
             Ok(semver::Version::new(0, 0, 2))
         } else {
-            Err(Error::MissingVersion)
+            // Rerun cloud schemas currently come without metadata,
+            // so we need to run the full migration just in case.
+            // TODO(emilk): Fix this: https://linear.app/rerun/issue/DPF-2097/send-sorbetversion-metdata-in-schema-requests
```
emilk:
```suggestion
            // TODO(rerun-io/dataplatform#1605): Always include version
```

teh-cmc (approved):
Regardless of all the other complexities we talked about and still have to address, I think that specific bit always makes sense anyhow.

teh-cmc:
This makes sense regardless of the big revert I believe

PR #11034, opened 2025-08-28
Use `ViewerImportUrl` for updating history, adds more unit tests

Wumpf:
_Mostly_ a refactor that got split out of other ongoing work

* integrate valid values of `LogDataSource` into `ViewerImportUrl` to avoid handling the same weird corner cases all the time
* `display_mode_to_content_url` is now integrated/split into `ViewerImportUrl` via `ViewerImportUrl::from_display_mode` and `ViewerImportUrl::sharable_url`
	* `sharable_url` already handles appending the a web viewer base to it which will be relevant in the next PR

But also:
* tons of new tests for these
* fix some corner cases with the example/welcome page where it was possible to get into a state where we add an invalid server & show the welcome page
* fix some formatting issue on intra recording urls

### DIFF
--- a/crates/viewer/re_viewer/src/open_url.rs
+++ b/crates/viewer/re_viewer/src/open_url.rs
@@ -278,7 +278,7 @@ impl ViewerImportUrl {
     /// Whenever possible you should provide a web viewer base URL so that the URL can be opened
     /// in the browser (this does *not* exclude native, web viewer URLs can still be opened there as well!)
     ///
-    /// This is roughly the inverse of [`Self::from_str`].
+    /// This is roughly the inverse of `Self::from_str`.
     #[allow(unused)] // TODO(rerun/dataplatform#1336): Only used on the web. About to change!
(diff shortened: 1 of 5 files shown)


crates/viewer/re_viewer/src/open_url.rs
```diff
@@ -121,14 +161,214 @@ fn parse_webviewer_url(url: &str) -> anyhow::Result<ViewerImportUrl> {
...
+    pub fn from_display_mode(
+        store_hub: &StoreHub,
+        display_mode: DisplayMode,
+    ) -> anyhow::Result<Self> {
+        match display_mode {
+            DisplayMode::Settings => {
+                // Not much point in updating address for the settings screen.
+                Err(anyhow::anyhow!("Can't share links to the settings screen."))
+            }
+            DisplayMode::LocalRecordings => {
```
abey79, line 193: nit

```suggestion

            DisplayMode::LocalRecordings => {
```

PR #11035, opened 2025-08-28
Universal copy direct link command that works on native & web

Wumpf:
### Related

* https://github.com/rerun-io/rerun/pull/11034
* Fixes https://github.com/rerun-io/dataplatform/issues/1336

### What

The `Copy Direct Link` command previously only worked on the Web and would only copy the base url unless you looked at an http recording. Now it copies the URL based on the current display mode (just like what we put in the addressbar on web).

Thanks to the unified logic link logic you can use this for aribtrary navigation on both web & native now

https://github.com/user-attachments/assets/0730ec74-847a-427b-8062-6ff6b5a74c90

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -1585,7 +1585,7 @@ impl App {
                     .success("Copied link to clipboard".to_owned());
             }
             Err(err) => {
-                re_log::error!("Can't produce a sharable link for the current screen: {err}");
+                re_log::error!("{err}");
             }
         }
     }


crates/viewer/re_viewer/src/app.rs
```diff
@@ -1552,27 +1556,38 @@ impl App {
...
+        )
+        .and_then(|content_url| content_url.sharable_url(base_url.as_ref()))
+        {
+            Ok(url) => {
+                self.egui_ctx.copy_text(url);
+                self.notifications
+                    .success("Copied link to clipboard".to_owned());
+            }
+            Err(err) => {
+                re_log::error!("Can't produce a sharable link for the current screen: {err}");
```
abey79: I trust you checked that the displayed error text is Good Enough(tm) (aka not a `tonic::Status`-level of garbage 😅)

Wumpf: uhoh. I better check on this

Wumpf: all good, the only error you can get here is `"Can't share more than one URL without a web viewer base URL"`
which zomg how can you get there even?
but also means I should only print the error

Issue #10866, opened 2025-08-12
Improve dataplatform navigation bar support further


Wumpf:
* Further improve upon [https://github.com/rerun-io/rerun/issues/10822](https://github.com/rerun-io/rerun/issues/10822)

Address bar should also update with working links for

- [X] tables & datasets (entries)
- [X] fragment: current selection, updated on click (https://github.com/rerun-io/rerun/pull/11153)
- [X] fragment: current time, updated on pause (https://github.com/rerun-io/rerun/pull/11153)
- [x] Keep time fragment from when it was last paused even when playing. (https://github.com/rerun-io/rerun/pull/11246)

Fragments should not append to the history, but just update the current link in place

PR #11038, opened 2025-08-28
Support sharing URLs for dataplatform datasets & tables

Wumpf:
### Related

* Part of #10866

### What

Web viewer URL bar reacts to this accordingly, [copy direct link](https://github.com/rerun-io/rerun/pull/11035) will pick it up and you can navigate to the resulting URL in all the places that take URLs (like command line or command pallette or `url` parameter)

https://github.com/user-attachments/assets/f9843337-70bc-4708-9256-5395686c0534

(ignore errors & missing loads. Sandbox is having some issues as of writing)

Also fixes an issue where navigation back to the server looked broken because it didn't select it as well as general missing redraws.

### DIFF
--- a/crates/viewer/re_context_menu/src/lib.rs
+++ b/crates/viewer/re_context_menu/src/lib.rs
@@ -384,8 +384,8 @@ trait ContextMenuAction {
                 }
                 Item::Container(container_id) => self.process_container(ctx, container_id),
                 Item::RedapServer(origin) => self.process_redap_server(ctx, origin),
-                Item::RedapEntry(entry_uri) => {
-                    self.process_redap_entry(ctx, &entry_uri.entry_id);
+                Item::RedapEntry(entry) => {
+                    self.process_redap_entry(ctx, &entry.entry_id);
                 }
             }
         }
--- a/crates/viewer/re_global_context/src/item.rs
+++ b/crates/viewer/re_global_context/src/item.rs
@@ -173,8 +173,8 @@ impl std::fmt::Debug for Item {
                 write!(f, "({view_id:?}, {instance_path}")
             }
             Self::Container(tile_id) => write!(f, "(tile: {tile_id:?})"),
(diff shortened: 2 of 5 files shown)


crates/viewer/re_global_context/src/lib.rs
```diff
@@ -71,7 +71,7 @@ pub enum DisplayMode {
     LocalTable(TableId),
 
     /// The Redap server/catalog/collection browser.
-    RedapEntry(re_log_types::EntryId),
```
abey79, line 74: If you haven't done so, you might want to grep for `RedapEntry(entry_id)` -> `RedapEntry(entry_uri)`, for the sake of reducing confusion.

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -102,9 +102,7 @@ impl View3DState {
         scene_bbox: &SceneBoundingBoxes,
         scene_view_coordinates: Option<ViewCoordinates>,
     ) {
-        // Mark as interaction since we want to stop doing any automatic interpolations,
-        // even if this is caused by a full reset.
-        self.last_eye_interaction = Some(Instant::now());
```
abey79: Did you mean to do that? Seems out of place.

abey79 (approved):
noice

crates/viewer/re_global_context/src/lib.rs
Wumpf: I did some searching, but now that you prompt me I really want to give it another pass just in case

crates/viewer/re_view_spatial/src/ui_3d.rs
Wumpf: thanks. That's from a different PR and slipped in

Wumpf:
ci failure is a known spurious issue where we loose the graphics adapter mid way. seems unrelated. Taking the risk 😄

Issue #10908, opened 2025-08-14
Remove deprecated methods from AnyValues


ntjohnson1:
Follow up to the deprecation https://github.com/rerun-io/rerun/issues/10879

Mostly here to tag from our zombie script. Searching for this issue number should highlight all the places with deprecation warnings.

PR #11045, opened 2025-08-29
Split AnyValues into AnyValues and DynamicArchetype

ntjohnson1:
### Related
* #10879

### What
Right now AnyValues is ambiguous. If we have two AnyValues with unique archetypes but the same component name we get collisions (since we don't combine the archetype with the name like we do for built ins). This deprecates the paths for archetypes to get attached to AnyValues (with [follow up to remove](https://github.com/rerun-io/rerun/issues/10908) in 0.26). Then creates a `DynamicArchetype` where we are explicit that we make the names unique like in built in types.

For both python and rust all the logic is in `DynamicArchetype` with private hooks to create one without an Archetype that we expose as `AnyValues`.


crates/store/re_types/src/archetype_builder.rs
```diff
@@ -0,0 +1,141 @@
...
+use crate::{
+    reflection::ComponentDescriptorExt as _, ArchetypeName, Component, ComponentDescriptor,
+    ComponentIdentifier, SerializedComponentBatch,
+};
+use re_types_core::{try_serialize_field, AsComponents, ComponentType, Loggable};
+
+/// A helper for logging a dynamically defined archetype to Rerun.
+/// component names will be modified in a way similar to Rerun
+/// internal types to avoid name collisions.
+//TODO(nick): Should this allow default or is THAT a footgun?
```
abey79: Well, if the question is worth asking, I'd default (pun unintended) to answering no.

crates/store/re_types/src/archetype_builder.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// Specifies an archetype name for this dynamically generated archetype.
+    #[inline]
+    pub fn new(archetype_name: impl Into<ArchetypeName>) -> Self {
+        Self {
+            archetype_name: Some(archetype_name.into()),
+            batches: Default::default(),
+        }
+    }
+
+    pub(crate) fn default_without_archetype() -> Self {
```
abey79: I'd drive the point home and call this `new_without_archetype`. It's also a more idiomatic naming imo.

crates/store/re_types/src/archetype_builder.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    use crate::components;
+
+    use super::*;
+
+    #[test]
+    fn with_archetype() {
+        let values = ArchetypeBuilder::new("MyExample")
+            .with_component::<components::Scalar>("confidence", [1.2f64, 3.4, 5.6])
+            .with_loggable::<components::Text>("homepage", "user.url", vec!["https://www.rerun.io"])
```
abey79: nit: it's slightly weird to use a `component` here. You could use `datatype::Utf8` instead, which kinds of drive the point home that this is generic data without attached semantics.

crates/store/re_types/src/archetype_builder.rs
```diff
@@ -0,0 +1,141 @@
...
+mod test {
+
+    use std::collections::BTreeSet;
+
+    use crate::components;
+
+    use super::*;
+
+    #[test]
+    fn with_archetype() {
```
abey79: might be nice to have one without archetype for completeness?

crates/store/re_types/src/archetype_builder.rs
```diff
@@ -0,0 +1,141 @@
...
+                    desc
+                },
+            },
+        );
+        self
+    }
+
+    /// Adds an existing Rerun [`Component`] to this archetype.
+    #[inline]
+    pub fn with_component<C: Component>(
```
abey79: I'm slightly uncomfortable with the naming here. All these methods add a `field` to the logged archetype, yet a single one actually has `field` in their name. I'm not sure  I'm 100% thrilled with the obvious suggest of `with_raw_field` (or `with_data_field`), `with_loggable_field`, `with_component_field`. Worth discussing with product minded people, since this ends up in the public API?

docs/content/howto/logging/custom-data.md
```diff
@@ -9,8 +9,8 @@ Rerun comes with many pre-built [Types](../../reference/types.md) that you can u
 For Python and Rust we have helpers for this, called `AnyValues`, allowing you to easily attach custom values to any entity instance.
 You find the documentation for these helpers here:
 
-* [`AnyValues` in Python](https://ref.rerun.io/docs/python/main/common/custom_data/)
-* [`AnyValues` in Rust](https://docs.rs/rerun/latest/rerun/struct.AnyValues.html)
+-   [`AnyValues` in Python](https://ref.rerun.io/docs/python/main/common/custom_data/)
+-   [`AnyValues` in Rust](https://docs.rs/rerun/latest/rerun/struct.AnyValues.html)
```
abey79: Worth linking to the newly introduced archetype builder as well?

rerun_py/rerun_sdk/rerun/archetype_builder.py
```diff
@@ -0,0 +1,429 @@
...
+    rr.log(
+        "some_type", rr.ArchetypeBuilder(
+            archetype="my_archetype",
+            confidence=[1.2, 3.4, 5.6],
+            description="Bla bla bla…",
+            # URIs will become clickable links
+            homepage="https://www.rerun.io",
+            repository="https://github.com/rerun-io/rerun",
+        ),
+    )
```
abey79: Also include a `send_columns` example?

abey79 (approved):
This is a great improvement! Did you have a chance to demo it? Would be great to have some feedback from product people.

I have a bunch of minor-ish comments. My main issue here is on the documentation side. This needs:
- coverage in the migration guide
- updated docs, at least incustom-data.md
- updated/new snippets to demo that stuff

I'm still ✅ because you might want to do that in a separate PR, but please have at least an issue with the correct milestone.

abey79:
Final though? What's the situation on the C++ front?

ntjohnson1:
>  Did you have a chance to demo it? Would be great to have some feedback from product people.

Not yet. I mostly wanted to get something concrete up for the technical side first since we've iterated in our discussions many times. I'll get your cleanup handled then try to run it by some people.

> Final though? What's the situation on the C++ front?

Final_final was mostly a joke based on how many times we've been discussing this.

crates/store/re_types/src/archetype_builder.rs
ntjohnson1: Sounds good. Can tag it into that discussion. This is mostly just mirroring the AnyValues existing API

crates/store/re_types/src/archetype_builder.rs
ntjohnson1: The only way to generate one without is to use the crate public `new_without_archetype` constructor that only exists for AnyValues and that has tests for that format.

crates/store/re_types/src/archetype_builder.rs
abey79: Sadly, the existing of something anywhere near `AnyValue` doesn't mean it's not utterly broken 😅

abey79:
> > Final though? What's the situation on the C++ front?
> 
> Final_final was mostly a joke based on how many times we've been discussing this.

That "?" was meant to be a ":", but I like that it shows how confused I am with everything C++ 😅

crates/store/re_types/src/archetype_builder.rs
ntjohnson1: I removed the default but will remove the todo.

ntjohnson1:
> My main issue here is on the documentation side.

I went ahead and addressed a bunch of the documentation concerns since I think that will hopefully make it easier to understand from a product perspective. I'll reach out for feedback for naming/API bits now.

rerun_py/rerun_sdk/rerun/archetype_builder.py
pweids: this example should have a `components={}` field instead, right?

ntjohnson1:
Ok I got some feedback on the design. https://rerunio.slack.com/archives/C08TEFXNHB2/p1756739649003149
Summary of changes from the functional review:
* Moved `ArchetypeBuilder` -> `DynamicArchetype`
* According to our [docs](https://rerun.io/docs/concepts/entity-component) we don't really have a `field` so I made everything reference `components`

The one thing that I'm not sure how to resolve is the fact I added links to the docs I added but all our docs are relative to main. Should I just remove the link and do a follow up PR?

rerun_py/rerun_sdk/rerun/any_value.py
```diff
@@ -1,220 +1,15 @@
...
-        value:
-            The data to be logged as a component.
-        drop_untyped_nones:
-            If True, any components that are None will be dropped unless they have been
-            previously logged with a type.
-
-        """
-        inst = cls(descriptor, value, drop_untyped_nones)
-        return ComponentColumn(descriptor, inst)
+from .dynamic_archetype import AnyBatchValue as AnyBatchValue, DynamicArchetype
```
abey79: weird import pattern here:
- `AnyBatchValue` is in `dynamic_archetype.py`
- imported in `any_value.py`
- re-imported as "pub" in `__init__.py`

I suggest moving `AnyBatchValue` to it's own file and streamline import pattern

abey79: PSA/if needed: type-related circular imports can be resolved by conditionally importing with `if TYPING:` and turning type hints into string (e.g. def fun() -> "Thing": ...`.

rerun_py/rerun_sdk/rerun/dynamic_archetype.py
```diff
@@ -0,0 +1,438 @@
...
+            elif hasattr(value, "as_arrow_array"):
+                self.pa_array = value.as_arrow_array()
+            else:
+                if pa_type is not None:
+                    if value is None:
+                        value = []
+                    self._maybe_parse_dlpack(value, pa_type)
+                    self._maybe_parse_string(value, pa_type)
+                    self._maybe_parse_scalar(value, pa_type)
+                    self._fallback_parse(value, pa_type, np_type)
```
abey79: I'm not a great fan of this stateful parsing logic, where each function needs to check for `self.pa_array is None`.

I think this logic would best be in standalone functions: `def _try_parse_dlpack(value, type) -> pa.Array | None: ...` along with a `def _parse_arrow_array(...) -> py.Array: ...` to combine them all.

Advantage: easier to write tests for these methods.

docs/content/reference/migration/migration-0-25.md
```diff
@@ -30,6 +28,18 @@ In the C++ and Python APIs, negative timeouts used to have special meaning. Now
...
+We used to encode blobs as `List<uint8>`, which was rather unidiomatic.
+Now they are instead encoded as `Binary`.
+Old data will be migrated on ingestion (zero-copy).
+
+Affects the following components:
+
+-   [`Blob`](https://rerun.io/docs/reference/types/components/blob)
+-   [`ImageBuffer`](https://rerun.io/docs/reference/types/components/image_buffer)
+-   [`VideoSample`](https://rerun.io/docs/reference/types/components/video_sample)
+
```
abey79: did we not revert that? I'm not sure what's the status for 0.25

docs/content/reference/migration/migration-0-25.md
```diff
@@ -38,3 +48,10 @@ Support for Python 3.9 is being deprecated. Python 3.9 is past end-of-life. See:
 In the next release, we will fully drop support and switch to Python 3.10 as the minimum supported version.
 
 See an overview for supported python versions [here](https://ref.rerun.io/docs/python/main/common#supported-python-versions).
+
+### `archetype` specification in `AnyValues`
+
+If `AnyValues` had the same `field_name` but unique `archetypes` then the view would disambiguate them but the `dataframe` api wouldn't.
+We split `AnyValues` into `AnyValues` with no archetype and `DynamicArchetype` requiring an `archetype` where the name in the `dataframe` api is made unique by combining
+the archetype with the field name similar to how built-in components are handled.
+In the next release we will remove the ability to specify an `archetype` when creating `AnyValues` to finalize the split.
```
abey79: I think we can improve the clarity of this bit. I'm not sure I understand all of it tbh. Also, we dont usually quote the "dataframe API".

```suggestion
Previously, logging two `AnyValues` with the same field name but different archetype name  under the same entity would lead to an inconsistency where the viewer would disambiguate them, but not the dataframe API.

TODO: small example snippet?

To address that, we split this functionality into two utilities:
- `AnyValues`, which has no archetype name
- `DynamicArchetype`, which requires an archetype name 

When using `DynamicArchetype`, the dataframe API will include the archetype the column names (similar to how built-in components are handled), which reduces the possibility for ambiguity.

In the next release we will remove the ability to specify an `archetype` when creating `AnyValues` to finalize the transition.
```

docs/content/howto/logging/custom-data.md
```diff
@@ -20,20 +20,57 @@ rr.log(
...
+
+rec.log("my_entity", &any_values)?;
+```
+
+If your values should be grouped together and that grouping isn't referred to from many places that need to stay aligned we have a helpers for this called, `DynamicArchetype` which adds some structural grouping to multiple values.
+
+You find the documentation for these helpers here:
+
+-   [`DynamicArchetype` in Python](https://ref.rerun.io/docs/python/main/common/custom_data/)
+-   [`DynamicArchetype` in Rust](https://docs.rs/rerun/latest/rerun/struct.DynamicArchetype.html)
```
abey79: This how we silence link errors for links that don't yet exist. We then remove these markers during the release process.

```suggestion
-   [`DynamicArchetype` in Rust](https://docs.rs/rerun/latest/rerun/struct.DynamicArchetype.html?speculative-link)
```

docs/content/howto/logging/custom-data.md
```diff
@@ -20,20 +20,57 @@ rr.log(
...
+    .with_component::<rerun::components::Text>("description", vec!["Bla bla bla…"]);
+
+rec.log("my_entity", &any_values)?;
+```
+
+If your values should be grouped together and that grouping isn't referred to from many places that need to stay aligned we have a helpers for this called, `DynamicArchetype` which adds some structural grouping to multiple values.
+
+You find the documentation for these helpers here:
+
+-   [`DynamicArchetype` in Python](https://ref.rerun.io/docs/python/main/common/custom_data/)
```
abey79: use a (speculative) direct link to the actual object?

docs/content/howto/logging/custom-data.md
```diff
@@ -9,8 +9,8 @@ Rerun comes with many pre-built [Types](../../reference/types.md) that you can u
 For Python and Rust we have helpers for this, called `AnyValues`, allowing you to easily attach custom values to any entity instance.
 You find the documentation for these helpers here:
 
-* [`AnyValues` in Python](https://ref.rerun.io/docs/python/main/common/custom_data/)
-* [`AnyValues` in Rust](https://docs.rs/rerun/latest/rerun/struct.AnyValues.html)
+-   [`AnyValues` in Python](https://ref.rerun.io/docs/python/main/common/custom_data/)
```
abey79: same

abey79 (approved):
Great work!

docs/content/reference/migration/migration-0-25.md
ntjohnson1: Must havebeen a bad merge. Will prune (I checked it's not on main) so people can re-add it if we put it back in.

rerun_py/rerun_sdk/rerun/any_value.py
ntjohnson1: I wasn't sure what our attempts to be backwards compatible were. I saw some weird re-import patterns other places (which I assumed were to keep import locations stable). I can move to it's own file and make things direct.

rerun_py/rerun_sdk/rerun/dynamic_archetype.py
ntjohnson1: Makes sense. This was just copied over from AnyValues where this was already a small refactor on a big if/else block. I can update this when I move AnyBatchValues later this morning.

rerun_py/rerun_sdk/rerun/any_value.py
abey79: Ah, I see what you mean. Well, we re-export everything public from `__init__.py`. The only "public" submodules are `catalog`, `dataframe`, and `blueprint` (top of my head). I'd consider any other submodule import incorrect. This is sadly not reflected by our naming conventions though. We should prefix all non-public submodule with  `_`.

PR #11046, opened 2025-08-29
always allow blueprint panel resizing

gavrelina:
now one needs to have 4 recordings open to be able to resize the blueprint panel. 
especially with a long list for catalog this rule doesnt work anymore as it's quickly too long and annoying. 
Hence the change!


Wumpf (changes requested):
It's behaves weirdly if there's only one item in the examples. I think that's what this was for


https://github.com/user-attachments/assets/45552b6e-c8aa-4591-8b6c-468f8c05bd7a

Wumpf:
@gavrelina _bump_

Issue #4812, opened 2024-01-15
Allow tracking arbitrary entities in spatial views


Wumpf:
In
* https://github.com/rerun-io/rerun/pull/4799 

all the necessary code for tracking (i.e. continuously refocusing on) entities was added, but we ended up not enabling it on double click:
The issue is that it's very intransparent when tracking is happening and there's nothing indicating the state:

* Make tracked entity an actual part of the camera that can be set from blueprint components
* Indicate ad-hoc tracking somehow?
* add a button upon selecting an entity to make it tracking (and revisit the decision to only foucs on double click)
   * related to #810

Wumpf:
the PR that fixed this was reverted

IsseW:
Re-implemented in https://github.com/rerun-io/rerun/pull/11554

PR #11048, opened 2025-08-30
Entity tracking feature

Gentlegg:
- eye.rs: new variable to track the ignored inputs so the update function in ui3d.rs wont escape the tracking
 - ui.rs: extra label to display if an entity is being tracked
 - ui_3d.rs: alt + double left clicking an entity makes the view_eye to track that entity without interpollating

### Related
* #4812

### DIFF
--- a/crates/viewer/re_view_spatial/src/eye.rs
+++ b/crates/viewer/re_view_spatial/src/eye.rs
@@ -192,6 +192,8 @@ pub struct ViewEye {
     eye_up: Vec3,
 
     velocity: Vec3,
+
+    track_ignore_input: bool,
 }
 
 impl ViewEye {
@@ -212,6 +214,7 @@ impl ViewEye {
             fov_y: Eye::DEFAULT_FOV_Y,
             eye_up,
             velocity: Vec3::ZERO,
+            track_ignore_input: false,
         }
     }
 
@@ -239,6 +242,10 @@ impl ViewEye {
             Eye3DKind::FirstPerson => None,
             Eye3DKind::Orbital => Some(self.center),
         }
+    }
+
+    pub fn get_track_ignore_input_pressed(&self) -> bool {
+        self.track_ignore_input
     }
 
     /// If in orbit mode, how far from the orbit center are we?
@@ -326,6 +333,7 @@ impl ViewEye {
                 // and even then it's not a big deal.
                 eye_up: self.eye_up.lerp(other.eye_up, t).normalize_or_zero(),
                 velocity: self.velocity.lerp(other.velocity, t),
+                track_ignore_input: false,
             }
         }
     }
@@ -391,6 +399,8 @@ impl ViewEye {
         // Otherwise we flicker in and out of "has interacted" too quickly.
         let mut did_interact = response.drag_delta().length() > 0.0;
 
+        self.track_ignore_input = false;
+
         if response.drag_delta().length() > drag_threshold {
             let roll = response.dragged_by(ROLL_MOUSE)
                 || (response.dragged_by(ROLL_MOUSE_ALT)
@@ -401,8 +411,10 @@ impl ViewEye {
                 if let Some(pointer_pos) = response.ctx.pointer_latest_pos() {
                     self.roll(&response.rect, pointer_pos, response.drag_delta());
                 }
+                self.track_ignore_input = true
             } else if response.dragged_by(ROTATE3D_BUTTON) {
                 self.rotate(response.drag_delta());
+                self.track_ignore_input = true
             } else if response.dragged_by(DRAG_PAN3D_BUTTON) {
                 // The pan speed is selected to make the panning feel natural for orbit mode,
                 // but it should probably take FOV and screen size into account
@@ -494,6 +506,7 @@ impl ViewEye {
                 if f32::MIN_POSITIVE < new_radius && new_radius < 1.0e17 {
                     self.orbit_radius = new_radius;
                     did_interact = true;
+                    self.track_ignore_input = true;
                 }
             }
             Eye3DKind::FirstPerson => {
@@ -501,6 +514,7 @@ impl ViewEye {
                 let delta = (zoom_factor - 1.0) * speed;
                 self.center += delta * self.fwd();
                 did_interact = true;
+                self.track_ignore_input = true;
             }
         }
 
--- a/crates/viewer/re_view_spatial/src/ui.rs
+++ b/crates/viewer/re_view_spatial/src/ui.rs
@@ -153,6 +153,12 @@ impl SpatialViewState {
                 self.state_3d.set_spin(spin);
             }
         }
+
+        let mut tracked_entity_path = String::from("None");
+        if let Some(tracked_entity) = self.state_3d.tracked_entity.clone() {
+            tracked_entity_path = tracked_entity.to_string();
+        }
+        ui.label(format!("Tracked entity: {tracked_entity_path}"));
     }
 
     pub fn fallback_opacity_for_image_kind(&self, kind: ImageKind) -> f32 {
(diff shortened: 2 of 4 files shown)


crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -241,6 +244,10 @@ impl ViewEye {
         }
     }
 
+    pub fn get_track_ignore_input_pressed(&self) -> bool {
+        self.track_ignore_input
+    }
```
Wumpf: I have the context of this PR right now ofc, but it's very unclear what this means from the outside imho. Needs a doc string and maybe a better name (not sure yet, I'll think about it 🤔 )

Also generally in rust getters don't start with `get_`

crates/viewer/re_view_spatial/src/ui.rs
```diff
@@ -153,6 +153,12 @@ impl SpatialViewState {
                 self.state_3d.set_spin(spin);
             }
         }
+
+        let mut tracked_entity_path = String::from("None");
+        if let Some(tracked_entity) = self.state_3d.tracked_entity.clone() {
+            tracked_entity_path = tracked_entity.to_string();
+        }
+        ui.label(format!("Tracked entity: {tracked_entity_path}"));
```
Wumpf: let's use `ui.horizontal` and put the label next to a "standard entity button" via `item_ui::entity_path_button`  if there is one

nice to have that existing property exposed btw! 👍

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -408,6 +410,10 @@ pub fn help(os: egui::os::OperatingSystem) -> Help {
             ),
         )
         .control("Focus", ("double", icons::LEFT_MOUSE_CLICK, "object"))
+        .control(
+            "Track",
+            ("alt", "+", "double", icons::LEFT_MOUSE_CLICK, "object"),
```
Wumpf: it's not alt on mac! should use `IconText` etc. for it

Wumpf (changes requested):
love you wired up the existing fragments of what's already there for tracking entities
Tried it out, works better than I thought it would :)

Just some suggestion improvements on naming & ui.

But beyond that I'm not sure we need the alt modifier shortcut here since it feels weirdly inconsistent or at least confusing now:
double clicking cameras tracks them regardless of the alt/option but everything else still gets focus but only tracks if the modifier is pressed.
It's probably fine to just have tracking always be happening on any focus event? The other advantage of that would be that you can initiate tracking by double clicking in the blueprint panel (just like you can for cameras). Thoughts?

Wumpf:
some ci failures here:
https://github.com/rerun-io/rerun/actions/runs/17348184271/job/49419675556?pr=11048

image for image comparision tests need updating, instructions for this can be found here https://github.com/rerun-io/rerun/blob/main/CONTRIBUTING.md#image-comparison-tests

Gentlegg:
> But beyond that I'm not sure we need the alt modifier shortcut here since it feels weirdly inconsistent or at least confusing now: double clicking cameras tracks them regardless of the alt/option but everything else still gets focus but only tracks if the modifier is pressed. It's probably fine to just have tracking always be happening on any focus event? The other advantage of that would be that you can initiate tracking by double clicking in the blueprint panel (just like you can for cameras). Thoughts?

Thanks for the feedback! I fixed the things that you pointed out.
I wanted to keep the focus on double click because its just an easy navigation option, but its a fair point its really similar if you just want to focus on something that is not moving.
On the blueprint panel you can focus as well with alt just like in the view.

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -412,7 +412,13 @@ pub fn help(os: egui::os::OperatingSystem) -> Help {
         .control("Focus", ("double", icons::LEFT_MOUSE_CLICK, "object"))
         .control(
             "Track",
-            ("alt", "+", "double", icons::LEFT_MOUSE_CLICK, "object"),
+            (
+                IconText::from_modifiers(os, RuntimeModifiers::slow_down(&os)),
```
Wumpf: nice try, but that's extra wrong now ;)
* this doesn't have the `slow_down` semantic
* `slow_down` is ctrl on mac and alt on Windows/Linux

what you want os `Modifiers::ALT` which on Windows/Linux is `alt` and on Mac is `option` (or rather the weird option symbol that's used there)

Wumpf:
> I wanted to keep the focus on double click because its just an easy navigation option, but its a fair point its really similar if you just want to focus on something that is not moving.

Correct me, but with the change currently the focus _is_ still there for both. The only drawback with my suggestion of dropping `alt` is that you'd always track unasked.

Maybe we should make it the opposite 💡: when double clicking something we always have track + (viewer-wide) focus, but if alt is pressed we don't track which then applies the same to cameras and other entities.
That way the behavior is less situational (no camera/non-camera distinction) and the more powerful track one is easy to discover with an opt-out.

Gentlegg:
> Maybe we should make it the opposite 💡: when double clicking something we always have track + (viewer-wide) focus, but if alt is pressed we don't track which then applies the same to cameras and other entities. That way the behavior is less situational (no camera/non-camera distinction) and the more powerful track one is easy to discover with an opt-out.

Sounds good, i'll erease the camera distinction, maybe except the drag threshold and swap the alt behavioure. I didn't change the way focus/track worked for cameras because i didn't think anyone would just focus a pinhole but an option for that couldn't hurt.

crates/viewer/re_view_spatial/src/ui_3d.rs
Gentlegg: Fixed it in the latest commit 👌

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -192,6 +192,9 @@ pub struct ViewEye {
     eye_up: Vec3,
 
     velocity: Vec3,
+
+    /// Tracks if any user input happened that shouldn't interrupt the entity tracking behavioure.
```
```suggestion
    /// Tracks if any user input happened that shouldn't interrupt the entity tracking behavior.
```
😄

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -588,17 +592,19 @@ impl SpatialView3D {
...
-                        entity_path,
-                        &state.bounding_boxes,
-                        space_cameras,
-                    );
+                    state
+                        .state_3d
+                        .track_entity(entity_path, &state.bounding_boxes);
+                    if let Some(tracked_camera) = find_camera(space_cameras, &entity_path) {
+                        state.state_3d.interpolate_to_eye(tracked_camera);
+                    }
```
Wumpf: this is aaaalmost what I've meant (I've struggled to express it properly): for a non-alt click we still want to call interpolate_eye_to_entity` for non-cameras and `interpolate_to_eye` for cameras (the later is already the case in your most recent version.

Also the `ctx.recording().is_logged_entity(entity_path)` check is in a bit of a weird place


So something like this. (untested)
```rs
if ctx.recording().is_logged_entity(entity_path) {
    if let Some(tracked_camera) = find_camera(space_cameras, &entity_path) {
        state.state_3d.interpolate_to_eye(tracked_camera);
    } else {
        state.state_3d.interpolate_eye_to_entity(entity_path, &state.bounding_boxes, space_cameras);
    }

    // Additionally track the entity if `alt` is NOT pressed.
    // (Note that this means slightly different things for cameras & regular entities)
    if !response.ctx.input(|i| i.modifiers.alt) {
        state.state_3d.track_entity(entity_path, &state.bounding_boxes);
    }
}
```

does this (still?) make sense?

Gentlegg: yep, its clear now, i just did some minor changes but i noticed if you want to track another moving entity while you're allready tracking something the interpollation is a little late :/

Wumpf (approved):
thanks for bearing with all the comments & iterations!
Looks good now, shipping time :)

Wumpf:
Reverted in https://github.com/rerun-io/rerun/pull/11147 as discussed

Issue #11049, opened 2025-08-31
Changing albedo factor on a mesh causes massive vram memory leaks


Wumpf:
Might be another cache eviction bug (we create new entires on the MeshCache for each change), but that would also be a bit surprising since we tested this recently, so maybe it's something else. Either way it's clear that the newly produces `GpuMaterial`s (or maybe even more) are accumulating. Could also be a bug in `re_renderer`'s texture & buffer retention mechanism.

https://github.com/user-attachments/assets/f52bf747-62b3-49ba-bc57-a191b25559e7

Naturally, the higher your framerate the quicker it gets bad (not helping that I switched to a 240hz panel...)

emilk:
It would be very useful with a part of the Memory Panel showing all our caches (e.g. MeshCashe), with stats about number of entries and size (bytes). We could even add `fn ui` to `trait Cache`

emilk:
@Wumpf I gave this a spin, and… that counter is not to be trusted. I pretty easily get up to 60+ GiB of VRAM… in a browser. I think not :)

emilk:
…also Query Caches is at 5.9 GiB in the browser. So lots of things are wrong 😭 

<img width="1201" height="288" alt="Image" src="https://github.com/user-attachments/assets/a82b8bbd-3997-456f-b546-8db3ad6090bf" />

Wumpf:
oops, those are surely also very annoying bugs then!

I didn't add that to the description, but I got the tab to crash/freeze before on my Windows machine @ Firefox @ WebGPU. But can't reproduce this on my mac right now 🤔 . Need to check if this was specific to my branch maybe

IsseW:
I added some more information to the mesh cache section, and the amount of meshes inside one row did increase:

<img width="445" height="216" alt="Image" src="https://github.com/user-attachments/assets/f1be62a7-7d2f-4353-9fe5-57ae48cb73c6" />

Wumpf:
so the mesh cache _is_ leaking after all
phew, that's much easier to debug than the renderer resources :)

emilk:
Turns our visualizing stuff is useful 😆

PR #11055, opened 2025-09-01
Add debug information about caches on the memory panel

IsseW:
### Related

To help debug what is going on in #11049

### What

Adds a header to the memory panel where different cache types can display UI.

Screenshot 2025-09-01 at 18 25 51

### Still TODO

Implement the ui for the various caches.


Wumpf (changes requested):
nice, great start!

for separation of ui code from the rest of the logic I'd really like to keep the ui code out of the caches though. I realize that makes it a bit harder to do flexible reports (like "number of meshes") but that way we can also look at accumulated memory from caches.
I.e. I'm thinking we could report something back like
```
struct CacheMemoryReport {
    bytes_cpu: usize,
    bytes_gpu: usize,
    /// helpful doc string that explains wth this is
    detailed_properties: BTreeMap<String, usize>, 
}
```
and then build a ui from that generically 🤔 . What do you think?

IsseW:
> or separation of ui code from the rest of the logic I'd really like to keep the ui code out of the caches though.

That makes sense, could also achieve the separation by having a secondary trait for it. But I think a unified UI for it would be good because there was some code-duplication between the different `ui` implementations.

crates/viewer/re_renderer/src/mesh.rs
```diff
@@ -200,6 +200,16 @@ pub struct GpuMesh {
     pub materials: SmallVec<[GpuMaterial; 1]>,
 }
 
+impl GpuMesh {
+    /// Returns the approximate byte size this `GpuMesh` contributes.
+    ///
+    /// The total buffer sizes are divided by how many strong references
+    /// they have.
```
Wumpf: outdated, ended up not doing that because it's messy with the way we ref count these

crates/viewer/re_view_spatial/src/mesh_cache.rs
```diff
@@ -97,8 +97,40 @@ impl Cache for MeshCache {
...
+                    .sum();
+                full_bytes_gpu += bytes_gpu;
+                CacheMemoryReportItem {
+                    item_name: row_id.short_string(),
+                    bytes_cpu: meshes.total_size_bytes(),
+                    bytes_gpu: Some(bytes_gpu),
+                }
+            })
+            .collect();
+        items.sort_by(|a, b| a.item_name.cmp(&b.item_name));
```
Wumpf, line 160: nice catch on avoiding unstable hashmap ordering 😄

crates/viewer/re_view_spatial/src/proc_mesh.rs
```diff
@@ -513,8 +521,21 @@ impl Cache for SolidCache {
...
+                    .values()
+                    .map(|mesh| mesh.as_ref().map_or(0, |m| m.gpu_mesh.gpu_byte_size()))
+                    .sum(),
+            ),
+            per_cache_item_info: Vec::new(),
+        }
+    }
+
+    fn name(&self) -> &'static str {
+        "Solid Cache"
```
Wumpf: in context that name makes sense, but if I see this in the panel a week from now I'd be confused
Maybe `Proc Mesh Solids`?

On that note, applies to everywhere, having cache in the name is a bit redundant. On the other hand nice because it resolves better to the type name in the code. No strong opinion either way (your call, feel free to ignore that aspect 🤷)

crates/viewer/re_viewer/src/ui/memory_panel.rs
```diff
@@ -82,20 +82,94 @@ impl MemoryPanel {
...
+            egui::ScrollArea::vertical()
+                .max_height(200.0)
+                .id_salt(name)
+                .show(ui, |ui| {
+                    egui::Grid::new(format!("{name} grid"))
+                        .num_columns(3)
+                        .show(ui, |ui| {
+                            ui.label(egui::RichText::new("Name").underline());
+                            ui.label(egui::RichText::new("Memory (cpu)").underline());
+                            ui.label(egui::RichText::new("Memory (gpu)").underline());
```
Wumpf: nit: inconsistent to the rest

```suggestion
                            ui.label(egui::RichText::new("Memory (CPU)").underline());
                            ui.label(egui::RichText::new("Memory (GPU)").underline());
```

Wumpf (approved):

crates/viewer/re_view_spatial/src/proc_mesh.rs
IsseW: Renamed them a bit

PR #11058, opened 2025-09-01
Update video documentation to clarify MP4 export process (remuxing)

pablovela5620:
Based on the original discussion on Slack https://rerunio.slack.com/archives/C04UBTMMGJK/p1756500919499429

Users searching for “rrd to mp4” may not recognize the term “remuxing.” This PR adds a beginner‑friendly heading (“Convert RRD to MP4 (remuxing)”) to the video docs, improving search/discoverability without changing content.


Wumpf:
I've set `deploy docs` so live docs should be updated with this right away

PR #11060, opened 2025-09-02
Add `.editorconfig` with correct settings for `.ron` files

abey79:
Add a `.editorconfig` file with proper settings for .ron


abey79:
> we're starting with `.editorconfig`? sure why not, at least those are widely understood these days

alternative was for me to have it in my global .gitignore. I figured it might be useful for other jet brain users 🤷🏻

Wumpf:
.editorconfig is actually recongized by most editors these days
https://editorconfig.org/

Issue #9949, opened 2025-05-09
Support to mcap files


patrickelectric:
**Is your feature request related to a problem? Please describe.**
When I run a mcap file it says:
```
rerun /tmp/demo_2025-05-08_17-58-25.mcap               
[2025-05-09T18:14:55Z INFO  re_data_loader::load_file] Loading "/tmp/demo_2025-05-08_17-58-25.mcap"…
Error: "/tmp/demo_2025-05-08_17-58-25.mcap" -> No data-loader support for "/tmp/demo_2025-05-08_17-58-25.mcap"
```

**Describe the solution you'd like**
Be able to run mcap files

**Describe alternatives you've considered**
Not that I'm aware

**Additional context**
Just recorded a file in https://mcap.dev/ and rerun failed to play

patrickelectric:
references: https://rerun.io/blog/rosbag

Wumpf:
Can't believe we didn't have an issue for this already! This is really high on our priority list. Stay tuned!

patrickelectric:
Just for curiosity.. and my sanity.. is there any timeline for this or prediction ? 
We are working to integrate zenoh + mcap into our system and would like to use rerun

nikolausWest:
Current plan is to include an early/incomplete/experimental mcap dataloader in the next release. The goal is to get it ready to handle external contributions in the form of message conversions, testing, and bug fixes.

patrickelectric:
Thanks @nikolausWest! I'll certainly help with that

nikolausWest:
@patrickelectric amazing, thanks!

git-pi-e:
Foxglove also has a native protobuf support, so MCAP and protobuf together would also be great. It will allow the data from protobuf messages, in this case from MCAP files, to be visualised directly onto the viewer.

pweids:
@patrickelectric @yinguobing @git-pi-e 

letting you know that we decided to prioritize bringing MCAP as a built-in data loader into the viewer. The first iteration is scheduled for the next release: 0.25. We are actively working on it, but will be descoping some aspects, especially around re-mapping custom messages. With that said, we are soliciting feedback for what is important to you in your MCAP workflows. Would appreciate if you can go a bit deeper into your workflows/must-haves/use cases. Happy to chat in other formats too!

git-pi-e:
That's fantastic news!

git-pi-e:
[...]

PR #11062, opened 2025-09-02
Add archetypes for MCAP metadata

oxkitsune:
### Related

* Part of #9949 

### What

This introduces dedicated types for the (previously) manually logged MCAP data (channel, stats, schema and raw messages). 
I've tried to re-use as many existing components/datatypes as possible, however this does introduce some new ones.


crates/store/re_types/definitions/rerun/archetypes/mcap_channel.fbs
```diff
@@ -0,0 +1,51 @@
...
+/// for all messages published to that topic.
+///
+/// See also [archetypes.McapMessage] for individual messages within a channel,
+/// [archetypes.McapSchema] for the data structure definitions, and the
+/// [MCAP specification](https://mcap.dev/) for complete format details.
+///
+/// \example archetypes/mcap_channel_simple !api title="Simple MCAP channel"
+table McapChannel (
+  "attr.docs.category": "MCAP",
+  "attr.rerun.state": "unstable",
```
emilk, line 19: We could add a `TODO(#…)` here so we remember to stabilize it when it is ready (for some definition of "ready"

crates/store/re_types/definitions/rerun/archetypes/mcap_channel.fbs
```diff
@@ -0,0 +1,51 @@
...
+  ///
+  /// Channel IDs must be unique within a single MCAP file and are used to associate
+  /// messages with their corresponding channel definition.
+  id: rerun.components.ChannelId ("attr.rerun.component_required", order: 1000);
+
+  /// The topic name that this channel publishes to.
+  ///
+  /// Topics are typically hierarchical paths (e.g., "/sensors/camera/image") that
+  /// categorize and organize different data streams within the system.
+  topic: rerun.components.Text ("attr.rerun.component_required", order: 2000);
```
emilk: How closely related are topics and entity paths?

crates/store/re_types/definitions/rerun/archetypes/mcap_channel.fbs
```diff
@@ -0,0 +1,51 @@
...
+
+  /// The encoding format used for messages in this channel.
+  ///
+  /// Common encodings include:
+  /// * `ros1` - ROS1 message format
+  /// * `cdr` - Common Data Representation (CDR) message format
+  /// * `protobuf` - Protocol Buffers
+  /// * `json` - JSON encoding
```
emilk: This would be a good place to explain what ROS2 does

crates/store/re_types/definitions/rerun/archetypes/mcap_channel.fbs
emilk: here and everywhere

crates/store/re_types/definitions/rerun/archetypes/mcap_schema.fbs
```diff
@@ -0,0 +1,53 @@
...
+  "attr.rust.derive": "PartialEq",
+  "attr.docs.unreleased"
+) {
+  // --- Required ---
+
+  /// Unique identifier for this schema within the MCAP file.
+  ///
+  /// Schema IDs must be unique within a single MCAP file and are used by channels
+  /// to reference the schema that defines their message structure.
+  id: rerun.components.ChannelId ("attr.rerun.component_required", order: 1000);
```
emilk: Is it a unique identifier for the _schema_ or for the _channel_? There is conflicting signals here

crates/store/re_types/definitions/rerun/components/channel_id.fbs
```diff
@@ -0,0 +1,19 @@
...
+  "attr.arrow.transparent",
+  "attr.docs.unreleased",
+  "attr.python.aliases": "int",
+  "attr.python.array_aliases": "int, npt.NDArray[np.uint8], npt.NDArray[np.uint16], npt.NDArray[np.uint32], npt.NDArray[np.uint64]",
+  "attr.rust.derive": "Copy, PartialEq, Eq, PartialOrd, Ord",
+  "attr.rust.repr": "transparent",
+  "attr.rust.custom_clause":
+    'cfg_attr(feature = "serde", derive(::serde::Serialize, ::serde::Deserialize))'
+) {
+  id: rerun.datatypes.ChannelId (order: 100);
```
emilk: I wonder if we really should add a new datatype here - I think it would probably be better with `rerun.datatypes.UInt16` here

emilk (approved):
Very nice! I appreciate the in-depth docstrings

crates/store/re_types/definitions/rerun/components/channel_id.fbs
oxkitsune: Yes, we can do that. I originally added a new datatype because the codegen was complaining about missing serde attributes. But I can just remove them from this type as well.

crates/store/re_types/definitions/rerun/archetypes/mcap_channel.fbs
oxkitsune: Topics are separate from Rerun's entity paths, but often map to them. I clarified this in the docs:

```rs
/// The topic name that this channel publishes to.
///
/// Topics are hierarchical paths from the original robotics system (e.g., "/sensors/camera/image")
/// that categorize and organize different data streams.
/// Topics are separate from Rerun's entity paths, but they often can be mapped to them.
```

crates/store/re_types/definitions/rerun/archetypes/mcap_schema.fbs
oxkitsune: Schema IDs must be unique within an MCAP file as they're used in channel records, multiple channels can refer to the same schema.
I clarified the docs a bit, and added a new component: `rerun.components.SchemaId`.

Does it make sense to turn `SchemaId` and `ChannelId` into a single component `Id`?

emilk: Having this be typesafe with `SchemaId` and `ChannelId` seems like the right choice 👍

emilk (approved):

PR #11065, opened 2025-09-02
Revert `BinaryArray` migration (#11039)

andrea-reale:
(cherry picking from https://github.com/rerun-io/rerun/pull/11039)

### Related
* Reverts #10875
* Reverts #11005
* Fixes https://github.com/rerun-io/rerun/issues/11028
* Closes https://github.com/rerun-io/rerun/pull/11032
* Re-opens https://github.com/rerun-io/rerun/issues/10929
* Re-opens https://github.com/rerun-io/rerun/issues/10973

### What
Reverts the the change from `List<u8>` to `Binary` because of Rerun Cloud migration woes.

Hopefully temporarily.

PR #11067, opened 2025-09-02
Re-apply some reverted things

emilk:
### Related
* https://github.com/rerun-io/rerun/pull/11065
* https://github.com/rerun-io/rerun/issues/11028

### What
We do not yet support migrating schemas on cloud (https://github.com/rerun-io/rerun/issues/11028) so we had to revert changing the datatype of `Blob` from `List<u8>` to `LargeBinary`.

However, in the revert we threw away a lot of good baby with the bath water.

This PR selectively cherry-pick back a bunch of things that we do want in, without actually changing any data formats.

This lets us:
* Get in some good stuff
* Reduce merge conflicts on other PRs to main
* Reduce future merge conflicts when we want to change the datatype again

NOTE: everything in here has already been reviewed before

### TODO
* [x] `@rerun-bot full-check`

### DIFF
--- a/crates/build/re_types_builder/src/codegen/cpp/array_builder.rs
+++ b/crates/build/re_types_builder/src/codegen/cpp/array_builder.rs
@@ -66,6 +66,11 @@ fn arrow_array_builder_type_and_declaration(
             );
             ident
         }
+        Type::Binary => {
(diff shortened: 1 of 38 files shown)


emilk:
@rerun-bot full-check

PR #11069, opened 2025-09-02
MCAP ROS2 loader test

aedm:


crates/store/re_data_loader/tests/mcap_test.rs
```diff
@@ -0,0 +1,166 @@
...
+            linear_acceleration: ros2msg::definitions::geometry_msgs::Vector3 {
+                x: 0.0,
+                y: 0.0,
+                z: 0.0,
+            },
+            linear_acceleration_covariance: [0.0; 9],
+        }
+    }
+
+    fn generate_ros2_mcap(imu: &ros2msg::definitions::sensor_msgs::Imu) -> Vec<u8> {
```
oxkitsune: Creating these ROS2 `.mcap` files like this is a bit painful. I think we should check in an example MCAP file that we can then write tests against.

aedm:
@rerun-bot full-check

crates/store/re_data_loader/tests/mcap_test.rs
aedm: I simplfied code. Based on our conversation, a few of these tests will stay as the input is easier to comprehend than an actual MCAP binary.

crates/store/re_data_loader/tests/mcap_test.rs
oxkitsune, line 1: Can these tests not live inside the `loader_mcap.rs` file? at the bottom?
That would allow us to keep a few modules private

crates/store/re_data_loader/tests/assets/simple-protobuf.mcap
oxkitsune, line 1: I don't think this file is used anywhere right now right?

crates/utils/re_mcap/src/parsers/cdr.rs
```diff
@@ -62,6 +64,18 @@ pub fn try_decode_message<'d, T: Deserialize<'d>>(msg: &'d [u8]) -> Result<T, Cd
...
+    let mut writer = std::io::Cursor::new(Vec::new());
+    let mut header = [0u8; 4];
+    header[0..2].copy_from_slice(&RepresentationIdentifier::CdrLittleEndian.to_bytes());
+    writer
+        .write_all(&header)
+        .map_err(|err| CdrError::Other(anyhow::Error::from(err)))?;
+
+    cdr_encoding::to_writer::<_, byteorder::LittleEndian, _>(&mut writer, msg)?;
+    Ok(writer.into_inner())
+}
```
oxkitsune: I think it's a good idea to make these methods a bit more generic, and let the caller pass in the representation, because the little/big endianness is important here.

```suggestion
/// Encode a serializable message into a writer with the given CDR representation.
pub fn encode_message_into<T, W>(
    msg: &T,
    mut out: W,
    representation: RepresentationIdentifier,
) -> Result<(), CdrError>
where
    T: Serialize,
    W: std::io::Write,
{
    let id = representation.to_bytes();
    out.write_all(&[id[0], id[1], 0, 0])
        .map_err(|err| CdrError::Other(anyhow::Error::from(err)))?;

    if representation.is_big_endian() {
        cdr_encoding::to_writer::<_, byteorder::BigEndian, _>(&mut out, msg)?;
    } else {
        cdr_encoding::to_writer::<_, byteorder::LittleEndian, _>(&mut out, msg)?;
    }
    Ok(())
}

/// Encode a serializable message into a new buffer with the provided CDR representation.
pub fn try_encode_message<T: Serialize>(
    msg: &T,
    representation: RepresentationIdentifier,
) -> Result<Vec<u8>, CdrError> {
    let mut buf = Vec::new();
    encode_message_into(msg, &mut buf, representation)?;
    Ok(buf)
}
```

crates/store/re_data_loader/tests/assets/simple-protobuf.mcap
aedm: Right, accidental commit from another branch... 😫

crates/store/re_data_loader/tests/mcap_test.rs
aedm: That doesn't seem to be the currect practice in this repository. Snapshot tests usually live in a `tests` directory. I found 25 `snapshots` folders, 23 of which are subfolders of `tests`.

I don't have strong opinions here and I'd gladly move it to `src`. Do you still prefer that?

crates/store/re_data_loader/tests/mcap_test.rs
```diff
@@ -0,0 +1,157 @@
...
+        let mut chunk_id = ChunkId::from_u128(123_456_789_123_456_789_123_456_789);
+        while let Ok(res) = rx.recv() {
+            if let LoadedData::Chunk(_, _, chunk) = res {
+                let chunk = chunk.with_id(chunk_id).zeroed();
+                chunks.push(chunk);
+                chunk_id = chunk_id.next();
+            }
+        }
+        chunks
+    }
```
grtlr, line 93: Similar to what @oxkitsune said, I'd move this to the `re_mcap` crate to avoid making too much of that crate `pub`.

Some context: There is a good chance that `re_mcap` will become a crate that people will depend upon directly when they creating their own semantic converters, so I think we should be careful to keep the API surface small and clean, as much as possible.

crates/store/re_data_loader/tests/mcap_test.rs
grtlr, line 1: These test cases are great to check that our overall decoding logic works, but they don't test if our definitions are correct. Iirc, @oxkitsune mentioned that he has a repo to generate an MCAP file with a bunch of messages? Is that using official (Python) libraries?

Maybe it makes more sense to convert that file into a snippet test that gets run as part of `compare_snippets.py` (or similar)? So that we are sure that our definitions are correct?

crates/store/re_data_loader/tests/mcap_test.rs
aedm: Makes sense. The reason it's here because `load_mcap`is part of `re_data_loader`. Maybe it makes more sense to move that function (and this test) to `re_mcap`?

crates/store/re_data_loader/tests/mcap_test.rs
aedm: Sounds reasonable to me. If you think this PR doesn't have the right approach, I'd rather not spend time on it.

grtlr: Understandably!

I have just had a look at @oxkitsune's "ground truth" MCAP file. He's making some minor changes to it, but I'm now convinced that using it as snippet is a better approach to end-to-end testing the ROS2 MCAP data loader.

Can you two please sync on getting it into the repo + CI?

I'm sorry for all time that you have lost implementing this.

grtlr:
Will be superseded by

* https://github.com/rerun-io/rerun/pull/11175

PR #11077, opened 2025-09-02
End-to-end out-of-band entry headers: kickoff

teh-cmc:
This introduces gRPC headers as first class citizens of the Rerun Data Protocol.

To demonstrate it in action, two endpoints have been ported to use them:
* `rerun.cloud.v1alpha1.RerunCloudService/ReadDatasetEntry`
* `rerun.cloud.v1alpha1.RerunCloudService/GetPartitionTableSchema`

Update: actually, a third surprise endpoint joined the party: `/WriteChunks` was already header based.

There are a bunch of reasons why we'd want to leverage headers more. They are explained in:
* Sibling: https://github.com/rerun-io/dataplatform/pull/1650

### DIFF
--- a/crates/store/re_datafusion/src/datafusion_connector.rs
+++ b/crates/store/re_datafusion/src/datafusion_connector.rs
@@ -3,10 +3,8 @@ use std::sync::Arc;
 use datafusion::{catalog::TableProvider, error::DataFusionError};
 use tracing::instrument;
 
-use re_log_types::{EntryId, external::re_tuid::Tuid};
-use re_protos::cloud::v1alpha1::{
-    DatasetEntry, EntryFilter, ReadDatasetEntryRequest, ext::EntryDetails,
-};
+use re_log_types::EntryId;
+use re_protos::cloud::v1alpha1::{EntryFilter, ext::EntryDetails};
 use re_redap_client::ConnectionClient;
 
 use crate::partition_table::PartitionTableProvider;
@@ -49,24 +47,6 @@ impl DataFusionConnector {
         TableEntryTableProvider::new(self.client.clone(), entry.id)
             .into_provider()
             .await
-    }
-
-    #[instrument(skip(self), err)]
-    pub async fn get_dataset_entry(
-        &mut self,
-        id: Tuid,
-    ) -> Result<Option<DatasetEntry>, tonic::Status> {
-        let entry = self
-            .client
-            .inner()
-            .read_dataset_entry(ReadDatasetEntryRequest {
-                id: Some(id.into()),
-            })
-            .await?
-            .into_inner()
-            .dataset;
-
-        Ok(entry)
     }
 
     #[instrument(skip(self), err)]
--- a/crates/store/re_datafusion/src/partition_table.rs
+++ b/crates/store/re_datafusion/src/partition_table.rs
@@ -51,16 +51,17 @@ impl GrpcStreamToTable for PartitionTableProvider {
 
     #[instrument(skip(self), err)]
     async fn fetch_schema(&mut self) -> DataFusionResult<SchemaRef> {
-        let request = GetPartitionTableSchemaRequest {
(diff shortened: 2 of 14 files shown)


crates/store/re_datafusion/src/datafusion_connector.rs
teh-cmc, line 1: dead code

crates/store/re_datafusion/src/partition_table.rs
```diff
@@ -51,16 +51,17 @@ impl GrpcStreamToTable for PartitionTableProvider {
 
     #[instrument(skip(self), err)]
     async fn fetch_schema(&mut self) -> DataFusionResult<SchemaRef> {
-        let request = GetPartitionTableSchemaRequest {
-            dataset_id: Some(self.dataset_id.into()),
-        };
+        let mut req = tonic::Request::new(GetPartitionTableSchemaRequest {});
+        re_protos::headers::RerunHeadersInjector(&mut req)
+            .with_entry_id(self.dataset_id.to_string())
+            .map_err(|err| DataFusionError::External(Box::new(err)))?;
```
teh-cmc, line 57: Actual usage example

crates/store/re_log_types/src/lib.rs
teh-cmc, line 1: So kind can be passed using HTTP headers, internally.

crates/store/re_redap_client/src/connection_client.rs
```diff
@@ -147,11 +147,12 @@ where
         &mut self,
         entry_id: EntryId,
     ) -> Result<DatasetEntry, StreamError> {
+        let mut req = tonic::Request::new(ReadDatasetEntryRequest {});
+        re_protos::headers::RerunHeadersInjector(&mut req).with_entry_id(entry_id.to_string())?;
```
teh-cmc, line 151: Another actual usage example

crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
```diff
@@ -9,6 +9,17 @@ import "rerun/v1alpha1/common.proto";
...
+// ## Headers
+//
+// Most endpoints in the Rerun Cloud service require specific gRPC headers to be set.
+//
+// The so-called "standard dataset headers" correspond to at least one of the following headers:
+// * x-rerun-entry-id: ID of the entry of interest, e.g. `1860390B087BC65F602d68eb646c385c`.
+// * x-rerun-entry-name-bin: Name of the entry of interest, e.g. `droid:sample2k`.
+//
+// Headers with a -bin suffix must be base64-encoded (HTTP only supports ASCII values, UTF8 strings must
+// binary encoded).
```
abey79, line 22: Couple of thoughts on this.

"Standard dataset header" could possibly be made more to the point: "entry identification header" maybe? (My thinking is there there will be other such "standard" set of headers, e.g. partition id and layer name).

Taking the perspective of the client, I will certainly want to expose the id-or-name flexibility in the API (e.g. in `ConnectionClient`). This means there will be some boilerplate (entry identification enum, base64 encoding, etc.)—which is ofc fine. That boilerplate would not be significantly more complex than resolving the entry name using an `/FindEntries` call though. My question is thus: does it make sense to expose this id-or-name complexity all the way through the protocol to the server implementations when the UX is easily achievable on the client side? The obvious upside is that is saves a `/FindEntries` call, but is it worth it? (Not implying either way, genuinely curious.)

teh-cmc: Answered live, and the answer was: I don't know what's best (roundtrip resolve vs. all-in-on resolve). I just did the all-in-one because it only took a couple minutes and makes our lives easier for now. Time will tell :shrug:

teh-cmc:
Reviewed live -- merging before the conflicts on the other end get even more out of hand.

PR #11078, opened 2025-09-02
Add MCAP parsers for scalar sensor messages

oxkitsune:
### Related

* Part of #9949 

### What

This adds a generic `ScalarMessageParser` which supports ROS2 messages that implement a new `ScalarExtractor` trait. This generic parser is then used for the following ROS2 messages that contain simple scalar values:

- `sensor_msgs/msg/Temperature`
- `sensor_msgs/msg/FluidPressure`
- `sensor_msgs/msg/RelativeHumidity`
- `sensor_msgs/msg/Illuminance`
- `sensor_msgs/msg/Range`
- `sensor_msgs/msg/BatteryState`

### DIFF
--- a/crates/utils/re_mcap/src/parsers/ros2msg/definitions/sensor_msgs.rs
+++ b/crates/utils/re_mcap/src/parsers/ros2msg/definitions/sensor_msgs.rs
@@ -497,16 +497,16 @@ pub struct Range {
     /// The type of radiation used by the sensor.
     pub radiation_type: RadiationType,
 
-    /// The size of the arc that the distance reading is valid for [rad].
+    /// The size of the arc that the distance reading is valid for (rad).
     pub field_of_view: f32,
 
-    /// Minimum range value [m].
+    /// Minimum range value (m).
     pub min_range: f32,
 
-    /// Maximum range value [m].
+    /// Maximum range value (m).
     pub max_range: f32,
 
-    /// Range data [m].
+    /// Range data (m).
     ///
     /// ### Note
     ///


crates/utils/re_mcap/src/parsers/ros2msg/scalar_parser.rs
```diff
@@ -0,0 +1,137 @@
...
+/// Trait for extracting scalar values from ROS2 messages.
+///
+/// This trait allows different message types to specify which fields should be
+/// extracted as scalar values for visualization in Rerun.
+pub trait ScalarExtractor: serde::de::DeserializeOwned {
+    /// Extract scalar values from the message.
+    ///
+    /// Returns a vector of (`field_name`, `value`) pairs where `field_name` is used
+    /// for labeling in the visualization and `value` is the scalar measurement.
+    fn extract_scalars(&self) -> Vec<(&str, f64)>;
```
aedm, line 23: A default implementation would be nice. :)

aedm (approved):
LGTM

crates/utils/re_mcap/src/parsers/ros2msg/scalar_parser.rs
oxkitsune: I don't think it makes sense to have a default implementation for that function, as each struct _should_ implement it

aedm: I mean there are structs that could utilize a serde-based field enumeration. But it's probably overkill.

PR #11079, opened 2025-09-02
Fix mesh cache eviction issue

### LINKED ISSUE #11049: Changing albedo factor on a mesh causes massive vram memory leaks

IsseW:
### Related

Fixes #11049.

### What

This adds cache-eviction similarly to how it's done in `crates/viewer/re_viewer_context/src/cache/image_decode_cache.rs`. Store a `last_used_generation` for each cache entry, and if it wasn't used within the last frame we evict it.

### DIFF
--- a/crates/viewer/re_view_spatial/src/mesh_cache.rs
+++ b/crates/viewer/re_view_spatial/src/mesh_cache.rs
@@ -117,9 +117,9 @@ impl MeshCache {
 
 impl Cache for MeshCache {
     fn begin_frame(&mut self) {
-        // We aggressivly clear caches that weren't used in the last frame because
+        // We aggressively clear caches that weren't used in the last frame because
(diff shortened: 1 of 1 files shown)


Wumpf (approved):
nice!
very happy we went with that much simpler route now rather than overengineering the cache entry update as we discussed first. For now 😁

PR #11080, opened 2025-09-02
cpp-build-snippets can now run in `min-wheel-test` environment, report ci disk usage

Wumpf:
poking our wheel test ci a bit since it consistently started failing to build cpp snippets on linux-arm
https://github.com/rerun-io/rerun/actions/runs/17392389529/job/49369446668

maybe we'll learn something

* [ ] green main ci, just to make sure I don't somehow break it completely here


Wumpf:
@rerun-bot full-check

PR #11082, opened 2025-09-02
[pure refactor] Picking configuration is now part of view builder setup

Wumpf:
Previously, to set up picking on a `ViewBuilder`, you'd call `schedule_picking_rect`. With this refactor, it's instead part of the views initial configuration. Surprisingly, that's practically no issue at all for our existing Viewer & example code 🤷 

Why the hassle though, what's wrong with the old way?

This is a small piece of prep work I split out of another renderer refactor/enhancement pet-project of mine. At the core the issue addressed here is that it's very useful to know what draw phases are active when enqueuing `DrawData`. With the `schedule_picking_rect` approach this information wasn't guaranteed to be available ahead of time.
(With this, the only draw phase we don't know about whether it's active is the "screenshot phase" which is a bit of a strange copy-shim, rather dubious that it's a full phase in the first place.)

### DIFF
--- a/crates/viewer/re_renderer/src/draw_phases/picking_layer.rs
+++ b/crates/viewer/re_renderer/src/draw_phases/picking_layer.rs
@@ -81,11 +81,13 @@ impl<T: 'static + Send + Sync> PickingResult<T> {
     }
 }
 
+type ReadbackBeltUserData = Box<dyn std::any::Any + Send + Sync>;
+
 /// Type used as user data on the gpu readback belt.
-struct ReadbackBeltMetadata<T: 'static + Send + Sync> {
+struct ReadbackBeltMetadata {
     picking_rect: RectInt,
     world_from_cropped_projection: glam::Mat4,
-    user_data: T,
+    user_data: ReadbackBeltUserData,
 
     depth_readback_workaround_in_use: bool,
 }
@@ -174,7 +176,7 @@ impl PickingLayerProcessor {
     /// `enable_picking_target_sampling` should be enabled only for debugging purposes.
     /// It allows to sample the picking layer texture in a shader.
     #[allow(clippy::too_many_arguments)]
-    pub fn new<T: 'static + Send + Sync>(
+    pub fn new(
         ctx: &RenderContext,
         view_name: &DebugLabel,
         screen_resolution: glam::UVec2,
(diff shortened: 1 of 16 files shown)


crates/viewer/re_renderer/src/draw_phases/picking_layer.rs
```diff
@@ -81,11 +81,13 @@ impl<T: 'static + Send + Sync> PickingResult<T> {
     }
 }
 
+type ReadbackBeltUserData = Box<dyn std::any::Any + Send + Sync>;
```
Wumpf: looking at this again the name of this is kinda bad; should express that this is about picking data.
Worth noting that we dont'actually ever use this right now here

crates/viewer/re_renderer/src/draw_phases/picking_layer.rs
```diff
@@ -174,15 +176,15 @@ impl PickingLayerProcessor {
     /// `enable_picking_target_sampling` should be enabled only for debugging purposes.
     /// It allows to sample the picking layer texture in a shader.
     #[allow(clippy::too_many_arguments)]
-    pub fn new<T: 'static + Send + Sync>(
+    pub fn new(
```
Wumpf, line 173: not having a huge generic method here is nice and well worth the extra boxing imho

crates/viewer/re_renderer/src/draw_phases/picking_layer.rs
```diff
@@ -442,15 +444,15 @@ impl PickingLayerProcessor {
                     picking_depth_data = picking_depth_data.into_iter().step_by(4).collect();
                 }
 
-                PickingResult {
+                Some(PickingResult {
                     picking_id_data,
                     picking_depth_data,
-                    user_data: metadata.user_data,
+                    user_data: *metadata.user_data.downcast::<T>().ok()?,
```
Wumpf: not happy with the lack of error handling here. But arguably it was worse before - if you'd specify the wrong type you wouldn't have gotten the picking result at all

crates/viewer/re_renderer/src/draw_phases/picking_layer.rs
aedm: Is there a case where `ReadbackBeltMetadata::user_data` is not empty? I can't find one.

Wumpf: nope, we never use it atm. Mentioned that in the comment above.

But now that you point it out as well, gosh let's just remove it. Just causes mental friction 😅

crates/viewer/re_renderer/src/draw_phases/picking_layer.rs
aedm: This function is always called with `T = ()`. Could it be that we don't use `user_data`?

Wumpf:
> My favorite kind of PR.

just barely 😅

PR #11085, opened 2025-09-03
Make archetype and URDF loaders honor `entity_path_prefix`

### LINKED ISSUE #8344: entity_path_prefix not working for log_file()
**Describe the bug**
`entity_path_prefix` does not seem to work on `rerun.log_file_from_path` on 0.20.3.

**To Reproduce**
```
rec = rr.new_recording("app", spawn=True)

rr.log_file_from_path(
    "cube.stl",
    entity_path_prefix="test",
    recording=rec.to_native(),
)

```

oxkitsune:
### Related

* Closes #8344

### What

Both `loader_archetype` and `loader_urdf` now honor `entity_path_prefix` when it is set. The LeRobot and MCAP loaders do not, since their entity paths directly reflect the structure of the data being loaded.

### DIFF
--- a/crates/store/re_log_types/src/path/entity_path.rs
+++ b/crates/store/re_log_types/src/path/entity_path.rs
@@ -150,8 +150,13 @@ impl EntityPath {
         Self::new(
             file_path
                 .clean()
-                .iter()
-                .map(|p| EntityPathPart::from(p.to_string_lossy().to_string()))
+                .components()
+                .filter_map(|component| match component {
+                    std::path::Component::Normal(os_str) => {
+                        Some(EntityPathPart::from(os_str.to_string_lossy().to_string()))
+                    }
+                    _ => None, // Skip root, prefix, parent dir, current dir components
+                })
                 .collect(),
         )
     }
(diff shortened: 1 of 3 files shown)


crates/store/re_log_types/src/path/entity_path.rs
```diff
@@ -796,4 +801,69 @@ mod tests {
...
+            EntityPath::from_file_path(std::path::Path::new("")),
+            EntityPath::root()
+        );
+        assert_eq!(
+            EntityPath::from_file_path(std::path::Path::new("/")),
+            EntityPath::root()
+        );
+
+        // Windows paths
+        #[cfg(target_os = "windows")]
```
Wumpf, line 843: shouldn't those paths also work when not building for windows?

Wumpf (approved):
lgtm!

crates/store/re_log_types/src/path/entity_path.rs
oxkitsune: No they do not 🫠 
For unix targets it freaks out at the `\` instead of `/` and turns it into a path component.

PR #11086, opened 2025-09-03
Use `opened_store_id_or_recommended()` in URDF loader

oxkitsune:
### What

Update URDF loader to prefer the opened store when available, falling back to the recommended store ID if none is opened.

This was causing the recording to split in two when loading an URDF file using the SDK.


Wumpf (approved):
sounds familiar 😄

Issue #11084, opened 2025-09-03
Ensure the datafusion python package is at the right version


abey79:
Datafusion FFI, which we use, it currently unstable across releases:
- https://github.com/apache/datafusion/issues/17374

For this reason, we must always use matching python <-> rust datafusion versions, lest we segfault. Our python sdk should check that, and fail to load if a version mismatch is detected.

PR #11089, opened 2025-09-03
Bump datafusion-python to 48.0.0

abey79:
### Related

* closes https://github.com/rerun-io/rerun/issues/11084

### What

Adds a datafusion compatibility check. Failing the check is _far_ better than the segfault ensuing a mismatch.

This PR also bumps datafusion-python to 48. Datafusion-rust remains at 47 for now, which is ok since both version are ffi compatible. The check introduced in this PR is aware of this.

### DIFF
--- a/rerun_py/rerun_sdk/rerun/catalog.py
+++ b/rerun_py/rerun_sdk/rerun/catalog.py
@@ -22,7 +22,7 @@ if TYPE_CHECKING:
 
 # Known FFI compatible releases of Datafusion.
 DATAFUSION_MAJOR_VERSION_COMPATIBILITY_SETS = [
-    {47, 46, 48},
+    {47, 48},
 ]
 
 


rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -27,13 +29,26 @@ class CatalogClient:
...
+        # Check that we have a compatible version of datafusion.
+        #
+        # Note: ideally, we would pull the spec from our `pyproject.toml` file or package metadata. I tried, and it's
+        # complex, flaky, and introduces additional dependencies (such as `packaging` and/or `tomli`). Also, if we have
+        # a mismatch here, tests will fail, so we shouldn't forget to update.
+        #
+        # TODO(ab): we could be more flexible here and allow versions that are known to be FFI compatible (e.g. 48 is
+        # compatible with 47). That would make the version check more complicated though, unless we start depending on
+        # the `packaging` package.
+        version_spec = "datafusion==47.0.0"
```
nikolausWest: This seems like a strange place to hardcode the datafusion version.

.editorconfig
```diff
@@ -2,3 +2,9 @@
 indent_style = space
 indent_size = 2
 tab_width = 2
+
+# keep in sync with `taplo.toml`
+[*.toml]
```
Wumpf, line 7: keep it coming, makes all editors happy :)

rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -27,13 +29,26 @@ class CatalogClient:
...
-                "The 'datafusion' package is required to use `CatalogClient`. "
-                "You can install it with `pip install datafusion`."
-            )
+            raise RerunMissingDependencyError("datafusion", "datafusion")
+
+        # Check that we have a compatible version of datafusion.
+        #
+        # Note: ideally, we would pull the spec from our `pyproject.toml` file or package metadata. I tried, and it's
+        # complex, flaky, and introduces additional dependencies (such as `packaging` and/or `tomli`). Also, if we have
+        # a mismatch here, tests will fail, so we shouldn't forget to update.
```
Wumpf: couldn't we just do a hack in lints.py? can't be that bad 🤔 
If I update datafusion, would we get the right error?

also are we sure that updating

rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -27,13 +29,26 @@ class CatalogClient:
...
+        #
+        # TODO(ab): we could be more flexible here and allow versions that are known to be FFI compatible (e.g. 48 is
+        # compatible with 47). That would make the version check more complicated though, unless we start depending on
+        # the `packaging` package.
+        version_spec = "datafusion==47.0.0"
+
+        datafusion_version = version("datafusion")
+        if datafusion_version != version_spec.split("==")[1]:
```
Wumpf: it's checking [1], isn't that a zero here, or what does the split do exactly?

Wumpf (dismissed):

rerun_py/rerun_sdk/rerun/catalog.py
abey79: Well, as per the comment above, it's tricky to do better, and doing it that way is not Bad(tm) (aka tests will not let us forget about updating it).

rerun_py/rerun_sdk/rerun/catalog.py
abey79: that's the correct thing, I want "47.0.0" here, not "datafusion"

rerun_py/rerun_sdk/rerun/catalog.py
abey79: again, the reference pin is in `pyproject.toml` (we were missing one btw). Ideally, we'd refer to that from here, but it's a friggin pain for ${REASONS}. From there:

1) When updating datafusion, you need to grep for `datafusion==`. I've specifically used that syntax so you'll find this place.
2) If you _still_ forget about it, then CI will fail because we now have test that use `CatalogClient`. The correct datafusion version would be installed as per `pyproject.toml`, and the present check would fail if not updated.

For these reasons, I don't feel particularly compelled to add more cruft to the existing one in `lint.py`.

rerun_py/rerun_sdk/rerun/catalog.py
abey79: To answer the question, the split occurs at `"=="`, as per the `"=="` literal passed to `.split()` :)

rerun_py/pyproject.toml
```diff
@@ -38,7 +38,7 @@ text = "MIT OR Apache-2.0"
 tests = ["pytest==8.4.1"]
 notebook = ["rerun-notebook==0.25.0-alpha.1+dev"]
 datafusion = ["datafusion==47.0.0"]
-all = ["notebook", "datafusion"]
+all = ["notebook", "datafusion==47.0.0"]
```
ntjohnson1: I think this `all` target was actually broken when I added it and has been silently ignored.

It looks like it should be:
`all = ["rerun-sdk[notebook]", "rerun-sdk[datafusion]"]` however that then hits the rerun-notebook bootstrapping issue our pixi config calls out so we probably need to remove the `[all]` in our pixi. If that's too big a detour here I can file a ticket (or just fix it) separately.

abey79: I totally agree, this `all` target is fubar. I tiny bit less so now, but I would much rather remove it entirely tbh. Out of scope here though.

ntjohnson1: Ok I added it so I can file a ticket post coffee. It seemed like a nice convenience but annoying to test with our pixi setup.

abey79: Related: we probably should have the datafusion dep for the `notebook` extra.

rerun_py/rerun_sdk/rerun/catalog.py
ntjohnson1: Since this is mostly about the rust version being compatible with python if we update the rust and forget to update the pyproject this check would pass but we'd still segfault right? I wonder if it makes sense for us to have some kind of `compatible_datafusion` function on the rust side we expose to python that we can call for this check. Whether that returns major versions or full strings tbd.

abey79: That's correct. Ideally we'd have some check. In practice, updating datafusion basically is a major deal, with pyarrow, arrow-rs, and lancedb updates needed. It takes Tim like half a day to sort out. So sure, some automation would be great, but would also have little value short of solving the entire thing (aka Good Luck(tm)).

rerun_py/rerun_sdk/rerun/catalog.py
Wumpf: ah, I was misreading it. So but that means we're checking the full `47.0.0` string, isn't that way too aggressive?

rerun_py/rerun_sdk/rerun/catalog.py
Wumpf: hmmm okay. Begrudgingly convinced :)

rerun_py/pyproject.toml
ntjohnson1: Ya? I don't see where we depend on datafusion for the rerun-notebook package. They seem separate to me. We can probably discuss elsewhere though.

rerun_py/rerun_sdk/rerun/catalog.py
Wumpf: I mean minor & patch updates should be fine?

abey79: In general, yes. But given our pin (which is exactly the same in pyproject) and datafusion version scheme, no.

abey79: If we agree to have `packaging` as additional dependency, we can have a spec compliant version check here instead.

rerun_py/rerun_sdk/rerun/catalog.py
timsaucer: Suggestion:

```python
        expected_df_version = CatalogClientInternal.datafusion_major_version()

        datafusion_version = int(version("datafusion").split(".")[0])
        if datafusion_version != expected_df_version:
            raise RerunIncompatibleDependencyVersionError("datafusion", datafusion_version, expected_df_version)
```

And then in `rerun_py/src/catalog/catalog_client.rs` in `PyCatalogClientInternal`

```rust

    #[staticmethod]
    pub fn datafusion_major_version() -> u64 {
        datafusion_ffi::version()
    }
```

timsaucer: I tested this locally, including verifying it fails with another DF version.

abey79: Ah, this is much nicer, yes! I'll make the change.

rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -27,13 +29,26 @@ class CatalogClient:
...
-            raise ImportError(
-                "The 'datafusion' package is required to use `CatalogClient`. "
-                "You can install it with `pip install datafusion`."
-            )
+            raise RerunMissingDependencyError("datafusion", "datafusion")
+
+        # Check that we have a compatible version of datafusion.
+        #
+        # Note: ideally, we would pull the spec from our `pyproject.toml` file or package metadata. I tried, and it's
+        # complex, flaky, and introduces additional dependencies (such as `packaging` and/or `tomli`). Also, if we have
```
timsaucer: I think in the comments here we want a link to the underlying issue https://github.com/apache/datafusion/issues/17374 is that the current DataFusion FFI interface is not stable and we are currently requiring the Python major version is fixed to the rust version used in our library.

timsaucer (approved):

rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -14,10 +14,45 @@
...
 
+from .error_utils import RerunIncompatibleDependencyVersionError, RerunMissingDependencyError
+
 if TYPE_CHECKING:
     import datafusion
 
 
+# Known FFI compatible releases of Datafusion.
+DATAFUSION_MAJOR_VERSION_COMPATIBILITY_SETS = [
+    {47, 46, 48},
```
timsaucer: I think 46 won't work

abey79: It kind of did for the longest time, I discovered today. But since there is a sliver of doubt, i've removed it.

Issue #5531, opened 2024-03-15
WebSocket proxy should keep static data around forever


teh-cmc:

slyfox3:
While we are on this, can we add also a new option to `connect()` to only receive static + future data, but not older data?

teh-cmc:
This has been a common user-request, coming in in many different forms, the last few months.

Slapping a `user-request` on it, and slowly creeping it in plannings.

PR #11091, opened 2025-09-03
Keep static log messages longer in the proxy

emilk:
### Related
* Follows https://github.com/rerun-io/rerun/pull/11088
* Closes https://linear.app/rerun/issue/RR-2207/persist-static-data-in-grpc-proxy
* A better solution would be https://github.com/rerun-io/rerun/issues/5531

### What
`re_grpc_server` throws away old messages when a memory limit is reached.

Now temporal log messages are garbage collected before any static data.
See code comments for motivation.

In order to support this I added a `is_static` field to the `ArrowMsg` proto definition.

### TODO
* ~Make a sibling PR to dataplatform~ No need - this is not a breaking change (I checked)!
* [x] Add an automated test
* [x] Do a manual test
* [x] Remove the SDK-side warning "Received a batch with an encode timestamp but no decode timestamp. Latency measurements will be incomplete."
* [x] `@rerun-bot full-check`

### DIFF
--- a/crates/store/re_grpc_server/src/lib.rs
+++ b/crates/store/re_grpc_server/src/lib.rs
@@ -629,6 +629,13 @@ impl MessageBuffer {
             "Dropped {} bytes in {messages_dropped} message(s)",
             re_format::format_bytes(bytes_dropped as _)
         );
+
+        if max_bytes < self.size_bytes() {
+            re_log::warn_once!(
+                "The gRPC server is using more memory than the given memory limit ({}), despite having garbage-collected all non-persistent messages.",
+                re_format::format_bytes(max_bytes as _)
+            );
+        }
     }
 }
 


emilk:
@rerun-bot full-check

crates/store/re_grpc_server/src/lib.rs
```diff
@@ -491,6 +491,146 @@ impl MsgQueue {
...
+
+        while self.disposable.pop_front().is_some() {
+            messages_dropped += 1;
+            if self.size_bytes() < max_bytes {
+                break;
+            }
+        }
+
+        if max_bytes < self.size_bytes() {
+            re_log::info_once!("Starting to drop static messages too");
```
oxkitsune: In theory this could occur a while after reaching the memory limit for the first time, so I think adding a bit more context to this message is a good idea.

```suggestion
            re_log::info_once!(
                "Memory limit ({}) exceeded. Dropping old *static* log messages as well. Clients connecting after this will no longer see the complete set of static data.",
                re_format::format_bytes(max_bytes as _)
            );
```

crates/store/re_grpc_server/src/lib.rs
```diff
@@ -491,6 +491,146 @@ impl MsgQueue {
...
+            while self.static_.pop_front().is_some() {
+                messages_dropped += 1;
+                if self.size_bytes() < max_bytes {
+                    break;
+                }
+            }
+        }
+
```
oxkitsune, line 625: In the unlikely scenario that the persistent data exceeds the memory limit, we should probably log a warning as well

crates/store/re_sorbet/src/lib.rs
```diff
@@ -92,3 +94,12 @@ pub fn chunk_id_of_schema(
         Err(SorbetError::MissingChunkId)
     }
 }
+
+/// If this is a [`ChunkBatch`]: does it contain static data?
+// TODO(#10343): remove this
+pub fn is_static_chunk(batch: &RecordBatch) -> Option<bool> {
```
oxkitsune, line 100: might make sense to `#[inline]` this?

crates/store/re_grpc_server/src/lib.rs
```diff
@@ -491,6 +491,146 @@ impl MsgQueue {
...
+        // NOTE: the order here is important!
+        // TODO(#6523): make this behavior configurable
+        itertools::chain!(persistent.iter(), static_.iter(), disposable.iter(),)
+            .cloned()
+            .collect()
+    }
+
+    fn add_table(&mut self, table: TableMsgProto) {
+        self.disposable.push_back(table.into());
```
oxkitsune, line 551: Are tables _always_ disposable? is there no way to send a static table?

crates/store/re_sorbet/src/lib.rs
emilk: No - the implementation does _a lot_ of work. I measured though - the overhead is minimal compared to the cost of IPC encode/decode.

crates/store/re_grpc_server/src/lib.rs
emilk: Good point

crates/store/re_grpc_server/src/lib.rs
emilk: Tables are outside the temporal/static categorization: https://rerun.io/docs/howto/logging/send-table

PR #11094, opened 2025-09-03
Make sure our python deprecation notices end up in docs

ntjohnson1:
### Related
https://linear.app/rerun/issue/RR-2210/clearly-mark-deprecated-python-functions

### What
Enables a plugin to print a big deprecation notice for methods we tag with the deprecation decorator.
Screenshot 2025-09-03 at 6 25 00 AM

While I'm here I also fixed the broken python target I added before `all` which adds our multiple optionals. https://github.com/rerun-io/rerun/pull/11089#discussion_r2318351092


pixi.toml
```diff
@@ -721,6 +721,7 @@ mkdocstrings = "==0.26.2"
 mkdocstrings-python = "==1.12.1"
 setuptools = ">75"
 sphobjinv = "==2.3.1"
+griffe-warnings-deprecated = "==1.1.0"
```
emilk: Sort 🙏

emilk (approved):
Nice - thank you!

ntjohnson1:
FYI @abey79 since we were talking about all being broken. Feel free to remove if you prefer but at least now it isn't a no-op

PR #11061, opened 2025-09-02
Implement full text filtering for string-based columns of the table widget

abey79:
### Related

* part of https://linear.app/rerun/project/ability-to-filter-tables-in-the-viewer-8dcb9ff4e6bc/overview
* part of https://github.com/rerun-io/rerun/issues/10560

### What

Initial take at filtering in the datafusion table widget. This PR establishes the structure, and introduces full text `contains` filters for string based columns (including `StringList`).

Issue #10560, opened 2025-07-08
Dataframe/Partition View: filter/search over column values


pweids:
It would be helpful to have a search bar and/or some kind of filter bar in the partition view. If trying to find a specific partition by the ID, it's often more convenient to type the ID in than it is to sort and scroll.

Feedback source:

* Ultra [1 ](https://www.notion.so/rerunio/Customer-Feedback-Ultra-22ab24554b1980259e5bfaec24b4e6df?source=copy_link#22ab24554b1980a985f8d6854cae5fff)and [2](https://www.notion.so/rerunio/Customer-Feedback-22ab24554b1980259e5bfaec24b4e6df?source=copy_link#22ab24554b19803ab67fd01f4b4c1c5c)
* [Maniro](https://www.notion.so/rerunio/Rolling-Updates-137b24554b1980bb9615f88ea21e2644?source=copy_link#242b24554b19800692ade9fd9eb3b735)

Figma explorations for filtering: [Frame 427322145 - Data platform (Figma)](https://www.figma.com/design/04eHlTWW361rIs3YesfTJo/Data-platform?node-id=1189-5085&t=CLpv0NGamivYkNYL-1)

Internally tracked by https://linear.app/rerun/project/ability-to-filter-tables-in-the-viewer-8dcb9ff4e6bc/overview

abey79:
Closing this as completed as I'm about to wrap up the initial work on this feature. Let's open dedicated issues for any bugs and/or follow-up work.

PR #11095, opened 2025-09-03
Add support for filtering on boolean columns

abey79:
### Related

* follow up to https://github.com/rerun-io/rerun/pull/11061

### What

Bool and List[Bool] column can now be filtered by `is true/false`. For List[Bool], all instances must be equal to the query value.

Added a bunch of test for the filter themselves.

This PR also goes all in on the "semantic" syntax highlighting, which does raise some questions/inconsistencies. We will likely need to:
- discuss the colour theme
- update the table rendering to also use apply the color scheme

image

### DIFF
--- a/crates/viewer/re_dataframe_ui/tests/filter_tests.rs
+++ b/crates/viewer/re_dataframe_ui/tests/filter_tests.rs
@@ -189,13 +189,13 @@ async fn test_string_contains() {
     filter_snapshot!(
         Filter::new("string", FilterOperation::StringContains("A".to_owned())),
         TestColumn::strings(false),
-        "A"
+        "a_uppercase"
     );
 
     filter_snapshot!(
         Filter::new("string", FilterOperation::StringContains("A".to_owned())),
         TestColumn::strings(true),
-        "nullable_A"
+        "nullable_a_uppercase"
     );
 
     filter_snapshot!(
--- a/crates/viewer/re_dataframe_ui/tests/snapshots/filter_tests__string_contains@a_uppercase.snap
+++ b/crates/viewer/re_dataframe_ui/tests/snapshots/filter_tests__string_contains@a_uppercase.snap
@@ -6,7 +6,7 @@ expression: "(filter, result)"
     Filter {
         column_name: "string",
         operation: StringContains(
-            "a",
+            "A",
         ),
     },
     RecordBatch {
--- a/crates/viewer/re_dataframe_ui/tests/snapshots/filter_tests__string_contains@nullable_a_uppercase.snap
+++ b/crates/viewer/re_dataframe_ui/tests/snapshots/filter_tests__string_contains@nullable_a_uppercase.snap
@@ -6,7 +6,7 @@ expression: "(filter, result)"
     Filter {
         column_name: "string",
         operation: StringContains(
-            "a",
+            "A",
         ),
     },
     RecordBatch {
@@ -15,7 +15,7 @@ expression: "(filter, result)"
                 Field {
                     name: "string",
                     data_type: Utf8,
-                    nullable: false,
+                    nullable: true,
                     dict_id: 0,
                     dict_is_ordered: false,
                     metadata: {},


crates/viewer/re_dataframe_ui/src/filter_ui.rs
```diff
@@ -277,14 +279,55 @@ impl SyntaxHighlighting for FilterOperation {
...
+        );
+
+        job.append(
+            &format!("{} ", self.column_name),
+            0.0,
+            normal_text_format.clone(),
+        );
+        job.append(self.operator_text, 0.0, operator_text_format);
+    }
+}
```
lucasmerlin: With https://github.com/rerun-io/rerun/pull/11099 you should be able to remove this and just do it inline with `append_identifier` and `append_with_format` (as long as it's only used in a single place of course).

crates/viewer/re_dataframe_ui/src/filter_ui.rs
```diff
@@ -277,14 +279,55 @@ impl SyntaxHighlighting for FilterOperation {
...
+    fn syntax_highlight_into(&self, style: &Style, job: &mut LayoutJob) {
+        let tokens = re_ui::design_tokens_of_visuals(&style.visuals);
+
+        let normal_text_format = egui::TextFormat::simple(
+            egui::TextStyle::Body.resolve(style),
+            egui::Color32::PLACEHOLDER,
+        );
+        let operator_text_format = egui::TextFormat::simple(
+            egui::TextStyle::Body.resolve(style),
+            tokens.table_filter_operator_text_color,
```
lucasmerlin: I think these tokens should be more general and named after the semantinc meaning. So if none of the existing tokens fits maybe we can add a `code_keyword` token.

crates/viewer/re_dataframe_ui/src/filter_ui.rs
```diff
@@ -311,13 +363,17 @@ impl FilterOperation {
     fn operator_text(&self) -> &'static str {
         match self {
             Self::StringContains(_) => "contains",
+
+            Self::BooleanEquals(_) => "is",
         }
     }
 
     /// Display text of the right-hand side operand (aka the user-provided query data).
     fn rhs_text(&self) -> String {
```
lucasmerlin: I feel this should return a WidgetText or maybe a `impl SyntaxHighlighted` or something, since the highlighting color should depend on the type of the value. For boolean that should be `code_primitive` and for string `code_string`.

crates/viewer/re_dataframe_ui/src/filters.rs
```diff
@@ -98,6 +108,22 @@ impl FilterOperation {
 
                 Ok(contains_patch(lower(operand), lower(lit(query_string))))
             }
+
+            Self::BooleanEquals(value) => match field.data_type() {
+                DataType::Boolean => Ok(col(column.clone())),
```
lucasmerlin: Shouldn't this do something like 
```suggestion
                DataType::Boolean => Ok(col(column.clone()).eq(value)),
```

crates/viewer/re_dataframe_ui/src/filter_ui.rs
```diff
@@ -302,6 +345,15 @@ impl FilterOperation {
...
                 }
             }
+
+            Self::BooleanEquals(query) => {
+                ui.label(top_text);
+                if ui.re_radio_value(query, true, "true").clicked()
+                    || ui.re_radio_value(query, false, "false").clicked()
+                {
+                    action = FilterUiAction::CommitStateToBlueprint;
+                }
```
lucasmerlin, line 305: Would be awesome if these could also have the syntax highlighting color

crates/viewer/re_dataframe_ui/src/filters.rs
abey79: Ah right!

crates/viewer/re_dataframe_ui/src/filter_ui.rs
abey79: This returns the actual operator text, so I don't think it should be type dependant. I actually like the idea of having `code_keyword` and use that (thought I may wait until Katya is back and we make all the color decisions).

lucasmerlin: Oh, this comment was meant for rhs_text 😅

crates/viewer/re_dataframe_ui/src/filter_ui.rs
abey79: let's keep as TBD :)

crates/viewer/re_dataframe_ui/src/filter_ui.rs
abey79: ah right. anyways, fixed.

crates/viewer/re_dataframe_ui/tests/snapshots/filter_tests__string_contains_list@c.snap.new
```diff
@@ -0,0 +1,53 @@
...
+            [
+              "c",
+              "ab",
+            ],
+              StringArray
+            [
+              "A B",
+              "aBc",
+            ],
+            ],
```
lucasmerlin: So, for boolean arrays, all values must match, while for string arrrays, it's enough if a single value matches? I feel like this is a bit unexpected. But on the other hand it makes sense, I guess. Eventually we should have an `all`/`any` toggle for arrays.

lucasmerlin (approved):
Awesome!

crates/viewer/re_dataframe_ui/tests/snapshots/filter_tests__string_contains_list@c.snap.new
abey79: That's correct, the semantics are not super consistent. TBH, i'm not expecting to see a whole lot of list of bool with more than a single one, so it's probably not a huge deal. But we definitely need to review/design the semantics of all of these, and make sure the UI is consistent.

FWIW, I'm about to introduce a 3rd semantics for scalar: only consider the 1st element 😬   (because I'm not sure how to deal with `[1, 2, 3] > 3` but we _must_ deal with `[1] > 3` because that's how our properties are typed.

cc @pweids

PR #11096, opened 2025-09-03
Cherry-pick version bump for alpha.2

abey79:
Because of git shenanigans, it's easier to not merge the 0.25.0-alpha.2 branch back to `main`. This is a cherry pick of its version bump commit, which we need.

Related:
- https://github.com/rerun-io/rerun/pull/11042


abey79:
well, this is not needed it turns out

Issue #11075, opened 2025-09-02
Ensure consistent syntax highlighting


lucasmerlin:
- related: https://github.com/rerun-io/rerun/issues/6315

The new arrow ui introduced syntax highlighting, but it's not used everywhere yet, e.g. some strings and some custom editable uis are not syntax highlighted yet.

_Originally posted by @emilk in https://github.com/rerun-io/rerun/pull/10777#discussion_r2315918268_

PR #11099, opened 2025-09-03
Change SyntaxHighlighting trait to use SyntaxHighlightedBuilder

lucasmerlin:
…and use design tokens in existing SyntaxHighlighting trait impls

### Related

* part of <https://github.com/rerun-io/rerun/issues/11075>

### What

A controvesial change here is that the entity path is now displayed in monospace (since it uses append_identifier). Maybe that isn't the right thing and we should add another append_* for rerun specific identifiers (since append_identifiers was initially used for user-data identifiers like arrow struct fields).

### DIFF
--- a/crates/viewer/re_ui/src/syntax_highlighting.rs
+++ b/crates/viewer/re_ui/src/syntax_highlighting.rs
@@ -17,7 +17,7 @@ pub trait SyntaxHighlighting {
         builder.job
     }
 
-    fn syntax_highlight_into(&self, builder: &mut SyntaxHighlightedBuilder);
+    fn syntax_highlight_into(&self, builder: &mut SyntaxHighlightedBuilder<'_>);
 }
 
 // ----------------------------------------------------------------------------
@@ -137,14 +137,6 @@ fn text_format(style: &Style) -> TextFormat {
     }
 }
 
-fn monospace_text_format(style: &Style) -> TextFormat {
-    TextFormat {
-        font_id: egui::TextStyle::Monospace.resolve(style),
-        color: style.visuals.text_color(),
-        ..Default::default()
-    }
-}
-
 /// Monospace text format with a specific color (that may be overridden by the style).
 fn monospace_with_color(style: &Style, color: Color32) -> TextFormat {
     TextFormat {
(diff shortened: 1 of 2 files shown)


crates/viewer/re_ui/src/syntax_highlighting.rs
```diff
@@ -40,48 +40,64 @@ impl<'a> SyntaxHighlightedBuilder<'a> {
...
+    pub fn with(mut self, portion: &dyn SyntaxHighlighting) -> Self {
+        portion.syntax_highlight_into(&mut self);
+        self
+    }
+
+    #[inline]
+    pub fn append(&mut self, portion: &dyn SyntaxHighlighting) -> &mut Self {
+        portion.syntax_highlight_into(self);
         self
     }
```
abey79, line 52: This is a bit confusion. I was definitely surprise to see `with` when all the interesting stuff is with `append_*`.  Shouldn't we just get rid of `with`?

lucasmerlin: I think with can be useful if you quickly want to construct something syntax highlighted inline, e.g. `SyntaxHighlightedBuilder::new(style).with(a).with(b).into_job()`.

Ideally the best api would have everything be duplicated so we had e.g. both `with_primitive` and `append_primitive`, one taking `mut self` and the other `&mut self`. 

This duplication is of course annoying but I'm not sure if there is a better way. Since this is such a low level building block I'd optimize for the most ergonomic api even if it leads to duplication.

lucasmerlin: I guess we could make `SyntaxHighlightedBuilder::into_job` take `&mut self` instead, then `SyntaxHighlightedBuilder::new(style).append(a).append(b).into_job()` should work. Since it's marked as inline, the compile should hopefully be able to optimize the clone?

abey79: Yeah I agree duplication is acceptable here 🤷🏻

abey79: But we could also force everyone into the `with_` version. Slightly more verbose for trait implementors, but that's more rare than actual use of the builder.

lucasmerlin: That'd be really annoying, since you can't move out of a &mut reference. Then you'd have to `mem::take` all the time so that you're able to call the builder functions.

Issue #9862, opened 2025-05-01
AnnotationContext for keypoints broken in 0.23.1


pablovela5620:
**Describe the bug**

AnnotationContext no longer works on 0.23.1 when trying to log link connections (on notebooks)

**To Reproduce**
Minimal script to reproduce on a jupyter notebook
```python
import numpy as np
import rerun as rr
import rerun.blueprint as rrb

left_joints = np.array([
    [-0.0956699,   0.00638343,  0.00618631],
    [-0.0715802,  -0.0091389,   0.03199915],
    [-0.0519469,  -0.0082476,   0.0556987 ],
    [-0.0297292,  -0.0136805,   0.07022282],
    [-0.0022312,  -0.018095,    0.0909145 ],
    [-0.0075726,   0.00118307,  0.0268723 ],
    [ 0.02510622,  0.00519243,  0.02908936],
    [ 0.04726213,  0.003894,    0.02897524],
    [ 0.0705249,   0.00461197,  0.03302451],
    [-0.0010094,   0.00490446,  0.00282876],
    [ 0.03017318,  0.00676579, -0.0027657 ],
    [ 0.05307782,  0.00551369, -0.0067102 ],
    [ 0.07899282,  0.00614665, -0.0120408 ],
    [-0.0139343,   0.00242601, -0.0204868 ],
    [ 0.0143799,   0.00449301, -0.0255854 ],
    [ 0.03790041,  0.0028049,  -0.0332192 ],
    [ 0.06080423,  0.00734306, -0.0402022 ],
    [-0.0268829,  -0.0035569,  -0.037023  ],
    [-0.0098685,  -0.003495,   -0.0495218 ],
    [ 0.00599835, -0.0041862,  -0.0598537 ],
    [ 0.02189885, -0.0016281,  -0.0701316 ]
])

MEDIAPIPE_IDS = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
MEDIAPIPE_ID2NAME = {
    0: 'WRIST',
    1: 'THUMB_CMC',
    2: 'THUMB_MCP',
    3: 'THUMB_IP',
    4: 'THUMB_TIP',
    5: 'INDEX_MCP',
    6: 'INDEX_PIP',
    7: 'INDEX_DIP',
    8: 'INDEX_TIP',
    9: 'MIDDLE_MCP',
    10: 'MIDDLE_PIP',
[...]

PR #11100, opened 2025-09-03
Add compatibility to both numpy1 and numpy2.

Benjamin-Tan:
### Related

https://github.com/rerun-io/rerun/pull/9109#issuecomment-3245297266

### What
Add compatibility to numpy1, to relax the requirement of numpy2 as the minimum version.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -603,7 +603,7 @@ jinja2 = ">=3.1.3,<3.2"           # For `build_screenshot_compare.py` and other 
 mypy = "1.14.1.*"
 nasm = ">=2.16"                   # Required by https://github.com/memorysafety/rav1d for native video support
 ninja = "1.11.1.*"
-numpy = ">=2"                     # Whenever upgrading here, also make sure to upgrade in `rerun_py/pyproject.toml`
+numpy = ">=1.24"                  # Whenever upgrading here, also make sure to upgrade in `rerun_py/pyproject.toml`
 prettier = ">=3.6"
 pyarrow = "18.0.0.*"              # Whenever upgrading here, also make sure to upgrade in `rerun_py/pyproject.toml`
 pytest = ">=8.4.1"
@@ -693,7 +693,7 @@ platforms = ["linux-64", "linux-aarch64", "osx-arm64", "osx-64", "win-64"]
 #
 # This is also redundantly defined in `python-pypi`
 opencv = ">4.6"
-numpy = ">=2"   # Rerun still needs numpy <2. Enforce this outside of the pypi dep tree so we pick up the conda version.
+numpy = ">=1.24" # Rerun still needs numpy <2. Enforce this outside of the pypi dep tree so we pick up the conda version.
 
 [feature.python-dev.pypi-dependencies]
 # Install the `rerun_py` as a package in editable mode.
@@ -748,7 +748,7 @@ platforms = ["linux-64", "linux-aarch64", "osx-arm64", "osx-64", "win-64"]
 [feature.python-pypi.dependencies]
 # Note these are the same as in `python-dev` but we need to repeat them here because the two are mutually exclusive.
 opencv = ">4.6"
-numpy = ">=2"   # Rerun still needs numpy <2. Enforce this outside of the pypi dep tree so we pick up the conda version.
+numpy = ">=1.24" # Rerun still needs numpy <2. Enforce this outside of the pypi dep tree so we pick up the conda version.
 
 
 [feature.python-pypi.pypi-dependencies]
--- a/rerun_py/pyproject.toml
+++ b/rerun_py/pyproject.toml
@@ -16,7 +16,7 @@ classifiers = [
 dependencies = [
   # Must match list in `.github/workflows/reusable_test_wheels.yml`
   "attrs>=23.1.0",
-  "numpy>=2",
+  "numpy>=1.24",
   "pillow>=8.0.0",          # Used for JPEG encoding. 8.0.0 added the `format` arguments to `Image.open`
   "pyarrow>=18.0.0",
   "typing_extensions>=4.5",
--- a/rerun_py/rerun_sdk/rerun/datatypes/angle.py
+++ b/rerun_py/rerun_sdk/rerun/datatypes/angle.py
@@ -12,6 +12,8 @@ import numpy as np
 import numpy.typing as npt
 import pyarrow as pa
 from attrs import define, field
+from packaging import version
+IS_NUMPY_2 = True if version.parse(np.__version__) >= version.parse("2.0.0") else False
 
 from .._baseclasses import (
     BaseBatch,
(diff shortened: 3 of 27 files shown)


Wumpf (changes requested):
Numpy 1.xx is known to be not working with at least some of our APIs see for instance https://github.com/rerun-io/rerun/issues/9862 for an issue that came up before we correctly speced the version

If we were to enable numpy 1 support again we can only land this if also our full test suite runs against it. That plus the general maintenance burden of supporting both numpy versions is not something I want us to take on right now.

Wumpf:
Ah didn't see the discussion over here so far https://github.com/rerun-io/rerun/pull/9109#issuecomment-3245297266

As @emilk pointed out, we'd need full test coverage of both numpy 1 and numpy 2 then and setting this up isn't entirely trivial (can be done ofc).
`pixi run py-test` will still use numpy 2 unless you also change the numpy version. And even so that's only part of our test suite.


> Mainly about the challenge of upgrading all the system dependencies to be numpy2 compatible (especially pybind-related components).

Maybe you can go a bit more into the details of what's breaking there? :)

Benjamin-Tan:
I was referring to the other pre-compiled libraries (outside of rerun) that are not compatible with numpy2.

As mentioned https://github.com/rerun-io/rerun/pull/9109#issuecomment-3249724998, I ran `pixi run -e py py-build && pixi run -e py py-test` with the following changes to run the unit tests locally with numpy < 2, and all the 541 unit tests passed (without committing it). Are there any other changes that need to be made to test it against both versions?

```
index 211439f7f..96b4f632e 100644
--- a/pixi.toml
+++ b/pixi.toml
@@ -126,7 +126,7 @@ wheel-test-min = [
   "base",
   "cpp",
   "py-test-deps",
-  "python-pypi",
+  "python-dev",
   "wheel-build",
   "wheel-test",
 ]
@@ -138,7 +138,7 @@ wheel-test = [
   "wheel-build",
   "py-test-deps",
   "examples-common",
-  "python-pypi",
+  "python-dev",
 ]

 # Example environments
@@ -151,7 +151,7 @@ examples = ["examples-common", "python-dev", "wheel-build", "examples-tasks"]

 # This environment uses the pypi-published version of the python package. This avoids the need to
 # build the package from source if you just want to try running the examples locally.
-examples-pypi = ["examples-common", "python-pypi"]
+examples-pypi = ["examples-common", "python-dev"]


 ################################################################################
@@ -603,7 +603,7 @@ jinja2 = ">=3.1.3,<3.2"           # For `build_screenshot_compare.py` and other
 mypy = "1.14.1.*"
 nasm = ">=2.16"                   # Required by https://github.com/memorysafety/rav1d for native video support
 ninja = "1.11.1.*"
-numpy = ">=1.24"                  # Whenever upgrading here, also make sure to upgrade in `rerun_py/pyproject.toml`
+numpy = ">=1.24,<2"               # Whenever upgrading here, also make sure to upgrade in `rerun_py/pyproject.toml`
 prettier = ">=3.6"
 pyarrow = "18.0.0.*"              # Whenever upgrading here, also make sure to upgrade in `rerun_py/pyproject.toml`
 pytest = ">=8.4.1"
@@ -693,7 +693,7 @@ platforms = ["linux-64", "linux-aarch64", "osx-arm64", "osx-64", "win-64"]
 #
 # This is also redundantly defined in `python-pypi`
 opencv = ">4.6"
-numpy = ">=1.24" # Rerun still needs numpy <2. Enforce this outside of the pypi dep tree so we pick up the conda version.
+numpy = ">=1.24,<2" # Rerun still needs numpy <2. Enforce this outside of the pypi dep tree so we pick up the conda version.

 [feature.python-dev.pypi-dependencies]
 # Install the `rerun_py` as a package in editable mode.
@@ -748,7 +748,7 @@ platforms = ["linux-64", "linux-aarch64", "osx-arm64", "osx-64", "win-64"]
 [feature.python-pypi.dependencies]
 # Note these are the same as in `python-dev` but we need to repeat them here because the two are mutually exclusive.
 opencv = ">4.6"
-numpy = ">=1.24" # Rerun still needs numpy <2. Enforce this outside of the pypi dep tree so we pick up the conda version.
+numpy = ">=1.24, <2" # Rerun still needs numpy <2. Enforce this outside of the pypi dep tree so we pick up the conda version.


 [feature.python-pypi.pypi-dependencies]
diff --git a/rerun_py/pyproject.toml b/rerun_py/pyproject.toml
index 58be73544..a2efbf9c2 100644
--- a/rerun_py/pyproject.toml
+++ b/rerun_py/pyproject.toml
@@ -16,7 +16,7 @@ classifiers = [
 dependencies = [
   # Must match list in `.github/workflows/reusable_test_wheels.yml`
   "attrs>=23.1.0",
-  "numpy>=1.24",
+  "numpy>=1.24,<2",
   "pyarrow>=18.0.0",
   "typing_extensions>=4.5",
```

Wumpf:
Given that we know numpy1 fails with Rerun as-is (see https://github.com/rerun-io/rerun/issues/9862 - we haven't done anything to mitigate this, so _at least_ that error will still happen) you need to make sure that pixi actually installed numpy1 in its environment. I don't have full overview of how the setup works there, there's likely more things involved than the changes you did. Probably best to start with a fresh virtual environment, install the adjusted Rerun version in there and then run all test suites involving python code from there

In order to actually ship this we'd also have to ...
* figure out how to run numpy 1 and 2 on CI - atm I don't even know how to do this, likely involves wrapping all our test runs in a way that allows running both
    * for python tests, snippet comparision tests, example builds, etc. 
* adjust codegen to not emit the code that caused issues in #9862
* probably add additional tests to catch things like

The problem with this is that it means that anyone working on Rerun python sdk will have to bear this cost of having to test & maintain both versions. And as mentioned I really don't want to pay this unless this becomes a major problem across our userbase, something we haven't observed yet.

So either way I don't see us landing this currently.

Benjamin-Tan:
Actually, this PR already resolved the issue of https://github.com/rerun-io/rerun/issues/9862 by having a compatible API rather than bumping the minimum version to above 2. I am happy to make the other changes as needed. How about recommended to have numpy>=2, but numpy<2 is the community supported effort?

emilk:
So we've discussed this a bit internally, and arrived to the following:

We don't want to officially promise numpy1 support, because that will make testing on CI a lot more complex
Therefore we want our `pyproject.toml` and `pixi.toml` to still require `numpy >= 2`.

BUT - You should be able to install `rerun-sdk` with numpy 1 using `pip install rerun-sdk --no-deps`, regardless of what our `pyproject.toml` says.

So, if there are some small code changes (like the ones in this PR) that would allow us to keep support for numpy1, we will be open to merging those.

In other words: please re-open this PR but revert the changes to `pixi.toml` and `rerun_py/pyproject.toml` :)

(but also: you can't just modify the contents of our code-generated files - you need to modify `crates/build/re_types_builder/src/codegen/python/mod.rs` instead and then run `pixi run codegen`)

Wumpf:
What @emilk said! :)
Tbh I totally missed that you've attempted to fix things because the generated files are hidden by default. It would be nice to find a way to do the version check only once and not in each .py upon load; that accumulates to quite a bit of startup overhead. Thanks!

Benjamin-Tan:
@emilk Noted with thanks.
@Wumpf I have made the changes. Could you reopen this PR, please?

Wumpf:
github doesn't seem to allow that because there's force pushes on it. That or some complications with creating a PR of `main` (creating a separate branch makes a lot of things easier!)
image

probably easiest to just recreate it 🤷

PR #11103, opened 2025-09-03
End-to-end out-of-band entry headers: datasets (part 1)

teh-cmc:
First batch of dataset APIs migrated to header-based entry routing.

* Sibling: https://github.com/rerun-io/dataplatform/pull/1662
* Requires: https://github.com/rerun-io/rerun/pull/11077


crates/store/re_protos/src/headers.rs
teh-cmc, line 1: Bunch of API improvements as I learned more and more while porting endpoints.

teh-cmc:
Reviewed live -- merging before the conflicts on the other end get even more out of hand.

PR #11108, opened 2025-09-04
Alpha release: better process for GH releases and changelog

abey79:
Title. See diff for the proposal

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -204,7 +204,7 @@ jobs:
                 - [ ] Delete the GH release draft
             - [ ] Merge or close the release PR
               - **IFF** the release job succeeds and the link checker is happy, you may merge the PR.
-              - Otherwise, close the PR without merging and cherrypick to `main` as required.
+              - Otherwise, close the PR without merging. Cherrypick any interesting commit to `main`, but *not* the version bump one (it would introduce bad links).
           - For non-alpha releases:
             - For any added commits, run the release workflow in 'rc' mode again
             - After testing, _ensure that this PR is mergeable to `main`_, then run the release workflow in 'release' mode
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -194,8 +194,18 @@ jobs:
           cat <<EOF > pr-body.txt
           ### Next steps
           - Test the release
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -194,8 +194,18 @@ jobs:
           cat <<EOF > pr-body.txt
           ### Next steps
           - Test the release
-          - If this is an 'alpha' release, merge the PR **only** if the release job succeeds and the link checker is happy. Otherwise, close the PR without merging and cherrypick to main as required.
-          - Otherwise:
+          - For alpha releases:
+            - [ ] Should the GH release be published for this alpha? Make a call (give extra love to alphas deployed to Rerun Cloud or when external testing is required).
```
andrea-reale, line 198: What is the downside in removing the ambiguity and say that we always generate a raw changelog for alphas?

.github/workflows/release.yml
```diff
@@ -194,8 +194,18 @@ jobs:
...
+              - If yes:
+                - [ ] Edit and publish the GitHub release
+                - [ ] Stretch goal: generate a raw changelog and add it to the GH release with this disclamer:
+                      **DISCLAIMER**: This is an unreviewed, automatically generated changelog. We only provide fully reviewed changelogs for final releases.
+              - If no:
+                - [ ] Delete the GH release draft
+            - [ ] Merge or close the release PR
+              - **IFF** the release job succeeds and the link checker is happy, you may merge the PR.
+              - Otherwise, close the PR without merging and cherrypick to `main` as required.
```
Wumpf: should probably clarify what should be cherry-picked. Is the version patching done by the bot something that has to be cherry-picked?

Wumpf (commented):
Thanks!

another thing we should add is that we should manual git tag all releases with the linear-tag associated with that release. I.e. the release we're on right now should get a git tag `release-oss-0.26-mid`
(I'm very sure I discussed this with a bunch of folks and there was lotsa agreement, but I think we didn't note it down 😱 )

.github/workflows/release.yml
abey79: Basically the friction. It's added steps for a use case that we intended to be super fast and smooth back then.

Taking a step back, it appears that we want:
- a way to generate a tagged build that is super smooth
- a way to generate a tagged build we're willing to invest _some_ love into, because it's definitely making it to customer's hand

For that we could introduce an alpha vs. beta distinction (though obv out-of-scope for this pr).

jprochazk (approved):

PR #11109, opened 2025-09-04
JS Publish: Always build all packages

jprochazk:


abey79 (approved):
Thanks! Let's test this with alpha.4

PR #11112, opened 2025-09-04
Constrained column projection for `/ScanPartitionTable`

teh-cmc:
This gets rid of the arbitrary `ScanParameters` on `/ScanPartitionTable`, which is a prerequisite for the upcoming new layer-compatible endpoint.

Columns can still be projected server-side using a dedicated parameter, but everything else (pagination, filtering, etc) must be off limit.

* Sibling: https://github.com/rerun-io/dataplatform/pull/1665


abey79 (approved):
Seems we weren't using them anyways, so yay?

PR #11114, opened 2025-09-04
Tell the user about the native viewer

emilk:
### What
It seems some of our users are using the web viewer not knowing we have a much faster native viewer.

So let's tell them!

* [x] Only shown in web viewer
* [x] Not shown when viewing examples
* [x] Not shown in notebooks
* [x] No toast is shown
* [x] If the user clicks "Dismiss" on it, then never show it again, _ever_

Screenshot 2025-09-05 at 13 19 24

Discussion: https://rerunio.slack.com/archives/C08TEFXNHB2/p1756389003633929


crates/viewer/re_viewer/src/app.rs
```diff
@@ -1983,6 +1989,26 @@ impl App {
...
                         }
                     }
+
+                    if cfg!(target_arch = "wasm32")
+                        && !self.startup_options.is_in_notebook
+                        && !is_example
+                    {
+                        use std::sync::Once;
+                        static ONCE: Once = Once::new();
+                        ONCE.call_once(|| {
```
jprochazk, line 1999: This will show up every time the user opens a new web viewer, or reloads the page. That seems overly spammy. I _hate_ when websites keep prompting me to do something, especially if I've already done it. I believe if it has to work this way, then it's better to not do this at all. It may drive users away.

(the other notification changes in this PR seem valuable, though)

jprochazk: From the Slack thread, the important part that would make this fine in my opinion (emphasis mine):

> when opening webviewer _first time_ with _your own data_

The important part is ensuring that this doesn't happen all the time. Once per domain (by persisting a `user_has_seen_native_viewer_tip` flag in `localStorage`) seems OK.

emilk: Yeah, good point… ideally it should be shown until the user has clicked "Dismiss" on it, and then never again.

crates/viewer/re_selection_panel/tests/snapshots/selection_panel_component_hybrid_overwrite.png
lucasmerlin, line 1: There seem to be a lot of unrelated snapshot changes 🤔 Are these related to your changes?

lucasmerlin (approved):
Nice!

crates/viewer/re_selection_panel/tests/snapshots/selection_panel_component_hybrid_overwrite.png
emilk: I fixed the notification icons - they had the wrong dimension.

Before:

Image

After:

Image

crates/store/re_log_types/src/index/timestamp.rs
```diff
@@ -36,6 +36,11 @@ impl Timestamp {
     pub fn nanos_since_epoch(self) -> i64 {
         self.0
     }
+
+    #[inline]
+    pub fn elsapsed(self) -> Duration {
```
benliepert, line 41: minor typo

emilk: Oops!
* https://github.com/rerun-io/rerun/pull/11134

PR #11117, opened 2025-09-04
Add data inline viewer validity test

### LINKED ISSUE #10831: Missing rrd file in online documentation
Web viewer of the `Updating a scalar over time, in a single operation` section fails to load due to a missing rrd file:
https://rerun.io/docs/howto/logging/send-columns#updating-a-scalar-over-time-in-a-single-operation

Image

IsseW:
### Related

* Closes #10831

### What

Fixes an invalid snippet reference and adds a test for checking that `data-inline-viewer` tags actually point to a valid reference.

### DIFF
--- a/docs/content/howto/logging/send-columns.md
+++ b/docs/content/howto/logging/send-columns.md
@@ -36,7 +36,7 @@ which can be translated to the column-oriented `send_columns` API as such:
 
 snippet: archetypes/scalars_column_updates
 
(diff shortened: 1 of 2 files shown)


IsseW:
Example for found error https://github.com/rerun-io/rerun/actions/runs/17467264344/job/49606110965?pr=11117#step:5:25

Issue #6523, opened 2024-06-08
Request: disable replay when connecting new gRPC clients


carueda:
**Is your feature request related to a problem? Please describe.**

(from [discord](https://discord.com/channels/1062300748202921994/1247634109992534097):)
I'm also facing the "issue" of having new clients (in my case, with embedded rerun viewer in browser) to receive, upon fresh connection, the re-playing from the beginning of whatever has previously been captured by the websocket server.  It would be great if one could indicate no such replaying for any new client connection.

I note that the `server_memory_limit` parameter in `rr.serve(),` may help address this at least to some extent.

**Describe the solution you'd like**

.. Thanks @jleibs , as you say in discord:

_An explicit option for the server to disable replay, or an option for the client when initiating the connection both sound like reasonable features though._

**Describe alternatives you've considered**

**Additional context**

Issue #10843, opened 2025-08-08
Video play broken with time ranged links


Wumpf:
`pixi run rerun-release 'rerun://sandbox.redap.rerun.io:443/dataset/1859951D49D9F54F053110352a835641?partition_id=ILIAD_49fe161f_2023_06_27_15h_52m_49s&time_range=real_time@2023-07-13T04:41:06.567641088Z..2023-07-13T04:41:15.114329088Z'`

This correctly fetches ~70MB / 16GB from this RRD. But the Video player gives a warning about "out of order samples" and then refuses to play.
https://rerunio.slack.com/archives/C07Q6HLG24S/p1754652218751459

(this rrd is merged together from a lot of different RRDs)

Could be that we have to improve how we handle out of order arriving samples. To be investigated what actually fails here

emilk:
There are two different issues to worry about:
* Individual chunks being "unordered" (not all timelines increasing monotonically with RowId)
* Chunks arriving out-of-order

I _believe_ the "out of order samples" refers to the first of these problems, and could probably be fixed by sorting the chunk by the active timeline.

IsseW:
Additional information from a deep-dive about this today:
- This issue is not specific to time ranged links and happens when just opening the recording manually as well.
- Chunks themselves are in order, looking at component time ranges. And samples within individual chunks are ordered. But in some cases different chunks' time ranges overlap, causing the first sample from a later chunk be earlier than the last sample in the earlier chunk. Which causes the warning, and that entire chunk to be ignored.

The working theory now is that since these recordings are merges of many different recordings video streams from multiple sources might've ended up at the same time. Causing the video not to load.

It might also be possible to detect that this specific case is happening and display a better warning for it.

IsseW:
Gotten some more insight into what's happening with this bug now. And there are two parts to it.
One part is the overlapping video like we thought, and that matches up with some of my debugging. Where two or more of the merged recordings were in the same time frame.

The other part of this is that for streamed videos the order in which the files were merged in matters. And for the streamed videos to work you have to merge the recording from an earlier time stamp before the one with the earlier timestamp.

So this will work:
`rerun rrd route --recording-id test Oct__7.rrd Oct_20.rrd > merged.rrd`
This won't work:
`rerun rrd route --recording-id test Oct_20.rrd Oct__7.rrd > merged.rrd`

Given that the same entity path in both recordings both have a video stream.

[...]

PR #11118, opened 2025-09-04
Add options to gRPC proxy server to replay _newest_ data first

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/6523

### What
This adds options (to the SDK and CLI) to control the replay behavior of the gRPC proxy server.

The default is (like before) to replay the oldest data first, and then start relaying live data.
But now you can opt-in to getting the live data _first_, and afterwards getting the historical data _starting with the newest_.

### TODO
* [x] Automated test
* [x] Manual test

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -45,6 +45,27 @@ See the many resources available online more details.
 You can ensure that everything is correctly installed by running `git lfs ls-files` from the repository root.
 It should list some test snapshot files.
 
+### Useful git-lfs commands
+```
+# Install git-lfs in the repo (installs git hooks)
+git lfs install
+
+# Move a file to git lfs
+git lfs track "path/to/file/or/pattern" # OR manually edit .gitattributes
+git add --renormalize . # Moves already added files to lfs (according to .gitattributes)
+
+# Move a file from lfs to regular git
+git lfs untrack "path/to/file/or/pattern" # OR manually edit .gitattributes
+git add --renormalize . # Moves already added files to regular git (according to .gitattributes)
+
+# Push to a contributor remote (see https://github.com/cli/cli/discussions/8794#discussioncomment-8695076)
+git push --no-verify
+
+# Push git lfs files to contributor remote:
+git push origin $(git branch --show-current) && git push --no-verify && git push origin --delete $(git branch --show-current)
+```
+
+See also [this section in the egui docs](https://github.com/emilk/egui/blob/eceb0b11c9f8e68d96c33c1f4af0335b3e66376d/CONTRIBUTING.md#working-with-git-lfs).
 
 ## Validating your environment
 You can validate your environment is set up correctly by running:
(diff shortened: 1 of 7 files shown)


crates/store/re_grpc_server/src/priority_stream.rs
```diff
@@ -0,0 +1,41 @@
...
+    type Item = S1::Item;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        if let Poll::Ready(item) = self.high_priority.as_mut().poll_next(cx) {
+            Poll::Ready(item)
+        } else {
+            self.low_priority.as_mut().poll_next(cx)
+        }
+    }
+}
```
emilk, line 41: Thank you Claude 🙏

crates/store/re_grpc_server/src/lib.rs
```diff
@@ -540,11 +584,23 @@ impl MessageBuffer {
...
+
+        match playback_behavior {
+            PlaybackBehavior::OldestFirst => {
+                itertools::chain!(persistent.iter(), static_.iter(), disposable.iter())
+                    .cloned()
+                    .collect()
+            }
+            PlaybackBehavior::NewestFirst => itertools::chain!(
+                persistent.iter().rev(),
+                static_.iter().rev(),
```
jprochazk, line 598: I'd be curious what happens if you have a `VideoFrameReference` and static `AssetVideo`, and then the frame references come before the video 🤔. But `persistent` and `static` probably don't need to be reversed, right?

emilk: Our viewer _should_ be robust to out-of-order chunks, so likely you will have an on-screen video-error until the video chunk is received. More likely this feature will be used by people doing video _streams_ though.

As for static: some people use static logging as a way to log "transient" data (e.g. many images), and in that case we again want to show the latest first, hence the .rev

jprochazk: > some people use static logging as a way to log "transient" data (e.g. many images), and in that case we again want to show the latest first, hence the .rev

if only we had an in-memory server which used an actual chunk store so we could discard static data when it is overwritten :sweat_smile:

Wumpf: While it will work fine'ish with `VideoFrameRefernece`, we actually completely break down with out of order chunks for `VideoStream`. Need to create a follow-up issue about this.

We have other places where we hit this limitation. E.g.:
* https://github.com/rerun-io/rerun/issues/10843

it's not entirely trivial to fix this, but we have to look into it

Wumpf (approved):
lgtm!

`VideoStream` issues are a major concerns though, we need to follow-up on this. Can you please create a dedicated ticket and prioritize it accordingly? thx

emilk:
* https://github.com/rerun-io/rerun/issues/11146

PR #11120, opened 2025-09-05
Force callers of `data_label` to think about syntax highlight

lucasmerlin:
### Related
* #11075 

### What

This removes `impl SyntaxHighlighted` for `&str` and `String`, since it's not obvious wheter this should be shown as a identifier, string literal or body text.
`data_label` now takes `impl Into<SyntaxHighlightedBuilder>` to force the caller to think about syntax highlighting.
I've updated all calls to `data_label` with proper syntax highlighting. 
This also fixes arrow string values not being shown as syntax highlighted.

### DIFF
--- a/crates/viewer/re_ui/src/syntax_highlighting.rs
+++ b/crates/viewer/re_ui/src/syntax_highlighting.rs
@@ -46,8 +46,8 @@ impl std::fmt::Debug for SyntaxHighlightedStyle {
             Self::Syntax => write!(f, "Syntax"),
             Self::Body => write!(f, "Body"),
             Self::BodyItalics => write!(f, "BodyItalics"),
-            Self::Custom(_) => write!(f, "Custom(...)"),
-            Self::CustomClosure(_) => write!(f, "CustomClosure(...)"),
+            Self::Custom(_) => write!(f, "Custom(…)"),
+            Self::CustomClosure(_) => write!(f, "CustomClosure(…)"),
         }
     }
 }


crates/viewer/re_arrow_ui/src/arrow_ui.rs
```diff
@@ -9,29 +8,9 @@ use crate::show_index::ArrayUi;
...
     re_tracing::profile_function!();
 
-    use arrow::array::{LargeStringArray, StringArray};
-
     ui.scope(|ui| {
         ui.style_mut().wrap_mode = Some(egui::TextWrapMode::Truncate);
 
-        // Special-treat text.
-        // This is so that we can show urls as clickable links.
-        // Note: we match on the raw data here, so this works for any component containing text.
```
emilk, line 19: So what happens when a user puts a link in a table now? Are they no longer clickable?

crates/viewer/re_arrow_ui/tests/snapshots/arrow_ui.png
emilk, line 1: Oh, it is still clickable? Nice!

crates/viewer/re_component_ui/src/datatype_uis/float_drag.rs
```diff
@@ -77,7 +78,13 @@ pub fn edit_f32_float_raw_with_speed_impl(
                 .suffix(suffix),
         )
     } else {
-        UiLayout::List.data_label(ui, format!("{}{}", re_format::format_f32(**value), suffix))
+        // UiLayout::List.data_label(ui, format!("{}{}", re_format::format_f32(**value), suffix))
```
emilk: rm

crates/viewer/re_dataframe_ui/tests/snapshots/filter_ui_string_contains.png
emilk, line 1: I looks like the space before "contains" used to be too thin, and now it is instead too wide?

crates/viewer/re_ui/src/syntax_highlighting.rs
```diff
@@ -6,36 +6,74 @@ use re_log_types::{
...
+}
+
 /// Easily build syntax-highlighted text.
-pub struct SyntaxHighlightedBuilder<'a> {
-    pub style: &'a Style,
-    pub tokens: &'a DesignTokens,
-    pub job: LayoutJob,
+pub struct SyntaxHighlightedBuilder {
+    text: String,
+    parts: Vec<SyntaxHighlightedPart>,
```
emilk: We could consider using a `SmallVec` here since most things with syntax-highlighting are rather short

crates/viewer/re_ui/src/ui_layout.rs
```diff
@@ -94,29 +95,35 @@ impl UiLayout {
     /// Show data while respecting the given UI layout.
     ///
     /// Import: for data only, labels should use [`UiLayout::label`] instead.
+    /// Make sure to use the right syntax highlighting.
```
emilk: How do I know which is the right syntax highlighting?

emilk (approved):
I love that Rerun is finally getting some color 😍

crates/viewer/re_ui/src/ui_layout.rs
lucasmerlin: Added a link to the SyntaxHighlightedBuilder docs

Issue #11009, opened 2025-08-26
Firefox @ WebGPU crashes on video playback


Wumpf:
```
panicked at C:\Users\Andreas\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\wgpu-25.0.2\src\backend\webgpu.rs:2562:14:
called `Result::unwrap()` on an `Err` value: JsValue(TypeError: GPUQueue.copyExternalImageToTexture: 'source' member of GPUCopyExternalImageSourceInfo could not be converted to any of: ImageBitmap, HTMLImageElement, HTMLCanvasElement, OffscreenCanvas.
__wbg_get_imports/imports.wbg.__wbg_copyExternalImageToTexture_a2037621913ee52d/<@http://localhost:9090/re_viewer.js:1204:18
handleError@http://localhost:9090/re_viewer.js:40:22
__wbg_get_imports/imports.wbg.__wbg_copyExternalImageToTexture_a2037621913ee52d@http://localhost:9090/re_viewer.js:1203:93
re_viewer.wasm.__wbg_copyExternalImageToTexture_a2037621913ee52d externref shim@http://localhost:9090/re_viewer.js line 816 > WebAssembly.instantiate:wasm-function[116472]:0x4bbb844
re_viewer.wasm.<wgpu::backend::webgpu::WebQueue as wgpu::dispatch::QueueInterface>::copy_external_image_to_texture::h3b37f1e3f536d2e0@http:/…
```

External video textures for WebGPU aren't implemented in Firefox yet.
We should figure out how to handle this issue (do we need to patch wgpu?) and not pick Firefox WebGPU unless explicitely opted in.

Confusingly, the video output callbacks may still fire, which can give you different, very confusing callstacks that drown out the original source. This led to:
* https://github.com/rerun-io/rerun/pull/11007

IsseW:
Created an issue on wgpu for the unwrapping that causes this crash https://github.com/issues/created?issue=gfx-rs%7Cwgpu%7C8186

IsseW:
With #11122 we now default to webGL for firefox, making firefox not crash with normal usage at least. But that should be reverted when this issue is fixed.

PR #11122, opened 2025-09-05
Default to webGL on firefox web-viewer

IsseW:
### Related

Related to #11009 and https://github.com/gfx-rs/wgpu/issues/8186

### What

Because videos crash on webgpu & Firefox we shouldn't default to using webgpu there. You can still restart the web-viewer with webgpu on Firefox after this. And when this has been fixed in Firefox we can revert this change.


Wumpf (approved):
👍

PR #11126, opened 2025-09-05
If the component filter is explicitly set, do not override it

timsaucer:
### Related

Closes internal report. When setting explicitly the filter component columns on the client connecter, it is getting overwritten by the filter pushdown.

### What

Check if filter is explicitly set. If not, they attempt filter pushdown.

### DIFF
--- a/rerun_py/tests/unit/test_datafusion_tables.py
+++ b/rerun_py/tests/unit/test_datafusion_tables.py
@@ -183,8 +183,8 @@ def test_component_filtering(server_instance: tuple[subprocess.Popen[str], Datas
 
     filter_on_dataframe = (
         dataset.dataframe_query_view(index="time_1", contents="/**")
-        .filter_is_not_null(component_path)
         .df()
+        .filter(col(component_path).is_not_null())
         .collect_partitioned()
     )
 


rerun_py/tests/unit/test_datafusion_tables.py
```diff
@@ -162,6 +162,40 @@ def test_df_aggregation(server_instance: tuple[subprocess.Popen[str], DatasetEnt
...
+        dataset.dataframe_query_view(index="time_1", contents="/**")
+        .filter_is_not_null(component_path)
+        .df()
+        .collect_partitioned()
+    )
+
+    filter_on_dataframe = (
+        dataset.dataframe_query_view(index="time_1", contents="/**")
+        .filter_is_not_null(component_path)
+        .df()
```
jleibs: I assume you meant this to be 
```suggestion
        .df()
        .filter(col(component_path).is_not_null())
```

jleibs (approved):

rerun_py/tests/unit/test_datafusion_tables.py
timsaucer: Yes, thank you

PR #11127, opened 2025-09-05
Changelog generation to use native gh CLI auth

bllchmbrs:
### What

`scripts/generate_changelog.py` now uses `gh` CLI auth instead of a personal token.

code was written by robot but seems straightforward enough.

I do not know what labels to add.


ntjohnson1 (commented):
It looks like the github cli is not included in our pixi environment. Could we add that to be self contained. I tried `pixi run scripts/generate_changelog.py...` and cli was unhappy.

I agree that gh auth is nicer than the token management. In the [zombie_todos](https://github.com/rerun-io/rerun/blob/7f8dfbe796c4be7c650455eca773b8c283179fee/scripts/zombie_todos.py#L11) we are actually about to use the github python package without any auth to check our open source stuff (with a lower request limit). But that's not a required change just a potential incremental improvement.

bllchmbrs:
gh is installed via brew. I figure since this is kind of a special script / not one that everyone runs all the time, it's (a) an easy fix for anyone using it and (b) simpler to not bundle.

Also, this is just a net improvement. The fact that right now when you run it you have to login to GH, get a token, add that token, etc. is a lot more work than `brew install gh` `gh auth login` and you're one.

ntjohnson1:
> gh is installed via brew. I figure since this is kind of a special script / not one that everyone runs all the time, it's (a) an easy fix for anyone using it and (b) simpler to not bundle.
> 
> Also, this is just a net improvement. The fact that right now when you run it you have to login to GH, get a token, add that token, etc. is a lot more work than `brew install gh` `gh auth login` and you're one.

The bundling feels fairly simple. But I don't feel strongly enough to block if you disagree. Was able to confirm the change.
```diff
diff --git a/pixi.toml b/pixi.toml
index f9daf9025..f4a128937 100644
--- a/pixi.toml
+++ b/pixi.toml
@@ -616,6 +616,7 @@ tomlkit = "0.12.3.*"
 tqdm = ">=4.66.2,<4.67"           # For displaying progress in various utility scripts.
 typing_extensions = ">4.5"
 typos = ">=1.16.20"
+gh = "*" # or pick a version
```

ntjohnson1 (approved):

scripts/generate_changelog.py
```diff
@@ -95,19 +75,25 @@ def fetch_pr_info_from_commit_info(commit_info: CommitInfo) -> PrInfo | None:
...
+            check=True,
+        )
+
+        pr_data = json.loads(result.stdout)
+        labels = [label["name"] for label in pr_data["labels"]]
+        gh_user_name = pr_data["author"]["login"]
+        return PrInfo(gh_user_name=gh_user_name, pr_title=pr_data["title"], labels=labels)
+
+    except subprocess.CalledProcessError as e:
+        eprint(f"ERROR fetching PR #{pr_number}: {e.stderr.strip()}")
```
emilk: This could provide a better help message like "You need to install the GitHub CLI tools: https://cli.github.com/"

bllchmbrs: fixed

bllchmbrs:
@ntjohnson1 , I made this change. PTAL.

ntjohnson1:
Confirmed locally with `pixi run scripts/generate_changelog.py --version 0.25.0`

scripts/generate_changelog.py
```diff
@@ -48,62 +49,38 @@ class CommitInfo:
...
-                f"ERROR fetching org members {url}: {response.status_code} - {response.json().get('message', 'Unknown error')}"
-            )
-
-        json_data = response.json()
-        if not json_data:  # Empty response means we've reached the end
-            break
+    try:
+        # Use gh CLI to fetch organization members
+        result = subprocess.run(
+            ["gh", "api", f"/orgs/{OWNER}/members", "--paginate", "--jq", ".[].login"],
```
emilk, line 62: This does not return all members, for unknown reason

emilk: Ah, it can only fetch _Public_ members. Users can change themselves to be public here: https://github.com/orgs/rerun-io/people

emilk: https://rerunio.slack.com/archives/C095BFLA4V6/p1759138523996419

Issue #702, opened 2023-01-09
Draw order & (order independent) transparency


Wumpf:
Current situation
===================

Currently we do not support sorting within a `Renderer`, instead all renderer have a fixed order.
This is primarily an issue for supporting transparency as transparent objects (with (premultiplied) alpha blending) need to be *at least* be ordered back to front.
Note that additive blended objects are order independent but still need to be drawn after all opaque objects.

Steps needed
===================

If (!!) we're keeping the current monolithic `Render` approach, we can split the way to fully supporting transparency in two steps:
* [x] allowing `Render` to be invoked several times for different "phases", a prototype is being developed on `andreas/re_renderer/phases`
* [ ] Sorting within (certain) phases either by..
   * invoking every `Renderer` an arbitrary amount of time with relatively fine grained draw data
   * implementing an order independent transparency technique

Pro/Con of fine grained object sorting
-------------------------

Sorting rendering primitives by distance to camera is very common in general purpose rendering engines!
However, regardless of the other implications of an order independent transparency algorithm, *not* sorting **within** `Renderer` actually has both advantages and disadvantages:

Drawbacks:
* no front-to-back order for opaque, meaning potentially lot of (discarded) overdraw
   * fragment shader with discard/depth changes may not profit from ordering
   * this is an issue if pixel shading complexity is high and can be alleviated with a depth-first pass
   * profiling is needed to determine if this actually helps, may be very scene dependent. Some of our key vertex shading is complex (points & lines), making depth-first less attractive
* no back-to-front order for transparent, meaning rendering artifacts
  until we implement some form of order independent transparency

Advantages:
* super simple, less data to transfer & compute (don't need to know distances)
    * allows fast path from database to gpu data, close to no data examination needed
* faster in some cases
* best possible batching, we invoke every `Renderer` only once per pass, meaning minimal stage changes without the need for any sorting

Order Independent transparency (OIT)
==========================
Employing an OIT technique is _partially_ independent of the question of draw order: As noted above draw order has advantages even with OIT in place *but* is no longer *required* then.
Furthermore, depending on our expected scenes draw order alone will not solve alpha blending ordering issues either because our batches are not fine grained enough (i.e. a single draw has several front/back faces in line of sight) or triangles are overlapping (happens often in our case - e.g. line penetrating transparent box).

[...]

Issue #8226, opened 2024-11-26
LineStrips2D displayed 3D in web


karolmajek:
# Bug description

LineStrips2D are displayed with non zero 3rd axis values in Web GUI
Instead of being displayed sticking to the image plane - as in native app

## Image+LineStrips2D natively [Reference - this wokrs good!]
```
rr.init("rerun_example_image", spawn=True)
# rr.serve_web()
```

![image](https://github.com/user-attachments/assets/ffe10182-5957-4e9c-a9fd-35ba32a68640)

## Image+LineStrips2D in Web GUI [this behavior is unexpected/surprising/just wrong]

```
rr.init("rerun_example_image", spawn=False)
rr.serve_web()
```
trajectories are in 3D in front of the image

![image](https://github.com/user-attachments/assets/69af2f31-c22c-4910-b511-6fcaff3de301)
![image](https://github.com/user-attachments/assets/3ad62b91-5c1e-4b3a-8f21-945fa19d0b6a)
![image](https://github.com/user-attachments/assets/1a113a24-133f-48a7-b525-91a21b359de2)
![image](https://github.com/user-attachments/assets/14404b07-8ef7-4ea3-8d79-0645aa44590a)
![image](https://github.com/user-attachments/assets/4a407e26-730e-4704-b940-3892de99b838)

**To Reproduce**
```
import numpy as np
import rerun as rr
import random

img = np.ones((1080, 1920, 3))

rr.init("rerun_example_image", spawn=False)
rr.serve_web()

rr.log("image", rr.Image(img[:, :, ::-1]))


# Parameters
n = 1000  # Number of trajectories
xmin, xmax = 0, img.shape[1]  # X range
ymin, ymax = 0, img.shape[0]  # Y range
step_min, step_max = 0.1, 2.0  # Minimum and maximum step size

# Generate and log trajectories
for i in range(n):
    # Generate random length for the trajectory
    trajectory_length = random.randint(3, 1000)

    # Start the trajectory at a random initial point
    x, y = random.uniform(xmin, xmax), random.uniform(ymin, ymax)
    trajectory = [[x, y]]

    for _ in range(trajectory_length - 1):
        # Generate a small random step
        step_x = random.uniform(-step_max, step_max)
        step_y = random.uniform(-step_max, step_max)

        # Update the current position with constraints
        x = max(xmin, min(xmax, x + step_x))
        y = max(ymin, min(ymax, y + step_y))

        trajectory.append(
            [
                x,
                y,
[...]

Issue #8066, opened 2024-11-11
&nbsp;Add a `GeoPolygon` archetype


abey79:

Wumpf:
@karanvir-jassal-ai no progress has been made, we didn't have time to extend the palette of supported primitives much. Is this a feature you're depending on?

Issue #7193, opened 2024-08-15
Filled Box2D


Famok:
**Describe the solution you'd like**
Very simple: give Box2D a fill option to fill with RGBA or RGB and an opacity value. (Boxes3D can be made solid)

**Describe alternatives you've considered**
There is no alternative as far as I know.

Wumpf:
Blocked by
* https://github.com/rerun-io/rerun/issues/1611
* https://github.com/rerun-io/rerun/issues/702

Related to
* https://github.com/rerun-io/rerun/issues/1361
* https://github.com/rerun-io/rerun/issues/6012

PR #11131, opened 2025-09-07
Add the ability to sort `Drawables` (new subdivision of `DrawData`) to `re_renderer`

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/702
* Unblocks https://github.com/rerun-io/rerun/issues/1025
* Unblocks https://github.com/rerun-io/rerun/issues/4787
* Unblocks https://github.com/rerun-io/rerun/issues/8226
* Unblocks https://github.com/rerun-io/rerun/issues/1611
* Unblocks https://github.com/rerun-io/rerun/issues/7193
* Unblocks https://github.com/rerun-io/rerun/issues/8066
* Unblocks https://github.com/rerun-io/rerun/issues/6009 (some more challenges here though)
* Unblocks https://github.com/rerun-io/rerun/issues/4106
* Somewhat related to https://github.com/rerun-io/rerun/issues/4529
* Based on https://github.com/rerun-io/rerun/pull/11082

### What

Bit of an architectural shift on how we schedule drawing in re_renderer. Previously, we'd only supplied `DrawData` and drew them for every draw phase that made sense.
Now we additionally collect `Drawables` which refer into `DrawData` but unlike `DrawData` they are fixed size & super lightweight, allowing us to sort them depending on a phase's need i.e. far to near for transparent and aggressively batching for everything else.
This also means that `Renderer::draw` is now only called once in opaque draw phases with all collected draw data & drawables.
(it would be great to collect statistics about this, but that hasn't been implemented yet)

An updated overview of the drawing life cycle can be found in re_renderer's lib.rs

* [x] pass full ci

### DIFF
--- a/crates/viewer/re_renderer/src/context.rs
+++ b/crates/viewer/re_renderer/src/context.rs
@@ -4,7 +4,7 @@ use std::sync::{
 };
 
 use parking_lot::{MappedRwLockReadGuard, Mutex, RwLock, RwLockReadGuard};
-use type_map::concurrent::{self, TypeMap};
+use type_map::concurrent::TypeMap;
 
 use crate::{
     FileServer, RecommendedFileResolver,
@@ -12,7 +12,7 @@ use crate::{
     device_caps::DeviceCaps,
     error_handling::{ErrorTracker, WgpuErrorScope},
     global_bindings::GlobalBindings,
-    renderer::Renderer,
+    renderer::{Renderer, RendererExt, RendererTypeId},
     resource_managers::TextureManager2D,
     wgpu_resources::WgpuResourcePools,
 };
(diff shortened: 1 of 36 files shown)


crates/viewer/re_renderer/src/context.rs
```diff
@@ -136,23 +136,71 @@ pub struct RenderContext {
...
-    pub fn get<R: 'static + Renderer>(&self) -> Option<&R> {
-        self.renderers.get::<R>()
+    pub fn get<R: 'static + Renderer>(&self) -> Option<&RendererWithKey<R>> {
+        self.renderers.get::<RendererWithKey<R>>()
+    }
+
+    /// Gets a renderer by its key.
+    ///
+    /// For this to succeed, the renderer must have been initialized prior.
+    /// (there would be no key otherwise anyways!)
```
IsseW, line 219: I feel like this comment would make more sense if `RendererTypeId` was a newtype that was only constructed in `get_or_create`.

crates/viewer/re_renderer/src/draw_phase_manager.rs
```diff
@@ -0,0 +1,473 @@
...
+use crate::{
+    GpuRenderPipelinePoolAccessor, QueueableDrawData, RenderContext,
+    context::Renderers,
+    renderer::{
+        DrawDataDrawable, DrawDataDrawablePayload, DrawInstruction, DrawableCollectionViewInfo,
+        RendererTypeId,
+    },
+};
+
+/// Darw data id within the [`DrawPhaseManager`].
```
IsseW:
```suggestion
/// Draw data id within the [`DrawPhaseManager`].
```

crates/viewer/re_renderer/src/draw_phase_manager.rs
```diff
@@ -0,0 +1,473 @@
...
+/// This tightly packs the two values into a single u32 for sorting.
+/// The renderer key forms the first 8 significant bits, the draw data the remaining 24.
+/// This way, sorting the this value ascending, will sort the draw data by renderer and then by draw data index.
+///
+/// Note that a single [`DrawDataIndex`] can only ever refer to a single [`RendererTypeId`].
+/// Therefore, we could alternatively pre-sort draw data by renderer so that the resulting
+/// [`DrawDataIndex`] are already grouped by renderer.
+/// However, using just the higher 8 bits for [`RendererTypeId`] makes the process a lot simpler.
+/// We may reconsider this if we change the design such that variations of renderers are
+/// expressed in the [`RendererTypeId`] such that 8bit are no longer sufficient.
```
IsseW:
```suggestion
/// expressed in the [`RendererTypeId`] such that 8 bit are no longer sufficient.
```

crates/viewer/re_renderer/src/draw_phase_manager.rs
```diff
@@ -0,0 +1,473 @@
...
+    }
+}
+
+/// A single drawable item within a given [`crate::renderer::DrawData`].
+///
+/// For more details see [`DrawDataDrawable`].
+/// This is an expanded version used for processing/sorting.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Drawable {
+    /// Distance sort key from near (0.0) to far (float max).
```
IsseW: Thinking about what you said in the meeting earlier, about not being specific. Is there any reason this can't be from `-INF` to `INF`

crates/viewer/re_renderer/src/queueable_draw_data.rs
```diff
@@ -1,59 +1,77 @@
...
+    type Target = dyn TypeErasedDrawData + Send + Sync;
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        self.0.as_ref()
+    }
+}
+
+impl QueueableDrawData {
+    /// Panics if the type `T` is not the underlying type of this draw data.
```
IsseW:
```suggestion
    /// Panics if the type `D` is not the underlying type of this draw data.
```

crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
```diff
@@ -89,11 +91,15 @@ mod gpu_data {
...
+    instance_end_index: u32,
 
-    /// Number of meshes out of `count` which have outlines.
-    /// We put all instances with outlines at the start of the instance buffer range.
-    count_with_outlines: u32,
+    /// Last index for meshes with outlines
+    ///
+    /// We put all instances with outlines at the start of the instance buffer range, so this is always
+    /// smaller or equal to `instance_end_index`.
+    /// If it is equal to `instance_end_index`, there are no meshes with outlines in this batch.
```
IsseW: Is this comment outdated? At line 125 it checks equality with `instance_start_index` instead of `instance_end_index`

crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
```diff
@@ -428,55 +447,60 @@ impl Renderer for MeshRenderer {
...
-            pass.set_index_buffer(
-                index_buffer.slice(mesh_batch.mesh.index_buffer_range.clone()),
-                wgpu::IndexFormat::Uint32,
-            );
-
-            let num_meshes_to_draw = if phase == DrawPhase::OutlineMask {
-                mesh_batch.count_with_outlines
-            } else {
-                mesh_batch.count
+        // TODO(andreas): use drawables to orchestrate drawing.
```
IsseW: What does this TODO mean? Because I feel like that's kind of what we're doing 😄

IsseW (changes requested):
very cool stuff!

crates/viewer/re_renderer/src/draw_phase_manager.rs
Wumpf: no reason, comment is just wrong :D

crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
Wumpf: it's to be fixed properly in the follow-up PR, but you're right it's wrong. dropping it

crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
Wumpf: oops! don't think this was ever true

Wumpf:
@rerun-bot full-check

Issue #1741, opened 2023-03-31
Allow hiding back/front faces of meshes


Wumpf:
All meshes are rendered double sided currently since we don't know what is front/back (it's hard to enforce a standard!). This is a resonable default for Rerun, but there should be a blueprint option for backface culling where one can specificy whether clockwise/counterclockwise triangles are shown. Need to figure out good names for this.

emilk:
Will help the ARKit example

Wumpf:
Bunch of things to do for the proper solution:

* separate semantics of mapping front/back <-> CW/CCW and showing All/front/back
* read this mapping from mesh format if defined (many formats do!)
* establish what our default mapping is
* APIs for all of the above

emilk:
This will blow up the required number of mesh pipelines by x3 if I'm not mistaken

emilk:
Related: https://github.com/rerun-io/rerun/issues/701

emilk:
This is both a component, and should be overridable in the blueprint store

Issue #1611, opened 2023-03-18
Varying opacity for spatial primitives


roym899:
**Is your feature request related to a problem? Please describe.**
Varying opacity can be useful to visualize overlapping or occluded data.

Seems like most primitives support RGBA arguments, however the alpha channel was ignored for the one I tested (`rr.log_line_segments`) or didn't correspond to opacity (`rr.log_mesh`).

**Describe the solution you'd like**
Would be great if all primitives (including 3D primitives) support some form of transparency / alpha compositing.

Wumpf:
Partially blocked by #702 - in planning this week we decided to get in some more transparency without order independence. I.e. we'll move forward to supporting it, but visual bugs will be expected when several transparent surfaces overlap on screen.

shabbirmarzban:
Hi there, what is the current status (of this feature)? I particularly would like to use it for meshes, thanks.

Wumpf:
@shabbirmarzban no progress and no eta so far sadly. But curious to hear what you're looking for exactly - e.g. if you'd get transparency for meshes how do you want to control it? (uniformly/per-vertex/via-texture, data via custom mesh or mesh format or set via viewer etc.)

julianschoep:
From a user perspective, for me the most straight-forward would be to have an opacity slider next to the entity visibility toggle, so entity-level control in the UI. Even if it's not ordered correctly it will add a lot of value for us.

Ipuch:
I have interest for this feature too

nshankar:
I also have interest in transparency. Are there any examples of how to correctly render 3D points with RGBA values? Here is a minimal example where the transparent point is rendered as opaque

```
import rerun as rr
from numpy.random import default_rng

rr.init("rerun_example_points3d_random", spawn=True)
rng = default_rng(12345)

positions = rng.uniform(-5, 5, size=[10, 3])
colors = rng.uniform(0, 255, size=[10, 3])
radii = rng.uniform(0, 1, size=[10])

rr.log(
    "transparent",
    rr.Points3D(positions=[(0, 0, 0)], colors=(0.5, 0.5, 0.5, 0.1), radii=1),
)
rr.log("random", rr.Points3D(positions, colors=colors, radii=radii))
```

Wumpf:
@nshankar this issue is about transparency not being available yet, i.e. you can't achieve this yet api code

georgematheos:
Hi, I just wanted to upvote this again.

[...]

PR #11132, opened 2025-09-07
Support basic transparency for `Mesh3D`/`Asset3D`/`Boxes3D`/`Ellipsoids3D`/`Capsules3D`/`Cylinders3D`

Wumpf:
### Related

* Part of https://github.com/rerun-io/rerun/issues/1611
* Exacerbates https://github.com/rerun-io/rerun/issues/1741
    * we just assume GLTF winding order for now. If you don't follow that transparency looks weird.

### What

Uses the newly established `Drawables` sorting mechanism to implement classic far to near rendering for all types of meshes.

Comes with image comparison tests for all archetypes where this is possible now, plus some unit tests to test the updated mesh renderer draw data buildup & drawable emission.

https://github.com/user-attachments/assets/3c182dc5-52cd-418b-9021-3db667bf8651


docs/snippets/all/archetypes/mesh3d_instancing.cpp
```diff
@@ -12,7 +12,7 @@ int main() {
         rerun::Mesh3D(
             {{1.0f, 1.0f, 1.0f}, {-1.0f, -1.0f, 1.0f}, {-1.0f, 1.0f, -1.0f}, {1.0f, -1.0f, -1.0f}}
         )
-            .with_triangle_indices({{0, 1, 2}, {0, 1, 3}, {0, 2, 3}, {1, 2, 3}})
+            .with_triangle_indices({{0, 2, 1}, {0, 3, 1}, {0, 3, 2}, {1, 3, 2}})
```
Wumpf, line 15: makes the winding order ccw so that if you make them transparent they don't look awful

crates/viewer/re_renderer/src/context_test.rs
```diff
@@ -1,20 +1,30 @@
 //! Extensions for the [`RenderContext`] for testing.
 //!
+use std::sync::LazyLock;
+
 use crate::{RenderConfig, RenderContext, device_caps};
 
+/// Adapter shared with all tests.
+///
+/// Devices are not re-used since we want to isolate wgpu resources.
+static TEST_ADAPTER: LazyLock<wgpu::Adapter> = LazyLock::new(|| {
```
Wumpf, line 10: could go into a separate PR. But on my Windows box this turned out necessary for executing the new tests - there seems to be some race condition in wgpu (or my driver?) when spamming creation of new adapters in parallel. (also, it's slow ;))
Worth deeper investigation from a wgpu pov, but need to isolate it first

crates/viewer/re_renderer/shader/instanced_mesh.wgsl
```diff
@@ -77,14 +79,15 @@ fn fs_main_shaded(in: VertexOut) -> @location(0) vec4f {
         default: { texture = vec3f(0.0); }
     }
 
-    let albedo = texture
-                 * in.color.rgb
-                 * material.albedo_factor.rgb
-                 + in.additive_tint_rgb;
+    var albedo = vec4f(texture * in.color, 1.0) * material.albedo_factor;
```
Wumpf, line 86: should explain why we ignore alpha on vertex color & texture (very not obvious!)

crates/viewer/re_renderer/src/context_test.rs
jprochazk: Drive-by: Recently saw a similarly shaped problem elsewhere, but on a different platform (loading libdrm amdgpu in parallel tests). The problem was calling `dlopen`/`LoadLibrary` in parallel. Putting it behind a lock was the solution there also

crates/store/re_types/definitions/rerun/archetypes/mesh3d.fbs
```diff
@@ -9,6 +9,9 @@ namespace rerun.archetypes;
 /// If there are multiple [archetypes.InstancePoses3D] instances logged to the same entity as a mesh,
 /// an instance of the mesh will be drawn for each transform.
 ///
+/// The viewer draws meshes always two-sided. However, for transparency ordering
```
jprochazk, line 12: Drive-by: Are there plans to expose an option to disable this? I was a bit surprised that the meshes in the demo posted on socials were two-sided 🤔 (Sorry if this is already answered elsewhere and I missed it)

jprochazk: I did miss it, it's even linked in the PR description https://github.com/rerun-io/rerun/issues/1741 disregard! :facepalm:

Wumpf: but I should also link the issue here

IsseW:
I noticed opacity of images in 3d is fully broken here:
Screenshot 2025-09-09 at 17 18 31

It is broken in the current version too, but you can see meshes through them:
Screenshot 2025-09-09 at 17 18 58

Since it was already quite broken I think it makes sense to put this in an issue.

Wumpf:
good callout. I think we can mitigate it a bit by drawing the images last in the opaque phase until we do the right thing and put transparent images on the transparent phase & take their draw order into account. There's some trickiness on how we combine 2D draw order in 3D, but I'm sure we'll figure something out

Wumpf:
filed an issue here
* https://github.com/rerun-io/rerun/issues/11156

should link it in the code

crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
```diff
@@ -284,19 +290,62 @@ impl MeshDrawData {
...
+                    for chunk in instances.chunk_by(|(_, phases_a), (_, phases_b)| {
+                        phases_a.contains(phase) == phases_b.contains(phase)
+                    }) {
+                        if !chunk[0].1.contains(phase) {
+                            continue;
+                        }
+
+                        // SAFETY: `chunk` and `instances` are both from the same allocation with chunk being a subset of instances.
+                        #[expect(unsafe_code)]
+                        let start_idx = unsafe { chunk.as_ptr().offset_from(instances.as_ptr()) };
```
IsseW: Not sure what the policy is on unsafe code here, but this could be pretty easily avoided by having an additional counter and adding `chunk.len()` to it each iteration.

<sub>This also led me to search for instances of unsafe and find several undocumented ones 🙈 </sub>

Wumpf: hmmmmm yes. will fix. That's the old C++ corner in my brain wanting to do pointer arithmetic real badly.

> This also led me to search for instances of unsafe and find several undocumented ones 🙈 

totally offtopic; but now I'm curious. Found any scary ones? Thought most of it is the weird storage engine thing, env vars and C FFI.

IsseW: In `non_min_i64.rs`, and `datatypes/selected_columns.rs` which looks to be the storage engine thing. Wouldn't really say scary, but some are a bit non-obvious at first glance.

IsseW:
I think these changes look good!

So after going through the checkboxes, and the things you wanted to note down in code this should be good to go.

crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
Wumpf: in `non_min_i64.rs` defense, that one is vendored 😅 

re commented-on code: fixing it I agree that this was very silly to resort to unsafe

Wumpf:
URDF example works just fine with this! 

https://github.com/user-attachments/assets/d53bb43e-cc8f-4e23-bfcb-8e158e3a3b65

PR #11133, opened 2025-09-07
Show welcome screen when all recordings are closed

### LINKED ISSUE #10872: Closing only app_id puts viewer in undefined state
## Step 1
Load a single .rrd, e.g. dna:

Image

## Step 2
Click the close-button for the app_id:

Image

## Step 3
Scream in horor
Image

IsseW:
### Related

Closes RR-1573

Closes #10872

### What

Fixes a regression where the viewer was put in an undefined state after all recordings were closed.

### DIFF
--- a/crates/viewer/re_viewer/src/navigation.rs
+++ b/crates/viewer/re_viewer/src/navigation.rs
@@ -27,6 +27,10 @@ impl Navigation {
         self.history.push(display_mode);
     }
 
+    pub fn push_default(&mut self) {
+        self.push(self.default.clone());
+    }
+
     pub fn replace(&mut self, new_mode: DisplayMode) -> Option<DisplayMode> {
         let previous = self.history.pop();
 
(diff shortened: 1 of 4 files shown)


crates/viewer/re_viewer/src/navigation.rs
```diff
@@ -27,6 +27,10 @@ impl Navigation {
         self.history.push(display_mode);
     }
 
+    pub fn push_default(&mut self) {
```
Wumpf: I know everything here is a bit thinly documented, but I think it would be good to call out what default effectively means here. Yes, the code is written in a way that leaves the concrete default open in the future, but it's clearly some notion of start page here.
Also not sure whether `default` is such a good name actually 🤔

IsseW:
> fyi: closing the gh issue will auto-close the associated linear issue

Yeah, but I don't think it actually links the PR to linear through that.

crates/viewer/re_viewer/src/navigation.rs
IsseW: Yeah, docs in that file could improve overall. But added some to that function and renamed `default` to `start_mode`.

Wumpf:
thanks!

PR #11135, opened 2025-09-08
Refactor setting selection to always use system commands

IsseW:
### Related

Part of #10866.

### What

To be able to handle selection updates in one place to update the viewer url I made all instances of setting the selection to instead send a `SystemCommand::SetSelection`. This also means we can make selection not be in a mutex.

Part of this is also moving `ItemCollection` to `re_global_context` to be able to send it in the `SetSelection` command.

### TODO
 - [x] implement the web url updating using these changes

### DIFF
--- a/crates/viewer/re_global_context/src/command_sender.rs
+++ b/crates/viewer/re_global_context/src/command_sender.rs
@@ -91,7 +91,7 @@ pub enum SystemCommand {
     EnableInspectBlueprintTimeline(bool),
 
     /// Set the item selection.
-    SetSelection(crate::Item),
+    SetSelection(crate::ItemCollection),
 
     /// Set the active timeline and time for the given recording.
     SetActiveTime {
@@ -129,6 +129,12 @@ pub enum SystemCommand {
     /// Add a task, run on a background thread, that saves something to disk.
     #[cfg(not(target_arch = "wasm32"))]
     FileSaver(Box<dyn FnOnce() -> anyhow::Result<std::path::PathBuf> + Send + 'static>),
+}
+
+impl SystemCommand {
+    pub fn clear_selection() -> Self {
+        Self::SetSelection(crate::ItemCollection::default())
+    }
 }
 
 impl std::fmt::Debug for SystemCommand {
--- a/crates/viewer/re_global_context/src/item_collection.rs
+++ b/crates/viewer/re_global_context/src/item_collection.rs
@@ -1,0 +1,299 @@
+use indexmap::IndexMap;
+use itertools::Itertools as _;
+use re_chunk::EntityPath;
+use re_entity_db::EntityDb;
+use re_log_types::StoreKind;
+use re_types::external::glam;
+
+use crate::{Item, ViewId, resolve_mono_instance_path_item};
+
+/// Context information that a view might attach to an item from [`ItemCollection`] and useful
+/// for how a selection might be displayed and interacted with.
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+pub enum ItemContext {
+    /// Hovering/Selecting in a 2D space.
+    TwoD {
+        space_2d: EntityPath,
+
+        /// Where in this 2D space (+ depth)?
+        pos: glam::Vec3,
+    },
+
+    /// Hovering/Selecting in a 3D space.
+    ThreeD {
+        /// The 3D space with the camera(s)
+        space_3d: EntityPath,
+
+        /// The point in 3D space that is hovered, if any.
+        pos: Option<glam::Vec3>,
+
+        /// Path to an entity that is currently tracked by the eye-camera.
+        /// (None for a free floating Eye)
+        tracked_entity: Option<EntityPath>,
+
+        /// Corresponding 2D spaces and pixel coordinates (with Z=depth)
+        point_in_space_cameras: Vec<(EntityPath, Option<glam::Vec3>)>,
+    },
+
+    /// Hovering/selecting in one of the streams trees.
+    StreamsTree {
+        /// Which store does this streams tree correspond to?
+        store_kind: StoreKind,
+
+        /// The current entity filter session id, if any.
+        filter_session_id: Option<egui::Id>,
+    },
+
+    /// Hovering/selecting in the blueprint tree.
+    BlueprintTree {
+        /// The current entity filter session id, if any.
+        filter_session_id: Option<egui::Id>,
+    },
+}
+/// An ordered collection of [`Item`] and optional associated context objects.
+///
+/// Used to store what is currently selected and/or hovered.
+#[derive(Debug, Default, Clone, PartialEq, serde::Deserialize, serde::Serialize)]
+pub struct ItemCollection(IndexMap<Item, Option<ItemContext>>);
+
+impl From<Item> for ItemCollection {
+    #[inline]
+    fn from(val: Item) -> Self {
+        Self([(val, None)].into())
+    }
+}
+
+impl ItemCollection {
+    pub fn from_items_and_context(
+        items: impl IntoIterator<Item = (Item, Option<ItemContext>)>,
+    ) -> Self {
+        Self(items.into_iter().collect())
+    }
+
+    /// Is this view the selected one (and no other)?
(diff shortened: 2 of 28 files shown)


crates/viewer/re_viewer_context/src/selection_state.rs
```diff
@@ -413,20 +116,26 @@ impl ApplicationSelectionState {
...
-        *self.selection_this_frame.lock() = items.into();
+    pub fn set_selection(&mut self, items: impl Into<ItemCollection>) {
+        self.selection_this_frame = items.into();
     }
 
-    /// Extend the selection with the provided items.
-    pub fn extend_selection(&self, items: impl Into<ItemCollection>) {
-        self.selection_this_frame.lock().extend(items.into());
+    /// Sends a command to select the current selection + `items`.
+    pub fn extend_selection(
```
Wumpf: let's keep a strong(er) separation of concerns and move `extend_selection` & `toggle_selection` out of here as well: my thinking is that the selection state is just the data structure and now that we say that it  no longer has interior mutability, we no longer offer ways of changing it without `mut`. This way the delay in setting selections becomes more obvious and we don't have operations that _look_ like they directly operate on the selection state when they actually don't.


Not a super strong position since this veers into taste-territory (after all we have those dependencies anyways and the utilities have to live _somewhere_), so if this feels clunky to you or is just lotsa overhead let me know!

crates/viewer/re_global_context/src/lib.rs
```diff
@@ -8,6 +8,7 @@ mod command_sender;
 mod contents;
 mod file_dialog;
 mod item;
+mod item_collection;
```
Wumpf, line 11: [not actionable] not excited about moving things further down the crate stack, but afaik it's rather rare that we depend on `re_global_context` and not on `re_viewer_context` :thinking:

crates/viewer/re_global_context/src/item_collection.rs
```diff
@@ -0,0 +1,300 @@
+use indexmap::IndexMap;
```
Wumpf, line 1: Trusting this file is just moved code, not going through everything ;)

crates/viewer/re_global_context/src/command_sender.rs
```diff
@@ -131,6 +131,12 @@ pub enum SystemCommand {
     FileSaver(Box<dyn FnOnce() -> anyhow::Result<std::path::PathBuf> + Send + 'static>),
 }
 
+impl SystemCommand {
+    pub fn clear_selection() -> Self {
```
Wumpf, line 135: convenient! but also used in exactly one place. Overkill?

Wumpf (changes requested):
I believe the core idea of forcing everything to route through the `SystemCommand` is a clear improvement since  it makes it easier to reason about what happens when - the fact that we had two paths of achieving selection change was just weird!

However, I think the refactor on the selection state is slightly incomplete and leaves us in an odd in-between state. A lil bit of touching up is needed imho, details see comments

crates/viewer/re_global_context/src/item_collection.rs
IsseW: Yeah more or less, `into_mono_instance_path_items` I also changed to instead of taking a `ViewerContext` as parameter it takes an `EntityDb` and a `LatestAtQuery`. In hindsight would've been good to put that as a note on the PR

crates/viewer/re_global_context/src/lib.rs
IsseW: > not excited about moving things further down the crate stack

Yeah that's understandable

crates/viewer/re_viewer_context/src/selection_state.rs
IsseW: Yep agreed!

Removed those two functions. And they were used in very few places so I don't think they need to be separate functions.

Also removed the double buffering of selections, so there's only one selection collection.

crates/viewer/re_global_context/src/command_sender.rs
IsseW: Used in two places now :)

I kind of like it as
```rust
SystemCommand::SetSelection(ItemCollection::default())
```
is not as obviously clearing the selection as
```rust
SystemCommand::clear_selection()
```

crates/viewer/re_viewer_context/src/selection_state.rs
```diff
@@ -104,267 +52,19 @@ impl InteractionHighlight {
...
-    /// Selection of the previous frame. Read from this.
-    selection_previous_frame: ItemCollection,
+    /// The selected items. Write to this with [`re_global_context::SystemCommand::SetSelection`].
+    selection: ItemCollection,
 
-    /// Selection of the current frame. Write to this.
+    /// Has selection changed since the previous frame?
     #[serde(skip)]
-    selection_this_frame: Mutex<ItemCollection>,
+    selection_changed: bool,
```
IsseW, line 65: It feels like we should come up with a better way to check if this has changed now... Which is used to fire off an event dispatcher.

crates/viewer/re_viewer_context/src/selection_state.rs
```diff
@@ -104,267 +52,19 @@ impl InteractionHighlight {
...
 ///
 /// Both hover and selection are double buffered:
 /// Changes from one frame are only visible in the next frame.
 #[derive(Default, serde::Deserialize, serde::Serialize)]
 #[serde(default)]
 pub struct ApplicationSelectionState {
-    /// Selection of the previous frame. Read from this.
-    selection_previous_frame: ItemCollection,
+    /// The selected items. Write to this with [`re_global_context::SystemCommand::SetSelection`].
+    selection: ItemCollection,
```
Wumpf: the serde situation here is still super weird. I thought I remembered a bug about us serializing selection but it not working out for some reason, but I can't find the PR about that anymore even via git blame
Either way, before we serialized only the "previous_frame" which then would be discarded, now we serialize the current one which we'd keep. I'd err on the side of caution and therefore skip serialization of `selection` (which is now the `selection_this_frame` effectively) -> this then skips everything, disabling serialization on the whole thing

crates/viewer/re_viewer_context/src/selection_state.rs
Wumpf: I agree it's not very elegant and we'd better off joining the system event processing with the event dispatcher (which AFAIK is today used for in-process (rust & py notebooks)callbacks only). But I wouldn't push this refactor iteration further 🤷

crates/viewer/re_viewer_context/src/selection_state.rs
```diff
@@ -391,47 +91,39 @@ impl ApplicationSelectionState {
         re_tracing::profile_scope!("SelectionState::on_frame_start");
 
         // Purge selection of invalid items.
-        let selection_this_frame = self.selection_this_frame.get_mut();
-        selection_this_frame.retain(|item, _| item_retain_condition(item));
-        if selection_this_frame.is_empty()
+        self.selection.retain(|item, _| item_retain_condition(item));
```
Wumpf, line 91: very subtle bug here: previously, if we purged a single element from a multi-selection, we'd fire a selection change event. Now, we only fire it if we purged all of them.
Have to check count before/after `retain` to set `selection_changed`

Wumpf (changes requested):
almost 🙃

crates/viewer/re_viewer_context/src/selection_state.rs
IsseW: Huh yeah, from `AppContext`:
```rust
    /// Selection & hovering state.
    ///
    /// Not serialized since on startup we have to typically discard it anyways since
    /// whatever data was selected before is no longer accessible.
    ///
    /// For dataplatform use-cases this can even be rather irritating:
    /// if previously a server was selected, then starting with a URL should no longer select it.
    #[serde(skip)]
    pub selection_state: ApplicationSelectionState,
```

So I'll remove the serialization derive entirely since it's unused

crates/viewer/re_viewer_context/src/selection_state.rs
IsseW: Ahh yeah good catch!

Wumpf (approved):
🚢

Issue #10862, opened 2025-08-12
MCAP `MessageLayers` should only look at relevant chunks


grtlr:
Currently, we still iterate over an entire MCAP file to retrieve all the messages that are relevant for a given `dyn MessageLayer`. We should abstract of the `::mcap::Summary` and `mcap_bytes` to allow indexed access + filtering of the MCAP file.

We should be able to:

* Filter by topic
* Filter by time range

Each `MessageLayer` should specify the `encoding` or `schema` of messages that it is interested in and then get presented only those messages. This is required large MCAP files, as well as presenting more useful errors on a per-layer basis.

oceanusxiv:
@grtlr Hi, I'm curious, is this being actively worked on right now? If not, I might be interested in taking a stab at this since this is very useful or me.

oceanusxiv:
FWIW, I spent some time with #12139, it includes time filtering and switches mcap access to using the indexed reader, it should be really simple to add topic filtering also, though I did not implement the `MessageLayer` changes.

PR #11136, opened 2025-09-08
Implement fallback mode for raw MCAP layer

### LINKED ISSUE #11074: Fallback to `raw` layer when encoding isn't supported
### What

- Make `raw` layer a fallback layer if we don't support a message encoding
- Avoid duplicating data

oxkitsune:
### Related

* Closes #11074

### What

Makes the `raw` layer run as fallback when `ros2` and `protobuf` don't support a message.

### DIFF
--- a/crates/utils/re_mcap/src/layers/mod.rs
+++ b/crates/utils/re_mcap/src/layers/mod.rs
@@ -366,33 +366,6 @@ impl LayerRegistry {
         self
     }
 
-    /// Custom factory for file layer.
-    pub fn register_file_with_factory<S: Into<LayerIdentifier>>(
-        mut self,
-        identifier: S,
-        factory: fn() -> Box<dyn Layer>,
-    ) -> Self {
-        let id = identifier.into();
-        if self.file_factories.insert(id.clone(), factory).is_some() {
-            re_log::warn_once!("Inserted file layer {} twice.", id);
-        }
-        self
-    }
-
-    /// Custom factory for message layer.
-    pub fn register_message_with_factory<S: Into<LayerIdentifier>>(
-        mut self,
-        identifier: S,
-        factory: fn() -> Box<dyn MessageLayer>,
-    ) -> Self {
-        let id = identifier.into();
-        if self.msg_factories.insert(id.clone(), factory).is_some() {
-            re_log::warn_once!("Inserted message layer {} twice.", id);
-        }
-        self.msg_order.push(id);
-        self
-    }
-
     /// Configure a global fallback message layer (e.g. `raw`).
     pub fn with_global_fallback<M: MessageLayer + 'static>(mut self) -> Self {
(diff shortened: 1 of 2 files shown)


crates/utils/re_mcap/src/layers/mod.rs
```diff
@@ -247,13 +254,32 @@ impl LayerRegistry {
...
+            registry = registry
+                .register_with_factory(<McapRawLayer as Layer>::identifier(), || {
+                    Box::new(McapRawLayer::default().with_fallback_enabled(true)) as Box<dyn Layer>
+                });
+        } else {
+            registry = registry
+                .register_with_factory(<McapRawLayer as Layer>::identifier(), || {
+                    Box::new(McapRawLayer::default().with_fallback_enabled(false)) as Box<dyn Layer>
+                });
+        }
```
grtlr: This code would be much nicer, if fallbacks were a property of the registry (i.e. via an `register_fallback` method).

grtlr (changes requested):
I really like the concept of fallbacks!

However, I think we should build that functionality into the registry. The registry would then become the source-of-truth for which layers get applied to which messages. `SelectedLayers` could then just generate a new, stripped down, version of a registry.

This should allow us to:

* #10862
* Allow easier printing of the current configuration, i.e. a `rerun mcap convert --dry-run`, or a UI in Rerun settings.
* Allow extending to per-encoding fallbacks in the future.
* Get rid of the fallback code in `McapRawLayer`. A layer should not need to know in which context it is used in.

crates/utils/re_mcap/src/layers/raw.rs
```diff
@@ -55,14 +55,25 @@ impl MessageParser for RawMcapMessageParser {
...
         "raw".into()
     }
 
+    fn init(&mut self, summary: &::mcap::Summary) -> Result<(), Error> {
+        let _ = summary; // nothing to do
+        Ok(())
+    }
+
+    fn supports_channel(&self, channel: &mcap::Channel<'_>) -> bool {
+        let _ = channel;
```
grtlr:
```suggestion
    fn supports_channel(&self, _channel: &mcap::Channel<'_>) -> bool {
```

grtlr: nit

crates/utils/re_mcap/src/layers/mod.rs
```diff
@@ -209,80 +252,277 @@ impl<T: MessageLayer> Layer for T {
...
+        identifier: S,
+        factory: fn() -> Box<dyn MessageLayer>,
+    ) -> Self {
+        let id = identifier.into();
+        if self.msg_factories.insert(id.clone(), factory).is_some() {
+            re_log::warn_once!("Inserted message layer {} twice.", id);
+        }
+        self.msg_order.push(id);
+        self
+    }
```
grtlr: These are unused, right?
```suggestion
```

PR #11137, opened 2025-09-08
New link sharing dialog

### LINKED ISSUE #10814: Add share button with "link builder" for creating dataplatform links
design in [https://www.figma.com/board/dsZcCUQxyrfj9YI8ltv3JY/share?node-id=8-1383&t=15DaAvP56ppiX3JH-4](https://www.figma.com/board/dsZcCUQxyrfj9YI8ltv3JY/share?node-id=8-1383&t=15DaAvP56ppiX3JH-4)

[Video Prototype](https://rerunio.slack.com/archives/C04MTSM2U91/p1755021712663899)

Options for the time range drop-down:

* Entire recording
* Highlight time selection
* Trim to selection

Default behavior: 

* time cursor: current
* playback: paused
* time range: depends on the context 
  * preselect to `entire recording` if no time range is selected 
  * preselect to `Highlight time selection` if time range is selected

Question: do we already have a flag to decide whether to make a snapshot or not? 

By default: we want to make a snapshot (as it is closest to what users see in the viewer at that moment)

Wumpf:
### Related

* Fixes #10814

### What

design WIP


crates/viewer/re_viewer/src/open_url.rs
```diff
@@ -565,6 +565,56 @@ impl ViewerOpenUrl {
             }
         }
     }
+
+    /// Fragments of the URL if supported.
+    pub fn fragments_mut(&mut self) -> Option<&mut re_uri::Fragment> {
```
IsseW, line 567: Just realized you added this here, in #11153 I added it as a parameter to `from_display_mode`. But think I prefer this as it's explicit about whether the fragment can exists or not.

crates/viewer/re_viewer/src/ui/share_modal.rs
```diff
@@ -0,0 +1,436 @@
...
+) {
+    ui.list_item_flat_noninteractive(PropertyContent::new("Link format").value_fn(|ui, _| {
+        ui.selectable_toggle(|ui| {
+            selectable_value_with_min_width(ui, MIN_TOGGLE_WIDTH, create_web_viewer_url, false, "Only source")
+                .on_hover_text("Link works only in already opened viewers and not in the browser's address bar.");
+            selectable_value_with_min_width(ui, MIN_TOGGLE_WIDTH, create_web_viewer_url, true, "Web viewer")
+                .on_hover_text("Link works in the browser's address bar, opening a new viewer. You can still use this link in the native viewer as well.");
+        });
+    }));
+
```
gavrelina, line 259: Can we add some space between the options? the borders currently overlap each other
``` ui.add_space(8.0);```

File: crates/viewer/re_viewer/src/ui/share_modal.rs [hunk 3]
```diff
@@ -0,0 +1,436 @@
...
+    if let Some(url_time_range) = url.time_range_mut() {
+        time_range_ui(ui, url_time_range, ctx.rec_cfg);
+    }
+    if let Some(fragments) = url.fragments_mut() {
```
gavrelina, line 264: and one for before Time cursor option! 
``` ui.add_space(8.0);```

File: crates/viewer/re_viewer/src/ui/share_modal.rs [hunk 4]
```diff
@@ -0,0 +1,436 @@
...
+                ui.list_item_scope("share_dialog_url_settings", |ui| {
+                    url_settings_ui(ctx, ui, url, &mut self.create_web_viewer_url);
+                });
+            },
+        );
+    }
+}
+
+// TODO(andreas): why is this not a thing, should we make it one?
+// TODO(andreas): What we _actually_ want is a group of toggles that are all vertically aligned and take up the entire width.
```
gavrelina, line 206: not sure if this is the right place but the width of the last one is a bit sticking out, almost as if it shall be truncated a bit earlier

crates/viewer/re_viewer/src/ui/share_modal.rs
```diff
@@ -0,0 +1,436 @@
...
+                MIN_TOGGLE_WIDTH,
+                &mut any_time,
+                false,
+                "At the start",
+            );
+            ui.add_enabled_ui(current_time_cursor.is_some(), |ui| {
+                let mut label = egui::Atoms::new(egui::Atom::from("Current").atom_grow(true));
+                if let Some((_, time_cell)) = current_time_cursor {
+                    label.push_right({
+                        let time = time_cell.format_compact(timestamp_format);
```
gavrelina: question: we display this format in the dialog as `+##.##### ms` but in the timeline of that same time we do not display the relative time and instead we show YYYY-MM-DD HH:MM:SS.blablabla. 

Looks like users will have no idea whether the time displayed in the dialog is the same as in the timeline, because we do not show it in the same format

Wumpf: depends on the type of timeline: we explicitely distinguish date based & time offset based ones

Wumpf: (that's somewhat recent)

Wumpf: so it should be the same actually. lemme double check

crates/viewer/re_ui/src/modal.rs
```diff
@@ -60,6 +60,11 @@ impl ModalHandler {
             None
         }
     }
+
+    /// Whet
```
IsseW: Unfinished comment?

crates/viewer/re_viewer/src/app.rs
```diff
@@ -1487,6 +1487,21 @@ impl App {
                 re_ui::apply_style_and_install_loaders(egui_ctx);
             }
 
+            UICommand::Share => {
+                let selection = self.state.selection_state.selected_items().clone();
```
IsseW: Is this cloned because of borrow stuff? I can't really tell what it'd be

crates/viewer/re_viewer/src/app.rs
```diff
@@ -1499,8 +1514,8 @@ impl App {
 
                 let rec_cfg = storage_context
                     .hub
-                    .active_recording()
-                    .and_then(|db| self.state.recording_config(db.store_id()));
+                    .active_store_id()
```
IsseW, line 1517: oh is this a better way to get the id? Or just more general

crates/viewer/re_viewer/src/ui/share_modal.rs
Wumpf: caught me, I feared that this could happen but didn't notice it and concluded good enough :/

crates/viewer/re_viewer/src/ui/share_modal.rs
```diff
@@ -0,0 +1,436 @@
...
+    modal::{ModalHandler, ModalWrapper},
+};
+use re_uri::Fragment;
+use re_viewer_context::{
+    DisplayMode, ItemCollection, RecordingConfig, StoreHub, UrlContext, ViewerContext,
+};
+
+use crate::open_url::ViewerOpenUrl;
+
+const COPIED_FEEDBACK_DURATION: Duration = Duration::from_millis(500);
```
IsseW: Not sure what the standard time to show messages like this is for, but to me it feels a bit quick.

I think how these usually work is that they say "copied link" until the cursor has left the button. Don't think this is something that has to be changed before merge though.

crates/viewer/re_ui/src/modal.rs
Wumpf: `Whet`?! xD
No idea what happened here, thanks for catching it

Wumpf: I promise, I'm a professional :|

crates/viewer/re_viewer/src/app.rs
Wumpf: it's just slightly shorter, `active_store_id` is essentially the exact same thing we had here before :)

crates/viewer/re_viewer/src/app.rs
Wumpf: yes. will add comment

Wumpf: correction: this was the case in an iteration of mine, but now is just wrong

crates/viewer/re_viewer/src/ui/share_modal.rs
Wumpf: I like that idea, lemme give it a stab

Wumpf: works like a charm and is simpler code

crates/viewer/re_viewer/src/ui/share_modal.rs
Wumpf: they're not, just as you observed. Looking into it!

Issue #10906, opened 2025-08-14
Improve Arrow rendering in tooltips


grtlr:
**Describe the bug**

The following PR adds very nice rendering of Arrow components:

* #10777 

Now that we might encounter deeply nested Arrow arrays our heuristics for when to output items during formatting (currently the threshold is set to 10 iirc) can be off.

**To Reproduce**
1. Open `arrow_example.rrd`
2. Hover over the `large_nested_list`
3. Screenshot below

> [!IMPORTANT]
> Happens when hovering over empty spaces in the Arrow values.

**Expected behavior**

* Use UI with syntax highlighting (cc @gavrelina 🙏)
* Keep track of how much output was already produced when visiting the nested fields and exit based on the length of encountered symbols instead of a hardcoded value of elements.

**Screenshots**

<img width="841" height="282" alt="Image" src="https://github.com/user-attachments/assets/4811d7be-743b-44f4-b1d1-78977cd87e62" />

lucasmerlin:
> Keep track of how much output was already produced when visiting the nested fields and exit based on the length of encountered symbols instead of a hardcoded value of elements.

I don't think we should do this. Say field a of some imaginary struct holds an array with 100s of values and field b and c hold much more valuable information. If we just limit based on the amount of output produced, the user will never see b and c. 
I think instead we should tune the limit of array / struct / map fields so it makes sense for most data. Maybe by decreasing the limit as data gets more nested.

grtlr:
That makes sense. The only thing I'm worried about is that we'd still have no guarantees over the size of the tooltip. I like the approach in your PR, let's see if this comes up again.

Off-topic: An (theoretical) ideal solution would probably mean using dynamic programming to look at the contents of a struct and find good cut-off points within the entire struct according to some heuristic.

PR #11139, opened 2025-09-08
Add arrow `DisplayOptions` and make limits configurable

lucasmerlin:
### Related

* related to https://github.com/rerun-io/rerun/issues/10906

### What

Smarter limits on how much data should be shown when formatting arrow data. With my test data this leads to 1-3 rows of data shown in tooltips, which I feel is reasonable.

Formatting in tooltips will need a egui release with 
- https://github.com/emilk/egui/pull/7514

### DIFF
--- a/crates/viewer/re_arrow_ui/tests/arrow_ui_test.rs
+++ b/crates/viewer/re_arrow_ui/tests/arrow_ui_test.rs
@@ -109,6 +109,6 @@ fn inline_formatting() {
         let highlighted =
             re_arrow_ui::arrow_syntax_highlighted(&data).expect("Failed to format data");
 
-        insta::assert_snapshot!(name, highlighted.text())
+        insta::assert_snapshot!(name, highlighted.text());
     }
 }


crates/viewer/re_arrow_ui/src/show_index.rs
```diff
@@ -28,24 +28,89 @@ use re_ui::{UiExt as _, UiLayout};
 use crate::arrow_node::ArrowNode;
 use crate::list_item_ranges::list_item_ranges;
 
-/// The maximum number of items when formatting an array to string.
+/// Arrow display options.
 ///
-/// If an array has more items, it will be truncated with `…`.
-pub const MAX_ARROW_LIST_ITEMS: usize = 10;
+/// Max item limits will not affect the `list_item` based ui, that will always show all items.
```
grtlr:
```suggestion
/// Max item limits will not affect the `list_item`-based ui, that will always show all items.
```

grtlr (commented):
Overall it looks good to me.

I really think we should have test cases in place if we ever want to update the parameters, though.

Since we already have it, I think it's time to check in (and potentially clean up) the code that produces the Arrow arrays from `arrow_example.rrd`. `re_arrow_util` might be a good place to put this.

What do you think?

lucasmerlin:
> add the auto-[] removal as an option

Definitely!

> expose some/all of theses options as (advanced) settings

Yes, I'd do that once the need arises

PR #11140, opened 2025-09-08
Fix wrong datatype displayed for non-list column in the table view

abey79:
### What

This PR does the following:
- fixes a bug where non-list columns had their datatype displayed as list
- properly display a table header with the display record batches are empty (e.g. when filtering a table down to nothing)
- attempt at splitting the supermassive `show()` method into two still quite massive functions

The root cause for the datatype bug is that by using sorbet batch, the column were auto-listified. This PR removes this step for `BatchType::Dataframe`.

image

image


crates/viewer/re_dataframe_ui/src/display_record_batch.rs
```diff
@@ -54,21 +51,19 @@ enum ComponentData {
         dict: ArrowInt32DictionaryArray,
         values: ArrowListArray,
     },
+    Scalar(ArrowArrayRef),
```
abey79, line 54: Do not worry too much about this change and anything with `ComponentData`. This thing will be removed in favour of the future `ArrayCell` thing, of which `ComponentData` was a precursor in hiding.

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -61,6 +61,19 @@ impl From<&TableBlueprint> for DataFusionQueryData {
     }
 }
 
+/// Result of the async datafusion query process.
+#[derive(Debug, Clone)]
+pub struct DataFusionQueryResult {
+    /// The resulting [`SorbetBatch`]s.
```
lucasmerlin:
```suggestion
    /// The resulting [`RecordBatch`]s.
```

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -61,6 +61,19 @@ impl From<&TableBlueprint> for DataFusionQueryData {
...
+    pub record_batches: Vec<RecordBatch>,
+
+    /// The schema of the record batches (useful if none is returned).
+    pub schema: SchemaRef,
+
+    /// The migrated schema of the record batches.
+    pub sorbet_schema: re_sorbet::SorbetSchema,
```
emilk: Migrating just the schema seems like a bug waiting to happen. What if we want to join columns in the migrator, for instance?

is the problem is the call to `make_all_data_columns_list_arrays`? If so, just comment it out (I don't think anyone is actually relying on it), or only call it for `BatchType::Chunk`

abey79: Well there is a bit of a philosophical questions here, I believe.

When `SorbetBatch` wears its "i'm a step towards deserialising a chunk" hat, then yes, it will likely end up making all sorts of columns splitting/merging/conversion.

When `SorbetBatch` wears its "best effort interpretation of sorbet metadata from user tables" hat, then it should actually have more respect for the original data imo.

I'll investigate reverting back to `SorbetBatch` but I do believe that this may also lead to bug where we display unintended transformation instead of the actual raw data (such as the one that this PR fixes).

abey79: @emilk my last commit does what you suggest: stop listifying on migrating `BatchType::Dataframe`, and revert to using `SorbetBatch`. It appears to be working fine for the dfn table widget. However:

1) I have no idea if there might be side effects to removing the listification. Cursory testing didn't show any, but I'm not sure where to look for—let's see what CI has to say.

2) My reservations above (which aren't exactly [new](https://github.com/rerun-io/rerun/issues/10421#issuecomment-3028163414)).

Please have a second look, and the last word ;)

abey79:
@rerun-bot full-check

abey79:
@rerun-bot full-check

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -61,6 +61,19 @@ impl From<&TableBlueprint> for DataFusionQueryData {
...
 }
 
+/// Result of the async datafusion query process.
+#[derive(Debug, Clone)]
+pub struct DataFusionQueryResult {
+    /// The record batches to display.
+    pub sorbet_batches: Vec<SorbetBatch>,
+
+    /// The schema of the record batches (useful if none is returned).
+    pub schema: SchemaRef,
```
emilk: maybe call this `original_schema` or something? Why do we still need it after migration? Shouldn't we keep the original record batches as well then?

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -61,6 +61,19 @@ impl From<&TableBlueprint> for DataFusionQueryData {
     }
 }
 
+/// Result of the async datafusion query process.
+#[derive(Debug, Clone)]
+pub struct DataFusionQueryResult {
+    /// The record batches to display.
+    pub sorbet_batches: Vec<SorbetBatch>,
+
+    /// The schema of the record batches (useful if none is returned).
```
emilk:
```suggestion
    /// The schema of the record batches (useful if the list of batches is empty).
```

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -629,7 +639,7 @@ impl egui_table::TableDelegate for DataFusionTableDelegate<'_> {
             let id = self.table_config.visible_column_ids().nth(column_index);
 
             if let Some((index, column)) = self.columns.index_and_column_from_id(id) {
-                let column_field = &self.fields[index];
+                let column_field = &self.query_result.schema.fields[index];
```
emilk: Why is this using the unmigrated schema and not the migrated one?

```suggestion
                let column_field = &self.query_result.sorbet_schema.fields[index];
```

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
abey79: We need the original schema because it contains the original column names, which I need to know about to later tell dfn to sort/filter according to a particular column.

BTW, this will all crumble the day we do column merging/etc in `SorbetBatch` migration, because how can I tell dfn to sort by column X if column X doesn't even exist in the original data?

One possibility would be to migrate the dataframe before even registering it to dfn. This would also solve the need to keep track of the original schema. This will become a world of pain with larger-than-memory tables though.

So, again, what we really need is a version of the sorbet migration that is "best effort interpretation of user-provided metadata", which is distinct from the existing "coerce all data into the current rerun format".

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
abey79: Because here I need the column that datafusion knows about, not the one we migrated (and display).

crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
emilk: > One possibility would be to migrate the dataframe before even registering it to dfn

That seems natural to me. In general, our approach is to migrate _early_.

If the problem is that the migration throws away the column names, then we can fix that by adding them back in to `SorbetBatchSchema`.

> what we really need is a version of the sorbet migration that is "best effort interpretation of user-provided metadata", which is distinct from the existing "coerce all data into the current rerun format"

I still don't get it. If the goal is to show the original record batches as the user sent them, then why are we doing _any_ migration?

emilk: Oh I think get it, its to read metadata (component name etc), which may need migration because of renames of those meta-keys 🤔

abey79: Yes! And to have the convenience of having repopulated `ColumnDescriptor` stuff. (In the future, we will likely introduce more rerun-type-specific formatting.)

abey79: I renamed `schema` to `original_schema` 👍🏻

emilk: feel free to merge this PR btw, I just wanted to understand the problem. Thanks for explaining it to me

abey79: Thanks, will do when ✅

PR #11142, opened 2025-09-08
Add support for filtering numeric (int/float) columns in tables

abey79:
### Related

* part of [RR-2260](https://linear.app/rerun/issue/RR-2260/complete-mvp-table-filtering-for-025)

### What

This PR adds support for filtering on numeric columns (int/float), including list array columns. In this case, the `any` semantics is used.

This PR also introduces a better format for the test snapshots (less verbose, includes unfiltered data for reference).


crates/viewer/re_dataframe_ui/src/filters.rs
```diff
@@ -126,6 +228,222 @@ impl FilterOperation {
...
+                    .iter()
+                    .map(|x| x.map(|v| operator.apply(v, *value as _)))
+                    .collect();
+
+                Ok(result)
+            }};
+        }
+
+        macro_rules! float_case {
+            ($conv_fun:ident, $op:expr) => {{
```
timsaucer, line 321: It looks like these can be condensed into one macro_rules with an extra ident passed in, `IntCompares` or `FloatCompares` but I didn't try it myself. Not a blocker.

abey79:
> someone else should review the UI portions

I'm not too worried about these. We already paired with Lucas, and they'll get rewritten 3x anyways :)

crates/viewer/re_dataframe_ui/src/filters.rs
abey79: Makes sense. I'll do that in a subsequent PR.

PR #11148, opened 2025-09-09
Remove custom number formatting on copy

lucasmerlin:
### Related
* closes RR-2238

### What

Removes the custom number formatting (thin spaces and custom minus sign) we apply in re_format when copying text.

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -336,11 +336,11 @@ impl App {
                     if !o.copied_text.is_empty() {
                         o.copied_text = re_format::remove_number_formatting(&o.copied_text);
                     }
-                    o.commands.iter_mut().for_each(|event| {
-                        if let egui::output::OutputCommand::CopyText(text) = event {
+                    for command in &mut o.commands {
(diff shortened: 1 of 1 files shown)


crates/viewer/re_viewer/src/app.rs
```diff
@@ -328,6 +328,23 @@ impl App {
...
+                ctx.output_mut(|o| {
+                    #[expect(deprecated)]
+                    if !o.copied_text.is_empty() {
+                        o.copied_text = re_format::remove_number_formatting(&o.copied_text);
+                    }
+                    o.commands.iter_mut().for_each(|event| {
+                        if let egui::output::OutputCommand::CopyText(text) = event {
+                            *text = re_format::remove_number_formatting(text);
+                        }
+                    });
```
emilk: style: what's wrong with a good old fashion for loop? 🤷 

```suggestion
                    for command in &mut o.commands {
                        if let egui::output::OutputCommand::CopyText(text) = command {
                            *text = re_format::remove_number_formatting(text);
                        }
                    }
```

emilk (approved):

PR #11149, opened 2025-09-09
Replace member list in changelog to `rerun-io` fetched list

Wumpf:
Tested locally, works like a charm (90% machine generated code)

Plus some unrelated stuff that I've been to lazy to create separate PRs for
* typo in link sharing description
* don't bother with the web viewer checkbox on the pr comments, nobody presses it and it gets overriden all the time

### DIFF
--- a/crates/viewer/re_viewer/src/open_url.rs
+++ b/crates/viewer/re_viewer/src/open_url.rs
@@ -558,7 +558,7 @@ impl ViewerOpenUrl {
                     url_parameters.first().open_description()
                 } else {
                     ViewerOpenUrlDescription {
-                        category: "Open several URLs",
+                        category: "Several URLs",
                         target_short: Some(format!("{} URLs", url_parameters.len())),
                     }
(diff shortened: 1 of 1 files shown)


crates/viewer/re_viewer/src/open_url.rs
```diff
@@ -558,7 +558,7 @@ impl ViewerOpenUrl {
                     url_parameters.first().open_description()
                 } else {
                     ViewerOpenUrlDescription {
-                        category: "Open several URLs",
+                        category: "Several URLs",
```
IsseW, line 561: I think this saying something like "Open N URLs" would be cleaner, but a bit out of scope of this PR maybe 😅

Wumpf: eh yeah maybe I shouldn't have put this into this PR, very very very offtopic. Guilty!
It actually works out fine because we always also print the extra info below alongside :)

PR #11152, opened 2025-09-09
feat: Improve error handling by replacing unwrap with expect  in `re_selection_panel`

Weijun-H:
### Related

Part of #6330


crates/viewer/re_selection_panel/src/view_entity_picker.rs
```diff
@@ -194,8 +194,9 @@ fn add_entities_line_ui(
     let entity_path = &entity_data.entity_path;
     let name = &entity_data.label;
 
-    #[allow(clippy::unwrap_used)]
-    let add_info = entities_add_info.get(entity_path).unwrap();
+    let add_info = entities_add_info
+        .get(entity_path)
+        .unwrap_or_else(|| panic!("HashMap should contain entry for {entity_path}"));
```
Wumpf: that's the exact same thing as an `expect`. also the error message isn't more helpful that the unwrap. why do we expect it to contain the element already? is there a way to get rid of the potential crash entirely by adjusting the method?

Weijun-H: For the missing element, should it return immediately, or should it be treated as the default `EntityAddInfo`? 🤔  af2a2e40994a03ec4765c60522258e2daf4de64f

Wumpf: checked this branch out locally to check on something, so I took the freedom to go with return immediately + debug assert

Wumpf (approved):

PR #11153, opened 2025-09-09
Update web address with focus and time fragment

### LINKED ISSUE #10866: Improve dataplatform navigation bar support further

IsseW:
### Related

- Depends on #11135
- Closes #10866.

### What

Makes the web url include, and update with selection and the current selected time.

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -521,6 +521,21 @@ impl App {
         }
     }
 
+    /// Updates the web address on web. Noop on native.
+    fn update_web_address_bar(&self, store_hub: &StoreHub) {
+        let rec_cfg = store_hub
+            .active_recording()
+            .and_then(|db| self.state.recording_config(db.store_id()));
+        let time_ctrl = rec_cfg.as_ref().map(|cfg| cfg.time_ctrl.read());
+        update_web_address_bar(
+            self.startup_options.web_history_enabled(),
+            store_hub,
+            self.state.navigation.peek(),
+            self.state.selection_state.new_selected_items().first_item(),
+            time_ctrl.as_deref(),
+        );
+    }
+
     #[allow(clippy::unused_self)]
     fn run_system_command(
         &mut self,
@@ -528,29 +543,16 @@ impl App {
         store_hub: &mut StoreHub,
         egui_ctx: &egui::Context,
     ) {
-        let update_address_bar = |this: &Self, store_hub: &mut StoreHub| {
-            let rec_cfg = store_hub
-                .active_recording()
-                .and_then(|db| this.state.recording_config(db.store_id()));
-            let time_ctrl = rec_cfg.as_ref().map(|cfg| cfg.time_ctrl.read());
-            update_web_address_bar(
-                this.startup_options.web_history_enabled(),
-                store_hub,
-                this.state.navigation.peek(),
-                this.state.selection_state.new_selected_items().first_item(),
-                time_ctrl.as_deref(),
-            );
-        };
         match cmd {
             SystemCommand::ActivateApp(app_id) => {
                 self.state.navigation.replace(DisplayMode::LocalRecordings);
                 store_hub.set_active_app(app_id);
-                update_address_bar(self, store_hub);
+                self.update_web_address_bar(store_hub);
             }
 
             SystemCommand::CloseApp(app_id) => {
                 store_hub.close_app(&app_id);
-                update_address_bar(self, store_hub);
+                self.update_web_address_bar(store_hub);
             }
 
             SystemCommand::ActivateRecordingOrTable(entry) => {
@@ -565,7 +567,7 @@ impl App {
                             .replace(DisplayMode::LocalTable(table_id.clone()));
                     }
                 }
-                update_address_bar(self, store_hub);
+                self.update_web_address_bar(store_hub);
             }
 
             SystemCommand::CloseRecordingOrTable(entry) => {
@@ -606,7 +608,7 @@ impl App {
 
                 store_hub.remove(&entry);
 
-                update_address_bar(self, store_hub);
+                self.update_web_address_bar(store_hub);
             }
 
             SystemCommand::CloseAllEntries => {
@@ -646,7 +648,7 @@ impl App {
                 // _which_ recording is about to be selected.
                 // I.e. if we update navigation bar here, this would become order dependent.
                 if display_mode != DisplayMode::LocalRecordings {
-                    update_address_bar(self, store_hub);
+                    self.update_web_address_bar(store_hub);
                 }
 
                 self.state.navigation.replace(display_mode);
@@ -788,7 +790,7 @@ impl App {
                 }
 
                 self.state.selection_state.set_selection(items);
-                update_address_bar(self, store_hub);
+                self.update_web_address_bar(store_hub);
                 egui_ctx.request_repaint(); // Make sure we actually see the new selection.
             }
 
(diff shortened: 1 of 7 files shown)


Wumpf:
(I've put do-not-merge so we don't merge into the parent branch while its pr is still open)

Wumpf:
It seems to break navigation: I end up in an endless loop when trying to navigate back

https://github.com/user-attachments/assets/a173b5df-f00d-45b6-9227-f5265b5bbff1

Haven't checked the code yet, but something is inserting new nav events.

Time selection on the URL doesn't update on scrubbing yet - if I click around on a paused timeline, it doesn't update the URL either. Happy to take that as a follow-up PR, but in that case let's not close https://github.com/rerun-io/rerun/issues/10866 just yet :)
I'd also not clear out the timepoint from the URL while playing and just keep the old state, but I realize that this might be annoying to implement, so no strong opinion on this.

What's also quite annoying when playing around with it, is that we literally trigger `focus` events which means that going back actually doesn't get the ui into the expected state but rather zooms in on stuff. That's not your fault since this is simply what having the path on the ui means right now! But seeing this in action I think we should actually not trigger a focus at all and instead just select. Yes, this breaks existing behavior, but it's not like this was all that usable before :)

IsseW:
> What's also quite annoying when playing around with it, is that we literally trigger `focus` events which means that going back actually doesn't get the ui into the expected state but rather zooms in on stuff.

Yeah, going through history it's definitely annoying. But I wonder if you send someone a link of a big scene, focusing is probably wanted in that case?

But in the current state removing the focus event is probably a good call. Do we want to rename that fragment part to `selection` then?

Wumpf:
> Do we want to rename that fragment part to selection then?

in favor!
but if we do that let's add a unit test that checks that we can deal with invalid fragments. I don't think there's old links in circulation _yet_ but would be silly to break over fragment parsing 🤔

IsseW:
> Time selection on the URL doesn't update on scrubbing yet - if I click around on a paused timeline, it doesn't update the URL either. Happy to take that as a follow-up PR, but in that case let's not close https://github.com/rerun-io/rerun/issues/10866 just yet

Looking into this it does seem kind of annoying so thinking taking it in a follow up is the way to go.

Also thinking if we could update call the update web address function each tick. It has a check for if the url actually changes before trying to do any history stuff anyway. It does make some of the refactoring I've done not needed for this though 🙃

crates/utils/re_uri/src/redap_uri.rs
```diff
@@ -252,6 +252,34 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_dataset_data_url_with_broken_fragment() {
```
Wumpf, line 256: nice!!

crates/viewer/re_viewer/src/app.rs
```diff
@@ -539,6 +538,21 @@ impl App {
         }
     }
 
+    /// Updates the web address on web. Noop on native.
+    fn update_web_address_bar(&self, store_hub: &StoreHub) {
+        let rec_cfg = store_hub
+            .active_recording()
+            .and_then(|db| self.state.recording_config(db.store_id()));
+        let time_ctrl = rec_cfg.as_ref().map(|cfg| cfg.time_ctrl.read());
+        update_web_address_bar(
```
Wumpf: nit: this confused me on review, looked recursive but isn't

crates/viewer/re_viewer/src/open_url.rs
```diff
@@ -198,12 +198,12 @@ impl ViewerOpenUrl {
     /// To produce a sharable url, from this result, call [`Self::sharable_url`].
     ///
     /// Returns Err(reason) if the current state can't be shared with a url.
-    // TODO(#10866): Should have anchors for selection etc. when supported. Need to figure out how this works together with the "share editor".
     // Does this method merely provide the starting point?
```
Wumpf: dead comment fragment

```suggestion
```

crates/viewer/re_viewer/src/open_url.rs
```diff
@@ -383,9 +382,9 @@ impl ViewerOpenUrl {
         } else if urls.len() == 1 {
             Ok(urls.split_off_first().0)
         } else {
-            Err(anyhow::anyhow!(
+            return Err(anyhow::anyhow!(
```
Wumpf: nit: unneeded `return`, leftover from a discarded approach?
surprised clippy isn't picking up on it

crates/viewer/re_viewer/src/app.rs
```diff
@@ -2859,6 +2827,21 @@ impl eframe::App for App {
...
+        // This is a kind of hacky way to fix that: If there are currently any
+        // inputs, don't update the web address bar. This works for most cases
+        // because you need to hold down pointer to aggressively scrub, need to
+        // hold down key inputs to quickly step through the timeline.
+        if !egui_ctx.is_using_pointer()
+            && egui_ctx.input(|input| !input.any_touches() && input.keys_down.is_empty())
+        {
+            // Update web address after commands have been applied.
+            self.update_web_address_bar(&store_hub);
+        }
```
Wumpf: despite the input based hack this is just so much better than the guessing game of when to update the address bar. nice!

crates/viewer/re_viewer/src/app.rs
```diff
@@ -2859,6 +2827,21 @@ impl eframe::App for App {
         }
         self.run_pending_system_commands(&mut store_hub, egui_ctx);
 
+        // We don't want to spam the history API with changes, because
+        // otherwise it will start complaining about it being an insecure
+        // operation.
```
Wumpf:
```suggestion
        // We don't want to spam the web history API with changes, because
        // otherwise it will start complaining about it being an insecure
        // operation.
```

crates/viewer/re_viewer/src/app.rs
```diff
@@ -3233,17 +3217,32 @@ fn update_web_address_bar(
...
+                .and_then(|when| {
+                    Some((
+                        *when.timeline().name(),
+                        re_log_types::TimeCell {
+                            typ: when.timeline().typ(),
+                            value: when.time_int()?.into(),
+                        },
+                    ))
+                }),
+        },
```
Wumpf: this approach is more than good enough for this iteration, but I'd hope we could move the fragment determination into `from_display_mode` (which ehrm isn't quite just from display mode anymore then).
That way we always get the same starting URL, both for the share dialog & and the "copy direct link" action. Making not having the fragment an opt-out; yes that's a bit more opinionated 🤷‍♀️, curious what you think would be best here going forward :)

IsseW: Ohh yeah, I think that's a good idea to have the determination in a function there.

I put a comment on your draft PR https://github.com/rerun-io/rerun/pull/11137#discussion_r2339350426 because I saw that you did passing the fragment differently there.

I did like how you do it there, because then in for example context menus I can possibly do
```rust
if let Some(fragment) = url.fragments_mut() {
	// add the copy time selection url button only if we support that in this display mode
}
```

Although that can be done in other ways like having a function to check if fragments are supported.

crates/viewer/re_viewer/src/app.rs
IsseW: hmm yeah, can probably have the wasm compile check on the method and inline the function

Wumpf: or just rename one of them? 🤷 
maybe this is just me, you can leave it also be as is, up to you!

IsseW: Tried inlining it, and think it turned out much nicer. Could move the skip history check to the top of that function.

PR #11154, opened 2025-09-09
Add RSS to opentelemetry and `/metrics` endpoint

emilk:


crates/utils/re_perf_telemetry/src/memory_telemetry.rs
```diff
@@ -0,0 +1,17 @@
...
+    meter
+        .i64_observable_gauge("RSS")
+        .with_description("Resident Set Size")
+        .with_unit("B")
+        .with_callback(|observer| {
+            let bytes_used = memory_stats::memory_stats().map(|usage| usage.physical_mem as i64);
+            if let Some(bytes_used) = bytes_used {
+                observer.observe(bytes_used, &[]);
+            } else {
+                observer.observe(-666, &[]);
```
alvaroaleman: Maybe -1 to indicate error/no data rather than -666? The latter looks like a miscalculation from the outside IMHO

thz: Can we not "observe" instead of observing bogus data?

thz: Having any number in case of error, will make any aggregation over the fleet useless.

crates/utils/re_perf_telemetry/src/memory_telemetry.rs
```diff
@@ -0,0 +1,15 @@
+pub fn install_memory_use_meters() {
+    let meter = opentelemetry::global::meter("memory-use");
+
+    meter
+        .i64_observable_gauge("RSS")
```
emilk:
```suggestion
        .i64_observable_gauge("memory_resident_set_size_bytes")
```

thz (approved):

PR #11157, opened 2025-09-09
Use partition id in the recording panel

abey79:
### What

With this PR, the recording panel (and everywhere we use `entity_db_button_ui`), we now use:

- the recording's name (from the properties) if any
- the partition id if any (obviously remote stuff only)
- the recording starting time if any
- `<unknown>` otherwise

Previously, we didn't have the partition id step.

Here is what i currently looks like with droid dataset:

image

### DIFF
--- a/crates/viewer/re_data_ui/src/item_ui.rs
+++ b/crates/viewer/re_data_ui/src/item_ui.rs
@@ -2,6 +2,7 @@
 //!
 //! TODO(andreas): This is not a `data_ui`, can this go somewhere else, shouldn't be in `re_data_ui`.
 
+use re_entity_db::entity_db::EntityDbClass;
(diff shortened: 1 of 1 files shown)


crates/viewer/re_data_ui/src/item_ui.rs
```diff
@@ -696,6 +697,8 @@ pub fn entity_db_button_ui(
         entity_db.recording_info_property::<Name>(&RecordingInfo::descriptor_name())
     {
         Some(recording_name.to_string())
+    } else if let EntityDbClass::DatasetPartition(url) = entity_db.store_class() {
+        Some(url.partition_id.clone())
```
emilk, line 706: It would be nice with a comment explaining _why_ we do what we do

PR #11158, opened 2025-09-09
Try to resolve snippet errors

ntjohnson1:
My [PR](https://github.com/rerun-io/rerun/pull/11045) broke main. Remove deprecated method usages in snippets.

I wasn't able to fully repro locally because I was running into some other errors that might be specific to Mac. If CI errors on my first run of validation I'll try to root cause that.


ntjohnson1:
@rerun-bot full-check

Wumpf (approved):

ntjohnson1:
@rerun-bot full-check

ntjohnson1:
@rerun-bot full-check

ntjohnson1:
@rerun-bot full-check

docs/snippets/all/tutorials/dynamic_archetype.cpp
```diff
@@ -0,0 +1,66 @@
+// Log arbitrary archetype data.
```
Wumpf, line 1: yay more examples. but it's a bit of an odd one, isn't this just a straight up copy of the anyvalues one? Could have opted out of C++ instead in snippets.toml
Well I guess it makes discoverability of how to do the equivalent better

ntjohnson1: It's a copy but with a manual workaround to with_builtin_archetype not existing. The error message and resolution with snippets.toml weren't super obvious last night.

PR #11161, opened 2025-09-10
Fix sharing selection as a fragment

IsseW:
### Related

 - Closes RR-2222
 - Based on #11153 because I made some changes there that would conflict with this otherwise.

### What

Fixes selection sharing, which before this was only applied too early to actually affect the selection.

This adds the variant `SetFragment` to `UiCommand` in `crates/store/re_redap_client/src/grpc.rs`, which I'm pretty sure is a breaking change (according to semver).

Also adds a system command `SetFragment`, which creates other system commands when invoked. Which might be the first system command which does that? The system command loop looks like it's made to handle that though.


crates/viewer/re_test_context/src/lib.rs
```diff
@@ -540,12 +540,54 @@ impl TestContext {
         result.expect("Function should have been called at least once")
     }
 
+    /// Applies a fragment.
+    ///
+    /// Does *not* switch the active recording.
+    fn go_to_dataset_data(&self, store_id: StoreId, fragment: re_uri::Fragment) {
```
IsseW, line 546: This is just a copied version of the one in `app.rs`, not sure if it's really worth creating some common function for those to share.

crates/utils/re_uri/src/fragment.rs
```diff
@@ -112,6 +112,14 @@ impl Fragment {
             }
         }
     }
+
+    /// True if this fragment doesn't contain any information.
+    pub fn is_empty(&self) -> bool {
+        // Keep this as a destruction so there is a compile error if a new field isn't handled here.
+        let Self { selection, when } = self;
```
Wumpf, line 119: 👍👍

crates/viewer/re_test_context/src/lib.rs
Wumpf: hmmmmm there's a lot of stuff in `TestContext` which at this point is just a better version of what we have in the older/more grown `app.rs`.
I think I would have left `SetFragment` for test context unimplemented, but I'm happy you didn't because that would cause surprises later for sure.

agree with your assessment

Wumpf: hmmmmm there's a lot of stuff in TestContext which at this point is just a better version of what we have in the older/more grown app.rs.
I think I would have left SetFragment for test context unimplemented, but I'm happy you didn't because that would cause surprises later for sure.

agree with your assessment

crates/viewer/re_global_context/src/command_sender.rs
```diff
@@ -90,6 +90,12 @@ pub enum SystemCommand {
     #[cfg(debug_assertions)]
     EnableInspectBlueprintTimeline(bool),
 
+    /// Load a [`re_uri::Fragment`].
+    SetFragment {
```
Wumpf: nit:
```suggestion
    /// Navigate to time/entities/anchors/etc. that are set in a [`re_uri::Fragment`].
    SetUrlFragment {
```
don't know about you, but when I hear `Fragment` without the context of our current workstream I'd be confused and first think about pixels 😄

crates/viewer/re_viewer/src/app.rs
```diff
@@ -468,7 +468,7 @@ impl App {
         if let SmartChannelSource::RedapGrpcStream { uri, .. } = rx.source() {
             self.command_sender
                 .send_system(SystemCommand::AddRedapServer(uri.origin.clone()));
-            self.go_to_dataset_data(uri);
+            self.go_to_dataset_data(uri.store_id(), uri.fragment.clone());
```
Wumpf: I thiiiiink this is redundant now? or at least I'm confused why it's needed.
Either way, comment above about adding servers doesn't apply to this

IsseW: yeah was able to remove this

PR #11162, opened 2025-09-10
re_perf_telemetry: Log memory stats as memory grows

emilk:
### Related
* Fixes RR-1965

### What
For Rerun Cloud.

Each time we cross another 1GiB boundary we log how much RAM is currently used, and the top 10 callstacks that are eating the most RAM (and how much they are eating). This should help us quickly diagnose OOM errors and find memory leaks.


crates/utils/re_memory/src/accounting_allocator.rs
```diff
@@ -82,6 +76,13 @@ struct GlobalStats {
     /// This is expensive, but sometimes useful!
     pub track_callstacks: AtomicBool,
 
+    /// Never capture callstacks for allocations smaller than this.
+    pub small_size: AtomicUsize,
+
+    /// Allocations smaller than are stochastically sampled.
```
IsseW:
```suggestion
    /// Allocations smaller than this are stochastically sampled.
```

crates/utils/re_memory/src/accounting_allocator.rs
```diff
@@ -95,38 +96,80 @@ struct GlobalStats {
...
+
+/// Controls which allocations to sample.
+///
+/// By "sample" we mean "capture a callstack".
+#[derive(Clone, Copy, Debug)]
+pub struct TrackingOptions {
+    /// Never capture callstacks for allocations smaller than this.
+    pub small_size: usize,
+
+    /// Allocations smaller than are stochastically sampled.
```
IsseW:
```suggestion
    /// Allocations smaller than this are stochastically sampled.
```

crates/utils/re_perf_telemetry/src/memory_telemetry.rs
```diff
@@ -12,4 +12,112 @@ pub fn install_memory_use_meters() {
...
+    // ---------------------------------------------------
+
+    // Configure the acounting allocator to only capture the big stuff.
+    // That way we get low overhead.
+    re_memory::accounting_allocator::set_tracking_options(
+        re_memory::accounting_allocator::TrackingOptions {
+            // Never capture callstacks for allocations smaller than this.
+            small_size: 1024,
+
+            // Allocations smaller than are stochastically sampled.
```
IsseW:
```suggestion
            // Allocations smaller than this are stochastically sampled.
```

crates/utils/re_perf_telemetry/src/memory_telemetry.rs
```diff
@@ -12,4 +12,112 @@ pub fn install_memory_use_meters() {
             }
         })
         .build();
+
+    // ---------------------------------------------------
+
+    // Configure the acounting allocator to only capture the big stuff.
```
IsseW:
```suggestion
    // Configure the accounting allocator to only capture the big stuff.
```

crates/utils/re_perf_telemetry/src/memory_telemetry.rs
```diff
@@ -12,4 +12,112 @@ pub fn install_memory_use_meters() {
...
+
+            // Allocations smaller than are stochastically sampled.
+            // Allocations larger than this are fully sampled.
+            medium_size: 1024 * 1024,
+        },
+    );
+    // Turn on the accounting allocator:
+    re_memory::accounting_allocator::set_tracking_callstacks(true);
+
+    // Perioidically check memory usage:
```
IsseW:
```suggestion
    // Periodically check memory usage:
```

crates/utils/re_perf_telemetry/src/memory_telemetry.rs
```diff
@@ -12,4 +12,112 @@ pub fn install_memory_use_meters() {
...
+async fn memory_monitor_task() {
+    const ONE_GIG: u64 = 1024 * 1024 * 1024;
+
+    // How often we check RAM use.
+    let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(30));
+
+    let total_ram_in_bytes = re_memory::total_ram_in_bytes();
+
+    let mut warn_at_gb = if let Some(total_ram_in_bytes) = total_ram_in_bytes {
+        let total_ram_gb = total_ram_in_bytes / ONE_GIG;
```
IsseW, line 40: I guess it's not really a problem that this breaks when ram < 2 GiB 😄 Although looking at the code I don't see any huge reason `warn_at_gb` can't check with bytes.

IsseW (changes requested):
Looks good! Mostly noticed doc things

crates/utils/re_perf_telemetry/src/memory_telemetry.rs
emilk: I don't want to print out the top allocators too often, so I do it every time we go up a full gigabyte.

Yes, that means this is pretty useless when running on a machine with < 2GiB of ram, but we're not running our cloud instances on potatoes :)

PR #11169, opened 2025-09-11
Unify button handling for CustomContent

lucasmerlin:
### Related

* follow-up to #11164 

### What

Implement the ItemButtons for CustomContent. This still has some problems around how overflow is handled (see the changed test image). I attempted to fix this using egui::Sides but this doesn't seem to fix anything yet. Maybe we could try to change the remember_property_content_max_width to work for all the content types.


Wumpf:
outdated

PR #11171, opened 2025-09-11
Retry problematic CI jobs

emilk:
It's a sad "fix", but may keep `main` green a bit more often.

### DIFF
--- a/.github/workflows/reusable_test_wheels.yml
+++ b/.github/workflows/reusable_test_wheels.yml
@@ -181,6 +181,7 @@ jobs:
         with:
           max_attempts: 2
           retry_wait_seconds: 30
+          timeout_minutes: 60
           command: cd rerun_py/tests && pixi run -e wheel-test-min pytest -c ../pyproject.toml
 
       - name: Run e2e test
--- a/.github/workflows/reusable_checks_rust.yml
+++ b/.github/workflows/reusable_checks_rust.yml
@@ -195,6 +195,7 @@ jobs:
         with:
           max_attempts: 2
           retry_wait_seconds: 30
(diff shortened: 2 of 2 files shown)


.github/workflows/reusable_test_wheels.yml
```diff
@@ -176,7 +176,12 @@ jobs:
 
       - name: Run unit tests
         if: ${{ inputs.PLATFORM != 'linux-arm64' }}
-        run: cd rerun_py/tests && pixi run -e wheel-test-min pytest -c ../pyproject.toml
+        # We've seen some spurious network errors here… easier to retry than to fix :(
+        uses: nick-fields/retry@v3
+        with:
+          max_attempts: 2
+          retry_wait_seconds: 30
+          command: cd rerun_py/tests && pixi run -e wheel-test-min pytest -c ../pyproject.toml
```
ntjohnson1, line 185: Can you file an issue for this? From the CI failure I saw was transport errors for our datafusion test where we run the OSS server. My assumption is our port checking isn't working correctly in the cloud. These hosted server type tests always tend to be a little flaky but we could probably do a better check for available ports to make it more stable.

I just worry once we start turning on retries for flaky tests it disincentivizes making stable tests. Not blocking this PR but voicing concern for the future.

ntjohnson1: Ya I see a note about `Error during waiting for server to start: Server port 51234 not ready within 30s` using randomized ports every time we launch the server and returning the port number from the fixture should help.

emilk: Good idea:
* #11173

Issue #10807, opened 2025-08-06
Create example MCAP files to test against


grtlr:
For `.rrd`s we have several layers of testing:

* Unit tests
* Snippet tests
* Backwards compatibility tests (`tests/assets`)

We really need to start testing our MCAP ingestion as well. This entails:

- [ ] Testing of `protobuf` reflection (MCAP message to Chunk snapshot)
- [ ] Testing of any other reflection-based appoaches
- [ ] Testing of our semantic extraction

grtlr:
Although not perfect, this was closed by #11175.

We can create actionable follow up issues for more coverage and improvements as needed.

PR #11175, opened 2025-09-11
Adds basic snapshots tests for MCAP binaries

aedm:
### Related

- part of #10807

### What

Snapshot tests for the MCAP loader.
Adds a basic protobuf-based MCAP and a detailed ROS2 MCAP by @gijs.

### DIFF
--- a/crates/store/re_data_loader/tests/snapshots/test_mcap_loader__tests__assert_chunk_snapshot@simple_protobuf_vec3.mcap.snap
+++ b/crates/store/re_data_loader/tests/snapshots/test_mcap_loader__tests__assert_chunk_snapshot@simple_protobuf_vec3.mcap.snap
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:558ea889070b41ee5d9a5145f90ecc0030c8722b051bd6a374a442ff616d3fc9
-size 24083
+oid sha256:0d754b1992f72daa7623c49ba4ce21ad7fe392f9e153e0a9958ed719271dd1e5
+size 17642
--- a/crates/store/re_data_loader/tests/snapshots/test_mcap_loader__tests__assert_chunk_snapshot@supported_ros2_messages.mcap.snap
+++ b/crates/store/re_data_loader/tests/snapshots/test_mcap_loader__tests__assert_chunk_snapshot@supported_ros2_messages.mcap.snap
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:e549b0e8f4609d763f4eb1d4c4feb719aeb66e0a226de62f2e0adb1a95196bc0
-size 581928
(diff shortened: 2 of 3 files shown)


crates/store/re_data_loader/src/loader_mcap.rs
```diff
@@ -154,7 +154,7 @@ fn load_mcap_mmap(
     load_mcap(&mmap, settings, tx, selected_layers)
 }
 
-fn load_mcap(
+pub fn load_mcap(
```
aedm, line 184: Eeek. This is a snapshot test suite, therefore it's under `tests/` as customary. But Rust considers tests in that folder integration tests that can't access `pub(crate)` methods. So I had to make it `pub`. I'm not happy about it. :-/

crates/store/re_data_loader/tests/assets/rosbag2_2025_09_10-13_03_56_0.mcap
grtlr, line 1: Maybe we could give this a more descriptive name, like `supported_ros2_messages.mcap`?

crates/store/re_data_loader/tests/.gitattributes
```diff
@@ -0,0 +1,2 @@
+*.mcap filter=lfs diff=lfs merge=lfs -text
+*.snap filter=lfs diff=lfs merge=lfs -text
```
grtlr, line 2:
```suggestion
```

I don't think we use `lfs` to manage text-based snapshots. We do so for `.png` though.

aedm: Can't. :( The snap file of the rosbag is almost 600k, which makes CI unhappy (`pixi run check-large-files`).

crates/store/re_data_loader/tests/snapshots/test.snap.new
oxkitsune, line 1: I dont think this should be here 😁

.github/workflows/reusable_checks_rust.yml
```diff
@@ -202,6 +202,10 @@ jobs:
         if: ${{ inputs.CHANNEL == 'nightly' }}
         run: pixi run rs-check
 
+      - name: DEBUG ME PLEASE
```
oxkitsune: Not sure what this is about, but I dont think it should be here either

crates/store/re_data_loader/src/loader_mcap.rs
emilk: If you really hate it you can hide it behind a `#[cfg(feature = "testing")]` feature flag, and add `#![cfg(feature = "testing")]` to the test file. (we run all our tests with --all-features`)

crates/store/re_data_loader/tests/snapshots/test.snap.new
abey79: @aedm yep, it really should not be there, yet it was merged :)

PR #11179, opened 2025-09-11
Separate timelines for message timestamps with a custom epoch

oxkitsune:
### Related

* Fixes issue introduced in #11172 

### What

Currently, Rerun does not support mixing different timestamp types on the same timeline. 

In #11172 this was overlooked, causing MCAP files with mixed timestamp epochs to drop topics. This fixes that, by assigning messages with custom-epoch timestamps to separate timelines.

### TODO
* [x] `@rerun-bot full-check`


oxkitsune:
@rerun-bot full-check

PR #11180, opened 2025-09-11
Remove `web_viewer` symlink

emilk:
For historical reasons, we always built the web viewer to the root `./web_viewer` folder.
This presented a problem when first publishing `re_web_viewer_server`: we are not allowed to publish a crate that includes files outside its own folder. So as a "quick fix" we created `crates/viewer/re_web_viewer_server/web_viewer` as a sym-link to the root `./web_viewer` folder.

We now encountered an error when publishing crate, probably related to this: https://github.com/rerun-io/rerun/actions/runs/17644901646/job/50140485971#step:5:186

So, this PR removes the symlink and just moves the web viewer to always be built in `crates/viewer/re_web_viewer_server/web_viewer`. For instance, running `pixi run rerun-web` will now build directly into that folder.

### Related
* https://github.com/rerun-io/rerun/issues/11199


emilk:
@rerun-bot full-check

jprochazk:
Though regarding the build failure:

> 2 files in the working directory contain changes that were not yet committed into git:

Those files are gitignored. So I'm not sure why cargo is complaining that they aren't committed to git?

emilk:
Yup - and why is cargo complaining now, but not last release? Maybe because of a recent version update of cargo?

Issue #10762, opened 2025-07-28
Multiple blobs on same entity leads to ambiguous UI


grtlr:
**Describe the bug**

When logging multiple blobs to the same entity (yay tagged components) the "Save blob..." button is placed not within its archetype but all the way at the bottom.

**To Reproduce**

1. Log multiple blobs under different archetypes to the same entity.
2. Look at the selection panel

**Expected behavior**
The download button should be shown grouped within each corresponding archetype.

**Screenshots** *(note the two "data" blobs)*

<img src="https://uploads.linear.app/93919c5e-abe4-4cd5-b77d-7e3eafe3e62d/c00eaebf-4f37-4325-93ef-47fb121c7894/124c2af2-c0db-4d25-8e9b-d409e60a20a9?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzkzOTE5YzVlLWFiZTQtNGNkNS1iNzdkLTdlM2VhZmUzZTYyZC9jMDBlYWViZi00ZjM3LTQzMjUtOTNlZi00N2ZiMTIxYzc4OTQvMTI0YzJhZjItYzBkYi00ZDI1LThlOWItZDQwOWU2MGEyMGE5IiwiaWF0IjoxNzU3NDk5MjIxLCJleHAiOjMzMzI4MDU5MjIxfQ.X440AVUvD75hTZkrBwGxWPYVfvoHHjKnP0Mzppd7wZU " alt="Image" width="482" data-linear-height="367" />

Wumpf:
this might lead us to a slightly architectural issue that we've been delaying on: those additional `save blob` buttons are, just like image & video previews, a bit of a hack on top rather than following a generic path.
Ideally we figure a way out to register some kind of archetype ui so we can solved this more generically

grtlr:
@emilk somewhat related to what we discussed yesterday.

Wumpf:
worst case we could just not show the button at all instead of showing several.

PR #11181, opened 2025-09-11
Move copy / save buttons inline with the relevant component

lucasmerlin:
### Related

* closes https://github.com/rerun-io/rerun/issues/10762
* should be merged after https://github.com/rerun-io/rerun/pull/11164

### What

This moves the copy / save buttons inline next to the component. In order to do this, I've split the blob / image / video uis into structs with a `data_ui` and `inline_button` fns. The `ExtraDataUi` is a wrapper for convenience. 

In hindsight this could have probably also been done using freestanding fns.

This also fixes a couple inconsistencies between `Image` and `EncodedImage`. Both now show a histogram and both show the image format.

EncodedImage:
Screenshot 2025-09-11 at 16 14 08

Image:
Screenshot 2025-09-11 at 16 16 03

AssetVideo:
Screenshot 2025-09-11 at 16 14 46

### DIFF
--- a/crates/viewer/re_data_ui/src/image.rs
+++ b/crates/viewer/re_data_ui/src/image.rs
@@ -345,20 +345,18 @@ impl ImageUi {
         ctx: &'a ViewerContext<'_>,
         property_content: list_item::PropertyContent<'a>,
     ) -> list_item::PropertyContent<'a> {
-        property_content
-            .with_action_button(&icons::COPY, "Copy image", move || {
-                if let Some(rgba) = self.image.to_rgba8_image(self.data_range.into()) {
-                    let egui_image = egui::ColorImage::from_rgba_unmultiplied(
-                        [rgba.width() as _, rgba.height() as _],
-                        bytemuck::cast_slice(rgba.as_raw()),
-                    );
-                    ctx.egui_ctx().copy_image(egui_image);
-                    re_log::info!("Copied image to clipboard");
-                } else {
-                    re_log::error!("Invalid image");
-                }
-            })
-            .with_help_text("Copy image to clipboard")
+        property_content.with_action_button(&icons::COPY, "Copy image", move || {
+            if let Some(rgba) = self.image.to_rgba8_image(self.data_range.into()) {
+                let egui_image = egui::ColorImage::from_rgba_unmultiplied(
+                    [rgba.width() as _, rgba.height() as _],
+                    bytemuck::cast_slice(rgba.as_raw()),
+                );
+                ctx.egui_ctx().copy_image(egui_image);
+                re_log::info!("Copied image to clipboard");
+            } else {
+                re_log::error!("Invalid image");
+            }
+        })
     }
 
     pub fn inline_download_button<'a>(
--- a/crates/viewer/re_ui/src/list_item/item_buttons.rs
+++ b/crates/viewer/re_ui/src/list_item/item_buttons.rs
@@ -153,6 +153,7 @@ where
     ///
     /// The `alt_text` will be used for accessibility (e.g. read by screen readers),
     /// and is also how we can query the button in tests.
+    /// The `alt_text` will also be used for the tooltip.
     ///
     /// See [`Self::with_button`] for more information.
     #[inline]
@@ -169,6 +170,7 @@ where
     ///
     /// The `alt_text` will be used for accessibility (e.g. read by screen readers),
     /// and is also how we can query the button in tests.
+    /// The `alt_text` will also be used for the tooltip.
     ///
     /// See [`Self::with_button`] for more information.
     #[inline]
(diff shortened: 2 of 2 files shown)


crates/viewer/re_data_ui/src/video.rs
```diff
@@ -572,3 +569,141 @@ fn source_image_data_format_ui(ui: &mut egui::Ui, format: &SourceImageDataFormat
...
+                // Unfortunately we can't easily detect here if the Blob was _supposed_ to be a video, for that we'd need tagged components!
+                // (User may have confidently logged a non-video format as Video, we should tell them that!)
+                None
+            }
+            Err(VideoLoadError::UnrecognizedMimeType) => {
+                // If we couldn't detect the media type,
+                // we can't show an error for unrecognized formats since maybe this wasn't a video to begin with.
+                // See also `MediaTypeIsNotAVideo` case above.
+                None
+            }
```
lucasmerlin: Since we have tagged components now, can we remove this?

I guess I'd check if `blob_component_descriptor` is `archetypes::AssetVideo::descriptor_blob()`?

emilk:
Screenshot 2025-09-11 at 16 31 46
I love this design

emilk:
I miss some sort of tooltip for the button

crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -281,3 +125,158 @@ fn exif_ui(ui: &mut egui::Ui, key: StoredBlobCacheKey, blob: &re_types::datatype
         });
     }
 }
+
+// TODO: Hold references instead?
```
emilk: All the data is shallow-clone, so I doubt this will show up in the profiling

crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -281,3 +125,158 @@ fn exif_ui(ui: &mut egui::Ui, key: StoredBlobCacheKey, blob: &re_types::datatype
...
+pub struct BlobUi {
```
emilk, line 130: A docstring would be nice

crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -281,3 +125,158 @@ fn exif_ui(ui: &mut egui::Ui, key: StoredBlobCacheKey, blob: &re_types::datatype
...
+    image: Option<ImageUi>,
+    video: Option<VideoUi>,
+    row_id: Option<RowId>,
+    descr: ComponentDescriptor,
+    blob: re_types::datatypes::Blob,
+    media_type: Option<MediaType>,
```
emilk: nit/style: to me it makes more sense if the non-optional things (blob/descr) comes first, and the "derived" things (image/video uis) come last

crates/viewer/re_data_ui/src/image.rs
```diff
@@ -181,3 +170,265 @@ fn largest_size_that_fits_in(aspect_ratio: f32, max_size: Vec2) -> Vec2 {
...
+                Ok(png_bytes) => {
+                    let file_name = format!(
+                        "{}.png",
+                        entity_path
+                            .last()
+                            .map_or("image", |name| name.unescaped_str())
+                            .to_owned()
+                    );
+                    ctx.command_sender().save_file_dialog(
+                        re_capabilities::MainThreadToken::i_promise_i_am_on_the_main_thread(),
```
emilk: How can you promise this here though? better to construct this with `MainThreadToken::from_egui_ui` and pass it in by argument

crates/viewer/re_data_ui/src/image.rs
```diff
@@ -181,3 +170,265 @@ fn largest_size_that_fits_in(aspect_ratio: f32, max_size: Vec2) -> Vec2 {
...
+            if let Some(rgba) = self.image.to_rgba8_image(self.data_range.into()) {
+                let egui_image = egui::ColorImage::from_rgba_unmultiplied(
+                    [rgba.width() as _, rgba.height() as _],
+                    bytemuck::cast_slice(rgba.as_raw()),
+                );
+                ctx.egui_ctx().copy_image(egui_image);
+            } else {
+                re_log::error!("Invalid image");
+            }
+        })
```
emilk, line 359: Should have a tooltip, like `Copy image to system clipboard` (that was the old tooltip)

emilk: Maybe `with_action_button` could use the string it is given as the tooltip?

crates/viewer/re_data_ui/src/blob.rs
```diff
@@ -281,3 +125,158 @@ fn exif_ui(ui: &mut egui::Ui, key: StoredBlobCacheKey, blob: &re_types::datatype
...
+    pub fn data_ui(
+        &self,
+        ctx: &ViewerContext<'_>,
+        ui: &mut egui::Ui,
+        ui_layout: UiLayout,
+        query: &re_chunk_store::LatestAtQuery,
+        entity_path: &EntityPath,
+    ) {
+        if let Some(row_id) = self.row_id {
+            if !ui_layout.is_single_line() && ui_layout != UiLayout::Tooltip {
```
emilk: Can be collapsed!

emilk:
Also: clicking the copy button provides no feedback. I would expect a "Image copied to clipboard!" or something. Right now it just feels dead/broken, even though it actually works.

Just a `re_log::info` will do it 🤷

crates/viewer/re_data_ui/src/video.rs
emilk: ping @Wumpf

crates/viewer/re_data_ui/src/video.rs
```diff
@@ -572,3 +569,141 @@ fn source_image_data_format_ui(ui: &mut egui::Ui, format: &SourceImageDataFormat
...
+                        time,
+                        &buffers,
+                    );
+                }
+            }
+            Self::Asset(video_result, timestamp, blob) => {
+                video_asset_result_ui(ui, ui_layout, video_result);
+                if !ui_layout.is_single_line() && ui_layout != UiLayout::Tooltip {
+                    // Show a mini video player for video blobs:
+                    if let Ok(video) = video_result.as_ref() {
```
emilk:
```suggestion
                // Show a mini video player for video blobs:
                if !ui_layout.is_single_line() && ui_layout != UiLayout::Tooltip
                    && let Ok(video) = video_result.as_ref() {
```

PR #11182, opened 2025-09-11
Assorted table filtering improvements

abey79:
### Related

### What

- make the filter delete button smaller
- allow empty text field for numeric filter (empty filter matches everything, like for string columns)
- make the comparison popup smaller in the numeric filter ui
- disable filter for "recording link" column (and indeed any column which as a variant ui, for now)
- fix incorect popup spacing in certain circumstances (due to unintended egui id reuse)

### DIFF
--- a/crates/viewer/re_dataframe_ui/src/filters/filter_ui.rs
+++ b/crates/viewer/re_dataframe_ui/src/filters/filter_ui.rs
@@ -143,8 +143,20 @@ impl FilterState {
                 for row in rows {
                     ui.horizontal(|ui| {
                         for index in row {
-                            let filter_id = ui.make_persistent_id(index);
                             let filter = &mut self.filters[index];
+
+                            // egui uses this id to store the popup openness and size information,
+                            // so we must invalidate if the filter at a given index changes its
+                            // nature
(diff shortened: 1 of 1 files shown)


crates/viewer/re_dataframe_ui/src/filters/filter.rs
```diff
@@ -108,15 +108,15 @@ pub enum FilterOperation {
     /// For columns of lists of integers, only the first value is considered.
     IntCompares {
         operator: ComparisonOperator,
-        value: i128,
+        value: Option<i128>,
```
emilk, line 111: What does "Less than None" mean??

emilk: Ah, I think some confusion on my part is that this enum does two things: it represents the UI state AND it is the actual filter. That's fine, but it would be very useful with a comment explaining that.

abey79: Yeah, it really is UI state, and as such is converted to a filter (the actual filter being a dfn expression). I'll clarify this in future PRs, as I want to rework this very enum.

Issue #10817, opened 2025-08-06
Improve context menu for timeline


Wumpf:
<img src="https://uploads.linear.app/93919c5e-abe4-4cd5-b77d-7e3eafe3e62d/0ba738c5-b0f5-4606-8f19-893c6bded812/0f8d349b-e01b-4155-a086-ed7fe98e884a?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzkzOTE5YzVlLWFiZTQtNGNkNS1iNzdkLTdlM2VhZmUzZTYyZC8wYmE3MzhjNS1iMGY1LTQ2MDYtOGYxOS04OTNjNmJkZWQ4MTIvMGY4ZDM0OWItZTAxYi00MTU1LWEwODYtZWQ3ZmU5OGU4ODRhIiwiaWF0IjoxNzU2MTMyODQxLCJleHAiOjMzMzI2NjkyODQxfQ.39DVDn0fDRRCubddspWwVZXOyH7GA76GEb0B1NIrxCI " alt="image.png" width="1726" data-linear-height="310" />

<img src="https://uploads.linear.app/93919c5e-abe4-4cd5-b77d-7e3eafe3e62d/bc9a9442-3475-4d7f-a5ae-4d43a0be3f00/1b10efe2-70bc-4f22-8d0f-5cf4631e2d83?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzkzOTE5YzVlLWFiZTQtNGNkNS1iNzdkLTdlM2VhZmUzZTYyZC9iYzlhOTQ0Mi0zNDc1LTRkN2YtYTVhZS00ZDQzYTBiZTNmMDAvMWIxMGVmZTItNzBiYy00ZjIyLThkMGYtNWNmNDYzMWUyZDgzIiwiaWF0IjoxNzU2MTMyOTI0LCJleHAiOjMzMzI2NjkyOTI0fQ.uX5SAroFipHbeSB5P5S6LgLUsghczq-5LJKfaEjh-EU " alt="image.png" width="1724" data-linear-height="300" />

Add short cuts to educate and enable power users to do these operations without too many clicks.

Other ideas for copy:

* Link to trimmed range | Link to highlighted range
* Copy direct link to time range | Copy link to highlighted time range
* Copy link to selection only | Copy link to full + highlight selection

See
[...]

PR #11183, opened 2025-09-11
Add UrlContext which handles getting information for ViewerOpenUrl

IsseW:
### Related

Part of #10817

### What

This is to create a unified way to specify what should go in URLs.

Somewhat discussed here https://github.com/rerun-io/rerun/pull/11153#discussion_r2339690844, but had to go another direction because `ViewerOpenLink` is in the `re_viewer` crate and we need a way to specify what should go in the URL in ui crates like `re_time_panel`.

### DIFF
--- a/crates/store/re_log_types/src/index/time_real.rs
+++ b/crates/store/re_log_types/src/index/time_real.rs
@@ -142,13 +142,6 @@ impl From<TimeReal> for crate::Duration {
     }
 }
 
-impl From<TimeReal> for TimeInt {
-    #[inline]
-    fn from(time_real: TimeReal) -> Self {
-        Self::new_temporal(time_real.0.to_num())
-    }
-}
-
 // ---------------
 
 impl std::ops::Neg for TimeReal {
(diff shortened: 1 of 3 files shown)


crates/store/re_log_types/src/index/time_real.rs
```diff
@@ -142,6 +142,13 @@ impl From<TimeReal> for crate::Duration {
     }
 }
 
+impl From<TimeReal> for TimeInt {
+    #[inline]
+    fn from(time_real: TimeReal) -> Self {
+        Self::new_temporal(time_real.0.to_num())
+    }
+}
```
Wumpf: better to be explicit what we're doing and use the existing utility:
```suggestion
impl From<TimeReal> for TimeInt {
    #[inline]
    fn from(time_real: TimeReal) -> Self {
        time_real.floor()
    }
}
```

lil bit on the fence with that conversion though since it's hiding what we're doing 🤔

IsseW: Ahh yeah I should remove the From then, couldn't find an existing conversion for it, probably add some `floor` method to the time range as well if it doesn't exist already

Wumpf: it does :)

Wumpf: pretty sure I also missed it somewhere and didn't use it

PR #11186, opened 2025-09-11
Add share link button to time panel context menu

IsseW:
### Related

- Part of #10817
- Based on #11183

### What

Adds share buttons to timeline context menus (left click) as described in #10817

 - [x] Look into adding keybind and keybind in ui (leave for future PR)

### DIFF
--- a/crates/viewer/re_global_context/src/command_sender.rs
+++ b/crates/viewer/re_global_context/src/command_sender.rs
@@ -97,6 +97,8 @@ pub enum SystemCommand {
     },
 
     /// Copies the given url to the clipboard.
+    ///
+    /// On web this adds the viewer url as the base url.
     CopyViewerUrl(String),
 
     /// Set the item selection.


.typos.toml
```diff
@@ -175,4 +175,6 @@ extend-ignore-re = [
   "cros-codecs",
 
   "muh_scalars", # Weird name introduced in `crates/viewer/re_view_time_series/tests/basic.rs` introduced in https://github.com/rerun-io/rerun/pull/10713
+
+  "isse", # Name
```
Wumpf, line 179: 😄

crates/viewer/re_time_panel/src/time_control_ui.rs
```diff
@@ -65,6 +65,25 @@ You can also define your own timelines, e.g. for sensor time or camera frame num
...
+                .id(egui::Id::new("timeline select context menu"))
+                .open_memory(if response.secondary_clicked() {
+                    Some(egui::SetOpenCommand::Bool(true))
+                } else if response.clicked() {
+                    // Explicitly close the menu if the widget was clicked
+                    // Without this, the context menu would stay open if the user clicks the widget
+                    Some(egui::SetOpenCommand::Bool(false))
+                } else {
+                    None
+                })
```
Wumpf, line 92: TIL

crates/viewer/re_time_panel/src/time_control_ui.rs
```diff
@@ -65,6 +65,25 @@ You can also define your own timelines, e.g. for sensor time or camera frame num
...
+                .at_pointer_fixed()
```
Wumpf, line 93: I'm curious why this is needed, maybe also warrants a comment (?)

crates/viewer/re_viewer_context/src/url_context.rs
```diff
@@ -12,7 +13,60 @@ pub struct UrlContext {
...
+}
+
 impl UrlContext {
+    /// Information about what this url can do, can be used to hide UI for
+    /// sharing when we know it won't work.
+    ///
+    /// This should be kept up to date with what `ViewerOpenUrl::new` does
+    // TODO(isse): I don't really like how this has to be kept up to date manually with
+    // `ViewerOpenUrl`, would be nice to consolidate.
+    pub fn features(&self, store_hub: &crate::StoreHub) -> Option<UrlFeatures> {
```
Wumpf: I get how you arrived here but I very much concur that this is nasty to maintain.
Alternative suggestion: Where this is used right now we can just create the url with all the needed context and then see if the data we put in "sticks"

crates/viewer/re_time_panel/src/time_panel.rs
```diff
@@ -1774,29 +1778,44 @@ fn interact_with_streams_rect(
...
+    if let Some(url_features) = url_context.features(ctx.storage_context.hub) {
+        if let Some(selected_time_range) = time_ctrl.active_loop_selection()
+            && selected_time_range.contains(hovered_time)
+            && url_features.time_range
+        {
+            if ui.button("Copy link to trimmed range").clicked() {
+                ctx.command_sender()
+                    .send_system(url_context.without_timestamp().into_copy_cmd());
+            }
+        } else if url_features.fragment && ui.button("Copy link to timestamp").clicked() {
```
Wumpf: why else? We could still offer that regardless? If Katya argued otherwise earlier ignore me ;)
(didn't follow everything there)

Wumpf: also under "ignore me if Katya says otherwise": I'd show those buttons even when we don't support it but with `enabled false` to hint at the possibility

(nice to advertise our features and hoooopefully soonish this is a thing almost everywhere since everything goes through OSS server)

crates/viewer/re_viewer_context/src/url_context.rs
IsseW: I think maybe better to just always show them for now tbh. The problem with creating the url and why I created this structure to begin with was to try avoid moving `ViewerOpenUrl` to a lower crate. And the time panel does not have access to it.

crates/viewer/re_time_panel/src/time_panel.rs
IsseW: Went by the design, but yeah I'll hear with Katya.

crates/viewer/re_time_panel/src/time_control_ui.rs
IsseW: Added a comment, I think I have to do this to give the context menu a custom ID. Which otherwise would conflict with the combo box's popup id. Which made egui complain

crates/viewer/re_viewer_context/src/url_context.rs
Wumpf: ohhh didn't realize this was in `re_viewer_context`. That complicates things 🤔

IsseW: Moved `ViewerOpenUrl` to `re_viewer_context` now, and I think it's much nicer

Wumpf: 🥳

crates/viewer/re_global_context/src/command_sender.rs
```diff
@@ -96,12 +96,10 @@ pub enum SystemCommand {
...
-    /// Copies an url to the clipboard to the given a display mode, selection range, and url fragment.
-    CopyUrlWithContext {
-        display_mode: crate::DisplayMode,
-        time_range: Option<re_uri::TimeSelection>,
-        fragment: re_uri::Fragment,
-    },
+    /// Copies the given url to the clipboard.
+    ///
+    /// On web this adds the viewer url as the base url.
+    CopyViewerUrl(String),
```
Wumpf, line 102: good workaround/disentanglement/clarification!

Wumpf:
I thiiiink this is a known bug in egui to do a line wrap here when it shouldn't, but worth a quick poke

image

crates/viewer/re_viewer_context/src/open_url.rs
```diff
@@ -1,20 +1,33 @@
...
 /// A URL that points to a selection (typically an entity) within the currently active recording.
 pub const INTRA_RECORDING_URL_SCHEME: &str = "recording://";
 
 /// An eventListener for rrd posted from containing html
 pub const WEB_EVENT_LISTENER_SCHEME: &str = "web_event:";
 
+/// Origin used to show the examples ui in the redap browser.
+///
+/// Not actually a valid origin.
+pub static EXAMPLES_ORIGIN: LazyLock<re_uri::Origin> = LazyLock::new(|| re_uri::Origin {
```
Wumpf, line 25: much better place for this, agreed.

crates/viewer/re_viewer_context/Cargo.toml
```diff
@@ -30,6 +30,11 @@ re_format.workspace = true
 re_global_context.workspace = true
 re_log_types.workspace = true
 re_log.workspace = true
+re_log_encoding = { workspace = true, features = [
+  "decoder",
+  "encoder",
+  "stream_from_http",
+] }
```
Wumpf, line 37: a bit unfortunate to have this and `re_data_source` come in here now (no idea if they were already transitive dependencies), but willing to pay that price for the refactors in this PR

IsseW:
> I thiiiink this is a known bug in egui to do a line wrap here when it shouldn't, but worth a quick poke
> image

Yeah I've looked at that, and tried to set the context menu to a fixed size.

But that happens specifically if you first open the context menu with the shorter text, i.e "Copy link to timestamp" and then open the longer one. If you do it the other way around it doesn't wrap.

Maybe I could somehow reset egui memory of the size before it opens, but feels really hacky :/

PR #11189, opened 2025-09-12
Add support for filtering on null with boolean columns

abey79:
### Related

### What

Title.

Noteworthy:
- I made the call that for a inner and outer nullable column, filtering for null matches both outer null, and any inner null.
- This PR introduce the pattern of moving per-type implementation to a dedicated submodule (in this case `boolean_filter`). I intend to backport this pattern to string and numeric columns next time I work on them.

DNM: chained

image

### DIFF
--- a/crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
+++ b/crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
@@ -133,7 +133,7 @@ impl BooleanFilter {
                     .re_radio_value(query, None, primitive(ui, "null"))
                     .clicked();
             }
-        };
+        }
 
         if clicked {
             *action = FilterUiAction::CommitStateToBlueprint;


crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
```diff
@@ -0,0 +1,122 @@
...
+                } else {
+                    "false"
+                }
+            }
+            Self::Nullable(value) => {
+                if let Some(value) = value {
+                    if *value { "true" } else { "false" }
+                } else {
+                    "null"
+                }
```
emilk: I would love if we could use the same syntax highlighting here as we do in the arrow table…

and on that point "null" should have a different color than true/false, but that's outside the scope of this PR

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use re_ui::UiExt as _;
+
+use super::{FilterError, FilterUiAction, Nullability};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BooleanFilter {
+    NonNullable(bool),
+    Nullable(Option<bool>),
+}
```
emilk: This definitely needs a docstring what the hell does `NonNullable(false)` mean? That is nullable? Then why not `Nullable(true)`?

abey79: Right, adding one.

For the sake of the mental model, these "filter" objects are a data model of the UI. For boolean, the UI should either be true/false (for strictly non-nullable columns), or true/false/null (for columns with any form of nullability). This is what this object is about.

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
abey79: it's even worse: we're quite inconsistent with the representation of null. Sometime it's "-", but we also show some "[true, null, false]" things 😬

abey79: https://linear.app/rerun/issue/RR-2280/consistently-display-nulls

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+use super::{FilterError, FilterUiAction, Nullability};
+
+/// A filter for a boolean column.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BooleanFilter {
+    /// Filter for strictly non-nullable columns.
+    ///
+    /// This includes non-nullable primitive datatypes, as well as outer AND inner non-nullable nested
+    /// datatypes. In that case, the UI should not display the `null` option.
+    NonNullable(bool),
```
emilk, line 17: Does `NonNullable(true)` mean "Show only non-nullable columns?"
What does `NonNullable(false)` then mean?

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    /// Filter for nullable columns.
+    ///
+    /// In this case, the UI should display a `null` option. A value of `None` means nulls should be
+    /// matched.
+    Nullable(Option<bool>),
```
emilk, line 23: So does `Nullable(true)` shows nullable columns, or columns with nulls in it?

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+}
```
emilk, line 24: Can all this be just a 

```
enum BoolFilter {
    /// Show only `true`
    True,
   
    /// Show only `false`
    False,
   
    /// Show only `null`
    Null,
}
```

And then in the UI function, wether or not to show the `Null` alternative depends on the nullability of the column?

Because right now this is all so confusing.

Some example confusion:
* `Nullable(true)` sounds like the opposite of `NonNullable(false)`… but isn't
* What does it mean to have a `Nullable` filter on a non-nullable column?

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
abey79: This enum represent UI state. Sometime, we want the UI to offer a `null` choice, and sometime we don't. We make this choice when creating said state, based on the datatype of the column.

So `NonNullable(true)` really means "in the UI don't display a null choice and select the "true" option".

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
abey79: For clarity, `FilterOperation` and related sub-struct such as `BooleanFilter` are only half of a filter. The full filter is `Filter`, with combines a column name with a filter operation.

abey79: For the sake of limiting the rebase gymnastics, I'll do the following on my WIP follow-up PR:
- rename that substructure to `BooleanFilterOperation` (consistent with this being a sub structure of `FilterOperation`)
- clarify that all of these are UI states
- clarify the meaning of `Nullable` / `NonNullable` (I'm taking rename suggestions)

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
emilk: Ah, well then I think we really should go with this suggestion:

https://github.com/rerun-io/rerun/pull/11189/files#r2348254211

OR we have to explicitly call this "BooleanFilterUiState" and explain this in a docstring. "A filter for a boolean column." made me think this was a filter operation, not the UI state of a filter operation.

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
abey79: I disagres because it would introduce a dependency in the UI code on the column datatype which really doesn't need to exists.

Really, the current version of `BooleanFilter` is the _correct_ model for the state it's supposed to represent. If it is confusing, we should make it less confusing. Not less correct + introduce unwanted dependancies somewhere else.

abey79: I could split it into to two separate structs if that would help:
- `NullabelBoolFilter(Option<bool>)`
- `NonNullableBoolFilter(bool)`

edit: that would be arguably even more correct, as it would be unexpected for a give filter to turn from e.g. nullable to non-nullable.

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
abey79: I'd rather avoid s/Filter/FilterUiState/. I've already added the docstring in my wip followup pr.

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
emilk: Discussed IRL:

Let's be super-explicit and do

```rs
/// Filtering of nullable bools
NullabelBoolFilter(NullabelBoolFilter),

/// Filtering of non-nullable bools
NonNullabelBoolFilter(bool),
```

with 

```rs
enum NullableBoolFilter {
    /// Show only `true`
    True,
   
    /// Show only `false`
    False,
   
    /// Show only `null`
    Null,
}
```

but can wait for a cleanup-PR

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+                DataType::List(field) | DataType::ListView(field)
+                    if field.data_type() == &DataType::Boolean =>
+                {
+                    // `ANY` semantics
+                    if let Some(value) = value {
+                        Ok(array_has(col(column.clone()), lit(*value)))
+                    } else {
+                        Ok(col(column.clone()).is_null().or(array_element(
+                            array_sort(col(column.clone()), lit("ASC"), lit("NULLS FIRST")),
+                            lit(1),
```
emilk, line 74: wait, datafusion is 1-indexed?

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+            Self::NonNullable(value) => {
+                if *value {
+                    "true"
+                } else {
+                    "false"
+                }
+            }
+            Self::Nullable(value) => {
+                if let Some(value) = value {
+                    if *value { "true" } else { "false" }
```
emilk:
```suggestion
                    value.to_string()
```

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+                } else {
+                    "null"
+                }
+            }
```
emilk: I would like for this to have syntax highlighting, and ideally different color for "null"

emilk: Or is this only used for datafusion? It is hard to understand from the function docstring (or lack thereof) :P

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+
+            Self::Nullable(query) => {
+                ui.re_radio_value(query, Some(true), primitive(ui, "true"))
+                    .clicked()
+                    || ui
+                        .re_radio_value(query, Some(false), primitive(ui, "false"))
+                        .clicked()
+                    || ui
+                        .re_radio_value(query, None, primitive(ui, "null"))
+                        .clicked()
```
emilk: This will make it so that if the first radio button is clicked, then the others won't show up that frame

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+            }
+        };
+
+        if clicked {
+            *action = FilterUiAction::CommitStateToBlueprint;
+        }
```
emilk, line 140: Should we really commit on each click instead of actual change?
```suggestion
        if old_value != self {
            *action = FilterUiAction::CommitStateToBlueprint;
        }
```

crates/viewer/re_dataframe_ui/src/filters/filter.rs
```diff
@@ -9,7 +9,70 @@ use datafusion::logical_expr::{
...
 };
-use datafusion::prelude::{Column, Expr, array_has, array_to_string, col, lit, lower};
+use datafusion::prelude::{Column, Expr, array_to_string, col, lit, lower};
+
+use super::BooleanFilter;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Nullability {
+    pub inner: bool,
+    pub outer: bool,
```
emilk: Docstrings!

crates/viewer/re_dataframe_ui/src/filters/filter.rs
```diff
@@ -9,7 +9,70 @@ use datafusion::logical_expr::{
...
+impl std::fmt::Display for Nullability {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match (self.inner, self.outer) {
+            (false, false) => write!(f, "no_null"),
+            (false, true) => write!(f, "outer_null"),
+            (true, false) => write!(f, "inner_null"),
+            (true, true) => write!(f, "both_null"),
+        }
+    }
+}
```
emilk: You should probably use `Debug` formatting if it is for testing, but 🤷

crates/viewer/re_dataframe_ui/src/filters/filter.rs
```diff
@@ -9,7 +9,70 @@ use datafusion::logical_expr::{
...
+    pub const OUTER: Self = Self {
+        inner: false,
+        outer: true,
+    };
+
+    pub const ALL: &'static [Self] = &[Self::NONE, Self::INNER, Self::OUTER, Self::BOTH];
+
+    pub fn from_field(field: &Field) -> Self {
+        match field.data_type() {
+            DataType::List(inner_field) | DataType::ListView(inner_field) => Self {
```
emilk, line 64: Also: LargeList

crates/viewer/re_dataframe_ui/src/filters/filter.rs
```diff
@@ -122,11 +188,14 @@ pub enum FilterOperation {
     //TODO(ab): parameterise that over multiple string ops, e.g. "contains", "starts with", etc.
     StringContains(String),
 
-    BooleanEquals(bool),
+    Boolean(BooleanFilter),
```
emilk, line 196: To me it would be more natural to order datatypes from simplest to most advanced, i.e. "Bool -> Int -> Float -> String -> …"

abey79: Added to https://linear.app/rerun/issue/RR-2294/table-filters-refactor-the-code-for-better-clarity (skipping for now to limit conflict with chained PRs)

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
abey79: It's a minor usability details but yes you want it, so when you click on the default selected `true`, you get the commit.

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
abey79: SQL is, and, as a result, so is dfn in expression stuff. I ofc fell into that trap first.

crates/viewer/re_dataframe_ui/src/filters/boolean_filter.rs
abey79: It's syntax highlighted. This flow will be reworked later. As for null display, this is tracked here: https://linear.app/rerun/issue/RR-2280/consistently-display-nulls

crates/viewer/re_dataframe_ui/src/filters/filter.rs
abey79: This requires some non trivial testing. I already have a todo, but I also opened an issue: https://linear.app/rerun/issue/RR-2296/table-filter-support-more-container-types

PR #11190, opened 2025-09-12
Fix tooltips sometimes being way too big

emilk:
### Related
* Better version of https://github.com/rerun-io/rerun/pull/11187
* https://github.com/emilk/egui/issues/7537
* See [Slack thread](https://rerunio.slack.com/archives/C08TEFXNHB2/p1757603292722869)

### What
During a egui sizing pass, we allow widgets to inform the parent ui how large they would _like_ to be.
Unfortunately this meant that tooltips could grow without bounds.
This PR fixes this.

To help debug this now and in the future I added `ui.sanity_check();` calls here and there.
The plan is to remove those explicit calls in the future and have egui call that for us instead:
* https://github.com/emilk/egui/issues/7537

### DIFF
--- a/crates/viewer/re_ui/src/ui_ext.rs
+++ b/crates/viewer/re_ui/src/ui_ext.rs
@@ -39,7 +39,7 @@ pub trait UiExt {
 
         if cfg!(debug_assertions)
             && ui.is_tooltip()
-            && ui.spacing().tooltip_width + 100.0 < ui.max_rect().width()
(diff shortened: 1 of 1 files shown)


emilk:
So sometimes tooltips with ListItems in them will still be larger (by 100-150 points) than the set `tooltip_width`, because of complicated layout reasons.

The proper fix will have to wait (maybe after Lucas finishes up all the “intrinsic size” work in egui),
so I’ll just increate the threshold (by 10x) before we panic, for now. It only panics in debug builds btw.

PR #11193, opened 2025-09-12
Extract extra fields from ROS2 `PointCloud2` messages

grtlr:
### What

This extracts additional values that are associated with points in the `PointCloud2` ROS2 messages. The fields will be added to the `Points3d` archetype as partially tagged.

### Screenshot

image

### Test

We can use `autoware_all_sensors.mcap` to test.

### DIFF
--- a/crates/store/re_data_loader/tests/snapshots/test_mcap_loader__tests__assert_chunk_snapshot@supported_ros2_messages.mcap.snap
+++ b/crates/store/re_data_loader/tests/snapshots/test_mcap_loader__tests__assert_chunk_snapshot@supported_ros2_messages.mcap.snap
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:3fe309bc865899cacf618eb9a2c38ba16aa457bce7032b44defb44ef3161b1ed
-size 441989
+oid sha256:a5007101f4ae92b50171c12aed1b51ceb482503aade827248c8d8c8fabb146f4
+size 442998


crates/utils/re_mcap/src/parsers/ros2msg/sensor_msgs/point_cloud_2.rs
```diff
@@ -103,6 +123,22 @@ fn access(data: &[u8], datatype: PointFieldDatatype, is_big_endian: bool) -> std
     }
 }
 
+impl From<PointFieldDatatype> for DataType {
+    fn from(value: PointFieldDatatype) -> Self {
+        match value {
+            PointFieldDatatype::Unknown => unreachable!(), // Not part of the MCAP spec
```
oxkitsune:
```suggestion
            PointFieldDatatype::Unknown => unreachable!(), // Not part of the ROS2 spec
```
nit

emilk: Is this really unreachable, no matter what is in an mcap file? Remember: we should NOT crash on bad input.

crates/utils/re_mcap/src/parsers/ros2msg/sensor_msgs/point_cloud_2.rs
```diff
@@ -204,6 +364,31 @@ impl MessageParser for PointCloud2MessageParser {
             &point_cloud.fields,
         );
 
+        // We lazily initialize the builders for the values in the point cloud.
+        if points.len() != point_cloud.fields.len() {
```
emilk: Is this only the first time, i.e. when `point.is_empty()`? Or could this check ever be `4 != 6`?

It currently reads as the latter, which makes me thing we also need to recreate `points` if the field names or datatype changes.

crates/utils/re_mcap/src/parsers/ros2msg/sensor_msgs/point_cloud_2.rs
```diff
@@ -362,6 +549,18 @@ impl MessageParser for PointCloud2MessageParser {
                 ),
             ]
             .into_iter()
+            .chain(points.into_iter().filter_map(|(name, mut builder)| {
+                points_3ds.as_ref()?;
+                if ["x", "y", "z"].contains(&name.as_str()) {
+                    None
+                } else {
```
emilk, line 564: Why though?

crates/utils/re_mcap/src/parsers/ros2msg/sensor_msgs/point_cloud_2.rs
grtlr: Yes, it's per the spec: https://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/PointField.html

grtlr: I'll add the link as a comment.

crates/utils/re_mcap/src/parsers/ros2msg/sensor_msgs/point_cloud_2.rs
grtlr: You're right, `points` was a bad name, changed it to `extracted_fields` to make it clearer.

grtlr:
Thank you @oxkitsune and @emilk for the feedback.

I've addressed your comments if you want to take another look.

crates/utils/re_mcap/src/parsers/ros2msg/sensor_msgs/point_cloud_2.rs
```diff
@@ -103,6 +123,24 @@ fn access(data: &[u8], datatype: PointFieldDatatype, is_big_endian: bool) -> std
     }
 }
 
+impl From<PointFieldDatatype> for DataType {
+    fn from(value: PointFieldDatatype) -> Self {
+        match value {
+            // Not part of the MCAP spec
+            // // https://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/PointField.html
+            PointFieldDatatype::Unknown => unreachable!(),
```
emilk: It still sounds like you can create an MCAP file (that is _not_ spec compliant) that can crash the Rerun viewer.

For instance: a new MCAP spec is released with a `PointFieldDatatype::Int128`, and the old parser will map that to `Unknown` and Rerun crashes. NOT GOOD.

Instead: return an error

grtlr: The loader runs in a separate thread that is specific to that MCAP file.

But overall I agree, we can't know in which context the loader will be running. I will add the error mechanism.

emilk: > The loader runs in a separate thread that is specific to that MCAP file

We compile with `panic=abort`, so a panic on any thread shuts down the whole process.

grtlr: TIL! I've made the required changes PR.

Issue #11194, opened 2025-09-12
Improve the code for wrapping filters in the filter bar


abey79:
egui currently doesn't support wrapping complex widget (aka `Frame`) in `ui.horizontal_wrapped`, so I had to roll manual wrapping code. This should be cleaned up when egui is improved.

cc @lucasmerlin

PR #11195, opened 2025-09-12
Wrap filter bar

abey79:
### Related

### What

Make the filter bar wrapping. 

This uses manual code, hopefully can be improved with proper support for that in egui:
- #11194

### DIFF
--- a/crates/viewer/re_dataframe_ui/src/filters/filter_ui.rs
+++ b/crates/viewer/re_dataframe_ui/src/filters/filter_ui.rs
@@ -126,20 +126,17 @@ impl FilterState {
                     .collect::<Vec<_>>();
                 let item_spacing = ui.style().spacing.item_spacing.x;
                 let available_width = ui.available_width();
-                let mut rows = vec![vec![]]; // single empty row
+                let mut rows = vec1::vec1![vec![]];
                 let mut current_left_position = 0.0;
                 for (index, prepared_ui) in prepared_uis.iter().enumerate() {
                     if current_left_position > 0.0
-                        && current_left_position + prepared_ui.desired_width() + item_spacing
-                            > available_width - item_spacing
+                        && current_left_position + prepared_ui.desired_width() > available_width
                     {
                         rows.push(vec![]);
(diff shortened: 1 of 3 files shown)


crates/viewer/re_dataframe_ui/src/filters/filter_ui.rs
```diff
@@ -112,27 +114,59 @@ impl FilterState {
...
+                    .map(|filter| filter.prepare_ui(ui))
+                    .collect::<Vec<_>>();
+                let item_spacing = ui.style().spacing.item_spacing.x;
+                let available_width = ui.available_width();
+                let mut rows = vec![vec![]]; // single empty row
+                let mut current_left_position = 0.0;
+                for (index, prepared_ui) in prepared_uis.iter().enumerate() {
+                    if current_left_position > 0.0
+                        && current_left_position + prepared_ui.desired_width() + item_spacing
+                            > available_width - item_spacing
```
emilk: looks like item_spacing is applied to both sides of the > by mistake

crates/viewer/re_dataframe_ui/src/filters/filter_ui.rs
```diff
@@ -112,27 +114,59 @@ impl FilterState {
...
+                    {
+                        rows.push(vec![]);
+                        current_left_position = 0.0;
+                    }
+
+                    rows.last_mut()
+                        .expect("we always have at least one row")
```
emilk: Can use [`vec1`](https://crates.io/crates/vec1) to avoid this but 🤷

crates/viewer/re_dataframe_ui/src/filters/filter_ui.rs
emilk: Also item_spacing is already applied to `current_left_position` so probably it shouldn't appear at all here?

abey79: not sure what i was thinking, exactly 🤦🏻

PR #11197, opened 2025-09-12
Fix incorrect application id when dragging and dropping files

oxkitsune:
### What

When drag-and-dropping a file into the web viewer, we only get the file name not the full path. This results in the store being created with a random application id. 
This makes the picked application id fall back to the file name, if there's no file path available.

### DIFF
--- a/crates/store/re_data_loader/src/loader_mcap.rs
+++ b/crates/store/re_data_loader/src/loader_mcap.rs
@@ -189,9 +189,7 @@ pub fn load_mcap(
     raw_fallback_enabled: bool,
 ) -> Result<(), DataLoaderError> {
     re_tracing::profile_function!();
-
-    // If there's an open store, use it. If there's no store yet, use the recommended id.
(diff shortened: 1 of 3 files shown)


grtlr:
Still seeing some problems. Will add more details here after standup.

crates/store/re_data_loader/src/loader_mcap.rs
```diff
@@ -189,9 +189,7 @@ pub fn load_mcap(
     raw_fallback_enabled: bool,
 ) -> Result<(), DataLoaderError> {
     re_tracing::profile_function!();
-
-    // If there's an open store, use it. If there's no store yet, use the recommended id.
-    let store_id = settings.opened_store_id_or_recommended();
```
emilk, line 194: Is `opened_store_id_or_recommended` still being used anywhere?

oxkitsune: Yes, in an example where it makes sense I think.

PR #11198, opened 2025-09-12
Bump our minimum macos for examples

ntjohnson1:
So we don't have to build opencv-contrib from scratch (see the opencv_contrib .tar.gz to .wheel in pixi.lock).

If we don't like this we should also be able to pin opencv-contrib to 4.10 where they still support macos12.


ntjohnson1:
CC: @aedm

abey79 (approved):
It's fair to expect non ancient OSes from contributors, if it means better DX across the board

ntjohnson1:
> It's fair to expect non ancient OSes from contributors, if it means better DX across the board

Technically I just bumped the version for running examples. Our base env is much more accepting of older macOS https://github.com/rerun-io/rerun/blob/af84f90070ccc84ecd3b133a70fec7a3fb32232c/pixi.toml#L30 (not sure if there is a need for that or if it causes other weird problems)

Issue #11191, opened 2025-09-12
Opening a link to a custom timeline doesn't work


IsseW:
Opening some a link that links to a specific timestamp on a timeline other than the default timeline or the `log_time` timeline instead opens the default timeline at the start.

`pixi run rerun rerun://sandbox.redap.rerun.io:443/dataset/186446580E79EA77776cd5d9a5842c6c?partition_id=AUTOLab_0d4edc83_2023_11_03_15h_45m_58s#when=log_tick@231`

Wumpf:
We talked about this and concluded that the timeline should support switching to timelines that aren't there yet. The custom timeline wouldn't show up in the timeline selection (i.e. you can't go back if you selected a valid one) but nothing in the system should prevent us from staying with the empty/invalid timeline.

To make this state easier discoverable, we should add some sort of space filling message on the empty/invalid timeline, ideally allowing you to pick another one easily.

Note that we gonna run into this regardless with blueprint/viewer instrumentation APIs as users may set timelines before setting any data.

PR #11213, opened 2025-09-15
Fix loading an url linking to a log-tick timeline

IsseW:
### Related

Part of #11191

### What

Fixes a bug where opening a link with a timestamp on the log tick timeline landed on the default timeline.


crates/store/re_entity_db/src/times_per_timeline.rs
```diff
@@ -45,14 +45,19 @@ impl TimesPerTimeline {
...
+                *Timeline::log_time().name(),
+                TimelineStats::new(Timeline::log_time()),
+            ),
+            (
+                *Timeline::log_tick().name(),
+                TimelineStats::new(Timeline::log_tick()),
+            ),
+        ]))
     }
 }
```
Wumpf, line 62: tbh at this point I'm surprised we need to have a default `log_time` timeline in the first place and I'm skeptical of more leaning in here and unsure whether this is the right fix:
* `log_time` & `log_tick` are today automatically generated on the `log` family of SDK methods but we don't do so if you use `send_columns`, so it is actually possible to create recordings without either.
* if I link you to you a timeline that never (??) exists and we switch back to default, then that wouldn't be a bug. But if I link you to one (might be a custom one) that comes in while loading the URL and Viewer switches to a default one that _is_ a bug --> does this mean this PR fixes this only for the corner case of `log_tick` link sbut not for others

IsseW: Yeah so currently this would only fix it for `log_tick`, and I'm guessing it only works for `log_time` because it is in that default function.

Took a deeper look into this, and from what I can tell we don't really receive information about the `log_tick` timeline (and any custom timeline I assume) after ~80 update functions after we try to go to that timeline.

So there's some async stuff happening here for the arrow chunk message that adds the timeline to actually be applied.

It would be nice if recordings had some "header" to tell about what timelines exist, right now I can't really think of any clean way to fix this properly.

Wumpf: > It would be nice if recordings had some "header" to tell about what timelines exist, right now I can't really think of any clean way to fix this properly.

definitely where we want to go! we want a phase where we first download the schema (all timelines, entities & component types) and then selectively download the rest. But we don't have that yet which makes this real hard 🤔Actually, @teh-cmc, I'm curious now how far are we away from that for dataplatform things (schema downloading I mean)

crates/viewer/re_time_panel/tests/snapshots/various_filters-_to_the_.png
emilk, line 1: Did the default timeline switch from `log_time` to `log_tick` now?

grtlr: No: https://github.com/rerun-io/rerun/blob/0c9ac82f0198b56337e0cfaa29cec7736bdcaf1a/crates/viewer/re_viewer_context/src/time_control.rs#L765C9-L765C29

So I would not expect this PR to change that. It seems that the test context uses a different heuristic for the default timeline (`BTreeMap` probably `tick` < `time`). In the viewer we use the `default_timeline` as linked above.

IsseW: In `TestContext::new` it explicitly sets the timeline to `log_tick`, which before this change didn't work I think?

emilk: Nice; thanks

emilk: > In `TestContext::new` it explicitly sets the timeline to `log_tick`, which before this change didn't work I think?

That makes me think we should probably set the default timeline in our tests to `log_time` instead, since in general that is preferred over `log_tick`. Not sure if this PR is the right place for it though.

emilk (approved):

emilk:
Maybe @Wumpf wants the final word?

IsseW:
I'm personally a bit on the fence whether we want to merge this, on one hand it does make sharing the `log_tick` timeline "work", the same way sharing the "log_time" timeline currently works. By assuming it will always be there.

On the other hand this is a hacky fix for this, and will only work for those timelines.

But if you think it's worth it @Wumpf, feel free to merge

Wumpf:
Agreed with @IsseW, let's not merge it; we don't want to move into a world where we assume _more_ timelines to be pre-existing and it doesn't solve the issue for anything but the hardcoded `log_tick` anyways which limits the fix too much.
I'd change my mind if this specifically comes up as a user pain :)

IsseW:
Preceded by #11333

PR #11214, opened 2025-09-15
Fix outlines not showing sometimes for individual instances

Wumpf:
Fixes a regression from
* https://github.com/rerun-io/rerun/pull/11212

that sometimes we'd outline the right mesh instance. Added a image comparison test - tried to add this to the unit tests but it proved tricky.
Confirmed that issue get caught by the comparison test.

TODO:
* [x] calibrate image threshhold to pass on ci

### DIFF
--- a/crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
+++ b/crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
@@ -312,21 +312,18 @@ impl MeshDrawData {
                     for chunk in instances.chunk_by(|(_, phases_a), (_, phases_b)| {
                         phases_a.contains(phase) == phases_b.contains(phase)
                     }) {
-                        if !chunk[0].1.contains(phase) {
-                            continue;
-                        }
-
                         let num_instances = chunk.len() as u32;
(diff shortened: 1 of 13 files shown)


crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
```diff
@@ -312,21 +312,18 @@ impl MeshDrawData {
                     for chunk in instances.chunk_by(|(_, phases_a), (_, phases_b)| {
                         phases_a.contains(phase) == phases_b.contains(phase)
                     }) {
-                        if !chunk[0].1.contains(phase) {
-                            continue;
```
IsseW, line 316: ahh, yeah that'd mess things up 😅

IsseW (approved):
Looks good, nice to have a test for it!

crates/viewer/re_renderer/src/renderer/mesh_renderer.rs
Wumpf: yup! I broke it when removing the `unsafe` thing and replaced with tallying. And then didn't test enough afterwards

Issue #10781, opened 2025-08-01
Send intermediate results/chunks for datafusion table providers


timsaucer:
**Is your feature request related to a problem? Please describe.**

This is a follow on to https://github.com/rerun-io/rerun/pull/10698 

In that PR we send a fully partition's worth of data at a time. Instead we should be able to stream out intermediate results once we have enough chunks to do so.

**Describe the solution you'd like**

From our first gRPC call, we know the chunk IDs and their associated time. We should be able to determine when we have received all prior rows for a given time of interest. Once all of those chunks have been received, stream out a row. This will give intermediate results instead of waiting for the entire partition to be received.

**Describe alternatives you've considered**

Leave as is, which is still better than collecting everything in memory

**Additional context**

- It's unclear to me how we can verify the timing of when the chunks arrived
- There is an opportunity to exit early when we've reached our row limit.

PR #11215, opened 2025-09-15
Use `FetchChunks` instead of `GetChunks` for chunk fetching on query execution path

zehiko:
### What

We've recently added support for the new ``FetchChunks`` API on the Rerun Cloud side that should be more optimal and avoid re-querying internal indexes. As part of initial ``QueryDataset`` request we're asking the server to give us ``ChunkKey`` information that holds location details we'll send back to the server as part of ``FetchChunks``. 

As part of this change I've also removed the remaining chunk tracking as it simplifies some pieces of code as it was currently unused, but happy to bring it back if we have some future plans for it. 

### Testing

Did local desktop perf testing for sample50 and sample500 and it's looking like a significant improvement (it's getting better as we have more chunks, and this is somewhat expected as chunk querying is expensive part of existing ``GetChunks`` server side path). I did confirm that we get exact same data back (although slightly different ordering). 

Before:
```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Test Name                                 ┃ duration_secs (avg) ┃ duration_secs (count) ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━┩
│ droid-droid-sample500:sample_index_values │ 4.9939              │ 1.0000                │
└───────────────────────────────────────────┴─────────────────────┴───────────────────────┘
```

After:
```

┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Test Name                                 ┃ duration_secs (avg) ┃ duration_secs (count) ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━┩
│ droid-droid-sample500:sample_index_values │ 0.4927              │ 1.0000                │
└───────────────────────────────────────────┴─────────────────────┴───────────────────────┘
```

### Few more notes:
* while testing I noticed that we fire 1 GetChunks request per partition in the current (before this change) path, but after this change there's grouping and we fire substantially less ``FetchChunks`` calls (like 50x less, so before for sample500 I saw 493 GetChunks requests vs 14 FetchChunks requests now). I believe I've kept the same hash based partitioning logic
* I removed ``dataset_id`` from wasm implementation and did the same we do in native one, again to simplify things and not carry things around, but I'm admittedly not sure about the consequence of this

### DIFF
--- a/crates/store/re_datafusion/src/dataframe_query_provider.rs
+++ b/crates/store/re_datafusion/src/dataframe_query_provider.rs
@@ -401,7 +401,11 @@ async fn chunk_store_cpu_worker_thread(
 /// to the gRPC channel for chunks coming in from the data platform. This loop is started
 /// up by the execute fn of the physical plan, so we will start one per output partition,
 /// which is different from the `partition_id`. The sorting by time index will happen within
-///  the cpu worker thread.
+/// the cpu worker thread.
+/// `chunk_infos` is a list of batches with chunk information where each batch has info for
+/// a *single partition*. We also expect these to be previously sorted by partition id, otherwise
+/// our suggestion to the query planner that inputs are sorted by partition id will be incorrect.
+/// See `group_chunk_infos_by_partition_id` and `execute` for more details.
 #[tracing::instrument(level = "trace", skip_all)]
 async fn chunk_stream_io_loop(
     mut client: ConnectionClient,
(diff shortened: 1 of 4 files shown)


crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -433,39 +399,38 @@ async fn chunk_store_cpu_worker_thread(
...
-            if output_channel.send(chunk_and_partition_id).await.is_err() {
-                break;
-            }
+    let fetch_chunks_response_stream = client
+        .inner()
+        .fetch_chunks(fetch_chunks_request)
+        .instrument(tracing::trace_span!("chunk_stream_io_loop"))
+        .await
+        .map_err(|err| exec_datafusion_err!("{err}"))?
+        .into_inner();
```
abey79: If feasible, please move this to a proper wrapper call in `client`. (In an ideal world, we never have to call `inner` outside of `re_redap_client`.)

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -505,38 +470,34 @@ impl ExecutionPlan for PartitionStreamExec {
...
-                chunk_rx,
-                batches_tx,
-                Arc::clone(&self.chunk_info),
-                query_expression,
-                projected_schema,
-            ),
+            // TODO(zehiko) - do we really need chunk infos inside the cpu worker thread? It seems it was only
+            // used to check if received chunks are those we expected, but that's a server side check.
+            // Remaining usage was tracking if all were received, but that was not used for anything.
+            chunk_store_cpu_worker_thread(chunk_rx, batches_tx, query_expression, projected_schema),
```
timsaucer: The reason to keep this info in the worker thread was to support sending out intermediate results, which we are not currently doing. There were a few bits of commented out code in this PR alluding to it, but it must not have been clear.

timsaucer:
Follow on work: https://github.com/rerun-io/rerun/issues/10781

crates/store/re_datafusion/src/dataframe_query_provider.rs
zehiko: ok to merge as is, or should I bring it back? (although what we now pass around is slightly different since we're no longer building ``ChunkInfo``s, so perhaps best to tackle this as part of #10781 . But of course, don't want to create extra work for you!)

timsaucer: I linked to #10781 for the follow on, so I think ok to merge as is

zehiko: that makes sense, I'll drop a todo with that PR in there as well. thanks!

crates/store/re_datafusion/src/dataframe_query_provider.rs
zehiko: (didn't forget about this, will address a in a small follow up)

zehiko:
Similarly as we were doing before for ``get_chunks``, I've switched to firing off a single request per partition, sequentially. This comes with perf losses (I already see 2x slower execution compared to before) that are probably going to be bigger for larger datasets. We could introduce firing concurrent ``fetch_chunks`` requests with preserved order, but then we need to be careful about the http2 window size and ensure that one of these concurrent background data stream doesn't exhaust the window while we're polling for the next (rust) stream in order. 

Doing ordering on the server side is possible, but that also might become much more complicated as that fetch path grows and scales. 

If nothing else, we know we have a lever to pull here for improving perf (and sacrificing something else).

Issue #10739, opened 2025-07-23
Simplify Row Count Inspection Viewer and SDK


ntjohnson1:
## Motivation
I wanted to confirm that I had processed data correctly so the first check was to see how many rows were in the column of interest.

In the viewer hovering over the streams window show how many times things were LOGGED to that.

<img width="1919" height="251" alt="Image" src="https://github.com/user-attachments/assets/8f2da17d-3575-40ed-ae06-82a57c01a148" />

Meanwhile in the sdk we can load an rrd and count rows like such.
```python
>>rec = rr.dataframe.load_recording(rrd_name)
>>view = rec.view(index="log_tick", contents={"/camera/ext1": "embeddings"})
>>rb = view.select()
>>total_rows = 0
>>for batch in rb:
...total_rows += batch.num_rows
>>print(total_rows)
5120
```

These two values can be different if someone has a typo where they log duplicates:
```python
def create_sample() -> str:
    batch_size = 1024
    embedding_dim = 768
    rrd_name = "<some_path>"

    stream = rr.RecordingStream(application_id="sample_name", recording_id="uuid", send_properties=True)
    stream.save(rrd_name)
    data = np.zeros((batch_size, embedding_dim), dtype=np.uint8)

    def write_n(n: int) -> None:
        for i in range(n):
            time_column = rr.TimeColumn("log_tick", sequence=np.arange(i * batch_size, (i + 1) * batch_size))
            stream.send_columns(
                entity_path="/camera/ext1",
                indexes=[time_column],
                columns=rr.AnyValues.columns(embeddings=data.flatten()).partition([embedding_dim] * batch_size),
            )
        stream.flush(True)

    write_n(5)
    write_n(5)
    return rrd_name
```

## Suggestions

It would be nice to get the actual row count in an easier fashion for a given column. In the viewer when we select an entity we get a nice side window that I believe comes from our schema. If our schema provided a row count that might make a simpler way to get the answer on the sdk and provide it through the gui as well.
<img width="605" height="161" alt="Image" src="https://github.com/user-attachments/assets/dc85820f-b7c2-4780-8857-590618515b68" />

ntjohnson1:
Ok based on the other issue it looks like the row count in the viewer is CORRECT but the dataframe API isn't reflective of what's actually in the dataset. Lower priority but I still think it might be a nice to have to show the row count in the selection.

grtlr:
Somewhat related to this request—we plan to show the number of events in the timeline picker:

[...]

PR #11217, opened 2025-09-15
Set timeline with most events as default timeline

grtlr:
### Related

* Closes https://linear.app/rerun/issue/RR-2282/better-default-timeline-when-loading-mcap.
* Related: https://github.com/rerun-io/rerun/issues/10739

### What

This PR changes the `default_timeline` logic (when no timeline was selected by the user) to always switch to the timeline with the most events on it.

It also adds the number of events to the timeline selector in the time panel.

> [!IMPORTANT]
> It would be very nice if we could right-align the subdued text. That way it would be much easier to see, at a glance, which of the timelines has the most events (because the text would be the longest). **If someone knows how to do this quickly, please point out in the review.**

| dark | light |
|--------|--------|
| image | image |

### DIFF
--- a/crates/store/re_entity_db/src/times_per_timeline.rs
+++ b/crates/store/re_entity_db/src/times_per_timeline.rs
@@ -25,6 +25,11 @@ impl TimelineStats {
     }
 
     pub fn num_events(&self) -> u64 {
+        debug_assert_eq!(
+            self.per_time.values().sum::<u64>(),
+            self.total_count,
+            "[DEBUG ASSERT] book keeping drifted"
+        );
         self.total_count
     }
 }
--- a/crates/viewer/re_viewer_context/src/time_control.rs
+++ b/crates/viewer/re_viewer_context/src/time_control.rs
@@ -825,6 +825,7 @@ mod tests {
             timeline,
             // Dummy `TimeInt` because were only interested in the counts.
             per_time: std::iter::once((TimeInt::ZERO, num)).collect(),
+            total_count: num,
         }
     }
 


crates/store/re_entity_db/src/times_per_timeline.rs
```diff
@@ -21,6 +21,10 @@ impl TimelineStats {
             per_time: Default::default(),
         }
     }
+
+    pub fn num_events(&self) -> u64 {
+        self.per_time.values().sum()
```
Wumpf: is this a fast operation...? could it be that there's many thousands of events here?

grtlr: Potentially, yes. But we are already computing the `BTreeMaps` that hold these events and passing them here:

* https://github.com/rerun-io/rerun/pull/11217/files#diff-3230cbadb3697984dea331036e90933d9881500b075c2bd7adcac47c4776c21cR552.

So adding over these elements should be a ~constant~ small factor. But I have not confirmed the actual performance impact. Looking into whether our benchmarks capture this.

Wumpf:
Fixed 😄 https://github.com/rerun-io/rerun/pull/11217/commits/da2627047955d09aa87f9ce1b17474df21720c5d
egui atoms just worked <3

image

crates/viewer/re_viewer_context/src/time_control.rs
```diff
@@ -733,23 +734,24 @@ fn range(values: &TimeCounts) -> AbsoluteTimeRange {
...
-    for timeline in timelines {
-        if timeline == &Timeline::log_tick() {
-            found_log_tick = true;
-        } else if timeline == &Timeline::log_time() {
-            found_log_time = true;
-        } else {
-            return *timeline;
+/// Pick the timeline that should be the default, by number of elements and prioritizing user-defined ones.
+fn default_timeline<'a>(timelines: impl IntoIterator<Item = &'a TimelineStats>) -> Timeline {
+    // Helper function that acts as a tie-braker.
```
grtlr:
```suggestion
    // Helper function that acts as a tie-breaker.
```

crates/store/re_entity_db/src/times_per_timeline.rs
grtlr: Ok, computing `default_timeline` twice per frame takes around 0.450ms in total on my machine (MBP M4) in for a quite representative MCAP file (4096 rows in `EntityDB`).

I think that's worth optimizing, so I did. I'm now caching the `total_count` in `TimelineStats`. Ofc this adds some overhead on ingestion, but probably negligible.

crates/store/re_entity_db/src/times_per_timeline.rs
```diff
@@ -8,19 +8,30 @@ use re_log_types::{TimeInt, Timeline};
...
+            total_count: 0,
         }
     }
+
+    pub fn num_events(&self) -> u64 {
+        debug_assert_eq!(
+            self.per_time.values().sum::<u64>(),
+            self.total_count,
+            "[DEBUG ASSERT] book keeping drifted"
+        );
```
Wumpf, line 32: nit: I miss the debug assert part like everyone else, but don't debug asserts print that they're debug asserts?
maybe nit: so if this sometimes takes half a millisecond, isn't that too slow for our debug build already?

emilk: > don't debug asserts print that they're debug asserts

nope:

https://doc.rust-lang.org/src/core/macros/mod.rs.html#306

Maybe someone should make a PR about it, or at least open an issue

grtlr: Here you go: https://internals.rust-lang.org/t/indicate-debug-assert-in-output-of-panic/23534

PR #11218, opened 2025-09-15
Group non-Unix epoch ROS 2 messages on a shared timeline

oxkitsune:
### What

Groups all ROS 2 messages without a Unix epoch timestamp onto a single shared timeline (`ros2_duration`), instead of assigning each one its own separate timeline.

### DIFF
--- a/crates/store/re_data_loader/tests/snapshots/test_mcap_loader__tests__assert_chunk_snapshot@supported_ros2_messages.mcap.snap
+++ b/crates/store/re_data_loader/tests/snapshots/test_mcap_loader__tests__assert_chunk_snapshot@supported_ros2_messages.mcap.snap
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:a5007101f4ae92b50171c12aed1b51ceb482503aade827248c8d8c8fabb146f4
-size 442998
+oid sha256:9edae2d9c157373606a3909fccc9ad3dca5c4f829c0b9e8a3a3f74594c9fdc8e
+size 443218


grtlr (changes requested):
Code looks good—thank you! Can you please at a paragraph about this to https://github.com/rerun-io/rerun/blob/main/docs/content/reference/mcap/message-formats.md?

docs/content/reference/mcap/message-formats.md
```diff
@@ -16,6 +16,10 @@ The `ros2msg` layer provides automatic visualization for common ROS2 message typ
 
 We are continually adding support for more standard ROS2 message types. For the complete list of currently supported messages, see the [ROS2 message parsers in our codebase](../../../../crates/utils/re_mcap/src/layers/ros2.rs).
 
+### Timelines
+
+Timestamps within Unix time range (1990-2100) create `ros2_timestamp` timelines. Values outside this range create `ros2_duration` timelines for relative time from custom epochs.
```
grtlr:
```suggestion
In addition to the `log_time` and `publish_time` timestamps that are part of an MCAP message, some ROS message payloads can have an additional [`Header`]( https://docs.ros.org/en/noetic/api/std_msgs/html/msg/Header.html) that may also contain timestamp information. These timestamps are put onto specific `ros2_*` timelines.

Timestamps within Unix time range (1990-2100) create a `ros2_timestamp` timeline. Values outside this range create a `ros2_duration` timeline representing relative time from custom epochs.
```

grtlr (approved):
Nice! I added a bit more detail on the source of the timestamps.

PR #11223, opened 2025-09-15
Nick/fix cpp arrow again

ntjohnson1:
### Related
Attempting to fix https://github.com/rerun-io/rerun/issues/10796 again. I'm not really sure what I messed up when testing locally last time but it definitely is still broken.

### What
Arrow doesn't pass through the CMAKE versioning flag to MIMALLOC so things still barf. This applies a patch to pass through the variable (the variable was already set [here](https://github.com/rerun-io/rerun/pull/10799))

Validation:
I just set our cmake in pixi to be 4.*
```diff
diff --git a/pixi.toml b/pixi.toml
index 505e335f5..486ff39fe 100644
--- a/pixi.toml
+++ b/pixi.toml
@@ -593,7 +593,7 @@ aiohttp = ">=3.9.3,<3.10"         # For `zombie_todos.py`
 attrs = ">=23.1.0"
 clang = "16.0.6.*"
 clang-tools = "16.0.6.*"          # clang-format
-cmake = "3.27.6.*"
+cmake = "4.*"
 colorama = ">=0.4.6,<0.5"
 doxygen = "1.9.7.*"               # Make sure to use a version that is compatible with the theme we're using, see https://github.com/jothepro/doxygen-awesome-css/blob/v2.2.1/README.md
 fd-find = ">=10.1.0"              # Used by `cpp-fmt` to find C++ files
```

```console
$ rm -rf build
$ cmake -G 'Ninja' -B build/debug -S . -DCMAKE_BUILD_TYPE=Debug
$ cmake --build build/debug --config Debug --target ALL
```


Wumpf:
previous patch attempt landed in 0.24.1 it seems, so adding to changelog

Wumpf (commented):
this makes sense to me I think.

We have to make sure this also works on our deployed build-zip 🤔 
CI should produce that & test it, I have a look to confirm

Wumpf:
Looks like only nightly is running `bundle-and-upload-rerun_cpp`
We should run `python3 ./scripts/ci/bundle_and_upload_rerun_cpp.py` locally - we have to ensure the new patch file is part of the distribution, don't think it is right now.

emilk:
@rerun-bot full-check

Wumpf (approved):
```
pixi run -e py python ./scripts/ci/bundle_and_upload_rerun_cpp.py --local-path rerun_cpp_sdk.zip --no-upload --git-hash 0d23a5149c6552ed2e222da13905fc9db1644371 --platform-filter linux-x64 --skip-test
```
contains patch in buildzip! so far so good

Ran `cargo build -p rerun_c --release` to build mac bindings and inject them into the build zip as `librerun_c__macos_arm64.a`
bundled it up as a zip again and ran it against a checkout of the opencv example.

works! actually didn't work from pixi because of clang shenanigans but invoking cmake directly and making sure I have a sane clang verison. But that's a problem with that sample's pixi setup if at all, not with this PR.
Not 100% sure it applied the patch, but the patch was in there and all the code is in place to use it, so that should do it I'd think!

PR #11224, opened 2025-09-15
Update mcap snippet again

oxkitsune:
### What

Updates MCAP snippet backwards test.


oxkitsune:
@rerun-bot full-check

grtlr:
It's doubly annoying because GitHub won't let us check in the actual text file with real diffs so we have no idea what's going on without running locally.

PR #11225, opened 2025-09-15
Fix wrong rendering of some texture formats in light mode

emilk:
### Related
* Closes https://linear.app/rerun/issue/RR-2293/f64f16u64-images-dont-display-correctly

### What
When uploading RGB and BGR images to the GPU we would fill the alpha with a high value, e.g. u8. The GPU would interpret that as being opaque… mostly. It didn't work for all types, like `f16` where the alpha value wasn't obvious (is it 1.0 or 255.0?).

This PR clean this up, fixing a bug that would make some rare images (like BGR F16) sometime show up as _additive_, meaning they would be invisible in light mode.

Tested by clicking around the normal examples.

### DIFF
--- a/crates/viewer/re_renderer/src/renderer/rectangles.rs
+++ b/crates/viewer/re_renderer/src/renderer/rectangles.rs
@@ -89,7 +89,7 @@ pub struct ColormappedTexture {
     /// Only applies to [`wgpu::TextureFormat::Rgba8Unorm`] and float textures.
     pub decode_srgb: bool,
 
-    /// How to treat the alpha channel
+    /// How to treat the alpha channel.
     pub texture_alpha: TextureAlpha,
 
     /// Raise the normalized values to this power (before any color mapping).
--- a/crates/viewer/re_renderer/src/renderer/rectangles.rs
+++ b/crates/viewer/re_renderer/src/renderer/rectangles.rs
@@ -56,7 +56,7 @@ pub enum ShaderDecoding {
 }
 
 /// Describes a texture and how to map it to a color.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct ColormappedTexture {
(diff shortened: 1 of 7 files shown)


crates/viewer/re_renderer/src/renderer/rectangles.rs
```diff
@@ -72,11 +89,8 @@ pub struct ColormappedTexture {
     /// Only applies to [`wgpu::TextureFormat::Rgba8Unorm`] and float textures.
     pub decode_srgb: bool,
 
-    /// Multiply color channels with the alpha channel before filtering?
-    ///
-    /// Set this to false for textures that don't have an alpha channel or are already pre-multiplied.
-    /// Applied after range normalization and srgb decoding, before filtering.
-    pub multiply_rgb_with_alpha: bool,
+    /// How to treat the alpha channel
```
grtlr:
```suggestion
    /// How to treat the alpha channel.
```

crates/viewer/re_viewer_context/src/gpu_bridge/image_to_gpu.rs
```diff
@@ -675,29 +674,30 @@ fn narrow_f64_to_f32s(slice: &[f64]) -> Cow<'static, [u8]> {
...
 
-fn pad_and_narrow_and_cast<T: Copy + bytemuck::Pod>(
+fn pad_and_narrow_and_cast<T: Copy + bytemuck::Pod + Default>(
     data: &[T],
-    pad: f32,
     narrow: impl Fn(T) -> f32,
 ) -> Cow<'static, [u8]> {
     re_tracing::profile_function!();
 
+    let alpha = 0.0; // The shader should just ignore it
```
grtlr, line 695: Nit: Please update the comment in line 571. And, I think we can close the TODO in line 572.

PR #11226, opened 2025-09-15
Bump re_mp4 to 0.4.0

ntjohnson1:
### Related

https://github.com/rerun-io/re_mp4/pull/20

### What
For mp4 videos of 1 frame rerun fails to load them at all with a weird error about indexes needing to be the same length
   * I fixed something that caused us to skip frames when they are all the same size (not sure how common that is outside of single frame samples). Right now this points to that branch to demo nothing breaks. Will need to update to latest version once we update the crate.

Pointing at the same directory containing episode 1 and episode 58
Before (Where we both couldn't load episode 58 ever but also couldn't load it because our vector was of length 2):   
Screenshot 2025-09-15 at 5 47 46 PM

After (where things appear to work):
Screenshot 2025-09-15 at 5 52 05 PM

NOTE: I put a link to the re_mp4 branch for validation but I assume we will land that then update the version here.


Cargo.toml
```diff
@@ -822,3 +822,4 @@ unnecessary_debug_formatting = "allow"
 #walkers = { git = "https://github.com/rerun-io/walkers", branch = "antoine/egui-0.32" } # https://github.com/podusowski/walkers/pull/345
 
 # dav1d = { path = "/home/cmc/dev/rerun-io/rav1d", package = "re_rav1d", version = "0.1.1" }
+re_mp4 = { git = "https://github.com/rerun-io/re_mp4.git", branch = "nick/fix_constant_sixe" }
```
emilk: We can't ship crates with this, so let's merge the re_mp4 fix and release a patch release of re_mp4:
* https://github.com/rerun-io/re_mp4/pull/20

emilk (approved):

ntjohnson1:
> Before when I tried to open a partial LeRobot dataset (if I only had a subset of the contents listed in the metadata) we got all kinds of weird errors.
> * For this I made the episode collection a map instead of a vector which allows us to display an arbitrary subset of episodes from a dataset

I opened a PR for the non-mp4 changes.
https://github.com/rerun-io/rerun/pull/11228

ntjohnson1:
Ok I pruned all the other lerobot stuff that landed in a separate PR. I updated the description and moved this to draft. Once `re_mp4` 0.4.0 is released I'll update this to point there and re-open.

PR #11227, opened 2025-09-16
Support filtering by timestamp columns

abey79:
### Related

* closes https://linear.app/rerun/issue/RR-2389/timestamp-filtering-ui-finishing-touches

### What

Add the ability to filter over timestamp columns.

Currently supports:
- `rerun.components.Timestamp` columns
- arrow `Timestamp` columns (all representations: s, ms, us, and ns)


crates/viewer/re_dataframe_ui/src/filters/filter_ui.rs
```diff
@@ -121,18 +140,22 @@ impl FilterState {
...
+                        let filter_id = ui.make_persistent_id(
+                            egui::Id::new(index).with(&filter.column_name).with(
+                                match filter.operation {
+                                    FilterOperation::NonNullableBoolean(_) => "non_nullable_bool",
+                                    FilterOperation::NullableBoolean(_) => "nullable_bool",
+                                    FilterOperation::IntCompares { .. } => "int",
+                                    FilterOperation::FloatCompares { .. } => "float",
+                                    FilterOperation::StringContains(_) => "string",
+                                    FilterOperation::Timestamp(_) => "timestamp",
+                                },
```
Wumpf: why do we need the name _and_ the type of operation? Shouldn't the name be sufficient?
(at least right now you can't have a different type of operation on a given column)

crates/viewer/re_dataframe_ui/src/filters/timestamp_formatted.rs
```diff
@@ -0,0 +1,30 @@
+use re_log_types::TimestampFormat;
+
+/// Wrapper structure to associate a [`TimestampFormat`] with a reference.
+///
+/// This is used to implement [`re_ui::SyntaxHighlighting`] to types which may need to display
+/// timestamps.
+///
+/// Note: this can't be moved to `re_log_types` along with `TimestampFormat`, because then we cannot
+/// implement traits for it.
```
Wumpf, line 9: confused why we would want to move these to re_log_types

crates/viewer/re_dataframe_ui/src/filters/timestamp.rs
```diff
@@ -0,0 +1,990 @@
+//! Implement a filter for timestamp columns.
```
Wumpf, line 1: nice, thanks for that overview

crates/viewer/re_dataframe_ui/src/filters/timestamp.rs
```diff
@@ -0,0 +1,990 @@
...
+    LastWeek,
+    Before,
+    After,
+    Between,
+}
+
+/// A filter for [`arrow::datatypes::DataType::Timestamp`] columns.
+///
+/// This represents both the filter itself, and the state of the corresponding UI.
+//TODO(ab): a nicer `Debug` implementation would make snapshot tests cleaner.
```
Wumpf: so.. any reason why not now? ;)

Wumpf (approved):
Did some very light reviewing of the code after we already went into depth during the call yesterday. Looks good!

crates/viewer/re_dataframe_ui/src/filters/filter_ui.rs
abey79: you're right I think. removing.

crates/viewer/re_dataframe_ui/src/filters/timestamp_formatted.rs
abey79: Some other code might want to use a combo of syntax highlighting and `TimestampFormat` elsewhere. However, it's a tricky combo because of orphan rule etc. Here I suggest `re_log_types` because that's where `TimestampFormat` lives.

crates/viewer/re_dataframe_ui/src/filters/timestamp.rs
abey79: none that I can think of but laziness.... okay okay, on it :)

abey79: nice 
image

PR #11228, opened 2025-09-16
Add Ability to Visualize Partial LeRobot Dataset

ntjohnson1:
### Related
This is a subset of https://github.com/rerun-io/rerun/pull/11226 where I am not including the MP4 changes that need more discussion.

### What
Before when I tried to open a partial LeRobot dataset (if I only had a subset of the contents listed in the metadata) we got all kinds of weird errors.

I made the episode collection a map instead of a vector which allows us to display an arbitrary subset of episodes from a dataset.

### DIFF
--- a/crates/store/re_data_loader/src/lerobot.rs
+++ b/crates/store/re_data_loader/src/lerobot.rs
@@ -10,6 +10,7 @@
 //! See [`LeRobotDataset`] for more information on the dataset format.
 
 use std::borrow::Cow;
+use std::collections::BTreeMap;
 use std::fmt;
 use std::fs::File;
 use std::io::BufReader;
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
lgtm as pointed out on other PR. Since we're close to a release, I'd like @oxkitsune to briefly confirm this looks alright :)

crates/store/re_data_loader/src/lerobot.rs
```diff
@@ -232,10 +242,16 @@ impl LeRobotDatasetMetadata {
...
+        let episodes_vec: Vec<LeRobotDatasetEpisode> =
+            load_jsonl_file(metadir.join("episodes.jsonl"))?;
         let mut tasks = load_jsonl_file(metadir.join("tasks.jsonl"))?;
 
-        episodes.sort_by_key(|e: &LeRobotDatasetEpisode| e.index);
+        // Convert episodes vec to HashMap for efficient lookup by index
+        let episodes = episodes_vec
+            .into_iter()
+            .map(|episode| (episode.index, episode))
+            .collect::<HashMap<EpisodeIndex, LeRobotDatasetEpisode>>();
```
oxkitsune: IIUC, episodes are no longer sorted by index here right. 

We should at least go through them in order, so that in the viewer when you're loading the episodes they are loaded from lowest episode id to highest.

oxkitsune (approved):

PR #11230, opened 2025-09-16
Fix: Patch blueprint id when forced_application_id is set

oxkitsune:
### What

Patch the blueprint id in outgoing `LogMsg`s, if a `forced_application_id` is set. This was implemented for `LogMsg::ArrowMsg` and `LogMsg::SetStoreInfo` in #10742 but not for `LogMsg::BlueprintActivationCommand`.

### DIFF
--- a/crates/store/re_data_loader/src/loader_rrd.rs
+++ b/crates/store/re_data_loader/src/loader_rrd.rs
@@ -204,7 +204,20 @@ fn decode_and_stream<R: std::io::Read>(
                 }
 
                 re_log_types::LogMsg::BlueprintActivationCommand(blueprint_activation_command) => {
-                    re_log_types::LogMsg::BlueprintActivationCommand(blueprint_activation_command)
+                    let mut blueprint_id = blueprint_activation_command.blueprint_id.clone();
+                    if let Some(forced_application_id) = forced_application_id {
+                        blueprint_id =
+                            blueprint_id.with_application_id(forced_application_id.clone());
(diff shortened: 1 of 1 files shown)


crates/store/re_data_loader/src/loader_rrd.rs
```diff
@@ -204,7 +204,20 @@ fn decode_and_stream<R: std::io::Read>(
                 }
 
                 re_log_types::LogMsg::BlueprintActivationCommand(blueprint_activation_command) => {
-                    re_log_types::LogMsg::BlueprintActivationCommand(blueprint_activation_command)
+                    let mut blueprint_id = blueprint_activation_command.blueprint_id.clone();
+                    if let Some(forced_application_id) = forced_application_id {
+                        blueprint_id =
+                            blueprint_id.with_application_id(forced_application_id.clone());
+                    }
+                    if let Some(forced_recording_id) = forced_recording_id {
```
oxkitsune: Earlier in the file where this `decode_and_stream` function is called, there's a comment saying:
```
// We never want to patch blueprints' store IDs, only their app IDs.
```

and `forced_recording_id` will be set to `None`. Maybe it doesn't make sense to even attempt to patch the recording id.

abey79: Yeah, blueprint's recording id are always random, and we never want to "merge" blueprint, so it doesn't make sense to patch it.

PR #11233, opened 2025-09-16
Remove `ItemButton`s on favor of new `OnResponse` helper

lucasmerlin:
### Related

* follow up of https://github.com/rerun-io/rerun/pull/11164

### What

Adds a new `OnResponseExt` trait that allows you to call `on_click`, `on_menu`, `enabled` and much more on anything that is `impl Widget`.

We want to experiment with this in rerun and eventually move this to egui (if we decide that this is a good pattern).
One blocker for moving to egui would be that we can't customize the default menu_styling right now.


crates/viewer/re_ui/src/list_item/custom_content.rs
```diff
@@ -105,7 +106,15 @@ impl<'a> CustomContent<'a> {
...
-        self.button(super::ItemActionButton::new(icon, alt_text, on_click).enabled(enabled))
+        let alt_text = alt_text.into();
+        self.button(move |ui: &mut Ui| {
+            ui.add(
+                ui.small_icon_button_widget(icon, &alt_text)
+                    .on_click(on_click)
+                    .enabled(enabled)
+                    .on_hover_text(alt_text),
+            )
+        })
```
lucasmerlin, line 117: It'd be nice if small_icon_button was a freestanding function, then we could remove the extra closure.

crates/viewer/re_ui/src/boxed_widget.rs
```diff
@@ -0,0 +1,38 @@
+//! [`BoxedWidget`] and [`BoxedWidgetExt::boxed`] helpers.
```
lucasmerlin, line 1: These could also be very useful in egui.

crates/viewer/re_ui/src/widget_ext.rs
```diff
@@ -0,0 +1,187 @@
+use egui::{Popup, Response, Widget, WidgetText};
+
+pub struct OnResponse<'a, T> {
```
lucasmerlin, line 6: Not sure OnResponse is still the best name considering that it now also handles `enabled`. Maybe `WrappedWidget`?

crates/viewer/re_ui/src/egui_ext/boxed_widget.rs
```diff
@@ -0,0 +1,39 @@
+//! [`BoxedWidget`] and [`BoxedWidgetExt::boxed`] helpers.
+//!
+//! [`Box<dyn Widget>`] cannot work because dyn Widget is not `Sized`.
```
abey79: I don't understand this comment. `FnOnce` isn't `Sized` either. Do you mean that `Widget` is not dyn compatible (because of the `self` method)?

PR #11234, opened 2025-09-16
Add support for displaying timezone with our timestamps

abey79:
### Related

- part of https://linear.app/rerun/issue/RR-2306/table-filter-support-timestamp-columns

### What

This PR does the following:
- Introduce an option to display timestamp in local time but still displaying time zone (which makes displayed timestamp unambiguous).
- Improve the related UI in the settings window
- Propagate and use this setting all the way to `arrow_ui`

image

image

image

### DIFF
--- a/crates/viewer/re_arrow_ui/src/show_index.rs
+++ b/crates/viewer/re_arrow_ui/src/show_index.rs
@@ -4,18 +4,18 @@
 use std::ops::Range;
 
 use arrow::array::cast::{
-    as_generic_list_array, as_map_array, as_struct_array, as_union_array, AsArray as _,
+    AsArray as _, as_generic_list_array, as_map_array, as_struct_array, as_union_array,
 };
 use arrow::array::types::{
-    ArrowDictionaryKeyType, Float16Type, Float32Type, Float64Type, Int16Type, Int32Type, Int64Type,
-    Int8Type, RunEndIndexType, UInt16Type, UInt32Type, UInt64Type, UInt8Type,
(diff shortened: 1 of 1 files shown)


crates/viewer/re_arrow_ui/src/show_index.rs
```diff
@@ -199,6 +237,49 @@ fn make_ui<'a>(
...
+
+                Ok(())
+            }
+
+            fn array(&self) -> &dyn Array {
+                self.array
+            }
+        }
+    };
+}
```
lucasmerlin: Maybe you could instead update the existing show_primitive so it's ShowIndexState and holds displayoptions and then add a variant of the macro that gets passed the display options and formats the value based on that?

crates/viewer/re_arrow_ui/src/show_index.rs
```diff
@@ -159,13 +166,41 @@ fn make_ui<'a>(
...
+        | DataType::Duration(_) | DataType::Interval(_)
         | DataType::Decimal128(_, _) | DataType::Decimal256(_, _)
         => {
             show_arrow_builtin(array, options)
         }
+        DataType::Timestamp(TimeUnit::Second, _) => {
+            Ok(show_timestamp(array
+                .as_any()
+                .downcast_ref::<TimestampSecondArray>()
+                .expect("TimestampSecondArray downcast failed"), options.timestamp_format))
```
lucasmerlin: `array.as_primitive::<TimestampSecondType>()` should also work for all of these

crates/viewer/re_viewer/tests/snapshots/settings_screen.png
lucasmerlin, line 1: Nice!

PR #11235, opened 2025-09-16
Add copy link context menu to server entries

### LINKED ISSUE #10816: Add link copies to various context menus
See project details [Sharing Data by Links](https://linear.app/rerun/project/sharing-data-by-links-786c3809af14/overview)

See [https://www.figma.com/board/dsZcCUQxyrfj9YI8ltv3JY/share?node-id=85-1093&t=15DaAvP56ppiX3JH-4](https://www.figma.com/board/dsZcCUQxyrfj9YI8ltv3JY/share?node-id=85-1093&t=15DaAvP56ppiX3JH-4)

IsseW:
### Related

Closes #10816

### What

Adds context menu buttons to copy link to dataset entries and partitions in the server entries UI.

### DIFF
--- a/crates/viewer/re_test_context/src/lib.rs
+++ b/crates/viewer/re_test_context/src/lib.rs
@@ -435,7 +435,9 @@ impl TestContext {
                 render_ctx,
 
                 connection_registry: &self.connection_registry,
-                display_mode: &DisplayMode::LocalRecordings(self.active_store_id()),
+                display_mode: &DisplayMode::LocalRecordings(
+                    store_context.recording_store_id().clone(),
+                ),
             },
             component_ui_registry: &self.component_ui_registry,
             view_class_registry: &self.view_class_registry,


crates/viewer/re_global_context/src/lib.rs
```diff
@@ -68,7 +68,7 @@ pub enum DisplayMode {
     Settings,
 
     /// Regular view of the local recordings, including the current recording's viewport.
-    LocalRecordings,
+    LocalRecordings(Option<StoreId>),
```
abey79: What does `LocalRecordings(None)` mean? From the code I'm seeing below, absence of a current local recording is an error condition. So I really think should not be an option (but having the `StoreId` here is great).

crates/viewer/re_redap_browser/src/servers.rs
```diff
@@ -463,7 +463,7 @@ impl RedapServers {
             viewer_ctx
                 .command_sender()
                 .send_system(SystemCommand::ChangeDisplayMode(
-                    DisplayMode::LocalRecordings,
+                    DisplayMode::LocalRecordings(None),
                 ));
```
abey79: The situation here is "I was supposed to display some server stuff here, but I don't have any, so please `re_viewer` display something else". It's kind of weird to begin with and should generally never happen, but who knows, so we must handle it.

Now, it's absolutely not `re_redap_browser`'s responsibility  to figure out _what_ to actually display. Having this hard-coded here was wrong to begin with imo.

My suggestion would be to have a `ViewerContext::revert_to_default_display_mode` helper which would figure out whether there exist a "current local recording" (and if so use it), defaulting to whatever we show on startup (e.g. welcome screen if not disabled).

crates/viewer/re_test_context/src/lib.rs
```diff
@@ -435,7 +435,7 @@ impl TestContext {
                 render_ctx,
 
                 connection_registry: &self.connection_registry,
-                display_mode: &DisplayMode::LocalRecordings,
+                display_mode: &DisplayMode::LocalRecordings(None),
```
abey79: We populate the store of the test context with some dummy recording. You can use its `StoreId` here.

abey79 (changes requested):
This generally looks great, but I'm going to push back against that `Option` in `DisplayMode::LocalRecording`. This `DisplayMode` stuff is a tool in an on-going effort to address the sorry state of the top-level viewer implementation (`AppState::show()` etc.). As such, I'm insistent we remain principled and not let any form of (further) debt creep in.

I do love adding a (non-`Option`) `StoreId` in `DisplayMode::LocalRecording`. This is a step is the right direction 👍🏻 

I've tried to leave breadcrumbs as to how you may fix this. Chances it'll end up being trickier. Don't hesitate to ping me if you need help!

crates/viewer/re_global_context/src/lib.rs
IsseW: Yeahh, it was mostly to avoid having to do some more refactoring but happy to make it only `StoreId` as that makes sense!

abey79:
While trying out stuff, I found a display logic bug. Fixing it is not necessarily in scope for this PR (also, it's kind of niche), but chances are it lurks not very far from it.

- https://github.com/rerun-io/rerun/issues/11243

IsseW:
> While trying out stuff, I found a display logic bug. Fixing it is not necessarily in scope for this PR (also, it's kind of niche), but chances are it lurks not very far from it.

With my newest changes I can't repro this unless I missunderstood, so might've fixed it?

abey79:
> > While trying out stuff, I found a display logic bug. Fixing it is not necessarily in scope for this PR (also, it's kind of niche), but chances are it lurks not very far from it.
> 
> With my newest changes I can't repro this unless I missunderstood, so might've fixed it?

I'll try it again after you merge this.

crates/viewer/re_test_context/src/lib.rs
IsseW: I think that using this recording is causing the tests to freeze for some reason

EDIT: found why 🤦‍♀️

PR #11238, opened 2025-09-16
Fix doc index build

Wumpf:
We were running with an old nightly version, causing trouble running against our current workspace rerun version. See job failure https://github.com/rerun-io/rerun/actions/runs/17772184999/job/50510938352

@jprochazk remind me, why do we have to use nightly on these? That makes this very easy to miss on rust updates 😬

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -54,6 +54,7 @@ deny = [
 skip = [
   { name = "base64" },           # Too popular
   { name = "block2" },           # Old version via rfd
+  { name = "cargo-platform" },   # Older version used by ply-rs. It's build-time only!
   { name = "cargo_metadata" },   # Older version used by ply-rs. It's small, and it's build-time only!
   { name = "core-foundation" },  # Currently, e.g. `webbrowser` and `winit` use different versions.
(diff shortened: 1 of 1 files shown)


jprochazk:
> why do we have to use nightly on these?

Because rustdoc json format is unstable.

.github/workflows/on_push_docs.yml
```diff
@@ -44,7 +44,7 @@ jobs:
     runs-on: ubuntu-latest-16-cores
     strategy:
       matrix:
-        toolchain: ["nightly-2025-02-05"]
+        toolchain: ["nightly-2025-09-16"]
```
jprochazk, line 48: Could you add a comment to `rust-toolchain` to also bump this version when bumping to new stable?

jprochazk (approved):

.github/workflows/on_push_docs.yml
jprochazk: Also, the crates associated with rustdoc (-json and -types) will probably have to be bumped to new versions. I don't know if that will cause any real failures.

This is unfortunately still the best way to collect items + docs for each crate programmatically.

PR #11241, opened 2025-09-16
Fix title bar height on macOS Tahoe

emilk:
Before:
image

After:
image


IsseW (approved):
Fix makes sense! I assume this won't break for older versions of mac, tested on Sequoia

Issue #10815, opened 2025-08-06
All URL components should have a copy button


Wumpf:
In particular links that are presented with `Open` should still have it

For details see
https://www.figma.com/board/dsZcCUQxyrfj9YI8ltv3JY/share?node-id=85-1032&t=15DaAvP56ppiX3JH-4

PR #11242, opened 2025-09-17
Add copy button to the recording link table item

IsseW:
### Related

Part of #10815

### What

Adds a copy button that's shown on hover in table url entries
image

- [x] look into if it's possible to not have to define the copy button rect

### DIFF
--- a/crates/viewer/re_component_ui/src/variant_uis/redap_uri_button.rs
+++ b/crates/viewer/re_component_ui/src/variant_uis/redap_uri_button.rs
@@ -3,9 +3,11 @@ use std::str::FromStr as _;
 
 use egui::{Align, Layout, Link, Ui, UiBuilder};
 use re_types_core::{ComponentDescriptor, RowId};
-use re_ui::UiExt as _;
+use re_ui::UiExt;
 use re_uri::RedapUri;
-use re_viewer_context::{SystemCommand, SystemCommandSender as _, ViewerContext};
+use re_viewer_context::{
+    SystemCommand, SystemCommandSender as _, ViewerContext, open_url::ViewerOpenUrl,
+};
 
 /// Display an URL as an `Open` button (instead of spelling the full URL).
 ///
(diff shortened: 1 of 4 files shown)


crates/viewer/re_component_ui/src/variant_uis/redap_uri_button.rs
```diff
@@ -52,16 +54,45 @@ pub fn redap_uri_button(
...
     // Show the link left aligned and justified, so the whole cell is clickable.
     let put_justified_left_aligned = |ui: &mut Ui, link| {
         ui.scope_builder(
             UiBuilder::new().max_rect(ui.max_rect()).layout(
-                Layout::left_to_right(Align::Center)
-                    .with_main_justify(true)
+                Layout::right_to_left(Align::Center)
                     .with_cross_justify(true)
                     .with_main_align(Align::Min),
             ),
```
abey79: I believe you could use `egui::Sides` for this. See e.g. how the table headers (which have a right aligned menu button) is implemented, in `header_cell_ui` (`datafusion_table_widget.rs`)

IsseW: ahh thanks! I felt like there had to be some better way to do it, still a bit finicky to get the whole cell clickable, but managed to do it :)

IsseW:
> Nit: the button isn't vertically centred. You may need to have some `Align::Center` there too.

Good catch, wish there was some less ugly way to do this, but it seems `Sides` for some reason has a smaller y range than the parent UI.

Actually maybe I can use the Sides::height function here

IsseW:
In the design I also noticed that this is something we want for a lot of stuff, so might be nice to be able to extract adding a copy button that's only visible on hover into some util function, but can probably be left for the future.

abey79:
> In the design I also noticed that this is something we want for a lot of stuff, so might be nice to be able to extract adding a copy button that's only visible on hover into some util function, but can probably be left for the future.

Yes, that's definitely something we want in the tables in general. In particular for data that is formatted for display (e.g. thousand separators, etc.). In this case, the copy button should copy the unformatted data represnetation.

Issue #11245, opened 2025-09-17
Component defaults menu is broken


Wumpf:
Can't edit anything there at alll anymore.

https://github.com/user-attachments/assets/40fa0913-2006-45a9-a332-edce956d4cf9

Also:
* buttons shouldn't show up just on hover but instead always be there
* rendering of archetypes in list shows full archetype names, should show "friendly" names instead

PR #11247, opened 2025-09-17
Fix some menu buttons closing unexpectedly

lucasmerlin:
### Related

* Closes https://github.com/rerun-io/rerun/issues/11245

### What

Changes the default for `should_always_show_buttons` to true when adding a menu button. 
Maybe we should change the default to be always true though?
I feel like usually you want to always show buttons, unless every list item has it (e.g. visibility in the blueprint tree) where it would add too much noise / use too much space.

When I worked on the new ItemButtons, I defaulted to false since that was the default in LabelContent.

### DIFF
--- a/crates/viewer/re_ui/src/list_item/item_buttons.rs
+++ b/crates/viewer/re_ui/src/list_item/item_buttons.rs
@@ -196,6 +196,8 @@ where
     /// The `alt_text` will also be used for the tooltip.
     ///
     /// See [`Self::with_button`] for more information.
+    ///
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
> Maybe we should change the default to be always true though?
> I feel like usually you want to always show buttons, unless every list item has it (e.g. visibility in the blueprint tree) where it would add too much noise / use too much space.

agreed! but it's probably also good to keep this iteration more minimal? 🤔 

a bit shocking that none of our image comparison tests triggers on this change. Strange!

Can be a separate PR, but it would be nice to also fix the formatting here:
image
(this was called out in the same issue, but more as a sidenote)

Issue #11244, opened 2025-09-17
Space origin editing in viewer is broken


Wumpf:
https://github.com/user-attachments/assets/89ce4f47-1904-4ef3-8720-2bb42e005bee

Repros in `nuscenes` example, haven't tried if it is an issue elsewhere yet.

PR #11251, opened 2025-09-17
Fix space origin editor resetting & closing immediately

lucasmerlin:
### Related
* close https://github.com/rerun-io/rerun/issues/11244

### What

Fixes the space origin editor. It will now always switch to editing state if focused, before, even without breaking this would have been weird and buggy if clicking the text again. Still not ideal but good enough for now. Maybe eventually we could make a generic autocomplete widget with better focus handling.


Wumpf (approved):
fix works nicely, thanks

Still feels fairly broken with this meaningless "hidden suggestion".. like c'mon what are you hiding?!  But that was surely pre-existing, just happy I can type things in there again
image

PR #11252, opened 2025-09-17
Don't go back to default screen when selecting something

IsseW:
### Related

Fixes regression from #11235

### What

Fixes regression that made selections go back to start screen


oxkitsune (approved):
Tested, works and wasn't able to break anything else :shipit:

PR #11253, opened 2025-09-17
Use egui from main

lucasmerlin:
### What

Updates egui to latest main.

I also had to update `notify` and `directories` to fix some cargo-deny errors, since they duplicated `windows-sys`.

### TODO
* [x] Merge in latest `main`
* [x] Run full check


crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -683,7 +683,11 @@ impl AppState {
             .memory(|mem| mem.focused())
             .and_then(|id| TextEditState::load(ui.ctx(), id))
             .is_none()
-            && !LabelSelectionState::load(ui.ctx()).has_selection()
+            && ui
+                .ctx()
+                .plugin::<LabelSelectionState>()
+                .lock()
+                .has_selection()
```
lucasmerlin, line 694: I don't really like this. Ideally this would be instead: `ui.plugin::<LabelSelectionState>().has_selection()`

I think in most cases you want to access the plugin right then and there so it should return the guard by default. Then for edge cases like cloning the plugin handle to send it to some thread or something we could add a `ctx.plugin_handle()` fn.

crates/viewer/re_renderer/src/renderer/lines.rs
```diff
@@ -131,7 +131,8 @@ pub mod gpu_data {
     #[repr(C, packed)]
     #[derive(Clone, Copy, bytemuck::Pod, bytemuck::Zeroable)]
     pub struct LineStripInfo {
-        pub color: Color32, // alpha unused right now
+        /// [`ecolor::Color32`] is `repr(align(4))` so we can't use it in `repr(packed)`.
+        pub color: [u8; 4], // alpha unused right now
```
lucasmerlin: This was also interesting to me, is this fine or should we revert the repr(align)?

abey79: My personal opinion is that you want control for these type of structure, so not using 3rd party types makes sense. But maybe double check with @Wumpf ?

abey79 (approved):
lgtm.

Wumpf:
> I also did a cargo update.

I think we should do that every once in a while, but @emilk argued against it in the past because it causes a lot of unnecessary churn and potential unexpected breakage (but I figure anyone doing `cargo install rerun` has the same experience)

Either way would be good to do it separately

crates/viewer/re_renderer/src/renderer/lines.rs
Wumpf: It's a bit unfortunate to not have this typed here because picking the wrong type of color is just so awfully error prone. So I think re_renderer should introduce a `UnalignedColor32` for this purpose 🤔

repr packed is used here to make sure that there's no unexpected padding between the 8bit fields below

Wumpf: ironically we really do want this to be align(4) still here 😅

crates/viewer/re_viewer/src/app_state.rs
emilk: `ui.ctx().plugin::<LabelSelectionState>(|p| p.has_selection())` would be more in line with the rest of egui

crates/viewer/re_renderer/src/renderer/lines.rs
lucasmerlin: I've added a `UnalignedColor32` struct

Cargo.lock
```diff
@@ -8353,7 +8390,7 @@ dependencies = [
  "wasm-bindgen-futures",
  "web-sys",
  "web-time",
- "wgpu",
+ "wgpu 25.0.2",
```
emilk: looks like we're doubling up on wgpu

lucasmerlin: Yup I went back to a commit before the egui wgpu update was merged, I'll do that in a separate PR

lucasmerlin:
@rerun-bot full-check

PR #11267, opened 2025-09-18
Fix `to_arrow_reader` causing an error for streaming providers

timsaucer:
### Related

The `DataFrameQueryView::to_arrow_reader` was broken after switching to a non-collecting table provider. This was discovered during release testing.

### What

This PR fixes two issues:

- `PartitionStreamExec` could not be copied using `with_children` because it would error even if the children vec was empty.
- `to_arrow_reader` was using a blocking call on the python tokio runtime, but we cannot spawn our additional streams in this way. As an overall improvement we can instead use the fact that DataFusion DataFrames already support the arrow C interface for streaming. Thus we can use `pyarrow.RecordBatchReader.from_stream` to turn a dataframe directly into a streaming provider. This also has the advantage of removing the collect operation, so it is not memory bound.


rerun_py/src/catalog/dataframe_query.rs
```diff
@@ -402,28 +402,12 @@ impl PyDataframeQueryView {
...
-
-        let reader = RecordBatchIterator::new(record_batches.into_iter().map(Result::Ok), schema);
-
-        Ok(PyArrowType(Box::new(reader)))
+    ) -> PyResult<Bound<'py, PyAny>> {
+        let df = Self::df(self_)?;
+
+        py.import("pyarrow")?
+            .getattr("RecordBatchReader")?
+            .call_method1("from_stream", (df,))
```
timsaucer, line 408: I would have preferred to do this in python but since we don't have wrapper classes and instead use `.pyi` files I need to call it from rust.

timsaucer:
```
def to_arrow_reader(self):
    return pa.RecordBatchReader.from_stream(self.df())
```

abey79: FWIW, the current political situation regarding pure python wrappers ranges from "very welcome" to "tolerated", depending who you talk to. After experimenting with `class CatalogClient`, I'm very much in the former's camp.

rerun_py/tests/unit/test_datafusion_tables.py
```diff
@@ -103,7 +103,7 @@ def is_port_open() -> bool:
 
 
 @pytest.fixture(scope="module")
-def server_instance() -> Generator[tuple[subprocess.Popen[str], DatasetEntry], None, None]:
+def server_instance() -> Generator[tuple[subprocess.Popen[str], CatalogClient, DatasetEntry], None, None]:
```
abey79: Might be time to roll a small @dataclass in place of this tuple 😅

rerun_py/tests/unit/test_datafusion_tables.py
```diff
@@ -234,10 +234,48 @@ def test_partition_ordering(server_instance: tuple[subprocess.Popen[str], Datase
                         prior_timestamp = timestamp
 
 
-def test_url_generation(server_instance: tuple[subprocess.Popen[str], DatasetEntry]) -> None:
+def test_arrow_rb_reader(server_instance: tuple[subprocess.Popen[str], CatalogClient, DatasetEntry]) -> None:
```
abey79: let maximise the legibility of that when it actually fails

```suggestion
def test_view_to_arrow_reader(server_instance: tuple[subprocess.Popen[str], CatalogClient, DatasetEntry]) -> None:
```

abey79 (approved):
Looks great, thanks!

rerun_py/tests/unit/test_datafusion_tables.py
timsaucer: This should test view, table, and table entry, but I will make it more descriptive.

PR #11268, opened 2025-09-18
Add module definition to all `pyclasses`

ntjohnson1:
### What
Without specifically listing module all of our classes show as if they are from `builtins`

Before
```python
from rerun import ComponentDescriptor
type(ComponentDescriptor("foo"))
<class 'rerun_bindings.rerun_bindings.ComponentDescriptor'>
```

After
```python
from rerun import ComponentDescriptor
type(ComponentDescriptor("foo"))
<class 'builtins.ComponentDescriptor'>
```

This also seems to resolve the issue with `pixi run py-docs-serve` not sure if pyo3 or griffe or some other thing in the toolchain updated which broke things since this worked on main 2 weeks ago.

### DIFF
--- a/rerun_py/src/dataframe/component_columns.rs
+++ b/rerun_py/src/dataframe/component_columns.rs
@@ -9,7 +9,11 @@ use re_sorbet::{ComponentColumnDescriptor, ComponentColumnSelector};
 /// Column descriptors are used to describe the columns in a
 /// [`Schema`][rerun.dataframe.Schema]. They are read-only. To select a component
 /// column, use [`ComponentColumnSelector`][rerun.dataframe.ComponentColumnSelector].
-#[pyclass(frozen, name = "ComponentColumnDescriptor")]
+#[pyclass(
+    frozen,
+    name = "ComponentColumnDescriptor",
+    module = "rerun_bindings.rerun_bindings"
+)]
 #[derive(Clone)]
 pub struct PyComponentColumnDescriptor(pub ComponentColumnDescriptor);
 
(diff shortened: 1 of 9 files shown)


rerun_py/src/dataframe/component_columns.rs
```diff
@@ -9,7 +9,11 @@ use re_sorbet::{ComponentColumnDescriptor, ComponentColumnSelector};
 /// Column descriptors are used to describe the columns in a
 /// [`Schema`][rerun.dataframe.Schema]. They are read-only. To select a component
 /// column, use [`ComponentColumnSelector`][rerun.dataframe.ComponentColumnSelector].
-#[pyclass(frozen, name = "ComponentColumnDescriptor")]
+#[pyclass(
+    frozen,
+    name = "ComponentColumnDescriptor",
+    module = "rerun_bindings.rerun_bindings"
```
oxkitsune, line 15: Since `__init__.py` in `rerun_bindings` re-exports everything in the `rerun_bindings.rerun_bindings` module, can we not drop one of these `rerun_bindings`?

ntjohnson1: I don't think that follows the general approach. As an example we flatten most of our namespaces around archetypes but the type is where it actually lives. This example comes from how pure python handles naming around modules.

```python
>>> import rerun as rr
>>> value = rr.Scalars(1)
>>> print(value)
  rr.Scalars(
    scalars=[1.0]
  )
>>> type(rr.Scalars(1))
<class 'rerun.archetypes.scalars.Scalars'>
```

ntjohnson1: Ya removing the one level of nesting I get

```
Could not resolve alias rerun._baseclasses.ComponentDescriptor pointing at rerun_bindings.ComponentDescriptor (in rerun_py/rerun_sdk/rerun/_baseclasses.py:11)
```

There might be a setting for griffe to resolve this but I'd prefer to land the improvement.

PR #11270, opened 2025-09-19
Show loading screen when opening link

IsseW:
### Related

Closes RR-2325

### What

Adds a loading screen when opening links, and improves the existing table loading screen by using the same UI.

### DIFF
--- a/crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
+++ b/crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
@@ -201,18 +201,6 @@ impl<'a> DataFusionTableWidget<'a> {
         self
     }
 
-    fn loading_ui(ui: &mut egui::Ui) -> egui::Response {
-        Frame::new()
-            .inner_margin(16.0)
-            .show(ui, |ui| {
-                ui.horizontal(|ui| {
-                    ui.spinner();
-                    ui.label("Loading table…");
-                });
-            })
-            .response
-    }
-
     /// Display the table.
     pub fn show(
         self,
@@ -232,13 +220,11 @@ impl<'a> DataFusionTableWidget<'a> {
         match session_ctx.table_exist(table_ref.clone()) {
             Ok(true) => {}
             Ok(false) => {
-                Self::loading_ui(ui).on_hover_text("Waiting…");
+                ui.loading_screen("Loading table...");
                 return;
             }
             Err(err) => {
-                Self::loading_ui(ui).on_hover_ui(|ui| {
-                    ui.label(err.to_string());
-                });
+                ui.loading_screen(&format!("Error while loading table:\n{err}"));
                 return;
             }
         }
@@ -291,7 +277,7 @@ impl<'a> DataFusionTableWidget<'a> {
                 // still processing, nothing yet to show
                 //TODO(ab): it can happen that we're stuck in the state. We should detect it and
                 //produce an error
-                Self::loading_ui(ui);
+                ui.loading_screen("Loading table...");
                 return;
             }
         };
--- a/crates/viewer/re_global_context/src/lib.rs
+++ b/crates/viewer/re_global_context/src/lib.rs
@@ -67,6 +67,8 @@ pub enum DisplayMode {
     /// The settings dialog for application-wide configuration.
     Settings,
 
+    Loading(String),
+
     /// Regular view of the local recordings, including the current recording's viewport.
     LocalRecordings(StoreId),
 
--- a/crates/viewer/re_viewer_context/src/open_url.rs
+++ b/crates/viewer/re_viewer_context/src/open_url.rs
@@ -192,6 +192,14 @@ pub fn base_url(url: &Url) -> Url {
     base_url
 }
 
+#[derive(Clone, Copy, Default)]
+pub struct OpenUrlOptions {
+    pub follow_if_http: bool,
+    pub select_redap_source_when_loaded: bool,
+    /// Shows the loading screen.
+    pub show_loader: bool,
+}
+
 impl ViewerOpenUrl {
     pub fn from_context(ctx: &ViewerContext<'_>) -> anyhow::Result<Self> {
         let time_ctrl = ctx.rec_cfg.time_ctrl.read();
@@ -255,6 +263,8 @@ impl ViewerOpenUrl {
                 // Not much point in updating address for the settings screen.
                 Err(anyhow::anyhow!("Can't share links to the settings screen."))
             }
+
+            DisplayMode::Loading(source) => source.parse(),
 
             DisplayMode::LocalRecordings(store_id) => {
                 // Local recordings includes those downloaded from rrd urls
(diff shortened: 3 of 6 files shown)


crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -232,13 +220,14 @@ impl<'a> DataFusionTableWidget<'a> {
...
+                    "Loading table:",
+                    url.as_deref().or(title.as_deref()).unwrap_or(""),
+                );
                 return;
             }
             Err(err) => {
-                Self::loading_ui(ui).on_hover_ui(|ui| {
-                    ui.label(err.to_string());
-                });
+                ui.loading_screen("Error while loading table:", &err.to_string());
```
Wumpf: wondering if this should be styled with the error message style 🤔

crates/viewer/re_global_context/src/lib.rs
```diff
@@ -67,6 +67,11 @@ pub enum DisplayMode {
     /// The settings dialog for application-wide configuration.
     Settings,
 
+    // TODO(isse): Use this to know what to switch to when sources are
+    // done loading.
```
Wumpf: not sure I parse the intention correctly, you mean this state should know its destination in the future, i.e. something like this?
```
Loading {
    message: String,
    target: Box<DisplayMode>,
}
```
(would make sense to me :))

crates/viewer/re_recording_panel/src/recording_panel_ui.rs
```diff
@@ -569,23 +569,32 @@ fn receiver_ui(
...
-    let Some(string) = receiver.loading_string() else {
+    let Some(string) = receiver.loading_name() else {
         return;
     };
 
-    let label_content = re_ui::list_item::LabelContent::new(string).with_buttons(|ui| {
-        let resp = ui
-            .small_icon_button(&re_ui::icons::REMOVE, "Disconnect")
-            .on_hover_text("Disconnect from this source");
+    let label_content = re_ui::list_item::LabelContent::new(string)
```
Wumpf, line 578: great love to detail in here <3

crates/viewer/re_viewer_context/src/open_url.rs
```diff
@@ -192,6 +192,15 @@ pub fn base_url(url: &Url) -> Url {
     base_url
 }
 
+#[derive(Clone, Copy, Default)]
+pub struct OpenUrlOptions {
```
Wumpf, line 196: good call splitting this out! 👍

crates/viewer/re_viewer_context/src/open_url.rs
```diff
@@ -238,6 +247,52 @@ impl ViewerOpenUrl {
         Ok(this)
     }
 
+    pub fn from_data_source(data_source: &SmartChannelSource) -> anyhow::Result<Self> {
```
Wumpf, line 253: could use a doc string, can refer to `from_display_mode`

crates/viewer/re_viewer_context/src/open_url.rs
```diff
@@ -436,12 +449,17 @@ impl ViewerOpenUrl {
...
-        select_redap_source_when_loaded: bool,
+        options: &OpenUrlOptions,
         command_sender: &CommandSender,
     ) {
         re_log::debug!("Opening URL: {:?}", &self);
 
+        if options.show_loader
+            && let Ok(url) = self.sharable_url(None)
+        {
+            command_sender.send_system(SystemCommand::ChangeDisplayMode(DisplayMode::Loading(url)));
```
Wumpf: interesting that nowhere in the `match` below we change the display mode directly, huh.
I think it would be best to put this last just in case, but I'm not 100% sure: if any of the other commands _also_ changes the display mode directly (not by enqueuing another command but by direct manipulation), then this would ofc break down again 😵‍💫 
Maybe best to leave it as is (seems to work fine!) and leave a comment instead that anything below is not expected to change display mode?

crates/viewer/re_global_context/src/lib.rs
Wumpf: reading on, I realize that this string isn't actually a message but rather the target URL :O
At the very least it should be documented! Ideally we'd be doing this typed, but I figure we don't have the right type here (and decoupling is worth _something_ ;))

crates/viewer/re_viewer_context/src/viewer_context.rs
```diff
@@ -163,6 +163,43 @@ impl ViewerContext<'_> {
         self.selection_state.selected_items()
     }
 
+    /// Returns if this item should be displayed as selected or not.
+    ///
+    /// This does not always line up with [`Self::selection`], if we
+    /// are currently loading something that will be prioritized here.
+    pub fn is_selected(&self, item: &Item) -> bool {
```
Wumpf: The semantics are a bit scary. Is calling this `is_selected_or_loading` too much of a stretch? 🤔 

Diverging the selection state from what is shown as selection sounds like asking for lotsa future trouble - why isn't the thing then somehow pinned on the selection state :/
I believe we should try to either sharpen semantics or find a better solution

Wumpf (changes requested):
love it, great look & feel!

But let's try to figure out the selection kerfuffle a bit better; makes me worried this becomes too complicated in the future as-is

crates/viewer/re_viewer_context/src/viewer_context.rs
IsseW: > why isn't the thing then somehow pinned on the selection state

I thiink we can't do that right now because the thing we're loading hasn't become an item that can be reffered to yet?

IsseW: Renaming this sounds good unless I find some better way to express loading items

crates/viewer/re_global_context/src/lib.rs
IsseW: Yeah this is even further down than `re_viewer_context`, should definitely add documentation.

> you mean this state should know its destination in the future

Yeah, so we only switch to the thing we've loaded if we're still on that page. And not if they have pressed to go to some other recording etc.

The problem with `Box<DisplayMode>` is I think again that the thing we're loading hasn't gotten a `StoreId` yet. But maybe it would be possible to change the order around to have that ahead of time?

Wumpf: > The problem with Box<DisplayMode> is I think again that the thing we're loading hasn't gotten a StoreId yet. But maybe it would be possible to change the order around to have that ahead of time?

actually making this happen might be a pretty deep issue. I believe we could make this happen for the server connection paths but for any others it might not be possible 🤔 
e.g. loading an rrd from http directly means we know only the url but not what store id that gonna lead to :/

crates/viewer/re_viewer_context/src/open_url.rs
IsseW: I think in some cases we do, for example `RedapEntry`. But those handle their own loading screens internally, which is a bit messy tbh.

And if something immediately changes display mode that's probably intended, and we don't want a loading screen in those cases.

crates/viewer/re_global_context/src/lib.rs
Wumpf: actually just learned that store ids are now derived from URLs `DatasetPartitionUri::store_id`
don't think that's actually for this PR, but related to the question :)

emilk:
For how long is the loading screen shown?
Can I abort the loading if it gets stuck?
Is it full-screen?

IsseW:
> For how long is the loading screen shown?

Until the first data from the source is received. [Arguably](https://rerunio.slack.com/archives/C08TEFXNHB2/p1758292637507359?thread_ts=1758267223.120069&cid=C08TEFXNHB2) it should be longer? But not something that should be done in this PR imo.

> Is it full-screen?

No you still see the side-panel with list of recordings that you can switch to.

> Can I abort the loading if it gets stuck?

Yeah, you can both switch to other things in the side-panel, and cancel the loading by closing the recording there. Although did find an issue making sure this was the case 😄

crates/viewer/re_viewer_context/src/viewer_context.rs
IsseW: Did the renaming

crates/viewer/re_global_context/src/lib.rs
Wumpf: great idea with the channel source!

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
IsseW: Not sure exactly what you mean with this? I tried the `error_label`, but that had lots of aligning issues :(

IsseW: Made the error text use the error fg color at least

PR #11281, opened 2025-09-19
Better warning for overlapping/out of order videos

IsseW:
### Related

* Part of #10843

### What

Adds better warning for when video streams are logged out of order.

### DIFF
--- a/crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
+++ b/crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
@@ -304,10 +304,6 @@ fn read_samples_from_chunk(
         return Ok(());
     };
 
-    let min_presentation_timestamp = samples
-        .front()
-        .map_or(re_video::Time::MIN, |s| s.presentation_timestamp);
-
     let mut previous_max_presentation_timestamp = samples
         .back()
         .map_or(re_video::Time::MIN, |s| s.presentation_timestamp);
@@ -320,6 +316,10 @@ fn read_samples_from_chunk(
         .and_then(|time_range| time_range.get(&sample_descr))
     {
         Some(time_range) => {
+            let min_presentation_timestamp = samples
+                .front()
+                .map_or(re_video::Time::MIN, |s| s.presentation_timestamp);
+
             // If the whole chunk is before the current stream time range, insert
             // it at the start.
             if time_range.max().as_i64() < min_presentation_timestamp.0 {


crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -316,9 +320,22 @@ fn read_samples_from_chunk(
         .and_then(|time_range| time_range.get(&sample_descr))
     {
         Some(time_range) => {
-            if time_range.min().as_i64() < previous_max_presentation_timestamp.0 {
+            // If the whole chunk is before the current stream time range, insert
+            // it at the start.
+            if time_range.max().as_i64() < min_presentation_timestamp.0 {
+                re_log::warn_once!(
```
IsseW, line 326: I think it would be ideal if these warns were "once per recording". Guess I could include the current recording in the warn somehow?

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -304,6 +304,10 @@ fn read_samples_from_chunk(
         return Ok(());
     };
 
+    let min_presentation_timestamp = samples
+        .front()
+        .map_or(re_video::Time::MIN, |s| s.presentation_timestamp);
```
aedm: Nitpick: this could go inside the `Some(time_range) => { ... }` block.

aedm (approved):
Better error messages! :)

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
Wumpf: agreed, but don't know how to do this atm without either ugly code or ugly id inside the error strong :/

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -316,9 +316,26 @@ fn read_samples_from_chunk(
...
+                     an earlier point in time. Or two recordings have been \
+                     merged where the trailing recording was at an earlier \
+                     point in time than the later one."
+                );
+
+                return Ok(());
+            } else if time_range.min().as_i64() < previous_max_presentation_timestamp.0 {
                 re_log::warn_once!(
-                    "Out of order logging on video streams is not supported. Ignoring any out of order samples."
+                    "Overlapping video streams are not supported, some logged streams have been ignored."
```
Wumpf, line 338:
```suggestion
                    "Late insertion of video frames within an established video stream is not supported. Some video data has been ignored."
```

That's the other more common way to get there I think: you log a bit of video and then later insert frames within the previously established range.
Also, I hope this sounds now less like we don't support multiple video streams (which we ofc do if they're on different entities). Better formulation still welcome

IsseW:
#11307 precedes this

Issue #11294, opened 2025-09-22
Remove the `blocking` argument of `def flush(…)`


emilk:
Kept for backwards compatibility, but is not necessary if we use the new `timeout_sec` instead

PR #11293, opened 2025-09-22
Improve migration guide about `connect_grpc`

emilk:
* Guide originally added in https://github.com/rerun-io/rerun/pull/11008

PR #11008, opened 2025-08-26
Add explicit timeout to flush API

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/9845

### What
This PR rewrites our flushing logic to be more robust and configurable.

When flushing a recording stream or a sink, you can now configure the timeout. The default timeout is ∞.

### TODO
* [ ] Add notice in migration guide
* [ ] `@rerun-bot full-check`

PR #11295, opened 2025-09-22
Python SDK: Add `timeout_sec` argument to `flush`

emilk:
### Related
* Somehow forgotten in https://github.com/rerun-io/rerun/pull/11008
* Related: https://github.com/rerun-io/rerun/pull/11293
* Follow-up: https://github.com/rerun-io/rerun/issues/11294

### What
This adds `timeout_sec` argument to the `flush` function in Python.

The change is backwards compatible, leaving the (now superfluous) `blocking` argument there too.


emilk:
@rerun-bot full-check

PR #11296, opened 2025-09-22
Add optional viewer url parameter to web app options

### LINKED ISSUE #11261: Share button doesn't make any sense on our documentation
Image

It works as the author intended: it gives you the base url + the rrd. Which .. doesn't work obviously:
https://rerun.io/docs/reference/types/archetypes/capsules3d?url=https%3A%2F%2Fapp.rerun.io%2Fversion%2F0.25.0%2Fexamples%2Fsnippets%2Fcapsules3d_batch.rrd

Not entirely sure what to do about this: likely when running off the js sdk we should opt-in to providing base URLs?
Fairly sure @jprochazk already pointed that problem out ahead of time and I forgot about it...

IsseW:
### Related

* Part of #11261

### What

Adds an optional field to the web viewer's app options for a viewer url that's used when sharing an url

### DIFF
--- a/crates/viewer/re_viewer/src/web.rs
+++ b/crates/viewer/re_viewer/src/web.rs
@@ -715,6 +715,10 @@ fn create_app(
 ) -> Result<crate::App, re_renderer::RenderContextError> {
     let build_info = re_build_info::build_info!();
 
+    let app_env = crate::AppEnvironment::Web {
+        url: cc.integration_info.web_info.location.url.clone(),
+    };
+
     let AppOptions {
         viewer_url,
         url,
@@ -732,10 +736,6 @@ fn create_app(
 
         fallback_token,
     } = app_options;
-
-    let app_env = crate::AppEnvironment::Web {
-        url: viewer_url.unwrap_or_else(|| cc.integration_info.web_info.location.url.clone()),
-    };
 
     if let Some(fallback_token) = fallback_token {
         match re_auth::Jwt::try_from(fallback_token) {
@@ -784,6 +784,7 @@ fn create_app(
         panel_state_overrides: panel_state_overrides.unwrap_or_default().into(),
 
         enable_history,
+        viewer_url,
     };
     crate::customize_eframe_and_setup_renderer(cc)?;
 
--- a/crates/viewer/re_viewer/src/web.rs
+++ b/crates/viewer/re_viewer/src/web.rs
@@ -661,6 +661,7 @@ impl From<PanelState> for re_types::blueprint::components::PanelState {
 // Keep in sync with the `AppOptions` interface in `rerun_js/web-viewer/index.ts`.
 #[derive(Clone, Default, Deserialize)]
 pub struct AppOptions {
+    viewer_url: Option<String>,
     url: Option<StringOrStringArray>,
     manifest_url: Option<String>,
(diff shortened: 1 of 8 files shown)


crates/viewer/re_viewer/src/web.rs
```diff
@@ -758,7 +774,14 @@ fn create_app(
...
             max_bytes: Some(2_500_000_000),
         },
-        location: Some(cc.integration_info.web_info.location.clone()),
+        location: Some(eframe::Location {
+            url: url
+                .as_ref()
+                .and_then(|s| s.first())
+                .unwrap_or_else(|| &cc.integration_info.web_info.location.url)
+                .clone(),
+            ..cc.integration_info.web_info.location.clone()
```
emilk: This updates only the `Location::url` but leaves all other fields with their incorrect values:

https://docs.rs/eframe/latest/wasm32-unknown-unknown/eframe/struct.Location.html

Adding a `eframe::Location::from_url` would probably be a good idea (and we can, once https://github.com/rerun-io/rerun/pull/11253 is landed).

eframe gets the `Location` from browser APIs btw, so if we change the browser location (I'm sure there is some API for that), this should already be correct

crates/viewer/re_viewer/src/web.rs
```diff
@@ -64,7 +64,11 @@ impl WebHandle {
     }
 
     #[wasm_bindgen]
-    pub async fn start(&self, canvas: JsValue) -> Result<(), wasm_bindgen::JsValue> {
+    pub async fn start(
+        &self,
+        canvas: JsValue,
+        viewer_url: JsValue,
```
emilk: Please add a docstring. Is this the url to the .wasm file, or what?

crates/viewer/re_viewer/src/web.rs
IsseW: Ah! This was meant to be opened as a draft PR sorry!

I don't think I'll end up actually needing to edit that location. But will keep this in mind if I do

crates/viewer/re_viewer/src/web.rs
IsseW: Moved that parameter to `AppOptions` and added doc string in typescript

crates/viewer/re_viewer/src/web.rs
IsseW: Yeah updating that part wasn't necessary

crates/viewer/re_viewer/src/web.rs
```diff
@@ -734,6 +733,10 @@ fn create_app(
         fallback_token,
     } = app_options;
 
+    let app_env = crate::AppEnvironment::Web {
```
jprochazk: `AppEnvironment` is used for analytics, for example:

https://github.com/rerun-io/rerun/blob/436ea5ca618ead9a6b26171552673900fb6f4ed3/crates/viewer/re_viewer/src/viewer_analytics/event.rs#L64-L67

This `url` field should store the current location in which the app is running. With these changes, it could point to `rerun.io/viewer` when the viewer is actually running in a `rerun.io/docs` page.

The `viewer_url` could be a new web-only field on `StartupOptions` instead

jprochazk (approved):

PR #11299, opened 2025-09-22
Automates BGR images support test

aedm:
### Related

https://linear.app/rerun/issue/RR-1120/replace-most-of-our-release-checklist-with-kittest

### What

Automates BGR image support test, part of the manual release checklist.
Checks all combinations of all int/float pixel formats, dark/light themes and alpha/opaque images.

- [x] full-check

### DIFF
--- a/crates/viewer/re_view_spatial/tests/snapshots/bgr_images_bgr_f16.png
+++ b/crates/viewer/re_view_spatial/tests/snapshots/bgr_images_bgr_f16.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:a7304950d0ce23b233f835c38d53e65f16b11026c3a9539c0a043eabdfe311b8
+size 42527
--- a/crates/viewer/re_view_spatial/tests/snapshots/bgr_images_bgr_f32.png
+++ b/crates/viewer/re_view_spatial/tests/snapshots/bgr_images_bgr_f32.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:a7304950d0ce23b233f835c38d53e65f16b11026c3a9539c0a043eabdfe311b8
+size 42527
--- a/crates/viewer/re_view_spatial/tests/snapshots/bgr_images_bgr_f64.png
+++ b/crates/viewer/re_view_spatial/tests/snapshots/bgr_images_bgr_f64.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:a7304950d0ce23b233f835c38d53e65f16b11026c3a9539c0a043eabdfe311b8
+size 42527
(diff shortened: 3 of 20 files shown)


aedm:
@rerun-bot full-check

crates/viewer/re_view_spatial/tests/bgr_images.rs
```diff
@@ -0,0 +1,165 @@
...
+
+fn convert_pixels_to<T: From<u8> + Copy>(u8s: &[u8]) -> Vec<T> {
+    u8s.iter().map(|u| T::from(*u)).collect()
+}
+
+fn run_bgr_test<T: ImageChannelType>(
+    image: &[T],
+    size: [u32; 2],
+    color_model: ColorModel,
+    is_light_theme: bool,
```
Wumpf: I actually think we shouldn't do this for dark & light ui here. The underlying issue was that those images ended up being transparent which was fine on black background and isn't on white. As light/dark theme just happens to change the default background color of the 2D view (there's no other difference), let's instead do that explicitly and avoid the permutation instead.
I'd suggest setting the view's blueprint such it uses some debugging friendly magenta. That should ofc then come with an explanation why we do that

crates/viewer/re_view_spatial/tests/bgr_images.rs
```diff
@@ -0,0 +1,165 @@
...
+        .with_size(size)
+        .build_ui(|ui| {
+            test_context.run_with_single_view(ui, view_id);
+        });
+    harness.ctx.set_theme(theme);
+
+    harness.run();
+    harness.snapshot_options(
+        name,
+        &SnapshotOptions::new().failed_pixel_count_threshold(OsThreshold::new(2).macos(2)),
```
Wumpf: 2.0 is a suspiciously large threshold, should see if we can use something smaller _if possible_. That said, this test probably needs a fair amount of that since texture may turn out quite differently per platform :/. So if you've already tried you can disregard that comment.

FYI: Highest threshold we've ever used so far is 2.2 + some small three digit value for failed pixels for video decoding tests (and also only for certain codecs, for AV1 we use a much smaller one)

emilk: This is a different threshold - this is the _number_ of failed pixels, not the threshold for failing a pixel. yes, this is confusing.

emilk: Me and @lucasmerlin talked today about a better long term solution for this, involving a global threshold set in a `.kittest.toml` config file, plus always using snapshots generated by CI (when we land an easy way of doing so). Stay tuned!

Wumpf: thanks for correcting - 2 pixel difference is very low, so if that works that's actually great and can stay as is!
Awesome news, looking forward to those changes!!

aedm: There is a one pixel difference between Linux and Macos with a threshold of about `0.7`. The default threshold is `0.6`afaict.

crates/viewer/re_view_spatial/tests/bgr_images.rs
aedm: Great idea! Done.

aedm:
> grinda.jpg should be uploaded as LFS file, it's a binary right now (all the snapshots are correctly done as LFS already). That's a picture you took yourself I reckon? :)

Yes, I wanted something where each color channel is immediately recognizeable (sky, grass, something red) and is guarantted to have no copyright issues. :)

Moved to LFS.

Issue #11274, opened 2025-09-19
Some .rrd files fail to load


emilk:
### Problem

Failing to show *some* snippet rrd:s on our web page, e.g. [https://rerun.io/docs/reference/types/archetypes/bar_chart](https://rerun.io/docs/reference/types/archetypes/bar_chart)

Repro: `pixi run rerun https://app.rerun.io/version/0.25.0/examples/snippets/bar_chart.rrd`

### Trailing data?

Is there actual trailing data in the `bar_chart.rrd`, or is it a bug in our parser?

If there is trailing data in it, how did it end up there?

### Why only some snippets?

Why are some snippets affected, but not all?

### Loading from disk and HTTP act differently

These two things act differently, and use different code paths:

`rerun https://app.rerun.io/version/0.25.0/examples/snippets/bar_chart.rrd`

vs

`wget https://app.rerun.io/version/0.25.0/examples/snippets/bar_chart.rrd && rerun bar_chart.rrd`

We should see if we can unify the code.

### Related

* [https://github.com/rerun-io/rerun/pull/11265](https://github.com/rerun-io/rerun/pull/11265)
* [https://github.com/rerun-io/rerun/issues/11266](https://github.com/rerun-io/rerun/issues/11266)
* [https://rerunio.slack.com/archives/C04UBTMMGJK/p1758181860652999](https://rerunio.slack.com/archives/C04UBTMMGJK/p1758181860652999)

teh-cmc:
### Status as of https://github.com/rerun-io/rerun/pull/11476

* `pixi run rerun https://app.rerun.io/version/0.25.0/examples/snippets/bar_chart.rrd`
    * Complains about `Trailing bytes in rrd stream: [0, 0, 0, 123, 34, 34, 0, 1, 8, 8, 144, 34] (Not an RRD file: expected FourCC header "RRF2", got "\0\0\0{")`.
    * Loads 2 different recordings and switches to the named one:
        <img width="459" height="147" alt="Image" src="https://github.com/user-attachments/assets/d7719354-b58e-4db5-ba12-43b62ea7b255" />
    * Output is a bunch of empty views:
        <img width="1309" height="1183" alt="Image" src="https://github.com/user-attachments/assets/ac9a022e-6c35-47d7-b36e-617a01bddce4" />
    * The `<unknown>` recording does contain the data and displays it appropriately though:
[...]

Issue #11266, opened 2025-09-18
Refactor .rrd loader: merge file and http loaders


emilk:
We currently have two different code paths when loading and .rrd from file, vs streaming it from http:

* File loader uses `crates/store/re_log_encoding/src/decoder/mod.rs`
* HTTP loader uses `crates/store/re_log_encoding/src/stream_rrd_from_http.rs`

These act slightly differently. It would be nice to unify them.

PR #11302, opened 2025-09-22
Improve robustness of rrd loader

emilk:
### Related
* Part of investigation of https://github.com/rerun-io/rerun/issues/11274
* Part of https://github.com/rerun-io/rerun/issues/11266

### What
Part of an aborted investigation into the bug linked above.

Mostly improved the logging, but also makes the streaming file loader more robust when there are trailing bytes in an .rrd file, logging a warning instead of returning an error.

### DIFF
--- a/crates/store/re_log_encoding/src/decoder/stream.rs
+++ b/crates/store/re_log_encoding/src/decoder/stream.rs
@@ -64,6 +64,9 @@ enum State {
     /// Compression is only applied to individual `ArrowMsg`s, instead of
     /// the entire stream.
     Message(crate::codec::file::MessageHeader),
+
+    /// Stop reading
+    Done,
 }
 
 impl StreamDecoder {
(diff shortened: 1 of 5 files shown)


crates/store/re_log_encoding/src/decoder/stream.rs
```diff
@@ -109,9 +112,30 @@ impl StreamDecoder {
...
                     // header contains version and compression options
-                    let (version, options) = options_from_bytes(header)?;
+                    let (version, options) = match options_from_bytes(header) {
+                        Ok(ok) => ok,
+                        Err(err) => {
+                            if is_first_header {
+                                return Err(err);
+                            } else {
+                                re_log::error!("Trailing bytes in rrd stream: {header:?} ({err})");
+                                self.state = State::Done;
```
Wumpf: seems strange that the only way to get into the `Done` state is via this failure case 🤔 
something missing?

Wumpf (approved):
lgtm except for strange state handling, see comment

crates/store/re_log_encoding/src/decoder/stream.rs
emilk: Yes, I find it confusing too. I renamed it `Aborted` now for clarity

PR #11216, opened 2025-09-15
Set parallelism for c++ snipepts

oxkitsune:
### What

Limits the parallelism for the building C++ snippets, to hopefully avoid OOMing on arm64

PR #11303, opened 2025-09-22
CI: Update Arm Linux runner from 24 GB -> 64 GB RAM

emilk:
It has been dying due to OOM more and more.

24 GB is huge, but 🤷 

NOTE: the biggest buildjet runner has 48 GB of RAM.

## Related
* https://github.com/rerun-io/rerun/pull/11216

#### TODO
*  Run nightly: https://github.com/rerun-io/rerun/actions/runs/17918934267
* [ ] Run nightly again: https://github.com/rerun-io/rerun/actions/runs/17937078521

### DIFF
--- a/.github/workflows/reusable_test_wheels.yml
+++ b/.github/workflows/reusable_test_wheels.yml
@@ -71,7 +71,7 @@ jobs:
         run: |
           case "${{ inputs.PLATFORM }}" in
             linux-arm64)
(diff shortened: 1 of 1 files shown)


emilk:
@rerun-bot full-check

emilk:
The arm build succeeds now, so that's good

PR #11304, opened 2025-09-22
Remove store info from `ChunkStore`

Wumpf:
It was redundant and not kept up to date in all circumstances with its parent `EntityDb`, something that caused me grief/confusion while working in the area.


Wumpf:
yeah, was surprised as well. Bit me when adding a property to it and then expecting it to be up to date in the ChunkStore

Issue #10692, opened 2025-07-17
Time-range urls don't work well


jleibs:
The following somewhat demonstrates the problem.

```
import rerun as rr
from rerun.utilities.datafusion.functions.url_generation import partition_url
import pyarrow as pa
import pyarrow.compute # Needed for https://github.com/rerun-io/rerun/issues/10689
import datafusion as dfn
from rerun.notebook import Viewer
```

```python
REDAP_URI="rerun://sandbox.redap.rerun.io"
DATASET_NAME="droid:raw"
WIDTH=1100
HEIGHT=800
```

```
client = rr.catalog.CatalogClient("rerun://sandbox.redap.rerun.io")
dataset = client.get_dataset(name="droid:raw")
```

```python
url = dataset.partition_url('ILIAD_5e938e3b_2023_07_28_10h_20m_36s')
```

```python
Viewer(width=WIDTH, height=HEIGHT, url=url)
```

<img src="https://uploads.linear.app/93919c5e-abe4-4cd5-b77d-7e3eafe3e62d/92d880b1-d116-40ca-b9eb-b01a82d825d1/99f49991-0069-453c-a521-3234f53c4eb8?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzkzOTE5YzVlLWFiZTQtNGNkNS1iNzdkLTdlM2VhZmUzZTYyZC85MmQ4ODBiMS1kMTE2LTQwY2EtYjllYi1iMDFhODJkODI1ZDEvOTlmNDk5OTEtMDA2OS00NTNjLWE1MjEtMzIzNGY1M2M0ZWI4IiwiaWF0IjoxNzUyNzg3Nzc3LCJleHAiOjMzMzIzMzQ3Nzc3fQ.kc8QiftqNTQGsgIKNfvutRUB3kPWErO3gvgJuwn-qXw " alt="image.png" width="1396" data-linear-height="262" />

```python
#TODO: Need a helper for this
url = dataset.partition_url('ILIAD_5e938e3b_2023_07_28_10h_20m_36s')+'&time_range=real_time@2023-07-28T17:20:40Z..2023-07-28T17:20:43Z'
```

```python
Viewer(width=WIDTH, height=HEIGHT, url=url)
```

[...]

Issue #11315, opened 2025-09-23
Stream data from servers on-demand for larger-than-RAM support


Wumpf:
TODO: There's various ideas about this floating around, at this point this is merely a placeholder issue for reference in affected code.

Follow-up ticket to
* https://github.com/rerun-io/rerun/issues/5247

Keywords for search-ability: larger than RAM, out of memory, on demand loading

Wumpf:
The Rerun dataplatform as well as the rerun opensource server address this by now!

PR #11318, opened 2025-09-24
Viewer now crops incoming chunks if time range was requested

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/10692
* This is in many ways just part of a workaround for inspecting large episodes, making it related to https://github.com/rerun-io/rerun/issues/11315

### What

Datastore URLs allow querying only a time range. This then leads us to pulling down the necessary chunks, causing arbitrarily more data in the Viewer to arrive, depending on how big the chunks are.

This PR introduces crop-on-insert in order to crop chunks to the specified range, thus honoring the previously requested time bounds more accurately.
@abey79 and me decided to crop only on the specified timeline (not cropping on any other) and allow recombining cropped chunks into an existing store. This video demonstrates what this means in practice and feature some slight ui changes:

https://github.com/user-attachments/assets/395223c3-4025-439f-a9ae-806f191713f2

While the cropping itself is a saimple range query on incoming chunks, in practice this gives rise to a whole slew of issues like how to deal with overlapping data, chunk ids, row ids, etc.. The most important ones like "what happens if I load the recording after all" & "what happens if I load overlapping ranges" are well-enough addressed here, but it's not a clean solution overall as you easily may end up in a situation with duplicated data in-memory.
I firmly believe that we should remove this mechanism again once we have dynamic occupancy handling for continuous streaming within a memory budget (#11315). We'll still want a way to highlight selected time ranges, but we can do that on a per-visualization basis as needed.

### DIFF
--- a/crates/store/re_chunk/src/range.rs
+++ b/crates/store/re_chunk/src/range.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use re_log_types::{AbsoluteTimeRange, TimeInt, TimelineName};
 use re_types_core::ComponentDescriptor;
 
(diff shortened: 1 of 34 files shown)


IsseW:
I also noticed that entities which don't move during the recording but are set up at some specific location don't work correctly in trimmed ranges, for example the gripper in [this](https://rerun.io/viewer/pr/11318?url=rerun%3A%2F%2Fsandbox.redap.rerun.io%3A443%2Fdataset%2F186446580E79EA77776cd5d9a5842c6c%3Fpartition_id%3DAUTOLab_0d4edc83_2023_12_02_13h_55m_02s%26time_range%3Dreal_time%402023-12-02T21%3A55%3A04.564623574Z..2023-12-02T21%3A55%3A05.894623575Z%23selection%3D%2Fpanda_link0%2Fpanda_link1%2Fpanda_link2%2Fpanda_link3%2Fpanda_link4%2Fpanda_link5%2Fpanda_link6%2Fpanda_link7%2Fpanda_link8%2Frobotiq_85_base_link%2Fvisual_0%26when%3Dreal_time%402023-12-02T21%3A55%3A05.318623574Z). Which is a bit similar to the "no keyframes" video issue, but doesn't feel like it's working as expected.

Expected would probably be start at the last logged location before the time selection. But not sure how to fix that in a good way.

Wumpf:
at first I thought this was a problem with the recording itself, but confirmed now that this is instead an issue regarding the cropping implementation. Looking into it

Wumpf:
Fixed, causes wider time ranges as we should expect

image

Wumpf:
drafted since there's some ongoing discussion on UX & future of the hacky parts

Issue #11320, opened 2025-09-24
Fix "Invalid null value" CI failure


emilk:
<img width="1802" height="1528" alt="Image" src="https://github.com/user-attachments/assets/342e696e-ba92-401a-83ee-75e4b4976876" />

emilk:
This is VERY hard to repro, but I managed to catch it again: `_try_parse_string` calls `pa.array([], type=None)` and that results in an array with `pa.null()` as datatype, which causes downstream errors.

Issue #11319, opened 2025-09-24
Improve Python linting: catch unused argument errors


emilk:
We currently have functions in our code with unused arguments. That smells like a bug to me:

https://github.com/rerun-io/rerun/blob/b5c3d7add12c02c3f5a6dd60dfebae8e85d10395/rerun_py/rerun_sdk/rerun/any_batch_value.py#L17-L29

emilk:
This seems to fix it

```diff
diff --git a/rerun_py/pyproject.toml b/rerun_py/pyproject.toml
index cb9afb9e51..0c5b6b4da9 100644
--- a/rerun_py/pyproject.toml
+++ b/rerun_py/pyproject.toml
@@ -143,6 +143,7 @@ line-length = 120
 lint.select = [
   "A",   # Avoid shadowing built-ins
   "ANN", # Ensure all code is properly annotated.
+  "ARG", # Detect unused arguments
   "B",   # flake8-bugbear lints
   "COM", # Trailing-comma-related rules
   "C4",  # Comprehension-related rules
```

ntjohnson1:
Yes but we have 22 failures in the repo so part of this should probably be to make that lint pass.

PR #11321, opened 2025-09-24
Improve error message when `AnyValue` fails

emilk:
Cleanup in pursuit of fixing
* https://github.com/rerun-io/rerun/issues/11320

### DIFF
--- a/rerun_py/rerun_sdk/rerun/any_batch_value.py
+++ b/rerun_py/rerun_sdk/rerun/any_batch_value.py
@@ -9,7 +9,7 @@ from pyarrow import ArrowInvalid
 from rerun._baseclasses import ComponentDescriptor
 
 from ._baseclasses import ComponentBatchLike, ComponentColumn
-from .error_utils import catch_and_log_exceptions
+from .error_utils import catch_and_log_exceptions, strict_mode
 
 ANY_VALUE_TYPE_REGISTRY: dict[ComponentDescriptor, Any] = {}
 
@@ -58,6 +58,11 @@ def _try_parse_string(
     if not isinstance(value, (str, bytes)):
         try:
             pa_array = pa.array(value, type=pa_type)
+            if strict_mode():
+                assert pa_array.type != pa.null(), (
+                    f"pa.array of value {value} and type {pa_type} resulted in type {pa_array.type}"
+                )
+
             if descriptor is not None:
                 ANY_VALUE_TYPE_REGISTRY[descriptor] = (None, pa_array.type)
             return pa_array


rerun_py/rerun_sdk/rerun/any_batch_value.py
```diff
@@ -9,28 +9,33 @@
...
 from ._baseclasses import ComponentBatchLike, ComponentColumn
-from .error_utils import catch_and_log_exceptions
+from .error_utils import catch_and_log_exceptions, strict_mode
 
 ANY_VALUE_TYPE_REGISTRY: dict[ComponentDescriptor, Any] = {}
 
 
 def _parse_arrow_array(
-    value: Any, pa_type: Any | None = None, np_type: Any | None = None, descriptor: ComponentDescriptor | None = None
+    value: Any, *, pa_type: Any | None = None, np_type: Any | None = None, descriptor: ComponentDescriptor | None = None
```
emilk, line 18: NOTE: `np_type` is unused here, which is SUS.

* https://github.com/rerun-io/rerun/issues/11319

rerun_py/rerun_sdk/rerun/any_batch_value.py
```diff
@@ -9,28 +9,33 @@
...
         return possible_array
-    possible_array = _try_parse_scalar(value, pa_type, descriptor)
+    possible_array = _try_parse_scalar(value, pa_type=pa_type, descriptor=descriptor)
     if possible_array is not None:
         return possible_array
-    return _fallback_parse(value, pa_type, None, descriptor)
+    return _fallback_parse(
+        value,
+        pa_type=pa_type,
+        np_type=None,  # TODO(emilk): np_type=None looks VERY suspicious here
```
ntjohnson1, line 32: Definitely a bug from this [commit](https://github.com/rerun-io/rerun/pull/11045/commits/b8c3c7f6a00531b91a82762000aa569017b9de16) when I shuffled the code around. Looks like I should definitely revisit since none of our tests screamed.

rerun_py/rerun_sdk/rerun/any_batch_value.py
```diff
@@ -46,13 +51,18 @@ def _try_parse_dlpack(
...
 def _try_parse_string(
-    value: Any, pa_type: Any | None = None, descriptor: ComponentDescriptor | None = None
+    value: Any, *, pa_type: Any | None = None, descriptor: ComponentDescriptor | None = None
 ) -> pa.Array | None:
     # Special case: strings are iterables so pyarrow will not
     # handle them properly
     if not isinstance(value, (str, bytes)):
         try:
             pa_array = pa.array(value, type=pa_type)
+            if strict_mode():
```
ntjohnson1, line 61: Isn't the whole point of strict mode that we can just raise and when not strict they just end up as warnings? This seems to duplicate that mechanism.

emilk: I don't understand the code well enough (yet) to know if it is fine to unconditionally raise here on null, so I wanted to keep the previous behavior of "continue and hope for the best". But I'm not 100% convinced it is a good idea.

PR #11322, opened 2025-09-24
Improve error message

emilk:
### Related
* Merge https://github.com/rerun-io/rerun/pull/11321 first
* Closes https://github.com/rerun-io/rerun/issues/11320

### What
When using `AnyValues` we try to infer the type from the argument.
This does not work if the argument is `None`, so in those cases we just ignore that column.
This can be controlled with `drop_untyped_nones` (default: `True`).

It _also_ doesn't work when given an empty sequence, `[]`. However, we did not previously take this count into consideration. When logging `[]` we would assign the column the type `pa.null()`. If the user later logged actual data on that column (e.g. `[1, 2, 3]`) that would lead to an exception: "Invalid null value" (because we expected `[null, null, …]` but got `[1, 2, 3]`).

The new code now drops any `[]` calls until we know their type.
This means logging `[]` followed by `[1, 2, 3]` will result in just `[1, 2, 3]` showing up in your recorded data.
Logging `[1, 2, 3]` followed by `[]` will NOT discard the `[]` though.

### TODO
* [ ] Add tests
* [ ] Are there other empty sequences beyond `[]` we need to worry about? Like `[[]]`?
  * [ ] Is there a more robust way to handle this?

### DIFF
--- a/rerun_py/rerun_sdk/rerun/any_batch_value.py
+++ b/rerun_py/rerun_sdk/rerun/any_batch_value.py
@@ -155,7 +155,7 @@ class AnyBatchValue(ComponentBatchLike):
         value:
             The data to be logged as a component.
         drop_untyped_nones:
-            If True, any components that are None will be dropped unless they have been
+            If True, any components that are either None or `[]` will be dropped unless they have been
             previously logged with a type.
 
         """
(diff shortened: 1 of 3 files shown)


rerun_py/rerun_sdk/rerun/any_batch_value.py
```diff
@@ -155,7 +155,7 @@ def __init__(self, descriptor: str | ComponentDescriptor, value: Any, drop_untyp
         value:
             The data to be logged as a component.
         drop_untyped_nones:
-            If True, any components that are None will be dropped unless they have been
+            If True, any components that are either None or `[]` will be dropped unless they have been
```
ntjohnson1: My guess is that this edge case is broader and we should instead check what gets coerced into a `pa.null()` I haven't tried it yet but I'd suspected `np.array([])` would similarly end up a null.

rerun_py/tests/unit/test_any_values.py
```diff
@@ -147,6 +149,29 @@ def test_iterable_any_value() -> None:
...
+    })
+
+    with pytest.warns(RerunWarning) as warnings:
+        batches = list(values.as_component_batches())
+        assert len(batches) == 0
+
+        assert "No valid component batches" in str(warnings[0].message)
+
+    batches = list(new_values.as_component_batches())
+    assert len(batches) == 3
```
emilk, line 172: ⭐

PR #11325, opened 2025-09-24
Only include web url for copies in web-viewer

IsseW:
### What

The native viewer currently copies a web viewer url on some UI actions, which it shouldn't do.


Wumpf (approved):
thx!

PR #11327, opened 2025-09-24
Warn about unused args in Python code

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/11319
* Also needs https://github.com/rerun-io/rerun/pull/11326

### DIFF
--- a/rerun_notebook/src/rerun_notebook/__init__.py
+++ b/rerun_notebook/src/rerun_notebook/__init__.py
@@ -200,7 +200,7 @@ class Viewer(anywidget.AnyWidget):  # type: ignore[misc]
         if fallback_token:
             self._fallback_token = fallback_token
 
-        def handle_msg(widget: Any, content: Any, buffers: list[bytes]) -> None:
+        def handle_msg(_widget: Any, content: Any, _buffers: list[bytes]) -> None:
             if isinstance(content, str):
                 if content == "ready":
                     self._on_ready()
--- a/rerun_py/pyproject.toml
+++ b/rerun_py/pyproject.toml
@@ -145,8 +145,8 @@ lint.select = [
   "ANN", # Ensure all code is properly annotated.
   "ARG", # Detect unused arguments
   "B",   # flake8-bugbear lints
-  "COM", # Trailing-comma-related rules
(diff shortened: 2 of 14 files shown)


rerun_notebook/src/rerun_notebook/__init__.py
```diff
@@ -200,7 +200,7 @@ def __init__(
         if fallback_token:
             self._fallback_token = fallback_token
 
-        def handle_msg(widget: Any, content: Any, buffers: list[bytes]) -> None:
+        def handle_msg(_widget: Any, content: Any, _buffers: list[bytes]) -> None:
```
ntjohnson1: NIT: Would prefer to just add no_qa here since this does change the api (I don't know if anyone is calling these by kwargs instead of positional)

ntjohnson1 (approved):

rerun_notebook/src/rerun_notebook/__init__.py
emilk:
```suggestion
        def handle_msg(widget: Any, content: Any, buffers: list[bytes]) -> None:  # noqa: ARG002
```

rerun_notebook/src/rerun_notebook/__init__.py
```diff
@@ -200,7 +200,7 @@ def __init__(
         if fallback_token:
             self._fallback_token = fallback_token
 
-        def handle_msg(widget: Any, content: Any, buffers: list[bytes]) -> None:
+        def handle_msg(widget: Any, content: Any, buffers: list[bytes]) -> None:  # noqa: ARG002
```
emilk:
```suggestion
        def handle_msg(widget: Any, content: Any, buffers: list[bytes]) -> None:  # noqa: ARG001
```

rerun_py/pyproject.toml
```diff
@@ -143,6 +143,7 @@ line-length = 120
 lint.select = [
   "A",   # Avoid shadowing built-ins
   "ANN", # Ensure all code is properly annotated.
+  "ARG", # Detect unused arguments
```
Wumpf, line 146: 🥳

PR #11328, opened 2025-09-24
Snapshot utility function in TestContext

aedm:
### What

Bikeshedding: replaces copies of the `run_view_ui_and_save_snapshot` function and adds it to TestContext.

Some of the snapshots had to be updated because the new utility function has one more `harness.run()`.


crates/viewer/re_view_bar_chart/tests/bar_chart_test.rs
```diff
@@ -24,12 +24,7 @@ fn test_bar_chart() {
     });
 
     let view_id = setup_blueprint(&mut test_context);
-    run_view_ui_and_save_snapshot(
-        &mut test_context,
-        view_id,
-        "tensor_1d",
-        egui::vec2(400.0, 300.0),
-    );
+    test_context.run_view_ui_and_save_snapshot(view_id, "bar_chart_1d", egui::vec2(400.0, 300.0));
```
oxkitsune: Might be good to change the entity path as well, when the name is changed here.

crates/viewer/re_test_viewport/src/lib.rs
```diff
@@ -166,4 +173,21 @@ impl TestContextExt for TestContext {
...
         self.handle_system_commands();
     }
+
+    fn run_view_ui_and_save_snapshot(
+        &mut self,
+        view_id: ViewId,
+        snapshot_name: &str,
+        size: egui::Vec2,
+    ) {
+        let snapshot_options = self.snapshot_options.clone();
```
Wumpf: I relate to the desire of having `run_view_ui_and_save_snapshot` to make some of the tests shorter. But having the snapshot options live on the test context live on the test context feels a bit arbitrary to me because:
* size is not on the test context, but the many of the snapshot options are sensitive to it
* this utility is the only thing that uses the options, but its only available on the extensions for views

not a biggy, afaik @lucasmerlin gonna refactor how we set those options soon anyways

aedm: Good catch. I removed the field, added an argument to the function instead.

PR #11329, opened 2025-09-24
Enable more Python lints

emilk:
### What
Enable a bunch more Python lints and then run automatic formatting based on them.

All code changes was made by `ruff`, but look OK to me (a non-Python expert)


examples/python/imu_signals/imu_signals.py
```diff
@@ -77,7 +77,7 @@ def _download_dataset(root: Path, dataset_url: str = DATASET_URL) -> None:
                 pb.update(len(data))
                 file.write(data)
 
-    if total_size != 0 and pb.n != total_size:
+    if total_size not in (0, pb.n):
```
emilk, line 80: Shouldn't this be `if total_size != pb.n` ?

rerun_py/rerun_bindings/types.py
```diff
@@ -10,12 +10,11 @@
...
-        ComponentColumnSelector as ComponentColumnSelector,
-        ComponentDescriptor as ComponentDescriptor,
-        IndexColumnDescriptor as IndexColumnDescriptor,
-        IndexColumnSelector as IndexColumnSelector,
-        VectorDistanceMetric as VectorDistanceMetric,
+        ComponentColumnDescriptor,
+        ComponentColumnSelector,
+        IndexColumnDescriptor,
+        IndexColumnSelector,
+        VectorDistanceMetric,
```
emilk: `ComponentDescriptor` got lost here 🤔

rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -4,14 +4,11 @@
...
-    EntryId as EntryId,
-    EntryKind as EntryKind,
-    TableEntry as TableEntry,
-    Task as Task,
-    VectorDistanceMetric as VectorDistanceMetric,
+    DatasetEntry,
+    Entry,
+    EntryId,
+    EntryKind,
+    TableEntry,
```
emilk: Here again some things went missing

examples/python/imu_signals/imu_signals.py
ntjohnson1: We should run this example. I wonder if pb.n sometimes isn't calculated and the initial condition protects against that?

examples/python/ocr/ocr.py
```diff
@@ -226,7 +226,7 @@ def get_table_markdown(line: dict[str, Any]) -> str:
             return markdown_table  # type: ignore[no-any-return]
 
         except Exception as e:
-            return f"Error processing the table: {str(e)}"
+            return f"Error processing the table: {e!s}"
```
ntjohnson1, line 229: This is fine but I think we don't even need the `str` wrapping in a formatted string.

rerun_py/pyproject.toml
```diff
@@ -141,27 +141,35 @@ lint.ignore = [
...
   "TID",     # flake8-tidy-imports
   "UP",      # pyupgrade (ensures idomatic code for supported python version)
   "W",       # pycodestyle warning codes: https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes
   "YTT",     # Various checks on the use of `sys.version_info` and related.
+
+  # TODO(emilk): Consider enable more of these lints:
+  # "ERA",     # eradicate dead code (WONT DO: we have lots of dead-looking code in comments)
+  # "LOG",     # flake8-logging (logging best practices) # NOT SURE
+  # "PERF",    # Perflint (performance anti-patterns) # MAYBE
+  # "PL",      # Pylint (comprehensive linting)
```
ntjohnson1, line 171: I like the pedantic nature of pylint. There are subsets of this that are less aggressive so if we move in this direction could enable piecemeal.

rerun_py/rerun_bindings/types.py
ntjohnson1: I think that's because it isn't used in the file.

rerun_py/rerun_sdk/rerun/components/media_type_ext.py
```diff
@@ -100,7 +100,7 @@ def guess_from_path(path: str | Path) -> MediaType | None:
         ext = Path(path).suffix.lower()
 
         # Images
-        if ext == ".jpg" or ext == ".jpeg":
+        if ext in {".jpg", ".jpeg"}:
```
ntjohnson1, line 103: Interesting that the other one checked in a tuple where this is a set. For small values shouldn't matter much. Maybe it does it by datatype and hashing the string makes sense.

rerun_notebook/src/rerun_notebook/__init__.py
```diff
@@ -172,11 +172,11 @@ class Viewer(anywidget.AnyWidget):  # type: ignore[misc]
     ).tag(sync=True)
 
     _ready = False
-    _event_queue: list[_Event] = []
+    _event_queue: ClassVar[list[_Event]] = []
```
ntjohnson1: Not opposed to this but need to look at the rule for the change.

ntjohnson1 (approved):
LGTM. Will look at the ClassVar rule for my own interest.

rerun_notebook/src/rerun_notebook/__init__.py
ntjohnson1: Ya so ruff "fixed" this but I don't think we want mutable defaults for these class types. I don't know the code THAT well but seems like a major footgun. Maybe a follow on to add to the rerun specific lints to yell if we see the ClassVar so people are really sure they want multiple class instances to mutate default values.

Wouldn't consider this blocking since this was already concerning before but clearer now. Unfortunately there isn't a way to just error on this for class AFAICT but at least there is for [functions](https://docs.astral.sh/ruff/rules/mutable-argument-default/) 

[RUF012](https://docs.astral.sh/ruff/rules/mutable-class-default/)

emilk: I opted-out of RUF012 for now

rerun_py/rerun_sdk/rerun/components/media_type_ext.py
emilk: I guess/hope this is mostly a syntactic change

rerun_py/pyproject.toml
emilk: Yes please! Feel free to enable any and all of these :)

examples/python/ocr/ocr.py
emilk: Yeah, probably not. Feel free to remove all `!s` in a separate PR if you feel confident enough (I don't 😅)

rerun_py/rerun_sdk/rerun/components/media_type_ext.py
ntjohnson1: Yes. Functionally should still be the same. Just interesting.

rerun_notebook/src/rerun_notebook/__init__.py
ntjohnson1: Fair not a blocker on this PR.

I mostly meant that the code pattern we are using that necessitates triggers this rule should be resolved but that's bigger than a linting thing.

For instance something like

```python
_event_queue: Optional[list[_Event]] = None
…
def __init__(…):
   _event_queue = []
```

examples/python/imu_signals/imu_signals.py
ntjohnson1:
```python
with tqdm(desc="Downloading dataset", total=0, unit="B", unit_scale=True) as pb:
        pass
print(pb.n)
```
Does print 0 so I think the 0 check is redundant.

PR #11330, opened 2025-09-24
Fix url edit warning spam

IsseW:
### Related

Closes RR-2369

### What

Fixes warnings spam when editing an invalid url in open url modal / command palatte.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8785,7 +8785,6 @@ dependencies = [
 name = "re_uri"
 version = "0.26.0-alpha.1+dev"
 dependencies = [
- "re_log",
  "re_log_types",
  "re_tuid",
  "serde",
--- a/crates/utils/re_uri/Cargo.toml
+++ b/crates/utils/re_uri/Cargo.toml
@@ -15,7 +15,6 @@ version.workspace = true
 workspace = true
 
 [dependencies]
-re_log.workspace = true
 re_log_types = { workspace = true, features = ["serde"] }
 re_tuid.workspace = true
 
--- a/crates/utils/re_uri/src/endpoints/dataset.rs
+++ b/crates/utils/re_uri/src/endpoints/dataset.rs
@@ -67,7 +67,7 @@ impl DatasetPartitionUri {
                     time_range = Some(value.parse::<TimeSelection>()?);
                 }
                 _ => {
-                    re_log::debug_once!("Unknown query parameter: {key}={value}");
+                    // We ignore unknown query keys that may be from urls from prior/newer versions.
                 }
             }
         }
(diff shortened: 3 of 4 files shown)


Wumpf (commented):
not an amazing solution obviously as this means that we now get log spam just on debug (which is only enabled by default on cli in the repo) instead of everywhere :/

Also, more importantly, if I start Rerun with a URL and there was stuff that's ignored, that's still nice to known about, it's just that in some situations we expect things to be more wrong than right 🤔 
How we solve this in other places is to have a `parse` & `parse_forgiving` version; that seems overkill here, but I guess we could have a parse method that takes a log level and then use the debug version on `from_str`? Once that's in place, we can opt-in to the warning again in key locations, i.e. the places where we actually open a url from a string.

Alternative idea I just had: we could have "notification off"-scopes in which log messages aren't forwarded to the toast notification system 🤔 

if this turns into a rabbithole we can also go with the solution the PR proposes as it _does_ address the problem at hand.

IsseW:
I'm not really sure the warnings would be useful to users, at least not with the way the warnings are written.

One other solution I started to attempt was to have a result enum with 3 variants `Ok`, `Err` and `Warn`. Where `Warn` does return a parsed url, but also has some warnings the caller can choose how to handle.

For example in the command palette / open url modal we could have a warning symbol that shows more details on hover. And for loading an url when starting the app we could actually warn with those warnings.

It looks like that'd be a somewhat big refactor to that though.

Wumpf:
> I'm not really sure the warnings would be useful to users, at least not with the way the warnings are written.

If we're convinced of that then maybe we should just remove them entirely and replacement with mere comments of the rationale why we ignore certain parts of the URL. That's also a solution I guess, hum.

> [...] It looks like that'd be a somewhat big refactor to that though.

Agreed that this would be useful and also agreed that that would be a too big refactor right now that we can't really justify

IsseW:
> If we're convinced of that then maybe we should just remove them entirely and replacement with mere comments of the rationale why we ignore certain parts of the URL. That's also a solution I guess, hum.

Yeah I agree that would make more sense than debug then.

> but I guess we could have a parse method that takes a log level and then use the debug version on from_str? Once that's in place, we can opt-in to the warning again in key locations, i.e. the places where we actually open a url from a string.

I think this could be something we could do, for actually opening the url the warnings aren't as invasive. But also feels a bit hacky :/

> Alternative idea I just had (that addresses toasts, not log spam in general): we could have "notification off"-scopes in which log messages aren't forwarded to the toast notification system

I feel like this could be a bit of a footgun, because maybe you then get some situation further in where you actually want to log errors to the toast. Should that then have a "notification on"-scope?

Wumpf:
Okay, unless you have new/further concerns I'd call it and say we just remove the log messages and only leave a small rationale-comment on why we ignore those bits. All my other suggestions are just unnecessary churn but I still think moving the spam from warn to debug isn't great.

PR #11333, opened 2025-09-24
Fix links for custom timelines

IsseW:
### Related

Closes https://github.com/rerun-io/rerun/issues/11191

### What

Fixes opening links to custom timelines and adds UI to show that the timeline isn't there yet.

https://github.com/user-attachments/assets/5813e981-0672-4b19-acb6-1b30fe9da753

### DIFF
--- a/crates/viewer/re_global_context/src/command_sender.rs
+++ b/crates/viewer/re_global_context/src/command_sender.rs
@@ -112,6 +112,8 @@ pub enum SystemCommand {
         store_id: StoreId,
         timeline: re_chunk::Timeline,
         time: Option<re_log_types::TimeReal>,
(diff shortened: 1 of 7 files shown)


crates/viewer/re_global_context/src/command_sender.rs
```diff
@@ -112,6 +112,9 @@ pub enum SystemCommand {
         store_id: StoreId,
         timeline: re_chunk::Timeline,
         time: Option<re_log_types::TimeReal>,
+
+        /// If this is true the timeline will persist even if it is invalid at the moment.
+        pending: bool,
```
Wumpf, line 117: good idea to make this opt-in 💡

PR #11335, opened 2025-09-25
Better gRPC errors

IsseW:
### Related

Closes RR-2338

### What

Splits the `StreamError::Tonic` error into multiple different variants, making the errors more readable and useful.

Also adds a details section to notification, which are automatically created if the notification contains a "<details>" part.

### DIFF
--- a/crates/viewer/re_ui/src/notifications.rs
+++ b/crates/viewer/re_ui/src/notifications.rs
@@ -5,7 +5,7 @@
 //!
 //! ## Special cased text
 //! - If a notifications text contains `"\nDetails:"` the section after that
-//!   will be displayed inside a collapsable details header.
+//!   will be displayed inside a collapsible details header.
 
 use std::time::Duration;
 
@@ -102,7 +102,7 @@ pub struct Notification {
     level: NotificationLevel,
     text: String,
 
-    /// if set this notifications will have a collapsable details section.
+    /// if set this notifications will have a collapsible details section.
     details: Option<String>,
     link: Option<Link>,
 
@@ -216,7 +216,7 @@ impl NotificationUi {
     ///
     /// ## Special cased text
     /// - If a notifications text contains `"\nDetails:"` the section after that
-    ///   will be displayed inside a collapsable details header.
+    ///   will be displayed inside a collapsible details header.
     pub fn add_log(&mut self, message: re_log::LogMsg) {
         let re_log::LogMsg { level, target, msg } = message;
 


Wumpf:
before/after screenshot plz :)

IsseW:
Toast before:
Screenshot 2025-09-26 at 10 03 33

Toast after:
Screenshot 2025-09-26 at 10 05 37

Panel after (details section is folded by default):
Screenshot 2025-09-26 at 10 07 25

IsseW:
This should apply to most stream errors, like the auth error:
Screenshot 2025-09-26 at 10 45 16

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -450,10 +450,20 @@ fn show_notification(
...
-                        ui.label(egui::RichText::new(text.clone()));
+                        let (split_text, details) = text.split_once("<details>").unzip();
+                        let text = split_text.unwrap_or(text);
+                        ui.label(egui::RichText::new(text));
+
+                        if let Some(details) = details {
+                            if mode == DisplayMode::Panel {
+                                ui.collapsing_header("Details", false, |ui| ui.label(details));
+                            } else {
+                                ui.weak("Open notification toast for more details");
```
Wumpf: Hmm but technically the "toast" is the temporary thing, not the card in the "permanent" menu.

```suggestion
                                ui.weak("Open notification menu for more details");
```

maybe better?

Anyways, as discussed on slack it would be cool to make the collapsing panel work consistently imho

crates/store/re_data_source/src/data_source.rs
```diff
@@ -261,10 +261,7 @@ impl LogDataSource {
 
                 spawn_future(async move {
                     if let Err(err) = stream_partition.await {
-                        re_log::warn!(
-                            "Error while streaming {uri}: {}",
-                            re_error::format_ref(&err)
-                        );
+                        re_log::warn!("Error while streaming: {}", &err);
```
Wumpf: removed because the error already contains the uri? Why remove `format_ref`, bad formatting?

crates/store/re_redap_client/src/connection_registry.rs
```diff
@@ -64,13 +64,13 @@ pub enum ClientConnectionError {
     #[error("server is expecting an unencrypted connection (try `rerun+http://` if you are sure)")]
     UnencryptedServer,
 
-    #[error("the server requires an authentication token, but none was provided: {0}")]
+    #[error("the server requires an authentication token, but none was provided<details>{0}")]
```
Wumpf: Likely out of scope, but I'm curious whether we can remove the `Tonic(#[from]` above as well and always pack it up in a `<details>` error.
At the very least I figure it might be a good idea to hide the exact transport error as well since it's likely not very readable 🤔

crates/store/re_redap_client/src/lib.rs
```diff
@@ -78,13 +78,55 @@ impl std::error::Error for TonicStatusError {
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum StreamEntryError {
```
Wumpf, line 82: 😍

Wumpf (changes requested):
I love this feature and the simplicity of how to put log messages into the details section of the toast!

Buuut the problem with the suggested `<details>` syntax is that it makes the errors on command line (where they still do show up) hard to read. In particular things from `re_redap_client` also used in the Python SDK, potentially with no viewer in the loop at all. I think we need to find a syntax that works better here.
Maybe it could be `\nDetails:`?

Also, we need to document this behavior somewhere. Maybe as a `//!` comment in `notification.rs`, not sure where I'd look for this tbh.

crates/store/re_data_source/src/data_source.rs
IsseW: > Why remove format_ref, bad formatting?

Ah think I just removed it for testing and forgot to add it back, doesn't seem to change anything though, but the function says to always use it. ✅ 

> removed because the error already contains the uri?

The error does say stuff like which partition wasn't found in the details. But mainly because our urls are huge 😬  I had the idea of maybe automagically shortening urls in notifications as well, and wonder if it'd be possible to make them clickable? Probably for another PR though.

I've seen other places shorten long urls by doing something like [https://rerun.io/viewer/pr/11340?url=rerun%3A[...]6369Z](https://rerun.io/viewer/pr/11340?url=rerun%3A%2F%2Fsandbox.redap.rerun.io%3A443%2Fdataset%2F186446580E79EA77776cd5d9a5842c6c%3Fpartition_id%3DAUTOLab_0d4edc83_2023_11_03_15h_45m_58s%26time_range%3Dreal_time%402023-11-03T22%3A46%3A01.455965998Z..2023-11-03T22%3A46%3A02.680965999Z%23when%3Dreal_time%402023-11-03T22%3A46%3A01.68266369Z)

Wumpf: uh yeah that would be cool, yes!

crates/viewer/re_ui/src/notifications.rs
```diff
@@ -193,11 +211,27 @@ impl NotificationUi {
         self.unread_notification_level
     }
 
+    /// Given that the log is relevant this creates a notification
+    /// based on that log.
+    ///
+    /// ## Special cased text
+    /// - If a notifications text contains `"\nDetails:"` the section after that
+    ///   will be displayed inside a collapsible details header.
```
emilk, line 219: It would be good to also document this somewhere in `re_log`

PR #11337, opened 2025-09-25
Copy button on some selection items

### LINKED ISSUE #10815: All URL components should have a copy button

IsseW:
### Related

Closes #10815

### What

Adds a copy button to more items in the selection panel.


Wumpf:
Doesn't quite work out here:
image

Wumpf (changes requested):
I like that this is a universal approach! I pictured something that would be implemented on component_ui basis.
But there's unfortunately cases where this obscures the little text there is and I this looks pretty hard to fix, but maybe there's a way to not show the button if there's definitely not enough space 🤔 

Would also be great to have a image comparison test that induces mouse hover to draw the copy button in a few different spacing situations (i.e. happy case, too much text, too little space, etc.)

Either way would be great to have @lucasmerlin 's  or @emilk 's eyes on this - this affects a lot of places in the ui and I don't entirely trust my sense of figuring out the consequences here.

PR #11340, opened 2025-09-25
Mark valid data ranges in timeline when loading data via range-limited URL

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/10692
* replaces previous chunk cropping based attempt https://github.com/rerun-io/rerun/pull/11318

### What

https://github.com/user-attachments/assets/eea14be3-6e38-49eb-9ebe-4a91e940380d

Introduces two new concepts that are used when opening dataplatform  URLs with a time range:
#### partial Store

* this is a flag on a store that signals that there's more data to be had
* Used to show "(partial)" in the ui and enable the download-the-rest button
    
#### valid time ranges

* for the moment this is stored as a viewer ephemeral property in the time controls
     * I originally modeled it as part of the recording properties, but I discarded that because it was just too hard to append to given the way we load recordings from the dataplatform right now
     * "good enough for now"
* this is rendered in the timeline as-if it was outside of the data-range + slightly greyed out data-histogram
     *  I tried three different approaches of how to model this inside the timeline drawing structure.  It's tricky because it has to behave correctly with the time-gap detection!
          1) push out more/smaller linear `Segment`s covering only the valid ranges. Drawbacks
			    * hard to make time cursor behave correctly in the invalid data ranges
				* compresses invalid data ranges between two valid ones to nothing since system thinks it's just a gap in the data
          2) introduce sub-ranges on linear `Segment`s (**this is what I went with**). Drawbacks:
                * usually we add a "unselectable" region on gaps that the time cursor can't reach. This is not the case for gaps within a valid range, thus these gaps become slightly different
                * for the same reason invalid-data induced gap zig-zag lines behave meh when zooming out, they can in fact overlap (...feature...?)
          3) make linear `Segment`s be either "valid" or "invalid" and adjust drawing
                * causes weird jumps when playing from valid to invalid parts of the data since these areas are now "real" gaps which the time cursor can't stay on
                * breaks a lot of the previous invariants around `Segment` never touching


crates/store/re_redap_client/src/grpc.rs
```diff
@@ -21,13 +21,16 @@ use crate::{
...
 /// If you're not in a ui context you can safely ignore these.
 pub enum UiCommand {
-    SetLoopSelection {
-        recording_id: re_log_types::StoreId,
-        timeline: re_log_types::Timeline,
-        time_range: re_log_types::AbsoluteTimeRangeF,
+    AddValidTimeRange {
+        store_id: StoreId,
+
+        /// If `None`, signals that all timelines are valid entirely.
```
IsseW:
```suggestion
        /// If `None`, signals that all timelines are entirely valid.
```
maybe?

IsseW:
Just to make sure I understood what the current approach is: The cursor is meant to be playing all the data we received? Even in the grayed out zone?

In the collapsed timeline I think this looks a bit confusing.

Wumpf:
> The cursor is meant to be playing all the data we received? Even in the grayed out zone?

Correct. I wanted it to be possible to click into the "invalid" data ranges. The fact that it also plays everything annoys me a bit as well though. Maybe we should make it so that when playing it stops at the last invalid piece, or even jumps over invalid pieces? I'm not entirely sure what's best tbh. Opinions welcome

IsseW:
I think stopping at the last valid piece would be good, as that matches up with what's last in the collapsed timeline as well.

IsseW:
I feel like this seeking to the next valid range turned out really good!

crates/viewer/re_viewer_context/src/time_control.rs
```diff
@@ -613,6 +638,78 @@ impl TimeControl {
...
+        &mut self,
+        timeline: Option<TimelineName>,
+        time_range: AbsoluteTimeRange,
+    ) {
+        if let Some(timeline) = timeline {
+            match self.valid_time_ranges.entry(timeline) {
+                std::collections::hash_map::Entry::Vacant(entry) => {
+                    entry.insert(Vec1::new(time_range));
+                }
+                std::collections::hash_map::Entry::Occupied(mut entry) => {
```
IsseW, line 662: The logic here *could* be optimized with binary search as well, but as I think you said in a comment it's not important as we won't have a lot of these.

PR #11341, opened 2025-09-25
Add support for `starts with` and `ends with` to string filter

abey79:
### Related

### What

This PR adds `starts_with` and `ends_with` support for string column filtering. To that end, this PR also:
- refactors the string filter to be inline with the rest of the code base
- introduces a dedicated UDF to perform the actual filtering

Note that `FilterKind` increasingly looks like it could be turned into a trait. I might look into it next week.


crates/viewer/re_dataframe_ui/src/filters/filter.rs
```diff
@@ -194,9 +192,11 @@ impl FilterKind {
                 Some(Self::Float(Default::default()))
             }
 
-            DataType::Utf8 | DataType::Utf8View => Some(Self::StringContains(String::new())),
+            data_type if is_supported_string_datatype(data_type) => {
```
Wumpf, line 195: this helps with keeping dupl down, but it does erode this match statement here quite a bit unfortunately :/
not sure if that's actionable, could inline it here (it's internally just a `matches`) but it's an annoying microtradeoff

crates/viewer/re_dataframe_ui/src/filters/filter_ui.rs
```diff
@@ -315,6 +315,7 @@ impl Filter {
 
 impl SyntaxHighlighting for TimestampFormatted<'_, FilterKind> {
     fn syntax_highlight_into(&self, builder: &mut SyntaxHighlightedBuilder) {
+        //TODO(ab): this is weird. this entire impl should be delegated to inner structs
```
Wumpf, line 318: agreed, very much inline with the traitification idea from the PR descr

crates/viewer/re_dataframe_ui/src/filters/filter_ui.rs
```diff
@@ -364,51 +365,22 @@ impl FilterKind {
         column_name: &str,
         popup_just_opened: bool,
     ) -> FilterUiAction {
-        let mut action = FilterUiAction::None;
-
-        let operator_text = self.operator_text();
-
         // Reduce the default width unnecessarily expands the popup width (queries as usually vers
         // small).
         ui.spacing_mut().text_edit_width = 150.0;
```
Wumpf, line 371: neat!

crates/viewer/re_dataframe_ui/tests/filter_tests.rs
```diff
@@ -552,50 +556,52 @@ async fn test_float_lists() {
 #[tokio::test]
 async fn test_string_contains() {
     filter_snapshot!(
-        FilterKind::StringContains(String::new()),
+        FilterKind::String(StringFilter::new(StringOperator::Contains, String::new())),
```
Wumpf, line 559: what's the reason/rationale we have a dedicated `test_string_contains` test but cover starts/ends only with the generic `test_string_list` one?

crates/viewer/re_dataframe_ui/src/filters/string.rs
```diff
@@ -0,0 +1,299 @@
...
+        match self {
+            Self::Contains => "contains".fmt(f),
+            Self::StartsWith => "starts with".fmt(f),
+            Self::EndsWith => "ends with".fmt(f),
+        }
+    }
+}
+
+impl StringOperator {
+    pub const ALL: &'static [Self] = &[Self::Contains, Self::StartsWith, Self::EndsWith];
```
Wumpf, line 40: we use a lot of those around here. Makes me nervous they outdate. Have you checked if we have some minimal procmacro around for this? I know we use `enumset` in some places, but that one seems like overkill. Almost tempted to roll our own just for the `ALL` property

crates/viewer/re_dataframe_ui/src/filters/string.rs
```diff
@@ -0,0 +1,299 @@
...
+pub fn is_supported_string_datatype(data_type: &DataType) -> bool {
+    matches!(
+        data_type,
+        DataType::Utf8 | DataType::Utf8View | DataType::LargeUtf8
+    )
+}
+
+/// Custom UDF for performing filtering on a column of strings, with support for list columns.
+///
+/// This UDF converts both the haystack and the needle to lowercase before performing the queries.
```
Wumpf, line 127: we probably should put the case insensitiveness on a tooltip somewhere? might get cluttery though tbf

crates/viewer/re_dataframe_ui/src/filters/string.rs
```diff
@@ -0,0 +1,299 @@
...
+    fn is_valid_input_type(data_type: &DataType) -> bool {
+        match data_type {
+            DataType::List(field) | DataType::ListView(field) => {
+                // Note: we do not support double nested types
+                is_supported_string_datatype(field.data_type())
+            }
+
+            //TODO(ab): support other containers
+            _ => is_supported_string_datatype(data_type),
+        }
```
Wumpf, line 164: this pattern keeps coming up in UDFs around here with the same todo. Time to unify?

crates/viewer/re_dataframe_ui/src/filters/string.rs
```diff
@@ -0,0 +1,299 @@
...
+    fn invoke_primitive_array(&self, array: &ArrayRef) -> DataFusionResult<BooleanArray> {
+        if !is_supported_string_datatype(array.data_type()) {
+            return exec_err!("Unsupported data type {}", array.data_type());
+        }
+
+        // We need to convert the haystack to lowercase first. We delegate this task to the existing
+        // datafusion `lower` UDF.
+        //
+        // Note that this _must_ happen here (and, e.g., not at the `Expr` level), because `lower`
+        // does not support nested datatypes such as lists.
```
Wumpf, line 176: curious, could there be benefits to do it on the expression level iff it is not nested?

crates/viewer/re_dataframe_ui/src/filters/string.rs
```diff
@@ -0,0 +1,299 @@
...
+                    // `ANY` semantics happening here
+                    Ok(element_results
+                        .iter()
+                        .map(|x| x.unwrap_or(false))
+                        .find(|x| *x)
+                        .unwrap_or(false))
+                })
+            })
+            .map(|x| x.transpose())
+            .collect::<DataFusionResult<BooleanArray>>()
```
Wumpf, line 240: same pattern as on the "all the other filters" UDF. unify? :thinking:

Wumpf (approved):
nice! looks like more and more things fall into place
I am however itching to promote further unification of things, expressed in various comments here. Given the overall state we're in it might be best to disregard those though - zooming out it still feels too early to solidify

overall almost all comments I left are "maybe?"-style, leaving it to you what to do with those ;-)

crates/viewer/re_dataframe_ui/tests/filter_tests.rs
abey79: Not a very strong one but the actual logic of the various string operator is pretty well tested in the list case, so the non-list test case are primarily focused on ensuring we correctly handle all arrow datatypes, so it's not super useful to spam more snapshots files.

I'll add a comment to that effect.

abey79: Or i'll actually improve that test, lessee.

crates/viewer/re_dataframe_ui/src/filters/string.rs
abey79: I have this exact same internal monologue every. single. time. 😓 

On one hand, we already depend on `strum` and there is that: https://docs.rs/strum/latest/strum/derive.VariantArray.html

On the other hand, are we really gonna forget updating those? I actually just did in some follow-up work and immediately ran into it missing, and promptly corrected it. I'm torn.

crates/viewer/re_dataframe_ui/src/filters/string.rs
abey79: It could be a note below the text entry field. All ask Katya what she thinks about that.

crates/viewer/re_dataframe_ui/src/filters/string.rs
abey79: I'll push back on that for now. We surely want dictionary stuff, but it's a bigger problem and will need a bunch of tests etc. so I'll leave it as future work.

crates/viewer/re_dataframe_ui/src/filters/string.rs
Wumpf: variant array does sound perfect. I believe it's worth the cost

crates/viewer/re_dataframe_ui/src/filters/string.rs
abey79: Much the contrary actually.

Datafusion's `Expr` implementation of `contains` is identical to mine (aka it lowers to using arrow's kernel), BUT does not handle the array vs. scalar scenario we have here. So it creates a vector with a duplicated version of the scalar with the same length as the column.

This is obviously bad for performance (not that we wouldn't have other problem in cases where that becomes noticeable).

Relevant slack thread: https://rerunio.slack.com/archives/C04DLS7CCE4/p1758809948817929

crates/viewer/re_dataframe_ui/src/filters/string.rs
abey79: agreed, but later, as I want to rewrite this bit to begin with. I'll add a todo.

abey79: also, part of the traitification, i'll probably split the other UDF in two. So good occasion to unify stuff.

crates/viewer/re_dataframe_ui/src/filters/filter.rs
abey79: yeah I'm not fan either. i'll revisit this when doing the traitification. I took a look at that today, and it should be feasible.

PR #11342, opened 2025-09-25
Fix running pytest in strict mode

emilk:
The pytests would fail if you had `RERUN_STRICT=1` set when running them 🤦 

Not sure what a nicer fix for this is. Global mutable state is… awful.


rerun_py/tests/unit/test_any_values.py
ntjohnson1, line 1: I think we need to explicitly disable strict mode on line 79 too? Anywhere we are checking for RerunWarnings means we assume we had strict mode disabled.

Otherwise LGTM

PR #11343, opened 2025-09-25
web viewer links produced by native viewer now are now best-effort version-pinned

Wumpf:
### Related

* Fixes RR-2413

### What

See title. It's not always perfect, but I think it's still better than before.

### DIFF
--- a/crates/viewer/re_viewer/src/startup_options.rs
+++ b/crates/viewer/re_viewer/src/startup_options.rs
@@ -113,7 +113,29 @@ impl StartupOptions {
         #[cfg(not(target_arch = "wasm32"))]
         {
             // TODO(RR-1878): Would be great to grab this from the dataplatform when available.
-            url::Url::parse("https://rerun.io/viewer").ok()
+            let version = re_build_info::build_info!().version;
+
+            // We generally put `dev` when we're on a version that hasn't been released yet.
+            // Therefore, go one version back!
+            let (major, minor, patch) = if version.is_dev() {
+                if version.patch == 0 {
+                    // There might be a patched version of the latest minor/major, but we don't know that unfortunately.
+                    if version.minor == 0 {
+                        (version.major - 1, 0, 0)
+                    } else {
+                        (version.major, version.minor - 1, 0)
+                    }
+                } else {
+                    (version.major, version.minor, version.patch - 1)
+                }
+            } else {
+                (version.major, version.minor, version.patch)
+            };
+
+            url::Url::parse(&format!(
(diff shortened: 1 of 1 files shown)


crates/viewer/re_viewer/src/startup_options.rs
```diff
@@ -113,7 +113,29 @@ impl StartupOptions {
...
+                } else {
+                    (version.major, version.minor, version.patch - 1)
+                }
+            } else {
+                (version.major, version.minor, version.patch)
+            };
+
+            url::Url::parse(&format!(
+                "https://rerun.io/viewer/version/{major}.{minor}.{patch}"
+            ))
```
emilk: I think this logic would make much more sense as a `impl CrateVersion {` in `crates/build/re_build_info/src/crate_version.rs`

```suggestion
            let version = re_build_info::build_info!().version;
            
            // We generally put `dev` when we're on a version that hasn't been released yet.
            // Therefore, go one version back!
            let version = version.latest_stable();
            
            url::Url::parse(&format!(
                "https://rerun.io/viewer/version/{version}"
            ))
```

emilk (changes requested):

crates/viewer/re_viewer/src/startup_options.rs
Wumpf: even gave it a doctest now!

crates/build/re_build_info/src/crate_version.rs
```diff
@@ -79,6 +79,58 @@ impl PartialOrd for CrateVersion {
...
+
+    /// If this version is stable returns it, otherwise returns the version prior to that.
+    ///
+    /// Doesn't have knowledge of release patched versions, so the returned version will be conservative,
+    /// and not contain any patched versions.
+    /// Similarly, it doesn't know whether a version release was skipped.
+    /// ```
+    /// # use re_build_info::CrateVersion;
+    /// assert_eq!(CrateVersion::parse("0.19.1").latest_stable(), CrateVersion::parse("0.19.1"));
+    /// assert_eq!(CrateVersion::parse("0.19.1-rc.1").latest_stable(), CrateVersion::parse("0.19.1"));
```
emilk: What is the logic for this? If we have just released 0.19.1-rc.1, it means we do NOT have a 0.19.1 final yet

Wumpf: ~right I guess we shouldn't just walk backwards for dev, but for any of dev/rc/alpha~
no, more correct to keep the meta if there's no `dev` flag

PR #11346, opened 2025-09-26
Support registering the table zoo to a local server

abey79:
Improve the table zoo script to support registering on a local server.

Note:
- OSS server doesn't support tables yet
- local only for now (for a remote, we'd need the ability to save the table to some bucket for which we have authorisation, yada)
- moved to `examples` (from `tests`) because there our dependency linting is more lenient, and I didn't want to introduce a "main" dependency on `lancedb`).


timsaucer (approved):
Overall this looks great. Moving it out of tests will lower our coverage, though. I think this will potentially be helpful for the OSS server once we implement the register APIs.

abey79:
> Overall this looks great. Moving it out of tests will lower our coverage, though. I think this will potentially be helpful for the OSS server once we implement the register APIs.

Yeah somehow depending on lance in pixi.toml is inevitable I guess. When that happens, we should probably move this back to `tests` where it really belongs, I agree.

PR #11350, opened 2025-09-26
Fix: enabling `auth` should not pull in all of `re_viewer`

emilk:
Building `rerun-cli` with the `auth` feature should not pull in `winit`, `wgpu`, etc…


IsseW (approved):
👍

PR #11352, opened 2025-09-26
Fix race condition for ui adjustments while loading recordings from redap client

Wumpf:
This caused these ui adjustments like timeline selection etc. to get lost!

Also, we sent those commands for blueprints as well, which didn't make any sense!
(if anything it's actually a bit dangerous since it could have messed with the blueprint's timeline)

### DIFF
--- a/crates/store/re_redap_client/src/grpc.rs
+++ b/crates/store/re_redap_client/src/grpc.rs
@@ -438,7 +438,7 @@ pub async fn stream_blueprint_and_partition_from_server(
             blueprint_partition,
             None,
             re_uri::Fragment::default(),
-            on_ui_cmd.as_deref(),
+            None, // Don't send UI commands for blueprint.
             on_msg.as_deref(),
         )
         .await?;
@@ -483,7 +483,7 @@ pub async fn stream_blueprint_and_partition_from_server(
         partition_id.into(),
         time_range,
         fragment,
-        on_ui_cmd.as_deref(),
+        on_ui_cmd,
         on_msg.as_deref(),
     )
     .await?;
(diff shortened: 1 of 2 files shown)


crates/store/re_redap_client/src/grpc.rs
```diff
@@ -590,6 +574,33 @@ async fn stream_partition_from_server(
...
+                        fragment,
+                    });
+                }
+            })
+            .is_err()
+        {
+            re_log::debug!("Receiver disconnected");
+            return Ok(());
+        }
+    }
```
emilk, line 602: Ooof, I don't like this at all. This is a callback calling a callback. Trying to figure out why the loop selection changed in the middle of receiving messages will not be fun.

I would much prefer that we converted this into data:


```rs
tx.send(LogMsg::UiCommand(UiCommand::SetLoopSelection { … }));
tx.send(LogMsg::UiCommand(UiCommand::SetUrlFragment { … }));
```

emilk (changes requested):

crates/store/re_redap_client/src/grpc.rs
Wumpf: Not saying my solution is pretty, but I'd like to avoid touching `LogMsg` though if possible. We wanted to remove it for a while and that makes it harder

Wumpf: very hairy overall to import that knowledge into `re_log_types`

Wumpf: trying to find a different solution....

Wumpf:
Replaced by https://github.com/rerun-io/rerun/pull/11365

PR #11353, opened 2025-09-26
Enable "Close current recording" only when there's a recording

MichaelGrupp:
Trivial change, but as co-author of the app kittest I also felt obliged to add a test 🙃

### Before

Bildschirmfoto 2025-09-26 um 18 46 54

### Now

https://github.com/user-attachments/assets/e75c89fd-c736-417c-a232-70ea6e100b51

### DIFF
--- a/crates/viewer/re_viewer/tests/app_kittest.rs
+++ b/crates/viewer/re_viewer/tests/app_kittest.rs
@@ -41,6 +41,16 @@ async fn settings_screen() {
     harness.snapshot("settings_screen");
 }
 
+/// Opens the Rerun menu without an active recording and snapshots the app.
+/// Tests that certain recording-related entries are disabled (e.g. save or close recording).
+#[tokio::test]
+async fn menu_without_recording() {
+    let mut harness = viewer_test_utils::viewer_harness();
+    harness.get_by_label("Menu").click();
+    harness.run_ok();
+    harness.snapshot("menu_without_recording");
+}
+
 /// Tests the colormap selector UI with snapshot testing.
 /// This is defined here instead of in `re_viewer/tests` because it depends on `re_test_context`,
 /// which depends on `re_viewer_context`.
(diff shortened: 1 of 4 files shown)


tests/rust/re_integration_test/tests/snapshots/dataset_ui_table.png
MichaelGrupp, line 1: Side effect of deactivating the examples in the test harness. Looks also cleaner IMO

crates/viewer/re_viewer/tests/app_kittest.rs
```diff
@@ -41,6 +41,19 @@ async fn settings_screen() {
...
+/// Opens the Rerun menu without an active recording and snapshots the app.
+/// Tests that certain recording-related entries are disabled (e.g. save or close recording).
+#[tokio::test]
+async fn menu_without_recording() {
+    let mut harness = viewer_test_utils::viewer_harness();
+    harness.get_by_label("Menu").click();
+    harness.run_ok();
+    harness.snapshot_options(
+        "menu_without_recording",
+        &SnapshotOptions::new().threshold(2.0),
```
MichaelGrupp: I was fighting against a diff in two pixels' shade of orange on Mac vs Linux here.... 
grafik

Wumpf: Common mistake: threshhold sets an abstract threshhold value for differences of any pixel, but not the number of broken pixels. 2.0 for `threshold` is actually quite high

```suggestion
        &SnapshotOptions::new().failed_pixel_count_threshold(2.0),
```



@lucasmerlin is working on a better way to deal with these :)

Wumpf:
oops, sorry for that typo.
Looks actually though like there's a massive difference `'menu_without_recording' Image did not match snapshot. Diff: 186977,`
You can download the results & diffs from here https://github.com/rerun-io/rerun/actions/runs/18126797050?pr=11353
menu_without_recording diff
menu_without_recording new
menu_without_recording


I think what's happening is that the ci runs against a hypothetically merged PR, not against where you branched off.
I.e. you have to merge & reupload the image :/
so for some reason the shade is completely different

MichaelGrupp:
Looks like the color change from https://github.com/rerun-io/rerun/pull/11331

Updated snapshot ... ⏳

MichaelGrupp:
Windows build failed on main despite green CI here 😅 This should fix it: #11387

PR #11356, opened 2025-09-27
Add table support to OSS server

timsaucer:
This PR adds support for lance tables in the OSS server, similar to the existing support for Rerun datasets.

Users can add a `--table file_path` option to provide a path to the table location.

Additionally it creates the `__entries` table that is required for using the dataframe APIs.

Unit tests were updated and verified to match the redap server.

This is an enabling feature for ongoing work such as catalog providers where we need to use the OSS server in CI.

### DIFF
--- a/crates/store/re_server/src/store.rs
+++ b/crates/store/re_server/src/store.rs
@@ -509,73 +509,84 @@ impl InMemoryStore {
     }
 }
 
-macro_rules! generate_entries_table {
-    ($fn_name:ident, $field_name:ident, $entry_kind:expr) => {
-        pub fn $fn_name(&self) -> Result<RecordBatch, Error> {
-            let (id, name, entry_kind, created_at, updated_at): (
-                Vec<Tuid>,
-                Vec<String>,
-                Vec<i32>,
-                Vec<i64>,
-                Vec<i64>,
-            ) = self
-                .$field_name
-                .iter()
-                .map(|(key, value)| {
-                    (
-                        key.id,
-                        value.name.clone(),
-                        $entry_kind as i32,
-                        value.created_at.as_nanosecond() as i64,
-                        value.updated_at.as_nanosecond() as i64,
-                    )
-                })
-                .multiunzip();
-
-            let id_arr = id
-                .to_arrow()
-                .map_err(|err| DataFusionError::External(Box::new(err)))?;
-            let name_arr = Arc::new(StringArray::from(name)) as ArrayRef;
-            let kind_arr = Arc::new(Int32Array::from(entry_kind)) as ArrayRef;
-            let created_at_arr = Arc::new(TimestampNanosecondArray::from(created_at)) as ArrayRef;
-            let updated_at_arr = Arc::new(TimestampNanosecondArray::from(updated_at)) as ArrayRef;
-
-            let schema = Arc::new(Schema::new_with_metadata(
-                vec![
-                    Field::new("id", Tuid::arrow_datatype(), false),
-                    Field::new("name", DataType::Utf8, false),
-                    Field::new("entry_kind", DataType::Int32, false),
-                    Field::new(
-                        "created_at",
-                        DataType::Timestamp(TimeUnit::Nanosecond, None),
-                        false,
-                    ),
-                    Field::new(
-                        "updated_at",
-                        DataType::Timestamp(TimeUnit::Nanosecond, None),
-                        false,
-                    ),
-                ],
-                HashMap::new(),
-            ));
-
-            let num_rows = id_arr.len();
-            let rb = RecordBatch::try_new_with_options(
-                schema,
-                vec![id_arr, name_arr, kind_arr, created_at_arr, updated_at_arr],
-                &RecordBatchOptions::default().with_row_count(Some(num_rows)),
+fn generate_entries_table(entries: &[EntryDetails]) -> Result<RecordBatch, Error> {
+    #[allow(clippy::type_complexity)]
+    let (id, name, entry_kind, created_at, updated_at): (
+        Vec<Tuid>,
+        Vec<String>,
+        Vec<i32>,
+        Vec<i64>,
+        Vec<i64>,
+    ) = entries
+        .iter()
+        .map(|entry| {
+            (
+                entry.id.id,
+                entry.name.clone(),
+                entry.kind as i32,
+                entry.created_at.as_nanosecond() as i64,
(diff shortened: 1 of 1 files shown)


abey79:
"Blocked" on https://github.com/rerun-io/rerun/pull/11361 ;)

tests/assets/table/lance/simple_datatypes/_transactions/0-828e81a3-bb94-44a2-bb8a-7cb9ea3e10b2.txn
abey79, line 1: It is maybe worth adding all of these to gitlfs as well, and add all these file extension to the `.gitattributes` file.

crates/store/re_server/src/store.rs
```diff
@@ -251,6 +334,109 @@ impl InMemoryStore {
...
+            Table {
+                id: entries_table_id,
+                name: ENTRIES_TABLE_NAME.to_owned(),
+                provider: entries_table,
+                created_at: prior_entries_table
+                    .map(|t| t.created_at)
+                    .unwrap_or(Timestamp::now()),
+                updated_at: Timestamp::now(),
+                system_table: Some(SystemTable {
+                    kind: SystemTableKind::Entries,
```
abey79, line 440: It feels that having a ad hoc table provider for the entries table would be cleaner, but I guess it involves arc-mutex-ing a few things. So maybe at least have a todo and open an issue?

crates/store/re_server/src/store.rs
```diff
@@ -299,4 +485,100 @@ impl InMemoryStore {
...
+        let entry_id = self.id_by_name.get(name).copied()?;
+        self.table(entry_id)
+    }
+
+    pub fn iter_tables(&self) -> impl Iterator<Item = &Table> {
+        self.tables.values()
+    }
+}
+
+macro_rules! generate_entries_table {
```
abey79: Could this be turned into a regular method? It seems to me that all it'd take is to have  `as_entry_details()` behind a trait for `Table` and `Dataset`.

crates/store/re_server/Cargo.toml
```diff
@@ -40,9 +40,12 @@ re_protos.workspace = true
 anyhow.workspace = true
 arrow.workspace = true
 clap = { workspace = true, features = ["derive", "env"] }
+chrono = { workspace = true, features = ["serde"] }
```
abey79, line 43: please have a `#TODO(#issue)` note here (and related places) so it gets picked up when we check for outdated TODOs

tests/assets/table/.gitattributes
```diff
@@ -0,0 +1,2 @@
+*.lance filter=lfs diff=lfs merge=lfs -text
```
abey79, line 1: *.txn and *.manifest as well?

deny.toml
```diff
@@ -58,6 +59,7 @@ skip = [
   { name = "cargo_metadata" },      # Older version used by ply-rs. It's small, and it's build-time only!
   { name = "core-foundation" },     # Currently, e.g. `webbrowser` and `winit` use different versions.
   { name = "core-graphics-types" }, # wgpu requires 0.2 while winit is still on 0.1
+  { name = "event-listener" },      # remove after https://github.com/lancedb/lance/pull/4834 closes
```
abey79, line 62: again, by using  `#TODO(#issue)` format, it gets picked up by our zombie todo script

crates/store/re_server/src/rerun_cloud.rs
abey79, line 1121: Any chance for a quick stab at this? This PR would love it :)
- https://github.com/rerun-io/rerun/pull/11346

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -995,23 +1122,71 @@ impl RerunCloudService for RerunCloudHandler {
...
+        let Some(entry_id) = request.into_inner().table_id else {
+            return Err(Status::not_found("Table ID not specified in request"));
+        };
+        let entry_id = entry_id.try_into()?;
+
+        let table = store
+            .table(entry_id)
+            .ok_or_else(|| Status::not_found(format!("Entry with ID {entry_id} not found")))?;
+
+        let lance_table = table.provider();
```
abey79: it's not necessarily lance, right?

```suggestion
        let table_provider = table.provider();
```

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -995,23 +1122,71 @@ impl RerunCloudService for RerunCloudHandler {
...
+        let ctx = SessionContext::default();
+        let plan = table
+            .provider()
+            .scan(&ctx.state(), None, &[], None)
+            .await
+            .map_err(|err| Status::internal(format!("failed to scan table: {err:#}")))?;
+
+        let stream = plan
+            .execute(0, ctx.task_ctx())
+            .map_err(|err| tonic::Status::from_error(Box::new(err)))?;
```
abey79, line 1175: taking note of that trick 🙏🏻

abey79 (approved):
Great stuff! I've left a bunch of comments, some of which might be out of scope. I'll let you judge :)

crates/store/re_server/src/store.rs
timsaucer: That is exactly my thought as well, but the scope of this PR was already getting large.

https://github.com/rerun-io/rerun/issues/11369

tests/assets/table/lance/simple_datatypes/_transactions/0-828e81a3-bb94-44a2-bb8a-7cb9ea3e10b2.txn
timsaucer: added

crates/store/re_server/Cargo.toml
timsaucer: added to the `shear` section which will cause this to fail in CI if it isn't updated also

tests/assets/table/.gitattributes
timsaucer: done

deny.toml
timsaucer: I don't think adding an issue to track this one is necessary since CI will fail as soon as the exception is no longer encountered

crates/store/re_server/src/rerun_cloud.rs
timsaucer: Correct. Updated.

PR #11364, opened 2025-09-29
fix formatting and parsing of component paths in links

IsseW:
### Related

Closes RR-2368

### What

Fixes sharing entity + component path.

### TODO

* [x] @rerun-bot full-check

### DIFF
--- a/crates/store/re_log_types/src/path/parse_path.rs
+++ b/crates/store/re_log_types/src/path/parse_path.rs
@@ -42,6 +42,9 @@ pub enum PathParseError {
     #[error("Found trailing hash (#)")]
     TrailingHash,
 
+    #[error("Found trailing at (@)")]
+    TrailingAt,
+
     // Escaping:
     #[error("Unknown escape sequence: \\{0}")]
     UnknownEscapeSequence(char),
@@ -94,7 +97,19 @@ impl std::str::FromStr for DataPath {
                 return Err(PathParseError::TrailingColon);
             }
 
-            let component_type_delimiter = &component_descriptor_tokens
+            let archetype_delimiter = component_descriptor_tokens
+                .iter()
+                .position(|&token| token == "@");
+            let archetype = if let Some(archetype_delimeter) = archetype_delimiter {
+                if let Some(archetype) = component_descriptor_tokens.get(archetype_delimeter + 1) {
+                    Some(archetype.to_owned())
+                } else {
+                    return Err(PathParseError::TrailingAt);
+                }
+            } else {
+                None
+            };
+            let component_type_delimiter = component_descriptor_tokens
                 .iter()
                 .position(|&token| token == "#");
             let component_type = if let Some(component_type_delimiter) = component_type_delimiter {
@@ -109,8 +124,9 @@ impl std::str::FromStr for DataPath {
                 None
             };
 
-            let component_tokens_end =
-                component_type_delimiter.unwrap_or(component_descriptor_tokens.len());
+            let component_tokens_end = archetype_delimiter
+                .or(component_type_delimiter)
+                .unwrap_or(component_descriptor_tokens.len());
             if component_descriptor_tokens.get(component_tokens_end - 1) == Some(&":") {
                 return Err(PathParseError::TrailingColon);
             }
@@ -119,7 +135,7 @@ impl std::str::FromStr for DataPath {
 
             component_descriptor = Some(ComponentDescriptor {
                 component: field.into(),
-                archetype: None,
+                archetype: archetype.map(Into::into),
                 component_type: component_type.map(Into::into),
             });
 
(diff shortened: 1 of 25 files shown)


Wumpf (commented):
I'm on the fence of landing this, not because of any problem with the impl here, but because `DataPath` using `component_descriptor` in the first place is the actual problem here 💡 
When we switched to component names being the only relevant required/unique/significant part of addressing a component, we should have changed this according. So this is essentially running into a piece of tech debt.

cc: @grtlr

Wumpf:
to clarify, the proposal is for `DataPath` to not use component descriptor but merely a component ~name~ identifier

crates/store/re_types_core/src/component_descriptor.rs
```diff
@@ -85,6 +85,22 @@ impl ComponentDescriptor {
...
+            f.write_str("@")?;
+            f.write_str(&archetype)?;
+        }
+        if let Some(ty) = self.component_type {
+            f.write_str("#")?;
+            f.write_str(&ty)?;
+        }
+
+        Ok(())
+    }
```
Wumpf: do we still need this now? I'm actually not sure about the exact convention we should go for here since rerun-component-identifiers also embed the archetype name.
Plus, starting with a `:` is strange

IsseW: Ah yeah this can be removed now, will do.

> Plus, starting with a `:` is strange

Yeah, without context it definitely is. The reason is that when parsing a datapath it expects the component to start with a `:` to separate it from entity path.

crates/store/re_log_types/src/path/parse_path.rs
```diff
@@ -31,17 +31,14 @@ pub enum PathParseError {
     UnexpectedInstance(Instance),
 
     #[error("Found an unexpected trailing component descriptor: {0:?}")]
-    UnexpectedComponentDescriptor(Box<ComponentDescriptor>),
+    UnexpectedComponent(Box<ComponentIdentifier>),
```
grtlr: Just curious: Why was this boxed in the first place? Should we get rid of it?

crates/store/re_log_types/src/path/parse_path.rs
```diff
@@ -84,46 +81,29 @@ impl std::str::FromStr for DataPath {
 
         let mut tokens = tokenize_data_path(path);
 
-        let mut component_descriptor = None;
+        let mut component = None;
         let mut instance = None;
 
         // Parse `:Points3D:Color#colors` suffix:
```
grtlr: This comment looks out of date.

crates/store/re_log_types/src/path/parse_path.rs
```diff
@@ -424,124 +405,119 @@ mod tests {
...
-        assert_eq!(
-            ComponentPath::from_str("world/points:Points3D:colors#my.custom.colors"),
-            Ok(ComponentPath {
-                entity_path: EntityPath::from("world/points"),
-                component_descriptor: ComponentDescriptor::partial("Points3D:colors")
-                    .with_component_type("my.custom.colors".into()),
-            })
-        );
+    #[track_caller]
+    fn test_parse_ok<T: FromStr<Err = PathParseError> + Display + PartialEq + Debug>(
```
grtlr, line 411: These roundtrip tests are very nice!

crates/utils/re_smart_channel/src/lib.rs
```diff
@@ -30,7 +30,6 @@ pub enum FlushError {
 /// Identifies in what context this smart channel was created, and who/what is holding its
 /// receiving end.
 #[derive(Clone, Debug, PartialEq, Eq, Hash, serde::Deserialize, serde::Serialize)]
-#[cfg_attr(not(target_arch = "wasm32"), expect(clippy::large_enum_variant))]
```
grtlr, line 33: Feels so good when that happens 😁

crates/viewer/re_data_ui/src/component_path.rs
```diff
@@ -15,16 +15,26 @@ impl DataUi for ComponentPath {
...
 
+        let Some(component_descriptor) = engine
+            .store()
+            .entity_component_descriptor(entity_path, *component)
+        else {
+            ui.label(format!(
+                "Entity {entity_path:?} has no component {component:?}"
+            ));
+            return;
+        };
```
grtlr, line 31: I guess these `else` clauses are there to mostly handle broken links that were shared?

We could use `ui.warning_label(...)` to draw more attention.

crates/viewer/re_data_ui/src/component_path.rs
```diff
@@ -34,12 +44,12 @@ impl DataUi for ComponentPath {
...
                 &query.timeline(),
                 entity_path,
-                component_descriptor,
+                &component_descriptor,
             ) {
                 ui.label("<unset>");
             } else {
                 ui.label(format!(
-                    "Entity {entity_path:?} has no component {component_descriptor:?}"
+                    "Entity {entity_path:?} has no component {component:?}"
```
grtlr, line 52: Same as above.

grtlr (approved):

grtlr:
@rerun-bot full-check

crates/store/re_log_types/src/path/parse_path.rs
IsseW: Ah probably, I'm guessing it was boxed because we have compile time asserts that error enums are less than a certain size. ✅

crates/viewer/re_data_ui/src/component_path.rs
IsseW: > I guess these else clauses are there to mostly handle broken links that were shared?

Yep! And looks a lot better with warning label

crates/store/re_log_types/src/path/parse_path.rs
IsseW: Updated

IsseW:
Full build passed, so think this should be good to go now

PR #11365, opened 2025-09-29
Fix race condition for ui adjustments while loading recordings from redap client

Wumpf:
### Related

* DNM: https://github.com/rerun-io/rerun/pull/11362
* Replaces previous attempt: https://github.com/rerun-io/rerun/pull/11352
* Fixes RR-2439

### What

Title is only for the changelog and the immediate outcome, but  the real change here is much wider:
Introduces a new `DataSourceMessage` that wraps `LogMsg` and the previously out-of-band sent `re_redap_client::UiCommand`. This allows data sources to send ui commands that have to be ordered relative to other data.

Refactor has been done in a minimal way so far, but I think we should as next step unify `TableMsg` into as well.

### DIFF
--- a/crates/store/re_log_types/src/data_source_message.rs
+++ b/crates/store/re_log_types/src/data_source_message.rs
@@ -39,11 +39,11 @@ pub enum DataSourceUiCommand {
         time_range: AbsoluteTimeRange,
     },
 
-    /// Navigate to time/entities/anchors/etc. that are set in a [`re_uri::Fragment`].
+    /// Navigate to time/entities/anchors/etc. that are set in a `re_uri::Fragment`.
     SetUrlFragment {
         store_id: StoreId,
 
-        /// Uri fragment, see [`re_uri::Fragment`] on how to parse it.
+        /// Uri fragment, see `re_uri::Fragment` on how to parse it.
         // Not using `re_uri::Fragment` to avoid further dependency entanglement.
         fragment: String, //re_uri::Fragment,
     },
--- a/crates/store/re_log_types/src/data_source_message.rs
+++ b/crates/store/re_log_types/src/data_source_message.rs
@@ -1,0 +1,54 @@
+// TODO(andreas): Conceptually these should go to `re_data_source`.
+// However, `re_data_source` depends on everything that _implements_ a datasource, therefore we would get a circular dependency!
+
+use crate::{AbsoluteTimeRange, LogMsg, StoreId, TimelineName};
+
+/// Message from a data source.
+///
+/// May contain limited UI commands for instrumenting the state of the receiving end.
+#[derive(Clone, Debug)]
+pub enum DataSourceMessage {
+    /// See [`LogMsg`].
+    LogMsg(LogMsg),
+
+    /// A UI command that has to be sorted relative to `LogMsg`s.
+    ///
+    /// Non-ui receivers can safely ignore these.
+    UiCommand(DataSourceUiCommand),
+}
+
+impl From<LogMsg> for DataSourceMessage {
+    #[inline]
+    fn from(msg: LogMsg) -> Self {
+        Self::LogMsg(msg)
+    }
+}
+
+impl From<DataSourceUiCommand> for DataSourceMessage {
+    #[inline]
+    fn from(cmd: DataSourceUiCommand) -> Self {
+        Self::UiCommand(cmd)
+    }
+}
+
+/// UI commands issued when streaming in datasets.
+///
+/// If you're not in a ui context you can safely ignore these.
+#[derive(Clone, Debug)]
+pub enum DataSourceUiCommand {
+    AddValidTimeRange {
+        store_id: StoreId,
+
+        /// If `None`, signals that all timelines are entirely valid.
+        timeline: Option<TimelineName>,
+        time_range: AbsoluteTimeRange,
+    },
+
+    SetUrlFragment {
+        store_id: StoreId,
+
+        /// Uri fragment, see [`re_uri::Fragment`] on how to parse it.
+        // Not using `re_uri::Fragment` to avoid further dependency entanglement.
+        fragment: String, //re_uri::Fragment,
+    },
+}
--- a/crates/store/re_log_types/src/data_source_message.rs
+++ b/crates/store/re_log_types/src/data_source_message.rs
@@ -1,7 +1,7 @@
 // TODO(andreas): Conceptually these should go to `re_data_source`.
 // However, `re_data_source` depends on everything that _implements_ a datasource, therefore we would get a circular dependency!
 
-use crate::{AbsoluteTimeRange, LogMsg, StoreId, TimelineName};
+use crate::{AbsoluteTimeRange, LogMsg, StoreId, TimelineName, impl_into_enum};
 
 /// Message from a data source.
 ///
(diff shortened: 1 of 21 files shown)


crates/store/re_log_types/src/data_source_message.rs
```diff
@@ -0,0 +1,54 @@
...
+
+/// Message from a data source.
+///
+/// May contain limited UI commands for instrumenting the state of the receiving end.
+#[derive(Clone, Debug)]
+pub enum DataSourceMessage {
+    /// See [`LogMsg`].
+    LogMsg(LogMsg),
+
+    /// A UI command that has to be sorted relative to `LogMsg`s.
```
emilk:
```suggestion
    /// A UI command that has to be ordered relative to `LogMsg`s.
```

crates/store/re_log_types/src/data_source_message.rs
```diff
@@ -0,0 +1,54 @@
...
+/// If you're not in a ui context you can safely ignore these.
+#[derive(Clone, Debug)]
+pub enum DataSourceUiCommand {
+    AddValidTimeRange {
+        store_id: StoreId,
+
+        /// If `None`, signals that all timelines are entirely valid.
+        timeline: Option<TimelineName>,
+        time_range: AbsoluteTimeRange,
+    },
```
emilk, line 40: Docstring `AddValidTimeRange`!

Also, I suggest breaking out as `struct AddValidTimeRange`

crates/store/re_log_types/src/data_source_message.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    SetUrlFragment {
+        store_id: StoreId,
+
+        /// Uri fragment, see [`re_uri::Fragment`] on how to parse it.
+        // Not using `re_uri::Fragment` to avoid further dependency entanglement.
+        fragment: String, //re_uri::Fragment,
+    },
```
emilk, line 49: same

crates/store/re_log_types/src/data_source_message.rs
```diff
@@ -0,0 +1,54 @@
+// TODO(andreas): Conceptually these should go to `re_data_source`.
+// However, `re_data_source` depends on everything that _implements_ a datasource, therefore we would get a circular dependency!
```
emilk, line 2: We should fix that 🤔

crates/store/re_log_types/src/data_source_message.rs
```diff
@@ -0,0 +1,54 @@
...
+        timeline: Option<TimelineName>,
+        time_range: AbsoluteTimeRange,
+    },
+
+    SetUrlFragment {
+        store_id: StoreId,
+
+        /// Uri fragment, see [`re_uri::Fragment`] on how to parse it.
+        // Not using `re_uri::Fragment` to avoid further dependency entanglement.
+        fragment: String, //re_uri::Fragment,
```
emilk, line 48: sad :(

crates/store/re_redap_client/src/grpc.rs
```diff
@@ -517,39 +496,55 @@ async fn stream_partition_from_server(
...
-        .send(LogMsg::SetStoreInfo(SetStoreInfo {
-            row_id: *re_chunk::RowId::new(),
-            info: store_info,
-        }))
+        .send(
+            LogMsg::SetStoreInfo(SetStoreInfo {
+                row_id: *re_chunk::RowId::new(),
+                info: store_info,
+            })
+            .into(),
```
emilk, line 504: I believe this should work too

```suggestion
            SetStoreInfo {
                row_id: *re_chunk::RowId::new(),
                info: store_info,
            }
            .into(),
```

crates/store/re_redap_client/src/grpc.rs
```diff
@@ -517,39 +496,55 @@ async fn stream_partition_from_server(
...
-                fragment,
-            });
+            if tx
+                .send(
+                    DataSourceUiCommand::SetUrlFragment {
+                        store_id: store_id.clone(),
+                        fragment: fragment.to_string(),
+                    }
+                    .into(),
+                )
```
emilk, line 542: style: we could make `send` take an `impl Into<…>` 🤷 

```suggestion
                .send(
                    DataSourceUiCommand::SetUrlFragment {
                        store_id: store_id.clone(),
                        fragment: fragment.to_string(),
                    }
                )
```

crates/top/rerun/src/commands/entrypoint.rs
```diff
@@ -1216,16 +1225,20 @@ fn assert_receive_into_entity_db(
...
-                            re_log_types::StoreKind::Recording => rec.get_or_insert_with(|| {
-                                re_entity_db::EntityDb::new(msg.store_id().clone())
-                            }),
-                            re_log_types::StoreKind::Blueprint => bp.get_or_insert_with(|| {
-                                re_entity_db::EntityDb::new(msg.store_id().clone())
-                            }),
-                        };
-
-                        mut_db.add(&msg)?;
+                        if let DataSourceMessage::LogMsg(msg) = msg {
```
emilk: an `if-let` without an `else` makes me nervous, especially in a test. I rather have a full match, with a comment explaining why some branches are ignored

crates/viewer/re_viewer/src/app.rs
```diff
@@ -1994,168 +1975,222 @@ impl App {
...
+                    self.receive_log_msg(&msg, store_hub, egui_ctx, &channel_source);
+                }
 
-            if store_hub.is_active_blueprint(store_id) {
-                // TODO(#5514): handle loading of active blueprints.
-                re_log::warn_once!(
-                    "Loading a blueprint {store_id:?} that is active. See https://github.com/rerun-io/rerun/issues/5514 for details."
-                );
+                DataSourceMessage::UiCommand(ui_command) => {
+                    self.receive_data_sourceui_command(ui_command, &channel_source);
```
emilk: I think this will schedule the commands to be executed at the end of the frame. Maybe we should add a call to `run_pending_system_commands` directly after `receive_messages` to avoid a frame delay?

crates/viewer/re_viewer/src/app.rs
```diff
@@ -1994,168 +1975,222 @@ impl App {
...
+            if let Some(event_dispatcher) = self.event_dispatcher.as_ref() {
+                event_dispatcher.on_recording_open(entity_db);
+            }
+        }
+    }
 
-                if let Some(event_dispatcher) = self.event_dispatcher.as_ref() {
-                    event_dispatcher.on_recording_open(entity_db);
-                }
+    fn receive_data_sourceui_command(
```
```suggestion
    fn receive_data_source_ui_command(
```

crates/store/re_log_types/src/data_source_message.rs
emilk: Just to explain: "sorted" and "ordered" means different things. All lists are ordered, but only some of them are sorted. Sets are never ordered (but a BTreeSet is sorted!).

crates/store/re_log_types/src/data_source_message.rs
Wumpf: yeah, I prodded a bit into that, but it's almost more about the question of "what is `re_data_source`" than anything else. If `re_data_source` provides an enum (`DataSource`) that unifies all the possible ways to implement `DataSource`, then this is what follows from it. I suppose if it was as more dynamic structure, a set of rules (a trait?) what _makes_ something a datasource, then it would be different!

crates/store/re_log_types/src/data_source_message.rs
Wumpf: very good point, makes sense!

crates/store/re_redap_client/src/grpc.rs
Wumpf: `into` can't do double-into, requires me to do another layer. But have to do that anyways now because of structification of enum variants
EDIT: chaning intos is actually really annoying. Can't figure it out quickly enough atm.

crates/store/re_log_types/src/data_source_message.rs
Wumpf: went with splitting out as a struct for a bit but decided against now, it just makes all the existing usages worse as there's no such thing as double-into: there's no way to generically implement an `into`/`from` on the split out struct such that it produces `DataSourceMessage` directly

crates/store/re_log_types/src/data_source_message.rs
Wumpf: not as sad as more dependencies imho
(didn't even look if that one is circular tbh; even if not it's not worth it imho)

crates/store/re_redap_client/src/grpc.rs
Wumpf: my gut feeling says this isn't a good place for more implicit conversions; hasn't been annoying enough yet to justify it. Leaving it be

crates/viewer/re_viewer/src/app.rs
Wumpf: good idea!

PR #11367, opened 2025-09-29
Add reflection based support for deserializing ROS2 MCAP

oxkitsune:
### What

Adds support for dynamically deserializing ROS2 MCAP messages into arrow data.

This is done by parsing the ROS2 message definition and using the type definition to deserialize the CDR-encoded data into arrow types.

TODO:
- [x] Remove extra data logged in ROS2 semantic parsers

### DIFF
--- a/tests/assets/rrd/snippets/howto/load_mcap.rrd
+++ b/tests/assets/rrd/snippets/howto/load_mcap.rrd
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:c8f6ee32d3d1833719f437a11bffbe46a3e31fcbb598ae9b7ac9eb78c81eb286
-size 19248664
+oid sha256:5d28084bca1c5d555924a09122fc1a9410f71e649caa99867c0252845818cefb
+size 19248634


jprochazk:
Is this the right "spec" for ROS2? (do we care about ROS1) https://docs.ros.org/en/kilted/Concepts/Basic/About-Interfaces.html#interfaces

oxkitsune:
> Is this the right "spec" for ROS2? (do we care about ROS1) docs.ros.org/en/kilted/Concepts/Basic/About-Interfaces.html#interfaces

This implements the spec for ROS2 .msg messages, which is only slightly different from the ROS1 message spec. Adding support for ROS1 in the future is trivial with this in place.

This does _not_ implement parsing of ROS2 .idl message definitions

ARCHITECTURE.md
```diff
@@ -233,6 +233,7 @@ Update instructions:
 | re_mcap            | Convert MCAP into Rerun-compatible data.                                             |
 | re_memory          | Run-time memory tracking and profiling.                                              |
 | re_perf_telemetry  | In and out of process performance profiling utilities for Rerun & Redap              |
+| re_ros_msg         | Parsing and deserializing ROS   messages                                             |
```
grtlr, line 236: Looks like we still need to update the Architecture figma (see component above table).

crates/utils/re_mcap/src/layers/ros2_reflection.rs
```diff
@@ -0,0 +1,524 @@
...
+            let matching_spec = find_matching_message_spec(schema, message_fields);
+
+            if let Some(spec) = matching_spec {
+                // Use the specification field order to iterate through struct builder fields
+                for (i, spec_field) in spec.fields.iter().enumerate() {
+                    if let Some(field_builder) = struct_builder.field_builders_mut().get_mut(i) {
+                        if let Some(field_value) = message_fields.get(&spec_field.name) {
+                            append_value(field_builder, field_value, schema)?;
+                        } else {
+                            //TODO(gijsd): Field is missing in the message, append null
```
grtlr: Do we still need that comment?

crates/utils/re_mcap/src/layers/ros2_reflection.rs
```diff
@@ -0,0 +1,524 @@
...
+        Value::F64(x) => downcast_builder::<Float64Builder>(builder)?.append_value(*x),
+        Value::String(x) => {
+            downcast_builder::<StringBuilder>(builder)?.append_value(x.clone());
+        }
+        Value::Message(message_fields) => {
+            let struct_builder = downcast_builder::<StructBuilder>(builder)?;
+
+            // For nested messages, we need to find the matching specification from dependencies
+            // Since we don't have type information here, we'll try to match by field names
+            let matching_spec = find_matching_message_spec(schema, message_fields);
```
grtlr: This is quadratic over the number of dependencies. `append_value` is called a lot, does it make sense to cache / precompute all specs when creating the parser? `append_value` could then take a `&self`?

grtlr (commented):
I've reviewed the interface of the reflection-based parser and `re_mcap` and that part LGTM. This is so cool to have!

Please let me know if you want a more detailed review of the parser too.

crates/utils/re_mcap/src/layers/ros2_reflection.rs
oxkitsune: Nope, deserialization will fail.

crates/utils/re_mcap/src/layers/ros2_reflection.rs
oxkitsune: Fixed it by wrapping `StructBuilder` and including the message spec in there.

oxkitsune:
@rerun-bot full-check

PR #11371, opened 2025-09-29
Add `is not` to nullable boolean filter

abey79:
### Related

* part of https://linear.app/rerun/issue/RR-2409/add-not-versions-to-existing-filters

### What

Add `is not` to nullable boolean filter. (Obviously, adding that to non-nullable boolean filters would make no sense as the only two choices are `true` and `false`).

DNM: chained

popup_ui_nullable_boolean_not_equals_true


Wumpf (approved):
seems straight forward 👍

PR #11372, opened 2025-09-29
Treat`!=` filtering of numerical column as the inverse of `==` (aka. outer-NOT and ALL semantics)

abey79:
### Related

* closes of https://github.com/rerun-io/rerun/issues/10560
* closes https://linear.app/rerun/issue/RR-2409/add-not-versions-to-existing-filters

### What

This PR introduces a special-casing of the `!=` operator in numerical column filtering. Contrary to all other operator, this one has "outer-NOT" semantics compared to `==`, which means that:
- for list, the ALL semantic applies instead of the ANY
- the `==` and `!=` filters are complementary (e.g. their results are disjoint and, together, exhaustive)
- (note that this is not true for other pairs such as `<` and `>=`, since both have ANY semantics and, e.g., filter out nulls)

This may appears somewhat unexpected, but it is consistent with the NOT operators recently introduced in other column types, and is an arguably less surprising behaviour (e.g. before this PR, `!= 2` would allow a row with `[2, 3]`).

DNM: chained

### DIFF
--- a/crates/viewer/re_dataframe_ui/src/filters/numerical.rs
+++ b/crates/viewer/re_dataframe_ui/src/filters/numerical.rs
@@ -149,9 +149,9 @@ impl IntFilter {
 
         // Consistent with other column types, we treat `Ne` as an outer-NOT, so we applies it here
         // while the UDF handles `Ne` and `Eq` in the same way (see `ComparisonOperator::apply`).
-        let apply_any_or_null_semantics = self.operator == ComparisonOperator::Ne;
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
thankful that this confusing thing is a small change. I thiiiink I got it 😄 . Helps that I just came from the other ones which did outer-not in the exact same way.

Not sure I'd call it `any_or_null_semantics`, confused me at first along the lines of why the answer to an or question is true/false. As of writing I'm a bit to fried to think of something better.

PR #11375, opened 2025-09-29
CI: run bash with '-euo pipefail'

emilk:
I noticed some warnings when running the release job:

image

Leading to missing words in the release PR description:

https://github.com/rerun-io/rerun/pull/11374

Check the commit messages for details

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -44,14 +44,14 @@ runs:
       if: ${{ runner.os == 'macOS' }}
       run: |
         echo "$(brew --prefix llvm@18)/bin" >> $GITHUB_PATH
-      shell: bash
+      shell: bash -euo pipefail
 
     - name: Check clang version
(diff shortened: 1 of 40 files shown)


emilk:
@rerun-bot full-check

thz (approved):
❤️

jprochazk:
I have a feeling that this might break something. We have many multi-step bash scripts, where parts of them may _not_ want `-e` (fail-fast) behavior, because they're checking for failure or something. For example:

https://github.com/rerun-io/rerun/blob/21753bbae37b7593e7c2423a4c23b74419492701/.github/workflows/release.yml#L184-L190

(I have no idea if doing things this way makes sense - that was just the approach I chose when writing that code :sweat_smile:)

It would be good to at least spot-check all instances of `run:` in the repository to see if there's anything that could depend on this.

emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

PR #11376, opened 2025-09-29
Change colors of arrow-rs syntax highlighting

gavrelina:
align colors brightness, so that neither of them popup too much, but they look more in synch with each other and in tables.


Wumpf:
@gavrelina you need to update the screenshot tests for a change like this :)

To do so, you can run `pixi run rs-update-snapshot-tests`. This is expected to take quiiiite a while, especially when it hasn't been run yet. It essentially runs all Rust tests and updates any kind of snapshot file if there's a change instead of failing on discrepancies.

(as I understand Lucas is working on making it possible to do so from failed ci runs, but as of now you have to do this locally)

PR #11377, opened 2025-09-29
Share button on notebook no longer uses the current base url for web viewer urls

Wumpf:
### Related

* Fixes RR-2414

### What

Previously, web viewer links produced by notebooks would always be incorrect as they'd treat the current url like a viewer application, which is practically always incorrect.

Fixed:
image

Better exposed the (renamed) `viewer_base_url` property for this. This was introduced after the last release, so not a breaking change.

### DIFF
--- a/rerun_js/web-viewer/index.ts
+++ b/rerun_js/web-viewer/index.ts
@@ -101,6 +101,12 @@ export interface WebViewerOptions {
    * enclosing notebook environment, it should be used to set the fallback token.
    */
   fallback_token?: string;
+
+  /** The url that's used when sharing web viewer urls.
+   *
+   * If not set, the viewer will use the current base url.
+   */
+  viewer_base_url?: string;
 }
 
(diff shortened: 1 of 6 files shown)


rerun_js/web-viewer/index.ts
```diff
@@ -101,6 +101,12 @@ export interface WebViewerOptions {
    * enclosing notebook environment, it should be used to set the fallback token.
    */
   fallback_token?: string;
+
+  /** The url that's used when sharing web viewer urls.
+   *
+   * If not set, the viewer will use the current base url.
+   */
+  viewer_base_url?: string;
```
jprochazk, line 109: This will need a landing PR, see https://github.com/rerun-io/landing/pull/1102

Wumpf:
Reconsidering this. Viewer has a broken (?) `is_in_notebook` property that should be used instead.

Wumpf:
replaced by https://github.com/rerun-io/rerun/pull/11379

PR #11378, opened 2025-09-30
Add context menu button to copy partition name

IsseW:
### Related

Closes RR-2412

### What

Adds context menu button to partitions and datasets to copy their names. And adds the same context menu as partitions have to loading partitions.

### DIFF
--- a/crates/viewer/re_recording_panel/src/recording_panel_ui.rs
+++ b/crates/viewer/re_recording_panel/src/recording_panel_ui.rs
@@ -406,6 +406,11 @@ fn dataset_entry_ui(
             ctx.command_sender()
                 .send_system(SystemCommand::CopyViewerUrl(url));
         }
+
+        if ui.button("Copy dataset name").clicked() {
+            re_log::info!("Copied {name:?} to clipboard");
+            ui.ctx().copy_text(name.clone());
+        }
     });
 
     if item_response.clicked() {
@@ -570,13 +575,13 @@ fn receiver_ui(
     receiver: &SmartChannelSource,
     show_hierarchal: bool,
 ) {
-    let Some(string) = receiver.loading_name() else {
+    let Some(name) = receiver.loading_name() else {
         return;
     };
 
     let selected = ctx.is_selected_or_loading(&Item::DataSource(receiver.clone()));
 
-    let label_content = re_ui::list_item::LabelContent::new(string)
+    let label_content = re_ui::list_item::LabelContent::new(&name)
         .with_icon_fn(|ui, rect, _| {
             ui.put(rect, egui::Spinner::new());
         })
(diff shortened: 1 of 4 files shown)


crates/viewer/re_recording_panel/src/recording_panel_ui.rs
```diff
@@ -590,13 +595,31 @@ fn receiver_ui(
...
-            .show_flat(ui, label_content);
-    }
+            .show_flat(ui, label_content)
+    };
+
+    response.context_menu(|ui| {
+        let url = ViewerOpenUrl::from_data_source(receiver).and_then(|url| url.sharable_url(None));
+        if ui
+            .add_enabled(url.is_ok(), egui::Button::new("Copy link to partition"))
+            .on_disabled_hover_text("Can't copy a link to this partition")
```
Wumpf: Should we print a reason? 🤔 
I forget whether that was a well printable error or not

Wumpf (approved):
lgtm!

crates/viewer/re_recording_panel/src/recording_panel_ui.rs
```diff
@@ -406,6 +406,11 @@ fn dataset_entry_ui(
             ctx.command_sender()
                 .send_system(SystemCommand::CopyViewerUrl(url));
         }
+
+        if ui.button("Copy dataset name").clicked() {
+            re_log::info!("Copied {name:?} to clipboard");
+            ui.ctx().copy_text(name.clone());
+        }
```
Wumpf, line 413: I literally ran into it _now_ that I needed to copy a dataset _id_ rather than name and I couldn't do it from the web build linked by this PR.
That risk making it cluttery, but can we please also add that here?

Wumpf:
partially offtopic to this PR, but we're strangle inconsistent of what type of URL we copy where. So for instanace this button
image
gives me without viewer URL

But these give me it with it
image
image


Tbh I'm not even sure which ones I expect where :/
We've recently err'ed on the side of adding viewer URLs because (unless they're totally wrong) they're usually more flexible than not 🤔. Maybe that should just be our policy for everything that isn't obviously a WYSIWYG text copy

crates/viewer/re_recording_panel/src/recording_panel_ui.rs
IsseW: We probably can, I'll try it

Wumpf:
thanks!

IsseW:
> We've recently err'ed on the side of adding viewer URLs because (unless they're totally wrong) they're usually more flexible than not 🤔. Maybe that should just be our policy for everything that isn't obviously a WYSIWYG text copy

I think it's more that recently we've started adding viewer url if we're on web. But we should have this be consistent. Changed it to be the same in the title now.

PR #11379, opened 2025-09-30
Share button on notebook no longer uses the current base url for web viewer urls

Wumpf:
### Related

* Much improved version of https://github.com/rerun-io/rerun/pull/11377
* Fixes https://linear.app/rerun/issue/RR-2414/link-sharing-is-strange-from-rerun-cloud-in-notebook

### What

Also, fixes the Viewer reporting "have you tired native" on notebooks. Turns out our settings transfer from JS/TS to Rust was a bit in a dire state and didn't transmitt `notebook`

Once fixed, it was easy to have `notebook` mean different web viewer base urls by default.

* [x] needs a `landing` PR to update due to renamed parameter
    * https://github.com/rerun-io/landing/pull/1105

### DIFF
--- a/rerun_notebook/src/js/widget.ts
+++ b/rerun_notebook/src/js/widget.ts
@@ -26,11 +26,7 @@ interface WidgetModel {
 
 type Opt<T> = T | null | undefined;
 
-function _resize(
-  el: HTMLElement,
-  width: number | string,
-  height: number | string,
-) {
+function _resize(el: HTMLElement, width: number | string, height: number | string) {
   const style = el.style;
 
   if (typeof width === "string" && width === "auto") {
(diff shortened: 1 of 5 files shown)


rerun_notebook/src/js/widget.ts
```diff
@@ -26,7 +26,11 @@ interface WidgetModel {
 
 type Opt<T> = T | null | undefined;
 
-function _resize(el: HTMLElement, width: number | string, height: number | string) {
+function _resize(
```
jprochazk: Weird how this passed `prettier --check` before, and still does now

Wumpf: ah sorry I actually didn't want to check in all that clutter and only the relevant changes

jprochazk (approved):

PR #11381, opened 2025-09-30
Fix resizing loading screen

IsseW:
### Related

Closes RR-2417

### What

Fixes resizing loading screen.


Wumpf (approved):
tried it. works.
black egui magic!

PR #11278, opened 2025-09-19
Format `RecordBatch` with redactions for non-deterministic values

grtlr:
### What

This redacts the following values when pretty-printing a `RecordBatch`:

* Metadata values (`sorbet:version`, `rerun:id`, `heap_size_bytes`)
* `RowId`s

This can be useful for snapshot testing.

PR #11383, opened 2025-09-30
Add `-` formatting option for redacted values

grtlr:
### Related

* Follow-up of #11278.

### What

This PR allows you to use the minus sign (`-`) formatting option in most `RecordBatch`-derived structs to get a string that has non-deterministic values redacted.

Usage:

```rs
let chunk = Chunk::new(/* ... */);
println!("{chunk:-}");
```

Using `#` instead of `-` would be an alternative, but I think it's better to reserve this for a more meaningful representation.


CODE_STYLE.md
```diff
@@ -131,6 +131,12 @@ Use debug-formatting (`{:?}`) when logging strings in logs and error messages. T
 
 Use `re_error::format(err)` when displaying an error.
 
+We make extensive use of snapshot testing. To work around non-deterministic values, such as TUIDs (time-prefixed unique IDs), many types (should) offer `std::fmt::Display` implementations with redactions that can be access via an overloaded `-` formatting option:
+
+```rs
+println!("{:-}, value"); // The `-` option stands for redaction.
+```
+
```
emilk, line 139:
```suggestion

Look for `f.sign_minus()` in the code for where we handle this.
```

PR #11384, opened 2025-09-30
Py-SDK: More kw-args

emilk:
Using a lot of positional arguments have many downsides. One is readability, the other is future-proofing. If we want the ability to change our interfaces in the future, we need to force users to _name_ the arguments.

I've written a migration guide.

None of our own examples or snippets needed updating, so I believe the impact on users will be rather small.

But this is really on of those "Best time to do it was three years ago, next best time is today"


docs/content/reference/migration/migration-0-26.md
```diff
@@ -10,6 +10,17 @@ For non-blocking, use `timeout_sec=0`.
 Mostly you can just call `.flush()` with no arguments.
 That will block until all writes either finishes or an error occurs (e.g. the gRPC connection is severed).
 
+## Python SDK: `ImageFormat` now uses kw-args for all optional ars
```
oxkitsune:
```suggestion
## Python SDK: `ImageFormat` now uses kw-args for all optional args
```

docs/content/reference/migration/migration-0-26.md
```diff
@@ -10,6 +10,26 @@ For non-blocking, use `timeout_sec=0`.
...
+Before:
+```py
+rr.ImageFormat(width, height, "YUV420")
+
+blueprint.spawn("my_app", 1234)
+```
+
+After:
+```py
+rr.ImageFormat(width, height, pixel_format="YUV420"
```
emilk:
```suggestion
rr.ImageFormat(width, height, pixel_format="YUV420")
```

docs/content/reference/migration/migration-0-26.md
```diff
@@ -10,6 +10,26 @@ For non-blocking, use `timeout_sec=0`.
...
+```
+
+After:
+```py
+rr.ImageFormat(width, height, pixel_format="YUV420")
+
+blueprint.spawn("my_app", port=1234)
+```
+
+[ruff](https://github.com/astral-sh/ruff) (or your preferred Python linter) will find all the places in your code that needs to be updated!
```
ntjohnson1:
```suggestion
[ruff](https://github.com/astral-sh/ruff) (or your preferred Python linter) will find all the places in your code that need to be updated!
```

oxkitsune (approved):
LGTM!

Very nice improvement!

crates/build/re_types_builder/src/codegen/python/mod.rs
```diff
@@ -2460,17 +2466,19 @@ fn compute_init_parameters(obj: &Object, objects: &Objects) -> Vec<String> {
...
-        if optional.is_empty() {
-            required
-        } else if obj.kind == ObjectKind::Archetype {
-            // Force kw-args for all optional arguments:
+        if 2 < required.len() {
+            // There's a lot of required arguments.
+            // Using positional arguments would make usage hard to read.
+            // better for force kw-args for ALL arguments:
+            chain!(std::iter::once("*".to_owned()), required, optional).collect()
+        } else if optional.is_empty() {
```
timsaucer, line 2474: It looks like this is going to set all arguments to kw-args if number of required args is 3 or more but down in the ruff rules we have `max-positional-args = 3`. So is 3 allowable or do you want to intentionally be more restrictive here or did I miss something about how this bit works?

ntjohnson1 (approved):
I'm a little on the fence here so excuse the verbose post (accompanied by the approval since I don't think there is anything WRONG here).

1. This probably will break lots of people's (and our own external repos)
2. I don't generally see this pattern of `*` to explicitly force people to switch to kwargs but it's not totally bizarre
3. I think the hope of `PLR0917` is to reduce the total number of arguments and that this is an intermediate step to [PLR0913](https://docs.astral.sh/ruff/rules/too-many-arguments/#too-many-arguments-plr0913)
4. I wonder if varying between 1 and 3 positional args will be weird to people, I guess if we choose the right args to expose should be fine


I definitely don't want KWARGS only but this is probably a reasonable compromise. `PLR0913` is probably a much larger refactor but if we are feeling adventurous we could try enabling it, exclude all current offenders and hope its existence prevents future argument sprawl.

crates/build/re_types_builder/src/codegen/python/mod.rs
timsaucer: Does `self` count as one?

emilk: I wanted to be more restrictive here, at least for now. We seem to have no archetypes with exactly three required fields, so it doesn't matter much in practice for now. I rather start restrictive and expand later (removing a `*` is never a breaking change).

emilk: No, `self` is not counted.

PR #11385, opened 2025-09-30
Enable the same datafusion features everywhere

emilk:
I'm sick to death of recompiling `datafusion` from scratch several times per day.

This should help.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -221,7 +221,9 @@ convert_case = "0.6"
 criterion = "0.5"
 cros-codecs = "0.0.6"
 crossbeam = "0.8"
-datafusion = { version = "49.0.2", default-features = false, features = [
+datafusion = { version = "49.0.2", default-features = true, features = [
+  # NOTE: we enable the same features everywhere (including the default ones!),
+  # because otherwise we will recompile datafusion all the time based on our current compile target.
   "nested_expressions",
 ] }
 datafusion-ffi = "49.0.2"


Wumpf:
especially given that web is failing, we should check if this affects wasm size (once fixed).

At this point I'm frustrated enough to be ok with for instance the rust sdk depending on too much by default, but needlessly bigger wasm since is still a concern.

abey79:
> At this point I'm frustrated enough to be ok with for instance the rust sdk depending on too much by default, but needlessly bigger wasm since is still a concern.

Agreed. Since compilation/caching its on a per-arch basis, a situation where all the native targets have  a given subset of features and all the wasm targets have _another_ subset of features would still be "optimal".

emilk:
New attempt!

I checked the wasm size before/after, and it was 34 MB both times

Cargo.toml
```diff
@@ -257,7 +267,7 @@ itertools = "0.14"
 jiff = { version = "0.2.3", features = ["js"] }
 js-sys = "0.3"
 jsonwebtoken = { version = "9", default-features = false }
-lance = { version = "0.36.0", default-features = false }
+lance = { version = "0.36.0", default-features = false } # When you update this, also update the list of features in `datafusion`!
```
emilk:
```suggestion
lance = { version = "0.36.0", default-features = false } # When you update this, also update the list of features enabled for `datafusion` (~50 lines up)
```

PR #11387, opened 2025-09-30
Mask platform-specific menu entry in app kittest

MichaelGrupp:
Failed on Windows because Alt-F4 is displayed there.


MichaelGrupp:
This fixes the Windows build on the main branch.

MichaelGrupp:
Yeah, I'm also confused now. I generate the base snapshot on a Mac.

Wumpf:
ah. found it.

```
            #[cfg(all(not(target_arch = "wasm32"), target_os = "windows"))]
            Self::Quit => smallvec![KeyboardShortcut::new(Modifiers::ALT, Key::F4)],
```

so yeah you're fix is correct :)

Wumpf (approved):
thanks for following up!

we really should fix that ci though to catch this...

PR #11388, opened 2025-09-30
Add DataFusion catalog provider

timsaucer:
This PR adds a DataFusion `CatalogProvider` implementation for interacting with a Rerun server.

With this PR, any tables registered with the server are immediately available for querying via the DataFusion `SessionContext`. This includes the default `__entries` table. For example, you can do operations such as

```python
client.ctx.sql("SELECT * FROM __entries") 
```

or equivalently from the DataFrame APIs

```python
client.ctx.table("__entries")
```

## Limitations

In this initial implementation, commands like `register_table` on the `SchemaProvider` do not persist across sessions. Follow on work will support storing DataFrame Views on the server. Users can register lance tables via the `ConnectionClient` API.

### DIFF
--- a/rerun_py/src/catalog/datafusion_catalog.rs
+++ b/rerun_py/src/catalog/datafusion_catalog.rs
@@ -10,9 +10,15 @@ use re_redap_client::ConnectionClient;
 
 use crate::utils::get_tokio_runtime;
 
-#[pyclass(frozen, name = "DataFusionCatalog")]
+#[pyclass(frozen, eq, name = "DataFusionCatalog")]
 pub(crate) struct PyDataFusionCatalogProvider {
     pub provider: Arc<RedapCatalogProvider>,
+}
+
+impl PartialEq for PyDataFusionCatalogProvider {
+    fn eq(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.provider, &other.provider)
+    }
 }
 
 impl PyDataFusionCatalogProvider {


timsaucer:
@rerun-bot full-check

crates/store/re_datafusion/src/catalog_provider.rs
```diff
@@ -0,0 +1,230 @@
...
+use datafusion::common::{DataFusionError, Result as DataFusionResult, TableReference, exec_err};
+use parking_lot::Mutex;
+use re_protos::cloud::v1alpha1::{EntryFilter, EntryKind};
+use re_redap_client::ConnectionClient;
+use std::any::Any;
+use std::iter;
+use std::sync::Arc;
+
+pub const DEFAULT_CATALOG_NAME: &str = "datafusion";
+const DEFAULT_SCHEMA_NAME: &str = "public";
```
timsaucer, line 15: These are to match the defaults in datafusion.

crates/store/re_datafusion/src/catalog_provider.rs
```diff
@@ -0,0 +1,230 @@
...
+fn get_table_refs(client: &ConnectionClient) -> DataFusionResult<Vec<TableReference>> {
+    let mut builder = tokio::runtime::Builder::new_current_thread();
+    builder.enable_all();
+    let rt = builder.build().expect("failed to build tokio runtime");
+
+    rt.block_on(async {
+        Ok::<Vec<_>, DataFusionError>(
+            client
+                .clone()
+                .inner()
```
timsaucer: Catalog methods are all synchronous, so we need to do some blocking here.

timsaucer:
I do not think the full build failure is related to this change, and I did not reproduce it locally. I'm trying that test again and pulled up to main.

timsaucer:
Moving to draft because I want to add tests for alternate catalog and schema names.

crates/store/re_datafusion/src/catalog_provider.rs
zehiko: nit: I'd add that comment with bit more context in the code

crates/store/re_datafusion/src/catalog_provider.rs
```diff
@@ -0,0 +1,239 @@
...
+#[derive(Debug)]
+struct GrpcSchemaProvider {
+    catalog_name: Option<String>,
+    schema_name: Option<String>,
+    client: ConnectionClient,
+    in_memory_tables: Mutex<HashMap<String, Arc<dyn TableProvider>>>,
+}
+
+impl GrpcSchemaProvider {}
+
```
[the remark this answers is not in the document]
abey79: fair :)

crates/store/re_datafusion/src/catalog_provider.rs
abey79:
```suggestion
// These are to match the defaults in datafusion.
pub const DEFAULT_CATALOG_NAME: &str = "datafusion";
const DEFAULT_SCHEMA_NAME: &str = "public";
```

crates/store/re_datafusion/src/catalog_provider.rs
```diff
@@ -0,0 +1,239 @@
...
+        let schemas: HashMap<_, _> = iter::once((None, default_schema)).collect();
+
+        Self {
+            catalog_name: name.map(ToOwned::to_owned),
+            client,
+            schemas: Mutex::new(schemas),
+        }
+    }
+
+    fn update_from_server(&self) -> DataFusionResult<()> {
```
zehiko, line 92: is my understanding correct that the catalog and the schema discovery is tightly related to the name of ``entries`` on the server side?  if that's so, is it worth documenting perhaps? (or this is very common way how catalog provider implementations and the entry/table names server side work?

crates/store/re_datafusion/src/catalog_provider.rs
```diff
@@ -0,0 +1,239 @@
...
+use re_redap_client::ConnectionClient;
+use std::any::Any;
+use std::iter;
+use std::sync::Arc;
+
+pub const DEFAULT_CATALOG_NAME: &str = "datafusion";
+const DEFAULT_SCHEMA_NAME: &str = "public";
+
+#[derive(Debug)]
+pub struct GrpcCatalogProvider {
```
abey79: Even at this uncertain point, I think this would be cleaner, especially given that we use GRPC for other things than redap (so it's ambiguous).

```suggestion
pub struct RedapCatalogProvider {
```

crates/store/re_datafusion/src/catalog_provider.rs
abey79: I kind of agree with Copilot that creating a runtime each time seems inefficient. It feels that the catalog provider should be given a runtime, and it's the responsibility of the caller to provide one.

The viewer has one available (`App::async_runtime`), and the SDK has `get_tokio_runtime`.

crates/store/re_datafusion/src/catalog_provider.rs
```diff
@@ -0,0 +1,239 @@
...
+fn get_table_refs(client: &ConnectionClient) -> DataFusionResult<Vec<TableReference>> {
+    let mut builder = tokio::runtime::Builder::new_current_thread();
+    builder.enable_all();
+    let rt = builder.build().expect("failed to build tokio runtime");
+
+    rt.block_on(async {
+        Ok::<Vec<_>, DataFusionError>(
+            client
+                .clone()
+                .inner()
```
abey79: Please avoid using `inner()`, and use instead the `ConnectionClient`  wrapper function. If the wrapper function is not good enough, please improve it :) 

Btw, maybe this _entire_ `get_table_refs` (except the blocking part) should be part of the `ClientConnection`, as it could be useful elsewhere? It'd return a `Vec<String>` though, probably.

crates/store/re_datafusion/src/catalog_provider.rs
```diff
@@ -0,0 +1,239 @@
...
+        };
+
+        schemas
+            .get(&schema_name)
+            .map(|s| Arc::clone(s) as Arc<dyn SchemaProvider>)
+    }
+}
+
+#[derive(Debug)]
+struct GrpcSchemaProvider {
```
abey79:
```suggestion
struct RedapSchemaProvider {
```

crates/store/re_datafusion/src/catalog_provider.rs
```diff
@@ -0,0 +1,239 @@
...
+    catalog_name: Option<String>,
+    schema_name: Option<String>,
+    client: ConnectionClient,
+    in_memory_tables: Mutex<HashMap<String, Arc<dyn TableProvider>>>,
```
abey79, line 180: The fact that this provider enables registering new table that are not persisted should ideally be better documented. In particular, please add a docstring that says that this is where such tables land.

Related: I assume there is plan to actually persist these tables on the server when redap (aka the protocol) allows it? If so, probably worth opening an issue and using it as TODO.

crates/store/re_datafusion/src/catalog_provider.rs
```diff
@@ -0,0 +1,239 @@
...
+            table_ref.catalog() == self.catalog_name.as_deref()
+                && table_ref.schema() == self.schema_name.as_deref()
+                && table_ref.table() == name
+        }) {
+            return exec_err!("{name} already exists on the server catalog");
+        }
+
+        self.in_memory_tables.lock().insert(name, table);
+        Ok(None)
+    }
```
abey79, line 250: Same, let's document better what's going on here.

Also, this gets me thinking. Does it _actually_ make sense to implement this. Shouldn't we always return `DataFusionError::NotImplemented()` and call it a day? There should be a product-driven reason to have this imo (and if it exists, worth mentioning it in docstring).

rerun_py/src/catalog/datafusion_catalog.rs
```diff
@@ -0,0 +1,39 @@
+use crate::utils::get_tokio_runtime;
+use datafusion::catalog::CatalogProvider;
+use datafusion_ffi::catalog_provider::FFI_CatalogProvider;
+use pyo3::types::PyCapsule;
+use pyo3::{Bound, PyResult, Python, pyclass, pymethods};
+use re_datafusion::GrpcCatalogProvider;
+use re_redap_client::ConnectionClient;
+use std::sync::Arc;
```
abey79: nit: we have this untold, pseudo-convention of splitting `use` by `std`/external crates/`re_*` crates/ `crate::`. I'm not loosing sleep over it but it sure appeases my OCD when it's all in order 😅 

```suggestion
use std::sync::Arc;

use datafusion::catalog::CatalogProvider;
use datafusion_ffi::catalog_provider::FFI_CatalogProvider;
use pyo3::types::PyCapsule;
use pyo3::{Bound, PyResult, Python, pyclass, pymethods};

use re_datafusion::GrpcCatalogProvider;
use re_redap_client::ConnectionClient;

use crate::utils::get_tokio_runtime;
```

rerun_py/src/catalog/datafusion_catalog.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[pyclass(frozen, name = "DataFusionCatalog")]
+pub struct PyDataFusionCatalogProvider {
+    pub name: Option<String>,
+    pub provider: Arc<GrpcCatalogProvider>,
+}
```
abey79: Is this intended to be made public in the python SDK? If so, we should add it to the SDK header and `.pyi`. If not let's document it and explain why it can't be just inlined within `PyCatalogClientInternal::new()`.

rerun_py/tests/unit/test_datafusion_tables.py
```diff
@@ -317,24 +329,56 @@ def test_query_lance_table(server_instance: ServerInstance) -> None:
...
+    assert rb.num_rows > 0
+
+    # Verify SQL parsing for catalog provider works as expected
+    df = ctx.sql("SELECT * FROM simple_datatypes")
+    rb = df.collect()[0]
+    assert rb.num_rows > 0
+
+    df = ctx.sql("SELECT * FROM datafusion.public.simple_datatypes")
+    rb = df.collect()[0]
+    assert rb.num_rows > 0
```
abey79, line 395: Nice! 🤩

crates/store/re_datafusion/src/catalog_provider.rs
timsaucer: We need to do this because right now we are using `register_table` ourselves every time we do a `.df()` call. Otherwise I agree.

abey79: Aaaah right, makes sense.

But then the right thing to do is to change all of our APIs go to through the catalog instead of spawning/registering table providers. Right? (Which would btw trivially be implemented on python side, should we have a python wrapper for `DatasetEntry`.)

abey79: worth having an issue and assorted todos?

crates/store/re_datafusion/src/catalog_provider.rs
timsaucer: Updated doc and the follow on work is in linear for tracking

crates/store/re_datafusion/src/catalog_provider.rs
timsaucer: https://github.com/rerun-io/rerun/issues/11424

PR #11394, opened 2025-10-01
Introduce `Lens` as a concept for semantic mapping of arbitrary data

grtlr:
### What

This PR introduces the concept of _Lenses_ as a way to express re-interpretation and manipulation of component columns. This can, for example, be useful to attach Rerun semantics to arbitrary Arrow components that are created when loading MCAP files.

For now, the only way to access the feature is via the `LensesSink`, which transforms component columns and then forwards aggregated chunks to the backing sink.

> [!IMPORTANT]
> Lenses are still a very experimental feature and we expect the API to change a lot in upcoming releases.


examples/rust/lenses/src/main.rs
```diff
@@ -0,0 +1,163 @@
...
+// Logging helpers
+
+fn log_flag(rec: &RecordingStream) -> anyhow::Result<()> {
+    let flags = ["ACTIVE", "ACTIVE", "INACTIVE", "UNKNOWN"];
+    for x in 0..10i64 {
+        let flag = StringArray::from(vec![flags[x as usize % flags.len()]]);
+        rec.set_time("tick", TimeCell::from_sequence(x));
+        rec.log(
+            "flag",
+            &DynamicArchetype::new("com.Example.Flag")
```
jleibs:
```suggestion
            &DynamicArchetype::new("example:Flag")
```
nit: I'm not generally a fan of reverse-DNS namespaces unless they're being forced on us for some reason. Also, we previously decided to use `:` for scoping.

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -0,0 +1,734 @@
...
+    fn apply(&self, chunk: &Chunk) -> Vec<Chunk> {
+        let found = chunk
+            .components()
+            .iter()
+            .find(|(descr, _array)| descr.component == self.input.component);
+
+        // This means we drop chunks that belong to the same entity but don't have the component.
+        let Some((_component_descr, list_array)) = found else {
+            return Default::default();
+        };
```
jleibs, line 131: There's not necessarily anything better we can do here, but flagging this as possibly worrisome.

I think the real question is whether a user (esp if using a `Func` op) can change the validity of a row from `null` -> some value. If they can, then input-side concatentation of chunks (which auto-fills missing components with nulls) would lead to different outputs.

If we say any null input row must also correspond to a null output row, then this makes sense as a performance optimization.

crates/top/re_sdk/src/lenses.rs
```diff
@@ -0,0 +1,745 @@
+use re_chunk::{
+    Chunk, ChunkComponents, ChunkId, ComponentIdentifier, EntityPath,
+    external::arrow::{array::ListArray, datatypes::DataType, error::ArrowError},
+};
+use re_log_types::{EntityPathFilter, LogMsg, ResolvedEntityPathFilter};
+use re_types::ComponentDescriptor;
+
+use crate::sink::LogSink;
+
+/// A sink which can transform a `LogMsg` and forward the result to an underlying backing `LogSink`.
```
Wumpf:
```suggestion
/// A sink which can transform a [`LogMsg`] and forward the result to an underlying backing [`LogSink`].
```

crates/top/re_sdk/src/lib.rs
```diff
@@ -116,6 +116,13 @@ pub use re_types::{
     SerializedComponentColumn,
 };
 
+/// Transformation and reinterpretation of components.
+///
+/// # Experimental
+///
+/// This is an experimental API and may change in future releases.
+pub mod lenses;
```
Wumpf, line 124: `pub mod` + the fact that everything lives directly in `lenses.rs` makes it a bit hard to reason what is part of the public API and what isn't.
I'd appreciate more explicit listing out of those, probably by means of having a `lenses/mod.rs` and then keeping `pub mod`

crates/top/re_sdk/src/lenses.rs
```diff
@@ -0,0 +1,745 @@
...
+        self.relevant(chunk)
+            .flat_map(|transform| transform.apply(chunk))
+            .collect()
+    }
+}
+
+/// Provides commonly used transformations of Arrow arrays.
+///
+/// These operations should not be exposed publicly, but instead be wrapped by the [`Op`] abstraction.
+mod op {
```
Wumpf: would be nicer & less right shifted as an `op.rs`

crates/top/re_sdk/src/lenses.rs
```diff
@@ -0,0 +1,745 @@
...
+                nulls,
+            ))
+        }
+    }
+}
+
+/// Different variants of errors that can happen when executing lenses.
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[allow(missing_docs)]
```
Wumpf: can't put that attribute on the enum as a whole?

should prefer the newer `expect` whenever possible

crates/top/re_sdk/src/lenses.rs
```diff
@@ -0,0 +1,745 @@
...
+            Self::Func(func) => func(list_array),
+        }
+    }
+}
+/// Private module with the AST-like definitions of lenses.
+///
+/// **Note**: We should not leak those into the public API interface,
+/// so that we can evolve the definition of lenses over time, if requirements
+/// change.
+mod ast {
```
Wumpf: again, I'd go with `ast.rs`. Slightly overkill now, but surely not for long 😄

crates/top/re_sdk/src/lenses.rs
```diff
@@ -0,0 +1,745 @@
...
+    }
+}
+
+/// Provides convenient function to create a [`Lens`].
+pub struct LensBuilder(Lens);
+
+impl LensBuilder {
+    /// The column on which this [`Lens`] will operate on.
+    pub fn input_column(
+        entity_path_filter: EntityPathFilter,
```
Wumpf: doc should point out that we don't do any substitutions right now

crates/top/re_sdk/src/lenses/ast.rs
grtlr: > If they can, then input-side concatentation of chunks (which auto-fills missing components with nulls) would lead to different outputs.

Hmm, I don't think I follow. What do you mean be "input-side concatenation of chunks" here? The batcher?

crates/top/re_sdk/src/lenses.rs
```diff
@@ -0,0 +1,745 @@
...
+/// **Note**: We should not leak those into the public API interface,
+/// so that we can evolve the definition of lenses over time, if requirements
+/// change.
+mod ast {
+    use super::{
+        ComponentDescriptor, ComponentIdentifier, EntityPath, Op, ResolvedEntityPathFilter,
+    };
+
+    pub(super) struct InputColumn {
+        pub entity_path_filter: ResolvedEntityPathFilter,
```
Wumpf: Arguably the input should still be unresolved at this stage - if we allow substitutions in the future, then they'd be applied late when we have more context 🤔 
As you pointed out ast is what we could serialize and resolved entity path filters is conceptually not something that should be serialized

crates/top/re_sdk/src/lenses.rs
```diff
@@ -0,0 +1,745 @@
...
+        Self(Lens {
+            input: ast::InputColumn {
+                entity_path_filter: entity_path_filter.resolve_without_substitutions(),
+                component: component.into(),
+            },
+            outputs: vec![],
+        })
+    }
+
+    /// Can be used to define one or more output columns that are derived from the [`Self::input_column`].
```
Wumpf: slightly misleading. this method call adds a single column, it's just that you are allowed to add many.
Maybe call it `add_output_column`?

crates/top/re_sdk/src/lenses.rs
```diff
@@ -0,0 +1,745 @@
...
+            entity_path: entity_path.into(),
+            component_descr,
+            ops: ops.into_iter().collect(),
+            is_static: false,
+        };
+        self.0.outputs.push(column);
+        self
+    }
+
+    /// Can be used to define one or more output columns that are derived from the [`Self::input_column`].
```
Wumpf: as above

crates/top/re_sdk/src/lenses.rs
```diff
@@ -0,0 +1,745 @@
...
+    /// Finalizes this builder and returns the corresponding lens.
+    pub fn build(self) -> Lens {
+        self.0
+    }
+}
+
+/// A lens that transforms component data from one form to another.
+///
+/// Lenses allow you to extract, transform, and restructure component data
+/// as it flows through the logging pipeline. They are applied to chunks
```
Wumpf: "logging" sounds like the `send_` APIs don't affect it.
Also, here we don't know how it's gonna be used

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -0,0 +1,745 @@
...
+            },
+            datatypes::{DataType, Field},
+        },
+    };
+    use re_types::{ComponentDescriptor, archetypes::Scalars};
+
+    use super::*;
+
+    /// Creates a chunk that contains all sorts of validity, nullability, and empty lists.
+    // ┌──────────────┬───────────┐
```
Wumpf, line 301: very helpful, thanks for including these

crates/top/re_sdk/src/lenses.rs
grtlr: Muscle memory is hard to change 🙈

crates/top/re_sdk/src/lenses/ast.rs
grtlr: Fun fact, Claude does a good job going from these schematics to `ArrayBuilder`s.

crates/top/re_sdk/src/lenses/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+mod error;
+mod op;
+mod sink;
+
+pub use self::{
+    // We should be careful not to expose to much implementation details here.
+    ast::{Lens, LensBuilder, Op},
+    error::Error,
+    sink::LensesSink,
+};
```
Wumpf, line 18: nice!

Issue #8948, opened 2025-02-06
Replace (most) of our release checklist with kittest


emilk:
Most of our release checklists should be replaceable with automated snapshot tests using kittest. That would significantly simplify our release process.

[https://github.com/rerun-io/rerun/blob/main/tests/python/release_checklist/README.md](https://github.com/rerun-io/rerun/blob/main/tests/python/release_checklist/README.md)

PR #11395, opened 2025-10-01
Blueprint save/load test

aedm:
### Related

* Part of https://github.com/rerun-io/rerun/issues/8948

### What

Adds save/load/snapshot tests for blueprints.

Note: the `check_rbl_import.py` can't yet be removed as it's testing blueprint imports with a different app id, this test doesn't cover it yet, this is where `App` and `TestContext` differ somewhat.


crates/viewer/re_viewer/tests/blueprint_test.rs
```diff
@@ -0,0 +1,142 @@
+// #![cfg(feature = "testing")]
+
```
Wumpf: some brief overview of what we're intending to test here would be nice!

crates/viewer/re_viewer/tests/blueprint_test.rs
```diff
@@ -0,0 +1,142 @@
...
+        re_entity_db::StoreBundle::from_rrd(file).expect("Failed to load blueprint store");
+    {
+        let mut lock = test_context.store_hub.lock();
+        let app_id = lock.active_app().expect("Missing active app").clone();
+        lock.load_blueprint_store(rbl_store, &app_id)
+            .expect("Failed to load blueprint store");
+    }
+
+    // Trigger recalculation of visualizable entities and blueprint overrides.
+    test_context.setup_viewport_blueprint(|_ctx, _blueprint| {});
```
Wumpf, line 76: oof that's the annoying thing you've mentioned on yesterday standup where App is recalculating it all the time and `TestContext` is not? Feels like `TestContext` is in the wrong here, but I figure we can't do that where test context's run method lives or something of that shape?

Wumpf (approved):
nice!! very happy to see this automated now

crates/viewer/re_viewer/tests/blueprint_test.rs
aedm: Yes, it is. :(
The new `App`-based harness work differently, it loads the current blueprint from the store at the start of each frame. The `TestContext` only does that when you explicitly call this function.

PR #11391, opened 2025-10-01
Header propagation fixes and improvements

teh-cmc:
TODO

* Sibling: https://github.com/rerun-io/dataplatform/pull/1774

PR #11399, opened 2025-10-01
Autopropagating versioning headers everywhere

teh-cmc:
TODO

* DNM: requires #11391 
* Sibling: https://github.com/rerun-io/dataplatform/pull/1780

### DIFF
--- a/crates/store/re_protos/src/headers.rs
+++ b/crates/store/re_protos/src/headers.rs
@@ -12,6 +12,18 @@ pub const RERUN_HTTP_HEADER_ENTRY_ID: &str = "x-rerun-entry-id";
 /// while HTTP2 headers only support ASCII.
 pub const RERUN_HTTP_HEADER_ENTRY_NAME: &str = "x-rerun-entry-name-bin";
 
+/// The HTTP header key that all our official gRPC clients use to specify their identity and version.
+///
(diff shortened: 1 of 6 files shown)


crates/store/re_protos/src/headers.rs
```diff
@@ -127,19 +139,114 @@ impl<T> RerunHeadersExtractorExt for tonic::Request<T> {
...
+
+    pub fn new_server(name: Option<String>, version: Option<String>) -> Self {
+        Self::new(false, name, version)
+    }
+
+    pub fn new(is_client: bool, name: Option<String>, version: Option<String>) -> Self {
+        let mut name = name
+            .or_else(|| std::env::var("OTEL_SERVICE_NAME").ok())
+            .or_else(|| {
+                let path = std::env::current_exe().ok()?;
```
andrea-reale, line 204: nit: isn't the exe and CARGO_PGK the same for all services for now since we have a single entry point? (I guess not for the CLI, but there was a dicussion where we said it would not be too bad of a an idea to also have the CLI in the same binary).

teh-cmc: `CARGO_PKG` is the worst of them all: it will always be `re_protos` in this instance. So yeah, it's just layers of best effort in increasing order of shittiness :melting_face:

PR #11400, opened 2025-10-01
Add how to connect the server to the CLI command output

bllchmbrs:
Another small nit like #11398 .

Just helping to make it easier to know what to connect to.


emilk (approved):
Nice

PR #11401, opened 2025-10-01
Add meaningful equality comparisons to many rust wrappers

ntjohnson1:
### What
pyo3 defaults `__eq__` to be equivalent to `is` but since we clone before returning most python objects that means
`dataset.schema()==dataset.schema()` is False even though the contents are identical.

I resolved this specific case and added a test.
Then I asked a robot to write me a lint to enforce this everywhere, updated for things I could easily `derive(PartialEq, Eq)` then added NOLINT to the rest.
The score is
| With eq | Without eq |
|---------|------------|
|      8       |      15           |


rerun_py/src/catalog/catalog_client.rs
```diff
@@ -12,7 +12,7 @@ use crate::catalog::{
 };
 
 /// Client for a remote Rerun catalog server.
-#[pyclass(name = "CatalogClientInternal")]
+#[pyclass(name = "CatalogClientInternal")] // NOLINT
```
emilk: I wish the `// NOLINT` comment had a brief description of what lint it is noping out of, and _why_

emilk:
```suggestion
#[pyclass(name = "CatalogClientInternal")] // NOLINT: intentionally does not support `eq`
```

maybe?

emilk (approved):
Thanks!

rerun_py/src/catalog/catalog_client.rs
ntjohnson1: As I added them I kind of wished that our linter had a way to just opt out of a given lint rule instead of all lint rules.

`// NOLINT:pyclass_eq` or something but that was outside the scope here. I'll apply a note across them though.

ntjohnson1: I went for `skip pyclass_eq, non-trivial implementation`
`intentionally does not support` makes me feel like it means we should NEVER support it which is stronger than the statement I was making. I think it would be nice to at least improve our `__eq__` everywhere but at much lower priority than the classes that have direct comparison in rust.

For instance the python wrappers around `Arc<T>` cannot derive `Eq` but we could push down the comparison a little deeper with something like

```rust
pub fn __eq__...
   Arc.ptr_eq(&self.foo, &other.foo)
```

to at least see if the clone is pointing to the same object (I didn't check this specifically but expect this to equate to false today since the python objects would be different but point to the same rust content).

No super strong opinion on exact language if someone has something better and wants to find/replace.

PR #11406, opened 2025-10-02
feat: Export BindGroupEntry

### LINKED ISSUE #10720: Expose missing types that are in `pub` interface of `BindGroupDesc`
**Is your feature request related to a problem? Please describe.**

Currently, `BindGroupDesc` is part of the public interface, but not all of the types of its `pub` fields are exposed respectively. An example for this is `BindGroupEntry`.

This currently limits a user on Discord from implementing their own version of `re_renderer`.

Here is the link to the discussion: https://discord.com/channels/1062300748202921994/1396235826920161311

**Describe the solution you'd like**

If we want `BindGroupDesc` to be part of our public API surface, all of its children should be public too.

Weijun-H:
### Related

Closes #10720


Wumpf (approved):
sure 👍

PR #11408, opened 2025-10-02
feat: Add convenience methods for static row data in `ChunkBuilder`

### LINKED ISSUE #10356: Utility for creating static chunks using `ChunkBuilder`
### What

The `ChunkBuilder` should have utility to build static chunks.

Weijun-H:
### Related

Closes #10356

### What
Introduce dds convenience methods to `ChunkBuilder` for creating static chunks (chunks with no timeline data). Previously, users had to manually pass `TimePoint::default()` to create static/timeless data. 
### Changes Made

New Methods Added to `ChunkBuilder`: 
- **`with_static_row()`**
- **`with_static_sparse_row()`**
- **`with_static_archetype()`**
- **`with_static_component()`**
- **`with_static_component_batch()`**
- **`with_static_component_batches()`**
- **`with_static_sparse_component_batches()`**
- **`with_static_serialized_batch()`**
- **`with_static_serialized_batches()`**
- **`with_static_sparse_serialized_batches()`**

### DIFF
--- a/crates/store/re_chunk/src/builder.rs
+++ b/crates/store/re_chunk/src/builder.rs
@@ -264,6 +264,157 @@ impl ChunkBuilder {
         )
     }
 
+    /// Add a static row's worth of data using the given component data.
+    ///
+    /// This is a convenience method that adds data with an empty [`TimePoint`], meaning
+    /// the data will be considered static/timeless.
+    ///
+    /// Equivalent to calling `with_row(row_id, TimePoint::default(), components)`.
+    #[inline]
+    pub fn with_static_row(
+        self,
+        row_id: RowId,
+        components: impl IntoIterator<Item = (ComponentDescriptor, ArrayRef)>,
+    ) -> Self {
+        self.with_row(row_id, TimePoint::default(), components)
+    }
+
+    /// Add a static row's worth of data using the given sparse component data.
+    ///
+    /// This is a convenience method that adds data with an empty [`TimePoint`], meaning
+    /// the data will be considered static/timeless.
+    ///
+    /// Equivalent to calling `with_sparse_row(row_id, TimePoint::default(), components)`.
+    #[inline]
+    pub fn with_static_sparse_row(
+        self,
+        row_id: RowId,
+        components: impl IntoIterator<Item = (ComponentDescriptor, Option<ArrayRef>)>,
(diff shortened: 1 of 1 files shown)


grtlr:
It's not possible to build a chunk that mixes static and non-static data. I think adding these methods could lead someone to expect that this were possible (even more than the current API).

I think the root cause for this is the `Debug`-derive on `TimePoint`. I think we should rather deprecate that in favor of the already existing `TimePoint::STATIC` constant.

Weijun-H:
> It's not possible to build a chunk that mixes static and non-static data. I think adding these methods could lead someone to expect that this were possible (even more than the current API).
> 
> I think the root cause for this is the `Debug`-derive on `TimePoint`. I think we should rather deprecate that in favor of the already existing `TimePoint::STATIC` constant.

Hi @grtlr, thanks for reviewing. I agree. However, `TimePoint::default()` is too widely used to remove without causing breaking changes. As an alternative, we could add a check guard in `with_sparse_row` to prevent mixing static and temporal data in the same chunk. This would catch mixed chunks earlier while keeping the new static method available.

crates/store/re_chunk/src/builder.rs
```diff
@@ -264,6 +264,157 @@ impl ChunkBuilder {
...
+    ///
+    /// Equivalent to calling `with_row(row_id, TimePoint::default(), components)`.
+    #[inline]
+    pub fn with_static_row(
+        self,
+        row_id: RowId,
+        components: impl IntoIterator<Item = (ComponentDescriptor, ArrayRef)>,
+    ) -> Self {
+        self.with_row(row_id, TimePoint::default(), components)
+    }
```
emilk, line 280: Instead of duplicating all of this API, the user can just call `.with_row(row_id, TimePoint::STATIC, …)`. Obviously that wasn't discoverable enough. Not sure how we fix that, but maybe adding it as part of the docstring for `.with_row` and friends?

Weijun-H: Maybe add a mode/state to the `ChunkBuilder` itself that determines whether it's building a static or temporal chunk 🤔

Wumpf:
closing this and the underlying ticket since this is getting stale and apparently there's no clear consensus

PR #11414, opened 2025-10-03
Use rust-lld for Windows builds

Wumpf:
### Related

* Workaround for https://github.com/rust-lang/rust/issues/141626

### What

Started having trouble on debug builds Works out of the box like this.
Also updated list of crates we generate debug symbols for.

* [x] full ci


.cargo/config.toml
```diff
@@ -20,3 +20,8 @@ IS_IN_RERUN_WORKSPACE = "yes"
 # https://docs.rs/getrandom/latest/getrandom/#webassembly-support
 [target.wasm32-unknown-unknown]
 rustflags = ['--cfg=web_sys_unstable_apis', '--cfg=getrandom_backend="wasm_js"']
+
+# Use rust-lld on Windows due problems with large debug symbols.
+# See https://github.com/rust-lang/rust/issues/141626
+[target.x86_64-pc-windows-msvc]
+linker = "rust-lld"
```
IsseW, line 27: Should it be mentioned somewhere that you need `rust-lld` for building on windows?

Wumpf:
@rerun-bot full-ci

Cargo.toml
```diff
@@ -458,51 +458,61 @@ debug = false
 "re_byte_size".debug = true
 "re_capabilities".debug = true
 "re_case".debug = true
-"re_chunk".debug = true
```
IsseW, line 461: This file was just sorted I guess?

Wumpf: yeah, tbh not sure what was added since then. I know that I forgot about `re_tf` but that's the only one.

.cargo/config.toml
Wumpf: I was wondering about that as well at first but it seems to come with regular rust install. There's no `rust-lld` file or symlink on my machine as far as I can tell. It's a thing directly recognized by rustc - it is called the "self contained" linker [here](https://doc.rust-lang.org/rustc/codegen-options/index.html#linker-features)
I suppose since rust comes with llvm for the backend, this makes sense; it already has the llvm compiler so it drags the linker with it always.
Might be wrong though, haven't found anything definitive. Ci Seems to be just happy though as well

Wumpf:
@rerun-bot full-check

PR #11420, opened 2025-10-03
`rerun auth token`: print newline

emilk:
Previously `rerun auth token` would not print a newline, which made it hard to accurately copy the token.
In my case, I got an extra `%` suffix on the token. The viewer didn't complain about it, but the server did.


jprochazk (approved):
For some reason I thought the trailing newline would be a problem for `REDAP_TOKEN=$(rerun auth token)`, but bash trims the trailing newline

PR #11421, opened 2025-10-03
Fix shader reloading detecting changes on mere file access on Linux

### LINKED ISSUE #11306: (Debug builds) continuously recompile shaders on Linux
Something in the filewatcher seems to go haywire and we detect changes non-stops to shaders when running debug Rerun on WSL

Wumpf:
### Related

* Fixes #11306

### What

how did this ever work?

* [x] can still reload shaders on Linux
* [ ] can still reload shaders on Windows

### DIFF
--- a/crates/viewer/re_renderer/src/file_server.rs
+++ b/crates/viewer/re_renderer/src/file_server.rs
@@ -259,12 +259,12 @@ mod file_server_impl {
                     #[allow(clippy::enum_glob_use)]
                     use notify::EventKind::*;
                     match ev.kind {
-                        Access(_) | Create(_) | Modify(_) | Any => ev
+                        Create(_) | Modify(_) | Any => ev
                             .paths
                             .into_iter()
                             .filter_map(canonicalize_opt)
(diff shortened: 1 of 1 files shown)


Wumpf:
Doesn't work for @jprochazk  :(

crates/viewer/re_renderer/src/file_server.rs
```diff
@@ -259,12 +259,16 @@ mod file_server_impl {
                     #[allow(clippy::enum_glob_use)]
                     use notify::EventKind::*;
                     match ev.kind {
+                        // If we don't react to any kind of `Access` event, we'll miss some changes on linux.
+                        Access(notify::event::AccessKind::Open(_)) | Remove(_) | Other => {
```
Wumpf: removed all of `Access` before and limited it further in an attempt to fix the issue for @jprochazk 
But while this removes all the spam, it still breaks shader reloading

jprochazk:
I think fixing the shader spam is more important than making sure the shader hotreloading works

Issue #10959, opened 2025-08-21
Remove Python 3.9 Support


ntjohnson1:
Python 3.9 reaches EOL October 2025. Remove support in our pyproject.
Check ruff tools to migrate to newer minimum and check for ruff lints that we can turn on now (optional vs <item>|None consistency).

PR #11426, opened 2025-10-03
Drop Python 3.9 Support

ntjohnson1:
### Related

Closes https://github.com/rerun-io/rerun/issues/10959

### What

Biggest churn is the `Union[type,type]` to `type | type`. At first I just replaced `, ` join to ` | ` but that doesn't work for groupings of `Tuple[type, type]`. So I went and updated the format of the fbs to minimize the complexity in the codegen.

I split out the commits to hopefully make this easier to review codegen then resulting changes. A few small commits towards the end to fix some new things that crop up once we follow the newer formatting.


examples/python/notebook_neural_field_2d/neural_field_2d.ipynb
```diff
@@ -1,261 +1,261 @@
...
+    "        self.num_layers = num_layers\n",
+    "        self.dim_hidden = dim_hidden\n",
+    "        self.dim_in = dim_in\n",
+    "        self.dim_out = dim_out\n",
+    "        self.activation = activation\n",
+    "        self.pe_sigma = pe_sigma\n",
+    "\n",
+    "        sizes = [dim_in] + [dim_hidden for _ in range(num_layers - 1)] + [dim_out]\n",
+    "        self.linears = torch.nn.ModuleList()\n",
+    "        for in_size, out_size in itertools.pairwise(sizes):\n",
```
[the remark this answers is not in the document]
ntjohnson1: These changes are from ruff now that we have python 3.10. https://docs.astral.sh/ruff/rules/zip-instead-of-pairwise/

rerun_py/tests/unit/blueprint_utils.py
```diff
@@ -16,5 +17,5 @@ def assert_blueprint_contents_are_equal(*contents: rrb.View | rrb.Container) ->
     def strip_id_field(d: dict[str, Any]) -> dict[str, Any]:
         return {k: v for k, v in d.items() if k != "id"}
 
-    for i, (c1, c2) in enumerate(zip(contents, contents[1:])):
+    for i, (c1, c2) in enumerate(itertools.pairwise(contents)):
```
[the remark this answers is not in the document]
ntjohnson1: These changes are from ruff now that we have python 3.10. https://docs.astral.sh/ruff/rules/zip-instead-of-pairwise/

timsaucer:
Why did the markdown formatting need to be changed? Seems unrelated but maybe I missed something.

CHANGELOG.md
```diff
@@ -1,13 +1,18 @@
 # Rerun changelog
 
+## 0.26.0 - TBD - TBD
+
+### ⚠️ Breaking changes
+
+-   Python 3.9 support has been dropped.
```
timsaucer:
```suggestion
-   Minimum supported Python is now 3.10.
```

timsaucer: I prefer framing this as a positive, not a negative ;)

crates/build/re_types_builder/src/codegen/python/mod.rs
```diff
@@ -1522,13 +1519,11 @@ fn quote_aliases_from_object(obj: &Object) -> String {
             format!(
                 r#"
                 if TYPE_CHECKING:
-                    {name}Like = Union[
-                        {name},
-                        {aliases}
-                    ]
+                    {name}Like = {name}{aliases}
```
timsaucer, line 1522: Are aliases comma or `|` separated?

crates/build/re_types_builder/src/codegen/python/mod.rs
```diff
@@ -1567,23 +1559,20 @@ fn quote_union_aliases_from_object<'a>(
...
         r#"
             if TYPE_CHECKING:
-                {name}Like = Union[
-                    {name},{union_fields},{aliases}
-                ]
-                {name}ArrayLike = Union[
-                    {name},{union_fields},
-                    Sequence[{name}Like],{array_aliases}
-                ]
+                {name}Like = {name} | {union_fields}{aliases}
```
timsaucer, line 1568: Do we need a `|` between `{union_fields}{aliases}`

crates/build/re_types_builder/src/codegen/python/mod.rs
```diff
@@ -991,16 +991,13 @@ def auto(cls, val: str | int | {enum_name}) -> {enum_name}:
         .join(", ")
     );
     code.push_unindented(
-        format!("{enum_name}Like = Union[{enum_name}, {variants}, int]"),
+        format!("{enum_name}Like = {enum_name} | {variants} | int"),
```
timsaucer, line 994: It looks like a breaking change to the `.fbs` files where the variants and aliases now must all be `|` delimited instead of `,`. Is there a way we can validate that his has happened in all files? Otherwise it requires manual verification and I could see a case where someone who has been working on those changes independent of this PR merges them in and doesn't catch the need to update the definitions. One thing we could do was a check to see if splitting the string by comma returns the same as the original string (so that we have no remaining `,` splits).

crates/build/re_types_builder/src/codegen/python/mod.rs
timsaucer: Answering my own question: the `.fbs` files are now `|` delimited

pixi.toml
```diff
@@ -756,8 +756,10 @@ numpy = ">=2"   # Rerun still needs numpy <2. Enforce this outside of the pypi d
 
 
 [feature.python-pypi.pypi-dependencies]
-rerun-sdk = { version = "==0.24.0", extras = ["all"] }
-rerun-notebook = "==0.24.0"
+# TODO(nick): When we release 0.26.0 we should bump this and put extras back to all
+# currently blocked on datafusion being too high (which means this env probably doesn't work)
```
timsaucer: I don't understand the statement about datafusion being too high

ntjohnson1: I'll roll back this change since it didn't strictly need to be here.

Basically our python-pypi env installs a published wheel. However, this wheel is older than main so it only supports datafusion 48 but the pixi env is at 49.

crates/build/re_types_builder/src/codegen/python/mod.rs
ntjohnson1: Right now aliases can be an empty string. My solution was a global one liner to prefix with a ` | ` then strip the trailing ` | ` if any. Happy to do a different approach if preferred. Could do a `separator` string that is also empty if `aliases` is empty. When the separators were `, ` inside of `Union[...]` a trailing comma wasn't an issue.

crates/build/re_types_builder/src/codegen/python/mod.rs
ntjohnson1: I added a note in the Pr description but I can update it to try to make that clearer.

crates/build/re_types_builder/src/codegen/python/mod.rs
ntjohnson1: Yes that is true. My understanding of the validation is:
* CI ensures codegen doesn't change anything (that the checked in content matches what we would generate)
* CI then confirms that linting passes. A comma is not acceptable in our type aliases with Python 3.10

Splitting the original string on `, ` isn't a valid strategy because that is why I chose to update the `fbs` instead of just replacing the separator in our code gen. Right now the aliases are a single string and a breaking case that we have in the code is `Tuple[type_1, type_2]` which is not equivalent to `Tuple[type_1 | type_2]`. A more comprehensive breaking change would be to list the aliases as a list of strings then we could apply any separator we want in the code gen but that seemed like too much churn.

ntjohnson1 (commented):
I rolled back the markdown changes and re-applied them without format on save. I didn't see any local tests fail so this might be a function of our `prettier` vscode settings in the repo (I don't think I customized it further). I can take a look more carefully next week and file an issue to either disable prettier on our markdown or add a tool to make it uniform.

ntjohnson1:
Also the notebook change appears to just be from ruff since I didn't touch that. However, I recommend enabling the rich diff on notebooks if you haven't already. At first I was confused about your comment until I toggled it to the source diff since the image below was all I saw. https://github.blog/changelog/2023-03-01-feature-preview-rich-jupyter-notebook-diffs/

Screenshot 2025-10-04 at 9 25 46 AM

ntjohnson1:
Moved this to draft since we decided to push back to 0.27

ntjohnson1:
I believe my previous notes addressed concerns. I rebased onto main for post 0.26 release and also removed all irrelevant white space changes. We can discuss further if my understanding is incorrect and there is an obvious issue with the validation that `.fbs` changes match the new `|` separator format.

PR #11427, opened 2025-10-04
Transform resolution is no longer a store subscriber

Wumpf:
### Related

* Fixes RR-2489

### What

Mostly a pure refactor. May change perf characteristics a bit, most notably no more transform processing occurs at all until a 3D view is created. Not expecting any major shifts though.

`TransformCacheStoreSubscriber` -> `TransformResolutionCache`

In a nutshell, this removes the tight coupling of transform logic with store subscribers and migrates keeping transforms up to date in the Viewer to a Viewer-Cache.
Chunk events are still used in `TransformResolutionCache` but are a merely a thin aspect of its frontend and can be bypassed for chunk-adding only.

Motivation:
* major step towards coalescing into a multi-faceted `TransformDatabase` object which internally will contain transform resolution cache as well as transform tree
    * `TransformCacheStoreSubscriber` being a subscriber dependency made this unwiedly as we couldn't keep all objects in the same place easily due to the static nature of store subscribers
* tightly control when what is updated and how parallel that happens
* allow transform resolution with partial knowledge - we want to feed in chunks manually and observe the results for the given set of chunks
    * this is important for a query aware future where we can't rely on chugging through fully downloaded stores 

Testing:
Seems to be covered well enough by unit tests on `TransformResolutionCache` as well as integration tests in `re_view_spatial` (both of which prevented mistakes here!)

Reviewing:
Lots of whitespace in existing tests due to nesting decrease (!) as things are no longer happening under a subscriber lock


crates/store/re_tf/src/lib.rs
```diff
@@ -1,12 +1,12 @@
 //! Rerun spatial transform processing
 
-mod transform_cache;
+mod component_type_info;
+mod transform_resolution_cache;
 mod transform_tree;
```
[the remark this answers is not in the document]
Wumpf: good catch, this isn't ready yet

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -99,6 +73,37 @@ bitflags::bitflags! {
     }
 }
 
+impl TransformAspect {
+    /// Converts a component type to a transform aspect.
+    pub fn from_component_type(component_type: ComponentType) -> Self {
```
oxkitsune, line 78: probably not important but: previously a `ComponentType` could be part of multiple aspects, because:

```rust
for component_type in event
    .chunk
    .component_descriptors()
    .filter_map(|c| c.component_type)
{
    if self.transform_components.contains(&component_type) {
        aspects |= TransformAspect::Tree;
    }
    if self.pose_components.contains(&component_type) {
        aspects |= TransformAspect::Pose;
    }
    if self.pinhole_components.contains(&component_type) {
        aspects |= TransformAspect::PinholeOrViewCoordinates;
    }
    if component_type == re_types::components::ClearIsRecursive::name() {
        aspects |= TransformAspect::Clear;
    }
 ...
}
```

with these changes, this isn't possible anymore.

Wumpf: ah yeah nice catch. I think there's another refactor actually missing here. As of now my change is correct, but having separate component types for e.g. pose & general translation should be refactored and instead be a property of the archetype

PR #11428, opened 2025-10-04
Walkers updated to 0.45

podusowski:
Checked the flight data demo and it look okay. This version has a [new `show` api](https://docs.rs/walkers/latest/walkers/struct.Map.html#method.show), though I don't see any benefit of using it in Rerun.


Wumpf (approved):
Thank you!
Gave it a brief spin as well, seems all fine :)

PR #11433, opened 2025-10-06
Print out the url that 'rerun auth login' opens

emilk:
Printing out the URL that is about to open has several benefits:
* It lets the user pick a different browser
* It acts as a contingency if the automatic opening of the browser fails


Wumpf (approved):

Wumpf:
ah maybe that was too hastily merged, after all @jprochazk was picked explicitly as a reviewer. Sorry!

PR #11435, opened 2025-10-06
Kittest utility toolbelt

aedm:
### Related

* Part of #8948 

### What

- Allows interacting with app state from an integration test using a test hook function
- Adds a bunch of Rerun-specific utility functions to Kittest
- Replaces a context menu test from the release checklist as a proof of concept

### DIFF
--- a/crates/viewer/re_viewer/tests/snapshots/menu_without_recording.png
+++ b/crates/viewer/re_viewer/tests/snapshots/menu_without_recording.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:0d708f2f212f2fe6a16e16ccfb9e39388b931e71425ff31d8eddaf8a7b8dd5af
-size 136512
+oid sha256:3a6ea227b5e9b374bca3143be195f2dfe54331b690aaf1a2d4791c474cceef7e
+size 121810
--- a/crates/viewer/re_viewer/tests/snapshots/settings_screen.png
+++ b/crates/viewer/re_viewer/tests/snapshots/settings_screen.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:605ea30ac8e1367162e47e42279605763c590d3619cfec7a70bb93d3355eadda
-size 106222
+oid sha256:3c85e493858367eb4225a39ef19ac25a4078a863e25507dec51eb061bf0980d9
+size 91649


crates/viewer/re_viewer/src/viewer_test_utils/app_testing_ext.rs
```diff
@@ -0,0 +1,22 @@
+use re_viewer_context::StoreHub;
+
+use crate::App;
+
+#[cfg(feature = "testing")]
```
emilk: Alternatively, put `#![cfg(feature = "testing")]` at the top of the file!

tests/rust/re_integration_test/src/kittest_harness_ext.rs
```diff
@@ -0,0 +1,232 @@
...
+    // Clicks a node in the UI by its label.
+    fn click_label(&mut self, label: &str);
+    fn right_click_label(&mut self, label: &str);
+
+    // Takes a snapshot of the current app state with good-enough snapshot options.
+    fn snapshot_app(&mut self, snapshot_name: &str);
+
+    // Prints the current viewer state. Don't merge code that calls this.
+    #[allow(unused)]
+    fn debug_viewer_state(&mut self);
```
emilk: Why not merge code that calls it? Is it very slow?

If it's called from a test, stdout/stderr only reaches the terminal output if the tests fails.

tests/rust/re_integration_test/src/kittest_harness_ext.rs
```diff
@@ -0,0 +1,232 @@
...
+        self.click_label("Blueprint panel toggle");
+    }
+
+    fn toggle_time_panel(&mut self) {
+        self.click_label("Time panel toggle");
+    }
+
+    fn toggle_selection_panel(&mut self) {
+        self.click_label("Selection panel toggle");
+    }
```
emilk, line 70: It would be a lot nicer to have idempotent commands here, i.e. `fn show_blueprint_panel(&mut self, open: bool)`

tests/rust/re_integration_test/src/kittest_harness_ext.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    fn click_label(&mut self, label: &str) {
+        self.get_by_label(label).click();
+        self.run_ok();
+    }
+
+    fn right_click_label(&mut self, label: &str) {
+        self.get_by_label(label).click_secondary();
+        self.run_ok();
+    }
```
emilk, line 201: I appreciate having these helpers to not having to call `.run_ok()` after each action… but I wish we had this in kittest instrad. `kittest` allows user to buffer up many inputs (e.g. push down some modifiers before clicking), which is powerful, but the price is all these manual calls to `.run()`. And realistically, you almost _shouldn't_ test clicking two things in the same frame, since realistically that shouldn't happen in real usage.

In other words: I believe kittest's API is a bit too low-level for most tests.

I don't think there is anything actionable here, except food for thought, and ping @lucasmerlin

tests/rust/re_integration_test/src/kittest_harness_ext.rs
```diff
@@ -0,0 +1,232 @@
...
+        });
+    }
+
+    fn snapshot_app(&mut self, snapshot_name: &str) {
+        self.run_ok();
+        // TODO(aedm): there is a nondeterministic font rendering issue.
+        self.snapshot_options(
+            snapshot_name,
+            &SnapshotOptions::new().failed_pixel_count_threshold(0),
+        );
```
emilk: Is there an open issue about this we can link to?

Did you investigate doing the manual texture filtering in egui-wgpu?

tests/rust/re_integration_test/tests/basic_tests.rs
```diff
@@ -0,0 +1,35 @@
...
+use re_viewer::external::re_types;
+use re_viewer::external::re_viewer_context::ViewClass as _;
+use re_viewer::viewer_test_utils;
+use re_viewport_blueprint::ViewBlueprint;
+
+#[tokio::test(flavor = "multi_thread")]
+pub async fn test_single_text_document() {
+    let mut harness = viewer_test_utils::viewer_harness();
+    harness.init_recording();
+    harness.toggle_selection_panel();
```
emilk, line 14: See this is the problem with non-idempotency. Is this opening or closing the selection panel? 🤷

tests/rust/re_integration_test/tests/basic_tests.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    // Set up the viewport blueprint
+    harness.clear_current_blueprint();
+    harness.setup_viewport_blueprint(|_viewer_context, blueprint| {
+        blueprint.add_view_at_root(ViewBlueprint::new_with_root_wildcard(
+            TextDocumentView::identifier(),
+        ));
+    });
+
+    harness.snapshot("single_text_document_2");
```
emilk: This is so cool

tests/rust/re_integration_test/tests/basic_tests.rs
```diff
@@ -0,0 +1,35 @@
...
+use re_sdk::TimePoint;
+use re_sdk::log::RowId;
+use re_view_text_document::TextDocumentView;
+use re_viewer::external::re_types;
+use re_viewer::external::re_viewer_context::ViewClass as _;
+use re_viewer::viewer_test_utils;
+use re_viewport_blueprint::ViewBlueprint;
+
+#[tokio::test(flavor = "multi_thread")]
+pub async fn test_single_text_document() {
```
emilk, line 11: Why do we need tokio here at all?

tests/rust/re_integration_test/tests/snapshots/single_text_document_1.png
emilk, line 1: Looks like the toggle button is still being hovered?

I think we should try to avoid that.
There should be some way to remove the mouse cursor after a click. Maybe kittest should even do so automatically?
Ping @lucasmerlin

tests/rust/re_integration_test/tests/snapshots/single_text_document_2.png
emilk, line 1: It feels like these are maybe unnecessarily large. What sets their default size?

emilk (approved):
Really awesome ⭐

tests/rust/re_integration_test/src/kittest_harness_ext.rs
lucasmerlin: Actually kittest runs all queued up events in seperate steps (so you could click a text edit, type some text and then click a button, and all the events would happen on seperate frames, even if you only call run() once). 
But this is really not obvious, so theres definitely room for improvements in that.

 And to make it run automatically on clicks, kittest would require a lot of interior mutability, which might make some other apis kind of awkward. But maybe it'd be nice, then we could also hold Nodes across run() calls (so you don't have to re-query them all the time).

tests/rust/re_integration_test/tests/snapshots/single_text_document_1.png
lucasmerlin: Not sure if we should do so by default but we could add a `harness.remove_cursor()` I guess

tests/rust/re_integration_test/tests/basic_tests.rs
lucasmerlin: I believe some parts of the viewer are very unhappy without a tokio runtime (e.g. the redap browser). 
If we're sure that we'll never need to await in tests, we could maybe make this part of `viewer_harness` setup (by manually spawning a tokio runtime there and making it current for the current thread).

tests/rust/re_integration_test/src/kittest_harness_ext.rs
emilk: > Actually kittest runs all queued up events in seperate steps

So why are the `.run_ok()` calls even needed here then?

tests/rust/re_integration_test/tests/snapshots/single_text_document_1.png
emilk: Yeah, that sounds good. In this case it's an "implementation detail" of how we toggle panels that makes that button _happen_ to be hovered, but if we change how we toggle the panels we shouldn't need to change the screenshot.

tests/rust/re_integration_test/tests/basic_tests.rs
aedm: `App::new` has an `AsyncRuntimeHandle` argument, can't be created without a Tokio runtime.

See `crates/viewer/re_viewer/src/viewer_test_utils/mod.rs`.

tests/rust/re_integration_test/tests/snapshots/single_text_document_2.png
aedm: Good call. I added `HarnessOptions` to control window size it and decreased the default value.

crates/viewer/re_viewer/src/viewer_test_utils/app_testing_ext.rs
aedm: TIL!

tests/rust/re_integration_test/src/kittest_harness_ext.rs
aedm: > Did you investigate doing the manual texture filtering in egui-wgpu?

Not yet. I made a tracking issue and added it to the comment. WIll investigate later.

tests/rust/re_integration_test/src/kittest_harness_ext.rs
aedm: After investigation, it's a bit more complex than I hoped for.
Added tracking issue, will implement as a standalone PR: https://github.com/rerun-io/rerun/issues/11451

tests/rust/re_integration_test/tests/snapshots/single_text_document_1.png
aedm: This particular issue will be addressed by https://github.com/rerun-io/rerun/issues/11451

In general, I'll add some manual cursor movement to utility functions if it gets troublesome.

PR #11416, opened 2025-10-03
refactor: Move examples for notebook viewer and callbacks to `examples/python`

Weijun-H:
### Related

Closes #10711

PR #11440, opened 2025-10-06
Add Minimal Notebook Validation

ntjohnson1:
### Related
https://github.com/rerun-io/rerun/pull/11416
I requested the notebooks get moved so our automated examples test would make sure the notebooks aren't broken. Unfortunately we don't test any of our notebooks so most were broken. 

A little bit related to broader notebook testing improvement https://linear.app/rerun/issue/RR-1880/test-our-jupyter-notebooks-on-ci

### What
I highly recommend reviewing with rich diff for notebooks enabled.
* Adds nbqa so we can run mypy on the notebooks to at least ensure the apis align
   * I couldn't get the config working so unfortunately I just have the directories to search listed in the pixi command
* Adds nbstripout but doesn't yet tie it to our py-fmt. Jupyter notebook outputs are just binary so that will bloat the repo and complicate diffs. Unfortunately nbstripout doesn't have a nice config setup so I just put together a simple wrapper.
* I made two actual code changes not just in tests since it made sense to resolve notebook mypy errors
   1. I made component_type a class method. We basically have component type raw strings everywhere probably because it wasn't easy before to grab them from the class definition rather than instance
   2. If we have a recording string the api should actually be a little different than asking globally. I assume we'll always find the id so we universally return a string.
   
This still doesn't execute our notebooks which is probably a worthwhile followup issue.

### DIFF
--- a/examples/rust/objectron/src/objectron.rs
+++ b/examples/rust/objectron/src/objectron.rs
@@ -96,7 +96,17 @@ pub struct AvDepthData {
 /// Nested message and enum types in `AVDepthData`.
 pub mod av_depth_data {
     /// Indicates the general accuracy of the depth_data_map.
-    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
+    #[derive(
+        Clone,
+        Copy,
+        Debug,
+        PartialEq,
+        Eq,
+        Hash,
+        PartialOrd,
+        Ord,
+        ::prost::Enumeration
+    )]
     #[repr(i32)]
     pub enum Accuracy {
         UndefinedAccuracy = 0,
@@ -129,7 +139,17 @@ pub mod av_depth_data {
         }
     }
     /// Quality of the depth_data_map.
-    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
+    #[derive(
+        Clone,
+        Copy,
+        Debug,
+        PartialEq,
+        Eq,
+        Hash,
+        PartialOrd,
+        Ord,
+        ::prost::Enumeration
+    )]
     #[repr(i32)]
     pub enum Quality {
         UndefinedQuality = 0,
@@ -178,7 +198,9 @@ pub struct ArLightEstimate {
     /// A vector indicating the orientation of the strongest directional light
     /// source, normalized in the world-coordinate space.
     #[prost(message, optional, tag = "4")]
-    pub primary_light_direction: ::core::option::Option<ar_light_estimate::DirectionVector>,
+    pub primary_light_direction: ::core::option::Option<
+        ar_light_estimate::DirectionVector,
+    >,
     /// The estimated intensity, in lumens, of the strongest directional light
     /// source in the scene.
     #[prost(float, optional, tag = "5")]
@@ -253,7 +275,17 @@ pub mod ar_camera {
     }
     /// The general quality of position tracking available when the camera captured
     /// a frame.
-    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
+    #[derive(
+        Clone,
+        Copy,
+        Debug,
+        PartialEq,
+        Eq,
+        Hash,
+        PartialOrd,
+        Ord,
+        ::prost::Enumeration
+    )]
     #[repr(i32)]
     pub enum TrackingState {
         UndefinedTrackingState = 0,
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -462,7 +462,7 @@ You can now log URDF files directly to Rerun using the `log_file` API.
 - Make `archetype_field_name` mandatory in `ComponentDescriptor` [#10082](https://github.com/rerun-io/rerun/pull/10082)
 
 #### 📦 Dependencies
-- Update to wgpu 0.25 [#10009](https://github.com/rerun-io/rerun/pull/10009) (thanks [@Vrixyz](https://github.com/Vrixyz)!)
+- Update to wgpu 0.25 [#10009](https://github.com/rerun-io/rerun/pull/10009) (thanks [@ThierryBerger](https://github.com/ThierryBerger)!)
```
Wumpf, line 465: curious, how did you notice that user changed their name?

examples/python/notebook_callbacks/README.md
```diff
@@ -19,7 +19,8 @@ Check out the [minimal notebook example](https://rerun.io/examples/integrations/
 
 ## Background
 
-This notebook spins up a colorful point cloud and pipes it into the viewer so you can experiment with callbacks in real time. As the camera, timeline, and selection change, [`Viewer.on_event`](https://www.rerun.io/docs/reference/sdk/rerun_notebook#rerun.notebook.Viewer.on_event) emits rich event payloads that we translate into friendly [`ipywidgets`](https://ipywidgets.readthedocs.io/) readouts.
+<!-- TODO(#11453): Add link to Viewer.on_event when page exists -->
```
Wumpf, line 22: we have https://rerun.io/docs/howto/callbacks at least

Wumpf: huh but why is there no `Viewer` documented. We have `notebook_show` in there. Strange!

examples/python/notebook_callbacks/notebook_callbacks.ipynb
Wumpf, line 1: what is `noqua A004`? Can we document that on the same line?

examples/python/notebook_viewer/notebook_viewer.ipynb
Wumpf, line 1: just cosmetic or why was this change needed?

rerun_py/rerun_sdk/rerun/_baseclasses.py
```diff
@@ -456,13 +456,14 @@ def arrow_type(cls) -> pa.DataType:
...
-    def component_type(self) -> str:
+    @classmethod
+    def component_type(cls) -> str:
         """
         Returns the name of the component.
 
         Part of the `ComponentBatchLike` logging interface.
         """
-        return self._BATCH_TYPE._COMPONENT_TYPE  # type: ignore[attr-defined, no-any-return]
+        return cls._BATCH_TYPE._COMPONENT_TYPE  # type: ignore[attr-defined, no-any-return]
```
Wumpf, line 466: yikes. not covered by a test then? seems bad as well

Wumpf (approved):
awesome, thank you so much

CHANGELOG.md
ntjohnson1: I saw the error and had to do some googling. Their mastadon/other socials point to the updated github. I also checked the PR linked here as well.

examples/python/notebook_viewer/notebook_viewer.ipynb
ntjohnson1: In the PR that moved the notebooks the author removed the `panel_states` deprecated argument from the Viewer. This put it back to maintain the original state but with the updated API https://github.com/rerun-io/rerun/pull/11416

examples/python/notebook_callbacks/notebook_callbacks.ipynb
ntjohnson1: This is how you skip ruff rules. I think redefining it seems redundant/non-standard to me.

I can make a note that we are skipping the rule to make mypy happy to ensure `display` is defined.

examples/python/notebook_callbacks/README.md
ntjohnson1: Howto callbacks links to the notebook page which links to this notebook example 😬

Right now our python docs are all opt-in so we are missing a bunch of components. See https://linear.app/rerun/issue/RR-2421/tracking-issue-documentation-process-improvements for details

rerun_py/rerun_sdk/rerun/_baseclasses.py
ntjohnson1: I'm not sure if we have test coverage but this should just extend the capability rather than break anything

```python
instance = Foo(…)
instance.component_type()
# before only the top worked
Foo.component_type()
```

examples/python/notebook_callbacks/notebook_callbacks.ipynb
ntjohnson1: https://docs.astral.sh/ruff/rules/builtin-import-shadowing/

Basically ipython defines display but mypy seems to not know that content when reviewing the notebooks so we explicitly import display from ipython to make mypy happy.

PR #11443, opened 2025-10-06
Update pixi to 0.55.0 & pin version in `pixi.toml`

Wumpf:
It's finally possible to pin the minimum required pixi version in `pixi.toml`! This caused issues in the past because it was unclear what version you needed to get things working.
Might as well do a round of upgrades now, which is happening here.

* [ ] pass full-ci


pixi.toml
```diff
@@ -25,6 +25,7 @@ platforms = ["linux-64", "linux-aarch64", "osx-arm64", "osx-64", "win-64"]
 readme = "README.md"
 repository = "https://github.com/rerun-io/rerun"
 version = "0.1.0"                                                              # TODO(emilk): sync version with `Cargo.toml` with help from `crates.py`
+requires-pixi = ">=0.55.0" # Make sure to keep this in sync with the version on our CI jobs!
```
Wumpf: this is why we're here!

Wumpf:
@rerun-bot full-check

PR #11446, opened 2025-10-07
Merge `DroppableEncoder` and `Encoder` into a single facility

teh-cmc:
Our RRD encoding facilities are currently implemented by two distinct types: `DroppableEncoder` & `Encoder`.

This makes understanding how to use the API unnecessarily painful. Turns out the only reason things are done this way is because of an implementation detail related to the `Drop` implementation.
Instead of passing off that complexity to end-users of the API, we now contain it to the implementation instead.

In addition to that, the `AlreadyFinished` logic was broken, so fix too while we're at it.

---


teh-cmc:
@rerun-bot full-check

zehiko (approved):

PR #11449, opened 2025-10-07
remove GetChunks and all its usage

zehiko:
### What

As per title, removing ``GetChunks`` API and replacing usage with 2-phase query (query + fetch chunks). 

### Testing done
* existing tests
* opening partitions in the viewer


crates/store/re_redap_client/src/grpc.rs
```diff
@@ -444,6 +373,91 @@ pub async fn stream_blueprint_and_partition_from_server(
...
+) -> Result<FetchChunksResponseStream, StreamError> {
+    let fields_of_interest = [
+        "chunk_partition_id",
+        "chunk_id",
+        "rerun_partition_layer",
+        "chunk_key",
+    ]
+    .into_iter()
+    .map(String::from)
+    .collect::<Vec<_>>();
```
abey79: please please introduce a `impl QueryDatasetResponse` in `rerun.cloud.v1alpha1.ext.rs`, following the pattern of e.g. `impl QueryTasksResponse`. There should be constant for field names that you can reuse here.

It's not great, but this is how we properly document our dataframe-level public API and I would love if we can stick to this pattern (until we find a better one).

crates/store/re_redap_client/src/grpc.rs
```diff
@@ -444,6 +373,91 @@ pub async fn stream_blueprint_and_partition_from_server(
...
+
+/// Fetches all chunks for a specified partition. You can include/exclude static/temporal chunks.
+pub async fn fetch_partition_chunks(
+    client: &mut ConnectionClient,
+    dataset_id: EntryId,
+    partition_id: PartitionId,
+    exclude_static_data: bool,
+    exclude_temporal_data: bool,
+    query: Option<re_protos::cloud::v1alpha1::Query>,
+) -> Result<FetchChunksResponseStream, StreamError> {
```
abey79: This entire function could be a `ConnectionClient` method. That would count as a win w.r.t its use of `inner` :)

Ideally it would be broken into its query and fetch parts for when that's useful, but it would already be nice to be left as a TODO.


```suggestion
pub async fn fetch_partition_chunks(
    &mut self,
    dataset_id: EntryId,
    partition_id: PartitionId,
    exclude_static_data: bool,
    exclude_temporal_data: bool,
    query: Option<re_protos::cloud::v1alpha1::Query>,
) -> Result<FetchChunksResponseStream, StreamError> {
```

crates/store/re_redap_client/src/grpc.rs
zehiko: yeah, definitely can introduce these constants, but unlike ``QueryTasksResponse`` I can't really impose a schema as it's dynamic i.e. you can ask for any of the fields. Hopefully your ask is just about the constant fields (and that's a good suggestion as there's already 2 copies of these strings constants around).

crates/store/re_redap_client/src/grpc.rs
zehiko: yeah, agree :) let me make that change.

crates/store/re_redap_client/src/grpc.rs
abey79: That makes sense for the schema. Having a clear definition of at least the column names (and eventually the datatype?) is already progress.

As for the multiple definitions: yeah I've struggled with that quite a bit already. We really need to be better and single sources of truth for all of these.

PR #11450, opened 2025-10-07
Encoder: simplified APIs and improved discovery

teh-cmc:
The Encoder was offloading so much complexity to end-users of the API by forcing them to somehow discover, pick and figure out how to use a bunch of complicated freestanding helpers such as `encode_ref_as_bytes_local`.

All these freestanding functions have now been replaced by a couple of simpler, more generic, more discoverable helper methods.

---

This PR is part of an upcoming series of PRs to pay off organic growth debt in our encoding/decoding stack.

* DNM: requires #11446

PR #11452, opened 2025-10-07
`StreamDecoder`: lift naming confusion regarding chunks

teh-cmc:
This is just a pure renaming pass.

`StreamDecoder`'s job is to deal with chunks of bytes, and it used to call those `chunks`. This was extremely confusing to me when trying to understand the code, since _chunk_ is such an overloaded term for us.
This PR just makes it as clear as possible that this is about _byte chunks_, not _Rerun chunks_.

---

* DNM: requires #11450


teh-cmc:
@rerun-bot full-check

zehiko (approved):

PR #11455, opened 2025-10-07
`StreamDecoder`: fix support for out-of-order legacy store ID

teh-cmc:
This PR ports the test suite from the legacy `Decoder` to the new `StreamDecoder`, to ensure they are fully compatible.

In doing so, I stumbled upon a bug where the `StreamDecoder` would enter an obscure irrecoverable state when faced with an out-of-order legacy store ID (i.e. a store ID missing an app ID).
This PR fixes that.

Also removed some duplicated code on the way.

---

* DNM: requires #11452

### DIFF
--- a/crates/store/re_log_encoding/src/decoder/stream.rs
+++ b/crates/store/re_log_encoding/src/decoder/stream.rs
@@ -600,7 +600,6 @@ mod tests_legacy {
 
     use crate::Compression;
     use crate::Encoder;
-    use crate::codec::arrow::encode_arrow;
 
     use super::*;
 
@@ -665,7 +664,9 @@ mod tests_legacy {
     fn legacy_fake_log_messages() -> Vec<LogMsgProto> {
         fake_log_messages()
             .into_iter()
-            .map(log_msg_to_proto)
+            .map(|msg| {
+                crate::protobuf_conversions::log_msg_to_proto(msg, Compression::Off).unwrap()
+            })
             .map(|mut log_msg| {
                 match &mut log_msg.msg {
                     None => panic!("Unexpected `LogMsg` without payload"),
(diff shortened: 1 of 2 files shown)


crates/store/re_log_encoding/src/decoder/stream.rs
```diff
@@ -575,3 +586,398 @@ mod tests {
...
+
+            let decoded_messages: Vec<_> = decoder_into_iter(decoder).collect();
+            assert_eq!(decoded_messages.len(), messages.len());
+        }
+    }
+
+    /// Test that legacy messages (aka `StoreId` without an application id) that arrive _before_
+    /// a `SetStoreInfo` are dropped without failing.
+    #[test]
+    fn test_decode_legacy_out_of_order() {
```
teh-cmc, line 786: This is the test that revealed the problem.

crates/store/re_log_encoding/src/decoder/stream.rs
```diff
@@ -170,11 +172,20 @@ impl StreamDecoder {
...
-                    )?;
+                    ) {
+                        Ok(msg) => msg,
+                        Err(err) => {
+                            // We successfully parsed a header, but decided to drop the message altogether.
+                            // We must go back to looking for headers, or the decoder will just be stuck in a dead
+                            // state forever.
+                            self.state = State::MessageHeader;
+                            return Err(err);
+                        }
```
teh-cmc, line 187: And this is the entire fix. its size is not reflective of the pain involved :upside_down_face:

teh-cmc:
@rerun-bot full-check

crates/store/re_log_encoding/src/decoder/stream.rs
zehiko: it never is... :)

zehiko (approved):
🙏

PR #11456, opened 2025-10-07
Improve warning message on accidental component type change

grtlr:
### What

Title.

### DIFF
--- a/crates/store/re_chunk_store/src/writes.rs
+++ b/crates/store/re_chunk_store/src/writes.rs
@@ -382,8 +382,10 @@ impl ChunkStore {
                 && old_typ != list_array.value_type()
             {
                 re_log::warn_once!(
-                    "Component column '{}' changed type from {old_typ:?} to {:?}",
+                    "Component '{}' with component type '{}' in entity '{}' changed type from {old_typ:?} to {:?}",
+                    component_descr.component,
                     component_type,
+                    chunk.entity_path(),
                     list_array.value_type()
                 );
             }


crates/store/re_chunk_store/src/writes.rs
```diff
@@ -382,8 +382,10 @@ impl ChunkStore {
                 && old_typ != list_array.value_type()
             {
                 re_log::warn_once!(
-                    "Component column '{}' changed type from {old_typ:?} to {:?}",
+                    "Component '{}' with component type '{}' in entity '{}' changed type from {old_typ:?} to {:?}",
```
oxkitsune:
```suggestion
                    "Component '{}' with component type '{}' on entity '{}' changed type from {old_typ:?} to {:?}",
```

oxkitsune (approved):

crates/store/re_chunk_store/src/writes.rs
emilk: Pre-existing problem but:

Use `re_arrow_util::format_data_type` when formatting arrow `DataType`!  NEVER `Debug` formatting (it is extremely verbose).

In the next arrow version we can start using `Display` formatting, but never `Debug`

grtlr:
```suggestion
                    "Component '{}' with component type '{}' on entity '{}' changed type from {old_typ:?} to {:?}",
```

grtlr: Oops I missed this because I was using primitive types.

Thank you for pointing that out.

PR #11457, opened 2025-10-07
Add panel toggles to integration test harness

### LINKED ISSUE #11451: Add panel toggles to integration test harness
The integration test harness relies on UI clicks to toggle panels like `HarnessExt::toggle_blueprint_panel`.

Add proper commands to set panel state.

aedm:
### Related

Fixes #11451

### DIFF
--- a/tests/rust/re_integration_test/src/kittest_harness_ext.rs
+++ b/tests/rust/re_integration_test/src/kittest_harness_ext.rs
@@ -10,17 +10,24 @@ use re_sdk::{
     },
     log::Chunk,
 };
+use re_ui::{UICommand, UICommandSender};
 use re_viewer::{
-    SystemCommand, SystemCommandSender as _,
+    AppBlueprint, SystemCommand, SystemCommandSender as _,
     external::{
         re_chunk::{ChunkBuilder, LatestAtQuery},
         re_entity_db::EntityDb,
         re_types,
-        re_viewer_context::{self, ViewerContext, blueprint_timeline},
+        re_viewer_context::{self, BlueprintUndoState, ViewerContext, blueprint_timeline},
     },
     viewer_test_utils::AppTestingExt as _,
 };
 use re_viewport_blueprint::ViewportBlueprint;
+
+pub struct PanelStates {
+    pub blueprint_panel_open: bool,
+    pub selection_panel_open: bool,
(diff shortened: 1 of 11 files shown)


tests/rust/re_integration_test/src/kittest_harness_ext.rs
```diff
@@ -230,4 +228,39 @@ impl HarnessExt for egui_kittest::Harness<'_, re_viewer::App> {
             &SnapshotOptions::new().failed_pixel_count_threshold(20),
         );
     }
+
+    fn set_panel_states(&mut self, panel_states: PanelStates) {
```
lucasmerlin: Hmm, this definitely works, but I think it'd be cooler if this would work by actually querying the current states of the buttons and clicking (or not clicking) to set the right states. State of the buttons should be exposed to AccessKit in some way, and if it isn't, we should fix that.

lucasmerlin (changes requested):

tests/rust/re_integration_test/src/kittest_harness_ext.rs
aedm: Printing the button gives me this:
```json5
Node {
    id: #8895902437236127807,
    role: Button,
    label: "Blueprint panel toggle",
    focused: false,
    hidden: false,
    disabled: false,
    children: [],
}
```
I'm not sure how to get the state out of the button using Kittest. What do you suggest?

tests/rust/re_integration_test/src/kittest_harness_ext.rs
```diff
@@ -230,4 +233,15 @@ impl HarnessExt for egui_kittest::Harness<'_, re_viewer::App> {
...
     }
+
+    fn set_panel_opened(&mut self, panel_label: &str, opened: bool) {
+        let node = self.get_by_label(panel_label);
+        let is_open = Some(Toggled::True) == node.accesskit_node().data().toggled();
+        if is_open != opened {
+            self.click_label(panel_label);
+        }
+        // The toggle button is hovered now, move the pointer away
+        self.input_mut().events.push(egui::Event::PointerGone);
```
emilk: Can also use this now:
https://docs.rs/egui_kittest/latest/egui_kittest/struct.Harness.html#method.remove_cursor

PR #11459, opened 2025-10-08
Reflection-based `protobuf` parser only returns single component

grtlr:
### Related

* Closes RR-2552.
* Related to RR-2553.
* Related to #11394.
* [x] Requires #11458.

 
### What

This changes our `protobuf` reflection code to put the entire message behind a single component to make it easier to extract it from lenses. This also guarantees that each protobuf message is always self-contained.


crates/utils/re_mcap/src/layers/protobuf.rs
```diff
@@ -166,6 +157,36 @@ fn downcast_err<'a, T: std::any::Any>(
     })
 }
 
+fn append_null_to_builder(builder: &mut dyn ArrayBuilder) -> Result<(), ProtobufError> {
```
oxkitsune, line 159: would be nice if `ArrayBuilder` implemented `append_null` 🥲

oxkitsune (approved):
very nice!

PR #11460, opened 2025-10-08
Represent `std_msgs/msg/Empty` as empty list

oxkitsune:
### What

Represent `std_msgs/msg/Empty` as empty list instead of null. If this isn't done like this, the message does _not_ show up in the timeline view or query api.


ntjohnson1 (approved):
This appears to have fixed the motivating issue. A test would be nice but maybe we punt on that for follow up if we run into issues here again.

PR #11461, opened 2025-10-08
Update to latest `egui/main`

emilk:
More testing before next egui release.

All the new changes:
* https://github.com/emilk/egui/compare/f6fe3bff180d8f93ebd4ccf7c29201cf2fb8894d...683214bfe8cb63bc9b5cc9a754ab80aab836b3e1

I updated a bunch of dependencies in Rerun using `cargo upgrade`, thus actually changing their minimum version in `Cargo.toml`. That means there is a lot of updated 3rd party crates here too… for better or for worse.

I can try to split those out in a separate PR… but it will be a lot of work methingks.


lucasmerlin (approved):
Nice!

PR #11465, opened 2025-10-08
Explain to developers how to type ellipsis

emilk:
Apparently @Wumpf had no idea that there was a way to type this on mac 🙈


ntjohnson1:
> Apparently @Wumpf had no idea that there was a way to type this on mac 🙈

I just copy paste it from the error message every time

PR #11466, opened 2025-10-08
Update some crate dependencies

emilk:
* Part of https://github.com/rerun-io/rerun/issues/11467

Using `cargo upgrade`: https://crates.io/crates/cargo-edit

only patch updates in this one

### DIFF
--- a/crates/build/re_dev_tools/src/build_examples/example.rs
+++ b/crates/build/re_dev_tools/src/build_examples/example.rs
@@ -333,10 +333,7 @@ impl Frontmatter {
         let start = start + START.len();
 
         let Some(end) = content[start..].find(END) else {
-            anyhow::bail!(
-                "{:?} has invalid frontmatter: missing {END:?} terminator",
-                path
-            );
+            anyhow::bail!("{path:?} has invalid frontmatter: missing {END:?} terminator");
         };
         let end = start + end;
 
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -186,33 +186,33 @@ walkers = "0.45"
...
-camino = "1.1"
+bytes = "1.10"
+camino = "1.2"
 cargo_metadata = "0.18"
 cargo-run-wasm = "0.3.2"
 cdr-encoding = "0.10.2"
 cfg_aliases = "0.2"
 cfg-if = "1.0"
-chrono = { version = "0.4.39", default-features = false } # Needed for datafusion, see `re_datafusion`'s Cargo.toml
+chrono = { version = "0.4.42", default-features = false } # Needed for datafusion, see `re_datafusion`'s Cargo.toml
```
abey79, line 209: are we sure we have no requirement to be in sync with datafusion here?

Cargo.toml
```diff
@@ -232,148 +232,148 @@ datafusion = { version = "49.0.2", default-features = false, features = [
...
-protoc-prebuilt = "0.3"
+prost-types = "0.13.5"
+protoc-prebuilt = "0.3.0"
 puffin = "0.19.1"
-puffin_http = "0.16"
-pyo3 = "0.24.1"
-pyo3-build-config = "0.24.1"
+puffin_http = "0.16.1"
+pyo3 = "0.24.2"
+pyo3-build-config = "0.24.2"
```
abey79, line 317: these always make me nervous since it is part of the pyo3-arrow-dfn-lance gang. but maybe wrongly so for patch version?

abey79 (approved):
whatever CI says :)

Cargo.toml
emilk: on a patch version? that would be very surprising

mosesjames7271-svg: Yeah it's what giving a try

mosesjames7271-svg (approved):

PR #11468, opened 2025-10-08
Refactor server implementation and introduce `Layer` structure

abey79:
### Related

* part of https://linear.app/rerun/issue/RR-2482/add-layer-support-to-oss-server

### What

This PR is 90% splitting OSS server implementation in multiple files and 10% introducing a `Layer` structure in the OSS server implementation. This does NOT yet actually implement layers—the actual API coverage is unchanged (hence no test or anything). This comes next, but I wanted to have the refactor out first to keep review sane.


crates/store/re_server/src/store/dataset.rs
```diff
@@ -0,0 +1,203 @@
...
+        self.updated_at = jiff::Timestamp::now();
+    }
+
+    /// Load a RRD using its recording id as partition id.
+    pub fn load_rrd(
+        &mut self,
+        path: &Path,
+        layer_name: Option<&str>,
+        on_duplicate: IfDuplicateBehavior,
+    ) -> Result<BTreeSet<PartitionId>, Error> {
```
timsaucer, line 163: One thing I don't quite have a handle on is what adding a layer looks like from the CLI. I do wonder if we will need to go to a config file approach instead of command line parameters.

abey79: I agree this is a bit of a blind spot.

TBH, these command line arguments are weird to begin with, because they essentially try to be a CLI frontend to grpc API (though currently it's taking a shortcut implementation-wise). In the future, we should split the `redap` CLI into a tool that only covers the public API, and is moved to this repo. Then it could be used to fully populate a local server from terminal.

In the mean time, I think it's find to have the CLI being vastly underpowered (e.g., not support layers), unless we really need it e.g. for CI.

Issue #1025, opened 2023-02-01
Treat camera projected primitives in 3D space more like 2D objects


Wumpf:
Consider a camera projection (typically perspective, but orthographic shouldn't make a difference) in 3D space.
![image](https://user-images.githubusercontent.com/1220815/215998740-ea2a24d3-d741-479c-8900-f79eab870699.png)
Today, the renderer renders all shapes that are part of the camera projection (there might not only be images but also lines, points or even complex meshes) just like anything else in the scene. I.e. the user is expected to set the respective projection matrix as world matrix for all these objects.

This works nicely for the most part but there are are some things we need to fix (the actual tasks of this issue!): 
* any "viewspace generated geometry" needs to behave like it is being looked with the camera in question. Points & lines which today follow along the `Eye`/main camera should actually follow their camera, making them appear flat on the surface instead
* auto sizes need to be treated like they live in the realm their camera
* _...TODO other?_

Open questions:
* how should we communicate these things to the renderer in a somewhat unified manner? Depending on depth handling, not all objects need the same amount of knowledge
* what with remaining depth? I.e. when we had 3D objects under the frustum that would have yielded a depth buffer?
  * Bunch of options, maybe should expose them?
     * flatten them onto the plane, i.e. ignore their depth for everything but depth offsetting (!), making it look like the same in a 2D space view
         * Primitive depth offsetting only goes that far for larger ranges
     * use the  "virtual depth buffer depth", i.e. objects are distorted within the confines of the camera frustum
     * use view space depth, making it more behave like one would expect a depth map that someone placed into 3D\
* Should there be clipping to the frustum? If so, probably as an option.
   * is there a way to achieve this without expensive `discard` instructions in the shader?
* _...TODO other?_

--------------

Note that the seemingly straight forward solution would be to have a literal 2d surface that we render in 3D space. There is a whole lot of drawbacks with this though as well:
* 2D surface would have limited resolution, meaning we're loosing a lot of quality (there is no "good enough" resolution in this resampling problem!)
* Need another render to texture pass which is something our renderer doesn't have either!
* Very hard to handle "depth after projection", if at all we get a literal depth buffer

Wumpf:
Since we're also hitting issues in compositing pure 2D views we need a more radical approach:

* allow defining an arbitrary amount of `RenderCamera` which are added to `ViewBuilder`
   * each camera has its own draw data objects, they are no longer added to `ViewBuilder`
[...]

PR #11470, opened 2025-10-08
Perform transform tree walk only once per frame, rather than for every View

### LINKED ISSUE #7025: Simplify transform hierachy walk & computation
Today, each space view has its own transform context which walks up and down the tree going from `$origin` in order to figure out transform from each entity/instance to the "reference space", i.e. the space whose origin is at the `$origin` entity.

So today the graphics pipeline conceptually does the following:
```
[instance -> entity -> origin-reference] -> camera -> projection
```
Where the `TransformContext` is providing `[instance -> entity -> origin-reference]`

Disregarding (do we have to?) any 2D->3D/3D->2D consideration we could do something much simpler:
First, resolve the transform tree starting at `/`, let's call this the world space. This leads to the following transform relationship:
```
[instance -> entity -> world] -> origin-reference -> camera -> projection
```
Note that now there is a global transform context providing `[instance -> entity -> world]`. From there on we can either infect re_renderer with this as well - or much simpler - go with this world space and bake `[origin-reference ->camera]` into the camera matrix we're setting.

-> We end up with a single transform tree pass and don't have to care about walking the tree upwards!

The main issue with doing this today is that 2D in 3D handling makes this complicated in some regards.
Also, there might be several independent 3D spaces that have no connection to the root!

Related, maybe even blocked by:
* https://github.com/rerun-io/rerun/issues/1025

Wumpf:
### Related

* Fixes #7025
* Fixes RR-2516
* Part of #11445 
* Related/unblocking for https://github.com/rerun-io/rerun/issues/1025

### What

Also, our transform tree is now a `TransformForest` as we treat spaces under pinholes as separate trees.
They are then recombined into a transform tree on a per view basis as this is now seen as a visualization detail, rather than a property of our transform.
Crucially though, even disregarding peculiarities of 2D subtrees, we're now walking only once from the root! Then, for any given space origin ("target") we re-orient the tree, not performing a full walk.

Initial motivation for this refactor is to allow more complex tree walks more easily & split out more state/work from the View and move it into the view-agnostic `re_tf`.
I've identified this as a pre-requisite for moving towards transform-frame-based transforms (ROS transform support)!

Future related work: our blueprint heuristic supporting topology analysis is rather redundant to what `TransformForest` does, we can likely at least share some code there

### Performance

All on Mac M1. Screenshots are zoomed in on `AppState::show`

```sh
pixi run rerun-release --threads 1 
pixi run -e py python ./tests/python/many_entity_transforms/main.py
```

Before: ~72ms for app show
image

After: ~66ms for app show
image

---

with `pixi run -e py python ./tests/python/many_entity_transforms/main.py --num-views 1`

Before @ 707c55a3b5e43a4976ac1489f29e5b3e3df893dc: ~13.3ms for app show

image

After @ [25a7b5e](https://github.com/rerun-io/rerun/pull/11470/commits/25a7b5e2d3d0f7a2e2d9656cf415c9a43f1e43f5): ~13.1ms for app show
image


crates/store/re_tf/src/transform_forest.rs
```diff
@@ -0,0 +1,872 @@
...
+                if let Some(pinhole_projection) = transforms_at_entity.pinhole_projection {
+                    let new_root = child_path.hash();
+                    let new_root_info = TransformTreeRootInfo::Pinhole(PinholeTreeRoot {
+                        parent_tree_root: root,
+                        pinhole_projection: pinhole_projection.clone(),
+                        parent_root_from_pinhole_root: root_from_child,
+                    });
+
+                    let previous_root = self.roots.insert(new_root, new_root_info);
+                    debug_assert!(previous_root.is_none(), "Root was added already"); // TODO(andreas): Build out into cycle detection (cycles can't _yet_ happen)
```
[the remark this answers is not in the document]
Wumpf: no, don't be silly

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -1,48 +1,89 @@
...
         "TransformContext".into()
     }
 }
 
+struct TransformTreeContextStaticExecResult {
+    transform_forest: Arc<re_tf::TransformForest>,
+}
+
 impl ViewContextSystem for TransformTreeContext {
+    fn execute_static(
```
Wumpf: via @grtlr `execute_once`?

Wumpf: maybe `execute_once_per_frame` & `execute_per_view` to make it obvious :)

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -51,34 +92,46 @@ impl ViewContextSystem for TransformTreeContext {
 }
 
 impl TransformTreeContext {
+    /// Traransform info to get from the entity's transform to the origin transform frame.
```
Wumpf:
```suggestion
    /// Transform info to get from the entity's transform to the origin transform frame.
```

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -0,0 +1,872 @@
...
+use re_log_types::EntityPathHash;
+use re_types::ArchetypeName;
+
+use crate::{
+    CachedTransformsForTimeline, PoseTransformArchetypeMap, ResolvedPinholeProjection,
+    TransformResolutionCache, image_view_coordinates,
+};
+
+/// Transform from an entity towards a given root.
+#[derive(Clone, Debug, PartialEq)]
```
emilk: What does "towards" mean here? Is it even relevant?

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -0,0 +1,872 @@
...
+pub struct TransformInfo {
+    /// Root this transform belongs to.
+    ///
+    /// ⚠️ This is the root of the tree this transform belongs to,
+    /// not necessarily what the transform transforms into.
+    root: EntityPathHash,
```
emilk, line 27: why aren't we storing the `target` too?

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -0,0 +1,872 @@
...
+    target_from_entity: glam::Affine3A,
+
+    /// List of transforms per instance including poses.
+    ///
+    /// If no poses are present, this is always the same as `root_from_entity`.
+    /// (also implying that in this case there is only a single element).
+    /// If there are poses there may be more than one element.
+    ///
+    /// Does not take into account archetype specific transforms.
+    target_from_instances_overall: SmallVec1<[glam::Affine3A; 1]>,
```
emilk: What does `overall` mean in this case?

emilk: Maybe a module-level `//!` docstring at the top of the file would be a good place to introduce some of these concepts?

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -0,0 +1,872 @@
...
+
+    /// Like [`Self::target_from_instances_overall`] but _on top_ also has archetype specific transforms applied
+    /// if there are any present.
+    target_from_archetype: IntMap<ArchetypeName, SmallVec1<[glam::Affine3A; 1]>>,
```
emilk: So a single entity could have some sphere-poses and some different box-poses, for instance?

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -0,0 +1,872 @@
...
+            TransformInfo::new_root(EntityPath::root().hash()),
+            transforms,
+        );
+
+        tree
+    }
+}
+
+impl TransformForest {
+    #[allow(clippy::too_many_arguments)]
```
emilk: you need to merge in `main` - I've forbidden `allow` attributes. You need to use `expect` now!

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -0,0 +1,872 @@
...
+        transform_root_from_parent: TransformInfo,
+        transforms_for_timeline: &CachedTransformsForTimeline,
+    ) {
+        let root = transform_root_from_parent.root;
+        let root_from_parent = transform_root_from_parent.target_from_entity;
+
+        let previous_transform = self
+            .root_from_entity
+            .insert(subtree.path.hash(), transform_root_from_parent);
+        debug_assert!(previous_transform.is_none(), "Root was added already"); // TODO(andreas): Build out into cycle detection (cycles can't _yet_ happen)
```
emilk: Maybe TODO should link to ROS-tf issue?

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -0,0 +1,872 @@
...
+                // Don't care about instance transforms on the target frame, as they don't tree-propagate.
+                target_from_instances_overall: _,
+                // Don't care about archetype specific transforms on the target frame, as they don't tree-propagate.
+                target_from_archetype: _,
+            } = &root_from_target;
+
+            TargetInfo {
+                id: target,
+                root: *target_root,
+                target_from_root: root_from_entity.inverse(),
```
emilk, line 397: The fact that we do inverses makes me think we should just use `f64` for the transform tree

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -0,0 +1,872 @@
...
+    // Our visualization interprets this as looking at a 2D image plane from a single point (the pinhole).
+
+    // Center the image plane and move it along z, scaling the further the image plane is.
+    let focal_length = image_from_camera.focal_length_in_pixels();
+    let focal_length = glam::vec2(focal_length.x(), focal_length.y());
+    let scale = pinhole_image_plane_distance / focal_length;
+    let translation =
+        (-image_from_camera.principal_point() * scale).extend(pinhole_image_plane_distance);
+
+    let image_plane3d_from_2d_content = glam::Affine3A::from_translation(translation)
```
emilk, line 628: There's a lot of `glam::Affine3A` in this file. Feels like `use glam::Affine3A;` could save a lot of noise

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
emilk: `static` makes me think of non-temporal data

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
emilk: "Pay the traransom, or we execute the code!"

Wumpf:
> I suspect it would be a good idea to use a [f64 affine transform](https://docs.rs/glam/latest/glam/f64/struct.DAffine3.html) for our CPU-side transforms, but let's save it for a different PR

Is on the plan already, yep.
Didn't realize `glam` has an f64 namespace (didn't get far enough to look 😄 ), that's great!

Wumpf:
Thanks for jumping in for @oxkitsune who's super been busy with release things - we had a [walkthrough](https://rerunio.slack.com/archives/C09J226VB46/p1760012330093169) but it's always great to spread the knowledge and get eyes & feedback from more sides! :)

crates/store/re_tf/src/transform_forest.rs
Wumpf: size consideration. I want to keep this thing small since we move it around a lot.

target would add only 8 bytes. And if we switch everything to double in here then that's like nothing actually :/

crates/store/re_tf/src/transform_forest.rs
Wumpf: yeah good idea, I'll better start with that now and adjust as we go.
That `overall` term was taken over from `TransformTree` but I can totally see how it's mega unclear coming into it.

crates/store/re_tf/src/transform_forest.rs
Wumpf: thanks for the hint

crates/store/re_tf/src/transform_forest.rs
Wumpf: precisely. I'll add that as an example to the doc string
This is essentially early-resolve of instance-pose-like transforms. It's sooort of a hack we had to introduce when making the transforms on sphere/box/etc. no longer be equal to instance poses

crates/store/re_tf/src/transform_forest.rs
Wumpf: there's so many I created on that project :D
will do

crates/store/re_tf/src/transform_forest.rs
Wumpf: RR-2493

crates/store/re_tf/src/transform_forest.rs
Wumpf: will push this into a follow-up PR that I have in the works already, but renaming this field to drop the `overall` part

crates/store/re_tf/src/transform_forest.rs
Wumpf: taste thing from my C++ days: small namespaces like `wgpu` or `glam` I like to never put in a use statement and be explicit about them :shrug:

PR #11471, opened 2025-10-08
Introduce `WrapperComponent` to make serialization impl for most components generic

Wumpf:
Not sure if that's a compile time win (haven't measured), more likely it is not since generics == bad. But it does remove a ton of boilerplate code from component implementation 🤷

### DIFF
--- a/crates/store/re_types_core/src/wrapper_component.rs
+++ b/crates/store/re_types_core/src/wrapper_component.rs
@@ -37,6 +37,7 @@ impl<T: WrapperComponent> Loggable for T {
         <Self as WrapperComponent>::Datatype::arrow_datatype()
     }
 
+    // NOTE: Don't inline this, this gets _huge_.
     fn to_arrow_opt<'a>(
         data: impl IntoIterator<Item = Option<impl Into<::std::borrow::Cow<'a, Self>>>>,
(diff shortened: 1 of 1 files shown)


crates/store/re_types_core/src/wrapper_component.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+impl<T: WrapperComponent> Loggable for T {
+    #[inline]
+    fn arrow_datatype() -> arrow::datatypes::DataType {
+        <Self as WrapperComponent>::Datatype::arrow_datatype()
+    }
+
+    fn to_arrow_opt<'a>(
```
emilk, line 41: I think the old "// NOTE: Don't inline this, this gets _huge_.` belongs here and on `from_arrow_opt`… maybe?

Wumpf:
For posterity: I experimented with a more aggressive variant that makes `WrapperComponent` a struct and all components just type aliases on `andreas/wrapper-component-struct`. It was sexy but ran into too many subtle issues

crates/store/re_types_core/src/wrapper_component.rs
Wumpf: good callout

PR #11478, opened 2025-10-09
Remove re_global_context

IsseW:
### What

The `re_global_context` crate was no longer used anywhere where `re_viewer_context` wasn't used, and created a lot of friction because of having to move stuff down to the `re_global_context` crate for `SystemCommand`s.

So this PR moves `re_global_context` into `re_viewer_context`.

### DIFF
--- a/crates/top/rerun/Cargo.toml
+++ b/crates/top/rerun/Cargo.toml
@@ -202,7 +202,4 @@ re_build_tools.workspace = true
 
 [package.metadata.cargo-shear]
 # We only depend on re_video so we can enable extra features for it
-ignored = [
-  "re_video",
-  "puffin",
-]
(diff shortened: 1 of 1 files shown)


crates/viewer/re_viewer_context/src/global_context.rs
```diff
@@ -0,0 +1,40 @@
+use crate::{AppOptions, CommandSender, DisplayMode};
+
+/// Application context that is shared across all parts of the viewer.
+///
+/// This context, in difference to [`crate::ViewerContext`] can exist for
+/// any arbitrary state of the viewer. And not only when there is an open
+/// recording.
+pub struct GlobalContext<'a> {
```
Wumpf, line 8: thanks for adding the doc!

PR #11482, opened 2025-10-09
Adds blueprint container handling in integration tests

aedm:
### Related

* Part of #8948 

### What

- Adds container handling to the integration test harness
- Adds a test that checks the context menu when clicking on views in the selection panel
- Removes `check_context_menu_single_selection.py` from the release checklist


tests/rust/re_integration_test/src/kittest_harness_ext.rs
```diff
@@ -249,4 +258,42 @@ impl HarnessExt for egui_kittest::Harness<'_, re_viewer::App> {
...
+    ) -> ContainerId {
+        // Collect list of container ids before adding new container
+        let old_container_ids = Arc::new(Mutex::new(BTreeSet::<ContainerId>::new()));
+        let old_container_ids_clone = old_container_ids.clone();
+        self.setup_viewport_blueprint(move |_viewer_context, blueprint| {
+            old_container_ids_clone
+                .lock()
+                .extend(blueprint.containers.keys().copied());
+        });
+
```
emilk, line 290: Having to use `Arc/Mutex` here smells to me.

I think we should instead set up our code to ensure this works:


```suggestion
        // Collect list of container ids before adding new container
        let old_container_ids: BTreeSet::<ContainerId> = 
	        self.setup_viewport_blueprint(move |_viewer_context, blueprint| {
	            blueprint.containers.keys().copied().collect()
	        });
```

`fn setup_viewport_blueprint` and/or `fn run_with_viewer_context` should take care of the details, ensuring that the given closure gets called exactly once.

tests/rust/re_integration_test/src/kittest_harness_ext.rs
```diff
@@ -249,4 +258,42 @@ impl HarnessExt for egui_kittest::Harness<'_, re_viewer::App> {
...
+        });
+        self.run_ok();
+
+        // Find the new container id
+        new_container_ids
+            .lock()
+            .difference(&old_container_ids.lock())
+            .next()
+            .copied()
+            .expect("Failed to find new container id")
```
emilk: Let's also ensure the difference is exactly 1

PR #11485, opened 2025-10-09
Wait for test server to shutdown

emilk:
I enabled the `clippy::unused_async` which surfaced a problem with our test shutdown: it looked like we where trying to wait for the shutdown to complete… but we weren't waiting.

### Related
* Sibling: https://github.com/rerun-io/dataplatform/pull/1843

### DIFF
--- a/crates/store/re_server/src/server.rs
+++ b/crates/store/re_server/src/server.rs
@@ -74,7 +74,7 @@ impl ServerHandle {
     }
 
     /// Signal to the gRPC server to shutdown, and then wait for it.
-    pub async fn shutdown_blocking(mut self) {
+    pub async fn shutdown_and_wait(mut self) {
         if let Some(shutdown) = self.shutdown.take() {
             shutdown.send(()).ok();
             self.wait_for_shutdown().await;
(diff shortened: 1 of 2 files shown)


crates/store/re_server/src/server.rs
```diff
@@ -73,9 +73,12 @@ impl ServerHandle {
         self.failed.recv().await;
     }
 
-    /// Signal to the gRPC server to shutdown.
-    pub async fn shutdown(self) {
-        let _ = self.shutdown.send(());
+    /// Signal to the gRPC server to shutdown, and then wait for it.
+    pub async fn shutdown_blocking(mut self) {
```
Wumpf:
```suggestion
    pub async fn shutdown(mut self) {
```

clearly it's not blocking. It returns immediately, so if anything it's `shutdown_async`

emilk: I renamed it `shutdown_and_wait`. I want to make it clear that it NOT just _triggers_ a shutdown, but `await`ing the future will wait until the shutdown is complete

PR #11487, opened 2025-10-09
Scaffolding for a shareable, reference test suite for redap

abey79:
### Related

* sibling https://github.com/rerun-io/dataplatform/pull/1844

### What

There is no point bringing the OSS server to parity if that parity is neither verified, nor enforced. This PR introduces the scaffolding for a reference Rerun Data Protocol (redap) test suite.

Quoting its README.md:

> This test suite is specifically focused on the redap layer.  In particular it aims to cover what our API's `*.proto` files leave implicit. This includes at least:
> - all the dataframes (schema, content)
> - all the stateful behaviors (e.g. chunk keys, tasks, etc.)
> 
> As such, it is implemented to be as close as possible to the actual API boundary, aka the (incorrectly named) `RerunCloudService` trait.
> 
> ### Goals
> 
> - Cover all aspects of the redap layer, including dataframe schemas and stateful behaviors.
> - Serve as the definitive reference of what redap is.
> - Ensure conformance of all implementations (including, possibly, third-party).
> 
> ### Non-goals
> 
> - Test layers outside the redap boundary, including `re_redap_client::ConnectionClient` or the Python SDK.
> - Test anything about the internals of the redap implementors (OSS server, Rerun Cloud, etc.)

Currently, this test suite contains a single test as proof-of-concept. I've picked one that I thought would pass with both implementations. Of course, it didn't, so this PR also fixes the OSS server so that test passes.

Notes:
- All the code in `util` has been copied straight from the `dataplatform`.
- The crate is marked as `publish = true`. Primarily because CI told me to, but it also kind of make sense given that this crates serves as reference on redap.

### DIFF
--- a/crates/store/re_redap_tests/src/lib.rs
+++ b/crates/store/re_redap_tests/src/lib.rs
@@ -3,7 +3,7 @@
 //! ## Usage
 //!
 //! In the crate containing your implementation of the
-//! [`re_protos::cloud::v1alpha1::rerun_cloud_service_server::RerunCloudService` trait, add an
+//! [`re_protos::cloud::v1alpha1::rerun_cloud_service_server::RerunCloudService`] trait, add an
 //! integration test with the following content:
 //!
 //! ```ignore


crates/store/re_server/tests/redap_tests.rs
teh-cmc, line 1: sweet

PR #11488, opened 2025-10-10
Construct an empty struct for empty ROS2 message definitions

oxkitsune:
### Related

* #11480 

### What

This treats empty messages as a special case again, and correctly constructs an empty struct.

### DIFF
--- a/crates/utils/re_mcap/src/layers/ros2_reflection.rs
+++ b/crates/utils/re_mcap/src/layers/ros2_reflection.rs
@@ -5,8 +5,8 @@ use arrow::{
     array::{
         ArrayBuilder, ArrowPrimitiveType, BooleanBuilder, FixedSizeListBuilder, Float32Builder,
         Float64Builder, Int8Builder, Int16Builder, Int32Builder, Int64Builder, ListBuilder,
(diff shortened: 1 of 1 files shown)


crates/utils/re_mcap/src/layers/ros2_reflection.rs
```diff
@@ -96,11 +96,21 @@ impl ArrayBuilder for MessageStructBuilder {
     }
 
     fn finish(&mut self) -> arrow::array::ArrayRef {
-        Arc::new(self.builder.finish())
+        // Special case for empty structs, e.g. `std_msgs/Empty`
```
emilk, line 99: I can see from the code that we're special-casing it, but I cannot see _why_.

Comment the _why_, not the _what_

oxkitsune: This is an attempt to fix an issue when querying out these empty columns. However, it looks like it might be a datafusion issue: RR-2653.

Marking this as draft until then.

ntjohnson1:
We can close this since we root caused this to a bug in datafusion and on our newer datafusion it is resolved.

PR #11490, opened 2025-10-10
Put playback speed and fps in blueprint

### LINKED ISSUE #5577: Set playback speed from (blueprint) SDK
The user should be able to set the playback speed through the SDK.

IsseW:
### Related

- Closes #5577

### What

Adds playback speed and fps to the time panel blueprint. Both allow to configure via python API and making it persisted state.

### DIFF
--- a/crates/store/re_types/src/blueprint/components/fps.rs
+++ b/crates/store/re_types/src/blueprint/components/fps.rs
@@ -19,7 +19,7 @@ use ::re_types_core::{ComponentBatch as _, SerializedComponentBatch};
 use ::re_types_core::{ComponentDescriptor, ComponentType};
 use ::re_types_core::{DeserializationError, DeserializationResult};
 
-/// **Component**: Frames per seconds for a sequence timeline.
+/// **Component**: Frames per second for a sequence timeline.
 ///
 /// ⚠️ **This type is _unstable_ and may change significantly in a way that the data won't be backwards compatible.**
 #[derive(Clone, Debug, Copy, PartialEq, PartialOrd)]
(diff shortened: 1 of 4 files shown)


rerun_py/rerun_sdk/rerun/blueprint/api.py
```diff
@@ -483,11 +491,16 @@ def __init__(
...
+        self.fps = fps
+
     def _log_to_stream(self, stream: RecordingStream) -> None:
         """Internal method to convert to an archetype and log to the stream."""
         arch = TimePanelBlueprint(
             state=self.state,
             timeline=self.timeline,
+            playback_speed=self.playback_speed,
+            fps=self.fps,
         )
```
oxkitsune, line 504: I know it wasn't there before this either, but I noticed this doesn't include `self.time`, which is part of the archetype right?

crates/store/re_types/definitions/rerun/blueprint/components/fps.fbs
```diff
@@ -0,0 +1,15 @@
+namespace rerun.blueprint.components;
+
+// ---
+
+/// Frames per seconds for a sequence timeline.
```
oxkitsune:
```suggestion
/// Frames per second for a sequence timeline.
```

rerun_py/rerun_sdk/rerun/blueprint/api.py
IsseW: Yep, that's included below because we keep time as a static component for it to not affect the undo buffer

PR #11492, opened 2025-10-10
Implement `Op::flatten` in lenses

grtlr:
### Related

* Part of RR-2554

### What

This adds a `flatten` operation that turns a row that contains a list array of list arrays into a single list array.


emilk (commented):
This should have a test!

One test case I'm curious about is an input where the underlying array has elements that the `offsets` skip, e.g.

```
values: [0, 1, 2, 3, 4, 5, 6]
validity: [true, false, true]
offsets: [0, 2, 4]

logically: [[0, 1], null, [4, 5, 6]]

flattened: [0, 1, null, 4, 5, 6] (I guess?)
```

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -75,20 +82,31 @@ impl Op {
         Self::func(move |_| Ok(value.clone()))
     }
 
+    /// Flattens a list array inside a component.
+    ///
+    /// Takes `List<List<T>>` and flattens it to `List<T>` by concatenating all inner lists
+    /// within each outer list row.
+    pub fn flatten() -> Self {
```
emilk, line 98: What happens to nulls?

like `[[1, 2, 3], [4, null, 5], null, [6]].flatten()` -> ???

grtlr:
It uses this transformation: https://github.com/rerun-io/rerun/blob/98b93fa7277a80b475234fdbb20467d2bdc2e8ee/crates/utils/re_arrow_util/src/transform.rs#L526

There's a test for this here: https://github.com/rerun-io/rerun/blob/98b93fa7277a80b475234fdbb20467d2bdc2e8ee/crates/utils/re_arrow_util/src/transform.rs#L971

crates/top/re_sdk/src/lenses/ast.rs
grtlr: It looks like the `[4, null, 5]` is not covered by my test case yet. I will add that—thank you for pointing that out.

The result should be `[1, 2, 3, 4, null, 5, 6]`.

emilk: Please document that in the docstring!

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -38,15 +41,19 @@ pub enum Op {
     /// Efficiently casts a component to a new `DataType`.
     Cast(op::Cast),
 
+    /// Flattens a list array inside a component.
+    Flatten,
```
emilk, line 51: Document what happens to inner and outer nulls!

emilk (approved):

PR #11494, opened 2025-10-10
Blueprint plot background

IsseW:
### Related

* Part of RR-2350

### What

Adds blueprint background options for time series view, and bar view. To change background color and enable/disable grid.

Screenshot 2025-10-10 at 13 50 46

### DIFF
--- a/crates/viewer/re_view_time_series/src/view_class.rs
+++ b/crates/viewer/re_view_time_series/src/view_class.rs
@@ -460,9 +460,6 @@ impl ViewClass for TimeSeriesView {
         };
         state.time_offset = time_offset;
 
-        // use timeline_name as part of id, so that egui stores different pan/zoom for different timelines
-        let plot_id_src = ("plot", &timeline_name);
-
         let lock_y_during_zoom = y_zoom_lock;
 
         // We don't want to allow vertical when y is locked or else the view "bounces" when we scroll and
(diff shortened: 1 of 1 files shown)


crates/store/re_types/definitions/rerun/blueprint/archetypes/background.fbs
```diff
@@ -17,3 +17,14 @@ table Background (
   // TODO(andreas): Can't link to [components.BackgroundKind.SolidColor] since blueprint components aren't part of the doc page yet.
   color: rerun.components.Color ("attr.rerun.component_optional", nullable, order: 2000);
 }
+
+/// Configuration of a background in a plot view.
+table PlotBackground (
```
oxkitsune, line 22: I think the view is called `TimeSeriesView`, wouldn't it be more consistent to call the component `TimeSeriesBackground`?

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -463,8 +480,11 @@ impl ViewClass for TimeSeriesView {
 
         let min_axis_thickness = ui.tokens().small_icon_size.y;
 
+        ui.style_mut().visuals.extreme_bg_color = background_color.into();
```
oxkitsune: shouldn't this be in a `ui.scope`? e.g. won't this leak?
in bar chart it is done, not sure why it wasn't done here before?

crates/store/re_types/definitions/rerun/blueprint/archetypes/background.fbs
IsseW: I was thinking for this `PlotBackground` as an archetype made more sense since it's also used for `BarChartView`

crates/viewer/re_view_time_series/src/view_class.rs
IsseW: Didn't see anything break, but yes this is a good point

PR #11495, opened 2025-10-10
Disable toast when copying component path

oxkitsune:
### Related

* Part of RR-2629

### What

Removes the notification that pops up when copying a component path.

Originally the toast said: 
```
Copied entity path to clipboard: /__properties
```

2025-10-10_14-32-12@2x

### DIFF
--- a/crates/viewer/re_viewer_context/src/item_collection.rs
+++ b/crates/viewer/re_viewer_context/src/item_collection.rs
@@ -210,7 +210,6 @@ impl ItemCollection {
             AppId,
             StoreId,
             EntityPath,
-            Component,
         }
 
         #[expect(clippy::match_same_arms)]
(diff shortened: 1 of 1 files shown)


oxkitsune:
Adding this for `Archetype` and `Component Type` is a bit difficult without a larger refactor, as there's no way to discern whether the text is a component or archetype:

image

I have no idea why the text that ends up on the clipboard is actually correct when copying an archetype, when I print out the `clipboard_text` here:
https://github.com/rerun-io/rerun/blob/e6c9fbe95a9c2b114c92d75f4cf64a64ebbf692f/crates/viewer/re_viewer_context/src/item_collection.rs#L292-L297

I get `McapStatistics:channel_count` but `McapStatistics` actually gets copied to the clipboard.

grtlr:
Ok, I've done some more testing. The behavior of this RP is an improvement to what was there before, because now at least it's correct for `component`.

The issue mentioned in RR-2629 now happens for `archetype` and `component_type`. I'm fine with merging this PR, but it should not automatically close the linked issue, instead we should refactor the code to do the right thing (maybe for `0.27` cc @Wumpf).

Wumpf:
Would be almost better if we disable the toast entirely here? It's extra wrong when you select like _anything_

this:
image

Actually copied
`eStrip3D`

oxkitsune:
> Would be almost better if we disable the toast entirely here? It's extra wrong when you select like _anything_

I think that might be a good short term solution.

grtlr:
Yeah, let's get rid of the toast then this is fine by me.

PR #11497, opened 2025-10-10
Implement automatic token discovery

jprochazk:
* Closes RR-2536

### Refactor

The first commit is a refactor of the entire `workos` module in `re_auth`. It was hastily put together for the sake of a prototype, but we're going forward with it, so I tried to shape it up a bit. There are three entrypoints:

- `load_credentials`, which a synchronous function, and only loads the credentials from storage (disk on native, wasm currently unimplemented), and return `None` if they aren't available.
- `refresh_credentials`, which accepts an instance of possibly expired `Credentials`, and returns a new instance of `Credentials` with an expiration date some time in the future.
  - If the credentials are not expired, then they are returned as-is.
- `load_and_refresh_credentials`, which does both in sequence.

### Credentials provider

Cloud clients now use a credentials provider mechanism instead of storing a JWT string. A `CredentialsProvider` implements an async `get_token` function. It is async so that it can potentially send requests, such as for refreshing expired credentials.

This PR adds two implementations:

- `StaticCredentialsProvider`, which accepts a static JWT to return (`REDAP_TOKEN` env, `fallback_token`, etc.)
- `CliCredentialsProvider`, which reads credentials produced by `rerun auth login`, and refreshes them if needed.
  - It doesn't store credentials directly, instead credentials are stored as a singleton in a `static` - this helps prevent situations where more than one instance of credentials exists in the app at the same time.
  - It also performs request coalescing: If two threads attempt to get a token simultaneously, and credentials are expired, they will only be refreshed once.

The order in which tokens are used now are:
- Explicit per-server token
- Fallback token
- CLI credentials

All existing authentication methods continue to work.

### DIFF
--- a/crates/utils/re_auth/src/oauth.rs
+++ b/crates/utils/re_auth/src/oauth.rs
@@ -1,7 +1,5 @@
 use std::collections::HashMap;
 
-use jsonwebtoken::{Algorithm, DecodingKey, Validation, decode, decode_header, jwk::JwkSet};
-use saturating_cast::SaturatingCast as _;
 use serde::{Deserialize, Serialize};
 
 use crate::Jwt;
@@ -262,8 +260,10 @@ impl AccessToken {
     }
 
     pub fn remaining_duration_secs(&self) -> i64 {
+        use saturating_cast::SaturatingCast as _;
+
         // Time in seconds since unix epoch
-        let now = jsonwebtoken::get_current_timestamp() as i64;
+        let now: i64 = jsonwebtoken::get_current_timestamp().saturating_cast();
         self.expires_at - now
     }
 


crates/utils/re_auth/src/oauth.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    #[error("failed to deserialize credentials: {0}")]
+    MalformedToken(#[from] MalformedTokenError),
+}
+
+/// Refresh credentials if they are expired.
+pub async fn refresh_credentials(
+    credentials: Credentials,
+) -> Result<Credentials, CredentialsRefreshError> {
+    eprintln!("{credentials:?}");
```
teh-cmc: leftover?

crates/utils/re_auth/src/oauth/storage.rs
```diff
@@ -0,0 +1,84 @@
...
+        Ok(None)
+    }
+
+    pub fn store(credentials: &Credentials) -> Result<(), StoreError> {
+        let _ = credentials;
+        // This shouldn't actually be called anywhere, because no tokens are stored
+        // in local storage, which means nothing to refresh yet, either.
+        unreachable!("storing credentials in localStorage is not yet supported")
+    }
+}
```
teh-cmc, line 84: Didn't you say in the PR description that this was supported? I probably misunderstood.

crates/utils/re_auth/src/service/client.rs
```diff
@@ -1,40 +1,124 @@
-use re_log::error;
+// use super::{AUTHORIZATION_KEY, TOKEN_PREFIX};
```
teh-cmc: leftovers

crates/utils/re_auth/src/wasm_compat.rs
teh-cmc, line 1: :see_no_evil:

teh-cmc (approved):
Nice job. I gave it a very, very cursory look and nothing jump out at me.

We should deploy this on OG asap and use it ourselves, that's the only way to know for sure.

crates/utils/re_auth/src/oauth/storage.rs
jprochazk: It's more like it just compiles on web. There is nothing to actually create or pass the token into the web viewer. That's more of a job for a frontend login redirect thing. I updated the PR description

PR #11502, opened 2025-10-11
Update `register_partition` test `re_redap_tests`

abey79:
### Related

* part of https://linear.app/rerun/project/enable-ingestion-workflow-testing-with-oss-server-a73703fc9a94/
* closes https://linear.app/rerun/issue/RR-2322/mock-task-related-endpoints
* sibling https://github.com/rerun-io/dataplatform/pull/1847

### What

This PR moves and adapt a couple of test covering part of registration, partition table and dataset manifest schema and data, and tasks.

It also updates the OSS server as follows:
- The task-related endpoints are mocked.
- The in-memory store use compaction as per env variable (finally not needed for this PR, but I figured I'd leave it anyways)

Notes:
- The dataset manifest is currently disabled if the endpoint is unimplemented, because that's the case in OSS. They will soon be though, and the test will be re-enabled then.
- I made the `size_bytes` field of the manifest implementation-defined (aka it's not covered in snapshots). OSS uses heap size while Cloud uses, I believe, size on disk, and they don't match.

### DIFF
--- a/crates/store/re_server/tests/redap_tests.rs
+++ b/crates/store/re_server/tests/redap_tests.rs
@@ -1,17 +1,7 @@
-use re_chunk_store::ChunkStoreConfig;
 use re_server::{RerunCloudHandler, RerunCloudHandlerBuilder};
 
 #[expect(clippy::unused_async)] // needed by the macro
 async fn build() -> RerunCloudHandler {
-    // SAFETY: it's a test
-    #[expect(unsafe_code)]
-    unsafe {
-        // Mimic the behavior of cloud
-        std::env::set_var(ChunkStoreConfig::ENV_CHUNK_MAX_BYTES, "0");
-        std::env::set_var(ChunkStoreConfig::ENV_CHUNK_MAX_ROWS, "0");
-        std::env::set_var(ChunkStoreConfig::ENV_CHUNK_MAX_ROWS_IF_UNSORTED, "0");
-    }
-
     RerunCloudHandlerBuilder::new().build()
 }
 


crates/store/re_redap_tests/src/tests/register_partition.rs
```diff
@@ -0,0 +1,246 @@
...
+        rerun_cloud_service_server::RerunCloudService,
+    },
+    headers::RerunHeadersInjectorExt as _,
+};
+
+use crate::tests::common::register_with_dataset;
+use crate::{RecordBatchExt as _, create_simple_recording};
+
+// We just want to make sure that the dataset resolution logic and the proxy in general both work as
+// expected: registering and listing partitions using dataset IDs is a good way to do that.
```
teh-cmc: That comment was meant for readers of the Frontend code, in the context of the Frontend code. Doesn't make much sense here.
```suggestion
```

crates/store/re_redap_tests/src/tests/register_partition.rs
```diff
@@ -0,0 +1,246 @@
...
+pub async fn register_and_scan_simple_dataset(fe: impl RerunCloudService) {
+    // NOTE: We're re-using the same prefix multiple times, but it's fine because we're
+    // only interested in the dataset manifest in this test suite anyhow.
+
```
teh-cmc:
```suggestion
```

crates/store/re_redap_tests/src/utils/rerun.rs
```diff
@@ -38,6 +38,111 @@ fn next_row_id_generator(prefix: u64) -> impl FnMut() -> re_chunk::RowId {
 
 // ---
 
+/// Creates a simple, clean recording with sequential data and no overlaps or duplicates.
+///
+/// This recording is made such that it cannot be compacted, so that the effect of compaction is
+/// ruled out in snapshots.
```
teh-cmc, line 44: Seems pretty weird to me to rely on the shape of the test data to achieve that. The server should expose a constructor that makes it possible to configure compaction settings.

crates/store/re_server/tests/redap_tests.rs
```diff
@@ -1,7 +1,17 @@
+use re_chunk_store::ChunkStoreConfig;
 use re_server::{RerunCloudHandler, RerunCloudHandlerBuilder};
 
 #[expect(clippy::unused_async)] // needed by the macro
 async fn build() -> RerunCloudHandler {
+    // SAFETY: it's a test
```
teh-cmc: Hmmm... I'm confused now. Why go through the trouble of crafting that special data if compaction is disabled via the environment anyhow?

crates/store/re_redap_tests/src/utils/rerun.rs
abey79: Well OSS server does now (via env var), though I'm not exposing that in the rig. Also, I did not manage to get the same chunk count with OSS as with cloud using the degenerate recording.

On one hand, I agree this setup is really not great. On the other hand:
- I don't think redap should make any promises on what will happen to your chunks in general.
- I want to have _some_ snapshot of data.
Right now I'm not sure how to achieve this in a better way, but surely we will have to iterate on this.

crates/store/re_server/tests/redap_tests.rs
abey79: Ah, I actually forgot to remove that. It's done now. (As I said, I tried—and failed—to have a data match on the nasty recording that way.)

PR #11393, opened 2025-10-01
Dynamically assign port and catalog URL in test setup

Weijun-H:
### Related

Closes #11173

### What
dynamically assign port and catalog URL in test setup

PR #11512, opened 2025-10-13
Use dynamically assigned port for test server

oxkitsune:
### Related

* Fixes bug in #11393

### What

This passes the dynamically assigned port to the test server process.


rerun_py/tests/unit/test_datafusion_tables.py
```diff
@@ -140,6 +140,8 @@ def server_instance() -> Generator[ServerInstance, None, None]:
         f"second_schema.second_table={TABLE_FILEPATH}",
         "--table",
         f"alternate_catalog.third_schema.third_table={TABLE_FILEPATH}",
+        "--port",
+        str(port),
```
emilk: nit/style: I tend to find this easier to read, with one line per option

```suggestion
        f"--port={port}",
```

PR #11513, opened 2025-10-13
Flatten connection_client error hierarchy

andrea-reale:
### Related

- Part of RR-2395

### Why
When it comes to exposing errors to client (and specifically thinking at the Python SDK) we want:
- To have meaningful error messages that humans (primarily users and secondarily developers) can understand
- To have actionable types that computers can use to react to errors

Our previous taxonomy of errors in the client was very deeply nested and made it hard to classify the errors
in a way that was machine friendly, i.e., it was hiding very deep the underlying source of the error, and mainly
highlighting the operation where the error happened.

### What

This is an attempt to flatten that taxonomy. All errors generated by the connection client will be of type `ApiError` and 
will have a `ApiErrorKind` of one of few types. The assumption is that those  types are generic across all operations,
and are machine actionable.

`ApiError` is expected to include a human understandable error message string as well as wrapping the underlying `source`
of the error, for further inspection.

### Next
In my ideal mental world, `ExternalError` in `rerun_py` should follow a similar model (or even disappear completely in favor of a more global `ApiError`-equivalent). This will allow to have all the possible error in just a bunch of useful classes instead of this odd mixed taxonomy.

I did not do it in this PR because I would have had too touch way more things I am willing to at this stage.

### DIFF
--- a/crates/store/re_redap_client/src/connection_registry.rs
+++ b/crates/store/re_redap_client/src/connection_registry.rs
@@ -86,11 +86,22 @@ impl From<ConnectionError> for ClientConnectionError {
 }
 
 impl ClientConnectionError {
+    #[inline]
     pub fn is_token_error(&self) -> bool {
         matches!(
             self,
             Self::UnauthenticatedMissingToken(_) | Self::UnauthenticatedBadToken(_)
         )
+    }
+
+    #[inline]
+    pub fn is_missing_token(&self) -> bool {
+        matches!(self, Self::UnauthenticatedMissingToken(_))
+    }
+
+    #[inline]
+    pub fn is_wrong_token(&self) -> bool {
+        matches!(self, Self::UnauthenticatedBadToken(_))
     }
 }
 
--- a/crates/store/re_redap_client/src/lib.rs
+++ b/crates/store/re_redap_client/src/lib.rs
@@ -190,6 +190,14 @@ impl ApiError {
             source: Some(Box::new(err)),
         }
     }
+
+    /// Helper method to downcast the source error to a `ClientConnectionError` if possible.
+    #[inline]
+    pub fn as_client_connection_error(&self) -> Option<&ClientConnectionError> {
+        self.source
+            .as_deref()?
+            .downcast_ref::<ClientConnectionError>()
+    }
 }
 
 impl std::fmt::Display for ApiError {
--- a/crates/viewer/re_redap_browser/src/entries.rs
+++ b/crates/viewer/re_redap_browser/src/entries.rs
@@ -16,40 +16,11 @@ use re_protos::cloud::v1alpha1::ext::{EntryDetails, TableEntry};
 use re_protos::cloud::v1alpha1::{EntryFilter, EntryKind, ext::DatasetEntry};
 use re_protos::external::prost;
 use re_protos::external::prost::Name as _;
-use re_redap_client::{
-    ApiError, ClientConnectionError, ConnectionClient, ConnectionRegistryHandle,
-};
+use re_redap_client::{ApiError, ConnectionClient, ConnectionRegistryHandle};
 use re_ui::{Icon, icons};
 use re_viewer_context::AsyncRuntimeHandle;
 
 pub type EntryResult<T> = Result<T, ApiError>;
-
-pub trait AuthApiErrorExt {
-    fn client_connection_error(&self) -> Option<&ClientConnectionError>;
-    fn is_missing_token(&self) -> bool;
-    fn is_wrong_token(&self) -> bool;
-}
-
-impl AuthApiErrorExt for ApiError {
-    fn client_connection_error(&self) -> Option<&ClientConnectionError> {
-        use std::error::Error as _;
-        self.source()?.downcast_ref::<ClientConnectionError>()
-    }
-
-    fn is_missing_token(&self) -> bool {
-        matches!(
-            self.client_connection_error(),
-            Some(ClientConnectionError::UnauthenticatedMissingToken(_))
-        )
-    }
-
-    fn is_wrong_token(&self) -> bool {
-        matches!(
-            self.client_connection_error(),
-            Some(ClientConnectionError::UnauthenticatedBadToken(_))
-        )
-    }
-}
 
 pub struct Dataset {
     pub dataset_entry: DatasetEntry,
--- a/crates/viewer/re_redap_browser/src/servers.rs
+++ b/crates/viewer/re_redap_browser/src/servers.rs
@@ -17,7 +17,7 @@ use re_ui::{UiExt as _, icons};
 use re_viewer_context::{AsyncRuntimeHandle, GlobalContext, ViewerContext};
 
 use crate::context::Context;
-use crate::entries::{AuthApiErrorExt as _, Dataset, Entries, Entry, Table};
+use crate::entries::{Dataset, Entries, Entry, Table};
 use crate::server_modal::{ServerModal, ServerModalMode};
 
 pub struct Server {
@@ -118,13 +118,15 @@ impl Server {
     fn server_ui(&self, viewer_ctx: &ViewerContext<'_>, ctx: &Context<'_>, ui: &mut egui::Ui) {
         if let Poll::Ready(Err(err)) = self.entries.state() {
             self.title_ui(self.origin.host.to_string(), ctx, ui, |ui| {
-                if err.is_missing_token() || err.is_wrong_token() {
-                    let message = if err.is_missing_token() {
+                if let Some(conn_err) = err.as_client_connection_error()
+                    && conn_err.is_token_error()
+                {
+                    let message = if conn_err.is_missing_token() {
                         "This server requires a token to access its data."
                     } else {
                         "The provided token is invalid for this server."
                     };
-                    let edit_message = if err.is_missing_token() {
+                    let edit_message = if conn_err.is_missing_token() {
                         "Add a token"
                     } else {
                         "Edit token"


crates/store/re_redap_client/src/lib.rs
```diff
@@ -76,100 +76,141 @@ impl std::error::Error for TonicStatusError {
...
-    PartitionError(#[from] StreamPartitionError),
-
-    #[error(transparent)]
-    TasksError(#[from] StreamTasksError),
+impl From<&ClientConnectionError> for ApiErrorKind {
+    fn from(err: &ClientConnectionError) -> Self {
+        match err {
+            ClientConnectionError::UnencryptedServer => Self::InvalidArguments,
+            ClientConnectionError::UnauthenticatedMissingToken(_)
+            | ClientConnectionError::UnauthenticatedBadToken(_) => Self::Unauthenticated,
```
abey79, line 119: I think this is a slightly unfortunate loss of information for the user. I had specifically split these cases so that the user could know whether their token is bad or missing (which requires different action). Even if we're moving regular use case to a more automated token workflow, it's less of an issue, but might still be in the context of long-lived token for machine tasks.

One way to deal with that is specify the nuance in the `message` instead of exploding the number of `ApiErrorKind`s. This gives me pause about this `impl From`. I think it would be a better pattern to implement `From` for `ApiError` instead `ApiErrorKind`, so that we have the full opportunity to provide nuance to the user.

edit: I now see that the missing/bad distinction can still be done based on the source error. Still, having it spelled in the message is greater chance to be seen by the user (see the most recent implementation of our in-viewer error notification, where the details are collapsed).

crates/store/re_redap_client/src/lib.rs
```diff
@@ -76,100 +76,141 @@ impl std::error::Error for TonicStatusError {
...
+            .as_deref()
+            .map(|e| e as &(dyn std::error::Error + 'static))
+    }
 }
 
 const _: () = assert!(
-    std::mem::size_of::<StreamError>() <= 80,
+    std::mem::size_of::<ApiError>() <= 80,
     "Error type is too large. Try to reduce its size by boxing some of its variants.",
 );
```
abey79: This is possibly slightly redundant with the `result_large_err` clippy lint (but fine to still have it).

crates/viewer/re_redap_browser/src/entries.rs
```diff
@@ -4,97 +4,46 @@ use std::task::Poll;
...
-);
-
-impl From<DataFusionError> for EntryError {
-    fn from(err: DataFusionError) -> Self {
-        Self::DataFusionError(Box::new(err))
-    }
+pub trait AuthApiErrorExt {
+    fn client_connection_error(&self) -> Option<&ClientConnectionError>;
+    fn is_missing_token(&self) -> bool;
+    fn is_wrong_token(&self) -> bool;
```
abey79: what is the motivation to have this implemented in an extension trait? These methods are of a general usefulness and should be in `re_redap_client` imo. In particular, though we currently don't, we could possibly use them in the python SDK.

As a reminder,  `re_redap_client` is slated to become the "official redap API binding crate" crate for redap. So third party implementation could also benefit from that, eventually.

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -5,6 +5,7 @@ use arrow::datatypes::Schema as ArrowSchema;
 use arrow::pyarrow::PyArrowType;
 use pyo3::exceptions::PyValueError;
 use pyo3::{PyErr, PyResult, Python, create_exception, exceptions::PyConnectionError};
+use re_protos::{invalid_schema, missing_field};
```
abey79: nit: we have this untold convention to sort `use` statement by std -> external -> re_* -> crate

crates/store/re_protos/src/lib.rs
```diff
@@ -97,6 +97,19 @@ pub enum TypeConversionError {
         reason: String,
     },
 
+    #[error("missing required dataframe column {column_name} in {package_name}.{type_name}")]
```
emilk:
```suggestion
    #[error("missing required dataframe column {column_name:?} in '{package_name}.{type_name}'")]
```

`:?` will add quotes around the string, as well as escape special characters.

Adding quotes can improve readability in many cases (consider e.g. `column_name` = "name"`).

More importantly, if there is an input error somewhere and there is a trailing space or a trailing newline, using `:?` is absolutely CRITICAL in us or the user being able to debug it.

crates/store/re_protos/src/lib.rs
```diff
@@ -97,6 +97,19 @@ pub enum TypeConversionError {
...
+    MissingColumn {
+        package_name: &'static str,
+        type_name: &'static str,
+        column_name: &'static str,
+    },
+
+    #[error("invalid dataframe schema in {package_name}.{type_name}")]
```
emilk:
```suggestion
    #[error("invalid dataframe schema in '{package_name}.{type_name}'")]
```

crates/store/re_redap_client/src/connection_registry.rs
```diff
@@ -70,7 +70,7 @@ pub enum ClientConnectionError {
     #[error("the server rejected the provided authentication token\nDetails:{0}")]
     UnauthenticatedBadToken(TonicStatusError),
 
-    #[error("failed to obtain server version\nDetails:{0}")]
+    #[error("failed to validate credentials\nDetails:{0}")]
```
emilk, line 73: wow, good fix

crates/store/re_redap_client/src/lib.rs
```diff
@@ -76,100 +76,141 @@ impl std::error::Error for TonicStatusError {
...
+        Self {
+            message: message.into(),
+            kind: ApiErrorKind::from(err.code()),
+            source: Some(Box::new(err)),
+        }
+    }
 
-    #[error(transparent)]
-    ChunkError(#[from] re_chunk::ChunkError),
+    pub fn serde(
```
abey79: I'm not a fan of using this given how loaded it is ("serde = a well known crate"), and also it doesn't match the actual enum arm name.

rerun_py/src/catalog/errors.rs
```diff
@@ -152,9 +143,19 @@ impl From<ExternalError> for PyErr {
...
+            ExternalError::ApiError(err) => match err.kind {
+                ApiErrorKind::Connection => PyConnectionError::new_err(err.to_string()),
+                ApiErrorKind::Unauthenticated | ApiErrorKind::PermissionDenied => {
+                    PyPermissionError::new_err(err.to_string())
+                }
+                ApiErrorKind::Serialization | ApiErrorKind::InvalidArguments => {
+                    PyValueError::new_err(err.to_string())
+                }
+                ApiErrorKind::NotFound => PyFileNotFoundError::new_err(err.to_string()),
+                ApiErrorKind::AlreadyExists => PyFileExistsError::new_err(err.to_string()),
```
abey79, line 155: I don't think this is a correct mapping, as the not found/already exists will apply to entries (dataset/table, etc.) and not file. I think `PyValueError` applies here, for the lack of a more appropriate built-in error.

https://docs.python.org/3/library/exceptions.html#ValueError: 
> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

crates/store/re_redap_client/src/lib.rs
andrea-reale: the `Display` implementation of `ApiError` already prints the `message` of the `source` error, so I think it already satisfies the requirement you ar etalking about.

The current implementation would print something like:

```
Unauthenticated: the server rejected the provided authentication token\nDetails..
```

vs

```
Unauthenticated: the server requires an authentication token but none was provided\nDetails
```

crates/store/re_redap_client/src/lib.rs
andrea-reale: Sure, changing it to avoid any confusion.

(I am used to see `serde` as a generic expression for "serialization-deserialization" in multiple non-rust contexts for a very long time, so I do not think about it that way).

crates/store/re_redap_client/src/lib.rs
andrea-reale: removed it, less code is good

crates/viewer/re_redap_browser/src/entries.rs
andrea-reale: Looking twice, I agree with you that the extension trait is a bad solution.

What I did is move those helper methods to `impl ClientConnectionError`, and add a helper to `ApiError` to help downcast when necessary. See : [`d6105db` (#11513)](https://github.com/rerun-io/rerun/pull/11513/commits/d6105dbdd5a3e891c4b3984d09130a97c43c64b3)

rerun_py/src/catalog/errors.rs
andrea-reale: But I would really  like for the client to be able to programmatically identify a "generic error" (which `PyValueError` ends up being) vs something they can understand and act upon.

For example you could do (in pseudo-python):
```python
  try:
    dataset = catalog.get_dataset_entry()
  except FileNotFoundError:
    dataset = catalog.create_dataset_entry()
  except e:
    raise e
```

I think that would not be possible with a generic "value error" unless someone starts to do string matching.

In practice, I do not think any customer would shout and say "but it's not a real file!".

More philosophically, this matches the unix philosofy where every "entity" is a file, and when an entity is not found you get an `ENOENT`

> Raised when a file or directory is requested but doesn’t exist. Corresponds to errno [ENOENT](https://docs.python.org/3/library/errno.html#errno.ENOENT).

[CAVEAT: The server does not return the right status code anyways nowadays, but it's my personal battle that it will in the future]

abey79: I agree with your intention. Then the Pythonic course of action is to create our own error types and raise that.

Ideally, every single method that may raise should document what it raises under which circumstances (otherwise the user wouldn't even know how to write their `except` blocks), but we're extremely sloppy with that. An advantage of a custom error types is that it helps with discoverability, and so buys us some time on the docs front.

crates/store/re_redap_client/src/lib.rs
abey79: Alright. Did you check what happens in the viewer's UI? At the end of the day, that's what counts (since programmatic users will be more prone to read the message(s) in details).

andrea-reale: The server swallows the exact message and provides a nicer dialog:
image

Programmatically, instead, you get this:
image

rerun_py/src/catalog/errors.rs
andrea-reale: As discussed on zoom, I'll make a fast-follow PR today creating a custom error type `EntryNotFoundError`/`EntryExistsError`

PR #11515, opened 2025-10-13
Replaces `check_context_menu_multi_selection` test with kittest

aedm:
### Related

Part of #8948 

### What

- Adds a test for context menus when selecting multiple items in the blueprint panel
- Removes the manual `check_context_menu_multi_selection.py` test


crates/viewer/re_viewport/src/viewport_ui.rs
```diff
@@ -155,43 +159,60 @@ impl ViewportUi {
...
+            // // We paint the stroke on the inside so the panel-resize lines don't cover the highlight rectangle.
+            // painter.rect_stroke(rect, 0.0, stroke, egui::StrokeKind::Inside);
+
+            // if should_display_drop_destination_frame {
+            //     painter.rect_filled(
+            //         rect.shrink(stroke.width),
+            //         0.0,
+            //         stroke.color.gamma_multiply(0.1),
+            //     );
+            // }
```
lucasmerlin: Whats with this?

tests/rust/re_integration_test/src/kittest_harness_ext.rs
```diff
@@ -289,6 +302,17 @@ impl HarnessExt for egui_kittest::Harness<'_, re_viewer::App> {
...
     }
 
+    // Some snapshots have significant differences, eg. 2D views when its entities are selected
+    fn snapshot_app_high_threshold(&mut self, snapshot_name: &str) {
+        self.run_ok();
+        self.snapshot_options(
+            snapshot_name,
+            &SnapshotOptions::new()
+                .threshold(2.0)
+                .failed_pixel_count_threshold(1000),
```
lucasmerlin: Not a fan of hardcoding `failed_pixel_count_threshold` here, especially since we can see how many pixels didn't match in the test output. I think we should try to get this as close as possible, to avoid cases where we miss unrelated changes. It's a really big footgun. 
Also I'd rather go higher with the normal, maybe as high as 10? If that would mean we don't need the failed_pixel_threshold at all.
With 1000 pixels, you'd definitely miss if e.g. a icon button disappears or something like that.

I'd even go as far as saying this should probably be an exact match. So that a test fails if for some reason the test needs less failed pixels threshold in the future and you need to change it.

crates/viewer/re_viewport/src/viewport_ui.rs
aedm: This is me pushing the "ready for review" button way too early...

tests/rust/re_integration_test/src/kittest_harness_ext.rs
emilk: Also note that I'm working on new thresholds in
* https://github.com/rerun-io/rerun/pull/11542

aedm: Sorry, this was experimental while I tried to fix the threshold issue. It's gone now.

emilk:
Please hold off on merging until I get these in:
* https://github.com/rerun-io/rerun/pull/11551
* https://github.com/rerun-io/rerun/pull/11542

emilk (approved):
code LGTM

PR #11517, opened 2025-10-13
Fix flaky `air_traffic_data` example

oxkitsune:
### What

This fixes the flaky `air_traffic_data` example by skipping aircrafts without any data.


examples/python/air_traffic_data/air_traffic_data.py
```diff
@@ -367,6 +367,9 @@ def log_ground_status(self, df: polars.DataFrame, icao_id: str) -> None:
         entity_path = f"aircraft/{icao_id}"
         df = df["timestamp", "ground_status"].drop_nulls()
 
+        if df.height == 0:
+            return
```
emilk, line 371: This is lacking a comment explaining _why_

PR #11522, opened 2025-10-14
Background of graph view is now blueprint configurable

IsseW:
### Related

- Closes RR-2350

### What

Adds blueprint controlled background color for graph views.

image

### DIFF
--- a/crates/store/re_types/src/blueprint/views/graph_view.rs
+++ b/crates/store/re_types/src/blueprint/views/graph_view.rs
@@ -26,7 +26,7 @@ use ::re_types_core::{DeserializationError, DeserializationResult};
 /// ⚠️ **This type is _unstable_ and may change significantly in a way that the data won't be backwards compatible.**
 #[derive(Clone, Debug)]
 pub struct GraphView {
-    /// Configures the background of the plot.
+    /// Configures the background of the graph.
     pub background: crate::blueprint::archetypes::GraphBackground,
 
     /// Everything within these bounds is guaranteed to be visible.
--- a/docs/content/reference/types/views/graph_view.md
+++ b/docs/content/reference/types/views/graph_view.md
@@ -9,7 +9,7 @@ A graph view to display time-variying, directed or undirected graph visualizatio
 ## Properties
 
 ### `background`
-Configures the background of the plot.
+Configures the background of the graph.
 ### `visual_bounds`
 Everything within these bounds is guaranteed to be visible.
 
--- a/rerun_py/rerun_sdk/rerun/blueprint/views/graph_view.py
+++ b/rerun_py/rerun_sdk/rerun/blueprint/views/graph_view.py
@@ -128,7 +128,7 @@ class GraphView(View):
             This will be addressed in <https://github.com/rerun-io/rerun/issues/6673>.
 
         background:
-            Configures the background of the plot.
+            Configures the background of the graph.
         visual_bounds:
             Everything within these bounds is guaranteed to be visible.
 


Wumpf:
screenshot or didn't happen 😄

crates/store/re_types/definitions/rerun/blueprint/archetypes/background.fbs
```diff
@@ -28,3 +28,11 @@ table PlotBackground (
...
   show_grid: rerun.blueprint.components.Enabled ("attr.rerun.component_optional", nullable, order: 2000);
 }
+
+/// Configuration of a background in a graph view.
+table GraphBackground (
+  "attr.rerun.scope": "blueprint"
+) {
+  /// Color used for the background.
+  color: rerun.components.Color ("attr.rerun.component_optional", nullable, order: 1000);
+}
```
Wumpf, line 38: was surprised at first that we need a new archetype for this given that time series & bar chart has it (which I didn't know because it's like less than an hour old :D). But yeah we don't have a grid there I figure 🤔

crates/store/re_types/definitions/rerun/blueprint/views/graph.fbs
```diff
@@ -6,6 +6,10 @@ namespace rerun.blueprint.views;
 table GraphView (
     "attr.rerun.view_identifier": "Graph"
 ) {
+
+    /// Configures the background of the plot.
```
Wumpf:
```suggestion
    /// Configures the background of the graph.
```
.. I guess?

crates/viewer/re_view_graph/src/properties.rs
```diff
@@ -70,4 +70,15 @@ impl TypedComponentFallbackProvider<Position2D> for GraphView {
...
 
-re_viewer_context::impl_component_fallback_provider!(GraphView => [VisualBounds2D, Enabled, ForceDistance, ForceStrength, ForceIterations, Position2D]);
+impl TypedComponentFallbackProvider<Color> for GraphView {
+    fn fallback_for(&self, ctx: &re_viewer_context::QueryContext<'_>) -> Color {
+        // Color is a fairly common component, make sure this is the right context.
+        if ctx.archetype_name == Some(GraphBackground::name()) {
+            ctx.viewer_ctx().tokens().viewport_background.into()
+        } else {
+            Color::default()
+        }
```
Wumpf, line 80: as evident by the screenshot comparison tests, this changes the background color slightly. Is that intentional?

IsseW:
> screenshot or didn't happen 😄

added! 😄

crates/viewer/re_view_graph/src/properties.rs
IsseW: I feel like this makes sense since it uses the `viewport_background` token as a fallback now, which is what I'd expect.

crates/store/re_types/definitions/rerun/blueprint/archetypes/background.fbs
IsseW: Yeah no grid, did almost want to add it just cause and default to off, but very out of scope 😅

crates/viewer/re_view_graph/src/properties.rs
Wumpf: hmm probably fine but worth pinging @gavrelina about it

Wumpf:
Something I've forgot before: can we add a change to this config to the graph view config snippet so that this ability is documented and once we automate & screenshot tests those this is covered as well?

IsseW:
> Something I've forgot before: can we add a change to this config to the graph view config snippet so that this ability is documented and once we automate & screenshot tests those this is covered as well?

Yeah good point, maybe I could add that for both this and for bar chart/time series as well in a follow up?

Created issue here RR-2668

crates/viewer/re_view_graph/src/properties.rs
gavrelina: No objections! eventually I want to make those a bit more aligned between views (now if you notice then there is a slight discrepancy in color of the "tab" and the background and if text document view is in the viewport - it also is slightly different. but those are actually issues to solve in themselves, so unless there is a "separation" to be able to change these flexible (which @IsseW creates here) Im happy!

Issue #11516, opened 2025-10-13
RRD decoder panics on allocation: `capacity overflow`


nathan-b-flya:
**Describe the bug**

When sending too much data, Rerun crashes with the error:
```
thread 'tokio-runtime-worker' panicked at 'capacity overflow'
alloc/src/raw_vec/mod.rs:558
stack backtrace:
   6: core::panicking::panic_fmt
   7: alloc::raw_vec::capacity_overflow
   8: alloc::raw_vec::handle_error
   9: alloc::raw_vec::RawVecInner<A>::reserve::do_reserve_and_handle
  10: re_log_encoding::codec::arrow::decode_arrow
  11: re_log_encoding::protobuf_conversions::arrow_msg_from_proto
```

**To Reproduce**

Here is the C++ code to reproduce the issue easily `rerun_test.cpp`:
```C++
#include <cstdlib>
#include <iostream>
#include <numeric>
#include <string>
#include <vector>

#include <rerun.hpp>

struct AppOptions
{
    std::size_t nb_scans = 1;
    bool api_columns = true;
};

/// Handmade argument parsing for portability and avoid 3rd party library
/// when submitting Github issues
static AppOptions parseOptions(int argc, char** argv)
{
    AppOptions options;
    for (int i = 1; i < argc; ++i)
    {
        std::string arg = argv[i];
        if (arg == "--nb_scans")
        {
            if (i + 1 < argc)
            {
                try
                {
                    options.nb_scans = std::stoull(argv[++i]);
                }
                catch (const std::exception& e)
                {
                    std::cerr << "Error: Invalid argument for --nb_scans. Please provide "
                                 "a non-negative integer."
                              << std::endl;
                    std::exit(EXIT_FAILURE);
                }
            }
            else
            {
                std::cerr << "Error: The --nb_scans option requires a value."
                          << std::endl;
                std::exit(EXIT_FAILURE);
            }
        }
        else if (arg == "--row_api")
        {
            options.api_columns = false;
        }
        else
        {
            std::cerr << "Warning: Unknown option '" << arg << "' will be ignored."
                      << std::endl;
        }
    }
    return options;
}

int main(int argc, char** argv)
{
    const AppOptions args = parseOptions(argc, argv);

    const auto rec = rerun::RecordingStream("github_issue");
    rec.spawn().exit_on_failure();

    const std::size_t nb_scans = args.nb_scans;
    const std::size_t nb_points_per_scan = 1024 * 128;
[...]

PR #11525, opened 2025-10-14
Fix data corruption during compression of large chunks

teh-cmc:
There are 3 problems involved:
1. `ArrowMsg::uncompressed_size` is defined as an `int32`. That issue speaks for itself.
2. The encoding path tracks payload sizes using `usize`, and then silently cast them to `int32` when encoding the Protobuf data, leading to gibberish and crashes during decoding.
3. `lz4_flex` has similar casting bugs: it can encode 64bit values but cannot decode them. Once again the `as` keyword is involved.

This PR fixes 1 and 2. All the changes are backward and forward compatible because:
* In protobuf, `uint64` is a strict superset of `int32`: both are unsigned varints on the wire.
* There exist no valid RRD files out there with negative values for `uncompressed_size`.

We can already log much bigger chunks than before, but not too big, otherwise `lz4` itself will during decoding:
```
thread 'decompress' panicked at 'attempt to add with overflow'
lz4_flex-0.11.5/src/block/decompress_safe.rs:39
stack backtrace:
   6: core::panicking::panic_fmt
             at core/src/panicking.rs:75:14
   7: core::panicking::panic_const::panic_const_add_overflow
             at core/src/panicking.rs:175:17
   8: std::sys::backtrace::__rust_begin_short_backtrace
   9: core::ops::function::FnOnce::call_once{{vtable.shim}}
  10: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
             at alloc/src/boxed.rs:1966:9
      <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
             at alloc/src/boxed.rs:1966:9
      std::sys::pal::unix::thread::Thread::new::thread_start
             at std/src/sys/pal/unix/thread.rs:97:17
  11: <unknown>
  12: <unknown>
```

I'll have a follow-up PR for the `lz4_flex` issue.

---

* Half of fixing https://github.com/rerun-io/rerun/issues/11516
  * The other half requires patching `lz4_flex` (on it)

---

Checks:
* [x] `rerun` with `--nb_scans 2000`
* [x] `rerun rrd stats` with `--nb_scans 2000`
* [x] :no_entry: `rerun` with `--nb_scans 3000` -> bug in lz4
* [x] :no_entry: `rerun rrd stats` with `--nb_scans 3000` -> bug in lz4
* [x] All dataplatform test suites pass
* [x] Can load old RRDs no problem

### DIFF
--- a/tests/rust/log_benchmark/src/main.rs
+++ b/tests/rust/log_benchmark/src/main.rs
@@ -82,10 +82,13 @@ struct Args {
     #[clap(long, default_value = "false")]
     profile: bool,
 
-    /// If true, connect to a running Rerun viewer
-    /// instead of writing to a memory buffer.
+    /// If true, connect to a running Rerun viewer instead of writing to a memory buffer.
     #[clap(long, default_value = "false")]
     connect: bool,
+
+    /// If true, perform an encode/decode roundtrip on the logged data.
+    #[clap(long, default_value = "false")]
+    check: bool,
 }
 
 fn main() -> anyhow::Result<()> {
@@ -98,6 +101,7 @@ fn main() -> anyhow::Result<()> {
         benchmark,
         profile,
         connect,
+        check,
     } = Args::parse();
 
     // Start profiler first thing:
@@ -130,7 +134,7 @@ fn main() -> anyhow::Result<()> {
 
     // Being able to log fast isn't particularly useful if the data happens to be corrupt at the
     // other end, so make sure we can encode/decode everything that was logged.
-    if let Some(storage) = storage {
+    if check && let Some(storage) = storage {
         use rerun::external::re_log_encoding;
 
         let msgs: anyhow::Result<Vec<_>> = storage


crates/store/re_log_encoding/src/protobuf_conversions.rs
```diff
@@ -210,7 +210,7 @@ pub fn arrow_msg_to_proto(
             }
             crate::codec::Compression::LZ4 => re_protos::log_msg::v1alpha1::Compression::Lz4 as i32,
         },
-        uncompressed_size: payload.uncompressed_size as i32,
```
teh-cmc, line 213: The original bug: `Payload::uncompressed_size` was a `usize` that could definitely grow beyond `i32::MAX`, so that silenced cast leads to disaster.

tests/rust/log_benchmark/src/main.rs
```diff
@@ -120,9 +123,40 @@ fn main() -> anyhow::Result<()> {
...
+            .map(|msg| {
+                Ok(re_log_encoding::protobuf_conversions::log_msg_from_proto(
+                    &mut app_id_injector,
+                    msg,
+                )?)
+            })
+            .collect();
+
+        let _ = msgs?;
+    }
```
teh-cmc, line 163: Ha, just realized this whole thing is meant to be run with an external timer: need to put this behind an extra flag.

teh-cmc:
@rerun-bot full-check

emilk:
Ouch - maybe we should turn on https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_truncation and https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_wrap

EDIT:
* https://github.com/rerun-io/rerun/pull/11534

crates/store/re_protos/proto/rerun/v1alpha1/log_msg.proto
```diff
@@ -58,7 +58,7 @@ message ArrowMsg {
   // Compression algorithm used.
   Compression compression = 2;
 
-  int32 uncompressed_size = 3;
+  uint64 uncompressed_size = 3;
```
emilk, line 61: is this a breaking change? If so, can we make it a non-breaking change by adding a new field instead?

tests/rust/log_benchmark/src/main.rs
```diff
@@ -120,9 +127,40 @@ fn main() -> anyhow::Result<()> {
...
         Benchmark::Boxes3D => boxes3d_batch::run(&rec)?,
         Benchmark::Image(cmd) => cmd.run(&rec)?,
+        Benchmark::VeryLargeChunk => very_large_chunk::run(&rec)?,
     }
 
     rec.flush_blocking()?;
 
+    // Being able to log fast isn't particularly useful if the data happens to be corrupt at the
+    // other end, so make sure we can encode/decode everything that was logged.
+    if check && let Some(storage) = storage {
```
emilk, line 137: Is this run on CI?

crates/store/re_protos/proto/rerun/v1alpha1/log_msg.proto
teh-cmc: No, see PR description

tests/rust/log_benchmark/src/main.rs
teh-cmc: No, I don't believe any of this stuff runs on CI. Maybe I'll have a look in a follow-up.

teh-cmc:
merge & pray

PR #11529, opened 2025-10-14
custom NotFound/AlreadyExists python errors

andrea-reale:
### Related
Follow-up to #11513

### What
Use custom python error types instead of of built-in `FileNotFoundError`/`FileExistsError`.

### Manual testing
```
In [1]: import rerun
DEV ENVIRONMENT DETECTED! Re-importing rerun from: /var/home/andrea/workspace/rerun.io/rerun/rerun_py/rerun_sdk

In [2]: exception = rerun.catalog.AlreadyExistsError("entry exists")

In [3]: raise exception
---------------------------------------------------------------------------
AlreadyExistsError                          Traceback (most recent call last)
Cell In[3], line 1
----> 1 raise exception

AlreadyExistsError: entry exists
```

### DIFF
--- a/rerun_py/rerun_bindings/rerun_bindings.pyi
+++ b/rerun_py/rerun_bindings/rerun_bindings.pyi
@@ -1839,4 +1839,3 @@ class NotFoundError(Exception):
 
 class AlreadyExistsError(Exception):
     """Raised when trying to create a resource that already exists."""
-


rerun_py/src/catalog/mod.rs
```diff
@@ -53,6 +54,10 @@ pub(crate) fn register(_py: Python<'_>, m: &Bound<'_, PyModule>) -> PyResult<()>
     m.add_class::<PyVectorDistanceMetric>()?;
     m.add_class::<PyRerunHtmlTable>()?;
 
+    // register exceptions generated with the [`pyo3::create_exception!`] macro
+    m.add("EntryNotFoundError", _py.get_type::<EntryNotFoundError>())?;
```
andrea-reale: note: I tried to get the rust type called `PyEntryFooError` and the corresponding python type `EntryFooError`.

It kind-of-worked when it came to writing code that used the type. However, when raising an exception and getting the stack trace printed by python, I would still get `PyEntryFooError` which is confusing from the point of view of the UX.

PR #11514, opened 2025-10-13
Replace `#[allow(…)]` atributes with `#[expect`

emilk:
### Related
* See https://rust-lang.github.io/rust-clippy/master/index.html#allow_attributes
* Closes https://github.com/rerun-io/rerun/issues/7485

### What
A little improvement to our code quality

PR #11531, opened 2025-10-14
Drop unused `#[expect(..)]`

oxkitsune:
Drop unused `#[expect(..)]` introduced in #11514


oxkitsune:
@rerun-bot full-check

PR #11532, opened 2025-10-14
Add support for layers in OSS server

abey79:
### Related

* part of https://linear.app/rerun/issue/RR-2482/add-layer-support-to-oss-server
* part of https://linear.app/rerun/issue/RR-2650/clarify-the-api-contracts-around-querydataset-and-fetchchunks
* sibling https://github.com/rerun-io/dataplatform/pull/1861

### What

This adds support for layers in the OSS server. To do so, this PR makes several refinement to the redap specs, notably around `QueryDataset` and `FetchChunks` (see RR-2650, and in particular [this comment](https://linear.app/rerun/issue/RR-2650/clarify-the-api-contracts-around-querydataset-and-fetchchunks#comment-d12cc6d6)). This is reflected by a bunch of new and updated tests in `re_redap_tests`.

The handlers for both `QueryDataset` and `FetchChunks` have been entirely rewritten. Notably, the per-timeline-per-entity stats are no longer emitted.

### DIFF
--- a/crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
+++ b/crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
@@ -28,6 +28,15 @@ use crate::common::v1alpha1::{
 };
 use crate::{TypeConversionError, invalid_field, missing_field};
 
+/// Helper to simplify writing `field_XXX() -> FieldRef` methods.
+macro_rules! lazy_field_ref {
+    ($fld:expr) => {{
+        static FIELD: std::sync::OnceLock<FieldRef> = std::sync::OnceLock::new();
+        let field = FIELD.get_or_init(|| Arc::new($fld));
+        Arc::clone(field)
+    }};
+}
+
 // --- RegisterWithDatasetRequest ---
 
 #[derive(Debug)]
@@ -177,47 +186,59 @@ impl QueryDatasetResponse {
     pub const FIELD_CHUNK_ENTITY_PATH: &str = "chunk_entity_path";
     pub const FIELD_CHUNK_IS_STATIC: &str = "chunk_is_static";
 
-    pub fn field_chunk_id() -> Field {
-        Field::new(Self::FIELD_CHUNK_ID, DataType::FixedSizeBinary(16), false).with_metadata(
-            [("rerun:kind".to_owned(), "control".to_owned())]
-                .into_iter()
-                .collect(),
+    pub fn field_chunk_id() -> FieldRef {
+        lazy_field_ref!(
+            Field::new(Self::FIELD_CHUNK_ID, DataType::FixedSizeBinary(16), false).with_metadata(
+                [("rerun:kind".to_owned(), "control".to_owned())]
+                    .into_iter()
+                    .collect(),
+            )
         )
     }
 
-    pub fn field_chunk_partition_id() -> Field {
-        Field::new(Self::FIELD_CHUNK_PARTITION_ID, DataType::Utf8, false).with_metadata(
-            [("rerun:kind".to_owned(), "control".to_owned())]
-                .into_iter()
-                .collect(),
+    pub fn field_chunk_partition_id() -> FieldRef {
+        lazy_field_ref!(
+            Field::new(Self::FIELD_CHUNK_PARTITION_ID, DataType::Utf8, false).with_metadata(
+                [("rerun:kind".to_owned(), "control".to_owned())]
+                    .into_iter()
+                    .collect(),
+            )
         )
     }
 
-    pub fn field_chunk_layer_name() -> Field {
-        Field::new(Self::FIELD_CHUNK_LAYER_NAME, DataType::Utf8, false)
+    pub fn field_chunk_layer_name() -> FieldRef {
+        lazy_field_ref!(Field::new(
+            Self::FIELD_CHUNK_LAYER_NAME,
+            DataType::Utf8,
+            false
+        ))
     }
 
-    pub fn field_chunk_key() -> Field {
-        Field::new(Self::FIELD_CHUNK_KEY, DataType::Binary, false)
+    pub fn field_chunk_key() -> FieldRef {
+        lazy_field_ref!(Field::new(Self::FIELD_CHUNK_KEY, DataType::Binary, false))
     }
 
-    pub fn field_chunk_entity_path() -> Field {
-        Field::new(Self::FIELD_CHUNK_ENTITY_PATH, DataType::Utf8, false).with_metadata(
-            [("rerun:kind".to_owned(), "control".to_owned())]
-                .into_iter()
-                .collect(),
+    pub fn field_chunk_entity_path() -> FieldRef {
+        lazy_field_ref!(
+            Field::new(Self::FIELD_CHUNK_ENTITY_PATH, DataType::Utf8, false).with_metadata(
+                [("rerun:kind".to_owned(), "control".to_owned())]
+                    .into_iter()
+                    .collect(),
+            )
         )
     }
 
-    pub fn field_chunk_is_static() -> Field {
-        Field::new(Self::FIELD_CHUNK_IS_STATIC, DataType::Boolean, false).with_metadata(
-            [("rerun:kind".to_owned(), "control".to_owned())]
-                .into_iter()
-                .collect(),
+    pub fn field_chunk_is_static() -> FieldRef {
+        lazy_field_ref!(
+            Field::new(Self::FIELD_CHUNK_IS_STATIC, DataType::Boolean, false).with_metadata(
+                [("rerun:kind".to_owned(), "control".to_owned())]
+                    .into_iter()
+                    .collect(),
+            )
         )
     }
 
-    pub fn fields() -> Vec<Field> {
+    pub fn fields() -> Vec<FieldRef> {
         vec![
             Self::field_chunk_id(),
             Self::field_chunk_partition_id(),
@@ -285,23 +306,23 @@ impl FetchChunksRequest {
         ]
     }
 
-    pub fn field_chunk_id() -> Field {
+    pub fn field_chunk_id() -> FieldRef {
         QueryDatasetResponse::field_chunk_id()
     }
 
-    pub fn field_chunk_partition_id() -> Field {
+    pub fn field_chunk_partition_id() -> FieldRef {
         QueryDatasetResponse::field_chunk_partition_id()
     }
 
-    pub fn field_chunk_layer_name() -> Field {
+    pub fn field_chunk_layer_name() -> FieldRef {
         QueryDatasetResponse::field_chunk_layer_name()
     }
 
-    pub fn field_chunk_key() -> Field {
+    pub fn field_chunk_key() -> FieldRef {
         QueryDatasetResponse::field_chunk_key()
     }
 
-    pub fn fields() -> Vec<Field> {
+    pub fn fields() -> Vec<FieldRef> {
         vec![
             Self::field_chunk_id(),
             Self::field_chunk_partition_id(),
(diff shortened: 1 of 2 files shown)


crates/store/re_server/src/store/dataset.rs
```diff
@@ -168,7 +254,7 @@ impl Dataset {
 
         let layer_name = layer_name.unwrap_or(DataSource::DEFAULT_LAYER);
 
-        let mut new_partition_ids = BTreeSet::default();
+        let new_partition_ids = BTreeSet::default();
```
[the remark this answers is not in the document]
abey79: good catch, bad suggestion

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -551,7 +519,7 @@ impl RerunCloudService for RerunCloudHandler {
 
         let mut request = request.into_inner();
 
-        let mut chunk_stores = HashMap::new();
+        let mut chunk_stores = HashMap::default();
```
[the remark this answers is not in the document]
abey79: `ahash::HashMap` don't have `new()`...

crates/store/re_redap_tests/src/tests/query_dataset.rs
```diff
@@ -0,0 +1,176 @@
...
+        merged_chunk_info
+            .schema()
+            .contains(&Schema::new_with_metadata(
+                required_field.clone(),
+                Default::default()
+            )),
+        "query dataset must return all guaranteed fields\nExpected: {:#?}\nGot: {:#?}",
+        &required_field,
+        merged_chunk_info.schema()
+    );
```
abey79, line 174: This relies on field ordering, which fails with cloud because there is no ordering guarantees. I need to fix it.

crates/store/re_redap_tests/src/tests/query_dataset.rs
```diff
@@ -0,0 +1,190 @@
...
+        .map(Arc::clone)
+        .sorted_by_key(|field| field.name().clone())
+        .collect_vec();
+
+    let sorted_required_fields = required_field
+        .clone()
+        .into_iter()
+        .sorted_by_key(|field| field.name().clone())
+        .map(Arc::new)
+        .collect_vec();
```
abey79: arf this is verbose af. I'm fixing this is the next PR because I need to do the same elsewhere

crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
```diff
@@ -129,12 +168,151 @@ impl TryFrom<crate::cloud::v1alpha1::QueryDatasetRequest> for QueryDatasetReques
...
+    // These columns are guaranteed to be returned by `QueryDataset`. Additional columns may also be
+    // returned.
+    pub const FIELD_CHUNK_ID: &str = "chunk_id";
+    pub const FIELD_CHUNK_PARTITION_ID: &str = "chunk_partition_id";
+    pub const FIELD_CHUNK_LAYER_NAME: &str = "rerun_partition_layer";
+    pub const FIELD_CHUNK_KEY: &str = "chunk_key";
+    pub const FIELD_CHUNK_ENTITY_PATH: &str = "chunk_entity_path";
+    pub const FIELD_CHUNK_IS_STATIC: &str = "chunk_is_static";
+
+    pub fn field_chunk_id() -> Field {
```
Wumpf: do you think `Lazy`s would make sense for these? Maybe premature optimization, but could save some repeated allocs :thinking:
(or at least for the higher level `Schema`)

crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
```diff
@@ -1696,6 +1876,27 @@ impl TryFrom<QueryTasksRequest> for crate::cloud::v1alpha1::QueryTasksRequest {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use arrow::datatypes::ToByteSlice as _;
+
+    #[test]
+    fn test_query_dataset_response_create_dataframe() {
```
Wumpf, line 1903: would it make sense to check for the expected failure cases as well? they are very unclear to me right now and could use some documenting as well

crates/store/re_server/src/store/layer.rs
```diff
@@ -18,6 +22,21 @@ impl Layer {
...
+    }
+
+    pub fn last_updated_at(&self) -> jiff::Timestamp {
+        //TODO(ab): change this if we ever mutate a layer somehow?
+        self.registration_time
+    }
+
+    #[expect(clippy::unused_self)]
+    pub fn layer_type(&self) -> &'static str {
+        //TODO(ab): what should that actually be?
```
Wumpf, line 36: guess we might have non-rrd layers, like mcap in the future?

crates/store/re_server/src/store/layer.rs
```diff
@@ -29,6 +48,35 @@ impl Layer {
             .map(|chunk| chunk.heap_size_bytes())
             .sum()
     }
+
+    pub fn schema(&self) -> Schema {
+        let fields = self.store_handle.read().schema().arrow_fields();
+        Schema::new_with_metadata(fields, HashMap::default())
+    }
+
+    pub fn schema_sha256(&self) -> Result<[u8; 32], ArrowError> {
```
Wumpf, line 57: nit: be nice if `[u8; 32]` was newtyped to a sha256. Or at least type aliased

crates/store/re_server/src/store/in_memory_store.rs
```diff
@@ -55,6 +54,71 @@ impl InMemoryStore {
             .unwrap_or(ChunkStoreConfig::CHANGELOG_DISABLED)
     }
 
+    /// Returns the chunks corresponding to the provided chunk keys.
+    ///
+    /// The chunks are returned in the same order as they are requested.
```
Wumpf: doesn't look like they are - the incoming keys are absorbed in that hashmap monstrosity and then iterated via that

crates/store/re_server/src/store/partition.rs
```diff
@@ -31,6 +34,19 @@ impl Partition {
...
+    /// Iterate over layers.
+    ///
+    /// Layers are iterated in (registration time, layer name) order, as per how they should appear
+    /// in the partition table.
+    pub fn iter_layers(&self) -> impl Iterator<Item = (&str, &Layer)> {
+        self.layers
+            .iter()
+            .sorted_by(|(name_a, layer_a), (name_b, layer_b)| {
+                (layer_a.registration_time(), name_a).cmp(&(layer_b.registration_time(), name_b))
+            })
```
Wumpf, line 50:
```suggestion
            .sorted_by_key(|(name, layer)| (layer.registration_time(), name))
```

crates/store/re_server/src/store/dataset.rs
```diff
@@ -44,6 +47,32 @@ impl Dataset {
...
+            .get(partition_id)
+            .ok_or_else(|| Error::PartitionIdNotFound(partition_id.clone(), self.id))
+    }
+
+    pub fn partitions_from_ids<'a>(
+        &'a self,
+        partition_ids: &'a [PartitionId],
+    ) -> Result<impl Iterator<Item = (&'a PartitionId, &'a Partition)>, Error> {
+        if partition_ids.is_empty() {
+            Ok(Either::Left(self.partitions.iter()))
```
Wumpf, line 63: what? why are we returning all of them if we asked for none?! That's not documented and if it was I'd be complaining about it :P

crates/store/re_server/src/store/dataset.rs
```diff
@@ -74,59 +103,116 @@ impl Dataset {
...
+    //TODO(RR-2604): add support for property columns
     pub fn partition_table(&self) -> arrow::error::Result<RecordBatch> {
-        let (partition_ids, last_updated_at, num_chunks, size_bytes): (
+        let (partition_ids, last_updated_at, num_chunks, size_bytes, layer_names, storage_urls): (
+            Vec<_>,
+            Vec<_>,
             Vec<_>,
             Vec<_>,
             Vec<_>,
             Vec<_>,
```
Wumpf: scary looking :/
Can we just do a regular old loop + insert here? I couldn't tell whether one of those fields is swapped

crates/store/re_server/src/store/dataset.rs
```diff
@@ -74,59 +103,116 @@ impl Dataset {
...
+        ): (
+            Vec<_>,
+            Vec<_>,
+            Vec<_>,
+            Vec<_>,
+            Vec<_>,
+            Vec<_>,
+            Vec<_>,
+            Vec<_>,
+            Vec<_>,
```
Wumpf: as above. I refuse to do the required eyeballing to check whether things match up!

crates/store/re_redap_tests/src/tests/query_dataset.rs
```diff
@@ -0,0 +1,197 @@
...
+        (
+            "cannot specify entity paths if `select_all_entity_paths` is true",
+            QueryDatasetRequest {
+                entity_paths: vec!["/entity/path".into()],
+                select_all_entity_paths: true,
+                ..Default::default()
+            },
+            tonic::Code::InvalidArgument,
+        ),
+        //TODO(ab): add more failure cases
```
Wumpf: any reason you didn't go ahead here? does this warrant a ticket?

crates/store/re_redap_tests/src/utils/arrow.rs
```diff
@@ -278,6 +302,7 @@ impl RecordBatchExt for arrow::array::RecordBatch {
         }
     }
 
+    /// Remove the named columns.
     fn unfiltered_columns(&self, columns: &[&str]) -> Self {
```
Wumpf, line 306: More like `columns_except`? Took me a bit to get what unfiltered means

crates/store/re_server/src/store/dataset.rs
abey79: That is not new. A common pattern in our protos, and a rare instance of something that _is_ actually documented:

https://github.com/rerun-io/rerun/blob/20ddeddbaf2d030b15d680738a4f40763c54e85b/crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto#L427-L430

It's also tested.

Adding a docstring to clarify.

crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
abey79: The result comes from arrow's record batch creation, so given that we ensure consistency between schema and array types, the only failure case I can imagine is mismatch vec length. We _could_ add a test for that, but at this point it's more like a arrow unit test :)

crates/store/re_server/src/store/layer.rs
abey79: When you do, you will have a much larger problems than just this function. This will be tested pretty directly by snapshots in `re_redap_tests` so there is zero chance we "forget" about updating this.

crates/store/re_server/src/store/layer.rs
abey79: On one hand, yeah. And that new type be shared with `dataplatform` (including the actual sha computation that I copy/pasted from over there).

In practice, too much abstraction creates unnecessary coupling in that case, maybe? End of the day, all that matters is this:

https://github.com/rerun-io/rerun/blob/20ddeddbaf2d030b15d680738a4f40763c54e85b/crates/store/re_redap_tests/src/tests/snapshots/re_redap_tests__tests__register_partition__simple_with_layers_manifest_schema.snap#L11

crates/store/re_server/src/store/in_memory_store.rs
abey79: good catch. they were at some point the the spec changed in the process.

crates/store/re_redap_tests/src/tests/query_dataset.rs
abey79: So much that needs to be done around here :)

I've started a shopping list here: https://github.com/rerun-io/rerun/issues/11591

crates/store/re_redap_tests/src/utils/arrow.rs
abey79: I agree it's bad, but this is shared with `dataplatform` and I'd rather not change that now.

crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
abey79: good call, rolling in a small helper

abey79: also, yes, most likely premature optimisation 😅

PR #11534, opened 2025-10-14
Turn on `clippy::cast_possible_wrap`

emilk:
### Related
* Sibling PR: https://github.com/rerun-io/dataplatform/pull/1850
* Would have caught https://github.com/rerun-io/rerun/pull/11525
* https://github.com/rerun-io/rerun/issues/11516
* See <https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_wrap>

### What
The `as` casting in Rust is as convenient as it is dangerous.

This caught a few legit issues (not as big as the original bug though).

[`saturating_cast`](https://docs.rs/saturating_cast/latest/saturating_cast/) was a very helpful crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -576,9 +576,12 @@ debug = true # enable debug symbols for build scripts
 [workspace.lints.rust]
 unsafe_code = "deny"
 
+absolute_paths_not_starting_with_crate = "warn"
 elided_lifetimes_in_paths = "warn"
 future_incompatible = { level = "warn", priority = -1 }
+non_ascii_idents = "warn"
 nonstandard_style = { level = "warn", priority = -1 }
+noop_method_call = "warn"
 rust_2018_idioms = { level = "warn", priority = -1 }
 rust_2021_prelude_collisions = "warn"
 semicolon_in_expressions_from_macros = "warn"
@@ -604,6 +607,7 @@ allow_attributes = "warn"
 as_ptr_cast_mut = "warn"
 await_holding_lock = "warn"
 bool_to_int_with_if = "warn"
+cast_possible_wrap = "warn"
 char_lit_as_u8 = "warn"
 checked_conversions = "warn"
 clear_with_drain = "warn"
@@ -766,6 +770,7 @@ map_unwrap_or = "allow" # so is this one
 ref_patterns = "allow" # It's nice to avoid ref pattern, but there are some situations that are hard (impossible?) to express without.
 
 # TODO(emilk): enable more of these lints:
+cast_possible_truncation = "allow" # Moo much noise, sadly
 iter_over_hash_type = "allow"
 let_underscore_untyped = "allow"
 missing_assert_message = "allow"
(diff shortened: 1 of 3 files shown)


crates/utils/re_memory/src/memory_limit.rs
```diff
@@ -22,7 +24,7 @@ impl MemoryLimit {
     /// Set the limit to some number of bytes.
     pub fn from_bytes(max_bytes: u64) -> Self {
         Self {
-            max_bytes: Some(max_bytes as _),
+            max_bytes: Some(max_bytes.saturating_cast()),
```
emilk, line 27: This was a likely bug, if someone did `from_bytes(u64::MAX)` intending for "no limit", and instead getting a _negative_ limit

crates/viewer/re_view_spatial/src/contexts/depth_offsets.rs
```diff
@@ -53,13 +54,14 @@ impl ViewContextSystem for EntityDepthOffsets {
             .values()
             .map(|entities| entities.len())
             .sum();
-        let mut depth_offset = -((num_entities_with_draw_order / 2) as re_renderer::DepthOffset);
+        let mut depth_offset =
+            -(num_entities_with_draw_order / 2).saturating_cast::<re_renderer::DepthOffset>();
```
emilk, line 58: This would have been buggy with more than 32ki entities

Cargo.toml
```diff
@@ -765,6 +770,7 @@ map_unwrap_or = "allow" # so is this one
 ref_patterns = "allow" # It's nice to avoid ref pattern, but there are some situations that are hard (impossible?) to express without.
 
 # TODO(emilk): enable more of these lints:
+cast_possible_truncation = "allow" # Moo much noise, sadly
```
teh-cmc: image

crates/utils/re_auth/src/workos.rs
```diff
@@ -407,7 +408,7 @@ impl AccessToken {
 
     pub fn is_expired(&self) -> bool {
         // Time in seconds since unix epoch
-        let now = jsonwebtoken::get_current_timestamp() as i64;
+        let now = jsonwebtoken::get_current_timestamp().saturating_cast::<i64>();
         let seconds_left = now - self.expires_at;
         seconds_left <= SOFT_EXPIRE_SECS
```
oxkitsune, line 413: This is unrelated, but isn't this supposed to be:
```rs
let seconds_left = self.expires_at - now;
```

currently if `self.expires_at` is in the future, `seconds_left` will be negative meaning this function will return `true`?

cc @jprochazk

emilk: Good catch - I'll leave it to @jprochazk to fix in another PR

jprochazk: Yep. This is fixed in https://github.com/rerun-io/rerun/pull/11497/files#diff-51313fc74be2df4970510ca8f86e048f098da206367ff01fc808bc9f811b2fa9R260-R268

Issue #9285, opened 2025-03-14
DataframePart and Schema need to have optional bytes field


zehiko:
Title.

Wumpf:
They have both payloads that are optional byte fields. I think that's what you meant?

PR #11536, opened 2025-10-14
Remove the entire concept of "wire encoding"

teh-cmc:
TODO: I'll go into more details once I'm 200% confident this is the right approach. Need to build a few more things on top to figure that out.

* Sibling: https://github.com/rerun-io/dataplatform/pull/1852

PR #11537, opened 2025-10-14
Mark `RerunChunk::payload` as explicitly optional

teh-cmc:
Before we end up with a BW nightmare on our hands.

* DNM: requires #11536 
* Actually fixes https://github.com/rerun-io/rerun/issues/9285 for real


teh-cmc:
I'm going to ship that as part of the re_log_encoding PRs to make the process easier on everyone.

PR #11538, opened 2025-10-14
Change the arrow up / arrow down so that ascending sort points up

pweids:
This caused significant mental confusion for a user we watched. We should change to align with convention.

See, for example, MSFT DataGrid

"If DataGridColumn.SortDirection property is set to Ascending, an ascending sort icon (upward facing arrow) will be shown to the right of the column header indicating that the specific column has been sorted in the ascending order."

https://learn.microsoft.com/en-us/windows/communitytoolkit/controls/datagrid_guidance/group_sort_filter#2-sorting

### DIFF
--- a/crates/viewer/re_chunk_store_ui/src/sort.rs
+++ b/crates/viewer/re_chunk_store_ui/src/sort.rs
@@ -40,6 +40,7 @@ pub(crate) fn sortable_column_header_ui<T: Default + Copy + PartialEq>(
     let is_sorted = &sort_column.column == column;
     let direction = sort_column.direction;
 
+    //TODO(ab): this UI is really bad, we need to trash all of this in favor of the new table widget
     let (left_clicked, right_clicked) = egui::Sides::new()
         .height(tokens.table_row_height(table_style))
         .show(
@@ -50,12 +51,16 @@ pub(crate) fn sortable_column_header_ui<T: Default + Copy + PartialEq>(
                 ui.button(egui::WidgetText::from(label).strong()).clicked()
             },
             |ui| {
-                ui.button(match (is_sorted, direction) {
-                    (true, SortDirection::Ascending) => "↑",
-                    (true, SortDirection::Descending) => "↓",
-                    _ => "",
-                })
-                .clicked()
+                if is_sorted {
+                    let (icon, text) = match direction {
+                        SortDirection::Ascending => (&re_ui::icons::ARROW_UP, "ascending sort"),
+                        SortDirection::Descending => (&re_ui::icons::ARROW_DOWN, "descending sort"),
+                    };
+
+                    ui.small_icon_button(icon, text).clicked()
+                } else {
+                    false
+                }
             },
         );
 


emilk:
Let's add a snapshot test to cover this @abey79 

I would expect these icons, btw:
- ▲  for ascending (small on top)
- ▼  for decending (large on top)

with the icons showing _the current state_ (rather than the state you would get after a click)

pweids:
@emilk the current state is when no sorting is applied, there's no arrow in the column header:
image

When ascending sorting is applied, we now get this arrow:
image

And this PR just switches the arrows, so now it looks like this for ascending sort:
image
image

lucasmerlin:
> * ▲  for ascending (small on top)
> 
> * ▼  for decending (large on top)

In my opinion these triangles are way more intuitive than the arrows. With the arrows I'd expect it to point into the opposite direction (with ascending the natural order is top to bottom, so arrow points down, with descending it's bottom to top, so arrow points up). I don't work much with tables though, so my opinion is probably flawed

crates/viewer/re_dataframe_ui/tests/snapshots/test_ascending.png
emilk, line 1: I find the use of arrows very confusing. Here it looks like the arrow points from Z to A in sorting order? So… backwards?

Image

(assuming the icon here shows the _current_ state)

I think using ▲ and ▼ instead of ↑ and ↓ is A LOT clearer. With `▲` you can clearly see "smallest on top, largest on bottom".

An icon like this is also good:

Image

Again, it's easy to see "largest on top, smallest on bottom"

abey79:
> > * ▲  for ascending (small on top)
> > * ▼  for decending (large on top)
> 
> In my opinion these triangles are way more intuitive than the arrows. With the arrows I'd expect it to point into the opposite direction (with ascending the natural order is top to bottom, so arrow points down, with descending it's bottom to top, so arrow points up). I don't work much with tables though, so my opinion is probably flawed

I'm not a fan of arrows either. I think common icons for this are the following, which would work well with our icon style.

image

gavrelina:
Please no icons like these lines. They are used for filters. The triangles are better, but they also look like chevrons or drop down menu. Keep the arrow as they are now for this specific task. I will think of another icon if needed and have time, because a new icon also should be in a hormonised style with the rest and not used from symbols or whatever. 
image

crates/viewer/re_chunk_store_ui/src/sort.rs
```diff
@@ -51,8 +51,8 @@ pub(crate) fn sortable_column_header_ui<T: Default + Copy + PartialEq>(
             },
             |ui| {
                 ui.button(match (is_sorted, direction) {
-                    (true, SortDirection::Ascending) => "↓",
-                    (true, SortDirection::Descending) => "↑",
+                    (true, SortDirection::Ascending) => "↑",
+                    (true, SortDirection::Descending) => "↓",
```
emilk: We should use icons here too, not emojis

crates/viewer/re_dataframe_ui/tests/ascending_descending.rs
```diff
@@ -0,0 +1,178 @@
...
+#[tokio::test]
+async fn test_descending() {
+    let (session_context, table_ref) = prepare_session_context();
+    let test_context = TestContext::new();
+    let runtime_handle = AsyncRuntimeHandle::from_current_tokio_runtime_or_wasmbindgen().unwrap();
+
+    let table_status = Arc::new(Mutex::new(None::<TableStatus>));
+    let mut harness = test_context
+        .setup_kittest_for_rendering()
+        .with_size([600.0, 400.0])
```
emilk, line 77: You will need to merge in main and change this code once https://github.com/rerun-io/rerun/pull/11542 is merged

gavrelina:
My proposal is something like this, will do in a separate PR myself after redesign it to match our style! 
image

crates/viewer/re_dataframe_ui/tests/ascending_descending.rs
abey79: Well, that is unless I'm allowed merging this first.

crates/viewer/re_dataframe_ui/tests/ascending_descending.rs
```diff
@@ -0,0 +1,178 @@
...
+        .register_batch(table_ref, batch)
+        .expect("Failed to register the table");
+
+    (session_context, table_ref)
+}
+
+async fn run_async_harness(
+    harness: &mut Harness<'_>,
+    table_status: &Arc<Mutex<Option<TableStatus>>>,
+) {
```
lucasmerlin, line 155: Note that we do have a [step_until](https://github.com/rerun-io/rerun/blob/d4f895066a405f6511d220f2bda1e5e586861c26/crates/viewer/re_viewer/src/viewer_test_utils/mod.rs#L53) that does basically this (but without async). Maybe we should combine those. I actually think the async version is nicer since for step_until we need to use a multi threaded tokio runtime since the main tokio thread is blocked by the test, which is kind of a code smell.

emilk:
I'll get the added tests merged in a different PR:
* https://github.com/rerun-io/rerun/pull/11561

emilk:
@pweids sorry for the long spammy thread here with so many people piling on. Your PR brought up an important observation though: obviously different people interpret the arrows differently, and so we need a better icon. @gavrelina will work on it later this week. In the meantime I'm gonna close this PR since I don't want to change the direction of the arrow in one release only to change it to a completely different icon in the next release.

PR #11540, opened 2025-10-15
Use more deterministic rendering in snapshot tests

emilk:
### What
Disable alpha-to-coverage and skybox dithering in snapshot tests.

This is to (hopefully) produce more consistent results across different platoforms/GPUs/drivers.

View the difference here: <https://rerun-io.github.io/kitdiff/?url=https://github.com/rerun-io/rerun/pull/11540>

### DIFF
--- a/crates/viewer/re_view_spatial/tests/draw_order.rs
+++ b/crates/viewer/re_view_spatial/tests/draw_order.rs
@@ -1,9 +1,6 @@
 use re_chunk_store::RowId;
 use re_log_types::TimePoint;
-use re_test_context::{
-    TestContext,
-    external::egui_kittest::{OsThreshold, SnapshotOptions},
-};
+use re_test_context::{TestContext, external::egui_kittest::SnapshotOptions};
 use re_test_viewport::TestContextExt as _;
 use re_viewer_context::{ViewClass as _, ViewId};
 use re_viewport_blueprint::ViewBlueprint;


crates/viewer/re_renderer/src/view_builder.rs
```diff
@@ -183,11 +183,27 @@ impl Projection {
...
+    Deterministic,
+}
+
 /// Basic configuration for a target view.
 #[derive(Debug)]
 pub struct TargetConfiguration {
     pub name: DebugLabel,
 
+    /// Aim for beauty or determinism?
+    pub render_mode: RenderMode,
```
Wumpf, line 205: Now that I found the global `RenderConfig` again I'm wondering whether that would be a better place since we already control MSAA there (pretty much for testing only). Obviously the placement in this PR is much more flexible 🤷 
(controlling MSAA has much deeper effects on all sort of resource creation than this new shader variable)

crates/viewer/re_renderer/shader/generic_skybox.wgsl
```diff
@@ -59,12 +59,14 @@ fn main(in: FragmentInput, @builtin(position) frag_coord: vec4<f32>) -> @locatio
...
+        // Apply dithering in gamma space.
+        // TODO(andreas): Once we switch to HDR outputs, this can be removed.
+        //                As of writing, the render target itself is (s)RGB8, so we need to dither while we still have maximum precision.
+        var rgb_gamma_dithered = dither_interleaved(rgb, 256.0, frag_coord);
 
-    return vec4f(linear_from_srgb(rgb_gamma_dithered), 1.0);
-    //return vec4f(linear_from_srgb(rgb), 1.0); // Without dithering
-    //return vec4f(camera_dir, 1.0);
+        return vec4f(linear_from_srgb(rgb_gamma_dithered), 1.0);
+    }
```
Wumpf, line 71: as pointed out on slack I believe we should leave the dithering on here. `dither_interleaved` doesn't have any trigonometry, so we can assume that any _somewhat_ IEEE respecting implementation will come to the same result. Also, skybox changes fell under the threshhold every time so far 🤷

Wumpf (approved):
nice!

crates/viewer/re_renderer/shader/generic_skybox.wgsl
emilk: I would like to leave dithering off for now, then lower our fail-thresholds, and then maybe re-evaluate once we care about generating _nice_ screenshots from out snapshots (which currently is _not_ a priority).

PR #11542, opened 2025-10-15
Consistent and lower thresholds on snapshot tests

emilk:
This introduces consistent error threshold across tests (mostly), with a slightly higher tolerance for tests that include non-ui elements (e.g. 3D rendering).

View diffs: https://rerun-io.github.io/kitdiff/?url=https://github.com/rerun-io/rerun/pull/11542

## Related
* Awaiting https://github.com/rerun-io/rerun/pull/11551

### DIFF
--- a/crates/viewer/re_selection_panel/src/selection_panel.rs
+++ b/crates/viewer/re_selection_panel/src/selection_panel.rs
@@ -1348,7 +1348,6 @@ mod tests {
         harness.snapshot("selection_panel_component_hybrid_overwrite");
     }
 
-    #[ignore = "TODO(#11546)"]
     #[test]
     fn selection_panel_view_snapshot() {
         let mut test_context = get_test_context();
@@ -1389,7 +1388,6 @@ mod tests {
         harness.snapshot("selection_panel_view");
     }
 
-    #[ignore = "TODO(#11546)"]
     #[test]
     fn selection_panel_view_entity_no_visualizable_snapshot() {
         let mut test_context = get_test_context();
@@ -1431,7 +1429,6 @@ mod tests {
         harness.snapshot("selection_panel_view_entity_no_visualizable");
     }
 
-    #[ignore = "TODO(#11546)"]
     #[test]
     fn selection_panel_view_entity_no_match_snapshot() {
         let mut test_context = get_test_context();


crates/viewer/re_selection_panel/src/selection_panel.rs
```diff
@@ -1373,6 +1362,7 @@ mod tests {
         harness.snapshot("selection_panel_component_hybrid_overwrite");
     }
 
+    #[ignore = "TODO(#11546)"]
```
emilk: * https://github.com/rerun-io/rerun/issues/11546

crates/viewer/re_ui/src/testing.rs
```diff
@@ -0,0 +1,49 @@
...
+    let num_total_pixels = viewport_size.x * viewport_size.y;
+
+    let broken_pixels_fraction = 1e-4;
+    let max_broken_pixels = (num_total_pixels * broken_pixels_fraction).round() as usize;
+
+    let threshold = 1.3; // Must be pretty high because of 3D grid :/
+
+    SnapshotOptions::default()
+        .threshold(threshold)
+        .failed_pixel_count_threshold(max_broken_pixels)
```
lucasmerlin, line 48: I'm not a huge fan of hardcoding this. Say we have a 500x500 snapshot test, that would allow for 25 broken pixels. 
I just tested it and adding/removing an icon button with the ADD icon only affects 21 pixels.
So if one of the 3d tests only had 4 broken pixels we could easily miss an icon disappearing. 

With a higher threshold you could miss slight changes in color on the whole snapshot. With the broken pixel threshold you could miss widgets appearing / disappearing. In my opinion the later case is worse.

I guess ideally we had some way to define a threshold for the broken pixels.
Like this:
```
default threshold: 1.0
broken_pixel_thresholds:
    10: 5.0     // 10 pixels may have changes as high as 5.0
    5: 10.0.    // 5 pixels can have changes as high as 10.0
```

emilk: I agree this is sub-optimal, but on `main` the `broken_pixel_fraction` is 40x higher than in this PR, so this is a _huge_ improvement.

We can experiment with increasing the `theshold` and lowering `broken_pixels_fraction` further in subsequent PRs.

lucasmerlin (approved):
Awesome!

PR #11543, opened 2025-10-15
Transform frame id concept definition & `TransformForest` migration

Wumpf:
### Related

* Part of RR-2514

### What

* Introduce `TransformFrameId`
* Introduce `TransformFrameIdHash`
* Make `TransformForest` work in terms of frame ids

At this point this is all just internal refactor, not changing any behavior towards the outside!

Next up: Make `TransformResolutionCache` work with frame ids

### DIFF
--- a/crates/store/re_tf/src/transform_frame_id_hash.rs
+++ b/crates/store/re_tf/src/transform_frame_id_hash.rs
@@ -40,8 +40,9 @@ impl std::cmp::PartialEq for TransformFrameIdHash {
 }
 
 impl From<&TransformFrameId> for TransformFrameIdHash {
+    #[inline]
     fn from(value: &TransformFrameId) -> Self {
-        Self(Hash64::hash(&value.0))
+        Self::new(value)
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/viewer/re_view_spatial/src/visualizers/transform3d_arrows.rs
```diff
@@ -107,18 +107,23 @@ impl VisualizerSystem for Transform3DArrowsVisualizer {
...
+                    pinhole_tree_root_info.parent_root_from_pinhole_root
+                } else {
                     // We're inside a 2D space. But this is a 3D transform.
                     // Something is wrong here and this is not the right place to report it.
                     // Better just don't draw the axis!
                     continue;
-                } else {
-                    // Don't apply the from-2D transform, stick with the last known 3D.
-                    pinhole_tree_root_info.parent_root_from_pinhole_root
                 }
```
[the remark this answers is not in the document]
Wumpf: inversion is correct

crates/store/re_tf/src/transform_frame_id_hash.rs
```diff
@@ -0,0 +1,150 @@
...
+impl std::cmp::PartialEq for TransformFrameIdHash {
+    #[inline]
+    fn eq(&self, other: &Self) -> bool {
+        self.0.eq(&other.0)
+    }
+}
+
+impl From<&TransformFrameId> for TransformFrameIdHash {
+    fn from(value: &TransformFrameId) -> Self {
+        Self(Hash64::hash(&value.0))
```
oxkitsune: This implementation is different than `TransformFrameId::new` feels feels like a bit of a footgun

Wumpf: true, let's equalize. It's actually just wrong...

PR #11544, opened 2025-10-15
Crate level documentation for `re_tf`

Wumpf:
Most of it is aspirational. Will iterate on it as we go, but should serve as an overview of what's going on in `re_tf`

Based on https://github.com/rerun-io/rerun/pull/11543 to make some doc links work


crates/store/re_tf/src/lib.rs
```diff
@@ -41,28 +39,65 @@
...
 //! ```text
-//! robot_frame -> rerun_tf#/world
-//! rerun_tf#/world/robot/right_arm -> robot_frame
+//!                                     ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐                                   ┌──────────────────┐
+//!                                            rerun_tf#world      ◀ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─│      world       │
+//!                                     └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘                                   └──────────────────┘
+//!                                                   │                                                           │
+//!         ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┼ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐                                 │
+//!                                                   │                                                           │
+//!         │                                         │                         │                                 │
```
Wumpf: missing `▼` down on the left

crates/store/re_tf/src/lib.rs
```diff
@@ -1,4 +1,140 @@
...
+//!               ▲                                                              │
+//!               │
+//!                ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘
+//! ```
+//!
+//! You can interact with these impliciately generated frames like with any other transform frame!
+//! For example, let's say we log a manual transform relationship between two new frames called `robot_frame`
+//! and `left_frame`, associate them with `world/robot` and `world/robot/left` respectively.
+//! That would create two unconnected trees, but this can be handled by specifying another
+//! relationship from `robot` to `rerun_tf#world/robot`, leading to this setup:
```
oxkitsune:
```suggestion
//! relationship from `robot_frame` to `rerun_tf#world/robot`, leading to this setup:
```

crates/store/re_tf/src/lib.rs
```diff
@@ -1,4 +1,140 @@
...
+//! #### Entity relationship & built-in transform frames
+//!
+//! TODO(RR-2487): Most things in this paragraph are planned but not yet implemented.
+//!
+//! Every entity is associated with a transform frame.
+//! The transform frame can be set with the `CoordinateFrame` archetype.
+//! TODO(RR-2486): Link to respective archetype.
+//!
+//! However, by default, it points to an implicit, entity-derived transform frame.
+//! The name of the implicit transform frames is the entity path, prefixed with `rerun_tf#`, e.g. `rerun_tf#/world/robot/arm`.
```
grtlr, line 29: Warning, bike shedding. Similar to how we remove the `rerun.archetype.` prefix from component identifiers, I think we should remove the `rerun_` from the entity-derived name here.

It will make it a bit nicer for our users (imagine seeing that identifier over and over again in a table, for example).

crates/store/re_tf/src/lib.rs
```diff
@@ -1,4 +1,140 @@
...
+//!               ╱                             ▼                                ╱                         ╲
+//! ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐ ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐        ┌──────────────────┐      ┌──────────────────┐
+//!   rerun_tf#world/robot/left    rerun_tf#world/robot/right          │ world/robot/left │      │world/robot/right │
+//! └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘ └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘        └──────────────────┘      └──────────────────┘
+//!               ▲                                                              │
+//!               │
+//!                ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘
+//! ```
+//!
+//! You can interact with these impliciately generated frames like with any other transform frame!
```
grtlr:
```suggestion
//! You can interact with these implicitly generated frames like with any other transform frame!
```

crates/store/re_tf/src/lib.rs
Wumpf: yeah I wanted to kick off some naming discussion on what we call the implicit transform frames. But I can just go with `tf#` right away in this pr to make it a bit saner already

Wumpf: pushing this off.

To be clear though: I don't think it makes sense to have a different display name than the underlying frame id string. Frame id strings are a thing you actually need to known exactly. Better implicit frame ids I'm all for

crates/store/re_tf/src/lib.rs
Wumpf: good catch!

crates/store/re_tf/src/lib.rs
grtlr: I think we are "arguing" for the same thing 😊. I meant actually changing the implicit frame ids and not just hiding them in the UI, just as we do with component identifiers (we don't abbreviate `rerun.archetypes.Point3D:positions` in the UI, the identifier _is just_ `Points3D:positions`).

PR #11551, opened 2025-10-15
More deterministic 3D snapshots

emilk:
* Fade out 3D grid slightly earlier (only in tests)
* Make the "you are rotating around this"-gizmo in the 3D view dependent on egui time instead of wall time
* Hide the "you are rotating around this"-gizmo in our tests (why was it ever there? 🤷‍♂️)

I _believe_ this fixed the last _major_ inconsistencies between platforms, allowing us to lower our thresholds in a subsequent PR:
* https://github.com/rerun-io/rerun/pull/11542

View diffs: https://rerun-io.github.io/kitdiff/?url=https://github.com/rerun-io/rerun/pull/11551

### DIFF
--- a/crates/viewer/re_view_spatial/tests/snapshots/transparent_capsules3d_0.png
+++ b/crates/viewer/re_view_spatial/tests/snapshots/transparent_capsules3d_0.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:8dc9288906b7cdfd45ea950872aac49619d194faa58cf5f0576744eb4ebd4ec8
-size 39862
+oid sha256:46c7ef1ca92398f99d7eafd2a3016ece33d29783bb0c5d7fac7608599d577d56
+size 34757
--- a/crates/viewer/re_view_spatial/tests/transparent_geometry.rs
+++ b/crates/viewer/re_view_spatial/tests/transparent_geometry.rs
@@ -81,6 +81,10 @@ fn test_transparent_geometry<A: AsComponents>(
                         ),
                         glam::Vec3::Z,
                     ));
+
+                    // Hack: without this the view will svivel to the default eye/view.
+                    // TODO(#8265): we shouldn't need this hack.
+                    view_state.state_3d.last_eye_interaction = Some(-f64::INFINITY);
                 }
 
                 test_context.run_with_single_view(ui, view_id);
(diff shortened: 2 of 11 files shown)


crates/viewer/re_view_spatial/tests/transparent_geometry.rs
```diff
@@ -81,7 +82,6 @@ fn test_transparent_geometry<A: AsComponents>(
                         ),
                         glam::Vec3::Z,
                     ));
-                    view_state.state_3d.last_eye_interaction = Some(std::time::Instant::now());
```
emilk, line 84: No idea why someone thought it would be a good idea to show the rotation gizmo in tests. If you don't add a why-comment to your code, don't be surprised if your code disappear 😈

crates/viewer/re_view_time_series/tests/snapshots/blueprint_overrides_and_defaults_with_time_series.png
emilk, line 1: I did a `UPDATE_SNAPSHOTS=force` just to make sure I didn't miss any 3D tests… I guess that meant some spurious updates to 2D tests too 🤷

crates/viewer/re_view_spatial/tests/transparent_geometry.rs
lucasmerlin: I guess it would make sense if it is tested somewhere? We would want to know if it changes / breaks

crates/viewer/re_view_spatial/tests/snapshots/transparent_capsules3d_0.png
lucasmerlin, line 1: Why did these change so much? The camera position shouldn't have changed right?

emilk: Good question 🤔

crates/viewer/re_view_spatial/tests/transparent_geometry.rs
emilk: This change is what caused the camera to move in the transparency checks O.o
WEIRD

emilk: Probably because of https://github.com/rerun-io/rerun/blob/1622e85d4ffee6fc29c2064c1e543394b7d9e7a9/crates/viewer/re_view_spatial/src/ui_3d.rs#L121-L130

lucasmerlin: Yeah I think this line is not only to show the gizmo but it also affects the actual rotation in the test

lucasmerlin (approved):
Nice!

PR #11405, opened 2025-10-02
Make time and timeline controlled by blueprint

IsseW:
### Related

Closes #7728

### What

Makes time cursor and current timeline be controlled by the blueprint.

### Details

We log the time as static for blueprints for it not to affect undo state.

This includes some refactors that felt necessary to make tests still work and not be overly complicated:
- `TestContext` doesn't require a mutable reference to run.
- Can update `TestContext` blueprints via `TestContext::with_blueprint_ctx`.
- The blueprint ctx is a struct that implements a helper trait for interacting with blueprints. App also has an equivalent `AppBlueprintCtx`, and `ViewerContext` also implements this trait.

The time panel was refactored a bit to allow the normal time panel and blueprint inspector time panel to still share code. This is mostly changing `&mut TimeControl` to `&mut impl TimeControlCtx` which handles writing to blueprints/just updating time control whether this is the blueprint inspector time panel or not.

PR #11554, opened 2025-10-15
Allow eye-camera tracking arbitrary entities in 3D view

Gentlegg:
Lot of similarities except when we enter tracking there is no interpolation, i kept the ALT + double click functionality and added a tracking option to the context menu actions.

### Related
* Part of #11048

### DIFF
--- a/crates/viewer/re_view_spatial/tests/snapshots/help_view_3d_view_mac.png
+++ b/crates/viewer/re_view_spatial/tests/snapshots/help_view_3d_view_mac.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:ea346bc082627e801eef666433dbb9cf1adf84643d7a7330774aa5702761ab5d
-size 21746
+oid sha256:95c42b7d29d11683114db8499d7c84b9fb7553ac7fbd9babbd9ded8588cae3f1
+size 25055
--- a/crates/viewer/re_view_spatial/tests/snapshots/help_view_3d_view_windows.png
+++ b/crates/viewer/re_view_spatial/tests/snapshots/help_view_3d_view_windows.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:20614b5981053aa9077c358b1bed8a78dcdc68d29e52f738f06e34388f3a7c21
-size 22870
+oid sha256:bc8d7fa2126a1a3600f3cd1fa84e596195de2e6c30543c4631bddea16b1de51f
+size 26365


crates/viewer/re_viewer/src/app.rs
```diff
@@ -1103,6 +1106,21 @@ impl App {
...
+
+                let state = self.state.view_states.get_mut_or_create(view_id, anyad);
+
+                let asdasd = state.downcast_mut::<SpatialViewState>().unwrap();
+
+                asdasd.state_3d.tracked_entity = Some(entity_path);
+
+                // asdasd
+                //     .state_3d
+                //     .track_entity(&entity_path, &asdasd.bounding_boxes);
```
Wumpf: please review & clean up your code before opening pull requests ;)

crates/viewer/re_viewer_context/src/command_sender.rs
```diff
@@ -132,6 +132,8 @@ pub enum SystemCommand {
     /// Add a task, run on a background thread, that saves something to disk.
     #[cfg(not(target_arch = "wasm32"))]
     FileSaver(Box<dyn FnOnce() -> anyhow::Result<std::path::PathBuf> + Send + 'static>),
+
+    SetTracked(ViewId, EntityPath),
```
Wumpf: I've been more lenient on this before since we had a lot of properties that are not part of the blueprint, but as we are [now](https://github.com/rerun-io/rerun/pull/11405) [actively](https://github.com/rerun-io/rerun/pull/11494) [moving](https://github.com/rerun-io/rerun/pull/11490) [things](https://github.com/rerun-io/rerun/pull/11522) over I really don't want to add more transient state. Especially not when they require adding new single-view-only commands like this.

Let's instead model the data not in the view state but in a view property (like background color etc.). @IsseW will soon move the rest of the camera state into blueprint, so we can add an optional tracking property to that same view-property archetype.
This then in turn eliminates the need for a special action and instead will be a blueprint write like any other. (plus all the other benefits of being able to control it from python code and persist it on disk)

crates/viewer/re_view_spatial/src/ui.rs
```diff
@@ -154,6 +159,25 @@ impl SpatialViewState {
                 self.state_3d.set_spin(spin);
             }
         }
+
+        ui.horizontal(|ui| {
+            if let Some(tracked_entity) = self.state_3d.tracked_entity.clone() {
+                let (latest_at_query, _) =
+                    item_ui::guess_query_and_db_for_selected_entity(ctx, &tracked_entity);
```
Wumpf: why not just read the current query from the `ctx` directly? `guess_query_and_db_for_selected_entity` does a lot more work and can produce the wrong result in this context

edit: With proposed blueprint driven property you'll get the UI for this fully automatically, so doesn't matter.
Not quite as nice with the entity path button IIRC but that can be solved separately then.

crates/viewer/re_context_menu/src/actions/track_entity.rs
```diff
@@ -0,0 +1,71 @@
...
+        view_id: &ViewId,
+        instance_path: &re_entity_db::InstancePath,
+    ) {
+        ctx.viewer_context
+            .global_context
+            .command_sender
+            .send_system(SystemCommand::SetTracked(
+                *view_id,
+                instance_path.entity_path.clone(),
+            ));
```
Wumpf: Using the information in `ctx` you should be able to build up `ViewProperty::from_archetypes::<EyeControls3D>(...ctx stuff...)` and then use that to set a newly introduced tracking field on that.

crates/viewer/re_context_menu/src/actions/track_entity.rs
```diff
@@ -0,0 +1,71 @@
...
+        .viewport_blueprint
+        .views
+        .iter()
+        .filter(|x| x.0 == view_id)
+    {
+        if current_views.1.class_identifier() == ViewClassIdentifier::new("3D") {
+            is_3d = true;
+        }
+    }
+    is_3d
```
Wumpf: Strange for loop as view ids are unique. Should just use `Iterator::any` instead

crates/viewer/re_context_menu/src/actions/track_entity.rs
```diff
@@ -0,0 +1,71 @@
...
+            | Item::ComponentPath(_)
+            | Item::InstancePath(_) => false,
+            Item::DataResult(view_id, instance_path) => {
+                let mut show = false;
+
+                if is_3d_view(ctx, view_id)
+                    && ctx
+                        .viewer_context
+                        .recording()
+                        .is_logged_entity(&instance_path.entity_path)
```
Wumpf, line 31: don't understand why this check is needed. If we have a context menu on a data result then surely it was logged on the current recording?

crates/viewer/re_viewer/src/app.rs
Gentlegg: sorry for that :p

crates/viewer/re_context_menu/src/actions/track_entity.rs
Gentlegg: its for the root and other containers that are not logged, for instance in the DNA example
image

Wumpf: ahh nice. That makes sense. Can you plz leave a comment (can be by example)? thx!

crates/viewer/re_context_menu/src/actions/track_entity.rs
```diff
@@ -0,0 +1,70 @@
...
+            | Item::RedapEntry(_)
+            | Item::RedapServer(_)
+            | Item::ComponentPath(_)
+            | Item::InstancePath(_) => false,
+            Item::DataResult(view_id, instance_path) => {
+                let mut show = false;
+
+                if is_3d_view(ctx, view_id)
+                    // we need to check if the focused entity is logged
+                    // because containers doesn't have bounding box or position
```
Wumpf:
```suggestion
                    // because entities without any data don't have bounding boxes or positions.
```

`container` is what we call the view containers like `horizontal` etc.

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -555,6 +571,16 @@ impl SpatialView3D {
...
         let mut view_builder = ViewBuilder::new(ctx.render_ctx(), target_config)?;
 
+        if let Ok(Some(entity_path)) = eye_property
+            .component_or_empty::<re_types::components::EntityPath>(
+                &EyeControls3D::descriptor_tracking_entity(),
+            )
+        {
+            state.state_3d.tracked_entity = Some(re_log_types::EntityPath::from_file_path(
+                std::path::Path::new(entity_path.as_str()),
+            ));
```
Wumpf: From file path makes no sense here; that's a utility for creating entity paths from file paths which is not what we have here. Just create the entity path directly from the underlying string?
`re_types_core::EntityPath` should probably have a conversion utility for this to go directly between the two kinds of entity paths

crates/viewer/re_view_spatial/src/ui.rs
```diff
@@ -154,6 +172,22 @@ impl SpatialViewState {
...
+                    ctx,
+                    &ctx.blueprint_query,
+                    ctx.recording(),
+                    ui,
+                    Some(view_id),
+                    &tracked_entity,
+                );
+            } else {
+                ui.label("Tracked entity: None");
+            }
```
Wumpf: `EyeControls3D` already gets automated ui for all its fields, so with that we have this field twice now
image

Wumpf (changes requested):
happy to see that the blueprint driven approach works out! Still some things that have to be fixed up though

crates/viewer/re_view_spatial/src/ui_3d.rs
Gentlegg: Thanks for the help, i fixed all thats left in the latest commit :D

Wumpf (approved):
thanks for all the patience & fixes! 🙇 
I merged in latest main, resolved conflicts and adjusted the name of the context menu action a lil bit to be more clear. Looks all good now I think, will merge once we have green ci on it!

IsseW:
Triggered a workflow to update kittest snapshots

Wumpf:
> Triggered a workflow to update kittest snapshots

doesn't work on non-rerun-team prs because of LFS :(
Updated them manually before but looks like I missed something >.<

IsseW:
> doesn't work on non-rerun-team prs because of LFS :(

oh right :c

PR #11558, opened 2025-10-16
CMake: Download Arrow from Tarball URL Instead of Git

threeal:
### What

Replaces downloading Arrow via `git clone` with downloading it from a tarball URL.
This approach is faster and less bloated than cloning the repository, even when using a shallow clone.


Wumpf:
seems that Cmake's impl for unpacking is platform independent; at least there's no indication otherwise. So landing once C++ builds on latest ci run goes green!

PR #11559, opened 2025-10-16
trim whitespace before/after token

andrea-reale:
### Related

* Closes RR-2562

### What

ignore spaces/newlines accidentally apsted before/after the token


crates/utils/re_auth/src/service/server.rs
```diff
@@ -40,7 +40,8 @@ impl TryFrom<&MetadataValue<Ascii>> for Jwt {
         let token = value.to_str().map_err(|_err| Error::MalformedToken)?;
         let token = token
             .strip_prefix(TOKEN_PREFIX)
```
IsseW, line 42: Maybe we'd want to trim start before `TOKEN_PREFIX` as well?

andrea-reale: `TOKEN_PREFIX` is `Bearer ` and I think it's added automatically by the gRPC middlware (tonic). That's why I am trimming only what remains after the prefix.

PR #11561, opened 2025-10-16
Add test for the table sort icons

emilk:
* Forked out from https://github.com/rerun-io/rerun/pull/11538/files

The current arrow icons are confusing, and we are planning on changing them.

This PR doesn't change them, but adds a snapshot tests for them.

All work by @abey79 - I just extracted it from the original PR.

### DIFF
--- a/crates/viewer/re_dataframe_ui/tests/ascending_descending.rs
+++ b/crates/viewer/re_dataframe_ui/tests/ascending_descending.rs
@@ -21,8 +21,7 @@ async fn test_no_sort() {
 
     let table_status = Arc::new(Mutex::new(None::<TableStatus>));
     let mut harness = test_context
-        .setup_kittest_for_rendering()
-        .with_size([600.0, 400.0])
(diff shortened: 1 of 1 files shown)


crates/viewer/re_chunk_store_ui/src/sort.rs
```diff
@@ -50,12 +51,16 @@ pub(crate) fn sortable_column_header_ui<T: Default + Copy + PartialEq>(
...
-                    _ => "",
-                })
-                .clicked()
+                if is_sorted {
+                    let (icon, text) = match direction {
+                        SortDirection::Ascending => (&re_ui::icons::ARROW_DOWN, "ascending sort"),
+                        SortDirection::Descending => (&re_ui::icons::ARROW_UP, "descending sort"),
+                    };
+
+                    ui.small_icon_button(icon, text).clicked()
```
emilk, line 60: This is a (small) actual change: using icons rather than emojis

PR #11563, opened 2025-10-16
Turn on 4x MSAA for snapshot tests

emilk:
I'm just curious to see how different this is on different platforms… waiting on CI.


emilk:
Ok, this went a lot better than I feared :)

PR #11564, opened 2025-10-16
implement registration by prefix: OSS server

andrea-reale:
### Related
- RR-1962
- dataplatform-side change: https://github.com/rerun-io/dataplatform/pull/1864

### What

This implements, in the OSS server, a variant of `register_with_dataset` that allows data sources to be `prefix`-based paths. E.g. (in python code):

```python
dataset.register_batch(['s3://bucket/my_episodes/'])
```

### What next
This won't be usable by the SDK as is (i.e., the SDK will not let you register by prefix until we add support for it)


andrea-reale:
pretty sure the test failure is unrelated

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -132,6 +132,76 @@ impl RerunCloudHandler {
             })
             .collect::<Result<Vec<_>, _>>()
     }
+
+    fn resolve_data_sources(data_sources: &[DataSource]) -> Result<Vec<DataSource>, tonic::Status> {
+        let mut resolved = Vec::<DataSource>::with_capacity(data_sources.len());
+        for source in data_sources {
+            let path = source.storage_url.path();
+            if path.ends_with('/') {
```
teh-cmc: I would just check that it's a dir directly instead. That way people (including us) won't run into horrible bugs when programmatically generating links to register (because no 2 libraries on earth behave the same when it comes to trailing slashes).

```suggestion
            if path.is_dir() {
```

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -132,6 +132,76 @@ impl RerunCloudHandler {
...
+                    }
+                    _ => tonic::Status::invalid_argument(format!(
+                        "Failed to read directory metadata {path:?}: {err:#}"
+                    )),
+                })?;
+                if !meta.is_dir() {
+                    return Err(tonic::Status::invalid_argument(format!(
+                        "Expected directory, got file: {path:?}"
+                    )));
+                }
```
teh-cmc, line 149: You don't need to go through the metadata APIs and such: https://doc.rust-lang.org/std/path/struct.Path.html#method.is_dir

teh-cmc (approved):

crates/store/re_server/src/rerun_cloud.rs
andrea-reale: There's two things about that:
* primarily, I want the thing to be as intentional as possible: I don't want someone to accidentally register a million RRDs from a given prefix (especially because this is recursive). The trailing `/` is a simple but explicit signal for that.
* secondarily, `is_dir()` is easy on the local file system. It's annoying on an object store because it requires two API calls. This really means that before registration, I need to make between N and 2xN API calls to "resolve" the paths we want to actually register.

That said what about getting the best of both words by:
1. Changing the proto and adding a "dir" flag to the `DataSource` definition, so that the customer needs to be explicit.
2. Dropping the trailing slash thing, so no horrible bugs.

teh-cmc: > That said what about getting the best of both words by:
> 
>     1. Changing the proto and adding a "dir" flag to the `DataSource` definition, so that the customer needs to be explicit.
> 
>     2. Dropping the trailing slash thing, so no horrible bugs.

Yeah that's a decent compromise I think, especially if we don't mark the flag as `optional`, so old clients just do the right thing.

crates/store/re_server/src/rerun_cloud.rs
andrea-reale: uhm.. I don't particularly like that it coerces errors to `false`

PR #11565, opened 2025-10-16
add table writing to lance table providers

timsaucer:
### Related

Addresses part of https://linear.app/rerun/project/write-tables-to-rerun-cloud-via-the-python-sdk-432c0243348f/overview

### What

This PR is foundational to our ability to write to any kind of table on the data platform. Included in this PR is enabling `insert_into` for our gRPC based table providers. This also includes updates to the OSS server to support table writing.

We cannot merge this PR until the following are completed:

- [x] Add identical support to data platform
- [x] Add python integration tests

### DIFF
--- a/crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
+++ b/crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
@@ -152,6 +152,9 @@ service RerunCloudService {
   // Write record batches to a table.
   //
   // This endpoint requires the standard dataset headers.
+  //
+  // TODO(#11645): endpoints with streaming input are not supported by `grpc-web`.
+  // A non-streaming shim will need to be added if/when the viewer uses this endpoint.
   rpc WriteTable(stream WriteTableRequest) returns (WriteTableResponse) {}
 
   // --- Tasks ---
--- a/crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.rs
+++ b/crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.rs
@@ -2228,6 +2228,9 @@ pub mod rerun_cloud_service_client {
         /// Write record batches to a table.
         ///
         /// This endpoint requires the standard dataset headers.
+        ///
+        /// TODO(#11645): endpoints with streaming input are not supported by `grpc-web`.
+        /// A non-streaming shim will need to be added if/when the viewer uses this endpoint.
         pub async fn write_table(
             &mut self,
             request: impl tonic::IntoStreamingRequest<Message = super::WriteTableRequest>,
@@ -2543,6 +2546,9 @@ pub mod rerun_cloud_service_server {
         /// Write record batches to a table.
         ///
         /// This endpoint requires the standard dataset headers.
+        ///
+        /// TODO(#11645): endpoints with streaming input are not supported by `grpc-web`.
+        /// A non-streaming shim will need to be added if/when the viewer uses this endpoint.
         async fn write_table(
             &self,
             request: tonic::Request<tonic::Streaming<super::WriteTableRequest>>,


crates/store/re_datafusion/src/table_entry_provider.rs
```diff
@@ -155,4 +174,265 @@ impl GrpcStreamToTable for TableEntryTableProvider {
...
+
+        // Start the gRPC streaming call immediately
+        runtime.spawn(async move {
+            if let Err(err) = async {
+                let stream = tokio_stream::wrappers::UnboundedReceiverStream::new(rx)
+                    .into_streaming_request()
+                    .with_entry_id(table_id)?;
+                let mut client = client;
+
+                client.inner().write_table(stream).await
```
timsaucer: Before @abey79 says anything I already have it on my radar to push some of this code up to the client interface.

crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
```diff
@@ -589,6 +594,26 @@ message ScanTableResponse {
...
+  TABLE_INSERT_MODE_REPLACE = 2;
+
+  // Overwrites all existing rows in the table with the new rows.
+  TABLE_INSERT_MODE_OVERWRITE = 3;
+}
+
+message WriteTableRequest {
+  rerun.common.v1alpha1.DataframePart dataframe_part = 1;
+  TableInsertMode insert_mode = 2;
+}
```
abey79, line 612: The obviously annoying thing here is that `insert_mode` is repeated for each part in the stream, leading to potential edge cases if they are inconsistent.

I'm unsure how to best address that, and feel ill-equipped to make a call (cc @jleibs @teh-cmc). I can submit the following proposals for consideration:

1) specify that only the _first_ `insert_mode` is considered and all others are ignored (alternatively: error if `insert_mode` is inconsistent—whichever is easier to implement on Cloud side)

2) pass the `insert_mode` as header

3) (least preferred) apply as per `insert_mode` on a per-part basis and properly document here what happens when.

<hr>

Also `TABLE_INSERT_MODE_REPLACE` is way too underspecified for my liking. Like, how is the key(s) specified? Can we remove this option altogether for now?

<hr>

Ultimately, what I'm asking here is to ensure we have large enough buy-in for this new API so we avoid unexpected political debates next time we change a small thing around here.

rerun_py/tests/unit/test_datafusion_tables.py
```diff
@@ -110,9 +131,9 @@ def __init__(self, proc: subprocess.Popen[str], client: CatalogClient, dataset:
 
 
 @pytest.fixture(scope="module")
-def server_instance() -> Generator[ServerInstance, None, None]:
+def server_instance(table_filepath: pathlib.Path) -> Generator[ServerInstance, None, None]:
```
abey79: We are messing with the content of the table in the new test. So should that scope be "test" here, so the server is reloaded clean for each test?

rerun_py/pyproject.toml
```diff
@@ -37,7 +37,7 @@ text = "MIT OR Apache-2.0"
 [project.optional-dependencies]
 tests = ["pytest==8.4.1"]
 notebook = ["rerun-notebook==0.27.0-alpha.1+dev"]
-datafusion = ["datafusion==49.0.0"]
+datafusion = ["datafusion==50.1.0"]
```
abey79, line 40: sneaky :) is it that easy this time?

crates/store/re_redap_client/src/connection_client.rs
```diff
@@ -691,4 +691,11 @@ where
             .into_inner();
         Ok(response)
     }
+
+    pub async fn write_table(
+        &mut self,
+        request: impl tonic::IntoStreamingRequest<Message = WriteTableRequest>,
+    ) -> Result<(), tonic::Status> {
+        self.inner().write_table(request).await.map(|_| ())
+    }
```
abey79, line 717: I think we could improve this one greatly by:
1) I have a proper `ext::TableInsertMode` type with the usual conversion
2) pass it explicitly as standalone argument
3) take an iterator (or a stream, or both version) of `RecordBatches`

crates/store/re_server/src/store/table.rs
```diff
@@ -78,4 +93,106 @@ impl Table {
...
+
+        self.table = TableType::LanceDataset(updated_table);
+        Ok(())
+    }
+
+    pub async fn write_table(
+        &mut self,
+        rb: RecordBatch,
+        insert_op: InsertOp,
+    ) -> Result<(), DataFusionError> {
```
abey79, line 196: For such `pub` methods used by the actual server hander (`rerun_cloud.rs`), I'm trying to use the `crate::Error` type. Motivation: this type tries to inject a high-level task semantic ("I'm trying to write a table") in top of whatever low-level error you get ("obscure metadata does not comply with moon phase").

I think that here this approach applies.

Note that for the reason above, I try to avoid `#[from]` stuff in `crate::Error`, relying on enum variant that explicitly capture the task at hand.

crates/store/re_server/src/store/table.rs
```diff
@@ -78,4 +93,106 @@ impl Table {
...
+
+    async fn write_table_provider(
+        &self,
+        rb: RecordBatch,
+        insert_op: InsertOp,
+    ) -> Result<(), DataFusionError> {
+        let schema = rb.schema();
+        let TableType::DataFusionTable(provider) = &self.table else {
+            return exec_err!("Expected DataFusion Table Provider");
+        };
```
abey79, line 129: possibly too early but might be worth having two sub-structures for lance and dfn tables, to avoid these annoying phantom error cases

crates/store/re_datafusion/src/table_entry_provider.rs
```diff
@@ -154,4 +176,248 @@ impl GrpcStreamToTable for TableEntryTableProvider {
...
+
+        match Pin::new(&mut self.input_stream).poll_next(cx) {
+            Poll::Ready(Some(Ok(batch))) => {
+                // Send to gRPC if we have a sender
+                if let Some(ref grpc_sender) = self.grpc_sender {
+                    let dataframe_part = batch.into();
+                    let request = WriteTableRequest {
+                        dataframe_part: Some(dataframe_part),
+                        insert_mode: self.insert_op.into(),
+                    };
```
abey79: with my proposed `ConnectionClient` change, we'd be moving `RecordBatch`es here, which feels more logical than obscure grpc types.

crates/store/re_server/src/store/table.rs
timsaucer: I was planning to write up an issue to use some kind of trait instead of the enum we have, somewhat like `RerunTableDriver` in redap

crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
timsaucer: From the 3 methods suggested above regarding insert mode, I would lean towards the first - adding an error on inconsistent settings.

Regarding the `TABLE_INSERT_MODE_REPLACE` specificity, that came straight from DF but I'm happy to adjust. @teh-cmc asked a question on the other PR about if we should be doing replace operations. Right now I don't have work set up for adding indices to these tables, which is what I think we would want to be the replacement key.

I think the best path forward right now is to remove the replace option and we can add it later.

rerun_py/pyproject.toml
timsaucer: There was only a change was in Aggregate UDFs and since we aren't using them right now we can do this.

crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
teh-cmc: In this specific scenario, 1 sounds like the best option by a margin imo.

teh-cmc: > Right now I don't have work set up for adding indices to these tables, which is what I think we would want to be the replacement key.

I'm not exactly sure what you meant by that, but just to be clear: upserts can be done without indices are far as Lance is concerned. I don't know about Datafusion.

timsaucer:
@rerun-bot full-check

crates/store/re_redap_tests/src/tests/mod.rs
```diff
@@ -4,6 +4,17 @@ mod entries_table;
...
+mod write_table;
+#[cfg(not(feature = "lance"))]
+mod write_table {
+    // This is a stub test so that we do not have issues with setting
+    // --no-default-options when running tests
+    pub async fn write_table(
+        _service: impl re_protos::cloud::v1alpha1::rerun_cloud_service_server::RerunCloudService,
+    ) {
+    }
+}
```
timsaucer, line 17: I am not happy about this because I can't simply `#[cfg]` it out in the proc macro. I opened https://linear.app/rerun/project/tech-debt-burndown-lance-feature-flag-3de575e29a5f/overview to address.

crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
```diff
@@ -149,6 +149,11 @@ service RerunCloudService {
 
   rpc ScanTable(ScanTableRequest) returns (stream ScanTableResponse) {}
 
+  // Write record batches to a table.
+  //
+  // This endpoint requires the standard dataset headers.
+  rpc WriteTable(stream WriteTableRequest) returns (WriteTableResponse) {}
```
abey79, line 158:
```suggestion
  // Write record batches to a table.
  //
  // This endpoint requires the standard dataset headers.
  //
  // TODO(#11645): endpoints with streaming input are not supported by `grpc-web`. A non-streaming
  // shim will need to be added if/when the viewer uses this endpoint.
  rpc WriteTable(stream WriteTableRequest) returns (WriteTableResponse) {}
```

crates/store/re_redap_tests/src/tests/mod.rs
abey79: I don't think its a good idea to introduce a `"lance"` feature to `re_redap_tests`. Rationale:
- We always run tests with `--all-features`.
- The server without lance is going to be increasingly meaningless as we improve table support.
- That feature is hopefully short-lived anyways (with the protoc issue is resolved upstream)

crates/store/re_redap_tests/Cargo.toml
```diff
@@ -15,7 +15,7 @@ version.workspace = true
 [features]
 
 default = []
-
+lance = ["dep:lance"]
```
abey79, line 18: Again, let's drop this.

crates/store/re_redap_tests/src/utils/tables.rs
```diff
@@ -0,0 +1,59 @@
+use crate::TempPath;
+use arrow::array::{RecordBatch, RecordBatchIterator, record_batch};
+use arrow::datatypes as arrow_schema;
+
+fn create_example_record_batch() -> RecordBatch {
+    record_batch!(
```
abey79, line 6: TIL. I need more of that in my life.

crates/store/re_server/src/store/table.rs
abey79: My experience with trait in that case it that it can cause friction when you want non-dyn-compatible stuff. So I usually traverse the "big enum" -> "sub-structures" -> "Box<dyn Trait>" step by step, advancing only when required. Most recently, I did go all the way to the trait with table filters, but the last step wasn't as nice as I hoped.

Regardless, we can keep that as future work.

crates/store/re_redap_client/src/connection_client.rs
```diff
@@ -691,4 +692,27 @@ where
             .into_inner();
         Ok(response)
     }
+
+    pub async fn write_table(
+        &mut self,
+        stream: impl Stream<Item = RecordBatch> + Send + 'static,
+        table_id: EntryId,
+        insert_mode: TableInsertMode,
+    ) -> Result<(), ApiError> {
```
abey79, line 701: ❤️

crates/store/re_redap_tests/src/tests/mod.rs
timsaucer: I introduced it because we do run tests without all features: It is defined [here](https://github.com/rerun-io/rerun/blob/bbc58b1a5d2544043e0929907ce61caa83993483/.github/workflows/reusable_checks_rust.yml#L135) This check is what made me add the feature flag.

abey79: Ah ok, TIL. Then I'd default opt-in the "lance" feature for re_server tests.

abey79: My less preferred alternative would be:
- make sure we issue a recognisable error on OSS server when "lance" is disabled
- in the test, intercept that error, and early-out the test before it fails
- have an issue and a todo about it

Issue #11556, opened 2025-10-15
Misleading `MemoryLimit` documentation - connecting to 0byte server doesn't switch to recording


ARizzo35:
I was using `serve_grpc_opts()` and noticed this mention about the MemoryLimit option:

```
It is highly recommended that you set the memory limit to 0B if both the server and client are running on the same machine, otherwise you’re potentially doubling your memory usage!
```

However, when I set the MemoryLimit to 0 and sink data to the gRPC server from within the same Rust app, when I open a Viewer which connects to the gRPC server from another machine it just says "Waiting for data". I confirmed setting a MemoryLimit higher than 0 fixes this issue.

Wumpf:
@ARizzo35 thanks for filing an issue, but I don't quite follow what problem you're running into.

> when I open a Viewer which connects to the gRPC server from another machine

You're saying no data ever arrives from the server when you connect remotely?
If the connection comes in with a delay to the actual data that's expected. What the doc tries to say is that having any buffering up of data on the server doesn't make sense if it is immediately forwarded to a local viewer. If there's some delay (because you're not connected yet or you expect to connect several), then this won't work as there's no place to buffer said data.
Can you please clarify the setup in each case?

ARizzo35:
My set up is with two machines, but this case also applies when both the client and server are running on the same machine:

- Machine 1 is running a gRPC server and sinking data to it locally.
- Machine 1 (or 2) is connecting to the gRPC server and running a viewer.

The viewer opens the home page but never actually receives any data. This is all using the latest version (0.25.1).

Below is a minimal example to demonstrate the situation I'm seeing:

```
pub fn main() {
    let rerun_app_id = "test_app";

    let _rec = rerun::RecordingStreamBuilder::new(rerun_app_id).serve_grpc_opts(
        "0.0.0.0",
        9876,
        rerun::ServerOptions {
            playback_behavior: rerun::PlaybackBehavior::OldestFirst,
            memory_limit: rerun::MemoryLimit::from_bytes(0),
        },
    );

    let rec = rerun::RecordingStreamBuilder::new(rerun_app_id)
        .batcher_config(rerun::log::ChunkBatcherConfig::ALWAYS)
        .set_sinks((
            rerun::sink::GrpcSink::default(),
            rerun::sink::FileSink::new(format!("{}.rrd", rerun_app_id)).unwrap(),
        ))
        .unwrap();

    loop {
        std::thread::sleep(std::time::Duration::from_secs(1));
[...]

PR #11568, opened 2025-10-16
Switch to newly received recordings even if `SetStoreInfo` message is missing

Wumpf:
### Related

* Mitigates https://github.com/rerun-io/rerun/issues/11556
    * well, the most immediate pain of it at least

### What

It can happen that the viewer becomes aware of new stores without ever seeing a `SetStoreInfo`. The easiest way to run into this is to connect to a 0byte GRPC server. Which we're actually recommending right now (should we really? is there more to fix here? 🤔)

Either way, it seems sensible that the viewer would do its "oh, ah new store"-procedure independent of `SetStoreInfo`

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -2131,8 +2131,8 @@ impl App {
             );
         }
 
-        let msg_will_add_new_store = matches!(&msg, LogMsg::SetStoreInfo(..))
-            && !store_hub.store_bundle().contains(store_id);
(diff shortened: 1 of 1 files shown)


jprochazk:
I think we should also update docs before closing the linked issue

jprochazk (approved):
Changes like this always make me so nervous, like what are the global consequences of this 😰

Wumpf:
> Changes like this always make me so nervous, like what are the global consequences of this 😰

not wrong. But now's the best time for it. Also it has some coverage by integration tests by now :)

PR #11569, opened 2025-10-16
`re_log_encoding` revamp - playground

teh-cmc:


teh-cmc:
@rerun-bot full-check

PR #11571, opened 2025-10-17
Arrow transform from `BinaryArray` to `List(UInt8Array)`

grtlr:
### Related

* Part of RR-2554

### What

Provides a conversion from `BinaryArray` to `List(UInt8Array)`. Also makes the `DisplayRB` helper more generic.

### DIFF
--- a/crates/utils/re_arrow_util/src/transform.rs
+++ b/crates/utils/re_arrow_util/src/transform.rs
@@ -674,7 +674,7 @@ pub struct BinaryToListUInt8<O1: OffsetSizeTrait, O2: OffsetSizeTrait = O1> {
 }
 
 impl<O1: OffsetSizeTrait, O2: OffsetSizeTrait> BinaryToListUInt8<O1, O2> {
-    /// Create a new transformation to convert a [`BinaryArray`] to a [`ListArray`] of `u8`.
+    /// Create a new transformation to convert a binary array to a list array of `u8` arrays.
     pub fn new() -> Self {
         Default::default()
     }


crates/utils/re_arrow_util/src/transform.rs
```diff
@@ -656,8 +663,67 @@ impl Transform for Flatten {
     }
 }
 
+/// Converts a [`BinaryArray`] to a [`ListArray`] where each binary element becomes a list of `u8`.
```
emilk: Probably worth implementing `LargeBinaryArray` while you're at it?

crates/utils/re_arrow_util/src/transform.rs
```diff
@@ -656,8 +663,67 @@ impl Transform for Flatten {
...
+        offsets.push(current_offset);
+
+        for i in 0..source.len() {
+            if source.is_null(i) {
+                offsets.push(current_offset);
+                continue;
+            }
+
+            let bytes = source.value(i);
+            byte_values.extend_from_slice(bytes);
```
emilk: You shouldn't need to copy the actual contents. All that should be needed is to massage the offset buffer, no?

For reference, here is my zero-copy **inverse** of this function:

```rs

/// `List[List[u8]]` -> `List[Binary]`
fn convert_list_list_u8_to_list_binary(list_array: &dyn Array) -> Option<ListArray> {

    // The outer `List[List[u8]]`
    let list_array = list_array.as_list_opt()?;

    // The inner List[u8] array
    let inner_list_array: &ListArray = list_array.values().as_list_opt()?;

    // The underlying u8 values
    let u8_array: &UInt8Array = inner_list_array.values().as_primitive_opt()?;

    // We consistently use 64-bit offsets for binary data in order to keep our backwards-compatibility checks simpler.
    // Create the binary array reusing existing buffers
    let binary_array = arrow::array::LargeBinaryArray::try_new(
        arrow::buffer::OffsetBuffer::new(
            inner_list_array
                .offsets()
                .iter()
                .map(|&o| o as i64)
                .collect(),
        ),
        u8_array.values().clone().into_inner(),
        inner_list_array.nulls().cloned(),
    )
    .ok()?;

    // Create the outer list array with binary inner type
    let outer_list = ListArray::try_new(
        Arc::new(Field::new("item", DataType::LargeBinary, true)),
        list_array.offsets().clone(),
        Arc::new(binary_array),
        list_array.nulls().cloned(),
    )
    .ok()?;

    debug_assert_eq!(list_array.len(), outer_list.len());

    Some(outer_list)
}
```

crates/utils/re_arrow_util/src/transform.rs
```diff
@@ -70,6 +72,12 @@ pub enum Error {
     #[error("At least one field name is required")]
     NoFieldNames,
 
+    #[error("Offset overflow: total byte count {byte_count} exceeds i32::MAX: {err}")]
+    OffsetOverflow {
```
emilk, line 76: Unused?

crates/utils/re_arrow_util/src/transform.rs
```diff
@@ -958,13 +1003,102 @@ mod test {
...
+    fn test_binary_to_list_uint8() {
+        let builder = arrow::array::BinaryBuilder::new();
+        impl_binary_test(builder);
+    }
+
+    #[test]
+    fn test_large_binary_to_list_uint8() {
+        let builder = arrow::array::LargeBinaryBuilder::new();
+        impl_binary_test(builder);
+    }
```
emilk, line 1160: It would be good to also test that a 2GiB+ `LargeBinaryArray` produces the correct overflow error

crates/utils/re_arrow_util/src/transform.rs
```diff
@@ -958,13 +1003,102 @@ mod test {
...
+
+    #[test]
+    fn test_binary_to_list_uint8() {
+        let builder = arrow::array::BinaryBuilder::new();
+        impl_binary_test(builder);
+    }
+
+    #[test]
+    fn test_large_binary_to_list_uint8() {
```
emilk: This tests LargeBinary to LargeListArray, which is fine, but worth calling out, since our chunk store does not support `LargeListArray`

crates/utils/re_arrow_util/src/transform.rs
emilk: Oh actually, I see we don't support `LargeBinaryArray` -> `ListArray` (yet).

Note that the plan is to use `LargeBinaryArray` for blobs, but still use `ListArray` for chunk store

crates/utils/re_arrow_util/src/transform.rs
```diff
@@ -656,16 +664,55 @@ impl Transform for Flatten {
...
+/// Converts binary arrays to list arrays where each binary element becomes a list of `u8`.
+///
+/// The underlying bytes buffer is reused, making this transformation almost zero-copy.
+#[derive(Clone, Debug, Default)]
+pub struct BinaryToListUInt8<OffsetSize: OffsetSizeTrait> {
+    _phantom: PhantomData<OffsetSize>,
+}
+
+impl<OffsetSize: OffsetSizeTrait> BinaryToListUInt8<OffsetSize> {
+    /// Create a new transformation to convert a [`BinaryArray`] to a [`ListArray`} of `u8`.
```
emilk:
```suggestion
    /// Create a new transformation to convert a [`BinaryArray`] to a [`ListArray`] of `u8`.
```

crates/utils/re_arrow_util/src/transform.rs
grtlr: Ah, that's good to know, I somehow missed that decision.

emilk: `LargeBinaryArray` will allow us 2GiB+ sized blobs, which is important. `ListArray` mostly limits the number of _instances_ in an entity, e.g. max 2B points in a single point cloud which is… fine :)

grtlr:
YOLO merge, because macOS CI timed out again.

PR #11572, opened 2025-10-17
Enable multiple instance poses for `Points3D`

grtlr:
### What

This enables instanced rendering of `Points3D` using `InstancePoses3D`.


crates/store/re_types/definitions/rerun/archetypes/points3d.fbs
```diff
@@ -4,6 +4,8 @@ namespace rerun.archetypes;
 
 /// A 3D point cloud with positions and optional colors, radii, labels, etc.
 ///
+/// If there's more instance poses than positions, the last points' position will be repeated for the remaining poses.
```
Wumpf: not what's happening!

crates/viewer/re_view_spatial/src/visualizers/points3d.rs
```diff
@@ -125,33 +126,37 @@ impl Points3DVisualizer {
...
-                entity_path,
-                &keypoints,
-            )?;
+                let obj_space_bounding_box =
+                    macaw::BoundingBox::from_points(positions.iter().copied());
+                self.data.add_bounding_box(
+                    entity_path.hash(),
+                    obj_space_bounding_box,
+                    world_from_obj,
+                );
```
Wumpf: Let's be perf-nice and compute the bounding box only once and then add it for each transform

crates/viewer/re_view_spatial/tests/transform_clamping.rs
```diff
@@ -132,6 +132,23 @@ pub fn test_transform_clamping() {
...
         });
+
+        test_context.log_entity("points/more_transforms_than_positions", |builder| {
+            builder
+                .with_archetype(
+                    RowId::new(),
+                    TimePoint::default(),
+                    &re_types::archetypes::Points3D::new([(0.0, 5.0, 0.0)])
+                        .with_colors([0x0000FFFF])
+                        .with_radii([-5.]),
```
Wumpf: Let's do a test for several points because the semantics a bit different than on boxes

crates/viewer/re_view_spatial/tests/transform_clamping.rs
```diff
@@ -144,7 +161,7 @@ pub fn test_transform_clamping() {
 }
 
 #[expect(clippy::unwrap_used)]
-fn setup_blueprint(test_context: &mut TestContext) -> (ViewId, ViewId) {
+fn setup_blueprint(test_context: &mut TestContext) -> (ViewId, ViewId, ViewId) {
```
Wumpf, line 167: nit: maybe better to return a list of views at this point?

crates/viewer/re_view_spatial/src/visualizers/points3d.rs
```diff
@@ -95,10 +95,11 @@ impl Points3DVisualizer {
             let colors =
                 process_color_slice(ctx, self, num_instances, &annotation_infos, data.colors);
 
-            let world_from_obj = ent_context
+            for world_from_obj in ent_context
```
Wumpf, line 104: I think this is the right quick fix and should land. But obviously processing all the point data again is pretty horrible, let's add a todo note to at least signal we're aware 😅

Wumpf:
I'd _try_ to make it so that the points take more visual space on the test. They're big enough not to fall under threshhold, but it always makes me nervous when the tested thing is too small

PR #11566, opened 2025-10-16
Update various crates

emilk:
* Part of #11467 

This includes an update to https://crates.io/crates/ply-rs-bw  which promises faster deserialization of .ply files

PR #11573, opened 2025-10-17
Update backwards compatibility files containing random numbers

emilk:
I updated `rand` in <https://github.com/rerun-io/rerun/pull/11566>. This means snippets with random number no longer produced the same result as they used to. It's only a few files though, so I update them in this PR.

## TODO
* [x] `@rerun-bot full-check`


emilk:
@rerun-bot full-check

Issue #6743, opened 2024-07-03
Transform hierarchy is not affected by blueprint overrides


Wumpf:
A prototype exists on the `andreas/transform-overrides` branch, but holding on to this prior to 0.17 since it requires a few rather ugly changes and isn't quite complete yet.

PR #11577, opened 2025-10-17
Fallback provider registry

### LINKED ISSUE #10112: Make fallback provider be globally registered by `ComponentIdentifier`
Implementing fallback providers gets tedious and repetitive quickly - for instance we have lots of fallback providers for `Color` all of them just returning the same entity path based color!
The problem is that currently either visualizers or views provide fallbacks. Now that all components are tagged, it's trivial for us to instead implement a fallback registry that can even support layered fallbacks.

This also allows fusing the "fallback" with the even lower level "placeholder" mechanism which makes up raw arrow data when needed. Thus making a confusing system less confusing and complex!

A draft of such a fallback registry exists on `andreas/improved-fallback-provider`

IsseW:
### Related

- Closes #10112

### What

Adds a fallback provider registry instead of doing it trait based per fallback provider.

Most of the interesting code is in `crates/viewer/re_viewer_context/src/component_fallbacks.rs`, and other stuff is just refactoring to use this.

- [x] Look into moving fallback registration to some (new?) higher crate.
- [x] Possibly a test to make sure all components have fallbacks?

### DIFF
--- a/crates/viewer/re_component_fallbacks/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.blueprint.archetypes.EyeControls3D.snap
+++ b/crates/viewer/re_component_fallbacks/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.blueprint.archetypes.EyeControls3D.snap
@@ -4,3 +4,4 @@ expression: arch_display
 ---
 kind: [2]
 speed: [0.0]
+tracking_entity: []


crates/viewer/re_component_fallbacks/src/component_fallbacks.rs
```diff
@@ -0,0 +1,564 @@
...
+    registry.register_default_type_fallback_provider::<blueprint::components::Corner2D>();
+    registry.register_default_type_fallback_provider::<blueprint::components::BackgroundKind>();
+
+    registry.register_default_type_fallback_provider::<components::Position2D>();
+
+    registry.register_default_type_fallback_provider::<components::Opacity>();
+
+    registry.register_type_fallback_provider::<components::Color>(|ctx| {
+        auto_color_for_entity_path(ctx.target_entity_path)
+    });
```
Wumpf, line 10: soooo good not to do this all over so many places (:

crates/viewer/re_viewer_context/src/component_fallbacks.rs
```diff
@@ -0,0 +1,351 @@
...
+>;
+
+/// A registry to handle component fallbacks.
+///
+/// This has two layers of fallbacks. The first one being for specific [`ComponentDescriptor`]s,
+/// i.e certain fields in archetypes. And the second being for [`ComponentTypes`]s.
+#[derive(Default)]
+pub struct FallbackProviderRegistry {
+    view_component_fallback_providers:
+        HashMap<(std::any::TypeId, ComponentIdentifier), ComponentViewFallbackProviderFn>,
```
Wumpf: haven't read far enough yet, but looking forward to learn more about this one!
given what you've mentioned so far I have a good hunch of why it's needed and what it does, although the typeid surprises me a bit :)

IsseW: > although the typeid surprises me a bit :)

Yeah, I think that maybee it'd work to have it as `ViewId` if we change some stuff around. The things that can be fallback contexts currently, are views and visualizers. There was one spot where `Eye` was used as a fallback provider prior to this, which initially prompted me to use type id. But I got rid of that later.

So maybe visualizers should just return the fallback context of the view they're in?

crates/viewer/re_component_fallbacks/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.archetypes.Tensor.snap
```diff
@@ -0,0 +1,97 @@
...
...: {} }), nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }), (7, Field { name: "I32", data_type: List(Field { name: "item", data_type: Int32, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }), nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }), (8, Field { name: "I64", data_type: List(Field { name: "item", data_type: Int64, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }), nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }), (9, Field { name: "F16", data_type: List(Field { name: "item", data_type: Float16, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }), nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }), (10, Field { name: "F32", data_type: List(Field { name: "item", data_type: Float32, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }), nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }), (11, Field { name: "F64", data_type: List(Field { name: "item", data_type: Float64, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }), nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} })], Dense))
```
IsseW: Should I do some custom display to get rid of these?

IsseW:
This should be ready for review now :) A lot of the added lines is the insta test

crates/viewer/re_component_fallbacks/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+use re_viewer_context::FallbackProviderRegistry;
```
Wumpf, line 18: misses crate level docs (yes they are somewhat redundant to readme)

crates/viewer/re_component_fallbacks/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+
+mod blueprint_component_fallbacks;
+mod component_fallbacks;
+
+/// Creates a new [`FallbackProviderRegistry`] and registers default
```
Wumpf:
```suggestion
/// Creates a new [`FallbackProviderRegistry`] and registers built-in
```

crates/viewer/re_component_fallbacks/src/component_fallbacks.rs
```diff
@@ -0,0 +1,458 @@
...
+    registry.register_type_fallback_provider::<components::HalfSize3D>(|_| [0.5; 3].into());
+    registry.register_type_fallback_provider::<components::Vector2D>(|_| [1.0, 0.0].into());
+    registry.register_type_fallback_provider::<components::Vector3D>(|_| [1.0, 0.0, 0.0].into());
+    registry.register_type_fallback_provider::<components::Timestamp>(|_| {
+        components::Timestamp(0.into())
+    });
+    registry.register_type_fallback_provider::<components::SeriesVisible>(|_| true.into());
+    registry.register_type_fallback_provider::<components::ViewCoordinates>(|_| {
+        archetypes::Pinhole::DEFAULT_CAMERA_XYZ
+    });
```
Wumpf: makes me almost wonder whether there should be a fast-path for constants
(food for thought, out of scope for this PR)

crates/viewer/re_component_fallbacks/src/component_fallbacks.rs
```diff
@@ -0,0 +1,458 @@
...
+        // See https://github.com/rerun-io/rerun/issues/3852
+        resolution_of_image_at(ctx.viewer_ctx(), ctx.query, ctx.target_entity_path)
+            // Zero will be seen as invalid resolution by the visualizer, making it opt out of visualization.
+            // TODO(andreas): We should display a warning about this somewhere.
+            // Since it's not a required component, logging a warning about this might be too noisy.
+            .unwrap_or(components::Resolution::from([0.0, 0.0]))
+    });
+}
+
+pub fn archetype_fallbacks(registry: &mut FallbackProviderRegistry) {
```
Wumpf: they're not really fallbacks for archetypes but fallbacks for archetype fields if we want to contrast them with type fallbacks. So probably `field_fallbacks` or `archetype_field_fallbacks`
Counter point is that we call those fields nowadays often just "components", but by that reasoning we have `component_type_fallbacks` and `component_fallbacks` which is more confusing not less :thinking:

crates/viewer/re_component_fallbacks/src/component_fallbacks.rs
```diff
@@ -0,0 +1,458 @@
...
+
+        // Fallback to empty tensor if we can't determine the values length
+        let tensor_data =
+            datatypes::TensorData::new(vec![0u64], datatypes::TensorBuffer::I64(vec![].into()));
+        components::TensorData(tensor_data)
+    });
+
+    // GraphNodes
+    registry.register_fallback_provider(&archetypes::GraphNodes::descriptor_show_labels(), |_| {
+        components::ShowLabels::from(true)
```
Wumpf: any particular reason you changed style here? In `type_fallbacks` you've went with turbo-fished functions and `.into` whereas here it's no turbo-fish and `::from`. Don't mind either, but would be nice to stay consistent

crates/viewer/re_component_fallbacks/src/blueprint_component_fallbacks.rs
```diff
@@ -0,0 +1,97 @@
...
+use re_viewer_context::FallbackProviderRegistry;
+
+pub fn type_fallbacks(registry: &mut FallbackProviderRegistry) {
+    registry
+        .register_type_fallback_provider::<blueprint::components::ForceDistance>(|_| (60.).into());
+    registry
+        .register_type_fallback_provider::<blueprint::components::ForceStrength>(|_| (1.).into());
+}
+
+pub fn archetype_fallbacks(registry: &mut FallbackProviderRegistry) {
```
Wumpf: see naming comment on `component_fallback.rs`

crates/viewer/re_component_fallbacks/src/blueprint_component_fallbacks.rs
```diff
@@ -0,0 +1,97 @@
...
+    );
+    registry.register_fallback_provider(
+        &blueprint::archetypes::Background::descriptor_kind(),
+        |ctx| match ctx.egui_ctx().theme() {
+            egui::Theme::Dark => blueprint::components::BackgroundKind::GradientDark,
+            egui::Theme::Light => blueprint::components::BackgroundKind::GradientBright,
+        },
+    );
+
+    // PlotBackground
```
Wumpf, line 16: Bit of a pity that we have these centralized here when _most_ are clearly just about a single view. But then again some of them aren't and the types are centralized to begin with. I just wish stuff was more apparently modular 💭. Definitely more than good enough for now, but maybe something we want to iterate on

Wumpf: Reading on I realize that, yes a lot of it _is_ modularized whenever it needs the context of a certain view type.

I think `lib.rs` here should have a brief guide of where to do your registration - when should it go here and when should it go to `on_register`.
Also, I thiiink maybe those graph related ones should all be in the graph view until it turns out they're used in more places? :thinking:

crates/viewer/re_viewer_context/src/component_fallbacks.rs
```diff
@@ -1,32 +1,43 @@
...
 
-use crate::QueryContext;
+use crate::{QueryContext, ViewerContext};
 
-/// Result for a fallback request to a provider.
-pub enum ComponentFallbackProviderResult {
-    /// A fallback value was successfully provided.
-    Value(ArrayRef),
+/// Tries to get a fallback for the type `C`.
+pub fn typed_fallback_for<C: Component>(
```
Wumpf, line 19: Should point out that this isn't a terribly performant method & why, encouraging users to not call it in tight loops

I did look out for that during the review and couldn't spot any places where this would be too hazardous, but it's a thing we have to keep in mind - the previous `typed_fallback_for` had a clear edge there and it _is_ unfortunate that we're have to call this per entity already. Well, surely we can come up with perf optimizations if needed (!) _in the future_ ;-)

crates/viewer/re_component_fallbacks/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+[package.metadata.docs.rs]
+all-features = true
+
+[dependencies]
+re_chunk_store.workspace = true
+re_log_types.workspace = true
+re_types.workspace = true
+re_viewer_context.workspace = true
+
+egui.workspace = true
```
Wumpf: supprised me at first, but makes sense. Let's comment it for the next confused person


```suggestion
egui.workspace = true  # Needed to retrieve properties (colors, strokes etc.) from egui style.
```

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -163,7 +163,6 @@ fn lookup_image_plane_distance(
                 )
                 .get_mono_with_fallback::<ImagePlaneDistance>(
```
Wumpf, line 164: will that one still find the fallback that is registered on Pinhole? Bit confused about that.

crates/viewer/re_view_spatial/src/visualizers/cameras.rs
```diff
@@ -217,6 +217,28 @@ impl VisualizerSystem for CamerasVisualizer {
         filter_visualizable_3d_entities(entities, context)
     }
 
+    fn on_register(&self, mut fallbacks: VisualizerFallbackRegistry<'_>) {
+        fallbacks.register_fallback_provider(&Pinhole::descriptor_image_plane_distance(), |ctx| {
```
Wumpf: related to comment on `TransformTreeContext`: why is this defined on the visualizer and not on the view, what's the guideline there?

crates/viewer/re_view_spatial/src/pinhole.rs
```diff
@@ -118,17 +116,15 @@ pub fn query_pinhole_and_view_coordinates_from_store_without_blueprint(
         .component_mono_quiet::<components::Resolution>(
             archetypes::Pinhole::descriptor_resolution().component,
         )
-        .unwrap_or_else(|| {
-            resolution_of_image_at(ctx, query, entity_path).unwrap_or([100.0, 100.0].into())
-        });
+        .unwrap_or_default();
```
Wumpf: this and below should use fallbacks rather than default otherwise we may get accidental divergence?
That's a problem we have in a bunch of places I believe :/

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -97,6 +98,12 @@ fn ease_out(t: f32) -> f32 {
     1. - (1. - t) * (1. - t)
 }
 
+/// Properties stored in blueprint.
+pub struct BlueprintProperties {
```
Wumpf: very confusingly named - name implies this is "all the blueprint properties" of _something_.
Also unclear why we need this and don't pass around the camera `ViewProperty` instead of lightly wrap over it if needed

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -450,13 +456,25 @@ impl SpatialView3D {
             query.view_id,
         );
 
+        let view_context = self.view_context(ctx, query.view_id, state);
+
+        let speed = **eye_property
+            .component_or_fallback::<LinearSpeed>(&view_context, &EyeControls3D::descriptor_speed())
+            .unwrap_debug_or_log_error() // Should never fail.
+            .unwrap_or(1.0.into());
```
Wumpf: a bit annoying; why do we have to handle that here and not for instance when using `get_mono_with_fallback`? Creates a very strange asymmetry of error handling expectations

crates/viewer/re_view_spatial/src/view_2d.rs
```diff
@@ -63,6 +65,78 @@ impl ViewClass for SpatialView2D {
...
+            VideoStream::descriptor_opacity(),
+            VideoFrameReference::descriptor_opacity(),
+        ] {
+            system_registry.register_fallback_provider(descr, opacity_fallback(ImageKind::Color));
+        }
+
+        system_registry.register_fallback_provider(
+            &SegmentationImage::descriptor_opacity(),
+            opacity_fallback(ImageKind::Segmentation),
+        );
```
Wumpf: none of these are exclusive to the 2D view. Unclear why they're here (and not in the respective visualizers)

ARCHITECTURE.md
```diff
@@ -132,29 +132,30 @@ Update instructions:
...
-| re_viewport           | The central viewport panel of the Rerun viewer.                                                            |
+| Crate                  | Description                                                                                                |
+| ---------------------- | ---------------------------------------------------------------------------------------------------------- |
+| re_arrow_ui            | Show arrow data in a tree of rerun list_items and format arrow with syntax highlighting.                   |
+| re_blueprint_tree      | The UI for the blueprint tree in the left panel.                                                           |
+| re_redap_browser       | The UI and communication to implement the in-viewer redap server browser.                                  |
+| re_chunk_store_ui      | A chunk store browser UI.                                                                                  |
+| re_component_fallbacks | Provides default fallbacks for Rerun components in the fallback provider registry.                         |
```
Wumpf, line 141: you also need to update the diagram picture

crates/viewer/re_viewport_blueprint/src/view_properties.rs
```diff
@@ -91,15 +91,12 @@ impl ViewProperty {
     }
 
     /// Get the value of a specific component or its fallback if the component is not present.
-    // TODO(andreas): Unfortunately we can't use TypedComponentFallbackProvider here because it may not be implemented for all components of interest.
-    // This sadly means that there's a bit of unnecessary back and forth between arrow array and untyped that could be avoided otherwise.
     pub fn component_or_fallback<C: re_types::Component>(
```
Wumpf, line 94: ironic that the todo comment was resolved by not caring about that back and forth of serialization :/

crates/viewer/re_view_tensor/src/view_class.rs
```diff
@@ -727,16 +715,6 @@ fn selectors_ui(
     }
 }
 
-impl TypedComponentFallbackProvider<Colormap> for TensorView {
-    fn fallback_for(&self, _ctx: &re_viewer_context::QueryContext<'_>) -> Colormap {
-        // Viridis is a better fallback than Turbo for arbitrary tensors.
-        Colormap::Viridis
```
Wumpf, line 731: I think this one got lost - specifically in the tenser view we have a different fallback for colormaps than elsewhere

crates/viewer/re_component_fallbacks/src/component_fallbacks.rs
```diff
@@ -0,0 +1,458 @@
...
+        .map_or(0, |array| array.len());
+    let num_labels = results
+        .component_batch_raw(text_component.component)
+        .map_or(0, |array| array.len());
+
+    components::ShowLabels::from(num_labels == 1 || num_instances < MAX_NUM_LABELS_PER_ENTITY)
+}
+
+/// Get a valid, finite range for the gpu to use.
+fn tensor_data_range_heuristic(
```
Wumpf, line 407: re_viewer_context has near identical `image_data_range_heuristic` method. Maybe this is the right time to merge those two!

crates/viewer/re_viewer/src/app.rs
```diff
@@ -565,14 +572,11 @@ impl App {
     }
 
     /// Adds a new view class to the viewer.
-    #[deprecated(
-        since = "0.24.0",
-        note = "Use `view_class_registry().add_class(..)` instead."
-    )]
     pub fn add_view_class<T: ViewClass + Default + 'static>(
```
Wumpf, line 607: [likely not actionable] not sure how I feel about the un-deprecation. I wanted to keep the interface of `app.rs` leaner. But I reckon this makes this method too useful again :/

crates/viewer/re_viewer_context/src/viewer_context.rs
```diff
@@ -86,11 +89,8 @@ impl ViewerContext<'_> {
 
     /// Runtime info about components and archetypes.
     ///
-    /// The component placeholder values for components are to be used when [`crate::ComponentFallbackProvider::try_provide_fallback`]
-    /// is not able to provide a value.
-    ///
-    /// ⚠️ In almost all cases you should not use this directly, but instead use the currently best fitting
-    /// [`crate::ComponentFallbackProvider`] and call [`crate::ComponentFallbackProvider::fallback_for`] instead.
+    /// ⚠️ In almost all cases you should not use this directly, but instead use
+    /// [`crate::FallbackProviderRegistry::fallback_for`].
```
Wumpf: guess we can remove that entirely since by now the reflection is used for all sort of things not just getting defaults

crates/viewer/re_viewer_context/src/global_context.rs
```diff
@@ -16,11 +16,8 @@ pub struct GlobalContext<'a> {
 
     /// Runtime info about components and archetypes.
     ///
-    /// The component placeholder values for components are to be used when `ComponentFallbackProvider::try_provide_fallback`
-    /// is not able to provide a value.
-    ///
-    /// ⚠️ In almost all cases you should not use this directly, but instead use the currently best fitting
-    /// `ComponentFallbackProvider` and call `ComponentFallbackProvider::fallback_for` instead.
+    /// ⚠️ In almost all cases you should not use this directly, but instead use
+    /// [`crate::FallbackProviderRegistry::fallback_for`].
```
Wumpf: as on viewer_context: that comment lost its meaningfulness since Reflection is handling so much more than placeholder

crates/viewer/re_viewer_context/src/component_fallbacks.rs
```diff
@@ -1,32 +1,43 @@
...
-    ///
-    /// Unlike [`ComponentFallbackProviderResult::ComponentNotHandled`], this is treated as an unexpected error.
-    SerializationError(re_types::SerializationError),
-}
+    let Some(v) = C::from_arrow(&array)
+        .ok()
+        .and_then(|v| v.into_iter().next())
+    else {
+        panic!(
+            "Missing fallback provider for `{}`",
```
Wumpf: incorrect message: if we get here, we failed deserializing. We did in fact have a fallback provider (well or not and went with "placeholder", not that it matterns ;-))

crates/viewer/re_viewer_context/src/component_fallbacks.rs
```diff
@@ -37,87 +48,241 @@ pub enum ComponentFallbackError {
...
+/// A registry to handle component fallbacks.
 ///
-/// Fallbacks can be based on arbitrarily complex & context sensitive heuristics.
-pub trait ComponentFallbackProvider {
-    /// Tries to provide a fallback value for a given component.
-    ///
-    /// If the provider can't handle the component or simply want to use a placeholder value,
-    /// it should return [`ComponentFallbackProviderResult::ComponentNotHandled`].
+/// This has 5 layers of fallbacks:
+/// - First try to use a fallback for the view context and [`ComponentIdentifier`].
```
Wumpf: changed a bit / can be more accurate:

```suggestion
/// - First try to use a fallback using the active type of view (identified via [`ViewClassIdentifier`]) and [`ComponentIdentifier`].
```

crates/viewer/re_viewer_context/src/component_fallbacks.rs
```diff
@@ -37,87 +48,241 @@ pub enum ComponentFallbackError {
...
+/// - And finally we try to give some sensible value based on the arrow type.
+///
+/// The first 3 of those fallbacks are registered to this registry.
+///
+/// For [`ComponentIdentifier`] and [`ComponentType`] specific that is usually done in the
+/// `re_component_fallbacks` crate. For view specific fallbacks that is usually done in
+/// the views or visualizers `on_register` function.
+#[derive(Default)]
+pub struct FallbackProviderRegistry {
+    view_component_fallback_providers:
```
Wumpf, line 66: can tell which one you added by this one not having a comment :P

crates/viewer/re_viewer_context/src/component_fallbacks.rs
```diff
@@ -37,87 +48,241 @@ pub enum ComponentFallbackError {
...
+        HashMap<(ViewClassIdentifier, ComponentIdentifier), ComponentFallbackProviderFn>,
+
+    /// Maps component identifier to fallback providers.
+    component_fallback_providers: IntMap<ComponentIdentifier, ComponentFallbackProviderFn>,
+
+    /// Maps component types to fallback descriptors, used if there's no matching
+    /// fallback in `exact_fallback_providers`.
```
Wumpf: outdated comment - what is a fallback descriptor, `exact_fallback_providers` doesn't exist

crates/viewer/re_viewer_context/src/component_fallbacks.rs
```diff
@@ -37,87 +48,241 @@ pub enum ComponentFallbackError {
...
+
+        self.register_dyn_fallback_provider(
+            descriptor.component,
+            Box::new(move |query_context| {
+                let value = provider(query_context);
+
+                C::to_arrow([Cow::Owned(value)])
+            }),
+        );
+    }
```
Wumpf: naming of this one is a bit weird. I mean isn't it the exact same as `register_type_fallback_provider` only that it takes a descriptor?
I think if possible we should just kill that one off since it's confusing - we're not actually indexing on anytyhing else but the component identifier

crates/viewer/re_viewer_context/src/component_fallbacks.rs
```diff
@@ -37,87 +48,241 @@ pub enum ComponentFallbackError {
...
+
+    /// Registers a fallback provider function for a given component identifier
+    /// in a specific view.
```
Wumpf:
```suggestion
    /// in a specific view class type.
```

crates/viewer/re_viewer_context/src/component_fallbacks.rs
```diff
@@ -37,87 +48,241 @@ pub enum ComponentFallbackError {
...
+            );
+        }
+    }
+
+    /// Registers a fallback provider function for a given component identifier
+    /// in a specific view.
+    pub fn register_view_fallback_provider<C: re_types::Component>(
+        &mut self,
+        view: ViewClassIdentifier,
+        descriptor: &ComponentDescriptor,
```
Wumpf: as above, let's not take descriptors when we don't index over them to avoid confusion
Yes, it's unfortunate that we can't assert on the type anymore then but I had to remove that on various type converting queries as well recently; no turning back and I think it's the lesser evil

crates/viewer/re_viewer_context/src/component_fallbacks.rs
```diff
@@ -37,87 +48,241 @@ pub enum ComponentFallbackError {
...
-                        return  $crate::TypedComponentFallbackProvider::<$component>::fallback_for(self, _ctx).into();
-                    }
-                )*
-                $crate::ComponentFallbackProviderResult::ComponentNotHandled
-            }
+/// This requires the component type to be known by either datastore or blueprint store and
+/// will return a placeholder for a nulltype otherwise, logging an error.
+/// The rationale is that to get into this situation, we need to know of a component type for which
+/// we don't have a datatype, meaning that we can't make any statement about what data this component should represent.
+// TODO(andreas): Are there cases where this is expected and how to handle this?
```
Wumpf: nope, pretty sure by now there really isn't :)

```suggestion
```

crates/viewer/re_viewer_context/src/view/visualizer_system.rs
```diff
@@ -90,7 +89,7 @@ impl VisualizerQueryInfo {
 /// Is populated after scene contexts and has access to them.
 ///
 /// All visualizers are expected to be able to provide a fallback value for any component they're using
-/// via the [`ComponentFallbackProvider`] trait.
+/// by registering fallbacks to the [`crate::FallbackProviderRegistry`] in the `on_register` method.
```
Wumpf: not quite true! They obviously don't register a fallback for any component they're using!
Since it's somewhat rare I'd just remove that section here and expand below.

crates/viewer/re_viewer_context/src/view/visualizer_system.rs
```diff
@@ -119,6 +118,10 @@ pub trait VisualizerSystem: Send + Sync + 'static {
         None
     }
 
+    /// Called just as this visualizer is registered.
```
Wumpf:
```suggestion
    /// Called just as this visualizer is registered.
    ///
    /// This can be useful to register additional component fallbacks that are specific to this visualizer's archetype.
```

or something like that

crates/viewer/re_viewer_context/src/view/view_class_registry.rs
```diff
@@ -32,9 +33,32 @@ pub enum ViewClassRegistryError {
     UnknownClassIdentifier(ViewClassIdentifier),
 }
 
+/// Utility which is passed to visualizer's `on_register`, to register fallbacks.
+pub struct VisualizerFallbackRegistry<'a> {
```
Wumpf: Let's call this `ViewClassFallbackRegistry` to make clear that the fallbacks are _not_ local to visualizers but rather to view class types.

neat mechanic as such, like it.

crates/viewer/re_component_fallbacks/tests/all_component_fallbacks.rs
```diff
@@ -0,0 +1,55 @@
+use re_log_types::EntityPath;
+use re_test_context::TestContext;
+use re_types::{
+    ViewClassIdentifier,
+    external::arrow::util::display::{ArrayFormatter, FormatOptions},
+};
+use re_viewer_context::{BlueprintId, DataQueryResult, QueryContext, ViewContext};
+
+#[test]
```
Wumpf, line 9: suprised you sticked it in here after all and not on `re_viewer` - I thought we wanted to have snapshots also for all views and have access to anything that's registered "late"?

crates/viewer/re_component_fallbacks/tests/all_component_fallbacks.rs
```diff
@@ -0,0 +1,55 @@
...
+
+                let values = (0..res.len())
+                    .map(|i| formatter.value(i).to_string())
+                    .collect::<Vec<_>>();
+
+                arch_display.push((field.name, values));
+            }
+
+            let name = format!("arch_fallback_{arch_name}");
+            insta::assert_debug_snapshot!(name, arch_display);
```
Wumpf: only if it's not too much effort, could we do more compact rendering of the arrays? newlines on the brackets makes it quite long right now :/
Haven't tried, maybe `assert_compact_debug_snapshot` does it? (probably has too little newlines then...)

Wumpf (approved):
awesome 🤩 

I think there's a few things to refine though a little bit before landing. The main concern I have right now is that it's unclear where new fallback callbacks should go when. Might be a pure doc issue or something we want to adjust more generally, happy to chat about it again if needed!

Wumpf (changes requested):
(eh sorry, I wanted to press 'request changes' because some of the stuff I'm curious how it ends up and I feel somewhat strongly about the "where do things go going forward" thing I've mentioned)

crates/viewer/re_component_fallbacks/src/component_fallbacks.rs
IsseW: Went for `archetype_field_fallbacks`

crates/viewer/re_component_fallbacks/tests/all_component_fallbacks.rs
IsseW: Currently views only have access to register view-specific fallbacks. So those wouldn't affect this test.

I think you mentioned doing the all views test as a follow-up? And I imagine that'd be separate from this

crates/viewer/re_view_spatial/src/view_2d.rs
IsseW: Yeah having them in their visualizers make more sense 👍

crates/viewer/re_view_spatial/src/pinhole.rs
IsseW: Don't think I can really use fallbacks here since that requires a data result. Can simplify the method a bit, and use `resolution_of_image_at` again though.

Related to this TODO:
> // TODO(#6743): We don't have a data-result or the other pieces
> // necessary to properly handle overrides, defaults, or fallbacks.

crates/viewer/re_view_spatial/src/ui_3d.rs
IsseW: `get_mono_with_fallback` requires the component to implement default and does a `unwrap_or_default` internally. I don't think that's something we can really do for blueprint properties. The failure case is if it's an empty array or deserialization error. But yeah it is annoying to have to handle like this. Maybe it's fine to do `?` here?

crates/viewer/re_view_spatial/src/ui_3d.rs
IsseW: In this case it's because `state` needs to be borrowed mutably for updating the eye, and the `view_context` borrows the state immutable which we need for fallbacks.

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
IsseW: Ahh I think this is something that broke when I removed fallback providers. Before this would explicitly pass a `CameraVisualizer` as a provider for this.

I thiink the view class is the best place to register this still then. But I think we have to register it for both 2d and 3d then?

crates/viewer/re_view_tensor/src/view_class.rs
IsseW: It is still on the `TensorScalarMapping::descriptor_colormap` in `blueprint_component_fallbacks.rs`, but I think with my new guidelines this should prefer being registered in the view

crates/viewer/re_viewer/src/app.rs
IsseW: I also felt some friction making the change, but it wasn't possible to call `add_view_class` without it since you'd have to call two methods on app that take `&mut self` to get both registries.

crates/viewer/re_viewer_context/src/component_fallbacks.rs
IsseW: oops 😄

crates/viewer/re_component_fallbacks/tests/all_component_fallbacks.rs
IsseW: Did this with some custom formatting

crates/viewer/re_viewer_context/src/view/view_class_registry.rs
IsseW: This was removed upon further discussion

crates/viewer/re_component_fallbacks/src/component_fallbacks.rs
IsseW: Yeah that would be nice, not sure what the best way to handle that in the future would be though

crates/viewer/re_component_fallbacks/src/component_fallbacks.rs
IsseW: Made the style consistent

crates/viewer/re_view_spatial/src/pinhole.rs
IsseW: > That's a problem we have in a bunch of places I believe :/

These other places, are those the transforms?

crates/viewer/re_view_spatial/src/ui_3d.rs
IsseW: Did rename it though

crates/viewer/re_viewport_blueprint/src/view_properties.rs
IsseW: Yeahh 😭

crates/viewer/re_component_fallbacks/src/component_fallbacks.rs
IsseW: Did

crates/viewer/re_view_spatial/src/pinhole.rs
Wumpf: transforms are extra weird. But I believe other visualizers are often just using some default, trusting that it's the same as the default
(not a concern for this PR)

crates/viewer/re_view_spatial/src/ui_3d.rs
Wumpf: probably.... I'd say we let it slide though now in order to get this in :)

PR #11582, opened 2025-10-17
`re_log_encoding` v2

teh-cmc:
A complete rewrite of `re_log_encoding` that pays off the organically grown debt and sprinkles a couple years of hindsight on top. This also closes the "one encoder/decoder to rule them all" story.

Apologies, I really wanted to split this into multiple PRs, but everything I've tried turned into a dependency nightmare.
All the code in `re_log_encoding` is new, so you can just ignore the diff entirely for that folder and look directly at the code, really.
The docs are shit, but there is so much hidden complexity and so much context required, I'm never going to get them right on first try anyway. Suggestions welcome.

I've put as reviewers all the people that got a live walkthrough.

For everyone else, here's an excerpt from the docs of the new `re_log_encoding` to get an idea of what's going on:
> This crate covers two equally important but orthogonal matters:
> * Converting between transport-level and application-level Rerun types.
> * Encoding and decoding Rerun RRD streams.
>
> If you are working with one of the gRPC APIs (Redap or SDK comms), then you want to be looking
> at the [`ToTransport`]/[`ToApplication`] traits. The [`rrd`] module is completely irrelevant in
> that case. You can learn more about these traits below.
>
> If you are working with actual RRD streams (i.e. everything that does not go through gRPC:
> files, standard I/O, HTTP, data loaders, etc), then have a look into the [`rrd`] module.
> The [`ToTransport`]/[`ToApplication`] traits will also be useful to you. You can learn more
> about these traits below.
>
> ## What are transport-level and application-level types?
>
> To put it in simple terms: transport-level types are the types that you find in `re_protos`, while
> application-level types are those that you find in `re_log_types`.
>
> More generally, transport-level types are Rust objects that represent the decoded value of some
> Rerun bytes, *and nothing more than that*. It's all they do: they map raw bytes to their native
> Rust representation and vice-versa. They never apply any application-level logic beyond that.
>
> Transport-level types are used to unopinionatedly transport Rerun data across the many mediums
> that Rerun support, while application-level types are used to build applications for end-users,
> such as the Rerun viewer itself.
>
> Application-level types on the other hand are *very* opinionated, and often perform a lot of
> transformations on the data, including but not limited to:
> * Chunk/Sorbet migrations
> * Application ID injection
> * SDK version patching
> * Backward-compatibility shenanigans
> * Etc
>
> Application-level can _not_ be encoded/decoded, only transport-level types can. To encode an
> application-level type, you must first convert it to transport-level type.
> You can do so by using the [`ToApplication`] & [`ToTransport`] traits exposed by this crate.
>
> ## How do I make sense of all these different `LogMsg` types?!
>
> There are 3 different `LogMsg`-related types that you will very often encounter: `re_log_types::LogMsg`,
> `re_protos::log_msg::v1alpha1::LogMsg` and `re_protos::log_msg::v1alpha1::log_msg::Msg`.
>
> Mixing them up is a common source of pain and confusion, so let's go over what each does:
> * `re_log_types::LogMsg` is the application-level type that we use all across the viewer
>   codebase. It can be obtained by calling `to_application()` on one of the transport-level
>   `LogMsg` types which, among many other things, will perform Chunk/Sorbet-level migrations.
>   `re_log_types::LogMsg` isn't used in Redap, where everything is done at the transport-level, always.
> * `re_protos::log_msg::v1alpha1::LogMsg` is the transport-level definition of `LogMsg`. It is an
>   artifact of how `oneof` works in Protobuf: all it does is carry a `re_protos::log_msg::v1alpha1::log_msg::Msg`.
>   For that reason, it is never directly used, except by the legacy SDK comms protocol.
> * Finally, `re_protos::log_msg::v1alpha1::log_msg::Msg` is the real transport-level type that we
>   care about. It is used all over the place when encoding and decoding RRD streams.
>
> ## What are the different protocols supported by Rerun?
>
> Rerun currently supports 3 protocols:
> * Redap (Rerun Data Protocol): our gRPC-based protocol used by our OSS and proprietary data platforms.
> * SDK comms: our legacy gRPC-based protocol, currently used by everything relying on the old
>   `StoreHub` model (logging, message proxy, etc).
> * RRD streams: the binary protocol that we use for all stream-based interfaces (files, stdio,
>   data-loaders, HTTP fetches, etc).
>
> *All these protocols use the exact same encoding*. There is only one encoding: the Rerun encoding.
> It often happens that one protocol makes use of some types while others don't (e.g. the
> top-level `LogMsg` object is never used in RRD streams, but is used in SDK comms), but for all
> the types they do share, the encoding will be the exact same.

* Sibling: https://github.com/rerun-io/dataplatform/pull/1873

### DIFF
--- a/crates/store/re_log_encoding/src/rrd/decoder/mod.rs
+++ b/crates/store/re_log_encoding/src/rrd/decoder/mod.rs
@@ -37,7 +37,7 @@ use crate::rrd::Decodable as _;
 
 /// Implemented for top-level types that can kickoff decoding.
 ///
-/// There are only two of them:
+/// There are only two of them in this crate:
 /// * [`re_log_types::LogMsg`]: application-level root message
 /// * [`re_protos::log_msg::v1alpha1::log_msg::Msg`]: transport-level root message
 ///
@@ -45,8 +45,6 @@ use crate::rrd::Decodable as _;
 /// See also:
 /// * [`DecoderTransport`]
 /// * [`DecoderApp`]
-///
-// TODO(cmc): technically this should be a sealed trait, but things are complicated enough as is.
 pub trait DecoderEntrypoint: Sized {
     fn decode(
         data_excluding_headers: bytes::Bytes,
--- a/crates/store/re_log_encoding/src/rrd/encoder.rs
+++ b/crates/store/re_log_encoding/src/rrd/encoder.rs
@@ -99,12 +99,20 @@ impl Encoder<Vec<u8>> {
 }
 
 impl<W: std::io::Write> Encoder<W> {
+    /// Creates a new [`Encoder`].
+    ///
+    /// This will immediately write the [`StreamHeader`] to the output stream as part of
+    /// initialization (hence `_eager`).
+    ///
+    /// There is no `_lazy` version. Make one if you need one.
     pub fn new_eager(
         version: CrateVersion,
         options: EncodingOptions,
         mut write: W,
     ) -> Result<Self, EncodeError> {
-        // TODO(cmc): the extra heap-alloc and copy could be easily avoided.
+        // TODO(cmc): the extra heap-alloc and copy could be easily avoided with the
+        // introduction of an InMemoryWriter trait or similar. In practice it makes no
+        // difference and the cognitive overhead of this crate is already through the roof.
         let mut out = Vec::new();
         StreamHeader {
             fourcc: crate::rrd::RRD_FOURCC,
@@ -136,7 +144,7 @@ impl<W: std::io::Write> Encoder<W> {
         re_tracing::profile_function!();
 
         let message = message.to_transport(self.compression)?;
-        // Safety: we _are_ the encoder.
+        // Safety: the compression settings of this message are consistent with this stream.
         #[expect(unsafe_code)]
         unsafe {
             self.append_transport(&message)
@@ -194,7 +202,9 @@ impl<W: std::io::Write> Encoder<W> {
 
         match self.serializer {
             Serializer::Protobuf => {
-                // TODO(cmc): the extra heap-alloc and copy could be easily avoided.
+                // TODO(cmc): the extra heap-alloc and copy could be easily avoided with the
+                // introduction of an InMemoryWriter trait or similar. In practice it makes no
+                // difference and the cognitive overhead of this crate is already through the roof.
                 let mut header = Vec::new();
                 MessageHeader {
                     kind: MessageKind::End,
--- a/crates/store/re_log_encoding/src/rrd/headers.rs
+++ b/crates/store/re_log_encoding/src/rrd/headers.rs
@@ -139,7 +139,9 @@ impl Encodable for StreamHeader {
         out.extend_from_slice(fourcc);
         out.extend_from_slice(version);
         {
-            // TODO(cmc): the extra heap-alloc and copy could be easily avoided.
+            // TODO(cmc): the extra heap-alloc and copy could be easily avoided with the
+            // introduction of an InMemoryWriter trait or similar. In practice it makes no
+            // difference and the cognitive overhead of this crate is already through the roof.
             let mut options_out = Vec::new();
             options.to_rrd_bytes(&mut options_out)?;
             out.extend_from_slice(&options_out);
--- a/crates/store/re_log_encoding/src/rrd/mod.rs
+++ b/crates/store/re_log_encoding/src/rrd/mod.rs
@@ -1,6 +1,6 @@
 //! Everything needed to encode/decode and serialize/deserialize RRD streams.
 //!
-//! ⚠️Make sure to familiarize yourself with the [crate-level docs] first. ⚠️
+//! ⚠️Make sure to familiarize yourself with the [crate-level docs](crate) first. ⚠️
 //!
 //! RRD streams are used everywhere gRPC isn't: files, standard I/O, HTTP fetches, data-loaders, etc.
 //! This module is completely unrelated to the Rerun Data Protocol (Redap) gRPC API.
@@ -14,8 +14,6 @@
 //! That is all these traits do. They do not perform any kind of IO, they do not keep track of any
 //! sort of state. That's the job of the `Encoder` and `Decoder`: they provide the IO and the
 //! state machines that turn collections of `Encodable`s and `Decodable`s into actual RRD streams.
-//!
-//! [crate-level docs]: [`crate`]
 
 mod errors;
 mod headers;
--- a/crates/store/re_log_encoding/src/transport_to_app.rs
+++ b/crates/store/re_log_encoding/src/transport_to_app.rs
@@ -6,7 +6,7 @@
 //! happens: Chunk/Sorbet migrations, data patching (app ID injection, version propagation, BW-compat hacks,
 //! etc).
 //!
-//! To go from a from a freshly decoded transport-level type to its application-level equivalent, use [`ToApplication`].
+//! To go from a freshly decoded transport-level type to its application-level equivalent, use [`ToApplication`].
 //! To prepare an application-level type for encoding, use [`ToTransport`].
 
 use re_build_info::CrateVersion;


teh-cmc:
@rerun-bot full-check

crates/store/re_log_encoding/src/rrd/decoder/iterator.rs
teh-cmc, line 1: This file implements a poll-based synchronous `Iterator` on top of the shared decoding core. It's nothing new, it's like 10 days old already!

crates/store/re_log_encoding/src/rrd/decoder/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+
+/// On failure to decode or serialize a `LogMsg`.
+#[derive(thiserror::Error, Debug)]
+pub enum DecodeError {
+    #[error("Codec error: {0}")]
+    Codec(#[from] crate::rrd::CodecError),
+
+    #[error("Failed to read: {0}")]
+    Read(#[from] std::io::Error),
+}
```
teh-cmc, line 22: Note how non-insane this error looks.

crates/store/re_log_encoding/src/rrd/decoder/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+/// * [`re_log_types::LogMsg`]: application-level root message
+/// * [`re_protos::log_msg::v1alpha1::log_msg::Msg`]: transport-level root message
+///
+/// This can be used to generically instantiate transport- and/or application-level decoders.
+/// See also:
+/// * [`DecoderTransport`]
+/// * [`DecoderApp`]
+///
+// TODO(cmc): technically this should be a sealed trait, but things are complicated enough as is.
+pub trait DecoderEntrypoint: Sized {
```
teh-cmc, line 48: I should really talk more about this thing in the crate-level docs.

crates/store/re_log_encoding/src/rrd/decoder/state_machine.rs
teh-cmc, line 1: Nothing new here, our good old push-based decoding state machine.

crates/store/re_log_encoding/src/rrd/decoder/stream.rs
teh-cmc, line 1: This file implements a poll-based asynchronous `Stream` on top of the shared decoding core. That is very new, and is how we achieve a single encoding/decoding stack across both OSS and Redap.

This is basically the exact same code as decoder/iterator.rs, just using the async IO traits instead.

crates/store/re_protos/src/v1alpha1/rerun.common.v1alpha1.ext.rs
teh-cmc, line 1: That's https://github.com/rerun-io/rerun/pull/11537 and https://github.com/rerun-io/rerun/pull/11536.

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -414,11 +413,7 @@ async fn chunk_stream_io_loop(
     chunk_infos: Vec<RecordBatch>,
     output_channel: Sender<Result<ChunksWithPartition, re_redap_client::ApiError>>,
 ) -> Result<(), DataFusionError> {
-    let chunk_infos = chunk_infos
-        .into_iter()
-        .map(|batch| batch.encode())
-        .collect::<Result<Vec<_>, _>>()
-        .map_err(|err| exec_datafusion_err!("{err}"))?;
+    let chunk_infos: Vec<_> = chunk_infos.into_iter().map(Into::into).collect();
```
abey79, line 416: small but neat improvement, thx

crates/store/re_protos/src/v1alpha1/rerun.common.v1alpha1.ext.rs
```diff
@@ -747,6 +747,162 @@ impl From<crate::common::v1alpha1::semantic_version::Meta> for re_build_info::Me
     }
 }
 
+// --- DataframePart / RerunChunk <-> RecordBatch ---
+
+impl TryFrom<crate::common::v1alpha1::RerunChunk> for arrow::array::RecordBatch {
```
abey79, line 752: Beside the point of this PR, but I find it slightly weird that we have a specific grpc type for `RerunChunk` that is strictly equal to `DataframePart`. We have half a dozen and counting usage of `DataframePart` for very specific tables, but for one of them we jump through the hoop of essentially aliasing it at the grpc level. Why this one only?

Also, I'm not sure there is any type-checker benefit of it. Case in point: it goes through an arrow `RecordBatch` like all other dataframe.

crates/store/re_log_encoding/src/rrd/mod.rs
```diff
@@ -0,0 +1,118 @@
+//! Everything needed to encode/decode and serialize/deserialize RRD streams.
+//!
+//! ⚠️Make sure to familiarize yourself with the [crate-level docs] first. ⚠️
```
abey79: malformed docstring link?

crates/store/re_log_encoding/src/rrd/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+///
+/// All it does is map transport types to their RRD representation. If you're wondering how to turn
+/// application-level types into transport-level types for encoding, have a look at the
+/// [`ToTransport`] trait.
+///
+/// The only way this can fail is due to invalid data.
+///
+/// [`ToTransport`]: crate::ToTransport
+//
+// TODO(cmc): technically this should be a sealed trait, but things are complicated enough as is.
```
abey79, line 85: i mean, isn't that like 4 additional lines here?

crates/store/re_log_encoding/src/rrd/errors.rs
```diff
@@ -111,3 +115,22 @@ impl std::fmt::Display for NotAnRrdError {
...
+/// On failure to decode [`crate::rrd::EncodingOptions`]
+#[derive(thiserror::Error, Debug)]
+pub enum OptionsError {
+    #[error("Reserved bytes not zero")]
+    UnknownReservedBytes,
+
+    #[error("Unknown compression: {0}")]
+    UnknownCompression(u8),
+
+    // TODO(jan): Remove this at some point, realistically 1-2 releases after 0.23
```
abey79, line 128: :)

crates/store/re_log_encoding/src/rrd/encoder.rs
```diff
@@ -106,17 +99,20 @@ impl Encoder<Vec<u8>> {
 }
 
 impl<W: std::io::Write> Encoder<W> {
-    pub fn new(
+    pub fn new_eager(
```
abey79, line 108: worth adding a docstring to hint at why you would want the `eager` version and what you'd do if you didn't. Or what this is not called `new` to being with.

crates/store/re_log_encoding/src/rrd/encoder.rs
```diff
@@ -128,31 +124,36 @@ impl<W: std::io::Write> Encoder<W> {
...
-        match self.serializer {
-            Serializer::Protobuf => {
-                crate::codec::file::encoder::encode(&mut self.scratch, message, self.compression)?;
-
-                w.write_all(&self.scratch)
-                    .map(|_| self.scratch.len() as _)
-                    .map_err(EncodeError::Write)
-            }
+        let message = message.to_transport(self.compression)?;
+        // Safety: we _are_ the encoder.
```
abey79:
```suggestion
// Safety: the compression settings of this message are consistent with this stream.
```

crates/store/re_log_encoding/src/rrd/headers.rs
```diff
@@ -0,0 +1,263 @@
...
+            version,
+            options,
+        } = self;
+
+        let before = out.len() as u64;
+
+        out.extend_from_slice(fourcc);
+        out.extend_from_slice(version);
+        {
+            // TODO(cmc): the extra heap-alloc and copy could be easily avoided.
```
abey79: why don't you then? :troll:

More seriously, I don't mind these kinds of TODO (culprit of the same), but I do try to leave a breadcrumb as to why I didn't (or open an issue, etc.)

crates/store/re_log_encoding/src/rrd/decoder/mod.rs
abey79: Yeah I'm slightly confused by this trait. Is it really worth exposing this as anything else than an implementation utility for `::decoder`? If so, best present it as such.

If not, I'm slightly bothered by the fact that its purpose span the full byte->transport->application chain, yet it is in ::rrd.

Caveat: I might have misunderstood the whole thing.

crates/store/re_log_encoding/src/rrd/decoder/iterator.rs
```diff
@@ -0,0 +1,572 @@
...
+
+                        // …and the underlying decoder already considers that it's done (i.e. it's
+                        // waiting for a whole new stream to begin): time to stop.
+                        Ok(None) if self.decoder.state == DecoderState::StreamHeader => {
+                            return None;
+                        }
+
+                        // …but the underlying decoder doesn't believe it's done yet (i.e. it's still
+                        // waiting for an EOS marker to show up): we continue.
+                        Ok(None) => {}
```
abey79, line 148: not super familiar with rust io. Is `fill_buf()` here the thing that guaranties the loop doesn't go 100% cpu nooping when in that situation?

crates/store/re_log_encoding/src/transport_to_app.rs
```diff
@@ -0,0 +1,382 @@
+//! Everything needed to convert back and forth between transport-level and application-level types.
+//!
+//! ⚠️Make sure to familiarize yourself with the [crate-level docs] first. ⚠️
+//!
+//! This is where all the complex application-level logic that precedes encoding / follows decoding
+//! happens: Chunk/Sorbet migrations, data patching (app ID injection, version propagation, BW-compat hacks,
+//! etc).
+//!
+//! To go from a from a freshly decoded transport-level type to its application-level equivalent, use [`ToApplication`].
```
abey79:
```suggestion
//! To go from a freshly decoded transport-level type to its application-level equivalent, use [`ToApplication`].
```

abey79 (approved):
Awesome work. Caveat: I spend much more time learning about stuff than actually checking code correctness 😬

Some merge conflict to be expected with https://github.com/rerun-io/rerun/pull/11532 as I rewrote large parts of `re_server` and `re_redap_tests`.

crates/store/re_protos/src/v1alpha1/rerun.common.v1alpha1.ext.rs
teh-cmc: :shrug: I'll open an issue.

teh-cmc: * https://github.com/rerun-io/rerun/issues/11603

crates/store/re_log_encoding/src/rrd/mod.rs
teh-cmc: And more cognitive overhead for the next soul, of which there is way too much already imo

crates/store/re_log_encoding/src/rrd/decoder/iterator.rs
teh-cmc: Yeah `fill_buf` lowers to standard IO under the hood

Issue #11581, opened 2025-10-17
`rerun-cli` now has `protoc` as an undocumented build requirement


kpreid:
**Describe the annoyance**
Compiling `rerun-cli` 0.26 now seems to require `protoc` to be already installed, which was not the case with previous versions. This requirement is reasonable but should be documented.

**To Reproduce**
1. Read <https://rerun.io/docs/getting-started/installing-viewer#rust>
2. `cargo install rerun-cli@0.26.0 --locked`

**Expected behavior**
Installation succeeds after following the documented steps.

**Actual behavior**
```
error: failed to run custom build command for `lance-encoding v0.36.0`

Caused by:
  process didn't exit successfully: `/var/folders/bm/qff3m4w11hj5p03sytlsm8nw0000gn/T/cargo-installi84nhZ/release/build/lance-encoding-955743dc6baceb0d/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-changed=protos

  --- stderr
  Error: Custom { kind: NotFound, error: "Could not find `protoc`. If `protoc` is installed, try setting the `PROTOC` environment variable to the path of the `protoc` binary. To install it on macOS, run `brew install protobuf`. It is also available at https://github.com/protocolbuffers/protobuf/releases  For more information: https://docs.rs/prost-build/#sourcing-protoc" }
```

**Desktop (please complete the following information):**
 - OS: macOS Sequoia 15.7.1

Wumpf:
thanks for flagging this! Totally missed that we dragged that in!

> This requirement is reasonable

not convinced about that yet...
But you're ofc right if we keep it that way we have to document it, at the very least for the affected version(s)

Wumpf:
documented it now, but keeping this issue open - want to investigate what we can do about this :/

emilk:
Looks like we just need to add this:

```
[build-dependencies]
protoc-prebuilt.workspace = true
```

to `crates/store/re_server/Cargo.toml`


**EDIT**: nope, the this needs to be added to _lance_

emilk:
So the problem is that `lance` depends on `lance-encoding` with default features, and `lance-encoding` by default requires a system `protoc`:

* https://github.com/lancedb/lance/issues/3073

There is no quick fix for this.

We avoid depending on crates that need more than `cargo` to be build (and doesn't install those dependencies in their build.rs). I'm not sure how to best detect that though.

My thinking right now is:

* Make `lance` an opt-in dependency and do a patch-release
[...]

PR #11583, opened 2025-10-17
Mention `protoc` requirement in migration guide & viewer install guide

Wumpf:
### Related

* Caught here #11581 

### What

Not closing above issue since I really don't want this requirement to begin with if we can avoid it. Let's see if we can patch it out again, not terribly hopeful right now.

Already added it to the 0.26 changelog.


jprochazk (approved):
Damn, that sucks... Have we tried to introduce them to our lord and savior, offline codegen?

jprochazk:
Worst case we can publish our own fork of lance, with every codegen checked in to the repo

PR #11588, opened 2025-10-20
When hovering an instance, focus on the non-splatted components

emilk:
### Before
image

### After
image

### DIFF
--- a/crates/viewer/re_data_ui/src/instance_path.rs
+++ b/crates/viewer/re_data_ui/src/instance_path.rs
@@ -124,7 +124,7 @@ impl DataUi for InstancePath {
             let mut show_only_instanced = false;
 
             if !self.is_all() {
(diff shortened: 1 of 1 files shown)


crates/viewer/re_data_ui/src/instance_path.rs
```diff
@@ -113,35 +116,92 @@ impl DataUi for InstancePath {
...
-                "{} archetype{}: {}",
-                archetype_count,
-                if archetype_count > 1 { "s" } else { "" },
-                components_by_archetype
-                    .keys()
-                    .map(|archetype| {
-                        if let Some(archetype) = archetype {
-                            archetype.short_name()
+            if !self.is_all() {
+                // Focus on the components that has different values per instance (non-splatted components):
```
oxkitsune:
```suggestion
                // Focus on the components that have different values per instance (non-splatted components):
```

PR #11589, opened 2025-10-20
Container hierarchy tests

aedm:
### Related

Part of #8948 

### What

- Adds automated tests for container hierearchy manipulation
- Removes `check_container_hierarchy.py` from release checklist


crates/viewer/re_ui/src/ui_ext.rs
```diff
@@ -1249,13 +1250,16 @@ pub trait UiExt {
...
+        let response = egui::ComboBox::from_id_salt(id_salt)
+            .selected_text(selected_text.clone())
             .show_ui(self.ui_mut(), |ui| {
                 list_item::list_item_scope(ui, "inner_scope", |ui| {
                     content(ui);
                 });
             });
+        response.response.widget_info(move || {
+            WidgetInfo::labeled(WidgetType::ComboBox, true, selected_text.clone())
+        });
```
lucasmerlin: I was going to comment 

> You should be able to find it via `get_by_value`

but turns out we never set the value. Opened a PR for egui: https://github.com/emilk/egui/pull/7658

Maybe add a TODO(emilk/egui#7658) so we don't forget to remove it.


Also it'd be awesome if the things in our PropertyContent list items could automatically be labeled by the list item. (But that would be a huge refactor since the value fn currenly doesn't return a Response)

tests/rust/re_integration_test/tests/multi_container_test.rs
```diff
@@ -190,3 +190,95 @@ pub async fn test_multi_container_drag_container() {
     harness.drop_nth_label("Viewport (Grid container)", 0);
     harness.snapshot_app("multi_container_drag_container_6");
 }
+
+#[tokio::test(flavor = "multi_thread")]
+pub async fn test_add_container_from_blueprint_panel_menu() {
+    let mut harness = make_multi_view_test_harness();
+    add_containers_recursive(&mut harness, None, 1, 4, 0);
+
+    harness.click_label("Open menu with more options");
```
lucasmerlin, line 202:
```suggestion
    // Blueprint panel "…" icon
    harness.click_label("Open menu with more options");
```

tests/rust/re_integration_test/tests/multi_container_test.rs
```diff
@@ -190,3 +190,95 @@ pub async fn test_multi_container_drag_container() {
...
+    harness.snapshot_app("add_container_from_blueprint_panel_menu_1");
+
+    // Blueprint panel "…" icon
+    harness.click_label_contains("Add view or container");
```
lucasmerlin:
```suggestion
    harness.click_label_contains("Add view or container");
```

tests/rust/re_integration_test/tests/multi_container_test.rs
```diff
@@ -190,3 +190,95 @@ pub async fn test_multi_container_drag_container() {
...
+
+    harness.click_label("Viewport (Grid container)");
+    harness.snapshot_app("add_container_from_selection_panel_1");
+
+    // Selection panel "+" icon
+    harness.click_label("Add a new view or container to this container");
+    harness.snapshot_app("add_container_from_selection_panel_2");
+
+    harness.click_label("Vertical");
+    harness.snapshot_app("add_container_from_selection_panel_3");
```
lucasmerlin, line 226: The screenshots are nice but I think we should also have asserts for the expected results. Something like
```rs
let selection_panel = harness.get_by_label("Selection");
assert_eq(selection_panel.query_all_by_label("Vertical container").count(), 3);
```

tests/rust/re_integration_test/tests/multi_container_test.rs
```diff
@@ -190,3 +190,95 @@ pub async fn test_multi_container_drag_container() {
...
+
+    harness.click_nth_label("Vertical container", 0);
+    harness.snapshot_app("change_container_type_1");
+
+    harness.click_label("Vertical");
+    harness.snapshot_app("change_container_type_2");
+
+    harness.click_label("Horizontal");
+    harness.snapshot_app("change_container_type_3");
+}
```
lucasmerlin, line 245: It'd be nice to have a `harness.combobox_selection("combobox_label", "new_value")` helper on HarnessExt.

tests/python/release_checklist/check_container_hierarchy.py
```diff
@@ -1,110 +0,0 @@
...
-* Select the "Viewport" root container, and click "Simplify hierarchy" in the Selection Panel.
-    * The entire tree should be clean of spurious empty container.
-
-
-### Drag-and-drop in the viewport
-
-TODO(ab): be _way_ more specific exact actions and expected outcomes when drag-and-dropping tiles.
-
-* Check that the tiles may be resized by dragging boundaries in the Viewport.
-* In the Viewport, drag tiles around to build a different container hierarchy.
```
lucasmerlin, line 58: Do we have a test for dragging of tiles?

crates/viewer/re_ui/src/ui_ext.rs
aedm: Changed the code so the dropdown has a fix `Container kind` label. Now the code to change it is:
```rs
harness.change_dropdown_value("Container kind", "Horizontal");
```

tests/rust/re_integration_test/tests/multi_container_test.rs
aedm: Currently, `Selection`has no children. I added an assertion with a reference to the ticket for section-based queries: https://github.com/rerun-io/rerun/issues/11628

tests/python/release_checklist/check_container_hierarchy.py
aedm: Good catch! Added a bunch of tests for that. I added labels to views to find them, used `egui::WidgetType::Panel`.

PR #11592, opened 2025-10-20
Fix `cargo install rerun-cli` requiring `protoc` on `PATH`

emilk:
* Fixes https://github.com/rerun-io/rerun/issues/11581
* Working around https://github.com/lancedb/lance/issues/3073
* Feature-flags what was added in https://github.com/rerun-io/rerun/pull/11356

## What
* `re_server` has a new opt-in feature `lance`
* `rerun-cli` has a new opt-in feature `lance`

These feature-gate loading in `lance` files/directories as tables

## Details
~Initially this PR feature-gated the whole `table` feature. That required moving a lot of file-wide `use` statements into individual functions. I did not move back all of these, because in general I think we should only use file-wide `use` statements for things that are repeatedly used throughout the file. If you're only using it once or twice, it is better to use a fully-qualified path. And if you're only using it within a single function, put the `use` statement inside that function. This makes it _a lot_ easier to refactor code (moving/removing functions, feature-gating, …).~

## TODO
* [x] @rerun-bot full-check
* [x] Make sure `cargo build -p rerun-cli` works with no `protoc` on path

### DIFF
--- a/crates/store/re_server/src/store/in_memory_store.rs
+++ b/crates/store/re_server/src/store/in_memory_store.rs
@@ -223,7 +223,7 @@ impl InMemoryStore {
         Ok(())
     }
 
-    #[cfg(feature = "lance")]
+    #[cfg(feature = "lance")] // only used by the `lance` feature
     fn add_table_entry(
         &mut self,
         entry_name: &str,


emilk:
@rerun-bot full-check

emilk:
@rerun-bot full-check

crates/store/re_server/src/store/in_memory_store.rs
```diff
@@ -331,7 +331,24 @@ impl InMemoryStore {
...
+            ArrayRef, Int32Array, RecordBatch, RecordBatchOptions, StringArray,
+            TimestampNanosecondArray,
+        },
+        datatypes::{DataType, Field, Schema, TimeUnit},
+    };
+    use datafusion::error::DataFusionError;
+    use itertools::Itertools as _;
+
+    use re_tuid::Tuid;
+    use re_types_core::{ComponentBatch as _, Loggable as _};
```
emilk: Local imports makes it easier to feature-gate and move code around

emilk:
@rerun-bot full-check

crates/store/re_server/src/store/in_memory_store.rs
timsaucer: If this is a style you want contributors to follow then I suggest adding it to `CODE_STYLE.md`.

timsaucer: Personally, I don't like it because it carries additional mental load of tracking how often you use certain imports throughout the code. If there's some kind of lint that would auto-check this, I don't know about it yet. But if this is how you want the repo organized I think it needs to be documented.

crates/store/re_server/src/store/in_memory_store.rs
```diff
@@ -220,11 +215,12 @@ impl InMemoryStore {
         Ok(())
     }
 
+    #[cfg(feature = "lance")]
```
abey79: I dont think this should be here.


```suggestion
```

crates/store/re_server/src/store/in_memory_store.rs
emilk: * https://github.com/rerun-io/rerun/pull/11597

abey79: Reviewing this particular PR, I should add that I need to focus quite a bit to spot the actual change vs. sprawling `use`/fq naming change.

crates/store/re_server/src/rerun_cloud.rs
abey79, line 1082: I was wondering why there is no diff around here 😅

crates/store/re_server/src/store/in_memory_store.rs
emilk: 100% agree it does not belong in this PR, but I ran out of time. I can undo the use-stuff tomorrow

Issue #10903, opened 2025-08-14
Viewer queries incorrectly distinguishes by component metadata like archetype & field name rather than just component identifier


Wumpf:
We still use `ComponentDescriptor` as a key. This is fine for builtin types, but for user provided types this can cause issues. Details see https://github.com/rerun-io/rerun/issues/10879#issuecomment-3182785580

Wumpf:
repro code, using a bunch of deprecated methods.
```py
import rerun as rr

rr.init("duplicate_tagged_components", spawn=True)
# These are two groupings of content with a variety of fields but at least one intersects
schema = rr.AnyValues().with_field(rr.ComponentDescriptor("data", "mcap.Schema"), 1.0)
image = rr.AnyValues().with_field(rr.ComponentDescriptor("data", "SomeCustomImage"), 2.0)
rr.log("/camera/data", schema)
rr.log("/camera/data", image)
```

also causes massive debug-only warning spam.
Some warnings I would expect since the descriptor changed over time.

Issue #10460, opened 2025-07-01
Use `ComponentIdentifier` instead of `ComponentDescriptor` indexing into chunks


grtlr:
### What

With the recent changes to `ComponentDescriptor` it should be enough to use the `ComponentIdentifier` from the `component` field of the descriptor to index in chunks.

This should simplify our logic in many places and increase performance. These changes might also be relevant for choosing which fields (and how) to intern.

### Related

* Follow up of #10245.

Wumpf:
started fixing this on `andreas/fix-handling-changing-descriptors`

Wumpf:
Addressed most of it in https://github.com/rerun-io/rerun/pull/11593
A few places, mostly about "hybrid results" (the thing that most visualizers operate on) are still using descriptors. But all the indexing happens via identifiers now.

PR #11593, opened 2025-10-20
Fix handling of components that only vary by descriptors

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/10903
* Aaaalmost all of https://github.com/rerun-io/rerun/issues/10460
    * scrolling over it it doesn't look like I've left opportunities out to not pass descriptors but oh my there's still more! 

### What

... by biting the bullet and pretty much everything that so far indexes of `ComponentDescriptor` instead index over `ComponentIdentifier`.
I'd like to believe this also improves performance, but I haven't proven it.

There's a lot of access that goes `descriptor_THING().component` now which strongly implies that we should generate `component_THING()` instead. But I've left that as an exercise for the reader.

Commit by commit if you're into this thing - every successive commit is curated to walk up the stack of crates

TODO:
* [x] confirm the repro on #10903 behaves to our liking now

Needed follow-ups:
* Generate `component_THING()` methods (use them inside of `descriptor_THING()`)
    * use it to tone down existing `.component` sillyness
* Use `component_THING` for queries in visualizers -> solve  #10460 for good

### DIFF
--- a/crates/store/re_chunk/tests/latest_at.rs
+++ b/crates/store/re_chunk/tests/latest_at.rs
@@ -531,7 +531,7 @@ fn static_unsorted() -> anyhow::Result<()> {
 
 // ---
 
-// TODO(andread): This doesn't have to take a full descriptor, but all our access methods are using descriptors right now.
+// TODO(andreas): This doesn't have to take a full descriptor, but all our access methods are using descriptors right now.
 fn query_and_compare(
     (component_desc, query): (ComponentDescriptor, &LatestAtQuery),
     chunk: &Chunk,
--- a/crates/store/re_chunk/tests/range.rs
+++ b/crates/store/re_chunk/tests/range.rs
@@ -500,7 +500,7 @@ fn static_unsorted() -> anyhow::Result<()> {
 
 // ---
 
-// TODO(andread): This doesn't have to take a full descriptor, but all our access methods are using descriptors right now.
+// TODO(andreas): This doesn't have to take a full descriptor, but all our access methods are using descriptors right now.
 fn query_and_compare(
     (component_desc, query): (ComponentDescriptor, &RangeQuery),
     chunk: &Chunk,
--- a/crates/store/re_chunk_store/tests/correctness.rs
+++ b/crates/store/re_chunk_store/tests/correctness.rs
@@ -10,7 +10,7 @@ use re_log_types::{
     Duration, EntityPath, TimeInt, TimePoint, TimeType, Timeline, Timestamp, build_frame_nr,
     build_log_time,
 };
-use re_types::ComponentDescriptor;
+use re_types::ComponentIdentifier;
 
 // ---
 
@@ -18,22 +18,20 @@ fn query_latest_component<C: re_types_core::Component>(
     store: &ChunkStore,
     entity_path: &EntityPath,
     query: &LatestAtQuery,
-    component_descr: &ComponentDescriptor,
+    component: ComponentIdentifier,
 ) -> Option<(TimeInt, RowId, C)> {
     re_tracing::profile_function!();
 
     let ((data_time, row_id), unit) = store
-        .latest_at_relevant_chunks(query, entity_path, component_descr.component)
+        .latest_at_relevant_chunks(query, entity_path, component)
         .into_iter()
         .filter_map(|chunk| {
-            let unit = chunk
-                .latest_at(query, component_descr.component)
-                .into_unit()?;
+            let unit = chunk.latest_at(query, component).into_unit()?;
             unit.index(&query.timeline()).map(|index| (index, unit))
         })
         .max_by_key(|(index, _unit)| *index)?;
 
-    unit.component_mono(component_descr.component)?
+    unit.component_mono(component)?
         .ok()
         .map(|values| (data_time, row_id, values))
 }
@@ -84,7 +82,7 @@ fn row_id_ordering_semantics() -> anyhow::Result<()> {
                 &store,
                 &entity_path,
                 &query,
-                &MyPoints::descriptor_points(),
+                MyPoints::descriptor_points().component,
             )
             .unwrap();
             similar_asserts::assert_eq!(point2, got_point);
@@ -157,7 +155,7 @@ fn row_id_ordering_semantics() -> anyhow::Result<()> {
                 &store,
                 &entity_path,
                 &query,
-                &MyPoints::descriptor_points(),
+                MyPoints::descriptor_points().component,
             )
             .unwrap();
             similar_asserts::assert_eq!(point1, got_point);
@@ -203,7 +201,7 @@ fn row_id_ordering_semantics() -> anyhow::Result<()> {
                 &store,
                 &entity_path,
                 &query,
-                &MyPoints::descriptor_points(),
+                MyPoints::descriptor_points().component,
             )
             .unwrap();
             similar_asserts::assert_eq!(point1, got_point);
@@ -247,7 +245,7 @@ fn row_id_ordering_semantics() -> anyhow::Result<()> {
                 &store,
                 &entity_path,
                 &query,
-                &MyPoints::descriptor_points(),
+                MyPoints::descriptor_points().component,
             )
             .unwrap();
             similar_asserts::assert_eq!(point1, got_point);
--- a/crates/store/re_chunk_store/tests/gc.rs
+++ b/crates/store/re_chunk_store/tests/gc.rs
@@ -3,7 +3,9 @@ use std::sync::Arc;
 use arrow::array::ArrayRef;
 use rand::Rng as _;
 
-use re_chunk::{Chunk, ChunkId, LatestAtQuery, RowId, TimeInt, TimePoint, TimelineName};
+use re_chunk::{
+    Chunk, ChunkId, ComponentIdentifier, LatestAtQuery, RowId, TimeInt, TimePoint, TimelineName,
+};
 use re_chunk_store::{
     ChunkStore, ChunkStoreConfig, ChunkStoreDiffKind, GarbageCollectionOptions,
     GarbageCollectionTarget,
@@ -22,23 +24,21 @@ use re_types::{
 fn query_latest_array(
     store: &ChunkStore,
     entity_path: &EntityPath,
-    component_descr: &ComponentDescriptor,
+    component: ComponentIdentifier,
     query: &LatestAtQuery,
 ) -> Option<(TimeInt, RowId, ArrayRef)> {
     re_tracing::profile_function!();
 
     let ((data_time, row_id), unit) = store
-        .latest_at_relevant_chunks(query, entity_path, component_descr.component)
+        .latest_at_relevant_chunks(query, entity_path, component)
         .into_iter()
         .filter_map(|chunk| {
-            let chunk = chunk
-                .latest_at(query, component_descr.component)
-                .into_unit()?;
+            let chunk = chunk.latest_at(query, component).into_unit()?;
             chunk.index(&query.timeline()).map(|index| (index, chunk))
         })
         .max_by_key(|(index, _chunk)| *index)?;
 
-    unit.component_batch_raw(component_descr.component)
+    unit.component_batch_raw(component)
         .map(|array| (data_time, row_id, array))
 }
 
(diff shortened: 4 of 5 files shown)


Wumpf:
The repro case for the original example has become a bit more pathological. but it has handled correctly.

```py
import rerun as rr

rr.init("duplicate_tagged_components", spawn=True)
# These are two groupings of content with a variety of fields but at least one intersects
schema = rr.DynamicArchetype("mcap.Schema")._with_descriptor_internal(
    rr.ComponentDescriptor("data", "mcap.Schema"), 1.0
)
image = rr.DynamicArchetype("SomeCustomImage")._with_descriptor_internal(
    rr.ComponentDescriptor("data", "SomeCustomImage"), 2.0
)
rr.log("/camera/data", schema)
rr.log("/camera/data", image)
```

The first descriptor wins - we viewer only shows/receives `mcap.Schema`.
No warning is logged right now which is a bit unfortunate. But given how hard it is to hit this now I'm ok with that.

crates/store/re_chunk/src/batcher.rs
```diff
@@ -784,18 +784,35 @@ pub struct PendingRow {
...
-    pub components: IntMap<ComponentDescriptor, ArrayRef>,
+    pub components: IntMap<ComponentIdentifier, (ComponentDescriptor, ArrayRef)>,
 }
 
 impl PendingRow {
     #[inline]
-    pub fn new(timepoint: TimePoint, components: IntMap<ComponentDescriptor, ArrayRef>) -> Self {
+    pub fn new(
+        timepoint: TimePoint,
+        components: IntMap<ComponentIdentifier, (ComponentDescriptor, ArrayRef)>,
```
grtlr: Now I wonder if `rustc` will monomorphize the `from_iter` for `IntMap` into the same function as this (probably not).

crates/store/re_chunk/src/chunk.rs
```diff
@@ -72,7 +72,7 @@ pub type ChunkResult<T> = Result<T, ChunkError>;
 // ---
 
 #[derive(Debug, Clone, Default, PartialEq)]
-pub struct ChunkComponents(pub IntMap<ComponentDescriptor, ArrowListArray>);
+pub struct ChunkComponents(pub IntMap<ComponentIdentifier, (ComponentDescriptor, ArrowListArray)>);
```
grtlr: I think this is called `SerializedComponentColumn`.
```suggestion
pub struct ChunkComponents(pub IntMap<ComponentIdentifier, SerializedComponentColumn>);
```

crates/store/re_chunk/src/chunk.rs
```diff
@@ -157,7 +180,10 @@ impl FromIterator<SerializedComponentColumn> for ChunkComponents {
         let mut this = Self::default();
         {
             for serialized in iter {
-                this.insert(serialized.descriptor, serialized.list_array);
+                this.insert(
+                    serialized.descriptor.component,
+                    (serialized.descriptor, serialized.list_array),
+                );
```
grtlr: See above, maybe just use it directly in the map?

crates/store/re_chunk/src/chunk.rs
grtlr, line 1: This could maybe be made more self-documenting by using `SerializedComponentColumn` in more places and adding some helpers on it. But I can understand not going there, after all these changes.

crates/store/re_chunk/src/batcher.rs
```diff
@@ -784,18 +784,35 @@ pub struct PendingRow {
     /// The component data.
     ///
     /// Each array is a single component, i.e. _not_ a list array.
-    pub components: IntMap<ComponentDescriptor, ArrayRef>,
+    pub components: IntMap<ComponentIdentifier, (ComponentDescriptor, ArrayRef)>,
```
grtlr: This then becomes a `SerializedComponentBatch`, I believe.
```suggestion
    pub components: IntMap<ComponentIdentifier, SerializedComponentBatch>,
```

crates/store/re_chunk/src/iter.rs
```diff
@@ -263,19 +260,19 @@ impl Chunk {
...
-        let Some(list_array) = self.components.get(&component_descriptor) else {
+        let Some(list_array) = self.components.get_array(component) else {
             return Either::Left(std::iter::empty());
         };
 
         let Some(struct_array) = list_array.values().downcast_array_ref::<ArrowStructArray>()
         else {
             if cfg!(debug_assertions) {
-                panic!("downcast failed for {component_descriptor}, data discarded");
+                panic!("downcast failed for {component}, data discarded");
```
grtlr: Nit, because this rang a bell: would be a nice drive-by fix to add source and target data types to the asserts/errors in this file, for example by adding a free function that receives the target as a generic parameter.

I've done this for one function below, but should have looked at the entire file back then. 🙈

crates/store/re_chunk/src/iter.rs
```diff
@@ -944,15 +933,9 @@ impl Chunk {
...
+        component: ComponentIdentifier,
     ) -> ChunkComponentIter<C, impl Iterator<Item = Span<usize>> + '_ + use<'_, C>> {
-        debug_assert_eq!(
-            component_descriptor.component_type,
-            Some(C::name()),
-            "component type mismatch"
-        );
-
-        let Some(list_array) = self.components.get(component_descriptor) else {
+        let Some(list_array) = self.components.get_array(component) else {
```
grtlr, line 902: The training wheels are off!

crates/store/re_chunk/tests/latest_at.rs
```diff
@@ -531,14 +531,15 @@ fn static_unsorted() -> anyhow::Result<()> {
 
 // ---
 
+// TODO(andread): This doesn't have to take a full descriptor, but all our access methods are using descriptors right now.
```
grtlr: Nice try 😉

```suggestion
// TODO(andreas): This doesn't have to take a full descriptor, but all our access methods are using descriptors right now.
```

crates/store/re_chunk/tests/range.rs
```diff
@@ -500,14 +500,15 @@ fn static_unsorted() -> anyhow::Result<()> {
 
 // ---
 
+// TODO(andread): This doesn't have to take a full descriptor, but all our access methods are using descriptors right now.
```
grtlr:
```suggestion
// TODO(andreas): This doesn't have to take a full descriptor, but all our access methods are using descriptors right now.
```

crates/store/re_chunk_store/tests/correctness.rs
```diff
@@ -23,15 +23,17 @@ fn query_latest_component<C: re_types_core::Component>(
     re_tracing::profile_function!();
 
     let ((data_time, row_id), unit) = store
-        .latest_at_relevant_chunks(query, entity_path, component_descr)
+        .latest_at_relevant_chunks(query, entity_path, component_descr.component)
```
grtlr: Nit: This is only a test, but maybe `query_latest_component` should take a `ComponentIdentifier`? That would make it symmetric to `latest_at_relevant_chunks`

crates/store/re_chunk_store/tests/gc.rs
```diff
@@ -28,15 +28,17 @@ fn query_latest_array(
     re_tracing::profile_function!();
 
     let ((data_time, row_id), unit) = store
-        .latest_at_relevant_chunks(query, entity_path, component_descr)
+        .latest_at_relevant_chunks(query, entity_path, component_descr.component)
```
grtlr: nit: same here

crates/store/re_dataframe/src/query.rs
```diff
@@ -345,13 +345,8 @@ impl<E: StorageEngineLike> QueryHandle<E> {
...
                             re_chunk::LatestAtQuery::new(TimelineName::new(""), TimeInt::STATIC);
 
-                        let component_descriptor = store
-                            .entity_component_descriptor(&descr.entity_path, descr.component)
-                            .into_iter()
-                            .next()?;
-
                         let results =
-                            cache.latest_at(&query, &descr.entity_path, [&component_descriptor]);
+                            cache.latest_at(&query, &descr.entity_path, [descr.component]);
```
grtlr, line 351: The deletion above is haunting me, but I guess that was just an early exit? Looking at the diff, we could have always just passed `descr` along.

Or did that line just become trivial because of @IsseW's recent refactor?

crates/store/re_dataframe/src/query.rs
```diff
@@ -1064,16 +1059,8 @@ impl<E: StorageEngineLike> QueryHandle<E> {
...
-                        .into_iter()
-                        .next()?;
-
-                    let results = cache.latest_at(
-                        &query,
-                        &descr.entity_path.clone(),
-                        [&component_descriptor],
-                    );
+                    let results =
+                        cache.latest_at(&query, &descr.entity_path.clone(), [descr.component]);
```
grtlr, line 1065: Same here.

crates/viewer/re_view/src/query.rs
```diff
@@ -24,27 +24,26 @@ use re_viewer_context::{DataResult, QueryRange, ViewContext, ViewQuery, ViewerCo
 ///
 /// Data should be accessed via the [`crate::RangeResultsExt`] trait which is implemented for
 /// [`crate::HybridResults`].
-pub fn range_with_blueprint_resolved_data<'a, 'b>(
+pub fn range_with_blueprint_resolved_data<'a>(
```
grtlr, line 27: A leftover, but the `component_types` in this function's comment is outdated.

crates/viewer/re_view/src/query.rs
```diff
@@ -66,29 +65,29 @@ pub fn range_with_blueprint_resolved_data<'a, 'b>(
 /// [`crate::HybridResults`].
 ///
 /// If `query_shadowed_components` is true, store components will be queried, even if they are not used.
-pub fn latest_at_with_blueprint_resolved_data<'a, 'b>(
+// TODO(#10460): Should take identifiers instead of descriptors.
+pub fn latest_at_with_blueprint_resolved_data<'a>(
```
grtlr, line 69: Same here: A leftover, but the `component_types` in this function's comment is outdated.

crates/viewer/re_view/src/results_ext.rs
```diff
@@ -94,15 +94,15 @@ impl HybridLatestAtResults<'_> {
     pub fn get_instance<C: re_types_core::Component>(
         &self,
         index: usize,
-        component_descr: &ComponentDescriptor,
+        component_descr: &ComponentDescriptor, // TODO(#10460): Take `ComponentIdentifier` instead of descriptor.
```
grtlr, line 97: Btw I was just browsing around, and we should probably do `blueprint_helpers.rs` (reset, ...) as part of #10460 too.

grtlr (approved):
Wow, this is already looking much nicer 🤩!

All of my comments are minor and only suggestions / notes to myself. Thank you for going through the effort of starting to make these changes.

crates/store/re_chunk/src/batcher.rs
Wumpf: Don't think so either. You're wondering whether the `from_iter` shortcut might be harmful?

crates/store/re_chunk/src/iter.rs
Wumpf: yep, my thought exactly. Keeping those checks was too hard and we haven't hit them that much

crates/viewer/re_view/src/results_ext.rs
Wumpf: many blueprint helpers write data and therefore need full descriptor

crates/store/re_dataframe/src/query.rs
Wumpf: I believe recently we had to add a brunch of descriptor lookups because of the chnage to component target, and those are gone now again because the descriptor is available in a more obvious fashion

don't see why it would be needed since we're asking the same store on the latest_at query then for the existance

crates/store/re_chunk/src/chunk.rs
```diff
@@ -161,13 +181,14 @@ impl std::ops::DerefMut for ChunkComponents {
     }
 }
 
+// TODO(andreas): Remove this variant, we should let users construct `SerializedComponentColumn` directly to sharpen semantics.
```
grtlr, line 184: Yes please (but not now, let's get this merged)!

PR #11594, opened 2025-10-20
Test adding entities to new view

aedm:
### Related

Part of #8948 

### What

- Adds automated tests adding entities to a new view
- Removes `check_context_menu_add_entity_to_new_space_view.py` from release checklist


tests/rust/re_integration_test/tests/add_entity_to_view_test.rs
```diff
@@ -0,0 +1,232 @@
...
+    harness.right_click_nth_label("boxes3d", 1);
+    harness.snapshot_app("add_entity_to_view_boxes3d_1");
+
+    harness.hover_label_contains("Add to new view");
+    harness.snapshot_app("add_entity_to_view_boxes3d_2");
+
+    harness.click_nth_label("3D", 0);
+    harness.snapshot_app("add_entity_to_view_boxes3d_3");
+
+    harness.right_click_nth_label("/", 2);
```
lucasmerlin: All this nth_label stuff feels a bit brittle to me.. I imagine it could become really annoying if we change unrelated code somewhere that maybe adds another place where we show the name of the selection and then we'd need to update a lot of random `nth_label` indexes in the integration tests. 

Maybe we could do something like `harness.blueprint().click_label("/")` via `HarnessExt`? 

I think we could even do this in a way that we can still automatically do `harness.run()` in the event helpers.

This should also help readability of the tests, as it's more obvious whats being interacted with.

tests/rust/re_integration_test/tests/add_entity_to_view_test.rs
```diff
@@ -0,0 +1,232 @@
...
+    harness.right_click_nth_label("text_log", 1);
+    harness.snapshot_app("add_entity_to_view_text_log_1");
+
+    harness.hover_label_contains("Add to new view");
+    harness.snapshot_app("add_entity_to_view_text_log_2");
+
+    harness.click_nth_label("Text log", 0);
+    harness.snapshot_app("add_entity_to_view_text_log_3");
+
+    // When adding a text log, to a new view, the origin is set to the entity path
```
lucasmerlin, line 237: Does this mean the Space origin should be `text_log`?

Maybe you can also add an assert for this?

tests/rust/re_integration_test/tests/add_entity_to_view_test.rs
aedm: I absolutely agree. I started with these when there was little hierarchy in our accessibility node tree, but now it should be fixed: https://github.com/rerun-io/rerun/issues/11628

aedm: Done.

tests/rust/re_integration_test/tests/add_entity_to_view_test.rs
aedm: Done.

tests/rust/re_integration_test/tests/add_entity_to_view_test.rs
```diff
@@ -0,0 +1,275 @@
+use egui::accesskit::Role;
```
emilk, line 3:
```suggestion
//! This file is all about…

use egui::accesskit::Role;
```

tests/rust/re_integration_test/tests/add_entity_to_view_test.rs
```diff
@@ -0,0 +1,275 @@
...
+            )),
+        )
+    });
+
+    harness.clear_current_blueprint();
+    harness
+}
+
+// Adds `count` views to the given container, names them sequentially from an index base.
+fn add_views_to_container(
```
emilk: `count` is what?

aedm: Updated to "`count` number of views".

PR #11595, opened 2025-10-20
Keyboard navigation in tree UIs (left panel, streams panel, etc.)

lucasmerlin:
### Related

* closes https://linear.app/rerun/issue/RR-2362/navigating-through-catalog-via-recordings-side-panel-with-hotkeys

### What

This adds keyboard navigation to all _interactive_ `list_items`.

- `up` / `down` to go to next / previous item
- `cmd+up` / `cmd+down` to expand / collapse collapsing items (freeing up left / right for timeline navigation)
- navigation is now based on egui focus
  - in some places focus is synced with selection (blueprint / streams / recording panel)
    - when navigating with the arrow keys, the focused item will be selected
    - when selection changes, the selected item will receive egui focus
    - when nothing has egui focus and a single item is currently selected, that item gets egui focus
  - tab focus now skips the list item chevrons
  - you can select list item buttons with a single tab button press (though order is confusing)
- I also made it so selecting something will scroll that item in the view
- also we ignored _all_ shortcuts when something was focused. I've changed to just ignore just the shortcuts without modifier key (and added special handling for space play/pause)

TODO:

- [x] different focus style
- [x] focus / select sync sometimes gets buggy when quickly switching through things
- [x] switch to app id on focus
- [x] fix pause key

### DIFF
--- a/crates/viewer/re_viewer_context/src/command_sender.rs
+++ b/crates/viewer/re_viewer_context/src/command_sender.rs
@@ -146,7 +146,7 @@ impl SystemCommand {
 
 /// What triggered this item to be selected?
 ///
-/// See [`handle_select_focus_sync`] why this is useful.
+/// See [`crate::ViewerContext::handle_select_focus_sync`] why this is useful.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum SelectionSource {
     ListItemNavigation,


crates/viewer/re_ui/src/list_item/list_item.rs
```diff
@@ -522,12 +574,14 @@ impl ListItem {
...
+            && !drag_target
+            && !egui::DragAndDrop::has_any_payload(ui.ctx());
+        let focused = style_response.has_focus();
         let visuals = ListVisuals {
             theme: ui.theme(),
-            hovered: (style_response.hovered() || style_response.contains_pointer())
-                && interactive
-                && !drag_target
-                && !egui::DragAndDrop::has_any_payload(ui.ctx()),
+            hovered: hovered || focused, // TODO: Should we have a separate focused style?
```
lucasmerlin: Should we have a seperate focused style or is the hovered style fine for this?

crates/viewer/re_ui/src/list_item/navigation.rs
```diff
@@ -0,0 +1,169 @@
...
+                gained_via_arrow.is_some_and(
+                    |GainedFocusViaArrowKey {
+                         widget_id: id,
+                         focused_on_pass,
+                     }| { id == list_item_id && focused_on_pass == pass - 1 },
+                )
+            } else {
+                false
+            }
+        })
```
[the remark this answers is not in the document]
lucasmerlin: Egui always fully executes passes, so this should be fine

crates/viewer/re_viewer_context/src/viewer_context.rs
```diff
@@ -401,8 +460,9 @@ impl ViewerContext<'_> {
                 .lookup_datatype(&component)
                 .or_else(|| self.blueprint_engine().store().lookup_datatype(&component))
                 .unwrap_or_else(|| {
-                         re_log::error_once!("Could not find datatype for component {component}. Using null array as placeholder.");
-                                    arrow::datatypes::DataType::Null})
+                    re_log::error_once!("Could not find datatype for component {component}. Using null array as placeholder.");
+                    arrow::datatypes::DataType::Null
+                })
```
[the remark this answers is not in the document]
lucasmerlin: It's what rustfmt wants

crates/viewer/re_ui/src/list_item/navigation.rs
```diff
@@ -0,0 +1,169 @@
...
+                let mut collapse_item = None;
+                let mut expand_item = None;
+                ctx.input_mut(|i| {
+                    if i.consume_key(Modifiers::COMMAND, egui::Key::ArrowUp) {
+                        match (navigation.focused_collapsed, navigation.focused_parent) {
+                            (Some(false), _) => {
+                                collapse_item = Some(current_focused);
+                            }
+                            (Some(true) | None, Some(parent)) => {
+                                focus_item = Some(parent);
```
lucasmerlin: Add some comments here

crates/viewer/re_ui/src/list_item/navigation.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    /// Is the focused item collapsible? What is its current state?
+    ///
+    /// None if the focused item is not collapsible.
+    pub focused_collapsed: Option<bool>,
+
+    /// What's the parent of the focused item?
+    ///
+    /// This should be the last item in `parent_stack` when `current_focused` is set.
+    pub focused_parent: Option<Id>,
```
abey79, line 23: Why does this need to be explicitly set? Could it be always derived from `parent_stack` and `current_focused`?

crates/viewer/re_ui/src/list_item/navigation.rs
```diff
@@ -0,0 +1,169 @@
...
+                d.get_temp_mut_or_default::<Self>(Id::NULL);
+            }
+
+            root
+        })
+    }
+
+    pub fn with_mut(ctx: &Context, f: impl FnOnce(&mut Self)) {
+        ctx.data_mut(|d| {
+            // Having a `has_temp` and/or `get_temp_mut` would be nice…
```
abey79, line 47: yes!

crates/viewer/re_ui/src/list_item/navigation.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /// Call this _only_ if [`Self::init_if_root`] returned true.
+    pub fn end_if_root(ctx: &Context) {
+        let navigation = ctx.data_mut(|d| d.remove_temp::<Self>(Id::NULL));
+
+        debug_assert!(navigation.is_some(), "Expected to find ListItemNavigation");
+
+        if let Some(navigation) = navigation {
+            if let Some(current_focused) = navigation.current_focused {
```
abey79: these need `let else` treatment to mitigate the right drift

crates/viewer/re_ui/src/list_item/navigation.rs
```diff
@@ -0,0 +1,169 @@
...
+                        state.store(ctx);
+                    }
+                }
+            }
+        }
+    }
+}
+
+/// Utility to check if focus was gained via arrow key navigation.
+#[derive(Debug, Clone)]
```
abey79, line 161:
```suggestion
/// Utility to check if focus was gained via arrow key navigation.
///
/// We need this because some UI parts (e.g. blueprint tree) will want to auto-select focused item _only_
/// if they were focused with arrows, not if they were focused e.g. with tab.
#[derive(Debug, Clone)]
```

crates/viewer/re_ui/src/list_item/navigation.rs
lucasmerlin: Not once we handle the navigation when the scope ends. 
I guess we could make it so we stop pushing / popping to it once we found the focused item, but I think the current way is more explicit and cleaner.

crates/viewer/re_viewer_context/src/command_sender.rs
```diff
@@ -136,7 +136,45 @@ pub enum SystemCommand {
...
+        Self::set_selection(crate::ItemCollection::default())
+    }
+
+    pub fn set_selection(selection: impl Into<SetSelection>) -> Self {
+        Self::SetSelection(selection.into())
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SelectionSource {
```
abey79, line 151:
```suggestion
/// What triggered this item to be selected?
///
/// See [`handle_select_focus_sync`] why this is useful.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum SelectionSource {
```

PR #11596, opened 2025-10-20
Functionality to `add_time_column` via lenses

grtlr:
### Related

* Closes RR-2692

### What

This adds functionality to extract a time column from a component column using lenses. This can, for example, be used to extract the `timestamp` that is present in some MCAP messages.

### DIFF
--- a/crates/top/re_sdk/src/lenses/ast.rs
+++ b/crates/top/re_sdk/src/lenses/ast.rs
@@ -100,7 +100,7 @@ impl Op {
 
     /// Ignores any input and returns a constant `ListArray`.
     ///
-    /// Commonly used with [`LensBuilder::add_static_output_column_entity`].
+    /// Commonly used with [`LensBuilder::add_static_component_column_entity`].
     /// When used in non-static columns this function will _not_ guarantee the correct amount of rows.
     pub fn constant(value: ListArray) -> Self {
         Self::func(move |_| Ok(value.clone()))


crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -196,13 +211,65 @@ impl Lens {
             ));
         }
 
+        let mut time_builders = ahash::HashMap::default();
+        for time in &self.time_columns {
+            let entity_path = time.new_entity_path.as_ref().unwrap_or(chunk.entity_path());
+
+            let time_entry = time_builders.entry(entity_path).or_insert_with(
+                <std::collections::HashMap<TimelineName, re_chunk::TimeColumn>>::default,
```
Wumpf: nit: doesn't `or_insert_default` do the trick here? If it's a type deduction issue, I'd rather type spec `let mut time_builders` :)

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -234,70 +301,131 @@ impl LensBuilder {
...
         };
-        self.0.outputs.push(column);
+        self.0.component_columns.push(column);
+        self
+    }
+
+    /// Can be used to extract a time column from the component specified via
+    /// [`Self::for_input_column`].
+    ///
+    /// The output column will be placed on the same entity path as the input column.
```
Wumpf: something is missing here: I was confused about whether this produces a time column on the output or is about input etc.; looking holistically at the API that got clear quickly, but maybe we can do a bit better:

on the output entity columns we call out that it's an output, so we should do the same here? 
-> `add_time_output_column`/ `add_output_time_column`?

Wumpf (approved):
nice, happy this is so straight forward! but some naming bike shedding might be in order :)

PR #11600, opened 2025-10-20
py-sdk: add register_prefix SDK call

andrea-reale:
### Related

* Closes RR-1962
* Follows:
  * https://github.com/rerun-io/rerun/pull/11564
  * https://github.com/rerun-io/dataplatform/pull/1864

### What

This PR builds up on the previous work and adds a new `register_prefix` python
SDK function that allows end-users to register all the recordings under a given
object store prefix (i.e., a directory).

### Why

In a previous version of this change, we allowed users to register a prefix
simply by implicitly interpreting any trailing '/' in the object store URL as meaning
the customer wanted to register a prefix.

As debated in #11564, however, this is error prone, so we moved to requiring an explicit signal.

Instead of modifying the signature of the existing `register_batch` function in non-obvious
and possibly odd ways, I decided to add a new function that is explicit in its intent, that is.
`register_prefix`.

### FAQs
* The redap API allows prefix and non-prefix `DataSources` to be mixed in the same call. Why does
  the python SDK not allow this?
  * We'd need to make the `register_batch` python function signature more complex and awkward.
    This keeps the user surface simple, while leaving room for future improvements if needed.
* Why not `register_prefixes` (plural)?
   * We could, but I don't see a strong use case for registering multiple prefixes
     at once.

### Testing
Manually

### DIFF
--- a/rerun_py/rerun_bindings/rerun_bindings.pyi
+++ b/rerun_py/rerun_bindings/rerun_bindings.pyi
@@ -1437,11 +1437,14 @@ class DatasetEntry(Entry):
 
         """
 
-    def register_prefix(self, recordings_prefix: str, recordings_layer: str | None = None) -> Tasks:
+    def register_prefix(self, recordings_prefix: str, layer_name: str | None = None) -> Tasks:
         """
         Register all RRDs under a given prefix to the dataset and return a handle to the tasks.
 
-        This method initiates the registration of multiple recordings to the dataset, and returns
+        A prefix is a directory-like path in an object store (e.g. an S3 bucket or ABS container).
+        All RRDs that are recursively found under the given prefix will be registered to the dataset.
+
+        This method initiates the registration of the recordings to the dataset, and returns
         the corresponding task ids in a [`Tasks`] object.
 
         Parameters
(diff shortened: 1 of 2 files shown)


rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1437,6 +1437,24 @@ class DatasetEntry(Entry):
 
         """
 
+    def register_prefix(self, recordings_prefix: str, recordings_layer: str | None = None) -> Tasks:
```
abey79: I'm not sure about the signature, see slack thread.

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1437,6 +1437,24 @@ class DatasetEntry(Entry):
...
+
+        This method initiates the registration of multiple recordings to the dataset, and returns
+        the corresponding task ids in a [`Tasks`] object.
+
+        Parameters
+        ----------
+        recordings_prefix: str
+            The prefix under which to register all RRDs.
+
+        recordings_layer: Optional[str]
```
abey79: This should be `layer_name` for consistency.

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1437,6 +1437,24 @@ class DatasetEntry(Entry):
 
         """
 
+    def register_prefix(self, recordings_prefix: str, recordings_layer: str | None = None) -> Tasks:
+        """
+        Register all RRDs under a given prefix to the dataset and return a handle to the tasks.
```
abey79: We should definitely expand on what we mean by `prefix` (maybe allude at s3, azure, file...?)

PR #11602, opened 2025-10-21
Expose view container visibility in python blueprint api

IsseW:
### What

Exposes the visibility option to containers (Horizontal, Vertical, …) in the python api.

### DIFF
--- a/crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
+++ b/crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
@@ -219,11 +219,11 @@ impl BlueprintTree {
                 .with_icon(icon_for_container_kind(&container_data.kind))
                 .subdued(!container_data.visible)
                 .with_buttons(|ui| {
+                    // If this has been hidden in a blueprint we want to be
+                    // able to make it visible again in the viewer.
                     if !container_data.visible {
                         let mut visible_after = container_data.visible;
                         visibility_button_ui(ui, true, &mut visible_after);
-                        // If this has been hidden in a blueprint we want to be
-                        // able to make it visible.
                         if visible_after != container_data.visible {
                             viewport_blueprint.set_content_visibility(
                                 ctx,


crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
```diff
@@ -216,7 +216,23 @@ impl BlueprintTree {
                     container_data.name.as_ref()
                 ))
                 .label_style(contents_name_style(&container_data.name))
-                .with_icon(icon_for_container_kind(&container_data.kind)),
+                .with_icon(icon_for_container_kind(&container_data.kind))
+                .subdued(!container_data.visible)
+                .with_buttons(|ui| {
+                    if !container_data.visible {
```
Wumpf, line 224: ohhh so if someone makes the root container invisible in blueprint, you make sure here to show that and give the user a way back? Would be good to comment the rationale here, took me a bit to understand what's going on

IsseW: Was confused because I do remember commenting that, it was just a bit below 😄 but yeah makes more sense to have the comment here

IsseW: > ohhh so if someone makes the root container invisible in blueprint, you make sure here to show that and give the user a way back?

Exactly, I ran into this exact thing in my testing :D

IsseW:
> Would be good to have some test coverage for container visibility, but I figure that's out of scope here

Yeah I agree, I tried looking and it doesn't seem we even have any snippets for view visibility.

Might also be nice to just have a snippet for just different containers in general?

Wumpf:
We could extend 
* https://github.com/rerun-io/rerun/pull/11589

once it lands :)

aedm:
> We could extend
> 
> * [Container hierarchy tests #11589](https://github.com/rerun-io/rerun/pull/11589)
> 
> once it lands :)

It landed and it adds finding a view by label. Also adds some panel drag&drop and resize tests if that's relevant.

But I'd wait a bit until I implement https://github.com/rerun-io/rerun/issues/11628, the harness should be more convenient to use after that.

PR #11606, opened 2025-10-21
Tests collapsing stream entities

aedm:
### Related

Part of #8948 

### What

- Adds automated tests for "Collapse all" menu for the stream tree, "Expand all" is already covered
- Removes `check_context_menu_collapse_expand_all.py` from release checklist

### DIFF
--- a/tests/rust/re_integration_test/tests/snapshots/collapse_stream_root_1.png
+++ b/tests/rust/re_integration_test/tests/snapshots/collapse_stream_root_1.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:f6a26a1f8a8ae925b9da391d11973eec1216614ab8972ef7037f74265eb46cc8
-size 57530
+oid sha256:1624b4cc2c77e8f6db988d681d9086e423fd1cb8460375ffb2dc25b9484ed325
+size 51493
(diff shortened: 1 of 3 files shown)


tests/rust/re_integration_test/tests/snapshots/xtemp.png
Wumpf, line 1: `xtemp.png`? Accidental checkin?

tests/rust/re_integration_test/tests/snapshots/collapse_stream_root_1.png
Wumpf, line 1: a bit unfortunate that we can't see the uncollapsed entities here - if for whatever reason the tree was collapsed already and collapsing breaks, we couldn't tell from this one.
Ah well, it still has the arrow down on the left of the root item, so I think this is good enough :)

tests/rust/re_integration_test/tests/snapshots/xtemp.png
aedm: Yes. :) Removed.

aedm:
> nice! I reckon the reason we can't test this on the time-panel in isolation is that it doesn't have the context menus registered, right? (If _possible_ this one would be nice to be isolated to `re_time_panel`)

We can now, but for the time panel though.

I updated the queries in this PR:
```rs
harness.streams_tree().right_click_label("txt/");
```

tests/rust/re_integration_test/tests/snapshots/collapse_stream_root_1.png
aedm: Fair. :) I added more screenshots.

Issue #11585, opened 2025-10-18
Reconnecting with same rerun id


lucasw:
**Describe the bug**

When I reconnect to rerun using the same name and rerun_id as previously used rerun becomes laggy and unresponsive, and I have to restart it to recover.  Using a fresh id works fine and as expected.

The play button is also behaving wrong, I can jump back in the timeline, press play, and then it fast forwards to the end instead of playing at a 1.0 rate.

**To Reproduce**

```
    let mut rec = rerun::RecordingStreamBuilder::new(rerun::ApplicationId::from(rerun_name.clone()));
    if let Some(rerun_id) = rerun_id {
        println!("using rerun id '{rerun_id}'");
        rec = rec.recording_id(rerun_id.to_string());
    }
    let rec = rec.spawn()?;
```

Every 1.0/60.0 seconds I'm updating the rerun time with current system time represented as an f64:

```
rec.set_timestamp_secs_since_epoch("view", t_seconds);
```
And then doing a bunch of rerun logs.

I thought I was confusing rerun by giving it a timestamp that was earlier in time than the end of the previous run, doubling up data every update, but even then the laggy behavior doesn't make sense, I would expect rerun to be robust against it and worst case is that the position of objects flickers between the old run and new run.

**Expected behavior**

I'd expect every time I reconnect using the same name and recording id that there would be a gap in the timeline for however long I was disconnected, and then the new telemetry would show up and play normally.

**Desktop (please complete the following information):**
 Linux Ubuntu 24.04

**Rerun version**

```
rerun-cli 0.26.0 (base map_view nasm native_viewer oss_server release_full release_no_web_viewer web_viewer) [rustc 1.88.0 (6b00bc388 2025-06-23), LLVM 20.1.5] x86_64-unknown-linux-gnu 11526/merge 831c187, built 2025-10-16T13:29:59.401210833Z
Video features: av1 default ffmpeg nasm serde
```

**Additional context**

lucasw:
I'm trying to create a reproducible example, but so far the example works fine (except for the play button jumping to the end of the timeline): https://github.com/lucasw/jolt_rust_cpp/tree/main/jolt_rerun

lucasw:
[...]

PR #11609, opened 2025-10-21
Fix pressing the play button sometimes going to follow mode

IsseW:
### Related

* Part of #11585

### What

Fixes a bug where pressing the play button in some contexts sometimes went to follow mode instead of playing mode.


crates/viewer/re_time_panel/src/time_control_ui.rs
```diff
@@ -155,7 +155,7 @@ You can also define your own timelines, e.g. for sensor time or camera frame num
             .on_hover_ui(|ui| UICommand::PlaybackTogglePlayPause.tooltip_ui(ui))
             .clicked()
         {
-            time_commands.push(TimeControlCommand::TogglePlayPause);
+            time_commands.push(TimeControlCommand::SetPlayState(PlayState::Playing));
```
grtlr: This looks suspicious (like we can't pause any more). There is probably a good reason for this, could we maybe document that?

Wumpf (approved):
so the bug was some timing issue in the command where toggle may cause issues somehow? Bit confused about why toggle doesn't work here. Also, isn't the play button changing its icon to pause... well clearly the code here is a pure play button as it always uses the `icons::PLAY` icon

IsseW:
> right so actually we could pause with the play button here, that was actually nice. Do we have to break this?

Yeah, I made that change some time ago. Before which it just switched to playing mode.

We could if we want to have a check like

```
if not playing mode {
	go to playing mode
} else {
	go to paused mode
}
```

IsseW:
> Bit confused about why toggle doesn't work here.

Toggle switches back to the last playing mode it seems, so if we were following last it switches to that.

aedm:
> Toggle switches back to the last playing mode it seems, so if we were following last it switches to that.

If we don't want that, we might fix it where it happens: https://github.com/rerun-io/rerun/blob/aa97658f1f09d66b8fa202eaa06b95b5ba045d99/crates/viewer/re_viewer_context/src/time_control.rs#L1085

IsseW:
> > Toggle switches back to the last playing mode it seems, so if we were following last it switches to that.
> 
> If we don't want that, we might fix it where it happens:
> 
> https://github.com/rerun-io/rerun/blob/aa97658f1f09d66b8fa202eaa06b95b5ba045d99/crates/viewer/re_viewer_context/src/time_control.rs#L1085

I think we do want that. When pressing space I think it makes sense to switch back to following at least

PR #11610, opened 2025-10-21
Prevent mutual access to a `Harness` in `TestContext`

grtlr:
### Related

* Follow-up of #11572.

### What

I ran into a weird problem where I accidentally had two active `Harness`es for a single `TestContext`.

The inputs of the two harnesses seemed to interact with each other. I've changed `TestContext` to require `&mut` to create a `HarnessBuilder`. This should prevent the probem in the future.

This PR changes the definitions in `crates/viewer/re_test_context/src/lib.rs`. The rest is following the compiler breadcrumbs.


grtlr:
You're right. And the multiple harnesses were not even the problem, which I just realized when creating a minimal reproducible example.

The problem was that the scrolling was applied twice, changing the view—and the view ofc outlives a given harness.

PR #11611, opened 2025-10-21
Pin `ipykernel` version for `rerun-notebook`

oxkitsune:
### What

This pins the version of `ipykernel` to `<7.0.0` for `rerun-notebook`. Version `7.0.0` introduced a bug (https://github.com/ipython/ipykernel/issues/1450) which prevents our embedded Viewer from receiving data that's been sent from a thread.

`7.1.0` [will include a fix for this,](https://github.com/ipython/ipykernel/pull/1451#issuecomment-3421859269) but until that's out we should pin the `ipykernel` to any version before `7.0.0`

### DIFF
--- a/rerun_notebook/pyproject.toml
+++ b/rerun_notebook/pyproject.toml
@@ -10,7 +10,7 @@ dependencies = [
   "anywidget",
   "jupyter-ui-poll",
   # ipykernel 7.0.0 has a bug that prevents it from sending display data from a different thread.
-# TODO(ipython/ipykernel/#1450)
+  # TODO(ipython/ipykernel/#1450)
   "ipykernel<7.0.0",
 ]
 readme = "README.md"


rerun_notebook/pyproject.toml
```diff
@@ -6,7 +6,13 @@ build-backend = "hatchling.build"
 name = "rerun-notebook"
 description = "Implementation helper for running rerun-sdk in notebooks"
 version = "0.27.0-alpha.1+dev"
-dependencies = ["anywidget", "jupyter-ui-poll"]
+dependencies = [
+  "anywidget",
+  "jupyter-ui-poll",
+  # ipykernel 7.0.0 has a bug that prevents it from sending display data from a different thread.
+  # https://github.com/ipython/ipykernel/issues/1450
```
ntjohnson1:
```suggestion
# TODO(ipython/ipykernel/#1450)
```

This should tag this into our zombie script so it's clearer when the root cause is resolved. (assuming I didn't mess up the format)

PR #11613, opened 2025-10-21
grpc tracing layer: add eos counter for the failure case

zehiko:
### What

I've noticed that for successful unary grpc requests we emit ``grpc_on_eos`` (as per code comment this happens for both streaming and unary requests), but for errors, we don't. Imho we should have it for both and ensure we can rely on this counter for both successful and failed requests. 

Added emitting this counter to the existing error handling path in ``on_response`` callback (thanks to @teh-cmc for going with a simpler approach rather than the original new ``on_failure`` callback)


teh-cmc:
How does this differ from the error handling in `on_response` (actual question)?

zehiko:
> How does this differ from the error handling in `on_response` (actual question)?

I have a disappointing answer - (for the practical purpose) I **believe** it doesn't. I ended up digging through the code and the docs on why ``on_eos`` callback doesn't get called in the failure case that it somehow slipped my mind I could add the counter to the existing error handling path. 

Let me simplify and get rid of ``on_failure``.

zehiko:
> > How does this differ from the error handling in `on_response` (actual question)?
> 
> I have a disappointing answer - (for the practical purpose) I **believe** it doesn't. I ended up digging through the code and the docs on why `on_eos` callback doesn't get called in the failure case that it somehow slipped my mind I could add the counter to the existing error handling path.
> 
> Let me simplify and get rid of `on_failure`.

now I do have a lingering question in my head - what is the use case for ``on_failure``?  (btw, I just looked into this middleware first time today, so maybe I'm missing some stuff). 
What happens if we have a midstream failure, will ``on_eos`` callback be called? Let me test this out quickly locally.

zehiko:
Posting a comment from the testing here as well:
* unary success - on_response and on_eos called
* unary failure - both on_response (error handling path) and on_failure called
* streaming success - on_response and on_eos called
* streaming immediate error (no stream started)  - on_response (error handling path) called and on_failure
* streaming mid stream error - on_response called (but only initially with OK code, no error detected here), on_eos called (and existing error handling logic executed). on_failure was not called.
looking at the code a bit, it seems that on_failure is called in immediate failure case, but also on polling frame error. I'm not sure if that should be transport errors. I wasn't able to simulate this (by having client panic midstream for example).

PR #11614, opened 2025-10-21
Update docs and migration guide notes on `protoc` requirements

abey79:
Small doc update cherry picked from the 0.26.1 branch.


Wumpf:
we should probably hold on for the patch release to go out so we can override via `deploy docs`

abey79:
> we should probably hold on for the patch release to go out so we can override via `deploy docs`

As discussed in slack, this is not necessary as these changes are already deployed.

image

PR #11616, opened 2025-10-21
Add timeout to GitHub Actions workflows

oxkitsune:
### Related

- https://github.com/rerun-io/rerun/actions/runs/18655380704/job/53191566210?pr=11426 

### What

Configures a time limit of 60 minutes for all workflows that take a bit longer on average to complete. This is done in order to avoid them from infinitely getting stuck.

### DIFF
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -18,7 +18,6 @@ on:
 
 jobs:
   label:
-    timeout-minutes: 60
     runs-on: ubuntu-latest
     steps:
       - name: Check for a "do-not-merge" label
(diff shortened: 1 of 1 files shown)


.github/workflows/cargo_shear.yml
```diff
@@ -9,6 +9,7 @@ on:
 
 jobs:
   cargo-shear:
+    timeout-minutes: 60
```
Wumpf, line 12: that would be interesting heh

oxkitsune: it would heh

PR #11619, opened 2025-10-22
Fixes hiding all tabs in a tab view spamming an undo warning

IsseW:
### What

Fixes a bug where hiding all the tabs in a tab view spammed an undo warning.

### Video of the issue

https://github.com/user-attachments/assets/b642c98a-87d0-4bdc-b699-b9e8bf5a819d

### Technical details

So there were actually two causes for this issue, first one being that we didn't clear the `active_tab` component when it's None. Meaning that queries would still pick up older values of the active tab.

The second issue was that if the active tab was none we were always assigning the tab to the first tab (something egui tiles already handles for us). And it was doing it without considering tab visibility.

### DIFF
--- a/crates/viewer/re_viewport_blueprint/src/container.rs
+++ b/crates/viewer/re_viewport_blueprint/src/container.rs
@@ -204,18 +204,12 @@ impl ContainerBlueprint {
             arch = arch.with_display_name(display_name.clone());
         }
 
-        // TODO(jleibs): The need for this pattern is annoying. Should codegen
-        // a version of this that can take an Option.
-        if let Some(active_tab) = &active_tab {
-            arch = arch.with_active_tab(&active_tab.as_entity_path());
-        } else {
-            // We have to clear the component if it's none, otherwise the query
-            // will still pick up older `active_tab` values.
-            ctx.clear_blueprint_component(
-                id.as_entity_path(),
-                re_types::blueprint::archetypes::ContainerBlueprint::descriptor_active_tab(),
-            );
-        }
+        // We want to write an empty array if `active_tab` is none. So can't use `arch.with_active_tab`
+        // here.
+        arch.active_tab = re_types::try_serialize_field::<re_types::blueprint::components::ActiveTab>(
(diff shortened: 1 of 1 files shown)


crates/viewer/re_viewport_blueprint/src/container.rs
```diff
@@ -208,6 +208,13 @@ impl ContainerBlueprint {
         // a version of this that can take an Option.
         if let Some(active_tab) = &active_tab {
             arch = arch.with_active_tab(&active_tab.as_entity_path());
+        } else {
+            // We have to clear the component if it's none, otherwise the query
+            // will still pick up older `active_tab` values.
+            ctx.clear_blueprint_component(
+                id.as_entity_path(),
+                re_types::blueprint::archetypes::ContainerBlueprint::descriptor_active_tab(),
+            );
```
Wumpf: quite related to the old todo comment abouve  comment: we don't need a separate clear write here, but rather the archetype we write should have an empty array set on `active_tab`. The generated `clear_fields` method on archetypes does just that - would that one be the right thing to do here as starting point or is that too aggressive? Either way would be much nicer to still have everything happening by logging this one archetype

IsseW: > we don't need a separate clear write here, but rather the archetype we write should have an empty array set on active_tab.

Ahh that makes sense.

> The generated clear_fields method on archetypes does just that - would that one be the right thing to do here as starting point or is that too aggressive?

If I understand what you mean correctly that would clear all fields, and there is a comment about saying to explicitly not clear the name

```rs
        // Note: it's important to _not_ clear the `Name` component if `display_name` is set to
        // `None`, as we call this function with `ContainerBlueprint` recreated from `egui_tiles`,
        // which is lossy with custom names.
        if let Some(display_name) = display_name {
            arch = arch.with_display_name(display_name.clone());
        }
```

So I think clearing all would be too aggressive.


~~I think we can just pass the option to `with_active_tab` though?~~ whoops didn't notice that was in an auto generated file

IsseW: Did the serialization inline instead.

PR #11620, opened 2025-10-22
Integration tests for origin heuristics

aedm:
### Related

Part of #8948 

### What

- Adds automated tests for origin heuristics when adding an entity to a new view
- Removes `check_context_menu_suggested_origin.py` from release checklist

### DIFF
--- a/tests/rust/re_integration_test/tests/origin_heuristics_test.rs
+++ b/tests/rust/re_integration_test/tests/origin_heuristics_test.rs
@@ -1,3 +1,8 @@
+//! Tests for the origin heuristics feature in the viewer.
+//!
+//! The origin heuristics feature is used to automatically set the view origin
+//! for views based on the data in the log.
+
 use egui_kittest::kittest::Queryable as _;
 use re_integration_test::HarnessExt as _;
 use re_sdk::TimePoint;


tests/rust/re_integration_test/tests/origin_heuristics_test.rs
```diff
@@ -0,0 +1,246 @@
...
+use re_viewer::viewer_test_utils::{self, HarnessOptions};
+use re_viewport_blueprint::ViewBlueprint;
+
+fn make_test_image() -> re_types::archetypes::Image {
+    let image = ndarray::Array3::from_shape_fn((256, 256, 3), |(y, x, c)| match c {
+        0 => x as u8,
+        1 => y as u8,
+        2 => 128,
+        _ => unreachable!(),
+    });
```
emilk, line 21: Image

tests/rust/re_integration_test/tests/origin_heuristics_test.rs
```diff
@@ -0,0 +1,246 @@
...
+pub async fn test_origin_image_3d() {
+    let mut harness = make_test_harness();
+
+    harness.blueprint_tree().right_click_label("3D view");
+    harness.click_label("Expand all");
+
+    // Close the selection panel before clicking on the image entity
+    // because rendering its download button requires the UI to be on
+    // the main thread. We are in a tokio test and it will crash.
+    harness.set_selection_panel_opened(false);
```
emilk, line 150: What is this about? Is this related to `MainThreadToken`? Do we really need to be on main thread to _render_ the button? I would assume we would only need it to process the save-action (i.e. handle clicking it)!

What happens if you remove this line? Does something crash?

tests/rust/re_integration_test/tests/origin_heuristics_test.rs
```diff
@@ -0,0 +1,246 @@
+use egui_kittest::kittest::Queryable as _;
```
emilk, line 6: Should have a module-level comment explaining roughly what this test is all about

emilk (approved):

tests/rust/re_integration_test/tests/origin_heuristics_test.rs
aedm: Yes, we need the `MainThreadToken` to render the button and the test will crash.
https://github.com/rerun-io/rerun/blob/22c37e8bf76f7ffb737021aecfc788daa254e0a3/crates/viewer/re_data_ui/src/instance_path.rs#L393

Issue #4818, opened 2024-01-15
TimeSeriesView Archetype


nikolausWest:
Add a (blueprint) archetype for the TimeSeriesView to hold the relevant options.

Example options that should be settable per view that aren't currently:
- axis / zoom / grid options
  - [ ] add axis label
  - [ ] hide axis
  - [ ] secondary Y axis
  - [ ] set Y axis bounds
  - [ ] axis tick options
  - [ ] hide grid
  - [ ] grid options (min / max size?)
  - [ ] allow zoom / drag for x / y
- legend 
  - [ ] set custom name (instead of entity name)
  - [ ] ability to hide the legend

jleibs:
> allow zoom / drag for x / y

@nikolausWest can you elaborate on what you mean by this? We can already zoom and drag to pan x / y so I assume this is something different?

nikolausWest:
> can you elaborate on what you mean by this? We can already zoom and drag to pan x / y so I assume this is something different
I meant making those things optional. In the egui.rs plots demo you can turn zooming and panning on and of per axis.

PR #11621, opened 2025-10-22
Expose `TimeSeries` time range & zoom in the Blueprint API

### LINKED ISSUE #8050: Allow configuring time series plot time axis from blueprint
Currently the time axis plot is ephemeral state that lives entirely in `egui_plot`. I.e. users can not set it from blueprint api and any left/right scrolling that has been manually is not stored in the blueprint state, getting lost on restart.

There *is* an "automatic" state that covers a lot of usecases already though: the time axis will adjust to the visible time range which determines the data that is queried. This should be the base `default provider` for this property!

Note also, that this auto-behavior doesn't always work as desired today: egui_plot zooms in on the available data rather than a defined range. This means that if your visible time range goes from e.g. 100 to 500 but your data only starts at 200, then the shown range becomes 200 to 500 when it should actually be 200 to 500

```
table TimeAxis (
    "attr.rerun.scope": "blueprint",
    "attr.rust.derive": "Default"
) {
    /// The range of the axis.
    ///
    /// If unset, the range well be automatically determined by the visible time range of the view.
    range: rerun.components.Range1D ("attr.rerun.component_optional", nullable, order: 2100);

    /// If enabled, the time axis range will remain locked to the specified range when zooming.
    zoom_lock: rerun.blueprint.components.LockRangeDuringZoom ("attr.rerun.component_optional", nullable, order: 2200);
}
```

Related to:
* https://github.com/rerun-io/rerun/issues/6265
* https://github.com/rerun-io/rerun/issues/7931

### LINKED ISSUE #11260: Give Time Series Views more follow/update options during playback
**Problem**
If I have very long recordings, my time series views can look like this:

Image

If I zoom in to get a better view of my data:

Image

Then soon enough, the time scrubber will run out of view:

Image

_one frame before the scrubber leaves the screen_

I have been unable to find anything in the selection panel to change the "follow" mode when zoomed into a chart. "Visible time range" just affects what's in view:

Image

Image

**Proposal**
Introduce two (blueprintified) ways to follow in charts:
1. No follow (or fixed positioning). Current behavior, where the scrubber will move out of the screen
2. Follow (or relative positioning). Provide a scrubber offset, defaulted to 0s. Position the scrubber always at the offset and move the chart behind it instead.

IsseW:
### Related

- Part of #4818.
- Closes #8050.
- Closes #11260.

- Moving the view with a trackpad has issues, see https://github.com/emilk/egui/issues/7657 and RR-2730. But this was a pre-existing issue. Temporarily fixed by https://github.com/rerun-io/rerun/pull/11627

### What

Adds a time range and zoom lock components to the time series view blueprint. Allowing you to change the view via the python API.

- [x] The UI code for the time range component used here more or less duplicates what is used for `VisibleTimeRange`. Looks difficult to deduplicate, but maybe look into it more?


crates/store/re_types/definitions/rerun/blueprint/archetypes/time_axis.fbs
```diff
@@ -1,15 +1,23 @@
 namespace rerun.blueprint.archetypes;
 
-
 // ---
 
-/// Configuration for the time (Y) axis of a plot.
+/// Configuration for the time (X) axis of a plot.
```
emilk, line 5: 🤦 

(I'm like 99% sure that was my typo)

crates/viewer/re_component_ui/src/time_range.rs
```diff
@@ -0,0 +1,332 @@
...
+            // see note above
+            let low_bound_override = if low_bound {
+                Some(re_log_types::TimeInt::MIN)
+            } else {
+                Some((other_boundary_absolute - current_time).into())
+            };
+
+            let mut edit_value = (*value).into();
+            time_drag_value
+                        .drag_value_ui(
```
emilk: weird indentation

emilk (dismissed):
Linking together the time axis of two plots is now broken 🤔 

Image

IsseW:
> Linking together the time axis of two plots is now broken 🤔

🤔  I thought I made sure it worked, might've broken it in a rebase

crates/viewer/re_component_ui/src/time_range.rs
IsseW: huh yeah, `cargo fmt` doesn't like those strings apparently

IsseW:
These snapshots updates are nice to not have to switch away from the branch I'm working on. But having to run them multiple times isn't ideal :/

emilk:
LinkToGlobal now works, but after enabling it double-clicking to reset the view doesn't work

crates/store/re_log_types/src/index/time_type.rs
```diff
@@ -240,8 +240,8 @@ mod tests {
     fn test_format_parse() {
         let cases = [
             (TimeInt::STATIC, "<static>"),
-            (TimeInt::MIN, "−∞"),
-            (TimeInt::MAX, "+∞"),
+            (TimeInt::MIN, "beginning"),
+            (TimeInt::MAX, "end"),
```
Wumpf, line 244: but we can still parse the old thing, so kinda strange that it falls out of the test here

crates/store/re_types/definitions/rerun/blueprint/components/time_range.fbs
```diff
@@ -0,0 +1,13 @@
+namespace rerun.blueprint.components;
+
+// ---
+
+/// A time range.
```
Wumpf:
```suggestion
/// A time range on an unspecified timeline using either relative or absolute boundaries.
```
.. right?

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -393,6 +417,70 @@ impl ViewClass for TimeSeriesView {
...
+
+        // Get the maximum time/X value for the entire plot
+        let max_time = all_plot_series
+            .iter()
+            .map(|line| line.points.last().map(|(t, _)| *t).unwrap_or(line.min_time))
+            .max()
+            .unwrap_or(0);
+
+        let timeline_start = ctx
+            .recording()
```
Wumpf: nit: we do a lot of `ctx.recording()` in this file, contributing to these call-towers. Let's pull it out to condense the code a lil bit :)

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -393,6 +417,70 @@ impl ViewClass for TimeSeriesView {
 
         let state = state.downcast_mut::<TimeSeriesViewState>()?;
 
+        let line_series = system_output.view_systems.get::<SeriesLinesSystem>()?;
```
Wumpf, line 442: this class' ui method has become _even_ more scary and hard to follow. Any chance we can split it up a little bit? E.g. have things like reset live in its own method?

Wumpf:
we clearly need a snapshot test for that time range ui.
Maybe if we split it out anyways, then we could do an isolated "pure" kittest test on it (no viewer test context needed)!

IsseW:
Link to global should only be on time axis.

crates/store/re_types/definitions/rerun/blueprint/archetypes/time_axis.fbs
```diff
@@ -1,15 +1,23 @@
...
 table TimeAxis (
     "attr.rerun.scope": "blueprint",
     "attr.rust.derive": "Default"
 ) {
     // --- Optional ---
 
     /// How should the horizontal/X/time axis be linked across multiple plots?
-    link: rerun.blueprint.components.LinkAxis ("attr.rerun.component_optional", nullable, order: 2200);
+    ///
+    /// Linking with global will ignore all the other options.
```
Wumpf:
```suggestion
    /// Linking with global will ignore `view_range`.
```

crates/store/re_log_types/src/index/time_type.rs
IsseW: Hmm maybe, the test checks both that the value formatted equals the given string. And that parsing the given string equals the value though.

IsseW:
> we clearly need a snapshot test for that time range ui.
Maybe if we split it out anyways, then we could do an isolated "pure" kittest test on it (no viewer test context needed)!

Are there any specific circumstances you have in mind to test?

crates/viewer/re_view_time_series/src/view_class.rs
IsseW: Split it up a bit

Wumpf:
a few combinations of different time ranges (closed/open intervals) and then small screenshots of the ui. Nothing too specific

IsseW:
Added snapshot tests that sets the time axis view range to a few different things

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -211,6 +193,46 @@ impl ViewClass for TimeSeriesView {
...
+                                re_types::datatypes::TimeRangeBoundary::Absolute(
+                                    s.max_time_view_range.max.into(),
+                                )
+                            },
+                        })
+                    })
+                    .unwrap_or(re_types::blueprint::components::TimeRange(TimeRange {
+                        start: re_types::datatypes::TimeRangeBoundary::Infinite,
+                        end: re_types::datatypes::TimeRangeBoundary::Infinite,
+                    }))
```
Wumpf, line 233: ah that's where the global view id would also end up in 💡

crates/viewer/re_view/src/view_property_ui.rs
```diff
@@ -16,10 +16,54 @@ pub fn view_property_ui<A: Archetype + ArchetypeReflectionMarker>(
...
     let view_property =
         ViewProperty::from_archetype::<A>(ctx.blueprint_db(), ctx.blueprint_query(), ctx.view_id);
-    view_property_ui_impl(ctx, ui, &view_property);
+    view_property_ui_impl(ctx, ui, &view_property, None);
 }
 
-fn view_property_ui_impl(ctx: &ViewContext<'_>, ui: &mut egui::Ui, property: &ViewProperty) {
+/// See [`view_property_ui`].
+///
+/// This will override
```
Wumpf: plz finish that comment ;)

crates/viewer/re_view/src/view_property_ui.rs
```diff
@@ -16,10 +16,54 @@ pub fn view_property_ui<A: Archetype + ArchetypeReflectionMarker>(
...
+pub fn view_property_ui_with_override<A: Archetype + ArchetypeReflectionMarker>(
+    ctx: &ViewContext<'_>,
+    ui: &mut egui::Ui,
+    override_component: ComponentIdentifier,
+    override_with_view_id: re_viewer_context::ViewId,
```
Wumpf: [not actionable] suspiciously specific to a single use case, but I appreciate the pragmatism :D

crates/viewer/re_view/src/view_property_ui.rs
Wumpf: use of the  term "override" is a bit audacious given that we already attached so much meaning to it elsewhere. Let's maybe go with "redirected" instead?

crates/viewer/re_selection_panel/src/visible_time_range_ui.rs
```diff
@@ -433,163 +404,3 @@ fn resolved_visible_history_boundary_ui(
 
     ui.label(label);
 }
-
```
Wumpf, line 436: yay, less code dupl! \o/

Wumpf (approved):
yay, happy to see that this works really well now!
tbh I lost a bit of track of what changed and what not, so I probably missed a few thigns in this review pass. Which goes to say we should move on and rather split fix things in upcoming iterations given that everything seems to be fine now

I still think we should add tests specifically for what is now `re_ui::RelativeTimeRange::ui` (yay, great win in re-use!), but it's probably better to take that as a separate iteration

Issue #11625, opened 2025-10-22
Deduplicate `reusable_publish_js`, `reusable_build_js` and `reusable_upload_js`


abey79:
We currently have two separate paths for JS publishing.

PR workflow goes through:
- `reusable_build_js.yml`
- `reusable_upload_js.yml`

Release workflow goes through:
- `reusable_publish_js.yml`

They are practically identical, with the difference being where on the build bucket the artefacts are uploaded.

On that note, the asset sync workflow expects everything to be in the `commit/$RELEASE_COMMIT/*` directory, but until recently `reusable_publish_js.yml` would not upload the artefacts there. So asset sync would fail unless the PR workflow would succeed before. This requires a mergeable PR, which is not always the case for e.g. patch release.

PR #11626, opened 2025-10-22
Fix js release

abey79:
Short term patch for an issue with asset sync encountered while releasing 0.26.1.

Better solution and background detiails:
- https://github.com/rerun-io/rerun/issues/11625

### DIFF
--- a/.github/workflows/reusable_publish_js.yml
+++ b/.github/workflows/reusable_publish_js.yml
@@ -92,7 +92,7 @@ jobs:
         uses: google-github-actions/upload-cloud-storage@v2
         with:
           path: "rerun_js_package"
-          destination: "rerun-builds/commit/${{ steps.get-commit-sha.outputs.short-sha }}/rerun_js"
+          destination: "rerun-builds/commit/${{ needs.get-commit-sha.outputs.short-sha }}/rerun_js"
           parent: false
           process_gcloudignore: false
           headers: |-


.github/workflows/reusable_publish_js.yml
```diff
@@ -86,3 +86,15 @@ jobs:
           headers: |-
             cache-control: no-cache, max-age=0
           gzip: false
+
+      # TODO(#11625): we need this for asset sync, in case the PR workflow fails (or doesn't run). See issue for details.
+      - name: "Upload rerun_js to GCS (commit)"
+        uses: google-github-actions/upload-cloud-storage@v2
+        with:
+          path: "rerun_js_package"
+          destination: "rerun-builds/commit/${{ steps.get-sha.outputs.sha }}/rerun_js"
```
jprochazk:
```suggestion
          destination: "rerun-builds/commit/${{ steps.get-commit-sha.outputs.short-sha }}/rerun_js"
```

jprochazk: That step is called something else in this workflow

abey79: 🤦🏻

.github/workflows/reusable_publish_js.yml
```diff
@@ -86,3 +86,15 @@ jobs:
           headers: |-
             cache-control: no-cache, max-age=0
           gzip: false
+
+      # TODO(#11625): we need this for asset sync, in case the PR workflow fails (or doesn't run). See issue for details.
+      - name: "Upload rerun_js to GCS (commit)"
+        uses: google-github-actions/upload-cloud-storage@v2
+        with:
+          path: "rerun_js_package"
+          destination: "rerun-builds/commit/${{ steps.get-commit-sha.outputs.short-sha }}/rerun_js"
```
jprochazk: Actually I think it's even worse, you need to use `needs` instead of `steps`, because it's a separate job
```suggestion
          destination: "rerun-builds/commit/${{ needs.get-commit-sha.outputs.short-sha }}/rerun_js"
```

jprochazk: (applied this suggestion myself, hope you don't mind)

jprochazk: Example of it being used correctly:

https://github.com/rerun-io/rerun/blob/58fee3dd461ab600f24e34d095334f86bdb4d6e3/.github/workflows/reusable_publish_web.yml#L134

abey79: > (applied this suggestion myself, hope you don't mind)

absolutely not, sorry for the sloppiness

PR #11632, opened 2025-10-23
Check that all examples on website have a thumbnail

grtlr:
### What

Title. We use the same logic as in https://github.com/rerun-io/landing/blob/main/src/lib/lang.ts to determine the default example (from which we pull the thumbnail).

Also fixes existing problems:

* Ignore `lenses` example because we don't have meaningful screenshot.
* Add thumbnail for `animated_urdf` example.
* Add missing URDF tag to `urdf_loader`.

> [!IMPORTANT]
> We ignore the link to `lang.ts` in our lints, because if that file moves, we should check if the corresponding logic still matches to avoid drift between CI and reality.

### DIFF
--- a/scripts/ci/thumbnails.py
+++ b/scripts/ci/thumbnails.py
@@ -110,7 +110,8 @@ def check() -> None:
     ignored = load_ignored_examples()
     missing_frontmatter = []
     missing_thumbnails = []
-    bad_dimensions = []
+    missing_dimensions = []
+    incorrect_dimensions = []
 
     # NOTE: This language priority must match the logic in the rerun-io/landing repo:
     # Priority order: Python > C++ > Rust > C.
@@ -151,12 +152,12 @@ def check() -> None:
 
             # If there's a thumbnail, check dimensions
             if not example.fm.get("thumbnail_dimensions"):
-                bad_dimensions.append((example.path, None, None))
+                missing_dimensions.append(example.path)
             else:
                 current = tuple(example.fm["thumbnail_dimensions"])
                 actual = get_thumbnail_dimensions(example.fm["thumbnail"])
                 if current != actual:
-                    bad_dimensions.append((example.path, current, actual))
+                    incorrect_dimensions.append((example.path, current, actual))
                 else:
                     print(f"✔ {example.path}")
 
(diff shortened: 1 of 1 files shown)


scripts/ci/thumbnails.py
```diff
@@ -101,6 +108,39 @@ def work(example: Example) -> None:
...
+                break
+
+        if default_path:
+            readme = (default_path / "README.md").read_text(encoding="utf-8")
+            fm = load_frontmatter(readme)
+            if fm is None or not fm.get("thumbnail") or not fm.get("thumbnail_dimensions"):
+                print(f"{default_path} is missing `thumbnail` and/or `thumbnail_dimensions`")
+                bad = True
+
+    # Check that existing thumbnail dimensions are correct (all languages)
```
jprochazk: I know none of this is exactly performance critical, but why isn't the check incorporated into the existing one? I would:

- Change `examples_with_thumbnails` to not exclude examples without thumbnails
- If an example is in `ignored_examples`, skip it
- If an example has no thumbnail, report that error
- If an example has a thumbnail but has incorrect `thumbnail_dimensions`, report that error



I also find the resulting error message a bit misleading/imprecise. The error message will be something like:
```
examples/python/minimal is missing `thumbnail` and/or `thumbnail_dimensions`
...
Please run `scripts/ci/thumbnails.py update`.
```

But missing thumbnails won't be fixed automatically. I think we should only say something like that if there are wrong `thumbnail_dimensions` (so instead of one `bad` state, we have `missing_thumbnails` and `bad_thumbnail_dimensions` - this can still be done in one pass).

The result that I'd expect:

```
examples/python/animated_urdf is missing `thumbnail`
examples/python/minimal `thumbnail_dimensions` are incorrect (current: [100, 100], actual: [480, 480])

Some example are missing thumbnails. Either add those thumbnails, or add those examples to the `ignore` list in `examples/manifest.toml`.

Incorrect thumbnail dimensions may be fixed automatically by running `scripts/ci/thumbnails.py update`.
```

We should be kind to whoever has to deal with this check 🙂

jprochazk (changes requested):

scripts/ci/thumbnails.py
grtlr: That makes sense—thank you!

scripts/ci/thumbnails.py
```diff
@@ -93,38 +99,99 @@ def work(example: Example) -> None:
...
+
+    # Build a map of example names to their default language version
+    example_defaults = {}
+    for lang in ["python", "cpp", "rust", "c"]:
+        lang_dir = Path(f"examples/{lang}")
+        if lang_dir.exists():
+            for path in lang_dir.iterdir():
+                if path.is_dir() and (path / "README.md").exists():
+                    # Only set if we haven't seen this example before (priority order)
+                    if path.name not in example_defaults and lang in ["python", "cpp", "rust"]:
```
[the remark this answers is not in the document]
grtlr: In `lang.ts` we don't even consider `c`, but why not 🤷. Should we report an error, if there's only a `c` example?

PR #11635, opened 2025-10-23
Reproduction of the Japanese Alley example for Foxglove MCAP files

grtlr:
### What

This is a testbed for the new Lenses feature. We reproduce the Foxglove MCAP example: https://foxglove.dev/blog/using-the-foxglove-sdk-to-generate-mcap.

To run this example you need to follow the instructions above to produce an MCAP file, or just contact me.

> [!IMPORTANT]
> We should not merge this example until we have `tf`-style transforms. At that point we might want to turn this into a more general Foxglove visualizer/data loader.


MichaelGrupp:
Replaced by https://github.com/rerun-io/rerun/pull/11789, which got now merged.

PR #11637, opened 2025-10-23
Add shortcuts to switch between recordings

lucasmerlin:
### Related

* part of https://linear.app/rerun/issue/RR-2362/navigating-through-catalog-via-recordings-side-panel-with-hotkeys

### What

Adds global hotkeys to switch through everything in the recording panel (except for the homepage and anything without it's own page, e.g. app ids).

This was surprisingly deep, the implementation feels quite complex but is the best thing I could come up with. Its the result of the following things:
- shortcuts should be a `UiCommand` for discoverability (so it can be found in the command palette)
- recording switching order should be defined in the recording panel, since it should match the order of the shortcuts

=> We introduce a new `RecordingPanelCommand` that can be send from a `UiCommand` and executed in the context of the recording panel. The latter is now a struct to hold the command queue.

Note:
- We only shift through recordings in the current "goup" (aka dataset or app id)
- This feature overlaps a lot with the new tree navigation (arrow keys)....
- ...but could be made massively more useful if we attempted to synchronise the viewer state:
  - https://github.com/rerun-io/rerun/issues/11792

### DIFF
--- a/crates/viewer/re_ui/src/command.rs
+++ b/crates/viewer/re_ui/src/command.rs
@@ -1,5 +1,5 @@
-use egui::{os::OperatingSystem, Key, KeyboardShortcut, Modifiers};
-use smallvec::{smallvec, SmallVec};
+use egui::{Key, KeyboardShortcut, Modifiers, os::OperatingSystem};
+use smallvec::{SmallVec, smallvec};
 
 use crate::context_ext::ContextExt as _;
 


crates/viewer/re_ui/src/command.rs
```diff
@@ -366,6 +375,9 @@ impl UICommand {
             Self::CloseCurrentRecording => smallvec![],
             Self::CloseAllEntries => smallvec![],
 
+            Self::NextRecording => smallvec![cmd(Key::CloseBracket)],
+            Self::PreviousRecording => smallvec![cmd(Key::OpenBracket)],
```
lucasmerlin: There needs to be some bike shedding on these shortcuts. 
- @pweids suggested it should be cmd + `[` / `]`
- @lucasmerlin thinks it should be `ctrl + tab` / `ctrl + shift + tab` (which switches tabs in my browser, finder and recent files in my ide)
- @abey79 would like something different

crates/viewer/re_recording_panel/src/recording_panel_command.rs
```diff
@@ -0,0 +1,90 @@
...
+/// have access to that, so we need to handle these commands "within" the frame, where we have
+/// access to the context.
+#[derive(Clone, Debug)]
+pub enum RecordingPanelCommand {
+    /// Switch to the next recording in the recording panel.
+    SelectNextRecording,
+
+    /// Switch to the previous recording in the recording panel.
+    SelectPreviousRecording,
+}
```
abey79: @Wumpf I'd like to have an architectural chat about this thing (that we discussed earlier with Lucas)

crates/viewer/re_ui/src/command.rs
emilk: Note: we can support multiple keyboard shortcuts (a primary and a backup; that's why this is a vec).

on **macOS**:
in my browser, I can cycle tabs with (at least) `ctrl+[shift]+tab` plus `cmd+option+side-arrows` plus `ctrl+pageup/down`.

on **Windows and Linux**, I think ctrl+[shift]+tab is the common one for switching between open tabs in an application?

Based on this, I suggest ctrl+[shift]+tab as the primary on all platforms, with `cmd+option+down/up arrows` as secondary on macOS.

But I'd like to know where @pweids got `cmd + [` from? Is there some precedence there?

abey79: ctrl-tab is application switching in Windows, isn't it? I guess we could check what's the browser tab shortcut there.

I'm not sure where `cmd + [` comes from, but I don't like it because it's not intl keyboard layout friendly. I need to do `alt + 5` to get a `[` on my swiss keyboard layout. (I guess never migrating to the US layout is top-3 reason I'll never be a 10x dev.)

emilk: No, in Windows Alt-Tab is _application_ switching (like cmd-tab in Mac)

abey79: Ah right, my bad!

gavrelina:
had a huddle with @abey79: 

we have now 2 ways going between recordings, by ↑ and ↓ and CMD+[ /CMD+] . 
the cmd+[ and cmd+] behaves weird, it goes through all the items even if they are not recordings. 
We decided on 2 things: 
1. Do not go to next dataset with cmd[/cmd], but loop through the ones belonging to current dataset. (To simply go through recordings panel one may use arrows, which is easier)

2. What is the added value of even having Cmd[ and cmd]? It adds value if the selection is "preserved" whilst switching to another recording, else it's repetitive and cumbersome. 

With this we keep it, and add a separate task for keeping a selection.

lucasmerlin:
I've changed the shortcuts to `cmd alt up / cmd alt down`. Ctrl tab conflicts with eguis tab handling to focus next / previous widget. We can change it to tab once we have https://github.com/emilk/egui/issues/7659.


> What is the added value of even having Cmd[ and cmd]? It adds value if the selection is "preserved" whilst switching to another recording, else it's repetitive and cumbersome.

It also adds value since this is a global shortcut that will also work when focus is somewhere else. Up/down will only work when the focus is in the recording panel.

PR #11648, opened 2025-10-24
Fix `concat_polymorphic_batches` discarding top-level metadata

teh-cmc:
This is our good old testing helper for "please take all of this and somehow turn it into a single batch that I can snapshot pls pls". It was dropping top-level metadata since forever, and we've never noticed because we've never really checked for top-level metadata in our tests. That's because we only cared about field-level metadata until now, but that is changing now that we're introducing dataset-level metadata in Redap!


teh-cmc:
@rerun-bot full-check (can't be too careful :no_mouth:)

PR #11646, opened 2025-10-24
Fix color picker on default & override disappearing when trying to edit

Wumpf:
### Related

* Fixes RR-2761

### What

Color picker on `Defaults` and the expanded override menu would disappear before:

https://github.com/user-attachments/assets/1aaccb1d-5a17-4c93-8413-234d8b7a1dca

This fixes this!

PR #11651, opened 2025-10-24
Fix color picker on default & override disappearing when trying to edit

emilk:
### Related

* Fixes RR-2761
* Closes https://github.com/rerun-io/rerun/pull/11646

### What
Creating a child ui advances the "auto-id" of the parent ui. So if the button goes from visible to invisible, that changes the `id` of e.g. the color picker widget


emilk:
I don't understand why CI fails, when the test passes locally 🤔

emilk:
The failing snapshot doesn't even have list-item stuff in it O.o

`crates/viewer/re_view_spatial/tests/snapshots/annotations_hover_region_red.png`:
image


EDIT: oh, the "new" snapshot created by the CI is… totally blank 🕵️

PR #11640, opened 2025-10-23
Create lance tables in the cloud

timsaucer:
### Related

This is a follow on to https://github.com/rerun-io/rerun/pull/11565

It addresses a portion of https://linear.app/rerun/project/write-tables-to-rerun-cloud-via-the-python-sdk-432c0243348f/overview

### What

With this PR we introduce a new server API to create a table from a given Schema. The current work will create a Lance based table. This table is registered with the server and available for immediate writing using #11565 

Additionally this PR adds the integration testing in `re_redap_tests`

PR #11657, opened 2025-10-24
Add client object table writing functions

timsaucer:
### Related

This is a follow on to https://github.com/rerun-io/rerun/pull/11565
This is part of https://linear.app/rerun/project/write-tables-to-rerun-cloud-via-the-python-sdk-432c0243348f/overview

### What

This PR adds two methods to the client interface in Python:

`write_table` takes in a table name and record batches and writes them to a table. It contains convenience functions to accept either a record batch, iterable, iterable of iterable, or a record batch reader. Additionally you can set it to append or overwrite the table.

`append_to_table` is a convenience function that takes Python objects and turns them into a row or rows of data. It uses pyarrow for type casting of the data into arrow arrays. Then it composes a record batch and writes it to the server by calling `write_table`.

### DIFF
--- a/rerun_py/src/catalog/connection_handle.rs
+++ b/rerun_py/src/catalog/connection_handle.rs
@@ -5,7 +5,6 @@ use arrow::pyarrow::PyArrowType;
 use pyo3::exceptions::PyValueError;
 use pyo3::{PyErr, PyResult, Python};
 use std::collections::BTreeSet;
-use std::sync::Arc;
 use tracing::Instrument as _;
 
 use crate::catalog::table_entry::PyTableInsertMode;
@@ -265,7 +264,6 @@ impl ConnectionHandle {
                     .await
                     .map_err(to_py_err)?
                     .ok_or(PyValueError::new_err("Unable to find EntryId for table"))?;
-                let schema = stream.schema();
 
                 // Since the errors occur during streaming, we cannot let this method
                 // fail without doing a collect operation. Instead, we log a warning to
--- a/rerun_py/tests/e2e_redap_tests/test_datafusion_tables.py
+++ b/rerun_py/tests/e2e_redap_tests/test_datafusion_tables.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-import pathlib
-import tempfile
 from typing import TYPE_CHECKING
 
 import pyarrow as pa
@@ -270,49 +268,6 @@ def test_datafusion_write_table(server_instance: ServerInstance) -> None:
     # Verify overwrite mode
     df_smaller.write_table(table_name, write_options=DataFrameWriteOptions(insert_operation=InsertOp.OVERWRITE))
     assert ctx.table(table_name).count() == smaller_count
-
-
-def test_create_table(server_instance: ServerInstance) -> None:
-    table_name = "created_table"
-
-    original_schema = pa.schema([("int64", pa.int64()), ("float32", pa.float32()), ("utf8", pa.utf8())])
-
-    with tempfile.TemporaryDirectory() as temp_dir:
-        temp_path = pathlib.Path(temp_dir).as_uri()
-
-        table_entry = server_instance.client.create_table_entry(table_name, original_schema, temp_path)
-        df = table_entry.df()
-
-        returned_schema = df.schema().remove_metadata()
-
-        assert returned_schema == original_schema
-
-
-def test_create_table_from_dataset(server_instance: ServerInstance) -> None:
-    table_name = "dataset_to_table"
-
-    df = server_instance.dataset.dataframe_query_view(index="time_1", contents="/**").df()
-    original_schema = df.schema()
-
-    with tempfile.TemporaryDirectory() as temp_dir:
-        temp_path = pathlib.Path(temp_dir).as_uri()
-
-        table_entry = server_instance.client.create_table_entry(table_name, original_schema, temp_path)
-        df = table_entry.df()
-
-        # Due to https://github.com/lancedb/lance/issues/2304 we cannot
-        # directly compare the returned schema. Verify we at least
-        # get back the same columns and metadata
-
-        returned_schema = df.schema()
-        for field in returned_schema:
-            assert original_schema.field(field.name) is not None
-        for field in original_schema:
-            assert returned_schema.field(field.name) is not None
-
-        for returned_field in returned_schema:
-            original_field = original_schema.field(returned_field.name)
-            assert returned_field.metadata == original_field.metadata
 
 
 def test_client_write_table(server_instance: ServerInstance) -> None:
--- a/rerun_py/tests/e2e_redap_tests/test_create_table.py
+++ b/rerun_py/tests/e2e_redap_tests/test_create_table.py
@@ -1,0 +1,53 @@
+from __future__ import annotations
+
+import pathlib
+import tempfile
+from typing import TYPE_CHECKING
+
+import pyarrow as pa
+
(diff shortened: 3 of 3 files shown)


rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -181,6 +185,123 @@ def register_table(self, name: str, url: str) -> TableEntry:
         """
         return self._raw_client.register_table(name, url)
 
+    def write_table(
+        self,
+        name: str,
+        batches: RecordBatchReader | RecordBatch | Sequence[RecordBatch] | Sequence[Sequence[RecordBatch]],
```
ntjohnson1, line 210: Is there a reason we don't also accepted a Dataframe directly and just call to_batches on it?

rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -181,6 +185,123 @@ def register_table(self, name: str, url: str) -> TableEntry:
...
+        insert_mode: TableInsertMode,
+    ) -> None:
+        """
+        Writes record batches into an existing table.
```
ntjohnson1, line 214: How to create table here https://github.com/rerun-io/rerun/pull/11640/files

rerun_py/rerun_sdk/rerun/catalog.py
```diff
@@ -181,6 +185,123 @@ def register_table(self, name: str, url: str) -> TableEntry:
...
+        name
+            The name of the table entry to write to. This table must already exist.
+
+        batches
+            One or more record batches to write into the table. For convenience, you can
+            pass in a record batch, list of record batches, list of list of batches, or
+            a [`pyarrow.RecordBatchReader`].
+
+        insert_mode
+            Determines how rows should be added to the existing table.
```
ntjohnson1, line 227: It's a little weird to me that Overwrite mode doesn't just create the table or that there isn't a create mode.

ntjohnson1:
Sorry for double note. I realized I didn't add this initially. Not a blocker here but are we able to view a table as an rrd if it has the right meta data or is that still dependent on @abey79 previous hack/demo?

rerun_py/rerun_sdk/rerun/catalog.py
timsaucer: We could do that. My goal with these was to be the non-datafusion users support. For dataframe, you don't need to do this at all you can just do `df.write_table("table_name")`

rerun_py/rerun_sdk/rerun/catalog.py
timsaucer: Topic for standup today is where to put those files. The reason create table is different is because it takes a storage location of where to create the files.

timsaucer:
> Sorry for double note. I realized I didn't add this initially. Not a blocker here but are we able to view a table as an rrd if it has the right meta data or is that still dependent on @abey79 previous hack/demo?

This still requires his demo.

rerun_py/rerun_sdk/rerun/catalog.py
ntjohnson1: Gotcha. I hadn't made the connection for your note below that the creation of the table is when we specify the storage location so that the newly added `write_table` in datafusion just works already.

rerun_py/rerun_sdk/rerun/catalog.py
bllchmbrs: Tim, these are fairly standard write semantics, right? 
As in, this isn't an area of innovation for us?

timsaucer: Correct.

ntjohnson1 (approved):
The python APIs LGTM. Nothing in the connection client jumped out as wild to me but if you're worried about those bits maybe get someone else to look at it or I can circle back to better understand what happens there later this afternoon.

crates/store/re_redap_client/src/connection_client.rs
```diff
@@ -695,6 +695,32 @@ where
         Ok(response)
     }
 
+    pub async fn get_entry_id(
+        &mut self,
+        entry_name: &str,
+        entry_kind: Option<EntryKind>,
+    ) -> Result<Option<EntryId>, ApiError> {
+        self.inner()
```
abey79, line 703: that would have been a very nice `ConnectionClient` helper ;)

rerun_py/src/catalog/catalog_client.rs
```diff
@@ -365,6 +368,22 @@ impl PyCatalogClientInternal {
...
+    ) -> PyResult<()> {
+        let connection = self_.borrow_mut(py).connection.clone();
+
+        let stream = ArrowArrayStreamReader::from_pyarrow_bound(batches)?;
+
+        connection.write_table(py, name, stream, insert_mode)?;
+
+        Ok(())
+    }
+
```
abey79, line 386: beautiful example of why the pure python API / internal object pattern is great

rerun_py/src/catalog/connection_handle.rs
```diff
@@ -244,6 +246,43 @@ impl ConnectionHandle {
...
+                    .map_err(to_py_err)?
+                    .ok_or(PyValueError::new_err("Unable to find EntryId for table"))?;
+                let schema = stream.schema();
+
+                // TODO(tsaucer) this is not ideal because it is dropping errors that originate in
+                // pyarrow. Since the errors occur during streaming, we cannot let this method
+                // fail without doing a collect operation.
+                let stream = futures::stream::iter(
+                    stream.map(move |rb| rb.unwrap_or(RecordBatch::new_empty(Arc::clone(&schema)))),
+                );
```
abey79: could we at least log a warning when that happens?

rerun_py/tests/e2e_redap_tests/test_datafusion_tables.py
```diff
@@ -313,3 +313,66 @@ def test_create_table_from_dataset(server_instance: ServerInstance) -> None:
         for returned_field in returned_schema:
             original_field = original_schema.field(returned_field.name)
             assert returned_field.metadata == original_field.metadata
+
+
+def test_client_write_table(server_instance: ServerInstance) -> None:
```
abey79, line 273: I think i'd rather have these in a dedicated file, maybe `test_write_table.py`. They don't have much to do with datafusion, and it's nice if the file more or less maps to a feature name and to a grpc endpoint. Far from always possible, but when it works its nice. (I try to do the same in `re_redap_tests`)

abey79 (approved):
Nice work!

crates/store/re_redap_client/src/connection_client.rs
timsaucer: I'm sorry - I don't understand this comment. Are you suggesting I add something different?

abey79: actually, i must be too tired. i meant it should have been a `ConnectionClient` helper and didn't see it already was one 🤦🏻  Basically, `inner` -> pitchfork reflex reaction.

rerun_py/src/catalog/connection_handle.rs
timsaucer: Yes, I switched it to log during the stream

rerun_py/tests/e2e_redap_tests/test_datafusion_tables.py
timsaucer: moved in latest push

PR #11658, opened 2025-10-25
When scrubbing time, intelligently round to a nice value

emilk:
### Related
* Part of https://linear.app/rerun/issue/RR-2768/improve-time-selection-experience-on-a-timeline

### What
When dragging the time cursor ("time scrubbing") we would previously pick arbitrarily precise time points, regardless of zoom level. Now, we snap to the "simplest" value that is within a pixel from the mouse cursor. The simplest value is defined as the one with _the most_ trailing zeroes (e.g. prefer `23.500000` over `23.484793`).

### Before
Even though we zoomed out to where we are viewing full hours, when we try to pick a time we select with microsecond precision:

dumb-aim

### After
Notice how we start by picking whole multiples of 10s, but as we zoom in we pick multiples of 0.5s, and then 5ms, 10us, etc.

smart-aim

Note that we only ever omit _zeroes_ in the display, so this is a real difference in how values are picked, NOT how they are displayed!

### Future work
* Coarser snapping when a modifier key is pressed
* Set the time display precision based on zoom level

### DIFF
--- a/crates/store/re_log_types/src/index/absolute_time_range.rs
+++ b/crates/store/re_log_types/src/index/absolute_time_range.rs
@@ -310,7 +310,8 @@ impl AbsoluteTimeRangeF {
             }
         }
 
-        min.into() // All digits are the same
+        // All digits are the same. Already handled earlier, but:
+        min.into()
     }
 }
 


crates/store/re_log_types/src/index/absolute_time_range.rs
```diff
@@ -208,6 +213,109 @@ impl AbsoluteTimeRangeF {
...
+                }
+                return TimeInt::new_temporal(
+                    #[expect(clippy::unwrap_used)] // Cannot fail
+                    re_format::parse_i64(&result_str).unwrap(),
+                )
+                .into();
+            }
+        }
+
+        min.into() // All digits are the same
```
Wumpf: that should be impossible since we already checked that min and max are different, no?

crates/store/re_log_types/src/index/absolute_time_range.rs
```diff
@@ -208,6 +213,109 @@ impl AbsoluteTimeRangeF {
...
+    ///
+    /// The simplest number is defined approximately as
+    /// "the number with the most trailing zeroes".
+    ///
+    /// This is used in GUI code.
+    ///
+    /// NOTE: this function only cares about simple _integers_.
+    /// We don't care about "smart" selection things smaller than a nanosecond or a single time step.
+    /// So if both ends are within the same integer, then [`Self::center`] is returned.
+    pub fn smart_aim(self) -> TimeReal {
```
Wumpf, line 227: why is this returning a `TimeReal` and not a `TimeInt`? Seems like everything in here starts out with `TimeInt`  and then goes `into`

crates/store/re_log_types/src/index/absolute_time_range.rs
```diff
@@ -208,6 +213,109 @@ impl AbsoluteTimeRangeF {
...
+
+                let rest_of_min_is_zeroes = min_str.as_bytes()[i + 1..].iter().all(|&c| c == b'0');
+
+                if !rest_of_min_is_zeroes {
+                    // There are more digits coming after `deciding_digit_min`, so we cannot pick it.
+                    // So the true min of what we can pick is one greater:
+                    deciding_digit_min += 1;
+                }
+
+                // if deciding_digit_min == b'1' && rest_of_min_is_zeroes {
```
Wumpf: what's the history with why this special case is no longer in use?

Wumpf:
I like those round numbers, but what I find super irritating is that I can't click on the time of an event anymore (edit: could I ever though? 🤔 ) even if the hover tooltip pretend that I do. Can we snap to exact times if something was hovered?

Demonstrated here:

https://github.com/user-attachments/assets/bc3d0ada-8060-41d2-a4d2-f61385e056eb

crates/store/re_log_types/src/index/absolute_time_range.rs
emilk: For the case of `min=max` and that's not an integer

emilk:
> I find super irritating is that I can't click on the time of an event anymore (edit: could I ever though?)

No, you couldn't ever. Try it on rerun.io/viewer - it's the same behavior there. I'll add it to the parent issue, but it is completely orthogonal to this PR.

PR #11664, opened 2025-10-27
Add play state, loop mode and loop selection to blueprint

### LINKED ISSUE #1381: Ability to configure default playing & loop modes for a given app_id from the SDK
Currently, the viewer will always start in loop mode when opening a file, and in follow mode when streaming data in.

In most cases this heuristic holds up, but sometimes it does not and multiple users have asked for a way to control that. 

See e.g. this [discord thread](https://discord.com/channels/1062300748202921994/1078019891338150030).

IsseW:
### Related

- Closes #1381

### What

Adds the ability to control playback state (paused, following, playing), and loop selection from blueprint.

- [x] Should play-state be a static component?

### DIFF
--- a/crates/store/re_types/definitions/rerun/datatypes/time_range.fbs
+++ b/crates/store/re_types/definitions/rerun/datatypes/time_range.fbs
@@ -3,7 +3,8 @@ namespace rerun.datatypes;
 /// Two [datatypes.TimeInt] describing a range of time.
 struct AbsoluteTimeRange (
   "attr.rust.derive": "Copy, PartialEq, Eq, PartialOrd, Ord",
-  "attr.rust.override_crate": "re_types_core"
+  "attr.rust.override_crate": "re_types_core",
+  "attr.docs.unreleased"
 ) {
   /// Start of the range.
   min: TimeInt (order: 100);
--- a/docs/content/reference/types/datatypes/absolute_time_range.md
+++ b/docs/content/reference/types/datatypes/absolute_time_range.md
@@ -26,8 +26,8 @@ Struct {
 ```
 
 ## API reference links
- * 🌊 [C++ API docs for `AbsoluteTimeRange`](https://ref.rerun.io/docs/cpp/stable/structrerun_1_1datatypes_1_1AbsoluteTimeRange.html)
- * 🐍 [Python API docs for `AbsoluteTimeRange`](https://ref.rerun.io/docs/python/stable/common/datatypes#rerun.datatypes.AbsoluteTimeRange)
- * 🦀 [Rust API docs for `AbsoluteTimeRange`](https://docs.rs/rerun/latest/rerun/datatypes/struct.AbsoluteTimeRange.html)
+ * 🌊 [C++ API docs for `AbsoluteTimeRange`](https://ref.rerun.io/docs/cpp/stable/structrerun_1_1datatypes_1_1AbsoluteTimeRange.html?speculative-link)
+ * 🐍 [Python API docs for `AbsoluteTimeRange`](https://ref.rerun.io/docs/python/stable/common/datatypes?speculative-link#rerun.datatypes.AbsoluteTimeRange)
+ * 🦀 [Rust API docs for `AbsoluteTimeRange`](https://docs.rs/rerun/latest/rerun/datatypes/struct.AbsoluteTimeRange.html?speculative-link)
 
 
--- a/docs/content/reference/types/datatypes/time_int.md
+++ b/docs/content/reference/types/datatypes/time_int.md
@@ -19,6 +19,6 @@ int64
 
 ## Used by
 
-* [`AbsoluteTimeRange`](../datatypes/absolute_time_range.md)
+* [`AbsoluteTimeRange`](../datatypes/absolute_time_range.md?speculative-link)
 * [`TimeRangeBoundary`](../datatypes/time_range_boundary.md)
 * [`Timestamp`](../components/timestamp.md)


crates/viewer/re_viewer/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.blueprint.archetypes.TimePanelBlueprint.snap
```diff
@@ -7,3 +7,6 @@ timeline: [log_time]
 time: [0]
 playback_speed: [0.0]
 fps: [0.0]
+play_state: [2]
+loop_mode: [1]
+time_selection: [{min: 0, max: 0}]
```
Wumpf, line 12: slightly confusing on enums but still love it :3

crates/store/re_types/definitions/rerun/blueprint/components/play_state.fbs
```diff
@@ -0,0 +1,23 @@
...
+
+/// The current play state.
+enum PlayState: ubyte (
+  "attr.rerun.scope": "blueprint",
+  "attr.rust.derive": "Copy, PartialEq, Eq",
+  "attr.rust.repr": "transparent",
+  "attr.rust.tuple_struct"
+) {
+  /// Invalid value. Won't show up in generated types.
+	Invalid = 0,
```
Wumpf: tabs :O. Well, we don't have formatting rules enforced on these so whatever ;-)

crates/store/re_types/definitions/rerun/blueprint/components/loop_mode.fbs
```diff
@@ -0,0 +1,25 @@
+namespace rerun.blueprint.components;
+
+// ---
+
+/// How the time should loop.
```
Wumpf: nit:
```suggestion
/// If playing, whether and how the playback time should loop.
```

crates/store/re_types/definitions/rerun/blueprint/archetypes/panel_blueprint.fbs
```diff
@@ -40,4 +40,17 @@ table TimePanelBlueprint (
 
   /// Frames per second. Only applicable for sequence timelines.
   fps: rerun.blueprint.components.Fps ("attr.rerun.component_optional", nullable, order: 2300);
+
+  /// If the time is currently paused, playing, or following.
+  ///
+  /// Defaults to playing.
```
Wumpf: it's more nuanced than that, no? depending on how data arrives we may also default to latest

crates/store/re_types/definitions/rerun/datatypes/time_range.fbs
```diff
@@ -0,0 +1,14 @@
+namespace rerun.datatypes;
+
+/// A 64-bit number describing either nanoseconds OR sequence numbers.
```
Wumpf: wrong comment

crates/store/re_types/definitions/rerun/datatypes/time_range.fbs
```diff
@@ -0,0 +1,14 @@
...
+struct AbsoluteTimeRange (
+  "attr.rust.derive": "Copy, PartialEq, Eq, PartialOrd, Ord",
+  "attr.rust.override_crate": "re_types_core"
+) {
+  /// Start of the range.
+  min: TimeInt (order: 100);
+
+  /// End of the range.
```
Wumpf:
```suggestion
  /// End of the range (inclusive).
```

examples/python/blueprint/blueprint.py
```diff
@@ -39,7 +39,14 @@ def main() -> None:
...
             rrb.SelectionPanel(state="collapsed"),
-            rrb.TimePanel(state="collapsed", timeline="custom", sequence_cursor=15),
+            rrb.TimePanel(
+                state="collapsed",
+                timeline="custom",
+                sequence_cursor=15,
+                time_selection=rrb.components.AbsoluteTimeRange(10, 25),
+                loop_mode="selection",
+                play_state="playing",
+            ),
```
Wumpf, line 49: very nice to have all these exposed!

crates/viewer/re_time_panel/src/time_selection_ui.rs
```diff
@@ -176,8 +177,8 @@ pub fn loop_selection_ui(
             // A zero-sized loop selection is confusing (and invisible), so remove it
             // (unless we are in the process of dragging right now):
             time_commands.push(TimeControlCommand::RemoveLoopSelection);
-        } else {
-            // Update it in case it was modified:
+        } else if Some(selected_range.to_int()) != time_ctrl.loop_selection().map(|s| s.to_int()) {
```
Wumpf, line 180: huh, so this spammed updates before?

crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -38,11 +38,14 @@ pub type TestHookFn = Box<dyn FnOnce(&ViewerContext<'_>)>;
 #[derive(serde::Deserialize, serde::Serialize)]
 #[serde(default)]
 pub struct AppState {
+    // TODO(isse): Just serialize app options since we skip everything else.
```
Wumpf: cc: @abey79 since we talked recently about how the serialization thing on AppState is becoming a bit laughable. Looks like this further pushed us in that direction! (for the better nonetheless).
The conclusion implied by the todo note is correct, only that there's probably some nuance on some of the other fields probably still having some serialized things (although lots of them skip also almost everything)

crates/viewer/re_viewer_context/src/time_control.rs
```diff
@@ -42,6 +45,16 @@ trait TimeBlueprintExt {
...
+
+    fn set_play_state(&self, play_state: PlayState);
+    fn get_play_state(&self) -> Option<PlayState>;
+
+    fn set_loop_mode(&self, loop_mode: LoopMode);
+    fn get_loop_mode(&self) -> Option<LoopMode>;
+
+    fn set_time_selection(&self, time_range: AbsoluteTimeRange);
+    fn get_time_selection(&self) -> Option<AbsoluteTimeRange>;
+    fn clear_time_selection(&self);
```
Wumpf, line 57: you didn't start the fire here, but still would be nice to be idiomatic:

```suggestion
    fn set_play_state(&self, play_state: PlayState);
    fn play_state(&self) -> Option<PlayState>;

    fn set_loop_mode(&self, loop_mode: LoopMode);
    fn loop_mode(&self) -> Option<LoopMode>;

    fn set_time_selection(&self, time_range: AbsoluteTimeRange);
    fn time_selection(&self) -> Option<AbsoluteTimeRange>;
    fn clear_time_selection(&self);
```

Wumpf (approved):
Awesome to have these exposed on the timeline blueprint. Makes it feel so much more complete.
Won't lie I found the code changes to the controls a bit scary, but both in reading and trying I didn't notice any bugs! 🤞 Test coverage in that area is so-so unfortunately, would be good to think a bit more about what we could do there.

>  Should play-state be a static component?

I think so. When trying things out I found it a bit strange.

crates/viewer/re_viewer/src/app_state.rs
IsseW: Might be better to actually create an issue with better detail for this then, and reference that there

crates/viewer/re_time_panel/src/time_selection_ui.rs
IsseW: Yeah, the blueprint undo warning caught it for me

PR #11665, opened 2025-10-27
Fixed some grammatical errors

zhwbubbles:
Made changes to the grammar, split longer paragraphs into smaller ones. (I think this doc has room for a lot more improvement though)


Wumpf:
thanks for wanting to contribute! But not gonna take this in since it's just mostly fairly opinionated style changes and I don't see a clear improvement with most of it - for instance I don't think the extra paragraphs in the rendered md file are that helpful for reading

PR #11666, opened 2025-10-27
Fix infinite recursion in `ci/crates.py`

abey79:
Make crate discovery more resistant against circular dependencies, which are apparently possible (via `dev-dependencies`).


scripts/ci/crates.py
```diff
@@ -153,36 +154,40 @@ def get_sorted_publishable_crates(ctx: Context, crates: dict[str, Crate]) -> dic
...
                 continue
-            if dependency.name in visited:
-                continue
-            helper(ctx, crates, dependency.name, output, visited)
+            helper(ctx, crates, dependency.name)
+
         # Insert only after all dependencies have been traversed
-        if name not in visited:
-            visited[name] = True
-            publish = crate.manifest["package"].get("publish")
```
abey79, line 174: that was the core issue: `visited` is updated _after_ recursion

PR #11668, opened 2025-10-27
Force `chrono`s `serde` feature

abey79:
Should fix nightly. Caused by:
- https://github.com/lancedb/lance/issues/5075

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9325,6 +9325,7 @@ version = "0.27.0-alpha.1+dev"
 dependencies = [
  "anyhow",
  "arrow",
+ "chrono",
  "datafusion",
  "futures",
  "insta",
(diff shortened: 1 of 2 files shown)


abey79:
> is it really just a test thing or should we enable it globally? I think I hit this also somewhere else where removing a seemingly unused chrono dependency broke the build :/
> 
> If it's somewhere else, maybe best to put it on the workspace settings also to lower number of variants?

I think our own effort of getting rid of chrono is clashing with our increasing reliance on lance, which is seemingly not opting into that feature despite using it.

Wumpf:
do we have a bug for this on lance?

abey79:
> do we have a bug for this on lance?

see PR desc ;)

PR #11671, opened 2025-10-27
Allow uploading non-images with `--single`

jprochazk:
This will now work:
```
pixi run upload-image --single path_to_video.mp4
```

Of course, the encoding of that video will depend on whatever recording/editing software was used. Whoever is uploading the video has to ensure that will work across browsers.

### DIFF
--- a/scripts/upload_image.py
+++ b/scripts/upload_image.py
@@ -127,10 +127,6 @@ class Uploader:
         gcs = storage.Client("rerun-open")
         self.bucket = gcs.bucket("rerun-static-img")
 
-    def _check_image_aspect_ratio(self, image: Path | Image) -> None:
-        if isinstance(image, Path):
-            image = PIL.Image.open(image)
-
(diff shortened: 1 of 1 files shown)


scripts/upload_image.py
```diff
@@ -127,7 +127,7 @@ def __init__(self) -> None:
         gcs = storage.Client("rerun-open")
         self.bucket = gcs.bucket("rerun-static-img")
 
-    def _check_aspect_ratio(self, image: Path | Image) -> None:
+    def _check_image_aspect_ratio(self, image: Path | Image) -> None:
         if isinstance(image, Path):
             image = PIL.Image.open(image)
```
abey79: wth? Looking at the history, I'm responsible for this 🤦🏻  Please remove entirely.

(back story is that we used to need to be careful about this, but no longer)

jprochazk: will do

PR #11672, opened 2025-10-27
Adds hierarchical queries to harness

### LINKED ISSUE #11628: Add hierarchical queries to Harness
Replace the current `nth` queries in `kittest_harness_ext.rs` with section-based selectors.
Add labels to panels wherever necessary.

Example: instead of 
```rs
harness.click_nth_label("boxes2d", 1);
```
one should be able to write:
```rs
harness.streams_panel().click_label("boxes2d");
```

aedm:
### Related

Closes #11628

### What

Adds section-based queries to the integration test harness.

Instead of 
```rs
harness.right_click_nth_label("boxes2d", 1);
```

...you can write this:
```rs
harness.blueprint_tree().right_click_label("boxes2d");
```

- Adds sections for blueprint tree, streams tree, selection panel
- Missing from this PR: a section for views. Egui doesn't always transfer parent-child relations to accesskit properly, I still need to find the root cause of that.

### DIFF
--- a/tests/rust/re_integration_test/src/viewer_section.rs
+++ b/tests/rust/re_integration_test/src/viewer_section.rs
@@ -2,15 +2,18 @@ use egui::PointerButton;
 use egui::accesskit::Role;
 use egui_kittest::kittest::Queryable as _;
 
-// A section of the viewer, e.g. the "Blueprint" or "Recording" panel. Every query and action in a section
-// only affects the children of the section.
+/// A section of the viewer, e.g. the "Blueprint" or "Recording" panel. Every query and action in a section
+/// only affects the children of the section.
 pub struct ViewerSection<'a, 'h> {
     pub harness: &'a mut egui_kittest::Harness<'h, re_viewer::App>,
     pub section_label: Option<&'a str>,
 }
 
 impl<'a, 'h: 'a> ViewerSection<'a, 'h> {
-    // Returns the root node of the section.
+    /// Returns the root node of the section.
+    ///
+    /// # Panics
+    /// Panics if the section label is not found.
     pub fn root<'n>(&'n self) -> egui_kittest::Node<'n>
     where
         'a: 'n,
@@ -22,7 +25,10 @@ impl<'a, 'h: 'a> ViewerSection<'a, 'h> {
             .get_by_role_and_label(Role::Pane, section_label)
     }
 
-    // Returns the only node with the given label.
+    /// Returns the only node with the given label.
+    ///
+    /// # Panics
+    /// Panics if there are zero or multiple nodes with the given label.
     pub fn get_label<'n>(&'n self, label: &'n str) -> egui_kittest::Node<'n>
     where
         'a: 'n,
@@ -30,7 +36,10 @@ impl<'a, 'h: 'a> ViewerSection<'a, 'h> {
         self.root().get_by_label(label)
     }
 
-    // Returns the nth node with the given label.
+    /// Returns the nth node with the given label.
+    ///
+    /// # Panics
+    /// Panics if there are fewer such nodes than `index`.
     pub fn get_nth_label<'n>(&'n self, label: &'n str, index: usize) -> egui_kittest::Node<'n>
     where
         'a: 'n,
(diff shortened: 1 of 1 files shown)


crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
```diff
@@ -157,6 +157,10 @@ impl BlueprintTree {
                                 root_container,
                             );
                         }
+                    })
+                    .response
+                    .widget_info(|| {
+                        WidgetInfo::labeled(WidgetType::Panel, true, "_blueprint_tree")
```
IsseW, line 163: If someone has a screen reader enabled on web, I think it will read this on focus (although would be good to double check). So maybe this should be a more human friendly label like "Blueprint Tree"? I guess you want to be sure that it's unique though...

Otherwise we could maybe only add these labels when the testing feature is enabled?

tests/rust/re_integration_test/src/viewer_section.rs
```diff
@@ -0,0 +1,159 @@
+use egui::PointerButton;
+use egui::accesskit::Role;
+use egui_kittest::kittest::Queryable as _;
+
+// A section of the viewer, e.g. the "Blueprint" or "Recording" panel. Every query and action in a section
```
IsseW: A lot of comments in this file can be doc comments `///`

tests/rust/re_integration_test/src/viewer_section.rs
```diff
@@ -0,0 +1,159 @@
...
+        'a: 'n,
+    {
+        let Some(section_label) = self.section_label else {
+            return self.harness.root();
+        };
+        self.harness
+            .get_by_role_and_label(Role::Pane, section_label)
+    }
+
+    // Returns the only node with the given label.
```
IsseW:
```suggestion
    /// Returns the only node with the given label.
    ///
    /// # Panics
    /// Panics if none or more than one node are found with the given label.
```

crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
aedm: To be fair, I'm not really sure how Rerun UI could be used with a screen reader in general. I don't think the app currently supports that in a practical way. Improving accessibility by adding screen reader support would be super nice but also quite challenging due to the nature of the app.

The purpose of these labels is to identify sections of the UI which are visually recognizable to the user. I labeled them like this so they are somewhat unique and internal. There might better ways to achieve this, maybe `$blueprint_tree`? What do you suggest?

I'd leave it there for release builds. It allows controlling the app by a 3rd party, which might be useful. :) But I don't have strong opinions.

IsseW: Ah you're right, I thought screen reader could be enabled from the outside in egui, but looks like it can't. 

I don't really have any opinion on `$name` vs `_name`, but should be consistent. Iirc there was one which didn't have either.

aedm: I think it definitely can be enabled outside egui, but in this case it wouldn't be much of a help to anyone trying to visualize data...

Labels starting with underscore are not visible on the screen. Every other label should be visible to the user.

IsseW: So this one is visible to users? https://github.com/rerun-io/rerun/pull/11672/files#diff-d0fe1cb35c39e2c0f03e819cd012d7b06e2cedf715f82d240c5ad6025259336eR597

tests/rust/re_integration_test/src/lib.rs
```diff
@@ -2,12 +2,15 @@
...
 mod kittest_harness_ext;
 mod test_data;
+mod viewer_section;
 
 pub use kittest_harness_ext::HarnessExt;
 use re_redap_client::{ApiError, ConnectionClient, ConnectionRegistry};
 use re_server::ServerHandle;
 use re_uri::external::url::Host;
 use std::net::TcpListener;
+// pub use viewer_section::GetSection;
```
IsseW, line 12: Dead code?

tests/rust/re_integration_test/src/kittest_harness_ext.rs
```diff
@@ -111,9 +99,72 @@ pub trait HarnessExt {
     fn set_time_panel_opened(&mut self, opened: bool) {
         self.set_panel_opened("Time panel toggle", opened);
     }
+
+    // Get a viewer section, eg. blueprint tree or selection panel.
```
IsseW, line 103: These can be doc comments as well

tests/rust/re_integration_test/src/kittest_harness_ext.rs
```diff
@@ -29,8 +29,12 @@ use re_viewer::{
 use re_viewer_context::ContainerId;
 use re_viewport_blueprint::ViewportBlueprint;
 
+// use crate::GetSection;
```
IsseW, line 32: Also dead code?

IsseW (commented):
Noticed some more stuff 😅

PR #11674, opened 2025-10-27
Introduce experimental `CoordinateFrame` archetype

Wumpf:
### Related

* Fixes RR-2698
    * probably? That one is about settling on a naming scheme for implicit transform frames
* Fixes RR-2510
* Fixes RR-2487
   * it's the last piece of "make existing transform logic work in terms of frames"
* Fixes RR-2700
* Part of RR-2486 

### What

Another important step towards fully supporting ROS tf style transforms: The `CoordinateFrame` archetype allows to set which coordinate frame an entity belongs to.
By default, every entity belongs to it's built-in/implicit coordinate frame. By logging or blueprint-overriding `CoordinateFrame` you can change what frame an entity maps to.
This internal (until now undocumented) automatic transform frame id got renamed from `rerun_tf#/some/entity/path` to `tf#/some/entity/path`.

To see what transform frame an entity uses, there's an experimiental marked UI flag under the settings menu which shows the frame
image
We're not quite sure yet where to put it. I experimented with putting it under `Visualizers` since it affects visualization and has override options (which are just the same as with regular visualizers!) but it didn't feel quite right either.

Comes with a simple example that shows the effect of switching transform frames:

[Try it live](https://rerun.io/viewer/pr/11674?url=https%3A%2F%2Fstatic.rerun.io%2Frrd%2F0.26.0%2Fcoordinate_frame_builtin_frames_869ed8c07b1e6686cdb59490fd967f71f17013a6.rrd)

demo

* [x] pass full-check

### DIFF
--- a/crates/viewer/re_viewer/src/ui/settings_screen.rs
+++ b/crates/viewer/re_viewer/src/ui/settings_screen.rs
@@ -181,9 +181,7 @@ fn settings_screen_ui_impl(ui: &mut egui::Ui, app_options: &mut AppOptions, keep
     // Experimental features
     //
 
-    // Currently, the wasm target does not have any experimental features. If/when that changes,
-    // move the conditional compilation flag to the respective checkbox code.
-    #[cfg(not(target_arch = "wasm32"))]
+    //#[cfg(not(target_arch = "wasm32"))]
     if true {
         separator_with_some_space(ui);
         ui.strong("Experimental features");


Wumpf:
@rerun-bot full-check

crates/viewer/re_viewer/src/ui/settings_screen.rs
```diff
@@ -181,13 +181,20 @@ fn settings_screen_ui_impl(ui: &mut egui::Ui, app_options: &mut AppOptions, keep
...
-    // move the conditional compilation flag to the respective checkbox code.
-    #[cfg(not(target_arch = "wasm32"))]
-    // Currently there are no experimental features
-    if false {
+    //#[cfg(not(target_arch = "wasm32"))]
+    if true {
         separator_with_some_space(ui);
         ui.strong("Experimental features");
+
+        ui.re_checkbox(
```
oxkitsune, line 189: Now that we have this `CoordinateFrame` archetype, I would expect this experimental UI to always show up in the selection panel if I log a `CoordinateFrame`, probably under the archetype itself though.

oxkitsune (approved):

crates/viewer/re_viewer/src/ui/settings_screen.rs
Wumpf: but `CoordinateFrame` itself is still experimental though 🤷

PR #11680, opened 2025-10-28
Add timestamps to printed output in `crates.py`

abey79:
Print timestamp so it's easier to understand what's going on when things are slow

### DIFF
--- a/scripts/ci/crates.py
+++ b/scripts/ci/crates.py
@@ -448,7 +448,7 @@ def publish_crate(crate: Crate, token: str, version: str, env: dict[str, Any]) -
         # TODO(#11199): remove this hack
         publish_cmd += " --allow-dirty"
 
-    print(f"{G}Publishing{X} {B}{name}{X}…")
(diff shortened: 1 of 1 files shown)


scripts/ci/crates.py
```diff
@@ -448,7 +448,7 @@ def publish_crate(crate: Crate, token: str, version: str, env: dict[str, Any]) -
         # TODO(#11199): remove this hack
         publish_cmd += " --allow-dirty"
 
-    print(f"{G}Publishing{X} {B}{name}{X}…")
+    print(f"{datetime.now()} {G}Publishing{X} {B}{name}{X}…")
```
emilk, line 451: What format is this? I'm thinking `HH:MM:SS` should suffice (or even "seconds since start"). I don't want _too_ verbose output.

abey79: It's a good old `YYYY-MM-DD HH:mm:ss` snapshot that is very nicely aligned (so easy to ignore) and is clean and non-verbose in the code.

emilk:
I just discovered this in the runner output 🤯 

image

PR #11681, opened 2025-10-28
Fix failing Pip index job

abey79:
This is caused by:
- https://github.com/googleapis/python-api-core/issues/848


Wumpf (changes requested):
can we also pin the version for google cloud? cleary not doing that in the first place is why we get this now

abey79:
> can we also pin the version for google cloud? cleary not doing that in the first place is why we get this now

makes sense, done

PR #11682, opened 2025-10-28
Don't use stored credentials in tests/on localhost

jprochazk:
* Closes RR-2793

NOTE: This is a temporary solution until we have proper auth flows which differentiate between using stored credentials and an explicit token.

We no longer use `CliCredentialsProvider` by default. It now must be explicitly enabled. It is currently enabled in all scenarios except tests.

We also check if the origin is `localhost` before actually using stored credentials, because it's rare that you'd want authentication on `localhost`. If that is necessary, it's possible to set an env var:

```
USE_STORED_CREDENTIALS=1 pixi run rerun
```

Since it's only really useful for testing authentication against a localhost redap, we're not going to publicly advertise that anywhere.

Also slightly improves error message when session expired, by not duplicating it:

Before:

```
$ rerun auth token
Error: Your session ended due to inactivity, run `rerun auth login` first -> Your session ended due to inactivity, run `rerun auth login` first
```

After:

```
$ rerun auth token
Error: Your session ended due to inactivity, run `rerun auth login` first
```

### DIFF
--- a/crates/store/re_redap_client/src/connection_registry.rs
+++ b/crates/store/re_redap_client/src/connection_registry.rs
@@ -54,7 +54,6 @@ impl ConnectionRegistry {
     ///
     /// Using stored credentials by default should be preferred in all other contexts,
     /// see [`Self::new_with_stored_credentials`].
-    #[expect(clippy::new_ret_no_self)] // intentional, to reflect the fact that this is a handle
     pub fn new_without_stored_credentials() -> ConnectionRegistryHandle {
         ConnectionRegistryHandle {
             inner: Arc::new(RwLock::new(Self {
@@ -69,7 +68,6 @@ impl ConnectionRegistry {
     /// Create a new connection registry and return a handle to it.
     ///
     /// This version explicitly enables using credentials stored on the local machine.
-    #[expect(clippy::new_ret_no_self)] // intentional, to reflect the fact that this is a handle
     pub fn new_with_stored_credentials() -> ConnectionRegistryHandle {
         ConnectionRegistryHandle {
             inner: Arc::new(RwLock::new(Self {


crates/viewer/re_viewer/src/web.rs
```diff
@@ -59,7 +59,7 @@ impl WebHandle {
 
         let app_options: Option<AppOptions> = serde_wasm_bindgen::from_value(app_options)?;
 
-        let connection_registry = re_redap_client::ConnectionRegistry::new();
+        let connection_registry = re_redap_client::ConnectionRegistry::new().with_stored_credentials();
```
abey79: I'm not fan of the builder api in this particular case. Feels too easy to forget about this.

My suggestion:
```
ConnectionRegistry::new_with_stored_credential()
ConnectionRegistry::new_without_stored_credential()
```
(with explanations on where to use what in docstrings)

Rationale: we use this call rarely, so making it verbose and ugly is worth so it's explicit and without footgun.

crates/utils/re_uri/src/origin.rs
```diff
@@ -145,11 +149,15 @@ fn is_host_unspecified(host: &url::Host) -> bool {
...
-        },
+        url::Origin::Tuple(_, host, _) => is_host_localhost(host),
+    }
+}
+
+fn is_host_localhost(host: &url::Host) -> bool {
+    match host {
+        url::Host::Domain(domain) => domain == "localhost",
+        url::Host::Ipv4(ip) => ip.is_loopback() || ip.is_unspecified(),
+        url::Host::Ipv6(ip) => ip.is_loopback() || ip.is_unspecified(),
```
abey79, line 160: (surprising `url::Host` doesn't have a such an api)

abey79 (commented):
lgtm, thanks for taking care of that. But please do consider the suggested API change.

PR #11684, opened 2025-10-28
make `TransformResolutionCache` test a bit more compact

Wumpf:
will add more tests soon, so I split this small housekeeping out


oxkitsune (approved):
neat

PR #11685, opened 2025-10-28
Test: deselect on escape

aedm:
### Related

Part of #8948 

### What

- Adds automated tests for deselecting items
- Removes `check_deselect_on_escape.py` from release checklist

### DIFF
--- a/tests/rust/re_integration_test/tests/deselect_on_escape.rs
+++ b/tests/rust/re_integration_test/tests/deselect_on_escape.rs
@@ -71,4 +71,8 @@ pub async fn test_deselect_on_escape() {
     harness.snapshot_app("deselect_on_escape_6");
     harness.key_press(egui::Key::Escape);
     harness.snapshot_app("deselect_on_escape_7");
+    // Whatever was selected should become deselected after pressing escape again
+    harness.key_press(egui::Key::Escape);
+    harness.set_selection_panel_opened(false);
+    harness.snapshot_app("deselect_on_escape_8");
 }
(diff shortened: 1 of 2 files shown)


tests/rust/re_integration_test/tests/snapshots/xtemp.png
emilk, line 1: "xtemp"?

tests/rust/re_integration_test/tests/deselect_on_escape.rs
```diff
@@ -0,0 +1,71 @@
...
+    harness.snapshot_app("deselect_on_escape_4");
+    harness.key_press(egui::Key::Escape);
+    harness.snapshot_app("deselect_on_escape_5");
+
+    // Select view, right click on blueprint tree, escape, view remains selected
+    harness.blueprint_tree().click_label("3D view");
+    harness.blueprint_tree().right_click_label("3D view");
+    harness.snapshot_app("deselect_on_escape_6");
+    harness.key_press(egui::Key::Escape);
+    harness.snapshot_app("deselect_on_escape_7");
```
emilk, line 73: Suggestion" add another snapshot after pressing escape yet again. Whatever is currently selected should become unselected, and the recording itself should become selected (which will be visible in the selection panel).

tests/rust/re_integration_test/tests/deselect_on_escape.rs
```diff
@@ -0,0 +1,71 @@
...
+}
```
emilk, line 78: An alternative approach here would be to replace the screenshot tests with actually checking the current selection `Item`

tests/rust/re_integration_test/tests/deselect_on_escape.rs
```diff
@@ -0,0 +1,71 @@
+use re_integration_test::HarnessExt as _;
```
emilk, line 3:
```suggestion
//! Test that pressing the Escape key will back out of dialogs etc.

use re_integration_test::HarnessExt as _;
```

PR #11686, opened 2025-10-28
Add a new `target_partition_num_rows` API parameter to the CreateIndex API 

jleibs:
### What

The old way of specifying `num_partitions` is now deprecated.
See: https://docs.rs/lance-index/0.38.2/lance_index/vector/ivf/builder/struct.IvfBuildParams.html#structfield.num_partitions

Lance exposes a new `target_partition_size` parameter which gives me predictable performance instead of users manually needing to increase the number of partitions as the size of the dataset increases. This works better for things like incremental indexing.

Resolves: RR-2772


crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
```diff
@@ -371,9 +371,16 @@ message InvertedIndex {
 }
 
 message VectorIvfPqIndex {
+  // num_partitions is deprecated. Remove in 0.6
+  // Use target_partition_size instead
   optional uint32 num_partitions = 1;
```
abey79: My understanding is that we don't give the `reserved` treatment to this now because we want to first deprecated it at the API level. Is that correct?

If so, please open an issue about the upcoming cleanup (what needs to be done and when, e.g. after 0.27 is released), and the associated TODOs.

rerun_py/src/catalog/dataset_entry.rs
abey79, line 631: Let's add a docstring with the parameters here, including a note on the deprecation

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1529,11 +1529,41 @@ class DatasetEntry(Entry):
...
+        Create a vector index on the given column.
+
+        This will enable indexing and build the vector index over all existing values
+        in the specified component column.
+
+        Results can be retrieved using the `search_vector` API, which will include
+        the time-point on the indexed timeline.
+
+        Only one index can be created per component column -- executing this a second
+        time for the same component column will replace the existing index.
```
teh-cmc, line 1547: Note to self that I will need to update that.

crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
```diff
@@ -383,9 +383,16 @@ message InvertedIndex {
...
+  // num_partitions is deprecated. Use target_partition_size instead
+  // TODO(RR-2798): Remove in 0.6
   optional uint32 num_partitions = 1;
+
   optional uint32 num_sub_vectors = 2;
+
   VectorDistanceMetric distance_metrics = 3;
+
+  // Target size of the IVF partition in rows
+  optional uint32 target_partition_size = 4;
```
teh-cmc: nit: I would have never expected this to be a number of rows, `size` immediately screams `bytes` to me.
```suggestion
  optional uint32 target_partition_num_rows = 4;
```

jleibs: Fair comment. The upside of `target_partition_size` is it matches the lance docs... :disappointed: 

https://docs.rs/lance-index/0.38.2/lance_index/vector/ivf/builder/struct.IvfBuildParams.html#structfield.target_partition_size

jleibs: Decided to go for clarity over consistency with lance.

PR #11688, opened 2025-10-28
Test: drag and drop streams to view

aedm:
### Related

Part of #8948 

### What

- Adds automated tests for dragging from stream tree to views
- Removes `check_entity_drag_and_drop.py` from release checklist

### DIFF
--- a/.github/workflows/reusable_checks.yml
+++ b/.github/workflows/reusable_checks.yml
@@ -263,19 +263,21 @@ jobs:
           key: cache-lychee-${{ github.sha }}
           restore-keys: cache-lychee-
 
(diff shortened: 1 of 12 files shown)


tests/rust/re_integration_test/tests/drop_stream_to_view.rs
```diff
@@ -0,0 +1,151 @@
+use std::f64::consts::TAU;
```
emilk, line 5: It would be nice if these tests starts with a `//!` comment explaining in English roughly what they are about. If this tests suddenly fails it will be very useful to be able to quickly assess what it is all about, without having to read 150 lines of Rust.

PR #11689, opened 2025-10-28
Add Ability to Launch OSS Server From Python

oxkitsune:
### Related

* Part of RR-2532
* Closes RR-2477

### What

This adds Python API for the OSS server, meaning you can now use it directly from python:

```py
import rerun as rr

server = rr.server.Server(port=9876, datasets={"my_data": "path/to/data.rrd"}) 
client = server.client()

# Use the client to interact with the catalog
datasets = client.datasets()
```

### DIFF
--- a/rerun_py/rerun_bindings/rerun_bindings.pyi
+++ b/rerun_py/rerun_bindings/rerun_bindings.pyi
@@ -1935,6 +1935,7 @@ class ServerInternal:
     """
 
     def __init__(
+        self,
         *,
         address: str = "0.0.0.0",
         port: int = 51234,
--- a/rerun_py/src/server.rs
+++ b/rerun_py/src/server.rs
@@ -21,6 +21,7 @@ pub struct PyServerInternal {
 impl PyServerInternal {
     #[new]
     #[pyo3(signature = (*, address="0.0.0.0", port=51234, datasets=None, tables=None))]
+    #[pyo3(text_signature = "(self, *, address='0.0.0.0', port=51234, datasets=None, tables=None)")]
     pub fn new(
         py: Python<'_>,
         address: &str,


rerun_py/Cargo.toml
```diff
@@ -15,14 +15,14 @@ name = "rerun_bindings" # name of the .so library that the Python module will im
...
-default = ["extension-module"]
+default = ["extension-module", "oss_server"]
 
 ## Extra features that aren't ready to be included in release builds yet.
 extra = []
 
 ## The features we turn on when building the `rerun-sdk` PyPi package
 ## for <https://pypi.org/project/rerun-sdk/>.
-pypi = ["extension-module", "nasm", "web_viewer", "server"]
+pypi = ["extension-module", "nasm", "web_viewer", "server", "oss_server"]
```
oxkitsune: Not sure if this is the right time already.

This makes our pypi packages depend on lance again.

rerun_py/src/server.rs
```diff
@@ -0,0 +1,108 @@
...
+use pyo3::{
+    Bound, PyResult, Python,
+    exceptions::PyValueError,
+    pyclass, pymethods,
+    types::{
+        PyAnyMethods as _, PyDict, PyDictMethods as _, PyModule, PyModuleMethods as _, PyString,
+        PyStringMethods as _,
+    },
+};
+use re_sdk::external::re_server::{self, Args as ServerArgs};
```
abey79: Any reason you included `re_server` in `re_sdk/Cargo.toml` and not in `rerun_py/Cargo.toml`?

oxkitsune: As discussed offline, no specific reason, i got misled by the `server` feature initially and followed that all the way through to `re_sdk`.

I think it makes sense to make `re_server` a direct dependency of `rerun_py`, and went ahead and implemented that.

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1907,3 +1907,67 @@ class NotFoundError(Exception):
...
     """Raised when trying to create a resource that already exists."""
+
+class ServerInternal:
+    """
+    Internal Rerun server instance.
+
+    This is the low-level binding to the Rust server implementation.
+    Users should typically use `rerun.server.Server` instead.
+    """
+
```
abey79, line 1936: FYI, you don't need to provide docstrings for `*Internal` objects here. They aren't checked like the others. I like this particular class level docstring, but I think i'd remove the method ones, because they will rot (since they are not checked).

rerun_py/rerun_sdk/rerun/server.py
```diff
@@ -0,0 +1,130 @@
...
+class Server:
+    """
+    A Rerun server instance.
+
+    This class allows you to start and manage a Rerun server programmatically.
+    The server hosts recordings and serves them via HTTP, and provides access to
+    the catalog through a client connection.
+
+    The server can be used as a context manager, which will automatically shut down
+    the server when exiting the context.
```
abey79, line 25: This is nice for tests and script, but is not going to work for notebook. So I think we should have a `__del__` that calls `shutdown()`. Also, I'm not 100% sure about exposing `start()` and `stop()`. What's the purpose of exposing this, vs. just tying the server life time with the object's? 

in any case, would love to have @ntjohnson1 take here

rerun_py/src/server.rs
```diff
@@ -0,0 +1,108 @@
...
+                        })
+                    })
+                    .collect::<Vec<_>>()
+            })
+            .unwrap_or_default();
+
+        // we can re-use the CLI argument to construct the server
+        let args = ServerArgs {
+            addr: address.unwrap_or("0.0.0.0".to_owned()),
+            port: port.unwrap_or(51234),
```
abey79: Let's use a random free port here, to avoid port clash issue. If only for DX (we often have a local server running on that exact port).

rerun_py/src/server.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Returns the address that clients can use to connect to this server,
+    /// formatted as a `rerun+http://` URL.
+    pub fn address(&self) -> String {
+        format!("rerun+http://{}", self.address)
+    }
+
+    /// Shutdown the server, blocking until it has fully stopped.
+    ///
+    /// If the server is not running, raises a `ValueError`.
+    pub fn shutdown(&mut self, py: Python<'_>) -> PyResult<()> {
```
abey79, line 66: again, we should probably call that when the object is dropped

rerun_py/Cargo.toml
abey79: i'm in favour of this

rerun_py/Cargo.toml
```diff
@@ -42,6 +42,9 @@ perf_telemetry = ["dep:re_perf_telemetry", "re_redap_client/perf_telemetry"]
 
 server = ["re_sdk/server", "dep:re_grpc_server"]
 
+## Support for running an open-source Rerun server instance.
+oss_server = ["dep:re_server"]
```
abey79: you need to activate the lance feature of the server as well

abey79:
Also, I think you should be able to fix the `e2e_redap_tests` features to use this now. Would be great way to add test coverage.

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1907,3 +1907,67 @@ class NotFoundError(Exception):
 
 class AlreadyExistsError(Exception):
     """Raised when trying to create a resource that already exists."""
+
+class ServerInternal:
```
ntjohnson1, line 1929: NIT: We don't currently have the convention but prefixing with `_` is more pythonic than calling it internal.

rerun_py/rerun_sdk/rerun/server.py
```diff
@@ -0,0 +1,130 @@
...
+        datasets = client.datasets()
+    ```
+
+    """
+
+    def __init__(
+        self,
+        address: str | None = None,
+        port: int | None = None,
+        datasets: dict[str, PathLike[str]] | None = None,
```
ntjohnson1, line 47: Doesn't the cli also take a table argument?

rerun_py/rerun_sdk/rerun/server.py
```diff
@@ -0,0 +1,130 @@
...
+        Create a new Rerun server instance and start it.
+
+        The server will host recordings and serve them via HTTP. If datasets are provided,
+        they will be loaded and made available when the server starts.
+
+        Parameters
+        ----------
+        address:
+            The address to bind the server to.
+            Defaults to `"0.0.0.0"`.
```
ntjohnson1: Opinion on just setting the defaults in the argument? On one hand it means we are managing in both rust and python. On the other it is self documenting where here if we change the rust default the comment rots.

rerun_py/rerun_sdk/rerun/server.py
```diff
@@ -0,0 +1,130 @@
...
+
+        Returns
+        -------
+        bool
+            `True` if the server is running, `False` otherwise.
+
+        """
+        return self._internal.is_running()
+
+    def stop(self) -> None:
```
ntjohnson1: Is there a reason we don't just call this shutdown? And if we are exposing this why are we calling the functionality directly in exit?

rerun_py/rerun_sdk/rerun/server.py
ntjohnson1: Why won't this work for a notebook?

IIUC we don't need to call shutdown on `__del__` since on delete of this wrapper it should clean up the internal rust object which should shutdown the server.

I think having `shutdown` exposed is nice in case someone wants to terminate early. Similar to the context manager.

```python
with Server() as srv:
   client = serv.client()
   # do some stuff
# jump back to not needing to hold a port
```

rerun_py/src/server.rs
ntjohnson1: Agree random free port is much nicer but if that delays the PR I'd kind of rather land and add that as an extension in follow up.

oxkitsune: Should be fairly straight forward, can have a go at it

rerun_py/rerun_sdk/rerun/server.py
oxkitsune: Exactly, we don't need `__del__`, however since we have the python wrapper, I don't think it can hurt

rerun_py/rerun_sdk/rerun/server.py
oxkitsune: It does, will add those as well.

rerun_py/rerun_sdk/rerun/server.py
abey79: > Why won't this work for a notebook?
The context manager specifically, I meant.

As for `__del__`, yeah, if the internal object cleans up properly it should be ok.

rerun_py/src/server.rs
abey79: I mean, unless dropping the ServerHandle does the right thing? Haven't checked.

rerun_py/rerun_bindings/rerun_bindings.pyi
abey79: I agree we should have `_`, but i'd keep the `Internal` suffix :)

rerun_py/rerun_sdk/rerun/server.py
ntjohnson1: > if the internal object cleans up properly it should be ok.

If the rust object deletion doesn't automatically shut down the server that probably means there are a bunch of extra bugs. I'm not opposed to do a belt and suspenders to try and explicitly call shut down at the end anyway.

> The context manager specifically, I meant.

I'm still not clear why the context manager wouldn't work in the notebook. Can you provide an example/elaborate?

rerun_py/rerun_bindings/rerun_bindings.pyi
ntjohnson1: No objection

rerun_py/rerun_sdk/rerun/server.py
abey79: > If the rust object deletion doesn't automatically shut down the server that probably means there are a bunch of extra bugs. I'm not opposed to do a belt and suspenders to try and explicitly call shut down at the end anyway.

Yeah, I'm probably being overly defensive with this. I agree it _should_ work.

> > The context manager specifically, I meant.
> 
> I'm still not clear why the context manager wouldn't work in the notebook. Can you provide an example/elaborate?

Well, it would work in a single cell. But you typically want to start the server in the first cell and use it throughout your notebook. Or am I missing something?

oxkitsune: Yea, but then you wouldn't use the context manager right? You can still construct the server manually.

abey79: Yes. My point is, the existence of a context manager signals that starting/shutdown might need to be manually managed (like eg file handle, which needs explicit closing if you don't use the CM). Since this is in fact not the case, we should be explicit.

ntjohnson1: I disagree and think the context manager is nice if you only need to server to do some file extraction then everything else is managed separately. However, if you feel strongly I'd rather drop it, land the PR and circle back to it later.

> eg file handle

Technically you don't have to close file handles since when it goes out of scope it closes so its similarly just a convenience for scoping access to things
```python
f = open(...)
del f # or just wait for the program to close
```

```python
with open(..) as f:
   pass
``

abey79: Oh, to clarify, I didn't mean we should drop the CM. I realise I have been utterly confusing in my communication about that, so I'm going start over.

When I see something like `shutdown()` or a CM in an API, _especially_ for something that looks like it involves another process (which it doesn't, but smells like it does*), my immediate reaction is "If I'm not using the CM, should I manually manage shutdown or anything???". And since we want to use this in a notebook environment, we cannot just say "always use the CM and you'll be fine".

So, from there, I wanted to (1) make sure that we indeed cleanup properly on `del` (settled), and (2) we address this in the docstring. Could be as simple as stating that "the server is automatically shutdown when the `Server` instance is garbage collected" (or sth to that effect). I'm otherwise very much in favour of keeping the CM.

ntjohnson1:
I added the do-not-merge label because this will push us over our pypi limit so until we get that resolved this is unfortunately paused.

abey79:
> I added the do-not-merge label because this will push us over our pypi limit so until we get that resolved this is unfortunately paused.

Removed, because our cap was just bumped to 200MB 🎉

PR #11691, opened 2025-10-28
Return optional DebugInfo with `memory_used` as part of CreateIndexResponse

jleibs:
### Related
- RR-2797

### What

This plumbs through an optional DebugInfo from the server. We expose this as part of a new IndexingResult returned from `create_vector_index`.  `debug_info()` itself is exposed to python simply as a dictionary so we can more easily add more data here without worrying about API changes.

### DIFF
--- a/rerun_py/src/catalog/dataset_entry.rs
+++ b/rerun_py/src/catalog/dataset_entry.rs
@@ -875,7 +875,7 @@ fn py_object_to_i64(py: Python<'_>, obj: &Bound<'_, PyAny>) -> PyResult<i64> {
     converted.extract::<i64>()
 }
 
-/// A dataset entry in the catalog.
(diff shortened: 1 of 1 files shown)


crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
```diff
@@ -332,6 +332,9 @@ message CreateIndexRequest {
 message CreateIndexResponse {
   reserved 1;
   reserved "data";
+
+  // Optional debug information about the index-creation task
+  optional DebugInfo debug_info = 2;
```
teh-cmc: message types are always optional
```suggestion
  DebugInfo debug_info = 2;
```

teh-cmc (approved):

PR #11692, opened 2025-10-28
Improve our Python e2e test suite and document our test infrastructure

abey79:
### What

This PR introduce an E2E test for `DatasetEntry.partition_ids`, which was broken until recently. It does that _the proper way_, that is:

- Refactor the e2e python test suite such that we can split it in multiple files
- Introduce snapshot testing to python using  [syrupy](https://github.com/syrupy-project/syrupy)
- Introduce `TESTING.md` to provide a birds-eye view about all our testing bits and pieces

### DIFF
--- a/rerun_py/tests/e2e_redap_tests/test_partition_id.py
+++ b/rerun_py/tests/e2e_redap_tests/test_partition_id.py
@@ -1,9 +1,14 @@
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 from .conftest import DATASET_FILEPATH, ServerInstance
 
+if TYPE_CHECKING:
+    from syrupy import SnapshotAssertion
 
-def test_partition_ids(server_instance: ServerInstance, snapshot) -> None:
+
+def test_partition_ids(server_instance: ServerInstance, snapshot: SnapshotAssertion) -> None:
     """Test that we can successfully collect information about partitions."""
     client = server_instance.client
 


rerun_py/tests/e2e_redap_tests/conftest.py
```diff
@@ -0,0 +1,179 @@
...
+DATASET_NAME = "dataset"
+
+DATASET_FILEPATH = pathlib.Path(__file__).parent.parent.parent.parent / "tests" / "assets" / "rrd" / "dataset"
+TABLE_FILEPATH = (
+    pathlib.Path(__file__).parent.parent.parent.parent / "tests" / "assets" / "table" / "lance" / "simple_datatypes"
+)
+
+
+@pytest.fixture(scope="module")
+def table_filepath() -> Generator[pathlib.Path, None, None]:
```
timsaucer, line 34: One thing I've found annoying in other repos is where it's not obvious to me where the fixtures are coming from. Can we label this file `fixtures.py` ? or maybe `conftest.py` has meaning to you that I don't have. I could also see that going down the route of blowing up if we add a lot more fixtures. It's not a strong opinion.

rerun_py/tests/e2e_redap_tests/test_partition_id.py
```diff
@@ -0,0 +1,23 @@
...
+
+from typing import TYPE_CHECKING
+
+from .conftest import DATASET_FILEPATH, ServerInstance
+
+if TYPE_CHECKING:
+    from syrupy import SnapshotAssertion
+
+
+def test_partition_ids(server_instance: ServerInstance, snapshot: SnapshotAssertion) -> None:
```
timsaucer, line 11: For my understanding, is there some kind of auto-fixture for the snapshot based on matching filename? Can I have multiple snapshots in a single file? I will take a look at the documentation for syrupy.

rerun_py/tests/e2e_redap_tests/conftest.py
abey79: `conftest.py` has a meaning to pytest: https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files

It automatically find them there to do its dependency injection magic. 

I'll add a comment in there to clarify, though.

rerun_py/tests/e2e_redap_tests/test_partition_id.py
abey79: The _name_ of the argument in a test function informs pytest of the dependency injection it's supposed to do.

It goes more or less like this:
- oh, this test wants a `snapshot` fixture
- bummer, I don't have any built-in fixture going by that name
- oh but I did load some plug-ins, better check 'em
- bingo! that sweet sweet syrupy plug in has a `snapshot` fixture, let's inject that

That's why pytest present itself primarily as a dependency injection framework.

Note that type hints have no play at all in all this. I've added them just to appease `mypy`. I would go as far as pretending that it is unidiomatic to annotate test function. Curious about @ntjohnson1's view on that, btw.

PR #11697, opened 2025-10-29
Update 'project' to 'workspace' in pixi.toml

swallez:
Pixi renamed `project` to `workspace` in [version 0.45](https://pixi.sh/v0.58.0/CHANGELOG/#0450-2025-04-07) and this causes a warning every time `pixi` is run.

We have `requires-pixi = ">=0.55.0"` so no one should be using a pre-0.45 version anyway.

Also updates the pixi version in `scripts/check_env.py` -- but is that check really needed given that `pixi.toml` also has it?

### DIFF
--- a/scripts/check_env.py
+++ b/scripts/check_env.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 import subprocess
 import sys
 
-PIXI_VERSION = "0.41.4"
+PIXI_VERSION = "0.55.0"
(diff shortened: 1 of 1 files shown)


grtlr (approved):
Thank you!

pixi.toml
grtlr, line 12: nit: Yeah I think we can just delete this line.
```suggestion
```

PR #11698, opened 2025-10-29
Fix crate publishing script

abey79:
This PR fixes the crate publishing script in the following ways:

- Ignore dev-dependencies (as per the cargo book)
- Sanitise the dependency graph (completeness, no cycles)
- Fix the token management in the dag
- Improve the dry run mode (which I heavily used to debug this)
- Add some printed output

### DIFF
--- a/scripts/ci/dag.py
+++ b/scripts/ci/dag.py
@@ -42,7 +42,7 @@ class RateLimiter:
 _T = TypeVar("_T", bound=Hashable)
 
 
-def _sanitize_dependency_graph(dependency_graph: dict[_T, list[_T]]):
+def _sanitize_dependency_graph(dependency_graph: dict[_T, list[_T]]) -> None:
     """
     Sanitize the dependency graph.
 
@@ -63,7 +63,7 @@ def _sanitize_dependency_graph(dependency_graph: dict[_T, list[_T]]):
     rec_stack = set()
     path = []
 
-    def find_cycle(node):
+    def find_cycle(node: _T) -> bool:
         visited.add(node)
         rec_stack.add(node)
         path.append(node)


scripts/ci/crates.py
```diff
@@ -508,6 +536,15 @@ def publish_unpublished_crates_in_parallel(all_crates: dict[str, Crate], version
     for name, crate in unpublished_crates.items():
         dependencies = []
         for dependency in crate_deps(crate.manifest):
+            # IMPORTANT: we must ignore dev-dependencies.
+            #
+            # As per the Cargo book, `dev-dependencies` are ignored by crates.io (contrary to `build-dependencies`).
+            # https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html
+            #
+            # Also, Cargo allows circular dependencies via `dev-dependencies`, but our publishing logic requires a DAG.
+            if dependency.kind == DependencyKind.DEV:
```
jprochazk, line 534: And I guess circular dependencies in normal `dependencies` isn't going to be a problem here, because that will fail at `cargo build`.

scripts/ci/dag.py
```diff
@@ -38,10 +38,60 @@ def get(self) -> bool:
         else:
             return False
 
+    def put(self) -> None:
+        """Return a token to the bucket when a task completes."""
```
jprochazk: Can you describe your reasoning behind this? This change seems wrong to me.

The rate limiter mimics the algorithm used on `crates.io`. If you complete a publish, you can't just publish another one right away unconditionally.

scripts/ci/crates.py
```diff
@@ -456,16 +478,20 @@ def publish_crate(crate: Crate, token: str, version: str, env: dict[str, Any]) -
                 publish_cmd,
                 cwd=crate.path,
                 env=env,
-                dry_run=False,
+                dry_run=dry_run,
                 capture=True,
             )
 
-            if not is_already_published(version, crate):
+            if dry_run:
```
jprochazk: IIUC, this is causes the following `is_already_published` check to always return `True` when `dry_run=True`. I don't understand why a global set is involved at all. What am I missing?

The point behind the dry run is to show what _would be_ published, which means knowing exactly what is already published. Doing those checks is harmless. Why are you disabling them here? These crates won't be published again, anyway

jprochazk (changes requested):

scripts/ci/crates.py
abey79: That. And also I now check we don't have a circular dependencies. So we'll know if that happens.

scripts/ci/crates.py
abey79: Well, in a dry run, things don't actually get published. So the crate check after `cargo publish` would always fail. Also, the initial check on for all crates is super long, which makes it super annoying when debugging the logic. So I'm basically mocking the entire logic of using `crates.io` to hold the "publishedness" of crates using this global variable.

jprochazk: I'm still missing what the point of the global is. To me, it looks like this is equivalent to:

```
if dry_run:
	break
```

You're adding the crate to that global set, and then calling `is_already_published` which will immediately early-exit because the crate was _just_ added to the set.

scripts/ci/dag.py
abey79: Yeah, I agree. I shouldn't have let claude do that. Still, at least in dry run mode, I still deadlock without it, so there is a logical bug somewhere 🤔

jprochazk: Is it a deadlock, or is it just waiting for the bucket to get more tokens?

jprochazk: You can test that by making the rate limit some absurdly high number in `dry_run` mode.

scripts/ci/crates.py
abey79: We call `is_already_published` elsewhere that also needs to know about fake-published crates.

scripts/ci/dag.py
abey79: Ah ofc, you're right. Yeah it works if I set it to a high number.

scripts/ci/crates.py
jprochazk: It's only called here, and also in `publish_unpublished_crates_in_parallel`, when it collects unpublished crates. At that point, the global set will be empty, so it will treat everything as unpublished - this is exactly what you don't want. The dry run should show which crates will be published, and the unpublished ones won't be, so they shouldn't be shown. But with this approach everything will be shown. There are no other calls that I can find

jprochazk: My suggestion is:
- Revert the global published set
- Exit early in `publish_crates` in `dry_run` mode:

```
cargo(
	publish_cmd
    # ...
)

if not dry_run and not is_already_published(version, crate):
	# ...

print("Published")

break
```

abey79: Well, this makes dry-run insufficient for debugging, but I guess I'm done with that now and next time I can always comment out the call to `publish_unpublished_crates_in_parallel` 🤷🏻

PR #11700, opened 2025-10-29
Test: check focus after double clicks

aedm:
### Related

Part of #8948 

### What

- Adds automated test checking what happens on double clicks
- Removes `check_focus.py` from release checklist

### DIFF
--- a/tests/rust/re_integration_test/tests/check_focus_test.rs
+++ b/tests/rust/re_integration_test/tests/check_focus_test.rs
@@ -1,3 +1,10 @@
+//! This test logs a few boxes and performs the following focus checks:
+//!
+//! - Double-click on a box in the first view
+//!   - check ONLY the corresponding view expands and scrolls
+//!   - check the streams view expands and scrolls
+//! - Double-click on the leaf "boxes3d" entity in the streams view, check both views expand (manual scrolling might be needed).
+
 use re_integration_test::HarnessExt as _;
 use re_sdk::TimePoint;
 use re_sdk::log::RowId;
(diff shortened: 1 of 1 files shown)


tests/rust/re_integration_test/tests/check_focus_test.rs
```diff
@@ -0,0 +1,125 @@
+use re_integration_test::HarnessExt as _;
```
MichaelGrupp, line 8: I think it would be nice to keep a high-level summary of what the test does at the top of this file. Basically what was previously the `README` in the Python file that this test automates. Otherwise it's a bit hard to understand the context / goal.

Should of course kept up to date if the test changes 🤔 

```suggestion
//! This test logs a few boxes and performs the following focus checks:
//!
//! - Double-click on a box in the first view
//!   - check ONLY the corresponding view expands and scrolls
//!   - check the streams view expands and scrolls
//! - Double-click on the leaf "boxes3d" entity in the streams view, check both views expand (manual scrolling might be needed).

use re_integration_test::HarnessExt as _;
```

tests/rust/re_integration_test/tests/check_focus_test.rs
```diff
@@ -0,0 +1,125 @@
...
+    harness.snapshot_app("check_focus_3");
+
+    // One of the boxes is at the center of the view
+    let pixel_of_a_box = harness.get_panel_position("3D view 1").center();
+
+    // Hover over the box
+    harness.hover_at(pixel_of_a_box);
+
+    // Let the app render. This will run the picking logic which needs the GPU
+    // and lets the app find the hovered box.
```
MichaelGrupp, line 99: Tiny consistency nit: this comment ends with period `.`, most others here not. I personally prefer ending sentences with `.`.

MichaelGrupp: Seems to be also the most common style in the repo from what I can see.

tests/rust/re_integration_test/tests/snapshots/xtemp.png
MichaelGrupp, line 1: Looks like an accidentally committed file?

tests/rust/re_integration_test/tests/check_focus_test.rs
aedm: Updated.

MichaelGrupp (approved):

PR #11705, opened 2025-10-29
Update the release workflow such that alpha releases are tagged

abey79:
Title. Before only rc and final releases would have a tag. This was particularly annoying given we might consider some alpha to be 0.XX-pre release.


.github/workflows/release.yml
```diff
@@ -370,6 +369,28 @@ jobs:
           token: ${{ secrets.RERUN_BOT_TOKEN }}
 
       - name: Release tag
+        env:
+          GH_TOKEN: ${{ secrets.RERUN_BOT_TOKEN }}
```
abey79, line 373: is this necessary?

jprochazk: I don't remember if it's redundant with the `token` on checkout step, but it's better to leave it here.

PR #11708, opened 2025-10-29
Split cli into separate package

ntjohnson1:
Currently we are above the pypi limit and waiting on an increase. Internally we already kind of treat our cli, bindings and sdk as three different packages under the same name. Explore splitting out the cli to a standalone package.

TODO:
- [ ] See if anything explodes in CI
- [ ] Look at how painful it would be to set up auto pushing this
- [ ] Get people to decide if they like this


ntjohnson1:
Closing since our pypi limit went up

PR #11709, opened 2025-10-29
Expose all the python apis I could find for web docs

ntjohnson1:
### Related
Closes: RR-2766

### What

Bascially just check that all of our inputs and outputs have cross references. There is so much log spam with out `Like` types but at least the API docs in the browser are searchable and usable now. Reorganizing to make nicer is future work and there's a ticket to make this less painful/manual. https://linear.app/rerun/issue/RR-2765/change-python-public-api-surface-pattern-to-support-easier-docs

Note: Our docs preview doesn't render API docs so you need to use `pixi run py-docs-serve` note this works even without the rerun shared object, but if you have a stale one you may get an error so either delete it or rebuild.

### DIFF
--- a/scripts/ci/check_large_files.py
+++ b/scripts/ci/check_large_files.py
@@ -9,6 +9,7 @@ FILES_ALLOWED_TO_BE_LARGE = {
     "Cargo.lock",
     "CHANGELOG.md",
     "crates/build/re_types_builder/src/codegen/cpp/mod.rs",  # TODO(andreas): This file should really be split up.
+    "crates/build/re_types_builder/src/codegen/python/mod.rs",
     "crates/build/re_types_builder/src/reflection.rs",
     "crates/store/re_dataframe/src/query.rs",
     "crates/store/re_protos/proto/schema_snapshot.yaml",


crates/build/re_types_builder/src/codegen/python/mod.rs
```diff
@@ -1566,7 +1582,9 @@ fn quote_union_aliases_from_object<'a>(
         r#"
             if TYPE_CHECKING:
                 {name}Like = {name} | {union_fields}{aliases}
+                """A type alias for any {name}-like object."""
                 {name}ArrayLike = {name} | {union_fields} | Sequence[{name}Like]{array_aliases}
+                """A type alias for any {name}-like array object."""
```
emilk, line 1637: nit/style: some vertical spacing can go a long way for readability (especially since Rust and Python are inconsistent with wether docstrings come before or after the item they document!)

```suggestion
               {name}Like = {name} | {union_fields}{aliases}
                """A type alias for any {name}-like object."""
                
               {name}ArrayLike = {name} | {union_fields} | Sequence[{name}Like]{array_aliases}
                """A type alias for any {name}-like array object."""
```

ntjohnson1:
TLDR: Nothing simple came to mind but happy to brainstorm

> For instance: Is there a lint we can enable that requires all public types be documented?

So the codegen change I did updated adding doc strings to our type aliases. There doesn't seem to be strong opinions on whether type aliases need docstrings or not (maybe I didn't dig deep enough into ruff rules).

> Is there more we can do to automatically find holes in our docs?

More broadly a lot of this complication and pain is because `gen_common_index` probably needs a moderate refactor. I proposed one approach to potentially simplify things https://linear.app/rerun/issue/RR-2765/change-python-public-api-surface-pattern-to-support-easier-docs. On some level our python api documentation is a little rough because our layout of our python package is a little arbitrary/organic then we do a ton of re-exports to the top level.

One thing I've done in the past is a package walk to enumerate all classes then do some test on that (we could probably compare the index we generate to that). However that generally is a bit brittle.

PR #11713, opened 2025-10-30
Merge `re_format_arrow` into `re_arrow_util`

abey79:
We had a (soft) dependency cycle between `re_arrow_util` and `re_format_arrow`, via `dev-dependencies`. Despite this being theoretically allowable, including for crate publishing, it made me loose literal days to try to work around.

This pr does what the title says.

Immediate follow up PR will add a CI check for dependency cycle detection.

- sibling https://github.com/rerun-io/dataplatform/pull/1942


crates/store/re_types_core/src/component_batch.rs
```diff
@@ -257,6 +258,24 @@ impl From<SerializedComponentBatch> for SerializedComponentColumn {
...
+    lengths: impl IntoIterator<Item = usize>,
+) -> arrow::error::Result<ListArray> {
+    let (field, _offsets, values, _nulls) = list_array.into_parts();
+
+    let offsets = OffsetBuffer::from_lengths(lengths);
+    let nulls = None;
+
+    ListArray::try_new(field, offsets, values, nulls)
+}
+
```
abey79, line 278: I had to move this here to break a cycle. It wasn't used anywhere else anyways.

PR #11714, opened 2025-10-30
Bump `arrow` to `56.1` and `datafusion` to `50`

teh-cmc:
Revival of https://github.com/rerun-io/rerun/pull/11248 which was reverted in https://github.com/rerun-io/rerun/pull/11712.

* DNM: Blocked on sibling: https://github.com/rerun-io/dataplatform/pull/1859

---

### Related

* Currently blocking #11221.
* Supersedes #11159.

### TODO

* [x] Double check `.expect()`. (copied over from #11159)
* [x] Pin `datafusion` to `50.0.0`?
* [x] Finish https://github.com/rerun-io/dataplatform/pull/1859
* [x] `@rerun-bot full-check`

### Reviewing

* Please expand files like `package_lock.json` (collapsed on GitHub) to see if everything makes sense.
* For good measure, please `git diff` against #11159 too.

PR #11719, opened 2025-10-30
Drop official support for Intel Macs

emilk:
### What
* Do not build any artifacts for macOS x86
* Do not run any CI tests macOS x86

### Why
Supporting Intel Mac is more trouble than its worth.

The recent Arrow update caused a build failure:
* https://github.com/rerun-io/rerun/pull/11714#issuecomment-3467868555
* https://github.com/rerun-io/rerun/actions/runs/18941355578/job/54080863849

Since nobody at Rerun has an Intel mac, fixing this issue is difficult. It also means we cannot test the viewer to guarantee it works as expected.

Furthermore, macOS x86 was recently demoted to a Tier 2 target for the Rust compiler ([link](https://blog.rust-lang.org/2025/08/19/demoting-x86-64-apple-darwin-to-tier-2-with-host-tools/)).

For reference, the first Arm Mac (M1) was launched in 2020.

### But I use an Intel Mac!
You can still build Rerun from source (but perhaps not with all features enabled). There are instructions in [`BUILD.md`](https://github.com/rerun-io/rerun/blob/main/BUILD.md).


emilk:
Good catch! My regexp-fu was not strong enough 😞

Issue #8340, opened 2024-12-06
Speed up creation of `DataQueryResultTree`


Wumpf:
Today, we spend a looot of time updating the tree of all "data results" for scenes with many entities.
This code has seen a lot of organic growth, and while I haven't dived in much yet it is resonable to expect that there is a lot of low hanging fruit.

That said, we need to look deeper in what is there, what can be made incremental and which concepts even make sense.

joelreymont:
How should performance be measured here?

That is, how do I tell that I have improved performance significantly?

joelreymont:
PR #11720 may improve performance by x3.5 here.

PR #11720, opened 2025-10-30
Small performance improvement for many entities

joelreymont:
Improves #8340

1. O(1) visualizer lookup instead of O(V) iteration
2. Single-pass filter evaluation instead of three separate passes

### DIFF
--- a/crates/store/re_log_types/src/path/entity_path_filter.rs
+++ b/crates/store/re_log_types/src/path/entity_path_filter.rs
@@ -1389,6 +1389,10 @@ mod tests {
         let eval = filter.evaluate(&EntityPath::from("/world"));
         assert!(eval.subtree_included, "/world subtree should be included");
         assert!(eval.matches, "/world should match");
+        assert!(
+            !eval.matches_exactly,
+            "/world should not match exactly (matched by /world/** subtree rule)"
+        );
 
         // Test /world/house - should be included by /world/**
         let eval = filter.evaluate(&EntityPath::from("/world/house"));
@@ -1397,6 +1401,10 @@ mod tests {
             "/world/house subtree should be included"
         );
         assert!(eval.matches, "/world/house should match");
+        assert!(
+            !eval.matches_exactly,
+            "/world/house should not match exactly (matched by /world/** subtree rule)"
+        );
 
         // Test /world/car - should not match but subtree should be included (has exception deeper)
         let eval = filter.evaluate(&EntityPath::from("/world/car"));
@@ -1405,6 +1413,10 @@ mod tests {
             "/world/car subtree should be included (has /world/car/driver exception)"
         );
         assert!(!eval.matches, "/world/car should not match");
+        assert!(
+            !eval.matches_exactly,
+            "/world/car should not match exactly (excluded by - /world/car/**)"
+        );
 
         // Test /world/car/driver - should be included
         let eval = filter.evaluate(&EntityPath::from("/world/car/driver"));
@@ -1425,11 +1437,19 @@ mod tests {
             "/world/car/hood subtree should be excluded"
         );
         assert!(!eval.matches, "/world/car/hood should not match");
+        assert!(
+            !eval.matches_exactly,
+            "/world/car/hood should not match exactly (excluded)"
+        );
 
         // Test /other - should be excluded (no matching rule)
         let eval = filter.evaluate(&EntityPath::from("/other"));
         assert!(!eval.subtree_included, "/other subtree should be excluded");
         assert!(!eval.matches, "/other should not match");
+        assert!(
+            !eval.matches_exactly,
+            "/other should not match exactly (no matching rule)"
+        );
 
         // Test exact match without subtree
         let filter = EntityPathFilter::parse_forgiving(
@@ -1455,6 +1475,10 @@ mod tests {
             "/world/house should not be included (parent has no subtree rule)"
         );
         assert!(!eval.matches, "/world/house should not match");
+        assert!(
+            !eval.matches_exactly,
+            "/world/house should not match exactly (no rule for this path)"
+        );
 
         // Test subtree_included when there's an include rule deeper in the tree
         let filter = EntityPathFilter::parse_forgiving(
@@ -1470,6 +1494,10 @@ mod tests {
             "/world should have subtree_included=true because /world/car/driver is included"
         );
         assert!(!eval.matches, "/world should not match directly");
+        assert!(
+            !eval.matches_exactly,
+            "/world should not match exactly (no rule for this path)"
+        );
 
         let eval = filter.evaluate(&EntityPath::from("/world/car"));
         assert!(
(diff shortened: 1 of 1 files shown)


tests/rust/test_query_tree_bench/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
```
Wumpf: this shouldn't be a separate crate but instead a `criterion` based benchmark in `crates/viewer/re_viewport_blueprint/benches` (or similar place)

tests/rust/test_query_tree_bench/src/main.rs
```diff
@@ -0,0 +1,374 @@
+use std::time::{Duration, Instant};
```
Wumpf: I don't get what you're benchmarking here. This doesn't use any of the data structures that were supposedly improved. It doesn't even depend on `re_viewport_blueprint`?!

Wumpf (changes requested):
didn't go through the details, but I can see that the list look-up inversion could help in some situations. However, I doubt that this is the main bottleneck of `DataQueryResultTree` build-up (is it?) and the benchmark you provided has seemingly nothing to do with the actual improvement?
Can you please add a benchmark that is actually checking on the respective data structure and show how it affects end-to-end per frame performance in select scenes (e.g. by providing before/after screenshot of frame profiling in puffin (see profiler option in the Rerun menu))

joelreymont:
I think there's at least a 10% improvement for complex queries here, e.g. benchmarking over the baseline (skipping the improvements commit). Please let me know if I'm mistaken!

```
data_query_tree/include_all
                        time:   [1.0338 ms 1.0350 ms 1.0362 ms]
                        thrpt:  [8.3382 Melem/s 8.3482 Melem/s 8.3574 Melem/s]
                 change:
                        time:   [+14.201% +14.640% +15.216%] (p = 0.00 < 0.05)
                        thrpt:  [-13.207% -12.770% -12.435%]
                        Performance has regressed.
...
Benchmarking data_query_tree/complex_filter: Collecting 100 samples in estimated 6.22data_query_tree/complex_filter
                        time:   [1.2033 ms 1.2157 ms 1.2386 ms]
                        thrpt:  [6.9756 Melem/s 7.1069 Melem/s 7.1805 Melem/s]
                 change:
                        time:   [-12.672% -11.961% -10.981%] (p = 0.00 < 0.05)
                        thrpt:  [+12.335% +13.586% +14.510%]
                        Performance has improved.
```

Puffin shows improvements on the many entities transform test, e.g. for baseline

many-entity-transform-baseline

vs improved 

many-entity-transform-update

crates/viewer/re_viewport_blueprint/benches/data_query.rs
```diff
@@ -0,0 +1,197 @@
...
+    StoreContext, ViewClassRegistry, VisualizableEntities, blueprint_timeline, Caches,
+    PerVisualizer,
+};
+use re_viewport_blueprint::ViewContents;
+
+// ---
+
+// `cargo test` also runs the benchmark setup code, so make sure they run quickly:
+#[cfg(debug_assertions)]
+mod constants {
```
Wumpf, line 23: we're not testing anything in here actually, so why not even lower values?

crates/viewer/re_viewport_blueprint/benches/data_query.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    group.finish();
+}
+
+// --- Helpers ---
+
+fn build_entity_tree() -> (EntityDb, PerVisualizer<VisualizableEntities>) {
+    use rand::Rng as _;
+    let mut rng = rand::rng();
```
Wumpf: not setting a seed makes the benchmark non-deterministic - the rng affects not just values but also visualizability here, so its important it stays the same

crates/viewer/re_viewport_blueprint/benches/data_query.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+#[cfg(not(debug_assertions))]
+mod constants {
+    pub const NUM_PARENTS: usize = 80;
+    pub const NUM_CHILDREN_PER_PARENT: usize = 18;
+    pub const NUM_GRANDCHILDREN_PER_CHILD: usize = 6;
+}
+
+#[expect(clippy::wildcard_imports)]
```
Wumpf: why not just do regular imports?

crates/viewer/re_viewport_blueprint/src/view_contents.rs
```diff
@@ -282,9 +283,14 @@ impl ViewContents {
 
         let mut data_results = SlotMap::<DataResultHandle, DataResultNode>::default();
 
+        let visualizers_per_entity = {
+            re_tracing::profile_scope!("visualizers_per_entity");
```
Wumpf: can be simplified by just putting `re_tracing::profile_function!();` inside of `visualizers_per_entity`

crates/viewer/re_viewport_blueprint/src/view_contents.rs
```diff
@@ -419,6 +446,56 @@ impl QueryExpressionEvaluator<'_> {
             None
         }
     }
+
+    fn evaluate_filter(&self, path: &EntityPath) -> FilterEvaluation {
```
Wumpf: this should be a method on `ResolvedEntityPathFilter` that comes with its own unit tests - those _are_ warranted here since this repeats some of the logic that's in existing functions today and I'm rather nervous that this misses subleties that lead to a change in behavior

crates/viewer/re_viewport_blueprint/src/view_contents.rs
```diff
@@ -419,6 +446,56 @@ impl QueryExpressionEvaluator<'_> {
...
+        let mut subtree_included = false;
+        let mut matches_exactly = false;
+        let mut last_match: Option<(RuleEffect, bool)> = None;
```
Wumpf: unclear from the name what the bool means

joelreymont:
Will address!

crates/store/re_log_types/src/path/entity_path_filter.rs
```diff
@@ -1287,6 +1363,95 @@ mod tests {
...
+            + /world/car/driver
+            ",
+        )
+        .resolve_forgiving(&subst_env);
+
+        // Test /world - should be included
+        let eval = filter.evaluate(&EntityPath::from("/world"));
+        assert!(eval.subtree_included, "/world subtree should be included");
+        assert!(eval.matches, "/world should match");
+
```
Wumpf, line 1396: why not check `eval.matches_exactly` also on all these evaluate runs? Unclear why it is checked sometimes and not other times

joelreymont:
Will fix!

joelreymont:
@Wumpf Merge?

PR #11721, opened 2025-10-30
Our IndexValuesLike type is overly strict. Allow np.datetime64 inputs

ntjohnson1:
### What
We accept indexes and time specs in a few different places and we seem to re-implement this handling in a few different ways. This doesn't resolve that but it does add a test highlighting the lack of ergonomics, then fixes it here.

A few that I am aware of are:
* Time column: https://github.com/rerun-io/rerun/blob/main/rerun_py/tests/unit/test_time_column.py
* Dataset entry start/stop: https://github.com/rerun-io/rerun/blob/7c3d2b05255883ef61d561c7923ea7716c402536/rerun_py/src/catalog/dataset_entry.rs#L229
* This here
* I know Andreas also mentioned a reference implementation for this but I think that was more internal than sdk python focused

I've contributed to some of the mess and I'll follow a ticket to try and unify our time handling so we can be as robust and flexible as possible while still being correct.


rerun_py/src/dataframe/type_aliases.rs
```diff
@@ -83,17 +83,55 @@ impl AnyComponentColumn {
...
+        }
+
+        // Try numpy i64 array
+        if let Ok(numpy) = obj.extract::<numpy::PyArrayLike1<'py, i64>>() {
+            return Ok(Self::NumPy(numpy));
+        }
+
+        // Check if this is a numpy array with datetime64 dtype
+        // First check if it has a dtype attribute to see if it's a numpy array
+        if let Ok(dtype) = obj.getattr("dtype") {
```
oxkitsune, line 109: Can this not be implemented on the python side? e.g. as a higher level wrapper around some rust internal structure?

ntjohnson1: It probably could but that's a much larger refactor than here.

Right now we have IndexValuesLike described in rust then each of the 
`filter_...` for two different classes each take that rust type in as argument and the conversion gets managed by the enum. If we made a Python wrapper around both RecordingView and DataFrameQueryView classes then we could take the inputs as IndexValuesLike and convert to int64 arrays and pass those to the underlying rust internal.

This is maybe the counterpoint to doing the wrapper classes everywhere.

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1620,6 +1622,47 @@ class TableInsertMode:
...
+    def len(self) -> int:
+        """
+        Get the number of unique index values.
+
+        Returns
+        -------
+        int
+            The number of unique index values.
+
+        """
```
abey79: `*Internal` object dont get their docstring checked against the rust version. So I generally recommend against having them here, because chances are they will rot in time.

rerun_py/src/dataframe/type_aliases.rs
abey79: Yeah, wrappers are indeed a bit "contagious". Which is good, we need them everywhere :) But it does require we once invest in a full refactor. Maybe worth leaving a todo/issue that this entire `IndexValueLike` think should be nuked from rust-side?

rerun_py/rerun_bindings/rerun_bindings.pyi
ntjohnson1: Actually the CI barfed on me because they didn't line up correctly. Let me try removing them and if they get to green.

rerun_py/src/dataframe/type_aliases.rs
ntjohnson1: I will reference this doc from our general python cleanup project. Are we already all in agreement on refactoring to the full wrapper pattern?

rerun_py/rerun_bindings/rerun_bindings.pyi
abey79: That is weird. Well, it is absolutely possible that there are bugs in `python_check_signature.py`.

PR #11726, opened 2025-10-30
Add a 'url' property to catalog and catalog entries

swallez:
Add a `url` property to `Catalog` and `Entry` in the Python SDK.


timsaucer:
Ok, I think we've got the wrong url here. I was hoping to see where the table was stored at. When I tested this against the oss server and called

```python
for t in client.table_entries():
    print(t.url)
```

I got

```
rerun+http://localhost:51234/entry/1873553423F8D7853390f7616be4247a
rerun+http://localhost:51234/entry/18735534263EFDA13390f7616be4247f
```

I didn't see the storage url.

swallez:
Hmm... on datasets we have `manifest_url` that returns something like `memory:///187355C59740F1A6119bb23ba40e2d73`.

Should we add it to tables too? Using the same pattern `memory:///<entry.id>`?

And what is it used for? 😅

timsaucer:
No - I wanted to get the physical storage url. I was hoping it was already available. I'm going to check the redap server to see what it gives.

timsaucer:
Targeting your branch to show how I would expect it to work for tables: https://github.com/rerun-io/rerun/pull/11734

Still need to do datasets

swallez:
I merged #11734 in this branch. Let's work here.

I now understand what URLs we're talking about. I initially thought it was those displayed in the viewer's dataset list (`rerun+http://<host>/entry/<id>`) when it's actually the asset's physical URL.

To avoid ambiguities I'll rename it from `url` to a more distinctive `asset_url` (unless there's an existing terminology for it).

swallez:
Closing this, work will continue in #11734

PR #11730, opened 2025-10-30
Add experimental `child_frame` & `parent_frame` to `Transform3D`

Wumpf:
### Related

* Fixes RR-2679
* Fixes RR-2684

### What

EDIT: fields got renamed in https://github.com/rerun-io/rerun/pull/11770

Another crucial stepping stone towards ROS style transform support!

`source` & `target` are singular right now in the API (since all of `Transform3D` is), but the underlying datastructures are already unit-tested against multiple sources (alas, not targets though; sources is where all the complexity is though anyways).

The doc of these new fields explains what's going on here:

```
  /// ⚠️ Experimental ⚠️: The frame this transform transforms from.
  ///
  /// The entity at which the transform relationship of any given source frame is specified musn't change over time.
  /// E.g. if you specified the source "robot_arm" on an entity named "my_transforms", you may not log transforms
  /// with the source "robot_arm" on any other entity than "my_transforms".
  /// An exception to this rule is static time - you may first mention a source on one entity statically and later on
  /// another one temporally.
  ///
  /// ⚠️ This also affects the source frame of [archetype.Pinhole] & [archetype.PoseTransforms3D].
  ///
  /// If not specified, this is set to the implicit transform frame of the current entity path.
  /// This means that if a [archetypes.Transform3D] is set on an entity called `/my/entity/path` then this will default to `tf#/my/entity/path`.
  ///
  /// To set the frame an entity is part of see [archetypes.CoordinateFrame].
  // TODO(RR-2777): Mark as stable when we're ready for it.
  source_frame: rerun.components.TransformFrameId ("attr.rerun.component_optional", nullable, order: 2000);

  /// ⚠️ Experimental ⚠️: The frame this transform transforms to.
  ///
  ///
  /// ⚠️ This also affects the target frame of [archetype.Pinhole].
  ///
  /// If not specified, this is set to the implicit transform frame of the current entity path's parent.
  /// This means that if a [archetypes.Transform3D] is set on an entity called `/my/entity/path` then this will default to `tf#/my/entity`.
  ///
  /// To set the frame an entity is part of see [archetypes.CoordinateFrame].
  // TODO(RR-2777): Mark as stable when we're ready for it.
  target_frame: rerun.components.TransformFrameId ("attr.rerun.component_optional", nullable, order: 2100);
```

⚠️ `TransformForest` does not yet walk arbitrary frames and is still bound to implicit frames. Addressing this is the next step after.
As a consequence the added example is unfinished and a little bit "weird": we can define `Transform3D` outside of the entity hierarchy, but we can not yet collapse said entity hierarchy!

image

---

This PR unfortunately makes queries a bit slower again as captured by existing benchmarks. Datastructure build-up is roughtly the same (I had fairly big swings on that one without doing anything)

Windows Ryzen 9 7950X3D comparing against a previous run on main (dab04a695674e3235668f4d7b1d01d41d220e765) via `pixi run cargo bench -p re_tf `
```
Benchmarking build_from_entity_db: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 35.9s, or reduce sample count to 10.
build_from_entity_db    time:   [347.14 ms 353.18 ms 359.98 ms]
                        change: [-1.5899% +0.4380% +2.7028%] (p = 0.70 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  3 (3.00%) high mild
  7 (7.00%) high severe

Benchmarking query_uncached_frame: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 37.5s, or reduce sample count to 10.
query_uncached_frame    time:   [26.557 ms 27.016 ms 27.504 ms]
                        change: [+36.134% +39.507% +42.906%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild

query_cached_frame      time:   [30.762 ns 30.990 ns 31.222 ns]
                        change: [-2.1408% -0.4344% +1.0960%] (p = 0.62 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe
```

Still a far cry from how slow it was before https://github.com/rerun-io/rerun/pull/11655

### DIFF
--- a/rerun_py/tests/unit/test_archetype_str.py
+++ b/rerun_py/tests/unit/test_archetype_str.py
@@ -17,6 +17,8 @@ import rerun as rr
                 "  scale=[],\n"
                 "  mat3x3=[],\n"
                 "  relation=[],\n"
+                "  source_frame=[],\n"
+                "  target_frame=[],\n"
                 "  axis_length=[]\n"
                 ")"
             ),
@@ -31,6 +33,8 @@ import rerun as rr
                 "  scale=[],\n"
                 "  mat3x3=[],\n"
                 "  relation=[],\n"
+                "  source_frame=[],\n"
+                "  target_frame=[],\n"
                 "  axis_length=[]\n"
                 ")"
             ),


Wumpf:
@rerun-bot full-check

crates/store/re_types/definitions/rerun/archetypes/transform3d.fbs
```diff
@@ -43,11 +44,42 @@ table Transform3D (
...
+  /// An exception to this rule is static time - you may first mention a source on one entity statically and later on
+  /// another one temporally.
+  ///
+  /// ⚠ This also affects the source frame of [archetypes.Pinhole] & [archetypes.InstancePoses3D].
+  ///
+  /// If not specified, this is set to the implicit transform frame of the current entity path.
+  /// This means that if a [archetypes.Transform3D] is set on an entity called `/my/entity/path` then this will default to `tf#/my/entity/path`.
+  ///
+  /// To set the frame an entity is part of see [archetypes.CoordinateFrame].
+  // TODO(RR-2777): Mark as stable when we're ready for it.
```
Wumpf, line 64: does the `attr.rerun.state: unstable` actuall work on fields? Use that?

crates/store/re_tf/src/entity_to_source_frame_tracking.rs
```diff
@@ -0,0 +1,288 @@
+use nohash_hasher::IntSet;
+use re_log_types::{AbsoluteTimeRange, EntityPath, TimeInt};
+use re_types::TransformFrameIdHash;
+use std::collections::BTreeMap;
+use std::ops::Range;
+use vec1::smallvec_v1::SmallVec1;
+
+/// Datastructures for tracking which transform relationships are specified by any moment in time for a given entity.
```
oxkitsune, line 8:
```suggestion
/// Datastructures for tracking which transform relationships are specified at any moment in time for a given entity.
```

crates/store/re_tf/src/entity_to_source_frame_tracking.rs
```diff
@@ -0,0 +1,288 @@
...
+            ))
+            .collect(),
+            all_sources: std::iter::once(fallback_source_frame).collect(),
+        }
+    }
+
+    /// Insert a new range-start for a set of sources.
+    ///
+    /// Every time the start of a range is inserted, an existing range gets split in two!
+    /// The return value is the range previously not using `new_sources` but now is.
```
oxkitsune, line 50:
```suggestion
    /// The return value is the range that was not previously using `new_sources`, but is now.
```
IIUC, the retrun value is the range that wasn't previously using the `new_sources`, but now is.
Wording it like this is a bit easier to understand.

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -64,21 +65,46 @@ pub struct SourceToTargetTransform {
...
 pub struct CachedTransformsForTimeline {
-    /// Which entities logged transforms about which source frame when.
-    ///
-    /// For any source mentioned, there should be an entry in [`Self::per_source_frame_transforms`].
+    /// Maps entity paths to [`EntityToAffectedSources`] datastructures.
     ///
-    /// Note that an entity may have information about multiple different source frames, varying over time.
-    /// (that said, we should optimize for the common case where an entity always informs about the same source)
-    per_entity_source_information: IntMap<EntityPath, SourceFrameUpdates>,
+    /// This allows us to keep track of which incoming (or removed) transform events on entities apply to which
```
oxkitsune:
```suggestion
    /// This allows us to keep track of which incoming (or removed) transform events on entities, apply to which
```

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -616,54 +766,115 @@ impl TransformResolutionCache {
...
-                min_time,
-                || time_column.times(),
-                entity_path,
-            );
+            // Now that our map of active sources is up to date, we can insert "event points" (invalidated cache entries)
+            // into the respective per-source data structures.
+            for (time_range, source_frames) in
+                affected_sources.iter_ranges(time_column.time_range())
+            {
+                // We can now filter out the times on the time column that we're interested in.
```
oxkitsune: We're not filtering these out right? We're filtering based on the time column we're interested in

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -840,57 +1070,56 @@ impl TransformResolutionCache {
...
+    }
 }
 
 /// Queries all components that are part of pose transforms, returning the transform from child to parent.
 ///
-/// If any of the components yields an invalid transform, returns a `Affine3A::ZERO`.
-/// (this effectively disconnects a subtree from the transform hierarchy!)
+/// If any of the components yields an invalid transform, returns `None`.
 // TODO(#3849): There's no way to discover invalid transforms right now (they can be intentional but often aren't).
 fn query_and_resolve_tree_transform_at_entity(
```
oxkitsune, line 1116: I feel like this is a bit scary, similar to #8952 people could be depending on `Affine3A::ZERO` for invalid transforms.

I think this is the right thing to do, but might be worth to make a note of this somewhere in the changelog.

Wumpf: agreed, the whole "what is an invalid transform" thing was a bit vague before still is :/

crates/store/re_tf/src/transform_resolution_cache.rs
Wumpf: comment is a bit misleading. It means to say that we look at all the times of the time column that are relevant for the range in question. Fixing it

PR #11732, opened 2025-10-30
Increase size of windows runners 

emilk:
### Related
* Also reverts https://github.com/rerun-io/rerun/pull/11680 because:

image

### What
The release job is just so painfully slow, and Windows is among the most slow parts of it. Maybe this will help a bit.


abey79 (approved):
FWIW, my recent tests with linux runners indicated that there is nothing to gain to go from 16 to 32, but still worth trying for windows 🤷🏻

PR #11734, opened 2025-10-30
set url from provider details

timsaucer:
Here is an example of what I was hoping to get. It's partially based on [this slack comment](https://rerunio.slack.com/archives/C045J1Z7DU7/p1761067484597669?thread_ts=1761065326.836149&cid=C045J1Z7DU7)

With this I get:

```python
for e in client.table_entries():
    print(e.url)
```

produces

```
file:///Users/tsaucer/src/rerun/tests/assets/table/lance/simple_datatypes/

```

As I would expect. The second entry is the system table which is in memory and doesn't have an associated URL, so I do expect it is blank. We could even improve that to return a None.


timsaucer:
This did require I make some changes in the OSS server, so we would have to be sure to do the same in our dataplatform repo.

timsaucer:
FYI If this PR goes in it will impact this work:  https://github.com/rerun-io/rerun/pull/11735

PR #11735, opened 2025-10-30
change ProviderDetails from trait to enum

timsaucer:
This PR changes the `ProviderDetails` extension from a trait into an enum. This makes usage down the pipe easier in a variety of contexts.

### DIFF
--- a/crates/store/re_server/src/rerun_cloud.rs
+++ b/crates/store/re_server/src/rerun_cloud.rs
@@ -25,7 +25,7 @@ use re_protos::{
         ScanPartitionTableResponse, ScanTableResponse,
         ext::{
             self, CreateDatasetEntryRequest, CreateDatasetEntryResponse, CreateTableEntryRequest,
-            CreateTableEntryResponse, DataSource, DatasetDetails, LanceTable, ProviderDetails,
+            CreateTableEntryResponse, DataSource, DatasetDetails, ProviderDetails,
             ReadDatasetEntryResponse, ReadTableEntryResponse, TableInsertMode,
             UpdateDatasetEntryRequest, UpdateDatasetEntryResponse,
         },


crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -1131,10 +1130,18 @@ impl RerunCloudService for RerunCloudHandler {
             return Err(tonic::Status::invalid_argument("Missing provider details"));
         };
         #[cfg_attr(not(feature = "lance"), expect(unused_variables))]
-        let lance_table = LanceTable::try_from_any(&provider_details)?
-            .table_url
-            .to_file_path()
-            .map_err(|()| tonic::Status::invalid_argument("Invalid lance table path"))?;
+        let lance_table = {
```
jleibs: Now that we have an enum, an exhaustive match statement like:
```
match ProviderDetails::try_from_any(&provider_details)? {
```
that specifically returned not support for the `ProviderDetails::SystemTable` case would be a bit more clear and future proof for if/when we add ParquetTable support.

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -1331,11 +1338,19 @@ impl RerunCloudService for RerunCloudHandler {
 
         let request: CreateTableEntryRequest = request.into_inner().try_into()?;
         let table_name = &request.name;
-        let provider_details = LanceTable::try_from_any(&request.provider_details)?;
+        let provider_details = ProviderDetails::try_from_any(&request.provider_details)?;
+        let table_url = match provider_details {
+            ProviderDetails::LanceTable(table) => table.table_url,
+            ProviderDetails::SystemTable(_) => {
+                return Err(tonic::Status::invalid_argument(
+                    "System table does not provide a URL",
```
jleibs: The issue here doesn't seem like it's that system table doesn't provide a URL, the problem is that a user isn't allowed to register a SystemTable.

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -1331,11 +1338,19 @@ impl RerunCloudService for RerunCloudHandler {
...
+                ));
+            }
+        };
         let schema = Arc::new(request.schema);
 
         let table = store
-            .create_table_entry(table_name, &provider_details.table_url, schema)
+            .create_table_entry(table_name, &table_url, schema)
```
jleibs: Given that this is hard-coded to a only work with lance, rather than extracting the url, this seems like it would be cleaner if we just passed through the LanceTable provider details directly.

Like this whole thing could be 
```
let schema = Arc::new(request.schema);

let table = match provider_details {
            ProviderDetails::LanceTable(lance_details) => {
                store..create_lance_table(table_name, &lance_details, schema)
             },
            ProviderDetails::SystemTable(_) => {
                return Err(tonic::Status::invalid_argument(
                    "Cannot create System Table"
                )
             }
        };
```

timsaucer:
I made the requested changes. I also figured that we should do this up right and switch out all the other places in proto_ext that are carrying around `Any` and turn them into `ProviderDetails`. I'm working on the corresponding PR for dataplatform.

crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
```diff
@@ -1118,21 +1126,62 @@ impl TryFrom<crate::cloud::v1alpha1::TableEntry> for TableEntry {
...
 
-pub trait ProviderDetails {
-    fn try_as_any(&self) -> Result<prost_types::Any, TypeConversionError>;
+#[derive(Debug, Clone)]
+pub enum ProviderDetails {
+    SystemTable(SystemTable),
+    LanceTable(LanceTable),
+}
+
+impl ProviderDetails {
```
teh-cmc: Why not use the standard `TryFrom`/`TryInto` then?

teh-cmc (approved):

crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
timsaucer: Good point. Changing.

timsaucer:
@rerun-bot full-check

timsaucer:
@rerun-bot full-check

Issue #4298, opened 2023-11-21
Parallelize data ingestion


teh-cmc:
(Random thoughts as I'm struggling with slow ingestion times)

Speaking of multi-threading: further down the line the bottleneck for real-time use cases is probably gonna be the speed of ingestion when using `connect()`, which is currently bounded by the time it takes to render the UI since they both run serially on the same thread (right?).

There are two main fronts that I can think of to improve things, though they are both far more challenging than parallelizing space-views:

1. Run insertion and UI rendering in parallel.
This requires a consistent view of the datastore for the entire duration of a frame, so results are consistent across views.
The standard approach would be to implement some kind of snapshotting/MVCC for the store and its views (which would actually be possible thanks to our `RowId`s), but thankfully unnecessary in our case if we follow our plan of handling all query work in one centralized place (just lock everything, get the data out, unlock, then render UI as usual).
This asynchronicity between the store and the UI rendering is something we need to move towards to in any case in order to ultimately get the database out in its own process (the "hub").

2. Parallelize insertions into the store.
This should not be too hard as long as we partition on `(EntityPath, Timeline)` so that it matches the natural indexing of the store (though Im sure we're going to discover some nasty race conditions on the way).
Similarly, most of our builtin store-subscribers should be capable of updating in parallel when following this partition scheme.

On the bright side, this also means that the upcoming work to multi-thread the UI rendering should actually improve ingestion speeds too by side-effect:
- https://github.com/rerun-io/rerun/issues/1325

teh-cmc:
- Blocked on #6066

teh-cmc:
We really need to move ingestion off of the UI thread -- this really makes using Rerun hard on some OSes / window managers (including mine).

emilk:
> We really need to move ingestion off of the UI thread -- this really makes using Rerun hard on some OSes / window managers (including mine).

An alternative (at least in the short term) is to ensure data ingestion still happens when Rerun is hidden:
* #7427

This _may_ be an easier fix, though I agree proper parallel data ingestion is more desirable.

teh-cmc:
Note blocked anymore: we have the chunks, and we have the storage handles.

Wumpf:
This naturally also happens when modal dialogs like the OS's save dialog is open

joelreymont:
There may be 2x improvements with a batch size of 10, with PR #11741

PR #11741, opened 2025-10-31
Parallel ingestion

joelreymont:
Improves #4298.

Strangely, it's batch size of 10 that's optimal (2x) for me. 

```
 name                                            baseline.txt ns/iter  current.txt ns/iter  diff ns/iter   diff %  speedup
 load_from_file_contents/rrd_loader              83,794,333            83,564,312               -230,021   -0.27%   x 1.00
 message_processing_batches/parallel_batch_10    72,747,021            36,069,833            -36,677,188  -50.42%   x 2.02
 message_processing_batches/parallel_batch_100   5,782,571             4,111,424              -1,671,147  -28.90%   x 1.41
 message_processing_batches/parallel_batch_1000  1,255,255             1,174,237                 -81,018   -6.45%   x 1.07
 message_processing_batches/parallel_batch_50    11,089,879            7,294,523              -3,795,356  -34.22%   x 1.52
 message_processing_batches/parallel_batch_500   1,857,767             1,488,916                -368,851  -19.85%   x 1.25
 message_processing_batches/sequential           693,867               693,470                      -397   -0.06%   x 1.00
```

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -2231,13 +2231,59 @@ impl App {
         #[cfg(not(target_arch = "wasm32"))]
         if let LogMsgKind::Arrow = msg_kind {
             if let LogMsg::ArrowMsg(_, arrow_msg) = msg {
-                if self.try_enqueue_arrow_ingestion(
-                    store_id,
-                    arrow_msg,
-                    channel_source.clone(),
-                    msg_will_add_new_store,
-                ) {
-                    return;
+                // In test mode, process synchronously to avoid race conditions
+                // In production, use async processing for better UI responsiveness
+                match self.app_env {
+                    crate::AppEnvironment::Test => {
+                        // Process synchronously in tests
+                        match re_entity_db::PreparedArrowChunk::from_arrow_msg(arrow_msg) {
+                            Ok(prepared) => {
+                                let (was_empty, entity_db_add_result) = {
+                                    let entity_db = store_hub.entity_db_mut(store_id);
+                                    let was_empty = entity_db.is_empty();
+                                    let result = entity_db.add_prepared_arrow_chunk(&prepared);
+                                    (was_empty, result)
+                                };
+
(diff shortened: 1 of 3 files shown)


crates/viewer/re_viewer/src/app.rs
```diff
@@ -2275,6 +2406,163 @@ impl App {
...
+        store_id: &StoreId,
+        arrow_msg: &re_log_types::ArrowMsg,
+        channel_source: Arc<SmartChannelSource>,
+        msg_will_add_new_store: bool,
+    ) -> bool {
+        let arrow_msg = arrow_msg.clone();
+
+        let promise = poll_promise::Promise::spawn_thread("prepare_arrow", move || {
+            PreparedArrowChunk::from_arrow_msg(&arrow_msg)
+        });
```
Wumpf: thread spawning is very expensive and for every single message you spawn a new one where the only thing it does is to unpack the chunk from the message?

Wumpf (changes requested):
After scrolling a bit over this I decided to dismiss this PR without doing a deep review. Reasons:
* rather dubious thread spawning strategy: the stated goal of the underlying ticket was to rework ingestion as a whole to as much as possible run in a background thread parallel to the ui. Instead, what seems to happen here is that _for every message_ a new thread is spawned that decodes messages into chunks which then have to be picked up later
    * with that strategy it is to be expected to have performance getting worse for small messages coming in as it happens in your benchmarks if I understood correctly
    * what we most likely need instead is more of a work queue design (likely with a backpressure mechanism)
* makes a tech debt riddled area of our code worse by adding another layer & new concepts like "prepared arrow chunks" and `LogMsgKind`
   * overall found it hard to follow
* strongly divergent code for wasm
* divergent code for test environment?!

I do think there's a path to make a somewhat similar process fly, but I don't think we have the capacity to guide a new contributor to make these kind of deep changes with this as a starting point.

Thanks for your understanding!

joelreymont:
Will redesign using a work queue with back pressure and resubmit!

Issue #7115, opened 2024-08-08
Rerun viewer becomes slow when there are several hundreds of entities


emilk:
We currently don't have a good example of this. The best we got is

`pixi run py-build-examples && pr -e examples open_photogrammetry_format --no-frames`

and maybe

* https://github.com/rerun-io/rerun/pull/5449

jleibs:
Here's a very simple example that makes the viewer totally unuable:
```python
import rerun as rr

from rerun.datatypes import RotationAxisAngle

rr.init("rerun_example_many_entities", spawn=True)


def one_strip(number_points: int):
    return rr.LineStrips3D(
        [[0.3 if i % 2 == 0 else -0.3, i, 0] for i in range(number_points)],
    )


def rotate(theta: float):
    return rr.Transform3D(
        rotation_axis_angle=RotationAxisAngle(axis=[0, 0, 1], degrees=theta),
        axis_length=0,
    )


def draw_lines(lines_count: int, pts_per_line: int):
    theta_delta = 360 / lines_count
    for i in range(lines_count):
        rr.log(f"line_{i}", rotate(theta_delta * i))
        rr.log(f"line_{i}/strip", one_strip(pts_per_line))

draw_lines(5000, 10)
```

joelreymont:
I'm trying to get a feel for what is slow. I tried to patch some things and cache others and this is what I see. 

Is this slow for the air traffic example?

https://github.com/user-attachments/assets/7f8ce393-2dff-42e2-962f-3b84cadd70aa

Wumpf:
@joelreymont "is this slow" is hard to answer without any context like what you're testing, what you're machine running on etc.. A blurry video recording is definitely not sufficient ;-). Also as far as I can tell from that one you're running in debug mode and while we do care about performance of unoptimized builds that's more of an afterthought and not what we profile against.
Today we generally think that airtraffic is an example that is too slow on both frame time/rendering as well as ingestion with exact numbers varying a lot on what's on screen of course.
If you want to explore _what_ takes time check the profiler in the menu

joelreymont:
Took a stab with PR #11743. See related PRs #11741 and #11720.

joelreymont:
[This branch](https://github.com/joelreymont/rerun/tree/all-together) combines the three PRs above.

PR #11743, opened 2025-10-31
Optimize performance for hundreds of entities (#7115)

joelreymont:
Improves #7115.

I'm not sure if I'm paying attention to the right metric but I decided to pay attention to the `VisualizerSystem::execute` scope. This is the baseline of main when running the RRD generated by `tests/python/many_entity_transforms`.

7715-many-transform-base

There's a fair bit of improvement after this PR. 

7715-many-transform-new

### DIFF
--- a/crates/viewer/re_viewer/src/app_state.rs
+++ b/crates/viewer/re_viewer/src/app_state.rs
@@ -514,6 +514,8 @@ impl AppState {
                     view_states,
                 );
 
+                // Update cache with post-override results. Order doesn't matter here.
+                #[expect(clippy::iter_over_hash_type)]
                 for (view_id, result) in &query_results {
                     if let Some(entry) = self.view_query_cache.get_mut(view_id) {
                         entry.result = result.clone();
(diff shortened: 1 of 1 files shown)


joelreymont:
What UI metrics would I pay attention if I combine this with PRs #11720 and #11741?

crates/viewer/re_test_context/src/lib.rs
```diff
@@ -494,6 +494,9 @@ impl TestContext {
         let drag_and_drop_manager =
             re_viewer_context::DragAndDropManager::new(ItemCollection::default());
 
+        // Tests don't compute visualizable entities per view, so we provide an empty map
```
Wumpf, line 497: why not? future tests will likely trip over that then

crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -69,6 +72,12 @@ pub struct AppState {
     #[serde(skip)]
     pub(crate) share_modal: crate::ui::ShareModal,
 
+    #[serde(skip)]
+    view_query_cache: HashMap<ViewId, ViewQueryCacheEntry>,
+
+    #[serde(skip)]
+    visualizable_entities_cache: HashMap<ViewId, VisualizableEntitiesCacheEntry>,
```
Wumpf, line 82: need to find a better place for this. Probably best to refactor this together with `ViewStates`, but that exceeds the scope of the PR a bit.

So for starters I think these could go into a shared datastructure that lives on a cache struct that is `impl Cache` (and lazily created with other caches) so we can also easily oversee its memory usage
The viewer's cache trait isn't a perfect fit for this since it's more geared towards changes in the store rather than blueprint, but as good as any place.

Either way I'd like to avoid making the situation on `ApState` any worse

crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -139,6 +150,58 @@ pub(crate) struct WelcomeScreenState {
...
+        blueprint_generation: re_chunk_store::ChunkStoreGeneration,
+        space_origin: &EntityPath,
+    ) -> Self {
+        Self {
+            recording_generation,
+            blueprint_generation,
+            space_origin: space_origin.clone(),
+        }
+    }
+}
```
Wumpf, line 204: should all go to a separate module

crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -288,39 +351,97 @@ impl AppState {
...
+                                VisualizableEntitiesCacheEntry {
+                                    key: cache_key,
+                                    result: PerVisualizer(visualizable_entities.0.clone()),
+                                },
+                            );
+
+                            (view.id, visualizable_entities)
+                        })
+                        .collect()
+                };
```
Wumpf, line 402: all of this should be encapsulated by the cache itself

crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -139,6 +150,58 @@ pub(crate) struct WelcomeScreenState {
...
+
+struct VisualizableEntitiesCacheEntry {
+    key: VisualizableEntitiesCacheKey,
+    result: PerVisualizer<VisualizableEntities>,
+}
+
+#[derive(Clone, PartialEq, Eq, Hash)]
+struct VisualizableEntitiesCacheKey {
+    recording_generation: re_chunk_store::ChunkStoreGeneration,
+    blueprint_generation: re_chunk_store::ChunkStoreGeneration,
```
Wumpf: afaik blueprint generation now changes continuously when playing since we store the time cursor position as a static component in it. That makes this cache _a lot_ less effective and really only useful for a standstill viewer. Need to come up with a better cache key

Wumpf (commented):
partial review. I think this could be a promising direction, but needs to be designed a bit better.

Something I'd like to see more of in the PR descriptions is 
* what exactly has changed
* why has it changed
* what are the impacts under what circumstances

it's nice that you provided a before/after screenshot of the method list in puffin for the many-entity test script, but a bit more context would be needed to evaluate this
* how was this measured _exactly_
* what overall impact does it have on frame time - flamegraph before/after _can_ be useful
* what exactly got faster
* tradeoffs

joelreymont:
It will be done! 🫡

crates/viewer/re_viewer/src/app_state.rs
joelreymont: What about removing blueprint_generation from VisualizableEntitiesCacheKey?

Visualizable entities only depend on recording data (which entities exist and what components they have) and view configuration (space_origin, view class). They do not depend on the blueprint time cursor.
    
It seems we can safely remove blueprint_generation from VisualizableEntitiesCacheKey (app_state.rs:184-189) and keep recording_generation since new entities being added should invalidate the cache.  This should make the visualizable entities cache effective during playback.

What am I missing?

Issue #11747, opened 2025-10-31
Is Pixi the standard and only way to build cpp lib, instead of cmake?


chengguizi:
**Describe the annoyance**

All was working well when i was using cmake to build rerun, including `make rerun_sdk` and `cmake --install`, until 0.24.1.

I am trying out the latest 0.26.2, and encounter error

```
CMake Error at thirdparty/rerun/rerun_cpp/cmake_install.cmake:58 (file):
  file INSTALL cannot find
  "<repo root>/thirdparty/rerun/target_pixi//release/librerun_c.a":
  No such file or directory.
Call Stack (most recent call first):
  thirdparty/rerun/cmake_install.cmake:57 (include)
  cmake_install.cmake:47 (include)

```

When i look through the folder, the correct path is <repo root>/thirdparty/rerun/target/release/librerun_c.a. There is an extra `_pixi`

It seems that it is changed to this behaviour in https://github.com/rerun-io/rerun/commit/4f208a0f0b80630a006a6f8b9a93ee76a785981e commit.

May I know if pixi is the only way to build the newer versions? Or we can have a quick fix on this to make vanilla cmake building & installing process work as the older versions?

**To Reproduce**
1. git clone rerun repo
2. checkout 0.26.2
3. do standard `mkdir build && cd build`
4. perform cmake config `cmake -DCMAKE_INSTALL_PREFIX=/tmp`
5. the step shall fail

one of the error
```
make[2]: *** No rule to make target '<repo root>/target_pixi//release/librerun_c.a', needed by 'examples/cpp/incremental_logging/example_incremental_logging'.  Stop.

```

**Expected behavior**

- cmake workflow should work
- if it doesn't probably the release note should say it?

**Your goals**

- hopefully rerun is still nicely integrated with cmake, so to integrated into larger cpp project in the most compatible way, if possible

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04 X86

**Additional context**

Wumpf:
C++ builds still happen via CMake, but it seems that the in-repo CMake script for tracking the rerun_c dependency now assumes `target_pixi` as output path for building `rerun_c` from (rust) source code. With the "official" zip package we ship on our releases you won't run into that problem because it has rerun_c prebuilt so you can skip the Rust build.
But either way I think you're right `crates/top/rerun_c/CMakeLists.txt` is defacto broken since it invokes cargo for the wrong with the wrong target folder now causing issues 🤔 
[...]

PR #11751, opened 2025-10-31
Fix in-repo CMake build not resolving builds of rerun_c that weren't made with pixi

ntjohnson1:
### Related
* Fixes https://github.com/rerun-io/rerun/issues/11747

### What
Dynamically set the target dir based on the env var and default to the old location `target`


Wumpf:
Huh I must have messed up something when I tested this outside of pixi
* windows ci build has the same failure I had before
* vsstudio project files point to `D:\dev\rerun-io\rerun\\\debug\rerun_c.lib` so clearly the env var setting thing didn't fly

Wumpf:
TIL you can have exclamation marks in branch names

Wumpf:
that at least seems to set the correct dependency path in a non-pixi environment 🤷 

Edit: and it also worked now on my windows machine after a proper clean. The exact incantation:
```
mkdir build2
cmake -B build2 -S . -DCMAKE_BUILD_TYPE=Debug
cmake --build build2 --config Debug --target rerun_sdk_tests
```

Wumpf:
```
C:/a/rerun/rerun/C:/a/rerun/rerun/target_pixi/debug
```
So it seems that `CARGO_TARGET_DIR` may be an absolute path sometimes 🤔 
Maybe we can detect that and then act accordingly

PR #11755, opened 2025-11-03
Fix timeline line drawing bug on (at least some) Nvidia GPUs

Wumpf:
Fixes a curious rasterization issue we ran into before that shows up in the image comparision tests (there's similar code nearby to the site of the fix)

Running timeline_panel fixes on my Windows Nvidia machine without this fix:
time_panel_two_sections_with_two_valid_ranges_zoomed_out new

What is should look like / what it does look like with the fix:
time_panel_two_sections_with_two_valid_ranges_zoomed_out

Diff:
time_panel_two_sections_with_two_valid_ranges_zoomed_out diff

### DIFF
--- a/crates/viewer/re_time_panel/src/time_panel.rs
+++ b/crates/viewer/re_time_panel/src/time_panel.rs
@@ -1675,13 +1675,17 @@ fn paint_time_ranges_gaps(
         .cloned()
         .collect::<Vec<_>>();
 
+    // Margin for the (left or right) end of a gap.
+    // Don't use an arbitrarily large value since it can cause platform-specific rendering issues.
+    const GAP_END_MARGIN: f32 = 100.0;
+
     if let Some(segment_subrange) = valid_time_ranges.first() {
         let gap_edge = *segment_subrange.start() as f32;
+        let gap_edge_left_side = ui.ctx().content_rect().left() - GAP_END_MARGIN;
 
         if zig_zag_first_and_last_edges {
-            // Careful with subtracting a too large number here. Nvidia @ Windows was observed not drawing the rect correctly for -100_000.0
             // Left side of first segment - paint as a very wide gap that we only see the right side of
-            paint_time_gap(gap_edge - 10_000.0, gap_edge);
+            paint_time_gap(gap_edge_left_side, gap_edge);
         } else {
             // Careful with subtracting a too large number here. Nvidia @ Windows was observed not drawing the rect correctly for -100_000.0
             painter.rect_filled(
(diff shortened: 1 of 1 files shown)


crates/viewer/re_time_panel/src/time_panel.rs
```diff
@@ -1700,8 +1700,9 @@ fn paint_time_ranges_gaps(
     if let Some(segment_subrange) = valid_time_ranges.last() {
         let gap_edge = *segment_subrange.end() as f32;
         if zig_zag_first_and_last_edges {
+            // Careful with adding a too large number here. Nvidia @ Windows was observed not drawing the rect correctly for -100_000.0
```
Wumpf:
```suggestion
            // Careful with adding a too large number here. Nvidia @ Windows was observed not drawing the rect correctly for +100_000.0
```

crates/viewer/re_time_panel/src/time_panel.rs
```diff
@@ -1700,8 +1700,9 @@ fn paint_time_ranges_gaps(
     if let Some(segment_subrange) = valid_time_ranges.last() {
         let gap_edge = *segment_subrange.end() as f32;
         if zig_zag_first_and_last_edges {
+            // Careful with adding a too large number here. Nvidia @ Windows was observed not drawing the rect correctly for +100_000.0
             // Right side of last segment - paint as a very wide gap that we only see the left side of
-            paint_time_gap(gap_edge, gap_edge + 100_000.0);
+            paint_time_gap(gap_edge, gap_edge + 10_000.0);
```
emilk: Much less hacky would be to use `ui.ctx().content_rect().right + small_margin` for the right edge, where `small_margin` is slightly larger than the thickness of the gap (100.0 should be sufficient).

crates/viewer/re_time_panel/src/time_panel.rs
```diff
@@ -1700,8 +1700,9 @@ fn paint_time_ranges_gaps(
...
         } else {
             painter.rect_filled(
                 Rect::from_min_max(pos2(gap_edge, top), pos2(gap_edge + 100_000.0, bottom)),
```
emilk:
```suggestion
                Rect::from_min_max(pos2(gap_edge, top), pos2(gap_edge + 10_000.0, bottom)),
```

emilk: Actually, there are more 100_000.0 is this function

Wumpf: Yikes, how did I not notice

Wumpf:
there, much better!

PR #11756, opened 2025-11-03
Transform computations are internally now using double precision

grtlr:
### Related

* Closes RR-2493.

### What

This makes the computations within the `re_tf` crate double precision. Existing single precision components as casted to double. The results of the computation are then casted to single precision in the visualizers before rendering.

### DIFF
--- a/crates/store/re_tf/src/transform_queries.rs
+++ b/crates/store/re_tf/src/transform_queries.rs
@@ -15,7 +15,10 @@ use re_types::{
 };
 use vec1::smallvec_v1::SmallVec1;
 
-use crate::{PoseTransformArchetypeMap, ResolvedPinholeProjection};
+use crate::{
+    PoseTransformArchetypeMap, ResolvedPinholeProjection,
+    transform_resolution_cache::SourceToTargetTransform,
+};
 
 /// Lists all archetypes except [`archetypes::InstancePoses3D`] that have their own instance poses.
 // TODO(andreas, jleibs): Model this out as a generic extension mechanism.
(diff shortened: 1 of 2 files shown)


crates/store/re_tf/src/transform_forest.rs
```diff
@@ -621,8 +621,8 @@ fn compute_root_from_archetype(
 
 fn pinhole3d_from_image_plane(
```
grtlr, line 624: A lot of the casting comes from `ResolvedPinholeProjection` storing `rerun::components` internally. I did not change that, because we get quite a bit of convenience from them.

crates/store/re_types/src/components/pose_rotation_axis_angle_ext.rs
grtlr, line 1: The `impl`s in this and the next files, also kind of document where our current data model is not flexible enough (or where we want to change things in the future).

emilk:
How does this affect our benchmarks?

grtlr:
Comparison on `build_from_entity_db` (run via `cargo bench -p re_tf`):
* `main` (3ba05a7dd5c98a4edcccdf20d8a0710550d9a529) ~311.96 ms.
* this branch (896afeb049ad38f0698c57c126777855571f7b78) ~288.16 ms.

Yes it looks like that got faster. Yes it ~was~ is confusing to me too. Looking at some profiles now to see if the `as_affine3a` in the visualizers makes much of a difference (but it's probably negilable).

Wumpf (approved):
lgtm!
I do wonder though whether the `Into` from f64 glam types to f32 Rerun components are gonna bite us later: when `Transform3D`'s interface becomes more generic we will want to have `from_translation(glam::DVec3(1,2,3))` just do the right thing and that becomes more difficult with those conversions 🤔

grtlr:
Regarding visualizer performance, I've looked at the profiler and also can't see anything show up in particular.

Issue #11745, opened 2025-10-31
Time selection behaves unexpectedly on log_tick and frame timelines


natasha-sk:
When attempting to move an existing time selection by dragging it, instead of moving, the selection area expands. The log_time timeline works correctly.

1. Switch to log_tick or frame timeline
2. Create a time selection
3. Attempt to move the time selection by clicking and dragging within the selected area
4. Observe that instead of moving, the selection expands

[https://drive.google.com/file/d/1HgsQzPVO15Qsg0xNRMMCbTNz4hxeARt_/view?usp=sharing](url)

![Image](https://github.com/user-attachments/assets/87bb5150-149a-402f-882d-eaaa9d14c7d9)

emilk:
What version of rerun is this?

natasha-sk:
latest release 026.2
and it's the same on main

PR #11757, opened 2025-11-03
Hold down shift for snap to grid when moving time cursor

emilk:
In the time panel, hold down shift when moving the time cursor (or selecting a loop region) to snap against a coarse grid. This makes it easier to pick nice, round time values.

* Fixes https://github.com/rerun-io/rerun/issues/11745

### DIFF
--- a/crates/viewer/re_time_panel/src/time_panel.rs
+++ b/crates/viewer/re_time_panel/src/time_panel.rs
@@ -1483,6 +1483,7 @@ fn help(os: egui::os::OperatingSystem) -> Help {
             "Select time segment",
             (icons::SHIFT, "+", "drag time scale"),
         )
+        .control("Snap to grid", icons::SHIFT)
         .control("Pan", (icons::LEFT_MOUSE_CLICK, "+", "drag event canvas"))
         .control(
             "Zoom",
--- a/crates/viewer/re_time_panel/tests/snapshots/help_view_timeline_windows.png
+++ b/crates/viewer/re_time_panel/tests/snapshots/help_view_timeline_windows.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:2c9b6d526075623fb4b36c71a705f887f23c7919d998e6f3cdc3c96ddfe4157b
-size 21389
+oid sha256:bfc51d02a552432d2244bd3aeea31342f31227cddcac4e4f42bf7a790124d876
+size 23206


crates/store/re_log_types/src/index/absolute_time_range.rs
```diff
@@ -213,106 +213,6 @@ impl AbsoluteTimeRangeF {
...
-    ///
-    /// The simplest number is defined approximately as
-    /// "the number with the most trailing zeroes".
-    ///
-    /// This is used in GUI code.
-    ///
-    /// NOTE: this function only cares about simple _integers_.
-    /// We don't care about "smart" selection things smaller than a nanosecond or a single time step.
-    /// So if both ends are within the same integer, then [`Self::center`] is returned.
-    pub fn smart_aim(self) -> TimeReal {
```
emilk, line 227: This was an over-complicated way of doing snapping, that in the end resulted in worse behavior. The new algo uses the fact that the time panel is (locally) linear.

The original smart aiming algo came from egui/emath, but I improved it when I re-implemented it here.
These improvements have now been upstreamed: https://github.com/emilk/egui/pull/7680

(it still makes sense to use the smart-aiming approach in egui for logarithmic sliders and other non-linear situations)

crates/viewer/re_time_panel/src/time_panel.rs
```diff
@@ -1902,7 +1902,10 @@ fn time_marker_ui(
             time_commands.push(TimeControlCommand::SetTime(time));
             time_commands.push(TimeControlCommand::Pause);
 
-            x = pointer_pos.x; // avoid frame-delay
+            // Avoid frame-delay:
+            x = time_ranges_ui
+                .x_from_time_f32(time)
+                .unwrap_or(pointer_pos.x);
```
emilk, line 1909: Show _actual_ time position

crates/store/re_log_types/src/index/absolute_time_range.rs
Wumpf: nice!

crates/viewer/re_time_panel/src/time_ranges_ui.rs
```diff
@@ -520,3 +531,38 @@ fn test_time_ranges_ui_2() {
...
 }
+
+/// Find the largest value, smaller than `max`, that is "round" (even multiple of 10, or half that).
+fn time_snap_value_smaller_than(max: f64) -> i64 {
+    let max = max.abs();
+    if max <= 1.0 {
+        return 1;
+    }
+
+    if 1_000_000_000_000_000.0 < max {
```
Wumpf: why is this clamping needed? Seems like a rather arbitrary value to me

Wumpf (approved):
Very nice! But really hard to find this feature! Can we put this please into the time panel's help-text?

PR #11759, opened 2025-11-03
Do not create a 2nd Tokio runtime to run the server

swallez:
The `rerun` command's `main()` creates a Tokio runtime, and the server's `run()` does the same. This PR calls `server.run_async()` instead so that we don't create an additional runtime.

Additionally, should we remove the `run` function? It's not called anymore.


timsaucer:
Yes, I would say this PR should also remove `run` now that it is no longer needed. This is a leftover from when we first implemented the server as a stand alone application instead of part of the rerun app.

timsaucer:
@rerun-bot full-check

PR #11761, opened 2025-11-03
Time panel: pick sub-second precision based on zoom level

emilk:
### Related
* Part o [RR-2768](https://linear.app/rerun/issue/RR-2768/improve-time-selection-experience-on-a-timeline)

### What
Automatically pick number of sub-second decimals to show based on zoom level.

On `main`, moving the time cursor can cause the number of decimals to jump, which is very distracting. Now it changes when zooming instead.

dynamic-precision


Wumpf (approved):
very fancy. feels really nice! :)

thanks for keeping these improvement PRs self-contained and small!

PR #11762, opened 2025-11-03
Improve recently regressed common case transform ingestion

Wumpf:
### Related

* Follow-up to https://github.com/rerun-io/rerun/pull/11730

### What

Just before merging https://github.com/rerun-io/rerun/pull/11730 I ran the benchmarks again and noticed that regression on chunk ingestion was quite a bit worse than I'd expected. This follow-up addresses this!
Also, I noticed that the benchmark wasn't using mimalloc yet which curiously affects query performance measurement the most.

Rough perf journey here on `build_from_entity_db` benchmark - average numbers on my windows box:
* before source/target PR (but with mimalloc): 45.32 ms 
* after source/target PR (but with mimalloc): 306.31ms
* af207634702726b5348df9cfe61daf79013a9aa0 274.83 ms
* 3628cff9d550fc5916e30920ac2347d8d83a3fbb 108.25ms

So not glamorous in terms of recovery, but afaik still fine compared to 0.26 we're still a lot faster regardless on ingestion thanks to
* https://github.com/rerun-io/rerun/pull/11655

The biggest potential here really is to not evaluate all timelines when chunks come in, doing that will give us _a lot_ more budget. I.e. we should build up knowledge about a timeline only when someone asks for transforms for said timeline!

(which is also why this is exclude-changelog)

### DIFF
--- a/crates/store/re_tf/src/transform_resolution_cache.rs
+++ b/crates/store/re_tf/src/transform_resolution_cache.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet, hash_map::Entry};
 use std::ops::Range;
 
@@ -837,15 +838,20 @@ impl TransformResolutionCache {
                 // to the current entity path yield information about the sources in `source_frames`.
                 let times_with_potential_update = if time_column.time_range().min
                     >= time_range.start
+                    // Careful, we're comparing a std `Range` with `AbsoluteTimeRange`!
+                    // `max` is inclusive, `end` is exclusive.
+                    // The reason we have to use `Range` here over `AbsoluteTimeRange` is that `time_range` may contain `TimeRange::STATIC`.
                     && time_column.time_range().max < time_range.end
                 {
-                    time_column.times().collect_vec()
+                    Cow::Borrowed(time_column.times_raw())
                 } else {
-                    time_column
-                        .times()
-                        // TODO(andreas): For sorted time columns we could speed this up a bit.
-                        .filter(|time| time_range.contains(time))
-                        .collect_vec()
+                    Cow::Owned(
+                        time_column
+                            .times()
+                            // TODO(andreas): For sorted time columns we could speed this up a bit.
+                            .filter_map(|time| time_range.contains(&time).then_some(time.as_i64()))
+                            .collect_vec(),
+                    )
                 };
 
                 // Note down that all these source frames were updated at the given times.
@@ -865,7 +871,11 @@ impl TransformResolutionCache {
                     frame_transforms.insert_invalidated_transform_events(
                         aspects,
                         time_range.start,
-                        || times_with_potential_update.iter().copied(),
+                        || {
+                            times_with_potential_update
+                                .iter()
+                                .map(|t| TimeInt::new_temporal(*t))
+                        },
                         entity_path,
                     );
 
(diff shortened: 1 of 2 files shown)


crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -828,11 +835,18 @@ impl TransformResolutionCache {
...
                 // Note that there may be more times than actual relevant updates, but crucially, all queries
                 // to the current entity path yield information about the sources in `source_frames`.
-                let times_with_potential_update = time_column
-                    .times()
-                    // TODO(andreas): For sorted time columns we could speed this up a bit.
-                    .filter(|time| time_range.contains(time))
-                    .collect_vec();
+                let times_with_potential_update = if time_column.time_range().min
+                    >= time_range.start
+                    && time_column.time_range().max < time_range.end
```
emilk, line 844:
```suggestion
                let times_with_potential_update = if
                    time_range.contains_range(time_column.time_range())
```

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -828,11 +835,18 @@ impl TransformResolutionCache {
...
+                {
+                    time_column.times().collect_vec()
+                } else {
+                    time_column
+                        .times()
+                        // TODO(andreas): For sorted time columns we could speed this up a bit.
+                        .filter(|time| time_range.contains(time))
+                        .collect_vec()
```
emilk: If you replace `.times()` with `.times_raw()` (in both branches), you could completely avoid allocations in the first branch (e.g. using a `let times_with_potential_update: Cow<[i64]>`).

crates/store/re_tf/src/transform_resolution_cache.rs
emilk: (note that `contains_range` handles `max` slightly different - and probably more correctly!)

grtlr (approved):
I'm still trying to wrap my head around the details of the cache, but these changes make sense to me.

crates/store/re_tf/src/transform_resolution_cache.rs
Wumpf: then I have to swap over to i64 everywhere instead of `TimeInt`. But that sounds worth it. Thanks!

crates/store/re_tf/src/transform_resolution_cache.rs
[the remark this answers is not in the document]
Wumpf: you're confusing max & end here

Wumpf: the problem is that we're comparing a `Range` with an `AbsoluteTimeRange` here which makes things confusing. Having a look whether I can fix that at its core

Wumpf: that went well until I realized I do intentionally have static time in those ranges and I _do_ want to allow that. 
Adding a `contains_range_std` to `AbsoluteTimeRange` instead

Wumpf: nah that doesn't work. Okay too much hassle. Adding a clarifying comment instead.

crates/store/re_tf/src/transform_resolution_cache.rs
Wumpf: was trivial and gives 5.5% perf on `build_from_entity_db`. Also just feels good :D

PR #11763, opened 2025-11-03
Update references to Arrow conda packages in docs

amoeba:
### What

Updates references in the docs to the deprecated arrow-cpp conda package to refer to the current libarrow package. The conda package was renamed from arrow-cpp to libarrow and arrow-cpp in [2022](https://github.com/conda-forge/arrow-cpp-feedstock/commit/1487ea3e7d19e75cd3e905d9b871e112542f17a3). Rerun appears to be compatible with Arrow C++ 18 which is only available with Conda (and Pixi) as libarrow. The pixi lockfile references libarrow which is correct.

This PR also makes a couple of very small edits to some of docs where Arrow C++ is referred to.


amoeba:
Sure thing @ntjohnson1. Thanks for the quick merge.

PR #11765, opened 2025-11-03
Viewer login flows

jprochazk:
Adds login flows to the native and web viewer. They are available through the add/edit server modal.

### DIFF
--- a/crates/utils/re_auth/src/cli.rs
+++ b/crates/utils/re_auth/src/cli.rs
@@ -124,10 +124,8 @@ pub async fn login(options: LoginOptions) -> Result<(), Error> {
         .await
         .map_err(|err| Error::Generic(err.into()))?;
 
-    #[expect(unsafe_code)]
-    // SAFETY: Credentials come from the auth API
-    let credentials = unsafe { Credentials::from_auth_response(auth.into())? };
-    let credentials = credentials.ensure_stored()?;
+    // 5. Store credentials
+    let credentials = Credentials::from_auth_response(auth.into())?.ensure_stored()?;
 
     p.finish_and_clear();
 
--- a/crates/utils/re_auth/src/oauth.rs
+++ b/crates/utils/re_auth/src/oauth.rs
@@ -74,10 +74,7 @@ pub async fn refresh_credentials(
     );
 
     let response = api::refresh(&credentials.refresh_token).await?;
-    #[expect(unsafe_code)]
-    // SAFETY: credentials come from a trusted source
-    let credentials = unsafe { Credentials::from_auth_response(response)? };
-    let credentials = credentials
+    let credentials = Credentials::from_auth_response(response)?
         .ensure_stored()
         .map_err(|err| CredentialsRefreshError::Store(err.0))?;
     re_log::debug!("credentials refreshed successfully");
@@ -207,21 +204,12 @@ impl Credentials {
     ///
     /// Assumes the credentials are valid and not expired.
     ///
-    /// # Safety
-    ///
-    /// Misusing this will not cause UB, but we're still marking it unsafe
-    /// to ensure it is not used lightly.
-    ///
     /// The authentication response must come from a trusted source, such
     /// as the authentication API.
-    ///
-    #[expect(unsafe_code)]
-    pub unsafe fn from_auth_response(
+    pub fn from_auth_response(
         res: api::RefreshResponse,
     ) -> Result<InMemoryCredentials, MalformedTokenError> {
-        // SAFETY: The token comes from a trusted source, which is the authentication API.
-        #[expect(unsafe_code)]
-        let access_token = unsafe { AccessToken::unverified(Jwt(res.access_token))? };
+        let access_token = AccessToken::unverified(Jwt(res.access_token))?;
         Ok(InMemoryCredentials(Self {
             user: res.user,
             refresh_token: RefreshToken(res.refresh_token),
@@ -277,13 +265,8 @@ impl AccessToken {
 
     /// Construct an [`AccessToken`] without verifying it.
     ///
-    /// ## Safety
-    ///
-    /// - The token should come from a trusted source, like the Rerun auth API.
-    // Note: Misusing this will not cause UB, but we're still marking it unsafe
-    // to ensure it is not used lightly.
-    #[expect(unsafe_code)]
-    pub(crate) unsafe fn unverified(jwt: Jwt) -> Result<Self, MalformedTokenError> {
+    /// The token should come from a trusted source, like the Rerun auth API.
+    pub(crate) fn unverified(jwt: Jwt) -> Result<Self, MalformedTokenError> {
         use base64::prelude::*;
 
         let (_header, rest) = jwt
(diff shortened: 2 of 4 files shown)


crates/utils/re_auth/src/oauth/api.rs
```diff
@@ -1,15 +1,24 @@
...
 
 static API_BASE_URL: LazyLock<String> = LazyLock::new(|| {
     std::env::var("RERUN_AUTH_API_BASE_URL")
         .ok()
         .unwrap_or_else(|| "https://rerun.io/api".into())
 });
 
+fn endpoint(endpoint: impl std::fmt::Display) -> String {
+    format!("{base_url}{endpoint}", base_url = *API_BASE_URL)
+}
```
aedm: Using `https://rerun.io/api` for native auth is concerning to me. I'm uncertain whether using a middleman for auth is safe as it's not part of the standard PKCE data flow.

The rerun.io server is not a proxy, it unpacks the payload thereby increasing the attack surface.

 Let's discuss this first.

jprochazk: I think I've addressed this now. We call WorkOS directly for everything, and use PKCE.

crates/viewer/re_redap_browser/src/server_modal/login_flow.rs
```diff
@@ -0,0 +1,181 @@
...
+        #[cfg(not(target_arch = "wasm32"))]
+        {
+            // On native, always show the buttons
+            self.state.ui(ui);
+            self.done(ui, cmd)
+        }
+    }
+
+    fn done(&mut self, ui: &egui::Ui, cmd: &CommandSender) -> Option<LoginFlowResult> {
+        match self.state.done() {
```
aedm, line 69: This will run a web server on the UI thread inside a frame. I understand it's just the login window, still it's rather undeterministic and could cause unexpected delays. I think it's a good idea to make the http server loop run independently of the UI and put the auth response into a channel instead.

crates/utils/re_auth/src/oauth/api.rs
```diff
@@ -19,88 +24,275 @@ pub enum Error {
...
+    code_challenge: String,
+}
+
+const CHARSET: &[u8] = b"\
+    ABCDEFGHIJKLMNOPQRSTUVWXYZ\
+    abcdefghijklmnopqrstuvwxyz\
+    0123456789-.~_\
+";
+
+impl Pkce {
```
aedm, line 157: Nice that you implemented it yourself instead of adding an entire crate :)

crates/utils/re_auth/src/oauth.rs
```diff
@@ -207,8 +206,18 @@ impl Credentials {
...
+    ///
+    /// # Safety
+    ///
+    /// Misusing this will not cause UB, but we're still marking it unsafe
+    /// to ensure it is not used lightly.
+    ///
+    /// The authentication response must come from a trusted source, such
+    /// as the authentication API.
+    ///
+    #[expect(unsafe_code)]
```
aedm: If the auth response doesn't come from a trusted source, it just won't work. I'm okay with removing the `unsafe`, it's a safe function in terms of Rust safety.

crates/utils/re_auth/src/oauth.rs
```diff
@@ -77,7 +74,9 @@ pub async fn refresh_credentials(
     );
 
     let response = api::refresh(&credentials.refresh_token).await?;
-    let credentials = Credentials::from_auth_response(response)?;
+    #[expect(unsafe_code)]
+    // SAFETY: credentials come from a trusted source
+    let credentials = unsafe { Credentials::from_auth_response(response)? };
```
aedm: See below.

crates/utils/re_auth/src/cli.rs
```diff
@@ -149,19 +100,33 @@ pub async fn login(options: LoginOptions<'_>) -> Result<(), Error> {
...
-    // 4. Deserialize credentials
-    let credentials = Credentials::from_auth_response(auth.into())?;
+    // 4. Exchange code for credentials
+    let auth = send_async(AuthenticateWithCode::new(&code, &pkce))
+        .await
+        .map_err(|err| Error::Generic(err.into()))?;
+
+    #[expect(unsafe_code)]
+    // SAFETY: Credentials come from the auth API
+    let credentials = unsafe { Credentials::from_auth_response(auth.into())? };
```
aedm: See below.

crates/utils/re_auth/src/cli.rs
```diff
@@ -121,23 +83,12 @@ pub async fn login(options: LoginOptions<'_>) -> Result<(), Error> {
...
-        addr = server.server_addr(),
-        n = BASE64_URL_SAFE.encode(nonce.to_le_bytes()),
-    );
-    // This is the URL the user should open to log in:
-    let login_url = format!(
-        "{login_page_url}?r={r}",
-        login_page_url = options.login_page_url.unwrap_or(DEFAULT_LOGIN_URL),
-        r = BASE64_URL_SAFE.encode(logged_in_url.as_bytes()),
-    );
+    let login_url = server.get_login_url();
```
aedm, line 91: I like that you removed the nonce <3

crates/utils/re_auth/src/oauth/api.rs
```diff
@@ -2,12 +2,17 @@
...
+use sha2::{Digest as _, Sha256};
 
-static API_BASE_URL: LazyLock<String> = LazyLock::new(|| {
-    std::env::var("RERUN_AUTH_API_BASE_URL")
+use crate::oauth::OAUTH_CLIENT_ID;
+
+use super::RefreshToken;
+
+static WORKOS_API: LazyLock<String> = LazyLock::new(|| {
+    std::env::var("WORKOS_API_BASE_URL")
```
aedm: Since it's standard OAuth2 at this point, maybe the env var could be called something not vendor specific. I like the `RERUN_OAUTH_CLIENT_ID` name you picked, so maybe `RERUN_OAUTH_SERVER_URL`?

aedm (approved):
Great job! Thanks for putting up with my nitpicks. :)

I approve but please take a look at my comments and decide which requests are worth addressing.

crates/viewer/re_redap_browser/src/server_modal/login_flow.rs
jprochazk: Sure. I don't think it's too much of a resource hog, since `tiny_http::Server` runs its accept loop on a separate thread already, but it's also not that big of a change to do this, and it makes sense.

crates/utils/re_auth/src/oauth/api.rs
jprochazk: I found a crate that implements this in roughly as many lines in much the same way, which gave me confidence to just implement it myself :sweat_smile:

crates/utils/re_auth/src/oauth.rs
jprochazk: Fair enough. Will change this

crates/viewer/re_redap_browser/src/server_modal/login_flow.rs
jprochazk: Hmm... After trying it, I changed my mind. I had to rewrite almost the complete file, and I'd rather not make large changes like that at this stage.

I also realised that the `check_for_browser_response` function very quickly exits if there are no requests to it, as the actual server runs in a separate thread already.

Issue #7200, opened 2024-08-15
Individual Timeline Points get lost when using send_columns


Famok:
**Describe the bug**
When sending (scalar) values via the send_columns the timeline might stop showing individual points and only a line.
Logging the same values using set_time_seconds and log Scalar will show these.

**To Reproduce**
```
import rerun as rr
import numpy as np

rr.init('test',spawn=True)

times = np.linspace(0,1,10000)
y = np.random.rand(len(times))

rr.send_columns(
                                f'myval',
                                times=[rr.TimeSecondsColumn("sensor_time", times)],
                                components=[rr.components.ScalarBatch(y)],
                            )

# And now the same using normal logcalls
for i,t in enumerate(times):
    rr.set_time_seconds('sensor_time',t)
    rr.log('anotherval', rr.Scalar(y[i]))
```

**Expected behavior**
I expect the timeline to look the same as with normal log calls

**Screenshots**
Would love to, but seems like I can't upload screenshots

**Desktop (please complete the following information):**
Windows 11 23H2

**Rerun version**
rerun_py 0.18.0-alpha.1+dev [rustc 1.76.0 (07dca489a 2024-02-04), LLVM 17.0.6] x86_64-pc-windows-msvc main 9e50d2c, built 2024-08-14T13:15:39Z

emilk:
<img width="621" alt="Screenshot 2024-08-15 at 14 01 14" src="https://github.com/user-attachments/assets/d3dbecfb-db1a-48b8-ab99-435814a20b21">

<img width="445" alt="Screenshot 2024-08-15 at 14 01 28" src="https://github.com/user-attachments/assets/e0c76de6-b12f-4ac7-8bee-5a3c173ac2c3">

jprochazk:
We recently changed how the timeline retrieves the data used to render those density graphs to be done on demand in the UI, which resulted in a large improvement in ingestion times.
But because we're now doing more a bit more work on each frame, we no longer render individual time points at all times, instead we only render individual points if the total number of events within the chunk is under a threshold. With 10000 individual events in a single chunk, you're exactly at that threshold:

https://github.com/rerun-io/rerun/blob/main/crates/viewer/re_time_panel/src/data_density_graph.rs#L478

We could perform a range query of the visible time range for each sorted chunk, and then test _that_ against the threshold, instead of the full chunk as we're currently doing. With that, zooming in on a large sorted chunk would still display individual events.

PR #11766, opened 2025-11-03
Improve data density visualization by sampling dense chunks

### LINKED ISSUE #10102: Sample rows from a chunk uniformly instead of giving up and filling the chunk
### What

The timeline panel's data density histogram currently abandons rendering when chunks contain too much data, falling back to no visualization to maintain performance (<1.5ms frame cost, [relevant code](https://github.com/rerun-io/rerun/blob/0645e11c46ccdf394ff6bb12f8434d3575dedcc3/crates/viewer/re_time_panel/src/data_density_graph.rs#L572)). This creates confusing gaps in the histogram that don't reflect actual data density, especially when recordings have periods without any logging.

Instead of giving up, I propose we uniformly sample `x` rows from the chunk and use that to create a better estimate.

### Example

In this example I stopped logging data for about 10 seconds between `12:04:40` and `12:04:50`:

https://github.com/user-attachments/assets/2b6dd448-b574-4eb7-9ecc-3dac2549443b

oxkitsune:
### Related

* Closes #10102 

### What

Rendering the data density graph now uniformly samples rows from chunks with many rows, instead of skipping them.
This does not solve the problem and there's a better, albeit more involved solution discussed in #7200.

Benchmarks (M3 Pro):

For chunks with 5k (around the threshold) I get:

```
// 5k
sampling/5000/sample_0        time:   [2.1716 µs 2.2432 µs 2.3375 µs]
sampling/5000/sample_4000 time:   [57.836 µs 57.937 µs 58.033 µs]
sampling/5000/sample_8000 time:   [62.420 µs 68.867 µs 76.966 µs]
```

So around the threshold, we see that for 5k rows with `max_sampled_events_per_chunk=4000`, we sample 4k events, which is actually faster than sampling the full chunk at `max_sampled_events_per_chunk=8000`.

For chunks with 20k-100k rows I get the following:

```
// 20k
sampling/20000/sample_0 time:           [2.1293 µs 2.1467 µs 2.1631 µs]
sampling/20000/sample_4000 time:    [130.47 µs 131.00 µs 131.39 µs]
sampling/20000/sample_8000 time:    [159.53 µs 159.68 µs 159.78 µs]

// 50k
sampling/50000/sample_0        time:    [2.1483 µs 2.2347 µs 2.3770 µs]
sampling/50000/sample_4000 time:    [278.60 µs 279.56 µs 281.00 µs]
sampling/50000/sample_8000 time:    [307.62 µs 317.30 µs 345.24 µs]

// 100k
sampling/100000/sample_0        time:   [2.1437 µs 2.2881 µs 2.6224 µs]
sampling/100000/sample_4000 time:   [521.40 µs 522.86 µs 525.53 µs]
sampling/100000/sample_8000 time:   [552.32 µs 553.91 µs 555.89 µs]
```

here `sample_0` is the original behavior. We'd be able to go with `8000` samples without much of a performance hit.

### DIFF
--- a/crates/viewer/re_time_panel/src/data_density_graph.rs
+++ b/crates/viewer/re_time_panel/src/data_density_graph.rs
@@ -410,28 +410,6 @@ fn smooth(density: &[f32]) -> Vec<f32> {
         .collect()
 }
 
-/// Uniformly sample events using the given sample size.
-///
-/// Each sampled event's count is reweighted to preserve the total density.
-pub fn uniform_sample_events(events: &[(TimeInt, u64)], sample_size: usize) -> Vec<(TimeInt, f32)> {
-    re_tracing::profile_function!();
-
-    let step = events.len() as f32 / sample_size as f32;
-    let mut sampled = Vec::with_capacity(sample_size);
-
-    for i in 0..sample_size {
-        let idx = (i as f32 * step) as usize;
-        // This means we might miss the last event if rounding down, but that's acceptable.
-        if let Some(&(time, count)) = events.get(idx) {
-            // Reweight the count to preserve total density
-            let weighted_count = count as f32 * step;
-            sampled.push((time, weighted_count));
-        }
-    }
-
-    sampled
-}
-
 // ----------------------------------------------------------------------------
 
 #[expect(clippy::too_many_arguments)]
@@ -611,16 +589,17 @@ pub fn build_density_graph<'a>(
             } else if config.max_sampled_events_per_chunk > 0 {
                 let events = chunk.num_events_cumulative_per_unique_time(timeline);
 
-                let sampled_events = if events.len() > config.max_sampled_events_per_chunk {
+                if events.len() > config.max_sampled_events_per_chunk {
                     // If there's more rows than the configured max, we sample events to get a fast, good enough density estimate.
-                    uniform_sample_events(&events, config.max_sampled_events_per_chunk)
+                    data.add_uniform_sample_from_chunk(
+                        &events,
+                        config.max_sampled_events_per_chunk,
+                    );
                 } else {
                     // No need to sample, we can use all events.
-                    events.into_iter().map(|(t, n)| (t, n as f32)).collect()
-                };
-
-                for (time, num_events) in sampled_events {
-                    data.add_chunk_point(time, num_events);
+                    for (time, num_events) in events {
+                        data.add_chunk_point(time, num_events as f32);
+                    }
                 }
             } else {
                 // Fall back to uniform distribution across the entire time range
@@ -755,6 +734,26 @@ impl<'a> DensityGraphBuilder<'a> {
             density_graph: DensityGraph::new(row_rect.x_range()),
             hovered_time: None,
             hovered_pos: None,
+        }
+    }
+
+    /// Uniformly sample events using the given sample size.
+    ///
+    /// Each sampled event's count is reweighted to preserve the total density.
+    fn add_uniform_sample_from_chunk(&mut self, events: &[(TimeInt, u64)], sample_size: usize) {
+        re_tracing::profile_function!();
+
+        let step = events.len() as f32 / sample_size as f32;
+
+        for i in 0..sample_size {
+            let idx = (i as f32 * step) as usize;
+            // This means we might miss the last event if rounding down, but that's acceptable.
+            if let Some(&(time, count)) = events.get(idx) {
+                // Reweight the count to preserve total density
+                let weighted_count = count as f32 * step;
+
+                self.add_chunk_point(time, weighted_count);
+            }
         }
     }
 


crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -649,6 +693,10 @@ impl Default for DensityGraphBuilderConfig {
 
             // Processing unsorted events is about 20% slower than sorted events.
             max_events_in_unsorted_chunk: 8_000,
+
+            // When chunks are too large to render all events, sample this many events uniformly
+            // to create a good enough density estimate.
+            max_sampled_events_per_chunk: 4_000,
```
emilk: What's the motivation behind this particular number?

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -582,10 +603,26 @@ pub fn build_density_graph<'a>(
...
+                    // If there's more rows than the configured max, we sample events to get a fast, good enough density estimate.
+                    uniform_sample_events(&events, config.max_sampled_events_per_chunk)
+                } else {
+                    // No need to sample, we can use all events.
+                    events
+                };
+
+                for (time, num_events) in sampled_events {
+                    data.add_chunk_point(time, num_events as usize);
+                }
```
emilk: The weight we add to each point should be multiplied by `events.len() / config.max_sampled_events_per_chunk`. That is: if we only sample 5% of all points, each point should count 20x stronger!

```suggestion
                let multiplier = events.len() as f32 / config.max_sampled_events_per_chunk as f32;
                for (time, num_events) in sampled_events {
                    data.add_chunk_point(time, multiplier * num_events as f32);
                }
```

oxkitsune: Good catch!
I dont think that your suggestion is in the correct place though? This code was already there and we render individual events on that code path

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -410,6 +410,31 @@ fn smooth(density: &[f32]) -> Vec<f32> {
...
+    let step = events.len() as f64 / sample_size as f64;
+    let mut sampled = Vec::with_capacity(sample_size);
+
+    for i in 0..sample_size {
+        let idx = (i as f64 * step) as usize;
+
+        // This means we might miss the last event if rounding down, but that's acceptable.
+        if let Some(&(time, count)) = events.get(idx) {
+            // Reweight the count to preserve total density
+            let weighted_count = (count as f64 * step).round() as u64;
```
emilk: No need to round here - just change `fn add_chunk_point` from taking an `num_events: usize` to a `weight: f32`

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -410,6 +410,31 @@ fn smooth(density: &[f32]) -> Vec<f32> {
...
 
+/// Uniformly sample events using the given sample size.
+///
+/// Each sampled event's count is reweighted to preserve the total density.
+fn uniform_sample_events(events: &[(TimeInt, u64)], sample_size: usize) -> Vec<(TimeInt, u64)> {
+    re_tracing::profile_function!();
+
+    // Use double precision to prevent rounding errors when handling many events.
+    // This ensures that we sample evenly across the entire range.
+    let step = events.len() as f64 / sample_size as f64;
```
emilk: f32 has 24 bits of precision, so as long as sample_size is < 2^24 we will get even enough sampling with f32 🤷

crates/viewer/re_time_panel/src/data_density_graph.rs
oxkitsune: The limit is configured at 8k for unsorted chunks, so I figured lets take half. However, after doing some benchmarking, it seems we can use 8k without much of a loss.

For chunks with 5k (around the threshold) I get:

```
// 5k
sampling/5000/sample_0        time:   [2.1716 µs 2.2432 µs 2.3375 µs]
sampling/5000/sample_4000 time:   [57.836 µs 57.937 µs 58.033 µs]
sampling/5000/sample_8000 time:   [62.420 µs 68.867 µs 76.966 µs]
```

So around the threshold, we see that for 5k rows with `max_sampled_events_per_chunk=4000`, we sample 4k events, which is actually faster than sampling the full chunk at `max_sampled_events_per_chunk=8000`.

For chunks with 20k-100k rows I get the following:

```
// 20k
sampling/20000/sample_0 time:           [2.1293 µs 2.1467 µs 2.1631 µs]
sampling/20000/sample_4000 time:    [130.47 µs 131.00 µs 131.39 µs]
sampling/20000/sample_8000 time:    [159.53 µs 159.68 µs 159.78 µs]

// 50k
sampling/50000/sample_0        time:    [2.1483 µs 2.2347 µs 2.3770 µs]
sampling/50000/sample_4000 time:    [278.60 µs 279.56 µs 281.00 µs]
sampling/50000/sample_8000 time:    [307.62 µs 317.30 µs 345.24 µs]

// 100k
sampling/100000/sample_0        time:   [2.1437 µs 2.2881 µs 2.6224 µs]
sampling/100000/sample_4000 time:   [521.40 µs 522.86 µs 525.53 µs]
sampling/100000/sample_8000 time:   [552.32 µs 553.91 µs 555.89 µs]
```

here `sample_0` is the original behavior. We'd be able to go with `8000` samples without much of a performance hit.

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -582,10 +604,26 @@ pub fn build_density_graph<'a>(
...
+                    // If there's more rows than the configured max, we sample events to get a fast, good enough density estimate.
+                    uniform_sample_events(&events, config.max_sampled_events_per_chunk)
+                } else {
+                    // No need to sample, we can use all events.
+                    events.into_iter().map(|(t, n)| (t, n as f32)).collect()
+                };
+
+                for (time, num_events) in sampled_events {
+                    data.add_chunk_point(time, num_events);
                 }
```
emilk: no need for that extra `collect` in this performance sensitive code

```suggestion
                    let sampled_events = uniform_sample_events(&events, config.max_sampled_events_per_chunk);
	                for (time, weight) in sampled_events {
	                    data.add_chunk_point(time, weight);
	                }
                } else {
                    // No need to sample, we can use all events.
	                for (time, num_events) in events {
	                    data.add_chunk_point(time, num_events as _);
	                }
                };
}
```

emilk: In fact, event `uniform_sample_events` doesn't need to collect. It can just call `add_chunk_point` directly

PR #11770, opened 2025-11-04
Rename source_frame/target_frame to child_frame/parent_frame

Wumpf:
### Related

* follow-up to https://github.com/rerun-io/rerun/pull/11730

### What

Pure rename + some drive-by comment improvements

After some internal discussion we concluded that we should continue to talk about child/parent relationships on individual transform relationships as this is more common.

However, when talking about how to get from one frame into the reference of another we keep talking about sources being expressed in a target. Because of that the edits to `TransformForest` and `TransformTreeContext` are much more limited!
(note however that there are some inconsistencies remaining in `TransformForest`. Didn't change those since I have more changes in that area staged on another branch and I don't want to conflict further)

* [x] pass full-ci (this caused issues on the last round here due to comparison not fully running)

### DIFF
--- a/crates/viewer/re_viewer/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.archetypes.Transform3D.snap
+++ b/crates/viewer/re_viewer/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.archetypes.Transform3D.snap
@@ -1,5 +1,5 @@
 ---
-source: crates/viewer/re_component_fallbacks/tests/all_component_fallbacks.rs
+source: crates/viewer/re_viewer/tests/all_component_fallbacks.rs
 expression: arch_display
 ---
 translation: [[0.0, 0.0, 0.0]]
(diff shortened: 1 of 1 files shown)


pixi.toml
```diff
@@ -680,7 +680,7 @@ numpy = ">=2" # Rerun still needs numpy <2. Enforce this outside of the pypi dep
 # This is very similar to `pixi run py-build`, and dispatches to maturin by way of PEP621.
 # However, pixi doesn't know how to track the rust dependencies of the python package, so
 # you still need to `pixi run py-build` in the correct environment if you change the rust code.
-rerun-sdk = { path = "rerun_py", editable = true }
+#rerun-sdk = { path = "rerun_py", editable = true }
```
Wumpf: oops, shouldn't be here

Wumpf:
@rerun-bot full-check

grtlr (approved):
Nice! Really makes it a lot cleaner in many places

PR #11774, opened 2025-11-04
Allow editing times in the time panel

emilk:
* Closes https://linear.app/rerun/issue/RR-2865/cant-edit-the-time-in-the-time-panel-input-box

When showing a time from _today_, we omit the date. But that meant we couldn't parse the results when the user edited the time. Now we can!


IsseW (approved):
Thanks!

The parser we use is not really great for human inputted stuff though

Image

Can't handle extra spaces in some places

emilk:
There are more improvements we can make yes, but at least we can now change a digit and it still parses (in this case)

PR #11775, opened 2025-11-04
Add extra check to fast lint and speed up mypy

ntjohnson1:
### What
* In my last pr I got an error about large files but the check can run locally in under a second
* We use a few different mypy configurations which is invalidating the cache. By using separate cache directories on my machine mypy goes from 20s to 2s (after the first run).


abey79 (approved):
lgtm. Curious, where do these .mypy_cache_* land? Should they be in the gitignore?

ntjohnson1:
> lgtm. Curious, where do these .mypy_cache_* land? Should they be in the gitignore?

They land in project root similar to the default `.mypy_cache` fun fact mypy's cache tool automatically ignores itself!

```console
$ tail .mypy_cache_rerun_sdk/.gitignore 
# Automatically created by mypy
*
```

PR #11776, opened 2025-11-04
Fix orbit eye camera no longer zooming in on double clicked entities

Wumpf:
### Related

* Closes  RR-2796

### What

At first I thought there was more wrong, but it was really just the _zoom_ not working. Sidy-sidy movement was still happening on double click


IsseW (approved):
Looks good to me!

PR #11779, opened 2025-11-04
Add the ability to lookup transform frame ids from transform id hashes

Wumpf:
Mostly needed for debugging, but also in order to identify when a frame is entity-path derived.
(split out of other ongoing work)

Furthermore, makes the cache-parts of the `TransformResolutionCache` have interior mutability. This is done so one can hold on to a `TransformsForChildFrame` while calling `TransformResolutionCache::lookup_frame_id`. Makes things a lot more convenient and as a pattern makes sense - we're not "editing" `TransformResolutionCache` through our lookups, we're merely happen to populate the cache if needed and an `RwLock` would be overkill from a practical standpoint (parking lot Mutex is super lightweight, RwLock less so). Right now this is very explicit in the API, making clear when locking occurs and I think I like it that way, but it's not unlikely we'll iterate on this in the future!

### DIFF
--- a/crates/store/re_tf/benches/transform_resolution_cache_bench.rs
+++ b/crates/store/re_tf/benches/transform_resolution_cache_bench.rs
@@ -80,7 +80,7 @@ fn transform_resolution_cache_query(c: &mut Criterion) {
                 cache
             },
             |cold_cache| {
-                let mut frame_transforms = cold_cache
+                let frame_transforms = cold_cache
                     .transforms_for_timeline(query.timeline())
                     .frame_transforms(queried_frame)
                     .unwrap();
@@ -102,7 +102,7 @@ fn transform_resolution_cache_query(c: &mut Criterion) {
 
     c.bench_function("query_cached_frame", |b| {
         b.iter(|| {
-            let mut frame_transforms = warm_cache
+            let frame_transforms = warm_cache
                 .transforms_for_timeline(query.timeline())
                 .frame_transforms(queried_frame)
                 .unwrap();


crates/store/re_tf/src/frame_id_registry.rs
```diff
@@ -0,0 +1,153 @@
...
+            // so compared to `TransformFrameIdHash,` is it _relatively_ expensive to compute.
+            match self
+                .frame_id_lookup_table
+                .entry(TransformFrameIdHash::from_entity_path(entity_path))
+            {
+                Entry::Occupied(_) => {
+                    break;
+                }
+                Entry::Vacant(e) => e.insert(TransformFrameId::from_entity_path(entity_path)),
+            };
```
grtlr, line 47: TIL: There is a nightly `try_insert`: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.try_insert

crates/store/re_tf/src/frame_id_registry.rs
```diff
@@ -0,0 +1,153 @@
...
+    ) -> Option<&TransformFrameId> {
+        self.frame_id_lookup_table.get(&frame_id_hash)
+    }
+
+    /// Registers all frame ids mentioned in a chunk, including frames implied by the chunk's entity and its parents.
+    ///
+    /// Implementation note:
+    /// Having the registration of frame ids separate from other frame id related bookkeeping makes things more modular
+    /// at the price of additional overhead. However, we generally assume that retrieving `TransformFrameId`/`TransformFrameIdHash` from a string is fast.
+    pub fn register_all_ids_in_chunk(&mut self, chunk: &re_chunk_store::Chunk) {
```
grtlr: nit: we have so many IDs already 🫠
```suggestion
    pub fn register_all_frames_in_chunk(&mut self, chunk: &re_chunk_store::Chunk) {
```

crates/store/re_tf/src/frame_id_registry.rs
Wumpf: want!!!

PR #11784, opened 2025-11-04
Update relative links

bllchmbrs:
Links were broken and it's not 100% clear why it wasn't caught before.


ntjohnson1 (approved):
If CI can't check it I'll trust you were able to check things manually

Issue #8265, opened 2024-12-02
Allow setting default camera position/direction in blueprint


Wumpf:
This is a small subtask of:
* https://github.com/rerun-io/rerun/issues/6825

3D spatial views decide their default camera pose heuristically based on a smoothed bounding box. This default pose is what is used on start & when resetting the camera.
The first step into configurable cameras is therefore to set the default position & direction for this default camera.

This is mostly orthogonal to all the other camera settings: it doesn't have any relation to how the camera is controlled (ego/orbit/other, follow-mode, movement/rotation sensitivity).
The accompanying ui should have two buttons for:
* setting the current camera dir/pos as the default
* resetting the camera to the default (same thing as double clicking in the view)

This was started in 
* https://github.com/rerun-io/rerun/pull/8211

Proposing to use keep the simple look-at camera style here that was used in that draft:
* position
* target position
* optional up axis (sourced from View Coordinates, this part isn't in the linked draft yet)

PR #11788, opened 2025-11-05
3D eye position and other properties in blueprint

IsseW:
### Related

- Closes https://github.com/rerun-io/rerun/issues/6825.
- Closes https://github.com/rerun-io/rerun/issues/8265.

### What

Stores more eye properties in blueprint, like eye position and eye look target. Making it controllable from python.

New selection UI for the 3d spatial view:
image

I've also updated tests which before this either directly wrote to the spatial view state, or artificially did inputs to zoom out the view. I did not bother to recreate the viewing angle of these exactly as I didn't think that was important to the tests.

### DIFF
--- a/rerun_py/rerun_sdk/rerun/blueprint/views/spatial3d_view.py
+++ b/rerun_py/rerun_sdk/rerun/blueprint/views/spatial3d_view.py
@@ -75,7 +75,6 @@ class Spatial3DView(View):
             spatial_information=rrb.SpatialInformation(
                 show_axes=True,
                 show_bounding_box=True,
-                show_smoothed_bounding_box=True,
             ),
         ),
         collapse_panels=True,


Wumpf:
co-review on build
* disable interpolation on change of position alone (i.e. rotating around in orbit mode)
* allow rotating out of camera tracking
* wasd doesn't work
* should add a test for some interactions having just the axis in the middle and see that certain button clicks do stuff

crates/store/re_types/definitions/rerun/blueprint/archetypes/spatial_information.fbs
```diff
@@ -0,0 +1,15 @@
...
+table SpatialInformation (
+    "attr.rerun.scope": "blueprint"
+) {
+  /// Whether axes should be shown at the origin.
+  show_axes: rerun.blueprint.components.Enabled ("attr.rerun.component_optional", nullable, order: 100);
+
+  /// Whether the bounding box should be shown.
+  show_bounding_box: rerun.blueprint.components.Enabled ("attr.rerun.component_optional", nullable, order: 200);
+
+  /// Whether the smoothed bounding box should be shown.
```
Wumpf: point out that we use this for heuristics internally

crates/store/re_types/definitions/rerun/blueprint/archetypes/eye_controls3d.fbs
```diff
@@ -6,6 +6,25 @@ namespace rerun.blueprint.archetypes;
 table EyeControls3D (
     "attr.rerun.scope": "blueprint"
 ) {
+    /// The cameras current position.
```
Wumpf, line 15: suggested order because it affects ui order as well:
* kind 
* position
* look_target
* eye_up
* speed
* tracking_entity
* spin_speed


... mostly I just want spin_speed at the bottom but I think the base config mode at the top might be nice :)

crates/store/re_types/definitions/rerun/blueprint/archetypes/eye_controls3d.fbs
```diff
@@ -6,6 +6,25 @@ namespace rerun.blueprint.archetypes;
...
+    position: rerun.components.Position3D ("attr.rerun.component_optional", nullable, order: 100);
+
+    /// The position the camera is currently looking at.
+    ///
+    /// If this is an orbital camera, this also is the center it orbits around.
+    look_target: rerun.components.Position3D ("attr.rerun.component_optional", nullable, order: 200);
+
+    /// What speed, if any, the camera should spin around the eye-up axis when in orbit mode.
```
Wumpf: should say that it defaults to 0, meaning no automatic spinning

crates/store/re_types/definitions/rerun/blueprint/archetypes/eye_controls3d.fbs
```diff
@@ -6,6 +6,25 @@ namespace rerun.blueprint.archetypes;
...
+    spin_speed: rerun.blueprint.components.AngularSpeed ("attr.rerun.component_optional", nullable, order: 300);
+
+    /// The up-axis of the eye itself, in world-space.
+    ///
+    /// Initially, the up-axis of the eye will be the same as the up-axis of the scene (or +Z if
+    /// the scene has no up axis defined).
+    ///
+    /// A zero vector is valid and will result in 3 degrees of freedom.
```
Wumpf: That's only the case for ego perspective though, no?

IsseW: Hmm it behaves weirdly in both instances I think, so can maybe just remove that comment?

crates/store/re_types/definitions/rerun/blueprint/archetypes/eye_controls3d.fbs
IsseW: Makes sense, updated!

crates/store/re_types/definitions/rerun/blueprint/archetypes/eye_controls3d.fbs
Wumpf: sure, don't recall having that feature 🤷

docs/snippets/all/views/spatial3d.py
```diff
@@ -24,6 +24,10 @@
         background=[100, 149, 237],
         # Configure the eye controls.
         eye_controls=rrb.EyeControls3D(
+            position=(0.0, 0.0, 2.0),
+            look_target=(0.0, 2.0, 0.0),
+            eye_up=(-1.0, 0.0, 0.0),
+            spin_speed=0.2,
             kind=rrb.Eye3DKind.FirstPerson,
             speed=20.0,
```
Wumpf, line 32: have you tried whether this looks any good in that scene? We build & show that example in the docs after all :)

crates/viewer/re_viewer/tests/snapshots/all_view_selection_uis/Dark/3D.png
Wumpf, line 1: ci hopefully complains about the changed field order here

crates/viewer/re_view_spatial/tests/transparent_geometry.rs
```diff
@@ -34,65 +39,62 @@ fn test_transparent_geometry<A: AsComponents>(
         });
     }
 
-    let view_id = test_context.setup_viewport_blueprint(|_ctx, blueprint| {
+    let pos = |d: f32| {
```
Wumpf: I admit being guilty of making this position logic arcane in the first place, but let's not make it worse, starting to read the diff I had no idea what this pos thing does ;)

```suggestion
    let get_eye_position = |orientation_y: f32| {
```

crates/viewer/re_view_spatial/tests/transparent_geometry.rs
```diff
@@ -34,65 +39,62 @@ fn test_transparent_geometry<A: AsComponents>(
...
+                ctx.blueprint_query(),
+                view_id,
+            );
+
+            let v = pos(d);
+
+            eye_property.save_blueprint_component(
+                &ctx,
+                &EyeControls3D::descriptor_position(),
+                &Position3D::new(v.x, v.y, v.z),
```
Wumpf: nit:
```suggestion
                &Position3D::from(position),
```

crates/viewer/re_view_spatial/tests/transform_clamping.rs
```diff
@@ -234,14 +235,17 @@ fn run_view_ui_and_save_snapshot(
...
+                ViewProperty::from_archetype::<EyeControls3D>(
+                    ctx.current_blueprint(),
+                    ctx.blueprint_query(),
+                    view_id,
+                )
+                .save_blueprint_component(
+                    &ctx,
+                    &EyeControls3D::descriptor_position(),
+                    &Position3D::new(15.0, 5.0, 15.0),
+                );
```
Wumpf, line 248: given how often this shows up it makes me wonder whether we should have a utility specificially in `re_view_spatial/tests` to set the camera :)

we can put that in a follow-up tho / whoever gets annoyed enough by this

crates/viewer/re_view_spatial/tests/snapshots/transparent_mesh_1.png
Wumpf, line 1: oh no, you swapped the transparency test pictures 😆

crates/viewer/re_view_spatial/tests/transform_hierarchy.rs
```diff
@@ -186,16 +187,18 @@ fn run_view_ui_and_save_snapshot(
...
+            ViewProperty::from_archetype::<EyeControls3D>(
+                ctx.current_blueprint(),
+                ctx.blueprint_query(),
+                view_id,
+            )
+            .save_blueprint_component(
+                &ctx,
+                &EyeControls3D::descriptor_position(),
+                &Position3D::new(0.0, 8.0, 10.0),
+            );
```
Wumpf, line 200: let's also fix the orientation on this test series- that will make the pictures on this one a lot easier to grasp

crates/viewer/re_view_spatial/tests/snapshots/transform_clamping_spheres.png
Wumpf, line 1: oh my gawd. so much better as a test also now.

crates/viewer/re_view_spatial/tests/snapshots/static_overwrite_color_override.png
Wumpf, line 1: should zoom in a bit more if possible here. I get nervous when the thing being tested (the colored points) is so small on screen.

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -150,235 +160,229 @@ impl Eye {
...
-/// Note: we use "eye" so we don't confuse this with logged camera.
-#[derive(Clone, Copy, Debug, PartialEq)]
-pub struct ViewEye {
-    /// First person or orbital?
-    kind: Eye3DKind,
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct LastTrackedEntity {
+    /// Currently tracked entity.
+    pub entity: EntityPath,
+}
```
Wumpf: not sure why this needs new-typing. Is the idea to later own store additional tracking state in here like what aspects are being tracked?

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -150,235 +160,229 @@ impl Eye {
...
+    spin: Option<f64>,
+
+    pub last_eye: Option<Eye>,
+
+    /// The time this was last interacted with in egui time.
     ///
-    /// A value of `Vec3::ZERO` is valid and will result in 3 degrees of freedom, although we never
-    /// use it at the moment.
+    /// None: Hasn't been interacted with yet.
+    pub last_interaction: Option<f64>,
```
Wumpf: maybe?

```suggestion
    pub last_interaction_time: Option<f64>,
```

the fact that this doesn't have a dimension like `Instant` is a bit annoying. but using egui time here makes sense

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -150,235 +160,229 @@ impl Eye {
...
+    pub last_look_target: Option<Vec3>,
+    pub last_orbit_radius: Option<f32>,
+    pub last_eye_up: Option<Vec3>,
+}
+
+struct RawEye {
```
Wumpf: what role does this fulfill?

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -150,235 +160,229 @@ impl Eye {
...
+            )
+            .unwrap_or_else(|| IsoTransform::from_translation(self.pos))
+            .inverse(),
+            fov_y: Some(self.fov_y.unwrap_or(Eye::DEFAULT_FOV_Y)),
         }
     }
 
-    pub fn kind(&self) -> Eye3DKind {
-        self.kind
+    /// World-direction we are looking at
```
```suggestion
    /// Normalized world-direction we are looking at.
```

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -150,235 +160,229 @@ impl Eye {
...
-    pub fn set_kind(&mut self, new_kind: Eye3DKind) {
-        if self.kind != new_kind {
-            // Keep the same position:
-            match new_kind {
-                Eye3DKind::FirstPerson => self.center = self.position(),
-                Eye3DKind::Orbital => {
-                    self.center = self.position() + self.orbit_radius * self.fwd();
-                }
-            }
+    fn up(&self) -> Vec3 {
```
Wumpf, line 353:
```suggestion
    /// Normalized up vector. Guaranteed to be orthgonal to [`Self::fwd`].
    fn up(&self) -> Vec3 {
```

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -150,235 +160,229 @@ impl Eye {
...
-            Eye3DKind::Orbital => Some(self.orbit_radius),
+        if res.dot(fwd).abs() > 0.9999 {
+            fallback
+        } else {
+            res
         }
     }
 
-    /// Set what we orbit around, and at what distance.
+    /// Only valid if we have an up-vector set.
```
Wumpf: a bit confusing with `up` always returning a valid vector

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -437,138 +462,379 @@ impl ViewEye {
...
+            });
+        }
     }
 
-    /// Rotate based on a certain number of pixel delta.
-    pub fn rotate(&mut self, delta: egui::Vec2) {
-        let sensitivity = 0.004; // radians-per-point. TODO(emilk): take fov_y and canvas size into account
-        let delta = sensitivity * delta;
+    /// Gets and updates the current target eye from/to the blueprint.
+    fn update_target_eye(
```
Wumpf: maybe

```suggestion
    fn control_and_sync_with_blueprint(
```

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -437,138 +462,379 @@ impl ViewEye {
...
+        )?;
+
+        let pos = Vec3::from_array(
+            eye_controls
+                .component_or_fallback::<Position3D>(
+                    ctx,
+                    EyeControls3D::descriptor_position().component,
+                )?
+                .0
+                .0,
```
Wumpf: I think we have those on deref, so you can go `**` instead

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -437,138 +462,379 @@ impl ViewEye {
...
+
+        let has_look_target = eye_controls
+            .component_or_empty::<Position3D>(EyeControls3D::descriptor_look_target().component)?
+            .is_some();
+
+        let has_position = eye_controls
+            .component_or_empty::<Position3D>(EyeControls3D::descriptor_position().component)?
+            .is_some();
+
+        let mut eye = RawEye {
```
Wumpf: seems like this would be a good bifurcation into two methods? First extract a `RawEye` from blueprint & fallbacks and then operate on it? 🤔 
This whole thing is a scary looking  and unwieldy right now. Sure from your pov you probably mostly moved around existing code and reconnected it, but it's kind hard to take it in atm

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -437,138 +462,379 @@ impl ViewEye {
...
-    }
 
-    /// Rotate around forward axis
-    fn roll(&mut self, rect: &egui::Rect, pointer_pos: egui::Pos2, delta: egui::Vec2) {
-        // steering-wheel model
-        let rel = pointer_pos - rect.center();
-        let delta_angle = delta.rot90().dot(rel) / rel.length_sq();
-        let rot_delta = Quat::from_rotation_z(delta_angle);
+        // Update blueprint if we've interacted and the component has changed.
+        if eye.did_interact {
```
Wumpf: would definitely be nice to have the saving part as a separate method

crates/viewer/re_view_spatial/src/picking_ui.rs
```diff
@@ -204,7 +204,13 @@ pub fn picking(
                 ItemContext::ThreeD {
                     space_3d: query.space_origin.clone(),
                     pos: hovered_point,
-                    tracked_entity: state.state_3d.tracked_entity.clone(),
+                    tracked_entity: state
+                        .state_3d
+                        .eye_state
```
Wumpf: quite the tower! worth adding an accessor?

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -852,13 +488,18 @@ fn show_orbit_eye_center(
...
+    let Some(look_target) = state_3d.eye_state.last_look_target else {
         return;
     };
-    let Some(orbit_radius) = view_eye.orbit_radius() else {
+    let Some(orbit_radius) = state_3d.eye_state.last_orbit_radius else {
+        return;
+    };
+
+    // We distinguish the eye up-axis from the other two axes:
+    // Default to RFU
```
Wumpf: it says to default but then early outs, looks like a bug?

crates/viewer/re_view_spatial/src/view_3d.rs
```diff
@@ -123,12 +137,130 @@ impl ViewClass for SpatialView3D {
...
+                    .up()
+                    .unwrap_or(SignedAxis3::POSITIVE_Z);
+
+                let eye_up: glam::Vec3 = scene_up.into();
+
+                let eye_dir = {
+                    // Make sure right is to the right, and up is up:
+                    let right = scene_right.into();
+                    let fwd = eye_up.cross(right);
+                    0.75 * fwd + 0.25 * right - 0.25 * eye_up
```
Wumpf, line 234: don't quite understand what's going on here, can you elaborate?

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -437,138 +462,379 @@ impl ViewEye {
...
+            eye_controls
+                .component_or_fallback::<Position3D>(
+                    ctx,
+                    EyeControls3D::descriptor_look_target().component,
+                )?
+                .0
+                .0,
+        );
+
+        // If we use fallbacks for position and look target, continue to
```
Wumpf: a bit confused about this: it says continue to interpolate but then calls _start_-interpolation. Loosing track here why that works

docs/snippets/all/views/spatial3d.py
IsseW: It looks different, but not bad I think. You're kind of in the middle of the spheres and spinning around slowly

crates/viewer/re_view_spatial/src/eye.rs
IsseW: Ah originally had more fields, but yes this would be better as just EntityPath now

crates/viewer/re_view_spatial/src/eye.rs
IsseW: > Guaranteed to be orthgonal to [`Self::fwd`]

This isn't necessarily true, but it's guaranteed to not be parallel to `Self::fwd`. Adding doc comments though!

crates/viewer/re_view_spatial/src/eye.rs
IsseW: Seems like `Vec3D` doesn't implement deref, but could use Vec3::from instead

crates/viewer/re_view_spatial/src/eye.rs
IsseW: tried to clarify it a bit, basically calling `start_interpolation` refreshes the interpolation state.

crates/viewer/re_view_spatial/src/view_3d.rs
IsseW: Tried to clarify it a bit from how I understand it, tell me if you think it's better. This is more or less copied from the old `default_eye` function.

crates/viewer/re_view_spatial/src/ui_3d.rs
IsseW: I think left over comment, added a comment on why those early returns are okay here.

crates/viewer/re_view_spatial/src/eye.rs
IsseW: Yeah good point! Tried making it a bit more readable

crates/viewer/re_view_spatial/src/eye.rs
IsseW: Added some comments on it, basically just utility to avoid passing a bunch of variables around instead.

crates/viewer/re_view_spatial/tests/snapshots/static_overwrite_color_override.png
IsseW: updated them

IsseW:
> Also plz grep for #8265 to check whether there's more tech debt associated with this if you haven't checked already.

Did grep for that and a few others, and didn't find anything. Also for tests that used raw_input

crates/viewer/re_view_spatial/src/view_3d.rs
Wumpf: it does. still kinda magic, but yeah if that's old behavior that's fine 🙈

Wumpf (changes requested):
`ControlEye` makes so much more sense and the method cuts makes this thing a lot easier to follow & make it feel more maintainable for the next person, thank you!

One nasty thing I noticed only now unfortunately is that the selection panel needs too much horizontal space now, _way_ more than we give it by default

Image

The reason for that is ofc "show smoothed bounding box".  We have to figure out how to make this shorter, not sure yet 🤔 

Btw. without that it's _almost_ alright, but something strange is going on with direction vs position. Seems those boxes are too wide. We can let that one slide though if we find a shorter name for the above

Image

PR #11789, opened 2025-11-05
Add example for loading MCAP with custom messages using Lenses

MichaelGrupp:
Builds on the draft #11635, i.e. you can load MCAP files with custom Protobuf
(here: Foxglove messages), convert and send them to Rerun via Lenses API.

This is an example for an in-progress API, so it's set to ignored in the
`manifest.toml`.

See `README.md` for more information & usage.

https://linear.app/rerun/issue/RR-2851

### DIFF
--- a/crates/utils/re_arrow_combinators/tests/semantic.rs
+++ b/crates/utils/re_arrow_combinators/tests/semantic.rs
@@ -15,7 +15,7 @@ use arrow::{
     datatypes::{DataType, Field, GenericBinaryType},
 };
 
-use re_types::components::VideoCodec;
+use re_types::{components::VideoCodec, reflection::Enum as _};
 
 mod testing_utils;
 
@@ -225,4 +225,30 @@ fn test_string_to_codec_uint32_unsupported() {
         };
         assert_eq!(actual, bad_codec);
     }
+}
+
+/// Tests that all codecs defined in `VideoCodec` are accepted.
+#[test]
+fn test_string_to_codec_uint32_all_supported() {
+    let variants = VideoCodec::variants();
+    let variant_names = variants
+        .iter()
+        .map(|v| format!("{v:?}").to_lowercase())
+        .collect::<Vec<String>>();
+    let input_array = StringArray::from(
+        variant_names
+            .iter()
+            .map(|name| Some(name.as_str()))
+            .collect::<Vec<Option<&str>>>(),
+    );
+    let output_array = StringToVideoCodecUInt32::default()
+        .transform(&input_array)
+        .expect("transformation failed - are all variants of VideoCodec supported?");
+    let expected_array = UInt32Array::from(
+        variants
+            .iter()
+            .map(|v| Some(*v as u32))
+            .collect::<Vec<Option<u32>>>(),
+    );
+    assert_eq!(output_array, expected_array);
 }


examples/rust/japanese_alley/src/main.rs
```diff
@@ -86,20 +72,40 @@ fn main() -> anyhow::Result<()> {
...
             .build();
 
     let lenses_sink = LensesSink::new(GrpcSink::default())
         .with_lens(image_lens)
         .with_lens(instance_pose_lens)
-        .with_lens(instance_poses_lens);
+        .with_lens(instance_poses_lens)
+        .with_lens(video_lens);
 
     let (rec, _serve_guard) = args.rerun.init("rerun_example_japanese_alley")?;
```
MichaelGrupp: I would prefer a better name for this example. It _can_ load the japanese alley MCAP, but it's actually also capable of loading other datasets, now also with the video type that's not in the alley example. What do you think?

Maybe `custom_mcap_lenses`?

MichaelGrupp: Changed it now, let me know if you have a better name or I should revert it.

examples/rust/custom_mcap_lenses/src/conversions.rs
```diff
@@ -0,0 +1,101 @@
...
+/// Converts a list of video codec strings to Rerun VideoCodec values (as u32).
+pub fn list_string_to_list_codec_uint32(list_array: &ListArray) -> Result<ListArray, Error> {
+    let pipeline = MapList::new(StringToCodecUint32::default());
+    Ok(pipeline.transform(list_array)?)
+}
+
+/// Transforms a StringArray of Foxglove video codec names to a UInt32Array,
+/// where each u32 corresponds to a Rerun VideoCodec enum value.
+#[derive(Default)]
+struct StringToCodecUint32 {}
```
grtlr:
```suggestion
struct StringToCodecUInt32 {}
```
nit: sorry this is triggering me 🙈

examples/rust/foxglove_lenses/src/conversions.rs
```diff
@@ -27,6 +28,75 @@ pub fn list_string_to_list_codec_uint32(list_array: &ListArray) -> Result<ListAr
...
+    type Target = Int64Array;
+
+    fn transform(
+        &self,
+        source: &StructArray,
+    ) -> Result<Self::Target, re_arrow_util::transform::Error> {
+        use re_arrow_util::transform::Error;
+
+        let available_fields: Vec<String> =
+            source.fields().iter().map(|f| f.name().clone()).collect();
```
grtlr: Probably better to do this on the unhappy path (i.e. in the `ok_or_else` calls).

examples/rust/foxglove_lenses/src/conversions.rs
```diff
@@ -27,6 +28,75 @@ pub fn list_string_to_list_codec_uint32(list_array: &ListArray) -> Result<ListAr
...
+        for i in 0..source.len() {
+            if source.is_null(i) {
+                output_builder.append_null();
+            } else {
+                let seconds = seconds_array.value(i);
+                let nanos = nanos_array.value(i);
+                let total_nanos = seconds * 1_000_000_000 + nanos as i64;
+                output_builder.append_value(total_nanos);
+            }
+        }
```
grtlr: I'd look to see if we can use `arrow::compute::binary` here, because it will give us better protection from `null` values.

Although both fields should always be present, in your example we can't guarantee a correct result if one of the arrays contains nulls.

examples/rust/foxglove_lenses/src/conversions.rs
MichaelGrupp: Makes sense! I copied that pattern from `transform.rs`, so we should change it also there. Made a small PR to fix that there, and then we can reuse it here without blowing up this PR here more: https://github.com/rerun-io/rerun/pull/11804

MichaelGrupp: Resolved by recent refactoring (will update here soon)

crates/utils/re_arrow_combinators/src/semantic.rs
```diff
@@ -62,3 +65,63 @@ impl<O1: OffsetSizeTrait, O2: OffsetSizeTrait> Transform for BinaryToListUInt8<O
         Ok(list)
     }
 }
+
+/// Converts `StructArray` of timestamps with `seconds` (i64) and `nanos` (i32) fields
+/// to `Int64Array` containing the corresponding total nanoseconds timestamps.
+#[derive(Default)]
+pub struct TimeSpecToNanos {}
```
MichaelGrupp, line 75: I called this `TimeSpecToNanos` now, following the common Linux time type https://man7.org/linux/man-pages/man3/timespec.3type.html.

(I had `SecondsNanosToNanos` before but that looked awful)

examples/rust/foxglove_lenses/src/conversions.rs
MichaelGrupp: I'll check

MichaelGrupp: Nice suggestion, works. See https://github.com/rerun-io/rerun/pull/11789/commits/695b3145e10c4b29e71bf0ddf11f5c00ec2b03bc

crates/utils/re_arrow_combinators/src/semantic.rs
grtlr: It would be great to have a test case for this. I'm particularly interested in seeing how `null` handling works (i.e. either `seconds` or `nanos`) missing.

Since the spec mandates these fields to be there, I think it's fine if the converted entry is null too.

crates/utils/re_arrow_combinators/src/semantic.rs
```diff
@@ -62,3 +67,122 @@ impl<O1: OffsetSizeTrait, O2: OffsetSizeTrait> Transform for BinaryToListUInt8<O
...
+                    field_name: "seconds".to_owned(),
+                    struct_fields: available_fields.clone(),
+                })?;
+        let nanos_array =
+            source
+                .column_by_name("nanos")
+                .ok_or_else(|| Error::MissingStructField {
+                    field_name: "nanos".to_owned(),
+                    struct_fields: available_fields,
+                })?;
```
grtlr: Maybe use our `GetField` transform?

crates/utils/re_arrow_combinators/src/semantic.rs
```diff
@@ -62,3 +67,122 @@ impl<O1: OffsetSizeTrait, O2: OffsetSizeTrait> Transform for BinaryToListUInt8<O
...
+                expected: "Int64Array".to_owned(),
+                actual: seconds_array.data_type().clone(),
+            })?;
+        let nanos_array = nanos_array
+            .as_any()
+            .downcast_ref::<Int32Array>()
+            .ok_or_else(|| Error::UnexpectedListValueType {
+                expected: "Int32Array".to_owned(),
+                actual: nanos_array.data_type().clone(),
+            })?;
```
grtlr: Should we create a `DowncastRef` transform in `cast.rs`? That way we could chain the getting and downcasting.

crates/utils/re_arrow_combinators/src/semantic.rs
```diff
@@ -62,3 +67,122 @@ impl<O1: OffsetSizeTrait, O2: OffsetSizeTrait> Transform for BinaryToListUInt8<O
...
+
+        Ok(
+            arrow::compute::binary::<Int64Type, Int32Type, _, Int64Type>(
+                seconds_array,
+                nanos_array,
+                |seconds, nanos| seconds * 1_000_000_000 + nanos as i64,
```
grtlr: Please use a checked version of this multiplication and output an error if we overflow.

crates/utils/re_arrow_combinators/src/semantic.rs
```diff
@@ -62,3 +67,122 @@ impl<O1: OffsetSizeTrait, O2: OffsetSizeTrait> Transform for BinaryToListUInt8<O
...
+                    unsupported => Err(Error::UnexpectedValue {
+                        expected: "'h264' or 'h265'".to_owned(),
+                        actual: unsupported.to_owned(),
+                    }),
+                }?;
+                output_builder.append_value(codec as u32);
+            }
+        }
+
+        Ok(output_builder.finish())
```
grtlr: Could this be an [`arrow::compute::try_unary`](https://docs.rs/arrow/latest/arrow/compute/fn.try_unary.html)?

crates/utils/re_arrow_combinators/src/semantic.rs
```diff
@@ -62,3 +67,122 @@ impl<O1: OffsetSizeTrait, O2: OffsetSizeTrait> Transform for BinaryToListUInt8<O
...
+    for &bad_codec in &unsupported_codecs {
+        let input_array = StringArray::from(vec![Some("h264"), Some(bad_codec)]);
+        let result = StringToVideoCodecUInt32::default().transform(&input_array);
+        assert!(result.is_err());
+        let Err(Error::UnexpectedValue { actual, .. }) = result else {
+            panic!("wrong error type");
+        };
+        assert_eq!(actual, bad_codec);
+    }
+}
```
grtlr: Do we want to move these tests into the `tests/` folder?

examples/rust/mcap_protobuf/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "mcap_protobuf"
```
grtlr, line 2: Sounded like @nikolausWest prefers this name? Then I'm fine with that too 😇.

examples/rust/mcap_protobuf/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.26.0-alpha.1+dev"
```
grtlr: Probably best to match the version number with the other crates. (`0.27-something`)

crates/utils/re_arrow_combinators/src/error.rs
```diff
@@ -34,6 +34,9 @@ pub enum Error {
         struct_fields: Vec<String>,
     },
 
+    #[error("Unexpected value: expected {expected}, got {actual}")]
+    UnexpectedValue { expected: String, actual: String },
```
grtlr: Nit: The `expected` could probably be a list of `&'static [&'static str]`. That way we would not need the `or` in the message nor the allocation.

crates/utils/re_arrow_combinators/src/semantic.rs
```diff
@@ -62,3 +67,122 @@ impl<O1: OffsetSizeTrait, O2: OffsetSizeTrait> Transform for BinaryToListUInt8<O
...
+    fn transform(&self, source: &StringArray) -> Result<Self::Target, Error> {
+        let mut output_builder = UInt32Builder::with_capacity(source.len());
+
+        for i in 0..source.len() {
+            if source.is_null(i) {
+                output_builder.append_null();
+            } else {
+                let codec = match source.value(i).to_lowercase().as_str() {
+                    "h264" => Ok(VideoCodec::H264),
+                    "h265" => Ok(VideoCodec::H265),
```
grtlr: We should be supporting more codecs in Rerun? If so we should add them all here. (you potentially need to `cfg` check for WASM)

examples/rust/mcap_protobuf/src/main.rs
```diff
@@ -0,0 +1,160 @@
...
+    let video_lens =
+        LensBuilder::for_input_column(EntityPathFilter::all(), "foxglove.CompressedVideo:message")
+            .add_time_column(
+                "timestamp",
+                TimeType::TimestampNs,
+                [
+                    Op::access_field("timestamp"),
+                    Op::func(list_timespec_to_list_nanos),
+                ],
+            )
```
grtlr: The video timeline is different then the other timelines (this will be duration, most of Foxglove look like timestamps).

grtlr: Btw, we should use `add_time_column` for all other  Foxglove types too. 🙏

crates/utils/re_arrow_combinators/src/semantic.rs
MichaelGrupp: Might be possible with `try_binary`. Checking... https://docs.rs/arrow/latest/arrow/compute/fn.try_binary.html

crates/utils/re_arrow_combinators/src/semantic.rs
MichaelGrupp: Confirmed with @Wumpf: We don't support more for `VideoStream`. (for the other video asset types etc, see here https://rerun.io/docs/reference/video)

crates/utils/re_arrow_combinators/src/semantic.rs
MichaelGrupp: Yes, forgot that in my own PR 😄

examples/rust/mcap_protobuf/Cargo.toml
MichaelGrupp: yes!

crates/utils/re_arrow_combinators/src/semantic.rs
MichaelGrupp: Implemented here: https://github.com/rerun-io/rerun/pull/11789/commits/4c786f8a19d0cbe3b59e3f57cdf52e837b0613aa

crates/utils/re_arrow_combinators/src/semantic.rs
MichaelGrupp: No, the input `StringArray` doesn't qualify as primitive array (because variable size string) ... `try_unary` only works with primitve types of fixed size. Not sure if there's a different Arrow builtin that could work here.

On the syntactic sugar side I found that `try_fold` could be a candidate (does the same as the raw loop in the end)

```rust
        Ok(source
            .iter()
            .try_fold(
                UInt32Builder::with_capacity(source.len()),
                |mut builder, maybe_str| {
                    if let Some(codec_str) = maybe_str {
                        let codec = match codec_str.to_lowercase().as_str() {
                            "h264" => VideoCodec::H264,
                            "h265" => VideoCodec::H265,
                            _ => {
                                return Err(Error::UnexpectedValue {
                                    expected: &["h264", "h265"],
                                    actual: codec_str.to_owned(),
                                });
                            }
                        };
                        builder.append_value(codec as u32);
                    } else {
                        builder.append_null();
                    }
                    Ok(builder)
                },
            )?
            .finish())
```

examples/rust/mcap_protobuf/src/main.rs
MichaelGrupp: So you mean it should be `TimeType::DurationNs` instead, but only here? I'm not sure about this. Looking  at https://docs.foxglove.dev/docs/sdk/schemas/timestamp#schema, it says that a `Timestamp` is "seconds since a **user-defined epoch**". So in general, can we ever assume that it's relative to the 1970 UNIX epoch? (although many people do that probably).

MichaelGrupp: As discussed in Slack. https://github.com/rerun-io/rerun/pull/11789/commits/90a15b4f90faaf922459f648a206c18c20f30ec3

crates/utils/re_arrow_combinators/src/semantic.rs
```diff
@@ -78,46 +79,20 @@ impl Transform for TimeSpecToNanos {
...
-
-        Ok(
-            arrow::compute::binary::<Int64Type, Int32Type, _, Int64Type>(
-                seconds_array,
-                nanos_array,
-                |seconds, nanos| seconds * 1_000_000_000 + nanos as i64,
-            )?,
-        )
+        let seconds_array = DowncastRef::<Int64Type>::new()
+            .transform(&GetField::new("seconds").transform(source)?)?;
```
grtlr: Nit: You should be able to write something like.

```suggestion
        let seconds_array = GetField::new("seconds").then(DowncastRef<Int64Type>::new()).transform(source)?;
```

crates/utils/re_arrow_combinators/src/semantic.rs
grtlr: Specifically, I think it would be great to create a dedicated file for each `Transform` there. We don't have to do all in this PR, but this might be a good start.

grtlr (approved):
That turned out really nice! 

Only some minor comments around the test case for `Transform`s.

grtlr:
I just realized that the following PR landed:

* https://github.com/rerun-io/rerun/pull/11849

crates/utils/re_arrow_combinators/src/semantic.rs
MichaelGrupp: test added

crates/utils/re_arrow_combinators/src/semantic.rs
MichaelGrupp: done (for semantic)

MichaelGrupp:
Added AV1 support and test to catch new codecs.

Issue #11174, opened 2025-09-11
ROS2 `tf` messages don't map to entity hierarchy


grtlr:
**Describe the bug**

We currently map MCAP topics to Rerun entities. But since `tf` messages are logged to separate topics, they don't show up in the correct places in the Rerun entity hierarchy and are therefore not correctly applied.

**To Reproduce**
1. Open a ROS2 file with `tf` messages
2. Transformations are not propagated correctly.

**Expected behavior**

We should have a way to restore the transformation hierarchy from MCAP files.

As a short-term solution we could supply the MCAP loader with a config (file) that explains how to map `tf` messages to corresponding topics.

A more long-term solution could be entity path _references_ that would allow us to "teleport" entities to other parts of the entity hierarchy.

grtlr:
The following issue:

* #9949

Has more information from users in this comment:

* https://github.com/rerun-io/rerun/issues/9949#issuecomment-3285360716

Wumpf:
Also related to:
* https://github.com/rerun-io/rerun/issues/1522
* https://github.com/rerun-io/rerun/issues/5242

PR #11790, opened 2025-11-05
Support arbitrary transform frame based hierarchies, fully independent of entity hierarchy

Wumpf:
### Related

* Fixes RR-2685
* Fixes https://github.com/rerun-io/rerun/issues/11174

### What

With this iteration the parent/child frames on `Transform3D` can now form fully independent graphs!

Practically, this means that you can flatten out your entity hierarchy, but still  have a deep transform hierarchy modeled solely with `Transform3D` & `CoordinateFrame` archetypes: 
image

The reference/target frame of the view is the _coordinate frame at `space origin`_. Which is why the example shown here has this line:
```py
rr.log("/", rr.CoordinateFrame("sun_frame"))
```

Alternatively you could also leave it at its default `tf#/` but then you have to connect it up:
```py
rr.log("wherever", rr.Transform3D(child_frame="sun_frame", parent_frame="tf#/"))
```

⚠️ there are still plenty of rough edges & surprises around pinholes, instance poses and (most suprisingly) the implicit transforms on boxes/capsules/cylinder/ellipsoids.

### DIFF
--- a/crates/store/re_tf/src/transform_forest.rs
+++ b/crates/store/re_tf/src/transform_forest.rs
@@ -843,14 +843,18 @@ fn transforms_at(
 
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use std::sync::Arc;
+
     use itertools::Itertools as _;
+
+    use super::*;
+
+    use crate::TransformFromToError::NoPathBetweenFrames;
     use re_chunk_store::Chunk;
     use re_entity_db::EntityDb;
-    use re_log_types::{StoreInfo, TimePoint, TimelineName};
+    use re_log_types::{StoreInfo, TimeCell, TimePoint, TimelineName};
     use re_types::components::TransformFrameId;
     use re_types::{Archetype as _, RowId, archetypes};
-    use std::sync::Arc;
 
     fn test_pinhole() -> archetypes::Pinhole {
         archetypes::Pinhole::from_focal_length_and_resolution([1.0, 2.0], [100.0, 200.0])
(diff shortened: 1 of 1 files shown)


crates/store/re_tf/src/snapshots/re_tf__transform_forest__tests__simple_entity_hierarchy__transform_from_to__top.snap
```diff
@@ -315,10 +315,10 @@ expression: "pretty_print_transform_frame_ids_in(&result, &all_transform_frame_i
         ),
     ),
     (
-        tf#/top/nonexistent,
+        TransformFrameIdHash(D01893A02DC02A5F),
```
Wumpf, line 318: this is a bit of collateral damage from switching the pretty printing over to the frame-id registry. Overall easier to read/better test code, so I'm ok with that.

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -228,118 +225,255 @@ pub struct TransformForest {
...
+        entity_db.tree().visit_children_recursively(|entity_path| {
+            unprocessed_sources.insert(TransformFrameIdHash::from_entity_path(entity_path));
+        });
+
+        let mut transform_stack = Vec::new(); // Keep pushing & draining from same vector as a simple performance optimization.
+
+        let mut forest = Self {
+            root_from_frame: IntMap::default(),
+            roots: IntMap::default(),
         };
```
grtlr:
```suggestion
        let mut forest = Self::default();
```

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -228,118 +225,255 @@ pub struct TransformForest {
...
-        let transform_frame_id = TransformFrameIdHash::from_entity_path(&subtree.path);
-        let previous_transform = self
-            .root_from_frame
-            .insert(transform_frame_id, transform_root_from_parent);
-        debug_assert!(previous_transform.is_none(), "Root was added already"); // TODO(RR-2667): Build out into cycle detection (cycles can't _yet_ happen)
+        // Figure out the root frame for this entire stack.
+        let (mut root_frame, mut root_from_target) = if let Some(parent_from_child) =
+            top_of_stack.parent_from_child.as_ref()
+        {
+            // We have a connection further up the stack. That might we must have stopped because we already know that target!
```
grtlr:
```suggestion
            // We have a connection further up the stack. That means we must have stopped because we already know that target!
```

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -228,118 +225,255 @@ pub struct TransformForest {
...
+        query: &LatestAtQuery,
     ) -> Self {
         re_tracing::profile_function!();
 
-        let entity_tree = recording.tree();
-        let transforms = transform_cache.transforms_for_timeline(time_query.timeline());
+        let transforms = transform_cache.transforms_for_timeline(query.timeline());
 
-        let entity_hierarchy_root = TransformFrameIdHash::entity_path_hierarchy_root();
+        let mut unprocessed_sources: IntSet<_> = transforms.all_child_frames().collect();
```
grtlr: The `IntSet` _feels_ problematic to me wrt. to worst case complexity because it gives us less control about the order in which we visit nodes. I cannot pin it down though, because also we _should_ memoize everything.

But because we don't control the order, we have O(n) worst case memory for the `transform_stack` (where it could be O(1) theoretically). That's because the algorithm here is not a graph _walk_ but rather dynamic programming instead.

Also, probably not relevant for our use case, but maybe worth pointing these things out in the overview/docs.

grtlr: The term "sources" threw me off here (because I immediately thought about leafs). Maybe just "node" to emphasize that they can be at arbitrary positions in the tree?

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -228,118 +225,255 @@ pub struct TransformForest {
...
 
-        let mut tree = Self {
-            roots: std::iter::once((entity_hierarchy_root, TransformTreeRootInfo::EntityRoot))
-                .collect(),
-            root_from_frame: Default::default(),
+        // We also have to add implicit frame ids for all entities in the entity _tree_.
+        // That's more than just the entity paths that have things logged on since there might be arbitrary steps without any data.
+        entity_db.tree().visit_children_recursively(|entity_path| {
+            unprocessed_sources.insert(TransformFrameIdHash::from_entity_path(entity_path));
+        });
```
grtlr: Small observation: Isn't this the `unprocessed_sources` now very similar to the id registry (except that it's missing the path segments)?

crates/store/re_tf/src/transform_forest.rs
grtlr: Also, what does that mean for cycle detection?

Wumpf: not controlling the order is a problem we run into with every approach. At some point we'll also have a "take a node, see how it fits in" step. I do agree though that the approach here is particularly temperamental since the transform stack size( lenght of a walk) is random

PR #11795, opened 2025-11-05
Trigger Rust tests on CI when a snapshot png changes

### LINKED ISSUE #11794: Snapshot png file changes don't trigger CI
If you only change a snapshot png, CI won't run the test that created it.
Example https://github.com/rerun-io/rerun/pull/11793

aedm:
Fixes #11794


Wumpf (approved):
seems legit and I guess self-explantory enough
do we trust this to work without testing it? 🤔

aedm:
> seems legit and I guess self-explantory enough do we trust this to work without testing it? 🤔

Tested, works as expected.

PR #11798, opened 2025-11-05
Parallel ingestion

joelreymont:
Redesign ingestion with dedicated ingestion worker

Addresses #4298 by moving CPU-intensive Arrow message processing off the UI thread.

Key changes:

1. Ingestion Worker (crates/viewer/re_viewer/src/ingestion_worker.rs):
   - Dedicated background thread for processing Arrow messages into chunks
   - Bounded channel (capacity: 2000) provides natural backpressure

2. Viewer Integration (crates/viewer/re_viewer/src/app.rs):
   - Arrow messages submitted to worker queue instead of blocking UI thread
   - Poll worker output during update() and add chunks to EntityDb

### DIFF
--- a/crates/store/re_entity_db/src/entity_db.rs
+++ b/crates/store/re_entity_db/src/entity_db.rs
@@ -668,7 +668,7 @@ impl EntityDb {
     /// On native: Lazily creates worker on first call, then queues message for processing
     /// On Wasm: This method doesn't exist (compile error if called)
     ///
-    /// The caller should periodically call [`Self::poll_worker_output`] to retrieve
+    /// The caller should periodically call [`Self::on_frame_start`] to retrieve
     /// processed chunks and add them to the store.
     #[cfg(not(target_arch = "wasm32"))]
     pub fn submit_arrow_msg(


abey79:
Can you please rebase your PR? Seems it's based on a rather old commit from `main`. Notably arrow has been updated since.

joelreymont:
> Can you please rebase your PR? Seems it's based on a rather old commit from `main`. Notably arrow has been updated since.

Done!

abey79:
> > Can you please rebase your PR? Seems it's based on a rather old commit from `main`. Notably arrow has been updated since.
> 
> Done!

I'm still seeing your fork's `main` being quite a few commits behind ours.

joelreymont:
Mea culpa! Should be good now. 

Also, I'm available on Discord if you have any questions.

abey79:
I'm observing a fairly large reduction of the "time to fully loaded" with a 2h air traffic data RRD: 30s -> 9s.

### `main`

https://github.com/user-attachments/assets/48c36fc3-9152-40ae-8c43-52114dcde426

### this PR

https://github.com/user-attachments/assets/9204d6c6-aeb5-4a80-9f6d-b0c08ef94176

abey79:
Based on the above benchmark, could you provide a more thorough analysis of the performance profile of this PR with Puffin? 

You should be able to produce the rrd using `pixi run -e examples air_traffic_data --dataset 2h --save output.rrd` (I can also send mine if needed).

crates/store/re_entity_db/src/entity_db.rs
```diff
@@ -554,16 +554,16 @@ impl EntityDb {
...
                 chunk.sort_if_unsorted();
-                self.add_chunk_with_timestamp_metadata(
-                    &Arc::new(chunk),
-                    &chunk_batch.sorbet_schema().timestamps,
-                )?
+
+                let chunk = Arc::new(chunk);
+                let timestamps = chunk_batch.sorbet_schema().timestamps.clone();
+
+                self.add_chunk_with_timestamp_metadata(&chunk, &timestamps)?
```
abey79, line 598: I'm seeing quite a few diffs in this file and they seem spurious. Am I missing something?

joelreymont:
CleanShot 2025-11-06 at 13 01 07@2x
CleanShot 2025-11-06 at 13 01 30@2x

The compressed Puffin file is [here](https://drive.google.com/file/d/1NyetZQ3pTrrULSep_5F3dq8s4bfK8Xav/view?usp=share_link).

joelreymont:
I'm working on it!

joelreymont:
This is what the old ingestion performance looks like, using the floating metrics window from PR #11820 against the `main` branch. 

https://github.com/user-attachments/assets/212cc794-db54-43d1-bb36-dc056d98dd5a

Time elapsed is ~65s to load the air traffic 2h data set. Throughput is 2017msg/s.

sequential

And this is what new ingestion performance looks like, using the same floating metrics window, adapted to parallel processing and included in this PR.

https://github.com/user-attachments/assets/559afd09-16bf-4f16-b0dc-de6e3f0e6ef7

You can see that ingestion time went down to ~12s and throughput went up >5x.

parallel

joelreymont:
@abey79 This confirms your findings from yesterday. It also confirms that performance did not decrease with the redesign.

crates/store/re_entity_db/src/entity_db.rs
```diff
@@ -629,6 +663,72 @@ impl EntityDb {
...
+
+        if let Some(worker) = &self.ingestion_worker {
+            worker.submit_arrow_msg_blocking(
+                self.store_id.clone(),
+                arrow_msg,
+                channel_source,
+                msg_will_add_new_store,
+            );
+        }
+    }
```
abey79, line 694: Good opportunity to use `Option::get_or_insert_with` here. It better expresses the intent and doesn't hide the important code behind a scary `if`

crates/viewer/re_viewer/src/app.rs
```diff
@@ -100,6 +100,9 @@ pub struct App {
     /// Interface for all recordings and blueprints
     pub(crate) store_hub: Option<StoreHub>,
 
+    /// Metrics for tracking RRD loading performance (when profiling is active)
+    pub(crate) rrd_loading_metrics: crate::rrd_loading_metrics::RrdLoadingMetrics,
```
abey79: Looks like part of  #11820 made it into this PR. Either it's a hard pre-requisite, then it should ideally be chained PRs, or it's not, then it should be removed from the present PR so we address both separately.

crates/viewer/re_viewer/src/app.rs
```diff
@@ -2276,6 +2765,149 @@ impl App {
...
     }
 
+    #[cfg(not(target_arch = "wasm32"))]
+    fn process_ingestion_worker_output(
+        &mut self,
+        store_hub: &mut StoreHub,
+        egui_ctx: &egui::Context,
+    ) {
+        re_tracing::profile_function!();
+
```
abey79: Everything about the ingestion worker should be fully encapsulated in entity dbs.

My understanding is that entity db needs to be "stepped" at each frames in order to process the result of their respective worker. I suggest to use the `on_frame_start` pattern that several subsystem have. EntityDb are stored in a `StoreBundle`. It should have a `on_frame_start` method, which calls into each of the enclosed entity dbs' `on_frame_start`.

Ideally, nothing related to the ingestion leaks from `EntityDb`, except for that "on_frame_start` method.

crates/viewer/re_viewer/src/app.rs
joelreymont: I would prefer chained PRs as there's a fair bit of difference in booking up the metrics and display between the PR on `main` and this branch.

crates/viewer/re_viewer/src/app.rs
joelreymont: Will fix!

joelreymont:
Before:
```
App ──> knows about workers
 └──> manually iterates EntityDbs
      └──> calls poll_worker_output()
           └──> processes results
```
After:
```
App ──> store_hub.on_frame_start()
          └──> store_bundle.on_frame_start()
                └──> entity_db.on_frame_start() (for each)
                      └──> worker.poll_processed_chunks() ✅ Encapsulated!
```

The metrics are key to the whole thing, in my opinion, as is performance tracking. I don't know how to tell if the PR provides true performance enhancements without metrics! 

I kept the metrics as the `HEAD` so I could drop them from this PR and create another one on top of this same branch.

abey79:
> The metrics are key to the whole thing, in my opinion, as is performance tracking. I don't know how to tell if the PR provides true performance enhancements without metrics!
> 
> I kept the metrics as the `HEAD` so I could drop them from this PR and create another one on top of this same branch.

I'm not saying metrics are not useful, and it's great to have them to document the performance increase. But I cannot review this PR, let alone land it, unless the diff only covers the actual ingestion part. Using git, you should be able to rebase the metrics UI stuff on top of this branch, such that you can use it locally without it appearing here.

joelreymont:
Cut the head off, please check!

joelreymont:
Should be ready to go!

PR #11801, opened 2025-11-06
Fix trace context propagation

zehiko:
### What

While looking at traces we realized that at least on the server side we don't have context propagation and so for example ``FetchChunks`` call would end up as 2 individual traces for each of our current services on the path. 

It seems that the problem was the ordering where span creation would happen after context extraction that was in ``TracingExtractorInterceptor``. Trying to reverse the order didn't work as well as I couldn't find a way to preserve the context except for some very hacky ways. Instead I stumbled upon ``HeaderExtractor`` and now using it in ``GrpcMakeSpan``.  There's admittedly way too many layers and crates in this ecosystem for me to really understand if I've done the right thing in this limited amount of time, but it seems to work. More than happy to learn there is a simpler/better approach. 

Before:
Screenshot 2025-11-06 at 11 07 36

After:
Screenshot 2025-11-06 at 11 05 51

e2e (CLI -> server -> server)

Screenshot 2025-11-06 at 11 57 49

### DIFF
--- a/crates/utils/re_perf_telemetry/src/grpc.rs
+++ b/crates/utils/re_perf_telemetry/src/grpc.rs
@@ -10,6 +10,8 @@ const RERUN_HTTP_HEADER_SERVER_VERSION: &str = "x-rerun-server-version";
 
 /// Implements [`tower_http::trace::MakeSpan`] where the trace name is the gRPC method name.
 ///
+/// See details in `make_span` on how we ensure context propagation.
+///
 /// We keep track of a bunch of relevant in-house state associated with the span in `SpanMetadata`.
 #[derive(Debug, Clone)]
 pub struct GrpcMakeSpan {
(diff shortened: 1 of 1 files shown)


crates/utils/re_perf_telemetry/src/grpc.rs
```diff
@@ -30,6 +30,12 @@ impl GrpcMakeSpan {
 
 impl<B> tower_http::trace::MakeSpan<B> for GrpcMakeSpan {
     fn make_span(&mut self, request: &http::Request<B>) -> tracing::Span {
+        // Extract `OpenTelemetry` context from headers before creating the span
+        let parent_ctx = opentelemetry::global::get_text_map_propagator(|prop| {
+            prop.extract(&opentelemetry_http::HeaderExtractor(request.headers()))
+        });
+        let _guard = parent_ctx.attach();
```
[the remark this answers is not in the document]
zehiko: this seems off. Context is still alive when we're creating the span.

teh-cmc, line 46: I assume this dark magic right here is why anything work at all. This needs a comment along the lines of "Having this guard in scope will automagically attach any newly created span in this scope to the OTEL context above".

teh-cmc (approved):
client-to-server-to-server and back seems to work, therefore :green_circle: :lotus_position_man:

crates/utils/re_perf_telemetry/src/grpc.rs
zehiko: yes, let me add a comment

PR #11804, opened 2025-11-06
Use GetField in StructToFixedList

MichaelGrupp:
Avoids code duplication for this common operation and makes sure that `available_fields` is only queried lazily on error.

See also comment in https://github.com/rerun-io/rerun/pull/11789#discussion_r2498549608


grtlr (changes requested):
This is probably to address my comment in the other PR? I wonder if we could just use the `GetField::transform` directly? That way we would not have to expose an extra function (that is also hard for discoverability).

MichaelGrupp:
True, that's way easier. Changed it and adapted the title of the PR. PTAL

PR #11807, opened 2025-11-06
Make last updated timestamp robust

abey79:
### Related

- part of https://linear.app/rerun/issue/RR-2640/repeated-dataframe-query-view-calls-slow-on-specific-rrd-for-oss
- part of https://linear.app/rerun/issue/RR-2715/registering-with-layers-does-not-do-an-update-to-the-dataset-updated
- sibling https://github.com/rerun-io/dataplatform/pull/1985

### What

I want to introduce a caching mechanism for dataset schema, since this operation is currently very expansive and often used in dataframe queries. A simple way to invalidate that cache would be to use the `last_updated_at` field that we maintain anyways. As a first step, this PR makes it more robust to maintain that `last_update_at` flag by introducing a `Tracked` helper which does that automagically.

Also introduces a couple of tests.

### DIFF
--- a/crates/store/re_server/src/store/dataset.rs
+++ b/crates/store/re_server/src/store/dataset.rs
@@ -117,7 +117,9 @@ impl Dataset {
     }
 
     pub fn set_dataset_details(&mut self, details: DatasetDetails) {
-        self.inner.modify().details = details;
+        if details != self.inner.details {
+            self.inner.modify().details = details;
+        }
     }
 
     pub fn as_entry_details(&self) -> EntryDetails {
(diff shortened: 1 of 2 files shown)


crates/store/re_server/src/store/dataset.rs
```diff
@@ -86,47 +98,49 @@ impl Dataset {
...
+                    .get(id)
+                    .map(|partition| (id, partition))
             })))
         }
     }
 
     pub fn set_dataset_details(&mut self, details: DatasetDetails) {
-        self.details = details;
-        self.updated_at = jiff::Timestamp::now();
+        self.inner.modify().details = details;
```
oxkitsune: Similarly to the `set_name` shouldn't it be trivial to check whether the new details are actually _new_? In the case where they are, this will update the `last_updated_timestamp` even though nothing changed.

`DatasetDetails` can implement `PartialEq` and `Eq`, all its components already do.

abey79: good call

PR #11809, opened 2025-11-06
Handle corner cases with registration of entity path derived transform frames

Wumpf:
* Fixes RR-2902

This fixes some very illusive bugs (== unexpected debug assertions triggering)

### DIFF
--- a/crates/store/re_types/src/transform_frame_id_hash.rs
+++ b/crates/store/re_types/src/transform_frame_id_hash.rs
@@ -83,7 +83,6 @@ impl TransformFrameIdHash {
     ///
     /// If the string was an entity path derived frame id, also returns that entity path.
     #[inline]
-    #[expect(clippy::should_implement_trait)]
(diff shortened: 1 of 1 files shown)


crates/store/re_tf/src/frame_id_registry.rs
```diff
@@ -32,9 +46,39 @@ impl FrameIdRegistry {
...
+                    );
+
+                if let Entry::Vacant(e) = self.frame_id_lookup_table.entry(frame_id_hash) {
+                    e.insert(TransformFrameId::new(frame_id_string.as_str()));
+
+                    // If this was an entity path, we didn't know about this one before.
+                    // Make sure we also add all parents!
+                    if let Some(entity_path) = entity_path.and_then(|path| path.parent()) {
+                        self.register_frame_id_from_entity_path(&entity_path);
+                    }
```
Wumpf, line 71: just as before I'm going pretty aggressive on avoiding unnecessary string copies and hashings in this registry. Makes this unfortunately a lil bit harder to follow, but I think it's fine

PR #11694, opened 2025-10-28
Create and write tables in the Rerun server

timsaucer:
This is the final PR in the group associated with creating and writing tables in the server. The previous PRs include:

- https://github.com/rerun-io/rerun/pull/11565
- https://github.com/rerun-io/rerun/pull/11640
- https://github.com/rerun-io/rerun/pull/11657

The above mentioned PRs add in the core functionality required for users to both create new tables and to write to existing tables in the Rerun server. This PR adds the final step, a usage example. The purpose of this example is to demonstrate a common work flow in data analysis where you wish to incrementally evaluate a dataset and write results into a table.

This PR closes https://linear.app/rerun/initiative/multistep-robotics-pipelines-f45b92ccebb6/overview

PR #11812, opened 2025-11-06
Add ability to replace/upsert in tables

timsaucer:
### Related

Closes https://linear.app/rerun/project/add-ability-to-upsert-data-in-tables-aad5989947ac

### What

This PR follows up on the work from https://github.com/rerun-io/rerun/pull/11694 and the issues linked inside it.

With this PR you can update existing tables if they have an index specified as a primary key. You do this by setting the metadata `rerun:is_table_index` on the field when creating the table. During `update` operations we will use these columns as the keys by which we update the underlying lance table.


timsaucer:
@rerun-bot full-check

jleibs (approved):

crates/store/re_server/src/store/table.rs
```diff
@@ -170,13 +168,40 @@ impl Table {
                     .map_err(|err| DataFusionError::External(err.into()))?;
             }
             InsertOp::Replace => {
-                exec_err!("Invalid insert operation. Only append and overwrite are supported.")?;
+                let key_columns: Vec<_> = dataset
```
jleibs, line 171: If `key_columns` is empty and/or disjoint with `reader.schema()`, should we / do we get an error?

timsaucer: Yes, you get the error `A merge insert operation must specify at least one on key`

Issue #11813, opened 2025-11-06
Spatial view's scroll-to-zoom gets stuck when speedup modifier is pressed


MichaelGrupp:
**Describe the bug**

The speedup modifier (shift) doesn't work with scroll-to-zoom. If you press shift while scrolling in a spatial view, it just stops.

All other inputs work with speedup (WASD, pinch zoom).

**To Reproduce**
1. open some 3D spatial view
2. scroll to zoom
3. press shift

**Expected behavior**

You can speed up also scroll-to-zoom.

**Desktop (please complete the following information):**
macOS Tahoe (probably irrelevant)

**Rerun version**
0.27 alpha / main branch

PR #11814, opened 2025-11-06
Fix 3D eye speed-up modifier not working with scroll-to-zoom

MichaelGrupp:
This is basically caused by egui's scroll behavior that changes when you press certain modifiers (see [here](https://github.com/emilk/egui/blob/31d313572b798d1924329cc074f0f22075fca712/crates/egui/src/input_state/mod.rs#L86)).

Pressing shift for horizontal scroll is useful in other views like the dataframe view, but in the spatial view it conflicts with the speed-up modifier.

This change ensures that scrolling always zooms the eye.

Fixes: #11813

### DIFF
--- a/crates/viewer/re_view_spatial/src/eye.rs
+++ b/crates/viewer/re_view_spatial/src/eye.rs
@@ -487,12 +487,7 @@ impl ControlEye {
             // egui's default horizontal_scroll_modifier is shift, which is also our speed-up modifier.
             // This means that a user who wants to speed up scroll-to-zoom will generate a horizontal scroll delta.
             // To support that, we have to check and use the horizontal delta if no vertical delta is present (see: #11813).
-            let is_maybe_horizontal = input.smooth_scroll_delta.y == 0.;
-            let scroll_delta = if is_maybe_horizontal {
-                input.smooth_scroll_delta.x
-            } else {
-                input.smooth_scroll_delta.y
-            };
+            let scroll_delta = input.smooth_scroll_delta.x + input.smooth_scroll_delta.y;
             input.zoom_delta() * (scroll_delta / 200.0).exp()
         });
 


crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -484,8 +484,16 @@ impl ControlEye {
     /// Handle zoom/scroll input.
     fn handle_zoom(&mut self, egui_ctx: &egui::Context) {
         let zoom_factor = egui_ctx.input(|input| {
-            let (zoom_delta, scroll_delta) = (input.zoom_delta(), input.smooth_scroll_delta.y);
-            zoom_delta * (scroll_delta / 200.0).exp()
+            // egui's default horizontal_scroll_modifier is shift, which is also our speed-up modifier.
+            // This means that a user who wants to speed up scroll-to-zoom will generate a horizontal scroll delta.
+            // To support that, we have to check and use the horizontal delta if no vertical delta is present.
+            let is_maybe_horizontal = input.smooth_scroll_delta.y == 0.;
```
MichaelGrupp: I first tried to check modifier states to determine this, but egui's `horizontal_scroll_modifier` is a private field and this simple numeric check does the trick too.

See web viewer of this PR to try it.

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -484,8 +484,16 @@ impl ControlEye {
...
-            zoom_delta * (scroll_delta / 200.0).exp()
+            // egui's default horizontal_scroll_modifier is shift, which is also our speed-up modifier.
+            // This means that a user who wants to speed up scroll-to-zoom will generate a horizontal scroll delta.
+            // To support that, we have to check and use the horizontal delta if no vertical delta is present (see: #11813).
+            let is_maybe_horizontal = input.smooth_scroll_delta.y == 0.;
+            let scroll_delta = if is_maybe_horizontal {
+                input.smooth_scroll_delta.x
+            } else {
+                input.smooth_scroll_delta.y
+            };
```
emilk: Maybe this will work just as well, and be a lot simpler?

```suggestion
            let scroll_delta = input.smooth_scroll_delta.x + input.smooth_scroll_delta.y;
```

crates/viewer/re_view_spatial/src/eye.rs
emilk: I recently changes some of this code in
* https://github.com/emilk/egui/pull/7678

I started making a distinction between raw _wheel_ input, and the scroll/pan vs zoom output of it. But not everything is public yet. Feel free to make a PR to egui with improvements though 🙏 

(note though that we're not currently using `egui/main`, and won't until we release Rerun 0.27 next week)

emilk (approved):

crates/viewer/re_view_spatial/src/eye.rs
MichaelGrupp: The eye zoom application here is maybe a bit special, because "zoom" here is implemented custom and not referring directly to what egui considers a zoom. If I understand correctly, egui (current release) provides the following:
* `zoom_delta()`: from pinch zoom, or from scroll delta if shift is pressed (e.g. for `egui::Scene`)
* `smooth_scroll_delta()`: smoothed scroll wheel or touchpad scroll delta, vertical delta unless shift is pressed (e.g. for `egui::ScrollArea` or tables).

In the eye control here, we want to be able to zoom both with scrolling and pinch-to-zoom, plus the modifiers have a different meaning here (slow-down / speed-up).

tldr: I think the current code is correct

PR #11815, opened 2025-11-07
Add missing `:` to file path regex for Windows in `e2e` tests

grtlr:
### Related

Attempts to fix recent CI error:

* https://github.com/rerun-io/rerun/actions/runs/19157311181/job/54764526386
* Introduced by https://github.com/rerun-io/rerun/pull/11782

### What

The regex was missing `:` which is important for Windows-based paths that usually include something like `C:`.

### TODO

* [x] `full-check`


grtlr:
@rerun-bot full-check

grtlr:
@rerun-bot full-check

PR #11816, opened 2025-11-07
Fix zooming to box and dragging axis for the time series view

IsseW:
### Related

* Closes RR-2905
* Fixes regression from #11621

### What

Fixes some time series controls not working.

### DIFF
--- a/crates/viewer/re_view_time_series/src/view_class.rs
+++ b/crates/viewer/re_view_time_series/src/view_class.rs
@@ -652,9 +652,7 @@ impl ViewClass for TimeSeriesView {
                 .id(plot_id)
                 .show_grid(**show_grid)
                 .auto_bounds(false)
-                .allow_zoom(false)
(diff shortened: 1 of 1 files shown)


Wumpf (commented):
"make egui control things again and react if things have changed", okay makes sense, that's nice!
But I thought it was moved out because of sync issue issues with blueprint? 🤔 Those turned out to be no longer a concern, because?

(not approved yet, want to check on the web viewer myself first; just to be sure :))

IsseW:
> But I thought it was moved out because of sync issue issues with blueprint? 🤔 Those turned out to be no longer a concern, because?

Yeah I thought I'd run into issues like that as well, but I'm pretty sure those were fixed by turning off auto bounds, something that also caused other issues. And properly passing zoom lock.

Wumpf (approved):
seems to work like a charm

PR #11820, opened 2025-11-07
Add RRD loading metrics tracking with UI visualization

joelreymont:
I've been looking for a better way to measure loading performance than eyeballing the UI to check when loading has ended. To that end, this PR implements metrics tracking for RRD file loading when profiling is active. 

It tracks the complete pipeline from file loading through playback completion with detailed timing breakdowns and a visual timeline display. Also prints performance breakdown to the console.

Metrics tracked:
- Time to first message
- Stream latency
- Ingestion lag time
- Playback time
- Message throughput (msgs/sec)
- Total end-to-end loading time

This is how it looks in action

https://github.com/user-attachments/assets/23bff513-36b1-4926-85ad-b1c6e5d93481

Final look

CleanShot 2025-11-07 at 13 07 26@2x

And console output

CleanShot 2025-11-07 at 13 08 03@2x

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -100,6 +100,9 @@ pub struct App {
     /// Interface for all recordings and blueprints
     pub(crate) store_hub: Option<StoreHub>,
 
+    /// Metrics for tracking RRD loading performance (when profiling is active)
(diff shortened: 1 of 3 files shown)


joelreymont:
Fixed bug where elapsed time kept running for the duration of playback.

PR #11822, opened 2025-11-07
Fix inputs when tracking entity

IsseW:
### Related

* Fixes regression in #11788

### What

Fixes regression when tracking entity where inputs don't work correctly when tracking an entity.

* [x] another round of manual testing on the web viewer build


vivekmenezes:
Since you're fixing an issue, is it possible for you to add a test to this PR?

IsseW:
~~It seems like canceling tracking with wasd doesn't work in the web-viewer 🤔~~

seems to be fixed after I cleared my browser cache

PR #11823, opened 2025-11-07
Remove time from blueprint

IsseW:
### Related

* Part of RR-2909

### What

Removes time from the time panel blueprint. Breaking change from previous version.


Wumpf:
exclude changelog since it is a no-change from 0.26 pov?

PR #11824, opened 2025-11-07
Consider `CoordinateFrame` when adding ids to `FrameIdRegistry`

grtlr:
### What

Title.

### DIFF
--- a/crates/store/re_tf/src/frame_id_registry.rs
+++ b/crates/store/re_tf/src/frame_id_registry.rs
@@ -240,9 +240,5 @@ mod tests {
             )),
             Some(&TransformFrameId::from_entity_path(&"root/surprise".into()))
         );
-        assert_eq!(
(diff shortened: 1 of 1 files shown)


crates/store/re_tf/src/frame_id_registry.rs
```diff
@@ -221,5 +240,9 @@ mod tests {
             )),
             Some(&TransformFrameId::from_entity_path(&"root/surprise".into()))
         );
+        assert_eq!(
+            registry.lookup_frame_id(TransformFrameIdHash::new(&"frame0".into())),
+            Some(&TransformFrameId::new("frame0"))
+        );
```
Wumpf: why is this assert in here again under the block that looks at all the entity paths?

Wumpf (changes requested):

crates/store/re_tf/src/frame_id_registry.rs
Wumpf:
```suggestion
```

PR #11825, opened 2025-11-07
Restructure re_arrow_combinators

MichaelGrupp:
Groups the transform implementations by the type of operations:

- `cast`: Transforms that cast arrays to different types.
- `map`: Transforms that apply operations to elements within arrays.
- `reshape`: Transforms that extract and reshape arrays.
- `semantic`: Semantic array transforms for concrete applications.


MichaelGrupp:
This only shuffles existing code around. I will add new stuff as part of the example PR #11789 to keep the diff manageable.

crates/utils/re_arrow_combinators/src/lib.rs
MichaelGrupp, line 1: Not sure if we should re-export more stuff here. I actually like that you need to import with the operation category as prefix, makes it easier to find stuff for me (`use re_arrow_combinators::{map::MapList, reshape::GetField};` instead of `use re_arrow_combinators::{MapList, GetField};`). But no strong opinion either.

grtlr: I like it the way it is in your PR! Also, exposing more is always possible, but exposing less later on usually means annoying code changes, so let's go with this!

grtlr (approved):
Awesome! Thank you for doing that 💪

PR #10828, opened 2025-08-07
Fix building Web Viewer on macOS

ntjohnson1:
### Related
Closes #9849

### What
Basically mac has a fun combination where you can have `clang`/`llvm` components but not `llvm-ar`. Mac's `ar` pads bytes in way that make llvm barf. Our pixi env for c++ has the `cxx-compiler` package which has a more full fledged llvm install. On mac on activation set `AR` to point to `llvm-ar`. It works on my machine! I'm not sure if we should have a second person try or if we should setup a test for the web build on mac just for this weird edge case.

PR #11826, opened 2025-11-07
Remove `target_pixi` again & address env var changes that caused rebuilds

Wumpf:
I got annoyed by having two (four with _wasm variants) target folders around because of unintentionally switching between the pixi environment & "raw" invocations.

This PR removes `target_pixi` again and addresses the reason we introduced this in the first place: there are certain environment variables in the pixi environment that get picked up by rust build scripts.
Using `CARGO_LOG=cargo::core::compiler::fingerprint=info` I identified two:
* `AR`: we set this on MacOS to solve some issues with our wasm build, see https://github.com/rerun-io/rerun/pull/10828
* `CONDA_PREFIX`: py03 uses this to govern which python install to use

The fix in both cases is to move those env vars to `.cargo/config.toml`. For `AR` that's trivial, for `CONDA_PREFIX` not so much! Every pixi environment has its own conda prefix (and yes that would also cause rebuilds between pixi environments!). So instead I'm setting `PYO3_PYTHON` which directly governs which python install pyo3 uses for linking. Here be dragons (a little bit ;-)): This means we no longer fix the python version used and instead rely on whatever `python` points to.

things I tried to verify this:
* [x] `pixi run cargo check` & `cargo check` not causing to rebuild each other
* [x] clean pixi environment, then run `pixi run py-build` then execute some script that needs a Rerun install
* [x] macos web build via `pixi run rerun-web`
* [x] full-ci

out of pragmatism I left the pixi target folder in git & prettier ignore lists 🤷

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -12,6 +12,17 @@ run-wasm = "run --release --package run_wasm --"
 # so that we don't run them on cargo publish or on users machines.
 IS_IN_RERUN_WORKSPACE = "yes"
 
+# Depending on the pixi environment, we get different `CONDA_PREFIX` values.
+# This means that changing the pixi environment (or not using it in the first place) will cause rebuilds.
+# To avoid unnecessary rebuilds, we set a fixed value here.
+# For details see:
+# https://pyo3.rs/main/building-and-distribution.html#configuring-the-python-version
+PYO3_PYTHON = "python"
+
+# TODO(rust-lang/stacker#127): Work around until wasm build on mac is resolved
+[target.osx-arm64.activation.env]
+AR = "llvm-ar"
+
 # `web_sys_unstable_apis` is required to enable the web_sys clipboard API which egui_web uses.
 # https://wasm-bindgen.github.io/wasm-bindgen/api/web_sys/struct.Clipboard.html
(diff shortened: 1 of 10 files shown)


Wumpf:
@rerun-bot full-check

.cargo/config.toml
```diff
@@ -12,6 +12,17 @@ run-wasm = "run --release --package run_wasm --"
...
 IS_IN_RERUN_WORKSPACE = "yes"
 
+# Depending on the pixi environment, we get different `CONDA_PREFIX` values.
+# This means that changing the pixi environment (or not using it in the first place) will cause rebuilds.
+# To avoid unnecessary rebuilds, we set a fixed value here.
+# For details see:
+# https://pyo3.rs/main/building-and-distribution.html#configuring-the-python-version
+PYO3_PYTHON = "python"
+
+# TODO(rust-lang/stacker#127): Work around until wasm build on mac is resolved
```
ntjohnson1, line 22: Mostly an FYI that stacker took my PR. Doing a `cargo tree --invert stacker` shows that everything depends on it indirectly through `recursive` which doesn't pin the patch version. That means on the next build of `datafusion` > 50.3 we should no longer need the AR workaround.

PR #11830, opened 2025-11-09
Use ": " separator in re_error::format to match anyhow's {:#} format

### LINKED ISSUE #8681: `anyhow::Error::to_string` doesn't print the full context
This has bitten us many times. That's why we have `re_error::format`.

Suggested fix:
* Switch to https://github.com/eyre-rs/eyre/tree

joelreymont:
Instead of using " -> " as the separator between error contexts, use ": " which matches the output of anyhow's `format!("{:#}", err)`.

Fixes #8681

### DIFF
--- a/crates/utils/re_error/src/lib.rs
+++ b/crates/utils/re_error/src/lib.rs
@@ -11,9 +11,11 @@ pub fn format(error: impl AsRef<dyn std::error::Error>) -> String {
 ///
 /// Always use this when displaying an error, especially `anyhow::Error`.
 pub fn format_ref(error: &dyn std::error::Error) -> String {
+    // Use ": " as separator to match anyhow's `format!("{:#}", err)` output
+    // See: https://github.com/rerun-io/rerun/issues/8681
     let mut string = error.to_string();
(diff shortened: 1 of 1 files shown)


emilk (approved):

Wumpf:
This doesn't really fix https://github.com/rerun-io/rerun/issues/8681 though, no? nothing changed in how much context is printed

joelreymont:
> This doesn't really fix #8681 though, no? nothing changed in how much context is printed

It does it _automatically_, I suppose, and does not require _building_ the context.

Do you want something else?

Wumpf:
don't get it. the ticket talks about `anyhow::Error::to_string` not printing everything. This change only changes what re_error::format looks like, not how `anyhow::Error::to_string` plays out in Rerun. Clearly @emilk was only tracking that problem for Rerun and not proposing any particular solution in the ticket yet (other than maybe use `re_error::format` more often). A solution _could_ be to catch and forbid `anyhow::Error::to_string` if that's possible via the clippy forbidden method list for example, but I doubt it

PR #11832, opened 2025-11-09
Forbid unwrap (#3408)

### LINKED ISSUE #3408: Forbid `unwrap()` in our code
All `unwrap()`s will eventually cause a panic. For example:
* https://github.com/rerun-io/rerun/issues/3270
* https://github.com/rerun-io/rerun/issues/3393

If you can prove the `unwrap()` cannot happen, then you can often rewrite it in a way to avoid the `unwrap()`. If you can't: add a comment explaining why it is impossible and/or replace it with `unreachable!()`.

I suggest we turn on `clippy::unwrap_used` to enforce this.
Right now there are too many `unwrap()`s (~350) to make this feasible, and that scares me. Those are crashes waiting to happen.

So the first step is to start removing `unwraps` in our code.

* For tests we can add `#[allow(clippy::unwrap_used)]` or use `?` instead.
* For examples we should use `?` instead.
* For production code we should use `?` or `.unwrap_or_warn` or similar.

We should also:
* Remove and the `try_` prefixes of functions

We can start working on this on a per-crate basis, e.g. adding `#![warn(clippy::unwrap_used)]` to `crates/re_types/src/lib.rs` first, and then take it from there.

joelreymont:
Closes #3408

### DIFF
--- a/crates/build/re_types_builder/CODEGEN_CHANGES.md
+++ b/crates/build/re_types_builder/CODEGEN_CHANGES.md
@@ -1,0 +1,71 @@
+# Code Generator Changes for expect() Migration
+
+## What Changed
+
+The code generator has been updated to emit `expect()` with descriptive messages instead of `unwrap()` with `#[expect(clippy::unwrap_used)]` attributes.
+
+### Modified Files
+
+1. **crates/build/re_types_builder/src/codegen/rust/serializer.rs**
+   - Line 636-639 and 655-658: Changed `offsets.last().copied().unwrap()` to use `expect()`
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -756,7 +756,7 @@ unused_peekable = "warn"
 unused_rounding = "warn"
 unused_self = "warn"
 unused_trait_names = "warn"
-unwrap_used = "warn"
+unwrap_used = "deny"
```
emilk, line 759: We don't want to deny it - we want to fix all the `// TODO(#3408)` in the code

PR #11834, opened 2025-11-09
When closing recordings (and blueprints), drop them on separate threads for UI responsiveness.

kpreid:
In `StoreHub::remove_store()`, execute drop of the store on a separate thread created for the purpose.

This prevents the UI from hanging while dropping a large recording, which is useful when the user is trying to close multiple recordings at once.

If threads are not available, automatically falls back to dropping on the current thread.

### DIFF
--- a/crates/viewer/re_viewer_context/src/store_hub.rs
+++ b/crates/viewer/re_viewer_context/src/store_hub.rs
@@ -345,7 +345,7 @@ impl StoreHub {
         // so that recursing through it and freeing the memory doesn’t block the UI thread.
         #[allow(
             clippy::allow_attributes,
-            clippy::disallowed_types
+            clippy::disallowed_types,
             reason = "If this thread spawn fails due to running on Wasm (or for any other reason),
                       the error will be ignored and the store will be dropped on this thread."
         )]


crates/viewer/re_viewer_context/src/store_hub.rs
```diff
@@ -340,6 +340,12 @@ impl StoreHub {
                     .retain(|_, id| id != store_id);
             }
         }
+
+        // Drop the store itself on a separate thread,
+        // so that recursing through it and freeing the memory doesn’t block the UI thread.
+        _ = std::thread::Builder::new()
+            .name("drop-removed-store".into())
+            .spawn(|| drop(removed_store));
```
emilk:
```suggestion
            .spawn(|| {
                re_tracing::profile_scope!("drop store");
                drop(removed_store)
            });
```

crates/viewer/re_viewer_context/src/store_hub.rs
```diff
@@ -340,6 +340,15 @@ impl StoreHub {
...
             }
         }
+
+        // Drop the store itself on a separate thread,
+        // so that recursing through it and freeing the memory doesn’t block the UI thread.
+        _ = std::thread::Builder::new()
+            .name("drop-removed-store".into())
+            .spawn(|| {
+                re_tracing::profile_scope!("drop store");
+                drop(removed_store)
```
emilk:
```suggestion
                drop(removed_store);
```

Wumpf:
CI is unhappy because this code gets hit on wasm (gets correctly stuck in our disallowed method list for wasm)

kpreid:
> CI is unhappy because this code gets hit on wasm (gets correctly stuck in our disallowed method list for wasm)

Whoops, I failed to notice that that lint was not a preexisting one but caused by my change, due to misreading what it was pointing at.

The code already works as intended on wasm: the thread spawn returns an error, and this error is discarded, causing the store to be dropped on the current thread. I’ve added an `#[allow]` with explanation of this, and adjusted the code to be lint-free.

crates/viewer/re_viewer_context/src/store_hub.rs
```diff
@@ -340,6 +340,21 @@ impl StoreHub {
                     .retain(|_, id| id != store_id);
             }
         }
+
+        // Drop the store itself on a separate thread,
+        // so that recursing through it and freeing the memory doesn’t block the UI thread.
+        #[allow(
+            clippy::allow_attributes,
+            clippy::disallowed_methods,
```
Wumpf:
```suggestion
            clippy::disallowed_types
```

nasty.

crates/viewer/re_viewer_context/src/store_hub.rs
```diff
@@ -340,6 +340,21 @@ impl StoreHub {
                     .retain(|_, id| id != store_id);
             }
         }
+
+        // Drop the store itself on a separate thread,
+        // so that recursing through it and freeing the memory doesn’t block the UI thread.
+        #[allow(
+            clippy::allow_attributes,
+            clippy::disallowed_types
```
Wumpf: sry, that was me 😅 
```suggestion
            clippy::disallowed_types,
```

PR #11835, opened 2025-11-09
Automatically build the web viewer

joelreymont:
It's annoying to have to interrupt the Rust build to build the web viewer. This PR replaces the hard assertion with automatic web viewer building when WASM files are missing. Also, it detects build profile and runs the appropriate `pixi` command with a custom `CARGO_TARGET_DIR` to prevent workspace lock conflicts.

### DIFF
--- a/crates/build/re_dev_tools/src/build_web_viewer/lib.rs
+++ b/crates/build/re_dev_tools/src/build_web_viewer/lib.rs
@@ -135,9 +135,9 @@ pub fn build(
             cmd.arg("--profile=web-release");
         }
 
-        // Note that we can't use RUSTFLAGS here directly since having more than one flag on set via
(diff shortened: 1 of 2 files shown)


emilk (changes requested):
We used to do this, but stopped it.

Unless we are VERY careful it will be triggered by:

* Rust analyzer
* Users adding `rerun` as a dependency in their `Cargo.toml`
* Users running `cargo install rerun-cli`

That's partially why we decided a long time ago to not do too many magic things in `build.rs` files, and instead move those to `pixi.toml`. In other words: pixi can invoke cargo, but not the other way around.

> It's annoying to have to interrupt the Rust build to build the web viewer.

I agree, but `pixi run rerun` should be your main way to build, and it will take care of building the web viewer for you.

PR #11836, opened 2025-11-10
Fix TODO(#3408): Remove all unwrap() calls

### LINKED ISSUE #3408: Forbid `unwrap()` in our code

joelreymont:
Replaced unwrap() with proper error handling:
- Use expect() with clear messages for invariants
- Use ? for Result propagation
- Add descriptive error context for build scripts
- Remove all clippy::unwrap_used annotations

Closes #3408


Wumpf (changes requested):
https://github.com/rerun-io/rerun/issues/3408 says explicitly:

> The purpose of this issue is NOT to replace unwrap with expect. The latter still crashes, just with a nicer message.
> The purpose of this issue is to AVOID crashes, and replace them with Err and Result.

But as far as I can tell from a quick scroll over things that's all you did in there? Sure for some situations that's fine, but that shouldn't be the rule especially not in places like the time panel

joelreymont:
> #3408 says explicitly:
> 
> > The purpose of this issue is NOT to replace unwrap with expect. The latter still crashes, just with a nicer message.
> > The purpose of this issue is to AVOID crashes, and replace them with Err and Result.
> 
> But as far as I can tell from a quick scroll over things that's all you did in there? Sure for some situations that's fine, but that shouldn't be the rule especially not in places like the time panel

Fair, I'm gonna have to go back to the drawing board.

To clarify, I would need to trace the paths upwards from each use of `unwrap` and figure out where the error could or should be caught. For example, in an existing error handler. Am I on the right track?

PR #11842, opened 2025-11-10
Fix double memcpy in Rust blob serialization of image and other blob-based types

### LINKED ISSUE #11590: Image constructors (from_color_model_and_bytes, etc) cause double memcpy
**Describe the bug**
Image is memcopied twice in Image::from_elements and other constructors.

You can see it in the profiler, first at from_color_model_and_bytes ->  `bytes.into()`, then in the new constructor on the line `try_serialize_field(Self::descriptor_buffer())`. Both functions call to libc memcpy, which is inefficient

**To Reproduce**
1. call any Image constructor

**Expected behavior**
Does 1 or 0 memcopies 

**Screenshots**

Image

**Backtrace**

Manjaro Rolling Release

**Rerun version**
0.25.1

joelreymont:
Optimize Blob serialization to avoid unnecessary concat() for single-blob case.

Changes:
- Code generator: Detect single-blob and clone ScalarBuffer directly
- Generated blob.rs: Apply optimization to to_arrow_opt
- Test single/multi-blob cases

Closes #11590.

### DIFF
--- a/crates/build/re_types_builder/src/codegen/rust/serializer.rs
+++ b/crates/build/re_types_builder/src/codegen/rust/serializer.rs
@@ -767,7 +767,8 @@ fn quote_arrow_field_serializer(
                     InnerRepr::ScalarBuffer => {
                         // Special optimization for Blob (Vec<u8>): reuse ScalarBuffer for single-blob case.
                         // For other ScalarBuffer types, use standard concat approach.
-                        let is_blob = matches!(inner_datatype, DataType::Atomic(AtomicDataType::UInt8));
+                        let is_blob =
+                            matches!(inner_datatype, DataType::Atomic(AtomicDataType::UInt8));
 
                         if is_blob {
                             // Blob optimization: avoid allocation for single-buffer case
--- a/crates/store/re_types/src/datatypes/blob_ext.rs
+++ b/crates/store/re_types/src/datatypes/blob_ext.rs
@@ -64,8 +64,8 @@ impl std::ops::Deref for Blob {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use arrow::array::Array;
-    use re_types_core::Loggable;
+    use arrow::array::Array as _;
+    use re_types_core::Loggable as _;
 
     #[test]
     fn test_single_blob_serialization() {
@@ -213,7 +213,9 @@ mod tests {
         let offsets = list_array.offsets();
         assert_eq!(offsets[0], 0);
         assert_eq!(offsets[1], 0);
-        assert_eq!(offsets[2], data.len() as i32);
+        #[expect(clippy::cast_possible_wrap)]
+        let expected_offset = data.len() as i32;
+        assert_eq!(offsets[2], expected_offset);
 
         let validity = list_array.nulls().expect("Null buffer is expected");
         assert!(!validity.is_valid(0));


crates/build/re_types_builder/src/codegen/rust/serializer.rs
```diff
@@ -765,15 +765,28 @@ fn quote_arrow_field_serializer(
...
                         quote! {
-                            #data_src
-                                .iter()
-                                #flatten_if_needed
-                                .map(|b| b as &[_])
-                                .collect::<Vec<_>>()
-                                .concat()
-                                .into()
+                            {
+                                let buffers: Vec<_> = #data_src
```
Wumpf: that's a needless Vec allocation if there's only a single buffer

crates/build/re_types_builder/src/codegen/rust/serializer.rs
```diff
@@ -765,15 +765,28 @@ fn quote_arrow_field_serializer(
 
                 match inner_repr {
                     InnerRepr::ScalarBuffer => {
-                        // TODO(emilk): this can probably be optimized
+                        // Reuse the ScalarBuffer for the single-blob case.
+                        // Fall back to concat for multple blobs.
```
Wumpf:
```suggestion
                        // Fall back to concat for multiple blobs.
```

crates/store/re_types/src/datatypes/blob_ext.rs
```diff
@@ -60,3 +60,196 @@ impl std::ops::Deref for Blob {
...
+
+        // Verify the size
+        let inner = list_array
+            .values()
+            .as_any()
+            .downcast_ref::<arrow::array::PrimitiveArray<arrow::datatypes::UInt8Type>>()
+            .expect("Inner should be UInt8Array");
+
+        assert_eq!(inner.len(), 1920 * 1080 * 3);
+    }
```
Wumpf, line 195: let's drop this one. don't see how this is any different from the previous test, only adds noise

joelreymont:
Done!

Wumpf:
I think your rebase went sideways

joelreymont:
Should be good now!

Wumpf:
if you keep rebasing like that it's really hard to keep track of what it is we're actually landing - practically I have to do a full sweep over the code again to make sure nothing strange is going on. Once a review is out it's better to a merge since they're a lot easier to incrementally diff

Also, needs CI fixes.

joelreymont:
Duly noted, thank you!

Do you make use of the `Compare` button to review the changes after a force-push? I find it really helpful!

CleanShot 2025-11-12 at 12 39 55@2x

See the lower right corner of the pic above.

Wumpf:
I usually use the "changes since last review" feature so I get everything relevant, not just a single push

crates/build/re_types_builder/src/codegen/rust/serializer.rs
```diff
@@ -765,15 +765,52 @@ fn quote_arrow_field_serializer(
...
+                                }
+                            }
+                        } else {
+                            // Standard path for other ScalarBuffer types
+                            quote! {
+                                #data_src
                                 .iter()
                                 #flatten_if_needed
-                                .map(|b| b as &[_])
+                                .map(|b| b.as_ref() as &[_])
```
Wumpf, line 810: why did this change now?

joelreymont: I updated both paths to use explicit `.as_ref()` for consistency, as it makes the conversion more explicit (not relying on `deref` coercion).

Both versions are functionally identical since `&ScalarBuffer<T>` implements `Deref<Target=[T]>`.

joelreymont:
Should be good now!

PR #11844, opened 2025-11-10
Create a proper raw mesh example (#1957)

### LINKED ISSUE #1957: raw_mesh example is misleading in that it loads gltf files
This is confusing in that far that we can load gltf/glb files natively. The raw_mesh example (both in Python and Rust) should load something we don't support out of the box, or even better a generated mesh.

joelreymont:
Python implementation:
- Replaced GLTF/trimesh loading with procedural primitive generation
- Added functions to generate: cube, pyramid, sphere, icosahedron, grid
- Removed trimesh and requests dependencies
- Removed download_dataset.py (no longer needed)
- Updated README to reflect procedural generation approach
- Added command-line options for sphere subdivisions and grid size

Rust implementation:
- Replaced GLTF parsing with procedural primitive generation
- Added MeshPrimitive struct with Clone derive
- Implemented same geometric primitives as Python version
- Removed gltf and bytes dependencies
- Updated README with new usage instructions
- Added command-line options matching Python version

Geometric primitives demonstrate:
- Colored Cube: per-vertex colors
- Textured Pyramid: UV coordinates with checkerboard texture
- Smooth Sphere: vertex normals for smooth shading
- Icosahedron: flat shading (no normals)
- Grid instances: transform hierarchy with translation, rotation, scale

Closes #1957


grtlr (changes requested):
Overall I think these changes are valuable, because they make the example more self-contained, so thank you very much for opening a PR!

There are a couple of things I think need to change though:

* Create a single mesh and then re-use that with a different subsets of parameters (`albedo_factor`, `albedo_texture`, …) to reduce the amount of geometry code in the example.
* I'm not really sure what the benefits of the _Grid Meshes_ are, they use the same `Transform3D` archetype as the other logging calls. That being said, I don't think we need to cram instancing into this example.

joelreymont:
Please check!

examples/python/raw_mesh/raw_mesh/__main__.py
```diff
@@ -2,138 +2,154 @@
...
-import trimesh
-
-from .download_dataset import AVAILABLE_MESHES, ensure_mesh_downloaded
 
 DESCRIPTION = """
 # Raw meshes
-This example shows how you can log a hierarchical 3D mesh, including its transform hierarchy.
+This example shows how to construct and log 3D meshes programmatically from scratch, including their transform hierarchy.
+
+It demonstrates how to create vertices, normals, colors, texture coordinates, and materials for various geometric primitives.
```
grtlr: We now only have a single geometric primitive (a sphere), right?

grtlr (changes requested):

grtlr:
@rerun-bot full-check

examples/python/raw_mesh/raw_mesh/__main__.py
joelreymont: Please check again!

grtlr:
Your recent force pushes seem to have overwritten the requested changed that you previously committed.

PR #11845, opened 2025-11-10
Removing time selection with context menu

IsseW:
### Related

- Part of RR-2768

### What

Adds functionality to remove an time selection either via context menu, or alt+click.

Screenshot 2025-11-10 at 18 55 31
Screenshot 2025-11-10 at 18 55 39
Screenshot 2025-11-10 at 18 55 52

https://github.com/user-attachments/assets/3cabd74c-8227-4e9d-bd42-fdff36f0466e

alt+click also works as context menu suggests

### DIFF
--- a/crates/viewer/re_time_panel/src/time_selection_ui.rs
+++ b/crates/viewer/re_time_panel/src/time_selection_ui.rs
@@ -6,7 +6,8 @@ use re_log_types::{
 use re_types::blueprint::components::LoopMode;
 use re_ui::{HasDesignTokens as _, UICommand, UICommandSender as _, UiExt as _, list_item};
 use re_viewer_context::{
-    SystemCommandSender, TimeControl, TimeControlCommand, ViewerContext, open_url::ViewerOpenUrl,
+    SystemCommandSender as _, TimeControl, TimeControlCommand, ViewerContext,
+    open_url::ViewerOpenUrl,
 };
 
 use super::time_ranges_ui::TimeRangesUi;
@@ -311,12 +312,12 @@ fn selection_context_menu(
             is_on_selection && copy_command.is_ok() && has_time_range,
             egui::Button::new("Copy link to trimmed time range"),
         )
-        .on_disabled_hover_text(if copy_command.is_err() {
-            "Can't share links to the current recording"
+        .on_disabled_hover_text(if let Err(err) = copy_command.as_ref() {
+            format!("Can't share links to the current recording: {err}")
         } else if !has_time_range {
-            "The current recording doesn't support time range links"
+            "The current recording doesn't support time range links".to_owned()
         } else {
-            "Open the context menu on selected time to copy link"
+            "Open the context menu on selected time to copy link".to_owned()
         })
         .clicked()
         && let Ok(copy_command) = copy_command


gavrelina:
nitty-gritty! 
according to the design the idea is to only have `remove` time selection option when clicking on the top of the timeline. 
and have it "dimmed" (deselected) when right mouse click outside of the time selection, but still on the top of the time panel.

so basically the mental model is: 
everything you could do with the time selection should be on top of the timepanel (remove, copy link, save (out of scope for this one))
and on the timepanel itself only actions related to single timestamp

crates/viewer/re_time_panel/src/time_selection_ui.rs
```diff
@@ -163,6 +165,15 @@ pub fn loop_selection_ui(
                         );
                     });
 
+                if let Some(pointer_pos) = pointer_pos
+                    && let Some(time) = time_ranges_ui.snapped_time_from_x(ui, pointer_pos.x)
+                {
+                    middle_response.context_menu(|ui| {
+                        let enabled = true;
+                        selection_context_menu(ui, ctx, time_commands, time_ctrl, time, enabled);
```
emilk: Why does the selection context menu need the time under the cursor? 🤔

crates/viewer/re_time_panel/src/time_selection_ui.rs
```diff
@@ -163,6 +165,15 @@ pub fn loop_selection_ui(
                         );
                     });
 
+                if let Some(pointer_pos) = pointer_pos
+                    && let Some(time) = time_ranges_ui.snapped_time_from_x(ui, pointer_pos.x)
+                {
+                    middle_response.context_menu(|ui| {
+                        let enabled = true;
```
emilk: hard to tell from reading this _what_ is enabled

crates/viewer/re_time_panel/src/time_selection_ui.rs
```diff
@@ -163,6 +165,15 @@ pub fn loop_selection_ui(
                         );
                     });
 
+                if let Some(pointer_pos) = pointer_pos
+                    && let Some(time) = time_ranges_ui.snapped_time_from_x(ui, pointer_pos.x)
+                {
+                    middle_response.context_menu(|ui| {
```
emilk: so this is the context menu for right-clicking the loop selection in the timeline…

crates/viewer/re_time_panel/src/time_selection_ui.rs
```diff
@@ -230,32 +245,80 @@ pub fn loop_selection_ui(
         }
     }
 
+    let right_clicked_time_id = egui::Id::new("__right_clicked_time_selection");
```
emilk: …but what is this? Right-clicking the events canvas is already handled elsewhere. Is this right-clicking on the timeline but outside of the loop selection? If so, you should just use `timeline_response.context_menu`

crates/viewer/re_viewer_context/src/time_control.rs
```diff
@@ -501,13 +502,22 @@ impl TimeControl {
...
+            let bp_loop_section = blueprint_ctx.time_selection();
+            if old_timeline != timeline {
+                match state.loop_selection {
+                    Some(selection) => blueprint_ctx.set_time_selection(selection.to_int()),
+                    None => {
+                        blueprint_ctx.clear_time_selection();
+                    }
+                }
+            } else {
+                state.loop_selection = bp_loop_section.map(|r| r.into());
```
emilk, line 521: What is going on here, and why?

crates/viewer/re_viewer_context/src/time_control.rs
```diff
@@ -816,9 +826,14 @@ impl TimeControl {
...
-                    && !self.states.contains_key(timeline_name)
-                {
+                if let Some(state) = self.states.get(timeline_name) {
+                    if let Some(blueprint_ctx) = blueprint_ctx {
+                        match state.loop_selection {
+                            Some(selection) => blueprint_ctx.set_time_selection(selection.to_int()),
+                            None => blueprint_ctx.clear_time_selection(),
+                        }
+                    }
+                } else if let Some(full_valid_range) = self.full_valid_range(times_per_timeline) {
```
emilk, line 838: What is happening here, and why?

crates/viewer/re_time_panel/src/time_selection_ui.rs
IsseW: Because it has the other option to copy link, and copy timestamp as well

crates/viewer/re_time_panel/src/time_selection_ui.rs
IsseW: Yes that's basically what I'm doing, but I want to use the `is_open` to more easily track the memory of where to draw the line for the context menu, and making sure it stays open even when the cursor leaves that upper timeline

IsseW:
> 🚨 I can no longer select time by dragging the timeline on the native app, but I can in the web viewer??

Very weird... Is that still the case?

crates/viewer/re_viewer_context/src/time_control.rs
IsseW: Added a comment to this

crates/viewer/re_viewer_context/src/time_control.rs
IsseW: Added a comment to this

crates/viewer/re_time_panel/src/time_selection_ui.rs
IsseW: Managed to remove this now

crates/viewer/re_time_panel/src/time_selection_ui.rs
IsseW: Removed that now with the new ui

emilk:
> > 🚨 I can no longer select time by dragging the timeline on the native app, but I can in the web viewer??
> 
> Very weird... Is that still the case?

Yes, still the case. Using a trackpad. It's very weird.

This is my blueprint for `dna`: [dna_blueprint.rbl.zip](https://github.com/user-attachments/files/23464423/dna_blueprint.rbl.zip)

EDIT: the default blueprint acts the same
EDIT2: I thought I had reset my blueprint, but I hadn't: https://rerunio.slack.com/archives/C08TEFXNHB2/p1762849099873379?thread_ts=1762847983.789149&cid=C08TEFXNHB2

emilk:
I'll try debugging it

crates/viewer/re_time_panel/src/time_selection_ui.rs
```diff
@@ -256,6 +272,59 @@ pub fn loop_selection_ui(
...
+    let mut url = ViewerOpenUrl::from_context(ctx);
+    let has_time_range = url.as_mut().is_ok_and(|url| url.fragment_mut().is_some());
+    let copy_command = url.and_then(|url| url.copy_url_command());
+    if ui
+        .add_enabled(
+            is_on_selection && copy_command.is_ok() && has_time_range,
+            egui::Button::new("Copy link to trimmed time range"),
+        )
+        .on_disabled_hover_text(if copy_command.is_err() {
+            "Can't share links to the current recording"
```
emilk: Could also use the error message in `copy_command`? 🤷

IsseW: Yeah good point, it's an anyhow which is not really ideal to create human readable stuff, because it could be anything. But I do add the message in the other place.

PR #11848, opened 2025-11-10
Add memory limit option to GrpcSink to prevent unbounded growth

### LINKED ISSUE #11519: GrpcSink Memory Usage Limit
**Is your feature request related to a problem? Please describe.**

When using `set_sinks()` with `GrpcSink()`, I'm seeing that my application memory continues to grow unbounded if there is no server for the sink to connect to.

**Describe the solution you'd like**

There should be a way to limit the amount of data that the `GrpcSink()` can buffer when it is failing to connect to a server.

**Describe alternatives you've considered**

**Additional context**

joelreymont:
Fixes #11519 by adding configurable memory limits to prevent unbounded buffer growth when the client cannot connect to a server.

Changes:
- Add max_pending_messages option (default: 10,000 messages)
- Use bounded channel when limit is set, unbounded otherwise
- Drop new messages with warning when buffer is full
- Add GrpcSink::new_with_options() for custom configuration
- Include automated tests verifying the fix

Current implementation is to drop newest messages:
* When buffer is full, new messages are discarded
* Old/stale messages remain in the buffer
* When connection resumes, old data is sent first
* Result: Viewer shows stale data, not current state

A better approach could be to drop oldest:
* When buffer is full, old messages are discarded
* New messages are kept
* When connection resumes, recent data is sent
* Result: Viewer shows current state

Since `GrpcSink` uses `ChunkBatcherConfig::LOW_LATENCY` and is designed for live streaming to a viewer, users probably care more about seeing the current state than preserving historical data when the connection is down.

Implementing drop-oldest properly is complex:
1. Tokio's bounded channel doesn't support _evict oldest_ from sender side
2. Would need custom queue implementation (VecDeque + Mutex + Condvar?)
3. Potential for subtle bugs and race conditions
4. Much more code to test and maintain

*Option A*: Accept current implementation
* Pros: Simple, solves memory growth, flush works
* Cons: Poor UX for live visualization use case

*Option B*: Implement drop-oldest
* Pros: Better UX, sends recent data when reconnected
* Cons: Complex, requires custom queue, more testing needed

I could implement _Option B_ as a follow-up PR or as part of this one, just let me know!

### DIFF
--- a/crates/store/re_grpc_client/src/write.rs
+++ b/crates/store/re_grpc_client/src/write.rs
@@ -126,7 +126,8 @@ pub struct Options {
     /// If we have not yet connected to the client, then
     /// do not block [`Client::flush_blocking`] for longer than this.
     ///
-    /// We will still retry connecting for however long it takes.
+    /// By default, we will retry connecting for however long it takes
+    /// (see [`Options::initial_connect_timeout`]).
     /// But blocking [`Client::flush_blocking`] forever when the
     /// server just isn't there is not a good idea.
     pub connect_timeout_on_flush: Duration,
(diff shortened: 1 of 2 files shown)


jprochazk (commented):
I believe we don't actually want to do this. It makes things a lot more complicated, and doesn't address the real problem, which is that we don't surface connection errors.

PR #11849, opened 2025-11-10
Add `av1` support to `VideoStream`

### LINKED ISSUE #10184: Av1 support for `VideoStream` archetype

oxkitsune:
### Related

* Closes #10184

### What

This adds `av1` support for the `VideoStream` archetype.

TODO:
- [x] Add tests 🤞 
- [x] Construct chroma subsampling from frame/sequence header
- [x] sanity check

### DIFF
--- a/crates/viewer/re_view_spatial/tests/video.rs
+++ b/crates/viewer/re_view_spatial/tests/video.rs
@@ -346,6 +346,7 @@ fn test_video_stream_codec_h265() {
 //     test_video(VideoType::VideoStream, VideoCodec::VP9);
 // }
 
+#[cfg(feature = "nasm")] // Need nasm for Av1 decoding on some platforms otherwise we error.
 #[test]
 fn test_video_stream_codec_av1() {
     test_video(VideoType::VideoStream, VideoCodec::AV1);


crates/utils/re_video/src/av1.rs
```diff
@@ -0,0 +1,86 @@
...
+                            dimensions =
+                                Some([header.frame_width as u16, header.frame_height as u16]);
+                        }
+                    }
+                    ParsedObu::FrameHeader(header) => {
+                        if is_gop_start(&header) {
+                            gop_found = true;
+                            dimensions =
+                                Some([header.frame_width as u16, header.frame_height as u16]);
+                        }
```
vivekmenezes, line 67: this code is repeated twice. Wanna encapsulate it in a single local function?

crates/utils/re_video/src/av1.rs
```diff
@@ -0,0 +1,86 @@
...
+        let action = parser
+            .read_obu(&data[offset..])
+            .map_err(DetectGopStartError::Av1ParserError)?;
+
+        match action {
+            ObuAction::Drop(num_bytes) => {
+                offset += num_bytes as usize;
+            }
+            ObuAction::Process(obu) => {
+                let bytes_used = obu.bytes_used;
```
vivekmenezes, line 46: no need to create a new var

offset += obu.bytes_used;

crates/utils/re_video/src/av1.rs
```diff
@@ -0,0 +1,86 @@
+use cros_codecs::codec::av1::parser::{FrameHeaderObu, FrameType, ObuAction, ParsedObu, Parser};
+
+use crate::{DetectGopStartError, GopStartDetection, VideoEncodingDetails};
+
+/// Try to determine whether an AV1 frame chunk is the start of a GOP.
+/// This is a simplified approach that only looks for keyframes, `INTRA_ONLY` frames are not considered GOP starts.
```
Wumpf: haven't looked at the spec but `INTRA_ONLY` also sounds like something we're not intersted in. here anyways :)
IIRC av1 has different types of GOPs and the one we're after are about decoder-rentrancy

crates/utils/re_video/src/av1.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    // If we found a GOP start, we should have dimensions either from the
+    // frame header or the sequence header.
+    let coded_dimensions = dimensions.unwrap_or_default();
+
+    Ok(GopStartDetection::StartOfGop(VideoEncodingDetails {
+        codec_string: "av01".to_owned(),
+        coded_dimensions,
+        bit_depth,
+        chroma_subsampling: None,
```
Wumpf, line 100: plz validate that, but I think we can comment here that the decoder will find this out on the fly
(that said, it would be _nice_ if we could already know ahead of time. mostly for UI purposes though 🤷 )

Wumpf (commented):
nice, this looks very promising

crates/utils/re_video/src/av1.rs
oxkitsune: I mention this specifically because AFAIK, `INTRA_ONLY` frames only do intra prediction, e.g. they only use information from that specific frame, no dependency on other frames. 
They also refresh reference frames, but it does not guarantee the result is valid unless `error_resilient_mode` is set. So if this frame or a future intra frame references data that's been refreshed or hasn't been decoded it can lead to artifacts.

So in theory we could use these as well. However, I think it is not worth opening pandora's box and I think you are correct in saying that we don't really care about this.

crates/utils/re_video/src/av1.rs
oxkitsune: I _think_ we can, it's marked as a todo on this PR, looking into it now.

crates/utils/re_video/src/av1.rs
Wumpf: makes sense, thanks for clarification

Wumpf:
I think there's some feature flag issue on when we enable av1 support on debug linux builds 🤔 
Since the impl is homogenous accross platforms I think it's fine to disable the test on Linux with a comment. But should check whether that's true and there's not something else up

crates/store/re_types/definitions/rerun/components/video_codec.fbs
```diff
@@ -37,10 +37,9 @@ enum VideoCodec: uint (
     /// AOMedia Video 1 (AV1)
     ///
     /// See <https://en.wikipedia.org/wiki/AV1>
```
Wumpf, line 39: We should write something about what the payload means here (something about whole OBUs?) similar to how we spec it the other formats.

crates/utils/re_video/src/av1.rs
```diff
@@ -0,0 +1,198 @@
...
+        // Subsampling in X only => 4:2:2
+        ChromaSubsamplingModes::Yuv422
+    } else {
+        // Subsampling in both X and Y => 4:2:0
+        ChromaSubsamplingModes::Yuv420
+    }
+}
+
+#[cfg(test)]
+mod test {
```
Wumpf, line 130: nice, these look great

crates/utils/re_video/src/av1.rs
```diff
@@ -0,0 +1,198 @@
...
+        let result = detect_av1_keyframe_start(&[]);
+        assert!(matches!(result, Ok(GopStartDetection::NotStartOfGop)));
+    }
+
+    #[test]
+    fn test_detect_av1_invalid_data() {
+        // Random invalid data, the parser will panic on invalid OBU structure.
+        // Make sure we handle that gracefully.
+        let invalid_data = &[0xFF, 0xFF, 0xFF, 0xFF, 0xFF];
+        let _result = detect_av1_keyframe_start(invalid_data);
```
Wumpf: shouldn't we check that the output is `Av1ParserError`?

crates/utils/re_video/src/av1.rs
```diff
@@ -0,0 +1,198 @@
...
+    #[test]
+    fn test_detect_av1_empty_data() {
+        let result = detect_av1_keyframe_start(&[]);
+        assert!(matches!(result, Ok(GopStartDetection::NotStartOfGop)));
+    }
+
+    #[test]
+    fn test_detect_av1_invalid_data() {
+        // Random invalid data, the parser will panic on invalid OBU structure.
+        // Make sure we handle that gracefully.
```
Wumpf:
```suggestion
        // Make sure we handle that gracefully and don't forward to the parser.
```

confused me because for me the whole thing was the parser

crates/utils/re_video/src/av1.rs
```diff
@@ -0,0 +1,198 @@
...
+        let invalid_data = &[0xFF, 0xFF, 0xFF, 0xFF, 0xFF];
+        let _result = detect_av1_keyframe_start(invalid_data);
+    }
+
+    #[test]
+    fn test_detect_av1_non_keyframe() {
+        // This would need a valid AV1 sequence with a non-keyframe
+        // For now, we just test that invalid/incomplete data doesn't panic
```
Wumpf, line 183: hm yeah getting good minimal data is hard. Makes me sad though. Well I figure our viewer based integration test has that covered

crates/utils/re_video/src/av1.rs
oxkitsune: good catch

PR #11850, opened 2025-11-11
add an ability to enrich `tracestate` from an env var

zehiko:
### What

This PR changes ``re_perf_telemetry`` and allows any consumer of this library to enrich the ``tracestate`` header (on the outbound path) by setting the relevant env var. This will be propagated e2e.

E2e trace context propagation continues to work as expected.

### Testing
* few simple unit tests
* e2e tests

Screenshot 2025-11-11 at 09 44 34

### DIFF
--- a/crates/utils/re_perf_telemetry/src/lib.rs
+++ b/crates/utils/re_perf_telemetry/src/lib.rs
@@ -131,7 +131,9 @@ pub fn current_trace_headers() -> Option<TraceHeaders> {
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct TraceHeaders {
     pub traceparent: String,
-    pub tracestate: Option<String>,
+    #[serde(skip)]
+    pub tracestate: HashMap<String, String>,
+    tracestate_header_raw: Option<String>,
 }
 
 impl TraceHeaders {
@@ -141,16 +143,8 @@ impl TraceHeaders {
     fn empty() -> Self {
         Self {
             traceparent: String::new(),
-            tracestate: None,
-        }
-    }
-
-    pub fn tracestate(&self) -> anyhow::Result<HashMap<String, String>> {
-        if let Some(tracestate_str) = &self.tracestate {
-            crate::tracestate::parse_pairs(tracestate_str)
-                .map(|pairs| pairs.into_iter().collect::<HashMap<String, String>>())
-        } else {
-            Ok(HashMap::default())
+            tracestate: HashMap::default(),
+            tracestate_header_raw: None,
         }
     }
 }
@@ -161,7 +155,9 @@ impl opentelemetry::propagation::Injector for TraceHeaders {
             Self::TRACEPARENT_KEY => self.traceparent = value,
             Self::TRACESTATE_KEY => {
                 if !value.is_empty() {
-                    self.tracestate = Some(value);
+                    // Parse tracestate, keeping only valid pairs
+                    self.tracestate = crate::tracestate::parse_pairs(&value);
+                    self.tracestate_header_raw = Some(value);
                 }
             }
             _ => {}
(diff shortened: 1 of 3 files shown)


crates/utils/re_perf_telemetry/src/args.rs
```diff
@@ -235,6 +235,19 @@ pub struct TelemetryArgs {
...
+    /// Expects a comma-separated string of key=value pairs, e.g. `bench_id=my_bench,env=prod`.
+    /// These will be added to the W3C tracestate header for distributed tracing.
+    ///
+    /// This is useful for passing application-specific context that should propagate
+    /// across service boundaries.
+    ///
+    /// Keys must conform to the W3C tracestate spec: lowercase letters, digits,
+    /// underscores, dashes, asterisks, and forward slashes only.
+    #[clap(long, env = "OTEL_PROPAGATORS_TRACESTATE", default_value = "")]
+    pub tracestate: String,
```
andrea-reale, line 249: is this intended to be used by clients  only? 

[the fact that we use this for both client and server confuses me. What happens if this is set for example on the server as an env var]?

crates/utils/re_perf_telemetry/src/lib.rs
```diff
@@ -141,6 +144,15 @@ impl TraceHeaders {
             tracestate: None,
         }
     }
+
+    pub fn tracestate(&self) -> anyhow::Result<HashMap<String, String>> {
+        if let Some(tracestate_str) = &self.tracestate {
+            crate::tracestate::parse_pairs(tracestate_str)
+                .map(|pairs| pairs.into_iter().collect::<HashMap<String, String>>())
```
andrea-reale: why not replace `Option<String>` with `Option<HashMap<String, String>>` or (maybe?) just `HashMap<String, String>`?

crates/utils/re_perf_telemetry/src/telemetry.rs
```diff
@@ -326,10 +327,22 @@ impl Telemetry {
...
-            opentelemetry::global::set_text_map_propagator(
+            // This will be used by the `TracingInjectorInterceptor` to encode the trace information into the request headers.
+            // Additional `tracestate` can be added through the relevant env var and the custom enricher below.
+            let mut propagators: Vec<
+                Box<dyn opentelemetry::propagation::TextMapPropagator + Send + Sync>,
+            > = vec![Box::new(
                 opentelemetry_sdk::propagation::TraceContextPropagator::new(),
+            )];
+
+            if !tracestate.is_empty() {
```
andrea-reale, line 338: what happens if a "downstream" request already has a `tracestate`? woudl we override?

andrea-reale: nvm, I think I got my answer below

crates/utils/re_perf_telemetry/src/args.rs
zehiko: you will just enrich the context further i.e. both should be respected. I thought it might be useful, but also didn't want to add the distinction in the code.

crates/utils/re_perf_telemetry/src/lib.rs
zehiko: you mean changing the actual field? yeah, that makes sense, let me change that

crates/utils/re_perf_telemetry/src/args.rs
zehiko: reading about it bit more, it seems that it's common for each service in the path to add additional context via tracestate, so if you don't mind I'd keep the support for both sides?

andrea-reale: but it's very odd that it's static / or via an env var server side? (It's somehow odd client side too, but at least we have a use-case for it)

zehiko: I can think of a use case where server's env vars are calculated on container init (hostname, region, az, etc) and then injected into the context. But in any case, we can make this only allowed on the client side if that makes more sense.

PR #11851, opened 2025-11-11
make sure we have the version in the release job

IsseW:
### What

Currently the release job doesn't get the "version" variable, this fixes that by making the tag and release different steps in the same job,

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -395,6 +395,7 @@ jobs:
           gh release create $version --verify-tag --draft --title $version $pre_arg
 
       - name: Create comment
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
abey79, line 411: this thing used to run only for rc and final, now it runs all the time. Is that intentional?

abey79: if not i guess you need to addd the same `if` as the previous step...

IsseW: Yeah need to copy the if there, good catch

Wumpf: nice, totally missed that!

Issue #11852, opened 2025-11-11
Alternate catalog doesn't work with registered table


abey79:
While working on the e2e test suite, I noticed that accessing the `"alternate_catalog"` fails when the table as created using `register_table`, while it works when the table is pre-populated in the OSS server.

cc @timsaucer

PR #11853, opened 2025-11-11
Make the e2e redap test suite more idiomatic

abey79:
### Related

- uncovered https://github.com/rerun-io/rerun/issues/11852

### What

Title. The goals are:
- make the fixtures more composable
- establish idiomatic patterns for future inspiration
- make it easier, in the future, to point the test suite at any redap implementation

### DIFF
--- a/rerun_py/tests/e2e_redap_tests/conftest.py
+++ b/rerun_py/tests/e2e_redap_tests/conftest.py
@@ -58,8 +58,15 @@ def table_filepath(tmp_path_factory: pytest.TempPathFactory) -> Generator[pathli
     yield temp_dir / "simple_datatypes"
 
 
+# TODO(ab): make this fixture configurable, such that the entire test suite can be run against different servers
 @pytest.fixture(scope="function")
 def catalog_client() -> Generator[CatalogClient, None, None]:
+    """
+    Return a `CatalogClient` instance connected to a test server.
+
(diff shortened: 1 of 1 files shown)


rerun_py/tests/e2e_redap_tests/conftest.py
```diff
@@ -30,68 +30,84 @@
...
-        # Copy all test data to the temp directory
-        shutil.copytree(TABLE_FILEPATH, temp_path / "simple_datatypes")
+    temp_dir = tmp_path_factory.mktemp("table_filepath")
+    shutil.copytree(TABLE_FILEPATH, temp_dir / "simple_datatypes")
+    yield temp_dir / "simple_datatypes"
 
-        # Yield the path to the copied data
-        yield temp_path / "simple_datatypes"
 
+# TODO(ab): make this fixture configurable, such that the entire test suite can be run against different servers
```
ntjohnson1, line 61: I'm confused by this note. The python server object picks a random port. And the function scope means this is built/torn down for every test that requests it. So that means the entire test suite runs with their own server.

abey79: The idea is to be able to point this test suite to eg a cloud local build or even a deployed version, including in the `dataplatform` CI.

ntjohnson1: Ahh different KINDS of servers not instances of servers. Got it

PR #11858, opened 2025-11-11
Don't smooth velocity after WASD input

IsseW:
### Related

- Follow-up from #11788
- Closes https://github.com/rerun-io/rerun/issues/11867

### What

Stops smoothing down velocity after moving with WASD.

This wasn't written to the blueprint so caused rubber banding that's visible on high speeds.


IsseW:
https://rerun.io/viewer?url=https%3A%2F%2Fapp.rerun.io%2Fversion%2Fnightly%2Fexamples%2Fnuscenes_dataset.rrd also jumps around a lot, and I'm pretty sure we are doing the same smoothing as that.

The velocity I set to zero here only affects WASDQE input. Will add a comment though!

PR #11859, opened 2025-11-11
Rename `SpaceCamera` and make its pinhole non-optional

Wumpf:
Minor refactor, just some stuff I couldn't unsee while looking over @IsseW 's camera blueprint refactor and now @MichaelGrupp 's orthographic upcoming orthographic mode.


crates/viewer/re_view_spatial/src/pinhole_wrapper.rs
```diff
@@ -24,13 +24,13 @@ pub struct SpaceCamera3D {
     // -------------------------
     // Optional projection-related things:
```
IsseW: I'm guessing this comment should be updated as it isn't optional anymore?

crates/viewer/re_view_spatial/src/ui_3d.rs
```diff
@@ -591,24 +591,22 @@ fn show_orbit_eye_center(
 
 fn show_projections_from_2d_space(
     line_builder: &mut re_renderer::LineDrawableBuilder<'_>,
-    space_cameras: &[SpaceCamera3D],
+    space_cameras: &[PinholeWrapper],
```
IsseW: Could change this to `cameras` like you did in other places

Wumpf: oops missed one

PR #11861, opened 2025-11-11
Fix pinhole color fallback showing the wrong color

MichaelGrupp:
Instead of the generic color fallback, we have to use the `frustum_color` from the design tokens. On top of that, this fixes a bug where the fallback wasn't actually used (instead some previous default).


crates/viewer/re_view_spatial/src/visualizers/cameras.rs
```diff
@@ -41,29 +41,28 @@ impl IdentifiedViewSystem for CamerasVisualizer {
 }
 
 struct CameraComponentDataWithFallbacks {
-    pinhole: crate::Pinhole,
+    image_from_camera: glam::Mat3,
+    resolution: glam::Vec2,
+    color: egui::Color32,
+    line_width: re_renderer::Size,
```
MichaelGrupp, line 47: Changed this to make it a bit more ergonomic to use.

MichaelGrupp:
This here shows the bug. The fallback showed a wrong color, although the pinhole was actually in the right color. Only when using the color picker with that wrong fallback, it got applied. Both doesn't make sense.

https://github.com/user-attachments/assets/3400e9b5-9670-484b-b417-76b391c6af7f

IsseW (approved):
Nice fix!

PR #11862, opened 2025-11-11
Remove frame lag when creating loop region

emilk:
This was already very noticeable at 60 Hz. My bad for only using 120Hz+ in my day-to-day

### DIFF
--- a/crates/viewer/re_time_panel/src/time_selection_ui.rs
+++ b/crates/viewer/re_time_panel/src/time_selection_ui.rs
@@ -1,4 +1,4 @@
-use egui::{Color32, CursorIcon, Id, NumExt as _, Rect};
+use egui::{Color32, CursorIcon, Id, NumExt as _, Rangef, Rect};
 
 use re_log_types::{
     AbsoluteTimeRange, AbsoluteTimeRangeF, Duration, TimeInt, TimeReal, TimeType, TimestampFormat,
@@ -64,8 +64,6 @@ pub fn loop_selection_ui(
     timeline_rect: &Rect,
     time_commands: &mut Vec<TimeControlCommand>,
 ) {
-    let tokens = ui.tokens();
-
     if time_ctrl.loop_selection().is_none() && time_ctrl.loop_mode() == LoopMode::Selection {
         // Helpfully select a time slice
         if let Some(selection) = initial_time_selection(time_ranges_ui, time_ctrl.time_type()) {
@@ -76,8 +74,6 @@ pub fn loop_selection_ui(
     if time_ctrl.loop_selection().is_none() && time_ctrl.loop_mode() == LoopMode::Selection {
         time_commands.push(TimeControlCommand::SetLoopMode(LoopMode::Off));
     }
-
-    let is_active = time_ctrl.loop_mode() == LoopMode::Selection;
 
     let pointer_pos = ui.input(|i| i.pointer.hover_pos());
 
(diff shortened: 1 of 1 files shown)


crates/viewer/re_time_panel/src/time_selection_ui.rs
```diff
@@ -271,6 +239,77 @@ pub fn loop_selection_ui(
...
+        .iter()
+        .rev()
+        .find_map(|c| {
+            if let TimeControlCommand::SetLoopSelection(range) = c {
+                Some(AbsoluteTimeRangeF::from(*range))
+            } else {
+                None
+            }
+        })
+        .or(time_ctrl.loop_selection())?;
```
Wumpf, line 272: I wish there was a nicer way to query this out, feels hacky :/

crates/viewer/re_time_panel/src/time_selection_ui.rs
```diff
@@ -271,6 +239,77 @@ pub fn loop_selection_ui(
...
+    ui: &egui::Ui,
+    top_y_range: Rangef,
+    full_y_range: Rangef,
+    time_commands: &[TimeControlCommand],
+) -> Option<()> {
+    // Use latest range to avoid frame delay
+    let selected_range = time_commands
+        .iter()
+        .rev()
+        .find_map(|c| {
```
IsseW, line 265: That's a neat way to do this!

crates/viewer/re_time_panel/src/time_selection_ui.rs
```diff
@@ -271,6 +239,77 @@ pub fn loop_selection_ui(
...
+
+    // Paint selection:
+    let tokens = ui.tokens();
+    let corner_radius = tokens.normal_corner_radius();
+    let corner_radius = egui::CornerRadius {
+        nw: corner_radius,
+        ne: corner_radius,
+        sw: 0,
+        se: 0,
+    };
```
Wumpf, line 296: pre-existing, but personally I found the corner radius here a bit strange

Image

Image

🤷 no strong opinion on it, but can't unsee now

crates/viewer/re_time_panel/src/time_selection_ui.rs
```diff
@@ -271,6 +239,77 @@ pub fn loop_selection_ui(
...
+    top_y_range: Rangef,
+    full_y_range: Rangef,
+    time_commands: &[TimeControlCommand],
+) -> Option<()> {
+    // Use latest range to avoid frame delay
+    let selected_range = time_commands
+        .iter()
+        .rev()
+        .find_map(|c| {
+            if let TimeControlCommand::SetLoopSelection(range) = c {
```
IsseW, line 266: Could possibly check for a `RemoveLoopSelection` here to also have that not be frame delayed. But not as important as setting it

IsseW: More of a thought, not something I think needs any actions

crates/viewer/re_time_panel/src/time_selection_ui.rs
IsseW: I don't think it's that hacky when you consider that `time_commands` will almost always be empty. Might be nice to have a function for this if similar pattern pops up more times though.

emilk: I think the nice way to re-arrange this at some point is to put the time commands into `ctx` and then have helpers on `ctx` to get the latest values, including pending commands. Maybe.

crates/viewer/re_time_panel/src/time_selection_ui.rs
emilk: True, but removal is much less sensitive to frame delays than dragging is, so I rather not add the complexity now

PR #11868, opened 2025-11-11
[auth] Handle failures to fresh credentials as being unauthenticated

jleibs:
### What

Trying to create connection client in python with an expired refresh token was causing authentication failures, even if connecting to a server that doesn't require auth.

```
RuntimeError: Internal: verifying credentials: status: 'Unknown error', self: "failed to refresh credentials: {\"error\":\"invalid_grant\",\"error_description\":\"Session has already ended.\"}"
```

This properly handles `SessionExpired` / `Custom` cases and maps them to `CredentialsProviderError`.

### DIFF
--- a/crates/store/re_redap_client/src/connection_registry.rs
+++ b/crates/store/re_redap_client/src/connection_registry.rs
@@ -1,7 +1,9 @@
 use std::collections::{HashMap, hash_map::Entry};
+use std::error::Error as _;
 use std::sync::Arc;
 
 use itertools::Itertools as _;
+use re_auth::credentials::CredentialsProviderError;
 use tokio::sync::RwLock;
 use tonic::Code;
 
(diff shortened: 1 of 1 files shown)


crates/store/re_redap_client/src/connection_registry.rs
```diff
@@ -332,7 +340,24 @@ impl ConnectionRegistryHandle {
...
+            Err(err) => {
+                if let Some(cred_error) = err.source().and_then(|s| {
+                    s.downcast_ref::<re_auth::credentials::CredentialsProviderError>()
+                }) {
+                    match cred_error {
+                        CredentialsProviderError::SessionExpired => Err(ApiError::credentials(
+                            ClientCredentialsError::SessionExpired,
+                            "session expired",
+                        )),
+                        CredentialsProviderError::Custom(_) => Err(ApiError::credentials(
```
jleibs, line 352: @jprochazk  I don't totally follow why this is Custom and not mapped to SessionExpired, but this still seems like preferred behavior.

jprochazk: It's past Jan's fault:

https://github.com/rerun-io/rerun/blob/eb4236f2e03dff8f3cb99b044fe913a5184a0335/crates/utils/re_auth/src/credentials.rs#L112-L113

PR #11869, opened 2025-11-12
URDF loader: Use global material if inline material is absent

oxkitsune:
### What

If a `visual`’s inline material does not specify a color or texture, use the matching global material instead (if any).

### DIFF
--- a/crates/store/re_data_loader/src/loader_urdf.rs
+++ b/crates/store/re_data_loader/src/loader_urdf.rs
@@ -495,7 +495,7 @@ fn log_link(
         let name = name.clone().unwrap_or_else(|| format!("visual_{i}"));
         let vis_entity = link_entity / EntityPathPart::new(name);
 
-        // Prefer inline defined material properties, if there's no properties fall back to global material.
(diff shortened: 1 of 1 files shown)


crates/store/re_data_loader/src/loader_urdf.rs
```diff
@@ -490,22 +495,19 @@ fn log_link(
         let name = name.clone().unwrap_or_else(|| format!("visual_{i}"));
         let vis_entity = link_entity / EntityPathPart::new(name);
 
-        // We need to look up the material by name, because the `Visuals::Material`
-        // only has a name, no color or texture.
-        let material = material
-            .as_ref()
-            .and_then(|m| urdf_tree.materials.get(&m.name).cloned());
+        // Prefer inline defined material properties, if there's no properties fall back to global material.
```
MichaelGrupp: Wording:
```suggestion
        // Prefer inline defined material properties if present, otherwise fall back to global material.
```

MichaelGrupp (approved):
nice!

PR #11871, opened 2025-11-12
Revise Lenses API and definitions to include scatter columns

grtlr:
### What

This PR revises the Lenses API and specification by introducing different types of lenses that are centered around the time columns:

* `OneToOne` lenses have the same amount of output rows as their input column.
* `Static` lenses don't have any time columns attached to them.
* `OneToMany` lenses produce multiple rows for any given row in their input column.

These kind of lenses are accessed in the `LensBuilder` via `columns`, `static_columns`, and `scatter_columns` methods, which create scopes that can be used to bundle columns together.

Although the concept of chunks is not exposed in the Lenses API, under the hood all columns (time or component) specified in a scope (`LensKind`) will end up in the same chunk.

### Context

The new scatter columns operating mode is required for "plural" MCAP messages that have multiple timestamps defined in a single message. An example for this is [`foxglove.FrameTransforms`](https://docs.foxglove.dev/docs/sdk/schemas/frame-transforms).

### DIFF
--- a/crates/top/re_sdk/src/lenses/ast.rs
+++ b/crates/top/re_sdk/src/lenses/ast.rs
@@ -164,7 +164,7 @@ impl Op {
 
     /// Ignores any input and returns a constant `ListArray`.
     ///
-    /// Commonly used with [`LensBuilder::add_static_component_column_entity`].
+    /// Commonly used with [`LensBuilder::output_static_columns`].
     /// When used in non-static columns this function will _not_ guarantee the correct amount of rows.
     pub fn constant(value: ListArray) -> Self {
         Self::func(move |_| Ok(value.clone()))
--- a/crates/top/re_sdk/src/lenses/builder.rs
+++ b/crates/top/re_sdk/src/lenses/builder.rs
@@ -197,7 +197,7 @@ impl ColumnsBuilder {
         self
     }
 
-    /// Builds a [`LensKind`], the `input` is passed for providing contextualized errors.
+    /// Builds a [`ast::LensKind`], the `input` is passed for providing contextualized errors.
     fn build(self, input: &ast::InputColumn) -> Result<ast::LensKind, LensError> {
         Ok(ast::LensKind::Columns(OneToOne {
             target_entity: self.target_entity,
@@ -246,7 +246,7 @@ impl StaticColumnsBuilder {
         self
     }
 
-    /// Builds a [`LensKind`], the `input` is passed for providing contextualized errors.
+    /// Builds a [`ast::LensKind`], the `input` is passed for providing contextualized errors.
     fn build(self, input: &ast::InputColumn) -> Result<ast::LensKind, LensError> {
         Ok(ast::LensKind::StaticColumns(Static {
             target_entity: self.target_entity,
@@ -320,7 +320,7 @@ impl ScatterColumnsBuilder {
         self
     }
 
-    /// Builds a [`LensKind`], the `input` is passed for providing contextualized errors.
+    /// Builds a [`ast::LensKind`], the `input` is passed for providing contextualized errors.
     fn build(self, input: &ast::InputColumn) -> Result<ast::LensKind, LensError> {
         Ok(ast::LensKind::ScatterColumns(OneToMany {
             target_entity: self.target_entity,


crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -4,49 +4,92 @@
...
+        timelines: Vec<TimeOutput>,
+    },
+
+    /// Each input row produces multiple output rows (1:N flat-map).
+    ///
+    /// Outputs are always temporal.
+    ToMany {
+        target_entity: TargetEntity,
+        components: Vec<ComponentOutput>,
+        timelines: Vec<TimeOutput>,
```
emilk: Does this mean `timelines` is never empty? Should we maybe use [`vec1`](https://crates.io/crates/vec1) for it?

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -4,49 +4,92 @@
...
+    },
+
+    /// Static lens: outputs have no timelines (timeless data).
+    ///
+    /// In many cases, static lenses will omit the input column entirely.
+    Static {
+        target_entity: TargetEntity,
+        components: Vec<ComponentOutput>,
+    },
```
emilk: style: usually I find it better to use 1-tuple enum variants that wrap named structs. It has several benefits:

* Better encapsulation
* Easier to refactor later
* You can add `impl` block on a specific enum variant
* Overall clearer code

In other words:

```rs
pub enum LensKind {
    OneToOne(OneToOneLens),
    …
}

struct OneToOneLens {
    target_entity: …
}
```

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -174,266 +217,270 @@ impl Lens {
...
-
-            let time_entry = time_builders.entry(entity_path).or_default();
+        self.outputs
+            .iter()
+            .filter_map(|output| match output {
+                LensKind::OneToOne {
+                    target_entity,
+                    components,
+                    timelines,
+                } => apply_one_to_one(chunk, column, target_entity, components, timelines),
```
emilk: Here's an example of where using structs for the variants would lead to cleaner code
```suggestion
                LensKind::OneToOne(one_to_one) => one_to_one.apply(chunk, column),
```

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -174,266 +217,270 @@ impl Lens {
...
+    components: &[ComponentOutput],
+) -> impl Iterator<Item = (ComponentDescriptor, ListArray)> {
+    components.iter().filter_map(
+        |output| match apply_ops(input.list_array.clone(), &output.ops) {
+            Ok(list_array) => Some((output.component_descr.clone(), list_array)),
+            Err(err) => {
+                re_log::error!(
+                    "Lens operations failed for component columns '{}': {err}",
+                    output.component_descr
+                );
```
emilk: Are we sure we should not abort the operation if this happens?

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -174,266 +217,270 @@ impl Lens {
...
+        let time_column = re_chunk::TimeColumn::new(
+            None,
+            Timeline::new(timeline_name, timeline_type),
+            time_vals.values().clone(),
+        );
+        Some((timeline_name, time_column))
+    } else {
+        re_log::error_once!(
+            "Output for timeline '{timeline_name}' must produce data type {}",
+            DataType::List(Arc::new(Field::new_list_field(DataType::Int64, false))),
```
emilk: yeeshus arrow-rs can be verbose sometimes…

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -174,266 +217,270 @@ impl Lens {
...
+        );
+        None
```
emilk: Again - are we sure we want to continue despite these errors?

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -174,266 +217,270 @@ impl Lens {
...
+        ChunkId::new(),
+        entity_path.clone(),
+        final_timelines,
+        output_component_columns.collect(),
+    )
+    .inspect_err(|err| {
+        re_log::error_once!("Failed to build lens output at entity path '{entity_path}': {err}");
+    })
+    .ok()
+}
```
emilk: Logging the error instead of returning it is very weird. Why not leave the choice to the caller?

Using `None` for error is unidiomatic

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -4,49 +4,92 @@
 //! we should not leak these elements into the public API. This allows us to
 //! evolve the definition of lenses over time, if requirements change.
```
emilk, line 5: If you feel strongly that it is the right thing to log errors and then continue, please document that design choice clearly here (including WHY), since it goes against most idiomatic Rust and our [`CODE_STYLE.md`](https://github.com/rerun-io/rerun/blob/main/CODE_STYLE.md)

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -174,266 +217,270 @@ impl Lens {
...
+        let end = window[1] as usize;
+        let count = end - start;
+
+        if reference_array.is_null(row_idx) || count == 0 {
+            // Null or empty list produces one output row
+            scatter_indices.push(row_idx as u32);
+        } else {
+            // Each element produces one output row
+            for _ in 0..count {
+                scatter_indices.push(row_idx as u32);
```
emilk: Why cast to usize only to cast back to u32? O.o

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -846,4 +893,281 @@ mod test {
         assert_eq!(my_timeline.times_raw()[3], 130);
         assert_eq!(my_timeline.times_raw()[4], 140);
     }
+
+    // Helper function to create test data: list of structs with {timestamp: i64, value: String}
+    fn create_test_struct_list() -> ListArray {
```
emilk: Since you log errors instead of returning them, I think these tests should add a hook-in to ensure that no errors were logged during their runtime

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -846,4 +893,281 @@ mod test {
...
+        struct_list_builder
+            .values()
+            .field_builder::<Int64Builder>(0)
+            .unwrap()
+            .append_value(timestamp_counter.next().unwrap());
+        struct_list_builder
+            .values()
+            .field_builder::<StringBuilder>(1)
+            .unwrap()
+            .append_value("three");
```
emilk: Is there anything we can do to make this code more readable, e.g. by creating some helpers in `re_arrow_utils`? 🤔

crates/top/re_sdk/src/lenses/builder.rs
```diff
@@ -0,0 +1,328 @@
...
+            component_descr,
+            ops: ops.into_iter().collect(),
+        });
+        self
+    }
+
+    fn build(self) -> ast::LensKind {
+        if self.components.is_empty() {
+            re_log::warn_once!(
+                "Lens output created without any components. At least one component is required."
```
emilk: If it is _required_, then we should return an error.
Logging a warning implies we can keep going: https://github.com/rerun-io/rerun/blob/main/CODE_STYLE.md#warning

crates/top/re_sdk/src/lenses/builder.rs
```diff
@@ -0,0 +1,328 @@
...
+            timeline_type,
+            ops: ops.into_iter().collect(),
+        });
+        self
+    }
+
+    fn build(self) -> ast::LensKind {
+        if self.components.is_empty() {
+            re_log::warn_once!(
+                "Lens scatter output created without any components. At least one component is required."
```
emilk: same

crates/utils/re_arrow_combinators/src/reshape.rs
```diff
@@ -277,3 +278,82 @@ impl Transform for StructToFixedList {
...
+                if len == 0 {
+                    // Preserve empty list as a single empty row
+                    new_validity.push(true);
+                    new_offsets.push(current_offset);
+                } else if let Some(inner_list) = inner_list_opt {
+                    // Nested case: scatter inner lists
+                    let inner_offsets = inner_list.value_offsets();
+
+                    for j in start..end {
+                        new_validity.push(!inner_list.is_null(j));
```
emilk:
```suggestion
                        new_validity.push(inner_list.is_valid(j));
```

crates/utils/re_arrow_combinators/src/reshape.rs
```diff
@@ -277,3 +278,82 @@ impl Transform for StructToFixedList {
...
+/// - `[[1, 2, 3], [4, 5]]` → `[[1], [2], [3], [4], [5]]` (each element becomes a row)
+/// - `[[1, 2], null, [], [3]]` → `[[1], [2], null, [], [3]]` (nulls and empties preserved)
+/// - `[[[1, 2], [3]], [[4, 5, 6]]]` → `[[1, 2], [3], [4, 5, 6]]` (flatten one level)
+pub struct Explode;
+
+impl Transform for Explode {
+    type Source = ListArray;
+    type Target = ListArray;
+
+    fn transform(&self, source: &Self::Source) -> Result<Self::Target, Error> {
```
emilk, line 298:
```suggestion
    fn transform(&self, source: &Self::Source) -> Result<Self::Target, Error> {
        re_tracing::profile_function!();
```

crates/utils/re_arrow_combinators/tests/snapshots/explode__primitives.snap
```diff
@@ -0,0 +1,21 @@
+---
+source: crates/utils/re_arrow_combinators/tests/explode.rs
+expression: "format!(\"{}\", DisplayRB(result))"
```
emilk, line 3: It would be a lot easier to check these tests if these snapshot tests included the input.

Alternatively, put the snapshot output into the test files with [inline snapshots](https://docs.rs/insta/latest/insta/#inline-snapshots)

crates/utils/re_arrow_combinators/tests/explode.rs
```diff
@@ -0,0 +1,90 @@
...
+        Some(vec![Some(1), Some(2), Some(3)]),
+        Some(vec![Some(4), Some(5)]),
+        Some(vec![Some(6)]),
+    ]);
+    println!("Input:\n{}", DisplayRB(input.clone()));
+
+    let explode = Explode;
+    let result = explode.transform(&input).unwrap();
+
+    insta::assert_snapshot!("primitives", format!("{}", DisplayRB(result)));
```
emilk, line 27: Consider using https://docs.rs/insta/latest/insta/#inline-snapshots

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -174,266 +217,270 @@ impl Lens {
...
+                    scattered_i64.values().clone(),
+                );
+                final_timelines.insert(*timeline_name, new_time_column);
+            }
+            Err(err) => {
+                re_log::error_once!(
+                    "Failed to replicate timeline '{}' for entity '{entity_path}': {err}",
+                    timeline_name
+                );
+                return None;
```
emilk: In some places you log errors and _continue_, but here you log an error and abort. Why not just return an `Err` instead?

crates/top/re_sdk/src/lenses/ast.rs
grtlr: Sorry the doc comment was confusing here. Time columns defined via Lenses are optional.

Regarding `vec1`: ~~Might make sense for component columns, but these structs would be backing a potential future config language so that would make them weird to use?~~ A just saw that `vec1` has a `serde` feature.

crates/top/re_sdk/src/lenses/ast.rs
grtlr: Agree. TIL: Rust analyzer even has a function for this:
image

crates/top/re_sdk/src/lenses/ast.rs
grtlr: Hmm, it would be great if the call site of Lenses could make that decision (a migration script should probably fail, logging SDKs or the Viewer should do their best effort). 

Let's sync on that during the walkthrough.

crates/top/re_sdk/src/lenses/ast.rs
grtlr: Amen

crates/top/re_sdk/src/lenses/ast.rs
grtlr: I'd love that, especially for those tests. I found this: https://docs.rs/arrow/latest/arrow/#serde-compatibility. Giving it a shot.

crates/top/re_sdk/src/lenses/ast.rs
```diff
@@ -149,299 +213,409 @@ impl Op {
...
+        let time_column = re_chunk::TimeColumn::new(
+            None,
+            Timeline::new(timeline_name, timeline_type),
+            time_vals.values().clone(),
+        );
+        Ok((timeline_name, time_column))
+    } else {
+        Err(LensError::InvalidTimeColumn {
+            timeline_name,
+            actual_type: format!("{}", list_array.values().data_type()),
```
emilk: Why not just store the datatype here, instead of converting to a string?

emilk: This is a more idiomatic way of creating a string

```suggestion
            actual_type: list_array.values().data_type().to_string()
```

crates/top/re_sdk/src/lenses/builder.rs
```diff
@@ -0,0 +1,335 @@
...
+
+use super::LensError;
+use super::Op;
+use super::ast;
+use super::ast::OneToMany;
+use super::ast::OneToOne;
+use super::ast::Static;
+
+/// Builder for lenses with support for multiple output modes.
+pub struct LensBuilder {
```
emilk, line 16: Usually a good idea to mark builders as `must_use`

```suggestion
#[must_use]
pub struct LensBuilder {
```

crates/top/re_sdk/src/lenses/builder.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    /// Builds a [`LensKind`], the `input` is passed for providing contextualized errors.
+    fn build(self, input: &ast::InputColumn) -> Result<ast::LensKind, LensError> {
+        Ok(ast::LensKind::Columns(OneToOne {
+            target_entity: self.target_entity,
+            components: self.components.try_into().map_err(|_err| {
+                LensError::MissingOutputComponent {
+                    // TODO(grtlr): This should not use debug formatting.
+                    input_filter: format!("{:?}", input.entity_path_filter),
```
emilk: Why use formatting at all here?

```suggestion
                    input_filter: input.entity_path_filter.clone(),
```

crates/utils/re_arrow_combinators/src/reshape.rs
```diff
@@ -277,3 +278,108 @@ impl Transform for StructToFixedList {
...
+/// Explodes a list by scattering each inner element to a separate row.
+///
+/// Takes a `List<T>` and returns a flattened `List<T>` where each inner element
+/// becomes its own row.
+///
+/// # Example
+///
+/// - `[[1, 2, 3], [4, 5]]` → `[[1], [2], [3], [4], [5]]` (each element becomes a row)
+/// - `[[1, 2], null, [], [3]]` → `[[1], [2], null, [], [3]]` (nulls and empties preserved)
+/// - `[[[1, 2], [3]], [[4, 5, 6]]]` → `[[1, 2], [3], [4, 5, 6]]` (flatten one level)
```
emilk, line 291: talked about this earlier, but this is pretty hard for my eyes to parse. I think it would be a lot easier if made super-verbose:


```
[
   [1, 2, 3],
   [4, 5],
]

→

[
  [1],
  [2],
  [3],
  [4],
  [5],
]
```

crates/top/re_sdk/Cargo.toml
```diff
@@ -84,8 +85,11 @@ webbrowser = { workspace = true, optional = true }
 libc.workspace = true
 
 [dev-dependencies]
+arrow = { workspace = true, features = ["json"] }
```
emilk: I think we should add this feature flag to the workspace `Cargo.toml`, so that we always compile arrow with the same features enabled, no matter which crate in the workspace we're compiling (just like we already do for datafusion!)

crates/utils/re_arrow_combinators/src/reshape.rs
grtlr: Other examples use the old annotation. If we polish the `re_arrow_combinators` crate in the future a bit more, we should do all of them at once.

PR #11872, opened 2025-11-12
Fix broken `delete_entries` and entries table not updated

abey79:
### What

I noticed that the entries table wasn't updated when dataset are created and deleted. This PR fixes this.

One thing leading to another, I also discovered that `delete_entry` was utterly broken:
- it would unconditionally attempt to delete a dataset (would fail to delete a table)
- it would not cleanup associated blueprint dataset

Both fixed too.

Note: new test checked to be passing with `dataplatform` main.


crates/store/re_server/src/store/in_memory_store.rs
```diff
@@ -287,6 +287,59 @@ impl InMemoryStore {
         self.update_entries_table()
     }
 
+    pub fn create_dataset(
```
IsseW, line 290: I feel like these probably should have doc comments? But not required here as this is mostly a moved function

PR #11873, opened 2025-11-12
Fix broken links from nightly

ntjohnson1:
### Related
https://github.com/rerun-io/rerun/actions/runs/19285496738/job/55145478542

### What
* Adds lychee to our pixi env so it can be run as `pixi run lychee`
   * Doesn't yet update the .github workflows to use this instead of from command line (it probably should but uncertain why lychee wasn't already in our env)
* Run nightly incantation locally and fix broken links


ntjohnson1:
Nightly run: https://github.com/rerun-io/rerun/actions/runs/19298544113

ntjohnson1:
Check links already passed on nightly

pixi.toml
```diff
@@ -563,6 +563,7 @@ flatbuffers = ">=23"
 gitignore-parser = ">=0.1.9"
 gitpython = ">=3.1.40"
 jinja2 = ">=3.1.3,<3.2" # For `build_screenshot_compare.py` and other utilities that build websites.
+lychee = "0.20.1.*"
```
Wumpf, line 566: ah, right I always used a local install but why would we!

PR #11874, opened 2025-11-12
Make getting started viewable on sidebar

ntjohnson1:
### Related
#11873

### What
* Adds pixi commands for the lychee install I added
* Also make the getting started side panel viewable but with the correct re-direct.
* I ran `pixi run link-check` and it looks like it is ok

## Before
Screenshot 2025-11-12 at 8 57 10 AM

## After
Screenshot 2025-11-12 at 9 49 35 AM


bllchmbrs (approved):

bllchmbrs:
Looks good, think you should go for it ASAP @ntjohnson1

PR #11876, opened 2025-11-12
Fix `dataset_ui_test` broken by the dataset name now appearing twice in the UI

abey79:
### Related

* triggered by #11872

### What

The test assumed that the dataset name appeared only once, which it did only because the OSS server's entry list was broken until #11872. Then it would also fail flakily since the entries table can take a while to refresh and appear in the UI.


abey79:
@rerun-bot full-check

Issue #11702, opened 2025-10-29
CompressedDepth decoding for re_mcap


makeecat:
**Is your feature request related to a problem? Please describe.**
CompressedDepth encoded with RVL does not work with mcap visualization with re_mcap

**Describe the solution you'd like**
Implement CompressedDepth decoding functionality in the [compressed_image.rs](https://github.com/rerun-io/rerun/blob/main/crates/utils/re_mcap/src/parsers/ros2msg/sensor_msgs/compressed_image.rs)

**Describe alternatives you've considered**
None

**Additional context**
I already have a working prototype; I would like to use this issue to discuss the scope of a potential PR.

Wumpf:
Related / overlapping with
* https://github.com/rerun-io/rerun/issues/9046

grtlr:
Yeah I think first having an `EncodedDepthImage` archetype (that mimics `EncodedImage`) like @oxkitsune proposed makes sense, from that the remaining changes would come more naturally.

makeecat:
I will start implementing the EncodedDepthImage archetype and prepare a PR. I will move the conversation to [#9046](https://github.com/rerun-io/rerun/issues/9046). Once we merge the EncodedDepthImage archetype we can come back to this issue and prepare a second PR

grtlr:
That would be awesome 🚀!

makeecat:
We are implementing the `re_mcap` support of RVL in one PR #11877, feel free to add comment.

PR #11877, opened 2025-11-12
Support `png` & `rvl` depth images via new `EncodedDepthImage` archetype + integration with viewer and `re_mcap`

makeecat:
### Related

* #11702
* #9046

### What

This is a draft PR to provide prototype for EncodedDepthImage. This PR needs to be split into sub-PRs once reviewed by @oxkitsune 

1. Implemented the EncodedDepthImage archetype
2. Introduced a new utility crate re_depth_compression to host rvl codecs implementation
3. Integrated EncodedDepthImage archetype into viewer
4. Supported the EncodedDepthImage archetype in re_mcap.

### DIFF
--- a/rerun_cpp/src/rerun/archetypes/encoded_depth_image.hpp
+++ b/rerun_cpp/src/rerun/archetypes/encoded_depth_image.hpp
@@ -58,8 +58,7 @@ namespace rerun::archetypes {
     ///
     ///     std::vector<uint8_t> bytes{
     ///         std::istreambuf_iterator<char>(file),
-    ///         std::istreambuf_iterator<char>()
-    ///     };
+    ///         std::istreambuf_iterator<char>()};
     ///     // Determine media type based on file extension
     ///     rerun::components::MediaType media_type;
     ///     if (depth_path.extension() == ".png") {
@@ -71,11 +70,9 @@ namespace rerun::archetypes {
     ///     rec.log(
     ///         "depth/encoded",
     ///         rerun::archetypes::EncodedDepthImage()
-    ///             .with_blob(
-    ///                 rerun::components::Blob(
-    ///                     rerun::Collection<uint8_t>::take_ownership(std::move(bytes))
-    ///                 )
-    ///             )
+    ///             .with_blob(rerun::components::Blob(
+    ///                 rerun::Collection<uint8_t>::take_ownership(std::move(bytes))
+    ///             ))
     ///             .with_media_type(media_type)
     ///             .with_meter(0.001f)
     ///     );


makeecat:
Hi @oxkitsune , I believe I am comfortable with the prototype, the proof concept works well with lint and my local mcap bag with RVL-encoded depth image. I am ready to submit a separate PR for EncodedDepthImage archetype. 

However, I am not very confident about my changes in `re_view_spatial` crate. I refactored the `crates/viewer/re_view_spatial/src/visualizers/depth_images.rs` to support `crates/viewer/re_view_spatial/src/visualizers/encoded_depth_images.rs` as a separate visualizer. I am looking for comments on that part.

If you are happy with the proof-of-concept demo and the design of the EncodedDepthImage archetype, I can prepare the first PR for EncodedDepthImage archetype.

oxkitsune:
This demo looks great, thanks for putting it together!

Go ahead and open the PR for the `EncodedDepthImage` archetype; I’ll leave any implementation notes there.

makeecat:
According to our discussion, we will not split this PR into small PRs.

There are a few decisions we need to make:
1. whether add mediatype
2. whether leave re_depth_compression as a crate or integrate it into exsting logic
fallback detection of different format for decoding if the data didn't provide type metadata
3. When RVL metadata disagrees with the logged ImageFormat, the code just logs a warning and still constructs an ImageInfo with the caller’s width/height.
4. I don't like the way I refactored the DepthImagevisualizer: `viewer/re_view_spatial/src/visualizers/depth_image.rs`, I would like to avoid changing too much of existing depth image visualizer, but I didn't come up with a solution. The reason I have to refactor it is to expose some components from the original depth image visualizer to support encoded depth image visualizer.

crates/utils/re_depth_compression/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+//! Utilities for decoding depth compression formats.
+
+pub mod ros_rvl;
+
+pub use ros_rvl::{
+    RvlDecodeError, RvlMetadata, decode_ros_rvl_f32, decode_ros_rvl_u16, parse_ros_rvl_metadata,
+};
```
oxkitsune: If the module's already `pub`, I think it's neater to avoid re-exporting. This means that in the future, when adding more formats, we can easily access it using `re_depth_compression::<format>`.

makeecat:
After discussing with @oxkitsune 
1. We plan to add support of `rvl` as a `media_type`
2. We plan to squash the logic of `re_depth_completion` crate into `re_viewer_context`.
3. We plan to support `MediaType::guess_from_data` to support fallback detection of RVL data
4. We need further discussion to avoid refactoring too much of `viewer/re_view_spatial/src/visualizers/depth_image.rs`

crates/utils/re_mcap/src/parsers/ros2msg/sensor_msgs/compressed_image.rs
```diff
@@ -12,19 +17,96 @@ use crate::parsers::{
...
+    ZeroResolution,
+
+    #[error("RVL image resolution would overflow")]
+    ResolutionOverflow,
+
+    #[error("RVL payload shorter than expected for resolution {width}x{height}")]
+    PayloadLengthMismatch { width: u32, height: u32 },
+}
+
+fn parse_ros_rvl_metadata(data: &[u8]) -> Result<RvlMetadata, RvlDecodeError> {
```
oxkitsune: IIUC this is the same metadata parsing from `crates/viewer/re_viewer_context/src/utils/depth.rs` right? Is there any way we can avoid duplicating this?

crates/viewer/re_viewer_context/src/cache/image_decode_cache.rs
```diff
@@ -119,6 +166,119 @@ fn decode_image(
...
+    let (buffer, decoded_format) = ImageBuffer::from_dynamic_image(dynamic_image)?;
+
+    if decoded_format.datatype() != format.datatype() {
+        return Err(ImageLoadError::DecodeError(format!(
+            "Encoded depth PNG datatype mismatch: blob is {:?}, expected {:?}",
+            decoded_format.datatype(),
+            format.datatype()
+        )));
+    }
+
```
oxkitsune, line 204: Would be a good place to validate that the `decoded_format` is actually single channel data. This now support RGB pngs

crates/store/re_sdk_types/src/components/media_type_ext.rs
```diff
@@ -169,11 +186,29 @@ impl MediaType {
...
+            // - RVL Payload: variable length compressed data
+
+            if buf.len() < 20 {
+                return false; // Minimum size check (12 + 8 bytes)
+            }
+
+            // Read width and height from bytes 12-15 and 16-19 respectively
+            let width = u32::from_le_bytes([buf[12], buf[13], buf[14], buf[15]]);
+            let height = u32::from_le_bytes([buf[16], buf[17], buf[18], buf[19]]);
+
```
oxkitsune, line 214: Given the structure given above, I think that this, in its current state, will lead to a lot of false positives.
I think we should also check that `depth_quant_a` and `depth_quant_b` are both valid floats with sane ranges (e.g. not `NaN` or `inf` and not absurdly large)


```suggestion
let depth_quant_a = f32::from_le_bytes([buf[4], buf[5], buf[6], buf[7]]);
let depth_quant_b = f32::from_le_bytes([buf[8], buf[9], buf[10], buf[11]]);

// Reject NaNs/infs
if !depth_quant_a.is_finite() || !depth_quant_b.is_finite() {
    return false;
}

// Reject insane magnitudes
if depth_quant_a <= 0.0 || depth_quant_a > 1e4 || depth_quant_b.abs() > 1e4 {
    return false;
}
```

crates/store/re_types/src/components/media_type_ext.rs
```diff
@@ -260,5 +297,79 @@ fn test_media_type_extension() {
     assert_eq!(MediaType::markdown().file_extension(), Some("md"));
     assert_eq!(MediaType::plain_text().file_extension(), Some("txt"));
     assert_eq!(MediaType::png().file_extension(), Some("png"));
+    assert_eq!(MediaType::rvl().file_extension(), Some("rvl"));
     assert_eq!(MediaType::stl().file_extension(), Some("stl"));
 }
+
+#[test]
+fn test_guess_from_data() {
```
oxkitsune: most of this test is for `MediaType::rvl()`, which I think is fine. No need to test the GLB/stl here in my opinion. Test can also be renamed to `test_guess_from_data_rvl`

crates/viewer/re_component_fallbacks/src/component_fallbacks.rs
```diff
@@ -300,6 +300,77 @@ pub fn archetype_field_fallbacks(registry: &mut FallbackProviderRegistry) {
...
+            let is_float_image = ctx
+                .recording()
+                .latest_at_component::<components::ImageFormat>(
+                    ctx.target_entity_path,
+                    ctx.query,
+                    archetypes::EncodedDepthImage::descriptor_format().component,
+                )
+                .is_some_and(|(_index, format)| format.is_float());
+
+            components::DepthMeter::from(if is_float_image { 1.0 } else { 1000.0 })
```
oxkitsune: I think that this behavior should be documented on the archetype field.

crates/utils/re_depth_compression/src/ros_rvl.rs
```diff
@@ -0,0 +1,369 @@
...
+        depth_quant_a,
+        depth_quant_b,
+        payload_offset,
+        num_pixels,
+    })
+}
+
+pub fn decode_ros_rvl_u16(data: &[u8], metadata: &RvlMetadata) -> Result<Vec<u16>, RvlDecodeError> {
+    let payload = metadata.payload(data)?;
+    let mut disparity = vec![0u16; metadata.num_pixels()];
```
oxkitsune, line 123:
```suggestion
    let mut disparity = Vec::with_capacity(metadata.num_pixels());
```

For `f32` you initialize the vec like this, and since `RvlDecoder::decode_into` explicitly writes zeros, we can avoid the allocation here.

crates/utils/re_rvl/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+            for value in output.iter_mut().skip(write_index).take(zeros) {
+                *value = 0;
+            }
+            write_index += zeros;
+            remaining -= zeros;
+
+            let nonzeros = self.decode_vle()? as usize;
+            if nonzeros > remaining {
+                return Err(RvlDecodeError::TooManyNonZeros);
+            }
```
oxkitsune, line 191: we should make sure this decoder always progresses forward. E.g. if `nonzeros` and `zeros` are both 0 (doesn't make sense and should never happen with valid data), we should raise an error and exit.

crates/store/re_sdk_types/definitions/rerun/archetypes/encoded_depth_image.fbs
```diff
@@ -0,0 +1,50 @@
...
+  // --- Required ---
+
+  /// The encoded depth payload (e.g. RVL, depth-PNG).
+  blob: rerun.components.Blob ("attr.rerun.component_required", order: 1000);
+
+  /// Image format (width, height, datatype).
+  ///
+  /// Unlike generic [archetypes.EncodedImage], most depth codecs do not carry full headers,
+  /// so width/height/datatype must be logged explicitly.
+  format: rerun.components.ImageFormat ("attr.rerun.component_required", order: 1100);
```
oxkitsune:
```suggestion
  /// Image format (width, height, datatype).
  ///
  /// Standard image formats like PNG or JPEG include this metadata in their headers,
  /// allowing [archetypes.EncodedImage] to be self-describing. Depth images, however,
  /// typically use headerless codecs, so this information must be provided explicitly.
  format: rerun.components.ImageFormat ("attr.rerun.component_required", order: 1100);
```

crates/store/re_types/src/archetypes/encoded_depth_image_ext.rs
```diff
@@ -0,0 +1,23 @@
+use crate::{
+    archetypes::EncodedDepthImage,
+    components::{Blob, ImageFormat, MediaType},
+    datatypes::ChannelDatatype,
+};
+
+impl EncodedDepthImage {
+    /// Construct from encoded bytes with explicit format metadata.
+    pub fn from_encoded_bytes(bytes: impl Into<Vec<u8>>, format: impl Into<ImageFormat>) -> Self {
```
oxkitsune:
```suggestion
    pub fn from_file_contents(bytes: impl Into<Vec<u8>>, format: impl Into<ImageFormat>) -> Self {
```

this is a bit more consistent with `EncodedImage`

crates/store/re_types/src/components/media_type_ext.rs
```diff
@@ -46,6 +46,14 @@ impl MediaType {
     /// <https://www.iana.org/assignments/media-types/model/stl>
     pub const STL: &'static str = "model/stl";
 
+    // -------------------------------------------------------
+    // Compressed Depth Data:
+
+    /// [RVL compressed depth](https://rgbd-data.org/): `application/rvl`.
```
oxkitsune: This website does not exist: https://rgbd-data.org/

crates/utils/re_rvl/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+        bytes.extend_from_slice(&depth_params.0.to_le_bytes());
+        bytes.extend_from_slice(&depth_params.1.to_le_bytes());
+        bytes.extend_from_slice(&dimensions[0].to_le_bytes());
+        bytes.extend_from_slice(&dimensions[1].to_le_bytes());
+        let compressed = encode_rvl(disparity);
+        bytes.extend_from_slice(&compressed);
+        bytes
+    }
+
+    fn encode_rvl(values: &[u16]) -> Vec<u8> {
```
oxkitsune, line 317: Since you essentially implement a full encoder here as well, maybe it makes sense to just bundle the decoder + encoder up in a separate rust crate, so that it can be used by others as well?

Just a suggestion as I was surprised to see an encoder here as well 😅

docs/snippets/all/archetypes/encoded_depth.png
oxkitsune, line 1: I think this file should be under `tests/assets/image`

oxkitsune (dismissed):

crates/viewer/re_view_spatial/src/visualizers/encoded_depth_image.rs
oxkitsune, line 1: as far as I can see, this is 99% the same as the `depth_images` visualizer. The main difference is the component identifiers (e.g. `DepthImage::meter()` vs `EncodedDepthImage::meter()`).

Can we move the implementation to a unified function that takes the component identifiers?

docs/snippets/all/archetypes/encoded_depth.png
makeecat: I leave it here because I saw ferris.png in the same folder

crates/utils/re_rvl/src/lib.rs
makeecat: refactored the parsing/encoding/decoding logic back to re_depth_compression crate

crates/store/re_types/src/components/media_type_ext.rs
makeecat: fixed, linked to the Microsoft RVL paper

crates/store/re_types/src/archetypes/encoded_depth_image_ext.rs
makeecat: done

crates/store/re_sdk_types/definitions/rerun/archetypes/encoded_depth_image.fbs
makeecat: done

crates/utils/re_rvl/src/lib.rs
makeecat: done

crates/utils/re_depth_compression/src/ros_rvl.rs
makeecat: I understood that you would like to avoid unecesary allocation. We can’t “avoid the allocation” for u16 because the decoder writes into the caller’s buffer; that buffer must be sized first. We should keep it this way, otherwise, it will fail the test of the current API design.

crates/store/re_types/src/components/media_type_ext.rs
makeecat: done

crates/store/re_sdk_types/src/components/media_type_ext.rs
makeecat: done

crates/viewer/re_viewer_context/src/cache/image_decode_cache.rs
makeecat: now we have enforced single-channel PNG

crates/viewer/re_component_fallbacks/src/component_fallbacks.rs
makeecat: done

crates/utils/re_depth_compression/src/lib.rs
makeecat: done

crates/utils/re_mcap/src/parsers/ros2msg/sensor_msgs/compressed_image.rs
makeecat: done, refactored into re_depth_compression

crates/utils/re_depth_compression/src/lib.rs
makeecat: done

crates/store/re_mcap/Cargo.toml
```diff
@@ -21,6 +21,7 @@ re_log_types.workspace = true
 re_ros_msg.workspace = true
 re_tracing.workspace = true
 re_sdk_types = { workspace = true, features = ["ecolor", "glam", "image", "video"] }
+re_depth_compression.workspace = true
```
Wumpf: should be sorted

crates/store/re_sdk_types/definitions/rerun/archetypes/encoded_depth_image.fbs
```diff
@@ -0,0 +1,53 @@
+namespace rerun.archetypes;
+
+/// A depth image encoded with a codec (e.g. RVL or PNG).
+///
+/// Rerun also supports uncompressed depth images with the [`archetypes.DepthImage`](https://rerun.io/docs/reference/types/archetypes/depth_image).
+///
+/// \example archetypes/encoded_depth_image
```
Wumpf: A Rerun employee should upload an image for this

crates/viewer/re_viewer_context/src/cache/image_decode_cache.rs
Wumpf: should probably rename this to `entry_encoded_color_image` to be consistent with `entry_encoded_image`

crates/viewer/re_viewer_context/src/cache/image_decode_cache.rs
```diff
@@ -57,36 +60,85 @@ impl ImageDecodeCache {
...
+            decode_color_image(
+                blob_row_id,
+                blob_component,
+                image_bytes,
+                media_type.as_str(),
+            )
+        })
+    }
+
+    pub fn entry_encoded_depth(
```
Wumpf, line 75: misses doc string

crates/utils/re_depth_compression/src/ros_rvl.rs
```diff
@@ -0,0 +1,375 @@
...
+    ValueOutOfRange { value: i32 },
+
+    #[error("RVL stream ended unexpectedly")]
+    UnexpectedEof,
+
+    #[error("RVL VLE value overflowed")]
+    ValueOverflow,
+}
+
+pub fn parse_ros_rvl_metadata(data: &[u8]) -> Result<RvlMetadata, RvlDecodeError> {
```
Wumpf: could use some doc string. I tried to read up on this ros specific (?) metadata but couldn't find anything

crates/utils/re_depth_compression/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+//! Depth compression utilities.
+
+pub mod ros_rvl;
```
Wumpf: the namespacing here seems rather unnecessary since `ros_rvl` is repeated everywhere. Should flatten it out (also I dislike `pub mod` in general because you never quite know what you're exporting ;))

crates/viewer/re_viewer_context/src/cache/image_decode_cache.rs
```diff
@@ -115,6 +167,135 @@ fn decode_image(
...
+        blob_component,
+        Blob::from(buffer),
+        format.0,
+        ImageKind::Depth,
+    ))
+}
+
+fn vec_into_bytes<T: Pod>(vec: &[T]) -> Vec<u8> {
+    bytemuck::cast_slice(vec).to_vec()
+}
```
Wumpf: the callsites should do direct vector conversion instead of doing a full memcpy to a new vec, use https://docs.rs/bytemuck/latest/bytemuck/allocation/fn.cast_vec.html

docs/snippets/all/archetypes/encoded_depth_image.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include <utility>
+#include <vector>
+
+namespace fs = std::filesystem;
+
+int main() {
+    const auto rec = rerun::RecordingStream("rerun_example_encoded_depth_image");
+    rec.spawn().exit_on_failure();
+
+    const auto depth_path = fs::path(__FILE__).parent_path() / "encoded_depth.png";
```
Wumpf: should pass in the filename via command line like we do in other samples with command line options. We can use that as a default ofc.

Applies to all variants of the sample

docs/snippets/all/archetypes/encoded_depth_image.cpp
```diff
@@ -0,0 +1,44 @@
+// Log an encoded depth image stored as a 16-bit PNG.
```
Wumpf: if (below comment) we make this sample more flexible, the comment needs to be adjusted

docs/snippets/all/archetypes/encoded_depth_image.cpp
```diff
@@ -0,0 +1,44 @@
...
+    if (!file) {
+        std::cerr << "Failed to open encoded depth image: " << depth_path << std::endl;
+        return 1;
+    }
+
+    std::vector<uint8_t> png_bytes{
+        std::istreambuf_iterator<char>(file),
+        std::istreambuf_iterator<char>()};
+
+    const rerun::WidthHeight resolution(64, 48);
```
Wumpf: hardcoded image size means this example is pretty much unusual with any other data

docs/snippets/src/snippets/encoded_depth.png
Wumpf, line 1: we usually don't push pngs to our snippet directory, I'd rather not start with that now

crates/viewer/re_view_spatial/src/visualizers/depth_images.rs
```diff
@@ -14,189 +15,193 @@ use re_viewer_context::{
...
-    fill_ratio: Option<FillRatio>,
-    colormap: Option<Colormap>,
-    value_range: Option<[f64; 2]>,
+pub(super) struct DepthImageComponentData {
+    pub(super) image: ImageInfo,
+    pub(super) depth_meter: Option<DepthMeter>,
+    pub(super) fill_ratio: Option<FillRatio>,
+    pub(super) colormap: Option<Colormap>,
+    pub(super) value_range: Option<[f64; 2]>,
 }
```
Wumpf: can just be pub since this doesn't get exported further. No need to complicate things with `pub(super)`

applies to all `pub(super)` in here

crates/viewer/re_view_spatial/src/visualizers/depth_images.rs
```diff
@@ -14,189 +15,193 @@ use re_viewer_context::{
...
+#[expect(clippy::too_many_arguments)]
+pub(super) fn process_depth_image_data(
+    data_store: &mut SpatialViewVisualizerData,
+    depth_cloud_entities: &mut DepthCloudEntities,
+    ctx: &QueryContext<'_>,
+    depth_clouds: &mut Vec<DepthCloud>,
+    ent_context: &SpatialSceneEntityContext<'_>,
+    transforms: &TransformTreeContext,
+    images: impl Iterator<Item = DepthImageComponentData>,
+    archetype_name: ArchetypeName,
```
Wumpf, line 61: query context should have archetyape name already afaik

crates/viewer/re_view_spatial/src/visualizers/encoded_depth_image.rs
```diff
@@ -0,0 +1,178 @@
...
+                                    EncodedDepthImage::descriptor_blob().component,
+                                    blob,
+                                    media_type.as_ref(),
+                                    &format,
+                                )
+                            },
+                        ) {
+                            Ok(image) => image,
+                            Err(err) => {
+                                re_log::warn_once!("Failed to decode EncodedDepthImage at path {entity_path}: {err}");
```
Wumpf: didn't exist when this PR started, but this should be a visualizer error now that goes on the output :)

crates/viewer/re_view_spatial/src/visualizers/depth_images.rs
```diff
@@ -334,6 +324,72 @@ impl VisualizerSystem for DepthImageVisualizer {
...
+
+    output.draw_data.push(PickableTexturedRect::to_draw_data(
+        ctx.viewer_ctx.render_ctx(),
+        &data.pickable_rects,
+    )?);
+
+    Ok(output)
+}
+
+pub(super) fn first_copied<T: Copy>(slice: Option<&[T]>) -> Option<T> {
```
Wumpf: can move to `mod.rs`

crates/viewer/re_view_spatial/src/visualizers/depth_images.rs
Wumpf, line 213: would be nice to separate out the visualizer from the visualization utilities for depth images. Makes the diff messier again ofc, so I'm kinda happy you didn't so far 😄

crates/viewer/re_view_spatial/src/visualizers/depth_images.rs
Wumpf: yes, but only as an option

crates/utils/re_rvl/src/lib.rs
```diff
@@ -132,12 +132,20 @@ pub fn decode_rvl_with_quantization(
...
+
     for value in disparity {
         if value == 0 {
             depth.push(f32::NAN);
-        } else {
+        } else if has_quantization {
             let quantized = value as f32;
             depth.push(metadata.depth_quant_a / (quantized - metadata.depth_quant_b));
+        } else {
+            depth.push(value as f32);
```
Wumpf, line 148: I guess in that case we could keep it as u16 just like in "the original PR"
But in the interest of expediency I guess this is good enough and less confusing ;)

PR #11878, opened 2025-11-12
Update catalog during registration

### LINKED ISSUE #11852: Alternate catalog doesn't work with registered table

timsaucer:
### Related

Closes #11852

### What

When new catalog providers are created, they are not automatically registered with the datafusion session context. This PR adds a check for new catalog providers when a table is created or registered.

### DIFF
--- a/rerun_py/tests/e2e_redap_tests/test_table_read.py
+++ b/rerun_py/tests/e2e_redap_tests/test_table_read.py
@@ -48,31 +48,31 @@ def test_datafusion_catalog_get_tables(prefilled_catalog: PrefilledCatalog) -> N
 
     # Note: as of DataFusion 50.0.0 this is not a DataFrame
     # but rather a python object that describes the table.
-    table = schema_provider.table("simple_datatypes")
+    table = schema_provider.table(prefilled_catalog.factory.apply_prefix("simple_datatypes"))
     assert table is not None
 
     schema_provider = catalog_provider.schema("second_schema")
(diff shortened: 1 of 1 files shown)


abey79 (commented):
Do I understand correctly that this means everywhere in the SDK implementation we must know if we might alter the catalog, and if we do, call `update_catalog_providers`? This seems extremely error prone. Matter of fact, I believe at least `Entry.delete()` should update the catalog provider as well. (I did the exact same mistake with our very fragile `__entries` table stuff on OSS server.)

timsaucer:
> Do I understand correctly that this means everywhere in the SDK implementation we must know if we might alter the catalog, and if we do, call `update_catalog_providers`? This seems extremely error prone. Matter of fact, I believe at least `Entry.delete()` should update the catalog provider as well. (I did the exact same mistake with our very fragile `__entries` table stuff on OSS server.)

Yes, I'm trying to figure out a better way. I have started a PR for upstream to expose `CatalogProviderList` via FFI which will resolve this problem. That has missed the boat for DF 51, but maybe could make it into a patch release.

timsaucer:
For reference: https://github.com/apache/datafusion/pull/18657

timsaucer:
@abey79 Ok, I think this is okay for now especially since the DF 51 is on it's way

PR #11879, opened 2025-11-12
Parse `PoseStamped` from ROS 2 MCAPs as `InstancePoses3D`

MichaelGrupp:
Parses `geometry_msgs::PoseStamped` ROS messages into `InstancePoses3D`
entities with associated `CoordinateFrame` named after the message's `frame_id`.

Part of https://linear.app/rerun/issue/RR-2852

---

Example showing a drone groundtruth trajectory from the [Euroc MAV dataset](https://docs.openvins.com/gs-datasets.html#gs-data-euroc) converted to MCAP PoseStamped via [evo](https://github.com/MichaelGrupp/evo):

https://github.com/user-attachments/assets/f8b35fbb-029b-4fd9-ad7b-f6f6357d726b

### DIFF
--- a/crates/store/re_mcap/src/parsers/ros2msg/geometry_msgs/pose_stamped.rs
+++ b/crates/store/re_mcap/src/parsers/ros2msg/geometry_msgs/pose_stamped.rs
@@ -1,4 +1,4 @@
-use super::super::definitions::geometry_msgs::PoseStamped;
+use super::super::definitions::geometry_msgs::{Pose, PoseStamped};
 use re_chunk::{Chunk, ChunkId};
 
 use re_sdk_types::archetypes::{CoordinateFrame, InstancePoses3D};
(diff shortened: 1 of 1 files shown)


crates/store/re_mcap/src/parsers/ros2msg/geometry_msgs/pose_stamped.rs
```diff
@@ -0,0 +1,93 @@
...
+            quaternions: Vec::with_capacity(num_rows),
+            frame_ids: Vec::with_capacity(num_rows),
+        }
+    }
+}
+
+impl MessageParser for PoseStampedMessageParser {
+    fn append(&mut self, ctx: &mut ParserContext, msg: &mcap::Message<'_>) -> anyhow::Result<()> {
+        re_tracing::profile_function!();
+        let PoseStamped { header, pose } = cdr::try_decode_message::<PoseStamped>(&msg.data)?;
```
emilk, line 34: Using the destruction pattern all the way ensures we don't ever forget any field. It is a good habit to get into.

```suggestion
        let PoseStamped { header, pose } = cdr::try_decode_message::<PoseStamped>(&msg.data)?;
        let Pose { position, orientation } = pose;
```

PR #11880, opened 2025-11-12
Some tweaks to draft table APIs

jleibs:
### Related

- related to https://linear.app/rerun/issue/RR-2901/pr-tracking-clean-up-cloud-apis-project

### What

Mostly making sure I understand the shim plumbing and fixing things so I get correct typing in vs code.

I know there are downsides to using multiple inheritance like this, but, I wanted to see what it would look like if TableEntry just acted like a dataframe with a few Entry-related extras.

### DIFF
--- a/rerun_py/tests/api_sandbox/test_current/test_table_api.py
+++ b/rerun_py/tests/api_sandbox/test_current/test_table_api.py
@@ -1,0 +1,60 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import datafusion
+import pyarrow as pa
+import rerun as rr
+from inline_snapshot import snapshot as inline_snapshot
+
+if TYPE_CHECKING:
+    import pytest
+
+
+def test_table_api(tmp_path_factory: pytest.TempPathFactory) -> None:
+    with rr.server.Server() as server:
+        client = server.client()
+
+        tmp_path = tmp_path_factory.mktemp("my_table")
+
+        table = client.create_table_entry(
+            "my_table",
+            pa.schema([
+                ("rerun_segment_id", pa.string()),
+                ("operator", pa.string()),
+            ]),
+            tmp_path.as_uri(),
+        )
+
+        assert isinstance(table.df(), datafusion.DataFrame)
+
+        assert str(table.df().schema()) == inline_snapshot("""\
+rerun_segment_id: string
+operator: string
+-- schema metadata --
+sorbet:version: '0.1.1'\
+""")
+
+        assert str(table.df().collect()) == inline_snapshot("[]")
+
+        client.append_to_table(
+            "my_table",
+            rerun_segment_id=["segment_001", "segment_002", "segment_003"],
+            operator=["alice", "bob", "carol"],
+        )
+
+        assert str(table.df().select("rerun_segment_id", "operator")) == inline_snapshot("""\
+┌─────────────────────┬─────────────────────┐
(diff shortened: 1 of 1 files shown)


rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1284,7 +1283,7 @@ class Entry:
         """The entry's name."""
 
     @property
-    def catalog(self) -> CatalogClient:
+    def catalog(self) -> CatalogClientInternal:
```
abey79, line 1287: That's correct, but also extremely broken. We shouldn't be exposing or returning the internal catalog here. I guess the best way to deal with this correctly is to roll the wrapper/internal pattern to entry.

```suggestion
    # TODO(RR-2938): this should return `CatalogClient`
    def catalog(self) -> CatalogClientInternal:
```

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -80,8 +83,12 @@ def register_table(self, name: str, url: str) -> TableEntry:
         """Registers a foreign Lance table as a new table entry."""
         return TableEntry(self._inner.register_table(name, url))
 
-    def create_table_entry(self, name: str, schema, url: str) -> TableEntry:
+    def create_table(self, name: str, schema, url: str | None = None) -> TableEntry:
         """Create and register a new table."""
+        if url is None:
+            tmpdir = tempfile.TemporaryDirectory()
+            self.tmpdirs.append(tmpdir)
+            url = Path(tmpdir.name).as_uri()
```
abey79, line 89: I love the utter nastiness we are allowed in here 😅

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -290,28 +305,33 @@ def do_maintenance(
         )
 
 
-class TableEntry(Entry):
+class TableEntry(Entry, datafusion.DataFrame):
```
abey79: As far as this PR goes, I'm obviously open to experimentation and this is fine. As for API design, I unsurprisingly the opposite of in love with this though 😅 Multiple inheritance is obviously broken in man ways, but it feels particularly nasty when one of the base class is outside of the control of the project. For example, `append` and `update` could easily conflict with future datafusion apis.

```suggestion
# TODO(RR-2939): can we do better than multiple inheritance?
class TableEntry(Entry, datafusion.DataFrame):
```

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -80,8 +83,12 @@ def register_table(self, name: str, url: str) -> TableEntry:
         """Registers a foreign Lance table as a new table entry."""
         return TableEntry(self._inner.register_table(name, url))
 
-    def create_table_entry(self, name: str, schema, url: str) -> TableEntry:
+    def create_table(self, name: str, schema, url: str | None = None) -> TableEntry:
```
abey79, line 84: Note: we tried to establish the convention that: `xyz` would return "raw data" (aka dataframe), while `xyz_entry` returns the python object representation. We couldn't really make it super consistent, but the df/object hybrid obviously doesn't help here.

abey79 (approved):
I obviously get shills up my spine with the multiple inheritance thing, but this is the point of this sandbox, so lgtm

PR #11884, opened 2025-11-13
Continuing to update the draft dataset APIs

jleibs:
### Related

- related to https://linear.app/rerun/issue/RR-2901/pr-tracking-clean-up-cloud-apis-project

### What

Work-in-progress. Introducing:
- Metadata join helpers
- DatasetViews to filter segments / entity_paths

### DIFF
--- a/rerun_py/tests/api_sandbox/test_draft/test_table_basics.py
+++ b/rerun_py/tests/api_sandbox/test_draft/test_table_basics.py
@@ -1,0 +1,216 @@
+from __future__ import annotations
+
+import pyarrow as pa
+import rerun_draft as rr
+from inline_snapshot import snapshot as inline_snapshot
+from rerun.catalog import TableInsertMode
+
+
+def test_create_table_and_append() -> None:
+    """Create a table and append data using Python values."""
+    with rr.server.Server() as server:
+        client = server.client()
+
+        # Create a table with a schema
+        schema = pa.schema([
+            pa.field("id", pa.int32()),
+            pa.field("value", pa.float64()),
+            pa.field("enabled", pa.bool_()),
+        ])
+
+        table = client.create_table("my_table", schema)
+
+        # Append single row with scalar values
+        table.append(id=1, value=10.5, enabled=True)
+
+        # Append multiple rows using lists
+        table.append(id=[2, 3, 4], value=[20.3, 15.7, 30.2], enabled=[False, True, False])
+
+        # Read the table back
+        df = table.reader()
+
+        assert str(df.sort("id")) == inline_snapshot("""\
+┌───────────────────────────────────────────────────────────────────┐
+│ METADATA:                                                         │
+│ * version: 0.1.1                                                  │
+├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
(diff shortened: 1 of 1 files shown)


rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -185,9 +188,31 @@ def schema(self) -> Any:
     def segment_ids(self) -> list[str]:
         return self._inner.partition_ids()
 
-    def segment_table(self) -> datafusion.DataFrame:
+    def segment_table(
+        self, join_meta: datafusion.DataFrame | None = None, join_key: str = "rerun_segment_id"
+    ) -> datafusion.DataFrame:
```
abey79, line 239: I really like this! (Also goes to show how powerful slapping some python code on top of internal object can be.)

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -304,6 +329,201 @@ def do_maintenance(
             unsafe_allow_recent_cleanup=unsafe_allow_recent_cleanup,
         )
 
+    def filter_dataset(
+        self,
+        *,
+        segment_ids: datafusion.DataFrame | Sequence[str] | None = None,
+        entity_paths: Sequence[str] | None = None,
+    ) -> DatasetView:
+        """Returns a new DatasetEntry filtered to the given segment IDs."""
```
abey79: With the mocking code getting increasingly fancy, might be worth adding comments to state the actual intended behaviour. In this case, are we actually introducing some kind of python-side `DatasetView`? Or server-side of the same? Or just a regular, server-side, filtered-and-materialised `Dataset`?

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -304,6 +329,201 @@ def do_maintenance(
...
+            return filtered_schema
+
+    def segment_ids(self) -> list[str]:
+        if self._lazy_state.filtered_segments is not None:
+            return [pid for pid in self._inner.partition_ids() if pid in self._lazy_state.filtered_segments]
+        else:
+            return self._inner.partition_ids()
+
+    def download_segment(self, segment_id: str) -> Any:
+        return self._inner.download_partition(segment_id)
```
abey79, line 495: Shouldn't this fail if the segment is filtered out?

rerun_py/tests/api_sandbox/test_draft/conftest.py
abey79, line 1: would be great to roll the same in `test_current`, such as to minimise the noise when diffing snippets

rerun_py/tests/api_sandbox/test_draft/test_dataset_basics.py
```diff
@@ -52,3 +53,49 @@ def test_dataset_basics(simple_dataset_prefix: Path) -> None:
...
+
+        # TODO(jleibs): Consider attaching this metadata table directly to the dataset
+        # and automatically joining it by default
+        meta = client.create_table(
+            "basic_dataset_metadata",
+            pa.schema([
+                ("rerun_segment_id", pa.string()),
+                ("success", pa.bool_()),
+            ]),
+        )
```
abey79, line 109: this should use the fixture, right?

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: worth having a knob to control the case where some segments are missing in the metadata table? I can see both "fill nulls" and "drop rows" being contextually useful. Footgun could be avoided by defaulting to "strict" mode that would fail on a segment missing a matching  metadata row.

PR #11886, opened 2025-11-13
Make the e2e redap test suite target server configurable

abey79:
### What

This PR make the e2d redap test suite configurable such that it can (optionally) be pointed at an arbitrary redap server.

In a nutshell, the default is to run with OSS servers (via the python `Server` object):
```
pixi run -e py pytest -c rerun_py/pyproject.toml rerun_py/tests/e2e_redap_tests
```

Alternatively, it can run against any running server:

```
pixi run -e py pytest -c rerun_py/pyproject.toml rerun_py/tests/e2e_redap_tests --redap-url=rerun+http://localhost:51234
```

For remote server, the test resources must be made available to said server, and the corresponding prefix provided:

```
pixi run -e py pytest -c rerun_py/pyproject.toml rerun_py/tests/e2e_redap_tests --redap-url=rerun://sandbox.redap.rerun.io --resource-prefix="s3://..."
```

For this magic to happen, the following has been done:

- complete revamp of `conftest.py` to add the custom pytest options
- update all the fixtures so entries are cleaned up after tests (no clean up happened before, because the OSS server was restarted for each test)
- mark some tests as `local_only` if they require the ability to create local rrd and register them (as opposed to using the standard resources)
- automatically skip test if the resource prefix is not `file://`
- fixed the table test so that they are either of:
  - read-only (in which case they can use the provided table resource)
  - write to a `create_table` table (fine, that gets cleaned up by the fixture)
  - write to `register_table` (special care needed: table files are first copied, and then discarded after mutation, thus local only)
- SDK's `Schama` is fixed to always iterate over index column before component column (for some reason that order was inconsistent between OSS and DPF)
- bunch of test had bug making them fail with DPF, including not blocking on tasks when `register_prefix`

### DIFF
--- a/rerun_py/tests/e2e_redap_tests/resources/dataset/generate_dataset.py
+++ b/rerun_py/tests/e2e_redap_tests/resources/dataset/generate_dataset.py
@@ -9,13 +9,17 @@ This script generates 20 .rrd files with test data containing:
 The generated files are used for testing gRPC calls and query operations.
 """
 
+import typing
 from datetime import datetime, timedelta
 import numpy as np
 import random
 import rerun as rr
 
 
-def maybe_val(val):
+_T = typing.TypeVar("_T")
+
+
+def maybe_val(val: _T) -> _T | None:
     """Return value with 80% probability, None otherwise."""
     if random.random() > 0.2:
         return val
@@ -23,14 +27,14 @@ def maybe_val(val):
         return None
 
 
-def generate_nanosecond_time(base_time, minute_delta: int):
+def generate_nanosecond_time(base_time: np.datetime64, minute_delta: int) -> np.datetime64:
     """Generate a timestamp with nanosecond precision."""
     base_ns = base_time.astype(np.int64)
     new_time_ns = base_ns + minute_delta * 60 * 1_000_000_000
     return np.datetime64(0, "ns") + np.timedelta64(new_time_ns, "ns")
 
 
-def generate_data(filename, n_rows):
+def generate_data(filename: str, n_rows: int) -> None:
     """Generate a single .rrd file with test data."""
     # Intentionally create a timestamp that has values all the way down to nanosecond
     base_time = np.datetime64("2024-01-15T10:30:45.123456789")


abey79:
@rerun-bot full-check

rerun_py/tests/e2e_redap_tests/resources/dataset/README.md
```diff
@@ -0,0 +1,104 @@
...
+These RRDs are an example dataset for testing gRPC calls between rerun and the OSS server.
+
+## Generating
+
+We could either generate the dataset on the fly or we can commit RRDs to the repository.
+By generating on the fly we run the issue of making our testing framework more complicated,
+but committing the files leads to a potential situation where we need to regenerate the
+files ad hoc. For the initial implementation of our testing framework we are choosing to
+create the files and commit them to the repository. The below python code was used
+to generate the first versions of these datasets.
```
Wumpf: why don't you add this as a regular script here and refer to it? If I want to run it myself then I have to copy that out now :(

rerun_py/tests/e2e_redap_tests/resources/.gitattributes
```diff
@@ -0,0 +1,4 @@
+*.rrd filter=lfs diff=lfs merge=lfs -text
```
Wumpf, line 1: Really liked your readme in `dataset`, can this folder have tone too?

rerun_py/tests/e2e_redap_tests/README.md
```diff
@@ -0,0 +1,40 @@
+# End-to-end redap tests
```
Wumpf, line 1: very nice overview, thank you!

rerun_py/tests/e2e_redap_tests/conftest.py
```diff
@@ -45,69 +112,182 @@ def setup_windows_tzdata() -> None:
...
-    ds.register_prefix(DATASET_FILEPATH.as_uri())
+    # Create the dataset directly (not using entry_factory since it's function-scoped)
+    ds = catalog_client.create_dataset(dataset_name)
+    tasks = ds.register_prefix(resource_prefix + "dataset")
+    tasks.wait(timeout_secs=50)
 
     yield ds
 
+    # Cleanup at session end
+    try:
```
Wumpf, line 268: shouldn't we put everything around a try/finally here? not sure if it still makes sense to delete if we hit the timeout

rerun_py/tests/e2e_redap_tests/README.md
abey79: Frankly, it's a perk of using AI. It kept tweaking it and keeping it up-to-date as "I" was making change. I did do some editing, but that's easier than summoning the will for writing one from scratch.

PR #11890, opened 2025-11-13
Fix non semver compliant breaking changes in `ply-rs` breaking Rust builds

teh-cmc:
Make sure we can compile with all 2.0.x versions of `ply-rs` out there.

See https://rerunio.slack.com/archives/C095BFLA4V6/p1763051394591979.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -297,7 +297,7 @@ pathdiff = "0.2.3"
 percent-encoding = "2.3"
 pico-args = "0.5.0"
 pin-project-lite = "0.2.16"
-ply-rs-bw = { version = "2.0.3", default-features = false }
+ply-rs-bw = { version = "=2.0.3", default-features = false } # ply-rs-bw has released semver breaking changes in patch releases before. Fool me once, shame on you…
 poll-promise = "0.3.0"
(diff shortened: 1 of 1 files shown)


Wumpf (approved):
thanks!

teh-cmc:
Those deny failures seem completely unrelated, wtf

Cargo.toml
```diff
@@ -297,7 +297,7 @@ pathdiff = "0.2.3"
 percent-encoding = "2.3"
 pico-args = "0.5.0"
 pin-project-lite = "0.2.16"
-ply-rs-bw = { version = "2.0.0", default-features = false }
+ply-rs-bw = { version = "2.0.3", default-features = false }
```
emilk: This might warrant pinning to avoid a repeat

```suggestion
ply-rs-bw = { version = "=2.0.3", default-features = false }  # ply-rs-bw has released semver breaking changes in patch releases before. Fool me once, shame on you… 
```

IsseW: @teh-cmc pointed out that locking the crate version might create problems for people depending on this crate as well

Wumpf (approved):

Issue #10575, opened 2025-07-09
Replace uses of `egui_extras::Table` with `egui_table`


emilk:
Let's migrate all our code to use the same table widget

PR #11893, opened 2025-11-14
Introduce re_table, implementing styling and basic table things

lucasmerlin:
### Related

* part of https://linear.app/rerun/issue/RR-2360/multi-selection-within-a-partition-table

### What

Simplifies the datafusion table code by pulling out the fuctionality that doesn't need to know about datafusion in a re_table wrapper. It handles styling, column order / visibility and eventually row selection. For now I'll leave it in the datafusion table crate, in case we do find something in re_table that does require tighter integration with datafusion. But eventually it could become a separate crate, and be used for the dataframe table as well to add selection / row numbers / table config.


emilk:
* Is this maybe be a step towards https://github.com/rerun-io/rerun/issues/10575 ?

lucasmerlin:
> * Is this maybe be a step towards [Replace uses of `egui_extras::Table` with `egui_table` #10575](https://github.com/rerun-io/rerun/issues/10575) ?

Could be, maybe we could add a compability layer that adds a egui_extras like api to this. On the other hand, maybe that compability layer could/should exist in egui_table? Then we could deprecate and eventually remove the egui extras table

crates/viewer/re_dataframe_ui/src/re_table.rs
```diff
@@ -0,0 +1,139 @@
...
+use std::sync::Arc;
+
+/// Wrapper around [`egui_table`] that handles styling, selection, column visibility, row numbers, etc.
+pub struct ReTable<'a> {
+    session_id: Id,
+    inner: &'a mut dyn egui_table::TableDelegate,
+    config: &'a TableConfig,
+    num_rows: u64,
+    table_style: TableStyle,
+    original_style: Arc<egui::Style>,
```
abey79, line 20: the purpose of this needs a docstring, it's not immediately obvious to me what it is exactly.

crates/viewer/re_dataframe_ui/src/re_table.rs
```diff
@@ -0,0 +1,139 @@
+use crate::re_table_utils::{TableConfig, apply_table_style_fixes, cell_ui, header_ui};
+use egui::{Context, FontSelection, Id, Rangef, RichText, TextWrapMode, Ui, WidgetText};
+use egui_table::{CellInfo, HeaderCellInfo, PrefetchInfo};
+use re_format::format_uint;
+use re_ui::{TableStyle, UiExt as _};
+use std::iter;
+use std::sync::Arc;
+
+/// Wrapper around [`egui_table`] that handles styling, selection, column visibility, row numbers, etc.
```
abey79:
```suggestion
/// Wrapper around [`egui_table::TableDelegate`] that handles styling, selection, column visibility, row numbers, etc.
```

abey79 (approved):
nice

PR #11895, opened 2025-11-14
Rename `ControlEye` to `EyeController`

MichaelGrupp:
Improves readability. The new name makes it clearer that it's only an utility for mutating eye controls, not an eye itself.

Also renames awkward constructs like `eye.get_eye()` to `eye_controller.get_eye()`.

This is a cleanup in preparation for https://linear.app/rerun/issue/RR-178, where we noticed that the current eye code is a bit confusing to read (this might not be the last pure refactor, but let's go step by step)

### DIFF
--- a/crates/viewer/re_view_spatial/src/eye.rs
+++ b/crates/viewer/re_view_spatial/src/eye.rs
@@ -294,7 +294,7 @@ impl EyeController {
         })
     }
 
-    /// Saves the eye controls that have changed their values to the blueprint.
+    /// Saves the subset of eye controls that can change through user input to the blueprint.
     /// Does nothing if no interaction happened.
     fn save_to_blueprint(
         &self,


crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -293,8 +294,8 @@ impl ControlEye {
         })
     }
 
-    /// Saves position, look target and eye up to blueprint if we have interacted
-    /// and that field has changed.
+    /// Saves the eye controls that have changed their values to the blueprint.
+    /// Does nothing if no interaction happened.
```
MichaelGrupp, line 298: Removed mentions of specific parameters in doc comment (impl detail). Makes it easier to extend/change.

IsseW: Hmm it doesn't actually save everything though. It's only those 3 specific ones

MichaelGrupp: True, it only saves the ones that can change via input interaction. How about this, I think it would be simple and clear:
```
Saves the eye controls that have changed through user input to the blueprint.
```

MichaelGrupp: Changed it now to 
```
    /// Saves the subset of eye controls that can change through user input to the blueprint.
    /// Does nothing if no interaction happened.
```
This should make it clear without having to mention each single parameter.

IsseW: Yep looks good!

Issue #6824, opened 2024-07-09
Make TextLogView customizable from Blueprint


Wumpf:
These settings should be part of the blueprint:

![image](https://github.com/rerun-io/rerun/assets/1220815/672b5351-d854-48f0-981c-a62444b6d54f)

56789KD:
Issue a change view comment

PR #11896, opened 2025-11-14
Text log view blueprint properties

IsseW:
### Related

* Closes https://github.com/rerun-io/rerun/issues/6824

### What

Moves text log view properties to blueprint.

Screenshot 2025-11-18 at 11 37 07

### DIFF
--- a/crates/viewer/re_component_ui/src/text_log_columns.rs
+++ b/crates/viewer/re_component_ui/src/text_log_columns.rs
@@ -1,6 +1,7 @@
-use re_types::blueprint::{components::TextLogColumn, datatypes::TextLogColumnKind};
-use re_ui::{HasDesignTokens as _, UiExt as _};
+use re_types::blueprint::components::TextLogColumn;
 use re_viewer_context::{MaybeMutRef, ViewerContext};
+
+use crate::visible_dnd::visible_dnd;
 
 pub fn edit_or_view_columns_singleline(
     _ctx: &ViewerContext<'_>,
@@ -38,77 +39,20 @@ pub fn edit_or_view_columns_multiline(
             .map(|column| ui.strong(column.kind.name()))
             .reduce(|a, b| a.union(b))
             .unwrap_or_else(|| ui.weak("Empty")),
-        MaybeMutRef::MutRef(columns) => {
-            let mut any_edit = false;
-
-            const ITEM_SPACING: f32 = 8.0;
-            let egui::InnerResponse { mut response, .. } = egui::Frame::new()
-                .corner_radius(ui.visuals().menu_corner_radius)
-                .fill(ui.visuals().tokens().text_edit_bg_color)
-                .inner_margin(egui::Margin {
-                    left: ITEM_SPACING as i8,
-                    right: ITEM_SPACING as i8,
-                    top: ITEM_SPACING as i8,
-                    bottom: (ITEM_SPACING * 0.5) as i8,
-                })
-                .show(ui, |ui| {
-                    let text_height = ui
-                        .style()
-                        .text_styles
-                        .get(&egui::TextStyle::Body)
-                        .map(|s| s.size)
-                        .unwrap_or(0.0);
-                    let sz = egui::vec2(ui.max_rect().size().x, ITEM_SPACING + text_height);
-                    let dnd_res = egui_dnd::dnd(ui, "text_log_columns_dnd").show_sized(
-                        // We include the index in the item here because the item
-                        // so doing this will make columns with the
-                        // same name not collide.
-                        columns.iter_mut().enumerate(),
-                        sz,
-                        |ui, (_idx, col), handle, _state| {
-                            ui.horizontal(|ui| {
-                                handle.ui(ui, |ui| {
-                                    ui.small_icon(
-                                        &re_ui::icons::DND_HANDLE,
-                                        Some(ui.visuals().text_color()),
-                                    );
-                                });
-
-                                let visible = col.visible.0;
-
-                                let col = &mut **col;
-                                egui::containers::Sides::new().shrink_left().show(
-                                    ui,
-                                    |ui| {
-                                        let column: &mut TextLogColumnKind = &mut col.kind;
-                                        let name = column.name();
-                                        if visible {
-                                            ui.strong(name);
-                                        } else {
-                                            ui.weak(name);
-                                        }
-                                    },
-                                    |ui| {
-                                        any_edit |= ui
-                                            .visibility_toggle_button(&mut col.visible.0)
-                                            .changed();
-                                    },
-                                );
-                            });
-                        },
-                    );
-
-                    if dnd_res.is_drag_finished() {
-                        any_edit = true;
-                        dnd_res.update_vec(columns);
-                    }
-                });
-
-            if any_edit {
-                response.mark_changed();
-            }
-
-            response
-        }
+        MaybeMutRef::MutRef(columns) => visible_dnd(
+            ui,
+            "text_log_columns_dnd",
+            columns,
+            |ui, col| {
+                let name = col.kind.name();
+                if *col.visible {
+                    ui.strong(name);
+                } else {
+                    ui.weak(name);
+                }
+            },
+            |col| *col.visible,
+            |col, v| col.visible = v.into(),
+        ),
     }
 }
--- a/crates/viewer/re_component_ui/src/timeline_columns.rs
+++ b/crates/viewer/re_component_ui/src/timeline_columns.rs
@@ -2,8 +2,9 @@ use re_data_ui::item_ui::timeline_button;
 
 use re_log_types::TimelineName;
 use re_types::blueprint::components::TimelineColumn;
-use re_ui::{HasDesignTokens as _, UiExt as _};
 use re_viewer_context::{MaybeMutRef, ViewerContext};
+
+use crate::visible_dnd::visible_dnd;
 
 pub fn edit_or_view_columns_singleline(
     ctx: &ViewerContext<'_>,
@@ -33,8 +34,8 @@ pub fn edit_or_view_columns_multiline(
             .reduce(|a, b| a.union(b))
             .unwrap_or_else(|| ui.weak("Empty")),
         MaybeMutRef::MutRef(columns) => {
-            let mut any_edit = false;
-
+            // Add new timelines to the end of the UI, if there is any edit
+            // these will be written to the component.
             let extra_columns = ctx
                 .recording()
                 .times_per_timeline()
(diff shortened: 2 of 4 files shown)


crates/store/re_types/definitions/rerun/blueprint/archetypes/text_log_columns.fbs
```diff
@@ -0,0 +1,16 @@
...
+table TextLogColumns (
+  "attr.rerun.scope": "blueprint"
+) {
+  // --- Optional ---
+
+  /// All columns to be displayed.
+  text_log_columns: rerun.blueprint.components.TextLogColumnList ("attr.rerun.component_optional", order: 1000);
+
+  /// What timeline the timeline column should show.
+  timeline: rerun.blueprint.components.TimelineName ("attr.rerun.component_optional", order: 2000);
```
Wumpf: should point out what we do by default

IsseW:
Updating the snippet here and making the python api a bit nicer for this now too. But I think the rust side should be good to review.

crates/build/re_types_builder/src/codegen/cpp/mod.rs
```diff
@@ -1050,6 +1050,7 @@ impl QuotedObject {
 
         hpp_includes.insert_system("utility"); // std::move
         hpp_includes.insert_system("cstring"); // std::memcpy
+        hpp_includes.insert_system("cassert"); // assert
```
Wumpf: interesting we're hitting this only now. Can you point me to where we emit asserts which were missing? Probably we can narrow this down a bit and only include when necessary

crates/store/re_types/definitions/rerun/blueprint/archetypes/text_log_columns.fbs
Wumpf: applies to all the fbs files

crates/store/re_types/definitions/rerun/blueprint/datatypes/text_log_column_list.fbs
```diff
@@ -0,0 +1,12 @@
...
+
+/// A list of text log columns.
+table TextLogColumnList (
+  "attr.rust.derive": "Default, PartialEq, Eq",
+  "attr.rerun.scope": "blueprint",
+  "attr.docs.unreleased"
+) {
+  /// All columns to be displayed.
+  text_log_columns: [rerun.blueprint.datatypes.TextLogColumn] (order: 100);
+}
```
Wumpf: Don't entirely understand why we need this as a datatype. Can't we just us component arrays of `TextLogColumn` instead? Is the problem that our fallbacks and comiponent uis don't produce lists today? If so I think that's something we can change without going totally over board.
Was always surprised this didn't catch up with us sooner...

crates/store/re_types/definitions/rerun/blueprint/datatypes/text_log_level_list.fbs
```diff
@@ -0,0 +1,12 @@
...
+
+/// A list of text log levels.
+table TextLogLevelList (
+  "attr.rust.derive": "Default, PartialEq, Eq",
+  "attr.rerun.scope": "blueprint",
+  "attr.docs.unreleased"
+) {
+  /// Log levels to display.
+  log_levels: [rerun.datatypes.Utf8] (order: 100);
+}
```
Wumpf: as above, we really shouldn't need this as this level

crates/store/re_types/src/blueprint/components/text_log_level_list_ext.rs
```diff
@@ -0,0 +1,8 @@
+use super::TextLogLevelList;
+
+impl Default for TextLogLevelList {
```
Wumpf: you can also define an attr default in the fbs file for this instead

crates/viewer/re_component_fallbacks/src/blueprint_component_fallbacks.rs
```diff
@@ -4,6 +4,12 @@ use re_viewer_context::FallbackProviderRegistry;
 pub fn type_fallbacks(registry: &mut FallbackProviderRegistry) {
     registry.register_type_fallback_provider(|_| blueprint::components::ForceDistance::from(60.));
     registry.register_type_fallback_provider(|_| blueprint::components::ForceStrength::from(1.));
+
+    registry.register_type_fallback_provider(|ctx| {
+        blueprint::components::TimelineName::from(
```
Wumpf: ah nice, I can see how this should come in handy in a lot of places

crates/viewer/re_view_text_log/src/view_class.rs
```diff
@@ -166,6 +183,66 @@ Filter message types and toggle column visibility in a selection panel.",
...
+            .map(|col| col.kind)
+            .collect();
+
+        let levels_list = rows_property.component_or_fallback::<TextLogLevelList>(
+            &view_ctx,
+            TextLogRows::descriptor_filter_by_log_level().component,
+        )?;
+        let levels = &levels_list.0.log_levels;
+
+        let reset_column_widths = if columns.as_slice() != state.last_columns.as_slice() {
```
Wumpf: fully resetting when adding a column seems to be too aggressive?

crates/viewer/re_view_text_log/src/view_class.rs
```diff
@@ -299,50 +323,39 @@ fn table_ui(
         .max_scroll_height(f32::INFINITY) // Fill up whole height
         .cell_layout(egui::Layout::left_to_right(egui::Align::Center));
 
+    if reset_column_widths {
+        table_builder.reset();
```
Wumpf: related to above, we never did a full reset before, right?

crates/viewer/re_view_text_log/src/view_class.rs
```diff
@@ -438,6 +444,131 @@ fn table_ui(
...
+    column: &bp_datatypes::TextLogColumnKind,
+    timeline: &TimelineName,
+) -> egui::Response {
+    match column {
+        bp_datatypes::TextLogColumnKind::Timeline => item_ui::timeline_button(ctx, ui, timeline),
+        _ => ui.strong(column.name()),
+    }
+}
+
+// We need this to be a custom ui to bew able to use the view state to get seen text log levels. This could potentially be avoided if we could add component ui's from this crate.
```
Wumpf:
```suggestion
// We need this to be a custom ui to be able to use the view state to get seen text log levels.
// This could potentially be avoided if we could add component ui's from this crate.
```

well. so.. we actually can add component uis from this crate. Would be a first, but don't think there's anything to prevent us doing that in on_register

crates/build/re_types_builder/src/codegen/cpp/mod.rs
IsseW: I hit it when I tried to use a union where I had which timeline stored in the column, but that didn't work out because later I hit that unions aren't supported. Can remove this line now that it isn't used.

crates/store/re_types/definitions/rerun/blueprint/datatypes/text_log_column_list.fbs
IsseW: I did try doing that with custom UI and supporting array fallbacks at the start. But another problem I ran into then is that it's not possible as far as I know to tell the difference between empty array and never having logged it then?

crates/viewer/re_view_text_log/src/view_class.rs
IsseW: Yeah, although the table itself doesn't know which column is which, so swapping order with say body and log level makes log level as wide as body was. Which in my opinion was even worse.

Not sure what the best way to improve this would be.

crates/build/re_types_builder/src/codegen/cpp/mod.rs
Wumpf: 👍 
we had a lot of trouble with unions in the past

crates/store/re_types/definitions/rerun/blueprint/datatypes/text_log_column_list.fbs
Wumpf: in the viewer we very often treat them the same way for practical reasons, but if you go "low level" enough it can be done

crates/viewer/re_view_text_log/src/view_class.rs
IsseW: Changed it to only do a reset for elements that changed order now, since the table only remembers column sizes by index.

crates/build/re_types_builder/src/codegen/python/mod.rs
```diff
@@ -2166,20 +2168,42 @@ fn quote_arrow_serialization(
...
+                code.push_indented(
+                    1,
+                    format!(
+                        "typed_data = [x if isinstance(x, {name}) else {name}(x) for x in data]"
+                    ),
+                    2,
+                );
+            } else {
+                code.push_indented(1, "typed_data = data", 2);
+            }
```
IsseW, line 2200: would we prefer _ext classes over this?

This applied to the two cases I had, with no existing cases that had both overrode init and a python alias.

Allows us to do `["some", "timelines"]`, which from the type hint of the function you'd expect.

crates/viewer/re_view_text_log/src/view_class.rs
IsseW: Didn't succeed in making it a registered UI because it would need a view context to get the view state. While component UI's get viewer context. So would need something more similar to the fallback registry in that case, where component UI's registered in a view's `on_register` get registered for that specific view. And used with that specific view context.

crates/build/re_types_builder/src/codegen/python/mod.rs
Wumpf: seems good to have in codegen instead of manual code :)

crates/viewer/re_viewer_context/src/component_ui_registry.rs
```diff
@@ -397,6 +525,44 @@ impl ComponentUiRegistry {
         ui.sanity_check();
     }
 
+    fn get_untyped_component_ui(
```
Wumpf:
```suggestion
    /// Tries to lookup a ui callback with the given constraints.
    fn untyped_component_ui_callback(
```

crates/viewer/re_viewer_context/src/component_ui_registry.rs
```diff
@@ -763,6 +902,19 @@ fn try_deserialize<C: re_types::Component>(value: &dyn arrow::array::Array) -> O
     }
 }
 
+fn try_deserialize_array<C: re_types::Component>(
```
Wumpf, line 881: nit: could use this in above `try_serialize` to ensure consistency of error handling

crates/viewer/re_component_ui/src/timeline_columns.rs
```diff
@@ -0,0 +1,126 @@
...
+        for col in columns.iter() {
+            if *col.visible {
+                timeline_button(ctx, ui, &TimelineName::new(&col.timeline));
+            }
+        }
+    })
+    .response
+}
+
+pub fn edit_or_view_columns_multiline(
```
Wumpf, line 24: would be _nice_ to do some code sharing with text_log_columns, looks very similar from afar 🤔

crates/viewer/re_component_ui/src/text_log_columns.rs
```diff
@@ -0,0 +1,111 @@
...
+                .show(ui, |ui| {
+                    let text_height = ui
+                        .style()
+                        .text_styles
+                        .get(&egui::TextStyle::Body)
+                        .map(|s| s.size)
+                        .unwrap_or(0.0);
+                    let sz = egui::vec2(ui.max_rect().size().x, ITEM_SPACING + text_height);
+                    let dnd_res = egui_dnd::dnd(ui, "text_log_columns_dnd").show_sized(
+                        columns.iter_mut().enumerate(),
```
Wumpf: should leave a comment on why this enumerates - we don't actually use it but it is used for the id (clever, but sneaky!)

PR #11897, opened 2025-11-14
cloud telemetry: enrich span fields with trace_id

andrea-reale:
This will add trace_id and benchmark_id as span fileds explicitly.

The main use of this is that they will be included in log lines span
contexts so that we can filter by it.

### DIFF
--- a/crates/utils/re_perf_telemetry/src/grpc.rs
+++ b/crates/utils/re_perf_telemetry/src/grpc.rs
@@ -104,6 +104,18 @@ impl<B> tower_http::trace::MakeSpan<B> for GrpcMakeSpan {
             headers = ?safe_headers,
             email,
             entry_id,
+            // Record trace_id and benchmark_id as top level span fields.
+            //
+            // At this stage we may not know yet the actual trace_id (depending on whether
+            // we're generating a new trace or continuing an existing one). However,
+            // we need to pre-declare these fields if we want to record values for them later.
+            //
+            // The fields will be filled in by a separate [`tracing_subscriber::Layer`] (see
+            // [`TraceIdLayer`]).
+            trace_id = tracing::field::Empty,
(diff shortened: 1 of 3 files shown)


crates/utils/re_perf_telemetry/src/grpc.rs
```diff
@@ -706,3 +718,50 @@ impl tonic::service::Interceptor for TracingInjectorInterceptor {
...
+use tracing_subscriber::{Layer, layer::Context};
+
+/// A `tracing_subscriber::Layer` that injects the opentelemetry `trace_id` as a `benchmark_id` field
+/// top level field on every span.
+///
+/// This allows us to use the upstream tooling to filter logs within a span by `trace_id`
+#[derive(Default)]
+pub struct TraceIdLayer {
+    _private: (),
+}
```
Wumpf, line 736: why not just 
```suggestion
pub struct TraceIdLayer;
```
as well here?

Wumpf (approved):
very unfamiliar with this stuff, but I think it makes sense :)

crates/utils/re_perf_telemetry/src/grpc.rs
andrea-reale: :D There's a good story - and no real good reason - for it!

While I was working on this, I found this pattern pretty much everywhere in the tracing crate, and that got me curious. I learned that they do it to make sure that if you ever add "stuff" to your struct, then no client of your struct would break when they instantiate the thing as `let x = TraceIdLayer`, because they would have all at least have to do `let x = TraceIdLayer::default()` .

Now, this thing is public, but still our own thing, so I could have just not done it :D 
(I did not do that on the other thing because not public).

Wumpf: ohh fascinating, that _is_ a really good reason for public types indeed

PR #11898, opened 2025-11-14
cpp: Export `compile_commands.json` via Pixi

MichaelGrupp:
This simplifies a lot the setup of dev tools, as it contains all the relevant include paths etc exactly as they are used by the actual build.

For example, you can configure the vscode C++ extension to use it for IntelliSense indexing by adding
`"compileCommands": ["${workspaceFolder}/build/debug/compile_commands.json"]` to the `.vscode/c_cpp_properties.json` (can be also done via settings UI, [more info here](https://code.visualstudio.com/docs/cpp/configure-intellisense#_compilecommandsjson-file)).

Then just run `pixi run cpp-prepare` (or a dependent Pixi command like `pixi run cpp-build-all`) to write/update the `compile_commands.json`.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -147,6 +147,9 @@ and build all C++ artifacts with:
 pixi run -e cpp cpp-build-all
 ```
 
+The Pixi build commands export a `compile_commands.json` compilation database to the build directory.
+This can be useful for developer tools, e.g. for [setting up IntelliSense in VSCode](https://code.visualstudio.com/docs/cpp/configure-intellisense#_compilecommandsjson-file).
+
 ## Building the docs
 
 High-level documentation for Rerun can be found at [http://rerun.io/docs](http://rerun.io/docs). It is built from the separate repository [rerun-docs](https://github.com/rerun-io/rerun-docs).


.gitignore
```diff
@@ -19,6 +19,9 @@ _deps
 **/rerun_cpp/docs/xml
 *.tgz
 
+# User-specific VSCode C++ extension config:
+.vscode/c_cpp_properties.json
```
MichaelGrupp, line 23: Example file that I use. All fields except for `compileCommands` are auto-generated defaults that get written when you configure vscode as explained in the guide https://code.visualstudio.com/docs/cpp/configure-intellisense#_compilecommandsjson-file
```json
{
    "configurations": [
        {
            "name": "Mac",
            "includePath": [],
            "defines": [],
            "compilerPath": "/usr/bin/clang",
            "cStandard": "c17",
            "cppStandard": "c++17",
            "intelliSenseMode": "macos-clang-arm64",
            "configurationProvider": "ms-vscode.cmake-tools",
            "compileCommands": [
                "${workspaceFolder}/build/debug/compile_commands.json"
            ]
        }
    ],
    "version": 4
}
```

Wumpf (approved):
uhh cool, didn't know about `compile_commands.json`!

MichaelGrupp:
> uhh cool, didn't know about `compile_commands.json`!

I try to use that wherever I can 😀  It's just so much nicer if you only need to worry about your `CMakeLists.txt` and not have to maintain additional configs just for developer tools.

PR #11899, opened 2025-11-14
Patch release changelogs

IsseW:
### What

Includes the changelog from 0.27.1 and 0.27.2. Merging those branches into main would be much more work than just doing this.


Wumpf (approved):
thanks!

PR #11901, opened 2025-11-14
Improved transform query semantics

Wumpf:
### Related

* Fixes RR-2906
* Fixes RR-2767
* Fixes RR-2897

### What

Makes query semantics on transform data "atomic": If any transform property changes, we consider all of them reset.

Historically, we achieved this SDK sided by logging empty arrays. However, this data-sided fix was always a bit hacky since it meant we sent & queries a lot more data than needed!

With the arrival of frame based transforms it's clear this is no longer sufficient since our regular latest-at queries create very strange & hard to reason scenarios: for example if on the same entity (!) you first log a translation for a->b and then a rotation for c->d, would the translation still apply to c->d? The natural answer seems to be no, and this is what these new atomic query semantics allow you to do.

This also now allows us to log multiple transform relations _on the same timestamp_. Example:

```py
from math import pi

import rerun as rr
from rerun.datatypes import Angle, RotationAxisAngle

rr.init("rerun_example_very_demo", spawn=True)

# Everything on the same timestamp!!!
rr.set_time("DEMO TIME", sequence=0)

for name, color in [
    ("red", [1, 0, 0, 0.2]),
    ("green", [0, 1, 0, 0.2]),
    ("blue", [0, 0, 1, 0.2]),
]:
    rr.log(
        name,
        rr.Boxes3D(
            half_sizes=[1, 1, 1],
            colors=color,
            fill_mode="solid",
        ),
        rr.CoordinateFrame(f"{name}_frame"),
    )

rr.log(
    "me_transforms",
    rr.Transform3D(
        scale=0.5,  #
        child_frame="red_frame",
        parent_frame="tf#/",
    ),
)
rr.log(
    "me_transforms",
    rr.Transform3D(
        scale=2.0,  #
        translation=[3, 0, 0],
        child_frame="green_frame",
        parent_frame="tf#/",
    ),
)
rr.log(
    "me_transforms",
    rr.Transform3D(
        translation=[0, 3, 0],  #
        rotation_axis_angle=rr.RotationAxisAngle(axis=[0, 0, 1], angle=pi / 4),
        child_frame="blue_frame",
        parent_frame="tf#/",
    ),
)
```
(not yet included as snippet, but there's a unit test added that does exactly this)
image

Will now show all three transforms instead of just the last one. With regular latest-at semantics this wasn't possible as the last one would "win" over all the others.

--

Benchmark reports about 10% improvement for cache buildup and 6.6% improvement for querying uncached frames. Both are still not great imho though.
```
build_from_entity_db    time:   [106.06 ms 107.27 ms 108.65 ms]
                        change: [-13.271% -12.118% -10.868%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) high mild
  5 (5.00%) high severe

query_uncached_frame    time:   [12.447 ms 12.563 ms 12.700 ms]
                        change: [-7.6062% -6.6114% -5.3861%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
```

--

Future work (RR-2940):
* mark archetypes as atomic and explain the concept
    * make it drop `update_fields` where possible 
* remove empty array logging from transform

--

### DIFF
--- a/crates/store/re_tf/src/transform_resolution_cache.rs
+++ b/crates/store/re_tf/src/transform_resolution_cache.rs
@@ -1486,6 +1486,7 @@ mod tests {
                     .latest_at_transform(&entity_db, &LatestAtQuery::new(*timeline.name(), 1)),
                 Some(ParentFromChildTransform {
                     parent: TransformFrameIdHash::entity_path_hierarchy_root(),
+                    // Due to atomic-latest-at, the translation is no longer visible despite being on the static chunk.
                     transform: DAffine3::from_scale(glam::dvec3(123.0, 234.0, 345.0)),
                 })
             );
@@ -1578,6 +1579,7 @@ mod tests {
             assert_eq!(
                 transforms
                     .latest_at_instance_poses(&entity_db, &LatestAtQuery::new(*timeline.name(), 1)),
+                // Due to atomic-latest-at, the translation is no longer visible despite being on the static chunk.
                 vec![DAffine3::from_scale(glam::dvec3(10.0, 20.0, 30.0)),]
             );
 


crates/store/re_tf/src/entity_to_frame_tracking.rs
```diff
@@ -13,7 +16,7 @@ use vec1::smallvec_v1::SmallVec1;
 ///
 /// The list of frame id hashes can never be empty.
 /// If a clear or empty array is logged, we insert the implicit frame again since this is what we always fall back to.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct EntityToFrameOverTime {
```
Wumpf, line 20: Actually we can massively simplify this structure and its uses downstream. Created a Linear issue for that, RR-2964.
Should link to that before landing this iteration!

---

Details (written down in there):

Previously, we had to keep track of when which frame is active. With atomic-query-semantics we no longer have to do this!

Not only are we doing a lot of unnecessary now confusing work, I also suspect it's ever so slightly wrong

I believe if we do that, we can also scrap the "entity binding" rule, i.e. transforms for any source may then appear on any entity without making things more complicated

crates/store/re_tf/src/transform_queries.rs
```diff
@@ -27,16 +29,166 @@ pub enum TransformError {
...
+                    // *Something* on this row has to be non-empty & non-null!
+                    // Example where this is not the case:
+                    //
+                    // ┌───────────────┬─────────────┬────────────┐
+                    // │ child_frame_id│ translation │ color      │
+                    // ├───────────────┼─────────────┼────────────┤
+                    // │ ["myframe"]   │ [[1,2,3]]   │ null       │
+                    // │ null          │ null        │ 0xFF00FFFF │
+                    // │ null          │ []          │ null       │
+                    // └───────────────┴─────────────┴────────────┘
```
grtlr: Sorry, I got triggered 🙈.
```suggestion
                    // ┌────────────────┬─────────────┬────────────┐
                    // │ child_frame_id │ translation │ color      │
                    // ├────────────────┼─────────────┼────────────┤
                    // │ ["myframe"]    │ [[1,2,3]]   │ null       │
                    // │ null           │ null        │ 0xFF00FFFF │
                    // │ null           │ []          │ null       │
                    // └────────────────┴─────────────┴────────────┘
```

crates/store/re_tf/src/transform_queries.rs
```diff
@@ -381,3 +568,432 @@ pub fn query_view_coordinates_at_closest_ancestor(
...
+    fn atomic_component_set() -> [ComponentIdentifier; 3] {
+        [
+            MyPoints::descriptor_points().component,
+            MyPoints::descriptor_colors().component,
+            MyPoints::descriptor_labels().component,
+        ]
+    }
+
+    fn frame_condition_component() -> ComponentIdentifier {
+        // We stick witih `MyPoints` all the way and its labels happen to be compatible with frame ids (it's just utf8!)
```
grtlr:
```suggestion
        // We stick with `MyPoints` all the way and its labels happen to be compatible with frame ids (it's just utf8!)
```

crates/store/re_tf/src/transform_resolution_cache.rs
grtlr, line 1: I think I'd extend both the `test_static_tree_transforms` and the `test_static_pose_transforms` with a temporal translation logged after the `final_static_chunk`, for completeness.

grtlr, line 2096: Should this maybe also test the regular constructor (not going through `update_fields`)?

Wumpf: my point there was actually that the translation is no longer visible. But I guess there's also a point to be made that it's still possible to log that translation which is almost the more interesting thing :thinking:

Wumpf: opted to clarify mentioned intention instead

Wumpf: good question! right now the regular constructor still logs tons of empty arrays, but update_fields doesn't. But very soon it's gonna be the same. At which point I should probably make all of those just be regular ctor :shrug:

PR #11905, opened 2025-11-17
Remove redundant `Pose*` components

grtlr:
### Related

* Follow-up of #6889.
* Closes RR-2874.
* Related: https://github.com/rerun-io/rerun/pull/12096

### What

This removes the, now redundant, ad-hoc `Pose*` components from the before tagged components time.

### TODO

* [x] `@rerun-bot full-check`

### DIFF
--- a/crates/store/re_sorbet/src/migrations/v0_1_1__to__v0_1_2.rs
+++ b/crates/store/re_sorbet/src/migrations/v0_1_1__to__v0_1_2.rs
@@ -29,7 +29,7 @@ fn migrate_pose_components(batch: RecordBatch) -> RecordBatch {
     let needs_migration = schema.fields().iter().any(|field| {
         field
             .metadata()
-            .get(re_types_core::FIELD_METADATA_KEY_COMPONENT_TYPE)
+            .get("rerun:component_type")
             .is_some_and(|component| {
                 component == "rerun.components.PoseTranslation3D"
                     || component == "rerun.components.PoseRotationAxisAngle"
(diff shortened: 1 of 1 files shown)


grtlr:
@rerun-bot full-check

crates/store/re_sorbet/src/migrations/v0_1_1__to__v0_1_2.rs
```diff
@@ -11,11 +11,111 @@ impl super::Migration for Migration {
...
+            }
+            "rerun.components.PoseRotationQuat" => "rerun.components.RotationQuat".to_owned(),
+            "rerun.components.PoseScale3D" => "rerun.components.Scale3D".to_owned(),
+            "rerun.components.PoseTransformMat3x3" => "rerun.components.TransformMat3x3".to_owned(),
+            _ => component.to_owned(),
+        }
+    }
+
+    // Create new schema with migrated component names
+    let new_fields: Vec<Arc<Field>> = schema
```
Wumpf, line 63: curious, is renaming component types a reoccurring thing in existing migrations such that we could have a utility where one just passes in the renaming rules?

crates/store/re_sorbet/src/migrations/v0_1_1__to__v0_1_2.rs
```diff
@@ -11,11 +11,111 @@ impl super::Migration for Migration {
...
+            if let Some(component) = metadata.get(re_types_core::FIELD_METADATA_KEY_COMPONENT) {
+                let new_component = migrate_component_name(component);
+                if new_component != *component {
+                    metadata.insert(
+                        re_types_core::FIELD_METADATA_KEY_COMPONENT.to_owned(),
+                        new_component,
+                    );
+                    modified = true;
+                }
+            }
```
Wumpf: wait a second, the component type should never show up on the `rerun:component` metadata. What is this block for?

tests/assets/rrd/snippets/archetypes/pinhole_projections.rrd
Wumpf, line 1: shouldn't the backwards compatibility rrds stay the same? Why did they change?
Also, why did they get much larger - this one went from 20316 bytes to 104468 bytes

Wumpf (changes requested):
awesome! this makes me very happy 😄 

However, I think I found a (fairly benign) bug in the migration code

tests/assets/rrd/snippets/archetypes/pinhole_projections.rrd
grtlr: Those `.rrd`s were already checked in with new Sorbet version `v0.1.2` so migration would never be run on them, which is why I updated them. No idea about the increase in size yet.

grtlr: Ahhh, footer probably

crates/store/re_sorbet/src/migrations/v0_1_1__to__v0_1_2.rs
grtlr: That would be nice, but it should be low-level Arrow helpers and not assume too much about `key` and `values` in the metadata.

Arguably, using `re_types_core::FIELD_METADATA` is already an antipattern, because what happens if we change that value? Will old migrations break all the sudden?

grtlr: I will backport these changes to older migrations.

tests/assets/rrd/snippets/archetypes/pinhole_projections.rrd
Wumpf: yeah that gotta be it

crates/store/re_sorbet/src/migrations/v0_1_1__to__v0_1_2.rs
grtlr: PR is up:

* https://github.com/rerun-io/rerun/pull/12096

grtlr:
@rerun-bot full-check

Issue #11865, opened 2025-11-11
ComponentBatch::from_arrow_array is not thread safe


kobeever:
**Describe the bug**
When logging data from multiple threads, there will be a crash

**To Reproduce**
Simply call the log function with some objects from multiple threads

**Expected behavior**
no crash

**Backtrace**
********** Crash dump: **********
Build fingerprint: 'Meta/diamond/diamond:14/UKQ1.250303.001/344830000002781:userdebug/test-keys'
Abort message: 'assertion `left != right` failed: Component type with the same descriptor already registered
#00 0x00000000000aa238 /apex/com.android.runtime/lib64/bionic/libc.so (tgkill+8) (BuildId: 250a5ac456b689dd08870d4bd16b1635)
#01 0x00000000054f1fd4 /data/local/tests/skadi/SkadiApp (BuildId: 1d5e51e8beb15b93edfbdf4be176af855721a973)
                                                         folly::symbolizer::(anonymous namespace)::callPreviousSignalHandler(int, siginfo*)
                                                         ./xplat/folly/debugging/symbolizer/SignalHandler.cpp:213:7
                                                         folly::symbolizer::(anonymous namespace)::signalHandler(int, siginfo*, void*)
                                                         ./xplat/folly/debugging/symbolizer/SignalHandler.cpp:551:3
#02 0x0000000000000860 [vdso]
#03 0x0000000000059aa0 /apex/com.android.runtime/lib64/bionic/libc.so (abort+160) (BuildId: 250a5ac456b689dd08870d4bd16b1635)
#04 0x00000000027a3b6c /data/local/tests/skadi/SkadiApp (BuildId: 1d5e51e8beb15b93edfbdf4be176af855721a973)
                                                         std::sys::pal::unix::abort_internal
                                                         ./xplat/rust/toolchain/sysroot/1.90.0/library/std/src/sys/pal/unix/mod.rs:382:14
#05 0x000000000279ace0 /data/local/tests/skadi/SkadiApp (BuildId: 1d5e51e8beb15b93edfbdf4be176af855721a973)
                                                         std::process::abort
                                                         ./xplat/rust/toolchain/sysroot/1.90.0/library/std/src/process.rs:2499:5
#06 0x00000000027a49c4 /data/local/tests/skadi/SkadiApp (BuildId: 1d5e51e8beb15b93edfbdf4be176af855721a973)
                                                         __rustc::__rust_abort
[...]

PR #11907, opened 2025-11-17
Make component type registration thread-safe in C++ SDK

MichaelGrupp:
* Closes: #11865

### DIFF
--- a/rerun_cpp/src/rerun/component_type_registry.cpp
+++ b/rerun_cpp/src/rerun/component_type_registry.cpp
@@ -1,5 +1,7 @@
 #include "component_type_registry.hpp"
 
+#include <mutex>
+
 namespace rerun {
 
     /// Returns the handle to the registered component type for the given descriptor.


rerun_cpp/src/rerun/component_type_registry.cpp
```diff
@@ -0,0 +1,27 @@
+#include "component_type_registry.hpp"
+
+namespace rerun {
+
+    /// Returns the handle to the registered component type for the given descriptor.
+    ///
+    /// Registers the component type when first encountered.
+    Result<ComponentTypeHandle> ComponentTypeRegistry::get_or_register(
+        const ComponentDescriptor& descriptor, std::shared_ptr<arrow::DataType> arrow_datatype
```
Wumpf: nit:

```suggestion
        const ComponentDescriptor& descriptor, const std::shared_ptr<arrow::DataType>& arrow_datatype
```

That way we won't needlessly bump the ref count if we don't have to

rerun_cpp/src/rerun/component_type_registry.cpp
```diff
@@ -0,0 +1,27 @@
...
+    ) {
+        std::lock_guard<std::mutex> lock(mutex_);
```
Wumpf: looking at this again I think we should actually use a [`shared_mutex`](https://en.cppreference.com/w/cpp/thread/shared_mutex.html): clearly folks are running this method from multiple threads and almost always we have everything we need already registered.

Makes the logic ofc a little bit more complicated but I think it's alright, let me know what you think :)
Essentially first acquire a [`std::shared_lock`](https://en.cppreference.com/w/cpp/thread/shared_lock.html) then drop it again if no value was found, then attempt again with a [`std::unique_lock`](https://en.cppreference.com/w/cpp/thread/unique_lock.html) and _potentially_ adding the element.

rerun_cpp/src/rerun/component_type_registry.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+        const auto descr_hash = descriptor.hashed();
+        if (const auto search = comp_types_per_descr_.find(descr_hash);
```
Wumpf, line 29: ah neat I've never much used these "new" condition scoped variables, but I realize now that there's some overlap top Rust's `if let`

rerun_cpp/src/rerun/component_type_registry.cpp
MichaelGrupp: Makes sense, was quite easy to change with an additional scope. PTAL

Wumpf (approved):
nice!

rerun_cpp/src/rerun/component_type_registry.cpp
Wumpf: better than I expected!

PR #11908, opened 2025-11-17
Add partition table context menu and row selection

lucasmerlin:
- closes https://linear.app/rerun/issue/RR-2360/multi-selection-within-a-partition-table
- closes https://linear.app/rerun/issue/RR-2361/open-as-a-distinct-partition-table-action

### What

Adds row selection to re_table and a context menu to partition tables to easily open multiple tables.

https://github.com/user-attachments/assets/87c5f487-ebfc-4de2-a4d2-b061bd1057f0

### DIFF
--- a/crates/viewer/re_dataframe_ui/src/table_selection.rs
+++ b/crates/viewer/re_dataframe_ui/src/table_selection.rs
@@ -26,7 +26,7 @@ impl TableSelectionState {
 
     pub fn clear(ctx: &egui::Context, id: Id) {
         ctx.data_mut(|data| {
-            data.remove::<TableSelectionState>(id);
+            data.remove::<Self>(id);
         });
     }
 


IsseW:
Screenshot 2025-11-17 at 16 36 19

I think when the upper left icon here is a `-` pressing it should deselect everything, but right now it will select everything. If @gavrelina has another opinion on this ignore me

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -695,33 +724,75 @@ impl egui_table::TableDelegate for DataFusionTableDelegate<'_> {
...
+                                == Some(&partition_links_spec.column_name)
+                        });
+
+                        // Let's open the tabs in order
+                        for row in selected_items.iter().copied().sorted() {
+                            if let Some((display_record_batch, batch_index)) =
+                                Self::with_row_batch(self.display_record_batches, row as usize)
+                            {
+                                if let Some(column) = column_index
+                                    .and_then(|col| display_record_batch.columns().get(col))
```
IsseW: Quite a lot of indentation happening here, could at least combine these if statements

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -695,33 +724,75 @@ impl egui_table::TableDelegate for DataFusionTableDelegate<'_> {
...
+                                {
+                                    match column {
+                                        DisplayColumn::RowId { .. }
+                                        | DisplayColumn::Timeline { .. } => {}
+                                        DisplayColumn::Component(col) => {
+                                            let string = col.string_value_at(batch_index);
+                                            if let Some(url) = string {
+                                                ui.ctx().open_url(OpenUrl::same_tab(url));
```
IsseW: Wonder if `new_tab` would be preferred here?

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -695,33 +724,75 @@ impl egui_table::TableDelegate for DataFusionTableDelegate<'_> {
...
+        }
+    }
 
-                break;
-            } else {
-                row_index -= row_count;
-            }
+    fn row_ui(&mut self, ui: &mut Ui, _row_nr: u64) {
+        let has_context_menu = self.blueprint.partition_links.is_some();
+        if has_context_menu {
```
IsseW: Could have early return here instead to reduce indenting a bit

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -695,33 +724,75 @@ impl egui_table::TableDelegate for DataFusionTableDelegate<'_> {
...
+            ui.response().container_context_menu(|ui| {
+                let selection = TableSelectionState::load(ui.ctx(), self.session_id);
+
+                // re_table will ensure that the right-clicked row is always selected.
+                let selected_items = selection.selected_rows;
+
+                if let Some(partition_links_spec) = &self.blueprint.partition_links {
```
IsseW:
```suggestion
                let Some(partition_links_spec) = &self.blueprint.partition_links else {
                	return;
                };
```

Could early return here as well.

crates/viewer/re_dataframe_ui/src/re_table.rs
```diff
@@ -120,7 +188,29 @@ impl egui_table::TableDelegate for ReTable<'_> {
...
+                    if response.changed() {
+                        // If the checkbox is clicked, the row will also detect the click.
+                        // Since we want the checkbox to have a different click behavior,
+                        // undo the row click:
+                        self.selection = self.previous_selection.clone();
+
+                        let mut modifiers = ui.input(|i| i.modifiers);
+                        if !modifiers.shift {
+                            // By default, the checkbox is like command clicking.
+                            modifiers = Modifiers::COMMAND;
```
IsseW: this feels a bit hacky, maybe could have a `checkbox_click` argument on `handle_row_click` instead to make it more explicit?

crates/viewer/re_ui/src/egui_ext/response_ext.rs
```diff
@@ -0,0 +1,67 @@
...
+pub trait ResponseExt {
+    fn _self(&self) -> &Response;
+
+    /// Is this response or any child widgets hovered?
+    ///
+    /// Will return false if some other area is covering the given layer, or if anything is being
+    /// dragged.
+    ///
+    /// This calls [`egui::Context::rect_contains_pointer`]. See also [`Response::hovered`].
+    fn container_hovered(&self) -> bool {
```
IsseW, line 13: Ah nice! I feel like I've done this manually in multiple places 😄

Cargo.toml
```diff
@@ -845,6 +845,9 @@ unnecessary_debug_formatting = "allow"
 # egui_tiles = { git = "https://github.com/rerun-io/egui_tiles", branch = "emilk/update-egui" }
 # egui_tiles = { path = "../egui_tiles" }
 
+egui_table = { git = "https://github.com/rerun-io/egui_table", branch = "lucas/add-row-ui" }
```
IsseW: Guessing we want the `lucas/add-row-ui merged before this?

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
lucasmerlin: I dunno, this feels dangerous if someone adds another context menu entry and misses the early return.

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
lucasmerlin: Good question. @gavrelina should we immediately switch to the first opened recording or should they be opened in the background?

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
lucasmerlin: Pulled out some fns to combat the indentation and added the early return there.

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
lucasmerlin: Thinking about it some more, I think we should stick to the convention. Cmd click -> new tab. Click -> Same tab.

lucasmerlin: Implemented this

pweids:
Found a few unexpected behaviors that I walk through here. @gavrelina would like your comments, some things to take a look at
https://drive.google.com/file/d/16FoWbW5ba-IBQ51uw_Re6FGsLcA_u3Xn/view?usp=sharing

lucasmerlin:
> Found a few unexpected behaviors that I walk through here. @gavrelina would like your comments, some things to take a look at

We decided to start without preserving the selection, as it adds extra complexity since then we will need some way to determine which column is the "primary key", and I don't believe we have that yet and I'm not even sure how that would work with custom tables. So for now it's just based on the row number. But we should of course clear the selection if the data changes, I'll fix that.

gavrelina:
@pweids @lucasmerlin did not check my gh updates.. hence this very late response..
Indeed, we discussed to not preserve the exact items for now — but current way of interaction – I agree — is very confusing. Instead we should remove the selection all together when user ascend/descend/remove filter/adds filter. 

And another point with losing the selection when right click elsewhere — I think the current one is not too bad! other option could be introduce the "active selection", make it similar to mac finder. But that one would need some extra states that we do not have yet in the table itself..

PR #11909, opened 2025-11-17
Simplify transform resolution cache's tracking based on new atomic-latest-at semantics

Wumpf:
### Related

* Fixes RR-2964

### What

* drop static exception for the "transform relation -> entity path pinning rule" (see .fbs doc update)
* no more entity-to-frame-tracking
* make clear tracking much more simple & precise (likely fixes some untested cases!)
* add test for first adding a chunk with a clear and then actual data it clears
    * haven't checked but I strongly suspect this was broken before

This makes stuff quite a bit faster! comparison of `cargo bench -p re_tf` against main line on my windows box:

```
build_from_entity_db    time:   [32.289 ms 32.945 ms 33.676 ms]
                        change: [-70.671% -69.621% -68.579%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe

query_uncached_frame    time:   [4.6857 ms 4.7923 ms 4.9035 ms]
                        change: [-62.224% -61.283% -60.414%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

query_cached_frame      time:   [27.953 ns 28.218 ns 28.601 ns]
                        change: [+6.9480% +10.017% +13.033%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 19 outliers among 100 measurements (19.00%)
  1 (1.00%) high mild
  18 (18.00%) high severe
```

Slower query is unfortunate but not all that significant in the big picture I believe. I'm a bit surprised by this, yes the logic has been rearranged, but there's less data to chugg there now and I expected the rearrangement to yield similar code in optimized builds, clearly that's not the case.

(these numbers are prior to landing https://github.com/rerun-io/rerun/pull/11911 which naturally an impact of the benchmarks but isn't really an optimization of anything in re_tf)

### DIFF
--- a/crates/store/re_tf/src/transform_resolution_cache.rs
+++ b/crates/store/re_tf/src/transform_resolution_cache.rs
@@ -17,7 +17,6 @@ use crate::{
 };
 
 use re_arrow_util::ArrowArrayDowncastRef as _;
-use re_chunk::ArrowArray as _;
 use re_chunk_store::external::arrow;
 use re_chunk_store::{Chunk, LatestAtQuery};
 use re_entity_db::EntityDb;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9677,6 +9677,7 @@ name = "re_tf"
 version = "0.28.0-alpha.1+dev"
 dependencies = [
  "ahash",
+ "arrow",
  "bitflags 2.9.4",
  "criterion",
  "glam",
@@ -9686,7 +9687,6 @@ dependencies = [
  "nohash-hasher",
  "parking_lot",
  "re_arrow_util",
- "re_chunk",
  "re_chunk_store",
  "re_entity_db",
  "re_log",
--- a/crates/store/re_tf/Cargo.toml
+++ b/crates/store/re_tf/Cargo.toml
@@ -21,7 +21,6 @@ all-features = true
 
 [dependencies]
 re_arrow_util.workspace = true
-re_chunk.workspace = true
 re_chunk_store.workspace = true
 re_entity_db.workspace = true # It would be nice not to depend on this, but we need this in order to do queries right now.
 re_log.workspace = true
@@ -30,6 +29,7 @@ re_tracing.workspace = true
 re_types = { workspace = true, features = ["glam"] }
 
 ahash.workspace = true
+arrow.workspace = true
 bitflags.workspace = true
 glam.workspace = true
 itertools.workspace = true


crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -2136,8 +1831,11 @@ mod tests {
 
     #[test]
     fn test_clear_non_recursive() -> Result<(), Box<dyn std::error::Error>> {
-        for clear_in_separate_chunk in [false, true] {
+        for (clear_in_separate_chunk, first_clear_then_data) in
+            [(false, false), (true, false), (true, true)]
+        {
             println!("clear_in_separate_chunk: {clear_in_separate_chunk}");
+            println!("first_clear_then_data: {clear_in_separate_chunk}");
```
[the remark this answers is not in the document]
Wumpf: nice catch lil bot! I'll bundle this with other comments later to reduce ci turn around

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -19,9 +16,11 @@ use crate::{
     },
 };
 
+use re_arrow_util::ArrowArrayDowncastRef as _;
+use re_chunk::ArrowArray as _;
```
grtlr: Should we depend on `arrow` directly?

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -548,22 +410,37 @@ pub struct ResolvedPinholeProjection {
...
+        associated_entity_path: EntityPath,
         child_frame: TransformFrameIdHash,
         _timeline: TimelineName,
         static_timeline: &mut CachedTransformsForTimeline,
+        non_recursive_clears: &IntMap<EntityPath, BTreeSet<TimeInt>>,
+        recursive_clears: &IntMap<EntityPath, BTreeSet<TimeInt>>,
     ) -> Self {
         let mut events = TransformsForChildFrameEvents::new_empty();
 
+        // Take over static events.
```
grtlr, line 421: I wonder about the sequencing here. Does that mean clears can overwrite static events?

crates/store/re_tf/src/transform_resolution_cache.rs
Wumpf: yeah why not, been cleaning up a few indirection here so why stop!

crates/store/re_tf/src/transform_resolution_cache.rs
Wumpf: yes, a temporal clear can "hide" a static event. Will add test for that!

Wumpf:
We talked about this a bit on Slack DM. So turns out my implementation and me are wrong here: one can't clear static data. Example:

```py
import rerun as rr

rr.init("rerun_example_points3d", spawn=True)
rr.set_time("timetime", duration=1.0)
rr.log("points", rr.Points3D([[0, 0, 0], [1, 1, 1]]), static=True)
rr.log("points", rr.LineStrips3D([[0, 0, 0], [1, 1, 1]]), static=False)
rr.set_time("timetime", duration=2.0)
rr.log("points", rr.Clear(recursive=False))
```

This will still show the points after t=2 but not the line.
Fixing this will require us to store a "static flag" every atomic set (transforms, instance poses, pinholes) which then makes us disregard all new temporal data

I'll fix this & test this in a separate iteration.

crates/store/re_tf/src/transform_resolution_cache.rs
Wumpf: that's what's happening, but it's incorrect, see comment below

PR #11911, opened 2025-11-17
Transform3D no longer sets all its components to empty array by default

Wumpf:
### Related

* Fixes RR-2940

### What

As described in
* #11901

we changed our query semantics such that this is no longer necessary. This greatly improves navigation of transforms int he Viewer & queries and has the potential to speed logging of transforms up significantly (especially SDK sided).

In other words.

Before:
image

After:
image

### DIFF
--- a/rerun_py/tests/unit/test_archetype_str.py
+++ b/rerun_py/tests/unit/test_archetype_str.py
@@ -8,7 +8,7 @@ import rerun as rr
     ["archetype", "expected"],
     [
         [
-            rr.Transform3D(),
+            rr.Transform3D().from_fields(clear_unset=True),
             (
                 "rr.Transform3D(\n"
                 "  translation=[],\n"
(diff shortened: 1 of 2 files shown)


crates/build/re_types_builder/src/lib.rs
```diff
@@ -181,7 +181,6 @@ pub const ATTR_ARROW_SPARSE_UNION: &str = "attr.arrow.sparse_union";
 pub const ATTR_RERUN_COMPONENT_OPTIONAL: &str = "attr.rerun.component_optional";
 pub const ATTR_RERUN_COMPONENT_RECOMMENDED: &str = "attr.rerun.component_recommended";
 pub const ATTR_RERUN_COMPONENT_REQUIRED: &str = "attr.rerun.component_required";
-pub const ATTR_RERUN_LOG_MISSING_AS_EMPTY: &str = "attr.rerun.log_missing_as_empty";
```
grtlr, line 184: Nice!

tests/assets/rrd/snippets/archetypes/coordinate_frame_builtin_frames.rrd
grtlr, line 1: Are we sure to update the snippets? Or should we add migration code?

crates/store/re_types/definitions/rerun/archetypes/instance_poses3d.fbs
grtlr, line 27: Makes me wonder: How did `InstancePose3D` get away without the missing attribute?

Wumpf: it didn't. We always accepted that this has different semantics there. Having the attribute in the first place was compromise nobody was happy with and somehow we arrived having it not there. I vaguely remember that being a conscious decision, but I can't summon the why....

tests/assets/rrd/snippets/archetypes/coordinate_frame_builtin_frames.rrd
Wumpf: as per Slack message: yeah because migration code would be complicated and brittle - nulling out all `[]` is only legal if _all_ transform components are missing.
Personally I also don't like it - the code IS producing different data now. Not in a viewer-breaking manner sure, but it's somewhat breaking

PR #11914, opened 2025-11-18
[JetBrain] Update watchers and remove/ignore rerun.iml

abey79:
### What

Fix the file watchers and remove/ignore `rerun.iml`


.idea/watcherTasks.xml
```diff
@@ -18,7 +18,27 @@
...
+      <option name="outputFilters">
+        <array />
+      </option>
+      <option name="outputFromStdout" value="false" />
+      <option name="program" value="pixi" />
+      <option name="runOnExternalChanges" value="true" />
+      <option name="scopeName" value="Project Files" />
+      <option name="trackOnlyRoot" value="false" />
+      <option name="workingDir" value="$ProjectFileDir$" />
       <envs />
```
Wumpf, line 42: looks like we have `run ruff format` twice now?

lucasmerlin:
Ooooh, so this is why my files disapper when switching branches?

Wumpf:
yes :(
caused by https://github.com/rerun-io/rerun/pull/11903
I didn't realize there's more of us and failed to announce, sorry!

.idea/watcherTasks.xml
abey79: yes, once for python file, once for python stub files (.pyi)

abey79: i haven't found a way to have a single watcher apply to both

Wumpf (approved):

.idea/watcherTasks.xml
Wumpf: ahhh makes sense

Issue #11912, opened 2025-11-17
Zooming out of double-clicked frustum resets the view


roym899:
**Describe the annoyance**
Before Rerun 0.27, double clicking a frustum, and then zooming out, would change the up-axis of the 3D view to the up-axis of the camera frustum. This was quite handy, because often images / cameras are somewhat upright; but the world might be oriented in some other (possibly unknown way).

In general it was possible to control the 3D camera after double clicking, which (to me) felt quite natural. Now, all controls are disabled and zooming out resets to the previous view.

Not sure if this was a deliberate change; if it was, it would be nice to keep both functionalities somehow (e.g., Ctrl+Z to restore the previous view; manipulating the 3D view should have the behavior from 0.26).

**To Reproduce**
Double click frustum, zoom out.
View resets to what it was before double clicking, instead of maintaining the new up-axis (zooming out from the camera view). 

**Expected behavior**
I preferred the old behavior. Now there is no way of quickly setting the up-axis of the world based on a camera in the world. Previously double clicking the 3D view already provided a way of resetting. Ctrl+Z works in the new version to restore the previous view.

IsseW:
Not having tracking an entity continuously update the view was intentional. But I agree that doing an input to cancel tracking should be done from the active view's perspective.

Got a fix in the works here https://github.com/rerun-io/rerun/pull/11915

PR #11915, opened 2025-11-18
Fix exiting tracking with inputs.

IsseW:
### Related

- Closes https://github.com/rerun-io/rerun/issues/11912

### What

Makes exiting tracking with inputs not return to the original camera location.

### DIFF
--- a/crates/viewer/re_view_spatial/src/eye.rs
+++ b/crates/viewer/re_view_spatial/src/eye.rs
@@ -743,7 +743,8 @@ impl EyeState {
                 self.last_tracked_entity = Some(tracking_entity.clone());
             }
 
-            let did_eye_change = match eye_controller.kind {
+            // If the position that doesn't move by changing the eye's rotation changed.
+            let did_eye_orbit_center_change = match eye_controller.kind {
                 Eye3DKind::FirstPerson => eye_controller.pos != old_pos,
                 Eye3DKind::Orbital => eye_controller.look_target != old_look_target,
             };
@@ -835,7 +836,7 @@ impl EyeState {
                     Some(eye_controller.pos.distance(eye_controller.look_target));
 
                 // When we stop tracking, set the blueprint eye state to the tracked view.
-                if eye_controller.did_interact && did_eye_change {
+                if eye_controller.did_interact && did_eye_orbit_center_change {
                     eye_controller.save_to_blueprint(
                         ctx.viewer_ctx,
                         eye_property,


crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -744,18 +753,22 @@ impl EyeState {
                     && (eye_controller.pos != old_pos
                         || eye_controller.look_target != old_look_target)
                 {
+                    // When we stop tracking, set the blueprint eye state to the tracked view.
```
MichaelGrupp, line 757: ~~The `if` conditions are duplicated? Or do I overlook something~~ 🤔 

`did_eye_change` derived from the `match` above does the same as `(eye_controller.pos != old_pos
|| eye_controller.look_target != old_look_target)`, so here it's better to write:

```rust
if eye_controller.did_interact && did_eye_change {
```

~~This however also means that your newly added `if eye_controller.did_interact && did_eye_change {` inside the `else` block is unreachable...?~~

~~See following comments....~~

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -744,18 +753,22 @@ impl EyeState {
...
                         EyeControls3D::descriptor_tracking_entity(),
                     );
                 } else {
                     return Some(target_eye);
                 }
-            } else if eye_controller.did_interact && did_eye_change {
-                eye_property.clear_blueprint_component(
-                    ctx.viewer_ctx,
-                    EyeControls3D::descriptor_tracking_entity(),
-                );
```
MichaelGrupp, line 758: ~~Removing this branch is correct as it's the same condition.~~

crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -821,7 +834,23 @@ impl EyeState {
                 self.last_orbit_radius =
                     Some(eye_controller.pos.distance(eye_controller.look_target));
 
-                return Some(eye_controller.get_eye());
+                // When we stop tracking, set the blueprint eye state to the tracked view.
+                if eye_controller.did_interact && did_eye_change {
```
MichaelGrupp: ~~Revert this, because we will never enter the `else` block when this is true?~~ 🤔

crates/viewer/re_view_spatial/src/eye.rs
MichaelGrupp: Sorry, scratch that. I looked at the wrong column. Nested ifs 😵‍💫

MichaelGrupp: You can still change it to `if eye_controller.did_interact && did_eye_change {` for readability here.

IsseW: `did_eye_change` and `(eye_controller.pos != old_pos || eye_controller.look_target != old_look_target)` are not always the same.

`did_eye_change` is `eye_controller.pos != old_pos` if it's first person. And `eye_controller.look_target != old_look_target` if it's an orbit camera.

Could maybe be better named though, `did_eye_orbit_center_change` maybe?

MichaelGrupp: Ah.. yes. A descriptive name would be good (or a comment).

MichaelGrupp: Thx, clearer now. In Orbital mode, an eye rotation doesn't move the look target, in FPV it doesn't move the eye position.

PR #11916, opened 2025-11-18
Introduce `creates_table` mark to the e2d redap test

abey79:
### Related

* workaround for https://linear.app/rerun/issue/RR-2969/add-default-storage-for-table
* unblocks https://github.com/rerun-io/dataplatform/pull/2029

### What

The above issue means that `create_table` needs to pass a URI for the table to be created. This s impossible to do for an arbitrary `REDAP_URL`, which the e2e test suite aims to target. As a workaround, this PR introduces a mark for all tests which create table, so they maybe disabled (using `-m "not creates_table"`, or more likely `-m not (creates_table or local_only)`)


abey79:
@rerun-bot full-check

abey79:
> Marker looks reasonable to me. Not digging into if there's some easier workaround for creating tables.

My opinion on that is that this project should be bubbled up quite high in the priority list:

https://linear.app/rerun/project/add-default-storage-option-for-created-tables-and-datasets-c47b7ebd5650

PR #11921, opened 2025-11-18
JIT compaction in ChunkStore can now split chunks in addition to merging them

ntjohnson1:
### Related
closes RR-2971

### What
If chunks are too big split them. Naive approach but helps with footguns made possible by send_columns (where everything is inserted into one chunk).

- [x] I should probably add tests to the chunk store now that we are splitting by default
- [x] By adding it to the chunk store does that mean send_columns will now get split by default at log time? Or does that not go through this path? (Confirmed locally that there is a different path)
- [x] @teh-cmc: Make sure the new snapshot sizes make sense


teh-cmc:
@rerun-bot full-check

ntjohnson1:
Hopefully for one last time

@rerun-bot full-check

crates/store/re_chunk_store/src/writes.rs
```diff
@@ -716,6 +730,70 @@ impl ChunkStore {
...
+        let mut cur_chunk_id = chunk.id();
+
+        while start_idx < chunk.num_rows() {
+            let remaining_rows = chunk.num_rows() - start_idx;
+            let chunk_size = remaining_rows.min(target_rows);
+
+            let split_chunk = chunk
+                .row_sliced(start_idx, chunk_size)
+                .with_id(cur_chunk_id);
+            cur_chunk_id = cur_chunk_id.next();
```
emilk, line 779: ⚠️  There is a latent `ChunkId` collision bug here!

Right now `decompact_chunk_if_needed` may be called _again_ on the previous output of `decompact_chunk_if_needed`. This can happen in "uneven" chunk, where only some rows are fat. Splitting such an uneven chunk can create pieces that are still above `chunk_max_bytes`.

emilk: * Fix in https://github.com/rerun-io/rerun/pull/12008

PR #11922, opened 2025-11-18
Limit record batch sizes during coalesce

timsaucer:
### Related

Closes https://linear.app/rerun/issue/RR-2925/querying-assetvideo-from-oss-server-crashes

### What

This resolves a problem of out of memory crashing python code. The current streaming will coalesce batches until they reach a specific size, 2048 rows. When querying rows that are very large, like video blobs, this causes an out of memory problem to happen.

This PR introduces a custom implementation of `CoalesceBatches` where we will emit batches when we either reach a memory size or number of rows. There is a [related upstream issue](https://github.com/apache/arrow-rs/issues/8850) which may more elegantly correct this problem, but the need to fix this is too urgent to wait for that upstream work.

This PR essentially copies the contents of `CoalesceBatches` and `CoalesceBatchesExec` from the datafusion repo and changes the logic to use either memory size or number of rows.

### DIFF
--- a/crates/store/re_datafusion/src/dataframe_query_common.rs
+++ b/crates/store/re_datafusion/src/dataframe_query_common.rs
@@ -1,3 +1,5 @@
+use crate::batch_coalescer::coalesce_exec::SizedCoalesceBatchesExec;
+use crate::batch_coalescer::coalescer::CoalescerOptions;
 use crate::pushdown_expressions::{apply_filter_expr_to_queries, filter_expr_is_supported};
 use ahash::HashSet;
 use arrow::array::{
@@ -15,8 +17,6 @@ use datafusion::datasource::TableType;
 use datafusion::logical_expr::{Expr, Operator, TableProviderFilterPushDown};
 use datafusion::physical_plan::ExecutionPlan;
 use futures::StreamExt as _;
-use crate::batch_coalescer::coalesce_exec::SizedCoalesceBatchesExec;
-use crate::batch_coalescer::coalescer::CoalescerOptions;
 use re_dataframe::external::re_chunk_store::ChunkStore;
 use re_dataframe::{Index, QueryExpression};
 use re_log_types::EntryId;
--- a/crates/utils/re_arrow_util/src/lib.rs
+++ b/crates/utils/re_arrow_util/src/lib.rs
@@ -22,8 +22,8 @@ pub use self::format::{
     format_record_batch_with_width,
 };
 pub use self::format_data_type::*;
-pub use self::test_extensions::*;
 pub use self::string_view::*;
+pub use self::test_extensions::*;
 
 /// Convert any `BinaryArray` to `LargeBinaryArray`, because we treat them logically the same
 pub fn widen_binary_arrays(list_array: &ListArray) -> ListArray {


timsaucer:
@rerun-bot full-check

crates/store/re_datafusion/src/dataframe_query_common.rs
```diff
@@ -14,46 +17,39 @@ use datafusion::common::{Column, DataFusionError, downcast_value, exec_datafusio
...
-use std::str::FromStr as _;
-use std::sync::Arc;
 
 /// Sets the size for output record batches in rows. The last batch will likely be smaller.
 /// The default for Data Fusion is 8192, which leads to a 256Kb record batch on average for
 /// rows with 32b of data. We are setting this lower as a reasonable first guess to avoid
 /// the pitfall of executing a single row at a time, but we will likely want to consider
 /// at some point moving to a dynamic sizing.
-const DEFAULT_BATCH_SIZE: usize = 2048;
+const DEFAULT_BATCH_BYTES: usize = 200 * 1024 * 1024;
```
emilk: nit: we usually measure bytes in u64. That way we don't accidentally confuse them with row-counts (usize) in function arguments etc

crates/store/re_datafusion/src/batch_coalescer/coalescer.rs
```diff
@@ -0,0 +1,605 @@
...
+    /// - `schema` - the schema of the output batches
+    /// - `target_batch_bytes` - the ideal maximum size in total bytes
+    /// - `target_batch_rows` - the minimum number of rows for each
+    ///   output batch (until limit reached)
+    /// - `fetch` - the maximum number of rows to fetch, `None` means fetch all rows
+    pub fn new(
+        schema: SchemaRef,
+        target_batch_bytes: usize,
+        target_batch_rows: usize,
+        fetch: Option<usize>,
```
emilk: Grouping these three into a `struct CalescerOptions` would have several benefits:

A) would allow "named" arguments to the function, making the calling of `new` more readable
B) would remove the risk of confusing one argument with another
C) would allow adding docstrings to each "parameter" individually
D) would reduce the number of [chunks](https://en.wikipedia.org/wiki/Chunking_(psychology)) inside of `SizedBatchCoalescer`

crates/store/re_datafusion/src/batch_coalescer/coalescer.rs
```diff
@@ -0,0 +1,605 @@
...
+    /// Checks if the buffer will reach the specified limit after getting
+    /// `batch`.
+    ///
+    /// If fetch would be exceeded, slices the received batch, updates the
+    /// buffer with it, and returns `true`.
+    ///
+    /// Otherwise: does nothing and returns `false`.
+    fn limit_reached(&mut self, batch: &RecordBatch) -> bool {
+        match self.fetch {
+            Some(fetch) if self.total_rows + batch.num_rows() >= fetch => {
```
emilk:
```suggestion
        if let Some(fetch) = self.fetch &&
            self.total_rows + batch.num_rows() >= fetch {
```

crates/store/re_datafusion/src/batch_coalescer/coalescer.rs
```diff
@@ -0,0 +1,605 @@
...
+    buffer: Vec<RecordBatch>,
+
+    /// Buffered size in bytes
+    buffered_bytes: usize,
+
+    /// Buffered row count
+    buffered_rows: usize,
+
+    /// Limit: maximum number of rows to fetch, `None` means fetch all rows
+    fetch: Option<usize>,
```
emilk: maybe this could be called something more explicit, like `max_rows`?

crates/store/re_datafusion/src/batch_coalescer/coalescer.rs
```diff
@@ -0,0 +1,605 @@
...
+    total_rows: usize,
+
+    /// Buffered batches
+    buffer: Vec<RecordBatch>,
+
+    /// Buffered size in bytes
+    buffered_bytes: usize,
+
+    /// Buffered row count
+    buffered_rows: usize,
```
emilk, line 71: these could also be grouped into one struct for less cognitive overhead. Keeping track of eight different states is a bit daunting, but with grouping it is reduced to four: the schema, the config, the buffer, and the total.

crates/store/re_datafusion/src/batch_coalescer/coalescer.rs
```diff
@@ -0,0 +1,605 @@
...
+    ///
+    /// Action: call [`SizedBatchCoalescer::finish_batch()`] to get the final
+    /// buffered results as a batch and finish the query.
+    LimitReached,
+
+    /// The specified minimum number of rows a batch should have is reached.
+    ///
+    /// Action: call [`SizedBatchCoalescer::finish_batch()`] to get the current
+    /// buffered results as a batch and then continue pushing batches.
+    TargetReached,
```
emilk: I feel the difference between Limit and Target are bigger than the words convey. I also find the ordering of them in the enum confusing (the final state should be last). Suggestion:


```suggestion
    /// The specified size limit of a single batch is reached.
    ///
    /// Action: call [`SizedBatchCoalescer::finish_batch()`] to get the current
    /// buffered results as a batch and then continue pushing batches.
    BatchFull,
    
    /// The end has been reached, as specified by [`CoalescerConfig::row_limit`].
    ///
    /// Action: call [`SizedBatchCoalescer::finish_batch()`] to get the final
    /// buffered results as a batch and finish the query.
    EndReached,
```

crates/store/re_datafusion/src/batch_coalescer/coalescer.rs
```diff
@@ -0,0 +1,605 @@
...
+        self.buffered_bytes = 0;
+        Ok(batch)
+    }
+}
+
+/// Indicates the state of the [`SizedBatchCoalescer`] buffer after the
+/// [`SizedBatchCoalescer::push_batch()`] operation.
+///
+/// The caller should take different actions, depending on the variant returned.
+pub enum CoalescerState {
```
emilk: nit: I would put this enum at the top of the file, as it is almost 'required reading' before one reads `SizedBatchCoalescer`

crates/store/re_datafusion/src/batch_coalescer/coalescer.rs
```diff
@@ -0,0 +1,605 @@
...
+/// If the average size of each view is larger than 32 bytes, we compact the array.
+///
+/// `StringViewArray` include pointers to buffer that hold the underlying data.
+/// One of the great benefits of `StringViewArray` is that many operations
+/// (e.g., `filter`) can be done without copying the underlying data.
+///
+/// However, after a while (e.g., after `FilterExec` or `HashJoinExec`) the
+/// `StringViewArray` may only refer to a small portion of the buffer,
+/// significantly increasing memory usage.
+fn gc_string_view_batch(batch: &RecordBatch) -> RecordBatch {
```
emilk: This seems like a nice helper that could go into `re_arrow_util`

crates/store/re_datafusion/src/batch_coalescer/coalescer.rs
```diff
@@ -0,0 +1,605 @@
...
+            for batch in input_batches {
+                match coalescer.push_batch(&batch) {
+                    CoalescerState::Continue => {}
+                    CoalescerState::LimitReached => {
+                        output_batches.push(coalescer.finish_batch().unwrap());
+                        break;
+                    }
+                    CoalescerState::TargetReached => {
+                        coalescer.buffered_rows = 0;
+                        coalescer.buffered_bytes = 0;
```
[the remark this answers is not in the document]
emilk: I agree

crates/store/re_datafusion/src/dataframe_query_common.rs
timsaucer: Changed.

timsaucer:
@rerun-bot reality-sync

PR #11924, opened 2025-11-19
MVP for a per-segment `using_index_values`

abey79:
### Related

### What

MVP for a per-segment `using_index_values`:
- use a `dict[str, IndexValuesLike]`
- yield a row per provided values (e.g. if a segment is missing from the `dict`, it is entirely skipped)

### DIFF
--- a/rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
+++ b/rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
@@ -4,6 +4,7 @@ import datetime
 from typing import TYPE_CHECKING
 
 import numpy as np
+from datafusion import col
 from inline_snapshot import snapshot as inline_snapshot
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
```diff
@@ -60,55 +61,97 @@ def test_dataframe_api_filter_partition_id(basic_dataset: DatasetEntry) -> None:
...
+                ],
+                dtype=np.datetime64,
+            ),
+            "complex_recording_1": np.array(
+                [
+                    datetime.datetime(2000, 1, 1, 0, 0, 2),
+                ],
+                dtype=np.datetime64,
+            ),
+            # complex_recording_2 unspecify, all rows are included
```
jleibs:
```suggestion
            # complex_recording_2 not included -- no rows are returned
```

jleibs (approved):
Nice.

The more I think about this the more I still think the Dataframe version of this is going to be quite important as welll.  The reason I believe this is because that's what lets us chain these operations.

e.g.
```
interventions = df.filter_contents("robot/alerts").reader("realtime").filter("robot/alerts:event" == "intervention").select("rerun_segment_id", "realtime")

df.filter_contents("robot/camera").reader("realtime", using_index_values=interventions)
```

abey79:
> Nice.
> 
> The more I think about this the more I still think the Dataframe version of this is going to be quite important as welll. The reason I believe this is because that's what lets us chain these operations.
> 
> e.g.
> 
> ```
> interventions = df.filter_contents("robot/alerts").reader("realtime").filter("robot/alerts:event" == "intervention").select("rerun_segment_id", "realtime")
> 
> df.filter_contents("robot/camera").reader("realtime", using_index_values=interventions)
> ```

I've added support for dataframe + a snippet. As I note in the code, your example above (and my snippet, which is based on it) feels like it's "misusing" `using_index_values`, since this is essentially a filter operation. Ideally we'd have a more fitting example, but that can certainly wait.

PR #11927, opened 2025-11-19
Add `Dataset[Entry|View].index_ranges()` to the API sandbox

abey79:
### Related

### What

Introduces `.index_ranges` to datasets and datasets view.

### DIFF
--- a/rerun_py/tests/api_sandbox/rerun_draft/catalog.py
+++ b/rerun_py/tests/api_sandbox/rerun_draft/catalog.py
@@ -287,7 +287,7 @@ class DatasetEntry(Entry):
         )
 
     def index_ranges(self, index: str | IndexColumnDescriptor) -> datafusion.DataFrame:
-        view = DatasetView(self._inner, LazyDatasetState())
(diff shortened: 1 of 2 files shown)


jleibs (commented):
I understand why you went with this, but not feeling like we've nailed the naming here (esp on them count columns). I'm somewhat inclined to leave the counts out for now until we figure this out.

For the index columns, I wonder if `timeline_name:min` / `timeline_name:max` would feel more consistent? 

Although I proposed the name statistics originally, would `.index_ranges()` be more concrete and obvious?

abey79:
> I understand why you went with this, but not feeling like we've nailed the naming here (esp on them count columns). I'm somewhat inclined to leave the counts out for now until we figure this out.

Sounds good. Done.

> For the index columns, I wonder if `timeline_name:min` / `timeline_name:max` would feel more consistent?

Feeling very unopinionated. I've made the change. 

> Although I proposed the name statistics originally, would `.index_ranges()` be more concrete and obvious?

Yeah I think it's a better naming. `statistics` convey the idea you might find stuff like means, median, etc. which is not what this is about.

PR #11650, opened 2025-10-24
Add documentation to dataframe query workflow and fix api rendering for bindings

ntjohnson1:
### Related

* Closes RR-2251

### What
* Primary task: first pass through query-data.md to give an overview of how to start and connect to OSS server and get to a dataframe with pointers to datafusion for follow on work
   * Will fill an issue to make this into a snippet/tutorial to make testable after we get a python object for the server
* Secondary task: Render docs for DataframeQueryView and lots of other bindings exposed through catalog
   *  This scope creeped a lot so I just did the manual class list will file a follow on ticket for proposal to fix this more globally
   1. Add lint codes so I can separate out  my ignores for pyclass_eq and pyclass_module requirements
      1. The lint.py changes (which aren't great since I'm not an experience parser writer but don't seem awful)
   3. Add a pyclass module requirement to specify `rerun_bindings.rerun_bindings` see https://github.com/rerun-io/rerun/pull/11268 for more context
      1. All the rust file changes 
   4. Fight mkdocs to render bindings properly
      1. The mkdocs related files
      2.  Adds the ability to generate our docs just off of the stubs without the shared object being present at all. NOTE: The docs actually look better this way, but I added some accommodation to make them look less awful if the shared object is present.

Filed an issue to fix all the other missing classes from our rendered api docs: https://linear.app/rerun/issue/RR-2766/cover-remaining-classes-in-python-api-rendered-docs
and an issue with a proposal to avoid this problem moving forward:
https://linear.app/rerun/issue/RR-2765/change-python-public-api-surface-pattern-to-support-easier-docs

PR #11928, opened 2025-11-19
Add lint to check if rust classes have __str__, fix or exclude all existing

ntjohnson1:
### Related
https://github.com/rerun-io/rerun/pull/11650

### What
I hate seeing `<rerun_bindings.rerun_bindings.DataFusionTable object at 0x10d4220b0>` when I print things.
* Updated our lint.py to use git for file discovery. This is ~1000x faster than the package walk. Don't have a way to confirm we aren't missing new things now 😬 if not correct by inspection I guess I could try doing a set diff and manually inspecting
* I also asked a robot to generate a new lint rule to check for __str__ in our pymethods
* Implemented a whole bunch and added no lint when I got tired of it

Here is my minimal repro script to exercise most code paths:
```python
import rerun as rr
client = rr.catalog.CatalogClient("rerun://sandbox.redap.rerun.io")
print(client)
print(client.table_entries())
dataset = client.get_dataset(name="droid:raw")
print(dataset)
print(dataset.partition_table())
print(dataset.list_indexes())
first_partition = dataset.partition_ids()[0]
qv = dataset.dataframe_query_view(index="real_time", contents="/observation/gripper_position").filter_partition_id(first_partition)
print(qv)
print(rr.RecordingStream("foo"))
print(rr.GrpcSink("rerun+http://127.0.0.1:9876/proxy"))
print(rr.FileSink("out.rerun"))
```

Before
```console
CatalogClient(rerun://sandbox.redap.rerun.io:443)
[Entry(Table, '__entries')]
Entry(Dataset, 'droid:raw')
<rerun_bindings.rerun_bindings.DataFusionTable object at 0x10d4220b0>
[<rerun_bindings.rerun_bindings.IndexingResult object at 0x10d1e5b00>]
<rerun_bindings.rerun_bindings.DataframeQueryView object at 0x10d0e7660>
<rerun.recording_stream.RecordingStream object at 0x103232cf0>
<rerun_bindings.rerun_bindings.GrpcSink object at 0x10cff4540>
<rerun_bindings.rerun_bindings.FileSink object at 0x10cff4540>
```

After ( a little verbose but definitely better than above and we can adjust as needed).

```console
CatalogClient(rerun://sandbox.redap.rerun.io:443)
[Entry(Table, '__entries')]
DatasetEntry(name='droid:raw', id='187514A5086F1F1D5906873e499c6436')
DataFusionTable(name='droid:raw_partition_table')
[IndexingResult(index='/camera/wrist/embedding:embeddings' on 'real_time': VectorIvfPq { num_sub_vectors: 16, metric: cosine })]
DataframeQueryView(
 dataset=DatasetEntry(name='droid:raw', id='187514A5086F1F1D5906873e499c6436'),
 query_expression=QueryExpression {
     view_contents: Some(
         ViewContentsSelector(
             {
                 /observation/gripper_position: None,
             },
         ),
     ),
     include_semantically_empty_columns: false,
     include_tombstone_columns: false,
     include_static_columns: Both,
     filtered_index: Some(
         "real_time",
     ),
     filtered_index_range: None,
     filtered_index_values: None,
     using_index_values: None,
     filtered_is_not_null: None,
     sparse_fill_strategy: None,
     selection: None,
 },
 partition_ids=["ILIAD_50aee79f_2023_07_12_20h_28m_36s"]
)
RecordingStream(Some(
    StoreInfo {
        store_id: StoreId(
            Recording,
            "foo",
            "935b03a47f834753b5329a70dc991556",
        ),
        cloned_from: None,
        store_source: PythonSdk(
            3.11.13,
        ),
        store_version: Some(
            CrateVersion {
                major: 0,
                minor: 28,
                patch: 0,
                meta: Some(
                    DevAlpha {
                        alpha: 1,
                        commit: None,
                    },
                ),
            },
        ),
        is_partial: false,
    },
))
GrpcSink(ProxyUri {
    origin: Origin {
        scheme: RerunHttp,
        host: Ipv4(
            127.0.0.1,
        ),
        port: 9876,
    },
})
FileSink("out.rerun")
```

### DIFF
--- a/rerun_py/src/catalog/dataframe_query.rs
+++ b/rerun_py/src/catalog/dataframe_query.rs
@@ -439,6 +439,18 @@ impl PyDataframeQueryView {
             self_.partition_ids.as_slice(),
         )
     }
+
+    pub fn __str__(&self, py: Python<'_>) -> String {
+        let dataset_str = PyDatasetEntry::__str__(self.dataset.borrow(py));
+        let query_expr_str = format!("{:#?}", self.query_expression);
+
+        let dataset_line = indent::indent_all_by(1, format!("dataset={dataset_str},"));
+        let query_line = indent::indent_all_by(1, format!("query_expression={query_expr_str},"));
+        let partition_line =
+            indent::indent_all_by(1, format!("partition_ids={:?}", self.partition_ids));
+
+        format!("DataframeQueryView(\n{dataset_line}\n{query_line}\n{partition_line}\n)")
+    }
 }
 
 impl PyDataframeQueryView {
--- a/rerun_py/src/catalog/dataframe_rendering.rs
+++ b/rerun_py/src/catalog/dataframe_rendering.rs
@@ -95,7 +95,7 @@ impl PyRerunHtmlTable {
     }
 }
 
-#[pymethods]
+#[pymethods] // NOLINT: ignore[py-mthd-str]
 impl PyRerunHtmlTable {
     #[new]
     #[pyo3(signature = (max_width=None, max_height=None))]
--- a/rerun_py/src/dataframe/schema.rs
+++ b/rerun_py/src/dataframe/schema.rs
@@ -18,7 +18,7 @@ pub struct SchemaIterator {
     iter: std::vec::IntoIter<PyObject>,
 }
 
-#[pymethods]
+#[pymethods] // NOLINT: ignore[py-mthd-str]
 impl SchemaIterator {
     fn __iter__(slf: PyRef<'_, Self>) -> PyRef<'_, Self> {
         slf
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10714,6 +10714,7 @@ dependencies = [
  "datafusion-ffi",
  "document-features",
  "futures",
+ "indent",
  "infer",
  "itertools 0.14.0",
  "jiff",
--- a/rerun_py/Cargo.toml
+++ b/rerun_py/Cargo.toml
@@ -80,6 +80,7 @@ crossbeam.workspace = true
 datafusion.workspace = true
 datafusion-ffi.workspace = true
 document-features.workspace = true
+indent.workspace = true
 futures.workspace = true
 itertools.workspace = true
 jiff.workspace = true
--- a/rerun_py/rerun_sdk/rerun/recording_stream.py
+++ b/rerun_py/rerun_sdk/rerun/recording_stream.py
@@ -1419,6 +1419,9 @@ class RecordingStream:
 
         send_columns(entity_path=entity_path, indexes=indexes, columns=columns, strict=strict, recording=self)
 
+    def __str__(self) -> str:
+        return str(self.inner)
+
 
 class BinaryStream:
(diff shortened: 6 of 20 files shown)


rerun_py/src/catalog/dataframe_query.rs
```diff
@@ -439,6 +439,18 @@ impl PyDataframeQueryView {
...
+
+    pub fn __str__(&self, py: Python<'_>) -> String {
+        let dataset_str = PyDatasetEntry::__str__(self.dataset.borrow(py));
+        let query_expr_str = format!("{:#?}", self.query_expression);
+
+        let dataset_line = indent::indent_all_by(1, format!("dataset={dataset_str},"));
+        let query_line = indent::indent_all_by(1, format!("query_expression={query_expr_str},"));
+        let partition_line =
+            indent::indent_all_by(1, format!("partition_ids={:?}", self.partition_ids));
+
```
abey79, line 451: I wonder if we shouldn't leave it to `pprint` to do the indent + multiline. Genuine question, I'm not sure what is more "right".

rerun_py/src/catalog/dataframe_rendering.rs
```diff
@@ -95,7 +95,7 @@ impl PyRerunHtmlTable {
     }
 }
 
-#[pymethods]
+#[pymethods] // NOLINT: ignore[py-mthd-str]
```
abey79, line 98: fancy! :)

rerun_py/src/dataframe/schema.rs
```diff
@@ -18,7 +18,7 @@ pub struct SchemaIterator {
     iter: std::vec::IntoIter<PyObject>,
 }
 
-#[pymethods]
+#[pymethods] // NOLINT: ignore[py-mthd-str]
```
abey79, line 21: worth adding a `TODO` for those you feel would benefit from a   `__str__`?

rerun_py/src/catalog/dataframe_query.rs
ntjohnson1: Can you elaborate or point to an example? Formatted string mangling is something I am very much still figuring out for rust. No strong opinions yet on the right way.

abey79: My thought comes from a logging context, where I generally dislike prints to be more than one line. E.g. it interferes with the timestamp prefix, etc.  Now after trying `pprint` a bit, I realise that it doesn't work at all like I expected (e..g. it won't wrap/indent the repr in the form of `MyObject(key="value", ...)`. It only does so when `list` and other containers are involved. So basically let's ignore that for now.

ntjohnson1: Consider it ignored. But I'm still interested if we have general best practices around rust string formatting.

rerun_py/src/dataframe/schema.rs
ntjohnson1: I did another pass. I think everything user facing should have one (hence the lint). All of the remaining exclusions either have a note, are internal, probably should be deprecated, or I wasn't sure how they're even used.

So at the moment I don't think there is anything obvious that is high priority to track.

Issue #11929, opened 2025-11-20
Transforms / pinhole not always correctly showing up in 3D view (since 0.26)


roym899:
**Describe the bug**
Sometimes (and I don't know what triggers this) frustums or transforms are not correctly visualized in the 3D view. I can see the data changing and being available in the timeline / right-side panel. But the view shows the frustum frozen at a prior timestamp.

When it's a moving camera, this bug manifests itself as the frustum freezing for some consecutive timesteps, and moving again. To give a sense of scale, I log a few hundred images with different timestamps (in order), but the frustum does not move for some significant section of it.

When it's multiple pinholes at the same time, this bug (or another bug?) shows up as some images not having a frustum, and instead show up as a huge billboard; basically breaking the 3D view. Again, it's the same code logging multiple pinholes, but some do not show up correctly. So there's no obvious reason on the logging side that might cause this. Rerunning the code twice will not show the same issue. Sometimes all pinholes show up fine, sometimes another random subset is broken.

Has anything changed from 0.25 to 0.26 that might cause what I'm observing?

**To Reproduce**
Unfortunately, I could not find a way to reproduce this issue reliably. But it's been stopping me from updating the viewer to anything >=0.26. 

The same code, sometimes produces this issue and sometimes it doesn't.

**Desktop (please complete the following information):**
Logging happens on a remote Ubuntu machine; visualization on MacOS (using `connect_grpc(...)`).

**Rerun version**
This happens for both 0.26 and 0.27. The same code works fine with 0.25.

Only the viewer version affects this. If I log from 0.26 / 0.27 to 0.25 things work; but not when logging to anything >= 0.26.

Wumpf:
Thanks for reporting this!! We have a lot of changes ongoing around transforms right now, in fact I'm working on things related to the frustum visualization right now. I'll keep this in mind and be on the lookout but not having seen the issue myself yet makes it really hard to figure out.
Since you mention this is spurious I reckon that it has to do something with how (timing/ordering/chunking etc.) data arrives in the viewer. Can you maybe share an rrd from the viewer's broken state?

roym899:
I don't have one at hand; I'll try to get one. But it might take a few days. I'm also not sure if saving an rrd / loading it again reproduces the problem or not.

Wumpf:
[...]

PR #11932, opened 2025-11-20
Fix transform tree walk sometimes ignoring pinhole-only connections

Wumpf:
### Related

* MAYBE addresses https://github.com/rerun-io/rerun/issues/11929

### What

Couldn't prove that this is causing any problems with existing transforms. As far as I can tell it isn't because we always fall back to identity connection within the entity graph. However, it *is* causing issues with the upcoming parent/child frame options on Pinhole. That (upcoming) branch also adds a unit test that runs reliably into problems without this fix. The extended test here does unfortunately not repro the problem, but still felt useful to me.

(-> exclude from changelog until I can prove this actually fixes a regression and not just a new feature)

### DIFF
--- a/crates/store/re_tf/src/transform_forest.rs
+++ b/crates/store/re_tf/src/transform_forest.rs
@@ -412,8 +412,7 @@ fn walk_towards_parent(
         && unprocessed_frames.remove(&current_frame)
     {
         // We either already processed this frame, or we reached the end of our path if this source is not in the list of unprocessed frames.
-        let transforms =
-            transforms_at(current_frame, entity_db, query, id_registry, transforms);
+        let transforms = transforms_at(current_frame, entity_db, query, id_registry, transforms);
         next_frame = transforms.parent_frame;
 
         // No matter whether there's a next frame or not, we push the transform information we got about this frame onto the stack
@@ -752,7 +751,7 @@ fn transforms_at(
     let mut parent_from_child;
     let child_from_instance_poses;
     let pinhole_projection;
-    
+
     if let Some(source_transforms) = transforms_for_timeline.frame_transforms(child_frame) {
         parent_from_child = source_transforms.latest_at_transform(entity_db, query);
         child_from_instance_poses = source_transforms.latest_at_instance_poses(entity_db, query);


crates/store/re_tf/src/transform_forest.rs
```diff
@@ -405,19 +412,22 @@ fn walk_towards_parent(
         && unprocessed_frames.remove(&current_frame)
     {
         // We either already processed this frame, or we reached the end of our path if this source is not in the list of unprocessed frames.
-        let mut transforms = transforms_at(current_frame, entity_db, query, transforms);
+        let mut transforms =
+            transforms_at(current_frame, entity_db, query, id_registry, transforms);
 
         // Maybe there's an implicit connection that we have to fill in?
         if transforms.parent_from_child.is_none()
+            && transforms.pinhole_projection.is_none()
```
Wumpf:
```suggestion
if transforms.parent_frame.is_none()
```

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -277,31 +279,29 @@ impl TransformForest {
...
                 // So if we don't add this root now to our `root_from_frame` map, we'd never fill out the required self-reference!
-                self.root_from_frame.insert(
-                    parent_from_child.parent,
-                    TransformInfo::new_root(parent_from_child.parent),
-                );
+                self.root_from_frame
+                    .insert(parent_frame, TransformInfo::new_root(parent_frame));
 
-                (parent_from_child.parent, glam::DAffine3::IDENTITY)
+                (parent_frame, glam::DAffine3::IDENTITY)
```
Wumpf, line 304: here we always use the new `parent_frame` field of `ParentChildTransforms` so that we don't get it wrong!

crates/store/re_tf/src/transform_forest.rs
```diff
@@ -738,6 +734,7 @@ fn pinhole3d_from_image_plane(
 }
 
 struct ParentChildTransforms {
```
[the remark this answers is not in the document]
Wumpf: nah, it's fine

PR #11933, opened 2025-11-20
Add shortcuts to navigate the timeline

lucasmerlin:
### Related

* closes https://linear.app/rerun/issue/RR-2391/faster-timeline-navigation-with-shortcuts

This adds shortcuts to navigate the timeline.

The playback speed chord is implemented as a UICommand, so that it is discoverable via the CommandPalette. It's a bit odd since it has a value,  but I think it works quite nicely.

Gaps aren't handled yet since that would require a bigger refactor of how we handle them. I've opened an issue here: https://linear.app/rerun/issue/RR-3023/correctly-handle-timeline-gaps

The hack to override eguis built in shortcuts are quite bad, I intend to fix this with the next egui release.

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -359,6 +359,8 @@ impl App {
                             UICommand::PlaybackEnd,
                             UICommand::PlaybackForwardFast,
                             UICommand::PlaybackBackFast,
+                            UICommand::PlaybackStepForward,
+                            UICommand::PlaybackStepBack,
                             UICommand::PlaybackForward,
                             UICommand::PlaybackBack,
                         ];


IsseW:
I think using arrow keys should clamp to start/end before looping around, if you do `shift + arrow right` on dna now it skips to the start. And skips to end for `shift + arrow left`.

Other than that from testing I think this is great!

crates/viewer/re_ui/src/command.rs
```diff
@@ -498,6 +556,100 @@ impl UICommand {
...
+
+            // Leading zeros should divide the speed by 10 for each zero.
+            // So e.g. 05 = 0.5x speed, 005 = 0.05x speed, etc.
+            let leading_zeros = chord_state
+                .accumulated
+                .chars()
+                .take_while(|&c| c == '0')
+                .count();
+
+            let factor = if leading_zeros > 0 {
```
IsseW: This could also just be `10usize.pow(leading_zeros as u32)`, pow already has a fast-path for exponent == 0

IsseW (approved):
Looks good!

emilk:
⚠️ When editing text, `cmd+arrows` will move the time cursor instead of moving the text cursor.
`shift+arrows` works correctly though

lucasmerlin:
> ⚠️ When editing text, `cmd+arrows` will move the time cursor instead of moving the text cursor. `shift+arrows` works correctly though

Oh, didn't realized that the pre-existing commands had the same problem. I've applied the same fix for them!

PR #11936, opened 2025-11-21
Order partitions by opening order

IsseW:
### Related

- Closes RR-2340
- Closes https://github.com/rerun-io/rerun/issues/9848

### What

Makes partitions order by opening order in the recordings panel, instead of loaded order.

### Details

This adds a static incrementing counter that's incremented for each new smart channel receiver. Which is then wrapped in a new-type struct which we can use to order receivers.

### DIFF
--- a/crates/viewer/re_recording_panel/src/data.rs
+++ b/crates/viewer/re_recording_panel/src/data.rs
@@ -376,9 +376,9 @@ impl<'a> ServerEntriesData<'a> {
                                 .collect();
 
                             if let Some(loading_partitions) = loading_partitions
-                                && let Some(sources) = loading_partitions.get(&entry.id())
+                                && let Some(smart_channels) = loading_partitions.get(&entry.id())
                             {
-                                displayed_partitions.extend(sources.iter().map(|source| {
+                                displayed_partitions.extend(smart_channels.iter().map(|source| {
                                     PartitionData::Loading {
                                         receiver: source.clone(),
                                     }
--- a/crates/viewer/re_recording_panel/src/recording_panel_ui.rs
+++ b/crates/viewer/re_recording_panel/src/recording_panel_ui.rs
@@ -427,9 +427,7 @@ fn dataset_entry_ui(
             .show_hierarchical_with_children(ui, id, true, list_item_content, |ui| {
                 for partition in displayed_partitions {
                     match partition {
-                        PartitionData::Loading { receiver, .. } => {
-                            receiver_ui(ctx, ui, receiver, true);
-                        }
+                        PartitionData::Loading { receiver } => receiver_ui(ctx, ui, receiver, true),
 
                         PartitionData::Loaded { entity_db } => {
                             let include_app_id = false; // we already show it in the parent item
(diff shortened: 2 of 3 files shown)


crates/utils/re_smart_channel/src/receiver.rs
```diff
@@ -18,14 +27,22 @@ impl<T: Send> Receiver<T> {
         stats: Arc<SharedStats>,
         source: Arc<SmartChannelSource>,
     ) -> Self {
+        static ORDER_COUNTER: AtomicU64 = AtomicU64::new(1);
```
IsseW: I don't think having this as a static is a problem since the only requirement is having a counter where the previous is smaller than the next.

But maybe we'd want to store this at a higher level in some way?

crates/store/re_entity_db/src/entity_db.rs
```diff
@@ -114,6 +114,9 @@ pub struct EntityDb {
     storage_engine: StorageEngine,
 
     stats: IngestionStatistics,
+
+    /// The order this should have with respect to other [`EntityDb`].
+    added_order: ReceiverAddOrder,
```
Wumpf: I'd rather not have this on entity db since we use that in places without any viewer awareness. order of how we go the store to know should live somewhere else imho

crates/utils/re_smart_channel/src/receive_set.rs
```diff
@@ -81,6 +81,18 @@ impl<T: Send> ReceiveSet<T> {
...
 
+    /// List of connected receiver sources.
+    ///
+    /// This gets culled after calling one of the `recv` methods.
+    pub fn sources_with_order(&self) -> Vec<(ReceiverAddOrder, Arc<SmartChannelSource>)> {
+        re_tracing::profile_function!();
+        let mut rx = self.receivers.lock();
+        rx.retain(|r| r.is_connected());
+        rx.iter()
+            .map(|r| (r.added_order(), r.source.clone()))
```
Wumpf: shouldn't the order be just the order within the vec, i.e. `enumerate`?

Wumpf:
realized only now that 
* #11940

was targeting this

crates/store/re_entity_db/src/entity_db.rs
IsseW: Put it at a higher level now, in the store hub, and it's a much simpler change now.

Wumpf:
does not close https://github.com/rerun-io/rerun/issues/9848 in its current form, right?

IsseW:
> does not close https://github.com/rerun-io/rerun/issues/9848 in its current form, right?

The change [here](https://github.com/rerun-io/rerun/pull/11936/files#diff-44bb05331f29327e5f731fae98364513cce0abb21357d2bcdd07ee3b58264c2dR279) should do that by clearing the active recording.

Wumpf:
ah, nice! I admittedly missed that. also realizing only now how dubious the previous code was before.
```rs
            if recording.is_none() && self.active_recording_or_table.is_none() {
                self.active_recording_or_table = None;
            }
```

Issue #11937, opened 2025-11-21
0.27.2 "Visible time range" override for time series views is broken


flightless-coder:
**Describe the bug**
I upgraded from 0.25.1 of the viewer to 0.27.2 and the visible timerange override now seems to be broken:

![Image](https://github.com/user-attachments/assets/e26da38a-3043-4a7d-a237-4af82eec4243)

**To Reproduce**
1. Plot a 2d data series
2. Try to use the "override visible time range" feature in the viewer.

**Expected behavior**
The time range of the 2d view should focus on the described range, like it did in 0.25.1

**Desktop:**
Windows 11

**Rerun version**
0.27.2

IsseW:
You can use the `TimeAxis` `View range` above for this now!

The data outside the "VisibleTimeRange" still being visible might be a bug though.

flightless-coder:
@IsseW Thanks so much! Yes that works. 
Then I guess yeah, the old one should probably be removed/respec:ed to avoid confusion 😄

IsseW:
The contextual meaning of them is a bit different, `Visible time range` exists for many views and controls what data should be shown in the view.

While this new one just controls the time axis range of the time series view.

flightless-coder:
Oh okay, that makes sense. But then it's still a bug that the data is visible outside the set time range I guess? 

Btw, I love the feature of having the adaptable time range in the plot. So great work with that!

PR #11938, opened 2025-11-21
Fix the visible time range for the time series view

IsseW:
### Related

* Closes https://github.com/rerun-io/rerun/issues/11937

### What

Fixes the visible time range for the time series view.

### DIFF
--- a/crates/viewer/re_view_time_series/tests/blueprint.rs
+++ b/crates/viewer/re_view_time_series/tests/blueprint.rs
@@ -151,6 +151,7 @@ pub fn test_custom_visible_time_range() {
 }
 
 const MAX_TIME: i64 = 31;
+
 fn log_data(test_context: &mut TestContext, timeline: re_log_types::Timeline) {
     for i in 0..=MAX_TIME {
         let timepoint = TimePoint::from([(timeline, i)]);


crates/viewer/re_view_time_series/src/util.rs
```diff
@@ -58,22 +86,14 @@ pub fn determine_time_range(
...
-        .map(|range| {
-            AbsoluteTimeRange::new(
-                range.start.start_boundary_time(current_time),
-                range.end.end_boundary_time(current_time),
-            )
-        })
-        // If we don't have an overridden time range, we want to show everything.
-        .unwrap_or(AbsoluteTimeRange::EVERYTHING))
+        .intersection(data_time_range)
+        .unwrap_or(AbsoluteTimeRange::EMPTY))
```
Wumpf, line 96: We don't usually do intersection here. In fact the data time range is meant to override the view's time range. Plus afaik data time range is set to the view's range already

See https://github.com/rerun-io/rerun/blob/604e5897bbfa75de676a8151e58062d29ebd30f7/crates/viewer/re_viewer_context/src/view/view_query.rs#L63

IsseW: Hmm we did do this before as well, but with plot memory. https://github.com/rerun-io/rerun/blob/release-0.25.1/crates/viewer/re_view_time_series/src/util.rs#L57

Think that's probably why I did this mistake to begin with.

I guess it is an optimization though to only process scalars we can actually see?

Wumpf: ah. Yes. You're right, I got stuff mixed up. As an optimization we did the intersection to the visible range 💡 
I got confused there what range is what.

Wumpf (approved):
Thanks for those thorough tests and well spotted on my false alarm there!! <3

PR #11940, opened 2025-11-21
Fix re-opening recording

IsseW:
### Related

- Fixes RR-2990
- Not necessarily dependent on #11936, but the fix for closing loading recordings in it was needed to test this.

### What

Fixes a bug where opening multiple recordings with the same app id, and re-opening one of them didn't work.

### DIFF
--- a/crates/viewer/re_viewer_context/src/store_hub.rs
+++ b/crates/viewer/re_viewer_context/src/store_hub.rs
@@ -322,6 +322,7 @@ impl StoreHub {
         let removed_store = self.store_bundle.remove(store_id);
 
         let Some(removed_store) = removed_store else {
+            re_log::error!("Failed to remove {store_id:?}");
             return;
         };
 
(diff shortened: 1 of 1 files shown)


Wumpf (approved):
makes you wonder how this ever worked.
I'd love to have a regression test for this, but don't want to hold it up over that :/

crates/viewer/re_viewer_context/src/store_hub.rs
```diff
@@ -334,10 +335,16 @@ impl StoreHub {
...
 
+                // If this is the only recording with this app id, remove the app.
                 if !any_other_recordings_for_this_app {
                     re_log::trace!("Removed last recording of {app_id}. Closing app.");
                     self.close_app(app_id);
                 }
+                // Otherwise remove stores with the same data source we're closing.
+                else if removed_store.data_source.is_some() {
+                    self.store_bundle
+                        .retain(|r| r.data_source != removed_store.data_source);
```
Wumpf, line 346: actually, why does `store_bundle.remove` not already do this work 🤔

Wumpf (changes requested):
on second thought, are we sure this makes sense?

PR #11944, opened 2025-11-21
Load URDF file into the current recording, if one exists already

MichaelGrupp:
Previous behavior: if you dragged the URDF file into the viewer, it created its own new recording. Same for "Import into current recording" menu (note: requires #11941), which is counter-intuitive.

A recording containing only an URDF is pretty useless, and it's more common that you have some recording and want to add the model to it.

Loosely related to https://linear.app/rerun/issue/RR-2995


MichaelGrupp:
`cargo run -p animated_urdf` that logs URDF via SDK works still fine with this (ran it with an open viewer that had already a recording, and the logging example created a new one as desired).

oxkitsune (approved):

Issue #11902, opened 2025-11-14
Name Recordings


roym899:
I often run experiments with slightly different configurations under the same application id.

Right now the only well to tell these apart in the recordings panel is by their timestamp.

<img width="398" height="316" alt="Image" src="https://github.com/user-attachments/assets/298c4245-05be-48d8-b933-3fc4ba0c6a72" />

It is possible to assign a human readable recording_id, but that is only visible on hover; so not very useful to quickly pick a recording I'm interested in.

**Describe the solution you'd like**
Either show the recording id (or the beginning of it) in recordings panel or allow setting an optional recording name that is displayed in addition or instead of the recording time.

**Describe alternatives you've considered**
One could also change the application id, but that would mean the blueprint does not apply to all. Especially when comparing different settings, it is useful when it does.

Wumpf:
Happy to see you still around here 😄 
You can do so with [`send_recording_name`](https://ref.rerun.io/docs/python/0.27.2/common/initialization_functions/#rerun.RecordingStream), check this snippet: https://github.com/rerun-io/rerun/blob/7ac9d608c483c5a5c58e883ad9a0cecc3881f82c/docs/snippets/all/concepts/recording_properties.py#L9

I think that one should show up directly, if not that's a bug

Was a bit hard to find in the docs even for me knowing it's there 🤔  cc: @ntjohnson1

roym899:
This works! Thank you. Funny it doesn't show up when searching for 'recording name'.

ntjohnson1:
Ugh 😠 

If we search from the landing page on the website our full text search works

<img width="690" height="650" alt="Image" src="https://github.com/user-attachments/assets/f316f7fd-51c7-4fc8-8783-645a1bbe2011" />

If we search on the python api docs we don't appear to do good fuzzy search

<img width="759" height="358" alt="Image" src="https://github.com/user-attachments/assets/bcaf02bf-6a09-4049-b661-11cc6c9786a4" />

<img width="759" height="85" alt="Image" src="https://github.com/user-attachments/assets/0a836b39-93e8-4545-bd4e-1b3467096f1f" />

ntjohnson1:
FYI just added a PR that fixes this specific instance. (I don't think it will go live until our next release 0.28). If you see something in the future that's weird like this feel free to ping me. I think this PR helps but there are more drastic measures if our search is still insufficient. Thanks for calling out this one!

PR #11946, opened 2025-11-21
Make python api doc search match more things

ntjohnson1:
### Related

#11902

### What

Our python search by default looked for perfect matches which is 😠. This does a lazy improvement to at least split things along various characters to make it easier to match approximately. There are other features we could continue to tweak if necessary (or could overhaul with a better external search support index).

I still don't really know how to get our python api docs to update besides releasing a new version so run `pixi run py-docs-serve` locally and see that we now math the motivating issue (or trust me).

I also added a share link so I can refer to searches in the future without screenshots.

Screenshot 2025-11-21 at 2 32 24 PM

### DIFF
--- a/rerun_py/mkdocs.yml
+++ b/rerun_py/mkdocs.yml
@@ -19,6 +19,9 @@ theme:
 
 plugins:
   - search: # https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-search/
+      # Copy-pasted from https://squidfunk.github.io/mkdocs-material/plugins/search/#config.separator
+      # with the addition of underscore. Splits things for better partial matching.
+      # Further exploration of the parts at the link.
       separator: '[\s\-,:!=\[\]()"/_]+|(?!\b)(?=[A-Z][a-z])|\.(?!\d)|&[lg]t;'
   - autorefs
   - mkdocstrings: # https://mkdocstrings.github.io/usage/#global-options


rerun_py/mkdocs.yml
```diff
@@ -15,9 +15,11 @@ theme:
     - navigation.tabs
     - navigation.tabs.sticky
     - navigation.tracking
+    - search.share
 
 plugins:
-  - search # https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-search/
+  - search: # https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-search/
+      separator: '[\s\-,:!=\[\]()"/_]+|(?!\b)(?=[A-Z][a-z])|\.(?!\d)|&[lg]t;'
```
emilk, line 25: Of course, how simple!

emilk, line 25:
```suggestion
      separator: '[\s\-,:!=\[\]()"/_]+|(?!\b)(?=[A-Z][a-z])|\.(?!\d)|&[lg]t;' # Don't just match whole-words
```

maybe?

or just add a link to explain that monstrosity

emilk (approved):

rerun_py/mkdocs.yml
ntjohnson1: I copied that blob from the mkdocs-material page and inserted the `_` to the separator list. Adding a comment to that effect.

PR #11947, opened 2025-11-22
Use different color for `null` values

emilk:


.gitignore
```diff
@@ -54,6 +54,7 @@ perf.data*
 
 # insta snapshot temp file:
 *.pending-snap
+*.snap.new
```
grtlr: Can we leave out that line please? 
```suggestion
```

It's quite useful for me to see info about snapshot related failures in Git tools.

PR #11948, opened 2025-11-22
Fix sharing links to time ranges on duration timelines

emilk:
This was severely broken because of failure to parse URLs with `+` in them (e.g. `+1.2s`). I've now fixed the problem(s) and added several tests.

On my way to fixing this I also attacked a bunch of places where we `debug` logged some very verbose stuff, that is now actually readable instead.

See commit messages for details.


crates/utils/re_uri/src/time_selection.rs
```diff
@@ -61,9 +61,72 @@ impl std::str::FromStr for TimeSelection {
...
+                TimeSelection {
+                    timeline: Timeline::new_duration("duration"),
+                    range: AbsoluteTimeRange {
+                        min: TimeInt::from_secs(-1.096),
+                        max: TimeInt::from_secs(2.097),
+                    },
+                },
+            ),
+            (
+                "duration@−1.096s..+2.097s",
```
Wumpf: had to use some website to identify this as the special minus character. plz add comment that this is different than the above because in gh's font I can't tell the difference :)

crates/utils/re_uri/src/endpoints/dataset.rs
```diff
@@ -64,6 +64,7 @@ impl DatasetPartitionUri {
                     partition_id = Some(value.to_string());
                 }
                 "time_range" => {
+                    let value = value.replace(' ', ""); // Seems like "+" becomes " " when parsed by `url::Url`
```
Wumpf: huh, is that a bug in `url::Url`? I would have expected some kind of error or escaping going on

crates/viewer/re_viewer_context/src/open_url.rs
```diff
@@ -1209,4 +1209,72 @@ mod tests {
...
+                    origin: "rerun+http://localhost:51234".parse().unwrap(),
+                }),
+            ),
+            // (
+            //     "rerun+http://localhost:51234/dataset/187A3200CAE4DD795748a7ad187e21a3",
+            //     ViewerOpenUrl::RedapEntry(re_uri::EntryUri {
+            //         origin: "rerun+http://localhost:51234".parse().unwrap(),
+            //         entry_id: "187A3200CAE4DD795748a7ad187e21a3".parse().unwrap(),
+            //     }),
+            // ),
```
Wumpf: what happened here?

PR #11949, opened 2025-11-22
Add food-for-thought around new query API

emilk:
Some random notes from a sleep-deprived discussion with @nikolausWest 

May be useful as a basis for a huddle. Definitely not ready for review.

### DIFF
--- a/rerun_py/tests/api_sandbox/rerun_draft/catalog.py
+++ b/rerun_py/tests/api_sandbox/rerun_draft/catalog.py
@@ -70,6 +70,8 @@ class CatalogClient:
         """Returns a DataFrame containing all table entries in the catalog."""
         return self._inner.tables()
 
+    # Why are _some_ functions prefixed with `get_` and not others?
+    # Is the logic "if it takes an argument"? Maybe this is Pythonic, IDK
     def get_dataset_entry(self, *, id: EntryId | str | None = None, name: str | None = None) -> DatasetEntry:
         """Returns a dataset entry by its ID or name."""
         return DatasetEntry(self._inner.get_dataset_entry(id=id, name=name))
@@ -210,6 +212,7 @@ class DatasetEntry(Entry):
     def __init__(self, inner: _catalog.DatasetEntry) -> None:
         self._inner = inner
 
+    # What is this? What is it for?
     @property
     def manifest_url(self) -> str:
         return self._inner.manifest_url
@@ -217,18 +220,22 @@ class DatasetEntry(Entry):
     def arrow_schema(self) -> pa.Schema:
         return self._inner.arrow_schema()
 
-    def blueprint_dataset_id(self) -> EntryId | None:
-        return self._inner.blueprint_dataset_id()
+    # Remove these blueprint stuff, and just have
+    def set_default_blueprint(self, blueprint: Blueprint | Path) -> None:
+        pass
 
-    def blueprint_dataset(self) -> DatasetEntry | None:
-        result = self._inner.blueprint_dataset()
-        return DatasetEntry(result) if result is not None else None
+    # def blueprint_dataset_id(self) -> EntryId | None:
+    #     return self._inner.blueprint_dataset_id()
 
-    def default_blueprint_segment_id(self) -> str | None:
-        return self._inner.default_blueprint_partition_id()
+    # def blueprint_dataset(self) -> DatasetEntry | None:
+    #     result = self._inner.blueprint_dataset()
+    #     return DatasetEntry(result) if result is not None else None
 
-    def set_default_blueprint_segment_id(self, segment_id: str | None) -> None:
-        return self._inner.set_default_blueprint_partition_id(segment_id)
+    # def default_blueprint_segment_id(self) -> str | None:
+    #     return self._inner.default_blueprint_partition_id()
+
+    # def set_default_blueprint_segment_id(self, segment_id: str | None) -> None:
+    #     return self._inner.set_default_blueprint_partition_id(segment_id)
 
     def schema(self) -> Schema:
         return Schema(self._inner.schema(), _LazyDatasetState())
@@ -242,7 +249,8 @@ class DatasetEntry(Entry):
         view = DatasetView(self._inner, _LazyDatasetState())
         return view.segment_table(join_meta=join_meta, join_key=join_key)
 
-    def manifest(self) -> Any:
+    # What is this? What is it for?
+    def manifest(self) -> DataFusionTable:
         return self._inner.manifest()
 
     def segment_url(
@@ -254,15 +262,35 @@ class DatasetEntry(Entry):
     ) -> str:
         return self._inner.partition_url(segment_id, timeline, start, end)
 
+    # wanted: def add_segments([recording])
+    # requires default_s3_bucket somewhere
+
+    # We have a bunch of potentially slow stuff here: register, create_fts_index, …
+    # What is the API we want?
+    # One option is to return a `Job` object with
+    # .block(), .id, .cancel()
+    # catalog.register(…).block(timeout=60)
+    # catalog.register(…) # non-blocking
+    # job = catalog.register(…)
+    # while not job.is_done():
+    #     …
+    # job.block_and_tqdm()
+
+    # Call these `register_segment` and `register_segments` for clarity
+    # What happens on timeout? Does the server still finish the registration?
     def register(self, recording_uri: str, *, recording_layer: str = "base", timeout_secs: int = 60) -> str:
         return self._inner.register(recording_uri, recording_layer=recording_layer, timeout_secs=timeout_secs)
 
+    # Do we really want/need a layer per URI?
     def register_batch(self, recording_uris: list[str], *, recording_layers: list[str] | None = None) -> Any:
         if recording_layers is None:
             recording_layers = []
         return self._inner.register_batch(recording_uris, recording_layers=recording_layers)
 
-    def register_prefix(self, recordings_prefix: str, layer_name: str | None = None) -> Any:
+    # I assume this registers everything in a s3 bucket?
+    # Why isn't this just the same call as `register`?
+    # Why is this `layer_name` but called `recording_laye` earlier?
+    def register_prefix(self, recordings_prefix: str, *, layer_name: str | None = None) -> Any:
         return self._inner.register_prefix(recordings_prefix, layer_name)
 
     def download_segment(self, segment_id: str) -> Any:
@@ -286,10 +314,13 @@ class DatasetEntry(Entry):
             fill_latest_at=fill_latest_at,
         )
 
+    # May be mistaking read in the imperative ("please index these ranges")
+    # No better suggestion though.
     def index_ranges(self, index: str | IndexColumnDescriptor) -> datafusion.DataFrame:
         view = DatasetView(self._inner, _LazyDatasetState())
         return view.index_ranges(index)
 
+    # Not part of MVP. "Experimental - talk to sales"
     def create_fts_index(
         self,
         *,
@@ -305,6 +336,7 @@ class DatasetEntry(Entry):
             base_tokenizer=base_tokenizer,
         )
 
+    # Not part of MVP. "Experimental - talk to sales"
     def create_vector_index(
         self,
         *,
@@ -322,9 +354,11 @@ class DatasetEntry(Entry):
             distance_metric=distance_metric,
         )
 
+    # Is this only secondary indices? Should the name reflect that?
     def list_indexes(self) -> list:
         return self._inner.list_indexes()
 
+    # Do we need this for MVP?
     def delete_indexes(self, column: Any) -> list[Any]:
         return self._inner.delete_indexes(column)
 
@@ -334,6 +368,7 @@ class DatasetEntry(Entry):
     def search_vector(self, query: Any, column: Any, top_k: int) -> Any:
         return self._inner.search_vector(query, column, top_k)
 
+    # Should also return Job
     def do_maintenance(
         self,
         optimize_indexes: bool = False,
@@ -723,7 +758,7 @@ class TableEntry(Entry):
         """
         return self._inner.df()
 
-    def schema(self) -> pa.Schema:
+    def arrow_schema(self) -> pa.Schema:
         """Returns the schema of the table."""
         return self.reader().schema()
 
--- a/rerun_py/tests/api_sandbox/test_draft/test_catalog_basics.py
+++ b/rerun_py/tests/api_sandbox/test_draft/test_catalog_basics.py
@@ -23,7 +23,7 @@ def test_catalog_basics(tmp_path: Path) -> None:
         assert str(df.schema()) == inline_snapshot("""\
 id: fixed_size_binary[16] not null
 name: string not null
-entry_kind: int32 not null
+entry_kind: int32 not null  # Why is this int32 and not an enum? Poor py-arrow support?
 created_at: timestamp[ns] not null
 updated_at: timestamp[ns] not null
 -- schema metadata --
@@ -43,7 +43,7 @@ sorbet:version: '0.1.1'\
 │ │ ---        ┆ ---        │ │
 │ │ type: Utf8 ┆ type: i32  │ │
 │ ╞════════════╪════════════╡ │
-│ │ __entries  ┆ 3          │ │
+│ │ __entries  ┆ 3          │ │ # TODO(emilk): Can we remove __entries?
 │ ├╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌┤ │
 │ │ my_dataset ┆ 1          │ │
 │ ├╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌┤ │
--- a/rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
+++ b/rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
@@ -23,8 +23,8 @@ def test_dataframe_api_filter_partition_id(basic_dataset: DatasetEntry) -> None:
 
     assert str(df) == inline_snapshot("""\
 ┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
-│ METADATA:                                                                                                                                       │
-│ * version: 0.1.1                                                                                                                                │
+│ METADATA:          # TODO(emilk): can we skip this?                                                                                             │
+│ * version: 0.1.1   # TODO(emilk): can we skip this? Or call it `rerun_schema_version`?                                                          │
 ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
 │ ┌────────────────────┬──────────────────────────────┬───────────────────────────────────┬─────────────────────────────────────────────────────┐ │
 │ │ rerun_segment_id   ┆ timeline                     ┆ /points:Points2D:colors           ┆ /points:Points2D:positions                          │ │
@@ -73,6 +73,9 @@ def test_dataframe_api_using_index_values(complex_dataset: DatasetEntry) -> None
         index="timeline",
     ).sort("rerun_segment_id", "timeline")
 
+    # Should row_id be part of this?
+    # An API like `dataset_view.select_row_ids(row_ids)` would be the fastest way to fetch those rows of data.
+
     assert str(df) == inline_snapshot("""\
 ┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
 │ METADATA:                                                                                                                                                                             │
@@ -110,6 +113,9 @@ def test_dataframe_api_using_index_values(complex_dataset: DatasetEntry) -> None
 """)
 
     # Create a view with all partitions
+    # TODO(Niko): Future API idea: "Sample the data at 20 Hz starting at this time point with latest-at"
+
+    # This is a cumbersome API. Can we do better?
     df = dataset_view.reader(
         index="timeline",
         using_index_values={
@@ -130,6 +136,24 @@ def test_dataframe_api_using_index_values(complex_dataset: DatasetEntry) -> None
         },
         fill_latest_at=True,
     ).sort("rerun_segment_id", "timeline")
+
+
+    dataset_view.where(col("image") != None).latest_at()
+
+    # df = dataset_view.reader(
+    #     segment_id="complex_recording_0",
+    #     timeline=[
+    #         datetime.datetime(1999, 12, 31, 23, 59, 59),
+    #         datetime.datetime(2000, 1, 1, 0, 0, 1, microsecond=500),
+    #     ],
+    #     fill_latest_at=True,
+    # ).concat(dataset_view.reader(
+    #     segment_id="complex_recording_2",
+    #     timeline=[
+    #         datetime.datetime(1999, 12, 31, 23, 59, 59),
+    #     ],
+    #     fill_latest_at=True,
+    # )).sort("rerun_segment_id", "timeline")
 
     assert str(df) == inline_snapshot("""\
 ┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐


rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -70,6 +70,8 @@ def tables(self) -> datafusion.DataFrame:
         """Returns a DataFrame containing all table entries in the catalog."""
         return self._inner.tables()
 
+    # Why are _some_ functions prefixed with `get_` and not others?
+    # Is the logic "if it takes an argument"? Maybe this is Pythonic, IDK
```
abey79, line 74: worth running a "pythonic review board" on the APIs when we're are done with defining the functional part. i'd at least volunteer @ntjohnson1 and @timsaucer for that.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -242,7 +249,8 @@ def segment_table(
         view = DatasetView(self._inner, _LazyDatasetState())
         return view.segment_table(join_meta=join_meta, join_key=join_key)
 
-    def manifest(self) -> Any:
+    # What is this? What is it for?
```
abey79, line 252: the dataset manifest, aka a table describing the content of a dataset where each row is a layer

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -254,15 +262,35 @@ def segment_url(
     ) -> str:
         return self._inner.partition_url(segment_id, timeline, start, end)
 
+    # wanted: def add_segments([recording])
+    # requires default_s3_bucket somewhere
```
abey79, line 266: 💯 and I really really want it (e.g. for tests, etc.). But imo this is new functionality, not update/fixing of existing APIs. Given that it's potentially a rabbit hole, i'd like to scope it out. (but again, I would love to have this priorities)

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -254,15 +262,35 @@ def segment_url(
...
+    # We have a bunch of potentially slow stuff here: register, create_fts_index, …
+    # What is the API we want?
+    # One option is to return a `Job` object with
+    # .block(), .id, .cancel()
+    # catalog.register(…).block(timeout=60)
+    # catalog.register(…) # non-blocking
+    # job = catalog.register(…)
+    # while not job.is_done():
+    #     …
+    # job.block_and_tqdm()
```
abey79, line 277: We have a `Tasks` api already, which implements `wait(timeout=...)`.

Support for tqdm (let's call it `progress` or sth) would be a very nice addition. I'm fine adding a tqdm dependency for this sole purpose because:
- tqdm is extremely widespread/standard
- we're splitting the redap sdk from the logging sdk anyways (right, RIGHT?)

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -254,15 +262,35 @@ def segment_url(
...
+
+    # Call these `register_segment` and `register_segments` for clarity
+    # What happens on timeout? Does the server still finish the registration?
     def register(self, recording_uri: str, *, recording_layer: str = "base", timeout_secs: int = 60) -> str:
         return self._inner.register(recording_uri, recording_layer=recording_layer, timeout_secs=timeout_secs)
 
+    # Do we really want/need a layer per URI?
```
abey79, line 284: that is (for now) the very definition of what a layer is

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -254,15 +262,35 @@ def segment_url(
...
     def register_batch(self, recording_uris: list[str], *, recording_layers: list[str] | None = None) -> Any:
         if recording_layers is None:
             recording_layers = []
         return self._inner.register_batch(recording_uris, recording_layers=recording_layers)
 
-    def register_prefix(self, recordings_prefix: str, layer_name: str | None = None) -> Any:
+    # I assume this registers everything in a s3 bucket?
+    # Why isn't this just the same call as `register`?
+    # Why is this `layer_name` but called `recording_laye` earlier?
```
abey79, line 292: good call. the official name is `layer_name`, the other is wrong

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -254,15 +262,35 @@ def segment_url(
...
+    # catalog.register(…) # non-blocking
+    # job = catalog.register(…)
+    # while not job.is_done():
+    #     …
+    # job.block_and_tqdm()
+
+    # Call these `register_segment` and `register_segments` for clarity
+    # What happens on timeout? Does the server still finish the registration?
     def register(self, recording_uri: str, *, recording_layer: str = "base", timeout_secs: int = 60) -> str:
         return self._inner.register(recording_uri, recording_layer=recording_layer, timeout_secs=timeout_secs)
```
abey79, line 282:
```suggestion
    def register(self, recording_uri: str, *, layer_name: str = "base", timeout_secs: int = 60) -> str:
        return self._inner.register(recording_uri, recording_layer=layer_name, timeout_secs=timeout_secs)
```

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -254,15 +262,35 @@ def segment_url(
...
 
+    # Do we really want/need a layer per URI?
     def register_batch(self, recording_uris: list[str], *, recording_layers: list[str] | None = None) -> Any:
         if recording_layers is None:
             recording_layers = []
         return self._inner.register_batch(recording_uris, recording_layers=recording_layers)
 
-    def register_prefix(self, recordings_prefix: str, layer_name: str | None = None) -> Any:
+    # I assume this registers everything in a s3 bucket?
+    # Why isn't this just the same call as `register`?
```
abey79, line 291: The FS analogy would be:
- `register` -> single file
- `register_batch` -> list of files
- `register_prefix` -> entire directory (somehow directories are called prefix in object store)

I would submit the idea of merging these two APIs into one (using kwargs shenanigans) to the "pythonic review board"

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -286,10 +314,13 @@ def reader(
             fill_latest_at=fill_latest_at,
         )
 
+    # May be mistaking read in the imperative ("please index these ranges")
+    # No better suggestion though.
     def index_ranges(self, index: str | IndexColumnDescriptor) -> datafusion.DataFrame:
```
abey79, line 319: `get_index_ranges` might be a better name, yeah

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -305,6 +336,7 @@ def create_fts_index(
             base_tokenizer=base_tokenizer,
         )
 
+    # Not part of MVP. "Experimental - talk to sales"
```
abey79, line 339: - https://github.com/rerun-io/rerun/pull/11954

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -322,9 +354,11 @@ def create_vector_index(
             distance_metric=distance_metric,
         )
 
+    # Is this only secondary indices? Should the name reflect that?
     def list_indexes(self) -> list:
         return self._inner.list_indexes()
 
+    # Do we need this for MVP?
```
abey79, line 361: Index management APIs were introduced as a direct consequence of customer issues/requests

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -210,25 +212,30 @@ class DatasetEntry(Entry):
     def __init__(self, inner: _catalog.DatasetEntry) -> None:
         self._inner = inner
 
+    # What is this? What is it for?
     @property
     def manifest_url(self) -> str:
         return self._inner.manifest_url
```
abey79, line 218: let's remove this

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: from the blueprint-related decision: we need something like this, but exposing it for data is not top-priority

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: Decision:
- make `.register()` return a `Tasks` for consistency
- we need build a nice API at the edge with clear error reporting (importantly OK/NOK on a per segment basis), cancelling, progress bar, etc.
- make sure we have trace ids, etc. to help support/debugging
- the existing redap low-level API is ok, but needs sugar coating on Python side.
- expose `Tasks` as `Job` to "hide" the underlying implementation details?

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: decision:
- we want:
  - register 1 segment with a given layer name (optionally)
  - register N segments with a single layer name (optionally) (use multiple calls if you want multiple layer names)
  - register all rrds under a given prefix with a single layer name (optionally)
- all of these returns `Tasks`
-  "pythonic review board" to decide how far to merge `register`, `register_batch` and `register_prefix` (the 3 of them sound way too much) (@abey79 to make a propose)
- ~`register` -> `register_segment/s` for clarity~ (these register a segment/layer combo, not a segment)

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -322,9 +354,11 @@ def create_vector_index(
             distance_metric=distance_metric,
         )
 
+    # Is this only secondary indices? Should the name reflect that?
     def list_indexes(self) -> list:
```
abey79, line 358: decision:
- `create_secondary_index` (kwargs to define which type)
- `list_secondary_indices`
- `delete_secondary_indices`

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -334,6 +368,7 @@ def search_fts(self, query: str, column: Any) -> Any:
     def search_vector(self, query: Any, column: Any, top_k: int) -> Any:
         return self._inner.search_vector(query, column, top_k)
 
+    # Should also return Job
```
abey79, line 371: yes

rerun_py/tests/api_sandbox/test_draft/test_catalog_basics.py
```diff
@@ -43,7 +43,7 @@ def test_catalog_basics(tmp_path: Path) -> None:
 │ │ ---        ┆ ---        │ │
 │ │ type: Utf8 ┆ type: i32  │ │
 │ ╞════════════╪════════════╡ │
-│ │ __entries  ┆ 3          │ │
+│ │ __entries  ┆ 3          │ │ # TODO(emilk): Can we remove __entries?
```
abey79, line 46: the entries table as a dataframe is essentially disappearing from the API, so it's ok for it to remain low-level

rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
```diff
@@ -73,6 +73,9 @@ def test_dataframe_api_using_index_values(complex_dataset: DatasetEntry) -> None
         index="timeline",
     ).sort("rerun_segment_id", "timeline")
 
+    # Should row_id be part of this?
+    # An API like `dataset_view.select_row_ids(row_ids)` would be the fastest way to fetch those rows of data.
```
abey79, line 77: This hinges on exposing the ability to run dataframe queries with `row_id` as index, which we currently don't expose (at least on python side). TBD, but probably out of scope for the API project.

rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
```diff
@@ -131,6 +137,24 @@ def test_dataframe_api_using_index_values(complex_dataset: DatasetEntry) -> None
...
+    dataset_view.where(col("image") != None).latest_at()
+
+    # df = dataset_view.reader(
+    #     segment_id="complex_recording_0",
+    #     timeline=[
+    #         datetime.datetime(1999, 12, 31, 23, 59, 59),
+    #         datetime.datetime(2000, 1, 1, 0, 0, 1, microsecond=500),
+    #     ],
+    #     fill_latest_at=True,
+    # ).concat(dataset_view.reader(
```
jleibs, line 150: This makes sense for merging 1 or 2 views but seems impractical as a replacement for using_index_values on 1000s of segments.

rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
```diff
@@ -110,6 +113,9 @@ def test_dataframe_api_using_index_values(complex_dataset: DatasetEntry) -> None
 """)
 
     # Create a view with all partitions
+    # TODO(Niko): Future API idea: "Sample the data at 20 Hz starting at this time point with latest-at"
+
+    # This is a cumbersome API. Can we do better?
```
jleibs, line 118: Do you find the version that directly takes a DataFrame to be cumbersome?

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
jleibs: > "pythonic review board" to decide how far to merge register, register_batch and register_prefix (the 3 of them sound way too much)

@nikolausWest this feels like the opposite direction of the table API design where you pushed for 3 explicitly separate APIs (append/overwrite/upsert) instead of just having a single API and a write_mode parameter.

We need to fundamentally decide if we want a minimal number of APIs with complex parameters and behaviors, or a larger number of APIs with very specific clear functionality. 

I am personally leaning toward the latter. e.g. in the table APIs a separate `append_batches` would make the docs and implementation of error handling more clear (none of this incompatible subsets of parameters business).

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -254,15 +262,35 @@ def segment_url(
...
+    # One option is to return a `Job` object with
+    # .block(), .id, .cancel()
+    # catalog.register(…).block(timeout=60)
+    # catalog.register(…) # non-blocking
+    # job = catalog.register(…)
+    # while not job.is_done():
+    #     …
+    # job.block_and_tqdm()
+
+    # Call these `register_segment` and `register_segments` for clarity
```
jleibs, line 279: The challenge here is it's not a segment. It's a segment-layer (which would benefit from a better name). Several registrations can all add to the same segment.

rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
abey79: FYI, we discussed that at length with Niko, and concluded that:
- the current proposal is not _great_
- it's not immediately obvious how to make it better
- a lot of this discussion hinges on future features (time resample helper,  alternate interpolation function, etc.)
- we are unable to make a guess now as to how these future features will look like (aka chances are we'd make wrong choices)

=> we're OK to roll with the proposal in it's current state

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -210,25 +212,30 @@ class DatasetEntry(Entry):
...
     def manifest_url(self) -> str:
         return self._inner.manifest_url
 
     def arrow_schema(self) -> pa.Schema:
         return self._inner.arrow_schema()
 
-    def blueprint_dataset_id(self) -> EntryId | None:
-        return self._inner.blueprint_dataset_id()
+    # Remove these blueprint stuff, and just have
+    def set_default_blueprint(self, blueprint: Blueprint | Path) -> None:
```
ntjohnson1, line 224: For the api can we just say `set_blueprint`? Because we either have the default blueprint or the heuristic on the viewer so this is the only configurable blueprint.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: After discussing with @ntjohnson1:
- we _already_ have a `Task` (singular) abstraction that I somehow overlooked this morning
- a good API inspiration may come from [`concurrent.futures.as_completed`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.as_completed), which we could add to `Tasks`

The latter would allow:
- very easy user-provided progress-bar integration
- get error on a per-task basis as they arise

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
timsaucer: I do think this is useful, especially for any users who leverage the layers heavily. The naming may need improvement though.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: Discussing this with @ntjohnson1, having a single `register` call is ok because all 3 can be used to achieved the exact same outcome (at the the cost of a loop or so), whereas append/overwrite/upsert achieve distinct tasks. So `there should be one-- and preferably only one --obvious way to do it` would apply here.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: @ntjohnson1 proposal: "search index"

- `create_search_index`
  - or `create_fts_search_index` and `create_fts_search_index`
- `list_search_indices`
- `delete_search_indices`
- `search`
  - or `fts_search`/`vector_search`  (docstring: "requires relevant search index")

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
timsaucer, line 74: I think in general Python prefers brevity and Rust does not. So we have a case here were I suspect we are mirroring the internal APIs. I think we should remove the `get_` prefixes where possible.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
timsaucer: 100% support this

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
timsaucer: Came on here to agree with @abey79 's last statement. I don't think this is the same argument as having different append/overwrite/upsert because these are exactly 1 concept with 3 different input types. Having different input types and automatically knowing how to handle them is one thing python should be good at.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
timsaucer: This goes to the above discussion for `get_` prefixes and the conflict between brevity and being explicit. I would not expect `index_ranges` to perform indexing.

rerun_py/tests/api_sandbox/test_draft/test_catalog_basics.py
```diff
@@ -23,7 +23,7 @@ def test_catalog_basics(tmp_path: Path) -> None:
         assert str(df.schema()) == inline_snapshot("""\
 id: fixed_size_binary[16] not null
 name: string not null
-entry_kind: int32 not null
+entry_kind: int32 not null  # Why is this int32 and not an enum? Poor py-arrow support?
```
timsaucer, line 26: The underlying data type is int32. We can create an extension type and then also update our rendering system to use those extension types.

rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
```diff
@@ -23,8 +23,8 @@ def test_dataframe_api_filter_partition_id(basic_dataset: DatasetEntry) -> None:
 
     assert str(df) == inline_snapshot("""\
 ┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
-│ METADATA:                                                                                                                                       │
-│ * version: 0.1.1                                                                                                                                │
+│ METADATA:          # TODO(emilk): can we skip this?                                                                                             │
+│ * version: 0.1.1   # TODO(emilk): can we skip this? Or call it `rerun_schema_version`?                                                          │
```
timsaucer, line 27: Yes, we can update the rendering. Easily done. The question is: what about other metatdata if the user adds it?

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: My proposal is as follows:

```python
def register(
    *,
    uri: str | Sequence[str] | None = None,
    prefix: str | None = None,
    layer_name: str | Sequence[str] = "base",
) -> Tasks: ...


# Example usage.

# Single partition
ds.register(uri="s3://bkt/data1.rrd").wait()

# Single partition with specific layer
ds.register(uri="s3://bkt/data1.rrd", layer_name="embeddings").wait()

# Multiple partitions with same layer name
ds.register(
    uri=[
        "s3://bkt/data1.rrd",
        "s3://bkt/data2.rrd",
        "s3://bkt/data3.rrd",
    ],
    layer_name="embeddings",
).wait()

# Multiple partitions with multiple layer names (list length must match)
ds.register(
    uri=[
        "s3://bkt/embedding.rrd",
        "s3://bkt/previews.rrd",
    ],
    layer_name=[
        "embeddings",
        "previewis",
    ],
).wait()

# Prefix (either `uri` or `prefix` must be provided, passing both will raise an error)
ds.register(prefix="s3://bkt/prefix/").wait()

# Prefix with layer name (`layer_name` cannot be a list)
ds.register(prefix="s3://bkt/prefix/", layer_name="base").wait()
```

Note that upon Nick suggestion, I'm keeping the ability to have multiple layer names as there are existing user workflow relying on that.

nikolausWest: How would prefix interact with multiple layer names?

nikolausWest: Regarding three separate actions vs one action with multiple input types I was thinking the same as @abey79 and @timsaucer

abey79: I'm disallowing it. List of layer name is allowed only when `uri` is a list as well, and length must match.

jleibs: I buy the argument for merging register and register batch. Promoting a single element to a list is fairly common. In both of them the total number of sub-tasks in the output is known and deterministic based on the input.

register_prefix, on the other hand, still feels like a legitimately odd-duck to me. It's actually not something where you can achieve the same thing with a loop. The user may not even have permissions for the s3 location they are trying to register. (For that matter the API itself also requires broader server permissions -- list vs read). The number of results is unknown when the user makes the call. The server behind the scenes also goes down a fairly distinct code path to scan the prefix first.

ntjohnson1: > The user may not even have permissions for the s3 location they are trying to register

This seems more like a bug than a feature to me, but it does provide a weird edge case around the equivalence of these methods.

One thing that I mentioned to Antoine that might be captured elsewhere is if we make our Tasks/Jobs object nicer/more informative that might help to make it clearer their prefix is not aligned with their expectations. However, if they just use boto3 and rglob the prefix this same footgun is pretty easy to add a lot more data than you thought so pretty hard to design that away even with different method.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
nikolausWest: Sounds good!

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
nikolausWest: Fair and more descriptive. Lets go with "search index"

rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
nikolausWest: Regarding the cumbersome comment. As Emil mentioned this was just some notes we wrote down between flights and not meant as a real full review on it's own. On first read through the combination of timeline + using_index_values + latest at seemed non-future proof and like there might be a more composable way to do it. One of the perceived problems was that we're likely to want more kinds of sampling methods that will force more kwargs. The other is that more fill methods will do the same. However, we weren't able to come up with something composable that was good (timeline, sampling, and fill methods belong together) and don't have enough clarity on the grounded needs for the sampling and fill methods to propose something better.

That's why we concluded it's best to keep it as is and then evolve this API as needed

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: The current plan is:

```python
def datasets() -> list[DatasetEntry]: ...
def get_dataset(*, id, name) -> DatasetEntry: ...
```
This morning, I got a bit annoyed with the inconsistency. Why is there a `get_` prefix in one and not in the other? Yet, I felt that `get_datasets()` is unnecessary verbose, and `dataset(id, name)` is not really pythonic, leading to a bit of cognitive dissonance.

I poked around a bit and ended up coming to peace with the status quo. Rationale:
- `datasets()` doesn't need an argument (it actually has one optional arg now to show the hidden entries, but that's niche), so it looks and feels like a property. It _cannot_ be a property given that network calls are involved, but it is one in spirit.
- `get_dataset()` requires an argument, so it is unambiguously an accessor, for which the `get_` prefix is extremely common in Python. The core inspiration here being `dict`: `my_dict.get("item")`

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: implemented in https://github.com/rerun-io/rerun/pull/11969

rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
nikolausWest: The main problem here is that _version_ is unclear. What version is it referring to? Can the user add their own metadata here?

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: Formal proposal here:
- https://linear.app/rerun/issue/RR-3018/improve-the-dataset-blueprint-apis-in-the-python-sdk

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: @ntjohnson1 currently, `register` returns the segment id of the segment that was just registered. This will be "lost" if we return `Tasks` instead.

Q: was this returned segment id useful? (My gut feeling is yes, but I'd love to have your take on this.)

`Tasks` should obviously have the ability to provide the segment ids of what was registered, but:
- from an API design perspective, it's tricky because segment registration is a special case of a task
- I'm not even sure how this information is made available by redap in the first place

Alternative: `register_*` calls do not return any segment ids information. Escape hatch is to list segments before/after and diff.

cc @nikolausWest

abey79: arf... as @jleibs noted below, registration is about `(segment, layer)`, not just segment. (I somehow keep forgetting about this.) So this pleads in favour of not attempting to return such information from `register_*` or `Tasks`. How bad does that sounds from a user's prespective?

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: I share the opinion that `register` and `register_batch` are very similar, and `register_prefix` a bit less so. Seeing this being debated, I'm personally undecided about merging them all vs. keeping `register` + `register_prefix`.

If I had to make the call, I would go for `register`/`register_prefix` because, facing a doubt, I pick explicit over full-on kwargs.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: @timsaucer I'm unsure about which direction you are trying to push.

As per my newfound "rule" of "skip `get_` when it looks like a `@property` (but cannot be one)", this should have `get_` as it takes a mandatory argument.

rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
abey79: From the code, it should be `sorbet:version`. I have no idea why we only see `version` here. Something about the formatting?

**edit**: confirmed, it's indeed `sorbet:version` in the data. I'm not sure why this gets printed as `version` for `str(datafusion.DataFrame)`. Maybe a bug @timsaucer? Related to the handling of extension types? (These involve `:` as well iirc.)

abey79: @timsaucer 
- https://github.com/apache/datafusion-python/issues/1312

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: mvp implemented in:
- https://github.com/rerun-io/rerun/pull/11969

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: changed to return a `datafusion.DataFrame` for consistency in:
- https://github.com/rerun-io/rerun/pull/11969

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: fixed in:
- https://github.com/rerun-io/rerun/pull/11969

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: renamed in `get_index_ranges` in:
- https://github.com/rerun-io/rerun/pull/11969

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: mvp of that change implemented in:
- https://github.com/rerun-io/rerun/pull/11969

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: ...well. The grpc currently doest return tasks. I'm not sure how this is even done in cloud. I suggest we leave this aside for this project, as this feature is entirely orthogonal to everything else.

rerun_py/tests/api_sandbox/test_draft/test_catalog_basics.py
abey79: I agree, but imo out of scope for this project.

rerun_py/tests/api_sandbox/test_draft/test_dataframe_api.py
timsaucer: Added notes to the issue in the datafusion-python repo, but here is the offending source: https://github.com/rerun-io/rerun/blob/128d37a175d5410b86ac6a0153252ecf366e906c/rerun_py/src/catalog/dataframe_rendering.rs#L160 The default options are to trim metadata, which is applied both at the header and for the columns.

abey79: Ah, right! I've been completely fooled by this machinery I wasn't aware of. Sorry for the noise.

emilk:
I think this has been processed by now

PR #11951, opened 2025-11-23
Support URLs with origins and path prefixes

methylDragon:
### What

This PR updates the URL parser to support path prefixes. This allows Rerun instances to be hosted at non-root paths (e.g., behind a reverse proxy).

Before: The parser assumed Rerun endpoints existed at the root:
- `http://example.com/catalog`
- `http://example.com/proxy`

After: The parser now accepts arbitrary sub-paths:
- `http://example.com/custom/prefix/catalog`
- `http://example.com/custom/prefix/proxy`

</details>

### Motivation

Currently, hosting Rerun behind a reverse proxy at a specific sub-path triggers a "Failed to parse URL" error.

For example, if `example.com` hosts a Rerun instance at `example.com/hosted_rerun/`, the current parser cannot handle the gRPC proxy link:
```
https://example.com/hosted_rerun/url?=rerun%2Bhttps://example.com/hosted_rerun_grpc_data/proxy
```

You get a "Failed to parse URL error"! Hence motivating this PR.

I am trying to host a Rerun web viewer behind such a reverse proxy and getting this issue.

### Additional Concerns

Some of the URL parsing logic needs to search for keywords to then extract arguments from.

Consider:
```
http://example.com/sub/path/entry/entry_id/dataset/dataset_id
```

Should this be an "entry" or a "dataset" URL?

I decided it would be a "dataset" URL, to support cases where an external page has a really long, accidentally clobbering path prefix, by having the **last** occurence of a keyword be what determines what kind of page it is e.g.:
```
http://example.com/path/that/contains/example/dataset/and/then/hosts/rerun/dataset/dataset_id
```

If we searched the first, the chance of an unintentional collision is higher.

### Tests

I added more unit tests and adjusted the pre-existing one.

### DIFF
--- a/crates/utils/re_uri/src/endpoint_addr.rs
+++ b/crates/utils/re_uri/src/endpoint_addr.rs
@@ -1,0 +1,102 @@
+use crate::Origin;
+
+/// A Rerun endpoint address, consisting of an origin and an optional path prefix.
+///
+/// Example: `https://rerun.io:443/custom/prefix`
+#[derive(Clone, Debug, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
+pub struct EndpointAddr {
+    pub origin: Origin,
+
+    /// An optional path prefix, e.g. `/my/prefix`.
+    ///
+    /// The prefix is guaranteed to start with a slash if it is not empty,
+    /// and guaranteed not to end with a slash.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub path_prefix: Option<String>,
+}
+
+impl EndpointAddr {
+    /// Create a new [`EndpointAddr`] with the given origin and no path prefix.
+    pub fn new(origin: Origin) -> Self {
+        Self {
+            origin,
+            path_prefix: None,
+        }
+    }
+
+    /// Add a path prefix to the endpoint address.
+    ///
+    /// The prefix is normalized:
+    /// - It will be ensured to start with a `/` if not empty.
+    /// - Trailing slashes will be removed.
+    pub fn with_path_prefix(mut self, path_prefix: impl Into<String>) -> Self {
+        let path_prefix = path_prefix.into();
+        if path_prefix.is_empty() || path_prefix == "/" {
+            self.path_prefix = None;
+            return self;
+        }
+
+        let mut path_prefix = path_prefix.trim_end_matches('/').to_owned();
+        if !path_prefix.starts_with('/') {
+            path_prefix.insert(0, '/');
+        }
+
+        self.path_prefix = Some(path_prefix);
+        self
+    }
+}
+
+impl std::fmt::Display for EndpointAddr {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let Self {
+            origin,
+            path_prefix,
+        } = self;
+
+        write!(f, "{origin}")?;
+        if let Some(prefix) = path_prefix {
+            write!(f, "{prefix}")?;
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::Scheme;
+
+    fn test_origin() -> Origin {
+        Origin::from_scheme_and_socket_addr(Scheme::Rerun, "127.0.0.1:1234".parse().unwrap())
+    }
+
+    #[test]
+    fn test_serialization() {
+        let origin = test_origin();
+        let addr = EndpointAddr::new(origin.clone());
+        assert_eq!(addr.path_prefix, None);
+        assert_eq!(addr.to_string(), "rerun://127.0.0.1:1234");
+
+        let addr = EndpointAddr::new(origin.clone()).with_path_prefix("");
+        assert_eq!(addr.path_prefix, None);
+
+        let addr = EndpointAddr::new(origin.clone()).with_path_prefix("/");
+        assert_eq!(addr.path_prefix, None);
+
+        let addr = EndpointAddr::new(origin.clone()).with_path_prefix("foo");
+        assert_eq!(addr.path_prefix.as_deref(), Some("/foo"));
+        assert_eq!(addr.to_string(), "rerun://127.0.0.1:1234/foo");
+
+        let addr = EndpointAddr::new(origin.clone()).with_path_prefix("/foo");
+        assert_eq!(addr.path_prefix.as_deref(), Some("/foo"));
+        assert_eq!(addr.to_string(), "rerun://127.0.0.1:1234/foo");
+
+        let addr = EndpointAddr::new(origin.clone()).with_path_prefix("foo/");
+        assert_eq!(addr.path_prefix.as_deref(), Some("/foo"));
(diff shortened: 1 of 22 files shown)


crates/utils/re_uri/src/endpoints/proxy.rs
```diff
@@ -3,17 +3,18 @@ use crate::{Origin, RedapUri};
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct ProxyUri {
     pub origin: Origin,
+    pub prefix: String,
```
grtlr: Can we please give this a more descriptive name like `path_prefix`? Same applies to the other `*Uri`s.

grtlr (changes requested):
Overall this sounds like a reasonable feature—thank you for opening the PR!

As a quick fix, you could now already host Rerun under a subdomain and it should work out of the box. But I can see how this can sometimes could be too limiting. I'm curious: Is there anything preventing you from doing this that motivates this PR?

> [!IMPORTANT]
> The way we currently use paths with GRPC endpoints is already a bit of a hack, so I'm a bit hesitant complicating things even more _at the current point in time_.

Code-wise, I think there are also some changes that we need to make if we go down that path (pun intended).

* We should separate the path handling + the origin into a new object to avoid having to add logic to every new `*Uri` variant.
* This new object could then replace the existing `origin` field in those structs.
* Given that `path_prefix` is more of a niche feature, I'd make it an `Option` too, and use the builder pattern to add it. This is also motivated by the following:
* Finally, we need to make sure that the implementation is robust against leading and trailing slashes. A method like `with_path_prefix` could be used to validate the inputs. We should also test those edge cases too.

methylDragon:
> Overall this sounds like a reasonable feature—thank you for opening the PR!
> 
> As a quick fix, you could now already host Rerun under a subdomain and it should work out of the box. But I can see how this can sometimes could be too limiting. I'm curious: Is there anything preventing you from doing this that motivates this PR?

The way we're hosting Rerun atm is:
- We're spinning up a different VM for each user who is requesting visualization of a file they have (MCAP). These VMs are ephemeral and separately authed.
- We also have different "organizations" that the users belong to

The number of users and orgs are relatively unbounded for us, and make using a subdomain pretty tricky. The URL we end up with is something like: `https://our_site.com/org/user/vm/rerun`, hence motivating use of subpaths. (Similarly for the grpc server)

crates/utils/re_uri/src/endpoints/entry.rs
```diff
@@ -6,19 +6,23 @@ use crate::{Error, Origin, RedapUri};
...
     pub origin: Origin,
+    pub prefix: String,
     pub entry_id: EntryId,
 }
 
 impl std::fmt::Display for EntryUri {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let Self { origin, entry_id } = self;
-        write!(f, "{origin}/entry/{entry_id}")
+        write!(f, "{}{}/entry/{}", self.origin, self.prefix, self.entry_id)
```
emilk: This is a bad change - using explicit destruction is preferred, as it forces us to consider new additions as they are made

emilk (commented):
Seems like adding the path-prefix to the `struct Origin` would make the code simpler, and less error-prone (though the name `Origin` would be a bit misleading in that case)

crates/utils/re_uri/src/endpoints/entry.rs
methylDragon: https://github.com/rerun-io/rerun/pull/11951/commits/e44fee4a07171e705a59b24646ac625b0b5f90d0

methylDragon:
> Seems like adding the path-prefix to the `struct Origin` would make the code simpler, and less error-prone (though the name `Origin` would be a bit misleading in that case)

Added a new struct `EndpointAddr` and used it, composing `Origin`

crates/utils/re_uri/src/endpoint_addr.rs
```diff
@@ -0,0 +1,102 @@
...
+pub struct EndpointAddr {
+    pub origin: Origin,
+
+    /// An optional path prefix, e.g. `/my/prefix`.
+    ///
+    /// The prefix is guaranteed to start with a slash if it is not empty,
+    /// and guaranteed not to end with a slash.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub path_prefix: Option<String>,
+}
```
grtlr, line 16: Shouldn't this just be a `url::Url`?

I remember there being some shenanigans around default ports though:

* https://github.com/servo/rust-url/issues/706

methylDragon: Technically speaking a URI is: `<scheme>://<origin>/<subpath>/<endpoint>`

I was taking the EndpointAddr to be `<origin>/<subpath>`, since it isn't including the final endpoint segment.

grtlr (changes requested):
If we want to land this change, we should also make sure that we add the prefix to all our SDKs, as handling otherwise becomes inconsistent.

The `partition_url` in the Python SDK is just one such example. There is also `ConnectionHandle` and probably more places.

crates/utils/re_uri/src/endpoint_addr.rs
methylDragon: Given that ^, what do you think? I'm trying to scope down the change as much as possible (I'm treating this as an extension to the origin, mostly)

methylDragon:
> If we want to land this change, we should also make sure that we add the prefix to all our SDKs, as handling otherwise becomes inconsistent.
> 
> The `partition_url` in the Python SDK is one such example.

Apologies, I'm a little unfamiliar with the code base, what do you mean by this?

grtlr:
Sorry, was just about to clarify my comment. What I meant is: We basically would need to start using the new `EndpointAddr` instead of `Origin` in many places to have consistency across all SDKs and use cases.

This looks like a pretty big task, so I wonder what @emilk's thoughts are here?

methylDragon:
Bump on this @grtlr / @emilk 

Alternatively, is there work done/underway for supporting hosted instances of Rerun, pointing to gRPC data located behind proxies? This issue is unfortunately preventing us from upgrading from Rerun v0.22.X

crates/utils/re_uri/src/endpoint_addr.rs
grtlr: Our `Origin` contains the scheme as well. So my idea was to use an URL to represent everything up to endpoint. Maybe we could therefore use a simple wrapper struct around `url::Url`.

grtlr: By replacing `Origin`, we are also ensuring that we catch all instances where the path segment functionality needs to be added.

grtlr:
Sorry for the slow response, most of us have been on out over the break.

The tricky thing here is that we need to ensure to add this functionality to all places where we currently use `Origin`. That means all SDKs and the link sharing in the browser, so this is a big undertaking. @abey79 has also been refactoring the API over the last couple of weeks, so he might have opinions too.

If these above points are addressed (also commented above), I think this would be a very nice addition from a technical point of view.

Wumpf:
seems this is too outdated by now, closing

PR #11952, opened 2025-11-23
Update Table APIs based on Niko Review

jleibs:
### Related

### What

I did a review of the table API surface with @nikolausWest and @emilk -- this was the first round of requested changes.

Overview of the requested changes:
- By default, hide the `__entries` Table and blueprint datasets.
- Don't expose DF-oriented APIs for listing Entries/Datasets/Tables -- only lists of Entry objects
  - e.g. `client.datasets() -> list[DatasetEntry]`
- Replace `Entry.update()` with `Entry.set_name()`. (update sounded like it could be modifying the table contents)
- Get rid of `cli.write_table()` in favor of `TableEntry`-oriented helpers
  - Add symmetric helpers for `table.append()`, `table.overwrite()`, and `table.upsert()` -- each of these helpers takes either a iterator of RecordBatches OR the named-column structure
  - Rationale: only 2 ways of doing things.  Datafusion (e.g. `ctx.table(...)` OR via the TableEntry objects)


rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -34,17 +35,32 @@ def __init__(self, address: str, token: str | None = None) -> None:
...
+            [
+                Entry(e)
+                for e in self._inner.all_entries()
+                if (not e.name.startswith("__") and e.kind != 5) or include_hidden
+            ],
+            key=lambda e: e.name,
+        )
 
-    def dataset_entries(self) -> list[DatasetEntry]:
+    def datasets(self, *, include_hidden=False) -> list[DatasetEntry]:
```
abey79, line 48: Repurposing an existing name for a different signature is obviously a very breaking change. The project already states that we wont be running a deprecation cycle on these API changes, which would be tricky to begin with. I'm 100% fine with this, I still wanted to call this out.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -705,12 +697,148 @@ def client(self) -> CatalogClient:
...
+        batches: pa.RecordBatch | Iterator[pa.RecordBatch] | Iterator[Iterator[pa.RecordBatch]] | None = None,
+        **named_params: Any,
+    ) -> None:
+        """
+        Upsert data into the Table.
+
+        To use upsert, the table must contain a column with the metadata:
+        ```
+            {"rerun:is_table_index" = "true"}
+        ```
```
abey79, line 819: Some thoughts/questions:
- It would be very useful to have a `TableEntry.set_table_index(column="my_column")` helper
- Would it useful for the segment table to have it's `rerun_segment_id` column pre-marked as table index?
- Which leads to the layer table. Its primary key is `(rerun_segment_id, rerun_layer_name)`. Do we want to support multi-column table indices?

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -58,22 +74,6 @@ def table_names(self) -> list[str]:
...
-        """Returns a DataFrame containing all entries in the catalog."""
-        return self._inner.entries()
-
-    def datasets(self) -> datafusion.DataFrame:
-        """Returns a DataFrame containing all dataset entries in the catalog."""
-        return self._inner.datasets()
-
-    def tables(self) -> datafusion.DataFrame:
-        """Returns a DataFrame containing all table entries in the catalog."""
-        return self._inner.tables()
```
abey79, line 71: Are we 100% sure these weren't useful? Would be nice to have confirmation from @ntjohnson1 .

abey79 (approved):
Code LGTM, matches the changes summary in the PR desc (thx for that).

In general, I like these API changes. Makes the whole thing more straightforward. My primary thoughts/questions on this are:
- Did we confirm with user space (aka @ntjohnson1) that the DF-entry API aren't useful?
- Calling out that changing signature of existing call is a strongly breaking change and we're running this without deprecation cycle. (I'm 100% fine with it.)
- Some thoughts/questions on table indices, see bellow.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -34,17 +34,32 @@ def __init__(self, address: str, token: str | None = None) -> None:
...
         return repr(self._inner)
 
-    def all_entries(self) -> list[Entry]:
+    def entries(self, *, include_hidden=False) -> list[Entry]:
         """Returns a list of all entries in the catalog."""
-        return [Entry(e) for e in self._inner.all_entries()]
+        return sorted(
+            [
+                Entry(e)
+                for e in self._inner.all_entries()
```
timsaucer, line 42: minor: if updating the naming of this function then I find it beneficial to also update the underlying function it mirrors. I think it leads to better maintainability and understanding for people new to the code base.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -34,17 +34,32 @@ def __init__(self, address: str, token: str | None = None) -> None:
...
+                if (not e.name.startswith("__") and e.kind != 5) or include_hidden
```
timsaucer, line 43: Good call on the `include_hidden` option

timsaucer, line 43: Below we do a `e.kind == _catalog.EntryKind.DATASET` which I find very intuitive. Can we do the same here instead of magic number 5 (which could potentially change)

rerun_py/tests/api_sandbox/test_draft/conftest.py
```diff
@@ -74,21 +74,21 @@ def populated_client_complex(complex_dataset_prefix: Path) -> Iterator[rr.catalo
         yield client
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
```
timsaucer, line 77: Are none of these tests reusable? It does make a big difference in how long it takes to run against the cloud server.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
timsaucer: I guess this comment is more for the actual implementation rather than this draft.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: After discussing with Nick, these are fine to remove.

rerun_py/tests/api_sandbox/test_draft/conftest.py
abey79: Yeah I wondered the same and I'm unsure how this change came to be. OTOH, I dont really care: these tests will have to be properly ported anyways to make it the `e2e_redap_tests`. For now, given they are just demonstration snippets, they only target the OSS server.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: Well this API is actually disappearing entirely in the next PR.

edit: actually that's wrong (not sure what I was thinking). The implementation in this throwaway mock matters little, but I'll clean it up anyway in the next PR.

PR #11925, opened 2025-11-19
Move `Transform3D::axis_length` into its own `TransformAxes3D` archetype

grtlr:
### Related

* Closes RR-2857.
* [x] Requires #11934.
* [x] Requires #11953.

### What

This moves `Transform3D::axis_length` to its own archetype (`TransformAxes3D`). This also allows us to get rid of the `DataBasedVisualizabilityFilter` trait.

### TODO

* [x] `@rerun-bot full-check`

PR #11953, opened 2025-11-24
Respect ignore patterns from `lychee.toml` in `pr_link_checker.py`

grtlr:
### Related

* Required for #11925.

### What

The `pr_link_check.py` scripts is picking up missing documentation in the snippet index, which prevents CI from being green when adding a new archetype.

This PR loads the `exclude_path` section from the `lychee.toml`.

> [!IMPORTANT]
> Not that there is also a `exclude` pattern in `lychee.toml` that allows matching on the actual URL. This is not supported in this PR.


Wumpf:
> Not that there is also a exclude pattern in lychee.toml that allows matching on the actual URL. This is not supported in this PR.

lychee still reads the toml so the url matching exclusion is already supported :)

PR #11954, opened 2025-11-24
Implement the search service

swallez:
Implements the search service in `re_server`. Fixes [RR-2512](https://linear.app/rerun/issue/RR-2512/)

Indices are created as Lance tables. It therefore requires to compile with the `lance` feature which is not enabled by default.

The lance tables are created in a temporary directory that is deleted when the process exists.

Todo:
- [x] implement list and delete indices
- [x] ~define an index optimization policy and implement it. Possible candidates: debounce (i.e. no more than once every `x` seconds), or every `x` updates~ left as a TODO. Search works without it, it just won't be optimal, but also avoids the cost of updating/reconstructing the index (can be costly with vectors)
- [x] unit tests for vector search
- [x] integration tests in `re_redap_tests`

### DIFF
--- a/crates/store/re_server/src/chunk_index/mod.rs
+++ b/crates/store/re_server/src/chunk_index/mod.rs
@@ -242,10 +242,7 @@ impl DatasetChunkIndexes {
 
         for (index, segment_id, layer_name, chunk) in worklist {
             index
-                .store_chunks(
-                    vec![(segment_id.clone(), layer_name, chunk.clone())],
-                    true,
-                )
+                .store_chunks(vec![(segment_id.clone(), layer_name, chunk.clone())], true)
                 .await?;
         }
 
@@ -353,8 +350,8 @@ mod tests {
 
     use super::*;
     use arrow::array::{
-        ArrayRef, FixedSizeBinaryArray, FixedSizeListArray, FixedSizeListBuilder,
-        Float32Array, Float32Builder, ListBuilder, RecordBatch,
+        ArrayRef, FixedSizeBinaryArray, FixedSizeListArray, FixedSizeListBuilder, Float32Array,
+        Float32Builder, ListBuilder, RecordBatch,
     };
     use arrow::buffer::ScalarBuffer;
     use nohash_hasher::IntMap;


swallez:
Ready for review. @teh-cmc and @abey79 I added you as reviewers, feel free to delegate to someone else if needed.

nikolausWest:
> Indices are created as Lance tables. It therefore requires to compile with the lance feature which is not enabled by default.

If someone uses this feature after doing `pip install rerun-sdk`, what happens? Are python wheels built with that flag on by default? Do we give good descriptive error messages or does it just crash when you run? Does it fail on import?

swallez:
@nikolausWest it only affects `cargo install rerun-cli`, but it's actually "affected" as [this issue should be resolved now](https://github.com/rerun-io/rerun/pull/11592). I'll check and remove this restriction if this indeed works.

swallez:
Misunderstood "Fix cargo install requiring protoc" as "protoc is not needed anymore" while it was actually "adding the `lance` feature". So the issue still stands for people installing with `cargo install` but doesn't affect `pip install`.

crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
```diff
@@ -37,6 +38,55 @@ macro_rules! lazy_field_ref {
...
+//             column: value.column
+//                 .ok_or_else(|| missing_field!(crate::cloud::v1alpha1::IndexConfig, "column"))?,
+//             time_index: value.time_index
+//                 .ok_or_else(|| missing_field!(crate::cloud::v1alpha1::IndexConfig, "time_index"))?,
+//             properties: value.properties
+//                 .ok_or_else(|| missing_field!(crate::cloud::v1alpha1::IndexConfig, "properties"))?,
+//         })
+//     }
+// }
+
```
abey79: dead code

.gitignore
```diff
@@ -93,3 +93,6 @@ justfile
 
 # Lychee link checker cache
 .lycheecache
+
+# IntelliJ/PyCharm/RustRover
+.idea
```
abey79: We do have stuff committed in `.idea/` now. You should probably .gitignore specific files in `.idea/.gitignore` instead.

crates/store/re_server/src/chunk_index/mod.rs
```diff
@@ -0,0 +1,622 @@
...
+pub const FIELD_INSTANCE: &str = "instance";
+// Position of the instance in the column cell
+pub const FIELD_INSTANCE_ID: &str = "instance_id";
+
+/// An index for a column of a dataset's chunks
+#[cfg(feature = "lance")]
+pub struct Index {
+    config: IndexConfig,
+    // Mutex because we need to update the lance object after writing and checking out the latest version.
+    lance_dataset: parking_lot::Mutex<lance::dataset::Dataset>,
```
abey79: Shouldn't this be a `tokio::Mutex`?

crates/store/re_server/src/chunk_index/mod.rs
```diff
@@ -0,0 +1,622 @@
...
+                })
+                .map_err(Into::into)
+        });
+
+        Ok(tonic::Response::new(Box::pin(stream)))
+    }
+
+    // ----- Called by Dataset
+
+    pub async fn chunks_loaded(
```
abey79: took me a bit to form a mental model of what that does. Rename `on_layer_write` or something like that?

crates/store/re_server/src/chunk_index/mod.rs
```diff
@@ -0,0 +1,622 @@
...
+        entity_path: &EntityPath,
+        component: &ComponentIdentifier,
+    ) -> Option<Arc<Index>> {
+        let indexes = self.indexes.read().await;
+
+        if let Some(path_indexes) = indexes.get(entity_path) {
+            if let Some(component_index) = path_indexes.get(component) {
+                return Some(component_index.clone());
+            }
+        }
```
abey79: PSA: short circuit operator `?` works for `Option`, so `indexes.get(entity_path)?.get(component)?.clone()` should do the trick.

crates/store/re_server/src/chunk_index/mod.rs
```diff
@@ -0,0 +1,622 @@
...
+                .await?;
+        }
+
+        Ok(())
+    }
+
+    // ---- implementation
+
+    /// Get the index for a path and component, if any.
+    pub async fn get(
```
abey79: i think

```suggestion
    async fn get(
```

crates/store/re_server/src/chunk_index/mod.rs
```diff
@@ -0,0 +1,622 @@
...
+            if let Some(component_index) = path_indexes.get(component) {
+                return Some(component_index.clone());
+            }
+        }
+
+        None
+    }
+
+    /// Add an index to a dataset
+    pub(crate) async fn add_index(
```
abey79: same?
```suggestion
	async fn add_index(
```

crates/store/re_server/src/chunk_index/mod.rs
```diff
@@ -0,0 +1,622 @@
...
+
+        let mut indexes = self.indexes.write().await;
+
+        // Do we have it already?
+        if let Some(path_indexes) = indexes.get(entity_path)
+            && path_indexes.contains_key(component)
+        {
+            return Err(StoreError::DuplicateEntryNameError(format!(
+                "{entity_path}#{component}",
+            )));
```
abey79: This error specifically refers to "entry names", which is not what this is. You should introduce a `DuplicateSearchIndexError` instead.

crates/store/re_server/src/store/error.rs
```diff
@@ -36,6 +36,16 @@ pub enum Error {
     #[error(transparent)]
     DataFusionError(#[from] datafusion::error::DataFusionError),
 
+    #[error(transparent)]
+    ArrowError(#[from] arrow::error::ArrowError),
+
+    #[cfg(feature = "lance")]
+    #[error(transparent)]
+    LanceError(#[from] lance::Error),
```
abey79, line 50: I have been culprit of that many times, but have grown wary of these "type-based" error variant. They invariably translate to `internal` status and end up being very bad error messages.

Ideally, this error enum would only contain variant that correspond to "high-level" operation or error conditions that are understandable by the user ("index missing", "duplicate index", "failed to create index", etc.). I'm totally fine for these variant to contain a `Box<dyn Error>` or a `String`—who cares about the root error representation, it all turns to a string anyway.

In the code, I would expect the the low-level index processing stuff to return whatever makes more sense (lance error, etc.), and those error to be converted into `store::Error` by the layer that execute an operation that can be understood by the user. (This is also the point at which we might actually check what the low-level error is such that we can map it to a specific high-level error variant.)

crates/store/re_server/src/chunk_index/mod.rs
```diff
@@ -0,0 +1,622 @@
...
+        layer_name: &str,
+        _overwritten: bool,
+    ) -> Result<(), StoreError> {
+        Err(StoreError::IndexingError(
+            "Indexing not available: Lance feature not enabled".to_owned(),
+        ))
+    }
+}
+
+//---- Helper functions
```
abey79: rm

crates/store/re_server/src/chunk_index/mod.rs
```diff
@@ -0,0 +1,622 @@
...
+            let next = next?;
+            assert_eq!(next.column(0).len(), 1);
+            //println!("{:?}", next);
+        }
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_vector_search() -> anyhow::Result<()> {
```
abey79, line 369: Might be neat to have an insta snapshot test of the index for this one. It's often convenient to have a textual representation of the table somewhere.

crates/store/re_server/src/chunk_index/index.rs
```diff
@@ -0,0 +1,429 @@
...
+use arrow::datatypes::{DataType, Field, Schema};
+use arrow::error::ArrowError;
+use lance::deps::arrow_array::UInt8Array;
+use re_chunk_store::Chunk;
+use re_log_types::{EntityPath, TimelineName};
+use re_protos::cloud::v1alpha1::ext::{IndexConfig, IndexProperties};
+use re_protos::common::v1alpha1::ext::PartitionId;
+use re_types_core::ComponentIdentifier;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
```
abey79: I know its nit-picky and it should be tool-based, but fyi we have the informal (and not yet(?) enforced) convention of grouping `use` by `std` -> external -> `re_*` -> `crate`/`super`.

crates/store/re_server/src/store/dataset.rs
```diff
@@ -50,6 +51,7 @@ impl Dataset {
                 name,
                 details,
                 partitions: HashMap::default(),
+                indexes: DatasetChunkIndexes::new(id),
```
abey79, line 54: I believe it's incorrect to have the indexes in the inner object. We don't want to bump the modified timestamp when we change (e.g. delete) indexes. They should live the outer object instead. Might not change much since they have interior mutability, but just cleaner.

teh-cmc:
I can have a look after the first wave of comments has been addressed

crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
swallez: Oops. Removed in 3342657.

.gitignore
swallez: Ah indeed. Restored in 2278759.

crates/store/re_server/src/chunk_index/mod.rs
swallez: That one is a bit tricky. A shared instance of Lance's `Dataset` can be used for concurrent reads, but writes require a mutable reference. Also, once a write is done, we have to move the current version of the Dataset to the latest version, and therefore update the shared reference.

So we don't technically need a `tokio::Mutex` as we don't need to synchronize across async calls. Now `parking_lot::Mutex` is error-prone as exposing it allows holding a sync lock across async calls, which could starve the Tokio thread pool.

To avoid this I have introduced in e476803 a dedicated `ArcCell` data structure that hides the mutex and handles the shared reference and replacing the Lance instance after an index update.

We could also have re-created the Lance `Dataset` at every operation, but that would probably have added some latency what users may not expect with the local in-memory store. Not maybe I'm overthinking it and re-creating it on demand is ok.

crates/store/re_server/src/chunk_index/mod.rs
swallez: Renamed to `on_layer_added` to follow the naming of `add_layer` where it's called.

crates/store/re_server/src/chunk_index/mod.rs
swallez: Good catch. I always forget about it 😅

crates/store/re_server/src/chunk_index/mod.rs
swallez: Fixed. I know now that "entry" is a _catalog_ entry 😉

crates/store/re_server/src/store/error.rs
swallez: Good point. Fixed in 84fa8ca.

crates/store/re_server/src/store/dataset.rs
swallez: Ah indeed. Fixed.

crates/store/re_server/src/chunk_index/index.rs
swallez: Done. I tend to just let `cargo fmt` handle this…

crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
teh-cmc, line 1: Shouldn't there be a sibling PR on dataplatform to make use of these?

crates/store/re_redap_tests/src/tests/indexes.rs
```diff
@@ -71,8 +67,9 @@ pub async fn index_lifecycle(service: impl RerunCloudService) {
                     .unwrap_err()
                     .code();
 
-                // TODO(RR-2779): implement indexing support in OSS
-                code == tonic::Code::InvalidArgument || code == tonic::Code::Unimplemented
+                // TODO(RR-2779): OSS returns AlreadyExists.
+                // This is more precise and Rerun Cloud should be updated to return it.
```
teh-cmc, line 73: Ha. So there is a sibling PR coming later then I guess.

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -929,25 +931,58 @@ impl RerunCloudService for RerunCloudHandler {
...
+            if #[cfg(feature = "lance")] {
+                let store = self.store.read().await;
+                let entry_id = get_entry_id_from_headers(&store, &request)?;
+                let dataset = store.dataset(entry_id)?;
+
+                dataset.indexes().create_index(dataset, request.into_inner().try_into()?).await
+            } else {
+                Err(tonic::Status::unimplemented("create_index requires the `lance` feature"))
+            }
+        }
```
teh-cmc, line 950: Here and everywhere else, I really struggle to see the added value of  this `cfg_if` dependency. 

How is this better than:
```suggestion
        #[cfg(feature = "lance")]
        {
            let store = self.store.read().await;
            let entry_id = get_entry_id_from_headers(&store, &request)?;
            let dataset = store.dataset(entry_id)?;

            dataset.indexes().create_index(dataset, request.into_inner().try_into()?).await
        }

        #[cfg(not(feature = "lance"))]
        {
            Err(tonic::Status::unimplemented("create_index requires the `lance` feature"))
        }
```
which achieves the same without any extra dependency, without forcing all the code into a macro scope, and without the extra rightward drift?

crates/store/re_server/src/store/dataset.rs
```diff
@@ -364,7 +374,8 @@ impl Dataset {
             .map(|layer| layer.store_handle())
     }
 
-    pub fn add_layer(
+    /// Adds a layer
```
teh-cmc:
```suggestion
```

crates/store/re_server/src/store/error.rs
```diff
@@ -69,9 +85,13 @@ impl From<Error> for tonic::Status {
             | Error::EntryNameNotFound(_)
             | Error::PartitionIdNotFound(_, _)
             | Error::LayerNameNotFound(_, _, _)
+            | Error::IndexNotFound(_)
             | Error::ChunkNotFound(_) => Self::not_found(format!("{err:#}")),
 
             Error::DataFusionError(err) => Self::internal(format!("DataFusion error: {err:#}")),
+            Error::ArrowError(err) => Self::internal(format!("Arrow error: {err:#}")),
+            #[cfg(feature = "lance")]
+            Error::LanceError(err) => Self::internal(format!("Lance error: {err:#}")),
```
teh-cmc, line 94: Feature-flagged branches should really be separate from the rest (like at the bottom or something), it's super hard to parse otherwise.

crates/store/re_server/src/chunk_index/mod.rs
teh-cmc, line 1: lib.rs and mod.rs files should really be kept as small as possible and act as manifests for the packages within. I.e. they should only really contain module declarations, constants and public exports (see e.g. the lib.rs of this crate).

I should be able to look at this mod.rs file and immediately tell what this package is about and what it exposes.

crates/store/re_server/src/chunk_index/mod.rs
```diff
@@ -0,0 +1,647 @@
...
+    use re_chunk_store::external::re_chunk;
+    use re_chunk_store::external::re_chunk::{ChunkComponents, TimeColumn};
+    use re_chunk_store::{ChunkStore, ChunkStoreConfig};
+    use re_log_types::{EntryId, StoreId, StoreKind, TimeType, Timeline, TimelineName};
+    use re_protos::cloud::v1alpha1::VectorDistanceMetric;
+    use re_protos::cloud::v1alpha1::ext::{IndexColumn, IndexProperties, IndexQueryProperties};
+    use re_protos::common::v1alpha1::ext::{IfDuplicateBehavior, ScanParameters};
+    use re_types_core::{ChunkId, ComponentDescriptor, Loggable as _, SerializedComponentColumn};
+
+    const RRD: &str = "../../../tests/assets/rrd/snippets/views/timeseries.rrd";
```
teh-cmc: That's a very weird data dependency to take on. I'd strongly advise to just your own recording the fly here, similar to e.g. where re_redap_tests does.

crates/store/re_server/src/chunk_index/mod.rs
```diff
@@ -0,0 +1,647 @@
...
+
+            assert_eq!(distances.value(0), 0.0);
+            assert!(distances.value(1) > 0.0);
+            assert_eq!(instances.value(0), 3.0);
+        }
+
+        Ok(())
+    }
+
+    #[expect(dead_code)]
```
teh-cmc:
```suggestion
```

crates/store/re_server/src/chunk_index/mod.rs
```diff
@@ -0,0 +1,647 @@
...
+        dataset
+            .load_rrd(
+                Path::new(RRD),
+                None,
+                IfDuplicateBehavior::Error,
+                StoreKind::Recording,
+            )
+            .await?;
+
+        //dump_dataset_info(&dataset);
```
teh-cmc:
```suggestion
        if false {
            dump_dataset_info(&dataset);
        }
```

crates/store/re_server/src/chunk_index/index.rs
```diff
@@ -0,0 +1,429 @@
...
+            .with_dict_is_ordered(true),
+            Field::new_dictionary(
+                FIELD_RERUN_PARTITION_LAYER,
+                DataType::UInt8,
+                DataType::Utf8,
+                non_nullable,
+            )
+            .with_dict_is_ordered(true),
+            // Will be repeated, but Lance doesn't support dictionaries for FixedSizeBinary because
+            // of stringly-typed checks in lance_core::data_types (look for "Unsupported dictionary type")
```
teh-cmc, line 292: Instinctively, I wouldn't use dictionaries at all here, for any column. Lance already stores everything in its own compressed format anyway.

crates/store/re_server/src/chunk_index/mod.rs
teh-cmc: Actually, why is all of this not part of re_redap_tests to begin with?

crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
swallez: Yes, I'll take care of that once this PR is merged.

crates/store/re_redap_tests/src/tests/indexes.rs
swallez: Indeed, same as above, I'll take care of that once this PR is merged.

crates/store/re_server/src/rerun_cloud.rs
swallez: It's just syntactic sugar to have an `else` block instead of having to use brain cycles to write the negative `cfg` clause. Granted, in that case it's not really hard.

crates/store/re_server/src/chunk_index/index.rs
swallez: At least it saves some memory when creating the data to be indexed (all data resides in memory in the OSS server). I'll write some tests to see if it makes any difference on storage size.

swallez:
> My only worry here is: are the semantics of search part the Redap API? I.e. are Cloud & OSS expected to return similar results for similar queries? And if so, where are the redap_tests that perform actual searches?

@teh-cmc yes, Cloud & OSS should return similar results. @abey79 and I agreed that this PR is already big enough and redap_tests for this will come in a follow-up PR.

crates/store/re_server/src/store/error.rs
swallez: I added it below `DataFusionError` and `ArrowError` which form a group of "error in 3rd-party library" category. 

The `lance` feature has been introduced to avoid the requirement for `protoc` to exist when installing the viewer with `cargo install rerun-cli` (see https://github.com/rerun-io/rerun/pull/11592 for context).

Now IMHO the `lance` feature should go away:
* Lance is increasingly becoming a requirement for rerun-oss (some significant features will not be available without it)
* people using `pip install` will have a version compiled with the `lance` feature
* people using `cargo install` should be tech-savvy enough to know how to install `protoc`
* the compilation/installation error message gives clear instructions on how to do that:
  >   Error: Custom { kind: NotFound, error: "Could not find `protoc`. If `protoc` is installed, try setting the `PROTOC` environment variable to the path of the `protoc` binary. To install it on macOS, run `brew install protobuf`. It is also available at https://github.com/protocolbuffers/protobuf/releases  For more information: https://docs.rs/prost-build/#sourcing-protoc" }

crates/store/re_server/src/chunk_index/mod.rs
swallez: Why should leaf and non-leaf modules be treated differently?

The `index` and `search` sub-modules were introduced because the `chunk_index` module was getting too large. The main `chunk_index` now essentially contains the public API and the data structures of the chunk indexing features and calls either `index` for the write path or `search` for the read path.

Without the submodules, we would have a single 1200 lines `chunk_index.rs` leaf module. Would that be better?

crates/store/re_server/src/chunk_index/mod.rs
swallez: Fair enough, although test files are for testing… This was meant to test the call chain through `re_server/store/dataset::load_rrd()`.

Test removed in f319b12, this is something we'll exercise more extensively in both server implementations in `redap_tests` in a follow-up PR.

PR #11955, opened 2025-11-24
Reformat all our imports

emilk:
We cannot _enforce_ this formatting:

* this formatting settings requires nightly rust fmt
* we don't want to depend on nightly, even for formatting

…but enough people have complained about our inconsistent formatting that doing a quick pass won't hurt

Discussion: https://rerunio.slack.com/archives/C0440MD5UAV/p1762951699148509


Wumpf:
if we're reformatting everything (which.. oof.. can we do that over the weekend and not on Monday?) I'd like to suggest that we use a formatting setting that aggressively uses more lines rather than merging: the `imports_granularity` `crate` is nice to look at but from a merge conflict avoidance pov `module` or even `item` are _waaaaaay_ better.

PR #11958, opened 2025-11-24
Reduce data in traces and disable traces from specific Lance paths (unless overridden)

zehiko:
### What

Pretty much the title - in the effort to reduce the traces size we're removing some fields from the spans and also filtering more lance traces (at the info level).  Also enabled gzip while exporting spans.

### Testing done
Deployed on a cluster and ran registration

### DIFF
--- a/crates/utils/re_perf_telemetry/src/telemetry.rs
+++ b/crates/utils/re_perf_telemetry/src/telemetry.rs
@@ -1,11 +1,10 @@
 use opentelemetry::trace::TracerProvider as _;
-use opentelemetry_otlp::WithTonicConfig;
+use opentelemetry_otlp::WithTonicConfig as _;
 use opentelemetry_sdk::trace::{BatchConfigBuilder, BatchSpanProcessor};
 use opentelemetry_sdk::{
     logs::SdkLoggerProvider, metrics::SdkMeterProvider, trace::SdkTracerProvider,
 };
 use std::sync::Arc;
-use std::time::Duration;
 use tracing_subscriber::layer::SubscriberExt as _;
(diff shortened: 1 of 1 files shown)


crates/utils/re_perf_telemetry/src/telemetry.rs
```diff
@@ -321,10 +327,21 @@ impl Telemetry {
         let (tracer_provider, layer_traces_otlp) = if otel_enabled {
             let exporter = opentelemetry_otlp::SpanExporter::builder()
                 .with_tonic() // There's no good reason to use HTTP for traces (at the moment, that is)
+                .with_compression(opentelemetry_otlp::Compression::Gzip)
                 .build()?;
 
+            let batch_config = BatchConfigBuilder::default()
+                .with_max_queue_size(5_000)
+                .with_max_export_batch_size(2048)
+                .with_scheduled_delay(Duration::from_secs(10))
```
teh-cmc: Can you add add some comments for each these things please, namely:
* What does it do?
* Why do we do it?
* What's the tradeoff?

Looking at https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/trace/struct.BatchConfigBuilder.html, my understanding is that these 3 lines have _a lot_ of impact.

zehiko: ah of course, in all the noise and gazzillion spans I forgot to do that.

zehiko:
> Ha, the `with_scheduled_delay` is gone now. That seems less scary.

yep, I also realized it didn't make sense to actually increase that value

PR #11962, opened 2025-11-24
Allow visualizers to report runtime errors

Wumpf:
Visualizers can fail for all sort of reasons, either as a whole for everything in a View or just for individual entities (visualizer execution is always in bulk for practical reasons). This change makes it possible to surface these errors directly in the UI.

To do so, the output of a visualizer becomes more structured and gets passed through the Viewer.

This PR already makes some use of this new ability, notably to show the various ways a Pinhole can be un-visualizable:
image

There's a lot more potential in this direction, but I wanted to keep this iteration minimal. Future work:
* emit all sort of other useful errors. This PR covers mostly transforms & pinhole
* make error reporting pretty
* make error reporting discoverable (highlight in blueprint tree for instance)
* use new structured output to remove "visualizer data" pointers

### DIFF
--- a/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
+++ b/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
@@ -149,53 +149,13 @@ impl ViewContextSystem for TransformTreeContext {
 }
 
 impl TransformTreeContext {
-    /// Retrieves the transform info for the given entity, or reports an error and returns `None` if
-    /// the entity was either not known or the transform couldn't be computed.
+    /// Returns a transform info describing how to get to the view's target frame for the given entity.
     #[inline]
-    pub fn transform_info_for_entity_or_report_visualizer_error(
+    pub fn target_from_entity_path(
         &self,
-        entity_path: &re_log_types::EntityPath,
-        output: &mut re_viewer_context::VisualizerExecutionOutput,
-    ) -> Option<&re_tf::TransformInfo> {
-        match self.transform_infos.get(&entity_path.hash()) {
-            None => {
-                output.report_error_for(
-                    entity_path.clone(),
-                    "No transform relation known for this entity.",
-                );
-                None
-            }
-
-            Some(Err(re_tf::TransformFromToError::NoPathBetweenFrames { .. })) => {
-                // TODO(RR-2997): Pretty print out the frames involved.
-                output.report_error_for(
-                    entity_path.clone(),
-                    "No transform path to the view's origin frame.",
-                );
-                None
-            }
-
-            Some(Err(re_tf::TransformFromToError::UnknownTargetFrame { .. })) => {
-                // The target frame is the view's origin.
-                // This means this could be hit if the view's origin frame doesn't show up in any data.
-                // TODO(RR-2997): Pretty print out the frames involved.
-                output.report_error_for(entity_path.clone(), "The view's origin frame is unknown.");
-                None
-            }
-
-            Some(Err(re_tf::TransformFromToError::UnknownSourceFrame { .. })) => {
-                // Unclear how we'd hit this. This means that when processing transforms we encountered a coordinate frame that the transform cache didn't know about.
-                // That would imply that the cache is lagging behind.
-                // TODO(RR-2997): Pretty print out the frames involved.
-                output.report_error_for(
-                    entity_path.clone(),
-                    "The entity's coordinate frame is unknown.",
-                );
-                None
-            }
-
-            Some(Ok(transform_info)) => Some(transform_info),
-        }
+        entity_path_hash: EntityPathHash,
+    ) -> Option<&Result<re_tf::TransformInfo, re_tf::TransformFromToError>> {
+        self.transform_infos.get(&entity_path_hash)
     }
 
     /// Returns the properties of the pinhole tree root at given frame if the entity's root is a pinhole tree root.
--- a/crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
+++ b/crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
@@ -11,7 +11,10 @@ use re_viewer_context::{
     VisualizableFilterContext, VisualizerExecutionOutput, VisualizerQueryInfo, VisualizerSystem,
 };
 
-use crate::{contexts::TransformTreeContext, view_kind::SpatialViewKind};
+use crate::{
+    contexts::TransformTreeContext, view_kind::SpatialViewKind,
+    visualizers::utilities::transform_info_for_entity_or_report_error,
+};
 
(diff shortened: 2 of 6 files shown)


crates/viewer/re_viewer_context/src/typed_entity_collections.rs
```diff
@@ -60,14 +61,56 @@ impl std::ops::Deref for VisualizableEntities {
...
+// Manual default impl, otherwise T: Default would be required.
+impl<T> Default for PerVisualizer<T> {
+    fn default() -> Self {
+        Self(IntMap::default())
+    }
+}
+
+/// Like [`PerVisualizer`], but ensured that all visualizers are relevant for the given view class.
+#[derive(Debug)]
+pub struct PerVisualizerInViewClass<T> {
```
Wumpf, line 89: FYI this one is gonna be more heavily used in the upcoming visualizability-simplification

Ported it "back" into this branch because it came in handy as well

Wumpf: .. but I'm actually a bit skeptical about it myself. The problem I wanted to express is that `PerVisualizer` alone can sometimes be misleading, as gonna be more obvious in said follow-up

Cargo.lock
```diff
@@ -9967,7 +9966,6 @@ dependencies = [
  "re_chunk_store",
  "re_entity_db",
  "re_log_types",
- "re_renderer",
```
grtlr, line 9970: A surprising turn of events

crates/viewer/re_viewer_context/src/typed_entity_collections.rs
grtlr: Ah thanks for the clarification!

crates/viewer/re_selection_panel/src/selection_panel.rs
```diff
@@ -635,8 +635,19 @@ fn entity_selection_ui(
         .cloned();
 
     if let Some(view) = viewport.view(view_id) {
-        let view_ctx = view.bundle_context_with_states(ctx, view_states);
-        visualizer_ui(&view_ctx, view, entity_path, ui);
+        let class = view.class(ctx.view_class_registry());
+        view_states.ensure_state_exists(*view_id, class);
+        let view_state = view_states
+            .get(*view_id)
+            .expect("State got created just now"); // Convince borrow checker we're not mutating `view_states` anymore.
```
emilk, line 642:
```suggestion
        let view_state = view_states.ensure_state_exists(*view_id, class);
```

emilk (commented):
Oops - I see now that @grtlr is already reviewing. 

Image

crates/viewer/re_selection_panel/src/selection_panel.rs
```diff
@@ -635,8 +635,19 @@ fn entity_selection_ui(
...
+        let view_ctx = view.bundle_context_with_state(ctx, view_state);
+
+        let empty_errors = PerVisualizer::default();
+        let visualizer_errors = view_states
+            .visualizer_errors(*view_id)
+            .unwrap_or(&empty_errors);
```
grtlr, line 648:
```suggestion
        let empty_errors = PerVisualizer::default();
        let visualizer_errors = view_states
            .visualizer_errors(*view_id)
            .unwrap_or_default();
```

Might be missing something, but would this work as well?

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -149,13 +149,53 @@ impl ViewContextSystem for TransformTreeContext {
...
+    /// Retrieves the transform info for the given entity, or reports an error and returns `None` if
+    /// the entity was either not known or the transform couldn't be computed.
     #[inline]
-    pub fn transform_info_for_entity(
+    pub fn transform_info_for_entity_or_report_visualizer_error(
         &self,
-        entity_path: EntityPathHash,
+        entity_path: &re_log_types::EntityPath,
+        output: &mut re_viewer_context::VisualizerExecutionOutput,
     ) -> Option<&re_tf::TransformInfo> {
```
grtlr: Nit: That function signature reads a bit clumsy. It looks like we're only ever returning a single error, why not make that function return a `Result`?

Cargo.lock
Wumpf: yeah. really weird. it's just because some types are no longer spelled out directly 🙄

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
Wumpf: yeah agreed, this is very heavy handed atm. Also polluting the transform context with knowledge about execution output is weird.

Wumpf:
> Part of the code reads a bit clumsy but this might come from not wanting to make too much changes at once?

there is for sure more refactors I'd like to do related to this, but apart from the ones you already pointed out on the review I'd like to learn what other areas invoked that sense to you. Just so we don't assume to talk about the same thing while not actually :)

crates/viewer/re_selection_panel/src/selection_panel.rs
Wumpf: no, we're getting a reference here because I didn't want to copy things out.
Guess a `clone` wouldn't be too bad though 🤷

crates/viewer/re_selection_panel/src/selection_panel.rs
Wumpf: well, I tried that. But the problem is that `ensure_state_exists` takes a `&mut` and that means that I can't borrow `view_states` read-only down the line. Thus the comment.

Any better idea how to solve that kind of thing? Really I want `get_or_create` but have the return value not depend on the mutable borrow

grtlr:
I just had another look, sorry for the cliffhanger. I think there were mostly two things that threw me off (no of them are blocking in any ways; mostly personal preference):

1. It should be the visualizers job alone to turn structured errors into human readable errors because that is just another form of "visualization" and visualizers show have the most context available. Right now the `TransformTreeContext` is doing stringly conversions.
2. Maybe a bit too philosophical but anyways: The `Result::Ok` variant containing `errors_per_entity` reads a bit weird. To me `Ok` means _everything_ went well, right now it means something like "I was able to produce something, with some caveats" (which also makes sense). But, moving all error reporting to the `Result::err` variant means introducing `QueueableDrawData` there, which you could argue is confusing for similar reasons, so 🤷.

crates/viewer/re_selection_panel/src/selection_panel.rs
emilk: No :(

Wumpf:
Addressed (1) by moving out the error handling to `transform_retrieval.rs` which happens to be the place where in the next pr more error handling lives

PR #11968, opened 2025-11-25
Modernize `ros_node` example and documentation

MichaelGrupp:
ros_node_example

* remove custom URDF loader, use builtin loader
* remove TF buffer and log directly transforms with parent/child frame
* port to ROS Kilted (Ubuntu 24)
* use Turtlebot 4 simulation instead of TB3
  * TB4 sim has a working depth camera, TB3 doesn't seem to have the RGBD cam anymore in Kilted
  * more interesting realistic warehouse environment
* several fixes, e.g.:
  * removed outdated `rr.log` API use
  * removed outdated comments mentioning ancient API like `rr.log_image`
  * missing / broken dependencies
  * deprecation warnings from other ROS modules
  * ...
* documentation
  * up-to-date
  * slimmer documentation page
    * only code explanations for interesting & important concepts
    **note:** we can't use snippets here because of ROS dependency
    * removed duplicate installation guide and refer to README instead
  * updated screenshot showing the new simulation

### DIFF
--- a/examples/python/ros_node/README.md
+++ b/examples/python/ros_node/README.md
@@ -37,8 +37,11 @@ For more information on future improved ROS support, see tracking issue: [#1527]
 
 > NOTE: Unlike many of the other examples, this example requires a system installation of ROS
 in addition to the packages from requirements.txt.
+>
+> The commands are focused on the official ROS installation path on an Ubuntu distro, but should also work analogously if you installed ROS through [Robostack](https://robostack.github.io/index.html).
 
 This example was developed and tested on top of [ROS2 Kilted Kaiju](https://docs.ros.org/en/kilted/index.html), with [Nav2](https://docs.nav2.org/) and the Turtlebot 4 simulation example from the [nav2_bringup](https://github.com/ros-navigation/navigation2/tree/main/nav2_bringup) package.
+If you use another version of ROS, e.g. [Jazzy](https://docs.ros.org/en/jazzy/index.html), you should be able to just replace `kilted` with `jazzy` in the commands below.
 
 Installing ROS is outside the scope of this example, but you will need the equivalent of the following packages:
 ```


examples/python/ros_node/main.py
```diff
@@ -139,97 +121,66 @@ def __init__(self) -> None:
...
-        self.model.fromCameraInfo(info)
+        self.model.from_camera_info(info)
 
         rr.log(
             "map/robot/camera/img",
             rr.Pinhole(
                 resolution=[self.model.width, self.model.height],
                 image_from_camera=self.model.intrinsicMatrix(),
+                parent_frame=info.header.frame_id,
+                child_frame=info.header.frame_id + "_image_plane",
```
MichaelGrupp, line 107: This is correct, but the Pinhole API was broken. We need to wait until this patch is released: https://github.com/rerun-io/rerun/pull/12360

MichaelGrupp: Update: see TODO comment

examples/python/ros_node/requirements.txt
```diff
@@ -1,6 +1,10 @@
-numpy
+# NOTE: numpy has to be downgraded to be compatible with ROS packages
+# that were built against the 1.x version of python3-numpy in Ubuntu 24.04.
+
+numpy<2.0
```
MichaelGrupp, line 4: This is ugly, but I don't have a better solution. Rerun pulls in numpy>2.0 and we downgrade this here afterwards. This is less painful than [installing with --no-deps](https://github.com/rerun-io/rerun/pull/11100#issuecomment-3257694687) and causing a ton of other dependency issues.

examples/python/ros_node/main.py
MichaelGrupp: Changed it back now that we have the patch ready.

nikolausWest:
Didn't read all the code but docs look good to me!

examples/python/ros_node/requirements.txt
ntjohnson1: This kind of raises the question if we should consider properly supporting numpy 1.0. The biggest headache is the testing but with the updated uv support that shouldn't be too bad. The fact this still works says our informal support for backwards compatibility it decent.

MichaelGrupp: At least it will stay relevant for a while for users that use Python packages from a non-bleeding edge distro. If not strictly necessary outside testing, I would lift that requirement but I also don't know all the details why it was set that way.

PR #11970, opened 2025-11-25
Instanced drawing of `TransformAxes3D` via `InstancePose3D`

grtlr:
### Related

* Closes RR-2916.

### What

https://github.com/user-attachments/assets/7f35a448-5b42-4ef4-a42d-f76a5dddb1a1

This PR makes it possible to draw instanced `TransformAxes3D`. This is quite helpful for MCAP messages that contain multiple instance poses at once.

### DIFF
--- a/examples/rust/mcap_protobuf/src/main.rs
+++ b/examples/rust/mcap_protobuf/src/main.rs
@@ -2,7 +2,7 @@ use arrow::array::{Float32Array, Float64Array, ListArray};
 
 use re_log_types::TimeType;
 use rerun::{
-    EncodedImage, InstancePoses3D, Points3D, Transform3D, TransformAxes3D, VideoStream,
+    EncodedImage, InstancePoses3D, Transform3D, TransformAxes3D, VideoStream,
     dataframe::EntityPathFilter,
     external::re_log,
     lenses::{Lens, LensesSink, Op, OpError},


Wumpf:
screenshot plz :)

crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
```diff
@@ -39,6 +39,7 @@ impl VisualizerSystem for TransformAxes3DVisualizer {
         query_info.required = RequiredComponents::Any(
             Transform3D::all_component_identifiers()
                 .chain(TransformAxes3D::all_component_identifiers())
+                .chain(InstancePoses3D::all_component_identifiers())
                 .chain(CoordinateFrame::all_component_identifiers())
                 .collect(),
```
Wumpf, line 47: so as-is I actually can't add one of these if there's only a Pinhole (without extrinsics). I think we used to allow that 🤔 
Not sure if that's a serious problem.

grtlr:
Was creating a video just as you wrote this.

crates/viewer/re_view_spatial/tests/transform_clamping.rs
Wumpf, line 1: the tests themselves are great, but I'm a bit on the fence to re-use the transform clamping test suite for this. this used to be about, well, clamping of transform properties.

edit: but we do have very nice facilities here. maybe just needs a rename or something 🤔

crates/viewer/re_view_spatial/tests/transform_clamping.rs
```diff
@@ -160,6 +160,35 @@ fn test_transform_clamping(base_transform: BaseTransform) {
...
+                                (-1.0, -1.0, 0.0),
+                                (1.0, -1.0, 0.0),
+                                (-1.0, 1.0, 0.0),
+                                (1.0, 1.0, 0.0),
+                            ])
+                            .with_scales([
+                                (1.75, 1.75, 1.75),
+                                (1.5, 1.5, 1.5),
+                                (1.0, 1.0, 1.0),
+                                (0.5, 0.5, 0.5),
```
Wumpf: in the spirit of transform clamping it would be interesting to still test out the existing clamping semantics
i.e. if you have more translations than scales, we still apply clamping logic there since we treat the "poses" as a whole as a single unit and within that unit clamping applies

examples/rust/mcap_protobuf/src/main.rs
Wumpf, line 1: don't mind snowballing this in here, but was that intentional? :)

examples/rust/mcap_protobuf/src/main.rs
```diff
@@ -95,12 +104,14 @@ fn main() -> anyhow::Result<()> {
     // Name of the timestamp field in Foxglove messages, and name of the corresponding Rerun timeline.
     const TIME_NAME: &str = "timestamp";
 
-    // The following could be improved with columnar archetype APIs.
-    let dummy_point = Points3D::new([[0.0f32, 0.0, 0.0]])
-        .columns_of_unit_batches()
-        .unwrap()
-        .next()
-        .unwrap();
+    // TODO(grtlr): This can be removed once we have proper support for Pinhole.
```
Wumpf: not quire following how this one relates to pinholes 🤔 
is this meant to be about the pose <-> frame_id rather?

crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
```diff
@@ -78,28 +79,35 @@ impl VisualizerSystem for TransformAxes3DVisualizer {
...
                 continue;
             };
-            let world_from_obj = if let Some(pinhole_tree_root_info) =
-                transforms.pinhole_tree_root_info(transform_info.tree_root())
-            {
-                if transform_info.tree_root()
-                    == re_tf::TransformFrameIdHash::from_entity_path(&data_result.entity_path)
+
+            // Determine which transforms to draw axes at.
+            // For pinhole cameras, we draw at the pinhole location only.
```
Wumpf, line 90: this gives rise to an intersting corner case: several poses where we also have a camera. Should we draw poses + pinhole location instead? 🤔

Wumpf (approved):
some food for thought, but looking good!

examples/rust/mcap_protobuf/src/main.rs
grtlr: If we remove this primitive entirely then the Spatial View 3D heuristics won't have anything to pick up and therefore not spawn a 3D view, which is annoying. Once we support the full scene, other stuff should get picked up (when we get to heuristics and we can remove this again).

crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
grtlr: Hmm good question. I think we can punt on that for now. Most files that I looked at seem to have a 1:1 mapping between pinhole and instance pose.

examples/rust/mcap_protobuf/src/main.rs
grtlr: Yes, intentional, this PR improves the visualization quite a bit (see video).

PR #11931, opened 2025-11-20
Support LeRobotDataset v3.0

oxkitsune:
### Related

* Closes #11678

### What

This adds support for the LeRobotDataset v3.0 (https://github.com/huggingface/lerobot/pull/1412) format to our LeRobot dataset loader. Additionally this now loads feature statistics such as min, max, mean values as raw arrow data, making it easier to query data out of a loaded dataset if needed.

Still left to do before this can land:
- [x] Make sure the checklist item still passes, and add a v3 dataset item
- [x] Another pass over the new code structure, I moved lots of things around and have a feeling some things aren't optimal
- [x] 	Compatibility testing, so far this has been tested on [lerobot/aloha_mobile_cabinet](https://huggingface.co/datasets/lerobot/aloha_mobile_cabinet) and [lerobot/aloha_mobile_elevator](https://huggingface.co/datasets/lerobot/aloha_mobile_elevator). I want to try some third party datasets as well.

PR #11972, opened 2025-11-25
Add utility for sampling video from `VideoDataDescription`

oxkitsune:
### Related

* Part of #11931 

### What

This adds a utility to sample video data from a `VideoDataDescription` given specific timestamps in the format expected by `VideoStream`. It will insert the required data at the start of the returned sample in order to successfully decode the sample on playback.


crates/utils/re_video/src/demux/mod.rs
```diff
@@ -347,6 +352,99 @@ impl VideoDataDescription {
 
         Ok(())
     }
+
+    /// Returns the encoded bytes for a sample in the format expected by [`VideoCodec`].
+    ///
+    /// MP4 stores H.264/H.265 samples using AVCC/HVCC length-prefixed NALs and relies on container
+    /// metadata for SPS/PPS/VPS. AV1 samples are stored as-is.
```
Wumpf:
```suggestion
    /// * H.264/H.265: MP4 stores samples using AVCC/HVCC length-prefixed NALs and relies on container
    /// metadata for SPS/PPS/VPS. This method makes sure to unpack this.
    /// * AV1 samples are stored as-is.
    /// * VP8/VP9: Not yet supported
```

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -347,6 +352,99 @@ impl VideoDataDescription {
...
+    pub fn sample_data_in_stream_format(
+        &self,
+        chunk: &crate::Chunk,
+        annexb_state: Option<&mut AnnexBStreamState>,
```
Wumpf: looking at this again I wonder whether we actually need `AnnexBStreamState`. All it does right now is to not emit SPS/PPS if there's two sync frames in a row.
Which.. is weird? Why are we even doing that? If we were to emit too much reentrancy information how would that hurt us actually?

Worth a stab because it would make this method also muuuch nicer. But I realize this goes beyond of what you were trying to achieve here, so no need to dwell on it for too long

crates/viewer/re_view_spatial/tests/video.rs
```diff
@@ -351,3 +351,53 @@ fn test_video_stream_codec_h265() {
...
 }
+
+/// Helper function to check if data contains Annex B start codes
+fn has_annexb_start_codes(data: &[u8]) -> bool {
+    const ANNEXB_NAL_START_CODE: &[u8] = &[0x00, 0x00, 0x00, 0x01];
+    data.windows(4).any(|w| w == ANNEXB_NAL_START_CODE)
+}
+
+#[test]
+fn test_full_video_sampling() {
```
Wumpf: Can we plz move that est up into re_video? There's no apparent reason really to have this be part of the end to end spatial view tests. The one drawback I can see right away is that we have to duplicate `video_test_file_mp4`, but that seems alright

crates/viewer/re_view_spatial/tests/video.rs
```diff
@@ -351,3 +351,53 @@ fn test_video_stream_codec_h265() {
...
+    let video_path = video_test_file_mp4(VideoCodec::H264, false);
```
Wumpf: let's loop over all available codecs

crates/utils/re_video/src/demux/mod.rs
oxkitsune: I got rid of the argument, and just create a new `AnnexBStreamState` everytime. Tested it in the lerobot loader and seems to work.

Wumpf: good enough for now I guess :)

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -1089,4 +1082,97 @@ mod tests {
...
+        assert!(non_idr_count > 0, "Should have at least one non-IDR frame");
+    }
+
+    #[test]
+    fn test_full_video_sampling_all_codecs() {
+        // TODO(#10186): Add VP9 once we have it.
+        for codec in [VideoCodec::H264, VideoCodec::H265, VideoCodec::AV1] {
+            test_video_codec_sampling(codec, false);
+        }
+    }
```
Wumpf, line 1177: find it rather strange to have these be part of `impl VideoDescription`. Can you please move them out to a `#[cfg(test)] mod tests {`?

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -1089,4 +1082,97 @@ mod tests {
...
+            &format!("test_{codec:?}_video_sampling"),
+        )
+        .unwrap();
+
+        let buffers: crate::StableIndexDeque<&[u8]> = std::iter::once(data.as_slice()).collect();
+
+        let mut idr_count = 0;
+        let mut non_idr_count = 0;
+
+        const ANNEXB_NAL_START_CODE: &[u8] = &[0x00, 0x00, 0x00, 0x01];
```
Wumpf, line 1138: how often do we have this constant by now? two times alone in this the diff of this file 😛

Wumpf (approved):
thanks for the improvements!

crates/utils/re_video/src/demux/mod.rs
oxkitsune: they are in a `#[cfg(test)] mod tests` riiight?

Wumpf: oops. sorry!

PR #11973, opened 2025-11-25
Relax visualizability constraints and handle invalid transforms as visualization errors instead

Wumpf:
### Related

* Fix RR-2008

### What

Renames `MaybeVisualizable` to `Visualizable`, removing the old machinery previously associated with per-frame visualizability. This makes the question of whether something is visualizable completely static and much much simpler!

The "maybe visualizable" -> "visualizable" transition previously entailed checking for a static assumed topology and how it interacts with 2D & 3D content. Now that we want to regard the topology as dynamic, this no longer works out all that well and we rather use runtime visualizer errors for this.

The direct effect of that is that entities with 2D/3D content will always show up in both 2D/3D views, but then may end up not being visualizable after all.

Note that static topology is still used for various heuristics. These remain unchanged in this PR.

---

Example:

```py
rr.init("example spawn=True)
rr.log("boxes2d", rr.Boxes2D(mins=[-1, -1], sizes=[2, 2]))
rr.log("boxes3d", rr.Boxes3D(half_sizes=[2.0, 2.0, 1.0]))
```

*unchanged*: Viewer spawns two views.  2D view with space origin boxes2d, 3D view with space origin /. Both use the standard entity path filter + $origin/**
*before*: Blueprint panel lists only boxes3d for the 3D view
*after*: Blueprint panel lists both  boxes3d & boxes2d. However, boxes2d reports a visualization error.

image

---

Commit by commit - I separated removal of the old system, addition of new runtime errors and update of tests.

### DIFF
--- a/crates/viewer/re_view_spatial/src/visualizers/utilities/transform_retrieval.rs
+++ b/crates/viewer/re_view_spatial/src/visualizers/utilities/transform_retrieval.rs
@@ -59,7 +59,7 @@ pub fn transform_info_for_entity_or_report_error<'a>(
             let target = transform_context.format_frame(*target);
             output.report_error_for(
                 entity_path.clone(),
-                format!("No transform path from {src} to the view's origin frame ({target})."),
+                format!("No transform path from {src:?} to the view's origin frame ({target:?})."),
             );
             None
         }
@@ -70,7 +70,7 @@ pub fn transform_info_for_entity_or_report_error<'a>(
             let target = transform_context.format_frame(*target);
             output.report_error_for(
                 entity_path.clone(),
-                format!("The view's origin frame {target} is unknown."),
+                format!("The view's origin frame {target:?} is unknown."),
             );
             None
         }
@@ -81,7 +81,7 @@ pub fn transform_info_for_entity_or_report_error<'a>(
             let src = transform_context.format_frame(*src);
             output.report_error_for(
                 entity_path.clone(),
-                format!("The entity's coordinate frame {src} is unknown."),
+                format!("The entity's coordinate frame {src:?} is unknown."),
             );
             None
         }
@@ -111,10 +111,13 @@ fn is_valid_space_for_content(
     //
     // Everything in this 3D view is technically 2D already, but we still have the 3D controls etc.
     // (We can however, still show some "agnostic" content like the Pinhole itself)
-    if view_kind == SpatialViewKind::ThreeD && target_frame_pinhole_root.is_some() {
+    if view_kind == SpatialViewKind::ThreeD
+        && let Some(target_frame_pinhole_root) = target_frame_pinhole_root
+    {
+        let origin = transform_context.format_frame(target_frame_pinhole_root);
         output.report_error_for(
             entity_path.clone(),
-            "The origin of the 3D view is under pinhole projection which is not supported by most 3D visualizations.",
+            format!("The origin of the 3D view ({origin:?}) is under pinhole projection which is not supported by most 3D visualizations."),
         );
         return false;
     }
--- a/crates/viewer/re_view_spatial/tests/snapshots/topology_errors__2d_view_at_pinhole.snap
+++ b/crates/viewer/re_view_spatial/tests/snapshots/topology_errors__2d_view_at_pinhole.snap
@@ -6,7 +6,7 @@ PerVisualizer(
     {
         "Ellipsoids3D": PerEntity(
             {
-                /disconnected_entity: "No transform path to the view's origin frame.",
+                /disconnected_entity: "No transform path from \"disconnected\" to the view's origin frame (\"tf#/pinhole_workaround/pinhole_entity\").",
             },
         ),
         "Boxes3D": PerEntity(
--- a/crates/viewer/re_view_spatial/tests/snapshots/topology_errors__2d_view_at_root.snap
+++ b/crates/viewer/re_view_spatial/tests/snapshots/topology_errors__2d_view_at_root.snap
@@ -11,7 +11,7 @@ PerVisualizer(
         ),
         "Ellipsoids3D": PerEntity(
             {
-                /disconnected_entity: "No transform path to the view's origin frame.",
+                /disconnected_entity: "No transform path from \"disconnected\" to the view's origin frame (\"world\").",
             },
         ),
         "Boxes3D": PerEntity(
--- a/crates/viewer/re_view_spatial/tests/snapshots/topology_errors__3d_view_at_pinhole.snap
+++ b/crates/viewer/re_view_spatial/tests/snapshots/topology_errors__3d_view_at_pinhole.snap
@@ -6,22 +6,22 @@ PerVisualizer(
     {
         "Points3D": PerEntity(
             {
-                /points3d_entity: "The origin of the 3D view is under pinhole projection which is not supported by most 3D visualizations.",
+                /points3d_entity: "The origin of the 3D view (\"tf#/pinhole_workaround/pinhole_entity\") is under pinhole projection which is not supported by most 3D visualizations.",
             },
         ),
         "Ellipsoids3D": PerEntity(
             {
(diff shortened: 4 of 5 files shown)


crates/viewer/re_context_menu/src/actions/add_entities_to_new_view.rs
```diff
@@ -89,34 +88,21 @@ fn recommended_views_for_selection(ctx: &ContextMenuContext<'_>) -> IntSet<ViewC
...
         // "visualizable" with it. By "visualizable" we mean that either the entity itself, or any
         // of its sub-entities, are visualizable.
 
         let covered = entities_of_interest.iter().all(|entity| {
-            visualizable_entities.0.iter().any(|(_, entities)| {
-                entities
-                    .0
-                    .iter()
-                    .any(|visualizable_entity| visualizable_entity.starts_with(entity))
+            visualizable_entities.iter().any(|(visualizer, entities)| {
```
grtlr: I wonder if we should encapsulate some of the these decisions in our context object? Might be better than having those access patterns sprinkled throughout the codebase.

crates/viewer/re_context_menu/src/actions/add_entities_to_new_view.rs
```diff
@@ -89,34 +88,21 @@ fn recommended_views_for_selection(ctx: &ContextMenuContext<'_>) -> IntSet<ViewC
 
     let view_class_registry = ctx.viewer_context.view_class_registry();
     let recording = ctx.viewer_context.recording();
-    let maybe_visualizable_entities = view_class_registry
-        .maybe_visualizable_entities_for_visualizer_systems(recording.store_id());
+    let visualizable_entities =
+        view_class_registry.visualizable_entities_for_visualizer_systems(recording.store_id());
```
grtlr: Don't we have this on the `viewer_context` already? Also see comment below.

crates/viewer/re_selection_panel/src/visualizer_ui.rs
```diff
@@ -612,32 +605,18 @@ fn menu_add_new_visualizer(
...
-        );
+    let view_class = ctx.view_class_entry();
 
-    visualizable_entities
+    ctx.viewer_ctx
+        .visualizable_entities_per_visualizer
         .iter()
         .filter(|&(vis, ents)| {
-            ents.contains(&data_result.entity_path) && !active_visualizers.contains(vis)
+            view_class.visualizer_system_ids.contains(vis)
```
grtlr: See comment above. 💊

crates/viewer/re_view_spatial/tests/topology_errors.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    test_context.log_entity("pinhole_workaround/pinhole_entity", |builder| {
+        builder.with_archetype_auto_row(
+            TimePoint::STATIC,
+            &archetypes::Pinhole::from_focal_length_and_resolution([1.0, 1.0], [100.0, 100.0]),
+            // TODO(RR-2680): set child/parent frames.
+        )
+    });
+
+    // TODO(RR-2997): Space origin can only be an entity, so we need something that points to the world frame.
```
grtlr: Is this the correct `TODO`? If so, how are they related?

crates/viewer/re_view_spatial/tests/topology_errors.rs
```diff
@@ -0,0 +1,305 @@
...
+    test_context.register_view_class::<re_view_spatial::SpatialView2D>();
+
+    setup_scene(&mut test_context);
+
+    // Most scenarios run into the unreachable & misplaced 3D entities not being visualizable!
+    let disconnected_entity_error = ExpectedError {
+        visualizer: "Ellipsoids3D",
+        entity_path: "disconnected_entity",
+        message_substring: "No transform path to the view's origin frame.",
+    };
```
grtlr: With the error reporting from

* #11985 

I wonder if these should become _inline_ `insta` snapshot tests? That would probably mean unrolling the entire test though, so maybe we just need some better constructors.

crates/viewer/re_view_tensor/src/view_class.rs
```diff
@@ -103,10 +101,14 @@ Set the displayed dimensions in a selection panel.",
     fn choose_default_visualizers(
         &self,
         entity_path: &EntityPath,
-        _maybe_visualizable_entities_per_visualizer: &PerVisualizer<MaybeVisualizableEntities>,
-        visualizable_entities_per_visualizer: &PerVisualizer<VisualizableEntities>,
+        visualizable_entities_per_visualizer: &PerVisualizerInViewClass<VisualizableEntities>,
         _indicated_entities_per_visualizer: &PerVisualizer<IndicatedEntities>,
     ) -> re_viewer_context::SmallVisualizerSet {
+        debug_assert_eq!(
```
grtlr: Should we move this assert to `update_overrides_recursive` (or in the constructor of `DataQueryPropertyResolver`) in `view_contents.rs`?

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -387,10 +387,14 @@ impl ViewClass for TimeSeriesView {
     fn choose_default_visualizers(
         &self,
         entity_path: &EntityPath,
-        _maybe_visualizable_entities_per_visualizer: &PerVisualizer<MaybeVisualizableEntities>,
-        visualizable_entities_per_visualizer: &PerVisualizer<VisualizableEntities>,
+        visualizable_entities_per_visualizer: &PerVisualizerInViewClass<VisualizableEntities>,
         indicated_entities_per_visualizer: &PerVisualizer<IndicatedEntities>,
     ) -> SmallVisualizerSet {
+        debug_assert_eq!(
```
grtlr: it's repeated quite often

crates/viewer/re_viewer_context/src/view/view_class_registry.rs
```diff
@@ -331,6 +341,30 @@ impl ViewClassRegistry {
...
+            view_class_identifier,
+            per_visualizer: visualizable_entities
+                .iter()
+                .filter_map(|(vis, ents)| {
+                    view_class
+                        .visualizer_system_ids
+                        .contains(vis)
+                        .then_some((*vis, ents.clone()))
+                })
+                .collect(),
```
grtlr: 💊

crates/viewer/re_viewport_blueprint/src/entity_add_info.rs
```diff
@@ -71,19 +70,13 @@ pub fn create_entity_add_info(
...
-        &view.space_origin,
-    );
+    let class = ctx
+        .view_class_registry()
+        .get_class_entry_or_log_error(view.class_identifier());
 
     tree.visit_children_recursively(|entity_path| {
         let can_add: CanAddToView =
-            if visualizable_entities.iter().any(|(_, entities)| entities.contains(entity_path)) {
+            if ctx.visualizable_entities_per_visualizer.iter().any(|(vis, entities)| class.visualizer_system_ids.contains(vis) && entities.contains(entity_path)) {
```
grtlr: :pill:

crates/viewer/re_context_menu/src/actions/add_entities_to_new_view.rs
Wumpf: agreed, will give this a pass!

crates/viewer/re_view_tensor/src/view_class.rs
Wumpf: yeah, good idea!

crates/viewer/re_view_spatial/tests/topology_errors.rs
Wumpf: clarifying comment

grtlr:
Needs to be rebased onto `main` to update the snapshots.

PR #11974, opened 2025-11-25
Native back and forth navigation

IsseW:
### Related

* Closes RR-2475

### What

Adds native back and forth navigation. Attempts to mirror almost exactly what web history does.

https://github.com/user-attachments/assets/99ee0add-3f18-4e03-8f86-39f668e20151


crates/viewer/re_viewer/src/app.rs
```diff
@@ -666,23 +663,48 @@ impl App {
...
+        };
+
+        self.state.history.update_current_url(url);
+    }
+
     /// Updates the web address on web. Noop on native.
     #[cfg(not(target_arch = "wasm32"))]
-    #[expect(clippy::unused_self)]
-    fn update_web_address_bar(&self, _store_hub: &StoreHub) {}
+    fn update_web_address_bar(&mut self, store_hub: &StoreHub) {
```
lucasmerlin: `update_web_address_bar` is a bit confusing if it also does something on native. Maybe just `update_history` and then rename the implementations to `update_web/native_history`?

crates/viewer/re_viewer/src/history.rs
```diff
@@ -1,305 +1,65 @@
...
-                .to_string()
+impl History {
+    pub fn update_current_url(&mut self, url: ViewerOpenUrl) {
+        // If just the fragment changed we don't push a new entry.
+        if let Some(current) = self.entries.get_mut(self.current_entry)
+            && current.clone().without_fragment() == url.clone().without_fragment()
         {
-            Self::default()
+            *current = url;
         } else {
```
lucasmerlin, line 23: I think this should be more explicit. Sometimes it might be useful to do a "replace" navigation even if more than the fragment changes. And sometimes you might want the fragment change to be part of the history. Maybe we need some kind of replace navigation event?

crates/viewer/re_viewer/src/history.rs
```diff
@@ -1,305 +1,65 @@
...
-pub struct HistoryEntry {
-    /// The URL to load/navigate to when using this entry.
-    ///
-    /// Note that the `url` parameter actually supports several urls, but
-    /// for navigation purposes we only support a single one.
-    url: String,
+/// History of url's the viewer has visited.
+#[derive(Default)]
+pub struct History {
+    entries: Vec<ViewerOpenUrl>,
```
lucasmerlin, line 12: What if the last item is a `IntraRecordingSelection` and we go back? Since it only holds an `Item`, I expect this to break in some way if we go back from a page that is not the recording that has that item.

crates/viewer/re_viewer/src/navigation.rs
```diff
@@ -1,19 +1,21 @@
 use re_viewer_context::DisplayMode;
+use vec1::smallvec_v1::SmallVec1;
 
 /// The navigation history of the viewer.
 ///
 /// This object should never be exposed to directly via contexts. Instead,
 /// we retrieve the display mode and pass that around.
 pub(crate) struct Navigation {
-    history: Vec<DisplayMode>,
+    stack: SmallVec1<[DisplayMode; 2]>,
```
lucasmerlin: This feels kind of duplicated with history.rs now... Maybe we can get rid of this and somehow construct the DisplayMode from the history url? Or we extend this to also hold the `ViewerOpenUrl`?

lucasmerlin (changes requested):
Works nicely!

But the back arrow doesn't work on the settings page. (I assume because the Settings page is not part of the OpenUrl enum?). 

Feel free to huddle me if you want to brainstorm on the History / Navigation thing

crates/viewer/re_viewer/src/history.rs
lucasmerlin: Ah, I guess this doesn't happen because those pages never make it to history.rs, due to the fragment filter? Still, feels kind of brittle.

crates/viewer/re_viewer/src/history.rs
lucasmerlin: After discussing this we decided to keep it as as for now, as it would likely require a large refactor.

crates/viewer/re_viewer/src/navigation.rs
IsseW: Updated it so navigation only has a current display mode now.

Can't really base it on the history url, as we use the current display mode to update history.

crates/viewer/re_ui/src/command.rs
```diff
@@ -378,6 +384,9 @@ impl UICommand {
             Self::NextRecording => smallvec![cmd_alt(Key::ArrowDown)],
             Self::PreviousRecording => smallvec![cmd_alt(Key::ArrowUp)],
 
+            Self::NavigateBack => smallvec![cmd_alt(Key::ArrowLeft)],
+            Self::NavigateForward => smallvec![cmd_alt(Key::ArrowRight)],
```
IsseW: Do we want key-binds for these?

crates/viewer/re_viewer/src/history.rs
IsseW: Yeah what's added to history is always something that's constructed from a display mode.

crates/viewer/re_ui/src/command.rs
lucasmerlin: Maybe `cmd [` / `cmd ]`? These go back and forth in chrome as well. Might just be annoying for keyboards that don't have [] keys. Not sure about arrow keys since those are so used by the timeline navigation.

crates/viewer/re_viewer/src/app.rs
```diff
@@ -655,23 +659,61 @@ impl App {
...
+                    url
+                })
+        else {
             return;
+        };
+
+        self.state.history.update_current_url(url);
+    }
+
+    /// Updates the web address on web. Noop on native.
```
lucasmerlin: The comment is wrong now

crates/viewer/re_test_context/src/lib.rs
```diff
@@ -746,6 +746,8 @@ impl TestContext {
                 | SystemCommand::AddReceiver { .. }
                 | SystemCommand::ResetViewer
                 | SystemCommand::ChangeDisplayMode(_)
+                | SystemCommand::Settings
+                | SystemCommand::ChunkStoreBrowser
```
lucasmerlin: Lets maybe call them `NavigateSettings` or `OpenSettings` for clarity

crates/viewer/re_viewer_context/src/display_mode.rs
```diff
@@ -5,7 +5,9 @@ use re_log_types::{StoreId, TableId};
 #[derive(Clone, PartialEq, Eq)]
 pub enum DisplayMode {
     /// The settings dialog for application-wide configuration.
-    Settings,
+    ///
+    /// The inner display mode is the one to return to when exiting this mode.
+    Settings(Box<DisplayMode>),
```
lucasmerlin, line 10: Does triggering a `UiCommand::NavigateBack` not work for `Settings` and `ChunkStoreBrowser`?

lucasmerlin (approved):
Awesome, works perfectly now!

crates/viewer/re_viewer_context/src/display_mode.rs
IsseW: It could work, on native now at least. But I think it's a bit weird since if you close it and go back you'd now be at an earlier point in history than expected I think.

Also to do that on web triggering a history back does feel very hacky.

crates/viewer/re_ui/src/command.rs
IsseW: Ah yeah! I've always used `cmd + arrows` in browsers as I don't have [] keys. But yeah changed it to those.

PR #11975, opened 2025-11-25
Analytics test CI 2

aedm:

PR #11976, opened 2025-11-25
Disable analytics in testing

aedm:
### What

Disables Posthog analytics events during testing, CI and in debug builds.

Tested on CI: https://github.com/rerun-io/rerun/pull/11975

### DIFF
--- a/crates/utils/re_analytics/src/native/pipeline.rs
+++ b/crates/utils/re_analytics/src/native/pipeline.rs
@@ -36,7 +36,16 @@ pub struct Pipeline {
 
 impl Pipeline {
     pub(crate) fn new(config: &Config, tick: Duration) -> Result<Option<Self>, PipelineError> {
+        if std::env::var("CI").is_ok() {
+            re_log::debug_once!("Analytics disabled on CI");
+            return Ok(None);
+        }
         if cfg!(feature = "testing") {
+            re_log::debug_once!("Analytics disabled in tests");
+            return Ok(None);
+        }
+        if cfg!(debug_assertions) {
+            re_log::debug_once!("Analytics disabled in debug builds");
             return Ok(None);
         }
 


crates/utils/re_analytics/src/native/pipeline.rs
```diff
@@ -36,6 +36,10 @@ pub struct Pipeline {
 
 impl Pipeline {
     pub(crate) fn new(config: &Config, tick: Duration) -> Result<Option<Self>, PipelineError> {
+        if cfg!(feature = "testing") {
+            return Ok(None);
+        }
```
emilk, line 50: This will handle `cargo test --all-features`, but not `cargo test`.

Maybe we should disable analytics in all debug builds?

And ALSO when `CI` env is set.

So something like


```suggestion
        if std::env::var("CI").is_ok() {
            re_log::debug_once!("Analytics disabled on CI");
            return Ok(None);
        }
        if cfg!(feature = "testing") {
            re_log::debug_once!("Analytics disabled in tests");
            return Ok(None);
        }
        if cfg!(debug_assertions) {
            re_log::debug_once!("Analytics disabled in debug builds");
            return Ok(None);
        }
```

aedm: Interestingly, the current version already decreases the number of analytics events to zero.

I made CI run in #11975 with two changes:
- I changed version to `alpha.62`, no other change: [284 posthog events](https://eu.posthog.com/project/1954/sql#q=select%20count(*)%20from%20events%20where%20properties.%24set.rerun_version%20like%20'%25alpha.62%25'%0A)
- I changed version to `alpha.61` with this PR merged: [0 posthog events](https://eu.posthog.com/project/1954/sql#q=select%20count(*)%20from%20events%20where%20properties.%24set.rerun_version%20like%20'%25alpha.61%25'%0A)

@emilk What do you think?

emilk: Our CI always uses `--all-features`. But what happens if you run `cargo test -p some-crate` (without `--all-features`) locally?

aedm: `cargo nextest run` locally:
- default: 52 events
- with this PR: 0 events

aedm: Anyway, I think your changes make sense nevertheless. I updated the PR.

PR #11977, opened 2025-11-25
Add `show_frame` option on `TransformAxes3D`

grtlr:
### Related

* Closes RR-2992.
* Related to RR-2997.

### What

This PR does two things:

1. It exposes a way to lookup a `FrameId` by `FrameIdHash` via the `TransformTreeContext`.
2. Builds a `show_frame` option for `TransformAxes3D` on top of that.

> [!IMPORTANT]
> The current version (8b4083dbf483812954bc2957437b18b75186d3b3) performs instancing of the labels, the initial version (94e47d7a92dd2bda873e3f194a04ef818b7efd7b) created a separate batch for each label. @Wumpf which of these two options do you prefer?

image

### DIFF
--- a/crates/viewer/re_viewer/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.archetypes.TransformAxes3D.snap
+++ b/crates/viewer/re_viewer/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.archetypes.TransformAxes3D.snap
@@ -3,3 +3,4 @@ source: crates/viewer/re_viewer/tests/all_component_fallbacks.rs
 expression: arch_display
 ---
 axis_length: [1.0]
+show_frame: [false]
--- a/crates/viewer/re_viewer/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.archetypes.TransformAxes3D.snap.new
+++ b/crates/viewer/re_viewer/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.archetypes.TransformAxes3D.snap.new
@@ -1,7 +1,0 @@
----
-source: crates/viewer/re_viewer/tests/all_component_fallbacks.rs
-assertion_line: 55
-expression: arch_display
----
-axis_length: [1.0]
-show_frame: [false]


crates/store/re_types/src/archetypes/transform_axes3d.rs
```diff
@@ -235,7 +273,8 @@ impl TransformAxes3D {
         self,
     ) -> SerializationResult<impl Iterator<Item = ::re_types_core::SerializedComponentColumn>> {
         let len_axis_length = self.axis_length.as_ref().map(|b| b.array.len());
-        let len = None.or(len_axis_length).unwrap_or(0);
+        let len_show_frame = self.show_frame.as_ref().map(|b| b.array.len());
+        let len = None.or(len_axis_length).or(len_show_frame).unwrap_or(0);
```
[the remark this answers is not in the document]
grtlr: This is codegen'ed.

crates/store/re_types/definitions/rerun/archetypes/transform_axes3d.fbs
```diff
@@ -15,4 +15,7 @@ table TransformAxes3D (
   /// The length is interpreted in the local coordinate system of the transform.
   /// If the transform is scaled, the axes will be scaled accordingly.
   axis_length: rerun.components.AxisLength ("attr.rerun.component_required", order: 1000);
+
+  /// Whether to show a text label with the corresponding frame.
```
Wumpf, line 19: should mention the default
Actually what _is_ the default? I guess thanks to globally registered fallbacks that's using some counting threshold? But not sure if that one actually is transferable x-archetypes 🤔. So probably just defaults to false?

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -73,8 +78,14 @@ impl ViewContextSystem for TransformTreeContext {
         let transform_forest =
             re_tf::TransformForest::new(ctx.recording(), &transform_cache, &ctx.current_query());
 
+        let frame_ids = transform_cache
```
Wumpf, line 81: any particular reason we're not just copying the frame registry as a whole? 🤷

crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
```diff
@@ -125,21 +133,60 @@ impl VisualizerSystem for TransformAxes3DVisualizer {
...
+                        },
+                        glam::Affine3A::IDENTITY,
+                    ));
+                } else {
+                    // It should not be possible to hit this path, but we keep the debug assert for good measure.
+                    re_log::warn_once!("Could not resolve frame id hash {frame_id_hash:?}");
+                    debug_assert!(
+                        false,
+                        "[DEBUG ASSERT] unable to resolve frame id hash {frame_id_hash:?}"
+                    );
```
Wumpf: let's use the new visualizer errors here! ✨

crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
```diff
@@ -125,21 +133,60 @@ impl VisualizerSystem for TransformAxes3DVisualizer {
...
+
+                if let Some(frame_id) = transforms.lookup_frame_id(frame_id_hash) {
+                    // Add label at the center of each transform instance if show_frame is enabled
+                    let num_instances = transforms_to_draw.len();
+                    let labels = vec![frame_id.0.clone().into(); num_instances];
+                    let annotation_infos = ResolvedAnnotationInfos::Same(
+                        num_instances,
+                        ResolvedAnnotationInfo::default(),
+                    );
+                    self.0.ui_labels.extend(process_labels_3d(
```
Wumpf: Best to use none of these `process_label` utilities here. Just looked at it and it doesn't have any value here to use it since youre not running into any of the clamping usecases it's supposed to handle. Just pushing a `UiLabel` likely also reads better!

Wumpf (changes requested):
nice! would like to make some adjustments though :)

crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
```diff
@@ -125,21 +133,60 @@ impl VisualizerSystem for TransformAxes3DVisualizer {
...
+            let show_frame: bool = results
+                .get_mono_with_fallback::<ShowLabels>(show_frame_identifier)
+                .into();
+
+            if show_frame {
+                let frame_id_hash =
+                    transforms.transform_frame_id_for(data_result.entity_path.hash());
+
+                if let Some(frame_id) = transforms.lookup_frame_id(frame_id_hash) {
+                    // Add label at the center of each transform instance if show_frame is enabled
```
MichaelGrupp: I think it should be called `show_label`. Reading `show_frame` without such documentation (which anyway explains it as a text _label_), I would at first think that it's related to the coordinate frame markers or so.

MichaelGrupp (changes requested):

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
grtlr: I thought about that but if we ever decide to add more stuff to the frame registry it might get cloned too. So I didn't want to couple them as closely. Maybe a comment would be enough though?

crates/store/re_types/definitions/rerun/archetypes/transform_axes3d.fbs
grtlr: Hmm, I thought `ShowLabels` would always default to `false` but let me check.

grtlr: Added a fallback provider.

Wumpf: > Hmm, I thought ShowLabels would always default to false but let me check.

probably does 😄

PR #11981, opened 2025-11-25
Fix opening non-recording URLs on startup

Wumpf:
Opening any URL that didn't point directly to a recording (tabels, datset entries etc.) selected the entry but didn't show the actual contents.

The fix for this is rather brittle. I explored making it less brittle and while the current situation is extra peculiar I concluded that this is hard to get perfect especially right now. Which is extra-why I added an integration test for it and made sure that it fails prior to the fix! :)

### DIFF
--- a/crates/viewer/re_viewer/src/viewer_test_utils/mod.rs
+++ b/crates/viewer/re_viewer/src/viewer_test_utils/mod.rs
@@ -15,6 +15,7 @@ pub struct HarnessOptions {
     pub window_size: Option<egui::Vec2>,
     pub max_steps: Option<u64>,
     pub step_dt: Option<f32>,
+    pub startup_url: Option<String>,
 }
 
 /// Convenience function for creating a kittest harness of the viewer App.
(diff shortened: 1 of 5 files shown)


crates/viewer/re_viewer_context/src/store_hub.rs
```diff
@@ -188,7 +188,13 @@ impl StoreHub {
...
             persistence,
             active_recording_or_table: None,
-            active_application_id: None,
+
+            // No active app is only ever transitional and we react by it to go back to the
+            // welcome/start screen.
+            // During application startup we may decide early to switch to a different screen,
+            // so make sure we start out with the welcome screen app already set, so we won't
+            // don't override this in the first frame.
+            active_application_id: Some(Self::welcome_screen_app_id()),
```
IsseW, line 197: Wonder if this even needs to be an option, but would require some further changes so I think this is good for this PR.

IsseW (approved):
That test was a lot simpler than I thought it'd be!

Wumpf:
> That test was a lot simpler than I thought it'd be!

you don't want to know how long it took me though to get it right 😆

PR #11983, opened 2025-11-26
Make `transform3d_hierarchy_frames` cross language

IsseW:
### Related

- Closes RR-2880

### What

Adds rust & cpp versions of `transform3d_hierarchy_frames`. Also does a TODO to change points into spheres.

image

* [x] pass full ci check ✅


Wumpf:
screenshot plz of what it looks like with spheres :)

docs/snippets/all/archetypes/transform3d_hierarchy_frames.cpp
```diff
@@ -0,0 +1,101 @@
...
+    rec.set_time_duration_secs("sim_time", 0.0);
+
+    // Planetary motion is typically in the XY plane.
+    rec.log_static("/", rerun::ViewCoordinates::RIGHT_HAND_Z_UP);
+
+    // Setup spheres, all are in the center of their own space:
+    rec.log(
+        "sun",
+        rerun::Ellipsoids3D::from_centers_and_half_sizes({{0.0f, 0.0f, 0.0f}}, {{1.0f, 1.0f, 1.0f}})
+            .with_colors({rerun::Color(255, 200, 10)})
```
Wumpf: I think this is overloaded such that

```suggestion
            .with_colors(rerun::Color(255, 200, 10))
```
should also work. But need to consult compiler

IsseW: Will check

IsseW: Yep works, but not for the arguments of `from_centers_and_half_sizes`

docs/snippets/snippets.toml
```diff
@@ -211,11 +211,6 @@ backwards_check = [
   "cpp", # Blueprint API doesn't exist for C++/Rust
   "rust", # Blueprint API doesn't exist for C++/Rust
 ]
-"archetypes/transform3d_hierarchy_frames" = [
```
Wumpf, line 214: spoiler: You'll likely still need a compare exclusion to pass tests because of trigonometry being inconsistent across languages

IsseW: Ahh that's what you meant by that

IsseW:
@rerun-bot full-check

docs/snippets/snippets.toml
IsseW: It doesn't seem like we already did that for `transform3d_hierarchy` though. So maybe not?

Wumpf: oh if it does work that's ofc even better \o/
very surprised by that

PR #11984, opened 2025-11-26
Add static/temporal chunk query filtering to OSS server

oxkitsune:
### Related

* Part of RR-2613

### What

This implements query filtering for temporal/static chunks in the OSS server.


crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -1009,10 +1009,10 @@ impl RerunCloudService for RerunCloudHandler {
...
             //TODO(RR-2613): we must do a much better job at handling these
-            chunk_ids: _,
+            chunk_ids: requested_chunk_ids,
             fuzzy_descriptors: _,
-            exclude_static_data: _,
-            exclude_temporal_data: _,
+            exclude_static_data,
+            exclude_temporal_data,
             scan_parameters: _,
             query: _,
```
emilk: These are still ignored - can we please add a logged warning about that while we're at it? 🙏

Silently ignoring stuff is what wasted an hour of your life today; let's stop doing that

abey79: I don't think it's a good idea. We _will_ (*) get lots of warning for something that is:
1) already on the todo list
2) best effort anyways

Debug logging _might_ make sense (though I'm not convinced), but warning on a TODO seems unproductively noisy.

(*) This will happen when using  dataframe queries.

oxkitsune: I think implementing the query/fuzzy scanning is pretty straight forward, so I can do that in this PR as well. For the scan parameters I'm a bit less sure and would have to look at it.

emilk: One line of `debug` logging of "The server does not yet support time queries" is the right level imho (using `debug_once`, and only when the query is set). But fixing it would ofc be better, as my larger-than-ram work needs it anyways

emilk: > I think implementing the query/fuzzy scanning is pretty straight forward, so I can do that in this PR as well. For the scan parameters I'm a bit less sure and would have to look at it.

For larger-than-ram, it is the time range query that is the clincher

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -1063,6 +1072,66 @@ impl RerunCloudService for RerunCloudHandler {
...
+                        // For range queries, filter chunks whose time range intersects with the query range
+                        if let Some(ref range) = query.range {
+                            let timeline_name = range.index.clone().into();
+                            let dominated = match chunk.timelines().get(&timeline_name) {
+                                Some(timeline_col) => {
+                                    timeline_col.time_range().intersects(range.index_range)
+                                }
+                                None => chunk.is_static(),
+                            };
+                            if !dominated {
```
emilk: "dominated" is an odd choice of word. Maybe I'm missing something

```suggestion
                            let include = match chunk.timelines().get(&timeline_name) {
                                Some(timeline_col) => {
                                    timeline_col.time_range().intersects(range.index_range)
                                }
                                None => chunk.is_static(),
                            };
                            if !include {
```

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -1063,6 +1072,66 @@ impl RerunCloudService for RerunCloudHandler {
...
+                                continue;
+                            }
+                        }
+
+                        // For latest_at queries, filter chunks whose time range contains data at or before the query timestamp
```
emilk:
```suggestion
                        // For latest_at queries, filter chunks whose time range contains data at or before the query timestamp.
                        // This may include too many chunks, but we are ok with that for now.
```

crates/store/re_redap_tests/src/tests/snapshots/re_redap_tests__tests__query_dataset__simple_dataset_query_range_15_25_data.snap
```diff
@@ -0,0 +1,31 @@
...
+│ 00000000000000030000000000000002 ┆ my_partition_id3   ┆ base                  ┆ /my/entity        ┆ true            │
+├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+│ 00000000000000030000000000000003 ┆ my_partition_id3   ┆ base                  ┆ /another/one      ┆ false           │
+├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+│ 00000000000000030000000000000004 ┆ my_partition_id3   ┆ base                  ┆ /another/one      ┆ true            │
+├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+│ 00000000000000030000000000000005 ┆ my_partition_id3   ┆ base                  ┆ /yet/another/one  ┆ false           │
+├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+│ 00000000000000030000000000000006 ┆ my_partition_id3   ┆ base                  ┆ /yet/another/one  ┆ true            │
```
emilk: Does this actually exclude any chunks? Hard to tell without a corresponding "everything" query

emilk: I added a test with no filtering at all, and it returns the same thing - so this seems broken

crates/store/re_redap_tests/src/tests/snapshots/re_redap_tests__tests__query_dataset__simple_dataset_query_latest_at_25_data.snap
```diff
@@ -0,0 +1,31 @@
...
+│ 00000000000000030000000000000002 ┆ my_partition_id3   ┆ base                  ┆ /my/entity        ┆ true            │
+├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+│ 00000000000000030000000000000003 ┆ my_partition_id3   ┆ base                  ┆ /another/one      ┆ false           │
+├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+│ 00000000000000030000000000000004 ┆ my_partition_id3   ┆ base                  ┆ /another/one      ┆ true            │
+├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+│ 00000000000000030000000000000005 ┆ my_partition_id3   ┆ base                  ┆ /yet/another/one  ┆ false           │
+├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+│ 00000000000000030000000000000006 ┆ my_partition_id3   ┆ base                  ┆ /yet/another/one  ┆ true            │
```
emilk: This also returns all partitions, with no visible filtering

oxkitsune:
There seem to be some deeper issues with the API endpoint, because the tests also do not pass on redap.
I have removed the fuzzy matching/time querying and plan to add those in a separate PR instead, so we can get the temporal/static filtering merged.

crates/store/re_redap_tests/src/tests/snapshots/re_redap_tests__tests__query_dataset__simple_dataset_query_range_15_25_data.snap
```diff
@@ -0,0 +1,31 @@
+---
+source: crates/store/re_redap_tests/src/tests/query_dataset.rs
```
emilk: These extra .snap files should be removed now

PR #11988, opened 2025-11-26
Add support to explicit child/parent transform frames for `Pinhole`

Wumpf:
### Related

* Fixes RR-2680

### What

Add `child_frame` & `parent_frame` to `Pinhole`, getting it in line with `Transform3D`'s components of the same name.

Additionally, introduces atomic semantics for _some_ of `Pinhole`'s components. This very rarely makes a difference, see migration guide update.

Surprisingly straight forward change given all the other things we landed earlier! But it triggered some rearrangement again in `TransformResolutionCache` and the testing around it is rather involved (and very necessary :). Existing tests helped a ton too!)

image

### DIFF
--- a/crates/store/re_types/definitions/rerun/archetypes/pinhole.fbs
+++ b/crates/store/re_types/definitions/rerun/archetypes/pinhole.fbs
@@ -5,6 +5,10 @@ namespace rerun.archetypes;
 ///
 /// If [archetypes.Transform3D] is logged for the same child/parent relationship (e.g. for the camera extrinsics), it takes precedence over [archetypes.Pinhole].
 ///
+/// If you use explicit transform frames via the `child_frame` and `parent_frame` fields, you don't have to use [archetypes.CoordinateFrame]
+/// as it is the case with other visualizations: for any entity with an [archetypes.Pinhole] the viewer will always visualize it
+/// directly without needing a [archetypes.CoordinateFrame] to refer to the pinhole's child/parent frame.
+///
 /// \example archetypes/pinhole_simple title="Simple pinhole camera" image="https://static.rerun.io/pinhole_simple/9af9441a94bcd9fd54e1fea44fb0c59ff381a7f2/1200w.png"
 /// \example archetypes/pinhole_perspective title="Perspective pinhole camera" image="https://static.rerun.io/pinhole_perspective/317e2de6d212b238dcdad5b67037e9e2a2afafa0/1200w.png"
 table Pinhole (
--- a/crates/store/re_types/src/archetypes/pinhole.rs
+++ b/crates/store/re_types/src/archetypes/pinhole.rs
@@ -25,6 +25,10 @@ use ::re_types_core::{DeserializationError, DeserializationResult};
 ///
 /// If [`archetypes::Transform3D`][crate::archetypes::Transform3D] is logged for the same child/parent relationship (e.g. for the camera extrinsics), it takes precedence over [`archetypes::Pinhole`][crate::archetypes::Pinhole].
 ///
+/// If you use explicit transform frames via the `child_frame` and `parent_frame` fields, you don't have to use [`archetypes::CoordinateFrame`][crate::archetypes::CoordinateFrame]
+/// as it is the case with other visualizations: for any entity with an [`archetypes::Pinhole`][crate::archetypes::Pinhole] the viewer will always visualize it
+/// directly without needing a [`archetypes::CoordinateFrame`][crate::archetypes::CoordinateFrame] to refer to the pinhole's child/parent frame.
+///
 /// ⚠️ **This type is _unstable_ and may change significantly in a way that the data won't be backwards compatible.**
 ///
 /// ## Examples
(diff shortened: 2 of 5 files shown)


crates/viewer/re_viewer/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.archetypes.Pinhole.snap
```diff
@@ -1,10 +1,12 @@
 ---
-source: crates/viewer/re_component_fallbacks/tests/all_component_fallbacks.rs
+source: crates/viewer/re_viewer/tests/all_component_fallbacks.rs
 expression: arch_display
 ---
 image_from_camera: [[100.0, 0.0, 0.0, 0.0, 100.0, 0.0, 100.0, 100.0, 1.0]]
 resolution: [[0.0, 0.0]]
 camera_xyz: [[3, 2, 5]]
+child_frame: [tf#/]
+parent_frame: [tf#/]
```
grtlr: It's a bit unfortunate that they are the same in this example.

Could we change the `target_entity_path` to `/my_entity` in `test_all_component_fallbacks`?

crates/viewer/re_view_spatial/src/visualizers/cameras.rs
```diff
@@ -56,35 +56,56 @@ struct CameraComponentDataWithFallbacks {
...
+
+        // We're currently NOT using `CoordinateFrame` component for this visualization but instead `Pinhole::child_frame`.
+        // Otherwise, you'd need to log a redundant `CoordinateFrame` to see the camera frustum which can be unintuitive.
+        //
+        // Note that `child_frame` defaults to the entity's implicit frame, so if no frames are set it doesn't make a difference.
+        //
+        // In theory `CoordinateFrame::frame_id` and `Pinhole::child_frame` could disagree making it unclear what to show.
+        // Sticking with the semantics of `CoordinateFrame::frame_id`, we should give it precedence,
+        // but this implies ignoring `CoordinateFrame::frame_id`'s fallback in all other cases which is arguably
+        // even more confusing. So instead, we rely _solely_ on `Pinhole::child_frame` for now.
```
grtlr, line 77: Should we make this more prominent in user-facing docs too? I don't think this will change for `0.28`?

crates/viewer/re_view_spatial/src/visualizers/cameras.rs
```diff
@@ -239,20 +261,23 @@ impl VisualizerSystem for CamerasVisualizer {
                 query_shadowed_components,
             );
 
-            let Some(pinhole_projection) = query_results
+            // `image_from_camera` _is_ the required component, but we don't process it further since we rely on the
+            // pinhole information from the transform tree instead, which already has this and other properties queried.
```
grtlr, line 265: We should probably create an issue and leave a TODO somewhere to make `atomic` queries / groups a real thing to track these assumptions throughout the code base.

crates/viewer/re_view_spatial/tests/project_2d_and_3d.rs
```diff
@@ -3,76 +3,101 @@
...
+        // Look at the middle box.
+        glam::Mat3::look_at_rh(eye_position, glam::vec3(0.0, 1.0, 0.0), glam::Vec3::Z),
+    )
+    .with_translation(eye_position);
+    let camera_intrinsics =
+        archetypes::Pinhole::from_focal_length_and_resolution([2., 2.], [3., 2.])
+            .with_image_plane_distance(1.0);
+    let camera_image = archetypes::Image::from_color_model_and_tensor(
+        re_types::datatypes::ColorModel::RGB,
+        Array::<u8, _>::zeros((2, 3, 3).f()),
```
grtlr: _something something checkerboard here_

crates/viewer/re_view_spatial/tests/snapshots/3d_in_2d_with_explicit_frames.png
grtlr, line 1: Not sure why the red cube is showing here (at least when looking at the picture above). Even if we don't clip/cull it away, it still looks like it's outside of the purple spheres.

grtlr: Ah, this image explains it better:

Image

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -685,28 +732,64 @@ impl TransformResolutionCache {
...
                             *timeline,
-                            &mut self.static_timeline,
-                            &per_timeline.non_recursive_clears,
-                            &per_timeline.recursive_clears,
+                            static_timeline,
                         )
-                    });
-
-                frame_transforms.insert_invalidated_transform_event(aspects, time);
+                        .invalidate_transform_at(time);
```
grtlr, line 758: This new API is so much nicer!

crates/viewer/re_view_spatial/tests/snapshots/2d_in_3d_with_explicit_frames.png
Wumpf, line 1: @grtlr we should probably think about making the arrow show up automatically here. I think we have to opt-in visualizability (!) for presence of any pinhole component

grtlr: Yeah, why not!

crates/viewer/re_viewer/tests/snapshots/all_component_fallbacks__arch_fallback_rerun.archetypes.Pinhole.snap
Wumpf: good idea. Will do! curious about the changes

crates/viewer/re_view_spatial/src/visualizers/cameras.rs
Wumpf: yeah don't intend to do anything about it in the next few months even.
I'll try to write something!

Issue #11467, opened 2025-10-08
Update our Rust dependencies


emilk:
We have a lot of old dependencies which needs to be bumped in `Cargo.toml`

GeorgeLifinRell:
I would like to work on this, can you please list out the deliverables?

emilk:
Thanks!

I'd like for us to not use old versions of crates. Some crates are very hard to update (`tonic`, `datafusion`, `lance`, `arrow`), bug some should be fairly trivial (like`infer` and `indicatif`).

You can list which crates needs updating with [`cargo-edit`](https://crates.io/crates/cargo-edit):

```
❯ cargo upgrade --compatible ignore --incompatible allow --dry-run
    Checking virtual workspace's dependencies
name                           old req compatible latest new req
====                           ======= ========== ====== =======
arrow                          55.2    55.2.0     56.2.0 56.2   
bincode                        1.3     1.3.3      2.0.1  2.0    
cargo_metadata                 0.18    0.18.1     0.23.0 0.23   
cargo-run-wasm                 0.3.2   0.3.2      0.4.0  0.4.0  
colored                        2.2     2.2.0      3.0.0  3.0    
convert_case                   0.6     0.6.0      0.8.0  0.8    
criterion                      0.5     0.5.1      0.7.0  0.7    
datafusion                     49.0.2  49.0.2     50.1.0 50.1.0 
datafusion-ffi                 49.0.2  49.0.2     50.1.0 50.1.0 
indicatif                      0.17.11 0.17.11    0.18.0 0.18.0 
infer                          0.16.0  0.16.0     0.19.0 0.19.0 
jsonwebtoken                   9.3     9.3.1      10.0.0 10.0   
lance                          0.36.0  0.36.0     0.38.2 0.38.2 
numpy                          0.24.0  0.24.0     0.26.0 0.26.0 
opentelemetry                  0.30.0  0.30.0     0.31.0 0.31.0 
opentelemetry-appender-tracing 0.30.1  0.30.1     0.31.1 0.31.1 
opentelemetry-otlp             0.30.0  0.30.0     0.31.0 0.31.0 
opentelemetry_sdk              0.30.0  0.30.0     0.31.0 0.31.0 
ordered-float                  4.6.0   4.6.0      5.1.0  5.1.0  
parquet                        55.2    55.2.0     56.2.0 56.2   
prost                          0.13.5  0.13.5     0.14.1 0.14.1 
[...]

PR #11990, opened 2025-11-26
Updating Rust deps to their latest version

kratos2377:
Part of #11467 

Updated some of the core dependencies to their latest version. 

I have tested the change by running rerun as both desktop and web application. The examples are working fine.

@Wumpf @emilk Please take a look at the PR when you have time and let me know if any other validation is also required.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -346,7 +346,7 @@ similar-asserts = "1.7.0"
 slotmap = { version = "1.0.7", features = ["serde"] }
 smallvec = { version = "1.15", features = ["const_generics", "union"] }
 static_assertions = "1.1"
-stl_io = "0.10.0"
+stl_io = "0.9.0"
 strum = { version = "0.26.3", features = ["derive"] } # need to update re_rav1d first
 strum_macros = "0.26.4" # need to update re_rav1d first
 sublime_fuzzy = "0.7.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,9 +20,9 @@ dependencies = [
 
 [[package]]
 name = "ab_glyph_rasterizer"
-version = "0.1.10"
+version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "366ffbaa4442f4684d91e2cd7c5ea7c4ed8add41959a31447066e279e432b618"
+checksum = "b2187590a23ab1e3df8681afdf0987c48504d80291f002fcdb651f0ef5e25169"
 
 [[package]]
 name = "abi_stable"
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -260,13 +260,13 @@ image = { version = "0.25.6", default-features = false }
...
 indicatif = "0.18.0" # Progress bar
-infer = "0.16.0" # infer MIME type by checking the magic number signaturefer MIME type by checking the magic number signature
+infer = "0.19.0" # infer MIME type by checking the magic number signaturefer MIME type by checking the magic number signature
 insta = "1.43"
 itertools = "0.14.0"
 jiff = { version = "0.2.15", features = ["js"] }
 js-sys = "0.3.77"
 jsonwebtoken = { version = "9.3", default-features = false }
-lance = { version = "0.38.2", default-features = false } # When you update this, also update the list of features enabled for `datafusion` (~50 lines up)
+lance = { version = "0.39.0", default-features = false } # When you update this, also update the list of features enabled for `datafusion` (~50 lines up)
```
emilk, line 269: did you remember to follow this instruction?

kratos2377: @emilk Added cargo.lock and there is no difference b/w the datafusion feature set in v0.39.0 and v0.38.0
v0.39.0 -> https://github.com/lance-format/lance/blob/ab5fdb362eda381bab5cb559cee94de8b3d63c6d/Cargo.toml#L106

v0.38.0 -> https://github.com/lance-format/lance/blob/v0.38.0/Cargo.toml#L99-L107

emilk:
Looks like a lot of crates no longer compile. Please make sure ` pixi run rs-check ` works!

Cargo.toml
emilk, line 269: @timsaucer is there any reason not to update lance right now?

kratos2377:
@emilk I have reverted a lot of deps back to their previous version or to a different version because there were lot of conflicts in crates. the command `pixi run rs-check` is still giving some errors though. Just wanted to confirm if the command nextest is valid or not. This is the recent result after my changes

Screenshot from 2025-11-27 22-13-11

kratos2377:
@emilk All the checks seems to be passing now in CI/CD , is there anything else you want me to validate ?

Cargo.toml
timsaucer: The Lance datafusion update isn't ready yet, so no surprise on these updates to the lock.

timsaucer: I don't see any blocker here, but we would want to update our platform repo in lock step.

emilk:
Sorry - I forgot about this PR and then now we have too many conflicts :(

PR #11992, opened 2025-11-26
Fix timeseries time view min fallback

IsseW:
### Related

* Fixes some stuff I missed in https://github.com/rerun-io/rerun/pull/11938

### What

Fixes the min fallback for the time axis min, which had issues if you changed the min visible time range.


crates/viewer/re_view_time_series/tests/snapshots/blueprint_overrides_and_defaults_with_time_series_absolute.png
Wumpf, line 1: this and the other off-by-one chnages in snapshots I don't quite understand to be honest. was `line.min_time` snapping incorrectly?

Both these

* crates/viewer/re_view_time_series/tests/snapshots/visible_time_range_absolute_until_end_view_data.png
* crates/viewer/re_view_time_series/tests/snapshots/visible_time_range_absolute_view_data.png

are as expected after checking the code again


Seems not too concerning, but would be nice to understand why

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -462,7 +463,7 @@ impl ViewClass for TimeSeriesView {
         // Get the minimum time/X value for the entire plot…
         let min_time = all_plot_series
             .iter()
-            .map(|line| line.min_time)
+            .filter_map(|line| line.points.first().map(|(t, _)| *t))
```
Wumpf: so when is `min_time` the correct thing to use when it's apparently not the actual minimum time of those points? Haven't checked it's documentation, maybe that's all clear there?

IsseW: Taking a closer look at this was causing the off by one seconds as well! This `min_time` should be used for the `time_offset`

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -620,18 +626,6 @@ impl ViewClass for TimeSeriesView {
             .map(|line| line.aggregator)
             .unwrap_or_default();
 
-        // …then use that as an offset to avoid nasty precision issues with
-        // large times (nanos since epoch does not fit into a f64).
```
IsseW, line 624: this was duplicated code…

PR #11996, opened 2025-11-26
Introduce Python wrapper for `Entry`, `DatasetEntry`, and `TableEntry`

abey79:
### Related

* part of RR-2901
* fixes RR-2938

### What

This PR introduces Python wrapper for `Entry`, `DatasetEntry`, and `TableEntry`.

In more details:
- No user-facing API changes except:
  - RR-2938 is fixed
  - `CatalogClient.all_entries()` is now implemented in python, and is basically just `datasets()` + `tables()`. That means it doesn't return blueprint dataset (if it at all did before). This aligns with the intention to hide as much as possible the internal blueprint dataset machinery.
  - The `__repr__` of `Entry` is now more explicit with the `EntryKind` (see snapshot diff). This comes from it now being implemented in python. I think it's more in line with staying close too "runnable code" for repr. 
- The `DatasetEntry` <- `Entry` -> `TableEntry` hierarchy is now in pure python.
- The previous Rust-based hierarchy is now flattened: `PyEntry` is removed and it's functionality is integrated to now-named `PyDatasetEntryInternal` and `PyTableEntryInternal`.
	- Not having a python hierarchy on rust side simplifies/removes a whole lot of boilerplate.
- Because it's not possible to expose a trait (e.g. `trait PyEntry`) across the FFI using pyo3, the following is done to mitigate code duplication:
  - A new `PyEntryDetailsInternal` object is introduced, which is used to expose the basic entry properties in `Entry`
  - An helper function for updating entries.
  - At the end of the day, there remain a rust-side `Entry` informal protocol implementation that is duplicated, but imo worth the overall simplifications

### DIFF
--- a/rerun_py/ARCHITECTURE.md
+++ b/rerun_py/ARCHITECTURE.md
@@ -174,8 +174,8 @@ TODO(ab):
 ## Internal/wrapper pattern
 
 Experience shows that the pattern of using a pure-Python wrapper around pyo3-based internal object is successful. It allows:
-- keeping the "accept anything" magic on the Python side
-- having simple, canonically typed methods for rust-based objects (making it more likely to benefit from pyo3 magic type conversions) 
+- keeping the "accept anything" magic on the Python side;
+- having simple, canonically typed methods for rust-based objects (making it more likely to benefit from pyo3 magic type conversions).
 
 See `rerun.catalog.CatalogClient` for an example of this pattern.
 


rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1272,251 +1272,70 @@ class EntryKind:
...
         """int(self)"""  # noqa: D400
 
-class Entry:
-    """An entry in the catalog."""
-
-    @property
-    def id(self) -> EntryId:
-        """The entry's id."""
-
+class EntryDetailsInternal:
```
ntjohnson1, line 1279: NIT: Not required in this PR but since this PR is pushing forward a pattern we could clean up a bit.

Right now we re-export all our rerun_bindings at the top level
https://github.com/rerun-io/rerun/blob/b5b313dc94ff0cf65e2b770c8a06810250615ddf/rerun_py/rerun_bindings/__init__.py#L3

I think all these `Internal` classes should be underscore prefixed to hide them then I think [this](https://github.com/rerun-io/rerun/blob/b5b313dc94ff0cf65e2b770c8a06810250615ddf/rerun_py/rerun_bindings/__init__.py#L6) is an antipattern (that I participated in) and instead the usage sites should import as

`from rerun_bindings.rerun_bindings import _EntryDetailsInternal`

rerun_py/rerun_sdk/rerun/catalog/__init__.py
```diff
@@ -20,4 +20,10 @@
     VectorDistanceMetricLike as VectorDistanceMetricLike,
 )
 
+# isort: off
+# Import order matters to avoid circular imports: Entry must come before DatasetEntry and TableEntry
```
ntjohnson1: Code smell. Can we not do a single entry file or something with all the variants? Our python project layout desperately needs love but once we start manually enforcing import orders things generally start falling apart. 😨 

Highlights our level of abstraction for things might be off.

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
```diff
@@ -83,41 +82,57 @@ def __init__(self, address: str, token: str | None = None) -> None:
...
+    def all_entries(self) -> list[DatasetEntry | TableEntry]:
         """Returns a list of all entries in the catalog."""
-        return self._raw_client.all_entries()
+
+        return self.dataset_entries() + self.table_entries()
 
     def dataset_entries(self) -> list[DatasetEntry]:
         """Returns a list of all dataset entries in the catalog."""
-        return self._raw_client.dataset_entries()
+        from . import DatasetEntry
```
ntjohnson1, line 113: Deferred import to avoid circular imports?

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
```diff
@@ -83,41 +82,57 @@ def __init__(self, address: str, token: str | None = None) -> None:
...
+        return [TableEntry(internal) for internal in self._internal.table_entries()]
 
     # ---
 
     def entry_names(self) -> list[str]:
         """Returns a list of all entry names in the catalog."""
-        return self._raw_client.entry_names()
+        return self._internal.entry_names()
 
     def dataset_names(self) -> list[str]:
```
ntjohnson1, line 129: No change needed here but its a little weird we have `all_entries`, `<dataset/table>_entries`, `<entry/dataset/table>_names`. I say weird because I didn't realize we had all these flavors until now.

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
```diff
@@ -141,13 +156,15 @@ def tables(self) -> datafusion.DataFrame:
 
     def get_dataset_entry(self, *, id: EntryId | str | None = None, name: str | None = None) -> DatasetEntry:
```
ntjohnson1, line 157: Another not this PR note. I wonder if it would be nicer/more pythonic for this to return `DatasetEntry | None` more of a don't look before you leap approach and if the dataset doesn't exist it returns none instead of raising an exception.

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
```diff
@@ -334,28 +358,31 @@ def write_python_objects_to_table(self, table_name: str, insert_mode: TableInser
...
 
     def _resolve_name_or_id(self, id: EntryId | str | None = None, name: str | None = None) -> EntryId:
         """Helper method to resolve either ID or name. Returns the id or throw an error."""
 
-        # TODO(ab): this screams for a `match` statement in Python 3.10+
-        if id is not None and name is not None:
-            raise ValueError("Only one of 'id' or 'name' must be provided.")
-        elif id is not None:
-            if isinstance(id, EntryId):
+        match id, name:
```
ntjohnson1, line 374: ❤️

rerun_py/rerun_sdk/rerun/catalog/_dataset_entry.py
```diff
@@ -0,0 +1,381 @@
...
+        Returns
+        -------
+        partition_id: str
+            The partition ID of the registered RRD.
+
+        """
+
+        return self._internal.register(recording_uri, recording_layer=recording_layer, timeout_secs=timeout_secs)
+
+    def register_batch(self, recording_uris: list[str], *, recording_layers: list[str] | None = None) -> Tasks:
```
ntjohnson1: NIT: Having the default none instead of the awful mutable empty list is a great reason to have the pure python wrapper. I assume we can also make our apis broader. At a minimum we should be able to support `Iterable[str]` instead of just `list[str]` no change required here necessarily but having a set of python side converters to make our input api as flexible as possible is a strong justification for adding another layer of indirection.

rerun_py/rerun_sdk/rerun/catalog/_dataset_entry.py
```diff
@@ -0,0 +1,381 @@
...
+    def download_partition(self, partition_id: str) -> Recording:
+        """Download a partition from the dataset."""
+
+        return self._internal.download_partition(partition_id)
+
+    def dataframe_query_view(
+        self,
+        *,
+        index: str | None,
+        contents: Any,
```
ntjohnson1: 😢 but I see we had these anys before this PR

rerun_py/rerun_sdk/rerun/catalog/_dataset_entry.py
ntjohnson1, line 1: I don't see a __str__ implementation for the python wrapper classes.

Can you run something like [this](https://github.com/rerun-io/rerun/pull/11928) printout to make sure we didn't just lose our pretty print for all of these? Maybe we should have a pytest for this (but I proeviously added the lint for the rust wrapper). I'm not sure how to add this with the new indiretion except maybe we just need `__str__` for all our python classes as well.

ntjohnson1: Can you also print out `client.all_entries` you mention making the repr more honest. However when we print a list of objects it seems to print the repr and I expect since we pass lists of entries they'll get printed often.

rerun_py/rerun_bindings/rerun_bindings.pyi
abey79: I actually had this thought, but figured that, from the perspective of `rerun_bindings`, `*Internal` object are _not_ private. They are the public API of the private package. If anything, we should instead name the internal package `_rerun_bindings` (or really, `_rerun_internal`, to be consistent with the "internal" wording).

Would you agree with this perspective? If so, I'll open a ticket to that effect.

rerun_py/rerun_sdk/rerun/catalog/__init__.py
abey79: On this being a code smell and level of abstraction comment, I don't entirely agree. The fact that there is an `Entry` hierarchy on python side is meant as a public API that conveys information on the underlying concept, not an implementation abstraction. As a matter of fact, the actual implementation no longer has this hierarchy. The circular import issues are a consequence of the desire to expose such an API.

On this lint exemption being painful: I agree. I think it's fine to merge everything in a single `_entry.py` file.

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
abey79: yes

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
abey79: These api will undergo a very significant brush-up (see `api_sandbox/rerun_draft/catalog.py`). I'm trying to minimise API changes here.

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
abey79: I dont have an opinion on this matter tbh. If you feel strongly about it, my suggestion would be make this change in the api_sandbox (along with a small demo snippet in `test_draft`) and open a PR for opinion-wielders to review.

rerun_py/rerun_sdk/rerun/catalog/_dataset_entry.py
abey79: Yeah, that's definitely the spirit. Have the rust-based method have a super straightforward, canonically typed, no magic API; then have all the python magics handled on python side. I'm not attempting to do any of this here, but anything that survives the api changes should evolve towards that goal. It really works wonder:
- simpler rust code which benefits more from the pyo3 magic conversions
- simpler stubs that are easier to write with proper, narrow type hints
- in turn, mypy is helpful with the python side "accept anything" glue

rerun_py/rerun_sdk/rerun/catalog/_dataset_entry.py
abey79: pinky finger will give it love when I rewrite that particular API

rerun_py/rerun_sdk/rerun/catalog/__init__.py
ntjohnson1: Abstraction not in the sense of the class hierarchy but rather in how the files and submodules are organized. I agree on the inheritance relationship. If there is a strict ordering in importing then maybe those files should be in their own submodule or in the same file.

Sounds like we're on the same page for merging the files as a reasonable short term solution to minimize the lint pain.

rerun_py/rerun_bindings/rerun_bindings.pyi
ntjohnson1: Hmm. Mostly just want something to highlight "rely on this API surface at your own risk". I'm ok with `_rerun_<something>`. If we end up splitting our python package into smaller more focussed packages we'll probably have a `_rerun_<something>` uniquely specified for each rather than the current single binary blob to rule them all.

rerun_py/rerun_sdk/rerun/catalog/_dataset_entry.py
abey79:
```
CatalogClient(rerun://sandbox.redap.rerun.io:443)
[Entry(EntryKind.TABLE, '__entries']
Entry(EntryKind.DATASET, 'droid:raw'
DataFusionTable(name='droid:raw_partition_table')
[IndexingResult(index='/camera/wrist/embedding:embeddings' on 'real_time': VectorIvfPq { target_partition_num_rows: 131072, num_sub_vectors: 16, metric: cosine })]
DataframeQueryView(
 dataset=DatasetEntry(name='droid:raw', id='187BCC812BF3DE98115e2d2a3b05a850'),
 query_expression=QueryExpression {
     view_contents: Some(
         ViewContentsSelector(
             {
                 /observation/gripper_position: None,
             },
         ),
     ),
     include_semantically_empty_columns: false,
     include_tombstone_columns: false,
     include_static_columns: Both,
     filtered_index: Some(
         "real_time",
     ),
     filtered_index_range: None,
     filtered_index_values: None,
     using_index_values: None,
     filtered_is_not_null: None,
     sparse_fill_strategy: None,
     selection: None,
 },
 partition_ids=["ILIAD_50aee79f_2023_07_12_20h_28m_36s"]
)
RecordingStream(Some(
    StoreInfo {
        store_id: StoreId(
            Recording,
            "foo",
            "cabac73b15f246299fc59ace2c4fb278",
        ),
        cloned_from: None,
        store_source: PythonSdk(
            3.12.11,
        ),
        store_version: Some(
            CrateVersion {
                major: 0,
                minor: 28,
                patch: 0,
                meta: Some(
                    DevAlpha {
                        alpha: 1,
                        commit: None,
                    },
                ),
            },
        ),
        is_partial: false,
    },
))
GrpcSink(ProxyUri {
    origin: Origin {
        scheme: RerunHttp,
        host: Ipv4(
            127.0.0.1,
        ),
        port: 9876,
    },
})
FileSink("out.rerun")
```

abey79: (I'm missing a closed parenthesis)

abey79: `print(client.all_entries())`:

```
[Entry(EntryKind.DATASET, 'random_sample_video_data_compacted'), Entry(EntryKind.DATASET, 'droid:70k'), Entry(EntryKind.DATASET, 'test_ds_500'), Entry(EntryKind.DATASET, 'rr_2925_repro'), Entry(EntryKind.DATASET, 'embedding_timeseries_demo:16'), Entry(EntryKind.DATASET, 'vecbench:wide:400K'), Entry(EntryKind.DATASET, 'test_ds_37'), Entry(EntryKind.DATASET, 'embedding_timeseries_demo:nick_0'), Entry(EntryKind.DATASET, 'random_sample_video_data'), Entry(EntryKind.DATASET, 'droid:sample500'), Entry(EntryKind.DATASET, 'embedding_timeseries_demo:17'), Entry(EntryKind.DATASET, 'vecbench:bench3:10K'), Entry(EntryKind.DATASET, 'embedding_timeseries_demo:18'), Entry(EntryKind.DATASET, 'benchmark_registration_sample500'), Entry(EntryKind.DATASET, 'droid:sample50'), Entry(EntryKind.DATASET, 'test_ds_627'), Entry(EntryKind.DATASET, 'sample_video_data_compacted'), Entry(EntryKind.DATASET, 'sample_video_data'), Entry(EntryKind.DATASET, 'droid:sample8k'), Entry(EntryKind.DATASET, 'droid:raw'), Entry(EntryKind.DATASET, 'test_ds_150'), Entry(EntryKind.DATASET, 'droid:sample2k'), Entry(EntryKind.TABLE, '__entries')]
```

ntjohnson1: Looks great!

rerun_py/rerun_bindings/rerun_bindings.pyi
abey79: https://linear.app/rerun/issue/RR-3053/rename-rerun-bindings-to-rerun-internal

PR #11997, opened 2025-11-26
Remove load_recording docs

ntjohnson1:
### Related
RR-3026

### What
The OSS server is the preferred route so start guiding people towards that.

### DIFF
--- a/docs/content/getting-started/data-out/query-data.md
+++ b/docs/content/getting-started/data-out/query-data.md
@@ -101,10 +101,10 @@ In order to add additional recordings to a dataset we use the `register` API.
 ```python
 # For OSS server you must register files local to your machine
 # To synchronously register a single recording
-dataset.register(f"file://{os.path.abspath('oss_demo.rrd')}")
+dataset.register(Path("oss_demo.rrd").resolve().as_uri())
 # To asynchronously register many recordings
 timeout_seconds = 100
-tasks = dataset.register_batch([f"file://{os.path.abspath('oss_demo.rrd')}"])
+tasks = dataset.register_batch([Path("oss_demo.rrd").resolve().as_uri()])
 tasks.wait(100)
 ```
 


abey79 (approved):
Thanks. Somewhat orthogonal, but could you drive-by change the remaining doc page to use `Path.as_uri()`? It's much cleaner than than the `f"file:/{...}"` thing imo.

ntjohnson1:
> Thanks. Somewhat orthogonal, but could you drive-by change the remaining doc page to use `Path.as_uri()`? It's much cleaner than than the `f"file:/{...}"` thing imo.

Done! This does raise the observation that our type hint for register say `str` but probably support `PathLike` since I tested passing a path in works.

abey79:
> > Thanks. Somewhat orthogonal, but could you drive-by change the remaining doc page to use `Path.as_uri()`? It's much cleaner than than the `f"file:/{...}"` thing imo.
> 
> Done! This does raise the observation that our type hint for register say `str` but probably support `PathLike` since I tested passing a path in works.

I'm surprised that it works. I'm pretty sure it doesn't work everywhere yet. But yeah, it would be nice to accept `PathLike` and implicitly call `.as_uri()` everywhere. I'm currently writing Python wrappers for `DatasetEntry` and `TableEntry`. It will be trivial to add that to the wrapping code.

ntjohnson1:
> > > Thanks. Somewhat orthogonal, but could you drive-by change the remaining doc page to use `Path.as_uri()`? It's much cleaner than than the `f"file:/{...}"` thing imo.
> > 
> > 
> > Done! This does raise the observation that our type hint for register say `str` but probably support `PathLike` since I tested passing a path in works.
> 
> I'm surprised that it works. I'm pretty sure it doesn't work everywhere yet. But yeah, it would be nice to accept `PathLike` and implicitly call `.as_uri()` everywhere. I'm currently writing Python wrappers for `DatasetEntry` and `TableEntry`. It will be trivial to add that to the wrapping code.

We can't implicitly call uri right? Since we don't know if it's local or a cloud destination?

abey79:
> > > > Thanks. Somewhat orthogonal, but could you drive-by change the remaining doc page to use `Path.as_uri()`? It's much cleaner than than the `f"file:/{...}"` thing imo.
> > > 
> > > 
> > > Done! This does raise the observation that our type hint for register say `str` but probably support `PathLike` since I tested passing a path in works.
> > 
> > 
> > I'm surprised that it works. I'm pretty sure it doesn't work everywhere yet. But yeah, it would be nice to accept `PathLike` and implicitly call `.as_uri()` everywhere. I'm currently writing Python wrappers for `DatasetEntry` and `TableEntry`. It will be trivial to add that to the wrapping code.
> 
> We can't implicitly call uri right? Since we don't know if it's local or a cloud destination?

Not sure I follow, If you pass a `Path` to register, then it has to be a local path that is also visible to OSS server. Aka only works for a local workflow. If you have a s3 uri, then it has to be a `str`, right?

ntjohnson1:
> Not sure I follow, If you pass a Path to register, then it has to be a local path that is also visible to OSS server. Aka only works for a local workflow. If you have a s3 uri, then it has to be a str, right?

They could store the S3 path as a path though. There appears to be a cloud pathlib we might want to encourage for that usecase.

PR #11998, opened 2025-11-26
`re_redap_client`: make it easy to query for chunk ids

emilk:
* Part of RR-3000
* Part of bigger-than-RAM support

This adds the ability to easily query for all the `ChunkId`s in a recording, or all in a specific time range.

Best reviewed commit-by-commit!

### DIFF
--- a/crates/store/re_redap_client/Cargo.toml
+++ b/crates/store/re_redap_client/Cargo.toml
@@ -42,7 +42,6 @@ re_smart_channel.workspace = true
 re_types_core.workspace = true
 re_uri.workspace = true
 
-anyhow.workspace = true # TODO(emilk): remove
 arrow.workspace = true
 itertools.workspace = true
 jiff.workspace = true
--- a/crates/store/re_redap_client/src/connection_client.rs
+++ b/crates/store/re_redap_client/src/connection_client.rs
@@ -440,7 +440,7 @@ where
     pub async fn query_dataset_chunk_index(
         &mut self,
         params: PartitionQueryParams,
-    ) -> anyhow::Result<Vec<ChunkIndex>> {
+    ) -> Result<Vec<ChunkIndex>, ApiError> {
         self.query_dataset_raw(params)
             .await?
             .collect::<Vec<_>>()
@@ -464,8 +464,11 @@ where
                 })
             })
             .map(|batch| {
-                let rb = arrow::array::RecordBatch::try_from(batch?)?;
+                let rb = arrow::array::RecordBatch::try_from(batch?).map_err(|err| {
+                    ApiError::serialization(err, "failed converting to RecordBatch")
+                })?;
                 ChunkIndex::from_record_batch(rb)
+                    .map_err(|err| ApiError::serialization(err, "failed creating ChunkIndex"))
             })
             .collect()
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9369,7 +9369,6 @@ dependencies = [
 name = "re_redap_client"
 version = "0.28.0-alpha.1+dev"
 dependencies = [
- "anyhow",
  "arrow",
  "itertools 0.14.0",
  "jiff",


crates/store/re_chunk_store/src/writes.rs
```diff
@@ -53,22 +57,15 @@ impl ChunkStore {
         }
 
         if self.chunks_per_chunk_id.contains_key(&chunk.id()) {
-            // We assume that chunk IDs are unique, and that reinserting a chunk has no effect.
-            re_log::debug_once!(
-                "Chunk #{} was inserted more than once (this has no effect)",
-                chunk.id()
-            );
+            // We assume that chunk IDs are unique, and that reinserting a
```
grtlr:
```suggestion
            // We assume that chunk IDs are unique, and that reinserting a chunk has no effect.
```

crates/store/re_chunk_store/src/writes.rs
grtlr, line 1: What's the motivation for removing information from the `log` calls? I'm guessing not wanting to overload the user with information that they don't care about?

Maybe it makes sense to have a `debug_once!` call in addition to the `warn_once!`?

crates/store/re_redap_client/src/grpc.rs
```diff
@@ -479,16 +479,14 @@ async fn stream_partition_from_server(
         Ok(true)
     };
 
-    let exclude_static_data = false;
-    let exclude_temporal_data = true;
     let static_chunk_stream = client
-        .fetch_partition_chunks(
+        .fetch_partition_chunks(PartitionQueryParams {
```
grtlr, line 483: I like the named arguments here!

crates/store/re_redap_client/src/connection_client.rs
```diff
@@ -375,43 +385,99 @@ where
...
         let query_request = QueryDatasetRequest {
             partition_ids: vec![partition_id.into()],
             chunk_ids: vec![],
             entity_paths: vec![],
             select_all_entity_paths: true,
             fuzzy_descriptors: vec![],
-            exclude_static_data,
-            exclude_temporal_data,
+            exclude_static_data: !include_static_data,
+            exclude_temporal_data: !incclude_temporal_data,
```
grtlr:
```suggestion
            exclude_temporal_data: !include_temporal_data,
```

crates/store/re_redap_client/Cargo.toml
```diff
@@ -39,8 +39,10 @@ re_log_types.workspace = true
 re_log.workspace = true
 re_protos.workspace = true
 re_smart_channel.workspace = true
+re_types_core.workspace = true
 re_uri.workspace = true
 
+anyhow.workspace = true # TODO(emilk): remove
```
grtlr: Why, when, how? If this `TODO` stays around, please add a proper comment.

crates/store/re_types_core/src/chunk_index.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+}
+
+fn chunk_ids_from_rb(rb: &RecordBatch) -> anyhow::Result<&[ChunkId]> {
+    let chunk_ids = rb
+        .column_by_name("chunk_id")
+        .ok_or_else(|| anyhow::format_err!("Missing chunk_id column in ChunkIndex"))?;
+
+    let chunk_ids = chunk_ids.as_fixed_size_binary_opt().ok_or_else(|| {
+        anyhow::format_err!("Expected ChunkIDd to be encoded as a FixedSizeBinary array")
```
grtlr:
```suggestion
        anyhow::format_err!("Expected ChunkId to be encoded as a FixedSizeBinary array")
```

crates/store/re_types_core/src/chunk_index.rs
grtlr, line 1: I'd say just remove the dependency on `anyhow` and just use `thiserror`.

crates/viewer/re_ui/src/design_tokens.rs
```diff
@@ -201,6 +201,8 @@ pub struct DesignTokens {
 
     pub density_graph_selected: Color32,
     pub density_graph_unselected: Color32,
+
+    /// This is the color of time rangres that has only been partially loaded.
```
grtlr:
```suggestion
    /// This is the color of time ranges that has only been partially loaded.
```

grtlr (changes requested):
`ChunkIndex` looks like a nice addition, but I don't think we should use the self-reference until we benchmarked that it really makes a difference.

 Until then I'd just use `.unwrap` + casts / `transmute` (as we already validated during construction).
 
 But also I don't feel overly strong about it, your call 👍.

crates/store/re_chunk_store/src/writes.rs
emilk: Yes - my terminal was filling up with spam with chunk ids and row ids that I could not act on

PR #12002, opened 2025-11-27
Make `InstancePoses3D` queries independent of `Transform3D::child_frame`

Wumpf:
### Related

* Part (!) of RR-2627

### What

While we were still figuring out the future of instance pose handling we tied their queries to be atomic along `Transform3D::child_frame`.
We've decided by now that `InstancePoses3D` should only be scoped to entity paths (which also means that for a single frame you can have many different sets of poses building on top!). This is a first step towards this, staring from the backend: queries are longer "stopping" along `Transform3D::child_frame`.

### DIFF
--- a/crates/store/re_tf/src/transform_queries.rs
+++ b/crates/store/re_tf/src/transform_queries.rs
@@ -169,7 +169,9 @@ fn atomic_latest_at_query(
                 continue;
             }
         } else {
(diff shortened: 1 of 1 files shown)


grtlr (approved):
Nice! Thank you for adding the additional test cases

grtlr (approved):
Puh, glad we caught that bug!

PR #12005, opened 2025-11-27
Load URDF with frame IDs and named transforms

MichaelGrupp:
- add `CoordinateFrame`s to link entities and visual/collision meshes, attach meshes via `InstancePoses3D` to their link
- log (default) joints as `Transform3D`, with `{parent, child}_frame` connecting a pair of links
- update `animated_urdf` example

Thanks to #12101, we can still load & visualize an URDF with default static transforms __and__ update them via TF-like transforms. 

Documentation will come separate (#12056).

---

> ⚠️ The `animated_urdf` is slightly less elegant now because we can't just update the rotation part.
>
> See the TODO there. We can probably make use of @oxkitsune parallel work that will expose the right utilities for the transforms in `urdf_rs`, so we don't need to use `glam` anymore. So I will leave it like that for now.

### DIFF
--- a/crates/store/re_data_loader/src/loader_urdf.rs
+++ b/crates/store/re_data_loader/src/loader_urdf.rs
@@ -7,7 +7,7 @@ use anyhow::{Context as _, bail};
 use itertools::Itertools as _;
 use re_chunk::{ChunkBuilder, ChunkId, EntityPath, RowId, TimePoint};
 use re_log_types::{EntityPathPart, StoreId};
-use re_sdk_types::archetypes::{Asset3D, CoordinateFrame, Transform3D};
+use re_sdk_types::archetypes::{Asset3D, CoordinateFrame, InstancePoses3D, Transform3D};
 use re_sdk_types::datatypes::Vec3D;
 use re_sdk_types::external::glam;
 use re_sdk_types::{AsComponents, Component as _, ComponentDescriptor, SerializedComponentBatch};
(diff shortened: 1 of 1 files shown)


crates/store/re_data_loader/src/loader_urdf.rs
```diff
@@ -282,6 +282,23 @@ fn log_robot(
...
+        &CoordinateFrame::update_fields().with_frame(urdf_tree.root.name.clone()),
+    )?;
+    // TODO: Do we always need to do that?
+    send_archetype(
+        tx,
+        store_id,
+        entity_path.clone(),
+        timepoint,
+        &Transform3D::update_fields().with_child_frame(urdf_tree.root.name.clone()),
+    )?;
```
MichaelGrupp: Connection to TF root shouldn't be implicit!

examples/rust/animated_urdf/src/main.rs
```diff
@@ -47,17 +47,16 @@ fn run(rec: &rerun::RecordingStream, _args: &Args) -> anyhow::Result<()> {
...
-                // NOTE: each joint already has a fixed origin pose (logged with the URDF file),
-                // and Rerun won't allow us to override or add to that transform here.
-                // So instead we apply the dynamic rotation to the child link of the joint:
-                let child_link = urdf.get_joint_child(joint);
-                let link_path = urdf.get_link_path(child_link);
+                // Rerun loads the URDF transforms with child/parent frame relations.
+                // In order to move a joint, we just need to log a new transform between two of those frames.
                 rec.log(
-                    link_path,
+                    "/transforms",
```
MichaelGrupp, line 71: This is currently not working if the URDF is logged together with the transforms. Waiting for "static-like" TF that can be updated.

MichaelGrupp: Will work with #12101 🎉

oxkitsune:
so nice :shipit:

PR #12006, opened 2025-11-27
Transform frame ancestors in UI

IsseW:
### Related

- Closes RR-3016

### What

UI to display transform frame ancestor in selection panel, and tooltips.

In selection panel:
Screenshot 2025-11-27 at 15 30 39

Hovering:
Screenshot 2025-11-27 at 15 30 59

Test with https://rerun.io/viewer/pr/12006?url=https%3A%2F%2Fapp.rerun.io%2Fversion%2F0.27.2%2Fexamples%2Fsnippets%2Farchetypes%2Ftransform3d_hierarchy_frames.rrd


crates/viewer/re_viewer_context/src/cache/transform_database_store.rs
IsseW, line 1: Some discussion on moving this: https://rerunio.slack.com/archives/C09J226VB46/p1764235418045279

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -480,6 +480,11 @@ impl TransformsForChildFrame {
         }
     }
 
+    /// The entity path that produces information for this frame.
+    pub fn associated_entity_path(&self) -> &EntityPath {
+        &self.associated_entity_path
```
Wumpf: TODO SELF NOTE: how is this used

crates/viewer/re_viewer_context/Cargo.toml
```diff
@@ -44,6 +44,7 @@ re_types_core.workspace = true
 re_ui.workspace = true
 re_uri.workspace = true
 re_video = { workspace = true, features = ["serde"] }
+re_tf.workspace = true
```
Wumpf: sorting!

crates/viewer/re_data_ui/src/extra_data_ui.rs
```diff
@@ -29,6 +30,17 @@ impl ExtraDataUi {
...
+            .or_else(|| {
+                transform_frames::TransformFramesUi::from_components(
+                    ctx,
+                    query,
+                    entity_path,
+                    descr,
+                    chunk,
+                    entity_components,
+                )
+                .map(Self::TransformHierarchy)
```
Wumpf, line 40: nit: if this is `Self::` the above should be as well

crates/viewer/re_data_ui/src/transform_frames.rs
```diff
@@ -0,0 +1,229 @@
...
+    ComponentDescriptor, TransformFrameIdHash, archetypes,
+    components::{self, TransformFrameId},
+};
+use re_ui::{HasDesignTokens as _, UiExt as _, UiLayout, icons};
+use re_viewer_context::{
+    Item, SystemCommand, SystemCommandSender as _, TransformDatabaseStoreCache, ViewerContext,
+};
+
+/// The max amount of ancestors we show before putting a '…'.
+const MAX_SHOWN_ANCESTORS: usize = 100;
```
Wumpf, line 13: good you added a limit, but.. that's a lot ;-). Maybe
```suggestion
const MAX_SHOWN_ANCESTORS: usize = 10;
```

crates/viewer/re_data_ui/src/transform_frames.rs
```diff
@@ -0,0 +1,229 @@
...
+impl TransformFramesUi {
+    pub fn from_components(
+        ctx: &ViewerContext<'_>,
+        query: &re_chunk_store::LatestAtQuery,
+        entity_path: &EntityPath,
+        transform_frame_descr: &ComponentDescriptor,
+        transform_frame_chunk: &UnitChunkShared,
+        entity_components: &[(ComponentDescriptor, UnitChunkShared)],
+    ) -> Option<Self> {
+        // All frame descriptor components, sorted by pirority.
```
Wumpf:
```suggestion
        // All frame descriptor components, sorted by priority.
```

crates/viewer/re_data_ui/src/transform_frames.rs
```diff
@@ -0,0 +1,229 @@
...
+        //
+        // The bool indicates if we should use the entity path as
+        // the current frame.
+        let frame_components = [
```
Wumpf: this seems very brittle to me. Right now this just lists out everything with type the compoennt type `TransformFrameId`, could use that for most things 🤔 
Generally I had a hard time following this thing, might be good to actually create a struct so we can have a more easy to read "spec" on this

crates/viewer/re_data_ui/src/transform_frames.rs
```diff
@@ -0,0 +1,229 @@
...
+                archetypes::Transform3D::descriptor_child_frame().component,
+                false,
+            ),
+            (
+                archetypes::Pinhole::descriptor_child_frame().component,
+                false,
+            ),
+            (
+                archetypes::CoordinateFrame::descriptor_frame().component,
+                false,
```
Wumpf: so if there is a coordinate frame I think we should use that instead of anything else. I'm still pondering whether we should _only_ ever use that since that's what governs what we actually visualize 🤔 
(with.. some weird exception on Pinhole actually :/)
Worth talking about

crates/viewer/re_data_ui/src/transform_frames.rs
```diff
@@ -0,0 +1,229 @@
...
+            .filter_map(|(desc, _)| find_frame_component(desc.component))
+            .any(|(p, ..)| p < priority)
+        {
+            return None;
+        }
+        let mut frame_id_hash = if use_entity_path {
+            TransformFrameIdHash::from_entity_path(entity_path)
+        } else {
+            let frame_id = transform_frame_chunk
+                .component_mono::<components::TransformFrameId>(transform_frame_descr.component)?
```
Wumpf: can we query blueprint overrides here if we know that we have a DataResult selected? 🤔

crates/viewer/re_data_ui/src/transform_frames.rs
```diff
@@ -0,0 +1,229 @@
...
+
+/// The max amount of ancestors we show in a tooltip.
+///
+/// We want this to be less because we don't use a scroll area for
+/// tooltips.
+const MAX_SHOWN_ANCESTORS_TOOLTIP: usize = 3;
+
+struct TransformFrameInfo {
+    frame_id: TransformFrameId,
+    source_entity: Option<EntityPath>,
```
Wumpf, line 23: should document that this is the entity that has connectects this frame_id with a parent frame. Got very confused on my reading and thought it was one of the entities having this frame as `CoordinateFrame`

also the `None` case is somewhat peculiar, as I understand so far that's really just an error case 🤔

Wumpf (changes requested):
promising direction, there's some stuff I'd like to talk through though. Also I thin it's worth looking into whether we can make the ui a bit more compact. But we can also wait until Katya has time for us to talk about this

crates/viewer/re_data_ui/src/transform_frames.rs
IsseW: > also the None case is somewhat peculiar, as I understand so far that's really just an error case 🤔

In the `transform3d_hierarchy_frames` snippet, I get that `sun` is None.

crates/viewer/re_data_ui/src/transform_frames.rs
```diff
@@ -0,0 +1,219 @@
...
+        ctx: &ViewerContext<'_>,
+        query: &re_chunk_store::LatestAtQuery,
+        transform_frame_descr: &ComponentDescriptor,
+        transform_frame_chunk: &UnitChunkShared,
+        entity_components: &[(ComponentDescriptor, UnitChunkShared)],
+    ) -> Option<Self> {
+        // All frame descriptor components, sorted by priority.
+        //
+        // The bool indicates if we should use the entity path as
+        // the current frame.
```
Wumpf: the bool is gone ;-)

crates/viewer/re_data_ui/src/transform_frames.rs
```diff
@@ -0,0 +1,219 @@
...
+
+impl TransformFramesUi {
+    pub fn from_components(
+        ctx: &ViewerContext<'_>,
+        query: &re_chunk_store::LatestAtQuery,
+        transform_frame_descr: &ComponentDescriptor,
+        transform_frame_chunk: &UnitChunkShared,
+        entity_components: &[(ComponentDescriptor, UnitChunkShared)],
+    ) -> Option<Self> {
+        // All frame descriptor components, sorted by priority.
```
Wumpf: it's not all of them anymore. it's a very select list and we should probably note down our learnings there on why we pick those :)

Wumpf (approved):
Design will definitely need some more love, but I think now with expandable & hidden by default this doesn't make our selection panel look much worse while at the same time be very useful for deliberate debugging the hierarchy 👍 

Plz clean up the code comments I pointed out before merging, but otherwise good to go I think!

Issue #11993, opened 2025-11-26
Cannot send 2GiB+ chunks over gRPC


nathan-b-flya:
**Describe the bug**

When sending too much data, Rerun crashes with the error:
```
[... ERROR re_grpc_client::write] Failed to flush gRPC messages during shutdown: gRPC connection is open, but flush still failed. Probably a bug in the Rerun SDK
[... ERROR re_grpc_client::write] Failed to gracefully shut down message proxy client: channel closed
[... ERROR re_grpc_server] Error while receiving messages: gRPC error, message: "h2 protocol error: error reading a body from connection"
```

**To Reproduce**
Here is the C++ code to reproduce the issue easily `rerun_test.cpp`:
```c++
#include <cstdlib>
#include <numeric>
#include <random>
#include <string>
#include <vector>

#include <rerun.hpp>
#include <rerun/components/color.hpp>
#include <rerun/components/position3d.hpp>

rerun::components::Color gray(uint8_t value) { return {value, value, value}; }

int main()
{
    const auto rec = rerun::RecordingStream("github_issue");
    rec.spawn().exit_on_failure();

    const std::size_t nb_scans = 2200;  // does not work
    // const std::size_t nb_scans = 2100;  // works
    const std::size_t nb_points_per_scan = 1024 * 64;
    const std::size_t nb_points = nb_scans * nb_points_per_scan;

    std::vector<uint64_t> times(nb_scans);
    std::iota(times.begin(), times.end(), 0);
    const auto timeline = rerun::TimeColumn::from_sequence("times", std::move(times));

    std::vector<uint32_t> lengths(nb_scans, nb_points_per_scan);

    std::mt19937 gen{42};
    std::uniform_real_distribution<float> distrib(0., 1.);
    std::uniform_int_distribution<int> distrib_bis(0, 255);

    std::vector<rerun::components::Position3D> points(nb_points);
    std::generate(points.begin(), points.end(),
                  [&]() {
                      return rerun::components::Position3D{distrib(gen), distrib(gen),
                                                           distrib(gen)};
                  });

    [[maybe_unused]] const rerun::components::Color BLUE{0, 0, 255};
    std::vector<rerun::components::Color> colors(nb_points);
    std::generate(colors.begin(), colors.end(), [&]()
                  { return gray(static_cast<uint8_t>(distrib_bis(gen))); });  // does not work
                  // { return BLUE; });  // works

    rec.send_columns("scans", timeline,
                     rerun::Points3D()
[...]

Issue #11971, opened 2025-11-25
Track lineage when merging/splitting chunks


teh-cmc:
We currently generate brand new `ChunkId`s whenever we merge/split chunks (e.g. during JIT compaction), because it's the easy way to prevent collisions with minimal effort, but that comes at the cost of losing track of all of the Chunk's lineage.

We should at the very least keep track of these things in the chunk's metadata, e.g.:
```
* rerun:parents: [chunk_xxx, chunk_yyy]
```

PR #12008, opened 2025-11-27
Fix `ChunkId` clash due to chunk splitting

emilk:
**Review commit-by-commit!**

We recently introduced chunk splitting on-ingest. The code used the very dangerous `ChunkId::next`, which can easily cause clashes.

The problem is this: we split chunks that are over some byte limit in even pieces, where "even" is measured in _row count_. This means that some of the smaller pieces may still be over the size limit. The caller (`ChunkStore::insert_chunk`) will in these cases recurse, i.e. split the smaller pieces further.

This case WILL cause `ChunkId` collisions, if the split algo uses `ChunkId::next`, which it did.

## Background
I wanted to understand the code a bit more, and also make it re-usable for fixing https://github.com/rerun-io/rerun/issues/11993

## Related
* Code added in https://github.com/rerun-io/rerun/pull/11921
* The refactor unblocks a quick-fix for https://github.com/rerun-io/rerun/issues/11993
* Related: https://github.com/rerun-io/rerun/issues/11971

### DIFF
--- a/crates/store/re_chunk/src/split.rs
+++ b/crates/store/re_chunk/src/split.rs
@@ -5,7 +5,7 @@ use re_types_core::ChunkId;
 
 use crate::Chunk;
 
-/// See [`Chunk::split`].
+/// See [`Chunk::split_chunk_if_needed`].
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct ChunkSplitConfig {
     /// Split chunks larger than this.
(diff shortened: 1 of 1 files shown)


teh-cmc:
> The problem is this: we split chunks that are over some byte limit in even pieces, where "even" is measured in _row count_. This means that some of the smaller pieces may still be over the size limit. The caller (`ChunkStore::insert_chunk`) will in these cases recurse, i.e. split the smaller pieces further.

Haaaaa, there we go. This is the proper explanation for the original collisions. Nice catch.

I completely missed the fact that this was a recursive call 🤦

PR #12012, opened 2025-11-28
Show progress bar when loading recordings from a server

emilk:
Best reviewed commit-by-commit

### Related
* Part of RR-3000
* Does part of https://github.com/rerun-io/rerun/issues/11971

### What
When streaming a recording from a gRPC server, we now first fetch all the chunk ids that are in the recording.
These are then put into a new `ChunkIndex` - a secondary index in `EntityDb`.
This lets us show a rough progress bar when loading recordings!

The progress is based on number of chunks received divided by total number of chunks in server.
In the future we can improve this by also taking the _size_ of each chunk into account.

image

The progress bar is shown the the top left panel, under the recording name.
The progress bar is only shown while loading is in progress. It disappear when the recording has fully loaded.

### DIFF
--- a/crates/viewer/re_selection_panel/tests/snapshots/selection_panel_recording.png
+++ b/crates/viewer/re_selection_panel/tests/snapshots/selection_panel_recording.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:093b18138000cdfa2f3fd974216a2a776a7323cb2ab9e3347e66118d39afdf16
-size 49271
+oid sha256:07021f9f969767c4115acaf21ec75b3c79af7ba6c6191e0d3b9f6d9d1612d232
+size 49405
--- a/crates/viewer/re_selection_panel/tests/snapshots/selection_panel_recording_hover_app_id.png
+++ b/crates/viewer/re_selection_panel/tests/snapshots/selection_panel_recording_hover_app_id.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:ac66c18b76a916a509013eb4d72a056485facfda96e733dd67dbf643b1ea7e1d
-size 61490
+oid sha256:1d11fd68ca605c87ff9b5cd66ba2795afa13d09aa9513c4ce7a6a7f6c6e19d31
+size 61814


lucasmerlin:
I think it's weird, that we first show a spinner, then nothing, and then the progress bar 🤔 Feels kinda inconsistent. Maybe we should show the progress bar from the beginning? Maybe as an [indeterminate](https://codepen.io/tmac/pen/QgVRKb) progress bar until we get "real" progress?

https://github.com/user-attachments/assets/b12b1c7a-19d6-49b2-80e6-a64c42327fd4

Also not sure about the placement, I feel like it throws off the balance in the list items? @gavrelina might want to try this. 

I really do like showing the progress though!

Wumpf:
agreed with Lucas. It's also a bit weirdly tight and moving over the x

image

I think I'd prefer a design where the entire background becomes the loading bar. But all that said I'm sure the first thing @gavrelina would come up with is better than any of that.
Buuut either way I don't think any of these problems are an argument for not having it at all. It's really great to have the ability to show progress there in the first place! <3

crates/store/re_entity_db/src/chunk_index.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /// [0, 1], how many chunks have been loaded?
+    ///
+    /// Returns `None` if we have already started garbage-collecting some chunks.
+    pub fn progress(&self) -> Option<f32> {
+        if self.has_deleted {
+            None
+        } else if self.num_chunks() == 0 {
+            Some(1.0)
```
lucasmerlin, line 74: Can we keep track of wheter we received the chunk index yet and return 0.0 here until we are sure that there are actually 0 chunks?

crates/store/re_entity_db/src/entity_db.rs
```diff
@@ -584,47 +598,58 @@ impl EntityDb {
...
-                self.tree
-                    .on_store_deletions(&engine, &entity_paths_with_deletions, &store_events);
-            }
+        self.on_store_events(&store_events);
 
-            // We inform the stats last, since it measures e2e latency.
-            self.stats.on_events(timestamps, &store_events);
-        }
+        // We inform the stats last, since it measures e2e latency.
+        self.stats.on_events(timestamps, &store_events);
```
Wumpf: why is this not part of the new `on_store_events`?

crates/store/re_entity_db/src/entity_db.rs
```diff
@@ -584,47 +598,58 @@ impl EntityDb {
...
 
         Ok(store_events)
     }
 
+    /// We call this on any changes, before returning the store events to the outsider caller.
+    fn on_store_events(&mut self, store_events: &[ChunkStoreEvent]) {
+        re_tracing::profile_function!();
+
+        let mut engine = self.storage_engine.write();
```
Wumpf, line 629: a bit _meh_ that there's repeated lock tacking now which was just one before.

crates/store/re_entity_db/src/entity_db.rs
```diff
@@ -657,6 +682,8 @@ impl EntityDb {
                 .write()
                 .cache()
                 .purge_fraction_of_ram(fraction_to_purge);
+        } else {
+            self.on_store_events(&store_events);
```
Wumpf, line 690: I'm confused about all the additional stuff that are happening now through this call. Why did we get away without it so far? Are we doing this twice for some reason now since all the handling already happens on `self.gc`? 🤔 
Conceptually it makes a lot of sense to me. There's events and we should react to them.

crates/store/re_entity_db/src/chunk_index.rs
```diff
@@ -0,0 +1,132 @@
...
+pub struct ChunkInfo {
+    /// Do we have the whole chunk in memory?
+    pub fully_loaded: bool,
+}
+
+/// A secondary index that keeps track of which chunks have been loaded into memory.
+///
+/// This is currently used to show a progress bar.
+/// TODO(RR-2999): use this for larger-than-RAM.
+#[derive(Default, Debug, Clone)]
```
Wumpf, line 24: I'm missing a quick overview of how this relates to ChunkIndexMessage and how it behaves in absence (filling in via regular store messages)

crates/store/re_entity_db/src/entity_db.rs
emilk: The `&mut self` guarantees there is no race here, so the only downside is a very minor performance hit

crates/store/re_entity_db/src/entity_db.rs
emilk: There was a bug in the old code. We should have called it here, but we didn't.

crates/store/re_entity_db/src/entity_db.rs
Wumpf: 👍  yeah no concern really. 
Since we have `&mut self`, can't we skip the lock and do `get_mut()`? Or was `storage_engine` more bespoke than a regular lock

crates/store/re_entity_db/src/entity_db.rs
Wumpf: yikes!

PR #12015, opened 2025-11-28
Better discovery for view's visualizer errors

IsseW:
### Related

* Closes RR-2996

### What

Adds an icon button to discover visualizer errors.

Pressing the button gives a list of errors:
Screenshot 2025-11-28 at 13 10 48

Hovering the button shows how many there are:
Screenshot 2025-11-28 at 13 10 56


Wumpf:
ohhh interesting. That goes much further than what I had in mind which is to merely highlight stuff in blue blueprint panel - we already do that today afaik if there's nothing visualizable. But this seems to be a good addition as well

IsseW:
> ohhh interesting. That goes much further than what I had in mind which is to merely highlight stuff in blue blueprint panel - we already do that today afaik if there's nothing visualizable. But this seems to be a good addition as well

I was thinking here it will always be visible, since you can have both blueprint panel & selection panel collapsed. Although could display error in the blueprint panel as well.

Wumpf:
Example screenshot for latest iteration with all three ways of highlighting the error:
image

crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
```diff
@@ -402,7 +417,21 @@ impl BlueprintTree {
         let is_item_hovered =
             ctx.selection_state().highlight_for_ui_element(&item) == HoverHighlight::Hovered;
 
-        let item_content = list_item::LabelContent::new(view_data.name.as_ref())
+        let has_error = errors.is_some_and(|errors| {
+            errors
+                .values()
+                .any(|error| matches!(error, VisualizerExecutionErrorState::Overall(_)))
```
Wumpf: let's make this a utility on the error state

crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
```diff
@@ -402,7 +417,21 @@ impl BlueprintTree {
         let is_item_hovered =
             ctx.selection_state().highlight_for_ui_element(&item) == HoverHighlight::Hovered;
 
-        let item_content = list_item::LabelContent::new(view_data.name.as_ref())
+        let has_error = errors.is_some_and(|errors| {
```
Wumpf: interesting, I didn't actually realize this possibility and that a "total visualizer failure" is indeed an overall viewer error. I think this makes sense.. buuuuut we don't show the error on click then. So we're in a situation where the view goes red here and clikcing it doesn't give us more information. To be fair we still have it in the top bar, so that's alright I think

crates/viewer/re_viewport/src/viewport_ui.rs
```diff
@@ -714,6 +720,110 @@ impl<'a> egui_tiles::Behavior<ViewId> for TilesDelegate<'a, '_> {
...
+        let visualizer_errors = self.view_states.visualizer_errors(view_id);
+
+        let error_count = visualizer_errors.map_or(0, |per_vis| {
+            per_vis
+                .values()
+                .map(|errors| match errors {
+                    re_viewer_context::VisualizerExecutionErrorState::Overall(_) => 1,
+                    re_viewer_context::VisualizerExecutionErrorState::PerEntity(errors) => {
+                        errors.len()
+                    }
```
Wumpf: would also make a nice utility on the error state.
btw. reason I keep bringing this up here is that I'm not set on the internals of that type/enum and having the variants show up less often might pay off 😅

crates/viewer/re_viewport/src/viewport_ui.rs
```diff
@@ -714,6 +720,110 @@ impl<'a> egui_tiles::Behavior<ViewId> for TilesDelegate<'a, '_> {
     }
 }
 
+impl TilesDelegate<'_, '_> {
+    fn visualizer_errors_button(&self, ui: &mut egui::Ui, view_id: ViewId) {
+        let visualizer_errors = self.view_states.visualizer_errors(view_id);
```
Wumpf: nit: right shift avoidance

```suggestion
        let Some(visualizer_errors) = self.view_states.visualizer_errors(view_id) else {
                return;
        };
```

crates/viewer/re_viewport/src/viewport_ui.rs
```diff
@@ -714,6 +720,110 @@ impl<'a> egui_tiles::Behavior<ViewId> for TilesDelegate<'a, '_> {
...
+                    re_viewer_context::VisualizerExecutionErrorState::PerEntity(errors) => {
+                        errors.len()
+                    }
+                })
+                .sum()
+        });
+
+        if let Some(visualizer_errors) = visualizer_errors
+            && error_count > 0
+        {
```
Wumpf: in line with above


```suggestion
        if error_count == 0 {
             return;
        }
```


If this was about errors in general and we expect others in the future then I guess this would be overzealous, but it's called `visualizer_errors_button`, so visualizer errors is the only thing we're intersted in 😜

crates/viewer/re_viewport/src/viewport_ui.rs
```diff
@@ -714,6 +720,110 @@ impl<'a> egui_tiles::Behavior<ViewId> for TilesDelegate<'a, '_> {
...
+            let errors = visualizer_errors.values().flat_map(|err| match err {
```
Wumpf: instead of counting errors first, why don't we just collect this list ans check on it? not perf critical here and would make stuff shorter/more concise

crates/viewer/re_viewport/src/viewport_ui.rs
```diff
@@ -714,6 +720,110 @@ impl<'a> egui_tiles::Behavior<ViewId> for TilesDelegate<'a, '_> {
...
+                                item_entity,
+                            )
+                            .clicked()
+                        {
+                            self.ctx
+                                .command_sender()
+                                .send_system(SystemCommand::set_selection(item));
+                            self.ctx
+                                .command_sender()
+                                .send_ui(re_ui::UICommand::ExpandSelectionPanel);
```
Wumpf, line 811: very nice! great detail, I think there's other places that should have done this elsewhere :)

PR #12017, opened 2025-11-28
Partition-to-segment rename (wave 1): redap layer

abey79:
### Related

- sibling https://github.com/rerun-io/dataplatform/pull/2080

### What

Title.

In this first wave, the starting point is the gRPC layer (`re_protos`, `re_redap_client`, etc.), and what ripples on the rest of the codebase. This leaves the UI and SDK mostly untouched for now.

This PR also includes a new `re_arrow_util` column rename utility that was needed by the sibling PR. I've marked it with a comment.

**NOTE**: this is a major breaking change at the redap layer.

### DIFF
--- a/crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
+++ b/crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
@@ -525,11 +525,6 @@ message QueryDatasetRequest {
   // all segments will be queried.
   repeated rerun.common.v1alpha1.SegmentId segment_ids = 11;
 
-  // DEPRECATED: Use segment_ids instead.
-  // Client can specify what segments are queried. If left unspecified (empty list),
-  // all segments will be queried.
-  repeated rerun.common.v1alpha1.PartitionId partition_ids = 2 [deprecated = true];
-
   // Client can specify specific chunk ids to include. If left unspecified (empty list),
   // all chunks that match other query parameters will be included.
   repeated rerun.common.v1alpha1.Tuid chunk_ids = 3;
@@ -576,6 +571,8 @@ message QueryDatasetRequest {
 
   reserved 1;
   reserved "dataset_id";
+  reserved 2;
+  reserved "partition_ids";
 }
 
 message QueryDatasetResponse {
@@ -974,9 +971,8 @@ message DatasetDetails {
   // The segment of the blueprint dataset corresponding to the default blueprint (if any).
   optional rerun.common.v1alpha1.SegmentId default_blueprint_segment = 5;
 
-  // DEPRECATED: Use default_blueprint_segment instead.
-  // The segment of the blueprint dataset corresponding to the default blueprint (if any).
-  optional rerun.common.v1alpha1.PartitionId default_blueprint = 4 [deprecated = true];
+  reserved 4;
+  reserved "default_blueprint";
 }
 
 message DatasetEntry {
--- a/crates/store/re_protos/proto/rerun/v1alpha1/common.proto
+++ b/crates/store/re_protos/proto/rerun/v1alpha1/common.proto
@@ -195,14 +195,6 @@ message SegmentId {
   optional string id = 1;
 }
 
-// DEPRECATED: Use SegmentId instead.
-// Kept for backward compatibility.
-message PartitionId {
-  option deprecated = true;
-
-  optional string id = 1;
-}
-
 message ComponentDescriptor {
   reserved 1;
   reserved "archetype_name";
--- a/crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
+++ b/crates/store/re_protos/src/v1alpha1/rerun.cloud.v1alpha1.ext.rs
@@ -108,15 +108,9 @@ impl Default for QueryDatasetRequest {
 }
 
 impl From<QueryDatasetRequest> for crate::cloud::v1alpha1::QueryDatasetRequest {
-    #[expect(
-        deprecated,
-        reason = "We need to set partition_ids for backward compatibility"
-    )]
     fn from(value: QueryDatasetRequest) -> Self {
         Self {
             segment_ids: value.segment_ids.into_iter().map(Into::into).collect(),
-            // Keep partition_ids empty - we only use segment_ids for new code
-            partition_ids: vec![],
             chunk_ids: value
                 .chunk_ids
                 .into_iter()
@@ -136,26 +130,13 @@ impl From<QueryDatasetRequest> for crate::cloud::v1alpha1::QueryDatasetRequest {
 impl TryFrom<crate::cloud::v1alpha1::QueryDatasetRequest> for QueryDatasetRequest {
     type Error = tonic::Status;
 
-    #[expect(
-        deprecated,
-        reason = "We need to read partition_ids for backward compatibility"
-    )]
     fn try_from(value: crate::cloud::v1alpha1::QueryDatasetRequest) -> Result<Self, Self::Error> {
         // Support both segment_ids (new) and partition_ids (deprecated) for backward compatibility
-        let segment_ids = if !value.segment_ids.is_empty() {
-            value
-                .segment_ids
-                .into_iter()
-                .map(TryInto::try_into)
-                .collect::<Result<Vec<_>, _>>()?
-        } else {
-            // Fallback to deprecated partition_ids
-            value
-                .partition_ids
-                .into_iter()
-                .map(TryInto::try_into)
-                .collect::<Result<Vec<_>, _>>()?
-        };
+        let segment_ids = value
+            .segment_ids
+            .into_iter()
+            .map(TryInto::try_into)
+            .collect::<Result<Vec<_>, _>>()?;
 
         Ok(Self {
             segment_ids,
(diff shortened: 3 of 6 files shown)


abey79:
@rerun-bot full-check

crates/utils/re_arrow_util/src/batches.rs
```diff
@@ -253,6 +256,29 @@ impl RecordBatchExt for RecordBatch {
...
+                Arc::clone(f)
+            })
+            .collect();
+
+        Self::try_new_with_options(
+            Arc::new(Schema::new_with_metadata(new_fields, metadata)),
+            columns,
+            &RecordBatchOptions::default().with_row_count(Some(row_count)),
+        )
+    }
```
abey79, line 281: new utility function here along with tests

crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
```diff
@@ -973,8 +971,12 @@ message DatasetDetails {
   // The blueprint dataset associated with this dataset (if any).
   optional rerun.common.v1alpha1.EntryId blueprint_dataset = 3;
 
-  // The partition of the blueprint dataset corresponding to the default blueprint (if any).
-  optional rerun.common.v1alpha1.PartitionId default_blueprint = 4;
+  // The segment of the blueprint dataset corresponding to the default blueprint (if any).
+  optional rerun.common.v1alpha1.SegmentId default_blueprint_segment = 5;
```
jleibs, line 972:
```suggestion
  // The id of the default blueprint (if any).
  optional rerun.common.v1alpha1.StoreId default_blueprint = 5;
```

Calling this a `blueprint_segment` (and even using SegmentId) feels wrong to me.

Logically blueprints don't make up a segmented dataset in the same way that recordings do.  They don't have an expectation of a shared schema. It's not something we intend to query, certainly not as a multi-blueprint query.

I've been throwing around the term "Collection" for a while as a schema-free "bag of recordings", and it's probably the right fit for blueprints as well. But, I don't think we should use "segment" to refer to elements of a collection.

 It's worth noting that I don't think of `segment_id` as a property of the recording. Rather, it is the id of a recording *within* a dataset.

"When adding a recording to the dataset, by default it is assigned a `segment_id` based on the recording_id, though this segment id can be overridden at registration time, such as is necessary when re-segmenting a view."

abey79: Some many things are mixed here:
- what dataset blueprint are logically?
- how they are implemented physically?
- recording id vs. segment id?
- "collection mode"?

In general, I agree with everything i'm reading in the above comment. However, the blunt reality is: today, blueprints are stored as segment in a (blueprint) dataset, and referred to as any other segment, aka `(entry id, segment id)`. I absolutely don't want to change how things are implemented in this PR. In particular, so long as we don't change how blueprint storage is implemented, using `StoreId` here doesn't make (technical) sense.

Again, I'm not saying I like how blueprint storage is implemented. Nor does Niko btw. This is why the API update project includes a drastic cleanup of how the dataset blueprints are interacted with at the Python SDK level (basically, implicit key-value store of blueprint attached to every dataset).

If we want to change the way blueprint are stored and presented at the redap level, sure, but we need a new project.

The only thing I'm willing to bike shed here the name of the field. Adding `_segment` actually was Claude's initiative. I'm happy to remove it if we feel better about it, but it won't change any of the above.

jleibs: Let's stick with `default_blueprint_segment` (otherwise you get a collision on the field name you're trying to deprecate), and clean up the whole thing in a future.

How about a related proposal then:
- What if we add code that forbids using APIs other than SegmentTable / FetchSegment on blueprint datasets.  I believe this will simplify a future move of blueprints to a non-dataset structure since I won't have to worry about backwards compatibility for those APIs.

abey79: I'm not against it, but i'm not sure this is necessary. Blueprint dataset are not really visible from the python side, except via `__entries`. Also, the API update will no longer expose anything about a dataset's matching blueprint dataset. The implementation will just assume that it's automatically created by redap implementors (which is the case, and is tested now). So again, blueprint dataset are increasingly a very discrete implementation detail, so I'm not too worried about our ability to swap the implementation entirely.

crates/store/re_protos/proto/rerun/v1alpha1/cloud.proto
```diff
@@ -528,9 +521,14 @@ message BTreeIndexQuery {
...
-  // all partitions will be queried.
-  repeated rerun.common.v1alpha1.PartitionId partition_ids = 2;
+  // Client can specify what segments are queried. If left unspecified (empty list),
+  // all segments will be queried.
+  repeated rerun.common.v1alpha1.SegmentId segment_ids = 11;
+
+  // DEPRECATED: Use segment_ids instead.
+  // Client can specify what segments are queried. If left unspecified (empty list),
+  // all segments will be queried.
+  repeated rerun.common.v1alpha1.PartitionId partition_ids = 2 [deprecated = true];
```
teh-cmc: Why are there both `segment_ids` and `partition_ids`? I thought this whole PR was a hard breaking change that required updating clients and servers in lockstep anyhow, so why would it need any kind of deprecation?

crates/utils/re_arrow_util/src/batches.rs
```diff
@@ -253,6 +256,29 @@ impl RecordBatchExt for RecordBatch {
...
+    fn rename_columns(self, renames: &[(&str, &str)]) -> arrow::error::Result<RecordBatch> {
+        let (schema_ref, columns, row_count) = self.into_parts();
+        let Schema { fields, metadata } = Arc::unwrap_or_clone(schema_ref);
+
+        let new_fields: Vec<_> = fields
+            .iter()
+            .map(|f| {
+                for (original_name, new_name) in renames {
+                    if f.name() == *original_name {
+                        return Arc::new(f.as_ref().clone().with_name(*new_name));
```
teh-cmc, line 269: See `Arc::unwrap_or_clone`

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -513,7 +513,7 @@ def segment_table(
     ) -> datafusion.DataFrame:
         # Get the partition table from the inner object
 
-        partitions = self._inner.partition_table().df().with_column_renamed("rerun_partition_id", "rerun_segment_id")
+        partitions = self._inner.partition_table().df()
```
teh-cmc, line 516: segments?

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -616,12 +616,7 @@ def reader(
                 else:
                     index_values = np.array([], dtype=np.datetime64)
 
-                other_df = (
-                    view.filter_partition_id(segment)
-                    .using_index_values(index_values)
-                    .df()
-                    .with_column_renamed("rerun_partition_id", "rerun_segment_id")
-                )
+                other_df = view.filter_partition_id(segment).using_index_values(index_values).df()
```
teh-cmc, line 619: filter_segment_id?

teh-cmc (approved):
> NOTE: this is a major breaking change at the redap layer.

I'm confused: the description says that this is intended to be a breaking change, presumably requiring to update clients and servers in lockstep, but the implementation seems riddled with backward compatibility codepaths? So which is it?

crates/store/re_protos/proto/rerun/v1alpha1/common.proto
```diff
@@ -188,10 +188,18 @@ message ScanParametersOrderClause {
...
 
-// Unique identifier for a partition. Can be user defined
+// Unique identifier for a segment. Can be user defined
 // which means it can be of any type. For simplicity we start
 // with a string, but we will probably revisit this.
+message SegmentId {
+  optional string id = 1;
+}
+
+// DEPRECATED: Use SegmentId instead.
```
teh-cmc: This is the exact same type as `SegmentId` and Protobuf ignores names during transport, so you probably don't need this.

abey79:
> > NOTE: this is a major breaking change at the redap layer.
> 
> I'm confused: the description says that this is intended to be a breaking change, presumably requiring to update clients and servers in lockstep, but the implementation seems riddled with backward compatibility codepaths? So which is it?

Right. There use to be many more (attempts at) clean deprecation support. I'm gonna do another pass.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: I'm keeping these API changes for a future PR, to minimise the amount of diff.

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
abey79: Same, future PR.

crates/utils/re_arrow_util/src/batches.rs
abey79: Nah, it doesn't work here. The reason is that `Fields = Arc<[Arc<Field>]`, so you can only get `&Arc<Field>`, so you never have a chance to unwrap.  I haven't found a way yet to "unpack-and-repack" `Fields` without `Arc::clone`ing.

PR #12021, opened 2025-11-28
Make `InstancePose3D` consistently interact with transform frames

Wumpf:
### Related

* Fixes RR-2627

### What

`InstancePoses3D` is now again tracked on a per entity basis rather than being tied to a specific frame. This allows you to have many different poses that are all relative to the same frame.

Example for illustrating purposes: (in a follow-up PR this example will be made available x-sdk)

```py
"""Shows using InstancePoses3D with explicit coordinate frames and mesh instancing."""

import math
import rerun as rr

rr.init("rerun_example_instance_poses3d_frames", spawn=True)

# Animate a coordinate frame called "world_frame" moving in a circle around the origin.
for i in range(100):
    rr.set_time("frame", sequence=i)
    rr.log(
        "tf",
        rr.Transform3D(
            translation=[[math.cos(i * 0.1) * 2.0, math.sin(i * 0.1) * 2.0, 0]],
            parent_frame="tf#/",  # Connect to the default root frame.
            child_frame="world_frame",
        ),
    )

# Add content to visualize.
rr.log(
    "mesh",
    rr.Mesh3D(
        vertex_positions=[[1, 1, 1], [-1, -1, 1], [-1, 1, -1], [1, -1, -1]],
        triangle_indices=[[0, 2, 1], [0, 3, 1], [0, 3, 2], [1, 3, 2]],
        vertex_colors=[[255, 0, 0], [0, 255, 0], [0, 0, 255], [255, 255, 0]],
    ),
    rr.CoordinateFrame("world_frame"),
    static=True,
)
rr.log(
    "some_markers",
    rr.TransformAxes3D(axis_length=0.5, show_frame=True),
    rr.CoordinateFrame("world_frame"),
    static=True,
)

# Add static poses for markers. These are relative to the "world_frame" defined above.
rr.log(
    "some_markers",
    rr.InstancePoses3D(
        translations=[[1, 0, 1], [0, 1, 1], [0, -1, 1], [-1, 0, 1], [1, 1, 1], [-1, -1, 1], [1, -1, 1], [-1, 1, 1]]
    ),
    static=True,
)

# Add animated poses for the meshes. These are relative to the "world_frame" defined above.
for i in range(100):
    rr.set_time("frame", sequence=i)
    scale = math.sin(i * 0.2) * 0.3 + 1.0
    rr.log(
        "mesh",
        rr.InstancePoses3D(translations=[[3, 0, 0], [0, 3, 0], [0, -3, 0], [-3, 0, 0]], scales=[[scale, scale, scale]]),
    )
```
image

Under the hood this means that we have to separate out tracking of poses to be entirely entity driven and combine them very late in the process: in fact the `re_tf::TransformForest` no longer has any knowledge of these non-tree pose transforms. Instead, the per-view `TransformTreeContext` is what picks them up and builds the `TransformInfo` object that is used by visualizers - note that this PR simply moves out `TransformInfo` from `re_tf`, leaving behind a much simpler `TreeTransform` construct!


crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -140,15 +219,40 @@ impl ViewContextSystem for TransformTreeContext {
...
+                                                    latest_at_query,
+                                                )
+                                            })
+                                            .unwrap_or_default();
+
+                                        TransformInfo::new(tree_transform, poses)
+                                    })
+                                    .map_err(|err| err.clone());
+
+                                (*entity_path_hash, transform_info)
```
grtlr: That's some heavy indentation, could we factor this out into 1-2 smaller methods?

crates/viewer/re_view_spatial/tests/transform_clamping.rs
grtlr, line 1: A bit unrelated, but: Would it make sense to set the `SpatialInformation::show_axes` view property for these tests?

Wumpf: sure why not. But as pointed out before I'd rather keep those tests as isolated as possible. So a new `show_axes` property should get its own test suite

PR #12022, opened 2025-11-28
Make transform axes always show on pinholes

Wumpf:
I believe this is a regression fix. Either way useful, since even in the absence of extrinsics _on_ a pinhole transform, we want to be able to show transform axes (and will do so by default since `choose_default_visualizer` method dictates this)


IsseW (approved):
Makes sense 👍

PR #12026, opened 2025-12-01
add tracing instrumentation to the pyo3 layer through ContextVars

zehiko:
### What

As part of benchmark testing, we'd like to be able to have a single e2e trace for each benchmark test. In order to do so, we need to pass trace context across pyo3 boundary and it seems (per Gemini) that the recommended way to do this is through ``ContextVar``s which is what standard tracing libraries in python do. 

This PR leverages that and adds a context extraction logic as well as an API (per @abey79 suggestion) to the rerun sdk so that we can easily access this trace context var and set it on the python side.

For now I've focused on the query paths and we can see both query and fetching phase under a single trace thanks to changes in the dataframe table provider and trace propagation between query execution phases.  Follow up will add instrumentation to the register and index creation paths. 

Example usage:
```
        from rerun.catalog import _rerun_trace_context
        from opentelemetry import context
        from opentelemetry.propagate import get_global_textmap

        trace_ctx = _rerun_trace_context()

        with tracer.start_as_current_span("sample_index_values") as span:
            # Add attributes to help identify this benchmark
            span.set_attribute("benchmark.name", "sample_index_values")
            span.set_attribute("dataset.name", self.dataset_name)
            span.set_attribute("sample_count", 6)

            wrist = self.dataset.dataframe_query_view(
                index="log_tick",
                contents="/camera/wrist/embedding /thumbnail/camera/wrist",
            )

            sampled_times = [0, 100, 200, 500, 1000, 2000]

            current_ctx = context.get_current()
            carrier = {}
            get_global_textmap().inject(carrier, current_ctx)

            if carrier:
                token = trace_ctx.set(carrier)

            result = (
                (wrist.filter_index_values(sampled_times).fill_latest_at()).df().drop("rerun_partition_id").collect()
            )

            if carrier:
                trace_ctx.reset(token)

            span.set_attribute("result_rows", len(result))
```

Screenshot 2025-12-01 at 10 38 18

### DIFF
--- a/crates/store/re_redap_tests/src/tests/indexes.rs
+++ b/crates/store/re_redap_tests/src/tests/indexes.rs
@@ -258,8 +258,7 @@ pub async fn column_doesnt_exist(service: impl RerunCloudService) {
             .unwrap_err()
             .code();
 
-        // TODO(RR-2779): OSS returns NotFound.
-        // This is more precise and Rerun Cloud should be updated to return it.
+        // TODO(RR-3100)
         assert!(code == tonic::Code::InvalidArgument || code == tonic::Code::NotFound);
     }
 
@@ -288,7 +287,8 @@ pub async fn column_doesnt_exist(service: impl RerunCloudService) {
             .unwrap_err()
             .code();
 
-        assert_eq!(code, tonic::Code::NotFound);
+        // TODO(RR-3100)
+        assert!(code == tonic::Code::InvalidArgument || code == tonic::Code::NotFound);
     }
 }
 
--- a/rerun_py/tests/e2e_redap_tests/test_dataset_query.py
+++ b/rerun_py/tests/e2e_redap_tests/test_dataset_query.py
@@ -133,7 +133,7 @@ def test_query_view_from_schema(readonly_test_dataset: DatasetEntry) -> None:
             assert contents.count() > 0
 
 
-def readonly_test_dataset_schema_comparison_self_consistent(readonly_test_dataset: DatasetEntry) -> None:
+def test_readonly_dataset_schema_comparison_self_consistent(readonly_test_dataset: DatasetEntry) -> None:
     schema_0 = readonly_test_dataset.schema()
     schema_1 = readonly_test_dataset.schema()
     set_diff = set(schema_0).symmetric_difference(schema_1)


rerun_py/src/catalog/dataframe_query.rs
```diff
@@ -292,11 +313,12 @@ impl PyDataframeQueryView {
         py: Python<'_>,
         values: crate::dataframe::IndexValuesLike<'_>,
     ) -> PyResult<Self> {
-        let values = values.to_index_values()?;
-
-        Ok(self.clone_with_new_query(py, |query_expression| {
-            query_expression.filtered_index_values = Some(values);
-        }))
```
abey79, line 299: what is the purpose of this trace? This is literally a builder method.

rerun_py/src/catalog/dataframe_query.rs
```diff
@@ -367,8 +389,10 @@ impl PyDataframeQueryView {
     ///
     ///     The original view will not be modified.
     fn fill_latest_at(&self, py: Python<'_>) -> Self {
-        self.clone_with_new_query(py, |query_expression| {
-            query_expression.sparse_fill_strategy = SparseFillStrategy::LatestAtGlobal;
```
abey79, line 371: same

rerun_py/src/catalog/dataframe_query.rs
zehiko: ah this was me just testing and trying to get anything to work and didn't yet remove it. If we're ok with the approach, I'll clean it up and only at the few critical places for the main paths.

crates/utils/re_perf_telemetry/src/lib.rs
```diff
@@ -192,6 +192,69 @@ impl From<&TraceHeaders> for opentelemetry::Context {
...
+///
+/// This function reads trace context headers stored in a Python ContextVar named "TRACE_CONTEXT"
+/// from the standardized "tracecontext" module. This enables distributed tracing across Python→Rust
+/// boundaries in PyO3 applications.
+///
+/// The Python application should create a `tracecontext.py` module containing:
+/// ```python
+/// import contextvars
+/// TRACE_CONTEXT = contextvars.ContextVar("trace_context")
+/// ```
```
abey79: TBH, TIL  `contextvars` and I know nothing of opentelemetry, but I have a hard time comprehending this. What is this `tracecontext.py` file? What "python application" should create it? What if it's not created? And which the below function doesn't just call `contextvars.ContextVar("trace_context")` directly? This all feels very strange to me.

rerun_py/src/catalog/dataframe_query.rs
```diff
@@ -17,7 +17,28 @@ use tracing::instrument;
 use re_chunk::ComponentIdentifier;
 use re_chunk_store::{QueryExpression, SparseFillStrategy, ViewContentsSelector};
 use re_datafusion::DataframeQueryTableProvider;
+
```
abey79:
```suggestion
```

rerun_py/src/catalog/dataframe_query.rs
```diff
@@ -406,17 +430,18 @@ impl PyDataframeQueryView {
...
 
-        let df = ctx.call_method1("read_table", (self_,))?;
+            drop(client);
+            drop(dataset);
 
-        Ok(df)
+            let df = ctx.call_method1("read_table", (self_,))?;
+            Ok(df)
+        })
     }
```
abey79: this is where it becomes more interesting, but the returned dataframe is still mostly lazy. Stuff mostly happens when the dataframe is interacted with via datafusion, which then calls into our `DataframeQueryTableProvider`, which should probably be instrumented.

abey79 (commented):
As I note below, I'm left quite perplex by some of these changes. I'm not familiar with the general area, which doesn't help, but some of that stuff is outright weird. If its actually the correct thing to do, it would require more explaining imo.

crates/utils/re_perf_telemetry/src/lib.rs
zehiko: from little exploration I've done it seems that without changing the rerun sdk APIs, the only non-invasive and pretty standard way of doing this is ``contextvars``. You know I have almost no python experience and this is a result of Gemini-aided approach that suggested `contextvars` which seems to be safe to use in async and concurrent context.  It is also leveraged by `opentelemetry` for this exact use case which made me think it's right approach. 

Now, why ``tracecontext.py``? Well we need to grab the context from some standard place and I could just hardcode this to "benchmark_tests" or whatever the module is that I saw this morning, but thought that being able to just add ``tracecontext.py`` to your python app (in our case I've added this file with the relevant contextvars logic in component-tests) might be simple solution (which for now stays "hidden" and not advertised to the outside world).  I'm happy to make this anything that sounds reasonable, or simply document more clearly, but overall I don't see this working without `contextvars`.

rerun_py/src/catalog/dataframe_query.rs
zehiko: (yes, as mentioned in comment beforre, I'd need to clean up and only add the tracing to the relevant paths).

crates/utils/re_perf_telemetry/src/lib.rs
abey79: I'm fine with the `contextvars` stuff. It's really this `tracecontext.py`. It really should be just a constant in `rerun_bindings`, which is exposed across the ffi if needed by the python side. Call it `_RERUN_OPENTELEMENTRY_CONTEXT = "trace_context"` with the `_` prefix to drive home that it's private (it will already be without, given the entirety of `rerun_bindings` is private).

I just fail to see why it makes sense to ask some bits of the code to create a file that other piece might not have. This feels fragile and I can't see what would justify it.

zehiko: I totally agree, will make that change.

zehiko:
had a sync with @abey79 yesterday and he had some great suggestions to improve this approach:
* we will indeed use contextvars, but instead of arbitrary placing, we will provide a (private for now) API in the rerun sdk for accessing the trace context contextvars
* implementation will be in `re_perf_telemetry` and its ``trace_context_vars`` (or whatever we name it) will be used in the ``rerun_bindings``, as well as in the same crate when extracting the trace context
* this way consumer doesn't need to know or worry about the python module where contextvars are in

rerun_py/src/catalog/dataframe_query.rs
zehiko: removed. only tracing ``df()`` now and with the relevant changes in the table provider, we now have e2e query execution trace.

crates/utils/re_perf_telemetry/src/lib.rs
zehiko: updated the approach as per yesterday's sync

crates/utils/re_perf_telemetry/src/lib.rs
```diff
@@ -192,6 +192,82 @@ impl From<&TraceHeaders> for opentelemetry::Context {
...
+/// Get the trace context `ContextVar` object.
+///
+/// This returns the same Python `ContextVar` instance every time, ensuring that
+/// values set on it can be read back later. It is up to the caller to ensure trace context
+/// is reset and cleared as needed.
+pub fn get_trace_context_var(py: pyo3::Python<'_>) -> pyo3::PyResult<pyo3::Bound<'_, pyo3::PyAny>> {
+    use pyo3::prelude::*;
+
+    static CONTEXT_VAR: parking_lot::Mutex<Option<pyo3::Py<pyo3::PyAny>>> =
+        parking_lot::Mutex::new(None);
```
abey79, line 208: I was expecting to see a `std::sync::OnceLock` here. Any reason not to?

crates/utils/re_perf_telemetry/src/lib.rs
```diff
@@ -192,6 +192,82 @@ impl From<&TraceHeaders> for opentelemetry::Context {
...
+    Ok(trace_ctx_var)
+}
+
+#[cfg(feature = "pyo3")]
+/// Extract trace context from Python `ContextVar` for cross-boundary propagation.
+pub fn extract_trace_context_from_contextvar(py: pyo3::Python<'_>) -> TraceHeaders {
+    use pyo3::prelude::*;
+    use pyo3::types::PyDict;
+
+    let result = (|| -> PyResult<TraceHeaders> {
```
abey79: is that intel-called closure shenanigan to use `?`?  I'd rather have a nested function here, feels less weird.

abey79 (approved):
couple of nits but otherwise lgtm

crates/utils/re_perf_telemetry/src/lib.rs
zehiko: ah, I actually wanted to comment this - because ``OnceLock``'s init only allows non-fallible function (the ``get_or_try_init`` is only available in nightly), and going for ``expect()`` i.e. panicking was a big no-no, hence the ``Mutex``.

crates/utils/re_perf_telemetry/src/lib.rs
zehiko: yeah, fair point, looks better with a nested function.

PR #12027, opened 2025-12-01
Fix transform3d_hierarchy_frames snippet error

IsseW:
### What

Fixes a visualizer error that was happening on the `transform3d_hierarchy_frames` snippet because `sun_frame` wasn't logged as a child frame.


IsseW:
> oh, if we got an error for that lacking then we have a bug in transform handling actually: this should be fine without this extra entity since the only thing this does is to create an identity relationship between tf#/sun_transform and sun_frame which shouldn't be necessary. Or at least I don't see why we'd do that.
Great though that this surfaces so easily now :)

Ahh thought it could be something like that. Will create an issue with some more context then.

Wumpf:
perfect, thanks!

PR #12031, opened 2025-12-01
Add `re_arrow_combinators::reshape::RowMajorToColumnMajor`

MichaelGrupp:
Required for example to read matrices from variable sized, row-major protobuf lists, e.g. the 3x3 or 3x4 camera matrices that are found in `foxglove.CameraCalibration` schema.

In order to do that, we need both a cast to fixed-size-list and a reshaping transformation.

---

This is a prerequisite for #12035.

Contributes to:
https://linear.app/rerun/issue/RR-3045
https://linear.app/rerun/issue/RR-2327

### DIFF
--- a/crates/store/re_arrow_combinators/src/cast.rs
+++ b/crates/store/re_arrow_combinators/src/cast.rs
@@ -1,7 +1,10 @@
 //! Transforms that cast arrays to different types.
 
+use std::sync::Arc;
+
 use arrow::array::{Array, ArrayRef, ArrowPrimitiveType, PrimitiveArray};
 use arrow::compute::cast;
+use arrow::datatypes::Field;
 
 use crate::{Error, Transform};
 
@@ -121,13 +124,11 @@ impl Transform for ListToFixedSizeList {
             }
         }
 
-        // We know that `source` is a `ListArray` by it's type. But Arrow won't expose its field directly.
-        let field = match source.data_type() {
-            arrow::datatypes::DataType::List(f) => f.clone(),
-            _ => unreachable!(),
-        };
-
         // Build the FixedSizeListArray.
+        let field = Arc::new(Field::new_list_field(
+            source.value_type().clone(),
+            source.is_nullable(),
+        ));
         Ok(arrow::array::FixedSizeListArray::try_new(
             field,
             self.value_length,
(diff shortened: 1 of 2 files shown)


crates/store/re_arrow_combinators/src/cast.rs
```diff
@@ -83,3 +83,56 @@ where
...
+                    });
+                }
+            }
+        }
+
+        // We know that `source` is a `ListArray` by it's type. But Arrow won't expose its field directly.
+        let field = match source.data_type() {
+            arrow::datatypes::DataType::List(f) => f.clone(),
+            _ => unreachable!(),
+        };
```
emilk: It does expose https://docs.rs/arrow/latest/arrow/array/struct.GenericListArray.html#method.value_type  and you can wrap that in a `Field` with https://docs.rs/arrow/latest/arrow/datatypes/struct.Field.html#method.new_list_field

emilk: Also, if you lack something in `arrow,` open an issue about it and
* Add it to https://github.com/rerun-io/opensource/issues/1
* Link to it with a `TODO(apache/arrow-rs#1233)`

crates/store/re_arrow_combinators/src/reshape.rs
```diff
@@ -383,3 +385,78 @@ impl Transform for Explode {
...
+            })
+            .collect();
+
+        // Reorder values into a new FixedSizeListArray.
+        // We explicitly allow `take` here because we care about nulls.
+        #[expect(clippy::disallowed_methods)]
+        let reordered_values = arrow::compute::take(source.values(), &indices_to_take, None)?;
+
+        let field = Arc::new(Field::new(
+            "item",
```
emilk: 🤷 

```suggestion
        let field = Arc::new(Field::new_list_field(
```

emilk (approved):
Nice!

crates/store/re_arrow_combinators/src/cast.rs
MichaelGrupp: Thanks for the hint! Looks indeed much better now. https://github.com/rerun-io/rerun/pull/12031/commits/8c69ab8a0b55387ee8804c9812460635ac9f9592

Don't think this needs more work on Arrow side.

PR #12035, opened 2025-12-01
Load pinhole in `mcap_protobuf` example

MichaelGrupp:
Part of https://linear.app/rerun/issue/RR-2327

Requires the combinators from: #12031 ✅


examples/rust/mcap_protobuf/src/main.rs
```diff
@@ -287,12 +309,46 @@ fn main() -> anyhow::Result<()> {
             })?
             .build();
 
+    // Simple pinhole camera calibration lens, setting `image_from_camera` from the `K` matrix.
+    // TODO(michael): set child_frame of Pinhole?
```
MichaelGrupp: Hardcode (just for the example)?

MichaelGrupp: I would leave this out for now. Adapted the TODO with better wording.

PR #12037, opened 2025-12-01
Add e2e test for creating a table that already exists

timsaucer:
If we attempt to create a table in the server that already exists, either by table name or location, we should fail.


rerun_py/tests/e2e_redap_tests/test_table_create.py
```diff
@@ -49,3 +49,29 @@ def test_create_table_from_dataset(prefilled_catalog: PrefilledCatalog, tmp_path
     for returned_field in returned_schema:
         original_field = original_schema.field(returned_field.name)
         assert returned_field.metadata == original_field.metadata
+
+
+def test_create_existing_table_fails(
```
abey79: iirc you're going to have to mark this one as `creates_table`, otherwise it'll break DPF ci.

rerun_py/tests/e2e_redap_tests/test_table_create.py
```diff
@@ -49,3 +49,29 @@ def test_create_table_from_dataset(prefilled_catalog: PrefilledCatalog, tmp_path
...
+    prefilled_catalog: PrefilledCatalog, entry_factory: EntryFactory, tmp_path: pathlib.Path
```
abey79: `prefilled_catalog` already contains an `EntryFactory`, so you can use `prefilled_catalog.factory` instead of having the explicit fixture. 


Also, not necessary for this PR, but `EntryFactory` could have introspection APIs to list entries created so far. Could be occasionally convenient, eg when you want to trigger a collision like in this test.

abey79 (approved):

rerun_py/tests/e2e_redap_tests/test_table_create.py
timsaucer: Really? It doesn't actually create any tables.

timsaucer:
@rerun-bot reality-sync

PR #12038, opened 2025-12-02
Implement sizebytes for things in the tf transform cache

IsseW:
### Related

* Closes RR-2517

### What

Implements size bytes for the transform cache so its memory usage shows up in the memory panel.

image


crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -424,6 +473,15 @@ pub struct ResolvedPinholeProjection {
...
 }
 
+impl SizeBytes for ResolvedPinholeProjection {
+    fn heap_size_bytes(&self) -> u64 {
+        self.parent.heap_size_bytes()
+            + self.image_from_camera.heap_size_bytes()
+            + self.resolution.heap_size_bytes()
+            + self.view_coordinates.heap_size_bytes()
+    }
+}
```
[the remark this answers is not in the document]
IsseW: N/A

crates/utils/re_byte_size/Cargo.toml
```diff
@@ -27,3 +27,4 @@ all-features = true
 arrow.workspace = true
 half.workspace = true
 smallvec.workspace = true
+glam.workspace = true
```
Wumpf: I think making this unconditional here means that the sdk can no longer opt out of glam 🤔 
https://docs.rs/rerun/0.27.2/rerun/#feature-flags

can you plz check and if so make it optional? I don't know whether `re_byte_size` leaks into the sdk for sure.
In the big picture of viewer/test/etc. crate having a feature flag on this crate is actually bad since it adds more crate variations to build. But I'd love our sdk crate to stay minimal if possible :)

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -61,6 +62,14 @@ impl Default for TransformResolutionCache {
     }
 }
 
+impl SizeBytes for TransformResolutionCache {
+    fn heap_size_bytes(&self) -> u64 {
```
Wumpf, line 65: I think all of these `heap_size_bytes` should do self deconstruct via `Self { .. } = self;` because I'm fairly sure that already my currently open PRs may add fields that are gonna get missed otherwise

crates/utils/re_byte_size/Cargo.toml
IsseW: Yeah it would have added this to the sdk, good point. Added it behind feature flag now.

Wumpf: thanks!

PR #12040, opened 2025-12-02
Add option to specify `target_frame` to `SpatialInformation`

grtlr:
### Related

* Closes RR-2876
* Closes RR-2907
* Follow up: RR-3076

### What

This PR adds `target_frame` view property to the `SpatialInformation` archetype.

image

To this this PR open the DNA example and enter `tf#/helix/structure` into the `target_frame` field of the view. Now only the beads should be animated when pressing play.

### DIFF
--- a/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
+++ b/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
@@ -6,7 +6,9 @@ use re_log_types::{EntityPath, EntityPathHash};
 use re_tf::{TransformFrameId, TransformFrameIdHash, TreeTransform};
 use re_types::components::ImagePlaneDistance;
 use re_types::{ArchetypeName, archetypes, blueprint};
-use re_view::{DataResultQuery as _, latest_at_with_blueprint_resolved_data};
+use re_view::{
+    DataResultQuery as _, HybridLatestAtResults, latest_at_with_blueprint_resolved_data,
+};
 use re_viewer_context::{
     IdentifiedViewSystem, TransformDatabaseStoreCache, ViewContext, ViewContextSystem,
     ViewContextSystemOncePerFrameResult, typed_fallback_for,


Wumpf:
screenshot plz :)

crates/store/re_types/definitions/rerun/blueprint/archetypes/spatial_information.fbs
```diff
@@ -4,6 +4,11 @@ namespace rerun.blueprint.archetypes;
 table SpatialInformation (
     "attr.rerun.scope": "blueprint"
 ) {
+  /// The target reference frame for all transformations.
+  ///
+  /// Defaults to the frame derived from the space origin.
```
Wumpf: nit: trying to be even more clear :)

```suggestion
  /// Defaults to the coordinate frame used by the space origin entity.
```

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -108,8 +107,35 @@ impl ViewContextSystem for TransformTreeContext {
...
+                        TransformFrameIdHash::from_str(target_frame.as_str())
+                    } else {
+                        self.transform_frame_id_for(query.space_origin.hash())
+                    }
+                }
+                Err(err) => {
+                    re_log::error_once!("Failed to query target frame: {err}");
+                    self.transform_frame_id_for(query.space_origin.hash())
+                }
+            }
```
Wumpf, line 233: Took me a bit that your fallback _tries_ to use a cached value and this is ONLY used when we didn't have a chance yet to do that caching. Should document that better, because my knee-jerk reaction was that the fallback is incomplete.

That said, I do think we can still have all the logic in the fallback and never have the fallback return empty string! We should be able to just do an adhoc query `latest_at_with_blueprint_resolved_data` on the origin entity

Also, status quo has very surprising behavior when intentionally setting an empty string (which ... is actually a valid name for a transform frame.. interesting 🤔 )

crates/viewer/re_view_spatial/src/shared_fallbacks.rs
```diff
@@ -123,4 +123,23 @@ pub fn register_fallbacks(system_registry: &mut re_viewer_context::ViewSystemReg
             components::ShowLabels(false.into())
         },
     );
+
+    // Target reference frame - fallback to the computed value from view state
+    //
+    // An empty string signals that nothing has been set yet.
```
Wumpf: see comment above, this is pretty ugly since we generally don't want the receiving end to be able to know that something is a fallback

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
Wumpf: we paired on this a bit and tried to make it work, but it turned out that had way more ripple effect than it's worth it!

crates/viewer/re_viewer/tests/snapshots/all_view_selection_uis/Dark/3D.png
Wumpf, line 1: this snapshot needs more space now

crates/viewer/re_view_bar_chart/src/view_class.rs
```diff
@@ -140,11 +140,11 @@ impl ViewClass for BarChartView {
         ctx: &ViewerContext<'_>,
         ui: &mut egui::Ui,
         state: &mut dyn ViewState,
-        _space_origin: &EntityPath,
+        space_origin: &EntityPath,
         view_id: ViewId,
```
Wumpf, line 144: ha, almost like this should be a view ctx.
(for another day ofc ;-))

PR #12050, opened 2025-12-02
Partition-to-segment rename (wave 2): `re_uri`

abey79:
### Related

### What

In this PR, we focus on `re_uri`:
- update everything to using "segment"
- remain backward compatible with url containing `?partition_id`

The rest of the diff is essentially dealing with consequences, with some minor adjacent renames.

### DIFF
--- a/crates/store/re_redap_tests/src/tests/indexes.rs
+++ b/crates/store/re_redap_tests/src/tests/indexes.rs
@@ -258,8 +258,7 @@ pub async fn column_doesnt_exist(service: impl RerunCloudService) {
             .unwrap_err()
             .code();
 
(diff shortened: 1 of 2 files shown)


crates/store/re_uri/src/endpoints/dataset.rs
```diff
@@ -87,7 +101,7 @@ impl DatasetPartitionUri {
         Ok(Self {
             origin,
             dataset_id,
-            partition_id,
+            segment_id,
             time_range,
             fragment,
         })
```
abey79, line 107: this is the only business logic part of this pr

emilk (approved):
Nice and simple

PR #12052, opened 2025-12-02
Refactor `re_smart_channel` -> `re_log_channel`

emilk:
* Part of https://linear.app/rerun/issue/RR-3003/rewrite-re-smart-channel-to-be-two-way

`re_smart_channel` is a monster that has evolved over time.

This takes a big step towards simplifying it

* Renamed `re_smart_channel` to `re_log_channel`
* Removes the generics - it always wraps a `DataSourceMessage` now
* Renamed `Sender<T>/Receiver<T>` to `LogSender/LogReceiver`
* Removed the latency metrics - we are no longer using it
* Use a much simpler design for the "ui waker" mechanism
* Combine `SmartMessageSource` and `SmartChannelSource` into a single `LogSource`

### DIFF
--- a/crates/store/re_log_channel/src/sender.rs
+++ b/crates/store/re_log_channel/src/sender.rs
@@ -82,6 +82,11 @@ impl LogSender {
     ///
     /// Using a [`LogSender`] after calling `quit` is undefined behavior: the receiving end is free
     /// to silently drop those messages (or worse).
+    ///
+    /// This is an _explicit_ quit.
+    /// Dropping all senders will also close the channel and notify the receiver,
+    /// but calling `quit` has the benefit of the receiver knowing
+    /// that we reached the end (or there was an error).
     pub fn quit(
         &self,
         err: Option<Box<dyn std::error::Error + Send>>,


crates/store/re_log_channel/src/lib.rs
```diff
@@ -268,33 +269,26 @@ impl std::fmt::Display for SmartMessageSource {
...
-/// Create a new channel using the same stats as some other.
-///
-/// This is a very leaky abstraction, and it would be nice to refactor some day
-pub(crate) fn smart_channel_with_stats<T: Send>(
-    sender_source: SmartMessageSource,
-    source: Arc<SmartChannelSource>,
-    stats: Arc<SharedStats>,
-) -> (Sender<T>, Receiver<T>) {
+    let channel = Arc::new(Channel::default());
     let (tx, rx) = crossbeam::channel::unbounded();
```
jprochazk, line 224: Any chance we can also make these bounded while we're at it, to add backpressure to ingestion?

emilk: I'm not gonna make that change in this PR, no. It's big enough, and its unrelated.

emilk: …but I will likely do something similar later as part of my ongoing work.

crates/store/re_log_channel/src/sender.rs
```diff
@@ -0,0 +1,114 @@
+use std::sync::Arc;
+
+use re_log_types::DataSourceMessage;
+
+use crate::{Channel, LogSource, SendError, SmartMessage, SmartMessagePayload};
+
+#[derive(Clone)]
+pub struct LogSender {
```
emilk, line 8: For some reason git didn't realize this is mostly a rename of the old sender file, but with generics removed

crates/store/re_log_channel/src/receiver.rs
```diff
@@ -0,0 +1,104 @@
+use std::sync::{
+    Arc,
+    atomic::{AtomicBool, Ordering::Relaxed},
+};
+
+use crate::{Channel, LogSource, SmartMessage, TryRecvError};
+
+pub struct LogReceiver {
```
emilk, line 8: This is mostly a rename of the old receiver file, but with generics removed

crates/store/re_log_channel/src/receiver.rs
```diff
@@ -0,0 +1,104 @@
...
+            connected: AtomicBool::new(true),
+        }
+    }
+
+    /// Call this on each sent message.
+    ///
+    /// Can be used to wake up the receiver thread.
+    pub fn set_waker(&self, waker: impl Fn() + Send + Sync + 'static) {
+        *self.channel.waker.write() = Some(Box::new(waker));
+    }
```
emilk, line 34: this is new though

crates/store/re_log_channel/src/sender.rs
```diff
@@ -0,0 +1,114 @@
...
+                payload: SmartMessagePayload::Quit(err),
+            })
+            .map_err(|SendError(msg)| SendError(Box::new(msg)))?;
+
+        if let Some(waker) = self.channel.waker.read().as_ref() {
+            (waker)();
+        }
+
+        Ok(())
+    }
```
lucasmerlin, line 106: Should dropping the Sender also call Quit?

crates/top/rerun/src/native_viewer.rs
```diff
@@ -7,6 +7,8 @@ use re_viewer::external::eframe;
 /// ⚠️  This function must be called from the main thread since some platforms require that
 /// their UI runs on the main thread! ⚠️
 pub fn show(main_thread_token: crate::MainThreadToken, msgs: Vec<LogMsg>) -> eframe::Result {
+    let log_source = re_log_channel::LogSource::Sdk; // Whatever. Nobody is really using this functions anyhow.
```
lucasmerlin, line 10: Then why not get rid of this? 🤔

lucasmerlin (approved):
Looks good!

crates/top/rerun/src/native_viewer.rs
emilk: It's part of our public API. We can mark it `#[deprecated]` but that feels out-of-bounds for this PR

crates/store/re_log_channel/src/sender.rs
emilk: No - the `Quit` shows intention. I'll add a comment to that effect.
Also the `LogSender` is clonable, so we really don't wanna quit on drop.

PR #12057, opened 2025-12-02
Make transform frame edit not experimental

IsseW:
### Related

* Closes RR-3044

### What

Improves the frame id edit in the selection panel, and makes it not experimental.

### Preview

When the frame hasn't been overridden the text is subdued.

Screenshot 2025-12-02 at 17 31 01

When starting to edit the frame it shows suggestions

Screenshot 2025-12-02 at 17 31 13

When editing the frame it keeps showing suggestion that match what you've written so far. The text is also red if the current value doesn't match an existing frame.

Screenshot 2025-12-02 at 17 31 28

When overridden it uses the default text color.

Screenshot 2025-12-02 at 17 31 48

### DIFF
--- a/crates/viewer/re_viewer/tests/snapshots/settings_screen.png
+++ b/crates/viewer/re_viewer/tests/snapshots/settings_screen.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:fbc37db5977d36c660a850abe78460e79a9441db2d2af92325d3a22679603f74
-size 105919
+oid sha256:6f6caa2f778d76f994fd9a6ac07ad4f1c5cfcc4d7b4e1c7373bcad927c1be60b
+size 99584


crates/viewer/re_selection_panel/src/selection_panel.rs
```diff
@@ -349,12 +349,7 @@ impl SelectionPanel {
                         // TODO(RR-2700): Come up with something non-experimental.
```
emilk: old TODO?

crates/viewer/re_selection_panel/src/selection_panel.rs
```diff
@@ -584,6 +607,85 @@ To learn more about coordinate frames, see the [Spaces & Transforms](https://rer
...
+    let response = ui.add(text_edit);
+
+    let mut suggestions = transform_cache
+        .frame_id_registry()
+        .iter_frame_ids()
+        // Only show named frames.
+        .filter(|(_, id)| !id.is_entity_path_derived())
+        .filter_map(|(_, id)| id.strip_prefix(&*frame_id))
+        .filter(|rest| !rest.is_empty())
+        .collect::<Vec<_>>();
```
grtlr: Consider scoping `transform_cache` and return `suggestions` and `text_edit`. We're sorting on the `suggestions` later, and shouldn't need to hold the lock for that.

crates/viewer/re_selection_panel/src/selection_panel.rs
```diff
@@ -349,12 +349,7 @@ impl SelectionPanel {
                         // TODO(RR-2700): Come up with something non-experimental.
                         let is_spatial_view =
                             view.class_identifier() == "3D" || view.class_identifier() == "2D";
-                        if ctx
-                            .global_context
-                            .app_options
-                            .experimental_coordinate_frame_display_and_override
-                            && is_spatial_view
-                        {
+                        if is_spatial_view {
```
grtlr, line 351: Small suggestion: @Wumpf and I [had the idea](https://linear.app/rerun/issue/RR-3044/make-coordinate-frame-ui-non-experimental#comment-47fc313b) of only showing the coordinate frame UI if explicit frames were logged.

What do you think about that?

IsseW: Yeah I do do that internally in the function, also documented it.

grtlr: Oh sorry, you're right.

PR #12058, opened 2025-12-02
Refactor auth state handling

lucasmerlin:
### What

This adds a way to add a callback to re_auth whenever auth state changes and uses that to add a AuthContext to the GlobalContext that is set when the user is logged in

### DIFF
--- a/crates/viewer/re_redap_browser/src/server_modal/login_flow/web.rs
+++ b/crates/viewer/re_redap_browser/src/server_modal/login_flow/web.rs
@@ -34,7 +34,6 @@ pub struct State {
     pkce: Rc<Pkce>,
     state: String,
 
-    login_hint: Option<String>,
     result: Rc<RefCell<Option<Result<Credentials, String>>>>,
 }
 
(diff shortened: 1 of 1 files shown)


crates/viewer/re_redap_browser/src/server_modal.rs
```diff
@@ -74,8 +62,8 @@ impl Authentication {
     }
 
     fn start_login_flow(&mut self, ui: &mut egui::Ui) {
-        let login_hint = self.email.as_deref();
-        match LoginFlow::open(ui, login_hint) {
+        // TODO: Is login hint required?
+        match LoginFlow::open(ui, None) {
```
lucasmerlin: Not sure how this should be handled. I assume the login hint will prefill the email in the browser is set? But now we should only have a email if the user is logged in, so this feels kind of useless. Did this do something before?

jprochazk: I think you're right that this is useless

jprochazk (approved):

Issue #12011, opened 2025-11-28
unit test of re_view_spatial fail at main branch (commit: f101cbd)


makeecat:
**Describe the bug**
cargo test of re_view_spatial fails for main branch at commit f101cbd

**To Reproduce**
```
cargo test -p re_view_spatial --all-targets --all-features
```

**Expected behavior**
The cargo test of `re_view_spatial` crate should pass

**Screenshots**
```
test test_spatial_view_2d_spawn_heuristics_like_detect_and_track_objects ... FAILED
test test_differing_image_sizes ... FAILED
test test_stacking_color_and_seg ... FAILED
test test_mixed_images ... FAILED
test test_not_stacking_color_images ... FAILED

failures:

---- test_spatial_view_2d_spawn_heuristics_like_detect_and_track_objects stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: crates/viewer/re_view_spatial/tests/snapshots/spawn_heuristics__detect_and_track_objects_like_scene_2d_view_heuristic.snap
Snapshot: detect_and_track_objects_like_scene_2d_view_heuristic
Source: crates/viewer/re_view_spatial/tests/spawn_heuristics.rs:113
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Expression: recommended_views
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────
    1     1 │ ViewSpawnHeuristics {
    2     2 │     recommended_views: [
    3     3 │         RecommendedView {
          4 │+            origin: /3D/camera,
          5 │+            query_filter: [
          6 │+                "+ $origin/**",
          7 │+            ],
          8 │+        },
          9 │+        RecommendedView {
    4    10 │             origin: /segmentation,
    5    11 │             query_filter: [
    6    12 │                 "+ $origin/**",
    7    13 │             ],
────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.

[...]

PR #12061, opened 2025-12-02
Fix tests failing with `cargo test` but not with `cargo nextest`

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/12011

### What

The difference between those two is whether each test gets its own process. So it was clear that we're leaking state. Turns out it's about store subscribers which often rely on unique store ids for distinguishing concrete stores.

Solving this on the store subscriber level with the current arch is hard because the subscriber correctly accumulate data from all parallel running tests that do writes do stores with the exact same id

### DIFF
--- a/crates/store/re_log_types/src/lib.rs
+++ b/crates/store/re_log_types/src/lib.rs
@@ -627,6 +627,7 @@ impl StoreInfo {
     /// Creates a new store info for testing purposes with a fixed store id.
     ///
     /// Most of the time we don't want to fix the store id since it is used as a key in static store subscribers, which might not get teared down after every test.
+    /// Use this only if the recording id may show up somewhere in the test output.
     pub fn testing_with_recording_id(recording_id: impl Into<RecordingId>) -> Self {
         // Do not use a version since it breaks snapshot tests on every update otherwise.
         Self::new_unversioned(


crates/store/re_log_types/src/lib.rs
```diff
@@ -617,9 +617,20 @@ impl StoreInfo {
...
         Self::new_unversioned(
-            StoreId::new(StoreKind::Recording, "test_app", "test_recording"),
+            StoreId::random(StoreKind::Recording, "test_app"),
+            StoreSource::Other("test".to_owned()),
+        )
+    }
+
+    /// Creates a new store info for testing purposes with a fixed store id.
+    ///
+    /// Most of the time we don't want to fix the store id since it is used as a key in static store subscribers.
```
grtlr:
```suggestion
    /// Most of the time we don't want to fix the store id since it is used as a key in static store subscribers, which might not get teared down after every test.
```

grtlr (approved):
Oh my, that looks hard to find!

Is there a rule of thumb when to use `StoreInfo::testing` and when to set the recording id explicitly?

Wumpf:
> Oh my, that looks hard to find!

it was surprisingly easy by just looking at the kind of test failures 🤷. that led directly to store subscribers and I unfortunately know how they're implemented....

Wumpf:
> Is there a rule of thumb when to use StoreInfo::testing and when to set the recording id explicitly?

use `testing` whenever possible. Use the other one if the id shows up for some reason.
I'll add a comment towards that.

PR #12064, opened 2025-12-03
Improve unknown/missing frame errors

IsseW:
### What

Adds overridden frame ids to lookup local to that view, so we can display better errors.

Before:
image

After:
Screenshot 2025-12-03 at 10 20 43

### DIFF
--- a/crates/viewer/re_selection_panel/src/selection_panel.rs
+++ b/crates/viewer/re_selection_panel/src/selection_panel.rs
@@ -607,7 +607,7 @@ fn frame_id_edit(
     frame_id: &mut String,
     frame_id_before: &String,
 ) {
-    let (frame_exists, mut suggestions) = {
+    let (mut suggestions, response) = {
         // In a scope to not hold the lock for longer than needed.
         let caches = ctx.viewer_ctx.store_context.caches;
         let transform_cache =
@@ -620,6 +620,12 @@ fn frame_id_edit(
             .lookup_frame_id(TransformFrameIdHash::from_str(&*frame_id))
             .is_some();
 
+        let mut text_edit = egui::TextEdit::singleline(frame_id).hint_text(frame_id_before);
+        if !frame_exists {
+            text_edit = text_edit.text_color(ui.tokens().error_fg_color);
+        }
+        let response = ui.add(text_edit);
+
         let suggestions = transform_cache
             .frame_id_registry()
             .iter_frame_ids()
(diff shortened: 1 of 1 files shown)


crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -111,6 +111,29 @@ impl ViewContextSystem for TransformTreeContext {
...
+            };
+
+            for frame in batch {
+                let frame_hash = TransformFrameIdHash::new(&frame);
+                if !self.frame_id_mapping.contains_key(&frame_hash) {
+                    // As overrides are local to this view we need to clone the whole map to add new hashes.
+                    Arc::make_mut(&mut self.frame_id_mapping).insert(frame_hash, frame);
+                }
+            }
+        }
```
Wumpf: in principle that's exactly what I had in mind, but we don't have to repeat the query here (which can get actually quite expensive!). We already have all the data in `self.entity_transform_id_mapping`

IsseW: `self.entity_transform_id_mapping` only stores `TransformFrameIdHash`. Do you mean to extend it to also store `TransformFrameId`?

Wumpf: oh, you're ofc right, we throw away the data right now!
Yeah I think we should either store that there or side-channel that out of the creation of `entity_transform_id_mapping` - the full query happens here alright after all https://github.com/rerun-io/rerun/blob/1f5f2c252fa0a81efe7bf6be8754a7ea7783c0c9/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs#L308

Wumpf: storing is probably fine, it's not that much data 🤷

IsseW: Collected results into a vec now

crates/viewer/re_selection_panel/src/selection_panel.rs
IsseW, line 1: Should probably have done this in another PR, but noticed here that a last minute change caused an off by 1 frame issue. And is fixed by simple re-ordering

Wumpf: k, fine by me 🙈 
thanks for pointing out what happened here

Wumpf (approved):
thanks for going the extra mile!

PR #12067, opened 2025-12-03
Simplify `Field` declarations in `re_arrow_combinators`

MichaelGrupp:
Arrow has `Field::new_list_field`, which needs one argument less for our use case:
https://docs.rs/arrow/latest/arrow/datatypes/struct.Field.html#method.new_list_field


MichaelGrupp:
Quick follow-up for the existing code in this crate based on suggestion by @emilk in #12031

Issue #12060, opened 2025-12-02
Moving camera behaves strange once spin speed != 0


roym899:
**Describe the bug**
Moving the camera seems to use an unexpected reference frame once the spin speed is not 0. This was not the case previously when spinning was toggled on.

**To Reproduce**
Activate spin speed; try to move the camera with right click + drag.

**Desktop (please complete the following information):**
 - OS: macOS Sequoia 15.7.2

**Rerun version**

> rerun-cli 0.28.0-alpha.1+dev (base lance map_view nasm native_viewer oss_server re_server release_full release_no_web_viewer web_viewer) [rustc 1.88.0 (6b00bc388 2025-06-23), LLVM 20.1.5] aarch64-apple-darwin main f101cbd, built 2025-11-28T03:42:26.691187Z

PR #12068, opened 2025-12-03
Do input from the perspective of spin

IsseW:
### Related

* Closes https://github.com/rerun-io/rerun/issues/12060

### What

Make eye inputs take eye spin into account.

### DIFF
--- a/crates/viewer/re_view_spatial/src/eye.rs
+++ b/crates/viewer/re_view_spatial/src/eye.rs
@@ -680,7 +680,7 @@ impl EyeState {
             }
         }
 
-        // Handle spinning after saving to blueprint to not continuously write to the blueprint.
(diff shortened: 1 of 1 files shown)


crates/viewer/re_view_spatial/src/eye.rs
```diff
@@ -888,11 +889,23 @@ impl EyeState {
...
+            apply_spin(*spin);
 
             // Request repaint if we're spinning.
             ctx.egui_ctx().request_repaint();
-        } else {
-            self.spin = None;
+        }
+        // If we just stopped spinning write new position to the blueprint.
+        else if let Some(spin) = self.spin.take() {
+            apply_spin(spin);
```
Wumpf, line 907: ah that's clever, very nice.

PR #12078, opened 2025-12-03
Logs organization of authenticated users

aedm:
### Related

Closes RR-2986

### What

- Adds the organization ID to post-login analytics event.
- Some ergonomics around the analytics framework.


crates/utils/re_analytics/README.md
```diff
@@ -18,6 +18,10 @@ Find out its location by running `rerun analytics config`.
 ### Opting out
 Run `rerun analytics disable` to opt out of all usage data collection.
 
+Alternatively, you can control analytics using an environment variable:
+- `RERUN_ANALYTICS=0`: off
+- `RERUN_ANALYTICS=force`: on (overrides configuration!)
```
Wumpf: Would it make sense to call it `FORCE_RERUN_ANALYTICS` since it always overrides? We then could use the `env_var_bool` to be consistent with other env vars we have
(`env_var_bool` would need to be made public and should probably live next to the already public `env_var_is_truthy`)

aedm: The idea was to be able to turn it off using the same env var. But `env_var_is_truthy` is nice, I'll use that for `FORCE_RERUN_ANALYTICS`.

aedm: > use the `env_var_bool` to be consistent with other env vars we have 

Fun fact: we only `env_var_bool` for a single env var. :)

Issue #6356, opened 2024-05-16
[Windows] Killing python script spawning Rerun Viewer also closes Rerun


Wumpf:
We recently explicitly changed this behavior on Mac & Linux to always detach so that if you close e.g. face_tracking, the viewer stays open. However, on Windows this is not the case and closing the face_tracking demo also closes the Viewer

Likely related to
* #3992

Wumpf:
Ran into this just now, still broken

PR #12079, opened 2025-12-03
Gracefully exit on `ctrl c`

lucasmerlin:
### Related

* closes https://linear.app/rerun/issue/RR-3060/closing-the-viewer-seems-to-not-always-save-list-of-servers

### What
Adds a ctrl c handler so that we can gracefully exit when closing rerun with ctrl c. This ensures the state is saved.

Testing with Rerun started by....
* [x] cli standalone
    * [x] Windows
    * [x] Linux
    * [x] Mac
* [x] via python package
    * [x] Windows
    * [x] Linux (if `python -m rerun` is correct here)
    * [x] Mac
* [x] via `spawn`
    * [x] Windows - still has https://github.com/rerun-io/rerun/issues/6356 
    * [x] Linux - behaves correctly and does not close the viewer
    * [x] Mac

### DIFF
--- a/crates/top/rerun/src/commands/entrypoint.rs
+++ b/crates/top/rerun/src/commands/entrypoint.rs
@@ -18,7 +18,6 @@ use crate::commands::AnalyticsCommands;
 #[cfg(feature = "data_loaders")]
 use crate::commands::McapCommands;
 use crate::commands::RrdCommands;
-use crate::external::re_ui::{UICommand, UICommandSender as _};
 
 // ---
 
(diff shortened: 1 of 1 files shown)


crates/top/rerun/src/commands/entrypoint.rs
```diff
@@ -1,6 +1,7 @@
 use std::net::IpAddr;
 use std::time::Duration;
 
+use crate::external::re_ui::{UICommand, UICommandSender};
```
emilk: Move down!

https://github.com/rerun-io/rerun/blob/main/CODE_STYLE.md#style

Wumpf:
added some testing check boxes because this sort of thing historically caused a loooot of grief in the past
Can help with windows testing later tonight

PR #12080, opened 2025-12-03
Fix missing rrd for tests

Wumpf:


IsseW:
Oh no, now I'll have to put all my rrd's somewhere else 😶

abey79:
> Oh no, now I'll have to put all my rrd's somewhere else 😶

fwiw, I have a `my_stuff` directory where I dump python scripts, etc. and it's gitignored in `~/.gitignore`. So I can apply that pattern in any repo checkout.

PR #12083, opened 2025-12-03
Add `MapProvider::MapboxLight` 

sectore:
### What

Adds support of [`Mapbox Light` maps](https://www.mapbox.com/maps/light).

It goes hand in hand with Rerun `light` theme.

### Example

(example is running with [fit-activities-rerun](https://github.com/sectore/fit-activities-rerun) pointing to this PR changes)

mb-light-example

### DIFF
--- a/crates/store/re_types/src/blueprint/components/map_provider.rs
+++ b/crates/store/re_types/src/blueprint/components/map_provider.rs
@@ -148,8 +148,8 @@ impl ::re_types_core::reflection::Enum for MapProvider {
             Self::OpenStreetMap,
             Self::MapboxStreets,
             Self::MapboxDark,
-            Self::MapboxLight,
             Self::MapboxSatellite,
+            Self::MapboxLight,
         ]
     }
 


MichaelGrupp (changes requested):
Hi, thanks for the PR! Adding this map provider makes sense to me. One remark regarding the code:

It looks like you manually edited the Rust/C++/Python code. We normally use automatic code generation for this type of updates, based on language-agnostic Flatbuffer definitions in the `re_types` crate.

In this specific case, you would need to add the new variant to [map_provider.fbs](https://github.com/rerun-io/rerun/blob/main/crates/store/re_types/definitions/rerun/blueprint/components/map_provider.fbs) and then run the codegen (`pixi run codegen`, or `cargo --quiet run --package re_types_builder`). This should generate almost the same output as your code, but will ensure that the changes are kept once this is merged 🙂

Wumpf:
> run codegen (no changes detected)

That's strange, and CI has a different opinion about it 😄 
`pixi run codegen` tries to do nothing if it thinks no files have changed, but the logic there isn't perfect especially when generated files were handedited before. Try running `pixi run codegen --force`

PR #12084, opened 2025-12-03
Add minimal support for variable-length stream footers

teh-cmc:
LLM summary:
> This PR adds minimal support for variable-length stream footers in the RRD encoding format. The main purpose is to prepare the codebase for future support of incremental RRD manifests by allowing stream footers to point to multiple RRD footers, while maintaining backward compatibility with the current single-footer approach.
> 
> **Key changes:**
> - Refactored `StreamFooter` to contain a vector of `StreamFooterEntry` objects instead of directly storing footer metadata
> - Introduced `StreamFooterEntry` struct to represent individual RRD footer references
> - Updated encoding/decoding logic to handle variable-length footers while maintaining the current constraint of exactly one entry

I have nothing to add.

---

It should work no problem but I'm gonna slap a `do-not-merge` on it until we can run `full-check` again, just in case.


teh-cmc:
@rerun-bot full-check

crates/store/re_log_encoding/src/rrd/frames.rs
```diff
@@ -276,74 +304,97 @@ impl StreamFooter {
...
-        out.extend_from_slice(&crc_excluding_header.to_le_bytes());
+        out.extend_from_slice(&num_rrd_footers.to_le_bytes());
 
         let n = out.len() as u64 - before;
-        assert_eq!(Self::ENCODED_SIZE_BYTES as u64, n);
+        assert_eq!(
+            Self::ENCODED_SIZE_BYTES as u64,
+            n,
+            "Stream footers always point to a single RRD footer at the moment"
+        );
```
Wumpf, line 377: So this means that today's builds (including release builds. the other checks on this are `debug_assert`) will break with future rrd footers without those rrd footers changing version. That seems wrong to me 🤔
Them not handling "all features" of those rrd footers is one thing, but I wouldn't expect encoder/decoder to outright crash.

teh-cmc: That's the encoding path, this assertion will always pass.

crates/store/re_log_encoding/src/rrd/frames.rs
```diff
@@ -361,17 +412,53 @@ impl Decodable for StreamFooter {
...
-        let crc = u32::from_le_bytes(data[24..28].try_into().expect("cannot fail, checked above"));
+        let num_rrd_footers = u32::from_le_bytes(
+            fixed_data[8..12]
+                .try_into()
+                .expect("cannot fail, checked above"),
+        );
+        debug_assert_eq!(
+            1, num_rrd_footers,
+            "Stream footers always point to a single RRD footer at the moment"
+        );
```
teh-cmc: > The variable length support as such makes sense to me, but inducing panics even in optimized builds on encode/decode seems weird/wrong/unecessary/dangerous to me - the version doesn't change but we still can't even read those footers at all then.

I don't think there's any realistic scenario where we will ever try to open a recording from that far into the future with today's Rerun, but I can remove the debug assertion, sure
```suggestion
```

PR #12085, opened 2025-12-03
Partition-to-segment rename (wave 4): rename everything else

abey79:
### Related

### What

This should be the final wave of renames. There are a couple of minor things that are pointed out as comments. The rest should be pure codebase rename.

### DIFF
--- a/rerun_py/rerun_bindings/rerun_bindings.pyi
+++ b/rerun_py/rerun_bindings/rerun_bindings.pyi
@@ -520,7 +520,7 @@ class Recording:
         monotonically increasing when data is sent from a single process.
 
         If `None` is passed as the index, the view will contain only static columns (among those
-        specified) and no index columns. It will also contain a single row per partition.
+        specified) and no index columns. It will also contain a single row per segment.
 
         Parameters
         ----------


docs/content/reference/migration/migration-0-28.md
```diff
@@ -90,3 +90,17 @@ The DataFusion utility functions in `rerun.utilities.datafusion.functions.url_ge
...
+# Before (0.27)
+def on_event(event):
+    print(event.partition_id)
+
+# After (0.28)
+def on_event(event):
+    print(event.segment_id)
+```
+
+This affects `PlayEvent`, `PauseEvent`, `TimeUpdateEvent`, `TimelineChangeEvent`, `SelectionChangeEvent`, and `RecordingOpenEvent`.
```
abey79, line 118: small breaking change here...

crates/store/re_sorbet/src/sorbet_schema.rs
```diff
@@ -166,7 +166,11 @@ impl SorbetSchema {
             None
         };
 
-        let partition_id = metadata.get("rerun:partition_id").map(|s| s.to_owned());
+        // Support both new "rerun:segment_id" and legacy "rerun:partition_id" keys
+        let segment_id = metadata
+            .get("rerun:segment_id")
+            .or_else(|| metadata.get("rerun:partition_id"))
+            .map(|s| s.to_owned());
```
abey79, line 173: sorbet metadata backward compatibility

ntjohnson1: NIT: Do we have a test on this? I mean it looks reasonable but might be nice to check?

ntjohnson1:
I'll bring it up at standup but if this is the last of it do we think it makes sense to cut an alpha off of this so we can advertise some of the small breaking changes/api updates ahead of time in case we missed anything?

crates/store/re_sorbet/src/sorbet_schema.rs
abey79: good call, Claude obliged.

PR #12087, opened 2025-12-03
Add exclusions for 2d visualizables in 3d and vice-versa

IsseW:
### Related

* Part of RR-3067

### What

To avoid errors in the view, automatically exclude entities from the view.

Example of an integration test that failed earlier:
Screenshot 2025-12-03 at 18 46 45


IsseW:
One issue this has now is that images inside pinholes are excluded

Fixed now

crates/viewer/re_view_spatial/src/heuristics.rs
```diff
@@ -1,17 +1,57 @@
...
 use crate::visualizers::SpatialViewVisualizerData;
+use crate::{SpatialView2D, SpatialView3D};
 
-/// Returns all entities for which a visualizer of the given kind would be picked.
-///
-/// I.e. all entities for which at least one visualizer of the specified kind is "maybe visualizable"
-/// *and* has a relevant archetype.
-/// (we can't reason with "visualizable" because that can be influenced by view properties like its origin)
-pub fn default_visualized_entities_for_visualizer_kind(
+pub struct VisualizedEntities {
```
Wumpf: would be nice to come up with a name that makes the heuristic nature of this more clear. Right now it sounds like this has authorative information on whether something can be visualized

crates/viewer/re_view_spatial/src/view_3d.rs
```diff
@@ -368,34 +368,39 @@ impl ViewClass for SpatialView3D {
...
+                //
+                // TODO(andreas):
+                // It's tempting to add a visualizer for view coordinates so that it's already picked up via `entities_with_indicator_for_visualizer_kind`.
+                // Is there a nicer way for this or do we want a visualizer for view coordinates anyways?
+                // There's also a strong argument to be made that ViewCoordinates implies a 3D space, thus changing the SpacialTopology accordingly!
+                let engine = ctx.recording_engine();
+                ctx.recording().tree().visit_children_recursively(|path| {
+                    if let Some(components) = engine.store().all_components_for_entity(path)
+                        && components.into_iter().any(|component| {
+                            archetypes::Pinhole::all_components().iter().any(|c| c.component == component)
```
Wumpf: did you perchance find out why the pinhole is still needed her? 🤔

crates/viewer/re_viewer_context/src/view/spawn_heuristics.rs
```diff
@@ -86,4 +86,15 @@ impl RecommendedView {
             .hash64()
             .into()
     }
+
+    pub fn exclude_entities(&mut self, excluded: &[EntityPath]) {
+        for e in excluded {
+            if e.is_descendant_of(&self.origin) {
```
Wumpf: this totally makes sense for your heurist enhancement on the spatial view, but in general it could be that a heuristic produces a `RecommendedView` that has a query filter that _doesn't_ use `+ $origin/**` in which case the behavior of subsequent calls to `exclude_entities` would be very confusing. I think this needs fixing in order to avoid potential future confusion!

crates/viewer/re_view_spatial/tests/snapshots/spawn_heuristics__detect_and_track_objects_like_scene_2d_view_heuristic.snap
```diff
@@ -2,19 +2,28 @@
...
+                },
             ],
         },
-        RecommendedView {
-            origin: /video,
-            query_filter: [
-                "+ $origin/**",
-            ],
+    ),
+    views_3d: Some(
```
Wumpf, line 24: nice side-effect to have it confirmed that we don't recommend 3d views here! 👍

crates/viewer/re_view_spatial/tests/snapshots/spawn_heuristics__should_be_three_2d_and_one_3d.snap
```diff
@@ -2,25 +2,41 @@
 source: crates/viewer/re_view_spatial/tests/spawn_heuristics.rs
 expression: recommended_views
 ---
-ViewSpawnHeuristics {
-    recommended_views: [
-        RecommendedView {
-            origin: /3D/camera,
-            query_filter: [
-                "+ $origin/**",
+RecommendedViews {
```
Wumpf, line 5: the title of this test needs to change now - definitely not 3 views here 😅
also we're not calling them "space views" for more than a year now 🙈

crates/viewer/re_view_spatial/tests/spawn_heuristics.rs
```diff
@@ -204,6 +222,35 @@ fn test_mixed_2d_and_3d() {
     run_heuristics_snapshot_test("should_be_three_space_views", &test_context);
 }
 
+#[test]
+fn test_mixed_2d_and_3d_at_root() {
```
Wumpf, line 227: nice test, exactly what I was hoping for :)

crates/viewer/re_view_spatial/tests/spawn_heuristics.rs
```diff
@@ -40,6 +40,7 @@ enum EntityKind {
 
 fn build_test_scene(entities: &[(&'static str, EntityKind)]) -> TestContext {
     let mut test_context = TestContext::new_with_view_class::<re_view_spatial::SpatialView2D>();
+    test_context.register_view_class::<re_view_spatial::SpatialView3D>();
```
Wumpf, line 44: nit: `new_with_view_class` is really just calling `register_view_class`. Would be nicer to read to do

```rs
    let mut test_context = TestContext::new()
    test_context.register_view_class::<re_view_spatial::SpatialView2D>();
    test_context.register_view_class::<re_view_spatial::SpatialView3D>();
```

crates/viewer/re_view_spatial/src/heuristics.rs
IsseW: Changed it to `IndicatedVisualizableEntities`, not fully happy with it because of
> (we can't reason with "visualizable" because that can be influenced by view properties like its origin)

But I don't want to bikeshed too much either

crates/viewer/re_viewer_context/src/view/spawn_heuristics.rs
IsseW: Ah yeah, good point! Changed it to actually consider the filter now.

crates/viewer/re_view_spatial/src/view_3d.rs
IsseW: Seems that in the current version it actually isn't needed. Will see if tests pass with it removed, but all examples I tested worked

IsseW: It still does have issues, for some reason pinholes aren't ending up in `indicated_entities` if I don't add them here.

At least the ones in the insta test.

Wumpf: huh okay. guess that's a mystery for another day then

PR #12089, opened 2025-12-03
Remove web viewer dependency from `mcap_protobuf`

MichaelGrupp:
Looks like a mistake? @grtlr This was already in your [draft](https://github.com/rerun-io/rerun/pull/11635/files#diff-90603a1da47d4c80784cb98cfaf61ac8a4312c51d47f25f070ae415b5728fe52), so let me know if this was intentional.

### DIFF
--- a/.github/workflows/contrib_rerun_py.yml
+++ b/.github/workflows/contrib_rerun_py.yml
@@ -34,6 +34,7 @@ defaults:
 
 permissions:
   contents: "read"
+  packages: "read"
 
 jobs:
   build-wheels:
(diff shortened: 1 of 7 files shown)


ntjohnson1:
We run the examples in our merge to main. So we can check that it still works

ntjohnson1 (approved):

grtlr:
I think we only _build_ C++ and Rust examples in the `channel` specified in the frontmatter of their `README.md`. For this example this is `main`.

A subset of Python examples is run using `scripts/run_all.py` (via `pixi run py-run-all-examples`).

That being said, I don't think we need the `web-viewer` feature flag here and I might have copied that from another example during scaffolding.

grtlr:
@rerun-bot full-check

grtlr (approved):

MichaelGrupp:
Full build passed, merged.

PR #6514, opened 2024-06-06
Make barchart legend settable via blueprint

jleibs:
### What
Also gets rid of more EntityProperties.

### Checklist
* [x] I have read and agree to [Contributor Guide](https://github.com/rerun-io/rerun/blob/main/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/rerun-io/rerun/blob/main/CODE_OF_CONDUCT.md)
* [x] I've included a screenshot or gif (if applicable)
* [x] I have tested the web demo (if applicable):
  * Using examples from latest `main` build: [rerun.io/viewer](https://rerun.io/viewer/pr/6514?manifest_url=https://app.rerun.io/version/main/examples_manifest.json)
  * Using full set of examples from `nightly` build: [rerun.io/viewer](https://rerun.io/viewer/pr/6514?manifest_url=https://app.rerun.io/version/nightly/examples_manifest.json)
* [x] The PR title and labels are set such as to maximize their usefulness for the next release's CHANGELOG
* [x] If applicable, add a new check to the [release checklist](https://github.com/rerun-io/rerun/blob/main/tests/python/release_checklist)!

- [PR Build Summary](https://build.rerun.io/pr/6514)
- [Recent benchmark results](https://build.rerun.io/graphs/crates.html)
- [Wasm size tracking](https://build.rerun.io/graphs/sizes.html)

To run all checks from `main`, comment on the PR with `@rerun-bot full-check`.

PR #12090, opened 2025-12-03
Allow specifying widths of `BarChart` bars

### LINKED ISSUE #4764: Allow to specify boundaries for BarChart
BarChart currently only supports discrete equal size bins starting with 0. It would be nice to support arbitrary x boundaries and hence bars of varying width.

### LINKED ISSUE #6265: Allow specifying horizontal range of barchart
Currently, the bar chart only tracks the heights of the bins and not the "edges" of each bin. An unfortunate downside of this is that when I have a distribution of scalar data that I would like to view in a histogram, I have no idea where the mean is. Consider an animation like below, where the _number_ of bins stays constant, but the mean moves.

example

I believe animations such as this could be achieved by extending the `BarChart` component to have `edges` and the `BarChart` blueprint to have view bounds.

bilsen:
### Related
* Closes #6265
* (Should) close #4764

### What
Adds an optional `widths` component to the `BarChart` archetype. Uses this when rendering the `BarChart` in the viewer.

This together with abscissa should allow logging arbitrary bar charts, to visualize distributions. 
From `bar_chart` snippet:

image


crates/viewer/re_view_bar_chart/src/view_class.rs
bilsen, line 1: this file is getting kinda clunky, might be nice to impl/use some conversion trait for TensorBuffer to simplify this

crates/store/re_types/definitions/rerun/archetypes/bar_chart.fbs
```diff
@@ -25,4 +25,7 @@ table BarChart (
 
   /// The abscissa corresponding to each value. Should be a 1-dimensional tensor (i.e. a vector) in same length as values.
   abscissa: rerun.components.TensorData ("attr.rerun.component_optional", nullable, order: 3000);
+
+  /// The width of the bins. Should be a 1-dimensional tensor (i.e. a vector) in same length as values.
+  widths: rerun.components.TensorData ("attr.rerun.component_optional", nullable, order: 4000);
```
Wumpf: The reason we ended up with `TensorData` on `abscissa` is it was really important to the person contributing that this has the same datatype flexibility as the `values`. Which makes sense since this is again still core data.

So, now we're not super happy with today's `TensorData` for a bunch of reasons: as you noted there's a lot of bloaty conversion code. But also, it uses arrow unions under the hood and they cause a lot of trouble when querying data back out in general.

All of which makes me wonder whether we could use a much simpler component type here that also describes the semantics better - in a way we don't have "raw numerical data" here, but rather a visual width.
We already have a `Length` component to define a length in a local coordinate system (which is opposed to `StrokeWidth` which is a width in UI  coordinates).
The big drawback here is though that it is limited to float values. Just going by your own usecase, is that too limiting or just fine?

---

Independent of the above, the docstring here should better describe how these widths are specified (not in UI points as many other things but rather in x axis units just like `values`) and that they default to `1`

crates/viewer/re_view_bar_chart/src/view_class.rs
Wumpf: agreed, but I think we can adjourn that (and depending on discussion outcome on comment above we won't make it worse here)

Wumpf (changes requested):
Thanks for contributing this feature! Looks good in principle, but I'd like to discuss whether a simpler component type would be appropriate, see comment. 🙂 

Otherwise, could you please add an example for this to https://github.com/rerun-io/rerun/blob/main/docs/snippets/all/archetypes/bar_chart.py (/cpp/rs) and add a screenshot the outcome to the PR description?
Changing those snippets will unfortunately also require updating https://github.com/rerun-io/rerun/blob/main/tests/assets/rrd/snippets/archetypes/bar_chart.rrd which we use for backwards compatibility tests so that  `compare_snippet_output.py` runs through. This bit can be a bit tedious, so happy to help with that if needed.

Wumpf:
If I understand correctly, https://github.com/rerun-io/rerun/issues/6265 is actually not closed by this since we'd need a property of the graph itself to be wider than the bars that show up in it. That should be a view property of bar chart view then.

However, it _also_ looks like I incorrectly closed https://github.com/rerun-io/rerun/issues/4764 when the abscissa PR came around which doesn't fully address that ticket 😳

crates/store/re_types/definitions/rerun/archetypes/bar_chart.fbs
bilsen: Thanks for the quick review :smile: 

will fix the doc

I would be fine with only allowing float values, although I would also feel its weird to have a different datatype than abscissa and values.

bilsen:
> If I understand correctly, #6265 is actually not closed by this since we'd need a property of the graph itself to be wider than the bars that show up in it. That should be a view property of bar chart view then.
> 
> However, it _also_ looks like I incorrectly closed #4764 when the abscissa PR came around which doesn't fully address that ticket 😳

I actually think that #6265 should already be closed, since the bar chart viewer currently has the x and y axis specified through blueprint. Maybe it was fixed here? https://github.com/rerun-io/rerun/pull/6514

crates/store/re_types/definitions/rerun/archetypes/bar_chart.fbs
bilsen: Some additional thought:
As a user, the ideal workflow would be to just log some table of floats (with named columns, perhaps) and to decide when using the viewer how to visualize the data (maybe using a python script :no_mouth:). Right now it feels like rerun has some (really great) default ways to visualize my data, but the data is pretty difficult to interact with/show in non-supported ways after it gets converted into a .rrd file.

In the case of `BarChart`s, what i would _really_ like, is to just log the entire set of points which represents samples from my distribution, and to look at the distribution evolved over time in the viewer.

Im not familiar enough with the project to make any concrete suggestions though

Wumpf: 🤔 

Re datatypes:
We generally want to get more flexible in regards to datatypes. It's been long the vision that components types are a semantic layer that allow generically a number of different datatypes. We made some steps towards that but aren't quite there yet.
Agreed that in this concrete case mixing types might be a bit unintuitive, but from a semantics pov it makes sense. So if floats on widths aren't limiting for you here I'd like to go through with it

Re visualization configuration:
Already today you can specify all these properties in blueprint overrides as well - it often makes sense to only put what you consider "data" to be in the rrd and everything else be a blueprint override, including stuff like bar widths. That said, the ux of practically loading an rrd and sending different blueprints is a bit lacking, but it's possible and iirc demonstrated in the docs.
Also, UI for blueprint overrides is lacking for more complex stuff like arrays, making it impossible to configure these things adhoc in a meaningful way without code.

Wumpf:
What's imho missing for https://github.com/rerun-io/rerun/issues/6265 is to set the plot window to a certain width, similar to what the time axis width setting on time series view does

Wumpf:
Thanks for adding examples & screenshot!

bilsen:
> What's imho missing for #6265 is to set the plot window to a certain width, similar to what the time axis width setting on time series view does

Hmm I'm not familiar with the time series view, but this is how it looks on my project (without the widths changes) when zoomed out and scrubbing the timeline:

[Screencast from 2025-12-05 10-42-47.webm](https://github.com/user-attachments/assets/d3d825d2-5adc-4048-a563-a4f2bfd7e9fd)

It's maybe not optimal, (might be a better default to specify the x-axis width to cover the entire timeline instead of dynamically changing the limits) but I think this is what was asked for in the issue?

Wumpf:
The issue author of https://github.com/rerun-io/rerun/issues/6265 specifically asks for bins moving within a given range as on the supplied gif, so I understood it as having to configure the range of the plots. Well, either way that's a feature disparity there with the time series, so if that ticket doesn't cover that we need a new one 😄 

On the time series view you can set x and y axis range from blueprint code, which on the bar chart you can only do via manual interaction which is also then not persisted over sessions

crates/store/re_types/definitions/rerun/archetypes/bar_chart.fbs
bilsen: Alright, will change the datatype. I am just a bit mystified regarding what the concept of a component is, since it seems like there's also something like a 'list of components' that can also be part of an archetype?
and the syntax to query these components seem to differ to regular components aswell. Think i sort of got it working now using [Length] as a datatype, will push some changes shortly

bilsen:
I also changed the component fallback, trying to match what I saw done with 'radii' and other similar 'list-components'

crates/viewer/re_view_bar_chart/src/view_class.rs
```diff
@@ -292,97 +293,14 @@ impl ViewClass for BarChartView {
                         TensorBuffer::F32(data) => data.iter().map(|v| *v as f64).collect(),
                         TensorBuffer::F64(data) => data.iter().copied().collect(),
                     };
-
-                    let chart = match &tensor.buffer {
```
bilsen, line 276: Was thinking this should be a pretty uncontroversial refactor. Can split into another PR if necessary however.

crates/store/re_types/definitions/rerun/archetypes/bar_chart.fbs
Wumpf: Thanks, will have a look soon!

>  since it seems like there's also something like a 'list of components' that can also be part of an archetype?

Internally we always store arrays for each components and the notation (whether array or not) in the fbs files is more or less just a hint for the api codegen. `TensorData` is a _very_ weird exception many ways right now. It is an array internally and I think our last remaining arrow union type, so having that as a starting point makes everything else confusing 😄
Generally the idea is that there's a bunch of raw data types that are then wrapped by component types which wrap them into some loose semantic.

tests/assets/rrd/snippets/archetypes/bar_chart.rrd
Wumpf, line 1: this file needs to be updated rather than deleted

crates/viewer/re_view_bar_chart/src/visualizer_system.rs
```diff
@@ -57,18 +57,32 @@ impl VisualizerSystem for BarChartVisualizerSystem {
...
             if tensor.is_vector() {
+                let length: u64 = tensor.shape().iter().product();
+
                 let abscissa: components::TensorData =
                     results.get_mono_with_fallback(BarChart::descriptor_abscissa().component);
                 let color = results.get_mono_with_fallback(BarChart::descriptor_color().component);
-                let widths: components::TensorData =
-                    results.get_mono_with_fallback(BarChart::descriptor_widths().component);
+                let widths = if let Some(chunked_widths) =
+                    results.get_required_chunks(BarChart::descriptor_widths().component)
```
Wumpf: should use `get_optional_chunks`

crates/viewer/re_view_bar_chart/src/visualizer_system.rs
```diff
@@ -54,21 +57,42 @@ impl VisualizerSystem for BarChartVisualizerSystem {
...
+                        widths.extend_from_slice(slice);
+                    }
+                    widths
+                } else {
+                    let fallback_width: Length = typed_fallback_for(
+                        &ctx.query_context(data_result, &view_query.latest_at_query()),
+                        BarChart::descriptor_widths().component,
+                    );
+                    vec![fallback_width.0.into(); length as usize]
+                };
```
Wumpf: This goes into the right direction, but something we usually do is to repeat the last element often enough to ensure that we have enough of them. This has two advantages:
* if there's less widths than values we won't just  loose values
* a user can just specify a single value that is then splatted out to everything
    * that's then consistent with the fallback you put in which is just returning a single value :) 

Unfortunately that's not handled by any query later and is done in-place most of everywhere.

My suggestion would be to use the fact that we always have a `LatestAtResults` where we know that the we get back a single row worth of data. This allows us also to set the semantics straight:
```rs
                let widths: &[f32] = results
                    .iter_as(view_query.timeline, BarChart::descriptor_widths().component)
                    .slice::<f32>()
                    .next() // It's a latest-at result, therefore this is a single row, therefore we there can't be more than 1 chunk involved.
                    .map_or(&[], |((_time, _row), slice)| slice);
                let fallback_width = typed_fallback_for::<Length>(
                    &ctx.query_context(data_result, &view_query.latest_at_query()),
                    BarChart::descriptor_widths().component,
                )
                .0;
                let widths = clamped_vec_or(fallback_width, length as usize, fallback_width);
````
(there's some other subtle differences in how this plays out, but it has mostly the same effect.

`clamped_vec_or` is a utility method that lives over in [re_view_spatial::entity_iterator.rs](https://github.com/rerun-io/rerun/blob/deb489e56fc4f1b36748364ee243b1d2334c83b8/crates/viewer/re_view_spatial/src/visualizers/utilities/entity_iterator.rs#L59) but I think it should be moved to `re_view` (can go to lib.rs for now, where there's already a related similar utility, `clamped_or_nothing`).
Guess a `clamped_vec_or_else` would be nice to not evaluate the fallback if not needed here 🤔 . But I've burdened you already enough with refactor asks

I realize that some of this stuff goes pretty deep into Rerun data model arcana and our query utilities are really poorly documented and a bit all over the place. Plus since the tensor view has a lot of "special" queries it has no good examples yet (unlike most views it only deals with latest-at and it has to deal with the relatively strange `TensorData` query).
I'm bit embarrassed we left things in this state; I think the underlying mechanics do work out, but it needs a thorough cleanup & documentation pass.

tests/assets/rrd/snippets/archetypes/bar_chart.rrd
Wumpf, line 1: this file needs updating rather than deleting. The most reliable (albeit very slow the first time around) way to do so is `pixi run -e py python ./docs/snippets/compare_snippet_output.py --write-missing-backward-assets archetypes/bar_chart`

crates/viewer/re_view_bar_chart/src/view_class.rs
Wumpf: no need to, appreciated! :)

tests/assets/rrd/snippets/archetypes/bar_chart.rrd
bilsen: Right, idk why it got removed... not so familiar with git lfs. 
Also, running the snippet comparisons made me run out of memory, had to increase swap size. Maybe some configuration error on my end, but compiling the c++ sdk seemed to be very tough.

crates/viewer/re_view_bar_chart/src/visualizer_system.rs
bilsen: Will take a look after work. Yea, coming from bevy the data model is not quite as intuitive to an outsider :smile: But thank you for the quick reviews, I think I have some better understanding now.

bilsen:
Did some refactoring for clamped vec utils (also changed to Cow<'a, [T]>) and used this utility

Wumpf:
Sorry for the delay, didn't have time to check on this last week. Will resolve conflicts and having a look over stuff today

Wumpf (approved):
looking great! Thanks for the additional refactoring!

Wumpf:
something is broken with contributor CI again -.-
running snippet tests locally over lunch, then probably good to merge (and if not I'll mop up after it)

PR #12092, opened 2025-12-03
Handle invalid server in connection registry

jprochazk:
When connecting to something that is not a Rerun server, you are likely to get a very confusing error message. This should help quite a bit by immediately telling you that what you're trying to connect to is not a Rerun server.

### DIFF
--- a/rerun_py/tests/e2e_redap_tests/test_client.py
+++ b/rerun_py/tests/e2e_redap_tests/test_client.py
@@ -27,6 +27,8 @@ def test_urls(prefilled_catalog: PrefilledCatalog) -> None:
     assert url.path.endswith("/simple_datatypes") or url.path.endswith("/simple_datatypes/")
 
 
+# TODO(#12122): It'd be nice if the connection timeout was configurable, so we would not have to wait for 30 seconds for this test.
+@pytest.mark.skip
 def test_network_unreachable() -> None:
     """Tests that the client raises an error when the server is unreachable."""
 
@@ -34,7 +36,6 @@ def test_network_unreachable() -> None:
         raise TimeoutError("the operation did not time out on time")
 
     signal.signal(signal.SIGALRM, timeout_handler)
-    # TODO(#12122): It'd be nice if the connection timeout was configurable, so we would not have to wait for 30 seconds for this test.
     signal.alarm(60)  # Our connection timeout is 30 seconds (ehttp default). Let's be generous to avoid flakiness
 
     try:


jleibs (approved):
This definitely improves things for me:
```
Telemetry disabled. All logging will be ignored.
  File "/home/jleibs/rerun/test-dev.py", line 5, in <module>
    cli = rr.catalog.CatalogClient(address=addr)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jleibs/rerun/rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py", line 85, in __init__
    self._internal = CatalogClientInternal(address, token)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ConnectionError: InvalidServer: rerun://rrc-jleibs-dev.dev.rerun.io:443 is not a valid Rerun server
```

I think the open question for me is if it's worth just auto-adding `api` prefix so that things "just work".

jprochazk:
New problem: The OSS server does not support `/version`, so the client can never connect to it.

crates/store/re_redap_client/src/connection_registry.rs
```diff
@@ -309,6 +309,21 @@ impl ConnectionRegistryHandle {
                 None => None,
             };
 
+        // It's a common mistake to connect to `asdf.rerun.io` instead of `api.asdf.rerun.io`,
+        // so if what we're trying to connect to is not a valid Rerun server, then cut out
+        // a layer of noise:
+        {
+            let Ok(res) =
+                ehttp::fetch_async(ehttp::Request::get(format!("{}/version", origin.as_url())))
```
andrea-reale: does `ehttp` have a sane connection timeout by default? 

Our tonic stack did not, I am adding it now. I just want to make sure this does.

jprochazk: Doesn't seem like it :/

jprochazk: Nevermind, looking deeper into the underlying `ureq`, there is a default connect timeout of 30 seconds.

jprochazk: Also adding a way to configure it, but I believe this is not an issue here: https://github.com/emilk/ehttp/pull/71

jprochazk:
This is now waiting on:
- https://github.com/rerun-io/rerun/pull/12108

Tests should pass after just a rebase, in theory

jprochazk:
I'm just going to increase the timeout in that test to be able to land this. I don't think the `ehttp` timeout change is going to happen today.

jprochazk:
After discussion in Slack, also skipping the test. We'll revive it once we have configurable timeouts.

PR #12065, opened 2025-12-03
Have perf_telemetry feature enabled by default

zehiko:
### What

This change makes ``perf_telemetry`` feature enabled by default. We want to use rerun sdk with perf telemetry enabled in our benchmark tests, but also in every day workflows. 

Logging strategy for the SDK, cli or the viewer is not impacted by this change and is controller by the env var ``TELEMETRY_ENABLED``. When this env var is set, logs, traces and metric pipelines are enabled (and configured with default values from ``re_perf_telemetry``).

Targets like ``py-build-perf-debug`` will now enable ``tracy`` feature of ``re_perf_telemetry``, but ``tracy`` feature is **disabled** by default. It seems that ``{crate}/{feature}`` syntax works in some places, but not others, hence I had to define ``perf_telemetry_tracy`` feature in some places.

### Testing done

```
pixi run rerun

[2025-12-03T09:27:00Z DEBUG rerun::commands::entrypoint] Detected 14 cores. Using 12 compute threads.
Telemetry disabled. All logging will be ignored.
[2025-12-03T09:27:00Z DEBUG rerun::commands::entrypoint] Parsing --server-memory-limit (for gRPC server)
[2025-12-03T09:27:00Z DEBUG rerun::commands::entrypoint] Server memory limit: 0B
[2025-12-03T09:27:00Z DEBUG rerun::commands::entrypoint] Parsing --memory-limit (for Viewer)
[2025-12-03T09:27:00Z DEBUG re_memory::memory_limit] Setting memory limit to 27.0 GiB, which is 75% of total available memory (36.0 GiB).
[2025-12-03T09:27:00Z DEBUG re_memory::memory_limit] Setting memory limit to 27.0 GiB, which is 75% of total available memory (36.0 GiB).
[2025-12-03T09:27:00Z INFO  re_grpc_server] Listening for gRPC connections on 0.0.0.0:9876. Connect by running `rerun --connect rerun+http://127.0.0.1:9876/proxy`
[2025-12-03T09:27:00Z DEBUG eframe] Using the wgpu renderer
```
```
TELEMETRY_ENABLED=true pixi run rerun

Turning off stderr logging because of perf_telemetry needs exclusive access to the global tracing subscriber
Disabling crash handler because of perf_telemetry/TELEMETRY_ENABLED
  2025-12-03T09:28:14.725987Z  INFO  Telemetry initialized
    at crates/utils/re_perf_telemetry/src/telemetry.rs:456 on main ThreadId(1)

[...]

PR #12093, opened 2025-12-03
make sure we don't crate spans when telemetry is not enabled

zehiko:
### Related
* Part of https://linear.app/rerun/issue/RR-3112/log-spam-in-viewer

### What

After enabling perf telemetry feature in https://github.com/rerun-io/rerun/pull/12065 we started spamming the viewer with tracing spans. This quick fix ensures we don't create the spans if tracing feature is not enabled. 

I tried doing this check earlier in the setup of the client telemetry, but since we have predefined client type, I just couldn't find a way without making this much more complicated. 

### Testing done
Tested adding local Redap server and with this fix the spammy messages are gone.

### DIFF
--- a/crates/utils/re_perf_telemetry/src/grpc.rs
+++ b/crates/utils/re_perf_telemetry/src/grpc.rs
@@ -24,6 +24,7 @@ pub struct GrpcMakeSpan {
 
 impl GrpcMakeSpan {
     pub fn new() -> Self {
+        // if telemetry is not explicitly enabled through an env var, we create noop spans
         let create_noop_spans = !std::env::var("TELEMETRY_ENABLED")
             .is_ok_and(|v| v == "1" || v.to_lowercase() == "true" || v.to_lowercase() == "yes");
 


crates/utils/re_perf_telemetry/src/grpc.rs
```diff
@@ -16,16 +16,26 @@ const RERUN_HTTP_HEADER_SERVER_VERSION: &str = "x-rerun-server-version";
 #[derive(Debug, Clone)]
 pub struct GrpcMakeSpan {
     gauge: opentelemetry::metrics::Gauge<u64>,
+    // unfortunately we can't have different implementation of `MakeSpan` as that creates a ripple effect
+    // through the entire hierarchy of types of the RedapClient and its usage, hence to disable the span
+    // creation, we create noop spans instead if telemetry is disabled at runtime
+    create_noop_spans: bool,
```
Wumpf, line 22: sounds pragmatic to me. also stuff not being conditionally compiled is always nice

crates/utils/re_perf_telemetry/src/grpc.rs
```diff
@@ -16,16 +16,26 @@ const RERUN_HTTP_HEADER_SERVER_VERSION: &str = "x-rerun-server-version";
...
 }
 
 impl GrpcMakeSpan {
     pub fn new() -> Self {
+        let create_noop_spans = !std::env::var("TELEMETRY_ENABLED")
+            .is_ok_and(|v| v == "1" || v.to_lowercase() == "true" || v.to_lowercase() == "yes");
```
Wumpf, line 29: Please use `!re_log::env_var_is_truthy` for this. We keep repeating these truthy checks in various forms :)
(assuming you have `re_log` available.. but the example there is literally with `TELEMETRY_ENABLED` 😄 )

zehiko: yeah, I added it and I wanted to, but this crate doesn't depend on ``re_log`` and I didn't want to further complicate the dependency tree. (And yeah, for a second there I thought - ok, but certainly I could move that check in ``re_perf_telemetry``, but also one of the creates using ``env_var_is_truthy`` doesn't depend on it).

crates/utils/re_perf_telemetry/src/grpc.rs
zehiko: 🙏

crates/utils/re_perf_telemetry/src/grpc.rs
zehiko: (left as is for now as we don't want to add dependencies over this)

PR #12095, opened 2025-12-04
Add helper type `ApiResult` for terser code

emilk:
Code readability is important


Wumpf (approved):
neat, thanks for splitting this out

PR #12096, opened 2025-12-04
Migrations should not rely on constants that may change over time

grtlr:
### Related

* Follow-up of #11905.

### What

Migrations need to be stable across time and changes to our crates. It's therefore an anti-pattern to use constants in migration code that come from external crates.

This PR removes the dependencies to the constants in `re_types_core` and hardcodes the values.

### TODO

* [x] full check


grtlr:
@rerun-bot full-check

grtlr:
@rerun-bot full-check

grtlr:
@rerun-bot full-check

PR #12097, opened 2025-12-04
Use new `/GetRrdManifest` endpoint

emilk:
### Related
* Closes RR-3000
* Follows #12076

### What 
* Rename `ChunkIndex` to `RrdManifest`
* Load it as a separate call first

Right now the viewer doesn't use the `RrdManifest` for anything other than the progress bar, but the goal is to use it for larger-than-RAM.

### DIFF
--- a/crates/store/re_redap_client/src/connection_client.rs
+++ b/crates/store/re_redap_client/src/connection_client.rs
@@ -1,5 +1,3 @@
-use std::sync::Arc;
-
 use arrow::array::RecordBatch;
 use arrow::datatypes::{Schema as ArrowSchema, SchemaRef};
 use re_arrow_util::ArrowArrayDowncastRef as _;


crates/utils/re_arrow_util/src/lib.rs
```diff
@@ -89,16 +89,63 @@ mod tests {
 
 // ----------------------------------------------------------------
 
+/// Error used when a column is missing from a record batcj
```
teh-cmc:
```suggestion
/// Error used when a column is missing from a record batch
```

crates/store/re_entity_db/src/rrd_manifest_index.rs
```diff
@@ -18,11 +18,11 @@ pub struct ChunkInfo {
 ///
 /// This is currently used to show a progress bar.
 ///
-/// This is constructed from one ore more [`ChunkIndexMessage`], which is what
+/// This is constructed from one ore more [`RrdManifestMessage`], which is what
 /// the server sends to the client/viewer.
 /// TODO(RR-2999): use this for larger-than-RAM.
 #[derive(Default, Debug, Clone)]
-pub struct ChunkIndex {
+pub struct RrdManifest {
```
teh-cmc: I'm not a fan of re-using the exact same name as the actual `RrdManifest` object, this is confusing.

crates/store/re_redap_client/src/connection_client.rs
```diff
@@ -365,12 +367,38 @@ where
...
+            .map_err(|err| ApiError::tonic(err, "/GetRrdManifest failed"))?
+            .into_inner()
+            .data
+            .ok_or_else(|| {
+                let err = missing_field!(GetRrdManifestResponse, "data");
+                ApiError::serialization(err, "missing field in /GetRrdManifest response")
+            })?
+            .pipe(RecordBatch::try_from)
+            .map_err(|err| ApiError::serialization(err, "failed parsing /GetRrdManifest response"))?
+            .pipe(RrdManifestMessage::try_from_record_batch)
```
teh-cmc: You should not be implementing transport-to-app conversion yourself. The provided one has a lot of extra checks that must be run.

crates/store/re_types_core/src/rrd_manifest_message.rs
teh-cmc, line 1: I don't understand the need to introduce this type. This looks like a less fleshed out version of the actual application-level `RrdManifest`. The fact that it implements its own unsafe deserialization is very problematic.

crates/store/re_entity_db/src/rrd_manifest_index.rs
emilk: Yeah, I should call this `RrdManifestIndex` or something

crates/store/re_redap_client/src/connection_client.rs
```diff
@@ -365,12 +369,45 @@ where
...
+                let err = missing_field!(GetRrdManifestResponse, "data");
+                ApiError::serialization(err, "missing field in /GetRrdManifest response")
+            })?
+            .pipe(RecordBatch::try_from)
+            .map_err(|err| ApiError::serialization(err, "failed parsing /GetRrdManifest response"))
+            .map(|data| {
+                RrdManifest {
+                    // TODO(cmc): fix this
+                    store_id: StoreId::empty_recording().with_recording_id(segment_id.to_string()),
+                    sorbet_schema: Arc::unwrap_or_clone(data.schema()),
```
teh-cmc: `data.schema()` is unrelated so that's very risky, better to just empty it:
```suggestion
                    sorbet_schema: Arc::new(Schema::empty()),
```

PR #12099, opened 2025-12-04
Adds analytics event for CLI commands

aedm:
### Related

Closes RR-2987

### What

- Adds anaytics events for CLI commands, subcommands and main flags used.


crates/top/rerun/src/commands/entrypoint.rs
```diff
@@ -1526,3 +1529,86 @@ impl ReceiversFromUrlParams {
         Ok(())
     }
 }
+
+/// Records analytics for the CLI command invocation.
+#[cfg(feature = "analytics")]
+fn record_cli_command_analytics(args: &Args) {
```
grtlr, line 1535: Should we destructure the entire `Args` to make sure that we don't miss any newly added flags? We could then also document the flags that we skip entirely.

aedm: I like that idea a lot, thanks! Done.

PR #12028, opened 2025-12-01
Mark frame based transform apis as stable

Wumpf:
### Related

* Fixes RR-2777
* Fixes RR-2699

### What

Also...

* new transform doc category https://landing-nf6b8q8x3-rerun.vercel.app/docs/reference/types/archetypes#transforms
* mark new transform fields/archetypes/examples as stable
* update example pictures
* include new hierarchy frame example on `CoordinateFrame` and `Transform3D`
    * on the later neither the new one nor the older non-frame hierarchy examples are part api doc since it got too much!

PR #12100, opened 2025-12-04
Remove most tf zombie todos

IsseW:
### Related

- Part of RR-3029

### What

Removes most zombie TODOs related to the tf project.


crates/store/re_types/src/components/transform_frame_id.rs
```diff
@@ -23,13 +23,13 @@ use ::re_types_core::{DeserializationError, DeserializationResult};
...
-/// Note that any transform logged at an entity path describes a relationship between the two transform
-/// frames represented by that entity path and its parent path,
-/// **not** the transform frame that the entity path may be using.
+/// These implicit transform frames look like `tf#path/to/entity`.
+///
+/// Note that any [`archetypes::Transform3D`][crate::archetypes::Transform3D]s logged at an entity path describes a relationship between the
+///  transform frames `parent` and `child`, **not** the transform frame that the entity path may be
+/// using. Which is defined by [`archetypes::CoordinateFrame`][crate::archetypes::CoordinateFrame].
 ///
 /// ⚠️ **This type is _unstable_ and may change significantly in a way that the data won't be backwards compatible.**
```
ntjohnson1: Is unstable different from experimental?

docs/content/reference/types/components/transform_frame_id.md
```diff
@@ -6,13 +6,13 @@ title: "TransformFrameId"
 ⚠️ **This type is _unstable_ and may change significantly in a way that the data won't be backwards compatible.**
```
ntjohnson1: Same note that you removed experimental but not unstable

crates/store/re_types/src/components/transform_frame_id.rs
IsseW: Good catch, yeah that should also be removed. Think it was missed here https://github.com/rerun-io/rerun/pull/12028

ntjohnson1: Great! I don't totally know what our process is for stabilizing things so mostly just wasn't sure if there was a distinction.

crates/store/re_types/definitions/rerun/components/transform_frame_id.fbs
```diff
@@ -2,15 +2,13 @@ namespace rerun.components;
...
-/// Note that any transform logged at an entity path describes a relationship between the two transform
-/// frames represented by that entity path and its parent path,
-/// **not** the transform frame that the entity path may be using.
-// TODO(RR-2486): Reference `CoordinateFrame` & `FrameTransform` once they exist.
-// TODO(RR-2698): Explain implicit transform frames in more detail.
+/// These implicit transform frames look like `tf#path/to/entity`.
+///
+/// Note that any [archetypes.Transform3D]s logged at an entity path describes a relationship between the
+///  transform frames `parent` and `child`, **not** the transform frame that the entity path may be
+/// using. Which is defined by [archetypes.CoordinateFrame].
```
MichaelGrupp: This sentence confuses me, had to reread a few times to make sure I get it right. Is this better?
```suggestion
/// Note that any [archetypes.Transform3D]s logged with both `parent_frame` and `child_frame` set
/// describes a relationship between these parent and child transform frames, **not** the transform frame
/// that the entity path may be using (defined by an [archetypes.CoordinateFrame]).
```

MichaelGrupp: (i.e. the statement is only true if the frame names are set, which is optional)

IsseW: I think it would technically be true even if it isn't set. If everything uses fallbacks `parent_frame` would be the entities parent implicit frame `tf#parent`. `child_frame` would be the entities implicit frame `tf#parent/child`. And the entities coordinate frame would also be `tf#parent/child`.

But I do think your phrasing is better either way

MichaelGrupp (approved):

PR #12101, opened 2025-12-04
Allow different associated entity path for static & temporal

Wumpf:
### Related

* Fixes RR-2803
* Fixes RR-3042

### What

... and log an error if we violate the invariant by changing the child_frame within static or temporal time.

Allowing different "associated entity paths" for static & temporal data greatly improves the usability story for URDF: the URDF file can log "default transforms" at static time and then we can later add animations on arbitrary timelines at a different path. I confirmed together with @MichaelGrupp that this PR has this effect! 🥳

### DIFF
--- a/crates/store/re_tf/src/transform_resolution_cache.rs
+++ b/crates/store/re_tf/src/transform_resolution_cache.rs
@@ -1309,7 +1309,7 @@ mod tests {
         PerStoreChunkSubscriber, RowId,
     };
     use re_log_types::{
-        StoreId, TimePoint, Timeline,
+        StoreId, StoreInfo, TimePoint, Timeline,
         example_components::{MyPoint, MyPoints},
     };
     use re_types::{ChunkId, archetypes};
@@ -1434,10 +1434,7 @@ mod tests {
     }
 
     fn new_entity_db_with_subscriber_registered() -> EntityDb {
-        let entity_db = EntityDb::new(StoreId::random(
-            re_log_types::StoreKind::Recording,
-            "test_app",
-        ));
+        let entity_db = EntityDb::new(StoreInfo::testing().store_id);
         let _ = TestStoreSubscriber::subscription_handle();
         entity_db
     }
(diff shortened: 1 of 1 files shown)


crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -680,19 +767,19 @@ impl TreeTransformsForChildFrame {
...
         // Separate check to work around borrow checker issues.
         if pinhole_projection == &CachedTransformValue::Invalidated {
             let transform = query_and_resolve_pinhole_projection_at_entity(
-                &self.associated_entity_path,
+                self.associated_entity_path(*time_of_last_update_to_this_frame),
                 self.child_frame,
                 entity_db,
-                query,
+                // Do NOT use the original query time since that may give us information about a different child frame!
+                &LatestAtQuery::new(query.timeline(), *time_of_last_update_to_this_frame),
```
Wumpf, line 782: this is actually a drive-by bugfix for a very complicated but also kinda hard to hit bug that is entirely unrelated to the goals of this PR. I pondered whether I add a test for this, but decided to let it slide so far. Having a look right now whether I can add a test for that easily

Wumpf: I experimented a bit with tests and I think we're actually very robust against this bug by now: since we're filtering the rows for the requested frame, even a too big time will always yield the correct data. 
I'm not 100% convinced though that using `&query` is _not_ a bug, conceptually it is and if a debug assertion would fire because of it I wouldn't be angry about it 😄

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -2602,6 +2696,216 @@ mod tests {
...
+                    .with_parent_frame("parent_frame"),
+            )
+            .build()?;
+
+        #[derive(Debug)]
+        enum Scenario {
+            StaticAndTemporalAtOnce,
+            StaticFirstThenTemporal,
+            TemporalFirstThenStatic,
+        }
```
MichaelGrupp, line 2736: 🥇

PR #12102, opened 2025-12-04
Better formatting or arrow schema in snapshot tests

emilk:
Motivation should be self-explanatory:

Image


emilk:
> This will probably need a companion PR in the other repo.

I tried it, and apparently not!

PR #12105, opened 2025-12-04
Add integration test for coordinate frame overrides

Wumpf:
### Related

* Fixes RR-3029
    * this was the last tf related "zombie todo" Isse found earlier and didn't address yet

### What

Closes a small gap in our tf test coverage

### DIFF
--- a/crates/viewer/re_view_spatial/tests/snapshots/transform_many_child_parent_relations_on_single_time_and_entity_with_coordinate_frame_overrides.png
+++ b/crates/viewer/re_view_spatial/tests/snapshots/transform_many_child_parent_relations_on_single_time_and_entity_with_coordinate_frame_overrides.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:752699e5c2fe04a3fcff638f94d1f3921d1b0f0b4feb58196e573dab60023587
+size 18758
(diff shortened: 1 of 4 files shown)


crates/viewer/re_view_spatial/tests/snapshots/transform_many_child_parent_relations_on_single_time_and_entity_with_coordinate_frame_overrides.png
Wumpf, line 1: exactly same image as its neighboring `crates/viewer/re_view_spatial/tests/snapshots/transform_many_child_parent_relations_on_single_time_and_entity.png`, as it should be

IsseW: Wonder if it'd be useful to explicitly test that?

Wumpf: been wondering this with a bunch of our screenshots by now. We have a lot of stuff where we have two tests that are expected to produce the same outcome. But for convenience of debugging we do want them to write separate images. Would be cool to have "those are the same" as a postprocessing step that runs after all the tests are through. cc: @lucasmerlin 
bit tricky to set up given that the other tests should also run in parallel

PR #12108, opened 2025-12-04
Expose `/version` from oss server

jprochazk:
```
$ cargo run -p rerun-cli -- server
$ curl http://localhost:51234/version
re_server 0.28.0-alpha.1+dev [rustc 1.88.0 (6b00bc388 2025-06-23), LLVM 20.1.5] x86_64-unknown-linux-gnu (debug)
```

All Rerun servers are now required to have the `GET /version` HTTP endpoint, for us to be able to remove one layer of noise when users connects to the wrong thing.

* Unblocks https://github.com/rerun-io/rerun/pull/12092


crates/store/re_server/src/server.rs
```diff
@@ -193,13 +195,32 @@ impl ServerBuilder {
...
+        self.axum_routes = self.axum_routes.route(path, handler);
+        self
+    }
+
     pub fn build(self) -> Server {
+        let grpc_routes = self.routes_builder.routes();
+        let grpc_routes = grpc_routes.into_axum_router();
+
+        // Apply GrpcWebLayer only to gRPC routes, not HTTP routes
+        let grpc_routes = grpc_routes.layer(tonic_web::GrpcWebLayer::new());
```
jprochazk, line 208: I was getting `400` before reordering things in this way. I'm not too familiar with tonic so let me know if there's a more straightforward way to do this.

crates/store/re_server/tests/redap_tests.rs
```diff
@@ -8,3 +8,36 @@ async fn build() -> RerunCloudHandler {
 }
 
 re_redap_tests::generate_redap_tests!(build);
+
+#[tokio::test(flavor = "multi_thread")]
+async fn version() {
```
jprochazk, line 13: This is not a gRPC endpoint, so it doesn't fit neatly into existing tests. So I made it separate here.

jprochazk:
Copilot helpful as always

PR #12109, opened 2025-12-04
Fix RRD manifest sanity check failing on data that is both static & temporal

teh-cmc:
Yep, forgot that was a thing. Needless to say, the fix does not reflect the pain involved, not even close :melting_face:

### DIFF
--- a/crates/store/re_log_encoding/src/rrd/footer/instances.rs
+++ b/crates/store/re_log_encoding/src/rrd/footer/instances.rs
@@ -513,7 +513,7 @@ impl RrdManifest {
                     );
 
                     if md.get("rerun:is_static").map(|s| s.as_str()) == Some("true") {
-                        // Static columns don't have :start nor :end columns... unless they exist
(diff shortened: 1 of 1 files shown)


crates/store/re_log_encoding/src/rrd/footer/instances.rs
```diff
@@ -501,6 +512,16 @@ impl RrdManifest {
...
 
+                    if md.get("rerun:is_static").map(|s| s.as_str()) == Some("true") {
+                        // Static columns don't have :start nor :end columns… unless they exist
+                        // both temporally and statically, something which is legal in Rerun, and
+                        // will end up with a final Sorbet schema that declares those column as
+                        // static (which is correct, since static overrides everything else), even
+                        // though there are still traces of temporal data for that same column!
+                        _ = rrd_manifest_fields.remove(&column_name);
+                        continue;
+                    }
```
teh-cmc, line 523: The actual fix -- everything else is just bonus while I was in there.

PR #12116, opened 2025-12-05
Add tests for various ways `target_frame` can be set

Wumpf:
### Related

* follow-up to https://github.com/rerun-io/rerun/pull/12115

### What

Tries to cover all ways how `target_frame` can be set.

### DIFF
--- a/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
+++ b/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
@@ -568,7 +568,7 @@ mod tests {
                 RecommendedView::new_single_entity("some/path"),
             ))
         });
-        let _view_id_some_path_overriden =
+        let _view_id_some_path_overridden =
             test_context.setup_viewport_blueprint(|ctx, blueprint| {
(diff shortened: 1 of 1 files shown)


crates/viewer/re_viewport_blueprint/src/view_contents.rs
```diff
@@ -383,8 +383,6 @@ impl QueryExpressionEvaluator<'_> {
             return None;
         }
 
-        // TODO(jleibs): If this space is disconnected, we should terminate here
```
Wumpf, line 386: drive by while reading stuff there. this comment no longer applies since we no longer have the concept of disconnected spaces

crates/viewer/re_view_spatial/Cargo.toml
```diff
@@ -71,6 +71,7 @@ vec1 = { workspace = true, features = ["smallvec-v1"] }
 re_component_ui.workspace = true
 re_test_context.workspace = true
 re_test_viewport.workspace = true
+re_viewport.workspace = true
```
Wumpf, line 74: was already transitively in

PR #12118, opened 2025-12-05
Move `DataSourceMessage` from `re_log_types` to `re_log_channel`

emilk:
* Part of https://linear.app/rerun/issue/RR-3131/clarify-inter-re-crate-dependencies

It makes much more sense in there, since it is the channel that uses it.


crates/store/re_log_channel/Cargo.toml
```diff
@@ -21,6 +21,7 @@ all-features = true
 [dependencies]
 re_log_types.workspace = true
 re_tracing.workspace = true
+re_types_core.workspace = true
```
Wumpf, line 24: (we already had `re_log_types` here, so this won't make anything worse)

Wumpf (approved):

PR #12119, opened 2025-12-05
Improve pinhole lookup error message

Wumpf:
### Related

* Fixes RR-3086

### What

also remove some strange redundant lookup that slipped in.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -9,6 +9,7 @@
 # RustRover generated files
 misc.xml
 prettier.xml
+rust.xml
 # We may want to share some inspection profiles, but by default ignore
 inspectionProfiles/*xml
 


crates/viewer/re_view_spatial/src/visualizers/cameras.rs
```diff
@@ -77,10 +77,11 @@ impl CamerasVisualizer {
         let Some(pinhole_tree_root_info) = transforms.pinhole_tree_root_info(pinhole_frame_id)
         else {
             // This implies that the transform context didn't see the pinhole transform.
-            // Should be impossible!
-            return Err(
-                "Transform context didn't register the pinhole transform, but `CamerasVisualizer` is trying to display it!".to_owned(),
-            );
+            // This can happen with various frame id mismatches.
+            return Err(format!(
+                "Expected well-defined pinhole transforming into the transform frame {:?}, but look-up in active transform tree failed.",
```
IsseW: This error still feels pretty unspecific, but maybe we don't have more context to give here?

Wumpf: hmmmmmm agreed. Having a hard time to make it better tbh.
One case I encountered that triggers this is some kinds malformed pinholes... lemme dig a bit more first, I'll come back with a new suggestion afterwards.

PR #12127, opened 2025-12-05
Better target frame fallback

IsseW:
### Related

* Closes RR-3079

### What

Improves the target frame fallback.

The current fallback is now:
1. Check if the space root has a defined coordinate frame.
2. (new) Check if all coordinate frames logged on entities included in the filter share the same root frame, if so use that frame.
3. Use the implicit frame for the space root.

### DIFF
--- a/crates/viewer/re_view_spatial/src/shared_fallbacks.rs
+++ b/crates/viewer/re_view_spatial/src/shared_fallbacks.rs
@@ -201,11 +201,11 @@ pub fn register_fallbacks(system_registry: &mut re_viewer_context::ViewSystemReg
                         return true;
                     };
 
-                    if let Some(root) = found_root {
-                        if root != node_root_frame.root {
-                            found_root = None;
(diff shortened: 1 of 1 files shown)


crates/viewer/re_view_spatial/src/shared_fallbacks.rs
```diff
@@ -151,6 +157,71 @@ pub fn register_fallbacks(system_registry: &mut re_viewer_context::ViewSystemReg
...
+                        })
+                    else {
+                        return true;
+                    };
+
+                    if let Some(root) = found_root {
+                        if root != node_root_frame.root {
+                            found_root = None;
+                            multiple_roots = true;
+                        }
```
MichaelGrupp: Can use an if-let-chain here:

```suggestion
                    if let Some(root) = found_root
                        && root != node_root_frame.root
                    {
                        found_root = None;
                        multiple_roots = true;
```

PR #12128, opened 2025-12-05
Rework notebook auth to use device code flow

jprochazk:
```py
rr.login()
```
```
Open this URL in your browser:
...
You should see this code:
ASDF-1234
```

### DIFF
--- a/rerun_py/rerun_sdk/rerun/auth.py
+++ b/rerun_py/rerun_sdk/rerun/auth.py
@@ -30,4 +30,5 @@ def login() -> None:
 
     # Wait for the flow to complete and store credentials
     credentials = flow.finish_login_flow()
+    print()
     print(f"Success! You're logged in as '{credentials.user_email}'.")


crates/utils/re_auth/src/oauth/api.rs
```diff
@@ -287,3 +321,93 @@ impl From<User> for crate::oauth::User {
...
+    type Res = AuthenticateWithDeviceCodeResponse;
+
+    const ALLOW_4XX: bool = true;
+
+    fn into_request(self) -> Result<ehttp::Request, Error> {
+        ehttp::Request::json(
+            format_args!("{base}/user_management/authenticate", base = *WORKOS_API,),
+            &self,
+        )
+        .map_err(Error::Serialize)
```
aedm, line 411: Neat!

crates/utils/re_auth/src/oauth/login_flow.rs
```diff
@@ -87,3 +95,132 @@ impl OauthLoginFlow {
...
+    pub fn get_login_url(&self) -> &str {
+        &self.verification_uri
+    }
+
+    pub fn get_user_code(&self) -> &str {
+        &self.user_code
+    }
+
+    pub async fn wait_for_user_confirmation(&mut self) -> Result<Credentials, Error> {
+        loop {
```
jleibs, line 167: If a user never finishes the flow and it just gets stuck in the pending state, does this eventually timeout or are we stuck here forever?

Should `wait_for_user_confirmation` maybe take a user-configurable timeout?

aedm (approved):
That was fast! :)

rerun_py/rerun_sdk/rerun/auth.py
```diff
@@ -22,10 +23,12 @@ def login() -> None:
 
     # Automatically open the login URL in the default browser
     login_url = flow.login_url()
-    print("Open the following URL in your browser:")
-    print(login_url)
-    webbrowser.open(login_url)
+    print("Click the button to login in your browser:")
+    display(HTML(f'<a href="{login_url}" target="_blank" style="">Login</a>'))
```
jleibs: This makes sense for notebook, but I assume this might have issues in a normal CLI application context?

crates/utils/re_auth/src/oauth/login_flow.rs
jprochazk: The device code has an expiration of 5 minutes, after that it will error out.

rerun_py/rerun_sdk/rerun/auth.py
jprochazk: If we want this to work in any SDK context, then yes. I can remove this and leave the plain URL in its place instead.

jprochazk: I would really prefer this to clear the output in a notebook context, though...

jleibs: Confirmed:
```
>>> import rerun as rr
DEV ENVIRONMENT DETECTED! Re-importing rerun from: /home/jleibs/rerun/rerun_py/rerun_sdk
Telemetry disabled. All logging will be ignored.
>>> rr.login()
Click the button to login in your browser:
<IPython.core.display.HTML object>
You should see the following code:
FRRC-RVTB
```

crates/utils/re_auth/src/oauth/login_flow.rs
jleibs: Ha, issue related to exactly this, given the below timeout issue.

Ctrl-C doesn't seem to be getting me out of this.
```
>>> rr.login()
Click the button to login in your browser:
<IPython.core.display.HTML object>
You should see the following code:
FRRC-RVTB
^C
^C^C^C
^C^C
```

jprochazk: It should now. I'm explicitly handling it, but this is not something we normally do. You may run into a similar issue with a `CatalogClient` timeout

rerun_py/rerun_sdk/rerun/auth.py
jprochazk: I decided to simplify the output, it now uses just `print` to ensure it works in normal `python`, `ipython`, and `jupyter notebook` environments.

PR #11855, opened 2025-11-11
Auth for Jupyter Notebooks

aedm:
### Related

Related to https://linear.app/rerun/issue/RR-2861/authentication-in-jupyter-notebooks

### What

Implements authentication in Jupyter notebooks.

- By running ``rr.login()`` the user can log in.
- Uses `re_auth` module, starts a web server in the background to listen to events.
- Automatically passes credentials to the wasm viewer.

PR #12129, opened 2025-12-05
Add python linter check for possibly-undefined and fix existing errors

aedm:
### What

Fixes notebook viewer crash when `REDAP_TOKEN` is set.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -34792,7 +34792,7 @@ packages:
 - pypi: ./rerun_py
   name: rerun-sdk
   version: 0.28.0a1+dev
-  sha256: 3d5a4a63e6473c1480b88832f2b32fa2303bab1a52eed9a136f38dbb18740f41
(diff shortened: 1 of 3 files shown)


ntjohnson1:
Where did this come from and why did our static tests not catch this? ruff or mypy should have screamed

aedm:
> Where did this come from and why did our static tests not catch this? ruff or mypy should have screamed

https://github.com/rerun-io/rerun/pull/11855

CI was green at merge as far as I can tell.

ntjohnson1:
Hope you don't mind I pushed a setting to enable our type checking to catch this and fixed the other offenders.

jleibs:
This looks like it's already included in https://github.com/rerun-io/rerun/pull/12128/files#diff-a54e02161d57616ec26ebfdd3035f0ba902fd31b1abd3adf5462a0a364cd5596

PR #12131, opened 2025-12-05
Create parent dir before storing credentials

jprochazk:


crates/utils/re_auth/src/oauth/storage.rs
```diff
@@ -68,6 +68,7 @@ mod file {
     pub fn store(credentials: &Credentials) -> Result<(), StoreError> {
         let path = credentials_path().ok_or(StoreError::UnknownConfigLocation)?;
         let data = serde_json::to_string_pretty(credentials)?;
+        std::fs::create_dir_all(path.parent().expect("config not in root"))?;
```
jleibs: Any reason a bad config somewhere would cause this to be violated?

It still seems like printing a warning and just no storing the credentials would be preferable to a panic.

jprochazk: So at this point we're already past `credentials_path`, so no, this should never actually panic. It'll always be at least `rerun/credentials.json`, which has a parent. But it's not much effort to not panic here, so I'll do that instead

PR #12132, opened 2025-12-05
Allow tables to be created without providing a URL

swallez:
Fixes [RR-3012](https://linear.app/rerun/issue/RR-3012/default-write-location-for-create-table) 

The `url` parameter in `create_table_entry` is now optional. When it's `None`, the OSS server will create the Lance table in a temporary directory that will be deleted when the process exists.

The path used is `$TMPDIR/rerun-data-<xxx>/lance-<some-tuid>`, e.g. `$TMPDIR/rerun-data-jYteWy/lance-187E63126233FA3A3f7ce35ea997f9c8`

### DIFF
--- a/scripts/ci/rust_checks.py
+++ b/scripts/ci/rust_checks.py
@@ -238,7 +238,7 @@ def cargo_deny(results: list[Result]) -> None:
     # Note: running just `cargo deny check` without a `--target` can result in
     # false positives due to https://github.com/EmbarkStudios/cargo-deny/issues/324
     # Installing is quite quick if it's already installed.
-    results.append(run_cargo("install", "--locked cargo-deny@^0.17"))
+    results.append(run_cargo("install", "--locked cargo-deny@^0.18"))
 
     for target in deny_targets:
         results.append(


crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -1478,7 +1492,27 @@ impl RerunCloudService for RerunCloudHandler {
 
         let schema = Arc::new(request.schema);
 
-        let table = match &request.provider_details {
+        let details = if let Some(details) = request.provider_details {
+            details
+        } else {
```
abey79, line 1528: nit: this looks like a `unwrap_or_else(|| {})

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
```diff
@@ -290,7 +290,7 @@ def register_table(self, name: str, url: str) -> TableEntry:
 
         return TableEntry(self._internal.register_table(name, url))
 
-    def create_table(self, name: str, schema: pa.Schema, url: str) -> TableEntry:
+    def create_table(self, name: str, schema: pa.Schema, url: str | None) -> TableEntry:
```
abey79: This should IMO be:
```suggestion
    def create_table(self, name: str, schema: pa.Schema, url: str | None = None) -> TableEntry:
```

cc @jleibs

rerun_py/tests/e2e_redap_tests/conftest.py
```diff
@@ -206,7 +206,7 @@ def create_dataset(self, name: str) -> DatasetEntry:
         self._created_entries.append(entry)
         return entry
 
-    def create_table(self, name: str, schema: pa.Schema, url: str) -> TableEntry:
+    def create_table(self, name: str, schema: pa.Schema, url: str | None) -> TableEntry:
```
abey79:
```suggestion
    def create_table(self, name: str, schema: pa.Schema, url: str | None = None) -> TableEntry:
```

abey79 (approved):
This looks good to me.

The big missing change is that the e2e_redap_tests should be updated to not use an explicit table path for most tests. This should allow us to remove the `creates_table` marker. There may be just 1-2 tests to actually verify that passing something else than `None` works, but these can be `local_only` (and maybe a version of it that is `cloud_only`). Still ✅ since that could happen in a separate PR.

rerun_py/tests/e2e_redap_tests/conftest.py
swallez: Good point. Indeed the goal is you have `url` out of the way if you don't want to set it explicitly.

crates/store/re_server/src/rerun_cloud.rs
swallez: Indeed. Now I used this approach because of the `url::Url::from_directory_path(...)?` in the `else` branch: exiting the enclosing function when an error happens in the closure would be more convoluted than this if/else…

PR #12133, opened 2025-12-05
We do not show the analytics disclaimer if a user is authenticating with a rerun account

jleibs:
### What

Without this, any time a user wants to create an authenticated session in a clean notebook environment, the experience ends up looking like.
image


ntjohnson1 (approved):
We probably need to have a longer discussion about this (or maybe this is already scoped somewhere else). I would imagine we get to a point where users select different levels of acceptable analytics and that we parse that from their login details.

PR #12135, opened 2025-12-07
Catalog API update 4: schema

abey79:
### Related

### What

Part of a set of PRs to apply the changes planned in the api_sandbox to the actual API.

This PR focuses on `Schema`:
- it's now a python wrapper over a rust internal object
- its moved, along with selectors and descriptors, to the `rerun.catalog` package (from `rerun.dataframe`, which is slated to disappear)
- should be backward compatible with existing usage of `rerun.dataframe`

Migration guide notes:

````md
## Python SDK: Schema and column types moved to `rerun.catalog`

The `Schema` class and related column descriptor/selector types have moved from `rerun.dataframe` to `rerun.catalog`.

| Old import (0.27)                                       | New import (0.28)                                      |
|---------------------------------------------------------|--------------------------------------------------------|
| `from rerun.dataframe import Schema`                    | `from rerun.catalog import Schema`                     |
| `from rerun.dataframe import ComponentColumnDescriptor` | `from rerun.catalog import ComponentColumnDescriptor`  |
| `from rerun.dataframe import ComponentColumnSelector`   | `from rerun.catalog import ComponentColumnSelector`    |
| `from rerun.dataframe import IndexColumnDescriptor`     | `from rerun.catalog import IndexColumnDescriptor`      |
| `from rerun.dataframe import IndexColumnSelector`       | `from rerun.catalog import IndexColumnSelector`        |

The previous import paths are still supported but will be removed in a future release.
````

### DIFF
--- a/rerun_py/rerun_sdk/rerun/catalog/_schema.py
+++ b/rerun_py/rerun_sdk/rerun/catalog/_schema.py
@@ -19,9 +19,6 @@ class Schema:
     The schema representing a set of available columns for a dataset.
 
     A schema contains both index columns (timelines) and component columns (entity/component data).
-
-    This class wraps the internal schema representation and provides a Pythonic interface
-    for inspecting the available columns in a dataset.
     """
 
     def __init__(self, inner: SchemaInternal) -> None:

### CODE REVIEW

File: rerun_py/rerun_bindings/rerun_bindings.pyi [hunk 1]
```diff
@@ -1298,7 +1253,7 @@ class DatasetEntryInternal:
 
     @property
     def manifest_url(self) -> str: ...
-    def schema(self) -> Schema: ...
+    def schema(self) -> SchemaInternal: ...
```
jleibs, line 1256: This should be Schema now -- the fact that nothing caught this is curious.

rerun_py/rerun_sdk/rerun/catalog/_schema.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class Schema:
+    """
+    The schema representing a set of available columns for a dataset.
+
+    A schema contains both index columns (timelines) and component columns (entity/component data).
+
+    This class wraps the internal schema representation and provides a Pythonic interface
+    for inspecting the available columns in a dataset.
```
jleibs: Is this part of the docstring important for users? This seems like an implementation detail.

rerun_py/rerun_bindings/rerun_bindings.pyi
abey79: This was actually correct. This is `DatasetEntryInternal`, which _does_ return a `SchemaInternal`, and mypy had the right to be happy about it.

We mixed that up with `Recording` and `Recording`. These used to, and continue to, return a `Schema`, which used to be rust, and is now the python wrapper (thus no diff in the .pyi)

rerun_py/rerun_sdk/rerun/catalog/_schema.py
abey79: removed

PR #12144, opened 2025-12-09
Refactor `VisualizerQueryInfo` and add `AnyPhysicalDatatypeRequirement`

grtlr:
### Related

* Part of RR-3151.

### What

Enhances `VisualizerQueryInfo` to allow queuing by any physical (Arrow) data type.


Wumpf (approved):
very nice, I like the abstraction level better regardless of the new exciting `Requirement::AnyPhysicalDatatype` :)

PR #12145, opened 2025-12-09
Rename MCAP timelines to avoid conflicts with SDK timestamps

oxkitsune:
### Related

* Closes RR-3056

### What

The MCAP loader previously exposed `publish_time` and `log_time` directly as timeline names. This caused issues when `log_time` was encoded as a duration instead of an absolute timestamp, conflicting with the logging sdk's requirement that the `log_time` timeline contains timestamps.

This PR renames timelines created by the MCAP loader to `message_publish_time` and `message_log_time` in order to avoid this conflict.

### DIFF
--- a/docs/content/howto/mcap.md
+++ b/docs/content/howto/mcap.md
@@ -42,7 +42,7 @@ To map MCAP messages to Rerun entities we make the following assumptions:
 * MCAP topics corresponds to Rerun entities.
 * Messages from the same topic within an MCAP chunk will be placed into a corresponding [Rerun chunk](../concepts/chunks.md).
 * The contents of an MCAP message will be extracted to Rerun components and grouped under a corresponding Rerun archetype.
-* `log_time` and `publish_time` of an MCAP message will be carried over to Rerun as two distinct [timelines](../concepts/timelines.md).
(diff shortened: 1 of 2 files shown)


oxkitsune:
Perhaps `msg_log_time`/`msg_publish_time` is better? or `mcap_log_time`/`mcap_publish_time`?

grtlr (approved):
Thank you! I think I like the long version, it does look a bit cleaner then `msg_` and might be more general than `mcap_`.

crates/store/re_mcap/src/layers/mod.rs
```diff
@@ -124,8 +124,8 @@ impl McapChunkDecoder {
         let log_time_cell = crate::util::TimestampCell::guess_from_nanos(msg.log_time);
         let publish_time_cell = crate::util::TimestampCell::guess_from_nanos(msg.publish_time);
         let timepoint = re_chunk::TimePoint::from([
-            ("log_time", log_time_cell.into_time_cell()),
-            ("publish_time", publish_time_cell.into_time_cell()),
+            ("message_log_time", log_time_cell.into_time_cell()),
```
ntjohnson1, line 127: Can we also update our doc page https://rerun.io/docs/reference/mcap/message-formats and add a note to the migration doc?

PR #12154, opened 2025-12-10
Replace time trimming urls with time selection

IsseW:
### Related

- Part of RR-3105

### What

In preparation for larger than ram, change time trimming to just copying time selection.

### Notes

I moved time selection to fragment now, as it doesn't affect the query anymore. Now we have two possible things that say which timeline we should be on. So I think we should maybe separate out timeline from the `when` and `time_selection` fragments into a new `timeline` fragment.

TODO:
- [x] Test web viewer url


crates/viewer/re_viewer/src/app.rs
```diff
@@ -1413,14 +1411,24 @@ impl App {
                 .send_system(SystemCommand::set_selection(item.clone()));
         }
 
+        let mut time_commands = Vec::new();
+        if let Some(time_selection) = time_selection {
+            time_commands.push(TimeControlCommand::SetActiveTimeline(
+                *time_selection.timeline.name(),
+            ));
+            time_commands.push(TimeControlCommand::SetTimeSelection(time_selection.range));
```
emilk, line 1437: I think we should also turn on looping of the time selection, to further highlight "this is what the link wanted you to see"

PR #12157, opened 2025-12-10
Improve instance hover card for long instance paths

MichaelGrupp:
Use the same type of label that's also used in the visualizers' tooltips, which gets shortened if needed.

Changes the label type to non-interactive in tooltips, so that the tooltip isn't sticky.

### DIFF
--- a/crates/viewer/re_view_spatial/tests/snapshots/annotations_hover_rect_green.png
+++ b/crates/viewer/re_view_spatial/tests/snapshots/annotations_hover_rect_green.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:5bd2344eef728ddbd9176ad817300bb2f8386f6faae9d229fe54218a33e417f2
-size 20447
+oid sha256:d200b527cf71db3af38fc1d24ef0d064e7af026fc2e511f1755c7b33d2ebb0ef
+size 20379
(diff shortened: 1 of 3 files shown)


crates/viewer/re_data_ui/src/item_ui.rs
```diff
@@ -567,15 +588,15 @@ pub fn instance_hover_card_ui(
...
-    });
+    instance_path_selectable_label(
+        ui,
+        false,
+        instance_path.syntax_highlighted(ui.style()),
+        query,
+        db,
+        instance_path,
+        true,
+    );
```
MichaelGrupp, line 599: ⬆️ This was the part that was causing long labels.

crates/viewer/re_ui/src/ui_ext.rs
```diff
@@ -771,25 +771,35 @@ pub trait UiExt {
                 )
             });
 
-            ui.is_enabled() && (enough_space || long_enough_text) && contains_pointer
+            ui.is_enabled()
+                && !ui.is_tooltip()
+                && (enough_space || long_enough_text)
+                && contains_pointer
```
MichaelGrupp, line 777: (never show copy button in tooltips)

MichaelGrupp:
This is pretty much resolved by #12453

PR #12159, opened 2025-12-10
e2e_tests: migrate droid  benchmarks

andrea-reale:
### Why
We are consodlidating `redap-component-tests` into `e2e_redap_tests`.

### What
The first step is moving the droid group into `e2e_redap_tests`. This change:
- Moves the tests and encodes them as `pytest-benchmark`s 
- Adds CLI options and fixtures for the tests
- Adds markers to select the tests (`cloud_only` and `slow`)

### What next
- Add telemetry (metrics, traces)
- Port the flask service for on demand services
- Missing `vecbench` group

### DIFF
--- a/.github/workflows/reusable_test_wheels.yml
+++ b/.github/workflows/reusable_test_wheels.yml
@@ -177,11 +177,11 @@ jobs:
       - name: Run unit tests (with linux arm64 opencv workaround)
         if: ${{ inputs.PLATFORM == 'linux-arm64' }}
         # Workaround for OpenCV TLS issue, see https://github.com/opencv/opencv/issues/14884#issuecomment-815632861
-        run: cd rerun_py/tests && LD_PRELOAD=/usr/lib/aarch64-linux-gnu/libgomp.so.1 pixi run -e wheel-test-min pytest -c ../pyproject.toml
+        run: cd rerun_py/tests && LD_PRELOAD=/usr/lib/aarch64-linux-gnu/libgomp.so.1 pixi run -e wheel-test-min pytest -c ../pyproject.toml -m "not cloud_only"
 
       - name: Run unit tests
         if: ${{ inputs.PLATFORM != 'linux-arm64' }}
-        run: cd rerun_py/tests && pixi run -e wheel-test-min pytest -c ../pyproject.toml
+        run: cd rerun_py/tests && pixi run -e wheel-test-min pytest -c ../pyproject.toml -m "not cloud_only"
 
       - name: Run e2e test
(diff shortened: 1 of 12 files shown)


rerun_py/tests/e2e_redap_tests/droid/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+
+import pytest
+from rerun.catalog import CatalogClient, DatasetEntry
+
+logger = logging.getLogger(__name__)
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from rerun.catalog import CatalogClient, DatasetEntry
```
abey79, line 16: nit

```suggestion
if TYPE_CHECKING:
    from collections.abc import Iterator

    from rerun.catalog import CatalogClient, DatasetEntry


logger = logging.getLogger(__name__)
```

.github/workflows/reusable_test_wheels.yml
```diff
@@ -177,11 +177,11 @@ jobs:
       - name: Run unit tests (with linux arm64 opencv workaround)
         if: ${{ inputs.PLATFORM == 'linux-arm64' }}
         # Workaround for OpenCV TLS issue, see https://github.com/opencv/opencv/issues/14884#issuecomment-815632861
-        run: cd rerun_py/tests && LD_PRELOAD=/usr/lib/aarch64-linux-gnu/libgomp.so.1 pixi run -e wheel-test-min pytest -c ../pyproject.toml
+        run: cd rerun_py/tests && LD_PRELOAD=/usr/lib/aarch64-linux-gnu/libgomp.so.1 pixi run -e wheel-test-min pytest -c ../pyproject.toml -m "not cloud_only"
```
abey79, line 180: is this needed since it's disabled by default? Also, do we want to force opt-in to `slow`? (genuine question, not sure it makes sense here)

rerun_py/tests/e2e_redap_tests/conftest.py
```diff
@@ -11,41 +11,67 @@
...
+        default="droid:sample50",
+        help="Dataset for Droid benchmarks (options: droid:raw, droid:sample50, etc.)",
+    )
+
+    droid_group.addoption(
+        "--droid-preregister-dataset",
+        action="store_true",
+        default=False,
+        help="should the test runner register the droid dataset before running tests",
     )
```
abey79, line 75: It would be slightly more readable if the `adoption` code was colocated with the `getoption` fixtures. Would it make sense to move that to the other conftest.py?

abey79 (approved):
great work!

PR #12162, opened 2025-12-10
Implement streaming for datafusion table

lucasmerlin:
### What

This implements streaming in the `DataFusionTableWidget` in order to speed up loading of huge tables.


crates/viewer/re_dataframe_ui/src/datafusion_adapter.rs
```diff
@@ -191,41 +194,92 @@ impl DataFusionQuery {
         // Collect record batches
```
timsaucer: stale comment, but can probably just remove anyway

PR #12163, opened 2025-12-10
Fix links to time points

emilk:
It worked _sometimes_, depending on wether or not we had already materialized a blueprint, something we do lazily (as late as possible = too late).


IsseW:
Tested on web viewer as well.

PR #12167, opened 2025-12-10
Viewer open url with fragment regression test

IsseW:
### Related

* Test for the regression fixed in https://github.com/rerun-io/rerun/pull/12163


emilk (approved):
Awesome!

PR #12171, opened 2025-12-11
RRD manifests: test and fix native map converters

teh-cmc:
Title.


crates/store/re_log_encoding/src/rrd/footer/instances.rs
```diff
@@ -442,8 +443,9 @@ impl RrdManifest {
                 .unwrap_or_else(|| NullBuffer::new_valid(col_end.len()));
 
             columns_per_index.insert(
-                index.to_owned(),
+                field.name().to_owned(),
```
teh-cmc, line 446: the actual fix

PR #12176, opened 2025-12-11
Switch to uv-based management of our python build+test environment

jleibs:
### What

Using our pixi environment for working with the `rerun-sdk` itself has led to all manner of pain.  In theory pixi-build may fix this some day, but it's still not optimized for an explicit maturin develop dev flow.

We move the entire "rerun-sdk" test-oriented environment over to uv, removing all examples and dedicated py-related environments from pixi.

### Building and running python code
BEFORE:
```
pixi run py-build
pixi run -e py ./my_script.py
```

NOW:
```
pixi run py-build

# Run through pixi if you don't have uv
pixi run uv ./my_script.py

# Or just use uv directly if you have it installed
uv run ./my_script.py
```


jleibs:
@rerun-bot full-check

jleibs:
@rerun-bot full-check

jleibs:
@rerun-bot full-check

jleibs:
@rerun-bot full-check

jleibs:
@rerun-bot full-check

crates/build/re_dev_tools/src/build_examples/snippets.rs
```diff
@@ -17,8 +17,26 @@ pub struct Snippets {
     output_dir: PathBuf,
 }
 
+fn install_snippet_deps() {
+    // uv sync --inexact --no-install-package rerun-sdk --group snippets
```
ntjohnson1, line 21: This selective inexact usage feels a little footgunny/susceptible to bit rot. I guess if we miss an inexact somewhere then rerun won't be in the environment and we hopefully fail early.

docs/snippets/compare_snippet_output.py
```diff
@@ -118,10 +118,17 @@ def main() -> None:
...
-        build_python_sdk(build_env)
+        if args.no_py_build:
+            print("Skipping building python rerun-sdk - assuming it is already built and up-to-date!")
+        else:
+            build_python_sdk(build_env)
+        # Use uv to install the snippet dependencies
+        run(
+            ["uv", "sync", "--group", "snippets", "--inexact", "--no-install-package", "rerun-sdk"],
+            env=build_env,
+            timeout=60,
```
ntjohnson1: Timeout magic number feels like it might lead to flaky CI. I'm not sure how long this typically takes so maybe this is already 10x of normal, in which case should be fine.

examples/manifest.toml
```diff
@@ -193,4 +191,8 @@ examples = [
   "table_zoo",
   "template",
   "viewer_callbacks",
+
+  # TODO(jleibs): We want to bring them back once opentelemtry is fixed
```
ntjohnson1, line 192: Is there an issue for this? Then zombie script can this.

pixi.toml
```diff
@@ -300,24 +270,26 @@ nb-strip = "python scripts/nbstripout.py --directories mypy examples/ scripts/ t
...
 # Get non-internal things that are ok to not check rerun on
-py-lint-non-sdk = { cmd = "mypy --config-file rerun_py/.non_sdk_mypy.ini --install-types --non-interactive --no-warn-unused-ignore --cache-dir .mypy_cache_rerun_other", env = { PYTHONPATH = "$PYTHONPATH:rerun_py/rerun_sdk/" } }
+py-lint-non-sdk = { cmd = "uv run mypy --config-file rerun_py/.non_sdk_mypy.ini --no-warn-unused-ignore --cache-dir .mypy_cache_rerun_other" }
 # I couldn't get the config to work correctly, so placing relevant directories in command line
 # Until we resolve nbqa config issue keep aligned with nb-strip dirs below
-py-lint-nb-rerun = { cmd = "nbqa mypy examples/notebook scripts/ tests/python/ docs/snippets/ --config rerun_py/pyproject.toml --cache-dir .mypy_cache_rerun_nb ", env = { PYTHONPATH = "$PYTHONPATH:rerun_py/rerun_sdk/" } }
+# MYPYPATH is set to allow mypy to find rerun and rerun_bindings without needing maturin develop.
+py-lint-nb-rerun = { cmd = "MYPYPATH=rerun_py:rerun_py/rerun_sdk:rerun_notebook/src uv run nbqa mypy --no-warn-unused-ignore examples/notebook scripts/ tests/python/ docs/snippets/ --cache-dir .mypy_cache_rerun_nb " }
```
ntjohnson1: NIT: the command in pixi allows an env argument rather than needing to just prefix it to the command. See line above

ntjohnson1: Same note throughout since I think it makes the already potentially complicated commands a little clearer

pixi.toml
```diff
@@ -373,125 +345,78 @@ js-docs = { cmd = "yarn --cwd rerun_js run docs", depends-on = ["js-build-packag
...
+py-build = "env -u CONDA_PREFIX RERUN_MATURIN_BUILD=1 RERUN_ALLOW_MISSING_BIN=1 uv run maturin develop --uv --manifest-path rerun_py/Cargo.toml --extras=tests,datafusion && uvpy scripts/ci/fix_rerun_pth.py"
+# Build the `rerun-notebook` package.
+py-build-notebook = { cmd = "uv sync --inexact --package rerun-notebook", depends-on = [
+  "js-build-base",
 ] }
-
-# Build and install the `rerun-sdk` package with the `web_viewer` & `server` feature.
-py-build-common-web-viewer = { cmd = "PIP_REQUIRE_VIRTUALENV=0 RERUN_ALLOW_MISSING_BIN=1 maturin develop --manifest-path rerun_py/Cargo.toml --features web_viewer,nasm,server --extras=tests", depends-on = [
-  "rerun-build-native-and-web", # We need to build rerun-cli since it is bundled in the python package.
+py-build-notebook-fast = { cmd = "uv sync --inexact --package rerun-notebook" }
```
ntjohnson1, line 366: Fast just allows skipping the JavaScript pieces?

scripts/ci/uv_install_wheel.py
```diff
@@ -0,0 +1,112 @@
...
+        wheels = [whl for whl in wheels if arch in whl]
+
+    if len(wheels) == 0:
+        if platform_independent:
+            print(f"No wheels found for package {pkg} (the package was expected to be platform independent)")
+        else:
+            print(f"No wheels found for package {pkg} on platform {plat} and architecture {arch}")
+        sys.exit(1)
+
+    if len(wheels) > 1:
```
ntjohnson1, line 69: I think this is overly strict. Did you try bumping the version and seeing if this barfs? For instance if we're on 0.27 then go to 0.28 we'd have duplicate wheels in our wheels dir but unique by version

timsaucer:
Something I ran into that I suspect others who frequently use uv/maturin will also is that I cannot just `source .venv/bin/activate` and then do `maturin develop --uv` from the `rerun_py` directory. I have to pass in that additional `RERUN_MATURIN_BUILD` environment variable and also run the `fix_rerun_pth.py`. I see that the `pixi run py-build` command will do those for me, but that means that even though I have this `uv` environment I can't just use it after setting up. I still need to go through `pixi`.

Maybe it's worth another look at the `fix_rerun_pth.py`. I don't have to do this in `datafusion-python`. There running `maturin develop --uv` just works.

So I guess what I'm finding is that this does work similar to what I had before except that there is a `uv` environment but it doesn't work like I would expect it to. I'll keep playing around with it on my flight.

jleibs:
We are going to hold off merging this until after 0.28 has been released to avoid debugging release-scripts during the release.

docs/snippets/compare_snippet_output.py
jleibs: Good call -- no reason for a timeout here.

examples/manifest.toml
jleibs: Good call. Created one: https://linear.app/rerun/issue/RR-3223/opentelemetry-package-conflicts-with-some-examples

scripts/ci/uv_install_wheel.py
jleibs: Historically this has only been used in CI where there will be a single wheel and the main purpose of this was to avoid needing to pick up a dependency on the version.  This is a direct port of the other pixi-oriented version of the same script. I'm going to leave it for now but flag it as a thing I clean up in the pending script cleanup.

jleibs:
@rerun-bot full-check

PR #12180, opened 2025-12-11
Update to egui 0.33.3

emilk:
* https://github.com/emilk/egui/releases/tag/0.33.3
* Full egui diff: https://github.com/emilk/egui/compare/0.33.2...0.33.3

This should make it much easier to updade snapshots with kitdiff: a single run should now update _all_ images.

Also fixes the text color of selected text:

image

Screenshot 2025-12-12 at 09 01 45


emilk:
The test failure is because in one of the tests, hovering the box no longer causes the box to be highlighted.

This is because the `.hover_at` function is now in kittest proper, but with slightly different semantics: it moves the cursor, but doesn't call .run().

PR #12182, opened 2025-12-11
Codegen archetype to blueprint mapping in Python

grtlr:
### Related

* Part of RR-1189
* Part of RR-1480
* Part of RR-3173
* Related #8368

### What

Fixes a bug in our Python SDK where not all visualizers where exported for use in blueprints.

Also establishes an (almost) 1-to-1 mapping from archetypes to visualizers (when applicable) and enforces it via `codegen`. It also creates the stubs for the new blueprint API.


crates/build/re_types_builder/src/codegen/python/visualizers.rs
```diff
@@ -0,0 +1,113 @@
...
+        1,
+    );
+    code.push_indented(3, "self.visualizer_type = visualizer_type", 1);
+    code.push_indented(3, "self.overrides = overrides", 1);
+    code.push_indented(3, "self.mappings = mappings or []", 2);
+    code.push_indented(
+        2,
+        "# TODO(grtlr,andreas): Implementation of new blueprint API goes here",
+        2,
+    );
```
Wumpf: This should not be part of the codegen, we want to handwrite this directly in python

crates/build/re_types_builder/src/codegen/common.rs
```diff
@@ -451,10 +467,15 @@ pub fn remove_orphaned_files(reporter: &Reporter, files: &GeneratedFiles) {
...
             }
 
+            // Only remove files that are:
+            // 1. Not in the files map (orphaned), AND
+            // 2. Either there's no .gitattributes OR the file is listed as generated in .gitattributes
             if !files.contains_key(&filepath) {
-                re_log::info!("Removing {filepath:?}");
-                if let Err(err) = std::fs::remove_file(&filepath) {
-                    panic!("Failed to remove {filepath:?}: {err}");
+                if generated_files_in_dir.is_empty() || generated_files_in_dir.contains(&filepath) {
```
Wumpf: plz revisit as discussed in person. thanks!

Wumpf (changes requested):

PR #12187, opened 2025-12-11
Catalog API update 6: register APIs

abey79:
### Related

### What

Part of a set of PRs to apply the changes planned in the api_sandbox to the actual API.

This PR:
- removes `Task`/`Tasks`, superseded by `RegistrationHandle`
- merges `register` and `register_batch` into a single `register` api, which returns `RegistrationHandle`
- `register_prefix` now returns `RegistrationHandle` as well

As per the updated migration guide:

````md
## Python SDK: `register()` and `register_batch()` merged into unified `register()` API

The `DatasetEntry.register()` and `DatasetEntry.register_batch()` methods have been merged into a single `register()` method that returns a `RegistrationHandle`. The `DatasetEntry.register_prefix()` now also returns a `RegistrationHandle`. The `Tasks` and `Task` classes have been removed. The `recording_layer` parameter has been renamed to `layer_name`.

Single URI registration:

```python
# Before (0.27)
segment_id = dataset.register("s3://bucket/recording.rrd")

# After (0.28)
segment_id = dataset.register("s3://bucket/recording.rrd").wait()[0]
```

Batch registration:

```python
# Before (0.27)
handle = dataset.register_batch(["file:///uri1.rrd", "file:///uri2.rrd"], recording_layers=["base", "base"])
segment_ids = handle.wait()

# After (0.28)
handle = dataset.register(["file:///uri1.rrd", "file:///uri2.rrd"], layer_name="base")
segment_ids = handle.wait()
```

````

### DIFF
--- a/crates/store/re_datafusion/src/dataframe_query_provider.rs
+++ b/crates/store/re_datafusion/src/dataframe_query_provider.rs
@@ -132,7 +132,7 @@ impl Stream for DataframeSegmentStream {
 
         #[cfg(not(target_arch = "wasm32"))]
         let _trace_guard = attach_trace_context(&this.trace_headers);
-        let _span = tracing::info_span!("poll_next").entered();
+        let _span = tracing::debug_span!("poll_next").entered();
 
         // If we have any errors on the worker thread, we want to ensure we pass them up
         // through the stream.
--- a/docs/snippets/snippets.toml
+++ b/docs/snippets/snippets.toml
@@ -170,6 +170,10 @@ backwards_check = [
   "cpp", # TODO(#2353): Doesn't exist
   "rust", # Runs indefinitely.
 ]
+"howto/load_urdf" = [
+  "cpp", # Not implemented
+  "rust", # Not implemented
+]
 "howto/micro_batching" = [
   "cpp", # TODO(#10661): Doesn't exist
 ]
--- a/tests/assets/rrd/snippets/howto/load_urdf.rrd
+++ b/tests/assets/rrd/snippets/howto/load_urdf.rrd
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:d5b906bc0b59158078a712a856a2aef740e516a4c61304c6f51d09552c8f360c
+size 63773

### CODE REVIEW

File: docs/content/reference/migration/migration-0-28.md [hunk 1]
```diff
@@ -330,3 +330,29 @@ df = table.reader()
...
+The `DatasetEntry.register()` and `DatasetEntry.register_batch()` methods have been merged into a single `register()` method that returns a `RegistrationHandle`. The `DatasetEntry.register_prefix()` now also returns a `RegistrationHandle`. The `Tasks` and `Task` classes have been removed. The `recording_layer` parameter has been renamed to `layer_name`.
+
+Single URI registration:
+
+```python
+# Before (0.27)
+segment_id = dataset.register("s3://bucket/recording.rrd")
+
+# After (0.28)
+segment_id = dataset.register("s3://bucket/recording.rrd").wait()[0]
```
jleibs: I thought we proposed:
```
 dataset.register("s3://bucket/recording.rrd").wait().segments[0]
```

rerun_py/src/catalog/registration_handle.rs
```diff
@@ -0,0 +1,322 @@
...
+    common::v1alpha1::TaskId,
+};
+use tokio::sync::mpsc;
+use tracing::Instrument as _;
+
+use super::{PyCatalogClientInternal, to_py_err};
+use crate::utils::{get_tokio_runtime, wait_for_future};
+
+/// Default timeout: 8 hours.
+const DEFAULT_TIMEOUT_SECS: u64 = 8 * 60 * 60;
```
jleibs: This seems like a surpriing/arbitrary value. It's sufficiently long that for a human it's indistinguishable from a blocking call.  But, if for some reason you did need it to be anywhere near this long though, then it would be extremely annoying to have it actually time out.

Long before getting to 8 hours I would expect to need to be handling things like the client itself dying. Which means the code needs to be structured so that it's tracking some sort of "unprocessed" state and should just be repeatedly querying the server for any new work, regardless of who called register.

jleibs (approved):

docs/content/reference/migration/migration-0-28.md
abey79: 🤦🏻 completely forgot about that. will update first thing in the morning

rerun_py/src/catalog/registration_handle.rs
abey79: Let's discuss this with @andrea-reale tomorrow morning. The idea is indeed to default to "blocking", but I think we cannot use `None` at the grpc level. Also, iirc there is some server-side internal timeout of 8h. But I'm unsure what the most correct thing to do.

In any case, I've put that in rust side because ultimately its an implementation details of the `None` case exposed on python side (which should aim for blocking semantics).

andrea-reale: The server side max timeout exists to avoid clients blocking indefinitely on the call (and haging on server resources forever): the idea is that when you need to wait for very long periods of time, you should handle some long-polling loop.  

Checking the code again the current server cap is set to 1 hour: that's arbitrary, but I think it's reasonable to think that way before that time client code should anyway handle exceptions and retries. 

That said I support the idea to have a default `None` timeout in the python SDK, that mimics indefinite blocking (you don't want to think about a timeout when just playing with the API). We should probably be pretty clear in our spec that our method can raise some `TimeoutException` that needs to be retried (but it's true that as any networked call this can return any sort of IO exception).

andrea-reale:
LGTM. I'd appreciate if we can somehow grow the number of registration-specific e2e_tests to also cover a few of the corner cases we discussed (as well as having dedicated registration tests, beyond those that happen as side effects of other tests?)

andrea-reale (approved):

rerun_py/src/catalog/registration_handle.rs
abey79: My proposal:
- I change the constant to 1h
- I make it clear that `None` is not a protection against `TimeoutException` in the SDK docstring

PR #12194, opened 2025-12-12
RRD manifests: implement `chunk_byte_size_uncompressed` support

teh-cmc:
Title.


crates/store/re_log_encoding/src/rrd/encoder.rs
```diff
@@ -148,6 +148,10 @@ impl FooterState {
                 // application layer where one can accessed the parsed, unmigrated data.
                 let chunk_batch = re_sorbet::ChunkBatch::try_from(&msg.batch)?;
 
+                // Not a totally accurate value (and it's unclear what would an accurate value even
+                // look like), but that's fine: this doesn't need to be perfectly precise.
+                let byte_size_uncompressed = msg.batch.heap_size_bytes();
```
teh-cmc: Hmm... I can probably get access to the transport-level message from here with a little bit more effort, let's see...

teh-cmc:
Gonna integrate that in the larger-than-RAM sprint + make a sibling on dataplatform to ignore the new field for now.

Issue #9334, opened 2025-03-20
Prototype tag overrides & generic datatypes to override where Time Series scalars are sourced from


Wumpf:
Part of the vision of how tagged components (see https://github.com/rerun-io/rerun/issues/6889) should play out in the viewer is that each archetype field should form a "slot" that is identified by its exact tag (including archetype name, field name and component/semantic name).
In this mental model, those slots are by default attached to component columns that have this tag. However, the slot assignment can be more flexibile: In absence of data a heuristic could attach other fitting data to the visualizers input or, more importantly, a user may decide in code or ui to specify where the data comes from.

As a first step exploring this we should look at the high value usecase of flexibly assigning scalar data: Scalars are often sourced from close to arbitrary data on the fly - vectors, matrices, etc. all may transitively be plotted. Note that directly implies that the encoding type of a scalar array must be more flexible: fixed arrays and both f32 & f64 need to be supported for starters to make this powerful.

As it turns out, despite tagged components not being fully implemented in the viewer, we are already in a fairly good shape to build an experimental version of this feature. In `andreas/hacky-flexible-plot-data-input` we override the component name used for scalar queries by reading it from a override of an ad-hoc name component.
**For this prototype the main hurdle was activating visualizers in absence of required components,** since it the viewer identified as the visualizer to be not applicable to the entity. (in the prototype this was solved by unhinging this restriction which breaks a few things)

Proposed plan of action:
* model an experimentally marked tag override archetype which essentially pretend that tagged components are already fully implemented. This isn't strictly needed but allows us to explore the space
* allow logging this new archetype as an override
* read in that override in series line and series point specifically for overrides of overriding the source for scalars
* make types that are accepted on scalars more flexible (fixed size arrays & f32) - this doesn't _have_ to be reflected in logging APIs at this point, documentation only is fine

Stretch goals:
* explore possible ui for changing changing the tag override live
* explore generalization of the concept on a blueprint "processing" level
* explore formalization of component generic

Wumpf:
We got a fairly sophisticated prototype now
Expecting to land this as a feature in 0.29 likely with more improvements from there on

PR #12195, opened 2025-12-12
Improve VisualizerOverrides ergonomics in Rust (#12265) + Some of the blueprint types required for component mapping (#12215)

grtlr:
### Related

* Part of #10631.
* Part of #9334.

### What

> [!IMPORTANT]
> This is a feature branch in which we collect commits that we will cherry-pick into `main` right after the `v0.28` release.

Please use PRs to merge into this branch and ensure that CI is green to make our future lives easier.

### DIFF
--- a/crates/build/re_types_builder/src/codegen/rust/api.rs
+++ b/crates/build/re_types_builder/src/codegen/rust/api.rs
@@ -22,10 +22,10 @@ use crate::codegen::{Target, autogen_warning};
 use crate::objects::{EnumIntegerType, ObjectClass};
 use crate::{
     ATTR_DEFAULT, ATTR_RERUN_COMPONENT_OPTIONAL, ATTR_RERUN_COMPONENT_RECOMMENDED,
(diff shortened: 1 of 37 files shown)


crates/viewer/re_viewer_context/src/app_options.rs
```diff
@@ -63,6 +63,9 @@ pub struct AppOptions {
     /// see [`AppOptions::cache_subdirectory`].
     #[cfg(not(target_arch = "wasm32"))]
     pub cache_directory: Option<std::path::PathBuf>,
+
+    /// Enables experimental visualization of arbitrary scalar values.
+    pub experimental_any_scalars: bool,
```
emilk: What is the point in adding this if it doesn't control any behavior?

grtlr: We'll add more commits until we have the release and can break API.

Wumpf:
I ended up dropping github.com/rerun-io/rerun/pull/12195/changes/7a30d6f5d7cccae8cf60ffe9eaf3b91eacce22d3
-> the other two commits look like they're actually ready to land now that a release isn't immediately imminent:

They are these two PRs respectively

* https://github.com/rerun-io/rerun/pull/12265
* https://github.com/rerun-io/rerun/pull/12215

PR #12199, opened 2025-12-12
Implement pycapsule for schema

timsaucer:
This PR adds support for converting internal rerun schema to pyarrow schema.

If you have a schema object you can do

```
now_schema = pa.schema(old_schema)
```


swallez (approved):
Nice!

PR #12204, opened 2025-12-14
Remove items marked as deprecated before 0.27

ntjohnson1:
I was getting ready to mark load_recording as deprecated and was looking over how we generally manage this. Saw we mark things deprecated in a few different ways and often forget to actually delete them. Try to make a quick pass to actually remove old cruft.


crates/top/re_sdk/src/recording_stream.rs
```diff
@@ -2677,8 +2543,6 @@ impl RecordingStream {
     /// See also:
     /// - [`Self::set_timepoint`]
     /// - [`Self::set_time_sequence`]
-    /// - [`Self::set_time_secs`]
-    /// - [`Self::set_time_nanos`]
```
abey79, line 2681: worth listing the other methods here? (or alternatively remove this list that is bound to be incomplete)

rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -424,8 +423,6 @@ def close_url(
 
         self._viewer.close_url(url)
 
-    @deprecated_param("nanoseconds", use_instead="duration or timestamp", since="0.23.0")
```
abey79, line 427: TIL deprecated_param

crates/top/re_sdk/src/recording_stream.rs
ntjohnson1: I think these are the relevant ones. Generally I support not enumerating things in docs since it always ends up stale but we do it all over 🤷

abey79: sharing the sentiment.

ntjohnson1:
> Might still be worth having a LLM write a small migration guide section, since this is technically breaking.

Added a section. Seemed good enough on first pass especially since most of these things have been deprecated for ~6months.

PR #12207, opened 2025-12-15
Turn off incremental Rust compilation

emilk:
Incremental compilation blows up the size of the target folder,
but it is also buggy, causing spurious "link failures" that requires a complete clean of the `target` folder.

Note that the incremental part is for each individual crate. Even with `CARGO_INCREMENTAL` we only recompile the crates that we need to recompile (but we will recompile them in their entirety).

https://doc.rust-lang.org/cargo/reference/profiles.html#incremental

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -7,6 +7,12 @@ codegen = "run --package re_types_builder --"
 # Temporary solution while we wait for our own xtasks!
 run-wasm = "run --release --package run_wasm --"
 
+
+[build]
+# Incremental compilation blows up the size of the target folder, and is also buggy.
+incremental = false
+
+
 [env]
 # Some of our build.rs files only run if this is set,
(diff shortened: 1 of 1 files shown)


oxkitsune (approved):

teh-cmc:
FYI: There's a less hacky way than an environment variable that I use in my ~/.cargo:
```toml
[build]
rustc-wrapper = "/home/cmc/.cargo/bin/sccache"
incremental = false

[target.x86_64-unknown-linux-gnu]
linker = "clang"
rustflags = [
  "-Clink-arg=-fuse-ld=/usr/bin/mold",
  "-Csplit-debuginfo=unpacked",
  "-Ctarget-cpu=native",
  "-Cforce-unwind-tables=yes",
  "--cfg", "tokio_unstable",
]
```

PR #12208, opened 2025-12-15
Fix timepanel size being rounded down

IsseW:
### Related

- Closes RR-3174

### What

Fixes time panel size being rounded down when dragging it.

### DIFF
--- a/crates/viewer/re_view_time_series/src/view_class.rs
+++ b/crates/viewer/re_view_time_series/src/view_class.rs
@@ -692,10 +692,11 @@ impl ViewClass for TimeSeriesView {
                 }
             }
 
+            let time_float_offset_id = plot_id.with("time_float_offset");
             // Since we store the x-axis view range with integers we want to store an extra temporary float offset for the view.
             let plot_x_range = ui
                 .ctx()
-                .memory(|mem| mem.data.get_temp::<Range1D>(plot_id))
+                .memory(|mem| mem.data.get_temp::<Range1D>(time_float_offset_id))
                 .map(|offset| {
                     Range1D::new(
                         x_range.start() + offset.start(),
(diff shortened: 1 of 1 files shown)


IsseW:
> Small fix! Would .round() make even more sense?

Good point, yeah I think so.

Also realized that this issue goes a bit further. Since we, in blueprint, store the x axis view with integers it gets buggy for small views. Especially with sequence timelines. So I added a temporary float value for the offset in the x axis that gets stored in egui memory.

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -692,6 +692,18 @@ impl ViewClass for TimeSeriesView {
...
+            let plot_x_range = ui
+                .ctx()
+                .memory(|mem| mem.data.get_temp::<Range1D>(plot_id))
+                .map(|offset| {
+                    Range1D::new(
+                        x_range.start() + offset.start(),
+                        x_range.end() + offset.end(),
+                    )
+                })
+                .unwrap_or(x_range);
```
emilk, line 706: Makes sense… I guess the alternative would be to switch to the `Decimal128` arrow datatype… but that seems like a lot of work.

The downside with the PR, I guess, is that sharing a blueprint wont _exactly_ match, since the Y/time range will be rounded to the closest integer. But… 🤷 . Good enough for now.

crates/viewer/re_view_time_series/src/view_class.rs
```diff
@@ -777,17 +789,28 @@ impl ViewClass for TimeSeriesView {
 
                 if unchanged_bounds != *transform.bounds() {
                     let new_x_range = transform_axis_range(transform, 0);
+                    let new_x_range_rounded =
+                        Range1D::new(new_x_range.start().round(), new_x_range.end().round());
+                    ui.ctx().memory_mut(|m| {
+                        m.data.insert_temp(
+                            plot_id,
```
emilk: I suggest creating a new ID, mostly just for readability

`let time_float_offset_id = plot_id.with("time_float_offset");`

```suggestion
                            time_float_offset_id,
```

(don't forget to update both places!)

emilk (approved):

crates/viewer/re_view_time_series/src/view_class.rs
IsseW: > since the Y/time range will be rounded to the closest integer.

would only be x axis, as that's the one that maps to time. And yeah I think it's better to keep that as an integer for now as that's also how we handle other times on the python api side. Also means there will never be detail on the plot that's smaller than 1 time unit.

PR #12210, opened 2025-12-15
Add documentation for sinks

oxkitsune:
### What

Adds a documentation page with an explanation of our sinks concept and some snippets.

### DIFF
--- a/docs/snippets/all/concepts/file_sink.cpp
+++ b/docs/snippets/all/concepts/file_sink.cpp
@@ -5,5 +5,5 @@
 int main() {
     const auto rec = rerun::RecordingStream("rerun_example_file_sink");
 
-    rec.set_sinks(rerun::FileSink{"recording.rrd"});
+    rec.set_sinks(rerun::FileSink{"recording.rrd"}).exit_on_failure();
 }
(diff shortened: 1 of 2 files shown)


docs/snippets/snippets.toml
```diff
@@ -165,6 +165,16 @@ backwards_check = [
   "cpp",
   "rust",
 ]
+"concepts/file_sink" = [ # pseudo-code
+  "py",
+  "cpp",
+  "rust",
+]
+"concepts/grpc_sink" = [ # pseudo-code
```
emilk: There's not really pseudo-code though, are they? Looks like they at least should be type checked/compiled. Maybe they are - I don't know  (do we have a level of "compile but don't run" for snippets? We should) - but "pseudo-code" is the wrong description imho.

oxkitsune: Ah yea you're right, copy pasta'd too hard
They should be type checked/compiled but we cannot run them

emilk (approved):
Nice - thanks!

docs/snippets/snippets.toml
emilk: Seeing the CI errors it looks like they are indeed compiled (or, at least _atempted_ to be 😆 )

Issue #12206, opened 2025-12-14
Segmentation Image and Video overlay Failure


pablovela5620:
**Describe the bug**  
`SegmentationImage` with semi‑transparent class colors correctly overlays on `Image`, but when overlaid on a logged video asset the alpha is ignored and the mask renders fully opaque, hiding the video.

**To Reproduce**
1. Ensure a small mp4 is at `data/hocap-example.mp4`.
2. Run: (minimal reproduction script below)
   ```bash
   python mini_repro.py
   ```
3. In the viewer, two stacked 2D views appear (`Video` and `Image`), both on the `video_time` timeline.
4. Observe the red square overlay:
   - In the `Image` view, the video frame is visible through the square (alpha respected).  
   - In the `Video` view, the underlying video is hidden (alpha ignored).

**Expected behavior**  
Both views should respect per-class alpha from `AnnotationContext`, showing the underlying content through the mask.

**Screenshots**  
See attached: video view shows fully opaque square; image view shows translucent square.
![Image](https://github.com/user-attachments/assets/021bd9a7-87e3-404f-845d-a097c11fc968)

also attached the rrd file you can view [here](https://app.rerun.io/version/0.27.3/index.html?url=https%3A%2F%2Fhuggingface.co%2Fdatasets%2Fpablovela5620%2Frrd-examples%2Fresolve%2Fmain%2Fseg-failure.rrd%2Brbl.rrd)

**Desktop (please complete the following information):**  
- OS: Linux (Ubuntu 24.04, kernel 6.14.0-37-generic, x86_64)  

**Rerun version**  
- 0.27.3 (pixi env, rerun-sdk wheel)

**Additional context**  
- Issue appears only when the base layer is a video asset (`AssetVideo` + `VideoFrameReference`). Using `Image` for the same frames renders correctly.  
- Draw order is set via `AnyValues(draw_order=...)`; behavior is the same if omitted.

## Minimal repro (matches steps above)

```python
"""Minimal repro: SegmentationImage opacity over video asset vs per-frame images."""

from pathlib import Path

import cv2
import numpy as np
import rerun as rr
import rerun.blueprint as rrb


def main() -> None:
    video_path = Path("data/hocap-example.mp4")
    if not video_path.exists():
        raise FileNotFoundError(f"Expected video at {video_path}")

    rr.init("seg-opacity-mini-repro", spawn=True)  # local viewer

    # Annotation context with transparent background and semi-transparent FG.
    rr.log(
        "/",
[...]

Issue #12023, opened 2025-11-28
Regression of RGB & Depth overlay


makeecat:
**Describe the bug**
overlay of RGB and depth image does not work as usual

**To Reproduce**
build on the main branch (commit: 99ffdbe), check RGBD example.

**Expected behavior**
The colorized depth image should overlay on top of RGB image.

**Screenshots**

<img width="492" height="734" alt="Image" src="https://github.com/user-attachments/assets/f0388ac7-8782-4cce-b34c-d4bf124a5a91" />

**Desktop (please complete the following information):**
 - OS: macOS Tahoe 26.1

**Rerun version**
main branch commit: 99ffdbe

makeecat:
After testing recent commits, commit 8572090 cause this regression.
Related PR: #11962 

When running RGBD example, the viewer report a bug: 

"Visualizers for view Named("RGB") haven't been executed prior to display. This should never happen, please report a bug."

A potential fix is proposed in https://github.com/makeecat/rerun/tree/users/yang/fix_transparent_overlay as a proof-of-concept and a reference to help the rerun team to triangulate the issue.

@Wumpf

Issue #11156, opened 2025-09-09
Opacity for images/depth-images/videos/etc. is broken in 3D


Wumpf:
The opacity feature on the re_renderer textured rectangle predates our new ability to sort drawables which got introduced in
* https://github.com/rerun-io/rerun/pull/11131

( this is not a regression of #11131, but it affects the behavior of this bug a bit)

<img width="344" height="233" alt="Screenshot 2025-09-09 at 17 18 58" src="https://github.com/user-attachments/assets/a4d7785f-b03f-40e6-8c8e-a556bb5b5a04" />

Currently images draw always as part of the opaque phase, even when they're transparent. We have to put them in the right phase and correctly handle draw order

-> This is strongly related to
* https://github.com/rerun-io/rerun/issues/1025

PR #12214, opened 2025-12-15
Fix transparency in 2D views

emilk:
* Closes https://github.com/rerun-io/rerun/issues/12023
* Closes https://github.com/rerun-io/rerun/issues/12206
* Closes https://github.com/rerun-io/rerun/issues/11156
* Closes RR-3179
* Closes RR-3057
* Bug introduced in https://github.com/rerun-io/rerun/pull/11962

~I don't understand _why_ this PR fixes the problem, but it does.~

----

@Wumpf : reverted the original fix and applied a more forward looking fix in the re_renderer: we're now doing the same transparency handling on `re_renderer::Rectangle` (the underlying primitive) that we do on other primitives like meshes. I previously assumed that this isn't necessary because we sort the rectangle draw data, was apparently wrong either way.

The situation is still a bit weird because we still rely on depth based draw ordering - our draw order gets translated into a depth offset. With this change, the ordering for transparent objects take care of it. But drawbacks with this approach remain:
* depth offset is inaccurate and causes glitches especially on web. This is probably reduced now since we no longer draw transparent rectangles with depth (which we did before, relying on their correct ordering). See  #4787
* draw order doesn't always interact correctly between opaque and non-opaque rectangles
* draw order doesn't always interact correctly between rectangles and other things with draw order

Solution to all these problems is to have a special 2D draw phase in which we don't rely on the depth buffer at all but exclusively on an explicit draw order (which the Viewer already has for 2D objects!). Other related issues: #1025, #11156

### DIFF
--- a/crates/viewer/re_renderer/src/renderer/rectangles.rs
+++ b/crates/viewer/re_renderer/src/renderer/rectangles.rs
@@ -421,7 +421,7 @@ impl DrawData for RectangleDrawData {
         view_info: &DrawableCollectionViewInfo,
         collector: &mut DrawableCollector<'_>,
     ) {
-        // TODO(#1025, #4787, #11156): Better handling of 2D objects, use per-2D layer sorting instead of depth offsets.
+        // TODO(#1025, #4787): Better handling of 2D objects, use per-2D layer sorting instead of depth offsets.
         // For 2D draw order based sorting, we should never enable depth write and always perform back to front sorting.
 
         for (index, instance) in self.instances.iter().enumerate() {
@@ -535,7 +535,7 @@ impl RectangleDrawData {
                     },
                 ),
                 draw_outline_mask: rectangle.options.outline_mask.is_some(),
-                has_transparency: rectangle.options.multiplicative_tint.a() < 1.0, // TODO(andreas): what about textures with alpha?
+                has_transparency: rectangle.options.multiplicative_tint.a() < 1.0, // TODO(#12223): what about textures with alpha?
             });
         }
 
--- a/crates/viewer/re_renderer/src/mesh.rs
+++ b/crates/viewer/re_renderer/src/mesh.rs
@@ -385,7 +385,7 @@ impl GpuMesh {
                     },
                 );
 
-                // TODO(andreas): handle texture transparency
+                // TODO(#12223): handle texture transparency
                 let is_transparent = material.albedo_factor.a() < 1.0;
 
                 materials.push(GpuMaterial {


oxkitsune:
A bit of clarification why this fixes the issue:

Since #11962, visualizer results are stored in a `HashMap` with a random iteration order.
This does not maintain insertion order, and our transparency blending requires a back-to-front order.

emilk:
> This does not maintain insertion order

Neither does BTreeMap!

oxkitsune (approved):
I'm not sure about the performance impact of this change, since `BTreeMap` lookups can be slower.

oxkitsune:
> > This does not maintain insertion order
> 
> Neither does BTreeMap!

Okay then I also have no idea what's going on 🫠

crates/viewer/re_view_spatial/tests/opacity_2d.rs
```diff
@@ -0,0 +1,84 @@
...
+                &re_sdk_types::archetypes::Image::from_color_model_and_tensor(
+                    re_sdk_types::datatypes::ColorModel::RGB,
+                    image,
+                )
+                .unwrap()
+                .with_draw_order(1.0),
+            )
+        });
+
+        test_context.log_entity("2d_layering/depth", |builder| {
```
emilk, line 38: Does this only work now because `2d_layering/depth < 2d_layering/gradient` lexicographically? 😬

oxkitsune: I hope not 🫠

emilk: Tried it and no - the name seems to have no effect

oxkitsune: Ohh, wait 🤦 the renderer still has its own sorting using the draw order. This fixes it because the ordering is consistent instead of being random?

emilk:
If it is truly the insertion order we should maintain, then `IndexMap` is probably the best solution. I'll hold off merging until @Wumpf can have a look tomorrow

Wumpf:
updated with different solution, see updated description

crates/viewer/re_renderer/src/renderer/rectangles.rs
```diff
@@ -529,6 +535,7 @@ impl RectangleDrawData {
                     },
                 ),
                 draw_outline_mask: rectangle.options.outline_mask.is_some(),
+                has_transparency: rectangle.options.multiplicative_tint.a() < 1.0, // TODO(andreas): what about textures with alpha?
```
emilk: Let's create a proper issue for this to track it ("Support alpha in images")

Wumpf:
Confirmed https://github.com/rerun-io/rerun/issues/12206 is also fixed with this latest incarnation

PR #12215, opened 2025-12-15
Adds some of the blueprint types required for component mapping

aedm:
### What

Adds a few blueprint types that will eventually be required by visualizer component mapping.

### DIFF
--- a/crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_narrow/VisualizerInstructionId_placeholder.png
+++ b/crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_narrow/VisualizerInstructionId_placeholder.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:09162333bf266dc369751e6bfa6f7376ad37bcc07ade5f4d8972d4a5988da0e7
-size 2687
+oid sha256:d7f26bf11900a737758ac7dc9d532d48d69d755d7a730da985f05370930ce628
+size 3580
--- a/crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_wide/VisualizerInstructionId_placeholder.png
+++ b/crates/viewer/re_component_ui/tests/snapshots/all_components_list_item_wide/VisualizerInstructionId_placeholder.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:e66d975032f08980bfdd4159c911a7686f2fdf9e58075206a0c93937814b2e2d
-size 3002
+oid sha256:32e651f2fe772da8c0a8ffc93af6d68f6f13d8a1d433fe2e4dab7ee2691010cf
+size 4935


.cargo/config.toml
```diff
@@ -7,6 +7,12 @@ codegen = "run --package re_types_builder --"
 # Temporary solution while we wait for our own xtasks!
 run-wasm = "run --release --package run_wasm --"
 
+
+[build]
+# Incremental compilation blows up the size of the target folder, and is also buggy.
+incremental = false
```
Wumpf, line 13: looks like we should first merge latest main into `experimental/plot-any-scalars`
or rebase it 🤷

crates/store/re_sdk_types/definitions/rerun/blueprint/components/visualizer_component_mapping.fbs
```diff
@@ -0,0 +1,10 @@
+namespace rerun.blueprint.components;
+
+/// Associate components of an entity to components of a visualizer.
```
Wumpf:
```suggestion
/// Associates components of an entity to components of a visualizer.
```

crates/store/re_sdk_types/definitions/rerun/blueprint/components/visualizer_instruction_id.fbs
```diff
@@ -0,0 +1,13 @@
...
+/// Single visualizer override the visualizers for an entity.
+table VisualizerInstructionId (
+  "attr.python.aliases": "str | list[str]",
+  "attr.python.array_aliases": "str",
+  "attr.rerun.scope": "blueprint",
+  "attr.rust.derive": "PartialEq, Eq, PartialOrd, Ord, Default",
+  "attr.rust.repr": "transparent"
+) {
+  /// Names of a visualizer that should be active.
+  visualizer: rerun.datatypes.Utf8 (order: 100);
```
Wumpf: that's the one we want to change to `rerun.datatype.Uuid`. I guess we could do so right away here? I'm also fine with leaving that for an iteration :)

crates/store/re_types/source_hash.txt
Wumpf, line 1: this file shouldn't be submitted

crates/store/re_sdk_types/definitions/rerun/blueprint/archetypes/visualizer_instruction.fbs
```diff
@@ -0,0 +1,13 @@
+namespace rerun.blueprint.archetypes;
+
+/// A visualizer instruction for an entity.
+table VisualizerInstruction (
```
Wumpf, line 4: wondering whether we should mark the new archetypes as experimental. Counter argument is that once we start using them they immediately become ubiquitous and mandatory in all blueprint files

Wumpf (changes requested):
lgtm minus submitted extra file

I think we can just land this one directly on main faster release since it doesn't touch any of the existing stuff

.cargo/config.toml
aedm: Yeah. :( We already have a ton of merge conflicts anyway, so I better merge main into the feature branch before I fix those.

crates/store/re_sdk_types/definitions/rerun/blueprint/components/visualizer_instruction_id.fbs
aedm: Good idea. I changed it to here.

crates/store/re_sdk_types/definitions/rerun/blueprint/archetypes/visualizer_instruction.fbs
aedm: To me, experimental means users can choose not to use it. So this is probably not experimental.

aedm:
> I think we can just land this one directly on main faster release since it doesn't touch any of the existing stuff

I'd rather do it after release.

Wumpf (approved):
👍 thanks

PR #12217, opened 2025-12-15
GC: target chunks furthest from time cursor

teh-cmc:
A very, very simple implementation of the `furthest_from` logic that just sorts and collects everything all at once + the necessary infrastructure to bench and test future performance improvements.

In practice, this is probably good enough for most use cases since the number of chunks in a single recording is never that large to begin with.

For 10k chunks with 1k rows each:
```
gc_everything/ordered_by_min_row_id
                        time:   [9.6599 ms 9.7254 ms 9.8152 ms]

gc_everything/ordered_by_distance_from_cursor
                        time:   [10.857 ms 10.910 ms 10.967 ms]
```

* Fixes https://linear.app/rerun/issue/RR-3002/chunkstore-impl-furthest-away-gc-strategy

TODO:
* [x] Try playing with this on the larger-than-ram branch

### DIFF
--- a/crates/store/re_chunk_store/src/query.rs
+++ b/crates/store/re_chunk_store/src/query.rs
@@ -89,7 +89,6 @@ impl ChunkStore {
     /// An implementation of `find_temporal_chunk_furthest_from` that focuses solely on correctness.
     ///
     /// Used to compare with results obtained from the optimized implementation.
-    #[cfg(test)]
     pub(crate) fn find_temporal_chunks_furthest_from_slow(
         &self,
         timeline: &TimelineName,


crates/viewer/re_viewer/src/app.rs
```diff
@@ -2725,7 +2725,13 @@ impl App {
...
                 );
             }
-            store_hub.purge_fraction_of_ram(fraction_to_purge);
+
+            let time_cursor_for =
+                |store_id: &StoreId| -> Option<(re_log_types::Timeline, re_log_types::TimeInt)> {
+                    let time_ctrl = self.state.time_controls.get(store_id)?;
+                    Some((*time_ctrl.timeline()?, time_ctrl.time_int()?))
+                };
+            store_hub.purge_fraction_of_ram(fraction_to_purge, &time_cursor_for);
```
teh-cmc, line 2734: Should we leave this as is, i.e. always enabled?

crates/store/re_chunk_store/src/gc.rs
```diff
@@ -270,57 +275,91 @@ impl ChunkStore {
...
             {
-                if let Some(chunk) = self.chunks_per_chunk_id.get(chunk_id) {
-                    if options.is_chunk_protected(chunk) {
-                        continue;
-                    }
+                let chunks = self.find_temporal_chunks_furthest_from(timeline, *time);
+
+                // This will only apply for tests run from this crate's src/ directory, which is good
+                // enough for our purposes.
+                #[cfg(test)]
```
emilk: nit/style: this reduces the chance that we commit a breaking change to the code that we don't detect until the tests are run on CI

```suggestion
                if cfg!(test)
```

crates/store/re_chunk_store/src/gc.rs
```diff
@@ -270,57 +275,91 @@ impl ChunkStore {
...
+                    });
+
+            let chunks_in_priority_order = chunks_furthest_away
+                .into_iter()
+                .chain(chunks_in_min_row_id_order);
+
+            for chunk in
+                chunks_in_priority_order.filter(|chunk| !protected_chunk_ids.contains(&chunk.id()))
+            {
+                if options.is_chunk_protected(&chunk) {
```
emilk: This seem to be applying the protection twice?

crates/store/re_chunk_store/src/query.rs
```diff
@@ -41,6 +43,77 @@ impl ChunkStore {
...
+                min_dist.map(|max_dist| (chunk, max_dist))
+            })
+            .sorted_by(|(_chunk1, dist1), (_chunk2, dist2)| std::cmp::Ord::cmp(dist2, dist1)) // descending
+            .map(|(chunk, _dist)| chunk.clone())
+            .collect_vec()
+    }
+
+    /// An implementation of `find_temporal_chunk_furthest_from` that focuses solely on correctness.
+    ///
+    /// Used to compare with results obtained from the optimized implementation.
```
emilk, line 91: nice

crates/viewer/re_viewer/src/app.rs
emilk: I think that's reasonable

crates/store/re_chunk_store/src/gc.rs
```diff
@@ -270,57 +275,91 @@ impl ChunkStore {
...
-            for chunk_id in self
-                .chunk_ids_per_min_row_id
-                .values()
-                .filter(|chunk_id| !protected_chunk_ids.contains(chunk_id))
+            // These chunks cannot be dangling by definition, since we need to access their data in
+            // order to sort them in the first place.
+            //
+            // TODO(cmc): we would very much like that to be iterative or at least paginated in
+            // some way, so that it doesn't eat away all of the mark phase's time budget for no
+            // reason, but that requires making things much more complicated, so let's see how fair
```
teh-cmc:
```suggestion
            // reason, but that requires making things much more complicated, so let's see how far
```

crates/store/re_chunk_store/src/gc.rs
teh-cmc: It's not, one is the the Chunk ID protection and the other is the time range protection iirc. I'll do some naming cleanup before merging.

crates/store/re_chunk_store/src/query.rs
```diff
@@ -41,6 +43,77 @@ impl ChunkStore {
...
+                        .get(pivot.saturating_sub(1))
+                        .map(|t| t.abs_diff(time.as_i64()));
+                    let min_value2 = times
+                        .times_raw()
+                        .get(pivot)
+                        .map(|t| t.abs_diff(time.as_i64()));
+
+                    // NOTE: Do *not* compare these options directly, if any of them turns out to
+                    // be None, it'll be a disaster.
+                    // min_value1.min(min_value2).min(min_value3);
```
teh-cmc:
```suggestion
                    // min_value1.min(min_value2);
```

crates/viewer/re_viewer/src/app.rs
teh-cmc: That might look a bit silly if e.g. you stop your time cursor at the leftmost spot in the timeline while ingesting data in real-time... so then at some point you'll be GC'ing the data that's coming in :melting_face: But eh, niche problems.

PR #12220, opened 2025-12-15
[python] Automatically disconnected recordings when they go fully out of scope

jleibs:
### Related
- RR-3065: SDK should flush both IO and app-level logic when a recording gets GC'd

### What

Previously, even if a recording had been fully GC'd from the python side of things, we didn't fully disconnect it until program shutdown, which made it easy to observe RRD's that weren't totally closed, even if they contained all the data.

This has become significantly more important with the introduction of RRD footers.

This PR adds a new binding helper `disconnect_orphaned_recordings()` that disconnects any recordings that we know are no longer reachable. We call this after cleaning up all of our references.

### DIFF
--- a/rerun_py/rerun_bindings/rerun_bindings.pyi
+++ b/rerun_py/rerun_bindings/rerun_bindings.pyi
@@ -857,7 +857,11 @@ def set_thread_local_blueprint_recording(
     """
 
 def check_for_rrd_footer(file_path: str | os.PathLike[str]) -> bool:
-    """Check if the given RRD file has a footer."""
+    """
+    Check if the RRD has a valid RRD footer.
+
+    This is useful for unit-tests to verify that data has been fully flushed to disk.
+    """
 
 def disconnect_orphaned_recordings() -> None:
     """


rerun_py/tests/unit/test_disconnect_on_cleanup.py
```diff
@@ -0,0 +1,39 @@
...
+        rec_path = f"{dirpath}/rec.rrd"
+
+        def create_recording() -> None:
+            with rr.RecordingStream("rerun_example_ctx") as rec:
+                rec.save(rec_path)
+                rec.log("x", rr.Points2D(positions=[(1, 2), (3, 4)]))
+
+        create_recording()
+
+        assert rerun_bindings.check_for_rrd_footer(rec_path)
```
teh-cmc, line 35: Wouldn't it be more interesting to run that `assert` just outside of the with block instead?
```suggestion
        def create_recording() -> None:
            with rr.RecordingStream("rerun_example_ctx") as rec:
                rec.save(rec_path)
                rec.log("x", rr.Points2D(positions=[(1, 2), (3, 4)]))
            assert rerun_bindings.check_for_rrd_footer(rec_path)

        create_recording()
```

rerun_py/src/python_bridge.rs
```diff
@@ -290,6 +292,32 @@ fn flush_and_cleanup_orphaned_recordings(py: Python<'_>) -> PyResult<()> {
...
+        // only referenced by the `all_recordings` list and thus can't be referred to by the Python SDK.
+        for recording in all_recordings().iter() {
+            if recording.ref_count() <= 1 {
+                re_log::error!("Disconnecting orphaned recording");
+                recording.disconnect();
+            } else {
+                re_log::error!(
+                    "Not disconnecting recording, refcount = {}",
+                    recording.ref_count()
+                );
```
teh-cmc: Why would that be an error? That's just some other recording living its best life, what's wrong with that?

teh-cmc (approved):
LGTM but needs someone with a better eye for Python's subtleties to review cc @abey79 @ntjohnson1

rerun_py/src/python_bridge.rs
```diff
@@ -2354,3 +2386,16 @@ fn get_credentials(py: Python<'_>) -> PyResult<Option<PyCredentials>> {
     .map_err(|err| PyRuntimeError::new_err(err.to_string()))?;
     Ok(Some(PyCredentials(credentials)))
 }
+
+#[pyfunction]
+/// Check if the RRD has a valid RRD footer.
+///
+/// This is useful for unit-tests to verify that data has been fully flushed to disk.
+fn check_for_rrd_footer(file_path: std::path::PathBuf) -> PyResult<bool> {
```
ntjohnson1, line 2395: If this is just meant for testing just prefix with underscore so its not easily discoverable?

rerun_py/src/python_bridge.rs
jleibs: Not an error. That was me feeling crazy for missing a debug statement and wanting to make sure I caught it. This one will go away and the other one is a debug. Thanks for catching.

rerun_py/src/python_bridge.rs
jleibs: It's only in the bridge, which is not part of the public API surface anyways.

PR #12222, opened 2025-12-16
`/GetRrdManifest` now returns actual `RrdManifest`s

teh-cmc:
Title.

In addition:
* `chunk_byte_size` & `chunk_byte_offset` are always mandatory
  * they are filled with uncompressed values in contexts where compressed values cannot be computed (e.g. the OSS server, where there is no disk involved and thus no compression)
    > ## Understand size/offset columns
    >
    > * `chunk_byte_size` & `chunk_byte_offset` are always reported using the backend's native
    >   storage size. For a backend that makes use of compression, such as an RRD file with
    >   compression enabled, these sizes are therefore compressed. For a backend that doesn't do any
    >   kind of compression, such as the OSS server that stores everything already decoded in memory,
    >   these sizes will correspond to heap memory usage.
    > * `chunk_byte_size_uncompressed` always corresponds to the size on the heap that the data would
    >   require once fully decoded, regardless of the backend.
    > * `chunk_key`, if specified, should always be used to fetch the associated data.
  * `chunk_key` is optional

gRPC breaking changes on purpose: unpublished API.

---

* Fixes https://linear.app/rerun/issue/RR-3182/getrrdmanifest-needs-to-return-an-actual-rrdmanifest

### DIFF
--- a/crates/store/re_redap_tests/src/tests/snapshots/re_redap_tests__tests__rrd_manifest__rrd_manifest_sorbet_schema.snap
+++ b/crates/store/re_redap_tests/src/tests/snapshots/re_redap_tests__tests__rrd_manifest__rrd_manifest_sorbet_schema.snap
@@ -5,14 +5,14 @@ expression: rrd_manifest.sorbet_schema.format_snapshot()
 top-level metadata: [
     sorbet:version: "0.1.2"
 ]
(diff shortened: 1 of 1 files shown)


emilk:
the barchart CI failure is on main too (someone is on it)
* Fix coming in https://github.com/rerun-io/rerun/pull/12228

PR #12224, opened 2025-12-16
Parse `tf2_msgs/TFMessage` from ROS 2 MCAPs as `Transform3D`s

MichaelGrupp:
Adds a parser for [TFMessage](https://docs.ros2.org/foxy/api/tf2_msgs/msg/TFMessage.html).
This message type is a bit special because it contains a variable size list of individual transforms, where each one has its own header timestamp.

The parser creates `Transform3D`s from that, and assigns each output transform a `ros2_timestamp` set to the individual transform's header stamp and a `log/publish_time` timepoint according to the input message.

To achieve this, the default log/pub timepoint handling had to be slightly adapted to allow a custom override for this parser that can produce that variable amount of log/pub timepoints.

Closes: https://linear.app/rerun/issue/RR-3181

### DIFF
--- a/crates/store/re_mcap/src/parsers/ros2msg/tf2_msgs/tf_message.rs
+++ b/crates/store/re_mcap/src/parsers/ros2msg/tf2_msgs/tf_message.rs
@@ -6,6 +6,8 @@ use re_sdk_types::components::{RotationQuat, Translation3D};
 use re_sdk_types::datatypes::Quaternion;
 
 use super::super::Ros2MessageParser;
+use crate::parsers::ros2msg::definitions::geometry_msgs::{Transform, TransformStamped};
+use crate::parsers::ros2msg::definitions::std_msgs::Header;
 use crate::parsers::{
     cdr,
     decode::{MessageParser, ParserContext},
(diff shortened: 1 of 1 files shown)


crates/store/re_mcap/src/parsers/ros2msg/tf2_msgs/tf_message.rs
```diff
@@ -0,0 +1,110 @@
...
+            transforms.len()
+        ])
+    }
+
+    fn append(&mut self, ctx: &mut ParserContext, msg: &mcap::Message<'_>) -> anyhow::Result<()> {
+        re_tracing::profile_function!();
+        let TFMessage { transforms } = cdr::try_decode_message::<TFMessage>(&msg.data)?;
+
+        // Each transform in the message has its own timestamp.
+        for transform in transforms {
```
emilk: can use destructor pattern here for `transform`

MichaelGrupp: ah, again. I need a clippy lint for this 😄

PR #12225, opened 2025-12-16
Add support for creating dataset from list of RRDs to `rerun.server.Server`

abey79:
### Related

### What

So far, the OSS server only add the ability to create dataset on start based on a prefix path, aka a path to a directory that would then be globed for `*.rrd`. This is true for both the CLI (`rerun server`) and the Python wrapper (`rerun.server.Server`).  Sometime, we want to instead create a dataset from a well defined list of path to RRDs instead.

This PR adds this ability to `Server`. It also does 90% of the work for the same in the CLI, but leaves the actual CLI UX as a todo since it's not immediately obvious and trivial how to do that.

This PR also:
- fixes a bug where dataset created at start did not have a matching blueprint dataset
- rename existing `datasets` to `dataset_prefixes` when directories of RRDs are involved
- simplifies a bit the code and signature of rust-based `_ServerInternal` (signature shenanigans are best left to the python wrapper)
- add a test suite for the `Server` object

### DIFF
--- a/crates/store/re_grpc_server/src/lib.rs
+++ b/crates/store/re_grpc_server/src/lib.rs
@@ -772,7 +772,7 @@ impl EventLoop {
             return;
         };
 
-        self.messages.gc(max_bytes as _);
+        self.messages.gc(max_bytes);
     }
 }
 
--- a/crates/utils/re_memory/src/memory_limit.rs
+++ b/crates/utils/re_memory/src/memory_limit.rs
@@ -14,7 +14,7 @@ pub struct MemoryLimit {
     /// This is primarily compared to what is reported by [`crate::AccountingAllocator`] ('counted').
     /// We limit based on this instead of `resident` (RSS) because `counted` is what we have immediate
     /// control over, while RSS depends on what our allocator (MiMalloc) decides to do.
-    pub max_bytes: Option<i64>,
+    pub max_bytes: Option<u64>,
 }
 
 impl MemoryLimit {
@@ -61,7 +61,7 @@ impl MemoryLimit {
         } else {
             re_format::parse_bytes(limit)
                 .map(|max_bytes| Self {
-                    max_bytes: Some(max_bytes),
+                    max_bytes: Some(max_bytes.max(0) as _),
                 })
                 .ok_or_else(|| format!("expected e.g. '16GB', got {limit:?}"))
         }
--- a/crates/utils/re_memory/src/memory_use.rs
+++ b/crates/utils/re_memory/src/memory_use.rs
@@ -1,5 +1,3 @@
-#![expect(clippy::cast_possible_wrap)] // u64 -> i64
-
 /// How much RAM is the application using?
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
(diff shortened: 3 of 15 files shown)


crates/store/re_server/src/store/in_memory_store.rs
```diff
@@ -306,15 +306,45 @@ impl InMemoryStore {
...
     }
 
+    /// Create a (regular) dataset with a matching blueprint dataset.
     pub fn create_dataset(
         &mut self,
-        name: &str,
+        dataset_name: String,
+        dataset_id: Option<EntryId>,
+    ) -> Result<&mut Dataset, Error> {
+        let dataset_id = dataset_id.unwrap_or_else(EntryId::new);
```
emilk, line 318: Why make it an option if the user can just call `EntryId::new` themselves?

crates/store/re_server/src/entrypoint.rs
```diff
@@ -42,6 +47,13 @@ pub struct NamedPath {
     pub path: PathBuf,
 }
 
+/// A named collection of paths.
+#[derive(Debug, Clone)]
+pub struct NamedCollection {
+    pub name: String,
+    pub paths: Vec<PathBuf>,
```
emilk, line 57: Docstring! Can this be a list of folders? Or just rrds? Can it be a mix?

rerun_py/rerun_bindings/rerun_bindings.pyi
```diff
@@ -1532,33 +1532,29 @@ class AlreadyExistsError(Exception):
     """Raised when trying to create a resource that already exists."""
 
 class _ServerInternal:
-    """
-    Internal Rerun server instance.
-
-    This is the low-level binding to the Rust server implementation.
-    Users should typically use `rerun.server.Server` instead.
-    """
```
emilk, line 1540: Isn't this a useful docstring to have?

crates/store/re_server/src/store/in_memory_store.rs
abey79: This argument shouldn't even exist. Assigning entry ids is a server implementation details. However, we surface the possibility of override all the way to redap for testing purposes. So I'd much rather keep that option to convey that this is "use at your own risk" (I'll add a docstring to that effect).

abey79: That said, the `_impl` function just below is internal, and in that case the option doesn't make sense. I'll fix that.

crates/store/re_server/src/entrypoint.rs
abey79: moving things around for clarity. i'll clarify the docstring of `datasets: Vec<NamedCollection>`. This is where the nature of the paths matters.

rerun_py/rerun_bindings/rerun_bindings.pyi
abey79: This is not used in the documentation, is not user facing, and subject to rot. Authoritative docstrings for internal objects and their methods are on the rust side.

PR #12226, opened 2025-12-16
Fix deprecated decorator ordering

oxkitsune:
### What

Puts the `@deprecated` decorator on the function instead of the `@property` decorator.
This will fix the CI failing to build/test examples


oxkitsune:
@rerun-bot full-check

abey79 (approved):
thanks

PR #12231, opened 2025-12-16
Component id reflection for any scalars

aedm:
### What

Codegen for `ComponentDescriptor` and `ComponentIdentifier`.


Wumpf (changes requested):
We found a better way to do this without emitting more reflection during the meetup. We should roll with that instead.
@grtlr I remember you saying it worked out fine, but I don't know where that went

grtlr:
It went onto [`main`](https://github.com/rerun-io/rerun/commit/9f0d1e76737a8dbd9add83bec0a4e1ee581ce08b):

* https://github.com/rerun-io/rerun/pull/12202

PR #12235, opened 2025-12-16
Do a lot more book-keeping in `RrdManifestIndex`

emilk:
## Related
* Part of RR-3105

## What
If we receive an `RrdManifestIndex`, we will now use it to pre-populate the `EntityTree` and `TimeHistogram`s. This lets us know where there will be data before it is loaded. It also adds a bunch of helper functions that will be used for prefetching chunks.

### DIFF
--- a/crates/store/re_entity_db/src/sorted_range_map.rs
+++ b/crates/store/re_entity_db/src/sorted_range_map.rs
@@ -55,22 +55,15 @@ impl<K: Ord + Copy, V> SortedRangeMap<K, V> {
     }
 
     #[inline]
-    #[allow(clippy::allow_attributes, dead_code)] // used for tests
+    #[cfg_attr(not(test), expect(dead_code))] // only used in tests
     pub fn len(&self) -> usize {
         self.entries.len()
     }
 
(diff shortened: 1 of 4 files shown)


crates/store/re_entity_db/src/sorted_range_map.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    /// Find the first index that could possibly overlap with the query.
+    #[inline]
+    fn find_first_possible(&self, query: &RangeInclusive<K>) -> usize {
+        // We need max_end[i] >= query.start for any overlap to be possible
+        self.max_end.partition_point(|max| *max < *query.start())
+    }
+
+    #[inline]
+    #[allow(clippy::allow_attributes, dead_code)] // used for tests
```
IsseW: Could maybe do `#[cfg_attr(not(test), expect(dead_code))]` (haven't tested) here, and other instances in this file, to be warned when we start to use them

emilk: good idea!

PR #12237, opened 2025-12-16
Updated release workflows to have access to container registry

ntjohnson1:
Changes required for the release workflow to run. Now that we use GHCR we need to specify package access when using container.


ntjohnson1:
Cherry-picked these back from my alpha branch https://github.com/rerun-io/rerun/actions/runs/20269086705

PR #12242, opened 2025-12-16
Compare sorted results in e2e tests

swallez:
`re_server` and rerun cloud return the same test result, but not necessarily in the same order, which functionally is not a problem but breaks snapshot tests. This PR sorts schema and dataframes to allow for reliable snapshot testing.

### DIFF
--- a/rerun_py/tests/e2e_redap_tests/test_dataset_views.py
+++ b/rerun_py/tests/e2e_redap_tests/test_dataset_views.py
@@ -4,13 +4,13 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-import datafusion
 import pyarrow as pa
 import pytest
 
 if TYPE_CHECKING:
     from pathlib import Path
 
+    import datafusion
     from rerun.catalog import DatasetEntry
     from syrupy import SnapshotAssertion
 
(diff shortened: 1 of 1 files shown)


rerun_py/tests/e2e_redap_tests/test_dataset_views.py
```diff
@@ -106,4 +107,10 @@ def test_dataset_view_reader(readonly_test_dataset: DatasetEntry, snapshot: Snap
     view = readonly_test_dataset.filter_segments([first_segment]).filter_contents(["/obj1/**"])
     df = view.reader(index="time_1")
 
+    df = sorted_df(df)
+
     assert str(df) == snapshot()
+
+def sorted_df(df: datafusion.DataFrame) -> datafusion.DataFrame:
+    sorted_fields = sorted([field.name for field in df.schema()])
+    return df.select(*sorted_fields)
```
emilk: for your own sanity, I suggest configuring your editor to always end files with a newline (as well as trim trailing whitespace).

https://github.com/rerun-io/rerun/blob/057e18c3faac300d8d265fa0b55ece51778fa697/CONTRIBUTING.md#L274

swallez: Indeed, and done. I started with a fresh install of RustRover and for some reason this isn't the default 🤷

PR #12243, opened 2025-12-16
bugfix: validate table names on create_table

timsaucer:
Closes https://linear.app/rerun/issue/RR-3203/create-table-fails-silently-if-dash-in-table-name

This PR adds a check that we have a SQL parseable name when provided to `create_table`. If not, return an error without creating the table.

### DIFF
--- a/rerun_py/tests/e2e_redap_tests/test_table_create.py
+++ b/rerun_py/tests/e2e_redap_tests/test_table_create.py
@@ -65,3 +65,15 @@ def test_create_table_in_custom_schema(catalog_client: CatalogClient, tmp_path: 
         assert returned_schema == original_schema
     finally:
         table_entry.delete()
+
+
+@pytest.mark.creates_table
+def test_create_table_invalid_name(entry_factory: EntryFactory, tmp_path: pathlib.Path) -> None:
+    table_name = "created-table"
+
+    schema = pa.schema([("int64", pa.int64()), ("float32", pa.float32()), ("utf8", pa.utf8())])
+    with pytest.raises(
+        ValueError,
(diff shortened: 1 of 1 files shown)


rerun_py/src/catalog/catalog_client.rs
```diff
@@ -242,6 +242,13 @@ impl PyCatalogClientInternal {
     ) -> PyResult<Py<PyTableEntryInternal>> {
         let connection = self_.borrow_mut(py).connection.clone();
 
+        // Verify we have a valid table name
+        let dialect = datafusion::logical_expr::sqlparser::dialect::GenericDialect;
+        let _ = datafusion::logical_expr::sqlparser::parser::Parser::new(&dialect)
+            .try_with_sql(name.as_str())
+            .and_then(|mut parser| parser.parse_multipart_identifier())
+            .map_err(|err| PyValueError::new_err(format!("Invalid table name. {err}")))?;
```
emilk, line 250: Please make sure the final error message shown to the user contains the (invalid) name, ideally quoted (`{name:?}`). Maybe that's already part of `{err}` here; I don't know

emilk (approved):

rerun_py/src/catalog/catalog_client.rs
emilk: Could maybe have a unit test to verify that, if not too much work

timsaucer: The error message does show what the problem is specifically. I can add the unit test.

andrea-reale (approved):
Thanks! It'd be great if we could have a pytest in the e2e tests package

PR #12245, opened 2025-12-17
Skip heuristic exclusions if transform frame ids have been logged

IsseW:
### What

Title.

Because these more or less define a seperate entity tree that is opaque when we generate heuristics. So can't make any assumptions about the entity tree.


Wumpf (approved):
thanks for adding a test right away!
I think that will do until we figure out something better for `SpatialTopology` :)

PR #12246, opened 2025-12-17
Fix missing error when drawing depth image for point cloud without pinhole

Wumpf:
We checked this correctly for images but not for depth images (since they're a bit more complex with their depth cloud and all)

Extended tests accordingly!

image

### DIFF
--- a/crates/viewer/re_view_spatial/src/visualizers/depth_images.rs
+++ b/crates/viewer/re_view_spatial/src/visualizers/depth_images.rs
@@ -15,7 +15,6 @@ use re_viewer_context::{
 use super::entity_iterator::process_archetype;
 use super::{SpatialViewVisualizerData, textured_rect_from_image};
 use crate::contexts::{SpatialSceneEntityContext, TransformTreeContext};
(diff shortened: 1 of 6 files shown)


crates/viewer/re_view_spatial/src/contexts/mod.rs
```diff
@@ -17,6 +17,8 @@ pub struct SpatialSceneEntityContext<'a> {
 
     pub highlight: &'a re_viewer_context::ViewOutlineMasks, // Not part of the context, but convenient to have here.
     pub view_class_identifier: ViewClassIdentifier,
+
+    pub output: &'a mut re_viewer_context::VisualizerExecutionOutput,
```
Wumpf, line 21: this is a pattern I also started on the `EncodedDepthImage` PR. Turned out to be very useful. There's surely more that can be done to make this cleaner still, but importantly it allows error reporting in more places easily and in a very non-intrusive manner

Wumpf:
merge broke things. fixing.

PR #12248, opened 2025-12-17
When linking to a time range, load those chunks _first_

emilk:
* Closes RR-3207

When opening a link to a specific time range, we now download the chunks needed for just that time range _first_, then continue downloading all the other chunks.

Tested locally: works as intended.

### DIFF
--- a/crates/store/re_redap_client/src/grpc.rs
+++ b/crates/store/re_redap_client/src/grpc.rs
@@ -469,7 +469,7 @@ async fn stream_segment_from_server(
     let mut already_loaded_chunk_ids: ahash::HashSet<ChunkId> = Default::default();
 
     if let Some(time_selection) = fragment.time_selection {
-        // Start be loading only the chunks required for the time selection:
+        // Start by loading only the chunks required for the time selection:
         let time_selection_batches = client
             .query_dataset_chunk_index(SegmentQueryParams {
                 dataset_id,


crates/store/re_redap_client/src/grpc.rs
```diff
@@ -324,15 +325,19 @@ pub async fn stream_blueprint_and_segment_from_server(
...
             &tx,
             blueprint_dataset,
             blueprint_segment,
             re_uri::Fragment::default(),
         )
-        .await?;
+        .await?
+        .is_break()
+        {
+            return Ok(());
```
emilk, line 339: Small correctness fix: if the receiver is dropped, we would previously continue trying to load the recording etc. Now we stop earlier.

crates/store/re_redap_client/src/grpc.rs
```diff
@@ -429,46 +438,103 @@ async fn stream_segment_from_server(
...
-                .send(DataSourceMessage::RrdManifest(
-                    store_id.clone(),
-                    rrd_manifest.into(),
-                ))
-                .is_err()
-            {
-                re_log::debug!("Receiver disconnected");
-                return Ok(()); // cancelled
+    // TODO(RR-3110): load RRD manifest first
+    if false {
```
emilk, line 442: Save some time by not loading RRD manifest. We make no use of it yet anyway

crates/store/re_redap_client/src/grpc.rs
```diff
@@ -429,46 +438,103 @@ async fn stream_segment_from_server(
...
         }
-        Err(err) => {
-            if err.kind == ApiErrorKind::Unimplemented {
-                // TODO(RR-3110): implement rrd manifest on cloud
+    }
+
+    let mut already_loaded_chunk_ids: ahash::HashSet<ChunkId> = Default::default();
+
+    if let Some(time_selection) = fragment.time_selection {
+        // Start be loading only the chunks required for the time selection:
```
emilk: This is new!

IsseW:
```suggestion
        // Start by loading only the chunks required for the time selection:
```

PR #12249, opened 2025-12-17
Make empty prefix error handling consistent with dpf

abey79:
This fixes the e2e test for the empty prefix directory error and the related OSS server implementation such that it is consistent with dpf. This should fix the test failure on dpf side.


abey79:
yolo merging this

PR #12186, opened 2025-12-11
Fix pending timeline

IsseW:
### Related

* Introduced in https://github.com/rerun-io/rerun/pull/12172

### What

Fixes the timeline from pending timeline being used directly.

PR #12250, opened 2025-12-17
Fix skipping updating overrides if no timeline is active

IsseW:
### Related

* Fixes bug introduced in https://github.com/rerun-io/rerun/pull/12186

### What

At the start `QueryRange` is assigned to default, with the expectation that it should be overriden before used. But if the current timeline is pending we currently skipped the update that would override it. Causing time series view to receive an unexpected `QueryRange::LatestAt`

### DIFF
--- a/crates/viewer/re_view_time_series/src/util.rs
+++ b/crates/viewer/re_view_time_series/src/util.rs
@@ -43,10 +43,12 @@ pub fn determine_time_range(
     let visible_time_range = match query_range {
         re_viewer_context::QueryRange::TimeRange(time_range) => time_range.clone(),
         re_viewer_context::QueryRange::LatestAt => {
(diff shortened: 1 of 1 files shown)


crates/viewer/re_view_time_series/src/util.rs
```diff
@@ -43,7 +43,7 @@ pub fn determine_time_range(
     let visible_time_range = match query_range {
         re_viewer_context::QueryRange::TimeRange(time_range) => time_range.clone(),
         re_viewer_context::QueryRange::LatestAt => {
-            re_log::error_once!(
+            re_log::error!(
```
emilk: Make this a debug-only log, so we don't risk showing this to actual users

PR #12253, opened 2025-12-17
make fetching chunks in the dfn table provider concurrent while maintaining order

zehiko:
### What

Sending 1 request per segment sequentially has a big impact on performance. The reason we did it this way is ordering, having that guarantee is beneficial in some common query patterns. 

This PR keeps that promise (segment ordering), while performing concurrent fetches in batches. We batch the requests and execute desired number of them concurrently and then we re-order responses before flushing them downstream.

This brings 5-10x (and in some cases more) better benchmark times (server side change is also required to achieve that), see numbers in [here](https://linear.app/rerun/issue/RR-3150/benchmarking-and-datset-size-numbers).

We will do a follow up to make target batch size and concurrency level configurable, for now went with safe and performant defaults.

### Testing done
* existing tests for correctness
* new unit tests for verify ordering and batching
* lots of benchmarking

### DIFF
--- a/crates/store/re_datafusion/src/dataframe_query_provider.rs
+++ b/crates/store/re_datafusion/src/dataframe_query_provider.rs
@@ -443,8 +443,9 @@ async fn chunk_store_cpu_worker_thread(
     Ok(())
 }
 
-/// Extract segment ID from a `chunk_info` `RecordBatch`. Currently standing assumption is that `chunk_info` batch contains
-/// chunks *for a single segment*, hence we can just take the first row's `segment_id`.
+/// Extract segment ID from a `chunk_info` `RecordBatch`. Each `chunk_info` batch contains
+/// chunks *for a single segment*, hence we can just take the first row's `segment_id`. This is
+/// guaranteed by the implementation in `group_chunk_infos_by_segment_id`.
 fn extract_segment_id(chunk_info: &RecordBatch) -> Result<String, DataFusionError> {
     let segment_ids = chunk_info
         .column_by_name(re_protos::cloud::v1alpha1::QueryDatasetResponse::FIELD_CHUNK_SEGMENT_ID)
@@ -490,7 +491,7 @@ fn create_request_batches(
     for chunk_info in chunk_infos {
         let segment_id = extract_segment_id(&chunk_info)?;
         let chunk_sizes = extract_chunk_sizes(&chunk_info)?;
-        let segment_size: u64 = (0..chunk_sizes.len()).map(|i| chunk_sizes.value(i)).sum();
+        let segment_size: u64 = chunk_sizes.iter().map(|v| v.unwrap_or(0)).sum();
 
         // Track original segment order
         if !segment_order.contains(&segment_id) {


crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -435,51 +443,288 @@ async fn chunk_store_cpu_worker_thread(
     Ok(())
 }
 
+/// Extract segment ID from a `chunk_info` `RecordBatch`. Currently standing assumption is that `chunk_info` batch contains
```
timsaucer: This assumption is guaranteed by the implementation in `group_chunk_infos_by_segment_id`

crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -435,51 +443,288 @@ async fn chunk_store_cpu_worker_thread(
...
+) -> Result<BatchingResult, DataFusionError> {
+    let mut request_batches = Vec::new();
+    let mut current_batch = Vec::new();
+    let mut current_batch_size = 0u64;
+    let mut segment_order = Vec::new();
+
+    for chunk_info in chunk_infos {
+        let segment_id = extract_segment_id(&chunk_info)?;
+        let chunk_sizes = extract_chunk_sizes(&chunk_info)?;
+        let segment_size: u64 = (0..chunk_sizes.len()).map(|i| chunk_sizes.value(i)).sum();
```
timsaucer: There's no direct iterator? feels non-idiomatic

timsaucer (commented):
The logic looks sound. We have the existing e2e test `test_segment_ordering` to validate ordering, but I'm not convinced that it has been demonstrated at scale. Maybe we should make a single run against a larger dataset that saturates the stream to verify?

crates/store/re_datafusion/src/dataframe_query_provider.rs
zehiko: thanks, updating the comment.

crates/store/re_datafusion/src/dataframe_query_provider.rs
zehiko: yeah, that was weird, fixed.

zehiko:
> The logic looks sound. We have the existing e2e test `test_segment_ordering` to validate ordering, but I'm not convinced that it has been demonstrated at scale. Maybe we should make a single run against a larger dataset that saturates the stream to verify?

I'll go ahead and try to run a few similar tests on one of the larger datasets.

zehiko:
> > The logic looks sound. We have the existing e2e test `test_segment_ordering` to validate ordering, but I'm not convinced that it has been demonstrated at scale. Maybe we should make a single run against a larger dataset that saturates the stream to verify?
> 
> I'll go ahead and try to run a few similar tests on one of the larger datasets.

ran it locally against sample500 (as even that takes 10 minutes for all 3 time indexes we have, test passed). I'd say there is some reasonable confidence this is doing the right thing when it comes to ordering. Will start 2k in the background just in case.

PR #12256, opened 2025-12-17
Refactor `AppOptions` and add larger-than-RAM feature flag

emilk:
Screenshot 2025-12-17 at 14 35 24

Only visible in debug builds.

I couldn't resist grouping all the video settings into its own struct. This unfortunately means the video settings will reset to default in the next Rerun release. But having a flat structure of everything just isn't scalable.

In order to be able to read the "larger-than-ram" setting in time, I also had to re-arrange the startup code for the viewer slightly.

Let's hold off merging until we cut the 0.28 release branch.


crates/viewer/re_viewer/tests/snapshots/settings_screen.png
abey79, line 1: The new snapshot has different highlight on the top right "show left panel" icon. Not sure why or if it even matters.

emilk: Maybe it was below the threshold of error on main? 😬

PR #12258, opened 2025-12-17
Fix `compare_snippet_output.py` if `--no-py-build` isn't set

MichaelGrupp:
`pixi build py-build` doesn't exist in the `wheel-test-min` environment.

Useful when running the tests locally and you wanna build everything. E.g.:
```sh
RUST_LOG=debug pixi run -e wheel-test-min python docs/snippets/compare_snippet_output.py --target aarch64-apple-darwin
```


docs/snippets/compare_snippet_output.py
```diff
@@ -314,7 +314,7 @@ def build_python_sdk(build_env: dict[str, str]) -> None:
     print("----------------------------------------------------------")
     print("Building rerun-sdk for Python…")
     start_time = time.time()
-    run(["pixi", "run", "py-build"], env=build_env, timeout=12000)
+    run(["pixi", "run", "py-build-common"], env=build_env, timeout=12000)
```
MichaelGrupp, line 327: `py-build` is an alias for `py-build-common` in the `py` env:
https://github.com/rerun-io/rerun/blob/819335b24925d82e1ea8ded749c97bf473c2b66e/pixi.toml#L426

So I assume this here is better for the `min-wheel-test` env.

PR #12260, opened 2025-12-17
Fix race condition in urdf snippet rrd comparison

ntjohnson1:
I thought `rerun rrd compare --unordered` handled chunks out of order but apparently not in this case. Since logging is async we can get a race where sometimes the urdf logs before/after the transform. Functionally this shouldn't be an issue but our rrd comparison test appears to be brittle.

Force a flush so the ordering is always fixed.

* [x] full check green


ntjohnson1:
@rerun-bot full-check

PR #12262, opened 2025-12-17
Show `TransformAxes3D` for out-of-hierarchy `Transform`s

grtlr:
### Related

* Closes RR-3055.

### What

| Axis length | Show frame |
|--------|--------|
| image | image | 

This allows visualizing of transformations that are logged with `parent` and `child` frames, even if no associated geometry is logged.

### DIFF
--- a/crates/utils/re_byte_size/src/std_sizes.rs
+++ b/crates/utils/re_byte_size/src/std_sizes.rs
@@ -73,9 +73,10 @@ impl<K: SizeBytes, S> SizeBytes for HashSet<K, S> {
         // NOTE: It's all on the heap at this point.
 
         if K::is_pod() {
-            (self.len() * std::mem::size_of::<K>()) as _
+            (self.capacity() * std::mem::size_of::<K>()) as _
         } else {
-            self.iter().map(SizeBytes::total_size_bytes).sum::<u64>()
+            (self.capacity() * std::mem::size_of::<K>()) as u64
+                + self.iter().map(SizeBytes::heap_size_bytes).sum::<u64>()
         }
     }
 }
--- a/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
+++ b/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
@@ -158,6 +158,8 @@ impl ViewContextSystem for TransformTreeContext {
     fn execute_once_per_frame(
         ctx: &re_viewer_context::ViewerContext<'_>,
     ) -> ViewContextSystemOncePerFrameResult {
+        re_tracing::profile_function!();
+
         let caches = ctx.store_context.caches;
         let (transform_forest, transform_cache) =
             caches.entry(|c: &mut TransformDatabaseStoreCache| {


crates/store/re_tf/src/frame_id_registry.rs
```diff
@@ -1,27 +1,35 @@
...
 use re_sdk_types::{TransformFrameIdHash, archetypes};
 
-/// Frame id registry for resolving frame id hashes back to frame ids.
+/// Provides context around frame id hashes.
 pub struct FrameIdRegistry {
+    /// A lookup table for resolving frame id hashes back to frame ids.
     frame_id_lookup_table: IntMap<TransformFrameIdHash, TransformFrameId>,
+
+    /// Holds all frame ids per entity that were logged as a child frame.
+    child_frames_per_entity: IntMap<EntityPathHash, IntSet<TransformFrameIdHash>>,
```
Wumpf, line 15: curious, should we for our purposes here also register parent frames? 🤔 
practically in a ROS scenario that probably means just including the root, so guess it doesn't matter! :)

crates/store/re_tf/src/frame_id_registry.rs
Wumpf, line 1: happy to see adding this here is as compact as I hoped it would be!

crates/utils/re_byte_size/src/std_sizes.rs
```diff
@@ -67,6 +67,19 @@ impl<K: SizeBytes, V: SizeBytes, S> SizeBytes for HashMap<K, V, S> {
...
 
+impl<K: SizeBytes, S> SizeBytes for HashSet<K, S> {
+    #[inline]
+    fn heap_size_bytes(&self) -> u64 {
+        // NOTE: It's all on the heap at this point.
+
+        if K::is_pod() {
+            (self.len() * std::mem::size_of::<K>()) as _
+        } else {
+            self.iter().map(SizeBytes::total_size_bytes).sum::<u64>()
```
Wumpf: ~that's incorrect: in the non-pod case you want to _add_ this to `(self.capacity() * std::mem::size_of::<K>()) as _`~ I'm wrong. it's using `total_size_bytes` so that's mostly fine. but the capacity portion is still relevant imho

crates/utils/re_byte_size/src/std_sizes.rs
```diff
@@ -67,6 +67,19 @@ impl<K: SizeBytes, V: SizeBytes, S> SizeBytes for HashMap<K, V, S> {
     }
 }
 
+impl<K: SizeBytes, S> SizeBytes for HashSet<K, S> {
+    #[inline]
+    fn heap_size_bytes(&self) -> u64 {
+        // NOTE: It's all on the heap at this point.
+
+        if K::is_pod() {
+            (self.len() * std::mem::size_of::<K>()) as _
```
Wumpf: slight bug
```suggestion
            (self.capacity() * std::mem::size_of::<K>()) as _
```

crates/utils/re_byte_size/src/std_sizes.rs
Wumpf: oh. a lot of the implementaitons have this bug. Guess I'll look into cleaning that up

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -171,9 +176,15 @@ impl ViewContextSystem for TransformTreeContext {
             .iter_frame_ids()
             .map(|(k, v)| (*k, v.clone()));
 
+        let child_frames_per_entity = transform_cache
```
Wumpf, line 181: bit painful to clone even more stuff, but eh at least we're only doing it once per frame 🤷 
Time to add a profile scope to this method though

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -249,12 +260,24 @@ impl ViewContextSystem for TransformTreeContext {
...
             let frame_hash = TransformFrameIdHash::new(&frame);
-            if !self.frame_id_mapping.contains_key(&frame_hash) {
+            if !self.frame_id_hash_mapping.contains_key(&frame_hash) {
                 // As overrides are local to this view we need to clone the whole map to add new hashes.
-                Arc::make_mut(&mut self.frame_id_mapping).insert(frame_hash, frame);
+                Arc::make_mut(&mut self.frame_id_hash_mapping).insert(frame_hash, frame);
             }
         }
 
+        let lookup_image_plane_distance = |transform_frame_id_hash: TransformFrameIdHash| -> f64 {
```
Wumpf, line 271: thanks for cleaning up a lil bit!

crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
```diff
@@ -69,44 +71,70 @@ impl VisualizerSystem for TransformAxes3DVisualizer {
...
-                continue;
-            };
-
-            // Determine which transforms to draw axes at.
-            // For pinhole cameras, we draw at the pinhole location only.
-            // For normal entities, we iterate over all instance poses.
-            let transforms_to_draw: smallvec::SmallVec<[glam::Affine3A; 1]> =
-                if let Some(pinhole_tree_root_info) =
-                    transforms.pinhole_tree_root_info(transform_info.tree_root())
+            // We first add the axes for the entity transforms, because we want them to be drawn below
```
Wumpf, line 74: what do you mean by draw below, you mean before? We don't do any distance sorting on opaque objects yet so this _is_ probably the case, but I'm not sure we'll guarantee that in the future 🤔, well I do think if we change to front to back sorting for opaque we'll use a stable sort so if you end up with the exact same transform twice it should still work :) 🤞

crates/utils/re_byte_size/src/std_sizes.rs
Wumpf: let's take the whole thing separately

Wumpf: ... turns out only `HashSet` was incorrect. BTreeMap & BTreeSet don't have `capacity`. See
https://github.com/rerun-io/rerun/pull/12268

crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
grtlr: Yeah, I know that we don't do any depth sorting for 3D primitives. The draw order is important for the labels though: We don't want the entity frame (which is almost always implicit and arguably not as useful in this context) to cover the more meaningful transform labels.

Wumpf: ahhh. yes makes sense

PR #12263, opened 2025-12-17
Fix compaction of recordings containing video streams

emilk:
### Related
* Broke in https://github.com/rerun-io/rerun/pull/11921
* together with https://github.com/rerun-io/rerun/pull/12008
* Proper fix coming with RR-3163

### Problem
Repro:

```
rerun rrd compact some_rrd_with_video_stream.rrd > compacted.rrd
RUST_LOG=info rerun compacted.rrd
```

Before:

> WARN  re_viewer_context::cache::video_stream_cache] Late insertions of video frames within an established video stream is not supported, some video data has been ignored.

After: No such warning

### DIFF
--- a/crates/top/rerun/src/commands/rrd/merge_compact.rs
+++ b/crates/top/rerun/src/commands/rrd/merge_compact.rs
@@ -358,7 +358,7 @@ fn sort_chunks_by_timepoint(iter_chunks: impl Iterator<Item = Arc<Chunk>>) -> Ve
     // Sorting by time point ensures that chunks are inserted in temporal order,
     // at least if all timelines increase monotonically.
     // This is currently required by our video playback.
-    // TODO(#RR-3163): handle out-of-order video samples
+    // TODO(RR-3163): handle out-of-order video samples
     let mut chunks: Vec<Arc<Chunk>> = iter_chunks.collect();
     chunks.sort_by_cached_key(|chunk| chunk.min_time_point());
     chunks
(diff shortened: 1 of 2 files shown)


crates/top/rerun/src/commands/rrd/merge_compact.rs
```diff
@@ -348,3 +351,15 @@ fn merge_and_compact(
 
     Ok(())
 }
+
+fn sort_chunks_by_timepoint(iter_chunks: impl Iterator<Item = Arc<Chunk>>) -> Vec<Arc<Chunk>> {
+    re_tracing::profile_function!();
+
+    // Sorting by time point ensures that chunks are inserted in temporal order,
+    // at least if all timelines increase monotonically.
```
jleibs, line 359: What is our sort definition what timelines are inconsistent?

jleibs (approved):

crates/top/rerun/src/commands/rrd/merge_compact.rs
emilk: Sorting on timestamps will sort by one timeline at a time, in alphabetical order. If they are inconsistent, then the videos may or may not work, depending on what timeline they are on. This is not a solid solution, but will work for most realistic cases. @IsseW and @Wumpf has a better solution cooking as part of https://linear.app/rerun/issue/RR-3163

Wumpf: even if we fix the video handling, it seems like a good idea reagardless? I mean the more time sorted the chunks are the better for all sort of streaming and access? 🤔

emilk: True! I'm off today, but change the comment to that effect before you merge

Wumpf:
superseeded by https://github.com/rerun-io/rerun/pull/12270

After long discussion we decided that this is too superficial because it only handles the rrd compact command itself and not ingestion into the store via other routes. The fact it fixes our problem is more of a happy accident. We generally want to handle arbitrary order and overlapping chunks everywhere

PR #12265, opened 2025-12-18
Improve `VisualizerOverrides` ergonomics in Rust

grtlr:
### Related

* Part of RR-3153
* Follow up of #12182.
* [x] Requires #12262.

### What

This adds a `fn visualizer()` method to applicable archetypes to prevent hardcoding their identifiers. We don't have a Rust blueprint API yet, but this is still a nice contract to uphold.

### DIFF
--- a/crates/build/re_types_builder/src/codegen/rust/api.rs
+++ b/crates/build/re_types_builder/src/codegen/rust/api.rs
@@ -22,10 +22,10 @@ use crate::codegen::{Target, autogen_warning};
 use crate::objects::{EnumIntegerType, ObjectClass};
 use crate::{
     ATTR_DEFAULT, ATTR_RERUN_COMPONENT_OPTIONAL, ATTR_RERUN_COMPONENT_RECOMMENDED,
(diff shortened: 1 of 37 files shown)


crates/build/re_types_builder/src/codegen/rust/api.rs
```diff
@@ -1116,6 +1120,21 @@ fn quote_trait_impls_for_archetype(reporter: &Reporter, obj: &Object) -> TokenSt
         })
         .collect_vec();
 
+    let visualizer_override = attrs.get_string(ATTR_RERUN_VISUALIZER).map(|visualizer| {
+        let doc_lines = ["Returns the visualizer that corresponds to this archetype."];
```
Wumpf: nit: trying to be even more precise :)
```suggestion
        let doc_lines = ["Returns the visualizer type name that corresponds to this archetype."];
```

Wumpf (approved):
lgtm, but I think we'll want to evolve that to return the visualizer type component once we have that in place
(type system will make sure we don't forget it, so it's a good concept to land that we'll maintain as we go :))

grtlr:
I've made the change to `VisualizerType` and left `TODO`s in the tests.

PR #12268, opened 2025-12-18
Fix `SizeBytes::heap_size_bytes` for `HashMap`

Wumpf:
and add one for `HashSet`.


grtlr (approved):
Thanks! Need to update the snapshots though 😢

Wumpf (commented):
those size changes on the snapshots are actually way more than I expected. Makes me a lil bit nervous of what effects that may have down the line

FYI: @teh-cmc

Wumpf:
Thanks for checking. Will set up a tandem PR, but leaving that for next year

crates/utils/re_byte_size/src/std_sizes.rs
```diff
@@ -52,15 +54,17 @@ impl<K: SizeBytes, V: SizeBytes, S> SizeBytes for HashMap<K, V, S> {
...
+                + self.keys().map(SizeBytes::heap_size_bytes).sum::<u64>()
         };
 
         let values_size_bytes = if V::is_pod() {
-            (self.len() * std::mem::size_of::<V>()) as _
+            (self.capacity() * std::mem::size_of::<V>()) as _
         } else {
-            self.values().map(SizeBytes::total_size_bytes).sum::<u64>()
+            (self.capacity() * std::mem::size_of::<V>()) as u64
+                + self.values().map(SizeBytes::heap_size_bytes).sum::<u64>()
```
emilk, line 67: If we wanna be pedantic about this:

The capacity is defined as the number of elements we can put in there before re-allocating.

The `std` implementation (which uses the `hashbrown` crate) actually reserves 12.5% of the slots as always being empty, so to get an even preciser estimate, use something like

`let num_buckets = self.capacity() * 8 / 7;`

… and hope it doesn't change in the future :)

Reference:

https://github.com/rust-lang/hashbrown/blob/9037471eb241119de665eb328030f4b19c63dcbe/src/raw/mod.rs#L190-L191

… but in any case, `capacity` is a lot better than `len`. In other words, no need to be pedantic unless you really want to

Wumpf: > If we wanna be pedantic about this

I really don't want to, was hard enough to update all those snapshots 😅 
But good insight, thanks!

PR #12270, opened 2025-12-18
Disable merging compaction for video samples

teh-cmc:
See inline comment:
> The video decoder does not support overlapping chunks yet, and the
> combination of chunk splitting and chunk merging during compaction can
> easily lead to chunks that end up overlapping (hint: think about what
> happens when you split a bunch of chunks multiple times, and then later on
> the compaction system realizes that it can merges those back into one bigger
> chunk while still staying under max_rows/max_size budget).
> 
> To prevent that situation from happening, and until we implement support for
> overlapping chunks in the decoder itself, we simply disable compaction for
> video samples.
> Because video samples are already large by nature, we still want to do splitting,
> but we can live with the lack of merge for a while.


Wumpf (approved):
thanks!

Wumpf:
I'm fixing the test

PR #12271, opened 2025-12-18
Force utf8 output in compare_snippet_output.py

oxkitsune:
### What

Forces utf8 output on all streams in `compare_snippet_output.py`.
A better solution might be to set the following environment variable: `PYTHONUTF8=1`.


oxkitsune:
@rerun-bot full-check

oxkitsune:
@rerun-bot full-check

PR #12276, opened 2025-12-18
Fix image planes not working correctly in some named-transform frame scenarios

Wumpf:
### Related

* Fixes RR-3209

### What

Previously, when looking for the image plane distance of a pinhole camera we were looking for the entity with that coordinate frame. Not only is it inconsistent to how our pinhole camera visualization work via `child_frame`, but also means that if your pinhole doesn't have the right coordinate frame in general we never found the plane distance.

Made sure the adjusted test captures this problem now.

### DIFF
--- a/tests/rust/re_integration_test/tests/snapshots/origin_keypoint_3d.png
+++ b/tests/rust/re_integration_test/tests/snapshots/origin_keypoint_3d.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:844240f8e88c89525a12ef149674f461fe6e2ff994c8b68c8ba6350a8e41bdb5
-size 192339
+oid sha256:cd11d93cb254b2039fb7124d4cf5c91e52b4d75d2d71f10ae7ae25a0364dfa57
+size 191801


Wumpf:
that difference is curious. https://rerun-io.github.io/kitdiff/?url=https://github.com/rerun-io/rerun/pull/12276
should take some time to understand whether that's an expected sideffect or not

MichaelGrupp:
Looks a like an origin axis marker that's now appearing?

Wumpf:
that's the one I changed, but the failure looks like a change in eye position which is a bit odd

Wumpf:
eh okay whatever, looks really like only a bounding box changed for some reason 🤷

crates/viewer/re_view_spatial/tests/snapshots/2d_in_3d_with_explicit_frames.png
grtlr, line 1: _Wahh my tab did not have the convo above 🙈._

It's hard to notice, but there's a new transform frame in there in the center of the plane. Did we regress with my PR landing?

* https://github.com/rerun-io/rerun/pull/12262

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -293,6 +306,8 @@ impl ViewContextSystem for TransformTreeContext {
             )
             .collect::<Vec<_>>();
 
+        // TODO(andreas, grtlr): We should not re-query all those transforms. We still have them around in `tree_transforms_per_frame` (and a bit below in later `self.transform_infos`).
```
grtlr, line 309: 👍

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -268,16 +268,29 @@ impl ViewContextSystem for TransformTreeContext {
             }
         }
 
+        let caches = ctx.viewer_ctx.store_context.caches;
+        let transform_cache = caches.entry(|c: &mut TransformDatabaseStoreCache| {
+            c.read_lock_transform_cache(ctx.recording())
+        });
+
         let lookup_image_plane_distance = |transform_frame_id_hash: TransformFrameIdHash| -> f64 {
```
grtlr, line 276: Just realized this although I touched the code yesterday: Kinda cool that this is not a name clash in the compiler. 🤓

PR #12277, opened 2025-12-18
Video stream changes to support out of order samples

IsseW:
### Related

* Closes RR-3212
* Closes RR-3163

### What

Support out of order samples for video streams.

### Implementation details

- We can pre-allocate chunk samples from the rrd manifest, which are then filled in as that chunk arrives.
- We no longer keep track of GOPs the same way, and no longer rely on stable gop indices. Instead we keep track of keyframes, and instead of storing indices into this we keep it sorted and  binary search keyframes.
- Which chunk and buffer a frame should use is now stored within the frame itself.

### DIFF
--- a/crates/utils/re_video/src/decode/webcodecs.rs
+++ b/crates/utils/re_video/src/decode/webcodecs.rs
@@ -265,13 +265,6 @@ impl AsyncDecoder for WebVideoDecoder {
             .send(OutputCallbackMessage::Reset)
             .ok();
 
-        // For all we know, the first frame timestamp may have changed.
-        self.first_frame_pts = video_descr
-            .samples
-            .iter()
-            .find_map(|s| s.sample())
-            .map_or(Time::ZERO, |s| s.presentation_timestamp);
-
         if *IS_FIREFOX {
             // As of Firefox 140.0.4 we observe frequent tab crashes when calling `reset` on a video decoder.
             // See https://bugzilla.mozilla.org/show_bug.cgi?id=1976929 for more details.
@@ -286,6 +279,13 @@ impl AsyncDecoder for WebVideoDecoder {
             self.decoder = decoder;
             self.output_callback_tx = output_callback_tx;
         }
+
+        // For all we know, the first frame timestamp may have changed.
+        self.first_frame_pts = video_descr
+            .samples
+            .iter()
+            .find_map(|s| s.sample())
+            .map_or(Time::ZERO, |s| s.presentation_timestamp);
 
         let encoding_details = video_descr
             .encoding_details


IsseW:
Since our lints are failing because of `re_video` depending on `re_core_types`. I'm considering either making chunk id a generic parameter used to identify source.

Or we could consider moving re_video out of util.

Resolved by using tuid instead of chunk id

crates/store/re_data_loader/src/lerobot/datasetv3.rs
```diff
@@ -468,33 +468,40 @@ impl LeRobotDatasetV3 {
 
         // Find the GOPs that contain our time range
         let start_gop = video
```
Wumpf: below got renamed to `end_keyframe`. This is inconsistent now. should rename this as well

crates/utils/re_video/src/decode/webcodecs.rs
```diff
@@ -169,6 +169,7 @@ impl WebVideoDecoder {
         let first_frame_pts = video_descr
             .samples
             .front()
+            .and_then(|s| s.sample())
             .map_or(Time::ZERO, |s| s.presentation_timestamp);
```
Wumpf, line 172: not having a first_frame_pts is fine most of the time but may cause issues. We may want to come up with a different strategy of obtaining that offset :thinking:
See large comment block in submit_chunk below why we need it in the first place

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -621,16 +609,17 @@ impl VideoDataDescription {
             VideoDeliveryMethod::Stream { .. } => match self.samples.num_elements() {
                 0 => std::time::Duration::ZERO,
                 1 => {
-                    let first = self.samples.front()?;
+                    let first = self.samples.iter().find_map(|s| s.sample())?;
```
Wumpf, line 628: method documentation should point out how we handle absent samples and gaps in sample ranges

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -670,37 +659,81 @@ impl VideoDataDescription {
...
 
-        // Segments are guaranteed to be sorted among each other, but within a segment,
-        // presentation timestamps may not be sorted since this is sorted by decode timestamps.
-        Some(self.gops.iter().flat_map(move |seg| {
+        Some(
             self.samples
-                .iter_index_range_clamped(&seg.sample_range)
-                .map(|(_idx, sample)| sample.presentation_timestamp)
+                .iter()
+                .filter_map(|sample| Some(sample.sample()?.presentation_timestamp))
```
Wumpf, line 682: doc string should be updated to mention that reserved-only samples are simply ignored

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -670,37 +659,81 @@ impl VideoDataDescription {
...
     ) -> Option<SampleIndex> {
-        samples.latest_at_idx(|sample| sample.decode_timestamp, &decode_time)
+        // First find what keyframe this decode timestamp is in.
+        let keyframe_idx = keyframes
+            .partition_point(|p| {
+                samples
+                    .get(*p)
+                    .map(|s| s.sample())
+                    .inspect(|_s| {
+                        debug_assert!(_s.is_some(), "Keyframes should always be loaded");
```
Wumpf: nit:
```suggestion
                        debug_assert!(_s.is_some(), "Keyframes mentioned in the keyframe lookup list should always be loaded");
```

because otherwise future-me would be confused about what magic is supposed to load all keyframes in existance

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -670,37 +659,81 @@ impl VideoDataDescription {
...
+                    .flatten()
+                    .is_some_and(|s| s.decode_timestamp <= decode_time)
+            })
+            .checked_sub(1)?;
+
+        let start = *keyframes.get(keyframe_idx)?;
+        let end = keyframes
+            .get(keyframe_idx + 1)
+            .copied()
+            .unwrap_or_else(|| samples.next_index());
```
Wumpf: isn't that incorrect for the end of a video? Let's say I have a single keyframe and frame 0, followed by 32 frames, then the last sample should be the one with index 31

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -670,37 +659,81 @@ impl VideoDataDescription {
...
     /// For a given decode (!) timestamp, returns the index of the first sample whose
     /// decode timestamp is lesser than or equal to the given timestamp.
     fn latest_sample_index_at_decode_timestamp(
-        samples: &StableIndexDeque<SampleMetadata>,
+        keyframes: &[usize],
+        samples: &StableIndexDeque<SampleMetadataState>,
         decode_time: Time,
     ) -> Option<SampleIndex> {
-        samples.latest_at_idx(|sample| sample.decode_timestamp, &decode_time)
+        // First find what keyframe this decode timestamp is in.
```
Wumpf: took me a while to realize _why_ we're searching inside the keyframe list - it double serves as a quick index to what sample ranges are present which shortens the search quite a lot. Worth pointing out

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -670,37 +659,81 @@ impl VideoDataDescription {
...
+
+        Some(shown_idx)
     }
 
     /// See [`Self::latest_sample_index_at_presentation_timestamp`], split out for testing purposes.
+    ///
+    /// The returned sample index is guaranteed to be [`SampleMetadataState::Present`].
     fn latest_sample_index_at_presentation_timestamp_internal(
-        samples: &StableIndexDeque<SampleMetadata>,
+        keyframes: &[usize],
```
Wumpf:
```suggestion
        keyframes: &[KeyframeIndex],
```

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -719,8 +752,10 @@ impl VideoDataDescription {
         // since smaller presentation timestamps may still show up further back!
         let mut best_index = SampleIndex::MAX;
         let mut best_pts = Time::MIN;
-        for sample_idx in (0..=decode_sample_idx).rev() {
-            let sample = &samples[sample_idx];
+        for sample_idx in (samples.min_index()..=decode_sample_idx).rev() {
```
Wumpf, line 775: huh interesting, how did this ever work

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -767,57 +803,100 @@ impl VideoDataDescription {
...
+        }
     }
 
-    /// For a given decode (!) timestamp, return the index of the group of pictures (GOP) index containing the given timestamp.
-    pub fn gop_index_containing_decode_timestamp(&self, decode_time: Time) -> Option<GopIndex> {
-        self.gops.latest_at_idx(
-            |gop| self.samples[gop.sample_range.start].decode_timestamp,
-            &decode_time,
-        )
+    pub fn sample_keyframe_idx(&self, sample_idx: SampleIndex) -> Option<KeyframeIndex> {
```
Wumpf, line 838: doc string plz

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -767,57 +803,100 @@ impl VideoDataDescription {
...
-        self.gop_index_containing_decode_timestamp(
-            self.samples[requested_sample_index].decode_timestamp,
-        )
+                    false
+                }
+            })
+            .checked_sub(1)
+    }
+
+    /// For a given decode (!) timestamp, return the index of the keyframe index containing the given timestamp.
```
Wumpf, line 861: I'm wondering whether we could add some more unit testing around all those bookkeeping changes. We have some coverage, but doesn't seem all that much :/

crates/utils/re_video/src/stable_index_deque.rs
```diff
@@ -164,6 +180,11 @@ impl<T> StableIndexDeque<T> {
         self.vec.front_mut()
     }
 
+    /// See [`VecDeque::clear`].
```
Wumpf:
```suggestion
    /// See [`VecDeque::clear`].
    ///
    /// Does not reset the underlying index tracking.
```

crates/utils/re_video/src/stable_index_deque.rs
```diff
@@ -241,6 +262,35 @@ impl<T> StableIndexDeque<T> {
         self.vec.get(index.checked_sub(self.index_offset)?)
     }
 
+    /// See [`VecDeque::range`].
```
Wumpf: not quite true apparently since `VecDeque::range` doesn't take a filter function ;-)

crates/utils/re_video/src/stable_index_deque.rs
```diff
@@ -241,6 +262,35 @@ impl<T> StableIndexDeque<T> {
...
+    #[inline]
+    pub fn partition_point_in_range(
```
Wumpf: this one definitely deserves a unit test

crates/utils/re_video/src/stable_index_deque.rs
```diff
@@ -340,6 +367,17 @@ impl<T> StableIndexDeque<T> {
         let num_elements = range.end - range.start;
         self.iter_indexed().skip(range_start).take(num_elements)
     }
+
+    /// Mutable version of [`Self::iter_index_range_clamped`].
```
Wumpf: let's at least copy the doc test from `iter_index_range_clamped`

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -505,48 +508,50 @@ impl VideoPlayer {
...
-            return Err(InsufficientSampleDataError::ExpectedSampleNotAvailable.into());
-        };
-
-        self.enqueue_samples_of_gop(video_description, gop_idx, &gop.sample_range, video_buffers)
+        let sample_range = video_description.keyframe_indices[keyframe_idx]
+            ..video_description
+                .keyframe_indices
+                .get(keyframe_idx + 1)
+                .copied()
+                .unwrap_or_else(|| video_description.samples.next_index());
```
Wumpf: this "give me the GOP range starting at keyframe_idx X" happens quite a bit, worth considering having a utility for that if only to name the thing we're doing here

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -875,10 +954,13 @@ pub struct SampleMetadata {
     /// May be unknown if this is the last sample in an ongoing video stream.
     pub duration: Option<Time>,
 
-    /// Index of the data buffer in which this sample is stored.
-    pub buffer_index: usize,
+    /// The raw buffer this sample is in.
+    pub buffer: arrow::buffer::Buffer,
```
Wumpf: I know why we arrived at this, but I don't particularly like having arrow buffer here:
* it increases the ref count for that arrow buffer _a lot_
* it drags in libarrow dependency, until now re_video didn't need any arrow at all to do its work
* arrow buffer under the hood is already a byte span into a ref counted buffer, that makes the `byte_span` below kinda pointless. I.e. if we're using arrow buffers here we might as well _only_ use arrow buffers

It also means that `SampleMetadata` is no longer really metadata but really forms the actual sample, which is what distinguished it previously from `Chunk` which had a copy of the data for transfer to the decoder. I figure at that point we might as well rename it to `Sample` then.


... so either we check again whether we can not depend on arrow here, keeping stable buffer indices around, or we go all in and drop `byte_span`, rename the thing and at a later point maybe even remove `Chunk`

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -597,6 +602,10 @@ impl VideoPlayer {
             for (_, sample) in video_description.samples.iter_index_range_clamped(
                 &(sample_idx_end.saturating_sub(allowed_delay + 1)..sample_idx_end),
             ) {
+                let Some(sample) = sample.sample() else {
+                    continue;
+                };
```
Wumpf, line 620: should we distinguish unloaded here and immediately treat things as live stream if we encounter unloaded samples?

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -192,19 +161,156 @@ impl VideoStreamCache {
...
+                            event.chunk.entity_path()
+                        );
+                    }
+                }
+            }
+        }
+    }
+}
+
+struct VideoData {
```
Wumpf: what is this struct and why is it needed? Name doesn't really tell me what it is used for

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -83,16 +47,18 @@ impl PlayableVideoStream {
 struct VideoStreamCacheEntry {
     used_this_frame: AtomicBool,
     video_stream: Arc<RwLock<PlayableVideoStream>>,
+    known_chunk_offsets: BTreeMap<ChunkId, ChunkSamples>,
```
Wumpf: isn't this more like chunk sample ranges?

crates/utils/re_video/src/demux/mod.rs
IsseW: Guess we didn't really drop samples at the start very often?

crates/utils/re_video/src/demux/mod.rs
IsseW: Since this one just cares about the first/last present sample at most, I don't think gaps matter?

crates/utils/re_video/src/decode/webcodecs.rs
IsseW: In other places we have done `find_map` to get it, so did that here too.

IsseW: Although, since this is done on construction we might not have the first frame of the entire video at this point… So maybe we'll need to look this up just in time instead

crates/utils/re_video/src/stable_index_deque.rs
IsseW: method ended up being unused so removed it instead.

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
IsseW: Removed it as it wasn't really making the code easier to read, and was just used as a return argument for one function.

crates/viewer/re_renderer/src/video/player.rs
IsseW: Possibly, but also I think only video streams would have unloaded samples now? And the inner instant is updated on new store events, which would happen if we're actively loading incoming data.

So at least with how it works now, I don't think anything would change if we convert non-streams to streams if I understand what you mean correctly.

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -1024,68 +1032,63 @@ fn load_known_chunk_offsets(
...
+        if let Some((chunk, rrd_entry)) = next_chunk
+            && !all_loaded_chunks.contains(chunk)
+        {
+            let idx = data_descr.samples.next_index();
+            // `num_rows` gives us the maximal amount of frames in this chunk.
+            known_chunk_ranges.insert(
+                *chunk,
+                ChunkSampleRange {
+                    first_sample: idx,
+                    last_sample: idx + rrd_entry.num_rows as usize,
```
Wumpf: I think this is an off-by-one error - `last_sample` is documented to be inclusive, but we're adding number of rows, i.e. resulting in the end_sample

crates/utils/re_video/src/demux/mod.rs
```diff
@@ -221,6 +221,18 @@ impl re_byte_size::SizeBytes for VideoDataDescription {
 }
 
 impl VideoDataDescription {
+    /// Get the range of samples which use a keyframe, including the keyframe sample itself.
+    pub fn get_keyframe_sample_range(&self, keyframe_idx: usize) -> Option<std::ops::Range<usize>> {
```
Wumpf: nit: `keyframe_sample_range`

or maybe better and call it a group of pictures?
`gop_sample_range_for_keyframe`

crates/viewer/re_view_spatial/src/visualizers/video/video_stream.rs
```diff
@@ -153,10 +154,42 @@ impl VisualizerSystem for VideoStreamVisualizer {
...
+                    // We want to figure out the byte offsets of each blob within the arrow buffer that holds all the blobs,
+                    // i.e. get out a Vec<ByteRange>.
+                    let inner_list_array =
+                        raw_array.downcast_array_ref::<arrow::array::ListArray>()?;
+
+                    let values = inner_list_array
+                                .values()
+                                .downcast_array_ref::<arrow::array::PrimitiveArray<arrow::array::types::UInt8Type>>()?;
+
+                    let values = values.values().inner();
```
Wumpf: deja vu. Mabe we should add this as a static utility method somewhere in `re_arrow_util`?

crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs
```diff
@@ -978,43 +977,52 @@ impl Cache for VideoStreamCache {
     }
 }
 
+/// `loaded_chunks` should be sorted by start time, and internally sorted on `timeline`.
 fn load_known_chunk_offsets(
```
Wumpf:
```suggestion
fn load_known_chunk_ranges(
```

PR #12281, opened 2025-12-19
Ensure that `flush` waits for all data loaders to finish

grtlr:
### Related

* Makes #12260 work as expected.
* Closes RR-3214.
* Related to RR-3215.
* Also should fix the [indeterministic CI failures](https://github.com/rerun-io/rerun/actions/runs/20359071884/job/58501585339) that we have been experiencing.

### What

This changes the `flush` logic to wait for all data loader (threads) to finish, which is the intuitive thing to do.

### Misc

* [x] `@rerun-bot full-check`

### DIFF
--- a/docs/snippets/all/howto/load_urdf.py
+++ b/docs/snippets/all/howto/load_urdf.py
@@ -9,7 +9,7 @@ with RecordingStream("rerun_example_load_urdf") as rec:
     # `log_file_from_path` automatically uses the built-in URDF data-loader.
     urdf_path = Path(__file__).parent / "minimal.urdf"
     rec.log_file_from_path(urdf_path, static=True)
-    
+
     # The `flush` call is optional, but it helps with logging consistency,
     # because it ensures that the URDF finishes loading before continuing.
     rec.flush()


grtlr:
@rerun-bot full-check

docs/snippets/all/howto/load_urdf.py
```diff
@@ -9,7 +9,7 @@
     # `log_file_from_path` automatically uses the built-in URDF data-loader.
     urdf_path = Path(__file__).parent / "minimal.urdf"
     rec.log_file_from_path(urdf_path, static=True)
-    # Ensure the URDF is fully processed for testing consistency.
+    # Wait for the URDF to finish loading before continuing.
```
ntjohnson1: From an end user perspective I don't think this is necessary. The rrd is the same both ways for the viewer. It is something just because rrd compare is unhappy so I do like some mention this is just for us. Or maybe a note that if logging order is important they do this. I don't want them to think they HAVE to.

Wumpf (approved):
how convenient that we already had `wait_for_dataloaders`

PR #12284, opened 2025-12-19
Fix video errors having black rectangle around them

Wumpf:
### Related

* RR-3218

### What

In particular in light mode this was very ugly!

0.28.0:
image

After:
image


ntjohnson1 (approved):
Looks much nicer!

Wumpf:
have regression test piggy backing on existing integration test with a simple color check. coming in a bit!

PR #12147, opened 2025-12-09
Add support for server side filtering of DataFusion DataFrames

timsaucer:
This PR adds support for server side (push down) filtering of DataFrames from a dataset query. Specifically it adds support for `rerun_segment_id` and time index. With this PR users can do complicated queries such as the following:

```python
df = view.reader(index="log_tick").filter(
    (col("rerun_segment_id") == "recording_0") & (col("log_tick") == 123456)
)
```

This will be converted into an update in the dataset query or will produce multiple queries as needed.

All of the results from the queries will be combined to find a unique set of required chunks and then sent out to the appropriate DataFusion partitions for requesting chunks from the server.

This PR adds support for boolean logic (AND/OR) to combine the segment and time parameters. Additionally it supports `functions.in_list` and `Expr.between`. If the filters include a non-supported expression, then we will fall back to querying all data.

PR #12285, opened 2025-12-19
Add redap tests for dataframe pushdown

timsaucer:
This is follow on work to #12147 . 

This adds explicit tests to see what data come back from our push down filters.

The difference between this and the e2e tests is that the e2e tests will check the result of the full output from datafusion collection. That will go through the FilterExec after our table provider, so the data filtered could be done by datafusion rather than our table provider.

We cannot mark our table provider as `Exact` because it will send back chunks that have more data than was requested.

This test will get record batches directly from our table provider without going through any of the other mechanisms of DataFusion. We can verify we get exactly the chunks we expect from the server and no more.

In addition to these checks I am adding a client util. Since the `TableProvider` requires a `ConnectionClient` we need a little more than just the server hook.

Also I updated one method in the `TableProvider` that removes a `block_on` call that makes integration with testing difficult.

### DIFF
--- a/crates/store/re_redap_tests/Cargo.toml
+++ b/crates/store/re_redap_tests/Cargo.toml
@@ -14,11 +14,9 @@ version.workspace = true
 
 [features]
 default = []
-perf_telemetry = ["re_redap_client/perf_telemetry", "re_perf_telemetry"]
 
 [lints]
 workspace = true
-
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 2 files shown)


timsaucer:
Since I ended up with a fair amount of scope creep on this PR, I'm going to open another one with just the updates to the OSS server and existing tests. Then this PR will rebase with only the new tests.

https://github.com/rerun-io/rerun/pull/12328

timsaucer:
@rerun-bot reality-sync

timsaucer:
@rerun-bot reality-sync

PR #12286, opened 2025-12-19
Document that URDF files can be loaded into existing recording

MichaelGrupp:
It's cool feature that deserves to be documented!


MichaelGrupp:
Rendered version: https://landing-ivfqzadvk-rerun.vercel.app/docs/howto/urdf#load-urdf-into-an-existing-recording

Wumpf (commented):
👍

PR #12287, opened 2025-12-19
Add dataloader for encoded depth image (to allow `log_from_file` & dragdrop)

Wumpf:
### Related

* Small follow-up to https://github.com/rerun-io/rerun/pull/11877

### What

Allows drag & drop for encoded depth images `rvl` files into the Viewer.

### DIFF
--- a/crates/store/re_data_loader/src/loader_archetype.rs
+++ b/crates/store/re_data_loader/src/loader_archetype.rs
@@ -114,6 +114,14 @@ impl DataLoader for ArchetypeLoader {
                 entity_path,
                 contents.into_owned(),
             )?);
+        } else if crate::SUPPORTED_DEPTH_IMAGE_EXTENSIONS.contains(&extension.as_str()) {
+            re_log::debug!(?filepath, loader = self.name(), "Loading depth image…",);
+            rows.extend(load_depth_image(
+                &filepath,
+                timepoint,
+                entity_path,
+                contents.into_owned(),
+            )?);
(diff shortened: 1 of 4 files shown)


crates/store/re_data_loader/src/loader_archetype.rs
```diff
@@ -191,6 +199,35 @@ fn load_image(
...
+    timepoint: TimePoint,
+    entity_path: EntityPath,
+    contents: Vec<u8>,
+) -> Result<impl ExactSizeIterator<Item = Chunk> + use<>, DataLoaderError> {
+    re_tracing::profile_function!();
+
+    let rows = [{
+        let mut arch = re_sdk_types::archetypes::EncodedDepthImage::from_file_contents(contents);
+
+        if let Ok(format) = image::ImageFormat::from_path(filepath) {
```
oxkitsune, line 213: If it's a `.png` depth image, we should only accept it here if it contains a exactly 1 channel. RGB pngs should be handled by `EncodedImage`

Wumpf: good point. Practically we won't get here because pngs are first handled by the image handler. Maybe we should stick to that explicitly, i.e. adding only rvl to the depth image extensions and then here also only handling rvl

oxkitsune: Yea that sounds reasonable to me. 

note that `.rvl` is not really an official file extension, and in the current state we only support the ROS specific format. So do we need this?

PR #12289, opened 2025-12-19
Fix RVL depth decoding for non-quantized images

oxkitsune:
### What

Fix RVL depth decoding to only apply inverse quantization when present.
Zero disparity is treated as NaN only for quantized (32FC1) images; non-quantized (16UC1) values now pass through unchanged.


Wumpf (approved):
great catch. Thanks for handling it and adding a unit test!

PR #12291, opened 2025-12-19
Fix mypy type issue in ARKit Scenes example

oxkitsune:
### What

Fix mypy lint error in ARKit Scenes example by casting `visit_id` to `float` to force the return type of `.iat`.


ntjohnson1:
This still seems unhappy. Can you just use typing cast if we know what it is? Or is there something else weird here?

ntjohnson1:
I ended up basically merging the same change Friday to make CI green https://github.com/rerun-io/rerun/pull/12293

oxkitsune:
Yea I saw, I'll see if I can make this green without type ignores.

PR #12294, opened 2025-12-20
docs: Fix parameter name from 'color' to 'colors'

Ivan-Zhong:
### Related

None.

### What

Fixed a small typo in the documentation code example. The parameter name was wrong which might confuse new users.


ntjohnson1:
Thanks for the catch!

PR #12296, opened 2025-12-21
Run nightly clippy and fix a few things

emilk:


crates/store/re_datafusion/src/dataframe_query_provider.rs
```diff
@@ -404,7 +404,7 @@ async fn chunk_store_cpu_worker_thread(
                 }
             }
 
-            let current_stores = current_stores.get_or_insert({
+            let current_stores = current_stores.get_or_insert_with(|| {
```
emilk, line 407: This might actually save us quite a lot of overhead

Issue #12295, opened 2025-12-21
panning breaks all timeseries view ranges


rrad5409:
**Describe the bug**

Manually panning the `TimeSeries` view causes all time series view ranges to break, and they can't get changed away from absolute time.
Changing to other time ranges (current time, beginning/end of timeline) will work for a single frame, but gets forced back to absolute time as soon as the viewer updates.
This breaks for the **entire viewer application**, for _all other timeseries views_, both _independent_ and _global_ view ranges, as well as _all recordings_ (current and future), requiring a full viewer restart to fix.

**To Reproduce**

1. Open a timeseries view
2. Scroll or drag to pan the view
3. Observe that it will now get stuck on absolute time

**Expected behavior**

... it doesn't get stuck ?

**Screen Recordings**

https://github.com/user-attachments/assets/ce38a0e1-24ae-419b-a6e1-a1c1dcc786ab

**Desktop (please complete the following information):**
 - OS: Ubuntu 25.04
 - Viewer: 0.28.1

**Rerun version**

```
rerun-cli 0.28.1 (base lance map_view nasm native_viewer oss_server perf_telemetry re_server release_full release_no_web_viewer web_viewer) [rustc 1.88.0 (6b00bc388 2025-06-23), LLVM 20.1.5] x86_64-unknown-linux-gnu prepare-release-0.28.1 ac98620, built 2025-12-19T21:06:08.748579589Z
Video features: av1 ffmpeg nasm
```

ntjohnson1:
Thanks for pointing this out!

It looks like this bug was introduced here https://github.com/rerun-io/rerun/pull/12208

ntjohnson1:
Ok I have a fix up in PR right now. We'll land that to main today but we won't cut a new patch release (0.28.2) until ~Friday.

ntjohnson1:
There's one other issue that came up I'd like to get in the patch so I'm going to bump back the patch release into next week.

ntjohnson1:
A few more things popped up so I didn't get this out this week. Sorry about that. More people are back online next week so we should hopefully get a 0.28.2 out shortly.

rrad5409:
All good, I've been ~suffering~ working on other things anyway

ntjohnson1:
Closed and fixed on main https://github.com/rerun-io/rerun/pull/12297 patch release pending

Issue #12193, opened 2025-12-12
Time series range drifts while panning


aedm:
# Reproduction steps

1. Open some time series data
2. Pan it right
At each frame when the mouse moves right, time range narrows: 

https://github.com/user-attachments/assets/4b91cf3b-af49-42bb-b6e2-4192aba996ff

It's visible in the "Plots" example, too:

https://github.com/user-attachments/assets/0669e792-9c64-4a8d-8663-43a01660d734

PR #12297, opened 2025-12-22
Don't force absolute time views after panning

ntjohnson1:
Should solve #12295
Related to #12208
Looks like it doesn't break #12193

I believe our updated rounding logic avoids the accumulation of floating point errors over time which was the original issue. It looks like forcing the new plot x was enforcing the absolute time instead of encoding all our different view variants. We could probably keep the float offset for precision and have a match statement but I don't know this area well enough without doing more digging.

https://github.com/user-attachments/assets/3c26b8e9-57ec-49bb-8598-194b91107aa6


ntjohnson1:
@IsseW if you wanted to take a look at this to confirm.

IsseW:
This specific fix of saving a smaller offset fixed being able to pan while being zoomed in on sequence timelines, but I do think this PR was correct since s you say the rounding logic actually fixed the original issue.

PR #12302, opened 2025-12-26
Add redirects for removed components

ntjohnson1:
Try to fix nightly link checker errors.


ntjohnson1:
Manually verified re-directs on vercel preview.

PR #12308, opened 2026-01-05
Feature-flagged chunk-prefetching

emilk:
Part of larger-than-RAM.

Feature-flagged in app settings.


crates/viewer/re_viewer/src/prefetch_chunks.rs
```diff
@@ -0,0 +1,89 @@
...
+        // Ideally it should depend on the actual channel bandwidth and latency.
+        delta_byte_budget: 500_000,
+    };
+
+    let data_source = recording.data_source.as_ref()?;
+    let rrd_manifest = &mut recording.rrd_manifest_index;
+
+    if !rrd_manifest.has_manifest() {
+        return None;
+    }
```
emilk, line 52: This is the feature-flag

PR #12309, opened 2026-01-05
Fix bad error reporting when registering a (faulty) RRD URI built from a relative path

abey79:
### Related

* fixes https://linear.app/rerun/issue/RR-3234/calling-registration-with-relative-path-on-oss-server-leads-to-opaque
* sibling https://github.com/rerun-io/dataplatform/pull/2226

### What

Sticking `"file://` in front of a relative path is a footgun, because the resulting URI is valid, but the first path element is considered a hostname (uri in the form of `file://hostname/path/to/file.txt`). These were previously accepted by `/RegisterWithDataset` and silently dropped by the OSS server.

This PR:
- hardens the URI handling in OSS server
- adds a redap test to ensure `not_found` errors are returned for bad URIs


grtlr:
I've removed the `consider-patch` label as per our discussion.

PR #12310, opened 2026-01-05
Add Python support for `UrdfTree`

oxkitsune:
### Related

* Closes RR-3141

### What

This adds:

- Python bindings for our internal `UrdfTree` types
- Python utilities for computing transforms using urdf
- A new python example using the new python apis

### DIFF
--- a/docs/content/howto/urdf.md
+++ b/docs/content/howto/urdf.md
@@ -55,7 +55,7 @@ joints = urdf_tree.joints()
 # Lookup by name
 urdf_tree.get_joint_by_name("shoulder")
 urdf_tree.get_link_by_name("base_link")
-urdf_tree.get_link_path_by_name("end_effector")  # Entity path for logging
+urdf_tree.get_link_path_by_name("end_effector")  # Entity path for logging additional data e.g. images
 ```
 
 ### UrdfJoint


examples/python/animated_urdf/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+```python
+import rerun as rr
+import rerun.urdf as rru
+
+# Log the URDF file once, as a static resource
+rec.log_file_from_path(urdf_path, static=True)
+
+# Load the URDF tree structure into memory
+urdf_tree = rru.UrdfTree.from_file_path(urdf_path)
```
MichaelGrupp: Since we only access the `UrdfTree` from the module here, I think this looks nicer than "rru":

```suggestion
from rerun.urdf import UrdfTree

# Log the URDF file once, as a static resource
rec.log_file_from_path(urdf_path, static=True)

# Load the URDF tree structure into memory
urdf_tree = UrdfTree.from_file_path(urdf_path)
```

rerun_py/rerun_sdk/rerun/urdf.py
```diff
@@ -0,0 +1,346 @@
...
+    def name(self) -> str:
+        """Name of the link."""
+        return self._inner.name
+
+    def __repr__(self) -> str:
+        return self._inner.__repr__()
+
+
+class UrdfTree:
+    """A URDF robot model with joints and links."""
```
nikolausWest: I wonder if we perhaps need to highlight here that this thing isn't an archetype and isn't loggable (although perhaps we should make it loggable?). If we don't make it loggable I think docs need to highlight that this is a utility class and what the purpose of that utility class is

docs/content/howto/urdf.md
```diff
@@ -31,13 +31,54 @@ Here is an example that demonstrates how to load and update a `URDF` with the Py
 
 snippet: howto/load_urdf
 
-For similar code in Rust, we have a full example [here](https://github.com/rerun-io/rerun/tree/main/examples/rust/animated_urdf).
+For a full animation example, see the [Python animated URDF example](https://github.com/rerun-io/rerun/tree/main/examples/python/animated_urdf). There's also a [Rust example](https://github.com/rerun-io/rerun/tree/main/examples/rust/animated_urdf).
+
+## Python API
```
MichaelGrupp: "Python API" reads a bit broad, I would narrow this a bit down since it's about an optional utility.
```suggestion
## URDF utility API (Python)
```

docs/content/howto/urdf.md
```diff
@@ -31,13 +31,54 @@ Here is an example that demonstrates how to load and update a `URDF` with the Py
...
+
+Rerun provides the [`rr.urdf`](../../../rerun_py/rerun_sdk/rerun/urdf.py) module for programmatically inspecting and animating URDF models.
```
MichaelGrupp: Could be a bit longer intro:
```suggestion
Rerun provides the [`rr.urdf`](../../../rerun_py/rerun_sdk/rerun/urdf.py) Python module that can facilitate the handling of URDF models in your code.
It can be used as an alternative to other 3rd-party packages like [yourdfpy](https://yourdfpy.readthedocs.io/en/latest/index.html) or [pytransforms3d](https://dfki-ric.github.io/pytransform3d/index.html).

As shown below, you can use it e.g. to access individual joints of the URDF model and to compute their respective transforms based on joint states (e.g. angles for revolute joints).
These transforms can be directly sent to Rerun.
```

examples/python/animated_urdf/main.py
```diff
@@ -0,0 +1,55 @@
...
+                # Usually this angle would come from a measurement, here we just fake something
+                sin_value = math.sin(step * (0.02 + joint_index / 100.0))
+
+                # Remap from [-1, 1] to the joint's valid range
+                dynamic_angle = joint.limit_lower + (sin_value + 1.0) / 2.0 * (joint.limit_upper - joint.limit_lower)
+
+                # Compute the full transform including origin and parent/child frames
+
+                # Rerun loads the URDF transforms with child/parent frame relations.
+                # To move a joint, we just need to log a new transform between those frames.
```
MichaelGrupp: That comment above looks out of place.
```suggestion
                # Rerun loads the URDF transforms with child/parent frame relations.
                # To move a joint, we just need to log a new transform between those frames.
                # Here, we use the `compute_transform` method that automatically takes care
                # of setting the frame names and calculating the full transform from the joint angle. 
```

rerun_py/rerun_sdk/rerun/urdf.py
```diff
@@ -0,0 +1,346 @@
...
+        from . import Transform3D
+        from .datatypes import Quaternion
+
+        joint_type = self.joint_type
+
+        if joint_type in ("revolute", "continuous"):
+            # Revolute and continuous joints rotate around their axis
+            if joint_type == "revolute":
+                # Check limits only for revolute (continuous has no limits)
+                if not (self.limit_lower <= angle <= self.limit_upper):
```
MichaelGrupp: Can be joined with outer `if` to reduce nesting.
```python
if joint_type == "revolute" and not (self.limit_lower <= angle <= self.limit_upper):
```

rerun_py/rerun_sdk/rerun/urdf.py
```diff
@@ -0,0 +1,346 @@
...
+                    translation=translation,
+                    parent_frame=self.parent_link,
+                    child_frame=self.child_link,
+                )
+            else:
+                return Transform3D(
+                    translation=translation,
+                    parent_frame=self.parent_link,
+                    child_frame=self.child_link,
+                )
```
MichaelGrupp: Can be simplified? (same below for fixed joint)
```suggestion
            quat = None
            if roll != 0.0 or pitch != 0.0 or yaw != 0.0:
                quat = Quaternion(xyzw=_euler_to_quat(roll, pitch, yaw))
            return Transform3D(
                translation=translation,
                quaternion=quat,
                parent_frame=self.parent_link,
                child_frame=self.child_link,
            )
```

rerun_py/rerun_sdk/rerun/urdf.py
```diff
@@ -0,0 +1,346 @@
...
+                    translation=self.origin_xyz,
+                    parent_frame=self.parent_link,
+                    child_frame=self.child_link,
+                )
+            else:
+                return Transform3D(
+                    translation=self.origin_xyz,
+                    parent_frame=self.parent_link,
+                    child_frame=self.child_link,
+                )
```
MichaelGrupp:
```suggestion
            quat = None
            if roll != 0.0 or pitch != 0.0 or yaw != 0.0:
                quat = Quaternion(xyzw=_euler_to_quat(roll, pitch, yaw))
            return Transform3D(
                translation=self.origin_xyz,
                quaternion=quat,
                parent_frame=self.parent_link,
                child_frame=self.child_link,
            )
```

examples/python/animated_urdf/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Animate joints by logging transforms
+for joint in urdf_tree.joints():
+    if joint.joint_type == "revolute":
+        # compute_transform gives you a complete transform ready to log
+        # (includes origin, dynamic motion, and frame relationships)
```
MichaelGrupp: "dynamic motion" sounds a bit strange here.
```suggestion
        # compute_transform gives you a complete transform that is ready to log,
        # calculated from joint origin and the current angle and with the frame names set.
```

rerun_py/rerun_sdk/rerun/urdf.py
```diff
@@ -0,0 +1,346 @@
...
+
+def _quat_multiply(q1: list[float], q2: list[float]) -> list[float]:
+    """Multiply two quaternions in XYZW format."""
+    x1, y1, z1, w1 = q1
+    x2, y2, z2, w2 = q2
+
+    w = w1 * w2 - x1 * x2 - y1 * y2 - z1 * z2
+    x = w1 * x2 + x1 * w2 + y1 * z2 - z1 * y2
+    y = w1 * y2 - x1 * z2 + y1 * w2 + z1 * x2
+    z = w1 * z2 + x1 * y2 - y1 * x2 + z1 * w2
```
MichaelGrupp, line 339: Not really optimal that we need a custom implementation here, but I guess we have to keep the dependencies low.

oxkitsune: Yea agreed it's rough but I'd rather have this than another dependency. 

However, maybe (🤞) we already depend on scipy and we can just use that

docs/content/howto/urdf.md
```diff
@@ -31,13 +31,54 @@ Here is an example that demonstrates how to load and update a `URDF` with the Py
...
+
+# Access properties
+robot_name = urdf_tree.name
+root_link = urdf_tree.root_link()
+joints = urdf_tree.joints()
+
+# Lookup by name
+urdf_tree.get_joint_by_name("shoulder")
+urdf_tree.get_link_by_name("base_link")
+urdf_tree.get_link_path_by_name("end_effector")  # Entity path for logging
```
MichaelGrupp: Not sure what the use case for the entity path of a link is, as it's not needed for updating transforms. I would either remove that line to avoid confusion or add a longer comment.

rerun_py/rerun_sdk/rerun/urdf.py
MichaelGrupp: I checked already, but it looks like scipy is only used in examples 😢

docs/content/howto/urdf.md
oxkitsune: Could be used for logging e.g. images to camera frame or something like that. I think it'd be useful

rerun_py/rerun_sdk/rerun/urdf.py
oxkitsune: Ideally we make `UrdfTree` loggable, I'm not sure how easy it is to add that. I've added a note explaining that this type isn't loggable directly for now.

docs/content/howto/urdf.md
MichaelGrupp: Ok, makes sense then. But would be good to have that example use case explained in the comment IMO.

MichaelGrupp (approved):
Super useful for our examples and looks good to me overall (feel free to address the remaining open comment).

PR #12311, opened 2026-01-05
Fix redap test snapshots

abey79:
Fix test on main.

Also sneak a minor fix for an error formatting.


abey79:
> LGTM. Full check is probably not required, right?

Dont think so no.

PR #12306, opened 2025-12-31
Fix chunk row_sliced for complex datatypes

ntjohnson1:
### Related
#12304, #12305

### What
When slicing a chunk we slice the components which are arrow ListArrays. Because of the layout slicing a listArray of structArrays doesn't slice the underlying children.

That means if we have 3 rows of 500KB tensors in one chunk slicing out each row leads to an RRD of 4500KB (all data copied in triplicate).

This works and passes a test. Validated on the example script linked in the issues above but currently as draft to highlight this probably doesn't pass the requisite quality bar.

The work around for the ListArray memory calculation was a misunderstanding about this slice error so I removed that (tests still pass) and I yelled at a robot to get something functional as POC. This probably needs discussion next week or insight from someone more familiar with arrow (this currently special cases ListArray, StructArray, and UnionArray since those are in tensors where this was discovered but there's probably a more generic recursive slice that doesn't require special casing).

PR #12312, opened 2026-01-05
Introduce deep Chunk slicing

### LINKED ISSUE #12304: Logging tensors to rerun following v0.22.1 -> v0.28.1 upgrade balloons .rrd size
**Describe the bug**

I upgraded the rerun version I was using from 0.22.1 -> 0.28.1 and noticed that the same recordings became about ~50x larger in size. For .rrd files that were previously in the <100MB range, they are now several GB, making them prohibitively large for use with the web viewer. 

In the minimal example below, logging 5000 tensors creates recordings that take 39.5 MB memory / 21MB .rrd (0.22.1) and 2.2GB memory / 1.19GB .rrd (0.28.1).

For the minimal examples (as with my other code), I iteratively log tensors since there does not appear to be a way to batch log them yet: https://github.com/rerun-io/rerun/issues/8361

**To Reproduce**

For rerun v0.22.1:
```
import rerun as rr
import numpy as np
import time

rr.init("testing")
rr.connect_tcp()

for i in range(5000):
    rr.set_time_nanos("local_ts_us", int(time.time() * 1e6))
    x = np.random.randint(0, 50000, size=1024)
    rr.log("test/tensors", rr.Tensor(x))
```

For rerun v0.28.1:

```
import rerun as rr
import numpy as np
import time

rr.init("testing")
rr.connect_grpc()

for i in range(5000):
    rr.set_time("local_ts_us", timestamp=np.datetime64(int(time.time() * 1e6), "us"))
    x = np.random.randint(0, 50000, size=1024)
    rr.log("test/tensors", rr.Tensor(x))
```

I then view & save the recording from the rerun viewer menu to compare the sizes

**Expected behavior**

Would've expected these .rrd files which contain the same data to remain at the <100MB size.

MacOS 15.6.1

**Rerun version**
v0.28.1

### LINKED ISSUE #12305: Tensor data breaks when saved to .rrd and loaded back into viewer, leaving most data unviewable
**Describe the bug**

After logging tensor data to the rerun reviewer, playback and interaction with it works perfectly. However, after saving the recording to file and attempting to load it back into the viewer, only a subset of the frames can be navigated to and visualized at all. In most places, it appears stuck on a prior tensor and the timeline scrub & step does nothing.

This appears to be a regression, as this does not happen on v0.21.1, which I just upgraded from. Not sure if unique to tensor data. 

**To Reproduce**

1. Build recording of tensors
2. Log to native viewer
3. Save recording to .rrd file
4. Load recording from the .rrd file saved above
5. Attempt to scrub or step through recording, data does not display

Recording created with:
```
import rerun as rr
import numpy as np
import time

rr.init("testing")
rr.connect_grpc()

for i in range(300):
    rr.set_time("local_ts_us", timestamp=np.datetime64(int((time.time() + i * 100)* 1e6), "us"))
    x = np.random.randint(0, 50000, size=1000)
    rr.log("test/tensors", rr.Tensor(x))
```

**Expected behavior**
The recording playback experience should be the same between freshly logged data and the saved .rrd file.

**Screenshots**

Screen recording of reproduction: https://www.youtube.com/watch?v=q3nXyLFfMuM

 MacOS Sequoia 15.6.1

**Rerun version**
0.28.1

teh-cmc:
## Why?

We are facing two distinct problems when dealing with sliced chunks:
1. Their size on disk is much larger than it should be, owing to the fact that data that is not part of the slice still gets serialized along with it.
1. Their reported physical in-memory size is also much larger than it should be, leading to the chunk splitter in the `ChunkStore` splitting chunks way more than it ever should.

The two issues, while distinct, are very related and together form a feedback loop that effectively turns the chunk splitter into a zip-bomb (well, an RRD-bomb).
Both of these issues vary in intensity depending on the datatype and nested-ness of the data (because it all has to do with offsets in the end).

### Issue 1: too much data on disk

The core of issue number 1 is that `Array::slice` is shallow, by which we really mean that it is guaranteed to run in O(1)-ish time. It achieves this by never actually physically slicing data, because doing so would require rewriting all the offsets that point at that data in the first place.
And since it doesn't patch the offsets, it must also serialize the entirety of the original data when writing to disk, not just the sliced part, else the offsets wouldn't make sense anymore.

Looking at the low-level slicing implementation in `ArrayData` gives a clear picture:
```rust
pub fn slice(&self, offset: usize, length: usize) -> ArrayData {
    if let DataType::Struct(_) = self.data_type() {
        let new_offset = self.offset + offset;
        ArrayData {
            data_type: self.data_type().clone(),
            len: length,
            offset: new_offset,
            buffers: self.buffers.clone(),
            child_data: self
                .child_data()
                .iter()
                .map(|data| data.slice(offset, length))
                .collect(),
            nulls: self.nulls.as_ref().map(|x| x.slice(offset, length)),
        }
    } else {
        let mut new_data = self.clone();

        new_data.len = length;
        new_data.offset = offset + self.offset;
        new_data.nulls = self.nulls.as_ref().map(|x| x.slice(offset, length));

        new_data
    }
}
```
Note how all the buffers are cloned as-is while the offsets, although sliced, are left untouched.

To make matters more confusing, there seem to exist a few datatype specific exceptions. Consider a pretty simple `list[i32]` for example:
```
array[0..]:          len=    100000  buf_size=    800004  data.slice_size=    800000 / IPC=825544
slice[25000..75000]: len=     50000  buf_size=    800004  data.slice_size=    600000 / IPC=413000
 deep[25000..75000]: len=     50000  buf_size=    400064  data.slice_size=    400000 / IPC=413000
```
Note how the reported in-memory sliced size (`data.slice_size`) for the shallow slice says `600000`, and yet somehow it ends up at `413000` once serialized to IPC. It seems that, in this specific case, the IPC writer is rewriting the offsets just-in-time in order to properly pack the data.

This is not always the case though, consider e.g. a `union#dense{u8,f32,i64}`:
```
array[0..]:          len=    100000  buf_size=   1000008  data.slice_size=   1000008 / IPC=1021832
slice[25000..75000]: len=     50000  buf_size=   1000008  data.slice_size=    750008 / IPC=771848
 deep[25000..75000]: len=     50000  buf_size=    900096  data.slice_size=    466670 / IPC=473864
```
Here we can see that the in-memory slice size reported for the shallow slice does match what eventually gets serialized to disk, except it's way more than it should be. Presumably because neither the shallow slicer nor the IPC writer make any attempt at removing unnecessary values from the underlying dense union arrays. Which makes sense, re-packing the underlying union arrays is a pretty heavy operation.

Anyway, what's important is that is this more of a fundamental design limitation rather than a bug: if you're not willing to hot patch offsets, then you cannot possibly serialize just the data you need. Therefore, we need a proper deep-slice implementation no matter what.

### Issue 2: reported physical in-memory size is completely off

The second issue, I believe, has more to do with bugs than fundamental design decisions. Indeed, I cannot think of any reason why [`ArrayData::get_slice_memory_size`](https://docs.rs/arrow/latest/arrow/array/struct.ArrayData.html#method.get_slice_memory_size) couldn't be made to work properly with all kinds of slicing setups.
Today, though, it seems to struggle a lot with anything that makes use of offsets (which is basically everything besides basic primitive arrays).

It is trivial to end up with extreme discrepancies by stacking multiple layers of offsets on top of each other. Here's a `list[list[i32]]`:
```
array[0..]:          len=    100000  buf_size=   1200008  data.slice_size=   1200000 / IPC=1238280
slice[ 5000.. 5005]: len=         5  buf_size=   1200008  data.slice_size=    800020 / IPC=   904
 deep[ 5000.. 5005]: len=         5  buf_size=       192  data.slice_size=        60 / IPC=   904
```
Note how far off the reported slice size is for the shallow case.

## What

This PR implements a slow but reliable `deep_slice_array` routine.

It works similar to `concat` in that it just allocates and copies what it needs. While technically much slower that it should be, this is A) likely not gonna matter in practice (`concat()` performs fine, after all) and, more importantly, B) this will serve as a baseline if and when we decide to optimize.
The entire implementation fits in a few lines, which literally just recreates an array from the ground up with just the desired slice of data, packed as much as possible.
```rust
/// Deep-slicing operation for Arrow arrays.
///
/// The data, offsets, bitmaps and any other buffers required will be reallocated, copied around, and patched
/// as much as required so that the resulting physical data becomes as packed as possible for the desired slice.
pub fn deep_slice_array<T: Array + From<ArrayData>>(array: &T, offset: usize, length: usize) -> T {
    let data = array.to_data();

    let use_null_optimization = false;
    let mut data_sliced =
        arrow::array::MutableArrayData::new(vec![&data], use_null_optimization, length);

    data_sliced.extend(0, offset, offset + length);

    T::from(data_sliced.freeze())
}
```
For now, this puts us in a stable state so that we can ship a patch release and breathe a little. 

This routine is exposed via a new `Chunk::row_sliced_deep` method. The old `Chunk::row_sliced` has been renamed to `Chunk::row_sliced_shallow`:
```rust
/// Shallow-slices the [`Chunk`] vertically.
///
/// The result is a new [`Chunk`] with the same columns and (potentially) less rows.
///
/// This cannot fail nor panic: `index` and `len` will be capped so that they cannot
/// run out of bounds.
/// This can result in an empty [`Chunk`] being returned if the slice is completely OOB.
///
/// WARNING: the returned chunk has the same old [`crate::ChunkId`]! Change it with [`Self::with_id`].
///
/// ## When to use shallow vs. deep slicing?
///
/// This operation is shallow and therefore always O(1), which implicitly means that it cannot
/// ever modify the values of the offsets themselves.
/// Since the offsets are left untouched, the original unsliced data must always be kept around
/// too, _even if the sliced data were to be written to disk_.
/// Similarly, the sizes reported might not always make intuitive sense, and should be used
/// very carefully.
///
/// For these reasons, shallow slicing should only be used in the context of short-term, in-memory storage
/// (e.g. when slicing the results of a query).
/// When slicing data for long-term storage, whether in-memory or on disk, see [`Self::row_sliced_deep`] instead.
fn row_sliced_shallow(&self, index: usize, len: usize) -> Self;

/// Deep-slices the [`Chunk`] vertically.
///
/// The result is a new [`Chunk`] with the same columns and (potentially) less rows.
///
/// This cannot fail nor panic: `index` and `len` will be capped so that they cannot
/// run out of bounds.
/// This can result in an empty [`Chunk`] being returned if the slice is completely OOB.
///
/// WARNING: the returned chunk has the same old [`crate::ChunkId`]! Change it with [`Self::with_id`].
///
/// ## When to use shallow vs. deep slicing?
///
/// This operation is deep and therefore always O(N).
///
/// The underlying data, offsets, bitmaps and other buffers required will be reallocated, copied around,
/// and patched as much as required so that the resulting physical data becomes as packed as possible for
/// the desired slice.
/// Similarly, the reported sizes would always match intuitive expectations.
///
/// These characteristics make deep slicing very useful for longer term data, whether it's
/// stored in-memory (e.g. in a `ChunkStore`), or on disk.
/// When slicing data for short-term needs (e.g. slicing the results of a query), whether in-memory or on
/// disk, prefer [`Self::row_sliced_shallow`] instead.
#[must_use]
fn row_sliced_deep(&self, index: usize, len: usize) -> Self
```

## What's next

There are definitely ways to make deep-slicing faster (we should only ever need to copy/reallocate/patch offsets), and I believe `get_slice_memory_size` could also be implemented in a way that works with all possible slicing setups.
In fact, I wonder if deep-slicing could be avoided entirely by hot-patching everything as needed during serialization. Again, hard to no whether there are design limitations preventing that without digging first.

Because these things are tricky, take time to get right and even more time to test properly, this PR instead focuses on providing a provably correct baseline that will always return the correct size and never writes one too many byte on disk. Optimizations can follow later.

---

* https://github.com/rerun-io/rerun/pull/12306 is an example of a PR that implements such optimizations, but [further testing](https://github.com/rerun-io/rerun/pull/12306#issuecomment-3710841239) has shown that its results differ in some cases at the moment. We shall come back to it.
* Fixes #12304
* Fixes #12305

---

```python
import rerun as rr
import numpy as np
import time

rr.init("testing")
rr.save("wtf.rrd")

for i in range(5000):
    rr.set_time("local_ts_us", timestamp=np.datetime64(int(time.time() * 1e6), "us"))
    x = np.random.randint(0, 50000, size=1024)
    rr.log("test/tensors", rr.Tensor(x))
```

Now compacts as expected:
```
pixi run rerun rrd compact wtf.rrd > /dev/null
✨ Pixi task (rerun in default): cargo run --package rerun-cli --no-default-features --features release_no_web_viewer -- rrd compact wtf.rrd
    Finished `dev` profile [optimized] target(s) in 0.54s
     Running `target/debug/rerun rrd compact wtf.rrd`
[2026-01-05T16:35:22Z DEBUG re_analytics::native::pipeline] Analytics disabled in debug builds
[2026-01-05T16:35:22Z DEBUG rerun::commands::entrypoint] Detected 32 cores. Using 30 compute threads.
[2026-01-05T16:35:22Z INFO  rerun::commands::rrd::merge_compact] merge/compaction started max_rows=4 096 max_rows_if_unsorted=1 024 max_bytes=384 KiB srcs=["wtf.rrd"]
[2026-01-05T16:35:22Z INFO  rerun::commands::rrd::merge_compact] processing input…
[2026-01-05T16:35:22Z DEBUG re_sorbet::migrations] Encountered batch without 'sorbet:version' metadata.
[2026-01-05T16:35:22Z DEBUG re_sorbet::migrations] Performing migrations from 0.0.1…
[2026-01-05T16:35:22Z DEBUG re_sorbet::migrations] Migrating record batch from Sorbet 'v0.0.1' to 'v0.0.2'.
[2026-01-05T16:35:22Z DEBUG re_sorbet::migrations] Migrating record batch from Sorbet 'v0.0.1' to 'v0.1.0'.
[2026-01-05T16:35:22Z DEBUG re_sorbet::migrations] Migrating record batch from Sorbet 'v0.0.1' to 'v0.1.1'.
[2026-01-05T16:35:22Z DEBUG re_sorbet::migrations] Migrating record batch from Sorbet 'v0.0.1' to 'v0.1.2'.
[2026-01-05T16:35:22Z INFO  rerun::commands::rrd::merge_compact] running extra compaction pass… pass=0
[2026-01-05T16:35:22Z INFO  rerun::commands::rrd::merge_compact] extra compaction pass completed pass=0 num_chunks_before=122 num_chunks_after=122 num_chunks_reduction="-0.000%" time=798.954µs
[2026-01-05T16:35:22Z INFO  rerun::commands::rrd::merge_compact] cannot possibly improve further, stopping early pass=0 time=808.384µs
[2026-01-05T16:35:22Z INFO  rerun::commands::rrd::merge_compact] preparing output…
[2026-01-05T16:35:22Z INFO  rerun::commands::rrd::merge_compact] encoding…
[2026-01-05T16:35:22Z INFO  rerun::commands::rrd::merge_compact] merge/compaction finished srcs=["wtf.rrd"] time=153.255563ms num_chunks_before=42 num_chunks_after=122 num_chunks_reduction="--190.476%" srcs_size_bytes=20.0 MiB dst_size_bytes=20.2 MiB size_reduction="--0.754%"
```

### DIFF
--- a/crates/store/re_chunk/src/slice.rs
+++ b/crates/store/re_chunk/src/slice.rs
@@ -49,8 +49,8 @@ impl Chunk {
     /// ever modify the values of the offsets themselves.
     /// Since the offsets are left untouched, the original unsliced data must always be kept around
     /// too, _even if the sliced data were to be written to disk_.
-    /// Similarly, the sizes reported might not always make intuitive sense, and should be used
-    /// very carefully.
+    /// Similarly, the byte sizes reported by e.g. `Chunk::heap_size_bytes` might not always make intuitive
+    /// sense, and should be used very carefully.
     ///
     /// For these reasons, shallow slicing should only be used in the context of short-term, in-memory storage
     /// (e.g. when slicing the results of a query).
@@ -78,12 +78,11 @@ impl Chunk {
     /// The underlying data, offsets, bitmaps and other buffers required will be reallocated, copied around,
     /// and patched as much as required so that the resulting physical data becomes as packed as possible for
     /// the desired slice.
-    /// Similarly, the reported sizes would always match intuitive expectations.
+    /// Similarly, the byte sizes reported by e.g. `Chunk::heap_size_bytes` should always match intuitive expectations.
     ///
     /// These characteristics make deep slicing very useful for longer term data, whether it's stored
     /// in-memory (e.g. in a `ChunkStore`), or on disk.
-    /// When slicing data for short-term needs (e.g. slicing the results of a query), whether in-memory or on
-    /// disk, prefer [`Self::row_sliced_shallow`] instead.
+    /// When slicing data for short-term needs (e.g. slicing the results of a query) prefer [`Self::row_sliced_shallow`] instead.
     #[must_use]
     pub fn row_sliced_deep(&self, index: usize, len: usize) -> Self {
         let deep = true;
--- a/crates/store/re_chunk/src/split.rs
+++ b/crates/store/re_chunk/src/split.rs
@@ -28,6 +28,9 @@ impl Chunk {
     /// Naively splits a chunk if it exceeds the configured thresholds.
     ///
     /// The resulting pieces may still be larger than [`ChunkSplitConfig::chunk_max_bytes`].
+    ///
+    /// The Chunk is *deeply* sliced, as opposed to shallowly. Refer to [`Chunk::row_sliced_deep`]
+    /// to learn more about that and why it matters.
     pub fn split_chunk_if_needed(chunk: Arc<Self>, cfg: &ChunkSplitConfig) -> Vec<Arc<Self>> {
         let ChunkSplitConfig {
             chunk_max_bytes,


crates/utils/re_arrow_util/src/snapshots/re_arrow_util__arrays__tests__deep_slice_comparisons.snap
teh-cmc, line 1: Screenshot for small screens:
image

crates/utils/re_arrow_util/src/snapshots/re_arrow_util__arrays__tests__deep_slice_comparisons.snap
```diff
@@ -0,0 +1,48 @@
+---
+source: crates/utils/re_arrow_util/src/arrays.rs
+expression: output
+---
+int32:
+array[0..]:          len=    100000 array_size=    400096 buf_size=    400000 data.len=    100000 data.array_size=    400160 data.buf_size=    400000 data.slice_size=    400000 / IPC=412872
+slice[25000..75000]: len=     50000 array_size=    400096 buf_size=    400000 data.len=     50000 data.array_size=    400160 data.buf_size=    400000 data.slice_size=    200000 / IPC=206600
+ deep[25000..75000]: len=     50000 array_size=    200096 buf_size=    200000 data.len=     50000 data.array_size=    200160 data.buf_size=    200000 data.slice_size=    200000 / IPC=206600
```
teh-cmc, line 8: Primitive arrays seem to be doing fine.

crates/utils/re_arrow_util/src/snapshots/re_arrow_util__arrays__tests__deep_slice_comparisons.snap
```diff
@@ -0,0 +1,48 @@
...
+
+list[int32]:
+array[0..]:          len=    100000 array_size=    800212 buf_size=    800004 data.len=    100000 data.array_size=    800324 data.buf_size=    800004 data.slice_size=    800000 / IPC=825544
+slice[25000..75000]: len=     50000 array_size=    800212 buf_size=    800004 data.len=     50000 data.array_size=    800324 data.buf_size=    800004 data.slice_size=    600000 / IPC=413000
+ deep[25000..75000]: len=     50000 array_size=    400272 buf_size=    400064 data.len=     50000 data.array_size=    400384 data.buf_size=    400064 data.slice_size=    400000 / IPC=413000
```
teh-cmc, line 13: List arrays already show invalid data:
* Vanilla `slice()` reports 50k entries in the slice, for a total slice size of 600kb: that doesn't make sense. The IPC size is correct though, so the writer actually sliced the data as expected in the end.

teh-cmc: I'm not testing it here, but it's fair to assume that if Lists fail, then everything that works similarly do too (BinaryArray, StringArray, etc).

crates/utils/re_arrow_util/src/snapshots/re_arrow_util__arrays__tests__deep_slice_comparisons.snap
```diff
@@ -0,0 +1,48 @@
...
+
+list[list[int32]]:
+array[0..]:          len=    100000 array_size=   1200328 buf_size=   1200008 data.len=    100000 data.array_size=   1200488 data.buf_size=   1200008 data.slice_size=   1200000 / IPC=1238280
+slice[ 5000.. 5005]: len=         5 array_size=   1200328 buf_size=   1200008 data.len=         5 data.array_size=   1200488 data.buf_size=   1200008 data.slice_size=    800020 / IPC=   904
+ deep[ 5000.. 5005]: len=         5 array_size=       512 buf_size=       192 data.len=         5 data.array_size=       672 data.buf_size=       192 data.slice_size=        60 / IPC=   904
```
teh-cmc, line 18: A clear demonstration of why the splitter goes insane: vanilla size reports 800kb in-memory size, which actually turns to 900 bytes on disk.

This is bug number 1: it causes the splitter to split much more than it should.

crates/utils/re_arrow_util/src/snapshots/re_arrow_util__arrays__tests__deep_slice_comparisons.snap
```diff
@@ -0,0 +1,48 @@
...
+
+struct{int32}:
+array[0..]:          len=    100000 array_size=    400200 buf_size=    400000 data.len=    100000 data.array_size=    400296 data.buf_size=    400000 data.slice_size=    400000 / IPC=425480
+slice[25000..75000]: len=     50000 array_size=    400200 buf_size=    400000 data.len=     50000 data.array_size=    400296 data.buf_size=    400000 data.slice_size=    200000 / IPC=212936
+ deep[25000..75000]: len=     50000 array_size=    200200 buf_size=    200000 data.len=     50000 data.array_size=    200296 data.buf_size=    200000 data.slice_size=    200000 / IPC=212936
```
teh-cmc, line 23: Standard struct arrays seem fine.

crates/utils/re_arrow_util/src/snapshots/re_arrow_util__arrays__tests__deep_slice_comparisons.snap
```diff
@@ -0,0 +1,48 @@
...
+list[struct{int32}]:
+array[0..]:          len=    100000 array_size=    800316 buf_size=    800004 data.len=    100000 data.array_size=    800460 data.buf_size=    800004 data.slice_size=    800000 / IPC=838216
+slice[25000..75000]: len=     50000 array_size=    800316 buf_size=    800004 data.len=     50000 data.array_size=    800460 data.buf_size=    800004 data.slice_size=    600000 / IPC=419400
+ deep[25000..75000]: len=     50000 array_size=    400376 buf_size=    400064 data.len=     50000 data.array_size=    400520 data.buf_size=    400064 data.slice_size=    400000 / IPC=419400
+
+union#dense{{u8,f32,i64}}::
+array[0..]:          len=    100000 array_size=   1000392 buf_size=   1000008 data.len=    100000 data.array_size=   1000672 data.buf_size=   1000008 data.slice_size=   1000008 / IPC=1021832
+slice[25000..75000]: len=     50000 array_size=   1000392 buf_size=   1000008 data.len=     50000 data.array_size=   1000672 data.buf_size=   1000008 data.slice_size=    750008 / IPC=771848
+ deep[25000..75000]: len=     50000 array_size=    900480 buf_size=    900096 data.len=     50000 data.array_size=    900760 data.buf_size=    900096 data.slice_size=    466670 / IPC=473864
```
teh-cmc, line 33: Now this is _very_ messed up: not only the in-memory size is wrong here, but so is the data actually written to disk. That's the reason why the RRD file sizes were balooning.

This is bug number 2: data that doesn't belong to the slice not only reports its memory size wrongly, but also makes it all the way to the disk.

crates/store/re_chunk/src/slice.rs
```diff
@@ -78,7 +78,7 @@ impl Chunk {
             entity_path: entity_path.clone(),
             heap_size_bytes: Default::default(),
             is_sorted,
-            row_ids: row_ids.clone().slice(index, len),
+            row_ids: re_arrow_util::deep_slice_array(row_ids, index, len),
```
emilk: We should add a docstring to the parent function (`fn row_sliced`) to explain that it is NOT a shallow slice, as one would expect. And probably explain WHY. And maybe add a TODO to make it shallow again in the future, if that would make sense

crates/utils/re_byte_size/src/arrow_sizes.rs
```diff
@@ -136,58 +136,11 @@ impl SizeBytes for UnionFields {
 
 // ---
 
-/// Returns the accurate memory size of an Arrow array, accounting for slicing.
-///
-/// For `ListArray`s, this manually calculates the size by only counting the memory used by
-/// the list entries that are actually in the slice, since Arrow's `get_slice_memory_size()`
-/// doesn't properly handle the case where sliced `ListArray`s still reference large inner data.
+/// Returns the memory size of an Arrow array.
 fn array_slice_memory_size(array: &dyn Array) -> u64 {
```
emilk: Is `array_slice_memory_size(a)` and `array_slice_memory_size(a.sliced(0, 0))` expected to always be the same, or different? Please add to docstring, and maybe add a unit test!

This seems to be a continuing point of confusion

teh-cmc:
I'll make _both_ shallow & deep slicing first-class citizens later this morning, with docs explaining why both are needed.

teh-cmc:
@rerun-bot full-check

crates/store/re_chunk/src/slice.rs
```diff
@@ -42,10 +42,57 @@ impl Chunk {
     /// This can result in an empty [`Chunk`] being returned if the slice is completely OOB.
     ///
     /// WARNING: the returned chunk has the same old [`crate::ChunkId`]! Change it with [`Self::with_id`].
+    ///
+    /// ## When to use shallow vs. deep slicing?
+    ///
+    /// This operation is shallow and therefore always O(1), which implicitly means that it cannot
+    /// ever modify the values of the offsets themselves.
```
emilk, line 49:
```suggestion
    /// ever modify the values of the ListArray offsets themselves.
```

crates/store/re_chunk/src/slice.rs
```diff
@@ -42,10 +42,57 @@ impl Chunk {
...
+    /// Since the offsets are left untouched, the original unsliced data must always be kept around
+    /// too, _even if the sliced data were to be written to disk_.
```
emilk, line 51: So the IPC encoder will write the full original buffer? I sort of wish the IPC encoder instead has a flag for re-computing the offsets for slices, or that we had a `record_batch.shrink_to_fit()` that did this 🤔

crates/store/re_chunk/src/slice.rs
```diff
@@ -42,10 +42,57 @@ impl Chunk {
...
+    /// Similarly, the sizes reported might not always make intuitive sense, and should be used
+    /// very carefully.
```
emilk:
```suggestion
    /// Similarly, the byte sizes reported by [`Chunk::byte_size`] might not always make intuitive sense, and should be used
    /// very carefully.
```

something like this?

crates/store/re_chunk/src/split.rs
```diff
@@ -78,7 +78,7 @@ impl Chunk {
 
             // TODO(#11971): keep track of the split chunks' lineage
             let split_chunk = chunk
-                .row_sliced(start_idx, chunk_size)
+                .row_sliced_deep(start_idx, chunk_size)
```
emilk, line 84: We should maybe docstring that the splitter is deep?

emilk (approved):
Awesome - thanks for digging into this 🙏

crates/store/re_chunk/src/slice.rs
teh-cmc: It's not specific to ListArrays, almost everything besides primitive arrays rely on offsets.

crates/store/re_chunk/src/slice.rs
teh-cmc: > So the IPC encoder will write the full original buffer?

Yes

>  I sort of wish the IPC encoder instead has a flag for re-computing the offsets for slices

It does perform that kind of just-in-time packing automatically in _some_ cases, as I've demonstrated in the PR description, but **not** all. Which made the whole thing even more confusing to debug...

> or that we had a `record_batch.shrink_to_fit()` that did this

Yeah I think `record_batch.shrink_to_fit` is a nice way to think about this. Effectively, this is what `deep_slice_array()` is today: it's an overly aggressive `shrink_to_fit`.

I believe the ideal world for us would be one where A) `get_memory_slice_size` would work in all scenarios (which I see I no reason why it couldn't) and B) the IPC serializer would always perform aggressive JIT shrink-to-fit/packing at serialization time.
This would make it so that things "just work" during normal operations, without ever having to think about shallow vs. deep semantics at all, and I'm pretty confident that the cost of JIT packing would be vastly dwarfed by the IO that follows it in any case.

Issue #12313, opened 2026-01-05
`.serve_grpc()` should be possible to use independently of recording stream


jleibs:
In the Python SDK, the only way to spawn an internal grpc_server is using the `.serve_grpc()` APIs.

The grpc server is stored in a LogSink and is automatically cleaned up if the corresponding recording goes out of scope or is disconnected.

This leads to confusing situations like the following:
```python
import rerun as rr

rr.init("rerun_example_grpc_server")

# Create a grpc server
uri = rr.serve_grpc()

# A user might want to connect to it with a dedicated stream
rec1 = rr.RecordingStream("rerun_example_grpc_server")
rec1.connect_grpc(uri)
rec1.log("object_stream", rr.TextLog("This works"))

# Later they set up the global recording to go to file isntead
rr.save("rerun_example_grpc_server.rerun")

# Doing the same thing, now fail because the grpc server was implicitly shut down
# when the global recording was switched to a file sin
rec2 = rr.RecordingStream("rerun_example_grpc_server")
rec2.connect_grpc(uri)
rec2.log("object_stream", rr.TextLog("This fail"))
```

It would be nice if a dedicated server object was created, separate of the stream concepts:
```python
import rerun as rr

# Create a grpc server. As long as the server object exists
server = rr.GrpcServer()

# A user can then connect to the server with a stream
rec1 = rr.RecordingStream("rerun_example_grpc_server")
rec1.connect_grpc(server)
rec1.log("object_stream", rr.TextLog("This works"))

# Changes to the global stream have no implicit impact on the server object
rr.save("rerun_example_grpc_server.rerun")

# A user can still connect and log to the server
rec2 = rr.RecordingStream("rerun_example_grpc_server")
rec2.connect_grpc(server)
rec2.log("object_stream", rr.TextLog("This still works"))
```

Issue #12301, opened 2025-12-25
RecordingStream disconnects when variable goes out of scope


JafarAbdi:
Hi folks, 

Really appreciate the package, I use it everyday and it makes my life so much easier.

I ran into a bug after upgrading to 0.28. Basically I get the following in my terminal. I think it's related to the `RecordingStream`s getting garbage collected when they go out of scope, causing gRPC disconnections.

```bash
[2025-12-25T19:59:50Z ERROR re_sdk::recording_stream] Failed to flush sink: gRPC has been unable to connect to rerun+http://127.0.0.1:9876/proxy for 6s
[2025-12-25T19:59:51Z ERROR re_sdk::recording_stream] Failed to flush previous sink: gRPC has been unable to connect to rerun+http://127.0.0.1:9876/proxy for 7s
[2025-12-25T19:59:52Z ERROR re_grpc_client::write] Failed to flush gRPC messages during shutdown: gRPC has been unable to connect to rerun+http://127.0.0.1:9876/proxy for 8s
```

And the following from the web

<img width="351" height="109" alt="Image" src="https://github.com/user-attachments/assets/12e6447f-ba24-4413-b110-ef524881f336" />

**To Reproduce**

1. Clone https://github.com/JafarAbdi/rerun_bug
2. Run `uv run bug_repro.py`
3. Bug
4. If you run `uv run bug_repro_init.py` or `uv run bug_repro_fixed.py` they work as expected

**Expected behavior**:

Recordings should remain connected until explicitly closed, not when Python GC collects the variable.

**Desktop**:
- OS: Ubuntu 24.04.3 LTS

**Rerun version**:
- rerun-sdk<0.28 - works
- rerun-sdk>=0.28 - broken

**Potential fix?**:
<details>

I tried running claude code to see if it can help fixing the issue. Here is the patch, but since I have no experience with the codebase, I don't trust it :D:D

```diff
diff --git a/crates/store/re_grpc_server/src/shutdown.rs b/crates/store/re_grpc_server/src/shutdown.rs
index 743c8c7864..13af92f706 100644
--- a/crates/store/re_grpc_server/src/shutdown.rs
+++ b/crates/store/re_grpc_server/src/shutdown.rs
@@ -26,12 +26,20 @@ impl Signal {
 pub struct Shutdown(Option<oneshot::Receiver<()>>);
 
 impl Shutdown {
-    /// Returns a future that resolves when the signal is sent.
[...]

PR #12314, opened 2026-01-05
Improve the docstrings related to serve_gpc APIs

jleibs:
When we introduced https://github.com/rerun-io/rerun/pull/12220 the recording-stream cleanup logic had unexpected side-effects for some users of grpc_server sink. See: https://github.com/rerun-io/rerun/issues/12301

Although this change in behavior is subtle, it's correct given the current architecture of the RecordingStream and Sink ownership model.

I've created a future issue for making it possible to avoid this by changing that ownership model rather than introducing new edge-cases in the cleanup code-paths: https://github.com/rerun-io/rerun/issues/12313

### DIFF
--- a/docs/content/reference/migration/migration-0-28.md
+++ b/docs/content/reference/migration/migration-0-28.md
@@ -366,3 +366,39 @@ The `send_dataframe()` and `send_record_batch()` functions have been moved to th
 |-----------------------------------------|-----------------------------------------|
 | `rr.dataframe.send_dataframe()`         | `rr.send_dataframe()`                   |
 | `rr.dataframe.send_record_batch()`      | `rr.send_record_batch()`                |
+
+## `RecordingStream` now cleans up when going out of scope in Python SDK
+
+`RecordingStream` objects are now cleaned up as they go out of scope. This specifically
+includes flushing and closing the associated sinks.
+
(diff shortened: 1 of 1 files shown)


docs/content/reference/migration/migration-0-28.md
```diff
@@ -366,3 +366,38 @@ The `send_dataframe()` and `send_record_batch()` functions have been moved to th
...
+    create_recording(data1, "data1.rrd")
+    create_recording(data2, "data2.rrd")
+    create_recording(data3, "data3.rrd")
+```
+
+**Before**
+All 3 recording files would stay open and unterminated until the application exit.
+
+**After**
+Now each recording is closed and terminated incrementally as the functions run and the
```
ntjohnson1: Not finishing a sentence talking about improper termination is pretty funny.

jleibs: Gah. I guess I need to actually hit save in my editor before I commit and push.

ntjohnson1 (approved):
Confirmed there is no race condition and all that is needed is holding onto the recording that is connected to the server. I think the documentation for 0.28 is fine but we should explore cleaning this up soon (thanks for already filing the follow up issue)

PR #12315, opened 2026-01-05
Remove a bunch of `fn as_any`

emilk:
No longer needed with recent Rust versions!


timsaucer:
We have a few build errors, but in general this looks like a very nice cleanup.

PR #12317, opened 2026-01-05
Put dataplatform examples in standard docs

ntjohnson1:
### Related
RR-3088

Most of this is just moving the files from our private repo where we initially debugged these. https://github.com/rerun-io/dataplatform-examples

### What
* Takes our standard set of dataplatform examples and runs them as snippets using the open source server
   * Checks in small datasets to compare with
* Don't yell about snippets generating empty RRDs. Our snippet comparison is super painful. Will file an issue to clean this up later. For now, add rust/cpp to don't run and add all languages to don't compare, and also add files to backwards check to run but don't verify against rrd 😢

### DIFF
--- a/docs/content/howto/query-and-transform/dataframe_operations.md
+++ b/docs/content/howto/query-and-transform/dataframe_operations.md
@@ -4,7 +4,7 @@ order: 50
 ---
 
 Dataframes are core to modern analytics workflows.
-Rerun provides a dataframe interface to your data via [datafusion](https://datafusion.apache.org/python/).
+Rerun provides a dataframe interface to your data via [DataFusion](https://datafusion.apache.org/python/).
 This example performs a series of joins, filters, etc that highlight a variety of common operations in context.
 Because datafusion has a lazy execution model it is generally more performant to use datafusion for processing,
 however datafusion does allow conversion to dataframes for popular tools (pandas, polars, pyarrow).
--- a/docs/content/howto/query-and-transform/overview.md
+++ b/docs/content/howto/query-and-transform/overview.md
@@ -9,4 +9,4 @@ The open source SDK connects to the cloud Data Platform, which allows you to sto
 The open source Rerun SDK includes a simplified Data Platform server, which is API compatible with the cloud platform.
 The open source server loads everything into memory, which makes it fast and simple to operate for very small datasets, which in turn makes it perfect for quick testing and local experimentation.
 
-See the [how to guide for the open source server](get-data-out.md) for more details on launching and connecting to the server.
+See the [how-to guide for the open source server](get-data-out.md) for more details on launching and connecting to the server.
--- a/docs/snippets/all/howto/query_images.py
+++ b/docs/snippets/all/howto/query_images.py
@@ -58,7 +58,7 @@ nearest_keyframe = nearest_keyframe.select(
     .over(Window(window_frame=WindowFrame("rows", None, 0), order_by="log_time"))
     .alias("log_time")
 )
-keyframe_ts = pa.table(nearest_keyframe)["log_time"].to_numpy()[0]
+keyframe_ts = pa.table(nearest_keyframe.select("log_time"))["log_time"].to_numpy()[0]
 
 df = video_df.filter((col("log_time") <= selected_time) | (col("log_time") >= keyframe_ts))
 rows = df.select("log_time", video_column)
--- a/docs/snippets/all/howto/time_alignment.py
+++ b/docs/snippets/all/howto/time_alignment.py
@@ -21,7 +21,7 @@ cheaper_column = (
     .filter_contents("/observation/joint_positions")
     .reader(index="real_time")
 )
-
+# TODO(RR-3189): Fix when we have a cleaner way to extract this time range efficiently
 min_max = cheaper_column.aggregate(
     "rerun_segment_id", [F.min(col("real_time")).alias("min"), F.max(col("real_time")).alias("max")]
 )


ntjohnson1:
Moving this back to draft because we are exploring shuffling our docs a bit. Right now this added `cloud` which was a bad name. But no reason to figure out a better one if we are about to reorganize

ntjohnson1:
Ugh preview doesn't load. Missing a link. Back to do not merge but otherwise this should be ok

ntjohnson1:
Moved this back to do not merge since the shuffle got delayed but is soon

docs/content/howto/query-and-transform/dataframe_operations.md
```diff
@@ -0,0 +1,44 @@
+---
+title: Common Dataframe Operations
+order: 50
+---
+
+Dataframes are core to modern analytics workflows.
+Rerun provides a dataframe interface to your data via [datafusion](https://datafusion.apache.org/python/).
```
abey79: I think the project like this casing better:

```suggestion
Rerun provides a dataframe interface to your data via [DataFusion](https://datafusion.apache.org/python/).
```

docs/content/howto/query-and-transform/overview.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Overview
+order: 10
+---
+
+Rerun is a Data Platform for Physical Data.
+The open source SDK connects to the cloud Data Platform, which allows you to store, retrieve, and query over large amounts of data, and integrates with the SDK so you can browse and inspect the data visually.
+
+The open source Rerun SDK includes a simplified Data Platform server, which is API compatible with the cloud platform.
+The open source server loads everything into memory, which makes it fast and simple to operate for very small datasets, which in turn makes it perfect for quick testing and local experimentation.
```
abey79, line 10: note: hoping to cover this in the future version of app-model.md

docs/content/howto/query-and-transform/overview.md
```diff
@@ -0,0 +1,12 @@
...
+
+See the [how to guide for the open source server](get-data-out.md) for more details on launching and connecting to the server.
```
abey79:
```suggestion
See the [how-to guide for the open source server](get-data-out.md) for more details on launching and connecting to the server.
```

docs/snippets/all/howto/dataframe_operations.py
```diff
@@ -0,0 +1,173 @@
...
+import numpy as np
+import pyarrow as pa
+import rerun as rr
+from datafusion import col, functions as F, lit
+
+sample_5_path = Path(__file__).parents[4] / "tests" / "assets" / "rrd" / "sample_5"
+
+server = rr.server.Server(datasets={"sample_dataset": sample_5_path})
+CATALOG_URL = server.address()
+client = rr.catalog.CatalogClient(CATALOG_URL)
```
abey79, line 16: an reason you are not using `client = server.client()`?

docs/snippets/all/howto/dataframe_operations.py
```diff
@@ -0,0 +1,173 @@
...
+dataset = client.get_dataset(name="sample_dataset")
+observations = dataset.filter_contents(["/observation/**"]).reader(index="real_time")
```
abey79, line 18: This line deserves a comment since there are no explanation around  it in the doc page.


```suggestion
# Obtain the observations as `datafusion.DataFrame`. This is cheap because DataFusion's dataframes are lazy query plans, aka no data is downloaded or processed yet.
observations = dataset.filter_contents(["/observation/**"]).reader(index="real_time")
```

docs/snippets/all/howto/dataframe_operations.py
```diff
@@ -0,0 +1,173 @@
...
+    col("rerun_segment_id"),
+    [
+        F.first_value(col("real_time")).alias("start"),
+        F.last_value(col("real_time")).alias("end"),
+    ],
+)
+
+# Sort for consistency here
+first_last = first_last.sort("start")
+pa.table(first_last)["start"][0]
```
abey79, line 32: Likewise, I'd add a note here to signal that here we are materialising the dataframe into a pa.table().

Also, isn't that suboptimal by materialising the full dataframe and then only keeping the `"start"` column. Maybe use `project` on the dfn.dataframe (or whatever it is)

Finally, you might want to explicitly call `print` here since we're not in a notebook.

docs/content/howto/query-and-transform/dataframe_operations.md
```diff
@@ -0,0 +1,44 @@
...
+## Group-by / aggregation
+
+Perform an aggregation on the episodes to track the first and last timestamp for the columns of interest.
+
+snippet: howto/dataframe_operations[group_by]
+
+## Join and query
+
+Some of our columns start much later than others.
+Find out how often this delay exceeds some threshold.
```
abey79, line 29: I'm having a hard parsing these without the context of what we're trying to do here.

docs/content/howto/query-and-transform/dataframe_operations.md
```diff
@@ -0,0 +1,44 @@
...
+
+⚠️ **Performance warning:**
+Even though datafusion pulls data lazily, we don't currently decouple our payload from its timeline.
+E.g. in this example this means that we have to pull the full camera data to inspect their min/max timestamps.
+This works quickly when the data is already local and in memory, but can be a bottleneck on cloud at scale.
```
abey79, line 34: I'd feel better if we worded this as a current limitation rather than what sounds like a fundamental shortcoming.

docs/content/howto/query-and-transform/dataframe_operations.md
```diff
@@ -0,0 +1,44 @@
+---
+title: Common Dataframe Operations
```
abey79, line 2: Based on this title, I was expecting some kind of collection of generally useful dataframe-related operations. Instead, it reads more like a fully fledged example of some processing (that does highlight a number of operation that have general usefulness). The framing tripped me up here.

My hopefully actionable suggestion would be to frame this as "how to use dataframes for common query tasks" (or something to that effect), and then clear up a couple of things before getting into the material.
- by "dataframe", we mean datafusion's dataframe, which are not, in fact, dataframe (but rather query plans) ([rant]very unfortunate naming from datafusion[/rant])
- this page uses a concrete example to illustrate the use of dataframes
- short description of what that task is about

docs/content/howto/query-and-transform/query_images.md
```diff
@@ -0,0 +1,38 @@
+---
+title: Query image & video types
+order: 60
+---
+
+Images and videos are incredibly useful, however there are many ways to store and manipulate them.
+For more details about the different video types we support see our [video reference](https://rerun.io/docs/reference/video).
+This example focuses on querying image frames from the Rerun dataplatform.
```
abey79, line 8: Add a note to say that Rerun supports various way to store streams of images/videos, and that this page covers them all.

docs/snippets/all/howto/query_images.py
```diff
@@ -0,0 +1,91 @@
...
+from PIL import Image
+
+sample_video_path = Path(__file__).parents[4] / "tests" / "assets" / "rrd" / "video_sample"
+
+server = rr.server.Server(datasets={"video_dataset": sample_video_path})
+CATALOG_URL = server.address()
+client = rr.catalog.CatalogClient(CATALOG_URL)
+dataset = client.get_dataset(name="video_dataset")
+df = dataset.filter_contents(["/compressed_images/**", "/raw_images/**", "/video_stream/**"]).reader(index="log_time")
+times = pa.table(df.select("log_time"))["log_time"].to_numpy()
```
abey79, line 23: ditto

docs/snippets/all/howto/query_images.py
```diff
@@ -0,0 +1,91 @@
...
+# Make sure keyframe is in the past
+nearest_keyframe = video_df.filter((col("log_time") <= selected_time) & col(keyframe_column).is_not_null())
+
+# Get the most recent keyframe timestamp
+nearest_keyframe = nearest_keyframe.select(
+    F.last_value(col("log_time"))
+    .over(Window(window_frame=WindowFrame("rows", None, 0), order_by="log_time"))
+    .alias("log_time")
+)
+keyframe_ts = pa.table(nearest_keyframe)["log_time"].to_numpy()[0]
```
abey79: same
```suggestion
keyframe_ts = pa.table(nearest_keyframe.select("log_time"))["log_time"].to_numpy()[0]
```

docs/content/howto/query-and-transform/view_operations.md
```diff
@@ -0,0 +1,33 @@
+---
+title: View Operations
+order: 70
+---
+
+Robotics data has many sensors and many columns.
+In order to more narrowly specify relevant content for further dataframe operations you first generate a view.
+This view can filter on episode, time, column name etc.
+This example shows specific instances highlighting these capabilities.
+
```
abey79, line 10: Would be nice to have a sentence to help solidify what we mean by "view" exactly. Aka the fact that they apply to dataset before they are "reshaped" into a dataframe-looking thing by the `reader()` operation.

docs/snippets/all/howto/time_alignment.py
```diff
@@ -0,0 +1,49 @@
...
+# region: extract_timepoints
+cheaper_column = (
+    dataset.filter_segments("ILIAD_sbd7d2c6_2023_12_24_16h_20m_37s")
+    .filter_contents("/observation/joint_positions")
+    .reader(index="real_time")
+)
+
+min_max = cheaper_column.aggregate(
+    "rerun_segment_id", [F.min(col("real_time")).alias("min"), F.max(col("real_time")).alias("max")]
+)
```
abey79, line 27: we absolutely need to fix this when Tim lands `get_index_ranges()`. Maybe a `TODO(RR-3189)`?

abey79: That would make this snippet no-longer usable for prod until 0.29 is out, though :/

docs/snippets/all/howto/dataframe_operations.py
ntjohnson1: Yes because this approach is the more general one. For instance if you are running a long running local server or connecting to the cloud the diff is smaller.

docs/content/howto/query-and-transform/dataframe_operations.md
ntjohnson1: Hmm I tried to split out different things to do with dataframes (joins, aggregations, etc). But without context it feels meaningless? I think Gijs has some follow ups on all this stuff that he is waiting until this lands to add.

docs/snippets/all/howto/dataframe_operations.py
ntjohnson1: Hmm I'd rather NOT have the explanation in the how to. We definitely need to add the explanation but that feels like it fits a different section

docs/snippets/all/howto/query_images.py
ntjohnson1: Ugh I can't wait for the next python datafusion release where we have
nearest_keyframe.select_column("log_time") to replace this big mess.

docs/snippets/all/howto/time_alignment.py
```diff
@@ -0,0 +1,49 @@
...
+dataset = client.get_dataset(name="sample_dataset")
+# endregion: setup
+
+# region: extract_timepoints
+cheaper_column = (
+    dataset.filter_segments("ILIAD_sbd7d2c6_2023_12_24_16h_20m_37s")
+    .filter_contents("/observation/joint_positions")
+    .reader(index="real_time")
+)
+
```
ntjohnson1:
```suggestion
# TODO(RR-3189): Fix when we have a cleaner way to extract this time range efficiently
```

docs/snippets/all/howto/dataframe_operations.py
ntjohnson1: I think that makes sense NOT in the how to. Materializing it isn't necessarily suboptimal depending on what we do with it. For things that are small just grabbing the column at the end of filtering isn't that bad.

ntjohnson1:
Annoying fun fact: If you mark a conversation as resolved it will no longer let you apply suggestions from the gui

PR #12323, opened 2026-01-06
`/GetRrdManifest` on Rerun Cloud

teh-cmc:
This implements all the necessary OSS changes so that `/GetRrdManifest` on Rerun Cloud can become a reality. The sibling PR takes care of the actual endpoint implementation in Rerun Cloud.

Specifically:
* Make `/GetRrdManifest` a proper end-to-end stream.
* Tweak the test suite so that it doesn't track any implementation-specific concerns.

The protobuf breaking changes are on purpose.

---

* Part of https://linear.app/rerun/issue/RR-3110/implement-get-rrd-manifest-on-rerun-cloud
* Sibling: https://github.com/rerun-io/dataplatform/pull/2206
* DNM: should be merged at the same time as the sibling


crates/store/re_redap_client/src/connection_client.rs
```diff
@@ -382,14 +387,44 @@ where
...
+                    ApiError::serialization(err, "missing field in /GetRrdManifest response")
+                })?
+                .to_application(())
+                .map_err(|err| {
+                    ApiError::serialization(err, "failed parsing /GetRrdManifest response")
+                })?;
+
+            if let Some(mut temp) = rrd_manifest.take() {
+                temp.data =
+                    re_arrow_util::concat_polymorphic_batches(&[temp.data, rrd_manifest_part.data])
```
emilk: ⚠️ This will call concat N-1 times, probably leading to O(N^2) behavior. Better to collect everything in a vector and then call concat only once

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -972,10 +973,12 @@ impl RerunCloudService for RerunCloudHandler {
         }))
     }
 
+    type GetRrdManifestStream = GetRrdManifestResponseStream;
+
     async fn get_rrd_manifest(
         &self,
         request: tonic::Request<re_protos::cloud::v1alpha1::GetRrdManifestRequest>,
-    ) -> tonic::Result<tonic::Response<re_protos::cloud::v1alpha1::GetRrdManifestResponse>> {
+    ) -> std::result::Result<tonic::Response<Self::GetRrdManifestStream>, tonic::Status> {
```
emilk: shorter 🤷 

```suggestion
    ) -> tonic::Result<tonic::Response<Self::GetRrdManifestStream>> {
```

PR #12326, opened 2026-01-06
Hide the 'Server tables' example by default

emilk:
Added by @timsaucer but it should not be on https://rerun.io/viewer


emilk:
* Dup: https://github.com/rerun-io/rerun/pull/12325

PR #12324, opened 2026-01-06
[PROPOSAL] Reorganize documentation

abey79:
### Related

RR-3246

### What

Proposal:
- provide more visibility to the querying and data pipeline parts
- "getting started": both a tutorial, and an overview of the main parts of the Rerun product, so organised along these boundaries:
  - "send data" <-> logging sdk
  - "see data" <-> viewer
  - query data" <-> data sdk
  - data pipeline <-> Rerun Dataplatform
- "concept": focuses on the basics concepts, which are transversal to the product part boundaries (e.g. data model applies to all bits of the product)
  - my proposal is to NOT have a "concept" section, and instead promote its content to the top level (`landing` doesn't seem to let me do that right now)
- renamed `How to` -> `Cookbook` to better reflect the fact that it's a task-oriented grab-bag of recipes (it should be easy to add more without caring too much on the structure—ideally this becomes a repository of tagged/searchable recipes in the future)
- slight reorganisation of the "install rerun"
- moved some stuff from `Reference` to more suitable places

PR #12327, opened 2026-01-06
Shuffle docs example

ntjohnson1:
Investigating: RR-3246
This is another consideration to https://github.com/rerun-io/rerun/pull/12324

Definitely need content changes but demonstrating what this layout looks like.
My strategy/approach is somewhat around laziness and pragmatism. We already have this concepts/howto/reference structure. So this mostly just adds one layer of nesting under that to make it a little clearer where to look.

The names of the sub-folders didn't get too much thought so can massage that further.


ntjohnson1:
closed in favor of https://github.com/rerun-io/rerun/pull/12355

PR #12328, opened 2026-01-06
Support latest-at and range queries in OSS server

timsaucer:
### What

This PR adds support for latest-at and range queries in the OSS server. Additional updates include

- Update ChunkStore to return static component chunks when `include_static` is enabled
- Update redap tests since sending a static + temporal component on an entity should only return the static component
- Add work around for data platform in unit tests when it returns additional chunks for the case of static + temporal component on the same entity

### DIFF
--- a/crates/store/re_server/src/rerun_cloud.rs
+++ b/crates/store/re_server/src/rerun_cloud.rs
@@ -12,7 +12,9 @@ use datafusion::logical_expr::dml::InsertOp;
 use datafusion::prelude::SessionContext;
 use nohash_hasher::IntSet;
 use re_arrow_util::RecordBatchExt as _;
-use re_chunk_store::{Chunk, ChunkStore, ChunkStoreHandle, LatestAtQuery, RangeQuery};
+use re_chunk_store::{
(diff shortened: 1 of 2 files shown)


timsaucer:
@rerun-bot full-check

crates/store/re_server/src/rerun_cloud.rs
```diff
@@ -1151,7 +1151,104 @@ impl RerunCloudService for RerunCloudHandler {
...
+                            .map(Clone::clone)
+                            .collect(),
+                    }
+                } else {
+                    store_handle
+                        .read()
+                        .iter_chunks()
+                        .map(Clone::clone)
+                        .collect()
+                };
```
abey79, line 1172: this makes sense to me, but I feel this bit of code is made unnecessarily hard to read because of the duplication and the "right drift".

timsaucer:
@rerun-bot reality-sync

timsaucer:
@rerun-bot full-check

timsaucer:
@rerun-bot full-check

PR #12329, opened 2026-01-06
Make it possible to build `rerun_sdk` through `uv` without invalidating `pyo3-config`

jleibs:
### What
- This introduces a new `.cargo/config.toml` dependency on `rerun_py/pyo3-build.cfg`

The pixi environment is set up to generate this file during environment activation, which keeps everything generally sane and means we always build our crates using a consistent pyo3 configuration based on the uv environment.

This now makes it so `uv sync` can actually work with `rerun-sdk` package without major cargo invalidations.


timsaucer:
Tested on macos and it worked as advertised. I did have to disable my system's env variable `PYO3_CONFIG_FILE` to test.

timsaucer:
I'm not sure I agree it's worth the effort to do this over suggesting users set up an environment variable to a global config file, but I also don't see any blockers in the implementation. We probably need someone with windows to test since we are at least getting CI errors for those runners.

.cargo/config.toml
```diff
@@ -25,6 +25,18 @@ IS_IN_RERUN_WORKSPACE = "yes"
...
+# executable to .venv/bin/python, ensuring consistent builds regardless of whether
+# uv uses an isolated build environment.
+# The `relative = true` makes cargo expand this to an absolute path.
+#
+# NOTE: This file must be generated before any cargo build involving pyo3.
+# Run `pixi run py-build-pyo3-cfg`
+#
+# If you need to build outside of pixi, you may run `python scripts/generate_pyo3_config.py`
+# manually. But, mixing this with any pixi invocations will result in spurious rebuilds.
+PYO3_CONFIG_FILE = { value = "rerun_py/pyo3-build.cfg", relative = true }
```
Wumpf, line 32: doesn't that make `PYO3_PYTHON` above redundant/bad? The config file should also pin python itself, no?

jleibs: Ah yes, good call.

Issue #11359, opened 2025-09-29
CI: snapshots test often fail on Mac runners.


emilk:
We have an intermittent CI failure on our Mac runners, related to our UI snapshot tests. Restarting the whole CI job fixes it, but adding just a retry to the individual step does not (https://github.com/rerun-io/rerun/pull/11171)

### Excerpt

```
  [2025-09-26T14:56:22Z DEBUG re_renderer::device_caps] Found the following adapters:
  [2025-09-26T14:56:22Z DEBUG re_renderer::device_caps] * backend: Metal, device_type: IntegratedGpu, name: "Apple Paravirtual device"
  [2025-09-26T14:56:22Z INFO  re_renderer::device_caps] Picked adapter: AdapterInfo { name: "Apple Paravirtual device", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }
  [2025-09-26T14:56:22Z DEBUG re_renderer::context] wgpu adapter backend: Metal, device_type: IntegratedGpu, name: "Apple Paravirtual device"
  
  thread 'test_text_documents' panicked at crates/viewer/re_view_text_document/tests/text_document_test.rs:68:13:
  Error rendering image: "Timeout"

…


  We timed out while waiting on the last successful submission to complete!
```

### Full
<details>
```
          SLOW [> 60.000s] re_view_spatial::video test_video_stream_codec_h265
          SLOW [> 60.000s] re_view_text_document view_class::test_help_view
          SLOW [> 60.000s] re_view_text_document::text_document_test test_text_documents
          SLOW [> 60.000s] re_view_text_log view_class::test_help_view
          SLOW [> 60.000s] re_view_time_series view_class::test_help_view
          FAIL [  60.287s] re_view_text_document::text_document_test test_text_documents
  ──── STDOUT:             re_view_text_document::text_document_test test_text_documents
  
  running 1 test
  test test_text_documents has been running for over 60 seconds
  test test_text_documents ... FAILED
  
  failures:
  
  failures:
      test_text_documents
  
  test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 60.25s
  
  ──── STDERR:             re_view_text_document::text_document_test test_text_documents
  [2025-09-26T14:56:22Z DEBUG re_renderer::device_caps] Found the following adapters:
[...]

PR #12333, opened 2026-01-07
Don't capture test output on mac runners

lucasmerlin:
### Related

- https://github.com/rerun-io/rerun/issues/11359

### What

Our mac ci runs sometimes hang. Since we capture the output we don't see which test are being executed before it hangs. 
This turns off capturing, which should hopefully show us if we always hang at the same test or if it's random.

### DIFF
--- a/scripts/ci/rust_checks.py
+++ b/scripts/ci/rust_checks.py
@@ -84,10 +84,13 @@ def run_cargo(
             f"{os.getcwd()}/{clippy_conf}"
         )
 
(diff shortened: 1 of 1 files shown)


lucasmerlin:
@rerun-bot full-check

Wumpf:
has the desired outcome: log now has all tests in there unconditionally
https://github.com/rerun-io/rerun/actions/runs/20775808210/job/59661190816

Wumpf (approved):

PR #12338, opened 2026-01-07
Larger than ram time panel UI

IsseW:
### Related

* Closes RR-3133
* Closes RR-3001

### What

Display loaded/unloaded state of chunks in the time panel.

https://github.com/user-attachments/assets/5f543b77-1001-464a-b162-82e015f32f4e

### DIFF
--- a/crates/store/re_entity_db/src/rrd_manifest_index.rs
+++ b/crates/store/re_entity_db/src/rrd_manifest_index.rs
@@ -453,13 +453,13 @@ impl RrdManifestIndex {
                     })
                 }));
 
-                ranges.extend(time_range_merger::merge_ranges(&mut scratch));
+                ranges.extend(time_range_merger::merge_ranges(&scratch));
 
                 scratch.clear();
             }
         }
 
-        time_range_merger::merge_ranges(&mut ranges)
+        time_range_merger::merge_ranges(&ranges)
             .into_iter()
             .filter(|r| r.loaded)
             .map(|r| r.range)


crates/store/re_entity_db/src/rrd_manifest_index.rs
```diff
@@ -417,37 +430,239 @@ impl RrdManifestIndex {
...
+                                // To not leave any gaps between states, expand the prioritized current state
+                                // to start at the end of the last state
+                                range.min = last_range.max;
+                                new_ranges.push((loaded, range));
+
+                                None
+                            }
+                        }
+                    }
+                };
```
emilk: I have a hard time following the logic here. Maybe some ASCII art could help?

It would also be a lot easier to follow if the closure with implicit capturing was replaced by a `struct` + member function

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -368,23 +391,105 @@ fn smooth(density: &[f32]) -> Vec<f32> {
...
             let mut sum = 0.0;
+            let mut loaded = LoadState::Loaded;
             for (j, &k) in kernel.iter().enumerate() {
-                if let Some(&density) = density.get((i + j).saturating_sub(2)) {
-                    debug_assert!(density >= 0.0);
-                    sum += k * density;
+                if let Some(bucket) = buckets.get((i + j).saturating_sub(2)) {
+                    debug_assert!(bucket.density >= 0.0);
+                    sum += k * bucket.density;
+                    loaded = loaded.and(bucket.loaded);
```
emilk, line 402: Won't this just always shrink the `loaded` state at the edges? I don't think that makes sense. No need to "blur" `loaded` imho

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -368,23 +391,105 @@ fn smooth(density: &[f32]) -> Vec<f32> {
...
+                density: sum,
+                loaded,
+            }
         })
         .collect()
 }
 
 // ----------------------------------------------------------------------------
 
+pub fn paint_loaded_indicator_bar(
```
emilk, line 422: docstring

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -368,23 +391,105 @@ fn smooth(density: &[f32]) -> Vec<f32> {
...
+            let dashed_line = egui::Shape::dashed_line_with_offset(
+                &[egui::pos2(range.min, y), egui::pos2(range.max, y)],
+                ui.visuals().widgets.noninteractive.fg_stroke,
+                &[line],
+                &[gap],
+                ui.input(|i| (i.time * speed) % (gap as f64 + line as f64) - line as f64) as f32,
+            );
+
+            ui.painter()
+                // Need to clip because offsetting the dashed line may end up outside otherwise
```
emilk, line 486: If `epaint::Shape::dashed_line_with_offset` paints outside of the given line range, then I would consider that a bug in `epaint`, and we should fix it there (as well). I would expect the `offset` there to be modulo the line length

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -446,13 +552,24 @@ pub fn build_density_graph<'a>(
...
+        &item.entity_path,
+        item.component,
+    ) {
+        data.add_chunk_range(entry.time_range, entry.num_rows, LoadState::Unloaded);
+    }
+
     // NOTE: These chunks are guaranteed to have data on the current timeline
-    let (chunk_ranges, total_events): (Vec<(Arc<Chunk>, AbsoluteTimeRange, u64)>, u64) = {
+    let (chunk_ranges, total_events): (
+        Vec<(Arc<re_chunk_store::Chunk>, AbsoluteTimeRange, u64)>,
```
emilk, line 573: Why the fully-qualified path now? Are there more than one `Chunk` type in scope?

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -368,23 +391,105 @@ fn smooth(density: &[f32]) -> Vec<f32> {
...
+    if should_load
+        && let Some(start) = time_ranges_ui.x_from_time(full_range.min.into())
+        && let Some(end) = time_ranges_ui.x_from_time(full_range.max.into())
+    {
+        let range = x_range.intersection(Rangef::new(start as f32, end as f32));
+        // How many pixels the gap is in the dashed line
+        let gap = 5.0;
+        // How many pixels each line is in the dashed line
+        let line = 3.0;
+        // Animation speed of the loading in pixels per second
```
emilk: In egui we use "pixels" to mean "physical display pixels", and "points" to mean "logical UI points", so these are all points, not pixels

crates/viewer/re_time_panel/src/data_density_graph.rs
IsseW: Probably just have touched the code a bit, and that's the only usage of `re_chunk_store::Chunk` in this module

crates/viewer/re_time_panel/src/data_density_graph.rs
IsseW: If we don't it starts looking like this (loaded edges on unloaded chunks)

Screenshot 2026-01-07 at 14 42 54

or reverse if we default to unloaded.

And also doesn't it kind of make sense that if we display some of the unloaded rows in a bucket that we show that whole bucket as unloaded?

crates/viewer/re_time_panel/src/data_density_graph.rs
IsseW: Looked a bit at the implementation in epaint and don't think that's the issue. I think it may just be that the line stroke is outside? The issue I noticed was a pixel at the edge of the loaded range flickering.

emilk: could be some pixel-rounding then (epaint rounds line segments to physical pixels in order to make them crisper). Carry on :)

crates/store/re_entity_db/src/rrd_manifest_index.rs
IsseW: Yeah agreed, a lot happening there. Factored it out into its own file and added some comments. Hopefully should be easier to read now.

crates/viewer/re_time_panel/src/data_density_graph.rs
emilk: Ah right… So that's because the default state for the zero-value buckets is `Loaded`. We could solve that by having zero-buckets be coded with a third "None" state value, that is weaker than both `Loaded` and `Unloaded` (tribool).

The smoothing is mostly there to make the graph prettier, but I think having the vertical coloring be precise is just _better_. As in: if this pixel is bright, we know we have all the data for that pixel.

But is is a very small blur radius so 🤷

crates/store/re_entity_db/src/rrd_manifest_index/time_range_merger.rs
```diff
@@ -0,0 +1,231 @@
...
+            //          range:    |----loaded----|
+            // result:
+            // new last_range: |------loaded------|
+            //
+            // case 2:
+            //     last_range: |--unloaded--|
+            //          range:                  |-unloaded-|
+            //
+            // result:
+            // new last_range: |---------unloaded----------|
```
emilk: So all gaps in the data gets filled in as "unloaded"? Then can't this whole algorithm just be replaced by merging of all loaded ranges, and keeping track of the global min/max of all unloaded ranges?

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -96,7 +102,13 @@ impl DensityGraph {
         let max_x = x_range.max + MARGIN_X;
         let n = ((max_x - min_x) * DENSITIES_PER_UI_PIXEL).ceil() as usize;
         Self {
-            buckets: vec![0.0; n],
+            buckets: vec![
+                Bucket {
+                    density: 0.0,
+                    loaded: LoadState::Loaded,
```
emilk, line 108: This is where we could do `LoadState::None`, and then adapt `LoadState::and` so that `None` always "looses"

crates/store/re_entity_db/src/rrd_manifest_index/time_range_merger.rs
IsseW: > So all gaps in the data gets filled in as "unloaded"? 

Not all gaps, for example we don't want to show gaps between loaded data as unloaded.

> Then can't this whole algorithm just be replaced by merging of all loaded ranges, and keeping track of the global min/max of all unloaded ranges?

Not entirely sure how you imagine that, but how I see it that wouldn't work.

crates/viewer/re_time_panel/src/data_density_graph.rs
IsseW: image

ending up with this, because data on one sub-entity is loaded. And getting some aliasing issues then. So think I'll revert to not having the `None` state

crates/store/re_entity_db/src/rrd_manifest_index/time_range_merger.rs
emilk: Ok, so we don't collapse gaps between loaded ranges, but we collapse gaps between unloaded ranges. Why exactly?

Each time point can have one of three states:
* no data
* unloaded data
* loaded data

And the point of this file is to do a logical AND on all these three states (using the tribool-and discussed else-where)

But if we collapse the gaps between unloaded ranges then things will be wrongly categorized as "unloaded" that should be categorized as "no data"?

IsseW: > Ok, so we don't collapse gaps between loaded ranges, but we collapse gaps between unloaded ranges. Why exactly?

We do collapse the gaps between loaded ranges into a larger loaded range. The intention of the algorithm is to mark everything as either `loaded` or `unloaded`.

```
// case 2:
//     last_range: |--loaded--|
//          range:                  |-loaded-|
//
// result:
// new last_range: |---------loaded----------|
```

Maybe a good example why I think this is needed is this:

Say we have this, which is a very common scenario for for example videos.

```
comp1:    |---------------------------loaded-------------------------------|
comp2: |unloaded|   |unloaded|   |unloaded|   |unloaded|   |unloaded|   |unloaded|
```

We don't want the end result to be that the gaps between the unloaded ranges are displayed as loaded. Although in this case this is handled by the fact that we first merge ranges per-component. Then merge them all.

IsseW: Should note that this function is not used for the timepoint buckets, and is only used for the top indicator for what sections has all data loaded.

emilk: > The intention of the algorithm is to mark everything as either `loaded` or `unloaded`.

So then we only need to look at the loaded ranges, and then assume everything else is unloaded. That should simplify the algorithm a lot: it's just a union of all loaded ranges.

emilk: Actually, my bad: all we need to look at are the _unloaded_ ranges and do a union on them. The result is everything that should have a loading animation; the rest should not.

PR #12339, opened 2026-01-07
Fix `animated_urdf` Python example

oxkitsune:
### What

Fixes the `animated_urdf` Python example spawning the viewer on CI


oxkitsune:
@rerun-bot full-check

PR #12342, opened 2026-01-07
Automatically cast to `Float64Array` in time series views

grtlr:
### Related

* Closes RR-3006

### What

This PR makes it possible to plot any data type (logged as `Scalars:scalars`) that can natively be cast to `Float64Array`. Note that the result of the cast is currently not cached and therefore recomputed on every frame.


crates/store/re_chunk/src/lib.rs
```diff
@@ -39,7 +39,12 @@ pub use self::chunk::{
 };
 pub use self::helpers::{ChunkShared, UnitChunkShared};
 pub use self::iter::{
-    ChunkComponentIter, ChunkComponentIterItem, ChunkComponentSlicer, ChunkIndicesIter,
+    // TODO(grtlr): Right place? Maybe `re_view`?
+    CastToPrimitive,
```
Wumpf: I think primitive casts is fine to have here for now, but I'd move casting to a separate cast module and not cram it into `iter.rs`
(... then again I'm always for moving stuff further up in the tree to keep things slimmer ;-))

crates/store/re_chunk/src/iter.rs
```diff
@@ -740,6 +740,106 @@ impl ChunkComponentSlicer for bool {
...
+    T: arrow::datatypes::ArrowNativeType + Clone,
+    I: Iterator<Item = Span<usize>>,
+{
+    type Item = Cow<'a, [T]>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let span = self.component_spans.next()?;
+        match &self.values {
+            Cow::Borrowed(values) => Some(Cow::Borrowed(&values[span.range()])),
+            // TODO(grtlr): This `clone` here makes me sad, but I don't see a way around it.
```
Wumpf: it's a tricky one because in this branch we get a lifetime dependency on `self` :thinking:
Neither my borrow checker nor llm fu is good enough to fix it right now

crates/viewer/re_view_time_series/src/series_query.rs
```diff
@@ -21,8 +24,10 @@ pub fn determine_num_series(all_scalar_chunks: &ChunksWithComponent<'_>) -> usiz
         .iter()
         .find_map(|chunk| {
             chunk
-                .iter_slices::<f64>()
-                .find_map(|slice| (!slice.is_empty()).then_some(slice.len()))
+                // TODO(grtlr): The comment above is even more important when we do casting (allocations) here.
+                // TODO(grtlr): Unless we're on the happy path this allocates and happens everytime the visualizer runs!
+                .iter_slices::<CastToPrimitive<arrow::datatypes::Float64Type, f64>>()
```
Wumpf, line 29: could we not just iterate "anonymous" slices here instead? :thinking:

Wumpf (approved):
delightfully simple

crates/viewer/re_view_time_series/src/series_query.rs
grtlr: I could not get the compiler to comply.

PR #12344, opened 2026-01-07
Add nextest timeout to prevent forever hanging tests on mac

lucasmerlin:


.config/nextest.toml
```diff
@@ -0,0 +1,3 @@
+[profile.default]
```
emilk, line 3:
```suggestion
# See https://nexte.st/docs/configuration/

[profile.default]
```

PR #12307, opened 2026-01-01
Rust SDK: `blueprint` support

sectore:
### Related

* Closes #6110
* Part of #5521

### What

This PR introduces `blueprint` support for the Rust SDK. Very similar to Python's implementation.

The Rust implementation is not fully completed, but this PR might be a good foundation to use it today and to work further on. Features have been tested with examples (see below).

## Configurable Panels:

- [x] (Blueprint) `Panel`
- [x] `SelectionPanel`
- [x] `TimePanel`

## Layout

- [x] `Horizontal`
- [x] `Grid`
- [x] `Tabs`
- [x] `Vertical`

## Views

- [x] `MapView`
- [x] `Spatial2DView`
- [x] `Spatial3DView`
- [x] `TextDocumentView`
- [x] `TimeSeriesView`

_Following views are still missing and might be added by another PRs._

- `BarChartView`
- `DataframeView`
- `GraphView`
- `TensorView`
- `TextDocumentView`
- `TextLogView`

## Examples:

- `blueprint` (similar to Python `blueprint` example)

   ```sh
   cargo run -p blueprint
   ```
   
   blueprint-example

- `blueprint_stocks` (similar to Python `blueprint_stock` example)

   ```sh
   cargo run -p blueprint_stocks -- --blueprint grid
   ```
  
   blueprint-stocks-example

 
- `fit-activities-rerun` (external)

   ```sh
   # clone branch `blueprint-rs`, which points `rerun` to current PR
   git clone -b blueprint-rs git@github.com:sectore/fit-activities-rerun.git
   cd fit-activities-rerun
   cargo run -- --fit ./data/fitdecode/Edge810-Vector-2013-08-16-15-35-10.fit --map dark
   ```

   
   far-example

PR #12347, opened 2026-01-07
Make `RecordingStreamBuilder::with_blueprint()` apply to everything, not just `spawn()`.

kpreid:
### Related

Further revises code that was just introduced in #12307.

### What

Makes `RecordingStreamBuilder::with_blueprint()` and `with_default_blueprint()` apply to every function that builds a stream. Previously, the field was only read by `serve()` and not any other case. I believe this behavior will be more what users expect; certainly I was baffled by why my blueprint was having no effect at all.

This change also introduces the helper function `create_recording_stream()`, which is not strictly necessary but greatly reduces code duplication.

### DIFF
--- a/crates/top/re_sdk/src/blueprint/api.rs
+++ b/crates/top/re_sdk/src/blueprint/api.rs
@@ -180,3 +180,14 @@ impl Blueprint {
         Ok(())
     }
 }
+
+impl BlueprintOpts {
+    /// Send the blueprint to the given recording stream.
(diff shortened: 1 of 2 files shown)


Wumpf (approved):
Thanks! This makes a lot more sense and with that refactor even more readable and compact <3

Wumpf:
ci fails because of cargo deny, is being handled elsewhere right now

PR #12348, opened 2026-01-07
Shuffle docs based on products

ntjohnson1:
### Related
Alternate to https://github.com/rerun-io/rerun/pull/12327


ntjohnson1:
https://github.com/rerun-io/rerun/pull/12355

PR #12349, opened 2026-01-08
Make `py-fmt` cover the entire repo

abey79:
### What

In the recent move to uv stuff, the scope of `py-fmt` was reduced to only some of our packages. This PR attempt to bring back coverage to the entire repo. This did incur some diff on some files, but they actually seem ok.


abey79:
merging because the ci failure is unrelated

PR #12351, opened 2026-01-08
Added missing PIL dependency for upload_image.py

danielduberg:
### What

Added `pillow` as a base PyPi dependency as `upload_image.py` script uses `PIL`. Without this, I get this error:

```sh
  File "/Users/dduberg/rerun_examples/rerun/scripts/upload_image.py", line 58, in <module>
    import PIL
ModuleNotFoundError: No module named 'PIL'
```


Wumpf:
@danielduberg you need to check in the changed pixi lock file

PR #12352, opened 2026-01-08
move tower on failure logs to debug level

zehiko:
### What

We get error logs on grpc failure from ``tower-http`` and this is unwanted verbosity that shows up in difference places, notebook for example. 

```
[2026-01-08T10:00:17Z ERROR tower_http::trace::on_failure] response failed classification=Code: 16 latency=203 ms
```

This is an attempt to make sure those logs are only shown at DEBUG level.  

I'm failing to test this as I can't use locally built rerun sdk for some reason anymore, but I believe the relevant callback is responsible for the spammy logs. 

Note: see @emilk's comment below, this will basically permanently remove on_failure logs for tower as we have a rule in ``re_log`` for the ``tower`` lib. Added a code comment in case someone wants to look at these specific logs.

### DIFF
--- a/crates/utils/re_perf_telemetry/src/grpc.rs
+++ b/crates/utils/re_perf_telemetry/src/grpc.rs
@@ -743,6 +743,8 @@ pub type ClientTelemetryLayer = tower::layer::util::Stack<
 // we ultimately can add all the same things that we have on the server as we need them.
 pub fn new_client_telemetry_layer() -> ClientTelemetryLayer {
     let trace_layer = tower_http::trace::TraceLayer::new_for_grpc()
+        // Note: we're actually disabling all DEBUG level logs for `tower` in re_log, so if you want to enable it
+        // you'll need to adjust that as well. See crates/utils/re_log/src/lib.rs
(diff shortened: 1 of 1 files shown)


zehiko:
@rerun-bot full-check

emilk:
Before:
image

After:
image

Looks like the logging is _gone_?

zehiko:
@emilk  thanks for verifying, I was failing to do so. Ok to merge this then?

emilk:
If we are ok with the log message being _completely_ gone. It is an improvement from my POV, but I think the intention of the PR was to demote it to DEBUG log level, and that seems to have failed.

Oh right, this is why:

https://github.com/rerun-io/rerun/blob/72429308ccfbd48d07b3a8bf049a25ea6872b543/crates/utils/re_log/src/lib.rs#L69-L81

We say "report errors on debug" and at the same time we mute all debug spam from tower :)

So maybe comment the new code to explain this

zehiko:
> We say "report errors on debug" and at the same time we mute all debug spam from tower :)

had no way to test it, hence I didn't know if this showed at DEBUG level, and had no idea about this code, thanks for the tip. 

Ok, will add a comment and merge as is.

PR #12354, opened 2026-01-08
Add option to hide all notification popups/toasts

emilk:
Requested by @nikolausWest 

Here is the new option: 
image

Tested. Works.


emilk:
Yes! We save app options to browser local storage, and that works for notebooks

PR #12355, opened 2026-01-08
Reshuffle documentation organization

abey79:
### Related

- closes https://linear.app/rerun/issue/RR-3258/reorder-sections-to-focus-on-robotics-learning
- takes advantage of https://github.com/rerun-io/landing/pull/1156

### What

Pure reshuffling, no content added or changed. The main goal is to structure everything around the same three main themes:
- Logging and ingestion
- Visualising
- Query and transform

This PR moves existing and new redirects in the new `_redirects.yaml` file.

This paves the way to possible larger reorganisation in the future.

Overview of the changes:
- "Installing Rerun" now has subsection, with languages SDK and viewer as items. Also, the "Troubleshooting" section moved here, because these are mostly install-related.
- "Concept" are broken up into the main themes
  - "Application model" is still there, we need to work more on this thing.
- Likewise, "How-to" are broken into the main themes.
  - "Integrations" remains stand alone in there. Hard to classify.
- A bunch of section that used to be in "Reference" were moved to "Concept"
  - "Visualization":
    - entity-queries.md
  - "Query and transform":
    - dataframes.md
  - "Logging and ingestion":
    - video.md
    - data-loaders.md
    - mcap.md

### DIFF
--- a/docs/content/_redirects.yaml
+++ b/docs/content/_redirects.yaml
@@ -73,7 +73,6 @@ howto/dataframe-api: howto/query-and-transform/get-data-out
 howto/embed-rerun-viewer: howto/integrations/embed-web
 howto/fixed-window-plot: howto/visualization/fixed-window-plot
 howto/get-data-out: howto/query-and-transform/get-data-out
-howto/integrations: howto/integrations/embed-notebooks
 howto/limit-ram: howto/visualization/limit-ram
 howto/logging: howto/logging-and-ingestion/send-columns
 howto/mcap: howto/logging-and-ingestion/mcap
--- a/docs/content/howto/integrations.md
+++ b/docs/content/howto/integrations.md
@@ -1,0 +1,6 @@
+---
+title: Integrations
+order: 400
+---
+
+How-to guides for integrating Rerun with other tools.


nikolausWest:
Structure looks good to me. Three categories under "Getting started" are differently named than in the other sections.

nikolausWest:
To make the language consistent I suggest:

- Log and Ingest
- Visualize
- Query and Transform

The two extra categories that would round it out and will make sense in the future might be _Store and Manage_ and _Train_ making it:

- Log and Ingest
- Store and manage
- Visualize
- Query and Transform
- Train

docs/content/_redirects.yaml
```diff
@@ -0,0 +1,99 @@
+# Redirect Table
```
ntjohnson1, line 1: Does the build validate that these paths make sense? If so I won't do the manual labor of comparing paths. Not sure how to review this.

docs/content/concepts/logging-and-ingestion.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Log and Ingest
+order: 100
+---
+
+This section covers how data is structured and ingested into Rerun.
+
+<!-- TODO(ab): it would be nice to auto-generate a TOC here -->
```
ntjohnson1, line 8: Yes autogenerated everything would be excellent

docs/content/concepts/logging-and-ingestion/data-loaders.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Data-loaders
+order: 800
+---
+
+Extending Rerun's file loading capabilities with custom data loaders.
```
ntjohnson1: NIT: data loader or data-loader? We use both in this file.

docs/content/overview/installing-rerun.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Installing Rerun
+order: 300
+---
+
+Choose what you want to install:
+
+-   [Viewer](./installing-rerun/viewer.md) - the standalone Rerun Viewer application
```
ntjohnson1: Nit this order doesn't match the side bar. Either put viewer at bottom of this list or change order in viewer.md

docs/content/_redirects.yaml
abey79: Seems easy enough to add a CI check. On it.

scripts/ci/check_doc_redirects.py
```diff
@@ -0,0 +1,129 @@
...
+
+import yaml
+
+DOCS_ROOT = Path(__file__).parent.parent.parent / "docs" / "content"
+REDIRECTS_FILE = DOCS_ROOT / "_redirects.yaml"
+
+
+def get_deleted_and_renamed_docs(base_branch: str = "main") -> tuple[list[str], list[str]]:
+    """Get lists of deleted and renamed doc paths relative to docs/content/."""
+    result = subprocess.run(
```
ntjohnson1: NIT: I think this is nicer and more stable if we use the git python sdk but for expedience definitely not blocking

abey79: ah good call, we use that elsewhere already iirc.

abey79:
The build-cli job is expected to fail, since there is a diff in the cli.md file. It's not due to it not being outdated wrt the actual CLI help, so ok to leave as is.

docs/content/_redirects.yaml
abey79: done

PR #12359, opened 2026-01-08
Virtual ChunkStore 1: shallow GC

teh-cmc:
## Broadly

This PR is the first in a series of incremental steps towards implementing a `ChunkStore` and in-memory query stack that fully support the notion of virtual/offloaded chunks.

Virtual/Offloaded chunks, as opposed to physical/loaded chunks, are chunks that are not currently loaded in local volatile memory. This can happen either because they were merely referenced by an RRD manifest but never actually loaded (aka `insert_rrd_manifest`, which will be introduced in a follow up), or simply because they were evicted by garbage collection.

The end goal we're aiming for is that a local `ChunkStore` should always be capable of answering the queries (latestat/range/dataframe) that are coming its way, regardless of whether the required data is still available locally at the time of the query.
When the data is not available, the query result should indicate so, and specify which bits of data must be downloaded first in order to compute the correct results for this query.
It is the responsibility of the caller to fetch that data however they wish: the store itself never performs any kind of IO.

## This PR

This PR introduces the first traces of the virtual/offloaded vs. physical/loaded terminology and in particular overhauls the garbage collector implementation so that it never clears the local chunk indices, effectively turning these previously physical indices into virtual ones instead.
A nice side effect of this is that the GC is now substantially faster, but virtue of having much less work to do.

This should work great both for legacy as well Redap-powered data paths:
* For Redap-based data paths, everything from the time panel to the store is aware of the RRD manifest and knows how to keep track of things.
* For legacy data paths, e.g. loading a file straight into the viewer without going through a Redap interface, things will still work fine because, even though the store indices advertise more data than there is, the time panel is still updated via the GC store events and therefore will not even try to query that data anyway.

---

TODO:
* [x] Try this out with an OSS server loading a giant recording with low memory limit
* [x] Try this out with by legacy-loading a giant recording with low memory limit

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -5,8 +5,6 @@ environments:
     - url: https://conda.anaconda.org/conda-forge/
     indexes:
     - https://pypi.org/simple
-    options:
-      pypi-prerelease-mode: if-necessary-or-explicit
     packages:
       linux-64:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
@@ -21,12 +19,14 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/binutils-2.44-h4852527_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/binutils_impl_linux-64-2.44-h4bf12b8_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/binutils_linux-64-2.44-h4852527_1.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/brotli-python-1.2.0-py311h66f275b_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/buf-1.57.0-ha8f183a_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/bzip2-1.0.8-hda65f42_8.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/c-ares-1.34.5-hb9d3cd8_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/c-compiler-1.6.0-hd590300_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/ca-certificates-2025.8.3-hbd8a1cb_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cairo-1.18.4-h3394656_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/cffi-2.0.0-py311h5b438cf_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/clang-16-16.0.6-default_hddf928d_15.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/clang-16.0.6-default_hfa515fb_15.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/clang-format-16-16.0.6-default_hddf928d_15.conda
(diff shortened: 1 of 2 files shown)


crates/store/re_chunk_store/src/drop_time_range.rs
teh-cmc, line 1: no semantic changes, just adapting to the new `remove_chunks` definition

crates/store/re_chunk_store/src/stats.rs
teh-cmc, line 1: Stats only account for physically loaded data.

crates/store/re_chunk_store/src/store.rs
```diff
@@ -420,32 +433,64 @@ pub struct ChunkStore {
...
-    /// All [`ChunkId`]s currently in the store, indexed by the smallest [`RowId`] in each of them.
+    /// All *physical* [`ChunkId`]s currently in the store, indexed by the smallest [`RowId`] in
+    /// each of them.
     ///
     /// This is effectively all chunks in global data order. Used for garbage collection.
+    ///
+    /// During garbage collection, physical chunks are offloaded from memory and become virtual
+    /// chunks instead. At the same time, their IDs are removed from this set, which is how we
+    /// distinguish virtual from physical chunks.
     pub(crate) chunk_ids_per_min_row_id: BTreeMap<RowId, ChunkId>,
```
teh-cmc, line 456: These 2 indices are the core of it all: they are the only purely physical indexes (i.e. they are the only remaining things actually affected by garbage collection). It's thanks to them that we know what is or isn't virtual in the first place.

crates/store/re_log_encoding/src/rrd/footer/instances.rs
teh-cmc, line 1: unrelated fix that was never backported from the `project/larger-than-ram` branch

crates/store/re_chunk_store/src/gc.rs
teh-cmc, line 1: This big overhaul can be summarized to: do less, much less.

Instead of tracking down every single index that refers to a chunk being dropped, there is now only 2 that we care about, and they are very simple to access.

crates/store/re_entity_db/src/entity_db.rs
```diff
@@ -666,14 +666,20 @@ impl EntityDb {
...
             // (e.g. when paused in the live camera example).
             // To be perfect it would need margins (because of latest-at), i.e. we would need to know
             // exactly how far back the latest-at is of each component at the current time…
             // …but maybe it doesn't have to be perfect.
+            //
+            // NOTE: This will only apply if the GC is forced to fall back to row ID based collection,
+            // otherwise timestamp-based collection will ignore it.
             protected_time_ranges: Default::default(),
 
             furthest_from: time_cursor.map(|(timeline, time)| (*timeline.name(), time)),
```
teh-cmc, line 685: We should really disable `furthest_from` when loading via a legacy path @emilk, it's just a horrible experience because the cursor lags behind on the left of the time panel, all while data gets loaded at high speed on right, only to immediately get garbage collected away because the cursor is all the way to the left, and then ofc there's no way to re-fetch it later :melting_face: 

(I guess that's not so much of a problem when the thing starts in "rightmost mode" (I forgot what we call it), but that's not the case with a file off of a disk, it seems)

teh-cmc:
@rerun-bot full-check

crates/store/re_chunk_store/src/gc.rs
```diff
@@ -307,438 +322,108 @@ impl ChunkStore {
...
-
-            for chunk in
-                chunks_in_priority_order.filter(|chunk| !protected_chunk_ids.contains(&chunk.id()))
-            {
-                if options.is_chunk_temporally_protected(&chunk) {
-                    continue;
-                }
-
+                // `find_all_protected_physical_chunk_ids` is rather expensive so make sure we only
+                // compute it if we couldn't find enough chunks to remove already.
```
emilk, line 340: nice

emilk (approved):
Super!

PR #12360, opened 2026-01-08
Fix missing `child_frame`/`parent_frame` arguments from pinhole constructor

Wumpf:
Quite the oversight. We have unit tests for using these, but nothing that goes through the python sdk. I'll likely follow up with a snippet using this.


Wumpf:
Thanks for asking, I had a look around again: the only ones I can think of are parent/child on pinhole & transform3d. On `InstancePoses3D` we didn't add anything. `TransformAxes3D` doesn't have an _ext file.

Pretty nasty issue as such - anywhere we have a custom constructor it can easily happen that one misses updating it :/
Don't have a good idea to reliably check for that

IsseW:
> Don't have a good idea to reliably check for that

I wonder if we can have a warning if all fields aren't initialized in that function somehow 🤔

ntjohnson1:
> > Don't have a good idea to reliably check for that
> 
> I wonder if we can have a warning if all fields aren't initialized in that function somehow 🤔

I've been vibe coding various parsers for things that annoy me with moderate success. I think these one off tools are less risky since the alternative is no checking. We already have `lint-rerun` which is kind of a good place for these. The downside is we may eventually have a big mess of too many custom tools. Either was I got nerd sniped a bit, didn't find an easy off the shelf answer but for this specific custom ext class with custom init not specifying all __attr_init__ this is was my vibe coded check says (spot checking the missing bits looks rightish). Note that this was before your fix here so the child/parent frame stuff should be fixed now.

✅ arrows2d_ext.py:62 - All arguments provided
✅ arrows3d_ext.py:57 - All arguments provided
✅ asset3d_ext.py:67 - All arguments provided
✅ asset_video_ext.py:65 - All arguments provided
✅ boxes2d_ext.py:166 - All arguments provided
✅ boxes3d_ext.py:128 - All arguments provided
✅ capsules3d_ext.py:68 - All arguments provided
✅ clear_ext.py:28 - All arguments provided
✅ cylinders3d_ext.py:68 - All arguments provided
✅ depth_image_ext.py:121 - All arguments provided
✅ ellipsoids3d_ext.py:78 - All arguments provided
✅ encoded_image_ext.py:94 - All arguments provided
✅ geo_line_strings_ext.py:41 - All arguments provided

❌ Issues in geo_points_ext.py:45
   Missing arguments: class_ids
✅ image_ext.py:155 - All arguments provided
✅ image_ext.py:184 - All arguments provided
✅ image_ext.py:255 - All arguments provided
✅ mesh3d_ext.py:120 - All arguments provided

❌ Issues in pinhole_ext.py:160
   Missing arguments: child_frame, parent_frame
✅ points2d_ext.py:67 - All arguments provided
✅ points3d_ext.py:63 - All arguments provided

❌ Issues in segmentation_image_ext.py:67
   Missing arguments: draw_order
✅ series_points_ext.py:64 - All arguments provided
✅ tensor_ext.py:59 - All arguments provided
✅ transform3d_ext.py:150 - All arguments provided

❌ Issues in video_frame_reference_ext.py:69
   Missing arguments: draw_order, opacity

❌ Found issues in 4 call(s)

PR #12361, opened 2026-01-08
Fix incorrect handling for cameras and depth images for 3D views that use target frames other than the scene root

Wumpf:
### Related

* Fixes RR-3280

### What

Previously, if you set something other than the root of the scene to a view's target, camera visualizers (and as I found out depth images) would not correctly place their visualization. This is because we checked the "how to get to the root of the 3D view I'm embedded in"-property of pinhole-roots, rather than resolving to the target space.

Before:
image

After:
image

### DIFF
--- a/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
+++ b/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
@@ -407,7 +407,7 @@ impl TransformTreeContext {
 
     /// Returns the transform from a pinhole root to the target frame if any.
     ///
-    /// This is effectively the extrinsic portion or the pinhole's transform to the view's target frame.
+    /// This is effectively the extrinsic portion of the pinhole's transform to the view's target frame.
     /// Note that this may otherwise not be accessible since we allow both extrinsics and intrinsics to be
     /// on a transform between two frames, with no transform frame representing just the extrinsics.
     ///


crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
```diff
@@ -88,28 +86,11 @@ impl VisualizerSystem for TransformAxes3DVisualizer {
...
 
-                    if let Some(pinhole_tree_root_info) =
-                        transforms.pinhole_tree_root_info(transform_info.tree_root())
+                    if let Some(target_from_camera) =
+                        transforms.target_from_pinhole_root(transform_info.tree_root())
                     {
-                        if transform_info.tree_root()
-                            == re_tf::TransformFrameIdHash::from_entity_path(
-                                &data_result.entity_path,
-                            )
```
Wumpf, line 97: removing this filtering here is technically a second fix: it didn't make sense to begin with because, again, the tree root may not be the target. The "we're _inside_ the 2d space" problem should already be filtered out here because in that case `transform_info_for_entity_or_report_error` should have already bailed out since it knows that we're doing a 3d visualization.

crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs
```diff
@@ -405,6 +405,38 @@ impl TransformTreeContext {
         self.transform_forest.pinhole_tree_root_info(frame)
     }
 
+    /// Returns the transform from a pinhole root to the target frame if any.
+    ///
+    /// This is effectively the extrinsic portion or the pinhole's transform to the view's target frame.
```
MichaelGrupp: typo?
```suggestion
    /// This is effectively the extrinsic portion of the pinhole's transform to the view's target frame.
```

PR #12356, opened 2026-01-08
Bump `google-cloud-storage` to `3.4.1`

grtlr:
### Related

* Attempt to fix https://github.com/rerun-io/rerun/actions/runs/20816542784/job/59794452838.

### What

Title.

### Todo

* [ ] @rerun-bot full-check

PR #12362, opened 2026-01-08
Pin urllib3 for now

ntjohnson1:
### Related
https://github.com/rerun-io/rerun/pull/12356
https://github.com/googleapis/google-resumable-media-python/issues/491

### What
Pin urllib3 until the google cloud storage sdk gets updated to support our compressed download.
There was a CVE but we only use this in 1 place on CI to download a file we host, so should be ok as a temporary detail.

* [x] full test


ntjohnson1:
@rerun-bot full-check

ntjohnson1:
@rerun-bot full-check

PR #12369, opened 2026-01-09
Improved auth section in server modal

lucasmerlin:
### Related

- closes https://linear.app/rerun/issue/RR-3277/improve-server-modal

### What

Improves the authentication section of the redap server modal. 
Screenshot 2026-01-09 at 13 12 31
Screenshot 2026-01-09 at 13 12 39

### DIFF
--- a/tests/rust/re_integration_test/tests/snapshots/dataset_ui_empty_form.png
+++ b/tests/rust/re_integration_test/tests/snapshots/dataset_ui_empty_form.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:52c9c6ca6f350ac6f6aba64e46af6520724e39a8a2f3bf442f3684f9404ca283
-size 60352
+oid sha256:78b37876a2ea0b4cc0c80eb06788f6e7a1e7b2003994fd41bdb854f564519e27
+size 60761


crates/store/re_redap_client/src/connection_registry.rs
```diff
@@ -227,24 +227,8 @@ impl ConnectionRegistryHandle {
...
-                    if err.is_client_credentials_error() {
-                        let mut inner = self.inner.write().await;
-
-                        // make sure that we're not deleting some token that another thread might
-                        // have just set
-                        if inner.saved_credentials.get(&origin) == saved_credentials.as_ref() {
-                            inner.saved_credentials.remove(&origin);
-                        }
-                    }
+            Self::try_create_raw_client(origin.clone(), credentials_to_try.into_iter()).await?;
```
lucasmerlin, line 244: I've removed this since it's really confusing to the user if they enter a token and it disappears (especially since they might want to check what they entered after getting the invalid token error).

gavrelina:
One last edit to iterate on a better copy, sounds a bit funny and inconsistent now.
- `Authenticate:`  change to `Authentication:`
- `Rerun account` change to `Account login`
- `With a token` change to `Access token`
- `No authentication` leave as is

PR #12370, opened 2026-01-09
Deprecate `Entry.update` in favor of `Entry.set_name`

abey79:
### Related

### What

Small planned API update that I forgot to do in 0.28.

As per the migration guide:

````md
## `Entry.update()` is deprecated in favor of `Entry.set_name()`

The `Entry.update()` method has been deprecated. Use `Entry.set_name()` instead for renaming entries.

```python
# Before (deprecated)
entry.update(name="new_name")

# After
entry.set_name("new_name")
```

The deprecated method will emit a `DeprecationWarning` and will be removed in a future release.
````


ntjohnson1 (approved):
Looks like we might do a 0.28.3 for another small missed api change. So consider patch?

abey79:
> Looks like we might do a 0.28.3 for another small missed api change. So consider patch?

I'd like to ship, but that's introducing soft-breaking changes in a patch release, with isn't great practice (e.g. we need a 0.28.3 migration guide?). All in all, I think i'm fine with waiting for 0.29. The stakes aren't super high, this is a rather niche API.

PR #12371, opened 2026-01-09
URDF loader: use `InstancePoses3D` for geometry scale

oxkitsune:
### What

In #12005 we started logging the origin transforms for geometry using `InstancePoses3D`. However the mesh scale was still being logged using a normal `Transform3D`, this fixes that and now meshes with a scale attribute should be scaled correctly.

Originally reported on our discord: https://discord.com/channels/1062300748202921994/1458859702359167106


MichaelGrupp:
oops... good catch! Thanks.

PR #12373, opened 2026-01-09
Draft: Add `transform_frame_prefix` to data loader API and use it for URDF

MichaelGrupp:
Relates to: [RR-3275](https://linear.app/rerun/issue/RR-3275)

Draft for now, because we might go for a different approach to
tackle the prefix option. But it works and can be tested, e.g. via
`py-build` and logging some URDF.

```python
import rerun as rr

if __name__ == "__main__":
    rr.init("test_urdf_prefix", spawn=True)
    rr.log_file_from_path("/your/robot.urdf", transform_frame_prefix="blabla_")
```

---

Adds `transform_frame_prefix` to the load file API, which allows
to prefix all transform frame IDs loaded by a data loader.

One use case, which is part of this PR, is to load a URDF robot
multiple times with different frame names via prefix. This is a
common practice in other tools like RViz.

### DIFF
--- a/crates/store/re_data_loader/src/lib.rs
+++ b/crates/store/re_data_loader/src/lib.rs
@@ -167,7 +167,7 @@ impl DataLoaderSettings {
         if let Some(transform_frame_prefix) = transform_frame_prefix {
             args.extend([
                 "--transform-frame-prefix".to_owned(),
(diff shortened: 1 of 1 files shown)


MichaelGrupp:
closing for now

PR #12376, opened 2026-01-09
Check for unset attr from extension class init

ntjohnson1:
### Related
Nerd sniped by #12360
I branched off of that to keep the pinhole fixes.

### What
Some of our code gen python code has custom __init__ functions that didn't expose all of the available parameters. Fix that and add a test to avoid future drift.

Ask claude to write an AST parser after looking at a few different options. Maybe there is something off the shelf 🤞 but it didn't bubble up immediately.

Run that AST parser to identify areas of mismatches. Manually confirm those were errors then fix them. Add parser to lint-rerun to enforce extension class stability moving forward.

* [x] land #12360 first

### DIFF
--- a/crates/build/re_types_builder/src/codegen/python/mod.rs
+++ b/crates/build/re_types_builder/src/codegen/python/mod.rs
@@ -429,14 +429,14 @@ fn check_ext_consistency(
             reporter.error(
                 ext_filepath.as_str(),
                 &obj.fqname,
-                "Could not find __attrs_init__ call".to_string(),
+                "Could not find __attrs_init__ call".to_owned(),
             );
         }
     } else {
         reporter.error(
             ext_filepath.as_str(),
             &obj.fqname,
-            "The __init__ method should call __attrs_init__ with all available fields".to_string(),
+            "The __init__ method should call __attrs_init__ with all available fields".to_owned(),
         );
     }
 }


scripts/ci/validate_attrs_init_calls.py
```diff
@@ -0,0 +1,506 @@
...
+class AttrsFieldVisitor(ast.NodeVisitor):
+    """AST visitor to extract attrs field definitions from a class."""
+
+    def __init__(self, target_class_name: str):
+        self.target_class_name = target_class_name
+        self.attrs_params: Set[str] = set()
+
+    def visit_ClassDef(self, node: ast.ClassDef) -> None:
+        if node.name == self.target_class_name:
+            # Look for field assignments and __attrs_clear__ method
```
Wumpf: why do we have to look for `__attrs_clear__`? it never has any parameters, it should be enough to check on the attributes alone

scripts/ci/validate_attrs_init_calls.py
```diff
@@ -0,0 +1,506 @@
...
+        type=Path,
+        nargs="?",
+        default=Path("."),
+        help="Path to scan for extension files (default: current directory)",
+    )
+    parser.add_argument("--file", type=Path, help="Validate a specific extension file instead of scanning")
+    parser.add_argument(
+        "--pattern", type=str, default="*_ext.py", help="File pattern to match when scanning (default: *_ext.py)"
+    )
+    parser.add_argument("--ci", action="store_true", help="CI mode: minimal output, exit with error code on issues")
```
Wumpf: all this seems like needless complexity, let's trim it down a bit? I have no idea when I'd ever user `--pattern` or `--ci`
`--file` maaaaybe, but it has a weird interaction with `path`

scripts/ci/validate_attrs_init_calls.py
ntjohnson1: This isn't looking at the signature for attrs_clear but rather the implementation where we basically call `__attrs_init__` with all arguments set to None.

ntjohnson1:
> Great stuff, but oof this is rather complex! I think it would have been a lot easier to do this with a bit of string matching in the codegen which is already insepecting `_ext.py` files. That said, it seems to work and find things and I don't want you to start over here, but would be nice to simplify some of the stuff the llm put in there so it's easier to check on stuff if the script doesn't do what we want after all

Ya, I put the fixes as a single commit in case we wanted to just merge the fixes then make a follow up issue to make a better check. I can spend 15 minutes or so looking into codegen approach to see if that's feasible/reuse any bits from there. The challenges I ran into for simpler string matching were:
* Inverted dependency: main class depends on ext
* The fact our `__init__` often has a different signature than `__attrs_init__` so I can just run inspect on both to compare
* The additional redirect to datatypes so ext has to know to look for main class and then potentially do an inheritance hop.

Wumpf:
the idea with a codegen based one would be that the `attrs_init` call in the extension has to assign all the fields that we _know_ the type has. The nice thing in codegen land is that we don't have to do anything fancy to find out what fields have to be written since we're the ones putting them there in the first place

crates/build/re_types_builder/src/codegen/python/mod.rs
```diff
@@ -325,6 +331,116 @@ impl ExtensionClass {
...
+    }
+
+    // Look for __attrs_init__ call using Python indentation structure
+    if contents.contains("__attrs_init__") {
+        let lines: Vec<&str> = contents.lines().collect();
+        let mut attrs_init_section = String::new();
+        let mut found_attrs_init = false;
+        let mut base_indent = 0;
+
+        for line in lines {
```
ntjohnson1, line 381: This definitely felt like it should have been regexable but I couldn't get it to play nice.

ntjohnson1:
> the idea with a codegen based one would be that the `attrs_init` call in the extension has to assign all the fields that we _know_ the type has. The nice thing in codegen land is that we don't have to do anything fancy to find out what fields have to be written since we're the ones putting them there in the first place

Ok I got this working in codegen. There are some gross bits but not as bad. I couldn't coerce extracting the field names from the `attrs_init`to be a regex so I just parsed it line by line. Sometimes our fields trace through recursively and sometimes they don't AFAICT archetypes and datatypes don't otherwise we do (but if I misunderstood then I'm not enforcing this correctly).

Adding it to codegen did find that we had a few datatypes places where we were using a different pattern for initializing. That seemed to work fine but I guess making things more standard probably doesn't hurt.

ntjohnson1:
> nice. I do like the added consistency here of always going through `__attrs_init__` and at the very least the business logic of enforcing this is much shorter. Great job and thanks for bearing with me!

No problem. The original bit was `fix the broken things` then finding the least bad enforcement approach. Definitely agree this isn't as bad as what I had before. Hopefully someone comes along and lets us know we missed the obviously better solution `<placeholder>` (fun fact github doesn't render < placeholder > with no spaces).

PR #12380, opened 2026-01-09
Use sitecustomize from rerun_dev_fixup to add RERUN_CLI_PATH to the environment

jleibs:
After moving to the uv environment, we broke the logic that used to handle running `rerun` from within the python environment using the python-based target.

As a rseult
```
uv run rerun
```
was pointing to the wrong location.

### DIFF
--- a/rerun_py/rerun_dev_fixup/__init__.py
+++ b/rerun_py/rerun_dev_fixup/__init__.py
@@ -1,10 +1,12 @@
 """Development helper for rerun-sdk editable installs."""
 
+from __future__ import annotations
+
 import os
 from pathlib import Path
 
 
-def init():
+def init() -> None:
     """
     Sitecustomize entrypoint that sets RERUN_CLI_PATH.
 


rerun_py/rerun_dev_fixup/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+
+def init() -> None:
+    """
+    Sitecustomize entrypoint that sets RERUN_CLI_PATH.
```
abey79, line 11: I'm assuming this is a typo but coffee hasn't kicked it yet.

```suggestion
    Customize entrypoint that sets RERUN_CLI_PATH.
```

oxkitsune: I think it refers to [this](https://docs.python.org/3/library/site.html#module-sitecustomize)?

abey79: Ah right, TIL

PR #12381, opened 2026-01-09
Clean up the pixi boostrapping via a python package

jleibs:
This fixes the wheel-test on windows nightly builds.
See example: https://github.com/rerun-io/rerun/actions/runs/20840495128/job/59887650747

### What

The problem is that our `uv` shim did not work consistently when launching `uv` as a subprocess within a python script. This is because python subprocess on windows won't consider `.bat` file if shell isn't set to true, leaving an annoying footgun.

## Solution

Instead, we introduce a new python package installed within the uv environment. This lets us use proper pypi-managed entry-points which act as executables on windows for the shim, side-stepping the problem.

This generally also cleans up the entry-point hackery by making it a proper python script instead of assorted bash shenagins. We only use the activation scripts to bootstrap into the single python activation script.


jleibs:
Test-run of nightly: https://github.com/rerun-io/rerun/actions/runs/20906416397

scripts/pixi/uvpy
```diff
@@ -1,8 +1,10 @@
-#!/bin/bash
-# Wrapper script to run "uv run python" using the local uv wrapper
-# This ensures we use .venv instead of pixi env
-
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-
-# Use our uv wrapper script with "run python"
-"${SCRIPT_DIR}/uv" run python "$@"
+#!/home/jleibs/rerun/.pixi/envs/default/bin/python3.11
```
grtlr: I think this might be causing some of the remaining CI failures.

grtlr:
Another nightly run: https://github.com/rerun-io/rerun/actions/runs/20912309547

PR #12385, opened 2026-01-12
URDF: Bundle pose + mesh scale in single `InstancePoses3D`

yujeong1jeong:
### What

The URDF loader now correctly applies `<mesh scale>` attributes to visual and collision geometries by emitting all pose components (translation, rotation, and scale) together in a single InstancePoses3D archetype instance.

- Why
Previously, mesh scale was logged as a separate component update in a different chunk row from the translation and rotation data. This violates the principle of atomic archetype emission and can result in incomplete or out-of-order component data, causing the renderer to receive poses without scale information. Scaled meshes would render at their unscaled size despite correct metadata being present in the store.

- How
1. Extract mesh scale at the point of use (`log_link()`) via `extract_instance_scale()`
2. Pass scale through the call chain: `log_link()` → `send_instance_pose_with_frame()` → `instance_poses_from_pose()`
3. Emit translation + rotation + scale as a single InstancePoses3D row via one `send_archetype()` call
4. Remove redundant scale handling from `log_geometry()` to eliminate duplication

- Result
  - Scaled mesh visuals and collisions render at correct dimensions
  - InstancePoses3D data is emitted atomically, preventing order-dependent rendering artifacts
  - Cleaner separation of concerns: mesh scale is handled at the pose level, not the geometry level

Doosan Robotics URDF example: https://github.com/DoosanRobotics/doosan-robot2/blob/humble/dsr_description2/urdf/p3020.urdf

### DIFF
--- a/crates/store/re_data_loader/src/loader_urdf.rs
+++ b/crates/store/re_data_loader/src/loader_urdf.rs
@@ -539,24 +539,12 @@ fn log_debug_format(
     )
 }
 
-fn is_identity_scale(x: f64, y: f64, z: f64) -> bool {
-    const EPS: f64 = 1e-12;
-    (x - 1.0).abs() < EPS && (y - 1.0).abs() < EPS && (z - 1.0).abs() < EPS
-}
-
 fn extract_instance_scale(geometry: &Geometry) -> Option<Vec3D> {
     match geometry {
         Geometry::Mesh {
             scale: Some(Vec3([x, y, z])),
             ..
-        } => {
-            // Only log a scale component if it isn't the identity.
-            if is_identity_scale(*x, *y, *z) {
-                None
-            } else {
-                Some(Vec3D::new(*x as f32, *y as f32, *z as f32))
-            }
-        }
+        } => Some(Vec3D::new(*x as f32, *y as f32, *z as f32)),
         _ => None,
     }
 }


oxkitsune:
Ah good catch! In #12371 I figured we can log the scale later as I figured the order doesn't matter.

I don't understand though; does your URDF not visualize correctly without this patch?

crates/store/re_data_loader/src/loader_urdf.rs
```diff
@@ -532,6 +539,28 @@ fn log_debug_format(
...
+        Geometry::Mesh {
+            scale: Some(Vec3([x, y, z])),
+            ..
+        } => {
+            // Only log a scale component if it isn't the identity.
+            if is_identity_scale(*x, *y, *z) {
+                None
+            } else {
+                Some(Vec3D::new(*x as f32, *y as f32, *z as f32))
+            }
```
oxkitsune: I'd argue that we can actually just always log the scale, shouldn't make much of a difference in data logged and it's more correct even.

Also makes the code a lot more neat!

yujeong1jeong:
> Ah good catch! In #12371 I figured we can log the scale later as I figured the order doesn't matter.
> 
> I don't understand though; 
> does your URDF not visualize correctly without this patch?

First of all, thank you for the quick response and for looking into this so promptly!

To clarify the technical details, the issue wasn't that the scale was missing from InstancePoses3D entirely—in commit 6063672, it was indeed being logged.

However, the problem was that the properties were being split across different RowIds:
`send_instance_pose_with_frame()` was logging translation + rotation (RowId A).
`log_geometry()` was logging the scale separately (RowId B).

Because these exist in separate chunk rows, Rerun treats them as distinct entries rather than merging them into a single pose. This caused the viewer to display two "partial" poses instead of one unified InstancePoses3D entry.

This PR fixes the behavior by extracting the scale before calling send_instance_pose_with_frame(). This ensures that translation, rotation, and scale are all bundled together in a single InstancePoses3D row for a complete and correct visualization.


**Before 6063672:** URDF geometry had no scale support - scaled meshes always rendered at 1x size.


Screenshot 2026-01-13 at 10 23 44 AM

**After 6063672:** Scale was extracted in `log_geometry()`, but InstancePoses3D (the pose that controls rendering) was already logged earlier without scale information. 



image

**Current Branch (fix):** Scale is extracted BEFORE creating InstancePoses3D and passed as a parameter. Now InstancePoses3D includes translation + rotation + **scale** from the start. Geometry renders correctly.


image

crates/store/re_data_loader/src/loader_urdf.rs
yujeong1jeong: Fixed in https://github.com/rerun-io/rerun/pull/12385/commits/bccbf91ae9581981ec70ae43e262efa89478bb05 .
Thanks for the suggestion, it definitely made the code much cleaner!

oxkitsune (approved):
LGTM, thanks!

PR #12390, opened 2026-01-12
Virtual ChunkStore 2: support partial query results at the store level

teh-cmc:
Second incremental step towards a virtual ChunkStore.

This time the focus is on making it possible to return partial results from the low-level ChunkStore query APIs:
```rust
/// The results of a latest-at and/or range relevancy query.
///
/// Since the introduction of virtual/offloaded chunks, it is possible for a query to detect that
/// it is missing some data in order to compute accurate results.
/// This lack of data is communicated using a non-empty [`QueryResults::missing`] field.
#[derive(Debug, Clone, Default, PartialEq)]
pub struct QueryResults {
    /// The relevant *physical* chunks that were found for this query.
    ///
    /// If [`Self::missing`] is non-empty, then these chunks are not enough to compute accurate query results.
    pub chunks: Vec<Arc<Chunk>>,

    /// The relevant *virtual* chunks that were found for this query.
    ///
    /// Until these chunks have been fetched and inserted into the appropriate [`ChunkStore`], the
    /// results of this query cannot accurately be computed.
    pub missing: Vec<ChunkId>,
}
```
We stop at the ChunkStore level for now: `re_query` is not ready to deal with these things, yet.

This is all straightforward except for one tricky edge-case: shallow removal is not enough when removing chunks that were compacted away (as opposed to GC'd away). That's because you don't want these pre-compacted IDs to linger on in the virtual indexes, else the results would always appear partial even though they aren't.

Some rare parts of the viewer-space code (such as the transform system) use the low-level query APIs directly. I've made sure to leave TODO breadcrumbs that we will address in the coming days/weeks.

---

Note: we do not make any effort to address lineage yet, which means that the returned missing chunk IDs might not make any sense from the caller's perspective. That's for a follow-up.

---

* Previous: https://github.com/rerun-io/rerun/pull/12359

### DIFF
--- a/crates/store/re_chunk_store/tests/correctness.rs
+++ b/crates/store/re_chunk_store/tests/correctness.rs
@@ -14,7 +14,6 @@ use re_sdk_types::ComponentIdentifier;
 
 // ---
 
-#[expect(clippy::unwrap_used)]
 fn query_latest_component<C: re_types_core::Component>(
     store: &ChunkStore,
     entity_path: &EntityPath,
@@ -28,7 +27,6 @@ fn query_latest_component<C: re_types_core::Component>(
         .latest_at_relevant_chunks(query, entity_path, component)
         .to_iter()
         .unwrap()
-        .into_iter()
         .filter_map(|chunk| {
             let unit = chunk.latest_at(query, component).into_unit()?;
             unit.index(&query.timeline()).map(|index| (index, unit))


crates/store/re_chunk/src/slice.rs
```diff
@@ -310,14 +310,14 @@ impl Chunk {
         chunk
     }
 
-    /// Densifies the [`Chunk`] vertically based on the `componentiptor` column.
```
emilk, line 313: "componentiptor" 😆 

Some fun search-and-replace is surely behind that name

crates/store/re_chunk_store/src/gc.rs
```diff
@@ -361,18 +361,152 @@ impl ChunkStore {
...
                 chunks_to_be_removed,
                 Some((start_time, options.time_budget)),
             )
         }
     }
 
-    /// Surgically removes a set of _temporal_ [`ChunkId`]s from all indices.
+    /// Surgically removes a set of _temporal_ [`ChunkId`]s from all *physical & virtual* indices.
+    ///
+    /// This only makes sense to use with compacted chunks: you don't want a chunk that was compacted
```
emilk: The term "compacted chunks" is ambiguous: it could be read either as chunks that are the _source_ or the _output_ of  compaction.

("Those chunks have been compacted and are no longer needed" vs "those chunks are compacted and not what is on the server").

Maybe we should introduce conceps like "source chunks" (chunks that exist in the source (e.g. redap)) vs "stored chunks" (chunks that could have been the output of compaction or splitting). Chunks that are neither compacted nor splitted would be _both_ source chunks and stored chunks. Or maybe we should always assign two IDs to stored chunks: the "source chunk id" and the "stored chunk id"? 🤔

crates/store/re_chunk_store/src/gc.rs
```diff
@@ -361,18 +361,152 @@ impl ChunkStore {
...
+    /// into a new larger chunk to linger on in our internal indices, both physical and virtual.
+    /// For garbage collection purposes, refer to [`Self::remove_chunks_shallow`] instead.
```
emilk: Does this logic apply only to compaction, or splitting too? 🤔

crates/store/re_chunk_store/src/gc.rs
```diff
@@ -361,18 +361,152 @@ impl ChunkStore {
...
+    ///
+    /// This is orders of magnitude faster than trying to `retain()` on all our internal indices,
+    /// when you already know where these chunks live.
+    pub(crate) fn remove_chunks_deep(
```
emilk, line 428: "deep" GC would also make sense for the blueprint store, since we just append to that forever, and have no need for virtual chunks without physical backing

crates/store/re_chunk_store/tests/correctness.rs
```diff
@@ -22,8 +22,10 @@ fn query_latest_component<C: re_types_core::Component>(
 ) -> Option<(TimeInt, RowId, C)> {
     re_tracing::profile_function!();
 
+    // NOTE: Purposefully ignoring virtual chunks -- these tests predate that.
```
emilk, line 25: Maybe add a `QueryResults::unwrap_chunks()` test-helper that panics if there are missing chunks?

crates/store/re_chunk_store/src/query.rs
```diff
@@ -606,6 +607,98 @@ impl ChunkStore {
...
+    /// them, load them, and then try the query again.
+    ///
+    /// [missing chunk IDs]: `Self::missing`
+    pub fn to_iter(&self) -> Option<impl Iterator<Item = &Arc<Chunk>>> {
+        if self.missing.is_empty() {
+            return Some(self.chunks.iter());
+        }
+
+        None
+    }
```
emilk, line 703:
```suggestion
    }
    
    /// Panics if there are missing chunks.
    #[cfg(feature = "testing")]
    pub fn unwrap_chunks(self) -> Vec<Arc<Chunk>> {
        …
    }
```

crates/store/re_chunk_store/src/query.rs
```diff
@@ -606,6 +607,98 @@ impl ChunkStore {
...
+}
+
+impl std::fmt::Display for QueryResults {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let Self { chunks, missing } = self;
+
+        let chunk_ids = chunks.iter().map(|c| c.id().to_string()).join(",");
+        let missing_ids = missing.iter().map(|id| id.to_string()).join(",");
+
+        f.write_fmt(format_args!("chunks:[{chunk_ids}] missing:[{missing_ids}]"))
```
emilk: maybe only write the `missing` part if there are any missing, to avoid ugly debug output in the common case

crates/store/re_query/src/latest_at.rs
```diff
@@ -636,6 +636,8 @@ impl LatestAtCache {
 
         let ((data_time, _row_id), unit) = store
             .latest_at_relevant_chunks(query, entity_path, component)
+            // TODO(RR-3295): what should we do with virtual chunks here?
```
emilk, line 639: maybe start off by at least doing a `re_log::debug_once` (or `warn_once` in debug-builds)?

emilk: here and elsewhere

emilk: could use another helper method for that

crates/store/re_chunk_store/src/gc.rs
teh-cmc: > The term "compacted chunks" is ambiguous: it could be read either as chunks that are the _source_ or the _output_ of compaction.
> 
> ("Those chunks have been compacted and are no longer needed" vs "those chunks are compacted and not what is on the server").

I'll try and clarify.

> Maybe we should introduce conceps like "source chunks" (chunks that exist in the source (e.g. redap)) vs "stored chunks" (chunks that could have been the output of compaction or splitting). Chunks that are neither compacted nor splitted would be _both_ source chunks and stored chunks. Or maybe we should always assign two IDs to stored chunks: the "source chunk id" and the "stored chunk id"? 🤔

Let's leave that discussion up for the lineage PRs.

crates/store/re_chunk_store/src/gc.rs
teh-cmc: Only compaction, split chunks never even make it into the store. I can add a comment.

crates/store/re_chunk_store/src/gc.rs
teh-cmc: I'll leave a note.

crates/store/re_chunk_store/src/query.rs
teh-cmc: I'd rather not introduce one of these cursed `testing` flags for a handful of unwraps tbh.

crates/store/re_chunk_store/src/writes.rs
```diff
@@ -386,7 +386,8 @@ impl ChunkStore {
                 // NOTE: The chunk that we've just added has been compacted already!
                 let srcs = std::iter::once((non_compacted_chunk.id(), non_compacted_chunk))
                     .chain(
-                        self.remove_chunks(vec![elected_chunk.clone()], None)
+                        // NOTE: deep removal, we don't want a GC'd chunk to linger on!
```
teh-cmc:
```suggestion
                        // NOTE: deep removal, we don't want a compacted chunk to linger on!
```

PR #12394, opened 2026-01-12
Implement welcome card analytics

lucasmerlin:
### Related

* closes https://linear.app/rerun/issue/RR-3299/add-analytics-to-new-cards

### What

Implements analytics for the welcome cards

### DIFF
--- a/crates/utils/re_analytics/src/event.rs
+++ b/crates/utils/re_analytics/src/event.rs
@@ -583,7 +583,7 @@ impl Properties for CliCommandInvoked {
 /// Tracks navigation clicks on the welcome screen cards.
 ///
 /// This event is sent when users click on cards on the welcome screen,
-/// such as documentation links or cloud-related CTAs.
+/// such as documentation links or cloud-related call-to-actions (CTAs).
 pub struct WelcomeScreenNavigation {
     /// Type of the card. E.g. "docs", "redap".
     pub card_type: String,


gavrelina:
there is no rerun_version tracked, could you add? 
also looks like all the links to docs are opened in another tab by default.. right? so having `new_tab` is pointless, lets remove

crates/utils/re_analytics/src/event.rs
```diff
@@ -580,6 +580,57 @@ impl Properties for CliCommandInvoked {
 
 // -----------------------------------------------
 
+/// Tracks navigation clicks on the welcome screen cards.
+///
+/// This event is sent when users click on cards on the welcome screen,
+/// such as documentation links or cloud-related CTAs.
```
emilk: what is a CTA?

crates/utils/re_analytics/src/event.rs
```diff
@@ -580,6 +580,57 @@ impl Properties for CliCommandInvoked {
...
+pub struct WelcomeScreenNavigation {
+    /// Whether the click was on the Cloud card (true) or a documentation card (false).
+    pub cloud_card: bool,
```
emilk: are these the only two types that are likely going to exist, or would it be better if this wad a card_category: String`?

crates/utils/re_analytics/src/event.rs
gavrelina: call to action, it is to indicate the main call to action of the button, instead of referring to exact wording that changes depending on your "situation" (logged in/have server/etc)

crates/utils/re_analytics/src/event.rs
gavrelina: good point! since one of the main reasons is to see if the commercial card is tracking any attention — this is a flag to split for that specifically. And as of now we don't know if any other types are gonna be there. 

Otherwise something like `card_type: docs` and `card_type: redap` could be more future proof?

lucasmerlin:
> there is no rerun_version tracked, could you add?

I thought it's enough that we track the version in the user properties. But I've now added it as a field in the new request as well.

crates/utils/re_analytics/src/event.rs
lucasmerlin: Changed it to string

crates/utils/re_analytics/src/event.rs
```diff
@@ -580,6 +580,53 @@ impl Properties for CliCommandInvoked {
 
 // -----------------------------------------------
 
+/// Tracks navigation clicks on the welcome screen cards.
+///
+/// This event is sent when users click on cards on the welcome screen,
+/// such as documentation links or cloud-related CTAs.
```
emilk: what is a CTA?
```suggestion
/// such as documentation links or cloud-related call-to-actions (CTAs).
```

crates/utils/re_analytics/src/event.rs
emilk: Ah, that makes much more sense than Chicago Transit Authority

emilk (approved):

PR #12396, opened 2026-01-12
Show loaded indicator when time panel is collapsed

IsseW:
### Related

- Closes RR-3295

### What

Also shows the time panel loaded indicator when timeline is collapsed.

Screenshot 2026-01-13 at 12 50 54

### DIFF
--- a/crates/viewer/re_time_panel/src/data_density_graph.rs
+++ b/crates/viewer/re_time_panel/src/data_density_graph.rs
@@ -452,7 +452,7 @@ pub fn paint_loaded_indicator_bar(
 
     let is_loading_at_current_time = time_ctrl.time_int().is_some_and(|time| {
         full_time_range.contains(time)
-            && loaded_ranges_on_timeline.iter().all(|r| !r.contains(time))
+            && !loaded_ranges_on_timeline.iter().any(|r| r.contains(time))
     });
 
     if is_loading_at_current_time


crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -441,41 +444,52 @@ pub fn paint_loaded_indicator_bar(
...
+                &[
+                    egui::pos2(
+                        x_range
+                            .intersection(Rangef::new(start as f32, end as f32))
+                            .min,
+                        y,
+                    ),
+                    egui::pos2(
+                        x_range
+                            .intersection(Rangef::new(start as f32, end as f32))
```
emilk: `x_range.intersection(Rangef::new(start as f32, end as f32))` is repeated three times.

Bind it once and reuse instead. I would suggest the name `x_range`, but that is already taken by… something? Maybe that parameter could be renamed something better (`full_x_range`?)

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -484,14 +498,32 @@ pub fn paint_loaded_indicator_bar(
 
             ui.painter()
                 // Need to clip because offsetting the dashed line may end up outside otherwise
-                .with_clip_rect(egui::Rect::from_x_y_ranges(range, Rangef::EVERYTHING))
+                .with_clip_rect(egui::Rect::from_x_y_ranges(
+                    x_range.intersection(Rangef::new(start as f32, end as f32)),
```
emilk: four times!

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -419,13 +419,16 @@ fn smooth(buckets: &[Bucket]) -> Vec<Bucket> {
...
+///
+/// `draw_ranges` indicates if the loaded ranges should be filled in.
 pub fn paint_loaded_indicator_bar(
     ui: &egui::Ui,
     time_ranges_ui: &TimeRangesUi,
     db: &re_entity_db::EntityDb,
     time_ctrl: &TimeControl,
     y: f32,
     x_range: Rangef,
+    draw_ranges: bool,
```
emilk: Maybe be more explicit and call this `paint_fully_loaded_ranges` or something?

(oh yeah and try to be consistent with naming - `paint`, and not `draw` or `render`)

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -441,41 +444,52 @@ pub fn paint_loaded_indicator_bar(
...
-            let end = time_ranges_ui.x_from_time(range.max.into())?;
-            debug_assert!(start <= end, "Negative x-range");
-            let x = Rangef::new(start as f32, end as f32).intersection(x_range);
-
-            if x.span() > 0.0 { Some(x) } else { None }
-        })
         .collect::<Vec<_>>();
 
+    let timeline_range = full_range;
+    let should_load = time_ctrl
```
emilk: `is_loading` is maybe a more accurate name?

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -441,41 +444,52 @@ pub fn paint_loaded_indicator_bar(
...
-            let start = time_ranges_ui.x_from_time(range.min.into())?;
-            let end = time_ranges_ui.x_from_time(range.max.into())?;
-            debug_assert!(start <= end, "Negative x-range");
-            let x = Rangef::new(start as f32, end as f32).intersection(x_range);
-
-            if x.span() > 0.0 { Some(x) } else { None }
-        })
         .collect::<Vec<_>>();
 
+    let timeline_range = full_range;
```
emilk: unused?

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -441,41 +444,52 @@ pub fn paint_loaded_indicator_bar(
...
+    let should_load = time_ctrl
+        .time_int()
+        .is_some_and(|time| full_range.contains(time))
+        && time_ctrl.time_int().is_some_and(|time| {
+            (&loaded_ranges_on_timeline)
+                .iter()
+                .all(|r| !r.contains(time))
+        });
```
emilk: This can be  a single `time_ctrl.time_int().is_some_and(|time|`

crates/viewer/re_time_panel/src/data_density_graph.rs
IsseW: ah yeah, think this was something that happened after automatically inlining a function… Fixed now

crates/viewer/re_time_panel/src/data_density_graph.rs
```diff
@@ -436,46 +440,37 @@ pub fn paint_loaded_indicator_bar(
...
-            if x.span() > 0.0 { Some(x) } else { None }
-        })
         .collect::<Vec<_>>();
 
-    if should_load
-        && let Some(start) = time_ranges_ui.x_from_time(full_range.min.into())
-        && let Some(end) = time_ranges_ui.x_from_time(full_range.max.into())
+    let is_loading_at_current_time = time_ctrl.time_int().is_some_and(|time| {
+        full_time_range.contains(time)
+            && loaded_ranges_on_timeline.iter().all(|r| !r.contains(time))
```
emilk: nit/style: "and not any contains" reads better to me than "and all does not contain"
```suggestion
            && !loaded_ranges_on_timeline.iter().any(|r| r.contains(time))
```

emilk (approved):

PR #12397, opened 2026-01-12
Improve add server ux

lucasmerlin:
### Related

* closes https://linear.app/rerun/issue/RR-3278/improve-add-a-server-ux

### What

Some small ux improvements when adding a server

### DIFF
--- a/crates/viewer/re_viewer/tests/snapshots/menu_without_recording.png
+++ b/crates/viewer/re_viewer/tests/snapshots/menu_without_recording.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:d78d5a30f3d56ebc894d68a6292bd01e65d57ec878dd8523cf5771b2e94b8319
-size 119994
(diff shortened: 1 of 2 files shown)


crates/viewer/re_recording_panel/src/recording_panel_ui.rs
```diff
@@ -52,9 +52,9 @@ impl RecordingPanel {
 
         ui.panel_content(|ui| {
             ui.panel_title_bar_with_buttons(
-                "Recordings",
+                "Sources",
                 Some(
-                    "These are the Recordings currently loaded in the Viewer, organized by application",
+                    "These are the Sources currently loaded in the Viewer, organized by application",
```
emilk: I think the tooltip should explain what we mean by "sources", and NOT capitalize it:

https://github.com/rerun-io/rerun/blob/5aeb0a6396c0ee69c1e318d0c18b8cd43134cbbc/DESIGN.md#L10-L13

emilk (approved):

PR #12398, opened 2026-01-12
Fix GC sometimes not collecting anything

emilk:
* Closes RR-3303

We had several bugs in our GC time budgeting

* Sometimes we spent the entire budget on prepatory work, leaving nothing for marking
* Sometimes we would spend the entire time budget on marking, and then have nothing left for sweeping

The result would be that sometimes the GC would spend a bunch of time only to find a single chunk as a candidate, and then not collect it, because the time budget was out.

### Changes
* During marking, we ignore setup-time when considering to break due to time budget constraints
* We always save at least half of the time budget for sweeping, so that the time spent marking is not wasted.


emilk:
Hopefully we can remove or optimize those upfront costs so that the time budgeting is useful/meaningful. Maybe it still is in some situations. I dunno

PR #12400, opened 2026-01-12
Virtual ChunkStore 3: initialize pure virtual store when receiving an RRD manifest

teh-cmc:
In addition to kicking off the pre-fetching machinery, we now also insert all the virtual chunks from that same manifest into the target store.

This doesn't achieve much on its own, but gets us closer to the pre-fetching vs. JIT partial results code paths colliding with one another.

---

* Previous: https://github.com/rerun-io/rerun/pull/12390

---

TODO:
* [x] Play with larger-than-RAM recordings and make sure nothing funky happens

### DIFF
--- a/crates/store/re_chunk_store/src/writes.rs
+++ b/crates/store/re_chunk_store/src/writes.rs
@@ -3,13 +3,11 @@ use std::sync::Arc;
 
 use ahash::HashMap;
 use arrow::array::Array as _;
-use arrow::datatypes::DataType;
 use itertools::Itertools as _;
 
 use re_byte_size::SizeBytes;
 use re_chunk::{Chunk, EntityPath, RowId};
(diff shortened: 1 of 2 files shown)


crates/store/re_chunk_store/src/writes.rs
```diff
@@ -16,6 +20,173 @@ use crate::{
...
+                    re_log::warn_once!(
+                        "Component '{}' on entity '{}' has unexpected non-list-array type: {}",
+                        descr.component,
+                        descr.entity_path,
+                        re_arrow_util::format_data_type(&dt),
+                    );
+                    dt
+                }
+            }
+        }
```
emilk: this could be a helper on `impl ComponentColumnDescriptor` 🤷

PR #12402, opened 2026-01-12
Rename `address` to `url` or `host` depending on context

emilk:
### Related
* Closes https://linear.app/rerun/issue/RR-3062/serveraddress-should-not-return-0000
* [Slack Discussion](https://rerunio.slack.com/archives/C045J1Z7DU7/p1768235441243609)

### What
* `rerun server --address …` is now `rerun server --host …`
* The `addr` argument of the `rr.Server` constructor is now called `host`
* The `address` function of `rr.Server` is now called `url`
* The `address` parameter of `rr.CatalogClient` is now called `url`

### DIFF
--- a/rerun_py/src/catalog/catalog_client.rs
+++ b/rerun_py/src/catalog/catalog_client.rs
@@ -46,8 +46,8 @@ impl PyCatalogClientInternal {
 
     /// Create a new catalog client object.
     #[new]
-    #[pyo3(text_signature = "(self, addr, token=None)")]
-    fn new(py: Python<'_>, addr: String, token: Option<String>) -> PyResult<Self> {
+    #[pyo3(text_signature = "(self, url, token=None)")]
+    fn new(py: Python<'_>, url: String, token: Option<String>) -> PyResult<Self> {
         // NOTE: The entire TLS stack expects this global variable to be set. It doesn't matter
         // what we set it to. But we have to set it, or we will crash at runtime, as soon as
         // anything tries to do anything TLS-related.
@@ -55,7 +55,7 @@ impl PyCatalogClientInternal {
         // but we removed that unused dependency, so now we must do it ourselves.
         _ = rustls::crypto::ring::default_provider().install_default();
 
-        let origin = addr.as_str().parse::<re_uri::Origin>().map_err(to_py_err)?;
+        let origin = url.as_str().parse::<re_uri::Origin>().map_err(to_py_err)?;
 
         let connection_registry =
             re_redap_client::ConnectionRegistry::new_with_stored_credentials();
--- a/rerun_py/src/recording/rrd.rs
+++ b/rerun_py/src/recording/rrd.rs
@@ -88,7 +88,7 @@ impl PyRecording {
 ///
 /// Parameters
 /// ----------
-/// path_to_rrd : str | os.PathLike[str]
+/// path_to_rrd:
 ///     The path to the file to load.
 ///
 /// Returns
@@ -120,7 +120,7 @@ pub fn load_recording(path_to_rrd: std::path::PathBuf) -> PyResult<PyRecording> 
 ///
 /// Parameters
 /// ----------
-/// path_to_rrd : str | os.PathLike[str]
+/// path_to_rrd:
 ///     The path to the file to load.
 ///
 /// Returns


rerun_py/tests/api_sandbox/rerun_draft/server.py
```diff
@@ -12,20 +12,20 @@
...
         self,
         *,
-        address: str = "0.0.0.0",
+        bind_ip: str = "0.0.0.0",
         port: int | None = None,
         datasets: dict[str, PathLike[str]] | None = None,
         tables: dict[str, PathLike[str]] | None = None,
     ) -> None:
         self._internal = _server.Server(
-            address=address,
```
emilk, line 28: I'm 99% sure there was a bug in the old code here

crates/store/re_server/src/entrypoint.rs
```diff
@@ -15,9 +16,9 @@ use crate::{ServerBuilder, ServerHandle};
 #[derive(Clone, Debug, clap::Parser)]
 #[clap(author, version, about)]
 pub struct Args {
-    /// Address to listen on.
+    /// IP address to listen on.
     #[clap(long, default_value = "0.0.0.0")]
-    pub addr: String,
+    pub ip: String,
```
abey79: this need to be covered in the migration guide as it might break existing user scripts

rerun_py/rerun_sdk/rerun/server.py
```diff
@@ -69,9 +73,19 @@ def __init__(
...
         """
 
+        if bind_ip == "0.0.0.0" and addr != "0.0.0.0":
+            bind_ip = addr
+            _send_warning_or_raise(
+                "The `addr` parameter is deprecated in Rerun 0.29, and has been renamed to `bind_ip`.",
+                depth_to_user_code=2,
+                warning_type=DeprecationWarning,
+            )
+
```
abey79: I think you could use `@deprecated_param` for this (grep for usage example, we have it somewhere else). Also, needs mention in migration guide.

rerun_py/rerun_sdk/rerun/server.py
```diff
@@ -69,9 +73,19 @@ def __init__(
         tables:
             Optional dictionary mapping table names to lance file paths which will be loaded and made available when the
             server starts.
+        addr:
+            DEPRECATED: Renamed to `bind_ip`
```
abey79: I'd remove this entirely.

rerun_py/rerun_sdk/rerun/server.py
```diff
@@ -102,24 +116,25 @@ def __init__(
...
+        return self._internal.url()
 
-        Returns
-        -------
-        str
-            The address of the server.
+    @deprecated("Renamed to `url`.")
+    def address(self) -> str:
+        """Get the URL of the server to which clients can connect."""
+        return self._internal.url()
```
abey79: nit

```suggestion
        return self.url()
```

rerun_py/src/server.rs
```diff
@@ -33,34 +36,54 @@ impl PyServerInternal {
...
+    pub fn url(&self) -> String {
+        self.url.clone()
+    }
+
+    /// The address of the server to which clients can connect.
+    #[deprecated(note = "Renamed to `url`")]
     pub fn address(&self) -> String {
-        format!("rerun+http://{}", self.address)
+        self.url.clone()
+    }
```
abey79: we don't need to keep that around, this is an internal API

nikolausWest:
How are you making sure this doesn't break existing code / handling breaking it?

rerun_py/tests/api_sandbox/rerun_draft/server.py
abey79: why is that?

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
```diff
@@ -59,7 +59,7 @@ class CatalogClient:
 
     __slots__ = ("_internal",)
 
-    def __init__(self, address: str, token: str | None = None) -> None:
+    def __init__(self, url: str, *, token: str | None = None) -> None:
```
abey79: This is a breaking change that will affect many users who use `address`. And I see no reason to force non-kw arg here tbh.

We need to roll deprecation correctly:

```suggestion
	@deprecated_param("address", use_instead="url", since="0.29")
    def __init__(self, url: str, token: str | None = None, address: str | None = None) -> None:
```

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
```diff
@@ -21,8 +21,8 @@
 class CatalogClient:
     """Client for a remote Rerun catalog server."""
 
-    def __init__(self, address: str, token: str | None = None) -> None:
-        self._inner = _catalog.CatalogClient(address, token)
+    def __init__(self, url: str, *, token: str | None = None) -> None:
```
abey79, line 24:
```suggestion
    def __init__(self, url: str, token: str | None = None) -> None:
```

abey79 (changes requested):
Generally LGTM, but we need to be more thorough with deprecation, including properly deprecating the api change around `CatalogClient` and having a note in the migration guide.

rerun_py/rerun_sdk/rerun/server.py
```diff
@@ -57,8 +61,8 @@ def __init__(
 
         Parameters
         ----------
-        address:
-            The address to bind the server to.
+        bind_ip:
+            The IP address to bind the server to.
```
grtlr: If we want to allow (special) DNS names like `localhost` in the future, we might want to use `host` here.

rerun_py/rerun_sdk/rerun/server.py
emilk: Then how would users know it is deprecated?
Also, I'm pretty sure we have a lint that all args must be documented (we should have)

rerun_py/rerun_sdk/rerun/server.py
emilk: I tired it, but it `@deprecated_param` doesn't seem work for `__init__` methods

rerun_py/tests/api_sandbox/rerun_draft/server.py
emilk: It called `Server.address` (which was an URL) and passed it to `Server::__init__` (which expects an IP)

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
emilk: > And I see no reason to force non-kw arg here tbh.

"Force non-kw"? My change forces kw-args for `token`, which I think is good: it makes the calling code a lot more explicit. Right now this compiles: `CatalogClient(my_token, my_url)` but is obviously buggy. Your suggested code would even accept `CatalogClient("foo", "bar", "baz")`

rerun_py/tests/api_sandbox/rerun_draft/catalog.py
emilk: See above

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
emilk: kw-args are also a lot easier to migrate in the future, which is why we prefer them in most of our APIs

CODE_STYLE.md
```diff
@@ -153,6 +153,21 @@ println!("{:-}, value"); // The `-` option stands for redaction.
 
 Look for `f.sign_minus()` in the code for where we handle this.
 
+## Python
+Prefer kw-args (key-word arguments) for non-obvious parameters.
```
ntjohnson1: NIT: non-obvious and infrequent parameters.

There are a few places where I think we are unpythonic in forcing always using kwargs. Maybe with agents and people writing less code it matters less. But if there is only a single parameter that users often have to specify then I think allowing that to be positional is much nicer.

If I could call serve() and all other arguments were optional with reasonable defaults then I think all kwargs is fine but putting the single required parameter as kwarg only doesn't align with my intuition for common practice.

emilk: I agree, and will clarify.

rerun_py/rerun_sdk/rerun/catalog/_catalog_client.py
abey79: ~Sure, but kw-args introduce friction for the users when used excessively. We recently had a case where we backtracked on this because of how annoying it was (`get_dataset(name=...)`/`get_table(name=...)`).~

nvm, coffee still needs to happen

rerun_py/rerun_sdk/rerun/server.py
abey79: huh, ok

rerun_py/tests/api_sandbox/rerun_draft/server.py
abey79: I'm not seeing where `Server.address` is called.

emilk: image

abey79: this is just binding `rerun_draft.server.Server.address` to the same method as `rerun.server.Server.address`. Nothing is called or anything.

PR #12404, opened 2026-01-12
Virtual ChunkStore 4: prefetch last frame's missing chunks in priority

teh-cmc:
This implements the first real bridge across the prefetch world and the JIT world.

The store now automatically tracks all the chunks that were missed while computing query results.
These "missing chunks" are then exposed via a poll-based API that viewer space can call once per frame in order to prioritize thos chunks during the next round of prefetching.

---

* Previous: https://github.com/rerun-io/rerun/pull/12400

---

TODO:
* [x] Play with larger-than-RAM recordings and make sure nothing funky happens

### DIFF
--- a/crates/store/re_entity_db/src/rrd_manifest_index.rs
+++ b/crates/store/re_entity_db/src/rrd_manifest_index.rs
@@ -432,12 +432,13 @@ impl RrdManifestIndex {
         let mut indices = vec![];
 
         let missing_chunk_ids = missing_chunk_ids.into_iter();
-        let prefetched_chunk_ids = chunks
-            .query(desired_range.into())
-            .map(|(_, chunk_id)| *chunk_id);
-
+        let prefetched_chunk_ids = || {
+            chunks
+                .query(desired_range.into())
+                .map(|(_, chunk_id)| *chunk_id)
+        };
         let chunk_ids_in_priority_order =
-            missing_chunk_ids.chain(std::iter::once_with(|| prefetched_chunk_ids).flatten());
+            missing_chunk_ids.chain(std::iter::once_with(prefetched_chunk_ids).flatten());
 
         // We might reach our budget limits before we enqueue all `missing_chunk_ids`.
         // That's fine: they will still be missing next frame, and therefore will still be reported


crates/store/re_chunk_store/src/store.rs
```diff
@@ -728,6 +743,19 @@ impl ChunkStore {
...
+    /// Returns the set of [`ChunkId`]s that were detected as missing since the last time since method was called.
+    ///
+    /// Chunks are considered missing when they are required to compute the results of a query, but cannot be
+    /// found in local memory.
+    ///
+    /// Calling this method is destructive: the internal set is cleared on every call, and will grow back as
+    /// new queries are run.
+    /// Callers are expected to call this once per frame in order to know which chunks were missing during
+    /// the previous frame.
+    pub fn gather_missing_chunk_ids(&self) -> HashSet<ChunkId> {
```
emilk: nit/naming

```suggestion
    pub fn take_missing_chunk_ids(&self) -> HashSet<ChunkId> {
```

emilk (approved):
nice!

crates/store/re_entity_db/src/rrd_manifest_index.rs
```diff
@@ -424,25 +431,38 @@ impl RrdManifestIndex {
...
 
-        for (_, chunk_id) in chunks.query(desired_range.into()) {
-            let Some(remote_chunk) = self.remote_chunks.get_mut(chunk_id) else {
+        let missing_chunk_ids = missing_chunk_ids.into_iter();
+        let prefetched_chunk_ids = chunks
+            .query(desired_range.into())
+            .map(|(_, chunk_id)| *chunk_id);
+
+        let chunk_ids_in_priority_order =
+            missing_chunk_ids.chain(std::iter::once_with(|| prefetched_chunk_ids).flatten());
```
teh-cmc: i've put the laziness in the wrong place, this achieves nothing as currently written :melting_face: 
```suggestion
        let prefetched_chunk_ids = || {
            chunks
                .query(desired_range.into())
                .map(|(_, chunk_id)| *chunk_id)
            };

        let chunk_ids_in_priority_order =
            missing_chunk_ids.chain(std::iter::once_with(prefetched_chunk_ids).flatten());
```

crates/store/re_chunk_store/src/store.rs
teh-cmc: i like it

PR #12409, opened 2026-01-12
Fix snippet compare

ntjohnson1:
### Related
Failing CI on main

### What
I don't actually understand how this overlaps with out snippet compare script. They look to both do quite similar things? The uv transition seems to have highlighted this or maybe I just needed the full check on my last PR and doing the snippet compare wasn't enough.

If we aren't doing a backwards comparison we shouldn't require an rrd gets generated. This is a bandaid to unblock doing cloud docs. We probably need a broader updated approach to make all the snippet/examples less rrd centric.


ntjohnson1:
The failing test from main works for me locally but just in case

@rerun-bot full-check

PR #12415, opened 2026-01-13
Fix axes visualizer complaining about entities with frame-based transforms

MichaelGrupp:
Closes: [RR-3311](https://linear.app/rerun/issue/RR-3311)

Instead of trying to visualize _always_ an implicit frame ID and then _additionally_ the potential explicit transform frames of that entity, do either:

- visualize _only_ explicit frames' axes if they exist
- visualize _only_ the implicit frame axes otherwise

This fixes an issue where the visualizer complains if you have an entity containing transforms with frame names (e.g. `/transforms/`) and then fails to find the connection from the implicit frame `tf#/transforms` to the target, although this implicit frame is irrelevant in this case because we are interested in the explicit frames.

Follow up to: https://github.com/rerun-io/rerun/pull/12262

### DIFF
--- a/crates/viewer/re_view_spatial/tests/snapshots/transform_axes_for_explicit_transforms.png
+++ b/crates/viewer/re_view_spatial/tests/snapshots/transform_axes_for_explicit_transforms.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:088e1b443eb58311fddfdd610996c996b78a23e38fe50c38b95a0d1a66659b9b
-size 17089
+oid sha256:9ca222b3a3373d1050f443b37bfbdd7a65669c5d428f8909cef8bb0b58010ce0
+size 16576


crates/viewer/re_view_spatial/tests/snapshots/2d_in_3d_with_explicit_frames_at_origin.png
MichaelGrupp, line 1: Could this be the solution to the mystery we observed in this PR? https://github.com/rerun-io/rerun/pull/12276/changes#r2635589941

MichaelGrupp:
Examples with implicit frames or explicit frames both work (e.g. animated_urdf and NuScenes in the web_viewer built here).

image

Bildschirmfoto 2026-01-13 um 13 54 39

crates/viewer/re_view_spatial/tests/snapshots/transform_axes_for_explicit_transforms.png
MichaelGrupp, line 1: This snapshot highlight another issue we have: the root frame ID is not visualized (we only iterate over children in the visualizer?). Here in this test, it looks like `tf#/all_the_transforms` was shown "accidentally" because because it's also the implicit frame ID and there was no conflict?

In the other more realistic examples, the root of the TF tree doesn't show up (independent of this patch here). E.g. in the animated_urdf example, the "base" axes is not visible in 0.28.

I think this should be investigated separately?

Wumpf:
Superseeded by https://github.com/rerun-io/rerun/pull/12419

PR #12416, opened 2026-01-13
Add `--latency-ms` option to OSS server

emilk:
Useful for testing

### Related
* https://linear.app/rerun/issue/RR-3297/show-redap-latency-and-throughput
* https://linear.app/rerun/issue/RR-3294/test-ltr-in-high-latency-low-bandwidth-situations

### DIFF
--- a/crates/store/re_server/src/latency_layer.rs
+++ b/crates/store/re_server/src/latency_layer.rs
@@ -55,7 +55,9 @@ where
         let Self { mut inner, rtt } = self.clone();
         Box::pin(async move {
             let resp = inner.call(req).await;
-            sleep(rtt).await;
+            if rtt != Duration::ZERO {
+                sleep(rtt).await;
+            }
             resp
         })
     }


crates/store/re_server/src/latency_layer.rs
```diff
@@ -0,0 +1,62 @@
...
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request) -> Self::Future {
+        let Self { mut inner, rtt } = self.clone();
+        Box::pin(async move {
+            let resp = inner.call(req).await;
+            sleep(rtt).await;
```
IsseW: Might be worth having this in an `if rtt > 0` since it happens on every response, from what I can tell a zero duration isn't special cased for tokio's sleep. https://users.rust-lang.org/t/does-awaiting-tokio-sleep-duration-zero-yield-back-to-the-runtime-or-poll-ready-immediately/116756/2

emilk: Good catch!

PR #12419, opened 2026-01-13
Don't show error for implicit transform axes if we're already showing named transforms

Wumpf:
### Related

* Alternative to https://github.com/rerun-io/rerun/pull/12415
* Fixes RR-3311

### What

When adding a `TransformAxes` to an entity with named `Transform3D` we correctly get to see all the axes.
However, if that entity didn't have `CoordinateFrames` and the implicit frame at that entity wasn't connected up, we'd show an error. Since in this case we're almost certainly not interested in the local implicit coordinate frame, we're now omitting this error.

image

### DIFF
--- a/crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
+++ b/crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
@@ -72,6 +72,7 @@ impl VisualizerSystem for TransformAxes3DVisualizer {
             let entity_path = &data_result.entity_path;
 
             // Draw all transforms defined _at_ this entity.
+            // TODO(RR-3319): consider also root frames here (not only child frames).
             let mut transforms_to_draw: smallvec::SmallVec<[_; 1]> = transforms
                 .child_frames_for_entity(entity_path.hash())
                 .map(|(frame_id_hash, transform)| {


grtlr (approved):
Makes sense, thank you for the quick fix!

Yeah, let's get this into a potential patch release.

crates/viewer/re_view_spatial/src/visualizers/transform_axes_3d.rs
```diff
@@ -69,48 +69,59 @@ impl VisualizerSystem for TransformAxes3DVisualizer {
         );
 
         for data_result in query.iter_visible_data_results(Self::identifier()) {
-            // We first add the axes for the entity transforms, because we want them to be drawn below
+            let entity_path = &data_result.entity_path;
+
+            // Draw all transforms defined _at_ this entity.
```
MichaelGrupp, line 74: Maybe you can mention the other ticket I created here, e.g.:

```suggestion
            // Draw all transforms defined _at_ this entity.
            // TODO(RR-3319): consider also root frames here (not only child frames).
```

MichaelGrupp (approved):
Looks good to me and doesn't change tests, so this one wins ;) One minor note.

PR #12422, opened 2026-01-13
Virtual ChunkStore 6: support virtual chunks in latest-at query cache

teh-cmc:
This makes the latest-at query cache ready for virtual chunks.

Contrary to the range cache, latest-at cache semantics are fairly complex, and even more so in a lazy loaded environment where `Clear` tombstones might be known to exist but not necessarily loaded in local memory at query time.
Most of the complexity of this PR comes from checking those ugly cases and getting them right. Other than that, it's business as usual.

---

> [!NOTE]
> 
> The focus here is only on making the cache itself (and cache invalidation especially), and its _direct_ callers, aware of partially missing data. There is and will be an infinitely long tail of propagating missing-ness information through the million little helpers and result classes that we have all over the viewer codebase, but that's out of the scope of this PR.
>
> Right now we only want 2 things: caches that behave properly in the face of partially missing data, and a store that properly reports missing chunks to the (pre)fetcher.

---

* Previous: https://github.com/rerun-io/rerun/pull/12420

### DIFF
--- a/crates/store/re_chunk_store/src/gc.rs
+++ b/crates/store/re_chunk_store/src/gc.rs
@@ -2,6 +2,7 @@ use std::collections::BTreeSet;
 use std::sync::Arc;
 use std::time::Duration;
 
+use itertools::Itertools as _;
 use nohash_hasher::IntMap;
 use web_time::Instant;
 
@@ -434,10 +435,9 @@ impl ChunkStore {
 
         // Make sure to not forward those diffs as-is: just because the shallow deletion yielded
         // nothing, doesn't mean that a deep one won't.
-        {
-            let diffs = self.remove_chunks_shallow(chunks_to_be_removed.clone(), time_budget);
-            debug_assert!(diffs.len() <= 1);
-        }
+        // The deep diff is always a superset of the shallow one (you can remove chunk physical
+        // chunks while keeping virtual ones, but not vice-versa).
+        let diffs_shallow = self.remove_chunks_shallow(chunks_to_be_removed.clone(), time_budget);
 
         let Self {
             id: _,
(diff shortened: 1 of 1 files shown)


crates/store/re_query/src/latest_at.rs
```diff
@@ -77,6 +77,9 @@ impl QueryCache {
         // the data.
         let mut max_clear_index = (TimeInt::MIN, RowId::ZERO);
         {
+            // TODO(cmc): this will currently fail with stores bootstrapped directly from an RRD manifest
+            // (`ChunkStore::insert_rrd_manifest`), since virtual insertions don't trigger any kind of store
+            // events yet.
```
teh-cmc, line 82: That will be the subject of my next follow up.

crates/store/re_chunk_store/src/gc.rs
```diff
@@ -432,8 +432,12 @@ impl ChunkStore {
     ) -> Vec<ChunkStoreDiff> {
         re_tracing::profile_function!();
 
-        let diffs = self.remove_chunks_shallow(chunks_to_be_removed, time_budget);
-        debug_assert!(diffs.len() <= 1);
+        // Make sure to not forward those diffs as-is: just because the shallow deletion yielded
+        // nothing, doesn't mean that a deep one won't.
+        {
+            let diffs = self.remove_chunks_shallow(chunks_to_be_removed.clone(), time_budget);
+            debug_assert!(diffs.len() <= 1);
```
emilk: But should the diffs be completely ignored like this? I guess the deep code below will generate the same diff again? Or a different/better one?

teh-cmc: The deep diff is always a strict superset of the shallow diff. I'll add a debug assertion to that effect.

teh-cmc: also why is that `debug_assert!(diffs.len() <= 1);` assertion still there, that doesn't make sense anymore

PR #12427, opened 2026-01-13
Check for uv lock being up-to-date

jleibs:
### Related

We spent a bunch of time debugging cryptic failures trying to publish docs in release.  Turns out this is something that happens in some edge cases when the uv.lock is out of sync.

This adds checks for that condition and also updates the locks when bumping the version info during release.

### DIFF
--- a/rerun_notebook/pyproject.toml
+++ b/rerun_notebook/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "hatchling.build"
 [project]
 name = "rerun-notebook"
 description = "Implementation helper for running rerun-sdk in notebooks"
-version = "0.29.0-alpha.2+dev"
+version = "0.29.0-alpha.1+dev"
 dependencies = [
   "anywidget",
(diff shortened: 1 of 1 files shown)


rerun_notebook/pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = "hatchling.build"
 [project]
 name = "rerun-notebook"
 description = "Implementation helper for running rerun-sdk in notebooks"
-version = "0.29.0-alpha.1+dev"
+version = "0.29.0-alpha.2+dev"
```
emilk: unintentional?

jleibs: intentional attempt to trigger the check failure

jleibs: This is what I wanted to see:
https://github.com/rerun-io/rerun/actions/runs/20969696281/job/60269516120?pr=12427

PR #12429, opened 2026-01-13
Display downloaded & max size of recording in selection panel

IsseW:
### Related

- Part of RR-3297

### What

Display downloaded & max size of recording in selection panel

Screenshot 2026-01-13 at 21 44 44
Screenshot 2026-01-13 at 21 45 45


crates/viewer/re_data_ui/src/entity_db.rs
```diff
@@ -118,13 +118,48 @@ impl crate::DataUi for EntityDb {
 
             {
                 ui.grid_left_hand_label("Size");
-                ui.label(re_format::format_bytes(self.total_size_bytes() as _))
+
+                let total_size_bytes = self.total_size_bytes();
+                let static_size = if self.rrd_manifest_index().has_manifest() {
+                    self.rrd_manifest_index().full_uncompressed_size().max(total_size_bytes)
```
lucasmerlin: Egui has a ext trait that adds a `.at_least`/`.at_most` function, which to me is way more readable than `.max`/`.min`.

lucasmerlin (approved):
Nice!

emilk:
image

I find this UI confusing. The size is 20 GB, but Rerun is just going to download 18 GB of it?
@gavrelina

IsseW:
> The size is 20 GB, but Rerun is just going to download 18 GB of it?

Yes because of the memory limit, should it be more clear that it's 18 GB at a time and not just 18 GB of the recording?

PR #12432, opened 2026-01-14
Help out `+nightly` type checker by specifying trait bounds

grtlr:
### What

Rerun does not currently build with `+nightly` due to type inference hitting the stack limit. We help out the poor type checker by pruning some of the deep paths.


Wumpf (approved):
weird, but sure 🤷

grtlr:
Unrelated failures 🤷

PR #12434, opened 2026-01-14
Load chunks in batches

emilk:
### Related
* https://linear.app/rerun/issue/RR-3322/latency-spikes-towards-rerun-cloud

### What
Previously, we would load all chunks of a store in one big request. This caused a big "blockage" with symptoms being a latency spike, and long time until first data arrived.

This is an attempt to address this.

### Results
Previously we would get everything in one or two big batches: [main behavior](https://rerun.io/viewer/version/main?url=rerun%3A%2F%2Fapi.latest.cloud.rerun.io%3A443%2Fdataset%2F1888C7A31503A35E35e832070692f160%3Fsegment_id%3DILIAD_50aee79f_2023_07_12_20h_28m_36s)

Now, data instead streams in: [PR behavior](https://rerun.io/viewer/pr/12434?url=rerun%3A%2F%2Fapi.latest.cloud.rerun.io%3A443%2Fdataset%2F1888C7A31503A35E35e832070692f160%3Fsegment_id%3DILIAD_50aee79f_2023_07_12_20h_28m_36s)

### DIFF
--- a/crates/store/re_redap_client/src/grpc.rs
+++ b/crates/store/re_redap_client/src/grpc.rs
@@ -611,7 +611,7 @@ fn chunk_id_column(batch: &RecordBatch) -> Option<&[ChunkId]> {
 
 /// Load only static chunks
 async fn load_static_chunks(
-    client: &mut ConnectionClient,
+    client: &ConnectionClient,
     tx: &re_log_channel::LogSender,
     store_id: &StoreId,
     rrd_manifest: re_log_encoding::RrdManifest,


crates/store/re_redap_client/src/grpc.rs
```diff
@@ -651,20 +651,87 @@ async fn load_static_chunks(
...
+        let end = usize::min(start + BATCH_SIZE, num_rows);
+        let small_batch = full_batch.slice(start, end - start);
+
+        let mut client = client.clone();
+        let tx = tx.clone();
+        let store_id = store_id.clone();
+
+        batch_handles.push(tokio::spawn(async move {
+            load_small_chunk_batch(&mut client, &tx, &store_id, &small_batch).await
+        }));
```
teh-cmc: Is there any particular reason why you're spawning tasks instead of relying on future-level concurrency here?

crates/store/re_redap_client/src/grpc.rs
```diff
@@ -651,20 +651,87 @@ async fn load_static_chunks(
...
+    // Batch requests in groups of N=32 rows.
+    const BATCH_SIZE: usize = 32;
+    let num_rows = full_batch.num_rows();
+
+    for start in (0..num_rows).step_by(BATCH_SIZE) {
+        let end = usize::min(start + BATCH_SIZE, num_rows);
+        let small_batch = full_batch.slice(start, end - start);
+        if load_small_chunk_batch(client, &tx, &store_id, &small_batch)
+            .await?
+            .is_break()
```
teh-cmc: Opposite question here: is there any particular reason why you're not leveraging future-level concurrency here?

teh-cmc (commented):
I believe both implementations can be merged into a single one by leveraging future-level concurrency. This should make the native one lighter and the web one faster in the process.

crates/store/re_redap_client/src/grpc.rs
```diff
@@ -651,20 +651,87 @@ async fn load_static_chunks(
...
+    Ok(ControlFlow::Continue(()))
+}
+
+async fn load_small_chunk_batch(
+    client: &mut ConnectionClient,
+    tx: &re_log_channel::LogSender,
+    store_id: &StoreId,
+    batch: &RecordBatch,
+) -> ApiResult<ControlFlow<()>> {
+    let chunk_stream = client.fetch_segment_chunks_by_id(batch).await?;
```
teh-cmc, line 701:
```suggestion
    // TODO(RR-3323): FetchChunks should expose a proper bidirectional streaming path on native.
    let chunk_stream = client.fetch_segment_chunks_by_id(batch).await?;
```

crates/store/re_redap_client/src/grpc.rs
emilk: What do you mean by future-level concurrency?

teh-cmc: In this instance, it would pretty much just be a matter of removing the `spawn` so that we collect the raw futures instead:
```suggestion
        batch_handles.push(async move {
            load_small_chunk_batch(&mut client, &tx, &store_id, &small_batch).await
        });
```
and then if you put that collection of pending futures into e.g. a [`FuturesUnordered`](https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html) set, you can poll all of them concurrently without having to spawn anything.

Another alternative is to turn the iterator of batches into a `Stream` and use something like [`Stream::buffer_unordered`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffer_unordered) instead. That will have the exact same outcome. We tend to prefer that on the server-side, but in this case both are fine imo.

PR #12441, opened 2026-01-14
Send store events when we add prefetched chunks

IsseW:
### What

The path for prefetched chunks was just discarding store events. This will send them forward.

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -2396,6 +2396,7 @@ impl App {
                     let entity_db = store_hub.entity_db_mut(&store_id);
                     entity_db.add_rrd_manifest_message(*rrd_manifest);
 
+                    // TODO(RR-3329): Should this run for all caches?
                     if let Some(caches) = store_hub.active_caches() {
                         // Downgrade to read-only, so we can access caches.
                         let entity_db = store_hub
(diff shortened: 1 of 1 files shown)


crates/viewer/re_viewer/src/app.rs
```diff
@@ -3082,15 +3082,39 @@ impl App {
     fn prefetch_chunks(&self, store_hub: &mut StoreHub) {
```
emilk, line 3101:
```suggestion
    ///
    /// There is logic duplicated between this and [`Self::receive_messages`].
    /// Make sure they are kept in sync!
    fn prefetch_chunks(&self, store_hub: &mut StoreHub) {
```

And then add the same to `fn receive_messages`

crates/viewer/re_viewer/src/app.rs
```diff
@@ -3082,15 +3082,39 @@ impl App {
...
                 }
 
+                // Downgrade to read-only, so we can access caches.
+                let db = store_hub
+                    .entity_db(&store_id)
+                    .expect("Just queried it mutable and that was fine.");
+
+                if let Some(caches) = store_hub.active_caches() {
+                    caches.on_store_events(&store_events, db);
+                }
```
emilk: I think we should update _all_ caches, not just the active one. Let's leave a TODO for this: 

https://linear.app/rerun/issue/RR-3329/transform-hierarchy-broken-when-loading-in-background

crates/viewer/re_viewer/src/app.rs
```diff
@@ -3082,15 +3082,39 @@ impl App {
...
+
+                self.validate_loaded_events(&store_events);
```
```suggestion
                self.validate_loaded_events(&store_events);
                
                // Note: some of the logic above is duplicated in `fn receive_messages`.
                // Make sure they are kept in sync!
```

And add the same to `fn receive_messages`.

…or figure out a way to deduplicate the code

PR #12442, opened 2026-01-14
Virtual ChunkStore 7: virtual store events & support for purely virtual tombstones in the cache

teh-cmc:
This A) introduces the notion of virtual store events, which are kicked off anytime `ChunkStore::insert_rrd_manifest` is called, and B) makes the latest-at query cache able to reason about tombstones that were never ever physically loaded by subscribing to these events.
These virtual events will likely be useful beyond the scope of the query cache in the future.

---

* Previous: https://github.com/rerun-io/rerun/pull/12422

### DIFF
--- a/crates/store/re_chunk_store/src/events.rs
+++ b/crates/store/re_chunk_store/src/events.rs
@@ -128,6 +128,8 @@ pub struct ChunkStoreDiff {
     /// Still, even though no physical data was ingested yet, it might be important for downstream
     /// consumers to know what kind of virtual data is being referenced from now.
     /// For example, query caches must know about pending tombstones as soon as they start being referenced.
+    ///
+    /// If this is set, all fields below are irrelevant (set to their default/empty values).
     pub rrd_manifest: Option<Arc<RrdManifest>>,
 
     /// The chunk that was added or removed.
@@ -199,7 +201,6 @@ impl ChunkStoreDiff {
         Self {
             kind: ChunkStoreDiffKind::Addition,
             rrd_manifest: Some(rrd_manifest),
-            // TODO: we need an enum for the 2 different payloads
             chunk: Arc::new(Chunk::empty(ChunkId::ZERO, "<empty>".into())),
             split_source: None,
             compacted: None,


crates/store/re_chunk_store/src/events.rs
```diff
@@ -121,6 +122,16 @@ pub struct ChunkStoreDiff {
...
     pub kind: ChunkStoreDiffKind,
 
+    /// A new [`RrdManifest`] was inserted into the store.
+    ///
+    /// This is very different from the usual chunk-related events as this is purely virtual.
+    /// Still, even though no physical data was ingested yet, it might be important for downstream
+    /// consumers to know what kind of virtual data is being referenced from now.
+    /// For example, query caches must know about pending tombstones as soon as they start being referenced.
+    ///
+    /// If this is set, all fields below are irrelevant (set to their default/empty values).
```
teh-cmc, line 132: I made the possibly controversial decision of taking a "uber-struct" approach here, as opposed to introducing a whole new layer with e.g. a tagged union.

I believe this a much more sane approach in this instance, rather than performing countless refactorings in all the downstream consumers all over the codebase.
Not only right now at this instant, but also for future changes: events like these just don't land themselves well to rigid tagged unions, which lead to massive amounts of code churn with no particular upside if the data within can instead just as well been rendered idempotent (like is the case here, using an empty chunk).

crates/store/re_query/src/cache.rs
```diff
@@ -304,13 +304,84 @@ impl ChunkStoreSubscriber for QueryCache {
...
+                                    .static_
+                                    .entry((entity_path.clone(), component))
+                                    .or_default()
+                                    .insert(chunk_id);
+                            }
+                        }
+                    }
+
+                    Err(err) => {
+                        re_log::error!(%err, ?store_id, "static data from RRD manifest couldn't be parsed");
```
emilk, line 335: I would really prefer if we parsed it _once_ on ingestion instead, so we wouldn't need this error handling everywhere in the code (i.e. in the constructor of `RrdManifest`). It's been painful so far, and is not getting less so

teh-cmc: Ye, agreed. I also would like to hash the manifest so that I can have some kind of unique ID, and I definitely want that memoized. I'll do a quick investigation of how we can improve things on that front.

PR #12443, opened 2026-01-14
Support enqueing video gop even if the whole gop isn't loaded

IsseW:
### Related

- Followup from https://github.com/rerun-io/rerun/pull/12277.

### What

Before this we would error if video couldn't enqueue a full gop, which could be an issue if a gop is split into multiple chunks.

### DIFF
--- a/crates/viewer/re_viewer_context/src/cache/video_stream_cache/test_player.rs
+++ b/crates/viewer/re_viewer_context/src/cache/video_stream_cache/test_player.rs
@@ -85,7 +85,7 @@ impl TestVideoPlayer {
     }
 
     #[track_caller]
-    fn expect_encoded_samples(&self, samples: impl IntoIterator<Item = SampleIndex>) {
+    fn expect_decoded_samples(&self, samples: impl IntoIterator<Item = SampleIndex>) {
         let received = self.sample_rx.try_iter().collect::<Vec<_>>();
         let expected = samples.into_iter().collect::<Vec<_>>();
 
@@ -259,19 +259,19 @@ fn create_video(
 fn test_simple_video(mut video: TestVideoPlayer, count: usize, dt: f64, max_time: f64) {
     video.play(0.0..max_time, dt).unwrap();
 
-    video.expect_encoded_samples(0..count);
+    video.expect_decoded_samples(0..count);
 
     // try again at 0.5x speed.
 
     video.play(0.0..max_time, dt * 0.5).unwrap();
 
-    video.expect_encoded_samples(0..count);
+    video.expect_decoded_samples(0..count);
 
     // and at 2x speed.
 
     video.play(0.0..max_time, dt * 2.0).unwrap();
 
-    video.expect_encoded_samples(0..count);
+    video.expect_decoded_samples(0..count);
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -378,7 +379,18 @@ impl VideoPlayer {
             }
 
             // Nothing more to enqueue / reached end of video?
-            if last_enqueued + 1 == video_description.samples.next_index() {
+            if video_description
+                .samples
+                .get(last_enqueued + 1)
+                .is_none_or(|sample| matches!(sample, re_video::SampleMetadataState::Unloaded(_)))
+            {
+                // If this is a sample we expect to have return an error.
```
Wumpf: grammar no sense to me

crates/viewer/re_renderer/src/video/player.rs
```diff
@@ -378,7 +379,18 @@ impl VideoPlayer {
...
+                if last_enqueued < requested_sample_idx {
```
Wumpf: the `last_enqueued >= requested_sample_id` would mean that we just did a jump backwards, right?
Maybe worth pointing out. If what I wrote is wrong it's extra worth pointing out 😄

Wumpf (approved):
nice! some comment nits, but I think this makes sense! :)
As always it would be _greeeaaaaat_ if we could investigate some time into provoking these errors in a unit test environment

crates/viewer/re_renderer/src/video/player.rs
IsseW: No we usually enqueue more than `requested_sample_id`. So it being ahead is the "usual" case when playing a video. Will add a clarifying comment.

If we make a jump backwards and not within the same gop I'm pretty sure we reset.

crates/viewer/re_viewer_context/src/cache/video_stream_cache/test_player.rs
```diff
@@ -0,0 +1,478 @@
...
+            &self.video_descr,
+            &mut |_, _| Ok(()),
+            &|_| &[],
+        )?;
+
+        Ok(())
+    }
+
+    #[track_caller]
+    fn expect_encoded_samples(&self, samples: impl IntoIterator<Item = SampleIndex>) {
```
Wumpf: aren't  these **de**coded samples? Those are the samples that were bounced back from the decoder after all
Guess depends on your point of view since we're sorta forwarding the encoded samples. But I'd still call it "decoded" since they came through our test decoder!

crates/viewer/re_viewer_context/src/cache/video_stream_cache/test_player.rs
```diff
@@ -0,0 +1,478 @@
...
+
+    let mut video = create_video(samples).unwrap();
+
+    video.play(0.0..1.0, 0.1).unwrap();
+
+    video.expect_encoded_samples(expected_indices);
+}
+
+#[test]
+fn player_with_unloaded() {
```
Wumpf, line 398: this one takes me a bit to digest and see where all the numbers come from, but uuhhhhh that one is super useful

crates/viewer/re_viewer_context/src/cache/video_stream_cache/test_player.rs
```diff
@@ -0,0 +1,478 @@
...
+                )
+            ),
+            "Expected {} got {err}",
+            VideoPlayerError::InsufficientSampleData(
+                InsufficientSampleDataError::ExpectedSampleNotAvailable
+            )
+        );
+    }
+
+    assert_loading(video.play(1.0..2.0, 0.25));
```
Wumpf: nit: would be nice to do this more in order. Or is there a particular test-relevant reason we start at 1?

crates/viewer/re_viewer_context/src/cache/video_stream_cache/test_player.rs
```diff
@@ -0,0 +1,478 @@
...
+
+    video.expect_encoded_samples([]);
+
+    video.play(0.0..0.75, 0.25).unwrap();
+    video.expect_encoded_samples(0..3);
```
Wumpf: what happens if we play from 0.0 to 1.0 before populating things? Should still give us the first 3 samples, right? Do you think this is worth adding?

IsseW: That would give us an error, since after 0.75 there is an unloaded frame with an unknown presentation/decode time. I think we could potentially store a "min time" for unloaded samples since we know their chunk's time range and play up until that time at least. But probably not for this PR.

crates/viewer/re_viewer_context/src/cache/video_stream_cache/test_player.rs
IsseW: Made the timestamps match the indices to make it easier to read

PR #12440, opened 2026-01-14
Use SwiftShader software rasterizer on MacOS CI

Wumpf:
### Related

* Fixes #11359 
* Partially #8475
* Depends on https://github.com/rerun-io/rerun/pull/12444
* Depends on https://github.com/rerun-io/rerun/pull/12449

### What

The setup software rasterizer script got extended to have an upload utility and there's now a vulkan upload script as well. Both utils are a bit wordy due to their vibecoded nature, but I found it pretty useful!
At CI runtime we just just grab a prebuilt swiftshader dylib and set up an ICD file that points to it. which is pretty straight forward.

Unfortunately SwiftShader seems to be a little bit buggy when it comes to accuracy of filtering and MSAA. I had to bump treshholds very high and upload larger bgr test images to workaround these problems

PR #12444, opened 2026-01-14
More robust world grid rendering

Wumpf:
### Related

* prerequisite of https://github.com/rerun-io/rerun/pull/12440

### What

The world grid rendering has been plagued by precision issues since its inception. This got particularly bad now when trying out the SwiftShader software rasterizer where the entire grid was significantly shifted. After some experimentation I found the culprit to be the large geometry size which some rasterizers struggle with. The solution is that we rely on a geometry independent raycast in the fragment shader now. This is a bit more ALU intensive, but yields much much more consistent results!

### DIFF
--- a/crates/viewer/re_ui/src/testing.rs
+++ b/crates/viewer/re_ui/src/testing.rs
@@ -32,13 +32,13 @@ fn default_snapshot_options_for_ui() -> SnapshotOptions {
     SnapshotOptions::default().failed_pixel_count_threshold(10) // we sometimes have a few wrong pixels in text rendering in egui for unknown reasons
 }
 
-fn default_snapshot_options_for_3d(viewport_size: Vec2) -> SnapshotOptions {
+pub fn default_snapshot_options_for_3d(viewport_size: Vec2) -> SnapshotOptions {
     // We sometime have "binary" failures, e.g. a pixel being categorized
     // as either inside or outside a primitive due to platform differences.
     // How many depend on the size of the image.
     let num_total_pixels = viewport_size.x * viewport_size.y;
 
-    let broken_pixels_fraction = 5e-4; // 0.05% of pixels.
+    let broken_pixels_fraction = 4e-4; // 0.04% of pixels.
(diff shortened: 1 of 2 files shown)


emilk (changes requested):
The intensity now jumps when viewed from below (chromium)
Jan-14-2026 14-51-13

Wumpf:
nice catch!! have a repro, taking a look

Wumpf:
@emilk should be fixed, was a fairly straight forward oversight

crates/viewer/re_renderer/shader/world_grid.wgsl
```diff
@@ -83,11 +82,20 @@ fn main_fs(in: VertexOutput) -> @location(0) vec4f {
     // We're not actually implementing the "pristine grid shader" which is a grid with world space thickness,
     // but rather the pixel space grid, which is a lot simpler, but happens to be also described very well in this article.
 
+    // Use a camera ray intersection instead of interpolating the plane vertex positions.
+    // Since those can be very far away, this approach ends up being much more precise!
+    // (also it has the added benefit that we're independent of the geometry in general!)
+    let camera_ray = camera_ray_from_fragcoord(in.position.xy);
+    let plane_world_position = intersect_ray_plane(camera_ray, config.plane) * camera_ray.direction + frame.camera_position;
+    let plane_y_axis = normalize(cross(config.plane.normal, select(vec3f(1.0, 0.0, 0.0), vec3f(0.0, 1.0, 0.0), config.plane.normal.x != 0.0)));
+    let plane_x_axis = cross(plane_y_axis, config.plane.normal);
```
Wumpf, line 91: those two depend entirely on uniforms. Guess we could put them into the uniforms directly and/or pass them from the vertex shader. But then again, who cares about a tiny bit of ALU usage 🤷. Having less uniforms & attributes is always nice imho

Wumpf:
> We can try lowering this now:

I give it a pass here, let's see what happens

Wumpf:
~much~ little bit lower threshhold now, caught some outdated screenshots

PR #12446, opened 2026-01-14
Fix links in message-formats.md for docs-latest

MichaelGrupp:


MichaelGrupp:
https://github.com/rerun-io/rerun/blob/michael/fix-docs-latest-links/docs/content/reference/mcap/message-formats.md links work now

PR #12448, opened 2026-01-14
Fix re-exported symbols being listed with `rerun_bindings.rerun_bindings` prefix in the Python docs

abey79:
Fix an issue which I believe was introduced in #11650 where some symbols where displayed with `rerun_bindings.rerun_bindings` prefix in the front page tables of the Python docs:

image

My fix appears to do the right thing based on `pixi run py-docs-serve`.


ntjohnson1 (approved):
This looks good to me. The docs and links weren't rendering before so that was my main focus. This looks like the content still lives and links still work.

Issue #8475, opened 2024-12-16
Make graphics adapter creation list all adapters one by one and inform why they weren't picked


Wumpf:
Closely related to:
* https://github.com/rerun-io/rerun/issues/6581
* https://github.com/rerun-io/rerun/issues/7385

Today, we (that's egui) use `request_adapter` to pick a graphics device. That's fine for many usecases, but it would be better if we went through all available adapters one by one using `enumerate_adapters` and list out the reasons why we didn't pick each of them.
The result should be available both on failure case (none is suitable) and on success case (in debug log and upon request in the viewer)

It's tempting to pull out adapter creation entirely out of egui and into re_renderer. However we need to take into account the _surface_ as well:
* this is a strict *requirement* when using WebGL
* on OpenGL & Linux it _helps_ to know the surface because either Vulkan or OpenGL may not be happy with all surfaces

PR #12449, opened 2026-01-14
Enable vulkan backend for Mac Viewer builds, select graphics adapter more explicitely internally

Wumpf:
### Related

* Partially #8475
* Split out of https://github.com/rerun-io/rerun/pull/12440
    * this is also what motivated this in the first place

### What

Enables `vulkan-portability`, i.e. Vulkan support on MacOS. This is needed for supporting software rasterization on Mac since there 's no Metal software rasterizer!
It's not a huge dependency so I'd rather not make this a feature flag variation and it gives us some more flexibility as it could happen that in the futures users may want to use MoltenVK to work around intermittent issues.

As a consequence I ran into an issue where my local MoltenVk install was used over Metal. So I decided to finally implement custom adapter enumeration (see #8475) to take this into our own hands. It looks a bit scary, but I'm fairly confident this does the right thing; nice thing is that if it doesn't we have much more control now.


Cargo.toml
```diff
@@ -427,6 +427,7 @@ wgpu = { version = "27.0.1", default-features = false, features = [
   # Backends (see https://docs.rs/wgpu/latest/wgpu/#feature-flags)
   "gles",
   "metal",
+  "vulkan-portability", # Enables use of Vulkan on MacOS. We use this for software rasterizing our CI (there aren't any Metal software rasterizers and we had issues with flaky tests when creating lots of Metal devices in parallel)
```
Wumpf, line 430: in this PR we don't enable said software rasterization yet, but it naturally is the prime motivation for enabling `vulkan-portability`

emilk (approved):

PR #12451, opened 2026-01-14
Display chunk download bandwidth

IsseW:
### Related

- Closes RR-3297

### What

Displays chunk download bandwidth on the time panel.

https://github.com/user-attachments/assets/fbeb880f-9991-4f2d-b595-1ebf92534492

Screenshot 2026-01-14 at 18 29 21


crates/store/re_entity_db/src/chunk_promise.rs
```diff
@@ -49,17 +90,20 @@ impl ChunkPromises {
...
 
+        let history = &mut self.download_size_history;
+        history.flush(time);
         self.batches.retain_mut(|batch| {
             let mut promise_opt = batch.promise.lock();
             if let Some(promise) = promise_opt.take() {
                 match promise.try_take() {
                     Ok(Ok(chunks)) => {
                         all_chunks.extend(chunks);
+                        history.add(time, ByteFloat(batch.size_bytes_uncompressed as f64));
```
emilk: Ouch, this is a bug that inflates the numbers quite a bit. Bandwidth should be _bytes on the wire_ over time.

Unfortunately that is less easy to find. Maybe @teh-cmc can help you here

emilk (changes requested):
Please make sure to validate it with an external tool first! I did, and found the values a lot higher than what e.g. [stats](https://github.com/exelban/stats) reported

crates/store/re_entity_db/src/chunk_promise.rs
teh-cmc: For all realistic cases that actually matter, the compressed size in the RRD manifest is what will be transferred over the wire, so we just need to glue that all the way up into `ChunkPromiseBatch`, really

emilk: But the RRDManifest always stores the uncompressed size, but not always the compressed size:

https://github.com/rerun-io/rerun/blob/eebeff73b292e538633c6d318d23cdfa555f5fc3/crates/store/re_log_encoding/src/rrd/footer/instances.rs#L138-L146

IsseW: I think this has to be done at a lower level if we want it to accurately represent compressed data on the wire. I was initially thinking that showing uncompressed made sense since that information ties in to the sizes we display in the selection panel.

For large chunks it's also very inaccurate as this only displays throughput as the chunks arrive

teh-cmc: `chunk_byte_size` _is_ an accurate representation of the data passed onto the wire. In the case of the OSS server, this happens to be an uncompressed value, but that's only because the OSS server passes uncompressed data on the wire.

IsseW: Changed it to use the byte size now, it's more inline on what I see on `stats`, but not fully. I think because we only add chunks to history as they have fully arrived.

emilk (approved):
Nice; seems accurate now!

PR #12452, opened 2026-01-14
Fix transforms & videos breaking when receiving new data in the background

Wumpf:
### Related

* Fixes RR-3329

### What

This keeps all caches up to date.
We first tried discarding them which also worked fine, but decided to go with keeping things up to date since most likely those recordings are still in used. If it gets too much memory used, GC will do.

### DIFF
--- a/crates/viewer/re_viewer/src/app.rs
+++ b/crates/viewer/re_viewer/src/app.rs
@@ -2549,22 +2549,16 @@ impl App {
 
     fn process_store_events_for_db(
         &self,
-        store_hub: &mut StoreHub,
+        store_hub: &StoreHub,
         store_id: &StoreId,
         store_events: &[re_chunk_store::ChunkStoreEvent],
     ) {
-        if let Some(caches) = store_hub.active_caches()
-            && &caches.store_id == store_id
-            && let Some(entity_db) = store_hub.entity_db(store_id)
+        // Keep all caches up to date, even if they're in the background.
+        // This ensures that when we switch to a different recording, the caches are already valid.
+        if let Some(entity_db) = store_hub.entity_db(store_id)
+            && let Some(caches) = store_hub.caches_for_store(store_id)
         {
-            // If there's active cahces for this store, notify about the new data, so things stay up to date.
             caches.on_store_events(store_events, entity_db);
-        } else {
-            // If there's no active caches for this store id, make sure that we don't have outdated ones around.
-            //
-            // Caches are supposed to be lazy access and fast to populate,
-            // but we don't want to pay a cost for continously updating data that we're not using right now!
-            store_hub.clear_caches_for_store(store_id);
         }
 
         self.validate_loaded_events(store_events);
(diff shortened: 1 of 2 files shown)


crates/viewer/re_viewer_context/src/store_hub.rs
```diff
@@ -619,6 +619,11 @@ impl StoreHub {
         caches
     }
 
+    /// Fully resets, i.e. removes, the caches for a given if there's any.
+    pub fn reset_caches_for_store(&mut self, store_id: &StoreId) {
```
emilk:
```suggestion
    /// Fully resets, i.e. removes, the caches for a given store if there's any.
    pub fn reset_caches_for_store(&mut self, store_id: &StoreId) {
```

crates/viewer/re_viewer/src/app.rs
```diff
@@ -2549,13 +2549,22 @@ impl App {
...
-        // TODO(RR-3329): Should this run for all caches?
-        if let Some(caches) = store_hub.active_caches() {
+        if let Some(caches) = store_hub.active_caches()
+            && &caches.store_id == store_id
+            && let Some(entity_db) = store_hub.entity_db(store_id)
+        {
+            // If there's active cahces for this store, notify about the new data, so things stay up to date.
             caches.on_store_events(store_events, entity_db);
+        } else {
+            // If there's not active caches for this store id, make sure that we don't have outdated ones around.
```
```suggestion
            // If there's no active caches for this store id, make sure that we don't have outdated ones around.
```

crates/viewer/re_viewer/src/app.rs
```diff
@@ -2549,13 +2549,22 @@ impl App {
...
+            //
+            // Caches are supposed to be lazy access and fast to populate,
+            // but we don't want to pay a cost for continously updating data that we're not using right now!
+            store_hub.reset_caches_for_store(store_id);
```
emilk: I think this is better naming

```suggestion
            store_hub.clear_caches_for_store(store_id);
```

Wumpf:
> The original problem is that not all store events reach all the cashes.

so but if we don't have a cache to begin with it should work fine since upon initial creation of any of these caches we restore all the data from the store. So I do think this is a viable approach!

Relevant "examples":
* transform cache populates itself from the store on first access https://github.com/rerun-io/rerun/blob/a65732fc2024fe6b79d815637e7c65f530d9617d/crates/viewer/re_viewer_context/src/cache/transform_database_store.rs#L35
* video stream cache doing a similar thing on a per video base https://github.com/rerun-io/rerun/blob/a65732fc2024fe6b79d815637e7c65f530d9617d/crates/viewer/re_viewer_context/src/cache/video_stream_cache.rs#L138


I meant to draft this PR though since I didn't have a repro before/after yet as noted on Slack, that's an oversight on my part!

emilk:
> upon initial creation of any of these caches we restore all the data from the store

Oh I missed that! Then this should work.

I will try to repro and test

emilk:
I created this issue, which _may_ be related: https://linear.app/rerun/issue/RR-3342/debug-crash-video-santity-check

Wumpf:
as discussed: alternative approach of always updating the caches. Confirmed with above snippet

PR #12453, opened 2026-01-14
Simplify entity path hierarchy of static URDF assets

MichaelGrupp:
Structure URDF visual and collision geometries in separate flat entity paths, and all default joint transforms on a separate entity.

Allows easier toggling of the two geometry/mesh types, and avoids super-long entity paths.

```
../robot/visual_geometries/...
../robot/collision_geometries/...
../robot/joint_transforms
```

**Backwards compatibility**: This is only a change of how we load the static assets and does not change the frame IDs, so it can be animated with the same dynamic transform data as before. Also, existing RRDs that were loaded with the previous hierarchy aren't affected of course and still work.

**But** we remove the previous entity path related methods of `UrdfTree`, because they don't make sense now anymore. Instead, there is now `get_visual_geometry_paths()` and `get_collision_geometry_paths()`. See the example in the docs for how this can be used e.g. to change mesh colors during runtime.

https://github.com/user-attachments/assets/1c514511-06c7-4f6d-b078-9fd866b36d6d

### DIFF
--- a/crates/store/re_data_loader/src/loader_urdf.rs
+++ b/crates/store/re_data_loader/src/loader_urdf.rs
@@ -429,10 +429,16 @@ fn walk_tree(
 fn get_joint_transform(joint: &Joint) -> Transform3D {
     let Joint {
         name: _,
+        joint_type: _,
         origin,
         parent,
         child,
-        ..
+        axis: _,
+        limit: _,
+        calibration: _,
+        dynamics: _,
+        mimic: _,
+        safety_controller: _,
     } = joint;
 
     transform_from_pose(origin, parent.link.clone(), child.link.clone())
@@ -530,15 +536,18 @@ fn log_link(
     link: &urdf_rs::Link,
 ) -> anyhow::Result<()> {
     let urdf_rs::Link {
-        name: link_name, ..
+        name: link_name,
+        inertial: _,
+        visual: _,
+        collision: _,
     } = link;
 
     for (visual_entity_path, visual) in urdf_tree.get_visual_geometries(link).unwrap_or_default() {
         let urdf_rs::Visual {
+            name: _,
             origin,
             geometry,
             material,
-            ..
         } = visual;
 
         let instance_scale = extract_instance_scale(geometry);
@@ -578,7 +587,9 @@ fn log_link(
         urdf_tree.get_collision_geometries(link).unwrap_or_default()
     {
         let urdf_rs::Collision {
-            origin, geometry, ..
+            name: _,
+            origin,
+            geometry,
         } = collision;
         let instance_scale = extract_instance_scale(geometry);
 


crates/store/re_data_loader/src/loader_urdf.rs
```diff
@@ -288,157 +287,124 @@ fn log_robot(
...
+        &visual_root_path,
+        &collision_root_path,
         &urdf_tree.root.name,
         timepoint,
     )?;
 
+    // Send all transforms as rows in a single chunk.
+    if !transforms.is_empty() {
+        send_transforms_batch(tx, store_id, &transforms_path, &transforms, timepoint)?;
+    }
```
MichaelGrupp, line 391: I first used our previous send_transform before, but it overwrote the (static) chunk row for each new transform because now the transforms aren't all on separate entities 😅 So we have to be careful now and send a batch... I'm not yet familiar with the chunk API, so please let me know if there's a better way to implement this.

(this here works tho)

crates/store/re_data_loader/src/loader_urdf.rs
```diff
@@ -288,157 +287,124 @@ fn log_robot(
...
-        joint_path.clone(),
-        origin,
-        timepoint,
-        parent.link.clone(),
-        child.link.clone(),
-    )?;
+    transform_from_pose(origin, parent.link.clone(), child.link.clone())
+}
 
-    log_debug_format(
```
MichaelGrupp: I removed these obscure log_debug thingies to clean up the code, or does anyone miss those?

MichaelGrupp:
@rerun-bot full-check

crates/store/re_data_loader/src/loader_urdf.rs
MichaelGrupp, line 1: **TODO:** update snippet test RRD to reflect entity paths change (see full build CI).

MichaelGrupp:
@rerun-bot full-check

docs/snippets/all/howto/minimal.urdf
MichaelGrupp, line 1: Added a collision mesh to this dummy URDF and updated the snippet RRD.

MichaelGrupp:
TODO: check if this breaks anything in `UrdfTree` Python API & docs

MichaelGrupp:
@rerun-bot reality-sync

MichaelGrupp:
@rerun-bot full-check

MichaelGrupp:
@rerun-bot full-check

crates/store/re_data_loader/src/loader_urdf.rs
oxkitsune: being able to inspect the joint type might be nice, but I’m not particularly attached to it.

crates/store/re_data_loader/src/loader_urdf.rs
oxkitsune: Is this a potential footgun for users? How easy is it to accidentally overwrite these static transforms when trying to log robot state? E.g. when using the python utility?

What happens if we log a URDF file with `static=False`, and then log transforms on the same entity?

crates/store/re_data_loader/src/loader_urdf.rs
```diff
@@ -318,127 +403,61 @@ fn walk_tree(
...
         origin,
         parent,
         child,
-        axis,
-        limit,
-        calibration,
-        dynamics,
-        mimic,
-        safety_controller,
+        ..
```
oxkitsune: when ingesting user data it is a bit risky to use a destructor like this, as we're essentially throwing away data. In this case we're no longer logging the "debug" information and I don't think the URDF spec will change anytime soon, but it's worth to keep in mind.

MichaelGrupp: I can readd them and the debug logs, now I understand better why it was done like that. Will add a comment to not get confused again.

crates/store/re_data_loader/src/loader_urdf.rs
MichaelGrupp: will readd, see comment above

crates/store/re_data_loader/src/loader_urdf.rs
MichaelGrupp: The staticness was already there before. Overwriting the static transforms is a feature, not a bug. Quote from transform docs: 

> Note: A current limitation to this is that once a Transform3D (or Pinhole) relating two frames has been logged to an entity, this particular relation may no longer be logged on any other entity. An exception to this rule is [static data](https://rerun.io/docs/concepts/logging-and-ingestion/static): if you log a frame to frame relationship on an entity with static time, you can later on use a different entity for temporal information. This is useful to specify "default" transforms without yet knowing what timeline and paths are going to be used for temporal transforms.

We specifically added this for the URDF use case, where you have static default transforms that you may wanna override later on some other `/tf` entity.

MichaelGrupp: > What happens if we log a URDF file with static=False, and then log transforms on the same entity?

If you put `static=False`, you **have** to log new transforms to exactly the same entity. This is however not new for this branch, this is already the behavior on `main` (e.g. if you do that with the `animated_urdf.py` example, it won't work because the transform updates are on a separate entity). Not sure if we have to document this clearer?

MichaelGrupp: Resolved as discusses on Slack: the transforms are now __always__ static, see `send_static_transforms_batch` and its comment.

crates/store/re_data_loader/src/loader_urdf.rs
MichaelGrupp: I decided to not readd them now, can be added later if anybody needs it really.

MichaelGrupp:
@rerun-bot full-check

MichaelGrupp:
@rerun-bot reality-sync

MichaelGrupp:
Closing because already synced via merge bot.

PR #12454, opened 2026-01-14
Refactor getting started to match product layout

ntjohnson1:
### Related
RR-3263

### What
* Remove quick-start
This was highly redundant with what we had in data-in / logging and ingestion. There is no way to provide a true multi language getting started anymore because different languages have different coverage in the various projects. Instead pointing people to their objective then splitting makes the most sense to me.

The content itself doesn't need to much adjustment besides a small amount of framing.

### DIFF
--- a/docs/content/_redirects.yaml
+++ b/docs/content/_redirects.yaml
@@ -28,9 +28,9 @@ getting-started/installing-viewer: overview/installing-rerun/viewer
 getting-started/data-out/analyze-and-log: getting-started/data-out/analyze-and-send
 getting-started/data-out/query-data: getting-started/data-out
 getting-started/quick-start: getting-started
-getting-started/quick-start/cpp: getting-started/cpp
-getting-started/quick-start/python: getting-started/python
-getting-started/quick-start/rust: getting-started/rust
+getting-started/quick-start/cpp: getting-started/data-in/cpp
+getting-started/quick-start/python: getting-started/data-in/python
+getting-started/quick-start/rust: getting-started/data-in/rust
 
 # Concepts section (old paths moved to subfolders)
 concepts/app-model: concepts/how-does-rerun-work


docs/content/getting-started/quick-start/cpp.md
```diff
@@ -1,135 +1,5 @@
 ---
-title: C++
-order: 1
+title: CPP
+hidden: true
+redirect: getting-started/data-in/cpp
```
abey79: hidden redirect are best moved to `_redirects.yaml` to declutter

docs/content/getting-started/quick-start/python.md
```diff
@@ -1,150 +1,5 @@
 ---
 title: Python
-order: 2
+hidden: true
+redirect: getting-started/data-in/python
 ---
```
abey79: ditto

docs/content/getting-started/quick-start/rust.md
```diff
@@ -1,78 +1,5 @@
 ---
-title: Rust
-order: 3
+title: CPP
+hidden: true
+redirect: getting-started/data-in/rust
```
abey79: ditto (it'll even save you to fix the cpp -> rust copy/paste typo ;))

docs/content/getting-started/quick-start.md
```diff
@@ -1,16 +1,5 @@
 ---
-title: Quick start
-order: 200
+title: Quick Start
+hidden: true
+redirect: getting-started
```
abey79: ditto

docs/snippets/all/howto/query_images.py
```diff
@@ -86,6 +86,6 @@
         pass
 if frame is None:
     raise RuntimeError("Failed to decode any frame from video stream.")
-image = np.asarray(frame.to_image())
+image = np.asarray(frame.to_image())  # type: ignore[attr-defined]
```
abey79: its not super great to have this in snippets. any idea why it showed up?

docs/content/getting-started/quick-start/cpp.md
ntjohnson1: Agreed but this wasn't working for me before. Seems to 2work now 🤷  so I'll move them all to redirects table (might be related to the website PR I put up to fix some other bugs)

docs/snippets/all/howto/query_images.py
ntjohnson1: I'm not sure what caused this to be unhappy. But I just annotated the variable which looks nicer.

PR #12458, opened 2026-01-14
Listen on both ipv4+ipv6 rather than ipv4 only by default

alvaroaleman:
### What

Listen on both ipv4+ipv6 rather than ipv4 only by default


emilk:
To fix the CI error, add `IPv4` and `IPv6` to `doc-valid-idents` in https://github.com/rerun-io/rerun/blob/42f2479854d591c21eea1eccb2c47a4a9b7ae39d/clippy.toml#L98-L125 and https://github.com/rerun-io/rerun/blob/42f2479854d591c21eea1eccb2c47a4a9b7ae39d/scripts/clippy_wasm/clippy.toml#L114-L142

Screenshot 2026-01-15 at 09 04 29

alvaroaleman:
> Nice! Have you tested it?

Yeah, I did a `./target/debug/rerun --serve-grpc`, with this change I can connect to that both using `[::1]:9876` and `127.0.0.1:9876`, without it only the latter works.

alvaroaleman:
@emilk could you have another look, please?

emilk:
NOTE: It would have been easier to re-review if there was a single commit with the changes since my first review. Instead there is a force-push, so now I need to re-review everything.

emilk (approved):
Nice!

emilk:
@rerun-bot reality-sync

PR #12466, opened 2026-01-15
Fix time point sharing for duration timelines

IsseW:
### Related

- Closes RR-3348

### What

Fixes & cleans up time formatting in urls.

### DIFF
--- a/crates/store/re_log_types/src/index/time_cell.rs
+++ b/crates/store/re_log_types/src/index/time_cell.rs
@@ -12,6 +12,9 @@ pub struct TimeCell {
     pub value: NonMinI64,
 }
 
+// We shouldn't implement display for this as it's too ambiguous, instead create specific functions.
+static_assertions::assert_not_impl_any!(TimeCell: std::fmt::Display);
+
 impl TimeCell {
     pub const ZERO_DURATION: Self = Self {
         typ: TimeType::DurationNs,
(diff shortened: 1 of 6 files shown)


crates/store/re_uri/src/time_selection.rs
```diff
@@ -17,6 +17,20 @@ impl TimeSelection {
             TimeCell::new(self.timeline.typ(), self.range.max).format_compact(timestamp_format),
         )
     }
+
+    pub fn url_format(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
emilk: docstring!

crates/store/re_uri/src/time_selection.rs
```diff
@@ -31,19 +45,6 @@ impl From<TimeSelection> for AbsoluteTimeRange {
     }
 }
 
-impl std::fmt::Display for TimeSelection {
```
emilk, line 34: Maybe add a `static_assertion` that we shouldn't implement `Display` for `TimeSelection`, and explain why

crates/store/re_log_types/src/index/time_cell.rs
```diff
@@ -190,15 +190,13 @@ impl TimeCell {
             TimeType::Sequence => typ.format(value, timestamp_format),
         }
     }
-}
 
-// TODO(andreas): Should remove this in favor or explicit format methods above.
-impl std::fmt::Display for TimeCell {
```
emilk, line 196: Same here: add a `static_assertion` so we don't regress!

https://docs.rs/static_assertions/latest/static_assertions/macro.assert_not_impl_any.html

PR #12469, opened 2026-01-15
Initial CLAUDE.md

Wumpf:
Handwrote one at first and then had Claude fill out the gaps. Ironic? yes, but it took fair amount of time of collecting the data for that which it will need to do less of now! I then slimmed it down a bit from what it put there and removed stuff that's too much in flux

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,160 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) and other LLMs when working with code in this repository.
+
+## Project overview
+
+Rerun is a time-aware multimodal data stack and visualizations tool used in robotics, spatial AI, computer vision, and similar domains. It provides SDKs (Python, Rust, C++) for logging rich data (images, point clouds, tensors, etc.) and a Viewer for visualization.
+
+## Build system
+
+We use `pixi` for task management and dependency installation. Check `pixi.toml` for a full list of tasks.
+
+### Essential commands
+
+**Building:**
+- `pixi run py-build` - Build Python SDK into local .venv (uses uv)
+- `pixi run rerun-build` - Build native viewer (without web viewer)
+- `pixi run rerun-build-web` - Build web viewer (wasm)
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,159 @@
...
+- **Immediate Mode**: The entire viewer is rendered from scratch each frame (no state management callbacks)
+- **Arrow Native**: Data is stored, transmitted, and queried as Apache Arrow arrays
+- **Multi-language**: Changes to .fbs files affect Rust, Python, and C++ simultaneously
+
+## Development references
+
+- [`ARCHITECTURE.md`](ARCHITECTURE.md) - Detailed architecture documentation
+- [`BUILD.md`](BUILD.md) - Full build instructions
+- [`CODE_STYLE.md`](CODE_STYLE.md) - Code style guidelines
+- [`CONTRIBUTING.md`](CONTRIBUTING.md) - Contribution guidelines
```
emilk, line 158:
```suggestion
- [`CONTRIBUTING.md`](CONTRIBUTING.md) - Contribution guidelines
- [`DESIGN.md`](DESIGN.md) - Guidelines for UI design, covering GUI, CLI, documentation, log messages, etc
```

CLAUDE.md
```diff
@@ -0,0 +1,159 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
```
emilk:
```suggestion
This file provides guidance to Claude Code (claude.ai/code) and other LLMs when working with code in this repository.
```

CLAUDE.md
Wumpf: nice catch

oxkitsune:
Maybe we should also include an `AGENTS.md` that just says: "Use `CLAUDE.md` instead of `AGENTS.md`" 🤔

Wumpf:
sure why not
we already have so many top level files anyways 🤪

Wumpf:
Actually we could just check in a symlink heh

PR #12470, opened 2026-01-15
Set `RERUN_FLUSH_TICK_SECS` in `micro_batching` snippet

grtlr:
### Related

* CI failures like this: https://github.com/rerun-io/rerun/actions/runs/21025520773/job/60452855381

### What

Turns out the, `RERUN_FLUSH_TICK_SECS` was (implicitly left) at the default of 200ms for the `micro_batching` snippet which _could_ be the reason for the CI failures that we're seeing.

The intuition is that if for some reason we need longer then `RERUN_FLUSH_TICK_SECS` to fill up `RERUN_FLUSH_NUM_ROWS` (for example because we send off recording properties or concurrency) we create an unnecessary chunk.

* [x] `full-check` passes

### DIFF
--- a/docs/snippets/all/howto/micro_batching.py
+++ b/docs/snippets/all/howto/micro_batching.py
@@ -4,6 +4,7 @@ Shows how to configure micro-batching directly from code.
 Check out <https://rerun.io/docs/reference/sdk/micro-batching> for more information.
 """
 
+from datetime import timedelta
 import rerun as rr
 
 # Equivalent to configuring the following environment:


docs/snippets/all/howto/micro_batching.py
```diff
@@ -9,9 +9,11 @@
 # Equivalent to configuring the following environment:
 # * RERUN_FLUSH_NUM_BYTES=<+inf>
 # * RERUN_FLUSH_NUM_ROWS=10
+# * RERUN_FLUSH_TICK_SECS=10
 config = rr.ChunkBatcherConfig(
     flush_num_bytes=2**63,
     flush_num_rows=10,
+    flush_tick=10,
```
jleibs: Why didn't we name this `flush_sick_secs`? 😭 

Maybe worth being clear using `timedelta`?
```suggestion
    flush_tick=timedelta(seconds=10),
```

grtlr:
@rerun-bot full-check

PR #12275, opened 2025-12-18
Allow multiple visualizers of same type per `Entity`

aedm:
### Related

* Closes RR-3153
* Closes RR-3154
* Closes RR-3155

### What

image

Really what this is all about is to change the blueprint datamodel to accommodate for:
* multiple visualizers of same type on the same entity
    * have overrides for each of these!
* component mappings - feature gated right now, not everything is enabled yet in this PR
    * we'll stabilize and connect the dots in follow-up iterations! 

**Breaking**: Unless we add migration code later, this breaks all existing blueprint data for overrides (including "entity global" settings like visibility)

This PR paves the way for component mappings, i.e. the ability to remap arbitrary components into visualizers so that you can for instance plot anything vaguely scalar shaped. Most of the datastructures & operations are put into place here, but we haven't flipped the switch yet!

An early prototype of setting component mappings is available behind an experimental feature flag. However, visualizers can't yet be spawned (automatically or explicit) unless the required component (like `scalars` for series types or `positions` for `Points3D)` is present with its builtin type.

* [x] pass full-ci check

PR #12475, opened 2026-01-15
Use visualizer objects in Rust blueprint override API

Wumpf:
This gets it back on par with Python!
One difference is that we don't write `VisualizerOverrides` out yet since I expect all this to be changed anyways again  in https://github.com/rerun-io/rerun/pull/12275

Manually tested with `cargo run -p blueprint`

### DIFF
--- a/crates/viewer/re_view_spatial/tests/transform_child_parent_single_time.rs
+++ b/crates/viewer/re_view_spatial/tests/transform_child_parent_single_time.rs
@@ -235,6 +235,7 @@ fn test_transform_axes_for_explicit_transforms() {
         ctx.save_blueprint_archetype(
             transforms_override_path.clone(),
             &blueprint::archetypes::VisualizerOverrides::new([
+                // TODO(RR-3153): remove the `as_str()`.
                 archetypes::TransformAxes3D::default()
                     .visualizer()
                     .visualizer_type


crates/store/re_sdk_types/src/visualizer.rs
```diff
@@ -0,0 +1,62 @@
+use crate::blueprint::components as bp_components;
+use crate::{AsComponents, SerializedComponentBatch};
+
+/// Configuration for a visualizer in Rerun.
+#[derive(Clone, Debug)]
+pub struct Visualizer {
+    /// Identifies this visualizer instance uniquely within its view.
+    pub id: uuid::Uuid,
```
grtlr: nit: consider adding type safe `VisualizerId(Uuid)` wrapper type here

crates/viewer/re_view_spatial/tests/transform_child_parent_single_time.rs
```diff
@@ -235,8 +235,10 @@ fn test_transform_axes_for_explicit_transforms() {
         ctx.save_blueprint_archetype(
             transforms_override_path.clone(),
             &blueprint::archetypes::VisualizerOverrides::new([
-                // TODO(RR-3153): remove the `as_str()`.
-                archetypes::TransformAxes3D::visualizer().as_str(),
+                archetypes::TransformAxes3D::default()
```
grtlr, line 239: I think the `TODO` should probably stay?

PR #12476, opened 2026-01-15
Fix: tap stream view to move time cursor

emilk:
Due to a bug, only clicks or drags that lasted longer than a frame caused the time cursor to move. A quick tap on a Mac trackpad would _not_ work. Now it does.


emilk:
* https://github.com/emilk/egui/pull/7851

PR #12479, opened 2026-01-15
Fix rerun-root calculations for reality

jleibs:
### What
- The old git-based strategy was incompatible with a monorepo stricture
- This uses a new pattern to search for a `.RERUN_ROOT` sentinel file

Future work: RR-3355 to do this with a shared utility.

### DIFF
--- a/scripts/lint.py
+++ b/scripts/lint.py
@@ -50,10 +50,14 @@ def get_rerun_root() -> str:
     # TODO(RR-3355): Use a shared utility for this
     current = Path(__file__).resolve().parent
     while current != current.parent:
+        # Look for sentinel file
         if (current / ".RERUN_ROOT").exists():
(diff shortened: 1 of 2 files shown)


scripts/lint.py
```diff
@@ -45,6 +45,17 @@
 Frontmatter = dict[str, Any]
 
 
+def get_rerun_root() -> str:
+    # Search upward for .RERUN_ROOT sentinel file
+    # TODO(RR-3355): Use a shared utility for this
+    current = Path(__file__).resolve().parent
```
thz, line 51: Can we please not only stop at `/` but at finding `.git`? This code should not continue traversing the parents outside the git repo.

jleibs: Done

PR #12480, opened 2026-01-15
Streaming cache for redap tables, adding instant filtering

lucasmerlin:
### Related

* closes https://linear.app/rerun/issue/RR-3073/1m-row-tables-cache-segment-table-for-filtering

### What

Implements a streaming cache for datafusion. Also ensures that the current datafusion stream is cancelled when the table is refreshed.

Changing filter is now instant if enough data is loaded: 

https://github.com/user-attachments/assets/56429df8-0ab9-4263-b209-3e2a89096cca

It's weird that the sort order changes when you change filters. I assume this is maybe due to some internal datafusion optimizations? It's not wrong since we don't specify any sort, but it's weeird. Maybe there is a datafusion setting to prevent this?

### DIFF
--- a/crates/viewer/re_dataframe_ui/src/streaming_cache.rs
+++ b/crates/viewer/re_dataframe_ui/src/streaming_cache.rs
@@ -83,6 +83,11 @@ impl StreamingCacheInner {
 ///
 /// When `refresh()` is called, a new inner cache is created. Old streams continue
 /// reading from the old cache until completion, while new scans use the fresh cache.
+///
+/// # Pushdown filters
+///
+/// This provider does not support filter pushdown since it caches the full table. If your
+/// table supports filter pushdown, you probably should not use this cache.
 pub struct StreamingCacheTableProvider {
     /// Input table provider
     input_table: Arc<dyn TableProvider>,


crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
```diff
@@ -127,13 +128,28 @@ pub struct DataFusionTableWidget<'a> {
...
-        session_ctx: &SessionContext,
+        runtime: &AsyncRuntimeHandle,
+        egui_ctx: egui::Context,
+        session_ctx: Arc<SessionContext>,
         table_ref: impl Into<TableReference>,
     ) {
-        let id = id_from_session_context_and_table(session_ctx, &table_ref.into());
+        let table_ref = table_ref.into();
+
+        // Unfortunately, getting a TableProvider is async, so we need to spawn here:
```
abey79, line 138: What's the implication of that? Could something wrong happen if the user spams the refresh button?

crates/viewer/re_dataframe_ui/src/streaming_cache.rs
```diff
@@ -0,0 +1,424 @@
...
+/// reading from the old cache until completion, while new scans use the fresh cache.
+pub struct StreamingCacheTableProvider {
+    /// Factory to create the [`DataFrame`] (called once on first scan).
+    df_factory: DataFrameFactory,
+
+    /// Schema for the cached data.
+    schema: SchemaRef,
+
+    /// Shared cache state. The outer Mutex allows swapping the inner cache on refresh.
+    cache: Mutex<Arc<Mutex<StreamingCacheInner>>>,
```
abey79, line 99: I'm still largely challenged by this stuff, but I've learned to always ask the question. Is it ok for this Mutex to not be async aware, given that the table provider is essentially async?

crates/viewer/re_dataframe_ui/src/streaming_cache.rs
```diff
@@ -0,0 +1,424 @@
...
+        projection: Option<&Vec<usize>>,
+        filters: &[Expr],
+        limit: Option<usize>,
+    ) -> DataFusionResult<Arc<dyn ExecutionPlan>> {
+        enum Action {
+            UseMemTable(Arc<MemTable>),
+            ReturnError(Arc<DataFusionError>),
+            StartStreaming(Arc<Mutex<StreamingCacheInner>>),
+            ReturnStreamingPlan(Arc<Mutex<StreamingCacheInner>>),
+        }
```
abey79, line 223: great pattern! clarifies upfront that what this function is about 👍🏻

crates/viewer/re_dataframe_ui/src/streaming_cache.rs
```diff
@@ -0,0 +1,424 @@
...
+                Ok(Arc::new(CachedStreamingExec::new(inner_cache)))
+            }
+        }
+    }
+
+    fn supports_filters_pushdown(
+        &self,
+        filters: &[&Expr],
+    ) -> DataFusionResult<Vec<TableProviderFilterPushDown>> {
+        // We don't push down filters since we cache the full table
```
abey79, line 276: i think this is worth pointing out in  the struct docstring (that would be a major shortcoming of this table provider in some context—just not how it's being used currently)

crates/viewer/re_redap_browser/src/entries.rs
```diff
@@ -158,11 +158,40 @@ async fn fetch_entries_and_register_tables(
...
+    // Create a schema for raw (uncached) table providers
+    const RAW_SCHEMA: &str = "__rerun_internal__";
+    session_ctx
+        .catalog("datafusion")
+        .expect("default catalog exists")
+        .register_schema(
+            RAW_SCHEMA,
+            Arc::new(datafusion::catalog::MemorySchemaProvider::new()),
+        )
+        .ok();
```
abey79: Does that work if the schema already exists in the session context?

crates/viewer/re_dataframe_ui/src/streaming_cache.rs
lucasmerlin: If you never hold your lock over an await point it should be fine. Since we never do this we're good. I believe it's even recommended to not use tokio::Mutex unless you really really need to hold a lock over an await point. 

From [docs.rs](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html): 
> Contrary to popular belief, it is ok and often preferred to use the ordinary [Mutex](https://doc.rust-lang.org/nightly/std/sync/poison/mutex/struct.Mutex.html) from the standard library in asynchronous code.

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
lucasmerlin: I believe it should be pretty much instant for us. I believe it could be slow if you have a custom schema provider?

crates/viewer/re_dataframe_ui/src/streaming_cache.rs
abey79: TIL. wish there was some kind of lint for that.

crates/viewer/re_dataframe_ui/src/datafusion_table_widget.rs
timsaucer: We use an in-memory schema provider, and there are no `.await` calls inside it so this should not be an issue.

crates/viewer/re_dataframe_ui/src/streaming_cache.rs
timsaucer: I was initially concerned about this but upon investigation we aren't pushing down filters in the outer segment table provider (as opposed to our dataset queries which do) so we're ok for now.

**IF** we ever use our dataset queries here then I would be concerned.

crates/viewer/re_redap_browser/src/entries.rs
timsaucer: In memory catalog doesn't throw an error and will just replace the existing schema with the new one. My question is: is this the right thing to do? I haven't thought enough about when we get multiple calls to this function with the same session context.

crates/viewer/re_dataframe_ui/src/streaming_cache.rs
abey79: Yeah that's exactly what I had in mind. Now it's ok, but let's leave breadcrumbs for when we start rolling this out left and right.

timsaucer:
Here is an example for how to remove registering the inner table, but I haven't been able to test it in the viewer yet: https://github.com/rerun-io/rerun/pull/12527

crates/viewer/re_redap_browser/src/entries.rs
lucasmerlin: It's gone now 👍🏻

lucasmerlin:
@rerun-bot sync-reality

lucasmerlin:
@rerun-bot reality-sync

lucasmerlin:
@rerun-bot reality-sync

lucasmerlin:
@rerun-bot reality-sync

lucasmerlin:
@rerun-bot reality-sync

Issue #12366, opened 2026-01-09
`--newest-first` / `rr.serve_grpc(newest_first=True)` is not serving any data


Wumpf:
Looks like the feature is completely broken?

Simple repro:
Apply the following diff to `examples/python/live_camera_edge_detection/live_camera_edge_detection.py`:

```diff
--- a/examples/python/live_camera_edge_detection/live_camera_edge_detection.py
+++ b/examples/python/live_camera_edge_detection/live_camera_edge_detection.py
@@ -63,12 +63,12 @@ def main() -> None:
     )
     parser.add_argument("--num-frames", type=int, default=None, help="The number of frames to log")
 
-    rr.script_add_args(parser)
+    #   rr.script_add_args(parser)
     args = parser.parse_args()
 
-    rr.script_setup(
-        args,
+    rr.init(
         "rerun_example_live_camera_edge_detection",
+        spawn=False,
         default_blueprint=rrb.Vertical(
             rrb.Horizontal(
                 rrb.Spatial2DView(origin="/image/rgb", name="Video"),
@@ -78,10 +78,11 @@ def main() -> None:
             row_shares=[1, 2],
         ),
     )
+    rr.serve_web_viewer(open_browser=True, connect_to=rr.serve_grpc(newest_first=True))
 
     run_canny(args.num_frames)
 
-    rr.script_teardown(args)
+    # rr.script_teardown(args)
```

With `newest_first=False` things are still working.

Discord thread: https://discord.com/channels/1062300748202921994/1458439690519707794/1458439690519707794

emilk:
This works as expected:

> pixi run rerun-web --newest-first
> pixi run uv run examples/python/live_camera_edge_detection/live_camera_edge_detection.py --connect

🕵️

emilk:
The provided code also works for me on `main`. In other words: I fail to reproduce this problem.

emilk:
What I tried:

```
git checkout emilk/try-reproduce
pixi run py-build
pixi run uv run examples/python/live_camera_edge_detection/live_camera_edge_detection.py 
```

Refresh the browser. It works as expected.

([branch diff](https://github.com/rerun-io/rerun/compare/emilk/try-reproduce?expand=1))

emilk:
Actually, with a high-res camera I *can* reproduce it, together with a bunch of other (serious) problems, like a huge memory leak. Will investigate!

emilk:
* The core problem was solved by #12484
[...]

PR #12484, opened 2026-01-15
Fix early disconnect with `--newest-first`

emilk:
### Related
* Part of https://github.com/rerun-io/rerun/issues/12366
(I'm still investigating that issue - there may be more bugs lurking there)

### What
When using the `newest_first` flag in the SDK or the CLI, clients (the viewer) would unexpectedly disconnect once the backlog of events were caught up to. If the backlog was empty, this meant the client would just never connect.

### Test setup
**server**:  `rerun --serve-grpc --server-memory-limit 2GB --newest-first`
**producer**: `uv run examples/python/live_camera_edge_detection/live_camera_edge_detection.py --connect`
**viewer**: `rerun --memory-limit 2GB --connect`

#### Before
The viewer would connect, and start catch up with the backlog. Once it caught up however, it would disconnect.

#### After
The viewer stays connected. Everything works as expected.

### DIFF
--- a/crates/store/re_grpc_server/src/priority_stream.rs
+++ b/crates/store/re_grpc_server/src/priority_stream.rs
@@ -3,55 +3,55 @@ use std::task::{Context, Poll};
 
 use tokio_stream::Stream;
 
-/// Merges two streams, favoring one above the other.
+/// Merges multiple streams, favoring earlier streams over later ones.
 ///
-/// The merged stream only terminates when the high priority stream terminates.
-/// When the low priority stream is exhausted, we continue polling only the high priority stream.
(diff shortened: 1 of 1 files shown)


teh-cmc (approved):
LGTM since this is a strict improvement...

...That said, and without any further knowledge of the surrounding context here, having something called `PriorityMerge(s1, s2)` that closes as soon as either stream is drained sounds very surprising to me.
I would expect this to continue polling as long as _any_ of the underlying streams are alive, just giving priority to `s1` as much as it's possible.

emilk:
Good point - I've made it much nicer now

PR #12491, opened 2026-01-16
Our search index rust calls python, run it in uv

ntjohnson1:
### Related
Failing docs-latest

### What
Calling a python script from rust is a little bit of an antipattern IMO because it leads to convoluted situations like this. For now run our script inside of our uv env so `mkdocs` is installed and available.

I don't think anything but docs-latest runs our search index build so I just ran it locally to debug/confirm.

### DIFF
--- a/crates/build/re_dev_tools/src/build_search_index/ingest/python.rs
+++ b/crates/build/re_dev_tools/src/build_search_index/ingest/python.rs
@@ -15,14 +15,39 @@ pub fn ingest(ctx: &Context) -> anyhow::Result<()> {
     // run `mkdocs` to generate documentation, which also produces a `objects.inv` file
     // this file contains every documented item and a URL to where it is documented
     progress.set("mkdocs build", ctx.is_tty());
-    Command::new("mkdocs")
+    let mkdocs_config = ctx.workspace_root().join("rerun_py/mkdocs.yml");
+    println!("Running mkdocs build with config: {mkdocs_config}");
+    println!(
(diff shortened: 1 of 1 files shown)


crates/build/re_dev_tools/src/build_search_index/ingest/python.rs
```diff
@@ -15,14 +15,39 @@ pub fn ingest(ctx: &Context) -> anyhow::Result<()> {
...
     progress.set("mkdocs build", ctx.is_tty());
-    Command::new("mkdocs")
+    let mkdocs_config = ctx.workspace_root().join("rerun_py/mkdocs.yml");
+    println!("Running mkdocs build with config: {mkdocs_config}");
+    println!(
+        "Current working directory: {:?}",
+        std::env::current_dir().unwrap_or_default()
+    );
+
+    let mkdocs_result = Command::new("mkdocs")
```
jleibs, line 25: An alternative would be to insert the `uv` here here (`uv run mkdocs ...`)

Somehow `pixi -> rust -> uv -> mkdocs-from-uv` feels like a more natural ordering to me than `pixi -> uv -> rust -> mkdocs-from-uv`

ntjohnson1: It feels like calling even more complex arbitrary executables from our rust makes this even more opaque. If we update our uv stuff in the future we'd never remember to update this. At least now it advertises that we need the environment for the rust call?

PR #12494, opened 2026-01-16
Fix previous login showing up on tests & screenshots

Wumpf:
.. this was messing with locally run integration tests!

### DIFF
--- a/crates/store/re_redap_client/src/connection_registry.rs
+++ b/crates/store/re_redap_client/src/connection_registry.rs
@@ -102,7 +102,7 @@ pub struct ConnectionRegistryHandle {
 
     /// Whether to use credentials stored on the host machine by default.
     /// Since some tests run on a single-threaded tokio runtime and this is never updated,
-    /// it lives outside the RwLock.
+    /// it lives outside the `RwLock`.
     use_stored_credentials: bool,
 }
 


Wumpf:
wait, there's more. dataset_ui also has trouble here

crates/viewer/re_viewer/src/ui/top_panel.rs
```diff
@@ -441,7 +441,10 @@ fn panel_buttons_r2l(
         selection,
     );
 
-    if let Some(auth) = &app.state.auth_state {
+    if let Some(auth) = &app.state.auth_state
+        && !app.is_screenshotting()
+        && !app.app_env().is_test()
```
lucasmerlin: Hmm, this feels like a super hacky fix. I think we should prevent re_auth from loading shared credentials in tests.

I think this piece of code _should_ prevent the auth credentials from being picked up in the test: 

https://github.com/rerun-io/rerun/blob/003019fa7571c57e00cc83f6b49e06ad4f24d803/crates/viewer/re_viewer/src/viewer_test_utils/mod.rs#L50

No idea why it isn't working though.

Wumpf: This drawing part here I'm actually kinda confident about: it's just like we drop all sort of other things like screenshot label etc. from screenshots out of the command palette as well. We do that in order to take screenshots for docs where we don't want that. It seems to make only sense to also drop this part as well?

The thing I'm more nervous about is the diff in app.rs which is geared towards `re_integration_test` not getting any credentials

lucasmerlin:
I tried running tests locally on this branch and it fails with 

```bash
thread 'time_control_emits_timeline_switch_event' panicked at crates/store/re_redap_client/src/connection_registry.rs:475:15:
can call blocking only when running on the multi-threaded runtime
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
🥲

Ah it's also failing with this in CI

Wumpf:
@rerun-bot reality-sync

lucasmerlin (approved):

Wumpf:
got merged and appears on `main`, but the bot responsible for closing isn't working right now

PR #12498, opened 2026-01-18
Add new `re_quota_channel`

emilk:
### Related
* Part of https://linear.app/rerun/issue/RR-3366/investigate-memory-leaks
* Part of https://github.com/rerun-io/rerun/issues/11024

### What
This is a new crate with a channel that can apply backpressure when a certain size limit is reached

### Next steps
Using this channel!

### DIFF
--- a/crates/utils/re_quota_channel/src/sync/mod.rs
+++ b/crates/utils/re_quota_channel/src/sync/mod.rs
@@ -7,7 +7,7 @@
 
 use std::sync::Arc;
 
-use parking_lot::Mutex;
+use parking_lot::{FairMutex, Mutex};
 
 #[cfg(not(target_arch = "wasm32"))]
 use parking_lot::Condvar;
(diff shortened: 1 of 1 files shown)


crates/utils/re_quota_channel/src/sync/mod.rs
```diff
@@ -0,0 +1,437 @@
...
+
+    #[cfg(not(target_arch = "wasm32"))]
+    fn send_impl(&self, msg: T, size_bytes: u64) -> Result<(), SendError<T>> {
+        let capacity = self.shared.capacity_bytes;
+
+        if size_bytes <= capacity {
+            // Normal case: wait until we have room
+
+            {
+                let mut current = self.shared.current_bytes.lock();
```
Wumpf: bit unfortunate that we always take a lock now even if we don't have to wait - crossbeam's channel are afaik totally lock free. But it's not obvious to me either immediately what a datastructure like that would look like. Condition variable (+Mutex) totally seems the right choice for choice for this kind of access pattern. I'm sure it can be done and I'm reasonably sure it would be a loooot more complicated 😅

crates/utils/re_quota_channel/src/sync/mod.rs
```diff
@@ -0,0 +1,437 @@
...
+
+                if capacity < *current + size_bytes {
+                    re_log::debug_once!(
+                        "{}: Channel byte budget ({}) exceeded. Blocking until space is available…",
+                        self.shared.debug_name,
+                        re_format::format_bytes(capacity as f64),
+                    );
+                    while capacity < *current + size_bytes {
+                        self.shared.space_available.wait(&mut current);
+                    }
```
Wumpf: Hmm small messages can starve out big messages. That's a bit of a problem, no? Let's say I spam you on one thread with tons of transforms and on another one I try to send an image, that image will never go through. I fear we have to implement some fairness criterium here? 🤔

crates/utils/re_quota_channel/src/sync/mod.rs
```diff
@@ -0,0 +1,437 @@
...
+            re_log::debug_once!(
+                "{}: Message size ({}) exceeds channel capacity ({}). \
+                 Waiting for channel to empty before sending.",
+                self.shared.debug_name,
+                re_format::format_bytes(size_bytes as f64),
+                re_format::format_bytes(capacity as f64),
+            );
+
+            {
+                // Wait until the channel is completely empty
```
Wumpf, line 138: similar to above, if I have a continous trickle of small messages, this one will never make it 🤔

PR #12501, opened 2026-01-19
Speed up size estimation of transform cache

emilk:
### Related
* Part of https://linear.app/rerun/issue/RR-3366/investigate-memory-leaks
* Related to https://linear.app/rerun/issue/RR-3349/prefetcher-sometimes-fights-gc

### What
As part of more accurate memory tracking I noticed that the `SizeBytes` trait implementation for the transfrom cache was extremely slow, because of it traversing many tiny objects in a `BTreeMap`. So I created a `BookeepingBTreeMap` that keeps track of its own byte size as it is being built and mutated.

This unblocks having a live view of the total memory use of all caches, and indeed most other things in the app (such a view is coming later). It will also unblock more fine-grained decision making when it comes to GC and prefetching.

### DIFF
--- a/crates/store/re_tf/src/transform_resolution_cache.rs
+++ b/crates/store/re_tf/src/transform_resolution_cache.rs
@@ -479,8 +479,6 @@ impl TransformsForChildFrameEvents {
 
 impl SizeBytes for TransformsForChildFrameEvents {
     fn heap_size_bytes(&self) -> u64 {
-        re_tracing::profile_function!();
-
         let Self {
             frame_transforms,
             pinhole_projections,
@@ -848,8 +846,6 @@ impl Clone for PoseTransformForEntity {
 
 impl SizeBytes for PoseTransformForEntity {
     fn heap_size_bytes(&self) -> u64 {
-        re_tracing::profile_function!();
-
         let Self {
             entity_path,
             poses_per_time,


crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -718,42 +721,52 @@ impl TreeTransformsForChildFrame {
...
-                }
-
-                Err(err) => {
-                    re_log::error_once!("Failed to query transformations: {err}");
-                    CachedTransformValue::Cleared
-                }
-            };
-        }
+            .mutate_entry_before(
+                &query.at().inc(),
```
emilk:
```suggestion
            .mutate_latest_at(
                &query.at(),
```

CLAUDE.md
```diff
@@ -71,6 +71,7 @@ To add custom functionality to generated types, create `_ext` files:
 
 - use `…` instead of `...` <!-- NOLINT -->
 - validate various custom conventions via `pixi run lint-rerun <file>` (not passing any file will check everything)
+- Use `format!("{x}")` over `format!("{}, x)` (same in log calls etc)
```
Wumpf, line 74: nice

crates/utils/re_byte_size/src/bookkeeping_btreemap.rs
```diff
@@ -0,0 +1,512 @@
...
+                let value_ref = vacant.insert(default_value);
+
+                // Call the mutator on the newly inserted value
+                mutator(value_ref);
+
+                // Compute the final size after mutation
+                let final_size = value_ref.total_size_bytes();
+
+                // Add key size and the final value size to heap_size_bytes
+                self.heap_size_bytes += key_size + final_size;
```
Wumpf: suspiciously "well" commented 😁 

nit: I'd tune that down. I consider excessive commenting as llms do (was it? ;)) harmful as it makes it _harder_ to read and comments outdate. e.g. "call the mutator" I can see that

crates/utils/re_byte_size/src/bookkeeping_btreemap.rs
```diff
@@ -0,0 +1,512 @@
...
+            Entry::Vacant(vacant) => {
+                // Insert the default value
+                let key_size = vacant.key().total_size_bytes();
+                let value_ref = vacant.insert(default_value);
+
+                // Call the mutator on the newly inserted value
+                mutator(value_ref);
+
+                // Compute the final size after mutation
+                let final_size = value_ref.total_size_bytes();
```
Wumpf: it's not a final size, it's the value size

crates/utils/re_byte_size/src/bookkeeping_btreemap.rs
```diff
@@ -0,0 +1,512 @@
...
+                // Measure size after mutation
+                let size_after = occupied.get().total_size_bytes();
+
+                // Update heap_size_bytes with the difference
+                self.heap_size_bytes = self.heap_size_bytes - size_before + size_after;
+            }
+        }
+    }
+
+    /// Finds and mutates the last entry smalelr or equalts to the given `key`.
```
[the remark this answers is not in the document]
Wumpf: you can keep that for the human touch 😄

crates/utils/re_byte_size/src/bookkeeping_btreemap.rs
```diff
@@ -0,0 +1,512 @@
...
+        K: Clone,
+    {
+        // Find the last key before the given key (need to clone it to avoid borrow issues)
+        let (key, value) = self.map.range_mut(..=key).next_back()?;
+
+        // Measure size before mutation
+        let size_before = value.total_size_bytes();
+
+        // Call the mutator
+        let ret = mutator(key, value);
```
Wumpf: yeah this is really silly imho, too much comment clutter imho

crates/utils/re_byte_size/src/bookkeeping_btreemap.rs
```diff
@@ -0,0 +1,512 @@
...
+
+        map1.insert(2, "two".to_owned());
+        assert_ne!(map1, map2);
+
+        map2.insert(2, "TWO".to_owned());
+        assert_ne!(map1, map2);
+
+        map2.insert(2, "two".to_owned());
+        assert_eq!(map1, map2);
+    }
```
Wumpf, line 407: pointless to test `Clone` & `PartialEq` since they're generated? Doesn't strictly hurt much but imho it's clutter

crates/utils/re_byte_size/src/bookkeeping_btreemap.rs
```diff
@@ -0,0 +1,512 @@
...
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn total_size_bytes_of_map<K: Ord + SizeBytes, V: SizeBytes>(
```
Wumpf: this is not the total size of the map actually. that's the expected heap size. The map itself takes space as well

crates/utils/re_byte_size/src/bookkeeping_btreemap.rs
```diff
@@ -0,0 +1,512 @@
...
+        // Test as_map returns the correct BTreeMap
+        let inner_map = map.as_map();
+        assert_eq!(inner_map.len(), 2);
+        assert_eq!(inner_map.get(&1), Some(&"one".to_owned()));
+        assert_eq!(inner_map.get(&2), Some(&"two".to_owned()));
+    }
+
+    #[test]
+    fn test_with_pod_types() {
+        // Test with POD (Plain Old Data) types that have zero heap size
```
Wumpf: misleading comment. Yes they have internally zero heap size, but they still add to the heap size of the map

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -87,6 +89,8 @@ pub struct ParentFromChildTransform {
 
 impl SizeBytes for ParentFromChildTransform {
     fn heap_size_bytes(&self) -> u64 {
+        re_tracing::profile_function!();
```
Wumpf, line 92: are all these profile_function still meaningful now that you fixed those methods? we still have issues with too many profile scopes on functions that don't take any time, and this may make it worse?

crates/store/re_tf/src/transform_resolution_cache.rs
```diff
@@ -442,14 +450,10 @@ impl TransformsForChildFrameEvents {
...
-            frame_transforms,
-            pinhole_projections,
-        } = self;
-
-        frame_transforms.extend(time.iter().map(|t| (*t, CachedTransformValue::Cleared)));
-        pinhole_projections.extend(time.iter().map(|t| (*t, CachedTransformValue::Cleared)));
+    fn insert_clears(&mut self, times: &BTreeSet<TimeInt>) {
+        for &time in times {
+            self.insert_clear(time);
+        }
```
Wumpf, line 456: since extend+iter can use the iter's size hint to do preallocation this is can be slower. Would be nice to add an extend to the bookkeeping btreemap. Practicalily not a huge difference though I'd guess

emilk:
> Curious, how slow was it exactly?

many hundreds of milliseconds on a 20GiB demo.rrd

crates/utils/re_byte_size/src/bookkeeping_btreemap.rs
emilk: Yeah, very much Claude's doing :) I can tell it to tone it down in CLAUDE.md :P

crates/store/re_tf/src/transform_resolution_cache.rs
emilk: BTreeMap doesn't do any of that anyhow: https://doc.rust-lang.org/stable/src/alloc/collections/btree/map.rs.html#2383-2384

PR #12507, opened 2026-01-19
Ensure the presence of `sorbet:version` for migrations

grtlr:
### What

This is a small sanity check that ensures changes to `sorbet:version` will show up in our test cases directly.


Wumpf (approved):
I see added decipherable test, I happy

EDIT: ... unlike the dependency graph

grtlr:
@rerun-bot reality-sync

PR #12508, opened 2026-01-19
Single method for creating `DataResult` tree

Wumpf:
While working on getting some of stuff we disabled in https://github.com/rerun-io/rerun/pull/12275 I got really annoyed with. the confusing two step buildup of the dataresult tree: we first create all the `DataResult` and then in a separate step determine all the overrides & properties of those data results.

That is _still_ the case after this PR, buuut it now has a single `build_data_result_tree` method which makes this an "internal affair". Crucially this means that it's not more trivial parallized (I checked, there's measurable, but miniscule perf gains on multi views with many entities) and it gets rid of a really old wart we had in `app.rs` where we'd create the `ViewerContext` twice for borrow-checker reasons.

Commit by commit review encouraged.

### DIFF
--- a/crates/viewer/re_viewport_blueprint/src/view_contents.rs
+++ b/crates/viewer/re_viewport_blueprint/src/view_contents.rs
@@ -257,6 +257,7 @@ impl ViewContents {
     /// Note that this result will not have any resolved overrides. Those can
     /// be added by separately calling `DataQueryPropertyResolver::update_overrides` on
     /// the result.
+    #[expect(clippy::too_many_arguments)]
     pub fn build_data_result_tree(
         &self,
         ctx: &re_viewer_context::StoreContext<'_>,


crates/viewer/re_viewer/src/app_state.rs
```diff
@@ -416,43 +440,6 @@ impl AppState {
                 // Update the viewport. May spawn new views and handle queued requests (like screenshots).
                 viewport_ui.on_frame_start(&ctx);
 
-                let query_results = update_overrides(&ctx, &viewport_ui.blueprint, view_states);
-
-                // We need to recreate the context to appease the borrow checker. It is a bit annoying, but
-                // it's just a bunch of refs so not really that big of a deal in practice.
-                let ctx = ViewerContext {
```
Wumpf, line 423: I'm really happy this thing is finally gone

crates/viewer/re_viewer_context/src/view/view_query.rs
```diff
@@ -22,7 +21,7 @@ pub struct VisualizerComponentMapping {
 }
 
 /// A list of component mappings for a visualizer instruction.
-pub type VisualizerComponentMappings = SmallVec<[VisualizerComponentMapping; 2]>;
```
Wumpf, line 25: these seemed overzealous, blowing up `DataResult` needlessly

grtlr:
@rerun-bot reality-sync

PR #12510, opened 2026-01-19
Store events should always be there for direct callers, regardless of cfg

teh-cmc:
> ## Pull request overview
> 
> This PR introduces a semantic change where store events are now always computed and returned to direct callers, while the `enable_changelog` configuration only controls whether global subscribers receive these events. Additionally, a new `enable_viewer_indexes` parameter is added to `EntityDb` to control whether secondary indexes should be maintained, improving performance for CLI tools that don't need these indexes.
> 
> **Changes:**
> - Added `enable_viewer_indexes` field to `EntityDb` to control maintenance of secondary indexes
> - Modified `ChunkStore` to always compute events but only emit them to subscribers when `enable_changelog` is enabled
> - Updated all `EntityDb::with_store_config` call sites to include the new `enable_viewer_indexes` parameter

image

### DIFF
--- a/crates/store/re_entity_db/src/entity_db.rs
+++ b/crates/store/re_entity_db/src/entity_db.rs
@@ -662,7 +662,12 @@ impl EntityDb {
 
         let mut engine = self.storage_engine.write();
 
+        // The query cache isn't specific to the viewer. Always update it.
         engine.cache().on_events(store_events);
+
+        if !self.enable_viewer_indexes {
+            return;
+        }
 
         let engine = engine.downgrade();
 


crates/top/rerun/src/commands/rrd/compare.rs
```diff
@@ -182,8 +182,10 @@ fn load_chunks(
         stores
             .entry(msg.store_id().clone())
             .or_insert_with(|| {
+                let enable_viewer_indexes = false; // that would just slow us down for no reason
```
teh-cmc, line 185: The fact that we have to explicitly disable it here is a feature: a CLI implementer has no way to know that enabling the store changelog will somehow result in e.g. a time panel datastructure being maintained, even though there are no viewers in sight. This makes it clear what's happening.

crates/store/re_entity_db/src/entity_db.rs
```diff
@@ -637,8 +662,13 @@ impl EntityDb {
 
         let mut engine = self.storage_engine.write();
 
+        // The query cache isn't specific to the viewer. Always update it.
```
teh-cmc, line 665: This means that currently, on `main`, if you instantiate an `EntityDb` with changelog disabled, and then run queries on it, the results you get are effectively undefined.

The relationship between store settings and entitydb behavior, and in general the usage of entitydb in headless contexts, need to be seriously reworked... but for now at least that's a strict improvement.

Issue #10717, opened 2025-07-21
Displaying real-time images results in significant memory growth


WorstCodeWay:
**Describe the bug**

1. rerun viewer is running on windows
2. rerun c++ sdk is running on jetson(aarch64)

Memory increases significantly(from 2.5GB to 7GB) when I logging real time image(about 30Hz, 1280*800) onto rerun viewer with rerun c++ SDK, using following code

```
    // Create a new `RecordingStream` which sends data over gRPC to the viewer process.
    const auto rec = rerun::RecordingStream("rerun_example_cpp");
    rec.connect_grpc("rerun+http://192.168.7.66:9876/proxy").exit_on_failure();

   while(true)
   {
       // get image from stream
       cv::Mat img;
       rec.log(it->first, rerun::Image::from_rgb24(
                        rerun::borrow(img.data, img.total() * img.elemSize()), {img.cols, img.rows}
                    )
          );
          usleep(33000);
   }
```
When the memory is used up, the program is killed by system.

**To Reproduce**
1. run rerun viewer with `rerun listen --bind 0.0.0.0`
2. run my program with rerun c++ sdk

**Expected behavior**

**Screenshots**

**Backtrace**

**Desktop (please complete the following information):**
 - OS: 

**Rerun version**
0.24.0

**Additional context**

grtlr:
Thank you for your report, that's super helpful!

How many images did you log in total? Does the same happen with `rerun.EncodedImage`?

WorstCodeWay:
> How many images did you log in total? 

Thanks for your reply @grtlr 

I'm new to rerun, I cannot tell how many images here, but maybe a scree shot help.

<img width="1916" height="250" alt="Image" src="https://github.com/user-attachments/assets/0229d9e4-7fb0-431f-9fe6-791694967fa4" />

There are 4 cameras capturing data simultaneously, and I want to display them all in `rerun` at the same time. Based on the screenshot, there should be 40 images per camera, totaling 160 images in about a few seconds. Each image is a 1280×800 RGB8 color image.

> Does the same happen with `rerun.EncodedImage`?

No. But I will try this, is there any document about it?

grtlr:
Awesome, thank you for the extra context. I've counted 40 images times 4 entities. Back of the napkin says this should not be 7.5G (40 * 4 * 1280 * 800 * 32 bit ~= 700MB).

We'll look into this 🙏!

[...]

PR #12511, opened 2026-01-19
Apply back-pressure in gRPC sink

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/10717
* Part of  #11024
* Future work: https://github.com/rerun-io/rerun/issues/9812

### What
Encoding arrow data for transport on the wire is unfortunately very slow:
* https://github.com/rerun-io/rerun/issues/9812

If we logged data faster than we could encode it, that lead to a queue growing indefinitely, leading to what looked like a memory leak in the Rerun logging SDK:

image
The 1.7 s here is just the start - it grows and grows forever, leading to more and more memory use and higher and higher latencies.

With this PR, we now apply back-pressure when the IPC encoder cannot keep up. This is ONLY applied to the gRPC sink thought. The rest of the sinks will follow in later PRs.

In practice, what this means is that the first call to `rec.log(…)` will be fast, but after a while it may block for a bit in order for the data encoder to have time to sent what was previously logged.

### In English please
If you connected a fast web-camera and logged images from it as fast as you can, you would get very high latency and high memory usage in the logging process.

After this PR, the logging will slow down to be just fast enough to be able to keep up. Latencies will still grow, but much less:

image

### How to test
Connect a very fast web camera (like the Elgato we have at the office. The normal Macbook camera is NOT enough):

* **server**:  pr rerun --serve-grpc --server-memory-limit 2GB --newest-first
* **producer**: pr py-build && RUST_LOG=debug pr uv run examples/python/live_camera_edge_detection/live_camera_edge_detection.py --connect
* **viewer**: pr rerun --memory-limit 2GB --connect

### DIFF
--- a/crates/top/re_sdk/src/log_sink.rs
+++ b/crates/top/re_sdk/src/log_sink.rs
@@ -38,7 +38,11 @@ impl SinkFlushError {
 pub trait LogSink: Send + Sync + 'static + std::any::Any {
     /// Send this log message.
     ///
-    /// This should optionally block in order to apply backpressure.
+    /// In order to apply backpressure, implementing sinks should block
+    /// the call until there is enough capacity to accept the message.
+    ///
+    /// It is however NOT expected that `send` will _always_ block,
+    /// just when the sink is overwhelmed.
     fn send(&self, msg: LogMsg);
 
     /// Send all these log messages.


emilk:
@rerun-bot reality-sync

emilk:
@rerun-bot reality-sync

emilk:
@rerun-bot reality-sync

crates/top/re_sdk/src/log_sink.rs
```diff
@@ -37,6 +37,8 @@ impl SinkFlushError {
 /// Where the SDK sends its log messages.
 pub trait LogSink: Send + Sync + 'static + std::any::Any {
     /// Send this log message.
+    ///
+    /// This should optionally block in order to apply backpressure.
```
IsseW: `should` -> `can`? I guess it reads a bit differently to the user of the function and the implementer of the trait.


```suggestion
    /// This can optionally block in order to apply backpressure.
```

emilk: I've improved the wording to be more explicit

emilk:
@rerun-bot reality-sync

PR #12513, opened 2026-01-19
Apply backpressure in video decoder

emilk:
### Related
* Part of https://github.com/rerun-io/rerun/issues/11024

### What
Apply backpressure if the video decoder is not fast enough

### DIFF
--- a/crates/viewer/re_viewer_context/src/cache/video_stream_cache/test_player.rs
+++ b/crates/viewer/re_viewer_context/src/cache/video_stream_cache/test_player.rs
@@ -63,6 +63,11 @@ struct TestVideoPlayer {
 
 impl TestVideoPlayer {
     fn play(&mut self, range: Range<f64>, time_step: f64) -> Result<(), VideoPlayerError> {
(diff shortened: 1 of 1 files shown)


crates/utils/re_video/src/lib.rs
```diff
@@ -47,3 +48,11 @@ pub fn enabled_features() -> &'static [&'static str] {
         "nasm",
     ]
 }
+
+/// Create a channel for use with [`new_decoder`].
+pub fn channel<T>(
+    debug_name: impl Into<String>,
+) -> (re_quota_channel::Sender<T>, re_quota_channel::Receiver<T>) {
+    let max_bytes_in_flight = 512 * 1024 * 1024; // TODO(emilk): make configurable
```
Wumpf, line 56: 512MiB is a lot :)
But I guess better to err on the too much side than suddenly getting warnings from some 8k hdr videos

emilk: 512 MiB is a lot less than infinity!

Issue #12495, opened 2026-01-16
Potential Bug with sending Arrows in bulk in different Frames


MaxiMaerz:
**Describe the bug**

When logging an Arrow3D entity with multiple arrows and multiple coordinate frames, they get mixed up.

It could be that I'm misunderstanding how CoordinateFrame should be used in this context, but If you look at the first picture, it is wrong the origin is at [0,0,0] in "A" while the parent frame is "C". Therefore [0,0,0] should be in C which would be 2 away.

I work around with the last code piece, however it is very confusing.

**To Reproduce**
```python
import rerun as rr
rr.init("test", spawn=True)
rr.log("/tf", rr.Transform3D(translation=[1,0,0], quaternion=[0,0,0,1], parent_frame="A", child_frame="B"))
rr.log("/tf", rr.Transform3D(translation=[1,0,0], quaternion=[0,0,0,1], parent_frame="B", child_frame="C"))
rr.log("some_arrow", rr.Arrows3D(origins=[[0,0,0], [0,0,0], [0,0,0]], vectors=[[1,0,0], [0,1,0], [0,0,1]]), rr.CoordinateFrame(frame=["A", "B", "C"]))
```

<img width="2660" height="950" alt="Image" src="https://github.com/user-attachments/assets/3a87f983-c047-46b7-96af-bb690e8cfafc" />

**Expected behavior**
Every Arrow points in one dimension on each Transform (A, B, C):
```python
import rerun as rr
rr.init("test", spawn=True)
rr.log("/tf", rr.Transform3D(translation=[1,0,0], quaternion=[0,0,0,1], parent_frame="A", child_frame="B"))
rr.log("/tf", rr.Transform3D(translation=[1,0,0], quaternion=[0,0,0,1], parent_frame="B", child_frame="C"))
rr.log("some_arrow_A", rr.Arrows3D(origins=[[0,0,0]], vectors=[[1,0,0]]), rr.CoordinateFrame(frame=["A"]), static=True)
rr.log("some_arrow_B", rr.Arrows3D(origins=[[0,0,0]], vectors=[[0,1,0]]), rr.CoordinateFrame(frame=["B"]), static=True)
rr.log("some_arrow_C", rr.Arrows3D(origins=[[0,0,0]], vectors=[[0,0,1]]), rr.CoordinateFrame(frame=["C"]), static=True)
```
[...]

PR #12514, opened 2026-01-19
Warn if an entity has multiple `CoordinateFrame` instances

MichaelGrupp:
Make it transparent for users that we don't support this and pick the first one.

We already do something similar for transforms [here](https://github.com/rerun-io/rerun/blob/d1c73f07c92c053ce8667199ef1c40383916c7dc/crates/viewer/re_view_spatial/src/contexts/transform_tree_context.rs#L76).

Relates to #12495.


MichaelGrupp:
There might be a smarter way to catch this and the SDK should probably be also stricter ideally, but this works with the example from the issue.
image

PR #12505, opened 2026-01-19
Fix detect and track objects example

lucasmerlin:
### What
Fixes examples: https://github.com/rerun-io/rerun/actions/runs/21124713009/job/60744818243

PR #12522, opened 2026-01-20
Fix detect and track objects example

lucasmerlin:
### Related

* try 2 of https://github.com/rerun-io/rerun/pull/12505

### What

Oops, even in the full check it didn't run the detect and track objects example. Is there a good way to test this before merging?

This is the fix claude came up with. Locally it works.

### DIFF
--- a/examples/python/detect_and_track_objects/detect_and_track_objects.py
+++ b/examples/python/detect_and_track_objects/detect_and_track_objects.py
@@ -176,28 +176,12 @@ class Tracker:
     next_tracking_id = 0
     MAX_TIMES_UNDETECTED = 2
 
-    @staticmethod
-    def _create_tracker() -> Any:
-        """Create a CSRT tracker, handling OpenCV API variations."""
-        # Try cv2.legacy first (most common in opencv-contrib-python >4.5)
-        try:
-            return cv2.legacy.TrackerCSRT_create()  # type: ignore[attr-defined]
-        except AttributeError:
-            pass
-        # Try cv2.TrackerCSRT.create() (some versions)
-        try:
-            return cv2.TrackerCSRT.create()  # type: ignore[attr-defined]
-        except AttributeError:
-            pass
-        # Try cv2.TrackerCSRT_create() (older API)
-        return cv2.TrackerCSRT_create()  # type: ignore[attr-defined]
-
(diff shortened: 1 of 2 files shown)


grtlr (approved):

lucasmerlin:
Nightly run to test if this works: https://github.com/rerun-io/rerun/actions/runs/21166101406

lucasmerlin:
Another try: https://github.com/rerun-io/rerun/actions/runs/21172417017

abey79:
Claude's position after digging a little bit. Seems credible. Digging.


  **Why it used to work:**                                                                                                                                                                                             
  - Before the uv switch (`bbc404178d9a` on Jan 5), pixi managed Python environments with better isolation                                                                                                             
  - Now with uv, all examples are installed together via `uv sync --inexact --package <all examples>` into a single shared `.venv`                                                                                     
                                                                                                                                                                                                                       
  **Why it's failing now:**                                                                                                                                                                                            
  1. `rerun-sdk` depends on `opencv-python`                                                                                                                                                                            
  2. `detect_and_track_objects` depends on `opencv-contrib-python`                                                                                                                                                     
  3. When uv installs both packages, `opencv-python` "wins" because it's a direct dependency of rerun-sdk                                                                                                              
  4. `opencv-python` doesn't have the tracking module (`TrackerCSRT`)                                                                                                                                                  
                                                                                                                                                                                                                       
  **Why the fix on Jan 19 "worked" briefly:**                                                                                                                                                                          
  Looking at the timeline, that fix merged but then immediately started failing again in nightly - the fix never actually worked in CI, it just passed some check that didn't run the full example.                    
                                                                                                                                                                                                                       
  **The real fix:** Change `opencv-python` to `opencv-contrib-python` in `rerun_py/pyproject.toml` and `pyproject.toml`, since `opencv-contrib-python` is a superset that includes all of `opencv-python` plus the     
  contrib modules (like tracking).

abey79:
Ah, I can repro the issue locally with the following:

```
rm -rf .venv
pixi run py-build
pixi run py-sync-examples
uv run detect_and_track_objects`
```

abey79:
image

lucasmerlin:
Closed via https://github.com/rerun-io/rerun/pull/12537

PR #12526, opened 2026-01-20
Incorporate native type check into `AnyPhysicalDatatype`

Wumpf:
### Related

* Part of RR-3317 / RR-3318

### What

Immediate effect is that with component mappings enabled, we no longer _recommend_ the time series view unless there's native scalars


Wumpf:
@rerun-bot reality-sync

PR #12527, opened 2026-01-20
Remove inner table

timsaucer:
This is an example of how we might just pass the table provider into the other.


lucasmerlin:
Thanks!

PR #12531, opened 2026-01-20
Handle error when trying to plot static data

Wumpf:
Previously, we'd crash in debug (due to overflows/underflows) and show a nonsensical plot in release builds. Now we report the problem through the visualizer error mechanism.

This can be hit when simply logging `Scalars` as static, but it becomes more important we're moving towards allowing to plot any data! 

Screenshot with early stage experimental component mapping, here pointing to a column with static data:
image

### DIFF
--- a/crates/viewer/re_view_time_series/src/lib.rs
+++ b/crates/viewer/re_view_time_series/src/lib.rs
@@ -114,7 +114,7 @@ enum LoadSeriesError {
     ViewPropertyQuery(ViewPropertyQueryError),
     EntitySpecificVisualizerError {
         entity_path: EntityPath,
-        error: String,
+        err: String,
     },
 }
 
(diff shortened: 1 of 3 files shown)


crates/viewer/re_view_time_series/src/util.rs
```diff
@@ -123,10 +123,20 @@ pub fn points_to_series(
...
+        return Ok(());
+    }
+
+    // Filter out static times if any slipped in.
+    // It's enough to check the first one since an entire column has to be either temporal or static.
+    if let Some(first) = points.first()
+        && first.time == re_log_types::TimeInt::STATIC.as_i64()
+    {
+        return Err("Can't plot data that was logged statically in a time series since there's no temporal dimension.".to_owned());
     }
```
grtlr, line 140: It's good to catch this here anyways, but I wonder if being interested in only non-static data is something that we should capture in the `VisualizerQueryInfo` and pass along to the `VisualizerEntitySubscriber`. That way it would not count towards heuristics either?

Wumpf: huh yeah we could actually encode that in our requirements, creating a note on the project
since staticness is regarded irreversible, our visualizability subscriber can make this assessment!

Wumpf:
@rerun-bot reality-sync

PR #12532, opened 2026-01-20
Early stage experimental component mapping & component casting in plots

Wumpf:
### Related

* Fixes RR-3318 (with caveat of being feature flag gated)
* Fixes RR-3317 (with caveat of being feature flag gated)

### What

We've previously landed the feature flag, but built back some more things to disabled things. With this PR experimental component mapping becomes available again, but we're still quite a bit out on making it non-experimental.

This applies various fixes for experimental component mappings:
* re-enable mapping under the feature flag
* fix heuristic mapping generation for plots
* fix experimental component mapping ui showing nothing and getting overall confused in the presence of component casts
    * it still has a lot of issues but no longer fails catastrophically as it did meanwhile 

Still need to refine the ui and datastructures for component mappings and add tests. Also more work on heuristics is needed as well.

Screenshot of `cargo run -p test_any_scalar` with enabled experimental flag on this branch (no manual viewer config needed!)
image

### DIFF
--- a/crates/viewer/re_view/src/query.rs
+++ b/crates/viewer/re_view/src/query.rs
@@ -141,11 +141,9 @@ pub fn latest_at_with_blueprint_resolved_data<'a>(
     let component_mappings_hash = if let Some(visualizer_instruction) = visualizer_instruction {
         for (target, selector) in &visualizer_instruction.component_mappings {
             if let Some(chunk) = results.components.remove(selector) {
-                let chunk = std::sync::Arc::new(
-                    chunk.with_renamed_component(*selector, *target),
-                )
-                .to_unit()
-                .expect("The source chunk was a unit chunk.");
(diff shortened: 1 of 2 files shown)


crates/viewer/re_viewer_context/src/view/view_query.rs
```diff
@@ -38,7 +33,9 @@ pub struct VisualizerInstruction {
     /// like `Visible`, `Interactive` or transform components.
     pub component_overrides: IntSet<ComponentIdentifier>,
 
-    /// List of component mapping pairs.
+    /// Component mappings from target to source (selector).
+    ///
+    /// Keys are target components, values are source components (selectors).
```
grtlr:
```suggestion
    /// Keys are target components, values describe where to source components (selectors).
```

Wumpf:
@rerun-bot reality-sync

PR #12533, opened 2026-01-20
Introduce new helper to only ensure the uv shims exist independently of pyo3-build-cfg

jleibs:
### What

This makes it possible to install the uv-shims in an environment that doesn't need the pyo3-config fixes


jleibs:
@rerun-bot reality-sync

jleibs:
@rerun-bot reality-sync

PR #12534, opened 2026-01-20
Prune my zombie

ntjohnson1:
### Related
#10828

### What
Ran the zombie script:
* Fix one incorrectly formatted todo
* TLDR: Just moved my todo to a comment
   * Looked to see if my stacker PR actually let us get rid of the llvm tools. Unfortunately, there are other Wasm targets that fail on the same error. Potentially we could look to try and land the fix into `cc` to see if that catches this issue across the ecosystem. But might not be worth is since a full llvm compiler is needed for the wasm32 target anyway.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -575,9 +575,7 @@ buf = "1.*" # not available for linux-aarch64
 buf = "1.*" # not available for linux-aarch64
 libgfortran5 = ">=14" # Fixes issues with OpenCV
 meilisearch = "1.5.1.*" # not available for linux-aarch64
-# Various Wasm targets fail on mac without
-# llvm-ar
-llvm-tools = "16.0.6.*"
+llvm-tools = "16.0.6.*" # Various Wasm targets fail on mac without llvm-ar
 
 [target.win-64.dependencies]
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -574,7 +566,9 @@ buf = "1.*" # not available for linux-aarch64
 buf = "1.*" # not available for linux-aarch64
 libgfortran5 = ">=14" # Fixes issues with OpenCV
 meilisearch = "1.5.1.*" # not available for linux-aarch64
-llvm-tools = "16.0.6.*" # TODO(rust-lang/stacker#127)
+# Various Wasm targets fail on mac without
+# llvm-ar
+llvm-tools = "16.0.6.*"
```
emilk: putting the comment on the same line makes it less likely to be outdated

```suggestion
llvm-tools = "16.0.6.*" # Various Wasm targets fail on mac without llvm-ar
```

ntjohnson1: Done for both locations. I was just annoyed that the cross OS wasm build support wasn't more robust and was venting on the cargo location.

ntjohnson1:
@rerun-bot reality-sync

PR #12537, opened 2026-01-21
Yet another attempt to fix detect-and-track example build

abey79:
### Related

* superseedes https://github.com/rerun-io/rerun/pull/12522

### What

Yet another attempt, the genesis of which is documented in #12522's comments.

This approach is still extremely fragile though. The root cause is that detect-and-track uses stuff from `opencv-contrib-python`, which is supposedly a super-set of `opencv-python`. Both race for the `cv2` module :old-man-yell-at-dumpster-fire:. What the PR does is to help contrib with the race _in controlled environment_.

The correct approach would be to migrate everything to `opencv-contrib-python`, but we can't control transitive dependencies. In particular, `nuscenes-devkit` pulls `opencv-python`.

One proposal would be to add a defensive  `"opencv-python>9999"` dependency and deal with the consequences, but this would likely mean removing some examples (or other some such headache).

### DIFF
--- a/crates/top/re_sdk/src/lenses/semantic.rs
+++ b/crates/top/re_sdk/src/lenses/semantic.rs
@@ -1,0 +1,144 @@
+//! Semantic array transforms for concrete applications.
+//!
+//! Note: These should not be exposed as part of the public API, but rather wrapped in [`super::Op`].
+
+use std::marker::PhantomData;
+use std::sync::Arc;
+
+use arrow::array::{
+    Array as _, ArrowNativeTypeOp as _, GenericBinaryArray, GenericListArray, Int64Array,
+    OffsetSizeTrait, StringArray, StructArray, UInt32Array, UInt32Builder,
+};
+use arrow::datatypes::{DataType, Field, Int32Type, Int64Type};
+use arrow::error::ArrowError;
+use re_sdk_types::components::VideoCodec;
+
+use re_arrow_combinators::cast::DowncastRef;
+use re_arrow_combinators::reshape::GetField;
+use re_arrow_combinators::{Error, Transform};
+
+/// Converts binary arrays to list arrays where each binary element becomes a list of `u8`.
+///
+/// The underlying bytes buffer is reused, making this transformation almost zero-copy.
+#[derive(Clone, Debug, Default)]
+pub struct BinaryToListUInt8<O1: OffsetSizeTrait, O2: OffsetSizeTrait = O1> {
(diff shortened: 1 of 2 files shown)


abey79:
Nightly attempt: https://github.com/rerun-io/rerun/actions/runs/21203108573

edit: Build example ✅

pyproject.toml
```diff
@@ -76,7 +76,7 @@ snippets = ["rerun-sdk", "av", "pandas"]
 examples = [
   # External deps used by examples
   "polars>=0.12.0",
-  "opencv-python>=4.6",
+  "opencv-contrib-python>=4.6", # contrib is a superset; needed for detect_and_track_objects tracking module
```
lucasmerlin, line 79: This means everyone using rerun sdk will get the larger contrib version now, could that be a problem?

abey79: No because this is the `examples` group. It's not part of the core rerun dependencies:

```
❯ pixi run py-build                                                                                                                                                                                                  
  ✨ Pixi task (py-build in default): uv run maturin develop --uv --manifest-path rerun_py/Cargo.toml --extras=tests,datafusion                                                                                        
  ⠁Using CPython 3.11.13 interpreter at: .pixi/envs/default/bin/python3.11                                                                                                                                             
  Creating virtual environment at: .venv                                                                                                                                                                               
  Installed 175 packages in 522ms                                                                                                                                                                                      
  🍹 Building a mixed python/rust project                                                                                                                                                                              
  🔗 Found pyo3 bindings with abi3 support                                                                                                                                                                             
  🐍 Found CPython 3.11 at /Users/hhip/src/rerun/.venv/bin/python                                                                                                                                                      
  📡 Using build options editable-profile, locked, config from pyproject.toml                                                                                                                                          
  Resolved 31 packages in 146ms                                                                                                                                                                                        
  Installed 5 packages in 293ms                                                                                                                                                                                        
  + mpmath==1.3.0                                                                                                                                                                                                      
  + networkx==3.6.1                                                                                                                                                                                                    
  + opencv-python==4.13.0.90                                                                                                                                                                                           
  + sympy==1.14.0                                                                                                                                                                                                      
  + torch==2.9.1                       
```

And this is because of the `test` extra. I believe opencv wouldn't even show up without it (which is what users would normally do.

lucasmerlin: Oh I see, thanks for the explanation!

lucasmerlin:
@rerun-bot reality-sync

PR #12539, opened 2026-01-21
Move the often-used `TrackingAllocator` into `re_byte_size`

emilk:
This is to unlock exact tests of size measurements in `re_byte_size`, so the size estimations for things like `BTreeMap` and `HashMap` can be improved.


emilk:
@rerun-bot reality-sync

PR #12543, opened 2026-01-21
Fix icon colors in light theme

oxkitsune:
### What

This fixes two small icons that did not apply the correct theme color, causing them to be hard to see:

- Transform frame id hierarchy view in selection panel:
   image

- Info icon for memory limit in the entity db view:
   image


oxkitsune:
@rerun-bot reality-sync

oxkitsune:
> This does seem like a footgun though. Maybe when none is provided we should set the tint to text_color in small_icon?

I think that would be good, I don't know what kind of side effects that has (if any). I can try doing that, and seeing whether any comparisons fail.

oxkitsune:
@rerun-bot reality-sync

oxkitsune:
Merged by sync bot: https://github.com/rerun-io/rerun/commit/edac1a15d6aec2e344017bcf9b8148af8e6389bd

PR #12544, opened 2026-01-21
Improve `re_byte_size` estimates

emilk:
This adds tests to `re_byte_size` to verify our assumptions. This in turn lead to some improvements in the accuracy of our size estimates. In particular, our estimates were over-optimistic (low) for many types, in particular BTreeMap/Set.

HashSet and HashMaps are no much more accurate, with accuracy improvements between 10-50%.

BTreeMap/Set is much harder to accurately estimate, since their final heap size depends a lot on how data were inserted. Still, the estimation is more accurate now (and more pessimistic!).

You can see the HashSet/Map improvements here: https://github.com/rerun-io/rerun/pull/12544/changes/32d5d204b1f1aee89c8bd6e396326594c65357ce#diff-1a3fbd8d3f8a5d3955784cc0a11383432b2a14fbe0f178319df6d9527ebf7ae8


crates/utils/re_byte_size/src/bookkeeping_btreemap.rs
```diff
@@ -65,17 +65,15 @@ where
 
         match self.map.entry(key) {
             Entry::Vacant(vacant) => {
-                let key_size = vacant.key().total_size_bytes();
+                self.heap_size_bytes += vacant.key().heap_size_bytes();
```
grtlr: I know, it's not part of this code. But I find the `heap_size_bytes` confusing. 

What if you have nested data structures? In that case the inner structures would live entirely on the heap too (just not at the same spot)?

crates/utils/re_byte_size/src/bookkeeping_btreemap.rs
```diff
@@ -110,15 +107,15 @@ where
     pub fn insert(&mut self, key: K, value: V) -> Option<V> {
         // In a BTreeMap, the keys and values themselves are stored on the heap,
         // so we count their total size (stack + heap).
-        let new_key_size = key.total_size_bytes();
-        let new_value_size = value.total_size_bytes();
+        let new_key_size = key.heap_size_bytes();
+        let new_value_size = value.heap_size_bytes();
```
grtlr, line 122: The comment above seems to be outdated (and it reflects what I would expect). So the change does not really make sense to me. Can we update it please?

_This is the requested change._ 😁

crates/utils/re_byte_size/tests/test_byte_sizes.rs
```diff
@@ -0,0 +1,162 @@
...
+        (0..len as u32)
+            .map(|i| (i, i as u8))
+            .collect::<BTreeMap<u32, u8>>()
+    });
+    run_many_sizes(&mut lines, "BTreeMap<u64, String> (collect)", |len| {
+        (0..len as u64)
+            .map(|i| (i, format!("value_{i}")))
+            .collect::<BTreeMap<u64, String>>()
+    });
+    run_many_sizes(&mut lines, "BTreeMap<u64, String> (insert)", |len| {
```
grtlr: Iirc, we have several places of nested `BTreeMaps` in the viewer, maybe add such a test case? It might highlight what I've written above.

emilk: It shouldn't matter - we already test a non-pod-type (String). Exactly _which_ non-pod-type we put in there should be 🤷

grtlr (approved):

emilk:
@rerun-bot reality-sync

PR #12546, opened 2026-01-21
Add coarse check that python apis end up in rendered docs and fix missing

ntjohnson1:
### Related
RR-3410

### What
Trying to reference our Dynamic Archetype for someone and saw it is missing from our API page. When I started to look deeper quite a few things are missing.
* Added a coarse check to see if we include add submodules at a depth of 1
   * Also add ability to list multiple submodules for one `Section` in our docs generation
* There are quite a few small files that are listed as public but not really referred to anywhere directly and we don't have them in the documentation. I moved a bunch of those to internal and opt-ed out internal things from our docs check.
*  We don't publish `rerun_notebook` docs 😢 so there is one remaining warning that we fail to cross link that.

We only publish python docs on release AFAICT. So please run `pixi run py-docs-serve` to review the end result and confirm it builds for you too.

### DIFF
--- a/rerun_py/docs/gen_common_index.py
+++ b/rerun_py/docs/gen_common_index.py
@@ -76,48 +76,49 @@ def all_submodules(max_depth: int | None = None, ignore_hidden: bool = True) -> 
         ignore_hidden: If True, skip modules starting with underscores (e.g., _baseclasses, __main__).
 
     """
-    import pkgutil
 
     rerun_package_path = Path(__file__).parent.parent.parent.joinpath("rerun_py/rerun_sdk/rerun")
 
-    # Add the parent directory to sys.path temporarily so we can import rerun
-    parent_path = str(Path(__file__).parent.parent.parent.joinpath("rerun_py/rerun_sdk"))
-    if parent_path not in sys.path:
-        sys.path.insert(0, parent_path)
+    # Walk the filesystem directly instead of importing packages
+    submodules = []
 
-    try:
-        # Import the rerun package
-        import rerun
+    # We do this because we build and test our docs without rerun_bindings built.
+    def _walk_package_dir(directory: Path, base_path: Path, current_relative_path: str = "") -> None:
+        """Recursively walk a Python package directory to find submodules."""
+        if not directory.is_dir():
+            return
 
-        submodules = []
+        # Check if this directory is a Python package (has __init__.py)
+        init_file = directory / "__init__.py"
+        if not init_file.exists():
+            return
 
-        # Walk through all submodules in the rerun package
-        for _, modname, _ in pkgutil.walk_packages(rerun.__path__, rerun.__name__ + "."):
-            # Remove the 'rerun.' prefix to get the relative module name
-            relative_name = modname[len("rerun.") :]
-
+        # If we're not at the root, add this as a submodule
+        if current_relative_path:
             # Skip hidden modules if requested
             if ignore_hidden:
-                # Check if any part of the module path starts with underscore
-                parts = relative_name.split(".")
+                parts = current_relative_path.split(".")
                 if any(part.startswith("_") for part in parts):
-                    continue
+                    return
 
             # Check depth if max_depth is specified
             if max_depth is not None:
-                depth = relative_name.count(".") + 1
+                depth = current_relative_path.count(".") + 1
                 if depth > max_depth:
-                    continue
+                    return
 
-            submodules.append(relative_name)
+            submodules.append(current_relative_path)
 
-        assert len(submodules) > 0, f"Found no submodules in {rerun_package_path}"
-        return sorted(submodules)
+        # Recursively walk subdirectories
+        for item in directory.iterdir():
+            if item.is_dir() and not item.name.startswith("."):
+                new_relative_path = current_relative_path + "." + item.name if current_relative_path else item.name
+                _walk_package_dir(item, base_path, new_relative_path)
 
-    finally:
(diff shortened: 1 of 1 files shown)


ntjohnson1:
Move to draft because we are missing a bunch of docs stuff 😢

rerun_py/rerun_sdk/rerun/notebook.py
```diff
@@ -454,10 +454,6 @@ def set_time_ctrl(
             Whether to start playing from the specified time point. Defaults to paused.
         timeline:
             The name of the timeline to switch to. If not provided, time will remain on the current timeline.
```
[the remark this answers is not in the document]
ntjohnson1: These arguments were previously removed (check the code) we just forgot to cleanup the doc string.

docs/content/reference/migration/migration-0-29.md
```diff
@@ -68,6 +68,16 @@ The deprecated `rerun.catalog` APIs that were marked for removal in 0.28 have no
 
 Please refer to the [0.28 migration guide section on catalog API overhaul](migration-0-28.md#python-sdk-catalog-api-overhaul) for more details on the new API patterns.
 
+### Multiple internal submodules were move to properly mark internal
```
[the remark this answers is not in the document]
ntjohnson1: This is a good catch. I'll wait to commit until human review to avoid extra CI

rerun_py/docs/gen_common_index.py
```diff
@@ -63,18 +66,77 @@ def all_archetypes() -> list[str]:
...
+
 @dataclass
 class Section:
     title: str
     sub_title: str | None = None
     func_list: list[str] | None = None
     class_list: list[str] | None = None
     gen_page: bool = True
-    mod_path: str = "rerun"
+    mod_path: list[str] | None = None
```
abey79, line 131: Not for this PR, but this draw my attention to the module path being displayed in the documentation (e.g. `rerun.dynamic_archetype` for `DynamicArchetype`), which really isn't very useful for symbols that are re-exported to the top level and expected to be used as such.

rerun_py/docs/gen_common_index.py
```diff
@@ -388,9 +446,21 @@ class Section:
             "VectorDistanceMetricLike",
         ],
     ),
+    Section(
+        title="Server",
+        show_tables=True,
+        mod_path=["rerun.server"],
+        show_submodules=True,
+    ),
```
abey79, line 454: oh, did we not have that? 🤦🏻

rerun_py/docs/gen_common_index.py
```diff
@@ -429,40 +523,41 @@ class Section:
         ],
         class_list=["LoggingHandler", "MemoryRecording", "GrpcSink", "FileSink"],
     ),
-    Section(
-        title="Utilities",
-        show_tables=False,
-        mod_path="rerun.utilities",
-        show_submodules=True,
-    ),
-    # We don't have any experimental apis right now, but when you add one again, you should add this here:
```
abey79, line 438: that worked out great didn't it 😅

rerun_py/docs/gen_common_index.py
```diff
@@ -546,22 +641,23 @@ def make_slug(s: str) -> str:
...
-                if not section.default_filters:
-                    fd.write("      filters: []\n")
-                if section.show_submodules:
-                    fd.write("      show_submodules: True\n")
+                for mod_path in section.mod_path:
+                    fd.write(f"::: {mod_path}\n")
+                    fd.write("    options:\n")
+                    fd.write("      show_root_heading: True\n")
+                    fd.write("      heading_level: 3\n")
+                    fd.write("      members_order: alphabetical\n")
```
abey79, line 649: Re my other comment, I guess we could just remove this header? What's your thoughts on that?

abey79 (approved):
great stuff, thx

rerun_py/docs/gen_common_index.py
ntjohnson1: I'm on board with the re-export thing but it's a bit of a can of worms.

I think we really need a pass over this whole thing since it's a bit of a mess. IMO the discussion is connected to the discussion on potentially breaking up our package into multiple sub-packages.

Also we should have a test that does more intelligent inspection that considers our re-export.

Also re-exporting nearly everything to top level for a basically flat namespace seems like something revisiting eventually

ntjohnson1: Or did you mean removing the header in this PR as a short term solution? The problem there is we don't have great tooling to check if it got re-exported or not

ntjohnson1:
@rerun-bot reality-sync

rerun_py/docs/gen_common_index.py
abey79: Two discussions here, none needs addressing now.

1) Should we continue spamming the `rerun` namespace as we've done so far?

Probably not. Matter of fact, we haven't much lately (see `rerun.server`, `rerun.catalog`, etc.). As you know, my position is that we should go further and actually split packages.

2) Should we display the _implementation_ sub-module in the docs?

We should obviously not, so long as everything is re-exported in `rerun`. Your prepending of _ for many submodules mostly address that concern.

There remain some hybrid dumpster fires, e.g. https://ref.rerun.io/docs/python/stable/common/errorutils/
`rerun.error_utils` is displayed as header., but some symbols are re-exported to `rerun`. We should be consistent here.

PR #12547, opened 2026-01-21
Fix Windows not listening on ipv4 by default

Wumpf:
### Related

* Fixes RR-3402

### What

We recently started binding by default to `::` instead of `0.0.0.0`. On Linux & Mac this will actually listen for both ipv4 & ipv6 addresses. Windows also has this ability but it's opt-in and not exposed on Rust's `TcpListener`. This then effectively causes _all our examples_ to no longer work out of the box since they try to connect to `127.0.0.1` which the (Windows) Viewer would then no longer accept.

This PR works around this!

For details see
* https://github.com/rust-lang/rust/issues/130668

Tested on my local windows machine only so far.

### DIFF
--- a/crates/top/rerun/src/commands/entrypoint.rs
+++ b/crates/top/rerun/src/commands/entrypoint.rs
@@ -91,7 +91,7 @@ struct Args {
 
     /// What bind address IP to use.
     ///
-    /// Defaults to `::` (all interfaces, IPv6 and IPv4).
+    /// `::` will listen on all interfaces, IPv6 and IPv4.
     #[clap(long, default_value = "::")]
     bind: IpAddr,
 
--- a/docs/content/reference/cli.md
+++ b/docs/content/reference/cli.md
@@ -41,6 +41,8 @@ The Rerun command-line interface:
 * `--bind <BIND>`
 > What bind address IP to use.
 >
+> `::` will listen on all interfaces, IPv6 and IPv4.
+>
 > [Default: `::`]
 
 * `--memory-limit <MEMORY_LIMIT>`


crates/store/re_grpc_server/src/lib.rs
```diff
@@ -154,17 +154,52 @@ async fn serve_impl(
     message_proxy: MessageProxy,
     shutdown: shutdown::Shutdown,
 ) -> anyhow::Result<()> {
-    let tcp_listener = TcpListener::bind(addr).await?;
-    let incoming = TcpIncoming::from(tcp_listener).with_nodelay(Some(true));
+    // TODO(rust-lang/rust#130668): When listening on `::` we want to listen to both ipv6 `::` and ipv4 `0.0.0.0`
```
MichaelGrupp, line 157: Could be worth to add this note also to the user-facing arg description in `entrypoint.rs`. I don't think everyone knows that.

MichaelGrupp (approved):
Looks clean to me!

crates/store/re_grpc_server/src/lib.rs
Wumpf: I figure if someone actually puts in explicit `--bind` argument then they expect to bind to that exact ip address. I'm mostly worried about our default out of the box behavior being flexible.

But I guess it can't hurt...

Wumpf:
@rerun-bot reality-sync

crates/store/re_grpc_server/src/lib.rs
MichaelGrupp: Yeah was more about someone reading `--help`. LGTM!

Wumpf:
@rerun-bot reality-sync

Wumpf:
(did a manual sync by now)

PR #12548, opened 2026-01-21
Source selector for visualizer components

aedm:
### Related

Closes RR-3338
Ref RR-3382

### What

- Add source selector to visualizer component UI.
- Removes the experimental flag for component mapping.

image


aedm:
@rerun-bot reality-sync

aedm:
@rerun-bot reality-sync

tests/rust/test_any_scalar/src/main.rs
```diff
@@ -85,6 +85,28 @@ fn log_scalar_data(rec: &RecordingStream) -> anyhow::Result<()> {
             .with_component_from_data("sigmoid", sigmoid_multi_array);
 
         rec.log("float32/multi", &float32_multi_archetype)?;
+
+        // 6. Log using DynamicArchetype with all integer types on same entity
```
Wumpf, line 89: that was me and tbh it kinda ruins the example 😅 
should probably undo again

crates/viewer/re_selection_panel/src/visualizer_ui.rs
```diff
@@ -466,28 +366,31 @@ fn visualizer_components(
                                 ctx.recording(),
                                 &data_result.entity_path,
                                 unmapped_component_descr,
-                                None,
-                                raw_fallback.as_ref(),
+                                current_value_row_id,
+                                raw_current_value.as_ref(),
```
Wumpf, line 370: this seems wrong we're in the else branch of `if let Some(result_default) = result_default {` therefore we have to use the fallback, no?

crates/viewer/re_selection_panel/src/visualizer_ui.rs
```diff
@@ -578,124 +484,156 @@ fn source_component_ui(
...
+fn source_component_ui(
+    ctx: &ViewContext<'_>,
+    ui: &mut egui::Ui,
+    entity_components_with_datatype: &[(ComponentIdentifier, DataType)],
+    component_descr: &ComponentDescriptor,
+    instruction: &VisualizerInstruction,
+    query_info: &VisualizerQueryInfo,
+    raw_override: &Option<ArrayRef>,
+) {
+    // TODO(aedm): In some cases, there should be no separate source selector. Eg. when the value is as enum:
```
Wumpf, line 513: ah, good point. didn't occur to me so far

crates/viewer/re_selection_panel/src/visualizer_ui.rs
```diff
@@ -563,6 +465,10 @@ fn source_component_ui(
         false
     };
 
+    if is_required_component {
```
Wumpf: this early out here doesn't make any sense. It means that you can't select custom components for required components at all. Also note that there's lotsa special handling on `is_required_component` just a few lines below

On `cargo run -p test_any_scalar` this leads to this situation:

Image

Wumpf: This return is also the reason we get offered `Scalars::scalars` which doesn't make sense because the entity doesn't have it

crates/viewer/re_selection_panel/src/visualizer_ui.rs
```diff
@@ -578,124 +484,156 @@ fn source_component_ui(
...
+                });
+            },
+        ));
     });
 }
 
+/// Determines which component source is currently active.
+///
+/// If none is encoded in the visualizer instruction, we apply the same logic as `re_view::query`.
+/// TODO(andreas): Can we deduplicate this somehow?
```
Wumpf, line 568: I think I have an answer to that by now - yes we can fairly easily. But that landed recently and might be still a bit hairy. I'm making a note to care care of it

tests/rust/re_integration_test/src/viewer_section.rs
```diff
@@ -166,6 +168,18 @@ impl<'a, 'h: 'a> ViewerSection<'a, 'h> {
         self.harness.run();
     }
 
+    /// Toggles the collapse triangle of a hierarchical list item. Eg. visualizer components in the selection panel.
+    pub fn toggle_nth_hierarchical_list(&mut self, label: &str, index: usize) {
```
Wumpf, line 172: fancy!

tests/rust/re_integration_test/tests/source_component_test.rs
Wumpf, line 1: nice!

Wumpf (changes requested):
great to have this covered by a integration test. However, it seems that you accidentally broke custom components on required components, curious how you ended up with this ;-)

crates/viewer/re_selection_panel/src/visualizer_ui.rs
aedm: We use the fallback value, see above. `raw_current_value` is `raw_fallback.clone()` in this case.

crates/viewer/re_selection_panel/src/visualizer_ui.rs
aedm: Yes, this is silly indeed. :)

crates/viewer/re_selection_panel/src/visualizer_ui.rs
Wumpf: ah got it. makes sense!

Wumpf (approved):

PR #12554, opened 2026-01-21
Make reality-sync incremental on pushes for rerun org members

jleibs:
### Related

RR-3389


jleibs:
@rerun-bot reality-sync

PR #12557, opened 2026-01-22
Don't gc chunks we last prefetched

IsseW:
### Related

- Part of RR-3349

### What

Stops the garbage collector from removing chunks the prefetcher would just fetch again.

### DIFF
--- a/crates/store/re_entity_db/src/rrd_manifest_index.rs
+++ b/crates/store/re_entity_db/src/rrd_manifest_index.rs
@@ -536,31 +536,26 @@ fn warn_when_editing_recording(store_kind: StoreKind, warning: &str) {
 impl re_byte_size::SizeBytes for RrdManifestIndex {
     fn heap_size_bytes(&self) -> u64 {
         let Self {
-            chunk_intervals,
-            chunk_promises: _, // TODO(emilk)
             entity_has_static_data,
             entity_has_temporal_data_on_timeline,
             entity_tree,
-            manifest_row_from_chunk_id,
             manifest,
             native_static_map,
             native_temporal_map,
             remote_chunks,
-            static_chunk_ids,
+            chunk_prioritizer,
             timelines,
             full_uncompressed_size: _,
         } = self;
 
-        chunk_intervals.heap_size_bytes()
-            + entity_has_static_data.heap_size_bytes()
+        entity_has_static_data.heap_size_bytes()
             + entity_has_temporal_data_on_timeline.heap_size_bytes()
             + entity_tree.heap_size_bytes()
-            + manifest_row_from_chunk_id.heap_size_bytes()
             + manifest.heap_size_bytes()
             + native_static_map.heap_size_bytes()
             + native_temporal_map.heap_size_bytes()
             + remote_chunks.heap_size_bytes()
-            + static_chunk_ids.heap_size_bytes()
+            + chunk_prioritizer.heap_size_bytes()
             + timelines.heap_size_bytes()
     }
 }


IsseW:
@rerun-bot reality-sync

crates/store/re_chunk_store/src/gc.rs
```diff
@@ -83,13 +87,18 @@ impl GarbageCollectionOptions {
...
             perform_deep_deletions: false,
         }
     }
 
     /// If true, we cannot remove this chunk.
-    pub fn is_chunk_temporally_protected(&self, chunk: &Chunk) -> bool {
+    pub fn is_chunk_protected(&self, chunk: &Chunk) -> bool {
+        if self.protected_chunks.contains(&chunk.id()) {
+            return true;
+        }
```
emilk, line 99: This doesn't seem to implement the "or that are derived from a chunk in the rrd manifest with this id" part of the docstring?

crates/store/re_chunk_store/src/lineage.rs
```diff
@@ -407,6 +407,35 @@ impl ChunkStore {
         roots
     }
 
+    pub fn physical_leaf_chunks_for<'a>(
```
emilk: Docstring!

crates/store/re_chunk_store/src/lineage.rs
```diff
@@ -407,6 +407,35 @@ impl ChunkStore {
...
+        &'a self,
+        chunk_id: &'a ChunkId,
+    ) -> impl Iterator<Item = ChunkId> {
+        let is_physical = |c: &&ChunkId| self.chunks_per_chunk_id.contains_key(c);
+
+        if is_physical(&chunk_id) {
```
emilk, line 419: It's physical sure, but is it also a "leaf"? What is the difference really? I guess that's something a docstring could explain. Maybe these terms should be explain at the `lineage.rs` module-level though

crates/store/re_chunk_store/src/query.rs
```diff
@@ -673,11 +673,18 @@ impl QueryResults {
...
+
+            if !this.missing.is_empty() {
+                tracker.missing.extend(this.missing.iter().copied());
+            }
+
+            if !this.chunks.is_empty() {
+                tracker
+                    .used_physical
+                    .extend(this.chunks.iter().map(|c| c.id()));
+            }
```
emilk: looks like the `if` was there to prevent an unnecessary call to `write()`, but now that we always want to write _something_ to the tracker, we might as well go with this simpler code:

```suggestion
        {
            let mut tracker = store.queried_chunk_id_tracker.write();
            tracker.missing.extend(this.missing.iter().copied());
            tracker.used_physical.extend(this.chunks.iter().map(|c| c.id()));
```

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
```diff
@@ -0,0 +1,347 @@
...
+
+#[derive(Default)]
+#[cfg_attr(feature = "testing", derive(Clone))]
+pub struct ChunkPrioritizer {
+    /// All loaded chunks that are 'in' the memory limit.
+    ///
+    /// These chunks are protected from being gc'd.
+    pub(super) in_limit_chunks: HashSet<ChunkId>,
+
+    chunk_promises: ChunkPromises,
```
emilk, line 189:
```suggestion
    /// Chunks that are in the progress of being downloaded.
    chunk_promises: ChunkPromises,
```

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
```diff
@@ -0,0 +1,347 @@
...
+
+    pub fn chunk_promises(&self) -> &ChunkPromises {
+        &self.chunk_promises
+    }
+
+    pub fn chunk_promises_mut(&mut self) -> &mut ChunkPromises {
+        &mut self.chunk_promises
+    }
+
+    pub fn take_protected_chunks(&mut self) -> HashSet<ChunkId> {
```
emilk, line 287: docstring

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
```diff
@@ -0,0 +1,347 @@
...
+        std::mem::take(&mut self.in_limit_chunks)
+    }
+
+    pub fn prioritize_and_prefetch(
```
emilk, line 346: docstring

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
```diff
@@ -0,0 +1,347 @@
...
+                Some(ChunkInfo {
+                    state: LoadState::Loaded,
+                    ..
+                })
+                | None => {
+                    {
+                        let Some(chunk) = store.physical_chunk(&chunk_id) else {
+                            re_log::warn_once!("Couldn't get loaded chunk from chunk store");
+                            continue;
+                        };
```
emilk, line 440: The message here is not really accurate for the `| None` case

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    pub fn prioritize_and_prefetch(
+        &mut self,
+        store: &ChunkStore,
+        options: &ChunkPrefetchOptions,
+        load_chunks: &dyn Fn(RecordBatch) -> ChunkPromise,
+        manifest: &RrdManifest,
+        remote_chunks: &mut HashMap<ChunkId, ChunkInfo>,
+    ) -> Result<(), PrefetchError> {
```
emilk, line 353:
```suggestion
    ) -> Result<(), PrefetchError> {
        re_tracing::profile_function!();

```

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
```diff
@@ -0,0 +1,347 @@
...
+                        if remaining_byte_budget == 0 {
+                            break; // We've already loaded too much.
+                        }
+
+                        skip_chunks.extend(
+                            store
+                                .find_root_rrd_manifests(&chunk_id)
+                                .into_iter()
+                                .map(|(id, _)| id),
+                        );
```
emilk: So this will ignore any root chunk that is at least _partially_ loaded? Is that correct? Looks like a comment would help here

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
```diff
@@ -0,0 +1,347 @@
...
+                .map(|(_, chunk_id)| *chunk_id)
+        };
+
+        let chunk_ids_in_priority_order = itertools::chain!(
+            used,
+            missing,
+            self.static_chunk_ids.iter().copied(),
+            std::iter::once_with(chunks_ids_after_time_cursor).flatten(),
+            std::iter::once_with(chunks_ids_before_time_cursor).flatten(),
+        );
```
emilk, line 329: Maybe this code block could be extracted to its own function to split out concerns (and make the current function less complex).

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
```diff
@@ -0,0 +1,347 @@
...
+
+        let entity_paths = manifest.col_chunk_entity_path_raw()?;
+
+        self.in_limit_chunks.clear();
+        let mut skip_chunks = HashSet::default();
```
emilk: This is a pretty long and complex function, with 6 parameters and 21 (!) additional `let` bindings at this point; far more than fits [in most peoples head](https://en.wikipedia.org/wiki/The_Magical_Number_Seven,_Plus_or_Minus_Two).

Is there anything we can do to reduce the complexity?

Some ideas:

* the actual prioritization ordering can be broken out (see earlier comment)
* the request batching can be broken out. If this function focuses on producing a stream or vec of `ChunkId`s, then some other function can be responsible for breaking that up into batches, starting the requests, and storing the promises

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
```diff
@@ -0,0 +1,347 @@
...
+
+    /// Batch together requests until we reach this size.
+    pub max_uncompressed_bytes_per_batch: u64,
+
+    /// Total budget for all loaded chunks.
+    pub total_uncompressed_byte_budget: u64,
+
+    /// Maximum number of bytes in transit at once.
+    pub max_uncompressed_bytes_in_transit: u64,
+}
```
emilk: Consider the reading order of your code! This would have been useful to read _before_ all the logic that actually used it

crates/store/re_entity_db/src/chunk_promise.rs
```diff
@@ -92,6 +93,22 @@ impl ChunkPromises {
         self.batches.iter().map(|b| b.size_bytes_uncompressed).sum()
     }
 
+    pub fn bandwidth(&self) -> Option<f64> {
```
emilk, line 97:
```suggestion
    /// Average of bytes/second over recent history
    pub fn bandwidth(&self) -> Option<f64> {
```

crates/store/re_entity_db/src/chunk_promise.rs
```diff
@@ -92,6 +93,22 @@ impl ChunkPromises {
...
+        self.download_size_history.bandwidth().map(|b| b.0)
+    }
+
+    pub fn bandwidth_data_age(&self, time: f64) -> f64 {
```
emilk: docstring

IsseW:
@rerun-bot reality-sync

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
IsseW: I don't think it's the most correct way, but I think it's more correct than always downloading a partially downloaded chunk. Since when we split a chunk it's because it's a large chunk. If one of those split chunks were then gc'd that most likely means we're at a pretty low memory budget. Or that was a **very** big chunk. So not really a situation where we want to re-download it.

Will add a comment!

IsseW:
> vs main:

Screenshot 2026-01-27 at 11 48 32

This is what my timing on main for prefetch chunks looks like on biggy after letting it fetch some chunks, so I don't think it's that much slower.

Also made some changes to make it faster, so should be ~the same now. But yes still way to slow for what we want.

crates/store/re_chunk_store/src/gc.rs
IsseW: Yeah, out of date doc string as I only have physical chunks in that now. Updated.

crates/store/re_chunk_store/src/lineage.rs
IsseW: Removed the term leaf and use `descendents` instead since that's what other lineage places use.

crates/viewer/re_selection_panel/tests/snapshots/selection_panel_recording.png
emilk, line 1: we should probably REDACT the Size here at some point

crates/store/re_entity_db/src/chunk_promise.rs
```diff
@@ -92,6 +93,27 @@ impl ChunkPromises {
         self.batches.iter().map(|b| b.size_bytes_uncompressed).sum()
     }
 
+    /// Average of bytes/second over recent history.
+    pub fn bandwidth(&self) -> Option<f64> {
+        self.download_size_history.bandwidth().map(|b| b.0)
+    }
+
+    /// Returns how fresh the bandwidth data is, as a normalized value from 0.0 to 1.0.
```
emilk, line 101:
```suggestion
    /// Returns how fresh the [`Self::bandwidth`] data is, as a normalized value from 0.0 to 1.0.
```

crates/store/re_entity_db/src/chunk_promise.rs
```diff
@@ -92,6 +93,27 @@ impl ChunkPromises {
...
+    ///
+    /// - `0.0` means the most recent download just completed.
+    /// - `1.0` means no downloads have completed within `Self.download_size_history.max_age()`.
```
emilk: So the most fresh data has a _low_ freshness value? That seems backwards. This function should probably be called `bandwidth_stalneness` instead, or invert its range

crates/viewer/re_data_ui/src/component_path.rs
```diff
@@ -21,15 +21,28 @@ impl DataUi for ComponentPath {
...
+                .unloaded_temporal_entries_for(
+                    &re_log_types::Timeline::new(
+                        query.timeline(),
+                        db.timeline_type(&query.timeline()),
+                    ),
+                    entity_path,
+                    Some(*component),
+                )
+                .iter()
+                .any(|chunk| chunk.time_range.contains(query.at()))
```
emilk: a helper function makes the code easier to read

```suggestion
            if !db.rrd_manifest_index().has_fully_loaded(query) {
```

crates/store/re_chunk_store/src/lineage.rs
```diff
@@ -376,35 +376,67 @@ impl ChunkStore {
...
+                    Either::Left(
+                        self.leaky_compactions
+                            .get(chunk_id)
+                            .filter(is_physical)
+                            .copied()
+                            .into_iter(),
+                    )
+                })
+        }
+        .into_iter()
```
emilk: Maybe @teh-cmc can take a quick peek at this

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
IsseW: Agreed, split it out a bit. Now `ChunkRequestBatcher` is responsible for batching requests. And `chunks_in_priority` that creates the iterator for chunks in priority order.

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
```diff
@@ -0,0 +1,474 @@
...
+        })
+    }
+
+    /// Returns (`uncompressed_size`, `byte_size`)
+    fn chunk_sizes(&self, row_idx: usize) -> (u64, u64) {
+        (
+            self.chunk_byte_size_uncompressed_raw[row_idx],
+            self.chunk_byte_size_raw[row_idx],
+        )
+    }
```
emilk: This feels like an unnecessary layer. The user can just reach in  and do `batcher.chunk_byte_size_uncompressed_raw[row_idx]` and get more readable code

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
```diff
@@ -0,0 +1,474 @@
...
+        let rb = take_record_batch(
+            &self.manifest.data,
+            &Int32Array::from(std::mem::take(&mut self.indices)),
+        )?;
+        self.chunk_promises.add(ChunkPromiseBatch {
+            promise: parking_lot::Mutex::new(Some((self.load_chunks)(rb))),
+            size_bytes_uncompressed: self.uncompressed_bytes_in_batch,
+            size_bytes: self.bytes_in_batch,
+        });
+        self.uncompressed_bytes_in_batch = 0;
```
emilk, line 111: we should also reset `self.bytes_in_batch` too, right?

crates/store/re_entity_db/src/rrd_manifest_index/chunk_prioritizer.rs
IsseW: We already have this on the `RrdManifestIndex` function that calls this function. Do you mean to move that span here, or have both?

crates/viewer/re_selection_panel/tests/snapshots/selection_panel_recording.png
IsseW: Might also be good to be warned if the size accidentally grows a lot though?

crates/store/re_chunk_store/src/lineage.rs
```diff
@@ -376,35 +376,67 @@ impl ChunkStore {
...
+        chunk_id: &'a ChunkId,
+    ) -> impl Iterator<Item = ChunkId> {
+        let is_physical = |c: &&ChunkId| self.chunks_per_chunk_id.contains_key(c);
+
+        if is_physical(&chunk_id) {
+            #[expect(
+                clippy::iter_on_single_items,
+                reason = "We need this to be an option below"
+            )]
+            Either::Left(Some(*chunk_id).into_iter())
```
teh-cmc: A comment as to why this is enough wouldn't hurt here I guess:
```suggestion
            // A physical chunk cannot have descendents. If it did, it would have been offloaded already.
            Either::Left(Some(*chunk_id).into_iter())
```

IsseW:
@rerun-bot reality-sync

IsseW:
@rerun-bot reality-sync

IsseW:
@rerun-bot reality-sync

crates/viewer/re_selection_panel/tests/snapshots/selection_panel_recording.png
emilk: True, but I don't think a screenshot test is the best way to enforce that

emilk: I'll redact the size in https://github.com/rerun-io/reality/pull/295

emilk:
@rerun-bot reality-sync

Issue #12530, opened 2026-01-20
Pinhole with named transform leads to a no transform path error on TransformAxes3D


danielduberg:
**Describe the bug**

When creating a `Pinhole` with a named transform, the implicitly created `TransformAxes3D` would generate a transform error.

**To Reproduce**
1. Run the code below.
2. Inspect entity `/camera` in the viewer.

```python
import rerun as rr
import numpy as np

rr.init("tf", spawn=True)

img_height, img_width = 12, 16
focal_length = 10

rr.send_columns(
    'camera',
    indexes=[rr.TimeColumn('time', sequence=[0])],
    columns=rr.Pinhole.columns(
        image_from_camera=[[[focal_length, 0, img_width / 2],
                           [0, focal_length, img_height / 2],
                           [0, 0, 1]]],
        resolution=[[img_width, img_height]],
        camera_xyz=[rr.ViewCoordinates.LEFT_HAND_Z_UP],
        child_frame=['world/base_link/camera/optical'],
        parent_frame=['world/base_link/camera'],
    ),
)

checkerboard = np.zeros((img_height, img_width, 1), dtype=np.uint8)
checkerboard[(np.arange(img_height)[:, None] +
              np.arange(img_width)) % 2 == 0] = 255
rr.set_time("time", sequence=0)
rr.log("camera/image", rr.Image(checkerboard),
   rr.CoordinateFrame('world/base_link/camera/optical'))

rr.log('/', rr.CoordinateFrame('world/base_link/camera'), static=True)
```

**Expected behavior**

There should be no error, as it should know from what has been speicified for the `Pinhole` what named transform to use for the `TransformAxes3D`.

**Screenshots**

The screenshot below shows the error.

<img width="1812" height="1058" alt="Image" src="https://github.com/user-attachments/assets/67a03f68-563b-488c-b059-17e5294e7a6a" />

**Backtrace**

Not a crash, but this is the error shown in the viewer:

```
No transform path from "tf#/camera" to the view's origin frame ("world/base_link/camera").
```

**Desktop:**
 - OS: macOS Tahoe 26.2 

**Rerun version**

```
rerun-cli 0.29.0-alpha.1+dev (e3b9abf)
aarch64-apple-darwin
rerun-cli features: base lance map_view nasm native_viewer oss_server perf_telemetry re_server release_full release_no_web_viewer web_viewer
rustc 1.90.0 (1159e78c4 2025-09-14), LLVM 20.1.8
[...]

PR #12558, opened 2026-01-22
Fix `TransformAxes` not picking up transform frame from `Pinhole` with named transforms

Wumpf:
### Related

* Fixes https://github.com/rerun-io/rerun/issues/12530

### What

.. and then complaining that we don't have a transform to show


grtlr (approved):

Wumpf:
@rerun-bot reality-sync

PR #12559, opened 2026-01-22
Fix case of unstable camera image plane distance

Wumpf:
While testing a very degenerated scene in https://github.com/rerun-io/rerun/pull/12558 I had some trouble with jumpy bounding boxes causing flickering in image plane lengths. This fixes it. Also, works around a debug assertion I occasionally ran into


grtlr:
@rerun-bot reality-sync

PR #12564, opened 2026-01-23
Use optical flow tracking in detect and track objects example

lucasmerlin:
This should hopefully fix our nightly example builds by removing the opencv-contrib dependency.

To me it even looks way smoother now:

https://github.com/user-attachments/assets/7e5a2ac6-d74e-409a-9946-76e26637dab9


lucasmerlin:
@rerun-bot reality-sync

PR #12565, opened 2026-01-23
Construct socket addr correctly

IsseW:
### Related

* Closes RR-3445

### What

Fixes the web viewer server not parsing the socket address correctly.

### DIFF
--- a/crates/viewer/re_web_viewer_server/src/lib.rs
+++ b/crates/viewer/re_web_viewer_server/src/lib.rs
@@ -106,7 +106,7 @@ impl WebViewerServer {
     /// # Ok(()) }
     /// ```
     pub fn new(bind_ip: &str, port: WebViewerServerPort) -> Result<Self, WebViewerServerError> {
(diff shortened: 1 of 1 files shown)


crates/viewer/re_web_viewer_server/src/lib.rs
```diff
@@ -106,7 +106,7 @@ impl WebViewerServer {
     /// # Ok(()) }
     /// ```
     pub fn new(bind_ip: &str, port: WebViewerServerPort) -> Result<Self, WebViewerServerError> {
-        let bind_addr: std::net::SocketAddr = format!("{bind_ip}:{port}").parse()?;
```
IsseW, line 109: In the default case this would become `:::{port}`, which is invalid. The correct syntax for ipv6 would be `[::]:{port}`. But we just need to parse the bind_ip here and can more easily construct it that way.

IsseW:
@rerun-bot reality-sync

PR #12567, opened 2026-01-23
Simplify queries with store/default/overrides to never ignore default values

Wumpf:
### Related

* Part of RR-3400

### What

After looking it over this distinction seems to be no longer necessary: for visualizability we need store components present anyways! This paves the way for explicitly selecting between default/override/store

It's possible to construct cornercases where this does make a difference, but it's pretty uncommon: you'd need to set a default for the exact (required) component first, then log some data, visualize it and scroll back in time before you've logged the value. Then the default is now active whereas before it wasn't. From that point of view this is a bugfix since we already advertised the defaults in the ui!
But practically I don't think this has any noticeable effect (famous last words ;))

### DIFF
--- a/crates/viewer/re_view_spatial/src/visualizers/assets3d.rs
+++ b/crates/viewer/re_view_spatial/src/visualizers/assets3d.rs
@@ -138,7 +138,8 @@ impl VisualizerSystem for Asset3DVisualizer {
             |ctx, spatial_ctx, results| {
                 use re_view::RangeResultsExt as _;
 
-                let all_blob_chunks = results.get_required_chunk(Asset3D::descriptor_blob().component);
+                let all_blob_chunks =
+                    results.get_required_chunk(Asset3D::descriptor_blob().component);
                 if all_blob_chunks.is_empty() {
                     return Ok(());
                 }
--- a/crates/viewer/re_view_spatial/src/visualizers/video/video_frame_reference.rs
+++ b/crates/viewer/re_view_spatial/src/visualizers/video/video_frame_reference.rs
@@ -72,8 +72,8 @@ impl VisualizerSystem for VideoFrameReferenceVisualizer {
                 let timeline = ctx.query.timeline();
                 let entity_path = ctx.target_entity_path;
 
-                let all_video_timestamp_chunks =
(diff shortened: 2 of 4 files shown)


Wumpf:
[7aa2478](https://github.com/rerun-io/rerun/pull/12567/commits/7aa247807db2a050d192265c702f6355360bcc30) is actually a little bit crazy: we always dropped the row ids on non-required chunks for latest at queries. Which turns out is almost never important because since it’s latest-at we only have a single row anyways. Buuuuut we use that rowid sometimes for cacheing, in particular images. Which is why a test failed (good test <3)

Edit: Not entirely correctly unfortunately. We did do this deliberately to make our range zips work out with both latest-at and range queries. I think they can still work if we just always discard the rows, but I have to adjust use of rowids for caching then... on it...

crates/viewer/re_view/src/results_ext.rs
```diff
@@ -287,23 +287,10 @@ impl RangeResultsExt for HybridLatestAtResults<'_> {
...
             let results_chunks = self.results.get_chunks(component);
-            let chunks = results_chunks
-                .chunks
-                .iter()
-                // NOTE: Since this is a latest-at query that is being coerced into a range query, we
-                // need to make sure that every secondary column has an index smaller then the primary column
-                // (we use `(TimeInt::STATIC, RowId::ZERO)`), otherwise range zipping would yield unexpected
-                // results.
-                .map(|chunk| chunk.clone().into_static().zeroed())
-                .collect_vec();
```
grtlr: Nice catch, that would have been so easy to miss!

Wumpf: my fix is also wrong T_T

Wumpf:
@rerun-bot reality-sync

crates/viewer/re_view/src/results_ext.rs
Wumpf: long story short: this was intentional. As the old comment says we _are_ treating all these outputs as-if they were range queries. range queries however have quite different semantics!

Wumpf:
we walked the latest changes on a call briefly. I think that counts as review, also now it's less invasive

PR #12568, opened 2026-01-23
Fix toggling visibility of plots via legend

grtlr:
### Related

* Closes RR-3447

### What

The `visible_series` was written at the wrong location (the base blueprint path). This PR fixes this by looking for the particular visualizer of interest. 

Alternative option would be piping the entire `VisualizerInstruction` through.


grtlr:
@rerun-bot reality-sync

crates/viewer/re_view_time_series/src/lib.rs
```diff
@@ -77,6 +80,9 @@ pub enum PlotSeriesKind {
 pub struct PlotSeries {
     pub instance_path: InstancePath,
 
+    /// Id of the visualizer instruction that is responsible for this series.
+    pub visualizer_instruction_id: VisualizerInstructionId,
+
     /// Id used for this series in the egui plot view.
     pub id: egui::Id,
```
Wumpf: nit: can kick this out now in favor of the visualizer instruction id. I think.

grtlr: @copilot address this feedback

PR #12574, opened 2026-01-23
Track orphaned_recordings separately

jleibs:
test


rerun_py/src/python_bridge.rs
```diff
@@ -52,6 +52,15 @@ fn all_recordings() -> parking_lot::MutexGuard<'static, Vec<RecordingStream>> {
     ALL_RECORDINGS.get_or_init(Default::default).lock()
 }
 
+// We separately track orphaned recordings. These have been disconnected and are flushing but
```
Wumpf, line 55: the comment on `all_recordings` is still explaining why *all* recordings have to be kept around. Should acknowledge that it's not all-all

PR #12575, opened 2026-01-23
Data Platform consistency

ntjohnson1:
### Related
RR-3450

### What
Adds a lint so we use Data Platform is most situations https://rerunio.slack.com/archives/C04N1AZ5D9S/p1769184232433139?thread_ts=1769169623.661129&cid=C04N1AZ5D9S

* [x] figure out why the snapshots are unhappy

### DIFF
--- a/tests/rust/re_integration_test/tests/snapshots/dataset_ui_empty_form.png
+++ b/tests/rust/re_integration_test/tests/snapshots/dataset_ui_empty_form.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:6a4fc9b2215a1fca8a05240fefb3e88a093967362c4c75f83997c2580710e3ab
-size 66490
+oid sha256:cd36aaa2f87ad831b53eb740053bd2e36da47509f7240e2bb41c19005ead4d5e
+size 66283


crates/store/re_redap_tests/src/tests/common.rs
```diff
@@ -251,7 +251,7 @@ impl LayerType {
                 segment_id,
                 // TODO(ab): avoid this annoying conversion (this requires a change to
                 // `create_recording_with_properties` which needs to be propagated to
-                // `dataplatform`.
+                // `Data Platform`.
```
abey79: I dont think this change is correct here. We have instances such as these to specifically refer to the "commercial version of the data platform" (as opposes to "anything that implements redap"). I'm unsure what to suggest concretely, since we haven't yet fully sorted out our naming of things. Maybe not lint when back-tick quoted? Or maybe just reword that comment.

ntjohnson1: Hmm I need to fix some snapshots so I am happy to add an exception.

However, now that `dataplatform` repo no longer exists does it matter if we say data platform vs Data Platform? On some level commercial version of the data platform is referring to the Data Platform. I don't feel super strongly on it so open to whatever you think is reasonable.

abey79: Looking again at this comment, I dont understand what it tries to say. I'd cut it short after "avoid this annoying conversion" :)

ntjohnson1:
@rerun-bot reality-sync

ntjohnson1:
@rerun-bot reality-sync

PR #12576, opened 2026-01-24
Fix CI: opt out segment_properties snippet from Rust/C++

jleibs:
It appears we have a PR testing gap as this was green in https://github.com/rerun-io/reality/pull/200

I created an issue to investigate that: RR-3466

But adding this to fix main.


jleibs:
@rerun-bot reality-sync

PR #12577, opened 2026-01-24
Always set the reality sync status to pending

jleibs:
This should help improve the way the pending check that blocks merging renders


jleibs:
Maybe this was a bad idea -- I see now it leaves the overall PR status blocked at yellow rather than green and just un-mergable

Wumpf:
seems stale

PR #12578, opened 2026-01-24
Fix broken snippets and snapshots

jleibs:
### Related

Not clear why this is broken on `main` but isn't breaking in CI.
- https://github.com/rerun-io/reality/pull/204 passed
- https://github.com/rerun-io/reality/commit/841374381e7c631ecdd524881c99cfd34ee7c432 failed

### DIFF
--- a/crates/store/re_server/src/rerun_cloud.rs
+++ b/crates/store/re_server/src/rerun_cloud.rs
@@ -1122,9 +1122,8 @@ impl RerunCloudService for RerunCloudHandler {
         } = request.into_inner().try_into()?;
 
         if scan_parameters.is_some() {
-            re_log::info_once!(
-                "query_dataset: scan_parameters are not yet implemented and will be ignored"
-            );
+            // Logged at a low debug-level, because of https://github.com/rerun-io/rerun/pull/12578
+            re_log::debug_once!("   scan_parameters are not yet implemented and will be ignored");
         }
 
         let entity_paths: IntSet<EntityPath> = entity_paths.into_iter().collect();


jleibs:
@rerun-bot full-check

jleibs:
@emilk in particular would be great if you could take a look since I don't know if you intended for that popup to show up in the diff or why this doesn't get caught in PRs but does get caught on main.

@Antoine I also wanted to point out that the snippet you added won't be tested because our snippets depend on the thing that forces rrds to go to a different location. Not sure if we want to add coverage somewhere else.

jleibs:
Oh, extra fun... so when I run locally I get the same behavior we were seeing in `main`... but my fix of course breaks on the PR   :sob: 

Something racy going on probably?.

emilk:
So this is the problem:

* The tests starts up a server in a background task
* It open a Viewer, an does a query
* The server logs an INFO message ("query_dataset: scan_parameters are not yet implemented and will be ignored")
* The viewer task shows that INFO log message as a toast… if it arrives at the right/wrong time

(we have a system that maps all log messages into notifications)

Some potential solutions:
* A) Demote log message from INFO to DEBUG
* B) Disable log->notification system for this particular test (…but that may leave other test vulnerable)
* C) Disable log->notification system for `re_server`
* D) let the test run until the first notification arrives
* E) Even more custom crazyness

I'll go with the simplest solution for now (A)

emilk:
@rerun-bot full-check

ntjohnson1:
@rerun-bot reality-sync

PR #12579, opened 2026-01-24
docs(sdk): add thread safety warning for log() with mutexes

ecto:
## Summary

Add documentation warning users about potential deadlocks when calling `RecordingStream::log()` while holding a `std::sync::Mutex`.

## Background

We encountered a deadlock in production on Linux aarch64 when:
1. A 100 Hz control loop held a `std::sync::Mutex`
2. Called `RecordingStream::log()` with LiDAR point cloud data
3. The rerun SDK's internal rayon thread pool caused a deadlock

The deadlock occurs because rayon's work-stealing behavior can cause implicit lock recursion when a mutex is held while calling into rayon-using code. This is a known rayon limitation documented in [rayon-rs/rayon#592](https://github.com/rayon-rs/rayon/issues/592).

## Debugging Evidence

GDB backtrace showed:
- Main thread blocked on `std::sys::sync::mutex::futex::Mutex::lock_contended`
- Rayon worker threads (from rerun) in sleep state
- Classic symptoms of the rayon + mutex deadlock pattern

## Reproduction

We created a [minimal reproduction attempt](https://gist.github.com/ecto/2749b99a6f899cc9f7caf3d36d7e8f0c) modeling the pattern. Note: the deadlock is timing/platform dependent and we could not reproduce on macOS x86_64 — it only manifested on Linux aarch64 under load.

## Changes

Added a "Thread Safety" section to `RecordingStream::log()` documentation with:
- Clear warning about the mutex + rayon interaction
- Code example showing the problematic pattern
- Code example showing the safe alternative (extract data before logging)
- Link to the relevant rayon issue

## Testing

Documentation change only - verified the example code is syntactically correct.


ecto:
Nope, I don't use rayon anywhere :) pls check the repro above

Wumpf:
I suspect the problem isn't rayon (which I'd be surprised if we use it on the path of `log`) but rather tokio - your repro uses a standard mutex and tokio has afaik the exact same issue, see https://turso.tech/blog/how-to-deadlock-tokio-application-in-rust-with-just-a-single-mutex
Did you perchance halt the application during a deadlock and check the threads that are stuck? They should all be named, so it should be easy to confirm whether it is tokio or rayon.
Outcome is of course still the same: standard mutex in the right constellation may causes deadlocks 😩 

EDIT: sorry I overlooked that you wrote that you already checked with gdb and it looked like it's the rayon theads blocking. Hum, very strange all that, I share Emil's confusion.

emilk:
@rerun-bot reality-sync

emilk:
No matter what is the cause, I think it is a good warning to have and heed. Thanks for making this PR!

PR #12582, opened 2026-01-25
Migrate serde-yaml to serde_yaml_bw

bourumir-wyngs:
This change replaces the unmaintained [`serde_yaml`](https://crates.io/crates/serde_yaml) crate with the maintained [`serde_yaml_bw`](https://crates.io/crates/serde_yaml_bw).

`rerun` uses YAML only in a limited capacity. Depending on the fast-moving `serde-saphyr` backend, which introduces frequent changes and additional features, does not appear necessary. `serde_yaml_bw` is more stable and better aligned with the project’s needs, while also offering improved robustness when handling malformed YAML.

The public APIs of the two crates are largely compatible, so the impact of this change should be minimal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -409,6 +409,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3d62b7694a562cdf5a74227903507c56ab2cc8bdd1f781ed5cb4cf9c9f810bfc"
 
 [[package]]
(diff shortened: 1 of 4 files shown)


bourumir-wyngs:
I could also implement this without the YAML library. Are regular expressions OK, or would you prefer a plain Rust implementation?

PR #12587, opened 2026-01-26
Fix false positive warnings about scoped IDs in DAE importer

MichaelGrupp:


MichaelGrupp:
@rerun-bot reality-sync

PR #12588, opened 2026-01-26
Warn when logging transform at `/` without set `parent_frame`

Wumpf:
Previously, we'd trigger a debug assertion and ignore the transform silently. Now we still ignore the transform, but log a warning. Also, this adds an automated test to check for this and removed the old release check list item that stumbled into that (other aspects of the release checklist manual test were already covered fairly well)


Wumpf:
@rerun-bot reality-sync

PR #12590, opened 2026-01-26
Visualizer queries (`HybridResults`) now apply `ComponentSourceKind` if present

Wumpf:
### Related

* Fixes RR-3400

### What

This allows to explicitly select default/fallback/store/override (tbd if we merge default/fallback) and have the effects be visible on all visualizers.

Note that this general ability is not limited to plots and not gated behind a feature flag. What is though is the (unfinished) UI as well as flexible remapping in the time series plot. Also, we don't expose writing these mappings in Rust/Python blueprint api.


Wumpf:
@rerun-bot reality-sync

grtlr:
Approved in reality.

Wumpf:
Will merge together with https://github.com/rerun-io/rerun/pull/12590 in one go

PR #12592, opened 2026-01-26
Remove distinction of `ComponentSourceKind::Fallback` & `ComponentSourceKind::Default`

Wumpf:
### Related

* Fixes RR-3399
* Direct follow-up to https://github.com/rerun-io/rerun/pull/12590

### What

I did the minimal thing I think it’s actually decent: removed `ComponentSourceKind::Fallback` and documented that _all_ the other variants today will pick fallback if there’s no fallback.

Failure to resolve `ComponentSourceKind::SourceComponent` should almost certainly be an error, but in the interest of small steps I haven't looked further into this and described status quo.

⚠️ as previously noted this removes the ability to enforce the fallback if there is a view default. Had to adjust my nascent mapping test for that a bit.


Wumpf:
@rerun-bot reality-sync

grtlr:
Looks like the sync bot botched some stuff here.

grtlr (approved):

PR #12595, opened 2026-01-27
Fix bug when closing modal

asinghka:
When calling `ModalHandler::close`, the `should_close` field is set to true and is never reset, causing the modal to remain in a permanently “closing” state. This PR ensures the modal does not immediately close due to stale state carried over from a previous interaction.


emilk:
@rerun-bot reality-sync

MichaelGrupp:
Thanks for the fix!

PR #12601, opened 2026-01-30
docs: update operating mode docs

### LINKED ISSUE #12580: docs: unclear operating mode & sink behaviour
Regarding the [operating mode docs](https://rerun.io/docs/reference/sdk/operating-modes) [here](https://github.com/rerun-io/rerun/blob/main/docs/content/reference/sdk/operating-modes.md). I think it'd help to add a note mentioning how to use multiple sinks at the same time, as the docs don't seem to mention it.

> All four of them are optional: when none of these modes are active, the client will simply buffer the logged data in memory, waiting for one of these modes to be enabled so that it can flush it.

The current behaviour seems to be that `connect_grpc()` and `save()` override each other and effectively destroy the other sink when called consecutively. There is no mention of [`set_sinks()`](https://ref.rerun.io/docs/python/stable/common/initialization_functions/#rerun.set_sinks), which seems to be the intended way to do this. Maybe we can add:

> Note: these will destroy any other active sinks; `set_sinks()` is the only way to use multiple sinks concurrently.

edit: I can make a PR for this if it gets approved

rrad5409:
- add note on using multiple sinks concurrently
- add docs for `set_sinks()`

### Related

Closes #12580

### DIFF
--- a/docs/content/reference/sdk/operating-modes.md
+++ b/docs/content/reference/sdk/operating-modes.md
@@ -93,7 +93,7 @@ Use [`rr.save`](https://ref.rerun.io/docs/python/stable/common/initialization_fu
 Use [`RecordingStream::save`](https://docs.rs/rerun/latest/rerun/struct.RecordingStream.html#method.save).
 
 
-### `stdout`
+### Standard Input/Output (`stdout`)
 
 Streams all logging data to standard output, which can then be loaded by the Rerun Viewer by streaming it from standard input.
 


docs/content/reference/sdk/operating-modes.md
```diff
@@ -90,7 +93,7 @@ Use [`rr.save`](https://ref.rerun.io/docs/python/stable/common/initialization_fu
 Use [`RecordingStream::save`](https://docs.rs/rerun/latest/rerun/struct.RecordingStream.html#method.save).
 
 
-### Standard Input/Output
+### `stdout`
```
grtlr:
```suggestion
### Standard Input/Output (`stdout`)
```

grtlr (approved):
Very nice, thank you!

grtlr:
@rerun-bot reality-sync

PR #12611, opened 2026-02-02
rerun-export: Fix url in example README

oxkitsune:
### What

The README for the `rerun-export` example used a relative url to the `VideoStream` archetype page. This isn't supported by our docs build for examples and therefore they fail.


oxkitsune:
@rerun-bot reality-sync

Issue #12642, opened 2026-02-17
Viewer fails to open LeRobot v3 datasets: "Sender blocked waiting for space in channel" (Backpressure/Timeout)


KnoblochMathis:
**Describe the bug**
rerun viewer does not open LeRobot v3 datasets above a certain size/episode count; 
it works with smaller datasets

**To Reproduce**
1. Open Terminal type: `rerun /path/`
2. Execute
3. After 5 seconds warning appears
4. Nothing happens

**Expected behavior**
Rerun Viewer should open and show all episodes with videos and joint data.

**Backtrace**
```powershell
lab@lab091145:~$ rerun ~/.cache/huggingface/lerobot/AIR/redBoxDataset
[2026-02-16T15:45:06Z INFO  winit::platform_impl::linux::x11::window] Guessed window scale factor: 1
[2026-02-16T15:45:06Z WARN  wgpu_hal::vulkan::instance] Unable to find extension: VK_EXT_physical_device_drm
libEGL warning: DRI2: failed to authenticate
[2026-02-16T15:45:07Z INFO  re_grpc_server] Listening for gRPC connections on 0.0.0.0:9876. Connect by running `rerun --connect rerun+http://127.0.0.1:9876/proxy`
[2026-02-16T15:45:07Z INFO  re_data_loader::load_file] Loading "/home/lab/.cache/huggingface/lerobot/AIR/redBoxDataset"…
[2026-02-16T15:45:07Z INFO  re_data_loader::loader_lerobot] Loading LeRobot v3 dataset from "/home/lab/.cache/huggingface/lerobot/AIR/redBoxDataset", with 399 episode(s)
[2026-02-16T15:45:16Z WARN  re_quota_channel::sync] log_channel(file:///home/lab/.cache/huggingface/lerobot/AIR/redBoxDataset): Sender has been blocked for over 5 seconds waiting for space in channel
``` 

**Desktop:**
 - OS: Ubuntu 22.04.5 LTS

**Rerun version**
rerun-sdk Version: 0.29.0

**Additional context**
This is not due to RAM limitations; `--memory-limit`has been raised; there is sufficient memory available.

danielholl14:
Hi,
I have the same problem with large lerobot datasets.

PR #12652, opened 2026-02-20
Fix deadlock when loading LeRobot datasets

emilk:
### Related
* Closes https://github.com/rerun-io/rerun/issues/12642

### What
The LeRobot dataset loader would sometimes create 1000+ recordings, which would create 1000+ SystemCommands. In my eagerness to enforce backpressure in our codebase, I had made the command senders bounded. BUT: most often we send and receive commands on the _same thread_ (the main GUI thread). This means the sender will never be unblocked!

So the fix was simple: make the command channels unbounded.


emilk:
@rerun-bot reality-sync

PR #12653, opened 2026-02-20
Drop cached videos once all referencing episodes are loaded

oxkitsune:
### What

This makes the LeRobot dataset loader a bit smarter, by dropping the cached video blobs from memory as soon as all episodes referencing the blob have been loaded.

This makes the loader a bit better at handling larger datasets.


oxkitsune:
@rerun-bot reality-sync

emilk (approved):
Code LGTM! How much difference does it make in practice?

oxkitsune:
I ran a test and found that Rerun now uses a little over half the memory it previously required when loading a dataset.

(This is based on looking at the memory number at the top of the viewer!)

oxkitsune:
@rerun-bot reality-sync

Issue #12602, opened 2026-01-30
Improve Python logging performance of small things like `Transform3D`


emilk:
## Related
* https://github.com/rerun-io/rerun/issues/12485
* https://github.com/rerun-io/reality/pull/184

## What
When logging many small things, e.g. transforms, the overhead of each log call can be significant, and on the order of 40µs. That's 20x slower than Rust.

## Why
Generated with the instructions in `rerun/tests/python/log_benchmark/test_log_benchmark.py`
<img width="1200" height="1274" alt="Image" src="https://github.com/user-attachments/assets/6a76df01-23b5-4184-a391-b31145c1cd8b" />

### Culprit 1: arrow encoding
A lot of that overhead is from encoding the data into arrow, which is done in Python using py-arrow. We're essentially generating a bunch of single-row columns - something Arrow is _not_ optimized for.

There are a few plausible avenues to explore:
* Batch many `rr.log` call _in Python_ so we can encode them all at once.
* Use codegen to generate more efficient arrow encoding code than what pyarrow provides
* Switch from arrow to something else for the Python->Rust interface

### Culprit 2: `_log_components` overhead
I'm not sure what we can do here, except to batch `rr.log` calls in Python. The problem then becomes when and how to flush those calls.

## Workaround: manual batching
Batch many small `rr.log` call into one big one. Most of our archetypes support logging many at once (e.g. log many points at once instead of a single one).

nikolausWest:
I'll add another workaround in the case of transforms. In many cases with many fast transforms the actual source data on the user side will be vectors of join angles. They will then use their URDF (robot model) to turn joint angles into transforms. Since we don't support multi-entity log calls, each joint angle will need to turn into one transform per entity. If the viewer supporter joint_angles mapped to URDFs directly, the user could just send a static joints_angles -> URDF mapping once and then log joint angle vectors. Simpler code, less stored data, and higher performance all around

ntjohnson1:
[...]

PR #12656, opened 2026-02-22
Log transforms faster with patterns for general faster python logging

ntjohnson1:
### Related
https://github.com/rerun-io/rerun/issues/12602

### What
Got nerd sniped on the plane and fell down this rabbit hole hard. Ended up with a pile of spaghetti so tried to get claude to reproduce my path a little cleaner.

If you look back at the commits I pruned a tests/hypotheses directory that shows the different things explored and their impact. Leaving as a draft since this still needs a little cleanup to decide if we want all the speedups or we want to remove some of the ones that are a bit messier.

Overall impact is ~4x performance improvement. From 35k to 135k samples per second on the benchmark Emil made.


Wumpf:
@ntjohnson1 any intention of making this still fly?

ntjohnson1:
No. We landed some different pieces but ultimately this is all a bit of a mess and we should refactor more in the future/probably needs better benchmarking to drive smart choices anyway

PR #12660, opened 2026-02-23
Fix mono8/mono16 image channel classification

oxkitsune:
### What

This correctly classifies `mono8` and `mono16` as single channel images, so our ROS2 message parser loads these images as depth image.

Additionally this adds the missing `CvF64C1` f64 single channel image from `sensor_msgs/image_encodings.h`


oxkitsune:
@rerun-bot reality-sync

PR #12663, opened 2026-02-23
Use entity path derived frame id for empty frame ids

oxkitsune:
### What

This makes parsing some MCAP files that contain messages with an empty frame id a bit less painful.

### DIFF
--- a/crates/store/re_tf/src/transform_resolution_cache/tests.rs
+++ b/crates/store/re_tf/src/transform_resolution_cache/tests.rs
@@ -1685,6 +1685,76 @@ fn test_error_on_changing_associated_path(time: TimeInt) -> Result<(), Box<dyn s
     Ok(())
 }
 
+/// Empty frame id strings (as commonly found in technically wrong, but painfully common ROS data)
+/// should be treated as entity-derived frames, not as a shared "" frame.
+///
+/// Multiple entities with empty frame ids must each get their own independent transform.
+#[test]
+fn test_empty_frame_id_falls_back_to_entity_path_derived() -> Result<(), Box<dyn std::error::Error>>
+{
+    let mut entity_db = new_entity_db_with_subscriber_registered();
+    let mut cache = TransformResolutionCache::default();
+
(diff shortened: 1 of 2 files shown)


oxkitsune:
We should hold off on this until we can discuss it with Michael next week:

> I'm not sure if reinterpreting Empty frame ID as an implicit one is the right approach

Wumpf:
thread https://rerunio.slack.com/archives/C09J226VB46/p1771858272990179

crates/store/re_tf/src/transform_resolution_cache/tests.rs
```diff
@@ -1685,6 +1685,76 @@ fn test_error_on_changing_associated_path(time: TimeInt) -> Result<(), Box<dyn s
     Ok(())
 }
 
+/// Empty frame id strings (as commonly found in technically wrong, but painfully common ROS data)
+/// should be treated as entity-derived frames, not as a shared "" frame.
```
MichaelGrupp, line 1689: Given this comment the main motivator here seems to be converted ROS data. As mentioned before [on Slack](https://rerunio.slack.com/archives/C09J226VB46/p1771920743390809?thread_ts=1771858272.990179&cid=C09J226VB46), **reinterpreting** an empty ROS frame_id (undefined) as a valid implicit Rerun-entity-path-hierarchy frame relationship is the wrong approach. It's silently changing the semantics of the input data (TF frames) to an alternative semantic that doesn't exist in ROS (entity path frames), which is most likely wrong and also can't be easily fixed by users.

I would rather treat it as what it is (undefined data) and have an explicit path:
* have a defined behavior for `CoordinateFrame("")` - e.g. a warning (see also #12718)
* potentially allow users to manually edit the frame ID in the UI

oxkitsune: Yea, this sounds good to me! Will close this!

oxkitsune:
Closing in favor of #12718

PR #12669, opened 2026-02-25
Upload extensionless HTML aliases to GCS for clean URLs

jleibs:
Resolves: RR-3687

## Summary
- Adds a CI step to both `reusable_upload_web.yml` and `reusable_publish_web.yml` that copies HTML files (e.g. `signed-in.html`) to extensionless objects (e.g. `signed-in`) before uploading to GCS
- This allows `app.rerun.io/signed-in` to work in addition to `app.rerun.io/signed-in.html`, since the GCS backend bucket + load balancer don't strip `.html` extensions automatically
- `index.html` and `index_bundled.html` are excluded since GCS handles directory index resolution for those

## Test plan
- [x] Manually copied `signed-in.html` → `signed-in` and `signed-out.html` → `signed-out` in `gs://rerun-web-viewer/version/latest/` and verified `app.rerun.io/signed-in` works

### DIFF
--- a/.github/workflows/reusable_publish_web.yml
+++ b/.github/workflows/reusable_publish_web.yml
@@ -104,6 +104,16 @@ jobs:
         run: |
           pixi run rerun-build-web-release
 
+      - name: "Create HTML aliases without extensions"
+        run: |
+          cd crates/viewer/re_web_viewer_server/web_viewer
+          for f in *.html; do
+            name="${f%.html}"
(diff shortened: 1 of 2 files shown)


jleibs:
> maybe there's a way to do this through GCS path rewrites or something

I looked into that and then created: https://linear.app/rerun/issue/RR-3871/apprerunio-should-be-deprecated-hosted-properly instead

jleibs:
@rerun-bot reality-sync

PR #12672, opened 2026-02-27
Fix incorrect byte unit conversion values in render_bench.py

jashshah999:
The `UNITS` dictionary in `render_bench.py` had incorrect values for MB, GB, TB, GiB, and TiB. They were all set to `1000` or `1024 * 1024` instead of the correct powers:

| Unit | Was | Should be |
|------|-----|-----------|
| mb | 1000 | 1000^2 |
| gb | 1000 | 1000^3 |
| tb | 1000 | 1000^4 |
| gib | 1024^2 | 1024^3 |
| tib | 1024^2 | 1024^4 |

This caused all benchmark size conversions above KB/KiB to produce incorrect values (e.g. bytes-to-MB would divide by 1000 instead of 1,000,000).


Wumpf:
@rerun-bot reality-sync

PR #12673, opened 2026-02-27
Reuse precomputed timelines dict in send_columns instead of re-iterating indexes

jashshah999:
`send_columns()` iterated over `indexes` to build a `timelines_args` dict with validation, but then passed a new dict comprehension to `send_arrow_chunk` that re-iterated `indexes` and called `as_arrow_array()` again.

Two problems:
1. If `indexes` is a generator or single-use iterable, the second iteration produces an empty dict, silently sending data without any timeline information.
2. Even with a list, it redundantly calls `as_arrow_array()` and `timeline_name()` on every index a second time.

Fix: use the already-built `timelines_args` dict.


ntjohnson1:
Thanks so much for the fix! This makes sense. I'm a little surprise that none of our ruff rules flag that we created timelines_args and never used it after. Will take a look into this to maybe catch if this happened elsewhere.

ntjohnson1:
@rerun-bot reality-sync

PR #12674, opened 2026-02-27
Fix wrong variable in error message for extra args in log()

jashshah999:
The error message for invalid items in the `extra` parameter of `log()` reported `type(entity)` (the already-validated first argument) instead of `type(ext)` (the actual offending extra argument). This made the error misleading when debugging.


Wumpf:
@rerun-bot reality-sync

PR #12676, opened 2026-02-27
[WIP] Custom shader support for Mesh3D

zalo:
https://github.com/user-attachments/assets/8face5d0-70b8-4ba0-be3b-8fcf036a80a0

Feel free to ignore or close this as a slop PR.  I am exploring adding an API which we frequently need, which is the ability to represent more complex volumetric data natively via raymarching.   Given how scary volume raymarching can be, I figured it made more sense to attempt to add WGSL shader injection for Mesh3D's, so the Python client can handle the custom trickiness.

This should be flexible enough to handle the terrain example and much more, but not quite enough for Gaussian Splatting or Marching Cubes.  Ideally, someday there would be a Slang integration for multipass compute pipelines (which _could_ handle GS and MC), but, for now, I figured this would enable >50% of custom visualizer needs.

While the code is all vibed, the ergonomics might be useful to frame future discussions on high-expressiveness/low-effort multi-language extensibility.

## Summary

This PR adds **custom WGSL fragment shaders on Mesh3D entities**, enabling effects like volumetric raymarching, stylized rendering, and data-driven visualization. The end-to-end pipeline is fully functional: custom shaders render correctly with dynamic uniform/texture bindings resolved from the entity store.

### What's included

**Type system & SDK:**
- New components (`unstable`): `ShaderSource` (inline WGSL code) and `ShaderParameters` (JSON metadata describing uniform/texture bindings and their source entity paths)
- Updated `Mesh3D` archetype with two new optional fields across Rust, Python, and C++
- Python SDK: `Mesh3D` accepts `shader_source`/`shader_parameters` kwargs; `slang_utils.py` helper for compiling Slang to WGSL

**Renderer infrastructure:**
- Inline WGSL shader modules with content-hash caching in the shader module pool
- 3D texture manager (`GpuTexture3D` / `TextureManager3D`) for volumetric data, following the existing `GpuTexture2D` pattern
- `CustomShaderMeshRenderer` — dynamic render pipeline that reuses the standard mesh vertex stage (`instanced_mesh.wgsl`) with user-provided fragment shaders
- Custom bind group builder for dynamic uniform buffers + 3D texture bindings

**Parameter binding:**
- JSON-driven reflection metadata maps shader uniforms to named entity paths (relative or absolute)
- Uniform buffer packing with std140 layout (float, vec2, vec3, vec4, mat4)
- `Tensor` → `GpuTexture3D` conversion for volumetric data
- Graceful handling of streaming data: rendering is skipped when required bindings aren't yet available, retrying each frame

**Vertex shader extensions:**
- Added `position_object` and `camera_pos_object` to `VertexOut` — enables custom fragment shaders to do object-space raymarching without exceeding the GPU inter-stage component limit (31 max)
- `camera_pos_object` is computed in the vertex shader via adjugate inverse of the 3x3 rotation portion of `world_from_mesh`, using only 3 inter-stage scalar components instead of the 12 that passing the full matrix would require

**DICOM volumetric example:**
- Updated `dicom_mri.py` with a raymarched volumetric rendering alongside the existing tensor view
- Logs a `[0,1]³` unit-cube mesh with a custom WGSL fragment shader that raymarches through a 3D volume texture
- Volume data logged as a downscaled 3D `Tensor` (≤128 per dimension) bound to the shader via entity path reference
- Shader parameters (`density_scale`, `value_range`) logged as `Scalars`/`Tensor` and resolved dynamically at each frame

### Architecture

```
SDK side (Python):                         Viewer side (Rust):
  slangpy compiles .slang → WGSL             Receives WGSL source + reflection JSON
  slangpy reflects params → JSON             Builds dynamic bind group from reflection
  rr.log("mesh", Mesh3D(                    Creates dynamic render pipeline
    ...,                                     Queries referenced entity data at current time
    shader_source=wgsl_str,                  Uploads uniforms/textures to GPU
    shader_parameters=json_str,              Renders with custom fragment shader
  ))
```

### Key files

| Area | Files |
|------|-------|
| Type definitions | `shader_source.fbs`, `shader_parameters.fbs`, `mesh3d.fbs` |
| Renderer infra | `custom_shader_mesh.rs`, `texture_manager_3d.rs`, `shader_module_pool.rs` |
| Parameter binding | `shader_params.rs`, `shader_param_resolver.rs`, `custom_shader_bind_group.rs` |
| Visualizer | `meshes.rs` (extended with shader component queries) |
| Vertex shader | `instanced_mesh.wgsl` (added `position_object`, `camera_pos_object`) |
| Python | `mesh3d_ext.py`, `slang_utils.py` |
| Example | `dicom_mri.py`, `volume_raymarch.wgsl`, `volume_raymarch.slang` |

## Test plan

- [x] `cargo clippy` passes on `re_renderer`, `re_view_spatial`, `re_sdk_types`
- [x] `cargo test -p re_sdk_types` — Mesh3D roundtrip test passes with new fields
- [x] `cargo test -p re_renderer` — all tests pass
- [x] `cargo test -p re_view_spatial` — all tests pass (including shader_params tests)
- [x] `cargo fmt` clean
- [x] `ruff` clean on Python files
- [x] End-to-end manual testing: DICOM volumetric raymarching renders correctly with dynamic parameter binding
- [x] Texture streaming race condition handled gracefully (skips rendering until data arrives)
- [ ] Snapshot regression tests for default Mesh3D rendering (unchanged behavior)
- [ ] CI validation (pending)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/python/dicom_mri/dicom_mri.py
+++ b/examples/python/dicom_mri/dicom_mri.py
@@ -139,10 +139,9 @@ def log_volumetric_rendering(voxels_volume: npt.NDArray[np.int16], shape: tuple[
     max_dim = 128
     from scipy.ndimage import zoom as ndimage_zoom
 
-    # The DICOM data axes are (right, back, up). Transpose to (back, up, right)
-    # so the 3D texture maps: X→right, Y→up, Z→back. This aligns the head's
-    # "up" direction with Rerun's Y-up world space (no Transform3D needed).
-    voxels_reoriented = np.ascontiguousarray(voxels_volume.transpose(1, 2, 0))
+    # Reverse the axis order so the head stands upright in Rerun's Y-up
+    # coordinate system. After transpose: World X = axis2, Y = axis1, Z = axis0.
+    voxels_reoriented = np.ascontiguousarray(voxels_volume.transpose(2, 1, 0))
     reoriented_shape = voxels_reoriented.shape
 
     scale_factors = tuple(max_dim / s if s > max_dim else 1.0 for s in reoriented_shape)


Wumpf:
Interesting experiment, thanks for sharing! I want to have a browse some point next week.

The bit off the bat I don't get is why this also tries to integrate Slang - are WGSL compute shader too limited for your purposes in some way? Also don't quite get how slang would solve any of the setup issues for marching cubes (which is typically ran on CPU only to generate a mesh, i.e. the opposite of ray marching) or gaussian splats. Generally, the problem with adjusting more than a (color-only) fragment shader is that this also required intricate knowledge of our current (and undocumented) pipeline for picking, outlines and many many more aspects of our admittedly very adhoc render pipeline.

zalo:
Thanks!

> The bit off the bat I don't get is why this also tries to integrate Slang - a re WGSL compute shader too limited for your purposes in some way? 

In the pure fragment shader context, it boils down to: I like C/HLSL/GLSL syntax for shaders and I don’t like rust-style syntax for shaders  🫠 The LLMs have a bit more experience with C-like syntax too, but that’s secondary.  I’m aiming to get the core of the PR to be WGSL native so I can just have my Python wrappers call out to SlangPy and not impose on the main pipeline, but I’d generally rather try to forget WGSL exists for myself 😅

> Also don't quite get how slang would solve any of the setup issues for marching cubes (which is typically ran on CPU only to generate a mesh, i.e. the opposite of ray marching) or gaussian splats. Generally, the problem with adjusting more than a (color-only) fragment shader is that this also required intricate knowledge of our current (and undocumented) pipeline for picking, outlines and many many more aspects of our admittedly very adhoc render pipeline.

Slang has relatively good support for defining full compute pipelines in a single file ala CUDA ( see the playground Gaussian splatter https://shader-slang.org/slang-playground/ and SlangPy Pathtracer examples); a proper deep slang integration could expand the mesh rendering process of the shader into emitting triangles for marching cubes, or emitting and sorting and rasterizing the splats for Gaussian splats. 🤔

I think WGSL also has reflection capabilities that can also sort of do this too?  I’ll admit, I haven’t kept up with it.  I saw the Naga Oil PR, but wasn’t sure if there was anything similar that has be integrated since…

With a dynamic compute + render pipeline assembler (even if it wasn’t quite exposed in the way this is), it would probably be a lot easier to add and maintain a larger constellation of whacky visualizer types (Dreams SDF particle splatting/raymarching, voxels, marching cubes, 3d texture raymarching, deformable meshes, raytraced refractive optical elements, etc.)

Wumpf:
Ah so this really was solely about cross compiling, gotcha. I could actually imagine to just leverage the shader transpiling features that Naga, wgpu's shader compiler, comes with. It can already take in GLSL/HLSL/MSL as well. Although keeping this as a pure sdk sided utility is probably better to avoid more idiosyncracies and having to enable more feature flags on Naga, further blowing up our binary side [...]


> I think WGSL also has reflection capabilities that can also sort of do this too? 

yep. And we have access to all of that already through Naga :)

re compute shader: we had to stay away from it so far because we still have to maintain a WebGL rendering path. This is also the reason for other idiosyncracies accross the code. I'm hoping we can drop that path some point later this year. But truth be told we generally haven't invested nearly as much time into our renderer as I would hoped we did at this stage.
WGSL compute shader support is generally more limited compared to cuda, but it is slowly becoming the safest, most portable and often even most well speced way of writing them (there's _a lot_ of hidden pitfalls in compute shading...). For everything rendering I'm fairly convinced it's the way to go for Rerun, but there might be heavy compute loads in the future where more platform specific things could make sense.

zalo:
Err, I should ask: Does Rerun have a sanctioned way for using custom shaders if they’re natively WGSL?  I promise I’m not _just_ trying to avoid WGSL with this PR 😅

Would the pure fragment shader volume visualizer make more sense to expose Python equivalent bindings for these?

https://github.com/rerun-io/rerun/pull/9820
https://rerun.io/docs/howto/visualization/extend-ui

Wumpf:
> Err, I should ask: Does Rerun have a sanctioned way for using custom shaders if they’re natively WGSL? I promise I’m not just trying to avoid WGSL with this PR 😅

Heh, no worries, your attempt here would still be the first hook that doesn't require doing a whole lot of rust building. The extend-ui doc page describes all sort of ways of how to extend the viewer by writing rust code - without forking, but still by essentially building the viewer yourself.
Sorry if I confused you there!

> Would the pure fragment shader volume visualizer make more sense to expose Python equivalent bindings for these?

I think eventually we should add some archetype/visualizer for volume visualizations that just works nice out of the box. But that ofc requires thinking about how the data is speced out and no one format will fit everyone. So I think some kind of escape hatch similar to what you're proposing here makes a lot of sense.

As mentioned, I haven't actually looked at any details here yet, but I'm sympathetic to the idea of adding it as an experimental feature. Likely also behind an experimental checkbox that one has to opt-in to the settings since I fear that is too easy for someone to shoot themselves into the foot and run into all sort of nasty portability issues, rendering artifacts, crashes and worse.

zalo:
https://github.com/user-attachments/assets/911e54b2-0df3-48bd-a60a-22908ff7c702

Oooh, I've almost got it... a few remaining issues, but they just seem to be shader bugs...  (this doesn't look right from any other angle 😅 )

zalo:
https://github.com/user-attachments/assets/954807cc-6ff5-4d9e-b654-9c5e79e52767

The shader works! 🥳  Wouldn't this be a fun demo for release...
Could probably put together a bunch of other fun ones too...

zalo:
https://github.com/user-attachments/assets/8fa81362-a9e7-43c9-835c-b4266509ebbe

(Deleted a few work in progress comments)  This PR now works as I originally envisioned!

I'm not terribly proud of how I had to mangle more uniforms onto the vertex shader yet, but I'll figure out a clean interface there at some point perhaps 😅

PR #12677, opened 2026-03-03
Fix selecting colormap by label in selection ui

oxkitsune:
### What

The colormap selection ui did not work when clicking the name of the colormap. This fixes that.


oxkitsune:
Already landed a fix in #[971](https://github.com/rerun-io/reality/pull/971)

PR #12682, opened 2026-03-09
feat(re_uri): Support path prefix in re_uri for sub-path reverse proxy routing

### LINKED ISSUE #10373: Preserve pathname prefix in `re_uri`

stefanwagnerdev:
### Related

* Closes #10373
* Part of #10409

### What

When connecting through a reverse proxy at a sub-path (e.g. rerun+http://host/my-prefix/proxy), the path prefix is dropped and gRPC connections go to the root URL instead.

This is a follow-up to #10409, reworked based on the review feedback there: the path prefix is now stored on Origin rather than on the endpoint structs, so origin.as_url() includes it automatically without needing to patch individual call sites. I'm using this branch successfully in my project which is served from https://HOST/cell/data-viewer

Includes the test case from @jprochazk in the #10409 review.

### DIFF
--- a/crates/store/re_uri/src/origin.rs
+++ b/crates/store/re_uri/src/origin.rs
@@ -2,7 +2,7 @@ use std::net::SocketAddr;
 
 use crate::{Error, Scheme};
 
-/// `scheme://hostname:port`
+/// `scheme://hostname:port[/path_prefix]`
 #[derive(
     Debug, PartialEq, Eq, Clone, Hash, PartialOrd, Ord, serde::Serialize, serde::Deserialize,
 )]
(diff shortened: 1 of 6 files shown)


stefanwagnerdev:
Hi @jprochazk thank you for the explanation. I think you are right and this needs a lot more knowledge about the complete codebase which I can't offer. 
As this fix works for me without visible side effects I will just fork it and hope that you can address this in the future. :)

PR #12689, opened 2026-03-11
Use GCS hosted raw_mesh files

Wumpf:
Fixes our sample builds!


Wumpf:
@rerun-bot reality-sync

PR #12690, opened 2026-03-11
Support focusing specific 3D points in viewer (#5382)

### LINKED ISSUE #5382: Double-click to focus on a point no longer works
**Describe the bug**

Double-clicking on a 3D point of a point cloud to refocus the camera no longer works.

- Version 0.12.1: it works ✅ (https://app.rerun.io/version/0.12.1)
- Version 0.13.0: it doesn't work ⛔ (https://app.rerun.io/version/0.13.0)
- Version 0.14.1: still doesn't work ⛔ (https://app.rerun.io/version/0.14.1)

**To Reproduce**
1. Go to an official example with a 3D point cloud (e.g. `Structure from Motion`)
2. Double-click on a 3D point on the side of the scene
3. The focus is made on the center of the 3D point cloud as a whole, not the individual point as before

**Expected behavior**

Up until v0.12.1, we could focus specific points of a 3D point cloud easily, and this is no longer the case.

**Screenshots**

N/A

**Backtrace**

N/A

 - OS: macOS Monterey 14.3.1
 - Browser: Firefox Version 123.0 (64-bit)

**Rerun version**

Versions 0.13.0 to 0.14.1

**Additional context**

N/A

bchretien:
### Related

* Closes #5382

### What

This is an attempt to solve #5382, that is the ability to center the camera on a specific 3D point when clicking (similar to how it was working before version 0.13.0) and not the whole entity. This is especially useful when analyzing large meshes or point clouds, where details may matter.

To do that, the focused item has now an optional `ItemContext` attached (see `FocusTarget` class that I can rename if necessary). If a 3D position is detected in the `ItemContext`, a new case is added to the dispatch logic to focus that point. For the new camera parameters, I went with the same forward vector, and the same distance to the point as when it was clicked, unless we're further away than the entity bounding sphere radius x 1.5. Thus, if we're far away we get the full entity in view while targeting the point, and if we were already close to the entity, we keep a similar distance to avoid large translations when clicking points that are nearby. UX is not my forte, so I'm not sure if that's the best approach for a wider user base.

I saw afterwards that the content of `FocusTarget` is kinda redundant with the `ItemCollection` internals, where `Item` and `ItemContext` are associated. Before attempting any cleanup, I'd appreciate some feedback :)

**DISCLAIMER**: as I'm not familiar with the rerun codebase, the heavy lifting was done by ChatGPT-5.4 Thinking, so there may be some inefficiencies there. I did not have much time to work on it, so I'm even surprised to have something that works for my use case at all :)

### DIFF
--- a/crates/viewer/re_viewer_context/src/focus_target.rs
+++ b/crates/viewer/re_viewer_context/src/focus_target.rs
@@ -1,3 +1,5 @@
+use re_log_types::EntityPath;
+
 use crate::{Item, ItemContext};
 
 /// One-shot focus payload.
@@ -18,5 +20,12 @@ impl From<Item> for FocusTarget {
             item,
             context: None,
         }
+    }
+}
+
+impl From<EntityPath> for FocusTarget {
+    #[inline]
+    fn from(entity_path: EntityPath) -> Self {
+        Self::from(Item::from(entity_path))
     }
 }
--- a/crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
+++ b/crates/viewer/re_blueprint_tree/src/blueprint_tree.rs
@@ -145,9 +145,10 @@ impl BlueprintTree {
             .show(ui, |ui| {
                 re_tracing::profile_scope!("blueprint_tree_scroll_area");
                 ui.panel_content(|ui| {
-                    self.blueprint_tree_scroll_to_item = ctx.focused_item().and_then(|item| {
-                        self.handle_focused_item(ctx, viewport_blueprint, ui, &item.item)
-                    });
+                    self.blueprint_tree_scroll_to_item =
+                        ctx.focused_item().and_then(|item| {
+                            self.handle_focused_item(ctx, viewport_blueprint, ui, &item.item)
+                        });
 
                     list_item::list_item_scope(ui, "blueprint tree", |ui| {
                         if let Some(root_container) = &blueprint_tree_data.root_container {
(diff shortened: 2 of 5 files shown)


crates/viewer/re_viewer_context/src/focus_target.rs
```diff
@@ -0,0 +1,22 @@
...
+
+/// One-shot focus payload.
+///
+/// Unlike selection, focus is cleared every frame. The optional context lets a receiver
+/// distinguish between “focus the whole entity” and “focus the exact 3D point that was
+/// double-clicked”.
+#[derive(Clone, Debug, PartialEq)]
+pub struct FocusTarget {
+    pub item: Item,
+    pub context: Option<ItemContext>,
```
Wumpf: nice, taking an optional context like selection itself makes sense I think

Wumpf:
sorry about the long delay, still on my radar, just lots of other things to tend to right now!

bchretien:
@Wumpf don't worry, take your time, there's no rush ;)

Wumpf:
oh and thanks for writing such a well motivated description which explains _why_ you're doing things a certain way. Kinda rare these days to see that with all those agents around 😅

Wumpf:
@rerun-bot reality-sync

IsseW:
Nice! Think the code in eye.rs looks good.

Trying it I noticed that double clicking an instance also still selects the whole entity. So feels like there's a disconnect between what is focused and what is selected.

A suggestion from @gavrelina was to kind of have this be a rolling event, which cycles between selecting & focusing a single point vs selecting & focusing the whole entity on double click.

Meaning:
- If only the instance is selected and it's double clicked, we select & focus the whole entity.
- Otherwise we select & focus the given instance on double click.

Wumpf:
@bchretien sorry for the long hold-up. We talked about this today internally and figured we'll take your contribution as is, I'll resolve the conflicts and shepard it in :)

Wumpf:
@rerun-bot reality-sync

Wumpf:
@rerun-bot reality-sync

Wumpf:
@rerun-bot reality-sync

Wumpf:
some trouble with our internal mono repo, but as ⬆️ says, the fix got in :)

PR #12692, opened 2026-03-13
Remove anyhow from store-layer crates

vaporif:
### Related

- Part of #1845

### What
Replace anyhow with thiserror in re_query, re_chunk_store, and re_entity_db library crates.

### DIFF
--- a/crates/store/re_chunk_store/src/lib.rs
+++ b/crates/store/re_chunk_store/src/lib.rs
@@ -78,11 +78,20 @@ pub enum ChunkStoreError {
     #[error(transparent)]
     Chunk(#[from] re_chunk::ChunkError),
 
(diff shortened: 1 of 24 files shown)


crates/store/re_chunk_store/src/lib.rs
```diff
@@ -78,12 +78,21 @@ pub enum ChunkStoreError {
     #[error(transparent)]
     Chunk(#[from] re_chunk::ChunkError),
 
-    #[error("Failed to load data, parsing error: {0:#}")]
-    Codec(#[from] re_log_encoding::CodecError),
```
vaporif, line 82: Was not being used anywhere. Safe to remove or still should have stayed?

Wumpf:
quite outdated by now unfortunately, that's on us for not processing this in time

PR #12696, opened 2026-03-18
Fix(web-viewer): clean up loader when startup is interrupted

### LINKED ISSUE #12695: web-viewer-react can leave a stale "Loading Rerun…" overlay after remount/unmount during async start
**Describe the bug**

`@rerun-io/web-viewer-react` can leave a stale `Loading Rerun…` overlay in the DOM if the viewer is unmounted or remounted while `WebViewer.start()` is still in progress.

In my repro, the underlying viewer actually starts successfully: a canvas is present, the console prints `Web app started`, and the `.rrd` finishes decoding. However, the loading overlay remains on top permanently, so the page looks like it is stuck loading.

This appears to be a lifecycle race between the async `start()` path in `@rerun-io/web-viewer` and `stop()` being called from the React wrapper during remount/unmount.

**To Reproduce**
1. [demo](https://stackblitz.com/edit/vitejs-vite-ylbpmhbq?file=src%2FApp.tsx)

**Expected behavior**

If the initial async startup is interrupted by unmount/remount, the loading overlay should always be cleaned up. Remounting the viewer should not leave a stale loader covering the canvas.

**Screenshots**

Image

**Backtrace**

No crash / no Rust backtrace. This is a stuck loading overlay rather than a process crash

 - OS: macOS 26.3.1

**Rerun version**

"@rerun-io/web-viewer-react": "0.30.2",

**Additional context**

I believe the bug is in the loader lifecycle inside @rerun-io/web-viewer:

- WebViewer.start() creates the loader as a local variable and appends it before async startup work begins.
- start() then awaits multiple async steps (delay(0), load(...), handle.start(...)).
- There are early-return checks after async steps if this.#state !== "starting".
- If React unmounts/remounts the component during that window, the wrapper calls stop(), which sets the state to "stopped" and removes the canvas/handle.
- When the original start() resumes, it hits the early return and exits before loader.remove() runs.
- stop() does not remove the loader, so the stale overlay remains in the DOM.

The React wrapper makes this easy to hit because it starts the viewer in componentDidMount() and calls stop() in componentWillUnmount(), without waiting for the async start() call to finish.

So the page looks like it is "stuck loading", but in practice the viewer may already be running under a leaked overlay.

I have a local fix and can open a PR for this.

Woodii1998:
### Related
Closes #12695

### What
This fixes a loader lifecycle race in `rerun_js/web-viewer`.

`WebViewer.start()` creates and appends the `Loading Rerun…` overlay before several async startup steps. If `stop()` is called during that window, the in-flight `start()` can later resume, detect that the state is no longer `"starting"`, and return early without removing the loader. In React integrations, this can happen during remounts or unmounts, leaving a stale overlay covering a working canvas.

- Promoted the loader from a local `start()` variable to an instance field
- Added a shared `#clearLoader()` helper
- Cleared the loader on all startup exit paths:
  - `load()` failure
  - `handle.start()` failure
  - early return after interrupted startup
  - normal startup completion
- Cleared the loader from `stop()` as well

### DIFF
--- a/rerun_js/web-viewer/index.ts
+++ b/rerun_js/web-viewer/index.ts
@@ -344,11 +344,6 @@ export class WebViewer {
     setupGlobalEventListeners();
   }
 
-  #clearLoader() {
-    this.#loader?.remove();
-    this.#loader = null;
-  }
-
   /**
    * Start the viewer.
    *
@@ -716,6 +711,11 @@ export class WebViewer {
     }
 
     this.stop();
+  }
+
+  #clearLoader() {
+    this.#loader?.remove();
+    this.#loader = null;
   }
 
   /**


rerun_js/web-viewer/index.ts
```diff
@@ -343,6 +344,11 @@ export class WebViewer {
     setupGlobalEventListeners();
   }
 
+  #clearLoader() {
```
jprochazk: Please move this method down with the other private methods.

Woodii1998:
@jprochazk  
done 

By the way, unrelated to this PR: while working in `rerun_js`, I noticed that `WebViewer.#id` is initialized from `randomId()` in `rerun_js/web-viewer/index.ts`, I couldn’t find any later reads of that field.

https://github.com/rerun-io/rerun/blob/f4ec69088b7ccf1276ea20bb62cf3be3d056c730/rerun_js/web-viewer/index.ts#L332

 More generally, it feels like the repo already has some baseline formatting conventions, but the JS/TS side in `rerun_js` is not enforced very consistently yet. In practice, format-on-save can introduce noisy, non-functional diffs, which makes small changes harder to keep focused.

Not something to handle in this PR, but it might be worth tightening or standardizing the JS/TS lint/format setup a bit more so future contributors can make changes without fighting style drift.

jprochazk:
@rerun-bot reality-sync

jprochazk:
> In practice, format-on-save can introduce noisy, non-functional diffs, which makes small changes harder to keep focused.

You can run `pixi run misc-fmt` to format files. That should probably be mentioned somewhere closer to the JS code. Note that there aren't any formatting issues in your PR right now.

The CI failure is unrelated to your changes

jprochazk:
Thanks!

Issue #10090, opened 2025-05-27
Support h.264/h.265 b-frames on `VideoStream` archetype


Wumpf:
Our `VideoAsset` archetype has no issues with b-frames at all. However, the (as of writing upcoming) `VideoStream` archetype won't support b-frames in its first few iterations.

B-frames for h.264/h.265 create additional complexity as they require differing decode timestamps (DTS) and presentation timestamps (PTS):
We have to add decode timestamps as a separate optional field and take care of the necessary reordering. An additional issue is that currently we assume that all video frames come in-order (rather special as all our other logging it out-of-order!). This greatly complicates access into rerun chunks containing frame data.

As encoders emit b-frames alongside both prior and successive frames, one way to solve this is to extend `VideoStream` to have _arrays_ of presentation time offsets and allow several frames to be in a single sample. This way, the stream of samples is still ordered in decoding order, keeping decoder ingestion simple. Since any use of b-frames naturally causes significant latency, this solution is likely truthful to the actual output.
(Note, that none of the decoders we worked with so far actually needs the actual value of the DTS, so we're not missing this information anywhere!)

See also:
* https://github.com/rerun-io/rerun/issues/7484

TODO: What about bframe support on AV1 and VP9? We probably support this fine already, as these codecs don't have the sample ordering issue AVC/HEVC have. But we should test it!

Wumpf:
@jleibs proposed an `decoder_timeline` field for `VideoStream` archetype. His motivation was more about the consequences of out of order logging of video data, but I think this could also be a great solution for the decode ordering issues of videos!

PR #12700, opened 2026-03-20
Video stream bframe support

caiopiccirillo:
### Related

* Closes [#1234](https://github.com/rerun-io/rerun/issues/10090)

### What

Adds B-frame support for H.264/H.265 streams on the `VideoStream` archetype.

#### New types:

`VideoPresentationTimestampOffset` datatype (transparent i64) and matching component — represents the per-sample offset from decode timestamp (DTS) to presentation timestamp (PTS), in timeline units.

#### Archetype changes (`VideoStream`):

`sample` field is now array-typed `VideoSample`, allowing multiple frames to be batched into a single row in decode order.

New optional `presentation_time_offset` field `VideoPresentationTimestampOffset — when present, PTS for sample i = DTS_i + offset_i. When absent, PTS == DTS (the common no-B-frame case, fully backward compatible).

#### Cache/decoder pipeline (`VideoStreamCache`):

Frame numbers are now assigned by presentation order (PTS-sorted) instead of decode order.
Sample durations are computed from PTS-sorted timestamps so scrubbing and seek work correctly.

read_pts_offsets_from_chunk extracts offset arrays from incoming chunks.

Incremental chunk additions properly update frame numbers and SamplesStatistics.
demux/mod.rs: video duration is now computed from actual min/max PTS to handle reordered frames.


Wumpf:
Thanks for wanting to contribute! This is a fairly complex and large piece that comes with quite some design choices & testing needs. Your description seems to be entirely generated, which can be fine but in the future make sure to add your own thoughts about the tradeoffs involved, what's good, what's bad, how this was tested, some screenshots & examples etc.

Either way we don't have time to process that right now, so I'll put it on draft

See also https://github.com/rerun-io/rerun/blob/main/CONTRIBUTING.md#what-to-contribute

Wumpf:
> Your description seems to be entirely generated, which can be fine 

To clarify, I don’t care that it’s generated but I deeply care about it being reviewable and indicating that thought went into the things where thought is required! Right now that's just the usual "agent did things" listing which doesn't help understanding the relevant pieces much

caiopiccirillo:
@Wumpf Sorry for the generated description. I was in a hurry yesterday and thought it was more important to review the code with a concise description.

But I took some time to describe the decisions/considerations that were made in the design.

I chose to implement DTS on the timeline with PTS offset per sample. The timeline timestamp is the decoding timestamp, but an optional field (presentation_time_offset, which is an int64 with the timeline unit) was added that gives the delta PTS - DTS per sample, and when  is absent, PTS == DTS enabling backward compatibility at no cost for no B-frames.

Two alternatives were considered for the problem, the first being the one suggested in the issue itself, which would be to add the decoder_timeline, but I see that it solves another problem that can be addressed in the future (which would be if the user logs the samples out of decoding order), but the current implementation assumes that the samples arrive in DTS order on the timeline. The second alternative would be to separate the DTS and PTS fields, but I thought that wouldn't be efficient in the sense of duplicating the metadata, and the decoder doesn't need the DTS.

Making `sample` an array was necessary because B-frame encoders naturally output batches, thus allowing multiple samples per row and keeping the sample stream in decoding order within a log call (which aligns with how encoder outputs data).

I see some limitations in my implementation:
1. I haven't implemented the decoder_timeline field.
2. I haven't tested for AV1/VP9.
2. The frame number assignment is done by PTS order, which can be negligible for "typical" streams, but is worth noting for very large batches.

Finally, I tested using unit tests for the cases I found relevant, but I also did a manual e2e test using an H.264 file (re-encoded with libx264 with max_b_frames=3) and streaming it through the Python SDK and it seems to be working.

I'm open to discuss further if changes are needed :smile:

PR #12702, opened 2026-03-23
Remove anyhow from `re_ros_msg` and `re_arrow_util`

### LINKED ISSUE #1845: Replace `anyhow` with `thiserror` in libraries

emadak47:
### Related
* Part of #1845 

### What
- removed `anyhow` from public fns in `utils/` crates

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8269,7 +8269,6 @@ dependencies = [
 name = "re_arrow_util"
 version = "0.31.0-alpha.1+dev"
 dependencies = [
- "anyhow",
  "arrow",
  "comfy-table",
  "datafusion",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9507,7 +9507,6 @@ dependencies = [
 name = "re_ros_msg"
 version = "0.31.0-alpha.1+dev"
 dependencies = [
- "anyhow",
  "serde",
  "thiserror 2.0.18",
 ]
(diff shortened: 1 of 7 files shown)


grtlr (changes requested):
Thank you for your PR! The motivation behind

* #1845

is to have more structured error reporting that can replace the functionality provided by `anyhow` while allowing downstream processing. Yet this PR does not add any structure beyond the single variant `ArrayComparisonError`.

For example, I would expect an `Error::ChildCountMismatch` (or similar) variant in `re_arrow_util`. The ROS-related errors go into the right direction.

emadak47:
thank you for the feedback. 

originally, I focused on current usage; since no callers match on variants, I went for a simple unit struct `ArrayComparisonError`. But that's not better than `anyhow::Error`, it's a string with a different hat. 

I've added as many error variants as I deemed necessary, reducing further would lose meaningful distinctions, but I'm happy to hear your opinion if you disagree.

grtlr:
Closing because it went stale. Please feel free to re-open after addressing the feedback.

PR #12703, opened 2026-03-23
Update README.md with working `run_wasm` example

filonik-ups:
The example is in a separate crate now, changed the build command to reflect that.


filonik-upsaclay:
Sorry, fails too many checks, I will report an issue instead.

PR #12704, opened 2026-03-23
Fix world grid drawing for orthographic cameras

s-nie:
It seems #12444 did not consider orthographic cameras. This adds support for those again by adding the camera ray origin offset in those cases.

### DIFF
--- a/crates/viewer/re_renderer/shader/debug_overlay.wgsl
+++ b/crates/viewer/re_renderer/shader/debug_overlay.wgsl
@@ -9,6 +9,7 @@
 
 #import <./types.wgsl>
 #import <./global_bindings.wgsl>
+#import <./utils/camera.wgsl>
 
 struct UniformBuffer {
     screen_resolution: vec2f,
@@ -42,7 +43,7 @@ fn main_vs(@builtin(vertex_index) vertex_index: u32) -> VertexOutput {
     // But this is not the point here - we want to debug this and other things rapidly by editing the shader.
     let screen_fraction = texcoord * (uniforms.extent_in_pixel / uniforms.screen_resolution) +
                         uniforms.position_in_pixel / uniforms.screen_resolution;
-    let screen_ndc = vec2f(screen_fraction.x * 2.0 - 1.0, 1.0 - screen_fraction.y * 2.0);
+    let screen_ndc = screenuv_to_ndc(screen_fraction);
 
     var out: VertexOutput;
     out.position = vec4f(screen_ndc, 0.0, 1.0);


crates/viewer/re_renderer/shader/utils/camera.wgsl
```diff
@@ -44,24 +44,37 @@ fn camera_ray_to_world_pos(world_pos: vec3f) -> Ray {
...
+// Converts screen UV coordinates (0 to 1) to Normalized Device Coordinates (-1 to +1)
+fn screenuv_to_ndc(texcoord: vec2f) -> vec2f {
+    return vec2f(texcoord.x - 0.5, 0.5 - texcoord.y) * 2.0;
+}
+
+// Converts pixel coordinates to Normalized Device Coordinates (-1 to +1)
+fn fragcoord_to_ndc(fragcoord: vec2f) -> vec2f {
+    let texcoord = fragcoord / frame.framebuffer_resolution.xy;
+    return screenuv_to_ndc(texcoord);
+}
```
Wumpf, line 64: good utils to have around, I bet there could be a bunch of other places where this could be used

crates/viewer/re_renderer/shader/utils/camera.wgsl
```diff
@@ -77,8 +90,23 @@ fn camera_ray_direction_from_fragcoord(fragcoord: vec2f) -> vec3f {
...
+        // Compute viewport extent in world space
+        let viewport_world_size = frame.pixel_world_size_from_camera_distance * frame.framebuffer_resolution;
+        let view_space_offset = vec3f(ndc * viewport_world_size * 0.5, 0.0);
+
+        // Offset the ray origin based on the fragment position in the camera plane
+        ray.origin = frame.camera_position + view_space_to_world_space(view_space_offset);
+    } else {
+        // For perspective cameras, all rays originate from the camera position
+        ray.origin = frame.camera_position;
+    }
```
Wumpf, line 108: generating the camera ray should be another utility - otherwise all camera rays everywhere else are just as wrong

Wumpf: actually, taking a step back here: we should change `frame.camera_position` to mean the position that was computed here. I can't imagine a case where a shader would be any other one 🤔 

Please document the adjusted semantics of the field to be the center of the near ortho plane for orthographic cameras)
(unless you find that this doesn't make sense for some reason)

s-nie: As I understand it, `camera_position` is part of `FrameUniformBuffer`, a uniform is computed on the CPU. So it's the shader's job to compute the actual per-fragment position. I don't feel comfortable making a change to rework this, at least not in this PR.

But you're right, it's used in multiple places, in some of which the actual ray origin may be a better fit. I opened https://github.com/rerun-io/rerun/issues/12706 as a reminder.

Wumpf: ah right I got confused. the code very explicitely finds where the position is on the plane in space and not just the center of said plane, yeah let's leave the uniforms alone 👍

crates/viewer/re_renderer/shader/utils/camera.wgsl
s-nie: Found another NDC computation and replaced it: https://github.com/rerun-io/rerun/pull/12704/commits/1ca8f7612bbb151197e1b3ec1d6f2b4032fa91af

Wumpf (approved):
nice, thanks for the followup. Yeah that will do I think, I was definitely wrong with my uniform-comment.

cc: @MichaelGrupp fairly sure this will fix at least some of the issues you had your prototype for a rerun orthographic camera mode

Wumpf:
putting 'exclude from changelog' since so far we don't have dedicated changelog for things that only affect re_renderer itself and only do changes about the things that affect the released artifacts (i.e. sdk & viewer)

Wumpf:
@rerun-bot reality-sync

PR #12705, opened 2026-03-24
Enable `clippy::iter_over_hash_type` in remaining viewer crates

emadak47:
### Related

* Part of #6198 

### What

  - Enable `#![warn(clippy::iter_over_hash_type)]` in 17 viewer crates
       - 12 crates with no violations and 5 crates with violations
  - Suppress 16 sites where, in my understanding, iteration order is irrelevant
  - Fix one site in `re_renderer/src/file_server.rs` where `?` made the iteration order-dependent

### DIFF
--- a/crates/viewer/re_renderer/src/file_server.rs
+++ b/crates/viewer/re_renderer/src/file_server.rs
@@ -212,10 +212,15 @@ mod file_server_impl {
                     .with_context(|| format!("couldn't resolve imports for file at {path:?}"))?
                     .imports;
 
+                #[expect(clippy::iter_over_hash_type)]
+                // Each import path is watched independently; errors are logged, not propagated.
                 for path in imports {
-                    self.watcher
+                    if let Err(err) = self
+                        .watcher
                         .watch(path.as_ref(), RecursiveMode::NonRecursive)
-                        .with_context(|| format!("couldn't watch file at {path:?}"))?;
+                    {
+                        re_log::error!(%err, ?path, "couldn't watch imported dependency");
+                    }
                 }
             }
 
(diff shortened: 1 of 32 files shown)


crates/viewer/re_renderer/src/file_server.rs
```diff
@@ -212,10 +212,15 @@ mod file_server_impl {
...
+                // Each import path is watched independently; errors are logged, not propagated.
                 for path in imports {
-                    self.watcher
+                    if let Err(err) = self
+                        .watcher
                         .watch(path.as_ref(), RecursiveMode::NonRecursive)
-                        .with_context(|| format!("couldn't watch file at {path:?}"))?;
+                    {
+                        re_log::error!(%err, ?path, "couldn't watch imported dependency");
+                    }
```
emadak47: if one path fails, the set of paths watched before the failure depends on iteration order, so I replaced with error logging which matches the pattern at line 279

Wumpf: that's a setup failure though whereas in ln279 it's a runtime issue on collecting the latest updates

Wumpf (changes requested):
sorry for the long delay, this fell a bit through the cracks!

Almost wanted to push back to make it opt-out now already, but there's a bit too many occasions yet in there and it's continuing a status quo pattern we have there.

So all that's left is to get it up to date again and roll back the failure change (which seems to be unrelated)

Wumpf (approved):

Wumpf:
@rerun-bot reality-sync

Wumpf:
@rerun-bot reality-sync

PR #12707, opened 2026-03-24
Improve text log view performance with cached indexing and virtualized rendering

vaishanth-rmrj:
### Related

- No linked issue yet.

### What

Improve text log view performance for large numbers of log entries by replacing the eager per-frame expansion path with a cached, virtualized one.

Previously, the text log view eagerly expanded all matching `TextLog` data into a flat in-memory list, sorted the entire list, recalculated row heights from full text bodies, and rendered through the old heterogeneous row path. That approach works for small recordings, but it becomes expensive when a recording contains many log rows or large multi-line messages.

This PR introduces a store-level `TextLogCache` plus a per-view `TextLogProjectionState` so the view can work from lightweight row metadata instead of eagerly materializing every visible row body on every frame.

Main changes:
- Add `TextLogCache`, which indexes text-log chunks once and stores lightweight per-row metadata:
  - row id / instance id
  - timepoint
  - log level
  - color
  - explicit line count
  
- Support incremental append-only updates in the cache, while falling back to a rebuild for non-additive store edits.

- Add a per-view projection that:
  - tracks the currently included entities and active timeline
  - preserves the legacy stable row ordering
  - applies log-level filtering over cached row handles
  - caches prefix sums for multi-line row heights
  - reuses cached discovered levels for blueprint fallback UI
  
- Switch rendering to `egui_table` with a table delegate that:
  - resolves text bodies lazily for visible rows only
  - uses cached prefix sums for row positioning
  - preserves current-time autoscroll behavior
  - preserves current-time marker rendering
  
- Simplify `TextLogSystem` so it remains responsible for discoverability of `TextLog` entities, while the view owns cached loading/projection logic.

- Add unit tests covering:
  - sparse carry-forward behavior for level/color metadata
  - lazy body resolution from indexed chunks
  - additive vs non-additive cache revisions
  - explicit newline counting
  - filtered projection prefix sums
  - sorted merge behavior
  - scroll target behavior
  - current-time marker placement

This is intended to preserve existing user-visible behavior while making large text-log views much more responsive.

Local validation:
- `cargo check -p re_view_text_log`
- `cargo test -p re_view_text_log`

Both passed locally.

Breaking changes:
- None.

Migration guide:
- Not needed.

### DIFF
--- a/crates/viewer/re_view_text_log/Cargo.toml
+++ b/crates/viewer/re_view_text_log/Cargo.toml
@@ -19,6 +19,8 @@ workspace = true
 all-features = true
 
 [dependencies]
+re_byte_size.workspace = true
+re_chunk.workspace = true
 re_chunk_store.workspace = true
 re_data_ui.workspace = true
 re_entity_db.workspace = true
@@ -32,8 +34,8 @@ re_ui.workspace = true
 re_viewer_context.workspace = true
 re_viewport_blueprint.workspace = true
 
-egui_extras.workspace = true
 egui.workspace = true
+egui_table.workspace = true
 itertools.workspace = true
 
 [dev-dependencies]
--- a/crates/viewer/re_view_text_log/src/cache.rs
+++ b/crates/viewer/re_view_text_log/src/cache.rs
@@ -1,0 +1,503 @@
+use std::cmp::Ordering;
+use std::collections::BTreeMap;
+use std::sync::Arc;
+
+use re_byte_size::{MemUsageTree, MemUsageTreeCapture, SizeBytes};
+use re_chunk::{Chunk, ChunkId, RowId};
+use re_chunk_store::ChunkStoreEvent;
+use re_entity_db::EntityDb;
+use re_log_types::{EntityPath, TimeInt, TimePoint, TimelineName};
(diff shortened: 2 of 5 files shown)


vaishanth-rmrj:
Hi @oxkitsune, any updates on this PR? Let me know if there are any issues. Thanks!

oxkitsune:
First off, apologies for the slow turnaround on this, and thanks for your patience.

The code doesn't compile as provided, so I had to patch it before I could run anything. Once I got it running, I noticed a deadlock on static text chunks! That should be fixed!

Direction-wise this is good, and I'm happy to move this forward once it's in a better state.
The PR mentions this was validated locally, but given the viewer doesn't compile as-is, I'm curious whether you tested the view itself? Just want to make sure we're testing the same thing.1

Thanks again!

Wumpf:
@vaishanth-rmrj are you still interested in working on this?

Wumpf:
@oxkitsune now landed a much more lightweight fix https://github.com/rerun-io/rerun/commit/98f0297aebe18b5362d8f677dd7af2d2f306d7e0

Perf could still be better but I think the now established cache is a much simpler starting point

PR #12719, opened 2026-04-02
Handle empty ROS2 camera `frame_id` without panicking

### LINKED ISSUE #12718: MCAP: empty ROS2 camera frame_id causes TransformForest panic instead of falling back to 2D image import
**Describe the bug**

When opening a ROS2 MCAP file that contains camera-related messages with an empty `header.frame_id`, the viewer crashes in `re_tf` instead of degrading gracefully.

  This is not just malformed JS/WASM behavior. The underlying problem is that empty camera frame IDs are currently treated as real transform frames, which can produce invalid pinhole /
  coordinate-frame topology such as a self-referential pinhole (`"" -> ""`) and shared empty frames across multiple camera entities.

  In practice, missing `frame_id` is common enough in MCAP recordings that it should not be fatal. The affected camera should fall back to being imported as a normal 2D image, with a
  warning, rather than crashing the viewer.

**To Reproduce**

1. Prepare a ROS2 MCAP file containing camera topics such as `sensor_msgs/CameraInfo`, `sensor_msgs/Image`, or `sensor_msgs/CompressedImage`.
2.  Ensure one or more of those messages have `header.frame_id == ""` .
3. Open the MCAP file in the Rerun viewer / web viewer.
4. Observe that the viewer panics while building the transform forest.

**Expected behavior**

  Missing camera `frame_id` should be treated as missing metadata, not as a valid transform frame.

  Expected fallback behavior:
  - `CameraInfo` with missing `frame_id` should be ignored for spatial camera setup.
  - `Image` / `CompressedImage` should still import and display as ordinary 2D images.
  - The viewer should emit a warning instead of panicking.

**Screenshots**

Image

**Backtrace**

  ```text
  re_viewer.js:1249 panicked at crates/store/re_tf/src/transform_forest.rs:363:17:
  DEBUG ASSERT: Root was added already at Some(TransformFrameId("")) as Some(TransformFrameRoot)

  Stack:

  Error
      at __wbg_new_72e7b12a3de206c6 (re_viewer.js:2232:29)
      at re_viewer.wasm.eframe::web::panic_handler::Error::new::__wbg_new_72e7b12a3de206c6::h2405d2c308c7c8ee externref shim (re_viewer.wasm-1fb94d82:0x635670a)
      at re_viewer.wasm.eframe::web::panic_handler::PanicSummary::new::h5975acdcf013b7c8 (re_viewer.wasm-1fb94d82:0x54b811d)
      at re_viewer.wasm.eframe::web::panic_handler::PanicHandler::install::{{closure}}::hcaf2ce3c7e3f00cd (re_viewer.wasm-1fb94d82:0x429ce0c)
      at re_viewer.wasm.std::panicking::panic_with_hook::hf90632d5c7102557 (re_viewer.wasm-1fb94d82:0x5df3f54)
      at re_viewer.wasm.std::panicking::panic_handler::{{closure}}::he5eb13023b639916 (re_viewer.wasm-1fb94d82:0x5fb74b3)
      at re_viewer.wasm.std::sys::backtrace::__rust_end_short_backtrace::hdaee13ec4dd4f348 (re_viewer.wasm-1fb94d82:0x6395188)
      at re_viewer.wasm.__rustc[d9b87f19e823c0ef]::rust_begin_unwind (re_viewer.wasm-1fb94d82:0x62ebfb1)
      at re_viewer.wasm.core::panicking::panic_fmt::hcb6b2b4be1f4be38 (re_viewer.wasm-1fb94d82:0x62ec03b)
      at re_viewer.wasm.re_tf::transform_forest::TransformForest::new::h577489312fa29c83 (re_viewer.wasm-1fb94d82:0xd2274d)
```

 - OS:  macOS 26.4 (25E246)

**Rerun version**

 Local build from main (commit: 4f1fbee83cf4407e6092b8ef81c74df50d59fe1b)

**Additional context**

 I inspected the MCAP file and confirmed that the relevant camera messages had empty frame_id values. The file itself was readable with mcap info; the issue is specifically in how missing camera frame metadata is imported.

 Root cause chain:

  - CameraInfo currently forwards header.frame_id directly into pinhole transform setup.
  - suffix_image_plane_frame_ids("") preserves the empty string, so this degenerates into:
      - parent_frame = ""
      - child_frame = ""
  - Image / CompressedImage can also map to the same empty frame.
  - Multiple camera entities may then reuse the same empty frame ID.
  - Later, TransformForest hits an invalid topology condition and panics.

  I think the importer should explicitly treat trim().is_empty() as “missing frame”, skip spatial camera setup for that topic, and keep images as plain 2D content.

Woodii1998:
### Related
Closes #12718

### What

Empty `header.frame_id` in ROS2 camera messages caused a `TransformForest` panic due to self-referential pinhole topology. At the MCAP parser level, empty frame IDs now trigger a graceful downgrade: `CameraInfo` import is skipped entirely, while `Image` and `CompressedImage` fall back to plain 2D import without spatial metadata. As a defense-in-depth measure, `TransformForest` now also ignores self-referential pinhole projections instead of panicking.

### DIFF
--- a/crates/store/re_mcap/src/parsers/ros2msg/sensor_msgs/camera_info.rs
+++ b/crates/store/re_mcap/src/parsers/ros2msg/sensor_msgs/camera_info.rs
@@ -3,7 +3,7 @@ use re_sdk_types::archetypes::{CoordinateFrame, Pinhole};
 
 use super::super::Ros2MessageParser;
 use super::super::definitions::sensor_msgs;
-use super::super::util::suffix_image_plane_frame_ids;
+use super::super::util::spatial_camera_frame_ids_or_log_missing;
 use crate::Error;
 use crate::parsers::cdr;
 use crate::parsers::decode::{MessageParser, ParserContext};
@@ -73,24 +73,29 @@ impl MessageParser for CameraInfoMessageParser {
         } = *self;
 
         let entity_path = ctx.entity_path().clone();
+        let Some(frame_ids) = spatial_camera_frame_ids_or_log_missing(
+            ctx.channel_topic(),
+            &entity_path,
+            "sensor_msgs/msg/CameraInfo",
+            "Skipping camera calibration import for this topic.",
+            frame_ids,
+        ) else {
+            return Ok(Vec::new());
+        };
         let timelines = ctx.build_timelines();
-
-        // We need a frame ID for the image plane. This doesn't exist in ROS,
-        // so we use the camera frame ID with a suffix here (and in the image parsers).
-        let image_plane_frame_ids = suffix_image_plane_frame_ids(frame_ids.clone());
 
         let mut components: Vec<_> = Pinhole::update_fields()
             .with_many_image_from_camera(image_from_cameras)
             .with_many_resolution(resolutions)
-            .with_many_parent_frame(frame_ids.clone())
-            .with_many_child_frame(image_plane_frame_ids)
+            .with_many_parent_frame(frame_ids.camera_frame_ids.clone())
+            .with_many_child_frame(frame_ids.image_plane_frame_ids)
             .columns_of_unit_batches()
             .map_err(|err| Error::Other(anyhow::anyhow!(err)))?
             .collect();
 
         components.extend(
             CoordinateFrame::update_fields()
-                .with_many_frame(frame_ids)
+                .with_many_frame(frame_ids.camera_frame_ids)
                 .columns_of_unit_batches()
                 .map_err(|err| Error::Other(anyhow::anyhow!(err)))?,
         );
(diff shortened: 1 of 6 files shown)


MichaelGrupp:
I think this should be handled on `re_tf` level instead of MCAP workarounds. An empty string can also come from elsewhere (ping @Wumpf: let me know if you agree). I'm a bit surprised that we are panicking, need to reproduce this myself (next week). Last time I looked into empty frame IDs was for https://github.com/rerun-io/rerun/commit/0e66bc0461e86ce6178f4e78fdcfcfba6f58884a

oxkitsune:
Related: There's #12663 that also solves this on the TF level

Woodii1998:
> I think this should be handled on `re_tf` level instead of MCAP workarounds. An empty string can also come from elsewhere (ping @Wumpf: let me know if you agree). I'm a bit surprised that we are panicking, need to reproduce this myself (next week). Last time I looked into empty frame IDs was for [0e66bc0](https://github.com/rerun-io/rerun/commit/0e66bc0461e86ce6178f4e78fdcfcfba6f58884a)

I agree with handling this in re_tf so that empty frame ids can never panic the transform system, regardless of the source.

But I think [#12663](https://github.com/rerun-io/rerun/pull/12663) goes a bit too far: it does not just make the system robust, it reinterprets empty frame ids as implicit/entity-derived frames. That changes the meaning of the data.

For example, if a camera topic has frame_id = "", I would rather treat that as missing spatial metadata and degrade it to plain 2D import. Under #12663, it instead starts behaving like a valid implicit frame, which hides the fact that the original frame metadata was missing.

So my preference would be:

re_tf handles robustness globally;
importers keep schema-specific fallback behavior where missing frame metadata has a well-defined meaning.

PR #12722, opened 2026-04-06
Spacebar toggles play/pause, never enables following

### LINKED ISSUE #12699: Spacebar play/pause is confusing
As I understand it, there's two modes for the spacebar:
1. If previously in "following" -> "pause", pressing again goes back to "following"
2. If previously in "play" -> "pause", pressing again goes back to "play"

I find this to be pretty frustrating when using rerun. I'm often looking at a log and scrubbing around the timeline after the data has stopped coming in, and I find an interesting moment and I hit spacebar (trying to see the behavior) and it jumps to the end of the timeline. I grumble, click somewhat randomly on the play/pause buttons to get it into the "good mode", find the event again, and then I'm happy. This happens to me _every time_ I record a log.

I get that going back to following mode should be easy, but I think the `>|` button makes sense for that (or perhaps manually scrubbing to the end of the timeline while playing?). Or the `Cmd+Right` suggestion below.

My personal vote is that spacebar should always pause if playing or following, and always go to playing if paused.

> Restarting should stop the "follow" behavior, but doesn't. Let's fix that first.
> 
> 
> 
> I also think we should have a shortcut for entering follow mode, like `Cmd+Right` (mirroring `Cmd+Left` for restart).
> 
> 
> 
> Pressing space/ toggles between the current mode and pause. If you are in play mode, it will toggle play->pause->play. If you are in follow it will toggle follow->pause->play. Why?
> 
> 
> 
> When following live data it is sometimes useful to pause to look at something, and then you want to _keep following_. Hitting the pause button again achieves that. It could otherwise be easy to think you are still following the live data when actually you are actually a few seconds behind.
> 
> 
> 
> I do agree that the current behavior is a bit confusing. I could be convinced to change this behavior so that pressing space twice would go from follow->pause->play, but then perhaps we should have a little notification popup "No longer following the live data" or something. …or we just switch the behavior and see if anyone complains. 

 _Originally posted by @emilk in [#1964](https://github.com/rerun-io/rerun/issues/1964#issuecomment-1523395287)_

wolfd:
### Related

Closes #12699

### What

This change removes what I believe to be very confusing behavior for the spacebar that constantly trips me up. While not a large sample size, I asked my coworkers if this also bothered them, and the common response I got was "yes this is confusing, but I've learned to never press spacebar at any point." For whatever reason, I haven't been able to learn that, and so I kept losing my location in the timeline.

This change does not add a way to get to following mode, because it's already been added since the original discussion happened: Alt+→

I removed a big explanation comment that justified the current behavior, if you deem that there should be a replacement comment, I can add one.


grtlr:
@rerun-bot reality-sync

gavrelina:
hey @wolfd thanks for this PR, 
based on [this](https://github.com/rerun-io/rerun/issues/12699) issue you've posted we also saw there were actually a few bugs, i.e. it was impossible to simply scrub while following and pause at a moment of your interest (the scrubber would immediately be forced back to the end). which is [now fixed ](https://github.com/rerun-io/reality/pull/1260). 
The issue that you're describing in here makes sense, and there are few more UX nuances we're going to address together while reviewing your PR. To give you an example: we have already a hotkey that goes to the end of the time line (`End` or `Fn →`). Which makes following mode kinda redundant, because as a user you simply play/pause and if you're in the very end of the timeline — then you are "following" but if you're somewhere in the middle of the timeline you simply play from where you are. and switching to "following mode" would simply mean — hit `play` and hit `go to the end of my timeline`. but once you hit "pause"  it would just pause and "play" would start from where you _paused_ without jumping to the end, unless you hit "end of the timeline". 
All of this to say we're fine tuning this behavior based on your input!

wolfd:
Hi @gavrelina, thanks for looking into this issue! What you're saying makes sense to me.

Do you want me to close out this PR?

IsseW:
We've decided to merge this along with some other changes to play mode, thanks @wolfd !

PR #12729, opened 2026-04-10
Feat/web viewer react follow if http

Woodii1998:
### Related

Closes #12728 

### What

Add `follow_if_http` support to `@rerun-io/web-viewer-react`.

This updates the React wrapper so the `follow_if_http` prop is passed through to the underlying web viewer when opening HTTP `.rrd` sources.

### DIFF
--- a/rerun_js/web-viewer-react/index.js
+++ b/rerun_js/web-viewer-react/index.js
@@ -75,19 +75,25 @@ export default class WebViewer extends React.Component {
 
       const prev = toArray(prevProps.rrd);
       const current = toArray(this.props.rrd);
-      if (prevProps.follow_if_http !== this.props.follow_if_http) {
-        this.#handle.close(prev);
-        this.#handle.open(current, {
-          follow_if_http: this.props.follow_if_http,
-        });
+      const { added, removed } = diff(prev, current);
+      const reopened =
+        prevProps.follow_if_http !== this.props.follow_if_http
+          ? intersection(prev, current).filter(isFollowIfHttpSource)
+          : [];
+
+      // While the viewer is still starting there is nothing to diff yet.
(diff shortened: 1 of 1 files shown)


Woodii1998:
@Wumpf 
I kept the fix in web-viewer-react because that let me address the bug with minimal surface area.

The startViewer change is working around a gap in web-viewer: today WebViewer.start(rrd, ...) performs the initial load internally, but that internal initial open() path cannot receive open-time options like follow_if_http. So the first recording load ignores follow_if_http, while later explicit open() calls do honor it.

That means the real root cause is arguably in web-viewer, not in the React wrapper. A cleaner architectural fix would be to change web-viewer so the initial load can also receive open options. But that felt like a larger API change than I wanted to make in this patch, since it would affect the base package interface, generated types, and docs.

So for now I scoped the fix to web-viewer-react: the wrapper explicitly controls the initial open flow and applies follow_if_http there, while also avoiding the startup-time race. If you think it makes sense, I’m happy to follow up with a separate web-viewer API change, but I wanted to leave that tradeoff to you rather than expanding the scope unilaterally.

jprochazk:
> A cleaner architectural fix would be to change web-viewer so the initial load can also receive open options. 

Could you open a separate PR with that, so we can compare? I'm not sure we want to accept something that we know is tech debt minute one

Woodii1998:
> > A cleaner architectural fix would be to change web-viewer so the initial load can also receive open options.
> 
> Could you open a separate PR with that, so we can compare? I'm not sure we want to accept something that we know is tech debt minute one

Ok PR is here https://github.com/rerun-io/rerun/pull/12731

jprochazk:
Closing in favor of:
- https://github.com/rerun-io/rerun/pull/12731

PR #12730, opened 2026-04-13
Complete support for PLY format (add mesh and 2D) also using more performant API of ply-rs-bw 4.0.0.

### LINKED ISSUE #12627: Implement PLY file reading using property accessor
PLY file reading is currently implemented in a simplified way (see [points3d_ext.rs](https://github.com/rerun-io/rerun/blob/main/crates/store/re_sdk_types/src/archetypes/points3d_ext.rs)). This could be optimized by using custom property accessors (you already have a TODO comment in the code referring to this). Such an improvement would  enhance performance when handling large and high-resolution PLY files, some of which can be very large in practice.

For your API, this would not introduce any breaking changes.

I would be happy to implement this optimization using property accessors (I am a maintainer of `ply-rs-bw` and know it). Although the change itself is not extensive, I would prefer to discuss the approach in advance before proceeding and submitting a PR.

bourumir-wyngs:
### Related

* Closes rerun-io/rerun#12627

### What

* Finish support for .PLY format, completing implementation also for mesh (not just point cloud)
* Support 2D PLY meshes, rendered flattened onto z=0, and 2D PLY point clouds as Points2D.
* Migrate to ply-rs-bw 4.0.0 that has API improvements allowing simpler and more efficient code in the client.
* Add tests, example and documentation (test pass and the example runs).
* Should not be any breaking changes, subject of possible bugs of course.

### Notes

* Initially scoped to property accessors, but expanded to cover missing PLY support (mesh and 2D).
* Found limitations in ply-rs-bw 3.x API; resolved properly in a new major release (4.0.0) instead of adding workarounds.
* Added an API compatibility check in ply-rs-bw CI (not in rerun) to enforce SemVer compliance.


crates/store/re_importer/src/lib.rs
```diff
@@ -543,9 +543,9 @@ pub const SUPPORTED_DEPTH_IMAGE_EXTENSIONS: &[&str] = &["rvl", "png"];
 
 pub const SUPPORTED_VIDEO_EXTENSIONS: &[&str] = &["mp4"];
 
+// .ply is both point cloud and mesh and should not be in this list for detection to work.
 pub const SUPPORTED_MESH_EXTENSIONS: &[&str] = &["glb", "gltf", "obj", "stl", "dae"];
 
-// TODO(#4532): `.ply` data loader should support 2D point cloud & meshes
 pub const SUPPORTED_POINT_CLOUD_EXTENSIONS: &[&str] = &["ply"];
```
Wumpf: [no action required] a bit odd, but given that so far our users use ply a lot for point clouds I think it makes sense

Wumpf (changes requested):
Sorry for letting this sit here for so long! This looks really good overall and has a lot of great tests and even an example <3, but I think there's a missunderstanding of Mesh3D <-> Asset3D going on here:
Instead of loading the mesh SDK sided (which we'd need to expose to C++, Python and Rust) we load encoded meshes into `Asset3D` and unpack those at runtime using one of the importers in `crates/viewer/re_renderer/src/importer/`

I.e. what that means is that we'd rather like to
* change documentation of `Asset3D` to point out PLY as supported (that's in `crates/store/re_sdk_types/definitions/rerun/archetypes/asset3d.fbs`, running codegen will mirror that over to all SDKs)
* have the dataloader branch to `Asset3D` if it finds any topology information
* wire up `crates/viewer/re_view_spatial/src/mesh_loader.rs` to a new `re_renderer::importer::PLY`

The reason we want to keep the raw mesh data as long as possible is that we generally like to preserve user input data as is, plus this delays the mesh parsing to the last moment meaning we don't have to "guard" all possible input paths (dataloader + 3x SDK)

there's ofc the future question whether `Asset3D` should just not render as point cloud if it sees a PLY without topology, but that's a design question that way beyond this iteration. Also, Asset3D is not very well inspectable in the Viewer today, but that's another problem entirely.

crates/store/re_importer/src/lib.rs
bourumir-wyngs: Understand, working on this. .ply is very versatile format that can hold 2D or 3D mesh, and if without vertices, this becomes the point cloud. It is internally a bit like "binary XML".

bourumir-wyngs:
Main changes since your review:
- topology-bearing PLY files are now preserved as `Asset3D` by the dataloader
- PLY is documented as supported by `Asset3D`
- `mesh_loader.rs` dispatches `MediaType::PLY` to the new renderer-side PLY importer
- PLY header/property classification was moved into a small shared `re_ply` helper crate
- docs now describe the `Asset3D` path instead of saying PLY meshes load as `Mesh3D`

bourumir-wyngs:
Could you please confirm the current status of this MR? I've implemented all requested changes. Before resolving the merge conflicts, I'd like to know whether the MR is still under consideration for merge.

Wumpf:
Sorry I've been first out on business travel and then for vacation and didn't hand this over to anyone else in the meantime! I can have a look at the conflicts and shepherd the rest in.

bourumir-wyngs:
Thanks for reply, would be great!

docs/content/getting-started/data-in/open-any-file.md
```diff
@@ -21,10 +21,21 @@ The following data types have built-in support in the Rerun Viewer and SDK:
...
+Once a `.ply` file is selected, its header is inspected to determine how to log it:
+
+-   if the header declares a non-empty `face` element, it is logged as an `Asset3D`
+-   otherwise, if the vertices contain `x`, `y`, and `z`, it is loaded as a `Points3D`
+-   otherwise, if the vertices contain `x` and `y`, it is loaded as a `Points2D`
+
+The Viewer renders PLY `Asset3D` data as a mesh when the face topology uses `vertex_indices` or `vertex_index`.
+For those topology-bearing PLY files, only `x` and `y` are required on vertices. If `z` is missing, the Viewer renders the mesh flattened onto the `z=0` plane.
+
+In practice, this means `.ply` is preserved as `Asset3D` whenever it declares topology, and rendered as a mesh when that topology is supported. Topology-free PLY files are loaded as point data.
```
Wumpf: that's a very long generated sounding section going into the details of `.ply` loading in a section that is supposed to be about file imports in general. should be reduced to just few subullets on the ply bullet point

crates/utils/re_ply/README.md
Wumpf, line 1: a whole new crate just for ply seems very heavy handed to me. I'm skeptical of whether this is actually needed, seems fine to inline most of these utilities? The non-test part is just 168 lines and most of it seem to be a bunch of standardized strings
Since you're the author of ply_rs, I wonder whether you'd be interested in instead up-streaming some of those constants?

crates/store/re_importer/src/importer_archetype.rs
```diff
@@ -150,12 +150,8 @@ impl Importer for ArchetypeImporter {
...
             )
         } else if crate::SUPPORTED_POINT_CLOUD_EXTENSIONS.contains(&extension.as_str()) {
-            re_log::debug!(?filepath, importer = self.name(), "Loading 3D point cloud…",);
-            self.send_chunks(
-                &tx,
-                &store_id,
-                load_point_cloud(timepoint, entity_path, &contents)?,
-            )
+            re_log::debug!(?filepath, importer = self.name(), "Loading .ply geometry…",);
+            self.send_chunks(&tx, &store_id, load_ply(timepoint, entity_path, &contents)?)
```
Wumpf: this doesn't make any sense: as is the code states that if an extension is contained in `SUPPORTED_POINT_CLOUD_EXTENSIONS` then it has to be a ply

crates/store/re_importer/src/lib.rs
```diff
@@ -540,9 +540,9 @@ pub const SUPPORTED_DEPTH_IMAGE_EXTENSIONS: &[&str] = &["rvl", "png"];
 
 pub const SUPPORTED_VIDEO_EXTENSIONS: &[&str] = &["mp4"];
 
+// .ply is both point cloud and mesh and should not be in this list for detection to work.
 pub const SUPPORTED_MESH_EXTENSIONS: &[&str] = &["glb", "gltf", "obj", "stl", "dae"];
 
-// TODO(#4532): `.ply` importer should support 2D point cloud & meshes
 pub const SUPPORTED_POINT_CLOUD_EXTENSIONS: &[&str] = &["ply"];
```
Wumpf: as per comment on importer_archetype: what's the point of saying it's a point cloud extension when it's not, it's a more general kind of extension that may yield to either.
So technically ply has to be indeed both on the mesh extensions and the point cloud extensions in case anyone probes from the outside what's supported (and they can since this is part of the public interface)
I'd do just that and then fix up the importer code

crates/store/re_importer/src/importer_archetype.rs
```diff
@@ -374,3 +389,447 @@ fn load_text_document(
...
+        assert!(
+            !chunk
+                .components()
+                .contains_component(Points2D::descriptor_positions().component)
+        );
+        assert!(
+            !chunk
+                .components()
+                .contains_component(Points3D::descriptor_positions().component)
+        );
```
Wumpf: seems redundant to the `ensure_similar`?

crates/store/re_importer/src/importer_archetype.rs
```diff
@@ -374,3 +389,447 @@ fn load_text_document(
...
+        assert!(
+            !chunk
+                .components()
+                .contains_component(Asset3D::descriptor_blob().component)
+        );
```
Wumpf: redundant

crates/store/re_importer/src/importer_archetype.rs
```diff
@@ -374,3 +389,447 @@ fn load_text_document(
...
+property uchar green
+property uchar blue
+end_header
+1 2 10 20 30
+4 5 11 21 31
+"#;
+
+        let chunk = load_single_chunk_via_archetype_importer("points_xy.ply", contents);
+
+        assert!(
```
Wumpf: this and following tests, why not keep using the the `ensure_similar` route?

crates/store/re_importer/src/importer_archetype.rs
```diff
@@ -374,3 +389,447 @@ fn load_text_document(
...
+            !chunk
+                .components()
+                .contains_component(Mesh3D::descriptor_vertex_positions().component)
+        );
+        assert!(
+            !chunk
+                .components()
+                .contains_component(Asset3D::descriptor_blob().component)
+        );
```
Wumpf: as above, seems redundant

crates/store/re_importer/src/importer_archetype.rs
```diff
@@ -374,3 +389,447 @@ fn load_text_document(
 
     Ok(rows.into_iter())
 }
+
+#[cfg(test)]
+mod tests {
```
Wumpf: let's move this to a separate test file under /tests/ - it's quite odd that `importer_archetype` would handle all sort of things and then have tests exactly only for ply assets.

apart from some redundancies very nice test suite!

crates/store/re_sdk_types/src/archetypes/points3d_ext.rs
```diff
@@ -119,181 +12,573 @@ const PROP_SH_DC_1: &str = "f_dc_1";
...
+                SEEN_OPACITY,
+                &mut self.opacity,
+            ),
+            _ => PropertyAccessResult::Ignored,
         }
     }
 }
 
-fn read_ply(reader: &mut impl std::io::BufRead) -> std::io::Result<Points3D> {
+impl Points2D {
```
Wumpf: why is this not in points2d_ext.rs

crates/store/re_sdk_types/src/archetypes/points3d_ext.rs
```diff
@@ -119,181 +12,573 @@ const PROP_SH_DC_1: &str = "f_dc_1";
...
     scale_z: Option<f32>,
     opacity: Option<f32>,
+    seen_props: u16,
 }
 
-impl PropertyAccess for Vertex {
-    fn new() -> Self {
-        Self::default()
+impl ParsedVertex {
+    fn note_ignored_props_for(seen_props: u16, ignored_props: &mut BTreeSet<String>, mask: u16) {
```
Wumpf: seems like we have to run this _a lot_ per vertex, doesn't that cause a significant perf hit?
also why are we stringifying them so eagerly when we have them as bitfields?

crates/store/re_sdk_types/tests/types/points2d.rs
```diff
@@ -62,3 +64,129 @@ fn roundtrip() {
     let deserialized = Points2D::from_arrow(serialized).unwrap();
     similar_asserts::assert_eq!(expected, deserialized);
 }
+
+fn example_ply_path() -> PathBuf {
+    Path::new(env!("CARGO_MANIFEST_DIR")).join("../../../examples/assets/example.ply")
+}
+
```
Wumpf, line 68: misses `mod tests`

I don't quite get how you decided which tests to put in here versus in the higher level re_importer. Arguably what re_importer adds is fairly trivial integration and almost all the tests should be in here rather?

crates/viewer/re_renderer/src/importer/ply.rs
```diff
@@ -0,0 +1,721 @@
...
+    ny: Option<f32>,
+    nz: Option<f32>,
+    red: Option<u8>,
+    green: Option<u8>,
+    blue: Option<u8>,
+    alpha: Option<u8>,
+}
+
+impl ParsedMeshVertex {
+    fn into_parts(self) -> std::io::Result<(glam::Vec3, Option<glam::Vec3>, Option<Rgba32Unmul>)> {
```
Wumpf: labling of those parts via a struct would be good - have to know from context that it's position & normal. but also why not just do getter individually for position/normal/color?

examples/rust/ply_stl_tetrahedrons/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+fn main() -> anyhow::Result<()> {
+    let main_thread_token = rerun::MainThreadToken::i_promise_i_am_on_the_main_thread();
+    re_log::setup_logging();
+
+    let (rec, storage) =
+        rerun::RecordingStreamBuilder::new("rerun_example_ply_stl_tetrahedrons").memory()?;
+    run(&rec)?;
+
+    rerun::native_viewer::show(main_thread_token, storage.take())?;
+    Ok(())
```
Wumpf: this is very unidiomatic rerun usage: yes it makes the sample self-contained and doesn't need an external rerun viewer, but this means that the sample _is_ the rerun viewer and has to compile the whole thing in. The normal Rerun execution model is that applications log to file/viewer/server and then consume data there.

Let's follow that here and use `re_log::setup_logging();`, `rerun::clap::RerunArgs` etc. like most other rust examples do

examples/rust/ply_stl_tetrahedrons/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+name = "ply_stl_tetrahedrons"
+version = "0.35.0-alpha.1+dev"
+edition = "2024"
+rust-version = "1.95"
+license = "MIT OR Apache-2.0"
+publish = false
+
+[dependencies]
+rerun = { path = "../../../crates/top/rerun", default-features = false, features = [
+  "native_viewer",
```
Wumpf: as above, we only use `native_viewer` when demonstrating viewer extensions etc.

examples/rust/ply_stl_tetrahedrons/README.md
Wumpf, line 1: nice!

Wumpf (changes requested):
I did a mostly llm driven rebase & fixup and then went onto reviewing the latest state. The handling of asset3d in that whole machinery seems fine to me now. And most other things too! However, there's quite a few things that I found a bit too messy as well as some questions. You're still up for having a look at these? 🙏

docs/content/getting-started/data-in/open-any-file.md
bourumir-wyngs: shortened

crates/store/re_importer/src/importer_archetype.rs
bourumir-wyngs: removed

crates/store/re_importer/src/importer_archetype.rs
bourumir-wyngs: The function was renamed to load_ply because it likely was not capable of loading any other kind of point cloud anyway. I currently see some more code there in the diff, maybe this came from some later contribution. Here something likely went wrong, trying to understand.

bourumir-wyngs: Worst is that this **re_importer seems missing completely in the current main branch.** I need to untagle where all the code needs to go.

Wumpf: eh sorry sounds like I might have messed up some stuff there as well

Wumpf:
ah I figure it was long time for another round on my end and to help cleaning up the mess I'm partially responsible. Re-request review would have been helpful :)

Very short on time this week but setttng a reminder for next

crates/build/re_type_definitions/rerun/archetypes/asset3d.def.rs
```diff
@@ -27,6 +27,7 @@ pub struct Asset3D {
     /// * `model/gltf-binary`
     /// * `model/gltf+json`
     /// * `model/obj` (.mtl material files are not supported yet, references are silently ignored)
+    /// * `application/x-ply`
     /// * `model/stl`
```
emilk, line 31: sort

crates/store/re_importer/src/lib.rs
```diff
@@ -589,10 +590,17 @@ pub const SUPPORTED_DEPTH_IMAGE_EXTENSIONS: &[&str] = &["rvl", "png"];
 
 pub const SUPPORTED_VIDEO_EXTENSIONS: &[&str] = &["mp4"];
 
-pub const SUPPORTED_MESH_EXTENSIONS: &[&str] = &["glb", "gltf", "obj", "stl", "dae"];
+/// The `.ply` extension, which is in both [`SUPPORTED_MESH_EXTENSIONS`] and
+/// [`SUPPORTED_POINT_CLOUD_EXTENSIONS`].
+pub const PLY_EXTENSION: &str = "ply";
```
emilk: lets just inline this for consistency

crates/store/re_sdk_types/src/components/media_type_ext.rs
```diff
@@ -233,6 +233,10 @@ impl MediaType {
             Some("stl") => {
                 return Some(Self::stl());
             }
+            // `mime_guess2` doesn't know about PLY, but we use `application/x-ply`.
+            Some("ply") => {
```
emilk, line 233: nit: sort

crates/store/re_sdk_types/tests/types/points2d.rs
```diff
@@ -62,3 +64,131 @@ fn roundtrip() {
     let deserialized = Points2D::from_arrow(serialized).unwrap();
     similar_asserts::assert_eq!(expected, deserialized);
 }
+
+fn example_ply_path() -> PathBuf {
+    Path::new(env!("CARGO_MANIFEST_DIR")).join("../../../examples/assets/example.ply")
```
emilk: this would be more robust as include_str!

crates/store/re_sdk_types/tests/types/points3d.rs
```diff
@@ -59,3 +61,249 @@ fn roundtrip() {
     let deserialized = Points3D::from_arrow(serialized).unwrap();
     similar_asserts::assert_eq!(expected, deserialized);
 }
+
+fn example_ply_path() -> PathBuf {
+    Path::new(env!("CARGO_MANIFEST_DIR")).join("../../../examples/assets/example.ply")
```
emilk: include_str here and elswhere

crates/store/re_sdk_types/src/archetypes/points2d_ext.rs
```diff
@@ -0,0 +1,183 @@
...
+        let radii = radii
+            .into_iter()
+            .map(|opt| opt.unwrap_or_else(|| Radius::from(1.0)));
+        arch = arch.with_radii(radii);
+    }
+    if labels.iter().any(|opt| opt.is_some()) {
+        // If some labels have been specified but not others, default the unspecified ones to "undef".
+        let labels = labels
+            .into_iter()
+            .map(|opt| opt.unwrap_or(Text("undef".into())));
```
emilk: use an empty string , here and in 3D

crates/store/re_sdk_types/src/archetypes/points3d_ext.rs
```diff
@@ -1,308 +1,220 @@
...
+            ply::PROP_Z => ply::set_required_f32(&property, &mut self.z),
+            ply::PROP_RED => ply::set_color(&property, &mut self.red),
+            ply::PROP_GREEN => ply::set_color(&property, &mut self.green),
+            ply::PROP_BLUE => ply::set_color(&property, &mut self.blue),
+            ply::PROP_ALPHA => ply::set_color(&property, &mut self.alpha),
+            ply::PROP_RADIUS => ply::set_f32(&property, &mut self.radius),
+            ply::PROP_LABEL => ply::set_text(&property, &mut self.label),
+            ply::PROP_SH_DC_0 => ply::set_f32(&property, &mut self.sh_dc_0),
+            ply::PROP_SH_DC_1 => ply::set_f32(&property, &mut self.sh_dc_1),
+            ply::PROP_SH_DC_2 => ply::set_f32(&property, &mut self.sh_dc_2),
```
emilk: I think we should remove all this gaussian specific stuff from point3d

crates/store/re_sdk_types/src/archetypes/points3d_ext.rs
```diff
@@ -1,308 +1,220 @@
...
+            ply::PROP_OPACITY => ply::set_f32(&property, &mut self.opacity),
+            ply::PROP_SCALE_X => ply::set_f32(&property, &mut self.scale_x),
+            ply::PROP_SCALE_Y => ply::set_f32(&property, &mut self.scale_y),
+            ply::PROP_SCALE_Z => ply::set_f32(&property, &mut self.scale_z),
```
emilk: these too

crates/store/re_sdk_types/src/ply.rs
```diff
@@ -0,0 +1,472 @@
...
+    if has_readable_faces(header) {
+        return Ok(PlyGeometryClass::MeshOrAsset3D);
+    }
+
+    match classify_vertex_layout(vertex_element) {
+        PlyVertexLayout::Xy => Ok(PlyGeometryClass::Points2D),
+        PlyVertexLayout::Xyz => Ok(PlyGeometryClass::Points3D),
+        PlyVertexLayout::Other => Err(std::io::Error::new(
+            std::io::ErrorKind::InvalidData,
+            "PLY vertex element must contain at least \"x\" and \"y\"",
```
emilk: can we include the name of all properties that _do_ exist in the error message?

emilk (approved):

emilk:
@bourumir-wyngs I took the freedom to rearrange this PR quite a bit. Please take a look and see if it still makes sense to you, and then we can merge it!

emilk:
@rerun-bot reality-sync

emilk:
@rerun-bot reality-sync

(merged latest `main` into the branch)

— Claude

emilk:
@rerun-bot reality-sync

Squashed the branch onto latest `main` (identical tree, 1 commit) - the sync bot replays PR commits one-by-one, and the oldest ones no longer applied.

— Claude

crates/viewer_support/re_renderer/src/importer/ply.rs
```diff
@@ -0,0 +1,810 @@
...
+    let mut ignored_props = BTreeSet::new();
+
+    let mut saw_faces = false;
+
+    for (_key, element_def) in &header.elements {
+        match element_def.name.as_str() {
+            ELEMENT_VERTEX => {
+                let vertices = vertex_parser
+                    .read_payload_for_element(&mut payload_reader, element_def, &header)
+                    .map_err(std::io::Error::from)?;
```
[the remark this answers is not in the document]
emilk: No change: `ply-rs-bw` 4.x already caps this. `read_payload_for_element` sizes the `Vec` through `cap_preallocated_size`, which clamps to `MAX_PREALLOCATED_SIZE = 65_536` elements, so a header claiming a huge `element vertex` count cannot drive the allocation. Past that cap the `Vec` only grows as real elements are parsed, and a truncated payload stops at `UnexpectedEof`.

The batching in `gaussian_splats3d_ext.rs` earns its keep for the other reason its comment gives: it frees each parsed batch instead of holding every splat alive alongside the archetype being built. A mesh needs every vertex and face at once anyway, so batching would not lower peak memory here.

— Claude

crates/store/re_sdk_types/src/archetypes/points3d_ext.rs
```diff
@@ -1,9 +1,104 @@
...
 use super::Points3D;
 use crate::components::{Color, Position3D, Radius, Text};
+use crate::ply;
+
+/// The `vertex` properties a `Points3D` is read out of.
+///
+/// Anything else in the header is reported as ignored, once. That includes the gaussian
+/// splatting parameters: a reconstruction is read by
+/// [`crate::archetypes::GaussianSplats3D`] instead.
+const SUPPORTED_PROPERTIES: [&str; 9] = [
```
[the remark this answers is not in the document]
emilk: No change: this is deliberate. A splat `.ply` is now read by `GaussianSplats3D`, which keeps the full reconstruction (centers, rotations, scales, opacity and spherical harmonics) instead of the lossy color/radius approximation `Points3D` used to derive. `classify_geometry_from_bytes` routes splat headers there before the point-cloud rules see them, so opening such a file gives the user more than it did before, not less.

— Claude

crates/viewer_support/re_renderer/src/importer/ply.rs
```diff
@@ -0,0 +1,810 @@
...
+            }
+            _ => PropertyAccessResult::Ignored,
+        }
+    }
+}
+
+fn triangulate_face(indices: &[u32]) -> impl Iterator<Item = glam::UVec3> + '_ {
+    indices[1..]
+        .windows(2)
+        .map(|pair| glam::uvec3(indices[0], pair[0], pair[1]))
```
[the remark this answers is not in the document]
emilk: Documented rather than fixed. `triangulate_face` now spells out that the fan is exact for triangles and quads, which is what `.ply` faces are in practice, and that a concave polygon of five or more vertices can come out wrong. Rejecting non-triangular faces would throw away valid convex n-gons, so the fan stays until a real file asks for ear clipping.

— Claude

emilk:
@rerun-bot reality-sync

emilk:
@rerun-bot reality-sync

Rebased on latest `main`, which moved `re_importer` to `crates/data_flow/` and `re_renderer` to `crates/viewer_support/`.

— Claude

emilk:
@rerun-bot reality-sync

emilk:
@rerun-bot reality-sync

emilk:
@rerun-bot reality-sync

emilk:
@rerun-bot reality-sync

bourumir-wyngs:
Thanks for finishing this. I am sorry I was too busy to contribute. Well done.

PR #12737, opened 2026-04-22
Unpin wasm-bindgen

anassinator:
The comment specifically specifies to never pin this dependency, but we do anyways. This prevents us from upgrading.

[>  Do not make this an `=` dependency, because that may break Rust users’ builds when a newer
](https://github.com/rerun-io/rerun/blob/main/Cargo.toml#L428)

This was done correctly prior to https://github.com/rerun-io/rerun/commit/af1f737838afe980ae98dd561a3cf4fffec50bd8 which pinned it without clear justification.


SeaOtocinclus:
@lucasmerlin @jleibs @Wumpf Any feedback on this?

emilk (approved):
Ah good catch. Yeah, Cargo.lock should handle this without the `=`

emilk:
@rerun-bot reality-sync

PR #12742, opened 2026-04-29
feat: support vp8 on the web

AyushAgrawal-A2:
### Related

Closes #8114 

### What

- [x] Update codec string to 'vp8'
- [x] Update docs
- [x] Add test

PR #12743, opened 2026-04-29
feat: VP8+VP9 support for VideoStream archetype

### LINKED ISSUE #8114: Support VP8 on the web
This should be very easy to do! All we have to do is to make sure we send the right codec string in - right now we pass in too much when it should actually just be "vp8". See

https://www.w3.org/TR/webcodecs-vp8-codec-registration/#fully-qualified-codec-strings

Solving this issue includes
* adding a vp8 video to the video release checklist
* update the table in the docs
    * remove incorrect comment that this isn't supported in Mp4! It is an officially recognized codec, see https://mp4ra.org/registered-types/codecs
* test this in a handful of browsers

--- 
Internal Slack thread on this https://rerunio.slack.com/archives/C07Q6HLG24S/p1731342080159589 (all the relevant information as of writing is paraphrased above)
Thanks to @jprochazk for digging up these things links

### LINKED ISSUE #10186: VP9 support for `VideoStream` archetype

AyushAgrawal-A2:
depends on https://github.com/rerun-io/re_mp4/pull/22

### Related
Closes #8114
Closes #10186

### What
- [x] Add IVF stream for VP8 + VP9 VideoStream archetype
- [x] Update all glue code, types and buffers for VP8, VP9 type definitions and plumbing
- [x] Add test and snapshots for VP8 + VP9 VideoStream


crates/store/re_lenses/src/op/semantic.rs
```diff
@@ -524,6 +526,10 @@ mod tests {
             Some("h264"),
             Some("H265"),
             Some("aV1"),
+            Some("vp8"),
+            Some("vp9"),
+            Some("Vp8"),
+            Some("vP9"),
```
Wumpf, line 532: what are the different variations for here?

crates/store/re_lenses/src/op/semantic.rs
```diff
@@ -545,7 +555,7 @@ mod tests {
     /// Tests that we return the correct error when an unsupported codec is in the data.
     #[test]
     fn test_string_to_codec_uint32_unsupported() {
-        let unsupported_codecs = ["vp9"];
+        let unsupported_codecs = ["vp08", "vp09"];
```
Wumpf, line 558: seems a bit odd - these are just codec strings we're not parsing correctly but _look_ valid 🤔 
We actually might as well just support them (cc: @grtlr ). Better to put in some properly unsupported codec in that list

docs/content/concepts/logging-and-ingestion/video.md
```diff
@@ -94,16 +92,11 @@ Codec support varies in the web & native viewer:
...
-<!--
-for web codecs see https://www.w3.org/TR/webcodecs-codec-registry/#video-codec-registry
-VP8 is only not in the list because VP9 doesn't support MP4 as a container and that's
-today the only container we take.
--->
+| AV1        | ✅      | 🟧     |
+| H.264/avc  | ✅      | ✅     |
+| H.265/hevc | 🟧      | ✅     |
+| VP8        | ✅      | ✅     |
+| VP9        | ✅      | ✅     |
```
Wumpf, line 99: 🥳

crates/utils/re_video/src/decode/ffmpeg_cli/ffmpeg.rs
```diff
@@ -1079,24 +1057,149 @@ fn sanitize_ffmpeg_log_message(msg: &str) -> String {
     msg
 }
 
-#[derive(Clone)]
 enum CodecMeta {
```
Wumpf, line 1060: nice, I really like how you abstracted those writes out!

crates/utils/re_video/src/decode/ffmpeg_cli/ivf.rs
Wumpf, line 1: haven't looked up any of the details of IVF, but this looks reasonable to me and apparently it works 👍

crates/utils/re_video/src/demux/mp4.rs
```diff
@@ -74,6 +90,28 @@ impl VideoDataDescription {
             }
         }
 
+        // VP8/VP9 MP4 files often have incorrect stss (sync sample) boxes,
+        // where the muxer marks non-keyframes as sync samples. Override is_sync
+        // by inspecting the actual bitstream, and rebuild keyframe_indices.
```
Wumpf, line 95: any source can cite for this? Could this be instead a bug in `re_mp4`?

crates/utils/re_video/src/demux/mp4.rs
```diff
@@ -227,11 +250,17 @@ fn codec_details_from_stds(
         _ => {}
     }
 
-    Ok(VideoEncodingDetails {
-        codec_string: stsd
+    let codec_string = match &stsd.contents {
+        // https://www.w3.org/TR/webcodecs-vp8-codec-registration/#fully-qualified-codec-strings
+        re_mp4::StsdBoxContent::Vp08(_) => "vp8".to_owned(),
+        _ => stsd
             .contents
```
Wumpf: as noted on the other PR, we should upstream this to `re_mp4`

crates/utils/re_video/src/vp8.rs
Wumpf, line 1: [no action] thanks for providing all the source here and in vp9 which should make it easy to debug this if something is wrong. Again I checked mostly for code sanity rather than following on all the source material trusting that the correctly augmented tests in this PR capture any glaring problems

Wumpf (changes requested):
excellent, thanks for filling in this gap!
Looks overall really well done! Gave it also a spin locally and seemed to work just fine.

Thanks for trying to do this step by step and basing this on a first iteration in https://github.com/rerun-io/rerun/pull/12742 but GitHub of course handles that poorly and we have both changes combined here. It did make reviewing a bit easier, but I think at this point we might as well land this as a combined PR and close the other one.

Repeating the still relevant note from #12742

---
Tried it locally on Mac in
* ✅ Firefox
* ✅ Chrome
* ❌ Safari

I know we don't have that for VP9 there yet either, but we might as well document the finding in the `### Web viewer` further below in the Readme
I can arrange more testing at a later point - but if you're on a different system you can put your findings there as well right away :)

Would you mind fixing `.codec_string()` https://github.com/rerun-io/re_mp4/ so we don't have this extra branch here? I can do a new release of re_mp4 right away so we get this in in a timely manner

---

With that and a the open questions I left I think we're good to go here!

crates/store/re_lenses/src/op/semantic.rs
AyushAgrawal-A2: Different casing, 531 has capital v , 532 has capital p
This is similar to different letter casing used just above these ones.

crates/store/re_lenses/src/op/semantic.rs
AyushAgrawal-A2: correct codec strings here would be `vp8` and not `vp08`, same for vp9
this test just check correct error is return when codec string passed in is not supported/incorrect

crates/utils/re_video/src/demux/mp4.rs
AyushAgrawal-A2: done in the base pr and changes are pulled in

AyushAgrawal-A2:
@Wumpf 

> awesome, didn't realize it's _that_ trivial to get it to work. Tried it locally on Mac in
> 
> * ✅ Firefox
> * ✅ Chrome
> * ❌ Safari
> 
> I know we don't have that for VP9 there yet either, but we might as well document the finding in the `### Web viewer` further below in the Readme I can arrange more testing at a later point - but if you're on a different system you can put your findings there as well right away :)

I am on Linux (Ubuntu 25.10), I tested and this works in Chrome and Firefox. I have added result for Linux and Mac (your results) in Readme for both vp8 and vp9. I have left columns for windows as empty.


> Would you mind fixing `.codec_string()` https://github.com/rerun-io/re_mp4/ so we don't have this extra branch here? I can do a new release of re_mp4 right away so we get this in in a timely manner

Done. PR for upstream changes in re_mp4 https://github.com/rerun-io/re_mp4/pull/22 , I have removed the codec_string branch for vp8 from this PR.

crates/utils/re_video/src/demux/mp4.rs
AyushAgrawal-A2: This happened for the vp8 sample video I added in this PR. stss was incorrect and video seek was erroring out in player. But the video works in web, just not native. Which led me to https://codereview.chromium.org/1319813002

Chrome ignores unreliable stss data and derives it by parsing the actual bitstream at demux time, which is being done here as well.

crates/store/re_lenses/src/op/semantic.rs
Wumpf: fair, let's leave it then

Wumpf (approved):
I've landed your re_mp4 changes and published a new version with it. Will have a run through CI & a few more quick local checks and I think then we're good to go. Thanks again for looking into this!

Wumpf:
@rerun-bot reality-sync

Wumpf:
did conflict resolution and a small ci test failure fix on the monorepo side which is why the sync is sad now. But it landed just fine https://github.com/rerun-io/rerun/commit/a1642b4f8213ac2c00d51bb6a630c30705d84a3b \o/

Thanks again @AyushAgrawal-A2 for taking care of this, great to have this gap closed! :)

PR #12749, opened 2026-05-01
Patch auto-downloaded arrow for C++ for latest macos libtool

### LINKED ISSUE #12748: c++: arrow + latest macos issue
I updated macOS and/or Xcode (not sure which one is at fault) and now I cannot use rerun with CMake's FetchContent.

The issue is that arrow tries to identify which `libtool` is in the PATH because they don't support gcc's one (or something like that). But the new `libtool` returns the version in a different format and their (old) regexp does not match anymore, so they error out.

The issue is fixed in arrow https://github.com/apache/arrow/pull/49370 

I think it's available in arrow version 24. Is it possible to update to the latest arrow (probably too early, it was just released), or can this patch be applied as part of the FetchContent-based setup? 

```diff
diff --git a/cpp/cmake_modules/BuildUtils.cmake b/cpp/cmake_modules/BuildUtils.cmake
index 692efa7..4f8f5e7 100644
--- a/cpp/cmake_modules/BuildUtils.cmake
+++ b/cpp/cmake_modules/BuildUtils.cmake
@@ -112,7 +112,7 @@ function(arrow_create_merged_static_lib output_target)
     execute_process(COMMAND ${LIBTOOL_MACOS} -V
                     OUTPUT_VARIABLE LIBTOOL_V_OUTPUT
                     OUTPUT_STRIP_TRAILING_WHITESPACE)
-    if(NOT "${LIBTOOL_V_OUTPUT}" MATCHES ".*cctools-([0-9.]+).*")
+    if(NOT "${LIBTOOL_V_OUTPUT}" MATCHES ".*cctools.+([0-9.]+).*")
       message(FATAL_ERROR "libtool found appears to be the incompatible GNU libtool: ${LIBTOOL_MACOS}"
       )
     endif()
```

nikolaseu:
### Related

Closes #12748 

### What

Latest macOS's libtool returns the version with a slightly different format so the regex used by arrow fails to recognize it, and aborts as if GNU's libtool is being used (which is not supported).

See https://github.com/apache/arrow/pull/49370

### DIFF
--- a/rerun_cpp/download_and_build_arrow.cmake
+++ b/rerun_cpp/download_and_build_arrow.cmake
@@ -81,7 +81,11 @@ function(download_and_build_arrow)
         set(VERSION_PATCH "-DCMAKE_POLICY_VERSION_MINIMUM=${CMAKE_POLICY_VERSION_MINIMUM}")
     endif()
 
-    set(MIMALLOC_PATCH ${CMAKE_CURRENT_LIST_DIR}/patches/mimalloc_cmake4.patch)
+    set(ARROW_PATCHES
+        # TODO(apache/arrow#45985): Arrow can't support CMake 4.0 yet
+        ${CMAKE_CURRENT_LIST_DIR}/patches/mimalloc_cmake4.patch
+        # TODO update to arrow 24 and remove this patch
+        ${CMAKE_CURRENT_LIST_DIR}/patches/arrow_libtool.patch)
 
     ExternalProject_Add(
         arrow_cpp
@@ -92,8 +96,8 @@ function(download_and_build_arrow)
         UPDATE_COMMAND "" # Prevent unnecessary rebuilds on every cmake --build
 
         # Apply patch after checkout but before configure
-        # TODO(apache/arrow#45985): Arrow can't support CMake 4.0 yet
-        PATCH_COMMAND git apply --check ${MIMALLOC_PATCH} && git apply ${MIMALLOC_PATCH} || true
+        # Since this is downloading a release tarball, we need to init a git repository in order to apply the patches.
+        PATCH_COMMAND git init && git apply --check ${ARROW_PATCHES} && git apply ${ARROW_PATCHES} || true
 
         # LOG_X ON means that the output of the command will
         # be logged to a file _instead_ of printed to the console.
--- a/rerun_cpp/patches/arrow_libtool.patch
+++ b/rerun_cpp/patches/arrow_libtool.patch
@@ -1,0 +1,13 @@
+diff --git a/cpp/cmake_modules/BuildUtils.cmake b/cpp/cmake_modules/BuildUtils.cmake
+index 692efa7..4f8f5e7 100644
+--- a/cpp/cmake_modules/BuildUtils.cmake
++++ b/cpp/cmake_modules/BuildUtils.cmake
+@@ -112,7 +112,7 @@ function(arrow_create_merged_static_lib output_target)
+     execute_process(COMMAND ${LIBTOOL_MACOS} -V
+                     OUTPUT_VARIABLE LIBTOOL_V_OUTPUT
+                     OUTPUT_STRIP_TRAILING_WHITESPACE)
+-    if(NOT "${LIBTOOL_V_OUTPUT}" MATCHES ".*cctools-([0-9.]+).*")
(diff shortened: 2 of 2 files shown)


rerun_cpp/download_and_build_arrow.cmake
```diff
@@ -93,7 +95,7 @@ function(download_and_build_arrow)
 
         # Apply patch after checkout but before configure
         # TODO(apache/arrow#45985): Arrow can't support CMake 4.0 yet
```
Wumpf: docs need to move to the actual patch and should also document the new patch (why it's needed, what version would make it go away etc.)

Wumpf (changes requested):
Thanks for looking into this and upstreaming your solution. See no reason not to take that in, other than needing to document it a bit better so we remove the patch!

That said, the version we pinned on arrow here is getting quite old, we should soon look into picking up a newer one. Afaik they switched to a newer C++ version and I'd like us to be C++17 compatible for a bit longer, so options might be a bit limited. (either way that's for a different iteration ofc)

nikolaseu:
Hi Andreas, thanks for taking a look.

I will check and continue with this again later. I marked it as draft because I can't really get it working locally, it seems like the patch was not being applied, but it didn't fail either. I'm not sure what was going on.

rerun-rmack:
I suspect copying lines 94-146 from latest BuildUtils.cmake to replace lines 100-121 would be an effective fix but I don't have a machine to test that on.

nikolaseu:
> I suspect copying lines 94-146 from latest BuildUtils.cmake to replace lines 100-121 would be an effective fix but I don't have a machine to test that on.

Sorry, I was not clear. The patch works when applied manually, but something is maybe wrong in the setup or my changes in `rerun_cpp/download_and_build_arrow.cmake`

There's no need to patch a lot more than this in the currently used version of arrow, even if latest arrow changed a lot more.

nikolaseu:
@Wumpf the issue was that the patches were not being applied, they were just skipped by git. The previous setup was not working, so I guess the mimalloc issue was not an issue anymore.

rerun_cpp/download_and_build_arrow.cmake
```diff
@@ -81,7 +81,11 @@ function(download_and_build_arrow)
         set(VERSION_PATCH "-DCMAKE_POLICY_VERSION_MINIMUM=${CMAKE_POLICY_VERSION_MINIMUM}")
     endif()
 
-    set(MIMALLOC_PATCH ${CMAKE_CURRENT_LIST_DIR}/patches/mimalloc_cmake4.patch)
+    set(ARROW_PATCHES
+        # TODO(apache/arrow#45985): Arrow can't support CMake 4.0 yet
+        ${CMAKE_CURRENT_LIST_DIR}/patches/mimalloc_cmake4.patch
+        # TODO update to arrow 24 and remove this patch
```
Wumpf, line 87: our CI flags `TODO` notes without ticket or name references.

Also, do you know whether this is fixed in earlier arrow versions as well? I checked and it looks like arrow 23 starts requiring C++20, so anything prior to that would be easier to upgrade for us.

rerun_cpp/download_and_build_arrow.cmake
```diff
@@ -92,8 +96,8 @@ function(download_and_build_arrow)
         UPDATE_COMMAND "" # Prevent unnecessary rebuilds on every cmake --build
 
         # Apply patch after checkout but before configure
-        # TODO(apache/arrow#45985): Arrow can't support CMake 4.0 yet
-        PATCH_COMMAND git apply --check ${MIMALLOC_PATCH} && git apply ${MIMALLOC_PATCH} || true
+        # Since this is downloading a release tarball, we need to init a git repository in order to apply the patches.
+        PATCH_COMMAND git init && git apply --check ${ARROW_PATCHES} && git apply ${ARROW_PATCHES} || true
```
Wumpf, line 100: afaik that's not entirely true, but the problem is rather that we're more likely than not already in an existing git repo

Digging a bit: at the time of the introducing PR #11223 it still did a full git clone, but at the same time I can build with cmake4 just fine off main for me :/. probably still best to leave the patch for now, cc @ntjohnson1

Wumpf (changes requested):

nikolaseu:
Dropping this, I'm using the system arrow with `RERUN_DOWNLOAD_AND_BUILD_ARROW=OFF` without issues.

PR #12753, opened 2026-05-08
Show bound and connect URLs when serving web viewer

### LINKED ISSUE #11584: Serving web viewer should print both server and connect IPs
Right now, if you do `rerun --serve-web --bind 0.0.0.0`, the web viewer server will print:

```
Hosting a web-viewer at http://localhost:9090?url=rerun%2Bhttp%3A%2F%2Flocalhost%3A9876%2Fproxy
```

This is not true, that's just the _connect_ URL, the one you should use to view it on the same machine. In the background, it is actually on `0.0.0.0`.

It should instead print:

```
Hosting a web-viewer at http://0.0.0.0:9090 - connect at http://localhost:9090?url=rerun%2Bhttp%3A%2F%2Flocalhost%3A9876%2Fproxy
```

terror:
Resolves #11584

When serving the web viewer on an unspecified bind address such as `0.0.0.0`, the log output previously only showed the local connect URL. This made it look like the server was hosted on `127.0.0.1` even though it was actually listening on `0.0.0.0`.

This adds a separate bound URL for `WebViewerServer` and uses it in the web viewer hosting log message, while keeping the browser/connect URL unchanged. The output now shows both where the server is listening and which URL should be opened locally.


oxkitsune (approved):
Awesome quality of life improvement! Thanks!

oxkitsune:
@rerun-bot sync-reality

ntjohnson1:
@rerun-bot sync-reality

PR #12754, opened 2026-05-08
Add iteration and length support to `RRDArchive`

### LINKED ISSUE #11206: Add __iter__ and __len__ to RRDArchive
With the OSS Server we probably should drop the dataframe api.

However, if we don't it would be nice to make RRDArchive a little more pythonic.

```python
for recording in archive.all_recordings():
   # Do something
```

To

```python
for recording in archive:
   # Do something
```

terror:
Resolves #11206

This diff makes `rerun.recording.RRDArchive` more Pythonic by adding `__len__` and `__iter__`. Archives can now be used directly with `len(archive)` and for recording in archive, matching the existing `num_recordings` and `all_recordings` behavior without changing the underlying loading logic.

### DIFF
--- a/rerun_py/rerun_sdk/rerun/recording/_recording.py
+++ b/rerun_py/rerun_sdk/rerun/recording/_recording.py
@@ -104,6 +104,12 @@ class RRDArchive:
     def __init__(self, inner: RRDArchiveInternal) -> None:
         self._internal = inner
 
+    def __len__(self) -> int:
+        return self.num_recordings()
+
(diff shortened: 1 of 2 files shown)


oxkitsune:
Apologies for getting to this so late please tag us on discord if we take this long to respond in the future! 

This is a good contribution! However, we have since deprecated this api in favour of the new `rerun.experimental.RrdReader` api, which contains `.recordings()` which returns a plain list (so `len(reader.recordings())` and iteration already work).

PR #12759, opened 2026-05-11
Rename `Compaction` label to `Compaction config`

### LINKED ISSUE #10265: The compaction field of store info shows the same value for all recordings
Reproduce:
1. Go to https://rerun.io/viewer
2. Open any examples
3. Click on the recording
4. look at the "Compaction" field in the Data section in the selection panel
5. Do the same for any other recording
6. Notice it always says "4 096 rows (1 024 if unsorted) or 384 KiB"

Image

Image

Image

lsy3:
### Related

* Closes #10265

### What

The `Compaction` row in the recording-properties Selection panel sits in the "Data" section alongside per-recording measurements (Rows, Size, Entities, etc.), but the value is actually the `ChunkStoreConfig` that governs in-memory compaction — defaulted from env vars but stored per-`EntityDb`. Without "config" in the label, a reader naturally categorizes the row as another measurement.

@Wumpf proposed this rename in the issue thread; @teh-cmc flagged that a longer label had previously offset the table. At 17 characters `Compaction config` is one less than the existing `Source RRD version` (18) in the same grid, so no new column offset.

Hover tooltip is untouched — @emilk added it in [`3d19ca7`](https://github.com/rerun-io/rerun/commit/3d19ca7bf63), already addressing the "expanded explanation on hover" that @nikolausWest suggested in the thread.

#### After

Screenshot from 2026-05-11 13-11-47

---

*First contribution to rerun — happy to iterate on style or scope if there are conventions I missed.*


lsy3:
Heads up — CI's `label` job is failing because outside contributors can't apply the required labels (silent permission denial — `gh pr edit --add-label` succeeds with no error but no label is applied).

Per `labels.yml` there are two label requirements:

1. **Changelog**: one of `include in changelog` / `exclude from changelog`. This is a user-visible string change, so `include in changelog` feels right, but happy to defer if you'd prefer `exclude from changelog`.
2. **Category**: at least one from the area list — `📺 re_viewer`, `ui`, or `🪳 bug` would all fit (those are the labels on the linked issue).

Could a maintainer add both when you get a moment? Thanks!

emilk:
@rerun-bot reality-sync

lsy3:
Snapshot tests failed because the label change affects the selection panel, which is visible in several full-viewer integration tests. Fixed by running `RUN_ID=25648440668 ./scripts/update_snapshots_from_ci.sh` per CONTRIBUTING.md and committing the updated PNGs.

PR #12767, opened 2026-05-13
Expose remaining blueprint view types in the Rust SDK

FredrikNoren:
## Summary

The Python SDK exposes 11 blueprint view types, but the Rust SDK in `re_sdk::blueprint` only exposes 6 of them. This adds the missing ones:

- `BarChartView`
- `DataframeView`
- `StateTimelineView`
- `TensorView`
- `TextLogView`

Each follows the existing minimal builder pattern used by `Spatial2DView`, `TimeSeriesView`, etc. — `with_origin` / `with_contents` / `with_visible` / `with_defaults` / `with_override(s)` — and gets a `From<…> for ContainerLike` impl so they can be placed inside `Vertical` / `Horizontal` / `Grid` / `Tabs`.

The original motivation was that `TextLogView` was unreachable from a downstream Rust crate building a blueprint — the others have the same gap.

View-specific property setters (e.g. `TextLog` columns/rows/format_options, `Dataframe` query options, `TimeSeries` axes, etc.) are *not* exposed in the Rust SDK for any view yet — that's a separate, larger gap.

## Test plan

- [x] `cargo clippy -p re_sdk` passes clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/top/re_sdk/src/blueprint/container.rs
+++ b/crates/top/re_sdk/src/blueprint/container.rs
@@ -365,6 +365,36 @@ impl From<crate::blueprint::TextDocumentView> for ContainerLike {
     }
 }
 
+impl From<crate::blueprint::TextLogView> for ContainerLike {
+    fn from(view: crate::blueprint::TextLogView) -> Self {
+        Self::View(view.0)
+    }
+}
+
+impl From<crate::blueprint::BarChartView> for ContainerLike {
+    fn from(view: crate::blueprint::BarChartView) -> Self {
+        Self::View(view.0)
+    }
+}
+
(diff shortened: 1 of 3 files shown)


Wumpf:
oof we really need codegen for this. someone else invoked an agent to do that over [here](https://github.com/rerun-io/rerun/pull/12841) but it's a lot to take in. this one here I can at least reason with quickly. Will probably do for the very moment, but should be replaced soon

Wumpf (approved):

Wumpf:
@rerun-bot reality-sync

